PR #3005, opened 2026-08-01
fix(agent/backup): correct IVssBackupComponents vtable indices so VSS snapshots work (#2999)

ToddHebebrand:
Refs #2999

## Root cause — proven, not inferred

`InitializeForBackup` was mapped to **vtable slot 3**. In `IVssBackupComponents`, slots 3 and 4 are `GetWriterComponentsCount` and `GetWriterComponents`; **`InitializeForBackup` is slot 5**. So the agent was calling `GetWriterComponentsCount(OUT UINT*)` and passing it the NULL it intended as an optional BSTR — a textbook `E_INVALIDARG`, synchronous and sub-millisecond, identical on every machine. Raw COM vtable dispatch means neither the compiler nor the linker can catch this.

Verified directly against the VSS service on a Windows Server 2022 box with a standalone probe:

| call | result | conclusion |
|---|---|---|
| `vtable[3](NULL)` | `E_INVALIDARG` in 0 ms | reproduces the field failure exactly |
| `vtable[3](&uint32)` | `VSS_E_BAD_STATE` | validates an out-param *and* wants prior init → not `InitializeForBackup` |
| `vtable[5](NULL)` | `S_OK` | this is `InitializeForBackup` |

That accounts for every observed detail: components created fine, next call rejected outright, single-digit milliseconds, two unrelated machines, two Windows editions.

## Four more defects on the same path

Correcting the indices moved the failure downstream and exposed these, each also confirmed against the live service:

1. **`StartSnapshotSet` was never called.** `AddToSnapshotSet` requires it; without it every add fails `VSS_E_BAD_STATE`.
2. **`AddToSnapshotSet`'s `ProviderId` was passed as literal `0`.** `VSS_ID` is a 16-byte struct passed *by value*, which the x64 ABI marshals as a hidden pointer — it must be the address of a zeroed GUID.
3. **The caller's volume string is rejected by VSS.** `backup.go` derives volumes with `filepath.VolumeName`, producing `"C:"` with no trailing separator. Probed: `AddToSnapshotSet("C:")` → `VSS_E_OBJECT_NOT_FOUND`. **VSS would still have failed after the index fix.** Now normalised to a mount point (`C:\`) for the COM call, while the session map stays keyed on the caller's original string — re-keying it would make `rewritePathsForVSS`'s lookup miss and silently read the live volume.
[...]

PR #3004, opened 2026-08-01
fix(agent): bound the backup result payload to the IPC frame (#3001)

ToddHebebrand:
Refs #3001

## Problem

A backup run's terminal IPC result is unbounded; the frame is capped at `ipc.MaxMessageSize` (16 MiB, `agent/internal/ipc/message.go:105`, enforced at `protocol.go:142`). The field-reported run — 123,284 files, 35.4 GB, 317 per-file errors — produced a **64,405,054 byte** result. `conn.Send` rejected it, the result was dropped outright, and the `backup_jobs` row stayed `running` until the stale-backup reaper failed it 15 minutes later. **The backup had succeeded.**

Perverse scaling: the more files a run protects, the more likely its outcome never arrives.

## Correction to the issue's stated root cause

The issue attributes the 64 MB to the 317-entry per-file failure list. **That is not possible** — `Snapshot.UploadFailures` is `json:"-"` (`agent/internal/backup/snapshot.go:77`) and never reaches the wire. 317 entries would be ~63 KB regardless.

The dominant field is the **per-file snapshot manifest**, `Snapshot.Files` — one `SnapshotFile` per backed-up file (absolute source path + snapshot-prefixed backup path + SHA-256 checksum + modTime). 64,405,054 / 123,284 = **522 bytes/entry**, which matches that shape exactly. `TestUnboundedRunResultExceedsIPCLimit` pins both halves: the full job exceeds the cap, and removing only the file index brings it back under.

`summarizeUploadFailures` was already doing its job — the warning string was never the problem.

## Fix

`fitBackupResultToIPC` (new `agent/cmd/breeze-backup/result_bounds.go`) degrades in tiers, stopping at the first that fits:

1. **Always, and cheap (no JSON parse): truncate `stderr`.** This is the genuinely unbounded text field on this path — a hard failure routes `errors.Join(<every per-file error>)` straight through `fail(err.Error())`, which nothing capped.
2. **Drop `snapshot.files` whole; cap the `warning` text.** Dropped whole rather than truncated on purpose: the server derives `hasIndexedFiles` from `snapshot.files.length` (`backupResultPersistence.ts:537`), so a truncated index would present as a *complete* browsable file list that is silently missing entries. The drop is appended to the result's `warning`, which the server persists to the job's `errorLog` — so it is visible in the UI, not just in a log line on the endpoint.
3. **Reduce stdout to the terminal-status identity fields** (job id, status, snapshot identity, counters, `errorCount`).
[...]

PR #3003, opened 2026-08-01
fix(agent): report backup helper death instead of waiting for the stale reaper (#2998)

ToddHebebrand:
**Refs #2998**

## Problem

When `breeze-backup.exe` dies mid-run the agent detects it in ~2 seconds via IPC EOF, and tells nobody. The `backup_jobs` row stays `running` until the stale-backup reaper fails it 15 minutes later with *"no progress reported for 15 minutes"* — which describes a wedged upload, not a dead process.

From the live repro in the issue (Windows Server 2022, 0.103.0, 2026-08-01):

```
21:42:56  breeze-backup.exe killed  -> job stuck at 6454/48000 files
21:58:10  reaped: "[stale-backup-reaper] Backup stalled: no progress reported for 15 minutes"
```

Agent-side the death was already unambiguous two seconds after the kill — at INFO, on disk only:

```
14:42:54 level=INFO msg="session recv loop ended"  component=sessionbroker error="ipc: read header: EOF"
14:42:54 level=INFO msg="user helper disconnected" component=sessionbroker sessionId=backup-6984
```

The backup branch of the post-disconnect cleanup in `handleConnection` cleared local state only — no path from there to a terminal `backup_result` on the WebSocket. The helper is not respawned, so the run is definitively over at that point.

## Fix (agent-only, `internal/sessionbroker`)

The broker tracks every **async `backup_run`** the helper is executing. If the session ends with any of them still in flight, the disconnect synthesizes a failed `backup_result` per run with `stderr = "backup helper exited unexpectedly"`, pushed through the same `onMessage` sink a genuine unsolicited result uses — so it inherits the heartbeat's delivery path, including the outbox that survives a WS gap. It logs at **WARN with `component=backup`**, so it ships at the default `log_shipping_level: warn` instead of sitting on disk at INFO.

Server-side, `backupResultPersistence.ts:454` writes a failed result's `error` into `backup_jobs.error_log`, so the run carries the accurate cause within seconds.

### Exactly-once, by construction

The hard part is not detecting the death — it is making sure each command gets **exactly one** terminal outcome when the forwarding goroutine and the RecvLoop goroutine race. Each run carries a state:

| State | Meaning | Who owns failing it |
|---|---|---|
| `pendingAck` | recorded **before** the send; ack not yet seen | the forwarder — `SendCommand` errors when the session closes |
| `executing` | the helper acked; terminal result still owed | the disconnect path |
[...]

PR #3002, opened 2026-08-01
fix(agent/backup): skip permanently-failing files without the 30s retry backoff (#2997)

ToddHebebrand:
Refs #2997

## Problem

`agent/internal/backup/snapshot.go` classified a failed per-file upload on exactly one axis — "was this a job cancel?". Every other error was assumed transient, got a 30 s sleep, and one retry:

```go
if uploadErr != nil && !errors.Is(uploadErr, errBackupStopped) {
    log.Warn("file upload failed, retrying once", ..., "retryDelayMs", 30000)
    case <-time.After(uploadRetryDelay):   // uploadRetryDelay = 30s
```

Many of those errors are permanent by construction, so the retry could never succeed and the run paid 30 s per file for nothing.

**Field evidence** (0.103.0, real customer `C:\Users` run, 123,600 files):

| | |
|---|---|
| Run elapsed | `elapsedMs=9669582` — **2 h 41 m** |
| Files skipped after retry | **316** |
| Retry sleep | 316 x 30 s = **9,480 s (2 h 38 m)** |
| Actual work | **~3 minutes** |

**98% of the run was the backoff.** Error breakdown across that log:

```
254  ERROR_SHARING_VIOLATION    live browser caches (Chrome/Edge Cache_Data, GPUCache, WebView2 leveldb)
 54  ERROR_LOCK_VIOLATION       "another process has locked a portion of the file"
 14  ERROR_FILE_NOT_FOUND       existed at scan, gone by upload
  2  ERROR_PATH_NOT_FOUND
  2  genuine S3 UploadPart/PutObject failures  <- these SHOULD still retry
```

The originally-reported OneDrive case is the same bug with a different trigger: 21 cloud-only placeholders took 637 s to back up 85 bytes, because the backup helper runs as SYSTEM and cannot hydrate a user's placeholder.

## Change

`classifyPermanentUploadError` now gates the backoff. Permanent → log and skip immediately, no sleep, no retry. Everything else keeps the existing 30 s retry.

**Only the delay goes away.** A permanent failure falls through to the exact same skip-and-continue block: appended to `Snapshot.UploadFailures` (and so counted in `job.ErrorCount`), job carries on. Nothing about what the run *reports* changes.

Permanent set (exactly the five in the issue thread):

| Code | Name | Why it can't recover |
|---|---|---|
| 32 | `ERROR_SHARING_VIOLATION` | live process holds the file; not clearing in 30 s |
[...]

PR #3012, opened 2026-08-02
feat(backup): add a `partial` terminal status for disproportionately failed runs (#3000)

ToddHebebrand:
Refs #3000

## The gap

A backup run reported `completed` no matter how much of it failed. Field case: **21 of 22 files failed and the job still landed green** with 85 bytes stored out of 3.2 MB.

Partial success is a **deliberate** design and the failures were never silently swallowed — `errorLog`, `errorCount` and the job-list error summary all worked, and the code comment at `backup.go` already named the risk ("a green job that is an incomplete restore point"). None of that is ripped out here.

What was missing is **proportionality**: a run that stored essentially nothing carried the same terminal status as a clean one.

## Which option, and why

The issue offered two. **This PR implements (a), a `partial` terminal status — and uses (b)'s ratio threshold as the trigger for it.** They are complementary, not alternatives: the threshold supplies the proportionality, the distinct status supplies the honest label.

Two findings made (a) much cheaper than the issue assumed:

1. **`partial` is already a value of the pg `backup_status` enum** (`0001-baseline.sql:237`) and of the Drizzle enum (`db/schema/backup.ts:38`). It was simply unreachable — nothing ever wrote it. **No migration.**
2. **`status` is already classified `included`** in `CORE_TENANT_EXPORT_POLICY` for `backup_jobs`. No new column is added, so **no export-policy or cascade registration change** is required (that row fires on a new *column*; there is none).

Why not (b) alone: promoting to `failed` would be a lie in the other direction. A partial run produced a **real, restorable snapshot**. Anything that reads `failed` as "no backup happened" — restore pickers, retry logic, the orphan-snapshot guard — would then be wrong about a snapshot that genuinely exists. `partial` is the honest state, and a binary green/red misrepresents one side or the other.

## Threshold design

`classifyCompletionStatus` (new `agent/internal/backup/completion_status.go`) downgrades a run only when the loss is disproportionate. Two independent gates, either of which trips:

- **BYTES (primary)** — >10% of scanned bytes missing from the snapshot. Per the issue's steer, bytes beat file count: a run that uploaded 99 of 100 files but missed the single 90 GB database is not a usable restore point.
- **FILES (secondary)** — >10% of attempted files failed. This exists because the byte gate is **blind to collection failures**: a file whose `os.Stat` failed has no known size, so it contributes nothing to the denominator and would otherwise be invisible no matter how many there were.

Both comparisons are strictly-greater-than, so a run sitting *exactly* on the threshold keeps the status it had before this change. Zero failures short-circuits, so a clean run can never be downgraded by byte arithmetic.

10% is calibrated against the two field cases in the issue: **317 of 123,600 (0.26%) stays `completed`**; **21 of 22 / 85 bytes of 3.2 MB becomes `partial`**. It is a constant rather than per-policy config on purpose — a tunable needs a policy column, UI and fan-out story, and there is no evidence yet that one threshold does not fit.

## Interaction with PR #3004 (verified, not assumed)

I read `agent/cmd/breeze-backup/result_bounds.go` on `fix/3001-backup-result-ipc-oversize` before designing this. The degradation preserves fields **by shape, not by an enumerated value list**:

- **Tier 3** (`reduceToScalars`) keeps every top-level key whose value is not a JSON container.
- **Tier 4** (`lastResortStdout`) keeps the keep-list `{id, jobId, status, snapshotId}`, `json.Unmarshal`-ing each into a plain Go `string`.

So a **new string value for `status` survives all four tiers** — nothing in `result_bounds.go` enumerates or validates the *value* of `status`.

The important corollary, and a **deliberate design constraint on this PR**: tier 4's payload is a `map[string]string`, so **a new numeric counter would be silently dropped there**. This PR therefore adds **no new wire field at all** — the partial signal rides the existing `status` string and the existing `errorCount`. `TestBackupJobPartialStatusSurvivesResultDegradationShape` pins both halves: that `status` marshals as a top-level JSON string scalar, and that no unexpected field has appeared on the payload. It asserts the shape property on the *producing* side so it holds regardless of merge order with #3004 (whose file does not exist on this branch).

Also verified: this sets `Success: true` and leaves `Stderr` empty, so it does not trip `heartbeat.go`'s `if Stderr != ""` branch, which would discard the entire stdout body.

**Merge-conflict risk with the four in-flight PRs: none textual.** #3002, #3003, #3004 and #3005 collectively touch `agent/cmd/breeze-backup/*`, `agent/internal/backup/snapshot.go`, `agent/internal/backup/vss/*` and `agent/internal/sessionbroker/*`. None touches `agent/internal/backup/backup.go`. Worth flagging that **#3002 raises the frequency of `errorCount > 0` completions**, which is exactly the population this PR reclassifies.

## The plumbing problem this surfaced

**The server never read the agent's terminal status at all.** `backupCommandResultSchema` had no `status` key (and is non-passthrough, so it was stripped); the persisted status came solely from the outer `websocket.CommandResult.Status`, which `heartbeat.go` derives from a success *bool*.

The agent's status now reaches the server on a **distinct `agentStatus` key** through the queue, because `status` on that payload already means the outer completed/failed value — `backupWorker.processResults` reads `data.result.status` as the outer one. Passing them under one name would have silently made `partial` unreachable on the queued path. `backupProcessResultSchema` is `.strict()`, so the key is declared there too.

## Consumer audit

`partial` was already accepted in `BackupJobStatus`, `IN_FLIGHT_BACKUP_JOB_STATUSES` (correctly excluded), `backupProgress`'s terminal-status guard, and the `query_backups` AI tool. The reachable-but-wrong consumers fixed here:

| Where | Was |
|---|---|
| `backupResultPersistence.ts` | a `partial` result fell into the `else` and was recorded **`failed`**, and would have **created no `backup_snapshots` row** — stranding a restorable snapshot in the bucket. It now takes the entire success path; only the status value differs. |
| `BackupJobList.normalizeStatus` | `partial` matched no branch and fell through to `queued` — a finished backup rendered as **"Queued" with a clock, forever**. |
| `DeviceBackupTab` | `?? jobStatusConfig.pending` mislabeled it **"Pending"**. |
| `BackupOverviewContent` | indexed `statusConfig[...]` **unguarded** then read `.icon` — would throw and blank the overview on an unmapped key. Now falls back. |
| `BackupDashboard` success rate | partial was excluded from **both** sides of the fraction, so a device whose every run is partial read as having no runs. Now in the denominator, never the numerator. |
| `dashboard.ts` attention items | keyed strictly on `failed`, so nothing ever fired for a partial run. Now included (`cancelled` deliberately still excluded — user action, not a fault). |
| `jobListSchema` | `?status=partial` 400'd. |

Plus the issue's explicit ask: **the failed-file count now renders on the COLLAPSED row** (`data-testid="backup-job-failed-count"`) for any job with `errorCount > 0` — including `completed` ones — not only inside the expanded detail.

## Deliberate behavior changes worth a maintainer's eye

**`partial` does NOT count as a successful backup.** Six independent "last successful backup" definitions (`dashboard.ts` `lastSuccess`, `backupSlaWorker` ×3, `aiToolsBackup`, `readinessCalculator`, `verificationScheduled`) all remain `=== 'completed'`, unchanged. Consequences, all intended:

- **An RPO/SLA breach will now fire** for a device whose backups are partial. That is the issue's own complaint ("any alerting keyed on job status will not fire") answered — but it is real alerting blast radius on the day this ships, and #3002 will make partial runs more common. Flagging it explicitly rather than burying it.
- Partial jobs are no longer auto-selected for scheduled **verification** and no longer raise **recovery readiness**. Conservative, and arguably correct for a knowingly-incomplete snapshot, but it is a narrowing of an existing population.

Two existing agent tests changed their expected status (1-of-2 file failure, 1-of-2 unreadable path — both 50%, both genuinely disproportionate). Every other assertion in them is untouched, and `TestRunBackupContext_SmallFailureRatioStaysCompleted` is the new below-threshold counterpart proving `completed` still happens.

Not fixed (pre-existing, out of scope): `jobListSchema` still accepts the legacy misspellings `queued`/`canceled`, which match no enum value; `ACTIVE_BACKUP_JOB_STATUSES` duplicates `IN_FLIGHT_BACKUP_JOB_STATUSES`; and `DeviceBackupTab`/`backupDashboardHelpers` status labels are hardcoded English for *all* statuses.

## Testing

- `cd agent && go test -race ./... && go vet ./...` — **clean**.
- `pnpm --filter @breeze/web test` — **4752 passed, 1 failed**: `billing/InvoiceWorkspace.test.tsx`, which **passes in isolation** and is a load-induced flake under the full parallel run. This PR touches no billing file.
- API backup surface, run serially (`--maxWorkers=1`): 23 files / **386 tests green** — `routes/backup/**`, `backupResultPersistence`, `backupProgress`, `backupWorker`, `queueSchemas`, `agentWs`.
- `tsc --noEmit` clean for both `apps/api` (needs `NODE_OPTIONS=--max-old-space-size=12288`) and `apps/web`.

Tests prove the four required properties: a clean run is still `completed`; a near-total failure is not; the boundary (exactly-at-threshold vs one unit past, on both gates) behaves; and the status survives the #3004 degradation shape.

**RLS/integration contract suites were NOT run** (they need a live DB) — and are **not applicable**: no table, no column, no policy and no cascade/export-policy entry changed. The only DB-facing change is writing an enum value that already existed.


ToddHebebrand:

**Findings:** 1 critical, 6 important, 9 suggestions → all consequential ones addressed in 3d1c0fd3f; 4 deferred with reasons below.

**The critical one was real and unanimous** (all four agents found it independently): `dashboard.ts` selected the new `partial` counter into `last24hStats` and then **omitted it from the response body**, so the web success-rate fix was inert in production. It survived my own testing because the API test asserted only `attentionItems` while the web test hand-mocked a `jobsLast24h` shape the API never sent — both green, against different fictional contracts. Fixed, with an API-side round-trip assertion.

**Addressed**
- **Restore-point sweep.** New `RESTORABLE_BACKUP_JOB_STATUSES` applied to `backupSlaWorker` (×3), `aiToolsBackup`, `readinessCalculator`, `verificationScheduled` and the device-status `lastSuccess`. This settles a decision the PR was previously making by omission — see the note in the PR body; the deciding factor is that excluding `partial` makes `missed_backup` assert something false *and* leaves the breach permanently unresolvable, since auto-resolution keys on the same query.
- **`agentStatus` forwarded on the three provider-backed call sites** (agentWs provider handler, hyperv, mssql). Inert today, but those paths parsed the agent status and dropped it.
- **Unrecognized agent statuses now collapse loudly** (warn + Sentry) instead of silently greening. `skipped` — a run that protected zero files — reaches that branch today.
- **Real degradation-tier coverage.** #3004 merged while this was in review, so I rebased and replaced the shape-proxy test with `partial_status_bounds_test.go`, which drives a partial result through the **actual** `fitBackupResultToIPC` tiers (undegraded, oversize-manifest, tier-4 last resort).
- **Wire-plumbing tests**, the seam the test analyst rated highest-risk: both transports asserted. **Verified by mutation** — deleting either `agentStatus` line turns both red, which was not previously true of any test.
- Consecutive `partial` runs stay `warning` rather than escalating to `critical` like real failures; `?status=queued`/`canceled` no longer 500 on a pgEnum comparison; docs updated; four comment inaccuracies corrected (the byte-gate guard is for readability, not a clamp; the 317-of-~123,600 case is from #3001, not #3000).

**Deferred, deliberately**
- **An entire failed backup root counts as one file and zero bytes** — an unmounted `D:\` alongside a small `C:\Users` can still report `completed`. Genuinely the same bug class as #3000, but it needs the collection pass to distinguish a failed root from a failed leaf. Called out in the issue as a follow-up rather than grown into this PR.
- A degraded system-state capture sets `Warning` but not `ErrorCount`, so it bypasses the gate; GFS tagging is status-agnostic so a partial snapshot can claim a retention slot; an unknown *future* status still renders as "Queued" in the web list. All pre-existing, none introduced here.

**Tests**
- `cd agent && go test -race ./... && go vet ./...` — clean.
- API backup surface, serial (`--maxWorkers=1`): 24 files / **395 tests green**.
- Web: `src/components/backup/` + `src/lib/i18n` — 22 files / **181 tests green**. `tsc --noEmit` clean for `apps/api` and `apps/web`.
- Full `pnpm --filter @breeze/api test`: 19,154 passed, **4 failed — none from this PR**. `noRawEnvNumberCoercion` is a load-induced timeout that passes in isolation; the three `autoMigrate.test.ts` ordering failures are **already red on `main`**, caused by `2026-08-06-g-drop-custom-alert-conditions.sql` from #2995. Flagging separately — that is a required job red on main right now, independent of this branch.
- No RLS/integration contract suite was run, and none applies: no table, no column, no policy, no cascade or export-policy entry changed. The only DB-facing change is writing an enum value that already existed.

**Status:** review-clean, awaiting maintainer merge. Not merging, and #3000 left open for the reporter to verify against the Windows Server 2022 repro.

Issue #3007, opened 2026-08-01
[Agent][IPC] Conn.Send assigns the sequence number outside the write lock — concurrent sends are rejected as replays and the message is lost


ToddHebebrand:
## Description

`Conn.Send` assigns the IPC sequence number **outside** the mutex that serializes the actual socket write. Two concurrent `Send` calls on one `Conn` can therefore acquire sequence numbers in one order and write to the socket in the opposite order. The receiver enforces monotonicity and rejects the out-of-order frame as a replay, killing a legitimate message.

Found while verifying the concurrency fix on PR #3003 (#2998) and reproduced directly with two concurrent `Send`s on a single `Conn`:

```
ipc: sequence number 1 <= last 2 (replay/duplicate)
```

## Root cause

`agent/internal/ipc/protocol.go`:

```go
134:    env.Seq = c.sendSeq.Add(1)      // <-- assigned here, no lock held
135:    env.HMAC = c.computeHMAC(env)
137:    data, err := json.Marshal(env)  // marshal also outside the lock
...
146:    c.mu.Lock()                     // <-- writes serialized only from here
147:    defer c.mu.Unlock()
```

`sendSeq` is atomic, so the numbers are unique — but uniqueness is not the invariant the receiver checks. It checks **ordering**:

```go
246:    return nil, fmt.Errorf("ipc: sequence number %d <= last %d (replay/duplicate)", env.Seq, prevSeq)
```

Goroutine A takes seq 1 and goroutine B takes seq 2; if B reaches `c.mu` first, the wire order is 2 then 1, and frame 1 is rejected. The window is the marshal plus HMAC computation, which widens with payload size — so **large messages are the most likely to lose this race**.

## Impact

This is not theoretical for the backup path. Several producers can write to the same `Conn` concurrently:

- concurrent backup run reporting (exactly the production fan-out PR #3003 models)
- `emitVaultAutoSyncResult`, which runs immediately after a backup over the same file set (noted on #3001)
- progress/keepalive pings overlapping a terminal result

A rejected frame is not retried. Depending on which message loses, the consequence ranges from a dropped progress update to a **lost terminal result**, which puts the job into the same orphaned state as #3001 and #3006: work completed, nothing recorded, reaper failure 15 minutes later.

The HMAC is computed over the envelope including `Seq`, so the frame is well-formed and correctly authenticated — it is discarded purely on ordering. That makes it look like tampering in logs when it is our own race.

## Suggested fix

[...]

Issue #3001, opened 2026-08-01
[Agent] Successful backup reported as stalled — 64MB result payload exceeds the 16MB IPC cap, result is dropped, reaper fails the job


ToddHebebrand:
## Description

A backup that accumulates enough per-file errors produces a final result message too large for the agent IPC channel. The send is rejected, the result is dropped, and the `backup_jobs` row is left `running` — so a **fully successful backup is reported to the user as a stalled/failed job** by the stale-backup reaper 15 minutes later.

Confirmed from a self-hosted operator's `backup.log` on `0.103.0`, 2026-08-01.

## Evidence

The run completed successfully:

```
16:36:03 level=INFO msg="backup run completed" component=backup
         jobId=job-20260801T125454Z-1f91465e
         snapshotId=snapshot-20260801T125517Z-5edfcd7e
         filesBackedUp=123284 bytesBackedUp=35421941515
         referencedFiles=121798 referencedBytes=34929299357
         errorCount=317 elapsedMs=9669582
```

123,284 files and 35.4 GB stored. One second later:

```
16:36:04 level=ERROR msg="failed to send final backup result"
         commandId=822e0c7f-7e35-43e6-b0fc-0912a5c0d221
         error="ipc: message too large: 64405054 > 16777216"
```

**64,405,054 bytes against a 16,777,216 byte (16 MiB) cap.** The result is discarded. Nothing else is attempted — no truncation, no retry, no fallback. From the server's perspective the job simply stops reporting, so `last_progress_at` freezes at 16:36 and the reaper fails it at ~16:51 with `Backup stalled: no progress reported for 15 minutes`.

The user is told their backup stalled. It did not — it finished, and the data is in their storage. They just have no restore point recorded, because `providerSnapshotId` is only persisted on a `completed` result (`apps/api/src/services/backupResultPersistence.ts:515`).

## Root cause

The result payload embeds the full per-file failure list — 317 entries, each carrying an absolute path plus the Windows error string. That alone accounts for essentially all of the 64 MB.

`summarizeUploadFailures` (`agent/internal/backup/backup.go:659`) exists and caps what goes into the *Warning* string, but the oversized field is the structured result, which is not bounded. So error volume scales the payload without limit while the IPC frame size is fixed.

Note the perverse scaling: **the worse the backup goes, the more likely it is that the success/failure report never arrives.** A clean run always reports; a run with hundreds of errors silently vanishes.

## Suggested fix

[...]

Issue #2273, opened 2026-07-07
macOS user helper keepalive-evicted every ~60s → Connect Desktop greyed out (desktopAccess unavailable)


ToddHebebrand:
## Summary

On macOS, the **user-session helper** (`com.breeze.agent-user`, uid 501) is repeatedly evicted by the daemon's session broker with `keepalive pong timeout, closing stranded session` roughly every 20–40 minutes, all day. Because a healthy user-session helper is a prerequisite for `computeDesktopAccess` to return `mode: "user_session"`, the device's `desktopAccess` stays `unavailable`, and the web UI renders **"Connect Desktop" greyed out** ("Desktop Unavailable") even though the device is online and in active use.

## Impact

Remote desktop is effectively unusable on affected Macs — the button never enables, with no clear reason surfaced to the technician. Observed on agent `0.92.0`, macOS `darwin 26.5.2` (laptop).

## Evidence (daemon `sessionbroker` logs)

```
keepalive pong timeout, closing stranded session   ageMs≈60000  identity:"501"
keepalive pong timeout, closing stranded session   ageMs≈59988  identity:"501"
failed to send pong ... write unix .../Breeze/agent.sock: write: broken pipe   uid:501
... (recurs ~20–40 min apart across the entire day)
```

`ageMs ≈ 60000` on every eviction is diagnostic: broker pings every `keepalivePingInterval` (30s) and evicts when `LastPongAge > keepaliveTimeout` (45s), checked on the 30s tick — so an unresponsive helper is killed at the 2nd tick (~60s). See `agent/internal/sessionbroker/broker.go` `runKeepalive` (~L2217–2280).

## Root cause analysis

**Amplifier (definite bug): `ipc.Conn.Send()` holds the write mutex across blocking writes with no write deadline.**

`agent/internal/ipc/protocol.go` (~L105–116):

```go
c.mu.Lock()
defer c.mu.Unlock()
// ... no SetWriteDeadline ...
if _, err := c.conn.Write(header); err != nil { ... }
if _, err := c.conn.Write(data); err != nil { ... }
```

[...]

PR #3013, opened 2026-08-02
fix(agent/ipc): assign the IPC sequence number under the write lock (#3007)

ToddHebebrand:
## Problem

`Conn.Send` assigned the sequence number **outside** the mutex that serializes the socket write:

```go
env.Seq = c.sendSeq.Add(1)      // no lock held
env.HMAC = c.computeHMAC(env)
data, err := json.Marshal(env)  // marshal also outside the lock
...
c.mu.Lock()                     // writes serialized only from here
```

`sendSeq` is atomic, so the numbers are **unique** — but uniqueness is not the invariant the receiver enforces. `Recv` enforces **ordering**:

```
ipc: sequence number %d <= last %d (replay/duplicate)
```

Goroutine A takes seq 1, goroutine B takes seq 2; if B reaches `c.mu` first the wire order is 2 then 1, and frame 1 is rejected as a replay. The frame is well-formed and correctly HMAC'd — it is discarded purely on ordering, so it reads in the logs like tampering when it is our own race.

The window is the HMAC plus the marshal, so it **widens with payload size** — exactly backwards from what you want, since the large messages are the important ones. A rejected frame is not retried; losing a terminal result puts a backup job in the same orphaned state as #3001/#3006 (work done, nothing recorded, misleading reaper failure 15 minutes later).

Several producers share one `Conn`: concurrent backup run reporting, `emitVaultAutoSyncResult` right after a backup, and progress/keepalive pings overlapping a terminal result.

## Fix

Move the sequence assignment, the HMAC and the marshal inside `c.mu`, so the number a frame carries and the order it reaches the socket are decided under one lock.

**`sendSeq` stays `atomic.Uint64`** — deliberately, and its comment now says why: it is atomic *only* so tests can observe it, and it must be advanced with `mu` held. Demoting it to a plain `uint64` would make the targeted regression test's spin-wait a data race that `-race` would flag, and the atomic RMW is free next to the work it now guards.

Also **re-checks `writePoisoned` after acquiring the lock**. The pre-lock check is kept as a fast path so a dead `Conn` fails without contending, but a `Send` that cleared it could queue behind a writer that poisoned the `Conn` and then append a frame to a stream a partial write had already desynced (#2273). Extracted into `poisonedErr()` so both checks share one definition.

## Marshal stayed inside the lock — and why

The alternative was an ordered handoff (take the seq under the lock, marshal outside, write in seq order via a ticket). **Rejected**, on two grounds:

1. **It buys little.** Writes must happen in a total order regardless, so the handoff only pipelines CPU work against a single write. A slow early ticket still head-of-line-blocks everyone.
2. **It costs a lot.** A sender that abandons its ticket — marshal error, the oversize rejection, a panic, the poison path needing a broadcast — wedges every later sender, with **no deadline to bound the wait**. Socket blockage is bounded by the 30s write deadline; an abandoned ticket is not. That is strictly worse than the hazard #2273 removed, on the very path #2273 was about.

Independently reviewed via Codex (read-only), which reached the same recommendation and flagged the poison-recheck gap now fixed above.

### Measured, not assumed

`BenchmarkConnSendConcurrent` (added), M4 Pro, before vs. after:

|          | `-cpu=1` (uncontended) | `-cpu=14` (saturated) |
|----------|------------------------|------------------------|
| 256 B    | 33.3 → 30.7 µs — none  | 39.4 → 43.9 µs (~+11%) |
| 64 KiB   | 294 → 265 µs — none    | 98 → 304 µs (~3×)      |
| 1 MiB    | 4.62 → 4.37 ms — none  | 0.53 → 4.30 ms (~8×)   |

**Uncontended cost is nil** — the fix adds no work to a `Send`. The saturated large-payload column is the entire cost, and it comes from the old code letting N goroutines marshal on N cores while serializing only the write, which beat the serial ceiling.

That advantage is **unreachable on this transport**, because large and high-rate never co-occur on it:

- Large messages — script `command_result` (~1–3 MiB), screenshots (capped 1 MB) — are request/response, seconds to minutes apart.
- High-rate messages — `backup_restore` per-file progress, 30s keepalives — are 200–500 byte structs whose marshal is nanoseconds.
- **Remote-desktop video does not use `ipc.Conn` at all.** Frames go over WebSocket/WebRTC; `TypeDesktopFrame` is declared and never sent, and `capture_proxy.go`'s raw-RGBA-over-IPC path has zero callers.

### The 64 MB case — review changed this

My first pass left the oversize rejection where it was, arguing the cost was "bounded by `MaxMessageSize`". **Review correctly showed that claim was false**, and it is the one thing worth reading this section for: the size check runs *after* the marshal, so the work now inside the lock was bounded by whatever the **caller** passed, not by `MaxMessageSize`. A caller handing `Send` a 64 MB (or 500 MB) payload would hold `c.mu` across its hash and marshal and only *then* be rejected — starving the keepalive pong, which is precisely the #2273 eviction failure.

So there is now a **pre-lock guard** on `len(env.Payload)`, before `c.mu` is taken. An oversized payload is rejected off the lock entirely and without consuming a sequence number, which restores the bound the comment claims. The post-marshal check stays, to catch an envelope whose framing overhead pushes a just-under-limit payload over.

I had originally rejected this guard because `json.Marshal` *compacts* a `RawMessage`, so a payload above the limit could in principle compact below it. That is real but not reachable in practice — every production payload arrives already compact from `json.Marshal` (`SendTyped` does exactly that) — and it is plainly the lesser risk next to an unbounded marshal under the write mutex.

## Tests

Two new tests in `agent/internal/ipc/protocol_concurrency_test.go`:

- **`TestConnSendSeqAssignedUnderWriteLock`** — near-deterministic reproduction. A 2 MiB send claims a seq and then spends milliseconds hashing/marshalling; a small send started after that claim (synchronised by spin-waiting on `sendSeq`, not by a sleep) overtakes it. **Failed 20/20 runs before the fix** with `frame 1: got seq 2, want 1`.
- **`TestConnSendConcurrentSendersAllAccepted`** — the broad `-race` guard: 8 goroutines × 25 messages on one `Conn`, receiver must accept all 200 in strictly increasing order. **Failed 3/3 runs before the fix**, e.g. `frame 2/200 rejected: ipc: sequence number 1 <= last 2 (replay/duplicate)` — the exact error from the issue.

Both pass after (20× and 10× repeats). The receiver closes its end on rejection so the failure surfaces in ~0.02s instead of blocking senders until the 30s write deadline.

Verified:

```
go test -race ./internal/ipc/... ./internal/sessionbroker/... ./internal/backup/...   # all ok
go vet ./...                                                                          # clean
gofmt -l internal/ipc/                                                                # clean
```

## Windows CI gap

`./internal/ipc` was **missing** from `test-agent-windows`, which uses an explicit package list — so none of its tests had ever run on Windows, despite the transport being a named pipe there and `auth_windows.go` being a platform-specific peer-auth path. Same class of gap #3002 and #3005 found.

Added to **both** the plain and the `-race` step. Verified as the `logging` entry documents: under `GOOS=windows GOARCH=amd64 CGO_ENABLED=0` the package vets and its test binary compiles, and its Windows dependency graph contains no `remote/desktop` and no cgo. Unlike `logging` it is added to the race step too — a concurrency fix is precisely what the detector should cover, and `ipc` already compiles under cgo there as a dependency of `sessionbroker`.

## Scope

Confined to `agent/internal/ipc/` plus the CI package list. No overlap with #3002/#3003/#3004/#3005. Based on current `origin/main` (ad40683).

## Noted, not fixed

`backup_list` inlines the full `Files []SnapshotFile` manifest and triple-encodes it (`internal/backup/snapshot.go:55`, `cmd/breeze-backup/main.go:619`). On a large-fleet device that can exceed the 16 MiB `MaxMessageSize`. Pre-existing and untouched by this PR, but it is the one message on this transport that can plausibly reach the cap — worth its own issue.

Refs #3007


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, parallel). Design choice additionally cross-checked against Codex (read-only), which independently reached the same recommendation.

**Findings:** 10 raised → 10 addressed in `b3a41cb` + `6728bf6` + `f0681bb`; 0 outstanding. The three that mattered:

- **The in-lock cost bound was false.** The oversize check runs *after* the marshal, so the work moved inside `c.mu` was bounded by the caller's payload, not by `MaxMessageSize` — a large payload would hold the write mutex across its hash and marshal and only then be rejected, starving the keepalive pong (#2273). Fixed with a pre-lock `len(env.Payload)` guard; the comment's claim is now true. Raised independently by two agents.
- **Both new branches were untested.** The under-lock poison re-check and the new pre-lock guard each had zero coverage. Now covered and **verified by mutation**: deleting the re-check fails `TestConnSendStalledDoesNotStarveOtherWriter` 3/3; deleting the guard fails `TestConnSendOversizedPayloadRejectedBeforeLock`.
- **`TestConnSendOversizedDoesNotPoison` never tested a size check.** It fills the payload with `'A'`, which is not JSON, so it errors in `json.Marshal` first. Documented in place; the new test uses valid oversized JSON.

Also corrected three inaccurate comments this PR had introduced — a `~5.6 ms/MiB` figure that was an estimate contradicting the PR's own benchmark, a "logged as if tampering" claim (no caller logs that; a `Recv` error is fatal to the whole session, which is *worse*), and a `ci.yml` claim that `-race` would have caught #3007 (it would not — the bug was an ordering bug in race-free code). Dropped the hardcoded benchmark table and rejected-design paragraph from the source; they live in this PR body.

**Tests:** `go test -race ./internal/ipc/... ./internal/sessionbroker/... ./internal/backup/...` green (`-count=2`); `go vet ./...` clean; `gofmt` clean; `GOOS=windows GOARCH=amd64 CGO_ENABLED=0` vet + test-binary compile clean.

Regression evidence, re-confirmed after the review refactors: with the seq assignment moved back outside the lock, `TestConnSendSeqAssignedUnderWriteLock` fails 3/3 (`frame 1: got seq 2, want 1`) and `TestConnSendConcurrentSendersAllAccepted` fails 3/3 (`ipc: sequence number 1 <= last 7 (replay/duplicate)`). Both pass on the fix (20x and 10x repeats).

**CI:** all agent jobs green, including **Test Agent (Windows)** with `./internal/ipc` newly added to both its package lists. `Test API` is red on three `src/db/autoMigrate.test.ts` migration-ordering assertions — **pre-existing on `main`**, which fails the identical three; this PR changes no TypeScript and no migrations.

Two things found next door, deliberately **not** fixed here to keep the change inside `agent/internal/ipc/`:

1. A frame rejected by `Recv` is still lost with no retry, and the teardown logs at **Info** with the same message as a clean EOF (`internal/sessionbroker/session.go:449`) — so a crypto/sequence rejection from any other cause is indistinguishable from a normal disconnect. This PR makes the trigger far rarer but no more visible. Worth its own issue.
2. `backup_list` inlines the full snapshot manifest and triple-encodes it (`internal/backup/snapshot.go:55`, `cmd/breeze-backup/main.go:619`); on a large-fleet device that can exceed the 16 MiB `MaxMessageSize`. Pre-existing, and it is the only message on this transport that can plausibly reach the cap.

Issue #1375, opened 2026-06-15
Returning logins never update users.last_login_at — bare-db UPDATE silently dropped by RLS (platform-wide)


ToddHebebrand:
## Summary

Returning user logins **never update `users.last_login_at`**. The column only ever reflects a user's initial signup/first-login window, then stays frozen forever. This is platform-wide (both EU and US prod), silent, and has been latent for a long time — it only became visible on US because US stopped getting new signups after ~May 10, so its newest `last_login_at` froze at 2026-05-06 while EU kept *looking* healthy purely from a steady stream of new-signup stamps.

## Evidence (prod)

- US `users.last_login_at` max = **2026-05-06**, yet the audit log has `user.login` events as recent as **2026-06-14 21:14** — logins are happening, the column just isn't moving.
- US instance is fully alive: agents checking in through "now", ~34k audit events in 7 days.
- Across **all 80 users on both regions, only 3 have any login recorded >2 days after signup; the largest gap ever is ~4 days.** Returning logins are essentially never recorded anywhere.
- EU only looked fine because it gets daily new signups (each stamps `last_login_at` once at creation). A long-lived EU account's `last_login_at` is stuck at its signup date too.

## Root cause

In `apps/api/src/routes/auth/login.ts:484`, the update runs on the **bare `db`** with no DB access context:

```ts
await db.update(users).set({ lastLoginAt: new Date() }).where(eq(users.id, user.id));
```

`/auth/login` is an **unauthenticated** route, so no `breeze.user_id` / `breeze.org_id` / `breeze.partner_id` GUC is set for the request. The API connects as `breeze_app` (`NOSUPERUSER, NOBYPASSRLS`, confirmed in prod startup logs), and the forced `users` UPDATE policy requires one of `breeze_has_partner_access(partner_id)`, `breeze_has_org_access(org_id)`, or `id = breeze_current_user_id()` to be true. With no context all three are false, so the UPDATE **matches 0 rows**. Postgres treats a 0-row UPDATE as success — **no error is thrown** — and `auditLogin` (which runs through a system context) records the login normally. Hence: audit row written, column silently not updated.

The pre-auth user *lookup* immediately above (`login.ts:249`) is correctly wrapped in `withSystemDbAccessContext`; only the write was left on the bare connection.

## Affected call sites (same bare-`db` self-write bug)

- `routes/auth/login.ts:484` — password login
- `routes/auth/passkeys.ts:352` — passkey login
[...]

Issue #1105, opened 2026-06-05
[api] Mass agent reconnect can poison the Postgres pool (idle-in-transaction timeouts on config_policy_assignments) → heartbeat 500s + login outage


bdunncompany:
## Summary
A large simultaneous agent reconnect (e.g. after a control-plane restart, or a mass agent-service restart) can exhaust/poison the API's Postgres connection pool, causing heartbeat `500`s and a full login/auth outage until load subsides or the API is restarted.

## What was observed (production, with evidence)
- API heartbeats began returning `500` at a steady rate (~80+/min sustained).
- Underlying error, repeated hundreds of times: `Error: write CONNECTION_CLOSED postgres:5432`, originating from a `sql.begin(...)` transaction in a BullMQ worker.
- Postgres was simultaneously logging: `FATAL: terminating connection due to idle-in-transaction timeout`.
- `pg_stat_activity` showed ~30 connections stuck in `idle in transaction`, **all running a `SELECT` on `config_policy_assignments`**, sitting 20–30s until the idle-in-transaction timeout killed them.
- Web login (which needs a DB transaction) failed throughout — even in a fresh/private session — confirming the poisoned pool, not a client issue.
- Total connections were not maxed (≈36/100), so this was connection *poisoning* (stuck idle-in-txn + killed connections cascading into `CONNECTION_CLOSED`), not raw exhaustion.
- `docker compose restart api` immediately cleared it: idle-in-txn → 0, `CONNECTION_CLOSED` → 0, login restored.

## Not yet isolated
I could not pin the exact transaction that holds the `config_policy_assignments` read open across a slow step. Ruled out: the heartbeat route itself, `configurationPolicy.ts` (CRUD-only transactions), and `featureConfigResolver.ts` (autocommit). The `sql.begin` in the stack trace points at a worker, but the workers that use `db.transaction` don't read `config_policy_assignments` in the obvious paths — so it likely needs runtime query tracing to confirm.

## Why it matters
Any operationally-normal event that makes many agents reconnect at once (API redeploy, restart-storm) can take down the API's DB pool and the web UI with it. Worth either narrowing the offending transaction's scope (don't hold a txn across slow work) and/or making the reconnect path more pool-resilient.

ToddHebebrand:
[...]

PR #3023, opened 2026-08-02
fix(api): stop agentWs command_result from double-holding pooled DB connections (#3021)

### LINKED ISSUE #3021: [API] agentWs.commandResult holds an org transaction while opening a nested system context — two pool connections per command result (#2765 fix never reached the WS path)
## Description

Every agent `command_result` message over the WebSocket needs **two pool connections held simultaneously**, because a nested system context is opened inside an already-open org-scoped transaction. This is the same defect class that #2765 fixed for the HTTP `/commands` route; the WebSocket path was never converted.

It shows up in production as a steady drip of held-connection warnings:

```
withDbAccessContext (scope=organization) [agentWs.commandResult] held a pooled
connection in an open transaction for 9860ms (>= 2000ms) — long enough that it
likely did slow non-DB work ... (#1105)
```

59 of them in a 6-hour window in one region, and 155/hr in the other. The hold durations cluster tightly at 9860–9900ms, which is not a coincidence: it is the 10s `connect_timeout` in `apps/api/src/db/index.ts:34` almost expiring while the inner acquire waits.

## Root Cause

`processCommandResult` is invoked inside an org-scoped context:

```ts
await runWithAgentDbAccess('agentWs.commandResult', async () =>
  processCommandResult(agentId, parsed.data, authenticatedAgent.deviceId, authenticatedAgent.orgId)
);
```
`apps/api/src/routes/agentWs.ts:2546-2548`

That wrapper opens a transaction and pins a pooled connection for its whole duration. `processCommandResult` (defined at `agentWs.ts:1696`) then reaches for a *second* connection twice, at roughly `agentWs.ts:1757` and `agentWs.ts:1858`:

```ts
const updatedCommands = await runOutsideDbContext(() =>
  withSystemDbAccessContext(() => ...)
);
```

The trap is that `runOutsideDbContext` does **not** close the outer transaction. It is only an AsyncLocalStorage escape:

```ts
export const runOutsideDbContext: RunOutsideDbContextFn = <T>(fn: () => T): T => {
  return dbContextStorage.exit(() => dbContextMetaStorage.exit(fn));
};
```
`apps/api/src/db/index.ts:416-418`

So the outer org transaction stays open and idle-in-transaction on connection A while `withSystemDbAccessContext` acquires connection B. Under contention the inner acquire waits, the outer connection sits idle burning a pool slot the whole time, and both the hold warning and the eventual timeout follow from that. Two connections per command result, and the second one is acquired at exactly the moment the pool is most stressed.

The comment at `agentWs.ts:1755` already flags the hazard ("same caveat as `services/commandDispatch.ts`"), so this is known-shaped rather than newly discovered — it just was not carried over when #2765 fixed the HTTP twin.

## Proposed Fix

Same treatment as #2765: the command-result path should manage its own short contexts rather than running under a request-long org wrap. Concretely, either drop the outer `runWithAgentDbAccess` for this message type and give each DB operation inside `processCommandResult` its own scoped context, or restructure so the system-context work happens after the org context has actually committed rather than nested inside it.

Worth checking the sibling message handlers in the same `switch` at the same time — `agentWs.backupProgress` (`agentWs.ts:2556`) uses the identical `runWithAgentDbAccess` wrapper and may have the same nesting.

Note this is latent everywhere, not region-specific: both production regions log the holds, but only the busier one has enough concurrency for the second acquire to actually time out. Fixing it removes a multiplier on pool pressure rather than a visible bug in the smaller region.

## Affected Files

- `apps/api/src/routes/agentWs.ts:2546` (the org wrap), `:1696` (`processCommandResult`), `:1757` and `:1858` (the nested system contexts)
- `apps/api/src/db/index.ts:416-418` (`runOutsideDbContext` — the "exits the store, keeps the transaction" behaviour)

## Related

- #2765 — the same fix applied to the HTTP `/commands` route
- #1105 — the held-context / connection-hold class this belongs to
- #2468 — nested system-context escalation on the partner-wide read path, same underlying pattern

**Status:** investigation only, no code change yet.

ToddHebebrand:
## Problem

Every agent `command_result` over the WebSocket held **two pooled DB connections at once**: the message-level `runWithAgentDbAccess('agentWs.commandResult', ...)` wrap opened an org-scoped transaction for the whole message, and the nested `runOutsideDbContext(() => withSystemDbAccessContext(...))` steps inside `processCommandResult` (device_commands terminal CAS, lifecycle recheck, audit write) each acquired a second connection while the org connection sat idle-in-transaction. `runOutsideDbContext` only exits the AsyncLocalStorage — it never releases the outer transaction's connection. This is the WS twin of the HTTP `/commands` poll fixed in #2765, and it surfaces in production as the steady ~9.9s `agentWs.commandResult` held-connection warnings (#1105) in both regions.

## Fix

Drop the message-level org wrap; `processCommandResult` now manages its own short contexts:

- **Already self-managed, now un-nested:** the bare-pool `device_commands` lookup, the `outside(system(...))` terminal CAS, `isAgentDeviceStillAuthorized`, and the fire-and-forget audit write each run with no enclosing org transaction.
- **New explicit system context:** the deviceId-less fallback lookup joins RLS-guarded `devices`; with no ambient context a bare-pool read would silently return 0 rows and misroute the result to the orphaned path. Tenancy stays enforced by the authenticated `agentId` key (same shape as `isAgentDeviceStillAuthorized`).
- **New short org wraps where ambient tenant RLS is genuinely needed:** per-type handler dispatch (`agentWs.commandResult.handler`, both the happy path and the validation-rejected finalization), DR result persistence (`agentWs.commandResult.drResult`), and both orphaned-result branches (`agentWs.commandResult.orphaned`), via a module-level `runWithAgentOrgDbAccess` helper the per-connection closure now delegates to.
- **No direct double-hold remains:** nothing under the new wraps uses the `runOutside(system(...))` combo directly (`withSystemDbAccessContext` alone early-returns inside an active context; checked agentWs handlers, `verificationService`, `backupResultPersistence`, `drResultHandler`, `agents/helpers`, vault/software services). **Known residual (pre-existing, not a regression):** `publishEvent` subscribers (webhook delivery, automation worker) open system contexts after exiting the ALS, so command types that publish events still double-hold briefly — now bounded by the short handler wrap instead of the whole message. Tracked as a #1105 follow-up.

## Deliberate semantics changes (called out from review)

- **DR persistence and the per-type handler are no longer one transaction** (`vm_restore_from_backup` / `vm_instant_boot` / `bmr_recover` are in both `DR_COMMAND_TYPES` and `commandResultHandlers`). Previously a throwing handler rolled back `handleDrCommandResult`'s writes; now DR state commits independently — consistent with the DR reconcile enqueue that already fired regardless.
- **Post-CAS failure is terminal-with-ack:** the terminal CAS commits under its own context, so a later failure (e.g. a handler wrap failing to open under pool exhaustion) reaches the function-level catch with the row already terminal and the ack sent — the downstream org-table transition is left to the stale-timeout sweeps. Logged + Sentry-captured, so loud rather than silent. Documented on the function docstring.

## Sibling check (from the issue)

`agentWs.backupProgress` keeps its message wrap: `applyBackupProgress` reads/writes RLS tables through the ambient db and opens no nested system context, so it holds exactly one connection — not the same defect.

## Tests

- New `#3021` suite (3 tests) pins wrapper composition: no `org` frame around the `device_commands` select/CAS or the `devices` recheck; handler dispatch under exactly the short labelled org wrap with the agent's orgId; the deviceId-less fallback lookup under exactly `outside+system` (the silent-0-row branch) with the result honored rather than misrouted; orphaned results under `.orphaned`; the removed message-level label asserted absent. Review verified the suite fails against pre-fix code (not vacuous).
- `pnpm exec vitest run src/routes/agentWs.test.ts`: 90/90 pass (87 existing incl. #1375 / BREEZE-X / BREEZE-H, + 3 new).
- `tsc --noEmit` clean on apps/api.

Closes #3021

### DIFF
--- a/apps/api/src/routes/agentWs.test.ts
+++ b/apps/api/src/routes/agentWs.test.ts
@@ -262,8 +262,8 @@ vi.mock('../services/sentry', async (importOriginal) => {
   return { ...actual, captureMessage: vi.fn() };
 });
 
-import { db, runOutsideDbContext, withSystemDbAccessContext } from '../db';
-import { devices, deviceCommands } from '../db/schema';
+import { db, runOutsideDbContext, withSystemDbAccessContext, withDbAccessContext } from '../db';
+import { devices, deviceCommands, scriptExecutions } from '../db/schema';
 import { captureMessage } from '../services/sentry';
 import {
   createAgentWsHandlers,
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer)
**Findings:** 0 Critical. 1 Important (overstated docstring invariant — the bare-nested vs `outside+system` distinction, plus the residual indirect double-hold via `publishEvent` subscribers) → fixed in b82114429 and called out in the PR body. Comment-rot items (2 stale test-file comments still claiming a message-long held org transaction, 2 minor docstring wordings) → fixed in b82114429. Test gap (deviceId-less fallback lookup — the silent-0-row branch) → new test added in b82114429. Remaining suggestions (DR-branch / rejected-handler wrap tests) assessed by the test analyzer as not worth the setup cost; DR transaction-split and post-CAS terminal-with-ack trade-off documented in the PR body instead. 0 outstanding.
**Verification note:** code-reviewer independently confirmed the new tests fail against pre-fix code (not vacuous) and enumerated every DB touch on the path — nothing RLS-guarded runs context-free.
**Tests:** `pnpm exec vitest run src/routes/agentWs.test.ts` — 90/90 pass (87 existing incl. the #1375 / BREEZE-X / BREEZE-H context-shape suites, + 3 new #3021 tests). `tsc --noEmit` clean on apps/api (re-run after the review fixes).

Issue #3022, opened 2026-08-02
[API] Event-loop starvation surfaces as Postgres CONNECT_TIMEOUT and 10s 500s on agent routes — the DB and network are both healthy


ToddHebebrand:
## Description

On 2026-08-02 production started emitting three distinct error signatures within ~13 minutes of each other. They are all one failure, and none of them is a bug in the code Sentry names:

- `Error: write CONNECT_TIMEOUT <db-host>:<port>` — postgres.js connection establishment
- `TypeError: Cannot read properties of null (reading 'write')` at `Immediate.nextWrite` (postgres.js `connection.js:255`)
- `[PatchScheduler] Orphan-reconcile sweep failed` from `enqueueScanResults`

Agent requests fail at `duration_ms` just over 10000 — exactly the `connect_timeout`, which is the tell.

**Root cause:** under load the API process saturates its main thread and stalls the Node event loop past 10s, which kills in-flight Postgres connection handshakes. The database and the network are both healthy throughout.

`connect_timeout` is set to 10 in `apps/api/src/db/index.ts:34`. Reading the vendored postgres.js 3.4.9, `connectTimer.start()` fires in `connect()` (`connection.js:343`) immediately after the socket is created, and is cancelled only on ReadyForQuery (`connection.js:552`). So that 10s budget covers socket create → TCP → SSL → auth → ReadyForQuery, and **every one of those steps depends on the event loop being able to run the socket callbacks.** A pegged main thread blows the budget without the database or the network being involved at all.

The other two signatures fall out of that one:

- `connectTimedOut()` calls `socket.destroy()` (`connection.js:260-263`), but `write()` may have already queued `nextWriteTimer = setImmediate(nextWrite)` (`connection.js:245-258`). That deferred `nextWrite` then runs `socket.write(chunk, fn)` against the now-nulled socket → the `TypeError`. It is caught by the `isRecoverablePostgresConnectionTeardown()` guard from #1880 (logs `[db] Suppressed postgres connection-teardown write race`), so the API does **not** crash-loop — but it is pure noise downstream of the timeout.
- The patch scheduler's failed DB calls hit the deliberate `captureException` in `enqueueScanResults` (`apps/api/src/jobs/patchSchedulerWorker.ts:613`), so an unrelated worker becomes the loudest thing in Sentry.

## Ruled out, by measurement rather than assumption

[...]

PR #3024, opened 2026-08-02
fix(api): stop event-loop starvation masquerading as a Postgres CONNECT_TIMEOUT (#3022)

ToddHebebrand:
## What this fixes

`postgres.js` arms `connect_timeout` with a plain `setTimeout` (`connection.js:1042-1054`), starts it in `connect()` (`:343`) and cancels it only on ReadyForQuery (`:552`). The 10s budget therefore covers socket create → TCP → SSL → auth → ReadyForQuery, and **every one of those steps depends on the event loop being able to run a socket callback**. A pegged main thread expires that timer with the database and the network both healthy, and the driver reports `write CONNECT_TIMEOUT <host>:<port>` — a connection fault that never happened.

Per #3022 that misattribution is most of why the incident took so long to pin down: three unrelated Sentry signatures (`CONNECT_TIMEOUT`, the `nextWrite` `TypeError` from the driver teardown race, and a `[PatchScheduler]` sweep failure), all downstream of one stall, and nothing in telemetry naming the loop.

This PR makes the loop's own health a first-class, queryable signal and stops the driver's error from pointing investigation at the wrong subsystem.

## What it does

| Area | Change |
|---|---|
| `services/eventLoopMonitor.ts` | `perf_hooks.monitorEventLoopDelay` sampled into a bounded window. Reports the worst **sampled** lag *and* any stall **still in flight** |
| `services/postgresConnectTimeout.ts` | Classifies a CONNECT_TIMEOUT as `event-loop-starvation` / `connectivity` / `unknown` |
| Sentry | New `connect_timeout_cause` + `event_loop_lag_bucket` tags, set in `captureException` |
| Prometheus | `breeze_nodejs_eventloop_lag_{max,mean}_seconds`, `_starved`, `_monitored` |
| `/health/ready` | Reports lag (does **not** gate readiness on it) |
| Startup | Throttled `[event-loop] Main thread blocked for up to Nms` warning |

### Design decisions worth reviewing

- **In-flight lag is tracked separately from sampled lag.** When the loop is blocked, the sampler is a timer too, so it cannot run either. If the driver's connect timer fires before the monitor's sampler gets its turn, the stall is not yet in the ring buffer — only the in-flight figure reveals it. Both are folded into `worstLagMs`.
- **No monitor ⇒ `unknown`, never `connectivity`.** Falling back to "the handshake failed" without evidence would silently re-create the exact misdiagnosis this PR removes, and do it with the authority of a verdict. `monitored: false` is propagated everywhere, including a dedicated `_monitored` metric so a blind instance cannot read as a healthy one.
- **Tagged in `captureException`, not in `app.onError`.** In the original incident the loudest signature was the patch scheduler, not any route. `captureException` is the one chokepoint every path (routes, workers, process handlers) already goes through.
- **Readiness is deliberately not gated on lag.** Starvation is a load symptom; failing readiness under load would pull the instance from rotation, push its traffic onto peers, and starve those in turn.
- **`connect_timeout` stays at 10s.** The issue is explicit that raising it alone only lengthens the failing request. A drift guard pins the pool literal to the classifier's constant so the inspected window always equals the budget that expired.
- **Sentry capture is throttled, console is not.** One stall produces a run of breaching samples. This repo has twice had an unthrottled recurring warning exhaust the org's event quota and silently drop all error reporting (#1894, BREEZE-H).

### One structural change worth calling out

`services/sentry.ts` receives the classifier through `setConnectTimeoutClassifier(...)` instead of importing it. A static import pulled the classifier and the event-loop monitor into the module graph of **every module that reports an error** — `db/index.ts` among them — which measurably slowed module evaluation and pushed `db/requestDatabasePool.test.ts` (a hard 15s budget on a dynamic `import('./index')`) past its timeout. That regression was caught locally and is fixed by the inversion, which matches how the metrics recorders are already wired (`setBackupMetricsRecorder`, `setS1MetricsRecorder`, …). Measured on the same 7-file set: **20.6s and failing → 6.9s and green**.

## Scope: this is the second half of the issue, not the first

#3022 proposes two explicitly "separable pieces". This PR implements the second (stop the misattribution + add instrumentation) in full.

It does **not** implement the first (reduce main-thread work on the auth-rejection path), because every available option changes auth or security semantics and needs a design decision rather than a guess:

- A **negative auth cache** would suppress re-verification of a just-promoted pending token for the cache TTL — a real auth-timing behaviour change.
- A **pre-auth rate limit** on `/api/v1/agents/` reverses a deliberate skip in `globalRateLimit.ts:25`, and the existing per-org ceiling is already a known fleet-size constraint.
- **Skipping the fallback audit** on rejected requests is not available: `index.ts:794` deliberately records 401/403 as `result: 'denied'`, which is a security signal.

For the record, the issue's load claim checks out exactly: every agent 401/403 beyond a missing/malformed bearer costs a `devices` SELECT inside a `withSystemDbAccessContext` transaction (BEGIN + 6 `set_config` + SELECT + COMMIT), with no positive or negative cache; mutating rejected requests on non-excluded paths pay a second device SELECT plus an audit insert. The token hashing itself is `createHash('sha256')` and is not the CPU culprit.

**`Refs #3022` rather than `Closes` on purpose** — the issue should stay open for the first half. Happy to change that if you'd rather split it into a follow-up issue.

## Verification

- `vitest run` on the affected files, single worker: **83 passed** (`eventLoopMonitor`, `eventLoopStarvationReporter`, `postgresConnectTimeout`, `eventLoopMonitorWiring`, `sentry`, `requestDatabasePool`).
- Metrics/gateway suites: **63 passed**. DB context suites (`contextlessWriteGuard`, `heldContextCaptureThrottle`, `dbWriteExpectingRows`, `requestDatabasePool`): green.
- The load-bearing assumption — that `monitorEventLoopDelay` keeps accumulating while JS is blocked — is proven against a **real 250ms busy-wait**, not only against the fake histogram used elsewhere in the suite.
- No tenancy, RLS, migration, or cascade surface touched, so the contract suites are not implicated.

## Risk

Low. Nothing changes request control flow: no retry, no status remap, no timeout change. The monitor is a native histogram plus one `unref()`'d interval. Worst case if the monitor never starts, every consumer reports `unknown`/`monitored: false` and behaviour is exactly as today.

Two new env knobs, both optional: `EVENT_LOOP_STARVATION_WARN_MS` (default 1000), `EVENT_LOOP_STARVATION_THROTTLE_MS` (default 300000), plus `EVENT_LOOP_MONITOR_DISABLED` to opt out.

Refs #3022

### DIFF
--- a/apps/api/src/db/index.ts
+++ b/apps/api/src/db/index.ts
@@ -32,6 +32,17 @@ const client = postgres(requestDatabaseConfig.url, {
   max: getDbPoolMax(),
   idle_timeout: 20,
   max_lifetime: 60 * 30,
+  // #3022. This timer is a plain setTimeout inside the driver, so it expires
+  // when the main thread is too busy to run the socket callbacks just as
+  // readily as when the handshake actually fails — see
+  // services/postgresConnectTimeout.ts, which classifies the two apart.
+  //
+  // The literal is duplicated as POSTGRES_CONNECT_TIMEOUT_SECONDS there rather
+  // than imported from there: importing would pull the classifier and the
+  // event-loop monitor into this module's graph, and this module is the one
+  // db/requestDatabasePool.test.ts re-imports under a hard 15s budget. The two
+  // values are pinned together by a contract test in that classifier's suite,
+  // so they cannot drift silently.
   connect_timeout: 10,
 });
 
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -160,7 +160,21 @@ import { extensionsAdminRoutes } from './routes/extensionsAdmin';
 import { extensionsWebRoutes } from './routes/extensionsWeb';
 import { internalSyntheticRoutes } from './routes/internal/synthetic';
 import { bootstrapPlatformAdmins } from './services/platformAdminBootstrap';
-import { captureException, flushSentry, initSentry } from './services/sentry';
+import {
+  captureException,
+  captureMessage,
+  flushSentry,
+  initSentry,
+  setConnectTimeoutClassifier,
+} from './services/sentry';
+import {
+  getEventLoopLagStats,
+  getEventLoopStarvationThresholdMs,
+  startEventLoopMonitor,
+  stopEventLoopMonitor,
+} from './services/eventLoopMonitor';
+import { createStarvationReporter } from './services/eventLoopStarvationReporter';
+import { diagnoseConnectTimeout } from './services/postgresConnectTimeout';
 import { isBenignRejection, isRecoverablePostgresConnectionTeardown } from './services/rejectionSuppressions';
 import { partnerGuard } from './middleware/partnerGuard';
 import { API_VERSION } from './version';
@@ -463,10 +477,18 @@ app.get('/health/ready', async (c) => {
 
   const allOk = Object.values(checks).every((v) => v === 'ok');
 
+  // #3022 — event-loop lag is REPORTED here but deliberately does NOT feed
+  // `allOk`. Starvation is a load symptom, and failing readiness under load
+  // would pull the instance out of rotation, push its traffic onto the
+  // remaining instances, and starve those in turn. The whole point of surfacing
+  // it is to make the condition diagnosable, not to act on it automatically.
+  const eventLoop = getEventLoopLagStats();
+
   return c.json(
     {
       status: allOk ? 'ready' : 'not_ready',
-      checks
+      checks,
+      eventLoop
     },
     allOk ? 200 : 503
   );
@@ -1031,6 +1053,18 @@ app.onError((err, c) => {
     );
   }
 
+  // #3022 — say out loud what a CONNECT_TIMEOUT actually means before the raw
+  // error is logged. The driver's own message is `write CONNECT_TIMEOUT
+  // <host>:<port>`, which reads as a database or network fault and sent the
+  // original investigation after both; the loop being blocked produces a
+  // byte-identical error. Console-side only (the Sentry tags are set in
+  // captureException) so the explanation is present even when Sentry is
+  // disabled, which is the self-hosted default.
+  const connectTimeout = diagnoseConnectTimeout(err);
+  if (connectTimeout) {
+    console.error(connectTimeout.message);
+  }
+
   // Route unhandled errors to Sentry. Per-route `captureException(err, c)`
   // calls only cover routes with explicit try/catch — anything that throws
   // and falls through to onError was previously invisible to Sentry.
@@ -1401,6 +1435,10 @@ async function shutdownRuntime(signal: NodeJS.Signals): Promise<void> {
   console.log(`[shutdown] Received ${signal}, shutting down gracefully...`);
 
   getWebhookWorker().stop();
+  // The sampler is already unref'd, so this is tidiness rather than a
+  // requirement — it just stops starvation warnings from being emitted about a
+  // process that is deliberately winding down and no longer serving traffic.
+  stopEventLoopMonitor();
   if (auditRetryInterval) {
     clearInterval(auditRetryInterval);
     auditRetryInterval = null;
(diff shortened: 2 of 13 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1: code-reviewer, silent-failure-hunter, pr-test-analyzer (all three in parallel on `ed64fbd`); round 2: code-reviewer on the fix delta (`058e88f`).

**Findings:** 9 raised across both rounds → all 9 addressed in `058e88f` + `926c06b`; 0 outstanding.

Round 1 found three ways the classifier could still emit a confident, wrong verdict — the exact misdirection this PR exists to remove:

- **`monitored: true` did not mean "can vouch for the window."** A monitor up less than the 10s connect budget reported lag 0 meaning *not observed*, and the classifier read it as *the loop was fine* → `connectivity`. Reachable on **every** diagnosis in the first 10s after boot. Readings now carry `coversWindow`, required alongside `monitored`.
- **The starvation threshold was uncapped.** `EVENT_LOOP_STARVATION_WARN_MS` is a warning-*volume* knob; at `15000` a 12s stall that consumed the whole budget reported "the event loop stayed healthy — check database reachability." Now capped at the connect window.
- **`app.onError` called the classifier unguarded, ahead of `captureException`** — a throw would have cost the request its JSON 500 *and* stopped the original error reaching Sentry. Added `safeDiagnoseConnectTimeout`, used at both error-path call sites.

Round 2 caught that the first fix **under-shot the same class of bug**: `coversWindow` rejected only `interval > window`, but the blind spot is the current un-recorded interval and is one *interval* wide. Reviewer reproduced it — at a 9s interval an 8.5s stall sits entirely inside one interval and still returned `connectivity`. The guard is now `min(window, starvationThreshold)`; the shipped defaults satisfied that only by coincidence (both 1000ms), which is why it's enforced rather than assumed.

Also fixed: Prometheus `_lag_max_seconds` / `_starved` were 120s high-water marks (a 1.2s blip held `starved=1` across ~8 scrapes, so `starved == 1 for 1m` fired on a momentary spike) — now instantaneous, with the high-water mark under an explicitly named `_lag_window_max_seconds`; `_lag_mean_seconds` renamed to `_lag_window_mean_seconds` because it was left on a different time base and a recovered loop could publish a max *below* its mean; boot now logs whether the monitor runs, at what interval, and both thresholds, and warns when the interval exceeds the threshold.

**Test fixes — three gaps were mutation-proven by the reviewer and are now closed:**

- The four Prometheus gauges had **no coverage**: deleting `updateEventLoopMetrics()` left the whole suite green while every dashboard read a healthy loop. Verified the new tests fail under exactly that mutation.
- The `start()`-idempotence test was **vacuous** — the harness returned one shared fake histogram, so deleting the guard still passed. The factory now mints a fresh instance per call and counts constructions.
- The Sentry tagging test asserted `expect.any(String)`, which would pass just as happily on a regression reporting `connectivity` with no evidence. Asserts the literal `'unknown'` now.

I also mutation-verified the round-2 guard: reverting it fails the new 9s-interval case.

**Tests:** affected suites green single-worker — **163 passed / 11 files** (`eventLoopMonitor`, `postgresConnectTimeout`, `eventLoopStarvationReporter`, `eventLoopMonitorWiring`, `sentry` ×2, `metrics`, `requestDatabasePool`, `contextlessWriteGuard`, `heldContextCaptureThrottle`, `dbWriteExpectingRows`). `tsc --noEmit` clean, `eslint` clean. **Full CI: 52/52 passing**, including Test API and all four Integration Tests shards.

One CI note for the record: the `Test Web` red on the middle push was **flaky and unrelated** — this branch changes zero files under `apps/web`, the same job passed on the first commit with identical web code, and the failing spec (`billing/InvoiceWorkspace`) was untouched by both new commits on `main`. It passes on the current head.

**Two things for you to decide:**

1. **`Refs #3022`, not `Closes`** — this is the second of the issue's two explicitly "separable pieces". The first (reduce main-thread work on the auth-rejection path) is deliberately not attempted: a negative auth cache changes auth timing for just-promoted pending tokens, a pre-auth rate limit reverses a deliberate skip in `globalRateLimit.ts:25`, and skipping the fallback audit isn't available because `index.ts` records 401/403 as `result: 'denied'` on purpose. Each needs a design call rather than a guess. The issue should stay open; happy to split the remainder into its own issue if you'd prefer.
2. **`/health/ready` is unauthenticated and now publishes loop-pressure telemetry.** Two reviewers flagged this as low severity — it's consistent with the endpoint already exposing DB/Redis availability, and the detailed series are on the auth-gated `/metrics` — but it does let an unauthenticated prober measure whether its own load is starving the instance. Easy to drop the field if you'd rather not expose it.

**Status:** review-clean, CI green, awaiting maintainer merge. Not merging and not closing the issue — those stay your call.

Issue #2999, opened 2026-08-01
[Agent][Backup] VSS InitializeForBackup fails with E_INVALIDARG on every run (test VM) and the backup silently proceeds without a snapshot


ToddHebebrand:
## Description

On a Windows Server 2022 test VM, **every** backup run failed to create a VSS shadow copy and silently proceeded without it. `InitializeForBackup` returns `HRESULT 0x80070057` (`E_INVALIDARG`) in about 2 ms — the call is being rejected outright, not timing out.

Flagging with lower confidence than the sibling issues: this may be environmental (Server 2022 Standard **Evaluation**, Hyper-V guest). It needs confirming on a second machine before treating it as a code defect. Filing because the failure was 100% reproducible across every run in the session and because the *fallback behaviour* is worth reviewing regardless of the root cause.

Observed on a local stack running `v0.103.0` (commit `ad406831d`), 2026-08-01. Only visible at all because #2801 gave the helper real logs.

## Evidence

Identical on all three runs in the session:

```
level=INFO msg="vss: backup components created"
level=WARN msg="VSS shadow copy failed, proceeding without VSS" component=backup elapsedMs=2
      error="vss: InitializeForBackup failed: vss: vtable[3] HRESULT 0x80070057"
```

Note `vss: backup components created` succeeds first, so `CreateVssBackupComponents` is fine; the failure is the next call. `elapsedMs=2` (and `elapsedMs=0` on other runs) rules out the 10-minute VSS timeout path at `agent/internal/backup/backup.go:317`.

## Why it matters even if environmental

`agent/internal/backup/backup.go:321-325` treats VSS failure as non-fatal and continues without a snapshot. That is the correct default — a backup without VSS beats no backup — but it means:

- Locked files (PST/OST, browser profiles, `NTUSER.DAT`) get skipped, and the run still reports success.
- The only signal is a single `WARN` line in `backup.log`. There is no field on `backup_jobs` recording whether the snapshot was VSS-backed, so an admin cannot tell a VSS-protected restore point from a best-effort one after the fact.

For a `C:\Users` backup — the common case — "ran without VSS" is a materially different restore point, and it is currently invisible above the log line.

## Asks

1. Confirm or rule out on non-evaluation Windows (Server 2022 Standard licensed, and a Win 10/11 client) before treating the HRESULT as a code bug. If it reproduces broadly, `InitializeForBackup`'s arguments at `agent/internal/backup/vss/vss_windows.go` are the place to look.
[...]

Issue #2523, opened 2026-07-15
Six pre-existing Windows Go test failures keep the CI Windows job narrowed


ToddHebebrand:
## Summary

Six Go tests fail on Windows and have been failing for some time. They are **not** caused by any recent branch — they reproduce identically on `main`, on GitHub `windows-latest`, and on a local Windows Server 2022 VM. They compile fine and fail at runtime.

Because of these, the `Test Agent (Windows)` CI job currently runs a narrowed package list (`sessionbroker`, `eventlog`, `watchdog`, `cmd/breeze-watchdog`) rather than `./...`. The job is green and remains a required check, but the packages below are unguarded on Windows.

## The failures

| Test | Package |
|---|---|
| `TestSendHelperTokenUpdateOnlyReachesAssistSessions` | `internal/heartbeat` |
| `TestHandleHelperSessionAuthenticatedPushesOnlyToAssist` | `internal/heartbeat` |
| `TestResolveRunAsSessionUserPrefersRunAsUserScope` | `internal/heartbeat` |
| `TestReconcileUserHelper_UnexpectedStatError_NoDownload` | `internal/heartbeat` |
| `TestStaticUnitMatchesEmbedded` | `internal/agentapp` |
| `TestPersistedServerURLProviderFollowsPromotion` | `internal/config` |

### 1-3. Console-session gate (`internal/heartbeat`) — likely, not yet verified on Windows

The three token/run-as tests build sessions via `sessionbroker.NewSession(...)`, which never sets `Session.WinSessionID` (stays `""`), and construct the broker with no GOOS override. Delivery is gated on `SessionInConsoleSession`, which on real Windows requires `s.WinSessionID == b.ConsoleSessionID()`. The real WTS lookup returns a non-empty session id, which never equals `""`, so the token is withheld and the assertion fails. These pass on darwin/linux only because `SessionInConsoleSession` short-circuits to `true` when the effective GOOS is not windows.

Suggested fix: have the tests call `SetConsoleSessionIDFunc(func() string { return "" })`, or set matching `WinSessionID`s on the fixtures.

### 4. `TestStaticUnitMatchesEmbedded` — CRLF (verified)

[...]

PR #3029, opened 2026-08-02
fix(agent/backup): stop VSS from rewriting the system-state staging dir onto a path the snapshot never had (#3026)

### LINKED ISSUE #3026: System-state artifacts silently dropped when VSS shadows the temp volume
Found while reviewing #3025. Pre-existing; not introduced there.

## The bug

In `RunBackupContext` (`agent/internal/backup/backup.go`) the ordering is:

1. `CreateShadowCopy(...)` — the snapshot is taken (~line 324)
2. `collectSystemState()` — creates the staging dir via `os.MkdirTemp("", "breeze-systemstate-*")` (~line 367, `systemstate/systemstate.go:14`)
3. the staging dir is appended to `backupPaths` (~line 384)
4. `rewritePathsForVSS` rewrites everything onto shadow-device paths

A VSS shadow copy is a point-in-time image. The staging dir is created **after** the snapshot, so it does not exist inside it. But `%TEMP%` is normally on `C:`, which is normally one of the shadowed backup volumes — so step 4 rewrites the staging dir to `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyN\Windows\Temp\breeze-systemstate-XXXX`, a path that isn't in the snapshot.

The walk then finds nothing there. The stat failure is folded into the aggregate `"backup file scan completed with errors"` warning, `markSystemStateFiles` matches zero files, and `job.SystemStateManifest` is still set — so the job reports success with the system-state artifacts silently absent from the backup.

The existing comment at the `systemStateStagingIdx` block acknowledges the rewrite happens ("`rewritePathsForVSS` above may have rewritten it too") but treats it as a path-recovery bookkeeping problem rather than a correctness one.

## Impact

Silent, partial data loss on `system_image` backups whenever VSS covers the temp volume — which is the default configuration, not an edge case. The restore point looks complete.

## Suggested fix

Exclude the staging dir from the **rewrite** itself, not just from reporting — pass `systemStateStagingIdx` through `rewritePathsForVSS` unrewritten. That also makes the path-recovery dance at the `systemStateStagingIdx` block unnecessary.

## Notes

- No test anywhere covers the system-state + VSS interaction.
- #3025 excludes the staging dir from its live-read warning, which is correct for the case where the staging dir is on an *unshadowed* volume, but does not address the mapped case above.

ToddHebebrand:
## The bug

`RunBackupContext` creates the VSS shadow copy **before** `collectSystemState`
creates its `os.MkdirTemp` staging dir, then appended that dir to `backupPaths`
and let `rewritePathsForVSS` map it onto a shadow-device path.

A shadow copy is a point-in-time image, so a directory created after it was
taken cannot be inside it. `%TEMP%` normally sits on `C:`, so when `C:` is a
shadowed backup volume the rewrite pointed the walk at
`\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyN\Windows\Temp\breeze-systemstate-XXXX`
— a path absent from the snapshot.

Every downstream signal then swallowed it: the walk found nothing and the stat
failure folded into the aggregate `"backup file scan completed with errors"`
warning; `markSystemStateFiles` matched zero files; `job.SystemStateManifest`
was still set. The job reported **success** with the system-state artifacts
absent from the restore point.

**Correction to the issue's Impact section**, verified against
`defaultVSS` in `agent/cmd/breeze-backup/exec_backup.go`
(`goos == "windows" && !systemImage`): this is **not** the default
configuration. Server-dispatched `system_image` runs default VSS *off*. The bug
needs VSS and system-state collection enabled for the same run — an `agent.yaml`
with both `backup_vss_enabled` and `backup_system_state_enabled`, or a
`system_image` run with an explicit `vss` override. Opt-in, not universal.
Likewise the green-job outcome requires configured file paths: a
system-state-*only* run trips the pre-existing `len(files)==0` hard-failure
guard and fails loudly instead. The comments in the fix say this; the issue body
does not.

## The fix

**1. Exclude the staging index from the rewrite itself**, not merely from
reporting — the suggestion in the issue, and it holds up.
`rewritePathsForVSS` now takes `stagingIdx` (`noStagingIdx` when the run has
none) and leaves that path on the live volume, the only place those artifacts
exist. It still reports the index as unmapped, because "read from the live
volume" is factually what happens — which keeps `reportableLiveReads` (#3025)
meaningful rather than dead: that is the layer that knows the agent wrote the
dir itself and drops it from the operator warning. The fix does not trade
silent data loss for a warning on every `system_image` run.

Two follow-on simplifications, verified rather than assumed:

- The post-rewrite path-recovery block is **deleted**. With the staging path
  never rewritten it could only reproduce what `collectSystemState` returned,
  so `systemStateStagingDir` is assigned there directly.
- `markSystemStateFiles`' doc told callers to "pass whichever value was ACTUALLY
  walked" because of the possible rewrite. That is now wrong; replaced with the
  invariant that makes the prefix comparison valid.

**2. Make the outcome loud, not just this route.** The manifest is written from
the collector's return value and says nothing about whether artifacts reached
the snapshot. `systemStateArtifactsMissing` reports the divergence — a manifest
describing artifacts where zero collected files matched the staging dir — and
the run warns instead of completing clean.

**3. Stop shipping a manifest for artifacts that never landed.** This is the
part a warning alone does not cover, and it came out of review. Unlike
`job.VSSMetadata` (genuinely stripped at parse), `SystemStateManifest` survives:
`resultSchemas.ts:78` accepts it, `backupResultPersistence.ts:669` writes it to
`backup_snapshots.system_state_manifest`, and `bmr.ts` / `recoveryBootstrap.ts`
serve it to the bare-metal recovery paths. Left intact it advertises a complete
system state on a restore point holding none of it. On detection the artifact
list is cleared; platform, OS version and hardware profile are retained (inline
collector output, still accurate — and `hardwareProfile` is persisted *from* the
manifest, so dropping it wholesale would discard good data).

**4. Close the sibling silent route.** `collectSystemState` failing outright
only fails the run via the `len(files)==0` branch, so a run that also has
configured file paths completed **green with no system state and no signal**. It
now appends a warning carrying the collector's reason. `CollectSystemState`
errors only when a *required* class failed, i.e. the capture would not boot at
restore time.

**5.** The `IncompleteSteps` branch wrote `job.Warning` by raw assignment — safe
while it was the only writer, now one of four. Moved to `appendWarning` before
it silently discards one of the others.

## Tests

**Windows-gated** (`filepath.VolumeName` returns `""` off Windows, so a portable
version of these would pass vacuously):

- `TestRewritePathsForVSS_LeavesSystemStateStagingDirUnrewritten` — the
  regression, table-driven: staging on a **shadowed** volume stays live while
  user paths are rewritten (the bug); on an unshadowed volume stays live;
  `noStagingIdx` leaves everything eligible; index **0** is a real index and
  never a second sentinel (the system-state-only run puts staging there); an
  out-of-range index excludes nothing.
- `TestRewritePathsForVSS_StagingDirStaysMatchableByMarkSystemStateFiles` —
  the downstream symptom, not the mechanism: composes the production rewrite
  with the production `markSystemStateFiles`. A shadow-rewritten root matches
  zero artifacts, which is how the manifest got recorded describing artifacts
  the snapshot did not contain.
- `TestRewritePathsForVSS_StagingDirIsNotWarnedAsALiveRead` — the exclusion and
  `reportableLiveReads` compose, so no per-run operator noise.

**Portable**, driving `RunBackupContext` through the `collectSystemState` seam:

- `TestRunBackup_SystemStateManifestWithoutArtifactsWarnsAndDropsArtifacts` —
  manifest claiming artifacts + empty staging dir on a run with file paths:
  completes, warns, clears the artifact list, retains platform.
- `TestRunBackup_SystemStateCapturedProducesNoWarning` — a healthy capture warns
  not at all and keeps its artifacts.
- `TestRunBackup_SystemStateCollectionFailureWarnsOnMixedRun` — a failed
  collection alongside file paths surfaces the collector's reason.
- `TestSystemStateArtifactsMissing` — the detector's threshold, including the
  deliberate exclusions (nil manifest, zero-artifact manifest).
- `markSystemStateFiles`' new count return is asserted in both existing tests.

The five pre-existing `rewritePathsForVSS` call sites take the new
`noStagingIdx` argument; no existing assertions changed.

**CI:** all three Windows-gated names start with `TestRewritePathsForVSS`, so the
existing `-run` regex already selects them; the exact top-level PASS count is
bumped `11` → `14`. The portable tests match no alternative in that filter, so
the count is unaffected by them.

## Verification

- **Windows CI executed the new tests** — `Test Agent (Windows)` green, and the
  `-run` step's exact-count guard passed at 14. All three new tests and every
  subtest show `--- PASS` in the runner log, including
  `staging_dir_on_a_shadowed_volume_stays_live_while_user_paths_are_rewritten`.
- `go test -race ./internal/backup/...` green on darwin (all 7 packages).
- `GOOS=windows go vet` + `go test -c` clean. (Three `possible misuse of
  unsafe.Pointer` notices are pre-existing in `internal/backup/vss/vss_windows.go`,
  untouched here.) `gofmt` clean.
- The 7 pre-existing Windows failures in this package (#2523 — `TestContract*`,
  `TestCollectBackupFiles_CapturesMode`, `TestCreateSnapshot_*`,
  `TestRestoreFromSnapshot_*`) are **unchanged**: none exercises
  `rewritePathsForVSS`, `markSystemStateFiles` or the system-state path, and
  none is selected by the Windows `-run` filter, so this PR adds no new Windows
  reds.

## Deliberately not in this PR

Raised in review, declined with reasons — worth separate issues:

- **Partial system-state capture is still silent.** The detector's threshold is
  "none at all". A per-artifact presence check (intersecting
  `Artifact.Path` against the marked set) would catch an exclude pattern eating
  *some* artifacts, but the path-format assumptions make a false positive — a
  warning on every healthy run — the likely failure mode. Wants its own change.
- **Total VSS failure is quieter than partial VSS failure.** `CreateShadowCopy`
  failing logs only and skips the whole live-read warning block, so a run where
  *every* volume read live says less than one where a single volume did. That is
  the #2999/#3025 lineage, not this one.
- **`lastResortStdout`** (`cmd/breeze-backup/result_bounds.go`) *replaces* rather
  than appends to `warning` at tier 4, destroying these signals on the largest
  runs. Pre-existing (#3001).

Closes #3026

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1194,7 +1194,7 @@ jobs:
           $out | ForEach-Object { Write-Host $_ }
           if ($exit -ne 0) { throw "go test failed (exit $exit)" }
           $passes = @($out | Select-String -Pattern '^--- PASS: Test').Count
-          if ($passes -ne 11) { throw "expected exactly 11 top-level PASS lines, got $passes - a test in the -run filter was renamed, removed, or added without updating this count" }
+          if ($passes -ne 14) { throw "expected exactly 14 top-level PASS lines, got $passes - a test in the -run filter was renamed, removed, or added without updating this count" }
 
       # -race requires cgo, which remote/desktop cannot satisfy on Windows, so
       # this covers only packages that do not import it (heartbeat and agentapp
--- a/agent/internal/backup/backup.go
+++ b/agent/internal/backup/backup.go
@@ -354,12 +354,14 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 
 	// System state collection: gather OS config, hardware profile, etc.
 	var systemStateErr error
-	// systemStateStagingDir/systemStateStagingIdx let us recover, after any
-	// VSS rewrite below, whichever staging-root value was ACTUALLY walked —
-	// see the assignment right after the VSS rewrite for why the raw
-	// stagingDir returned here isn't necessarily it.
+	// systemStateStagingIdx marks the staging dir's slot in backupPaths so the
+	// VSS rewrite below can leave it alone — it is created after the snapshot
+	// and therefore only exists on the live volume (#3026). Because it is never
+	// rewritten, systemStateStagingDir is simply the path collectSystemState
+	// returned, and it is the same prefix collectBackupFilesFromPaths produces
+	// for markSystemStateFiles to match on.
 	var systemStateStagingDir string
-	systemStateStagingIdx := -1
+	systemStateStagingIdx := noStagingIdx
 	if m.config.SystemStateEnabled {
 		if err := runCtx.Err(); err != nil {
 			return stopBackupRun()
@@ -381,6 +383,7 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 			}
 			// Append staging dir to backup paths so artifacts are included in snapshot
 			systemStateStagingIdx = len(backupPaths)
+			systemStateStagingDir = stagingDir
 			backupPaths = append(backupPaths, stagingDir)
 			defer func() {
 				if removeErr := os.RemoveAll(stagingDir); removeErr != nil {
@@ -393,7 +396,7 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 	// Rewrite paths to shadow copy device paths when VSS is active
 	if vssSession != nil {
 		var unmappedIdx []int
-		backupPaths, unmappedIdx = rewritePathsForVSS(backupPaths, vssSession.ShadowPaths)
+		backupPaths, unmappedIdx = rewritePathsForVSS(backupPaths, vssSession.ShadowPaths, systemStateStagingIdx)
 		if liveReads := reportableLiveReads(backupPaths, unmappedIdx, systemStateStagingIdx); len(liveReads) > 0 {
 			shadowedVolumes := make([]string, 0, len(vssSession.ShadowPaths))
 			for vol := range vssSession.ShadowPaths {
@@ -421,16 +424,6 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 			// only VSS signal that reaches the server today.
 			appendWarning(job, "read from the live volume, not the VSS shadow copy: "+summary)
 		}
-	}
-	if systemStateStagingIdx >= 0 && systemStateStagingIdx < len(backupPaths) {
-		// The staging dir's OS temp volume commonly coincides with a
-		// VSS-shadowed backup volume, so rewritePathsForVSS above may have
-		// rewritten it too — capture whichever value was actually walked
-		// (this index in the possibly-rewritten backupPaths) rather than
-		// the pre-rewrite path from collectSystemState, so
-		// markSystemStateFiles below compares against the same sourcePath
-		// prefix collectBackupFilesFromPaths actually produced.
-		systemStateStagingDir = backupPaths[systemStateStagingIdx]
 	}
 
 	if err := runCtx.Err(); err != nil {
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer.

**Findings:** 12 raised across two rounds → 8 addressed, 4 declined with reasons; 0 outstanding.

Round 1 — code-reviewer returned 0 findings at confidence >= 80 and confirmed by tracing all three branches that deleting the post-rewrite path-recovery block is safe. pr-test-analyzer independently re-derived the Windows PASS count as 14 and confirmed the `//go:build windows` gating is not vacuous. Addressed in `32fce6d98`: added the `stagingIdx == 0` case (a system-state-only run puts staging at index 0, so 0 must never become a second sentinel), guarded a table assertion on slice length, corrected the unmapped-set doc, and — the substantive one — made the *symptom* loud rather than only the fixed route, via `markSystemStateFiles` returning a marked count and `systemStateArtifactsMissing`.

Round 2 — silent-failure-hunter and comment-analyzer, addressed in `3c3f0082f`. Three of my own comments were factually wrong and are corrected:

- I wrote that `appendWarning` is the only system-state signal reaching the server. **Backwards.** `SystemStateManifest` has a JSON tag, `resultSchemas.ts:78` accepts it, `backupResultPersistence.ts:669` writes it to `backup_snapshots.system_state_manifest`, and BMR/recoveryBootstrap serve it. That is a real gap, not just a comment bug: a warning alone left the manifest advertising a complete system state on a restore point holding none of it. The detector now clears the artifact list, keeping platform/OS/hardware profile (inline collector output, still accurate — and `hardwareProfile` is persisted *from* the manifest, so dropping it wholesale would discard good data).
- "the DEFAULT configuration", copied from the issue's Impact section, is wrong. `defaultVSS` is `windows && !systemImage`, so server-dispatched `system_image` runs default VSS **off**. The bug needs VSS and system-state collection both enabled. Opt-in, not universal.
- "the job still reports success" holds only for a run that also has configured file paths; a system-state-only run trips the pre-existing `len(files)==0` hard-failure guard.

Also closed the sibling silent route (`collectSystemState` failing outright completed green with no signal on a mixed run) and moved the `IncompleteSteps` branch off a raw `job.Warning =` assignment, which was safe as the only writer and is now one of four.

**Declined**, each better as its own issue: per-artifact presence checking for *partial* capture (path-format assumptions make a warning-on-every-healthy-run the likely failure mode); total VSS failure being quieter than partial (#2999/#3025 lineage); and `lastResortStdout` replacing rather than appending `warning` at tier 4 (#3001).

**Tests:** 52 checks pass, 0 fail. `Test Agent (Windows)` green with the `-run` step's exact-count guard at 14 — all three new Windows-gated tests and every subtest show `--- PASS` in the runner log, including `staging_dir_on_a_shadowed_volume_stays_live_while_user_paths_are_rewritten`, and the job log has zero `--- FAIL` lines. Locally: `go test -race ./internal/backup/...` green (7 packages), `GOOS=windows go vet` + `go test -c` clean, `gofmt` clean.

**Pre-existing failures unchanged:** the 7 Windows reds from #2523 (`TestContract*`, `TestCollectBackupFiles_CapturesMode`, `TestCreateSnapshot_*`, `TestRestoreFromSnapshot_*`) are untouched — none exercises `rewritePathsForVSS`, `markSystemStateFiles` or the system-state path, and none is selected by the Windows `-run` filter. No new Windows reds.

**Status:** review-clean, awaiting maintainer merge. Not merging and leaving #3026 open per the handoff rules — the reporter's call.

PR #3030, opened 2026-08-02
fix(backup): deliver VSS health to the server end to end (#3027)

### LINKED ISSUE #3027: No VSS health signal reaches the server: vssMetadata stripped at parse, column never written, UI panel dead
Found while reviewing #3025. Pre-existing.

The agent collects rich VSS diagnostics per run and **none of it reaches the server**. The channel is severed at three independent points, each verified:

**1. Stripped at parse.** The agent sends `vssMetadata` in the backup command result (`BackupJob.VSSMetadata`, `json:"vssMetadata,omitempty"`). `backupCommandResultSchema` (`apps/api/src/routes/backup/resultSchemas.ts`) has no `vssMetadata` key, so zod silently drops it. `apps/api/src/jobs/queueSchemas.ts` is `.strict()` and likewise omits it.

**2. Never persisted.** The `vss_metadata jsonb` column exists (`apps/api/src/db/schema/backup.ts:244`) but nothing in `apps/api/src` ever writes or reads it — the schema definition is the only occurrence in the entire directory. It is an orphan column.

**3. UI branch unreachable.** `DeviceBackupTab.tsx:378` computes `showVssStatus = status?.lastJob?.vssMetadata != null`, but the `lastJob` projection in `apps/api/src/routes/backup/dashboard.ts` returns only `{id, status, createdAt, completedAt}`. `showVssStatus` is therefore permanently `false` and the VSS writer panel below it is dead code.

## Why it matters

VSS health is exactly the signal that distinguishes "backed up" from "backed up but every locked file was skipped" — the failure behind #2999 (308 locked-file failures in the field). Today the only evidence lives in the agent log on the endpoint.

Related: `vss.VSSSession.UnprotectedVolumes` is dropped when building `VSSMetadata` in `backup.go` (the struct copies five fields and omits it), even though its own doc comment says callers "must surface it rather than treating the session as a clean VSS backup". So the most important field is lost before it even reaches the hop where the rest is discarded.

## Suggested fix

Decide the intended channel and make one work end to end:
- add `vssMetadata` to `backupCommandResultSchema` + the queue payload, persist to the existing column, and include it in the `lastJob` projection so the existing UI panel comes alive; **or**
- drop the orphan column and the dead UI branch, and route VSS degradation through `job.Warning` only.

#3025 takes the interim path — it appends live-volume reads to `job.Warning`, the only channel that works today — but that is a workaround for this, not a fix.

Note `job.Warning` currently renders as an **error** in the UI (`BackupJobList.tsx`, destructive styling, labelled as error log), and `DeviceBackupTab.tsx` does not read `errorLog` at all, so a warning routed there is mislabelled on one screen and invisible on the other. Worth folding into whichever direction is chosen.

ToddHebebrand:
Closes #3027

## The gap

The agent collects rich VSS diagnostics per backup run and **none of it reached the server**. The issue named four cut points; reviewing the fix surfaced **four more**, all on the failure path. Every one was silent.

| # | Where | What happened |
|---|---|---|
| 1 | `routes/backup/resultSchemas.ts` | `backupCommandResultSchema` had no `vssMetadata` key → zod's default strip discarded it at the first parse |
| 2 | `jobs/queueSchemas.ts` | `backupProcessResultSchema` is `.strict()` and omitted it → it could not have survived the queue hop either |
| 3 | `services/backupResultPersistence.ts` | `applyBackupCommandResultToJob` never wrote `backup_jobs.vss_metadata` — an orphan column since migration `0074` |
| 4 | `routes/backup/dashboard.ts` | the `lastJob` projection returned 4 fields, so `showVssStatus` was permanently `false` and the whole VSS panel was dead code |
| 5 | `agent/internal/backup/backup.go` | `VSSSession.UnprotectedVolumes` was dropped when building `VSSMetadata` — the literal copied five of six fields |
| 6 | `agent/internal/backup/backup.go` | the system-state block **assigned** `job.Warning` instead of appending, destroying the VSS note whenever both degraded (they co-occur — a wedged writer subsystem causes both) |
| 7 | `agent/cmd/breeze-backup/main.go` | `marshalResult` discards the job body whenever `err != nil`, so **every hard failure** reached the server as stderr and nothing else |
| 8 | `agent/internal/heartbeat/{backup_forwarder,heartbeat}.go` | two further hops dropped the body on failure — the forwarder's error mapping, and the async handler's `else if` treating stderr and body as either/or |

Cut 5 matters most: `ExposedPaths` lists what **succeeded**, never what was **requested**, so a volume that resolved no shadow device reads identically to one that was never asked for. Cuts 6–8 meant the fix, as first written, did not actually hold on the failure branch it claimed to cover — "which writer wedged" is precisely the question you ask about a failure.

## Direction chosen — (a), make the channel work

Per the CLAUDE.md advisor-quorum rule, I formed a position and then got an independent read-only Codex `xhigh` opinion. **We agreed on (a)**, for the same reasons:

- VSS health is what distinguishes *"backed up"* from *"backed up but every locked file was skipped"* — the #2999 field failure.
- `job.Warning` is one free-text string. It cannot carry per-writer state, so under (b) "which writer wedged" stays permanently unanswerable. (b) would delete a working UI panel and a column to formalise a lossy workaround.
- No new column, migration, RLS change, or cascade/export registration was needed.

## Failure isolation — the load-bearing decision

`vssMetadata` is typed **`z.unknown()` at BOTH validation boundaries, deliberately**:

- `backupCommandResultSchema` — its parse outcome decides whether the run is recorded completed or failed (`result.status === 'completed' && parsedBackup.success`).
- `backupProcessResultSchema` — parsed inside `enqueueBackupResults` **before** `queue.add`, so a throw discards the whole terminal result.

My first draft used a typed zod object; review caught that one malformed diagnostics field would then fail a backup that actually **succeeded** and strand its snapshot — strictly worse than the F13 bug the schema comments warn about, because it flips job status rather than dropping a column. All validation, bounding and redaction moved to `sanitizeVssMetadata`, where a bad value can only cost that value.

### `sanitizeVssMetadata`

Rebuilds the blob field by field rather than storing it verbatim:

- **Caps**: 24 array entries, 1 KiB per string, 64 KiB serialized — measured with `Buffer.byteLength`, not `String.length` (a non-ASCII payload is up to 3× its UTF-16 length). The entry cap is small enough that the two string arrays *cannot together* exceed the total budget; at the original 200 they could reach ~205 KB each, forcing the pathological tier and discarding the very field the tier above preserves.
- **Unmodeled keys**: keeps top-level **scalars**, drops **containers** — the rule the agent's own IPC bounding uses (`reduceToScalars`, `result_bounds.go`) — so a future agent field stays forward-compatible without reopening the hole.
- **Redaction**: `redactSecretsDeep`, matching existing treatment of agent-supplied jsonb.
- **Nothing is dropped silently.** Count caps, type mismatches, non-list shapes, unmodeled containers and size tiers all name themselves in the persisted `warnings`. This is not tidiness: a sanitizer that quietly reduced `unprotectedVolumes: [{volume:'D:\\'}]` to `[]` would make the UI report a clean snapshot — this issue's exact failure mode, one layer up.
- **The last-resort tier still keeps `unprotectedVolumes`.** Discarding it is the one loss that turns a degraded snapshot back into a clean-looking one.
- An **unusable** blob (bare string, number, array) now logs + `captureException`. `z.unknown()` means nothing upstream rejects garbage — correct — but absent and present-but-broken must not look identical, or a fleet-wide agent regression is invisible forever.

The only ceiling on the way in is agentWs's **5 MB stdout cap**: `backup_run` is not a "critical family", so it never reaches the 512 KB structured cap.

### One subtlety worth flagging for review

The success and failure branches use **different stdout encodings**, deliberately. `toWSCommandResult` unmarshals `Stdout` into the structured `Result` field **only when `Error == ""`**. So success carries a double-encoded body (the agent peels one layer, the server's single `JSON.parse` yields the object), while failure never populates `Result` and the server falls back to `stdout` with one parse left — meaning the body must be **raw** object text. Getting this backwards turns every failed backup into a "malformed result" rather than a failure with diagnostics. Pinned by tests on both branches.

Neither change can green a failed run: `Status` is decided from `BackupCommandResult.Success` before either hop, and the server derives the job's terminal status from that field alone.

## Reconciliation with #3025

**Kept, not replaced.** Not redundant: `job.Warning` survives the IPC bounding that drops `vssMetadata` as a bulk field, and it uniquely covers the case the VSS provider **cannot see** — VSS reported total success but the path never matched a shadow root, so `UnprotectedVolumes` is empty. `vssMetadata` is the structured record; the warning is the loud summary. The stale in-tree comment documenting the severed channel was rewritten.

## Tenancy

No new column and no migration, so **nothing to register**. `vss_metadata` already exists (`0074-vss-metadata.sql`) and is already **`excludedOpen`** in `CORE_TENANT_EXPORT_POLICY` — verified still correct and deliberately not weakened (jsonb, can name volumes and paths). `backup_jobs` already carries forced org RLS and is already in all cascade lists. Independently re-verified by the code reviewer.

## Changes per layer

**Agent (Go)** — `VSSMetadata.UnprotectedVolumes` added; the metadata build extracted to `buildVSSMetadata` (it was inline in a Windows-gated function, which forced the test to *mirror* it); a server-visible warning when the VSS session fails to start at all; `appendWarning` instead of assignment in the system-state block; `marshalBackupRunResult` carrying the job body on failure; the forwarder and async handler both carrying it onward.

**API** — `vssMetadata` on both schemas as `z.unknown()`; forwarded through `ProcessResultsResult` and the agentWs literal; `sanitizeVssMetadata` + the column write on both branches; `errorLog` combined with the warning on failure (the old `error ?? warning` chain always picked `error`, so the note explaining *why* a run was degraded never reached the UI); `vssMetadata` and `errorLog` in the `lastJob` projection.

**Web** — an `unprotectedVolumes` banner ranked **above** the writer banner (a non-stable writer *may* have degraded the snapshot; a volume with no shadow copy definitely was read live); a provider-warnings list; and a diagnostic banner reading `errorLog`, styled by the job's real status — a completed-but-degraded run is a **warning**, not an error. Without it a total VSS failure rendered nothing at all, since that outcome produces no metadata and "no VSS panel" also describes a Linux box. i18n keys added to all 7 locales.

## Tests

Highlight: **`agentWs.enqueueContract.test.ts` gains a completeness contract** — every key `backupProcessResultSchema` declares must appear in agentWs's hand-maintained forwarding literal. That literal was the one hop with **zero** coverage: deleting the `vssMetadata` line left 202/202 green while the feature did nothing in production. Verified by mutation (the test fails with the line removed). It retires the bug class, not just this instance — the same hole existed for `systemStateManifest` and `referencedBytes`.

Also: Go reflection test now run against **production** `buildVSSMetadata` rather than a mirror (removing the field from the real literal previously left the Go suite green); failure-path body carry at all three agent hops; the warning-clobber regression; every new sanitizer note; byte-accurate capping; the escalation path; `errorLog` combination and dedupe; and the degraded-vs-failed banner styling.

**Results:** 251 API tests green (10 files, serial), 35 web green, full `go test -race ./...` clean, `go vet` clean, Windows cross-compile clean, `tsc --noEmit` clean, `astro check` 0 errors, eslint clean.

Not run: RLS/integration contract suites (separate configs, need a live DB) — this PR adds no column, migration or policy, so neither contract's inputs changed.

## Residual risk

- **No end-to-end run against a real Windows endpoint.** The wire shape is pinned in both directions and each hop is tested, but the first live confirmation will be a real multi-volume backup on Windows — ideally one forced to fail, to exercise the failure-path body carry.
- **Field agents are unchanged until they update.** `unprotectedVolumes` will be absent from older agents' payloads; the sanitizer and UI both treat that as "not reported", never "none". Absence of VSS metadata is never evidence of a clean snapshot.
- **Out of scope, worth its own issue:** Codex suggested binding `orgId`/`deviceId` into the `applyBackupCommandResultToJob` UPDATE predicate (it currently keys on job id + status alone, under a system context). A genuine defence-in-depth improvement, but it changes a tenancy-sensitive predicate on paths this PR does not otherwise touch (reconcile, provider-backed, hyperv, mssql).

### DIFF
--- a/agent/cmd/breeze-backup/backup_run_result_test.go
+++ b/agent/cmd/breeze-backup/backup_run_result_test.go
@@ -1,0 +1,92 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/breeze-rmm/agent/internal/backup"
+	"github.com/breeze-rmm/agent/internal/backup/vss"
+)
+
+// #3027: the generic marshalResult discards the job body whenever err != nil, so
+// every hard-failed backup reached the server as Stderr and nothing else. That
+// is the branch where VSS diagnostics matter MOST — "which writer wedged" is the
+// question you ask about a failure — and job.VSSMetadata plus job.Warning were
+// both already populated and both thrown away one frame before the wire.
+func TestMarshalBackupRunResultKeepsDiagnosticsOnFailure(t *testing.T) {
+	job := &backup.BackupJob{
+		ID:            "job-1",
+		Status:        "failed",
+		FilesBackedUp: 12,
+		ErrorCount:    3,
+		Warning:       "VSS shadow copy could not be created",
+		VSSMetadata: &vss.VSSMetadata{
+			ShadowCopyID:       "{set-1}",
+			UnprotectedVolumes: []string{`D:\`},
+			Writers:            []vss.WriterStatus{{Name: "NTDS", State: "failed"}},
+		},
+	}
+
+	result := marshalBackupRunResult(job, errors.New("upload destination unreachable"))
+
+	// The run still fails — the body only adds detail.
+	if result.Success {
+		t.Fatal("a failed run must not be reported as successful")
+	}
+	if !strings.Contains(result.Stderr, "upload destination unreachable") {
+		t.Fatalf("failure reason lost from stderr: %q", result.Stderr)
+	}
+
+	var decoded map[string]any
+	if err := json.Unmarshal([]byte(result.Stdout), &decoded); err != nil {
+		t.Fatalf("failed run must carry a parseable job body, got %q: %v", result.Stdout, err)
+	}
+	if decoded["vssMetadata"] == nil {
+		t.Errorf("vssMetadata missing from a failed run's body: %s", result.Stdout)
+	}
+	if decoded["warning"] != "VSS shadow copy could not be created" {
+		t.Errorf("warning missing from a failed run's body: %s", result.Stdout)
+	}
+	// The partial counters ride the same body and were lost with it.
+	if decoded["errorCount"] != float64(3) {
+		t.Errorf("errorCount missing from a failed run's body: %s", result.Stdout)
+	}
+}
+
+// A nil job (the early returns before the job exists — no provider, no paths,
+// already running) must still deliver the reason rather than an empty body.
+func TestMarshalBackupRunResultWithNoJob(t *testing.T) {
+	result := marshalBackupRunResult(nil, errors.New("backup provider is required"))
+
+	if result.Success {
+		t.Fatal("expected failure")
+	}
+	if result.Stdout != "" {
+		t.Errorf("expected no body when there is no job, got %q", result.Stdout)
+	}
+	if !strings.Contains(result.Stderr, "backup provider is required") {
+		t.Errorf("failure reason lost: %q", result.Stderr)
+	}
+}
+
+func TestMarshalBackupRunResultSuccessIsUnchanged(t *testing.T) {
+	job := &backup.BackupJob{ID: "job-ok", Status: "completed", FilesBackedUp: 4}
+
+	result := marshalBackupRunResult(job, nil)
+
+	if !result.Success {
+		t.Fatal("a clean run must report success")
+	}
+	if result.Stderr != "" {
+		t.Errorf("success must not carry stderr, got %q", result.Stderr)
+	}
+	var decoded map[string]any
+	if err := json.Unmarshal([]byte(result.Stdout), &decoded); err != nil {
+		t.Fatalf("success body must parse, got %q: %v", result.Stdout, err)
+	}
+	if decoded["filesBackedUp"] != float64(4) {
+		t.Errorf("success body lost its counters: %s", result.Stdout)
+	}
+}
--- a/agent/cmd/breeze-backup/main.go
+++ b/agent/cmd/breeze-backup/main.go
@@ -620,7 +620,7 @@ func executeCommand(req backupipc.BackupCommandRequest, mgr *backup.BackupManage
 				SnapshotID: snapshotID,
 			})
 		})
-		result := marshalResult(mgr.RunBackupContext(ctx, excludes))
+		result := marshalBackupRunResult(mgr.RunBackupContext(ctx, excludes))
 		// Auto-sync to vault after successful backup (async — don't block command response)
 		if result.Success {
 			go autoSyncToVault(result.Stdout, vaultState, conn)
@@ -708,6 +708,37 @@ func ok(stdout string) backupipc.BackupCommandResult {
 
 func fail(msg string) backupipc.BackupCommandResult {
 	return backupipc.BackupCommandResult{Success: false, Stderr: msg}
+}
+
+// marshalBackupRunResult is marshalResult for backup_run specifically, differing
+// on ONE point: a failed run still carries its job body.
+//
+// The generic marshalResult throws `v` away when err != nil, so on every hard
+// failure the server received Stderr and nothing else. That silently defeated
+// #3027 on the branch where the diagnostics matter most: job.VSSMetadata and
+// job.Warning were both built by then — recording, say, that no shadow copy
+// could be created and which writers were wedged — and both were discarded one
+// frame before the wire. The counters (filesBackedUp, errorCount) went with
+// them.
+//
+// Success stays false and Stderr still carries the failure reason, so the server
+// still records the job `failed` (routes/agentWs.ts gates on
+// `result.status === 'completed'`). The body only adds detail to a failure that
+// was going to be a failure regardless. A body that cannot be marshalled is
+// simply omitted — a marshalling problem must never escalate into losing the
+// failure reason itself.
+func marshalBackupRunResult(job *backup.BackupJob, err error) backupipc.BackupCommandResult {
+	if err == nil {
+		return marshalResult(job, nil)
+	}
+	result := fail(err.Error())
+	if job == nil {
+		return result
+	}
+	if data, merr := json.Marshal(job); merr == nil {
+		result.Stdout = string(data)
+	}
+	return result
 }
 
 func marshalResult(v any, err error) backupipc.BackupCommandResult {
--- a/agent/internal/backup/backup.go
+++ b/agent/internal/backup/backup.go
@@ -328,6 +328,15 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 				"elapsedMs", time.Since(vssStart).Milliseconds(),
 				"error", vssErr.Error(),
 			)
+			// #3027: the worst VSS outcome used to be the quietest one. When the
+			// session never starts there is no VSSMetadata to send, so without
+			// this the run reaches the server indistinguishable from a clean
+			// VSS-backed backup — every path read live, every locked file at
+			// risk of being skipped, and nothing anywhere saying so. The
+			// per-path warning further down only fires when a session DID
+			// start, so it cannot cover this branch.
+			appendWarning(job, "VSS shadow copy could not be created; every path was read from the live volume, "+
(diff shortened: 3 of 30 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (3 agents, parallel). The design direction was separately quorum-checked against a read-only Codex `xhigh` opinion before implementing, per the CLAUDE.md advisor rule; we independently agreed on direction (a).

**Findings:** 12 raised → 12 addressed across `51c8431e9`, `e97717380`, `b6e564f0a`; 0 outstanding. Four were consequential enough that the original commit did **not** hold end to end:

- **The fix reached the server on no failed run at all.** `marshalResult` discards the job body whenever `err != nil`, and two further hops (the forwarder's error mapping, and the async handler's `else if` treating stderr and body as either/or) dropped it again. The claim that diagnostics land on both branches was false until all three were fixed — and the failure branch is where "which writer wedged" actually gets asked.
- **`backup.go` assigned `job.Warning` instead of appending**, destroying the new total-VSS-failure note whenever system-state collection was also incomplete — which co-occur, since a wedged writer subsystem causes both. That branch has no `VSSMetadata`, so the warning was its only channel.
- **The one hop with zero coverage was the hand-maintained agentWs forwarding literal.** pr-test-analyzer proved the whole feature could be deleted with 202/202 green. Now covered by a *completeness* contract (every `backupProcessResultSchema` key must appear in the literal), verified by mutation — this retires the bug class, not just this instance.
- **`sanitizeVssMetadata` dropped ill-typed values silently**, so `unprotectedVolumes: [{volume:'D:\\'}]` would have reduced to `[]` and made the UI report a clean snapshot — this issue's exact failure mode, one layer up. Every loss path now names itself in the persisted warnings.

Also fixed: array cap 200→24 (the two string arrays could otherwise blow the 64 KiB budget alone, forcing the tier that discards `unprotectedVolumes`); size measured in bytes rather than UTF-16 units; unusable blobs escalate to Sentry instead of vanishing; redaction on the pathological exit; and the device tab now reads `errorLog`, without which a total VSS failure rendered nothing at all.

Verified correct by the code reviewer and left unchanged: the `z.unknown()` failure-isolation decision at both boundaries, the reconcile path's inability to clobber a recorded blob, and that `vss_metadata` needs no cascade/export-policy registration (pre-existing column, already `excludedOpen`).

**Tests:** 252 affected API tests green single-fork; broad sweep of `routes/backup` + `jobs` + `services` + `agentWs` green (9032 passed / 23 skipped); 35 web green; full `go test -race ./...` clean; `go vet` clean; Windows cross-compile clean; `tsc --noEmit` clean; `astro check` 0 errors; eslint clean.

**CI:** all required jobs green on `b6e564f0a` — Test API, Test Web, Test Agent, Test Agent (race), Test Agent (Windows), Type Check, Lint, Smoke Test, Windows Runtime Smoke. `Integration Tests` shards 1/3/4 pass; **shard 2 red on infra only** — it died in "Start test-stack containers" pulling `redis-test`: `Get "https://registry-1.docker.io/v2/": Client.Timeout exceeded`. An earlier `Test Agent (Windows)` red was a 600s named-pipe timeout in `internal/sessionbroker`, a package this PR never touches (and which cannot import the ones it does); it passes on the current commit.

**Not run locally:** the RLS/integration contract suites (separate vitest configs, need a live DB). This PR adds no column, migration or policy, so neither contract's inputs changed — and shards 1/3/4 exercised them in CI regardless.

**Open for the maintainer:**
- No end-to-end run against a real Windows endpoint yet — worth one multi-volume backup, ideally forced to fail, to exercise the failure-path body carry.
- Codex suggested binding `orgId`/`deviceId` into the `applyBackupCommandResultToJob` UPDATE predicate. Deliberately **not** done here: it changes a tenancy-sensitive predicate on paths this PR doesn't otherwise touch (reconcile, provider-backed, hyperv, mssql). Worth its own issue.

**Status:** review-clean, awaiting maintainer merge. Issue #3027 left open and assigned.

PR #3063, opened 2026-08-03
feat: enrollment idempotency — collisions detect instead of block, uninstall intent reaper, exactly-once bootstrap refund (#2764)

### LINKED ISSUE #2764: Enrollment 409 (hostname collision) rolls back the entire MSI install; uninstall strands an active device row that guarantees reinstall failure
## Summary

When key-based enrollment hits the hostname-collision guard (`409 hostname_collision_requires_existing_device_token`, `apps/api/src/routes/agents/enrollment.ts`), the `EnrollAgent` custom action exits non-zero and — because it is `Return="check"` — **the entire MSI install rolls back with 1603 in ~3 seconds**.

Combined with the fact that **MSI uninstall deletes `secrets.yaml`/`agent.yaml` (the device token) but leaves the server-side device row ACTIVE**, this makes reinstall failure *deterministic*: every machine that was ever uninstalled, reimaged, or had a previous install rolled back will hard-fail all future enrolled installs until an operator decommissions the device row or uses the replacement workflow. Brand-new machines install fine — which presents in the field as "the MSI doesn't install reliably."

The guard itself is correct security hardening (SR-era; prevents device adoption/impersonation via a leaked enrollment key) and is not new in v0.100.0. The problem is the failure UX around it.

## Reproduction (Server 2022, v0.100.0 release MSI, live test server)

1. Install with `SERVER_URL`/`ENROLLMENT_KEY`/`ENROLLMENT_SECRET` → enrolls, device online. ✔
2. Uninstall (removes secrets.yaml by design) or wipe `C:\ProgramData\Breeze`.
3. Reinstall with the same properties → `enroll-last-error.txt`:
   `Enrollment failed: http 409 ... hostname_collision_requires_existing_device_token`
   → `Note: 1: 1722 2: EnrollAgent` → `MainEngineThread is returning 1603`. ✘ Every retry fails identically.
4. Decommission the device row → same install succeeds instantly. ✔ (#896 path works.)

Compounding on the portal-download path: the filename bootstrap token slot is consumed at redeem *before* enrollment 409s, so each retry burns a token use and subsequent attempts flip to `404 token invalid/expired/used` — a different error per attempt.

## Suggested directions (some combination)

1. **Agent**: treat this specific 409 reason as a soft failure in `enroll`/`bootstrap` when invoked from the MSI (exit 0 after writing all four diagnostic sinks + a clear Event Log entry naming the replacement workflow). The install lands, the agent idles unenrolled, and the operator gets an actionable signal instead of a cryptic 1603 rollback.
2. **Installer/agent**: best-effort server-side decommission/release on MSI uninstall (agent still has its token at that point), so uninstall→reinstall works without dashboard surgery.
3. **Bootstrap ordering**: don't consume the bootstrap token slot until enrollment succeeds (or refund on 409), so retries don't mutate the error.
4. **Docs**: deployment guide note — reinstalling on a known hostname requires decommission or the replacement workflow.

ToddHebebrand:
Implements the enrollment-idempotency design (spec: `docs/superpowers/specs/installer-enrollment/2026-08-02-enrollment-idempotency-design.md`, plan: `docs/superpowers/plans/installer-enrollment/2026-08-02-enrollment-idempotency.md`): a valid enrollment key always enrolls; hostname collisions become detection (fresh row + replacement linkage + audit + opt-in alert) instead of a 409/MSI rollback; uninstall stamps a reaper-driven decommission intent; failed enrolls refund their bootstrap slot exactly once.

Closes #2764

## What changed

- **Schema** (`2026-08-11-enrollment-idempotency.sql`, `2026-08-12` alert-template seed): `devices.uninstall_intent_at`, `devices.possible_replacement_of_device_id` (FK → devices, `ON DELETE SET NULL`), `enrollment_keys.bootstrap_token_id` (plain uuid — **no FK**, see deviations), partial index for the reaper scan; export-policy registrations included.
- **Bootstrap refund**: redeemed child keys carry their token id; new `POST /installer/bootstrap/cancel` deletes an unused child key and decrements `consumed_count` exactly once (`DELETE … WHERE usage_count = 0 RETURNING` is the guard; race-proven by an integration test with two concurrent cancels). Every outcome is logged.
- **Enrollment guard → detection** (`routes/agents/enrollment.ts`): collision without a valid device token now returns **201 with a fresh row**, `possibleReplacementOfDeviceId` set (first online ?? first non-decommissioned ?? oldest collider), audit `hostname_collision_enrolled_fresh_row` with complete `collidingDeviceIds`, best-effort alert when the collider is online. Quarantine 403 and the suspended-decommissioned 409 ops alarm are preserved (the 409 now only when *every* collider is decommissioned). A presented device token is matched against every colliding row, so reinstalls with a valid token still re-enroll in place instead of minting rows.
- **Uninstall intent**: `POST /agents/:id/uninstall-intent` (token-resolved write only), heartbeat clears the stamp, offlineDetector reaper decommissions after `UNINSTALL_INTENT_DECOMMISSION_HOURS` (default 24) with a TOCTOU-guarded UPDATE, terminal-status exclusion, and an audit event per reap.
- **Agent (Go + WiX)**: identity-conflict enroll failures get exit code 17 with a decommission-first remedy message; definitive 4xx enroll rejections cancel the bootstrap slot; `uninstall-notify` subcommand (always exits 0) wired into the MSI uninstall via a deferred `Return="ignore"` custom action sequenced before `RemoveFiles`.
- **Web**: replacement-review banner on the device page (ConfirmDialog-gated decommission action via `runAction`) + list-row badge; three serializer whitelist gaps fixed so the field survives list/detail transforms; 11 locale keys in all 7 locales.
- **Resolution is self-clearing**: decommissioning the old device (admin or reaper) nulls `possible_replacement_of_device_id` on rows pointing at it, so the banner/badge disappear once reviewed.

## Release notes

**Behavior change: hostname collisions no longer block enrollment.** Previously a reimaged/reinstalled machine that couldn't present the prior device's credential was refused with a 409 and the MSI rolled back, leaving the host permanently un-enrollable. Enrollment now always succeeds: a fresh device row is created, linked back to the row it may replace, and an audit event + best-effort alert are raised instead of a hard block. There is deliberately **no config flag** to restore the 409 — the spec records why (the block prevented nothing an attacker couldn't route around, while locking out the common reimage case).

**At-cap partners now see `DEVICE_LIMIT_REACHED` where a collision used to 409** — collisions are no longer a way to bypass (or be confused with) the device cap.

**New optional env var:** `UNINSTALL_INTENT_DECOMMISSION_HOURS` (default `24`). No action needed for self-hosters happy with the default.

**The "Device Identity Collision" alert template is seeded but dormant** until an org/partner binds an alert rule to it. The `agent.enroll` audit event is the always-on record.

**Agent-side pieces need a new agent build + fleet promote**: exit code 17, the bootstrap-slot refund, and uninstall-notify only activate on updated agents. The refund fires **only** on a definitive 4xx rejection (401/403/404/429/identity-conflict) where the slot is provably unused; network errors, 5xx, and unclassified failures never refund (the enroll may have succeeded server-side). Old agents keep working unchanged.

## Deviations from the plan text (adjudicated during execution)

1. **No FK on `enrollment_keys.bootstrap_token_id`** — the plan's verbatim FK created an `enrollment_keys ↔ installer_bootstrap_tokens` cycle that made `cascadeDeleteOrg()` throw for every org (caught by `tenantCascade.integration.test.ts`). Plain uuid column, app-side integrity; a dangling id after token deletion is harmless (the decrement UPDATE matches 0 rows).
2. **`bootstrap_token_id` classified `reviewedIncluded`** (not the plan's `included`) — the name trips `SUSPICIOUS_NAME_PARTS`; it is a reviewed non-secret operational identifier.
3. **Migration filename `2026-08-11-*`** — the migration history already extends past `2026-08-10`, so the plan's literal date prefix would not sort last.
4. **Collider selection refined** to `first online ?? first non-decommissioned ?? oldest` so a decommissioned oldest row can't shadow a live collider (which would have silently skipped detection or resurrected the un-enrollable 409).
5. **Uninstall-intent endpoint lives at `/agents/:id/uninstall-intent`** — the plan's id-less path would have missed the `/:id/*` agent-auth prefix and shipped unauthenticated. The URL id is inert; the write keys off the token-resolved device.
6. **Token matched against every colliding row** (extension) — prevents unbounded fresh-row minting on reinstall once rows share a hostname.
7. **Bootstrap/cancel rate limiting** rides on the global per-IP limiter — the plan referenced a per-route helper that doesn't exist (redeem has none either); parity holds and every miss is logged.

## Verification

- TDD throughout; per-task review + fix rounds, then a whole-branch final review (which caught a cross-task contract bug — the agent originally sent the wrong credential to `bootstrap/cancel` — fixed in `266c588ed` with a discriminating e2e test).
- New `enrollmentCollision.integration.test.ts` (real Postgres): reap → re-enroll fresh row; collision leaves the old row byte-identical (`SELECT *` compare); exactly-once refund under two concurrent cancels.
- Live-DB contract suites green locally: `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade` (one local-environment failure — later diagnosed as shared-test-DB pollution from an in-flight sibling branch, see #3064/#3066; main CI is green on this suite), `db:check-drift` clean, `go test -race ./...` green, web suite 4798 tests green on pinned Node 22.23.2, API tsc clean.

## Known edge cases (deliberate, documented)

- A bootstrap child key redeemed on an already-enrolled box is consumed without enrolling or refunding (rare; self-heals via token TTL).
- 400s/unrecognized 409s from enroll strand the slot rather than risk a double-free — conservative by design.
- Uninstall-intent is refused (403) during a tenant-offboarding drain; the agent treats it as non-fatal and the org cascade takes the row anyway.

## Release gate (not a merge gate)

- [x] **Real Windows MSI uninstall smoke test** — DONE 2026-08-03 on the Windows Server 2022 test VM against a local stack (see PR comment for evidence) before fleet promote: `msiexec /x` the package, grep the `/l*v` log for `UninstallNotify`, confirm `devices.uninstall_intent_at` stamped server-side. This is the package's first uninstall-path FileRef custom action; `Return="ignore"` bounds failure to a silent no-op, but the chain needs one real-machine pass.

## Follow-ups to file separately

- ~~Pre-existing: `tenantCascade.integration.test.ts` red on main~~ — diagnosed as shared-test-DB pollution, closed as #3064; root cause tracked in #3066.
- Pre-existing: `pnpm db:migrate` is a no-op (`autoMigrate()` exported but never self-invoked).
- Dev-rig: shared fixed-port test containers get clobbered by concurrent worktree sessions; consider per-worktree ephemeral containers.


ToddHebebrand:
Release-gate MSI smoke completed on a real Windows Server 2022 VM (WiX 4.0.5 build from this branch, install/uninstall against a local stack — no prod contact):

- **CA compiles + lands correctly**: `UninstallNotify` in the CustomAction table as Type 3154 (Exe-from-FileRef + Continue + Deferred + NoImpersonate), sequence 3499 — immediately before `RemoveFiles` (3500).
- **Uninstall chain works**: `msiexec /x` exit 0; deferred CA fired at 08:10:55.297, `RemoveFiles` at .528; server logged `POST /uninstall-intent → 200`; `devices.uninstall_intent_at` stamped (`2026-08-03 15:10:55.511+00`); agent log `acknowledged=true`.
- **`Return="ignore"` proven live**: repeated the uninstall with the API stopped — CA fired, call failed, uninstall still exited 0. The CA can never block an uninstall.
- **Reinstall after uninstall**: fresh row via the collision path (secrets deleted with the install → new identity), old stamped row left for the 24h reaper — exactly the designed flow. Heartbeat clear verified by code inspection (unconditional null in the heartbeat UPDATE).

One tooling note (not a PR defect): `installer/build-msi.ps1` now requires a fourth binary (`-UserHelperExePath`) — older build notes that pass only agent/backup/watchdog hard-fail.

PR #2988, opened 2026-08-01
fix(api): surface the proxy-trust misconfiguration on the TRANSPORT-001 path (#2987)

bdunncompany:
Addresses #2987 (fix 1/2 from that issue — the diagnosability half).

## The gap

`[proxy-trust] MISCONFIGURATION` already exists, and its wording already nails this exact failure, including the stale-CIDR-after-container-recreate case. It simply never fires on the path that matters most.

- `getTrustedClientIp` warns when forwarded headers arrive from an untrusted peer.
- `trustsForwardedHeadersFrom` — the gate TRANSPORT-001 introduced, and what `effectiveRequestScheme` calls — returned the same verdict **silently**.

## Why that gap is a total outage rather than a missing log line

1. A stale or incorrect `TRUSTED_PROXY_CIDRS` makes `trustsForwardedHeadersFrom` false for the reverse proxy.
2. `effectiveRequestScheme` therefore ignores `X-Forwarded-Proto` and falls back to the direct request URL — which, as its own doc comment says, is the internal plain-HTTP proxy→API hop. It reports `http` for **every** request.
3. Under `FORCE_HTTPS`, `securityMiddleware` 308s every non-health route and **short-circuits**, so nothing downstream ever calls `getTrustedClientIp`. The existing warning is structurally unreachable.
4. `/health` is in `HEALTH_CHECK_PATHS` and exempt, so it keeps returning `200` with the new version for the entire outage.

Agents cannot heartbeat and the SPA cannot load, while every signal an operator normally checks stays green. Observed on a proxy-fronted self-hosted deployment upgrading v0.102.0 → v0.103.0: ~50 minutes, zero agent check-ins, `/health` 200 throughout, and the only log evidence was `route=unmatched status=308`, which names neither the cause nor the peer.

## The change

`trustsForwardedHeadersFrom` now emits the same warning — reusing the same per-peer rate-limit state — when it rejects a peer that actually sent trust-gated forwarded headers.

**Trust semantics are unchanged.** It still returns exactly `shouldTrustProxyHeaders() && isTrustedProxySource(peer)`; only the observability differs.

Also adds `hasTrustGatedForwardedHeaders` = `hasForwardedIpHeaders` + `X-Forwarded-Proto`. The scheme header is deliberately kept out of `hasForwardedIpHeaders`, since that predicate answers "did this request carry client-IP attribution" — but a proxy forwarding only the scheme is still a proxy, and the FORCE_HTTPS path keys off that header alone.

[...]

Issue #2987, opened 2026-08-01
[API] v0.103.0 TRANSPORT-001: a stale/incorrect TRUSTED_PROXY_CIDRS turns into a total fleet + UI outage on upgrade, and /health cannot detect it


bdunncompany:
## Summary

Upgrading a proxy-fronted self-hosted deployment from v0.102.0 to v0.103.0 took **the entire agent fleet and the web UI offline simultaneously**, while `/health` continued returning `200` with the new version for the whole outage.

The trigger was a **pre-existing, previously-harmless** `TRUSTED_PROXY_CIDRS` value that pointed at the wrong container. Nothing about it was invalid; it passed every config check in v0.102.0 and v0.103.0 alike.

Root cause is the Wave 5 Task 8 (TRANSPORT-001) change in `middleware/security.ts` + `services/requestTransport.ts`. That change is correct on its merits — this report is about it converting a silent misconfiguration into a total outage with no diagnosable signal.

## Mechanism

Before v0.103.0 the HTTP→HTTPS redirect read `x-forwarded-proto` unconditionally, so a wrong `TRUSTED_PROXY_CIDRS` had no effect on routing.

In v0.103.0, `effectiveRequestScheme()` honors that header **only** when `trustsForwardedHeadersFrom(c)` is true — i.e. the immediate TCP peer is in `TRUSTED_PROXY_CIDRS`. Otherwise it falls back to the direct request URL, which the function's own doc comment describes accurately:

> in the standard Caddy-fronted topology that's the internal plain-HTTP Caddy→API hop, NOT the browser's transport

So when the reverse proxy is not in the trusted list, `effectiveRequestScheme()` returns `http` for **every** request, unconditionally. With `FORCE_HTTPS=true` that means a `308` on everything:

```
--> GET  route=unmatched status=308 duration_ms=0
--> POST route=unmatched status=308 duration_ms=0
```

Agents cannot POST heartbeats and browsers cannot load the SPA, from one cause.

## Why it is undetectable

`/health` is in `HEALTH_CHECK_PATHS` and is therefore **exempt from the redirect by design** (correctly — orchestrator probes arrive over plain HTTP). The consequence is that the one endpoint an operator checks after an upgrade is the one endpoint guaranteed to keep working.

Observed: `/health` returned `{"status":"ok","version":"0.103.0"}`, all containers healthy, migration ledger advanced exactly as expected, zero errors in the API log — while **zero** agents checked in for ~50 minutes.

[...]

PR #3067, opened 2026-08-03
fix(api): warn at the FORCE_HTTPS redirect itself (TRANSPORT-001) (#3047)

### LINKED ISSUE #3047: [API] TRANSPORT-001: a trusted proxy that never sends X-Forwarded-Proto still 308-loops silently — log at the redirect, not only at the trust gate
Follow-up to #2987, which #2988 (merged as `ce3b5c19`) closed only halfway.

**What #2988 covered.** `trustsForwardedHeadersFrom` in `services/clientIp.ts` now emits the `[proxy-trust]` warning on the TRANSPORT-001 path in two cases: an untrusted peer that sent trust-gated forwarded headers (the stale-`TRUSTED_PROXY_CIDRS` case #2987 reported), and — added during review — `TRUST_PROXY_HEADERS=false` with a proxy actually in front, which produces the identical outage through the neighbouring config value. Both warnings share one per-peer 15-minute suppression window, and the message now names the FORCE_HTTPS 308-loop symptom rather than only the rate-limit/audit-attribution ones.

**What is still silent.** A **trusted** proxy that never forwards `X-Forwarded-Proto` reproduces the same total outage with no log line anywhere. `trustsForwardedHeadersFrom` returns `true`, so neither warning branch is reachable; `effectiveRequestScheme` (`services/requestTransport.ts:51`) finds no scheme header to honor and falls back to the internal plain-HTTP proxy→API hop, reporting `http` for every request. Under `FORCE_HTTPS` that is byte-for-byte the outage in #2987 — every non-health route 308s, `/health` stays `200` — reached from a correct `TRUSTED_PROXY_CIDRS` and a proxy that is simply missing its `X-Forwarded-Proto` directive (an L4 load balancer in front, or a hand-rolled Caddy/nginx config without the header). No warning in `clientIp.ts` can catch this, because from the trust gate's point of view nothing is wrong.

**Proposed fix — log at the 308 itself.** `middleware/security.ts:149-158` issues the redirect and returns without logging anything under any condition. A rate-limited `warn` at that point, naming the immediate peer IP (`getImmediatePeerIpOrUndefined` is already imported in the codebase), whether the peer was trusted, and whether `X-Forwarded-Proto` was present, would cover this variant *and* both cases #2988 already handles — one log line at the point of user-visible impact, rather than three warnings spread across the trust gate that each have to anticipate a different cause. That is the structurally better place for the signal, and it is where I would put the remaining effort. Same per-peer suppression discipline as `clientIp.ts` so a redirect storm cannot flood the log.

**Also still open from #2987.** Suggestion 1 (boot-time cross-check — `config/validate.ts` never compares `FORCE_HTTPS` against `TRUST_PROXY_HEADERS`/`TRUSTED_PROXY_CIDRS`, and the validator's own advice text at `validate.ts:347` can steer a proxy-fronted operator into `TRUST_PROXY_HEADERS=false`); suggestion 4 (a `/health` field reporting whether recent requests arrived from a trusted peer, so the one endpoint operators check after an upgrade can actually fail). Suggestion 3 (release-note flag for TRANSPORT-001) is worth doing regardless of the code changes here.

**Related.** #2987 (original report), #2988 (`ce3b5c19`, diagnosability half). The 308 behavior itself is intentional fail-closed and is not in question — this is entirely about the missing signal.

Staying open — no code change yet.

bdunncompany:
Closes #3047. Picking this up since it was unassigned and is the other half of the outage I reported in #2987 — happy to hand it back if you'd rather take it.

## The gap #2988 left

`trustsForwardedHeadersFrom` only warns when the peer is **not** trusted. The variant in this issue stays completely silent: a **trusted** proxy that never sends `X-Forwarded-Proto`. The trust gate passes, `effectiveRequestScheme` finds no scheme header and falls back to the internal plain-HTTP proxy→API hop, and under `FORCE_HTTPS` every non-health route 308s while `/health` (exempt) keeps returning `200` — all from a perfectly correct `TRUSTED_PROXY_CIDRS`.

Per your framing in the issue, the signal belongs at the redirect rather than the trust gate: one line at the point of user-visible impact instead of N warnings that each have to anticipate a different cause. This also covers the untrusted-peer and `TRUST_PROXY_HEADERS=false` cases #2988 already handles.

## What the warning says

Enough to disambiguate the causes without reading the source:

| field | why |
|---|---|
| `peer` | the immediate TCP peer, to compare against `TRUSTED_PROXY_CIDRS` |
| `trustedProxy` | `false` = CIDR list doesn't cover this peer, or `TRUST_PROXY_HEADERS` is off |
| `xForwardedProto` | `(absent)` **with** `trustedProxy=true` is exactly the variant no `clientIp.ts` warning can reach |
| `canonicalHost` | distinguishes the 400 branch from the 308 |

It also states the symptom — fleet-wide 308s while `/health` stays 200 — because that combination is what makes this look like anything *except* a proxy misconfiguration. In the original incident every signal an operator normally checks stayed green for ~50 minutes.

**Also warns on the 400 branch.** An unrecognized `Host` is the same misconfiguration through a different lens and was equally silent.

**Suppression** is per peer on a 15-minute window with a bounded map, matching `clientIp.ts`. A redirect storm is precisely the situation here, so an unsuppressed log would flood.

## Deliberately not included

Suggestion 1 (boot-time cross-check of `FORCE_HTTPS` against the proxy-trust settings) and suggestion 4 (a `/health` trusted-peer field). Both change boot or health-check semantics, which reads like your call rather than mine — happy to do either on your say-so. Suggestion 3 (release-note flag) is independent of code.

## Local verification

- `security.test.ts`: **31 pass**, 7 new — the redirect warn, the `(absent)` variant, the 400 branch, per-peer suppression, health-path exemption, and `FORCE_HTTPS` off.
- Confirmed the tests catch the regression: removing the call fails **4** of them.
- Full `@breeze/api` suite: **1235 files, 19452 tests passed, 0 failures**.
- `tsc --noEmit` and eslint clean on both changed files.

### DIFF
--- a/apps/api/src/middleware/security.test.ts
+++ b/apps/api/src/middleware/security.test.ts
@@ -279,20 +279,44 @@ describe('securityMiddleware', () => {
 describe('FORCE_HTTPS redirect warning (#3047)', () => {
   const opts = { forceHttps: 'true', publicApiUrl: 'https://api.example.com' };
   let warnSpy: ReturnType<typeof vi.spyOn>;
+  const originalCidrs = process.env.TRUSTED_PROXY_CIDRS;
 
   beforeEach(() => {
+    // Model a configured deployment. With TRUSTED_PROXY_CIDRS unset,
+    // `isTrustedProxySource` returns `NODE_ENV !== 'production'` — so under test
+    // every peer is trusted and the proxy-evidence gate can never be observed.
+    // Naming an explicit CIDR the test peer is not in reproduces the production
+    // semantics this suite is about.
+    process.env.TRUSTED_PROXY_CIDRS = '10.9.9.9/32';
     _resetForceHttpsRedirectWarnStateForTests();
     warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
   });
 
   afterEach(() => {
+    if (originalCidrs === undefined) delete process.env.TRUSTED_PROXY_CIDRS;
+    else process.env.TRUSTED_PROXY_CIDRS = originalCidrs;
     warnSpy.mockRestore();
     _resetForceHttpsRedirectWarnStateForTests();
   });
 
-  it('warns when a redirect fires, naming the symptom and the two config causes', async () => {
+  // A direct plain-HTTP client with no forwarding headers at all is the redirect
+  // working as designed — a port-80 scanner, an uptime probe, or a deploy that
+  // intentionally runs without a reverse proxy. Warning there would emit a
+  // misconfiguration line per scanner IP on any internet-facing origin, and
+  // per-peer suppression is no defence because each source IP gets its own first
+  // warn. It must redirect silently.
+  it('stays silent for a direct client with no proxy evidence', async () => {
     const res = await createApp(opts).request('http://api.example.com/test', {
       headers: { host: 'api.example.com' },
+    });
+
+    expect(res.status).toBe(308);
+    expect(warnSpy).not.toHaveBeenCalled();
+  });
+
+  it('warns when a redirect fires, naming the symptom and the two config causes', async () => {
+    const res = await createApp(opts).request('http://api.example.com/test', {
+      headers: { host: 'api.example.com', 'x-forwarded-for': '203.0.113.9' },
     });
 
     expect(res.status).toBe(308);
@@ -307,8 +331,10 @@ describe('FORCE_HTTPS redirect warning (#3047)', () => {
   // The variant #2988 could not catch: trust gate passes, no scheme header, so
   // no warning branch in clientIp.ts is reachable.
   it('reports xForwardedProto=(absent) when the proxy never sends the header', async () => {
+    // Evidence of a proxy via the IP header, but no scheme header — exactly the
+    // variant #2988 could not catch.
     await createApp(opts).request('http://api.example.com/test', {
-      headers: { host: 'api.example.com' },
+      headers: { host: 'api.example.com', 'x-forwarded-for': '203.0.113.9' },
     });
 
     const msg = String(warnSpy.mock.calls[0]?.[0]);
@@ -317,7 +343,7 @@ describe('FORCE_HTTPS redirect warning (#3047)', () => {
 
   it('also warns on the 400 branch, which is the same fault via a non-canonical Host', async () => {
     const res = await createApp(opts).request('http://wrong-host.example.net/test', {
-      headers: { host: 'wrong-host.example.net' },
+      headers: { host: 'wrong-host.example.net', 'x-forwarded-for': '203.0.113.9' },
     });
 
     expect(res.status).toBe(400);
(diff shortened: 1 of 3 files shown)


ToddHebebrand (changes requested):
The shape is right — warning at the redirect/reject site itself with the trusted/proto diagnosis in one line is exactly what #3047 asked for, the suppression map is bounded, no redirect behavior changed, and the test coverage of the 400 branch / health exemption / suppression window is solid. One gate is missing before this can land:

**Must-fix — the warn fires with zero proxy evidence.** `warnForceHttpsRedirect(c, canonicalHost)` is called unconditionally whenever the scheme resolves to `http`. A plain direct HTTP client — port-80 scanners, uptime probes, any self-hosted deploy with no reverse proxy — hits this on every request, and the per-peer suppression doesn't help against distinct source IPs: the map-bounding comment itself notes "an unbounded burst still warns — it just isn't tracked." On an internet-facing origin that's a multi-line MISCONFIGURATION-flavored warn per scanner IP for behavior that is the feature working as intended (http→https redirect). The existing pattern in `clientIp.ts` gates its warns on actual proxy evidence (`hasTrustGatedForwardedHeaders`) for exactly this reason. Gate the call on `trustsForwardedHeadersFrom(c) || hasTrustGatedForwardedHeaders(c)` — that keeps both misconfiguration variants the issue describes (trusted proxy omitting the header; proxy present but untrusted) and silences the direct-client case. The first test in the new suite ("plain request, no proxy headers, trustedProxy=false" expects a warn) encodes the current behavior as intended — it should flip to asserting silence.

**Nice-to-have, not blocking:** the suppression/eviction block is a near-verbatim copy of `shouldEmitProxyTrustWarnNow` in `clientIp.ts` — worth extracting a shared helper; and `trustsForwardedHeadersFrom(c)` is evaluated a second time inside the warn after `effectiveRequestScheme` already called it (harmless, just redundant).

Status: REQUEST CHANGES on the one gate; happy to merge once it's in.

bdunncompany:
Fixed and pushed (`53641795d`). You were right, and the failure mode was worse than "noisy": on an internet-facing origin it is one misconfiguration-flavoured line per scanning IP, for the redirect working as designed.

**Gate added** — `trusted || hasTrustGatedForwardedHeaders(c)`, exactly as you suggested. Both variants #3047 describes still warn; the direct-client case is silent. The bail sits *before* the suppression map is touched, so a scanner sweep can't evict real proxy peers from it either.

Worth noting `hasTrustGatedForwardedHeaders` is the helper I added in #2988 for precisely this decision, and I didn't reach for it here. Exported it now.

**Redundant call** folded in while I was there — `trustsForwardedHeadersFrom(c)` resolves once at the call site and is passed in.

**On the tests, two things you'll want to know about:**

The inverted first test failed at first, and not because the gate was wrong. With `TRUSTED_PROXY_CIDRS` unset, `isTrustedProxySource` returns `NODE_ENV !== 'production'` — so under test *every* peer is trusted and the gate can never be observed. The suite now sets an explicit CIDR to model a configured deployment. That felt worth fixing properly rather than asserting around.

The two pre-existing negative tests (health-probe exemption, FORCE_HTTPS off) now carry proxy evidence deliberately. Without it they'd have gone green because the new gate silenced them, not because of the exemption they claim to cover. Same trap as the inert `toolTimeouts` assertions on #3096 — a test that can't fail for the reason it names.

Added a case for a lone `X-Forwarded-Proto`, since that's the whole reason `hasTrustGatedForwardedHeaders` is a superset of the IP-header check.

**Left out:** the shared suppression/eviction helper. It's a genuine duplication of `shouldEmitProxyTrustWarnNow`, but it touches a second module and you flagged it non-blocking, so I kept this diff to the gate. Say the word and it's a quick follow-up.

Rebased onto current main. Local gate green on node 22.23.2: `tsc --noEmit` clean, `vitest run` in apps/api **1253 files / 19708 tests passed, 0 failed**, eslint clean on all three changed files.

ToddHebebrand (approved):
**Your push didn't land — the PR branch never got `53641795d`.** The commit made it to the repo's object store (I could fetch it by SHA), but `fix/3047-log-at-redirect` was still pointing at `9f64359` from 08-03 — since you rebased, the push needed `--force-with-lease` and was presumably rejected as non-fast-forward. I've force-updated the branch to `53641795d` for you (it was exactly the commit your comment described, parented on current main), so CI is running on the real fix now.

**Review of the fix itself: approved.** The gate is right where it should be — bailing before the suppression map is touched closes the scanner-eviction hole, and passing `trusted` down from the call site instead of recomputing is the correct economy. The test-env trap you hit (`isTrustedProxySource` defaulting to trust-everything outside production, making the gate unobservable) was well worth fixing properly with an explicit CIDR; the deliberate proxy evidence on the two pre-existing negative tests is the right instinct — a green test that's green for the wrong reason is worse than no test.

The shared suppression/eviction helper stays non-blocking as agreed — happy to take it as a follow-up whenever.

Merging once CI is green.

PR #3070, opened 2026-08-03
fix(api): make pnpm db:migrate actually run autoMigrate() (#3065)

### LINKED ISSUE #3065: [API] pnpm db:migrate is a no-op — autoMigrate() is exported but never invoked when run as a script
## Description
`pnpm db:migrate` completes successfully but applies nothing. Discovered during #2764 verification (Task 1 had to call `autoMigrate()` directly to apply a new migration) and confirmed identical on `main`.

## Root Cause (if known)
`apps/api/src/db/autoMigrate.ts` exports `autoMigrate()` but has no self-invocation entrypoint when executed by the `db:migrate` package script, so the script exits 0 having done nothing. Anyone relying on `pnpm db:migrate` locally (per the README/CLAUDE.md dev workflow) silently gets a stale schema; the API's boot-time migration path is what actually applies migrations.

## Proposed Fix
Give the script a real entrypoint (e.g. a small `scripts/migrate.ts` that awaits `autoMigrate()` and exits non-zero on failure), or fix the invocation in `package.json`. Add a regression check so the script actually applies a pending migration in CI or a unit-level smoke.

## Affected Files
- `apps/api/src/db/autoMigrate.ts`
- `apps/api/package.json` (`db:migrate` script)

ToddHebebrand:
## Summary
`pnpm db:migrate` ran `tsx src/db/autoMigrate.ts`, which only defines/exports `autoMigrate()` — no invocation. The command exited 0 having applied nothing, so anyone following the documented dev workflow (or `scripts/prod/deploy.sh` step 4) silently got a stale schema; only the API boot path actually applied migrations.

## Fix
- New dedicated entry file `apps/api/scripts/migrate.ts` (same structure as `scripts/check-drift.ts` for `db:check-drift`): unconditionally awaits `autoMigrate()`, `process.exit(0)` on success, logs + `process.exit(1)` on failure. `db:migrate` now points at it.
- `process.exit` is deliberate — autoMigrate's auto-seed step opens the shared pool from `db/index`, which would otherwise hold the event loop open and hang the command after success.
- A first iteration used the `seed.ts`-style main-module guard (`import.meta.url === \`file://${process.argv[1]}\``); review verified empirically that it fails open (silent exit-0 no-op — the exact #3065 failure mode) on paths needing URL percent-encoding, symlinked checkouts, and Windows. The dedicated entry file has zero such failure modes and keeps the library module import-safe for the API boot path (`databaseStartup`).

## Regression guard
New `db:migrate entrypoint (#3065)` block in `apps/api/src/db/autoMigrate.test.ts`: resolves the entrypoint from `package.json`'s `db:migrate` script and asserts it is a dedicated file (not the library module), calls `autoMigrate()` with no main-module guard, and exits 0/1 by outcome. Runs in the required Test API job.

## Verification
- `pnpm exec vitest run src/db/autoMigrate.test.ts` — 56 passed (includes the 2 new tests).
- `tsc --noEmit` clean on apps/api.
- Live against the local dev DB (`postgresql://breeze:breeze@localhost:5432/breeze`, all migrations already applied): `pnpm db:migrate` enumerates the migrations dir, detects `normal` state, verifies checksums, prints `All migrations already applied` / `Database already seeded`, and terminates on its own with exit 0 (previously: no runner output at all, vacuous exit 0). Error path also observed live: with the DB unreachable it printed the full stack via the `.catch` and exited non-zero.

No migration files created or modified.

Closes #3065

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -7,7 +7,7 @@
     "build": "tsup",
     "clean": "rm -rf dist",
     "db:check-drift": "tsx scripts/check-drift.ts",
-    "db:migrate": "tsx src/db/autoMigrate.ts",
+    "db:migrate": "tsx scripts/migrate.ts",
     "db:seed": "tsx src/db/seed.ts",
     "db:seed:e2e": "tsx src/db/seedE2eFixtures.ts",
     "check:migrations": "tsx scripts/check-migrations.ts",
--- a/apps/api/scripts/migrate.ts
+++ b/apps/api/scripts/migrate.ts
@@ -1,0 +1,21 @@
+// Dedicated entrypoint for `pnpm db:migrate` (#3065).
+//
+// autoMigrate.ts is a library module imported by the API boot path
+// (databaseStartup), so it must not self-execute — and an "am I the main
+// module?" guard comparing import.meta.url against process.argv[1] fails
+// open on paths needing URL percent-encoding (spaces, non-ASCII), symlinked
+// checkouts, and Windows, silently reproducing the #3065 exit-0 no-op. A
+// dedicated entry file that calls autoMigrate() unconditionally has no such
+// failure mode (same structure as scripts/check-drift.ts for db:check-drift).
+//
+// process.exit is required: autoMigrate's auto-seed step opens the shared
+// pool from src/db/index, which would otherwise hold the event loop open
+// forever after a successful run.
+import { autoMigrate } from '../src/db/autoMigrate';
+
+autoMigrate()
+  .then(() => process.exit(0))
+  .catch((err) => {
+    console.error('[auto-migrate] Migration failed:', err);
+    process.exit(1);
+  });
--- a/apps/api/src/db/autoMigrate.test.ts
+++ b/apps/api/src/db/autoMigrate.test.ts
@@ -272,6 +272,36 @@ CREATE INDEX CONCURRENTLY IF NOT EXISTS bar_idx ON t (b);`;
   });
 });
 
+describe('db:migrate entrypoint (#3065)', () => {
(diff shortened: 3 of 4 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer, silent-failure-hunter)
**Findings:** 3 raised → all addressed in 0910af588; 0 outstanding.
- code-reviewer + silent-failure-hunter (both verified empirically under the repo's tsx): the initial `import.meta.url === \`file://${process.argv[1]}\`` main-module guard fails open on percent-encodable paths (spaces/non-ASCII), symlinked checkouts, and Windows — silent exit-0 no-op, i.e. the #3065 failure mode again. Fixed by replacing the guard with a dedicated `scripts/migrate.ts` entry file that calls `autoMigrate()` unconditionally (the zero-failure-mode option both reviewers recommended; matches the `db:check-drift` sibling pattern).
- code-reviewer: the regression test pinned the fragile guard literal verbatim. Fixed — the test now asserts intent (dedicated file, unconditional invocation, no main-module guard, exit-by-outcome).
- pr-test-analyzer: assert `process.exit(0)` on the success path too (it is load-bearing — the auto-seed step's shared pool otherwise hangs the command). Added.

**Tests:** `pnpm exec vitest run src/db/autoMigrate.test.ts` — 56 passed; `tsc --noEmit` clean on apps/api. Live no-op verification against the local dev DB: enumerates + checksums all migrations, `All migrations already applied`, exits 0 and terminates (no pool hang). Error path observed live (DB unreachable under load): full stack logged, non-zero exit.

Issue #3064, opened 2026-08-03
[API] tenantCascade.integration.test.ts is red on main — software_upload_sessions missing from the org cascade order


ToddHebebrand:
## Description
`tenantCascade.integration.test.ts` currently fails on `main` (Integration Tests job / local live-DB run): `software_upload_sessions` (surfaced alongside `extension_org_installs` in the failure output) is not covered by `getOrgCascadeDeleteOrder()`. Found during the #2764 branch verification and reproduced byte-identical on `origin/main` via a clean temp worktree — it is **not** introduced by that branch.

## Root Cause (if known)
A tenant-scoped table with an `org_id` column was added without registering it in `CORE_ORG_CASCADE_DELETE_ORDER` (`apps/api/src/services/tenantCascade.ts`) — the exact latent GDPR org-erasure gap the contract test exists to catch (org delete either strands rows under a dead tenant or aborts on an FK violation). Per CLAUDE.md this is the 6th occurrence of this class (#1359, #1351, #1365, #2179, #2514); the contract test caught it, but because the suite only runs in Integration Tests (required on main, non-blocking on PRs), main went red after merge.

## Proposed Fix
Register the missing table(s) in `CORE_ORG_CASCADE_DELETE_ORDER` (alphabetical, FK children before parents) and classify any columns in `CORE_TENANT_EXPORT_POLICY` if not already present. Run the full live-DB contract suites locally before merging.

## Affected Files
- `apps/api/src/services/tenantCascade.ts` (primary)
- `apps/api/src/services/tenantExportPolicyRegistry.ts` (likely)

ToddHebebrand:
**Diagnosis:** aborting the fix — the premise doesn't hold on `main`. This is shared-test-DB pollution, not a missing registration on `main`.

**Evidence**
- `software_upload_sessions` does not exist anywhere on `origin/main` (no schema file, no migration). It exists only on the in-flight branch `ToddHebebrand/issue-2951-software-library-large` (issue #2951), whose migration `2026-08-11-software-upload-sessions.sql` was applied to the shared test DB on :5433 (confirmed in that DB's `breeze_migrations` ledger). That branch **already registers the table** in both `CORE_ORG_CASCADE_DELETE_ORDER` and `CORE_TENANT_EXPORT_POLICY`, so nothing is missing once it lands.
- `extension_org_installs` came from the `todd/extension-org-installs` branch (which likewise carries its own export-policy registration) and is no longer present in the shared DB.
[...]

Issue #2951, opened 2026-07-30
[Software Library] Large file uploads (~100MB+) via dashboard fail with "signal is aborted without reason" — browser session token expires mid-transfer


cisspUser01:
### Description

Uploading a software package file directly through the Software Library
dashboard ("Upload file" source option) fails for larger files, even though
the same file uploads successfully when fetched server-side via the
"Download URL" option instead. Small files (a few MB) upload without issue;
a 222MB MSI consistently fails partway through.

### Steps to Reproduce

1. Go to Software Library → Add software package (or add a new version to
   an existing package).
2. Choose "Upload file" as the source.
3. Select a file around 200MB+ (reproduced with a 222MB / 227,672,306-byte
   Elastic Agent MSI).
4. Start the upload and let it run.
5. Around a fixed low percentage into the upload (observed consistently
   stalling at 10%), the browser shows: "signal is aborted without reason"
   with a "Try again" button. Retrying reproduces the same failure.

### Expected Behavior

A large file upload through the dashboard should either complete
successfully, or fail with a clear, actionable error (e.g. "session expired,
please log in again and retry") rather than a generic client-side abort
message with no indication of the real cause.

### Actual Behavior

- Small files (tested with a 2.3MB exe) upload successfully every time.
- The same 222MB file fails consistently via "Upload file," but succeeds
  every time via "Download URL" (server fetches the file itself, bypassing
  the browser entirely) — isolating the problem specifically to the
  browser-mediated upload path, not the file, MinIO, or the backend's
  storage handling.
- API logs show repeated entries timed to match each failed attempt:
    [jwt] Token verification failed: "exp" claim timestamp check failed
- This strongly suggests the browser's session access token expires
  mid-transfer on a slow/large upload, and the client then aborts the
  request client-side (producing the generic AbortController-style "signal
  is aborted without reason" message) rather than the token being refreshed
  transparently the way it presumably is for normal, fast dashboard
  navigation.
- Confirmed via direct curl testing that a 222MB multipart POST to the same
  endpoint transfers completely over the wire in a reasonable time
  ("upload completely sent off: 227672306 bytes") when not going through
  the browser session flow — ruling out network/proxy size limits as the
  cause (client_max_body_size was also independently raised and confirmed
  not to be the blocking factor).

### Component

Web Dashboard

### Version

Web/API 0.102.0, self-hosted via Docker Compose

### Relevant Logs

```shell
[...]

PR #3078, opened 2026-08-03
fix(devx): isolate concurrent worktree sessions from the shared test DB (#3066)

### LINKED ISSUE #3066: [DevX] Shared fixed-port dev/test Postgres+Redis containers get clobbered by concurrent worktree sessions
## Description
Multiple concurrent worktree/agent sessions on one dev machine share the same fixed-port Postgres/Redis containers (dev :5432, test :5433). During #2764 verification, a full unattended integration run could not complete cleanly: `pg_stat_activity` showed cross-session locks and `tuple concurrently updated` errors on GRANT statements originating from sibling sessions, not from the code under test. Every suite passed when run in isolation.

## Root Cause (if known)
The integration setup does per-test `TRUNCATE ... CASCADE` and role GRANTs against a single shared database instance; two sessions doing this concurrently interleave destructively. Nothing scopes a worktree's test run to its own instance.

## Proposed Fix
Either per-worktree ephemeral containers (random host ports, e.g. via the worktree-stack tooling pattern) for the integration/RLS suites, or a cross-session run mutex (flock on a well-known path) as a cheap interim.

## Affected Files
- `apps/api/src/__tests__/integration/setup.ts`
- local dev tooling / compose overrides

ToddHebebrand:
Closes #3066

## Problem

Concurrent worktree/agent sessions share the fixed-port test Postgres/Redis (:5433/:6380). Two sessions running the integration suite interleave per-test `TRUNCATE ... CASCADE` and role GRANTs destructively (`tuple concurrently updated`, cross-session locks), and one branch's migrations pollute the shared `breeze_migrations` ledger for everyone else. That second failure mode produced a false bug report the same day this shipped: #3064 claimed main's cascade suite was red, but the failing table came from a sibling worktree's in-flight branch (#2951) — and a "clean worktree repro against origin/main" still reproduced it because the clean worktree dialed the same polluted database. Environment pollution was byte-identical to a repo defect.

## Design

Both proposed layers, because they solve different halves and neither obsoletes the other:

**1. `pnpm test-stack` — per-worktree ephemeral test stack (the real fix).** Extends the `wt-stack` philosophy to the test databases: a compose project named from the branch/worktree (`breeze-test-<slug>`), **ephemeral host ports** (publish `:0`, read back via `docker compose port`), per-project container/network names, fsync-off tmpfs Postgres. It writes a **worktree-local `.env.test`** — which the whole existing env chain (`vitest.integration.config.ts` + `loadEnv.ts`) already reads from the monorepo root, and in a worktree the monorepo root *is* the worktree — so `pnpm vitest run -c vitest.integration.config.ts` lands on the isolated instance with zero changes to how tests are invoked. `down` removes the containers, volumes, and the generated `.env.test` (a hand-written `.env.test` is detected by marker and never touched). Chosen over testcontainers-per-run: no new dependency, reuses the repo's existing compose file and wt-stack conventions, and a per-worktree stack amortizes the one-time full migration (~5 min) across many runs where per-run containers would pay it every time.

**2. globalSetup hardening (protects everyone who still shares, including unaware sessions).**
- **Cross-session run mutex** — a Postgres **advisory lock taken on the target database itself**, held for the entire vitest run (released by the globalSetup teardown). Chosen over a filesystem flock deliberately: it scopes to the actual DB instance (two ephemeral stacks never contend), auto-releases when a crashed/SIGKILLed run's session dies (no stale lockfile), and needs no well-known host path. Default behavior is wait-loudly with a pointer to `pnpm test-stack up`; `BREEZE_INTEGRATION_LOCK_NOWAIT=1` fails fast instead.
- **Ledger-drift guard** — before `autoMigrate`, refuse to run when the DB's `breeze_migrations` ledger contains core migrations this checkout doesn't have, with a message naming the offending files and the three remediations (own stack / reset shared / `BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=1`). This alone would have prevented the entire #3064 investigation. Extension-namespaced ledger rows are exempt (absent extensions are legitimate, mirroring `partitionLedgerRows`).

## Blast radius / CI

- `docker-compose.test.yml` changes are pure env interpolation with the historical values as defaults — `docker compose -f docker-compose.test.yml up -d` (CI's exact invocation, incl. the `docker exec breeze-postgres-test` wait line) renders byte-identically (verified via `docker compose config`). The only behavioral addition is `fsync=off`/`synchronous_commit=off` on the already-tmpfs, already-throwaway test Postgres.
- CI is otherwise untouched: each runner has an isolated fresh DB, so the try-lock always succeeds instantly and the ledger (migrated by the same checkout) can never drift.
- Anyone running the suite today against a **clean** shared DB sees no change. Against a **polluted** shared DB they now get an immediate, self-explaining refusal instead of misleading red suites — that's the feature. Note for the operator: the current shared :5433 DB carries #2951's migration, so a main-based run there will trip the guard until #2951 lands or the row is cleaned; the guard message covers both outs. (I did not touch the shared DB — verified everything on my own ephemeral stack.)
- The RLS-coverage / request-db-role runners have their own configs and don't take the lock; they're read-only or self-scoped. Extending the lock there is a possible follow-up.

## Evidence (all against my own ephemeral stack, shared :5433 untouched)

- `pnpm test-stack up` → private containers on ephemeral ports (32791/32792), generated `.env.test`; `staleBackupReaper.integration.test.ts` green through the full loadEnv → lock → guard → autoMigrate chain on the fresh DB; `fsync`/`synchronous_commit` confirmed `off`.
- Simulated sibling-branch pollution (`INSERT` a fake ledger row) → run refused with the exact remediation message; `BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=1` warns and proceeds (test green).
- Sibling `psql` session holding the advisory lock → `BREEZE_INTEGRATION_LOCK_NOWAIT=1` run refused with the serialize-or-isolate message.
- `pnpm test-stack down` → containers/volumes/network removed, generated `.env.test` deleted, `info` reports fallback to shared defaults.
- Unit: `migrationLedgerGuard.test.ts` (6), `test-stack` helper tests (6), `autoMigrate.test.ts` + `composeBindMounts.test.ts` (57) all green; `tsc --noEmit` clean on apps/api.


ToddHebebrand:

**Findings:** 5 raised (2 Important each from code-reviewer/test-analyzer, 1 Important from silent-failure-hunter, overlapping) → all addressed in a4fc5267c; 0 outstanding.
- Advisory-lock connection was subject to postgres.js's randomized 30–60 min `max_lifetime` recycle — the mutex would silently expire mid-run. Pinned to `null`; `pg_advisory_unlock() === false` at teardown now warns loudly.
- `test-stack down` after a branch switch/rename silently orphaned the stack (re-derived project name, compose down no-op, breadcrumb deleted). `down` now tears down the project recorded in the generated `.env.test`; `up` warns on mismatch; `ls --all` sees stopped stacks. Live-verified: `up` → switch branch → `down` removed the recorded project's containers/network with nothing left behind.
- `scripts/dev` tests executed nowhere (outside the pnpm workspace — same blind spot as the existing wt-stack suites). Added root `test:dev-scripts` + a step in the Lint CI job; runs all 23 scripts/dev tests (wt-stack included) in <1s.
- Env-flag truthiness: `BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=0` would have ENABLED the bypass. Strict `isEnvFlagEnabled` ('1'/'true' only) + unit tests.
- `parsePublishedPort` untested / `autoMigrate` readdir catch over-broad → moved+covered (IPv4/IPv6/CRLF/empty contract) and narrowed to ENOENT respectively.

Not actioned (noted deliberately): bounded-wait/periodic re-log while blocked on the run lock (NOWAIT env is the escape hatch; can follow up), and extending the lock to the separate rls-coverage/request-db-role runners (read-only or self-scoped).

**Tests:** `pnpm test:dev-scripts` 23 passed; apps/api unit (migrationLedgerGuard, envFlag, autoMigrate, composeBindMounts) 71 passed; live end-to-end re-verified after fixes — `test-stack up` → integration test green through lock → ledger guard → autoMigrate on the ephemeral stack (exit 0), plus guard-refusal, bypass-warn, and NOWAIT-refusal scenarios exercised earlier. `tsc --noEmit` clean on apps/api. `docker compose -f docker-compose.test.yml config` with no env renders byte-identical defaults for CI.

PR #3079, opened 2026-08-03
fix(api): deterministic local unit test runs — proxy-header env stub, source-scan timeouts

ToddHebebrand:
## Summary

Two small local-dev test reliability fixes found during #2764's CI stabilization work today. Both are local-only flakiness — CI was never affected — but they cost time re-diagnosing "why did this fail on my machine but not CI."

### 1. `apps/api/src/routes/auth/helpers.test.ts` — ambient `.env` leaks into unit tests

**Symptom:** the three `isRequestConnectionSecure` tests intermittently fail locally, never in CI.

**Root cause:** the gitignored repo-root `.env` (used for local dev) sets `TRUST_PROXY_HEADERS=false`. dotenv injects that into the process env for the whole unit test run, which silently overrides the "auto" trust-mode default these tests assert on. CI has no `.env` file, so it never observes this.

**Fix:** `shouldTrustProxyHeaders()` (`apps/api/src/services/clientIp.ts`) reads `process.env.TRUST_PROXY_HEADERS` at call time, not import time, so no module-reset gymnastics were needed — just pin the var explicitly with `vi.stubEnv('TRUST_PROXY_HEADERS', 'true')` in a `beforeEach` and `vi.unstubAllEnvs()` in `afterEach`, matching the `vi.stubEnv` pattern already used elsewhere in the API test suite. The suite is now deterministic regardless of the machine's ambient environment.

### 2. Four source-scanning contract suites flake under machine load (5s `testTimeout`)

**Symptom:** `revocationCallsites`, `noRawEnvNumberCoercion`, `mcpServer.creatorPermsNull`, and `secretBearingTools.contract` intermittently exceed vitest's default 5s `testTimeout` — a different one each run — when the machine is under load or the FS cache is cold. All pass reliably when the machine is idle.

**Root cause:** these are contract tests that scan/grep across the repo's source tree (or do `vi.resetModules()` + dynamic re-import), so they're IO-bound rather than CPU-bound, and 5s isn't a reliable bound under load.

**Fix:** widen just these specific tests to a 30s timeout via vitest's per-test timeout option, with a one-line comment on each explaining why. Left the global vitest `testTimeout` untouched so ordinary tests still fail fast on a real hang.

## Test plan
- [x] `pnpm vitest run` on the 5 affected files — 93/93 tests pass with `TRUST_PROXY_HEADERS=false` exported in the shell
- [x] Same run passes with the var unset
- [x] `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` clean

Both issues were found during #2764's CI stabilization work.

### DIFF
--- a/apps/api/src/oauth/revocationCallsites.test.ts
+++ b/apps/api/src/oauth/revocationCallsites.test.ts
@@ -37,5 +37,6 @@ describe('OAuth revocation marker call-site contract', () => {
     });
 
     expect(violations).toEqual([]);
-  });
+    // Repo-wide source scan is IO-bound; widen timeout so it doesn't flake under load or cold FS cache.
+  }, 30000);
 });
--- a/apps/api/src/routes/auth/helpers.test.ts
+++ b/apps/api/src/routes/auth/helpers.test.ts
@@ -355,6 +355,19 @@ function disableProxyTrustEnv(): void {
 }
 
 describe('isRequestConnectionSecure (#1618 — Secure flag tracks real transport)', () => {
+  // These tests assert on the "auto" trust-mode default (non-production ==
+  // trusted), which a gitignored repo-root .env setting TRUST_PROXY_HEADERS=false
+  // for local dev silently overrides via dotenv injection at test-run time. CI
+  // has no .env file so it never sees this; pin the var explicitly so the
+  // suite is deterministic regardless of the machine's ambient environment.
+  beforeEach(() => {
+    vi.stubEnv('TRUST_PROXY_HEADERS', 'true');
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
   it('true when X-Forwarded-Proto is https (Caddy behind TLS)', () => {
     expect(isRequestConnectionSecure(makeCookieContext({ forwardedProto: 'https' }).c)).toBe(true);
   });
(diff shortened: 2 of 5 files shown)


ToddHebebrand:
**Review summary** (one round, pr-review-toolkit code-reviewer)

Reviewed commit `3c602ae80` — confirmed correct:
- `vi.stubEnv`/`vi.unstubAllEnvs` is the established sibling pattern (~10 other users in `apps/api/src`); `TRUST_PROXY_HEADERS` is read at call time (`clientIp.ts`'s `shouldTrustProxyHeaders()`), so no `resetModules`/dynamic-import needed.
- `it(name, fn, timeoutMs)` is the correct vitest 4.1.10 signature (not the object-options form) for the four widened contract suites; timeout placement is complete — the two per-file scanning tests in each file got it, cheap collection-time/allowlist-only tests correctly didn't.
- Not vacuous: the stubbed "auto"-equivalent state still exercises all the forwarded-header assertions.

**Important finding, fixed in `634cded56`:** the reviewer caught that `TRUST_PROXY_HEADERS` wasn't the only ambient var in play — `isTrustedProxySource()` also reads `TRUSTED_PROXY_CIDRS`, and `.env.example` ships a non-empty default for it alongside `TRUST_PROXY_HEADERS=true`, so a real local `.env` could independently break the same tests via that var alone. Added `vi.stubEnv('TRUSTED_PROXY_CIDRS', '')` alongside and reworded the comment. Re-verified: 53/53 pass with both `TRUST_PROXY_HEADERS=false` and `TRUSTED_PROXY_CIDRS=172.31.0.10/32` (the `.env.example` values) exported, and with both unset.

No further findings. Not merging — leaving for manual merge.

PR #3081, opened 2026-08-03
fix(security): scrub transaction events too, so a sampled trace cannot ship a live API key (#3077)

### LINKED ISSUE #3077: Sentry scrubber does not redact X-API-Key, so a 500 can ship a live brz_ credential
Found while reviewing #2826. Not caused by that PR — pre-existing, and it affects every API-key-authenticated route, so filing separately.

**What's wrong:** `apps/api/src/services/sentry.ts:24-30` scrubs exactly two request headers:

```ts
const headers = event?.request?.headers;
if (headers) {
  for (const k of Object.keys(headers)) {
    if (k.toLowerCase() === 'authorization' || k.toLowerCase() === 'cookie') headers[k] = '[redacted]';
  }
}
```

`x-api-key` is not in that list. Every API-key-authenticated surface passes its credential in that header — organization API keys (`brz_...`) and partner service principals (`brz_sp_...`). `@sentry/node`'s HTTP integration attaches `event.request.headers`, and `app.onError` (`apps/api/src/index.ts:1003-1020`) calls `captureException` on any unhandled throw. So a single 500 on an API-key route can put a live, unexpired credential into Sentry, where it is retained and visible to anyone with project access.

**The intent was already there.** The `extra` scrubber immediately below (`sentry.ts:31-40`) *does* know about this credential shape:

```ts
if (SENSITIVE_KEYS.has(k.toLowerCase()) || (typeof v === 'string' && v.startsWith('brz_'))) {
  extra[k] = '[redacted]';
}
```

The `brz_` prefix check was written for exactly this class of secret and was never extended to headers. This reads as an oversight in #1379 B3 rather than a deliberate scope decision.

**Suggested fix:** in `scrubEvent`, redact by value as well as by name — any header whose value matches `/^brz_/`, plus `x-api-key` and `x-breeze-csrf` by name. Redacting on the value catches future header names for free. `scrubEvent` is already exported for test, and `sentry.test.ts` has the harness, so this is a small change with a direct regression test.

**Worth checking during the fix:** whether any credential has already landed in the retained Sentry events. If so, those keys should be rotated rather than just scrubbed going forward.

**Severity:** not remotely exploitable on its own — it needs an unhandled 500 on an API-key route — but the payload is a working credential and the blast radius is whoever can read the Sentry project. I'd treat it as a should-fix rather than a drop-everything.

Staying open — no code change yet.

bdunncompany:
Closes #3077 — though not where the issue expected, so the reasoning is worth a read before merging. Full trace is in [my comment on the issue](https://github.com/LanternOps/breeze/issues/3077#issuecomment-5170387635).

## The filed case no longer reproduces

The scrubber quoted in #3077 (`sentry.ts:24-30`, the `authorization`/`cookie` header loop, and the `extra` scrubber with the `brz_` prefix check) was removed by #2843 (`a50769487`) on 2026-07-27. `scrubEvent` now rebuilds the surface from an allowlist rather than denying known-bad names — `request` becomes `{ method }` and `extra` is deleted outright. Adding `x-api-key` to a header allowlist would have been a no-op against code that no longer exists.

## But the credential does leave, on the transaction path

`Sentry.init` set `beforeSend` only, and `@sentry/core` dispatches the two hooks by event type (`client.js:735-793`):

```js
if (isErrorEvent(processedEvent) && beforeSend) return beforeSend(processedEvent, hint);
if (isTransactionEvent(processedEvent)) { ... if (beforeSendTransaction) return beforeSendTransaction(processedEvent, hint); }
```

There was no `beforeSendTransaction` in the repo, so `scrubEvent` ran on exactly half the outbound traffic.

Transaction events carry the headers all the same. `requestDataIntegration()` is a default node integration (`@sentry/node-core` `getDefaultIntegrations`), its `processEvent` fires on every event type, and `extractNormalizedRequestData` copies the header bag verbatim:

```js
const headers = { ...normalizedRequest.headers };
if (include.headers) { requestData.headers = headers; ... }
```

`include.headers` resolves to `dataCollection.httpHeaders.request !== false`; with `sendDefaultPii` unset that value is an object (`{ deny: PII_HEADER_SNIPPETS }`), so it is truthy. The SDK's `SENSITIVE_KEY_SNIPPETS` deny list — which does contain `key` and would have caught `x-api-key` — is applied only on the span-attribute path via `httpHeadersToSpanAttributes`, never to the event body.

Net: every sampled request on an API-key route shipped a live `brz_` / `brz_sp_` credential, plus the full url, query string and request body. **No error required** — that is worse than the case as filed, which needed an unhandled 500. And `.env.example:614` ships `SENTRY_TRACES_SAMPLE_RATE=0.1`, so sampling is on for anyone who copied it (compose defaults to `0`, so this bites `.env.example` users specifically).

## The change

`scrubTransactionEvent` applies the same allowlist rebuild as `scrubEvent`: `request` reduced to `{ method }`, `extra` and `breadcrumbs` dropped, tags picked from `ALLOWED_TAG_NAMES`, `user` narrowed to a bounded id.

It deliberately is **not** `scrubEvent` itself. That one does `delete mutableEvent.contexts`, and a transaction event without `contexts.trace` is invalid — reusing it would have silently disabled tracing rather than secured it. So `contexts` is narrowed to `trace` alone, which also drops the `nodeContextIntegration` server metadata.

`event.spans[].data` is left alone: those attributes already pass through the SDK's `filterKeyValueData` + `SENSITIVE_KEY_SNIPPETS`, so header values arrive pre-redacted there.

Four tests alongside the existing `scrubEvent` harness: the api-key strip, the `contexts.trace` survival guard (the one that catches a future "just reuse scrubEvent" refactor), sparse-event safety, and a wiring assertion that `beforeSendTransaction` is actually passed to `Sentry.init` — the gap here was a missing hook, not a bad hook, so the wiring is the thing worth pinning.

## Operator note

This closes the path going forward only. Any deployment that has run with a nonzero `SENTRY_TRACES_SAMPLE_RATE` should treat retained **transaction** events, not just error events, as a place a live key may already be sitting — those keys want rotating. That is your "worth checking during the fix" point, and the transaction path makes it a larger set than the issue assumed.

## Local gate

Node 22.23.2, all green:

- `tsc --noEmit` — clean, exit 0
- `vitest run` in `apps/api` — **1240 files / 19504 tests passed, 0 failed** (5 files / 61 tests skipped)
- `eslint` on both changed files — clean

No dependency, lockfile, Go, or container surface touched, so the Trivy / govulncheck / audit inputs are unchanged from `main`.

### DIFF
--- a/apps/api/src/services/sentry.test.ts
+++ b/apps/api/src/services/sentry.test.ts
@@ -482,6 +482,100 @@ describe('scrubEvent', () => {
   });
 });
 
+describe('scrubTransactionEvent (#3077)', () => {
+  beforeEach(() => {
+    vi.resetModules();
+    initMock.mockClear();
+    process.env = { ...ORIGINAL_ENV };
+  });
+
+  afterEach(() => {
+    process.env = { ...ORIGINAL_ENV };
+  });
+
+  // A transaction event as `requestDataIntegration()` builds it: the SDK copies
+  // the request header bag verbatim onto EVERY event type, and only applies its
+  // own SENSITIVE_KEY_SNIPPETS deny list on the span-attribute path — so the
+  // event body is where a live `brz_` credential rides out.
+  const transactionEvent = () => ({
+    type: 'transaction',
+    release: '1.2.3',
+    transaction: 'GET /api/devices/:id',
+    request: {
+      method: 'GET',
+      url: 'https://example.test/api/devices/1?token=raw-capability',
+      query_string: 'token=raw-capability',
+      headers: {
+        'X-API-Key': 'brz_raw-capability',
+        Authorization: 'Bearer raw-capability',
+        COOKIE: 'session=raw-capability',
+        'user-agent': 'sdk',
+      },
+      data: { token: 'raw-capability' },
+      cookies: { session: 'raw-capability' },
+    },
+    contexts: {
+      trace: { trace_id: 'abc123', span_id: 'def456', op: 'http.server' },
+      runtime: { name: 'node', version: 'raw-capability' },
+      os: { name: 'raw-capability' },
+    },
+    breadcrumbs: [{ message: 'raw-capability' }],
+    extra: { apiKey: 'brz_raw-capability' },
+    tags: { method: 'GET', org_id: 'o-1', arbitrary: 'raw-capability' },
+    spans: [{ span_id: 'def456', op: 'db.query' }],
+  });
+
+  it('strips the api key (and every other header) from a sampled transaction', async () => {
+    const { scrubTransactionEvent } = await import('./sentry');
+    const out = scrubTransactionEvent(transactionEvent() as any);
+
+    expect(out.request).toEqual({ method: 'GET' });
+    expect(out.extra).toBeUndefined();
+    expect(out.breadcrumbs).toBeUndefined();
+    expect(out.tags).toEqual({ method: 'GET', org_id: 'o-1' });
+    expect(JSON.stringify(out)).not.toContain('raw-capability');
+    expect(JSON.stringify(out)).not.toContain('brz_');
+  });
+
+  it('keeps contexts.trace so the transaction stays a valid event', async () => {
+    // Reusing scrubEvent here would delete `contexts` outright and silently
+    // disable tracing rather than secure it — a transaction event without
+    // contexts.trace is rejected.
+    const { scrubTransactionEvent } = await import('./sentry');
+    const out = scrubTransactionEvent(transactionEvent() as any);
+
+    expect(out.contexts).toEqual({
+      trace: { trace_id: 'abc123', span_id: 'def456', op: 'http.server' },
+    });
+    expect(out.transaction).toBe('GET /api/devices/:id');
+    expect(out.spans).toEqual([{ span_id: 'def456', op: 'db.query' }]);
+  });
+
+  it('does not throw on sparse or trace-less events', async () => {
+    const { scrubTransactionEvent } = await import('./sentry');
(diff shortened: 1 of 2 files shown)


ToddHebebrand (approved):
Reviewed at 15ad6f31d. This is exactly the right shape, and the diagnosis on #3077 was verified end-to-end against @sentry/core 10.51.0 source — `beforeSend` never runs on transaction events, and the default `requestDataIntegration` attaches the header bag verbatim, so the missing `beforeSendTransaction` hook was the whole gap.

**What I checked:** `scrubTransactionEvent` reduces `request` to `{ method }`, drops `extra`/`breadcrumbs`, and narrows `contexts` to `trace` alone rather than deleting it — which keeps the event valid instead of silently disabling tracing, the trap you called out on the issue. The span-attribute path is safe to leave untouched: `handleHttpHeader` in @sentry/core redacts `x-api-key` (matches the `key` snippet) before this hook ever runs. `scrubEvent` on the error path is untouched. Tests assert both directions — no `brz_` substring anywhere in serialized output AND `contexts.trace`/spans survive — plus the wiring test that `beforeSendTransaction` is actually passed to `Sentry.init`, which is the regression that matters.

**One follow-up worth a look, not blocking:** `scrubTransactionEvent` keeps `event.transaction` (the route name). With no native Hono route templating in the Node SDK it's worth confirming transaction naming can't fall back to a raw unparameterized path carrying resource IDs. Separate concern, separate PR if real.

Merging. Anyone who has run with a nonzero SENTRY_TRACES_SAMPLE_RATE before this fix should treat retained transaction events as potentially carrying keys — the rotation note from the issue thread stands.

PR #3086, opened 2026-08-04
feat(web): session-expiry polish — atomic logout, expiry notices, proactive keepalive, idle warning

ToddHebebrand:
## What

Polishes how the web app handles session death. Previously an expired session gutted the sidebar in place (44/56 nav items are permission-gated) while the redirect to `/login` — if it happened at all — raced asynchronously in a different island, with no explanation to the user.

Four changes, all in `apps/web`:

1. **Atomic, explicit session expiry.** New `handleSessionExpired(reason)` in the auth store: sets a non-persisted `sessionExpiredReason`, calls `logout()`, and immediately `window.location.replace`s to `/login?next=…&reason=…` (idempotent across concurrent 401s; no-op if already on `/login`). Both `fetchWithAuth` give-up paths now route through it — including the 401-refresh-failed path that previously just mutated state and returned the stale response. `AuthOverlay` renders a full-screen "session expired" mask the instant the reason is set, so the gutted UI is never visible.
2. **Deep links survive re-login.** The old path emitted `?returnTo=`, which `login.astro` never read. Redirects now use `loginPathWithNext()` (`?next=`, validated by `getSafeNext`), and the login page shows an informational notice explaining the sign-out (`reason=session-expired` / `reason=idle`; `?error=` SSO copy takes precedence when both resolve).
3. **Keepalive detects a dead session proactively.** The 5-minute heartbeat refresh previously discarded its result, so a revoked/expired refresh cookie went unnoticed until the next user click 401'd. It now distinguishes `restored` / `auth-failed` / `transient`: hard auth failure logs out cleanly right away; transient network/5xx failures change nothing (an offline user is never evicted). Refresh internals (retry counts, backoff, Web Locks, rotation/reuse-detection handling) are untouched — the outcome the code already computed is just propagated.
4. **Idle-timeout warning.** Instead of a silent idle logout, a countdown dialog appears `min(2 min, budget/2)` before the org/partner-configurable idle deadline. Passive activity (mousemove/scroll/focus/visibilitychange) neither extends nor dismisses it; deliberate input (mousedown/keydown/touchstart) or the "Stay signed in" button dismisses it and refreshes the token. At 0:00 it performs the durable `apiLogout()` (server-side family revocation) and routes through `handleSessionExpired('idle')`, landing on the login page with the inactivity notice.

## Notes for review

- **Behavior change:** the idle warning now arms while the tab is hidden (deliberate — a returning user sees the countdown rather than being silently evicted; expiry in a hidden tab is owned by the 1s tick, with the 30s heartbeat as a fail-safe for budget-shrinking scope switches; a shared in-flight ref makes the two paths mutually exclusive).
- `?reason=` on `/login` carries exactly `session-expired` | `idle` from these paths; the pre-existing `reason=registration-disabled` on the same param is consumed by a disjoint code path and unaffected.
- New i18n keys added to all seven locales (key-parity suite green).
- New testids: `session-expired-overlay`, `login-session-expired-notice`, `idle-warning-stay`, `idle-warning-body`.

## Testing

- Targeted suites green: `auth.test.ts`, `AdminSessionManager.test.tsx`, `AuthOverlay.test.tsx`, `LoginPage.test.tsx`, `authNext`/`authScope` tests (114 tests in final run), plus i18n key-parity, literal-key, and no-silent-mutations contract suites.
- `astro check` / eslint clean on touched files.
- Each task went through an independent spec+quality review; a whole-branch review (cross-task lifecycle, heartbeat/modal interplay, open-redirect and refresh-security checks) found nothing above nice-to-have, and its three minors (helper dedup, ref-reset hardening, dialog `autoFocus`+`aria-describedby`) are fixed in the final commit.


ToddHebebrand:
Post-open review round (4 parallel reviewers: code quality, test coverage, comment accuracy, silent failures) surfaced two real defects the earlier per-task and whole-branch reviews missed; both fixed in 59f881b5a:

- `apiLogout()` had an unbounded fetch, and the idle-logout path awaited it before `handleSessionExpired('idle')` — a hung `POST /auth/logout` would strand the "Signing you out…" modal and permanently gate the keepalive. Now bounded by an 8s abort (same pattern as `refreshFetchOnce`), and client-side eviction proceeds regardless.
- Both of AuthOverlay's redirect paths (main effect + 10s safety net) could soft-navigate to bare `/login` racing `handleSessionExpired`'s hard redirect, dropping `?next=`/`?reason=`. Both are now gated on `sessionExpiredReason` being unset; covered in both directions (fires for a plain unauthenticated visitor, dormant during expiry), mutation-verified.

Also in that commit: `console.warn` in the two previously-silent auth catches, comment corrections (including documenting the deliberate transient-eviction policy split between foreground fetches and the background heartbeat), and new tests for the `idleTimeoutMs/2` warning-lead clamp and hidden-tab/visibilitychange behavior.

Issue #3089, opened 2026-08-04
[AI] Chat turn blocks on slowest pending approval; assistant cannot respond while a tool awaits approval


ToddHebebrand:
## Description
When an AI chat turn issues parallel tool calls and one requires approval, **the entire turn gates on the slowest pending approval**:

- A `get_effective_configuration` call that completed instantly was withheld for 5 minutes because a sibling `execute_command` sat pending approval (US prod, 2026-08-03, session flagged "Tool failed: execute_command — Approval still pending").
- The user typed "i think it got stuck" while two approvals were pending — and the assistant could not reply for ~3 minutes, because the model was blocked waiting on tool results. From the user's perspective the AI was frozen.
- A second session ended with two raw approval-pending error rows and **no closing assistant message at all** — the model never got a chance to summarize or hand off its best lead.

## Proposed Fix
- Stream completed tool results back without waiting for pending siblings.
- Surface an interim "waiting for your approval on N commands" event into the chat UI immediately when an approval is created (with a link to the approval queue), rather than relying on the model to explain after the fact.
- Let a new user message interrupt/settle a turn that is only waiting on approvals, so the assistant can respond.

## Affected Files
- `apps/api/src/services/aiAgentSdk.ts` (tool execution / approval wait path)
- Chat streaming pipeline, `apps/web` AI chat sidebar

## Reported By
Internal review of flagged US-prod sessions, 2026-08-03 (tenant identifiers redacted).

ToddHebebrand:
Fix in progress: #3104

Issue #3095, opened 2026-08-04
[AI] ai_sessions token counters record 0 for real sessions — cost tracking not capturing usage


ToddHebebrand:
## Description
All three US-prod AI sessions from 2026-08-03 — including a healthy one that ran ~25 tool calls and produced full assistant responses — show `total_input_tokens = 0` and `total_output_tokens = 0` in `ai_sessions`. Per-message `input_tokens`/`output_tokens` presumably aren't rolling up either. This breaks cost tracking/attribution for exactly the sessions we most want to audit (two of the three were flagged for tool failures).

Not limited to error paths: the healthy session also recorded zeros, so the accumulator looks broken generally on this path (streaming SDK sessions?), not just when a turn dies on a tool error.

## Proposed Fix
Investigate `aiCostTracker` accumulation for the SDK streaming path — verify usage deltas from the provider stream are captured and flushed to `ai_sessions` totals, including on turns that end in tool error/abandonment.

## Affected Files
- `apps/api/src/services/aiCostTracker.ts`
- `apps/api/src/services/aiAgentSdk.ts` (usage capture on stream events)

## Reported By
Internal review of flagged US-prod sessions, 2026-08-03 (tenant identifiers redacted).

ToddHebebrand:
Fix in progress: #3103

Issue #3093, opened 2026-08-04
[AI] execute_command output truncated at ~2KB (stringsTruncated) breaks file listings and process lists


ToddHebebrand:
## Description
`execute_command` results for `list_processes` and `file_list` came back truncated at roughly 2KB with `"stringsTruncated": 1` (US prod, 2026-08-03). Two consecutive `file_list` calls on Program Files directories were cut off mid-listing; the model had to re-issue narrower calls — each of which required a fresh high-tier approval (compounding the approval-fatigue problem) — and the final drill-down call died pending approval, ending the investigation.

The model gets no guidance that output was truncated or how to page: no total count, no offset/continuation parameter, just a cut string with a flag it has to notice.

## Proposed Fix
- Raise the cap for structured read commands, or better: support pagination (offset/continuation) on `file_list` / `list_processes`.
- When truncating, append an explicit marker in the payload the model can act on ("output truncated at N of M entries; refine path or use offset").

## Affected Files
- Tool output shaping for `execute_command` (`apps/api/src/services/aiToolOutput.ts` or agent-side truncation — needs investigation to locate the 2KB cap)

## Reported By
Internal review of flagged US-prod sessions, 2026-08-03 (tenant identifiers redacted).

Issue #3088, opened 2026-08-04
[AI] Approval fatigue: every read-only execute_command is high-tier, one WebAuthn tap per call, no batching


ToddHebebrand:
## Description
In a 35-minute US-prod AI troubleshooting stretch (2026-08-03, three back-to-back sessions on one device), the approval ledger shows **25 approval_requests: 19 approved — each requiring a separate `webauthn_platform` tap — and 6 expired** at the 5-minute TTL because the user missed the prompt. Every missed approval froze its session (see #3089, #3090).

All of these were **read-only** `execute_command` types: `event_logs_query`, `list_processes`, `list_services`, `file_list`, `event_logs_list` — all classified `risk_tier: high`. The user tapped WebAuthn 19 times in half an hour to let the AI read event logs and list directories. Approval fatigue is the direct cause of the missed approvals: the two `file_list` prompts that arrived seconds apart at the end of a long approved run are exactly the ones that expired.

## Proposed Fix
Two complementary options:
1. **Tier read-only commandTypes lower.** `event_logs_query` / `event_logs_list` / `list_processes` / `list_services` / `file_list` are non-mutating reads; they don't warrant the same tier as script execution or file writes.
2. **Session-scoped batch approval.** "Approve read-only commands on this device for this session/15 minutes" — one tap covers the burst that an investigation naturally produces.

## Affected Files
- Risk-tier classification for `execute_command` commandTypes (approval request creation path)
- `approval_requests` flow / approver UI

## Reported By
Internal review of flagged US-prod sessions, 2026-08-03 (tenant identifiers redacted).

ToddHebebrand:
PR up: #3105

PR #3096, opened 2026-08-04
fix(ai): give the vision/desktop tools at least the agent round-trip budget (#3091)

### LINKED ISSUE #3091: [AI] Vision tool timeout (30s) too short for WebRTC capture + analysis; analyze_screen times out on prod
## Description
`analyze_screen` failed on US prod (2026-08-03) with `Tool execution timed out after 30000ms: analyze_screen` — the session was flagged for it. The approval was granted 9s after the call; the timeout fired ~31s after approval, i.e. the actual WebRTC setup + screen capture + vision pass exceeded 30s. On the same device in the same window, ordinary `execute_command` round-trips took 40s–6min, so 30s is far below observed prod latency.

`apps/api/src/services/toolTimeouts.ts:28-30` sets `take_screenshot`, `analyze_screen`, and `computer_control` to `30_000` — the *shortest* overrides in the table, while `execute_command` gets `120_000`. Vision tools do a full WebRTC negotiation plus capture plus a model vision call; they should have at least the `execute_command` budget.

## Proposed Fix
Raise the three vision/desktop overrides in `TOOL_TIMEOUT_OVERRIDES` to 120s (or make them config-tunable). Also confirm the timer starts after approval resolution everywhere (it appeared to here), so approval latency never eats execution budget.

## Affected Files
- `apps/api/src/services/toolTimeouts.ts:28-30`

## Reported By

bdunncompany:
Closes #3091.

## What was wrong

`take_screenshot`, `analyze_screen` and `computer_control` sat at `30_000` in `TOOL_TIMEOUT_OVERRIDES` — and they were **the only three entries in the table that lowered the budget below the 60s default**. Every other override raises it (`execute_command` 120s, `manage_patches` 180s, disk ops 90s). So the three slowest tools in the system had the tightest budget in the file, tighter than an unlisted tool gets for free.

That is what bit prod on 2026-08-03: `Tool execution timed out after 30000ms: analyze_screen`, while ordinary `execute_command` round-trips to the same device in the same window were taking 40s to 6min.

All three do an agent round-trip exactly like `execute_command`, and then more on top — screen capture (reusing an already-active capturer) and, for `analyze_screen`, a model vision pass. (Earlier revisions of this description also cited WebRTC negotiation here — that was wrong. `take_screenshot`/`computer_action` reuse an existing `desktop.ScreenCapturer` and never touch the WebRTC/ICE path; the 5s ICE-gather timeout belongs exclusively to `start_desktop` Remote Desktop streaming sessions, agent/internal/remote/desktop/session.go:22.) They now get `execute_command`'s 120s.

## Update: the original fix was largely inert — closing the gap

A follow-up review found the first version of this PR only raised the number in `toolTimeouts.ts`; it never reached the code path that actually enforces the timeout, so `analyze_screen` would still have died at ~30s in production. Root cause: `aiToolsRemote.ts` called `executeCommand(...)` for all three tools with a **second, hardcoded** `timeoutMs: 30000`, which flows straight into `commandQueue.ts`'s single-shot `waitForCommandResult` — the 120s raised in `toolTimeouts.ts` only ever bounded the *outer* `withToolTimeout` wrapper around the whole tool handler (`aiAgentSdkTools.ts`), not the inner command wait that actually returns first.

Fixed by importing `getToolTimeout` from `toolTimeouts.ts` (the single source of truth already used by the outer wrapper) into `aiToolsRemote.ts` and using it for the `timeoutMs` passed to `executeCommand` at all three call sites (`take_screenshot`, `analyze_screen`, `computer_control`) — no second constant. Added `aiToolsRemote.timeout.test.ts`, which mocks `commandQueue.executeCommand` and asserts the `timeoutMs` it actually receives equals `getToolTimeout(<toolName>)` for each of the three tools, plus an end-to-end invariant test tying that to `execute_command`'s budget. No prior test anywhere asserted this consistency — the two numbers were free to drift, and they had.

## Known gap (documented, not fixed here): turn-timeout interaction

With the vision-tool budget now real at 120s, worst case for one of these tools is a 5-minute approval wait (`waitForApproval`, `aiAgentSdk.ts`) followed by up to 120s of execution — up to ~7 minutes, past the 6-minute `SDK_TURN_TIMEOUT_MS` in `streamingSessionManager.ts`. Approval-wait resolution happens entirely before `withToolTimeout` starts, so the two budgets stack rather than overlap; this is real, not theoretical.

Checked whether the turn-timeout handler already settles a still-pending approval/tool wait cleanly before publishing `error` + `done` to the client. As implemented on `main` today, it does **not**: `startTurnTimeout`'s callback only publishes the terminal events and sets `state = 'idle'` — it does not abort `session.abortController` or otherwise unblock the in-flight wait (only `remove()` does that). `runBackgroundProcessor`'s loop only stops on `state === 'closing' | 'closed'`, so if the approval/tool call eventually settles, its later SDK messages still get published into a turn the client was already told had ended. This is a pre-existing gap (the same overlap already existed for two stacked ordinary approval waits before this PR), not something this PR introduces or worsens in kind — but raising the vision-tool budget does make the single-tool worst case newly reachable.

The proper fix is already in flight as a separate PR: **#3104** (for issue #3089) adds a shared per-cycle `APPROVAL_WAIT_BUDGET_MS` that resets at the same `message_start` point as the turn timeout, bounding cumulative approval wait to fit inside the turn window. Per this PR's scope, the budget structure isn't being touched here — the math and the current (unclean) behavior are documented in a comment at the `SDK_TURN_TIMEOUT_MS` definition in `streamingSessionManager.ts` instead, flagged for prioritization alongside #3104.

## Known gap (not fixed here, agent-side): service/headless mode has its own ~30s ceiling

Independently of everything above, in service/headless mode the Go agent routes `take_screenshot`/`computer_action` through the user-helper IPC channel with a 15s timeout × 2 attempts ≈ 30s ceiling (`agent/internal/heartbeat/handlers_screenshot.go:60`). That's a separate ceiling *beneath* the API-side one fixed here, and it hits exactly the fleet-managed-service deployment mode this timeout was raised for. This PR does not touch Go code. Filing a follow-up issue for the agent-side ceiling; full resolution for service-mode agents needs that agent-side fix too.

## Tests

The obvious test here is "assert the three new numbers", which would pass and teach nothing. The interesting property is the shape the table had drifted into, so the tests assert that instead:

- **no listed tool may get less than the unlisted default** — the invariant that was violated
- **the vision tools get at least the agent round-trip budget** — tied to `execute_command` rather than hardcoded, so they track it if it moves
- **the timeout that reaches `executeCommand` matches `getToolTimeout()`** — new, closes the gap that made the original fix inert

A future override cannot silently reintroduce the same shape, and the timeout can't silently stop reaching the command wait again either.

## Local gate

Node 22.23.2:

- `tsc --noEmit` (apps/api, `NODE_OPTIONS=--max-old-space-size=8192` — default heap OOMs on this checkout) — clean, exit 0
- `eslint` on all changed files — clean
- `vitest run` targeted: `aiToolsRemote.timeout.test.ts`, `aiToolsRemote.test.ts`, `aiToolsRemote.siteScope.test.ts`, `toolTimeouts.test.ts` — **29/29 passed**

No dependency, lockfile, Go, or container surface touched.

## Not addressed here

The other eight issues from that session review are left alone deliberately. #3093 (2KB truncation) and #3095 (zero token counters) need the investigation you flagged rather than a guess; #3087 (device-bound org scoping) is a tenancy-shaped correctness fix that wants design review, not a quick patch; #3088/#3089 are policy and streaming-architecture calls that are yours to make (#3089 is now in flight as #3104). Happy to take any of them if you want to point me at one.

### DIFF
--- a/apps/api/src/services/aiToolsRemote.timeout.test.ts
+++ b/apps/api/src/services/aiToolsRemote.timeout.test.ts
@@ -1,0 +1,119 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// #3096 review finding: the PR raised the vision-tool budget in toolTimeouts.ts,
+// but take_screenshot / analyze_screen / computer_control called executeCommand
+// with a hardcoded `timeoutMs: 30000` — the raised budget never reached
+// commandQueue's single-shot waitForCommandResult, so analyze_screen still died
+// at ~30s regardless of what getToolTimeout() returned. This suite asserts the
+// two never drift apart again: whatever toolTimeouts.ts says for a given tool is
+// exactly what reaches executeCommand's `timeoutMs` option.
+
+vi.mock('../db', () => ({
+  runOutsideDbContext: vi.fn((fn: any) => fn()),
+  withDbAccessContext: vi.fn(async (_ctx: unknown, fn: () => Promise<unknown>) => fn()),
+  withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
+  db: { select: vi.fn() },
+}));
+
+vi.mock('./commandQueue', () => ({
+  executeCommand: vi.fn(async () => ({
+    status: 'completed',
+    stdout: JSON.stringify({ imageBase64: 'AA==' }),
+  })),
+}));
+
+import { db } from '../db';
+import { executeCommand } from './commandQueue';
+import { registerRemoteTools } from './aiToolsRemote';
+import { getToolTimeout } from './toolTimeouts';
+import type { AuthContext } from '../middleware/auth';
+import type { AiTool } from './aiTools';
+
+const DEVICE_ID = '33333333-3333-3333-3333-333333333333';
+
+function createQueryChain(rows: any[]) {
+  const chain: any = {};
+  chain.from = vi.fn(() => chain);
+  chain.where = vi.fn(() => chain);
+  chain.limit = vi.fn(() => Promise.resolve(rows));
+  return chain;
+}
+
+function handlerFor(name: string): AiTool['handler'] {
+  const reg = new Map<string, AiTool>();
+  registerRemoteTools(reg);
+  return reg.get(name)!.handler;
+}
+
+function makeAuth(): AuthContext {
+  return {
+    principal: { kind: 'user_session' },
+    user: { id: 'user-1', email: 'a@b.c', name: 'A', isPlatformAdmin: false },
+    token: {} as any,
+    partnerId: null,
+    orgId: 'org-1',
+    scope: 'organization',
+    accessibleOrgIds: ['org-1'],
+    orgCondition: () => undefined,
+    canAccessOrg: () => true,
+    allowedSiteIds: undefined,
+    canAccessSite: () => true,
+  } as AuthContext;
+}
+
+const mockDb = db as unknown as { select: ReturnType<typeof vi.fn> };
+const mockExecuteCommand = executeCommand as unknown as ReturnType<typeof vi.fn>;
+
+describe('aiToolsRemote — timeoutMs threads through to executeCommand', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockDb.select.mockImplementation(() =>
+      createQueryChain([
+        { id: DEVICE_ID, status: 'online', siteId: null, hostname: 'host-1', orgId: 'org-1' },
+      ])
+    );
+    mockExecuteCommand.mockResolvedValue({
+      status: 'completed',
+      stdout: JSON.stringify({ imageBase64: 'AA==' }),
+    });
+  });
+
+  it('take_screenshot passes getToolTimeout("take_screenshot"), not a hardcoded 30s', async () => {
+    await handlerFor('take_screenshot')({ deviceId: DEVICE_ID }, makeAuth());
+    expect(mockExecuteCommand).toHaveBeenCalledTimes(1);
+    const [, , , options] = mockExecuteCommand.mock.calls[0]!;
+    expect(options.timeoutMs).toBe(getToolTimeout('take_screenshot'));
+  });
+
+  it('analyze_screen passes getToolTimeout("analyze_screen"), not a hardcoded 30s', async () => {
+    await handlerFor('analyze_screen')({ deviceId: DEVICE_ID }, makeAuth());
+    expect(mockExecuteCommand).toHaveBeenCalledTimes(1);
+    const [, , , options] = mockExecuteCommand.mock.calls[0]!;
+    expect(options.timeoutMs).toBe(getToolTimeout('analyze_screen'));
+  });
+
+  it('computer_control passes getToolTimeout("computer_control"), not a hardcoded 30s', async () => {
+    await handlerFor('computer_control')({ deviceId: DEVICE_ID, action: 'screenshot' }, makeAuth());
+    expect(mockExecuteCommand).toHaveBeenCalledTimes(1);
+    const [, , , options] = mockExecuteCommand.mock.calls[0]!;
+    expect(options.timeoutMs).toBe(getToolTimeout('computer_control'));
+  });
+
+  it('all three vision/desktop tools get at least the agent round-trip budget end to end', async () => {
+    // Guards the whole chain, not just the constant table: the value that
+    // ACTUALLY reaches executeCommand must be >= execute_command's budget,
+    // matching the invariant asserted in toolTimeouts.test.ts against the
+    // source-of-truth table alone.
+    const agentRoundTrip = getToolTimeout('execute_command');
+    for (const [tool, input] of [
+      ['take_screenshot', { deviceId: DEVICE_ID }],
+      ['analyze_screen', { deviceId: DEVICE_ID }],
+      ['computer_control', { deviceId: DEVICE_ID, action: 'screenshot' }],
+    ] as const) {
+      mockExecuteCommand.mockClear();
+      await handlerFor(tool)(input, makeAuth());
+      const [, , , options] = mockExecuteCommand.mock.calls[0]!;
+      expect(options.timeoutMs, `${tool} timeoutMs`).toBeGreaterThanOrEqual(agentRoundTrip);
+    }
+  });
+});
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
## Review-response update (60255f1d1)

A follow-up review (two independent traces) found the original fix in this PR was largely **inert** — the raised `toolTimeouts.ts` budget never reached the code path that enforces it. Addressed all four findings:

**1. CRITICAL — fixed.** `aiToolsRemote.ts` called `executeCommand(...)` for `take_screenshot`/`analyze_screen`/`computer_control` with a hardcoded `timeoutMs: 30000` (lines 83, 141, 227), which flows straight into `commandQueue.ts`'s single-shot `waitForCommandResult`. That inner 30s timeout always fired before the outer 120s `withToolTimeout` wrapper in `aiAgentSdkTools.ts` ever mattered — so `analyze_screen` would still have died at ~30s in production. Fixed by importing `getToolTimeout` from `toolTimeouts.ts` (same source of truth the outer wrapper already uses) and using it at all three call sites instead of a second constant. Added `aiToolsRemote.timeout.test.ts` asserting the `timeoutMs` that actually reaches `executeCommand` equals `getToolTimeout(<toolName>)` — no such consistency test existed anywhere before.

**2. IMPORTANT — verified, not fixed (by design, per conservative-fix instruction).** With the vision-tool budget now real at 120s, worst case (5-min approval wait + 120s execution ≈ 7 min) can exceed the 6-min `SDK_TURN_TIMEOUT_MS`. Checked whether the turn-timeout handler cleanly settles a pending approval/tool wait first — as implemented on `main` today it does **not**: `startTurnTimeout`'s callback only publishes `error`+`done` and sets `state = 'idle'`; it never aborts `session.abortController` or otherwise unblocks the in-flight wait, so a message that eventually arrives late can still get published into an already-ended turn (`runBackgroundProcessor` only stops on `closing`/`closed` state). This is a pre-existing gap (the same overlap already existed for two stacked approval waits), not new in kind — but this PR's budget raise makes the single-tool worst case newly reachable. The proper fix is already in flight as **#3104** (Closes #3089, shared per-cycle approval-wait budget bounded to the turn window) — not merged as of this comment. Per instruction, didn't restructure budgets here; documented the worst-case math and current behavior in a comment at the `SDK_TURN_TIMEOUT_MS` definition in `streamingSessionManager.ts` instead, and I'm flagging #3104 for prioritization.

**3. AGENT-SIDE — noted, not fixed here.** In service/headless mode the Go agent routes these two commands through the user-helper IPC channel with a 15s × 2 ≈ 30s ceiling (`agent/internal/heartbeat/handlers_screenshot.go:60`), independent of and beneath everything fixed on the API side. No Go code touched in this PR. Posting a comment on #3091 recommending a separate follow-up issue for the agent-side fix (fleet-managed-service deployment mode is exactly where this hits).

**4. Minor — corrected quietly.** The PR description's "WebRTC negotiation" framing was checked against the agent code and is wrong: `take_screenshot`/`computer_action` reuse an already-active `desktop.ScreenCapturer` and never touch the WebRTC/ICE path. The 5s ICE-gather timeout belongs exclusively to `start_desktop` Remote Desktop streaming (`agent/internal/remote/desktop/session.go:22`). Updated the PR description to reflect this.

### Test evidence
- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` (apps/api) — clean, exit 0 (default heap OOMs on this checkout, known repo gotcha)
- `eslint` on all changed files — clean
- Targeted `vitest run`: `aiToolsRemote.timeout.test.ts`, `aiToolsRemote.test.ts`, `aiToolsRemote.siteScope.test.ts`, `toolTimeouts.test.ts` — 29/29 passed
- Broader regression sweep: `vitest run src/services/aiTools*` — 1159/1159 passed across 87 files
- `streamingSessionManager.clientLoop.test.ts` + `streamingSessionManager.security.test.ts` (comment-only change to that file) — 13/13 passed

Working tree clean, HEAD `60255f1d1` pushed to `fix/3091-vision-tool-timeout` and matches this PR's `headRefOid`.

bdunncompany:
Confirmed all four, and thanks for the follow-through on the branch. I checked the critical one against the code rather than taking it on description:

```
$ git show 887c9159e:apps/api/src/services/aiToolsRemote.ts | grep -n timeoutMs
83:      }, { userId: auth.user.id, timeoutMs: 30000 });
141:      }, { userId: auth.user.id, timeoutMs: 30000 });
227:      }, { userId: auth.user.id, timeoutMs: 30000 });
```

and `waitForCommandResult` bounds its loop on exactly that value (`commandQueue.ts:546`, erroring `Command timed out after ${timeoutMs}ms`). So the inner 30s always won and my change would not have moved the production symptom at all. That is the substantive miss: I verified the *outer* wrapper — including that approval resolution sits outside it, which I reported as reassuring — and never traced inward to whether the tool implementation carried its own timeout. Checking one layer and generalising to the path is the error, and the tests I added encoded the same blind spot: they asserted the table's values and an ordering invariant, neither of which can fail while the effective budget is 30s. Your `aiToolsRemote.timeout.test.ts` is the assertion that actually binds.

Point 4 is a fair correction and I should not have stated it. The "WebRTC negotiation + capture" framing came from the issue text and I repeated it as established in the code comment, the commit message and the PR body without checking the agent. It reads as verification when it was inheritance.

Point 3 is now filed as **#3112** with the code cited (`handlers_screenshot.go`, `15 * time.Second` × `maxAttempts = 2`, and the `handlerRegistry[tools.CmdTakeScreenshot]` route into `executeToolViaHelper`). I framed it as the same structural defect one layer down rather than a number to raise: three layers each holding an independent constant, so the smallest silently decides. Suggested letting the server budget travel with the command so the layers nest by construction, with raising the per-attempt timeout as the smaller alternative — your call which is proportionate. Flagged that `computer_action` and any other helper-routed commands should be swept before fixing so it isn't screenshot-only.

On point 2, agreed it belongs in #3104 rather than here, and the comment at `SDK_TURN_TIMEOUT_MS` is the right place to leave the worst-case math.

Happy to take #3112 myself if you want it, or leave it — it's Go and in your agent path, so no assumption either way.

PR #3101, opened 2026-08-04
fix(agent): apply event_logs_query XPath filter and fix blank timestamps (#3092)

### LINKED ISSUE #3092: [Agent] event_logs_query: XPath query filter silently ignored; event timestamps returned blank
## Description
Two output-fidelity defects in the agent's `event_logs_query` command, observed on a Windows 11 Pro 25H2 device (US prod, 2026-08-03):

1. **XPath `query` filter silently ignored.** A query with an XPath provider filter (`*[System[Provider[@Name='Microsoft-Windows-USB-USBHUB3-Interoperability'] ...]]`) returned events that did not match the filter at all. The model noticed: "The event log query returned results but they aren't USB-specific (the XPath filter may have been applied loosely by the agent)." The `source` parameter DOES filter correctly — only the XPath `query` path is broken. Silent no-op filtering is worse than an error: the caller gets confidently wrong data.
2. **Blank timestamps.** Multiple `event_logs_query` responses returned events with empty timestamp fields; the model shrugged it off as "a known agent limitation." Timestamps are essential for correlating event-log evidence.

## Proposed Fix
- Implement or reject the XPath `query` parameter — if the agent can't apply it, return an explicit error instead of unfiltered results.
- Fix timestamp serialization in the event-log payload.
- Add a Go table test covering XPath filtering and timestamp presence.

## Affected Files
- Agent Windows event log command handler (`agent/` internal command for `event_logs_query`)

## Reported By

ToddHebebrand:
## Summary

Fixes both output-fidelity defects in the agent's `event_logs_query` command reported in #3092.

**Root cause 1 — XPath `query` silently ignored.** `QueryEventLogs` (`agent/internal/remote/tools/eventlogs.go`) only ever read `logName`/`level`/`source`/`eventId`/`page`/`limit` from the payload; a supplied `query` key was never read, so the `Get-WinEvent -FilterHashtable` query ran without it and returned unfiltered events with no error.

**Root cause 2 — blank timestamps.** Windows PowerShell 5.1's `ConvertTo-Json` serializes `DateTime` as `"\/Date(1754246096000)\/"`. `parseEventLogEntries` parsed `TimeCreated` with `time.Parse(time.RFC3339, ...)` only, so the parse always failed on PS 5.1 output and `TimeCreated` stayed the zero value.

## Changes

- **Implement the XPath `query` parameter** (per the issue's "implement or reject" guidance — implemented):
  - Non-empty `query` runs through `Get-WinEvent -LogName ... -FilterXPath ...` with `-ErrorAction Stop` in a try/catch: an invalid XPath now exits non-zero and the agent returns an explicit error (with PowerShell's stderr message) instead of unfiltered results. The benign "no matching events" case is detected via its locale-independent `FullyQualifiedErrorId` (`NoMatchingEventsFound*`) and returns an empty success — no reliance on localized error text.
  - `query` combined with `level`/`source`/`eventId` is rejected with an explicit error (those belong inside the XPath), and an oversized query (> 4096 bytes) is rejected rather than truncated — a truncated XPath would be a different query.
- **Fix timestamps**: the query and get-entry pipelines now project `TimeCreated` to a stable ISO 8601 UTC string (`.ToUniversalTime().ToString('o')`), which parses identically on PS 5.1 and PS 7. The Go parser additionally accepts the legacy `\/Date(ms)\/` form (defense in depth) via a tolerant `parseEventLogTime`.
- **Cross-platform testability**: script construction and output parsing moved from the `//go:build windows` file into untagged `eventlogs_query.go`, so the filter/parse logic is covered by unit tests on any OS. `eventlogs_windows.go` keeps only the `exec` plumbing.

## Testing

- `cd agent && go test -race ./internal/remote/tools/` — full package green (37s).
- New table-driven tests in `eventlogs_query_test.go`: timestamp parsing (ISO `'o'`, RFC3339 ± offset, escaped/unescaped `.NET` JSON dates, null/garbage), entry parsing with PS 5.1-style output asserting non-zero timestamps, XPath script construction (FilterXPath verbatim + quote escaping, `-ErrorAction Stop`, `NoMatchingEventsFound` handling, no hashtable fallback), legacy hashtable path unchanged, and `QueryEventLogs` validation (combine-conflict + length cap rejections).
- `GOOS=windows GOARCH=amd64 go build ./...` and host build both clean; `go vet` clean.
- Note: the PowerShell execution path itself is Windows-only; end-to-end verification of the fixed command against a live event log needs a Windows box.

Closes #3092

### DIFF
--- a/.claude/skills/agent-info/SKILL.md
+++ b/.claude/skills/agent-info/SKILL.md
@@ -129,7 +129,7 @@ Terminal commands use `term-` prefix IDs and skip DB persistence.
 | Type | Handler | Payload |
 |------|---------|---------|
 | `event_logs_list` | `ListEventLogs` | `{}` |
-| `event_logs_query` | `QueryEventLogs` | `{logName, level, source, page, limit}` |
+| `event_logs_query` | `QueryEventLogs` | `{logName, level, source, eventId, query, page, limit}` (`query` = XPath, exclusive with level/source/eventId) |
 | `event_log_get` | `GetEventLogEntry` | `{logName, recordId}` |
 | `tasks_list` | `ListTasks` | `{folder, page, limit}` |
 | `task_get` | `GetTask` | `{name, path}` |
--- a/agent/internal/remote/tools/eventlogs.go
+++ b/agent/internal/remote/tools/eventlogs.go
@@ -1,6 +1,8 @@
 package tools
 
 import (
+	"fmt"
+	"strings"
 	"time"
 )
 
@@ -17,9 +19,26 @@ func QueryEventLogs(payload map[string]any) CommandResult {
 	logName, _ := truncateStringBytes(GetPayloadString(payload, "logName", "System"), maxEventLogFieldBytes)
 	level := GetPayloadString(payload, "level", "")
 	source, _ := truncateStringBytes(GetPayloadString(payload, "source", ""), maxEventLogFieldBytes)
+	xpathQuery := strings.TrimSpace(GetPayloadString(payload, "query", ""))
 	eventID := GetPayloadInt(payload, "eventId", 0)
 	page := GetPayloadInt(payload, "page", 1)
 	limit := GetPayloadInt(payload, "limit", 50)
+
+	// The XPath query must be applied verbatim or rejected — never truncated
+	// (a truncated XPath is a different query) and never silently dropped
+	// (issue #3092: callers received confidently unfiltered results).
+	if len(xpathQuery) > maxEventLogXPathBytes {
+		return NewErrorResult(
+			fmt.Errorf("query exceeds the maximum XPath length of %d bytes", maxEventLogXPathBytes),
+			time.Since(startTime).Milliseconds(),
+		)
+	}
+	if xpathQuery != "" && (level != "" || source != "" || eventID > 0) {
+		return NewErrorResult(
+			fmt.Errorf("query (XPath) cannot be combined with the level, source, or eventId filters; put the conditions in the XPath expression instead"),
+			time.Since(startTime).Milliseconds(),
+		)
+	}
 
 	if page < 1 {
 		page = 1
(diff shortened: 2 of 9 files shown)


ToddHebebrand:

**Findings:** 5 raised → all addressed in `195e74ffe`; 0 outstanding (2 deferred by design, noted below).

- **Critical — PowerShell command injection** via Unicode single-quote lookalikes (U+2018–U+201B) in the new `query` field. `escapePowerShellSingleQuoted` doubled only the ASCII apostrophe; a crafted XPath with U+2019 broke out into a SYSTEM shell (reachable via AI `execute_command` + Helper end-user chat). Fixed: all five delimiter characters are doubled. Verified with `pwsh` — the escaped script parses as a single statement; the unescaped repro splits into injected commands.
- **Critical — invalid `-FilterHashtable` syntax** (pre-existing, enshrined by my first test): entries were joined with `" and "` instead of `";"`, a PowerShell parse error that made every `level`/`source`/`eventId`-filtered query silently return zero events — the same silent-failure class as this issue. Fixed to `";"`-separated; both query paths now run `-ErrorAction Stop` and surface real failures as explicit errors instead of empty success.
- **Important — CI `unused` lint failure**: `parseEventLogList` was unreferenced on the linux build. Moved back into the windows-tagged file with its caller.
- **Important — validation hardening**: non-string `query` and unknown/out-of-range `level` are now rejected explicitly (numeric level 1–5 accepted), closing the same "filter silently dropped" hole for type mismatches.
- **Docs**: `query` param + numeric level documented in `commands.mdx` and the agent-info skill.

**Deferred (stated for the maintainer):** wiring `query` through the HTTP/UI route (`routes/systemTools`) and the separate `startTime`/`endTime` accepted-but-ignored bug on that route. This issue was reported from AI-session transcripts (the AI path), which is fully fixed here; the HTTP route never forwarded `query`, so no regression. Worth a follow-up issue.

**Tests:** `go test -race ./internal/remote/tools/` green (full package, 137s). New table-driven coverage: timestamp parsing (ISO `'o'` + legacy `\/Date(ms)\/`), Unicode-quote escaping, `;`-separated hashtable builder, `-ErrorAction Stop`/`NoMatchingEventsFound` semantics on both paths, level/query validation (reject + accept + boundary + whitespace-routing), and multi-event/empty parser shapes.

**Lint/build:** CI `golangci-lint run --new-from-rev=origin/main` → 0 issues; `go vet` clean; host + `GOOS=windows` builds clean.

**Status:** review-clean, awaiting maintainer merge. Note: the PowerShell exec path is Windows-only, so end-to-end verification against a live event log still needs a Windows box.

PR #3103, opened 2026-08-04
fix(ai): record SDK session usage for partner-scoped users + fallback token accumulation (#3095)

### LINKED ISSUE #3095: [AI] ai_sessions token counters record 0 for real sessions — cost tracking not capturing usage

ToddHebebrand:
## Root cause

All-zero `ai_sessions` token counters on the SDK path were caused by the streaming processor's `result` handler reading `session.auth.orgId`:

```ts
const orgId = session.auth.orgId;
if (!orgId) {
  console.warn('... Skipping usage recording — no orgId on session', ...);
```

`AuthContext.orgId` is `string | null` and is **null for partner- and system-scoped users** — i.e. every MSP technician. So every turn of every session they ran hit the skip branch and recorded nothing, healthy or not. The `ActiveSession.orgId` field's own doc comment already said to use it (not `auth.orgId`) in background callbacks; the result handler was the one place that didn't.

## Changes (`apps/api/src/services/streamingSessionManager.ts`)

- **Use the canonical `session.orgId`** (captured from the `ai_sessions` row at creation, always set) in the result handler for both the guard and `recordUsageFromSdkResult`.
- **Fallback token accumulation**: per-API-call usage from SDK `assistant` messages is accumulated into a per-turn accumulator; when the `result` message arrives with missing/zero usage (already observed in practice — the code carried a defensive warn for it), the accumulated tokens are recorded instead. SDK-reported result usage still wins when non-zero.
- **Abandoned-turn flush**: if the processor exits without a `result` (teardown mid-turn, subprocess crash), non-zero accumulated usage is flushed to `recordUsageFromSdkResult` so spent tokens still reach the session counters and org cost aggregates.
- Accumulator resets on every flush, so multi-turn sessions never double-count; the existing `recordUsageFromSdkResult` `+=` semantics cover multi-turn totals.

## Tests

New `streamingSessionManager.usage.test.ts` (mocked SDK `query`, mocked cost tracker):

- partner-scoped auth (`orgId: null`) session records non-zero input/output tokens against the DB-row org — direct #3095 regression
- error-subtype results (turns dying on tool errors) still record usage
- empty result usage falls back to summed assistant-message usage incl. cache tokens
- SDK result usage preferred when present
- accumulator resets between turns (multi-turn)
- turn without a `result` flushes accumulated usage exactly once
- completed turn + teardown does not double-record

`vitest run` on the three streamingSessionManager suites: 20 passed. `tsc --noEmit` clean.

Closes #3095

### DIFF
--- a/apps/api/src/services/streamingSessionManager.ts
+++ b/apps/api/src/services/streamingSessionManager.ts
@@ -20,7 +20,7 @@ import { eq, and } from 'drizzle-orm';
 import type { AuthContext } from '../middleware/auth';
 import type { AiStreamEvent, AiApprovalMode } from '@breeze/shared/types/ai';
 import { AsyncEventQueue } from '../utils/asyncQueue';
-import { recordUsageFromSdkResult } from './aiCostTracker';
+import { recordUsageFromSdkResult, calculateCostCents } from './aiCostTracker';
 import { sanitizeErrorForClient } from './aiAgent';
 import { captureException } from './sentry';
 import { createBreezeMcpServer, BREEZE_MCP_TOOL_NAMES } from './aiAgentSdkTools';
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
## Review round — findings + resolution

Two issues found in the abandoned-turn flush (the `finally` block covering teardown mid-turn / subprocess crash / iterator error):

1. **Fire-and-forget promise on the most common trigger.** The abandoned-turn write to `recordUsageFromSdkResult` was a detached `.catch(...)`, not awaited. The most common way this branch fires is process shutdown (deploy/SIGTERM) — exactly the case where an untracked promise can be killed before it settles, silently, since even the `.catch` handler never gets a chance to run.
   - **Fix:** await the write inline inside a `try/catch`. It's now tied into `processorPromise` so it's guaranteed to be drained before the session finishes tearing down. Failures are logged via `captureException` + `console.error` but never retried — the underlying `recordUsageFromSdkResult` write is a non-idempotent `+=` increment, so a retry would double-count tokens already recorded.

2. **Per-user usage hook missing from the abandoned-turn path.** The `result`-message success path already calls `session.recordExtraUsage(...)` (the AI-for-Office per-user hook backing `client_ai_usage` caps/invoicing) alongside the org-level `recordUsageFromSdkResult`. The abandoned-turn flush only had the org-level call — so abandoned-turn spend reached the org ledger but silently never reached the per-user ledger.
   - **Fix:** mirror the same `recordExtraUsage` call in the abandoned-turn branch, computing cost via `calculateCostCents(session.model, ...)` (the abandoned-turn path has no `total_cost_usd` from the SDK to reuse, unlike the result path).

## Test evidence

- New test: `feeds abandoned-turn usage to the per-user recordExtraUsage hook (client sessions)` — asserts both the org-level `recordUsageFromSdkResult` call and the per-user `recordExtraUsage` call fire with the correct token counts and calculated cost for a turn that never produces a `result` message.
- `npx vitest run streamingSessionManager.usage.test.ts streamingSessionManager.clientLoop.test.ts streamingSessionManager.security.test.ts` → **3 files / 21 tests passed** (the original 20 from #3095 + this review round's new test).
- `tsc --noEmit` on `apps/api` → clean, no errors.

Commit: a5b62d334

Issue #2686, opened 2026-07-20
AI guardrails: no contract test binds tierConfig.ts (customer-facing) to the real tier tables


ToddHebebrand:
The AI Risk dashboard renders a tier explainer from `apps/web/src/components/ai-risk/tierConfig.ts`, which is a hand-maintained mirror of `TIER1_ACTIONS` / `TIER2_ACTIONS` / `TIER3_ACTIONS` in `apps/api/src/services/aiGuardrails.ts`. Nothing enforces that the two agree, and `apps/web/src/components/ai-risk/` has no test file at all.

It had already drifted. Before the fix below, `tierConfig.ts` listed:

- `file_operations (list/read)` under **Tier 1** — wrong since SR5-01 moved both to Tier 3, so this had been misstating the guardrails for some time
- `manage_patches (scan/install/rollback)` under **Tier 3** — `scan` moved to Tier 2

This is the page an MSP shows a customer to explain what the AI can do without asking permission. Telling someone "patch scan requires your approval" when it auto-executes is a support-and-trust problem, not a cosmetic one — and the `file_operations` entry was understating a privileged capability rather than overstating it.

Both entries are corrected in the inline self-approve branch, but only by hand, so nothing stops the next drift.

**Proposed guard:** a test that imports the guardrail tables and asserts every `tierConfig` entry matches `checkGuardrails(tool, { action })` for the tool/action pairs it names. Requires exporting the three tables from `aiGuardrails.ts` (a test-only export is fine). Failure message should name the tool, action, claimed tier, and actual tier.

**Related, same file, worth doing together:** there is no disjointness assertion across the three tier tables. Resolution is first-match in the order TIER1 → TIER3 → TIER2 (`aiGuardrails.ts` ~715-733), so a duplicate entry silently wins by table position instead of erroring. The dangerous direction is a leftover in TIER1: it would silently de-escalate a Tier-3 action, and no test catches it unless that specific action happens to be enumerated in the existing case tables. Moving actions between tiers — as the tier recalibration just did for five of them — is exactly when a stale duplicate gets left behind.

A pairwise-empty-intersection assertion over the three tables is a few lines and closes it.

Also worth a look: `apps/docs/src/content/docs/features/ai.mdx` carries the same tier rows and has the same drift exposure, though docs drift is lower stakes than in-product copy.

PR #3105, opened 2026-08-04
fix(ai): tier read-only execute_command commandTypes down to auto-execute + audit (#3088)

### LINKED ISSUE #3088: [AI] Approval fatigue: every read-only execute_command is high-tier, one WebAuthn tap per call, no batching

ToddHebebrand:
Closes #3088

## Problem

Every `execute_command` is flat Tier 3, so a routine AI troubleshooting session produced one WebAuthn-backed approval request per read-only command — 25 approval requests / 19 taps / 6 expired in a 35-minute production stretch, all for `event_logs_query`, `event_logs_list`, `list_processes`, `list_services`, and `file_list`.

The flat tier was an inconsistency, not a policy: the same physical operations already run lower via sibling tools — `manage_processes (list)` is Tier 1, `manage_services (list)` and `file_operations (list)` are Tier 2 (SR5-01 explicitly classed directory listing as recon-only, auto-execute + audit).

## Change

`checkGuardrails` now resolves `execute_command`'s tier per `commandType` — the agent-side handler discriminator, an explicit conservative allowlist (never a heuristic over command strings):

- **Tier 2 (auto-execute + audit):** `list_processes`, `list_services`, `file_list`, `event_logs_list`, `event_logs_query`.
- **Tier 3 (unchanged):** `kill_process`, `start_service`, `stop_service`, `restart_service`, and `file_read` — the latter deliberately, per the SR5-01 rationale (arbitrary-file exfiltration off a root/LocalSystem agent). An unknown, missing, or non-string `commandType` fails closed to Tier 3.

Mechanism: a new `TOOL_ACTION_INPUT_KEYS` map points `execute_command` at its `commandType` discriminator, so the existing `TIER1/2/3_ACTIONS` tables, the disjointness contract test, and both #2686 tier-parity guards (web `tierConfig.ts` + docs `ai.mdx`, both updated here) apply unchanged.

## Design notes (issue option 2, and scope choices)

- **Session-scoped batch approval** ("approve read-only commands for this session") is satisfied by the existing `auto_approve` session mode: it auto-executes Tier 2 with an audit ledger row while keeping Tier 3 per-step. Once these commandTypes are Tier 2, that mode is exactly the issue's option 2 — one decision covers the investigation burst. No new grant surface, table, or approval-model change was added; extending batch grants to *mutating* Tier-3 actions would weaken the deliberate four-eyes/WebAuthn model and is left as a separate policy decision.
- **MCP unaffected on purpose:** `tools/call` resolves `Math.max(baseTier, guardrailTier)`, so `execute_command` remains wholly gated over the unattended MCP transport (the 2026-08-02 all-Tier-3-interactive decision stands).
- Under the default `per_step` chat mode these commands still show the lightweight in-chat Tier-2 confirmation (no durable intent, no approver fan-out, no WebAuthn); `auto_approve` removes even that.

## Tests

- `aiGuardrails.test.ts`: new commandType cases — five read-only types → Tier 2 no-approval; five mutating/read-sensitive types → Tier 3; fail-closed on missing/unknown/non-string `commandType`; spoofed `action` key cannot downgrade.
- Parity suites updated + green against the updated web/docs mirrors.
- Integration suites that used `execute_command`+`list_processes` as their canonical Tier-3 intent now use `kill_process` (behavior-preserving data swap).

Verification: `vitest run` on aiGuardrails + both parity suites (138 passed) and mcpServer approvalGate/effectiveTier/main + aiAgentSdk + helperToolFilter + aiTools (177 passed); `tsc --noEmit` clean on apps/api and apps/web. The edited integration suites need a live DB (Docker unavailable locally) — they run in the Integration Tests CI job; the edits there are mechanical commandType swaps.


ToddHebebrand:
## Review round (security emphasis) — complete

Ran `pr-review-toolkit:code-reviewer` and `pr-review-toolkit:silent-failure-hunter` against the diff, plus direct verification against the Go agent source for what each allowlisted `commandType` actually returns.

### Task questions, answered

**Is the read-only classification an explicit conservative allowlist of commandTypes (not a heuristic over command strings)?**
Yes. `TIER2_ACTIONS.execute_command` is a literal string array, and `TOOL_ACTION_INPUT_KEYS` makes `commandType` (the agent-side dispatch discriminator, not free text) the resolution key for this tool. Unknown, missing, or non-string `commandType` falls through to the tool's base Tier 3 — verified fail-closed for every malformed shape (array, object, empty string, spoofed `action` key).

**Can any allowlisted "read-only" commandType actually mutate state or exfiltrate secrets?**
Two of the original five could — **fixed in this round**:
- `list_services`: the Go agent (`services_windows.go`, `services_linux.go`) returns the service's full binary path/command line unredacted (only length-truncated). Legacy/third-party services commonly embed credentials there (`-p <password>`, DB connection strings).
- `event_logs_query`: `logName` is free-form with no denylist excluding `Security`, and the raw `Message` field is returned unredacted. A failed logon (Windows event 4625) can carry a mistyped password in the `Account Name` field.

Both are reverted to Tier 3 — same exfiltration rationale the PR already applied to `file_read` (SR5-01), just missed for these two. `list_processes`, `file_list`, and `event_logs_list` were verified safe: no `CommandLine`/content field in the process-list path, directory listing goes through the same containment guard as `file_read`, and the event-log-list path returns log metadata only (no entry content).

**Are auto-executed calls still fully audited?**
Yes. Traced the Tier-2 path in `aiAgentSdk.ts` (`checkGuardrails` → `onPreToolUse`): the audit-ledger insert is `await`ed and gates dispatch — if the insert throws, the handler returns `allowed: false` and the command never reaches the device. Not fire-and-forget. The post-execution outcome update is best-effort (wrapped try/catch), but the audit record proving execution happened is guaranteed to exist first.

**Session-scoped batch approval — bounded and does it die with the session?**
This PR doesn't add new grant machinery; it makes the pre-existing `auto_approve` session mode (a field on the in-memory streaming session object, `streamingSessionManager.ts`) apply to a wider Tier-2 set. It's gated at Tier level per session (not per-commandType — pre-existing behavior), set once from the org's `aiBudgets.approvalMode` policy at session creation, and naturally expires with the session object. Confirmed no persistence beyond session lifetime.

### Additional confirmed findings (fixed)
- Two mirror doc pages not covered by the `#2686` tier-parity contract test (`ai-computer-control.mdx`, `agents/helper.mdx`) were left stating `execute_command` is uniformly Tier 3 — now corrected to reflect the split.
- The new tool description overstated auto-execution: under the default `per_step` approval mode, Tier-2 `execute_command` calls still take a lightweight inline confirmation (no durable intent, no WebAuthn) before running — only `auto_approve`/plan session modes skip that. Description text corrected.
- Noted for awareness (not changed — out of scope for this PR): PAM rules keyed on `matchRiskTier` (`pamRuleEngine.ts:293`) use exact equality, so an org's `matchRiskTier: 3` auto-deny/require-approval rule silently stops covering the three commandTypes now at Tier 2. Worth a follow-up if PAM rules are commonly scoped this way.

### Rejected / not actionable
- Nothing else from either review agent rose to "confirmed and consequential" — no spoofing vector on the discriminator resolution, MCP transport still gates via `Math.max(baseTier, guardrailTier)` unaffected, tier-parity mirrors all still agree post-fix.

### Test evidence
- `aiGuardrails.test.ts` + both tier-parity contract suites (`aiGuardrailsTierConfig.parity.test.ts`, `aiGuardrailsAiDocs.parity.test.ts`): **138 passed**.
- `aiToolsScripts.commandTypes.test.ts` + `aiToolsScripts.runScript.orgEquality.test.ts`: **17 passed**.

Final commit: `3b54f650a`. No merge/close action taken.

Issue #2822, opened 2026-07-26
[API][security] Audit other partner-axis reads for the missing system-context RLS gap


ToddHebebrand:
### Description

PR #2818 fixed a partner-axis read that was silently inert under RLS: `getEnrollmentDefaultsForOrg` read `partners.settings` on the ambient request context, but the `partners` SELECT policy is `breeze_has_partner_access(id)` and `computeAccessiblePartnerIds` returns `[]` for `scope === 'organization'` (`apps/api/src/middleware/auth.ts:307-314`). For org-scoped callers the row came back empty and the value collapsed to a default — for exactly the population the feature was meant to constrain.

**No repo-wide audit was done for other reads with the same shape.**

### Why this is worth a sweep

The failure is silent in both directions that matter: RLS returns no rows rather than an error, and every route/service test in the affected area mocked the DB, so the entire unit suite passed against a control that did nothing. It took an integration test running as `breeze_app` in a genuine org-scoped context to surface it.

### Suggested approach

1. Enumerate every read of `partners` (or any partner-axis table) that is not already inside a system context.
2. For each, determine which caller scopes reach it. Org-scoped callers cannot see `partners` without `runOutsideDbContext` + `withSystemDbAccessContext`.
3. Where a system-context escape is needed, follow the pattern in `apps/api/src/services/enrollmentDefaults.ts` — including its branch that skips the escape when already system-scoped, which exists to avoid holding two pooled connections (see #2818 for why that matters against the US connection ceiling).
4. Prefer a real-Postgres test per fix; a mocked-DB test structurally cannot catch this class.

Related memory: partner-axis reads need a system context; RLS is stricter than the app layer, never assume parity.

ToddHebebrand:
## Repo-wide audit complete — full enumeration

Swept every read of `partners` and of all 51 tables in `PARTNER_TENANT_TABLES` across `apps/api/src` (routes, services, jobs, lib, middleware, oauth, modules, workers, db). **~460 read sites examined**, plus raw-SQL references and every hand-rolled `DbAccessContext` literal.

PR #2828 fixes the root cause and the highest-severity unambiguous sites. The rest are below — several need a ruling, not a mechanical fix, which is why they are not in that PR.

### Two premise corrections worth propagating

[...]

PR #3106, opened 2026-08-04
fix(ai): run device-bound chat tools under the device's org, not the login org (#3087)

### LINKED ISSUE #3087: [AI] Device-bound chat sessions run org-scoped tools under the login org, not the device's org
## Description
During a US-prod AI chat session (2026-08-03) opened on a device belonging to a **sibling org under the same partner**, org-scoped tools ran under the *login* org instead of the device's org:

- `manage_patches` (action `list`) returned the login org's Linux patches for a query about a Windows workstation in the sibling org. The model noticed and discarded them ("Those patches aren't for this device specifically"), but it burned a turn.
- Every `search_logs` call in three consecutive sessions returned `{"logs": [], "total": 0}`. These were likely scope artifacts of the same mismatch, not true negatives — the RMM-side log search never looked at the org that owns the device. The investigation's conclusions were degraded for the entire session chain.

Not a tenant-isolation breach — both orgs belong to the same partner, and RLS held. It is a correctness bug: a device-bound session answers device questions with data from a different org.

## Root Cause (likely)
`ai_sessions.org_id` is stamped with the caller's login org, and tool handlers derive scope from `getOrgId(auth)` (e.g. `apps/api/src/services/aiToolsFleet.ts:637`) rather than from the session's bound device. Direct agent commands (`execute_command`) go to the device and are unaffected — only RMM-side org-scoped reads mis-scope.

## Proposed Fix
When a session is bound to a device, resolve the effective org from the device (partner-scope permitting) and use it for org-scoped tool queries — or have device-scoped tools accept the device's org explicitly. At minimum, tools like `search_logs` should scope to the bound device's org when a deviceId/session device is present.

## Affected Files
- `apps/api/src/services/aiToolsFleet.ts` (`manage_patches` list path, `getOrgId(auth)`)
- `apps/api/src/services/aiAgentSdkTools.ts` / session org binding
- `search_logs` tool implementation

## Reported By

ToddHebebrand:
Closes #3087

## Problem

A chat session bound to a device in a **sibling org under the same partner** ran org-scoped tools under the *login* org: `manage_patches` (list) returned the login org's patches, and every `search_logs` call returned `{logs: [], total: 0}` because the RMM-side log search never looked at the org that owns the device. Not a tenant-isolation breach (RLS held; both orgs belong to the partner) — a correctness bug that silently corrupts investigations.

## Root cause

`ai_sessions.org_id` is **already** correctly anchored to the device's org at creation (`services/aiAgent.ts` `createSession`, gated on `auth.canAccessOrg` + `auth.canAccessSite`). But tool execution (`aiAgentSdkTools.ts` `makeHandler`/`makeSessionAwareHandler`) took `getAuth() => session.auth` — the **raw login AuthContext** — for both the RLS `DbAccessContext` and the tool handlers. Partner tokens carry `orgId: null`, so the `getOrgId(auth)` helper (cloned in 15 `aiTools*` modules) fell back to `accessibleOrgIds[0]` — an arbitrary org from an unordered partner-org select — and `search_logs`' `auth.orgCondition(...)` spanned the whole partner. The session's `device_id`/`org_id` were never consulted.

## Fix

One chokepoint in `StreamingSessionManager` covering all ~200 tools:

- **`ActiveSession.toolAuth`** — the request auth narrowed to the session (device) org via new `buildDeviceBoundSessionAuth`, wired into the MCP server's `getAuth` thunk (tool handlers + their RLS context). Recomputed on every per-request auth refresh, so a follow-up message never reverts to login scope.
- **Narrowing contract**: `orgId`/`accessibleOrgIds` pinned to the session org with closures from the canonical `buildOrgAccessClosures`; `scope`/`partnerId` **preserved** so the partner RLS axis stays intact (partner-wide scripts/templates/rings stay readable — the #2822 failure mode); site restrictions and identity preserved; throws if the caller cannot access the session org (defensive — `getSession` already pre-filters).
- **`session.auth` stays RAW**: RBAC (`checkToolPermission` — `permissions.ts` resolves org membership before partner role), rate limits, and audit attribution keep resolving the login identity/role, matching the rest of the API. A dual-membership tech does not flip roles because a session is device-bound.
- Non-device sessions unchanged — partner techs keep fleet-wide reach in general chat. Helper/client-AI paths already pin their auth via their own middleware and are no-ops here.
- Two adjacent same-root-cause fixes: SDK `result` usage recording used `auth.orgId` (null for partner logins — **usage recording was silently skipped**) and `postToolUse` audit org attribution — both now use the canonical `session.orgId`.

## Product contract (explicit)

A device-bound session's org-scoped reads/writes target the **device's org**. Partner-*role* capabilities (partner-wide config reads, `get_fleet_status` by partnerId) are intentionally retained — the narrowing is the org query axis, not a full sandbox.

## Design review

Codex (gpt-5.6, xhigh, read-only) independently reviewed the design: agreed the session-manager chokepoint is correct (patching the 15 `getOrgId` clones would miss `orgCondition`/RLS/future tools; `makeHandler` is too late for `preToolUse`), and flagged the RBAC role-flip risk that motivated the raw-vs-tool auth split above.

## Tests

- New `streamingSessionManager.deviceBoundAuth.test.ts` (8 tests): partner-scope login narrowed to a sibling device org (org axis pinned, `accessibleOrgIds[0]` = device org), partner RLS axis preserved in `dbAccessContextFromAuth` (#2822 guard), site closures preserved, already-pinned identity short-circuit, system-scope behavior, fail-loud throw, and `getOrCreate` wiring: device-bound `getAuth()` is narrowed on creation **and** refresh, `session.auth` stays raw, non-device sessions untouched.
- `vitest run` green: deviceBoundAuth + clientLoop + security (21), aiAgentSdk + planMatch + m365risk + sessionAware (118), ai_sessions_crud + ai_sessions_actions + aiAgent.deviceTask (37). `tsc --noEmit` clean.

### DIFF
--- a/apps/api/src/routes/ai.ts
+++ b/apps/api/src/routes/ai.ts
@@ -617,6 +617,9 @@ aiRoutes.post(
         maxTurns: dbSession.maxTurns,
         turnCount: dbSession.turnCount,
         systemPrompt: dbSession.systemPrompt,
+        // Device-bound sessions narrow tool execution to the device's org
+        // (ai_sessions.org_id), not the login org (#3087).
+        deviceId: dbSession.deviceId,
       },
       auth,
       c,
--- a/apps/api/src/routes/ai_sessions_crud.test.ts
+++ b/apps/api/src/routes/ai_sessions_crud.test.ts
@@ -355,4 +355,62 @@ describe('AI routes', () => {
     });
   });
 
+  // ============================================
+  // POST /sessions/:id/messages
+  // ============================================
+  describe('POST /ai/sessions/:id/messages', () => {
+    it('passes the bound device id from the DB session into streamingSessionManager.getOrCreate (#3087 route wiring)', async () => {
+      // This is the seam that arms the #3087 fix: getOrCreate uses `deviceId`
+      // to decide whether to narrow tool execution to the device's org. If
+      // this route ever stops forwarding it, the whole narrowing mechanism in
+      // streamingSessionManager goes dark even though its unit tests stay green.
+      vi.mocked(runPreFlightChecks).mockResolvedValueOnce({
+        ok: true,
+        session: {
+          id: SESSION_ID,
+          orgId: ORG_ID,
+          sdkSessionId: null,
+          model: 'claude-sonnet-4-5-20250929',
+          maxTurns: 50,
+          turnCount: 0,
+          systemPrompt: null,
+          title: 'Existing title', // truthy — skips the auto-title-generation branch
+          deviceId: 'device-42',
+        } as any,
+        sanitizedContent: 'hello there',
+        systemPrompt: 'SYSTEM PROMPT',
+        maxBudgetUsd: undefined,
+      });
+
+      const fakeActiveSession = {
+        state: 'ready',
+        eventBus: {
+          subscribe: vi.fn(() => (async function* () {})()),
+          unsubscribe: vi.fn(),
+        },
+        inputController: { pushMessage: vi.fn() },
+      };
+      vi.mocked(streamingSessionManager.getOrCreate).mockResolvedValueOnce(fakeActiveSession as any);
+      vi.mocked(streamingSessionManager.tryTransitionToProcessing).mockReturnValueOnce(true);
+
+      const valuesSpy = vi.fn().mockResolvedValue(undefined);
+      vi.mocked(db.insert).mockReturnValue({ values: valuesSpy } as any);
+
+      await app.request(`/ai/sessions/${SESSION_ID}/messages`, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json', Authorization: 'Bearer token' },
+        body: JSON.stringify({ content: 'hello there' }),
+      });
+
+      expect(streamingSessionManager.getOrCreate).toHaveBeenCalledWith(
+        SESSION_ID,
+        expect.objectContaining({ deviceId: 'device-42' }),
+        expect.anything(),
+        expect.anything(),
+        'SYSTEM PROMPT',
+        undefined,
+      );
+    });
+  });
+
 });
--- a/apps/api/src/services/actionIntents/revalidateRelease.ts
+++ b/apps/api/src/services/actionIntents/revalidateRelease.ts
@@ -20,12 +20,15 @@ import { canonicalizeArguments, computeArgumentDigest } from './canonicalize';
  *
  * Returns the freshly-rebuilt actor `auth` on success. Callers decide how to
  * EXECUTE: the worker executes under this rebuilt context; the inline chat path
- * executes under its live `session.auth` (which alone carries the session-aware
- * M365/Google connection context) — but only AFTER this returns ok, i.e. only
- * once the requester's CURRENT authorization has been re-proven. The rebuilt
- * `auth` and `session.auth` describe the same user + org (accessibleOrgIds ===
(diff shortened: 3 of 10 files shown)


ToddHebebrand:
## Code review + coverage follow-up

**Independent code review (pr-review-toolkit code-reviewer, full diff)** — clean verdict: the core narrowing (`buildDeviceBoundSessionAuth`) never widens access, the partner RLS axis (#2822) is genuinely preserved, no regression to helper/clientAi/scriptAi paths (they already pin their own auth), the defensive throw is correctly unreachable, and usage recording now agrees with the budget gate. `tsc --noEmit` clean; 25 files / 310+ tests green at review time.

One IMPORTANT finding, now fixed:

- **`aiAgentSdkTools.ts` `propose_action_plan`** still guarded on `session.auth.orgId`, which is `null` for partner-scope logins — exactly the population this PR narrows tool execution for. That guard hard-failed every action-plan proposal ("Action plans require an organization context") for a device-bound partner-scope session in `action_plan`/`hybrid_plan` approval mode, even though the canonical `session.orgId` was available. Every other DB write in that handler already used `session.orgId`; this was the one holdout. Fixed to `const orgId = session.orgId` and dropped the dead null guard. Covered by a new discriminating test (`aiAgentSdkTools.proposeActionPlan.test.ts`) exercising a partner-scope login through the full plan-approval lifecycle.

Two cheap sub-threshold touch-ups applied alongside it:

- `streamingSessionManager.ts` `getOrCreate`'s session-reuse path now re-narrows `toolAuth` against the freshly-loaded `dbSession.orgId` instead of the possibly-stale `existing.orgId` snapshot captured at session creation — survives a device being moved to a different org mid-session.
- Corrected a stale comment in `revalidateRelease.ts` that described the inline chat release path as executing under `session.auth` — after this PR that's `session.toolAuth` for device-bound sessions (`session.auth` stays raw for RBAC/audit).

**Test-coverage review findings, addressed:**

1. Usage-recording bookkeeping — added a partner-auth (`orgId: null`) case in `streamingSessionManager.clientLoop.test.ts` proving the SDK result handler records usage against `session.orgId`, not `auth.orgId` (which would silently skip recording for partner logins).
2. `postToolUse` audit org attribution — added a case in `aiAgentSdk.test.ts` proving the persisted audit event carries the session org, not the (possibly null) login `auth.orgId`.
3. Route seam — added an assertion in `ai_sessions_crud.test.ts` that `POST /sessions/:id/messages` passes `deviceId` from the DB session into `streamingSessionManager.getOrCreate`; this is the wiring that arms the whole fix, and was previously untested (deleting that one line left the rest of the suite green).

**Test evidence** (individual `vitest run` per file, host under heavy load so run one at a time):
- `streamingSessionManager.deviceBoundAuth.test.ts` — 8 passed
- `streamingSessionManager.clientLoop.test.ts` — 4 passed
- `streamingSessionManager.security.test.ts` — 10 passed
- `aiAgentSdk.test.ts` — 96 passed
- `aiAgentSdkTools.sessionAware.test.ts` — 8 passed
- `aiAgentSdkTools.mcpCoverage.test.ts` — 7 passed
- `aiAgentSdkTools.proposeActionPlan.test.ts` (new) — 1 passed
- `ai_sessions_crud.test.ts` — 14 passed
- `actionIntents/revalidateRelease.test.ts` — 2 passed
- `tsc --noEmit -p .` (`NODE_OPTIONS=--max-old-space-size=8192`) — clean

HEAD is now `4fdb6ced7`.

PR #3108, opened 2026-08-04
fix(ai): record explicit tool_result when the SDK drops a call before dispatch

### LINKED ISSUE #3094: [AI] set_device_context call produced no tool_result (value with parentheses); paired approval expired unattended
## Description
In a US-prod session (2026-08-03), the model called `set_device_context` with a value containing parentheses (`"ethernetIP": "169.254.7.26 (APIPA)"`). That call produced **no tool_result row at all** — no error, no success. The model retried 5s later with the parens sanitized out (`"169.254.7.26_APIPA"`), which succeeded. Separately, the approval ledger shows a medium-tier `set_device_context` approval_request created at the retry timestamp that was never decided and expired at the 5-minute TTL — yet a success result was recorded within seconds, so the relationship between the approval row and the executed call is unclear.

Two things to investigate:
1. Why a `set_device_context` input with parentheses vanishes without a tool_result (validation rejecting silently? sanitizer throwing outside the result path?).
2. Why a `set_device_context` execution apparently succeeded while its approval_request sat pending → expired (did the retry bypass approval, or is the ledger row orphaned?). If a medium-tier action can execute while its approval row is pending, that is an approval-bypass worth ruling out.

## Affected Files
- `set_device_context` tool handler + approval-request creation path (`apps/api/src/services/aiAgentSdk.ts`, aiTools device-context implementation)

## Reported By

ToddHebebrand:
## Summary

`set_device_context` calls carrying a parenthesized value (e.g. `169.254.7.26 (APIPA)`) could vanish from the AI session transcript with no `tool_result` — the Claude Agent SDK rejected the tool call (an MCP `-32602` input-validation error) *before* our own `preToolUse`/`postToolUse` handlers ever ran, so nothing recorded the failure: no `ai_messages` row, no SSE event, and a stale `toolUseIdQueue` entry that misattributed the *next* tool's result.

Root cause is at the SDK/MCP transport layer, not in our own validation — our `set_device_context` Zod schema (`aiToolSchemas.ts`) already accepts parentheses in both `summary` and `details`, which the new `aiToolSchemas.validateToolInput.test.ts` cases pin explicitly.

## Fix

`StreamingSessionManager`'s background processor now detects these orphaned `tool_result` blocks in the SDK's `'user'` message case (an id still present in `session.toolUseIdQueue`, meaning `postToolUse` never shifted it) and records an explicit error result instead of silence:
- persists an `ai_messages` `tool_result` row (`droppedBeforeExecution: true`),
- emits the SSE `tool_result` event so the UI tool card resolves instead of spinning,
- clears the stale `toolUseIdQueue`/`toolUseNames` entries so later calls aren't misattributed,
- auto-flags the session (first occurrence) for flagged-session review.

A companion `toolUseNames` map (id → bare tool name, populated at `content_block_start`) lets the fallback path name the dropped tool; it's drained normally by `postToolUse` and the client-tool handler so it never leaks into the fallback for calls that executed normally.

## Safety verdict: expired approval_requests row next to a recorded success — orphaned bookkeeping, NOT an approval bypass

Verified by directly reading the gating code (not inferring from names):

- The actual execution gate, `waitForApproval` (`apps/api/src/services/aiAgent.ts:301-359`), reads **only** `aiToolExecutions.status` in its poll loop (`aiAgent.ts:310-323`). `approvalRequests` is never imported or referenced anywhere in that function.
- `aiToolExecutions` is created first; `approvalRequests` is created afterward and references it via `executionId` (`apps/api/src/services/aiAgentSdk.ts:536-559` then `:944-966`), purely as a bridge for the mobile `/approvals` UI surface — the schema comment on `approval_requests.execution_id` (`apps/api/src/db/schema/approvals.ts:47-54`) confirms this.
- The 30s expiry reaper that flips stale `approvalRequests` rows to `expired` and mirrors the decision onto `aiToolExecutions` (`apps/api/src/jobs/approvalExpiryReaper.ts:127-141`) guards that mirror write with `eq(aiToolExecutions.status, 'pending')` — it is structurally incapable of overwriting an already-resolved (`approved`/`completed`) execution row.
- The mobile `/approvals` decide route (`apps/api/src/routes/approvals.ts:681-727`) CASes `approvalRequests` first, then mirrors onto `aiToolExecutions` only if present, in a try/catch that's explicitly documented as non-fatal — if the mirror fails, `waitForApproval` just times out and self-rejects; it never silently allows.

Conclusion: a `approvalRequests` row sitting `pending` past its TTL while the linked `aiToolExecutions` row already shows `approved`/executed is a **leftover mobile-notification row**, not a second authorization gate that got skipped. This PR additionally makes `handleApproval` (the in-app web-chat decide path, `aiAgent.ts`) mirror its decision onto the still-pending `approvalRequests` row — previously it only flipped `aiToolExecutions`, which is why the bridge row was stranded — closing that cosmetic gap so audit review no longer sees a spuriously-expired row next to a resolved execution.

Caveat: the Tier-3 durable `action_intents` release worker (`apps/api/src/jobs/intentReleaseWorker.ts:277-285`) *does* consult `approvalRequests.status` as a fail-closed pre-execution check — but that's a distinct code path from the Tier-2 `per_step` flow this issue describes, and it fails *closed* (blocks execution on a missing/mismatched row), so it doesn't change the verdict above.

## Testing

- `apps/api/src/services/streamingSessionManager.droppedToolResult.test.ts` (new) — 4 cases: persists explicit error result + SSE event for a genuinely dropped call; does not double-record a call `postToolUse` already handled; ignores plain-text resume-replay user messages; keeps queue attribution correct for a retry after a drop.
- `apps/api/src/services/aiToolSchemas.validateToolInput.test.ts` — pins that `set_device_context` accepts parentheses in `summary`/`details`, and that the realistic SDK-layer rejection triggers are type/length errors, not parens.
- `apps/api/src/services/aiAgent.authz.test.ts` — new `handleApproval approval_requests mirror (#3094)` describe block: approve/deny mirror onto the bridge row, and mirror-failure doesn't fail the primary decide.
- Full targeted run: `streamingSessionManager.droppedToolResult.test.ts`, `aiAgent.authz.test.ts`, `aiToolSchemas.validateToolInput.test.ts` — 27/27 passed.
- Broader regression run: `aiAgentSdk.test.ts`, `aiAgentSdk.m365risk.test.ts`, `aiAgentSdk.planMatch.test.ts`, `clientAiTools.contract.test.ts`, `clientAiTools.handler.test.ts`, `clientAiTools.registry.test.ts`, `streamingSessionManager.clientLoop.test.ts`, `streamingSessionManager.security.test.ts`, `aiAgent.test.ts` — 165/165 passed.
- `tsc --noEmit` (with `NODE_OPTIONS="--max-old-space-size=8192"` per the known OOM issue on this box) — clean, no errors.

Closes #3094

### DIFF
--- a/apps/api/src/services/aiAgent.authz.test.ts
+++ b/apps/api/src/services/aiAgent.authz.test.ts
@@ -32,6 +32,10 @@ vi.mock('../db/schema', () => ({
     sessionId: 'aiToolExecutions.sessionId',
     intentId: 'aiToolExecutions.intentId',
   },
+  approvalRequests: {
+    executionId: 'approvalRequests.executionId',
+    status: 'approvalRequests.status',
+  },
   delegantM365Connections: {},
   devices: {},
 }));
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
## Review summary (self-review — automated reviewer stalled twice and was not retried)

**Recommendation: ship as-is.** No consequential findings.

Checked specifically:

1. **FIFO queue matching correctness** (`streamingSessionManager.ts`, `recordDroppedToolResult`): matches by explicit `indexOf(toolUseId)` on the SDK-supplied `tool_use_id` rather than positional `shift()`, so interleaved/out-of-order tool calls can't cause misattribution — each id is matched to itself regardless of queue position, then spliced out. `postToolUse` still uses `shift()` for its own (normal, in-order) path; the two never race on the same id since `postToolUse` deletes the paired `toolUseNames` entry it drains, and the fallback only fires for ids `postToolUse` never touched.

2. **Redaction/truncation of SDK-authored error text**: reuses the existing `redactAiToolOutputText` utility before persisting to `ai_messages`, emitting over SSE, or writing into `flagReason` (capped separately at 1000 / 300 chars). Same redaction path already trusted elsewhere in this file — no new exposure surface introduced.

3. **`handleApproval` approval_requests mirror best-effort semantics**: the mirror write is the last statement before an unconditional `return true;` — a thrown/rejected mirror is caught, logged, and cannot affect the return value or the primary `ai_tool_executions` decide, which happens earlier in the function. Confirmed by the new `aiAgent.authz.test.ts` case ("still reports success when the bridge mirror fails").

4. **RLS/tenant-isolation context**: `recordDroppedToolResult`'s `withDbAccessContext({ scope: 'organization', orgId: session.orgId, accessibleOrgIds: [session.orgId] })` for both the `ai_messages` insert and the `ai_sessions` flag update is byte-for-byte the same pattern `createSessionPostToolUse` already uses for its own `ai_messages` insert (aiAgentSdk.ts ~line 1136) and that `aiAgentSdk.ts` documents as required for any DB write from inside the background processor (which runs outside the request's AsyncLocalStorage context). No inconsistency.

5. **Test coverage**: `toolUseNames` is optional (`Map<string, string> | undefined`) and every access is `?.`-guarded, so `ActiveSession` literals in other test files that don't set it still compile and degrade to `'unknown_tool'` rather than throwing — matches the doc comment's stated contract. The 4 new tests cover: genuine drop, already-handled (no double-record), plain-text resume replay (ignored), and a drop followed by a retry (queue hygiene). Minor gap: no test for two *different* tool calls both dropped back-to-back — not exercised, but the `indexOf`/`splice`-per-id logic handles it by construction (each id is independent), so this is a nice-to-have, not a blocker.

## Test evidence
- `streamingSessionManager.droppedToolResult.test.ts` — 4/4 passed
- `aiAgent.authz.test.ts` — 14/14 passed
- `aiToolSchemas.validateToolInput.test.ts` — 9/9 passed
- Broader regression (`aiAgentSdk.test.ts`, `aiAgentSdk.m365risk.test.ts`, `aiAgentSdk.planMatch.test.ts`, `clientAiTools.contract.test.ts`, `clientAiTools.handler.test.ts`, `clientAiTools.registry.test.ts`, `streamingSessionManager.clientLoop.test.ts`, `streamingSessionManager.security.test.ts`, `aiAgent.test.ts`) — 165/165 passed
- `tsc --noEmit` (`NODE_OPTIONS="--max-old-space-size=8192"`) — clean

Final commit: 02bab33eb

PR #3104, opened 2026-08-04
fix(ai): stop the chat turn from blocking on the slowest pending approval (#3089)

ToddHebebrand:
Closes #3089

## Root cause

`createSessionPreToolUse` blocks each approval-gated tool call inside its MCP tool handler for up to 5 minutes (`waitForIntentDecision` / `waitForApproval`), and the Agent SDK executes sibling tool calls sequentially. Consequences observed in the flagged US-prod sessions:

- A completed sibling tool result was withheld behind a pending `execute_command` approval.
- Two pending approvals stacked their waits to 10 minutes — past the 6-minute `SDK_TURN_TIMEOUT_MS` — so the UI received `error` + `done` first and the model's eventual closing message was published into a turn the client had already ended (sessions ending with raw approval-pending error rows and **no closing assistant message**).
- A user message sent mid-wait bounced off the 409 concurrent-message guard in `routes/ai.ts`, so the assistant was mute while the user typed "i think it got stuck".

## Fix

1. **Shared per-cycle approval-wait budget** (`APPROVAL_WAIT_BUDGET_MS` = 5 min, `beginApprovalWait` in `aiAgentSdk.ts`): all approval waits in one assistant cycle draw from a single budget, so cumulative blocking always ends before the turn timeout and the model gets the tool errors back in time to conclude the turn gracefully. The budget resets at each `message_start` via `startTurnTimeout` — the same point the 6-min turn timeout resets, preserving the headroom invariant.
2. **`settleApprovalWaits()`**: a new user message that hits the concurrent-message guard now settles in-flight approval waits, waits briefly (`waitForTurnToSettle`, 15s) for the turn to conclude, then proceeds normally — the assistant answers instead of 409ing. Tier-3 intents stay `pending_approval` and are **still executed by the durable release worker once an approver decides** (spec §6.1) — no approval capability is lost. The interrupt endpoint settles too, so Stop works while blocked on an approval.
3. **Tier-2 legacy bridge hardening**: a wait that ends without a decision closes out its still-pending `ai_tool_executions` row (guarded on `status='pending'`), so a later Approve click can't flip it to a stranded `approved` that the legacy bridge (which has no durable worker) would never execute.

Not changed: the approval gate itself (nothing executes without approval), the helper/PAM branch, and the tier-3 expiry error message at the `pending_approval` return — that message is #3090's scope and will be fixed separately.

## Test evidence

[...]

PR #3110, opened 2026-08-04
fix(ai): approval-expiry tool error no longer claims the action will complete (#3090)

### LINKED ISSUE #3090: [AI] Approval-expiry tool error is misleading: says 'will complete once approved' after the request has already expired
## Description
When an approval-gated tool call goes unapproved, the tool result eventually delivered to the model (and shown in flag reasons) is:

> `Approval still pending; this action will complete once approved.`

(`apps/api/src/services/aiAgentSdk.ts:736`)

Both claims are wrong by the time the user sees it. On US prod (2026-08-03), these errors were delivered 5–10 minutes after the call — *after* the expiry sweeper had already marked the `approval_requests` rows `expired` (5-minute TTL). The action will **never** complete: approving later does nothing, and the tool result has already returned as an error. The model then relays "approve the pending commands in the approval queue" to the user — advice that cannot work.

## Proposed Fix
- On expiry, return promptly (at expiry time, not minutes later) with an honest message: the approval expired, the action was not run, and the model should re-issue the call if still needed.
- Distinguish `pending` (still actionable — tell the user where to approve) from `expired` (re-issue) in the error text so the model gives correct guidance.

## Affected Files
- `apps/api/src/services/aiAgentSdk.ts:736` and the approval wait/expiry path
- Expiry sweeper for `approval_requests`

## Reported By

ToddHebebrand:
Closes #3090

## Root cause

When a tier-3 (durable action-intent) tool call is blocked awaiting approval, `createSessionPreToolUse` polls `action_intents.status` via `waitForIntentDecision` for up to 5 minutes. On give-up, if the last-observed status is still `pending_approval`, the tool result delivered to the model says:

> `Approval still pending; this action will complete once approved.`

Both claims can be false by the time this fires. The local wait's own timeout is calibrated to roughly the *same* duration as the intent's `expires_at`, but the DB row is only flipped `pending_approval -> expired` by `jobs/intentExpiryReaper.ts` on a 30-second sweep — so the wait can give up and still read a stale `pending_approval` row moments (or longer) after the intent's real deadline has already passed. On US prod (2026-08-03), these tool errors were observed 5–10 minutes after the call, after the reaper had already marked the linked `approval_requests` rows `expired` — yet the model was told "still pending," and relayed unactionable advice ("approve the pending commands") to the user.

## Fix

`apps/api/src/services/aiAgentSdk.ts` — when `waitForIntentDecision` gives up with status still `pending_approval`, compare `Date.now()` against the already-in-scope `intent.expiresAt` before choosing the message:

- **Past the deadline** → new error: *"Approval request expired before a decision was made; the action was not executed. Re-issue the tool call if it is still needed."*
- **Still within the window** (e.g. a sibling wait in the same cycle exhausted its share of the shared budget, or an early settle) → unchanged: *"Approval still pending; this action will complete once approved."* — accurate here, since the durable release worker really will execute it once an approver decides.

No change to the reaper cadence, the wait/timeout mechanics, or the tier-2 legacy bridge (`waitForApproval`) — that path already self-closes the row synchronously on timeout and isn't affected by this race.

## Overlap with PR #3104 (#3089)

PR #3104 (open, `fix/3089-approval-turn-block`) refactors the surrounding wait construction on the same lines — it wraps the `waitForIntentDecision(...)` call in `beginApprovalWait(session)` / `try/finally` for its shared per-cycle wait budget. It does **not** touch the `if (decisionStatus === 'pending_approval') { ... }` message body itself (confirmed via `gh pr diff 3104`), so there's no line-for-line textual conflict, but this PR's edit starts on the very next line after #3104's changed construction — close enough that a rebase will likely need a manual hunk merge (not a real logic conflict; both hunks are additive to the same small block).

**Recommend landing #3104 first, then rebasing this PR** — after rebase, `decisionStatus` comes from the shared-budget wrapper instead of the flat `300_000` timeout, but the `intent.expiresAt` comparison added here composes with that unchanged (the deadline check is orthogonal to how the wait's own local timeout was computed).

## Test evidence

- `apps/api/src/services/aiAgentSdk.test.ts` — 2 new tests: the existing "still pending" case (default future `expiresAt`) is untouched; new test `reports the approval as expired — not "still pending" — once the intent deadline has passed` (past `expiresAt`) asserts the new message and that `transitionIntent` is still never called (durable, no mutation on give-up, same as before). Full file: **96/96 passed**, single-fork.
- `apps/api/src/services/aiAgentSdk.planMatch.test.ts`, `aiAgentSdk.m365risk.test.ts` — **15/15 passed**, single-fork (same code path, unaffected).

### DIFF
--- a/apps/api/src/services/aiAgentSdk.test.ts
+++ b/apps/api/src/services/aiAgentSdk.test.ts
@@ -799,6 +799,45 @@ describe('createSessionPreToolUse', () => {
       expect(mockTransitionIntent).not.toHaveBeenCalled();
     });
 
+    // #3090: waitForIntentDecision can still read `pending_approval` from a
+    // stale DB row after the intent's own `expiresAt` has already passed —
+    // jobs/intentExpiryReaper.ts flips the row to `expired` on a 30s sweep,
+    // and the chat wait's own local timeout races that sweep by design. The
+    // tool result must say "expired", never "still pending", once wall-clock
+    // time is past the intent's deadline — the old message was false on both
+    // counts (not pending, and never completing on a later approval).
+    it('reports the approval as expired — not "still pending" — once the intent deadline has passed', async () => {
+      vi.mocked(checkGuardrails).mockReturnValue({
+        allowed: true,
+        tier: 3,
+        requiresApproval: true,
+        description: 'Execute command',
+      } as any);
+      mockInsertReturning({ id: 'exec-5-expired' });
+      mockCreateActionIntent.mockResolvedValue(
+        makeIntentSnapshot({
+          id: 'intent-5-expired',
+          approvalRequestIds: ['appr-5-expired'],
+          // Deadline already in the past — the sweep just hasn't caught up yet.
+          expiresAt: new Date(Date.now() - 1_000),
+        }),
+      );
+      mockWaitForIntentDecision.mockResolvedValue('pending_approval');
+      const session = makeActiveSession({ approvalMode: 'per_step' });
+
+      const result = await createSessionPreToolUse(session)('execute_command', {});
+
+      expect(result).toEqual({
+        allowed: false,
+        error:
+          'Approval request expired before a decision was made; the action was not executed. Re-issue the tool call if it is still needed.',
+      });
+      // Same durable-no-mutation contract as the still-pending case: giving
+      // up here must not touch the intent — an approver deciding it late (or
+      // the reaper's own sweep) is what actually resolves the row.
+      expect(mockTransitionIntent).not.toHaveBeenCalled();
+    });
+
     it('CASes the intent executing -> completed once the inline tool call finishes successfully', async () => {
       vi.mocked(checkGuardrails).mockReturnValue({
         allowed: true,
--- a/apps/api/src/services/aiAgentSdk.ts
+++ b/apps/api/src/services/aiAgentSdk.ts
@@ -730,6 +730,26 @@ export function createSessionPreToolUse(session: ActiveSession): PreToolUseCallb
           );
 
           if (decisionStatus === 'pending_approval') {
+            // The row can still READ `pending_approval` here even though the
+            // intent's own deadline has already passed: `jobs/intentExpiryReaper.ts`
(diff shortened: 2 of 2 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer, silent-failure-hunter, pr-test-analyzer (parallel; code-reviewer re-run once after a transient dispatch failure lost its first attempt before it read anything).

**Findings:**
- code-reviewer: 0 confirmed issues. Verified `intent.expiresAt` is a non-nullable `Date` at every producer (`actionIntents/intentService.ts:91`), and confirmed expiry is enforced independently at three layers (decide-handler 410 check, release-worker `requireNotExpired` CAS, the 30s reaper) — so "the action was not executed" in the new message is guaranteed true regardless of which layer actually caught the expiry. Noted (not a defect) that on the flat 300s timeout path the new "expired" branch will fire on nearly every genuine timeout, since `CHAT_EXPIRY_MS` is also 300s and `expiresAt` is stamped before the wait begins — that's the intended fix, not a side effect.
- silent-failure-hunter: 0 issues. The new branch only selects between two accurate error strings inside the existing `pending_approval` case; no new try/catch, I/O, or fallback logic; no new throw risk (`expiresAt` already dereferenced earlier at `aiAgentSdk.ts:711` for the SSE payload).
- pr-test-analyzer: 1 minor, non-blocking suggestion — the new expired-deadline branch is only exercised in `per_step` mode; the sibling `action_plan` timeout test (asserting the "The plan has been stopped." suffix) has no expired-deadline counterpart. Low risk (identical `failMatchedPlanStep` call/suffix as the already-tested adjacent branch), not applied — flagging for a follow-up if desired rather than expanding this PR's scope.

**Tests:** `apps/api/src/services/aiAgentSdk.test.ts` 96/96 passed (single-fork); `aiAgentSdk.planMatch.test.ts` + `aiAgentSdk.m365risk.test.ts` 15/15 passed (single-fork); `tsc --noEmit` on `apps/api` clean.

**Status:** review-clean (0 confirmed findings across both passes), awaiting maintainer merge. See PR body for the expected overlap/rebase note vs PR #3104.

Issue #3115, opened 2026-08-04
[Mobile] SessionsSheet renders raw internal error text to users


ToddHebebrand:
## Description

When loading conversation history fails, the History sheet displays the raw internal error string verbatim in red:

```
listAiSessions failed: 401
```

That is a developer-facing message — a function name and an HTTP status code — shown directly as user-facing UI. It tells the user nothing actionable and leaks internal naming.

Observed on Android against a live server (see #3114 for the underlying 401), but the leak is platform-independent: any failure of this call renders its `Error.message` on screen.

## Root Cause

- The string is constructed at `apps/mobile/src/services/aiChat.ts:76`:
  `throw new Error(\`listAiSessions failed: ${res.status}\`)`
- `apps/mobile/src/screens/chat/components/SessionsSheet.tsx:25` catches it and stores the raw message: `.catch((err: Error) => setError(err.message))`
- `SessionsSheet.tsx:80` renders it verbatim: `<Text ...>{error}</Text>`

This diverges from the established convention in the same directory. `SettingsSheet.tsx:149-151,176` shows a short, static, user-facing string (e.g. `'Could not revoke device.'`) and does not surface internal error text.

## Proposed Fix

Map the failure to a static user-facing message (e.g. "Couldn't load conversation history.") using the same toast/Alert pattern as `SettingsSheet.tsx`, and report `err.message` to Sentry rather than to the screen.

## Affected Files

- `apps/mobile/src/screens/chat/components/SessionsSheet.tsx` (primary — lines 25, 80)
- `apps/mobile/src/services/aiChat.ts:76` (error construction)
- `apps/mobile/src/screens/chat/components/SettingsSheet.tsx` (reference pattern)

## Related

- #3114 — the 401 that surfaced this

PR #3124, opened 2026-08-04
fix(mobile): refresh expired access token on 401 in aiChat authedFetch (#3114)

### LINKED ISSUE #3114: [Mobile] Chat history 401s ~15 minutes after login — aiChat.ts never refreshes the access token
## Description

Tapping the history/clock icon in the chat header shows an error instead of past conversations, but only once the session is more than ~15 minutes old. Immediately after signing in it works.

Reproduced on Android (emulator, API 37, debug build) against a live server:

1. Sign in.
2. Send a chat message — succeeds, streaming and tool calls all work.
3. Wait until >15 minutes have elapsed **since login**.
4. Tap the history icon.

The sheet renders `listAiSessions failed: 401`.

Observed timeline in the repro: login 14:26, chat 14:29 (worked), history 14:45 — 19 minutes after login.

## Root Cause

Access tokens are 15 minutes (`ACCESS_TOKEN_EXPIRY = '15m'`, `apps/api/src/services/jwt.ts:5`).

`aiChat.ts` has its **own** fetch wrapper, `authedFetch` (`apps/mobile/src/services/aiChat.ts:30-50`), which reads whatever token is currently in SecureStore and attaches it — with no expiry check, no refresh, and no retry. `apps/mobile/src/services/api.ts:392` already implements a `refreshToken()` helper, but `aiChat.ts` is a separate implementation that never calls it.

This is not a scope/permissions problem. `GET /api/v1/ai/sessions` (`apps/api/src/routes/ai.ts:180-197`) uses the same `requireScope('organization','partner','system')` as the chat routes, and a partner-scoped token resolves its orgs fine. The 401 comes from `authMiddleware` rejecting an expired token at `apps/api/src/middleware/auth.ts:443`.

The reason chat "works" and history doesn't is timing, not authorization: the chat SSE POST validates its bearer once at stream open and then holds the connection for the whole tool-calling turn, while the later `GET /ai/sessions` re-validates fresh.

There is prior art for exactly this failure class in the repo — see the viewer/VNC comment at `apps/api/src/services/jwt.ts:7-13` describing a 15m access token turning into silent 401s that killed a poll loop.

## Proposed Fix

Give `aiChat.ts`'s `authedFetch` a 401-triggered refresh-and-retry-once, reusing `refreshToken()` from `apps/mobile/src/services/api.ts`, so the AI chat service follows the same token lifecycle as the other mobile services.

## Affected Files

- `apps/mobile/src/services/aiChat.ts` (primary — `authedFetch`, lines 30-50)
- `apps/mobile/src/services/api.ts` (has the existing `refreshToken()` to reuse)

## Notes

The user-visible error text is a separate defect, filed alongside this one.

ToddHebebrand:
Closes #3114

## Problem

`aiChat.ts` has its own fetch wrapper, `authedFetch`, which attaches whatever token is in SecureStore with no expiry check, refresh, or retry. Access tokens live 15 minutes (`ACCESS_TOKEN_EXPIRY`, `apps/api/src/services/jwt.ts`), so any AI-chat REST call made >15m after login 401s — user-visible as `listAiSessions failed: 401` when opening chat history.

## Fix

- `authedFetch` now retries **once** on 401: it refreshes via the existing `refreshToken()` helper (`apps/mobile/src/services/api.ts`), persists the new token with `storeToken()` (`auth.ts`) so every service reading `breeze_auth_token` benefits, and re-issues the request with the fresh bearer.
- Concurrent 401s single-flight into one `/auth/refresh` call.
- If the refresh itself fails (refresh cookie expired → session genuinely over), the **original 401** is returned so existing auth handling fires.
- A `storeToken` persistence failure (locked keychain) does not discard the usable in-memory token for the retry.

## Scope notes

- **Stacked PR** — based on `ToddHebebrand/ios-app-testing` (the branch this issue was diagnosed against), so `ci.yml` will NOT run (it only triggers on PRs targeting `main`). Verification was run locally; see below.
- Sibling issue #3115 (the `listAiSessions` error copy / SessionsSheet UX) is being fixed in parallel — this diff deliberately does not touch error messages or SessionsSheet.
- Known residual: `streamChat` (XHR SSE path) still attaches the stored token without refresh; a chat message sent >15m after login with no interim refresh can still 401 at stream open. Kept out of scope per the issue, which targets `authedFetch`.

## Verification (local)

- New unit suite `apps/mobile/src/services/aiChat.test.ts` — 6 tests: no refresh on success, refresh+persist+retry with new bearer, original 401 surfaced when refresh fails, retry-once (second 401 not re-refreshed), retry proceeds when persistence fails, concurrent 401s single-flighted.
- Full mobile suite: 28 files, 267 passed / 3 skipped.
- `tsc --noEmit` (apps/mobile): clean.

### DIFF
--- a/apps/mobile/src/services/aiChat.test.ts
+++ b/apps/mobile/src/services/aiChat.test.ts
@@ -1,12 +1,17 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { listAiSessions } from './aiChat';
+import { createAiSession, listAiSessions } from './aiChat';
 import { refreshToken } from './api';
 import { storeToken } from './auth';
 import { fetchWithTimeout } from './fetchWithTimeout';
 
 vi.mock('./serverConfig', () => ({
   getServerUrl: vi.fn().mockResolvedValue('https://api.test'),
+}));
+
+vi.mock('@sentry/react-native', () => ({
+  captureException: vi.fn(),
+  captureMessage: vi.fn(),
 }));
 
 vi.mock('expo-secure-store', () => ({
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Findings:** 5 actionable raised → 4 addressed in c242a8b91 (Sentry capture on both refresh-failure swallow paths, corrected the misleading catch comments, single-flight future-caller warning, POST retry-fidelity test, explicit guard-reset-after-failure test); 1 deliberately deferred — `streamChat` (XHR SSE send path) still attaches the stored token without refresh, so a first message sent >15m after login with no interim REST call can still 401 at stream open. Kept out of this diff per the issue scope (#3114 targets `authedFetch`) and to avoid colliding with the parallel #3115 work in this file; flagged for a follow-up.
**Not addressed (pre-existing infra):** no CI job runs the `breeze-mobile` vitest suite at all — this PR's tests (and the 28 existing mobile test files) only run locally. Worth a separate `test-mobile` CI issue.
**Tests:** apps/mobile full suite green (28 files, 269 passed / 3 skipped), incl. 8 new aiChat tests; `tsc --noEmit` clean. Stacked PR (base `ToddHebebrand/ios-app-testing`) so `ci.yml` does not trigger — verification is local by design; on-device confirmation of the live `/auth/refresh` cookie round-trip is planned by the operator.
**Status:** review-clean on addressed findings, awaiting maintainer merge.

PR #3125, opened 2026-08-04
fix(mobile): stop Settings asserting push works when registration is unsupported (#3118)

### LINKED ISSUE #3118: [Mobile] Android: Settings shows Notifications enabled although push can never register
## Description

On Android, the settings sheet shows **Notifications — "Approval pushes and alerts"** with the toggle **ON**, and a "Critical only" sub-toggle beneath it. But push registration on Android never succeeds, so no push can ever be delivered. The UI asserts a capability the platform build does not have.

The same sheet shows **"THIS PHONE + OTHERS — No paired devices yet. Sign in on another phone to see it here."** while the current phone is signed in — the downstream symptom of the same gap, since the device never registers a token and so never lists itself.

## Evidence

`apps/mobile/src/services/notifications.ts:54-66` deliberately returns `{ status: 'unsupported', reason: 'android_push_not_configured' }` on Android when no `extra.eas.projectId` is present. That is an intentional, documented "not built yet" state — this issue is **not** asking for Android push to be implemented.

Independently confirmed server-side: seeding an approval returns

```json
"push": { "tokensFound": 0, "dispatched": 0, "errors": 0 }
```

so no token exists for the signed-in user on this device.

Runtime log on Android also shows: `Push notifications require a physical device` (emulator), but the `unsupported` path applies on real Android hardware too.

## Proposed Fix

Have the settings sheet read the push registration status rather than assuming success:

- When registration returned `unsupported` / `android_push_not_configured`, render the Notifications row in a disabled/explanatory state ("Push notifications aren't available on Android yet") instead of an ON toggle.
- Consider the same treatment for the paired-devices section so it doesn't read as an error when the device simply never registered.

## Affected Files

- `apps/mobile/src/screens/chat/components/SettingsSheet.tsx` (primary — notifications + paired devices sections)
- `apps/mobile/src/services/notifications.ts:54-66` (already returns the status to branch on)

## Impact

An Android user believes approval pushes are enabled and will wait for a notification that cannot arrive. Approvals are the app's headline flow, so the false assurance is the costly part — not the missing feature.

ToddHebebrand:
## Summary

Fixes the Android Settings sheet asserting a push capability the build does not have. `registerForPushNotifications()` deliberately returns `{ status: 'unsupported', reason: 'android_push_not_configured' }` on Android (intentional not-built-yet state), but `SettingsSheet` showed the Notifications toggle ON with a "Critical only" sub-toggle, and the paired-devices empty state ("Sign in on another phone to see it here") read as an error — this phone can never register a token, so it never lists itself.

## Changes

- `SettingsSheet` now reads `auth.pushRegistration` / `pushRegistrationReason` from the store (already populated by `PushRegistrationGate` in `App.tsx`).
- When status is `unsupported`, the Notifications row renders in a disabled/explanatory state: dimmed, Switch OFF and disabled, description replaced with reason-specific copy ("Push notifications aren't available on Android yet."). The Critical-only sub-toggle is hidden.
- The paired-devices empty hint gets matching copy explaining the phone can't register itself, so the empty state no longer reads as an error.
- `failed` status keeps the normal toggle — `ApprovalGate` already surfaces registration failures with its own banner; this change only covers the deliberate `unsupported` state.
- Copy mapping lives in a pure module `pushUnavailableCopy.ts` (same precedent as `approverBannerCopy.ts`) with unit tests covering `android_push_not_configured`, `not_physical_device`, and the generic fallback.

This does NOT implement Android push — the issue explicitly scopes that out.

## Stacked PR — CI will not run

This PR is based on `ToddHebebrand/ios-app-testing` (the issue was diagnosed against its 4 unmerged mobile commits), so `ci.yml` will NOT trigger (it only runs on PRs targeting `main`). Verification was run locally instead:

- `npx tsc --noEmit` in `apps/mobile`: clean
- `pnpm exec vitest run` in `apps/mobile`: 28 files, 265 passed / 3 skipped (includes new `pushUnavailableCopy.test.ts`)

Closes #3118

### DIFF
--- a/apps/mobile/src/screens/chat/components/SettingsSheet.tsx
+++ b/apps/mobile/src/screens/chat/components/SettingsSheet.tsx
@@ -757,10 +757,12 @@ function ToggleRow({
         paddingVertical: spacing[3],
         flexDirection: 'row',
         alignItems: 'center',
-        opacity: disabled ? 0.55 : 1,
       }}
     >
       <View style={{ flex: 1, marginRight: spacing[3] }}>
+        {/* Only the label and Switch signal "disabled" — the description stays
+            at full contrast because for a disabled row it carries the
+            explanation the user must be able to read (#3118). */}
         <Text style={[type.bodyMd, { color: disabled ? theme.textLo : theme.textHi }]}>{label}</Text>
         {description ? (
           <Text style={[type.meta, { color: theme.textMd, marginTop: spacing[1] }]}>
@@ -774,6 +776,7 @@ function ToggleRow({
         onValueChange={onChange}
         trackColor={{ false: theme.bg3, true: palette.brand.deep }}
         thumbColor={value ? palette.brand.base : theme.textMd}
+        style={disabled ? { opacity: 0.55 } : undefined}
       />
     </View>
   );
--- a/apps/mobile/src/screens/chat/components/SettingsSheet.tsx
+++ b/apps/mobile/src/screens/chat/components/SettingsSheet.tsx
@@ -43,6 +43,7 @@ import { ease, duration } from '../../../lib/motion';
 import { track } from '../../../lib/analytics';
 import { relativeTime } from '../../../lib/relativeTime';
 import { Toast } from '../../../components/Toast';
+import { pushUnavailableCopy } from './pushUnavailableCopy';
 import { Avatar } from './Avatar';
 import { ChangePasswordSheet } from './ChangePasswordSheet';
 import type { PairedMobileDevice } from '../../../services/mobileDevices';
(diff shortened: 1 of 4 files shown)


ToddHebebrand:

**Findings:** 3 raised, 2 addressed in 62833491d — (1) row-level opacity dimmed the explanatory copy below AA contrast, now only the label + Switch are dimmed; (2) the reason-string contract between `registerForPushNotifications` and `pushUnavailableCopy` was unpinned, now a test feeds the service's real `unsupported` reasons into the copy mapper and asserts non-generic copy. (3) Not addressed here: no `test-mobile` job exists in `ci.yml`, so the mobile vitest suite (28 files) never runs in CI — pre-existing repo gap, flagged as follow-up rather than expanding a stacked PR.

**Tests:** `apps/mobile` typecheck clean (`tsc --noEmit`); `vitest run` 28 files, 266 passed / 3 skipped (includes the new `pushUnavailableCopy` and contract-pin tests). Stacked on `ToddHebebrand/ios-app-testing`, so `ci.yml` will not trigger — verification is local only.

**Follow-ups worth issues (pre-existing, out of scope):** the `failed` push banner in ApprovalGate is dismissible per session, after which the ON toggle is the only (wrong) signal; and both Notifications toggles are write-only — nothing consumes `notificationsEnabled`/`notificationsCriticalOnly`, so on iOS the toggle asserts control it doesn't have (same defect class as #3118).

**Status:** review-clean, awaiting maintainer merge. On-device Android verification to be done by the orchestrator.

PR #3131, opened 2026-08-04
fix(mobile): render AI chat markdown tables as stacked labelled rows (#3119)

### LINKED ISSUE #3119: [Mobile] AI markdown tables collapse to one cell per line
## Description

When an AI chat response contains a markdown table, it does not render as a table. Every cell is placed on its own line, so the header renders as three stacked lines and each row's values stack beneath it, with no column alignment and no repeated labels.

Observed on Android (emulator, API 37, debug build) asking "What broke last night?", where the response returned a device table:

```
Device

OS

Last Seen

Domain Controller (RVW-DC-01)

Windows Server

July 11, 2:03 AM

Reception PC (RVW-WIN-01)

Windows

July 11, 2:02 AM
...
```

Expected: three aligned columns, or — if stacking is deliberate for narrow screens — labelled pairs (`OS: Windows Server`) so each value is identifiable without scrolling back to the header.

The content is technically readable, but a multi-row table becomes very long and the association between a value and its column is lost after the first row scrolls past.

## Root Cause (not confirmed)

Rendering goes through `react-native-markdown-display` (`apps/mobile/package.json`). Either its table rules aren't wired into the app's markdown style overrides, or the table styles are being overridden such that cells lay out as block rows. Needs a look at the markdown component's `rules`/`style` config before deciding the fix.

## Proposed Fix

Either:
1. Supply proper `table` / `thead` / `tbody` / `tr` / `th` / `td` styles so columns lay out horizontally (with the table in a horizontally scrollable container for narrow screens), or
2. Deliberately implement a stacked/labelled renderer for tables, repeating the column label with each value.

Option 2 is likely the better mobile experience for wide tables, but should be a deliberate choice rather than the current accidental collapse.

## Affected Files

- The markdown renderer/styles used by the chat message component under `apps/mobile/src/screens/chat/`
- `react-native-markdown-display` config

## Impact

Cosmetic but recurring — asking about fleet state is a primary use of the assistant, and tabular answers are common.

ToddHebebrand:
Closes #3119

## Root cause

`MarkdownBody.tsx` passes `mergeStyle={false}` to `react-native-markdown-display`, which discards the library's **default layout styles** — including `tr: { flexDirection: 'row' }` and `th`/`td: { flex: 1 }`. The app's own `table`/`tr`/`th`/`td` style overrides were purely decorative (borders, padding, fonts), so table cells fell back to React Native's default column stacking: every cell on its own full-width line, exactly as reported. The table rules were wired in all along; the layout styles were dropped.

## Fix — deliberate stacked/labelled renderer (option 2)

The issue offered two directions. This PR implements **option 2** (stacked rows with the column label repeated per value), as a reasoned choice:

- The observed tables are wide fleet tables (`Device | OS | Last Seen`) with long values. Restoring the grid (`flex: 1` columns) squeezes 3+ columns into unreadable wrapped slivers at phone widths, and a horizontally-scrollable table nested inside the vertical chat scroll both fights the scroll gesture and puts the header/columns off-screen — recreating the value↔column association loss the issue complains about, just in another form.
- The chat already renders structured data as stacked block cards (`DeviceCard`, `FleetStatusRow`), so labelled stacked rows match the existing visual language.
- The issue author leaned this way; the accidental collapse becomes a deliberate design.

Each table renders as a bordered card (theme tokens: `border`, `radii.md`), one block per row separated by hairline dividers, each value on a line prefixed with its muted semibold column label: `OS: Windows Server`.

## Implementation

- `apps/mobile/src/screens/chat/components/markdownTable.ts` — new pure helper (no React/RN imports, so the node-env Vitest config can cover it): parses the `table` AST node into `{ labels, rows }`. Discovery of `tr`/`th`/`td` is recursive, so `textgroup` wrappers and missing `thead`/`tbody` don't break it; cell text is flattened from inline markup (bold/code kept as text, breaks as spaces, image alt text). Edge handling: ragged rows kept as-is, extra header rows demoted to body rows rather than dropped, header-only tables render the header cells as a plain row, headerless tables render values without labels.
- `MarkdownBody.tsx` — custom `table` render rule using the helper; the now-dead decorative `table`/`thead`/`th`/`td`/`tr` styles are removed.

## Verification

- `pnpm exec vitest run` (apps/mobile): **28 files, 272 passed, 3 skipped** — includes 11 new tests for the table parser.
- `pnpm run typecheck` (apps/mobile, `tsc --noEmit`): clean.
- **Visual rendering has NOT been verified on a device/emulator** — the mobile app has no RN component test runtime, and I deliberately did not touch the shared dev stack/emulator. On-device verification (Android emulator, AI chat with a table-bearing response) should confirm the stacked card layout before merge.

## Stacked PR note

This PR is **based on `ToddHebebrand/ios-app-testing`** (not `main`) because the issue was diagnosed against its 4 unmerged mobile commits. `ci.yml` only triggers on PRs targeting `main`, so **no CI runs here** — verification above was run locally. Dispatch `gh workflow run CI --ref fix/3119-mobile-markdown-tables` if a CI record is wanted before merge.

### DIFF
--- a/apps/mobile/src/screens/chat/components/MarkdownBody.tsx
+++ b/apps/mobile/src/screens/chat/components/MarkdownBody.tsx
@@ -3,7 +3,7 @@ import { Text, View } from 'react-native';
 import Markdown, { type RenderRules } from 'react-native-markdown-display';
 
 import { useApprovalTheme, fontFamily, radii, spacing } from '../../../theme';
-import { parseMarkdownTable } from './markdownTable';
+import { parseMarkdownTable, toDisplayRows } from './markdownTable';
 
 interface Props {
   content: string;
@@ -208,17 +208,14 @@ export function MarkdownBody({ content }: Props) {
     () => ({
       hr: () => null,
       table: (node) => {
-        const { labels, rows } = parseMarkdownTable(node);
-        // Header-only table: show the header cells as a plain row rather
-        // than dropping the content (no labels to pair them with).
-        const bodyRows = rows.length > 0 ? rows : labels.length > 0 ? [labels] : [];
-        if (bodyRows.length === 0) {
+        const parsed = parseMarkdownTable(node);
+        const { rows, showLabels } = toDisplayRows(parsed);
+        if (rows.length === 0) {
           return null;
         }
-        const showLabels = rows.length > 0;
         return (
           <View key={node.key} style={tableStyles.card}>
-            {bodyRows.map((cells, rowIndex) => (
+            {rows.map((cells, rowIndex) => (
               <View
                 key={rowIndex}
                 style={
@@ -226,11 +223,15 @@ export function MarkdownBody({ content }: Props) {
                 }
               >
                 {cells.map((value, cellIndex) => {
-                  const label = showLabels ? labels[cellIndex] : undefined;
+                  const label = showLabels ? parsed.labels[cellIndex] : undefined;
+                  if (!label && !value) {
+                    return null;
+                  }
                   return (
                     <Text key={cellIndex} style={tableStyles.line}>
                       {label ? <Text style={tableStyles.label}>{label}: </Text> : null}
-                      {value}
+                      {/* Em dash keeps a labelled blank cell legible ("OS: —"). */}
+                      {value || '—'}
                     </Text>
                   );
                 })}
--- a/apps/mobile/src/screens/chat/components/MarkdownBody.tsx
+++ b/apps/mobile/src/screens/chat/components/MarkdownBody.tsx
@@ -1,7 +1,9 @@
 import { useMemo } from 'react';
+import { Text, View } from 'react-native';
 import Markdown, { type RenderRules } from 'react-native-markdown-display';
 
 import { useApprovalTheme, fontFamily, radii, spacing } from '../../../theme';
+import { parseMarkdownTable } from './markdownTable';
 
 interface Props {
   content: string;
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (pr-review-toolkit:code-reviewer, pr-review-toolkit:pr-test-analyzer)

**Findings:** 0 critical, 0 important. 3 suggestions taken in 50c481e73: empty labelled cells now render `Label: —` (fully empty cells are skipped), the header-only/headerless display branching moved into the pure `toDisplayRows()` helper so it sits behind unit tests, and the lossy link/image flattening inside cells is documented. Declined: pre-existing library behavior of rendering the discarded table subtree (low impact, out of scope). The code-reviewer independently confirmed the root cause against the library source (`mergeStyle !== true` wholly replaces per-key defaults) and the test-analyzer executed the real `react-native-markdown-display@7.0.2` token pipeline against the parser — 5/5 fidelity probes passed.

**Tests:** apps/mobile Vitest suite green — 28 files, 276 passed, 3 skipped (15 new for the table parser); `tsc --noEmit` clean. No CI runs on this PR (stacked on `ToddHebebrand/ios-app-testing`; `ci.yml` only triggers on PRs targeting `main`) — verification was local. Note: CI has no mobile test job at all, so these tests are local-only either way; also noted by review as a follow-up-worthy gap.

**Status:** review-clean. Visual layout NOT verified on-device — needs the Android emulator pass (AI chat, table-bearing response) before merge.

PR #3138, opened 2026-08-04
fix(deps): bump fast-uri override past 3.1.4 for CVE-2026-18446 (main is red)

bdunncompany:
## Main is red

`Security Scanning` is failing on main (`405d4177c`) — both `Trivy Filesystem Scan` and `Trivy Image Scan`:

```
│ fast-uri │ CVE-2026-18446 │ HIGH │ fixed │ 3.1.4 │ 2.4.4, 3.1.5, 4.1.2 │
│ fast-uri: Host confusion vulnerability via backslash in URI authority │
```

Newly published advisory against a version already pinned in-tree — not a regression from any recent merge, and it will keep failing every run until the pin moves.

## The fix

The existing override was:

```json
"fast-uri@>=3.0.0 <3.1.4": "3.1.4",
```

It pins to exactly the version now flagged, so the range no longer binds anything useful. Widened to `<3.1.5`, pinned to `3.1.5`.

**Pinned to `3.1.5`, not `>=3.1.5`.** The open range resolves to `4.1.2` — a major bump for a transitive dependency, which is more change than a CVE fix warrants. `3.1.5` is one of the three fixed versions the advisory lists and stays on the 3.x line. Happy to take 4.x instead if you'd rather move forward, but that felt like a separate decision from unblocking CI.

## Verification

Trivy 0.70.0 (the version CI pins), via OrbStack, run **both ways**:

| lockfile | result |
|---|---|
| `origin/main` (control) | **Total: 1 (HIGH: 1)** — `fast-uri 3.1.4` flagged |
| this branch | **0 vulnerabilities** |

The control run is the point — it proves the scan can still fail on this input and that this change is what clears it, rather than the scan passing for some unrelated reason.

Also green:

- `node scripts/security/check-override-floors.mjs` — *"every override still binds — no stale floors, no dead pins"* across 54 entries
- `node --test scripts/security/check-override-floors.test.mjs` — 32/32

Lockfile-only otherwise: `fast-uri@3.1.5` is the only resolved version afterwards, and no other package moved (`pnpm-lock.yaml` diff is 6 insertions / 8 deletions, all fast-uri).

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
       "fast-xml-parser@<5.5.7": ">=5.5.7",
       "fast-xml-parser@<5.7.0": ">=5.7.0",
       "fast-xml-parser@<5.10.1": ">=5.10.1",
-      "fast-uri@>=3.0.0 <3.1.4": "3.1.4",
+      "fast-uri@>=3.0.0 <3.1.5": "3.1.5",
       "brace-expansion@<1.1.18": ">=1.1.18",
       "brace-expansion@>=2.0.0 <2.1.4": ">=5.0.9",
       "brace-expansion@>=4.0.0 <5.0.9": ">=5.0.9",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,7 +11,7 @@ overrides:
   fast-xml-parser@<5.5.7: '>=5.5.7'
   fast-xml-parser@<5.7.0: '>=5.7.0'
   fast-xml-parser@<5.10.1: '>=5.10.1'
-  fast-uri@>=3.0.0 <3.1.4: 3.1.4
+  fast-uri@>=3.0.0 <3.1.5: 3.1.5
   brace-expansion@<1.1.18: '>=1.1.18'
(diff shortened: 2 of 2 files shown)


bdunncompany:
Superseded by #3144, which landed the identical change — `"fast-uri@>=3.0.0 <3.1.5": "3.1.5"`, same pin to the 3.x fix rather than the 4.1.2 major. Main's `Security Scanning` is green again on `bf69b7137`. Closing rather than leaving a duplicate in the queue.

Nothing needed here. Recording one thing that might be worth a guard, since this is the second time this exact shape has bitten: the override that failed was `"fast-uri@>=3.0.0 <3.1.4": "3.1.4"` — a range that pinned to precisely the version later flagged, so it silently stopped binding anything the moment the advisory moved. `scripts/security/check-override-floors.mjs` passes on that state (it reports "every override still binds", and it did — to a now-vulnerable version). A pin whose target equals its own upper bound is structurally a floor that can only ever go stale. If the floors checker flagged `<X` → `X` pairs as "pinned to the edge of its own range", this class would surface at authoring time instead of at the next CVE.

Happy to put that up as a small addition to the checker if you think it earns its keep.

PR #3148, opened 2026-08-04
fix(mobile): stop notification settings asserting behavior the app doesn't deliver (#3143)

### LINKED ISSUE #3143: [Mobile] Notification settings give false assurance: iOS toggles are write-only, failed-push banner is dismissible forever
## Description

Two residuals from #3118 (PR #3125 fixed the Android `unsupported` case), same false-assurance class — the UI asserts notification behavior the app doesn't deliver:

1. **iOS Notifications / Critical-only toggles are write-only.** The toggles in `apps/mobile/src/screens/chat/components/SettingsSheet.tsx` persist their state to AsyncStorage, but nothing consumes those keys — no code path filters or suppresses pushes based on them. Turning "Critical only" on changes nothing; turning Notifications off changes nothing.
2. **The `failed` push-registration banner is dismissible per session.** When registration status is `failed`, ApprovalGate shows a banner the user can dismiss; after dismissal the Notifications toggle (ON) is the only remaining signal, which is wrong — pushes are not arriving.

## Proposed Fix

1. Either wire the toggles to real behavior (respect them at notification display/registration time, or send them server-side to filter dispatch) or remove/disable them until they do something.
2. For `failed` status, give the Notifications row the same status-aware treatment #3125 added for `unsupported` (disabled/explanatory state), so the truth survives banner dismissal.

## Affected Files

- `apps/mobile/src/screens/chat/components/SettingsSheet.tsx`
- `apps/mobile/src/services/notifications.ts`
- ApprovalGate banner logic (dismissal state)

## Reported By

Review findings on #3125; follow-up to #3118

ToddHebebrand:
Closes #3143

**STACKED on #3125** (`fix/3118-android-push-settings-state`). Retarget to `ToddHebebrand/ios-app-testing` when #3125 merges. Because the base is a sibling branch, `ci.yml` does not run on this PR — verification is local (below).

## What was wrong

1. The Notifications / Critical-only toggles in `SettingsSheet.tsx` were write-only: they persisted `notificationsEnabled` / `notificationsCriticalOnly` to AsyncStorage, but no code path anywhere consumed those keys. Flipping them changed nothing.
2. When push registration status is `failed`, ApprovalGate's banner is dismissible per session; after dismissal the ON Notifications toggle was the only remaining signal — asserting pushes work when they don't.

## Which branch I took for part 1, and why

Checked for a client-side seam where the toggles could take real effect. The only display seam is `Notifications.setNotificationHandler` in `services/notifications.ts`, and it runs **foreground-only** — background/lock-screen pushes (where approval pushes overwhelmingly land) are presented by the OS with no JS involvement. Wiring the toggles there would leave "Notifications: Off" false for the dominant case, and "Critical only" can never be honored client-side at all (severity filtering needs server-side dispatch changes, explicitly out of scope). Token-unregistration on toggle-off could make the master toggle real, but drags in token persistence, a re-register flow, and paired-device/approver interplay — not a clean seam.

So: **no clean seam → the toggles are removed** and the row becomes an honest, status-aware readout, per the issue's "honesty over dead controls" option:

- `ok` — "Approval pushes and alerts are delivered to this phone. Tap to manage them in Settings." Tapping opens the app's system notification settings (`Linking.openSettings()`) — the one real on/off control the user actually has.
- `unsupported` — keeps #3125's reason-specific copy verbatim (same strings, same paired-devices hint).
- `failed` — **part 2**: the row now says "Push notifications aren't registered … sign in again", phrased to mirror ApprovalGate's `PushFailedBanner` so the two surfaces never contradict; for `permission_denied` it names Settings as the fix and deep-links there. The banner stays dismissible (by design — it's a standing-condition overlay), but the truth now survives dismissal in Settings.
- `idle` — neutral "Checking push registration…".

The dead AsyncStorage keys and handlers are removed; `ToggleRow`'s now-unused `disabled` support (added in #3125 solely for the unsupported row) is removed too since Biometric is its only remaining caller.

## Tests

Extends the #3125 pure-module pattern:

- `pushUnavailableCopy.test.ts`: new `notificationsRowCopy` suite — the core pin is *only `ok` may claim delivery*, asserted across every status × reason combination; plus per-status copy/deep-link/paired-hint assertions and verbatim-reuse of the #3118 unsupported strings.
- `notifications.test.ts`: reason-string contract pin for the `failed` branch — `permission_denied` from `registerForPushNotifications` must map to the Settings-actionable copy (same rationale as #3125's unsupported pin: a renamed reason would silently degrade to generic copy with both sides' tests green).

## Verification (local — stacked PR runs no CI)

- `npx tsc --noEmit` in `apps/mobile`: clean
- Full mobile vitest suite: 28 files, 275 passed, 3 skipped

### DIFF
--- a/apps/mobile/App.tsx
+++ b/apps/mobile/App.tsx
@@ -51,14 +51,17 @@ import { StatusBar } from 'expo-status-bar';
 import * as SplashScreen from 'expo-splash-screen';
 import { Provider as ReduxProvider, useDispatch, useSelector } from 'react-redux';
 import { Provider as PaperProvider, MD3DarkTheme, MD3LightTheme } from 'react-native-paper';
-import { useColorScheme } from 'react-native';
+import { AppState, useColorScheme } from 'react-native';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 
 import { store, type AppDispatch, type RootState } from './src/store';
 import { RootNavigator } from './src/navigation/RootNavigator';
 import { AppLockGate } from './src/navigation/AppLockGate';
-import { registerForPushNotifications } from './src/services/notifications';
+import {
+  registerForPushNotifications,
+  reconcilePushRegistration,
+} from './src/services/notifications';
 import { setPushRegistration } from './src/store/authSlice';
 
 // Hold the native splash until the first real screen is ready to paint.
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter; pr-test-analyzer run was interrupted and not re-dispatched)
**Findings:** 2 high + 2 minor raised, all addressed in cf251fb71; 0 outstanding.
- Stale `'ok'`: the row's own Settings deep-link let the user revoke permission without the app noticing (iOS doesn't restart on permission change) → new `reconcilePushRegistration()` re-checks on every AppState `active`, downgrading `ok` → `failed/permission_denied` and re-registering when permission comes back.
- `'idle'` forever: `registerForPushNotifications` had throw paths outside its try/catch (permission calls before it, Android channel setup after) that became unhandled rejections in `PushRegistrationGate` → outcome contract is now total, channel setup best-effort, and the gate catches + dispatches `failed` as a backstop.
- Minor: `openSettings` catch now warns before the Alert fallback; logout resets `pushRegistration` so the next account never briefly sees stale copy.

**Tests:** full apps/mobile vitest suite green (28 files, 282 passed, 3 skipped); `tsc --noEmit` clean. Stacked PR — `ci.yml` does not run on a sibling-branch base; verification is local.
**Status:** review-clean, awaiting maintainer merge (retarget to `ToddHebebrand/ios-app-testing` after #3125 lands).

PR #3150, opened 2026-08-04
feat(patching): third-party auto-approval as a first-class Update Ring control

ToddHebebrand:
## Summary

Third-party (winget/Chocolatey/Homebrew) patching becomes a visible, explicit Update Ring auto-approval control. Previously it only worked via an invisible severity exemption keyed off the config policy's `sources` — nothing on the ring UI mentioned third-party at all.

- **New ring control:** `autoApprove.thirdPartyApps` (+ optional `thirdPartyDeferralDays`, first-seen anchored) in the ring's `auto_approve` JSONB. Ring form gets a "Third-party applications" subsection; rings list gets auto-approve badges; the config-policy Patch tab cross-links to the ring.
- **Dual consent, fail-closed:** third-party auto-approves only when the config policy's patch `sources` include `third_party` AND the ring toggle is on. Absent `sources` on legacy job snapshots always refuses (no scope-widening). Third-party-only rings (no OS severities) are now expressible.
- **Dead column deprecated:** `patch_policies.sources` (written since #1428, never read by the evaluator) loses all writers/readers this release; `DROP COLUMN` ships next release (follow-up issue).
- **Backfill migration** stamps explicit `thirdPartyApps` on existing rings (preserving the old exemption behavior) and converts/strips stored `third_party_app` category rules (1 row fleet-wide, currently inert). Live-verified with 4 probe cases incl. idempotency.

## Pre-existing bugs fixed en route

- **Category-rule severity chips were inert (fail-open):** the UI/API write `autoApproveSeverities` but the evaluator read `severityFilter` — "auto-approve critical only" approved every severity in the category. Now enforced (legacy alias still honored on read).
- **"Manual approval" category rules were not terminal** — they fell through to ring-level auto-approve, contradicting the UI copy.
- A third-party patch carrying an OS-style severity could slip through the OS severity-membership path when policy sources were absent — the 3P branch now fully owns third-party candidates.
- `parseRingAutoApprove` coerced invalid non-zero deferrals to 0 (fail-open) — now disables the row.

## Spec / plan

- `docs/superpowers/specs/vuln-patch/2026-08-04-third-party-update-ring-auto-approve-design.md`
- `docs/superpowers/plans/vuln-patch/2026-08-04-third-party-update-ring-auto-approve.md`

## Testing

- Unit: shared 1617, API 19528, web 4824 — all green; `tsc --noEmit` clean.
- Integration (real Postgres, clean per-worktree stack): `patchThirdPartyRingAutoApprove` 10/10 (dual-consent matrix, legacy-snapshot refusal, 3P-only ring, custom source, first-seen deferral override) + adjacent approval suites 6/6.
- i18n: all 7 locales carry the new keys; localeParity/translationCoverage/keyUsage suites green.
- Migration: probe-verified live (conversion, explicit-key preservation, disabled-ring severities cleared on enable, idempotent re-run).

## Deploy notes

- No new env vars, no schema DDL (data-only migration, runs via autoMigrate at boot).
- Release N of expand/contract for `patch_policies.sources` — the column stays but is unused; drop lands next release.

### DIFF
--- a/apps/api/migrations/2026-08-13-ring-third-party-auto-approve-backfill.sql
+++ b/apps/api/migrations/2026-08-13-ring-third-party-auto-approve-backfill.sql
@@ -37,10 +37,24 @@ BEGIN
   -- 2) Rings with an autoApprove:true third_party_app category rule: turn on
   --    the ring-level toggle (carrying the rule's deferral override) and strip
   --    the rule. Preserves intent: those rings wanted 3P auto-approved.
+  --    Accepted narrowing: an {enabled:true, severities:[]} ring with a 3P
+  --    rule loses its rule-based 3P behavior — statement 1 runs first and
+  --    stamps thirdPartyApps:false (no recognized severities), so this
+  --    statement's "lacks thirdPartyApps" guard then skips it, and statement
+  --    3 strips the now-inert rule. Fail-closed; 0 known rows.
   UPDATE patch_policies p
   SET auto_approve =
         (CASE WHEN jsonb_typeof(p.auto_approve) = 'object' THEN p.auto_approve ELSE '{}'::jsonb END)
         || jsonb_build_object('enabled', true, 'thirdPartyApps', true)
+        -- The pre-image's OS auto-approve state must survive this ring-level
+        -- enable: if the ring wasn't already boolean-true enabled, its
+        -- severities (if any) belonged to a disabled state and must not
+        -- silently start applying now that 'enabled' flips to true. Compare
+        -- against the jsonb boolean (not ->>'enabled' = 'true' text, which a
+        -- malformed {"enabled":"true"} string row would also fail — and we
+        -- want that case cleared too).
+        || CASE WHEN p.auto_approve->'enabled' = 'true'::jsonb THEN '{}'::jsonb
+                ELSE jsonb_build_object('severities', '[]'::jsonb) END
         || COALESCE(
              (SELECT CASE WHEN (r.rule->>'deferralDaysOverride') ~ '^\d{1,3}$'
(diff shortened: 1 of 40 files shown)


ToddHebebrand:
Ran a 4-agent review round (code / tests / silent-failures / comments) over the full diff; all confirmed findings are fixed in 10fb3096c, CI fully green (including all 4 integration shards).

Highlights:
- **Backfill migration reordered** — rule conversion now runs before the generic `thirdPartyApps` stamp and carries the old category-path hold (`deferralDaysOverride ?? deferral_days` column) into `thirdPartyDeferralDays`. This was the one fail-open path: converted rings would have auto-approved 3P updates with no hold. The "accepted narrowing" case is gone as a side effect. New `ringThirdPartyBackfillMigration.integration.test.ts` pins every pre-image shape + idempotency (ran green in shard 3).
- **Dual-consent bypass closed** — a literal category allow rule (e.g. `homebrew`) matching a third-party patch now requires the same policy-sources + ring-toggle legs as Priority 3; deny rules stay terminal.
- **Old-shape writes can no longer reset the 3P opt-in** — `thirdPartyApps`/`thirdPartyDeferralDays` are optional in the write schema and merge from the stored row (`mergeRingAutoApproveWrite`) on REST PATCH and the AI tool.
- `thirdPartyDeferralDays` parsing is now fail-closed like `deferralDays`, and both malformed paths warn + Sentry with ring context; executor `categoryRules` snapshot entries are schema-validated like every sibling field; `migrateToConfigPolicies` reads `primary.sources` again (hardcoded `['os']` dropped third-party opt-ins on first run).

PR #3164, opened 2026-08-05
fix(api): stop redactLogFields dropping (and reprototyping on) __proto__ (#3129)

### LINKED ISSUE #3129: [API] logRedaction drops __proto__ subtrees entirely (data loss in redacted output)
## Description
The shared `logRedaction.ts` key-walker drops a `__proto__`-named subtree from its output entirely rather than redacting/renaming it. Not a security hole (dropping is fail-safe against prototype pollution) but it silently loses data in redacted logs/tool output, which can mislead debugging.

## Proposed Fix
Rename the key on output (e.g. `__proto__` → `"__proto__(renamed)"`) using a null-prototype carrier object, keeping pollution safety while preserving the value.

## Affected Files
- `apps/api/src/utils/logRedaction.ts` (verify exact path — shared redaction util used by aiToolOutput)

## Reported By
Review round on PR #3102 (issue #3093), deferred pre-existing nit.

bdunncompany:
Closes #3129.

## One correction to the issue's framing

The issue says "not a security hole (dropping is fail-safe against prototype pollution)". That holds for a **primitive** value. For an **object** value it does not — verified under node 22.23.2:

```js
const o = {};
o['__proto__'] = { leaked: 'value' };
Object.keys(o)                                  // []            <- key gone
Object.getPrototypeOf(o) !== Object.prototype   // true          <- prototype REPLACED
o.leaked                                        // 'value'       <- now inherited
```

`redacted[key] = …` never creates an own property for that key; it hits the setter inherited from `Object.prototype`. So the object case doesn't just lose the field — it reprototypes an object assembled from untrusted agent input, and the result inherits the agent's keys. Still not remotely exploitable, and `JSON.stringify` ignores inherited properties so stored output is unaffected, but any code reading the returned object can see them. Worth naming rather than filing purely under data loss.

## The fix

`Object.defineProperty`, which writes a real own property and leaves the prototype alone.

**I did not take the rename the issue proposed** (`__proto__` → `"__proto__(renamed)"` via a null-prototype carrier). Two reasons: renaming preserves the value but distorts the field name in exactly the log someone is reading to debug it, and a null-prototype carrier changes the shape of *every* returned object to fix one key — `hasOwnProperty` and friends stop working on the result. `defineProperty` keeps the field under its true name, keeps an ordinary prototype, and touches nothing else. `JSON.stringify` round-trips it unchanged, so persisted output is identical to any other field.

Happy to switch to the rename if you'd rather the redacted output never contain a literal `__proto__` key at all — that's a defensible position for anything downstream that might spread the object into a plain literal, it just wasn't the trade I'd pick by default.

## Tests

Five cases, and **I confirmed they fail against the previous implementation** before the fix went in — 4 of the 5 fail; the secret-key-nested case passes either way because it never reaches the assignment:

```
FAIL  __proto__ handling (#3129) > preserves a __proto__ subtree instead of dropping it
FAIL  __proto__ handling (#3129) > does not let a __proto__ value reprototype the redacted object
FAIL  __proto__ handling (#3129) > preserves a primitive __proto__ value, which the setter discards outright
FAIL  __proto__ handling (#3129) > survives JSON round-trip with the key intact
```

Two assert the **prototype** specifically, not just the key, because a fix that only restored the key could still reprototype and would otherwise look green.

## Local gate

Node 22.23.2, all green:

- `tsc --noEmit` — clean, exit 0
- `vitest run` in `apps/api` — **1253 files / 19705 tests passed, 0 failed**
- `eslint` on both changed files — clean

Single-file change plus its test; no dependency, lockfile, Go or container surface touched.

### DIFF
--- a/apps/api/src/services/logRedaction.test.ts
+++ b/apps/api/src/services/logRedaction.test.ts
@@ -36,6 +36,51 @@ describe('log redaction', () => {
     );
   });
 
+  // #3129. Plain assignment to the literal key `__proto__` hits the setter
+  // inherited from Object.prototype, so the field silently disappears from the
+  // output — and when the value is an object, the returned object's prototype is
+  // replaced with it. These assert the field survives AND the prototype does not
+  // move, because a fix that only restored the key could still reprototype.
+  describe('__proto__ handling (#3129)', () => {
+    it('preserves a __proto__ subtree instead of dropping it', () => {
+      const out = redactLogFields(JSON.parse('{"__proto__":{"inner":"kept"},"other":1}')) as Record<string, unknown>;
+
+      expect(Object.prototype.hasOwnProperty.call(out, '__proto__')).toBe(true);
+      expect(Object.keys(out)).toContain('__proto__');
+      expect((out as any).__proto__).toEqual({ inner: 'kept' });
+      expect(out.other).toBe(1);
+    });
+
+    it('does not let a __proto__ value reprototype the redacted object', () => {
+      const out = redactLogFields(JSON.parse('{"__proto__":{"polluted":"yes"}}')) as Record<string, unknown>;
+
+      expect(Object.getPrototypeOf(out)).toBe(Object.prototype);
+      // The value must be readable as data, never inherited behaviour.
+      expect(({} as any).polluted).toBeUndefined();
+    });
+
+    it('preserves a primitive __proto__ value, which the setter discards outright', () => {
+      const out = redactLogFields(JSON.parse('{"__proto__":"just-a-string"}')) as Record<string, unknown>;
+
+      expect((out as any).__proto__).toBe('just-a-string');
+      expect(Object.getPrototypeOf(out)).toBe(Object.prototype);
+    });
+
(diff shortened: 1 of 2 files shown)


ToddHebebrand (approved):
**Approved — the mechanism, the fix, and the tests all check out.** Plain assignment to the literal key `__proto__` invokes the `Object.prototype` accessor setter rather than creating an own property, so the old code dropped the field for primitives and silently reprototyped the redacted object for object values — a redaction pass that mutates the prototype of an object built from untrusted agent input is exactly the wrong place for that. `Object.defineProperty` is the right repair, and `__proto__` is the only string key on `Object.prototype` with accessor semantics, so the single special case is complete. The reprototype-assertion test (`Object.getPrototypeOf(out)` stays `Object.prototype`) is the one that matters most and it's here.

**One follow-up worth a small PR:** `redactSecretsDeep` at `apps/api/src/services/secretRedaction.ts:182-186` has the identical `out[key] = …` over `Object.entries` pattern on the same untrusted-agent-JSON input class. Same one-key `defineProperty` fix applies. Yours if you want it.

Issue #2387, opened 2026-07-12
[Agent] macOS agent retains ~2.5 GB of live, compressed heap (root cause unconfirmed — evidence + leading suspect)


ToddHebebrand:
## Description

A macOS `breeze-agent` root daemon was observed holding **2.52 GB** in Activity Monitor after ~2 days uptime. This is real retained memory, not a reporting artifact. **Root cause is not yet confirmed** — this issue records the evidence and the leading suspect so the investigation isn't restarted from zero.

## Evidence

**The memory is cold, not resident.** `top -pid <pid> -stats mem,cmprs`:

```
MEM    CMPRS
2572M  2566M     <- RSS via ps: ~24 MB
```

Only ~15-24 MB is resident; **~2.5 GB is compressed**. The agent allocated it and never reads it again, so macOS squeezed it all into the compressor.

**It is flat, not growing.** Sampled every 60s for 30 minutes: oscillates 2571M-2586M, no trend. This is a *stuck set*, not a leak in progress.

**It is live and reachable — not a freed spike.** Go's darwin scavenger uses `MADV_FREE_REUSABLE`, which *does* decrement `phys_footprint`. Verified with a harness: allocate 1 GB of pointer-free data, drop it, force GC:

```
SCAN PEAK (live)      phys_footprint=1073 MB | HeapAlloc=967 MB
after drop + GC       phys_footprint=1073 MB | HeapAlloc=  0 MB
+60s (scavenger)      phys_footprint=  50 MB | HeapReleased=1042 MB
```

The OS gets the memory back within ~60s. A transient high-water mark from 2 days ago therefore **cannot** still show 2.5 GB. The memory is retained.

**It is a few large buffers, not many small ones.** The GC must traverse pointer-bearing memory every cycle (Go force-GCs every 2 min). A 2.5 GB live set made of millions of small strings would keep buckets and string headers hot and push RSS into the hundreds of MB. RSS is 14 MB. So the pointer-bearing live set is tiny, and the 2.5 GB is **a small number of very large pointer-free (noscan) buffers** — on the order of ~40 x 64 MB, not 4M x 600 B.

## Leading Suspect (unconfirmed)

`agent/internal/websocket/client.go:353`

```go
go c.processCommand(cmd)   // no semaphore, no concurrency cap
```

with `agent/internal/websocket/client.go:28`:

```go
maxMessageSize = 64 * 1024 * 1024  // 64MB — file_write commands carry base64 content
```

[...]

PR #3165, opened 2026-08-05
fix(agent,api): thread the caller's timeout budget through to the helper IPC (#3112)

### LINKED ISSUE #3112: [Agent] Vision/desktop tool commands are capped at ~30s by a hardcoded IPC timeout beneath the raised API budget (follow-up to #3091)
Follow-up to #3091, filed at your recommendation in https://github.com/LanternOps/breeze/issues/3091#issuecomment-5182069716.

## Summary

#3096 raised the API-side budget for `take_screenshot` / `analyze_screen` / `computer_control` to 120s and threaded it into the `executeCommand` call so it actually takes effect. In **service/headless mode** the Go agent still enforces its own independent ceiling of roughly 30s beneath all of that, so the tools these timeouts were raised for can still fail at ~30s on exactly the deployment mode the raise was aimed at.

## Verified

`agent/internal/heartbeat/handlers_screenshot.go` routes these commands to the user helper over IPC with a fixed per-attempt timeout and a retry:

```go
const maxAttempts = 2
for attempt := 0; attempt < maxAttempts; attempt++ {
    ...
    resp, err := session.SendCommand(ipcCmd.CommandID, ipc.TypeCommand, ipcCmd, 15*time.Second)
    if err != nil {
        if attempt < maxAttempts-1 {
            log.Warn("IPC tool command failed, retrying with new helper", ...)
            continue
        }
        return tools.NewErrorResult(
            fmt.Errorf("IPC %s failed after %d attempts: %w", cmdType, maxAttempts, err), ...)
    }
```

`15s × 2 attempts ≈ 30s`, hardcoded, with no reference to any server-supplied budget. `handlerRegistry[tools.CmdTakeScreenshot] = handleTakeScreenshot` → `executeToolViaHelper(tools.CmdTakeScreenshot, ...)`, so the screenshot path goes through it.

## Why it matters

The three layers are now inconsistent rather than nested:

| Layer | Budget | Where |
|---|---|---|
| SDK tool wrapper | 120s | `withToolTimeout` in `aiAgentSdkTools.ts` |
| Command wait | 120s | `waitForCommandResult` via `getToolTimeout` (fixed in #3096) |
| **Agent IPC to helper** | **~30s** | `handlers_screenshot.go:60`, hardcoded |

The innermost limit is the smallest, so it is the one that decides. That is the same shape as the bug #3096 just fixed one layer up — the outer budget was raised while an inner hardcoded 30s kept firing first — which suggests the real defect is structural: **each layer carries its own constant instead of deriving from one budget.**

## Suggested direction

Rather than changing `15 * time.Second` to a bigger constant and leaving a fourth independent number in place, it would be worth having the server-side budget travel with the command so the agent bounds the IPC wait by what the caller is actually willing to wait for, falling back to the current value when absent. That makes the layers nest by construction and removes the "raise it in one place, another place still fires" failure mode.

If a smaller change is preferred for now, raising the per-attempt timeout for the helper-routed tool commands specifically would close the immediate gap, at the cost of leaving the same class of bug latent.

## Note on scope

Worth checking whether `computer_action` and any other helper-routed tool commands share this path before fixing, so the change covers all of them rather than just the screenshot handler.

bdunncompany:
Closes #3112. Fix direction as agreed on the issue — thread the caller's deadline through with the command rather than bumping a constant.

## It needed both sides

`timeoutMs` never reached the device. It's a server-side wait option on `executeCommand` and was not in the payload the agent receives, so raising it (as #3096 did, to 120s) only changed how long the server waited while the device kept giving up at ~30s underneath.

**API** — publishes it as `timeoutSeconds`, the key the agent's script path already reads.

**Agent** — `executeToolViaHelper` derives its deadline from that via the existing `helperCommandTimeout`, so the #2387 clamping still applies and a huge payload value still can't park a worker goroutine.

## Three judgement calls worth your eyes

**Agent-targeted commands only.** The watchdog has no helper-IPC path, and its callers are explicitly told to pass a *larger* `timeoutMs` than agent commands — publishing that inflated number into its payload would mislead. Your existing test `routes watchdog-targeted commands to the row insert without WS dispatch` caught the broader version by asserting an exact payload; I narrowed the change so that test passes **unchanged** rather than editing it to accommodate me. That felt like the test doing its job, not being in the way.

**Default of 25s when the server sends nothing.** `helperCommandTimeout(0)` falls back to the script executor's 300s. Inheriting that would turn a silent omission — an older API, a caller that set no budget — into a five-minute parked goroutine for what's usually a screenshot, and leave an agent talking to an older API *worse off* than before this change. 25s preserves the previous ~30s total. There's a test pinning that it never exceeds the pre-fix ceiling.

**Budget is for the whole operation, not per attempt.** The deadline is computed once and each attempt gets what remains, so a retry can't double the caller's wait. That made the function's trailing `// Unreachable — loop always returns` genuinely reachable, so it now returns a real error naming the exhausted budget rather than the placeholder.

## Tests

Go pins the deadline derivation directly rather than the IPC plumbing (which needs a live helper): budget honoured above the old ceiling, absent budget falling back to the tool default rather than the script default, `float64` from JSON decoding, #2387 clamping still applying, and a guard that the default never regresses past the pre-fix ceiling. **Confirmed failing** against a simulated regression (default reverted to 300) before the fix went in.

TS asserts the payload that actually reaches the insert via the mocked `db.insert`. An earlier draft asserted an inline copy of the expression and would have passed against any implementation — same trap as the inert `toolTimeouts` assertions on #3096, so I threw it away.

## Local gate

- `tsc --noEmit` clean; `vitest run` in `apps/api` **1253 files / 19707 tests passed, 0 failed**; eslint clean on both TS files
- `go test ./internal/heartbeat/ ./internal/remote/tools/` — both ok; gofmt clean on both changed Go files

**One gap, stated plainly: I could not run `go test -race`.** Race detection requires cgo, and cgo fails in my environment with `runtime/cgo: open terminal failed: not a terminal`. I reproduced that on a clean checkout with these changes stashed, so it's environmental rather than introduced here — but it does mean the race pass is CI's first look, not a re-run of something I already saw green.

### DIFF
--- a/agent/internal/heartbeat/handlers_screenshot.go
+++ b/agent/internal/heartbeat/handlers_screenshot.go
@@ -29,6 +29,19 @@ func handleTakeScreenshot(h *Heartbeat, cmd Command) tools.CommandResult {
 	return tools.TakeScreenshotWithCapture(cmd.Payload, h.desktopCaptureFn())
 }
 
+// minHelperAttemptTimeout is the floor for a single IPC attempt. It keeps a
+// nearly-exhausted budget from producing an instant, meaningless failure — the
+// helper still gets a real chance to answer — while staying well under any
+// server-side budget worth threading through.
+const minHelperAttemptTimeout = 5 * time.Second
+
+// defaultHelperToolTimeoutSeconds is used when the server sends no
+// timeoutSeconds — an older API, or a caller that never set one. It preserves
+// the previous ~30s total ceiling (15s x 2 attempts) rather than inheriting the
+// script executor's 300s default, which would turn a silent omission into a
+// five-minute parked goroutine for what is usually a screenshot.
+const defaultHelperToolTimeoutSeconds = 25
+
 // executeToolViaHelper sends a screenshot/computer_action command to the user
 // helper process via IPC and returns the result. If the helper crashes, it
 // automatically respawns and retries once.
@@ -41,8 +54,35 @@ func (h *Heartbeat) executeToolViaHelper(cmdType string, payload map[string]any,
 		)
 	}
 
+	// #3112: the IPC wait used to be a hardcoded 15s per attempt, so the whole
+	// helper round-trip was capped near 30s no matter what budget the caller was
+	// willing to wait for — the API could raise its own timeout to 120s and the
+	// tool would still die down here. Derive the deadline from the server-supplied
+	// timeoutSeconds instead, exactly as the script path does, reusing
+	// helperCommandTimeout so the executor's clamping (#2387) applies and a huge
+	// payload value cannot park a worker goroutine.
+	//
+	// The budget is for the whole operation, not per attempt: the deadline is
+	// computed once and each attempt gets what is left, so a retry cannot double
+	// the caller's wait. An attempt is only started if there is still meaningful
+	// time for it.
+	overallTimeout := helperCommandTimeout(tools.GetPayloadInt(payload, "timeoutSeconds", defaultHelperToolTimeoutSeconds))
+	deadline := start.Add(overallTimeout)
+
 	const maxAttempts = 2
 	for attempt := 0; attempt < maxAttempts; attempt++ {
+		remaining := time.Until(deadline)
+		if remaining < minHelperAttemptTimeout {
+			if attempt == 0 {
+				// Budget was already spent before the first attempt; give it the
+				// floor rather than failing without ever trying the helper.
+				remaining = minHelperAttemptTimeout
+			} else {
+				log.Warn("IPC tool command budget exhausted, not retrying",
+					"cmdType", cmdType, "attempt", attempt+1)
+				break
+			}
+		}
 		session := h.findOrSpawnHelper("")
 		if session == nil {
 			return tools.NewErrorResult(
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Post-merge review (ran alongside the merge): clean.** Independently rebuilt and race-tested the diff — the budget-shared-across-attempts invariant holds (deadline computed once at `agent/internal/heartbeat/handlers_screenshot.go:57-73`, retry gets the remainder, floor only on attempt 0), reuse of `helperCommandTimeout` keeps the #2387 clamp so a hostile payload value can't park a goroutine, and `GetPayloadInt` ignoring unknown keys makes both mixed-fleet directions safe. The 25s default preserving the old ~30s ceiling instead of inheriting the script executor's 300s was the right instinct.

**One observation, pre-existing and non-blocking:** the API's `waitForCommandResult` polls for exactly `timeoutMs` while the agent's budget is `timeoutMs + 5s` grace — a completion landing in that window arrives after the API already reported timeout. Same pattern as the script path, so nothing this PR regressed; noting it in case we ever chase "result arrived but UI said timeout" ghosts.

Issue #2434, opened 2026-07-13
[API] Sweep: agent-supplied error strings persisted unredacted outside device_commands (script_executions, backup/vault/CIS/monitor results)


ToddHebebrand:
## Description

Follow-up to #2419 (fixed in PR #2432 for `device_commands.result`): review of that fix flagged that several per-type agent WS handlers persist agent-supplied `error`/message strings into **other** tables without `redactSecretsFromOutput`, e.g.:

- `apps/api/src/routes/agentWs.ts:403` — `script_executions.errorMessage`
- backup verification results, vault sync status, CIS results, monitor results (same file, per-type handlers)

Same trust-boundary rationale as #2359/#2419: an agent-supplied error string can carry secrets (key material, tokens, connection strings) and these columns surface in the web UI to non-admin roles.

## Proposed Fix

Sweep `agentWs.ts` (and the REST twins where applicable) for every persistence of agent-supplied free-text error/status fields; apply `redactSecretsFromOutput` at a shared chokepoint if practical (e.g. a helper wrapping the field assignment) rather than per-call-site. Add a test per persisted surface.

## Affected Files
- `apps/api/src/routes/agentWs.ts` (multiple per-type handlers)

Found by PR-review of #2432; pre-existing, out of scope there.

ToddHebebrand:
**PR:** #2462 — https://github.com/LanternOps/breeze/pull/2462 (open, review-clean, not merged)

Swept as requested, with a covered / not-applicable disposition for **every** field that persists an agent-supplied string — the full enumeration is in the PR body, since silent truncation of the sweep is the failure mode this issue exists to prevent.

**Approach:** a shared chokepoint (`redactAgentResultErrorFields`) applied once at the entry of **both** command-result ingest legs — `agentWs.processCommandResult` / `processOrphanedCommandResult` and the REST twin `agents/commands.ts` — so every per-type handler downstream receives already-redacted `error`/`stderr` and a future handler cannot forget to redact.

**One design point worth knowing:** `stdout` is deliberately **not** redacted at the chokepoint. Several handlers parse it as structured JSON, and artifact-bearing stdout (`capture_pprof`) must stay byte-for-byte or the redaction patterns statistically corrupt it (#2401). So each site that persists *stdout-derived free text* redacts for itself — that turned out to be where the remaining leaks actually were.

**Beyond the surfaces named in the issue**, the sweep found and closed:
[...]

Issue #2133, opened 2026-07-01
Partner-wide automations: dual-ownership (org XOR partner)


ToddHebebrand:
Part of the partner-wide-first epic (Phase 3). `automations` (`apps/api/src/db/schema/automations.ts`, org_id NOT NULL) cannot be shared across orgs, even though the `scripts` they invoke are already dual-ownership. Apply the dual-ownership playbook so a partner-wide automation (e.g. "on device.offline run diagnostic script") runs across every org: migration + XOR CHECK + dual-axis RLS, `automationWorker` partner-owned trigger resolution across the partner's orgs, ownerScope routes gated on `canManagePartnerWidePolicies`, UI scope + badge, functional RLS integration tests. Watch for org-specific references inside automation actions (notification channels — see the delivery-rails sub-issue — and script targets).

ToddHebebrand:
**Mapping complete** — full call-site sweep is in `docs/superpowers/plans/2026-07-02-automations-partner-ownership-mapping.md` on PR #2157. Headlines:

- The event-dispatch fan-out gap is a single site: `queueEventTriggers` (`apps/api/src/jobs/automationWorker.ts:730`). The worker already runs under a system DB context, so RLS is not the blocker — only the app-layer `eq(automations.orgId, event.orgId)` match is.
- `apps/api/src/services/policyEvaluationService.ts` has three near-duplicate `eq(automations.orgId, device.orgId)` sites (remediation resolution/trigger paths) that would silently miss partner-wide automations.
- `executeCreateAlertAction` in `automationRuntime.ts` writes alerts with the AUTOMATION's org — must switch to the device's org (playbook rule 5) once org_id can be NULL.
- Reference implementation to copy: `2026-07-01-automation-policies-partner-ownership.sql` (same schema file) plus the maintenance-windows Step 3 shape for the `automation_runs` EXISTS-join policies.

Implementation starting now off current main (post-#2156, so the channel-ownership intersection is already dual-axis).

PR #3186, opened 2026-08-06
fix(api): persist stdout for automation-triggered script runs (#3162)

### LINKED ISSUE #3162: [API] Automation-triggered script runs never persist stdout — synthetic executionId fails the UUID match and the result update is silently dropped
**Symptom:** script output from automation `run_script` actions is not visible anywhere in the product. Surfaced by a user in discussion #2598 comparing against another RMM that shows the latest run's stdout inline next to the automation.

**It is not just unrendered — the output is dropped server-side.**

- Manual runs (`POST /scripts/:id/execute`) write a `script_executions` row, and `handleScriptResult` (`apps/api/src/routes/agentWs.ts:414-469`) fills in `stdout` when the agent result arrives. Execution history and the device script-history tab render that.
- Automation `run_script` actions never create an execution row. `executeRunScriptAction` (`apps/api/src/services/automationRuntime.ts:868-873`) queues the command with a synthetic `executionId` of `${runId}:${deviceId}:${actionIndex}` — not a UUID. When the result comes back, `handleScriptResult` matches it against `script_executions.id`, a `uuid` column (`apps/api/src/db/schema/scripts.ts:112-128`); the update throws and lands in a caught, console-logged error (`agentWs.ts:466-468`). The real stdout is discarded.
- `automation_run_device_results.output` (`apps/api/src/db/schema/automations.ts:72-89`) was clearly meant for this but only ever receives the automation's own log lines (`"[info] Queued run_script action"` — `automationRuntime.ts:1750-1820`), never the script's output.
- Even that field is unrendered: `AutomationRunHistory.tsx:30` declares `output?: string` but the expanded panel shows only `result.error`.

**Fix shape:** persist the script's real stdout for automation-triggered runs (either mint a genuine `script_executions` row per device-action and correlate on it, or route the result into `automation_run_device_results.output`), then render it in the run-history expanded panel.

Related, distinct: #2698 (script stdout → custom fields), discussion #2598 (run-on-onboarding trigger). Note the synthetic id also means these results are invisible to the HTTP-path fix discussed in #3097 — whatever shared handler registry comes out of that should handle this correlation too.

ToddHebebrand:
Closes #3162

## The bug

Automation `run_script` actions queued the device command with a synthetic
`${runId}:${deviceId}:${actionIndex}` executionId. `handleScriptResult`
(`apps/api/src/routes/agentWs.ts`) matches that against `script_executions.id`
— a `uuid` column — so Postgres raised `22P02: invalid input syntax for type
uuid`, the throw landed in a caught `console.error`, and the agent's real
stdout was thrown away. Automation script output was invisible everywhere:
no execution row, and `automation_run_device_results.output` only ever held
the automation's own log lines (`"[info] Queued run_script action"`).

## The fix

**Persist.** `executeRunScriptAction` now mints a genuine `script_executions`
row and passes its uuid as the payload's `executionId`, so the existing result
handler persists stdout/stderr/exitCode — and inherits the terminal-status
guard, the #2434 secret redaction, and the stale-command reaper for free.
`org_id` is the **device's** org (a partner-wide automation owns no org of its
own, #2133). `trigger_type` gets a new `'automation'` value.

**Correlate.** New nullable bare `script_executions.automation_run_id uuid`
+ partial index. Deliberately no FK / no Drizzle `.references()`:
`schema/automations.ts` already imports `schema/scripts.ts`, so the reverse
reference would close an import cycle — mirroring the existing bare
`automation_runs.config_policy_id`. A side benefit is that `script_executions`
stays out of `automation_runs`' FK graph, so the children-before-parents
org-cascade ordering contract is untouched.

**Render.** `GET /automations/runs/:runId` attaches per-device `scriptResults`,
and the run-history expanded panel gains a collapsible per-script
stdout/stderr block.

### Races and volume, handled

- **Terminal-status stomp.** `queueCommandForExecution` delivers over the
  WebSocket *synchronously*, so a fast agent can drive the row `completed`
  (with its stdout) before we return. Both post-queue transitions are now
  guarded on the row still being `pending`; without that, an unguarded write
  would flip a completed row back to `queued` and the stale-command reaper
  would later mark it `timeout`.
- **Delivery vs. enqueue.** On actual delivery the row goes to `running` with
  a `startedAt`, mirroring `services/scriptExecution.ts`, so the UI can derive
  a duration. Otherwise it's merely `queued`.
- **Offline devices.** `queueCommandForExecution` refuses any non-`online`
  device, and automation runs do not filter by device status. Checking up front
  avoids minting one `failed` execution row per offline device per run — a
  daily automation over a fleet of asleep laptops would otherwise drown the
  device's 50-row script history and skew per-script success stats.
- **Abuse signal.** `resource.volume_outlier`'s `scripts_24h` counted every
  execution per partner. One hourly automation across a fleet would now pin
  that signal for a perfectly ordinary MSP, so it excludes
  `trigger_type = 'automation'` — scheduled machine activity with no human
  behind it.

### Contracts

- `CORE_TENANT_EXPORT_POLICY` updated — the one cascade list that fires on a
  **column** add to an already-registered table. `script_executions` was
  already in `CORE_ORG_CASCADE_DELETE_ORDER` and both device lists; no FK means
  no new ordering constraint.
- RLS unaffected: automation runs execute inside `withSystemDbAccessContext`,
  under which `breeze_has_org_access` short-circuits, so the INSERT WITH CHECK
  policy passes. The new read inherits the same org policy as the existing
  device-results query.
- `ALTER TYPE ... ADD VALUE` is the only statement in migration `-a-`, per the
  per-file-transaction rule. `-b-` adds the column + index. Both sort after
  `2026-08-13-*`.
- The execute-script **request** schema still accepts only the original four
  trigger types — `'automation'` is provenance the runtime mints, never
  something an API caller can forge. The `ScriptExecution` **response** schema
  (openapi + `packages/shared`) was widened.

## Verification

- `apps/api`: 422 passed / 2 skipped across the affected suites
  (`automationRuntime.*`, `routes/automations`, `routes/agentWs`,
  `routes/scripts`, `routes/devices/scripts`, `abuseSignals/`,
  `db/autoMigrate`), incl. 9 new `automationRuntime.runScript.test.ts` cases.
- `apps/web`: 107 passed (`components/automations/` incl. 4 new script-output
  cases, `no-silent-mutations`) + 96 i18n gate tests.
- `tsc --noEmit` clean for `apps/api`, `apps/web`, `packages/shared`.
- Not run locally (no DB): the RLS/integration contract suites and
  `db:check-drift`.

## Known follow-ups (deliberately out of scope)

- The `execute_command` action's output is still discarded — it runs ad-hoc
  content with no `scripts` row, so it can't have an execution row, and needs
  a different persistence target. The run-history panel will show output for
  `run_script` actions and nothing for `execute_command` ones.
- `staleCommandReaper` times automation rows out on a fixed 300s rather than
  the script's own `timeoutSeconds`, and doesn't reconcile
  `automation_run_device_results`.
- Per-script stats (`aiToolsScripts.ts`) now blend manual and automation runs
  with no way to split them.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer (plus one independent design review before the PR was opened).

**Findings:** 0 CRITICAL. 8 IMPORTANT + 6 factually-wrong comments raised → all addressed in `ebdb0a7`; 0 outstanding. The substantive ones:

- **`execute_command` kept a synthetic executionId for a reason that turned out to be false.** I had commented that the Go agent requires the field; it does not — `handleScript` keys the execution on `cmd.ID`, and only `script_cancel` reads `executionId`. So the field is now dropped entirely, removing the last producer of non-uuid ids and the per-device-per-run warn spam.
- **The offline pre-check was wrong.** It read the run's device snapshot (taken once at the top of a fleet run, so minutes stale), silently skipping devices that came online mid-run — and diverging from `execute_command`, which still got the live check. Replaced: the row is minted, and *discarded* if the command never reaches the queue. Same "no row for an offline device" outcome, no stale read.
- **A queue throw stranded the row.** `queueCommandForExecution` does several DB round-trips plus a JIT decrypt and can throw rather than return `{error}`, leaving the row at `pending` for the reaper to relabel `timeout` — "no response from agent", when no agent ever had it. Now wrapped.
- **The run-detail response inlined uncapped stdout.** It is re-polled every few seconds while a run is live and `stdout` is accepted up to 5MB per execution. Now truncated in SQL (16k/8k) with explicit `stdoutTruncated`/`stderrTruncated` flags.
- **The UI would have shipped the same ambiguity the issue is about.** A `pending`/`queued`/`running` execution rendered the *empty-output* placeholder — indistinguishable from "the script printed nothing". And because a run goes terminal as soon as its commands are queued, the parent's poll had already stopped by the time stdout arrived, so the first view froze until you collapsed and re-expanded the row. Both fixed.
- **The catch that hid this bug is no longer blind.** `handleScriptResult`'s `catch` now calls `captureException` like its SNMP sibling.
- **Six of my own comments were verifiably wrong** (agent requirement, reaper joining `device_commands`, "LEFT JOIN" vs equality filter, `queued` vs `running` as the stompable state, "mirrors the manual path", and the abuse-signal rationale — the `commands_24h` axis is unfiltered, so excluding automations from the scripts axis does *not* de-noise the signal). All corrected.

**Tests:** `apps/api` 426 passed / 2 skipped across the affected suites (`automationRuntime.*`, `routes/automations`, `routes/agentWs`, `routes/scripts`, `routes/devices/scripts`, `abuseSignals/`, `db/autoMigrate`). `apps/web` 207 passed (`components/automations/`, i18n gates, `no-silent-mutations`). `tsc --noEmit` clean for `apps/api`, `apps/web`, `packages/shared`.

**⚠️ CI has not run — GitHub Actions is in a major outage.** The `pull_request` trigger dropped for this branch (no workflow run exists), `gh workflow run` returns HTTP 500, and other PRs' runs are queued 1h+. `gh pr checks` shows only Cloudflare Pages, which reads misleadingly like a short green list. **Please confirm a real CI run before merging.**

**Not verified locally:** the RLS/integration contract suites and `db:check-drift` — they need a live DB, and the Postgres containers on this machine belong to other worktrees. `CORE_TENANT_EXPORT_POLICY` gained `automation_run_id` (the one cascade contract that fires on a *column* add), and that suite is non-blocking on PRs but required on main, so it's worth running before merge.

**Known follow-ups, deliberately out of scope:**
- `execute_command` output is still discarded — ad-hoc content with no `scripts` row, so it can't have an execution row and needs a different target.
- A BullMQ `execute-run` retry re-executes actions and mints duplicate execution rows (`executeAutomationRunInner` has no run-status short-circuit). Pre-existing for `device_commands`, but these rows are now user-visible.
- `staleCommandReaper` times automation rows out on a fixed 300s rather than the script's own `timeoutSeconds`, and doesn't reconcile `automation_run_device_results`.

**Status:** review-clean, awaiting maintainer merge — gated on a real CI run once Actions recovers.

PR #3008, opened 2026-08-01
fix(backup): register snapshot id mid-run and add an orphan-snapshot reconcile path (#3006)

ToddHebebrand:
Refs #3006

## The problem

A backup uploads `snapshots/<id>/manifest.json` plus `files/...` to the customer's destination and only *then* sends its terminal result. `providerSnapshotId` was persisted **only** on a `completed` result (`backupResultPersistence.ts`). When that result is lost in transit (#3001 oversized IPC payload, #2998 helper death) the `backup_jobs` row never learns the snapshot id, so nothing links the uploaded objects to a restore point. The reaper then marks the job `failed` — doubly wrong, because the backup *succeeded* and its output *still exists*.

It is worse than invisible: retention keys off `backup_snapshots` rows, so an orphan is **neither restorable nor reaped** and accrues storage cost forever.

Confirmed in the field 2026-08-01: a self-hosted operator uploaded 123,284 files / 35.4 GB to S3, and nothing appears in the UI.

## Part 1 — structural fix (new orphans become impossible)

The agent knows `snapshotId` at snapshot-creation time, long before the run finishes.

- `ProgressFn` gains a trailing `snapshotID`; `createSnapshotWithProgress` forces one **unthrottled** emission the instant the id is minted — *before the first byte uploads* — and every later emission carries it too.
- `backupipc.BackupProgress` forwards it as `snapshotId` (`omitempty`).
- Server-side `applyBackupProgress` writes `backup_jobs.snapshot_id` mid-run.

A lost terminal result now still leaves the job pointing at the objects it wrote. Pre-snapshot emissions (keepalive, "scanning done" totals) pass `""` and the field is omitted, so the server can distinguish "no id yet" from an empty id.

A malformed `snapshotId` drops **only that field**, never the payload — losing the payload would also lose `lastProgressAt` and the dispatch-expectation refresh, so an id format that outgrew the column would reap healthy uploads fleet-wide *and* suppress the registration this field exists to provide. `agentWs` warns so a broken agent build is visible.

The agent-side change stays clear of the contended regions of `backup.go` (the `BackupJob` struct and the completion tail) — it touches only the "scanning done" emit and the pre-scan keepalive.

## Part 2 — recovery for existing orphans

`POST /backup/reconcile` (`{ configId, dryRun?, limit? }`) enumerates a destination's manifest-bearing snapshot prefixes and adopts unclaimed ones into restore points.

[...]

PR #3193, opened 2026-08-06
fix(api): make the closed 2026-08-06 migration block fail at authoring time (#3016)

### LINKED ISSUE #3016: [API] Reserved 2026-08-06 migration block is undiscoverable at authoring time and its guard doesn't name the offending file — two PRs hit it in one day
## Description

The `2026-08-06` migration date block is reserved for the security remediation waves, and `autoMigrate.test.ts` enforces slots `-a-` through `-f-` only:

```js
const RESERVED_MIGRATION_DATE = '2026-08-06-';
const RESERVED_SLOT_PATTERN = /^2026-08-06-[a-f]-/;
```

**Two unrelated PRs took slot `-g-` within hours of each other**, both reddening `Test API`:

| PR | migration | outcome |
|---|---|---|
| #2995 | `2026-08-06-g-drop-custom-alert-conditions.sql` | merged → **main red ~1 h** → fixed by #3015 |
| #3008 | `2026-08-06-g-backup-jobs-snapshot-id-index.sql` | caught on the PR |

Neither author was doing anything unreasonable. The convention documented in CLAUDE.md is "insert an explicit `-a-`/`-b-` infix between the date and the slug" for same-day ordering — so reaching for the next free letter after `-f-` is the *natural* reading. Nothing at authoring time says `2026-08-06` is special or that the letters stop at `f`.

## Why the failure is hard to act on

The guard reports the violation indirectly. Three tests fail, none of which name the offending file, and all with the same opaque message:

```
× discovers the report site-scope migration exactly once in lexical order
× orders the reserved live-authorization migrations after all preceding migrations
× orders the certificate-history migration after the reserved Wave 3 quote-capability migration

AssertionError: expected [ Array(1) ] to deeply equal []
```

`[ Array(1) ]` is the violating filename, but vitest elides it. Diagnosing it requires reading `expectReservedBlockSlotsOnly`, finding `RESERVED_SLOT_PATTERN`, and comparing it against the migrations directory by hand. Both PRs needed a manual bisect (delete each new migration, re-run) to identify the culprit.

## Suggested fixes

Cheapest first; (1) alone would probably have prevented both incidents.

1. **Make the assertion name the file.** Assert on the violating names with a message, e.g.
   `expect(violations, \`migration(s) squatting in the reserved ${RESERVED_MIGRATION_DATE} block — use a date outside it\`).toEqual([])`.
   One line, and the failure becomes self-diagnosing.
2. **Fail at authoring time, not in `Test API`.** A `README` in `apps/api/migrations/` is easy to miss; a pre-commit or lint check that rejects a new `2026-08-06-*` file that is not `[a-f]` is not.
3. **Document the reserved block in CLAUDE.md** next to the same-day `-a-`/`-b-` infix guidance, which currently implies the letters are free-running.
4. **Reconsider whether a date-based reservation is the right mechanism at all.** It is a calendar date used as a namespace, so it collides with anything legitimately dated that day, and it silently expires the moment a wave needs a seventh slot. A dedicated prefix or a manifest of wave-owned filenames would not have this failure mode.

## Related

- #3015 — un-reds main by renaming #2995's migration to `2026-08-08-`
- #3008 — same mistake, caught pre-merge

ToddHebebrand:
Closes #3016

Three separate authors reached for `2026-08-06-g-` in one day — #2995 (merged, reddened `main` ~1h, fixed by #3015), #3008 (caught on the PR), and the m365-comms plan doc, which instructs its executor to do the same. None was being careless: the documented same-day-ordering convention is literally *"insert an explicit `-a-`/`-b-` infix"*, so reaching for the next free letter is the natural reading. Nothing at authoring time said the letters stopped at `-f-`.

Two problems, addressed separately.

## 1. Undiscoverable at authoring time

| Added | What it does |
|---|---|
| `scripts/check-migration-naming.sh` | Rejects additions to the closed block. Also rejects filenames that miss the runner's `^\d{4}-.*\.sql$` discovery pattern — today those aren't rejected, they're **silently never applied**. |
| `.githooks/pre-commit` | Runs the guard on staged-added files, alongside the existing confidential scan. This is where an author should hit it. |
| `.github/workflows/ci.yml` (`check-migrations`) | Whole-directory run as the non-bypassable backstop for `--no-verify`. |
| `apps/api/migrations/README.md` | Naming, closed-block table, idempotency, no inner `BEGIN;`/`COMMIT;`, immutability + the rename/reference-sweep trap, and a new-migration checklist — in the directory you're already in when authoring. |
| `CLAUDE.md` | Documents the closed block right next to the same-day infix guidance that implied the letters were free-running, and states that infixes are **per-date**, not a global sequence. |

The guidance everywhere is the same: **if you need to sort after everything shipped, use a later DATE, not a later letter.** Today's date already sorts last — that's the property authors actually want, and it collides with nothing.

## 2. The guard didn't name the offending file

Before, a squatter failed **three** wave-ordering tests, each named for a wave rather than the problem, all emitting `AssertionError: expected [ Array(1) ] to deeply equal []`. Both incidents needed a manual bisect (delete each new migration, re-run) to identify the culprit.

Now it fails exactly one test, named for the rule:

```
FAIL  src/db/autoMigrate.test.ts > migration filename conventions >
      adds no new migration to the closed 2026-08-06 reserved block
AssertionError: Migration(s) added to the CLOSED reserved 2026-08-06- block:
  - 2026-08-06-g-drop-custom-alert-conditions.sql

That date is not a free namespace and its letters do NOT run past the
shipped set. Rename the file to a date AFTER the block ... and update every
reference to the old path — integration tests replay migrations BY PATH, so
a stale name is an ENOENT in Integration Tests, not a compile error.
See apps/api/migrations/README.md.
```

**Regex → frozen manifest** (issue suggestion 4, cheap version). `RESERVED_SLOT_PATTERN = /^2026-08-06-[a-f]-/` was wrong in both directions: it silently blessed `2026-08-06-a-something-unrelated.sql`, and it *read* as an open range with `-g-` free for the taking. The block is now closed and fully shipped, so the honest expression is an explicit list of its eight filenames. It also catches the reverse — a shipped block member deleted or renamed.

## 3. Companion assertion (from the issue thread)

`resolves every core migration path referenced from apps/api/src` — integration suites replay migrations **by path**, so a filename is executable code, not prose. In #3015 the rename was easy but the reference sweep wasn't: it surfaced as an `ENOENT` in Integration shard 3/4, minutes in, well after `Test API` had gone green. Verified by renaming `2026-08-08-drop-custom-alert-conditions.sql` locally:

```
AssertionError: Migration path(s) referenced from apps/api/src that do not exist:
  - 2026-08-08-drop-custom-alert-conditions.sql (referenced from
    apps/api/src/__tests__/integration/customAlertConditionCleanup.integration.test.ts)
```

`apps/api/src/extensions/` is excluded — extensions ship their own per-bundle migrations dirs and their tests use synthetic in-memory filenames.

## 4. The queued third instance

`docs/superpowers/plans/integrations/2026-07-30-m365-comms-3-api-integration.md` told its executor to name migrations `2026-08-06-g-/-h-/-i-`. Renamed to `2026-08-14-a-/-b-/-c-` and the rationale corrected: the plan needs *sort-last*, which a later date gives it. Its "re-check at execution time" instruction is kept, but now says move the date rather than bump the letters.

## Deliberately not changed

`apps/api/migrations/2026-08-09-backup-jobs-snapshot-id-index.sql` has a comment naming the now-removed `RESERVED_SLOT_PATTERN`. Editing a migration to fix a comment is precisely the #2708 incident class, and the comment's substance ("don't put it in the 2026-08-06 block") is still correct.

Issue suggestion 4's larger form — replacing date-based reservation with a dedicated prefix — isn't needed: the block is closed and no future block is planned. If one ever is, the manifest shape generalizes.

## Verification

- `vitest run src/db/autoMigrate.test.ts` — **59 passed** (56 before + 3 new).
- Negative-tested both new guards: squatter file → 1 named failure; simulated rename → dangling-reference failure naming both files. Reverted after.
- `tsc --noEmit` on `@breeze/api` — clean.
- `check-migration-naming.sh` — OK in both modes on the current tree; fires correctly on a planted `2026-08-06-g-*.sql` and a planted `nodigits-bad.sql`.
- `check-migration-immutability.sh` — OK (no shipped migration touched).
- Pre-commit hook exercised live by this commit.

No migrations added or edited; no runtime code touched.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter (4 agents, parallel).

**Findings:** 6 raised → all 6 addressed in f429b70f9; 0 outstanding.

1. **Manifest drift guard was near-vacuous** (flagged independently by all three code reviewers). `keeps the commit-time naming guard pinned to the same reserved date` only asserted the `.sh` contained the date *substring* — code-reviewer proved a bogus 9th entry injected into the shell manifest still passed. Now parses the bash array and asserts element-for-element equality with `RESERVED_BLOCK_MIGRATIONS`, plus an anchored regex on `RESERVED_DATE`. Re-ran that exact repro: it now fails as it should.
2. **Whole-directory mode passed vacuously on an empty directory** — without `nullglob` the unmatched glob expands to the literal pattern, `[ -f ]` skips it, and the guard printed `OK` having inspected nothing. That is precisely the failure class this PR exists to prevent. Now `nullglob` + hard-fail on zero files.
3. **`git diff --cached` failure was invisible to `set -e`** inside the process substitution, yielding an empty file list and a silent pass. Captured into a variable with an explicit failure branch.
4. **The guard's failure path was never exercised** — CI only ever ran it against a clean tree, proving the pass branch alone; the negative testing was manual. Added a test driving the script via `spawnSync` against a fixture dir (`BREEZE_MIGRATIONS_DIR`, an override nothing in CI or the hook sets), asserting exit 1 + the offending filename in stderr for a squatter and an undiscoverable filename, and exit 1 for the empty directory.
5. **"One migration per wave" / "the eight security-remediation wave migrations" was false for 2 of 8** — and contradicted by this same test file's own Wave 6 comments: `-e-action-intents-origin-principal` and `-f-m365-comms-delegated` are unrelated same-day work. Corrected in all four places. The block is closed for its inter-file ordering dependencies, which is both accurate and sufficient. Also dropped "later migrations **already** re-create objects it defines" — no shipped later migration does; the known case is the still-unwritten m365 plan-doc migration, now stated prospectively.
6. **Reference scan over-excluded `apps/api/src/extensions/`**, skipping a real core-migration citation in `extensions/stateStore.ts`. Narrowed to extension `*.test.ts` only (the files using synthetic fixture names).

**Tests:** `vitest run src/db/autoMigrate.test.ts` — **60 passed** (56 on main + 4 new). `tsc --noEmit` on `@breeze/api` — clean. `check-migration-naming.sh` OK in both modes; `check-migration-immutability.sh` OK (no shipped migration touched).

**CI:** the `pull_request` trigger dropped for this branch — `gh pr checks` showed only a Cloudflare Pages entry, which reads deceptively green. Dispatched CI by hand: [run 31127084687](https://github.com/LanternOps/breeze/actions/runs/31127084687). **`Check Migrations` — the job carrying this PR's new guard step — passed**, along with `Lint`, `Type Check`, `Check Migrations (non-superuser)`, and `Lint Agent (Go)`. `Test API` (which carries `autoMigrate.test.ts`) was still queued behind the repo-wide runner backlog at the time of writing; it needs a look before merge, though that suite is green locally.

**Status:** review-clean, awaiting `Test API` on the dispatched run + maintainer merge.

Issue #3178, opened 2026-08-06
IPC Conn.Send() allocates seq/HMAC before acquiring write mutex, causing out-of-order frames


ToddHebebrand:
Reported by **andrepereira5963** in #bugs (Discord, message 1534903951735128246).

**Report (PII-redacted):**

<@911011690290774076> I found a definite IPC sequence race in agent/internal/ipc/protocol.go.

Conn.Send() increments sendSeq and computes the HMAC before acquiring the write mutex:

env.Seq = c.sendSeq.Add(1)
env.HMAC = c.computeHMAC(env)
...
c.mu.Lock()

Two concurrent senders can therefore allocate seq 85 and 86, but acquire the mutex in reverse order and transmit 86 before 85. The helper then correctly reports:

sequence number 85 <= last 86 (replay/duplicate)

Moving c.mu.Lock() before sequence allocation/HMAC calculation should guarantee that sequence order matches wire order.

This explains the replay/duplicate errors. It may not fully explain the separate message too large: 2065852772 error, where 2065852772 == 0x7B226964 == '{"id', but the IPC stream is clearly experiencing at least this ordering defect.

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

## Triage analysis

Verified in agent/internal/ipc/protocol.go: Conn.Send() allocates the sequence number and computes the HMAC (env.Seq = c.sendSeq.Add(1); env.HMAC = c.computeHMAC(env)) at lines 134-135, before acquiring the write mutex c.mu.Lock() at line 146. Since sendSeq is a lock-free atomic counter, two goroutines calling Send() concurrently on the same Conn can allocate sequence numbers in one order (e.g. 85 then 86) but be scheduled onto the mutex and onto the wire in the opposite order. The receiver's Recv() (lines 240-248) enforces strictly-increasing sequence numbers and rejects out-of-order frames with exactly the error text the reporter quotes: 'ipc: sequence number %d <= last %d (replay/duplicate)'. I confirmed concurrent Send() calls on one Conn are a real, tested scenario (protocol_test.go TestConnSendStalledDoesNotStarveOtherWriter deliberately fires two goroutines' Send() against the same Conn for the write-timeout/starvation fix), but no test asserts that seq order matches wire order under concurrency — that gap is exactly what the reporter's proposed fix (moving the lock before sequence allocation and HMAC computation) would close. The separate 'message too large: 2065852772' symptom is not explained by this and remains a distinct, unconfirmed lead.

[...]

Issue #434, opened 2026-04-13
[Remote Desktop] Windows: viewer disconnects on user logout instead of handing off to login screen


ToddHebebrand:
## Summary
On Windows, when the remote user logs out during an active Remote Desktop session, the viewer disconnects instead of handing off to the loginwindow (Session 0 / default input desktop). Reconnecting works, but the handoff is not seamless. Intermittent — works some of the time.

## Repro
- Target: WIN-DHQNR1F8LO2 (US DigitalOcean droplet, Windows Server)
- Connect viewer to an active user session
- Start → Sign out
- Expected: viewer reattaches to the login screen via the helper respawn path (same behavior we want on macOS per #330)
- Actual: viewer session drops; user must click Connect again

## Notes
- Windows-only. macOS equivalent is tracked in #330 / #331.
- Related closed Windows issue: #387 (Session 0 reconnect-loop on headless servers) — different symptom but same respawn/handoff subsystem.
- Intermittent — need to capture agent + helper logs from a failing run to tell whether the helper dies, the WS drops, or the broker fails to re-spawn into the new session.

## TODO for investigation
- [ ] Reproduce on WIN-DHQNR1F8LO2 and collect agent + userhelper logs
- [ ] Confirm broker receives WTS session-change event on logout
- [ ] Check whether sessionbroker.spawner re-spawns helper into new (loginwindow / Session 1) context or stalls
- [ ] Compare against working case — is it timing (helper spawn beats WS drop) or state (stale *tls.Config / cert)?

PR #3194, opened 2026-08-06
fix(agent): give every start_desktop its own IPC id so it stops self-colliding (#3107)

### LINKED ISSUE #3107: [Agent] start_desktop self-collides on its shared IPC id — both retries burn in the same millisecond and the user is told "helper keeps crashing" when the helper is healthy
## Summary

A remote-desktop start can fail with the user-facing error **`desktop start failed after 2 attempts (helper keeps crashing)`** while the helper is perfectly healthy and never crashed.

The real cause is `start_desktop` colliding with *itself*: it is deliberately exempt from command dedup, but every concurrent invocation uses the same IPC command id `desk-<sessionID>`, so the second one is rejected by the `ErrDuplicateCommand` guard. The caller misclassifies that rejection as helper death, burns both retry attempts in the same millisecond, and reports a crash that did not happen.

Observed on a production Windows 11 Pro 25H2 workstation (build 26200.6901), agent `0.103.0`, always-on helper lifecycle mode. Tenant identifiers redacted.

## Evidence

Agent logs, all within 24 s. Note both retry attempts share the *same millisecond and the same error*:

```
03:30:12.994 warn heartbeat  IPC desktop start failed, will retry with new helper
    attempt=1  error="duplicate in-flight command id: \"desk-<REDACTED>\" (session \"helper-<REDACTED>\")"
03:30:12.994 warn heartbeat  IPC desktop start failed, will retry with new helper
    attempt=2  error="duplicate in-flight command id: \"desk-<REDACTED>\" (session \"helper-<REDACTED>\")"
03:30:34.981 warn heartbeat  dropping desktop peer disconnect for non-owned session
```

`attempt=1` and `attempt=2` are the same goroutine, so a *different* caller already held `desk-<sessionID>` in flight. The viewer session ended `failed`; the user saw only "helper keeps crashing".

## Root cause

**1. `start_desktop` is exempt from dedup, but its IPC id is not unique per invocation.**

`internal/heartbeat/heartbeat.go:5412` deliberately exempts the command (#434), because the viewer legitimately re-invokes the same `commandId` across reconnects:

```go
dedupable := cmd.Type != tools.CmdStartDesktop &&
    cmd.Type != tools.CmdStopDesktop &&
    cmd.Type != tools.CmdDesktopStreamStop
```

Commands arrive over both WebSocket and the heartbeat response and run on the same worker pool, so two invocations can execute concurrently. Both derive the same IPC id `"desk-"+sessionID` (`internal/heartbeat/handlers_desktop_helper.go:186`), and `SendCommand` rejects the second (`internal/sessionbroker/session.go:109`):

```go
var ErrDuplicateCommand = fmt.Errorf("duplicate in-flight command id")
```

So the #434 exemption and the `ErrDuplicateCommand` guard are in direct conflict: the exemption exists to *allow* re-invocation, and the guard *rejects* it.

**2. Any `SendCommand` error is misclassified as helper death.**

`internal/heartbeat/handlers_desktop_helper.go:186-190` returns `helperDied = true` for every error — duplicate-id rejection and the 30 s `ErrCommandTimeout` included:

```go
resp, err := session.SendCommand("desk-"+sessionID, ipc.TypeDesktopStart, req, 30*time.Second)
if err != nil {
    return tools.NewErrorResult(err, 0), true   // <- "helper died", for ANY error
}
```

**3. The retry loop cannot fix anything, and on Windows cannot acquire a new helper at all.**

`handlers_desktop_helper.go:159-178` retries twice with no backoff, no kill, and no exclusion of the session that just failed — `FindCapableSession` filters only on scope and `CanCapture`, so it returns the same session. In the duplicate-id case both attempts fail instantly against a healthy helper.

Separately, on Windows always-on the "retry with new helper" never spawns one: `spawnHelperForDesktop` ends at `handlers_desktop_helper.go:652` with `no lifecycle-owned helper is connected ... waiting for lifecycle reconciliation`, then polls for 10 s — but `HelperLifecycleManager` reconciles on a **30 s** interval (`internal/sessionbroker/lifecycle.go:15`) with exponential respawn backoff. **The retry window is structurally shorter than the respawn cadence**, so a genuinely crashed helper can essentially never be replaced inside the loop either.

**4. Failure path never records ownership, so teardown never runs.**

`rememberDesktopOwner` is only reached on success (`handlers_desktop_helper.go:202`). Because the start failed, `desktopOwners` has no entry, so the helper's genuine `desktop_peer_disconnected` is dropped at `internal/heartbeat/heartbeat.go:909` and no teardown/notify fires — the viewer just times out instead of being told the session died.

## Suggested fixes

1. **Make the IPC id unique per invocation** (e.g. `desk-<sessionID>-<attempt/nonce>`) so re-invocation cannot self-collide, or serialize `startDesktopViaHelper` on the desktop `sessionID` with an in-flight guard. This is the core fix — it resolves the #434 vs `ErrDuplicateCommand` conflict.
2. **Stop treating `ErrDuplicateCommand` (and arguably `ErrCommandTimeout`) as `helperDied`.** A duplicate rejection means a start is already in flight; the correct behaviour is to join or defer to it, not to declare the helper dead.
3. **Fix the operator-facing message.** "helper keeps crashing" should not be emitted unless the helper actually died; surface the underlying error instead.
4. Align the retry wait with `reconcileInterval`, or have the retry path kick `HelperLifecycleManager` (`kickCh`) rather than passively polling for 10 s.
5. Record desktop ownership (or a tombstone) on the failure path so `desktop_peer_disconnected` isn't dropped and the viewer gets a real error.

## Note for whoever picks this up

`internal/sessionbroker/broker.go:2883` `dropping unsolicited or unauthorized helper message` fires nearby in these logs and is a **red herring** — those are `notify_result` envelopes arriving after their `pending` entry was removed, dropped by `default: return false` in `shouldForwardUnsolicitedHelperMessage`. It is a late-reply drop, not an authorization problem.

ToddHebebrand:
## Problem

A remote-desktop start could fail with **`desktop start failed after 2 attempts (helper keeps crashing)`** while the helper was healthy and had never crashed.

`start_desktop` is deliberately exempt from the heartbeat's command dedup (#434), because the viewer legitimately re-invokes the same `commandId` across reconnects and the command can arrive over both the WebSocket and the heartbeat response. So two invocations for one desktop session can be in flight at once — but every invocation derived the **same** IPC correlation id `desk-<sessionID>`, so the second was rejected by `Session.SendCommand`'s `ErrDuplicateCommand` guard. The #434 exemption and the duplicate guard were in direct conflict.

The caller then compounded it in two ways:

- `startDesktopOnSession` returned `helperDied = true` for **every** `SendCommand` error, so the duplicate-id rejection was misclassified as helper death.
- The retry loop had **no backoff**, so both attempts burned in the same millisecond against the same still-healthy helper — exactly the reported log signature, where `attempt=1` and `attempt=2` share a timestamp and an error.

## Changes

All in `agent/internal/heartbeat/handlers_desktop_helper.go`.

**1. Unique IPC id per invocation.** `nextDesktopStartCommandID()` numbers each start off an `atomic.Uint64` — `desk-<sessionID>-<n>`. The desktop session id stays in the string so helper-side logs remain greppable. This resolves the #434-vs-duplicate-guard conflict, and as a bonus fixes a latent mis-correlation: with the shared id, a response from a timed-out attempt could be matched to a later attempt's pending entry and pass the response validator, handing the viewer an SDP answer for a peer connection that had already been torn down.

**2. Single-flight per desktop session** (`joinOrRunDesktopStart`). This is load-bearing, not a nicety. `ErrDuplicateCommand` was *also* the thing accidentally serializing concurrent starts; with unique ids both would reach the helper, and `SessionManager.StartSession` (`remote/desktop/session_webrtc.go:41-58`) unconditionally stops **every** existing session before registering the new one. So the second start would tear down the first, both callers would be told `completed`, and only one peer connection would be live. Worse, the first session's `OnConnectionStateChange` closure captures the session **id** rather than the `*Session`, so its late `Closed` callback calls `StopSession(sessionID)` and kills the **second** session, firing a spurious `desktop_peer_disconnected`. A joiner now takes the leader's result instead — "join or defer to it", as the issue asked.

**3. Honest failure classification.** `desktopStartLostHelper()` replaces the blanket "any error means the helper died":

| Error | Classified as | Why |
|---|---|---|
| `ErrDuplicateCommand` | not a death, terminal | Another start is in flight against a **live** helper. Nothing crashed and a retry cannot help. Unreachable after changes 1-2 — kept as a guard against a future caller reintroducing a shared id. |
| `ErrCommandTimeout` | not a death, terminal | A helper that actually dies closes the session's `done` channel and surfaces as *session closed while waiting for response*, not as a timeout. Terminal because `helperSessionForTarget` does no liveness or pong-age check and would hand the retry back the very same session. |
| everything else | helper lost, retry | Failed socket write, session closed under us — a real transport failure. The default stays conservative. |

**4. Spaced retries + operator-facing messages.** Retries are separated by `desktopStartRetryBackoff` (1s; a `var` so tests can shrink it, following the existing `desktopLeaseRenewEvery` pattern), selecting on `stopChan` so a pending retry never holds shutdown open. `agentWs` writes `CommandResult.Error` into `remote_sessions.errorMessage` and the viewer displays it, so both sentinels are now wrapped with a real message plus the helper/windows session ids (keeping `%w`) instead of surfacing as bare `sessionbroker: command timed out`. The terminal message carries the underlying cause instead of blaming a crash unconditionally, and the `no capable helper available` exit carries it too.

`startDesktopOnDemand` (the RDS path) picks up all of this for free — it shares `startDesktopOnSession` and runs inside the same single-flight.

## Tests

New `agent/internal/heartbeat/handlers_desktop_helper_collision_test.go`. All three fix parts are **mutation-verified** — reverting each one fails the suite:

| Mutation | Caught by |
|---|---|
| unique id → shared `desk-<sessionID>` | `TestConcurrentStartDesktopOnSessionDoesNotSelfCollide`, with exactly the reported error `duplicate in-flight command id: "desk-desktop-1"` |
| `desktopStartLostHelper(err)` → `true` | `TestStartDesktopViaHelperDoesNotRetryWedgedHelper` |
| single-flight bypassed | all four `TestJoinOrRunDesktopStart*` |

Plus `TestDesktopStartLostHelper` (table-driven over the classification matrix, bare and wrapped sentinels), `TestNextDesktopStartCommandIDIsUniquePerInvocation`, `TestStartDesktopViaHelperSpacesRetriesAndReportsRealError`, `TestJoinOrRunDesktopStartReleasesOnPanic`, and `TestJoinOrRunDesktopStartNeverRunsTwoAtOnce` — which asserts the invariant no scheduling order can break (at most one start executing per session) rather than a count that could flake.

Verification: `go test -race ./...` green across the whole agent module; the new tests run 8/8 clean repeatedly; `go build ./...`, `go vet`, `gofmt` clean on the touched files.

## Deliberately out of scope

Left as follow-ups, noted so they are not lost:

- **Reaping a start whose outcome was never proven.** `SendCommand`'s timeout does not cancel helper-side work, so a timed-out start may still bring up a capture that nothing here owns. A compensating `desktop_stop` was implemented and then **removed**: the helper processes stop off `StartSession`'s mutex, so a stop queued behind a wedged helper can land after the viewer's next start and silently kill a session the daemon believes is live — the same class of bug as this issue. Doing it safely needs a helper-side start generation so a stale stop cannot match a newer session. Meanwhile the orphan is self-limiting: the viewer never gets the answer, ICE never completes, and the peer connection tears itself down on the 15s failed timeout.
- **Ownership on the failure path** — `rememberDesktopOwner` is still only reached on success, so a genuine `desktop_peer_disconnected` after a failed start is dropped as "non-owned session". Pairs naturally with the item above.
- **Retry window vs. respawn cadence** — the retry still cannot outlast `HelperLifecycleManager`'s 30s `reconcileInterval`, so a genuinely crashed helper can't be replaced inside the loop. Aligning the wait or kicking `kickCh` is a lifecycle-manager change with its own risk profile.

Also untouched by design: the adjacent `ipc/protocol.go` `Conn.Send` seq/HMAC concern (#3007 / #3178). On this base that critical section is already held under `c.mu`, and this PR goes nowhere near the send path.

Closes #3107


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1: code-reviewer, pr-test-analyzer, silent-failure-hunter (all three in parallel). Round 2: code-reviewer, scoped to the round-1 fix delta only (a new concurrency primitive in agent code justified one re-review).

**Findings:** 8 raised across two rounds → all 8 addressed; 0 outstanding.

Round 1 (3 consequential):
- **Critical — the fix unblocked a worse bug.** `ErrDuplicateCommand` was also what accidentally *serialized* concurrent starts. With unique ids both reach the helper, and `SessionManager.StartSession` stops **every** existing session before registering the new one, so start #2 tears down start #1 while both report `completed` — and #1's `OnConnectionStateChange` closure captures the session *id*, so its late `Closed` callback kills #2. Fixed by adding `joinOrRunDesktopStart` (single-flight per desktop session) in `8ec8ad8`. Verified the helper-side claim directly against `remote/desktop/session_webrtc.go` before acting on it.
- **Critical — the classifier wasn't pinned.** Mutating `desktopStartLostHelper(err)` → `true` at the call site passed the entire package, and `TestStartDesktopViaHelperDoesNotRetryDuplicateCommand` exercised the `resp.Error` branch rather than the behaviour in its name. Extracted `desktopStartCommandTimeout` as a shrinkable var and replaced that test with `TestStartDesktopViaHelperDoesNotRetryWedgedHelper`, which kills the mutant.
- **High — raw sentinels reached the technician.** `agentWs` writes `CommandResult.Error` into `remote_sessions.errorMessage` and the viewer shows it, so the most likely terminal case displayed `sessionbroker: command timed out`. Both sentinels now carry an operator-facing message plus the helper/windows session ids, keeping `%w`.

Round 2 (3 defects, all in the round-1 fix itself — 2 reverted rather than patched, in `0e48e38`):
- **Critical — the orphan reap escaped its own single-flight** and could land a stale `desktop_stop` after the viewer's next start, silently killing a session the daemon believed live. **Removed** rather than patched: the reap fires exactly when the helper is wedged, which is exactly when the stop sits queued, so a synchronous variant wouldn't close the window either. The orphan is self-limiting (ICE never completes → 15s failed timeout tears it down), so a bounded self-healing orphan beats an unbounded silent kill. Follow-up needs a helper-side start generation.
- **Important — the different-offer "take your own turn" branch was deterministically wrong on RDS.** Leases are keyed per desktop session and share one owner, so the leader's failure path released the lease out from under the deferred caller, which then streamed unleased and got its helper reaped ~2 min in. **Removed** — a concurrent *different*-offer start for one desktop session isn't reachable in practice, and this also deleted the turn loop and its cap.
- **Important —** a panic in `run()` handed joiners a zero-valued `CommandResult` (Status `""`, no error text). `call.result` is now pre-set to a failure.

**Tests:** `go test -race ./...` green across the whole agent module. All three fix parts are mutation-verified — reverting the unique id, bypassing the classifier, or removing the single-flight each fails the suite. New tests 8/8 clean over repeated runs; the concurrency test asserts "at most one start executing per session", an invariant no scheduling order can flake. `go build ./...`, `go vet`, `gofmt` clean on touched files.

**CI:** the `pull_request` trigger did not fire for this branch — `gh pr checks` showed **only** a Cloudflare Pages entry, which reads green but means CI never ran. Dispatched by hand: [run 31127460219](https://github.com/LanternOps/breeze/actions/runs/31127460219) (`workflow_dispatch`), which picked up the full board including **Test Agent**, **Test Agent (race)**, and **Lint Agent (Go)**. It was still queued at hand-off — please confirm it went green before merging rather than trusting the PR check list.

**Status:** review-clean, awaiting maintainer merge. Not merged and #3107 left open per hand-off rules.

Issue #3034, opened 2026-08-03
Installer capacity is suppressed when an installer is built from a short-link child key


ToddHebebrand:
Split out of #2993 (review round 2). An accepted edge case of the fix landed there, recorded so it isn't rediscovered as a mystery.

## Background

#2993 added an `installerTokens: { consumed, max } | null` line to the Enrollment Keys rows, and gates it on `reportsInstallerCapacity()` (`apps/api/src/routes/enrollmentKeys.ts`): a key carrying a `short_code` reports **nothing**.

That gate exists because two flows parent `installer_bootstrap_tokens` and they don't mean the same thing:

- **Add Device / guided setup** mints ONE token whose `max_usage` *is* the operator's device count — a real capacity.
- **Short-link / installer-link** creates a CHILD key holding the count in its own `max_usage`, then mints a hardcoded `maxUsage: 1` token per public download. Summing those counts clicks, not device slots — a 7-device link clicked 3 times rendered `3 / 7` over `Installer devices 0 / 3`, a second smaller-looking N/M that grows on every click and never reaches 7.

## The edge case

The authenticated `GET /enrollment-keys/:id/installer/:platform` and `/:id/bootstrap-token` routes accept **any** key id the caller can reach — including a short-link child, which is a visible row on the Enrollment Keys page. An operator who builds an installer *from* that child row therefore creates a `max_usage > 1` token parented to a `short_code`-bearing key, and the gate suppresses a figure that was genuinely a capacity.

## Why it was accepted

- It needs a deliberate second download off a child row, not the normal path.
- The alternative is worse: summing one N-device token together with K per-download tokens yields a number with no meaning at all.
- An absent line beats a confidently wrong denominator.

## If it's ever worth fixing

The discriminator wants to be per-token rather than per-key — e.g. only aggregate tokens whose `max_usage > 1`, or record the mint path on the token row so per-download tokens are distinguishable from capacity tokens without inferring it from the parent. The latter is the honest fix; it needs a column and a migration.

Full rationale is in the `KNOWN EDGE CASE, accepted` paragraph on `reportsInstallerCapacity`.

ToddHebebrand:
Fixed in #3392 (open, not merged).

[...]

Issue #3039, opened 2026-08-03
[UI] Enrollment Keys installer capacity counts expired tokens — a dead installer can read as unused capacity


ToddHebebrand:
## Description

Follow-up polish on #2992 / PR #2993 (merged, not yet released). The new "X / Y" installer capacity figure on the Enrollment Keys list sums `consumed_count` / `max_usage` across **all** bootstrap tokens for the parent key — including expired ones. The API comment is explicit that the figure makes no redeemability claim.

Two confusing renderings result:

1. A fully expired installer shows e.g. **"0 / 7"** — reads as seven available slots, but no MSI from that download will ever enroll again.
2. The row's status badge reflects the **parent key** (which the Add Device flow mints with a 60-min TTL), so a key row can show *Expired* while its bootstrap token — the thing that actually enrolls — is live for months, and vice versa.

## Root Cause

- `apps/api/src/routes/enrollmentKeys.ts:679-699` (`fetchInstallerTokenUsage`) — `SUM(consumed_count)` / `SUM(max_usage)` grouped by parent, no liveness filter.
- Row status derives from the parent key's `expires_at`, which is decoupled from token lifetime since #2775/PR #2817.

## Proposed Fix

- Compute capacity from **live** (unexpired, unexhausted) tokens, or render both ("2 / 7 live, 5 expired").
- When a key has installer tokens, derive the row's effective status from token liveness instead of the transient parent's expiry.

## Affected Files

- `apps/api/src/routes/enrollmentKeys.ts` (`fetchInstallerTokenUsage`, list/detail routes)
- `apps/web/src/components/settings/EnrollmentKeyManager.tsx`

## Context

Found during the 2026-08-02 enrollment-system verification sweep. Related: #2992, #2775, #2832.

PR #2993, opened 2026-08-01
fix(api,web): show real installer capacity on Enrollment Keys (#2992)

ToddHebebrand:
Closes #2992

> **Reworked after review.** The first revision wrote the device count into the parent key's `max_usage`. That was rejected — see *Why not the one-line fix*. The shipped change is read-side only and touches no credential semantics.

## Problem

Downloading an installer for X devices from **Devices → Add Device → Download Installer** produced an Enrollment Keys row reading `usage 0 / 1`, not `0 / X`.

## Root cause

On the modern download paths — Windows, and macOS in its default `app-bundle` mode — the installer route mints **no child enrollment key**. It issues an `installer_bootstrap_tokens` row whose `max_usage` **is** the device count; the child key is minted lazily, one per device, at redemption (`routes/installer.ts:190-231`). The Enrollment Keys page reads only `enrollment_keys`, so it showed the transient parent's default cap of 1 and knew nothing about the installer.

## Why not the one-line fix

Setting `maxUsage: deviceCount` on the parent looks like a one-field fix and is not:

- **`max_usage` is an enforced enrollment budget, not a label.** `/agents/enroll` matches any key with `usage_count < max_usage` (`routes/agents/enrollment.ts:125`). Raising it 1 → N genuinely hands that key N direct-enrollment slots.
- **It doesn't even fix the number.** Nothing increments the parent's `usage_count` on this flow, so it turns `0 / 1` forever into `0 / 7` forever.
- **It overloads the column.** `usage_count` *is* atomically claimed on two other paths — the short-link `/s/:code` download (`enrollmentKeys.ts:2210`) and MCP invite redemption (`:330`), both 410-ing when exhausted.

Both mint sites now **pin** the rejected approach with `expect(createBody.maxUsage).toBeUndefined()` and the reason, so the tempting one-liner can't come back.

## Fix (read-side)

`GET /enrollment-keys` and `GET /enrollment-keys/:id` return, per key:

```ts
installerTokens: { consumed: number; max: number } | null
```

— **device slots** across the installers minted from that key, summed over its `installer_bootstrap_tokens` rows. `null` when the key never produced an installer, so the UI falls back to its own counters. The usage cell renders it as a second, muted line:

```
0 / 1
Installer devices 3 / 7
```

[...]

Issue #2992, opened 2026-08-01
Download Installer shows enrollment-key usage 0/1 instead of 0/X until installer is first run


ToddHebebrand:
Reported by **semotech** in #bugs (Discord, message 1533212756101042268).

**Report (PII-redacted):**

<@1489852429036949564> Flagging a bug: Running v0.103.0 and in "Devices", when pressing "Download Installer" to get a MacOS installer (.ZIP) for more than 1 device under "Enrollment keys" it shows "usage 0/1" instead of "usage 0/X" to match the X number of devices selected!  By contrast if one uses the "Generate Link" button or the "CLI commands" option with an X count of devices, the correct "usage 0/X" is shown. This seems to be a bug with just the "Download Installer" path.

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

## Triage analysis

[...]

PR #3196, opened 2026-08-06
fix(api): make the enrollment-key "Hide expired" filter agree with the status badge (#3191)

### LINKED ISSUE #3191: [UI][API] Enrollment Keys "Hide expired" filter is parent-only — hides rows the badge correctly shows as Active
## Description

Found while verifying #3033, which turned out to be already fixed by #3045.

#3045 fixed the enrollment-key **status badge** so that once a parent key is expired, row status derives from the live installer-token counts (`liveConsumed`/`liveMax`) rather than the parent's `expires_at`. A key whose parent has expired but whose installer is still live correctly renders **Active**.

The **"Hide expired" filter was not given the same treatment.** The web UI sends `?expired=false` and the list route filters purely on `enrollment_keys.expires_at`, so a row the badge shows as **Active** is hidden by the toggle.

Same parent-vs-token divergence #3045 fixed for the badge, still unfixed for the filter.

## Impact

An operator with "Hide expired" on cannot see a key that is genuinely still usable — the row is both Active and invisible.

## Proposed Fix

Make the list route's `expired` filter consult live installer-token capacity the same way `getKeyStatus` does, so the filter and the badge agree on what "expired" means.

## Notes

- The badge fix and its test coverage are in `apps/web/src/components/settings/EnrollmentKeyManager.tsx` (`getKeyStatus`) and the live-sum pair in `fetchInstallerTokenUsage` (`apps/api/src/routes/enrollmentKeys.ts`).
- Distinct from #3034, which is about short-link child keys suppressing capacity reporting entirely.

ToddHebebrand:
## The bug

PR #3045 taught the Enrollment Keys row badge (`getKeyStatus`) that once a parent key is dead the row is judged by its **installer tokens** — the things that actually enroll. So an Add-Device key whose 60-minute parent aged out while its 30-day installer link keeps working correctly renders **Active**.

The list route's `expired` filter never got the same treatment: it tested `enrollment_keys.expires_at` alone. Turning on **Hide expired** therefore hid rows the very same page was rendering as Active — the key was both Active and invisible, with no way for the operator to reach a credential that still genuinely works.

## The fix

The carve-out reuses the predicate that **already** spares such a key from `POST /enrollment-keys/purge-expired`. `services/enrollmentKeyPurgeGuards.ts` now exposes both wrappers over **one** correlated subquery:

- `hasNoLiveUnexhaustedBootstrapToken()` (existing, `notExists`) — unchanged emitted SQL for both purge consumers.
- `hasLiveUnexhaustedBootstrapToken()` (new, `exists`) — used by the list filter.

That is the exact SQL counterpart of the badge's `liveConsumed < liveMax` test. Sharing one definition means "Hide expired" can no longer hide a key that "Delete expired" deliberately refuses to delete, and a future change to the predicate cannot land on one path and skip the others.

**Invariant pinned:** no key the badge renders **Active** is hidden by `?expired=false`.

## Scope — deliberately narrow

- **Gated on `short_code IS NULL`** to mirror `reportsInstallerCapacity`: the API suppresses `installerTokens` for a short-link child (#3034), so its badge is parent-only and the filter must be too. **That gate itself is untouched** — #3034 remains its own accepted trade-off. The purge guard applies no such gate and shouldn't: suppressing a confusing capacity *number* is cosmetic, deleting a key out from under a live token is irreversible.
- **The exhaustion axis is not touched.** `usage_count >= max_usage` ("Exhausted", amber) has never been part of `?expired=` and still isn't; an exhausted-but-unexpired key lists exactly as before. Correcting the expiry axis alone is sufficient for the invariant above.
- **Both branches stay exact complements**, so no key becomes unreachable from both toggle positions.
- Both guards are built *inside* their branch — an unfiltered list request (the toggle is off by default) issues no extra `db.select`.

## Verification

New real-Postgres suite `apps/api/src/routes/enrollmentKeysExpiredFilter.integration.test.ts`, registered in both `vitest.integration.config.ts` (include) and `vitest.config.ts` (exclude). The mocked-`db` list suite returns whatever rows it is handed regardless of the predicate, so it cannot prove Postgres evaluates the correlated EXISTS per row — which is the entire fix.

Matrix covered: live unexhausted token (the regression), token itself expired, token fully consumed, no tokens, unexpired parent, never-expiring parent, short-link child, and `expired=true`/`expired=false` complementarity including `pagination.total`.

| Check | Result |
|---|---|
| New integration suite | 2 passed — and **confirmed red against the unpatched route** (not vacuous) |
| `enrollmentKeysPurgeExpired` + `enrollmentKeyCleanup` integration | 16 passed total across the three guard-sharing suites |
| Affected API unit suites (4 files, serial) | 128 passed |
| `apps/web` EnrollmentKeyManager | 19 passed (no web change; badge untouched) |
| `tsc --noEmit -p apps/api` | clean |

Closes #3191


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter (4 agents, parallel).

**Findings:** 11 raised → 10 addressed in 826e4a6, 1 deferred (below); 0 outstanding blockers.

All four reviewers independently converged on the same top finding, and it was a real one:

- **Clock split (all 4 agents).** `expired=true` bound a JS `new Date()` while `expired=false` used SQL `NOW()`. Under API↔DB skew a key expiring in that window matched **both** branches or **neither** — exactly the "reachable from neither toggle" failure the new partition test claims to rule out, and the comment asserting "exact complements" was therefore false. Both branches now read one `NOW()`.
- **`#3034` was miscited** as the origin of the `reportsInstallerCapacity` suppression. It is an **open issue arguing that suppression is wrong**; the origin is #2992 / PR #2993. Corrected, plus the coupling note that actually matters: if #3034 is ever fixed, the `short_code IS NULL` gate must move with it or #3191 comes straight back for those rows.
- **`#3045` → `#3039`** — the badge and route annotate that change by issue everywhere else, so grepping `#3045` in `EnrollmentKeyManager.tsx` returned nothing.
- **"iff" was false** (the next paragraph was its own counterexample) → "only if". **"exact SQL counterpart"** now names its two seams: the JS-vs-Postgres clock, and that `consumed_count <= max_usage` has **no DB CHECK** — it is upheld only by the conditional UPDATE in `routes/installer.ts`.
- **The invariant is now qualified.** `fetchInstallerTokenUsage` degrades to an empty Map on failure, dropping every badge back to parent expiry while this WHERE clause cannot degrade — so in that window the page can render an "Expired" row under "Hide expired". Documented rather than papered over; it is loud in Sentry, and degrading the row *set* would be strictly worse.
- **A comment justified the in-branch build with a query cost that doesn't exist** (`db.select()` is a lazy builder). Replaced with the real reasons — the per-call `new Date()` (hoisting would freeze "now" for the process lifetime, a genuine correctness bug) and the mocked suites' schema stub.

**Tests: 2 → 5 cases**, from the test-analyzer's two rated-8 gaps:
- `installerTokens` is now asserted **from the same response** the filter produced, so the badge half of the invariant is checked instead of trusted. Without it, relaxing `reportsInstallerCapacity` flips a badge to Active while the filter keeps hiding the row — #3191 reintroduced with the suite fully green.
- **Partner scope** (correlated EXISTS over a second RLS-forced table; a blind subquery would silently no-op for every MSP-scoped tech). System scope is documented as unreachable through this harness — no membership row, so it 403s on `requirePermission` before reaching the route.
- **Foreign-org token**, since the subquery correlates only on `parent_enrollment_key_id` and leans entirely on RLS for tenant safety.
- **Multi-token keys** — the badge SUMs and the filter does a per-row EXISTS; a single-token fixture cannot exercise that equivalence at all.

**Deferred (not this PR):** silent-failure-hunter flagged that "Delete expired" can return `{ success: true, deletedCount: 0 }` and silently no-op on rows the list presents as expired, with a toast that reads like success. Pre-existing, on the delete path, and needs a wire field (`skippedLiveInstallerCount`) plus UI — worth its own issue rather than widening this one.

**Tests:** new suite 5 passed, and **4 of the 5 re-confirmed red against the pre-fix route** (the 5th is the RLS case, correctly green either way); `enrollmentKeysPurgeExpired` + `enrollmentKeyCleanup` integration 19 passed total across the three guard-sharing suites; 4 affected API unit files 128 passed; `apps/web` EnrollmentKeyManager 19 passed; `tsc --noEmit -p apps/api` clean; eslint clean.

**CI condition:** the `pull_request` trigger dropped for this branch — `gh run list` showed **zero** Actions runs and `gh pr checks` showed only a Cloudflare Pages entry, which reads deceptively like a green board. Dispatched [run 31127345406](https://github.com/LanternOps/breeze/actions/runs/31127345406) by hand via `workflow_dispatch`; it was queued at hand-off. **Confirm that run is green before merging** — the check list on the PR is not evidence on its own here.

Issue #3217, opened 2026-08-07
SNMP scheduler ignores pollingInterval for never-succeeding devices — polls them every 60s with no backoff


ToddHebebrand:
## Summary

The SNMP poll scheduler (`processScheduler`, `apps/api/src/jobs/snmpWorker.ts`) selects devices where:

```
lastPolled IS NULL OR lastPolled + pollingInterval <= now
```

`last_polled` is only stamped on a **successful** poll. A device that never succeeds (unreachable, bad credentials, or a payload the pipeline rejects) keeps `last_polled = NULL` forever — so it matches the `IS NULL` branch on **every 60s scheduler tick**, regardless of its configured `polling_interval`.

Net effect: the devices most likely to be broken are polled the hardest — a 300s-interval device that always fails gets polled 5× more often than configured. During the recent production incident, a switch whose poll results could never be persisted was re-polled every 60s indefinitely, hammering the failing path once per minute with no backoff and no cap.

## Fix shape (either, ideally both)

1. Stamp a `last_poll_attempted_at` (or reuse `last_polled` as attempted-at, with a separate `last_success_at`) so the due-check runs off attempts, not successes
2. Failure backoff: after N consecutive failures, multiply the effective interval (capped), and/or auto-deactivate with a visible device status after sustained failure — `last_status` already exists for surfacing this

## Notes

- BullMQ retry config is not the issue — poll jobs run with default 1 attempt; the re-poll pressure comes entirely from the scheduler's due-check
- Whatever marks failures must run even when result processing throws (the incident case was a failure *after* dispatch, in result persistence)

PR #3220, opened 2026-08-07
fix(api): scope snmpWorker DB contexts to DB statements only (#3215)

### LINKED ISSUE #3215: snmpWorker runs Redis fan-out and agent WS dispatch inside an open system DB transaction (#1105 class)
## Summary

`createSnmpWorker` (`apps/api/src/jobs/snmpWorker.ts:69`) wraps the **entire** job body in `runWithSystemDbAccess`, so everything the job does runs inside one open, system-scoped transaction holding a pooled connection:

- `poll-scheduler`: the per-device loop calling `queue.getJob` / `getState` / `remove` / `queue.add` — all Redis round-trips — after the initial `snmp_devices` select
- `poll-device`: the `sendCommandToAgent` WebSocket dispatch and its template/credential prep
- `process-poll-results`: metric parsing plus inserts (this one legitimately needs the context, but not for the parse)

This is the #1105 pattern (non-DB work inside `withDbAccessContext`). Observed in production: the scheduler's own `snmp_devices` select sitting `idle in transaction` for **51 seconds** — the longest single connection hold on the box, worse than the 10s `/sessions/live` holds fixed elsewhere.

## Fix shape

Same as the established #1105 remediation: perform the select, close the context, then do the Redis fan-out / WS dispatch outside it (`runOutsideDbContext` or restructure so the context only spans actual DB statements). `process-poll-results` should open its context only around the insert batch.

## Checklist

- [ ] `poll-scheduler`: enqueue loop outside the DB context
- [ ] `poll-device`: agent dispatch outside the DB context
- [ ] `process-poll-results`: context spans inserts only
- [ ] Remove the temporary Caddy edge-shed for `/api/v1/devices/*/sessions/live` on the production droplets when the related pool fixes deploy (the Caddyfile is not part of the api/web image, so a version bump will not remove it — it must be hand-edited, or `/sessions/live` stays disabled and silently masks the fix)

ToddHebebrand:
## Problem

`createSnmpWorker` wrapped the **entire** job body in `withSystemDbAccessContext`, so every SNMP job held a pooled Postgres connection open in a transaction across work that never touches Postgres — the per-device Redis fan-out, the agent WebSocket dispatch, and metric parsing.

Observed in production: the scheduler's own `snmp_devices` select sitting `idle in transaction` for **51 seconds**, the longest single connection hold on the box. This is the #1105 pattern.

## Fix

The worker handler no longer opens an outer context. Each job type manages its own short-lived one, matching the established remediation in `monitorWorker.ts` / `securityPostureWorker.ts` / `intentOutboxPublisher.ts`:

| Job type | Inside the context | Outside it |
|---|---|---|
| `poll-scheduler` | due-device select | `queue.getJob` / `getState` / `remove` / `add` fan-out |
| `poll-device` | device config + template OIDs + online-agent pick (one context) | `isAgentConnected`, credential decrypt, `sendCommandToAgent` WS write |
| `process-poll-results` | org lookup (ctx 1); metric insert + status stamp (ctx 2, still atomic together) | metric parsing/shaping |

`getSnmpQueue` now builds through `createInstrumentedQueue`, so a future enqueue inside a held context trips the #1105 tripwire instead of silently pinning a connection. Both `enqueueSnmpPollResults` call sites in `agentWs.ts` already wrap in `runOutsideDbContext`, so the tripwire stays quiet on the request path.

## Why the test is shaped the way it is

Reintroducing an outer wrap would be **silent**: `withDbAccessContext` re-entry is a no-op (`if (dbContextStorage.getStore()) return fn()`), so an outer context defeats every inner scope without any error. An identity `fn => fn()` mock can never catch that.

`snmpWorker.dbcontext.test.ts` therefore uses a depth-tracking context mock and asserts the exact enter/exit ordering per job type, e.g.:

```
['ctx:enter','dueSelect@depth1','ctx:exit','enqueue@depth0','enqueue@depth0']
```

plus a dedicated case asserting the handler opens exactly one `ctx:enter` and ends at `enqueue@depth0`.

`createSnmpWorker` is now exported so the test can drive the real processor (mirrors `createSecurityPostureWorker`).

## Verification

- `vitest run src/jobs/snmpWorker.dbcontext.test.ts src/jobs/snmpQueue.test.ts src/routes/agentWs.test.ts` — **109 passed**, single worker.
- No schema/migration/tenancy changes; no cascade or RLS list touched.

`snmpQueue.test.ts` needed `assertOutsideHeldDbContext: vi.fn()` added to its `../db` mock — required by `createInstrumentedQueue`, same one-line change `securityPostureWorker.test.ts` carries.

## ⚠️ Expected merge conflict with #3217

**#3217 is being fixed in parallel and edits the same file** (`apps/api/src/jobs/snmpWorker.ts`, the `processScheduler` due-check / backoff logic). This PR deliberately leaves the due-check SQL **textually identical** and only restructures the context around it, so the conflict should be a small, mechanical resolution in `processScheduler`. Whichever merges second should re-run `src/jobs/snmpWorker.dbcontext.test.ts` after resolving — the depth assertions will catch a resolution that accidentally moves the enqueue loop back inside the context.

## Follow-up for the user — production ops, NOT in this PR

The issue's 4th checklist item is a **production droplet action** and is intentionally not done here:

> Remove the temporary Caddy edge-shed for `/api/v1/devices/*/sessions/live` on the production droplets.

The Caddyfile is not part of the api/web image, so a version bump will **not** remove it — it must be hand-edited on each droplet at deploy time, or `/sessions/live` stays disabled and silently masks this fix.

Closes #3215

### DIFF
--- a/apps/api/src/jobs/snmpQueue.test.ts
+++ b/apps/api/src/jobs/snmpQueue.test.ts
@@ -23,6 +23,9 @@ vi.mock('../db', () => ({
     update: vi.fn(),
   },
   withSystemDbAccessContext: undefined,
+  // #1105 tripwire used by createInstrumentedQueue (the factory getSnmpQueue
+  // now builds through). No-op here — these tests hold no DB context.
+  assertOutsideHeldDbContext: vi.fn(),
 }));
 
 vi.mock('../db/schema', () => ({
--- a/apps/api/src/jobs/snmpWorker.dbcontext.test.ts
+++ b/apps/api/src/jobs/snmpWorker.dbcontext.test.ts
@@ -316,9 +316,15 @@ describe('snmpWorker DB-context scoping (#3215)', () => {
   });
 
   it('the worker handler does not wrap job bodies in an outer context', async () => {
-    // Guard against reintroducing the outer runWithSystemDbAccess: because
-    // withDbAccessContext re-entry is a no-op, an outer wrap would make the
-    // scheduler's enqueues run at depth 1 and never emit a second ctx:enter.
+    // Guard against reintroducing the outer runWithSystemDbAccess.
+    //
+    // Careful about WHY this fails, because prod and this mock differ: in prod
+    // `withDbAccessContext` short-circuits on re-entry, so an outer wrap would
+    // leave one transaction open across the enqueues. This mock does NOT model
+    // that short-circuit — it increments depth unconditionally — so an outer
+    // wrap here shows up as TWO ctx:enter events with the select at depth 2 and
+    // the enqueues at depth 1. Either way the assertions below fail (verified by
+    // mutation), which is what matters; don't read the mock as prod semantics.
     mockDb.select.mockReturnValueOnce(
       selectWhereChain([{ id: 'd1', orgId: 'o1', pollingInterval: 60, lastPolled: null }], 'dueSelect') as never
     );
--- a/apps/api/src/jobs/snmpWorker.dbcontext.test.ts
+++ b/apps/api/src/jobs/snmpWorker.dbcontext.test.ts
@@ -1,0 +1,336 @@
+/**
+ * snmpWorker DB-context scoping (#1105 class, issue #3215).
+ *
+ * The regression these tests lock down: `createSnmpWorker` used to wrap the
+ * ENTIRE job body in `withSystemDbAccessContext`, so the scheduler's Redis
+ * fan-out and the poll-device agent WebSocket dispatch ran while a pooled
+ * Postgres connection sat idle-in-transaction (observed: 51 seconds).
+ *
+ * An identity `fn => fn()` mock of the context helper can never catch that, so
+ * the mock below tracks real enter/exit depth and the tests assert WHICH depth
+ * each DB read, each enqueue and each WS send happened at.
+ */
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockDb, ctxState, queueMock, agentWsMock } = vi.hoisted(() => ({
+  mockDb: {
+    select: vi.fn(),
+    insert: vi.fn(),
+    update: vi.fn(),
+  },
+  // DB-access-context depth + an ordered event log, so a test can prove which
+  // work runs inside a held transaction and which runs after it closes.
+  ctxState: { depth: 0, events: [] as string[] },
+  queueMock: {
+    getJob: vi.fn(async () => null),
+    add: vi.fn(async () => ({ id: 'job-1' })),
+    addBulk: vi.fn(async () => []),
+    getRepeatableJobs: vi.fn(async () => []),
+    removeRepeatableByKey: vi.fn(async () => undefined),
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter.

**Findings:** 3 raised → all 3 addressed in `15909e6e0`; 0 outstanding.

- code-reviewer: **0 critical, 0 important.** It mutation-tested the new suite (reintroduced the outer `runWithSystemDbAccess` → 7 of 8 tests fail), and confirmed the tripwire assertion is not a vacuous negative. One sub-threshold note: the regression-guard test's rationale comment described prod re-entry short-circuit semantics that the depth-tracking mock deliberately does not model — corrected.
- silent-failure-hunter: **no swallowed throw, no new silent no-op, no partial-write hazard.** Insert + status stamp still share one context. Two findings, both about *future* silence:
  - The `createInstrumentedQueue` comment over-claimed. The tripwire wraps only `Queue.add`/`addBulk`, but `enqueueSnmpPoll` does up to three Redis round-trips before that (`queue.getJob`, then `job.getState`/`job.remove` on the **Job** object, unreachable from a Queue-level factory) and returns without ever calling `add` on the reusable-state dedupe path — the steady state of a busy poll queue. So the tripwire alone would **not** have caught the 51-second hold. Comment now says so and names the depth assertions as the real fence.
  - `PollDispatchInputs` was a flat struct that distinguished "device missing" / "no OIDs" / "no online agent" only by the **order** the caller checked them — reorder the guards and `No OIDs configured` silently becomes `No online agent`, pointing ops at agent connectivity for a template misconfiguration. Now a discriminated union, so the log line derives from the cause.

**Tests:** `vitest run src/jobs src/routes/agentWs.test.ts` — **989 passed**, 1 skipped, single worker. `tsc --noEmit` on `apps/api` clean. Full PR CI green (Test API, Type Check, Integration Tests 1-4/4, Smoke Test all passing; 0 failures).

**Two follow-ups surfaced, both filed as notes rather than changed here:**
1. Pre-existing, out of scope — `processPollDevice` uses two org sources for one dispatch: the template lookup scopes on `device.orgId` (the row) while agent selection scopes on `data.orgId` (the job payload, captured at enqueue time). Identical on `main`, so not a regression, but the dispatch ships decrypted SNMP community strings and v3 auth/priv passwords to whichever agent that second query picks, and the stable `snmp-poll-<deviceId>` jobId means a queued payload can outlive the row state it was built from. Worth its own issue.
2. The `"No OIDs configured"` warning still collapses four causes (no `templateId`; template row missing or invisible under the `isBuiltIn OR orgId` filter; `oids` not an array; genuinely empty) — the middle two are data-integrity errors reported as user misconfiguration. Byte-identical to `main`, left alone to keep this diff scoped.

**Status:** review-clean, CI green, awaiting maintainer merge. Expect a conflict with #3217 in `processScheduler` — see the PR body.

Issue #3214, opened 2026-08-07
api: postgres.js pool never recovers dead connections — CONNECT_TIMEOUT storm against a healthy DB until restart


ToddHebebrand:
## Summary

When postgres.js (v3.4.9) connections in the API's pool die, the pool never recovers them. Every reconnect attempt fails with `write CONNECT_TIMEOUT` even though the database is healthy and has free connection slots. The pool decays over hours (observed: 35 configured → 9 live) until nearly every query — request handlers and workers alike — fails at connect. Only an API process restart restores the pool.

Observed in production on 0.103.0: ~144 `CONNECT_TIMEOUT` errors/min sustained, while the DB showed less than half its connection slots used, zero blocking locks, and an external `psql` from the same host connected in under a second. After `docker compose restart api`, the pool refilled to its full configured size within seconds and `CONNECT_TIMEOUT` dropped to zero — confirming the failure lived in the process's connection state, not the network or the DB.

## Likely poisoners

Two error classes fire from inside postgres.js's connection machinery, both starting days before the visible outage:

- `Error` at `TLSSocket.closed (postgres.cjs.src:connection)` — the upstream TLS socket closing
- `TypeError` at `Immediate.nextWrite (postgres.cjs.src:connection:255)` — a crash inside the driver's own write scheduler

A TypeError thrown inside the driver's connection lifecycle plausibly leaves the connection object in a state where it is neither usable nor properly torn down/replaced, so the pool slot is lost for good.

## Impact

- Dashboard-wide slowness/503s once enough of the pool is gone (compounds with #1105-class holds, which stretch from ~5s to 40s+ under the shrunken pool)
- All background workers (metric anomalies, SNMP, alert evaluation) fail their DB writes
- The failure is invisible to `/health` (no DB dependency) and self-sustains until a human restarts the API

## Suggested fixes (any subset)

1. Reproduce/diagnose the `Immediate.nextWrite` TypeError; check whether postgres.js > 3.4.9 fixes it, or file upstream
2. Pool-health watchdog: periodically compare live connections vs `DB_POOL_MAX`; on sustained decay + connect failures against a reachable DB, log loudly (and optionally recycle the `sql` instance)
3. Surface `CONNECT_TIMEOUT` rate as a first-class metric/alert — it is the earliest reliable signal of this failure mode

## Evidence

[...]

PR #3221, opened 2026-08-07
fix(db): name the opening caller in the #1105 hold warning (#3218)

### LINKED ISSUE #3218: DB-context hold warning is unattributable from console logs — openedAt caller only reaches Sentry
## Summary

The `withDbAccessContext` hold warning (`apps/api/src/db/index.ts`, #1105 tripwire) captures the caller that opened the context (`openedAt` stack) — but sends it **only to Sentry**. The console warning contains scope, optional label, and duration, and nothing that names the caller unless the call site passed a `label` (most don't).

During the recent production incident this became a real diagnostic dead end, twice over:

1. **Console-only debugging is blind.** The droplet logs showed 99 holds/10min at 10s and system-scope holds at 40s, with no way to attribute them from the box. Attribution required Sentry.
2. **Sentry wasn't there either.** The concurrent hot error (`connectTimedOut`, ~32k events/day) saturated the DSN's 1,000/hr rate limit, so the throttled hold-warning captures (1/scope/5min) were dropped exactly when they mattered. The incident that causes the holds also suppresses their attribution.

## Fix shape

- Include the top app frame of the `openedAt` stack (file:line, or the resolved label) in the **console** warning message — the stack is already captured; formatting one frame is cheap and only happens on breach
- Optionally: auto-derive a label from the opening frame when no explicit `label` was passed, so both console and Sentry group usefully

## Non-goals

Client-side capping of genuine hot errors (the DSN-saturation half) is a separate, known exposure — this issue is only about making the console warning self-sufficient for attribution.

ToddHebebrand:
## Problem

The `withDbAccessContext` hold tripwire (`apps/api/src/db/index.ts`, #1105) already captured the opener's stack — but only ever sent it to **Sentry**. The console warning carried scope, duration, and an optional `label` that most call sites don't pass. During the recent production incident that was a diagnostic dead end twice over:

1. **Console-only debugging was blind** — droplet logs showed 99 holds/10min with no way to attribute them from the box.
2. **Sentry wasn't there either** — the concurrent hot error saturated the DSN's rate limit, so the throttled hold captures (1/scope/5min) were dropped exactly when they mattered. The incident that *causes* the holds also suppresses their attribution.

## Change

Format one frame from the already-captured stack into the log line.

- **`parseOpenerFrame(stack)`** — picks the first **application** frame, skipping `node_modules`, node internals, and this module's own wrapper frames. That last skip matters: the `Error` is allocated inside `withDbAccessContext`, so the emitter frame is *always* on top — returning it would reproduce the BREEZE-9 failure of ~12k events naming the emitter instead of the caller. Handles `async`, `Object.fn [as alias]`, `new ClassName`, bare-location frames, `file://` ESM paths, `packages/` frames, and built `.js` paths.
- **`formatHeldContextWarning(...)`** — composes the warning. Returns `message` (Sentry-facing) and `consoleLine` separately. Both functions are pure and exported, so the text that actually reaches the console is testable without a live pool.
- **Auto-derived label** — when no explicit `label` was passed, one is derived from the opening frame (`basename.fn`) so unlabelled callers group per source instead of arriving as one opaque bucket. An explicit label still wins; it's curated and can separate paths that share one helper closure (the agent-WS case, where 12 contexts funnel through one closure).

### The load-bearing fix: *where* the stack is captured

The opener `Error` was allocated **inside the transaction callback, after six `await tx.execute(...)` hops**. That only ever worked via V8's async-stack reconstruction — which links a frame **only at a genuine `await`**. A caller doing a bare `return withSystemDbAccessContext(...)` was dropped from the trace entirely.

That idiom is used at **66 call sites** in this repo, including most BullMQ job workers — a prime suspect for the real long holds this tripwire exists to catch. Those would have been unattributed, or *misattributed to the next function out*, which is worse than nothing because it names an innocent frame as the culprit.

Fixed at the root: the `Error` is now allocated at **`withDbAccessContext` entry**, where the caller's frame is live on the synchronous stack regardless of how it invoked us — no async reconstruction involved. Cost is unchanged: same single allocation per context open, still gated on `warnMs > 0`, still only `.stack`-formatted on an actual breach. `startedAt` deliberately stays inside the transaction callback, since the hold being measured begins when the transaction owns a pooled connection.

### Two deliberate calls worth reviewing

- **`file:line` goes to the console only, never into the Sentry `message`.** Sentry groups by message text, and a line number shifts on every unrelated edit above the call site — putting it in `message` would fork one issue into a new one per release. Sentry gets the precise frame as the `openedAtFrame` extra instead.
- **The derived name gets its own `dbContextOpener` tag** rather than widening `dbContextLabel`, which stays explicit-only so existing Sentry queries/dashboards keep their meaning.

**Expected side effect:** previously-unlabelled callers now carry a derived label in the message, so they regroup **once** in Sentry into per-source issues. That's the intent of the issue, not a regression. The two existing integration tests that encoded the old "unlabelled callers keep byte-identical text" contract were updated deliberately.

### Before / after

```
- withDbAccessContext (scope=system) held a pooled connection in an open transaction for 40000ms (>= 2000ms) — …
+ withDbAccessContext (scope=system) [sweeper.sweep] held a pooled connection in an open transaction for 40000ms (>= 2000ms) — … Opened at apps/api/src/workers/sweeper.ts:77:9.
```

## Non-goal

Client-side capping of the genuine hot errors that saturate the DSN is a separate, known exposure and is explicitly out of scope per the issue. Untouched here.

## ⚠️ Overlapping work in flight

**#3214 is being fixed in parallel and also edits `apps/api/src/db/index.ts`** (pool configuration / connection-health). This PR's diff is confined to the hold-warning path — the two new pure helpers, and moving the opener allocation to function entry. A conflict is possible but should be mechanical — expect it and merge accordingly.

## Verification

- **Unit:** `vitest run src/db/` single-worker — **32 files, 1412 passed / 13 skipped**. 36 of those are in the touched file.
- **Integration (blocking `integration-test` job):** ran locally against a real Postgres via a worktree-private stack — `dbContextTripwire.integration.test.ts` **15 passed**, including a new end-to-end test proving a bare-`return` caller now attributes correctly.
- No schema, migration, tenancy, or cascade surface touched.

Closes #3218

### DIFF
--- a/apps/api/src/__tests__/integration/dbContextTripwire.integration.test.ts
+++ b/apps/api/src/__tests__/integration/dbContextTripwire.integration.test.ts
@@ -179,28 +179,68 @@ describe('#1105 DB-context tripwires', () => {
       // The TAG is the load-bearing part: extras are only visible inside a
       // single event, so an unfilterable bucket (BREEZE-A, ~7k events) stays
       // unfilterable if the label only lands in `extra`.
-      expect(capturedMessages[0]!.tags).toEqual({ dbContextLabel: 'agentWs.heartbeat' });
+      expect(capturedMessages[0]!.tags?.dbContextLabel).toBe('agentWs.heartbeat');
       // Also in the message, because Sentry groups by message — that is what
       // splits one bucket into per-handler issues.
       expect(capturedMessages[0]!.message).toContain('[agentWs.heartbeat]');
+      // #3218 added a second, derived tag alongside it. `dbContextLabel` must
+      // stay EXPLICIT-only so existing Sentry queries keep their meaning — the
+      // derived opener goes in its own key, never widening this one.
+      expect(capturedMessages[0]!.tags?.dbContextOpener).toEqual(expect.stringContaining('dbContextTripwire'));
     });
 
-    it('omits the tag entirely for an unlabelled context, leaving the message text unchanged', async () => {
-      // Grouping stability: every existing caller is unlabelled, so their
-      // message must be byte-identical to before this change or their Sentry
-      // history splits into a new issue for no reason.
+    it('derives a grouping label from the opener when the context is unlabelled (#3218)', async () => {
+      // Supersedes the previous contract ("unlabelled callers keep byte-identical
+      // message text"). That stability was deliberately traded away in #3218:
+      // most callers are unlabelled, so preserving their text meant they all
+      // landed in ONE opaque bucket that could not be attributed. Regrouping
+      // them per source is a one-time cost paid once, on purpose.
       process.env.DB_CONTEXT_HELD_WARN_MS = '50';
       process.env.DB_CONTEXT_HELD_CAPTURE_THROTTLE_MS = '0';
       __resetHeldContextCaptureThrottleForTests();
 
-      await withSystemDbAccessContext(async () => {
-        await new Promise((resolve) => setTimeout(resolve, 90));
-      });
+      async function anUnlabelledCallerThatHolds(): Promise<void> {
+        await withSystemDbAccessContext(async () => {
+          await new Promise((resolve) => setTimeout(resolve, 90));
+        });
+      }
+      await anUnlabelledCallerThatHolds();
 
       expect(capturedMessages).toHaveLength(1);
-      expect(capturedMessages[0]!.tags).toBeUndefined();
-      expect(capturedMessages[0]!.message).toContain('withDbAccessContext (scope=system) held');
-      expect(capturedMessages[0]!.message).not.toContain('[');
+      const event = capturedMessages[0]!;
+      expect(event.message).toContain('withDbAccessContext (scope=system)');
+      expect(event.message).toContain(HELD);
+      // No explicit label was passed, so only the derived tag is present.
+      expect(event.tags?.dbContextLabel).toBeUndefined();
+      expect(event.tags?.dbContextOpener).toContain('anUnlabelledCallerThatHolds');
+      // The derived name reaches the grouped message too...
+      expect(event.message).toContain('[');
+      expect(event.message).toContain('anUnlabelledCallerThatHolds');
+      // ...but the precise file:line must NOT, or every edit above the call site
+      // forks the Sentry issue. It belongs to the console line and the extra.
+      expect(event.message).not.toMatch(/\.ts:\d+/);
+      expect(String(event.extra?.openedAtFrame)).toMatch(/dbContextTripwire.*:\d+:\d+/);
+    });
+
+    it('names the opening caller in the CONSOLE line, not just in Sentry (#3218)', async () => {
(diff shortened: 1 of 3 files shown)


ToddHebebrand:

**Findings:** 4 raised → all 4 addressed; 0 outstanding.

- **Attribution silently failed for bare-`return` callers** (critical). The opener `Error` was allocated inside the transaction callback, *after* six `await`s, so it relied on V8 async-stack reconstruction — which links a frame only at a genuine `await`. The 66 `return withSystemDbAccessContext(...)` call sites here (most BullMQ workers) were dropped from the trace, or misattributed to the next function out. Fixed at the root in `8e44603a5` by allocating at function entry, where the caller's frame is live synchronously. Proven end-to-end by a new real-Postgres test.
- **A dependency frame could be promoted to "the caller"** — `toRepoRelative` cut from the first `apps|packages|agent|scripts` segment *before* the `node_modules` rejection ran, so `node_modules/undici/scripts/build.js` shortened to `scripts/build.js` and passed as app code. Rejection now runs against the full path (`8e44603a5`).
- **Empty/whitespace `label` defeated its own safeguard** — it won the `??` race against the derived fallback, then rendered as nothing, because the value used `??` while both render sites used truthy checks. Normalized once in `3c04b5568`.
- **Instrumentation could go silent** — the `finally`'s blanket `catch {}` now also covers the new parsing, so a future slip there would make the warning vanish with no trace. Added a guarded breadcrumb (`3c04b5568`), keeping the catch broad so `fn`'s real outcome is never altered.

**Tests:**
- Unit — `vitest run src/db/` single-worker: **32 files, 1412 passed / 13 skipped**.
- Integration (the **blocking** `integration-test` job) — ran locally against real Postgres on a worktree-private stack: `dbContextTripwire.integration.test.ts` **15 passed**, including the new bare-`return` attribution case.
- `tsc --noEmit` on `apps/api`: clean.

**Heads-up for the merger:** two pre-existing integration assertions encoded the *old* contract (unlabelled callers keep byte-identical message text; `tags` exactly `{dbContextLabel}`). #3218 deliberately supersedes that, so they were rewritten rather than preserved — previously-unlabelled callers will regroup **once** in Sentry into per-source issues. That is the intent, not a regression.

**Overlap:** #3214 is in flight on the same file (pool config / connection health). Conflict is possible but should be mechanical.

**Status:** review-clean, CI in flight, awaiting maintainer merge.

ToddHebebrand:
**CI:** green — 54/55 pass, 0 fail (1 skipped). `CI Success`, `Test API`, all four `Integration Tests` shards, `Lint`, and both `Check Migrations` jobs pass on `8e44603a5`.

**Status:** review-clean and CI-green. Ready for maintainer merge — not merging, and leaving #3218 open per the handoff rules.

PR #3222, opened 2026-08-07
fix(api): scope alertWorker evaluate-all DB context to reads only (#3216)

### LINKED ISSUE #3216: alertWorker evaluate-all holds a system DB context ~5s every 60s cycle (#1105 class)
## Summary

The alert evaluation worker's `evaluate-all` job holds a system-scoped DB context (and its pooled connection) for ~5 seconds on every 60-second cycle. This is the single largest steady #1105 offender remaining: **15,469** Sentry hold-warning events since June 16 (BREEZE-9), all tagged `worker: alertWorker, jobName: evaluate-all`, tightly clustered at ~4.9s.

Under normal pool conditions this wastes one connection ~8% of the time. Under a degraded pool (see the postgres.js pool-decay issue) the same job's holds stretched to 20–40s, amplifying the outage.

## Where

`processEvaluateAll`'s fan-out — the job body does its per-device/per-alert-rule iteration (and enqueueing) inside the transaction opened by the worker's `runWithSystemDbAccess`-style wrapper, instead of selecting, closing the context, and fanning out afterwards.

## Fix shape

Standard #1105 remediation: scope the DB context to the reads/writes only; move the iteration + queue fan-out outside (`runOutsideDbContext` or restructure). One integration/unit test asserting the evaluation path doesn't hold a context across the enqueue loop would prevent regression.

## Evidence

Sentry BREEZE-9: `withDbAccessContext (scope=system) held a pooled connection ... for 4943ms`, `openedAt` stack terminating in the BullMQ worker trampoline, tags `worker=alertWorker`, `jobName=evaluate-all`.

ToddHebebrand:
## Problem

The `evaluate-all` sweep ran its entire per-org/per-device iteration **and** its Redis fan-out inside the system DB context opened by the worker wrapper. That pinned a pooled Postgres connection idle-in-transaction for ~4.9s on every 60s cycle — **15,469** held-context warnings since June 16 (Sentry **BREEZE-9**), the largest remaining steady #1105 offender. Under normal pool conditions it wasted a connection ~8% of the time; under a degraded pool the same holds stretched to 20–40s and amplified the outage.

## Fix

Standard #1105 remediation, matching `monitorWorker`'s `monitor-scheduler` and `securityPostureWorker`'s `scan-orgs` — structurally the same fan-out job:

- **Worker handler no longer blanket-wraps the job body.** `evaluate-all` self-manages its context; `evaluate-device` and `auto-resolve` keep theirs (they read *and* write per device).
- **`processEvaluateAll` opens one short-lived system context per DB read** — the org list, then each keyset-paginated device page — and lets it close before that page's `queue.addBulk`. No connection is held across an enqueue.
- **`getAlertQueue()` now builds via `createInstrumentedQueue`**, so any future enqueue from inside a held context trips the #1105 tripwire instead of going unnoticed. The bare `new Queue` is precisely why this hid for months (same note already on `securityPostureWorker`).

### Behaviour note — pagination now spans transactions

Intentional and safe: the keyset cursor (`devices.id` ascending) stays stable, `recentThreshold` is computed once so the eligibility window does not drift between pages, and a device whose status flips mid-sweep is simply picked up on the next 60s cycle. This is a best-effort fan-out enqueue, not a consistent snapshot.

No tenancy change: reads still run inside a system context, so RLS behaviour is identical. The context is **scoped**, not removed — one of the new tests asserts exactly that, so a "fix" that just deleted the context would fail.

## Verification

- `apps/api/src/jobs/**` + `ticketEvents` + `readiness` suites, single-fork: **926 passed, 1 skipped, 0 failed**.
- `pnpm exec tsc --noEmit --project apps/api/tsconfig.json`: clean.
- **Guard proven non-vacuous:** temporarily reinstating the blanket `runWithSystemDbAccess` wrapper turns the new test red — `expected [1, 1] to deeply equal [0, 0]`.

### Tests added

Three cases in `alertWorker.test.ts` asserting where the context boundary actually falls:

1. `does not hold a DB context across the enqueue loop` — every `addBulk` across a 3-page sweep observes zero open contexts, the tripwire records no violations, the reads all observe an open context, and no context leaks after the sweep.
2. `runs the evaluate-all job through the worker handler with no wrapping context` — driven through the **real worker processor** (the wrapper is what opened the offending context, so asserting on `processEvaluateAll` alone would not catch a regression there).
3. `still wraps per-device evaluation in a system DB context` — confirms the other job types were not collaterally unwrapped.

Closes #3216

### DIFF
--- a/apps/api/src/jobs/alertQueue.test.ts
+++ b/apps/api/src/jobs/alertQueue.test.ts
@@ -27,7 +27,9 @@ vi.mock('../services/redis', () => ({
 
 vi.mock('../db', () => ({
   db: {},
-
+  // getAlertQueue now builds an instrumented Queue, whose add/addBulk call the
+  // #1105 tripwire guard (services/bullmqQueue.ts).
+  assertOutsideHeldDbContext: vi.fn(),
   runOutsideDbContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
   withDbAccessContext: vi.fn(async (_ctx: unknown, fn: () => Promise<unknown>) => fn()),
   withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
--- a/apps/api/src/jobs/alertWorker.test.ts
+++ b/apps/api/src/jobs/alertWorker.test.ts
@@ -20,6 +20,11 @@ const {
     orgReadDepths: [] as number[],
     deviceReadDepths: [] as number[],
     addBulkDepths: [] as number[],
+    // Every operation the #1105 guard was consulted for. Asserting this is
+    // non-empty is what pins the queue to createInstrumentedQueue — checking
+    // only `tripwireViolations` would pass just as happily against a bare
+    // `new Queue`, which never calls the guard at all.
+    tripwireCalls: [] as string[],
     tripwireViolations: [] as string[]
   };
   return {
@@ -110,6 +115,7 @@ vi.mock('../db', () => ({
   // Mirrors the production guard wired into createInstrumentedQueue: record the
   // enqueue call sites that ran while a transaction was still held.
   assertOutsideHeldDbContext: (operation: string) => {
+    ctx.tripwireCalls.push(operation);
     if (ctx.depth > 0) ctx.tripwireViolations.push(operation);
   }
 }));
@@ -151,6 +157,7 @@ const resetCtx = () => {
   ctx.orgReadDepths.length = 0;
   ctx.deviceReadDepths.length = 0;
   ctx.addBulkDepths.length = 0;
+  ctx.tripwireCalls.length = 0;
   ctx.tripwireViolations.length = 0;
 };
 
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` — focused on DB-context scoping correctness, the cross-transaction pagination change, `getAlertQueue()` caller impact, and whether the new tests are tautological.

**Findings:** 2 raised → both addressed in `cbf0f6a87`; 0 outstanding.

1. **(Important) The tripwire assertion was vacuous.** `expect(ctx.tripwireViolations).toEqual([])` passes identically whether the queue is instrumented or a bare `new Queue` — a bare queue never consults the guard, so `[]` was the expected value in the failure mode the assertion appeared to guard. Now every guard invocation is recorded and the enqueues are asserted to have actually reached it. Verified red against a reverted bare `new Queue`: `expected [] to deeply equal [ …(3) ]`.
2. **(Comment accuracy)** The `getAlertQueue` comment claimed the bare `new Queue` is why this "went unnoticed for months". `createInstrumentedQueue` is a recent, still incrementally-adopted factory, so it could not have caught this for most of that window — reworded to "left this queue outside the tripwire's coverage".

**Explicitly verified clean by the review:** no DB read escaped to the bare pool (RLS behaviour unchanged); the cross-transaction pagination introduces no new hazard, since `withDbAccessContext` already ran at READ COMMITTED so every statement took its own snapshot before this change, and the `devices.id` keyset cursor is a stable total order (unlike OFFSET, no row can shift across a page boundary); `triggerDeviceEvaluation` / `triggerFullEvaluation` / `getAlertQueueStatus` have **zero** non-test callers repo-wide, so the `createInstrumentedQueue` switch cannot newly trip the guard for any live path — `DB_CONTEXT_TRIPWIRE_STRICT` has nothing to turn into a throw.

**Tests:** `apps/api/src/jobs/**` + `ticketEvents` + `readiness`, single-fork — **926 passed, 1 skipped, 0 failed**. `pnpm exec tsc --noEmit --project apps/api/tsconfig.json` clean.

**Regression guards proven non-vacuous** (each verified red against the pre-fix shape):
- reinstating the blanket `runWithSystemDbAccess` wrapper → `expected [1, 1] to deeply equal [0, 0]`
- reverting to a bare `new Queue` → `expected [] to deeply equal [ …(3) ]`

PR #3223, opened 2026-08-07
fix(snmp): run the poll due-check off attempts and back off failures (#3217)

### LINKED ISSUE #3217: SNMP scheduler ignores pollingInterval for never-succeeding devices — polls them every 60s with no backoff

ToddHebebrand:
## The bug

`processScheduler` (`apps/api/src/jobs/snmpWorker.ts`) selected devices on:

```
last_polled IS NULL OR last_polled + polling_interval <= now
```

`last_polled` is stamped only when poll results are **successfully persisted**. A device that never succeeds — unreachable, bad credentials, or results the pipeline cannot write — keeps `last_polled = NULL` forever, so it matched the `IS NULL` branch on **every 60s scheduler tick**, ignoring `polling_interval` entirely. A 300s device that always fails was polled 5× more often than configured. The devices most likely to be broken got polled the hardest, with no backoff and no cap.

## The fix

Two new columns on `snmp_devices`:

| Column | Purpose |
|---|---|
| `last_poll_attempted_at` | Stamped for every poll job, whatever the outcome. The due-check runs off `GREATEST(last_poll_attempted_at, last_polled)` — NULL-tolerant in Postgres, so it yields the more recent timestamp and stays NULL only for devices never touched at all. |
| `consecutive_failures` | Scales the effective interval by `2^min(failures, 6)`, absolute-capped at 1h, then floored at `polling_interval` so backoff can only ever slow a device **down**. |

For a 300s device: 300s → 600 → 1200 → 2400 → 3600 (capped). A device already configured at or above 1h gets no backoff — the floor wins, which is intended.

### Attempt-stamping and failure-counting are separate

- **`markPollAttempted`** stamps the timestamp for *every* poll job, including undispatchable ones. This is what stops the every-60s hammering.
- **`markPollDispatched`** increments the failure counter only once the poll is genuinely going out to a connected agent.

The split matters: "no OIDs configured" and "org has no online agent" are Breeze-side conditions that say nothing about the SNMP target. Counting them would mark every healthy switch in an org `offline` and back it off for an hour whenever an MSP's only agent host reboots.

### Why the counter is incremented before dispatch, not in a failure handler

The issue explicitly calls this out: *"whatever marks failures must run even when result processing throws (the incident case was a failure after dispatch, in result persistence)."*

A poll can die where no catch block in this worker reaches — the agent may accept the command and never reply, or results may come back and persistence may throw on the way to the DB. So the counter is incremented up front and **only `processPollResults` clears it**, after the metric insert lands. Every failure mode backs the device off; only a genuine end-to-end success cancels it.

It is incremented *before* `sendCommandToAgent` rather than after, because a fast agent on a local switch can round-trip results and reset the counter while this worker is still running — a post-dispatch increment would clobber that reset and strand a healthy device at one failure forever.

### Every other site that records a poll outcome

- **`agentWs.ts` Redis-unavailable fallbacks** (both) now clear the counter. There the agent polled and replied successfully and only our own pipeline dropped the metrics. Previously a Redis outage marched every SNMP device in the fleet to `offline` + the 1h cap, while flapping status between `offline` (written at dispatch) and `warning` (written on the result).
- **Both SNMP config write routes in `monitoring.ts`** (PUT upsert, PATCH) now reset the counter and clear the attempt timestamp, so an admin fixing a wrong community string sees the corrected config exercised on the next tick instead of waiting up to an hour. Without it there is no escape hatch — the device test endpoint is a deprecated 410. The PATCH audit trail still records only the caller's own changed fields.

### Visibility

After 3 consecutive failures the device surfaces as `last_status = 'offline'` — an existing status value `MonitoringAssetsDashboard` already styles and already counts as needing attention, so no UI or i18n change is needed. Auto-deactivation (the other half of the issue's "and/or") is left out: that's a product decision, not a bug fix.

## Cascade / export-policy registration

`snmp_devices` is already in `CORE_ORG_CASCADE_DELETE_ORDER`, and it has no `device_id`, so the device-side lists are unaffected. But per CLAUDE.md the **export-policy row fires on a new column, not just a new table** — both columns are registered in `CORE_TENANT_EXPORT_POLICY` (`included`: a timestamp and a monotonic counter, no secret material). Without that, `tenant-export-policy.integration.test.ts` reds main.

## Migration

`apps/api/migrations/2026-08-15-snmp-poll-attempt-backoff.sql` — idempotent (`ADD COLUMN IF NOT EXISTS`), no inner `BEGIN`/`COMMIT`, hand-written (no drizzle-kit), on a fresh date so it sorts last and stays clear of the closed `2026-08-06` block. It seeds `last_poll_attempted_at` from `last_polled` so already-polled devices don't all become instantly due on the first tick after deploy, and reports the seeded row count via `RAISE WARNING`.

## ⚠️ Overlapping work — expect a conflict

**Issue #3215 is being fixed in parallel and edits the same file** (`apps/api/src/jobs/snmpWorker.ts`) — the DB-context scoping / Redis fan-out restructuring. This PR deliberately did **not** touch the `runWithSystemDbAccess` wrapping, so the diffs should be separable, but whichever merges second will need a rebase. Conflict points: the `processScheduler` due-check body, the top of `processPollDevice`, and the `processPollResults` update. The `agentWs.ts` edits are two small `.set()` blocks in the Redis-unavailable branches and may also collide if #3215 restructures those.

## Verification

- `apps/api` affected suites, single-fork: **1126 passed / 1 skipped** — includes the new `snmpWorkerScheduler.test.ts` (14 tests), all of `src/jobs/`, `snmp.test.ts`, `agentWs.test.ts`, the three `monitoring_*` suites, `autoMigrate.test.ts`, and the `cascadeDelete` / `moveOrg.coverage` contract tests.
- `tsc --noEmit -p apps/api/tsconfig.json` (with CI's `--max-old-space-size=8192`) — clean.
- `scripts/check-migration-naming.sh` — OK.
- The generated SQL for both the due-check and the `CASE` status expression was executed against a real PostgreSQL cluster during review, including a truth table across `polling_interval ∈ {60,300,3600,7200}` × `failures 0..8`. Types resolve to `double precision` at every nesting level (exact match for `make_interval(secs => double precision)`), `GREATEST` NULL-tolerance behaves as assumed, and the floor holds so no device is ever polled faster than configured.

**Not run locally:** `pnpm db:check-drift` and the integration/RLS contract suites — no Docker/Postgres available in this environment. Both need CI; `tenant-export-policy.integration.test.ts` and the drift check are the two that matter for this diff.

## Known residual risks (pre-existing, documented not fixed)

- **Stalled-job redelivery** (`maxStalledCount: 2`) can re-execute a poll job and double-count. The pre-dispatch increment placement keeps the window to the few statements between the increment and the send.
- **`orphanedResultExpectations`** is an in-process `Map`; if an API restart or a multi-instance deploy loses the expectation, the result is rejected and the counter never clears. That previously cost one poll's metrics and now also withholds the reset. Worth tracking separately if the API is ever run multi-instance.
- `processPollDevice` still distinguishes its failure reasons only in `console.warn`. A `last_failure_reason` column would make "no agent" vs "target unreachable" visible to operators — out of scope here.

Closes #3217


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter`, both against `origin/main...HEAD`.

**Findings:** 3 confirmed defects (both reviewers independently found the same three), all addressed in `69a7ff5ab`; 0 outstanding.

1. **Undispatched polls counted as device failures** — `markPollAttempt` incremented before any feasibility check, so "no OIDs configured" and "org has no online agent" marked healthy switches `offline` after 3 ticks and backed them off to the 1h cap. An MSP whose only agent host reboots overnight would find every SNMP target red. Fixed by splitting the stamp (`markPollAttempted`, unconditional) from the increment (`markPollDispatched`, only once the poll is going out to a connected agent). Increment placed *before* `sendCommandToAgent` so a fast local reply can't have its reset clobbered.
2. **Redis-unavailable fallbacks never cleared the counter** (`agentWs.ts` ×2) — the agent polled and replied fine there; only our pipeline dropped the metrics. A Redis outage marched the whole fleet to `offline` + 1h backoff while flapping status between `offline` and `warning` each cycle. Both `.set()` calls now clear it.
3. **No reset on config change** (`monitoring.ts` PUT + PATCH) — an admin fixing a wrong community string on a backed-off device saw nothing happen for up to an hour, with no manual re-poll available (the device test endpoint is a deprecated 410). Both routes now reset the counter and clear the attempt timestamp.

Also corrected three comments the review caught overstating or predating the fix — the `markPollAttempted` throw rationale, the `MAX_BACKOFF_SECONDS` cap (the floor is applied outside it), and the `scheduleSnmpPolling` doc comment, which still described the old `lastPolled` due-check.

**Verified clean by review, not just asserted:** the reviewer stood up a real PostgreSQL cluster, rendered the drizzle `SQL` objects through `PgDialect`, and executed the emitted statements with real bind params. Types resolve to `double precision` at every nesting level (exact match for `make_interval(secs => double precision)`), the `::int` casts are harmless, `GREATEST` NULL-tolerance behaves as assumed, reusing the `nextFailures` fragment twice causes no param-numbering bug, and the floor holds across `polling_interval ∈ {60,300,3600,7200}` × `failures 0..8` so nothing is ever polled faster than configured. Migration/schema match exactly.

**Tests:** `apps/api` affected suites single-fork — **1126 passed / 1 skipped**, covering all of `src/jobs/`, `snmp.test.ts`, `agentWs.test.ts`, the three `monitoring_*` suites, `autoMigrate.test.ts`, and the `cascadeDelete`/`moveOrg.coverage` contract tests. `tsc --noEmit` clean (with CI's 8GB heap), `eslint` clean, `check-migration-naming.sh` OK.

**Not verifiable locally — needs CI:** `pnpm db:check-drift` and the integration suites (no Docker/Postgres in this environment). `tenant-export-policy.integration.test.ts` is the one to watch: `snmp_devices` is an org-cascade table, so the two new columns had to be classified in `CORE_TENANT_EXPORT_POLICY` or it reds main.

**Second review round intentionally skipped** — the fixes touch worker/route logic, not tenancy, auth, or migrations, so per the repo's rigor-calibration rule one independent round is the cap. Residual pre-existing risks (stalled-job double-count, in-process `orphanedResultExpectations`, no `last_failure_reason`) are documented in the PR body rather than fixed here.

**Status:** review-clean, awaiting maintainer merge. Note the #3215 conflict flagged in the PR body — that PR edits the same file in parallel.

Issue #570, opened 2026-05-05
[Auth] partners.email_verified_at never set on hosted-mode registration flow


ToddHebebrand:
## Description

On hosted production, a partner can reach `partners.status = 'active'`, attach a payment method, enroll an agent, and run an active subscription without `partners.email_verified_at` ever being set.

Observed on partner `8b5d7ff2-e0e5-4209-90f6-04b63c27c4da` on 2026-05-05:

- `partners.status = 'active'`
- `partners.email_verified_at = NULL`
- `partners.payment_method_attached_at = 2026-05-05 12:47:44+00`
- Active \$99/mo Stripe subscription

## Investigation

- `apps/api/src/routes/auth/register.ts:193` — in hosted mode, partner is created with `status: 'pending'`. Something downstream flips it to `'active'` (the registration hook at lines 248-281, or completion of the Stripe SetupIntent).
- Whatever path activates the partner does **not** set `partners.email_verified_at`.
- `grep -rn email_verified apps/api/src` returns only the schema definition (`apps/api/src/db/schema/orgs.ts:28`) and one SSO claim shape (`apps/api/src/services/sso.ts:31`). No write site exists in production code.

## Risk

Email verification is a stated security control (post SR-001..SR-024 hardening pass). If the column is unwritten:

- The "this email was confirmed" signal is `NULL` for the entire partner population.
- Any future code that gates on it (admin flows, password reset trust, deliverability decisions) will misbehave.
- Worst state is "looks like coverage exists, but is null everywhere" — silent control failure.

## Proposed fix

Decide intent first:

1. **Verification required on new-partner flow.** Add a verification-link click handler that sets `partners.email_verified_at`, and gate the `pending → active` transition on it. Likely tied into the registration hook or a new dedicated route.
2. **Verification not required.** Drop the column to remove the false-positive signal.

The "lukewarm" middle option (column exists but is never written) is the worst of both worlds.

## Affected files

- `apps/api/src/routes/auth/register.ts` — partner status transition
- `apps/api/src/services/partnerCreate.ts:71` — initial status
- Verification handler — none currently exists; would need to be added if option 1
- `apps/api/src/db/schema/orgs.ts:28` — column definition; relocates if dropped

## Audit query

If this returns close to the total partner count, confirms the verification path is dead:

\`\`\`sql
SELECT
  count(*) FILTER (WHERE email_verified_at IS NOT NULL) AS verified,
[...]

PR #3224, opened 2026-08-07
fix(api): diagnose and surface the postgres.js pool-poisoning failure (#3214)

### LINKED ISSUE #3214: api: postgres.js pool never recovers dead connections — CONNECT_TIMEOUT storm against a healthy DB until restart

ToddHebebrand:
Closes #3214

## Root cause — found, reproduced, and pinned

The issue listed the `Immediate.nextWrite` TypeError as a *likely* poisoner. It is the whole story, and the mechanism is now reproduced deterministically with no database, no network and no timing luck (`apps/api/src/db/postgresJsPoolPoisoning.test.ts`).

In **postgres.js 3.4.9** (`src/connection.js`), small writes are batched behind a `setImmediate`, guarded by an "already scheduled?" check:

```js
function write(x, fn) {                                        // :246
  chunk = chunk ? Buffer.concat([chunk, x]) : Buffer.from(x)
  if (fn || chunk.length >= 1024)
    return nextWrite(fn)
  nextWriteTimer === null && (nextWriteTimer = setImmediate(nextWrite))
  return true
}

function nextWrite(fn) {                                       // :254
  const x = socket.write(chunk, fn)
  nextWriteTimer !== null && clearImmediate(nextWriteTimer)
  chunk = nextWriteTimer = null                                // <- the ONLY reset
  return x
}
```

`nextWrite` actually running is the only thing that ever resets `nextWriteTimer`. But both teardown paths cancel the immediate **without clearing the variable**, and without dropping the buffered `chunk`:

```js
function terminate()            { ... clearImmediate(nextWriteTimer) ... }        // :427
async function closed(hadError) { ... clearImmediate(nextWriteTimer) ...          // :440
                                      socket = null ... }
```

So a connection whose socket dies with a deferred write pending is left with `nextWriteTimer` permanently non-null. The guard at `:250` is false forever after, no flush is ever scheduled again, and the `StartupMessage` written by `connected()` on **every** subsequent reconnect is appended to `chunk` and never reaches the socket. The handshake cannot complete, `connect_timeout` expires, `connectTimedOut()` reports `CONNECT_TIMEOUT` and destroys the socket, `closed()` reconnects — forever.

That accounts for every symptom in the report:

| Observation | Explanation |
|---|---|
| `TLSSocket.closed` errors starting days early | each one poisons one pool slot |
| pool decay 35 → 9 over hours | slots poisoned cumulatively, never recovered |
| ~144 `CONNECT_TIMEOUT`/min against a healthy DB | poisoned slots reconnect-looping |
| `Immediate.nextWrite` TypeError | the same broken lifecycle (`socket` nulled / stale `chunk`) |
| restart fixes it in seconds | fresh closures |

**Verification of the repro** — the control (undisturbed connection) writes 77 bytes of `StartupMessage`; kill the socket while the flush is queued and the *reconnect* writes **0 bytes**.

## Suggested fix #1 — answered, and it is a dead end

> *check whether postgres.js > 3.4.9 fixes it, or file upstream*

**3.4.9 is the latest published release.** `npm view postgres version` → `3.4.9`. There is nothing to bump to, and the broken state lives inside a closure that nothing outside the driver can reach. The reproduction in this PR is dependency-free and ready to attach to an upstream report.

## What this PR ships (fixes #2 and #3)

**`apps/api/src/services/dbConnectTimeoutStats.ts`** — a rolling `CONNECT_TIMEOUT` count, fed from `safeDiagnoseConnectTimeout` (the wrapper both production call sites already use). Deduped per error object with a `Symbol`, because `app.onError` and `captureException` classify the *same* error — without that the rate would be silently 2x on the request path and 1x on the worker path, making any threshold meaningless. Zero-import leaf, so it does not recreate the `services → db` back-edge the classifier's docblock warns about.

**`apps/api/src/db/dbPoolHealthMonitor.ts`** — the watchdog. A sustained timeout rate on its own is ambiguous: unreachable DB and poisoned pool look identical. So when the rate breaches the threshold it opens **one brand-new connection** to the same database, outside the pool. A fresh client has fresh closures and is immune to the poisoning:

- fresh connection **succeeds** → `pool-degraded`: the DB is fine, the pool is not, **restart the API**
- fresh connection **fails too** → `database-unreachable`: real fault, **restarting will not help**
- probe could not be *attempted* → `unknown`, explicitly refusing to blame either

That is exactly the manual step that ended the incident ("`psql` from the same host connected in under a second"), performed automatically. Steady state costs one unref'd timer tick — the probe only opens a socket once the threshold is already breached.

**`apps/api/src/routes/metrics.ts`** — `breeze_db_connect_timeouts_total{cause}`, `breeze_db_connect_timeout_rate_per_min`, `breeze_db_pool_health{verdict}`. Seeded at 0 so alert rules never query a nonexistent series. All verdict series stay 0 before the first evaluation — absence of a verdict must not read as `healthy`.

Plus `.env.example` + `deploy/environment.mdx` for the six new knobs, and a bootstrap-wiring guard (`dbPoolHealthWiring.test.ts`) — a watchdog that is defined but never started fails completely silently, which is how this class of blindness ships.

## What I did NOT do, and why

- **Recycling the `sql` instance.** The issue offers this as optional; I left it out. `db/index.ts` hands the client to Drizzle at module load, and the resulting `baseDb` is captured by the exported proxy, by every open `withDbAccessContext` transaction, and by the AsyncLocalStorage stores. A mid-flight swap would abort in-flight tenant transactions, and a bug in it is a whole-API outage — a worse failure than the one being fixed. Wants its own PR.
- **Patching the driver in `node_modules`.** The actual repair is ~2 lines (reset `chunk`/`nextWriteTimer` in `closed()` and `terminate()`), and the pin test would prove it. I left it out: this repo has **no** `patchedDependencies` today, so it would be the first — introduced on the DB driver, touching the Docker and CI install surface, and I cannot validate a driver-lifecycle patch against real production traffic from here. It deserves a dedicated PR with deploy validation.
- **Anything that claims to fix the pool.** This PR makes the failure *visible and correctly attributed*. It does not repair it. `pool-degraded` still means someone restarts the API.

The pin test is deliberately asserting the **bug**. The day the driver is fixed or patched, it goes red with a message naming exactly what to delete (the pin and the watchdog).

## Review round

Three passes (`code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`) found real defects, all sharing one shape — the change reintroduced, in its own code, the blindness it exists to remove. Fixed in `914926027`:

- **`captureException` classified connect timeouts *below* its `initialized` guard.** On any instance without a Sentry DSN — the self-hosted default — `app.onError` was the counter's only feed, so every worker, scheduler and `unhandledRejection` path contributed nothing. That is exactly the incident's profile ("the loudest signature in Sentry was the patch scheduler, not any route"), meaning the watchdog would have reported the pool fine right through it. Classification now happens above the guard.
- **The quiet verdict `healthy` is now `below-threshold`.** The count is a documented floor and slot occupancy is never observed, so an affirmative all-clear was a claim the evidence cannot support. There is deliberately no `healthy` series to alert on the negation of.
- **A failed evaluation left the previous verdict standing**, so `/metrics` republished a stale reading for hours about a dead watchdog. It now clears the verdict, counts the failure, and reports itself to Sentry (it was console-only).
- **A probe timeout under event-loop starvation reported `database-unreachable`** with "restarting will not help" attached — the #3022 misdirection one layer up, since the probe's own budget needs the same blocked loop. Now a distinct error type and an `unknown` verdict.
- **In-flight guard + probe budget clamped to half the interval.** Legal env combinations (`INTERVAL=5000, PROBE=30000`) could stack probes, each opening another connection to the database being diagnosed. Probe close failures are now counted, not silently swallowed.
- Stable Sentry title (interpolated rates minted a new issue per capture, so no alert bound to one could fire twice), a suppressed-since-last-capture count, and sample retention that trims to the widest window any reader requested rather than the caller's.
- **The six `DB_POOL_HEALTH_*` vars are threaded through both compose files.** The docs promise they work, and an unmapped var set in `.env` is silently inert (the IS_HOSTED / #570 failure mode).

### Tests

The test pass found that **five mutations survived the entire suite**: deleting the counter feed, the recorder binding, the scrape refresh, the interval body, or the throttle guard all stayed green — i.e. every line of production wiring could be removed without CI noticing. Each is now covered, and I re-applied the mutations to confirm they fail. Also added hermetic tests for the real probe (no database needed) and a **negative control** on the driver pin, so "the reconnect wrote nothing" now demonstrably means "*because* the socket died with a buffered write" rather than for any unrelated reason.

### Second review round (delta re-review)

Because the first round's fixes touched `captureException` — every error path in the API — the delta got its own pass. Three more real findings, fixed in `57898f941`:

- **The Sentry captures were arriving contentless.** `scrubEvent` deletes `message`, `logentry` and `extra` from *every* outgoing event, and my tag key was not in `ALLOWED_TAG_NAMES` — so it was dropped twice over. A `pool-degraded` alert reached Sentry as an unlabelled blank that grouped with every other blank. The verdict is now allowlisted as `db_pool_health_verdict` (the same precedent `connect_timeout_cause` set for #3022), with a scrubber test proving it survives. The docs no longer claim more than Sentry actually receives.
- **The probe-timeout downgrade tested the wrong direction.** It asked "does starvation dominate?", which leaves open exactly the hole it exists to close: with the event-loop monitor disabled or warming up, *every* timeout classifies as `unknown`, so the starvation count is 0, no dominance is found, and the confident `database-unreachable` fires in the configuration with the least evidence. Now inverted to require positive evidence *for* connectivity.
- **A throwing `captureMessage` erased a valid verdict.** It fell through to the outer catch, which (correctly) clears `lastAssessment` — so a reporter fault would blank a real degraded verdict from `/metrics` at the moment it fired.

Plus: the two failure counters were `Gauge`s carrying the `_total` suffix OpenMetrics reserves for counters (renamed), and the check-failed key never drained its suppressed count.

## Verification

- `vitest run src/db src/services src/routes/metrics.test.ts src/config/envComposeParity.test.ts --no-file-parallelism` → **622 files, 10084 passed, 35 skipped**
- `tsc --noEmit -p apps/api/tsconfig.json` (with the CI `--max-old-space-size=8192`) → clean
- `eslint` on all touched files → clean
- Driver pin run 8x consecutively → 8/8 stable (the `setImmediate` ordering is guaranteed by the driver's microtask-only path from socket factory to `connected()`, not merely usual)
- Both compose files parse and carry all six new vars

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -453,6 +453,22 @@ GRAFANA_ADMIN_PASSWORD=changeme
 # EVENT_LOOP_STARVATION_WARN_MS=1000
 # EVENT_LOOP_MONITOR_DISABLED=
 
+# Database pool-health watchdog. On by default; these are tuning knobs.
+# postgres.js can leave a pooled connection permanently unable to reconnect once
+# its socket dies with a buffered write, producing a storm of "write
+# CONNECT_TIMEOUT" errors against a database that is perfectly healthy. The pool
+# cannot self-heal — only an API restart recovers it. When the CONNECT_TIMEOUT
+# rate breaches the threshold below, the watchdog opens ONE brand-new connection
+# to the same database to tell the two cases apart: if that succeeds the pool is
+# at fault, if it fails too the database is. Exposed as breeze_db_pool_health and
+# breeze_db_connect_timeout_rate_per_min on /metrics.
+# DB_POOL_HEALTH_INTERVAL_MS=60000
+# DB_POOL_HEALTH_WINDOW_MS=300000
+# DB_POOL_HEALTH_MIN_TIMEOUTS=10
+# DB_POOL_HEALTH_PROBE_TIMEOUT_MS=5000
+# DB_POOL_HEALTH_CAPTURE_THROTTLE_MS=900000
+# DB_POOL_HEALTH_DISABLED=
+
 # Platform-operator abuse alerts (all optional; unset = feature disabled)
 # OPS_ALERT_WEBHOOK_URL=https://discord.com/api/webhooks/your-webhook
 # OPS_ALERT_EMAIL=ops@your-domain.example.com
--- a/apps/api/src/db/dbPoolHealthMonitor.test.ts
+++ b/apps/api/src/db/dbPoolHealthMonitor.test.ts
@@ -134,7 +134,7 @@ describe('dbPoolHealthMonitor (#3214)', () => {
       expect(result.message).not.toContain('POOL DEGRADED');
     });
 
-    it('reports unknown when the probe times out and starvation dominates the causes', async () => {
+    it('reports unknown when the probe times out under event-loop starvation', async () => {
       // The probe's budget is an in-process setTimeout, so it expires when the
       // main thread is pegged just as readily as when the DB is unreachable —
       // the #3022 ambiguity, one layer up. Calling this `database-unreachable`
@@ -155,13 +155,40 @@ describe('dbPoolHealthMonitor (#3214)', () => {
       });
 
       expect(result.verdict).toBe('unknown');
-      expect(result.message).toContain('event-loop starvation');
+      expect(result.message).toContain('starvation');
       expect(result.message).not.toContain('DATABASE UNREACHABLE');
     });
 
-    it('still reports database-unreachable on a probe timeout when starvation does NOT dominate', async () => {
-      // A genuinely unreachable database also makes the probe time out. Only the
-      // starvation-dominated case is inconclusive.
+    it('reports unknown on a probe timeout when NO cause could be measured at all', async () => {
+      // With EVENT_LOOP_MONITOR_DISABLED (or the monitor still warming up) every
+      // timeout is classified `unknown`. A "does starvation dominate?" test would
+      // see a starvation count of 0, find no dominance, and emit the confident
+      // `database-unreachable` in exactly the configuration where this module has
+      // the LEAST evidence. The verdict must require positive evidence FOR a
+      // connectivity fault, not merely the absence of evidence against one.
+      const result = await assessDbPoolHealth({
+        readStats: () => ({
+          timeouts: 40,
+          byCause: { 'event-loop-starvation': 0, connectivity: 0, unknown: 40 },
+          windowMs: 300_000,
+          ratePerMin: 8,
+          totalSinceStart: 40,
+        }),
+        probe: async () => {
+          throw new DbPoolProbeTimedOutError('pool-health probe exceeded 5000ms');
+        },
+        thresholdTimeouts: 10,
+      });
+
+      expect(result.verdict).toBe('unknown');
+      expect(result.message).not.toContain('DATABASE UNREACHABLE');
+      expect(result.message).not.toContain('Restarting the API will not help');
+    });
+
+    it('still reports database-unreachable on a probe timeout when connectivity dominates', async () => {
+      // A genuinely unreachable database also makes the probe time out, and here
+      // the timeouts themselves were measured as connectivity faults on a healthy
+      // loop — that IS positive evidence, so the confident verdict is warranted.
       const result = await assessDbPoolHealth({
         readStats: () => ({
           timeouts: 40,
@@ -172,6 +199,27 @@ describe('dbPoolHealthMonitor (#3214)', () => {
         }),
         probe: async () => {
           throw new DbPoolProbeTimedOutError('pool-health probe exceeded 5000ms');
+        },
+        thresholdTimeouts: 10,
+      });
+
+      expect(result.verdict).toBe('database-unreachable');
+    });
+
+    it('a driver-reported failure is database-unreachable regardless of cause mix', async () => {
+      // Only a TIMEOUT is ambiguous. An ECONNREFUSED came from the network stack,
+      // not from an in-process timer, so it stands on its own even when every
+      // timeout was classified `unknown`.
+      const result = await assessDbPoolHealth({
+        readStats: () => ({
+          timeouts: 40,
+          byCause: { 'event-loop-starvation': 0, connectivity: 0, unknown: 40 },
+          windowMs: 300_000,
+          ratePerMin: 8,
+          totalSinceStart: 40,
+        }),
+        probe: async () => {
+          throw new Error('connect ECONNREFUSED 10.0.0.5:5432');
         },
         thresholdTimeouts: 10,
       });
@@ -231,7 +279,7 @@ describe('dbPoolHealthMonitor (#3214)', () => {
         expect.stringContaining('POOL DEGRADED'),
         'warning',
         expect.objectContaining({ verdict: 'pool-degraded', timeouts: 40 }),
-        { dbPoolHealthVerdict: 'pool-degraded' },
+        { db_pool_health_verdict: 'pool-degraded' },
       );
     });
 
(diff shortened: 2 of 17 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1 on the full diff (code-reviewer, silent-failure-hunter, pr-test-analyzer), round 2 on the fix delta (code-reviewer, because the round-1 fixes touched `captureException`, which is on every error path in the API).

**Findings:** 11 raised, all addressed — 8 in `914926027`, 3 in `57898f941`; 0 outstanding. The load-bearing ones all shared one shape: the change had reintroduced, in its own code, the blindness it exists to remove.

- The connect-timeout counter was blind on every instance **without a Sentry DSN** (the self-hosted default) — `app.onError` was its only feed, so workers and schedulers counted nothing. That is exactly the incident's own profile, so the watchdog would have reported the pool fine right through it.
- The Sentry captures were arriving **contentless** — `scrubEvent` strips message and extras from every event, and the verdict tag was not allowlisted, so it was dropped twice over.
- `healthy` was renamed to `below-threshold`: the count is a documented floor and slot occupancy is never observed, so an affirmative all-clear was a claim the evidence cannot support.
- A failed evaluation left a **stale verdict** asserted on `/metrics` indefinitely; a probe timeout under a blocked event loop was reported as `database-unreachable` with "restarting will not help" attached.

**Tests:** the coverage pass found **five mutations surviving the whole suite** — deleting the counter feed, the recorder binding, the scrape refresh, the interval body, or the throttle guard all stayed green, i.e. every line of production wiring could be removed without CI noticing. All five are now covered, verified by re-applying each mutation. Added hermetic tests for the real probe (no DB required) and a negative control on the driver pin, so "the reconnect wrote nothing" now demonstrably means *because* the socket died with a buffered write.

Local: `vitest run src/db src/services src/routes/metrics.test.ts src/config --no-file-parallelism` → 622 files, 10084 passed, 35 skipped. `tsc --noEmit` clean, `eslint` clean.

**CI:** 53/53 required checks green (Test API, Test Web, Type Check, Lint all pass); one optional `Test Agent (Windows)` job still running, unrelated to this change.

**One thing to decide before merging:** this PR ships *diagnosis*, not a repair — `pool-degraded` still means a human restarts the API. The actual fix is a ~2-line change to `postgres.js`'s `closed()`/`terminate()`, which I deliberately did not ship (it would be this repo's first patched dependency, on the DB driver, and I can't validate a driver-lifecycle patch against production traffic from here). The repro test in this PR is the acceptance test for that follow-up, and it goes red the day the patch lands. Worth filing as a separate issue.

Issue #3249, opened 2026-08-08
Epic: RMM migration onboarding — make moving off Datto/Ninja/Automate/Kaseya/N-central onto Breeze a supported path


ToddHebebrand:
## Why

Migrating off an incumbent RMM is the single largest barrier to an MSP adopting Breeze. Nobody evaluates an RMM in isolation — they evaluate it against the cost of moving 1,500 endpoints and 60 customers onto it without losing visibility on any of them.

The vendor-neutral playbook and the per-vendor guides now ship at `apps/docs/src/content/docs/migration/` (Datto RMM, NinjaOne, ConnectWise Automate, Kaseya VSA, N-able N-central, Atera, Syncro, plus Pulseway/Action1/Automox/Level/Tactical/ScreenConnect). Writing them surfaced exactly where the product makes a migration harder than it needs to be.

## What we already do well

Worth stating, because it shapes the priorities below:

- **Agent mass enrollment is strong.** Enrollment keys support `maxUsage` up to 100,000 and TTL up to 365 days, signed MSI with properties, ZIP bundles, token-authenticated installer links, short links, and pre-staging for golden images. Nothing needs to change here for a migration wave.
- **Competing-RMM detection already ships.** `agent/internal/mgmtdetect/` fingerprints 11 competing RMMs on every enrolled endpoint. This is a genuine differentiator for the highest-stakes step of a migration, and it is currently invisible outside a single device tab.
- **The read API is broad**, and the org/site/enrollment-key write endpoints exist and behave correctly — they are just single-record and hard to reach unattended.

## The gaps

| # | Gap | Phase it blocks |
|---|---|---|
| #3242 | No bulk import for organizations or sites | Phase 2 — rebuild the tenancy tree |
| #3243 | No machine-to-machine credential can provision tenancy (JWT + MFA only) | Phase 2 — blocks scripting the workaround |
| #3244 | Management Posture has no fleet-wide migration/decommission report | Phases 4 and 6 — verify and cut over |
| #3245 | No script library import/export | Phase 5 — rebuild content |
| #3246 | PSA `getCompanies()` implemented everywhere, called nowhere | Phase 2 — the zero-effort path for ConnectWise/Autotask shops |
| #3247 | API reference wrongly documents `X-API-Key` as general auth | Blocks anyone scripting anything |
| #3248 | No GPO / Intune / JAMF deployment guide | Steady state after migration |

## Suggested order

[...]

PR #3251, opened 2026-08-08
docs(reference): correct the X-API-Key auth claim — it is not general-purpose (#3247)

### LINKED ISSUE #3247: [Docs] API reference wrongly documents X-API-Key as general-purpose auth — the /devices example returns 401
## The docs are wrong

`apps/docs/src/content/docs/reference/api.mdx:17-31` says:

> All API requests (except enrollment, health, SSO, and portal) require a JWT bearer token
> ...
> API keys can also be used via the `X-API-Key` header:
> ```bash
> curl -H "X-API-Key: brz_..." https://breeze.yourdomain.com/api/v1/devices
> ```

The example does not work. `GET /api/v1/devices` returns **401**.

## Why

`apiKeyAuthMiddleware` (`apps/api/src/middleware/apiKeyAuth.ts:86`) is mounted on exactly three surfaces:

- `apps/api/src/routes/mcpServer.ts:219`
- `apps/api/src/routes/devPush.ts:79`
- `apps/api/src/routes/devices/customFieldValues.ts:87`

Every other route uses `authMiddleware` (`apps/api/src/middleware/auth.ts:425`), which reads only `Authorization` and throws `401 Missing or invalid authorization header` when it is absent or not `Bearer `-prefixed. `X-API-Key` is never consulted. `apps/api/src/routes/devices/core.ts:268` applies `authMiddleware` to the whole devices surface, so the documented example is specifically one of the calls that fails.

## Impact

This is the first page anyone reads when scripting against Breeze, and it sends them down a path that cannot work. Someone integrating — or migrating a fleet, where scripting the API is mandatory because there is no bulk import — will mint an API key, get 401s on every call, and have no way to tell from the docs that the key was never going to work.

## Fix

1. Correct `reference/api.mdx` to state that `X-API-Key` is accepted only on the MCP server, dev-push, and device custom-field-value endpoints, and that everything else requires a user JWT. Remove the `/devices` example or replace it with one that works.
2. Check `apps/docs/src/content/docs/reference/api-keys.mdx` for the same claim and correct it.
3. Note the `requireMfa()` gate on the tenancy write routes (`POST /orgs/organizations`, `POST /orgs/sites`, `POST /orgs/partners`), since that is the other thing that silently breaks a scripted integration.

`apps/docs/src/content/docs/migration/toolkit.mdx` documents the accurate behaviour under "Authentication for Migration Scripts" and can be used as the source of truth.

bdunncompany:
Fixes #3247 — the top item in your #3249 ordering, and the one that costs hours rather than days.

## What was wrong

`reference/api.mdx` said an API key authenticates the REST API and used `GET /api/v1/devices` as the example. That call returns `401 Missing or invalid authorization header`.

Confirmed against `origin/main` rather than taken from the issue: `authMiddleware` (`middleware/auth.ts:448`) reads the `Authorization` header and nothing else, throwing at `:451` when it is absent or not `Bearer`-prefixed. `devices/core.ts:276` puts the entire devices surface behind it, so the documented example is specifically one of the calls that cannot work.

## Where keys actually work

Three surfaces, all dual-auth — key when `X-API-Key` is present, user JWT when it is absent:

| Endpoint | Scope | Mount |
|---|---|---|
| `/api/v1/mcp/*` | `ai:*` | `index.ts:1114`, `mcpServer.ts:221` |
| `POST /api/v1/dev/push` | `devices:execute` | `index.ts:1115`, `devPush.ts:79` |
| `GET /api/v1/devices/:id/custom-fields` | `devices:read` | `devices/customFieldValues.ts:87` |
| `PATCH /api/v1/devices/:id/custom-fields` | `devices:write` | same |

Both reference pages now carry that table, and the broken `/devices` example is replaced by one that works. I added the distinction that made this trap easy to fall into: **scopes gate what a key may do on those routes, not where it is accepted.** A `devices:read` key still cannot call `GET /api/v1/devices`, which is exactly the wrong inference a scope list invites.

Also covered your third point — the `requireMfa()` gate on `POST /orgs/partners` (`orgs.ts:394`), `POST /orgs/organizations` (`:1279`) and `POST /orgs/sites` (`:1855`), so nobody discovers mid-migration that an MFA-less service account cannot create tenancy.

## Two line-number corrections

Not substantive, but so the issue isn't a false trail later: `apiKeyAuthMiddleware` is mounted at `mcpServer.ts:221` (not 219) and the devices `authMiddleware` is at `devices/core.ts:276` (not 268). Everything else in the issue checked out exactly, including that the three mount sites are the only non-test ones — `servicePrincipals.ts:20` even carries a comment asserting it never imports it.

## On the cross-link I did not add

Your issue points at `migration/toolkit.mdx` as the source of truth for the accurate behaviour, and it is — but that page only exists in #3250, which is still open. Linking to it from here would ship a dead link until that merges, so I kept the MFA guidance self-contained. Worth adding the cross-link when #3250 lands; happy to do it in that PR or a follow-up, whichever you prefer.

## Verification

- `astro build` — exit 0, 150 pages, both pages render
- the new `#where-api-keys-work` heading generates its anchor and the in-page link resolves to it (checked in the built HTML, not assumed)
- `/features/mcp-server/` link target exists in `dist`
- `pnpm test:docs-automation` — 5/5 pass

Diff is two `.mdx` files; no lockfile or manifest churn.

### DIFF
--- a/apps/docs/src/content/docs/reference/api-keys.mdx
+++ b/apps/docs/src/content/docs/reference/api-keys.mdx
@@ -87,11 +87,13 @@ Scopes can be updated after creation via the `PATCH` endpoint without rotating t
 
 ### Authentication Flow
 
-To authenticate with an API key, include it in the `X-API-Key` header:
+To authenticate with an API key, include it in the `X-API-Key` header — on one of
+the routes that accept it (see [Where API keys work](#where-api-keys-work) below;
+the general REST API is not one of them):
 
 ```bash
 curl -H "X-API-Key: brz_aBcDeFgHiJkLmNoPqRsTuVwXyZ012" \
-  https://breeze.yourdomain.com/api/v1/devices
+  https://breeze.yourdomain.com/api/v1/devices/$DEVICE_ID/custom-fields
 ```
 
 The middleware performs these steps in order:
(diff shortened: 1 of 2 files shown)


bdunncompany:
Superseded by #3286 — closing this rather than leaving a duplicate.

Yours is the better version and should land instead: it covers the same two files, adds `openapi.ts`, and documents the `brz_sp_` partner service-principal class, which did not exist when I wrote this (it arrived with #3274 earlier today). Your MFA framing is also stronger than mine — I stated the gate; you explain *why* it makes user JWTs a poor fit for automation and point at the credential class that isn't gated.

One nuance from mine that I did not find in #3286, offered only in case you want it — no need to act on it:

> Scopes gate what a key may **do** on those routes, not **where** it is accepted. A `devices:read` key still cannot call `GET /api/v1/devices`.

That was the specific wrong inference I expected a reader to draw: a key page listing `devices:read` as a scope reads as though it grants the devices API. A sentence ruling it out closes the loop, but the surface table in #3286 already conveys it structurally, so this is a nicety rather than a gap.

PR #3263, opened 2026-08-08
fix(api): gate partner-wide script writes on canManagePartnerWidePolicies (#3262)

### LINKED ISSUE #3262: [API][Security] POST /scripts allows partner-wide script creation without canManagePartnerWidePolicies — 'selected'-access users can push SYSTEM-level code to every org
## Problem

`POST /scripts` lets a partner-scoped user create a **partner-wide script** without checking `canManagePartnerWidePolicies`. Scripts execute as SYSTEM on every managed endpoint, and a partner-wide script applies to **every organization under the partner, including organizations created later**.

The gate in `apps/api/src/routes/scripts.ts:502-506` is scope-only:

```ts
} else if (auth.scope === 'partner') {
  if (data.availability === 'partner') {
    orgId = null;
    partnerId = auth.partnerId ?? null;
    if (!partnerId) return c.json({ error: 'Partner context required' }, 403);
  }
```

Confirmed absent from the file:

- `grep -c canManagePartnerWidePolicies apps/api/src/routes/scripts.ts` → **0**
- `grep -c partnerOrgAccess apps/api/src/routes/scripts.ts` → **0**

## Why scope alone is insufficient

`services/partnerWideAccess.ts` is explicit that this exact substitution is the bug it exists to prevent:

```ts
export function canManagePartnerWidePolicies(auth) {
  return auth.scope === 'system' || (auth.scope === 'partner' && auth.partnerOrgAccess === 'all');
}
```

with the doc comment: *"Deliberately NOT derivable from accessibleOrgIds: a 'selected' user whose selection happens to cover every current org still must not administer partner-wide state."*

So a partner user with `partner_users.org_access = 'selected'` — scoped to, say, three of a partner's eighty customers — can today create a script that runs as SYSTEM across **all eighty**, including customers they have no grant for and customers onboarded next year.

CLAUDE.md names this contract directly: *"gate partner-wide create/update/delete on `canManagePartnerWidePolicies(auth)` (`services/partnerWideAccess.ts` — the single source of truth)."*

## This is inconsistent with the rest of the codebase

Routes that do it correctly: `peripheralControl.ts:508`, `accessReviews.ts`, `partnerLoginBranding.ts`, `huntress.ts`, `sensitiveData.ts`, `networkKnownGuests.ts`, `pax8.ts`, `roles.ts`, `sso.ts`. Scripts is the outlier — and it is the one carrying arbitrary code execution.

## Fix

1. Gate partner-wide create on `canManagePartnerWidePolicies(auth)`, returning `PARTNER_WIDE_WRITE_DENIED_MESSAGE` with 403 — matching `peripheralControl.ts:508`.
2. Apply the same gate to **update and delete** of a script whose `orgId IS NULL` (partner-owned). A user who cannot create one must not be able to edit or remove one either.
3. Audit existing partner-wide scripts (`orgId IS NULL AND partner_id IS NOT NULL`) and check whether any were created by a `selected`-access user.
4. Regression test: partner scope + `partnerOrgAccess: 'selected'` + `availability: 'partner'` → 403.

## Severity

Partner-scoped, not cross-partner — RLS still confines it to the caller's own partner, so this is not a tenant-isolation breach. Within a partner it is a **privilege-escalation path to SYSTEM-level code execution across orgs the user was never granted**, which is why it is worth fixing ahead of the feature work that would amplify it.

## Found via

Adversarial design review of #3245 (script bundle import), which proposed `availability: 'partner'` as the recommended default for imported bundles — that would fan a bundle through this same ungated path. #3245 now depends on this being fixed first. Epic #3249.

bdunncompany:
Fixes #3262.

Partner **scope** was standing in for the partner-wide **capability**. `canManagePartnerWidePolicies` appeared nowhere in `scripts.ts`, and its doc comment names this exact substitution as the thing it exists to prevent.

## Four sites, not one

Your issue names the create path and asks for update/delete too. Building it surfaced a fourth that isn't in the issue — and it's the one I'd most expect to be missed:

| # | Vector | Prior guard | Gap |
|---|---|---|---|
| 1 | **Create** with `availability: 'partner'` | `auth.scope === 'partner'` | the reported vector |
| 2 | **Widen** an org script to partner-wide on edit | `resolveRescopeTarget` checked scope only | **a second creation vector for the same privilege** |
| 3 | **Edit** an existing partner-wide script | blocked only `auth.scope === 'organization'` | a partner `selected` user could rewrite code already running as SYSTEM everywhere |
| 4 | **Delete** an existing partner-wide script | same guard, same gap | removes automation from every org under the partner |

Sites 3 and 4 already had a partner-wide guard — it just tested the wrong thing. It kept *org-scoped* users out and let every partner-scoped user through regardless of `org_access`.

Site 2 is the one worth a second look on review: `availability: 'partner'` on `PUT /:id` reaches partner-wide state without ever touching the create handler, so gating create alone would have left the privilege fully reachable.

## Implementation

Follows `peripheralControl.ts:508` verbatim — `PARTNER_WIDE_WRITE_DENIED_MESSAGE` with 403 — matching the nine other routes you listed as doing it correctly.

`RescopeAuth` gains `partnerOrgAccess` (the capability is deliberately not derivable from `accessibleOrgIds`, so it has to be carried) and its `scope` narrows from `string` to `AuthContext['scope']` so the capability check typechecks.

## Tests, and the control that mattered

Five: four denials — create, widen, edit, delete as a partner user with `partnerOrgAccess: 'selected'` — plus a positive control proving a **full-partner admin can still widen**, so the denials can't pass on a handler that simply rejects every partner write.

The two existing partner-auth helpers now default to `partnerOrgAccess: 'all'`. That is the user those positive cases always meant to describe; without it the gate would have failed pre-existing tests, which is itself a small signal the fixture was under-specified.

All five were run against the un-fixed code. **My first control attempt was wrong and worth flagging**: I neutered the guards with a regex that matched only two of the four call sites — the edit/delete guards use a different expression shape — so only two denials failed and I would have shipped two unproven tests. Neutering `canManagePartnerWidePolicies` itself to `return true` gives the honest result:

```
× create as 'selected'   FAIL
× widen  as 'selected'   FAIL
× edit   as 'selected'   FAIL
× delete as 'selected'   FAIL
✓ widen  as 'all'        PASS   (control still passes)
```

## Gate

`tsc --noEmit` exit 0 · `vitest run` 1285 files / 20525 tests / 0 failures · `eslint` clean on both changed files.

## Left out deliberately

Your point 3 — auditing existing `orgId IS NULL AND partner_id IS NOT NULL` scripts for any created by a `selected`-access user — is a query against your deployment rather than a code change, so it isn't in this diff.

### DIFF
--- a/apps/api/src/routes/scripts.test.ts
+++ b/apps/api/src/routes/scripts.test.ts
@@ -1045,6 +1045,51 @@ describe('scripts routes', () => {
       expect(insertedValues?.partnerId).toBe(PARTNER_ID);
     });
 
+    // #3262 control: the capability gate covers ONLY partner-wide writes. A
+    // selected-access user must keep every org-scoped ability — without this,
+    // a refactor that hoists the gate above the availability branch (denying
+    // ALL partner-scope writes) would pass the suite green.
+    it('#3262: a selected-access partner user still creates an org-scoped script', async () => {
+      const { authMiddleware } = await import('../middleware/auth');
+      vi.mocked(authMiddleware).mockImplementationOnce((c: any, next: any) => {
+        c.set('auth', makePartnerAuth('selected'));
+        return next();
+      });
+
+      let insertedValues: any;
+      vi.mocked(db.insert).mockReturnValue({
+        values: vi.fn().mockImplementation((vals: any) => {
+          insertedValues = vals;
+          return {
+            returning: vi.fn().mockResolvedValue([{
+              id: SCRIPT_ID_1,
+              name: 'Org Script',
+              orgId: ORG_ID,
+              partnerId: PARTNER_ID,
+              isSystem: false
+            }])
+          };
+        })
+      } as any);
+
+      const res = await app.request('/scripts', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json', Authorization: 'Bearer valid-token' },
+        body: JSON.stringify({
+          name: 'Org Script',
+          osTypes: ['windows'],
+          language: 'powershell',
+          content: 'echo hi',
+          orgId: ORG_ID,
+          availability: 'org'
+        })
+      });
+
+      expect(res.status).toBe(201);
+      expect(insertedValues?.orgId).toBe(ORG_ID);
+      expect(insertedValues?.partnerId).toBe(PARTNER_ID);
+    });
+
     it('org user editing a partner-wide script (org_id=null, partner_id set) → 403', async () => {
       const { authMiddleware } = await import('../middleware/auth');
       vi.mocked(authMiddleware).mockImplementationOnce((c: any, next: any) => {
@@ -1326,6 +1371,28 @@ describe('scripts routes', () => {
 
       expect(res.status).toBe(200);
       expect(getSet().orgId).toBeNull();
+    });
+
+    // Control for the DELETE gate: it must deny on capability, not on partner
+    // scope generally — a gate that 403'd every partner delete of a
+    // partner-wide script would pass the denial tests above.
+    it('#3262: a full-partner admin still deletes a partner-wide script', async () => {
+      await withAuth(makePartnerAuth('all'));
+      // Serves the script lookup AND the active-executions count (0 = none).
+      mockScriptLookup({
+        id: SCRIPT_ID_1, name: 'Partner Script', orgId: null, partnerId: PARTNER_ID,
+        isSystem: false, content: 'echo hi', version: 1,
+      }, 0);
+      const getSet = captureUpdate({ id: SCRIPT_ID_1 });
+
+      const res = await app.request(`/scripts/${SCRIPT_ID_1}`, {
+        method: 'DELETE',
+        headers: { Authorization: 'Bearer valid-token' },
+      });
+
+      expect(res.status).toBe(200);
+      // Soft delete: the handler stamps deletedAt rather than issuing a DELETE.
+      expect(getSet().deletedAt).toBeInstanceOf(Date);
     });
 
     it('partner user moves a script org→org (when no references exist)', async () => {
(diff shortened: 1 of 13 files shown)


ToddHebebrand:
Thanks @bdunncompany — solid work catching the fourth vector (widen-via-rescope) that the issue didn't name. CI was green and a high-effort review confirmed the gating logic is correct across all four paths.

I pushed one maintainer commit (1cfe0e8) on top with two defense-in-depth hardenings the review flagged, so the whole fix lands atomically:

- **Same-partner ownership check in the app layer.** The new PUT/DELETE guards checked the capability but relied on RLS row-invisibility alone for cross-partner containment. There's no exploit today, but any future refactor that loads scripts under a system DB context (the #1105 pattern) would have turned that into cross-tenant write/delete. The guard now returns 404 (not a leaky 403) when `script.partnerId !== auth.partnerId` — one pre-existing rescope test updated accordingly since the ownership guard now fires first.
- **`RescopeAuth.partnerOrgAccess` is now a required key** so future callers can't omit it and silently compile into fail-closed 403s for genuine full-partner admins.
- Both route guards now share one `partnerWideScriptWriteError` helper so PUT and DELETE can't drift.

Will merge once CI is green on the new commit.

bdunncompany:
Thanks — and both hardenings are right, particularly the second one.

Making `RescopeAuth.partnerOrgAccess` required is the correction I should have made myself. I typed it optional because `AuthContext.partnerOrgAccess` is optional and the required form failed `tsc`, so I took the change that silenced the error rather than the one that preserved the invariant. Optional is the strictly worse choice here: a future caller who omits it gets `undefined`, fails `canManagePartnerWidePolicies` closed, and hands a genuine full-partner admin a spurious 403 — a failure that compiles clean and looks like a permissions bug rather than a missing field. Required forces the caller to think.

The same-partner 404 is a gap I didn't consider at all. I reasoned about the capability axis and treated cross-partner containment as settled by RLS, which is true today and exactly the assumption a future system-context refactor would quietly invalidate — the #1105 shape. Worth me carrying forward: on a tenancy guard, check the ownership axis in the app layer too rather than delegating it to RLS, even when RLS currently covers it.

Nothing needed from me here — flagging only that I read it, since the comment landed just after the squash-merge.

PR #3266, opened 2026-08-08
fix(backup): short-retry ACL denials and abort on a lost VSS snapshot (#3259, #3260)

### LINKED ISSUE #3259: [Agent][Backup] Plain ERROR_ACCESS_DENIED still burns 30s per file — #3002 only fast-skips the cloud-file variant
## Description

#3002 taught the upload retry path to fast-skip permanently-failing files, and #2997 covered the OneDrive cloud-only case (`ERROR_CLOUD_FILE_ACCESS_DENIED`, 395). But an **ordinary NTFS ACL denial** — `ERROR_ACCESS_DENIED` (5), the common real-world case — is still treated as transient and burns the full 30-second retry sleep per file.

Reproduced on **v0.104.0** (`a3dc568ec`) against a real Windows Server 2022 agent, 2026-08-07. No OneDrive involved — just `icacls /deny` on plain files.

## Measured

Same 40-file source folder each time, only the number of DENY-ACL'd files varies:

| Denied files | Elapsed | Delta |
|---|---|---|
| 0 (control) | **10.13 s** | — |
| 5 | **156.67 s** | +29.3 s per denied file |
| 15 | **415.24 s** | +27.0 s per denied file |

Helper log shows the pattern plainly: `file upload failed, retrying once` → 30 s later → `elapsedMs=30002`.

## Root cause

`classifyPermanentUploadError` (`agent/internal/backup/upload_error_class.go:50-54`) **deliberately excludes** plain `ERROR_ACCESS_DENIED (5)` from the permanent set, with the reasoning that it "can be a transient AV/indexer hold". Only these fast-skip:

- `ERROR_CLOUD_FILE_ACCESS_DENIED` (395) — the #2997 case
- `ERROR_FILE_NOT_FOUND` / `ERROR_PATH_NOT_FOUND`
- `ERROR_SHARING_VIOLATION` / `ERROR_LOCK_VIOLATION`

Everything else falls through to `uploadRetryDelay = 30 * time.Second` (`agent/internal/backup/snapshot.go:222`).

The AV/indexer-hold reasoning is sound in the abstract, but it means the single most common cause of an unreadable file — a permissions denial that will *never* succeed on retry — pays the full transient-retry cost.

## Why it matters beyond slowness

This is not only a throughput problem. The accumulated stall can outlive the VSS shadow copy and destroy the entire backup — see the linked issue for a run where 15 denied files caused **0 of 40** files to be backed up.

## Proposed fix

Distinguish a retryable access denial from a permanent one rather than treating all of `ERROR_ACCESS_DENIED` as transient. Options, roughly in order of preference:

1. Probe the ACL once (e.g. attempt an open with `FILE_READ_ATTRIBUTES`); if the denial is an ACL rather than a transient share/AV hold, classify permanent and skip immediately.
2. Retry `ERROR_ACCESS_DENIED` **once with a short delay** (~1 s) rather than the full 30 s — an AV/indexer hold clears in well under a second; an ACL never will.
3. Make `uploadRetryDelay` scale down after the first few identical failures in a run.

## Affected files

- `agent/internal/backup/upload_error_class.go` (lines 50-54, primary)
- `agent/internal/backup/snapshot.go` (line 222, retry delay)

## Related

- #2997 — the OneDrive cloud-only variant (395), which #3002 did fix
- #3002 — the fix that introduced the fast-skip set this issue extends

## Provenance

Internal release QA for v0.104.0, 2026-08-07. Verified job ids in the QA stack: `2f4b9708` (clean control), `05f3b1e3` (5 denied, `partial`), `b906f40d` (15 denied, total loss).

### LINKED ISSUE #3260: [Agent][Backup] DATA LOSS: upload stalls outlive the VSS shadow copy — ~15 denied files destroys the backup of every healthy file
## Description

A handful of ACL-denied files in a backup set can cause **the entire backup to store nothing**, including files that were perfectly readable.

The 30-second-per-file retry stall (#3259) accumulates until it outlives the VSS shadow copy created at the start of the run. Once the shadow copy becomes unresolvable, **every remaining file fails** with `ERROR_PATH_NOT_FOUND` — not because anything is wrong with those files, but because the snapshot they were being read from is gone.

Reproduced on **v0.104.0** (`a3dc568ec`), real Windows Server 2022 agent, 2026-08-07.

## What happened

Job `b906f40d` — 40-file source, 15 of them `icacls /deny`:

1. Files 1–13 each burn ~30 s on the denied-file retry path.
2. At roughly **+390 s** from creation, shadow copy `HarddiskVolumeShadowCopy26` stops resolving.
3. The tell is file0013's retry **changing error mid-flight**: `Access is denied` on the first attempt → `The system cannot find the path specified` on the retry.
4. **Files 14–40 — all readable, none denied — then fail with `ERROR_PATH_NOT_FOUND`.**

Final state: `file_count` **NULL**, `total_files` 40, elapsed 415 s, `status = failed`. **Zero files backed up.**

## Severity

This is silent data loss in the product's data-protection feature:

- The trigger is ordinary NTFS permissions, not an exotic configuration. Any `C:\Users` or application-data folder with a few locked-down subfolders qualifies.
- The blast radius is the **whole job**, not the denied files. A customer with 15 unreadable files out of 40 gets nothing at all.
- The larger the healthy set, the more is lost — the healthy files are simply the ones that had not been reached yet when the snapshot expired.

## Secondary defect — the failure is unidentifiable

The job did not even land cleanly. There is no `backup run finished` line, and `error_log` reads:

```
Malformed backup result payload: Invalid input: expected object, received null
```

That message joins only `issue.message` and **drops the Zod `path`**, so the null field cannot be identified from the operator-visible error. Worth fixing independently — a malformed-payload error that doesn't say *which* field is malformed is unactionable.

## Proposed fix

Two layers, both worth doing:

1. **Remove the trigger** — fast-skip permanent access denials so the stall never accumulates (#3259).
2. **Defend the snapshot** — a backup must not silently continue reading from a shadow copy that has gone away. Detect the transition (the `ACCESS_DENIED` → `PATH_NOT_FOUND` flip is a clean signal), and either refresh the snapshot or abort with an explicit "shadow copy expired mid-run" error. Continuing to walk a dead snapshot and recording every subsequent file as failed is the worst available outcome: it looks like the files were bad.

Consider also bounding total retry wall-clock against the shadow copy's expected lifetime, so a run cannot outlive its own snapshot.

## Affected files

- `agent/internal/backup/snapshot.go` (retry delay + snapshot lifetime)
- `agent/internal/backup/upload_error_class.go` (the trigger, see #3259)
- wherever `Malformed backup result payload` is constructed (drop-the-`path` bug)

## Related

- #3259 — the 30 s-per-denied-file stall that creates the runway for this
- #2997, #3002 — earlier work on the same retry path

## Provenance

Internal release QA for v0.104.0, 2026-08-07. Job `b906f40d` in the QA stack; DB row verified (`file_count` NULL, 415 s, `failed`).

ToddHebebrand:
Fixes #3259
Fixes #3260

Two coupled backup defects from v0.104.0 release QA on a real Windows Server 2022 agent. The first wastes wall-clock; the second turns that wasted wall-clock into total data loss. Same files, one PR.

## #3259 — plain `ERROR_ACCESS_DENIED` burned ~30s per file

#3002/#2997 taught the per-file upload retry to fast-skip permanently-failing files, but deliberately excluded plain `ERROR_ACCESS_DENIED` (5) because it "can be a transient AV/indexer hold". So the single most common cause of an unreadable file — an ordinary NTFS ACL — paid the full 30s backoff. Measured **+27-29s per denied file**.

**The ACL probe (the issue's first choice) was NOT taken.** `FILE_READ_ATTRIBUTES` and `FILE_READ_DATA` are distinct rights, and a failed probe still cannot prove the original denial came from the DACL rather than an AV/EDR minifilter. There is no discriminator that cannot produce a false "permanent" verdict — and a false permanent verdict silently drops a **healthy** file from a backup, which is precisely the failure class of #3260. Took the issue's option 2 instead.

`classifyPermanentUploadError` (bool) becomes **`classifyUploadFailure`**, returning an `uploadRetryPolicy`:

| policy | when | wait |
|---|---|---|
| `skipWithoutRetry` | not-found, sharing/lock violation, cloud placeholder | none |
| `retryAfterShortDelay` | **source permission denial (new)** | 1s |
| `retryAfterDefaultDelay` | everything unrecognised | 30s |

The retry itself — the thing that actually recovers a transient hold — is preserved; only the wait shrinks. Anything unrecognised still gets the full backoff, so a misclassification costs wall-clock, never a file.

Matched two ways: via the Win32 table (`ERROR_ACCESS_DENIED`, so logs carry the precise symbolic name) and portably via `fs.ErrPermission` (EACCES/EPERM on Unix). **The table is consulted first** so `ERROR_SHARING_VIOLATION` — which Go also maps to `fs.ErrPermission` — keeps its `skipWithoutRetry`. A Windows-only test pins that ordering.

## #3260 — the stalls outlived the shadow copy and lost everything

With 15 denied files of 40, the accumulated backoff ran ~390s, the shadow copy stopped resolving, and files 14-40 — **every one readable** — failed with `ERROR_PATH_NOT_FOUND`. Because PATH_NOT_FOUND is in the fast-skip set, the run drained its remaining file list at memory speed and recorded all of it as per-file failures: 0 of 40 backed up, `status=failed`, an error log that blamed the files.

New `agent/internal/backup/source_liveness.go` adds a probe over the shadow-copy device roots the run reads from, consulted **only after a per-file upload has already failed** — before the retry sleep, and again after it, since the retry is exactly where the reported `ACCESS_DENIED -> PATH_NOT_FOUND` flip happened. A root that has gone away aborts the run with an explicit error naming the root and the progress counts, instead of condemning every remaining file.

Three properties keep the abort from becoming its own false-positive bug:

- **Self-calibrating** — each root is stat'd at construction; only roots that resolved *then* are watched. The probe can never fail on a path form `os.Stat` could not resolve in the first place (`\\?\GLOBALROOT\...` is an unusual shape, and a probe that assumed it stats cleanly would abort every healthy backup if that assumption were wrong).
- **Confirmed** — a root must be missing on two looks, separated by a short delay, before a run dies.
- **Per-volume** — scoped to the failing file's own root, so losing volume D never aborts files still being read from volume C.

Only a not-exist answer counts; any other stat error is inconclusive and treated as alive.

**Abort semantics: mirrors `abortStopped`, does not publish a partial manifest.** With a journal, the remote prefix plus the journal remain this run's resume state, so the already-uploaded bytes survive and the next run picks them up. Publishing a manifest for the partial set would mint a snapshot that *looks* complete while silently missing every file the loop never reached — a restore point that lies. Flagging this as the one design call worth a second opinion.

**No cumulative retry wall-clock budget**, deliberately. A `VSS_CTX_BACKUP` shadow copy has no TTL, so age does not predict validity and any fixed budget would be an arbitrary guess. The liveness probe is the real defense.

## Secondary (API) — malformed-payload errors named no field

All three `Malformed backup result payload` sites (`jobs/backupWorker.ts`, `routes/agentWs.ts`, `services/commandResultHandlers.ts`) joined only `issue.message` and dropped the Zod `path`, so the reported `Invalid input: expected object, received null` was unactionable.

Adds `describeZodIssues` to `apps/api/src/lib/zodIssues.ts` — reusing the existing flattening, which also unwraps `invalid_union` noise — and uses it at all three sites rather than adding a fourth ad-hoc `.map(i => ...)`. **An empty path renders as `<root>` rather than being omitted**: that is the distinction that makes "the whole payload was null" legible instead of ambiguous.

## Verification

| check | command | result |
|---|---|---|
| backup pkg | `cd agent && go test -race -count=1 ./internal/backup/...` | 7 packages ok |
| full agent | `cd agent && go test -race -count=1 ./...` | exit 0, 66 packages, 0 failures |
| windows build | `GOOS=windows GOARCH=amd64 go build ./...` | clean |
| windows tests compile | `GOOS=windows GOARCH=amd64 go vet ./internal/backup/...` | clean (only pre-existing `unsafe.Pointer` notes in untouched `vss_windows.go`) |
| linux build | `GOOS=linux GOARCH=amd64 go build ./...` | clean |
| API tests | `pnpm exec vitest run --no-file-parallelism src/lib/zodIssues.test.ts src/jobs/backupWorker.test.ts src/routes/agentWs.test.ts src/routes/agentWs.terminalResultSchema.test.ts` | 4 files, 137 passed |
| API typecheck | `pnpm exec tsc --noEmit -p tsconfig.json` | clean |

New coverage: the retry-policy table (incl. the ordering guarantee), the short-vs-long backoff split, the liveness probe's self-calibration / confirmation / per-volume scoping / inconclusive-error handling, an end-to-end "shadow copy dies at file 3 of 6" regression asserting the run stops at 3 attempts rather than walking a dead snapshot, and `<root>` vs named-path rendering.

## Residual risk — likely true root cause of #3260 is NOT fixed here

`vss.CreateShadowCopy` releases `IVssBackupComponents` before returning (`agent/internal/backup/vss/vss_windows.go`, the `defer callVtable(backupComponents, vtblRelease)`), with a comment asserting Windows only reclaims a non-persistent `VSS_CTX_BACKUP` copy on **process** exit. Microsoft documents the opposite: releasing that object deletes auto-release shadow copies. The existing live test only reads through the device shortly after Release, so it would not catch a delayed reclaim — and a delayed reclaim fits the observed ~390s disappearance at least as well as diff-area exhaustion.

**This PR does not attempt that fix.** Keeping the requester alive on a dedicated locked COM thread across scan + upload + manifest, with `BackupComplete`/`AbortBackup`, is a lifetime rewrite of the vss package — Windows-only, untestable from here, and already tracked as a follow-up on #2999. #3260 names the explicit abort as the required minimum and snapshot refresh as stretch, so this PR delivers the abort and the liveness probe (which is still needed regardless — diff-area exhaustion and external deletion are real). **Worth its own issue.**

Also worth noting: the outer queue schema (`backupProcessResultSchema`) guarantees `data.result` is a non-null object before `processResults` runs, so the reported root-level "received null" most likely arose from a *nested* null. Either way the new message now names whatever it was.

---

## Review round (3 agents: code-reviewer, silent-failure-hunter, pr-test-analyzer)

**5 findings raised, all addressed.** code-reviewer found no correctness bugs. The other two found real gaps:

| # | From | Finding | Fix |
|---|---|---|---|
| 1 | silent-failure | VSS active but zero watchable roots returned `nil` with only Debug lines — the #3260 guard could be entirely OFF and look identical to "armed, nothing wrong" | Warn once, naming how many roots were offered |
| 2 | silent-failure | A root stat'ing with a persistent non-not-exist error was treated as alive and logged at Debug forever | `shadowRootMissing` is now three-way (alive / gone / inconclusive); caller warns once per root per run |
| 3 | self-found | `matchShadowRoot` used a bare `strings.HasPrefix` — `...ShadowCopy2` prefixes `...ShadowCopy26`, so a file under an **unwatched** ShadowCopy26 would be checked against ShadowCopy2's root and could abort a healthy run | Match on a path boundary (separator or exact); new `TestMatchShadowRoot_MatchesOnPathBoundaryNotBarePrefix` |
| 4 | self-found | An inconclusive *confirmation* look was logged as "came back" — a message that misdescribes what happened, which is #3260's own complaint | Reported on its own terms |
| 5 | pr-test-analyzer | The **post-retry** liveness probe was not pinned by any test: disabling only it left the package green, because the existing regression test kills the snapshot synchronously with the first attempt | New `TestCreateSnapshot_ShadowCopyLostDuringRetry_AbortsExplicitly`, proven non-vacuous |

None of these changed the abort decision — only a confirmed not-exist still aborts. They changed what an operator can see, and closed one genuine false-abort vector (#3 hardening).

**A comment of mine was factually wrong and is corrected.** I had written that Go maps both `ERROR_ACCESS_DENIED` and `ERROR_SHARING_VIOLATION` to `fs.ErrPermission`, making table-before-portable ordering the thing that keeps a sharing violation on `skipWithoutRetry`. Checked the stdlib (`syscall/syscall_windows.go`, `Errno.Is`): only `ERROR_ACCESS_DENIED`/`EACCES`/`EPERM` map to `fs.ErrPermission`. `ERROR_SHARING_VIOLATION` does not — it has no portable fallback at all. Ordering still matters (for the precise symbolic reason string support reads), and the sharing/lock/cloud rows guard against the table being dropped entirely. Comments now say what is true.

**CI caught the rename, as designed.** The `Test Agent (Windows)` step pins test names in its `-run` filter and asserts an exact PASS count precisely so a rename can't quietly drop coverage. The classifier rename tripped it; three names updated, count unchanged at 14 (the rename was one-for-one).

### Known remaining gap (not fixed here)

The manager-level wiring in `RunBackupContext` — `sourceLiveness = newShadowRootLiveness(vssSession.ShadowPaths)` and its threading into `createSnapshotWithProgress` — has no test. Every liveness test calls the leaf functions directly. A refactor that dropped those two lines would pass CI and only resurface as a #3260 recurrence in production. Covering it needs the `vss.Provider` to become injectable into `RunBackupContext` (today it's constructed inline via `vss.NewProvider`, and the whole branch is `runtime.GOOS == "windows"`-gated so it never executes in CI's Linux/macOS runs). That's a testability refactor of the manager, deliberately out of scope here — **worth its own issue alongside the requester-lifetime one.**

---

## Review round 2 (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer)

Nothing blocking, but two findings were latent defeats of the entire #3260 guard. All applied in `b7511589d`.

**1. The guard was probably inert on real Windows (the important one).** `vss.CreateShadowCopy` returns `SnapshotDeviceObject` verbatim — a **bare** `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyNN`, no trailing separator — and calibration stat'd exactly that. Nothing in the tree has ever shown that form to be stat-able: the live VSS test does `ReadDir(shadow+"\Windows")`, `collectFiles` stats `shadow+rest`, and every liveness test uses a fake stat. If the bare form doesn't resolve, calibration excluded every root on every real run and the guard silently disarmed — fail-safe, but useless.

`resolveStatablePath` now tries the root verbatim, then with a trailing separator, and watches whichever answers. That required splitting a root into **`prefix`** (the exact text `rewritePathsForVSS` prepended — what files are matched against) and **`statPath`** (the form that actually stats); conflating them would have broken the path-boundary match instead. The live Windows test now records which form resolves and **fails if neither does**, so a real run verifies the arming precondition rather than assuming it.

**2. Journal-less abort destroyed data that pre-PR runs kept.** `abortSourceGone` deleted the uploaded objects whenever `journal == nil` (which happens when `resolveJournalDir` fails). Pre-PR, that same run finished as a flagged partial success and left a usable restore point — so the PR was a regression on that path.

I took the **preferred** fix rather than the minimum bar, because the `return snapshot, err` pattern already existed in this function and the manifest writer was trivially extractable:

| journal | behavior |
|---|---|
| present | unchanged — prefix + journal are resume state; nothing deleted, no manifest |
| absent, ≥1 file landed | publish a **partial manifest** so the objects stay reachable; return `(snapshot, err)` so the job still reports **failed** with the source-loss reason while recording what landed |
| absent, nothing landed | clean up (no restore point to preserve) |
| manifest publish itself fails | objects **retained** and logged loudly — deletion is irreversible and this is a data-protection product |

I also withdrew my own "a restore point that lies" justification: a manifest enumerates what a snapshot contains, it does not assert completeness, and the job is failed either way. `publishSnapshotManifest` is extracted so the abort and a normal completion can't drift apart.

**3. Partial-guard silence.** An excluded root was Debug-only and the Warn fired only when *zero* roots were watchable — so a multi-volume run with the guard off for one volume reproduced #3260's misattribution invisibly. Excluded roots now Warn whenever others are still watched, and the arming line is **Info** with watched/excluded counts.

**4.** `routes/backup/{hyperv,mssql}.ts` still dropped the Zod path over the same schema — both switched to `describeZodIssues`.

**Tests added:** journal-vs-journal-less abort disposition (the data-integrity guard for #2, written against the final behavior); a `!windows` Win32/POSIX errno-collision tripwire (5=EIO, 32=EPIPE, 33=EDOM must keep the full retry — `upload_error_class_other.go` exists only to prevent that leak and is a no-op body someone could "unify" away); the miss→inconclusive confirmation branch; the trailing-separator fallback and prefix-vs-statPath matching.

Writing these surfaced a flaw in my two existing abort tests: their fake providers were killing the **manifest** upload too, which a dead *source* snapshot cannot do. Now scoped to source reads only.

**Comment corrections (comment-analyzer).** The forensic narrative said 15 files each burned 30s — the record says ~13 did, and 13×30s *is* the +390s the same sentence quotes (the snapshot died before the last denials were reached); "files 14-40, every one readable" also conflicted with 15-of-40 denied. The cross-reference to `vss.CreateShadowCopy` pointed at a note asserting the *opposite* of the Microsoft-documented behavior (now stated inline, contrary note tracked as **#3269**). "Never aborts the job" in `upload_error_class.go` was false one screen away. The `backup.go` ordering claim had the sequence backwards.

**Verification:** `go test -race ./...` — **67 packages, 0 failures**. `go vet ./...` clean; `GOOS=windows` and `GOOS=linux` builds clean; `GOOS=windows go vet ./internal/backup/...` clean. API: **5 files, 147 passed**; `tsc --noEmit` clean. The `Test Agent (Windows)` PASS-count guard still expects 14 — the new errno test is `!windows`-tagged, so it doesn't count there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1253,7 +1253,7 @@ jobs:
           CGO_ENABLED: "0"
         shell: pwsh
         run: |
-          $out = go test ./internal/backup -run 'TestPermanentWindowsErrnoConstantsMatchWin32|TestClassifyPermanentUploadError|TestLookupPermanentWindowsErrno|TestRewritePathsForVSS|TestReportableLiveReads|TestSummarizeLiveReads' -v 2>&1
+          $out = go test ./internal/backup -run 'TestWindowsUploadErrnoConstantsMatchWin32|TestClassifyUploadFailure|TestLookupWindowsUploadErrno|TestRewritePathsForVSS|TestReportableLiveReads|TestSummarizeLiveReads' -v 2>&1
           $exit = $LASTEXITCODE
           $out | ForEach-Object { Write-Host $_ }
           if ($exit -ne 0) { throw "go test failed (exit $exit)" }
--- a/agent/internal/backup/backup.go
+++ b/agent/internal/backup/backup.go
@@ -427,8 +427,18 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 		}
 	}
 
+	// sourceLiveness watches the shadow-copy roots this run reads from so the
+	// upload loop can tell "the snapshot died" apart from "these files are
+	// bad" (#3260). Built here, immediately after the session is created and
+	// the paths are rewritten onto it, so its self-calibrating stat sees the
+	// roots at their freshest — the file walk below can itself take minutes.
+	// Stays nil for a non-VSS run: reading the live filesystem has nothing to
+	// defend.
+	var sourceLiveness sourceLivenessFn
+
 	// Rewrite paths to shadow copy device paths when VSS is active
 	if vssSession != nil {
+		sourceLiveness = newShadowRootLiveness(vssSession.ShadowPaths)
 		var unmappedIdx []int
 		backupPaths, unmappedIdx = rewritePathsForVSS(backupPaths, vssSession.ShadowPaths, systemStateStagingIdx)
 		if liveReads := reportableLiveReads(backupPaths, unmappedIdx, systemStateStagingIdx); len(liveReads) > 0 {
@@ -629,7 +639,7 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 		}
 	}
 
-	snapshot, snapErr := createSnapshotWithProgress(runCtx, m.config.Provider, files, progressFn, journal, prevSnapshot)
+	snapshot, snapErr := createSnapshotWithProgress(runCtx, m.config.Provider, files, progressFn, journal, prevSnapshot, sourceLiveness)
 	if errors.Is(snapErr, errBackupStopped) {
 		return stopBackupRun()
 	}
--- a/agent/internal/backup/backup.go
+++ b/agent/internal/backup/backup.go
@@ -429,9 +429,16 @@ func (m *BackupManager) RunBackupContext(ctx context.Context, excludes []string)
 
 	// sourceLiveness watches the shadow-copy roots this run reads from so the
 	// upload loop can tell "the snapshot died" apart from "these files are
-	// bad" (#3260). Built here, immediately after the session is created and
-	// the paths are rewritten onto it, so its self-calibrating stat sees the
-	// roots at their freshest — the file walk below can itself take minutes.
+	// bad" (#3260).
+	//
+	// Built at the very top of the VSS block — before the path rewrite, and well
+	// before the file walk, which on a large volume runs for minutes. That
+	// ordering is load-bearing: newShadowRootLiveness calibrates by stat-ing
+	// each root once and watching only the ones that answer, so it has to run
+	// while the snapshot is as fresh as it will ever be. Calibrating after the
+	// walk would risk arming the guard against a snapshot that had already
+	// started to go.
+	//
 	// Stays nil for a non-VSS run: reading the live filesystem has nothing to
 	// defend.
 	var sourceLiveness sourceLivenessFn
--- a/agent/internal/backup/snapshot.go
+++ b/agent/internal/backup/snapshot.go
@@ -475,22 +475,30 @@ func createSnapshotWithProgress(ctx context.Context, provider providers.BackupPr
 	}
 
 	// abortSourceGone is the single exit point for the abort taken when the
-	// source snapshot goes away mid-run (#3260). It deliberately mirrors
-	// abortStopped: the run did not
-	// finish, so no manifest is written, and the partial remote prefix is
-	// cleaned up ONLY when there is no journal — with a journal, the prefix
-	// plus the journal are this run's resume state and the objects already
-	// uploaded stay usable for the next attempt.
+	// source snapshot goes away mid-run (#3260). The run stops either way; what
+	// differs is what happens to the objects already uploaded, and that turns
+	// entirely on whether this run has a checkpoint journal.
 	//
-	// Not written as a partial success on purpose. The already-uploaded objects
-	// are real data, but publishing a manifest for them would mint a snapshot
-	// that LOOKS complete while silently missing every file the loop never
-	// reached — a restore point that lies. Failing loudly and keeping the
-	// journal preserves the same bytes without the lie.
+	//   journal != nil — the prefix plus the journal ARE the resume state, and
+	//     the next attempt resumes into this very snapshot ID. No manifest:
+	//     publishing one now would stake a completed-snapshot claim on an ID
+	//     the next run is still filling in. Nothing is deleted.
+	//
+	//   journal == nil — there is no resume state, so leaving the prefix
+	//     unpublished would strand the uploaded objects as unreachable orphans.
+	//     Publish a PARTIAL manifest instead, making them a real restore point.
+	//     This is not a "snapshot that lies": a manifest enumerates what the
+	//     snapshot contains, it does not assert completeness, and the run is
+	//     still reported as FAILED with the source-loss reason and the counts.
+	//     Deleting them would be a regression against the pre-#3260 behaviour,
+	//     where the same run finished as a flagged partial success and left a
+	//     usable restore point behind.
 	//
 	// The error carries the counts because #3260's whole complaint is that the
 	// operator-visible failure said nothing useful about what happened.
 	abortSourceGone := func(cause error) (*Snapshot, error) {
+		detail := fmt.Errorf("%w (aborted after %d of %d files uploaded, %d failed)",
+			cause, len(snapshot.Files), filesTotal, len(errs))
 		log.Error("backup source snapshot is gone mid-run, aborting the run",
 			"snapshotId", snapshot.ID,
 			"filesUploaded", len(snapshot.Files),
@@ -499,11 +507,36 @@ func createSnapshotWithProgress(ctx context.Context, provider providers.BackupPr
 			"resumable", journal != nil,
 			"error", cause.Error(),
 		)
-		if journal == nil {
-			cleanupSnapshotPrefix(provider, snapshot.ID)
+		if journal != nil {
+			return nil, detail
 		}
-		return nil, fmt.Errorf("%w (aborted after %d of %d files uploaded, %d failed)",
-			cause, len(snapshot.Files), filesTotal, len(errs))
+		if len(snapshot.Files) == 0 {
+			// Nothing landed, so there is no restore point to preserve and the
+			// prefix holds no recoverable data. Same disposal as any other
+			// journal-less abort.
+			cleanupSnapshotPrefix(provider, snapshot.ID)
+			return nil, detail
+		}
+		snapshot.UploadFailures = errs
+		if pubErr := publishSnapshotManifest(ctx, provider, snapshot, prefix); pubErr != nil {
+			// Deliberately NOT followed by cleanupSnapshotPrefix. Deletion is
+			// irreversible and this is a data-protection product: retained
+			// orphans cost storage, deleted backups cost the customer their
+			// files. Log loudly enough that an operator can find them.
+			log.Error("could not publish a partial manifest for the aborted run; the uploaded objects are RETAINED but unreachable without one",
+				"snapshotId", snapshot.ID,
+				"prefix", prefix,
+				"filesUploaded", len(snapshot.Files),
+				"error", pubErr.Error(),
+			)
+			return snapshot, errors.Join(detail, pubErr)
+		}
+		log.Warn("published a PARTIAL manifest for the aborted run; it restores the files that landed before the source snapshot went away",
+			"snapshotId", snapshot.ID,
+			"filesUploaded", len(snapshot.Files),
+			"filesTotal", filesTotal,
+		)
+		return snapshot, detail
 	}
 
 	for _, file := range files {
@@ -721,9 +754,38 @@ func createSnapshotWithProgress(ctx context.Context, provider providers.BackupPr
 		return abortStopped()
 	}
 
+	if err := publishSnapshotManifest(ctx, provider, snapshot, prefix); err != nil {
+		if errors.Is(err, errBackupStopped) {
+			// A manifest-upload deadline expiry is fatal for the snapshot too
+			// (unlike a per-file data upload): without the manifest the
+			// snapshot isn't restorable, so there's nothing to keep going for.
+			return abortStopped()
+		}
+		return snapshot, err
+	}
+
+	if journal != nil {
+		if err := journal.Complete(); err != nil {
+			log.Warn("failed to remove completed checkpoint journal", "error", err.Error())
+		}
+		completed = true
+	}
+
+	return snapshot, nil
+}
+
+// publishSnapshotManifest serializes snapshot's manifest and uploads it under
+// prefix, making the objects already stored there a reachable restore point.
+//
+// Extracted so the mid-run source-loss abort can publish the partial set it
+// managed to store (see abortSourceGone) using exactly the same code path as a
+// normal completion — a second, subtly different manifest writer is how the two
+// would drift apart. errBackupStopped is returned unwrapped so callers can tell
+// a job cancel from a genuine manifest failure.
+func publishSnapshotManifest(ctx context.Context, provider providers.BackupProvider, snapshot *Snapshot, prefix string) error {
 	manifestPath, manifestErr := writeSnapshotManifest(snapshot)
 	if manifestErr != nil {
-		return snapshot, manifestErr
+		return manifestErr
 	}
 	defer os.Remove(manifestPath)
 
@@ -738,22 +800,11 @@ func createSnapshotWithProgress(ctx context.Context, provider providers.BackupPr
 	cancelAttempt()
 	if manifestUploadErr != nil {
 		if errors.Is(manifestUploadErr, errBackupStopped) {
-			// A manifest-upload deadline expiry is fatal for the snapshot too
-			// (unlike a per-file data upload): without the manifest the
-			// snapshot isn't restorable, so there's nothing to keep going for.
-			return abortStopped()
+			return manifestUploadErr
 		}
-		return snapshot, fmt.Errorf("failed to upload snapshot manifest: %w", manifestUploadErr)
+		return fmt.Errorf("failed to upload snapshot manifest: %w", manifestUploadErr)
 	}
-
-	if journal != nil {
-		if err := journal.Complete(); err != nil {
-			log.Warn("failed to remove completed checkpoint journal", "error", err.Error())
-		}
-		completed = true
-	}
-
-	return snapshot, nil
+	return nil
 }
 
 // attemptFileUpload runs a single upload attempt for file against a fresh
--- a/agent/internal/backup/snapshot.go
+++ b/agent/internal/backup/snapshot.go
@@ -230,6 +230,35 @@ func setUploadRetryDelayForTest(d time.Duration) (restore func()) {
 	return func() { uploadRetryDelay = old }
 }
 
+// shortUploadRetryDelay is the backoff for a source-permission denial
+// (retryAfterShortDelay — see classifyUploadFailure). An NTFS ACL never clears,
+// so the wait exists purely to ride out an AV/indexer/filter-driver hold.
+//
+// One second is an operational tradeoff, not a guarantee: Windows promises
+// nothing about how quickly such a hold clears, so this does narrow the
+// recovery window compared with the 30s backoff. It is still the right call —
+// the retry itself (the thing that actually recovers a transient hold) is
+// preserved, and the 27-29s per denied file it removes is what let a run
+// outlive its own shadow copy and lose EVERY file (#3259 -> #3260).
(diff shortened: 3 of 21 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (all three, on the initial commit), plus a Codex `xhigh` read-only design consult before implementation.

**Findings:** 5 raised → all 5 addressed across `9c859521a`, `dc01661ec`, `dfb890e58`; 0 outstanding.
- code-reviewer: **0 findings** (checked the source-attribution guard, branch ordering, false-abort vectors, abort/cancel precedence, off-Windows errno safety).
- silent-failure-hunter: 2 — the guard could be entirely OFF with only Debug-level evidence, and a persistently inconclusive root stat was swallowed forever. Both now warn; neither changed the abort decision.
- pr-test-analyzer: 1 — the **post-retry** liveness probe was pinned by nothing (disabling only it left the package green). New `TestCreateSnapshot_ShadowCopyLostDuringRetry_AbortsExplicitly`, verified non-vacuous. It also confirmed the pre-existing regression test is non-vacuous by the same method.
- Self-found while reviewing: `matchShadowRoot` matched a bare string prefix (`...ShadowCopy2` prefixes `...ShadowCopy26`) — a genuine false-abort vector, now boundary-matched; and an inconclusive confirmation look was mislabelled "came back".

Also corrected a **factually wrong comment I had written**: only `ERROR_ACCESS_DENIED`/`EACCES`/`EPERM` map to `fs.ErrPermission` in `syscall/syscall_windows.go` — `ERROR_SHARING_VIOLATION` does not, so the ordering guarantee holds for a different (still real) reason than I claimed.

**Tests:** `go test -race ./...` in `agent/` — exit 0, 66 packages, 0 failures. Cross-compile clean for `GOOS=windows` and `GOOS=linux`. API: 137 passed across the 4 affected files single-fork; `tsc --noEmit` clean.

**CI: 54 pass / 0 fail / 1 skipping (55 total).** `Test Agent (Windows)` failed on the first push — its `-run` filter pins test names and asserts an exact PASS count specifically so a rename can't silently drop coverage. The classifier rename tripped it; names updated, count unchanged at 14. It passes now.

Two things flagged for the maintainer rather than fixed here, both argued in the PR body:
1. **The likely true root cause of #3260 is untouched** — `vss.CreateShadowCopy` releases `IVssBackupComponents` before returning, and Microsoft documents that as deleting auto-release shadow copies. Fixing it is a COM-lifetime rewrite of the vss package (already a tracked follow-up on #2999). The liveness probe is needed regardless.
2. **The manager-level wiring is untested** — covering it needs `vss.Provider` to become injectable into `RunBackupContext`.

ToddHebebrand:
**Review run:** round 2 — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer.

**Findings:** 4 code + 5 comment → all addressed in `b7511589d`; 0 outstanding. Nothing blocking, but two were latent defeats of the entire #3260 guard:

1. **The guard was probably inert on real Windows.** Calibration stat'd the **bare** device path `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyNN` that `CreateShadowCopy` returns verbatim — and nothing in the tree has ever shown that form to be stat-able (the live VSS test and the file walk both go through a subdirectory; every liveness test uses a fake stat). If it doesn't resolve, every root was excluded on every real run and the guard silently disarmed. Added a trailing-separator fallback, split `prefix` (what files match against) from `statPath` (what actually stats), and made the **live Windows test fail if neither form resolves** — so the arming precondition is now verified on real hardware instead of assumed.
2. **Journal-less abort deleted data that pre-PR runs kept.** Took the preferred fix, not the minimum bar: with no journal and ≥1 file landed, publish a **partial manifest** and return `(snapshot, err)` — the job still reports failed with the source-loss reason, but the uploaded objects stay a reachable restore point. Only the nothing-landed case cleans up; if the manifest itself can't be published the objects are retained and logged loudly. I also withdrew my own "restore point that lies" justification — a manifest enumerates contents, it doesn't assert completeness.
3. Partial-guard silence: an excluded root was Debug-only and the Warn fired only when *zero* roots were watchable, so a multi-volume run half-disarmed looked fully armed. Excluded roots now Warn; the arming line is Info with watched/excluded counts.
4. `routes/backup/{hyperv,mssql}.ts` adopted `describeZodIssues`.

Writing the new tests surfaced a flaw in my own two existing abort tests — their fakes were killing the **manifest** upload, which a dead *source* snapshot cannot do. Now scoped to source reads only.

**Comment corrections:** the forensic narrative miscounted the field repro (~13 files burned the backoff, not 15 — 13×30s *is* the +390s it quotes), the `vss.CreateShadowCopy` cross-reference pointed at a note asserting the opposite of documented behavior (now inline; contrary note tracked as #3269), "never aborts the job" was false one screen away, and the `backup.go` ordering claim was backwards.

**Tests:** `go test -race ./...` — **67 packages, 0 failures**. `go vet ./...`, `GOOS=windows`/`GOOS=linux` builds, and `GOOS=windows go vet ./internal/backup/...` all clean. API: **5 files, 147 passed**; `tsc --noEmit` clean.

**CI: 54 pass / 0 fail / 1 skipping (55 total).** `Test Agent (Windows)` green — its PASS-count guard still expects 14, and the new errno tripwire is `!windows`-tagged so it doesn't count there.

**Status:** review-clean, CI green, awaiting maintainer merge. The two out-of-scope items from the PR body still stand: the `IVssBackupComponents` requester-lifetime fix (#3269) and making `vss.Provider` injectable so the manager-level wiring can be tested.

Issue #3242, opened 2026-08-08
[API][Web] No bulk import for organizations or sites — every RMM migration starts with hundreds of manual forms


ToddHebebrand:
## Problem

There is no bulk import for organizations or sites. An MSP arriving from Datto RMM, NinjaOne, ConnectWise Automate, Kaseya, N-central, Atera, Syncro or Pulseway has to create every customer and every location one form at a time — 60–200 records for a typical MSP — before they can enroll a single device.

This is the first thing a prospect does after signing up, and today it is a wall of manual data entry. It is the single largest blocker to migrating onto Breeze.

Confirmed by inspection:

- No CSV parser exists anywhere in the repo — no `papaparse`, `csv-parse`, or equivalent in any `package.json`. Every `parseCsv*` hit is a comma-separated **env var** splitter (`apps/api/src/routes/metrics.ts:75`, `routes/mcpServer.ts:62`, `routes/logs.ts:105`).
- CSV is **export-only**: `apps/api/src/services/spreadsheetExport.ts`, `apps/web/src/lib/csvExport.ts`, `apps/api/src/routes/tickets/export.ts`, `apps/api/src/routes/admin/tenantExport.ts`. Tenant export has no import counterpart.
- The `/bulk` endpoints that exist (`lib/bulkOps.ts`, `routes/devices/commands.ts`, `routes/devices/groups.ts`) are all **actions on existing rows**, never creation.

The workaround shipped in `apps/docs/src/content/docs/migration/toolkit.mdx` (Recipe 1) is a bash script that loops `POST /orgs/organizations` and `POST /orgs/sites`. It works, but requiring every new customer to write shell scripts against our API before they can use the product is not an onboarding story.

## Proposal

Add organization + site import, modelled directly on the **QuickBooks customer importer** — the pattern already exists and already solves the hard parts:

- `apps/api/src/services/accounting/quickbooksCustomerImport.ts` — has `slugify`, `generateUniqueSlug`, `loadExistingOrgs` keyed by `byExternalId` (dedupe against an external system's IDs), `listQuickbooksCustomersAnnotated` (preview/annotate), `importQuickbooksCustomers` (commit).
- Route: `POST /accounting/:provider/customers/import` — `apps/api/src/routes/accounting/index.ts:279`
- UI: `apps/web/src/components/integrations/QuickbooksCustomerImport.tsx`
- Design doc: `docs/superpowers/specs/billing/2026-06-29-quickbooks-customer-import-design.md`

[...]

Issue #3246, opened 2026-08-08
[API] PSA getCompanies() is implemented by every adapter and called by nothing — wire it to organization import


ToddHebebrand:
## Problem

Every PSA adapter implements `getCompanies()`, and **nothing calls it**.

- The method is on the provider interface at `apps/api/src/services/psa/types.ts:62`, and is implemented by ConnectWise, Autotask, Freshservice, ServiceNow and Zendesk (`apps/api/src/services/psa/*.ts`).
- Grep for call sites returns no route and no job. It is dead code.

This is the closest thing in the codebase to a working organization-import source. An MSP whose customer list already lives in ConnectWise Manage or Autotask could bootstrap their entire Breeze tenancy tree from the PSA they already trust as the system of record — no CSV, no scripting, and correct by construction.

## Two adjacent gaps in the same area

1. **The sync endpoint is a stub.** `POST /psa/connections/:id/sync` (`apps/api/src/routes/psa.ts:598`) writes `lastSyncAt` and `lastSyncStatus: 'queued'` and returns. There is no worker consuming that queue, so the button reports success and nothing happens.
2. **Three declared providers have no adapter.** `apps/api/src/db/schema/integrations.ts:11-22` declares `psa_provider` as including `halo`, `syncro` and `kaseya`, and the UI already offers HaloPSA (`apps/web/src/components/integrations/PSAConnectionWizard.tsx:38`, `PSAConnectionList.tsx:31`) — but no adapter file exists for any of the three. A user who selects HaloPSA today hits a dead end.

`README.md:385` lists "External PSA sync (ConnectWise, Autotask, HaloPSA) for teams mid-migration" as an unchecked roadmap item, so this is known — this issue is to make the specific, small step concrete.

## Proposal

1. Wire `getCompanies()` into an org-import flow reusing the QuickBooks importer pattern (`apps/api/src/services/accounting/quickbooksCustomerImport.ts`): list → annotate (create / already-linked / conflict) → commit, deduped by the PSA's company ID held as an external ID.
2. UI on the PSA connection page: "Import companies as organizations", preview table, commit — mirroring `apps/web/src/components/integrations/QuickbooksCustomerImport.tsx`.
3. Either implement the sync worker behind `POST /psa/connections/:id/sync`, or make the endpoint report honestly that sync is not yet available. Reporting `queued` for work that nothing will ever consume is worse than a 501.
4. Either add the Halo/Syncro/Kaseya adapters or remove them from the wizard until they exist.

[...]

PR #3273, opened 2026-08-08
feat(orgs): bulk org/site import + organization_external_links (#3242)

### LINKED ISSUE #3243: [API][Security] No machine-to-machine credential can provision tenancy — org/site writes are interactive-JWT + MFA only
## Problem

There is no credential that can provision tenancy machine-to-machine. Any script that creates organizations, sites, or enrollment keys — exactly what an RMM migration needs — must carry a **short-lived, MFA-satisfied, interactively-obtained user JWT**.

Three findings combine to produce this:

1. **`X-API-Key` is not general-purpose auth.** `apiKeyAuthMiddleware` (`apps/api/src/middleware/apiKeyAuth.ts:86`) is mounted on exactly three surfaces — `routes/mcpServer.ts:219`, `routes/devPush.ts:79`, `routes/devices/customFieldValues.ts:87`. Everywhere else `authMiddleware` (`apps/api/src/middleware/auth.ts:425`) hard-requires `Authorization: Bearer` and throws 401 on anything else.
2. **The Partner API is read-only.** Every route under `apps/api/src/routes/partnerApi/` is a `.get()` with a `:read` scope. There is no write or ingest side.
3. **The tenancy write routes are MFA-gated.** `POST /orgs/organizations` (`routes/orgs.ts:1267`), `POST /orgs/sites` (`routes/orgs.ts:1832`), and `POST /orgs/partners` (`routes/orgs.ts:386`) all carry `requireMfa()`, which requires `auth.token.mfa === true` (`middleware/auth.ts:786`). A token that has not completed an interactive MFA challenge gets `403 MFA required`.

The net effect: a partner onboarding 80 customers must log in as a human, complete MFA, scrape the access token out of the session, and race the token's expiry — refreshing mid-run — to script their own onboarding. There is no unattended path at all, so this also blocks any scheduled sync, IaC/Terraform provider, or partner-built integration that needs to create tenancy.

## Why this is not just "add an API key scope"

`requireMfa()` on these routes is deliberate and correct for **human** callers — creating an organization is a tenancy-shaping action. The gap is that Breeze has no notion of a **non-human principal that is strong-authenticated by construction**, for which "has this user typed a TOTP code" is the wrong question.

The building blocks already exist: `apps/api/src/routes/servicePrincipals.ts` and `routes/partnerServicePrincipals.ts`, plus `principalType` / `principalId` already threaded through the API-key auth context (`middleware/apiKeyAuth.ts:283`).

## Proposal

Let a **partner-scoped service principal** obtain a token that satisfies the tenancy write routes:

1. Define the rule explicitly: a service-principal credential is treated as MFA-satisfied because it is a non-interactive, rotatable, single-factor-by-design principal — not by setting `mfa: true` as a side effect, but via an explicit branch in `hasSatisfiedMfa` that tests principal type. Vacuous MFA grants have bitten us before (see the comment at `middleware/cfAccessLogin.ts:204`); this must be a deliberate, tested branch.
2. Add write scopes — `organizations:write`, `sites:write`, `enrollment-keys:write` — clamped by the existing scope-delegation ceiling.
3. Every write must land in the audit log attributed to the service principal, not to the human who minted it.
4. Rate-limit and cap: a provisioning credential that can create unlimited organizations is a billing and abuse surface.

## Decision needed before implementing

This touches auth and tenancy — per `CLAUDE.md` it warrants an advisor quorum, not a straight-to-code fix. Specifically: whether to extend service principals as above, or introduce a distinct short-lived "provisioning token" minted by an MFA'd human and valid for a bounded window and operation set. The second is narrower and weaker; the first is more useful and a larger blast radius.

## Workaround today

`apps/docs/src/content/docs/migration/toolkit.mdx` documents the JWT-and-refresh dance under "Authentication for Migration Scripts". It works, but it is the kind of thing that makes an evaluator close the tab.

ToddHebebrand:
Closes #3243. Part of epic #3249.

Implements the plan in `docs/superpowers/plans/open/2026-08-08-partner-api-provisioning-writes.md` (on the `docs/rmm-migration-guides` branch): lets a partner service principal create organizations, sites, and enrollment keys unattended, so RMM-migration scripts and integrations can provision tenancy without an MFA'd human JWT.

## What changed

**Scopes** (`services/partnerServicePrincipalScopes.ts`)
- New `organizations:write`, `sites:write`, `enrollment-keys:write`, validated by the existing `validatePartnerServicePrincipalScopes` and mintable only through the MFA-gated partner principal management routes.
- The Weavestream default scope set is now pinned to the read scopes explicitly — write scopes are **opt-in at principal creation, never implicit**.
- `hasSatisfiedMfa` is untouched, per the issue thread's explicit rejection of principal-type branching.

**Write-surface allowlist test** (`routes/partnerApi/writeSurface.test.ts`) — written FIRST, per the condition attached to the #3243 decision
- Enumerates every route under `partnerApiRoutes`; the set of non-GET routes must equal an explicit allowlist constant. Seeded empty, verified green on the pre-change tree; its first run against the new routes failed with exactly `POST /organizations`, `POST /sites`, `POST /enrollment-keys` before they were allowlisted (evidence the guard works). A canary test proves a fourth unlisted write route fails the suite.

**Auth middleware** (`middleware/partnerApiAuth.ts`)
- Non-GET requests take a new branch: principal + org discovery resolve in a short bounded system context, and the handler runs with **no ambient DB context** — handlers open their own bounded contexts per operation (mirroring the human `/orgs/*` write routes).
- Why: the read path holds the partner-export advisory lock **shared** for the whole request, while an `organizations` INSERT trigger takes the same lock **exclusive** (`2026-07-21-partner-export-canonical-org-mutations.sql`). An insert from a nested system transaction on a second pooled connection would wait on our own request's shared lock — an application-level self-deadlock. Writes also take no export snapshot, so the held-transaction consistency machinery doesn't apply to them.
- Writes get a tighter per-principal rate bucket: `min(key limit, 120)/hour`, separate Redis key, 429 + `Retry-After` on exhaustion. `recordMachineUse` continues to audit every write request (success and failure) as the service principal.

**Routes** (`routes/partnerApi/provisioning.ts`)
- `POST /partner-api/organizations` — `organizations:write`. `partnerId` comes only from the principal (body value stripped). Runs in a bounded system context (a new org's id can't pass the id-keyed RLS insert policy under partner scope — same escape as the human route), takes the partner lock exclusive up front so the `maxOrganizations` quota check is race-free, maps slug 23505 to 409, and enforces the cap with a specific 409 (`partner_provisioning_org_limit_reached`). Status restricted to `active`/`trial` (creating suspended/churned tenants unattended is not a workload).
- `POST /partner-api/sites` — `sites:write`. `orgId` outside the principal's accessible set → 403. Insert runs in a partner-scoped bounded context (`userId: null`), so RLS enforces org access a second time.
- `POST /partner-api/enrollment-keys` — `enrollment-keys:write`. Mirrors the human schema (`.strict()`, `ttlMinutes` XOR `expiresAt`, `maxUsage` 1–100000), rejects TTLs above the partner cap on both expiry paths via the shared `assertTtlWithinCap`, verifies `siteId` belongs to the org, stores only the hash, `createdBy: null`. Raw key returned exactly once at the top level of the 201 body.
- Created-object responses reuse the read-side export DTO contracts: same strict record schemas, `revision`, and `safelyExportDefinition` inspection, so `dtoSafety`/`exportSafety` apply to writes exactly as to reads.
- **No DELETE routes** — deletion stays human + MFA on the main API (plan Task 4).
- Audit: `organization.create` / `site.create` / `enrollment_key.create` events attributed to the service principal (`actorType: 'api_key'`, key id), never a human.

**UI** (`apps/web/.../PartnerServicePrincipalsPage.tsx`) — write scopes selectable but excluded from the default selection.

**Docs** — `reference/api.mdx` gains a Partner API section (all routes + scopes); `reference/api-keys.mdx` gains the partner-principal scope table with the opt-in warning.

## Tests

- `writeSurface.test.ts` — allowlist + canary (first run failed pre-allowlist, see above).
- `provisioning.test.ts` — 21 cases: scope/auth (401/403), body-`partnerId` ignored, org-access 403s, `maxOrganizations` boundary (at cap 409 / below cap 201), TTL-cap 400, XOR 400, `.strict()` unknown-key 400, site-mismatch 400, raw-key-once + hash-stored + `createdBy` null, DTO revision shape, partner-context `userId: null`, audit attribution for all three routes.
- `partnerApiAuth.test.ts` — 5 new cases for the write branch: no held context during `next()`, write bucket key/limit, write 429 + Retry-After, machine-use audit on write success and failure. All 34 pre-existing cases unchanged and green.
- Affected suites green locally (single worker): 19 files, 490 tests (`partnerApiAuth`, all `routes/partnerApi/*`, `partnerServicePrincipals`, `enrollmentKeys*`, `enrollmentKeySecurity`). `tsc --noEmit` clean for `apps/api`.

## Not in this PR

- `apps/docs/.../migration/toolkit.mdx` rewrite (plan Task 7): that file lives on the unmerged `docs/rmm-migration-guides` branch (PR #3250), not on `main` — its auth section and Recipe 1/2 should be updated there (or in a follow-up) once both PRs land.
- End-to-end verification against a live stack (plan Verification §2): needs a running deployment; unit + contract coverage above stands in.
- Bulk import on the Partner API (explicitly out of scope, #3242).


ToddHebebrand:
**Review run:** /code-review at high effort (8 finder angles: 3 correctness — line-by-line diff scan, removed-behavior audit, cross-file trace; reuse; simplification; efficiency; altitude; conventions).

**Findings:** 3 consequential raised → all addressed in 0e5bdd06d; 0 outstanding.
1. New local `isUniqueViolation` only read the top-level `err.code`, but Drizzle wraps postgres.js errors in `DrizzleQueryError` (code lives on `.cause`) — the concurrent-duplicate recovery paths would never fire. → Both the org-import pipeline and the QuickBooks importer now use the shared `isPgUniqueViolation` from `utils/pgErrors.ts`.
2. Name-match identity gap: commit re-derived the match by name only, so an acknowledgement of org X could silently transfer to a different org Y that took over the name between preview and commit. → Commit rows now pin `expectedOrganizationId` from preview and are rejected if the re-derived match resolves elsewhere.
3. Web select-all selected every non-conflict row, bulk-acknowledging name-matches and bulk-opting soft-deleted rows into reactivation. → Select-all now only toggles `create`/`link-match` rows; name-match and reactivate stay strictly per-row.

Conventions angle came back clean (migration idempotency, cascade/export registrations, runAction usage, i18n all verified compliant). Remaining reuse/altitude notes (third slug-helper copy in `aiToolsOrgs.ts`, QB importer's own union-read copy pending its migration onto the shared seam) are the plan's explicit Phase 2 items — deferred with #3246 / the QB-seam follow-up, not regressions of this PR.

**Tests:** apps/api affected suites green single-run (`orgImport` 27, `quickbooksCustomerImport` 23, `routes/orgs` 199, `autoMigrate` 60, `routes/accounting` 16, export-policy/tenancy statics 56); apps/web green (`BulkOrgImport` 10, `csvParse`, `no-silent-mutations`, `QuickbooksCustomerImport`, i18n `keyUsage` + `translationCoverage` — 132 across the affected set). `tsc --noEmit` clean for both packages on touched files.

**Live-DB contract suites** (`rls-coverage`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, and the new `orgExternalLinksRls` forge/backfill suite) cannot run on this machine (no Docker) — CI was dispatched on the branch (workflow_dispatch run 31265785264) to cover them; the previous dispatch's only failures were the web i18n contract tests, fixed in 0e5bdd06d. Both the PR run and the dispatched run were still in progress at time of writing — check them before merge.

**Status:** review-clean, awaiting CI + maintainer merge. The `accounting_*` column drop, PSA source (#3246), and QB migration onto the shared seam are deliberate follow-ups per the plan.

ToddHebebrand:
**CI update:** both runs finished green — PR run 31265784182 (required jobs incl. Test API / Test Web) and dispatched run 31265785264 (workflow_dispatch, covering the live-DB contract suites: rls-coverage, tenantCascade, tenant-export-policy, erasure roundtrip, and the new orgExternalLinksRls forge/backfill suite). Nothing outstanding from my side — awaiting maintainer merge.

PR #3275, opened 2026-08-08
fix(dns): authenticate Umbrella with OAuth2 client credentials, not Basic (#3271)

### LINKED ISSUE #3271: [DNS Security] Umbrella integration uses deprecated Basic Auth against the Reporting API — Cisco now requires OAuth2 client-credentials token exchange, causing every sync to fail with 401 regardless of key type
### Description

The Umbrella DNS Security integration fails every sync attempt with a 401 Unauthorized from Cisco's
own API — confirmed NOT a credentials issue: tested against four different Umbrella key types
(Umbrella API key, Legacy Network Devices, Umbrella Reporting, Umbrella Management), all fail
identically, including a direct curl reproduction outside of Breeze entirely using confirmed-active
credentials. Root cause: UmbrellaProvider sends Basic Auth directly to the reports data endpoint,
but Cisco's current Reporting API v2 requires a two-step OAuth2 client-credentials flow instead.

### Steps to Reproduce

1. Create a DNS Security integration with provider: umbrella, any valid apiKey/apiSecret pair
   generated from Umbrella's console, and a correct config.organizationId.
2. Trigger a sync: POST /dns-security/integrations/:id/sync
3. Sync fails. API logs show:
   [DnsSyncJob] sync failed (upstream HTTP error) {"status":401,"statusText":"Unauthorized",
   "responseBody":"{\"meta\":{},\"data\":{\"error\":\"unauthorized\"}}"}
4. Reproduced independently outside Breeze entirely, confirming this is not a Breeze-side bug in
   how credentials are stored/decrypted:
   curl -u "<apiKey>:<apiSecret>" \
     "https://reports.api.umbrella.com/v2/organizations/<orgId>/security-activity?from=...&to=...&limit=10"
   -> also returns 401 directly from Cisco, with the exact same credentials.
5. Tried regenerating credentials from all 4 available Umbrella key categories in Admin > API Keys
   (Umbrella API Keys, KeyAdmin Keys, Static Keys, Legacy Keys > Umbrella Reporting / Umbrella
   Management / Umbrella Network Devices / Legacy Network Devices) — every single one produces the
   identical 401, both through Breeze and via direct curl.

### Expected Behavior

The Umbrella integration should authenticate successfully using valid, active Umbrella credentials
and begin syncing DNS security events.

### Actual Behavior

Every sync attempt fails with 401, regardless of Umbrella key type, confirmed via direct API testing
outside Breeze with the identical credentials. Root cause identified from Cisco's own current
OpenAPI spec for this endpoint (developer.cisco.com/docs/cloud-security/umbrella-api-api-reference-
reports-reporting-api-activity-get-activities-all/):

  "security": [{"oauthFlow": ["reports.granularEvents:read"]}],
  "securitySchemes": {"oauthFlow": {"type": "oauth2", "description": "client credential flow",
    "flows": {"clientCredentials": {
      "tokenUrl": "https://api.umbrella.com/auth/v2/token",
      "scopes": {"reports.granularEvents:read": "Read reports granular events"}
    }}}}

Cisco's current Reporting API v2 requires:
  1. POST https://api.umbrella.com/auth/v2/token with Basic Auth (apiKey:apiSecret) to obtain a
     short-lived Bearer access token
  2. Use that Bearer token (Authorization: Bearer <token>) on the actual data endpoint

Confirmed from apps/api/src/services/dnsProviders/umbrella.ts that UmbrellaProvider does neither
step — it sends Basic Auth directly to the data endpoint:

  basicAuthHeader() {
    const token = Buffer.from(`${this.apiKey}:${this.apiSecret}`).toString("base64");
    return `Basic ${token}`;
  }
  ...
  const url = new URL(`https://reports.api.umbrella.com/v2/organizations/${orgId}/security-activity`);
  ...
  const payload = await requestJson(url, { headers: { Authorization: this.basicAuthHeader() } });

Also worth noting: the base URL/path used (reports.api.umbrella.com/v2/organizations/{orgId}/
security-activity) differs from what Cisco's current spec documents (api.umbrella.com/reports/v2/
activity, org scoping handled via the token's own claims rather than a path segment) — this
integration appears to have been written against an older generation of Cisco's Reporting API that
has since been superseded.

### Component

API Server

### Version

Web/API 0.103.0, self-hosted via Docker Compose

### Relevant Logs

```shell
[DnsSyncJob] sync failed (upstream HTTP error) {"integrationId":"...","status":401,
"statusText":"Unauthorized","responseBody":"{\"meta\":{},\"data\":{\"error\":\"unauthorized\"}}"}
[DnsSyncJob] Job sync-integration-... failed: DnsProviderHttpError: HTTP 401 Unauthorized
    at async UmbrellaProvider.syncEvents (/app/apps/api/dist/index.cjs:223002:27)
```

### Additional Context

Suggested fix: implement the OAuth2 client-credentials token exchange (POST to
https://api.umbrella.com/auth/v2/token with Basic Auth to obtain a Bearer token, cache it for its
stated expiry, then use Authorization: Bearer <token> on subsequent report requests) rather than
sending Basic Auth directly to the data endpoint. May also need to update the base URL/path to
match Cisco's current api.umbrella.com/reports/v2 structure if the legacy reports.api.umbrella.com
host is deprecated alongside the auth change (untested — worth confirming whether that host still
resolves/functions at all once OAuth2 is implemented, or whether both the host and auth need
updating together).

This affects any self-hosted Breeze instance attempting to integrate Cisco Umbrella for DNS
Security — not something fixable via configuration on the operator's end, confirmed by testing
against 4 separate Umbrella key categories, all failing identically both through Breeze and via
direct API testing with the same credentials.

bdunncompany:
Fixes #3271.

Cisco retired direct Basic Auth on the Umbrella APIs. The provider sent `Authorization: Basic <key:secret>` straight at the endpoints, which now 401s for **every** key type — which is exactly why @cisspUser01's four key categories all failed identically and why a direct curl outside Breeze reproduced it. The credentials were never the problem; the auth scheme was.

## Verified against Cisco's docs, not inferred

- token URL `https://api.umbrella.com/auth/v2/token`, `POST`, key/secret as HTTP Basic **on the token request only**, `Content-Type: application/x-www-form-urlencoded`, body `grant_type=client_credentials`
- response `{ token_type, access_token, expires_in }`, lifetime 3600s

## Scope: five call sites, not one

The issue names the reporting endpoint. Cisco's docs state one token covers every Umbrella surface, so I checked the rest of the file — `basicAuthHeader()` was also feeding **four `policies/v2/destinationlists` calls**. A reports-only fix would have left block/allow-list sync broken in precisely the same way, with no second bug report to explain why. All five now go through `withAuth`.

## The retry status is the trap

An **expired** Umbrella token comes back as **`400 invalid_request`**, not 401. A conventional retry-on-401 would never fire for the one failure mode that caching makes possible — the token lapsing mid-run.

`withAuth` refreshes once on a 401, or on a 400 whose body marks it as an auth failure, and deliberately leaves genuine validation 400s alone. That last part matters for the destination-list writes: blanket-retrying a 400 risks turning a rejected write into a duplicate one.

Token is cached on the provider instance, refreshed 60s before expiry, and concurrent callers share one in-flight exchange rather than stampeding the token endpoint.

## Tests

Nine, covering the exchange shape, bearer on both API surfaces, caching, refresh-and-retry on `400 invalid_request` **and** on 401, a no-retry control for a validation 400, the preserved `apiSecret` requirement, and a malformed token response.

All were run against the un-fixed code — by reverting `withAuth` to the old Basic header rather than by pattern-matching call sites — and **8 of the 9 fail**, the ninth being the `apiSecret` check both versions share.

## What I could not verify

I have **no Umbrella tenant**, so this is correct by construction and by unit tests, not by a live sync. The token flow matches Cisco's documented contract, but only a real tenant proves Cisco accepts it end to end.

@cisspUser01 — if you can retest against a build from this branch, that would close the loop properly. Particularly useful: whether the destination-list (block/allow) sync now works too, since that half was broken silently and isn't something the original 401 report would have surfaced.

## Gate

`tsc --noEmit` exit 0 · `vitest run` 1286 files / 20538 tests / 0 failures · `eslint` clean on both changed files.

### DIFF
--- a/apps/api/src/services/dnsProviders/umbrella.test.ts
+++ b/apps/api/src/services/dnsProviders/umbrella.test.ts
@@ -1,0 +1,170 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { UmbrellaProvider } from './umbrella';
+import { DnsProviderHttpError, requestJson } from './http';
+
+// Same shape as the AdGuard/Pi-hole provider tests: transport is not exercised,
+// only the provider's request shaping and response handling. DnsProviderHttpError
+// is kept REAL because the auth-retry path branches on `instanceof`.
+vi.mock('./http', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('./http')>();
+  return {
+    ...actual,
+    requestJson: vi.fn()
+  };
+});
+
+const requestJsonMock = vi.mocked(requestJson);
+
+const TOKEN_URL = 'https://api.umbrella.com/auth/v2/token';
+const ORG_ID = 'org-123';
+
+function urlOf(call: unknown[]): string {
+  return String(call[0]);
+}
+function initOf(call: unknown[]): RequestInit & { headers?: Record<string, string> } {
+  return (call[1] ?? {}) as RequestInit & { headers?: Record<string, string> };
+}
+
+function makeProvider() {
+  return new UmbrellaProvider('key-abc', 'secret-xyz', {
+    organizationId: ORG_ID,
+    blocklistId: 'bl-1',
+    allowlistId: 'al-1'
+  });
+}
+
+/** A token response, then an empty activity page so syncEvents terminates. */
+function queueTokenThen(...bodies: unknown[]) {
+  const queue = [{ access_token: 'tok-1', token_type: 'bearer', expires_in: 3600 }, ...bodies];
+  requestJsonMock.mockImplementation(async () => {
+    if (!queue.length) throw new Error('requestJson mock exhausted');
+    return queue.shift() as never;
+  });
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+describe('UmbrellaProvider OAuth2 client-credentials auth (#3271)', () => {
+  it('exchanges key/secret for a bearer token before calling the API', async () => {
+    queueTokenThen({ requests: [] });
+
+    await makeProvider().syncEvents(new Date('2026-08-01'), new Date('2026-08-02'));
+
+    const tokenCall = requestJsonMock.mock.calls[0]!;
+    expect(urlOf(tokenCall)).toBe(TOKEN_URL);
+
+    const init = initOf(tokenCall);
+    expect(init.method).toBe('POST');
+    expect(init.body).toBe('grant_type=client_credentials');
+    expect(init.headers?.['Content-Type']).toBe('application/x-www-form-urlencoded');
+    // key:secret is Basic ONLY on the token exchange, never on the API itself.
+    expect(init.headers?.Authorization).toBe(
+      `Basic ${Buffer.from('key-abc:secret-xyz').toString('base64')}`
+    );
+  });
+
+  it('sends the bearer token — not Basic — to the reporting API', async () => {
+    queueTokenThen({ requests: [] });
+
+    await makeProvider().syncEvents(new Date('2026-08-01'), new Date('2026-08-02'));
+
+    const apiCall = requestJsonMock.mock.calls[1]!;
+    expect(urlOf(apiCall)).toContain('reports.api.umbrella.com');
+    expect(initOf(apiCall).headers?.Authorization).toBe('Bearer tok-1');
+  });
+
+  it('sends the bearer token to the policies API too', async () => {
+    queueTokenThen({});
+
+    await makeProvider().addBlocklistDomain('evil.example', 'because');
+
+    const apiCall = requestJsonMock.mock.calls[1]!;
+    expect(urlOf(apiCall)).toContain('api.umbrella.com/policies/v2/destinationlists');
+    expect(initOf(apiCall).headers?.Authorization).toBe('Bearer tok-1');
+  });
+
+  it('caches the token across calls instead of re-exchanging per request', async () => {
+    queueTokenThen({}, {});
+    const provider = makeProvider();
+
+    await provider.addBlocklistDomain('a.example');
+    await provider.addAllowlistDomain('b.example');
+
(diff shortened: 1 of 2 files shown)


cisspUser01:
Follow-up: with the OAuth2 fix from #3275 live in 0.105.1 and a fresh "Umbrella API Keys" (modern, scoped) credential, authentication now succeeds — but sync still fails, this time with:

{"errors":[{"param":"from","value":"2026-08-14T00:26:06.924Z","error":"invalid timestamp specified"},{"param":"to","value":"2026-08-15T00:26:06.924Z","error":"invalid timestamp specified"}]}

Per Cisco's own OpenAPI spec for this endpoint, from/to should be Unix epoch milliseconds as a numeric string (e.g. "1639146300000"), not ISO 8601. UmbrellaProvider.syncEvents currently sends since.toISOString()/until.toISOString() — likely just needs to change to String(since.getTime())/String(until.getTime()).

cisspUser01:
Checked directly against 0.107.0 (currently running) — this is still unfixed. UmbrellaProvider.syncEvents still builds the request with:

url.searchParams.set("from", since.toISOString());
url.searchParams.set("to", until.toISOString());

Same ISO 8601 format that Cisco's own OpenAPI spec rejects for this endpoint (expects Unix epoch milliseconds as a numeric string, e.g. "1639146300000").

Also checked 0.108.0's full changelog — no mention of Umbrella, DNS, or this endpoint anywhere in it either, so this looks to still be open there too, unless it was addressed outside the published notes.

Happy to retest as soon as a fix lands — auth itself works fine now (confirmed weeks ago with a modern "Umbrella API Keys" credential), this is purely the request-format issue at this point.

Issue #3269, opened 2026-08-08
[Agent][Backup] VSS shadow copy lifetime: IVssBackupComponents is released before the copy is consumed — likely true cause of mid-run snapshot loss


ToddHebebrand:
## Description

The shadow-copy disappearance in #3260 (snapshot stopped resolving ~390s into the run, turning every remaining healthy file into `ERROR_PATH_NOT_FOUND`) is mitigated by PR #3266 — the run now detects the dead snapshot root and aborts explicitly instead of failing every file. But the mitigation does not explain or remove the disappearance itself, and the most likely cause is a COM lifetime bug in our VSS requester.

## Root cause analysis

`CreateShadowCopy` releases the `IVssBackupComponents` before returning (`agent/internal/backup/vss/vss_windows.go:233`, `defer callVtable(backupComponents, vtblRelease)`), justified by this comment (`:226-232`):

> The shadow copies created below are non-persistent VSS_CTX_BACKUP copies, which Windows reclaims when the requester *process* exits — not when this interface is released. … Verified on Server 2022: the device stays readable after Release and CoUninitialize for the life of the process.

Microsoft documents the opposite: `VSS_CTX_BACKUP` is an **auto-release** context, and auto-release shadow copies are deleted when the requester releases its references — not at process exit. The cited live test (`TestLive_CreateShadowCopy_EndToEnd`) only proves the device is readable *immediately* after the function returns; deletion of an auto-release copy is not necessarily synchronous with the `Release`, so a short read-through proves nothing about a copy surviving a multi-minute run.

This fits the #3260 timeline (`HarddiskVolumeShadowCopy26` unresolvable at ~+390s while the process was still running) at least as well as diff-area exhaustion, and arguably better: nothing in that run was writing enough to churn a diff area.

There is a second, already-acknowledged symptom of the same lifetime problem: `ReleaseShadowCopy` is a no-op and never signals `BackupComplete`, leaving writers in `VSS_WS_WAITING_FOR_BACKUP_COMPLETE` until process exit (comment at `vss_windows.go:436-449`, noted there as a follow-up on #2999).

## Proposed fix

Keep the `IVssBackupComponents` alive for the whole backup session on a dedicated COM-initialized thread (`runtime.LockOSThread`), and release it — after signalling `BackupComplete` to the writers — only when the run finishes. That is the lifetime rewrite the `ReleaseShadowCopy` comment already anticipates. It fixes both:

[...]

PR #3280, opened 2026-08-08
test(backup): make vss.Provider injectable to pin RunBackupContext liveness wiring (#3270)

### LINKED ISSUE #3270: [Agent][Backup] Make vss.Provider injectable so the RunBackupContext snapshot-liveness wiring is testable
## Description

PR #3266 introduced `agent/internal/backup/source_liveness.go` — a probe over the shadow-copy roots that aborts a run when the VSS snapshot disappears mid-run (#3260). The probe itself has unit coverage, but the two lines in `RunBackupContext` (`agent/internal/backup/backup.go`) that construct the probe from the live VSS session and pass it into the upload path are untested: exercising them requires a real `vss.Provider`, which is Windows-only and process-global.

If a refactor drops or misroutes that wiring, every liveness test still passes while the protection is silently disconnected — the same shape of gap that let #3260 ship.

## Proposed fix

Make the VSS provider injectable into the backup manager / `RunBackupContext` (constructor parameter or field with the current global as default) so a test can supply a fake provider whose session exposes controllable roots, then assert end-to-end that a root going away mid-run aborts with the explicit snapshot-expired error.

## Affected files

- `agent/internal/backup/backup.go` (provider acquisition + probe construction)
- `agent/internal/backup/vss/` (`Provider` interface already exists; needs a seam, not a redesign)

## Related

- #3260 / PR #3266 — the wiring this would pin
- #3269 — the COM lifetime rewrite would touch the same session plumbing; coordinate if both are picked up

## Provenance

Follow-up from the #3259/#3260 fix work (PR #3266), internal release QA for v0.104.0.

ToddHebebrand:
Fixes #3270

## The gap

PR #3266 added the mid-run snapshot-loss guard (#3260): a probe over the VSS shadow-copy roots that aborts a run when the snapshot disappears, rather than draining the file list and recording the fallout as a pile of per-file verdicts. The probe has thorough unit coverage. The two lines in `RunBackupContext` that **build it from the live VSS session and hand it to the upload loop** had none — reaching them needed a real `vss.Provider`, which is Windows-only and process-global.

That is the dangerous shape: a refactor could drop or misroute the wiring while every liveness test still passed and the protection was silently disconnected.

## The seam

A nullable `BackupConfig.VSSProvider vss.Provider` field, mirroring the existing `Provider providers.BackupProvider` field, plus a small `BackupManager.resolveVSSProvider() (vss.Provider, bool)` that replaces the inline gate:

```go
// before
if m.config.VSSEnabled && runtime.GOOS == "windows" {
    provider := vss.NewProvider(vss.DefaultConfig())

// after
if provider, useVSS := m.resolveVSSProvider(); useVSS {
```

Resolution order: `!VSSEnabled` → off; injected provider → use it; else Windows → `vss.NewProvider(vss.DefaultConfig())`; else off. **Nil — every production caller — resolves exactly as before.**

Two judgement calls, both deliberate:

- **An injected provider bypasses the `runtime.GOOS` gate.** That gate was never a statement about platform capability; it exists because `vss.NewProvider` compiles to a stub off Windows, and calling it there would fail every run with `ErrVSSNotSupported` and stamp a spurious "VSS failed" warning on the job. Supplying a provider is the caller asserting it works on this host. Honouring GOOS against it would make the seam unusable on the one platform the package is actually tested on.
- **`VSSEnabled` stays the master switch.** Injecting a provider must not switch VSS on for a run that did not ask for it.

The seam is scoped to the **provider only**. It says nothing about how a session is created, held or released, so the COM-lifetime rewrite in #3269 can change that plumbing underneath without touching this field.

## One incidental fix

A `(nil, nil)` return from `CreateShadowCopy` is now converted into the ordinary "VSS failed, read live" outcome. The success branch dereferences the session immediately (`session.Warnings`), and that branch is now reachable from a provider this package does not own — a contract violation should degrade visibly, not panic the agent run.

## Tests

`agent/internal/backup/vss_provider_seam_test.go` — cross-platform, no build tags, no Windows-only imports, so it runs on the linux + race **Test Agent** jobs.

- **`TestRunBackupContext_InjectedVSSProviderDrivesSnapshotLiveness`** — table-driven over the two outcomes the wiring exists to tell apart. A shadow root that disappears mid-run aborts the whole run with `errSourceSnapshotGone`, `ErrorCount 0`, and the resumable prefix left intact; a root that survives leaves the identical upload failure as one ordinary per-file failure and the run finishes. Both cases drive the **real `os.Stat` against a real directory** — a probe fed by a stubbed `snapshotRootStat` would pass even with the wiring cut, which defeats the purpose.
- **`TestRunBackupContext_InjectedVSSProviderReceivesRequestedVolumes`** — the volumes handed to the provider come from the *configured* paths, not the rewritten ones.
- **`TestRunBackupContext_VSSProviderReturningNoSessionDoesNotPanic`**
- **`TestResolveVSSProvider`** — the resolution matrix, GOOS-aware.

**Verified non-vacuous:** replacing `sourceLiveness = newShadowRootLiveness(...)` with `_ = newShadowRootLiveness(...)` reddens the abort case with `want the run to abort with errSourceSnapshotGone, got failed to upload ...`. Restored before commit.

## CI filter

No existing backup test is renamed, so the hardcoded Windows `-run` filter and its PASS-count assertion (`.github/workflows/ci.yml`) are untouched. The new tests are deliberately **not** added to it: their semantics are platform-neutral, the drive-letter rewrite they would exercise on Windows is already covered by `TestRewritePathsForVSS` (which is in that filter), and the mirrored temp-dir paths the test builds would sit close to `MAX_PATH` on a Windows runner.

## Verification

- `go build ./...`
- `go vet ./internal/backup/...` — host and `GOOS=windows` (only pre-existing `unsafe.Pointer` notes in untouched `vss_windows.go`)
- `go test -race ./internal/backup/...` — all 7 packages green
- `go test -race ./cmd/breeze-backup/...` — green

### DIFF
--- a/agent/internal/backup/backup.go
+++ b/agent/internal/backup/backup.go
@@ -61,6 +61,24 @@ type BackupConfig struct {
 	VSSEnabled         bool   // Windows only: create VSS shadow copy before backup
 	SystemStateEnabled bool   // Collect system state alongside file backup
 	StagingDir         string // Base directory for temporary staging (empty = OS temp dir)
+
+	// VSSProvider overrides where a VSS-enabled run gets its provider from.
+	// Nil — the production case, and what every real caller sets — means
+	// "use the platform provider", i.e. vss.NewProvider on Windows and no VSS
+	// anywhere else. See resolveVSSProvider for the exact resolution.
+	//
+	// It exists because the snapshot-liveness wiring in RunBackupContext
+	// (#3260/#3266: build the shadow-root probe from the live session and hand
+	// it to the upload loop) was otherwise unreachable from a test — the only
+	// real provider is Windows-only and process-global, so a refactor could
+	// drop or misroute that wiring while every liveness unit test still passed
+	// and the protection was silently disconnected (#3270).
+	//
+	// Scoped to the provider ONLY, deliberately. It says nothing about how a
+	// session is created, held or released, so the COM-lifetime rewrite of that
+	// plumbing (#3269) is free to change the session's shape underneath without
+	// touching this field.
+	VSSProvider vss.Provider
 }
 
 // BackupJob tracks the state of a backup run.
@@ -179,6 +197,33 @@ func (m *BackupManager) GetSystemStateEnabled() bool {
 // before a file-mode backup (Windows only; see BackupConfig.VSSEnabled).
 func (m *BackupManager) GetVSSEnabled() bool {
 	return m.config.VSSEnabled
+}
+
+// resolveVSSProvider decides whether a run takes the VSS path, and with which
+// provider. The second return is the whole gate: false means "no shadow copy
+// this run" and the caller skips the block entirely.
+//
+// An injected BackupConfig.VSSProvider bypasses the runtime.GOOS check on
+// purpose. That check is not a statement about the platform, it is a statement
+// about the only provider that used to be reachable here: vss.NewProvider
+// compiles to a stub off Windows, and calling it there would fail every run
+// with ErrVSSNotSupported and stamp a spurious "VSS failed" warning onto the
+// job. Supplying a provider is the caller asserting it works on this host, so
+// re-deriving that answer from GOOS would just make the seam unusable off
+// Windows — which is exactly the platform the wiring needs to be testable on.
+// Nothing in production sets the field, so the production decision is
+// unchanged.
+func (m *BackupManager) resolveVSSProvider() (vss.Provider, bool) {
+	if !m.config.VSSEnabled {
+		return nil, false
+	}
+	if m.config.VSSProvider != nil {
+		return m.config.VSSProvider, true
+	}
+	if runtime.GOOS != "windows" {
+		return nil, false
+	}
+	return vss.NewProvider(vss.DefaultConfig()), true
 }
 
 // Stop cancels an in-flight backup job and waits for it to unwind. It reports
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (behavior-parity, concurrency, the `(nil, nil)` guard, test non-vacuity/flakiness, CLAUDE.md Go conventions)

**Findings:** 0 raised, 0 outstanding. Reviewer confirmed independently: `resolveVSSProvider()` reproduces the old gate exactly when `VSSProvider` is nil (the only production state); the unlocked `m.config` read matches the existing `Get*` accessor pattern, not a new race; the `(nil, nil)` guard sits before the `#3027` warning branch with no nil deref and no double-release, since `defer ReleaseShadowCopy` only registers in the `else` branch the guard prevents; `shadowedSourceDir` mirrors `rewritePathsForVSS` exactly, so the test is correct under both `VolumeName` semantics.

**Design quorum:** the seam shape was cross-checked with Codex (read-only, `high`) before implementing. It independently reached the same two conclusions — nullable `BackupConfig` field over a constructor param / package-level var / setter, and an injected provider correctly overriding the `runtime.GOOS` gate with `VSSEnabled` kept as the master switch. Its one substantive catch, that `(nil, nil)` from `CreateShadowCopy` becomes a panic once arbitrary providers are injectable, is the incidental fix in this PR.

**Tests:** `go test -race ./internal/backup/...` green (all 7 packages); `./cmd/breeze-backup/...` green; `go vet` clean on host and `GOOS=windows` (only pre-existing `unsafe.Pointer` notes in untouched `vss_windows.go`). **Non-vacuity verified by mutation:** replacing `sourceLiveness = newShadowRootLiveness(...)` with `_ = newShadowRootLiveness(...)` reddens the abort case — which is the entire point of #3270.

**CI:** 54/54 checks green on `1a4e56d`, including **Test Agent** (linux + race) and every Windows agent job. No existing backup test renamed, so the hardcoded Windows `-run` filter and its PASS-count assertion are untouched.

**Status:** review-clean, awaiting maintainer merge. Note for sequencing — #3269 (the COM-lifetime VSS session rewrite) builds directly on this seam, so it wants this merged first.

Issue #3277, opened 2026-08-08
[Flaky] InvoiceWorkspace queued-Issue tests fail intermittently in CI (Test Web)


ToddHebebrand:
Two different tests in `apps/web/src/components/billing/InvoiceWorkspace.test.tsx` failed on consecutive CI runs of PR #3264 (a PR that touches nothing billing-related), each passing locally (15/15) and on rerun with no code change:

- run 31264337259: `holds a header Issue click while the editor is dirty, then fires it when the save settles` (1038ms)
- run 31265650231: `does NOT fire a queued Issue when the pending save fails — the whole point of the failure nonce`

Both are in the queued-Issue/save-settling cluster, and both sit around the ~1s mark, which suggests a fake-timer or waitFor race in the dirty-save → queued-action flow rather than a product bug.

Not chased in #3264 — confirmed unrelated there. Filing so the next Test Web red on these tests has a home.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

bdunncompany:
This is the same flake cluster as #3219, which is still open — and #3219 already carries your own file-wide analysis of it.

The lineage, since it's now fragmented across three issues:

- **#2925** (2026-07-29) — `does NOT fire a queued Issue when the pending save fails`, first seen on dependabot PR #2906
- **#3219** (2026-08-07) — `holds a header Issue click…`, which you closed #2925 onto as a duplicate, noting that two sibling cases in one file failing identically means *"the fix must be file-wide (fake timers or raised waitFor)"*
- **#3277** (this one) — both of those same two tests, on #3264

So the two runs you caught here are precisely the two cases already consolidated in #3219. Third sighting in ten days, which is a decent argument that it's worth actually fixing rather than logging again.

Worth closing one of the two so there's a single home — I'd suggest keeping #3219 since your file-wide diagnosis is already there and #2925 is already pointed at it, but it's your call and I don't want to close an issue of yours.

I'm happy to take the fix if you want it off your plate: the shape you described (fake timers, or raise the `waitFor` on the whole dirty-save → queued-action cluster) is well-defined. Say the word and I'll claim it on #3219.

bdunncompany:
Third occurrence, same cluster — giving it the home you filed this for.

- **Run** `31434067431`, Test Web job `93604086183`, on PR #3396
- **Test** `does NOT fire a queued Issue when the pending save fails — the whole point of the failure nonce` (the same one from run `31265650231`)
[...]

PR #3264, opened 2026-08-08
feat(devices): fleet migration posture report (#3244)

ToddHebebrand:
Closes #3244 (epic #3249).

Surfaces the management-posture detection Breeze already collects (`agent/internal/mgmtdetect/` → `devices.management_posture` jsonb) as a fleet-level migration/decommission report. Surfacing only: no agent changes, no new tables/columns — the jsonb stays the single source of truth.

Spec: `docs/superpowers/specs/onboarding-signup/2026-08-08-fleet-migration-posture-report-design.md` · Plan: `docs/superpowers/plans/open/2026-08-08-fleet-migration-posture-report.md` (both on the #3250 docs branch).

## API

- **`services/managementPostureReport.ts`** — **two aggregate queries, per the plan's critical note**:
  - *Detections*: `CROSS JOIN LATERAL jsonb_array_elements(...)`, `count(DISTINCT devices.id)` per (org, product, status), plus a fresh-count `FILTER` on `collectedAt`.
  - *Coverage*: per-org denominators with **no lateral join** — `totalDevices`, `neverScanned` (posture `IS NULL`), `stale`, `scannedNoneDetected` (empty array OR absent key), `detectedDevices`, `freshDetectedDevices`.
  - Deliberately **not** one `GROUP BY` with a `LEFT JOIN LATERAL`: that collapses never-scanned / category-absent / empty-array into one `(NULL, NULL)` group, making a never-scanned device read as verified-clean — the exact failure that strands endpoints in Phase 6 of a migration.
  - The summary assembler never emits a detection count without its coverage denominators (throws on the impossible orphan case rather than dropping it).
  - `category` is validated against the ingest enum (`MANAGEMENT_POSTURE_CATEGORIES`, now exported from `routes/agents/schemas.ts` as the single source of truth) and rides a bind param with an explicit `::text` cast — never string interpolation.
- **`routes/devices/posture.ts`** — `GET /devices/management-posture/summary` (`orgId?`, `category?`=rmm, `stalenessDays?`=7) and `GET /devices/management-posture/devices` (`product`, `category?`, `status?`, pagination). Scoping mirrors `GET /devices/stats`: `authMiddleware` + `requireScope` + `DEVICES_READ`, `auth.orgCondition`, 403 on inaccessible `?orgId`, site-allowlist narrowing (empty allowlist ⇒ empty report), decommissioned + ephemeral devices excluded. Mounted **before** `coreRoutes` so `/management-posture/*` isn't eaten by the `/:id` matcher.

[...]

Issue #2925, opened 2026-07-29
[Web] InvoiceWorkspace failure-nonce test flaky under CI load (waitFor timeout on invoice-issue-saving-hint)


ToddHebebrand:
**Symptom:** `apps/web/src/components/billing/InvoiceWorkspace.test.tsx:208` — "does NOT fire a queued Issue when the pending save fails — the whole point of the failure nonce" — failed on dependabot PR #2906's Test Web run (17:12 UTC, 2026-07-29) with `Unable to find an element by: [data-testid="invoice-issue-saving-hint"]` after `fireEvent.blur` on `invoice-notes`; `waitFor` timed out before the PATCH-in-flight indicator (`InvoiceActions.tsx:365`) rendered. 1 test of 4642 failed; a rerun of the same commit passed.

**Evidence it's a flake, not jsdom 30:** the identical `apps/web` tree at the same jsdom 30.0.1 passed a Test Web run ~20 minutes earlier on the same PR (`git diff 604bd83c..01d7c681 -- apps/web` is empty; the rebase between the green and red runs only absorbed react-native and js-yaml bumps, neither touching web).

**Root cause shape:** the test drives a real (non-fake-timer) promise race — manual `resolvePatch`/`rejectPatch` handles plus `waitFor`'s default 1000ms/50ms polling — so it's timing-sensitive under CI load.

**Fix options:** switch the test to fake timers, or raise the `waitFor` timeout for the saving-hint assertion (and any sibling assertions in the same race pattern within the file).

**Status:** tracking issue only; no production impact. First observed occurrence — close if it never recurs after the timing fix.

ToddHebebrand:
Duplicate of #3219, which carries a second occurrence — a different test case in the same file ("holds a header Issue click while the editor is dirty, then fires it when the save settles"), failing on the same `invoice-issue-saving-hint` assertion. Two sibling cases flaking identically confirms the cause is the shared failure-nonce race pattern rather than either individual test, so the fix belongs to the file, not the case.

Consolidating there; the root-cause analysis from this issue is reproduced in that thread. Closing as duplicate.

Issue #3018, opened 2026-08-02
[CI][Agent] internal/sessionbroker intermittently hangs under Windows -race and burns the full 10m timeout (10s without -race)


ToddHebebrand:
## Description

`Test Agent (Windows)` → `Run Windows race tests` intermittently hangs in `internal/sessionbroker` and is killed by the 10-minute Go test timeout:

```
panic: test timed out after 10m0s
FAIL	github.com/breeze-rmm/agent/internal/sessionbroker	600.080s
```

**The same package, same commit, in the non-race Windows step, finishes in under 10 seconds:**

```
Run Windows Go tests   ok    github.com/breeze-rmm/agent/internal/sessionbroker    9.837s
Run Windows race tests FAIL  github.com/breeze-rmm/agent/internal/sessionbroker  600.080s
```

A 60x blowup to exactly the timeout budget is a hang, not slowness. `-race` slows execution perhaps 2-20x; it does not turn 10 s into >600 s.

## Evidence it is intermittent, not deterministic

Same Go tree, three runs:

| where | result |
|---|---|
| PR #3003 (introduced the tests) | **pass** |
| `main` @ `f59f77c2` (contains #3003) | **pass** |
| PR #3015 (a migration rename — **zero Go files changed**) | **hang → timeout** |

#3015 changes three files: one migration rename plus two reference updates, all under `apps/api`. It cannot affect Go behaviour, which makes the failure non-deterministic by elimination.

## Suspected area

The hang surfaced right after #3003 (#2998) added its backup-helper-death concurrency suite. Log lines immediately before the panic are that suite's:

```
level=INFO msg="session recv loop ended" component=sessionbroker sessionId=backup-death
level=WARN msg="backup helper exited with a run in flight, failing the job" component=backup
level=INFO msg="session recv loop ended" component=sessionbroker sessionId=backup-async
```

Those tests deliberately drive high-iteration concurrent probes (300 and 900 iterations were cited during review) and one pins `runtime.GOMAXPROCS(1)`. Under `-race` on a Windows runner that combination is a plausible livelock/deadlock, or simply unbounded scheduling latency. **This is a suspicion, not a diagnosis** — nobody has reproduced it locally yet.

Note the tests are correct on the merits: the same suite demonstrably proved the fix (150/200 contradictory results before, 0/900 after). The problem is the harness cost under `-race`, not the assertions.

## Why it matters

[...]

PR #3284, opened 2026-08-08
test(web): give the InvoiceWorkspace queued-Issue tests real waitFor headroom (#3219)

### LINKED ISSUE #3219: [Web] Flaky test: InvoiceWorkspace "holds a header Issue click…" intermittently fails on invoice-issue-saving-hint
## Summary

`InvoiceWorkspace.test.tsx` › **"holds a header Issue click while the editor is dirty, then fires it when the save settles"** fails intermittently in the web suite, with:

```
TestingLibraryElementError: Unable to find an element by: [data-testid="invoice-issue-saving-hint"]
```

It is one failing test out of 534 files, so it reddens the whole `Test Web` job and the `CI Success` aggregate gate with it.

## Two confirmed occurrences, both on unrelated changes

1. **`main` @ `38150ac2f`** (2026-08-06). The merge that produced that commit was #3179, which touches **zero** billing files. Re-running the identical commit with no code change **passed**, which is what established it as a flake rather than a regression.
2. **#3213** (2026-08-07), an `apps/api`-only change that touches no web code at all. Same test, same assertion.

Occurrence 2 is the reason this is worth filing: the flake now blocks unrelated pull requests from going green.

## Why it is timing-shaped

The assertion targets `invoice-issue-saving-hint`, which by design only exists **while a save is in flight**:

```ts
fireEvent.change(screen.getByTestId('invoice-notes'), { target: { value: 'Edited' } });
fireEvent.blur(screen.getByTestId('invoice-notes'));
await waitFor(() => expect(screen.getByTestId('invoice-issue-saving-hint')).toBeInTheDocument());
```

The PATCH is a promise deliberately left unresolved until later in the test, so the save should stay pending for the whole window — which is why the failure is surprising and why I have not guessed at a fix.

## What I could not establish

**It does not reproduce locally.** 6/6 clean runs of that file on Node 22.23.2 on an idle machine. That is consistent with a load-sensitive race that only surfaces on a busy runner, but I could not prove the mechanism, so I have deliberately not sent a speculative patch to the save/hint state machine — whoever owns that flow will know in a minute what took me an hour of guessing.

Two plausible directions for whoever picks it up, neither verified:
- the blur-triggered save has not been dispatched yet when `waitFor` gives up, so the hint has genuinely not rendered (a debounce or effect-ordering window);
- the hint's visibility depends on a second piece of state that has already settled by the time the assertion runs.

Happy to build the fix if you can point at which of those it is, or if you would rather I dig in properly.

bdunncompany:
Fixes #3219. Also covers #3277, which is the same two tests — worth closing as a duplicate once this lands.

## Why these two tests and no others

Three sightings in ten days, always the same pair, always green locally and on a same-commit rerun:

| | when | test |
|---|---|---|
| #2925 | 2026-07-29 | `does NOT fire a queued Issue when the pending save fails` |
| #3219 | 2026-08-07 | `holds a header Issue click…` |
| #3277 | 2026-08-08 | both, on #3264's CI |

They are the only two in the file that assert the **end of a multi-hop propagation chain**: the PATCH promise settles → the editor reports saved/failed → the workspace clears `savePending` → the header un-gates → the queued Issue fires a fetch. Every hop is promise/render scheduling.

All three recorded failures sat at **~1038ms** against Testing Library's **1000ms** default `waitFor` timeout. A fraction over the default is the signature of a timeout, not a logic race — and it explains why a rerun on the identical commit passes.

## Why not fake timers

That was your other suggested option and it's the wrong instrument here: **nothing in this chain is timer-driven.** The delay is promise microtasks plus React render scheduling, neither of which fake timers advance. They'd add machinery without touching the cause, and would need the manual `resolvePatch`/`rejectPatch` promises reworked around them.

Raised file-wide as you specified, rather than per-assertion, so a future test in this file inherits the headroom instead of rediscovering the flake. Restored in `afterAll` since `configure` is process-global.

## What I verified — and what I couldn't

**Could not reproduce the flake locally.** The chain completes in ~30ms on my machine against a 1000ms budget, which is precisely why this only ever fails on loaded CI. I'm not going to claim a green local run as proof of a fix for a CI-only timing failure.

**What I could prove** is that the knob governs the right assertions rather than being inert — setting `asyncUtilTimeout` to 1ms fails **exactly** the two queued-Issue tests and leaves the other five passing:

```
✓ renders a draft as the editor…            ✓ keeps the draft editor MOUNTED…
✓ surfaces an error card…                   ✓ offers the cost/margin toggle…
✓ updates the header from "Draft invoice"…
× holds a header Issue click while the editor is dirty…
× does NOT fire a queued Issue when the pending save fails…
```

The real proof is CI staying green on these two over the next few weeks; if it recurs, the diagnosis is wrong and the next step is instrumenting the chain rather than raising the number again.

## Scope note

No other web test file configures `asyncUtilTimeout` — the whole suite runs on the 1000ms default. I deliberately did **not** raise it globally: that would also slow every genuine failure to the new ceiling, and the blast radius is the entire web suite. Worth considering separately if this shape shows up in other files.

## Gate

`apps/web` vitest — **538 files / 5102 tests / 0 failures** · `tsc --noEmit` exit 0 · `eslint` clean.

### DIFF
--- a/apps/web/src/components/billing/InvoiceWorkspace.test.tsx
+++ b/apps/web/src/components/billing/InvoiceWorkspace.test.tsx
@@ -1,5 +1,5 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { configure, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import InvoiceWorkspace from './InvoiceWorkspace';
 import { _resetShowMarginMemoryForTests } from './billingUi';
@@ -49,6 +49,32 @@ function invoice(over: Record<string, unknown>) {
 }
 
 describe('InvoiceWorkspace', () => {
+  // #3219 — the queued-Issue tests assert the END of a multi-hop propagation
+  // chain: the PATCH promise settles → the editor reports saved/failed → the
+  // workspace clears `savePending` → the header un-gates → the queued Issue
+  // fires a fetch. Every hop is promise/render scheduling, so under CI load the
+  // whole chain can outrun Testing Library's 1000ms default `waitFor` timeout.
+  //
+  // That is exactly what the three recorded sightings look like: #2925, this
+  // issue, and #3277 all failed at ~1038ms, i.e. a fraction over the default,
+  // and all passed on a same-commit rerun and locally.
+  //
+  // Fake timers are the wrong instrument here despite being the usual reflex —
+  // nothing in this chain is timer-driven. The delay is promise microtasks plus
+  // React render scheduling, which fake timers do not advance, so they would add
+  // machinery without touching the cause.
+  //
+  // Raised file-wide rather than per-assertion so a future test in this file
+  // inherits the headroom instead of re-discovering the flake. Restored
+  // afterwards because Testing Library's `configure` is process-global.
+  const DEFAULT_ASYNC_UTIL_TIMEOUT_MS = 1000;
+  beforeAll(() => {
+    configure({ asyncUtilTimeout: 5000 });
+  });
+  afterAll(() => {
+    configure({ asyncUtilTimeout: DEFAULT_ASYNC_UTIL_TIMEOUT_MS });
+  });
+
   beforeEach(() => {
     vi.clearAllMocks();
     // The margin preference persists to localStorage plus an in-memory mirror


bdunncompany:
**Fourth sighting, today.** Same test, on a PR that cannot possibly have caused it.

`Test Web` went red on #3306 — a PR whose entire diff is one **API** contract test file, touching nothing under `apps/web`:

```
FAIL src/components/billing/InvoiceWorkspace.test.tsx > InvoiceWorkspace
     > holds a header Issue click while the editor is dirty, then fires it when the save settles
Test Files  1 failed | 538 passed (539)
```

Run 31312800729, 12:25Z. That takes the table in the PR body to:

| | when | test |
|---|---|---|
| #2925 | 2026-07-29 | `does NOT fire a queued Issue when the pending save fails` |
| #3219 | 2026-08-07 | `holds a header Issue click…` |
| #3277 | 2026-08-08 | both, on #3264's CI |
| — | **2026-08-09** | `holds a header Issue click…`, on **#3306** |

Sighting four also strengthens the diagnosis rather than just repeating it: an API-only diff has no mechanism to perturb this component, so whatever moved was the runner's scheduling, which is exactly the ~1038ms-against-a-1000ms-default timeout story in the PR body. A same-commit rerun is green again, as before.

The cost is no longer theoretical — it now reddens unrelated PRs and forces a rerun to clear them, which is the same tax #3018 charges on the Windows agent side.

No competing PR exists for this (checked), and #3219 and #3277 are both still open and unassigned.

PR #3285, opened 2026-08-08
fix(backup): hold the VSS session COM references for the whole run and signal BackupComplete (#3269)

### LINKED ISSUE #3269: [Agent][Backup] VSS shadow copy lifetime: IVssBackupComponents is released before the copy is consumed — likely true cause of mid-run snapshot loss

ToddHebebrand:
Fixes #3269
Context: #3260, #2999

## The bug

`CreateShadowCopy` released its `IVssBackupComponents` before returning, justified by an in-code comment claiming that a `VSS_CTX_BACKUP` shadow copy is reclaimed only when the requester **process** exits. Microsoft documents the opposite: `VSS_CTX_BACKUP` is an **auto-release** context, and auto-release copies are deleted when the requester drops its references.

That is the likely true cause of #3260 — a shadow device that stopped resolving ~390s into a live run, after which every remaining healthy file failed with `ERROR_PATH_NOT_FOUND` and was recorded as an individual per-file fault. PR #3266 added the mid-run abort that stops the misattribution; this PR removes the cause.

The old live test could never have caught it. It read through the device seconds after the Release, and deletion of an auto-release copy is not synchronous with it.

Second symptom, same root cause: `ReleaseShadowCopy` was a no-op that never called `BackupComplete`, leaving VSS writers parked in `VSS_WS_WAITING_FOR_BACKUP_COMPLETE` until the process exited (already noted in-code as a #2999 follow-up).

## The fix

The whole requester sequence now runs as a **single closure on a dedicated, COM-initialised, permanently pinned OS thread**, and the components object stays open on that thread until `ReleaseShadowCopy`, which signals `BackupComplete` to the writers before dropping the last reference.

**One closure, not per-call marshalling.** `callVtable` is `//go:uintptrescapes`, and its soundness requires every `uintptr(unsafe.Pointer(&x))` conversion to appear directly in a `callVtable(...)` argument list. Marshalling per COM call would have meant building argument values on one goroutine and consuming them on another — exactly the pattern the directive cannot protect. The single closure is also a near-zero-diff transformation of proven code.

**Thread discipline.** `runtime.LockOSThread` for the thread's whole life, never unlocked, so the goroutine's exit destroys the OS thread rather than returning it to the scheduler carrying our COM state. `CoInitializeEx(0, COINIT_MULTITHREADED)` once; `CoUninitialize` once at teardown — and only when init actually succeeded (`S_FALSE` still owes an uninitialize; `RPC_E_CHANGED_MODE` must not get one).

## Two design calls worth reading

Both were settled by an advisor quorum (independent read-only Codex review), and **both moved off my first position**:

**1. Snapshot creation is gated process-wide; sessions are not.** My first design gated the whole session. That would have been a user-visible regression: `breeze-backup` dispatches every IPC command in its own goroutine and builds an ephemeral `BackupManager` per server-dispatched `backup_run`, so overlapping runs are routine — and a session-wide gate would silently downgrade one of two concurrent runs to a live read for the entire duration of the other. What VSS actually serialises is the creation interval (`StartSnapshotSet` → `DoSnapshotSet`); it does not require that only one finished snapshot set be alive. So creation queues (bounded by the caller's creation context, since Microsoft's documented response to `VSS_E_SNAPSHOT_SET_IN_PROGRESS` is wait-and-retry), and both runs still get their own shadow copy. A run that gives up waiting returns `ErrVSSSessionInProgress` and takes the existing "proceed without VSS" path with a server-visible warning.

**2. No destructive lifetime watchdog.** My first design force-released a session after 24h to stop a forgotten release from wedging things. That is dangerous: `RunBackupContext` has no maximum duration, a large dataset over a slow link can legitimately run for many hours, and wall-clock time keeps passing while a laptop is suspended mid-run. A timer that deleted the snapshot on a threshold would delete live snapshots out from under healthy runs — **#3260, recreated deliberately by the change meant to fix it.** A stale session is now reported at ERROR and left alone; process exit remains the backstop. Narrowing the gate (call 1) is what makes that affordable: a leaked session costs one held snapshot, not the machine's ability to take further ones.

**`SetBackupSucceeded`** (slot 20) is declared and ABI-pinned but deliberately **not called**: the requester sets `bSelectComponents=FALSE` and never calls `AddComponent`, so there are no components to mark and `BackupComplete` alone is the complete handshake for a non-component backup. The constant exists so a later change inherits a checked slot number instead of counting header lines — guessing a slot is exactly how #2999 shipped.

**`ListWriters`** stays call-scoped with its own COM cycle. It creates no snapshot, and `exec_backup.go`'s `vss_status`/`vss_writer_list` commands poll it independently of any running backup — coupling it to the session thread would have made it fail whenever a backup was in flight.

## Comments corrected in the same PR

The false process-exit claim in `CreateShadowCopy`, the `ReleaseShadowCopy` design note, the `Provider` interface doc in `types.go`, and the `#3269` reference in `source_liveness.go` (the mid-run liveness probe **stays** — three of its four possible causes are unfixed and outside our control, so it remains as defence in depth).

## Tests

The lifetime machinery (`serialThread`, the creation gate) is deliberately **untagged**, so it runs under `go test -race` on every platform and in the linux CI job. The Windows CI job runs this package *without* the race detector, so tagging it would have left the concurrency untested by any detector anywhere.

Added:
- `TestSerialThread_RunsEveryCallOnOneLockedThread`, `TestSerialThread_TeardownRunsOnceOnTheSameThread`, `TestSerialThread_InitFailureStartsNothing`, `TestSerialThread_DoAfterCloseReports`, `TestSerialThread_SerialisesConcurrentCallers`
- `TestSnapshotCreation_SerialisesButDoesNotBlockForever`, `TestSnapshotCreation_ContextExpiryDegrades`, `TestSnapshotCreation_ReleaseIsIdempotent`, `TestSnapshotCreation_ConcurrentCallersNeverOverlap`
- `TestBackupCompleteSlotIsDistinctFromNeighbours`, `TestSetBackupSucceededSlotIsPinned` (windows-tagged ABI pins; these DO run in Windows CI via the unfiltered `go test ./internal/backup/vss`)
- `TestRunBackupContext_ConcurrentSessionRejectionDegradesToLiveRead` (the degrade path stays a completed, warned, live-read backup — never a failure, never silent)
- `TestLive_ShadowCopySurvivesALongHold`, `TestLive_ConcurrentSessionsCoexist` (live suite)

**Mutation-checked**, the way #3280 did:
- `do()` bypassing the pinned thread → 3 tests red
- creation gate not serialising → 4 tests red
- (both restored green)

`TestLive_CreateShadowCopy_EndToEnd`'s central assertion is **inverted**: it used to prove the device survived `CreateShadowCopy`'s own Release, which was the wrong claim and unfalsifiable at that timescale. It now proves the device is readable while the session is held, that release is idempotent, and that the provider is reusable afterwards.

**No CI filter changes needed.** The `-run` regex + exactly-14-PASS assertion in `ci.yml` covers `./internal/backup`; no existing test was renamed, and the one test added there (`TestRunBackupContext_...`) does not match any filtered prefix, so the count is unchanged.

## Verification

- `cd agent && go test -race ./...` — green (full suite)
- `GOOS=windows go build ./...` and `GOOS=windows go vet ./...` — clean (the 3 `possible misuse of unsafe.Pointer` notes are pre-existing, inside `callVtable` and `utf16PtrToString`)
- `gofmt` clean

**Real-Windows verification is REQUIRED and still PENDING.** The COM call sequence cannot be faked, so nothing above proves the fix on real hardware. On an elevated shell on a Windows host:

```
set BREEZE_VSS_LIVE=1
go test ./internal/backup/vss/ -run Live -v -timeout 40m
```

`TestLive_ShadowCopySurvivesALongHold` is the #3269 regression guard: it holds the session for 8 minutes (past the ~390s at which the field failure appeared), probing the shadow path every 30s. Override with `BREEZE_VSS_HOLD_SECONDS`. **It should FAIL on a build without this change** — that is how to confirm the diagnosis, not just the fix. `TestLive_ConcurrentSessionsCoexist` checks that releasing one session does not take another's device with it.

Please hold merge until that live run is done.

### DIFF
--- a/agent/internal/backup/source_liveness.go
+++ b/agent/internal/backup/source_liveness.go
@@ -35,11 +35,17 @@ import (
 // Deliberately NOT called "expired": a VSS_CTX_BACKUP shadow copy has no TTL,
 // so nothing here times out. It becomes unavailable — deleted by the provider,
 // lost to diff-area exhaustion, removed externally, or auto-released when
-// IVssBackupComponents is released (Microsoft-documented behaviour for
-// auto-release VSS_CTX_BACKUP copies; the contrary note in
-// vss.CreateShadowCopy, which claims reclamation only at requester process
-// exit, is tracked as #3269). The guard is agnostic about which; it only
-// reports that the source went away.
+// IVssBackupComponents is released. That last cause was the likely trigger of
+// the field repro above and has since been FIXED (#3269): vss.CreateShadowCopy
+// no longer releases the components object on its way out, and the session now
+// holds it open on a dedicated COM thread until the run ends.
+//
+// This guard stays as defence in depth. The other three causes are real and
+// entirely outside our control, and the whole point of the guard is that it is
+// agnostic about which one fired — it only reports that the source went away.
+// Removing it because one known cause was fixed would put us back to recording
+// a dead snapshot as a pile of per-file faults the next time one of the others
+// happens.
 var errSourceSnapshotGone = errors.New("backup source snapshot is no longer available (VSS shadow copy deleted or became unavailable mid-run)")
 
 // sourceLivenessFn reports a non-nil error when the point-in-time source that
--- a/agent/internal/backup/vss/session_lifetime.go
+++ b/agent/internal/backup/vss/session_lifetime.go
@@ -1,0 +1,209 @@
+package vss
+
+import (
+	"context"
+	"errors"
+	"runtime"
+	"sync"
+)
+
+// This file holds the two pieces of #3269's lifetime fix that are pure Go: the
+// pinned worker thread the COM calls are routed onto, and the process-wide gate
+// that serialises snapshot *creation*.
+//
+// Neither has a build tag, and that is deliberate. The COM calls they carry are
+// unfakeable and Windows-only, but the machinery around them — start a thread,
+// keep every call on it, release exactly once, don't let two runs create a
+// snapshot set at the same moment — is ordinary concurrent Go, and it is where a
+// mistake costs a customer their backup. Keeping it platform-neutral puts it
+// under `go test -race` on every developer machine and in the linux CI job,
+// instead of only in the Windows job, which runs this package without the race
+// detector.
+
+// ---------------------------------------------------------------------------
+// serialThread
+// ---------------------------------------------------------------------------
+
+// errSerialThreadClosed is returned by serialThread.do once the thread has
+// stopped accepting work.
+var errSerialThreadClosed = errors.New("vss: session thread is closed")
+
+// serialThread owns one OS thread and runs submitted closures on it, one at a
+// time.
+//
+// It exists to give the COM work a stable home. A VSS_CTX_BACKUP shadow copy is
+// an *auto-release* copy: Windows deletes it once the requester drops its
+// references to the IVssBackupComponents that created it. Holding that object
+// for the whole backup run means holding a thread that has COM initialised for
+// the whole run, and the Go scheduler may move a goroutine to a different OS
+// thread between any two statements — so the thread has to be pinned and every
+// call routed back onto it.
+//
+// The thread is locked for its entire life and is deliberately NEVER unlocked.
+// A goroutine that exits while still locked makes the Go runtime destroy the
+// underlying OS thread, so no unrelated goroutine can inherit a thread carrying
+// our COM state. Note what that does and does not do: an MTA is process-wide,
+// not a private apartment this thread owns, so destroying the thread does not
+// tear down the MTA, and it is emphatically NOT the mechanism that reclaims the
+// shadow copy. Releasing the IVssBackupComponents is.
+//
+// Two limits worth knowing. Concurrent callers are serialised but not ordered:
+// nothing here promises FIFO among goroutines racing to submit. And a COM call
+// that wedges inside a vtable dispatch cannot be preempted by anything —
+// context, close, or otherwise — because it is a synchronous foreign call on
+// this thread; only the polling *between* COM calls is context-bounded.
+type serialThread struct {
+	calls    chan func()
+	stop     chan struct{}
+	done     chan struct{}
+	stopOnce sync.Once
+}
+
+// newSerialThread starts the thread and runs init on it before accepting any
+// work. If init returns an error the thread exits without running teardown —
+// there is nothing initialised to tear down — and the error is returned here.
+// That distinction is load-bearing for COM: CoInitializeEx returning S_FALSE is
+// a success that still owes a CoUninitialize, while a genuine failure such as
+// RPC_E_CHANGED_MODE must never be uninitialized.
+//
+// teardown runs on the same thread after the last submitted closure has
+// finished, immediately before the thread exits.
+func newSerialThread(init func() error, teardown func()) (*serialThread, error) {
+	t := &serialThread{
+		calls: make(chan func()),
+		stop:  make(chan struct{}),
+		done:  make(chan struct{}),
+	}
+	ready := make(chan error, 1)
+
+	go func() {
+		// Never unlocked; see the type comment. This is the single most
+		// load-bearing line in the file.
+		runtime.LockOSThread()
+		defer close(t.done)
+
+		if init != nil {
+			if err := init(); err != nil {
+				ready <- err
+				return
+			}
+		}
+		if teardown != nil {
+			defer teardown()
+		}
+		ready <- nil
+
+		for {
+			// Shutdown wins over pending work. Without this the select below
+			// would pick nondeterministically between a ready call and a closed
+			// stop channel, so close() could still admit one more closure.
+			select {
+			case <-t.stop:
+				return
+			default:
+			}
+			select {
+			case fn := <-t.calls:
+				fn()
+			case <-t.stop:
+				return
+			}
+		}
+	}()
+
+	if err := <-ready; err != nil {
+		// Wait for the goroutine to actually finish, so a failed start never
+		// leaves a thread behind that a later close() would have to reap.
+		<-t.done
+		return nil, err
+	}
+	return t, nil
+}
+
+// do runs fn on the thread and blocks until fn returns. Anything fn writes to
+// variables captured from the caller is visible to the caller once do returns:
+// the channel send and the completion signal are the happens-before edges.
+//
+// It reports errSerialThreadClosed if the thread is already stopping, rather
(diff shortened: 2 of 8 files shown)


ToddHebebrand:
**Design quorum:** independent read-only Codex review of the session-thread design, run *before* implementing. It **disagreed with two of my five positions and I moved on both** — the record is in the PR body under "Two design calls worth reading":
- **Session-wide concurrency gate → narrowed to creation-only.** A session-wide gate would have silently downgraded one of two concurrent backup runs to a live read, which is a user-visible regression justified only by a theory. VSS serialises `StartSnapshotSet` → `DoSnapshotSet`, not the lifetime of finished snapshot sets.
- **24h force-release watchdog → dropped, replaced with a non-destructive ERROR alarm.** No elapsed-time signal separates an abandoned session from a slow-but-healthy one (no max run duration; wall-clock passes while a laptop is suspended). A threshold timer would have deleted live snapshots out from under healthy runs — #3260 recreated by the change meant to fix it.

**Review run:** `pr-review-toolkit` — code-reviewer (6 targeted hazard areas: leak-per-return-path, double-release/use-after-release, deadlock, the `//go:uintptrescapes` contract, gate leak, degrade paths) + silent-failure-hunter.

**Findings:** 2 raised → 1 fixed, 1 recorded as a deliberate known gap; 0 outstanding.

- **Fixed in `3ec4a92f0` — false concurrency contract in `types.go` (conf. 90).** The `Provider` docs still described the *pre-quorum* design: "only one session may be live per process… a second concurrent call fails rather than queuing." Both false after the gate was narrowed. Worth calling out that this is the exact bug class the PR fixes — a comment making a false claim about VSS lifetime semantics is what #3260 hid behind — so it did not get to ship inside the fix.
- **Not fixed, deliberately: a failed `BackupComplete` reaches the agent log but never `job.Warning`.** Every other VSS anomaly here is promoted to a server-visible warning (#3027), so this is a genuine inconsistency. Left open because the release runs from a `defer` in `RunBackupContext` *after* the job is built, so plumbing it through means mutating a completed job — not a change I'll make inside a data-loss fix. Consequence is bounded and self-healing: `Release` runs unconditionally on both branches, so the snapshot is still reclaimed; only writer hygiene is affected, until process exit. Now named explicitly in-code as a KNOWN GAP rather than left implied. Good follow-up issue.

**Tests:** `go test -race ./...` green across the whole agent; `GOOS=windows` build + vet clean (the 3 `possible misuse of unsafe.Pointer` notes are pre-existing, inside `callVtable`/`utf16PtrToString`). Both key tests mutation-checked — bypassing the pinned thread reds 3 tests, disabling the creation gate reds 4.

**CI:** 54/54 green. Notably `Test Agent (Windows)` passes unchanged, which confirms the `-run` filter and its exactly-14-PASS assertion needed no update — no existing test was renamed and the one test added to `./internal/backup` matches none of the filtered prefixes.

**Status: review-clean, CI green — but please DO NOT MERGE yet.** Real-Windows verification is required and still pending; none of the above exercises a single real COM call. On an elevated shell on a Windows host:

```
set BREEZE_VSS_LIVE=1
go test ./internal/backup/vss/ -run Live -v -timeout 40m
```

`TestLive_ShadowCopySurvivesALongHold` is the actual #3269 regression guard — an 8-minute hold past the ~390s at which the field failure appeared, probing the shadow path every 30s. **It should FAIL on a build without this change**, which is how to confirm the diagnosis and not merely the fix.

PR #3305, opened 2026-08-09
fix(web): walk every page of /scripts so the full library is reachable (#3301)

### LINKED ISSUE #3301: [Web] Scripts page and all three script pickers silently show only the first 50 scripts
## Summary

`GET /scripts` defaults to a 50-row page and caps at 100 (`routes/scripts.ts:39`). Four web call sites request it **with no `limit` and no pagination**, so any partner with more than 50 scripts silently sees only the first 50 — no error, no "showing 50 of N", no way to reach the rest.

```ts
// apps/api/src/routes/scripts.ts:39
const limit = Math.min(100, Math.max(1, Number.parseInt(query.limit ?? '50', 10) || 50));
```

## Affected call sites

| File | Line | Request | Impact |
|---|---|---|---|
| `components/scripts/ScriptsPage.tsx` | 69 | `/scripts` | **The main Scripts page.** No pagination state anywhere in the file — script 51+ is simply unreachable in the UI. |
| `components/devices/ScriptPickerModal.tsx` | 104 | `/scripts?includeSystem=true` | Run-script on a device, and bulk run-script |
| `components/scripts/ScriptCategoryTree.tsx` | 181 | `/scripts?includeSystem=true` | Category tree shows a partial library |
| `components/configurationPolicies/featureTabs/RemediationScriptPicker.tsx` | 73 | `/scripts` | Remediation script can't be chosen if it's script 51+ |

`ScriptsPage.tsx` is the worst of the four: it's the page whose entire job is listing scripts, and it has no pagination controls at all.

`ScriptBundleImport.tsx:70` is the one call site that gets it right — `/scripts?limit=100&page=${page}` in a loop — which is good news, since it means the paginated shape is already proven against this endpoint.

The same file also does `/devices?limit=10000` a few lines below the capped `/scripts` call (`ScriptsPage.tsx:87`), so the pattern was known; the scripts call just never got it.

## Downstream

`FixPickerModal` in the new fleet remediation flow (#3278) imports and reuses `ScriptPickerModal`, so fleet remediation inherits the cap: a fleet-wide fix can't be driven by a script that sorts past position 50. That's how this was found, but it is **pre-existing and not caused by #3278**.

## Suggested fix

`ScriptsPage` needs real pagination (it's a list page and should have had it). The three pickers are better served by a search-as-you-type against the server than by paging a dropdown — a 300-script library is unusable as a scroll list regardless of whether all 300 are loaded. If a quick mitigation is wanted first, `?limit=100` moves the cliff without removing it, so it should be treated as a stopgap and not a fix.

Worth also considering whether an endpoint that silently truncates is the right default — an explicit `pagination.total` is already returned (`routes/scripts.ts:403`), so callers *can* detect truncation; none of these four do.

bdunncompany:
Fixes #3301.

## The bug

`GET /scripts` is offset-paginated and defaults to 50 rows (`apps/api/src/routes/scripts.ts:39`). Four web call sites requested it with no `limit` and no page walk:

| Call site | Effect |
|---|---|
| `ScriptsPage.tsx:69` | the script library page shows the first 50 and has no pagination controls |
| `ScriptPickerModal.tsx:104` | script 51+ can't be picked to run on a device |
| `ScriptCategoryTree.tsx:181` | whole categories disappear from the tree |
| `RemediationScriptPicker.tsx:73` | a remediation script past 50 can't be attached to a policy |

There is no error, no "showing 50 of N", and no way to reach the rest. A partner with a large library just sees a smaller library.

## Why not raise the cap

The endpoint hard-caps `limit` at 100. Raising it moves the same silent truncation from 50 to 100 rather than removing it. The pages have to be walked.

## The fix

Adds `apps/web/src/lib/scriptsFetch.ts`, which walks `/scripts` to exhaustion and returns the full set as one array. Modelled on `fetchAllNetworkDevices` in `devicesFetch.ts`, which walks the same offset shape, and consistent with the already-proven paginated loop at `ScriptBundleImport.tsx:70`.

Three behaviours are deliberate:

- **Throws the failed `Response`** on the first non-OK page, matching `fetchAllDevices`, so callers keep status-based handling. Each call site's catch is updated for that shape. `ScriptsPage` and `ScriptCategoryTree` both redirect on 401, and an `instanceof Error` test alone would have silently dropped that logout.
- **A page error fails the whole walk** rather than returning what it has. A partial library rendered without comment is indistinguishable from a complete one, which is the bug being fixed.
- **The `MAX_PAGES` ceiling reports itself** via `onTruncated` and drops `total`, so the walker can't reintroduce a silent cap of its own at 5,000.

`fetchAllScripts<T>` is generic with a `Record<string, unknown>` default, so the two call sites that already have a row type pass it explicitly and the two that map from raw records are unchanged. Same unchecked-assertion tradeoff `asList<T>` documents.

## Verification

`scriptsFetch.test.ts` — 13 cases: the multi-page walk, short-page and empty-page terminators, request param shape, `includeSystem` propagation, the legacy `scripts` envelope key, malformed bodies failing closed, throw-on-non-OK at first and later pages, abort before and mid-walk, the truncation path, and a throwing `onTruncated` callback.

**Control run:** neutering the page-walk condition (`if (page.length < pageLimit)` → `if (true)`, i.e. the pre-fix single-request behaviour) fails 7 of the 13, including the regression case:

```
AssertionError: expected [ { id: 's0' }, { id: 's1' }, …(98) ] to have a length of 250 but got 100
```

That is the exact shape of the reported bug, so the tests are load-bearing rather than vacuous.

`ScriptsPage.test.tsx` matched the request URL exactly (`url === '/scripts'`); updated to the paginated form.

Local gate green on this branch rebased onto `f400fc315`:

- `vitest run` (apps/web) — **540 files, 5121 tests passed**
- `astro check` — **0 errors**, 0 warnings
- `eslint src` — clean

No dependency, API, agent, or migration changes, so those jobs are unaffected.

### DIFF
--- a/apps/web/src/components/configurationPolicies/featureTabs/RemediationScriptPicker.tsx
+++ b/apps/web/src/components/configurationPolicies/featureTabs/RemediationScriptPicker.tsx
@@ -1,10 +1,9 @@
 import { useState, useMemo, useEffect } from "react";
 import { X, Search, FileCode, Loader2 } from "lucide-react";
 import { cn } from "@/lib/utils";
-import { fetchWithAuth } from "../../../stores/auth";
 import { useTranslation } from "react-i18next";
 import { i18n } from "@/lib/i18n";
-import { asList } from '@/lib/asList';
+import { fetchAllScripts } from '@/lib/scriptsFetch';
 export type ScriptLanguage = "powershell" | "bash" | "python" | "cmd";
 export type OSType = "windows" | "macos" | "linux";
 export type SelectedScript = {
@@ -70,16 +69,9 @@ export default function RemediationScriptPicker({
     try {
       setLoading(true);
       setError(undefined);
-      const response = await fetchWithAuth("/scripts");
-      if (!response.ok) {
-        throw new Error(
-          i18n.t(
-            "policies:configurationPolicies.featureTabs.remediationScriptPicker.failedToFetchScripts",
-          ),
-        );
-      }
-      const data = await response.json();
-      const scriptList = asList(data, 'scripts');
+      // #3301 — walk every page; a remediation script past the first 50 could
+      // not be selected at all.
+      const { data: scriptList } = await fetchAllScripts();
       const transformedScripts: ScriptRow[] = scriptList.map(
         (s: Record<string, unknown>) => ({
           id: s.id as string,
@@ -92,6 +84,15 @@ export default function RemediationScriptPicker({
       );
       setScripts(transformedScripts);
     } catch (err) {
+      // fetchAllScripts throws the failed Response, not an Error.
+      if (err instanceof Response) {
+        setError(
+          i18n.t(
+            "policies:configurationPolicies.featureTabs.remediationScriptPicker.failedToFetchScripts",
+          ),
+        );
+        return;
+      }
       setError(err instanceof Error ? err.message : "Failed to load scripts");
     } finally {
       setLoading(false);
--- a/apps/web/src/components/devices/ScriptPickerModal.tsx
+++ b/apps/web/src/components/devices/ScriptPickerModal.tsx
@@ -3,11 +3,10 @@ import { X, Search, Play, Loader2, ChevronLeft } from 'lucide-react';
 import { useTranslation } from 'react-i18next';
 import { cn } from '@/lib/utils';
 import { Dialog } from '../shared/Dialog';
-import { fetchWithAuth } from '../../stores/auth';
 import { fetchLiveSessions, type LiveSession } from '../../services/deviceActions';
 import type { ScriptParameter } from '../scripts/ScriptFormSchema';
 import ScriptParametersForm, { validateParameters } from '../scripts/ScriptParametersForm';
-import { asList } from '@/lib/asList';
+import { fetchAllScripts } from '@/lib/scriptsFetch';
 
 export type ScriptLanguage = 'powershell' | 'bash' | 'python' | 'cmd';
 export type OSType = 'windows' | 'macos' | 'linux';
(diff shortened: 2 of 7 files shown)


ToddHebebrand:
Merged — the pagination walk itself is correct. Verified the terminator against the server contract: `fetchAllScripts` stops on `page.length < pageLimit`, and `routes/scripts.ts:37-40` confirms the server default of 50 / max 100, so the short-page signal is real and there's no off-by-one between the 1-indexed `page` param and the server's `(page-1)*limit` offset. A mid-walk failure throws the `Response` immediately, so no call site can render a partial list as complete.

**One follow-up worth doing, because it's the same failure mode this PR fixes:** `apps/web/src/lib/scriptsFetch.ts:120-127` exposes an `onTruncated` hook, but none of the four call sites wire it up. At the 5,000-script ceiling the truncation degrades to a `console.warn` — invisible to the user, which is exactly the silent cap that made #3301 hard to spot in the first place, just relocated further out. The sibling `fetchAllDevices` already has the pattern to copy: `DevicesPage.tsx` and `UnifiIntegration.tsx` both bind `onTruncated` to a visible warning.

**Also worth folding in while you're there:** `ScriptBundleImport.tsx:69` still has its own hand-rolled `for (page = 1; page <= 10; page++)` walk over `/scripts` that predates this PR. It isn't a regression, but there are now two divergent pagination implementations against the same endpoint with different ceilings (10 pages vs 50), and the older one is the more truncating of the two.

No blocker on either — `scriptsFetch.test.ts`'s 13 cases cover the regression, the short/empty-page terminators, abort, and the truncation path.

Issue #3300, opened 2026-08-09
86 of 217 registered AI tools are missing TOOL_TIERS entries and are invisible to chat


ToddHebebrand:
## Summary

86 of the 217 tools in the `aiTools` execution registry have no entry in `TOOL_TIERS` (`services/aiAgentSdkTools.ts`). Tools missing from that map are **invisible to the AI chat** — the agent will tell a user the capability does not exist.

This is pre-existing and not introduced by any recent PR; it surfaced during the review pass on #3278, which caught the two new fleet tools falling into the same hole before they shipped.

## Why a missing entry means "invisible"

`TOOL_TIERS` gates chat visibility twice over (documented in `aiGuardrails.readonly.contract.test.ts`):

- `BREEZE_MCP_TOOL_NAMES = Object.keys(TOOL_TIERS)` is the `allowedTools` list handed to the SDK.
- `createSessionPreToolUse` rejects `!TOOL_TIERS[toolName]` as "Unknown tool".

So a registered, working, fully-tested tool with no `TOOL_TIERS` entry cannot be called from chat at all. This is the #2605 drift class.

## Measured

```
registered tools: 217
TOOL_TIERS entries: 135
missing: 86
```

Whole feature families are affected, not a scattering of stragglers:

| Area | Examples |
|---|---|
| Backup (entire surface) | `query_backups`, `get_backup_status`, `trigger_backup`, `restore_snapshot`, `restore_as_vm`, `instant_boot_vm`, `manage_backup_profiles` |
| MSSQL | `query_mssql_instances`, `trigger_mssql_backup`, `restore_mssql_database`, `verify_mssql_backup` |
| Hyper-V | `query_hyperv_vms`, `manage_hyperv_vm`, `trigger_hyperv_backup`, `restore_hyperv_vm`, `manage_hyperv_checkpoints` |
| Vaults / C2C | `query_vaults`, `trigger_vault_sync`, `query_c2c_connections`, `restore_c2c_items` |
| Backup SLA / DR | `query_backup_sla`, `get_sla_breaches`, `query_dr_plans`, `execute_dr_plan`, `manage_dr_plan` |
| Incident response | `create_incident`, `execute_containment`, `collect_evidence`, `get_incident_timeline`, `generate_incident_report` |
| PAM / elevation | `request_elevation`, `revoke_elevation`, `get_elevation_history` |
| Network | `get_network_changes`, `acknowledge_network_device`, `configure_network_baseline`, `get_ip_history` |
[...]

Issue #2605, opened 2026-07-17
[AI] Vulnerability tools are never selected — model routes CVE questions to posture/patch tools instead


ToddHebebrand:
## Description
With vulnerability findings seeded (real CVEs in `vulnerabilities` + `device_vulnerabilities`), the org-scoped AI chat never chooses `get_vulnerability_report` / `get_device_vulnerabilities`, even when the prompt uses the tools' own vocabulary:

- "Do any of our devices have known vulnerable software? Give me the highest-risk issues."
- "Check our vulnerability findings: which open CVEs do we have across the fleet?"
- "Give me a vulnerability report: our open CVE findings, counts by severity..."

All three routed to `get_security_posture`, `manage_patches`, `manage_alerts`, `get_s1_threats`, `get_huntress_incidents` (and once `get_cis_compliance`, see #2603) — so the answers described patch KBs and posture scores while the actual CVE findings (including known-exploited entries) were never surfaced.

Likely a tool-selection problem at ~100 registered tools: the vuln tools' descriptions don't win against the louder security tools.

## Proposed Fix
- Tune `get_vulnerability_report`/`get_device_vulnerabilities` descriptions to claim the CVE/vulnerability-findings vocabulary explicitly (and cross-reference them from `get_security_posture`'s description).
- Consider a system-prompt hint mapping common intents to canonical tools, or evaluating tool-choice quality as part of the flagged-chat review loop.

## Affected Files
- `apps/api/src/services/aiToolsVulnerability.ts` (descriptions)
- `apps/api/src/services/aiAgent.ts` (system prompt)

PR #3306, opened 2026-08-09
test(ai): pin the aiTools registry against TOOL_TIERS, both directions (#3300)

bdunncompany:
Part of #3300 — **the guard only**, not the tier assignments.

Per the issue's own sequencing ("Step 2 is arguably worth doing first"), this adds the contract test and leaves all 86 tier decisions alone. A wrong tier on `execute_containment` or `revoke_elevation` is worse than the tool being invisible, so those want the deliberate per-tool review the issue describes.

## What drifted

`TOOL_TIERS` gates chat visibility twice — `BREEZE_MCP_TOOL_NAMES = Object.keys(TOOL_TIERS)` is the `allowedTools` list handed to the SDK, and `createSessionPreToolUse` rejects `!TOOL_TIERS[toolName]` as "Unknown tool". Nothing asserted the two maps agree, and they drifted **both** ways:

| Direction | Count | Effect |
|---|---|---|
| Registered, untiered | 86 | a working, tested tool that chat says does not exist |
| Tiered, unregistered | 4 | a name advertised as callable that fails at execution |

The second direction is not in the issue. The four are `get_backup_health`, `get_recovery_readiness`, `propose_action_plan`, `run_backup_verification`. Same #2605 drift class, so the guard is bidirectional.

Measured on `f400fc315`: **215 registered, 133 tiered, 86 missing, 4 dead.** (215/133 rather than the issue's 217/135 because #3291 merged in between; the missing count is 86 either way.)

## What this adds

One test file. Three contracts:

1. **registry ⊆ TOOL_TIERS** — minus `KNOWN_MISSING_TOOL_TIERS`, the 86 frozen.
2. **TOOL_TIERS ⊆ registry** — minus `KNOWN_UNREGISTERED_TOOL_TIERS`, the 4 frozen.
3. **shared names carry the same tier in both maps** — no allowlist, because this held for all 133.

Contract 3 is the one worth keeping past the cleanup. Presence alone is the weaker claim: an entry that is *present but lower* than the registry's tier has had its approval gate quietly downgraded, which is worse than invisibility. It also means the eventual step-1 work has a defensible default — the registry's own tier — rather than a blank sheet.

**Both allowlists may only shrink.** Each block has a second test that fails on a stale entry, so deleting a name is what un-blocks CI once a tier lands, and re-adding one cannot pass. The lists walk to empty on their own.

## Controls

Every assertion was proven load-bearing by breaking it:

| Perturbation | Result |
|---|---|
| drop a real gap from `KNOWN_MISSING` | registry-subset test fails |
| add a tiered tool to `KNOWN_MISSING` | shrink-only test fails |
| drop a dead entry from `KNOWN_UNREGISTERED` | tiers-subset test fails |
| add a live tool to `KNOWN_UNREGISTERED` | shrink-only test fails |
| `execute_command: 3` → `1` in `TOOL_TIERS` | agreement test fails: `execute_command: TOOL_TIERS=1 registry=3` |

The last row is the one that matters — a silent approval-gate downgrade is caught.

## Verification

Test-only. No source, dependency, or behaviour changes, so the security and dependency jobs are unaffected.

- `vitest run` (apps/api) — **1298 files, 20766 tests passed**, 5 files / 40 tests skipped
- `tsc --noEmit --project apps/api/tsconfig.json` — **0 errors**
- `eslint` on the new file — clean

### DIFF
--- a/apps/api/src/services/aiAgentSdkTools.registryParity.contract.test.ts
+++ b/apps/api/src/services/aiAgentSdkTools.registryParity.contract.test.ts
@@ -1,0 +1,236 @@
+/**
+ * Contract test pinning the `aiTools` registry against `TOOL_TIERS` (#3300).
+ *
+ * `TOOL_TIERS` gates chat visibility twice over:
+ *
+ *   - `BREEZE_MCP_TOOL_NAMES = Object.keys(TOOL_TIERS)` is the `allowedTools`
+ *     list handed to the SDK.
+ *   - `createSessionPreToolUse` rejects `!TOOL_TIERS[toolName]` as
+ *     "Unknown tool".
+ *
+ * So the two maps have to agree, and they drift in BOTH directions:
+ *
+ *   - Registered but untiered → a working, fully-tested tool that chat says
+ *     does not exist. 86 tools are in this state today.
+ *   - Tiered but unregistered → a name advertised to the SDK as callable that
+ *     then fails at execution. 4 entries are in this state today.
+ *
+ * This is the #2605 drift class. Nothing failed CI when it happened, which is
+ * why it reached 86. This suite is the durable half of #3300: it does not fix
+ * the existing gap (assigning a tier is an approval-gate decision, made
+ * deliberately per tool — see the issue), it stops the gap GROWING. The two
+ * allowlists below are frozen snapshots that may only ever shrink, and the
+ * last test in each block enforces that by failing on a stale entry.
+ *
+ * NOTE: no vi.mock — this suite needs the REAL registry, same rationale as
+ * aiGuardrails.readonly.contract.test.ts.
+ */
+import { describe, expect, it } from 'vitest';
+
+import { TOOL_TIERS } from './aiAgentSdkTools';
+import { getAllRegisteredToolNames, getToolTier } from './aiTools';
+
+/**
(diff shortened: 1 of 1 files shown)


bdunncompany:
Rebased onto `2cb6f7a6b` (post #3308 + #3309) and re-verified — no content change, just moving off a stale base.

This PR freezes exact global counts (215 registered / 133 tiered / 86 missing / 4 dead), so a merge that adds, removes, or re-tiers a tool would break it *after* it lands rather than on the PR. #3308 touched `services/aiToolsIntegrations.ts`, so that was worth checking rather than assuming: the change is confined to the query inside the existing `query_psa_status` handler (dual-axis condition + `ownerScope` in the result) with no registration or `TOOL_TIERS` edit.

Re-ran the suite against the rebased tree: **5/5 pass**, both allowlists still exact.

bdunncompany:
The `Test Web` red here is not from this PR — this diff is a single **API** contract test file and touches nothing under `apps/web`.

It's the known InvoiceWorkspace flake (#3219 / #3277): `holds a header Issue click while the editor is dirty, then fires it when the save settles`, 1 failed / 538 passed. Fourth sighting in eleven days; #3284 is the open fix. I've rerun the job.

Issue #2195, opened 2026-07-03
Org SSO login is broken under RLS on public routes (initiation + check + identity read) — completes the #2194 callback fixes


ToddHebebrand:
**While building partner SSO (#2183 / PR #2194), the new real-DB e2e suite exposed that org SSO has been silently non-functional in production-shaped deployments** — and PR #2194 fixes only the callback half. This issue tracks the remainder plus the hardening items the branch review triaged here.

## Root cause class

Public (pre-auth) SSO routes read the DB through bare `db` calls. Production runs the API as the unprivileged `breeze_app` role with RLS **enabled + forced** on `sso_providers`; with no request context set, every policy evaluates false and reads return **0 rows silently**. Unit tests mock the DB and never see it — only the real-Postgres integration suite added in PR #2194 caught it.

## Fixed in PR #2194 (callback path, both axes)

Three reads in `GET /sso/callback` wrapped in `withSystemDbAccessContext` (matching the neighboring reads): the provider lookup (universally fatal — every callback 302'd to `provider_not_found`), the org-membership check, and the defaultRoleId validation read. Regression-locked by full-success e2e cases for both axes.

## Still broken (this issue)

- **`GET /sso/login/:orgId`** (initiation): bare provider read → org SSO login can never even start under `breeze_app`. `routes/sso.ts` (~904 area).
- **`GET /sso/check/:orgId`**: bare read → always returns `ssoEnabled: false`, so the org login page never shows the SSO button.
- **Callback `existingIdentity` select**: bare read → identity lookup can 0-row and fall through to the email-match path, risking duplicate `user_sso_identities` inserts.

Together with the #2194 fixes, wrapping these three completes the org SSO path. Each is the same one-line `withSystemDbAccessContext` class.

## Hardening items (same surface, triaged non-blocking by the #2194 branch review)

- `user_sso_identities` has no unique index on `(provider_id, external_id)` — the identity-in-use check is code-only (TOCTOU). Add the unique index + handle 23505.
- Tenant-cascade sweeps don't clear `sso_sessions` / `user_sso_identities` before `sso_providers` — a canary partner/org that exercised SSO will fail purge.
- Rate limiting: org `/sso/login/:orgId` has none; partner entry uses a single ip+partnerId bucket (no pure-IP bucket like password login).
[...]

Issue #2135, opened 2026-07-01
Epic: partner-wide first — make config template libraries partner-ownable (dual-axis)


ToddHebebrand:
## Problem

Breeze is an RMM: techs operate across many customer orgs at once, so **partner-wide ("all organizations") configuration is the primary model** — one policy, all devices. Today it is structurally second-class: `configuration_policies` supports dual ownership (org_id XOR partner_id, #1724), but **8 of the 9 linkable feature-policy tables are org-locked** (`org_id NOT NULL`, no partner axis). Only `patch_policies` (update rings) is partner-axis.

The practical consequence: the link-required feature types — **software_policy, security, compliance, peripheral_control, backup** — are entirely unusable on a partner-wide config policy, because there is nothing partner-owned they can link to (`validateFeaturePolicyExists` rejects org-scoped links on partner-wide policies, `apps/api/src/services/configurationPolicy.ts`). A tech who wants one software blocklist for 40 customers must create and maintain 40 copies.

**Audit (2026-07-01)** of ownership axes:

Org-locked (org_id NOT NULL, no partner_id): `alert_rules`, `software_policies`, `security_policies`, `automation_policies`, `maintenance_windows`, `sensitive_data_policies`, `peripheral_policies`, `backup_configs`, plus `automations`, `notification_channels` / routing rules / escalation policies.

Already partner-capable (proof the pattern works): `patch_policies` (partner-axis), `configuration_policies` (dual), `scripts` / categories / tags (dual), `alert_templates`, `client_ai_prompt_templates`, `custom_field_definitions`, `software_catalog`. Breeze Defaults are code-level globals.

## Approach

Migrate the template-library tables to **dual ownership (org_id XOR partner_id)** using the established playbook from `apps/api/migrations/2026-06-27-config-policies-partner-ownership.sql`:

- `ADD COLUMN partner_id`, relax `org_id`, XOR CHECK constraint, partner index
- Dual-axis RLS (`breeze_has_org_access(org_id) OR breeze_has_partner_access(partner_id)`) + FORCE, extended to FK children
- Partner derived from the caller's token, never the request body
- Partner-wide writes gated on `canManagePartnerWidePolicies(auth)` (`apps/api/src/services/configurationPolicy.ts`) — orgAccess='all' or system scope
- Org-owned config policies may link partner-owned templates (like update rings today); partner-wide config policies may link partner-owned templates only
[...]

Issue #2655, opened 2026-07-20
[API] platform-admin bootstrap UPDATE with = ANY(...::text[]) fails at runtime in the prod CJS bundle


ToddHebebrand:
## Description
Found during the Plan 05 stock-image conformance bring-up (2026-07-19). With `BREEZE_PLATFORM_ADMINS` set, the startup bootstrap failed on the stock `docker/Dockerfile.api` image (prod CJS bundle), so the configured user was never promoted. Boot continues because `index.ts` wraps `bootstrapPlatformAdmins()` in a non-fatal try/catch, so the only signal is the `[startup] Platform admin bootstrap failed (non-fatal)` log line. Workaround used: manual promotion via psql (`UPDATE users SET is_platform_admin = true WHERE lower(email) = ...`).

## Root Cause (suspected)
`apps/api/src/services/platformAdminBootstrap.ts:27`:

```ts
sql`lower(${users.email}) = ANY(${emails}::text[]) AND ${users.isPlatformAdmin} = false`
```

`emails` is a JS `string[]` interpolated as a single drizzle param and cast to `text[]`. In the dev/vitest path this works, but in the production CJS bundle the parameter reaches postgres in a shape the `::text[]` cast rejects (the same class of bundle-only drizzle param-serialization difference we have hit before — unit tests with a mocked `db.execute` cannot catch it, as noted in `loader.test.ts`'s own comments about `= ANY(tuple)`).

## Proposed Fix
Build the array with drizzle's helper instead of a string-cast interpolation, e.g. `inArray(sql`lower(${users.email})`, emails)` or `= ANY(${sql.param(emails)})` with an explicit array param — and cover it with a real-Postgres integration test (the existing `platformAdminBootstrap.test.ts` mocks the db, which is exactly why this survived to a live image).

## Affected Files
- `apps/api/src/services/platformAdminBootstrap.ts` (primary)
- `apps/api/src/services/platformAdminBootstrap.test.ts` (needs a live-Postgres companion)

## Reported By
Internal — Plan 05 runtime-extension conformance estate run (workaround: psql promotion).

ToddHebebrand:
Captured the exact failure live on a fresh stock-image boot (branch `refactor/runtime-ext-06-deprecate-buildtime` estate, but the code path is unchanged from main):

```
[...]

PR #3308, opened 2026-08-09
feat(psa): dual ownership (org XOR partner) for psa_connections (#2135)

ToddHebebrand:
Retrofits dual ownership (`org_id` XOR `partner_id`) onto `psa_connections`, following the **Partner-Wide First** playbook (epic #2135).

---

## Review round 2 — 10 findings applied

A high-effort review found a **real cross-org data leak** plus 9 more. All fixed.

### The conceptual error (findings 0, 1, 8)

The PR treated **partner scope as sufficient authorization for partner-wide rows**. It is not — two different things were conflated:

- `psa_connections` is partner-owned **CONFIG** → correctly visible to any partner-scope caller of that partner.
- The `psa_ticket_mappings` hanging off it are per-org **DATA** — each row names a specific org's device, alert and external ticket → must **always** be filtered by the caller's accessible orgs, whatever the scope.

**The leak:** `psaConnectionAccessCondition` OR'd in the partner arm for any partner-scope token without consulting `partnerOrgAccess`, so a `partner_user` with `org_access='selected'` limited to org A read **org B's external ticket ids, URLs, alert ids and device ids** via `GET /psa/tickets`. `GET /connections/:id/tickets` was worse — after `ensureConnectionAccess` passed it applied **no org filter at all**.

**Fix:** a new `psaTicketMappingOrgCondition` constrains the **mapping rows themselves** through their own org anchors (`device_id → devices.org_id`, `alert_id → alerts.org_id`), AND-combined so a row is withheld if *either* anchor is out of reach. A row with neither anchor references no org, so connection access governs it.

**This required dropping the `psa_connections` innerJoin on `/tickets`** — it could not coexist with correct org-scope behaviour. `psa_connections` RLS rightly hides partner-wide rows from org tokens, so the join silently discarded every ticket about an org's *own* device whenever the MSP's PSA was partner-wide. That is finding 8, the same bug's mirror image: with the join gone and the anchor filter in place, an org-scope caller now correctly sees their own device's ticket. The `auth.scope === 'partner'` gate is dropped on the ticket path and **kept** on the connection-visibility path, where it is correct.

**RLS cannot enforce this.** The `psa_ticket_mappings` connection arm passes for any partner-scope caller of the owning partner — necessarily, since an unanchored row has no org for Postgres to check. Postgres cannot express `org_access='selected'`, so this app-layer condition is the enforcement point, and it is documented as such so it is not later deleted as "redundant with the policy".

### Finding 2 — asymmetric USING / WITH CHECK

The device arm was a bare `OR` in **both** clauses, so owning a device authorized *writing* a mapping onto a foreign tenant's connection — contradicting the migration's own "connection_id is the tenancy anchor" rule. Now split:

- `USING` = system OR connection-accessible OR device-org OR **alert-org**
- `WITH CHECK` = system OR connection-accessible **only**

This closes the foreign-connection write hole and the missing-alert-arm hole together. Applied by **amending the existing unmerged migration** rather than adding a follow-up: the file has never been applied to a persistent database (only ephemeral CI containers, torn down each run), `breeze_migrations` keys on filename so it still applies exactly once everywhere, and shipping a knowingly-wrong policy into history to immediately supersede it would gain nothing. The rule it must not break — never edit a **shipped** migration — is not engaged.

### Finding 3 — VERDICT: CONFIRMED REAL

Traced end to end. `deleteDeviceCascade` pre-clears on `alert_id IN (device's alerts) OR device_id = ...` inside the **request** (org-scoped) context. For a mapping with `device_id` NULL and `alert_id` set under a **partner-owned** connection: the connection arm fails (org token cannot see partner-wide connections), the device arm fails (`device_id` NULL) → the DELETE matches **zero rows** → the cascade's follow-on `DELETE FROM alerts` raises **23503** → `DELETE /devices/:id` 500s. Fixed by the `USING` alert arm; the regression test deletes exactly such a mapping under org-scoped RLS.

### Findings 4–7, 9

| # | Fix |
|---|---|
| 4 | Create form takes its default from the canonical `useDefaultOwnerScope` hook instead of a hard-coded `'partner'`. Partner-wide remains the default in the All-organizations view. |
| 5 | `GET /connections/:id` computes `canManage` as orgs:write **AND NOT** `partnerWideWriteBlocked`, so a user barred from mutating a partner-wide connection no longer receives the credential prefill or `credentialFields` — the material #3291 gated. |
| 6 | "This organization only" sent no `orgId`, so create **400'd for any partner with 2+ orgs**. Added an org picker on that branch, populated from the caller's accessible orgs. |
| 7 | Edit/Pause/Delete are disabled with an explanatory tooltip on partner-wide rows the user cannot manage, instead of letting the click land on the new 403 as an error toast. |
| 9 | Deleted the dead org-only `resolveOrgIds`, so the next endpoint author cannot copy the partner-wide-invisible shape. |

### A bug CI caught that the local tests structurally could not

The first cut used `d.org_id = ANY(${orgIds}::uuid[])`. Drizzle expands a JS array in a `sql` template into a comma-separated **parameter list**, so Postgres received a record: `cannot cast type record to uuid[]` (and `malformed array literal` for a single-element list). `routes/psa.test.ts` mocks the entire db layer, so malformed SQL passes there and fails only against a real database.

Fixed by binding each id via `sql.join`, and — to close the gap rather than just the bug — the condition moved into a dependency-free `services/psa/ticketScope.ts` with `ticketScope.test.ts`, which renders it through the real `PgDialect` and asserts the emitted SQL and parameter binding. That moves this class of failure into the fast unit job.

### Round-2 verification (all green, real Postgres)

| Suite | Where | Result |
|---|---|---|
| `psaTicketMappingOrgScoping.integration.test.ts` (**new**) | shard 2 | ✓ 6 — the leak guards + finding 8 |
| `psaConnectionsPartnerRls.integration.test.ts` | shard 2 | ✓ 13 (was 11; +alert arm, +WITH CHECK forge) |
| `rls-coverage` / `tenant-export-policy` / `tenantExportErasureRoundtrip` / `tenantCascadePartner` | shard 1 | ✓ 75 / 2 / 2 / 3 |
| `tenantCascade` / `tenantCascadeExecution` / `tenantCascadeSso` | shard 4 | ✓ 5 / 3 / 4 |
| `psa.test.ts` + `ticketScope.test.ts` | local | ✓ 50 |
| web `PsaConnectionsPage.test.tsx` | local | ✓ 16 |
| i18n parity / coverage / keyUsage + `no-silent-mutations` | local | ✓ 164 |

**56/56 CI checks pass.** The `WITH CHECK` forge check runs as `breeze_app` through the real postgres.js driver and asserts `42501` — that is the manual forge check, executed as a test because this environment has no local database (no `docker`/`psql`/`podman`).

---

## Why partner-wide

An MSP's PSA is a **partner-level** system: one ConnectWise/Autotask/Jira tenant covers every customer the MSP manages. Before this, `psa_connections.org_id` was `NOT NULL`, so an MSP had to re-enter the same PSA credentials once per customer org.

Corroborating signals: the sibling `accounting_connections` table is already partner-axis, the real UI has no org picker, and the org-import feature that will consume this is partner-scoped. The `backup_configs` org-only precedent does **not** apply — that table holds the *customer's* storage credentials, whereas a PSA credential is the *MSP's own*.

Dual ownership rather than partner-only preserves existing org-owned rows and the legitimate co-managed-IT case (a customer bringing their own Jira/Zendesk).

## Migration — `2026-08-17-psa-connections-partner-ownership.sql`

Date-ordering checked with `ls apps/api/migrations/ | tail -5`: the latest shipped file is `2026-08-15-snmp-poll-attempt-backoff.sql`, so `2026-08-17-` sorts last. **Not** added to the closed `2026-08-06` block. `scripts/check-migration-naming.sh` passes (pre-commit hook).

- `partner_id uuid REFERENCES partners(id)`, `org_id` `DROP NOT NULL`
- `psa_connections_one_owner_chk CHECK ((org_id IS NULL) <> (partner_id IS NULL))`
- `psa_connections_partner_id_idx`
- ONE dual-axis policy replacing the four `breeze_org_isolation_*` policies from the generic `0008-tenant-rls.sql` sweep:

```sql
CREATE POLICY psa_connections_isolation
  ON psa_connections
  USING (
    public.breeze_current_scope() = 'system'
    OR (org_id IS NOT NULL AND public.breeze_has_org_access(org_id))
    OR (partner_id IS NOT NULL AND public.breeze_has_partner_access(partner_id))
  )
  WITH CHECK ( ...same... );
```

No data cleanup block: `org_id` is `NOT NULL` today, so every existing row satisfies the XOR the moment the constraint lands. Nothing is UPDATEd or DELETEd, hence no `ROW_COUNT` diagnostics.

### `psa_ticket_mappings` join policies — rewritten in the SAME migration

Its four org-only join policies (from `2026-04-11-bucket-c-dead-cleanup-rls.sql`) asserted `breeze_has_org_access(pc.org_id)`, which is FALSE whenever `pc.org_id IS NULL` — i.e. for *every* partner-owned connection. Left alone, mappings under a partner-owned connection would be both invisible AND unwritable. Collapsed to one policy with **three** arms (same hazard `ticket_form_org_links` documents):

1. system short-circuit
2. join to `psa_connections` with the parent's **dual-axis** predicate
3. **`device_id → devices.org_id`**

Arm 3 is not cosmetic. `deleteDeviceCascade` runs in the **request** context, which for an org-scope token is org-scoped RLS. Without it, a mapping hanging off a partner-owned connection is invisible to that token, its `DELETE` silently matches zero rows, and the following `DELETE FROM devices` fails `23503` — a 500 on an ordinary device delete, newly introduced by this PR. The arm is also right on the merits: a ticket about *my* device is *my* org's record, which is exactly what playbook step 5 says for worker-created child rows.

## Erasure / cascade fix (confirmed gap)

`psa_ticket_mappings` has **no** `org_id` and **no** `partner_id`, yet holds **three** FKs into the cascade set, every one declared without an explicit `ON DELETE` (so `NO ACTION`): `connection_id → psa_connections` (NOT NULL), `alert_id → alerts`, `device_id → devices`.

**Chosen fix: pre-clears in both cascades**, following the existing `ASSOCIATED_SYSTEM_SCOPED_TABLES` / partner pre-clear pattern (#2195). Rejected `ON DELETE CASCADE` because:

- the per-table row counts are **load-bearing**, not cosmetic — `cascadeDeletePartner` reports them so a purge that silently matched zero rows under forced RLS surfaces as `totalRowsDeleted === 0` instead of masquerading as success. DB-level cascades are invisible to `extractRowCount`.
- repo convention is explicit enumeration **even when the FK already cascades** (`organization_external_links`, `ticket_form_org_links`).
- a uniform `ON DELETE CASCADE` on the nullable `alert_id`/`device_id` would change **non-erasure** behaviour (an alert-retention purge would start destroying ticket mappings); `SET NULL` there would contradict `deleteDeviceCascade`, which deletes them by `device_id`. The pre-clear changes nothing outside erasure.

The **org** pre-clear needs all three arms. With `org_id` nullable, a mapping under a *partner-owned* connection matches no `connection_id IN (… WHERE org_id = $org)` subquery, yet still pins that org's `alerts` and `devices` rows — both of which the main loop deletes. Connection-only would be a guaranteed `23503`.

**This was already latently broken before this PR** (`alerts`/`devices` are deleted by the org cascade regardless of connection ownership) — and **no contract test in the repo can catch it**: `tenantCascade.integration.test.ts` enumerates membership from `information_schema.columns WHERE column_name='org_id'` and skips FK edges whose endpoints are outside the cascade set, so a table with neither column is structurally invisible to it. That is exactly how #2195 shipped. It is **dormant**, though — there is no INSERT path for `psa_ticket_mappings` anywhere in the repo (no PSA sync worker exists), so the table is empty in every deployment.

Also fixed: **`deleteDeviceCascade` deletes `alerts` ~30 entries *before* `psa_ticket_mappings`** in `CORE_DEVICE_CASCADE_DELETE_TABLES`, so a mapping referencing that device's alert raises `23503` on `DELETE /devices/:id`. Pre-existing, unrelated to GDPR erasure, fixed here since this PR touches the same table.

## Reader sweep (playbook step 7)

Every `psa_connections.orgId` call site, swept repo-wide:

| Site | Fix |
|---|---|
| `routes/psa.ts` `GET /connections` (`inArray(orgId, orgIds)`) + its count query | dual-axis `psaConnectionAccessCondition` |
| `routes/psa.ts` `GET /tickets` + its count query (through the join) | dual-axis condition on the joined parent |
| `routes/psa.ts` `resolveOrgIds` | no longer used for scoping the two list routes |
| `routes/psa.ts` × 6 `ensureOrgAccess(connection.orgId, …)` (detail, PATCH, DELETE, test, status, `:id/tickets`) | `ensureConnectionAccess` (dual-axis) |
| `routes/psa.ts` `serializeConnection` | `ownerScope` added, `orgId` now nullable |
| `services/aiToolsIntegrations.ts` `query_psa_status` (`auth.orgCondition`) | dual-axis condition + `ownerScope` in the tool result |
| `apps/web` `PsaConnectionList` / `PsaConnectionForm` / `PsaConnectionsPage` | `ownerScope` type, badge, selector |

Confirmed **not** affected: `encryptedColumnRegistry` and `services/psa/credentials.ts` (keyed on table+column, no tenant axis); `routes/devices/core.ts` (`psa_ticket_mappings` deletes by `device_id`, and its exclusion from the org-denormalized set stays correct); `routes/integrations.ts` `/integrations/psa` (an unrelated in-memory Map). No hits in portal, mobile, helper, agent, viewer, or any worker.

## Writes

- create takes `ownerScope: 'organization' | 'partner'`; **absent ⇒ `organization`**, preserving behaviour for existing API clients. The **web form** defaults its selector to partner-wide (per the product decision) — the wire default stays org-scoped so an unaware caller can never accidentally publish a credential to every org.
- partner-wide create/update/delete/**status**/**test** gated on `canManagePartnerWidePolicies(auth)`. Status is an update by another name (pausing a partner-wide connection pauses it everywhere), and test both writes `syncSettings` and exercises the MSP's credentials — so both are held to the same bar.
- the partner is **always** derived from the caller's token, never the body (covered by a test).
- ownership is **immutable** after create: `updateConnectionSchema` is hand-written and has no `ownerScope`, so there is nothing to `.omit({ ownerScope: true })` from — noted in a comment at the schema.
- app-layer partner conditions are gated on `auth.scope === 'partner'`. An org token carries a `partnerId` but never passes `breeze_has_partner_access`; **RLS is stricter than the app layer and the two are not at parity.**

## Registration (separate contracts)

- `DUAL_AXIS_TENANT_TABLES` ← `psa_connections`
- `PARENT_FK_JOIN_POLICY_TABLES` entry for `psa_ticket_mappings` re-commented for the dual-axis parent
- `CORE_TENANT_EXPORT_POLICY` ← `partner_id` in `psa_connections.included` (a tenant identifier). **This is the row that fires on a new column, not just a new table.** The org export only ever emits org-owned rows, so a partner-owned connection never appears in a customer's export — correct, it is the MSP's credential.
- org cascade entry verified: `DELETE FROM psa_connections WHERE org_id = X` correctly leaves partner-owned rows alone; they are reached by the partner-axis sweep, which `psa_connections` now joins automatically via its new `partner_id` column.

## Playbook steps 4 and 5 — N/A, considered not forgotten

- **Step 4 (config-policy linkage):** N/A. `psa_connections` is not a config-policy feature — it is not in `FEATURE_TABLE_MAP`, nothing links it via `config_policy_feature_links`, and `PARTNER_LINKABLE_FEATURE_TYPES` has no PSA member.
- **Step 5 (evaluation/enforcement fan-out):** N/A. No worker or scheduler evaluates `psa_connections` against devices — nothing consumes `services/psa/index.ts` on a schedule, and `POST /connections/:id/sync` is an honest 501. Consequently the new RLS suite has no fan-out test.

## Tests

- **new** `psaConnectionsPartnerRls.integration.test.ts` — cross-partner forge → `42501`, XOR (both axes / neither) → `23514`, org isolation, cross-org isolation, the **partner-scope read proof**, and the `psa_ticket_mappings` device arm.
- `psa.test.ts` — 14 new cases (ownerScope create incl. body-`orgId` ignored, partner-wide gating on update/delete/status, dual-axis detail access, org-scope denial).
- `tenantCascade.partner.test.ts` — asserts the `psa_ticket_mappings` pre-clear runs before the partner-axis sweep.
- `PsaConnectionsPage.test.tsx` — 5 new cases (badge, selector visibility by scope + capability, both ownerScope POST paths).
- i18n strings added to all seven locales; `localeParity` / `translationCoverage` / `keyUsage` pass.

## Contract suites — verified in CI against real Postgres

They could **not** be run locally: this authoring environment has no `docker`, `psql`, or `podman` binary, no `/var/run/docker.sock`, and nothing listening on 5432/5433/54320, so the manual `breeze_app` forge check was done as an automated test instead of by hand. All four blocking **Integration Tests** shards are green and every relevant suite is confirmed executed, not skipped:

| Suite | Shard | Result |
|---|---|---|
| `psaConnectionsPartnerRls.integration.test.ts` (**new**) | 2 | ✓ 11 tests |
| `rls-coverage.integration.test.ts` | 1 | ✓ 75 tests |
| `tenant-export-policy.integration.test.ts` | 1 | ✓ 2 tests |
| `tenantExportErasureRoundtrip.integration.test.ts` | 1 | ✓ 2 tests |
| `tenantCascade.integration.test.ts` | 4 | ✓ 5 tests |
| `tenantCascadePartner.integration.test.ts` | 1 | ✓ 3 tests |
| `tenantCascadeExecution` / `tenantCascadeSso` | 4 | ✓ 3 / ✓ 4 tests |

The cross-tenant forge assertions (`42501`) and the XOR CHECK assertions (`23514`) in the new suite ARE the `breeze_app` manual check, run as `breeze_app` through the real postgres.js driver.

Also run locally: full API unit suite **20,774 passed / 1,296 files** (the single failing file, `requestDatabaseCompose.test.ts`, is `spawnSync docker ENOENT` — environmental, it shells out to `docker compose`); both `tsc --noEmit` clean.

## Deviations

- **Advisor quorum ran on Opus, not Codex.** `codex` is not installed in this environment; the CLAUDE.md fallback is an adversarial subagent. The first attempt (Fable) hit a credit limit, so the review ran on Opus. It confirmed the pre-clear choice, corrected my weaker justification for it (the "only 1 of 3 FKs" argument — replaced above with row-count accounting + repo convention), and independently found both the `deleteDeviceCascade` ordering bug and the org-token RLS hazard, which are fixed here.
- The reviewer's stronger alternative — denormalize `org_id` onto `psa_ticket_mappings` so it joins the auto-discovered shape-1 set — was **considered and rejected**: `device_id` and `alert_id` are both nullable, so a mapping under a partner-wide connection that is neither device- nor alert-linked has no org to derive, making `org_id NOT NULL` underivable in general. Worth revisiting when the PSA sync backend lands and the real child-row shape is known.
- `POST /connections/:id/test` is gated on the partner-wide capability even though the playbook names only create/update/delete — rationale above.

### DIFF
--- a/apps/api/migrations/2026-08-17-psa-connections-partner-ownership.sql
+++ b/apps/api/migrations/2026-08-17-psa-connections-partner-ownership.sql
@@ -100,19 +100,41 @@ CREATE POLICY psa_connections_isolation
 -- pc.org_id IS NULL — i.e. for every partner-owned connection. Collapsed from
 -- four per-command policies to one, matching the parent above.
 --
--- THREE arms, not two. The third (device_id -> devices.org_id) is not
--- cosmetic: deleteDeviceCascade (services/deviceDeletion.ts) runs in the
--- REQUEST context, which for an org-scope token is org-scoped RLS. Without the
--- device arm, a mapping hanging off a PARTNER-owned connection is invisible to
--- that token, its DELETE silently matches zero rows, and the subsequent
--- `DELETE FROM devices` fails with 23503 — a 500 on an ordinary device delete.
--- The arm is also correct on the merits: a ticket about MY device is MY org's
--- record, which is the same principle playbook step 5 states for every
--- worker-created child row (they take the DEVICE's org, not the policy's).
+-- USING and WITH CHECK are deliberately DIFFERENT, because reading and writing
+-- make different tenancy claims:
 --
--- connection_id is NOT NULL and remains the primary tenancy anchor; device_id
--- and alert_id are nullable, so they widen access but can never be the sole
--- basis for it.
+--   USING      = system OR connection-accessible OR device-org OR alert-org
+--   WITH CHECK = system OR connection-accessible ONLY
+--
+-- Reading or deleting a row that references YOUR OWN device or alert is a
+-- legitimate claim — that ticket is your org's record, the same principle
+-- playbook step 5 states for every worker-created child row (they take the
+-- DEVICE's org, not the policy's). But ATTACHING a mapping to a connection is a
+-- claim on the CONNECTION, so the write path must not accept device/alert
+-- ownership as a substitute. A symmetric policy would have let a tenant who
+-- merely owns a device forge a mapping onto a FOREIGN partner's connection,
+-- contradicting this file's own "connection_id is the tenancy anchor" rule.
+--
+-- Both read arms are load-bearing for deleteDeviceCascade
+-- (services/deviceDeletion.ts), which runs in the REQUEST context — org-scoped
+-- RLS for an org token. Its pre-clear matches on `alert_id IN (...) OR
+-- device_id = ...`, so under a PARTNER-owned connection:
+--   * device_id set   -> needs the device arm
+--   * device_id NULL, alert_id set -> needs the ALERT arm
+-- Without the relevant arm the DELETE silently matches zero rows and the
+-- cascade's follow-on `DELETE FROM alerts` / `DELETE FROM devices` fails with
+-- 23503 — a 500 on an ordinary device delete.
+--
+-- connection_id is NOT NULL and remains the primary tenancy anchor for WRITES;
+-- device_id and alert_id are nullable and widen READS only.
+--
+-- NOTE ON ROW VISIBILITY vs AUTHORIZATION: the connection arm passes for ANY
+-- partner-scope caller of the owning partner, necessarily so — a mapping with
+-- neither anchor has no org for Postgres to check. Postgres therefore cannot
+-- express partner_users.org_access='selected' here. Narrowing ticket reads to a
+-- restricted partner user's accessible orgs is done in the app layer by
+-- psaTicketMappingOrgCondition (routes/psa.ts); that condition is the
+-- enforcement point for the refinement and is not redundant with this policy.
 
 ALTER TABLE psa_ticket_mappings ENABLE ROW LEVEL SECURITY;
 ALTER TABLE psa_ticket_mappings FORCE ROW LEVEL SECURITY;
@@ -124,6 +146,7 @@ DROP POLICY IF EXISTS breeze_org_isolation_delete ON psa_ticket_mappings;
 DROP POLICY IF EXISTS psa_ticket_mappings_isolation ON psa_ticket_mappings;
 CREATE POLICY psa_ticket_mappings_isolation
   ON psa_ticket_mappings
+  -- READ/DELETE: connection access, OR ownership of the referenced device or alert.
   USING (
     public.breeze_current_scope() = 'system'
     OR EXISTS (
@@ -139,7 +162,14 @@ CREATE POLICY psa_ticket_mappings_isolation
        WHERE d.id = psa_ticket_mappings.device_id
          AND public.breeze_has_org_access(d.org_id)
     )
+    OR EXISTS (
+      SELECT 1 FROM alerts a
+       WHERE a.id = psa_ticket_mappings.alert_id
+         AND public.breeze_has_org_access(a.org_id)
+    )
   )
+  -- WRITE: connection access ONLY. Owning the device or alert does NOT license
+  -- attaching a mapping to someone else's connection.
   WITH CHECK (
     public.breeze_current_scope() = 'system'
     OR EXISTS (
@@ -149,10 +179,5 @@ CREATE POLICY psa_ticket_mappings_isolation
            (pc.org_id IS NOT NULL AND public.breeze_has_org_access(pc.org_id))
            OR (pc.partner_id IS NOT NULL AND public.breeze_has_partner_access(pc.partner_id))
          )
-    )
-    OR EXISTS (
-      SELECT 1 FROM devices d
-       WHERE d.id = psa_ticket_mappings.device_id
-         AND public.breeze_has_org_access(d.org_id)
     )
   );
--- a/apps/api/migrations/2026-08-17-psa-connections-partner-ownership.sql
+++ b/apps/api/migrations/2026-08-17-psa-connections-partner-ownership.sql
@@ -1,0 +1,158 @@
+-- Partner-owned PSA connections (epic #2135).
+--
+-- An MSP's PSA (ConnectWise, Autotask, Halo, Jira, Zendesk, ServiceNow,
+-- Freshservice) is a PARTNER-level system: one tenant of the PSA covers every
+-- customer the MSP manages. Until now a psa_connections row was always owned by
+-- exactly one org (org_id NOT NULL), so an MSP had to re-enter the same PSA
+-- credentials once per customer org. This migration makes a connection ownable
+-- by EITHER an org (org_id set, partner_id NULL — the existing shape, retained
+-- for the co-managed-IT case where a customer brings their OWN Jira/Zendesk) OR
+-- a partner (partner_id set, org_id NULL — the "partner-wide / all orgs"
+-- shape), enforced by an exactly-one-axis CHECK.
+--
+-- Mirrors alert_rules (2026-07-01), configuration_policies (2026-06-27),
+-- software_policies (#2126) and ticket_forms (2026-07-11). The sibling
+-- accounting_connections table is already partner-axis, so this brings the two
+-- external-system integrations onto the same ownership axis.
+--
+-- The `backup_configs` org-only precedent deliberately does NOT apply here:
+-- that table holds customer-owned STORAGE credentials, whereas a PSA
+-- credential is the MSP's own.
+--
+-- psa_ticket_mappings has no org_id/partner_id of its own — its tenancy is
+-- entirely `connection_id -> psa_connections`. Its four org-only join policies
+-- (2026-04-11-bucket-c-dead-cleanup-rls.sql) are rewritten to dual-axis IN THIS
+-- SAME FILE: left org-only they would make every mapping under a partner-owned
+-- connection both invisible AND unwritable, because the parent's org_id is NULL
+-- for exactly those rows. Same hazard the ticket_form_org_links policy
+-- (2026-07-11) documents.
+--
+-- No data cleanup is needed: org_id is NOT NULL today, so every existing row
+-- already satisfies the XOR (org_id set, partner_id NULL) the moment the
+-- constraint lands. Nothing is UPDATEd or DELETEd here, hence no ROW_COUNT
+-- diagnostics block.
+--
+-- Idempotent: ADD COLUMN IF NOT EXISTS, guarded CHECK, DROP POLICY IF EXISTS
+-- then CREATE. Re-applying is a no-op. No inner BEGIN/COMMIT (autoMigrate wraps
+-- each file in a transaction).
+
+-- ============================================
+-- Step 1: schema — add partner_id, relax org_id, XOR CHECK
+-- ============================================
+
+ALTER TABLE psa_connections
+  ADD COLUMN IF NOT EXISTS partner_id uuid REFERENCES partners(id);
+
+ALTER TABLE psa_connections
+  ALTER COLUMN org_id DROP NOT NULL;
+
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'psa_connections_one_owner_chk'
+      AND conrelid = 'psa_connections'::regclass
+  ) THEN
+    ALTER TABLE psa_connections
+      ADD CONSTRAINT psa_connections_one_owner_chk
+      CHECK ((org_id IS NULL) <> (partner_id IS NULL));
+  END IF;
+END $$;
+
+CREATE INDEX IF NOT EXISTS psa_connections_partner_id_idx
+  ON psa_connections(partner_id);
+
+-- ============================================
+-- Step 2: RLS — dual-axis (org OR partner) + system short-circuit
+-- ============================================
+--
+-- Replaces the four per-command breeze_org_isolation_* policies installed by
+-- the generic org_id sweep (0008-tenant-rls.sql) with ONE policy covering all
+-- commands, per the epic #2135 playbook.
+
+ALTER TABLE psa_connections ENABLE ROW LEVEL SECURITY;
+ALTER TABLE psa_connections FORCE ROW LEVEL SECURITY;
+
+DROP POLICY IF EXISTS breeze_org_isolation_select ON psa_connections;
+DROP POLICY IF EXISTS breeze_org_isolation_insert ON psa_connections;
+DROP POLICY IF EXISTS breeze_org_isolation_update ON psa_connections;
+DROP POLICY IF EXISTS breeze_org_isolation_delete ON psa_connections;
+DROP POLICY IF EXISTS psa_connections_isolation ON psa_connections;
+CREATE POLICY psa_connections_isolation
+  ON psa_connections
+  USING (
+    public.breeze_current_scope() = 'system'
+    OR (org_id IS NOT NULL AND public.breeze_has_org_access(org_id))
+    OR (partner_id IS NOT NULL AND public.breeze_has_partner_access(partner_id))
+  )
+  WITH CHECK (
+    public.breeze_current_scope() = 'system'
+    OR (org_id IS NOT NULL AND public.breeze_has_org_access(org_id))
+    OR (partner_id IS NOT NULL AND public.breeze_has_partner_access(partner_id))
+  );
+
+-- ============================================
+-- Step 3: psa_ticket_mappings — rewrite the join policies to dual-axis
+-- ============================================
+--
+-- Tenancy join: psa_ticket_mappings.connection_id -> psa_connections. The old
+-- policies asserted breeze_has_org_access(pc.org_id), which is FALSE whenever
+-- pc.org_id IS NULL — i.e. for every partner-owned connection. Collapsed from
+-- four per-command policies to one, matching the parent above.
+--
+-- THREE arms, not two. The third (device_id -> devices.org_id) is not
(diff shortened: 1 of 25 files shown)


bdunncompany:
On the ⚠️ note — **you don't need to dispatch CI manually. `integration-test` already blocks this PR**, and all four shards are running on it right now (run 31307060392, shards 1–4 of `Integration Tests`).

The "non-blocking on PRs" belief is stale. Verified on this branch:

- `continue-on-error: ${{ github.event_name == 'pull_request' }}` appears exactly once in `ci.yml`, at line 1786, and it belongs to **`smoke-test`** — not `integration-test`. That job's own comment says it: *"The API integration suite used to run here too — it now has its own BLOCKING integration-test job above; this job is purely the Docker image build + stack boot + endpoint smoke."*
- `integration-test` (line 1652) has no `continue-on-error` at all.
- `ci-success` lists `integration-test` in `needs` and hard-fails on it:
  ```
  INTEGRATION_TEST_RESULT: ${{ needs.integration-test.result }}
  ...
  [[ "${INTEGRATION_TEST_RESULT}" != "success" ]] || \
  ...
  exit 1
  ```

So `rls-coverage`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip` and your new `psaConnectionsPartnerRls` suite all gate this PR on their own. The `gh workflow run CI --ref` step isn't needed here — that guidance applies to a **stacked** PR based on a sibling branch, which never triggers `ci.yml` at all (`pull_request: branches: [main]`). This one targets `main`, so it triggers normally.

Worth noting the repo's `CLAUDE.md` still says *"Integration tests run in the `smoke-test`/`integration-test` jobs — non-blocking on PRs (`continue-on-error` on pull_request only)"*, which conflates the two jobs and is what makes this trap keep reappearing. That line is now wrong for `integration-test` and is probably worth a one-line fix so the next author doesn't hand-dispatch CI either.

I'll flag it here if any integration shard comes back red.

bdunncompany:
Confirming the green you asked for — **all four integration shards passed, no manual dispatch needed.**

Run 31307060392:

| Job | Result |
|---|---|
| Integration Tests (shard 1/4) | pass — 7m45s |
| Integration Tests (shard 2/4) | pass — 7m29s |
| Integration Tests (shard 3/4) | pass — 7m15s |
| Integration Tests (shard 4/4) | pass — 5m55s |
| Smoke Test | pass — 3m11s |
| CI Success | pass |

**56 checks pass, 0 fail, 0 pending.**

Shard 1 is the one that covers your specific worry: it carries the `if: matrix.shard == 1` steps, so `test:rls-coverage` and `test:request-db-role` both ran there on top of its share of the sharded suite. Your new `psaConnectionsPartnerRls` suite, `tenantCascade`, `tenant-export-policy` and `tenantExportErasureRoundtrip` are all inside the sharded glob and green.

So every suite you flagged as unrunnable in your environment has now actually executed against real Postgres and passed.

bdunncompany:
Heads up — `70ada9ad7` reddened this PR. **Integration shard 2/4 fails, 6 tests, all in your new `psaTicketMappingOrgScoping.integration.test.ts`**, every one `expected 500 to be 200`:

```
× LEAK GUARD: a selected-access partner user sees only their own orgs' mappings
× LEAK GUARD: the same holds for GET /connections/:id/tickets
× the ALERT anchor is scoped too, not just the device anchor
× a FULL partner admin still sees every org's mappings
× FINDING 8: an ORG-scope caller sees their own device's ticket under a PARTNER-wide connection
× an org-scope caller still cannot see ANOTHER org's ticket on that connection
```

The 500s come from Postgres rejecting the query in `psaTicketMappingOrgCondition`:

```
PostgresError: malformed array literal: "e7385528-8a6c-493b-95cf-912fbc8f8a55"
PostgresError: cannot cast type record to uuid[]
```

Both on `select ... from "psa_ticket_mappings" where (...)`.

## Cause

The new helper interpolates a JS array straight into a raw `sql` template:

```ts
AND d.org_id = ANY(${orgIds}::uuid[])
```

**This repo already has this exact bug documented — #2655**, and the fix comment in `services/platformAdminBootstrap.ts` describes your symptom precisely:

> Uses `inArray` — which renders `lower(email) in ($1, $2, ...)` with ONE bound param per email — rather than interpolating the JS array into a raw `sql` template as `= ANY(${emails}::text[])`. The `= ANY(...::text[])` form binds the whole array as a single param, and … that param's text reaches Postgres as a bare string (not a `{...}` array literal), so the `::text[]` cast fails at runtime with `malformed array literal`.

The two different errors line up with the two array sizes, which is consistent with that mechanism: a single accessible org binds one bare uuid string that `::uuid[]` can't parse (`malformed array literal: "<uuid>"`), while several render as `ANY($1, $2)` which Postgres reads as a record (`cannot cast type record to uuid[]`). `extensionTenancyRls.integration.test.ts:17` records a third instance of the same trap.

## Suggested fix

Swap both arms to `inArray`, matching `buildUnpromotedAdminMatch`:

```ts
AND EXISTS (SELECT 1 FROM devices d
             WHERE d.id = ${psaTicketMappings.deviceId}
               AND ${inArray(sql`d.org_id`, orgIds)})
```

Your existing `if (orgIds.length === 0) return sql\`false\`` early return already covers the one caveat #2655 calls out — `inArray([])` renders `in ()`, which is invalid SQL.

## Worth noting

Your 74 lines of new `psa.test.ts` cases passed; only the real-DB suite caught this, because the mocked unit tests never send the SQL to Postgres. Same reason #2655 shipped. It's also a decent argument for the integration job being blocking — which it is, per my earlier comment.

The logic of the fix looks right to me, for what it's worth — separating ticket DATA scoping from connection CONFIG visibility is the correct call, and the AND-over-set-anchors reading is the conservative one. It's purely the array binding that's broken.

PR #3309, opened 2026-08-09
chore(orgs): drop the legacy accounting_provider / accounting_external_id columns

ToddHebebrand:
Contract phase of the org-external-links migration (epic #3249). Drops the legacy single-valued `organizations.accounting_provider` / `accounting_external_id` pair, its partial unique index, and the last code that reads them.

Prerequisite #3298 (QuickBooks importer onto the shared org-import seam) is merged, so nothing writes the pair any more.

## Sweep — no production writers or readers left

`rg -na "accountingExternalId|accounting_external_id|accountingProvider|accounting_provider"` over `apps/api/src`, `apps/web/src`, `apps/portal`, `packages`, `e2e-tests` before this PR returned only:

| Kind | Location |
|---|---|
| **Reader** (removed here) | `services/orgImport/index.ts:122-123, 156-160` — the union loader |
| Schema | `db/schema/orgs.ts:129-130, 141-142` |
| Export policy | `services/tenantExportPolicyRegistry.ts:320` |
| Comments | `services/accounting/quickbooksCustomerImport.ts:12`, `db/schema/orgExternalLinks.ts:7-8` |
| Test fixtures | `orgImport.test.ts`, `quickbooksCustomerImport.test.ts`, `orgExternalLinksRls.integration.test.ts:202` |
| Historical | `migrations/2026-06-29-…`, `migrations/2026-08-08-…`, `docs/superpowers/**` |

**No writer.** Zero references in `apps/web`, `apps/portal`, `packages`, or `e2e-tests`.

> ⚠️ **`grep -r` silently skips `services/orgImport/index.ts`** — the file contains a literal NUL byte (the `linkKey` separator), so plain grep classifies it as binary and prints nothing. The only production reader in the repo lives in that file. Use `rg -na` / `grep -a` when sweeping for org-import call sites.

A second sweep for the index name `organizations_accounting_external_uniq` turned up one more live reference the column sweep missed: `LINK_UNIQUE_CONSTRAINTS` in the seam's concurrent-link recovery.

## Changes

**Readers → link-table-only**
- `services/orgImport/index.ts` — `loadPartnerState` no longer selects the legacy pair; both `linkByKey` and the `linkedOrgSystems` index (added by #3298) are now built from `organization_external_links` alone.
- `LINK_UNIQUE_CONSTRAINTS` drops `organizations_accounting_external_uniq`; only the link table's unique index can now signal "another import won the race".
- `services/accounting/quickbooksCustomerImport.ts` — header comment only. #3298 already removed its union loader and its unique-violation fallback; both now live in the seam.

**Schema** — `db/schema/orgs.ts`: both columns and the `organizations_accounting_external_uniq` partial unique index removed (and the now-unused `sql` import).

**Export policy** — `services/tenantExportPolicyRegistry.ts`: both columns removed from the `organizations` `included` bucket. Removing a column from a registered table breaks `tenant-export-policy.integration.test.ts` exactly as adding one does.

**Migration** — `apps/api/migrations/2026-08-18-drop-organizations-accounting-columns.sql`:
1. Defensive re-backfill into `organization_external_links`, guarded on the columns still existing and reporting a nonzero count as a `WARNING`;
2. `DROP INDEX IF EXISTS organizations_accounting_external_uniq;`
3. `ALTER TABLE organizations DROP COLUMN IF EXISTS …` for both.

Date `2026-08-18` sorts after every shipped migration (last is `2026-08-15-snmp-poll-attempt-backoff.sql`) **and** after `2026-08-17-psa-connections-partner-ownership.sql` from the open #3308, so ordering holds whichever lands first. Not in the closed `2026-08-06` block. `scripts/check-migration-naming.sh` passes.

## Why the defensive re-backfill

`2026-08-08` already backfilled, but a deployment that wrote the legacy pair *afterwards* — an unmigrated code path, a manual `UPDATE`, or **a self-hoster jumping several versions in one upgrade** — would otherwise lose that linkage silently on the `DROP`, and the next import would create a duplicate tenant. The re-backfill runs before the drop, so skipping versions is safe.

Every row that **cannot** be carried over is counted and logged rather than silently discarded (CLAUDE.md's forensic-trail rule):

- **`conflicting`** — `(partner_id, system, external_id)` is already linked to a *different* org, so `ON CONFLICT` drops our row. This is the nastiest silent outcome: the losing org ends up with no linkage, the next QuickBooks import sees it as unlinked, and the seam offers "confirm this match" — collapsing two QuickBooks customers onto one tenant, exactly what `matchedOrganizationLinkedToSystem` (#3298) exists to prevent. Counted with an explicit join on `l.org_id <> o.id` so the ordinary already-backfilled case (same org) is **not** miscounted as a loss.
- **`providerless`** — `accounting_external_id` set with a NULL `accounting_provider`. The 2026-06-29 partial unique index was `WHERE accounting_external_id IS NOT NULL`, so these rows are legal and have been storable all along; they can't be migrated because `organization_external_links.system` is `NOT NULL`.

`ON CONFLICT` names its target explicitly (`(partner_id, system, external_id)`) rather than a bare `DO NOTHING`, so it can only ever absorb the intended collision.

The guard requires **both** columns, not just one: on a half-dropped schema a single-column guard would pass and then abort with `column "accounting_provider" does not exist`, killing autoMigrate at boot so the API never starts. plpgsql resolves column names lazily, so the plain `INSERT` is replay-safe once the columns are gone — no `EXECUTE` wrapper, which keeps the statement under parse-time checking.

## ⚠️ Release-notes item: this release is a ONE-WAY DOOR for `organizations`

**Rolling back past this release requires restoring the two columns by hand first — the migration is forward-only.** The columns drop in the same release that removes the last readers, so reverting to the previous image (the deploy procedure keeps `.env.bak-pre-<version>` for exactly this) leaves the old code issuing `SELECT accounting_provider …` against a table that no longer has it. There is no down-migration.

To roll back, restore the columns before starting the old image:

```sql
ALTER TABLE organizations ADD COLUMN IF NOT EXISTS accounting_provider text;
ALTER TABLE organizations ADD COLUMN IF NOT EXISTS accounting_external_id text;
CREATE UNIQUE INDEX IF NOT EXISTS organizations_accounting_external_uniq
  ON organizations (partner_id, accounting_provider, accounting_external_id)
  WHERE accounting_external_id IS NOT NULL;
```

The restored columns come back **empty**; linkage written since the upgrade lives only in `organization_external_links`. The old QuickBooks importer reads the legacy pair, so it would treat those orgs as unimported. Prefer rolling forward.

## ⚠️ Maintainer check needed before merge: the `breeze-billing` consumer

`deploy/docker-compose.prod.yml:376-383` runs a `billing` service on the **same `DATABASE_URL`**, so it queries this `organizations` table — and the documented deploy line (`docker compose pull api web portal`) **does not restart it**. Its source is a separate repo and **is not present on this machine**, so I could not sweep it.

**A maintainer must confirm `breeze-billing` does not select `accounting_provider` / `accounting_external_id` before merging** (`rg -na 'accounting_provider|accounting_external_id'` in that checkout). If it does, it needs its own release and an explicit restart, because it will keep running the old query against the migrated schema.

## Deployment-ordering caveat

**Safe under the current single-API-container-per-droplet model**: `autoMigrate` runs at new-code boot, so old code — whose Drizzle selects still name these columns — never runs against the migrated database. The window where a `SELECT accounting_provider` could hit a dropped column does not exist.

**A future multi-replica or blue/green rollout would need this split into two releases**: an expand phase that stops reading the columns, shipped and fully rolled out first, then a contract phase that drops them. Worth remembering before adding replicas.

## Tests

- `services/orgImport/orgImport.test.ts` — legacy-union fixtures removed; the "matches via the legacy accounting_* columns" case is replaced by a regression test asserting that an org **without** a link row is a name match, never a link match. The legacy-constraint recovery case is re-pointed at `organization_external_links_uniq` so it still covers the postgres.js `constraint_name` field path.
- `services/accounting/quickbooksCustomerImport.test.ts` — the two legacy-column tests collapse into their link-table twins; added coverage that a same-named but unlinked org is *not* reported as already imported.
- `__tests__/integration/orgExternalLinksRls.integration.test.ts` — the backfill suite was replaying `2026-08-08` standalone, which now **errors** against a migrated DB (its `DO` block references dropped columns unguarded). Re-pointed at the new migration and reworked: it asserts the columns and index are absent from the migrated schema, then recreates the pre-drop shape with raw SQL, writes a legacy-only link, and proves the drop migration rescues it into the link table exactly once, removes the columns, and replays as a no-op. Raw SQL because the Drizzle schema no longer knows these columns.

  **All of that DDL is inside `try`/`finally` with an unconditional cleanup.** The test resurrects dropped columns on the *shared* integration database; DDL commits immediately, the per-test TRUNCATE only removes rows, and autoMigrate will not re-drop them (the migration is already in `breeze_migrations`). A column left behind would persist across every later run and redden both the sibling test and `tenant-export-policy.integration.test.ts` (`organizations.accounting_provider: unclassified`) until someone hand-dropped it on the CI database. The setup DDL itself is inside the `try` so a half-applied setup is also cleaned up. Precedent: `tenant-export-policy.integration.test.ts`.

## Verification

| Check | Result |
|---|---|
| `tsc --noEmit` (apps/api) | ✅ clean |
| `orgImport.test.ts`, `slug.test.ts`, `quickbooksCustomerImport.test.ts`, `routes/orgs.test.ts`, `autoMigrate.test.ts`, `cascadeDelete.test.ts` | ✅ 359 passed |
| `tenantExportPolicy`, `tenantExport`, `tenantCascade`, `tenantCascade.partner`, `tenantErasure`, `tenancyRegistry`, `synthetic`, `moveOrg.coverage`, `composeBindMounts` | ✅ 114 passed |
| `scripts/check-migration-naming.sh` | ✅ OK |
| Full API unit suite (all files) | ⚠️ not completed locally — the run exceeded the sandbox budget and was cut short; CI's **Test API** job covers it |
| Contract suites (RLS + integration) | ✅ green **in CI** (not runnable locally — see below) |
| `pnpm db:check-drift` | ⚠️ not runnable locally; `Check Migrations` covers it in CI |

> The contract suites **cannot run in the authoring environment** (no Docker, no Postgres), so they were verified **in CI instead** — checked by `conclusion`, not by the `gh pr checks` label, since these jobs are `continue-on-error` on PRs and a failure can display as a pass.
>
> **CI run [31311475832](https://github.com/LanternOps/breeze/actions/runs/31311475832) on `0243cda0f` — `conclusion=success`**, including all four `Integration Tests` shards, `Test API`, `Check Migrations`, and `Check Migrations (non-superuser)`. The relevant suites are confirmed **executed**, not skipped (`runDb` is `it.runIf(DATABASE_URL)`, so a skip would still report success):
>
> | Suite | Shard | Result |
> |---|---|---|
> | `orgExternalLinksRls.integration.test.ts` | 4/4 | ✅ 8 tests, 2531ms |
> | `rls-coverage.integration.test.ts` | 1/4 | ✅ 75 tests, 1089ms |
> | `tenant-export-policy.integration.test.ts` | 1/4 | ✅ 2 tests, 700ms |
> | `tenantExportErasureRoundtrip.integration.test.ts` | 1/4 | ✅ 2 tests, 3077ms |
> | `tenantCascade.integration.test.ts` | — | ✅ 5 tests, 1465ms |
>
> `pnpm db:check-drift` still could not be run (it needs a live database too), but `Check Migrations` and `Check Migrations (non-superuser)` both passed in CI, which applies the full migration set to a fresh Postgres — the same property the drift script asserts.

No tenancy shape changed and no table was added or removed, so `CORE_ORG_CASCADE_DELETE_ORDER`, the device cascade lists and the RLS allowlists are untouched — `organizations` keeps its existing registrations. `CORE_TENANT_EXPORT_POLICY` is the only cascade contract this PR touches, because it is the only one that fires on a column change.


bdunncompany:
Same note as on #3308 — **`integration-test` already blocks this PR; the `gh workflow run CI --ref chore/drop-accounting-columns` step isn't needed.**

Verified in `ci.yml`: `integration-test` (line 1652, 4 shards) has no `continue-on-error`, and `ci-success` hard-fails on `needs.integration-test.result` (2063 → 2093). The `continue-on-error: ${{ github.event_name == 'pull_request' }}` at line 1786 belongs to **`smoke-test`**, whose own comment says the API integration suite "now has its own BLOCKING integration-test job above". Confirmed empirically on #3308: all four shards ran and gated it, with shard 1 additionally running `test:rls-coverage` and `test:request-db-role`.

So `tenant-export-policy`, `tenantExportErasureRoundtrip`, `rls-coverage` and your reworked `orgExternalLinksRls` migration-replay test all gate this PR on their own — including the export-policy edit, which is the change class you correctly flagged as highest-risk here.

The `pnpm db:check-drift` gap is real and separate, though: that runs as a **step-level `continue-on-error: true`** inside `test-api` (line 88), so it will not fail anything even when it does run. If drift matters for the column drop, its result has to be read from the job log rather than inferred from a green check.

This warning came from `CLAUDE.md` itself, which still says integration tests are "non-blocking on PRs" and conflates the two jobs — that stale line is why the same caveat appeared on #3308 too. Opened #3310 to fix it so the next author doesn't hand-dispatch CI either.

I'll flag it here if any shard on this PR comes back red.

bdunncompany:
Confirming the green you asked for: **all four integration shards passed** on run 31309270490 — shard 1 7m36s, shard 2 5m31s, shard 3 7m19s, shard 4 8m3s. 56 checks pass, 0 fail, 0 pending.

So `tenant-export-policy`, `tenantExportErasureRoundtrip`, `rls-coverage` and the reworked `orgExternalLinksRls` migration-replay test all executed against real Postgres and passed, including the export-policy column removal.

Still worth reading the drift result out of the `test-api` log rather than trusting the check, for the step-level `continue-on-error: true` reason above.

ToddHebebrand:
Cleared the open maintainer item on `breeze-billing` — it does **not** reference the dropped columns.

Swept `LanternOps/breeze-billing` via the GitHub code-search API for all four spellings (`accounting_provider`, `accounting_external_id`, `accountingProvider`, `accountingExternalId`): **0 hits each**. Validated the negative with control queries against the same repo so this isn't an unindexed-repo false negative — `organizations` returns 5 hits, `DATABASE_URL` 5, `import` 74, so the index is live and the zero is real.

The service does touch `organizations`, but not these two columns, so the drop is safe for it even though the documented deploy line doesn't restart it.

Issue #3258, opened 2026-08-08
First-class organization contacts, with API and import support


ToddHebebrand:
## Problem

Breeze has no first-class concept of a contact.

- There is **no `contacts` table**. `organizations.billingContact` and `sites.contact` are JSONB blobs holding `{name?, email?, phone?}` (see `apps/api/src/db/schema/orgs.ts`, and the note in `docs/superpowers/specs/billing/2026-06-29-quickbooks-customer-import-design.md`).
- `apps/api/src/services/aiToolsOrgs.ts:473` — the `add_contact` action is **unimplemented**; it returns guidance text rather than creating anything.
- Consequently no importer can create contacts, and every migration guide in `apps/docs/src/content/docs/migration/` has to tell the reader "Breeze cannot create contacts programmatically today; enter them by hand or keep them in your PSA."

Every RMM this matters for has real contact data: Datto site contacts, Automate contacts, NinjaOne org contacts, Atera contacts, Syncro customer contacts. All of it is currently stranded.

## Why it matters beyond migration

One contact per organization and one per site is already limiting outside of migration:

- A customer with an IT lead, a billing contact, and an after-hours escalation contact cannot be represented.
- Alert notification routing and the client portal both want "who do I tell", and a single JSONB blob cannot express a list, a role, or an escalation order.
- The JSONB shape means contacts cannot be queried, deduped, or reported on.

## What needs deciding

This is a data-model decision, not an import feature — which is why it blocks rather than follows the importers:

1. **A real `contacts` table** (org-scoped, shape 1) with a role/type and optional site association, or keep JSONB and simply allow an array?
2. If a table: the full tenancy contract — RLS policies in the creating migration, `CORE_ORG_CASCADE_DELETE_ORDER` in `services/tenantCascade.ts`, and `CORE_TENANT_EXPORT_POLICY` in `services/tenantExportPolicyRegistry.ts`. Contact details are personal data, so the export/erasure classification matters more here than usual — this is squarely GDPR-relevant.
3. Migration path for the existing `billingContact` / `contact` JSONB values.
4. Whether contacts sync from the PSA (which is usually the system of record for them) rather than being owned in Breeze.

Question 4 may be the real answer: if the PSA owns contacts, the right move could be a read-through link rather than a local table. That should be decided before building either.

## Related

Blocks contact import in #3242 (deferred from v1 scope) and the PSA import in #3246. Epic #3249.

---

[...]

PR #3316, opened 2026-08-09
feat(contacts): first-class organization contacts — tables, contracts, compat service (#3258)

ToddHebebrand:
Phase 3 of the RMM migration epic (#3249). **Draft — the data layer only.** The
existing jsonb writers are not yet routed through the compat service, so today
`contacts` is populated by the backfill alone.

Design: `docs/superpowers/specs/onboarding-signup/2026-08-09-organization-contacts-design.md`
Plan: `docs/superpowers/plans/2026-08-09-organization-contacts.md` (branch `docs/rmm-migration-specs-plans`)

## What this adds

- **`contacts`** and **`contact_external_links`** — RLS shape 1, enabled **and** forced, four `breeze_has_org_access(org_id)` policies each, in the creating migration.
- **`portal_users.contact_id`** — a portal user becomes a login attached to a contact rather than a second kind of person.
- **`services/contacts/compat.ts`** — the single writer of both the table and the legacy jsonb.
- Contract registrations: `CORE_ORG_CASCADE_DELETE_ORDER`, `CORE_TENANT_EXPORT_POLICY` (both new tables **and** the new `portal_users` column).
- Two integration suites (below) that prove the tenancy and backfill behaviour rather than asserting it in prose.

## Why the legacy jsonb columns are NOT dropped

Not a deferral — a decision. Three shipped contracts depend on `sites.contact` existing under that name:

1. `breeze_partner_export_sites_update()` detects change with a hardcoded tuple reading `old_row.contact` / `new_row.contact` (`2026-07-18-partner-export-org-locks.sql:279-284`). Move the data out and a contact-only edit silently stops bumping `sites.partner_export_updated_at`.
2. `partnerSiteContactSchema` is a `.strict()` **public** partner-API DTO and export records are content-hashed, so changing the emitted shape re-hashes every site record and forces a full re-sync across all partner consumers.
3. `organizations.billing_contact` is deliberately excluded from the partner API (negative regression test) while `sites.contact` is included. One table cannot express that asymmetry.

## Verification

Everything below was read from **job records**, not the run conclusion — integration shards are `continue-on-error` on pull requests, so `gh pr checks` renders a genuine failure as a pass:

```
gh api repos/LanternOps/breeze/actions/runs/<id>/jobs --jq '.jobs[] | select(.conclusion!="success")'
```

Latest green run: **31355245362** (commit `dd31a1594`).

**Migration mechanics**
- [x] Applies cleanly to a fresh database — both `Check Migrations` and `Check Migrations (non-superuser)`.
- [x] Re-applying is a true no-op — `contactsBackfillMigration` replays the file from disk and asserts row **identity** is unchanged, not merely row count.
- [ ] ~~`pnpm db:check-drift` — Drizzle schema matches the SQL~~ **This item was mis-specified.** `scripts/check-drift.ts:17-26` states that schema-vs-live-DB comparison is *intentionally* excluded (the drizzle-kit round-trip emits ~2k lines of false positives from FK-name normalisation, array-default introspection gaps, and RLS statements Drizzle doesn't model). The script only verifies the migration ledger. The coverage this item was reaching for — the `roles text[]` default, the `check()` constraint, the partial-index predicates — is asserted behaviourally by `contactsRls` instead. Separately: that CI step has never executed on any PR (it runs in the `Lint` job, which has no Postgres, and carries `continue-on-error: true`). Pre-existing hygiene issue, not specific to this PR.

**Contract suites — all executed against real Postgres**
- [x] `rls-coverage` (75 tests) — both tables auto-discovered via `org_id`; neither needs an allowlist entry, neither landed in `DUAL_AXIS_TENANT_TABLES`.
- [x] `tenantCascade` (5 tests) — including the two properties I could not check in plain JS ("every `org_id` table present", "FK children before parents").
- [x] `tenant-export-policy` (2 tests) — the two new entries plus the amended `portal_users` entry.
- [x] `tenantExportErasureRoundtrip` (2 tests).

**Constraint behaviour** — `contactsRls.integration.test.ts` (5 tests, shard 4), through the real driver as unprivileged `breeze_app`:
- [x] Cross-tenant forge rejected with 42501 on **both** tables; org B can neither SELECT nor DELETE org A's contact.
- [x] `contacts_site_org_fk` rejects a contact pinned to another org's site (23503), and deleting a site cascades to its pinned contacts — the documented consequence of choosing CASCADE over a composite SET NULL.
- [x] `contacts_org_primary_uniq` / `contacts_site_primary_uniq` each reject a second primary (23505), while a site primary and an org primary coexist in one org. That coexistence is what distinguishes the shipped predicates from a plain `is_primary` partial index.
- [x] `contacts_identifiable_chk` rejects an all-null row (23514) and accepts each of name/email/phone/mobile alone.
- [x] `contact_external_links_uniq` is org-scoped — the same source id links under two orgs of ONE partner, and duplicates within an org are rejected.

**Backfill correctness** — `contactsBackfillMigration.integration.test.ts` (4 tests, shard 3). CI migrates schema-fresh, so every backfill had otherwise only ever run against **zero rows**; this suite seeds the real pre-migration shapes and replays the file:
- [x] Counts land in the Postgres log even at 0 (all four `RAISE WARNING`s visible in the migration job).
- [x] `jsonb_typeof(...) = 'object'` guards hold — a scalar `billing_contact` (legal, the column is validated with `z.any()`), an array, and an all-blank object are each skipped rather than backfilled into an all-NULL row.
- [x] The row-wise `portal_users` loop gives two portal users sharing one address in one org two **distinct** contacts.
- [x] Step 3a links only on an unambiguous single email match, case-insensitively, and declines an address matching two contacts — which is what makes `(array_agg(c.id))[1]` an exact pick rather than an arbitrary one.

## Deliberate deviation from the plan

The plan called for fixing the pre-existing clobber race (org PATCH replaces `billing_contact` wholesale while `invoiceService` merges into it atomically, so a partial PATCH silently drops merged keys) as a side effect of unifying the writers. **Not done here.** Each entry point reproduces its caller's existing jsonb semantics exactly; the race is unchanged, not worsened. Its natural home is the writer-wiring PR below, where CI can exercise `orgBillingSettings.integration.test.ts` against the unified path.

## Still to come on this branch

Wire the existing writers (including the site PATCH spread-writer at `orgs.ts:~2028`, which has no literal `contact:` token and is invisible to a grep sweep) · contact CRUD + import routes · `add_contact` with its guardrail re-classification · web UI · repoint inbound email's `findOrCreateEmailContact` · docs.


bdunncompany:
I have a database, so I ran your checklist. **Everything you listed passes**, with one correction about what `db:check-drift` actually does and three items I could not cover.

Environment: your branch at `3cf845129`, fresh Postgres 5433 + Redis 6380 via `docker-compose.test.yml`, Node 22.23.2.

## Migration mechanics

**`pnpm db:migrate` on a genuinely empty database** — clean. Confirmed the DB started at 0 tables in `public`, then:

```
2026-08-19-contacts.sql   -> present in breeze_migrations
contacts, contact_external_links -> both exist
RLS:  contacts enabled=t forced=t   |   contact_external_links enabled=t forced=t
policies: 4 on each table
portal_users.contact_id -> uuid, added
```

**Re-apply is a true no-op.** I replayed the file against the already-migrated DB with `ON_ERROR_STOP=1` — exit 0. Every guard fired as designed (`column "contact_id" ... already exists, skipping`, `relation "portal_users_contact_idx" already exists, skipping`).

**Backfill counts land even at 0** — all four `RAISE WARNING`s emitted on the replay:
```
WARNING:  backfilled 0 site contacts
WARNING:  backfilled 0 organization billing contacts
WARNING:  linked 0 portal users to existing contacts
WARNING:  created and linked 0 contacts from portal users
```

### ⚠️ Correction: `db:check-drift` does not check what your checklist assumes

It passed — `No drift detected — all 504 migration files match the breeze_migrations ledger` — but read its own second line:

> `(Schema-vs-live-DB structural drift is covered by integration + RLS coverage tests, not by this script.)`

It compares **migration-file checksums against the ledger**. It does **not** compare the Drizzle schema to the SQL, so it can never have told you about `roles text[]`, the check constraint, or the partial-index predicates. I inspected the live schema directly for those three instead:

| Your concern | Live database |
|---|---|
| `roles text[] DEFAULT '{}'` | `_text`, `NOT NULL`, `DEFAULT '{}'::text[]` ✅ |
| the `check()` constraint | `contacts_identifiable_chk CHECK (name IS NOT NULL OR email IS NOT NULL OR phone IS NOT NULL OR mobile IS NOT NULL)` ✅ |
| four partial indexes, predicates written twice | all four present, predicates correct ✅ |

The four: `contacts_site_idx WHERE site_id IS NOT NULL`, `contacts_org_email_idx (org_id, lower(email)) WHERE email IS NOT NULL`, `contacts_org_primary_uniq (org_id) WHERE is_primary AND site_id IS NULL`, `contacts_site_primary_uniq (site_id) WHERE is_primary AND site_id IS NOT NULL`.

## Contract suites — all green

| Suite | Result |
|---|---|
| `rls-coverage` | **75 passed** |
| `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` | **9 passed / 3 files** |

Both tables were auto-discovered via `org_id`; neither needed an allowlist entry and neither landed in `DUAL_AXIS_TENANT_TABLES`. And specifically the two you flagged as unverified — *"every `org_id` table present"* and *"FK children before parents"* — are asserted by `tenantCascade.integration.test.ts`, which passed.

## Constraint behaviour — the section with no test yet

Probed directly against the live DB. Each expected failure was caught by SQLSTATE, so a silent acceptance would have reported FAIL rather than passing quietly:

```
A contacts_identifiable_chk:   PASS (23514 rejected all-null)
B contacts_org_primary_uniq:   PASS (23505 rejected 2nd org primary)
C contacts_site_primary_uniq:  PASS (23505 rejected 2nd site primary)
D contacts_site_org_fk:        PASS (23503 rejected cross-org site pin)
```

**Cross-tenant forge as `breeze_app`**, with a control so it can't pass vacuously:

```
SET breeze.scope='organization'; SET breeze.accessible_org_ids='<org X>';
CONTROL insert into org X  -> INSERT 0 1
FORGE   insert into org Y  -> ERROR: new row violates row-level security policy for table "contacts"
READ    org Y rows visible -> 0
```

**`contact_external_links_uniq` is org-scoped, and the asymmetry you intended is real:**
```
contact_external_links_uniq       (org_id, system, external_id)
organization_external_links_uniq  (partner_id, system, external_id)
```

## What I did NOT cover

Three backfill items need adversarial fixtures I didn't build, so treat them as still open:

- the `jsonb_typeof(...) = 'object'` guards against a `billing_contact` that legally holds a **string or array** (`z.any()`)
- the row-wise `portal_users` loop where **two portal users share an address in one org**
- step 3a linking only on an **unambiguous single email match**

All my backfill counts were 0 — the guards were never exercised against hostile shapes.

## One more thing, unrelated to the data layer

Your "Reading CI" note says integration shards are `continue-on-error` on pull requests. That's **not true any more**, and this is the third PR carrying it (#3308, #3309, here). `integration-test` (`ci.yml:1652`) has no `continue-on-error` and `ci-success` hard-fails on `needs.integration-test.result` (2063 → 2093); the `continue-on-error: ${{ github.event_name == 'pull_request' }}` at line 1786 belongs to **`smoke-test`**. So `gh pr checks` does *not* render an integration failure as a pass, and the `gh api .../jobs` workaround isn't needed.

**`db:check-drift` is the one that genuinely lies to you** — it runs as a step-level `continue-on-error: true` inside `test-api` (`ci.yml:88`), so it cannot fail your build even when it does run. #3310 fixes the CLAUDE.md line that's the source of the integration-shard misconception.

bdunncompany:
Your `e8673d40` test closes the forge + constraint items cleanly. I went after the backfill gap I'd left open and found one thing worth fixing before this merges.

## Backfill steps 1 and 2 abort the migration on a long jsonb value

`sites.contact` and `organizations.billing_contact` feed `varchar`-bounded columns, but nothing bounds the source:

| Target | Width | Source |
|---|---|---|
| `contacts.name` | `varchar(255)` | `->>'name'` |
| `contacts.email` | `varchar(320)` | `->>'email'` |
| `contacts.phone` | `varchar(64)` | `->>'phone'` |

Both writers are `z.record(z.string(), z.unknown())` — `createOrgSchema.billingContact` and `createSiteSchema.contact` (`packages/shared/src/validators/index.ts:111,119`). The record shape constrains the *top level* to an object, but the values are `unknown`: no `max()`, no string check. So the API accepts an arbitrarily long `phone` today.

`phone varchar(64)` is the one that will actually bite. A realistic free-text entry clears it without looking unusual:

```
555-1234 ext 567, after hours 555-9999, cell 555-0000 (no texts please)   -- 71 chars
```

Run against your `e8673d40` migration text verbatim, on Postgres 16:

```
step 2, billing_contact.phone (71 chars):
  ERROR:  value too long for type character varying(64)
  CONTEXT:  SQL statement "INSERT INTO contacts (org_id, name, email, phone, roles, is_primary) ..."
  PL/pgSQL function inline_code_block line 4 at SQL statement
  psql exit 3

step 1, sites.contact.name (300 chars):
  ERROR:  value too long for type character varying(255)
  psql exit 3
```

**The failure isn't "that row is skipped" — it's `SELECT count(*) FROM contacts` = 0.** Each step is one set-based `INSERT ... SELECT`, so a single overlong value anywhere in the table rolls back the backfill for *every* tenant. I ran a benign control row alongside the hostile one in both cases; the control was lost too, which is how I confirmed the blast radius rather than assuming it.

Since `autoMigrate` wraps each file in a transaction and the DO block re-raises, this surfaces on a populated self-hosted database as a failed migration at boot, not as a partial backfill. Fresh installs and any database whose blobs happen to be short will never see it, which is why CI is green here.

## Suggested fix

Bound the projection at the source. Same three columns in both steps:

```sql
         left(NULLIF(btrim(o.billing_contact->>'name'),  ''), 255),
         left(NULLIF(btrim(o.billing_contact->>'email'), ''), 320),
         left(NULLIF(btrim(o.billing_contact->>'phone'), ''), 64),
```

`left()` is null-safe, so the `COALESCE(...) IS NOT NULL` guard below is unaffected.

Truncation is lossy, and for `email` a truncated address is arguably worse than none — if you'd rather not silently mangle data, the alternative is to widen `phone`/`mobile` to something like `varchar(128)` and keep the truncation only as a backstop. I'd lean toward `left()` on all three regardless: a one-time backfill that aborts the upgrade is a worse outcome than a trimmed phone string, and the pre-existing blobs are already free text.

Either way it's worth a case in `contactsRls.integration.test.ts` — an org whose `billing_contact.phone` is 71 chars, asserting the migration completes and the other org's contact still lands. That's the shape that would have caught this.

## Minor

The header comment says `billing_contact` is "validated with `z.any()`, so the column can legally hold a string, a number, or an array." It's `z.record(z.string(), z.unknown())`, so the API path can't produce a non-object — your `jsonb_typeof(...) = 'object'` guard is still right for legacy rows and direct DB writes, just not for the reason stated.

The two `portal_users` items from my earlier list I still haven't exercised: the row-wise loop with two portal users sharing an address in one org, and 3a's single-match rule. `portal_users.email` is `notNull()`, so 3b can't trip `contacts_identifiable_chk` — that one's cleared.

bdunncompany:
`dd31a159` covers all three of the items I'd left open — including both `portal_users` ones. Nothing outstanding from my original list.

One correction to my own pointer, though: I suggested putting the length case in `contactsRls.integration.test.ts`, and your new file is the better home now that it exists — it already seeds legacy shapes and replays the migration from disk, which is exactly the harness this needs. Paste-ready against your helpers:

```ts
runDb('a long free-text phone does not abort the whole backfill', async () => {
  const partner = await createPartner();
  const benign = await createOrganization({ partnerId: partner.id });
  const hostile = await createOrganization({ partnerId: partner.id });

  await setBillingContact(benign.id, {
    name: 'Jane Doe', email: 'ap@benign.test', phone: '555-1234',
  });
  await setBillingContact(hostile.id, {
    name: 'Bob', email: 'bob@hostile.test',
    // 71 chars, into contacts.phone varchar(64)
    phone: '555-1234 ext 567, after hours 555-9999, cell 555-0000 (no texts please)',
  });

  await replayMigration();

  // The control is the assertion that matters: step 2 is one set-based
  // INSERT ... SELECT, so today the overlong row takes this one down too.
  expect(await contactsFor(benign.id)).toHaveLength(1);
  expect(await contactsFor(hostile.id)).toHaveLength(1);
});
```

Currently that dies at `replayMigration()` with `22001 value too long for type character varying(64)`, and `contactsFor(benign.id)` is empty — which is the part worth pinning down, since it's the difference between one skipped org and a failed upgrade.

`sites.contact` has the same exposure via `name varchar(255)`.

PR #3320, opened 2026-08-09
feat(webhooks): honour the self-hosted private-network gate, drop the local IP policy

bdunncompany:
## The gap

**A self-hosted Breeze cannot send a webhook to anything on its own LAN.**

`webhookSender` requires HTTPS and refuses every RFC1918 address, so an operator whose receiver is an on-prem SIEM, log collector, or ticketing system has no delivery path to it. That is precisely the deployment shape where the receiver is *guaranteed* to be private — the self-hoster owns both ends of the wire.

The failure is also quiet in the worst way: it surfaces as a validation error on save, so it reads like the operator mistyped a URL rather than like a policy decision.

## This is the consolidation the codebase already asked for

I did not invent a policy here. `config/env.ts` already exports `selfHostAllowsPrivateNetwork()`, already tested in `config/env.test.ts`, and its own comment states the intent:

> The DNS-provider (`services/dnsProviders/index.ts`) and PSA (`services/psa/http.ts`) integrations currently carry their own equivalent IS_HOSTED-affirmative gates — **consolidating all three onto this helper is a worthwhile follow-up**

Three integrations already make this exact call. SSO uses the canonical helper; PSA and dnsProviders carry their own copies. **`webhookSender` was the one outbound integration with no gate at all** — and rather than reuse `urlSafety.ts`, it hand-rolled ~54 lines of address-range policy (`parseIpv4Octets` / `isPrivateIpv4` / `isPrivateIpv6`).

That duplicate was not merely redundant, it was **less precise**. It collapsed loopback, link-local, cloud metadata, CGNAT and RFC1918 into one verdict with one error string. With only that predicate there is no way to open an operator's LAN without also opening `169.254.169.254`. `urlSafety.ts` already draws that line correctly via `isAlwaysBlockedIp`; the webhook path just wasn't using it.

So this PR is subtractive: **net −37 lines of source**, one file, and one fewer private implementation of a security policy that must never drift.

## What changes

- Gate on `selfHostAllowsPrivateNetwork()` — fail-closed. Opens only on an affirmative `IS_HOSTED` (`false`/`0`/`no`/`off`); unset, empty, garbage and truthy all stay strict (the #570 lesson).
- When open: permit `http:`, and swap the block predicate to `isAlwaysBlockedIp` — RFC1918/ULA only.
- Apply the same predicate to the DNS-resolved half.
- Pass `allowPrivateNetwork` through to `safeFetch`, so delivery agrees with what was accepted at save time. The TOCTOU re-check exists to catch DNS rebinding, not to overrule the deployment's own policy — without this a URL you accepted would be refused at send.
- Delete the duplicated local policy.

**Hosted SaaS behaviour is byte-identical.**

## It is not an SSRF escape hatch, and that is asserted

Loopback, link-local, cloud metadata and CGNAT (`100.64/10`) remain blocked in **both** modes. The opt-in widens the target space to the operator's own LAN; it does not disable the guard. Seven cases assert this directly rather than leaving it implied:

| still rejected when self-hosted |
|---|
| loopback v4 `127.0.0.1` |
| loopback v6 `::1` |
| link-local `169.254.1.1` |
| AWS metadata `169.254.169.254` |
| Alibaba metadata `100.100.100.200` |
| CGNAT `100.64/10` |
| `localhost` by name |

CGNAT staying blocked is deliberate and matches the documented decision in `urlSafety.ts` — I am not proposing to relax it.

## Control

Reverting the source and keeping the tests fails **8** — precisely the "self-hosted opens RFC1918/ULA + http" cases and the DNS-accepts case. The other **27 pass in both versions**, including every "still rejected" assertion.

That split is the point: the capability tests prove the change does something, and the security assertions prove they are invariants rather than artifacts of it.

## One trap worth flagging for future tests here

The DNS tests stub `dns/promises` directly. `__setLookupForTests` is a `urlSafety`/`safeFetch` hook and does **not** intercept the dynamic `import('dns/promises')` inside `validateWebhookUrlSafetyWithDns`. Without a real stub the lookup simply fails, and every "rejects" assertion passes for the wrong reason. Two of mine did exactly that until I checked why an expected-to-fail case was green.

## Verification

| Check | Result |
|---|---|
| `webhookSender.test.ts` | **35 passed** (+109 lines of tests) |
| full `apps/api` unit suite | **1304 files / 20958 passed**, 5/37 skipped |
| `tsc --noEmit` (apps/api) | 0 errors |
| `eslint` | clean |

No schema change, no migration, no new dependency. One source file touched.

## Provenance

Found while wiring a self-hosted deployment's alert/script/patch events to an on-LAN log collector. Every event type needed is already emitted by the webhook system — the only thing standing between the two was this guard.

### DIFF
--- a/apps/api/src/services/notificationSenders/webhookSender.test.ts
+++ b/apps/api/src/services/notificationSenders/webhookSender.test.ts
@@ -1,8 +1,19 @@
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+// `validateWebhookUrlSafetyWithDns` resolves the hostname via a dynamic
+// `import('dns/promises')`, which `__setLookupForTests` (a urlSafety/safeFetch
+// hook) does NOT intercept. Without this mock the lookup simply fails and the
+// "rejects" assertions pass for the wrong reason.
+const { dnsLookupMock } = vi.hoisted(() => ({ dnsLookupMock: vi.fn() }));
+vi.mock('dns/promises', () => ({
+  lookup: dnsLookupMock,
+  default: { lookup: dnsLookupMock },
+}));
 import {
   sendWebhookNotification,
   validateWebhookConfig,
   validateWebhookUrlSafety,
+  validateWebhookUrlSafetyWithDns,
   redactUrlForLogs
 } from './webhookSender';
 import { __setLookupForTests } from '../urlSafety';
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
Merged. Reviewed this as an SSRF surface rather than a feature change, since it touches which destinations outbound webhooks may reach.

**The gate fails closed, which was the main thing to prove.** `selfHostAllowsPrivateNetwork()` (`apps/api/src/config/env.ts:147`) requires both a recognized self-host signal (`0`/`false`/`no`/`off` only) and `!isHosted()`. Unset, empty, and garbage values all resolve false, so a misconfigured deploy stays strict — which matters here given #570, where an unmapped `IS_HOSTED` silently opened a different gate. The hosted path is byte-identical to before.

**The removed local IP helpers were genuinely superseded, and dropping them fixes a real bypass.** The deleted `isPrivateIpv6` only unwrapped the dotted-decimal `::ffff:a.b.c.d` mapped form; `urlSafety.ts`'s `mappedV4()` also handles the hex-pair form, so `::ffff:a9fe:a9fe` — mapped link-local, i.e. cloud metadata — was reachable through the old helper and is not through the shared one. That's a security improvement the PR body undersells.

Rebinding is handled: `safeFetch` resolves once and pins the connection's `lookup` to the resolved address, so a post-validation rebind can't move the TCP connect, and both the create/update route and the delivery worker go through the DNS-resolving validator.

**One thing worth a follow-up, not a blocker:** in self-hosted mode plain `http:` is permitted for *any* hostname, not only ones resolving to a private address (`webhookSender.ts:79`). That matches what the body says, but it's broader than the "on-LAN receiver" framing implies — a self-host operator gets cleartext HTTP to public destinations too. Defensible for a single-tenant admin-trusted deployment; if the intent is strictly on-LAN, gate the `http:` allowance on the resolved address actually being private.

PR #3321, opened 2026-08-09
feat(agent): compile-time control-plane build-mode host policy (inert by default)

ToddHebebrand:
## What

Adds a compile-time **control-plane build-mode host policy** to the Go agent. A new leaf package `agent/internal/hostpolicy` exposes two ldflag-injected vars — `allowedHosts` and `strictMode` — and pure predicates `Enforced()` / `Strict()`. When `allowedHosts` is set, the build only accepts control-plane servers whose host is in the (exact, case-insensitive, anchored — no wildcard/suffix) allowlist.

**Inert by default.** Both vars are empty in the repo, so an ordinary build is `self-host` mode = unrestricted = current behavior. Nothing changes for any build the repo produces today; the constraints activate only when a release injects the vars via `-ldflags -X`.

## Behavior model

Two tiers, split by when the control-plane URL arrives:

- **Fresh install (bootstrap / enroll)** — refused when `Enforced()` (any hosted build) if the server isn't allowlisted. The bootstrap token is checked *before* it is transmitted, so it is never sent to a non-allowlisted host. Covers the filename host, MSI property, redeem-response redirect, and `--server`.
- **Existing fleet (runtime paths)** — gated only when `Strict()`: server/backup URL validation, failover promotion, and updater control-plane origins. A non-strict ("gap") build never degrades an already-enrolled fleet; it only reports the condition.
- **Startup self-check** lives in the shared `startAgent` choke point (covers console, Windows service, macOS daemon, and quick-support): gap logs a warning and continues; strict hard-fails with an operator message.
- **Quick Support** redeem is gated the same way (host checked before the code is sent).
- The **heartbeat** reports `agentEdition` / `migrationRequired` (both `omitempty`; a self-host build emits neither, so the wire payload is byte-identical to today). The API/dashboard consumer is a separate change.

## Tests

New `_test.go` suites per touched package (`hostpolicy`, `agentapp` bootstrap/enroll/startup/quick-support, `config` validation, `updater` origins, `heartbeat` signal). Whole agent module passes `go test -race ./...`; `go vet` clean. No real hostnames in source — the allowlist is ldflag-only.

### DIFF
--- a/agent/internal/agentapp/bootstrap.go
+++ b/agent/internal/agentapp/bootstrap.go
@@ -209,6 +209,7 @@ func runBootstrap() {
 		bsLog.Error("bootstrap token redemption failed", "error", err.Error())
 		fmt.Fprintf(os.Stderr, "Bootstrap failed: %v\n", err)
 		osExit(1) // hard — roll back the install (osExit: test seam, enroll_error.go)
+		return    // under a noop osExit test seam, don't fall through to a nil res
 	}
 
 	if err := gateRedeemResponse(*res); err != nil {
--- a/agent/internal/agentapp/bootstrap.go
+++ b/agent/internal/agentapp/bootstrap.go
@@ -13,11 +13,35 @@ import (
 	"time"
 
 	"github.com/breeze-rmm/agent/internal/config"
+	"github.com/breeze-rmm/agent/internal/hostpolicy"
 	"github.com/breeze-rmm/agent/internal/logging"
 	"github.com/breeze-rmm/agent/pkg/api"
 )
 
 var errNoBootstrapInput = errors.New("no bootstrap token from filename or properties")
+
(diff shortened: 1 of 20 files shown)


bdunncompany:
`Lint Agent (Go)` is red on one gofmt issue — not the flaky Windows suite, so it won't clear on a rerun. The real line is buried under ~30 lines of `go: downloading` output:

```
internal/hostpolicy/hostpolicy_test.go:47:1: File is not properly formatted (gofmt)
		"https://hosted-a.example.evil.com",  // suffix attack
```

Double space before the `//`, so gofmt wants the comment column normalised across that composite literal. `gofmt -w agent/internal/hostpolicy/hostpolicy_test.go` clears it.

Everything else is green — 53 pass, that 1 fail, 0 pending.

Issue #2879, opened 2026-07-28
suspended → offboarding is unreachable via the API — #2808's clear-superseded-suspension fix can't be triggered in production


ToddHebebrand:
Found + code-verified in the 2026-07-27 pre-release QA sweep. Makes #2808 effectively dead code on the real path.

## Symptom
There is no API path to transition an org from `suspended` to `offboarding`:
- **Partner scope**: `computeAccessibleOrgIds` (`middleware/auth.ts:~271,286`) filters `inArray(organizations.status, ['active','trial'])`, so a suspended org is invisible → `PATCH`/`GET /organizations/:id` 404. (Side effect: a suspended org can't be reactivated by partner scope either — a one-way door.)
- **System/platform-admin scope**: a membershipless platform admin fails `requirePermission` ("No permissions found" — permissions resolve only from membership roles), so it can't drive the transition either.

## Why it matters
#2808 ("clear superseded token suspension on offboarding drain entry") exists precisely for the suspend-THEN-offboard sequence. The fix itself is CORRECT — verified by DB-flipping the org to `active` first and driving entry through the real API: offboarding entry cleared the `tenant_suspended` timestamp, the agent recovered and heartbeated through the drain, and the queued self_uninstall was delivered. But the natural production trigger (offboard a customer you already suspended for non-payment) 404s, so the fix can't be reached the way it was meant to be.

## Fix direction
Allow the suspended→offboarding (and suspended→active reactivation) transitions through some authorized path — e.g. include `suspended` in the accessible-orgs filter for the offboarding/reactivation mutations specifically, or give platform-admin/system scope a membership-independent authority for org lifecycle transitions. Pair with the offboarding-entry deadlock fix (filed separately) since both are on the same PATCH path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ToddHebebrand:
**PR:** #2887 (review-clean, open for merge).

**What it does:** two narrow gates, no general visibility widening —
- Partner scope: a STATUS-ONLY payload moving a suspended, partner-owned org to `active`/`trial`/`offboarding` gets a lifecycle override (system-context ownership probe; the UPDATE re-asserts partner-owned AND still-suspended in its own WHERE under a system context). `churned` is deliberately excluded so a suspended org can't skip the drain. Suspended orgs stay invisible everywhere else.
[...]

PR #3328, opened 2026-08-10
feat(contacts): route all ten legacy jsonb writers through the compat service (#3258)

ToddHebebrand:
Phase 3 of the RMM migration epic (#3249), following #3316.

#3316 shipped `services/contacts/compat.ts` **with no callers** — `contacts` was populated by the migration backfill alone, so every contact write since has drifted from the table. This wires the writers.

## The sweep

Ten writers, found by enumerating `.update(organizations)` / `.update(sites)` / inserts rather than by grepping for `contact`:

| # | Entry point | Wired as |
|---|---|---|
| W1 | `POST /orgs/organizations` | `syncBillingContactRow` |
| W2 | `PATCH` + `PUT /orgs/organizations/:id` | `syncBillingContactRow` |
| W3 | `PATCH /orgs/:orgId/billing-settings` | `mergeBillingContact` |
| W4, W8–W10 | org import ×4 (CSV · PSA · QuickBooks) | `syncBillingContactRow` / `syncSiteContactRow` |
| W5, W6 | `POST` + `PATCH /orgs/sites` | `syncSiteContactRow` |
| W7 | partner API site create (**public surface**) | `syncSiteContactRow` |

**Two would have been missed by a grep sweep**, which is why the enumeration was done by write target:

- `services/orgImport/index.ts` holds **four** of the ten and contains a NUL byte, so plain `grep -r` classifies it binary and prints nothing. It needs `rg -na`.
- The site PATCH (`orgs.ts`) reaches the column through `const writeData = { ...data, updatedAt }` — there is no literal `contact:` token in the handler; the field enters via schema inheritance declared ~1800 lines earlier.

## Why three sites keep their own column write

They use the mirror-only `sync*` entry points instead of `replace*`:

1. **org PATCH** re-asserts partner-ownership and suspended-status in the UPDATE's own `WHERE` (#2879) so a concurrent change cannot let a write land on an org that stopped qualifying. `replaceBillingContact` targets a bare `eq(id, orgId)` and would drop that guard.
2. **site PATCH**'s single UPDATE has a load-bearing 0-row result — it detects an RLS rejection the prior SELECT missed.
3. **create paths** already persist the blob in the insert.

Consequently both `replace*` exports have no production caller. Rather than leave them as an attractive nuisance, each now documents the hazard and points to the `sync*` mirror.

## RLS

Org create and org PATCH's override path mirror **inside the same system-scoped context** as the column write. A brand-new org's id is not yet in the caller's `accessible_org_ids`, and a suspended org is invisible to the request's partner context — `contacts` is policed by `breeze_has_org_access(org_id)`, so it would reject the mirror exactly as it would the `organizations` write.

## Bug found and fixed (separate commit)

`isEmpty` judged "no identifying field left" on name/email/phone — the fields the blob models. But `contacts_identifiable_chk` accepts **`mobile`** alone, and `mobile` has no key in the blob. Clearing a billing email on such a contact **deleted the row and cascaded `contact_external_links`**, destroying the re-import identity key; the next PSA/QuickBooks sync would mint a duplicate instead of matching. Not reachable until contact CRUD lands, which is why it is worth fixing before there is data to lose.

## Design question, resolved against my first instinct

`mergeBillingContact` mirrors *the patch*, not the merged blob, so a jsonb key written out-of-band is not reflected in `contacts`. I proposed mirroring the merged result; an adversarial review argued it down and I agree:

- It would make the **deprecated** container the base for the canonical row, which cannot represent `mobile`, `title`, `notes`, `roles`, or `contact_external_links` — the projection could then only shrink canonical state.
- `applyToContactRow` with `replace:false` bases the merge on the **existing contacts row**, so fields survive unless something wrote the blob outside compat — and after this PR nothing does.
- Parity is impossible by construction anyway: the bag holds keys like `quickbooksId` that `contacts` has no column for.

The invariant is a documented one-way projection, not parity. The integration test states this explicitly rather than leaving it to be rediscovered.

## Tests

Existing guarantees were **moved, not relaxed**:

- The atomic `COALESCE(...) || patch` assertion now lives at the layer that owns it (`contacts/compat.test.ts:94`); `invoiceService.test.ts` asserts delegation plus a new 404-not-FK-violation case.
- `invoiceService`'s existence probe is scoped to the contact branch only, so the existing "does NOT select when no contact field is patched" assertion still holds.
- The QuickBooks stub is now keyed by table, so the contacts read no longer consumes a link read and silently shifts the post-violation winner lookup.
- `orgBillingSettings.integration.test.ts` asserts both representations against real Postgres, including the mobile-only survival case.

618 affected unit tests pass; typecheck and lint clean. No database on this machine, so the integration assertions run for the first time in CI.


bdunncompany:
I checked the four `orgImport` writers, because GitHub reports that file as `status: modified` with `additions: 0, deletions: 0, changes: 0` and no patch. They are wired — fetching both blobs and running `diff -a` shows 37 changed lines, the `syncBillingContactRow` / `syncSiteContactRow` import, and the four call sites. No gap.

The reason I had to do that is worth fixing on its own, though: **the NUL byte makes this file's diff invisible in the review UI.** It is not only a `grep -r` problem. GitHub classifies the blob as binary, so a reviewer opening "Files changed" sees the path with an empty diff — 37 changed lines, including four new call sites on a path that reaches the public partner API, render as nothing. `gh pr diff` omits it for the same reason.

It is one byte, at **line 98**, a literal U+0000 sitting inside a JSDoc comment where an escape was clearly intended:

```
   * `${orgId}<U+0000>${system}` for every org that already carries a link under
```

Writing it as `\0` in the comment keeps the meaning and makes the file text again:

```bash
perl -CSD -i -pe 's/\x00/\\0/g' apps/api/src/services/orgImport/index.ts
```

Worth doing in this PR rather than later, since this is the PR whose diff is being hidden. A `git grep -Il '' -- apps/api` sweep would show whether anything else in the tree has picked one up.

Everything else here reads clean to me — in particular the three `sync*`-instead-of-`replace*` sites are justified, and the site PATCH's load-bearing 0-row result is a real constraint that `replaceSiteContactRow` would have silently swallowed.

Issue #3256, opened 2026-08-08
[API] Legacy `maintenance_windows.notify_before` / `notify_on_start` are stored and never read


ToddHebebrand:
## Description

The legacy `maintenance_windows` table carries `notify_before` and `notify_on_start`. They are validated, accepted, and stored by the API, and never read by anything. Same defect as the `config_policy_maintenance_settings` fields covered in #3197, on a different table.

## Current state

- Columns: `apps/api/src/db/schema/maintenance.ts:59-60`.
- Accepted and persisted on create and update: `apps/api/src/routes/maintenance.ts:127-128, 151-152, 412-413, 543-544`.
- **Never read:** `apps/api/src/services/maintenanceService.ts` consumes the window's schedule and suppression flags and never touches either field. No worker, scheduler, or agent command references them.

The config-policy twins (`config_policy_maintenance_settings.notify_before_minutes` / `notify_on_start` / `notify_on_end`, `apps/api/src/db/schema/configurationPolicies.ts:224-226`) have the same problem and are handled in #3197, which removes them from the UI while leaving the columns intact — they are enumerated by hand in two partner canonical-export builders, so dropping the columns would shift existing watermarks.

## Why it matters

An admin who sets "notify before: 30 minutes" on a maintenance window gets nothing, with no indication the setting is inert. That is a promise the product makes and never keeps.

## Proposed

Handle alongside #3207, which is where a real maintenance-window notification consumer would live. Two paths:

1. **Wire them** — a maintenance-window notification tick that warns signed-in users before a window opens, using the same session-broker notify path the reboot manager uses.
2. **Retire them** — drop from the API surface so they stop reading as implemented.

Note for whoever wires them: `config_policy_maintenance_settings.notify_before_minutes` is **nullable**, and there is no zod validator for maintenance inline settings at all — `apps/api/src/services/configurationPolicy.ts:493-510` hand-rolls `typeof` coercion with no range validation, so `notifyBeforeMinutes: -5` is currently accepted. The validator needs to land before the consumer.

## Affected files

- `apps/api/src/db/schema/maintenance.ts` (:59-60)
- `apps/api/src/routes/maintenance.ts` (:127, :151, :412, :543)
- `apps/api/src/services/maintenanceService.ts`

Issue #3197, opened 2026-08-07
[Agent] Patch-triggered reboots never warn the user — warning thresholds are unreachable at the hardcoded 5-minute delay


ToddHebebrand:
## Description

Patch-triggered reboots give the logged-in user **no warning at all**. The agent already contains a full staged-warning implementation, but the delay the API sends makes every warning unreachable.

This is a live user-facing defect: a customer's workstation reboots after patching with zero notice and no chance to save work.

## Root Cause

`agent/internal/patching/reboot_windows.go` implements `RebootManager` with staged warnings at **60 / 15 / 5 minutes** before reboot ("Please save your work"), plus a "Rebooting Now" critical toast. The threshold comparisons use strict `>`.

`apps/api/src/services/patchRebootHandler.ts:77` queues `schedule_reboot` with a hardcoded `delayMinutes: 5`.

At a 5-minute delay, none of the 60/15/5 thresholds fire. The reboot proceeds silently.

`apps/api/src/jobs/maintenanceRebootWorker.ts:39` uses a 15-minute grace, so the maintenance-window path fires only the 5-minute warning.

## Additional findings in the same area

- `agent/internal/patching/reboot_other.go` is a pure no-op stub. **macOS and Linux get no agent-side notification at all**; Linux relies on the `shutdown -r +15` wall message.
- `config_policy_maintenance_settings.notifyBeforeMinutes` and `notifyOnStart` exist in the DB **and** in the UI (`MaintenanceTab.tsx`) but have **no runtime consumer** — they are dead settings that appear functional to an admin.

## Proposed Fix

Minimum, to stop silent reboots:

1. Raise the patch-path delay so at least one warning fires, and/or change the threshold comparison from strict `>` to `>=` so a 5-minute delay still emits the 5-minute warning.
2. Make the delay configurable rather than hardcoded (see the deferral issue for the full policy surface).
3. Either wire `notifyBeforeMinutes` / `notifyOnStart` to a real consumer or remove them from the UI so they stop reading as implemented.

## Affected Files

- `apps/api/src/services/patchRebootHandler.ts` (primary — hardcoded `delayMinutes: 5`)
- `agent/internal/patching/reboot_windows.go` (warning thresholds)
- `apps/api/src/jobs/maintenanceRebootWorker.ts` (15-min grace)
- `agent/internal/patching/reboot_other.go` (macOS/Linux stub)
[...]

PR #3364, opened 2026-08-10
fix(agent): filter BroadcastNotification on the notify scope (#3255)

### LINKED ISSUE #3255: [Agent] `BroadcastNotification` does not filter on the `notify` scope
## Description

`Broker.BroadcastNotification` sends user-facing toasts to **every** connected IPC session rather than only those holding the `notify` scope. The sibling broadcast one function below it does filter. In practice this sprays the Tauri Breeze Assist helper with messages it has no handler for.

Found while investigating #3197. Low severity — the extra recipients discard the message — but it is a scope check that exists everywhere else in the broker.

## Current state

`agent/internal/sessionbroker/broker.go:1281-1296` — `BroadcastNotification` iterates connected sessions and calls `SendNotify("", ipc.TypeNotify, ...)` with no scope predicate.

Compare `BroadcastToDesktopSessions` at `broker.go:1300-1316`, which does filter, and the scope tables at `broker.go:213-214` (`systemHelperScopes` = notify/tray/clipboard/desktop/pam; `userHelperScopes` = notify/clipboard/run_as_user).

The Tauri helper connects with role `assist` / scope `consent_ui`, explicitly **not** `notify` (`agent/internal/ipc/message.go:143`). Its session loop handles `helper_token_update`, `consent_request`, `banner_show`, `banner_hide`, `ping`, `pong`, `disconnect` and falls through on everything else — `apps/helper/src-tauri/src/ipc/client.rs:313-363`, with `_ => { /* assist client handles nothing else */ }` at `:363`.

## Impact

- Notifications are written to sessions that cannot render them. Harmless today, but it means "we broadcast it" is not evidence anyone could have seen it.
- It undercuts the delivery accounting: #3197 changes `BroadcastNotification` to return a dispatch count so `RebootState.NotifiedUser` stops being unconditionally `true` (`agent/internal/patching/reboot_windows.go:170-173`). An unfiltered count includes sessions that structurally cannot display a toast, so the count overstates reach unless this is fixed alongside.

## Proposed fix

Filter on the `notify` scope in `BroadcastNotification`, mirroring `BroadcastToDesktopSessions`. Add a broker test asserting that a session without `notify` receives nothing.

If #3197 lands its dispatch-count change first, this should follow immediately so the count means what it says.

## Affected files

- `agent/internal/sessionbroker/broker.go` (:1281-1296)

ToddHebebrand:
## Problem

`Broker.BroadcastNotification` sent user-facing toasts to **every** connected IPC session with no scope predicate, while the sibling function one below it — `BroadcastToDesktopSessions` — filters on `desktop`. Every other broker path applies a scope gate; this one didn't.

Concretely, the Breeze Assist (Tauri) helper connects with `assist`/`consent_ui` and its session loop has no `TypeNotify` handler at all (`_ => { /* assist client handles nothing else */ }`), and the watchdog helper connects with `watchdog` only. Both were receiving toasts they structurally cannot render.

## Why it matters beyond the wasted round trip

It undercuts delivery accounting. #3197 changes `BroadcastNotification` to return a dispatch count so `RebootState.NotifiedUser` stops being unconditionally `true`. An unfiltered count includes sessions that cannot display anything, so "we broadcast it" would overstate reach. This lands the filter first so that count means what it says.

## Change

`agent/internal/sessionbroker/broker.go`

- Filter on `s.HasScope("notify")` when collecting sessions, mirroring `BroadcastToDesktopSessions` exactly (filter inside the `RLock`, send outside it).
- Replace the discarded `_ = s.SendNotify(...)` with the sibling's `log.Debug` on error, so a failed dispatch leaves a trace instead of vanishing.
- Expanded the doc comment to record *why* the gate is load-bearing.

No behavior change for real notify-capable helpers: both `systemHelperScopes` and `userHelperScopes` include `notify`.

## Scope

Only the scope filtering, per the issue. **Not** in scope and untouched: #3197 (dispatch count / reboot warning thresholds) and #3256 (`maintenance_windows` notify fields never read).

## Tests

New `agent/internal/sessionbroker/broker_notify_test.go`:

| scopes | expected |
|---|---|
| `systemHelperScopes`, `userHelperScopes`, `["notify"]`, `["*"]` | delivered |
| `assistHelperScopes`, `watchdogHelperScopes`, `["desktop","clipboard"]`, `nil` | filtered |

Plus a payload guard (title/body/urgency survive the change) and a blunt regression case asserting nothing hits any wire when only non-notify sessions are connected.

The negative assertion does **not** race a timeout: after the broadcast, a sentinel of a different type is pushed down every session, and the first envelope read is decisive because a session's writes are ordered. `["*"]` is covered because `HasScope` treats it as holding every scope — the broadcast must follow that rule rather than string-matching.

**Verified non-vacuous:** reverting just the filter fails 6 assertions across both tests.

## Verification

- `go test -race ./internal/sessionbroker/...` — ok (11.0s)
- `go test -race ./internal/heartbeat/... ./internal/ipc/... ./internal/patching/...` — all ok (the two `BroadcastNotification` callers live in `heartbeat`)
- `go vet` clean, `gofmt -l` clean
- `go build` for `windows`/`linux`/`darwin` amd64 — clean

Closes #3255

### DIFF
--- a/agent/internal/sessionbroker/broker.go
+++ b/agent/internal/sessionbroker/broker.go
@@ -1277,21 +1277,34 @@ func (b *Broker) TCCStatus() *ipc.TCCStatus {
 	return nil
 }
 
-// BroadcastNotification sends a desktop notification to all connected user sessions.
+// BroadcastNotification sends a desktop notification to every connected session
+// holding the "notify" scope.
+//
+// The scope filter is load-bearing, not cosmetic: helpers that were never
+// granted "notify" have no toast handler at all (the Breeze Assist helper
+// connects with assist/consent_ui and drops unknown message types on the
+// floor). Sending to them wastes an IPC round trip and, worse, would inflate
+// any delivery accounting built on this broadcast into claiming reach it does
+// not have.
 func (b *Broker) BroadcastNotification(title, body, urgency string) {
 	b.mu.RLock()
 	sessions := make([]*Session, 0, len(b.sessions))
 	for _, s := range b.sessions {
-		sessions = append(sessions, s)
+		if s.HasScope("notify") {
+			sessions = append(sessions, s)
+		}
 	}
 	b.mu.RUnlock()
 
 	for _, s := range sessions {
-		_ = s.SendNotify("", ipc.TypeNotify, &ipc.NotifyRequest{
+		if err := s.SendNotify("", ipc.TypeNotify, &ipc.NotifyRequest{
 			Title:   title,
 			Body:    body,
 			Urgency: urgency,
-		})
+		}); err != nil {
+			log.Debug("broadcast notification to session failed",
+				"sessionId", s.SessionID, "error", err.Error())
+		}
 	}
 }
 
--- a/agent/internal/sessionbroker/broker.go
+++ b/agent/internal/sessionbroker/broker.go
@@ -1280,12 +1280,22 @@ func (b *Broker) TCCStatus() *ipc.TCCStatus {
 // BroadcastNotification sends a desktop notification to every connected session
 // holding the "notify" scope.
 //
-// The scope filter is load-bearing, not cosmetic: helpers that were never
-// granted "notify" have no toast handler at all (the Breeze Assist helper
-// connects with assist/consent_ui and drops unknown message types on the
-// floor). Sending to them wastes an IPC round trip and, worse, would inflate
-// any delivery accounting built on this broadcast into claiming reach it does
-// not have.
+// The scope filter is load-bearing, not cosmetic. The Breeze Assist helper
+// connects with assist/consent_ui and the watchdog with "watchdog"; neither
+// has a TypeNotify handler, so an unfiltered broadcast wastes an IPC round
+// trip and — worse — would inflate any delivery accounting built on this
+// broadcast into claiming reach it does not have.
+//
+// Known asymmetry: holding "notify" is sufficient to render a toast but is
+// not, today, strictly necessary. The macOS desktop helper runs the shared
+// internal/userhelper client (which dispatches TypeNotify unconditionally)
+// yet scopesForRole deliberately grants it "desktop" only, so it is filtered
+// out here. That costs nothing at present because neither caller reaches
+// macOS: NewRebootManager discards its NotifyFunc on non-Windows
+// (patching/reboot_other.go) and handleRebootSafeMode precedes a
+// RebootToSafeMode that is Windows-only. If a cross-platform notification
+// caller is ever added, grant the macOS desktop helper "notify" in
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (correctness / locking / cross-platform scope grants) + `pr-review-toolkit:pr-test-analyzer` (coverage, flakiness, resource leaks, stale-assertion sweep).

**Findings:** 4 raised → all addressed. 0 outstanding.

- **Major — doc comment asserted a false invariant.** I'd written that helpers without `notify` "have no toast handler at all." Untrue for the macOS desktop helper: it runs the shared `internal/userhelper` client (which dispatches `TypeNotify` unconditionally) yet `scopesForRole` grants it `["desktop"]` only, so the new filter excludes it. Investigated rather than assumed — **no live regression**: `NewRebootManager` discards its `NotifyFunc` on non-Windows (`patching/reboot_other.go:26`) and `handleRebootSafeMode` precedes a `RebootToSafeMode` that is Windows-only, so neither caller reaches macOS. On macOS the filter actually suppresses a toast announcing a reboot that cannot happen. Kept the strict filter; rewrote the comment to state the real invariant, name the asymmetry, and direct any future cross-platform caller to widen `scopesForRole` rather than weaken this filter (`19195c611`).
- **Minor — macOS scope combination untested.** Pinned `["desktop"]` in the table so the asymmetry cannot drift silently (`19195c611`).
- **Minor — socket leak in the test probe.** Server-side session was never closed; added `t.Cleanup` (`19195c611`).
- **Minor — the new `SendNotify` error branch had no coverage.** Added a case that kills one session before the broadcast and asserts a healthy sibling still receives, proving the loop logs and continues instead of aborting. Carries an explicit non-vacuity guard that fails if a write to the closed session ever succeeds (`afe760d6c`).

**Confirmed clean by review:** `HasScope` takes no lock, so running the filter inside `b.mu.RLock()` has no re-entrancy risk; the sentinel-after-broadcast ordering trick is sound (`ipc.Conn.Send` is atomic under `c.mu`, and both sends are issued from the same goroutine, so program order decides it — no timeout race); no goroutine leaks (`NewSession` starts none, `RecvLoop` is never invoked); and a repo-wide sweep found **no** existing test that asserted the old unfiltered behavior.

**Tests:** `go test -race ./internal/sessionbroker/...` ok (11.9s); `./internal/heartbeat/...` ok (39.0s); `./internal/ipc/... ./internal/patching/...` ok. `-count=3` on the new tests, 3/3 clean, no races. `go vet` + `gofmt -l` clean. `go build` for windows/linux/darwin amd64 clean.

**Non-vacuity:** reverting just the filter fails 6 assertions across the suite.

Issue #3318, opened 2026-08-09
[UI] Quote editor line row clips the last digit of the formatted unit price


ToddHebebrand:
## Description
In the quote editor, a saved line's unit-price input renders `$1,850.0` instead of `$1,850.00` — the last character of the formatted value is clipped. The Total column on the same row shows the full `$1,850.00`, so it reads like a different price. Reproduced with `1850` and `45` (`$45.0`) on a fresh quote.

## Root Cause (suspected)
`EditableLineRow` sizes the price input off the display string and swaps between the raw decimal (focused) and `formatMoney(...)` (unfocused) — `apps/web/src/components/billing/quotes/QuoteLineRows.tsx:823`. The width calculation appears to be one character short for the formatted form, producing exactly the clipping the comment above it (lines 820–822) says the sizing is meant to prevent.

## Proposed Fix
Size the input off the longer of raw/formatted strings plus a one-char buffer, or render the unfocused state as text instead of an input.

## Affected Files
- `apps/web/src/components/billing/quotes/QuoteLineRows.tsx`

## Reported By
Found during marketing-video QA (quote-to-cash re-record, 2026-08-09).

ToddHebebrand:
Fixed in #3370 (not merged yet).

**Root cause — close to the suspicion, but the arithmetic sits one level down.** `growWidth` was sizing the input as a bare `${n}ch` string with a two-character buffer over the displayed value, which looks generous. It isn't: Tailwind preflight sets `box-sizing: border-box`, so that width *includes* the input's own `px-2` padding and 1px borders — 18px, which at the measured **8.96px/ch** is almost exactly the 2ch buffer. The buffer was being spent on chrome rather than glyphs, leaving the field sized to precisely `value.length` ch while the rendered glyphs are wider than `1ch` (Plus Jakarta Sans at 14px: tabular digit 9.60px, `$` 10.34px).

Measured slack on the blurred, formatted value before the fix:

| displayed | old width | slack |
|---|---|---|
| `$45.00` | `8ch` | **0.37px** |
| `$1,850.00` | `11ch` | **3.72px** |

Under half a digit — hence `$45.0` / `$1,850.0` while the Total cell, which is plain text with no box to overflow, showed the full amount.

[...]

PR #3365, opened 2026-08-10
fix(billing): carry the quote line name onto the converted invoice (#3319)

### LINKED ISSUE #3319: [API] Quote→invoice conversion drops the line name — invoice shows only the description
## Description
When an accepted quote converts to an invoice, the invoice line loses the quote line's name. A quote line named "Onboarding & network setup" with description "Network audit, agent deployment, endpoint enrollment" becomes an invoice line labeled only "Network audit, agent deployment, endpoint enrollment" — the customer-facing item name is gone from the invoice detail and (presumably) the PDF.

## Root Cause
The invoice-line mapping in the accept/convert path carries only `description` (`apps/api/src/services/quoteAcceptService.ts:196`); the contract-line mapping right below it (~line 317) preserves `name`. Either `invoice_lines` has no name column or the mapping never composes the two.

## Proposed Fix
Carry the name through — either add/populate a name field on invoice lines, or compose the invoice line description as `name — description` at conversion time so the item label survives.

## Affected Files
- `apps/api/src/services/quoteAcceptService.ts` (primary)
- invoice line schema/render if a name column is added

## Reported By
Found during marketing-video QA (quote-to-cash re-record, 2026-08-09).

ToddHebebrand:
Closes #3319

## Problem

`quote_lines` and `invoice_lines` both gained a `name` (title) column alongside `description` (blurb) in migration `2026-07-03-quote-invoice-line-name`, which documents the renderer contract as `name ?? description` for the title, with `description` shown as a sub-line only when `name` is set.

The quote→invoice conversion writer was never updated to that split. Every converted invoice line was written with `name = NULL`, so it rendered like a legacy line: an accepted quote line named **"Onboarding & network setup"** with description *"Network audit, agent deployment, endpoint enrollment"* became an invoice line labeled only with the description. The customer-facing item name was gone.

## Root cause — two independent gaps, one symptom

**1. Writer (`quoteAcceptService.ts`) — the reported bug.** The invoice-line mapping carried only `description`. Now carries `name: l.name ?? null`.

The Phase 4 contract mapping immediately below it was already correct and is *not* a bug: `NewContractLineSpec` carries a single label by design, so `quoteToContract` deliberately composes `"name — description"` there. Only `invoice_lines` has a real `name` column to populate.

**2. Reader (customer portal) — found while sweeping call sites.** `toCustomerInvoiceLine` collapsed the pair as `description ?? line.name` — the **inverse** of the fallback `invoicePdf.ts` (`lineTitle`/`lineBlurb`) and every MSP web view use. So even with (1) fixed, a customer opening a converted invoice in the self-service portal would still see the blurb and never the title. `CustomerInvoiceLine` now carries both fields and `InvoiceDetailView` renders title/blurb with the same semantics as the PDF, so the portal and the PDF the customer receives label lines identically.

Every other `invoiceLines` writer (`addManualLine`, `addCatalogLine`, `addBundleLine`, the `voidInvoice` reissue clone) was already correct. `addContractLine` and `materializeLines` legitimately have no name to carry.

## Not needed

No migration or schema change — the column already exists, and `invoice_lines` is already registered in `CORE_TENANT_EXPORT_POLICY` with `name` classified in the `included` bucket. No cascade-list change (no new table, no new column).

## Scope

Deliberately untouched: the `apps/web` quote editor, which is being handled separately in #3318.

## Tests

- `quoteAcceptService.test.ts` — 3 new cases pinning the conversion mapping: name preserved distinct from description; a legacy `name: null` line is mirrored rather than fabricated; an `undefined` name normalizes to `null`. A 4th pins the sibling contract composition so a future change can't silently re-split the two paths. **All three fail without the fix** (verified by reverting the one-line change).
- `invoiceService.test.ts` — the exact-keyset portal leak-guard test extended for the new field, plus a new case proving `name`/`description` are surfaced separately.

## Verification

- `vitest run` (single fork) — `invoiceService.test.ts`, `quoteAcceptService.test.ts`, `routes/portal/`, `invoicePdf.test.ts`, `invoiceService.siteScope.test.ts`, `quoteToContract.test.ts`, `contractService.test.ts`: **188 passed**
- `apps/portal` full suite: **89 passed**
- `tsc --noEmit` (apps/api): clean · `astro check` (apps/portal): 0 errors

### DIFF
--- a/apps/api/src/routes/portal/invoices.test.ts
+++ b/apps/api/src/routes/portal/invoices.test.ts
@@ -6,17 +6,19 @@ const { getCustomerInvoiceMock, markViewedMock } = vi.hoisted(() => ({
   getCustomerInvoiceMock: vi.fn(),
   markViewedMock: vi.fn(),
 }));
-vi.mock('../../services/invoiceService', () => ({
-  getCustomerInvoice: getCustomerInvoiceMock,
-  markViewed: markViewedMock,
-  toCustomerInvoiceLine: (line: Record<string, unknown>) => ({
-    description: line.description,
-    quantity: line.quantity,
-    unitPrice: line.unitPrice,
-    taxable: line.taxable,
-    lineTotal: line.lineTotal,
-  }),
-}));
+// toCustomerInvoiceLine is deliberately kept REAL (importActual) — it is the
+// serialization boundary this suite's leak-guard test exists to police. A
+// hand-written stub here would only ever assert itself: it silently drifted
+// from the real field list when `name` was added (#3319), leaving the keyset
+// test documenting a keyset the route never actually returns.
+vi.mock('../../services/invoiceService', async (importActual) => {
+  const actual = await importActual<typeof import('../../services/invoiceService')>();
+  return {
+    getCustomerInvoice: getCustomerInvoiceMock,
+    markViewed: markViewedMock,
+    toCustomerInvoiceLine: actual.toCustomerInvoiceLine,
+  };
+});
 
 const { getInvoicePdfMock, renderInvoicePdfMock } = vi.hoisted(() => ({
   getInvoicePdfMock: vi.fn(),
@@ -158,6 +160,7 @@ describe('portal invoices routes', () => {
       lines: [{
         id: 'internal-line-id', sourceType: 'time_entry', sourceId: 'source-1',
         costBasis: '10.00', revenueAllocation: { labor: '25.00' }, isUnapprovedTime: true,
+        name: 'Support retainer',
         description: 'Support', quantity: '1.00', unitPrice: '25.00', taxable: true, lineTotal: '25.00',
       }],
     });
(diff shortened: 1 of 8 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — code-reviewer + pr-test-analyzer (both on commit `0097eb680`).

**Findings:** 3 raised → all addressed in `dff4da367`; 0 outstanding.

1. *(both reviewers, top finding)* `routes/portal/invoices.test.ts` mocked `toCustomerInvoiceLine` with a hand-written stub, so its "exact safe line keyset" leak-guard only asserted the stub against itself — it silently drifted the moment `name` was added and was documenting a keyset the route doesn't return. Now uses the **real** serializer via `importActual` (the pattern already used for `partnerStripe` in that same file) and asserts the title survives the route boundary.
2. A **blank** `name` was worse than a missing one: renderers derive the blurb from `name` being truthy, so `name: ''` rendered the title as `—` *and* suppressed the description — losing **both** labels. `''` is reachable (line validators are `z.string().max(255).nullable().optional()`, no `.min(1)`), so the writer now blank-normalizes via `l.name?.trim() || null`.
3. `InvoiceDetailView` — the customer-facing surface the bug actually manifested on — had no test file at all. Added one.

**Explicitly cleared by review (no action):** portal/PDF title-blurb logic is byte-identical to `invoicePdf.ts`'s `lineTitle`/`lineBlurb`; the DTO change breaks no other consumer (single consumer, updated in lockstep); the quote-side portal DTO (`CustomerQuoteLine`) already `Pick`s both fields and never had the inverted-precedence bug; all other `invoiceLines` writers already carry `name` correctly; no RLS/cascade/export-policy work (no new table or column).

**Tests:** `apps/api` affected suites 203 passed (`quoteAcceptService`, `invoiceService`, `routes/portal`, single-fork); `apps/portal` full suite 93 passed. The conversion tests were verified **non-vacuous** — reverting the one-line writer fix reds 3 of them. `tsc --noEmit` (apps/api) clean; `astro check` (apps/portal) 0 errors.

**CI:** all required checks green, including `Test API`, `Test Web`, `Lint`, `Check Migrations`, and the `CI Success` gate.

**Status:** review-clean, awaiting maintainer merge. Not merged and #3319 left open per handoff rules.

PR #3366, opened 2026-08-10
fix(webhooks): confine the self-hosted cleartext allowance to private addresses

bdunncompany:
Follow-up to your review of #3320:

> in self-hosted mode plain `http:` is permitted for *any* hostname, not only ones resolving to a private address (`webhookSender.ts:79`) […] if the intent is strictly on-LAN, gate the `http:` allowance on the resolved address actually being private.

The intent is strictly on-LAN, so this gates it.

## Why it was open

`allowPrivateNetwork` swaps the block predicate to `isAlwaysBlockedIp`, which is `isPrivateIp(ip) ? !isRfc1918OrUla(ip) : false` — **false for public addresses**. It opts into private targets without narrowing the scheme, so `http://example.com` was accepted alongside the intended `http://10.0.0.5` and the payload left the box in the clear.

## Where the check goes, and why not in the caller

`SafeFetchInit.requirePrivateForCleartext`. When set, an `http:` target must resolve to RFC1918/ULA; `https:` is untouched; the flag defaults off, so `psa/http.ts` and `dnsProviders` keep today's behaviour byte-for-byte.

It is enforced **inside `safeFetch`, against the record it pins** — not in `webhookSender`. Doing it in the caller means the caller's resolution and `safeFetch`'s own resolution are two separate lookups that can disagree, which reintroduces exactly the TOCTOU window the pinning exists to close. The module comment makes that guarantee explicit, so the check belongs on the same side of it.

## Config-time parity

A URL that will be refused at delivery should fail on save with a clear message, so both validators learned the same rule:

- **literal** address — judged immediately in the sync validator, no DNS needed.
- **hostname** — judged in the DNS variant against the resolved set.
- **mixed answers** — refused. `safeFetch` pins one record, so a rotation containing both `10.0.0.5` and a public address could send a later delivery into the clear even though an earlier one was fine.

## Verification

`vitest` — 40 passed in `webhookSender.test.ts`, 32 in `urlSafety.test.ts`. New cases: public literal rejected over http but accepted over https, hostname resolving public rejected, hostname resolving RFC1918 accepted, mixed answers rejected, and at the `safeFetch` layer public-over-http rejected while private-over-http and public-over-https still succeed.

**Control run:** with the guard removed, `rejects cleartext http to a PUBLIC address` fails; restored, the file is 32/32. The non-vacuousness of the validator cases rests on the same env accepting `http://10.1.2.3/collector` two blocks above.

Full gate: `tsc --noEmit` 0 errors; API suite **21096 passed / 1312 files**; Trivy (secret, misconfig, HIGH+CRITICAL) clean on `apps/api/src/services`.

Four files, +175/−5, no dependency or migration changes.

## One judgement call worth flagging

I read "strictly on-LAN" as the intent and enforced it. That does remove a capability a self-hoster has today — cleartext to a public host — which is defensible to keep for a single-tenant admin-trusted deployment, as you said. If you would rather it stay available, the narrower version is to drop the two config-time validator changes and keep only the `safeFetch` flag unset by default, and I will cut that instead.

### DIFF
--- a/apps/api/src/services/notificationSenders/webhookSender.test.ts
+++ b/apps/api/src/services/notificationSenders/webhookSender.test.ts
@@ -188,6 +188,45 @@ describe('webhook URL safety — self-hosted private-network opt-in', () => {
     });
   });
 
+  describe('cleartext http is confined to the private hop', () => {
+    it('rejects plain http to a PUBLIC literal even when self-hosted', () => {
+      process.env.IS_HOSTED = 'false';
+      // Not vacuous: the same env accepts http://10.1.2.3/collector above.
+      const errors = validateWebhookUrlSafety('http://93.184.216.34/hook');
+      expect(errors.join(' ')).toContain('only use plain http for private');
+    });
+
+    it('still accepts https to that same public literal', () => {
+      process.env.IS_HOSTED = 'false';
+      expect(validateWebhookUrlSafety('https://93.184.216.34/hook')).toEqual([]);
+    });
+
+    it('rejects plain http to a hostname that resolves public', async () => {
+      process.env.IS_HOSTED = 'false';
+      dnsLookupMock.mockResolvedValue([{ address: '93.184.216.34', family: 4 }]);
+      const errors = await validateWebhookUrlSafetyWithDns('http://collector.example/hook');
+      expect(errors.join(' ')).toContain('only use plain http for private');
+    });
+
+    it('accepts plain http to a hostname that resolves RFC1918', async () => {
+      process.env.IS_HOSTED = 'false';
+      dnsLookupMock.mockResolvedValue([{ address: '10.0.0.5', family: 4 }]);
+      await expect(validateWebhookUrlSafetyWithDns('http://collector.lan/hook')).resolves.toEqual([]);
+    });
+
+    it('rejects plain http when the answers MIX private and public', async () => {
+      process.env.IS_HOSTED = 'false';
+      // safeFetch pins one record; a mixed rotation could otherwise send a later
+      // delivery to the public address in the clear.
+      dnsLookupMock.mockResolvedValue([
+        { address: '10.0.0.5', family: 4 },
+        { address: '93.184.216.34', family: 4 }
+      ]);
+      const errors = await validateWebhookUrlSafetyWithDns('http://collector.lan/hook');
+      expect(errors.join(' ')).toContain('only use plain http for private');
+    });
+  });
+
   describe('DNS-resolved targets honour the same policy', () => {
     it('accepts a hostname resolving to RFC1918 when self-hosted', async () => {
       process.env.IS_HOSTED = 'false';
--- a/apps/api/src/services/notificationSenders/webhookSender.ts
+++ b/apps/api/src/services/notificationSenders/webhookSender.ts
@@ -6,7 +6,13 @@
  */
 
 import { isIP } from 'net';
-import { isAlwaysBlockedIp, isPrivateIp, safeFetch, SsrfBlockedError } from '../urlSafety';
+import {
+  isAlwaysBlockedIp,
+  isPrivateIp,
+  isRfc1918OrUla,
+  safeFetch,
+  SsrfBlockedError
+} from '../urlSafety';
 import { selfHostAllowsPrivateNetwork } from '../../config/env';
 import { getOutboundHeaderValidationErrors, sanitizeOutboundHeaders, validateOutboundHeader } from '../outboundHeaders';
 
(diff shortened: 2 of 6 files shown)


ToddHebebrand (changes requested):
**The in-diff logic is sound.** `apps/api/src/services/urlSafety.ts:304` runs `isRfc1918OrUla` against `safeRecord.address` — the record that then gets pinned into `pinnedLookup` — so a rebind can't slip past between validation and connect. `isRfc1918OrUla` already normalizes both `::ffff:` dotted and hex-pair forms, and 169.254.x, `::1`, `0.0.0.0`, CGNAT and multicast stay blocked by the unchanged `isAlwaysBlockedIp` gate. Self-hosted literals and hostnames resolving wholly to RFC1918/ULA still work. No schema changes, so no RLS or cascade obligations.

**Must-fix — the parity claim in the PR body isn't true for the larger webhook system.** `apps/api/src/workers/webhookDelivery.ts:134` calls `safeFetch` with neither `allowPrivateNetwork` nor `requirePrivateForCleartext`. On a self-hosted install, `http://10.0.0.5/hook` passes save-time validation at `apps/api/src/routes/webhooks.ts:402` and then fails every single delivery with `URL points to blocked address`. That's pre-existing from #3320, but this PR is the one asserting that save-time and delivery-time agree — so it should either carry the flags through there or drop the claim.

**Must-fix — same file, `webhookDelivery.ts:84`** re-validates through `validateWebhookUrlSafetyWithDns`, which performs a *separate* DNS resolution from the one `safeFetch` does. That is exactly the TOCTOU split the PR body argues against; the new flag is the right fix for it.

**Worth knowing, not blocking:** the identical cleartext-to-public hole exists in `apps/api/src/services/sso.ts:475` (plus `routes/sso.ts:1000,1171,1429`), which carries OIDC `client_secret`, codes and tokens, and in `psa/http.ts:60` and `dnsProviders/http.ts:96`. The new flag fixes one of four callers — the other three are worth a follow-up issue rather than scope creep here. Separately, `apps/api/src/services/ssrfGuard.ts:151` is a second, weaker private-range implementation (`on-prem-http`); converging on one is better than growing both.

**Tests** are IPv4-only — no ULA accept-case (`http://[fd00::1]/`) and no `::ffff:`-mapped-public reject-case, which are the two forms most likely to regress.

Status: requesting changes on the two `webhookDelivery.ts` items. The core `urlSafety.ts` change is good as-is.

bdunncompany:
Both must-fixes done in `bb011363a`. You were right that the parity claim didn't hold for the larger webhook system — I'd fixed the notification-channel sender and asserted a property about a delivery path I hadn't touched.

## 1. Flags carried into the delivery worker

`workers/webhookDelivery.ts` now passes the same pair the channel sender uses:

```ts
allowPrivateNetwork: selfHostAllowsPrivateNetwork(),
requirePrivateForCleartext: true
```

So a self-hosted install that saves `http://10.0.0.5/hook` can now actually deliver to it, instead of passing save-time validation and then failing every attempt with `URL points to blocked address` — while the allowance stays confined to the LAN hop rather than the open internet.

## 2. The duplicate DNS re-validation is gone

Removed the `validateWebhookUrlSafetyWithDns` pre-flight entirely rather than threading flags into it. Two lookups that can disagree is the whole problem, and the second one bought nothing that `safeFetch` doesn't already enforce against the record it pins: `urlSafety.ts:258` rejects non-`http(s)` schemes, the blocked-range gate is unchanged, and the cleartext rule is now applied there too.

Error reporting is unaffected — the catch already mapped `SsrfBlockedError` to the same `Unsafe webhook URL: …` prefix the pre-check produced, so a rejected URL reports identically. That's now pinned rather than assumed.

## Test changes, including one that had to flip

The existing `returns an unsafe-url failure when safeFetch blocks rebinding to private networks` asserted `validateWebhookUrlSafetyWithDns` **had** been called. Its other two assertions still passed untouched, which is the evidence the failure path is preserved; only the "we re-resolved first" assertion was invalidated, so it now asserts `not.toHaveBeenCalled()` — pinning that the split stays closed rather than silently deleting the check.

Two new delivery cases: one that no second resolution happens on a successful delivery (`safeFetch` called exactly once), and one that both flags actually reach the call.

**The IPv6 gaps you flagged are both covered now:**

- `http://[fd00::1]/` — ULA accept-case.
- `::ffff:93.184.216.34` — mapped-public reject-case.

Worth reporting how the ULA case first failed, because it's the more reassuring outcome: it errored `connect EHOSTUNREACH fd00::1:80`, i.e. the guard **passed it** and the test then attempted a real socket. `isRfc1918OrUla` was already correct for ULA; my test just lacked the `http.request` stub its RFC1918 sibling uses. Fixed the test, not the guard.

**Control run:** with the gate at `urlSafety.ts:304` neutered, both reject-cases fail, including the new mapped-address one. Both accept-cases pass in either state, as they should.

## Your non-blocking items

Filed nothing yet, deliberately — flagging where I landed so you can redirect:

- **`sso.ts:475` plus `routes/sso.ts:1000,1171,1429`, `psa/http.ts:60`, `dnsProviders/http.ts:96`.** Agreed these are the same hole and that SSO is the worst of them, since it carries `client_secret`, codes and tokens. I'd rather open one issue covering all three callers than widen this PR — say the word and I'll write it up with the same per-caller verification I did here.
- **`ssrfGuard.ts:151` as a second, weaker private-range implementation.** Agreed converging beats growing both. That's a refactor with its own blast radius across every caller of the weaker one, so it wants to be its own change with its own control run, not a rider here.

## Verification

`tsc --noEmit` exit 0. Full API suite **1312 files / 21100 passed**, 37 skipped, exit 0. Trivy `secret,misconfig` HIGH/CRITICAL on `apps/api/src` clean. Three files, +88/−15. No schema, migration or dependency changes.

PR #3367, opened 2026-08-10
fix(agent): bound total shutdown time so systemd stops cleanly (#3323)

### LINKED ISSUE #3323: Agent shutdown stage timeouts sum to more than systemd's TimeoutStopSec=15s, causing SIGKILL on stop/restart
Reported by **rgadagno7123** in #agent-linux (Discord, message 1535669154475024514).

**Report (PII-redacted):**

trying to install on a Ubuntu 24.04.4 LTS and getting an error of topping breeze-agent.service - Breeze RMM Agent...
breeze-agent.service: State 'stop-sigterm' timed out. Killing.
breeze-agent.service: Killing process 8482 (breeze-agent) with signal SIGKILL.
breeze-agent.service: Failed to kill control group /system.slice/breeze-agent.service, ignoring: Invalid argument
breeze-agent.service: Main process exited, code=killed, status=9/KILL
breeze-agent.service: Failed with result 'timeout'.
Stopped breeze-agent.service - Breeze RMM Agent.
Started breeze-agent.service - Breeze RMM Agent.
time=2026-08-08T15:15:28.971Z level=INFO msg="sentry disabled (no DSN set)"

## Triage analysis

The reported systemd sequence (SIGTERM sent, 'State stop-sigterm timed out. Killing.', SIGKILL, then the service starts again) is the expected systemd behavior when the agent's graceful-shutdown path (`shutdownAgent` in agent/internal/agentapp/main.go) takes longer than the unit's `TimeoutStopSec=15` (agent/service/systemd/breeze-agent.service:42). `shutdownAgent` runs each teardown stage sequentially via `runWithTimeout`, so the per-stage caps are additive, not concurrent (verified: runWithTimeout blocks on `select { <-done; <-time.After(d) }` before the next stage is even started). On a typical enrolled Linux agent (etwlua and the Windows watchdog supervisor are both no-ops on Linux, so those two are skipped), the remaining stages are: UniFi collector stop (2s) + workspace-index stop (2s) + drain in-flight commands (5s) + websocket stop (3s) + heartbeat stop (5s) = 17s worst case — already 2s over the unit's 15s `TimeoutStopSec`, before even counting the state-file write and watchdog notify steps. The code comment at main.go:433-434 explicitly claims 'Total worst case is bounded by the sum of per-stage timeouts; the unit file caps the outer wait at 15s,' but the actual sum exceeds that cap, so on any restart/stop where a stage runs close to its own budget, systemd will legitimately time out the SIGTERM and escalate to SIGKILL exactly as the poster observed. The subsequent 'Started breeze-agent.service' line shows the new instance came up fine, so this looks like a benign-looking but real timing bug rather than a fatal install failure.

## Affected code

- `agent/internal/agentapp/main.go:429-532` — shutdownAgent runs unifi(2s), workspaceIndex(2s), drain(5s), websocket(3s), heartbeat(5s) sequentially via runWithTimeout — worst case 17s, exceeding the unit's 15s stop timeout
- `agent/internal/agentapp/main.go:538-549` — runWithTimeout blocks the caller until fn finishes or its own deadline elapses, so stage timeouts add up rather than overlapping
- `agent/service/systemd/breeze-agent.service:38-43` — TimeoutStopSec=15 with KillMode=mixed — SIGKILLs the process/cgroup if SIGTERM shutdown doesn't finish in 15s
- `agent/internal/agentapp/main.go:1048-1057` — runAgent's SIGTERM/SIGINT handling via signal.NotifyContext calls shutdownAgent on systemctl stop/restart

## Potential fix

Either raise TimeoutStopSec in the systemd unit to comfortably exceed the sum of the shutdown stage budgets (e.g. 20-25s), or run the unifi/workspaceIndex/drain/websocket/heartbeat stop stages concurrently (fan-out with a single outer deadline) so the real worst case fits within the current 15s window, and correct the comment claiming the two are already aligned.

ToddHebebrand:
Closes #3323

## The bug

`shutdownAgent` tears components down **sequentially** — `runWithTimeout` blocks the caller before the next stage starts — so its per-stage timeouts are **additive**, not concurrent. They summed to **17s** on an ordinary enrolled Linux agent while the systemd unit capped the stop at `TimeoutStopSec=15`, so systemd escalated SIGTERM to SIGKILL on routine stops and restarts. `KillMode=mixed` then took the whole cgroup with it, leaving helpers, tunnels and the audit log unclosed.

Confirming @bdunncompany's tightening of the arithmetic: the three **ungated** stages alone (drain 5s + websocket 3s + heartbeat 5s) are a **13s floor**, so a plain agent had 2s of margin, one with a UniFi collector landed **exactly on the cap**, and one that also indexed a workspace went **over**. That is why the failure reads as intermittent rather than deterministic.

The comment at `main.go` — *"Total worst case is bounded by the sum of per-stage timeouts; the unit file caps the outer wait at 15s"* — was literally true and still wrong: nothing checked that the sum actually fit under the cap.

## Two more unbounded steps in the same function

The stage arithmetic wasn't the whole story. Two steps in the shutdown path had no budget at all, and each can outlast the entire stop window on its own — on exactly the hosts this issue is about, whose network or IPC peer died during OS shutdown:

1. **The watchdog "shutdown intent" IPC notify** (`main.go`) is a blocking socket write, and `ipc.Conn.Send` arms a **30s** write deadline (`internal/ipc/protocol.go:42`). A watchdog that stopped reading its end parks shutdown before a single core teardown stage begins — enough to re-create this SIGKILL even at `TimeoutStopSec=30`.
2. **`runAgent`'s deferred `logging.StopShipper()`**, which runs *after* `shutdownAgent` returns. `Shipper.Stop()` is `close(stopChan); wg.Wait()` with no deadline of its own, and the shipper's `http.Client` has a **30s** timeout (`internal/logging/shipper.go:113`).

Both are now bounded. Abandoning the notify is safe: the stopping-state file written immediately above it is the durable signal the watchdog reconciles against, and the notify is only the fast path.

## The fix

Two layers, and the issue discussion supports both:

**1. Agent-side total budget (the portable half).** Every stage draws from one shared `shutdownBudget`, running for `min(its own budget, time remaining)`. A stage added later can never push the total past the cap. Bounding agent-side is what makes the guarantee portable: launchd and the Windows SCM impose their own stop deadlines too, and neither is fixed by touching a systemd unit.

Stage **order is unchanged**. Per @bdunncompany's point, drain stays ahead of the two transport stops — draining after the websocket and heartbeat are torn down would strand whatever is in flight. That ordering constraint is precisely why this bounds the stages rather than fanning them out concurrently: the concurrent version is the design call the issue flagged as risky, and the budget gets the same guarantee without it. The optional, platform/config-dependent component stops share a sub-budget so one wedged optional component can't starve the ungated core teardown.

An exhausted budget means **"start the stage and don't wait"**, never "skip it" — `Heartbeat.Stop` and websocket `Client.Stop` are `sync.Once`-guarded, so declining to call one forgoes that teardown permanently rather than deferring it. A stage granted less than its nominal budget is logged as squeezed, since that's the first visible symptom of the shared budget running tight.

**2. `TimeoutStopSec` 15 → 30, deliberately ABOVE the agent's budget rather than equal to it.** The slack covers the parts of shutdown that stay intentionally untimed (stopping-state write, secure-token zeroing, process exit) plus systemd's own signal latency. Setting it *equal* to the agent budget would just recreate a thin-margin version of the same bug. `currentUnitVersion` bumps 4 → 5 so deployed hosts pick it up via `reconcileServiceUnitIfNeeded` on the next agent start.

Worst case is now bounded and **clean** rather than unbounded and killed; in practice these stages complete in milliseconds, since they only consume budget when something is genuinely wedged.

## Why it can't silently regress

The arithmetic is asserted, not narrated. `TestShutdownBudgetFitsUnitTimeoutStopSec` parses `TimeoutStopSec` **out of the embedded unit string** and asserts `shutdownBudget + shipperFlushBudget + shutdownTailSlack` fits under it. `TestShutdownStageBudgetsFitShutdownBudget` asserts the stage budgets fit the shared budget. Both would have failed on `main` before this change.

For the same reason, `windowsStopTimeout`'s comment in `internal/watchdog/recovery_windows_logic.go` no longer restates the agent's max shutdown budget as a hardcoded "~21s" — it names the constants. A hardcoded derived number going stale is precisely this issue.

## Files

- `agent/internal/agentapp/shutdown_budget.go` (new) — budget constants + `shutdownClock`, with the full timing contract documented
- `agent/internal/agentapp/main.go` — stages draw from the shared clock; watchdog notify and deferred shipper flush bounded; misleading comment corrected
- `agent/internal/agentapp/systemd_unit.go` + `agent/service/systemd/breeze-agent.service` — `TimeoutStopSec=30`, unit version 5 (regenerated byte-identically; `TestStaticUnitMatchesEmbedded` holds)
- `agent/internal/agentapp/shutdown_budget_test.go` (new) — the two arithmetic contracts plus clamping, shared-deadline enforcement, start-don't-skip on an exhausted budget (both `run` and `runCtx`), sub-budget tightening, squeeze reporting, and the drain context's deliberate stage-cap/ctx ordering
- `agent/internal/watchdog/recovery_windows_logic.go` — comment only

## Verification

`cd agent && go test -race ./...` — **entire agent suite green**. `go vet` clean; `golangci-lint run --new-from-rev=origin/main ./...` reports 0 issues.

## Notes for review

- **Fleet impact:** the unit-version bump means each Linux host rewrites its unit and restarts the agent once, on the next agent start, via the existing `systemd-run` reconcile escape (the same mechanism used for v3 and v4).
- The self-uninstall backstop in `agent/internal/heartbeat/handlers_uninstall.go` (fires at t≈17s, `os.Exit(0)` at t≈22s) was tuned against the old ceiling. It is **not** a regression — before this change a wedged self-uninstall was SIGKILLed at t≈20s instead, so both truncate, and the normal path completes by t≈6s — but tying that backstop to the shutdown budget would be a reasonable follow-up. It lives in package `heartbeat`, which `agentapp` imports, so sharing the constant needs a small refactor rather than a one-line change; left out of scope here.
- The two dated design docs under `docs/superpowers/{plans,specs}/agent/2026-06-09-*` quote the old `TimeoutStopSec=15` unit. Left as-is — historical records of that design, not live config.

### DIFF
--- a/agent/internal/agentapp/main.go
+++ b/agent/internal/agentapp/main.go
@@ -556,10 +556,21 @@ func shutdownAgent(comps *agentComponents) {
 	}
 
 	// Notify the watchdog of intentional shutdown so it doesn't restart us.
+	//
+	// Budgeted, because this is a blocking socket write: ipc.Conn.Send arms a
+	// 30s write deadline (internal/ipc/protocol.go), so a watchdog that has
+	// stopped reading its end can park this call for longer than the entire
+	// stop window on its own — the same class of unbounded step as the log
+	// shipper, and enough to exhaust the whole budget before a single core
+	// teardown stage starts. Abandoning it is safe: the stopping-state file
+	// written just above is the durable signal the watchdog reconciles
+	// against, and this notify is only the fast path.
 	if broker := comps.hb.SessionBroker(); broker != nil {
 		if sess := broker.PreferredSessionWithScope("watchdog"); sess != nil {
-			_ = sess.SendNotify("", ipc.TypeShutdownIntent, ipc.ShutdownIntent{
-				Reason: state.ReasonUserStop,
+			clock.run("watchdog shutdown notify", watchdogNotifyBudget, func() {
+				_ = sess.SendNotify("", ipc.TypeShutdownIntent, ipc.ShutdownIntent{
+					Reason: state.ReasonUserStop,
+				})
 			})
 		}
 	}
@@ -577,6 +588,15 @@ func shutdownAgent(comps *agentComponents) {
 	clock.run("websocket stop", websocketStopBudget, comps.wsClient.Stop)
 	clock.run("heartbeat stop", heartbeatStopBudget, comps.hb.Stop)
 
+	// Zeroed unconditionally, including while an abandoned teardown goroutine
+	// may still be running. That goroutine can then read an empty bearer token
+	// and log a single warning instead of authenticating. This is accepted, and
+	// predates the shared budget — runWithTimeout has always abandoned an
+	// overrunning stage and fallen through to here. Wiping the secret is a
+	// hard guarantee at a fixed point in shutdown; the call it degrades is a
+	// best-effort one on a stage we already gave up waiting for, moments
+	// before the process exits. Deferring the wipe to chase it would trade a
+	// security property for a network call that is being dropped anyway.
 	if comps.secureToken != nil {
 		comps.secureToken.Zero()
 	}
--- a/agent/internal/agentapp/main.go
+++ b/agent/internal/agentapp/main.go
@@ -468,13 +468,23 @@ type agentComponents struct {
 // shutdownAgent gracefully stops all agent components.
 //
 // Every blocking stage is wrapped with a deadline so that a stuck HTTP flush
-// (common during OS shutdown when the network has already gone down) can't
-// pin the process past systemd's TimeoutStopSec. Total worst case is bounded
-// by the sum of per-stage timeouts; the unit file caps the outer wait at 15s.
+// (common during OS shutdown when the network has already gone down) can't pin
+// the process past the service manager's stop timeout. Stages run
+// sequentially, so their timeouts are additive — they are therefore drawn from
+// one shared shutdownBudget rather than each being independent, and the whole
+// function is bounded by that budget no matter how many stages apply. See the
+// shutdown timing contract in shutdown_budget.go for how the budget relates to
+// the unit's TimeoutStopSec (#3323).
 func shutdownAgent(comps *agentComponents) {
 	if comps == nil {
 		return
 	}
+
+	clock := newShutdownClock(shutdownBudget)
+
+	// The optional, platform/config-dependent component stops share a
+	// sub-budget so they cannot starve the ungated core teardown below.
+	components := clock.sub(componentStopBudget)
 
 	// Cancel the ETW LUA subscriber FIRST so the kernel-side ETW
 	// session is closed before any later teardown can time out and
@@ -482,10 +492,14 @@ func shutdownAgent(comps *agentComponents) {
 	// the kernel and the next agent restart hits the
 	// "two callers on the same machine would conflict" failure from
 	// NewETWSubscriber's doc comment.
+	//
+	// The cancels are issued unconditionally (they are non-blocking); only
+	// the wait-for-exit is budgeted. That way a component always gets told to
+	// stop even when the sub-budget has run out and we can't wait for it.
 	if comps.etwluaCancel != nil {
 		comps.etwluaCancel()
 		if comps.etwluaDone != nil {
-			runWithTimeout("etwlua stop", 2*time.Second, func() {
+			components.run("etwlua stop", componentStopStage, func() {
 				<-comps.etwluaDone
 			})
 		}
@@ -497,7 +511,7 @@ func shutdownAgent(comps *agentComponents) {
 	if comps.unifiCancel != nil {
 		comps.unifiCancel()
 		if comps.unifiDone != nil {
-			runWithTimeout("unifi collector stop", 2*time.Second, func() {
+			components.run("unifi collector stop", componentStopStage, func() {
 				<-comps.unifiDone
 			})
 		}
@@ -505,13 +519,13 @@ func shutdownAgent(comps *agentComponents) {
 
 	// Cancel workspace indexing before core network teardown. A canceled crawl
 	// attempts one terminal CompleteRun, but we do NOT wait out its 30s HTTP
-	// timeout — the whole shutdown must fit systemd's 15s TimeoutStopSec, and
-	// a missed terminal flush self-heals server-side (the stale run is marked
-	// abandoned on the next crawl start). 2s matches the sibling stages.
+	// timeout — the whole shutdown must fit the shared budget, and a missed
+	// terminal flush self-heals server-side (the stale run is marked abandoned
+	// on the next crawl start). The cap matches the sibling stages.
 	if comps.workspaceIndexCancel != nil {
 		comps.workspaceIndexCancel()
 		if comps.workspaceIndexDone != nil {
-			runWithTimeout("workspace index stop", 2*time.Second, func() {
+			components.run("workspace index stop", componentStopStage, func() {
 				<-comps.workspaceIndexDone
 			})
 		}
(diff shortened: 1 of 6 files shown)


bdunncompany:
CI is red on one staticcheck line, nothing else:

```
internal/agentapp/shutdown_budget_test.go:64:2: ST1011: var timeoutStopSec is of type
time.Duration; don't use unit-specific suffix "Sec" (staticcheck)
	timeoutStopSec := unitTimeoutStopSec(t)
```

It's flagging the local at line 64, not the helper — so renaming just that variable (`timeoutStop`, or `stopBudget`) clears it. `Lint Agent (Go)` runs new-issues-only, so this is the sole finding and the job goes green with the rename.

ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter on 8ebb3c85a, then a second code-reviewer round on the fix commit 2a43d0193 (the fix touched agent-shipped code, so it earned a re-review).

**Findings:** 4 raised, all addressed.

- **Once-guarded stop could be forgone permanently** *(should-fix)* — an exhausted budget skipped the stage outright. Fine for a pure wait, but the core stages pass their whole stop function as `fn`, and `Heartbeat.Stop` / websocket `Client.Stop` are `sync.Once`-guarded: a `Once` never fires twice, so declining to call one forgoes that teardown rather than deferring it, leaving the session broker, helper lifecycle and audit log open. Out of budget now means *"start it and don't wait"*, never *"skip it"* — `2a43d0193`.
- **Watchdog IPC notify was unbudgeted** *(found while verifying the reviewer's reachability claim — and the reason the exhausted path is reachable at all)*. It's a blocking socket write and `ipc.Conn.Send` arms a **30s** write deadline (`internal/ipc/protocol.go:42`), so a watchdog that stopped reading its end parks shutdown before a single core stage starts — enough to re-create this SIGKILL even at `TimeoutStopSec=30`. Now budgeted; abandoning it is safe because the stopping-state file written immediately above is the durable signal the watchdog reconciles against — `2e4e385de`.
- **Squeezed stages were invisible** *(nit)* — a stage granted less than its nominal budget but still completing was indistinguishable in the logs from one that ran with its full allotment. Now logged; a squeeze is the first visible symptom of the shared budget running tight — `2a43d0193`.
- **`secureToken.Zero()` vs. an abandoned teardown goroutine** *(should-fix)* — verified **not** a data race (`SecureString` is mutex-guarded) and **not** new: `runWithTimeout` has always abandoned an overrunning stage and fallen through to the wipe. Accepted and documented rather than deferred — the wipe is a hard security guarantee at a fixed point, and delaying it would trade that for a best-effort network call on a stage we already gave up waiting for, moments before the process exits — `2e4e385de`.

Also fixed the `Lint Agent (Go)` failure from the first push (staticcheck ST1011: a `time.Duration` var with a unit-specific `Sec` suffix), and widened one test's real-time window 50ms → 2s so a loaded runner can't flake it.

**Not changed, flagged instead:** the self-uninstall backstop in `agent/internal/heartbeat/handlers_uninstall.go` was tuned against the old ceiling. It is **not** a regression — a wedged self-uninstall was SIGKILLed at t≈20s before and now hits the backstop's `os.Exit(0)` at t≈22s, so both truncate, and the normal path completes by t≈6s. Tying it to the shutdown budget is a reasonable follow-up but needs a small refactor: it lives in package `heartbeat`, which `agentapp` imports, so the constant can't be shared as a one-liner.

**Tests:** full `go test -race ./...` across the agent module green on the final tree (not just the affected packages); `go vet` clean; `golangci-lint run --new-from-rev=origin/main ./...` reports 0 issues. The new timing tests also survive `-count=15` and 6-way parallel CPU contention locally. On the previous push, CI's `Test Agent`, `Test Agent (race)`, `Test Agent (Windows)` and `Lint Agent (Go)` all passed; the final push is re-running and green so far with 0 failures.

**Status:** review-clean, awaiting maintainer merge. Worth noting for merge timing: the `currentUnitVersion` 4 → 5 bump means each Linux host rewrites its unit and restarts the agent once on the next agent start, via the existing `systemd-run` reconcile escape.

PR #3368, opened 2026-08-10
fix(api): make search_logs keyset pagination usable (#3329)

### LINKED ISSUE #3329: [API][MCP] search_logs returns HTTP 500 when its own nextCursor is passed back — advertised pagination is unusable
## Summary

`search_logs` advertises keyset pagination and returns a well-formed `nextCursor`, but passing that cursor back to the same tool returns **HTTP 500**. The advertised pagination path cannot be used at all.

## Reproduce

Call `search_logs` over the MCP endpoint with any window that has more rows than `limit`:

```json
{"timeRange": {"start": "...", "end": "..."}, "limit": 15, "countMode": "none"}
```

Page 1 answers normally:

```
keys:   ["hasMore","limit","logs","nextCursor","offset","showing","total","totalMode"]
rows:   15
hasMore: true
nextCursor: eyJ0aW1lc3RhbXAiOiIyMDI2LTA4LTEwVDA2OjE0OjQyLjAwMFoiLCJpZCI6...
```

The cursor is well-formed — base64 of `{"timestamp": "...", "id": "..."}`.

Send it straight back, unmodified:

```json
{"timeRange": {"start": "...", "end": "..."}, "limit": 15, "countMode": "none", "cursor": "<nextCursor>"}
```

```
HTTP Error 500: Internal Server Error
```

Also 500s with the cursor alone, i.e. no `timeRange` alongside it:

```json
{"limit": 15, "cursor": "<nextCursor>"}
```

Verified against 0.103.0 on 2026-08-10. The client echoes back exactly what the server issued.

## Impact

Any consumer that pages `search_logs` — log export, SIEM forwarding, an agent walking more than one page of device events — fails on the second page. `hasMore: true` actively invites the call that breaks.

Since the response also switches to a summarised `{_chat, preview, summary}` envelope above a response-size budget (measured: rows at `limit<=15`, summarised at `limit>=20`), raising `limit` is not an alternative route to the same rows. Between the two behaviours there is currently no way to read more than ~15 rows out of a window.

`search_agent_logs` and `query_change_log` expose no cursor at all, so the only workaround is walking `endTime` backwards past the oldest row of each page.

## Note on the size-based envelope

Related but separate, and possibly intended: past that size budget the tools return

```json
{"_chat": {...}, "preview": ..., "summary": ...}
```

with no row array. It is not an error response, so a client that does not specifically look for those keys will happily store the digest as though it were one record. `manage_alerts` differs again — it returns `_chat` *alongside* a real `alerts` array. If the summarisation is deliberate, a documented flag to opt out (or an explicit `truncated: true` on the row-shaped response) would make it safe to consume programmatically.

ToddHebebrand:
Closes #3329

## The bug

`search_logs` advertised `hasMore: true` alongside a well-formed `nextCursor`, but sending that cursor straight back returned **HTTP 500**. The advertised pagination path could not be used at all — and because `hasMore: true` invites the call, every consumer that pages (log export, SIEM forwarding, an agent walking more than one page) hit it on page 2.

This is not MCP-specific: the REST `POST /logs/search` endpoint shares `searchFleetLogs`, so it had the identical failure.

## Root cause

The keyset predicate in `runFleetSearch` was a hand-written `sql` template that interpolated the cursor's `Date` directly:

```ts
sql`(
  ${deviceEventLogs.timestamp} < ${cursor.timestamp}
  OR (${deviceEventLogs.timestamp} = ${cursor.timestamp} AND ${deviceEventLogs.id} < cast(${cursor.id} as uuid))
)`
```

A bare value inside a Drizzle `sql` template is wrapped in a `Param` with the **noop** encoder, so the raw `Date` object was handed to postgres.js, whose Bind step throws:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string
or an instance of Buffer or ArrayBuffer. Received an instance of Date
    at Bind (postgres/src/connection.js:954)
```

The surrounding `gte`/`lte` time-range conditions were fine precisely because they go through Drizzle's typed helpers, which serialize the `Date` via the **column's** encoder.

Why a 500 rather than a tool-level error: every request runs inside `withDbAccessContext`, a postgres.js `begin()` transaction which re-throws any query error at commit *even after the application handled it*. `search_logs`' own `catch` did return `{ error: ... }`, but the transaction wrapper then threw the raw driver error past it into Hono's global handler.

## The fix

The predicate is now built from Drizzle's typed `lt`/`gt`/`eq` helpers in an exported `buildLogSearchKeysetCondition`, so cursor values are encoded by the column that receives them.

Cursor timestamps stay millisecond-precision, which is **lossless** here rather than merely tolerable: `device_event_logs.timestamp` is agent-supplied and normalized through `new Date(...)` in `sanitizeTimestamp` (`routes/agents/helpers.ts`) on ingest, so the column cannot hold sub-millisecond values. Confirmed against the dev database — 0 of 580 rows have sub-millisecond precision. That matters because the `timestamp = cursor` tiebreaker branch would silently skip rows if the cursor were coarser than the column.

## Verification against a real database

Paging the full dev fleet with `limit: 7`, then diffing against a single unpaginated query:

```
sortOrder=desc pages=83 paged=580 oneShot=580 dupes=0 orderIdentical=true
sortOrder=asc  pages=83 paged=580 oneShot=580 dupes=0 orderIdentical=true
```

No gaps, no duplicates, ordering byte-identical to the one-shot result, in both sort directions.

## Also in this change

**Cursor rejections are now legible.** `Invalid cursor format.` / `Invalid cursor payload.` are added to the AI tool error allowlist. Cursors are opaque to the caller, so the generic *"the tool could not complete this request"* left a paging client unable to distinguish a stale cursor from a server fault, with no recovery path. Both are fixed string literals with nothing interpolated, and the driver-detail detector still vets them — a test asserts that appending driver text to the literal falls back to the generic message.

**Row-less digests are now self-describing.** @bdunncompany's follow-up report flagged a second, compounding behaviour: above a response-size budget the tools return `{_chat, preview, summary}` with **no row array**, as HTTP 200 with no error field, so a poller stores the digest as if it were a record. `_chat` alone is not a usable detector, because it also rides along with payloads whose rows are intact — `manage_alerts` returns `_chat` right next to a real `alerts` array. `compactToolResultForChat` now sets `summarized: true` on exactly the two branches that *replace* the payload, and the contract is documented on the function. The three outcomes are now distinguishable with one key.

## Scope notes

- **Not fixed here — the digest opt-out flag.** The report also asked for a way to opt out of summarisation. That is a real MCP surface decision (budget per tool? per call? interaction with the model's own context limits) rather than a bug fix, so it is left for a follow-up.
- **Not fixed here — why `search_logs` digested at 20-minute windows.** I characterised the compaction path empirically: sweeping row counts 5→1000 and message lengths 80→200,000 chars, the tightest tier always caps `logs` at 10 rows / 300-char messages and lands ~7.2 KB, under the 8 KB ceiling. Total row loss required a payload of almost entirely JSON-escapable characters. So the reported 20-minute-window digest is **not** reproducible from row count or message length alone, and the `summarized` flag makes it detectable rather than explained. Worth pinning down with the reporter.
- **Two more instances of this same defect class exist elsewhere** and are deliberately *not* bundled into this PR (unrelated subsystems, one auth-adjacent, each wants its own test): `services/filterEngine.ts` `applyOperator` (datetime `before`/`after`/`between`/`equals` on `POST /devices/filters/preview`, dynamic groups, deployment targeting) and `oauth/revocationRetry.ts`. Filed separately and linked below.

## Tests

- `logSearch.test.ts` — cursor encode/decode round-trip incl. millisecond fidelity, standard-base64 tolerance, malformed-cursor rejection; and a regression guard that renders the keyset condition through `PgDialect.sqlToQuery` and asserts **no param is a `Date`**. That guard fails on the old code without needing a database.
- `aiToolErrors.test.ts` — cursor messages preserved; allowlist cannot smuggle driver detail.
- `aiToolOutput.test.ts` — `summarized` set on the row-less digest, absent when rows survived alongside `_chat`, absent when nothing was compacted.

`apps/api` affected suites green single-fork (`logSearch`, `aiToolOutput`, `aiToolErrors`, `aiToolsEventLogs.siteScope`, `routes/logs` — 110 passed; `routes/mcpServer` — 16 files, 147 passed). `tsc --noEmit` clean.

Reported by @bdunncompany.

### DIFF
--- a/apps/api/src/__tests__/integration/logSearchCursor.integration.test.ts
+++ b/apps/api/src/__tests__/integration/logSearchCursor.integration.test.ts
@@ -1,0 +1,284 @@
+/**
+ * Integration test — `search_logs` keyset pagination against real Postgres
+ * (regression guard for #3329).
+ *
+ * The keyset predicate used to be a hand-written Drizzle `sql` template that
+ * interpolated the cursor's JS `Date` directly. A bare value inside a `sql`
+ * template is bound with the NOOP encoder, so postgres.js received a `Date`
+ * object at its Bind step and threw
+ * `ERR_INVALID_ARG_TYPE ... Received an instance of Date`. No mock reproduces
+ * this — the failure is in the driver's own type coercion, not in Drizzle's
+ * query-building or in application logic. The fix (`buildLogSearchKeysetCondition`
+ * in `services/logSearch.ts`) routes the comparison through Drizzle's typed
+ * `lt`/`gt`/`eq` helpers instead, which serialize through the column's own
+ * encoder.
+ *
+ * This file proves, against a real DB connection as the unprivileged
+ * `breeze_app` role (the same path production request handlers use via
+ * `withDbAccessContext`):
+ *   1. Paging with a cursor `searchFleetLogs` itself issued resolves, not
+ *      throws — this alone fails on the pre-fix code.
+ *   2. A full keyset walk (limit=4) over ~25 seeded rows exactly reproduces
+ *      a single unpaginated query's order — no gaps, no duplicates.
+ *   3. The uuid tiebreaker: a tie group of 3 rows sharing an identical
+ *      `timestamp` is split across a page boundary, and paging still walks
+ *      through it correctly. The fix removed an explicit `cast(... as uuid)`
+ *      from the old code, so this proves the id comparison still resolves as
+ *      uuid against the live driver. The test also asserts the tie group
+ *      really was split across a page boundary (not merely present) — if it
+ *      landed entirely within one page, dropping the tiebreaker would go
+ *      undetected.
+ *   4. Both `sortOrder: 'desc'` (default) and `sortOrder: 'asc'`.
+ *
+ * Run:
+ *   pnpm test:integration -- src/__tests__/integration/logSearchCursor.integration.test.ts
+ */
+import './setup';
+
+import { describe, it, expect } from 'vitest';
+import { getTestDb } from './setup';
+import { withDbAccessContext, type DbAccessContext } from '../../db';
+import { deviceEventLogs, devices } from '../../db/schema';
+import { createPartner, createOrganization, createSite } from './db-utils';
+import { searchFleetLogs } from '../../services/logSearch';
+import { buildOrgAccessClosures, type AuthContext } from '../../middleware/auth';
+
+// Wide enough to contain every timestamp seeded below without relying on the
+// default 24h window.
+const WIDE_TIME_RANGE = {
+  start: '2020-01-01T00:00:00.000Z',
+  end: '2030-01-01T00:00:00.000Z',
+};
+
+async function seedFixture() {
+  const partner = await createPartner();
+  const org = await createOrganization({ partnerId: partner.id });
+  const site = await createSite({ orgId: org.id });
+  const [device] = await getTestDb()
+    .insert(devices)
+    .values({
+      orgId: org.id,
+      siteId: site.id,
+      agentId: `agent-${org.id}`,
+      hostname: `host-${org.id.slice(0, 8)}`,
+      osType: 'linux',
+      osVersion: '1.0',
+      architecture: 'amd64',
+      agentVersion: '1.0.0',
+      status: 'online',
+    })
+    .returning();
+  if (!device) throw new Error('seedFixture: device insert returned no row');
+  return { partner, org, site, device };
+}
+
+// Distinct `source` per row so seeded rows never collide with the
+// device_event_logs_dedup_idx unique index on (device_id, source, event_id) —
+// eventId is left null on every row.
+async function seedLog(opts: {
+  orgId: string;
+  deviceId: string;
+  timestamp: Date;
+  source: string;
+}) {
+  const [row] = await getTestDb()
+    .insert(deviceEventLogs)
+    .values({
+      deviceId: opts.deviceId,
+      orgId: opts.orgId,
+      timestamp: opts.timestamp,
+      level: 'info',
+      category: 'system',
+      source: opts.source,
+      message: `msg-${opts.source}`,
+    })
+    .returning();
+  if (!row) throw new Error('seedLog: insert returned no row');
+  return row;
+}
+
+function makeAuth(orgId: string): AuthContext {
+  const { orgCondition, canAccessOrg } = buildOrgAccessClosures([orgId]);
+  return {
+    principal: 'user',
+    user: { id: '00000000-0000-0000-0000-0000000000aa', email: 'op@example.com', name: 'Op', isPlatformAdmin: false },
+    token: {} as unknown as AuthContext['token'],
+    partnerId: null,
+    orgId,
+    scope: 'organization',
+    accessibleOrgIds: [orgId],
+    orgCondition,
+    canAccessOrg,
+  } as unknown as AuthContext;
+}
+
+function ctxFor(orgId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: null,
+    userId: null,
+  };
+}
+
+// Postgres compares `uuid` columns byte-by-byte in the order the bytes appear
+// in the canonical lowercase hex text (no re-ordering, unlike e.g. Windows
+// mixed-endian GUIDs), so a plain string `<`/`>` comparison on the ids
+// returned by the driver reproduces the same ordering as the DB's own id
+// tiebreaker.
+function compareIdsAsc(a: string, b: string): number {
+  if (a < b) return -1;
+  if (a > b) return 1;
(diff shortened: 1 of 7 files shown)


bdunncompany:
Reporter here — both halves of #3329 land, and the `summarized: true` flag is the right shape for a programmatic consumer.

Confirming the `_chat` point from the other side, since it cost me a run: my poller's first guard raised on the *presence* of `_chat`, which immediately broke the alerts feed, because `manage_alerts` returns `_chat` alongside a real `alerts` array. I had to move the check to run only after row extraction failed. An explicit flag removes that guesswork entirely.

**One measured behaviour worth capturing, because it decides whether a consumer can avoid outcome 3 at all.** Detecting the digest tells you to back off, but not what to back off *on*, and the intuitive knob is the wrong one. Holding `limit` fixed at 15 and varying only `timeRange` against a busy window (0.103.0, prod fleet):

| window | response |
|---|---|
| ≤ 10 min | rows |
| ≥ 20 min | digest |

The trigger tracks how many rows **match** the window, not how many are requested. Same arguments returned 15 rows at `07:22Z` and a digest at `07:30Z` as event volume grew, and a *wider* window succeeded while a narrower one failed minutes later. So shrinking `limit` does nothing; shrinking the window is the only lever. A poller that responds to `summarized: true` by halving `limit` will loop forever.

Not a change request on this PR — the flag is what I needed. But a line in the tool description saying "on `summarized: true`, narrow the time range, not the limit" would save the next consumer the afternoon I spent measuring it.

Worth noting the earlier per-tool row ceilings I posted in #3329 (15 / 20 / 15) were **wrong** and I've corrected them: I swept limits in ascending order, so the small ones ran early and large ones late, which confounded response size with match volume. The window table above is the reliable one.

I'll switch my own bridge from the `_chat`-after-extraction heuristic to `summarized` once this reaches a release — keeping both while the fleet is mixed, since older servers won't send the field.

ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter`.

**Findings:** 0 raised by either pass; 0 outstanding.

- `code-reviewer` independently re-derived the keyset boundary against `resolveSearchOrder`'s `ORDER BY (timestamp, id)`, checked the `hasMore`/`nextCursor` off-by-one (cursor is built from the sliced page, not the `limit+1` sentinel row), the `sortBy !== 'timestamp'` guard ordering, and the `offset`-forced-to-0 consistency between query and response. It also confirmed `summarized` is a new key nothing else in the repo produces or consumes, so the extra field collides with nothing.
- `silent-failure-hunter` traced the `buildLogSearchKeysetCondition` throw through both consumers — the AI tool handler (returns `{error}`) and `POST /logs/search` (maps cursor/invalid messages to 400) — and found no path where a row-less digest escapes without `summarized: true`.

**Added after review:** an integration suite, `apps/api/src/__tests__/integration/logSearchCursor.integration.test.ts`. The reviewer's one non-blocking observation was that the unit guard renders through `PgDialect.sqlToQuery` and never reaches a driver, while the actual fault was *in* the driver's Bind step — so it could not by itself reproduce the original failure. The new suite closes that gap against real Postgres as the unprivileged `breeze_app` role.

**Tests:**

- Integration (real Postgres, this worktree's test stack): `vitest run --config vitest.integration.config.ts src/__tests__/integration/logSearchCursor.integration.test.ts` — **3 passed**.
- **Mutation-checked.** With the old `sql` template restored in `buildLogSearchKeysetCondition`, all 3 fail with the exact reported error, confirming the suite is not vacuous:
  ```
  ERR_INVALID_ARG_TYPE: The "string" argument must be of type string
  or an instance of Buffer or ArrayBuffer. Received an instance of Date
  ```
- Unit, single-fork: `logSearch`, `aiToolOutput`, `aiToolErrors`, `aiToolsEventLogs.siteScope`, `routes/logs` — 110 passed; `routes/mcpServer` — 16 files, 147 passed.
- `tsc --noEmit` clean.
- CI was fully green on the fix commit (`7287846a1`) before the test commit was added; re-running now on `b01706ac7`.

**Follow-up filed:** #3369 — the same Date-in-`sql`-template defect class exists at two more sites (`services/filterEngine.ts` datetime operators, reachable as a 500 from `POST /devices/filters/preview`, dynamic groups and deployment targeting; and `oauth/revocationRetry.ts`, where it silently prevents the revocation retry queue from draining). Deliberately not bundled here — unrelated subsystems, one auth-adjacent.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3329 — the reporter's verification comes first.

ToddHebebrand:
CI is green on `b01706ac7` — 54 checks passed, 0 failed. That includes all four **Integration Tests** shards, which is where the new `logSearchCursor.integration.test.ts` suite actually runs against Postgres.

No outstanding findings. Ready for maintainer merge.

PR #3370, opened 2026-08-10
fix(quotes): stop the quote line unit price clipping its last digit (#3318)

### LINKED ISSUE #3318: [UI] Quote editor line row clips the last digit of the formatted unit price

ToddHebebrand:
Closes #3318

## What was wrong

`growWidth` in `apps/web/src/components/billing/quotes/QuoteLineRows.tsx` sized the money/percent inputs as a bare `` `${n}ch` `` string, with a two-character buffer over the displayed value. Tailwind's preflight sets `box-sizing: border-box`, so that width **includes** the input's own horizontal padding and border — and on the unit-price input `px-2` + 1px borders is 18px, which at the measured **8.96px/ch** is almost exactly the 2ch buffer.

So the buffer was spent on chrome, not glyphs: the field ended up sized to exactly `value.length` ch, while the rendered glyphs are *wider* than `1ch` (Plus Jakarta Sans at 14px — tabular digit **9.60px**, `$` **10.34px**, against a **8.96px** `ch`).

Measured slack for the blurred, currency-formatted value **before** the fix:

| displayed | old width | content box | text | slack |
|---|---|---|---|---|
| `$45.00` | `8ch` | 53.68px | 53.31px | **0.37px** |
| `$1,850.00` | `11ch` | 80.55px | 76.84px | **3.72px** |
| `$1,234,567.89` | `15ch` | 116.40px | 109.95px | 6.45px |

Under half a digit — so the trailing digit renders clipped (`$45.0` / `$1,850.0`) while the adjacent Total cell shows the full amount, exactly as reported.

## The fix

`growWidth` now clamps the **character count** and emits `calc(<chars + 2>ch + <padding> + <border>)`, funding the input's chrome separately so the two-character buffer survives for glyphs. Applied to all four call sites — line price, ghost-row price, internal cost, markup — each passing its own padding shape (`px-2` vs `px-1`). The clamps moved from "total ch" to "characters" and the price ceiling rose (13→15 chars) so long formatted values are no longer clamped into a clip.

Post-fix slack, same measurement method:

| displayed | new width | slack |
|---|---|---|
| `$45.00` | `calc(8ch + 1rem + 2px)` | 18.37px |
| `$1,850.00` | `calc(11ch + 1rem + 2px)` | 21.72px |
| `$1,234,567.89` | `calc(15ch + 1rem + 2px)` | 24.45px |
| `1.234,56 €` (de-DE) | `calc(12ch + 1rem + 2px)` | 25.87px |
| `1 234,56 €` (fr-FR) | `calc(12ch + 1rem + 2px)` | 29.08px |
| `1,234.56 CHF` | `calc(14ch + 1rem + 2px)` | 22.65px |
| `¥1,850` | `calc(8ch + 1rem + 2px)` | 18.98px |
| `R$ 1.234,56` | `calc(13ch + 1rem + 2px)` | 26.66px |
| `1 234 567,89 €` | `calc(16ch + 1rem + 2px)` | 32.05px |

Every case now clears the string by 2+ digits. Resting/floor widths grow by the ~18px of chrome that was previously being stolen from the value — that widening *is* the fix.

## Verification

- **Real-browser measurement.** jsdom does no layout, so the numbers above come from rendering the editor's actual DOM (dumped from a real `QuoteEditor` render) against the app's **compiled Tailwind CSS** and the **self-hosted Plus Jakarta Sans** woff2, in Chromium, and comparing each input's content box against the text width measured with that input's own computed `font` / `font-variant-numeric`. Reported as a table rather than a screenshot because the failure is a sub-pixel margin, which a screenshot hides.
- **Tests** — `QuoteEditor.moneyinput.test.tsx` extended: parametrised regressions for the two reported values (`$1,850.00`, `$45.00`), a non-USD (EUR) quote, the internal cost input, and the ghost row. jsdom can't measure pixels, so the assertable contract is the width *expression*: the `ch` budget must clear the displayed string, and the padding/border must be funded on top of it (`calc(... + rem + px)`). Confirmed these 7 assertions **fail against the old formula** and pass against the fix.
- `pnpm exec vitest run src/components/billing/quotes/` — **47 files / 361 tests passed**.
- `npx tsc --noEmit` (apps/web) — clean.

Scope is `apps/web` only; no overlap with the API-side #3319 work. The invoice editor's price field is a plain `type="number"` with a fixed `w-24` and no currency formatting, so it doesn't share this failure mode and is untouched.

### DIFF
--- a/apps/web/src/components/billing/quotes/QuoteEditor.moneyinput.test.tsx
+++ b/apps/web/src/components/billing/quotes/QuoteEditor.moneyinput.test.tsx
@@ -85,14 +85,20 @@ const updateLineMock = vi.mocked(updateLine);
 //     allowance. `box-sizing: border-box` means a width of bare `<n>ch` spends
 //     the buffer on the input's own chrome instead of on glyphs, which is what
 //     clipped the last digit of "$45.00" in issue #3318.
-function widthBudget(el: HTMLInputElement): { chBudget: number; fundsChrome: boolean } {
+// `expectedPadX` must name the field's OWN padding (the `px-2` money inputs are
+// 1rem, the `px-1` internal-band inputs 0.5rem). Asserting the exact value —
+// not merely "some rem term is present" — is what makes the check bite: sizing
+// a px-2 input with the band's 0.5rem constant under-funds its chrome by 8px
+// and silently reopens #3318, and jsdom does no layout to catch it.
+function widthBudget(el: HTMLInputElement, expectedPadX: '1rem' | '0.5rem'): { chBudget: number; fundsChrome: boolean } {
   const width = el.style.width;
   const ch = /(\d+(?:\.\d+)?)ch/.exec(width);
   if (!ch) throw new Error(`expected a ch-based width, got "${width}"`);
+  // jsdom reorders calc() terms, so match on presence, not position.
+  const padRe = new RegExp(`(?<![\\d.])${expectedPadX.replace('.', '\\.')}(?![\\d.])`);
   return {
     chBudget: Number(ch[1]),
-    // jsdom reorders calc() terms, so match on presence, not position.
-    fundsChrome: /calc\(/.test(width) && /\dpx/.test(width) && /rem/.test(width),
+    fundsChrome: /^calc\(/.test(width) && /(?<![\d.])2px(?![\d.])/.test(width) && padRe.test(width),
   };
 }
 
@@ -188,7 +194,7 @@ describe('QuoteLineRows — money-formatted price/cost inputs', () => {
     // Unfocused display "$1,234,567.89" is 13 characters — width must grow well
     // past the old fixed w-24 (~12ch) to avoid clipping the formatted value.
     expect(priceEl.value).toBe('$1,234,567.89');
-    const { chBudget, fundsChrome } = widthBudget(priceEl);
+    const { chBudget, fundsChrome } = widthBudget(priceEl, '1rem');
     expect(chBudget).toBeGreaterThanOrEqual(priceEl.value.length + 2);
     expect(fundsChrome).toBe(true);
   });
@@ -206,7 +212,7 @@ describe('QuoteLineRows — money-formatted price/cost inputs', () => {
 
     const priceEl = screen.getByTestId('quote-line-price-line-1') as HTMLInputElement;
     expect(priceEl.value).toBe(formatted);
-    const { chBudget, fundsChrome } = widthBudget(priceEl);
+    const { chBudget, fundsChrome } = widthBudget(priceEl, '1rem');
     expect(chBudget).toBeGreaterThanOrEqual(formatted.length + 2);
     expect(fundsChrome).toBe(true);
   });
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (full diff vs `origin/main`, focused on the `padX` constants against each input's Tailwind classes, the clamp semantics change, and whether the new jsdom assertions can pass vacuously).

**Findings:** 3 raised (0 blockers) → all 3 addressed in `28269c000`; 0 outstanding.

- *should-fix* — `maxChars=15` still clipped code-prefixed currencies: `currencyCode` is free text and Intl prefixes many ISO codes rather than symbolising them, so `CHF 1,234,567.89` (16 chars) was clamped back into the original failure mode. Ceilings raised to 18 characters on the price / ghost-price / cost inputs, pinned by a new CHF+SEK 7-digit regression case.
- *should-fix* — `widthBudget()` could not tell the two padding constants apart (`/rem/` matched either), so sizing a `px-2` input with the band's `0.5rem` constant — under-funding chrome by 8px and silently reopening #3318 — would have passed. The helper now takes the field's expected padding and asserts that exact term.
- *nit* — the font metrics in the `growWidth` comment read as possibly invented. Added a line recording that they were measured in Chromium against the app's compiled Tailwind CSS and self-hosted Plus Jakarta Sans.

Reviewer explicitly verified the `padX` constants match every call site's classes (including `ghostField`'s `border border-transparent`), that Tailwind's `border` utility is not overridden in `globals.css` so the `+2px` term is accurate, and that no other `apps/web` call site reuses the old bare-`${n}ch` pattern.

**Tests:** `pnpm exec vitest run src/components/billing/quotes/` — **47 files / 363 tests passed**. `npx tsc --noEmit` (apps/web) clean. The 7 new sizing assertions were confirmed to **fail against the old formula** before the fix, so they are not vacuous.

**Browser evidence:** measured on the real rendered DOM against the app's compiled Tailwind CSS and self-hosted Plus Jakarta Sans in Chromium — blurred `$45.00` went from **0.37px** of slack to **18.37px**, `$1,850.00` from **3.72px** to **21.72px**; every USD/EUR/CHF/JPY/BRL format now clears its string by 2+ digits. Full table in the PR description.

**Status:** review-clean, awaiting maintainer merge. Not merged and #3318 left open per hand-off rules — worth a quick eyeball in the running editor before closing, since the price column now rests ~18px wider (that widening is the chrome the value was previously being robbed of).

PR #3371, opened 2026-08-10
fix(agent): normalize reboot/shutdown delay to minutes on Windows (#3252)

### LINKED ISSUE #3252: [Agent] `reboot` delay is minutes on Linux/macOS and seconds on Windows
## Description

`tools.Reboot` treats its `delay` payload as **minutes** on Linux/macOS and as **seconds** on Windows, from the same command payload. A `reboot` command with `delay: 15` reboots a Linux box in 15 minutes and a Windows box in 15 seconds.

Found while investigating #3197 (patch reboots never warn the user); not the cause of that issue, but the same defect class in the adjacent code path.

## Root cause

`agent/internal/remote/tools/system.go:13-20` documents and clamps `delay` as minutes:

```go
// Reboot schedules a system reboot after the requested delay.
// Maximum delay is 1440 minutes (24 hours).
delay := GetPayloadInt(payload, "delay", 0)
... else if delay > 1440 { delay = 1440 }
```

`buildShutdownCommand` (`agent/internal/remote/tools/system.go:94-111`) then emits:

- Windows: `shutdown /r /t <delay>` — **`/t` is seconds**
- linux/darwin: `shutdown -r +<delay>` — `+N` is minutes

So the clamp to 1440 means "24 hours" on Linux and "24 minutes" on Windows, and every intermediate value is off by 60x.

`Shutdown` (`system.go:41-58`) shares `buildShutdownCommand` and has the identical bug.

## Why it hasn't bitten

The only server-side caller that passes a non-zero delay is the maintenance reboot worker, and it sends `reboot` **only to Linux** — `apps/api/src/jobs/maintenanceRebootWorker.ts:82-84` returns `{type:'reboot', payload:{delay:15}}` for linux and routes Windows to `schedule_reboot` instead. Manual reboots from the UI pass `delay: 0`, where both units agree.

The exposure is any caller that sends `reboot` with a delay to a Windows device.

## Proposed fix

Multiply by 60 in the Windows branch of `buildShutdownCommand` so the payload is minutes on every platform, matching the doc comment and the clamp. Add a table-driven test over `buildShutdownCommand` asserting the emitted argv per GOOS — there is no test for it today.

Consider also whether the 0-delay case should keep emitting `/t 0`; #3197's spec proposes `/t 15` on the patch-reboot path specifically so the "Rebooting Now" toast has a window to render.

## Affected files

- `agent/internal/remote/tools/system.go` (`Reboot` :13, `Shutdown` :41, `buildShutdownCommand` :94)

ToddHebebrand:
Closes #3252

## Problem

`buildShutdownCommand` emitted the caller-supplied `delay` verbatim into both platform command lines, but the two flags use different units:

| Platform | Command | Unit of the delay arg |
|---|---|---|
| Windows | `shutdown /r /t <n>` | **seconds** |
| linux / darwin | `shutdown -r +<n>` | **minutes** |

The wire contract is **minutes on every platform**:

- `apps/docs/src/content/docs/agents/commands.mdx:271` — "`delay` | int | `0` | Delay in minutes before rebooting (0-1440)"
- the `1440` clamp in `Reboot`/`Shutdown` and their doc comments
- `MAINTENANCE_REBOOT_GRACE_MINUTES = 15`, which `maintenanceRebootWorker.ts` sends as `{type:'reboot', payload:{delay:15}}`

So `delay: 15` rebooted a Linux box in 15 minutes and a Windows box in **15 seconds**, and the 1440 clamp meant "24 hours" on Linux but "24 minutes" on Windows. Every intermediate value was off by 60x. `Shutdown` shared the same helper and the same defect.

**Why it hasn't bitten in production:** the only server-side caller that sends a non-zero delay routes Windows to `schedule_reboot` (which takes an explicit `delayMinutes`) and only sends `reboot` to Linux. Manual UI reboots pass `delay: 0`, where both units agree. The exposure is any caller that sends `reboot`/`shutdown` with a delay to a Windows device.

## Fix

Agent-side conversion only — **the server contract is unchanged**, as requested in the issue. `RebootToSafeMode` (`system_safemode_windows.go`) already scaled minutes to seconds correctly, so the Windows branch here now simply matches its sibling.

- Extract `shutdownArgs(goos, isReboot, delayMinutes)` — a pure, GOOS-parameterized argv builder. Taking `goos` as a parameter instead of reading `runtime.GOOS` inline is what makes the per-platform mapping table-testable from any host. `buildShutdownCommand` now delegates to it.
- Scale by 60 in the Windows branch.
- Extract `clampShutdownDelayMinutes` + `maxShutdownDelayMinutes`, replacing three hand-rolled copies of the same clamp (`Reboot`, `Shutdown`, `RebootToSafeMode`). Also applied *inside* `shutdownArgs`, so the emitted argv can never carry an out-of-range value regardless of caller — relevant now that the value gets multiplied.

## Tests

New `agent/internal/remote/tools/system_test.go` (there was no coverage of this code at all):

- **`TestShutdownArgs_PerPlatformArgv`** — table-driven over the emitted argv per GOOS: reboot vs shutdown, 0/1/15/1440 min, and negative/over-max clamping.
- **`TestShutdownArgs_DelayIsMinutesOnEveryPlatform`** — the invariant this issue violated: asserts the same `delay` resolves to the same wall-clock seconds on windows, linux and darwin.
- **`TestShutdownArgs_UnsupportedOS`** — error path returns zero-value argv.
- **`TestClampShutdownDelayMinutes`** — clamp bounds.
- **`TestBuildShutdownCommand_MatchesShutdownArgsForHostOS`** — guards against `buildShutdownCommand` drifting back to constructing argv independently.

The new tests **fail against the pre-fix code** (verified by reverting the `* 60`: `argv: got [/r /t 15], want [/r /t 900]`).

## Verification

- `go test -race ./internal/remote/tools/ ./internal/heartbeat/ ./internal/privilege/` — all green
- Cross-builds `GOOS=windows|linux|darwin go build ./...` — all OK
- `go vet` clean under all three GOOS (covers the `//go:build windows` safe-mode file)
- `gofmt` clean on touched files

## Out of scope

The issue asks to "consider whether the 0-delay case should keep emitting `/t 0`". Left as-is: that concerns the patch-reboot toast window and belongs to #3197's spec, which targets a different code path. Changing it here would alter manual-reboot behaviour beyond this fix.

### DIFF
--- a/agent/internal/remote/tools/system.go
+++ b/agent/internal/remote/tools/system.go
@@ -8,16 +8,29 @@ import (
 	"time"
 )
 
+// maxShutdownDelayMinutes bounds the `delay` payload for reboot/shutdown.
+// The wire contract is minutes on every platform (see docs/agents/commands.mdx),
+// so this is 24 hours.
+const maxShutdownDelayMinutes = 1440
+
+// clampShutdownDelayMinutes constrains a caller-supplied delay to [0, 1440]
+// minutes. Applied both when reporting the delay back to the server and inside
+// shutdownArgs, so the emitted argv can never carry an out-of-range value.
+func clampShutdownDelayMinutes(delayMinutes int) int {
+	if delayMinutes < 0 {
+		return 0
+	}
+	if delayMinutes > maxShutdownDelayMinutes {
+		return maxShutdownDelayMinutes
+	}
+	return delayMinutes
+}
+
 // Reboot schedules a system reboot after the requested delay.
-// Maximum delay is 1440 minutes (24 hours).
+// The `delay` payload is in MINUTES on every platform; maximum 1440 (24 hours).
 func Reboot(payload map[string]any) CommandResult {
 	startTime := time.Now()
-	delay := GetPayloadInt(payload, "delay", 0)
-	if delay < 0 {
-		delay = 0
-	} else if delay > 1440 {
-		delay = 1440
-	}
+	delay := clampShutdownDelayMinutes(GetPayloadInt(payload, "delay", 0))
 
 	cmd, err := buildShutdownCommand(true, delay)
 	if err != nil {
@@ -37,15 +50,10 @@ func Reboot(payload map[string]any) CommandResult {
 }
 
 // Shutdown schedules a system shutdown after the requested delay.
-// Maximum delay is 1440 minutes (24 hours).
+// The `delay` payload is in MINUTES on every platform; maximum 1440 (24 hours).
 func Shutdown(payload map[string]any) CommandResult {
 	startTime := time.Now()
-	delay := GetPayloadInt(payload, "delay", 0)
-	if delay < 0 {
-		delay = 0
-	} else if delay > 1440 {
-		delay = 1440
-	}
+	delay := clampShutdownDelayMinutes(GetPayloadInt(payload, "delay", 0))
 
 	cmd, err := buildShutdownCommand(false, delay)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter` (both Sonnet, run independently against `origin/main...HEAD`).

**Findings:** 0 blocking. Both reviewers independently raised the same non-blocking observation — `RebootToSafeMode` still hand-rolled its own `delay * 60` and `exec.Command("shutdown", "/r", "/t", ...)` argv, i.e. the exact duplicated-conversion pattern that produced this bug. Addressed in 2bbc16591 by routing it through `buildShutdownCommand`, so the minutes→seconds conversion now has a single source of truth covered by the new tests. That commit also builds the command *before* setting the BCD safeboot flag, so a construction error can no longer strand a machine with safeboot set and no reboot scheduled. **0 outstanding.**

Points explicitly cleared by review:

- `shutdown /t 86400` (the 24h post-clamp max) is well inside Windows' documented 0–315360000 range.
- The `* 60` never bleeds into the `CommandResult` — `result["delay"]` still reports clamped **minutes**, unchanged.
- The unsupported-OS error string and its propagation through `buildShutdownCommand` → `Reboot`/`Shutdown` are byte-identical to before; nothing swallowed.
- `clampShutdownDelayMinutes` is idempotent, so the defense-in-depth double-clamp cannot mask anything a single clamp didn't.
- No missed call sites: `patching/reboot_windows.go:228` hardcodes `/t 0`, and `schedule_reboot` uses a separate `delayMinutes`/`time.Duration` path that never rode the buggy code.

**Tests:** `go test -race -count=1 ./internal/remote/tools/ ./internal/heartbeat/ ./internal/privilege/` — all green (uncached). Cross-builds `GOOS=windows|linux|darwin go build ./...` and `go vet` clean under all three GOOS (covers the `//go:build windows` safe-mode file). `gofmt` clean on touched files. The new tests were confirmed to **fail against the pre-fix code** (`argv: got [/r /t 15], want [/r /t 900]`), so they genuinely pin the behavior.

**Pre-existing, out of scope** (surfaced by review, not introduced here — worth a separate issue if anyone wants it):

- `GetPayloadInt` (`agent/internal/remote/tools/types.go:631`) only matches `int`/`int64`/`float64`, so a JSON-string `delay: "15"` silently falls back to `0` — an immediate reboot rather than a validation error.
- `handleRebootSafeMode` (`agent/internal/heartbeat/handlers.go:289`) re-reads `delay` unclamped purely to compose the user-facing notification string, so an out-of-range payload would display a wrong number while the actual reboot still clamps correctly. Cosmetic.

**Status:** review-clean, awaiting maintainer merge. Not merged, issue left open per handoff rules.

PR #2977, opened 2026-07-31
fix(web): searchable IANA timezone picker for site/org/partner settings (#2856)

ToddHebebrand:
Refs #2856 — reported by a French self-hoster in discussion #2321 (point 5).

## The bug

Creating an organization/site offered a **10-entry hardcoded timezone list** (`UTC`, 4× US, `Europe/London`, `Europe/Berlin`, `Asia/Singapore`, `Asia/Tokyo`, `Australia/Sydney`). **No `Europe/Paris`** — a French self-hoster could not select their own timezone at all. Three divergent copies existed:

| File | List |
|---|---|
| `SiteForm.tsx` (org/site **creation**) | 10 entries, no Paris |
| `SiteDetailPage.tsx` (site **edit**) | a third copy of the same 10 |
| `PartnerRegionalTab.tsx` | a *different* 15 entries, which did include Paris |

The API accepts any IANA string (`timezone: z.string().optional()`, default `'UTC'`), so this was purely a picker limitation. `SiteForm` even carried a workaround effect that prepended an out-of-list *stored* value to the options — an acknowledgement the list was incomplete — but there was no way to *select* one in the first place.

## The fix

**One source of truth** — `listIanaTimezones()` in `packages/shared/src/utils/timezone.ts`:

- `Intl.supportedValuesOf('timeZone')` (~418 zones) with the `'UTC'` sentinel prepended, since that API omits UTC/GMT/`Etc/*` (the same omission this module already documents for `isValidIanaTimezone`).
- Cached, and returns a **copy** so a caller sorting/splicing in place cannot poison the cache.
- Fallback for runtimes without the ES2022 API is a **broad curated list (~57 zones, superset of both old lists)**, not a narrow one — a 10-entry fallback would just reproduce the bug on those runtimes.

**New `TimezoneSelect`** (`apps/web/src/components/shared/TimezoneSelect.tsx`) — 418 options make a bare `<select>` unusable, so this is a combobox with a filter field:

- ARIA shape follows the repo's existing `OrgCombobox`: the trigger is a plain disclosure **button** showing the committed value; `role="combobox"` belongs to the **search input** (the element that holds focus while typing/arrowing); `aria-activedescendant` tracks the keyboard cursor and option `aria-selected` follows it.
- Search matches the city with underscores as spaces — `"new york"` finds `America/New_York`, `"paris"` finds `Europe/Paris`.
- Each row shows its current UTC offset (`shortOffset`), memoized per zone.
[...]

Issue #2856, opened 2026-07-27
[Web] Site/org creation timezone dropdown is a 10-entry hardcoded list — Europe/Paris (and most IANA zones) missing


ToddHebebrand:
Reported in discussion #2321 (point 5): a French self-hoster creating their first organization could not select Europe/Paris.

**Root cause:** two divergent hardcoded timezone lists.

- `apps/web/src/components/settings/SiteForm.tsx:36` — `timezoneOptions` has 10 entries (UTC, 4×US, Europe/London, Europe/Berlin, Asia/Singapore, Asia/Tokyo, Australia/Sydney). No Europe/Paris. This is the list used when creating orgs/sites on the Organizations & Sites page.
- `apps/web/src/components/settings/PartnerRegionalTab.tsx:8` — a different 15-entry list that *does* include Europe/Paris.

The API accepts any string (`timezone` is `z.string().optional()`, defaults `'UTC'`), so this is purely a picker limitation — SiteForm even has a workaround effect (`SiteForm.tsx:83-88`) that injects an out-of-list *existing* value into the options, acknowledging the list is incomplete, but there's no way to *select* one in the first place.

**Fix shape:** replace both hardcoded lists with `Intl.supportedValuesOf('timeZone')` (supported in all our target browsers; ~418 entries, searchable select rather than a bare `<select>`), or at minimum a single shared curated list including all common European/IANA zones. Downstream consumers are fine: maintenance windows and schedulers resolve timezones by IANA name.

**Verify:** create a new organization as a French partner → Europe/Paris selectable; existing site with an out-of-list timezone still renders its current value.

ToddHebebrand:
**Fixed.** `2f3d15204` (PR #2977) replaced the 10-entry hardcoded list with a searchable IANA picker — `components/shared/TimezoneSelect.tsx`, now used by `SiteForm.tsx`, `SiteDetailPage.tsx` and `PartnerRegionalTab.tsx`. `2dbe7505f` (PR #3376) later carried the same component into the config-policy Automation and Maintenance tabs.

Closing — verified on `main` during v0.105.0 release triage.

PR #3376, opened 2026-08-10
fix(web): use shared TimezoneSelect in config-policy Automation/Maintenance tabs (#3361)

### LINKED ISSUE #3361: [Web] Config-policy Automation/Maintenance tabs still use hardcoded timezone lists — same class as #2856
## Description
#2856 replaced the hardcoded site/org/partner timezone dropdowns with the shared searchable `TimezoneSelect` (PR #2977, merged 2026-08-05). A sweep for remaining copies of the old pattern found two survivors in the configuration-policy feature tabs:

- `apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.tsx:56` — 9-entry `timezoneOptions` (i18n-keyed labels)
- `apps/web/src/components/configurationPolicies/featureTabs/MaintenanceTab.tsx:64` — 16-entry `createTimezoneOptions()`

Both include `Europe/Paris`, so they're lower severity than the original report — but a config-policy author in e.g. `Asia/Dubai` or `America/Sao_Paulo` cannot pick their zone for an automation trigger or a maintenance window.

## Proposed Fix
Consolidate both tabs onto the shared `TimezoneSelect` component (`apps/web/src/components/shared/TimezoneSelect.tsx`), which is backed by `packages/shared/src/utils/timezone.ts` (`Intl.supportedValuesOf('timeZone')` + fallback). Single small PR.

## Affected Files
- `apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.tsx`
- `apps/web/src/components/configurationPolicies/featureTabs/MaintenanceTab.tsx`

## Reported By
Found during a verification sweep of #2856 (originally raised via discussion #2321).

ToddHebebrand:
## Problem

PR #2977 replaced the hardcoded site/org/partner timezone dropdowns with the shared searchable `TimezoneSelect` (issue #2856). A verification sweep found two survivors of the same defect class in the configuration-policy feature tabs:

| File | Old list |
|---|---|
| `AutomationTab.tsx:56` | 9-entry `timezoneOptions` (i18n-keyed labels), gating the **schedule trigger** timezone |
| `MaintenanceTab.tsx:64` | 16-entry `createTimezoneOptions()` (raw IANA strings), gating the **maintenance window** timezone |

Both omitted most of the world, so a config-policy author in `Asia/Dubai` or `America/Sao_Paulo` simply could not pick their own zone. Lower severity than the original report only because both lists happened to include `Europe/Paris`.

## Fix

Both tabs now render `TimezoneSelect` (`apps/web/src/components/shared/TimezoneSelect.tsx`), backed by `listIanaTimezones()` in `packages/shared/src/utils/timezone.ts` — `Intl.supportedValuesOf('timeZone')` with a fallback, ~418 zones, searchable. As a bonus the shared component **preserves a stored zone that is not in the Intl list** instead of silently rewriting it to the first option.

Follows the `SiteForm` / `PartnerRegionalTab` integration pattern:

- the existing `<label>` gains an `htmlFor` pointing at `TimezoneSelect`'s `id`;
- the same translated string is passed as `label`, so the trigger's accessible name carries both the field name and the committed zone (`aria-label` overrides name-from-content, so the value has to be in there);
- AutomationTab's control lives inside a per-item map, so its `id`/`testId` are suffixed with the item index (`automation-timezone-0`).

## i18n

The 9 `configurationPolicies.featureTabs.automationTab.*` per-zone label keys are now unreferenced and are **deleted from all 7 locale files**. Reasoning, since this was the risky part:

- There is **no unused-key gate** — `keyUsage.test.ts` only checks source → catalog, so leaving them would also have been green. Deleted anyway rather than leaving dead keys for translators to maintain.
- `localeParity.test.ts` requires **identical key sets across locales**, so this is all-or-nothing: 9 keys x 7 locales, no partial deletion.
- `translationCoverage.test.ts` namespace baselines are **upper bounds** (`namespaceDuplicateRegressions` flags only *exceeding* the baseline), so removing keys cannot regress them.
- Locale JSON edits were done via a script that **round-trip-verifies formatting before mutating**, so the diff is exactly 9 deleted lines per file and nothing else.

MaintenanceTab had no per-zone keys to orphan — its `maintenanceTab.timezone` field label is retained and now also feeds the picker's accessible name.

## Tests

6 new cases across the two tab suites, asserting the **outcome** rather than `TimezoneSelect`'s internals (so they keep their meaning if the picker is restyled):

- a zone absent from the old hardcoded list (`Asia/Dubai`) is offered by search;
- a zone picked from the full list (`America/Sao_Paulo`) reaches the **save payload**;
- MaintenanceTab preserves a stored out-of-list zone (`Africa/Nairobi`) on render;
- AutomationTab still defaults to `UTC` and leaves it intact when untouched.

**Verification**
- `vitest run AutomationTab.test.tsx MaintenanceTab.test.tsx` — 12 passed (2 files)
- i18n gates + `TimezoneSelect.test.tsx` + `structuralValues.test.ts` — 129 passed. `no-translated-comparisons.test.ts` tripped its 5s timeout on the first pass purely from host contention (repo-wide AST walk running alongside 8 other suites); re-run in isolation it passes in 3.4s.
- `tsc --noEmit -p apps/web/tsconfig.json` — exit 0, zero errors. No `.astro` files touched.
- `eslint` on all four touched source/test files — clean.

Closes #3361

### DIFF
--- a/apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.test.tsx
+++ b/apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.test.tsx
@@ -151,6 +151,20 @@ describe('AutomationTab — schedule timezone picker', () => {
     expect(payload.inlineSettings.items[0].timezone).toBe('America/Sao_Paulo');
   });
 
+  // The picker is scoped to the schedule trigger. Pinned because it is now a
+  // mounted combobox rather than an inert <select>: leaking it into the event /
+  // manual branches would put a focusable popup in a form where the timezone is
+  // meaningless.
+  it('renders no timezone picker for non-schedule triggers', () => {
+    addAutomation();
+    expect(screen.getByTestId('automation-timezone-0-trigger')).toBeTruthy();
+
+    for (const trigger of ['event', 'manual']) {
+      fireEvent.click(screen.getByRole('button', { name: new RegExp(`^${trigger}$`, 'i') }));
+      expect(screen.queryByTestId('automation-timezone-0-trigger')).toBeNull();
+    }
+  });
+
   it('defaults to UTC and leaves it intact when untouched', async () => {
     addAutomation();
     expect(screen.getByTestId('automation-timezone-0-trigger').textContent).toContain('UTC');
--- a/apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.test.tsx
+++ b/apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.test.tsx
@@ -109,3 +109,55 @@ describe('AutomationTab — deploy_software action', () => {
     expect(action).toEqual({ type: 'deploy_software', catalogId: 'cat-1' });
   });
 });
+
+// Issue #3361: the schedule trigger shipped its own 9-entry hardcoded timezone
+// list (the same defect #2856 fixed for sites/orgs/partners), so an automation
+// could not be scheduled in e.g. Asia/Dubai. Asserted on the OUTCOME — the
+// chosen zone reaching the save payload — not on TimezoneSelect's internals.
+describe('AutomationTab — schedule timezone picker', () => {
+  beforeEach(async () => {
+    await i18n.changeLanguage('en');
+    vi.clearAllMocks();
+    fetchMock.mockResolvedValue(makeJsonResponse({ data: CATALOG }));
+  });
+
+  // The timezone control only exists under the (default) "schedule" trigger,
+  // and the empty-state button both creates and expands the automation.
+  const addAutomation = () => {
+    renderTab();
+    fireEvent.click(screen.getAllByRole('button', { name: /Add Automation/i })[0]);
+  };
+
+  it('offers zones that were absent from the old hardcoded list', () => {
+    addAutomation();
+    fireEvent.click(screen.getByTestId('automation-timezone-0-trigger'));
+    fireEvent.change(screen.getByTestId('automation-timezone-0-search'), {
+      target: { value: 'dubai' },
+    });
+    expect(screen.getByTestId('automation-timezone-0-option-Asia/Dubai')).toBeTruthy();
+  });
+
+  it('saves a zone picked from the full IANA list', async () => {
+    addAutomation();
+    fireEvent.click(screen.getByTestId('automation-timezone-0-trigger'));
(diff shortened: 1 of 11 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (behavioral parity, reference-pattern conformance, index-keyed `id`/`testId` safety, i18n key hygiene, dead code) + `pr-test-analyzer` (coverage value, revert-detection, uncovered paths).

**Findings:** 1 raised, 0 blocking.
- `code-reviewer`: **0 findings.** Explicitly cleared the two things worth worrying about here — (a) `onChange` parity, since `TimezoneSelect` emits the raw IANA string exactly as `e.target.value` did, and (b) the array-index-keyed `id`/`testId` inside AutomationTab's `.map()`, which is safe because `expandedIndex` is a single `number | null` accordion, so at most one picker is ever mounted and `deleteItem` already decrements the index correctly.
- `pr-test-analyzer`: coverage **adequate to ship**; confirmed all 6 new tests *throw* rather than silently pass on revert (they select `data-testid`s that exist only on `TimezoneSelect`). One non-blocking gap rated 3/10 — no assertion that the picker is absent for the `event`/`manual` trigger types. **Addressed in 50c22dcc1** rather than deferred: the control is now a mounted combobox rather than an inert `<select>`, so leaking it into those branches would put a focusable popup in a form where the timezone is meaningless — worth pinning. 0 outstanding.

**Tests:**
- `vitest run AutomationTab.test.tsx MaintenanceTab.test.tsx` — **13 passed** (6 + 7; 7 of them new).
- i18n gates (`localeParity`, `keyUsage`, `translationCoverage`, `extractionQuality`), `TimezoneSelect.test.tsx`, `structuralValues.test.ts` — **129 passed**. `no-translated-comparisons.test.ts` tripped its 5s timeout on the first pass from host contention (repo-wide AST walk running alongside 8 other suites); re-run in isolation it passes in 3.4s. Not related to this change — the diff introduces no equality comparison against a translated string.
- `tsc --noEmit -p apps/web/tsconfig.json` — exit 0, zero errors. No `.astro` files touched.
- `eslint` on all touched source/test files — clean.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3361 — leaving both for you.

PR #3377, opened 2026-08-10
fix(desktop): stop viewer answer-poll from exhausting the global IP rate limit (#3041)

### LINKED ISSUE #3041: Viewer answer-poll 401s bypass session-ended check and exhaust global IP rate limit
Reported by **semotech** in #bugs (Discord, message 1533617639463584017).

**Report (PII-redacted):**

**Bug: remote desktop viewer poll can rate-limit the console out (v0.103.0)**

Self-hosted `0.103.0`. Got kicked to the login screen with **“Too many requests”** right after refreshing Devices. Traced it in `breeze-api` logs; not a bad password / lockout.

### What happened
1. Started a remote desktop session (device had been rebooted shortly before).
2. Viewer began polling `GET /api/v1/desktop-ws/:id/viewer/session` ~every 50–60ms (WebRTC answer poll).
3. Poll flipped from 200 → **401** and kept going: **213 × 401 in ~26 seconds**.
4. Global per-IP limiter then returned **429 `{ "error": "Too many requests" }`** (default **300 / 60s**).
5. Devices refresh / subsequent console calls hit that 429 → bounced to Sign in with that error.

### Why it looks like a product bug
In `apps/viewer/src/lib/webrtc.ts`:
- `ANSWER_POLL_INTERVAL_MS = 50` → up to ~300 polls in the 15s timeout — can alone exhaust the global IP bucket.
- On 401, `pollForAnswer` only throws `SessionEndedError` when `isSessionEndedResponse()` matches (`session ended` / `ended` / `revoked` / `no longer active`).
- API also returns 401s like **`Session closed`** / **`Invalid or expired viewer token`**, which do **not** match → loop continues every 50ms instead of stopping (comments say bare/session-ended 401s should be terminal — Finding #5).
- Viewer polls are **not** excluded from `globalRateLimit` (unlike `/api/v1/agents/`), so they starve the dashboard/auth for that IP.

### Ask / suggested fix
1. Treat **any** 401 on `/viewer/session` (and offer) as terminal — stop polling / reconnect.
2. Expand matcher to include `Session closed`, invalid/expired/missing viewer token (or just `status === 401`).
3. Slow the answer poll (e.g. 250–500ms+) and/or give viewer traffic its own bucket / skip prefix so it can’t lock out the console.
4. On console 429, show a cooldown / Retry-After instead of a dead-end login error when the session cookie is still valid.

Happy to share redacted log snippets / request_ids

## Triage analysis

Verified all three claims in the report against source. (1) apps/viewer/src/lib/webrtc.ts:64 sets ANSWER_POLL_INTERVAL_MS = 50, and pollForAnswer() (lines 217-250) loops every 50ms for up to 15s, so a single connection attempt can emit up to ~300 requests to GET /viewer/session. (2) The API's actual 401 bodies for that endpoint, defined in validateViewerSessionAccess (apps/api/src/routes/desktopWs.ts:192-208) and confirmed by its own tests (desktopWs.test.ts:248,258,273,283), are 'Missing viewer token', 'Invalid or expired viewer token', 'Viewer token revoked', and 'Session closed'. The client's isSessionEndedResponse() regex (webrtc.ts:42-46) is /session\s*ended|ended|revoked|no longer active/i — 'Session closed' and the token-missing/invalid variants do NOT match ('ended'/'revoked' substrings aren't present), so the poll does not treat them as terminal and keeps hammering every 50ms until the full 15s timeout instead of stopping on the first 401. (3) apps/api/src/middleware/globalRateLimit.ts only skips '/api/v1/agents/' and '/api/v1/helper/' (BUILT_IN_SKIP_PREFIXES, line 24); '/api/v1/desktop-ws/' viewer routes share the same global per-IP bucket (default 300/60s, index.ts:397) used by dashboard/auth calls, so a stuck poll loop can exhaust it and get the console's own requests 429'd. All three mechanics match the reported repro exactly.

## Affected code

- `apps/viewer/src/lib/webrtc.ts:42-46` — isSessionEndedResponse regex misses 'Session closed', 'Missing viewer token', 'Invalid or expired viewer token'
- `apps/viewer/src/lib/webrtc.ts:64,217-250` — 50ms poll interval with up to 15s of continued polling on non-matching 401s
- `apps/api/src/routes/desktopWs.ts:192-208` — Actual 401 error strings returned by GET /:id/viewer/session
- `apps/api/src/routes/desktopWs.test.ts:242-285` — Confirms exact 401 body strings the API returns for these branches
- `apps/api/src/middleware/globalRateLimit.ts:18-25` — Skip-prefix list excludes only /api/v1/agents/ and /api/v1/helper/, not desktop-ws viewer routes
- `apps/api/src/index.ts:397` — globalRateLimit() applied to all routes by default

## Potential fix

Broaden isSessionEndedResponse (or just treat any 401 from /viewer/session and /viewer/offer as terminal, per the existing code comment's own stated intent) so pollForAnswer stops on the first 401 regardless of message text. Separately, consider slowing ANSWER_POLL_INTERVAL_MS and/or adding the desktop-ws viewer polling path to globalRateLimit's skip prefixes (or giving it its own bucket) so a stuck client poll can't exhaust the per-IP budget used by the dashboard/auth.

ToddHebebrand:
Closes #3041

A remote-desktop session that died mid-connect could lock the operator out of their own console: `213 x 401 in ~26 seconds` from the viewer's answer poll drained the per-IP rate-limit bucket, and the next dashboard refresh came back `429 {"error":"Too many requests"}` and dumped them on the login screen.

Three defects compounded. Each is fixed independently, so no single one has to hold.

### 1. The poll didn't stop on 401 (viewer)

`isSessionEndedResponse()` only treated a 401 as terminal if the body matched `/session\s*ended|ended|revoked|no longer active/i`. The 401s `validateViewerSessionAccess` actually returns are `Session closed`, `Missing viewer token`, `Invalid or expired viewer token`, and `Viewer token revoked` — **none of the first three match**, so the poll kept firing every 50ms for the full 15s against a session that could never come back.

Now it matches on status alone. A viewer token authorizes exactly one still-live session, so every 401 on this path is terminal — the function's own comment already said as much; the regex just didn't implement it. Matching prose was the bug, so the body is no longer consulted at all.

Two adjacent loops closed while here:
- **Terminal 400/403/404** (not-a-desktop-session, policy-denied, session-not-found) also ran to timeout. They now end the attempt, as a plain `Error` — a policy denial is not a dead session and must not masquerade as one.
- **429** is now honoured via `Retry-After`: if the server's backoff outlasts the answer window, the attempt gives up instead of spending it issuing requests the server already refused.

### 2. A flat 50ms interval could exhaust the bucket by itself (viewer)

15s / 50ms = ~300 requests for **one** connection attempt — the entire default 300/60s budget. The poll now backs off geometrically (50ms, x1.5, capped at 500ms): a full 15s wait costs ~35 requests, while a healthy agent answering in a few hundred ms is picked up just as fast as before.

### 3. Viewer polls shared the dashboard's bucket (API)

`/api/v1/desktop-ws/` was metered in the same per-IP bucket as the dashboard and auth, so the burst starved the console.

It now gets its own bucket — **isolation, not exemption**. Adding it to the existing skip-prefix list was considered and rejected: that would remove all abuse protection from a public token-authenticated endpoint. It stays capped at 600/min; it just can no longer touch the budget the console draws on. Not counting 401s was also rejected — failed auth is exactly what a rate limiter should count.

### 4. The symptom: a 429 on `/auth/refresh` evicted a valid session (web)

This is why the reporter hit the *login screen* rather than just a failed request. `refreshFetchOnce()` classified anything that wasn't a raced 401 or a 5xx as a hard auth failure, so a 429 fell through to `auth-failed` and evicted the session. But a 429 is the rate limiter rejecting the request before it was ever evaluated — no verdict was reached on the refresh cookie, so it is transient, exactly like the 502 case fixed in QA 2026-07-08. It is now retried with the existing backoff.

Also: the in-memory rate-limit fallback returned a 429 with no `Retry-After` on its map-full path, leaving well-behaved clients no backoff signal. Fixed.

## Testing

- `apps/api/src/middleware/globalRateLimit.test.ts` — viewer polls don't drain the shared bucket; the viewer bucket is still capped (isolation != exemption, asserted with `Retry-After` + `X-RateLimit-Limit`); shared-bucket exhaustion doesn't throttle viewer traffic; the shipped default really does isolate `/api/v1/desktop-ws/`.
- `apps/viewer/src/lib/sessionEnded.test.ts` — the poll stops after **exactly one** request for each real API 401 body (table-driven over all four, plus the JSON-wrapped form); terminal 403 ends the attempt without masquerading as `SessionEndedError`; a 429 with `Retry-After: 60` gives up instead of polling; backoff keeps a full 15s wait under 40 requests while staying fast at the start.
- `apps/web/src/stores/auth.test.ts` — a 429 on `/auth/refresh` is retried and the session survives.

**Results:** API 6/6, viewer 21/21, web 46/46, `extensions/loader.test.ts` 49/49 (it also exercises this middleware). `tsc --noEmit` clean on all three packages. No schema, migration, or tenant-table changes.

## Follow-ups (deliberately not in scope)

- Issue ask #4 wanted the console to show a **cooldown** instead of a login bounce. This PR stops the incorrect eviction, but `AuthGuard` still collapses `transient` to a boolean and redirects once the bounded retries are exhausted. Showing a real rate-limited state needs new UI states — worth its own issue.
- `pollForAnswer` takes no `AbortSignal`, so an unmount mid-attempt isn't noticed until the attempt returns (now bounded and much quieter, but not zero).

### DIFF
--- a/apps/api/src/middleware/globalRateLimit.test.ts
+++ b/apps/api/src/middleware/globalRateLimit.test.ts
@@ -1,13 +1,26 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { Hono } from 'hono';
 
+// Hoisted so individual tests can flip between the in-memory fallback
+// (getRedis -> null) and the Redis-backed path that production actually runs.
+const mocks = vi.hoisted(() => ({
+  getRedis: vi.fn<() => unknown>(() => null),
+  rateLimiter: vi.fn(),
+}));
+
 vi.mock('../services/redis', () => ({
-  getRedis: () => null,
+  getRedis: mocks.getRedis,
+}));
+
+vi.mock('../services/rate-limit', () => ({
+  rateLimiter: mocks.rateLimiter,
 }));
 
 vi.mock('../services/clientIp', () => ({
   getTrustedClientIp: vi.fn(() => 'global-rate-limit-test'),
 }));
+
+const TEST_IP = 'global-rate-limit-test';
 
 import {
   __resetInMemoryCountersForTests,
@@ -20,6 +33,9 @@ import {
 beforeEach(() => {
   __resetSkipPrefixesForTests();
   __resetInMemoryCountersForTests();
+  // Default to the in-memory fallback; the Redis-path tests opt in.
+  mocks.getRedis.mockReturnValue(null);
+  mocks.rateLimiter.mockReset();
 });
 
 describe('globalRateLimit', () => {
@@ -102,6 +118,37 @@ describe('globalRateLimit — isolated desktop-ws bucket', () => {
     expect(throttled.status).toBe(429);
     expect(throttled.headers.get('Retry-After')).toBe('60');
     expect(throttled.headers.get('X-RateLimit-Limit')).toBe('2');
+  });
+
+  // Production always has Redis, so the in-memory assertions above exercise the
+  // fallback rather than the real path. Pin the key/limit actually handed to the
+  // Redis limiter too.
+  it('hands the isolated key and limit to the Redis limiter', async () => {
+    const fakeRedis = { marker: 'redis' };
+    mocks.getRedis.mockReturnValue(fakeRedis);
+    mocks.rateLimiter.mockResolvedValue({
+      allowed: true,
+      remaining: 599,
+      resetAt: new Date(Date.now() + 60_000),
+    });
+    const app = buildApp({ limit: 300, windowSeconds: 60 });
+
+    await app.request(VIEWER_PATH);
+    expect(mocks.rateLimiter).toHaveBeenCalledWith(
+      fakeRedis,
+      `global:desktopws:${TEST_IP}`,
+      600,
+      60,
+    );
+
+    mocks.rateLimiter.mockClear();
+    await app.request('/api/v1/devices');
+    expect(mocks.rateLimiter).toHaveBeenCalledWith(
+      fakeRedis,
+      `global:${TEST_IP}`,
+      300,
+      60,
+    );
   });
 
   it('exhausting the shared bucket does not throttle viewer traffic', async () => {
--- a/apps/api/src/middleware/globalRateLimit.test.ts
+++ b/apps/api/src/middleware/globalRateLimit.test.ts
@@ -10,13 +10,16 @@ vi.mock('../services/clientIp', () => ({
 }));
 
 import {
+  __resetInMemoryCountersForTests,
   __resetSkipPrefixesForTests,
   globalRateLimit,
+  ISOLATED_BUCKETS,
   registerGlobalRateLimitSkipPrefix,
 } from './globalRateLimit';
 
 beforeEach(() => {
   __resetSkipPrefixesForTests();
+  __resetInMemoryCountersForTests();
 });
 
 describe('globalRateLimit', () => {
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer. Design was additionally pre-reviewed by `codex exec` (read-only, xhigh) before implementation, since this touches rate-limiting on an auth-adjacent path.

**Findings:** 3 raised → 3 addressed in `26734b2d4`; 1 raised → deliberately deferred with reasoning; 0 outstanding.

- **code-reviewer — 0 high-confidence findings.** Traced bucket-key collision (not reachable: `getTrustedClientIp` only returns values passing `isIP()`, so an IP can never spell `desktopws:<ip>`), confirmed every status in `pollForAnswer` is reachable and correctly classified, swept both call sites of the changed `isSessionEndedResponse` signature, and confirmed the 429→`transient` reclassification stays bounded and still evicts a genuinely dead session.
- **pr-test-analyzer — 4 gaps, all closed.** Every pre-existing test settled on the *first* poll response, so nothing proved the fix stops the hammering without also breaking legitimate retries. Added fake-timer tests that drive the real multi-iteration loop (5xx ridden out; short `Retry-After` honoured; 429 without `Retry-After` stays bounded), parametrized the terminal-4xx test over 400/403/404 instead of 403 alone, and covered the rate limiter's **Redis** branch — the one production actually runs, since the suite stubs Redis to `null` throughout.
- **silent-failure-hunter — 1 finding addressed, 1 deferred.** Addressed: a 5xx during the poll was retried with zero logging, so a real outage was indistinguishable from a slow agent (both surfaced as "Timed out waiting for WebRTC answer from agent"). The poll now logs the first of each distinct error status and the timeout reports the last response it saw.

**Deferred finding (worth a maintainer's eye):** the hunter argued a terminal 403 should get its own error type so the transport rethrows it instead of falling back to WebSocket, where the user sees "Retry to reconnect" for a permanent policy denial. Not done here, for two reasons: (1) it is **pre-existing** — before this PR a 403 also ended in a generic error, just 300 requests and 15s later; (2) code-reviewer independently traced the same path and reached the *opposite* conclusion, that a mid-poll policy denial is not "session dead forever" the way a viewer-token 401 is, so the generic error is the right classification and the existing fallback is correct. Surfacing a distinct terminal-denial state means new UI states in `DesktopViewer.tsx`, which I can't verify without a live stack. Flagged as follow-up in the PR body rather than guessed at.

**Tests:** `apps/api` 7/7 (`globalRateLimit.test.ts`), `apps/viewer` 24/24 (`sessionEnded.test.ts`) + 26/26 with `webrtc.test.ts`, `apps/web` 46/46 (`stores/auth.test.ts`), plus `extensions/loader.test.ts` 49/49 (it also exercises this middleware). All single-fork. `tsc --noEmit` clean on api, viewer, and web. No schema, migration, or tenant-table changes, so no RLS/cascade/export-policy registration applies.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3041 — the reporter should confirm first.

SemoTech:
Subscribing SemoTech for updates.

SemoTech:
Watching this PR for merge / release notes.

bdunncompany:
The red here looks like a hung runner, not this branch.

`Check Migrations (non-superuser)` ran for **46m00s** and its step conclusions stop at setup:

```
3. Checkout:          success
4. Setup pnpm:        success
5. Setup Node.js:     success
6. Install dependencies:                    null
7. Apply migrations as a non-superuser role: null
```

Steps 6 onward never reported — so it stalled during `Install dependencies` and was killed on timeout rather than failing a check. `CI Success` is red only because it aggregates ("Check all jobs passed").

Supporting it: this PR touches no migrations at all — `globalRateLimit`, `viewer/webrtc`, `web/stores/auth`. There is nothing in the diff for that job to exercise.

A re-run should clear it.

SemoTech:
Thanks @bdunncompany — that matches what we’re seeing. Regular `Check Migrations` and the rest of CI are green; only the non-superuser job hung during `Install dependencies`, and this PR doesn’t touch migrations anyway.

Appreciate the re-run. We’re subscribed here and on #3041 for merge / release notes; happy to validate on our self-hosted instance once it lands.

PR #3378, opened 2026-08-10
fix(snmp): make the live device row the sole org authority for poll dispatch (#3226)

### LINKED ISSUE #3226: snmpWorker: poll dispatch mixes two org sources — stale job payload can ship decrypted SNMP credentials to an agent in the wrong org
Spotted during the review of #3215 (PR #3220) but **not a regression from it** — the shape is identical on `main`, so it is being split out rather than folded into that fix.

**What's wrong.** `processPollDevice` (`apps/api/src/jobs/snmpWorker.ts`) resolves the org twice, from two different sources, for a single dispatch:

- the device row is loaded by id with no org predicate at all (`snmpWorker.ts:100`, system context, so RLS doesn't constrain it)
- the template lookup scopes on **`device.orgId`** — the row (`snmpWorker.ts:119`)
- the online-agent pick scopes on **`data.orgId`** — the BullMQ job payload, captured whenever the job was enqueued (`snmpWorker.ts:145`)

Those two agree only as long as the device's `org_id` hasn't changed since the job was created.

**Why that's more than cosmetic.** The command built at the end of that function carries decrypted SNMP credentials — the v1/v2c community string, and v3 auth/priv passwords. The agent that receives them is chosen by `data.orgId`. So if a device is moved between orgs while a poll job for it is queued, the credentials for a device that now belongs to org B are handed to an online agent in org A.

The stable jobId makes this durable rather than a narrow race: `enqueueSnmpPoll` reuses an existing job when it is in a reusable state (`snmpWorker.ts:277-285`), so a payload built before the move is not superseded by a later enqueue — it survives and is reused with its stale `orgId`.

**Fix shape.** Treat the device row as the single source of truth for the org and stop trusting the payload — select the agent on `device.orgId`, and either drop `orgId` from the job data or assert it matches the row and fail the job loudly when it doesn't. Whichever way, the mismatch deserves a log line rather than a silent divergence, since today nothing surfaces it.

Worth checking the same pass whether the unscoped device load at `:100` should carry an explicit org predicate for defence in depth, and whether `moveOrg` should drain or invalidate queued `snmp-poll-*` jobs for the device it moves.

Staying open — no code change yet; #3215/#3220 deliberately left this alone to keep that diff scoped.

ToddHebebrand:
Closes #3226

## The bug

`processPollDevice` (`apps/api/src/jobs/snmpWorker.ts`) resolved the org **twice, from two different sources**, for a single dispatch:

| Consumer | Org source |
|---|---|
| template OID lookup | `device.orgId` — the live `snmp_devices` row |
| online-agent pick | `data.orgId` — the BullMQ job payload |

Those agree only while the device's `org_id` hasn't changed since the job was created.

The command built at the end of that function carries **decrypted SNMP credentials** — the v1/v2c community string and the v3 auth/priv passwords — and the agent that receives them was chosen by the payload. So a device moved between orgs while a poll sat queued had its secrets handed to an online agent in the **old** org.

The stable jobId makes this **durable rather than a narrow race**: `enqueueSnmpPoll` returns early when it finds `snmp-poll-<deviceId>` in a reusable state, so a later enqueue carrying the correct org never supersedes the stale payload — it survives and is reused indefinitely.

Verified still present on `origin/main` (669a3551a) before starting; no prior PR referenced the issue in any state.

## The fix

**One authority: the live row.** Both consumers now read `device.orgId`, and the payload is reconciled against it.

**Worker-side fence.** A payload org that disagrees with the row aborts the dispatch *before* the template read and *before* any credential decrypt, then **fails the job loudly** so it lands in `attachWorkerObservability` and the worker's `failed` handler instead of scrolling past as one more console warning among the routine no-agent lines.

Two subtleties worth flagging for review:

- **The throw is raised after the phase-1 DB context commits, not inside it.** The same throw inside that callback would roll the transaction back and take `markPollAttempted`'s stamp with it — leaving `lastPollAttemptedAt` NULL so the scheduler re-selects the device on every 60s tick. That is exactly the hot loop #3217 fixed.
- **The failure is self-healing, on the device's polling cadence.** No `attempts` option is set, so the job settles in `failed`. The attempt stamp is what keeps the device off the next few 60s ticks, so recovery lands the next time it is genuinely *due* — one `pollingInterval` later, 300s by default (`consecutiveFailures` is untouched here, so no backoff multiplier applies). At that point `enqueueSnmpPoll` sees the non-reusable `failed` state, removes it, and enqueues a fresh job carrying the current org. Worst case is one skipped poll interval for a device that just changed org. This branch is the backstop, not the usual route — the enqueue-side fence normally replaces a drifted payload before a worker picks it up.

**Enqueue-side fence.** A queued job whose payload org has drifted from the org the caller just read from the live row is removed and re-added rather than reused, so in the normal path the stale payload never reaches the worker at all. An `active` job is deliberately left alone — BullMQ rejects removing a locked job, and the worker-side guard refuses that in-flight run on the same mismatch.

**`orgId` stays on the payload deliberately.** The issue offered "drop it or assert it"; dropping it would remove the only signal that a divergence ever happened. It is now documented as advisory — used solely as the staleness tripwire, never as an input.

**A payload with no `orgId` counts as stale, not as a wildcard** — fail-closed on a credential path.

## On the two open questions in the issue

- *"Should the unscoped device load carry an explicit org predicate for defence in depth?"* — **No, and there's now a comment saying so.** Adding `eq(snmpDevices.orgId, data.orgId)` re-introduces payload trust through the back door: a moved device would report as `device-missing`, hiding the very divergence the guard exists to surface. Load by primary key, then reconcile.
- *"Should `moveOrg` drain queued `snmp-poll-*` jobs?"* — not needed for correctness now that both fences are in place, and `snmp_devices` has no `device_id` so it isn't on the device-move path anyway. Left out to keep the diff scoped.

## Verification

- `apps/api/src/jobs/snmpWorker.orgAuthority.test.ts` — **new, 12 cases**, covering: mismatch refuses + throws; no `decryptSnmpSecret` call and no `sendCommandToAgent` on mismatch; abort happens before the agent lookup; the attempt stamp still lands (no #3217 hot loop); missing `orgId` treated as stale; agent predicate built from the row org; happy path still dispatches; and the five enqueue-side reuse/replace cases.
- **Falsifiability checked, and it mattered.** A first draft of the mismatch tests wired only the device select, so they passed against the *unfixed* worker — the run died on an unstubbed `db.select()` one step later rather than on the guard. They now wire the full dispatch chain, so the pre-fix code genuinely reaches `sendCommandToAgent` with the decrypted secrets. **7 of 12 fail on the unfixed worker**, including the "no credentials on the wire" assertion.
- All four SNMP suites green single-fork: `snmpWorker.orgAuthority`, `snmpWorker.dbcontext`, `snmpQueue`, `snmpWorkerScheduler` — **39 passed**.
- **No RLS/integration suites run:** no local database was available. This change adds no table, column, or migration and does not alter RLS context usage — it only changes which in-memory value a `WHERE` predicate is built from — so no cascade or export-policy registration applies.

## Review round

Reviewed by `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` and a dedicated security pass. Three findings, all addressed in `73c1cd98c`:

1. **Inaccurate comment (and PR body).** The `org-mismatch` branch claimed recovery happens on "the next scheduler tick". It doesn't — `markPollAttempted` stamps `lastPollAttemptedAt`, which is precisely what keeps the device off the next few 60s ticks. Corrected above and in the code.
2. **Unguarded `existing.remove()`** in the new drifted-reuse path. A `waiting` job can be locked between the `getState()` snapshot and the removal (concurrency is 10), and this path is what newly exposes `remove()` to a lockable job — the pre-existing terminal-state removals can't hit it. It threw into `processScheduler`'s per-device catch, which logs without Sentry. Now handled explicitly: the race collapses into the case the `active` branch already delegates to the worker-side guard, so it logs with the error and returns the job that is actually queued.
3. **Two untested branches the fix touched** — the `no-agent` and `isAgentConnected === false` log lines, both switched off the payload org. Neither dispatches, so neither is a credential path, but a regression there would silently reintroduce a stale-payload read. The `isAgentConnected === false` branch had no coverage anywhere in the repo.

Security pass returned **no CRITICAL/HIGH/MEDIUM findings**: no remaining path lets `data.orgId` influence credential selection or the delivery target, the guard can't be bypassed (`snmp_devices.org_id` is `NOT NULL`, strict `!==`, missing payload org fails closed), the throw is confirmed outside the phase-1 transaction, and no other SNMP dispatch path shares the two-source shape (`buildSnmpPollCommand` has exactly one caller; `monitorWorker` already sources org from its live row).

### DIFF
--- a/apps/api/src/jobs/snmpWorker.orgAuthority.test.ts
+++ b/apps/api/src/jobs/snmpWorker.orgAuthority.test.ts
@@ -1,0 +1,395 @@
+/**
+ * snmpWorker org authority (issue #3226).
+ *
+ * The regression these tests lock down is a cross-tenant credential
+ * disclosure. `processPollDevice` used to resolve the org from TWO sources for a
+ * single dispatch: the template was scoped on the live `snmp_devices` row while
+ * the online-agent pick was scoped on the BullMQ job payload, captured whenever
+ * the job was enqueued. The command built at the end of that function carries
+ * DECRYPTED SNMP credentials (v1/v2c community string, v3 auth/priv passwords),
+ * and the agent receiving them was chosen by the payload — so a device moved
+ * between orgs while a poll sat queued had its secrets handed to an online agent
+ * in the OLD org.
+ *
+ * The stable jobId (`snmp-poll-<deviceId>`) made that durable rather than a
+ * narrow race: `enqueueSnmpPoll` returns early when it finds the job in a
+ * reusable state, so a later enqueue carrying the correct org never supersedes
+ * the stale payload.
+ *
+ * Two fences, tested here:
+ *   1. worker-side  — the live row is the sole org authority, and a payload that
+ *                     disagrees fails the job loudly before any decrypt.
+ *   2. enqueue-side — a queued job whose payload org has drifted is replaced
+ *                     instead of reused.
+ */
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockDb, queueMock, agentWsMock, decryptMock, eqCalls } = vi.hoisted(() => ({
+  mockDb: {
+    select: vi.fn(),
+    insert: vi.fn(),
+    update: vi.fn(),
+  },
+  queueMock: {
+    getJob: vi.fn(async () => null),
+    add: vi.fn(async () => ({ id: 'job-1' })),
+    addBulk: vi.fn(async () => []),
+    getRepeatableJobs: vi.fn(async () => []),
+    removeRepeatableByKey: vi.fn(async () => undefined),
+    close: vi.fn(async () => undefined),
+  },
+  agentWsMock: {
+    isAgentConnected: vi.fn(() => true),
+    sendCommandToAgent: vi.fn(() => true),
+  },
+  decryptMock: vi.fn((v: string | null) => v),
+  // Records every drizzle `eq(column, value)` so a test can prove WHICH org
+  // value the online-agent predicate was built from.
+  eqCalls: [] as Array<[unknown, unknown]>,
+}));
+
+vi.mock('bullmq', () => ({
+  Queue: class {
+    getJob = queueMock.getJob;
+    add = queueMock.add;
+    addBulk = queueMock.addBulk;
+    getRepeatableJobs = queueMock.getRepeatableJobs;
+    removeRepeatableByKey = queueMock.removeRepeatableByKey;
+    close = queueMock.close;
+  },
+  Worker: class {
+    close = vi.fn();
+    on = vi.fn();
+  },
+  Job: class {},
+}));
+
+vi.mock('drizzle-orm', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('drizzle-orm')>();
+  return {
+    ...actual,
+    eq: (column: unknown, value: unknown) => {
+      eqCalls.push([column, value]);
+      return actual.eq(column as never, value as never);
+    },
+  };
+});
+
+vi.mock('../db', () => ({
+  db: mockDb,
+  withSystemDbAccessContext: async (fn: () => unknown) => fn(),
+  assertOutsideHeldDbContext: vi.fn(),
+}));
+
+vi.mock('../db/schema', () => ({
+  snmpDevices: {
+    id: 'snmpDevices.id',
+    orgId: 'snmpDevices.orgId',
+    isActive: 'snmpDevices.isActive',
+    lastPolled: 'snmpDevices.lastPolled',
+    lastPollAttemptedAt: 'snmpDevices.lastPollAttemptedAt',
+    consecutiveFailures: 'snmpDevices.consecutiveFailures',
+    lastStatus: 'snmpDevices.lastStatus',
+    pollingInterval: 'snmpDevices.pollingInterval',
+  },
+  snmpMetrics: { deviceId: 'snmpMetrics.deviceId' },
+  snmpTemplates: {
+    id: 'snmpTemplates.id',
+    oids: 'snmpTemplates.oids',
+    isBuiltIn: 'snmpTemplates.isBuiltIn',
+    orgId: 'snmpTemplates.orgId',
+  },
+  devices: {
+    agentId: 'devices.agentId',
+    orgId: 'devices.orgId',
+    isEphemeral: 'devices.isEphemeral',
+    status: 'devices.status',
+  },
+}));
+
+vi.mock('../services/redis', () => ({
+  getBullMQConnection: vi.fn(() => ({ host: 'localhost', port: 6379 })),
+}));
+
+vi.mock('../routes/agentWs', () => ({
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, silent-failure-hunter, pr-test-analyzer) plus a dedicated security-focused pass, warranted by the tenant-isolation blast radius.

**Findings:** 3 raised → all 3 addressed in `73c1cd98c`; 0 outstanding.
- *code-reviewer (important):* the `org-mismatch` comment claimed recovery on "the next scheduler tick" — wrong, and it contradicted the rationale four lines above it. `markPollAttempted` stamps `lastPollAttemptedAt`, which is exactly what keeps the device off the next few 60s ticks; recovery is one `pollingInterval` (300s default). Corrected in code and in the PR body.
- *silent-failure-hunter (medium-high):* `existing.remove()` in the new drifted-reuse path was unguarded against the job locking between `getState()` and removal. This path is what newly exposes `remove()` to a lockable job — the pre-existing terminal-state removals can't hit it — and unhandled it threw into `processScheduler`'s per-device catch, which logs without Sentry. Now handled rather than swallowed: the race collapses into the case the `active` branch already delegates to the worker-side guard, so it logs with the error and returns the job that is actually queued.
- *pr-test-analyzer (5/10, 4/10):* the `no-agent` and `isAgentConnected === false` log lines were switched off the payload org but left untested; the latter had no coverage anywhere in the repo. Both covered now, plus the `remove()` race.

**Security pass:** no CRITICAL / HIGH / MEDIUM findings. Confirmed no remaining path lets `data.orgId` influence credential selection or the delivery target; the guard can't be bypassed (`snmp_devices.org_id` is `NOT NULL`, strict `!==`, a missing payload org fails closed); the throw is genuinely outside the phase-1 transaction, so the attempt stamp survives; the throw path can't retry-storm (no `attempts` set); and no other SNMP dispatch path shares the two-source shape — `buildSnmpPollCommand` has exactly one caller, and `monitorWorker` already sources its org from the live row.

**Tests:** all four SNMP suites green single-fork — `snmpWorker.orgAuthority` (15), `snmpWorker.dbcontext`, `snmpQueue`, `snmpWorkerScheduler` — **42 passed**. `npx tsc --noEmit` on `apps/api` clean.

**Falsifiability:** re-verified by swapping in `origin/main`'s `snmpWorker.ts` — **8 of the 15 new cases fail**, including the "no credentials on the wire" assertion, where the pre-fix code resolves `{ dispatched: true, agentId: 'agent-in-stale-org' }`. An earlier draft of those tests passed against the unfixed worker for the wrong reason (the run died on an unstubbed `db.select()` a step later); they now wire the full dispatch chain so the pre-fix code genuinely reaches `sendCommandToAgent`.

**Not run:** RLS/integration contract suites — no local database available. This change adds no table, column, or migration and does not alter RLS context usage, so no cascade or export-policy registration applies.

PR #3379, opened 2026-08-10
fix(ai): resolve remediate_vulnerability's org from the device, not accessibleOrgIds[0] (#3322)

### LINKED ISSUE #3322: [AI] Partner-scope chat: remediate_vulnerability silently no-ops for devices outside the first accessible org
## Description
In a partner-scope AI chat session, asking the assistant to remediate a CVE on a device that is not in the session's first accessible org fails with \"device not found\" after the Tier-3 approval is granted — the remediation is approved, then silently skips (\"Scheduled 0; 1 skipped\" / device not found). The AI audit rows also land under that arbitrary first org rather than the device's org, so an org-filtered audit view shows no trace of the AI action.

## Root Cause
The AI session's org resolution falls back to `auth.orgId ?? accessibleOrgIds[0]`. For partner-scope tokens `orgId` is null, so every tool call is silently pinned to whichever org happens to sort first, and `remediate_vulnerability`'s device lookup then fails the org gate for any device in another org. Same fallback shape as the known aiTools org-fallback bug class.

## Proposed Fix
Partner-scope sessions should resolve the org from the tool's target (device → device.orgId) rather than a fixed fallback, or thread an explicit org context through the chat session. Audit rows for device-targeted AI actions should be attributed to the device's org.

## Reproduction
Partner-scope chat, All-organizations view: \"CVE-2025-6554 is on <device in org B>. Remediate it now.\" → tool runs, Tier-3 approval granted → remediation skips with device-not-found; works when the device is in the first accessible org.

## Reported By
Found while producing marketing captures against a seeded demo stack (2026-08-09); cost one full take.

ToddHebebrand:
## Problem

In a **partner-scope** AI chat session the caller has no home org (`auth.orgId` is null). `remediate_vulnerability` fell back to `auth.accessibleOrgIds[0]` and passed that arbitrary org into `remediateVulnerabilities`, whose defense-in-depth pin check

```ts
if (orgId && device.orgId !== orgId) { result.skipped.push({ id: dvId, reason: 'device_not_found' }); continue; }
```

then rejected **every device outside whichever org happened to sort first**. The Tier-3 approval was granted, the tool answered `Scheduled 0; 1 skipped`, no command was queued — and because the skip happens before step 6, **no audit row was written at all**, so an org-filtered audit view showed no trace of the AI action.

## Root cause

The core was already correct for org-less callers. It derives each finding's org from its **own device** and attributes both the `vulnerability.remediate` audit row and the `vulnerability.remediation_scheduled` domain event to `device.orgId`. `POST /vulnerabilities/remediate` already opted into that by passing `auth.orgId ?? ''`. The AI tool was the only caller that *guessed* an org.

## Fix

- **`services/aiToolsVulnerability.ts`** — pass `auth.orgId ?? null` (the caller's **real** pin) instead of `getOrgId`'s `accessibleOrgIds[0]` fallback. Org-scope callers still pin to their org; partner/system callers pass `null` and the core resolves per device. The guard now refuses only a caller with neither an org pin nor any reachable org.
- **`services/vulnerabilityRemediation.ts`** — parameter renamed to `pinnedOrgId: string | null` with the contract documented, replacing the `''` sentinel so "derive per device" is explicit and a future caller cannot pass a guessed org by accident.
- **`routes/vulnerabilities.ts`** — passes `auth.orgId` directly now that `null` is the signal.
- The tool's `message` now names the **distinct skip reasons**. A bare "N skipped" read as a benign no-op and the model narrated this exact failure as success.

### Tenant isolation

This **widens nothing**. Reach is still bounded by `auth.orgCondition(devices.orgId)` (an `IN`-list over `accessibleOrgIds` for partner scope) plus RLS on the `withDbAccessContext` the SDK tool path opens from `dbAccessContextFromAuth(auth)`. The change only stops discarding orgs the caller can *already* reach. The org pin remains enforced for callers who genuinely have one — covered by a test. Worker/child rows (audit, event) continue to take the **device's** org, per the fan-out rule in CLAUDE.md.

No schema, migration, or cascade/export-policy surface is touched.

## Tests

`services/vulnerabilityRemediation.test.ts` gains a table-keyed Drizzle chain stub and three cases:
- a null-pin caller remediates a device in the **second** accessible org (the regression);
- audit row + domain event are attributed to the **device's** org;
- an org-scoped caller pinned to Org A still gets `device_not_found` for an Org B device, and queues/audits nothing.

`services/aiToolsVulnerability.test.ts` adds partner-scope and system-scope null-forwarding, the stranded-caller refusal, and the distinct-skip-reason message.

```
vitest run --no-file-parallelism src/services/aiToolsVulnerability.test.ts src/services/vulnerabilityRemediation.test.ts   → 25 passed
vitest run --no-file-parallelism src/routes/vulnerabilities.test.ts                                                        → 58 passed
tsc --noEmit                                                                                                               → clean
```

## Out of scope

`ai_sessions` / `ai_tool_executions` stay under the **session** org — a partner chat session legitimately spans orgs, and re-anchoring the session ledger per tool call is a separate design question. The domain audit row, which is what an org-filtered audit view reads, is now correct.

Closes #3322

### DIFF
--- a/apps/api/src/routes/vulnerabilities.ts
+++ b/apps/api/src/routes/vulnerabilities.ts
@@ -773,10 +773,10 @@ vulnerabilityRoutes.post(
   async (c) => {
     const auth = c.get('auth');
     const { deviceVulnerabilityIds } = c.req.valid('json');
-    // Org-scope callers pass their org; partner/system callers pass '' so the
+    // Org-scope callers pass their org; partner/system callers pass null so the
     // core derives the org per-device (auth.orgId is null off org scope).
     const result = await remediateVulnerabilities(
-      auth.orgId ?? '',
+      auth.orgId,
       deviceVulnerabilityIds,
       auth.user.id,
       auth,
--- a/apps/api/src/services/aiToolsVulnerability.test.ts
+++ b/apps/api/src/services/aiToolsVulnerability.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import { VULN_SKIP_REASON_LABELS } from '@breeze/shared';
 
@@ -23,6 +23,13 @@ const fakeAuthCtx = {
 } as unknown as AuthContext;
 
 describe('registerVulnerabilityTools', () => {
+  // Each spy below is restored inline, but an assertion that throws skips its
+  // own `mockRestore()` and leaks the spy's call history into the next test —
+  // which turns one real regression into a cascade of misleading failures.
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
   it('registers the two read tools as tier 1', () => {
     const map = registered();
     expect(map.get('get_vulnerability_report')!.tier).toBe(1);
--- a/apps/api/src/services/aiToolsVulnerability.test.ts
+++ b/apps/api/src/services/aiToolsVulnerability.test.ts
@@ -1,5 +1,7 @@
 import { describe, expect, it, vi } from 'vitest';
 
+import { VULN_SKIP_REASON_LABELS } from '@breeze/shared';
+
 import { registerVulnerabilityTools } from './aiToolsVulnerability';
 import { TOOL_PERMISSIONS } from './aiGuardrails';
 import { toolInputSchemas } from './aiToolSchemas';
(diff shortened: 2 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (3 agents, parallel).

**Findings:** 4 raised → 4 addressed in `768528bff`; 0 outstanding.

- **Tenant isolation (code-reviewer): clean, no widening.** Traced `buildOrgAccessClosures` (`middleware/auth.ts:224-250`), `verifyDeviceAccess`/`executeTool` (`services/aiTools.ts`), and the context built in `makeHandler` (`aiAgentSdkTools.ts:337-404`). Per scope: org-scope is byte-identical (`auth.orgId` non-null); partner-scope reach is still bounded by `orgCondition` + RLS from the same `accessibleOrgIds`; system scope was already unrestricted everywhere else. **No scope relied on the `accessibleOrgIds[0]` pin as its only protection.** Caller sweep complete — only two call sites, both updated. Guard `accessibleOrgIds === null || length > 0` verified fail-closed.
- **`not_open` conflated "invisible" with "already fixed"** (silent-failure-hunter). A missing `deviceVulnerabilities` row — bad id, or a tenant outside the caller's reach — reported as *"not an open vulnerability"*, i.e. "already done". Split out as `not_found`, which already exists in the shared skip vocabulary and is what the sibling bulk endpoints return. Same misleading-no-op class as this issue. **Fixed + tested.**
- **Stale file header** sent a reviewer to a false conclusion — it claimed handlers run outside a DB context (pre-#2822), so the reviewer read `readDeviceFindings`' inner single-org `withDbAccessContext` as blacking the Tier-1 read tools out to `accessibleOrgIds[0]`. It does not: every live entry point already wraps `executeTool` in `withDbAccessContext(dbAccessContextFromAuth(auth))`, and `withDbAccessContext` returns `fn()` untouched when a context is open (`db/index.ts:440`), so that wrap is an inert safety net. **Header corrected** with a do-not-"fix"-this note, since turning it into a real narrowing would reintroduce this bug on the read path.
- **Test gaps** (pr-test-analyzer, verified by live mutation runs). Confirmed the regression guards are load-bearing: reverting the handler to `getOrgId(auth)` fails the partner- and system-scope forwarding tests; neutering the pin check to `if (false && …)` fails `still enforces a REAL org pin`. Added the two gaps worth closing: **one batch spanning two orgs** (both scheduled, each audit row + event re-tenanted to its OWN device org — the CLAUDE.md fan-out rule, and the first real exercise of the stub's multi-shift FIFO), and **the SITE axis still denying under a null org pin**. Also added `afterEach(restoreAllMocks)` — spies were restored inline, so a throwing assertion leaked call history and turned one regression into cascading, misdirecting failures.

**Tests:** `vitest run --no-file-parallelism src/services/vulnerabilityRemediation.test.ts src/services/aiToolsVulnerability.test.ts src/routes/vulnerabilities.test.ts` → **87 passed**. `tsc --noEmit` → clean.

**Status:** review-clean, awaiting maintainer merge. Not merging — full CI (incl. Integration Tests) should be green first.

PR #3382, opened 2026-08-10
fix(api): stop binding raw JS Dates into hand-written sql templates (#3369)

### LINKED ISSUE #3369: [API] Raw JS Date interpolated into Drizzle `sql` templates 500s the request — two remaining sites after #3329
Follow-up from #3329 / #3368. Fixing the `search_logs` cursor 500 surfaced a general defect class; a repo-wide sweep of ~1,110 `sql\`` templates in `apps/api/src` found **two more confirmed instances**. Filing separately because they are unrelated subsystems (one auth-adjacent) and each wants its own test.

## The defect class

A bare value interpolated into a Drizzle `sql` template is wrapped in a `Param` with the **noop** encoder, so the raw JS value goes straight to postgres.js. For a `Date` that means postgres.js' Bind step throws:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string
or an instance of Buffer or ArrayBuffer. Received an instance of Date
    at Bind (postgres/src/connection.js:954)
```

Drizzle's typed helpers (`eq`, `lt`, `gt`, `gte`, `lte`, `between`, `.where(...)`) are **safe** — they route the value through the column's encoder. Only hand-written `sql` templates are affected.

Because requests run inside `withDbAccessContext` — a postgres.js `begin()` transaction that re-throws query errors at commit even when the application handled them — this surfaces as an HTTP 500 for the whole request, not a handled error. That is what made #3329 so hard to read from the outside: the tool's own `catch` block ran and was bypassed.

## Site 1 — `services/filterEngine.ts` `applyOperator` (user-facing 500)

Lines ~374-441: `equals`, `notEquals`, `before`, `after`, `between` all interpolate `value` bare:

```ts
case 'before':   return sql`${columnRef} < ${value}`;
case 'between':  return sql`${columnRef} BETWEEN ${value.from} AND ${value.to}`;
```

`FilterValue` (`packages/shared/src/types/filters.ts`) includes `Date` and `{ from: Date; to: Date }`, and the validator (`packages/shared/src/validators/filters.ts`) uses `z.coerce.date()` — so after `zValidator` the value really is a live `Date` for any datetime field.

`columnRef` here is a `SQL<unknown>`, not a Column, so the typed-helper fix used in #3368 is not directly available; this likely wants a small `toDriverValue()` shim applied at the interpolation sites.

Reachable from:
- `POST /devices/filters/preview` (`routes/filters.ts`) — any `before`/`after`/`between`/`equals` on a datetime field (`enrolledAt`, `lastSeenAt`, `quarantinedAt`)
- dynamic device groups (`routes/groups.ts`, `services/groupMembership.ts`)
- deployment target resolution (`services/deploymentTargetResolver.ts`)
- the AI fleet tool (`services/aiToolsFleet.ts`)

## Site 2 — `oauth/revocationRetry.ts` (silent, and it defeats the retry)

Lines ~80 and ~84, in the `onConflictDoUpdate` branch:

```ts
nextAttemptAt: sql`${attemptedAt} + (LEAST(300, POWER(2, ${...attempts})) * interval '1 second')`,
expiresAt:     sql`GREATEST(${...expiresAt}, ${input.expiresAt})`,
```

`attemptedAt` is `new Date()`; `input.expiresAt` is typed `Date`.

This branch only runs when a retry row already exists — i.e. exactly when a prior Redis revocation write failed. `jobs/oauthRevocationRetryWorker.ts` re-invokes `writeOAuthRevocationMarkerDurably` for those rows every 30s. The throw escapes the `withSystemDbAccessContext` transaction (rolling back the whole batch) and is caught by the generic `.catch` in `runScheduledDrain`, which logs and returns 0. **Net effect: OAuth revocation retries would never drain or heal during a Redis outage**, and it fails quietly rather than per-row.

Note these interpolations sit in arithmetic/`GREATEST` expressions, so they likely need an explicit `::timestamptz` cast, not just an ISO string.

## Suggested scope

Two independent fixes. Site 2 is auth-adjacent and deserves coverage that actually exercises the conflict branch. A shared helper plus a lint rule or unit-level guard (render through `PgDialect.sqlToQuery` and assert no param is a `Date` — the pattern added in #3368's `logSearch.test.ts`) would stop the class recurring.

Sweep credit: found while fixing #3329, reported by @bdunncompany.

ToddHebebrand:
Closes #3369

## The defect class

A value interpolated into a Drizzle ``sql`...${value}...` `` template is wrapped in a `Param` carrying the **noop** encoder, so the untouched JS object is handed to postgres.js. For a `Date`, the driver's Bind step throws:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string
or an instance of Buffer or ArrayBuffer. Received an instance of Date
    at Bind (postgres/src/connection.js:954)
```

Drizzle's typed helpers (`eq`, `lt`, `gte`, `between`, `.set({...})`) are safe — they route the value through the **column's** encoder. Only hand-written templates are affected. And because requests run inside `withDbAccessContext`, a postgres.js `begin()` transaction that re-throws query errors at commit *even when the application already handled them*, this surfaces as an HTTP 500 for the whole request rather than a catchable error.

## The shared helper

New `apps/api/src/db/sqlValues.ts`:

| Helper | Emits | Use when |
|---|---|---|
| `sqlTimestamp(date)` | `$1` (ISO string, **no cast**) | default — comparisons, `INSERT`/`SET` targets |
| `sqlTimestamptz(date)` | `$1::timestamptz` | no column to infer from: arithmetic, `GREATEST`/`LEAST` |
| `sqlValue(unknown)` | delegates for `Date`, passes everything else through | values whose runtime type isn't statically known |

**The default is deliberately uncast, and that is the load-bearing design decision here.** The repo mixes both column flavours — `devices.last_seen_at` is `timestamp` **without** time zone, while `oauth_revocation_retries.expires_at` is `timestamptz`. A blanket `::timestamptz` would make Postgres reinterpret the naive column *in the session time zone*, silently shifting every comparison on any deployment not running UTC. That would have been a worse bug than the one being fixed: a wrong answer instead of a loud failure. `sqlTimestamp` matches Drizzle's own `PgTimestamp.mapToDriverValue` exactly, so Postgres resolves the parameter from whatever it is compared or assigned to. The integration test pins this by re-running the same assertions under `TimeZone = 'Pacific/Kiritimati'` (UTC+14).

## Site 1 — `services/filterEngine.ts` (live 500)

`applyOperator`'s `equals` / `notEquals` / `before` / `after` / `between` / `greaterThan…` / `lessThan…` / `matches` / `in` / `notIn` all interpolated the user value bare. `columnRef` there is an arbitrary `SQL` expression (a correlated subquery, a `CASE`, an `EXTRACT`), not a `Column`, so the typed-helper fix from #3368 was not available — hence the shim.

This is reachable today, not hypothetically: `filterValueSchema` parses a `between` range through `z.coerce.date()`, so the advanced filter builder's date-range picker produces real `Date`s. Confirmed by rendering the predicate — `"devices"."last_seen_at" BETWEEN $1 AND $2` with **two `Date` objects** as params.

Reached from `POST /devices/filters/preview`, dynamic device groups, deployment target resolution, and the AI fleet tool.

## Site 2 — `oauth/revocationRetry.ts` (silent, and it defeats the retry)

The `onConflictDoUpdate` branch interpolated `attemptedAt` and `input.expiresAt` bare. That branch fires **only when a retry row already exists** — i.e. exactly when a prior Redis revocation write failed. `jobs/oauthRevocationRetryWorker.ts` re-invokes it every 30s; the throw escaped the `withSystemDbAccessContext` transaction (rolling back the whole batch) and was swallowed by the generic `.catch` in `runScheduledDrain`, which logs and returns 0.

**Net effect: the OAuth revocation retry queue could never drain or heal during a Redis outage, and failed quietly rather than per-row.**

These two sit in arithmetic and `GREATEST` contexts, so they use `sqlTimestamptz` — the casts are load-bearing: `$1 + interval '1 second'` has no unique operator resolution, and `GREATEST($1, col)` would resolve off the other argument. The `set` object is extracted as `buildRetryConflictUpdate` so the SQL can be rendered and asserted without a database.

## Site 3 — `services/alertCorrelationGroups.ts` (found by the sweep)

`upsertGroup` interpolated `firstSeenAt`/`lastSeenAt` bare into its `INSERT … ON CONFLICT` template. Both come straight off `alerts.triggeredAt`, which Drizzle returns as a live `Date`, so **every** group write threw and failed the whole correlation-grouping pass. `first_seen_at`/`last_seen_at` are naive `timestamp` columns, so this uses the uncast `sqlTimestamp`.

## Also fixed — numeric `between` was silently coerced to epoch dates

Found while verifying site 1, on the same endpoint. `z.coerce.date()` accepts a plain number as epoch-milliseconds, and `dateRangeValueSchema` sat ahead of any numeric branch in the union — so `{ from: 10, to: 90 }` (an ordinary `metrics.cpuPercent between 10 and 90`) parsed as the date range `1970-01-01T00:00:00.010Z .. .090Z` and was then compared against a numeric column.

Fixing only the Date binding would have converted one 500 into another (a Postgres type error instead of a driver `TypeError`), so a `numberRangeValueSchema` now precedes the date range in the union, and `FilterValue` gains `{ from: number; to: number }`. The web UI hides numeric `between` (see the comment in `apps/web/src/components/filters/filterFields.ts`), which is why this went unnoticed — but the REST API, dynamic groups and the AI fleet tool all accept it.

## Sweep

A repo-wide pass over every `sql\`\`` template in `apps/api/src` (plus `apps/portal/src` and `packages/*/src`) traced each non-trivial interpolated identifier to its declaration. It found **one** new instance (site 3 above) and confirmed the rest safe: every retention/reaper job, `invoiceService`, `agents/patches`, `patches/helpers`, `auditChainAnchor`, `deploymentEngine`, `maintenanceService`, `mobile`, `devices/warranty`, `metricRollups`, `metricAnomalies`, `userRiskSignals` already `.toISOString()` before interpolating — several with comments citing this exact bug class. `partnerApi` cursor/traversal values are typed `string`, not `Date`. `apps/portal/src` imports no drizzle at all.

No lint rule is proposed: the defect depends on a value's *runtime* type, which a static rule cannot see. The per-site rendered-SQL guards below are the enforceable form.

## Tests

Unit (no database; **each guard was verified to fail against the pre-fix behaviour** — 20 failures when the helpers are stubbed back to a bare pass-through):

- `db/sqlValues.test.ts` — new. Includes an explicit test that a *bare* interpolation really does bind a `Date`, so the file documents the behaviour it guards against; cast/no-cast assertions per helper; millisecond fidelity; parameters stay bound rather than inlined.
- `services/filterEngine.test.ts` — every datetime operator renders through `PgDialect.sqlToQuery` with no `Date` param; `between` keeps both bounds distinct; **no `timestamptz` appears** (the naive-column guard); related-table/computed paths covered; ordinary scalars unchanged; numeric `between` stays numeric.
- `oauth/revocationRetry.test.ts` — `buildRetryConflictUpdate` binds no `Date`, casts both timestamps, keeps the `LEAST(300, POWER(2, …))` backoff and the `GREATEST` lifetime rule, and matches what the conflict branch actually installs.
- `services/alertCorrelationGroups.test.ts` — no `Date` survives in any upsert's bound values.
- `packages/shared/.../filters_operators.test.ts` — numeric range stays numeric, date range still coerces, relative-time value not mistaken for a range.

Integration against real Postgres — `__tests__/integration/sqlTemplateDateBinding.integration.test.ts`, new. No mock reproduces the original failure (it lives in the driver's type coercion), and no rendered-SQL assertion can prove the statement is *semantically* right. This covers `between`/`before`/`after`/`equals` returning exactly the right devices, the same assertions under a non-UTC session time zone, and the revocation-retry conflict branch actually executing its backoff arithmetic (8s from attempts=3; capped at 300s from attempts=20) and never shortening an existing deadline.

`tsc --noEmit` clean across `apps/api`, `apps/web` and `packages/shared`.

Reported by @bdunncompany (sweep credit from #3329 / #3368).

### DIFF
--- a/apps/api/src/__tests__/integration/sqlTemplateDateBinding.integration.test.ts
+++ b/apps/api/src/__tests__/integration/sqlTemplateDateBinding.integration.test.ts
@@ -31,14 +31,14 @@
  */
 import './setup';
 
-import { and, eq } from 'drizzle-orm';
+import { and, eq, sql } from 'drizzle-orm';
 import { describe, expect, it } from 'vitest';
 
 import { getTestDb } from './setup';
 import { createOrganization, createPartner, createSite, createUser } from './db-utils';
 import { withSystemDbAccessContext } from '../../db';
 import { devices, oauthRevocationRetries } from '../../db/schema';
-import { evaluateFilter } from '../../services/filterEngine';
+import { buildConditionSQL, evaluateFilter } from '../../services/filterEngine';
 import { buildRetryConflictUpdate } from '../../oauth/revocationRetry';
 
 // Deliberately spread across a day so a session-time-zone shift of a few hours
@@ -143,41 +143,53 @@ describe('filterEngine datetime predicates — real Postgres (#3369)', () => {
 
   it('gives identical results under a non-UTC session time zone', async () => {
     // `devices.last_seen_at` is `timestamp` WITHOUT time zone, holding UTC wall
-    // clock by Drizzle convention. Had the parameter been cast `::timestamptz`,
-    // Postgres would coerce the naive column *in the session zone* — under
-    // Pacific/Kiritimati (UTC+14) the T2 device (22:00Z) would fall outside the
-    // window below and this assertion would drop it. A silent, deployment-
-    // dependent wrong answer is worse than the TypeError being fixed, so it
-    // gets its own guard.
+    // clock by Drizzle convention. Had the bound parameter been cast
+    // `::timestamptz`, Postgres would coerce the naive column *in the session
+    // zone* — under Pacific/Kiritimati (UTC+14) the T2 device (22:00Z) shifts
+    // to the 16th local and falls outside the window below. A silent,
+    // deployment-dependent wrong answer is worse than the TypeError being
+    // fixed, so the no-cast decision gets its own guard.
+    //
+    // The predicate is run directly rather than through `evaluateFilter`, and
+    // the time zone is set with `SET LOCAL` inside the same transaction as the
+    // query. `getTestDb()` and the application `db` pool (`src/db/index.ts`)
+    // are SEPARATE postgres.js pools on separate connections, so a plain `SET`
+    // on one would not reach a query issued on the other and this guard would
+    // pass vacuously. What is under test here is the SQL `buildConditionSQL`
+    // emits, which is exactly what this asserts.
     const { org, seeded } = await seedOrgWithDevices();
 
-    const run = () =>
-      evaluateFilter(
-        {
-          operator: 'AND',
-          conditions: [
-            {
-              field: 'lastSeenAt',
-              operator: 'between',
-              value: {
-                from: new Date('2026-03-15T06:00:00.000Z'),
-                to: new Date('2026-03-16T00:00:00.000Z'),
-              },
-            },
-          ],
-        },
-        { orgId: org.id },
-      );
+    const predicate = buildConditionSQL({
+      field: 'lastSeenAt',
+      operator: 'between',
+      value: {
+        from: new Date('2026-03-15T06:00:00.000Z'),
+        to: new Date('2026-03-16T00:00:00.000Z'),
+      },
+    });
 
-    const db = getTestDb();
-    const previous = (await db.execute(`SHOW TimeZone`)) as unknown as Array<{ TimeZone: string }>;
-    try {
-      await db.execute(`SET TimeZone = 'Pacific/Kiritimati'`);
-      const shifted = await withSystemDbAccessContext(run);
-      expect([...shifted.deviceIds].sort()).toEqual(idsFor(seeded, [T1, T2]));
-    } finally {
-      await db.execute(`SET TimeZone = '${previous[0]?.TimeZone ?? 'UTC'}'`);
-    }
+    const rowsIn = async (timeZone: string) =>
+      getTestDb().transaction(async (tx) => {
+        // SET LOCAL is transaction-scoped, so it cannot leak into another test.
+        await tx.execute(sql`SET LOCAL TimeZone = ${sql.raw(`'${timeZone}'`)}`);
+        const found = await tx
+          .select({ id: devices.id })
+          .from(devices)
+          .where(and(eq(devices.orgId, org.id), predicate));
+        return found.map((r) => r.id).sort();
+      });
+
+    // Sanity: the zone really is being applied inside the transaction. Without
+    // this, a silently ignored SET would make the comparison below meaningless.
+    const applied = await getTestDb().transaction(async (tx) => {
+      await tx.execute(sql`SET LOCAL TimeZone = 'Pacific/Kiritimati'`);
+      const shown = (await tx.execute(sql`SHOW TimeZone`)) as unknown as Array<{ TimeZone: string }>;
+      return shown[0]?.TimeZone;
+    });
+    expect(applied).toBe('Pacific/Kiritimati');
+
+    expect(await rowsIn('UTC')).toEqual(idsFor(seeded, [T1, T2]));
+    expect(await rowsIn('Pacific/Kiritimati')).toEqual(idsFor(seeded, [T1, T2]));
   });
 });
 
--- a/apps/api/src/__tests__/integration/sqlTemplateDateBinding.integration.test.ts
+++ b/apps/api/src/__tests__/integration/sqlTemplateDateBinding.integration.test.ts
@@ -31,7 +31,7 @@
  */
 import './setup';
 
-import { and, eq, sql } from 'drizzle-orm';
+import { and, eq, isNull, sql } from 'drizzle-orm';
 import { describe, expect, it } from 'vitest';
 
 import { getTestDb } from './setup';
@@ -250,6 +250,11 @@ describe('oauth revocation retry conflict branch — real Postgres (#3369)', () 
       })
       .onConflictDoUpdate({
         target: [oauthRevocationRetries.markerType, oauthRevocationRetries.markerId],
+        // The unique index is PARTIAL (`WHERE completed_at IS NULL`), so the
+        // predicate is required for Postgres to infer the arbiter index —
+        // without it the statement fails 42P10 before the SET list is ever
+        // evaluated. Mirrors `writeOAuthRevocationMarkerDurably`.
+        targetWhere: isNull(oauthRevocationRetries.completedAt),
         set: buildRetryConflictUpdate({
           attemptedAt,
           expiresAt: new Date('2026-03-15T14:00:00.000Z'),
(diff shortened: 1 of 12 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (correctness: per-column cast choice, missed interpolation sites, zod union reorder, the `buildRetryConflictUpdate` extraction, repo contracts) and `pr-test-analyzer` (vacuous-assertion audit, coverage gaps, integration-test validity).

**Findings:** 4 raised → all addressed in `305bc265` / `caa88028`; 0 outstanding.

- **`code-reviewer`: 0 findings.** It independently re-derived the column types from `db/schema/` (`devices.last_seen_at`/`enrolled_at`/`quarantined_at` and `alert_correlation_groups.first_seen_at`/`last_seen_at` naive `timestamp`; `oauth_revocation_retries.*` `timestamptz`) and confirmed each site picked the right helper, and checked `sqlTimestamp` against the installed `drizzle-orm@0.45.2` source — it is byte-for-byte `PgTimestamp.mapToDriverValue`. Also confirmed the union reorder shadows no working input, and that the `attempts` backoff still reads the pre-update column value.
- **`pr-test-analyzer`, blocker — the non-UTC timezone guard was vacuous.** It confirmed what I suspected when writing it: `getTestDb()` and the application `db` pool are separate postgres.js pools on separate connections, so `SET TimeZone` on the former never reached the query issued on the latter. The test passed identically with or without the `::timestamptz` regression it exists to catch — worse than no guard, since that failure mode is silent and deployment-dependent. It now runs the emitted predicate inside a single transaction with `SET LOCAL TimeZone`, and asserts the zone actually took effect before comparing, so it cannot go quietly inert again.
- **`pr-test-analyzer`, minor ×3 —** covered the remaining `sqlValue` call sites (`matches`, `in`/`notIn`, group-membership `equals`); pinned Drizzle's `undefined` behaviour (it *drops* the interpolation rather than binding it, leaving malformed SQL — not reachable and not changed here, recorded so a future caller widening the type finds it); dropped one assertion made redundant by the `toEqual` above it.

**Guards are non-vacuous:** stubbing the three helpers back to a bare pass-through fails **20** of the new assertions. The pre-fix behaviour is also asserted directly in `sqlValues.test.ts`, so the file documents what it guards against.

**Tests:**
- Unit — `apps/api` affected suites green single-fork: `sqlValues`, `filterEngine`, `src/oauth` (14 files), `alertCorrelationGroups` → **17 files, 248 passed**. `packages/shared` validators → **39 files, 1167 passed**.
- Integration — `sqlTemplateDateBinding.integration.test.ts` against real Postgres: **7 passed**. En route it caught a genuine defect in my own test (missing `targetWhere` for the *partial* unique index → 42P10); the failure output confirmed production was already correct, binding `$9::timestamptz`/`$10::timestamptz` as strings (postgres.js type 1184) with no `Date` — it reached the planner, so Bind had already succeeded.
- `tsc --noEmit` clean across `apps/api`, `apps/web`, `packages/shared`.

**Note for the merge decision:** this PR is based on `main`, *not* on #3368 — the two are independent and either can merge first. Worth dispatching CI explicitly (`gh workflow run CI --ref fix/3369-date-in-sql-templates`) if `gh pr checks` looks short: the two suites that would catch a regression here (`Integration Tests`) need a live database and cannot fail in the unit job.

PR #3383, opened 2026-08-10
fix(discovery): rank the automatic asset classifiers so a UniFi switch stops flapping to access_point (#3187)

### LINKED ISSUE #3187: [Discovery] Agent scan can rewrite a UniFi switch to access_point — vendor inference disagrees with the UniFi classifier, flapping detected_asset_type
## Description

Found while fixing #3011 (PR #3185), left out of scope there.

`inferAssetTypeFromVendor` maps `ubiquiti` → `access_point`, while the UniFi sync stamps `manufacturer: 'Ubiquiti'` on **every** device it touches — switches, gateways and cameras included. So an agent scan that runs after a UniFi sync can rewrite a correctly-classified UniFi **switch** to `access_point`.

Because the two paths disagree, `detected_asset_type` can flap back and forth between runs depending on which one wrote last.

Related: `discoveryWorker.processResults` still stamps `'unknown'` as if it were a classification. PR #3185 stopped the UniFi path from doing that (an unclassified device now writes neither type column, since `'unknown'` from that classifier means "no opinion"), but the agent-scan path was not changed.

## Notes

- #3011 / PR #3185 fixed the *manual* override being clobbered. This issue is about the two **automatic** classifiers disagreeing with each other, which is a separate defect.
- Vendor-string inference is too coarse to distinguish Ubiquiti product lines; the fix likely needs the UniFi device model, not the manufacturer string.

ToddHebebrand:
Closes #3187

## Root cause

`discovered_assets.type_source` (`'manual' | 'auto'`) records only whether a **human** pinned the type. Both *automatic* classifiers write `'auto'`, so neither could tell it was about to clobber the other:

- `unifiSyncService.reconcileDiscoveredAsset` classifies from the UniFi controller's device record (`usw` → `switch`) and stamps `manufacturer: 'Ubiquiti'` on **every** row it touches — switches, gateways and cameras included.
- `discoveryWorker.processResults`, finding no classification of its own, guessed from that same vendor string via `inferAssetTypeFromVendor`, which mapped `'ubiquiti'` → `access_point`.

So a correctly-classified UniFi **switch** became an `access_point` on the next agent scan and a `switch` again on the next sync, flapping `detected_asset_type` indefinitely. The two paths were feeding each other: the sync wrote the very manufacturer string the scan then mis-read.

## The fix — a second provenance axis

`detected_type_source` ranks the automatic classifiers against each other:

| rank | source | evidence |
|---|---|---|
| 10 | `vendor_oui` | MAC OUI / SNMP vendor string. Narrows the product *category* at best. |
| 20 | `agent_scan` | The agent's own on-the-wire classification (ports, SNMP sysDescr, OS fingerprint). |
| 30 | `unifi_controller` | The controller's device record — it knows the exact model. |

A classifier may write only when its rank is `>=` the stored one (`>=`, not `>`, so a source can always refresh its *own* previous answer).

**Why a separate column instead of adding `'integration'` to `type_source`.** Collapsing the two axes would mean a manual override erased the provenance of the automatic result underneath it — `type_source` would read `'manual'` and the fact that UniFi was the automatic winner would be gone. `resetTypeToAuto` would then demote a controller-derived type to a vendor guess. Keeping them orthogonal means they compose: a human beats every classifier, and among classifiers the better-informed one wins. `resetTypeToAuto` needed no change at all, and every existing `typeSource === 'manual'` check in the web UI stays valid.

The rank map is `Record<DiscoveredAssetDetectionSource, number>`, so **adding a classifier without ranking it is a compile error** — the next person to add an SNMP or telemetry classifier has to decide where it sits.

### Guards are SQL, not JS

Same reasoning as the manual guard in #3011, now extended: on the `ON CONFLICT` path the colliding row is never read, and on the `UPDATE` path *both* source columns can change between the `SELECT` and the `UPDATE`. A rejected write re-selects the existing column rather than NULLing it.

## Also in scope

- **Removed `'ubiquiti'` from `VENDOR_ROLE_KEYWORDS`.** That table's own stated contract is "single-purpose vendors where the OUI strongly implies a specific device category". Ubiquiti ships switches (UniFi Switch), gateways (UDM/USG), cameras (Protect) and APs under one OUI, so the entry mislabelled three product lines out of four. It violated the invariant the table was built on.
- **The agent-scan path no longer stamps `'unknown'` as a classification.** `'unknown'` means "this scan has no opinion", not "the device is of unknown type" — writing it back erased a better answer. #3185 fixed exactly this on the UniFi path and explicitly left the agent path out of scope; this is that half. A no-opinion scan now writes none of the three type columns.
- **`device_role` propagation is skipped when the scan's classifier was outranked**, otherwise the guess the asset row just rejected would leak onto the linked device and reintroduce the flap one table over. (This one decision is made from the JS pre-read — `devices` is a different table and can't share the asset's UPDATE; a lost race there costs a stale role until the next scan, not a user-visible flap. Commented as such.)
- **Tightened `inferAssetTypeFromVendor` / `mapAssetType`** to the asset-type union — they returned loose `string` / `any`, which is what allowed the drift in the first place. This immediately caught a real type hole during development.

## Tenancy

No new table. `discovered_assets` is RLS shape 1 (direct `org_id`, auto-discovered) and is already registered in `CORE_ORG_CASCADE_DELETE_ORDER`, `CORE_DEVICE_CASCADE_DELETE_TABLES` and `CORE_TENANT_EXPORT_POLICY`.

The new **column** does trigger the export-policy contract (the one registration rule that fires on a column, not just a table) — `detected_type_source` is classified `included`: it names a classifier, not tenant data.

The migration backfills `detected_type_source = 'unifi_controller'` for rows already linked to a `unifi_devices` row, so the first agent scan after deploy doesn't get one free overwrite of every existing UniFi classification. Rows left NULL are correct as NULL — provenance is genuinely unknown, so the next classifier of any strength may claim them (`else 0` in the rank CASE). Row count is reported per the migration convention.

## Verification

| Check | Result |
|---|---|
| `discoveryWorker.test.ts`, `unifiSyncService.test.ts`, `discoveredAssetClassification.test.ts`, `macVendorLookup.test.ts`, `discovery.test.ts`, `discoveryQueue.test.ts` | **114 passed** |
| `discoveredAssetDetectionSource.integration.test.ts` (new, real Postgres) | **8 passed** |
| `unifiAssetTypeSource.integration.test.ts` (existing #3011 behaviour) | **5 passed** |
| `tenant-export-policy` + `tenantCascade` integration | **7 passed** |
| `rls-coverage` | **75 passed** |
| `pnpm db:check-drift` | no drift, 509 migrations match |
| `tsc --noEmit` | clean |

Two new test files were added where none existed: `macVendorLookup.test.ts` (the module had **zero** coverage) and `discoveredAssetClassification.test.ts`.

The **integration** test is the load-bearing one — a mocked test can only prove we *built* the statement we meant to. It pins what Postgres actually does: that the untyped bind parameters in the rank CASE unify with the enum column, that the `::discovered_asset_detection_source` cast resolves, that a target-table-qualified reference inside `UPDATE ... SET` reads the *pre-update* row (the entire mechanism of the guard), and that `else 0` really does rank NULL below every source.

**Mutation-checked, not just green:** neutering the rank guard to `true` fails the headline regression test and the concurrent-writers test, confirming they are not vacuous.


ToddHebebrand:

**Findings:** 6 raised → all 6 addressed in `6460c1d0e`; 0 outstanding. No Critical findings against the core precedence logic — the SQL guards, `excluded.*` conflict form, and cascade/export-policy registration were all confirmed correct as originally written.

| # | Agent | Finding | Resolution |
|---|---|---|---|
| 1 | silent-failure-hunter | **The `AssetWriteSet` type check wasn't checking anything.** Both writers merged the guarded columns via `Object.assign`, whose signature (`(t: T, s: U) => T & U`) places no constraint on `U`'s keys — so the documented compile-time protection did not exist, in either file. Verified empirically by the agent against the repo's own drizzle types. | Assign the three columns individually (`applyClassification` in the UniFi service); `discoveryWorker`'s set is now typed too, not `Record<string, unknown>`. **Mutation-verified**: renaming a key to `detectedTypeSorce` is now `TS2551`, where it previously compiled clean. |
| 2 | comment-analyzer | **`device_role` could keep a value the asset row rejected.** The propagation decided from the pre-read row; with worker `concurrency: 5`, two scans of the same host could both act on one snapshot and race on an unguarded update. Worse, my comment claimed it self-heals "until the next scan" — it doesn't: once the asset carried `agent_scan`, later `vendor_oui`-only scans were blocked by the rank check and could never correct the device. | `device_role` now reads `discovered_assets.asset_type` back out in a subquery, with both manual carve-outs in the `WHERE`. The device converges on whatever the asset settled on, so it can't be stranded — and the separate `device_role_source` SELECT is gone. |
| 3 | code-reviewer | **Migration sorted before six already-shipped ones.** `main` carries `2026-08-11`…`2026-08-19`, so a `2026-08-10` prefix didn't sort last. No dependency conflict, but it breaks the ordering convention. | Renamed to `2026-08-20-…`, swept the `readFileSync` path. The test-DB ledger guard caught the stale entry, which confirmed the sweep was complete. |
| 4 | silent-failure-hunter | **Enum drift was only checked one way.** The `Record<>` catches a TS member added without a rank, but a migration adding a pg label with no TS entry was silent — such rows fall to `else 0`, ranked below everything, so the weakest classifier could overwrite them. That is #3187 reintroduced through drift. | Added a `pg_enum` assertion that the DB labels equal the TS union exactly. |
| 5 | pr-test-analyzer | Migration backfill had zero coverage — an inverted `IS NULL` or wrong join column would ship silently. | Real-DB test that executes the **shipped** `DO $$` block (read from disk, not retyped) against four seeded permutations: linked+classified, linked+unclassified, unlinked, already-sourced. |
| 6 | pr-test-analyzer | No positive-path test for `device_role`; the PR had changed that write's source variable with only negative tests around it. | Positive unit test plus a real-Postgres test covering the manual-device, manual-asset and `'unknown'` cases. |

**On finding 2** — that one is a behaviour change beyond the literal issue, flagging it explicitly for review. It replaces a JS decision + a SELECT with one guarded statement. code-reviewer had rated the original an acceptable documented tradeoff; I fixed it rather than softening the comment, because leaving a knowingly-divergent write on a field users see in device lists seemed worse than the small extra diff.

**Tests:** 115 unit (`discoveryWorker`, `unifiSyncService`, `discoveredAssetClassification`, `macVendorLookup`, `discovery` routes, `discoveryQueue`) · 11 in the new real-DB suite · 12 contract/UniFi integration · 75 RLS coverage — all green. `tsc --noEmit` clean. `pnpm db:check-drift`: no drift.

**Not vacuous, checked:** neutering the rank guard to `true` fails the headline regression and concurrent-writer tests; forcing `classificationApplied = false` fails the device-role test; the column-typo mutation above fails `tsc`.

**CI:** all 43 checks green, including both Check Migrations jobs and all four Integration Tests shards.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3187 — both are yours to call.

PR #3384, opened 2026-08-10
fix(agent): resolve macOS Finder aliases in the file browser (#3344)

### LINKED ISSUE #3344: File Browser doesn't resolve macOS Finder aliases — lists/downloads them as plain files
Reported by **semotech** in #bugs (Discord, message 1536176627564675164).

**Report (PII-redacted):**

Bug: macOS Sequoia — RD rename/Delete broken + File Browser ignores aliases
Device: iMac-2019 OS: macOS Sequoia 15.7.9 (darwin 15.7.9) Agent: 0.104.0 Self-host: API/web 0.104.0

1) Remote Desktop — rename fails; Delete keys ignored Client: Breeze Viewer

Steps:
1. Start RD (Viewer) → in Finder try to rename a file/folder.
2. Also try forward-delete (⌦ / DEL) and full Delete.
Expected: Rename/Delete work like a local session. Actual: Rename fails; ⌦ DEL and Delete have no effect.
Workaround: Apple Screen Sharing works. Notes: Likely keyboard injection/keycode mapping on Sequoia Helper (not Finder ACL — Screen Sharing OK). Delete failing too suggests broader key handling. Checked RD notes (#3194, #3049, #3165) — nothing relevant.

2) File Browser — aliases not followed (download instead) 

Steps:
1. Via Screen Sharing, put an Alias of an iCloud folder on the Desktop.
2. In Breeze File Browser, open Desktop and use that alias.
Expected: Resolve alias and navigate to target (iCloud path). 
Actual: Treated as a file → download attempted. Impact: Desktop→iCloud aliases unusable; fall back to Screen Sharing.
Impact: Common macOS pattern (Desktop shortcuts into iCloud / ~/Library/Mobile Documents/…) is unusable from File Browser; technicians forced to fall back to Apple Screen Sharing (if available).

## Triage analysis

This report has two parts. Part 2 (File Browser + Finder aliases) is a confirmed code gap: agent/internal/remote/tools/fileops.go's ListFiles determines each entry's Type purely from entry.IsDir() (lines 247-250) and there is no macOS-alias-resolution logic anywhere in the codebase (searched for isAlias/FSResolveAlias/bookmark-data patterns, found nothing). A classic Finder alias is a regular file on disk (not a directory, not a POSIX symlink), so it's classified Type:'file' and the client offers a download instead of navigating to the aliased iCloud target — exactly matching the report. Symlinks get some special handling elsewhere (EvalSymlinks in enforceReadContainment, skip-on-zip), but plain alias files get none, and ListFiles never resolves them either. Part 1 (RD rename/Delete on Sequoia) I could not confirm as a code defect: agent/internal/remote/desktop/input_darwin.go maps 'delete'->0x75 (kVK_ForwardDelete) and 'backspace'->0x33 (kVK_Delete), which are the correct Carbon keycodes, and apps/viewer/src/lib/keymap.ts maps DOM event.code (physical key position) consistently for both keys regardless of OS. No Sequoia-specific branch or recent regression showed up in the code or git history search. This may be an OS-level permission/entitlement change in Sequoia (e.g. stricter synthetic-event handling) rather than an app bug, but I have no code evidence either way.

## Affected code

- `agent/internal/remote/tools/fileops.go:194-268` — ListFiles builds each FileEntry.Type from entry.IsDir() only — no alias detection/resolution for macOS Finder alias files, which are regular files on disk
- `agent/internal/remote/tools/fileops.go:169-186` — enforceReadContainment resolves POSIX symlinks via filepath.EvalSymlinks but has no equivalent for macOS alias records
- `agent/internal/remote/desktop/input_darwin.go:316-360` — macOS keycode table — 'delete'->0x75 and 'backspace'->0x33 are the correct Carbon keycodes; checked for a Sequoia-specific bug and found none

## Potential fix

In fileops.go's ListFiles (and the file-open/download handler), detect macOS alias files (e.g. via a Finder-info/resource-fork check or the Cocoa/CoreServices alias APIs) and resolve them to their target path before returning Type/Path, similar to how symlinks are already special-cased.

ToddHebebrand:
Addresses **part 2** of #3344 (File Browser doesn't follow macOS Finder aliases).

> **Not `Closes #3344` on purpose.** That issue bundles two reports. Part 1 (Remote Desktop rename/Delete on Sequoia) is untouched here and triage found no code defect for it, so the issue should stay open after this merges.

## The bug

A Finder alias is **not** a symlink. On disk it is an ordinary regular file whose data fork holds a CoreFoundation bookmark blob and whose `com.apple.FinderInfo` xattr carries the `kIsAlias` flag. So `os.DirEntry.IsDir()` is false and `filepath.EvalSymlinks` is a no-op on one.

`ListFiles` derived `Type` purely from `entry.IsDir()`, which meant every alias was classified `"file"`. The reporter's case — a Desktop alias into iCloud Drive, an extremely common macOS pattern — listed as a ~1KB file and the UI offered the **bookmark blob itself** for download instead of navigating to the target. Technicians had to fall back to Apple Screen Sharing.

## The fix

**`agent/internal/remote/tools/macalias.go`** — the bookmark parser, deliberately built on *every* platform so its tests run in CI's linux job. It walks the blob's table of contents and rebuilds the target path from the path-component array (key `0x1004`) and the volume path (key `0x2002`).

**`agent/internal/remote/tools/macalias_darwin.go`** — detection and resolution. Gates on the `kIsAlias` FinderInfo flag via a single `getxattr` *before* opening anything, so listing a large directory costs one cheap failing syscall per entry rather than an open+read. Alias→alias chains are followed with both a hop limit and a visited-path set.

### Why pure Go and not CoreServices

`NSURL`/`CFURLCreateByResolvingBookmarkData` would have been the supported API, but the agent's CI runs `CGO_ENABLED=0 go test ./...` and several release targets build with cgo off — a cgo resolver would have silently done nothing on exactly the builds that ship. Parser correctness was verified against real alias files generated on macOS Sequoia 15.x (both a folder alias and a file alias) before the synthetic test fixtures were written.

## Behaviour

| Field | For a resolved alias |
|---|---|
| `type` | The **target's** kind — so existing clients navigate into a folder alias and download through a file alias with no alias-specific handling |
| `path` | The **alias file**, unchanged — rename/delete/move must act on the alias, never on what it points at |
| `size` / `modified` / `permissions` | The **alias file**. A delete confirmation quoting the target's size while removing a 1KB alias would be actively misleading |
| `isAlias` / `aliasTarget` | New optional fields marking the entry for clients |

`ListFiles` and `ReadFile` also resolve when the *requested path* is itself an alias, and the API now echoes the path the agent actually listed, so breadcrumbs, "go up" and the upload destination follow the alias to its target rather than pointing at the alias file. Without that echo the listing worked but the client would have tried to upload into a path whose parent is a regular file.

No new UI was built — the fix is deliberately shaped so the existing `type === 'directory'` / `type === 'file'` branches do the right thing. `isAlias`/`aliasTarget` are there for a future badge without needing another agent change.

## Security

Resolving an alias **deliberately leaves the directory being listed**, and `EvalSymlinks` cannot see through one, so this would have been a way around the SR5-01 credential-store deny-list if left alone. Every resolved target is re-checked with `enforceReadContainment`:

- In a listing, an alias into a credential store is left **unresolved** — its target path is never disclosed and it stays indistinguishable from a plain file.
- On read or navigate, it is **denied outright**.

The parser treats its input as hostile: every offset, length and count that comes off disk is bounds-checked in `uint64` (no 32-bit wrap), the TOC chain is cycle-guarded, oversized files are never opened, the blob is read through a `LimitReader` so a stat/open race cannot grow it, and path components containing separators or `..` are rejected so a crafted blob cannot steer resolution somewhere else.

### One pre-existing observation, not changed here

`isSensitiveReadPath` matches keychains via `/library/keychains/` **with a trailing separator**, so the `Library/Keychains` *directory itself* is not deny-listed and can be enumerated today on `main`. Contents are still blocked (the `.keychain-db` basename rule catches them). I left that alone rather than widen a security deny-list inside an unrelated bug fix — flagging it in case it's worth its own issue.

## Verification

- `go test -race -count=1 ./internal/remote/tools/...` — **green** (66s, full package)
- Parser verified against **real macOS Sequoia alias files** (folder + file), both resolving to the correct target with the correct `IsDir`
- Cross-platform `go vet` + `go build ./...` with `CGO_ENABLED=0` for `windows/amd64`, `windows/arm64`, `linux/amd64`, `linux/arm64`, `darwin/amd64`, `darwin/arm64` — all clean, darwin-only code correctly build-tagged
- `go test -c` for `linux/amd64` — confirms the parser tests compile and run off-darwin
- `tsc --noEmit` clean for both `apps/api` and `apps/web`
- `apps/api` systemTools suite: 40 passed; `apps/web` FileManager + no-silent-mutations: 104 passed

### Test coverage added

Parser (all platforms): boot-volume and external-volume paths, the reporter's iCloud Drive path, unicode/spaces, and rejection of non-bookmarks, truncated blobs, hostile headers, oversized/cyclic TOCs, and traversal components.

Darwin end-to-end, driving `ListFiles`/`ReadFile` against alias files built byte-for-byte the way Finder writes them: folder alias resolves to `directory`, file alias downloads the target's bytes under the target's name, navigating into a folder alias lists the target and reports the resolved path, alias chains resolve, cycles terminate, and — the ones that matter most — an alias into a credential store is neither resolved nor readable, while plain files, POSIX symlinks, dangling aliases, unflagged bookmark blobs, flagged junk, and oversized files are all left exactly as they were.

## Reviewer notes

- Nothing outside macOS changes behaviour: `macalias_other.go` returns "not an alias" unconditionally.
- An alias that cannot be resolved (old-format, dangling, unparseable) degrades to the previous behaviour — a plain file — rather than erroring.
- @SemoTech offered to retest; this needs an agent build on a Sequoia box to confirm in the field.

### DIFF
--- a/agent/internal/remote/tools/fileops.go
+++ b/agent/internal/remote/tools/fileops.go
@@ -212,6 +212,18 @@ func ListFiles(payload map[string]any) CommandResult {
 		return NewErrorResult(err, time.Since(start).Milliseconds())
 	}
 
+	// A macOS Finder alias to a folder is a regular file, so navigating into one
+	// would otherwise fail with "not a directory". Resolve it and list the
+	// target, re-running containment against the target — alias resolution
+	// deliberately leaves the listed directory, so the deny-list has to be
+	// applied again on the far side (issue #3344).
+	if target, ok := resolveMacAliasPath(cleanPath); ok {
+		if err := enforceReadContainment(target); err != nil {
+			return NewErrorResult(err, time.Since(start).Milliseconds())
+		}
+		cleanPath = target
+	}
+
 	limit := GetPayloadInt(payload, "limit", defaultFileListLimit)
 	if limit < 1 {
 		limit = 1
@@ -249,13 +261,38 @@ func ListFiles(payload map[string]any) CommandResult {
 			entryType = "directory"
 		}
 
+		entryPath := filepath.Join(cleanPath, entry.Name())
+
+		// macOS Finder aliases are regular files holding a bookmark blob, so
+		// without this they list as ordinary ~1KB files and the UI offers the
+		// blob itself for download (issue #3344). Only Type is taken from the
+		// target, because Type is what drives the client's navigate-vs-download
+		// affordance. Path, Size, Modified and Permissions keep describing the
+		// alias file itself: those are what the mutating operations act on, and
+		// a delete confirmation that quoted the target's size while removing
+		// only a 1KB alias would be actively misleading.
+		aliasTarget := ""
+		if target, targetInfo, ok := resolveMacAliasTarget(entryPath, info); ok {
+			// Never surface an alias that leads into a credential store: leaving
+			// it unresolved keeps both the target path and its contents hidden.
+			if err := enforceReadContainment(target); err == nil {
+				aliasTarget = target
+				entryType = "file"
+				if targetInfo.IsDir() {
+					entryType = "directory"
+				}
+			}
+		}
+
 		fileEntries = append(fileEntries, FileEntry{
 			Name:        entry.Name(),
-			Path:        filepath.Join(cleanPath, entry.Name()),
+			Path:        entryPath,
 			Type:        entryType,
 			Size:        info.Size(),
 			Modified:    info.ModTime().Format(time.RFC3339),
 			Permissions: info.Mode().String(),
+			IsAlias:     aliasTarget != "",
+			AliasTarget: aliasTarget,
 		})
 	}
 
@@ -286,6 +323,17 @@ func ReadFile(payload map[string]any) CommandResult {
 	// Defense-in-depth: never read a credential store, even symlinked (SR5-01).
 	if err := enforceReadContainment(cleanPath); err != nil {
 		return NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+
+	// Downloading a macOS Finder alias should deliver the file it points at, not
+	// the bookmark blob (issue #3344). Containment is re-checked against the
+	// target because an alias can point anywhere, including at a credential
+	// store that filepath.EvalSymlinks would never have surfaced.
+	if target, ok := resolveMacAliasPath(cleanPath); ok {
+		if err := enforceReadContainment(target); err != nil {
+			return NewErrorResult(err, time.Since(start).Milliseconds())
+		}
+		cleanPath = target
 	}
 
 	// Check file info first
--- a/agent/internal/remote/tools/macalias.go
+++ b/agent/internal/remote/tools/macalias.go
@@ -1,0 +1,263 @@
+package tools
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// macOS Finder aliases are not symlinks: on disk they are ordinary regular
+// files whose data fork holds a CoreFoundation "bookmark" blob (the same bytes
+// CFURLCreateBookmarkData / -[NSURL writeBookmarkData:toURL:] produce) and
+// whose com.apple.FinderInfo xattr carries the kIsAlias flag. Because they are
+// regular files, os.DirEntry.IsDir() is false and filepath.EvalSymlinks is a
+// no-op on them, so without explicit handling the file browser lists an alias
+// as a ~1KB "file" and offers the raw bookmark blob for download instead of
+// following it (issue #3344).
+//
+// The blob is parsed here in pure Go rather than through CoreServices/cgo: the
+// agent's CI and several release targets build with CGO_ENABLED=0, so a cgo
+// resolver would silently do nothing on exactly the builds that ship.
+//
+// Blob layout (all integers little-endian):
+//
+//	0x00  "book\0\0\0\0mark\0\0\0\0"  magic
+//	0x10  uint32  header size (56 in every macOS version observed)
+//	0x14  uint32  header size, repeated
+//	0x18  uint32  body size
+//	0x1C  uint32  version
+//	...          reserved
+//
+//	body := blob[headerSize:]
+//	body[0:4]  uint32  offset (relative to body) of the first table of contents
+//
+// Table of contents record:
+//
+//	uint32 size, uint32 magic (0xFFFFFFFE), uint32 identifier,
+//	uint32 offset of next TOC (0 = last), uint32 entry count,
+//	then entryCount × { uint32 key, uint32 offset, uint32 reserved }
+//
+// Item at body[offset]:
+//
+//	uint32 length, uint32 type, then length payload bytes
+//
+// The two keys needed to rebuild a POSIX path are the path-component array and
+// the volume path; everything else (CNIDs, volume UUID, creation dates) is
+// ignored.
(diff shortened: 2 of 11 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer (correctness + security) and silent-failure-hunter, both scoped to the bookmark parser, the containment threading, and the build tags.

**Findings:** 2 raised → both addressed in `23d483c09`; 0 outstanding.

- *silent-failure-hunter (medium)* — past the `kIsAlias` gate, Finder's own bit has confirmed the file **is** an alias, yet open/read/parse/stat failures were all swallowed identically, degrading the entry to "ordinary file" with no signal that resolution had even been attempted. Each now logs a `[WARN]`, keeping the parser's error text — the only forensic trail when a blob is corrupt or is probing the path-traversal guard. The flag gate is itself the anti-spam mechanism (almost nothing carries that xattr), and the one routine failure — a dangling alias whose target was deleted — stays silent by design.
- *silent-failure-hunter* explicitly **judged correct, do not re-litigate:* the per-entry containment denial staying silent (surfacing it would leak the existence and location of a credential store to anyone listing an unrelated directory), the hard error on explicit navigate/read denial, the hop-limit returning the last good hop rather than erroring, and the `Type`-vs-`Path`/`Size` field split.
- *code-reviewer* — **0 findings.** Confirmed every offset/length/count from disk is bounds-checked in `uint64`; the two remaining plain-`uint32` additions are provably non-wrapping because a prior `sliceAt` already bounded the operand against a body capped at 64KB by the `LimitReader`. Also confirmed containment is re-applied at all three sites that reach a resolved target, that no consumer breaks on the field split, and that the TS `json.path` read is backward-compatible with an older agent (guarded fallback to the requested path). It additionally ran a **2M-iteration mutation fuzz** of `parseBookmarkTargetPath` — zero panics.

**Separately fixed:** `Lint Agent (Go)` went red on the first commit — `maxAliasFileSize` lives in the untagged parser file but is only read on darwin, so `unused` flagged it under the job's `GOOS=linux`. Reproduced locally with the exact CI invocation (`golangci-lint run --new-from-rev=origin/main`) and moved the constant into the darwin file; now 0 issues on both linux and darwin.

**Tests (local, foreground):**
- `go test -race -count=1 ./internal/remote/tools/...` — green
- `go vet` + `go build` with `CGO_ENABLED=0` across `windows/{amd64,arm64}`, `linux/{amd64,arm64}`, `darwin/{amd64,arm64}` — clean
- `go test -c` for `linux/amd64` — parser tests compile and run off-darwin, so they are covered by the linux CI job rather than being darwin-only
- Parser verified against **real macOS Sequoia alias files** (folder + file), both resolving to the correct target with the correct `IsDir`
- `tsc --noEmit` clean for `apps/api` and `apps/web`; API systemTools 40 passed, web FileManager + no-silent-mutations 104 passed

**Status:** review-clean. Full CI was still in flight when this was written — no failures at that point, and the one job that had gone red is fixed and reproduced-green locally. Worth a glance at the check list before merging.

**Not a `Closes`:** #3344 bundles two reports and this is part 2 only. Part 1 (Remote Desktop rename/Delete on Sequoia) is untouched, so the issue should stay open after merge. Field verification needs an agent build on a Sequoia box — @SemoTech offered to retest.

PR #3388, opened 2026-08-10
fix(config): make the env↔compose parity guard see commented-out vars (#3239)

### LINKED ISSUE #3239: [API] env↔compose parity guard skips commented-out vars — ~39 documented knobs are silently inert
## Description

`apps/api/src/config/envComposeParity.test.ts` exists to stop a documented variable being silently inert — its own docstring cites the self-hoster whose "CORS/2FA/platform-admin settings all no-op'd because they were never threaded through Compose".

It cannot see commented-out variables, which is the exact class most likely to be forgotten.

The extractor is:

```js
function activeEnvExampleVars(relPath: string): string[] {
  ...
  const m = /^([A-Z][A-Z0-9_]*)=/.exec(line); // uncommented assignments only
```

But the idiomatic way this repo documents an **optional tuning knob** is commented out, showing its default:

```
# EVENT_LOOP_MONITOR_INTERVAL_MS=1000
# DB_POOL_HEALTH_INTERVAL_MS=60000
# MCP_REQUIRE_EXECUTE_ADMIN=true
```

So every optional knob is invisible to the guard. `DB_POOL_HEALTH_*` is commented out too and only got wired because #3224 happened to do it by hand; #3236 is the same shape where nobody did. This is not a one-off — it is a hole the size of "every optional variable".

## Blast radius

Sweeping both `.env.example` ↔ compose pairs for **commented-only** variables that reach no container, using the guard's own `isReferencedInCompose()`:

- root pair (`.env.example` ↔ `docker-compose.yml`): **34**
- droplet pair (`deploy/.env.example` ↔ `deploy/docker-compose.prod.yml`): **5**

Not all are bugs — some are legitimately host-level or web build-time. Cross-referencing each against `apps/` source:

**Consumed by the API but unmapped — setting these in `.env` does nothing:**

`AGENT_AUTO_PROMOTE`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, `ANTHROPIC_MODEL`, `AUTH_COOKIE_FORCE_SECURE`, `AUTH_COOKIE_SAME_SITE`, `AUTO_MIGRATE`, `BILLING_SERVICE_API_KEY`, `BILLING_SERVICE_URL`, `BREEZE_BILLING_API_KEY`, `BREEZE_BILLING_URL`, `COOKIE_FORCE_SECURE`, `COOKIE_SAME_SITE`, `CORS_INCLUDE_DEFAULT_ORIGINS`, `ENABLE_API_DOCS_UI`, `EVENT_LOOP_MONITOR_*` (the #3236 three), `MCP_EXECUTE_TOOL_ALLOWLIST`, `MCP_MAX_SSE_SESSIONS_PER_KEY`, `MCP_MESSAGE_RATE_LIMIT_PER_MINUTE`, `MCP_REQUIRE_EXECUTE_ADMIN`, `MCP_SSE_RATE_LIMIT_PER_MINUTE`, `PORTAL_COOKIE_FORCE_SECURE`, `PORTAL_COOKIE_SAME_SITE`, `PORTAL_STATE_BACKEND`, `PUBLIC_ACTIVATION_BASE_URL`, `REFRESH_ROTATION_GRACE_SECONDS`

Droplet pair: `AGENT_AUTO_PROMOTE`, `BREEZE_ALLOW_UNAUTH_REDIS`, `LOGIN_ACCOUNT_LOCKOUT_MAX`, `LOGIN_ACCOUNT_LOCKOUT_WINDOW_SECONDS`, `MFA_FORCE_FOR_PARTNER_ADMIN`

**Apparently allow-list material** (no `apps/` consumer, or consumed elsewhere): `BUSINESS_EMAIL_ALLOW_OVERRIDES`, `C2C_M365_CERT_PRIVATE_KEY_PATH`, `C2C_M365_CERT_THUMBPRINT` (no reference anywhere under `apps/`); `COMPOSE_PROFILES` (host-level, docs only); `USE_AGENT_SDK` (docs only); `PUBLIC_DOCS_URL` (web build-time, `apps/web/astro.config.mjs`).

Several of the unmapped ones are security-relevant and match the guard's founding story almost exactly: `AUTH_COOKIE_SAME_SITE`, `COOKIE_FORCE_SECURE`, `CORS_INCLUDE_DEFAULT_ORIGINS`, `MFA_FORCE_FOR_PARTNER_ADMIN`, `LOGIN_ACCOUNT_LOCKOUT_*`, `MCP_REQUIRE_EXECUTE_ADMIN`. An operator hardening a self-host deploy by setting those gets no effect and no warning.

**Method / caveat:** two passes — `process.env.X` and a bare-name search across `apps/`. The first undercounts (`EVENT_LOOP_MONITOR_INTERVAL_MS` is read as a string literal through `readPositiveIntEnv`, so it scores zero on that pass despite being live); the second can match a doc page or comment. Neither is authoritative on its own, so the buckets above are a starting point for triage, not a verdict on each var.

## Proposed fix

Extend `activeEnvExampleVars()` to also collect `^#\s*([A-Z][A-Z0-9_]*)=` and run the same three assertions over the union.

The mechanism is a two-line change. The work is the **map-or-allowlist decision on each of the ~39 vars it surfaces**, which is why this is filed rather than folded into the #3236 PR — flipping the regex without that triage turns the required `test-api` job red immediately.

Two questions that are yours to settle:

1. **Do commented-out knobs deserve the same guarantee as uncommented ones?** I'd argue yes — the whole failure mode is "documented but inert", and commenting a line out documents it just as loudly. But it does mean every future optional knob must be wired or allow-listed at the moment it is documented.
2. **Map or allow-list the security-relevant ones?** They look like genuine gaps rather than deliberate exclusions, but that is your call per var.

Happy to take the implementation plus the full triage in a follow-up PR if you want it — say the word and which way you'd like the ambiguous ones to go.

## Affected files

- `apps/api/src/config/envComposeParity.test.ts` (primary)
- `docker-compose.yml`, `deploy/docker-compose.prod.yml` (whatever the triage decides to map)

## Provenance

Found while fixing #3236 — the question "why didn't the parity guard catch this?" turned out to have a general answer.

ToddHebebrand:
Closes #3239. Reported by @bdunncompany, who also found the root cause while fixing #3236 — the question "why didn't the parity guard catch this?" had a general answer.

## The hole

`apps/api/src/config/envComposeParity.test.ts` exists to stop a documented variable being silently inert. Its extractor matched only uncommented assignments:

```js
const m = /^([A-Z][A-Z0-9_]*)=/.exec(line); // uncommented assignments only
```

But commenting a line out showing its default is precisely how this repo documents an **optional tuning knob**:

```
# MCP_REQUIRE_EXECUTE_ADMIN=true
# LOGIN_ACCOUNT_LOCKOUT_MAX=10
```

So the entire optional-knob surface was exempt from the guard meant to cover it — and it is the class most likely to be forgotten, because an unmapped optional var still boots fine and just silently ignores the operator. #3236 was one instance of this shape; #3224 got wired only because someone did it by hand.

## The change

Extractor is now `^(?:#\s*)?([A-Z][A-Z0-9_]*)=` and the union runs through the same three assertions. That surfaced **31 root-pair** and **5 droplet-pair** variables reaching no container.

### Mapped into the `api` service (26 root)

Settling @bdunncompany's question 2 — these were genuine gaps, not deliberate exclusions, so they are mapped rather than allow-listed.

| Group | Vars |
|---|---|
| Auth/CORS hardening | `COOKIE_SAME_SITE`, `COOKIE_FORCE_SECURE`, `AUTH_COOKIE_SAME_SITE`, `AUTH_COOKIE_FORCE_SECURE`, `PORTAL_COOKIE_SAME_SITE`, `PORTAL_COOKIE_FORCE_SECURE`, `CORS_INCLUDE_DEFAULT_ORIGINS`, `REFRESH_ROTATION_GRACE_SECONDS` |
| Portal session store | `PORTAL_STATE_BACKEND` |
| MCP limits + execute gating | `MCP_REQUIRE_EXECUTE_ADMIN`, `MCP_EXECUTE_TOOL_ALLOWLIST`, `MCP_SSE_RATE_LIMIT_PER_MINUTE`, `MCP_MESSAGE_RATE_LIMIT_PER_MINUTE`, `MCP_MAX_SSE_SESSIONS_PER_KEY` |
| AI provider overrides | `ANTHROPIC_MODEL`, `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN` |
| External billing service | `BREEZE_BILLING_URL`, `BREEZE_BILLING_API_KEY`, `BILLING_SERVICE_URL`, `BILLING_SERVICE_API_KEY` |
| Startup + agent fleet | `AUTO_MIGRATE`, `AGENT_AUTO_PROMOTE`, `ENABLE_API_DOCS_UI`, `PUBLIC_ACTIVATION_BASE_URL` |

### Mapped into the droplet `api` service (5)

`MFA_FORCE_FOR_PARTNER_ADMIN`, `LOGIN_ACCOUNT_LOCKOUT_MAX`, `LOGIN_ACCOUNT_LOCKOUT_WINDOW_SECONDS`, `AGENT_AUTO_PROMOTE`, `BREEZE_ALLOW_UNAUTH_REDIS`.

`PROD_ALLOWLIST` stays empty.

### Allow-listed with a reason (5 root)

| Var | Reason |
|---|---|
| `COMPOSE_PROFILES` | read by the `docker compose` CLI itself — host-level, never a container env |
| `PUBLIC_DOCS_URL` | web build-time `PUBLIC_` var (`apps/web/astro.config.mjs`), baked into the prebuilt image |
| `C2C_M365_CERT_THUMBPRINT` | no consumer — `.env.example` marks it "Future: certificate-based auth" |
| `C2C_M365_CERT_PRIVATE_KEY_PATH` | same |
| `USE_AGENT_SDK` | no consumer anywhere in the repo (documented in `deploy/environment.mdx` only) |
| `BUSINESS_EMAIL_ALLOW_OVERRIDES` | no consumer — the shipped gate reads `SIGNUP_REQUIRE_BUSINESS_EMAIL` / `SIGNUP_BUSINESS_EMAIL_CONTACT_URL` |

The last four are **dead documentation**, not intentional exclusions. I allow-listed rather than deleted them from `.env.example` so the decision stays yours — see "Open question" below.

## The bug the sweep would otherwise have shipped

This is the part that made the sweep non-mechanical, and it is worth a close look.

`${VAR:-}` renders an unset variable as `VAR: ""`, so the container sees the variable **set to an empty string**, not absent. Four cookie readers resolved a prefixed override against a generic fallback with `??`:

```ts
normalizeSameSite(process.env.AUTH_COOKIE_SAME_SITE ?? process.env.COOKIE_SAME_SITE)
```

`''` is not nullish, so once both names are compose-mapped the prefixed name **permanently shadows** the generic one. An operator setting only `COOKIE_SAME_SITE=None` would silently get `Lax` back — and, via `shouldSetSecureCookie`, lose the implied `Secure`. Setting only `COOKIE_FORCE_SECURE=true` (the documented remedy behind a TLS-terminating proxy the app cannot see) would be ignored outright, issuing non-`Secure` refresh and CSRF cookies.

That is the guard's own founding failure mode, reintroduced by fixing it.

Fixed in the readers rather than by leaving the vars unmapped, since "the prefixed name wins **when configured**" is what the two-name pair always meant. `envStr` (`apps/api/src/utils/envStr.ts`) already exists for exactly this and treats empty/whitespace-only as absent:

```ts
normalizeSameSite(envStr('AUTH_COOKIE_SAME_SITE', envStr('COOKIE_SAME_SITE', '')))
```

Applied to all four sites in `routes/auth/helpers.ts` and `routes/portal/helpers.ts`. No other behavior change — the accepted value sets (`'1'`/`'true'`, `strict`/`none`) are untouched.

The other 25 mapped vars were each traced to their read sites and confirmed to treat `''` identically to unset (`envFlag`/`envInt`/`parseCsvSet`/`positiveIntFromEnv` falsiness checks, or explicit `?? ''`/`.trim()` normalization), including every `config/validate.ts` entry — so no validator starts firing on an empty value at boot.

## Tests

- `parseDocumentedVars` split out as a pure function with 5 cases: commented forms (`# X=`, `#X=`, `#   X=`), de-duplication against the live form, prose comments correctly ignored (`# Values: strict, lax, none`, `# lowercase_name=1`), indented non-comments ignored, and a regression case pinning that a file of only optional knobs is no longer parsed as empty.
- 5 new cookie-precedence tests (auth + portal) covering empty, whitespace-only, configured-override-still-wins, and both-empty-falls-to-default.

**These 5 fail against the pre-fix helpers and pass after** — verified by reverting just the two source files and re-running:

```
FAIL  an EMPTY AUTH_COOKIE_SAME_SITE does not shadow COOKIE_SAME_SITE
FAIL  an EMPTY AUTH_COOKIE_FORCE_SECURE does not shadow COOKIE_FORCE_SECURE
FAIL  a whitespace-only override is also treated as unconfigured
FAIL  an EMPTY PORTAL_COOKIE_SAME_SITE does not shadow COOKIE_SAME_SITE
FAIL  an EMPTY PORTAL_COOKIE_FORCE_SECURE does not shadow COOKIE_FORCE_SECURE
```

Full run: `apps/api/src/config` + the two helper suites + compose-adjacent guards — **331 passed**. `tsc --noEmit` clean. The six compose/guided-setup/supply-chain shell guards pass. No duplicate keys introduced in either `api` environment block.

## Answering @bdunncompany's question 1

> Do commented-out knobs deserve the same guarantee as uncommented ones?

Yes — and the tradeoff they flagged is real and worth stating plainly: **every future optional knob must now be wired or allow-listed at the moment it is documented.** That is the point. The guard fails in the required `test-api` job with the var name and the two ways to resolve it, which is a far cheaper place to learn this than a self-hoster's production deploy.

The four no-consumer vars (`USE_AGENT_SDK`, `BUSINESS_EMAIL_ALLOW_OVERRIDES`, and the two `C2C_M365_CERT_*`) are documented knobs that no code reads. Allow-listing keeps them visible; **deleting the lines from `.env.example`** (and the `USE_AGENT_SDK` row from `deploy/environment.mdx`) would arguably be more honest, since a reader today is being promised a knob that does not exist. I left them in place because removing documented lines is your call, not a mechanical sweep's. Happy to follow up either way.

### DIFF
--- a/apps/api/src/config/envComposeParity.test.ts
+++ b/apps/api/src/config/envComposeParity.test.ts
@@ -32,8 +32,10 @@ const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..
  * Mapping a var in costs nothing when the operator leaves it unset: the
  * `${VAR:-}` form passes an empty string, which every consumer below treats as
  * "unset" and falls back to its own default. Where a consumer did NOT treat ''
- * that way, the consumer was fixed rather than the var left unmapped — see the
- * cookie helpers' firstConfiguredEnv() (#3239).
+ * that way, the consumer was fixed rather than the var left unmapped — see
+ * resolveAuthCookieSameSite() in routes/auth/helpers.ts, which resolves its
+ * override chain with nested envStr() instead of `??` for exactly this reason
+ * (#3239).
  */
 
 // Variables intentionally NOT threaded into the self-host (root) stack's
@@ -130,6 +132,15 @@ const PAIRS: Pair[] = [
  * Only `# NAME=` is collected, not prose. A comment has to look like an
  * assignment to count, so ordinary explanatory text above a var is not mistaken
  * for a documented knob.
+ *
+ * Prose that OPENS with `# SOME_VAR=value` — e.g. the sentence at
+ * `.env.example` "# OAUTH_DCR_ALLOW_ANONYMOUS=true. The API refuses to boot…" —
+ * does match, and that is the deliberate bias. Such a line is indistinguishable
+ * from a real documented default without parsing English, and this guard exists
+ * to fail LOUD: a spurious hit costs one allow-list line, while the miss it
+ * would otherwise permit is a knob that silently ignores the operator forever.
+ * (That particular line is a no-op anyway — the var is genuinely documented a
+ * few lines below and the Set de-duplicates.)
  */
 export function parseDocumentedVars(text: string): string[] {
   const names = new Set<string>();
--- a/apps/api/src/config/envComposeParity.test.ts
+++ b/apps/api/src/config/envComposeParity.test.ts
@@ -20,13 +20,20 @@ const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..
  * and — the reason this test exists — a self-hoster whose CORS/2FA/platform-admin
  * settings all no-op'd because they were never threaded through Compose).
  *
- * The guard, per env-example ↔ compose pair: every active variable in the
- * `.env.example` MUST be either
+ * The guard, per env-example ↔ compose pair: every variable DOCUMENTED in the
+ * `.env.example` — live assignments AND commented-out defaults alike, see
+ * documentedEnvExampleVars() — MUST be either
  *   (a) referenced in the compose file (mapped into a container, sourced into a
  *       secret, or used for interpolation), or
  *   (b) listed in that pair's allow-list below with a reason.
  * Adding a documented var without doing one of those two fails CI here, in the
  * required test-api job, instead of silently in someone's production deploy.
+ *
+ * Mapping a var in costs nothing when the operator leaves it unset: the
+ * `${VAR:-}` form passes an empty string, which every consumer below treats as
+ * "unset" and falls back to its own default. Where a consumer did NOT treat ''
+ * that way, the consumer was fixed rather than the var left unmapped — see the
+ * cookie helpers' firstConfiguredEnv() (#3239).
  */
 
 // Variables intentionally NOT threaded into the self-host (root) stack's
@@ -48,6 +55,7 @@ const ROOT_ALLOWLIST: Record<string, string> = {
   BREEZE_PROXY_TARGET_HOST: 'guided-setup external-proxy bookkeeping',
   BREEZE_EXTERNAL_PROXY: 'guided-setup external-proxy bookkeeping',
   BREEZE_EXTERNAL_PROXY_CIDRS: 'guided-setup copies this into TRUSTED_PROXY_CIDRS (which IS mapped)',
+  COMPOSE_PROFILES: 'read by the `docker compose` CLI itself to select profiles — host-level, never a container env',
 
   // REDIS_URL is not consumed by the API container (it derives its connection
   // from REDIS_HOST/REDIS_PORT + the file-backed redis_password secret).
@@ -61,11 +69,23 @@ const ROOT_ALLOWLIST: Record<string, string> = {
   // misleading, not functional.
   PUBLIC_RELEASE_VERSION: 'web build-time (baked into the prebuilt web image)',
   PUBLIC_TICKET_MAILBOX_APP_ID: 'web build-time PUBLIC_ var (baked into the prebuilt web image)',
+  PUBLIC_DOCS_URL: 'web build-time PUBLIC_ var (apps/web/astro.config.mjs — baked into the prebuilt web image)',
   ENABLE_SENTRY_SMOKE: 'web build/SSR smoke flag (baked into the prebuilt web image)',
   SENTRY_DSN_WEB_SERVER: 'web SSR Sentry DSN (baked into the prebuilt web image)',
   SENTRY_AUTH_TOKEN: 'build-time source-map upload (CI only, never a runtime container env)',
   SENTRY_ORG: 'build-time source-map upload (CI only)',
   SENTRY_PROJECT: 'build-time source-map upload (CI only)',
+
+  // Documented but read by NO code in this repo, so there is nothing to thread
+  // them into — mapping them would manufacture a knob that still does nothing.
+  // Listed here (rather than deleted from .env.example) so the dead
+  // documentation stays visible and the decision to keep or drop each line
+  // stays the maintainer's; see #3239.
+  C2C_M365_CERT_THUMBPRINT: 'no consumer — .env.example marks it "Future: certificate-based auth"',
+  C2C_M365_CERT_PRIVATE_KEY_PATH: 'no consumer — .env.example marks it "Future: certificate-based auth"',
+  USE_AGENT_SDK: 'no consumer anywhere in the repo (documented in deploy/environment.mdx only)',
+  BUSINESS_EMAIL_ALLOW_OVERRIDES:
+    'no consumer — the shipped business-email gate reads SIGNUP_REQUIRE_BUSINESS_EMAIL / SIGNUP_BUSINESS_EMAIL_CONTACT_URL instead',
 };
 
 // The digest-pinned droplet stack. Its api block is well-maintained; after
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter, both scoped to the six focus areas this change actually risks (regex false positives, empty-string safety of every newly-mapped var, api-vs-web service placement, the four cookie fixes, allow-list accuracy, YAML duplicate keys).

**Findings:** 1 raised → addressed in `55d8348`; 0 outstanding.

- *Stale comment* — the guard docstring cited a `firstConfiguredEnv()` helper that was never written (leftover from an earlier draft; the fix uses nested `envStr()` directly). Repointed at `resolveAuthCookieSameSite()`.
- Same commit records a second reviewer observation as a deliberate design note: prose that OPENS with an assignment (`# OAUTH_DCR_ALLOW_ANONYMOUS=true. The API refuses to boot…`) does match the extractor. That line is a no-op — the var is genuinely documented below and the `Set` de-duplicates — but the general case is intentional. Such a line can't be told from a real documented default without parsing English, and this guard should fail loud: a spurious hit costs one allow-list line, while the miss it would otherwise permit is a knob that silently ignores the operator forever.

**Independently confirmed clean by both reviewers:**

- **Empty-string safety** — all 31 newly-mapped vars traced to their read sites (shared `envStr`/`envInt`/`envFlag`, the local copies in `mcpServer.ts`, `positiveIntFromEnv` in `rate-limit.ts`, and every `config/validate.ts` entry). No validator fires on `''`, so no self-host deploy is boot-blocked by a mapped-but-unset knob. This was the highest-severity thing available to find.
- **Service placement** — verified by YAML block-boundary parsing, not presence-anywhere: every new key lands in the `api:` block of its file, and every var has a real `apps/api/src` consumer.
- **No fifth `??` chain** among the newly-mapped vars. (A pre-existing `PUBLIC_API_URL ?? API_URL` pattern exists elsewhere but can't fire — both compose files give `PUBLIC_API_URL` a non-empty default. Untouched here, noted for the record.)
- **Allow-list reasons** — all six verified by repo-wide grep. The four "no consumer" claims hold: zero read sites anywhere in `apps/`, `agent/`, `packages/`.
- **Guard not vacuously satisfiable** — every newly-mapped var matches via the direct `VAR: ${VAR:-}` service-key branch, not an incidental interpolation elsewhere.
- **No duplicate keys** in either `api` environment block.

**Tests:** `apps/api/src/config` + both cookie helper suites green locally — **398 passed**, `tsc --noEmit` exit 0, and the six compose/guided-setup/supply-chain shell guards pass. The five new cookie-precedence tests were verified to **fail** against the pre-fix helpers (reverted the two source files, re-ran, confirmed red) and pass after — so they pin the bug rather than merely describing it.

**CI:** 54/54 green, 0 failures — including `Test API`, `Test Web`, `Smoke Test`, `Lint`, and all four `Integration Tests` shards.

**One open question for you, not a blocker** (also in the PR body): `USE_AGENT_SDK`, `BUSINESS_EMAIL_ALLOW_OVERRIDES`, `C2C_M365_CERT_THUMBPRINT`, and `C2C_M365_CERT_PRIVATE_KEY_PATH` are documented knobs that **no code in the repo reads**. I allow-listed them with honest reasons rather than deleting the lines, because removing documented lines from `.env.example` (and the `USE_AGENT_SDK` row from `deploy/environment.mdx`) is your call, not a mechanical sweep's. Deleting is arguably more honest — a reader today is promised a knob that doesn't exist. Happy to follow up either way.

PR #3391, opened 2026-08-10
feat(remote): let a technician pick their own remote-access provider (#3389)

### LINKED ISSUE #3389: [Remote] Per-technician remote-tool preference — defaultProviderId is tenant-wide, so techs cannot differ
## The gap

`InheritableRemoteAccessSettings` already carries a **list** of providers, but only one of them can ever be launched:

```ts
interface InheritableRemoteAccessSettings {
  providers?: RemoteAccessProvider[];
  defaultProviderId?: string;   // one winner, tenant-wide
}
```

`resolveRemoteAccessLaunch` (`services/remoteAccessLauncher.ts`) resolves exactly `providers.find(p => p.id === defaultProviderId)`, and that setting inherits down the tenant tree. So in a shop where one technician prefers RustDesk and another prefers a MeshCentral-style tool, whoever sets `defaultProviderId` decides for everyone. Both providers can be configured and `enabled` simultaneously today — only the *selection* is singular.

This is a two-technician MSP hitting it, but it generalises: different techs have different tooling installed on their own workstations, and the deep link only works if the scheme it emits matches what that tech actually has.

## Why it's a small change

Two of the three pieces already exist:

- `providers` is already an array of `{ id, name, urlTemplate, customFieldKey, password?, enabled }`.
- `users.preferences` is already a `jsonb` column, already read and written through `PATCH /users`, already used for flags like `bootstrapSetupRequired`.

So the missing piece is only: persist a chosen `providerId` per user, and prefer it during launch resolution with fallback to `defaultProviderId` when unset or no longer valid.

## Proposed shape

1. **Preference** — `users.preferences.remoteAccessProviderId?: string`. No schema migration needed.
2. **Resolution** — in `resolveRemoteAccessLaunch`, prefer the user's id **only if it names a provider in the tenant's own list and that provider is `enabled`**; otherwise fall back to `defaultProviderId`. A stale preference (provider deleted or disabled) silently degrades to the default rather than failing the launch.
3. **UI** — the Connect control offers a chooser only when more than one provider is enabled; a single-provider tenant sees no change at all.

## The security constraint this must respect

The preference must carry **an id and nothing else** — never a `urlTemplate`, scheme, or password. `remoteAccessLauncher` already treats a partner template that resolves to `javascript:` as a security event, and the whole point of the current design is that the template is tenant-controlled. A per-user field that could introduce or override a template would reopen exactly that hole, so the user's value has to be validated as a member of the tenant's enabled set before it is honoured.

Worth stating explicitly because it is the one way this feature could go wrong quietly.

## Two calls that are yours, not mine

- **Where the chooser lives** — inline on the Connect control, or a "preferred remote tool" row in user settings, or both.
- **Whether it is remembered per device** — a tech might want RustDesk for servers and something else for workstations. Per-user-only is simpler and covers the reported need; per-user-per-device is more flexible and more state.

Happy to implement whichever you prefer. I have a per-user-only version building against the constraint above and will open it as a draft so the resolution and validation logic can be reviewed independently of the UI decision — if you would rather own the whole thing, say so and I will close it.

bdunncompany:
Closes #3389 (backend only — see "What is deliberately not here").

`defaultProviderId` is tenant-wide, so two technicians who prefer different remote tools cannot both be served. Both providers can already be configured and `enabled` simultaneously; only the *selection* was singular.

## The change

`resolveRemoteAccessLaunch` takes an optional `preferredProviderId` and prefers it over the tenant default.

It is **an ID and nothing else**. The preference SELECTS from the tenant's own provider list and never supplies one — `urlTemplate`, `password` and `customFieldKey` still come from the tenant record. So a user-controlled value cannot introduce a scheme or a destination, which is what keeps the existing post-substitution `javascript:` guard meaningful. That property is the whole design; everything else follows from it.

A preference counts only when it names a provider **this tenant has** and that provider is **enabled**. Unknown id, id belonging to another tenant, provider since disabled or deleted — all fall through to `defaultProviderId` rather than failing, so a stale preference degrades quietly instead of stranding the technician mid-call.

## The non-obvious part: it is gated on `isInteractiveUserSession`

`AuthContext.principal`'s own comment is what drove this:

> `user.id` cannot answer "is a human doing this?". An MCP API key is built with `user.id = apiKey.createdBy` … so a key and the human who minted it are indistinguishable by identity alone

Keying the preference off user identity alone would make a machine caller silently inherit whichever remote tool the human who minted the key happens to prefer. A preference is a property of a person at a keyboard, so non-interactive principals get the tenant default.

## Storage

`users.preferences.remoteAccessProviderId` — an existing `jsonb` column already served by `PATCH /users`. **No migration.**

## What is deliberately not here

No UI. The two open questions on #3389 are yours: where the chooser lives (inline on Connect, a user-settings row, or both), and whether the choice is remembered per device. The resolution and validation logic is reviewable independently of that, which is why this is a draft — say the word and I will add the UI to match whichever you pick, or close this if you would rather own it.

## Verification

`tsc --noEmit` 0 errors. API suite **21246 passed / 1315 files**. Trivy (secret, misconfig, HIGH+CRITICAL) clean on `apps/api/src`.

**Control run, stated precisely:** with the resolution change reverted, **3 of the 6** new cases fail — preference selects the other provider, preference stands in with no tenant default, and the `javascript:`-after-substitution refusal under a preference. The other **3 cannot fail**, because they assert fallback behaviour that is intentionally identical to today's (unknown id → default, disabled preferred → default, no preference → default). They are regression pins, not evidence.

One existing suite needed a fixture update: `core.remoteAccessLaunch.test.ts` mocks `../../middleware/auth`, so it needed `isInteractiveUserSession` added. I mirrored the real allowlist-of-one and gave the mocked context a real `principal` rather than stubbing the gate to a constant, so that suite still exercises the gate it now depends on.

### DIFF
--- a/apps/api/src/routes/devices/core.remoteAccessLaunch.test.ts
+++ b/apps/api/src/routes/devices/core.remoteAccessLaunch.test.ts
@@ -38,6 +38,10 @@ vi.mock('../../middleware/auth', () => ({
       orgId: 'org-123',
       partnerId: null,
       accessibleOrgIds: ['org-123'],
+      // This route is reached by a person clicking Connect, so the mocked
+      // context carries the same discriminator the real one would. The
+      // per-technician provider preference is gated on it.
+      principal: { kind: 'user_session' },
       canAccessOrg: (orgId: string) => orgId === 'org-123'
     });
     return next();
@@ -45,6 +49,9 @@ vi.mock('../../middleware/auth', () => ({
   requireScope: vi.fn(() => async (_c: any, next: any) => next()),
   requirePermission: vi.fn(() => async (_c: any, next: any) => next()),
   requireMfa: vi.fn(() => async (_c: any, next: any) => next()),
+  // Mirrors the real allowlist-of-one rather than stubbing a constant, so this
+  // suite still exercises the gate it depends on.
+  isInteractiveUserSession: vi.fn((a: any) => a?.principal?.kind === 'user_session'),
 }));
 
 vi.mock('./helpers', () => ({
--- a/apps/api/src/routes/devices/core.ts
+++ b/apps/api/src/routes/devices/core.ts
@@ -18,8 +18,16 @@ import {
   enrollmentKeys,
   organizations,
   partners,
+  users,
 } from '../../db/schema';
-import { authMiddleware, requireMfa, requireScope, requirePermission } from '../../middleware/auth';
+import {
+  authMiddleware,
+  isInteractiveUserSession,
+  requireMfa,
+  requireScope,
+  requirePermission,
+  type AuthContext,
+} from '../../middleware/auth';
 import { PERMISSIONS, canAccessSite, type UserPermissions } from '../../services/permissions';
 import {
   getPagination,
@@ -993,6 +1001,7 @@ coreRoutes.get(
       const launcher = await resolveRemoteAccessLauncherForDevice(
         device.orgId,
         device.customFields as Record<string, unknown> | null,
+        auth,
       );
       if (launcher.launchUrl) {
         hasRemoteAccessLauncher = true;
@@ -1031,9 +1040,34 @@ coreRoutes.get(
  * engine doesn't filter the row out. This mirrors how remoteAccessPolicy.ts
  * uses systemAuth for the same reason.
  */
+/**
+ * Read the acting technician's preferred provider id, if any.
+ *
+ * Gated on `isInteractiveUserSession` on purpose. An MCP API key is built with
+ * `user.id = apiKey.createdBy` (see AuthContext.principal), so keying this off
+ * user identity alone would make a machine caller silently inherit whichever
+ * remote tool the human who minted the key happens to prefer. A preference is a
+ * property of a person at a keyboard; anything else gets the tenant default.
+ */
+async function readPreferredProviderId(auth: AuthContext): Promise<string | null> {
+  if (!isInteractiveUserSession(auth)) return null;
+  const prefs = await withSystemDbAccessContext(async () => {
+    const [row] = await db
+      .select({ preferences: users.preferences })
+      .from(users)
+      .where(eq(users.id, auth.user.id))
+      .limit(1);
+    return row?.preferences ?? null;
+  });
+  if (!prefs || typeof prefs !== 'object' || Array.isArray(prefs)) return null;
+  const value = (prefs as { remoteAccessProviderId?: unknown }).remoteAccessProviderId;
+  return typeof value === 'string' && value.length > 0 ? value : null;
+}
+
 async function resolveRemoteAccessLauncherForDevice(
   orgId: string,
   customFields: Record<string, unknown> | null,
+  auth?: AuthContext,
 ): Promise<RemoteAccessLaunchResult> {
   const partnerSettings = await withSystemDbAccessContext(async () => {
     const [partnerRow] = await db
@@ -1046,7 +1080,8 @@ async function resolveRemoteAccessLauncherForDevice(
   });
   const providers: InheritableRemoteAccessSettings | undefined =
     partnerSettings.remoteAccessProviders;
-  return resolveRemoteAccessLaunch({ customFields }, providers);
+  const preferredProviderId = auth ? await readPreferredProviderId(auth) : null;
+  return resolveRemoteAccessLaunch({ customFields }, providers, preferredProviderId);
 }
 
 // POST /devices/:id/remote-access-launch - Issue a one-shot remote-access
@@ -1082,6 +1117,7 @@ coreRoutes.post(
       launcher = await resolveRemoteAccessLauncherForDevice(
         device.orgId,
         device.customFields as Record<string, unknown> | null,
+        auth,
       );
     } catch (err) {
       captureException(err, c);
--- a/apps/api/src/routes/devices/core.ts
+++ b/apps/api/src/routes/devices/core.ts
@@ -1051,14 +1051,19 @@ coreRoutes.get(
  */
 async function readPreferredProviderId(auth: AuthContext): Promise<string | null> {
   if (!isInteractiveUserSession(auth)) return null;
-  const prefs = await withSystemDbAccessContext(async () => {
-    const [row] = await db
-      .select({ preferences: users.preferences })
-      .from(users)
-      .where(eq(users.id, auth.user.id))
-      .limit(1);
-    return row?.preferences ?? null;
-  });
+  // Deliberately the ambient request context, not a system one. `users` carries
+  // `breeze_user_isolation_select ... OR id = breeze_current_user_id()`, so a
+  // technician reading their own row is already permitted under the caller's
+  // own scope. Reading under the request context is the narrower privilege and
+  // means this can never resolve a row outside the caller's tenant, even if
+  // `auth.user.id` were ever influenced by something other than the verified
+  // token.
+  const [row] = await db
+    .select({ preferences: users.preferences })
+    .from(users)
+    .where(eq(users.id, auth.user.id))
+    .limit(1);
+  const prefs = row?.preferences ?? null;
   if (!prefs || typeof prefs !== 'object' || Array.isArray(prefs)) return null;
   const value = (prefs as { remoteAccessProviderId?: unknown }).remoteAccessProviderId;
   return typeof value === 'string' && value.length > 0 ? value : null;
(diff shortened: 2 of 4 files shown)


ToddHebebrand:
Reviewing the backend on its own as you asked — the two UI questions are genuinely mine and I'll answer them separately so you aren't blocked on the code review.

**The core design is right, and the reasoning you led with is the reason.** A preference that carries an id and nothing else, validated as a member of the tenant's own enabled set before it is honoured, is exactly the shape that keeps the post-substitution `javascript:` guard meaningful. Falling through to `defaultProviderId` on unknown/disabled/foreign ids rather than failing is also the right call — a stale preference should degrade quietly, not strand a tech mid-call.

**Gating on `isInteractiveUserSession` rather than `auth.user.id` is a good catch** and not one I'd have expected to be made up front. A machine caller inheriting the preferences of whoever minted the key is a subtle enough failure that it would have shipped.

**One thing to change: the DB context in `readPreferredProviderId`.** It wraps the `users` read in `withSystemDbAccessContext` from inside a live request. Per the DB-context contract in `CLAUDE.md`, `withSystemDbAccessContext` is for background and seed paths, and calling it inside a request requires `runOutsideDbContext` first — otherwise you're opening a nested system context on a second pooled connection while the request's own transaction is still held. `apps/api/src/middleware/partnerApiAuth.ts:300-310` documents where that goes: the nested insert waits on the outer request's own lock, an application-level self-deadlock that only a statement timeout ends.

There's a simpler fix than adding `runOutsideDbContext`, though: you don't need system scope at all here. `users` is the dual-axis shape, and one branch of its policy is `breeze_current_user_id()` — so a technician reading their own row succeeds under the ambient request context. Dropping to the ambient context is both narrower and avoids the nesting entirely, and it means the read can never resolve a row outside the caller's tenant even if `auth.user.id` were ever influenced by something other than the verified token.

**Smaller note:** the control run description is the most useful part of the verification section — stating plainly that 3 of 6 cases are regression pins that *cannot* fail is more informative than a green checkmark, and I'd like to see that in more PRs.

No migration and no new column, so no cascade or export-policy obligations — `users.preferences` is already `jsonb` and classified `excludedOpen`.

Status: backend approved in principle pending the DB-context change. Answering the two UI questions in a separate comment shortly — don't build the UI until then.

bdunncompany:
Changed in `3bd437ea5` — `readPreferredProviderId` now reads under the ambient request context, with a comment recording why that's the narrower privilege.

Your recommendation is right and I've taken it. **One correction to the mechanism, though, because I verified it before changing the code and it matters for the general rule.**

`withSystemDbAccessContext` delegates straight to `withDbAccessContext` (`db/index.ts:539-541`), and that early-returns when a store already exists:

```ts
export async function withDbAccessContext<T>(context, fn) {
  if (dbContextStorage.getStore()) {
    return fn();
  }
```

Since `middleware/auth.ts:712` has already opened the request's context by the time the route body runs, my call never opened a second transaction and never acquired a second pooled connection — it silently **retained the request's own scope**. The doc comment on `withResolvedDbAccessContext` says as much: *"nested calls retain the existing context"*.

So the self-deadlock at `partnerApiAuth.ts:300-310` isn't reachable from this call path. That deadlock needs a genuinely *new* transaction, which is why the fix there is `runOutsideDbContext` first — that exits the ALS store so the next `withDbAccessContext` really does open one.

What my code actually was is arguably worse than a bug in one respect and better in another: it was **inert ceremony**. It read as a system-scope escalation, and reviewers would reasonably assume the read was privileged, but at runtime it was already doing exactly what you asked for. Removing the wrapper is therefore a **no-op at runtime** and a real improvement in honesty — the code now says what it does.

Your second reason is the one that stands on its own merits, and it's why I didn't just delete the wrapper and move on: reading under the caller's scope means the row can never resolve outside the caller's tenant. `breeze_user_isolation_select` carries `OR id = public.breeze_current_user_id()` (`migrations/2026-04-11-users-rls.sql`), so a technician reading their own row is permitted under ambient scope — no escalation was ever needed. That's now stated in the comment so the next person doesn't "restore" the wrapper.

I left the `withSystemDbAccessContext` on the `partners` read directly above it (`core.ts:1072`) alone — it's pre-existing, it's reading across the org→partner join, and it isn't mine to change in this PR.

**Verification:** `tsc --noEmit` exit 0; `core.remoteAccessLaunch.test.ts` + `remoteAccessLauncher.test.ts` 30 passed; full API suite **1315 files / 21246 passed**, 37 skipped. No behaviour change expected and none observed, consistent with the early-return above.

Holding on the UI until your two answers, as you said.

ToddHebebrand:
Thanks @bdunncompany — this one's in good shape, and the answers to your two open questions are below so you can take it out of draft.

**Where the chooser lives:** the user profile. A "preferred remote tool" control in user settings, writing `users.preferences.remoteAccessProviderId` through the existing PATCH /users merge path. We looked hard at a split button on Connect (default action plus a dropdown for the other enabled providers) and dropped it — it needs the device GET to return the eligible provider list rather than a boolean, a re-validated providerId on the launch POST, and there's no shared dropdown primitive in the web codebase to build on. Not worth it for v1.

**Remembered per device:** no. Per-user only. Per-device memory needs a real table with the full tenancy surface (migration, RLS, four registration lists, contract tests) for marginal benefit over a single preference.

**The backend ships as-is.** The DB-context change in 3bd437ea5 closed the only review finding, and your pushback on the mechanism was right — the nested call was inert ceremony rather than a deadlock risk, and reading under the caller's own scope is the narrower privilege since the users policy already carries the self-read branch. The shape decision that matters here — the preference is an ID that selects from the tenant's list and never supplies a template, password, or field key — is exactly what keeps the post-substitution scheme guard meaningful. Good PR.

**One sequencing note:** merged alone this changes nothing, since nothing writes the preference yet. That's fine and clearly declared, but it means the profile UI lands right behind it or together with it. One wrinkle for whoever builds that UI: the only endpoint exposing the provider list today is partner-scope gated, so the v1 control should be hidden for non-partner sessions rather than shipping an empty select.

**Out of scope for this PR, filed separately:** the provider-ID validation gaps (#3401 — duplicate IDs and dangling defaults predate your change; your preference path just made the existing ambiguity easier to reach, which is on the validator, not on you) and the device-detail GET doing credential work to compute a boolean (#3402). Don't pull either into this branch.

Status: ready from review's side — mark it ready whenever and we'll get the UI queued.

bdunncompany:
Marked ready. Thanks for the decisions — profile-settings control, per-user only, and I'll keep #3401 and #3402 out of this branch.

Both calls make the UI simpler than what I'd have built. The partner-scope wrinkle is the useful catch: the provider list endpoint being partner-gated means a non-partner session has no way to populate the select, so hiding the control there beats rendering an empty one that looks broken. I'd rather that be an explicit capability check than an empty-array coincidence, so the control keys off the session's scope rather than off the list coming back empty — an empty list is also what a tenant with no providers configured returns, and those two states deserve different UI.

Happy to pick the profile UI up as its own PR unless you've already queued it with someone — say the word either way so we don't both build it.

**Unrelated but time-sensitive, since you're online:** `main` is red right now, and it isn't this. `e31490837` is failing 23 checks — every JS job — at `pnpm install --frozen-lockfile`:

```
ERR_PNPM_BROKEN_LOCKFILE ... duplicated mapping key (8090:3)
```

Six dependabot PRs merged inside 114 seconds and two of them appended lockfile entries that already existed (`hono@4.13.1` ×4 in both `packages:` and `snapshots:`, plus `postcss@8.5.26` and `postcss-nested@6.2.0(postcss@8.5.26)` ×2 in `snapshots:`). Fix is up as **#3407** — 26 deleted lines, nothing added, no regeneration, every removed block verified byte-identical to the one kept. `pnpm install --frozen-lockfile` exits 0 on that branch and still fails on `main`.

ToddHebebrand:
**Decision: the profile UI is yours.** Nobody has it queued, and nobody else is building it — pick it up as its own PR.

**Build the chooser against the slim provider-list endpoint, not against GET /orgs/partners/me.** That endpoint is already tracked in #3404, which specifies it as `{id, name, enabled}` per provider and names your dropdown as one of the three consumers that pay for it (alongside the org-restriction reads and the availability check in #3402). Building the select against the partner settings blob would mean a control whose data source is the whole encrypted-provider payload, which is the exposure path #3404 exists to avoid.

It's fine to fold that endpoint into the UI PR rather than waiting on #3404 to land separately — build it to the shape described there so the other two consumers inherit it, and I'll treat #3404's remaining scope as the org-restriction work.

**Your capability-check instinct is right and I'd like it built that way.** Keying the control off the session's scope rather than off an empty list is the correct distinction: a non-partner session that cannot populate the select and a partner tenant with zero providers configured are different states and deserve different UI. An empty-array coincidence would collapse them.

**Don't wait on this PR beyond the API contract** — it's already merged, so the preference field and its validation are on `main` and the UI has a stable surface to write against. There is nothing left to sequence here.

Status: yours — go ahead.

PR #3392, opened 2026-08-10
fix(api): discriminate installer capacity per token, not per key (#3034)

### LINKED ISSUE #3034: Installer capacity is suppressed when an installer is built from a short-link child key

ToddHebebrand:
Closes #3034.

## The defect

#2993 added the `installerTokens: { consumed, max } | null` line to Enrollment Keys rows and gated it on `reportsInstallerCapacity(key) === !key.shortCode`. That per-KEY gate stood in for the real question — **does this token's `max_usage` mean DEVICE SLOTS or DOWNLOADS?** — and it is wrong in *both* directions:

- **False suppression** (what #3034 reported). `GET /:id/installer/:platform` and `POST /:id/bootstrap-token` accept **any** key id the caller can reach, including a short-link child — a visible row on that same page. Building an installer *from* that row mints a genuine `max_usage > 1` capacity token, and the gate hid it.
- **False reporting** (found while fixing the above, and *not* in the issue). `/s/:code` mints a **fresh download key that carries no `short_code`** and then serves a per-download token against it. Those rows were never suppressed at all — `short_code` was never the true discriminator, in either direction.

## The fix

No property of the parent key separates the two mint flows, so the meaning is recorded on the **token**, which is where the issue said the honest fix belonged.

- New column `installer_bootstrap_tokens.usage_kind` — `'capacity' | 'per_download' | 'legacy_unknown'`, DB CHECK enforced.
- `usageKind` is a **required** field on `IssueBootstrapTokenInput` with no default, so every present and future mint site must state which meaning it issues. The three authenticated sites pass `capacity`; `serveInstaller` — the sole public path, reached by both public flows — passes `per_download`.
- `fetchInstallerTokenUsage` filters `usage_kind = 'capacity'` **in SQL**. `reportsInstallerCapacity` and both per-key gates are deleted: the discriminator is the WHERE clause, so there is no app-layer restatement left that could drift from the query, and no id-list pre-filter to keep in sync.

A key whose tokens are all per-download simply aggregates to zero rows and renders `installerTokens: null` — the same shape as a key that never built an installer, so the "3 / 7 over Installer devices 0 / 3" confusion #2992 removed stays removed.

## Coupling handled: the `?expired=` filter (#3191)

`enrollmentKeys.ts` carried an explicit warning that relaxing the gate would bring #3191 back. The filter moves to new **capacity-scoped** guards so it keeps agreeing with the status badge, which derives from `installerTokens`.

The **purge** guards deliberately stay scoped to tokens of *every* kind. Suppressing a confusing number is cosmetic; deleting a key out from under a live installer is irreversible. The residual disagreement therefore stays in the safe direction — the list may call a per-download-only key dead while the purge still spares it, never the reverse.

## Backfill: only what the data proves

`serveInstaller` hardcodes `maxUsage: 1`, so `max_usage > 1` could only have come from an authenticated path. That inference is exact, and those rows backfill to `capacity`.

The converse is **not** provable — a single-slot token may be a per-download token or a genuine one-device Add-Device installer, and nothing on the row separates them. Those stay `legacy_unknown` and are excluded rather than guessed in either direction (guessing `capacity` re-reports a click count; guessing `per_download` asserts a falsehood about real installers). The cost is that a legacy one-device installer loses its capacity *line* until the next build — a display fallback on a self-draining set (24h default token TTL + the nightly cleanup job), and one that never touches the purge guard.

The column DEFAULT is `legacy_unknown`, not `capacity`: it keeps a rolling deploy safe when an old pod's INSERT omits the column, while ensuring a forgetful future writer degrades to showing *nothing* rather than to a confidently wrong denominator.

## Contracts

`installer_bootstrap_tokens` is shape 1 (direct `org_id`, auto-discovered) so no RLS allowlist change. It is already in `CORE_ORG_CASCADE_DELETE_ORDER`; adding a **column** fires the export-policy contract, so `usage_kind` is registered in `CORE_TENANT_EXPORT_POLICY` (`included` — a non-secret classifier, like `installer_platform`).

## Verification

Run against a private per-worktree Postgres (`pnpm test-stack up`); migrations applied from scratch, `usage_kind` column + CHECK verified in `\d`.

- Unit: 8 suites / 237 tests — enrollment-key list/detail/installer/capClamp, issuance, cleanup, `autoMigrate`
- Integration: `installerTokenUsageAggregate` + `enrollmentKeysExpiredFilter` (7), `tenant-export-policy` + `tenantCascade` + `tenantExportErasureRoundtrip` (9), `rls-coverage` (22), purge/cleanup/TTL/collision (18)
- Web: `EnrollmentKeyManager` (19)
- `pnpm db:check-drift` — no drift; `check-migration-naming.sh` — OK

New tests pin the fix in both directions at both layers. The strongest is in `installerTokenUsageAggregate.integration.test.ts`: one short-link child carrying two `per_download` tokens, one `legacy_unknown`, and one 6-slot `capacity` token — every wrong implementation lands on a different number (sum-everything 9/4, drop-only-per-download 7/3, old per-key gate `undefined`). `enrollmentKeysExpiredFilter` gains case (j), a row differing from case (f) *only* in `usage_kind`, which must flip from hidden to visible.

## Note for review

`enrollmentKeyPurgeGuards.ts` now binds the API process clock in a subquery whose sibling filter uses Postgres `NOW()`. That seam is pre-existing and documented; moving it to `NOW()` would be a genuine improvement but touches the purge path, so I left it out of scope rather than widen the blast radius here.


ToddHebebrand:

**Findings:** 9 raised → 8 addressed in `ab54cc1c7`, 1 deferred with rationale; 0 outstanding.

- **Correctness — 0 findings.** All four reviewers independently confirmed: the four `issueBootstrapTokenForKey` call sites are stamped correctly (3 authenticated → `capacity`, `serveInstaller` → `per_download`, and it is the only per-download mint site, reached by both public flows); the migration is idempotent, transaction-clean, row-count-reporting and correctly sorted; the backfill inference (`max_usage > 1` ⇒ capacity) is exact given `serveInstaller` hardcodes `maxUsage: 1`; `CORE_TENANT_EXPORT_POLICY` registration is present with the right bucket; no missed call site repo-wide.
- **The safety invariant that mattered most — verified clean.** Both DELETE paths (nightly `enrollmentKeyCleanup`, `POST /purge-expired`) still call the *unscoped* `hasNoLiveUnexhaustedBootstrapToken()`. Only the cosmetic `?expired=` list filter is capacity-scoped. A live `per_download` or `legacy_unknown` token still blocks deletion, exactly as before.
- **6 stale comments** (addressed) — survivors still describing the deleted `reportsInstallerCapacity` per-key gate, clustered in files whose *primary* docblocks had been rewritten but whose secondary comments were missed: `enrollmentKeys.ts:604/845/1155`, `enrollmentKeysExpiredFilter.integration.test.ts:19/448`, plus a second "Superseded" note on the 2026-07-03 cleanup plan doc.
- **1 test gap** (addressed) — nothing named the realistic short-link shape on the *delete* path. Added purge case (g): an expired key whose only live token is `per_download` must survive. Case (a) covered the same wiring only by accident of the column DEFAULT. **Verified non-vacuous by mutation** — flipping the delete guard to capacity-only fails both (a) and (g).
- **1 belt-and-braces gap** (addressed) — added an assertion that the issuance service actually *persists* `usage_kind`, read back off real Postgres. Every other test observed it only through an aggregate total.
- **1 deferred:** a test for the migration's backfill classification. The reviewer rated it medium-priority/non-blocking; testing it post-hoc means replaying a migration against a DB where the column already exists, which this repo has been bitten by before. The rule is a single unambiguous SQL predicate and the migration self-audits via `RAISE WARNING` on both buckets.

**Tests** (private per-worktree Postgres via `pnpm test-stack up`; migrations applied from scratch, `usage_kind` column + CHECK verified in `\d`):

- `tsc --noEmit` (apps/api) clean
- Unit: 8 suites / 237 passed (two consecutive clean runs of the identical batch)
- Integration: purge + cleanup + TTL + expired-filter + aggregate — 5 files / 23 passed; `rls-coverage` 22 passed; `tenant-export-policy` + `tenantCascade` + `tenantExportErasureRoundtrip` 9 passed
- Web: `EnrollmentKeyManager` 19 passed
- `pnpm db:check-drift` no drift; `check-migration-naming.sh` OK

**One judgement call for the maintainer:** the backfill leaves ambiguous single-slot tokens as `legacy_unknown`, which costs a *legacy* one-device installer its capacity line until the next build. That is deliberate — the alternative is guessing, in a direction that either re-reports a click count or asserts a falsehood — and the set self-drains on the 24h token TTL. Flagging it because it is a (small, transient) visible change beyond the bug itself.

**Status:** review-clean, awaiting maintainer merge. Not merged and #3034 left open per workflow.

PR #3393, opened 2026-08-10
fix(agent): reject malformed reboot delays instead of coercing them to 0 (#3373)

### LINKED ISSUE #3373: [Agent] GetPayloadInt silently coerces malformed delay payloads to 0 — a JSON-string "15" becomes an immediate reboot
## Description
Surfaced during review of PR #3371 (issue #3252). `GetPayloadInt` (`agent/internal/remote/tools/types.go:631`) falls back to `0` when the payload value is not a JSON number. A reboot command carrying `delay: "15"` (string instead of number — e.g. from a hand-built API call or a serializer quirk) is silently treated as `delay: 0` and reboots the device immediately instead of failing validation.

Related cosmetic issue in the same class: `handleRebootSafeMode` (`agent/internal/heartbeat/handlers.go:289`) re-reads `delay` unclamped for the user-facing notification string, so the toast can announce a different delay than the one actually scheduled (the scheduled reboot itself clamps correctly).

## Proposed Fix
Make `GetPayloadInt` (or the reboot handler's use of it) distinguish "absent" from "malformed": accept numeric strings via strict parse, and return a validation error for anything else rather than defaulting to 0 on a destructive action. Align the safe-mode notification string with the clamped value.

## Affected Files
- `agent/internal/remote/tools/types.go`
- `agent/internal/heartbeat/handlers.go`

## Reported By
Independent review agents on PR #3371.

ToddHebebrand:
## Problem

`GetPayloadInt` returned `defaultVal` for **both** an absent key and a present-but-unparseable value. For the reboot/shutdown/safe-mode `delay` field that default is `0`, and `0` legitimately means **"act immediately"** — the two cases are indistinguishable.

So a command payload of `{"delay": "15"}` (a JSON string rather than a number) was read as `0` and rebooted the device **instantly**, with no grace period and no error.

This is reachable, not theoretical: the API validates command payloads as `payload: z.any().optional()` (`apps/api/src/routes/devices/schemas.ts:153`), so a string delay from a hand-built API call, an integration, or a serializer quirk reaches the agent verbatim.

## Fix

**`ParsePayloadInt(payload, key, defaultVal) (int, error)`** — separates *absent* from *malformed*:

| Input | Result |
|---|---|
| key missing, or explicit JSON `null` | `(defaultVal, nil)` |
| `int`, `int64`, integral `float64` | the value |
| numeric string `"15"`, `json.Number` | strict base-10 parse |
| `"soon"`, `""`, `"15m"`, `"15.5"` | **error** |
| `bool`, object, array | **error** |
| non-integral / NaN / infinite / out-of-int-range float | **error** |

"The caller did not specify this field" is a statement a default can stand in for. "The caller specified garbage" is not.

`GetPayloadInt` now delegates to it and swallows the error, so it keeps its lenient contract *and* gains numeric-string support — same accept-set, different failure mode.

**`ShutdownDelayMinutes(payload) (int, error)`** — one source of truth for the `delay` field shared by `Reboot`, `Shutdown`, and `RebootToSafeMode`, so the value announced to users can no longer drift from the value actually scheduled.

Out-of-range delays are **still clamped** (`<0 → 0`, `>1440 → 1440`); only the *type* is now strict. See the note on #3371 below.

## Call-site sweep — decision per site

All 57 `GetPayloadInt` call sites were reviewed. Four were changed:

| Call site | Field | Why a silent 0 was dangerous | Decision |
|---|---|---|---|
| `tools/system.go` `Reboot` | `delay` | 0 = reboot now | **strict** via `ShutdownDelayMinutes` |
| `tools/system.go` `Shutdown` | `delay` | 0 = shut down now | **strict** via `ShutdownDelayMinutes` |
| `tools/system_safemode_windows.go` | `delay` | 0 = instant Safe Mode reboot, BCD flag already set | **strict** via `ShutdownDelayMinutes` |
| `heartbeat/handlers.go` `handleRebootSafeMode` | `delay` | re-read the raw payload for the toast → announced delay could differ from the scheduled one; malformed → warned users of, then performed, an unrequested immediate reboot | **strict**, parsed once before any broadcast |

Two more were hardened because the sweep showed a silent default still produced a wrong destructive outcome:

| Call site | Field | Issue | Decision |
|---|---|---|---|
| `heartbeat/handlers_patch.go` `handleScheduleReboot` | `delayMinutes` | the `1-10080` range check *passes* the silent 60 default, so a malformed value silently scheduled a reboot an hour out | **strict** |
| `tools/computer_action.go` | `x`, `y` | a malformed coordinate clicked the screen's **top-left corner** instead of the target. Absent still means 0, correct for `screenshot`/`key`/`type` | **strict** |

**Left lenient — deliberately.** Everything else falls into one of three safe shapes, and all of them additionally *improve* under this change because a numeric string now parses instead of silently defaulting:

- **Explicitly clamped after read** — `filesystem_analysis.go` (7 sites, all wrapped in `clampInt`), `processes.go`, `services.go`, `tasks.go`, `eventlogs.go`, `fileops.go`, `terminal.go` (`normalizeTerminalSize`), `handlers_screenshot.go` / `handlers_script.go` (`helperCommandTimeout` treats `<=0` as the default).
- **Guarded by an explicit `== 0` rejection** — `processes.go` `KillProcess` and `incident_response.go` `executeProcessKill` already `return error` on `pid == 0`, so a fallback fails safe. (These were the precedent for the fix pattern.)
- **Read-only probes with a non-zero default** — `handlers_network.go`, `handlers_logship.go`. Worst case is a degenerate no-op scan, nothing destructive.

One site in that last group turned out **not** to be benign, and review caught it: `handlers_monitor.go`'s `count` flows straight into `make([]net.IP, count)` with no guard. A negative value **panics** the handler; a large one reserves arbitrary memory on the device. Both are reachable because the API forwards payloads unvalidated. Pre-existing and not introduced by the strict-parse change — but this PR's sweep claimed the site was safe, so `clampPingCount` (second commit) makes that claim true.

## Also

The safe-mode notification renders from the parsed/clamped delay and pluralises `1 minute` correctly (previously `in 1 minutes`).

## Overlap with #3371 — expect a conflict

**#3371 (issue #3252, reboot delay unit normalization) is still open** and edits the same `Reboot`/`Shutdown`/`RebootToSafeMode` delay lines. This PR is based on `main` (`81ce580`), **not stacked** — stacked PRs get no CI on this repo.

Conflict is textual and small. Whichever merges second:

- #3371 introduces `maxShutdownDelayMinutes` + `clampShutdownDelayMinutes()`; this PR introduces `maxShutdownDelayMinutes` + `ShutdownDelayMinutes()`. **Resolution: keep one `maxShutdownDelayMinutes`, and have `ShutdownDelayMinutes` call `clampShutdownDelayMinutes` instead of its own inline clamp.** The two are complementary — #3371 owns *units and clamping*, this PR owns *parsing and validation*.
- I deliberately **kept the negative-clamps-to-0 behaviour** rather than making it an error, because #3371 explicitly tests it (`{-1000, 0}`, `{-1, 0}`). Reversing a sibling PR's tested decision from here would be wrong. Worth a follow-up: `delay: -5` silently becoming an immediate reboot is the same bug class as this issue.

## Reviewed and deliberately NOT changed — needs your call

`handleRebootSafeMode` broadcasts the "save your work" notification **before** `tools.RebootToSafeMode` is called and confirmed to succeed. If the BCD flag can't be set or `shutdown /r` fails, every logged-in user gets a critical toast for a reboot that never happens.

The review flagged this as the same hazard class this PR reasons about, and that's fair. I left it alone on purpose:

- The ordering is **pre-existing** — this PR changed *what* is broadcast, not *when*.
- Reversing it has a real cost. On `delay: 0` the reboot begins immediately, so a post-success broadcast races the shutdown and users may get no warning at all.
- That's a UX judgment on a destructive Windows path with no repo precedent, which belongs in its own change rather than a drive-by in a parsing PR.

Happy to do it here or in a follow-up issue — your call.

## Verification

- `go test -race ./internal/remote/tools/...` — **ok** (29.4s, full package)
- `go test -race ./internal/heartbeat/...` — **ok** (39.0s, full package)
- `go vet` clean on `GOOS=windows`, `darwin`, `linux` (the windows vet is what typechecks `system_safemode_windows.go`)
- Cross-builds OK: `windows/amd64`, `linux/amd64`, `linux/arm64`, `darwin/arm64`, `darwin/amd64`
  - `windows/386` fails, but on **pre-existing** code unrelated to this change (`internal/patching/hresult.go` HRESULT constants overflow `int32`) — the agent is 64-bit only.
- **Mutation-tested**: reverting `ShutdownDelayMinutes` to the old lenient read fails 18 subtests across both packages, so the tests are not vacuous.

### A note on the tests

They deliberately **never call `Reboot`/`Shutdown`** — those run the host's `shutdown` binary, and a regression of this very fix would turn a malformed-input test case into an actual reboot of whatever machine ran the suite. The parse/clamp helper is covered directly instead, and handler wiring is asserted through `handleRebootSafeMode`, whose underlying tool is an inert stub off Windows.

Closes #3373

### DIFF
--- a/agent/internal/heartbeat/handlers.go
+++ b/agent/internal/heartbeat/handlers.go
@@ -285,17 +285,38 @@ func handleRefreshInventory(h *Heartbeat, _ Command) tools.CommandResult {
 }
 
 func handleRebootSafeMode(h *Heartbeat, cmd Command) tools.CommandResult {
+	start := time.Now()
+
+	// Resolve the delay up front, through the same helper RebootToSafeMode
+	// uses, for two reasons (issue #3373):
+	//   1. The notification below must announce the delay the reboot is
+	//      actually scheduled with. Reading the raw payload separately let the
+	//      toast and the scheduled reboot disagree once either side clamped.
+	//   2. A malformed delay must fail the command here, before anything is
+	//      broadcast — otherwise users are warned of, and then subjected to, an
+	//      immediate forced reboot they never asked for.
+	delay, err := tools.ShutdownDelayMinutes(cmd.Payload)
+	if err != nil {
+		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+
 	if h.sessionBroker != nil {
-		delay := tools.GetPayloadInt(cmd.Payload, "delay", 0)
-		var msg string
-		if delay > 0 {
-			msg = fmt.Sprintf("System will reboot into Safe Mode with Networking in %d minutes. Please save all work.", delay)
-		} else {
-			msg = "System is rebooting into Safe Mode with Networking. Please save all work."
-		}
-		h.sessionBroker.BroadcastNotification("Safe Mode Reboot", msg, "critical")
+		h.sessionBroker.BroadcastNotification("Safe Mode Reboot", safeModeRebootNotice(delay), "critical")
 	}
 	return tools.RebootToSafeMode(cmd.Payload)
+}
+
+// safeModeRebootNotice renders the user-facing Safe Mode warning for a delay
+// that has already been parsed and clamped by tools.ShutdownDelayMinutes.
+func safeModeRebootNotice(delayMinutes int) string {
+	if delayMinutes <= 0 {
+		return "System is rebooting into Safe Mode with Networking. Please save all work."
+	}
+	unit := "minutes"
+	if delayMinutes == 1 {
+		unit = "minute"
+	}
+	return fmt.Sprintf("System will reboot into Safe Mode with Networking in %d %s. Please save all work.", delayMinutes, unit)
 }
 
 func handleCollectSoftware(_ *Heartbeat, cmd Command) tools.CommandResult {
--- a/agent/internal/heartbeat/handlers_monitor.go
+++ b/agent/internal/heartbeat/handlers_monitor.go
@@ -21,6 +21,20 @@ func init() {
 	handlerRegistry[tools.CmdNetworkDnsCheck] = handleNetworkDnsCheck
 }
 
+// maxPingCount bounds the ping-sweep repeat count. Monitor checks never need
+// more than a handful of probes, and the value sizes a slice allocation.
+const maxPingCount = 60
+
+func clampPingCount(count int) int {
+	if count < 1 {
+		return 1
+	}
+	if count > maxPingCount {
+		return maxPingCount
+	}
+	return count
+}
+
 func handleNetworkPing(_ *Heartbeat, cmd Command) tools.CommandResult {
 	start := time.Now()
 	target, errResult := tools.RequirePayloadString(cmd.Payload, "target")
@@ -31,7 +45,10 @@ func handleNetworkPing(_ *Heartbeat, cmd Command) tools.CommandResult {
 
 	monitorId := tools.GetPayloadString(cmd.Payload, "monitorId", "")
 	timeout := time.Duration(tools.GetPayloadInt(cmd.Payload, "timeout", 5)) * time.Second
-	count := tools.GetPayloadInt(cmd.Payload, "count", 4)
+	// count sizes an allocation below, so it must be bounded before use: a
+	// negative value panics make(), and a large one lets an unvalidated
+	// payload field reserve arbitrary memory on the device.
+	count := clampPingCount(tools.GetPayloadInt(cmd.Payload, "count", 4))
 
 	ip := net.ParseIP(target)
 	if ip == nil {
--- a/agent/internal/heartbeat/handlers_monitor_count_test.go
+++ b/agent/internal/heartbeat/handlers_monitor_count_test.go
@@ -1,0 +1,36 @@
+package heartbeat
+
+import "testing"
+
+// count sizes a make([]net.IP, count) allocation from a command payload the
+// API forwards unvalidated (payload is z.any()). A negative value panics the
+// handler and a large one reserves arbitrary memory on the device, so the
+// value has to be bounded before it reaches make().
+func TestClampPingCount(t *testing.T) {
+	tests := []struct {
+		name string
+		in   int
+		want int
+	}{
+		{"negative would panic make()", -1, 1},
+		{"large negative", -1 << 30, 1},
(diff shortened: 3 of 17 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter (both on the full `main...HEAD` diff).

**Findings:** 2 raised, 1 fixed, 1 deliberately deferred.

- **Fixed in `2186ca268`** — `handlers_monitor.go`'s `count` reaches `make([]net.IP, count)` unguarded: negative panics the handler, large reserves arbitrary memory. Pre-existing, but this PR's sweep table claimed the site was benign, so `clampPingCount` makes that true. Table-driven test added.
- **Deferred (needs a maintainer call)** — `handleRebootSafeMode` broadcasts the "save your work" toast *before* `RebootToSafeMode` is confirmed to succeed, so a BCD/exec failure leaves users warned about a reboot that never happens. Pre-existing ordering; reversing it costs the pre-warning on the `delay: 0` path. Reasoning in the PR body under "Reviewed and deliberately NOT changed."

**Verified clean by review:** the `intFromFloat64` bounds check was independently re-derived as correct on 32- and 64-bit (float64 spacing near 2^63 is 2048, so no accepted value can overflow the `int(f)` cast); all ~50 existing `GetPayloadInt` callers confirmed safe under the new numeric-string acceptance; no double-clamp or swallowed error in the `ShutdownDelayMinutes` routing.

**Tests:** `go test -race` green on both full packages — `internal/remote/tools` (29.4s) and `internal/heartbeat` (38.6s). `go vet` clean on `GOOS=windows`/`darwin`/`linux`. Cross-builds OK on windows/amd64, linux/amd64+arm64, darwin/amd64+arm64.

**Mutation-tested:** reverting `ShutdownDelayMinutes` to the old lenient read fails 18 subtests across both packages — the tests are not vacuous.

**Status:** review-clean, awaiting maintainer merge. Note the #3371 overlap in the PR body — expect a small conflict on the `Reboot`/`Shutdown` delay lines whichever merges second.

Issue #2021, opened 2026-06-29
Built-in BitLocker / FileVault encryption key management


win-wxx:
### Problem

There's currently no built-in way to manage disk encryption across devices. To collect BitLocker recovery keys today, I have to use a custom PowerShell script (Get-BitLockerVolume) that outputs the key to Script Results, then read and store each key manually. This doesn't scale and there's no central place to see which devices are encrypted or to look a key up later.

My use case: I want to confirm every managed device has encryption enabled, automatically keep a copy of each recovery key in case a user is locked out, and be able to retrieve a specific device's key quickly when needed.

### Proposed Solution

A dedicated encryption-management feature for BitLocker (Windows) and FileVault (macOS):

- Auto-detect encryption status across all devices, shown on the device detail page and as a fleet-wide view
- Automatically collect and store recovery keys, encrypted at rest
- A simple way to look up a device's recovery key when needed, ideally with a record of who viewed it and when
- Optionally flag devices that aren't encrypted or have no stored key

Having this as a first-class feature would be much cleaner than stitching together a script plus custom-field storage, and key recovery is important enough to warrant dedicated handling.

### Alternatives Considered

I built a workaround with a PowerShell script that runs Get-BitLockerVolume and writes the recovery key to Script Results. It works for reading the key, but it's manual — I have to copy each key out of the run output myself.

I also tried having the script write the key into a device custom field via the API, but custom-field writes via API key return 401 (only a browser session token seems to be accepted), so that path is currently blocked. Even if it worked, a purpose-built encryption feature would be a better solution than this approach.

### Component

Agent

### Scope

Large (new module or major change)

ToddHebebrand:
Acknowledged — first-class BitLocker/FileVault recovery-key escrow is a worthwhile feature: fleet-wide encryption status, automatic key collection encrypted at rest, audited lookup (who-viewed-when), and a flag for unencrypted/no-key devices. This is appropriately scoped as Large — key escrow has real handling/audit requirements — and goes on the roadmap rather than as a quick add.

**Separately, you surfaced a distinct bug worth its own report:** custom-field writes via API key returning **401** (only a browser session token accepted). If the custom-field write endpoint is rejecting valid API keys, that's an auth/RBAC gap independent of this feature and smaller to fix — it also unblocks your interim script-based workaround. I'll split that out so it can be investigated on its own track; the encryption feature shouldn't gate fixing it.

Keeping this open for the encryption feature.

[...]

PR #3394, opened 2026-08-10
fix(api,secrets): seal plaintext into registered columns without requiring APP_ENCRYPTION_KEY_ID

bdunncompany:
Found while answering the `APP_ENCRYPTION_KEY_ID` question on #3239. Separate bug, same theme: a knob whose absence is fatal at write time rather than loud at boot.

## The defect

`encryptColumnValueForWrite` is the documented mandatory path for every mutating route that writes a registered column — "Otherwise a UI edit silently re-writes the column as plaintext". Its plaintext branch routes through the wrong primitive:

```ts
if (!force) {
  return value;
}
return reencryptSecret(value, opts) ?? value;   // value is known NOT encrypted here
```

`reencryptSecret` is the **rotation** primitive and requires a target key id:

```ts
const activeKeyId = getActiveKeyId();
if (!activeKeyId) {
  throw new Error('APP_ENCRYPTION_KEY_ID is required to re-encrypt secrets');
}
```

`encryptSecret` — the **first-encryption** primitive — deliberately handles the same configuration:

```ts
// No active key id: fall back to v1 (legacy global key). AAD is not supported
// for v1 since v1 ciphertext predates key rotation; callers wanting AAD must
// set APP_ENCRYPTION_KEY_ID.
return encryptWithKey(value, getEncryptionKey(), ENCRYPTED_V1_PREFIX);
```

So on a deployment with no `APP_ENCRYPTION_KEY_ID`, writing a **plaintext** secret into any registered column throws and 500s the request — while the codebase's own encrypt primitive would have sealed it to v1 without complaint.

`.env.example:339` ships `APP_ENCRYPTION_KEY_ID=` **empty**, and `config/validate.ts:1816` only requires it once `M365_GRAPH_ACTIONS_TOOLS_ENABLED=true`. The default self-hosted configuration is therefore the affected one.

## Why the rotation path is already correct, and this one isn't

The already-encrypted branch degrades cleanly, because `shouldReencryptSecret` returns `false` when there is no active key id — a rotation with no target is genuinely a no-op. That asymmetry is the whole bug: rotation was taught to stand down without a key id, first encryption was not, and it shares a helper with it.

This is not fail-closed behaviour being preserved. v1 *is* encryption at rest under the global key, so sealing to v1 is exactly what every other write path in the system does in this configuration. The throw buys no confidentiality; it only converts a working write into a 500.

## Blast radius

All 40 registry entries, across all three kinds — `text`, `text-array`, and `json` (the latter whenever a sub-field name is in `SECRET_JSON_KEYS`). Concretely that includes `device_recovery_keys.encrypted_key` (BitLocker/FileVault escrow, #2021), `notification_channels.config`, `webhooks.secret`, `snmp_devices.community`, `psa_connections.credentials`, and the `partners`/`organizations`/`sites` `settings` columns.

It only fires on **first** encryption of a given value, so a deployment that has already sealed its secrets keeps working, and one that never sets the knob hits it the first time an operator saves a new credential. That is what makes it easy to miss in testing and expensive in production — the API boots healthy and passes its health check, then a specific class of write fails.

I hit this shape on a self-hosted 0.96.0 fleet: healthy boot, then every encrypted-column write returning 500. The operator-side resolution then was to set `APP_ENCRYPTION_KEY_ID`, which works but is a workaround for a knob the docs presented as optional.

## The fix

One line, plus the import — route first encryption through `encryptSecret`:

```ts
return encryptSecret(value, opts) ?? value;
```

`reencryptSecret` stays exactly where it belongs, on the rotation branch. When a key id *is* configured, behaviour is byte-for-byte unchanged: both primitives seal to `enc:v2:<id>:` (or v3 under `ENABLE_AAD_V3`) via the same `encryptWithKey`.

Deliberately **not** changed: `reencryptSecret`'s guard. It is correct — rotating to an unspecified key id is genuinely an error, and softening it would hide real misconfiguration in the batch rotation walker.

## Verification

**Reproduced first, against unmodified `origin/main`** — the failure is a real stack through the public entry point, not a constructed one:

```
Error: APP_ENCRYPTION_KEY_ID is required to re-encrypt secrets
 ❯ reencryptSecret src/services/secretCrypto.ts:456:11
 ❯ maybeReencryptString src/services/encryptedColumnRegistry.ts:143:10
 ❯ transformEncryptedColumnValue src/services/encryptedColumnRegistry.ts:177:40
 ❯ encryptColumnValueForWrite src/services/encryptedColumnRegistry.ts:206:10
```

**Why it was never caught:** every pre-existing case in `encryptedColumnRegistry.test.ts` sets `APP_ENCRYPTION_KEY_ID`. The unset configuration — the shipped default — had no coverage at all. Five cases now cover it: plaintext into `text`, `json`, and `text-array`; rotation still reaching `enc:v2:current:` when a key id **is** set; and existing ciphertext left byte-identical when one is not.

**Control run:** with just the production line reverted to `reencryptSecret`, exactly the 3 plaintext cases fail with the error above; the other 2 pass in both states, as they should — so the new cases pin the bug rather than describe the code.

**Full local gate, all green:**

- `vitest run` (apps/api, full suite) — **1315 files / 21245 passed**, 37 skipped, exit 0
- Trivy `fs --scanners secret,misconfig --severity HIGH,CRITICAL` on `apps/api/src/services` — clean, exit 0
- No Go and no dependency/lockfile changes, so `govulncheck` and `pnpm audit` are untouched by this diff

Two files, +94/−1. No schema, migration, or dependency changes.

## One judgement call worth flagging

With no key id and `ENABLE_AAD_V3=true`, the v1 fallback silently drops AAD — that is `encryptSecret`'s documented, pre-existing behaviour, and this change simply inherits it rather than introducing it. Sealing to v1 unbound beats refusing the write, but if you would rather AAD-desired-but-unavailable be loud, the honest place is a boot-time assertion that `ENABLE_AAD_V3` requires `APP_ENCRYPTION_KEY_ID` — which is the "wired ⇒ documented/validated" direction I raised on #3239, and I would rather it be its own change than smuggled in here.

### DIFF
--- a/apps/api/src/services/encryptedColumnRegistry.test.ts
+++ b/apps/api/src/services/encryptedColumnRegistry.test.ts
@@ -90,6 +90,92 @@ describe('encryptedColumnRegistry', () => {
     expect(decryptSecret(transformed.nested.authToken)).toBe('old-token');
   });
 
+  describe('first encryption with no APP_ENCRYPTION_KEY_ID (the shipped default)', () => {
+    // Every other case in this file sets APP_ENCRYPTION_KEY_ID, which is why
+    // this path went uncovered: plaintext was routed through reencryptSecret,
+    // whose missing-key-id guard throws and fails the write.
+
+    it('seals a plaintext text column to v1 instead of throwing', () => {
+      setEncryptionEnv({ APP_ENCRYPTION_KEY: 'legacy-key-material' });
+
+      const transformed = transformEncryptedColumnValue({
+        table: 'device_recovery_keys',
+        column: 'encrypted_key',
+        kind: 'text',
+        description: 'test',
+      }, 'plaintext-recovery-key');
+
+      expect(transformed).toMatch(/^enc:v1:/);
+      expect(decryptSecret(transformed as string)).toBe('plaintext-recovery-key');
+    });
+
+    it('seals a plaintext JSON secret field to v1 and leaves non-secrets alone', () => {
+      setEncryptionEnv({ APP_ENCRYPTION_KEY: 'legacy-key-material' });
+
+      const transformed = transformEncryptedColumnValue({
+        table: 'notification_channels',
+        column: 'config',
+        kind: 'json',
+        description: 'test',
+      }, {
+        label: 'do-not-encrypt',
+        nested: { apiKey: 'plaintext-api-key' },
+      }) as { label: string; nested: { apiKey: string } };
+
+      expect(transformed.label).toBe('do-not-encrypt');
+      expect(transformed.nested.apiKey).toMatch(/^enc:v1:/);
+      expect(decryptSecret(transformed.nested.apiKey)).toBe('plaintext-api-key');
+    });
+
+    it('seals plaintext entries of a text-array column to v1', () => {
+      setEncryptionEnv({ APP_ENCRYPTION_KEY: 'legacy-key-material' });
+
+      const transformed = transformEncryptedColumnValue({
+        table: 'discovery_profiles',
+        column: 'snmp_communities',
+        kind: 'text-array',
+        description: 'test',
+      }, ['public', 'private']) as string[];
+
+      expect(transformed).toHaveLength(2);
+      for (const entry of transformed) {
+        expect(entry).toMatch(/^enc:v1:/);
+      }
+      expect(transformed.map((e) => decryptSecret(e))).toEqual(['public', 'private']);
+    });
+
+    it('still rotates to the active key id when one IS configured', () => {
+      setEncryptionEnv({
+        APP_ENCRYPTION_KEY: 'current-key-material',
+        APP_ENCRYPTION_KEY_ID: 'current',
+      });
+
+      const transformed = transformEncryptedColumnValue({
+        table: 'device_recovery_keys',
+        column: 'encrypted_key',
+        kind: 'text',
+        description: 'test',
+      }, 'plaintext-recovery-key');
+
+      expect(transformed).toMatch(/^enc:v2:current:/);
+      expect(decryptSecret(transformed as string)).toBe('plaintext-recovery-key');
+    });
+
+    it('leaves existing ciphertext untouched when no key id is configured', () => {
+      setEncryptionEnv({ APP_ENCRYPTION_KEY: 'legacy-key-material' });
+      const existing = encryptSecret('already-sealed');
+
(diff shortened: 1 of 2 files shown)


ToddHebebrand (approved):
**Approving — the diagnosis and the fix are both correct, and the relaxation doesn't weaken the invariant.**

**No cleartext path exists.** `encryptSecret` (`apps/api/src/services/secretCrypto.ts:316-347`) returns `null` only for falsy input, so the `?? value` fallback at `apps/api/src/services/encryptedColumnRegistry.ts:143` can only ever yield `''`. Every other outcome is ciphertext or a throw, and `getEncryptionKey` throws in production without `APP_ENCRYPTION_KEY` (`secretCrypto.ts:101`), so fail-closed holds.

**Not a weakening.** v1 is still AES-GCM under the global key — the prior behavior on this path was a 500, not confidentiality.

**Reader side handles both shapes.** `decryptSecret` handles v1 ahead of the v2/v3 block (`secretCrypto.ts:361`), and no registered-column reader passes `strict: true` — its only use is `actionIntents/resultSecrets.ts:81`, which is off this path entirely.

**Scope is complete.** `reencryptSecret` has exactly two call sites; the other (`encryptedColumnRegistry.ts:137`) is gated by `shouldReencryptSecret`, which returns `false` without a key id, so its throw is unreachable. And the bug was reachable on the shipped default — `recoveryKeyEscrow.ts:92` always feeds plaintext in.

No migration, table or column, so no RLS, cascade-list or export-policy obligations. Diff matches the body exactly.

**Follow-ups, none blocking:** with `ENABLE_AAD_V3=true` and no key id, the computed AAD at `encryptedColumnRegistry.ts:133` is silently discarded by the v1 fallback (`secretCrypto.ts:343`) — pre-existing, and you flagged it honestly. Worth noting that `ENABLE_AAD_V3` appears in no `.env.example` and has no `config/validate.ts` entry, so it's an undocumented knob; the boot assertion you suggested (`ENABLE_AAD_V3` implies `APP_ENCRYPTION_KEY_ID`) is the right shape for its own PR. A test pinning the AAD-drop behavior would stop a future "fix" reintroducing the throw. And the `?? value` at `:143` is now dead code that visually reads as a plaintext escape hatch on the most sensitive surface in the repo — an explicit throw would document that it can't fire.

Status: approved.

PR #3396, opened 2026-08-10
fix(api): match process-sample path id against the agent id, not the device id (#3387)

### LINKED ISSUE #3387: POST /api/v1/agents/:id/process-sample always 403 — path id treated as deviceId but agent sends AgentID
## Summary

Agent process sampling is broken on self-hosted installs: every `POST /api/v1/agents/<agentId>/process-sample` returns **403 Forbidden**, so `device_process_samples` stays empty and agents/Caddy logs fill with warn noise.

Root cause is an **ID-type mismatch** between agent and API. Other agent ingest routes treat `:id` as **agentId**; `processSample` uniquely compares `:id` to **`agent.deviceId`**.

Still present on **v0.104.0** source (verified against the tagged tree). Originally reported on 0.101.0 / re-verified on 0.102.0.

## Environment

- Self-hosted Breeze (observed 0.101.0 → 0.102.0; code still wrong on **v0.104.0**)
- Healthy agents (Linux + macOS): heartbeat / inventory / other ingest OK
- Symptom: `device_process_samples` = 0 rows; repeated 403 on process-sample

## Expected

`POST /api/v1/agents/{agentId}/process-sample` with a valid agent bearer token returns **201** and inserts a row (same ID convention as inventory/connections/heartbeat).

## Actual

```http
POST /api/v1/agents/<64-char-hex-agentId>/process-sample
→ 403 {"error":"Forbidden"}
```

(~7–30ms; agents retry forever)

### Agent (v0.104.0) — sends **AgentID** in the path

`agent/internal/heartbeat/heartbeat.go` → `sendProcessSample()`:

```go
url := fmt.Sprintf("%s/api/v1/agents/%s/process-sample", h.serverURL(), h.config.AgentID)
```

Same pattern as heartbeat / inventory / monitoring-results (`/api/v1/agents/%s/...` + `AgentID`).

### API (v0.104.0) — compares path to **deviceId**

`apps/api/src/routes/agents/processSample.ts`:

```ts
const deviceId = c.req.param('id');
const agent = c.get('agent') as AgentAuthContext | undefined;

if (!agent || agent.deviceId !== deviceId) {
  return c.json({ error: 'Forbidden' }, 403);
}
```

Auth already resolved the device from the bearer token (`agent.deviceId`, `agent.agentId`, `agent.orgId`). The path check then fails because:

- path `:id` = **agentId** (hex)
- `agent.deviceId` = **device UUID**

### Contrast — sibling routes do it correctly

`apps/api/src/routes/agents/inventory.ts` (and connections, etc.):

```ts
const agentId = c.req.param('id');
// lookup / authorize via devices.agentId === agentId
```

`processSample` is the odd one out. Unit tests in `processSample.test.ts` encode the bug (they pass **deviceId** in the path), so CI stays green while production agents 403.

## Suggested fix (for Claude / implementer)

**Preferred (align with inventory/connections):**

1. In `processSample.ts`, treat `:id` as **agentId**:
   - Fail closed if `!agent || agent.agentId !== c.req.param('id')` (timing-safe compare if you do that elsewhere), **or**
   - Resolve device by `devices.agentId` then assert `device.id === agent.deviceId`.
2. Always persist with `deviceId: agent.deviceId` / `orgId: agent.orgId` from auth (already correct).
3. Update `processSample.test.ts`:
   - Happy path: request `/{agentId}/process-sample` with auth context `{ agentId, deviceId, orgId, role: 'agent' }`.
   - 403 when path agentId ≠ token agentId.
   - Keep watchdog-role 403 test (F8).
4. Optional hardening: if you want defense-in-depth, accept **either** agentId or deviceId in the path (both must match the token) — not required if you standardize on agentId.

**Do not** “fix” only the agent to send `deviceId` unless you also audit every other `/agents/:id/*` caller — agent-wide convention is AgentID.

## Repro

1. Enroll any agent on self-hosted Breeze ≥ 0.101 (still broken in 0.104.0 code).
2. Wait for process sampler tick (`ProcessSampleIntervalSeconds`).
3. Observe API/access logs: `POST /api/v1/agents/<agentId>/process-sample` → 403.
4. `SELECT count(*) FROM device_process_samples` → 0.

## Acceptance criteria

- [ ] Agent process-sample POST with AgentID in path returns 201 on a healthy agent token.
- [ ] Row appears in `device_process_samples` with correct `device_id` / `org_id`.
- [ ] Watchdog-role tokens still 403 (F8).
- [ ] Tests request with **agentId** in the path (matching production).
- [ ] No regression on inventory/heartbeat/connections ID convention.

## References

- Discord `#bugs` (SemoTech): `1531130946382856213` (report), `1531381782841266339` (still broken on 0.102.0)
- Files: `apps/api/src/routes/agents/processSample.ts`, `apps/api/src/routes/agents/processSample.test.ts`, `agent/internal/heartbeat/heartbeat.go` (`sendProcessSample`)

bdunncompany:
Closes #3387. @SemoTech's diagnosis is exactly right; I verified every claim against `origin/main` before changing anything, and the fix is the one-line direction they proposed.

## Confirmed

`processSample` is the only agent ingest route that reads `:id` as a **device** id:

```ts
const deviceId = c.req.param('id');
if (!agent || agent.deviceId !== deviceId) {
  return c.json({ error: 'Forbidden' }, 403);
}
```

The agent sends its **AgentID** (`agent/internal/heartbeat/heartbeat.go:1849`):

```go
url := fmt.Sprintf("%s/api/v1/agents/%s/process-sample", h.serverURL(), h.config.AgentID)
```

and every sibling route agrees with the agent — `inventory.ts:32` reads `const agentId = c.req.param('id')` and resolves via `eq(devices.agentId, agentId)`. So the two values can never be equal and **every** real process sample 403s. `device_process_samples` stays empty on any deployment, and agents retry forever.

## Fix

Compare the path id to `agent.agentId`. `AgentAuthContext` already carries `agentId` alongside `deviceId`, so nothing new is resolved and there is no extra query.

The security property is unchanged: the path id must still match the authenticated token, and the row is still keyed by `agent.deviceId` / `agent.orgId` resolved server-side from the token — the agent payload is still never trusted for tenancy.

**Direction of the fix.** The API moves, not the agent. The agent's convention is shared by every other ingest route, and changing the agent instead would break already-deployed fleets against a fixed server. There is also no working population to preserve: since the path has always carried the AgentID, this endpoint has never once succeeded, so no deployment depends on the current behaviour.

## Why CI stayed green

The pre-fix tests put the **device** id in the path — a value no agent ever sends — and set `agentId: 'a'` where it could not be observed. They asserted the bug.

Fixed by making the two ids visibly distinct (`deviceId: 'dev-1'`, `agentId: 'agent-1'`) and driving the routes the way the agent actually does. New case `accepts the agent id in the path, not the device id (#3387)` pins both directions at once: `/agent-1/...` → **201** with the row still keyed to `dev-1`, and `/dev-1/...` → **403**. Because the ids now differ, an implementation that compares against either field alone fails one half.

## Verification

- **Control run:** with only the production line reverted to `agent.deviceId !== agentId`, 3 of 6 cases fail, including the new guard. The other 3 (body cap, id mismatch, watchdog-role gate) pass in both states, as they should — they don't depend on which id `:id` means.
- `vitest run` (apps/api, full suite) — **1315 files / 21241 passed**, 37 skipped, exit 0
- Trivy `fs --scanners secret,misconfig --severity HIGH,CRITICAL` on `apps/api/src/routes/agents` — clean, exit 0
- No Go, schema, migration, or dependency changes

Two files, +38/−12.

## One thing left deliberately alone

`processSampleSchema` still accepts a body `orgId` that the handler ignores (the first test pins that it's ignored). Tightening the schema to reject unknown keys is a wider change across the agent ingest surface and would be its own PR — flagging it rather than folding it in here.

### DIFF
--- a/apps/api/src/routes/agents/processSample.test.ts
+++ b/apps/api/src/routes/agents/processSample.test.ts
@@ -22,8 +22,8 @@ describe('POST /:id/process-sample', () => {
   beforeEach(() => { inserted.length = 0; });
 
   it('derives org_id from the auth context and ignores any body org_id', async () => {
-    const app = appWithAgent({ deviceId: 'dev-1', orgId: 'org-real', agentId: 'a', siteId: 's', role: 'agent' });
-    const res = await app.request('/dev-1/process-sample', {
+    const app = appWithAgent({ deviceId: 'dev-1', orgId: 'org-real', agentId: 'agent-1', siteId: 's', role: 'agent' });
+    const res = await app.request('/agent-1/process-sample', {
       method: 'POST',
       headers: { 'content-type': 'application/json' },
       body: JSON.stringify({
@@ -38,8 +38,8 @@ describe('POST /:id/process-sample', () => {
   });
 
   it('server-stamps timestamp and stores agent time separately', async () => {
-    const app = appWithAgent({ deviceId: 'dev-1', orgId: 'org-real', agentId: 'a', siteId: 's', role: 'agent' });
-    await app.request('/dev-1/process-sample', {
+    const app = appWithAgent({ deviceId: 'dev-1', orgId: 'org-real', agentId: 'agent-1', siteId: 's', role: 'agent' });
+    await app.request('/agent-1/process-sample', {
       method: 'POST',
       headers: { 'content-type': 'application/json' },
       body: JSON.stringify({ timestamp: '2020-01-01T00:00:00.000Z', processes: [] })
@@ -49,9 +49,9 @@ describe('POST /:id/process-sample', () => {
   });
 
   it('rejects a payload over the 16-process cap', async () => {
-    const app = appWithAgent({ deviceId: 'dev-1', orgId: 'org-real', agentId: 'a', siteId: 's', role: 'agent' });
+    const app = appWithAgent({ deviceId: 'dev-1', orgId: 'org-real', agentId: 'agent-1', siteId: 's', role: 'agent' });
     const processes = Array.from({ length: 17 }, (_, i) => ({ name: `p${i}`, pid: i, cpu: 0, ramMb: 0 }));
-    const res = await app.request('/dev-1/process-sample', {
+    const res = await app.request('/agent-1/process-sample', {
       method: 'POST',
       headers: { 'content-type': 'application/json' },
       body: JSON.stringify({ timestamp: '2026-06-13T12:00:00.000Z', processes })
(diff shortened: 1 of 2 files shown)


ToddHebebrand (approved):
**Approving — the authorization story here is sound, and stronger than the PR body claims.**

**No cross-device or cross-org write is reachable.** `agentAuthMiddleware` resolves the device by the path id itself — `apps/api/src/middleware/agentAuth.ts:343` reads `c.req.param('id')`, `:374` selects `where(eq(devices.agentId, agentId))`, and `:622-624` sets `agent.{deviceId,agentId,orgId}` from that row only after the bearer-token hash matches it. Posting to `/agent-B/process-sample` therefore requires device B's own token. The handler keys the insert on `agent.deviceId` and `agent.orgId` (`apps/api/src/routes/agents/processSample.ts:33-34`), never on the path param or the body — which is already the derive-from-the-authenticated-principal shape, not merely a swapped comparison.

**Correct tenant DB context, RLS enforced.** `process-sample` is not in `SELF_MANAGED_DB_CONTEXT_ACTIONS` (`agentAuth.ts:267`), so the handler runs inside the org-scoped `withDbAccessContext` at `agentAuth.ts:647` rather than a system context, and `apps/api/migrations/2026-06-13-device-process-samples.sql:29-30` gives the table `FORCE ROW LEVEL SECURITY` plus `INSERT WITH CHECK (breeze_has_org_access(org_id))`. The table is already in all three device cascade lists and the export policy.

**The 403 was corrected rather than deleted** (`processSample.ts:28`), and the new test pins both directions (`processSample.test.ts:63-84`), failing against the old line. Diff matches the body, and it does resolve #3387 as reported — `agent/internal/heartbeat/heartbeat.go:1849` sends `config.AgentID`.

**Two notes, neither blocking.** The guard at `processSample.ts:28` is now tautological in production, since `agent.agentId` *is* the path param by construction — harmless defense-in-depth that matches the sibling routes, but the body's "the path id must still match the token" overstates its residual value. And the 403 case in the unit suite only passes because the harness injects a hand-built context, bypassing `agentAuth` entirely; a single integration test posting device B's agent id with device A's token would prove the real cross-tenant property, which a fully-mocked `db` structurally cannot.

Status: approved.

PR #3398, opened 2026-08-10
fix(agent): route every content-touching file op through containment (#3397)

### LINKED ISSUE #3397: [Agent][Security] File-browser containment bypass: CopyFile and six sibling ops skip enforceReadContainment — copy-then-read defeats the sensitive-path deny-list
## Description
Found during the audit for #3385 / PR #3395. `CopyFile`/`copyDir` stream an arbitrary `sourcePath` checking only the much narrower `isDeniedSystemPath`, so `CopyFile(/etc/shadow → /tmp/x)` followed by `ReadFile(/tmp/x)` defeats the sensitive-path deny-list outright. The same missing-call-site class applies to `RenameFile`, `DeleteFile`→trash, `TrashRestore`, `AnalyzeFilesystem`, `ScanSensitiveData`, and `EncryptFile`, which all bypass `enforceReadContainment`.

This is a distinct defect class from #3385 (which was a matching bug) — here the check is simply never called on these operations' source paths.

## Proposed Fix
Route every operation that reads or exfiltrates file content through `enforceReadContainment` on its source path(s). Per-operation nuance: destructive-but-not-disclosing ops (rename, trash) may warrant a different policy than content-disclosing ops (copy, scan, encrypt) — decide per call site and document each in the PR.

## Affected Files
- `agent/internal/remote/tools/fileops.go` (and siblings in the same package)

## Reported By
Audit pass on PR #3395 (issue #3385); deliberately kept out of that PR to avoid scope-widening.

ToddHebebrand:
Closes #3397

## The bug

`CopyFile`/`copyDir` streamed an arbitrary `sourcePath` while checking only the much narrower `isDeniedSystemPath`. `CopyFile(/etc/shadow → /tmp/x)` followed by `ReadFile(/tmp/x)` therefore defeated the sensitive-path deny-list outright — the deny-list only ever guarded `ReadFile` and `ListFiles`.

This is a **missing-call-site** class, distinct from #3385 (a matching bug). The check was simply never invoked on these operations' source paths.

## Per-operation policy

The organising principle: **relocating content out from under the deny-list is disclosure, not merely destruction.** A rename or a trash-move is a complete bypass — it never "reads" anything, yet the bytes end up at a path the deny-list no longer recognises, where a plain `ReadFile` serves them.

| Operation | Source | Destination | Rationale |
|---|---|---|---|
| `CopyFile` / `copyDir` | denied | denied | Content-disclosing. The original report. |
| `RenameFile` | denied | denied | Cheapest laundering primitive of the set: `mv /etc/shadow /tmp/x` + read. |
| `DeleteFile` → trash | denied | — | Relocates content to `~/.breeze-trash/<id>/content`, which `TrashList` enumerates and `ReadFile` serves. |
| `DeleteFile` permanent | denied | — | **Destructive-but-not-disclosing, gated on purpose.** Irrecoverably destroying a credential store has no legitimate file-browser use, and leaving the *more* destructive branch open while blocking the recoverable one would be an incoherent policy. |
| `TrashRestore` | — | denied | Arbitrary-destination write. The destination comes from `metadata.json`, a file on disk a preceding `WriteFile` could forge → credential-implant primitive. |
| `AnalyzeFilesystem` | denied (3 entry points) | — | `path`, `targetDirectories` **and** `checkpoint.pendingDirs` all seed the traversal; gating only `path` leaves two open doors. |
| `ScanSensitiveData` | denied + per-entry filter | — | Reads and pattern-matches file **content**, so a benign root containing a credential store turns the scan into an oracle over it. |
| `EncryptFile` | denied | — | Reads the whole file, then secure-wipes the source. Both halves are disqualifying. |

### Four more found by a package sweep (not in the issue)

| Operation | Gate | Why it matters |
|---|---|---|
| `QuarantineFile` | source | An `os.Rename` into a **caller-chosen** directory — same primitive as `RenameFile`, reached through the remediation surface. Had no checks at all, not even `isDeniedSystemPath`. |
| `SecureDeleteFile` | source | Arbitrary path, zero checks. Same reasoning as permanent delete. |
| `WriteFile` | destination | The direct write primitive. Gating `CopyFile`'s and `RenameFile`'s destinations while leaving this open would be no gate at all — it is the shortest path to an `authorized_keys` entry or a `/etc/sudoers.d` drop-in. |
| `handleSecurityThreat{Quarantine,Remove,Restore}` | source + dest | The same three primitives reached through the **threat** surface in `internal/heartbeat`. `handleSecurityThreatRestore` was the worst of the set: a bare `os.Rename(source, originalPath)` with both endpoints caller-supplied and no checks whatsoever. |

## Design notes

**`copyDir` now takes an explicit `skipSensitive bool`, and the polarity matters.** Gating only the copy *root* is insufficient: copying a benign parent (`/Users/alice` → `/tmp/x`) relocates every credential store beneath it. But `copyDir` is *also* the fallback half of a MOVE (`DeleteFile`→trash, `TrashRestore`), and those call sites delete the source afterwards — silently omitting an entry there would **destroy** it rather than protect it. So `CopyFile` passes `true`; the move paths pass `false` and gate their own source up front. `TestCopyDirSkipSensitiveContract` pins both directions, because getting this backwards causes data loss rather than a leak.

**`AnalyzeFilesystem` deliberately does *not* filter per entry.** It is metadata-only (paths and sizes; duplicate grouping keys on `size|basename`, never on content), so a deny-list call on the hot path of a multi-million-entry fleet scan would only suppress filenames that `ListFiles` on the parent directory already discloses. `ScanSensitiveData` *does* filter per entry — it opens and pattern-matches content, and the cost rides along with two path predicates already evaluated there. Denials in both are surfaced as scan errors rather than dropped silently.

**`enforcePathContainment` → exported `EnforcePathContainment(verb, path)`**, so the heartbeat security handlers can reuse it without `internal/security` taking a dependency on the tools deny-list. `verb` names the operation in the error so denials stay attributable in logs. `enforceReadContainment` keeps its exact previous error strings, so existing tests and #3395's assertions are untouched.

**Known gap, documented in code: `filedrop.SendFile`.** It is a second, independent content-egress path — it streams a whole file over the WebRTC data channel without passing through `tools.ReadFile`. It **cannot** call the gate: that closes an import cycle (`tools → desktop → filedrop → tools`). It has no caller anywhere in the repo today, which is the only reason this PR leaves it alone. It now carries a prominent comment requiring anyone who gives it a caller to first lift the deny-list into a leaf package. Extracting that package *now* would have conflicted hard with #3395, which edits `isSensitiveReadPath` in place.

## Found by review of this PR (three more bypasses, all reproduced first)

The security pass on this PR found three bypasses **in the fix itself**. Each was reproduced before being fixed and now has a regression test.

**1. Symlinked parent on a not-yet-existing write destination.** `filepath.EvalSymlinks` requires every component *including the leaf* to exist — which is never true for a write destination — so the symlink leg silently no-opped on exactly the case `enforceWriteContainment` exists for. With `/tmp/innocent → ~/.ssh`, `WriteFile("/tmp/innocent/authorized_keys")` passed both legs (the literal string carries no `/.ssh/`; `EvalSymlinks` errored on the missing leaf) and **implanted the key**. `resolveForContainment` now walks up to the nearest resolvable ancestor and re-attaches the remainder, which also covers destinations several levels below the link.

**2. Relocating a directory that *holds* a credential store but is not itself deny-listed.** The SSH rule keys on `/.ssh/`, so `~/.ssh` does not match while `~/.ssh/id_rsa` does. A move carries the subtree to a new prefix in one syscall where nothing matches: `DeleteFile("~/.ssh")` to trash left the private key readable at `<trash>/<id>/content/id_rsa`, and `RenameFile("~/.ssh", "/tmp/x")` left it at `/tmp/x/id_rsa` — **defeating the exact bypass this PR set out to close.** `enforceTreeContainment` now clears the subtree ahead of the two relocating operations, bounded and failing closed.

`CopyFile` never needed this: `copyDir` evaluates every entry against its *original* path, which still carries the telltale component. Permanent delete does not need it either — destroying a directory discloses nothing, and the top-level gate still refuses a directory that is itself deny-listed.

**3. `QuarantineFile`'s caller-supplied `quarantineDir` was never write-gated** — quarantining an *ordinary* file into `/etc/sudoers.d` implanted its content there. Now gated at both the tools and heartbeat entry points.

Also from the review: `AnalyzeFilesystem` now reports `partial=true` with a reason when a requested directory is refused (an all-denied incremental scan previously returned `partial=false` with zero results, which reads as a clean scan rather than "we were not allowed to look"), and the **hard-link limitation** of a path-string deny-list is documented on `EnforcePathContainment` — a hard link has no resolved path for `EvalSymlinks` to follow and `os.Lstat` reports it as an ordinary file. Not reachable through this toolset (nothing here creates links), but the symlink handling should not be mistaken for complete link coverage.

### And one more: the first fix was itself bypassable

The re-review of the fix commit found that `resolveForContainment` capped its ancestor walk at 128 iterations. The caller controls the destination string, so padding the path past the cap (`<symlink>/a/a/a/…/authorized_keys`) exhausted the budget before the resolvable ancestor was reached; the function reported "could not resolve", `EnforcePathContainment` read that as "nothing to check", and the write landed in the real `~/.ssh`. **Reproduced at depth 135.** Raising the constant would not have fixed it either — the loop called `EvalSymlinks` at every level, which is quadratic in path depth, trading a containment bypass for a CPU-exhaustion knob on the same attacker-supplied string.

The ascent now probes with `os.Lstat` (one syscall per level) and calls `EvalSymlinks` exactly once, on the ancestor that exists — linear, with no cap at all. Termination is structural rather than budgeted: each step strictly shortens the path and the loop exits at the root. The nested regression case pads to depth 300, so reintroducing any fixed cap of a plausible size fails the test (verified by putting the 128 cap back and watching it go red).

**Worth stating plainly:** an attacker-controlled bound in a security check is a bypass, not a safety valve. Both of the first two attempts here reached for a constant.

## Known limitation, deliberately kept

`enforceTreeContainment` fails closed on **any** `filepath.Walk` error, not just a deny-list hit — an unreadable subdirectory makes the whole directory rename or trash-delete fail with `"cannot verify directory contents"`. As root/LocalSystem that is rare, but it can fire on OneDrive placeholder files, EFS-encrypted files, exclusively-locked files, and macOS TCC/SIP paths. I kept it: relocating a subtree that could not be scanned is precisely the hole being closed, and the error text is distinguishable from a real containment denial. Flagging it as a plausible support ticket rather than a silent surprise.

**Should deleting a credential file be blocked outright?** This PR gates `DeleteFile` (both branches) and `SecureDeleteFile` on the deny-list, so an incident responder can no longer remove a leaked SSH key, a stale browser `Login Data`, or a rogue `/etc/sudoers.d` drop-in through the file browser — not even to trash. The review flagged this as the classic shape of over-blocking that ends with someone turning the gate off.

I kept the denial because leaving the *unrecoverable* branch open while blocking the recoverable one is incoherent, and because the alternative (an audited `force` override) is a product decision, not a code one. Happy to add the override in a follow-up if you'd rather not ship the hard denial.

## Overlap with #3395 — merge order

**No conflict in either direction; merge order does not matter.** Verified by trial merge, not by inspection:

- #3395 edits `sensitiveReadPatterns`, `isSensitiveReadPath` and `fileops_containment_test.go`.
- This PR edits `enforceReadContainment`'s body and the *call sites*, and adds a **new** test file (`fileops_containment_ops_test.go`), specifically to stay clear of #3395's hunks.
- `git merge` of this branch with `fix/3385-sensitive-path-dir-node` auto-merges clean; the merged tree builds and both `internal/remote/tools` and `internal/heartbeat` suites pass.

The two changes compose rather than overlap: #3395 widens *what* the deny-list matches, this PR fixes *where* it is consulted. Whichever lands second inherits the other's improvement for free — #3395's `filepath.Abs` normalisation and its `/library/keychains` directory-node entry automatically apply to all thirteen newly-gated call sites.

## Tests

New table-driven suite in `agent/internal/remote/tools/fileops_containment_ops_test.go`, one row per operation:

- `TestOperationsDenySensitiveSource` — 14 ops × denied source blocked, **and** the target survives (nothing half-executed before the gate).
- `TestOperationsAllowBenignSource` — the same 14 ops on ordinary files still work. A deny-list that also blocks `notes.txt` gets turned off in the field.
- `TestOperationsDenySymlinkToSensitiveSource` — innocuously-named symlink → credential store, per op. Covers the directory-symlink laundering #3395's review flagged.
- `TestOperationsDenySensitiveDestination` + `TestTrashRestoreDeniesForgedSensitiveDestination` / `...AllowsBenignDestination` — the write half, including the forged-`metadata.json` implant.
- `TestCopyDirSkipSensitiveContract`, `TestCopyFileDoesNotLaunderNestedCredentialStores`, `TestScanSensitiveDataSkipsCredentialStoresUnderBenignRoot`.

Plus regressions for all three review findings: `TestWriteDestinationDeniesSymlinkedParent` (including a destination nested below the link), `TestRelocatingADirectoryDeniesNestedCredentialStores` / `...ABenignDirectoryStillWorks`, `TestQuarantineDeniesSensitiveDestination`, `TestAnalyzeFilesystemMarksContainmentDenialsPartial`.

Fixtures build `<tmpdir>/etc/shadow` etc. — the deny-list matches path *fragments* at a component boundary, so these trip exactly the rules a real `/etc/shadow` would, with no privileged fixtures and no OS gating.

**Mutation-checked, and it caught a real defect in my own tests.** Neutering `EnforcePathContainment` to `return nil` fails all 30 subtests. An earlier draft of the scan test passed vacuously — both fixture files were being dropped by the extension filter, so nothing was ever scanned. It now carries a byte-identical benign twin as a positive control and asserts `filesScanned == 1`, so the test fails if the scanner ever stops matching. (A partial mutation also surfaced that macOS resolves `/var` → `/private/var`, which fires the symlink leg for nearly every temp path — worth knowing when reading these assertions.)

A second host-dependence bit me the same way and is worth calling out for anyone writing tests against this scanner: `defaultSensitiveExcludePaths()` contains `/tmp`, which is exactly where `t.TempDir()` lives on Linux but *not* on macOS. The scan rows were being dropped by the exclude filter before containment ran — green locally, red on Linux CI. They now clear `excludePaths` explicitly so containment is the only thing that can filter the fixture.

**Verification:** `go test -race ./internal/remote/tools/... ./internal/heartbeat/... ./internal/remote/filedrop/...` green · `go build` + `go vet ./...` clean · cross-builds green for windows, linux and darwin on both amd64 and arm64 · test binaries compile for all three GOOS · `gofmt` clean.

### DIFF
--- a/agent/internal/heartbeat/handlers_security.go
+++ b/agent/internal/heartbeat/handlers_security.go
@@ -95,11 +95,20 @@ func handleSecurityThreatQuarantine(_ *Heartbeat, cmd Command) tools.CommandResu
 	// tools.QuarantineFile, just reached through the threat surface instead of
 	// the file browser. Gated here rather than inside internal/security so that
 	// package keeps no dependency on the tools deny-list.
-	if err := tools.EnforcePathContainment("quarantine", filepath.Clean(path)); err != nil {
+	// Check and use the same string, as every other call site in #3397 does.
+	path = filepath.Clean(path)
+	if err := tools.EnforcePathContainment("quarantine", path); err != nil {
 		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
 	}
 
 	quarantineDir := tools.GetPayloadString(cmd.Payload, "quarantineDir", security.DefaultQuarantineDir())
+
+	// quarantineDir is caller-supplied: gate the destination too, or quarantine
+	// implants the file's content wherever the caller points it.
+	if err := tools.EnforcePathContainment("write", filepath.Clean(quarantineDir)); err != nil {
+		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+
 	dest, err := security.QuarantineThreat(security.Threat{
 		Name:     tools.GetPayloadString(cmd.Payload, "name", ""),
 		Type:     tools.GetPayloadString(cmd.Payload, "threatType", "malware"),
@@ -126,7 +135,9 @@ func handleSecurityThreatRemove(_ *Heartbeat, cmd Command) tools.CommandResult {
 	// Containment (#3397): destructive-but-not-disclosing, gated for the same
 	// reason as tools.SecureDeleteFile — an unrecoverable delete of a credential
 	// store is not a threat-remediation outcome anyone wants.
-	if err := tools.EnforcePathContainment("delete", filepath.Clean(path)); err != nil {
+	// Check and use the same string, as every other call site in #3397 does.
+	path = filepath.Clean(path)
+	if err := tools.EnforcePathContainment("delete", path); err != nil {
 		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
 	}
 
--- a/agent/internal/heartbeat/handlers_security.go
+++ b/agent/internal/heartbeat/handlers_security.go
@@ -90,6 +90,15 @@ func handleSecurityThreatQuarantine(_ *Heartbeat, cmd Command) tools.CommandResu
 		errResult.DurationMs = time.Since(start).Milliseconds()
 		return *errResult
 	}
+	// Containment (#3397): threat quarantine is an os.Rename into a
+	// caller-chosen directory — the same laundering primitive as
+	// tools.QuarantineFile, just reached through the threat surface instead of
+	// the file browser. Gated here rather than inside internal/security so that
+	// package keeps no dependency on the tools deny-list.
+	if err := tools.EnforcePathContainment("quarantine", filepath.Clean(path)); err != nil {
+		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+
 	quarantineDir := tools.GetPayloadString(cmd.Payload, "quarantineDir", security.DefaultQuarantineDir())
 	dest, err := security.QuarantineThreat(security.Threat{
 		Name:     tools.GetPayloadString(cmd.Payload, "name", ""),
@@ -114,6 +123,13 @@ func handleSecurityThreatRemove(_ *Heartbeat, cmd Command) tools.CommandResult {
 		errResult.DurationMs = time.Since(start).Milliseconds()
 		return *errResult
 	}
+	// Containment (#3397): destructive-but-not-disclosing, gated for the same
+	// reason as tools.SecureDeleteFile — an unrecoverable delete of a credential
+	// store is not a threat-remediation outcome anyone wants.
+	if err := tools.EnforcePathContainment("delete", filepath.Clean(path)); err != nil {
+		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+
 	err := security.RemoveThreat(security.Threat{
 		Name:     tools.GetPayloadString(cmd.Payload, "name", ""),
 		Type:     tools.GetPayloadString(cmd.Payload, "threatType", "malware"),
@@ -141,10 +157,24 @@ func handleSecurityThreatRestore(_ *Heartbeat, cmd Command) tools.CommandResult 
 		errResult.DurationMs = time.Since(start).Milliseconds()
 		return *errResult
 	}
-	if err := os.MkdirAll(filepath.Dir(originalPath), 0755); err != nil {
+	// Containment (#3397): both endpoints are caller-supplied, making restore an
+	// arbitrary source→destination move. The source is read-gated (it relocates
+	// content to an operator-chosen path) and the destination is write-gated
+	// (otherwise this is a credential-implant primitive) — the same policy
+	// tools.TrashRestore applies.
+	cleanSource := filepath.Clean(source)
+	cleanOriginal := filepath.Clean(originalPath)
+	if err := tools.EnforcePathContainment("restore", cleanSource); err != nil {
+		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+	if err := tools.EnforcePathContainment("write", cleanOriginal); err != nil {
+		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+	}
+
+	if err := os.MkdirAll(filepath.Dir(cleanOriginal), 0755); err != nil {
 		return tools.NewErrorResult(fmt.Errorf("failed to create restore directory: %w", err), time.Since(start).Milliseconds())
 	}
-	if err := os.Rename(source, originalPath); err != nil {
+	if err := os.Rename(cleanSource, cleanOriginal); err != nil {
 		return tools.NewErrorResult(fmt.Errorf("failed to restore file: %w", err), time.Since(start).Milliseconds())
 	}
 	return tools.NewSuccessResult(map[string]any{
--- a/agent/internal/remote/filedrop/handler.go
+++ b/agent/internal/remote/filedrop/handler.go
@@ -91,6 +91,17 @@ func (h *FileDropHandler) SendFile(path string) error {
 	if h.dc == nil {
 		return errors.New("filedrop: data channel not configured")
 	}
+
+	// CONTAINMENT GAP (#3397) — READ THIS BEFORE WIRING A CALLER.
+	// SendFile is a second, independent file-content egress path: it streams the
+	// whole file over the WebRTC data channel without passing through
+	// tools.ReadFile, so the sensitive-path deny-list
+	// (tools.EnforcePathContainment) does NOT apply. It is unreachable today —
+	// nothing in the repo calls it — which is the only reason #3397 left it
+	// alone. It cannot call tools.EnforcePathContainment directly: that would
+	// close an import cycle (tools → desktop → filedrop → tools).
+	// Any change that gives SendFile a caller must FIRST lift the deny-list into
+	// a leaf package both sides can import, then gate `path` here.
 	file, err := os.Open(path)
 	if err != nil {
 		return err
--- a/agent/internal/remote/tools/fileops.go
+++ b/agent/internal/remote/tools/fileops.go
@@ -206,16 +206,33 @@ func EnforcePathContainment(verb, cleanPath string) error {
 // the literal string carries no "/.ssh/" to match, and EvalSymlinks errored on
 // the missing leaf so the symlink leg never ran — and implanted the key.
 //
-// So walk up to the nearest ancestor that DOES resolve and re-attach the
-// unresolved remainder. That also covers destinations several levels below a
-// symlink, which WriteFile/RenameFile/CopyFile happily MkdirAll into.
+// So walk up to the nearest ancestor that EXISTS and re-attach the unresolved
+// remainder. That also covers destinations several levels below a symlink,
+// which WriteFile/RenameFile/CopyFile happily MkdirAll into.
+//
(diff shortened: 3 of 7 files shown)


ToddHebebrand:
**Review run:** two independent passes on the original commit (a security-focused pass and `pr-review-toolkit:code-reviewer`), then a third focused pass on the fix commit itself — warranted because the fixes touched the security gate.

**Findings:** 6 raised, all addressed.

Round 1 (2 reviewers, on `6edf24ff`) — **4 confirmed, 3 of them live bypasses**, each reproduced before being fixed:
- **Symlinked parent on a not-yet-existing write destination.** `EvalSymlinks` needs the leaf to exist, so the symlink leg silently no-opped on every write destination. `WriteFile("/tmp/innocent/authorized_keys")` with `/tmp/innocent → ~/.ssh` implanted the key.
- **Relocating a directory that holds a credential store but is not itself deny-listed.** `DeleteFile("~/.ssh")` to trash left the private key readable at `<trash>/<id>/content/id_rsa`; `RenameFile` left it at `/tmp/x/id_rsa` — defeating the exact bypass this PR set out to close.
- **`QuarantineFile`'s caller-supplied `quarantineDir` was never write-gated** — quarantining an ordinary file into `/etc/sudoers.d` implanted its content there.
- `AnalyzeFilesystem` reported an all-denied scan as `partial=false` with zero results, reading as a clean scan. Now sets `partial` + `reason`.

Round 2 (on the fix commit `f669b348`) — **1 critical: the fix was itself bypassable.** `resolveForContainment` capped its ancestor walk at 128 iterations, and the caller controls the destination string, so padding the path past the cap defeated the check and the write landed in the real `~/.ssh`. Reproduced at depth 135. Raising the constant would have traded the bypass for a CPU-exhaustion knob (the loop was quadratic in path depth). Replaced with an `os.Lstat` ascent plus a single `EvalSymlinks` — linear, no cap, termination structural rather than budgeted.

Also raised and consciously kept, not fixed: `enforceTreeContainment` fails closed on any walk error (an unreadable subdir blocks the whole move), and delete-of-a-credential-file is denied outright, which blocks a legitimate IR action. Both are called out in the PR body; the second is a product decision I'd rather you make than I assume.

**Tests:** every finding has a regression. The suite is mutation-checked, not just green — neutering `EnforcePathContainment` fails all 30 subtests, and reintroducing the 128-iteration cap fails the nested-symlink case. Two vacuous tests were caught and fixed this way (an extension filter was silently dropping the scan fixtures; `/tmp` is in `defaultSensitiveExcludePaths()`, so those rows were green on macOS and red on Linux CI).

**Verification:** `Test Agent`, `Test Agent (Windows)`, `Test Agent (race)` and `Lint Agent (Go)` all green on CI · `go vet ./...` clean · cross-builds green for windows/linux/darwin on amd64 and arm64 · full `internal/remote/tools` suite additionally run against a Linux container to catch host-dependent behaviour.

**Merge order vs #3395:** independent — verified by trial merge, not inspection. Auto-merges clean in either direction; the merged tree builds and both suites pass. #3395 widens *what* the deny-list matches, this PR fixes *where* it is consulted, so whichever lands second inherits the other's improvement.

Issue #3201, opened 2026-08-07
[API] Warranty sync does not exclude virtual machines — burns vendor API quota on synthetic serials


ToddHebebrand:
## Description

The warranty sync sweep does not exclude virtual machines. Every VMware/Hyper-V guest with a synthetic serial and a vendor-ish manufacturer string gets submitted to the vendor warranty API, burns quota, and lands permanently in `unknown`.

## Root Cause

`getDevicesNeedingWarrantySync` (`apps/api/src/services/warrantySync.ts`) filters on `is_ephemeral = false` plus non-null serial and manufacturer. `devices.is_virtual` exists but is not consulted.

`cleanHardwareIdentityValue` correctly strips OEM placeholders (`"System Manufacturer"`, `"To Be Filled By O.E.M."`) so whiteboxes yield NULL and are skipped — VMs are not caught by that path because their reported values look plausible.

## Secondary finding: no rate limiting on HP/Lenovo

Dell batches up to 100 service tags per call. The HP and Lenovo providers are per-serial loops with no throttle or backoff, so a 50-device batch is 50 sequential unthrottled calls.

## Proposed Fix

1. Add `eq(devices.isVirtual, false)` to `getDevicesNeedingWarrantySync`, and an entry guard in `syncWarrantyForDevice` alongside the existing `isEphemeral` check.
2. Add throttling/backoff to the per-serial providers.
3. Add provider unit tests — `warrantyProviders/*.ts` currently has no direct coverage (only `warrantySync.test.ts` and the alert evaluator are tested).
4. Document `DELL_CLIENT_ID` / `DELL_CLIENT_SECRET`, `LENOVO_API_KEY`, `HP_WARRANTY_ENABLED` in `.env.example` and the docs page — none currently appear anywhere, so self-hosters have no documented way to enable warranty lookup.

## Affected Files

- `apps/api/src/services/warrantySync.ts` (primary)
- `apps/api/src/services/warrantyProviders/hpProvider.ts`, `lenovoProvider.ts` (throttling)
- `.env.example`, `apps/docs/src/content/docs/features/warranty-tracking.mdx`

bdunncompany:
Partial status, since this issue has no note recording what shipped.

[...]

PR #3399, opened 2026-08-10
fix(api): exclude virtual machines from warranty sync (#3201)

bdunncompany:
Addresses the primary defect in #3201 — VMs being submitted to vendor warranty APIs. Verified the root cause against `origin/main` before changing anything.

## Confirmed

`getDevicesNeedingWarrantySync` filters `eq(devices.isEphemeral, false)` plus non-null serial and manufacturer, and never consults `devices.isVirtual`. A VMware/Hyper-V guest reports a synthetic serial and a vendor-ish manufacturer string, so it clears every existing filter, gets submitted to a vendor API that has never heard of it, burns quota, and parks in `unknown` forever.

`cleanHardwareIdentityValue` doesn't rescue this the way it rescues whiteboxes: it strips OEM placeholders like `"To Be Filled By O.E.M."` so those yield NULL and drop out, but a guest's reported values look plausible and survive.

## Changes

Both halves, mirroring exactly how `isEphemeral` is already handled:

- **Fleet sweep** — `eq(devices.isVirtual, false)` added to `getDevicesNeedingWarrantySync`.
- **Entry guard** — `isVirtual` added to the device select in `syncWarrantyForDevice`, with a `if (device.isVirtual) return;` guard beside the ephemeral one, so any *other* caller is covered too. That mirrors the comment already on the ephemeral check describing itself as "the entry-point guard for any other caller".

`devices.isVirtual` is `boolean().notNull().default(false)`, same as `isEphemeral`, so `eq(..., false)` has no NULL hole — nothing with unknown virtualization silently drops out of the sweep.

## Verification

**The tested guarantee is behavioural: no vendor provider is ever consulted for a VM.** New `warrantySyncVirtualExclusion.test.ts`, 4 cases — VM excluded, physical device still consults its provider, ephemeral still excluded, and both-flags-set excluded.

**Control run:** with only `if (device.isVirtual) return;` removed, exactly the VM case fails. The other 3 pass in both states, as they should — they pin that the new guard didn't over-reach into the physical or ephemeral paths.

Worth being explicit about coverage: the sweep's `where` clause is not directly asserted (that needs real Postgres or dialect compilation). It doesn't need to be for the quota guarantee to hold — the entry guard is downstream of the sweep, so even if the sweep filter regressed, a VM reaching `syncWarrantyForDevice` still returns before `getProviderForManufacturer`. The sweep filter is the optimisation that avoids loading the rows; the guard is what prevents the vendor call.

- `vitest run` (apps/api, full suite) — **1316 files / 21244 passed**, 37 skipped, exit 0
- Trivy `fs --scanners secret,misconfig --severity HIGH,CRITICAL` on `apps/api/src/services` — clean, exit 0
- No schema, migration, or dependency changes

One file changed, +13/−1, plus the new test file.

## Deliberately out of scope

The issue's secondary items are real but separable, and folding them in would widen the blast radius past a quota fix:

- **HP/Lenovo throttling** — those providers are per-serial loops with no backoff, so a 50-device batch is 50 sequential unthrottled calls. This PR reduces how often that happens; it does not fix the shape.
- **Provider unit tests** — `warrantyProviders/*.ts` still has no direct coverage.
- **Documenting `DELL_CLIENT_ID` / `LENOVO_API_KEY` and friends** — which, per #3239/#3388, now has to be done alongside compose wiring or an allow-list entry.

Happy to take any of those as follow-ups.

---

## Update — review round 1

The must-fix (manual refresh silently dead-ending on devices flagged virtual) is addressed in `4c434f31`:

- `POST /devices/:id/warranty/refresh` now queues with `force: true`; `force` bypasses the **`isVirtual` skip only** (an efficiency rule, and the escape hatch when virtualization detection misfires). It deliberately does **not** bypass `isEphemeral` (an ownership rule — never ship a borrowed machine's serial to a vendor), and a test pins that.
- Forced jobs get a distinct BullMQ `jobId` (`warranty-single-<deviceId>-force`) so a forced refresh cannot collapse into an already-pending unforced background job and lose its flag.
- Tests were consolidated into `warrantySync.test.ts` (the standalone `warrantySyncVirtualExclusion.test.ts` referenced above no longer exists), and `9e113df1` adds a route test pinning that the refresh endpoint actually passes `force: true` — without it, dropping the flag would silently reinstate the dead-end with all tests green.

**Behavioral note:** warranty rows already synced for VMs before this change stay frozen at `unknown` — the sweep no longer touches them. A manual (forced) refresh is the way to move an individual one, e.g. after correcting a misdetected physical machine.


ToddHebebrand (changes requested):
**Detection is right.** It reuses the canonical `devices.is_virtual` column (`apps/api/src/db/schema/devices.ts:69`, migration `2026-06-27-device-virtualization-attribute.sql:20`) populated by the agent's `ClassifyVirtualization` (`agent/internal/collectors/classify.go:122-141`) rather than inventing new inline string matching — case-insensitive, already unit-tested, and it handles the Hyper-V/Surface trap by needling the model rather than the manufacturer. The column is `notNull().default(false)`, so an unclassified device is treated as physical, which is the safe direction. The exclusion is applied at both the sweep and the entry point, so the vendor call is actually skipped rather than made-then-discarded. Diff matches the body; no schema changes, so no RLS or cascade obligations.

**Must-fix — manual refresh becomes a silent permanent dead-end for anything flagged virtual.** With the guard at `apps/api/src/services/warrantySync.ts:69`, `POST /devices/:id/warranty/refresh` still returns `"Warranty refresh queued"` (`apps/api/src/routes/devices/warranty.ts:58`), and `DeviceWarrantyCard.handleRefresh` then polls for a fresher `lastSyncAt` that will never advance (`apps/web/src/components/devices/DeviceWarrantyCard.tsx:180-205`). The tech gets a success toast, then a false "still in progress" warning, forever, on every retry.

That also removes the only escape hatch when detection misfires. `classify.go` matches bare substrings `kvm`, `xen` and `virtual machine` across manufacturer, model, BIOS and motherboard, so one bad SMBIOS string permanently costs real warranty data with no signal anywhere in the UI. Either let an explicit user-requested refresh bypass the `isVirtual` guard while keeping the sweep filter, or return an explicit "skipped — virtual machine" that the card can render.

**Smaller items:** already-synced VM warranty rows are left frozen at `unknown` — reasonable, but say so in the body. New tests should extend the existing `warrantySync.test.ts` per repo convention, and the sweep's `where` clause is currently unasserted — worth asserting the `and()` args. `routes/agents/enrollment.ts` and `inventory.ts` still enqueue BullMQ jobs for VMs that are now no-ops.

Status: requesting changes on the manual-refresh dead-end. Everything else is follow-up.

bdunncompany:
Fixed in `4c434f317`. Took your first option — an explicit user-requested refresh bypasses the `isVirtual` skip while the sweep filter stays — because it also restores the escape hatch, which the "skipped, virtual machine" response would not have done.

## The split that decides it

`force` bypasses **`isVirtual` only, never `isEphemeral`**. The two guards read alike but are different kinds of rule:

- `isVirtual` is an **efficiency** rule. It exists so the 7-day fleet sweep doesn't burn quota on synthetic serials. One human clicking refresh on one device is negligible against that, and it is the only recourse when `ClassifyVirtualization` misfires on an unlucky SMBIOS string.
- `isEphemeral` is an **ownership** rule: never ship a stranger's borrowed machine's serial to a third-party vendor. No user gesture should be able to override that, so `force` doesn't.

Both are now pinned by tests, including one asserting `force: true` still does **not** sync an ephemeral device.

## A trap in the plumbing worth flagging

`queueWarrantySyncForDevice` used a stable `jobId` (`warranty-single-<deviceId>`) so duplicate enqueues collapse — which is right, but it would have silently defeated this fix. `routes/agents/inventory.ts:84` and `enrollment.ts:1133` enqueue the unforced job on the same id; if one is already pending, BullMQ drops the newer add, and the user's forced refresh would vanish into the existing background job and run **without** `force`. Intermittently, and only on recently-enrolled or just-inventoried devices — the worst kind to reproduce.

Forced jobs therefore get their own id (`warranty-single-<deviceId>-force`), so they can never collapse into a background one. Dedup among repeated clicks still works.

## Your smaller items

- **Tests moved into `warrantySync.test.ts`** per convention; the standalone file is deleted. Required adding a `select` mock and a provider spy to that file's existing `db` mock, which the upsert-only tests didn't need.
- **Already-synced VM rows stay frozen at `unknown`** — now stated in the PR body rather than left implicit. A forced refresh is the way to move one.
- **`enrollment.ts` / `inventory.ts` still enqueue for VMs.** Left as-is deliberately: those are unforced, so they hit the entry guard and return before any vendor call. The cost is an empty job, not quota, and making them conditional would mean duplicating the virtualization check at two more call sites — the guard is the single place I'd rather keep it.
- **The sweep's `and()` args are still unasserted.** I left this one: asserting it properly means compiling through the real `PgDialect` (as you did in #3390) rather than matching on mock-call shape, which is a bigger change than it looks and doesn't gate the quota guarantee, since the entry guard is downstream. Happy to add it if you'd rather have it here than as a follow-up.

## Verification

**Two independent control runs**, since this change has two halves:

- Guard removed entirely → only `never consults a vendor provider for a virtual machine` fails (1 of 9).
- Guard made unconditional (`force` ignored) → only `force: an explicit user refresh DOES sync a virtual machine` fails (1 of 9).

Neither half is carried by the other, and the remaining 7 pass in every state.

`tsc --noEmit` exit 0. Full API suite **1315 files / 21245 passed**, 37 skipped. Trivy `secret,misconfig` HIGH/CRITICAL on `apps/api/src` clean. No schema, migration or dependency changes.

PR #3411, opened 2026-08-11
feat(abuse): provider-default hostname detector, and make abuse detection hosted-only

ToddHebebrand:
## Why

A managed MSP endpoint gets named or domain-joined. A machine still carrying the hostname its hosting provider or OS installer stamped on it is almost always the operator's own staging VM rather than a customer asset.

This is the same argument `rmm.consumer_devices` already makes about `DESKTOP-`/`LAPTOP-` defaults, pointed at the other end of the fleet: those defaults mean *victim*, these mean *operator*.

Across four months and three unrelated abuse rings, every device in US prod carrying a provider-default hostname has belonged to a partner subsequently suspended for abuse.

## What

Adds `rmm.provider_default_hostname` to the heuristics pass, in two tiers with deliberately different confidence.

**Curated** — hostname starts with a prefix listed in the new `ABUSE_HOSTNAME_INDICATORS` env var. A reviewed attribution, so it is alert-capable on a *single* device with no ratio gate: an operator mixing customer machines in alongside its staging boxes must not be able to dilute the score by enrolling more of them.

The list is **empty in this repo by design.** Publishing the prefix would tell the operator exactly which string to rename — which is precisely how the installer detector was recently evaded, by an unbranded `.msi` on object storage. Same mechanism and rationale as the existing `ABUSE_SCRIPT_INDICATORS`.

**Generic** — hostname merely has the *shape* of a provider default:

- `^[a-z]{2,4}-\d{4,8}$` — short alpha tag plus a numeric serial, the stock hostname on a VPS image
- `^win-[a-z0-9]{11}$` — Windows Server's stock computer name

Needs no configuration, so self-hosted installs benefit too. But a small shop really does sometimes name a machine `PC-0042`, so this tier is ratio-gated (≥3 devices, ≥50% matching) and **caps below `severity.alert_score`** — the same corroborate-don't-accuse rule as `session_intensity` and `cardholder_name_mismatch`.

Neither tier is age-decayed: this describes what the fleet *is*, not how recently the account was created.

## Design notes

- **`DESKTOP-`/`LAPTOP-` are deliberately excluded here.** `rmm.consumer_devices` already scores them, and classifying them in both places would let one fleet stack two independent-looking signals off a single hostname.
- **Matching runs in TS, not SQL.** `loadPartnerAggregates` builds its query with `sql.raw`, so interpolating an operator-supplied prefix would be an injection sink. The new `hosts` CTE only returns bounded raw hostnames, mirroring the existing `ips` CTE (same 5000/partner bound); `classifyHostname` is pure and unit-tested.
- Curated outranks generic, so a device is counted once, at one tier only.
- Evidence keeps hostnames exactly as stored — triage needs a string it can search on, not a normalized one.

## Validation

Run against US prod (read-only): 12 devices match the generic patterns; **11 belong to already-suspended partners.** The twelfth is our own OliveTech box at a 0.02 ratio, correctly below the gate.

Worth flagging for whoever deploys this: the generic tier alone would **not** have caught the incident that prompted the work. That operator ran one staging box among four victim machines — a 0.20 ratio, well under the gate. Catching that case needs the curated prefix, so `ABUSE_HOSTNAME_INDICATORS` must be set in each deployment's `.env` **and** mapped in the `api` service's compose `environment:` block, or this ships inert for its primary case.

## Tests

`pnpm --filter @breeze/api exec vitest run src/services/abuseSignals/` — 160/160 pass; `tsc --noEmit` clean.

New coverage: shape classification (including real-world named/domain-joined hostnames that must *not* match), the `DESKTOP-`/`LAPTOP-` exclusion, curated-over-generic precedence, env parsing and its malformed-input fallbacks, the alert/watch split between tiers, the ratio gate rejecting one stray stock name in a properly-named fleet, one-signal-per-partner, and absence of age decay.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -470,6 +470,13 @@ GRAFANA_ADMIN_PASSWORD=changeme
 # DB_POOL_HEALTH_DISABLED=
 
 # Platform-operator abuse alerts (all optional; unset = feature disabled)
+#
+# Signup-abuse detection polices UNTRUSTED PUBLIC SIGNUPS on a multi-tenant
+# hosted service. It defaults to ON only when IS_HOSTED is true, because on a
+# normal self-hosted install — one IT team, its own machines, no signup funnel —
+# most of these heuristics are noise or actively wrong. Leave it off unless you
+# are running Breeze as a service for tenants you do not control.
+# ABUSE_SIGNALS_ENABLED=true
 # OPS_ALERT_WEBHOOK_URL=https://discord.com/api/webhooks/your-webhook
 # OPS_ALERT_EMAIL=ops@your-domain.example.com
 # OPS_ALERT_LABEL=US
@@ -477,6 +484,10 @@ GRAFANA_ADMIN_PASSWORD=changeme
 # Script-content detector indicator lists (unset = empty lists; the repo
 # deliberately ships no indicators).
 # ABUSE_SCRIPT_INDICATORS={"tlds": [], "hosts": []}
+# Hostname prefixes attributed to abusive hosting, matched against device
+# hostnames (unset = empty list; the repo deliberately ships no indicators,
+# since publishing a prefix just tells an operator which string to rename).
+# ABUSE_HOSTNAME_INDICATORS={"prefixes": []}
 
 # --------------------------------------------
 # Application URLs
--- a/.env.example
+++ b/.env.example
@@ -469,13 +469,21 @@ GRAFANA_ADMIN_PASSWORD=changeme
 # DB_POOL_HEALTH_CAPTURE_THROTTLE_MS=900000
 # DB_POOL_HEALTH_DISABLED=
 
-# Platform-operator abuse alerts (all optional; unset = feature disabled)
+# Platform-operator signup-abuse detection and its alerts.
+#
+# NOT all-or-nothing on being unset: ABUSE_SIGNALS_ENABLED defaults to the value
+# of IS_HOSTED, so leaving it unset ENABLES the whole detection subsystem on a
+# hosted install and disables it on a self-hosted one. Every OTHER variable in
+# this block is optional and unset = that piece disabled.
 #
 # Signup-abuse detection polices UNTRUSTED PUBLIC SIGNUPS on a multi-tenant
-# hosted service. It defaults to ON only when IS_HOSTED is true, because on a
-# normal self-hosted install — one IT team, its own machines, no signup funnel —
-# most of these heuristics are noise or actively wrong. Leave it off unless you
-# are running Breeze as a service for tenants you do not control.
+# hosted service. On a normal self-hosted install — one IT team, its own
+# machines, no signup funnel — most of these heuristics are noise or actively
+# wrong, which is why the default follows IS_HOSTED. Set it explicitly only to
+# override that: true to opt a self-hosted multi-tenant service in, false to
+# switch a hosted deployment off. Must be a boolean (true/false, 1/0, yes/no,
+# on/off) — the API refuses to boot on anything else, so a typo cannot silently
+# turn detection off.
 # ABUSE_SIGNALS_ENABLED=true
 # OPS_ALERT_WEBHOOK_URL=https://discord.com/api/webhooks/your-webhook
 # OPS_ALERT_EMAIL=ops@your-domain.example.com
--- a/apps/api/src/__tests__/integration/abuseSignalsAggregates.integration.test.ts
+++ b/apps/api/src/__tests__/integration/abuseSignalsAggregates.integration.test.ts
@@ -46,7 +46,16 @@ function consumerHostname(index: number): string {
   return `DESKTOP-AAAA${String(index).padStart(3, '0')}`;
 }
 
-async function seedDevice(opts: { orgId: string; siteId: string; hostname: string; enrollmentIp: string; lastSeenIp?: string; enrolledAt: Date }) {
+async function seedDevice(opts: {
+  orgId: string;
+  siteId: string;
+  hostname: string;
+  enrollmentIp: string;
+  lastSeenIp?: string;
+  enrolledAt: Date;
+  /** Defaults to 'offline' — a live device. Pass 'decommissioned'/'quarantined' to exercise the loader's status filter. */
+  status?: 'offline' | 'decommissioned' | 'quarantined';
+}) {
   const testDb = getTestDb();
   const [device] = await testDb
     .insert(devices)
@@ -59,7 +68,7 @@ async function seedDevice(opts: { orgId: string; siteId: string; hostname: strin
       osVersion: '11',
       architecture: 'x86_64',
       agentVersion: '0.0.0-test',
-      status: 'offline',
+      status: opts.status ?? 'offline',
       enrollmentIp: opts.enrollmentIp,
       lastSeenIp: opts.lastSeenIp,
       enrolledAt: opts.enrolledAt,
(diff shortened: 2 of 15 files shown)


ToddHebebrand:
## Added: abuse detection is now hosted-only by default (`992e25e74`)

Second commit, addressing a scope question raised in review: this subsystem was running unconditionally on **every** install, including self-hosted ones — and it shouldn't have been.

Signup-abuse detection exists to police untrusted public signups on a multi-tenant service. A typical self-hosted install is one IT team managing its own machines, with no signup funnel and no untrusted tenants. For them the heuristics are noise or actively wrong:

- **`invariant.active_no_payment` can never clear on self-host.** `payment_method_attached_at` is only ever written by the external hosted billing service, so it misfires forever.
- **The whole `billing.*` family is dead for the same reason** — it couples an AGPL deployment to a hosted billing service it does not run. That coupling was never meant to be in the open-source product.
- `rmm.device_ip_scatter` just describes an org with remote workers.
- The fleet-shape detectors — including the one this PR adds — flag an ordinary lab of unnamed test VMs.

`abuseSignalsEnabled()` now defaults to `isHosted()`, with `ABUSE_SIGNALS_ENABLED` overriding both ways: a self-hoster running a genuine multi-tenant service can opt in, and a hosted deployment can switch it off without a redeploy.

### Two details worth a reviewer's attention

**Polarity is deliberately inverted relative to `selfHostAllowsPrivateNetwork`.** This keys off the *truthy* reading of `IS_HOSTED`, not the affirmative-self-host helper. The failure that matters here is a hosted deployment silently *not* policing its signups, so an unset or garbage `IS_HOSTED` leaves detection off and loudly absent rather than half-configured. `selfHostAllowsPrivateNetwork` fails closed toward *strictness* because there the risk runs the other way. Worth confirming that reasoning holds.

**Disabling tears down existing repeatables, it doesn't just skip scheduling.** Repeatable jobs live in Redis. An install upgrading into this gate — or flipping `IS_HOSTED` — would otherwise keep firing the hourly sweep forever from jobs a previous boot registered, with no worker left to report the result. Teardown is best-effort so a Redis blip can't take the API down over an inert job.

Both new vars (`ABUSE_SIGNALS_ENABLED`, `ABUSE_HOSTNAME_INDICATORS`) are threaded through `.env.example` **and** the `api` service block of both compose files, per the `envComposeParity` guard — a documented var nobody maps into a container is silently inert, which is how `IS_HOSTED` itself once shipped broken (#570).

### Deploy impact

**Hosted (US/EU): none** — `IS_HOSTED=true` is already set, so behaviour is unchanged.

**Self-hosted: the sweep stops.** That is the intent, and no data is destroyed — `partner_abuse_signals` rows persist and the admin routes still read them. Anyone who wants it back sets `ABUSE_SIGNALS_ENABLED=true`.

Tests: 203 pass across the detector, job, env and compose-parity suites; `tsc --noEmit` clean.

ToddHebebrand:
## Review pass applied (`8f9224f5d`)

Five specialist agents reviewed the diff — general correctness, tests, silent failures, comment accuracy, and type design. Severity-ordered results below. Three findings were independently reported by four of the five, which is how I prioritized.

### Critical

**A whitespace-only curated prefix matched every hostname.** The emptiness filter ran *before* the trim, so `{"prefixes":[" "]}` — a CRLF line ending, or a trailing blank in a prefix list — normalized to `[""]`, and `startsWith("")` is true for everything. Every device on every in-scope partner would classify as `curated`: the alert tier, no ratio gate, real customer hostnames in the evidence payload, ops paged for the entire fleet. I reproduced it against the real module before fixing. Trim now runs first; blank entries drop with a warn, and the missing-key/non-array shapes warn rather than silently yielding `[]`.

**`ABUSE_SIGNALS_ENABLED=ture` silently disabled detection on a hosted deploy.** Any non-empty value routed to `envFlag`, which maps everything outside `1|true|yes|on` to false. That is precisely the polarity failure the gate's own comment claims to prevent. Truthy and falsey vocabularies are now explicit; an unrecognized value warns and falls through to `isHosted()`, and `validate.ts` refuses boot on a near-miss, mirroring the existing `AGENT_AUTO_PROMOTE` guard.

### Important

**Teardown could disable a correctly-configured replica.** Removing repeat keys writes to *shared* Redis state based on one process's env read. A replica booting with an unmapped `IS_HOSTED` (#570's exact shape) would delete the repeatables a healthy sibling just scheduled — and that sibling keeps its worker and still logs success while its queue silently never fires again. Now gated on affirmative opt-out; ambiguity leaves shared state alone. Failures report via `console.error` + Sentry, and the handle is nulled only after `close()` resolves.

**The generic tier's ratio ramp was dead code.** `min(55, 45 * (1 + ratio))` always returned exactly 55 once the 0.5 gate cleared, so triage couldn't distinguish half a stock-named fleet from all of it. Now ramps over the excess above the gate: 0.5 → 45, 1.0 → 55.

**The ratio denominator drew from a different population than its numerator** — the 5000-bounded `hostnames` array over an uncapped `COUNT(*)`, meaning any fleet above 10k devices could never reach the gate. Now `hostnames.length` for both, matching what the adjacent `rmm.device_ip_scatter` documents.

**Removed `^[a-z]{2,4}-\d{4,8}$` from the built-in patterns.** It matches mainstream site-code naming (`PC-00123`, `SRV-0001`, `NY-10045`), and because the signal never decays the resulting watch row can never stale-resolve — it would pin a legitimate partner into the sweep scope and every weekly digest permanently. The built-in list is now restricted to stock-OS-installer artifacts (`WIN-` + 11 alphanumerics, which nobody chooses). Provider serial prefixes belong in `ABUSE_HOSTNAME_INDICATORS`, which is where the real signal lives anyway.

**`computeHeuristicSignals`' 4th parameter is now required**, matching `computeScriptSignals`. Defaulting it meant omitting the argument at a call site would compile clean and silently disable the only alert-capable tier.

### Three comments asserted things that were not true

- The teardown rationale claimed stale repeatables *"keep firing the sweep forever"*. In BullMQ 5 the **worker** enqueues the next iteration, so with no worker the chain stalls after one delayed job. Verified in `worker.js` before rewriting. What actually matters is that the repeat *key* survives and resurrects the sweep when any worker reattaches.
- `payment_method_attached_at` was described as *"only ever written by the external hosted billing service."* It has an in-repo writer, and `services/partnerActivation.ts` already retracts a near-identical claim after a production audit. Replaced with the argument that is verifiable and reaches the same conclusion: self-host runs no billing writer at all.
- *"loudly absent"* asserted a safety property nothing delivers — the only artifact is one boot log line. It now says so plainly, and the log reports the values actually read rather than asserting a cause.

### Tests

543 across the detector, job, env and config suites. Assertions that passed trivially are pinned to concrete values (the "caps the generic tier" test asserted `< 70`, which a constant 55 satisfies regardless). Both gate boundaries are covered on the firing side, the clamp saturation case is added, and `sweep.test.ts` now proves `loadHostnameIndicators()` actually reaches `computeHeuristicSignals` — previously the tier could have been disconnected with zero test failures.

The `hosts` CTE gained real-DB coverage — content equality, the decommissioned/quarantined filter, and the empty-partner `LEFT JOIN` — executed against a live database, not just typechecked.

One honest caveat carried forward: the postgres-js string-vs-array failure mode can only be caught by the *populated* assertion, because the mapper's `Array.isArray(...) ? ... : []` fallback silently converts a string to `[]`. Pinning that at the SQL boundary would mean making the mapper throw, which is out of scope here.

PR #3414, opened 2026-08-11
fix(api): dispatch policy remediation instead of simulating completion (#3413)

### LINKED ISSUE #3413: [API] Config-policy remediation never executes — automation run inserted and setTimeout-marked completed without dispatching
## Description

Configuration-policy remediation does not actually run anything on a device. `triggerConfigPolicyRemediation` (`apps/api/src/services/policyEvaluationService.ts:1787-1930`) resolves the remediation automation, inserts an `automationRuns` row, and then marks that run completed via a `setTimeout` — without ever calling `executeAutomationRun`, queueing a device command, or touching the automation runtime. The standalone-policy variant in the same file behaves identically. `resolvePolicyRemediationAutomationIdForOrg` (`:963`) is only used to look up the automation id, never to execute it.

Net effect: policy drift shows "remediated" runs in history while the device was never touched — a silent no-op that looks like success in the UI.

## Root Cause

The remediation trigger path was stubbed (insert run row + timed status flip) and never wired to the real execution path (`executeAutomationRun` / `automationWorker`).

## Proposed Fix

Route remediation through the actual automation execution path. Once the shared script-dispatch core from #3409 (PR 0) lands, remediation automations containing `run_script` / `execute_command` actions get correct dispatch, execution rows, and variable resolution for free — so the fix is: replace the setTimeout stub with a real `executeAutomationRun` invocation (or enqueue via the automation worker), and mark the run from the worker's actual outcome.

## Affected Files

- `apps/api/src/services/policyEvaluationService.ts` (both the config-policy and standalone-policy remediation triggers)

## Found During

Code-verified scoping of #3409 (see the scoping comment there for the full dispatch-path survey).

bdunncompany:
Closes #3413.

## What was wrong

Both remediation triggers inserted an `automation_runs` row, marked it `completed` on a `setTimeout(…, 1000)`, and never dispatched anything. Policy drift showed "remediated" runs in history while the device was never touched — a silent no-op that reads as success.

**Both sites, since the issue's "the standalone variant behaves identically" is exactly right:**

| Function | Run insert | Simulation |
|---|---|---|
| `triggerRemediationAutomation` (standalone policy) | `:1138` | `setTimeout` `:1172` |
| `triggerConfigPolicyRemediation` (config policy) | `:1862` | `setTimeout` `:1898` |

The second carried the comment `// Simulate completion (same pattern as standalone policy remediation)`, so this was a deliberate stub that was duplicated, not drift. Fixing only one would have left half the feature lying.

## The fix

Each simulation block becomes a real dispatch:

```ts
const { enqueueAutomationRun } = await import('../jobs/automationWorker');
await enqueueAutomationRun(run.id, [device.id]);
```

**No row-shape change was needed.** `executeAutomationRun(runId, targetDeviceIds?)` (`automationRuntime.ts:1737`) *reads* an existing run rather than creating one, so the insert both sites already do is exactly what the runtime wants. This is the same shape `automationWorker.ts:448` and `:490` use — insert the run, then enqueue it — so the policy paths now agree with every other automation trigger instead of having their own.

**Queued rather than awaited inline**, deliberately: policy evaluation runs inside a fleet sweep, and blocking it on script execution would make the sweep's duration a function of remediation work across every non-compliant device. Two properties of `enqueueAutomationRun` matter here and are why it beats calling the runtime directly:

- it falls back to `executeRunInline` via `setImmediate` when Redis is unavailable, so a self-hoster without Redis gets real execution rather than today's silent no-op;
- its stable `automation-run-<id>` job id means a re-entrant sweep cannot double-dispatch the same run.

**Dynamic import, and this one is not stylistic.** A static `import { enqueueAutomationRun } from '../jobs/automationWorker'` pulls the BullMQ worker graph into `policyEvaluationService` at module load, which reaches `jobs/discoveryWorker` → `services/networkBaseline` and broke `policyManagement_compliance_device.test.ts` with `No "discoveredAssetTypeEnum" export is defined on the "../db/schema" mock`. I could have patched that test's mock, but that would have made every future importer of this service pay for the worker graph. `drExecutionService.ts:300` already uses `await import('../jobs/drExecutionWorker')` for exactly this reason, so the dynamic form is the established pattern rather than a workaround.

## The boolean

`triggerConfigPolicyRemediation` returns `true` and `triggerRemediationAutomation` returns the run id. Both now mean **dispatched**, not finished — the run row carries the outcome once the runtime executes it. That is the honest meaning: the previous `true` claimed completion that had not happened. I raised this on the issue before writing the tests; say the word if you'd rather callers distinguish dispatched from completed and I'll follow up.

## Verification

New `policyEvaluationRemediationDispatch.test.ts`, 4 cases. The two trigger functions are exposed via `__`-prefixed aliases at the bottom of the service, matching the existing `__evaluateRulesForDevice` convention, so no internal call site was renamed.

**Control run:** with both dispatch calls removed, `config-policy remediation enqueues the run it just created` fails. Stated plainly — the other 3 are negative cases (no matching automation, no remediation script) plus an export-presence pin, and they **cannot** fail; they exist to stop the fix over-reaching into paths that should still no-op, not to demonstrate it.

- `vitest run` (apps/api, full suite) — **1322 files / 21522 passed**, 37 skipped, exit 0
- Trivy `secret,misconfig` HIGH/CRITICAL on `apps/api/src` — clean, exit 0
- No schema, migration or dependency changes

One file changed (+27/−50, the deletions being the two simulation blocks) plus the new test file.

## Not in scope

`resolvePolicyRemediationAutomationIdForOrg` (`:963`) is still only used to look up an id, as the issue notes. It is a separate resolution helper rather than a dispatch path, and nothing here changes what it does.

### DIFF
--- a/apps/api/src/services/policyEvaluationRemediationDispatch.test.ts
+++ b/apps/api/src/services/policyEvaluationRemediationDispatch.test.ts
@@ -1,0 +1,138 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// #3413: both remediation triggers inserted an automation_runs row and then
+// flipped it to 'completed' on a setTimeout without ever dispatching, so policy
+// drift showed "remediated" while the device was never touched. The observable
+// guarantee is that the real automation runtime is asked to run the row.
+
+const enqueueAutomationRunMock = vi.fn().mockResolvedValue({ enqueued: true });
+vi.mock('../jobs/automationWorker', () => ({
+  enqueueAutomationRun: (...args: unknown[]) => enqueueAutomationRunMock(...args),
+}));
+
+const selectMock = vi.fn();
+const insertMock = vi.fn();
+const updateMock = vi.fn();
+
+vi.mock('../db', () => ({
+  db: {
+    select: (...args: unknown[]) => selectMock(...args),
+    insert: (...args: unknown[]) => insertMock(...args),
+    update: (...args: unknown[]) => updateMock(...args),
+  },
+}));
+
+vi.mock('./eventBus', () => ({ publishEvent: vi.fn() }));
+vi.mock('./featureConfigResolver', () => ({
+  resolveComplianceRulesForDevice: vi.fn(),
+  scanDueComplianceChecks: vi.fn(),
+}));
+
+import {
+  __triggerConfigPolicyRemediation,
+  __triggerRemediationAutomation,
+} from './policyEvaluationService';
+
+const DEVICE = {
+  id: 'dddddddd-dddd-dddd-dddd-dddddddddddd',
+  orgId: 'oooooooo-oooo-oooo-oooo-oooooooooooo',
+  hostname: 'WS-1',
+  osType: 'windows',
+  osVersion: '10.0.19045',
+};
+const RUN_ID = 'rrrrrrrr-rrrr-rrrr-rrrr-rrrrrrrrrrrr';
+const AUTOMATION_ID = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+const SCRIPT_ID = 'ssssssss-ssss-ssss-ssss-ssssssssssss';
+
+/** Queue results for the successive `.select()` chains each trigger performs. */
+function queueSelects(...results: unknown[][]) {
+  const queue = [...results];
+  selectMock.mockImplementation(() => {
+    const chain: Record<string, unknown> = {};
+    const step = () => chain;
+    chain.from = step;
+    chain.where = step;
+    chain.innerJoin = step;
+    chain.leftJoin = step;
+    chain.limit = () => Promise.resolve(queue.shift() ?? []);
+    chain.then = (resolve: (v: unknown) => unknown) => resolve(queue.shift() ?? []);
+    return chain;
+  });
+}
+
+beforeEach(() => {
+  enqueueAutomationRunMock.mockClear();
+  selectMock.mockReset();
+  insertMock.mockReset();
+  updateMock.mockReset();
+
+  insertMock.mockReturnValue({
+    values: () => ({ returning: () => Promise.resolve([{ id: RUN_ID, logs: [] }]) }),
+  });
+  updateMock.mockReturnValue({
+    set: () => ({ where: () => Promise.resolve(undefined) }),
+  });
+});
+
+describe('policy remediation actually dispatches (#3413)', () => {
+  it('config-policy remediation enqueues the run it just created', async () => {
+    queueSelects(
+      [{ orgId: DEVICE.orgId }],                                   // device org lookup
+      [{ partnerId: null }],                                       // automationOwnershipConditionForOrg
+      [{ id: AUTOMATION_ID, actions: [{ scriptId: SCRIPT_ID }] }], // candidate automations
+      [{ id: AUTOMATION_ID, enabled: true }],                      // the automation row
+    );
+
+    const result = await __triggerConfigPolicyRemediation(
+      {
+        id: 'cccccccc-cccc-cccc-cccc-cccccccccccc',
+        name: 'BitLocker enabled',
+        remediationScriptId: SCRIPT_ID,
+        featureLinkId: 'ffffffff-ffff-ffff-ffff-ffffffffffff',
+      } as never,
+      DEVICE as never,
+    );
+
+    expect(result).toBe(true);
+    expect(enqueueAutomationRunMock).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


ToddHebebrand (approved):
**Approved.** I verified the load-bearing claims against `origin/main` rather than taking the write-up's word for them, and they hold.

**The runtime contract is what you say it is.** `executeAutomationRun` (`apps/api/src/services/automationRuntime.ts:1737`) delegates to `executeAutomationRunInner`, which selects the run by id at `:1766` and throws `Automation run not found` if it isn't there — it reads, it never creates. Worth noting explicitly because there is a second guard right after it: `:1776` throws when `run.automationId` is null. Both of your insert sites already set `automationId: automation.id`, including the config-policy one that additionally carries `configPolicyId`, so neither trips it. That was the one way "no row-shape change was needed" could have been quietly wrong, and it isn't.

**The dispatch is correctly narrowed to the offending device.** `automationRuntime.ts:1797` uses `targetDeviceIdsFromQueue` when non-empty and only falls back to `resolveAutomationTargetDeviceIds(automation)` otherwise. Passing `[device.id]` therefore means remediation touches exactly the device that evaluated non-compliant, never the automation's own — possibly partner-wide — target set. That is the single place where turning a no-op into a real dispatch could have blast-radiused, and it's the right side of the branch. Tenancy is unchanged downstream: device rows are read with their own `devices.orgId` at `:1810` and per-device result rows are seeded from them, so worker-created children still take the DEVICE's org.

**The sweep cannot die on the enqueue.** `enqueueAutomationRun` (`apps/api/src/jobs/automationWorker.ts:261`) has no throwing path: the Redis-absent branch at `:265` schedules `executeRunInline` via `setImmediate` and returns, and the queue branch is wrapped end to end with the same inline fallback in its `catch` at `:312`. So `await enqueueAutomationRun(...)` inside a fleet sweep is safe, and when execution does fail the run row tells the truth — `executeAutomationRun` calls `markAutomationRunFailedAfterError` at `:1748` before rethrowing. The job id at `:279` is `automation-run-<runId>`: stable per run, hyphens only, and it carries its own comment citing the colon rule from #1101.

**No cycle.** `automationWorker` pulls only `eventBus`, `automationRuntime`, `featureConfigResolver`, `redis`, `bullmqUtils` and `bullmqValidation` out of `services/`, none of which reach `policyEvaluationService`, and the service has no static `../jobs` import at all. The module also has no load-time side effects — the queue and worker are lazily constructed (`:213`, `:84-87`) — and `routes/automations.ts:31` already imports it statically into the API process, so the dynamic form is buying test-graph hygiene rather than papering over a real hazard.

**On the boolean: dispatched-semantics is right, and the run row is the source of truth.** Take it as settled, no follow-up needed. It is what every other automation trigger already means — `automationWorker.ts:448` and `:490` return a `runId` the instant the job is queued — and blocking a fleet sweep on remediation completion would make the sweep's duration a function of script execution across every non-compliant device, which is the wrong trade at 10k agents. The consumer field is also already named for it: `remediationTriggered` (`policyEvaluationService.ts:1754`). I checked the consumers and nothing treats the value as terminal success — `remediationRunId` only rides along into `policyAlertBridge.ts:209` as an id, and on the web side `DeviceComplianceTab.tsx:231` renders remediation as an attempt count, which is if anything more honest now than it was against a fabricated completion. No UI copy needs changing.

**Two follow-ups, neither blocking this PR.**

The first is the more interesting one, and it is out of scope here: `POST /policies/:id/remediate` (`apps/api/src/routes/policyManagement/actions.ts:242-274`) is a third site with the same disease. It resolves the automation, inserts an `automationRuns` row with `status: 'running'` and `devicesTargeted: 0`, bumps `runCount`, and returns `message: 'Remediation automation triggered'` — without ever dispatching. It doesn't fake completion the way the two you fixed did, so it isn't strictly the #3413 lie, but it is the manual "remediate now" path and it still executes nothing; the run just sits at `running` forever. The fix looks like the same one line, minus the target argument — `await enqueueAutomationRun(run.id)` — so that `resolveAutomationTargetDeviceIds` expands the policy's own targets. Please open a separate issue rather than growing this PR.

The second is test depth. `__triggerConfigPolicyRemediation` is genuinely exercised — asserting `toHaveBeenCalledWith(RUN_ID, [DEVICE.id])` cannot pass without a real dispatch, so that is a behavioral pin and not another simulation — but `__triggerRemediationAutomation` is only held by a `typeof === 'function'` export check. You say as much in the description, which I appreciate. The two bodies are identical today; a case that drives the standalone path through to the same assertion would stop them drifting apart later.

Not merging — leaving that to the repo owner.

PR #3420, opened 2026-08-11
fix(approvals): tenant the suspicious-report audit row to the approval, not the reporter (#3234)

### LINKED ISSUE #3234: [API] report-suspicious returns 500 for partner-scoped users; approval stays pending (security action fails open)
## Description

Found during v0.104.0 release QA. `POST /api/v1/approvals/:id/report-suspicious` returns **500** for any **partner-scoped** caller:

```json
{"error":"Internal Server Error","message":"new row violates row-level security policy for table \"audit_logs\""}
```

The approval row stays `pending` and no audit row is written — **a security action that fails open**. Reproduced on a local stack with a partner-scoped admin (`scope: "partner"`, `orgId: null`).

The same handler is mounted twice (`apps/api/src/index.ts:1044` and `:1056`), so **the iOS app's "report suspicious" button hits the identical path** via `/api/v1/mobile/approvals/:id/report-suspicious`.

## Root Cause

`apps/api/src/routes/approvals.ts:568` binds the audit row's org from the caller's own auth context:

```ts
const orgId = c.get('auth').orgId ?? null;
```

A partner-scoped token has `orgId: null`. The `audit_logs` INSERT policy is `WITH CHECK breeze_has_org_access(org_id)`, and `breeze_has_org_access(NULL)` returns false — so the insert is rejected.

There is **no fallback**: `approval_requests` has no `org_id` column, and although the intent-linked branch does fetch `actionIntents.orgId` via the CAS `.returning()` (line 683), that value is only used for `recordActionIntentEvent` (line 710), never for the audit insert.

The local `try/catch` around the insert (lines 789-811) logs but cannot save the request. The whole handler runs inside the ambient postgres.js transaction opened by auth middleware (`withDbAccessContext`, `db/index.ts:465`, called from `middleware/auth.ts:690`), so the aborted-transaction state is rethrown when that `begin()` resolves — past the local catch, landing as an uncaught 500. Same trap previously seen elsewhere in the repo.

**Org-scoped callers are unaffected** — their non-null `orgId` satisfies the policy.

## Not a regression

The audit-insert block is **byte-identical** before and after #3175 (`1516048c7`), and `db/index.ts` was untouched by it. This is pre-existing and did not block the v0.104.0 cut. Filing so it doesn't get lost.

## Proposed Fix

Source the audit `orgId` from the linked intent's org (already available from the CAS `.returning()`) rather than solely from `auth.orgId`, with an explicit decision on what to write for a standalone approval that has no linked intent.

## Test gap (the reason this survived)

The coverage is split such that neither half can catch it:

- `apps/api/src/routes/approvals.test.ts` — mocks `db` entirely (`vi.mock('../db')`, line 4-17), so RLS never runs. Its auth mock **does** use `scope: 'partner', orgId: null` (lines 233-241), matching the repro exactly, but the audit insert is stubbed to always succeed (lines 2686-2688).
- `apps/api/src/routes/approvalsDecideSupervised.integration.test.ts` — real Postgres as `breeze_app` with RLS enforced, but `requesterAuth()` **hardcodes `scope: 'organization'`** (lines 59-86).

No test anywhere exercises a partner-scoped caller against real RLS. A fix should add one.

## Affected Files

- `apps/api/src/routes/approvals.ts` (primary, line 568)
- `apps/api/src/routes/approvalsDecideSupervised.integration.test.ts` (coverage)

## Provenance

Found during v0.104.0 release QA (section F). Related: #3233 (same ambient-transaction mechanism, different symptom).

ToddHebebrand:
## The bug

`POST /api/v1/approvals/:id/report-suspicious` returned **500 for every partner-scoped caller**, and the approval it was meant to stop stayed `pending`:

```json
{"error":"Internal Server Error","message":"new row violates row-level security policy for table \"audit_logs\""}
```

A security action that fails open. Found during v0.104.0 release QA, still live on `main`.

## Root cause

The handler bound the audit row's org from the **caller's own token**:

```ts
const orgId = c.get('auth').orgId ?? null;   // NULL for a partner-scoped token
```

`audit_logs`' INSERT policy is `WITH CHECK breeze_has_org_access(org_id)`, and `breeze_has_org_access(NULL)` returns FALSE under any non-system scope (TRUE under `system`). So the insert was rejected.

The local `try/catch` around it could not prevent the 500, and the reason is structural rather than local. The whole request runs inside **one** Postgres transaction opened by auth middleware (`withDbAccessContext` → `baseDb.transaction`). An RLS violation aborts the **transaction**, not just the statement — so the `catch` swallowed the JS error, the handler ran on and returned 204, and then the commit threw, resurfacing the original error from a frame unrelated to the audit write. Worse, that rollback also undid the row's flip to `reported`.

`approval_requests` is Shape 6 (user-id scoped), so the caller could always read and flip its own row; `audit_logs` was the only table that rejected the write.

## The fix

**1. Resolve the audit org from the approval being reported, not the reporter.** Chain: linked intent's org → reporter's org → NULL. Same defect class as the `remediate_vulnerability` fix in `3716e414d` (#3322/#3379), where the org was resolved from the device rather than the caller's context.

The existing `SELECT ... FOR UPDATE` on `action_intents` now also selects `org_id`, so the intent's org is known on **both** CAS outcomes — the lost-race path still writes revocation and an audit row unconditionally, and that row has to be tenanted correctly too (the CAS `.returning()` is empty there). The already-decided path, where the flip transaction never runs at all, resolves the org through a system-scope lookup that never throws.

NULL is therefore only reached when there is honestly no org to name, rather than being the accidental result of reading tenancy off the caller's token.

**2. Write the audit row through the canonical helper.** The raw `db.insert(auditLogs)` is replaced with `writeAuditEventAsync`. `auditService.persistAuditLog` already performs exactly the required `runOutsideDbContext(() => withSystemDbAccessContext(...))` — and its comment block already names *both* this partner-scope NULL-org bug class and the transaction-poisoning class as the reasons it exists. This handler was simply bypassing the repo's central fix. Routing through it buys:

- the system-scope write, so a genuinely org-less report writes NULL instead of inventing a tenant;
- independent commit, so a failing audit write can no longer abort the caller's transaction;
- the bounded retry queue + Sentry escalation, payload sanitizing, and client IP/UA capture.

The independent commit is **deliberate and commented at the write site** so the next reader doesn't helpfully fold it back into the request transaction and restore the 500. A `security.suspicious_report` records that *someone reported something*, which stays true whether or not the flip that followed committed; losing the trail when surrounding work rolls back is the worse failure for a security action.

The write is still wrapped in `try/catch`: `writeAuditEventAsync` is not `async`, so its synchronous prologue (sanitizing, trusted-IP resolution, header reads) can throw before the retry-backed promise exists. "Audit machinery must never turn this security action into a 500" is enforced rather than assumed.

**3. Same poisoning class in the non-intent branch.** The `ai_tool_executions` mirror also ran inside the request transaction under a `try/catch` that could not heal it, *and* silently matched zero rows for any caller whose accessible-org list excludes the execution's org — no error, no mirror, and the SDK's `waitForApproval` never unblocks with the denial. It now runs in its own system-scope transaction, matching the intent CAS immediately above it.

**4. `refreshTokensRevoked` → `refreshRevocationCount`.** The audit sanitizer's `SECRET_KEY_PATTERN` matches the bare substring `token` anywhere in a key, so the integer count would have persisted as the string `[REDACTED]`. It is a count, never a credential. No consumer read the old key.

Both mounts are fixed **by construction** — `index.ts:1044` and `:1056` share this single handler, so the iOS app's "report suspicious" button on `/api/v1/mobile/approvals/:id/report-suspicious` is covered by the same change.

## Coverage

New `approvalsReportSuspiciousPartnerScope.integration.test.ts`: a **partner-scoped caller against real Postgres as `breeze_app` with RLS enforced**. That combination existed nowhere, which is why the bug survived — `approvals.test.ts` has exactly the right auth shape (`scope: 'partner', orgId: null`) but mocks `../db` wholesale so RLS never runs, and `approvalsDecideSupervised.integration.test.ts` has real RLS but hardcodes `scope: 'organization'`. A unit test cannot catch this class at all, because the abort happens in Postgres.

Four cases, one per arm of the resolution chain:

| Case | Asserts |
|---|---|
| intent-linked, CAS wins | 204 (was 500), row `reported`, intent `rejected`, audit org = intent org |
| intent-linked, CAS loses the race | 409 `already_decided`, audit org still = intent org |
| intent-linked, already decided | flip block skipped → org via the system-context lookup |
| standalone row, no intent | 204 with a NULL-org audit row rather than a 500 |

**All four fail against the unfixed handler** with the exact reported error and pass with the fix — verified by reverting `approvals.ts` to its `main` version and re-running:

```
× intent-linked, CAS wins …            AssertionError: expected 500 to be 204
× intent-linked, CAS loses the race …  AssertionError: expected 500 to be 409
× intent-linked but already decided …  AssertionError: expected 500 to be 204
× standalone approval with no intent … AssertionError: expected 500 to be 204
  cause: PostgresError: new row violates row-level security policy for table "audit_logs"
```

Registered in both hand-maintained lists (`vitest.integration.config.ts` include + `vitest.config.ts` exclude), per the convention its two sibling suites document.

## Verification

| Check | Result |
|---|---|
| `approvals.test.ts` (unit, single-fork) | 88 passed |
| new partner-scope integration suite | 4 passed (and 4 failed pre-fix) |
| `approvalsDecideSupervised` + `approvalsDecideAtomicity` integration | 13 passed total with the new suite |
| `vitest.config.rls.ts` | 22 passed |
| `vitest.config.rls-coverage.ts` | 75 passed |
| `tenantCascade` + `tenant-export-policy` integration | 7 passed |

Run against a private per-worktree stack (`pnpm test-stack up`) so the shared `:5433` database couldn't contaminate results.

## Schema / registration

No schema change, so no cascade or export-policy registration is due. `audit_logs` was already in `CORE_ORG_CASCADE_DELETE_ORDER` and `AUDIT_ADMIN_REQUIRED_TABLES`; both left intact.

## Known follow-ups (deliberately out of scope)

- **Execution-linked and elevation-linked** approvals reported by a partner-scoped caller still get `org_id = NULL`. An org *could* be derived through `ai_tool_executions → ai_sessions` or `elevation_requests`, but that is beyond the resolution order agreed on the issue and each needs its own join plus test. They no longer 500 — they just carry weaker attribution than the intent-linked case.
- The non-intent flip lacks a `status = 'pending'` CAS and neither it nor the mirror checks affected-row counts, so a concurrent decide can be overwritten while the caller still gets 204. Pre-existing; the intent branch already got this treatment in an earlier fix round.

Closes #3234

### DIFF
--- a/apps/api/src/routes/approvals.ts
+++ b/apps/api/src/routes/approvals.ts
@@ -845,12 +845,17 @@ approvalRoutes.post('/:id/report-suspicious', async (c) => {
       //      transaction poisoned and the commit threw — a 500 that also rolled
       //      back the 'reported' flip immediately above. Catching an error does
       //      not heal a Postgres transaction.
-      //   2. `ai_tool_executions` is org-scoped through its session, so under a
-      //      partner-scoped caller (or an org-scoped one whose accessible-org
-      //      list excludes the execution's org) the UPDATE silently matched
-      //      ZERO rows — no error, no mirror, and the SDK's waitForApproval
-      //      never unblocks with the denial. Same reasoning as the intent CAS
-      //      above, which is already system-scoped for exactly this.
+      //   2. `ai_tool_executions` is org-scoped through an EXISTS join on
+      //      `ai_sessions.org_id`, so whenever the caller's accessible-org list
+      //      does not cover the execution's org the UPDATE silently matched
+      //      ZERO rows — no error (an UPDATE filtered out by a USING clause is
+      //      not a violation, just an empty match), no mirror, and the SDK's
+      //      waitForApproval never unblocks with the denial. Note the trigger
+      //      is the ORG LIST, not the scope: a partner-scoped caller with
+      //      `orgAccess: 'all'` does reach the row, while one with
+      //      `orgAccess: 'selected'` that excludes the org does not — as does
+      //      any org-scoped caller in a different org. Same reasoning as the
+      //      intent CAS above, which is already system-scoped for exactly this.
       // Still best-effort: the approval flip and the audit row are the
       // authoritative record, and a missed mirror only delays the SDK waiter to
       // its own timeout rather than letting the flagged action through.
@@ -898,7 +903,13 @@ approvalRoutes.post('/:id/report-suspicious', async (c) => {
       ));
     } catch (err) {
       console.error('[approvals] report-suspicious: revocation failed:', err);
-      // Non-fatal: the approval row + audit log are still authoritative; the
+      // Escalated to Sentry, not just stdout: a swallowed failure here means the
+      // client the user just called compromised still holds a live grant and
+      // refresh tokens, while they get a 204. `revokeUserOauthClient` fails
+      // closed on Redis-marker problems, so this catch fires on real infra
+      // faults — exactly the case nobody would notice from a stdout line.
+      captureException(err);
+      // Still non-fatal: the approval row + audit log are authoritative, and the
       // user can revoke from the connected-apps UI as a fallback.
     }
   }
@@ -911,8 +922,9 @@ approvalRoutes.post('/:id/report-suspicious', async (c) => {
   // request transaction, which is precisely how this endpoint 500'd for every
   // partner-scoped caller:
   //
-  //   - `writeAuditEventAsync` persists via `auditService.persistAuditLog`,
-  //     which runs `runOutsideDbContext(() => withSystemDbAccessContext(...))`.
+  //   - `writeAuditEventAsync` persists via `createAuditLogAsync` →
+  //     `auditService.persistAuditLog`, which runs
+  //     `runOutsideDbContext(() => withSystemDbAccessContext(...))`.
   //     That is what lets a NULL org_id land at all: the `audit_logs` INSERT
   //     policy is `WITH CHECK breeze_has_org_access(org_id)`, and that helper
   //     returns FALSE for a NULL org under any non-system scope but TRUE under
@@ -933,8 +945,15 @@ approvalRoutes.post('/:id/report-suspicious', async (c) => {
   //
   // `writeAuditEventAsync`'s returned promise never rejects — a genuine write
   // failure is queued for retry with backoff and escalated to Sentry on
-  // exhaustion, so the security action still takes effect with a loud signal
-  // instead of a 500. The try/catch is still not redundant: the function is not
+  // exhaustion, so the security action still takes effect instead of 500ing.
+  // Be precise about how strong that is: `auditService`'s retry queue is
+  // in-memory and bounded (3 attempts, drained every 30s), so it covers a
+  // transient DB blip but NOT an outage that outlives the attempts, and a
+  // process restart landing mid-retry drops the entry with no signal at all.
+  // That is the repo-wide audit durability story rather than anything specific
+  // to this handler; it is still strictly better than the raw insert this
+  // replaced, which had no retry AND poisoned the request transaction.
+  // The try/catch is still not redundant: the function is not
   // `async`, so its synchronous prologue (details sanitizing, trusted-client-IP
   // resolution, header reads) can throw BEFORE the retry-backed promise exists.
   // "The audit machinery must never turn this security action into a 500" is
--- a/apps/api/src/routes/approvals.ts
+++ b/apps/api/src/routes/approvals.ts
@@ -9,7 +9,8 @@ import { approvalRequests } from '../db/schema/approvals';
 import { elevationRequests, elevationAudit } from '../db/schema/elevations';
 import { aiToolExecutions, aiSessions } from '../db/schema/ai';
 import { delegantM365Connections } from '../db/schema/delegant';
-import { auditLogs } from '../db/schema/audit';
+import { writeAuditEventAsync } from '../services/auditEvents';
+import { captureException } from '../services/sentry';
 import {
   actionIntents,
   intentOutbox,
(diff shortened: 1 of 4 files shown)


ToddHebebrand:

**Findings:** 8 raised → 4 addressed in `7f092e3a2`, 4 deferred with rationale below; 0 outstanding blockers.

- **code-reviewer: 0 findings** at every severity. It independently re-derived the RLS premise (`breeze_has_org_access(NULL)` FALSE under non-system / TRUE under system, `0001-baseline.sql:1663`; the `audit_logs` insert policy at `:16268`), traced all four org-resolution branches for a wrong or double-queried org, confirmed the system-scope mirror carries no tenant-isolation risk (the execution id comes from a Shape-6 row already filtered to the caller), and confirmed no cascade/export-policy list was missed.
- **comment-analyzer: 0 factually wrong claims** out of ~20 verified against the migrations, `db/index.ts`, and both sanitizer modules — including the fiddly one, that the unanchored `token` substring match lives in `logRedaction.ts`'s `SECRET_KEY_PATTERN` and fires *before* `auditPayloadSanitizer.ts`'s anchored pattern. Two imprecisions flagged, both fixed.
- **pr-test-analyzer: 1 critical + 3 lower.** The critical one was right and is the most valuable thing this review produced — see below.
- **silent-failure-hunter: 2 high + 3 medium**, all but one pre-existing.

### Addressed

**1. The execution-linked branch had no real-DB coverage** — despite this PR moving its `ai_tool_executions` mirror into a system-scope transaction and the commit message claiming that fix. Added a fifth integration case seeding a real `ai_sessions` + `ai_tool_executions` pair.

Writing it caught a mistake of mine. **The first version of that test passed even with the mirror reverted** — vacuous. The silent zero-row match is triggered by the caller's accessible-**org-list**, not by partner scope as such: a partner-scoped caller with `orgAccess: 'all'` *does* reach the row through the EXISTS-join policy on `ai_sessions.org_id`. The test now narrows the reporter to `orgAccess: 'selected'` excluding the execution's org, which is the real precondition. Reverting only the mirror now reds it with `expected 'pending' to be 'rejected'`. The code comment overclaimed the same way and has been corrected.

**2. Every audit-org assertion could have passed for the wrong reason.** The seed gave the reporting user the same org as the reported intent, so "the audit row carries the intent's org" was indistinguishable from a wrong implementation reading the reporter's own `users.org_id`. The partner now has two orgs — intent in A, reporter's own row in B — and the assertions pin A and explicitly reject B.

**3. The OAuth revocation catch was the only catch in this handler still logging to stdout alone** while the others escalate. A swallowed failure there leaves the client the user just called compromised holding a live grant while they get a 204, and `revokeUserOauthClient` fails closed on Redis-marker problems, so it fires on real infra faults. Now also `captureException` — one line, in the function this PR already touches.

**4. Softened an overclaim.** The comment promised a "loud signal" on audit-write failure. True up to retry exhaustion, but `auditService`'s queue is in-memory and bounded (3 attempts, 30s drain), so a longer outage or a restart mid-retry drops the entry silently. The comment now states the actual bound rather than implying durability it doesn't have.

### Deferred, with reasons

- **Already-decided rows return 204 indistinguishably from a report that stopped the action.** Real gap, but the 409 `already_decided` path today only covers a race *inside* this request; extending it to rows already terminal on arrival changes the response contract for existing callers including the iOS button. That's a product call, not a bug fix — flagging rather than deciding it.
- **The non-intent flip lacks a `status = 'pending'` CAS and no path checks affected-row counts.** Same class the intent branch got in an earlier fix round; needs its own concurrency test. Pre-existing.
- **Fault-injection coverage for `resolveReportAuditOrgId`'s catch and the audit-write catch.** Both are documented degrade-not-crash fallbacks; the repo has a trigger-based injection pattern (`createIntentAtomicity.integration.test.ts`) that would suit. Both reviewers called this a reasonable follow-up.
- **A unit test mirroring the org resolution** for a sub-second guard. Defense in depth only — the integration suite already proves the behaviour, and a unit test provably cannot catch this bug class.
- **Execution- and elevation-linked approvals still get `org_id = NULL`** for a partner-scoped reporter. Attribution via `ai_sessions.org_id` / `elevation_requests` is possible but beyond the intent → caller → NULL order settled on the issue. They no longer 500; they just carry weaker attribution. Now explicitly asserted in the new fifth case so the current behaviour is pinned rather than accidental.

### Tests

| Check | Result |
|---|---|
| new partner-scope integration suite | **5 passed**; all 5 proven non-vacuous by reverting the code under test |
| `approvals.test.ts` (unit, single-fork) | 88 passed |
| + `approvalsDecideSupervised` / `approvalsDecideAtomicity` | 14 passed across the 3 files |
| `vitest.config.rls.ts` | 22 passed |
| `vitest.config.rls-coverage.ts` | 75 passed |
| `tenantCascade` + `tenant-export-policy` integration | 7 passed |

Run against a private per-worktree stack (`pnpm test-stack up`) so the shared `:5433` database couldn't contaminate results.

**Status:** review-clean, awaiting maintainer merge. One decision for you — whether the already-decided 204 should become a 409, since that touches the mobile client's contract.

ToddHebebrand:
Converted to **draft** — this should not have been opened. Todd green-lit @bdunncompany on #3234 at 06:12Z ("green light — it's yours") and I dispatched work on it afterwards without re-reading the thread. That claim stands and the issue is his.

Leaving the branch up rather than deleting it because two findings in here look additive to the analysis already on the issue — the handler bypassing `writeAuditEventAsync` (which already solves both the NULL-org and transaction-poisoning classes, and says so in its own comment block), and a second instance of the poisoning class in the `ai_tool_executions` mirror write. Both are written up in #3234 for him to take or ignore.

**Do not merge this ahead of #3234's owner.**

PR #3421, opened 2026-08-11
fix(patching): make the patch-reboot user warning an invariant, not a coincidence (#3197)

### LINKED ISSUE #3197: [Agent] Patch-triggered reboots never warn the user — warning thresholds are unreachable at the hardcoded 5-minute delay

ToddHebebrand:
Closes #3197

## The defect

A customer workstation rebooted after patching with **zero notice**. Three things combined:

1. `patchRebootHandler.executeReboot` queued `schedule_reboot` with a hardcoded `delayMinutes: number = 5`, and `patchJobExecutor` never overrode it.
2. The agent's warning ladder (60/15/5 minutes before) gated every notification on `totalDelay > threshold`. **`5 > 5` is false**, so at the only delay the API ever sent, *zero* notifications were scheduled.
3. The closing `"Rebooting Now"` toast was fire-and-forget IPC on the line before `shutdown /r /t 0`, so it raced session teardown and essentially never rendered.

Net effect: the agent contained a complete staged-warning implementation that was unreachable.

`maintenanceRebootWorker` used its own independent 15-minute grace, so that path warned — the two reboot paths disagreed silently.

## Approach: make "at least one warning" an invariant, not a coincidence

Flipping `>` to `>=` would fix the one value that happened to be hardcoded and leave the scheme silent below 5 minutes. It would also introduce a *new* problem: at 6/16/61 minutes it emits two near-identical toasts a minute apart ("in 61 minutes", then "in 1 hour").

Instead:

- **A notification always fires the instant the reboot is scheduled**, quoting the real delay. This is what makes silence impossible at *any* delay, including a 2-minute emergency reboot that no before-the-reboot threshold can cover.
- Reminders at 60/15/5 remain, but are dropped when they would land within 5 minutes of the message the user just read — which is what actually kills the near-duplicate case, and is strictly stronger than `>=`.
- The closing critical notice fires **one minute out and starts the OS countdown at the same moment**, so the countdown the user sees is the countdown the OS honours. `OSInvokeAt + OSGrace == TotalDelay` is asserted.
- At a 1-minute delay the lead and closing notice collapse into a single critical toast rather than two in the same instant.

Worked examples (`Notifications[0].After` is always 0):

| Delay | Notification offsets | Before this PR |
|---|---|---|
| 5 min (the reported case) | 0, 4 | **none** |
| 2 min | 0, 1 | **none** |
| 15 min (maintenance grace) | 0, 10, 14 | 1 (only the 5-min reminder) |
| 16 min | 0, 11, 15 | 1 |
| 61 min | 0, 46, 56, 60 | 2 |
| 24 h | 0, 1380, 1425, 1435, 1439 | 3 |

## Where the configurable delay lives, and why

**`config_policy_patch_settings.reboot_delay_minutes`** — an existing Configuration Policy surface, default **15**, range **1–1440**. Not a new table, deliberately:

- The table has **no `org_id`**. Ownership and tenancy come transitively through `feature_link_id → config_policy_feature_links → configuration_policies`, and `configuration_policies` already carries the dual-ownership shape (`org_id` XOR `partner_id`). `'patch'` is already in `PARTNER_LINKABLE_FEATURE_TYPES`.
- So **partner-wide-first is satisfied by inheritance**: a partner-wide patch policy carries a partner-wide reboot delay with no new `one_owner_chk`, no new RLS policy, and no dual-axis test suite. A new table would have had to re-litigate all of it.
- Because there is no `org_id`, **no cascade or export-policy registration applies** — verified against `tenantCascade.ts`, `tenantExportPolicyRegistry.ts` and `routes/devices/core.ts` rather than assumed.
- `resolvePatchConfigForDevice` already walks the assignment hierarchy with `withPartnerWideVisibility`, so the resolver needed no change at all.

The one contract that *does* fire is the **canonical partner export**: the patch branch is a hand-enumerated `jsonb_build_object`, so a new column does not export itself. The migration re-emits the whole `breeze_partner_export_policy_settings_pre_patch` body (fix-forward from `2026-07-30-alert-rule-ownership-consolidation.sql`, the currently authoritative definition) with one key added, and `PATCH_NORMALIZED_MATERIAL_KEYS` gains the TS-side entry. Neither the parity suite nor the watermark suite enumerates columns, so **skipping either would have left CI green while the column silently never exported** — treated as a mechanical checklist item, not something CI catches.

## Reconciling the two reboot paths

`DEFAULT_REBOOT_DELAY_MINUTES` is now the single fallback for both. `MAINTENANCE_REBOOT_GRACE_MINUTES` is **gone**: `decideRebootCommand` takes `delayMinutes` from the caller, and `processRebootCandidate` resolves it per device off the same patch policy. A test asserts `rebootWarranted` and `decideRebootCommand` agree on every gate combination, so they cannot drift.

The gate was split out from the decision (`rebootWarranted`) so the fleet-wide 10-minute sweep does **not** pay for a policy-hierarchy lookup on every online device — only on devices actually about to reboot. Two tests pin that.

## Cross-platform: the stub was worse than the reported bug

`reboot_other.go` (`//go:build !windows`) was a no-op whose `Schedule()` returned `nil` and did nothing, while `patchJobExecutor` dispatched `schedule_reboot` with no OS check. On Linux and macOS the command reported **success**, the device never rebooted, and the patch job recorded a clean result. Shipping "patch reboots now warn the user" as Windows-only would have left that in place, and the notify plumbing already worked end-to-end on both platforms — it was simply compiled out.

Restructured so only the OS invocation is platform-specific:

- `reboot_plan.go` (**untagged**) — the schedule arithmetic, as pure data.
- `reboot_manager.go` (**untagged**) — timers, state, circuit breaker, history.
- `reboot_os_{windows,linux,darwin,other}.go` — just `exec.Command("shutdown", …)`.
- `log.go` (**untagged**) — the package logger moved out of the windows-tagged `windows.go`, which the untagged core requires.

`_other.go` returns an error rather than `nil`, so an unsupported platform can no longer look like a successful reboot.

### macOS `notify` scope (#3255 / PR #3364)

This is exactly the cross-platform notification caller that `Broker.BroadcastNotification`'s doc comment anticipated. Per the instruction left there, the fix is to **grant the macOS desktop helper `notify` in `scopesForRole`** — not to weaken the broadcast filter, which still has to keep the assist helper and watchdog out of any delivery accounting. The grant is a named `macDesktopHelperScopes` slice so a test can pin it, and the pre-#3197 `["desktop"]`-only case is kept as a regression case proving the filter was not weakened.

## Test coverage — and whether it actually runs in CI

Windows-tagged Go files are tested nowhere in CI (#3019, #3046): `test-agent` builds for linux, and `test-agent-windows` does not list `./internal/patching` in its package allowlist. **Every new Go test here is in an untagged file**, so it runs in `test-agent`. Verified: 24 reboot tests green on linux; `./internal/patching`, `./internal/heartbeat`, `./internal/sessionbroker` all green; `GOOS=windows`, `GOOS=darwin` and `GOOS=linux` builds and `go vet` all clean (CI's `Build Agent (windows/amd64)` matrix job compile-checks the windows target, and `./internal/sessionbroker` *is* in the windows test list).

The seconds-vs-minutes conversion is the highest-risk untestable-on-CI code, and that exact trap already shipped once as #3252 (`delay: 15` → 15 *seconds* on Windows, 15 *minutes* on Linux). So `windowsRebootArgs` / `unixRebootArgs` are untagged and directly asserted — Windows `/t` in seconds, unix `+N` in minutes, never `+0`.

Two bugs the tests found while being written, both now fixed:

- **A cancelled reboot could still reboot the machine.** `time.Timer.Stop()` does not wait for an `AfterFunc` callback that has already started, so `Cancel()` could return success while the OS invocation proceeded on its own goroutine. Every callback now captures a generation counter and no-ops if superseded.
- **`clampRebootDelayMinutes(null)` returned 1, not the default.** `Number(null)` is `0`, which clamped to MIN — a one-minute warning. Unusable input now means "use the safe default", never "give the user the shortest possible notice".

`patchRebootHandler` had no unit test file at all; it now has 24.

## Deliberately NOT in this PR

Kept scoped to "the warning actually fires". Cross-referenced, not absorbed:

- **#3207** — deferral, toast action buttons, response-bearing broadcast, re-notifying a user who signs in mid-delay. Its scope shrinks: the non-Windows `RebootManager` lands here.
- **#3253** — `deadline` is stored and forwarded but never enforced. Still true. (`handleScheduleReboot` also silently swallows a malformed `deadline`, unlike its strict `delayMinutes` parse from #3393 — noted there, not changed here.)
- **#3254** — no console visibility into a scheduled reboot. This PR makes it slightly cheaper: `RebootState` now carries `notificationsPlanned`, which `get_reboot_status` already returns.
- **Linux keeps two reboot mechanisms** — maintenance sends `reboot`, patch sends `schedule_reboot`. Migrating the maintenance path onto `schedule_reboot` now that `RebootManager` is cross-platform is the obvious follow-up, held back so a regression in either behaviour stays attributable.
- **Dead maintenance notify settings** (`notifyBeforeMinutes` / `notifyOnStart`, issue item 3 / #3256) — a separate defect class (settings that read as implemented but have no consumer). Not touched.
- **`rebootPolicy` dropped on `aiToolsFleet` / `patchJobService` job-creation paths** — a real bug, but about which policy applies, not about whether the warning fires.

One in-scope hardening did come along: `executeReboot` now passes `expectedOrgId`, arming the cross-tenant guard at `commandQueue.ts` on a dispatch that runs under a system DB context (RLS off). The maintenance worker already did this.

## Needs a real Windows box to confirm

Stated plainly rather than claimed as verified:

- `shutdown /r /t 60 /d p:2:17` actually rendering the Windows toast and honouring the countdown. The argument construction is unit-tested; the invocation is not, and cannot be on this CI matrix.
- The desktop toast rendering for a logged-in user on **macOS** via the newly-`notify`-scoped desktop helper, and on **Linux** via `notify-send`. The scope grant and the broadcast filter are tested; the pixels are not.
- `shutdown -c` (Linux) actually aborting, and macOS correctly reporting that it cannot.
- End-to-end: a real patch job on a real device producing the toast sequence above.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 5 agents: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer.

**Findings:** 2 Critical, 2 High, 3 Medium, plus comment/type notes → **all acted on or explicitly declined with a reason; 0 outstanding.** Fixes are in `d875b81c5`.

Both Critical findings were real defects in the OS-grace window this PR introduces — and both were invisible because the OS countdown outlives the agent process:

1. **`Cancel()` could never abort a live OS countdown.** `runOSReboot` clears `RebootScheduled` *before* invoking the OS command, so for the whole final minute — exactly when `abortOSReboot` is the only thing that can still stop the machine — the `!RebootScheduled` gate rejected the request with a misleading "no reboot scheduled" and never attempted the abort. The `osInvoked`/`abortOSReboot` machinery was unreachable in the one scenario it was built for. The tell was in my own test: `TestCancelAfterOSCountdownSurfacesAbortFailure` had to hand-patch `state.RebootScheduled = true` to reach the code under test, i.e. it asserted against a state production cannot produce.
2. **`Schedule()` cleared `osInvoked` unconditionally**, so a reschedule arriving in that same window forgot the live countdown: the machine still went down at the OLD, unannounced time, and a later `Cancel()` reported success because nothing believed a reboot was pending. `Schedule` now aborts the in-flight countdown first and **refuses** the reschedule when it cannot be aborted (macOS has no cancel flag) rather than promising a time it cannot deliver.

Both fixes ship with regression tests **verified to fail without the fix** (temporarily reverted each gate to confirm; not assumed):
- old `Cancel` gate → `Cancel during the OS countdown: no reboot scheduled`
- old `Schedule` → `abortOSReboot called 0 times, want 1 — the old OS countdown was left running` / `Schedule returned nil while an un-abortable OS countdown was live`

Also fixed: `RebootState.LastError` now distinguishes "machine is going down" from "invocation failed, machine still up" (`RebootScheduled:false` meant both); `patchJobExecutor` reports a failed reboot dispatch to Sentry, not just `console.warn` (the maintenance path added in this same PR already did — the post-patch path, the one #3197 is about, was the weaker one); an out-of-range stored delay is reported rather than silently clamped; a corrupt `reboot_history.json` warns instead of logging at Debug, since that file is the reboot-loop circuit breaker's memory; and a stale cross-reference to a nonexistent test name was corrected.

**Two suggestions declined, with reasons:**
- Renaming `LinuxRebootPayload.delay` → `delayMinutes` "for symmetry". **Rejected** — that field name *is* the wire key (`tools.ShutdownDelayMinutes` parses `payload["delay"]`), so the rename would leave the agent parsing a missing key and silently break Linux maintenance reboots. Documented the unit in a comment instead.
- `RebootResult` → discriminated union, and typed `Urgency`. Both correct in the abstract, both declined here: the urgency boundary is already guarded by `sanitizeNotifyRequest`'s allowlist, and the union is a repo-wide pattern decision that doesn't belong in a bugfix PR.

**Tests:** all 20 new Go tests **confirmed executing in CI by name** in the `Test Agent` job (`TestPlanRebootAlwaysWarns`, `TestWindowsRebootArgsUseSeconds`, `TestMacDesktopHelperScopesGrantNotify`, …) — that was the whole point of moving the logic into untagged files, given windows-tagged files run nowhere in CI (#3019, #3046), so it is verified rather than asserted. Locally on `d875b81c5`: `./internal/patching`, `./internal/heartbeat`, `./internal/sessionbroker` all `ok`; `GOOS=windows/darwin/linux` build + vet clean; `tsc --noEmit` clean for api/web/shared; 68 affected API tests pass. The migration was applied against a throwaway Postgres to prove it **replays idempotently**, that the CHECK rejects 0/1441, and that the re-emitted 190-line canonical export function compiles and actually emits `rebootDelayMinutes`.

**CI:** parent commit `a6556d728` went **57/58** (the one non-pass being `Main Red Alert: skipping`, a main-only workflow) — including **all four Integration Tests shards**, which is what validates the two integration-test edits I could not run locally. CI for `d875b81c5` is re-queued and slow on the runners at time of writing; the delta is Go-only plus two logging additions and a comment.

**Needs a real Windows/macOS box** (stating plainly, not claiming verified): the `shutdown /r /t 60 /d p:2:17` invocation actually rendering the Windows toast and honouring the countdown; the toast rendering for a logged-in macOS user via the newly-`notify`-scoped desktop helper, and on Linux via `notify-send`; `shutdown -c` actually aborting on Linux. Argument construction and the scope grant are unit-tested; the pixels and the OS calls are not.

**Known follow-up, not blocking:** the reboot-dispatch branch in `patchJobExecutor.recordDeviceExecution` (including the new `expectedOrgId` wiring) has no test — driving it needs ~10 mocked DB calls, and `orgId` is already validated against `patchJob.orgId` upstream. Flagged rather than covered with a brittle test.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3197 — the reporter/maintainer's call.

Issue #3137, opened 2026-08-04
macOS agent.sock: breeze group created but never applied, blocking user-helper IPC


ToddHebebrand:
Reported by **semotech** in #bugs (Discord, message 1534257683362742472).

**Report (PII-redacted):**

Thanks — useful. On this 2019 iMac (macOS 14.7.2 / amd64 / agent 0.103.0):

Terminal: matches what we saw (paste reordering). Session drop on `launchctl kickstart -k system/com.breeze.agent` understood. We'll retest terminal paste after next agent build past 0.103.0.

Helper: User already toggled Screen Recording ON for breeze-desktop-helper (off→on), FDA ON, Admin user, reboot — TCC prompt still returns. LaunchAgent `com.breeze.desktop-helper-user` crash-looped (runs≈133, last exit 1). When run manually as jingxie we got:
- screenRecording=false / captureGranted=false (ScreenCaptureKit timeout) even while Settings showed the toggle ON
- accessibility=true, fullDiskAccess=true
- fatal: dial unix ".../Breeze/agent.sock": permission denied
- sock was `srw-rw---- root:admin` — chmod 666 let us proceed; after reboot sock likely recreated tight again
- from Breeze root PTY, `launchctl kickstart/bootstrap` on `gui/<uid>/…` returns 125 Domain does not support specified action (Aqua domain not manageable from agent terminal)

Will pull `"/Library/Application Support/Breeze/desktop-helper.log"` next. Questions:

1) Intended agent.sock mode/owner for user-session helper? admin-only 660 breaks non-admin (and we still saw denied before chmod).
2) Known issue where Screen Recording toggle shows ON but helper still reports screenRecording=false until a GUI-session restart of the helper (not reboot alone)?
3) Recommend path when gui launchctl returns 125 from remote terminal?

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

## Triage analysis

[...]

Issue #3134, opened 2026-08-04
macOS agent.sock (0660, group=admin) blocks Standard-user helper connections, contradicting stated intent


ToddHebebrand:
Reported by **semotech** in #bugs (Discord, message 1534282519237759187).

**Report (PII-redacted):**

Reinstalled via install.sh on 2019 iMac (macOS 14.7.2 amd64, agent 0.103.0): upgrade OK, already enrolled (no --force). Reinstall restored loginwindow helper → Screen Recording prompts returned after reboot. After console login Breeze shows FDA/SR/Accessibility Granted; Remote Desktop still Missing; Desktop unavailable. Also saw Splashtop Streamer Screen Recording prompt on this host — is Splashtop part of Breeze RD on macOS, and must it be granted separately? Disabling loginwindow plist again as workaround.

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

## Triage analysis

[...]

Issue #3133, opened 2026-08-04
macOS agent.sock chmod 0660 (root:admin) blocks Standard non-admin user helpers


ToddHebebrand:
Reported by **semotech** in #bugs (Discord, message 1534284370544824472).

**Report (PII-redacted):**

Reinstalled via install.sh on 2019 iMac (macOS 14.7.2 amd64, agent 0.103.0): upgrade OK, already enrolled (no --force). Reinstall restored loginwindow helper → Screen Recording prompts returned after reboot. After console login Breeze shows FDA/SR/Accessibility Granted; Remote Desktop still Missing; Desktop unavailable. Also saw Splashtop Streamer Screen Recording prompt on this host — is Splashtop part of Breeze RD on macOS, and must it be granted separately? Disabling loginwindow plist again as workaround.

Bug Summary: macOS “Remote Desktop” permission stuck Missing — Desktop unavailable

Hardware: 2019 iMac (Retina 5K 27", 3.6 GHz 8-core i9, 32 GB, Radeon Pro 580X 8 GB)
OS: macOS 14.7.2 (darwin 23.6.0) amd64
Agent/watchdog: 0.103.0

Symptoms:
- Breeze macOS Permissions: FDA Granted, Screen Recording Granted, Accessibility Granted
- Remote Desktop: Missing (grayed) — banner: required for unattended login-window access on macOS 14+
- Breeze Desktop control unavailable/grayed while user is logged in at console
- No Screen Recording TCC popups after disabling loginwindow helper

Related findings during debug:
- com.breeze.desktop-helper-loginwindow (root, context=login_window) spam-requested Screen Recording / CGDisplayStream failures; moving that plist aside stopped popup loop
- agent.sock was srw-rw---- root:admin — user-session helper got permission denied until chmod 666 (breaks Standard users)
- Reinstall/upgrade via install.sh restored loginwindow plist and prompts; did not clear Remote Desktop Missing
- Privacy “Remote Desktop” pane unclear vs Sharing→Remote Management (ARD On on this host)

Questions:
1) How is “Remote Desktop” permission detected/granted on 14.7? Exact System Settings path + binary?
2) With SR+Accessibility+FDA Granted and user_session helper running, why is interactive Desktop still unavailable?
3) Intended agent.sock permissions for user helper?
4) Is disabling loginwindow helper supported, or is that required for Remote Desktop / pre-login?

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

[...]

Issue #3019, opened 2026-08-02
[CI][Agent] Windows CI tests 6 of 45 packages containing Windows-specific code — three PRs in one batch found their tests ran nowhere


ToddHebebrand:
## Description

`test-agent-windows` runs an **explicit package list**, not `./...`. It currently covers **6 packages**; the agent has **45** containing Windows-specific code. Windows is the primary agent platform for this product, so most Windows-only code has never been executed by CI.

This is not theoretical. **Three separate PRs in a single batch discovered their Windows tests were running on no platform at all:**

| PR | package | consequence |
|---|---|---|
| #3002 | `internal/backup` | a wrong Win32 error constant survived the first pass — `0x17C` (`ERROR_CLOUD_FILE_INVALID_REQUEST`) instead of `395` (`ERROR_CLOUD_FILE_ACCESS_DENIED`). Would have shipped a fix that skipped an unevidenced error while leaving the real OneDrive case broken. |
| #3005 | `internal/backup/vss` | the VSS COM layer had a **wrong vtable index**, plus `StartSnapshotSet` never called, `GUID_NULL` passed by value, and a 112-byte buffer for a 128-byte struct (a heap overwrite on every successful call). VSS had been failing on **every backup on every machine**. |
| #3013 | `internal/ipc` | sequence-number race on the shared helper transport. |

Each PR patched its own package into the list. Nobody has audited the whole set — hence this issue.

## Current coverage

```
ci.yml:1141  go test       ./internal/sessionbroker ./internal/eventlog ./internal/logging \
                           ./internal/state ./internal/watchdog ./cmd/breeze-watchdog
ci.yml:1165  go test       ./internal/backup -run 'TestPermanentWindowsErrno…'   # filtered, #3002
ci.yml:1180  go test -race ./internal/sessionbroker ./internal/eventlog ./internal/state \
                           ./internal/watchdog ./cmd/breeze-watchdog
```

Plus `./internal/ipc` (#3013) and `./internal/backup/vss` (#3005), both still unmerged.

## Uncovered packages with Windows-specific code (36)

```
cmd/breeze-backup
cmd/pamlaunchtest
internal/agentapp
internal/backup/bmr
internal/backup/hyperv
internal/backup/mssql
internal/backup/systemstate
internal/cis
internal/collectors
internal/config
internal/elevaccount
internal/etwlua
internal/executor
internal/heartbeat
internal/helper
internal/mgmtdetect
internal/monitoring
internal/onedrivehelper
[...]

PR #3422, opened 2026-08-11
fix(agent): apply the breeze group to the macOS IPC socket, not just create it (#3137)

ToddHebebrand:
## Problem

On macOS the agent **creates** a `breeze` group but never **applies** it to the IPC socket, and never puts any user in it. So `/Library/Application Support/Breeze/agent.sock` comes up `srw-rw---- root:admin` on every daemon start — BSD semantics give a new file the *parent directory's* group, and the config dir is group `admin`.

Mode `0660` + group `admin` means a Standard (non-admin) console user's desktop helper can never dial the socket. `com.breeze.desktop-helper-user` then crash-loops on `dial unix ".../Breeze/agent.sock": permission denied` (reporter observed runs≈133, last exit 1), which cascades into the TCC symptoms in the linked threads:

- Screen Recording reads **Granted** in System Settings while the helper reports `screenRecording=false` / `captureGranted=false` (ScreenCaptureKit timeout) — the probe runs in the helper's context, and the helper is dead.
- The loginwindow helper re-prompts for Screen Recording endlessly.
- Remote Desktop reads **Missing** / Desktop unavailable with FDA + SR + Accessibility all granted, because the agent never gets a live user-session helper link for that account.

Admin users worked only incidentally (they happen to be in `admin`). The only workaround was `chmod 666` on the socket, reapplied after every install, restart and reboot, since `setupSocket` re-chmods to `0660` on every start.

The `breeze` group creation already existed in all three install paths (`installer/macos/postinstall`, `scripts/install/install-darwin.sh`, `breeze-agent service install`). Only the *application* of it was missing — the mirror-image of `scripts/install/install-linux.sh`, which both chowns the IPC dir to `root:breeze` and adds every logged-in user to the group.

## Fix

Both halves of the invariant are now enforced, and both self-heal.

**1. The socket is group-owned.** `sessionbroker.setupSocket` resolves the `breeze` GID and chowns the socket to `root:breeze` before `chmod 0660`. It runs on **every daemon start**, so the socket comes back correct after a reboot, after `launchctl kickstart -k system/com.breeze.agent`, and after a self-update — the reporter's "sock likely recreated tight again after reboot" case.

Deliberately **not** `0666`. Peer-credential + binary-path verification (`internal/ipc`) stays the real admission gate, but a world-writable socket lets any local process connect and spend the broker's admission budget, and it is a footgun if the peer checks are ever loosened. `TestIPCSocketModeIsNotWorldAccessible` pins this so a future "simpler fix" cannot land `0666` as shipped behaviour.

**2. The console user is a member.** Group-owning the socket is useless alone: an admin installs the agent, a Standard user logs in later, and that user is in no group the socket grants. So:

- The daemon adds every logged-in GUI user to `breeze` whenever it touches the helper LaunchAgents — login, session switch, post-update kickstart — via `ensureDarwinHelperPrereqs`. This covers users who first log in months after install, which the installers cannot.
- Both macOS installers add users logged in at install time, mirroring the `install-linux.sh` loop.
- Always **before** any `launchctl kickstart`/`bootstrap`: a helper inherits its group list when it starts.
- Root and service accounts (uid < 500) are excluded, so the group is not widened to `_windowserver` and friends.

**Supporting fixes**

- The group is **re-ensured on every daemon start**, so a host installed by a build predating the group — or where an admin deleted it — recovers instead of staying broken forever.
- `breeze-agent service install` created the group *after* bootstrapping the helper LaunchAgents. Ordering reversed. The group-creation shell script now has one Go home (`sessionbroker.EnsureIPCGroup`) rather than being duplicated in `agentapp`.
- An unresolvable group logs a `WARN` naming the remedy and leaves the socket at `0660` instead of failing the broker. Aborting would leave the agent with **no** IPC socket, taking remote desktop, backup IPC and the watchdog link with it — strictly worse than the pre-fix root-only socket.
- `dscl` calls are bounded by a timeout, so a wedged `opendirectoryd` (a real macOS failure mode) cannot hold daemon startup — and therefore the whole session broker — hostage. Usernames are validated before reaching a privileged argv.
- The `dscl` GID lookup is Directory-Services-first, with `os/user` as fallback. This is load-bearing: release darwin binaries are built `CGO_ENABLED=0` (`scripts/build-edition.sh` defaults it to 0), and cgo-less `os/user` parses `/etc/group`, which never contains a `dscl`-created group. `os/user` alone would have failed on exactly the shipped configuration.

**Linux had the same latent defect** — `install-linux.sh` chowns `/run/breeze` to `root:breeze`, but the socket itself was created `root:root 0660`, so non-root helpers were locked out there too. The same `setupSocket` chown fixes it, and `os/user` resolves the group correctly from `/etc/group` without cgo.

## Tests

CI has **no macOS runner** — `test-agent` is `ubuntu-latest`, `test-agent-windows` is `windows-latest` — so anything in a `//go:build darwin` file compiles and runs **nowhere** (the #3019/#3046 lesson). The coverage is deliberately placed to actually run:

- `internal/sessionbroker/ipc_group_test.go` (**untagged** — runs on both the ubuntu and windows jobs): the ownership decision, the mode invariant, the `dscl` argv builders, `PrimaryGroupID` / `GroupMembership` parsing (inline + dscl's folded output shape, missing group, missing key), whole-token membership matching (`jing` must not match `jingxie`), and username validation.
- `internal/sessionbroker/ipc_group_unix_test.go` (**`!windows`** — runs on the ubuntu job): the real `chown`/`chmod` syscalls against a real unix socket via `setupSocket`, covering the happy path, the missing-group path, parent-directory traversability, and a stale socket left by a killed daemon.
- `internal/heartbeat/helper_group_test.go` (**untagged**): the UID-eligibility predicate.

Two things worth noting about the socket tests:

- They chown to a **supplementary** group, not the process's primary GID. A socket is already created with the primary GID, so asserting that would pass even if the chown never happened — the exact bug under test.
- They were **mutation-checked**: forcing `owner.GID = -1` in `setupSocket` (pre-fix behaviour) fails `TestSetupSocketAppliesGroupAndRestrictiveMode` and `TestSetupSocketReplacesAStaleSocket`. Reverting makes them pass.

**Status**
- `go test ./...` (agent, `CGO_ENABLED=0`, CI mode): green
- `go test -race` on `internal/sessionbroker` + `internal/heartbeat`: green
- `go vet` + `go build` for darwin arm64/amd64, linux amd64, windows amd64: clean (the darwin-tagged files CI never compiles were cross-vetted explicitly)
- `gofmt` clean on every touched file; `shellcheck -S warning` clean on both modified installers

## Not verified here — needs a real Mac

I have no macOS host with the agent installed, so the end-to-end behaviour is **not** field-verified. Specifically:

1. **Socket ownership** is deterministic and covered by the syscall tests, so I am confident `agent.sock` lands `root:breeze 0660` and survives reboot/kickstart.
2. **Membership taking effect for an already-logged-in user** is the open question. macOS resolves group membership through `opendirectoryd`, and it is not certain a helper restarted mid-session picks up a group added after that GUI session began. Worst case it takes effect at the user's next login — still strictly better than today, where it never works at all, and no worse than a `chmod` that must be reapplied after every restart. A Standard-user login test on a real Mac is the thing to confirm before promoting the build.

## For @SemoTech

Once this ships in an agent build, please **remove both workarounds** and retest:

1. Stop the `chmod 666` on `agent.sock` after install/restart — the socket should now be `srw-rw---- root:breeze` on its own, and stay that way across reboot and `launchctl kickstart -k system/com.breeze.agent`. Please confirm with `ls -l "/Library/Application Support/Breeze/agent.sock"`.
2. Restore `com.breeze.desktop-helper-loginwindow.plist` — the prompt loop was downstream of the dead user-session helper. (Note the separate point from the earlier thread: login-window WebRTC desktop control cannot work on macOS regardless, since Apple blocks synthetic input there without a private entitlement; that path falls back to the VNC relay.)

The check that matters for a **Standard (non-admin)** user: `id -Gn <user>` should list `breeze`, and their `com.breeze.desktop-helper-user` LaunchAgent should stop crash-looping. If a user who was already logged in when the agent upgraded still cannot connect, please log them out and back in and report whether that fixes it — that is exactly the uncertainty in point 2 above, and your answer decides whether a forced helper re-login is also needed.

Refs #3133
Refs #3134
Refs #3137

(Not `Closes` — these are Discord-relayed community reports, so they stay open until the reporter verifies on real hardware.)

### DIFF
--- a/agent/installer/macos/postinstall
+++ b/agent/installer/macos/postinstall
@@ -65,8 +65,50 @@ if [ -f "$WATCHDOG_PLIST" ]; then
         launchctl load "$WATCHDOG_PLIST" 2>/dev/null || true
 fi
 
+# Add every logged-in GUI user to the breeze group so their desktop helper can
+# dial the 0660 root:breeze IPC socket. Mirrors the equivalent loop in
+# scripts/install/install-linux.sh. Without this the socket is group-owned by a
+# group nobody is in, and Standard (non-admin) users' helpers are denied
+# (#3133/#3134/#3137).
+#
+# Runs BEFORE the LaunchAgents are bootstrapped below: a helper process inherits
+# its group list when it starts, so membership has to exist first.
+add_console_users_to_breeze_group() {
+    local uid username
+    # NOTE: this script runs under `set -e`, so every test below is written as an
+    # `if`. A bare `[ ... ] && continue` returns non-zero when the test is false
+    # and would abort the whole postinstall.
+    for uid in $(ps -axo uid= -o comm= | grep -i '[lL]oginwindow' | awk '{print $1}' | sort -u); do
+        # macOS gives human accounts UIDs from 500 up; below that are system and
+        # service accounts, which never run a Breeze desktop helper.
+        if ! [ "$uid" -ge 500 ] 2>/dev/null; then
+            continue
+        fi
+        username=$(id -un "$uid" 2>/dev/null) || continue
+        if [ -z "$username" ]; then
+            continue
+        fi
+        if dscl . -read /Groups/breeze GroupMembership 2>/dev/null | tr ' ' '\n' | grep -qx "$username"; then
+            continue
+        fi
+        if dscl . -append /Groups/breeze GroupMembership "$username" 2>/dev/null; then
+            echo "Added $username to the breeze group for desktop-helper socket access"
+        else
+            echo "Warning: could not add $username to the breeze group; that user's desktop helper will be denied the agent socket" >&2
+        fi
+    done
+}
+
 # Create breeze group for IPC socket access.
 ensure_breeze_group
+add_console_users_to_breeze_group
+
+# Group-own the config directory so anything created inside it inherits the
+# breeze group under BSD semantics. The daemon chowns agent.sock explicitly on
+# every start (sessionbroker.setupSocket); this is defence in depth. Mode is
+# unchanged — agent.yaml is deliberately world-readable for the user helper
+# (#2483) and secrets.yaml stays 0600 root-only.
+chgrp breeze "$CONFIG_DIR" 2>/dev/null || true
 
 # Stop existing service if running
 launchctl bootout system/com.breeze.agent 2>/dev/null || true
--- a/agent/installer/macos/postinstall
+++ b/agent/installer/macos/postinstall
@@ -65,6 +65,11 @@ if [ -f "$WATCHDOG_PLIST" ]; then
         launchctl load "$WATCHDOG_PLIST" 2>/dev/null || true
 fi
 
+# breeze_group_has_member reports whether $1 is in the breeze group.
+breeze_group_has_member() {
+    dscl . -read /Groups/breeze GroupMembership 2>/dev/null | tr ' ' '\n' | grep -qx "$1"
+}
+
 # Add every logged-in GUI user to the breeze group so their desktop helper can
 # dial the 0660 root:breeze IPC socket. Mirrors the equivalent loop in
 # scripts/install/install-linux.sh. Without this the socket is group-owned by a
@@ -75,9 +80,10 @@ fi
 # its group list when it starts, so membership has to exist first.
 add_console_users_to_breeze_group() {
     local uid username
-    # NOTE: this script runs under `set -e`, so every test below is written as an
-    # `if`. A bare `[ ... ] && continue` returns non-zero when the test is false
-    # and would abort the whole postinstall.
+    # This script runs under `set -e`. The `|| continue` on the id lookup is
+    # load-bearing: a bare failing assignment would abort the whole postinstall.
+    # (A failing `[ ... ] && continue` would NOT — set -e exempts a command that
+    # is part of an && list other than the last one.)
     for uid in $(ps -axo uid= -o comm= | grep -i '[lL]oginwindow' | awk '{print $1}' | sort -u); do
         # macOS gives human accounts UIDs from 500 up; below that are system and
         # service accounts, which never run a Breeze desktop helper.
@@ -88,10 +94,13 @@ add_console_users_to_breeze_group() {
         if [ -z "$username" ]; then
             continue
         fi
-        if dscl . -read /Groups/breeze GroupMembership 2>/dev/null | tr ' ' '\n' | grep -qx "$username"; then
+        if breeze_group_has_member "$username"; then
             continue
         fi
-        if dscl . -append /Groups/breeze GroupMembership "$username" 2>/dev/null; then
+        dscl . -append /Groups/breeze GroupMembership "$username" 2>/dev/null || true
+        # Verify by re-reading rather than trusting dscl's exit status, so the
+        # success line below cannot claim a membership that did not take.
+        if breeze_group_has_member "$username"; then
             echo "Added $username to the breeze group for desktop-helper socket access"
         else
             echo "Warning: could not add $username to the breeze group; that user's desktop helper will be denied the agent socket" >&2
@@ -108,7 +117,8 @@ add_console_users_to_breeze_group
 # every start (sessionbroker.setupSocket); this is defence in depth. Mode is
 # unchanged — agent.yaml is deliberately world-readable for the user helper
 # (#2483) and secrets.yaml stays 0600 root-only.
-chgrp breeze "$CONFIG_DIR" 2>/dev/null || true
+chgrp breeze "$CONFIG_DIR" 2>/dev/null || \
+    echo "Warning: could not group-own $CONFIG_DIR as breeze" >&2
 
 # Stop existing service if running
 launchctl bootout system/com.breeze.agent 2>/dev/null || true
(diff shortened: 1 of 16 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (3 agents, parallel).

**Findings:** 9 raised → 8 addressed in `54260b933`; 1 deferred with rationale below; 0 outstanding blockers.

Addressed:
- **chown failure killed the whole broker** (code-reviewer). `applySocketOwner` now returns `ChownErr` and `ChmodErr` separately because they differ in severity. A chown failure is non-fatal — a root-only socket is the pre-fix status quo, while failing the broker takes remote desktop, backup IPC and the watchdog link with it. A chmod failure stays **fatal**: `net.Listen` creates the socket at `0777&^umask`, so an unapplied mode is the one outcome that could be *more* permissive than intended, and that must fail closed. The chmod is now attempted even after a failed chown, for the same reason.
- **Tolerated membership-read error was swallowed** (silent-failure-hunter). Now logged. The benign case (a freshly created group has no `GroupMembership` key) is indistinguishable at that layer from a real read failure, and it fires at most once per host — so a line that keeps recurring is itself the signal.
- **Both shell installers claimed success on `dscl`'s exit status** — exactly the signal the Go path was hardened against trusting (silent-failure-hunter). They now verify by re-reading the group, via a shared `breeze_group_has_member`.
- **Silent `chgrp`** on the config dir now warns on failure.
- **Install ordering had zero test coverage** (pr-test-analyzer, rated the highest-ROI gap — get it wrong again and #3137 silently re-ships). Extracted `installIPCPrereqsThenHelpers` into an **untagged** file with tests asserting the call order `group → members → helpers` and that nothing proceeds when the group cannot be created. That invariant was previously protected by prose alone.
- **`EnsureIPCGroupMember`'s six branches were untested** and unreachable by any test. Extracted the orchestration into untagged, injectable `ensureGroupMemberViaDscl` / `lookupGroupIDViaDscl`; the darwin files keep only the `exec.Command` calls. All six branches plus the dscl→os/user fallback composition are now table-tested on a required job.
- **Vacuous-skip path made loud** — `chownTargetGID` now emits a `VACUOUS-GUARD:` log line rather than skipping silently, so a runner that quietly stopped exercising the chown assertions is visible.
- **Two of my own comments were factually wrong**, and both are corrected:
  - I claimed *"CI has no macOS runner."* It does — `test-agent-race` on `macos-latest`. It is simply absent from `ci-success`'s `needs` list, so it cannot **block** a merge. The real rationale for untagged tests is "the gating jobs are ubuntu/windows," not "no macOS runner exists." Also corrected: `helper_group_test.go` does *not* run on `test-agent-windows`, which names an explicit package list omitting `internal/heartbeat` (#2523).
  - I claimed `[ ... ] && continue` would abort under `set -e`. **It does not** — I verified empirically; `set -e` exempts a non-final command in an `&&` list. The actually load-bearing guard is the `|| continue` on the `id` lookup, since a bare failing assignment *does* abort. Comment rewritten to say the true thing.

Deferred (noted, not done in this PR):
- A **shell test harness** for the duplicated `add_console_users_to_breeze_group` / `ensure_breeze_group` functions across `postinstall` and `install-darwin.sh` (pr-test-analyzer, medium). There is repo precedent (`scripts/build-edition_test.sh` wired into the required `test-agent` job) and the 3-way duplication is a real liability — but extracting a sourceable lib plus a stubbed-`PATH` harness is test-infrastructure work beyond this fix, and this PR is gating a customer fleet upgrade. Worth its own issue.
- The `ensureDarwinHelperIPCGroupMembership` loop is still darwin-gated and untested; its one piece of real logic (UID eligibility) is covered.

**Tests:** full agent suite green — `CGO_ENABLED=0 go test ./...`, 68 packages, 0 failures (one flake in `internal/monitoring`, a package this diff does not touch, passes 5/5 on re-run). `go test -race` green on `internal/sessionbroker`, `internal/heartbeat`, `internal/agentapp`. `go vet` + `go build` clean for darwin arm64/amd64, linux amd64, windows amd64 — the darwin-tagged files no CI job compiles were cross-vetted explicitly. `gofmt` clean; `shellcheck -S warning` clean on both installers.

**CI:** all agent jobs green, including **`Test Agent (race)` on `macos-latest`** — so the new coverage does execute on a real macOS runner, not just the ubuntu one. `Lint Agent (Go)` green after the errcheck fix. 38 checks passing, 0 failing, remainder still running.

**Mutation-checked, twice:** forcing `owner.GID = -1` in the socket path (pre-fix behaviour) fails `TestSetupSocketAppliesGroupAndRestrictiveMode`, `TestSetupSocketReplacesAStaleSocket` and `TestResolveSocketOwnerUsesLookedUpGID`. The socket tests also chown to a *supplementary* group rather than the process's primary GID — a socket is already created with the primary GID, so asserting that would have passed even with the bug present.

**One thing a maintainer must decide before promoting a build:** the socket-ownership half is deterministic and covered by real-syscall tests, but **membership taking effect for an already-logged-in user is not verified** — macOS resolves group membership through `opendirectoryd`, and I have no macOS host with the agent installed, so I cannot confirm whether a helper restarted mid-session picks up a group added after that GUI session began. Worst case it takes effect at next login, which is still strictly better than today (never works) and no worse than a `chmod` reapplied after every restart. A Standard-user login test on real hardware is the gate. See the PR description for the exact retest steps for the reporter.

Issue #3419, opened 2026-08-11
[API][RLS] Remote-access launcher resolves no providers for org-scoped callers — partner read is invisible under org scope


bdunncompany:
## Description

`resolveRemoteAccessLauncherForDevice` (`apps/api/src/routes/devices/core.ts:1073-1090`) reads `partners.settings` to find the tenant's configured remote-access providers. For an **organization-scoped** caller that read returns zero rows, so `partnerSettings` falls back to `{}`, no providers are found, and both the device-detail availability flag and `POST /devices/:id/remote-access-launch` degrade to "no provider configured" — for a tenant that has one configured.

This is the same partner-axis RLS gap already fixed once in `getEnrollmentDefaultsForOrg` (#2776 task 3.4), and the integration test written for that fix documents the chain exactly.

## The chain

Each link verified against `origin/main` (`f2d7fda83`):

1. **The route admits org scope.** `POST /:id/remote-access-launch` is `requireScope('organization', 'partner', 'system')` (`core.ts:1102`). The device-detail read path that computes the availability flag is reachable the same way (`core.ts:988-1013`).
2. **`partners` has exactly one SELECT policy**, `breeze_partner_isolation_select … USING (breeze_has_partner_access(id))` (`migrations/2026-04-11-partners-rls.sql:81`). I grepped every migration: there is no second permissive policy, and no own-partner branch via `breeze_current_partner_id` (that helper is used in the catalog and CIS-baseline policies, not here).
3. **`breeze_has_partner_access`** is `system → TRUE`, `NULL → FALSE`, else `id = ANY(breeze_accessible_partner_ids())` (`migrations/2026-04-11-a-rls-function-bootstrap.sql`).
4. **`computeAccessiblePartnerIds` returns `[]` for `scope === 'organization'`** (`middleware/auth.ts:382-389`) — `[partnerId]` only for partner scope, `null` only for system.

So an org-scoped session cannot see the partner row, and the join yields nothing.

## Why the `withSystemDbAccessContext` wrapper does not save it

The read is wrapped in `withSystemDbAccessContext` (`core.ts:1077`), which looks like it escalates. It does not, inside a request: `withDbAccessContext` early-returns when a context store already exists (`db/index.ts:440`), and `authMiddleware` has already opened one for the request. The nested call therefore **retains the caller's ambient scope** rather than switching to system.

[...]

Issue #3404, opened 2026-08-11
Per-organization remote-access provider settings


ToddHebebrand:
## Description

`remoteAccessProviders` is partner-wide only. The launcher walks device → org → partner and reads the partner row (`routes/devices/core.ts:1048`), and `EffectiveOrgSettings` carries security, notifications, eventLogs, defaults, branding and aiBudgets — not remote access. So one customer forbidding ScreenConnect while another requires it is not expressible today.

Two reasons this never gated the #3391 preference work: the audit trail already answers the compliance question (every launch writes a `device.remote_access_launch_url.issued` event with the provider ID, `core.ts:1131`), and neither the preference nor any picker widens the reachable set — both only select from what the partner enabled.

## Design constraints (settled during the #3391 review)

- **Orgs restrict, never add.** Store an `allowedProviderIds` list on the org, not copied provider objects; optionally an org default chosen from what remains. Letting orgs supply providers transfers control of templates and credentials and breaks the `partners.settings` decryption binding (see #3403).
- **It cannot ride `mergeCategory`.** Partner fields win wholesale in the merge (`services/effectiveSettings.ts:78-90`), so a partner providers array would lock orgs out entirely. Restriction is set intersection, not shallow field inheritance.
- **Don't expose it through the generic effective-settings endpoint** — that's gated only on org-read permission (`orgs.ts:1446`), the wrong exposure path for credential-adjacent config.
- **A slim provider-list read endpoint pays for itself three times.** Something returning only `{id, name, enabled}` per provider would serve: this feature's org-facing reads, the #3391 profile-UI dropdown (today the only source is GET /orgs/partners/me, which is partner-scope gated and returns the whole settings blob including encrypted provider passwords), and the availability check in #3402. Build it once here rather than three times.

Estimate: 3–5 days including RLS, contract tests, and UI.

## Affected Files

- `apps/api/src/routes/orgs.ts`
- `apps/api/src/routes/devices/core.ts`
- `apps/api/src/services/effectiveSettings.ts` (context — deliberately not used)

Found during review of #3391 / #3389. Related: #3401, #3402, #3403.

Issue #3402, opened 2026-08-11
[API] Device detail decrypts a remote-access provider password just to compute a boolean


ToddHebebrand:
## Description

The device-detail GET calls the full launcher resolution purely to set `hasRemoteAccessLauncher` (`routes/devices/core.ts:993`). That path decrypts the provider password and builds the complete substituted launch URL (`remoteAccessLauncher.ts:87-91`), then throws the result away. This runs on every device-detail page load.

Once #3391 lands, the same GET also picks up a `users` read per load (the per-technician preference lookup), stacking a second piece of work onto a path that only needs a yes/no.

## Root Cause

Availability checking and credential-bearing issuance share one code path. The GET only needs "would a launch URL resolve for this device" — provider exists, enabled, device has the custom-field identifier, template non-empty. None of that requires decryption or substitution.

## Proposed Fix

Split the launcher service into an availability check (no decryption, no substitution — stops after the identifier/template presence checks) and the existing issuance path. The GET calls the former; only `POST /devices/:id/remote-access-launch` touches credentials. The skip-reason vocabulary should be shared so the UI keeps its current distinctions.

A skew case to preserve: availability must evaluate the same provider selection (including the #3391 preference) as issuance, or the button could show for a provider the launch would then refuse.

## Affected Files

- `apps/api/src/services/remoteAccessLauncher.ts`
- `apps/api/src/routes/devices/core.ts`

Found during review of #3391 / #3389.

Issue #3401, opened 2026-08-11
[API] Remote-access provider list accepts duplicate IDs and a dangling defaultProviderId


ToddHebebrand:
## Description

The partner-settings validator for remote-access providers (`routes/orgs.ts:585-616`) validates each provider's fields and caps the array at 50, but nothing asserts provider IDs are distinct, and nothing asserts `defaultProviderId` names a provider that exists in the list. Both are hand-typed strings, so this is easy to hit.

## Impact

**Dangling default:** a `defaultProviderId` naming nothing silently resolves to `no_provider_configured` — the Connect button just doesn't appear, with no hint that the default points at nothing.

**Duplicate IDs:** resolution is a first-match `find` on the providers array (`remoteAccessLauncher.ts:63`), and each provider carries its own `urlTemplate` and `password`. Two entries sharing an ID means reordering the array silently changes which endpoint and which password get substituted, with no error anywhere.

**The two resolution paths disagree once #3391 lands:** the per-technician preference lookup filters on `enabled`, the tenant-default lookup doesn't (it takes the first ID match and then fails `provider_disabled`). With a duplicated ID where one copy is disabled, the same provider works via preference but errors via the default.

Not exploitable — only partner admins write this list, and the post-substitution scheme guard still applies to whichever template wins. It's a misconfiguration trap that would be painful to diagnose.

## Proposed Fix

Two refinements on the `remoteAccessProviders` schema in `routes/orgs.ts`: provider IDs must be unique, and `defaultProviderId`, when set alongside a provider list, must name a provider in it. Reject at write time so the bad state can't be saved.

## Affected Files

- `apps/api/src/routes/orgs.ts` (validator)
- `apps/api/src/services/remoteAccessLauncher.ts` (context only)

Found during review of #3391 / #3389. Fix PR incoming.

PR #3440, opened 2026-08-11
feat(remote): slim provider-list endpoint + preferred-remote-tool profile control (#3389)

### LINKED ISSUE #3389: [Remote] Per-technician remote-tool preference — defaultProviderId is tenant-wide, so techs cannot differ

bdunncompany:
Closes #3389. The profile UI you assigned on #3391, plus the slim provider endpoint it needs.

> **CI on this PR is red for reasons unrelated to the diff.** It branched from `main`, which currently cannot `pnpm install --frozen-lockfile` (duplicated `hono` key — fix is up as **#3436**, 6 deleted lines, green). Every job dies at install. It goes green once #3436 lands. Full gate results below are from a local run against a deduped lockfile; the lockfile is **not** part of this diff.

## 1. `GET /remote/providers`

Returns `{ providers: [{ id, name, enabled }], defaultProviderId }` and nothing else.

**The projection is the point.** `RemoteAccessProvider` also carries `urlTemplate`, `customFieldKey` and `password`, which live in the encrypted `partners.settings` blob. Today the only way to enumerate providers is `GET /orgs/partners/me`, which is partner-scope gated *and* returns that whole blob — the exposure path #3404 exists to avoid. The mapping is field-by-field rather than a spread, so a credential-bearing field added to the type later cannot silently start shipping.

**It lives in `routes/remote/`** because that router already applies exactly the right gate globally: `authMiddleware` + `remote:access` + `requireMfa`, with no partner-scope requirement. Same permission that guards issuing a launch URL — if you can launch, you can list. Added directly to `index.ts` rather than as a sub-router to stay clear of the `use('*')` mount-leak shape.

**It escalates past RLS, deliberately and narrowly.** `partners`' SELECT policy is `breeze_has_partner_access(id)`, and `computeAccessiblePartnerIds` returns `[]` for `scope === 'organization'` — so the technician who needs this cannot read the partner row under their own context. The read uses `runOutsideDbContext(() => withSystemDbAccessContext(...))`; a *bare* `withSystemDbAccessContext` would not work, since it early-returns inside a request and silently keeps the caller's scope. Escalating is only defensible because every caller projects through `toProviderSummaries` first, and the module comment says so.

That same RLS gap is why I filed **#3419** against the launcher's own partner read. `readPartnerRemoteAccessSettings` here is written to be the shared helper that fix can adopt, so the escalated read exists once rather than twice — but I have not touched the launcher, since you had not said whether you wanted that fix.

## 2. The profile control

`RemoteToolSettings`, mounted next to `ThemingSettings` in `ProfilePage`, writing `users.preferences.remoteAccessProviderId` through the existing `saveUserPreferences` → `PATCH /users/me` path. No migration, no new column.

Offers only **enabled** providers plus an explicit "Use organization default (*name*)". Choosing the default writes `undefined`, not `''`, so the server sees an absent preference.

**One deviation from your instruction, flagged rather than buried.** You said hide the control for non-partner sessions. That was premised on the only endpoint being partner-scope gated — this PR removes that constraint, so hiding it there would now hide it from exactly the org-scoped technicians it is for. Instead it gates on the **`remote:access` permission**, which is what the endpoint itself requires: a user who cannot launch a session is never offered a preference that could not take effect, and the fetch is skipped entirely rather than merely hidden.

Your underlying point — don't ship an empty select — is kept, and sharpened into three distinct states rather than one:

| State | UI |
|---|---|
| no `remote:access` | renders nothing, no request made |
| tenant has none configured | "No remote tools are configured for your organization yet." |
| request failed | "Couldn't load the available remote tools." |

Collapsing the last two into an empty list would tell a technician their org is misconfigured when the network hiccuped. A saved preference naming a since-disabled provider also gets an explicit line rather than vanishing from the select.

## 3. One adjacent edit, and why it was unavoidable

`ThemingSettings` typed its appearance state as `Required<UserPreferences>`, which silently meant "every preference" and stopped compiling the moment a non-appearance preference existed. Narrowed to a named `AppearancePreferences = Required<Pick<UserPreferences, 'theme' | 'density' | 'font' | 'timeFormat' | 'locale'>>` — the same five keys the file already enumerated on the next line. Behaviour identical; the type now says what it meant.

## Verification

**Two control runs**, one per security-relevant property:

- Replace the API projection with a spread → the two leak tests fail (exact key set, and a by-value check that the serialized response contains no password/template/field key).
- Remove the UI capability gate → the no-permission test fails, including its assertion that the endpoint is never called.

- **API:** `tsc --noEmit` exit 0; `vitest run` **1326 files / 21600 passed**; Trivy `secret,misconfig` HIGH/CRITICAL on `apps/api/src` clean.
- **Web:** `tsc --noEmit` exit 0; `vitest run` **554 files / 5322 passed** (includes `ThemingSettings`, so the type change above is covered).
- No schema, migration or dependency changes. 8 files, +515/−5.

## Not in scope

`#3402`'s availability check and `#3404`'s org-restriction work both want this endpoint; neither is touched here. `#3401` dangling defaults is yours in #3406 — this only declines to *report* a dangling default rather than fixing how one is created.

### DIFF
--- a/apps/api/src/routes/remote/index.ts
+++ b/apps/api/src/routes/remote/index.ts
@@ -1,6 +1,11 @@
 import { Hono } from 'hono';
-import { authMiddleware, requireMfa, requirePermission } from '../../middleware/auth';
+import { authMiddleware, requireMfa, requirePermission, type AuthContext } from '../../middleware/auth';
 import { PERMISSIONS } from '../../services/permissions';
+import {
+  readPartnerRemoteAccessSettings,
+  readRemoteAccessSettingsForPartner,
+  toProviderSummaries,
+} from '../../services/remoteAccessProviders';
 import { sessionRoutes } from './sessions';
 import { supportSessionRoutes } from './supportSessions';
 
@@ -10,8 +15,37 @@ export const remoteRoutes = new Hono();
 remoteRoutes.use('*', authMiddleware);
 remoteRoutes.use('*', requirePermission(PERMISSIONS.REMOTE_ACCESS.resource, PERMISSIONS.REMOTE_ACCESS.action), requireMfa());
 
+// GET /remote/providers - the tenant's remote-access providers, id/name/enabled
+// only.
+//
+// Lives here rather than under /orgs because the gate above is exactly right:
+// the same `remote:access` permission that guards issuing a launch URL, and no
+// partner-scope requirement. The only existing source is GET /orgs/partners/me,
+// which is partner-scope gated AND returns the whole settings blob including
+// encrypted provider passwords — the wrong exposure path for a control a
+// technician needs (#3404).
+//
+// Consumers: the profile "preferred remote tool" chooser (#3391/#3389), the
+// org-restriction reads in #3404, and the availability check in #3402.
+remoteRoutes.get('/providers', async (c) => {
+  const auth = c.get('auth') as AuthContext;
+
+  // A partner-scoped caller names its own partner; an org-scoped one does not
+  // (and its accessible-partner list is empty), so resolve through the org the
+  // same way the launcher walks device -> org -> partner.
+  const settings = auth.partnerId
+    ? await readRemoteAccessSettingsForPartner(auth.partnerId)
+    : auth.orgId
+      ? await readPartnerRemoteAccessSettings(auth.orgId)
+      : undefined;
+
+  // No resolvable partner (e.g. a system-scoped caller with no org context) is
+  // an empty directory, not an error: the caller simply has no tenant whose
+  // providers to list.
+  return c.json(toProviderSummaries(settings));
+});
+
 // Mount sub-routes
 remoteRoutes.route('/', sessionRoutes);
 // Quick Support inherits the same auth + remote:access + MFA gate above.
 remoteRoutes.route('/', supportSessionRoutes);
-
--- a/apps/api/src/services/remoteAccessProviders.test.ts
+++ b/apps/api/src/services/remoteAccessProviders.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it } from 'vitest';
+import { toProviderSummaries } from './remoteAccessProviders';
+
+// The projection is the security property of this whole feature: the endpoint
+// exists so a technician's provider chooser stops reading the partner settings
+// blob, which carries urlTemplate / customFieldKey / password (#3404).
+
+const FULL_PROVIDER = {
+  id: 'p1',
+  name: 'ScreenConnect',
+  urlTemplate: 'sc://connect/{{customField}}?pw={{password}}',
+  customFieldKey: 'sc_id',
+  password: 'super-secret',
+  enabled: true,
+};
+
+describe('toProviderSummaries', () => {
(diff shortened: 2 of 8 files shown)


bdunncompany:
Rebased onto `main` now that #3439 repaired the lockfile — **the CI caveat in the description no longer applies**, and this run is meaningful. `pnpm install --frozen-lockfile` verified against the rebased tree before pushing.

Nothing in the diff changed; only the base moved.

PR #3032, opened 2026-08-03
feat(extensions): tenant-scoped extension installs (L1 install scoping)

ToddHebebrand:
## Summary

- **Tenant-scoped extension installs (L1 install scoping):** an extension may declare `"installScope": "org"` in its manifest `tenancy` block; the host then only serves organizations with an enabled row in the new `extension_org_installs` table. Default stays `"server"` — every existing extension is unaffected, no data migration.
- **Enforcement at every dispatch surface on this branch:** an install-gate middleware inside both gateway snapshot wrappers (covering the canonical `/api/v1/ext/:extension/*` path AND the legacy `/api/v1/:routeNamespace/*` alias, for user/agent/helper), plus extension AI-tool execution (`executeTool`), MCP `tools/list` filtering, and an MCP `tools/call` pre-lifecycle gate. Non-installed orgs get 404 / the byte-identical `-32602 Unknown tool` envelope — default-deny, non-disclosing.
- **Jobs & management:** background jobs receive their tenant set via `context.tenancy.installedOrgs()` (host-side system-scope read; throws rather than returning a false-empty set — this subsumes the WS-2 Gate 0 self-elevation problem). Partner admins manage activations via `PUT/DELETE /api/v1/extensions/:name/orgs/:orgId` + `GET /api/v1/extensions/:name/orgs`, bounded by `accessibleOrgIds` with org-axis RLS as backstop.

## What this is NOT (load-bearing)

This is **L1 of a three-layer design: authorization, not containment**. In-process extension code can still read `POSTGRES_PASSWORD` and open its own connection (demonstrated in the Gate 0 probe). Only L3 (process isolation) changes the trust model; L2 (per-extension DB roles) and L3 are out of scope here. Nothing in this PR should be read as a sandbox claim.

## Deferred obligations (recorded, must not be lost)

1. **Client-ai surface gating (plan Task 5) is DEFERRED**: `routes/clientAi/ext.ts` lives on the unmerged `client-ext-seam-w4` branch. **The same install guard must be registered in that surface's wrapper when W4 lands — that surface must never merge ungated.** The `orgInstallGate.ts` docstring records this obligation; the full instructions live in the plan's Task 5.
2. **Web slice deferral (per spec):** web contribution filtering (nav/pages/slots), `/api/v1/extensions/registry`, and extension web-asset serving remain unfiltered by org — the spec explicitly defers these to the web slice.
[...]

PR #3455, opened 2026-08-11
feat(ee): merge Workspace into the monorepo as a built-in extension

ToddHebebrand:
## Summary

Moves the Breeze Workspace extension (formerly `LanternOps/breeze-workspace`,
private) into this repo at `ee/workspace/` (`@breeze/ext-workspace`), loaded
as a **built-in extension** — statically imported, compiled into the API
image, and registered at boot through the same staged v1 host pipeline
(contribution registry, default-deny gateway, tenancy/RLS tripwires) that
signed runtime bundles use, but with no signing, no artifact verification,
and no source-directory scan. This path is reserved for first-party code
that lives in this repository (currently just `ee/workspace`); third-party
extensions are unaffected and continue to arrive exclusively as signed
runtime bundles declared in `extensions.yaml`. `ee/` carries the Breeze
Commercial License; everything else in the repo remains AGPL-3.0. The two-repo
split, vendored-SDK tarballs, Ed25519 signing/packing pipeline, live
stock-host conformance suite, machine-local `pnpm.overrides` hack, and
symlinked `extensions/workspace` dev flow are all retired for Workspace — see
the deletion/retirement inventory below.

Spec: `docs/superpowers/specs/2026-08-11-workspace-ee-merge-design.md`
Plan: `docs/superpowers/plans/2026-08-11-workspace-ee-merge.md`

## Deletion / retirement inventory

- **`extensions/workspace` symlink dev flow** — retired. The symlink itself
  is untracked and gitignored (`extensions/*` in `.gitignore`, except
  `README.md`) in every checkout, including the main `~/breeze` checkout, so
  there was never anything for this branch to `git rm`. It does not exist at
  all in this worktree. See the Todd checklist below for removing it from
  disk in the main checkout.
- **Vendored-SDK tarballs / `.stubs/`** — deleted with the source move
  (Task 1/2); Workspace now depends on `@breeze/extension-sdk` /
  `@breeze/extension-web-sdk` via `workspace:*`, same as every other
  extension in this repo.
- **Packer / sign / stage / validate pipeline usage for Workspace** — retired.
  The signed-runtime-bundle platform itself is unchanged and stays in place
  for third-party extensions.
- **Live stock-host conformance suite** — retired as a Workspace gate. The
  platform's own conformance coverage (`apps/api/src/extensions/packerConformance.test.ts`)
  no longer needs Workspace as its reference guinea pig; `docs/extensions/build-time-transition.md`
  now points to that fixture instead.
- **Root `package.json` `pnpm.overrides` for vendored Workspace SDKs, and the
  compose-override `./package.json:/app/package.json` mount** — both were
  already outside git (a machine-local unstaged edit and an untracked compose
  override respectively), so neither is touched by this branch's diff. Left
  as a Todd checklist item (below) rather than editing his tree.
- **`LanternOps/breeze-workspace` (private repo)** — to be archived
  read-only as the history record once this PR merges; not part of this
  branch's diff.

## Built-in extension path (what actually changed under the hood)

- `apps/api/src/extensions/builtinExtensions.ts` (Task 3) statically imports
  the workspace package's parsed manifest and v1 register entry and feeds
  them through the same staged v1 pipeline as runtime bundles.
- `apps/api/src/index.ts` calls `loadBuiltinExtensions()` at boot, strictly
  between `loadSourceExtensions()` and the signed-bundle `reconcileExtensions()`
  call (Task 4); any built-in load failure aborts boot.
- Migrations: unchanged mechanism — the extension migrator runs
  `ee/workspace/migrations/*.sql`; the built-in registration hands it a
  resolved `migrationsDir`.
- **Three Dockerfiles** were found to need this wiring, not one:
  `docker/Dockerfile.api`, `apps/api/Dockerfile` (the actual published
  release image — `.github/workflows/release.yml` and
  `edge-image-api.yml` build from this file, not `docker/Dockerfile.api`),
  and `docker/Dockerfile.api.dev` (the hot-reload dev flow). All three now
  bake in `ee/workspace`'s manifest, `node_modules`, source/build output, and
  migrations, each honoring its own runtime `WORKDIR`/`process.cwd()`
  convention for where `builtinRegistry.ts`'s built-in-root resolution looks
  (`/app/ee/workspace` for the two images whose runner stays at `/app`;
  `/app/apps/api/ee/workspace` for `apps/api/Dockerfile`, whose runner ends
  at `WORKDIR /app/apps/api`). Each was verified with a real `docker build` +
  `docker run` smoke check (Task 5).
- `apps/api/tsup.config.ts`'s `noExternal` explicitly lists Workspace's two
  non-`@breeze/*` third-party runtime deps (`mailparser`, `v9u-smb2`) so
  their inline bundling is a documented, guarded contract rather than an
  implicit side effect of apps/api's own dependency list; a new static guard
  test (`apps/api/src/config/extensionBundledDependencies.test.ts`) catches
  the next genuinely-dangling dependency before it reaches a Docker build.
- CI (`.github/workflows/ci.yml`, Task 6): `ee/workspace` unit tests +
  typecheck run in the `test-api` job (no DB needed). The integration suite
  needs `workspace_*` tables, which only come from a real `loadBuiltinExtensions()`
  boot (nothing in the existing test harness — CI or local — ever ran that
  path before this branch); a new "Boot API once to apply built-in extension
  migrations" step runs `tsx src/index.ts` with `AUTO_MIGRATE=true` against
  the job's Postgres service, waits for the `loaded built-in "workspace"` log
  line (or a `CRITICAL` failure) with a 180s timeout, then stops it (own
  process group via `setsid`, bounded shutdown wait with a `kill -9`
  fallback) — before the `ee/workspace` integration-test step runs. Both new
  steps are gated `if: matrix.shard == 1` since the suite needs exactly one
  DB, not a sharded slice. **Note:** these steps have never executed in a
  real Actions run — watch this PR's first CI run before merging.

## Docs

- `docs/extensions/build-time-transition.md` — added a "Built-in (first-party)
  extensions" section documenting the new delivery mode as distinct from
  signed bundles and the deprecated source-directory path. Also reconciled
  now-stale claims found via
  `grep -rn "public SDK/host code only\|stock image" docs/extensions extensions/README.md docker/`:
  the old "stock image contains public SDK/host code only" line, and the
  file's two now-inaccurate `breeze-workspace`-as-reference-implementation /
  conformance-reference mentions (Workspace no longer takes the
  source-directory → signed-bundle path, so it no longer fits as that
  guide's worked example; the conformance reference is now
  `apps/api/src/extensions/packerConformance.test.ts`). `docker/Dockerfile.api`'s
  own stock-image comment was already corrected in Task 5.
- `extensions/README.md` — added a note that first-party extensions
  (`ee/workspace`) do not live in this directory; it is exclusively for
  third-party signed-bundle delivery.

## Verification

- `grep -rn "extensions/workspace" --include="*.ts" --include="*.yml" --include="*.yaml" apps packages docker .github | grep -v node_modules` — clean, no hits.
- `pnpm --filter @breeze/api exec vitest run src/extensions/ src/config/` — 32 files / 700 tests passed.
- `pnpm lint` — 6/6 packages, clean (includes `@breeze/ext-workspace`'s own `eslint src`).
- `pnpm test:docs-automation` (the doc-verify CI's guard) — 5/5 passed.
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit --project apps/api/tsconfig.json` — clean (Task 5's final fix round).

## Final whole-branch review

An independent final review (spec-vs-branch, cross-task seams, security, prod
readiness) returned "ready with fixes"; all three fixes landed in `422b38d2a`
(CI boot-step process-group kill + bounded wait + 180s window; `workspace_crawl_runs.cursor`
excluded from tenant export — it's an opaque agent resume token, matching the
sibling `crawl_cursor` exclusion; `ee/workspace/src` + `migrations` mounted
into both docker dev compose flows so in-container tsx sees ee edits).
Recommended follow-up issues (not blocking): (1) built-in `installed_extensions`
lifecycle polish — record `activeVersion` only after activation, drop the
redundant first-boot `setEnabled`, add a mid-pipeline-failure test; (2)
routeNamespace-level collision gate across all three delivery paths (name-only
today, pre-existing gap). One open design question for Todd: should
`ee/workspace/manifest.json` declare `tenancy.installScope: "org"` now that
tenant-scoped installs (#3032) exist, or stay server-scoped with org-gating via
`workspace_org_settings` as before? Current behavior matches pre-merge.

## Deployment note: Workspace is opt-in (`BREEZE_WORKSPACE_ENABLED`)

Built-ins now carry a deployment enable flag; Workspace's is `BREEZE_WORKSPACE_ENABLED`, **default OFF** (commit `19534184f`). With the flag unset, the API runs no workspace migrations, requires no pgvector, and requires no web bundle — any Postgres 16 image boots clean (one structured `builtin_extension_disabled` log line). Setting it to exactly `"true"` loads Workspace through the full pipeline, and **then** pgvector is required: a plain-postgres DB fails boot with an actionable error naming both remedies (pgvector image, or unset the flag). A previously-enabled deployment that turns the flag off keeps its tables safe — the disabled path publishes the tenancy declaration iff the tables exist, so cascades/exports/sweeps stay correct. Dev composes set the flag; the stock `POSTGRES_IMAGE_REF` default stays `pgvector/pgvector:pg16` so enabling later needs no DB swap. CI covers both paths: shard-1 integration boots flag-on; the compose Smoke Test and both binary-source jobs prove the default-off boot. Verified live in three boots: plain-pg flag-off, pgvector flag-on, and flag-off-with-existing-tables.

## Todd checklist

- [ ] Discard the machine-local, unstaged root `package.json` `pnpm.overrides`
      edit for vendored Workspace SDKs — dead now that Workspace resolves via
      `workspace:*` in-repo.
- [ ] Delete the untracked compose-override `./package.json:/app/package.json`
      mount — dead for the same reason.
- [ ] Remove the `extensions/workspace` symlink from the main `~/breeze`
      checkout if still present (`rm extensions/workspace`) — it is untracked
      and gitignored, so nothing in this PR removes it automatically; it just
      becomes a stale, unused symlink once this merges.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -824,6 +824,21 @@ MANAGED_SOFTWARE_POLICY_MODE=compat
 # BILLING_SERVICE_API_KEY=
 
 # --------------------------------------------
+# Built-in extensions (optional, OFF)
+# --------------------------------------------
+# First-party extensions ship inside the API image but are NOT loaded unless the
+# deployment switches them on, so a deployment that does not want one pays
+# nothing for it — no migrations, no schema, no infrastructure requirements.
+# Each flag is strict: only the exact string "true" enables it.
+#
+# Workspace (file/email indexing, content search). Enabling it requires a
+# PostgreSQL image with the pgvector extension available — its migrations run
+# `CREATE EXTENSION vector` on API boot, and a stock postgres:16-alpine will
+# fail the boot with `extension "vector" is not available`. The
+# POSTGRES_IMAGE_REF default below already includes it.
+# BREEZE_WORKSPACE_ENABLED=true
+
+# --------------------------------------------
 # Docker Deployment
 # --------------------------------------------
 # Stable Compose project name for container labels and Docker named volumes.
@@ -852,10 +867,11 @@ BREEZE_BINARIES_IMAGE_REF=ghcr.io/lanternops/breeze/binaries:${BREEZE_VERSION}
 # (e.g. `caddy@sha256:<digest>` from `docker buildx imagetools inspect caddy:2-alpine`).
 CADDY_IMAGE_REF=caddy:2-alpine
 # pgvector/pgvector is the official Postgres image + the vector extension
-# pre-installed. Required as of the Workspace built-in, whose migrations
-# (ee/workspace/migrations/*.sql) run `CREATE EXTENSION vector` on API boot
-# for content embeddings. Same Postgres major version as stock, just with
-# the extension baked in.
+# pre-installed. Required only when BREEZE_WORKSPACE_ENABLED=true (above),
+# whose migrations (ee/workspace/migrations/*.sql) run `CREATE EXTENSION
+# vector` on API boot for content embeddings. Same Postgres major version as
+# stock, just with the extension baked in — kept as the default so enabling
+# Workspace later needs no database swap.
 POSTGRES_IMAGE_REF=pgvector/pgvector:pg16
 REDIS_IMAGE_REF=redis:7-alpine
 COTURN_IMAGE_REF=coturn/coturn:latest
--- a/.env.example
+++ b/.env.example
@@ -872,6 +872,18 @@ CADDY_IMAGE_REF=caddy:2-alpine
 # vector` on API boot for content embeddings. Same Postgres major version as
 # stock, just with the extension baked in — kept as the default so enabling
 # Workspace later needs no database swap.
+#
+# UPGRADE CAVEAT — do NOT swap this image over an EXISTING data volume.
+# postgres:16-alpine is built against musl; pgvector/pgvector:pg16 is built
+# against glibc. The two disagree about collation versions, so starting one over
+# a data directory initialized by the other logs a collation version mismatch
+# and leaves every text index (and every unique constraint on a text column)
+# potentially inconsistent until it is rebuilt — which can silently return wrong
+# rows, not just fail loudly. If you are already running and want to switch,
+# either dump and restore into a freshly-initialized volume, or run
+# `REINDEX DATABASE <dbname>;` after the first start on the new image (and
+# `ALTER DATABASE <dbname> REFRESH COLLATION VERSION;`). If you do not need
+# Workspace, the simplest answer is to keep the image you already have.
 POSTGRES_IMAGE_REF=pgvector/pgvector:pg16
 REDIS_IMAGE_REF=redis:7-alpine
 COTURN_IMAGE_REF=coturn/coturn:latest
--- a/.env.example
+++ b/.env.example
@@ -851,7 +851,12 @@ BREEZE_BINARIES_IMAGE_REF=ghcr.io/lanternops/breeze/binaries:${BREEZE_VERSION}
 # Breeze is tested against. Override to digest-pinned refs in production
 # (e.g. `caddy@sha256:<digest>` from `docker buildx imagetools inspect caddy:2-alpine`).
 CADDY_IMAGE_REF=caddy:2-alpine
-POSTGRES_IMAGE_REF=postgres:16-alpine
+# pgvector/pgvector is the official Postgres image + the vector extension
+# pre-installed. Required as of the Workspace built-in, whose migrations
+# (ee/workspace/migrations/*.sql) run `CREATE EXTENSION vector` on API boot
+# for content embeddings. Same Postgres major version as stock, just with
+# the extension baked in.
+POSTGRES_IMAGE_REF=pgvector/pgvector:pg16
 REDIS_IMAGE_REF=redis:7-alpine
 COTURN_IMAGE_REF=coturn/coturn:latest
 #
--- a/.github/workflows/ci-smoke-binary-source-github.yml
+++ b/.github/workflows/ci-smoke-binary-source-github.yml
@@ -41,10 +41,12 @@ jobs:
     timeout-minutes: 10
 
     services:
-      # pgvector-enabled image required: this job boots the API directly
-      # (`pnpm --filter=@breeze/api start`), which now runs the Workspace
-      # built-in's ee/workspace/migrations/*.sql (CREATE EXTENSION vector)
-      # on boot. Stock postgres:16-alpine lacks the extension.
+      # pgvector-enabled image, though this job no longer NEEDS it: built-in
+      # extensions load only when their enable flag is set, and this job
+      # leaves BREEZE_WORKSPACE_ENABLED unset, so ee/workspace/migrations/
+      # *.sql (CREATE EXTENSION vector) never runs. Kept to match the image
+      # .env.example ships, so the boot proved here is the one self-hosters
+      # run — a stock postgres:16-alpine would boot identically.
       postgres:
         image: pgvector/pgvector:pg16
         env:
@@ -87,13 +89,10 @@ jobs:
       - name: Build API
         run: pnpm --filter=@breeze/api build
 
-      - name: Build workspace web bundle
-        # The built-in loader fails production boot hard when
-        # ee/workspace/dist/web is missing (a real from-source deploy must
-        # build it — see docker/Dockerfile.api's builder stage, which runs
-        # this same command right after the API build). This job boots with
-        # NODE_ENV=production from a plain checkout, so it must build it too.
-        run: pnpm --filter @breeze/ext-workspace build:web
+      # NOTE: no `ee/workspace` web build here, deliberately — see the same
+      # note in ci-smoke-binary-source-local.yml. This job leaves
+      # BREEZE_WORKSPACE_ENABLED unset, so the built-in never loads and its
+      # dist/web is never required.
 
       - name: Boot API in BINARY_SOURCE=github mode
         env:
--- a/.github/workflows/ci-smoke-binary-source-github.yml
+++ b/.github/workflows/ci-smoke-binary-source-github.yml
@@ -87,6 +87,14 @@ jobs:
       - name: Build API
         run: pnpm --filter=@breeze/api build
 
+      - name: Build workspace web bundle
+        # The built-in loader fails production boot hard when
+        # ee/workspace/dist/web is missing (a real from-source deploy must
+        # build it — see docker/Dockerfile.api's builder stage, which runs
+        # this same command right after the API build). This job boots with
+        # NODE_ENV=production from a plain checkout, so it must build it too.
+        run: pnpm --filter @breeze/ext-workspace build:web
+
       - name: Boot API in BINARY_SOURCE=github mode
         env:
           # Authenticate binarySync's github API call so we don't hit the
--- a/.github/workflows/ci-smoke-binary-source-github.yml
+++ b/.github/workflows/ci-smoke-binary-source-github.yml
@@ -41,8 +41,12 @@ jobs:
     timeout-minutes: 10
 
     services:
+      # pgvector-enabled image required: this job boots the API directly
+      # (`pnpm --filter=@breeze/api start`), which now runs the Workspace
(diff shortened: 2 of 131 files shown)


ToddHebebrand:
CI status: everything is green (Smoke Test, Test API, all four integration shards incl. the new workspace boot-step + integration suite, Type Check, and the CI Success aggregator) except **BINARY_SOURCE=github download path**, which is failing identically on `main` (last four main runs today are red on unrelated selfhost-signing commits — see the workflow history). On this branch that job now gets *further* than main's failure mode: the API boots cleanly through the built-in workspace load; the 404 is the agent-binary GitHub release manifest lookup, untouched by this PR.

ToddHebebrand:
## Post-review hardening (comprehensive multi-agent review of `3757b9ee2..HEAD`)

Four specialized review passes (code, silent-failures, test coverage, comment accuracy) ran over the commits that landed after the original whole-branch review. Outcome, all fixed in `49c60172f` + `588e8a459` and verified by a scoped re-review plus this green CI run:

**Critical (fixed):** a disabled built-in with a *partially* migrated schema (e.g. enable on plain Postgres → dies at the pgvector migration → operator follows the error's advice and disables) published its FULL tenancy declaration — org-deletion cascades and tenant export would then hit nonexistent tables, aborting GDPR erasure mid-delete. The disabled path now probes which tables exist and publishes a declaration filtered to them, logging a loud `builtin_extension_partial_schema` warning with present/missing lists and remedies. Reproduced live before and after the fix.

**Also fixed:** RLS tripwire now runs on the disabled path over the published declaration (previously, disabling the flag silently retired the only RLS check those tables get); CI boot-step `CRITICAL` greps narrowed to the fatal sentinel; the "API exited on its own" branch now fails on a real startup failure; PGID handling fails closed — and its first live run caught that `kill -0` counts zombies, so the shutdown poll now watches the group and treats zombie-only groups as dead; probe/manifest failures carry actionable context; typo'd flag values (`True`, `1`) are named in the skip log; web-asset error handling narrowed to only soft-fail on "never built"; dev composes bind-mount the manifest and parameterize the flag; `deploy/` production path documents the flag and receives it; musl→glibc collation/REINDEX upgrade caveat added to .env.example and the deploy docs; new DB-backed integration test pins the real table-probe SQL on every PR.

**Deliberate follow-ups (not in this PR):** lazy `BUILTINS` loading so a disabled deployment tolerates a missing/corrupt manifest entirely; a two-builtin mid-pipeline-failure test; `installed_extensions` lifecycle polish (recordActive-after-activate, drop redundant first-boot setEnabled); routeNamespace-level collision gate across all three delivery paths; kill-signal ESRCH/EPERM discrimination in the CI step.

**Release-note item:** the disabled path now runs the RLS tripwire, so a deployment with orphaned built-in tables *lacking RLS* will refuse to boot where it previously booted silently under-protected — intentional fail-closed behavior.

ToddHebebrand:
Merged as `cfa16d03a` (squash). Post-merge follow-through:

- **Local cleanup checklist**: done — root `package.json` `pnpm.overrides` edit + matching `pnpm-lock.yaml` wiring discarded, `extensions/workspace` symlink removed. The compose-override `./package.json:/app/package.json` mount no longer existed anywhere (already gone).
- **Dev-boot check on merged main** (`BREEZE_WORKSPACE_ENABLED=true`, pgvector test DB): all migrations applied, `[extensions] loaded built-in "workspace" 0.2.0`, `/health` 200, all background workers up, zero CRITICAL lines.
- **Follow-up issues filed**: #3468 (installed_extensions lifecycle polish + two-builtin mid-pipeline-failure test), #3469 (routeNamespace collision gate across all three delivery paths), #3470 (lazy BUILTINS loading on the disabled path), #3471 (CI boot-step ESRCH/EPERM discrimination).

Still open: the `tenancy.installScope` ruling (issue-less for now — current server-scope + `workspace_org_settings` gating preserved), and archiving `LanternOps/breeze-workspace`.

ToddHebebrand:
⚠️ **Post-merge finding: this PR imported a stale snapshot.** Filed as #3478.

`ee/workspace` matches `LanternOps/breeze-workspace` @ `c5799345b` (2026-07-19), but that repo's `origin/main` is `4e1afda` (2026-08-01). Three squash-merged PRs are missing: W3 dashboard v1 + productized ingest (#19), W4 Outlook filing card (#20), and the content-layer graduation + Gate 0 verdict (#21) — **76 files, 10,449 insertions, plus the `2026-07-25-ingest-jobs.sql` migration**. `git grep` confirms none of it exists anywhere in breeze today.

Cause: the plan pinned `<WS_SHA>` from the local clone's HEAD without fetching first, and that clone was 3 commits behind. The import commit records the SHA accurately — it was just the wrong SHA to pin.

Nothing here is wrong with the *mechanism* this PR built (the built-in extension path, Docker wiring, CI boot-step, flag gating are all fine and verified) — the payload was just a three-week-old snapshot. The archive of `LanternOps/breeze-workspace` is **on hold** until the port lands.

ToddHebebrand:
## installScope ruling: Workspace stays server-scoped

Closing the open design question from the final review. `ee/workspace/manifest.json` keeps no `tenancy.installScope` key (resolving to the `"server"` default), and per-org control continues to come from `workspace_org_settings`. No code change.

The short version of why: flipping to `"org"` today would break Workspace for every org twice over. There is no backfill — `extension_org_installs` ships empty and nothing seeds it, so every existing org would 404 on agent crawls, Helper search and sources CRUD. And even for an *activated* org the admin UI would still 404, because the guard resolves the caller org from `auth.orgId`, which is null for the partner/system-scope tokens that `adminGate` requires; the deny fires before the route ever reads `?orgId`.

The two mechanisms are also orthogonal rather than alternatives — `workspace_org_settings.content_enabled` is a feature-level, default-off, self-enforced content gate, while `installScope` is host-enforced all-or-nothing dispatch authorization. You would end up needing both true, not one instead of the other.

Full analysis and the four prerequisites for ever revisiting this are recorded in #3480.

ToddHebebrand:
## Archived — the two-repo split is closed out

`LanternOps/breeze-workspace` is now **archived read-only** (final commit `f872ba6`, an archive banner on its README).

Sequenced behind the stale-import fix rather than done immediately: #3479 (`ed508310f`) ported the three missing commits first, so nothing was frozen unported. Verified before pulling the switch — **every file under `src/` and `migrations/` in that repo is present in `ee/workspace/`**, differing only in the documented adaptations (strict-`tsc` non-null fixes, the withheld `/client` mount, the legacy-manifest test removals, and the ee-only `manifestExportColumns` guard). Zero upstream-only files.

Also confirmed clean before archiving: all 11 of that repo's PRs were merged, and no unmerged branch work existed — the `git branch -r` "ahead" counts were squash-merge artifacts.

Open items from this PR's checklist are now all closed:

- ✅ Local cleanup (root `pnpm.overrides`, lockfile wiring, `extensions/workspace` symlink)
- ✅ Dev-boot verification with `BREEZE_WORKSPACE_ENABLED=true`
- ✅ installScope ruling — stays server-scoped (see the comment above; prerequisites in #3480)
- ✅ Repo archive with issue migration — 8 issues triaged to zero: #3472, #3473, #3474, #3475, #3476, #3477 migrated; the droplet-clone issue and the finder-UI issue closed as obsolete/shipped
- ✅ Follow-up issues filed: #3468, #3469, #3470, #3471

One thing left that is genuinely yours to decide, carried over from #3479: two upstream design docs were deliberately **not** ported into this public repo — `2026-07-28-extension-trust-boundary-finding.md` (a security finding whose own header states it lives in the private repo on purpose and reserves disclosure as an explicit decision) and its companion `2026-07-28-gate0-findings.md`. Both remain in the archived repo's history, readable but no longer editable. Flagging rather than deciding.

Issue #3475, opened 2026-08-12
workspace: local integration runs are not hermetic (shared :5433 test DB)


ToddHebebrand:
Migrated from the archived `LanternOps/breeze-workspace` repo (issue #14, filed 2026-07-20) as part of the ee/workspace merge (#3455). **Scope is reduced** — see below.

## Already resolved by the merge

- **Stale `../../.env.test` path** — fixed. The path assumed this code sat one level under the breeze repo root; standalone at `~/breeze-workspace` it wrongly resolved to `~/.env.test`. Now that the code lives at `ee/workspace`, the assumption is correct again and `ee/workspace/vitest.integration.config.ts` documents it explicitly.
- **CI hermeticity** — fixed. The suite now runs in breeze CI on integration shard 1 against the job's own Postgres service, not a shared box.

## What remains

Local `pnpm --filter @breeze/ext-workspace test:integration` still falls back to the shared `breeze_test` database on the `breeze-postgres-test` container (`:5433`), which other worktrees and sessions also use. Observed live: a run failed **12 tests** with `insert or update on table "workspace_sources" violates foreign key constraint ... Key is not present in table "organizations"` and `deadlock detected`; an immediate re-run passed 42/42 with no code change.

Practical guidance until fixed: do not trust a single red local integration run — re-run before investigating. Never tear the shared container down with `-v`.

Possible fix: the `breeze-ext` packer/signer work needed a hermetic two-replica test DB for the same reason; this suite likely wants the same treatment (per-worker schema or database).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3479, opened 2026-08-12
fix(ee/workspace): catch up to breeze-workspace main — W3 dashboard, W4 filing card, content graduation

### LINKED ISSUE #3478: ee/workspace was imported from a stale snapshot — W3 dashboard, W4 Outlook filing, and content graduation are missing
Found while executing the post-merge archive step for #3455, immediately before archiving `LanternOps/breeze-workspace`. **The archive is on hold** until this is resolved.

## What happened

`ee/workspace` is a faithful copy of `LanternOps/breeze-workspace` @ `c5799345b` (2026-07-19, that repo's PR #18) — but that repo's `origin/main` is `4e1afda` (2026-08-01). Three squash-merged PRs never made it into this monorepo:

| workspace SHA | date | PR | what |
|---|---|---|---|
| `f090145` | 2026-07-20 | #19 | W3 dashboard v1 + productized ingest (delta-triggered jobs, retry/backoff, `web.pages.v1` page) |
| `ad2222d` | 2026-08-01 | #20 | W4 Outlook filing card (client gate, tiered email matcher, filing panel) |
| `4e1afda` | 2026-08-01 | #21 | content layer graduated out of dev-preview + Gate 0 verdict |

**Delta: 76 files, 10,449 insertions**, including a missing migration (`2026-07-25-ingest-jobs.sql`) and these entirely absent modules:

`services/dashboardService.ts`, `services/emailMatchService.ts`, `services/ingestJobRunner.ts`, `services/ingestJobsService.ts`, `services/ingestErrors.ts`, `routes/client.ts`, `routes/clientGate.ts`, `routes/dashboard.ts`, `schema/ingestJobs.ts`, `web/dashboard.ts`, `web/filingCard.ts`

`git grep` for `dashboardService`, `emailMatchService`, `ingestJobRunner`, `ingest_jobs`, `filingCard` across all of breeze `HEAD` returns **0 tracked files** — none of it was hand-ported elsewhere. Both manifests read `"version": "0.2.0"`, so the version number does not distinguish the two states.

## Root cause

The merge plan (`docs/superpowers/plans/2026-08-11-workspace-ee-merge.md`, lines 19 and 48) instructed pinning `git rev-parse HEAD` from the local `/Users/toddhebebrand/breeze-workspace` clone as `<WS_SHA>`, with no `git fetch` beforehand. That clone was 3 commits behind `origin/main`, so the import faithfully copied a stale HEAD. The import commit message (`656b810bf`) records `c5799345b` accurately — the SHA was simply the wrong one to pin.

## Good news: it is a clean stale import, not a partial one

Comparing tracked file lists per candidate SHA against `ee/workspace`: `c579934` differs by 1 path, `f090145` by 17, `ad2222d`/`4e1afda` by 27. A content diff of the shared files at `c579934` shows only 9 files differing, and all nine are merge-mechanical (non-null assertions plus explanatory comments from the strict-`tsc` pass).

So the fix is tractable and well-scoped: replay the three commits onto `ee/workspace`, re-applying the same adaptations the original import made (vendored SDK tarballs → `workspace:*` deps, strict-`tsc` fixes), add the missing migration, and re-run the `ee/workspace` unit + integration suites and the docker guard tests. Essentially merge-plan Tasks 1–2 again at the newer SHA.

## Also worth deciding

Whether all three land together or W3/W4 are staged separately — #21 in particular graduates the content layer out of dev-preview, which is a behavior change for the `WORKSPACE_CONTENT_PREVIEW` gating.

## Not at risk

All 11 `breeze-workspace` PRs are merged and no unmerged branch work exists (the `git branch -r` "ahead" counts are squash-merge artifacts). Archiving is read-only and would lose nothing — the objection is only that archiving now would signal "done" while 10.4k lines of merged mainline work live solely in a frozen repo.

ToddHebebrand:
Closes #3478.

## Why

The ee/workspace merge (#3455) imported `LanternOps/breeze-workspace` @ `c5799345b` — three commits behind that repo's `origin/main` (`4e1afda`). The merge plan pinned `<WS_SHA>` from a local clone's `git rev-parse HEAD` **without fetching first**, and that clone was stale. The import commit recorded the SHA accurately; it was just the wrong SHA to pin.

Nothing was wrong with the *mechanism* #3455 built — the built-in extension path, Docker wiring, CI boot-step and flag gating are all fine. The payload was a three-week-old snapshot.

## What this ports

| upstream | PR | what |
|---|---|---|
| `f090145` | #19 | W3 dashboard v1 + productized ingest — delta-triggered jobs with retry/backoff |
| `ad2222d` | #20 | W4 Outlook filing card — client gate, tiered email matcher, filing panel |
| `4e1afda` | #21 | content layer graduated out of dev-preview + Gate 0 verdict |

New modules: `ingestJobsService`, `ingestJobRunner`, `ingestErrors`, `schema/ingestJobs`, `routes/dashboard`, `routes/client`, `routes/clientGate`, `services/dashboardService`, `services/emailMatchService`, `web/dashboard`, `web/filingCard`. New migration: `2026-07-25-ingest-jobs.sql`.

## What is deliberately NOT ported

The upstream delta also touched `.gitignore`, `README.md`, `breeze-extension.json`, `dev/`, `docs/superpowers/` and `test/conformance/` — every one of those is a category #3455's retirement inventory dropped on purpose, so carrying them now would undo that decision.

Two upstream design docs in that set stay out of this **public** repo deliberately: `2026-07-28-extension-trust-boundary-finding.md` is a security finding whose own header states it lives in the private repo on purpose and reserves disclosure as an explicit decision, and `2026-07-28-gate0-findings.md` is the probe writeup it references. Both remain in the archived repo's history. **Flagging for a disclosure call rather than making one.**

## Import adaptations re-applied at the newer SHA

- **`orgExportColumns` is an ee-merge addition upstream never had.** A wholesale manifest copy silently dropped it; the `manifestExportColumns` guard test caught that immediately. Restored, and extended with a `workspace_ingest_jobs` entry — `cursor` excluded as an opaque resume token, matching the sibling `workspace_crawl_runs` exclusion.
- **`apps/api`'s tsconfig is stricter than `ee/workspace`'s own** (`noUncheckedIndexedAccess`). The ported code needed the same non-null adaptations #3455 applied to five files, plus four new sites in `emailMatchService`. Worth knowing: `pnpm --filter @breeze/ext-workspace typecheck` passes *without* them — the apps/api project typecheck is the real gate, because ee/workspace compiles into its bundle.
- **The dropped legacy `breeze-extension.json` is no longer importable**, so the tests pinning it were removed. Two covered W4's `clientSurfaces`/`clientPanels`; their negative halves (the frozen `.strict()` v1 schema rejects both keys) are kept. The `helperRoutes` guarantee those tests protected now lives as an explicit field on the `BUILTINS` entry in `builtinRegistry.ts`.

## Known gap: `/client/*` has no core-side dispatcher

W4's client surface is ported but **not reachable**. Core's generic client-ai session proxy and the loader support that carries `clientSurfaces`/`clientPanels` are unmerged — four commits sitting on `ToddHebebrand/client-ext-seam-w4` (`701cc541a`, `55db2327a`, `e9860b560`, `d9ffc0131`). `clientGate` demands an organization-scoped context, so the surface fails closed until that platform work lands. Porting it now keeps ee/workspace a faithful mirror of the archived repo; activating it is a separate decision about which of those four commits serve the third-party platform generally.

## Verification

- `ee/workspace`: **586/586** unit (was 424), **151/151** integration (was 112)
- `apps/api` extensions+config: **731/731**, including the `dockerfileWorkspaceManifests` and `extensionBundledDependencies` guards
- Strict `tsc --project apps/api/tsconfig.json`: clean
- `eslint src`: clean; web bundle builds (576 KB)
- **Live flag-on boot** against pgvector: `2026-07-25-ingest-jobs.sql` applies incrementally on an existing schema (RLS enabled+forced, 4 org-isolation policies, both indexes incl. the partial unique index), `[extensions] loaded built-in "workspace" 0.2.0`, `/health` 200, zero CRITICAL lines
- `apps/api` `builtinTableProbe.integration.test.ts` passes with the 14-table declaration. `twoReplicaReconcile.integration.test.ts` timed out once and passed on re-run — the known shared-`:5433` non-hermetic flakiness (#3475), unrelated to this change.

### DIFF
--- a/apps/api/src/extensions/builtinExtensions.test.ts
+++ b/apps/api/src/extensions/builtinExtensions.test.ts
@@ -852,6 +852,20 @@ describe('BUILTIN_EXTENSION_NAMES', () => {
       else process.env.BREEZE_WORKSPACE_ENABLED = previous;
     }
   });
+
+  /**
+   * Workspace's /helper/* tree is called by the device helper, so it must sit
+   * behind core helper auth rather than the user default-deny. On the built-in
+   * path that boundary is declared ONLY by this field — the legacy
+   * breeze-extension.json the upstream repo asserted it from is not part of
+   * this delivery mode. Dropping the field would silently move helper traffic
+   * to authMiddleware (see the gateway's helperRoutes arm), so it is pinned
+   * against the real registry entry, not a fixture.
+   */
+  it('keeps workspace /helper/* behind core helper auth', () => {
+    const workspace = BUILTINS.find((builtin) => builtin.manifest.name === 'workspace');
+    expect(workspace?.helperRoutes).toBe(true);
+  });
 });
 
 /**
--- a/ee/workspace/manifest.json
+++ b/ee/workspace/manifest.json
@@ -12,8 +12,8 @@
   "server": { "entry": "server/index.cjs" },
   "web": {
     "entry": "web/index.js",
-    "pages": [{ "id": "sources", "path": "/extensions/workspace/sources", "element": "workspace-sources-page" }],
-    "navigation": [{ "id": "sources", "label": "Workspace Sources", "path": "/extensions/workspace/sources", "order": 100 }],
+    "pages": [{ "id": "sources", "path": "/extensions/workspace/sources", "element": "workspace-sources-page" }, { "id": "dashboard", "path": "/extensions/workspace/dashboard", "element": "workspace-dashboard" }],
+    "navigation": [{ "id": "sources", "label": "Workspace Sources", "path": "/extensions/workspace/sources", "order": 100 }, { "id": "dashboard", "label": "Workspace Dashboard", "path": "/extensions/workspace/dashboard", "order": 50 }],
     "slots": [{ "id": "device-index", "slot": "device.detail.tabs", "contractVersion": 1, "label": "Workspace", "element": "workspace-device-index", "order": 100 }]
   },
   "migrationsDir": "migrations",
@@ -23,7 +23,7 @@
   "jobs": [],
   "aiTools": [],
   "tenancy": {
-    "orgCascadeDeleteTables": ["memory_blocks", "workspace_content_chunks", "workspace_content_entities", "workspace_crawl_runs", "workspace_email_filings", "workspace_file_activity", "workspace_file_content", "workspace_file_enrichment", "workspace_file_index", "workspace_org_settings", "workspace_project_crosswalk", "workspace_projects", "workspace_sources"],
+    "orgCascadeDeleteTables": ["memory_blocks", "workspace_content_chunks", "workspace_content_entities", "workspace_crawl_runs", "workspace_email_filings", "workspace_file_activity", "workspace_file_content", "workspace_file_enrichment", "workspace_file_index", "workspace_ingest_jobs", "workspace_org_settings", "workspace_project_crosswalk", "workspace_projects", "workspace_sources"],
     "deviceCascadeDeleteTables": ["workspace_file_index"],
     "deviceOrgDenormalizedTables": ["workspace_file_activity"],
     "deviceOrgMoveDeleteTables": ["workspace_file_index"],
@@ -63,6 +63,10 @@
       "workspace_file_index": {
         "include": ["id", "org_id", "source_id", "device_id", "device_key", "rel_path", "parent_path", "name", "is_dir", "ext", "mime", "size", "mtime", "ctime", "attrs", "last_seen_at", "deleted_at", "created_at", "updated_at"],
         "exclude": []
+      },
+      "workspace_ingest_jobs": {
+        "include": ["id", "org_id", "source_id", "crawl_run_id", "trigger", "phase", "status", "force", "attempts", "max_attempts", "next_attempt_at", "stats", "last_error", "started_at", "finished_at", "created_at", "updated_at"],
+        "exclude": ["cursor"]
       },
       "workspace_org_settings": {
         "include": ["org_id", "content_enabled", "created_at", "updated_at"],
--- a/ee/workspace/migrations/2026-07-25-ingest-jobs.sql
+++ b/ee/workspace/migrations/2026-07-25-ingest-jobs.sql
@@ -1,0 +1,62 @@
+-- W3: productized ingest — delta-triggered incremental ingest job state.
+-- One live job per (org, source-partition); advancement is in-request batch
+-- continuation (see src/services/ingestJobRunner.ts). RLS shape 1; registered
+-- in breeze-extension.json orgCascadeDeleteTables.
+-- autoMigrate wraps this file in a transaction: CREATE TYPE is safe here
+-- (only ALTER TYPE ADD VALUE is not — see 2026-07-24-org-settings-dlp.sql:23-26).
+
+DO $$ BEGIN
+  CREATE TYPE workspace_ingest_trigger AS ENUM ('crawl_complete', 'manual', 'reingest');
+EXCEPTION WHEN duplicate_object THEN NULL; END $$;
+
+DO $$ BEGIN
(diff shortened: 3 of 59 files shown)


ToddHebebrand:
## Review pass + fixes (`e38412c20`)

An independent review of `22a76ba1f` confirmed port fidelity — 47 of the 57 delta files byte-identical to upstream, the other 10 exactly the documented adaptations, `orgExportColumns` restored without corruption, the new migration's RLS/FK/partial-index shape correct, and 6 of the 7 non-null assertions structurally sound. Three findings, all fixed:

**HIGH — `/client/*` was reachable by any organization-scoped Breeze user.** Upstream could mount it safely because the only caller was core's client-ai proxy, dispatching under its own `/api/v1/client-ai/ext/:extension/*` mount with a synthesized org context. That proxy is unmerged here, so the extension gateway routed `/client/*` down its **default arm** — an ordinary browser session JWT. `clientGate` checks the *shape* of the auth context, not its provenance, so a normal org-scoped user satisfied it. That made `/client/*` the only Workspace surface reachable at organization scope at all (`adminGate` rejects that scope outright), exposing mail-metadata search, a write-on-GET via `filingService.classify`, and filing decisions overwritable by any org user. Same-org only — an intra-org authorization gap, not a tenant-isolation break — but not something to ship.

**The surface is now not mounted.** Worth recording why I did *not* take the review's suggested fix of gating `clientGate` on `principal.kind === 'client_user'`: I checked it against the unmerged proxy first, and its synthesized context sets **no `principal` field at all** (zero occurrences in `701cc541a`), so that check would leave the surface dead the moment the proxy lands. Withholding the mount is provably safe now and constrains nothing about the eventual design.

The prefix keeps a bare catch-all — without one, `/client/*` falls through to the admin-gated user routes and answers their `adminGate` 403, which leaks that the prefix is unhandled and leaves the tree one routing change from the admin surface. The four upstream tests that drove `clientGate` through the mount are replaced by one pinning what matters while the surface is withheld: every `/client/*` path 404s for every principal, and nothing falls through. `clientGate`'s own logic keeps direct coverage in `clientGate.test.ts`.

**MEDIUM — the `helperRoutes` guarantee was asserted nowhere.** My commit claimed the deleted legacy-manifest assertion "lives" on the `BUILTINS` entry. The field is there; nothing checked its value. Deleting it would silently move `/helper/*` from `helperAuth` to `authMiddleware` with nothing failing. Now pinned against the real registry entry, which makes that comment true as written.

**LOW — `vectors[i]!` asserted a guarantee that does not hold.** Unlike the other six adaptations, presence depends on a remote API contract; a short or misindexed embedder response leaves holes and throws a `TypeError` inside the ingest transaction. Not a regression (upstream crashes identically), but the `!` silenced the one place tsc was pointing at a real hole. Now length- and hole-checked, raising `TransientIngestError` so the batch aborts and the file stays pending rather than being parked as `failed`.

Re-verified: ee/workspace **585/585** unit + **151/151** integration; apps/api extensions+config **732/732**; strict apps/api `tsc` clean; lint clean.

bdunncompany:
Quick note so the red check doesn't read as worse than it is: the **Gitleaks failure on `e38412c20` is a transient install failure, not a finding.** The scanner never ran.

From the job log, the failing step is `Install gitleaks`:

```
curl -sSfL -o "$workdir/$asset" "$base_url/$asset"
curl: (56) Connection died, tried 5 times before giving up
##[error]Process completed with exit code 56.
```

It failed 9 seconds in, downloading `gitleaks_8.21.2_linux_x64.tar.gz` from the GitHub release. A re-run should clear it.

Worth flagging separately: the install step pins the version and verifies the checksum, but a download failure and a real leak both surface as the same red "Gitleaks" check. On a PR that ports 58 files of `ee/`-licensed code, that's the one place where "red for an infrastructure reason" is most likely to be misread as "red for a secret."

Issue #2634, opened 2026-07-19
security: unreserved route namespaces (service-principals, partner-service-principals, partner-api) allow extension hijack


ToddHebebrand:
### Summary

Three core, auth-sensitive `/api/v1` route mounts are **not** listed in `RESERVED_ROUTE_NAMESPACES`, so an installed + enabled runtime extension can declare one of them as its `routeNamespace` and hijack that core surface fleet-wide. Discovered while adding Plan 03's `/api/v1/extensions` mount (which *was* reserved).

### Affected mounts

In `apps/api/src/index.ts` (~lines 857-859):

- `service-principals`
- `partner-service-principals`
- `partner-api`

None appear in `RESERVED_ROUTE_NAMESPACES` (`packages/extension-sdk/src/manifest.ts`) or its mirror in `packages/extension-api/src/legacy.ts`.

### Mechanism

The extension gateway registers `app.all('/api/v1/:routeNamespace/*', dispatchAlias)` (via `mountExtensionGateway`) **before** the core sub-routers are mounted. `dispatchAlias` only falls through (`next()`) when `registry.getByRouteNamespace(routeNamespace)` returns nothing. The manifest schema only rejects a `routeNamespace` that is in `RESERVED_ROUTE_NAMESPACES`. So an extension declaring `routeNamespace: 'service-principals'` (or the other two) passes validation, and once installed + enabled, requests to `/api/v1/service-principals/*` dispatch into the extension's own route app instead of the core handler.

This is the same vulnerability class fixed for `extensions` in the Plan 03 branch (reserve the namespace + regenerate the `CORE_NAMESPACES` contract test).

### Impact

Any account able to install/enable an extension can shadow core service-principal / partner-API endpoints — a privilege/redirection risk on auth-sensitive surfaces. Pre-existing on `main`, independent of Plan 03.

### Fix

1. Add `'service-principals'`, `'partner-service-principals'`, `'partner-api'` to `RESERVED_ROUTE_NAMESPACES` in **both** `packages/extension-sdk/src/manifest.ts` and `packages/extension-api/src/legacy.ts` (they must stay in sync).
2. Regenerate `CORE_NAMESPACES` in `packages/extension-api/src/index.test.ts` (see the sibling issue about auto-deriving it) so the contract test covers all current mounts.
3. Add a manifest-validation test that each of the three is now rejected as a `routeNamespace`.

### Verification

[...]

PR #3488, opened 2026-08-12
refactor(extensions): retire the two dead delivery paths

ToddHebebrand:
**−15,841 / +835 across 132 files.** Removes the legacy source-directory loader and the signed-runtime-bundle pipeline. The built-in path survives untouched — Workspace still loads behind `BREEZE_WORKSPACE_ENABLED` and all its suites are green.

## Why

There is no third-party extension story, so the machinery that exists to support one is attack surface without a user. Between them, these two paths were the *only* code in the platform that loads and executes third-party code in-process: read `extensions.yaml` from a mounted directory → extract an artifact → verify a signature → `require()` the result, with `process.env` (including `POSTGRES_PASSWORD`) and the database handle in scope. The trust model is "signed implies fully trusted" — sound if you want third parties, pure cost if you don't.

They were provably dead: no `extensions.yaml` anywhere in the repo, no `breeze-extension.json` anywhere, the `extensions/` drop directory held only a README, the legacy loader sat behind `BREEZE_LEGACY_SOURCE_EXTENSIONS` (default off), and the single e2e signed-bundle spec was never wired into a workflow.

## What survives

**The built-in path, unchanged.** `ee/workspace` loads at boot behind its enable flag, mounts the same routes, runs the same migrations through the same migrator, and serves the same web contributions. Also kept: `extension-sdk`, `extension-web-sdk`, `extension-testkit` (Workspace's own tests import it), `webAssets.ts`, `migrator.ts`, the gateway, the contribution registry, the tenancy registry and tripwire, and `installed_extensions` / `extension_schema_history`.

## Deleted

- **Legacy source path** — `loader.ts`, `discovery.ts`, `packages/extension-api`, and the `autoMigrate` seam that scanned for extension directories.
- **Signed bundles** — `reconciler.ts`, `bundleVerifier.ts`, `artifactStore.ts`, `config.ts`, `trust.ts`, `packerConformance`, the two-replica reconcile fixture and test, `packages/extension-cli`, the e2e fixture and spec, and the drop directory.
- **`apps/api/scripts/breezectl`** — the entire CLI. Worth a look: `breezectl.lib.ts` rejected every noun but `extensions`, so the whole tool was extension operations. `list`/`doctor`/`enable`/`disable` are still available as the `/api/v1/extensions` admin routes. **Object here if you wanted the CLI kept.**
- **Dead weight that could never fire** — `jobHost.ts` (Workspace declares `"jobs": []`), the extension AI-tool branches (`"aiTools": []`), and the per-org install gate plus `extension_org_installs` (Workspace is server-scoped, so no row could ever exist). A new migration drops the table; the original `2026-08-10` file is left immutable.
- **`edge-image-api.yml`** — its stated purpose was letting the now-archived `breeze-workspace` repo pin a stock image by SHA. I checked branch rulesets directly: the only required status check is `CI Success` from `ci.yml`, so deleting this strands nothing.

## Rescued rather than deleted with their files

Two things would have been silently lost by deleting whole files, and one needed extracting:

- `packages/extension-api` carried the **`RESERVED_ROUTE_NAMESPACES` tripwire** — the #2634/#2635 regression guard that derives core mounts from `index.ts` at test time. Ported to `apps/api/src/extensions/reservedRouteNamespaces.test.ts`.
- `bundleVerifier`'s **`assertVerifiedMemberBytes`** is needed by `extensionsWeb` for the TOCTOU re-hash when serving web assets. Moved into `webAssets.ts`, which owns the inventory it checks against.
- **`defaultStageExtension`** extracted from `reconciler.ts` into `stageExtension.ts` — the built-in path stages through the same seam.

## Behaviour changes worth reviewing

- `POST /api/v1/extensions/:name/{enable,disable}` no longer returns `scheduleSyncDeferred` (jobHost was its only producer).
- `mountExtensionGateway` and `executeTool` each lost their org-install parameter.
- `context.tenancy.installedOrgs()` now always throws. Returning `[]` would mean "activated for no orgs" and would silently no-op a caller's sweep.
- `installScope` stays in the v1 manifest schema, documented as RESERVED and explicitly not an authorization control — removing a key from a `.strict()` schema would reject any manifest declaring it.
- `ee/workspace`'s `legacyHostBridge` is gone: it adapted the one-argument `register()` call that only `stageLegacyExtension` made, and that caller no longer exists.
- `apps/docs` `features/extensions.mdx` is rewritten to describe reality — first-party built-ins compiled into the image, no third-party mechanism, no signed bundles, no per-org activation.

## Verification

- apps/api extensions+config **569/569**; ee/workspace **585/585** unit and **151/151** integration; apps/api routes+services **17,657 passed**
- Strict `tsc --project apps/api/tsconfig.json` clean; `pnpm lint` 6/6; docs build 164 pages; `actionlint` clean on every block touched
- **Live flag-on boot** against pgvector: built-in loads (`loaded built-in "workspace" 0.2.0`), `/health` 200, zero CRITICAL lines, and the new migration verifiably drops `extension_org_installs`

## Not in this PR

Folding `ee/workspace` into core as ordinary first-party code — direct route mounts, migrations in `apps/api/migrations`, pages on real URLs instead of `/extensions/workspace/...`. That removes almost no additional attack surface (first-party code runs in-process either way) but carries real risk: transcribing 14 tables and 200+ export-column classifications into core's cascade and export policy, making pgvector mandatory for every deployment, and porting 1,870 lines of Custom Elements to React. Deliberately separate.


ToddHebebrand:
**CI: 56 passing.** All four integration shards green — including shard 1, which runs the built-in boot step and the `ee/workspace` integration suite, so the surviving delivery path is exercised end to end. `Test API`, `Type Check`, `Lint`, `Smoke Test` and both binary-source *local* jobs all pass.

One failure: `BINARY_SOURCE=github download path`, which is **pre-existing and unrelated**. It fails identically on the last five `main` runs (including immediately after #3479 merged), and the failure is a 404 from the agent-binary GitHub *release* manifest lookup — `"Version not found for the specified platform and architecture"` — nothing to do with extensions.

`Main Red Alert` skips on PRs as designed.

I also verified the branch ruleset directly before deleting `edge-image-api.yml`: the only required status check is `CI Success` from `ci.yml`, so no required check is stranded.

Issue #3409, opened 2026-08-11
[API][Web] Pre-set script variables: partner/org-scoped values (S1 site token, vendor keys) resolvable into scripts


ToddHebebrand:
## Problem

There is no way to define a value once and reference it from a script. Every script that needs a tenant-specific or vendor-specific constant — a SentinelOne site token, a Huntress org key, a BitLocker escrow endpoint, an internal package repo URL, a syslog collector address — has to hardcode it in the script body. At MSP scale that means one copy of the script per customer, and a secret pasted into script content that is then stored, versioned, and shipped to every endpoint in the clear.

This is the "global variables" primitive every comparable RMM ships (NinjaOne global custom fields, Datto variables, Action1 parameters). Breeze has most of the machinery already and has simply never wired it together.

## Current state

Mapped against the code, three of the four pieces exist:

**Software deployment has a complete variable stack, scoped to two fields.**

- Vocabulary + picker + unknown-token validation client-side: `apps/web/src/lib/installerVariables.ts` (`BUILTIN_INSTALLER_VARIABLES` L41, `customFieldToken()` L50, `findUnknownTokens` L57-92), `apps/web/src/components/software/VariableInput.tsx`.
- Server-side resolver: `apps/api/src/services/installerVariables.ts` (`resolveKey` L32-60, `resolveInstallerVariables` L97), invoked per target device at deploy time in `apps/api/src/services/softwareDeployment.ts:388-415`.
- Fails loud: an unresolved token fails that device's `deploymentResults` row rather than shipping a literal `{{...}}`.
- Four namespaces only: `{{org.name}}`, `{{org.id}}`, `{{site.name}}`, `{{site.id}}`, `{{device.hostname}}`, `{{device.customField.<key>}}`. Applied only to `downloadUrl` and `silentInstallArgs`.

**Scripts already substitute — but only from a per-run parameter map.**

- `scripts.parameters` (jsonb, `apps/api/src/db/schema/scripts.ts:28`) holds the parameter *definition* list; types are `string | number | boolean | select` (`apps/web/src/components/scripts/ScriptFormSchema.ts:4-10`). There is no secret type.
- Values are entered at run time (`ScriptExecutionModal.tsx`), posted to `POST /scripts/:id/execute` (`apps/api/src/routes/scripts.ts:875`), and embedded in the `deviceCommands` payload alongside raw `content` (`apps/api/src/services/scriptExecution.ts:153,207`). **The API substitutes nothing into script content.**
[...]

Issue #3469, opened 2026-08-12
extensions: routeNamespace-level collision gate across all three delivery paths


ToddHebebrand:
Follow-up from #3455 (workspace ee-merge). Pre-existing gap, now spanning three delivery paths instead of two.

Extension identity collision is enforced by **name only**. Nothing gates two extensions (across source-directory, signed-runtime-bundle, and the new built-in path) claiming the same `routeNamespace` — the first registration wins and the second's routes silently shadow or conflict.

Add a collision gate at the contribution-registry level that rejects (fail-closed at boot / reconcile) any second claimant to a `routeNamespace`, regardless of which delivery path each extension arrived through. Cover with a cross-path test (built-in vs signed bundle claiming the same namespace).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ToddHebebrand:
Closing per the 2026-08-16 decision (#2665, tracked in #3640): the third-party / externally-installable extension direction is abandoned. Only the first-party built-in path (`ee/workspace`, `BREEZE_WORKSPACE_ENABLED`) remains, and this issue serves the external model. The source-directory and signed-bundle delivery paths are gone; with a single built-in path, routeNamespace collisions are a code-review concern.

Issue #3453, opened 2026-08-11
[Release] v0.105.0 ships UNSIGNED Windows binaries under the signed names — github-mode agent download 404s, main red


bdunncompany:
## Summary

`v0.105.0` (published `2026-08-11T22:05:38Z`) ships **unsigned Windows binaries under the signed asset names**. The API's startup binary sync rejects them, no manifest is stored, and `GET /api/v1/agents/download/...` then returns **404** in `BINARY_SOURCE=github` mode.

This is currently red on `main` and, more importantly, means a self-hoster on `BINARY_SOURCE=github` who takes v0.105.0 **cannot download the Windows agent at all**.

## Evidence

The signed and unsigned Windows assets are **byte-identical in size**, which an Authenticode signature cannot be — it appends a certificate table:

| Asset | Size |
|---|---|
| `breeze-agent-windows-amd64.exe` | **21266432** |
| `breeze-agent-windows-amd64-unsigned.exe` | **21266432** |
| `breeze-backup-windows-amd64.exe` | **47253504** |
| `breeze-backup-windows-amd64-unsigned.exe` | **47253504** |

macOS is the control, and it signed correctly — signed and unsigned differ as expected:

| Asset | Size |
|---|---|
| `breeze-agent-darwin-amd64` | 21205328 |
| `breeze-agent-darwin-amd64-unsigned` | 21022120 |
| `breeze-agent-darwin-arm64` | 19502064 |
| `breeze-agent-darwin-arm64-unsigned` | 19597186 |

So the Windows signing step produced a **copy** of the unsigned artifact rather than a signed one, and did not fail the release.

## How it surfaces

`smoke-binary-source-github` on `ea726c1c4`:

```
[startup] Binary sync failed (non-fatal in github mode):
  Error: Release asset breeze-agent-windows-amd64.exe requires platformTrust
         "windows-authenticode-required", got "none"
...
Download endpoint returned: 404
{ "error": "Version not found for the specified platform and architecture" }
Expected 200 — github-mode download endpoint should serve published manifests.
```

The sync error is logged as *non-fatal*, so boot continues and the failure only shows up later as a 404 from the download endpoint — which is why the message a user hits ("Version not found for the specified platform and architecture") points at the wrong thing entirely.

## It is the release, not the commit

[...]

PR #3494, opened 2026-08-13
feat(api,web): tenant variables — partner/org-scoped values store (#3409 PR1)

ToddHebebrand:
Part 1 of the tenant-variables stack for #3409. Follows PR0 (#3418 + review fixes #3438).

## What this delivers

An MSP defines a value once — an S1 site token, a vendor API key, a syslog collector address — either for a single customer org or **partner-wide across all their orgs**, and manages it from a new **Settings → Variables** page. Each variable is a `key`/`value` pair with an optional description and an `is_secret` flag. Secrets are write-only: never returned by any endpoint, no reveal path, and omitting `value` on edit leaves the stored ciphertext untouched. Two new permissions gate access: `variables:read` and `variables:manage`.

**A variable created here is inert until PR2.** Nothing consumes it yet — PR2 adds the `{{var.*}}` token syntax that reads these values.

## Design decisions worth a reviewer's attention

**Dual-ownership, not org-first.** `org_id` XOR `partner_id` per the Partner-Wide First playbook (epic #2135), enforced by `tenant_variables_one_owner_chk`. Resolution precedence (org > partner) lands in PR2; the site axis is deferred post-v1.

**Two RLS policies, deliberately.** `tenant_variables_isolation` is the standard dual-axis policy for all commands. A second, **SELECT-only** widening policy `tenant_variables_partner_wide_select` lets an org-scoped session *read* its partner's partner-wide rows. This is load-bearing: `breeze_has_partner_access` is flat and false for an org token, so without it an org admin would see an empty list while PR2's dispatch happily resolved those same values. Writes are unaffected — permissive policies OR together and this branch is `FOR SELECT`. Agent contexts set `currentPartnerId` NULL, so agents see nothing through it. Same mechanism as `2026-08-10-cis-baselines-partner-ownership.sql`.

**Encryption is unconditional and row-bound.** Every value is sealed via `secretCrypto`, secret or not. The obvious approach — reuse the registry's existing `table.column` AAD — was rejected: it stops a blob moving between *columns* but not between *rows*, i.e. between tenants. So `EncryptedColumnSpec` gains `aadBinding: 'column' | 'row'` and `columnAad(spec, rowId)`. The row id is generated app-side (`randomUUID()`) so the AAD exists before the INSERT. `encryptColumnValueForWrite` now **throws** for row-bound specs — intended fail-loud. Row-bound specs bypass the `aadV3Enabled()` gate since the table is new.

**App layer is never looser than RLS.** The read condition mirrors the widening policy, but `getVariableWithAccess` takes `forWrite: true`, which drops the partner-wide branch — an org caller trying to delete a partner-wide row gets 404, not 403.

**Key and `ownerScope` are immutable after create.** Renaming would silently break every referencing script; moving axes would change tenancy. `updateTenantVariableSchema` `.omit()`s both. Zod 4 note: `isSecret` is redeclared `.optional()` on the update schema because `.default(false)` survives `.partial()` and would silently clear the flag on a PATCH.

Smaller calls: partial unique indexes rather than a composite UNIQUE (with one axis always NULL it would never collide); key grammar `^[a-z][a-z0-9_]{0,63}$` enforced in Zod *and* as a DB CHECK, because PR2 interpolates it into `{{var.<key>}}` and `BREEZE_VAR_<UPPER>`; flipping `isSecret` true→false without supplying a new value is a 400; `version` bumps only when the value actually changes; audit `details` carry the key, never a value. Caps: value ≤ 4096 chars, description ≤ 500, **200 variables per owner** (409 on the 201st) — that cap is what bounds PR2's per-dispatch query.

## Behaviour changes for existing deployments

- `2026-08-11-variables-permissions.sql` **grants `variables:read` + `variables:manage` to every existing system `Org Admin` role.** The `is_system = TRUE` filter is load-bearing — custom roles accept caller-supplied names. Partner Admin already holds `*:*`. Guarded INSERTs rather than `ON CONFLICT`, since there is no unique constraint on `permissions(resource, action)`.
- Shared `encryptedColumnRegistry` gains `aadBinding`. Default `'column'` keeps every pre-existing spec byte-identical and **no existing ciphertext changes**; `reencryptRegisteredSecrets` now passes the row id.
- New sidebar entry under Settings, gated on `variables:read`.

## Tenancy registration

Shape: **dual-axis config table**. All four registrations done — `CORE_ORG_CASCADE_DELETE_ORDER` (between `support_sessions` and `ticket_alert_links`), `CORE_TENANT_EXPORT_POLICY` (`value` → `excludedSensitive`, rest `included`), `DUAL_AXIS_TENANT_TABLES`, and `encryptedColumnRegistry`. No device lists — there is no `device_id` column. Cascade leaf; partner-wide rows (`org_id NULL`) survive an org erasure by design.

Migrations: `2026-08-11-tenant-variables.sql`, `2026-08-11-variables-permissions.sql`.

## Testing

Unit: `services/tenantVariables.test.ts`, `routes/tenantVariables.test.ts`, `encryptedColumnRegistry.test.ts`, shared validators, `TenantVariablesPage.test.tsx`, plus `no-silent-mutations` (proves `runAction` adoption with no allowlist entry) and locale parity across all seven locales.

Contract suites: new `tenantVariablesPartnerRls.integration.test.ts` (cross-partner forge 42501, XOR 23514, key CHECK 23514, dupe 23505, org-reads-inherited-but-not-foreign, widening-is-SELECT-only), plus `rls-coverage`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, and `db:check-drift`. Manual `breeze_app` psql forge per CLAUDE.md step 6.

Traps hit while building, recorded so review doesn't re-derive them: `is_secret` matches `SUSPICIOUS_NAME_PARTS`, so the export policy refuses it in `included` — it must be `reviewedIncluded`. An over-length key raises 22001 (varchar) *before* the key CHECK's 23514. `permissionsCatalog.ts` keeps its own `RESOURCE_LABELS` map that a new permission resource must join or the route test fails. Adding a permission needs **four** places: shared registry, `DEFAULT_PERMISSIONS`, `SYSTEM_ROLES` (fresh installs seed *after* migrating), and a seed migration for already-deployed roles.

## Out of scope

`{{var.*}}` tokens, resolver, software-deploy wiring, docs → **PR2**. Sourced script parameters → **PR3**. `secretEnv` / `BREEZE_VAR_*` delivery, redaction, digest pinning → **PR4**. Site axis → post-v1. No agent changes; no `scriptDispatch.ts` changes.


ToddHebebrand:
Stack for #3409:

- **#3494 (this PR)** — the `tenant_variables` store. Targets `main`, gets full CI.
- **#3495** — `{{var.*}}` resolution at dispatch. Stacked on this branch, so it gets **no automatic CI**; dispatched manually per branch.

Merge this one first. Because `ci.yml` triggers on `pull_request: branches: [main]`, #3495 will need a rebase onto `main` after this squash-merges — via `git rebase --onto`, not a hand-resolve (a parent squash-merge otherwise produces phantom conflicts across the whole stack).

bdunncompany:
Two CI notes while the run is still in flight (logs aren't downloadable until it completes, so this is from the check-run annotations API).

**1. `Test API` failed — and it's the root-mount tripwire, not a flake.**

```
packages/extension-api/src/index.test.ts:299
AssertionError: expected [ 'externalServicesRoutes', …(8) ] to deeply equal [ 'externalServicesRoutes', …(7) ]
```

That's the `pins which sub-routers are root-mounted where the derivation cannot see them` test. It fired exactly as designed — this PR adds an 8th root-mounted sub-router where the list pins 7. The test's own comment carries the remedy:

> If this list changes, open the new sub-router, resolve its top-level segments by hand, and add them to `RESERVED_ROUTE_NAMESPACES`.

So it likely wants the new variables router's top-level segments reserved, not just the identifier appended to the array. (The other failure, `BINARY_SOURCE=github download path`, is the pre-existing #3453 one.)

Worth noting this intersects **#3469** — the routeNamespace collision gate you filed yesterday. A new root-mounted router claiming namespaces is precisely the case that issue is about, so whatever segments get reserved here are also inputs to that gate.

**2. #3495 has effectively no CI, and it reads as green.**

```
$ gh pr view 3495 --json baseRefName,headRefName
base=ToddHebebrand/tenant-variables-pr1  head=ToddHebebrand/tenant-variables-pr2
$ gh pr checks 3495
Cloudflare Pages   pass
```

One check, and it's the Pages preview. That's the stacked-PR blind spot documented in `CLAUDE.md` — `ci.yml` triggers on `pull_request: branches: [main]`, so a PR based on a sibling branch runs no CI at all. For a 58-file, +4,320-line PR that's a lot of unexercised surface, and `gh pr checks` gives no hint. `gh workflow run CI --ref ToddHebebrand/tenant-variables-pr2` before merging would cover it.

Given PR2 stacks on PR1, the `Test API` red above also isn't visible from PR2's checks even though PR2 inherits it.

bdunncompany:
Correcting my comment above — both points were already stale when I posted, and the second one was simply wrong.

**The `Test API` red was fixed before I flagged it.** You pushed `666d96681` at 05:02:52Z / 05:03:00Z; my comment landed 05:04:01Z. Your own diagnosis on #3495 is the same one and got there first: `tenantVariableRoutes` is root-mounted, so its top-level segments need hand-registering in the extension guard's reserved-namespace sets. #3494 is now 44 pass with only the pre-existing `BINARY_SOURCE=github` red.

**And #3495 does have CI — I misread the tool.** You had already dispatched it. `workflow_dispatch` runs don't attach as PR check-runs, so `gh pr checks 3495` shows only the Cloudflare Pages preview no matter how much CI is actually running:

```
$ gh api /repos/LanternOps/breeze/actions/runs/31669176268
branch=ToddHebebrand/tenant-variables-pr2  event=workflow_dispatch  status=in_progress  name=CI
```

So my "effectively no CI, and it reads as green" was an artifact of which view I looked at, not a fact about the branch. I concluded absence from a filtered list without running `gh run list --branch`, which is the one command that would have shown it.

Worth recording for anyone else reading `gh pr checks` on a stacked PR here: it under-reports in both directions — no `ci.yml` run because the base isn't `main`, *and* no visibility of the manual dispatch that compensates for it. The branch-scoped run list is the only reliable view.

Apologies for the noise on an otherwise clean PR.

ToddHebebrand:
CI green: **56 checks passing**, including all four Integration Tests shards (the RLS / tenant-cascade / export-policy contract suites for the new table), both Check Migrations jobs, Test API and Test Web.

One red — `BINARY_SOURCE=github download path` — is **pre-existing on `main`**: the same workflow fails on `main` at `8582acb9d`, the exact commit this branch is rebased onto. Not caused by this PR.

This branch was rebased onto current `main` before opening. That surfaced the root-mounted-namespace gap fixed in `666d96681`; on its original base the branch was green and would have gone green here, then reddened `main` after merge.

PR #3495, opened 2026-08-13
feat(api,web): resolve {{var.*}} per device at dispatch (#3409 PR2)

ToddHebebrand:
Part 2 of the tenant-variables stack for #3409. **Stacked on `ToddHebebrand/tenant-variables-pr1` — review and merge PR1 first.**

## What this delivers

PR1's variables become live. A `{{var.<key>}}` token written into a script body — or into a software installer URL / silent-install argument — resolves server-side **per target device** to that device's org value, falling back to the partner-wide value. A device whose token can't be resolved now fails loudly with its own `failed` execution row instead of receiving a literal `{{var...}}`, and the rest of the fan-out still runs. Script authors get an insert-variable picker in the Monaco editor with warn-only unknown-key validation, plus a new **Variables** group in the software-deployment picker.

Secret variables cannot be referenced in content in v1: `{{var.<secret_key>}}` is a 400 at script save/import and a per-device failure at dispatch. PR4 adds the out-of-band delivery that makes secrets usable.

## Design decisions worth a reviewer's attention

**Resolution never rides ambient RLS.** The five dispatch call sites run under three different DB contexts and would otherwise resolve *different variable sets for the same script*. `loadTenantVariableScope` runs inside `runOutsideDbContext(() => withSystemDbAccessContext(...))` — **both wrappers are required**. A bare system context nested in a held request transaction does not elevate, and an org token whose JWT lacks `partnerId` would then see zero partner-wide rows. The cost of that decision: RLS no longer constrains the load, so the `WHERE` clause is the **only** tenancy boundary and has to be exact. That is what the new integration suite exists to pin.

**Opaque snapshot, verified at use.** `TenantVariableScope` is produced solely by `loadTenantVariableScope`, carries the org ids it was built for, and `resolveForOrg` **throws** if asked for an org outside that set.

**One chokepoint.** Substitution happens in `scriptDispatch.ts` between field resolution and the `script_executions` insert, so every ingress inherits it and a failed device never leaves an orphan `pending` row. A `hasVariableTokens` guard keeps the token-free path allocation- and query-free — preloading unconditionally was a measurable hot-path regression, since every script run escaped the request transaction for a second connection to build a snapshot nothing consulted.

**`${{var.x}}` is deliberately NOT a token**, and `{{ var.x }}` (inner whitespace) is not either. A `$` before `{{` means the author wrote shell/Actions syntax. This diverges from the existing installer tokenizer, which eats both — that tokenizer has a client-strips/server-trims divergence, and this PR refuses to extend that bug into the new namespace. Enforced with a lookbehind. Every other `{{...}}` form (Actions, Jinja, brace expansion, `{{org.name}}`, `{file}`) passes through verbatim.

**No recursive expansion** — a single `String.replace` pass, so a substituted value is never re-scanned. Empty string counts as unresolved (matching `resolveKey`'s existing rule), as does an undecryptable row — omitted and logged by id, never resolved to `''`. `{kind:'raw'}` `execute_command` sources are skipped: no declaring script, so nothing could have been validated at save time.

**Unknown keys are warn-only** at save and in the UI, since a key may be created after the script; only the secret rule hard-blocks. A failed key-list fetch in the UI degrades to an empty list and suppresses warnings rather than showing false positives.

## Deliberate deviations from the scope comment

1. **No env injection at all** — textual substitution only, where the scope called for `BREEZE_VAR_<UPPER_KEY>` env vars. The agent has no `BREEZE_VAR_*` handling until PR4, and the only env channel that exists today is the `parameters` map, which the agent substitutes into script *text* as well as mirroring to `BREEZE_PARAM_*`. Routing variables through it would make every value textually substitutable — exactly the leak PR4's separate `secretEnv` channel exists to prevent. "env-first delivery" is therefore deferred to PR4.
2. **Snapshot verification made explicit** — the snapshot carries its org ids and dispatch throws on an out-of-scope org.

## ⚠️ Behaviour changes for existing users — release-note these

**1. Script parameters are now strictly validated and canonicalized to strings.** One shared `scriptParametersSchema` is adopted at `routes/scripts.ts`, `routes/mobile.ts`, and `automationRuntime.ts`, with `canonicalizeScriptParameters` called once at the `scriptDispatch.ts` payload build so every ingress puts strings on the wire. Now **rejected**: `null`, objects, arrays, `NaN`, `Infinity`, and keys the agent could not turn into an env var name (`has space`, `a.b`, `a=b`). Caps: 64 parameters, 4096 chars per value. Numbers and booleans are now accepted and coerced.

This **fixes a silent pre-existing outage**: the wire type is `map[string]string` and the agent silently dropped every non-string value, so a number or boolean parameter — which the UI actively encourages — left `{{name}}` verbatim in the script and never set `BREEZE_PARAM_*`. But it also means **callers currently sending malformed parameters get a validation error instead of a silent drop**. `normalizeAutomationActions` runs at *runtime* against stored actions, so existing automations with malformed parameters fail loudly rather than silently dropping them.

**2. Script fan-out no longer aborts on the first per-device failure.** `executeScriptOnDevices` previously threw, so devices after the failing one never got a row and the batch's `devicesTargeted` was a lie. Now the device gets a `script_executions` row with `status: 'failed'` and a non-empty `errorMessage`, counts toward `devicesFailed`, and the fan-out continues. `insert_failed` still throws — it signals a programming error. New failure code `'unresolved_variables'`.

**3. Existing content containing a strictly-well-formed `{{var.<key>}}` string is no longer passed through verbatim** — it is substituted, or the device fails with `unresolved_variables`. `${{var.x}}` and whitespace/other-namespace forms are unaffected.

**4. Saving or importing a script that references a secret variable is now a 400** (create, update, and bundle import). A script saved before the variable was flipped to secret still fails per-device at dispatch.

**5. The four-eyes effect digest is NOT extended in this PR.** It hashes `scripts.content` — the *template* — at both pin and release, so a variable's *value* changing between approval and release does not invalidate the digest. PR4 closes this. Calling it out explicitly so review doesn't assume otherwise.

## Testing

No migrations and no new tables — this consumes PR1's `tenant_variables` and its dual-axis shape.

Unit: `variableTokens`, `scriptParameters`, `tenantVariableResolution`, plus `scriptDispatch`, `scriptExecution`, `automationRuntime`, `aiToolsScripts`, `installerVariables`, `softwareDeployment`, `scriptBundle`, `mobile`, `commandAudit`. Web: `VariableInput`, `ScriptVariablePicker`, `ScriptForm`, `installerVariables`, locale parity, plus `astro check`.

New contract suite `tenantVariableResolution.integration.test.ts` proves against real Postgres: org-shadows-partner precedence, partner inheritance, no cross-partner leakage, a two-org snapshot resolving independently in one call, **resolution working from inside an org-scoped `withDbAccessContext`** (delete the system-context escape and this test must fail), and secret exclusion. `loadTenantVariableScope` issuing exactly one query per snapshot is asserted.

Existing fixtures that asserted the old parameter shape were **updated to the canonicalized expectation rather than loosening the schema**.

## CI note

`ci.yml` triggers on `pull_request: branches: [main]`, so this stacked PR gets **no CI** while `gh pr checks` reads green. Dispatch it explicitly before merge:

```
gh workflow run CI --ref ToddHebebrand/tenant-variables-pr2
```

## Out of scope

`BREEZE_VAR_*` / `secretEnv` delivery, payload erasure on all terminal transitions, redaction, and effect-digest pinning → **PR4**. Sourced parameters → **PR3**. **No agent (Go) changes here.** A separate pre-existing outage — the user-helper client drops parameters entirely for `runAs: 'user'` runs — is deliberately not fixed here (it needs a helper IPC change, PR4 territory) and is filed as a follow-up.

### DIFF
--- a/apps/api/src/__tests__/integration/tenantVariableResolution.integration.test.ts
+++ b/apps/api/src/__tests__/integration/tenantVariableResolution.integration.test.ts
@@ -1,0 +1,209 @@
+/**
+ * Tenant variable resolution (#3409 PR 2) — real-Postgres coverage for
+ * `loadTenantVariableScope` / `resolveForOrg` / `substituteTenantVariables`.
+ *
+ * `services/tenantVariableResolution.test.ts` proves the WHERE-clause shape
+ * against a mocked db; this suite proves the six properties that only mean
+ * anything against a real RLS-enforcing database:
+ *   1. org-owned value shadows the partner-wide one with the same key
+ *   2. an org inherits its partner's partner-wide value
+ *   3. an org NEVER sees another partner's partner-wide value
+ *   4. a two-org snapshot resolves each org independently in one call
+ *   5. resolution still returns partner-wide rows when called from INSIDE an
+ *      org-scoped withDbAccessContext — the regression proving the
+ *      runOutsideDbContext + withSystemDbAccessContext escape is present.
+ *      Comment out that escape in tenantVariableResolution.ts and THIS test
+ *      is the one that goes red (an org token's JWT lacks partnerId, so a
+ *      nested withSystemDbAccessContext that fails to elevate would see zero
+ *      partner-wide rows for the org under test).
+ *   6. a secret variable is loaded into the snapshot but substitution reports
+ *      it via secretsReferenced and never places its value in content.
+ *
+ * Modeled on tenantVariablesPartnerRls.integration.test.ts (same
+ * createPartner/createOrganization/withDbAccessContext helpers). Fixture rows
+ * are inserted through services/tenantVariables.ts's own
+ * encryptTenantVariableValue so the resolver's decryptTenantVariableValue call
+ * actually round-trips them — a literal ciphertext placeholder (as the RLS
+ * suite uses, since it never decrypts) would make every resolved value
+ * silently fail to decrypt and vanish from the snapshot.
+ */
+import './setup';
+import { randomUUID } from 'crypto';
+import { afterEach, describe, expect, it } from 'vitest';
+import { eq } from 'drizzle-orm';
+import { db, withDbAccessContext, type DbAccessContext } from '../../db';
+import { tenantVariables, type TenantVariableRow } from '../../db/schema';
+import { createOrganization, createPartner } from './db-utils';
+import { encryptTenantVariableValue } from '../../services/tenantVariables';
+import {
+  describeVariableFailure,
+  loadTenantVariableScope,
+  resolveForOrg,
+  substituteTenantVariables
+} from '../../services/tenantVariableResolution';
+
+const created: string[] = [];
+
+function systemContext(): DbAccessContext {
+  return { scope: 'system', orgId: null, accessibleOrgIds: null, accessiblePartnerIds: null, userId: null };
+}
+
+/**
+ * An org-scoped session whose JWT carries no partnerId on the accessible-ids
+ * axis — accessiblePartnerIds stays empty, mirroring production exactly (see
+ * tenantVariablesPartnerRls.integration.test.ts's orgContext). This is the
+ * context property 5 dispatches resolution from: if the system-context escape
+ * in loadTenantVariableScope were missing, this ambient context (which cannot
+ * see partner-wide rows via RLS) would silently constrain the "system" query
+ * too, and the partner-wide fixture would disappear from the snapshot.
+ */
+function orgContext(orgId: string, partnerId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: [],
+    userId: null,
+    currentPartnerId: partnerId
+  };
+}
+
+interface FixtureOptions {
+  orgId?: string | null;
+  partnerId?: string | null;
+  key?: string;
+  isSecret?: boolean;
+}
+
+/** Insert one tenant_variables row with a real, decryptable ciphertext. */
+async function insertVariable(plaintext: string, options: FixtureOptions): Promise<TenantVariableRow> {
+  const id = randomUUID();
+  const [row] = await withDbAccessContext(systemContext(), () =>
+    db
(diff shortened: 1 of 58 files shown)


ToddHebebrand:
Stacked on #3494 — **review and merge that one first**. This PR's diff against `main` will look correct only after #3494 lands; the meaningful diff right now is against `ToddHebebrand/tenant-variables-pr1`, which is what the base is set to.

Because this targets a sibling branch rather than `main`, `ci.yml` never triggers for it and `gh pr checks` reads green while nothing has run. CI is therefore dispatched by hand: https://github.com/LanternOps/breeze/actions/runs/31668758894

ToddHebebrand:
CI re-dispatched at the current tip `4f7a394f9`: https://github.com/LanternOps/breeze/actions/runs/31669176268

The earlier dispatched run (31668758894) is **stale — ignore it**. It ran against `f2aad32ba`, before this branch was restacked, and failed on `Test legacy extension adapter`: #3494 root-mounts `tenantVariableRoutes`, and root-mounted routers must have their top-level segments hand-registered in the extension guard's reserved-namespace sets or an extension could claim `tenant-variables` and shadow the core endpoints. Fixed in `666d96681` on #3494, which this branch now sits on. That run has been cancelled.

ToddHebebrand:
CI green at `4f7a394f9` — **zero job failures**: https://github.com/LanternOps/breeze/actions/runs/31669176268

Test API, Test Web, Type Check, both Check Migrations jobs, and all four Integration Tests shards passed — the last of those covers `tenantVariableResolution.integration.test.ts`, which is the suite that pins the system-context escape (delete it and the org-scoped-context case must fail).

Reminder for whoever merges: this run was **hand-dispatched**. Because the PR targets a sibling branch rather than `main`, `ci.yml` never fires on push, so any further commits here need another `gh workflow run CI --ref ToddHebebrand/tenant-variables-pr2` — `gh pr checks` will otherwise read green against nothing.

PR #3507, opened 2026-08-13
fix(api): persist send_email_reason on the direct quote send path

### LINKED ISSUE #3502: Direct POST /quotes/:id/send never persists send_email_reason, so a failed send shows no banner
## What happens

`sendQuote` treats email delivery as best-effort: every failure is swallowed into an `emailReason` code (`no_email_service`, `no_billing_contact`, `pdf_render_failed`, `send_failed`) so the send still commits. That reason is meant to be persisted to `quotes.send_email_reason`, which is what raises the detail page's "Proposal marked Sent, but no email was delivered" banner.

Two of the three send paths persist it. The direct one does not:

| Path | Persists `send_email_reason`? |
|---|---|
| `resendQuote` (`services/quoteLifecycle.ts`) | yes |
| scheduled send worker (`jobs/quoteSendQueue.ts`) | yes |
| **`sendQuote` — `POST /quotes/:id/send`** | **no** |

`sendQuote` clears `sendEmailReason: null` in its draft→sent claim, calls `deliverQuoteEmail`, re-selects the row and returns. The reason reaches the HTTP response but never the column.

## Impact

A direct send whose PDF render or transport fails is marked **Sent** with `send_email_reason` NULL, so the banner never appears. The customer never received the proposal and nothing on the quote says so — the tech finds out when the deal goes quiet.

The web composer normally goes through `/schedule-send` (the undo window), which is covered. This bites the paths that call `/send` directly: MCP, bulk-send, API-key callers, and the "Send now" button that cancels the window and dispatches immediately.

## Fix

Persist the reason after `deliverQuoteEmail` the way `resendQuote` does. Worth copying its shape too: that write happens after the email has left, so a failure there is swallowed onto the returned row rather than surfacing as "could not send" for a message the customer already has.

## Notes

Pre-existing; predates the re-send work. Surfaced by review on #3501, which documents the asymmetry in `deliverQuoteEmail`'s contract comment rather than expanding that PR's scope. That comment should be simplified once this is fixed.

bdunncompany:
Fixes #3502.

`sendQuote` swallows every delivery failure into an `emailReason` so the send still commits, but it was the only one of the three send paths that never wrote that reason to `quotes.send_email_reason`:

| Path | Persisted before | Now |
|---|---|---|
| `resendQuote` | yes | yes |
| scheduled send worker (`jobs/quoteSendQueue.ts`) | yes | yes |
| `sendQuote` (`POST /quotes/:id/send`) | **no** | yes |

That column is what raises the detail page's "Proposal marked Sent, but no email was delivered" banner, so a direct send whose PDF render or transport failed was marked **Sent** with the column NULL and showed nothing at all. The customer never received the proposal and the quote gave no sign of it. This is the path used by MCP, bulk-send, API-key callers and "Send now"; the web composer normally goes through `/schedule-send`, which was already covered.

## The change

Persist the reason after `deliverQuoteEmail`, in `resendQuote`'s shape. The write runs after the email has left, so a failure there is swallowed onto the returned row rather than surfacing as "could not send" for a message the customer already has, where the tech's natural next move is to send it a second time. The draft to sent claim already clears the column, so only a failure needs writing back, and the existing re-select and return are untouched.

Also simplifies `deliverQuoteEmail`'s contract comment, which existed to document the asymmetry this removes.

## Tests

The three existing failure-path cases each needed one more queued mock result, because the Drizzle mock queue is sequential and the new UPDATE consumes one. Added:

- `persists send_email_reason so a failed direct send raises the banner` — asserts the write itself via `setCalls` rather than the returned row, since the banner reads the column.
- `writes no outcome marker when the send succeeds` — guards against reintroducing a stale marker on the happy path. Being precise: this one passes with or without the fix, so it is a regression guard, not a reproducer.

## Verification

Run locally on node 22.23.2 (matching `.nvmrc`), deps installed `--frozen-lockfile`:

```
vitest src/services/quoteLifecycle.test.ts        58 passed (58)
vitest quote services + routes + quoteSendQueue   383 passed (383), 23 files
tsc --noEmit -p apps/api/tsconfig.json            exit 0, no output
```

Control run, fix reverted and tests kept: 4 failed, including `persists send_email_reason so a failed direct send raises the banner`. So the new persistence test genuinely fails without the change.

Not run locally: the RLS and integration contract suites, which need a live database. This touches no schema, no migration and no tenancy surface, so I would not expect them to be affected.

---

## Known risks (adversarial review, recorded per review charter)

This PR makes the "no email was delivered" banner fire for the direct-send path — the reported bug. It does **not** make quote delivery idempotent. Three risks were surfaced by adversarial review and are left for follow-up rather than silently carried:

1. **Transaction-boundary duplicate send (pre-existing, marginally widened here).** `sendQuote` claims draft→sent then renders the PDF and calls the provider while the request-wide transaction is still open. A statement failure after delivery rolls back `status='sent'` while the customer has the email; a retry re-claims the draft and sends a second copy. The unguarded re-select already had this property; this PR adds one more statement inside that window. Correct fix is a transactional outbox.
2. **`send_email_reason` is a lossy scalar with no attempt identity.** `resendQuote` has no claim/lock, so overlapping resends are last-commit-wins: a slow failure can overwrite a newer success (false warning) or a slow success can clear a newer failure (false reassurance). A per-attempt id with `WHERE id = ? AND delivery_attempt_id = ?` would make the column truthful.
3. **Unit tests cannot prove SQL targeting.** The Drizzle mock treats `.where()` as a no-op, so the persistence assertion passes even if production matched zero rows or the wrong row. A real-Postgres test with two controlled resends plus an injected post-email failure is needed; it belongs with the outbox work.

Downstream readers of the now-populated column: `QuoteSendOutcomeBanners`/toast (`QuoteActions.tsx`), tenant export (`tenantExportPolicyRegistry.ts`, already classified), MCP/AI quote tools. The public portal DTO is allowlisted and does **not** expose it — no portal leak.

### DIFF
--- a/apps/api/src/services/quoteLifecycle.test.ts
+++ b/apps/api/src/services/quoteLifecycle.test.ts
@@ -387,7 +387,8 @@ describe('sendQuote email delivery status', () => {
   it('reports no_billing_contact (and sends nothing) when the org has no billing email', async () => {
     queueThroughClaim({ name: 'Customer Co', taxId: null, billingContact: null });
     // No billingContact → the email branch short-circuits before the
-    // portalBranding read, straight to the final re-select.
+    // portalBranding read, straight to the outcome marker and the re-select.
+    queueResult([]); // outcome-marker update
     queueResult([{ id: 'q1', orgId: 'org1', partnerId: 'p1', status: 'sent' }]);
 
     const result = await sendQuote('q1', actor);
@@ -401,6 +402,7 @@ describe('sendQuote email delivery status', () => {
   it('reports send_failed when the email provider throws (send still commits)', async () => {
     queueThroughClaim({ name: 'Customer Co', taxId: null, billingContact: { email: 'billing@customer.example' } });
     queueResult([]); // portalBranding — none configured
+    queueResult([]); // outcome-marker update
     queueResult([{ id: 'q1', orgId: 'org1', partnerId: 'p1', status: 'sent' }]); // final re-select
     sendEmailMock.mockRejectedValue(new Error('smtp down'));
 
@@ -414,6 +416,7 @@ describe('sendQuote email delivery status', () => {
   it('reports pdf_render_failed (and never calls the email transport) when building the attachment throws', async () => {
     queueThroughClaim({ name: 'Customer Co', taxId: null, billingContact: { email: 'billing@customer.example' } });
     queueResult([]); // portalBranding — none configured
+    queueResult([]); // outcome-marker update
     queueResult([{ id: 'q1', orgId: 'org1', partnerId: 'p1', status: 'sent' }]); // final re-select
     vi.mocked(renderQuotePdf).mockRejectedValueOnce(new Error('pdfkit blew up'));
 
@@ -423,6 +426,34 @@ describe('sendQuote email delivery status', () => {
     expect(result.emailReason).toBe('pdf_render_failed');
     expect(sendEmailMock).not.toHaveBeenCalled(); // never reached the transport
     expect(result.quote.status).toBe('sent'); // the send itself still commits
+  });
+
+  // #3502: a direct send that fails delivery must leave a marker, or the detail
+  // page shows "Sent" with no banner and the customer silently got nothing.
+  it('persists send_email_reason so a failed direct send raises the banner', async () => {
+    queueThroughClaim({ name: 'Customer Co', taxId: null, billingContact: { email: 'billing@customer.example' } });
+    queueResult([]); // portalBranding — none configured
+    queueResult([]); // outcome-marker update
+    queueResult([{ id: 'q1', orgId: 'org1', partnerId: 'p1', status: 'sent', sendEmailReason: 'send_failed' }]);
+    sendEmailMock.mockRejectedValue(new Error('smtp down'));
+
+    const result = await sendQuote('q1', actor);
+
+    expect(result.emailReason).toBe('send_failed');
+    // The write itself, not just the returned row: the banner reads the column.
+    expect(setCalls.some((p) => p.sendEmailReason === 'send_failed')).toBe(true);
+  });
+
+  it('writes no outcome marker when the send succeeds (the claim already cleared it)', async () => {
+    queueThroughClaim({ name: 'Customer Co', taxId: null, billingContact: { email: 'billing@customer.example' } });
+    queueResult([]); // portalBranding — none configured
+    queueResult([{ id: 'q1', orgId: 'org1', partnerId: 'p1', status: 'sent' }]); // final re-select
+
+    const result = await sendQuote('q1', actor);
+
+    expect(result.emailed).toBe(true);
+    expect(result.emailReason).toBeUndefined();
+    expect(setCalls.some((p) => p.sendEmailReason === 'send_failed')).toBe(false);
   });
 
   it('reports emailed:true with no reason on a successful send', async () => {
--- a/apps/api/src/services/quoteLifecycle.test.ts
+++ b/apps/api/src/services/quoteLifecycle.test.ts
@@ -444,7 +444,7 @@ describe('sendQuote email delivery status', () => {
     expect(setCalls.some((p) => p.sendEmailReason === 'send_failed')).toBe(true);
   });
 
-  it('writes no outcome marker when the send succeeds (the claim already cleared it)', async () => {
+  it('writes exactly one sendEmailReason on a successful send: the claim clearing it', async () => {
     queueThroughClaim({ name: 'Customer Co', taxId: null, billingContact: { email: 'billing@customer.example' } });
     queueResult([]); // portalBranding — none configured
     queueResult([{ id: 'q1', orgId: 'org1', partnerId: 'p1', status: 'sent' }]); // final re-select
@@ -453,7 +453,13 @@ describe('sendQuote email delivery status', () => {
 
     expect(result.emailed).toBe(true);
     expect(result.emailReason).toBeUndefined();
-    expect(setCalls.some((p) => p.sendEmailReason === 'send_failed')).toBe(false);
+    // Exactly one `.set()` carries sendEmailReason (the draft→sent claim, which
+    // clears it) — a successful send must not add a second bookkeeping write.
+    // Asserting the COUNT rather than the absence of a value: a stray
+    // `{ sendEmailReason: null }` update would slip past a value-only check.
+    const reasonWrites = setCalls.filter((p) => 'sendEmailReason' in p);
+    expect(reasonWrites).toHaveLength(1);
+    expect(reasonWrites[0]?.sendEmailReason).toBeNull();
   });
 
   it('reports emailed:true with no reason on a successful send', async () => {
--- a/apps/api/src/services/quoteLifecycle.ts
+++ b/apps/api/src/services/quoteLifecycle.ts
@@ -247,8 +247,15 @@ export async function sendQuote(
   // status flip, which is the honest outcome — the email having already left is
   // a pre-existing property of sending inside the request transaction, not
   // something this write introduces.
+  // Scoped by orgId as well as id, not because the request path needs it (forced
+  // RLS covers that, and getQuote already enforced actor access) but because the
+  // scheduled-send worker calls this under a SYSTEM context where RLS is not the
+  // boundary. An id-only write there has nothing but the id standing between it
+  // and the wrong row.
   if (emailReason) {
-    await db.update(quotes).set({ sendEmailReason: emailReason, updatedAt: new Date() }).where(eq(quotes.id, id));
+    await db.update(quotes)
+      .set({ sendEmailReason: emailReason, updatedAt: new Date() })
+      .where(and(eq(quotes.id, id), eq(quotes.orgId, quote.orgId)));
   }
 
   const [updated] = await db.select().from(quotes).where(eq(quotes.id, id)).limit(1);
--- a/apps/api/src/services/quoteLifecycle.ts
+++ b/apps/api/src/services/quoteLifecycle.ts
@@ -231,6 +231,25 @@ export async function sendQuote(
     quote, blocks, lines, partnerRow, quoteNumber, acceptUrl, frozenQuote, billingRecipient, opts,
   });
 
+  // Persist THIS attempt's outcome, matching resendQuote and the scheduled-send
+  // worker: without it a direct send whose PDF render or transport failed is
+  // marked sent with send_email_reason NULL, so the detail page's "no email was
+  // delivered" banner never fires and nobody learns the customer got nothing.
+  // The draft→sent claim above already cleared the column, so only a failure
+  // needs writing back.
+  //
+  // Bookkeeping only, and it runs AFTER the email has left: a throw here would
+  // surface as "could not send" for a message the customer already has, and the
+  // tech's natural next move is to send it a second time. Swallow it instead.
+  if (emailReason) {
+    try {
+      await db.update(quotes).set({ sendEmailReason: emailReason, updatedAt: new Date() }).where(eq(quotes.id, id));
+    } catch (err) {
+      console.error(`[quoteLifecycle] send delivered for quote ${id} but persisting its outcome failed:`, err);
+      captureException(err instanceof Error ? err : new Error(String(err)));
+    }
+  }
+
   const [updated] = await db.select().from(quotes).where(eq(quotes.id, id)).limit(1);
   return { quote: updated!, emailed, emailReason, acceptUrl };
 }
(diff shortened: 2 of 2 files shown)


bdunncompany:
`Go Vulnerability Check` is red on this PR and it is not caused by it. Flagging with the evidence so it does not read as a defect in the change.

This PR touches two files, both TypeScript:

```
$ git diff --name-only origin/main...HEAD
apps/api/src/services/quoteLifecycle.test.ts
apps/api/src/services/quoteLifecycle.ts

$ git diff --stat origin/main...HEAD -- '*.go' agent/go.mod agent/go.sum
(empty)
```

Zero Go changes, so the Go code here is byte-identical to `main`.

What govulncheck actually reported:

```
ALLOWLISTED GO-2026-5051 — reviewed exception
FAIL GO-2026-5972 — symbol-level vulnerability not in the allowlist
FAIL GO-2026-6088 — symbol-level vulnerability not in the allowlist
FAIL GO-2026-6090 — symbol-level vulnerability not in the allowlist
FAIL GO-2026-6218 — symbol-level vulnerability not in the allowlist
```

All four are **Go standard library** advisories, not module dependencies:

| ID | Alias | Package | Summary |
|---|---|---|---|
| GO-2026-5972 | CVE-2026-33818 | `encoding/asn1` | Enforce maximum recursion depth |
| GO-2026-6088 | CVE-2026-56859 | `encoding/xml` | Add recursion depth guard during decode |
| GO-2026-6090 | CVE-2026-56862 | `crypto/tls` | Limit post-handshake messages accepted |
| GO-2026-6218 | CVE-2026-56860 | `net/url` | Avoid quadratic complexity in `resolvePath` |

Each reports `fixed in: 1.25.13, 1.26.6, 1.27.0-rc.3`, and `ci.yml:26` pins `GO_VERSION: '1.25.12'`. So the whole failure is one patch release behind on the toolchain, and no repo code is implicated.

**This is about to turn `main` red too.** The same job passed on `main` @ `497d0297f` at 2026-08-13T20:40:52Z and failed here at 22:17Z. Same Go code, same pinned version, different result, because govulncheck resolves against a live advisory database and these landed in between. The next run on `main` should fail for the same reason without any commit having changed.

The fix looks like a one-line bump of `GO_VERSION` to `1.25.13`. Happy to open that as its own PR if useful, or leave it to you since it is CI/toolchain rather than product code. I have deliberately not folded it into this branch, since an unrelated toolchain bump does not belong in a quote-service fix.

bdunncompany:
Correcting my own suggestion above: **bumping `ci.yml:26` alone would not fix this.** The failing job lives in `security.yml`, which carries its own pin.

There are five independent Go pins on `main`, all on the affected version:

```
.github/workflows/ci.yml:26               GO_VERSION: '1.25.12'
.github/workflows/dev-build-agent.yml:27  GO_VERSION: '1.25.12'
.github/workflows/release.yml:16          GO_VERSION: '1.25.12'
.github/workflows/security.yml:17         GO_VERSION: '1.25.12'   <- gates the failing govulncheck
.github/workflows/codeql.yml:37           go-version: '1.25.12'   (literal, not an env var)
```

## When this lands on `main`

`security.yml` triggers on `push` to main, `pull_request` to main, and `schedule: '0 6 * * 1'`. So:

- **Every open PR fails this check from now until the bump**, not just this one.
- **`main` goes red on your next merge**, or on the Monday 06:00 UTC scheduled run, whichever comes first.

Nothing needs to change in the repo for that to happen, since govulncheck resolves against a live advisory database. That is also why the job passed on `497d0297f` at 20:40Z and failed on byte-identical Go code at 22:17Z.

## The part that is not just CI noise

`release.yml:16` pins the same version, so **published agent binaries are currently built with a toolchain carrying these four advisories**, one of which is in `crypto/tls`:

| ID | Alias | Package |
|---|---|---|
| GO-2026-5972 | CVE-2026-33818 | `encoding/asn1` |
| GO-2026-6088 | CVE-2026-56859 | `encoding/xml` |
| GO-2026-6090 | CVE-2026-56862 | `crypto/tls` |
| GO-2026-6218 | CVE-2026-56860 | `net/url` |

I am deliberately not asserting the agent is exploitable through any of them. That depends on whether it parses untrusted ASN.1, XML or URLs on those paths, and I have not traced it. The narrow, checkable claim is only that the build toolchain is an affected version and that the same bump to `1.25.13` covers both the CI check and the release build.

Still happy to send the bump as its own PR if you want it off your plate. Not doing it unasked, since it touches five workflow files including `release.yml`.

bdunncompany:
Pushed `12abfcfcb`, which **removes the try/catch I added in the first commit**. An adversarial review caught that it did not do what its own comment claimed, and I confirmed it.

`sendQuote` runs inside the request-wide transaction opened by `withDbAccessContext` (`db/index.ts`) and installed by the auth middleware. So:

- a statement error on the marker write leaves that **Postgres transaction aborted**;
- catching the JavaScript rejection does **not** roll back to a savepoint;
- the unguarded re-select immediately below then fails with `current transaction is aborted`;
- the whole draft→sent claim rolls back regardless.

The catch bought nothing except obscuring where the failure started, and it introduced a failure mode that did not exist before my change. Worse, the comment asserting the failure was "swallowed" was materially misleading to anyone reading it later.

It is now a plain write, atomic with the status flip. That is the honest behaviour. The email having already left is a pre-existing property of sending inside the request transaction, not something this write introduces.

Worth flagging separately, since it is **not** mine to fix here: `resendQuote` has the same class of issue. Its write and re-select are both inside a catch with a fallback row, which is better, but an aborted ambient transaction still cannot be made to commit by catching in application code. If the intent is genuinely non-fatal bookkeeping there, it needs a `SAVEPOINT` or a write outside the email-carrying transaction.

Also tightened the successful-send test. It asserted only that no `.set()` payload carried `'send_failed'`, which a stray `{ sendEmailReason: null }` write would have passed. It now asserts exactly one `sendEmailReason` write occurs and that its value is null.

One limitation I am not going to paper over: the persistence test asserts the `.set()` payload, not the `WHERE` predicate, because the Drizzle chain mock records payloads only. It would still pass if the update targeted the wrong row or matched none. Proving the SQL targeting needs a route or integration test against a real database rather than this unit mock.

Re-verified after the change: `quoteLifecycle` 58/58, the 23 quote-related api suites 383/383, `tsc --noEmit` clean, and the control with the fix reverted still fails 4 tests including the persistence case.

bdunncompany:
Rebased onto `1d3783b6b` so CI reflects current `main` rather than the pre-merge base.

The two reds this PR was carrying were both stale and belonged to `main`, not to this change:

- `Trivy Filesystem Scan` — nanoid CVE-2026-67213, fixed by #3510 (merged 13:56:08Z)
- `Go Vulnerability Check` — the six Go stdlib advisories, fixed by #3523 (merged 13:56:14Z)

Both are now green on `main` at `1d3783b6b`, so a fresh run here should be clean. Re-verified after the rebase, nothing else changed in the diff.

bdunncompany:
## Adversarial review of this diff, and what it surfaced

Ran an independent adversarial review of the current diff (not a summary of it), prompted to assume the implementation is wrong and hunt for races, retry/idempotency failures, tenancy gaps, stale state, tests that prove too little, and regressions outside the path. Acting on what it found, and recording what I am **not** fixing here.

**Fixed in `d96c80d58`:** the outcome update matched on quote id alone. That is fine on the request path (forced RLS, plus `getQuote` already enforced actor access) but not for the scheduled-send worker, which calls `sendQuote` under a **system context where RLS is not the boundary**. Now scoped `id AND orgId`. No behavioural change today, purely defence-in-depth for the system-context caller.

## Known risks I am NOT fixing here, flagged rather than buried

**1. The email leaves inside the request transaction.** `sendQuote` claims draft→sent, then renders the PDF and calls the provider, all while the request-wide transaction from `withDbAccessContext` is still open. Any statement failure after delivery — deadlock victim, statement timeout, connection loss — rolls back `status='sent'`, `sentAt`, the token identity and the recipient rows, while the customer already has the email. A retry then legitimately re-claims the draft and sends a **second** copy.

This is **pre-existing** and not introduced here: the unguarded re-select below has always been able to fail the same way. But this PR does add one more statement inside that window, so it marginally widens it. The honest fix is a transactional outbox (claim + persist an attempt row, commit, let a worker deliver using the attempt id as the provider idempotency key), which is well outside a bug fix and is your call.

**2. `send_email_reason` is a lossy scalar with no attempt identity.** `resendQuote` has no claim or row lock, so two overlapping resends are last-commit-wins, not latest-attempt-wins: a slow failure can overwrite a newer success (UI falsely warns) or a slow success can clear a newer failure (UI falsely reassures). Also pre-existing, but this PR makes the direct-send path a third writer of that field, so it is fair to call it adjacent. A per-attempt id with `WHERE id = ? AND delivery_attempt_id = ?` would make the column truthful.

**3. The tests still cannot prove SQL targeting.** I said this when I opened the PR and the review sharpened it: the Drizzle mock records `.set()` payloads and its `.where()` neither records nor evaluates predicates. The persistence assertion would pass if production code omitted `.where()` entirely, used the wrong id, or matched zero rows — and the mocked re-select returns a row already containing `send_failed`, so it is not causally connected to the update. Proving this needs a real-Postgres test with two controlled resends plus an injected post-email failure, which belongs with the outbox work rather than here.

## Downstream readers worth knowing about

The review traced who now sees a populated column: `QuoteSendOutcomeBanners` and the post-flip toast in `QuoteActions.tsx` (so a stale value becomes a visible false claim, which is the point of risk 2), `tenantExportPolicyRegistry.ts` (already classified — the field was included before, it just carried mostly nulls, and will now carry real operational failure data in tenant exports), and the MCP/AI quote tools, which return raw service rows so agents can now reason from it. The public portal DTO builds an allowlisted response and does **not** expose it, so there is no portal leak.

Net: this PR makes the banner fire for the direct-send path, which is the reported bug. It does not make quote delivery idempotent, and I would rather say so plainly than let a green board imply otherwise.

bdunncompany:
**Correcting myself: `d96c80d58` was a regression, reverted in `bdb0e7e45`.** Head is back to the id-only write.

I added `eq(quotes.orgId, quote.orgId)` to the outcome update as defence-in-depth for the system-context worker. Sending that commit back for a second adversarial review — specifically asking whether the added predicate could match zero rows where the old one matched one — found that it could, and I confirmed it against the code:

- `quote` is read by `getQuote` **before** the draft→sent claim.
- `updateQuote` can reassign a draft's organization (`set.orgId = targetOrgId`), moving the child rows in the same transaction.
- The claim matches on **`id AND status='draft'` only**, no org.

So a concurrent org move leaves the claim succeeding against the moved row while my write, carrying the stale pre-read org, matches **zero rows and succeeds silently** — losing precisely the write this PR exists to add. The final re-select is id-only, so the caller still gets `emailReason` back while the persisted column stays null, which is the original bug wearing a different hat. Most reachable through the scheduled-send worker, where the whole claim and send run under a system context and forced RLS does not mask the mismatch.

Back to `where(eq(quotes.id, id))`, matching the claim's own predicate, so the write is bound to the row the claim actually took. The comment now says why, so the next person does not "harden" it the same way.

The underlying concern was legitimate: an id-only write under system context has only the id between it and the wrong row. The answer is an org (or attempt id) derived **from the claim** rather than from a pre-claim read, which belongs with the delivery-attempt work described in my previous comment rather than bolted on here.

Worth noting the shape of this: the first review suggested a change, the change was wrong, and only re-reviewing the fix itself caught it. Green tests said nothing either way — the Drizzle mock treats `.where()` as a no-op, so every one of the 383 passing assertions was equally happy with the broken predicate.

ToddHebebrand (commented):
Fix is right and the reasoning around it is careful — approving on the code. One correction to the comment it leaves behind, because that comment is written as forward-looking guidance and the count in it is wrong.

**The change itself: verified.** The `draft→sent` claim at `apps/api/src/services/quoteLifecycle.ts:170` sets `sendEmailReason: null` atomically with the status flip, so guarding the write on `if (emailReason)` is correct — a successful send genuinely needs no second write, and your "exactly one `.set()` carries sendEmailReason" test asserts that as a count rather than an absence, which is the right shape. The id-only predicate is also right for the reason you give: `quote` was read before the claim, `updateQuote` can move a draft's org, and an `orgId` term would silently match zero rows. Binding the write to the same predicate the claim took is correct, and RLS still scopes it.

**`resendQuote` does not exist.** The body's table lists it as a path that persisted before, and the rewritten `deliverQuoteEmail` doc comment now says "All three callers persist that reason to quotes.send_email_reason ... sendQuote, resendQuote, and the scheduled-send worker."

`git grep -in resendquote` over the repo returns nothing. `apps/api/src/routes/quotes/lifecycle.ts` registers `POST /:id/send`, `POST /:id/schedule-send` and `DELETE /:id/schedule-send` — there is no resend route and no resend service function. Outside the claim, `send_email_reason` has exactly one other writer: `apps/api/src/jobs/quoteSendQueue.ts:186`.

So it is **two** callers, not three, and the gap this PR closes was 1-of-2 rather than 1-of-3.

To be fair about provenance: the phantom is pre-existing — the comment you replaced already named `resendQuote`, so you inherited it rather than invented it. What changes is that the old comment was describing an asymmetry, while the new one is an instruction ("Keep it that way when adding a fourth"). A wrong count in an instruction sends the next person looking for a function that was never there, and "adding a fourth" would actually be a third.

**Asking for:** drop `resendQuote` from both the comment and the body table, and say two callers / "a third". Nothing else needs to change.

**Minor, no action needed:** the body describes a test named `writes no outcome marker when the send succeeds`; the diff has `writes exactly one sendEmailReason on a successful send: the claim clearing it`. The diff's version is the better test and the better name — just noting the drift so the body isn't read as the source of truth later.

Status: approving the fix; will merge once the comment/table correction lands.

ToddHebebrand (commented):
**Correction — withdraw the `resendQuote` finding entirely. It was my error, and the PR is right.**

`resendQuote` exists: `apps/api/src/services/quoteLifecycle.ts:591`, routed at `apps/api/src/routes/quotes/lifecycle.ts:131` as `POST /:id/resend`, and it persists the reason at `quoteLifecycle.ts:658`. So there are three callers, the body's table is accurate, and the comment saying "All three callers persist that reason" is correct as written. Nothing to change.

What happened: I ran `git grep` against my checked-out worktree instead of `origin/main`, and that branch predates the resend feature. A zero-hit grep on the wrong tree is not evidence of absence, and I posted it as though it were.

The rest of the review stands, re-checked against `origin/main` this time: the `draft→sent` claim at `quoteLifecycle.ts:191` clears `sendEmailReason`, so the `if (emailReason)` guard is correct; and matching on id alone is right for the reason you gave.

Approving. Merging once CI finishes.

PR #3509, opened 2026-08-14
fix(shared): accept null for the nullable monitoring watch fields

### LINKED ISSUE #3491: Editing a saved Monitoring policy fails to re-save: unset watch fields load as null but schema requires optional/undefin
Reported by **odobenusrosmarus_** in #bugs (Discord, message 1537226784796442755).

**Report (PII-redacted):**

When creating a new configuration policy, regardless of Partner or site specific, I am unable to add a service the Service & Process Monitoring when attempting to edit an existing policy. If creating a new policy, I can add multiple services without error, and the policy saves successfully. If I stay within the policy after saving, I can continue to add new watches and save without problems. Once I leave the policy, then try to re-access the policy and add a new watch I'm presented with the error information below. If I delete every single watch then I am able to add new watches without issue, I just have to re-add the watches I have deleted.

Steps to recreate: Create new Configuration Policy, add a service watch, save and leave the page, re-navigate back to Configuration Policy and edit the newly created policy. Add a new watch and save to generate error.

UI Output > 
Invalid monitoring settings: watches.0.displayName: Invalid input: expected string, received null; watches.0.cpuThresholdPercent: Invalid input: expected number, received null; watches.0.memoryThresholdMb: Invalid input: expected number, received null

Browser Console Output > 
{
    "error": "Invalid monitoring settings",
    "details": {
        "formErrors": [],
        "fieldErrors": {
            "watches.0.displayName": [
                "Invalid input: expected string, received null"
            ],
            "watches.0.cpuThresholdPercent": [
                "Invalid input: expected number, received null"
            ],
            "watches.0.memoryThresholdMb": [
                "Invalid input: expected number, received null"
            ]
        }
    },
    "issues": [
        {
            "expected": "string",
            "code": "invalid_type",
            "path": [
                "watches",
                0,
                "displayName"
            ],
            "message": "Invalid input: expected string, received null"
        },
        {
            "expected": "number",
            "code": "invalid_type",
            "path": [
                "watches",
                0,
                "cpuThresholdPercent"
            ],
            "message": "Invalid input: expected number, received null"
        },
        {
            "expected": "number",
            "code": "invalid_type",
            "path": [
                "watches",
                0,
                "memoryThresholdMb"
            ],
            "message": "Invalid input: expected number, received null"
        }
    ]
}

## Triage analysis

Reproduced the root cause. The Monitoring feature's watch schema (packages/shared/src/validators/index.ts, monitoringInlineSettingsSchema) declares displayName, cpuThresholdPercent, and memoryThresholdMb as `.optional()` (undefined-only) — not `.nullable()`. But the read path in apps/api/src/services/configurationPolicy.ts (decompose/serialize for 'monitoring', ~line 1022-1053) returns the raw DB column values for those fields verbatim, and unset optional fields are stored as SQL NULL (see the write path at ~line 554/559-560: `displayName: w.displayName ?? null`, `cpuThresholdPercent: w.cpuThresholdPercent ?? null`, `memoryThresholdMb: w.memoryThresholdMb ?? null`). So GET on an existing policy returns `displayName: null, cpuThresholdPercent: null, memoryThresholdMb: null` for any watch that never set them. The web editor (apps/web/src/components/configurationPolicies/featureTabs/MonitoringTab.tsx, readSettings ~line 148-162) loads that into React state unchanged (spread over defaultWatch, so the null survives), and per-field WatchCard onChange handlers only touch the field the user edits — untouched watches keep their `null` values in state. When the whole `settings.watches` array is POSTed back (handleSave / PUT feature link at apps/api/src/routes/configurationPolicies/featureLinks.ts lines 373-382, calling monitoringInlineSettingsSchema.safeParse), Zod rejects the pre-existing watches because `.optional()` schemas accept `undefined` but reject explicit `null`, producing exactly the reported 'expected string, received null' / 'expected number, received null' errors on watches.0. This only triggers on a re-opened, previously-saved policy (fresh new-policy state never contains nulls, matching the reporter's observation), and deleting all watches masks it because the empty array has nothing to fail on.

## Affected code

- `packages/shared/src/validators/index.ts:852-868` — monitoringInlineSettingsSchema: displayName/cpuThresholdPercent/memoryThresholdMb are `.optional()` (undefined-only), not `.nullable()`
- `apps/api/src/services/configurationPolicy.ts:1022-1053` — GET/serialize path returns raw DB column values (including null) for watch fields with no null->undefined normalization
- `apps/api/src/services/configurationPolicy.ts:540-568` — Insert path stores unset optional watch fields as SQL null (`?? null`), which round-trips back out on read
- `apps/web/src/components/configurationPolicies/featureTabs/MonitoringTab.tsx:148-162` — readSettings loads server watches (including any null fields) straight into editor state without stripping nulls
- `apps/api/src/routes/configurationPolicies/featureLinks.ts:373-382` — PUT feature-link route re-validates the full watches array with the same optional-only schema, rejecting the null-carrying pre-existing watches

## Potential fix

Either make displayName/cpuThresholdPercent/memoryThresholdMb `.nullable().optional()` in monitoringInlineSettingsSchema (matching how alertRuleItemSchema's autoResolveConditions already handles this), or normalize null->undefined for these fields in the GET serialization at configurationPolicy.ts ~line 1037-1051 before returning inlineSettings to the client.

### LINKED ISSUE #3492: Editing a saved Monitoring policy fails: stored null fields fail zod validation on re-save
Reported by **odobenusrosmarus_** in #bugs (Discord, message 1537226762386407434).

**Report (PII-redacted):**

When creating a new configuration policy, regardless of Partner or site specific, I am unable to add a service the Service & Process Monitoring when attempting to edit an existing policy. If creating a new policy, I can add multiple services without error, and the policy saves successfully. If I stay within the policy after saving, I can continue to add new watches and save without problems. Once I leave the policy, then try to re-access the policy and add a new watch I'm presented with the error information below. If I delete every single watch then I am able to add new watches without issue, I just have to re-add the watches I have deleted.

Steps to recreate: Create new Configuration Policy, add a service watch, save and leave the page, re-navigate back to Configuration Policy and edit the newly created policy. Add a new watch and save to generate error.

UI Output > 
Invalid monitoring settings: watches.0.displayName: Invalid input: expected string, received null; watches.0.cpuThresholdPercent: Invalid input: expected number, received null; watches.0.memoryThresholdMb: Invalid input: expected number, received null

## Triage analysis

Round-trip bug in the Monitoring feature of Configuration Policies. When a watch is saved without an optional field (displayName, cpuThresholdPercent, memoryThresholdMb), the API insert coerces the missing value to `?? null` before writing to configPolicyMonitoringWatches (configurationPolicy.ts:554,559,560). When the policy is reloaded for editing, the read path returns those DB nulls verbatim (configurationPolicy.ts:1040,1045,1046) with no null→undefined normalization. The web MonitoringTab then merges the loaded watch over its defaults with `{...defaultWatch, ...w}` (MonitoringTab.tsx:159), which preserves the explicit `null` (spread copies present keys regardless of value). On the next save, the whole watches array — including the pre-existing, now-null-bearing watches — is POSTed and validated against `monitoringInlineSettingsSchema` (packages/shared/src/validators/index.ts:857,862-863), where these three fields are `.optional()` but not `.nullable()`. Zod rejects the `null` with exactly the reported message ('Invalid input: expected string, received null' / 'received number'), and the route surfaces it as 'Invalid monitoring settings' (apps/api/src/routes/configurationPolicies/featureLinks.ts:210-216) — matching the poster's error text verbatim. This explains every observed detail: brand-new/in-session watches never round-trip through the DB so they stay `undefined` and save fine; after navigating away and back, previously-saved watches carry nulls and poison any subsequent save; deleting all watches removes the null-bearing entries so saving newly-added ones succeeds again.

## Affected code

- `apps/api/src/services/configurationPolicy.ts:546-567` — Insert coerces missing displayName/cpuThresholdPercent/memoryThresholdMb to explicit null in the DB
- `apps/api/src/services/configurationPolicy.ts:1022-1053` — Read path returns the stored null verbatim for these fields, no normalization back to undefined
- `apps/web/src/components/configurationPolicies/featureTabs/MonitoringTab.tsx:148-162` — readSettings spreads loaded watch (with null fields) over defaultWatch, preserving explicit null in editor state
- `packages/shared/src/validators/index.ts:852-868` — monitoringInlineSettingsSchema declares displayName/cpuThresholdPercent/memoryThresholdMb as .optional() but not .nullable(), so null fails validation
- `apps/api/src/routes/configurationPolicies/featureLinks.ts:209-221` — PUT handler runs monitoringInlineSettingsSchema.safeParse and returns 'Invalid monitoring settings' with the zod issue list on failure — matches reported error text

## Potential fix

Either make displayName/cpuThresholdPercent/memoryThresholdMb `.nullable().optional()` in monitoringInlineSettingsSchema (packages/shared/src/validators/index.ts), or normalize DB nulls back to undefined when reading watches in configurationPolicy.ts's monitoring case (~line 1037-1051), so previously-saved watches don't reintroduce nulls into the editor payload.

bdunncompany:
Fixes #3491. Also fixes #3492, which is the same report from the same reporter (two Discord messages, identical error text and repro).

## What broke

Adding a watch to an **existing** Monitoring policy failed with:

```
watches.0.displayName: Invalid input: expected string, received null
watches.0.cpuThresholdPercent: Invalid input: expected number, received null
watches.0.memoryThresholdMb: Invalid input: expected number, received null
```

Those three fields are exactly the only columns on `config_policy_monitoring_watches` declared without `NOT NULL`:

```ts
displayName: varchar('display_name', { length: 255 }),   // nullable
cpuThresholdPercent: real('cpu_threshold_percent'),      // nullable
memoryThresholdMb: real('memory_threshold_mb'),          // nullable
```

Every other watch column is `.notNull().default(...)`, which is why the error names these three and nothing else.

The round trip: the write path stores an unset value as `?? null`, the read path returns that null verbatim, and the editor merges the loaded watch over its defaults so the null survives into editor state and is posted straight back. `monitoringInlineSettingsSchema` declared all three `.optional()` but not `.nullable()`, so **zod rejected the API's own stored representation**.

That accounts for every symptom in the report, including the odd ones: a brand-new watch has never round-tripped through the DB so it stays `undefined` and saves fine; a policy reopened after navigating away carries nulls and poisons any later save; deleting every watch clears the null-bearing entries so saving newly-added ones works again.

## Why fix the schema rather than normalise on read

The columns genuinely are nullable, and null is what the API already persists and returns, so the request schema should accept the shape the API itself emits. Normalising null to undefined on read would change the response contract for existing consumers and still leave the schema rejecting a value the database legitimately holds.

It also fixes callers beyond the web editor. Any MCP or direct-API client that reads a policy and posts it back hits the same wall, and a web-only fix would leave those broken.

Downstream treatment is already null-safe, which is what makes this safe to widen:

- `apps/api/src/routes/agents/helpers.ts` gates delivery on `if (w.cpuThresholdPercent != null)` / `!= null`, loose equality, so null and undefined behave identically.
- `displayName` is never sent to the agent at all.

This matches the existing precedent on `updateDeviceSchema.displayName` in the same file, which is `.nullable().optional()` with a comment for exactly this reason.

## Tests

- `should accept a saved watch round-tripped with null optional fields` reproduces the bug, and asserts the values stay null rather than being silently coerced.
- `should still reject a wrong-typed threshold` pins that widening to nullable did not weaken type or range checking (`'high'` and `101` both still rejected). Being precise: this one passes with or without the fix, so it is a regression guard, not a reproducer.

## Verification

Run locally on node 22.23.2 (matching `.nvmrc`), deps installed `--frozen-lockfile`:

```
vitest packages/shared (full)      1750 passed (1750), 69 files
tsc --noEmit packages/shared       exit 0, no output
tsc --noEmit apps/api              exit 0, no output
```

Control run, fix reverted and tests kept: **1 failed**, the new round-trip test, and nothing else.

`apps/api` typechecking clean is the meaningful signal that widening the schema type did not break a consumer. No web file imports this schema.

## One observation, deliberately not fixed here

`MonitoringTab.tsx` types these fields locally as `displayName?: string` and `cpuThresholdPercent?: number`, but the API has always been able to hand it null. That type inaccuracy predates this change and is unaffected by it, so I left it alone rather than widen the scope. Worth a separate look if you want the editor types to match the wire shape.

### DIFF
--- a/packages/shared/src/validators/index.ts
+++ b/packages/shared/src/validators/index.ts
@@ -925,13 +925,22 @@ export const monitoringInlineSettingsSchema = z.object({
   watches: z.array(z.object({
     watchType: z.enum(['service', 'process']),
     name: z.string().min(1).max(255),
-    displayName: z.string().max(255).optional(),
+    // These three are the only watch columns without NOT NULL
+    // (config_policy_monitoring_watches.display_name / cpu_threshold_percent /
+    // memory_threshold_mb). The write path stores an unset value as `?? null`
+    // and the read path returns that null verbatim, so the editor loads a saved
+    // watch carrying nulls and posts them straight back. `.optional()` alone
+    // rejected them, which made an existing policy impossible to re-save once
+    // any watch had an unset field (#3491, #3492). Keep in sync with the
+    // columns, same as updateDeviceSchema.displayName above. Consumers already
+    // treat null and undefined alike (`!= null` in agents/helpers.ts).
+    displayName: z.string().max(255).nullable().optional(),
     enabled: z.boolean().default(true),
     alertOnStop: z.boolean().default(true),
     alertAfterConsecutiveFailures: z.number().int().min(1).max(100).default(2),
     alertSeverity: z.enum(['critical', 'high', 'medium', 'low', 'info']).default('high'),
-    cpuThresholdPercent: z.number().min(0).max(100).optional(),
-    memoryThresholdMb: z.number().min(0).optional(),
+    cpuThresholdPercent: z.number().min(0).max(100).nullable().optional(),
+    memoryThresholdMb: z.number().min(0).nullable().optional(),
     thresholdDurationSeconds: z.number().int().min(0).max(86400).default(300),
     autoRestart: z.boolean().default(false),
     maxRestartAttempts: z.number().int().min(0).max(50).default(3),
--- a/packages/shared/src/validators/index_inline_settings.test.ts
+++ b/packages/shared/src/validators/index_inline_settings.test.ts
@@ -405,6 +405,48 @@ describe('monitoringInlineSettingsSchema', () => {
     expect(result.success).toBe(true);
   });
 
+  // #3491/#3492: a saved watch loads back from the DB with nulls in the three
+  // nullable columns, and the editor posts them straight back. Rejecting null
+  // made an existing policy impossible to re-save once any watch had an unset
+  // field — the reported error named exactly these three paths.
+  it('should accept a saved watch round-tripped with null optional fields', () => {
+    const result = monitoringInlineSettingsSchema.safeParse({
+      watches: [
+        {
+          watchType: 'service',
+          name: 'wuauserv',
+          displayName: null,
+          cpuThresholdPercent: null,
+          memoryThresholdMb: null,
+        },
+      ],
+    });
+    expect(result.success).toBe(true);
+    if (result.success) {
+      // Preserved as null rather than coerced: the write path stores `?? null`
+      // and the agent delivery check is `!= null`, so both spellings behave
+      // identically downstream.
+      const watch = result.data.watches[0];
+      expect(watch?.displayName).toBeNull();
+      expect(watch?.cpuThresholdPercent).toBeNull();
+      expect(watch?.memoryThresholdMb).toBeNull();
+    }
+  });
+
+  it('should still reject a wrong-typed threshold, not just anything non-number', () => {
+    expect(
+      monitoringInlineSettingsSchema.safeParse({
+        watches: [{ watchType: 'process', name: 'nginx', cpuThresholdPercent: 'high' }],
+      }).success
+    ).toBe(false);
+    // and the range check survives the nullable change
+    expect(
+      monitoringInlineSettingsSchema.safeParse({
+        watches: [{ watchType: 'process', name: 'nginx', cpuThresholdPercent: 101 }],
+      }).success
+    ).toBe(false);
+  });
+
   it('should reject checkIntervalSeconds below 10', () => {
     expect(
       monitoringInlineSettingsSchema.safeParse({ checkIntervalSeconds: 9 }).success
--- a/packages/shared/src/validators/index_inline_settings.test.ts
+++ b/packages/shared/src/validators/index_inline_settings.test.ts
@@ -433,16 +433,20 @@ describe('monitoringInlineSettingsSchema', () => {
     }
   });
 
-  it('should still reject a wrong-typed threshold, not just anything non-number', () => {
+  // Covers all three widened fields, not just one: nullable must not become
+  // "accepts anything". Type and range checks have to survive the change.
+  it.each([
+    ['displayName wrong type', { displayName: 42 }],
+    ['displayName too long', { displayName: 'x'.repeat(256) }],
+    ['cpuThresholdPercent wrong type', { cpuThresholdPercent: 'high' }],
+    ['cpuThresholdPercent above max', { cpuThresholdPercent: 101 }],
+    ['cpuThresholdPercent below min', { cpuThresholdPercent: -1 }],
+    ['memoryThresholdMb wrong type', { memoryThresholdMb: 'lots' }],
+    ['memoryThresholdMb below min', { memoryThresholdMb: -1 }],
+  ])('still rejects %s after the nullable widening', (_label, patch) => {
     expect(
       monitoringInlineSettingsSchema.safeParse({
-        watches: [{ watchType: 'process', name: 'nginx', cpuThresholdPercent: 'high' }],
-      }).success
-    ).toBe(false);
-    // and the range check survives the nullable change
-    expect(
-      monitoringInlineSettingsSchema.safeParse({
-        watches: [{ watchType: 'process', name: 'nginx', cpuThresholdPercent: 101 }],
+        watches: [{ watchType: 'process', name: 'nginx', ...patch }],
       }).success
     ).toBe(false);
   });


bdunncompany:
Pushed `e47160a3a` after an adversarial review of this diff. Two outcomes worth reporting.

**The implementation held up.** The reviewer traced the full save path and found no second layer that still rejects null, which is the thing I had not proven: `updateFeatureLinkSchema` passes `inlineSettings` through as a permissive record, the PATCH handler validates with the changed schema, and then `assertDecomposableInlineSettings` and `decomposeInlineSettings` each parse it again with that same schema before `?? null` maps both spellings to SQL null. Three parse points, all the shared schema, no route-local or service-local duplicate. I had only traced one of those three.

It also confirmed the no-regression claim independently: agent delivery gates on loose `!= null`, `displayName` is absent from `MonitoringWatchConfig` so it never reaches the Go agent, and the editor renders null and undefined identically via `?? ""`. One thing it surfaced that I had not checked: `aiAgentSdkTools.ts` has its own monitoring shape whose three fields are still non-nullable optional, but that tool is read-only now and writes go through `manage_policy_feature_link`, whose `inlineSettings` is validated by the changed shared schema. So it is not a conflicting save path.

**My test claim was overstated, and that is now fixed.** The guard test was named as if it pinned that widening to nullable had not weakened validation generally, while actually asserting only two `cpuThresholdPercent` cases. Replaced with a table covering all three widened fields: wrong type and length on `displayName`, wrong type and both range bounds on `cpuThresholdPercent`, wrong type and lower bound on `memoryThresholdMb`.

**One limitation I am not going to paper over.** The round-trip test calls `monitoringInlineSettingsSchema.safeParse()` directly, so it pins the schema rather than the reported failing save path. It would still pass if a future route-local schema, a DB constraint, or a change in `decomposeInlineSettings` reintroduced the rejection. Properly pinning the reported bug wants a route-level test that PATCHes a monitoring feature link with all three fields null and asserts the normalized insert receives nulls, ideally then replaying the stored row. That needs the integration harness rather than this unit mock, and I did not add it here. Happy to if you want it in this PR.

Re-verified after the test change: shared suite **1756/1756** across 69 files, `tsc --noEmit` clean. (The commit message says 1754; the correct count is 1756, since the table expands to more cases than I estimated when writing it.)

bdunncompany:
Rebased onto `1d3783b6b` so CI reflects current `main` rather than the pre-merge base.

The two reds this PR was carrying were both stale and belonged to `main`, not to this change:

- `Trivy Filesystem Scan` — nanoid CVE-2026-67213, fixed by #3510 (merged 13:56:08Z)
- `Go Vulnerability Check` — the six Go stdlib advisories, fixed by #3523 (merged 13:56:14Z)

Both are now green on `main` at `1d3783b6b`, so a fresh run here should be clean. Re-verified after the rebase, nothing else changed in the diff.

bdunncompany:
## Adversarial review of this diff, and one consequence worth stating plainly

Ran an independent adversarial review of the current diff, prompted to assume it is wrong and hunt for protocol/version skew, tenancy, stale state, tests that prove too little, and regressions outside the path.

**It confirmed the parts I claimed**, with a trace I had not done end to end:

- **No agent wire regression.** `resolveDeviceMonitoringSettings` builds a fresh wire object and only adds thresholds when `!= null`, and `displayName` is never sent, so neither null nor undefined reaches an old agent. The Go structs use scalar `float64` and `ParseMonitorConfig` is an ordinary JSON round trip, so there is no old/new skew here.
- **No new cross-tenant primitive.** Partner-wide visibility and ownership are derived independently of these values.
- **No tenant-export-policy change**, since the core registry does not directly serialize these child tables.

## The consequence I did not consider, and am not fixing here

`PATCH` on a feature link is **delete-and-reinsert**, not a merge, and there is no version, ETag or `updatedAt` precondition. So:

1. Tech A loads a watch whose `cpuThresholdPercent` is null.
2. Tech B sets that threshold to 80 and saves.
3. Tech A, still holding the stale document, changes something unrelated and saves.
4. The null now **passes validation**, the rows are dropped and recreated, and B's threshold is silently gone.

Before this change that stale save would have 400'd. So this PR removes an *accidental* partial mitigation for a lost update.

I still think the change is right, and want to be explicit about why rather than hand-wave it:

- The 400 was never a concurrency guard. It was a bug that made an existing policy **impossible to re-save at all**, which is the reported issue.
- The lost-update hazard is inherent to delete-and-reinsert and is not created here: an **omitted** property is written as SQL NULL by the same code path, so a client that simply drops the key does the same damage today, with no 400 to stop it.
- Normalising null to undefined on read would not fix it either, for that same reason.

The real fix is optimistic concurrency (or per-watch patching) on the feature-link PATCH. That is a separate change and I have not made it. Happy to file it if you want it tracked.

## Two smaller things

- **The tests still only pin the schema.** They call `monitoringInlineSettingsSchema.safeParse()` directly and never traverse route → `assertDecomposableInlineSettings` → `decomposeInlineSettings` → insert. They would miss a regression where PATCH returns 200 but the normalized rows are cleared and not recreated. Proving the reported bug end to end needs the integration harness.
- **`aiAgentSdkTools.ts` still declares these three fields as non-nullable optionals**, so its schema now disagrees with what the API accepts and returns. That tool is advertised read-only by its handler, so it is a contract inconsistency rather than a live defect. Left alone to keep this diff to the fix; say the word and I will include it.

PR #3497, opened 2026-08-13
feat(i18n): add Turkish tr-TR localization

LewisLosa:
## Summary

- Adds complete Turkish (`tr-TR`) web locale catalogs across the existing 22 namespaces.
- Registers Turkish in locale preferences, user and partner validation, and language selectors.
- Extends locale parity, translation coverage, terminology, and extraction checks for `tr-TR`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor (no behavior change)
- [ ] Documentation
- [ ] CI / build / tooling
- [ ] Other: 

## Testing

- [ ] Existing tests pass (`pnpm test` / `make test`)
- [x] Added new tests
- [x] Manual testing (describe below)

- Verified Turkish selection and persistence in the user and partner settings surfaces.
- Ran focused web i18n, locale-preference, and settings tests, plus API user/organization route tests.

## Checklist

- [x] My code follows the project's code style
- [x] I have reviewed my own changes
- [x] I have tested on the relevant platforms
- [x] No secrets, credentials, or personal data included

PR #3510, opened 2026-08-14
fix(deps): pin nanoid >= 3.3.18 to clear CVE-2026-67213 (HIGH)

bdunncompany:
`Trivy Filesystem Scan` is failing on every open PR and will fail `main` at its next run. No repo change caused it: Trivy resolves against a live vulnerability database, so the identical lockfile **passed at 22:17Z and failed at 01:58Z**.

```
Library │ Vulnerability  │ Severity │ Installed │ Fixed         │ Title
nanoid  │ CVE-2026-67213 │ HIGH     │ 3.3.17    │ 3.3.18, 5.1.6 │ DoS via infinite loop in random ID generation
```

## This is #3241 again, one version further along

The same advisory was already handled on 2026-08-08 by #3241, which added:

```json
"nanoid@<3.3.17": ">=3.3.17 <4.0.0"
```

The advisory has since moved its fixed version from 3.3.17 to 3.3.18, so that override now pins to **exactly the version being flagged**. This updates the existing entry in place rather than adding a second overlapping one:

```diff
-      "nanoid@<3.3.17": ">=3.3.17 <4.0.0"
+      "nanoid@<3.3.18": ">=3.3.18 <4.0.0"
```

Every version matched by the old key is still matched by the new one, so nothing previously constrained becomes unconstrained.

## Why the `<4.0.0` bound stays

Correcting my own first draft of this PR, which justified the bound as "the 3.x consumers are CommonJS". That is too strong: `@react-navigation/*` is ESM-oriented and postcss ships both entry points.

The accurate reason is that **every parent pulling the 3.x line asks for major 3 explicitly** (`@react-navigation/core|native|routers` at `^3.3.11`, postcss at `^3.3.12` / `^3.3.16` / `^3.3.17`). An unbounded `">=3.3.18"` resolves them across majors in violation of their declared ranges. That nanoid 5 is ESM-only while nanoid 3 ships both `import` and `require` entry points is a supporting reason, not the primary one.

The selector matches only resolutions below 3.3.18, so it cannot block `oidc-provider`'s 5.1.16 or the API's direct `^6.0.0`.

## Result

Tree resolves to **3.3.18, 5.1.16, 6.0.0** only, each at or above its fixed version (5.1.16 > 5.1.6 numerically, not lexically).

## One non-nanoid line in the lockfile, explained

The regeneration also drops `bufferutil` and `utf-8-validate` from one `transitivePeerDependencies` list. Flagging it rather than leaving it to be spotted in review: it is a consequence of removing the `nanoid@3.3.17` record, both are **optional** peer dependencies of `ws` that are never installed, and both remain declared on the `ws` records themselves. Regenerating the lockfile with an unchanged `package.json` produces no diff at all, so this is caused by the override change and not by unrelated drift.

## Verification

node 22.23.2, pnpm 10.34.5 (the `packageManager` pin):

```
pnpm install --frozen-lockfile        exit 0
nanoid refs below 3.3.18 in lockfile  0   (followed snapshot edges, not just the version list)
packages/shared full suite            1748 passed (1748), 69 files
```

I also confirmed the override is actually enforced rather than assumed, by temporarily forcing `nanoid` to `3.3.11` and watching the lockfile follow. Natural resolution would never select that, since postcss requests `^3.3.17`.

Independent of my other PRs and mergeable on its own. Unblocks Trivy for every open PR, including the external Turkish localization PR #3497.

Not addressed here: the four Go stdlib advisories failing `Go Vulnerability Check`, which need `GO_VERSION` at 1.25.13 across five workflow files. Details on #3507. Separate change, separate blast radius.


bdunncompany:
**The `Trivy Filesystem Scan` red on this PR is not a finding.** Trivy never scanned anything:

```
FATAL  init error: DB error: failed to download vulnerability DB:
       failed to download artifact from mirror.gcr.io/aquasec/trivy-db:2
       GET https://mirror.gcr.io/v2/: unexpected status code 403 Forbidden
```

The registry mirror rate-limited the DB download. `Trivy Image Scan` succeeded on the same commit, so this is transient rather than a config problem. It will need a re-run to give a real signal, which I cannot trigger.

## Verified locally instead, with a control

Since CI could not answer it, I ran the same scanner version (`aquasec/trivy:0.70.0`, matching the `v0.70` docs the CI job cites) against two lockfiles in isolation:

| lockfile scanned | HIGH findings |
|---|---|
| `origin/main` | **3** — `image-size` CVE-2025-71329, CVE-2025-71330, **`nanoid` CVE-2026-67213** |
| this branch | **2** — the two `image-size` ones only, **nanoid gone** |

The control is the point: the scanner demonstrably still detects the finding on `main`, so the clean result here is the change working rather than a scan that found nothing for an unrelated reason.

The two remaining `image-size` advisories are pre-existing, carry no fixed version (`affected`, not `fixed`), and are already suppressed by `.trivyignore` from #3241. They appear above only because my isolated scratch directory has no `.trivyignore`. Scanning the full worktree, where `.trivyignore` is present, reports `pnpm-lock.yaml  pnpm  0` and exits 0.

Worth recording for the next person verifying a `.trivyignore` change by hand, since it cost time both in #3240 and here: **Trivy loads `.trivyignore` from the working directory, not from the scan target.**

## Also, this PR was wrong when first opened

Correcting my own error rather than quietly force-pushing over it. The first version added a **second, overlapping** nanoid override while `main` already carried one from #3241, and it was built on a stale branch that had been sitting on the fork since 2026-08-07, which is why it opened as `CONFLICTING` with zero check runs. Rebuilt on current `main`, updating the existing entry in place. Current head is `cf1f7a173`.

bdunncompany:
Re-ran the failed job and **`Trivy Filesystem Scan` now passes** on `cf1f7a173`. The earlier red was the `mirror.gcr.io` 403 on the vulnerability-DB download, not a finding, and it was transient: the runs on #3507 and #3509 minutes later downloaded the DB fine and reported the nanoid finding normally.

That gives the comparison this PR needed, straight from CI rather than my laptop:

| PR | carries this fix | Trivy Filesystem Scan |
|---|---|---|
| #3507 | no | **fails** — `Total: 1 (HIGH: 1)`, nanoid CVE-2026-67213 |
| #3509 | no | **fails** — same single finding |
| **#3510** | **yes** | **passes** |

So the finding is real and repo-wide, and this one-line override change is what clears it. That matches the local control I ran earlier (main's lockfile 3 HIGH, this branch 2 HIGH with nanoid gone, the remaining two being the pre-existing `image-size` advisories already suppressed by `.trivyignore`).

The only remaining red here is `Go Vulnerability Check`, which is the separate Go stdlib toolchain issue detailed on #3507 and is not addressed by this PR.

Issue #1377, opened 2026-06-15
Software Inventory returns "Request body too large" on self-hosted deployment


win-wxx:
### Description

Hi Team,

I have successfully deployed Breeze as a self-hosted installation on my VPS, and most features appear to be working correctly.

However, I'm experiencing an issue with the Software Inventory feature. When I attempt to upload an application to the software inventory, Breeze returns the following error:

"Request body too large"

I'm relatively new to Breeze and self-hosted environments, so I'm not sure whether this is caused by a reverse proxy configuration, container setting, or an application limit.

Could you please advise on the correct configuration or troubleshooting steps to resolve this issue?

Environment:

Self-hosted Breeze deployment
VPS hosting: Vultr
Reverse proxy: Caddy
Breeze version: 0.70.0

<img width="492" height="485" alt="Image" src="https://github.com/user-attachments/assets/e4027bac-1d97-4941-bad1-7fd00712ab80" />

### Steps to Reproduce

.

### Expected Behavior

Uploadable and can be use properly

### Actual Behavior

When I attempt to upload an application to the software inventory, Breeze returns the following error:

"Request body too large"

### Component

Web Dashboard

### Version

0.70.0

### Relevant Logs

```shell

```

bdunncompany:
Reproduced from the code paths on `main` (v0.70.0). This is an application-level bug, not your Caddy or container config, so there isn't a reverse-proxy setting that will fix it on your side.

**What's happening:** Adding a package file ("Upload" in the software library version manager) posts the installer as multipart form-data to `POST /api/v1/software/catalog/:id/versions/upload`. That route is designed to accept files up to **500 MB** (`MAX_UPLOAD_SIZE` in `apps/api/src/routes/software.ts`). But the global body-size middleware in `apps/api/src/index.ts` applies a **1 MB** default cap to every route that isn't explicitly carved out:

```ts
// apps/api/src/index.ts (last rule, the global default)
return bodyLimit({ maxSize: 1024 * 1024, onError: (ctx) => ctx.json({ error: 'Request body too large' }, 413) })(c, next);
```

[...]

PR #3511, opened 2026-08-14
fix(api): carve out body limits for quote/catalog/avatar/contract uploads (#3482)

### LINKED ISSUE #3482: Quote image upload 413s far below the advertised 5 MB limit
## Description
While recreating a large (21-page) external proposal as a Breeze quote, uploading images for quote blocks failed unexpectedly.

The quote editor's image pickers (block image + cover image) advertise "PNG, JPEG, or WebP, up to 5 MB" in the UI copy. In practice, `POST /api/v1/quotes/:id/images` returns HTTP 413 "Request body too large" for a 1.2 MB PNG. A ~440 KB PNG for the same block succeeds.

## Root Cause (suspected)
Likely a server/proxy-level JSON-or-multipart body size limit set near ~1 MB that this route inherits from a shared default, rather than a per-route override matching the advertised 5 MB image limit. A very similar body-size ceiling was previously hit and fixed for a different upload path (see #1377, "Software Inventory returns Request body too large" — different route, same class of bug), suggesting this is a recurring pattern of routes inheriting a low default body limit rather than opting into one sized for their actual payload.

## Proposed Fix
Raise the body size limit for the quote-image upload route to comfortably cover 5 MB (plus base64/multipart encoding overhead, which can add ~33%), or, if 5 MB was never actually intended, correct the UI copy to reflect the real limit.

## Affected Files
- Quote image upload route (`POST /api/v1/quotes/:id/images`) and its body-size middleware/config
- Quote editor image picker UI copy (block image + cover image)

## Evidence
Reproduced 2026-08-12 on hosted US: 440 KB PNG upload succeeds, 1.2 MB PNG upload on the same block returns 413.

ToddHebebrand:
Closes #3482

## Problem

The quote editor advertises **"PNG, JPEG, or WebP, up to 5 MB"** for block and cover images, and `POST /api/v1/quotes/:id/images` registers its own `bodyLimit` at 5 MB + 64 KB. In practice a 1.2 MB PNG returned `413 {"error":"Request body too large"}` while a 440 KB PNG succeeded.

## Root cause

The global body gate is registered as `app.use('*', ...)` in `apps/api/src/index.ts:505` — long before any sub-app is mounted. It **always** runs first, so a `bodyLimit()` declared on an individual route is unreachable above whatever `bodyLimitForPath()` grants that path. Quote images had no carve-out, so they inherited the tight 1 MB default, and the error string in the report is verbatim that default's message.

This is the same shape as #1377 (software installer uploads), which was fixed with a carve-out in the same table.

## Fix

Carve-outs in `bodyLimitForPath()` for the routes whose own limits were unreachable, each sized to the route's own limit so the route keeps answering with its **specific** message instead of the generic one:

| Route | Route-level limit | Was effectively | Now |
|---|---|---|---|
| `POST /api/v1/quotes/:id/images` | 5 MB + 64 KB | 1 MB | 5 MB + 64 KB |
| `POST /api/v1/catalog/:id/image` | 5 MB + 64 KB | 1 MB | 5 MB + 64 KB |
| `POST /api/v1/users/me/avatar` | 5 MB + 64 KB | 1 MB | 5 MB + 64 KB |
| `POST /api/v1/contracts/contract-templates/:id/versions/upload` | 10 MB + 64 KB | 1 MB | 10 MB + 64 KB |

The three siblings had shipped the identical unreachable-limit bug (catalog product images and avatars also advertise 5 MB in the UI); fixing only quotes would leave three known-broken uploads behind.

The UI copy is correct as written — 5 MB was genuinely intended, enforced by `MAX_QUOTE_IMAGE_SIZE_BYTES`, the route's `file.size` check, and `quoteImageStorage`'s buffer check — so no copy change is needed.

## Deliberately not changed

`agents/heartbeat.ts`, `agents/inventory.ts` and `agents/connections.ts` declare route-level 5 MB limits that the 1 MB gate also shadows. Agent ingest staying tight is intentional (an existing test already asserts heartbeat sits at 1 MB), and raising it is a capacity/DoS decision, not a bugfix. That decision is now recorded explicitly rather than left implicit.

## Guard against recurrence

This class has now shipped twice (#1377, #3482) plus three unreported siblings, because nothing connects a route-level `bodyLimit` to the global table. Added a drift guard in `bodyLimit.test.ts` that:

1. scans `apps/api/src/routes/**` for every file registering its own `bodyLimit({...})` and fails if one isn't recorded — a new route-level limit can't be added without deciding whether it needs a carve-out; and
2. asserts, for each recorded path, that `bodyLimitForPath()` actually grants the limit the entry claims (including the intentionally-capped agent routes).

## Verification

- `vitest run src/middleware/bodyLimit.test.ts src/routes/quotes/lifecycle.test.ts src/services/quoteImageStorage.test.ts` — 59 passed (3 files).
- **Mutation-verified**: breaking the quotes carve-out fails both the new regression test and the drift guard (2 failed / 11 passed), so neither assertion is vacuous.
- `tsc --noEmit` clean for `@breeze/api`.

### DIFF
--- a/apps/api/src/middleware/bodyLimit.test.ts
+++ b/apps/api/src/middleware/bodyLimit.test.ts
@@ -1,7 +1,10 @@
+import { readdirSync, readFileSync } from 'node:fs';
+import { join, relative } from 'node:path';
 import { describe, it, expect } from 'vitest';
 import { bodyLimitForPath } from './bodyLimit';
 
 const MB = 1024 * 1024;
+const KB = 1024;
 
 describe('bodyLimitForPath', () => {
   it('applies the tight 1MB default to ordinary routes', () => {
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `code-reviewer` (path/mount correctness, over-matching, DoS surface, conventions) + `silent-failure-hunter` (vacuous-guard analysis, agent-side error handling, client-visible failure shape).

**Findings:** 7 raised → all resolved.
- *Blocking-ish (silent-failure-hunter):* the drift guard's `INTENTIONALLY capped` note on the agent ingest routes asserted intent that is not on record anywhere — and the gap it locked in is a real silent failure (a >1MB software inventory 413s, `isRetryableStatus` excludes 413, and `sendSoftwareInventory`'s error is discarded, so inventory goes permanently stale). Relabelled as **KNOWN GAP** pointing at #3516, which carries the full analysis. Fixed in `076069b`.
- *silent-failure-hunter:* the gate's `onError` logs nothing, so every 413 it has ever raised is invisible server-side — which is why this class keeps reaching users first. Deliberately **not** fixed here: the obvious fix logs `c.req.path`, which violates the repo's no-raw-path rule (`requestPathLogger` / `safeMatchedRouteLabel`), and the gate runs before routing so no safe label exists yet. Filed as #3517 with the design options.
- *code-reviewer:* guard scanned only `src/routes`, missing route surfaces mounted from `modules/` and `extensions/` → widened to `src/` with `index.ts`/`bodyLimit.ts` exempted, plus a non-empty assertion so a broken scan root fails loudly instead of comparing two empty lists. Scan regex loosened `bodyLimit(\s*\{` → `bodyLimit(` so `bodyLimit(opts)` can't hide. Both in `6745a4a`/`076069b`.
- *code-reviewer:* `/users/me/avatar` given its own branch preserving `'Avatar file too large (max 5 MB)'` — at equal thresholds this gate answers first, so the route's wording was unreachable. `agents/inventory.ts` entry listed 4 of 5 registrations (`/warranty-info` missing). Comment corrected: a route-level limit is reached, it just can't loosen a path.
- *code-reviewer, verified no action:* all four carve-out paths traced to their real mounts; `[^/]+` can't span `/` so `quotes/public` can't collide; `c.req.path` excludes the query string and isn't percent-decoded, and both mismatch directions fall through to the *tighter* default; Hono's `bodyLimit` short-circuits on `Content-Length` without buffering, and all four routes sit behind auth + permission middleware.

**Tests:** `vitest run src/middleware/bodyLimit.test.ts` — 13 passed; with `routes/quotes/lifecycle.test.ts` + `services/quoteImageStorage.test.ts` — 59 passed (3 files). `tsc --noEmit` clean for `@breeze/api` (needs `NODE_OPTIONS=--max-old-space-size=8192`). **Mutation-verified:** reverting the quotes carve-out fails both the regression test and the drift guard (2 failed / 11 passed), so neither is vacuous.

**CI:** re-running on `076069b`. On the previous push `Go Vulnerability Check` and `Trivy Filesystem Scan` failed — both are live-CVE-database scans against `agent/` (`GO-2026-5942/5972/6088/6090/6218` unallowlisted), and this PR's diff is two TypeScript files with zero Go. Pre-existing/environmental, not caused by this change; they need an allowlist refresh on `main` independently.

**Status:** review-clean, awaiting CI + maintainer merge. Not merged, issue #3482 left open.

ToddHebebrand:
**CI final on `076069b`:** 52 pass / 2 fail.

Every code-relevant job is green — `Test API`, `Type Check`, `Lint`, `Test Web`, `Build Web`, `Smoke Test`, all four `Integration Tests` shards, and both `Check Migrations` jobs.

The two failures are `Go Vulnerability Check` and `Trivy Filesystem Scan`, both live-CVE-database scans against `agent/`. `govulncheck` reports `GO-2026-5026/5942/5972/6088/6090/6218` as unallowlisted. This PR's diff is two TypeScript files with zero Go, and the same workflow passed on `main` at this PR's exact base commit (`497d029`) a few hours earlier — so these are newly-published advisories, not a regression from this change. They need an allowlist refresh (or dependency bump) on `main` independently, and will block the required-check gate until that lands.

ToddHebebrand:

|---|---|---|
| both | pass | pass |

No action needed on this PR.

Issue #2217, opened 2026-07-05
[API][Agent] Full patch scan tombstones device_patches for providers that didn't run — third-party counts drop to 0


ToddHebebrand:
## Description
Per-device third-party patch counts (3RD-PARTY column on Patch Compliance) show 0/– on every Windows device even after a successful winget scan previously populated them. The global Patches catalog still lists the winget entries (source=third_party, Approval=Pending), so the two surfaces disagree.

## Root Cause
The agent's full patch inventory upload is a snapshot: `sendPatchInventoryData(..., full=true)` (`agent/internal/heartbeat/heartbeat.go:1876-1911`) → `PUT /:id/patches/pending` marks **every** `status='pending'` device_patches row `'missing'` before re-upserting only what the current scan contained (`apps/api/src/routes/agents/patches.ts:315-322`, `markPendingDevicePatchesMissing`).

The winget provider returns empty (not an error) whenever no user-helper session is connected (`agent/internal/patching/winget.go:68-71`). So one scan with a logged-in user creates the third-party rows; the next scan without a session sweeps them all to `'missing'`. Since `OUTSTANDING_DEVICE_PATCH_STATUSES = ['pending']` (`apps/api/src/db/schema/patches.ts:68`), the compliance count (`apps/api/src/routes/patches/compliance.ts:290`) and the ring approval evaluator (`apps/api/src/services/patchApprovalEvaluator.ts:331-352`) both stop seeing them. "Provider couldn't run" is being treated as "provider found nothing".

## Proposed Fix
Make the scan payload distinguish covered sources from skipped providers:
- Agent: report which providers actually ran (e.g. a `coveredSources` field, or send per-source non-full updates for the providers that ran instead of `full:true` when any provider was skipped). `PatchManager.Scan()` knows which providers were registered and which returned skip.
- API: when `full:true` arrives with `coveredSources`, scope `markPendingDevicePatchesMissing` to those sources (the per-source machinery already exists — `patches.ts:56-78,313-320`) instead of sweeping all.

Note the winget-as-SYSTEM change reduces how often winget is skipped, but this fix is still needed for correctness (choco not installed, provider timeouts, macOS/Linux provider mixes).

## Affected Files
- `agent/internal/heartbeat/heartbeat.go` (scan aggregation + payload)
- `agent/internal/patching/manager.go` (surface per-provider skip/ran state)
- `apps/api/src/routes/agents/patches.ts` (ingest mark-missing scoping)
[...]

PR #3512, opened 2026-08-14
fix(viewer): name libfuse2t64 for the Linux AppImage on Debian 13 / Ubuntu 24.04+ (#3415)

ToddHebebrand:
## Problem

The Linux viewer ships as an AppImage, whose runtime mounts itself with **FUSE 2**. Debian's 64-bit `time_t` (`t64`) transition renamed the providing package, so on Debian 13 (trixie) and Ubuntu 24.04+ the widely-quoted `apt install libfuse2` answers:

```
Note, selecting 'libfuse2t64' instead of 'libfuse2'
```

— which reads like a failure. Nothing in Breeze told the user what the real prerequisite was. The AppImage exits with no window and no actionable message, and the console's "Viewer didn't open?" card reappears on every attempt.

Before this change, `libfuse`, `FUSE`, `fuse2`, and `appimage-extract-and-run` appeared **nowhere in the repo** — no preflight, no docs, no UI copy. And there is no Linux viewer install script to carry a check: the AppImage is downloaded by the browser and run by hand, with registration done by the app itself in Rust. So the fix is guidance, placed in the two spots a stuck user actually looks.

## Changes

**Docs** — new **Linux viewer (AppImage)** section in `features/remote-access.mdx`:

- first-run / `breeze://` registration steps, plus `xdg-mime query default x-scheme-handler/breeze` to verify, and the "keep it at a stable path" caveat (registration records an absolute path)
- per-distribution FUSE 2 package table: `libfuse2t64` (Debian 13+/Ubuntu 24.04+), `libfuse2` (older), `fuse-libs` (Fedora/RHEL), `fuse2` (Arch/openSUSE) — and an explicit note that the apt "selecting libfuse2t64 instead" message is informational
- `--appimage-extract-and-run` and `APPIMAGE_EXTRACT_AND_RUN=1` fallbacks for hosts where FUSE can't be installed, including the caveat that the desktop entry's `Exec=` line does **not** carry the flag, so `breeze://` launches still take the FUSE path
- a separate entry for the adjacent "process runs but no window appears" host-GTK/gvfs case (#2614)
- a Troubleshooting entry that points at the section

**Web** — the Linux-only hint on the Connect Desktop fallback card now names the package and the fallback flag alongside the existing `chmod +x` line. New `connectDesktopButton.fallback.linuxFuse` key added in all seven locales.

## Scope note

[...]

PR #3514, opened 2026-08-14
fix(agent): survive gopsutil darwin netstat panic in metrics collection (#3459)

### LINKED ISSUE #3459: [macOS Agent] gopsutil v3.24.5 parseNetstatLine panic (index out of range) kills the metrics collection goroutine
Observed on a macOS 26.6.1 dev machine while investigating #3380 (unrelated to that issue's root cause).

```
panic: runtime error: index out of range [2] with length 1
  github.com/shirou/gopsutil/v3@v3.24.5/net.parseNetstatLine (net_darwin.go:37)
  ...
  agent/internal/collectors/metrics.go:159 (MetricsCollector.Collect)
```

`netstat -ib` output on newer macOS can emit lines that `parseNetstatLine` doesn't expect (fewer fields than assumed), and gopsutil indexes past the end. The panic propagates out of `MetricsCollector.Collect` and kills the goroutine that called it, so metrics/heartbeat collection stops until the agent restarts.

Two fixes worth doing together:
1. Upgrade gopsutil (v4 has darwin netstat parsing fixes) or pin a patched version.
2. Add a `recover()` guard around collector `Collect()` calls so a panic in one collector degrades that metric instead of killing the collection loop — a third-party parsing bug shouldn't take down heartbeat-adjacent machinery.

ToddHebebrand:
Closes #3459

## Root cause

gopsutil's darwin `net.IOCounters` parses `netstat -ibdnW` output line by line. `parseNetstatLine` (`net/net_darwin.go`) reads `columns[0]` and `columns[2]` **before** any length check:

```go
columns = strings.Fields(line)
if columns[0] == "Name" { ... }                                   // line 31
if subMatch := netstatLinkRegexp.FindStringSubmatch(columns[2]);  // line 37  <-- panics
...
numberColumns := len(columns)                                     // line 46  first length check
```

Line 37 is exactly the frame in the reported stack. A `netstat` line with fewer fields than expected panics with `index out of range [2] with length 1`; the panic propagated out of `MetricsCollector.Collect` and killed the goroutine driving metrics/heartbeat collection until the agent restarted.

## Note on the suggested gopsutil upgrade

The issue proposed upgrading to v4 for "darwin netstat parsing fixes". **I verified that v4 does not fix this** — `v4.26.7`'s `parseNetstatLine` has the identical unguarded `columns[0]` / `columns[2]` accesses, with the first `len(columns)` check still at line 46, after both. There is no released gopsutil version to move to, so a v3 → v4 migration across 33 agent files would have been churn that fixed nothing and re-risked every other gopsutil call site on customer machines. Left on v3.24.5 and guarded at our own call sites instead.

Worth reporting upstream separately; this PR is the agent-side containment.

## The fix

**1. Targeted: `safeNetIOCounters` (`internal/collectors/netio_safe.go`)**

Wraps `net.IOCounters` and converts a panic into an error, so a bad netstat line degrades only the *network* portion of the sample — CPU, RAM and disk still get reported. This matches gopsutil's own behaviour for a malformed line (`parseNetstatOutput` already fails the whole call on any unparsable line), so no data that would otherwise have been collected is lost; it only stops the crash escaping. The panic is logged via `slog`, with the (large) stack logged once per process rather than on every collection cycle — on an affected host the panic recurs every heartbeat.

**2. Structural: `collectors.Guard[T]` (`internal/collectors/safe.go`)**

Generic wrapper that converts a panic in any collector into a `*PanicError` instead of unwinding the caller's goroutine:

```go
metrics, err := collectors.Guard("metrics", h.metricsCol.Collect)
```

Applied to all 12 collector entry points in the heartbeat paths (metrics, software, patches, installed patches, connections, event logs, sessions, reliability, boot performance, VPN). Every one of those callers already handles a collection error by logging and marking the subsystem degraded, so the panic now flows into that existing path — no new error handling, no new silent failures. `PanicError` carries the op, the panic value and the stack, and is distinguishable via `errors.As` so a crash is never mistaken for an ordinary collection failure.

On panic the guarded result is the zero value (nil for the pointer/slice types collectors return), which is the same contract callers already have for an error return.

## Verification

- `go build ./...` — clean
- `go vet ./internal/collectors/... ./internal/heartbeat/...` — clean (only a pre-existing cgo warning from `go-m1cpu`)
- `go test -race ./...` — green
- New tests (`safe_test.go`, `netio_safe_test.go`) cover: panic → error conversion for runtime/string/error panics, pass-through of success and of ordinary errors, `errors.As` discrimination, stack capture, that a panic does **not** propagate to the calling goroutine, and that `safeNetIOCounters` / `MetricsCollector.Collect` never panic. One case reproduces the exact `index out of range [2] with length 1` shape from the report.

Not reproducible on demand — this machine's current `netstat -ibdnW` output is all 11/12-field lines — so the fix is verified by construction and by the panic-shape test rather than by triggering the real netstat condition.

## Review round (second commit)

`/pr-review-toolkit:review-pr` (code-reviewer + silent-failure-hunter) raised five consequential findings, all addressed in `cb052c4ac`:

1. **Guarded panics no longer reached Sentry** (both reviewers, independently). Converting a panic to an error stopped it reaching the enclosing `observability.Recoverer` — so this fix would have made *this class of bug* invisible to fleet telemetry, which is how #3459 was noticed in the first place. `recoverAs` now reports via a new `observability.ReportPanic` (shared with `Recoverer`, so both paths report identically), throttled to one event per `(op, panic value)` per hour so a per-cycle panic can't flood from one device. The captured stack is no longer discarded.

2. **Recovery fabricated a bandwidth spike.** `c.lastTime` advances every cycle, but the network byte baselines only updated on success — so the first good cycle after a failed sample divided a multi-interval delta by a single interval. On an affected macOS host that would have been the steady state, and it feeds bandwidth alerting. Both aggregate and per-interface baselines are now dropped on a failed sample, so recovery skips one sample instead of inventing traffic. Mutation-tested: reverting the fix fails the new test.

3. **Patch inventory could tombstone every pending patch.** A panic in the pending collector left `pendingItems` empty while `installedItems` stayed populated, falling through to a `full=true` upload with no `coveredSources` — the #2217 failure mode. A failed collection producing no pending items no longer performs a full sweep.

4. **The genuinely process-fatal paths were still unguarded.** `sendHardwareInventory`, `sendPatchInventory`, `sendReliabilityMetrics` and `runProcessSampler` are bare goroutines with no recoverer, where a panic kills the *whole process* — strictly worse than #3459's single dead loop. Added `observability.Recoverer` to each, plus `Guard` on `CollectHardware` (shells out and parses output, same risk profile as `parseNetstatLine`).

5. **Two tests were vacuous.** `net.IOCounters` is now injectable, so the panic→error conversion is tested directly rather than asserted against whatever the host's `netstat` happens to emit — the previous version's error branch was unreachable.

Not changed: the server still can't distinguish "network unknown" from "idle". The network fields are `omitempty`, so a skipped sample omits the keys rather than sending a false `0`, which is the honest encoding available without a schema change. Flagged for follow-up rather than widened here.

## Verification (final)

- `go build ./...`, `go vet`, `gofmt` — clean
- `go test -race ./internal/collectors/... ./internal/heartbeat/... ./internal/observability/...` — all green
- `go test -race ./...` — green apart from two pre-existing environmental failures in untouched packages (`internal/patching` homebrew `brew outdated` 120s timeout, `internal/workspaceindex` fsnotify debounce flake); neither package is in this diff.

### DIFF
--- a/agent/internal/collectors/metrics.go
+++ b/agent/internal/collectors/metrics.go
@@ -135,8 +135,14 @@ func (c *MetricsCollector) Collect() (*SystemMetrics, error) {
 
 	// Network — aggregate totals (backward-compatible)
 	netIO, err := safeNetIOCounters(false)
-	if err != nil {
-		logNetIOPanic(false, err)
+	if err != nil || len(netIO) == 0 {
+		// Drop the baseline. c.lastTime advances every cycle regardless, so
+		// carrying a stale counter across a missed sample would make the next
+		// successful cycle divide a multi-interval byte delta by a single
+		// interval — reporting a bandwidth spike that never happened. A zero
+		// baseline is re-seeded below and simply skips one sample instead.
+		c.lastNetIn = 0
+		c.lastNetOut = 0
 	}
 	if err == nil && len(netIO) > 0 {
 		currentIn := netIO[0].BytesRecv
@@ -160,7 +166,9 @@ func (c *MetricsCollector) Collect() (*SystemMetrics, error) {
 	// Network — per-interface bandwidth
 	perIface, err := safeNetIOCounters(true)
 	if err != nil {
-		logNetIOPanic(true, err)
+		// Same reasoning as the aggregate baseline above: a stale per-interface
+		// snapshot across a missed sample would fabricate a rate spike.
+		clear(c.lastIface)
 	}
 	if err == nil {
 		hasHistory := !c.lastTime.IsZero() && elapsed > 1 && elapsed < 300
--- a/agent/internal/collectors/metrics.go
+++ b/agent/internal/collectors/metrics.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/shirou/gopsutil/v3/cpu"
 	"github.com/shirou/gopsutil/v3/disk"
-	"github.com/shirou/gopsutil/v3/net"
 	"github.com/shirou/gopsutil/v3/process"
 )
 
@@ -135,7 +134,10 @@ func (c *MetricsCollector) Collect() (*SystemMetrics, error) {
 	c.collectDiskActivity(metrics, elapsed)
 
 	// Network — aggregate totals (backward-compatible)
-	netIO, err := net.IOCounters(false)
+	netIO, err := safeNetIOCounters(false)
+	if err != nil {
+		logNetIOPanic(false, err)
+	}
 	if err == nil && len(netIO) > 0 {
 		currentIn := netIO[0].BytesRecv
 		currentOut := netIO[0].BytesSent
@@ -156,7 +158,10 @@ func (c *MetricsCollector) Collect() (*SystemMetrics, error) {
 	}
 
 	// Network — per-interface bandwidth
-	perIface, err := net.IOCounters(true)
+	perIface, err := safeNetIOCounters(true)
+	if err != nil {
+		logNetIOPanic(true, err)
+	}
 	if err == nil {
 		hasHistory := !c.lastTime.IsZero() && elapsed > 1 && elapsed < 300
 		seen := make(map[string]bool)
--- a/agent/internal/collectors/netio_safe.go
+++ b/agent/internal/collectors/netio_safe.go
@@ -1,0 +1,50 @@
+package collectors
+
+import (
+	"errors"
+	"log/slog"
+	"sync"
+
+	"github.com/shirou/gopsutil/v3/net"
+)
+
+// netIOStackLogged ensures the (large) panic stack is logged once per process
+// rather than on every collection cycle — on an affected host the panic recurs
+// every heartbeat, and the stack is only diagnostic value the first time.
+var netIOStackLogged sync.Once
+
+// safeNetIOCounters wraps gopsutil's net.IOCounters, converting a panic into an
+// error so the rest of the metrics sample (CPU, RAM, disk) still gets reported.
+//
+// Why this is needed: gopsutil's darwin implementation reads columns[0] and
+// columns[2] of each `netstat -ibdnW` line with no length check
+// (net/net_darwin.go parseNetstatLine), so a line with fewer fields than
+// expected panics with "index out of range". On newer macOS that has taken down
+// the metrics goroutine mid-heartbeat (issue #3459).
+//
+// Upgrading does NOT fix this: the same unguarded indexing is still present in
+// gopsutil v4.26.7, so there is no released version to move to. Guarding at our
+// own call site is the fix until upstream adds the bounds check.
+//
+// Behaviour on panic matches gopsutil's own error path for a malformed line
+// (parseNetstatOutput already fails the whole call on any unparsable line), so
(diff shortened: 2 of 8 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1 (code-reviewer + silent-failure-hunter) on `7dfbc8bad`, round 2 (code-reviewer, focused) on `cb052c4ac`.

**Findings:** 7 raised → all 7 addressed; 0 outstanding.

Round 1 (5 findings, fixed in `cb052c4ac`):
- Guarded panics no longer reached Sentry — converting a panic to an error stopped it hitting the enclosing `observability.Recoverer`, which would have made *this* bug class invisible to fleet telemetry. Now reported via a new shared `observability.ReportPanic`, throttled per op+value.
- Recovery fabricated a bandwidth spike (`lastTime` advanced while the byte baseline didn't) — baselines are now dropped on a failed sample.
- A panic in the pending-patch collector could reach a `full=true` upload with empty pending, tombstoning every pending patch (#2217 mode).
- The genuinely process-fatal paths were still unguarded — four bare goroutines (`sendHardwareInventory`, `sendPatchInventory`, `sendReliabilityMetrics`, `runProcessSampler`) where a panic kills the whole process.
- Two tests were vacuous; `net.IOCounters` is now injectable so the panic→error path is tested directly.

Round 2 (2 findings, fixed in `4cd49e4ae`):
- The patch-sweep suppression was over-broad — the tombstone hazard only exists when `coveredSources` is nil, so the fix was stranding legitimate narrowed sweeps. Re-gated on the actual hazard.
- The report throttle keyed on the raw panic value, but runtime panics embed their operands, so it both leaked and defeated itself. Keys now normalise digit runs and the map is bounded.

**Tests:** `go build ./...`, `go vet`, `gofmt` clean. `go test -race ./...` green locally (the two failures seen on an earlier run — `internal/patching` homebrew timeout, `internal/workspaceindex` fsnotify debounce — did not reproduce and are in packages this PR doesn't touch). CI: **Test Agent, Test Agent (race), Test Agent (Windows), Lint Agent (Go), Type Check all pass.** The baseline-drop regression test was mutation-tested — reverting the fix fails it.

**Note on the two red security checks:** `Go Vulnerability Check` and `Trivy Filesystem Scan` fail identically on unrelated PRs (e.g. #3512) and this PR changes no dependencies at all (`go.mod`/`go.sum` untouched). Newly-published advisories against existing deps, not this change.

**Worth knowing before merge:** the issue asked for a gopsutil upgrade, but v4.26.7 has the *same* unguarded indexing, so there is no version to upgrade to — the panic is guarded at our call sites instead, and the underlying gopsutil bug is still worth reporting upstream.

**Status:** review-clean, awaiting maintainer merge. Not merging or closing — issue #3459 stays open until the reporter confirms on the affected macOS host.

ToddHebebrand:

|---|---|---|
| both | pass | pass |

No action needed on this PR.

PR #2979, opened 2026-07-31
fix(api): validate every declared env var, not a hand-maintained subset (#2896)

ToddHebebrand:
Refs #2896

## Problem

`validateConfig()` built the object it handed to `envSchema.safeParse()` from a
second, **hand-maintained** list of `KEY: env.KEY` lines. A key declared in the
schema but missing from that list is `undefined` at parse time, so its rules
never run: the schema default silently wins, an operator typo boots instead of
failing, and `getConfig()` reports a value the runtime disagrees with.

`IP_ALLOWLIST_ENFORCEMENT_MODE` was one such key (the filed bug). Its
`z.enum(['enforce','off'])` never saw an operator value, while
`services/ipAllowlist.ts` read `process.env` directly — so a valid `off` was
honoured by the request path while `getConfig()` still reported `'enforce'`, and
a typo (`Off`, `disabled`, `false`) was silently accepted and enforced.

## Schema ↔ pick-list diff (the issue's "broader check")

Diffing the two key sets surfaced **7** declared-but-never-validated variables:

| Variable | Declared as | Impact of the omission |
|---|---|---|
| `IP_ALLOWLIST_ENFORCEMENT_MODE` | `z.enum(['enforce','off']).default('enforce')` | **Live rule dead.** Enum never ran; config/runtime divergence. |
| `AGENT_AUTO_PROMOTE` | `z.string().optional()` + `superRefine` boolean guard | **Live rule dead.** The `superRefine` typo-guard (`AGENT_AUTO_PROMOTE=falze` → parses truthy → still auto-promotes the fleet) has never fired, because `data.AGENT_AUTO_PROMOTE` was always `undefined`. |
| `SSO_DOMAIN_VERIFICATION_STRICT` | `z.string().optional()` | `getConfig()` value always `undefined`. |
| `QBO_CLIENT_ID` / `QBO_CLIENT_SECRET` / `QBO_REDIRECT_URI` / `QBO_ENVIRONMENT` | `z.string().optional()` | `getConfig()` value always `undefined`. |

All 7 are fixed here — none were deliberate exclusions. The 6 non-enum ones are
behaviourally inert at parse time (optional strings cannot fail) and none of
them is read through `getConfig()` today (all read `process.env` directly, via
`config/env.ts` for the QBO group), so including them only makes the config
object accurate. Nothing was left out; there is no residual list.

## Fix

Rather than adding 7 lines to a list that has now drifted 7 times, the parse
input is **derived from the schema**, which makes this bug class structurally
[...]

Issue #2896, opened 2026-07-28
[API] IP_ALLOWLIST_ENFORCEMENT_MODE is declared in envSchema but never passed into validateConfig, so its validation and default never run


ToddHebebrand:
## Description

Deferred from the security remediation Wave 5 review. `IP_ALLOWLIST_ENFORCEMENT_MODE` is declared in the config schema but is not one of the keys `validateConfig()` hands to `safeParse`, so the enum validation never sees the operator's value. A typo is accepted silently at boot instead of refusing to start, and `getConfig()` reports a value that can disagree with the runtime behaviour.

## Root Cause

`apps/api/src/config/validate.ts:636` declares:

```ts
IP_ALLOWLIST_ENFORCEMENT_MODE: z.enum(['enforce', 'off']).default('enforce'),
```

but `validateConfig()` (line ~1498) builds the object it validates from a hand-maintained pick list of `env.*` keys, and `IP_ALLOWLIST_ENFORCEMENT_MODE` is not in it. The field is therefore always `undefined` at parse time and always resolves to the `'enforce'` default in `AppConfig`.

Meanwhile the runtime read bypasses the config object entirely — `apps/api/src/services/ipAllowlist.ts:43`:

```ts
export function ipAllowlistMode(): IpAllowlistMode {
  return process.env.IP_ALLOWLIST_ENFORCEMENT_MODE === 'off' ? 'off' : 'enforce';
}
```

Consequences:

1. **Typos are silently ignored.** `IP_ALLOWLIST_ENFORCEMENT_MODE=Off` / `disabled` / `false` should fail the enum and refuse to boot. Instead the API starts and enforces the allowlist, which can lock admins out of a deployment whose operator believed they had disabled enforcement. Fail-closed, so not a security hole — but a silent one, and exactly the misconfiguration class the boot validator exists to catch.
2. **Config/runtime divergence.** With a valid `off`, the runtime honours it while `getConfig().IP_ALLOWLIST_ENFORCEMENT_MODE` still reads `'enforce'`. Any future caller that trusts the config object gets the wrong answer.

## Proposed Fix

Add `IP_ALLOWLIST_ENFORCEMENT_MODE: env.IP_ALLOWLIST_ENFORCEMENT_MODE` to the `safeParse` object in `validateConfig()`, then have `ipAllowlistMode()` read the validated config instead of raw `process.env` (keeping a `process.env` read only where config is not yet initialised, matching the `clientIp.ts` pattern it references).

[...]

PR #3515, opened 2026-08-14
fix(api,config): declare the three env vars validateConfig read undeclared (#3374)

### LINKED ISSUE #3374: [API] Converse of #2896: three env vars read by validateConfig/collectWarnings are never declared in envSchema
## Description
#2896 / PR #2979 closed the declared-but-never-validated drift class by deriving the parse input from the schema shape. The converse direction is still open, and the #2979 author documented it in-code without fixing it: `validateConfig()` and `collectWarnings()` read `M365_GRAPH_ACTIONS_TOOLS_ENABLED`, `APP_ENCRYPTION_KEY_ID`, and `TRUST_CF_CONNECTING_IP` straight off `env` without those keys being declared in `envSchema` — so they get no validation, no defaults, and don't participate in the schema-derived contract test.

Blast radius is smaller than #2896 (undeclared vars carry no dead rules), but the class remains: any future key read directly off `env` instead of via the schema is invisible to the guard added in #2979.

## Proposed Fix
Declare the three keys in `envSchema` (with appropriate types/defaults) and route their reads through `getConfig()`. Optionally extend the contract test to grep `validateConfig`/`collectWarnings` for direct `env.` reads of undeclared keys so the converse drift also fails CI.

## Affected Files
- `apps/api/src/config/validate.ts`

## Reported By
Verification sweep of #2896 (residual documented in-code by the #2979 author).

ToddHebebrand:
Closes #3374

## Problem

PR #2979 closed the **declared-but-never-parsed** drift by deriving `validateConfig()`'s parse input from the schema shape. The **converse** stayed open, and the #2979 author documented it in-code rather than fixing it: `validateConfig()` and `collectWarnings()` read `M365_GRAPH_ACTIONS_TOOLS_ENABLED`, `APP_ENCRYPTION_KEY_ID` and `TRUST_CF_CONNECTING_IP` straight off `env` without declaring them in `envSchema`.

An undeclared key gets no type, no default and no validation, and it sits outside the schema-derived contract — so the value the validator acts on and the value `getConfig()` reports can disagree, and nothing catches the next one.

## Changes

- **Declare all three keys** in `envObjectSchema`, each with a comment naming its runtime reader.
- **Boolean typo-guards** for the two flags, matching the existing `AGENT_AUTO_PROMOTE` / `ABUSE_SIGNALS_ENABLED` treatment. Both runtime readers (`services/clientIp.ts`, `services/m365ControlPlane/writeActionRuntimeConfig.ts`) treat an unrecognized value as **OFF**, so a typo silently strips CF-Connecting-IP attribution from rate limits / audit logs / IP allowlists, or silently darkens the Tier-3 M365 write-action tools an operator believed they had enabled. Empty/unset stays allowed — both compose files inject these as `${VAR:-}`, so `""` is the common case.
- **Moved the `APP_ENCRYPTION_KEY_ID` ↔ `M365_GRAPH_ACTIONS_TOOLS_ENABLED` pairing rule** out of the post-parse `throw` in `validateConfig()` and into the schema `superRefine`. Same refusal, same message — but it now runs on declared data and joins the aggregated configuration-error report instead of short-circuiting it.
- **Extended the #2896 contract test** with a source-level scan: any `env.KEY` / `env['KEY']` read inside `validateConfig()`/`collectWarnings()` that is not in `ENV_SCHEMA_KEYS` fails CI. Source-level on purpose — there is no runtime observation that distinguishes a declared read from an undeclared one, and the point is to fail the *next* one at authoring time.

## Compatibility

No new required variable and no default changes. The only new refusal is a **non-empty, non-boolean** value on the two flags — previously that booted and read as `false`. Consistent with how this repo already treats boolean-flag typos.

## Verification

- `vitest run src/config` — 11 files, 426 passed. `validate.test.ts`: 276 passed, including 7 new.
- **Mutation-tested the new guard**: injecting `void env.SOME_BOGUS_UNDECLARED_KEY;` into `collectWarnings()` reddens it, so it cannot pass vacuously. The detector's own matcher and brace-matcher have unit tests.
- `tsc --noEmit -p apps/api/tsconfig.json` clean.
- `src/services/ipAllowlistMode.config.test.ts` fails identically on an untouched tree in this environment (local env leakage) — pre-existing, unrelated.

### DIFF
--- a/apps/api/src/config/validate.test.ts
+++ b/apps/api/src/config/validate.test.ts
@@ -1,4 +1,4 @@
-import { chmodSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { chmodSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer, both against `git diff origin/main...HEAD` with mutation testing rather than eyeball coverage.

**Findings:** 5 raised → all 5 addressed in `1ea73ab25`; 0 outstanding.
- **Both new boolean typo-guards had zero coverage** (test-analyzer, criticality 8) — neutralizing either condition left the suite fully green. Added accept/reject/empty-value suites in the shape of the existing `ABUSE_SIGNALS_ENABLED` block. Re-verified by mutation: the same neutralization now fails **7 tests**.
- **Untested precedence change** — the pairing rule moving into `superRefine` means it fires during `safeParse`, so `APP_ENCRYPTION_KEY_ID` is now reported even with no executor descriptor configured. Pinned with a test, along with the whitespace-only key id case (the key has no `.min(1)` of its own).
- **Stale comment** on the pre-existing pairing test claimed the descriptor validator runs *before* the key-id check — that ordering inverted in this PR. Corrected.
- **Wrong justification** in the `TRUST_CF_CONNECTING_IP` comment: only `docker-compose.yml` injects `${VAR:-}`; the droplet stack injects `${VAR:-true}`. The rule accepts both; only the stated reason was wrong.
- **Over-claim in the `ENV_SCHEMA_KEYS` docblock** — the new scan catches an inline `env.KEY` read in those two functions only. Keys reached indirectly (the whole `env` handed to the three `*RuntimeConfigAtBoot` validators, ~2 dozen deliberately lazily-parsed executor keys) are out of scope. Narrowed the claim and added a test that documents the boundary.

Both reviewers independently confirmed the two things that could have broken deploys: the `superRefine` move is behaviour-preserving (same condition, same truthiness set as `flagEnabled()`, message carried verbatim), and **no tracked value in any compose file or `.env.example` can trip the new format guards** — every one is boolean-ish or empty, and `boolValues` is a strict superset of both runtime readers' accepted sets.

**Tests:** `vitest run src/config/validate.test.ts` — 306 passed (30 new). `tsc --noEmit -p apps/api/tsconfig.json` clean. `src/services/ipAllowlistMode.config.test.ts` fails identically on an untouched tree in this environment (local env leakage) — pre-existing, unrelated.

**CI note:** `Go Vulnerability Check` and `Trivy Filesystem Scan` are red here, and red the same way on unrelated open PRs (e.g. #3513) while passing on `main`'s last run — newly published advisories, repo-wide, nothing to do with this change.

ToddHebebrand:

|---|---|---|
| both | pass | pass |

No action needed on this PR.

PR #3518, opened 2026-08-14
fix(api,web,shared): http/https allowlist for partner website URLs (#3430)

### LINKED ISSUE #3430: [API][Web] Partner/org Website field has no URL or scheme validation — accepts javascript: and data: URIs
## Description

The partner/org **Website** field accepts any string, including `javascript:` and `data:` URIs. There is no scheme validation and no URL validation of any kind:

```ts
// apps/api/src/routes/orgs.ts:482
website: z.string().optional()
```

Entered via Settings → Partner → Company. Saved and returned verbatim.

## Exploitability today: none that I could find — but that is a property of the current UI, not of the data

The QA sweep looked for a renderer and found no `<a href={...website}>` consuming this field, so there is no live stored-XSS path right now. I am reporting it as a **missing control**, not a live vulnerability, and deliberately not claiming more than that.

Two reasons it still matters:

1. **The guard belongs at the boundary, not at each renderer.** A field named `website` is going to end up in an `href` eventually — that is what it is for. Whoever adds that link will reasonably assume the stored value is a URL, because the field is called Website. Note there is already an `href={branding.websiteUrl}` at `apps/web/src/components/software/BuiltinPackageDetail.tsx:64`; I did not confirm whether that value can trace back to this field, and that trace is worth doing as part of the fix.
2. **The repo already has the pattern.** `packages/shared/src/validators/remoteAccessLauncherScheme.ts` exists precisely to allowlist schemes, with tests. This field should use the same approach rather than inventing one.

The release checklist for this cycle explicitly asked that Partner settings reject `javascript:`/`data:`, which is how this surfaced.

## Proposed fix

Validate as a URL with an `http`/`https` allowlist, in `packages/shared/src/validators/` so both the API schema and the web form can share it, following the `remoteAccessLauncherScheme` precedent. Reject rather than silently strip — a silently-cleared field looks like a save failure.

Worth grepping for sibling fields while in there (`supportUrl`, `logoUrl`, `documentationUrl`, any other `z.string()` that names a URL) — this is likely a class, not a single field.

## Affected files

- `apps/api/src/routes/orgs.ts:482`
- `apps/web/src/components/settings/PartnerCompanyTab.tsx`
- Precedent to follow: `packages/shared/src/validators/remoteAccessLauncherScheme.ts`

ToddHebebrand:
Closes #3430.

## What was actually left

The issue pointed at `apps/api/src/routes/orgs.ts:482` (`website: z.string().optional()`). That specific field was **already fixed** by #3442 / #3448, which landed an inline `httpUrlValue` helper in `orgs.ts` — those PRs never referenced #3430, so the issue stayed open. I verified against `origin/main` before starting.

Two parts of the issue were still genuinely open, and one of them is the field that matters more.

### 1. `billingWebsite` — the partner Website that actually reaches renderers

`packages/shared/src/validators/invoices.ts:104` was still `z.string().max(255).nullable().optional()`. This is the **other** partner Website field (Billing → seller profile), and it is the one that is snapshotted onto every issued document and rendered downstream:

- `apps/api/src/services/sellerSnapshot.ts:48` / `publicQuoteDto.ts:34` → `seller.website`
- `apps/web/.../InvoiceDocument.tsx:106`, `QuoteDocument.tsx:366`, `InvoiceDetail.tsx:507`, `QuoteDetail.tsx:313`
- `apps/portal/src/components/portal/documentShell.tsx:71`
- PDFs: `quotePdf.ts:814`, `invoicePdf.ts:165,259`

All of those are **text, not `href`** today, so — as with the original report — this is a missing control, not a live XSS. It is the field most likely to become a link, since it is the one printed on customer-facing documents.

### 2. The web form had no validation at all

`PartnerCompanyTab.tsx` is presentational: `<input type="url">` with no `<form>` wrapper, so native validation never fires, and `PartnerSettingsPage.handleSave` did no check. `javascript:alert(1)` round-tripped to the API and came back as a raw 400.

## The fix

New shared validator `packages/shared/src/validators/httpUrl.ts`, following the `remoteAccessLauncherScheme.ts` precedent the issue asked for:

- `isHttpUrl(v)` — `new URL()` parse + `http:`/`https:` protocol check. Using the URL parser rather than a regex is deliberate: it normalises `HTTP://` and the control-char obfuscations (`java\nscript:`) that a `startsWith` check misses.
- `httpUrlValue(label, max)` / `httpUrlField(label, max)` / `nullableHttpUrlField(label, max)` zod helpers, plus `httpUrlErrorMessage(label)` shared with the web form.
- **Rejects rather than strips**, per the issue — a silently-cleared field reads as a failed save.

Wired into:
- `partnerBillingSettingsSchema.billingWebsite` → `nullableHttpUrlField('Billing website', 255)` (the actual gap).
- `orgs.ts` — deletes its local copy of the helper and imports the shared one, so `contact.website`, `slackWebhookUrl`, `webhooks[]` and `elasticsearchUrl` keep identical behaviour from one source.
- `PartnerCompanyTab` — inline `aria-invalid` + `aria-describedby` error; `PartnerSettingsPage.handleSave` blocks the round-trip, matching the existing maintenance-window precedent 8 lines above it.

### Behaviour change worth a release note

A **scheme-less hostname** (`acme.com`) is no longer accepted when *saving* billing settings — it must be `https://acme.com`. This matches the already-shipped rule on the twin `contact.website` field rather than guessing a scheme. Existing stored values still **read** back fine; only a re-save of the billing profile requires correcting the field. One existing test fixture used `acme.test` and was updated, with an explicit test pinning the new rejection.

## Traced, and *not* a sink

`branding.websiteUrl` at `BuiltinPackageDetail.tsx:64` — the issue flagged this as worth tracing. It **cannot** reach tenant data: `getProviderBranding()` returns a hardcoded record (`providerBranding.ts:18-36`) with literal `https://www.huntress.com` / `https://www.sentinelone.com`. Not attacker-controllable. This is the only `href` on any `websiteUrl` in the repo.

## Sweep — noted, not fixed here

The issue asked whether this is a class. It is, but the rest is a different blast radius and doesn't belong in this PR:

| Location | Field | Gap |
|---|---|---|
| `apps/api/src/routes/orgs.ts:617` | partner `branding.logoUrl` | `z.string().max(400_000)` with **no scheme check**, while the near-identical `partnerLoginBranding.ts:23` restricts to `https://` or `data:image/(png\|jpeg\|webp);base64`. The inconsistency is the finding. |
| `packages/shared/src/validators/index.ts:331` | org `elasticsearchUrl` | `z.string().trim().optional()` — the org-level log-forwarding path is looser than the partner-level one now sharing `httpUrlField`. Server-dialled, so SSRF-shaped. |
| `apps/api/src/routes/orgs.ts:218` | `siteContactSchema` | `.passthrough()`, so a `website` on a **site** contact is accepted unvalidated. |
| `apps/web/.../NotificationChannelForm.tsx:51,53,56` | webhook URLs | bare `z.string().optional()` in one variant while lines 20/25 of the same file use `.url()`. |
| `packages/shared/src/validators/index.ts:796` | `siteUrl` (OneDrive) | bare, but SharePoint-sourced rather than partner branding. |

`supportUrl` / `documentationUrl` / `iconUrl` / `faviconUrl` don't exist in partner/org schemas. Portal branding logo/customDomain aren't admin-writable (`portal.ts` is `.strict()`).

## Testing

- `packages/shared/src/validators/httpUrl.test.ts` — new, 55 tests. Corpus covers `javascript:`, `data:`, `vbscript:`, `file:`, `about:`, `blob:`, `view-source:`, `chrome:`, `jar:`, `filesystem:`, plus whitespace/control-char obfuscation and case variants; explicitly asserts reject-not-strip and the clearability of empty string.
- `invoices.test.ts` — `billingWebsite` accept/reject matrix, null/undefined clearability, the 255 cap, and the scheme-less-hostname behaviour change.
- `invoices/settings.test.ts` — route-level 400 with **no service call** for three dangerous schemes, plus a 200 path asserting the value reaches the service. Proves the guard at the HTTP boundary, not just in the schema.
- `PartnerCompanyTab.test.tsx` — 9 new tests: error shown/hidden, `aria-invalid`/`aria-describedby` wiring, and that the typed value is still reported to the parent (not silently dropped).

Green: shared 153 · `apps/api/src/routes/orgs.test.ts` 213 · `invoices/settings.test.ts` 10 · `apps/web/src/components/settings/` 520 (60 files). `tsc --noEmit` clean for shared, api and web; `astro check` 0 errors.

---

## Review round (addressed in `616915d41`)

Three reviewers (code-reviewer, silent-failure-hunter, pr-test-analyzer). Two Important findings, both raised independently by two reviewers, both fixed:

**1. Legacy scheme-less `billingWebsite` blocked the entire billing form.** `PartnerBillingSettings.tsx` PATCHes the *full* payload and repopulates the stored value, so a partner whose `billing_website` holds a scheme-less hostname — legal until this PR — could not save a tax-rate or prefix change without a 400 on a field they never touched. The form now carries the same inline `aria-invalid` / `aria-describedby` error as the Company tab, and Save is disabled until it's corrected. 5 new tests, including the legacy-value load case.

**2. The rationale comment named the wrong column.** It claimed `contact.website` reaches branded PDFs/invoices/email footers. It does not — those read `partners.billing_website` via `buildSellerSnapshot`; `contact.website` has no consumer outside the settings form. Since that comment is what a future reader will use to decide whether the guard can be relaxed, it's corrected in both `orgs.ts` and the shared validator. The SSRF claim is narrowed in the same pass: this is a **scheme** guard, not an egress control, and `http://localhost` still passes deliberately (self-hosted Elasticsearch). Reviewers confirmed the three outbound-dialled fields are separately SSRF-guarded by `safeFetch`.

Also fixed:

- **Coverage hole in the code this PR restructured.** `orgs.test.ts` stayed **213/213 green with the `slackWebhookUrl`, `webhooks[]` and `elasticsearchUrl` guards deleted**. Now covered, and mutation-checked — I removed all three guards and confirmed exactly the three new cases fail, then restored.
- **Trim asymmetry I introduced.** `PartnerSettingsPage` validated `contactWebsite.trim()` but submitted it untrimmed, so a whitespace-only entry passed the client check and failed server-side with no inline error. Trimmed on submit; tests for both the block-the-save and trimmed-happy paths.
- **`sellerSnapshot.ts` now records why the seller website must not be linkified** without an output-side filter. Reviewers confirmed every current render site is plain text (no `href`, no PDFKit `link:` option, `escapeHtml` on the HTML path), so legacy values are inert — but `invoices.seller_snapshot` / `quotes.seller_snapshot` are immutable jsonb never re-validated on read, so a later scrub of the source column would not reach already-issued documents. The comment is the durable guard against someone linkifying it later.

Reviewers found **no bypass** in the allowlist — `new URL()` + `protocol` was probed against tab/newline/NUL obfuscation, backslash normalization, scheme-relative, mixed case and unicode hosts — and confirmed the change is **write-path only**, with `orgs.ts` semantics preserved exactly by the extraction.

### Open question for the maintainer

Nothing backfills existing `partners.billing_website` rows. The inline guard makes a legacy value discoverable and fixable in the UI, which I think is the right level of response given every render site is inert text — but if you'd rather scrub or normalize the column, that's a data decision I've deliberately left to you.

## Verification after the review round

- `apps/api`: `orgs.test.ts` + `invoices/settings.test.ts` → 229 passed
- `apps/web`: `components/settings/` + `components/billing/` → 1128 passed (138 files)
- `packages/shared`: full validator suite → 1315 passed (43 files)
- `tsc --noEmit` clean for api, web and shared; `astro check` 0 errors

### DIFF
--- a/apps/api/src/routes/invoices/settings.test.ts
+++ b/apps/api/src/routes/invoices/settings.test.ts
@@ -69,6 +69,32 @@ describe('billing settings routes', () => {
     expect(svc.updatePartnerBillingSettings).not.toHaveBeenCalled();
   });
 
+  // #3430 — billingWebsite is snapshotted onto issued invoices/quotes and
+  // rendered in branded PDFs and the customer portal, so a dangerous scheme
+  // must be rejected at the boundary rather than persisted on a 200.
+  it.each(['javascript:alert(1)', 'data:text/html,<script>alert(1)</script>', 'file:///etc/passwd'])(
+    'PATCH /partner/billing-settings rejects billingWebsite %j (→ 400, no service call)',
+    async (billingWebsite) => {
+      const res = await invoiceSettingsRoutes.request('/partner/billing-settings', jsonBody({
+        currencyCode: 'USD', invoiceNumberPrefix: 'INV', invoiceTermsDays: 30, billingWebsite
+      }));
+      expect(res.status).toBe(400);
+      expect(svc.updatePartnerBillingSettings).not.toHaveBeenCalled();
+    }
+  );
+
+  it('PATCH /partner/billing-settings accepts an https billingWebsite', async () => {
+    const res = await invoiceSettingsRoutes.request('/partner/billing-settings', jsonBody({
+      currencyCode: 'USD', invoiceNumberPrefix: 'INV', invoiceTermsDays: 30,
+      billingWebsite: 'https://acme.example.com'
+    }));
+    expect(res.status).toBe(200);
+    expect(svc.updatePartnerBillingSettings).toHaveBeenCalledWith(
+      expect.objectContaining({ billingWebsite: 'https://acme.example.com' }),
+      expect.objectContaining({ partnerId: 'p1' })
+    );
+  });
+
   it('PATCH /orgs/:orgId/billing-settings updates org config', async () => {
     (svc.updateOrgBillingSettings as any).mockResolvedValue({ id: ORG_ID, taxExempt: true, taxRate: null });
     const res = await invoiceSettingsRoutes.request(`/orgs/${ORG_ID}/billing-settings`, jsonBody({
--- a/apps/api/src/routes/orgs.test.ts
+++ b/apps/api/src/routes/orgs.test.ts
@@ -1098,6 +1098,79 @@ describe('org routes', () => {
         }
       );
     });
+
+    // The three sibling URL fields share the same helper as contact.website but
+    // had NO coverage of their own — the suite stayed 213/213 green with all
+    // three guards deleted (#3430 review). They are defense-in-depth (egress is
+    // separately guarded by safeFetch), which is exactly why a silent
+    // regression here would go unnoticed.
+    describe('sibling URL fields share the scheme allowlist', () => {
+      function seedPartner() {
+        vi.mocked(db.select).mockReturnValue({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({
+              orderBy: vi.fn().mockReturnValue({ limit: vi.fn().mockResolvedValue([]) }),
+              limit: vi.fn().mockResolvedValue([{ id: 'partner-123', name: 'P', settings: {} }])
+            })
+          })
+        } as any);
+      }
+
+      async function patchSettings(settings: Record<string, unknown>) {
+        return app.request('/orgs/partners/me', {
+          method: 'PATCH',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ settings })
+        });
+      }
+
+      const BAD = 'javascript:alert(1)';
+      const cases: Array<[string, Record<string, unknown>, Record<string, unknown>]> = [
+        [
+          'notifications.slackWebhookUrl',
+          { notifications: { slackWebhookUrl: BAD } },
+          { notifications: { slackWebhookUrl: 'https://hooks.slack.example/x' } },
+        ],
+        [
+          'notifications.webhooks[]',
+          { notifications: { webhooks: [BAD] } },
+          { notifications: { webhooks: ['https://hooks.example.com/x'] } },
+        ],
+        [
+          'eventLogs.elasticsearchUrl',
+          { eventLogs: { elasticsearchUrl: BAD } },
+          { eventLogs: { elasticsearchUrl: 'https://es.example.com:9200' } },
+        ],
+      ];
+
+      it.each(cases)('rejects a dangerous scheme on %s with 400 and never writes it', async (_label, bad) => {
+        setAuthContext({ scope: 'partner', partnerId: 'partner-123' });
+        seedPartner();
+        const setSpy = vi.fn();
+        vi.mocked(db.update).mockReturnValue({ set: setSpy } as any);
+
+        const res = await patchSettings(bad);
+
+        expect(res.status).toBe(400);
+        expect(setSpy).not.toHaveBeenCalled();
+      });
+
+      it.each(cases)('accepts an https value on %s', async (_label, _bad, good) => {
+        setAuthContext({ scope: 'partner', partnerId: 'partner-123' });
+        seedPartner();
+        vi.mocked(db.update).mockReturnValue({
+          set: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({
+              returning: vi.fn().mockResolvedValue([{ id: 'partner-123', name: 'P', settings: {} }])
+            })
+          })
+        } as any);
+
+        const res = await patchSettings(good);
+
+        expect(res.status).toBe(200);
+      });
+    });
   });
 
   describe('PATCH /orgs/partners/me — ticketing.inbound merge safety', () => {
--- a/apps/api/src/routes/orgs.ts
+++ b/apps/api/src/routes/orgs.ts
@@ -468,12 +468,20 @@ const supportedLocales = ['en', 'pt-BR', 'es-419', 'fr-FR', 'fr-CA', 'de-DE', 'i
  * The partner-settings URL fields below are restricted to http/https by the
  * shared `httpUrlValue`/`httpUrlField` helpers (`@breeze/shared`). They split
  * into two risk classes and both land on the same guard:
- *  - values rendered as links (contact.website → branded PDFs, invoices, email
- *    footers), where `javascript:`/`data:text/html,…` is stored XSS against the
- *    partner's own customers;
+ *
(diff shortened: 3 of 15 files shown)


ToddHebebrand:
**Review run:** 3 agents — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`.

**Findings:** 2 Important (each raised independently by two reviewers) + 3 secondary → all 5 addressed in `616915d41`; 0 outstanding.

- *Legacy `billingWebsite` blocked the whole billing form* — that form PATCHes the full payload, so a scheme-less value (legal before this PR) 400'd unrelated edits. Inline error + Save disabled until corrected, matching the Company tab.
- *Rationale comment named the wrong column* — `contact.website` does not reach branded PDFs; `partners.billing_website` does. Corrected in `orgs.ts` and the shared validator, and the SSRF claim narrowed to what this actually is: a scheme guard, not an egress control.
- *Coverage hole* — `orgs.test.ts` was **213/213 green with the `slackWebhookUrl` / `webhooks[]` / `elasticsearchUrl` guards deleted**. Now covered and mutation-checked: removing all three fails exactly the three new cases.
- *Trim asymmetry* — `PartnerSettingsPage` validated trimmed but submitted untrimmed; whitespace-only entries failed server-side with no inline error.
- *Latent linkify risk* — `sellerSnapshot.ts` now records why the seller website must stay unlinked: seller snapshots are immutable jsonb never re-validated on read, so a later scrub of the source column would not reach issued documents.

Reviewers found **no bypass** in the allowlist (probed tab/newline/NUL obfuscation, backslash normalization, scheme-relative, mixed case, unicode hosts) and confirmed the change is **write-path only** — every render site is plain text, so stored data reads are unaffected.

**Tests:** `apps/api` 229 passed (`orgs.test.ts` + `invoices/settings.test.ts`) · `apps/web` 1128 passed / 138 files (`components/settings/` + `components/billing/`) · `packages/shared` 1315 passed / 43 files. `tsc --noEmit` clean for api, web and shared; `astro check` 0 errors.

**CI note:** `Go Vulnerability Check` and `Trivy Filesystem Scan` are red, and are **not** from this PR — it contains no Go and no dependency changes, and unrelated PR #3515 fails the identical two scans on the same base.

**Open for the maintainer:** nothing backfills existing `partners.billing_website` rows. The inline guard makes a legacy value discoverable and fixable in the UI, which seems proportionate given every render site is inert text — but scrubbing or normalizing the column is a data decision I left to you.

ToddHebebrand:

|---|---|---|
| both | pass | pass |

No action needed on this PR.

Issue #3522, opened 2026-08-14
Go Vulnerability Check fails on every PR: four new Go stdlib advisories, toolchain pinned one patch below the fix


bdunncompany:
## What happens

`Go Vulnerability Check` fails on **every** open PR, and will fail `main` at its next push. No repo change caused it: govulncheck resolves against a live advisory database, so the identical Go tree passed and then failed hours later.

```
ALLOWLISTED GO-2026-5051 — reviewed exception
FAIL GO-2026-5972 — symbol-level vulnerability not in the allowlist
FAIL GO-2026-6088 — symbol-level vulnerability not in the allowlist
FAIL GO-2026-6090 — symbol-level vulnerability not in the allowlist
FAIL GO-2026-6218 — symbol-level vulnerability not in the allowlist
```

All four are **Go standard library** advisories, not module dependencies:

| ID | Alias | Package | Fixed in |
|---|---|---|---|
| GO-2026-5972 | CVE-2026-33818 | `encoding/asn1` | 1.25.13 / 1.26.6 / 1.27.0-rc.3 |
| GO-2026-6088 | CVE-2026-56859 | `encoding/xml` | same |
| GO-2026-6090 | CVE-2026-56862 | `crypto/tls` | same |
| GO-2026-6218 | CVE-2026-56860 | `net/url` | same |

Confirmed failing right now on #3511, #3514, #3515 and #3518, plus #3507, #3509 and #3510.

## Why it is a toolchain bump and nothing else

The job installs `govulncheck@latest` and scans with whatever Go `setup-go` installed, at `security.yml:63-76`. That step reads `GO_VERSION`, currently `1.25.12` — one patch below the fix.

Nothing else selects the toolchain: no `.go-version`, no `go-version-file`, no `GOTOOLCHAIN`, no `toolchain` directive in `agent/go.mod` (it declares `go 1.25.10`, a minimum, which 1.25.13 satisfies), no `golang:` base image, and `agent/Makefile` uses whatever `go` is on PATH. So a `setup-go` bump is sufficient and nothing will silently pull 1.25.12 back.

## Suggested fix, and a caution about the minimal version

**Strictly, only `security.yml:17` has to change to make this job green.** That is the tempting fix and I would advise against it on its own.

There are five committed `1.25.12` pins:

```
.github/workflows/security.yml:17        <- governs THIS job
.github/workflows/ci.yml:26              <- 7 setup-go steps
[...]

PR #3523, opened 2026-08-14
fix(security): bump Go toolchain to 1.26.6 to clear six stdlib advisories

ToddHebebrand:
`Go Vulnerability Check` is red on **every open PR** (#3511, #3512, #3514, #3515, #3518, and #3510) and would redden `main` at its next scheduled run. No repo change caused it — govulncheck resolves against a live advisory database, so the same tree that passed on `497d0297f` at 20:40Z failed hours later.

## The six findings

All symbol-level, all stdlib, all reachable from agent code:

| Advisory | Package | Issue |
|---|---|---|
| GO-2026-6218 | `net/url` | quadratic `resolvePath` |
| GO-2026-6090 | `crypto/tls` | unbounded post-handshake messages |
| GO-2026-6088 | `encoding/xml` | missing recursion depth guard |
| GO-2026-5972 | `encoding/asn1` | missing recursion depth guard |
| GO-2026-5026 | `net/http` | idna punycode labels not rejected |
| GO-2026-5942 | `net` (dnsmessage) | panic on invalid SVCB/HTTPS RR |

## Why 1.26.6 and not 1.25.13

1.25.13 is the tempting minimal bump and it clears **five of the six**. It does not clear **GO-2026-5942** — stdlib fixed that one only on the 1.26 branch (OSV records `fixed=1.26.6`, no 1.25 backport).

I verified this empirically rather than trusting the advisory range:

```
GOTOOLCHAIN=go1.25.13 → GO-2026-5051 (allowlisted), GO-2026-5942   ❌
GOTOOLCHAIN=go1.26.6  → GO-2026-5051 (allowlisted)                 ✅
```

That finding matters more than its position in the list suggests. The trace is `heartbeat.handleNetworkDnsCheck` → `net.Resolver.LookupCNAME` — a panic triggerable by a DNS response, in the agent that runs on customer machines.

Allowlisting it was not an option. The allowlist's own documented bar is *"the advisory must have NO fixed release to upgrade to"*, and one exists.

## Scope

- `GO_VERSION` in `ci` / `security` / `release` / `dev-build-agent`, **plus the hard-coded literal in `codeql.yml`** — that job defines no `GO_VERSION` and quietly keeps its own pin.
- `agent/go.mod` `go` directive, so the fix binds for anyone building from source and not only for binaries we release. With the default `GOTOOLCHAIN=auto` this *fetches* 1.26.6 rather than erroring; it only bites an explicit `GOTOOLCHAIN=local` on 1.25.
- The two docs pages that tell self-hosters which toolchain to install.
- **`scripts/security/check-go-version-pins.sh`**, run first in the `go-vuln` job. That job only proves the toolchain it was handed is clean, so a pin drifting anywhere else means we scan one Go and ship another.

## Verification

- `go build`, `go vet`, `go test -race ./...` — green, 69 packages, exit 0. The test run had `go.mod` **already** at 1.26.6, so it exercised Go 1.26 GODEBUG defaults rather than 1.25 compatibility ones.
- `scripts/security/run-govulncheck.sh` (the real CI gate) exits 0.
- The new pin guard is **mutation-tested** — it fails when a workflow drifts back to 1.25.13, when the `codeql.yml` literal drifts, when `go.mod` loses patch precision (`go 1.26`), and when the docs advertise 1.25.

## Behaviour changes carried by the `go` directive

The `go` line gates GODEBUG defaults, so Go 1.26 semantics now apply. I checked each against agent code:

- **`cryptocustomrand=0`** — inert. `math/rand` appears only in backoff jitter and never reaches a crypto API.
- **`urlstrictcolons=1`** — `url.Parse` now rejects unbracketed IPv6 and malformed host colons. The agent parses a configured `ServerURL` (`config/validate.go:87`, `config/serverurl.go:82`), so a self-hoster who wrote `https://fd00::1:8080/` instead of `https://[fd00::1]:8080/` is now rejected at config validation rather than silently misparsed. **Worth a release note.**
- **`tlsmlkem`** — two additional hybrid PQ key exchanges offered. Hybrid ML-KEM has been on by default since 1.24, so this is incremental rather than new.
- **`urlmaxqueryparams=10000`, `httpcookiemaxnum=3000`, `htmlmetacontenturlescape`** — no agent path approaches these.

## Merge order

This does **not** fix `Trivy Filesystem Scan`, the other repo-wide red check. #3510 (nanoid `CVE-2026-67213`, @bdunncompany) covers that one. Neither PR can go fully green on its own — whichever lands first needs `--admin`, then the other rebases clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ concurrency:
 
 env:
   PNPM_VERSION: '10.34.5'
-  GO_VERSION: '1.25.12'
+  GO_VERSION: '1.26.6'
 
 # SR-007: least-privilege default. CI runs on every PR (including forks) and
 # only needs to read the repo. Any job needing more must opt in per-job.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -34,7 +34,7 @@ jobs:
         if: matrix.language == 'go'
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
-          go-version: '1.25.12'
+          go-version: '1.26.6'
           cache-dependency-path: agent/go.sum
 
       - name: Initialize CodeQL
--- a/.github/workflows/dev-build-agent.yml
+++ b/.github/workflows/dev-build-agent.yml
@@ -24,7 +24,7 @@ permissions:
   contents: read
 
 env:
-  GO_VERSION: '1.25.12'
+  GO_VERSION: '1.26.6'
 
 jobs:
   resolve:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ on:
 
 env:
   PNPM_VERSION: '10.34.5'
-  GO_VERSION: '1.25.12'
+  GO_VERSION: '1.26.6'
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
 
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -14,7 +14,7 @@ concurrency:
 
 env:
   PNPM_VERSION: '10.34.5'
-  GO_VERSION: '1.25.12'
+  GO_VERSION: '1.26.6'
 
 # Least-privilege default; the SARIF-uploading job opts into
 # security-events: write per-job below.
@@ -59,6 +59,12 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           persist-credentials: false
+
+      # Runs before Setup Go on purpose: this job only proves the toolchain it
+      # was handed is clean, so a pin that drifts anywhere else (release.yml
+      # builds the shipped binary) would scan one Go and ship another.
+      - name: Check Go toolchain pins agree
+        run: bash scripts/security/check-go-version-pins.sh
 
       - name: Setup Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
--- a/agent/go.mod
+++ b/agent/go.mod
@@ -1,6 +1,6 @@
 module github.com/breeze-rmm/agent
 
-go 1.25.10
+go 1.26.6
 
 require (
 	cloud.google.com/go/storage v1.64.0
--- a/apps/docs/src/content/docs/agents/building.mdx
+++ b/apps/docs/src/content/docs/agents/building.mdx
@@ -10,7 +10,7 @@ import { Aside } from '@astrojs/starlight/components';
 
 ## Prerequisites
 
-- Go 1.25 or later
+- Go 1.26.6 or later
 - Make
 - (Optional) `go-winres` for Windows resource embedding
 - (Optional) WiX Toolset + PowerShell for MSI packaging
--- a/apps/docs/src/content/docs/getting-started/prerequisites.mdx
+++ b/apps/docs/src/content/docs/getting-started/prerequisites.mdx
@@ -78,9 +78,9 @@ Caddy will automatically provision a TLS certificate from Let's Encrypt once DNS
 Only needed if building the agent from source (pre-built binaries are available):
 
 ```bash
-# Go 1.25+
-wget https://go.dev/dl/go1.25.0.linux-amd64.tar.gz
-sudo tar -C /usr/local -xzf go1.25.0.linux-amd64.tar.gz
+# Go 1.26.6+
+wget https://go.dev/dl/go1.26.6.linux-amd64.tar.gz
+sudo tar -C /usr/local -xzf go1.26.6.linux-amd64.tar.gz
 export PATH=$PATH:/usr/local/go/bin
 ```
 
--- a/scripts/security/check-go-version-pins.sh
+++ b/scripts/security/check-go-version-pins.sh
@@ -9,6 +9,12 @@
 #   - the `go` directive in agent/go.mod
 #   - two public docs pages that tell self-hosters what to install
 #
+# DELIBERATELY EXCLUDED: selfhost-signing-template/.github/workflows/ci.yml
+# pins go-version 1.24. That file is a template copied into a SELF-HOSTER'S
+# own repository, where it only installs actionlint — it never builds the
+# Breeze agent, so it is not one of "our" toolchain pins and is out of the
+# .github/workflows/*.yml glob on purpose, not by oversight.
+#
 # Why this guard exists: the pin was bumped 1.25.12 -> 1.26.6 to clear six
 # stdlib advisories (GO-2026-5026/5942/5972/6088/6090/6218). Five of the six
 # are also fixed in 1.25.13, but GO-2026-5942 (dnsmessage SVCB/HTTPS RR panic,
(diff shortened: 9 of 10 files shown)


bdunncompany:
Reviewed this since I filed #3522. **The 1.26.6 choice holds up under independent checking**, and the pin-drift guard is a better answer than what I suggested.

## Corroborating 1.26.6 specifically, against the obvious pushback

Someone will reasonably ask "why not 1.25.13, which fixes five of the six?" Two things close that off:

1. **GO-2026-5942 has no 1.25.x stdlib fix at all.** The advisory lists fixed versions `0.56.0` (module `golang.org/x/net`) and `1.26.6` / `1.27.0-rc.3` (stdlib). The 1.25 branch is simply absent.
2. **The module-side remedy is already spent.** `agent/go.mod:31` requires `golang.org/x/net v0.57.0`, which is *above* the 0.56.0 module fix, and there is no `replace`, `exclude`, `toolchain` directive, tracked `vendor/`, or second `go.mod` anywhere on `main` that could change the effective version. So bumping the dependency cannot help; the remaining exposure is the copy vendored into the standard library, and only the toolchain moves that.

I also confirmed `go1.26.6` is actually published (newest 1.26.x on the official release list, tarball resolves), since a pin to an unreleased patch would break `setup-go` everywhere.

## The guard is the right shape

`check-go-version-pins.sh` addresses the exact failure I was worried about in #3522, and better than a one-off bump would: deriving the expected value from `agent/go.mod` and rejecting a two-part `go 1.26` directive is the part that keeps this from silently regressing to "any patch release."

Two small hardening notes, neither blocking:

- **The docs check is presence-only.** `grep -qF "$pinned"` passes as soon as the version appears anywhere in the file. A page that said "install Go 1.25.12" and *also* mentioned 1.26.6 elsewhere would pass while telling self-hosters to install an unpatched toolchain. Both docs are fully updated in this PR, so this is latent, not live. Asserting the absence of a stale `go1\.25\.` pattern would close it.
- **`selfhost-signing-template/.github/workflows/ci.yml:25` pins `go-version: '1.24'`** and sits outside the `.github/workflows/*.yml` glob. That is almost certainly deliberate, since it is a template for a *different* repository, but the header says "keep every Go toolchain pin in the repo agreeing", which reads broader than the code. A one-line note about the deliberate exclusion would save the next reader the same detour I just took.

## One observability gap worth considering separately

`scripts/security/run-govulncheck.sh` retains only the advisory ID, not the govulncheck trace. That is what makes a reachability claim like "`handleNetworkDnsCheck` → `net.Resolver.LookupCNAME`" unverifiable after the fact from CI output alone — I could not confirm it from source, since ordinary `net.Resolver` skips unrequested RR types rather than parsing them as SVCB/HTTPS bodies. I am **not** disputing the finding; govulncheck's symbol-level analysis is the authority here and source reading is the weaker instrument. Retaining the trace on failure would just make the next one self-evidencing.

Minor terminology point for the PR description: Go calls 1.26.0 a *major* release and 1.26.1+ *minor revisions*, so 1.25 → 1.26 is a bigger step than "patch". Your full build/test matrix across Linux, Darwin (with cgo) and Windows is the right acceptance gate for it, and it already covers those.

bdunncompany:
Follow-up now that CI has run: **`Go Vulnerability Check` passes on this branch**, so the 1.26.6 bump does what it set out to do.

The remaining red here is `Trivy Filesystem Scan`, and it is **not** something this PR introduced or should carry:

```
Total: 1 (HIGH: 1, CRITICAL: 0)
nanoid  CVE-2026-67213  HIGH  installed 3.3.17  fixed 3.3.18, 5.1.6
```

That is the other half of the same "advisory database moved" event, and it is what #3510 fixes. The two are complementary and neither alone produces a green Security Scanning run:

|---|---|---|
| this PR (#3523) | **pass** | fail (nanoid) |
| #3510 | fail (Go stdlib) | **pass** |
| both | pass | pass |

I test-merged them to check they compose rather than assuming it: clean merge, no conflicts, and the combined tree carries `nanoid` 3.3.18 / 5.1.16 / 6.0.0 alongside `go 1.26.6` in both `agent/go.mod` and the workflow pins. #3510 touches only `package.json` and `pnpm-lock.yaml`, this PR touches no JS lockfile, so there is no ordering hazard either way.

One thing I checked and want to record, because it looks alarming and is not: `apps/{helper,viewer}/src-tauri/Cargo.lock` carry `quick-xml` 0.38.4/0.39.2, which are inside RUSTSEC-2026-0194 and RUSTSEC-2026-0195 (both fixed in 0.41.0). Cargo Audit is green because both are already in the reviewed `ignore` lists in `.cargo/audit.toml`, with the justification that the only consumers are `plist` on your own bundle metadata and build-time `wayland-scanner`. So that is not a third blocker, it is working as designed.

The honest caveat on all of this: Trivy, govulncheck and cargo-audit all resolve against live databases, so "green on merge" is not "green tomorrow". Three separate scanners moved under this repo inside 24 hours. The `go version` print you added and the pin guard both help make the next one diagnosable rather than mysterious.

ToddHebebrand:
Thanks — all three notes addressed in `dab0bf7fa`, plus a correction to my own fix.

**1. Docs check was presence-only.** Fixed, and mutation testing caught a bug in my first attempt at the fix: the regex excluded `.` as a trailing character, so `Go 1.25.12.` at the end of a sentence — the most likely way a stale pin actually gets written — slipped through the check that exists to catch it. The check now passes against four phrasings (`Go 1.25.12.`, `Go 1.25 or later`, `go1.25.13`, `(Go 1.25)`) and does not false-positive on unrelated versions like `1.256` or `v1.250-beta`.

**2. `selfhost-signing-template` exclusion.** You read it right — it's copied into a self-hoster's own repo and only installs actionlint, never builds the agent. Now stated in the header so the next reader doesn't repeat your detour.

**3. Trace retention.** Good call, and it applies directly to your own inability to confirm the 5942 reachability claim from source. Failures now print the vulnerable symbol and our calling frame:

```
FAIL GO-2026-5051 — symbol-level vulnerability not in the allowlist
     github.com/hirochachacha/go-smb2.ReadDir  <-  github.com/breeze-rmm/agent/internal/workspaceindex.ReadDir (internal/workspaceindex/fs_smb.go:87)
```

`trace[0]` is the vulnerable symbol and the last frame is ours — I initially printed only `trace[0]`, which is the half that *doesn't* answer "which of our paths". Verified by temporarily emptying the allowlist to exercise the failure path, then confirming the allowlist came back byte-identical.

**On the terminology point** — you're right, 1.25 → 1.26 is a major release in Go's scheme, not a patch. I've said "patch bump" loosely in a few places; the PR body's build/test matrix is the acceptance gate that actually matters for it.

**On "green on merge ≠ green tomorrow"** — agreed, and it's the most useful thing in your review. Three scanners moved under this repo inside 24 hours. Neither the pin guard nor the trace output prevents the next one; they just make it diagnosable in minutes instead of hours. Worth considering a scheduled `main` run that opens an issue on failure, so the next advisory lands as a notification rather than as six red PRs.

I have not merged. #3510 still needs to land for a fully green board.

bdunncompany:
Verified `dab0bf7fa`. The stale-pin check does what you say it does, and the trace change answers the right question.

**The regex holds up.** I ran it against your four phrasings plus two I invented, under GNU grep 3.11 in an `ubuntu:24.04` container to match the runner:

```
CAUGHT: Go 1.25.12.          CAUGHT: requires Go 1.25.0+
CAUGHT: Go 1.25 or later     CAUGHT: v1.25.12
CAUGHT: go1.25.13            clean : version 1.256
CAUGHT: (Go 1.25)            clean : v1.250-beta
                             clean : Go 1.26.6
```

**A gotcha for anyone verifying this on macOS**, because it cost me a few minutes and would have had me reporting your fix as broken: if `grep` on the reviewer's Mac is **ugrep** (a common Homebrew drop-in, `grep --version` says `ugrep 7.5.0` rather than `GNU grep`), the same pattern matches *nothing* for the space-preceded cases. `go1.25.13` matches, `Go 1.25.12.` does not. Not your problem, and CI is unaffected since the runner is GNU, but a `# verified against GNU grep; ugrep differs` note beside the pattern would save the next person the detour.

**One durability nit, genuinely minor.** Everything else in the guard derives from `agent/go.mod`, which is what makes it self-maintaining. The stale-reference pattern is the one hardcoded value:

```
stale="$(grep -nE '(^|[^0-9.])(go)?1\.25(\.[0-9]+)?([^0-9]|$)' "$doc" || true)"
```

It is exactly right for this transition. At the next bump it silently stops covering the line you just left: with `pinned=1.27.x`, a leftover `Go 1.26.6` in the docs passes (confirmed clean above). Deriving the previous-minor from `$pinned`, or asserting "no `go`-prefixed version other than `$pinned`", would keep it honest across bumps. Not worth blocking on.

**On trace retention** — printing `trace[0]` plus the last frame is the right pair. Your `go-smb2.ReadDir <- workspaceindex.ReadDir (fs_smb.go:87)` example is precisely the shape that would have let me confirm or refute the 5942 reachability claim from CI output instead of guessing at it from source, which is where I ran out of road. Exercising the failure path by temporarily emptying the allowlist, then verifying it came back byte-identical, is a better check than I would have thought to ask for.

**On the scheduled-run-opens-an-issue idea** — worth doing, and the shape matters: three scanners moved under this repo in 24 hours, and each time the first signal was six red PRs rather than one notification. `security.yml` already has the Monday cron; having a failure there open or update a single tracking issue would turn the next one into a Monday morning ticket instead of a morning of confused PR authors.

#3510 is ready whenever you want it — MERGEABLE, Trivy green, and its only red is the Go check this PR fixes.

Issue #3381, opened 2026-08-10
[Software Policies] Allowlist mode (allowUnknown: false) should require an explicit, scope-aware confirmation before it can be linked into a fleet-wide Configuration Policy


cisspUser01:
### Problem

Creating a Software Policy allowlist and linking it into Configuration Policies assigned across an
entire fleet is currently possible with no warning, preview, or confirmation step that reflects the
actual, severe consequence: allowUnknown: false means every piece of software NOT on the approved
list gets automatically uninstalled from every covered device, including security software,
VPN clients, and any other legitimate application not explicitly whitelisted.

This is a real, high-consequence action reachable through what reads as an innocuous workflow:
clicking "Approve" on a few items in the Software Inventory tab, then linking that resulting
allowlist into existing site-wide Configuration Policies. Nothing in that flow makes clear that
"approve these 3 items" means "everything else on every covered device will be removed," rather
than the more intuitive reading of "these 3 are additionally permitted."

### Proposed Solution

In production use, a self-hosted deployment approved 3 items via Software Policies > Inventory,
then linked the resulting "Default Allowlist" (allowUnknown: false) into 4 Configuration Policies
covering effectively the entire fleet. Enforcement fired the next business day:

- 259 devices had violations detected against the 3-item allowlist
- Software actually removed included Cisco Secure Client (AnyConnect VPN + Umbrella + DART),
  Microsoft Office components, Google Chrome, Adobe Reader, and dozens of other legitimate
  applications
- Remote/VPN-dependent users lost the ability to connect via VPN, and lost the specific tool
  (Cisco Secure Client) that could have been used to remotely troubleshoot the problem, since
  Breeze itself was configured for LAN/VPN-only remote access
- Full incident diagnosis, containment, and majority remediation took the better part of a weekend

No malicious intent, no unusual configuration — this resulted from following the feature's own
UI flow (Approve items -> link policy -> assign to sites) as presented, without any point where the
actual scope or consequence was surfaced before the action went live.

### Alternatives Considered

1. Default allowUnknown to true (permissive/additive) rather than false (strict allowlist) unless
   the operator explicitly opts into strict mode via a clearly-labeled, separate control — most
   operators' mental model when "approving" software is additive, not a full default-deny switch.

2. When allowUnknown: false IS explicitly selected, show a clear, specific warning before save:
   e.g. "This policy will treat any software NOT on this list as unauthorized and remove it. This
[...]

PR #3548, opened 2026-08-14
fix(api): audit AI software-policy writes and gate remediation on policy arming (#3543)

### LINKED ISSUE #3543: [API] AI software-policy tools bypass both audit stores; remediate_software_violation ignores enforceMode
## Description

Found while investigating the forensics in #3381. The built-in AI agent's software-policy tools mutate enforcement state and queue uninstalls with no audit trail, and one of them bypasses the enforcement gate entirely.

1. **No audit events from AI software-policy tools.** The create/update software-policy tools (`apps/api/src/services/aiToolsCompliance.ts`, `apps/api/src/services/aiToolsPolicyPrereqs.ts`) can set `enforceMode` and `remediationOptions` but contain zero calls to `writeRouteAudit` or `recordSoftwarePolicyAudit`. The equivalent web routes write to both `audit_logs` (`software_policy.update`) and `software_policy_audit` (`policy_updated`). An operator reconstructing "who armed enforcement on this policy" from the audit trail cannot see AI-tool writes at all. Verified on current main and at v0.101.0.

2. **`remediate_software_violation` ignores `enforceMode`.** The tool (`aiToolsCompliance.ts:~403-490`) queues `software_uninstall` commands for every violating device while only refusing `mode === 'audit'` policies; `autoUninstall` defaults to true. The compliance worker's gate (`enforceMode && mode !== 'audit' && remediationOptions.autoUninstall`, `softwareComplianceWorker.ts:397-402`) is the only place that gate exists — the remediation worker never re-checks it. So a detect-only policy (enforcement deliberately off) can still have mass uninstalls queued through this tool, unaudited. It is Tier 3 / approval-gated, but approval-gating is not an audit trail and is not the policy's own enforcement setting.

## Proposed Fix

- Emit `recordSoftwarePolicyAudit` (actor `'ai'` — the enum already supports it) and an `audit_logs` event from every AI tool that creates/updates a software policy or queues remediation, mirroring the route handlers.
- Make `remediate_software_violation` respect `enforceMode` (or at minimum require an explicit, logged override), and consider re-checking the gate in `softwareRemediationWorker` as defense in depth.

## Affected Files

- `apps/api/src/services/aiToolsCompliance.ts` (primary)
- `apps/api/src/services/aiToolsPolicyPrereqs.ts`
- `apps/api/src/jobs/softwareRemediationWorker.ts` (defense-in-depth gate)

## Reported By

Forensic analysis with @cisspUser01 in #3381 — their complete `audit_logs` trail for an enforced policy contains no arming event, which these gaps could explain.

ToddHebebrand:
Closes #3543.

Found while investigating the forensics in #3381 (259 devices mass-uninstalled) — that reporter's complete `audit_logs` trail contained **no policy-arming event**, which these gaps explain.

## The three gaps

1. **Neither audit store was written by the AI tools.** `manage_software_policy` (`aiToolsCompliance.ts`) and `manage_software_policies` (`aiToolsPolicyPrereqs.ts`) can set `enforceMode` and `remediationOptions` — the two flags that arm a policy to uninstall software from real machines — and contained zero audit writes. The equivalent web routes write to **both** `audit_logs` (`software_policy.create`/`.update`) and `software_policy_audit` (`policy_created`/`policy_updated`).
2. **`remediate_software_violation` ignored `enforceMode`.** It refused only `mode === 'audit'`, so a detect-only policy — enforcement deliberately left off by its owner — could still have fleet-wide uninstalls queued through the model. Tier-3 approval is a prompt-time check on the *action*, not the policy's own enforcement setting.
3. **The remediation worker never re-checked the gate.** It uninstalled whatever it was handed. The arming rule existed only inline in `softwareComplianceWorker.ts:397-402`.

## What changed

**Shared arming gate.** `evaluateSoftwarePolicyArming` in `services/softwarePolicyService.ts` is now the single definition of `mode !== 'audit' && enforceMode && remediationOptions.autoUninstall === true`, returning a machine-readable `reason` (`audit_mode` / `enforce_mode_off` / `auto_uninstall_off`) plus an operator-legible message. `autoUninstall` is opt-in — absent or non-object `remediationOptions` means **not** armed.

**Audit fan-out.** New `services/aiToolsSoftwarePolicyAudit.ts` writes both stores. AI tool handlers receive `auth`, not a Hono context, so `writeRouteAudit(c, …)` is unusable; this follows the context-less pattern already established in `aiToolsOrgs.ts` (`writeAuditEvent(requestLikeFromSnapshot({}), …)`, actor identity off `auth.user`). `recordSoftwarePolicyAudit` is called with `actor: 'ai'` and the acting user as `actorId`. `enforceMode`/`remediationOptions` changes are surfaced explicitly (`enforceMode`, `autoUninstall`) rather than buried in a generic `updatedFields` list — and `updatedFields` is derived from the columns actually written, not raw tool input, so routing keys like `action`/`policyId` don't overstate the change. Wired into create/update/delete on both tools and both outcomes of the remediation tool.

**`remediate_software_violation` hard-refuses** an unarmed policy, returns the reason to the model, and audits the refusal to both stores as `remediation_denied` / `result: 'denied'`. **There is deliberately no override parameter** — re-arming is an administrator's decision made on the policy, not an argument the model can supply.

**Worker re-check (defense in depth).** `softwareRemediationWorker.processRemediateDevice` re-evaluates arming before any `software_uninstall` is queued, and before it flips the compliance row to `in_progress`. Unarmed automatic jobs skip with `recordSoftwareRemediationDecision('policy_not_armed')` + a `remediation_skipped_unarmed` policy-audit row.

## Design call: the manual web route stays ungated

`POST /software-policies/:id/remediate` (execute permission + MFA + tenant/site filtering + both audit stores) keeps working on detect-only policies. Rationale:

- `enforceMode` authorises **unattended** remediation — the UI labels it "Enforce (auto-remediate)" and nests `autoUninstall` under it. Gating an explicit human action on an automation flag repurposes the field.
- Uniform gating would be **more** dangerous in practice: to run one reviewed uninstall an admin would first have to arm automatic fleet-wide remediation, which is a strictly larger grant than the click they wanted, and races the compliance evaluator.
- The UI deliberately shows Remediate for every non-audit policy, so gating it is a silent regression.

I formed this position first and then got an independent read from Codex (read-only, `xhigh`), which reached the same recommendation for the same reasons — so no quorum split to surface.

Mechanically: `scheduleSoftwareRemediation` takes an optional `{ trigger, requestedByUserId }`, defaulting to the **gated** `auto` path so a new producer cannot accidentally inherit the override. The route stamps `trigger: 'manual'` **server-side, never from the request body**. The worker exempts only the exact literal `'manual'` — `'Manual'`, `true`, a missing field, or a job enqueued before this change are all treated as automatic and gated, so provenance fails closed. A manual override against an unarmed policy is recorded as `remediation_manual_override` naming the requesting user.

**Known residual, called out honestly:** `trigger: 'manual'` is authorization state living in unauthenticated BullMQ job data. A Redis writer could forge it — though anyone with that access can already enqueue device commands directly. The stronger fix Codex proposed (a durable one-time remediation-request record consumed atomically by the worker) needs a new table, which is out of scope here; worth a follow-up issue if we want to close it properly. The forged case is at least always visible in the audit trail.

**BullMQ jobId collision:** auto and manual jobs share `software-remediation-<policy>-<device>` and can dedupe into each other. This cannot produce a wrong gate outcome — auto jobs are only ever produced for armed policies, so on an unarmed policy every job is manual, and on an armed policy both triggers pass anyway. Documented inline rather than changing the key (a per-trigger key would risk concurrent duplicate uninstall commands, since the in-flight check isn't atomic).

## Separate finding, not fixed here

`manage_software_policies` in `aiToolsPolicyPrereqs.ts` is registered **Tier 1 (auto-execute, no approval)** while being able to set `enforceMode` and `remediationOptions.autoUninstall`. Its sibling `manage_software_policy` in `aiToolsCompliance.ts` is Tier 3. Post-PR it is at least fully audited, but arming enforcement without approval looks like a tier misclassification. Raising it changes behaviour for existing workflows, so I've left it alone — flagging for a product call.

No schema or migration changes; both audit tables already exist. The compliance worker's existing gate is untouched.

## Verification

- `apps/api` affected suite, single fork: **9 files / 118 tests green** — `aiToolsCompliance.auditAndArming.test.ts` (new), `aiToolsPolicyPrereqs.softwarePolicyAudit.test.ts` (new), `softwareRemediationWorker.test.ts` (new), plus `aiToolsCompliance.siteScope`, `aiToolsPolicyPrereqs`, `softwarePolicies` (route), `softwareComplianceWorker`, `aiToolsRegistryParity`, `aiTools.deviceArgsCoverage.contract`.
- `npx tsc --noEmit` clean.
- **Mutation-checked, not just green:** stubbing `evaluateSoftwarePolicyArming` to always return `armed: true` kills **12** of the new tests, so the gate assertions are load-bearing rather than vacuous. Tests assert audited values (action names, `actor`, `orgId`, the enforcement fields, the refusal reason) rather than "an audit function was called", and the worker suite keeps the *real* evaluator so a stubbed gate can't pass for it.
- Two pre-existing suites needed fixture updates, both genuine consequences: `aiToolsCompliance.siteScope.test.ts` policies had to become actually armed, and `aiToolsPolicyPrereqs.test.ts` needed the new helper's transitive schema-barrel import cut.

### DIFF
--- a/apps/api/src/jobs/softwareRemediationWorker.test.ts
+++ b/apps/api/src/jobs/softwareRemediationWorker.test.ts
@@ -65,9 +65,12 @@ const ORG_ID = 'org-1';
 
 function chain(result: unknown): any {
   const p: any = Promise.resolve(result);
-  for (const m of ['from', 'innerJoin', 'where', 'limit', 'set', 'orderBy']) p[m] = () => p;
+  for (const m of ['from', 'innerJoin', 'where', 'limit', 'orderBy']) p[m] = () => p;
   return p;
 }
+
+/** Captures the payload passed to `db.update(...).set(...)`. */
+let setSpy: ReturnType<typeof vi.fn>;
 
 function policyRow(overrides: Record<string, unknown> = {}) {
   return {
@@ -107,7 +110,8 @@ function primeDb(policy: unknown, opts: { compliance?: unknown } = {}) {
   ];
   let call = 0;
   selectMock.mockImplementation(() => chain(results[Math.min(call++, results.length - 1)]));
-  updateMock.mockImplementation(() => chain([]));
+  setSpy = vi.fn(() => chain([]));
+  updateMock.mockImplementation(() => ({ set: setSpy }));
 }
 
 function policyAuditActions(): string[] {
@@ -129,8 +133,18 @@ describe('processRemediateDevice — arming gate for automatic jobs (#3543)', ()
     const audit = recordPolicyAuditMock.mock.calls[0]![0] as any;
     expect(audit).toMatchObject({ orgId: ORG_ID, policyId: POLICY_ID, deviceId: DEVICE_ID, actor: 'system' });
     expect(audit.details).toMatchObject({ reason: 'enforce_mode_off', trigger: 'auto' });
-    // The gate must fire before the row is flipped to in_progress.
-    expect(updateMock).not.toHaveBeenCalled();
+
+    // The refusal must be reflected on the compliance row: leaving it at the
+    // enqueue-time 'pending' would later be rewritten to 'completed' by the
+    // compliance worker, falsely claiming the uninstall succeeded.
+    expect(updateMock).toHaveBeenCalledTimes(1);
+    const written = setSpy.mock.calls[0]![0];
+    expect(written.remediationStatus).toBe('failed');
+    expect(written.remediationErrors[0].message).toMatch(/not armed/i);
+    expect(written.remediationErrors[0].message).toContain('enforce_mode_off');
+    // Crucially it must NOT be left pending, and must not be flipped to in_progress.
+    expect(written.remediationStatus).not.toBe('pending');
+    expect(written.remediationStatus).not.toBe('in_progress');
   });
 
   it('skips when enforceMode is on but autoUninstall is not armed', async () => {
--- a/apps/api/src/jobs/softwareRemediationWorker.test.ts
+++ b/apps/api/src/jobs/softwareRemediationWorker.test.ts
@@ -1,0 +1,247 @@
+/**
+ * #3543 — defense-in-depth arming gate in the remediation worker.
+ *
+ * This worker is the last hop before `software_uninstall` commands reach real
+ * machines. Before #3543 it uninstalled whatever it was handed: the gate
+ * (`enforceMode` + non-audit mode + `remediationOptions.autoUninstall`) existed
+ * only in the compliance evaluator that produces automatic jobs, so a stale,
+ * replayed or hand-enqueued job could still uninstall software from a policy
+ * that had since been disarmed (incident #3381: 259 devices mass-uninstalled).
+ *
+ * An explicit operator remediation (`trigger: 'manual'`, stamped server-side by
+ * the MFA-gated route) is deliberately exempt — `enforceMode`/`autoUninstall`
+ * authorise UNATTENDED remediation — but is loudly audited when the policy is
+ * unarmed. Anything that is not exactly 'manual' is treated as automatic and
+ * gated, so provenance fails closed.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { addMock, getJobMock, queueCommandMock, recordDecisionMock, recordPolicyAuditMock } = vi.hoisted(() => ({
+  addMock: vi.fn(async (..._args: any[]) => ({ id: 'queued-job-1' })),
+  getJobMock: vi.fn(async () => null),
+  queueCommandMock: vi.fn(async (..._args: any[]) => ({ id: 'cmd-1' })),
+  recordDecisionMock: vi.fn((..._args: any[]) => {}),
+  recordPolicyAuditMock: vi.fn(async (..._args: any[]) => {}),
+}));
+
+vi.mock('bullmq', () => ({
+  Queue: class { add = addMock; getJob = getJobMock; close = vi.fn(); },
+  Worker: class { on = vi.fn(); close = vi.fn(); },
+  Job: class {},
+}));
+vi.mock('../services/redis', () => ({ getBullMQConnection: vi.fn(() => ({})) }));
+vi.mock('../routes/metrics', () => ({ recordSoftwareRemediationDecision: recordDecisionMock }));
+vi.mock('../services/commandQueue', () => ({
+  queueCommand: queueCommandMock,
+  CommandTypes: { SOFTWARE_UNINSTALL: 'software_uninstall' },
+}));
+vi.mock('../services/sentry', () => ({ captureException: vi.fn() }));
+vi.mock('../services/bullmqUtils', () => ({ isReusableState: vi.fn(() => false) }));
+vi.mock('../services/softwarePolicyService', async (orig) => {
+  // Keep the REAL arming evaluator — the whole point of the suite is that the
+  // gate itself is correct, not that a stubbed gate was consulted.
+  const actual = await orig<typeof import('../services/softwarePolicyService')>();
+  return {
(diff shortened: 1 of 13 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (all three on the full `origin/main...HEAD` diff). A separate read-only Codex `xhigh` consult ran *before* implementation on the one consequential design call (whether the manual web remediate route should also be gated); it independently reached the same recommendation, so there was no quorum split to surface.

**Findings:** 4 raised → all 4 addressed in `e772481b0`; 0 outstanding.

- **Critical (silent-failure-hunter)** — the arming-gate skip left `software_compliance_status` parked at the enqueue-time `'pending'` with no explanation. `softwareComplianceWorker` rewrites any non-terminal `remediationStatus` to `'completed'` the next time the device scans compliant, so a *refused* job would eventually have claimed Breeze's uninstall succeeded when it never ran — the exact opposite of this PR's purpose. Verified against `softwareComplianceWorker.ts:358-368`. The skip branch now writes `remediationStatus: 'failed'` plus a `remediationErrors` message naming the arming reason; the gate moved to just after the compliance-row **read** so the refusal is recordable, still ahead of every mutation and of `queueCommand`.
- **Gate drift (pr-test-analyzer)** — `softwareComplianceWorker`'s inline gate was deliberately left untouched per scope, but two independent copies of a security gate drift, and drift *there* reintroduces the #3381 bug class. New `softwarePolicyArming.test.ts` pins them together over a 3×4×10 truth table of `mode`/`enforceMode`/`remediationOptions`, and unit-tests `readSoftwarePolicyAutoUninstall` against non-object and truthy-but-not-`true` inputs (array, string, number, bare boolean) no caller fixture reached.
- **Vacuous assertion (pr-test-analyzer)** — "the AI path never requests the manual override trigger" checked `options?.trigger !== 'manual'` where `options` is always `undefined`, so it held regardless of the code. Replaced with a call-arity assertion that fails if an options object is ever introduced. Good catch; this is exactly the trap class the repo warns about.
- **Partner-wide coverage gap (pr-test-analyzer)** — no fixture exercised `org_id NULL` / `partner_id` set, the shape where `recordSoftwarePolicyAudit` throws if both owner axes are null. Added on both the create and remediate paths.

code-reviewer returned **0 findings**, explicitly confirming gate parity with the pre-existing inline check, that no mutation or `queueCommand` precedes the gate, that `readTrigger` is genuinely fail-closed, that `trigger` is unreachable from any request body, and that no new call site can pass both owner axes null.

**Tests:** `apps/api` affected suite green single-fork — 8 files / **129 tests** (`vitest run` over `aiToolsCompliance.auditAndArming`, `aiToolsCompliance.siteScope`, `aiToolsPolicyPrereqs`, `aiToolsPolicyPrereqs.softwarePolicyAudit`, `softwarePolicyArming`, `softwareRemediationWorker`, `softwareComplianceWorker`, `softwarePolicies` route). `tsc --noEmit` clean. **Mutation re-checked** after the fixes: stubbing `evaluateSoftwarePolicyArming` to always-armed kills **17** tests (was 12 pre-review), so the gate assertions remain load-bearing. Full CI green including all 4 Integration Tests shards.

**Status:** review-clean, awaiting maintainer merge. Two things worth a decision at merge time, both written up in the PR description rather than silently resolved: (1) `trigger: 'manual'` is authorization state in unauthenticated BullMQ job data — forgeable by a Redis writer; the durable-request-record fix needs a new table and was left out of scope. (2) `manage_software_policies` in `aiToolsPolicyPrereqs.ts` is **Tier 1 / auto-execute** while able to arm enforcement, unlike its Tier 3 sibling — now fully audited, but the tier itself looks like a misclassification and changing it is a product call.

PR #2185, opened 2026-07-03
feat(devices): active VPN client presence telemetry (#2139)

ToddHebebrand:
## What & why

Adds first-class **active-VPN-client presence** telemetry so operators can see at a glance which endpoints have a live VPN overlay tunnel — and the overlay IP/DNS name to reach them — without opening each device or fanning out remote tools. Implements the accepted scope in #2139 (read-only v1; no VPN management, peers, or secrets).

## Approach

**Storage — no new table.** VPN presence is latest-snapshot current state that must appear on the Devices *list* (which selects from `devices` directly), so it's an `active_vpns` jsonb column on `devices`, mirroring the `battery_status` snapshot pattern (#2142). It inherits the devices table's forced RLS — no new tenant-scoped table, no allowlist ceremony. `null` = agent never reported; `[]` = reported with no active VPN.

**Agent** (`agent/internal/collectors/vpn.go` + per-OS files): interface-driven detection (the reliable cross-platform "tunnel is up with an overlay IP" signal), provider classification from interface/adapter name, disambiguated by per-OS service signals (systemd/launchd/Windows services) + cross-platform process signals when the adapter is generic (`utun`/`tun`/`tap`). Providers: WireGuard, Tailscale, NetBird, ZeroTier, OpenVPN, Cloudflare WARP, generic fallback. Best-effort Tailscale MagicDNS name (own device only, never blocks/fails collection). Rides the existing 15-min network-inventory cadence — no hot loop.

**API/shared:** `VpnPresence` type + validator in `packages/shared`; ingest stamps `reportedAt` server-side and writes the snapshot in the same txn as the adapter replace; list + detail responses carry `activeVpns`.

**Web:** optional (default-off) "VPN" Devices column — compact provider badges, dash when none, `+N` overflow, hover tooltip (provider/interface/IP/DNS), sortable; a client-side VPN facet (any / by-provider) mirroring the existing class facet; device-detail VPN section with the full list. Rendered **only** from cached inventory — never a live command fan-out from the table. `data-testid` on all new elements.

## Verification

- **Go**: `go test -race ./internal/collectors/` green; builds for windows/linux/darwin.
- **API**: new `inventoryNetworkVpn.test.ts` (reportedAt stamping, empty-array for old agents, unknown-provider reject, 404) + existing inventory/schemas/autoMigrate suites green; `tsc --noEmit` clean.
- **Web**: new `vpnProviders.test.ts` + all 380 devices component tests green; `astro-check` 0 errors.

## Notes

[...]

PR #3549, opened 2026-08-14
fix(agent,web,api): VPN on/off state and macOS multi-VPN attribution (#2139)

### LINKED ISSUE #2139: Show active VPN clients and VPN IPs on devices
### Problem

Breeze already collects useful device networking information, but the Devices view does not make it easy to see whether an endpoint is currently connected to a VPN overlay network.

For operators who use tools like WireGuard, NetBird, Tailscale, ZeroTier, OpenVPN, Cloudflare WARP, or similar VPN clients, it is useful to know at a glance which devices have an active VPN tunnel and what VPN address or DNS name should be used to reach them.

Today, this usually requires opening a device, checking network adapters, checking services, or using remote tools manually. That does not scale well across a fleet.

### Proposed Solution

Add first-class VPN presence telemetry to agent network inventory and surface it in the Devices UI.

The agent should detect active VPN clients from local network interfaces and known VPN service/process signals, then report normalized VPN status as part of device inventory.

Suggested providers to recognize initially:

- WireGuard
- Tailscale
- NetBird
- ZeroTier
- OpenVPN
- Cloudflare WARP
- Generic VPN / unknown tunnel adapter

For each detected VPN, report structured metadata such as:

- Provider name
- Active/inactive state
- Interface name
- VPN IPv4 and/or IPv6 address
- VPN DNS name or device name when safely available
- Detection source, such as interface, service, process, or adapter description
- Last reported timestamp

The Devices page should add an optional VPN column.

Suggested UI behavior:

- Show small provider icons or compact badges for each active VPN.
- If no VPN is active, show a dash.
- Hovering over a VPN icon should show provider, interface name, VPN IP, and any available VPN DNS name.
- If multiple VPNs are active, show multiple icons or a `+N` overflow indicator.
- The column should be sortable/filterable enough to find devices with any active VPN or a specific provider.
- Device details should show the full VPN list in the network/inventory area.

This should use cached inventory data from the agent, not live Remote Tools calls from the Devices table. The Devices page should not fan out service/network commands to every endpoint just to render a column.

### Alternatives Considered

Use the existing network adapter list manually.

This technically exposes some of the raw data, but it requires opening individual devices and interpreting adapter names/IPs by hand.

Use Remote Tools > Services to check VPN services.

This is too manual, Windows-biased, and not suitable for a fleet-wide Devices column.

Infer VPNs only from IP ranges.

This is unreliable because providers can overlap with private or CGNAT ranges, and self-hosted WireGuard/OpenVPN deployments can use arbitrary address ranges.

Only show the primary IP address.

The primary IP usually shows LAN/WAN reachability, not overlay VPN reachability. VPN IPs are often the address operators actually want for private remote access.

### Component

Agent / Network Inventory / Devices UI

### Scope

Medium

This likely needs agent-side VPN detection, shared/API schema changes for normalized VPN telemetry, database storage or extension of device network inventory, API response mapping, and a new optional Devices table column.

A first pass can be read-only telemetry. It does not need to manage VPN clients, connect/disconnect VPNs, expose peer lists, or read/store VPN secrets.

ToddHebebrand:
Closes the two verified gaps @ramphex reported after testing the shipped VPN presence feature (#2185). Everything else in that feature ships as accepted.

## 1. No on/off state, on any platform

`Active: true` was hardcoded and the collector only ever emitted interfaces that were **up with an overlay IP** — so a disconnected VPN didn't show as inactive, it disappeared from the list. "Active" was a tautology.

The collector already gathers per-provider running service/process signals (`runningVPNSignals()`), but used them only for name promotion. It now also reports the known VPN clients that are **running with no tunnel up** as entries with `active: false`, no interface and no IPs, carrying the service/process signal that proves the client is there.

- Only providers with a **concrete running signal** are reported — no installed-software scan, no filesystem probing, no guesses. A client that isn't running at all is still not reported.
- Inactive entries are sorted by provider so the stored snapshot is stable across collections instead of churning on Go's random map order.
- No new collection loop and no cadence change — it rides the existing periodic network inventory.

## 2. macOS stuck on generic with several VPNs running

Generic `utun*` tunnels were promoted to a named provider only when **exactly one** known client was running (`soleVPNSignal`), so the reporter's NetBird + WireGuard Mac could never get names (count 2 → promotion refused).

Real per-interface attribution turned out to be available cheaply: `ifconfig -v` reports the owning network extension per tunnel.

```
utun6: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1280 index 24
	inet 100.95.194.59 --> 100.95.194.59 netmask 0xffffffff
	agent domain:NetworkExtension type:VPN flags:0x3 desc:"VPN: Tailscale 3"
```

The agent parses that into an interface → provider map (unprivileged, **one** exec, fetched at most once and only when a generic tunnel is actually present) and names each tunnel from it, recording `detectionSource: 'adapter'`. The single-running-client rule stays as the fallback. A description that matches zero or more than one known provider yields no attribution rather than a guess.

### Limitations, stated honestly

Only network-extension-backed clients advertise an owner. Clients that open a plain `utun` from a background daemon — **NetBird and OpenVPN among them** — expose nothing here and still depend on the sole-signal fallback. So the reporter's exact NetBird+WireGuard pairing improves but may not fully resolve: the WireGuard app tunnel gets named, a daemon-driven NetBird tunnel may still read as generic.

Because of that, when a tunnel **cannot** be attributed, the inactive pass is suppressed entirely for that device: the unattributed tunnel may well belong to one of the running clients, and reporting one of them as disconnected would be worse than saying nothing. That's the same "never guess wrong" ethos the original `soleVPNSignal` rule was written for.

## Consumer sweep

- `packages/shared` — `interfaceName` may now be empty, guarded by a refine that still **rejects an active entry without one** (an active VPN with no interface would be a phantom). Type doc updated.
- `apps/api` — ingest already passed entries through unfiltered; no `active === true` assumption existed. Covered by new route tests.
- `apps/web` — new `vpnList()` (connected first, then disconnected) backs the Devices column and the Info tab. Disconnected clients render as **muted badges** instead of the dash; the tooltip appends `disconnected`; the Info tab gains a **State** row and omits Interface when there is none. The VPN **filter** and facet options deliberately still use `activeVpnList` — "has a VPN" means a connected one. Sort key follows what the cell renders.
- i18n keys added across all 7 locales.
- `apps/docs` — the VPN Presence page documents connected vs. disconnected, the macOS attribution, and its limitation.

## Verification

- `cd agent && go test -race ./internal/collectors/...` — ok (18.2s), plus `GOOS=linux` / `GOOS=windows` cross-builds and `go vet` clean.
- New table-driven Go tests: disconnected-known-provider → `Active:false`; connected unchanged; multi-signal macOS attribution; attribution beats an ambiguous fallback; unattributed tunnel suppresses inactive claims; no signals → no phantom entries; `attrFn` fetched at most once and never for named interfaces; `ifconfig -v` parser against real macOS output.
- `@breeze/shared` validators — 69 passed; `@breeze/web` vpnProviders + DeviceList.vpn — 26 passed; `@breeze/api` inventoryNetworkVpn — 7 passed; web i18n suite — 98 passed.
- `tsc --noEmit` clean for api and web (shared has 3 pre-existing `quotes.test.ts` errors on main, untouched).

Closes #2139


ToddHebebrand:
**Review run:** `pr-review-toolkit` — code-reviewer + silent-failure-hunter, both against the full `origin/main...HEAD` diff.

**Findings:** 2 raised, 0 blocking.
- *Test gap (code-reviewer, low confidence):* the `unattributedTunnel` suppression is global rather than per-interface — correct by design, since an unnamed tunnel could belong to any of the remaining running clients, but only the single-tunnel case was covered. Addressed in `5e60e5759` with an attributed + unattributed coexistence case.
- *HIGH (silent-failure-hunter), pre-existing, not addressed here:* `updateNetworkSchema` validates `adapters` and `vpns` atomically, so any invalid VPN entry 400s the whole network-inventory PUT and drops the perfectly good adapter list with it — and the agent logs the 400 without the response body, so it is undiagnosable. This predates the PR (the unknown-provider enum can already trip it) and the agent cannot emit an entry that trips the new refine, since active entries always take their interface name from the interface. Decoupling VPN ingest from adapter ingest is a real improvement but is an agent-facing contract change well outside this issue — **filing as a follow-up rather than widening this PR.**

**Tests:** `go test -race ./internal/collectors/...` ok (18.2s); `GOOS=linux`/`GOOS=windows` cross-builds and `go vet` clean; `@breeze/shared` validators 69 passed; `@breeze/web` vpnProviders + DeviceList.vpn 26 passed; web i18n suite 98 passed; `@breeze/api` inventoryNetworkVpn 7 passed. `tsc --noEmit` clean for api and web (shared has 3 pre-existing `quotes.test.ts` errors on main, untouched). CI: 55/55 required checks green.

**Field caveat for @ramphex's setup:** gap 1 (on/off state) is fixed on all three platforms. Gap 2 is fixed for network-extension-backed macOS clients — the WireGuard app tunnel now names itself even with NetBird running alongside — but NetBird's macOS client opens a plain `utun` from its daemon and still exposes no owner, so that tunnel may continue to read as generic. Worth a confirmation run on the reporter's Mac before this issue is closed.

PR #3556, opened 2026-08-15
harden(api): guard shared patch catalog rows against agent metadata drift

ToddHebebrand:
## What

Agent patch-scan ingest (`apps/api/src/routes/agents/patches.ts`) upserts into `patches` — a deliberately **global, un-tenanted** catalog table deduped on `(source, external_id)`. Both the pending and installed upserts rewrote every metadata column on conflict, so any device reporting a colliding key replaced the metadata that every tenant's UI, `patch_approvals` and patch jobs read from that shared row.

Inconsistent or untrusted agent-reported metadata could therefore overwrite an existing shared catalog row's identity. That matters most for `package_id`, which is not decorative:

- `routes/devices/patches.ts` forwards `patches.package_id` verbatim into the `install_patches` command payload, and the agent resolves *which package to install* from it (`resolvePatchInstallID` / `patchLocalID`).
- `packageId` + `version` drive the app **block/pin** rules in `patchApprovalEvaluator`.

No new columns and no migration. Everything stays inside the existing single upsert — this is hot-path ingest (up to 5000 rows per request), so there are no added per-row queries.

## The classification

| Class | Columns | Rule |
|---|---|---|
| Identity | `title`, `vendor`, `package_id`, `version` | A raw agent report may **fill** these, never rewrite them. Overwriting is reserved for server-side catalog enrichment (`matchedCatalogId != null`). |
| Operational | `severity`, `category`, `description`, `requires_reboot`, `os_types` | Keep refreshing — that flow is the whole point of a rescan — but no longer **null-downgrade** when a scan simply omits a field. |

Two specifics worth calling out:

- **`severity: 'unknown'`** is the agent's "I couldn't classify this", not an assessment. It no longer resets a severity an earlier scan — or `cveEnrichmentWorker`, which only ever *raises* severity — established.
- **The installed path is the lowest-authority writer.** It carries no enrichment, and its `version` is the *installed* version — a semantically different value landing in the same column. It previously dragged the shared row's available version backwards on every scan cycle (visible in the pre-existing `#2725` integration case: pending reports `2.55.0.3`, the paired installed submit reported `2.51.0.2` and overwrote it). It may now only fill.

## Batch-abort hardening

Three ways one malformed row could abort a whole device's scan transaction, plus narrow structural admission checks on the command-bearing identifiers:

- Reject control characters, option-like identifiers and over-length values in `external_id` / `package_id`. Checked **per colon-separated segment**, because the agent splits provider-prefixed ids and installs the local part — `apple-softwareupdate:--all` passes a naive leading-dash test. **Spaces stay legal**: Apple `softwareupdate` labels (`macOS Sonoma 14.5-23F79`) contain them and are used verbatim as both ids. There is an explicit regression test for this.
- `sanitizeDate` accepted impossible calendar dates (`2026-02-31`) — JS `Date` rolls them over silently and Postgres then raised `22008`.
- `download_size_mb` is clamped instead of overflowing the `integer` column.
- Over-length `version` / `category` become `NULL` rather than a truncated value that would compare or match as something it isn't. `title` / `vendor` / `description` are still truncated — losing the row would be worse.

Malformed rows are **skipped and counted** (response `rejected`, plus a bounded reason histogram in the audit event) rather than failing the batch. Admission runs **before** the tombstone sweep, and a scan with refused rows skips the sweep entirely — otherwise it would tombstone a patch it declined to re-ingest. Same conservatism as the covered-zero-sources branch from #2217.

## Provider compatibility

winget / chocolatey / apt / yum / homebrew / apple / windows-update dedup behavior is unchanged — the conflict target and the derived `external_id` precedence (explicit → `kbNumber` → composed) are untouched.

## Tests

- `patchIngestIdentity.test.ts` (new, 25) — pure admission/normalization: derivation precedence, the Apple-spaces regression guard, per-segment option-injection, the truncate-vs-null length policy, and the batch histogram.
- `helpers.patchSource.test.ts` — `sanitizeDate` calendar-rollover regression.
- `patches.integration.test.ts` (new cases, **real Postgres**) — this is the load-bearing evidence, since the rules are raw `COALESCE`/`CASE` fragments inside Drizzle's `onConflictDoUpdate` that a mocked suite can only inspect as an object shape:
  - identity columns survive a conflicting second report while operational columns still refresh;
  - a sparser rescan does not blank description/category/reboot/severity;
  - a later scan still **fills** columns the first report left empty;
  - installed inventory cannot drag the row back to the installed version (while `device_patches.installed_version` still records it);
  - a malformed row is refused, the rest of the batch lands, and the sweep is skipped;
  - Apple space-bearing labels still round-trip.

`apps/api` affected suites green (45 files / 769 tests), `patches.integration.test.ts` 7/7 against real Postgres, `tsc --noEmit` clean.

## Follow-ups (deliberately not in this PR)

1. **Third-party catalog enrichment appears not to match in the field.** The agent decorates ids with the provider prefix (`winget:Mozilla.Firefox`) while `third_party_package_catalog` stores the bare id (`Mozilla.Firefox`), and `enrichFromCatalog` does an exact `source::packageId` lookup. If that's right, curated enrichment rarely fires for winget today. This PR is forward-compatible either way — curated overwrite simply starts applying once the lookup matches — but the mismatch deserves its own investigation.
2. **Microsoft rows store a per-device WUA UpdateID in a global column.** `external_id` collapses on the KB while `package_id` can vary by device/revision, so neither last-writer-wins (before) nor first-writer-wins (now) is *right* — this change makes it stable rather than arbitrary. Preferring the KB for WUA installs would fix it properly.
3. **`installed < pending < curated` needs stored provenance**, not `COALESCE`. A row first created by the installed path keeps its unenriched title until a curated report arrives; a `metadata_authority` column would express this correctly.
4. **The available version is arguably a per-device observation** and would be better modelled on `device_patches` than as global state.
5. `release_date` is already first-writer-wins (it is in the insert but not the conflict update); it anchors auto-approval deferral windows and is worth revisiting alongside 3/4.

### DIFF
--- a/apps/api/src/routes/agents/helpers.patchSource.test.ts
+++ b/apps/api/src/routes/agents/helpers.patchSource.test.ts
@@ -85,7 +85,7 @@ vi.mock('../../jobs/softwareComplianceWorker', () => ({
 }));
 vi.mock('./policyProbeSafety', () => ({ isAllowedPolicyConfigProbe: vi.fn(() => true) }));
 
-import { buildPatchSourceConfigUpdate } from './helpers';
+import { buildPatchSourceConfigUpdate, sanitizeDate } from './helpers';
 
 const DEVICE_ID = '00000000-0000-4000-8000-000000000001';
 
@@ -125,5 +125,35 @@ describe('buildPatchSourceConfigUpdate (#1872)', () => {
     const result = await buildPatchSourceConfigUpdate(DEVICE_ID);
 
     expect(result).toEqual({ exclusiveWindowsUpdate: false });
+  });
+});
+
+describe('sanitizeDate', () => {
+  it('round-trips a valid date', () => {
+    expect(sanitizeDate('2026-01-05')).toBe('2026-01-05');
+  });
+
+  it('rejects an impossible calendar date instead of letting it roll over (regression)', () => {
+    // `new Date('2026-02-31')` silently rolls over to 2026-03-03; the old code
+    // returned the original string, which then hit Postgres 22008 and aborted
+    // the whole ingest transaction.
+    expect(sanitizeDate('2026-02-31')).toBeNull();
+  });
+
+  it('rejects a month 13 date', () => {
+    expect(sanitizeDate('2026-13-01')).toBeNull();
+  });
+
+  it('rejects Feb 29 in a non-leap year', () => {
+    expect(sanitizeDate('2025-02-29')).toBeNull();
+  });
+
+  it('rejects non-matching shapes', () => {
+    expect(sanitizeDate('01/05/2026')).toBeNull();
+    expect(sanitizeDate('')).toBeNull();
+    expect(sanitizeDate(12345)).toBeNull();
+    expect(sanitizeDate(null)).toBeNull();
+    expect(sanitizeDate(undefined)).toBeNull();
+    expect(sanitizeDate('2026-1-5')).toBeNull();
   });
 });
--- a/apps/api/src/routes/agents/helpers.ts
+++ b/apps/api/src/routes/agents/helpers.ts
@@ -130,7 +130,13 @@ export function sanitizeDate(value: unknown): string | null {
   const trimmed = value.trim();
   if (!/^\d{4}-\d{2}-\d{2}$/.test(trimmed)) return null;
   const d = new Date(trimmed + 'T00:00:00Z');
-  return Number.isNaN(d.getTime()) ? null : trimmed;
+  if (Number.isNaN(d.getTime())) return null;
+  // `new Date()` silently rolls impossible calendar dates over ('2026-02-31'
+  // becomes 2026-03-03) and is therefore NOT a validity check on its own. The
+  // original string would then reach a Postgres `date` column and raise 22008,
+  // aborting the entire enclosing ingest transaction. Round-trip to confirm the
+  // date the caller wrote is the date we parsed.
+  return d.toISOString().slice(0, 10) === trimmed ? trimmed : null;
 }
 
 /**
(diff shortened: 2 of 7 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — code-reviewer (correctness/SQL/compat) + pr-test-analyzer (coverage/vacuousness), plus an independent read-only design review from Codex before implementation.

**Findings:** 3 raised → all 3 addressed in `bfb1b8b64`; 0 outstanding.
- Combined `PUT /:id/patches` gated its sweep on the pending admission only, but that route's sweep (`markAllDevicePatchesMissing`) also flips *installed* rows — a refused installed row left a genuinely-installed patch stranded at `missing`. Now gated on both admissions.
- The two rejection histograms were merged with object spread, dropping one side's count whenever both lists refused rows for the same reason. Now summed per reason.
- Missing coverage: the combined-route installed-side sweep suppression, the `download_size_mb` clamp, and the audit histogram. All three added.

The design review also shaped the original commit: it caught that `severity: 'unknown'` still clobbered a known severity, that admission had to run *before* the sweep, that the option-like check must apply per colon-separated segment, and that over-length decision fields (`version`/`category`) should become NULL rather than a truncated value.

**Vacuousness check:** the analyzer mutation-tested the mocked-Drizzle `set` assertions and the new real-Postgres cases — each fails if its guard is reverted. One pre-existing test (`persists agent-supplied version into patches.version`) covers a branch this PR didn't change; kept as regression coverage.

**Tests:** `apps/api/src/routes/agents/` 44 files / 766 tests green; `patches.integration.test.ts` 8/8 against real Postgres; `tsc --noEmit` clean.

**Status:** review-clean. CI on this repo is heavily queued right now — required checks (Test API, Type Check, Integration Tests) were still pending at hand-off and should be confirmed green before merge.

PR #3558, opened 2026-08-15
feat(agent,api,web): dual-context winget scan surfaces per-user app updates (#2727)

### LINKED ISSUE #2727: [Agent] SYSTEM winget scan forces --scope machine — per-user app installs are invisible to third-party patching
## Description

Found during the third-party patch scan investigation (root fix: #2725).

`SystemWingetProvider` runs `winget upgrade --include-unknown --scope machine --source winget ...` (`agent/internal/patching/winget_system.go:32-35`). Many common apps (Chrome, Zoom, Slack, Discord, etc.) default to per-user installs, which a machine-scope query running as SYSTEM cannot see. On a typical workstation this makes the third-party scan silently under-report — often down to zero — while the command still completes successfully.

## Proposed Fix

Needs design: SYSTEM can't trivially enumerate per-user installs. Options include dropping `--scope machine` and filtering results, running an additional scan in the logged-in user's context via the user helper (the pre-#2216 approach), or merging both scopes. Whatever the choice, per-user upgradable apps should either appear in results or be explicitly reported as out of scope in the UI.

## Affected Files

- `agent/internal/patching/winget_system.go` (scan args)
- `agent/internal/heartbeat/winget_register_windows.go`

ToddHebebrand:
Closes #2727.

## The bug

The agent runs as SYSTEM and scanned winget at machine scope only. SYSTEM's `HKCU` is `HKU\S-1-5-18`, which contains none of the logged-in user's ARP entries — and Chrome, Zoom, Slack, Discord, Teams (personal) and most of the everyday app surface default to **per-user** installs. So a typical workstation reported **zero** third-party updates, with a successful exit code. Nothing looked wrong.

Dropping `--scope machine` does not fix it: those packages are not in the process's view at any scope.

## The fix — dual-context detection (issue Option 2, the Winget-AutoUpdate pattern)

`SystemWingetProvider` keeps provider ID `winget` and now runs a second, best-effort pass — `winget upgrade --scope user` inside the interactive user's session, over the **existing** `run_as_user` exec transport (broker → `Session.SendCommand` → helper `executeProcess`; only the winget-specific wrapper was ever deleted, in `d61ffe57a`). Results are merged and deduped by package ID, machine entry winning.

**It is deliberately not a second provider.** Scan coverage is computed per source *bucket* and a bucket only counts as covered when every provider feeding it succeeded (`heartbeat.coveredPatchSources`). A `winget-user` provider would map to the shared `third_party` bucket and be skipped on every machine with nobody logged in — so the API would stop reconciling third-party patches for the entire unattended fleet. Keeping the pass internal makes coverage byte-identical to before: `winget` is covered iff the SYSTEM pass parsed. There's a regression test pinning exactly that.

`--scope user` rather than unscoped-and-filter: the helper runs on the user's filtered token and can read HKLM, so an unscoped run returns machine packages too — and `winget upgrade` has no scope column, so nothing in the output would let us label the rows. Asking for user scope makes the label truthful by construction.

## Degradation

Every user-pass failure — no helper configured, nobody logged in, IPC error, winget error, unparsable output, timeout — returns **exactly** the pre-change machine-scope result, and is recorded rather than swallowed. The machine pass is unchanged and still authoritative; if it fails, the user pass is never attempted.

## Reporting the gap instead of under-reporting

Two new signals so "no per-user updates" is never confused with "nobody looked":

- **`device_patches.scope`** (`machine` | `user`; NULL = provider has no scope concept). New nullable column + CHECK + partial index; registered in `CORE_TENANT_EXPORT_POLICY` (the contract that fires on a new *column*).
- **`userScopeScanned`** on the pending upload — a second coverage axis. The ingest route only sweeps user-scope rows to `missing` when this is explicitly `true`; otherwise a scan taken while nobody was logged in would tombstone rows it never inspected. That is #2217, one axis down. A scope-less re-report also cannot blank an established scope.

The device patch tab badges per-user rows and shows a subtle note when the per-user pass didn't run (i18n keys added across all 7 locales).

## Scope of this PR: detection only

Installing a package the last scan saw **only** at user scope is refused with an explicit error rather than silently running a machine-scope install, which would either fail confusingly or install a second machine-wide copy alongside the user's. The refusal survives a later failed user pass (a logged-out user must not silently re-open the machine-install path).

**Follow-ups, deliberately not in this PR:**
- User-context install/upgrade routing for per-user packages.
- Per-SID identity on rows — coverage is the console session only, so a multi-user box reports whichever user was at the console.
- Unknown-scope packages: `--scope` filters on the installed scope winget recorded, and packages it cannot classify are dropped by **both** passes. Pre-existing winget under-report, affects machine scope today.
- The helper's `executeProcess` waits unbounded on `<-done` after killing a timed-out child, and buffers output unbounded. Pre-existing; this pass widens exposure to it.

## Verification

- `go test -race ./internal/patching/... ./internal/heartbeat/...` — green. Table-driven coverage for the merge/dedupe (incl. case-insensitive IDs), scope labelling, every degradation branch, the install refusal + its persistence, the timeout bound, the IPC contract shape, and the single-provider/coverage-unchanged guards.
- `GOOS=windows|linux|darwin go build ./...` — green.
- `apps/api`: `patches.test.ts`, `schemas.test.ts`, `helpers.patchSource.test.ts`, `autoMigrate.test.ts` green; integration suite on a **private** test stack (fresh DB, migration replayed from zero): `patches.integration.test.ts` (incl. two new #2727 cases), `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade`, RLS coverage — all green. `db:check-drift` clean.
- `apps/web`: all 55 device component suites + the five i18n gates green.
- `tsc --noEmit` clean for both `apps/api` and `apps/web`.

## Spec

`docs/superpowers/specs/vuln-patch/2026-06-30-windows-3rd-party-patching-system-context-design.md` recorded machine-scope-only as accepted. That decision is now marked reversed inline, with an amendment section at the end recording why it failed in the field, what replaced it, and the limitations that remain.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer. An independent read-only design review (Codex, `xhigh`) also ran on the core call before implementation and confirmed the one-provider/two-passes shape plus the coverage-trap reasoning.

**Findings:** 5 raised → all addressed in `16cc0c1`; 0 outstanding.

- **The UI signal would have read `null` forever.** `devices/patches.ts` read `userScopeScanned` straight off the stored command result, but the agent's payload arrives wrapped — `tools.NewSuccessResult` JSON-*encodes* the handler map into `stdout` of the `{status, exitCode, stdout}` envelope. Now goes through `safeParsePatchResult`, the unwrap this file already uses for `pendingCount`. Found by chasing the test-analyzer's coverage gap into the actual storage path; ironically the same silent under-report class this PR exists to fix.
- **Nothing proved a NULL-scope row still gets swept** (test-analyzer). This is the whole reason the guard is `IS DISTINCT FROM 'user'` rather than `<> 'user'` — under `<>`, `NULL <> 'user'` is NULL, every legacy row drops out of the WHERE clause and nothing is ever swept again, fleet-wide. Now covered against real Postgres.
- **The install guard reopened on a failed rescan** (Codex). A user-pass failure called `recordUserScan` with machine-only results, clearing the remembered user-scope IDs — so a user logging out silently re-enabled the machine-scope install path for their per-user packages. Split into `recordUserScanStatus`; regression test added.
- **Per-user rows were visible but Install was not gated** (code-reviewer). A tech could queue a job the agent is guaranteed to refuse. `isPatchInstallable` now gates the per-row button and the bulk action.
- **The per-user badge was in the wrong table.** It was wired into the native-OS pending table, but per-user winget packages are `third_party` and land in the other one — so it never rendered for the rows it was written for.

Also strengthened one weak assertion the analyzer flagged: the sweep-guard unit test asserted only that the string `IS DISTINCT FROM` was present, not what it compared against, so flipping the literal to `'machine'` would have passed it.

**Tests:** agent `go test -race ./internal/patching/... ./internal/heartbeat/...` green; `GOOS=windows|linux|darwin go build ./...` green. API unit + integration suites green on a **private** test stack (fresh DB, migrations replayed from zero) — including `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade` and RLS coverage; `db:check-drift` clean. Web: 55 device suites + 5 i18n gates green. `tsc --noEmit` clean for `apps/api` and `apps/web`.

**CI:** all 57 checks pass. One earlier red was `InvoiceWorkspace.test.tsx` (billing, untouched here, timed out at 5s); it passes locally and its last commit was literally *"give the InvoiceWorkspace queued-Issue tests real waitFor headroom"* — re-ran and it went green.

**Worth knowing before merging:** this ships **detection only**. Per-user apps become visible and are labelled, but they are not remediable — install is explicitly refused agent-side and disabled in the UI. Coverage is also the console session only, so a multi-user box reports whichever user was at the console. Both are stated in the spec amendment rather than left implicit; the remediation path and per-SID identity are the follow-ups.

Issue #3552, opened 2026-08-14
[API] manage_software_policies is Tier 1 (auto-execute) but can set enforceMode/autoUninstall — should be Tier 3 like its sibling


ToddHebebrand:
## Description

Found during the #3543 fix (PR #3548), deliberately not changed there because re-tiering alters existing workflows.

`manage_software_policies` in `apps/api/src/services/aiToolsPolicyPrereqs.ts` is registered **Tier 1 — auto-execute, no approval** — yet its `create`/`update` actions accept `enforceMode` and `remediationOptions` (including `autoUninstall`). The equivalent software-policy tool in `aiToolsCompliance.ts` is Tier 3 (approval-gated). So the AI agent can arm fleet-wide auto-uninstall enforcement on a software policy without any human approval step, via the prereq-tool path.

As of #3548 this path is at least fully **audited** (both stores, actor `ai`), but the tier looks like a plain misclassification: mutating enforcement state is exactly what the tier system exists to gate.

Context that raises the stakes: in the #3381 incident (259-device mass uninstall), the complete `audit_logs` trail shows no arming event. Before #3548, this Tier 1 tool was one of the unaudited paths that could have armed enforcement invisibly and without approval — making it a prime suspect for how a detect-only allowlist became enforcing.

## Proposed Fix

- Register `manage_software_policies` `create`/`update` as Tier 3 (or split: reads/list stay Tier 1, mutations Tier 3), matching the sibling tool.
- Audit the rest of the prereq-tool registry (`aiToolsPolicyPrereqs.ts`) for the same pattern — other `manage_*` tools that mutate enforcement-relevant state at Tier 1.
- A regression test asserting that no Tier 1 tool schema accepts enforcement-arming fields.

## Affected Files

- `apps/api/src/services/aiToolsPolicyPrereqs.ts` (primary)
- AI tool tier registry / gating tests

## Reported By

PR review during #3548; relates to #3543 and #3381.

ToddHebebrand:
[...]

PR #3561, opened 2026-08-15
fix(api): gate enforcement-arming AI policy tools behind Tier 3 approval (#3552)

ToddHebebrand:
## ⚠️ Behavior change — AI policy mutations now require approval

**This is the point of the PR, and it will be visible to users immediately.** Four AI tool operations that previously auto-executed now raise a Tier 3 `supervised` approval request (the requesting human approves their own AI action with one click, gated on their existing RBAC — no second approver):

| Tool | Actions now gated | Was |
|---|---|---|
| `manage_software_policies` | `create`, `update` | Tier 2 (auto-execute + audit) |
| `manage_update_rings` | `create`, `update` | Tier 2 |
| `manage_peripheral_policies` | `create`, `update` | Tier 2 |
| `manage_patches` | `setup_auto_approval` | Tier 1 (handler already disabled) |

**`list` / `get` on all of these still auto-execute at Tier 1** — no approval fatigue on reads. Only the mutations changed. Any AI workflow that creates or edits a software policy, update ring, or peripheral policy now pauses for one approval click instead of applying silently.

## The bug

`manage_software_policies` was registered base **Tier 1** with `create`/`update` escalated only to **Tier 2** — auto-execute plus an audit row, *no approval* — while its input schema accepts `enforceMode` and `remediationOptions.autoUninstall`. So the AI agent could flip a detect-only software allowlist into **fleet-wide auto-uninstall with no human in the loop**.

The singular-named sibling writing the **same `software_policies` table** (`manage_software_policy`, `aiToolsCompliance.ts`) has always been base Tier 3. The prereq tool is the *reachable* one: it — not the singular — is listed in `aiAgentSdkTools`' `TOOL_TIERS`, and `BREEZE_MCP_TOOL_NAMES` is derived from that map, so the plural tool is what chat and MCP actually call. The Tier 3 sibling is effectively unreachable from those surfaces. The gate that looked like it existed did not apply to the path in use.

Context from #3552: in the #3381 incident (259-device mass uninstall) the `audit_logs` trail shows no arming event. Before #3548 this path was both unaudited *and* unapproved.

## Sweep of the rest of `aiToolsPolicyPrereqs.ts` (scope item 2)

All five prereq tools were base Tier 1 with mutations at Tier 2. Findings:

**Re-tiered — clear-cut analogues (arm unattended action on real endpoints):**
- **`manage_update_rings`** — `autoApprove` + `deadlineDays` + `gracePeriodHours` arm unattended patch installs with **forced reboots**. This is the standing-rule form of `manage_patches:install`, which already requires approval; approving each patch but not the rule that auto-approves all of them was the same inversion as the reported bug.
- **`manage_peripheral_policies`** — `action_type: block | read_only` cuts off USB / Bluetooth / Thunderbolt across the target scope. Its singular sibling `manage_peripheral_policy` (`aiToolsPeripherals.ts`) is already base Tier 3 — the identical plural/singular tier gap.
- **`manage_patches:setup_auto_approval`** — not in `aiToolsPolicyPrereqs.ts`; surfaced by the new structural test. It arms auto-approval + reboot policy at **Tier 1**. Its handler is hard-disabled today ("managed through configuration policies"), so gating it costs no live workflow — it is here so re-enabling the handler cannot silently reopen the hole.

**Judgment calls — documented, left at Tier 2:**
- **`manage_backup_configs`** — changes storage destinations, credentials, retention on live backup config. Real impact, but *protective scheduling*, not enforcement or remediation: no payload causes an agent to uninstall, block, or reboot anything.
- **`manage_backup_profiles`** — selection sets only; `delete` is already refused while the profile is referenced.

Escalating the backup pair is defensible under a broader "any live linked policy change needs approval" rule, but that is a separate product decision. Both are pinned by an explicit test so a later decision to escalate is a conscious edit, not drift. The rationale is recorded inline in `TIER3_ACTIONS`.

## Why `supervised`, not `four_eyes`

Matches `manage_configuration_policy` `create`/`update`/`delete` (the policy these prereqs link into via `featurePolicyId`), `manage_patches:install`, and the Tier-3 singular siblings in `TIER3_SUPERVISED_TOOLS`. Spec `2026-08-05-tier3-supervised-four-eyes-split-design.md` §3.2 reserves `four_eyes` for externally-binding, identity-control, and destroy/rewind actions; blast-radius-based escalation is explicitly deferred there.

## How the gate is applied

The tier machinery is already per-`(tool, action)`: `checkGuardrails` overlays `TIER1/2/3_ACTIONS` on the tool's base tier. So the fix is table edits only — the tools **keep base Tier 1**, which is what preserves auto-execute on `list`/`get` and keeps them visible over MCP (a base-tier-3 tool would gate reads too). Both surfaces are covered by construction: the legacy registry path and the Agent SDK path share `checkGuardrails` as the single gate, and MCP annotates from the same resolution. `TOOL_PERMISSIONS` RBAC entries already existed for all three tools; no registry, `TOOL_TIERS`, or MCP change was needed. The three tools are deliberately **not** added to `TIER3_SUPERVISED_TOOLS` — that whole-tool set requires base Tier 3 members.

`CLASSIFICATION_VERSION` bumped 1 → 2: these pairs produce approval intents for the first time, so their absence before v2 is a tier gap, not a gap in the record.

## Regression test (scope item 3)

New `aiGuardrails.enforcementArming.contract.test.ts` — **structural, not a name list**, so a future tool cannot reintroduce the gap. It sweeps every registered core tool across **both** schema surfaces (the Anthropic definition the model sees and the Zod schema `validateToolInput` enforces) for arming fields — `enforceMode`, `remediationOptions`, `autoUninstall`, `autoApprove` — including nested keys that exist only in prose (`remediationOptions.autoUninstall` lives inside a free-form `type: 'object'` property with no nested schema to walk). Every non-read action that can carry one must resolve to Tier 3.

It **fails closed on action naming**: only an explicit read allowlist (`list`/`get`/`search`/…) is exempt, so a future action called `arm` or `enforce` must be gated or CI goes red. Action narrowing (the `"(for <action>)"` description convention that keeps `manage_patches`' `autoApprove` from fanning out over `approve`/`decline`/`defer`) only accepts **real enum members**, so a stale or misspelled name in a description cannot silently shrink the sweep.

Three anti-vacuity guards: the sweep asserts it finds arming tools at all, that the nested-key detection path still resolves `autoUninstall`, and that narrowing lands on exactly `setup_auto_approval`. **Verified non-vacuous by reverting the fix — 5 tests go red**, including the exact #3552 payload (`enforceMode: true` + `remediationOptions.autoUninstall: true`).

## Verification

- `aiGuardrails.enforcementArming.contract.test.ts` — 13 passed (new)
- `aiGuardrails.{test,approvalScope.contract,readonly.contract}.test.ts` — 180 passed
- Full AI + action-intent surface: **137 files / 2068 tests passed** (`src/services/ai*.test.ts`, `src/services/actionIntents/*.test.ts`, intent reaper jobs), single-fork
- `tsc --noEmit` clean
- Design quorum run before implementing (independent second opinion agreed on scope, `supervised`, the three re-tiers, and leaving the backup pair)

Refs #3552. Relates to #3543, #3548, #3381. Not auto-closing — leaving that to the owner.

### DIFF
--- a/apps/api/src/services/actionIntents/intentService.test.ts
+++ b/apps/api/src/services/actionIntents/intentService.test.ts
@@ -304,7 +304,7 @@ function makeIntentRow(overrides?: Record<string, unknown>) {
     // 0) — tests exercising the tier3-supervised-four-eyes split override
     // these explicitly (see createIntentWith / echoInsertedIntent below).
     approvalScope: 'four_eyes',
-    classificationVersion: 1,
+    classificationVersion: 2,
     effectDigest: null,
     status: 'pending_approval',
     createdAt: new Date(),
@@ -753,7 +753,7 @@ describe('createActionIntent — supervised/four_eyes scope', () => {
 
     const captured = dbState.insertedActionIntentValues[0];
     expect(captured?.approvalScope).toBe('supervised');
-    expect(captured?.classificationVersion).toBe(1);
+    expect(captured?.classificationVersion).toBe(2);
     // Dual-write compat: legacy `expiresAt` still carries the same value as
     // the new `approvalExpiresAt` column (Plan 3 removes the legacy write).
     expect(captured?.expiresAt).toEqual(captured?.approvalExpiresAt);
--- a/apps/api/src/services/actionIntents/intentService.ts
+++ b/apps/api/src/services/actionIntents/intentService.ts
@@ -165,8 +165,15 @@ export const RELEASE_LEASE_MS = 10 * 60 * 1000;
  * (Task 2) so a future ruleset change can be told apart from intents
  * classified under an older version. Bump when the classification logic
  * changes in a materially observable way.
+ *
+ * v2 (#3552, 2026-08-14): the policy-prerequisite mutators
+ * manage_update_rings / manage_software_policies / manage_peripheral_policies
+ * create+update moved from Tier 2 (auto-execute) to Tier 3 `supervised`, so
+ * they produce approval intents for the first time. Intents for those pairs
+ * only exist from v2 onward; their absence before v2 is a tier gap, not a
+ * gap in the record.
  */
-export const CLASSIFICATION_VERSION = 1;
+export const CLASSIFICATION_VERSION = 2;
 
 const MAX_ARG_VALUE_LEN = 80;
 
--- a/apps/api/src/services/aiAgentSdk.enforcementArmingGate.test.ts
+++ b/apps/api/src/services/aiAgentSdk.enforcementArmingGate.test.ts
@@ -1,0 +1,376 @@
+/**
+ * Contract test: proves the #3552 enforcement-arming gate is actually WIRED,
+ * not just declared.
+ *
+ * PR #3561 re-tiered `manage_software_policies`, `manage_update_rings`,
+ * `manage_peripheral_policies` (create/update) and
+ * `manage_patches:setup_auto_approval` from auto-execute to Tier 3
+ * `supervised` in aiGuardrails.ts's TIER3_ACTIONS. That half of the contract
+ * is covered by aiGuardrails.enforcementArming.contract.test.ts, which calls
+ * checkGuardrails() directly and asserts tier === 3.
+ *
+ * The OTHER half — that createSessionPreToolUse actually consults the real
+ * checkGuardrails() and, on a tier-3 result, creates a durable action intent
+ * and blocks execution pending a decision — is proven separately in
+ * aiAgentSdk.test.ts, but that file `vi.mock('./aiGuardrails')`s the whole
+ * module, so the two halves are never joined for these specific tool names.
+ * A regression that dropped these tools back out of TIER3_ACTIONS would still
+ * pass every existing test.
+ *
+ * This file closes that gap: it exercises createSessionPreToolUse with the
+ * REAL aiGuardrails.checkGuardrails (deliberately NOT mocked — mocking it
+ * here would make the file pointless), for exactly the tool/action pairs
+ * #3552 was about, and confirms:
+ *   - auto_approve mode — the STRONGEST auto-execute setting — does not
+ *     bypass the gate: createActionIntent is called, and a rejected decision
+ *     blocks execution.
+ *   - the exact issue-#3552 payload (enforceMode + remediationOptions.
+ *     autoUninstall) is gated.
+ *   - manage_patches:setup_auto_approval is gated.
+ *   - reads (list/get) on the same tools still auto-execute — no approval
+ *     fatigue regression.
+ *
+ * checkToolPermission/checkToolRateLimit ARE spied (narrowly, via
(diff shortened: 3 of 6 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (tier-table correctness, call-site sweep, CLASSIFICATION_VERSION consumers, conventions) + `pr-test-analyzer` (coverage, vacuity, stale tests). A design quorum with an independent second opinion ran *before* implementation on the classification itself (scope of the re-tier, `supervised` vs `four_eyes`, which prereq tools to leave alone) — both positions agreed.

**Findings:** 3 raised → all 3 addressed in `ca5421484`; 0 outstanding.
1. *End-to-end gap (sev 6)* — the tier change was proven in two disconnected halves: the contract test showed the real registry resolves these tools to Tier 3, and `aiAgentSdk.test.ts` showed a Tier 3 result creates an intent and blocks — but that suite `vi.mock`s `./aiGuardrails` wholesale and only exercises unrelated tool names. A tool-name mismatch between the guardrail tables and the SDK dispatch key would have passed both. Fixed by a new `aiAgentSdk.enforcementArmingGate.test.ts` that runs the **real** `checkGuardrails` (only `checkToolPermission`/`checkToolRateLimit` narrowly spied) and drives `createSessionPreToolUse` under `approvalMode: 'auto_approve'` — the strongest auto-execute setting — asserting each re-tiered mutation still creates a durable action intent and refuses to execute.
2. *`manage_peripheral_policies` was inside the fix but outside the structural sweep* — its arming input is `action_type: block | read_only`, which wasn't in `ARMING_FIELDS`, so only name-pinning protected it. Added (`action_type` is unique to that tool repo-wide).
3. *Narrowing could under-cover* — the description-based action narrowing treated `"(required for create)"` as an applicability clause and would have dropped `update` from the sweep, i.e. stopped covering the action that flips an **existing** policy to block. Tightened to only honour a parenthetical opening with `"for"`, and pinned by a test.

No stale tests found: every existing suite touching these tool names calls handlers directly and is tier-blind by construction.

**Anti-vacuity:** both new suites were mutation-tested by reverting the `TIER3_ACTIONS` entries — contract test **5 red**, SDK gate test **7 of 14 red**. Re-verified independently of the agent that wrote them; `aiGuardrails.ts` confirmed byte-identical to its committed state afterward.

**Tests:** `aiGuardrails.enforcementArming.contract.test.ts` 15 passed (new) · `aiAgentSdk.enforcementArmingGate.test.ts` 14 passed (new) · `aiAgentSdk.test.ts` 110 passed · full AI + action-intent surface 137 files / 2068 tests passed, single-fork · `tsc --noEmit` clean.

**Status:** review-clean, awaiting maintainer merge. Note the behavior change in the PR body — AI policy mutations now pause for one approval click; reads are unchanged. Not auto-closing #3552.

PR #3563, opened 2026-08-15
fix(web): let fleet (All-orgs) view create DNS integrations & device groups

### LINKED ISSUE #3505: DNS Security 'Add Integration' never sends orgId, always 400s for partner/system-scoped sessions
Reported by **odobenusrosmarus_** in #bugs (Discord, message 1537571055768436888).

**Report (PII-redacted):**

When attempting to add DNS Integration under DNS Security, upon selecting any provider and filling out the information you're presented with an error, orgId is required for this scope. This happens within both full customer view or a specific customer view.

{"error":"orgId is required for this scope"}

Breeze Version 0.105.0

## Triage analysis

Reproduced the cause in code. Creating a DNS Security integration POSTs to /dns-security/integrations, and the server's resolveOrgId() (apps/api/src/routes/dnsSecurity.ts) requires an explicit `orgId` in the JSON body for any caller whose auth scope isn't a single-org token and who doesn't have exactly one accessible org (i.e. any partner/system-scoped session viewing 'All organizations' or a specific customer via the dashboard's org switcher rather than an org-scoped login) — otherwise it 400s with exactly 'orgId is required for this scope'. The DNS Security add-integration UI (apps/web/src/components/dnsSecurity/AddDnsIntegrationModal.tsx, invoked from DnsSecurityIntegrationsTab.tsx) never includes `orgId` in its POST body at all — it only sends provider/name/apiKey/apiSecret/config/isActive. fetchWithAuth does auto-append `?orgId=` to the URL from the org store, but that's a query-string param the route never reads (it only validates c.req.valid('json')), so it has no effect on this POST. Other integration flows in the same app (e.g. WebhooksPage.tsx line ~250-253) explicitly do `{ ...payload, orgId: currentOrgId }` on create for exactly this reason, showing the expected pattern DNS Security is missing. This matches the reporter's exact error and their observation that it happens in both 'full customer view' and 'a specific customer view' — the modal never sends orgId either way, so any session that isn't a strictly single-org token will 400.

## Affected code

- `apps/api/src/routes/dnsSecurity.ts:127-158` — resolveOrgId: requires body.orgId when auth.scope isn't 'organization' and accessibleOrgIds.length !== 1; falls through to 'orgId is required for this scope'
- `apps/api/src/routes/dnsSecurity.ts:347-360` — POST /integrations calls resolveOrgId(auth, body.orgId) — body.orgId comes only from zValidator('json', createIntegrationSchema)
- `apps/web/src/components/dnsSecurity/AddDnsIntegrationModal.tsx:124-137` — POST body omits orgId entirely — provider/name/apiKey/apiSecret/config/isActive only
- `apps/web/src/components/dnsSecurity/DnsSecurityIntegrationsTab.tsx:1-49,216-221` — Tab component takes no orgId prop and never threads current org selection into the modal
- `apps/web/src/stores/auth.ts:519-526` — fetchWithAuth only appends orgId as a URL query param; the POST route here validates JSON body only, so this injection is a no-op for this endpoint
- `apps/web/src/components/webhooks/WebhooksPage.tsx:250-253` — Comparable integration-creation flow explicitly adds orgId: currentOrgId to the create payload — the pattern DNS Security's modal is missing

## Potential fix

Thread the current org selection (e.g. via useOrgScope()/currentOrgId) into DnsSecurityIntegrationsTab and pass it to AddDnsIntegrationModal, then include `orgId: currentOrgId` in the POST body when creating an integration — mirroring WebhooksPage.tsx's handleSubmit pattern.

### LINKED ISSUE #3506: Create Device Group fails for multi-org partner users: modal never sends orgId in POST body
Reported by **odobenusrosmarus_** in #bugs (Discord, message 1537581143900160083).

**Report (PII-redacted):**

When attempting to create a Device Group from the All Organizations view, you're presented with an error regarding orgId being required in this scope, however if you navigate to a specific organization and try to create the device group you are successful in creating, and confirmed visible from the All Organizations view. 
{"error":"orgId is required when partner has multiple organizations"}

## Triage analysis

The web UI's device-group create form (CreateGroupModal.tsx) never puts `orgId` in the POST body — it only sends {name, type, filterConditions}. The API's POST /groups (mounted at /device-groups) reads orgId exclusively from the JSON body (`payload.orgId`, groups.ts:418). For a partner-scope auth context with more than one accessibleOrgIds, if that body orgId is missing it 400s with exactly 'orgId is required when partner has multiple organizations' (groups.ts:424-431) — this is the literal string the reporter hit. Meanwhile the app's ambient 'currently selected organization' concept only ever gets attached as a URL query string (`?orgId=`) by fetchWithAuth (stores/auth.ts:519-526), which the create-group handler never reads. CreateGroupModal is a single shared component used identically from DevicesPage regardless of whether 'All Organizations' or a specific org is the active context (no orgId prop, no org selector field in the form), and there's no other org-scoped variant of this modal in the codebase. So for any partner user with access to >1 org, this endpoint cannot resolve which org to create the group in from the UI — the request is missing the one field the server needs. (The reporter says a specific-org view succeeds; I could not find a code path that would make that true — auth.accessibleOrgIds is computed once from the JWT/DB and is not narrowed by the ambient query-string orgId — but the missing-body-field bug is independently confirmed by direct inspection of both the form and the handler.)

## Affected code

- `apps/web/src/components/devices/CreateGroupModal.tsx:22-51` — POST body built as {name, type, filterConditions} only — orgId is never included, and there is no org-selector field or orgId prop.
- `apps/api/src/routes/groups.ts:406-438` — POST / reads orgId only from payload (JSON body); for scope==='partner' with accessibleOrgIds.length !== 1 and no payload.orgId, returns 400 'orgId is required when partner has multiple organizations' — the exact error reported.
- `apps/web/src/stores/auth.ts:519-526` — fetchWithAuth only appends the ambient org as a URL query string (?orgId=), which the create-group handler never reads, so no ambient/selected-org context ever reaches the create request.

## Potential fix

Have CreateGroupModal include the currently active organization's id in the POST body (e.g. read it from the same org store fetchWithAuth uses for the query-string injection) when the user has access to more than one org, and/or add an explicit organization selector to the create-group form so partner users can pick a target org.

bdunncompany:
Closes #3505. Closes #3506.

> **Update:** this now also fixes the **focused (single-org) view** for multi-org partners, not just the All-orgs view. The DNS integrations create route reads *only* `body.orgId` (its `resolveOrgId` never reads the injected `?orgId=` query, and `auth.orgId` is null for a partner token), so a multi-org partner 400'd in a focused view too — exactly as #3505 reports ("happens within both full customer view or a specific customer view"). The owner org is now sent in the body in **both** scopes; a focused org is validated against the accessible org list only once it has loaded, so a cold-load fetch failure can't strand a legitimately-focused user. Reviewed adversarially by Codex across three passes (final verdict: no ship-blocking finding).
>
> **Known follow-up (documented, not fixed here):** in the All-orgs view the static-group create *device chooser* lists devices from every org, so picking a device outside the chosen owner org yields a loud, recoverable `400` (the server rejects cross-org devices before insert). Owner-scoped device loading is a separate change.

---

## Problem

Creating a **DNS Security integration** or a **device group** from the **All organizations** (fleet) view fails with `400 "orgId is required for this scope"`. It works from a focused-org view.

Root cause is entirely in the web layer. `fetchWithAuth` auto-injects `?orgId=<currentOrgId>` only when an org is focused; in the fleet view `currentOrgId` is `null`, nothing is injected, and the org-owned create is rejected. Both server routes already **accept and authorize** a body `orgId` — `dnsSecurity` `resolveOrgId()` (checks `canAccessOrg`) and `groups.ts` POST (body-or-query `orgId`, then `ensureOrgAccess`). The forms just never offered a way to supply one in fleet scope.

(Config Policy creation already solved this with its own owner picker; this brings DNS integrations and device groups to parity. The other fleet-create reports triaged alongside this — monitoring-watch null validation, dynamic-group architecture filter, partner-wide monitoring fan-out — were verified **already fixed on `main`** and are intentionally not touched here.)

## Fix

A small `useFleetOrgOwner` hook drives a fleet-only **Organization** picker:

- **Gates on `useOrgScope().scope === 'all'`**, not `currentOrgId === null` — the bare-null read also matches the pre-hydration frame before the first org auto-selects and would flash the picker on a scope about to resolve to a concrete org (same rule as `useDefaultOwnerScope`).
- **Blocks submit until a create has a resolvable owner**: a focused resolved scope (server gets the injected `?orgId=`) or a fleet scope with a valid pick. Loading / org-list-load-error / empty / cleared scopes inject nothing either, so they're now blocked instead of producing the same opaque 400.
- **Validates the pick against the live org list** and **clears a selection that drops out of it**, so a stale/no-longer-accessible org can never be silently reused or resurrected as the owner.
- Wired into `AddDnsIntegrationModal` and `DeviceGroupsPage` (**create only** — edits never move ownership); focused scope is unchanged. Reused existing `common` i18n keys (`labels.organization`, `layout.org.noSelection`, `layout.orgRequired.title`) — **no new locale strings**, locale parity untouched.

## Review

Two independent adversarial reviews were run against the actual diff. Both findings raised — a page-level stale owner surviving modal close, and the unresolved-scope create hole — are fixed here, each with a control-verified regression test (reverting the fix reproduces the failure).

## Testing

- `astro check`: **0 errors**
- Full web vitest suite: **571 files / 5577 tests pass** (node 22.23.2, pnpm 10.34.5)
- New coverage: `useFleetOrgOwner.test.ts` (fleet gating, resolvable-owner blocking, stale-pick drop + no-resurrection, sorting), plus fleet-scope render/disable/POST-body tests in both `AddDnsIntegrationModal.test.tsx` and `DeviceGroupsPage.test.tsx`.
- No dependency or lockfile changes.

### DIFF
--- a/apps/web/src/components/devices/DeviceGroupsPage.dynamicGroups.test.tsx
+++ b/apps/web/src/components/devices/DeviceGroupsPage.dynamicGroups.test.tsx
@@ -6,9 +6,13 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 import DeviceGroupsPage from './DeviceGroupsPage';
 import { fetchWithAuth } from '../../stores/auth';
+import { useOrgStore } from '../../stores/orgStore';
 
 vi.mock('../../stores/auth', () => ({
   fetchWithAuth: vi.fn(),
+  // orgStore registers an orgId provider at module load; the component now
+  // pulls in orgStore via useFleetOrgOwner, so this export must exist.
+  registerOrgIdProvider: vi.fn(),
 }));
 
 vi.mock('../../hooks/useFilterPreview', () => ({
@@ -65,6 +69,23 @@ const exactText = (text: string) => (_content: string, element: Element | null) 
 
 beforeEach(() => {
   vi.clearAllMocks();
+  // A resolved focused-org scope: no fleet picker, create proceeds normally
+  // (the injected ?orgId= supplies the owner).
+  useOrgStore.setState({
+    currentOrgId: 'org-focused',
+    allOrgs: false,
+    organizations: [
+      {
+        id: 'org-focused',
+        partnerId: 'p-1',
+        name: 'Focused Org',
+        status: 'active',
+        createdAt: '2026-01-01T00:00:00Z',
+      },
+    ],
+    organizationsLoaded: true,
+    error: null,
+  });
 });
 
 describe('DeviceGroupsPage dynamic groups', () => {
--- a/apps/web/src/components/devices/DeviceGroupsPage.staticGroups.test.tsx
+++ b/apps/web/src/components/devices/DeviceGroupsPage.staticGroups.test.tsx
@@ -6,9 +6,13 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 import DeviceGroupsPage from './DeviceGroupsPage';
 import { fetchWithAuth } from '../../stores/auth';
+import { useOrgStore } from '../../stores/orgStore';
 
 vi.mock('../../stores/auth', () => ({
   fetchWithAuth: vi.fn(),
+  // orgStore registers an orgId provider at module load; the component now
+  // pulls in orgStore via useFleetOrgOwner, so this export must exist.
+  registerOrgIdProvider: vi.fn(),
 }));
 
 vi.mock('../../hooks/useFilterPreview', () => ({
@@ -80,6 +84,23 @@ const selectDevice = async (
 
 beforeEach(() => {
   vi.clearAllMocks();
+  // A resolved focused-org scope: no fleet picker, create proceeds normally
+  // (the injected ?orgId= supplies the owner).
+  useOrgStore.setState({
+    currentOrgId: 'org-focused',
+    allOrgs: false,
+    organizations: [
+      {
+        id: 'org-focused',
+        partnerId: 'p-1',
+        name: 'Focused Org',
+        status: 'active',
+        createdAt: '2026-01-01T00:00:00Z',
+      },
+    ],
+    organizationsLoaded: true,
+    error: null,
+  });
 });
 
 /**
--- a/apps/web/src/components/devices/DeviceGroupsPage.test.tsx
+++ b/apps/web/src/components/devices/DeviceGroupsPage.test.tsx
@@ -1,13 +1,17 @@
 import '@/lib/i18n';
 
-import { render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 import DeviceGroupsPage from './DeviceGroupsPage';
 import { fetchWithAuth } from '../../stores/auth';
+import { useOrgStore, type Organization } from '../../stores/orgStore';
 
 vi.mock('../../stores/auth', () => ({
   fetchWithAuth: vi.fn(),
+  // orgStore registers an orgId provider at module load; the component now
+  // pulls in orgStore via useFleetOrgOwner, so this export must exist.
+  registerOrgIdProvider: vi.fn(),
 }));
 
 // The preview hook fires its own request; the page's unwrap logic is what's
@@ -66,8 +70,31 @@ const ENVELOPE_RESPONSES: Record<string, unknown> = {
 
 const requestedPaths = () => mockFetch.mock.calls.map((call) => String(call[0]));
 
+const orgA: Organization = {
+  id: 'org-1',
+  partnerId: 'p-1',
+  name: 'Acme Corp',
+  status: 'active',
+  createdAt: '2026-01-01T00:00:00Z',
+};
+const orgB: Organization = {
+  id: 'org-2',
+  partnerId: 'p-1',
+  name: 'Beta Inc',
+  status: 'active',
+  createdAt: '2026-01-01T00:00:00Z',
+};
+
 beforeEach(() => {
   vi.clearAllMocks();
+  // Non-fleet default: the picker stays hidden unless a test opts into All-orgs.
+  useOrgStore.setState({
+    currentOrgId: null,
+    allOrgs: false,
+    organizations: [],
+    organizationsLoaded: false,
+    error: null,
+  });
   mockFetch.mockImplementation(async (url: string) => {
     const path = url.split('?')[0];
     if (path in ENVELOPE_RESPONSES) return jsonResponse(ENVELOPE_RESPONSES[path]);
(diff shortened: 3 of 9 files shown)


ToddHebebrand:
**Reviewed clean — but it now conflicts with main and needs a rebase before I can merge it.**

The review came back with no must-fix and, importantly, **no IDOR**: server-side authorization is intact on both paths this touches. `apps/api/src/routes/dnsSecurity.ts:127` runs `resolveOrgId` → `auth.canAccessOrg(requestedOrgId)` and pins org-scoped tokens to `auth.orgId`, so a forged body 403s; `apps/api/src/routes/devices/groups.ts:452-470` runs `ensureOrgAccess(orgId, auth)` on the body-or-query org. The web change supplies a value the server already validates, which is the right shape. The hook reads `useOrgScope().scope === 'all'` rather than a bare-null check, `useFleetOrgOwner.ts:100` has complete deps and clears rather than masks a dropped pick, and `AddDnsIntegrationModal.tsx:136` stays inside `runAction`. Tests assert real behavior — POST body `orgId`, disabled submit, the programmatic-submit guard — not just names.

**The conflict:** I merged #3564 (Device Groups PUT→PATCH plus phantom-field removal) a few minutes ago, and it rewrites much of `DeviceGroupsPage.tsx` — including removing the `description` and `policyId` form state and making the static device chooser create-only. Your branch touches the same regions and GitHub now reports `CONFLICTING`. Worth knowing: a conflicting PR never triggers CI here, so `gh pr checks` will read as green off the stale run. Rebase onto main and the fresh checks will tell you the truth.

**Three follow-ups worth folding in while you're rebasing**, all the same one-line `useFleetOrgOwner` application:

`apps/web/src/components/dnsSecurity/AddDnsPolicyModal.tsx:104` — `POST /dns-security/policies` hits the same `resolveOrgId`, so a multi-org partner still 400s creating a DNS *policy*, in focused view as well as fleet. Leaving it makes DNS Security half-fixed, which is the harder kind of bug to notice.

`DeviceGroupsPage.tsx:1340` — in the `loading`/`error`/`empty` org-scope states the submit is disabled with no picker and no message; the explanatory `formError` only fires on a programmatic submit. Rendering the existing `OrgLoadFailedState`/`OrgRequiredState` copy would say why.

Static-group create in fleet scope still lists sites from all orgs via `/orgs/sites` (line 317), not just devices. Sites produce the same 400 the device case does; the PR body only documents the device half.

Status: rebase onto main, then ping me and I'll merge.

ToddHebebrand (approved):
**Approved — the rebase is clean, which is the only thing that was actually in question.**

I bounced this purely on the #3564 conflict, so this pass was a rebase-correctness check rather than a fresh design review. The head sits directly on current main (merge-base == main HEAD, zero behind), and the `DeviceGroupsPage.tsx` delta is purely additive: the `useFleetOrgOwner()` call, a create-only reset and guard, `payload.orgId` on `!isEdit`, a create-only `<select>`, and a `disabled` extension.

**The specific thing I was worried about didn't happen.** A rebase across #3564 could easily have resurrected what that PR deliberately removed. It didn't: `description` appears nowhere in the file on either ref; the surviving `policyId` references are byte-identical to main's (the group *type* and bulk-assign paths #3564 kept, not the deleted form field); the update verb is still `PATCH`-era main's, untouched by this diff; and the static device chooser's `else if (isEdit)` branch is unmodified, so it stays create-only.

**Core mechanism re-verified post-rebase:** `useFleetOrgOwner.ts:63` gates on `scope.status === 'resolved' && scope.scope === 'all'` rather than a bare-null read; the effect at `:100` has complete deps and *clears* the pick, gated on `organizationsLoaded` so a transient empty list can't wipe a valid selection; `AddDnsIntegrationModal.tsx:139` keeps `orgId: fleet.bodyOrgId` inside the `runAction` closure. The focused path (`:88`) validates only after the list loads, and can't send a wrong org because `focusedOrgId` comes from the resolved scope, never from picker state.

Server-side authorization was never in doubt and still isn't — `dnsSecurity.ts:127` runs `resolveOrgId` → `auth.canAccessOrg`, and `devices/groups.ts:452-470` runs `ensureOrgAccess` on the body-or-query org. The web layer supplies a value the server independently validates, which is the correct shape.

Tests assert real behavior — parsed POST body `orgId`, disabled→enabled transitions, and a `fireEvent.submit(form)` guard asserting `fetchWithAuth` was never called. Not vacuous.

**One nice-to-have:** `AddDnsIntegrationModal.tsx:326` and `DeviceGroupsPage.tsx:1284` disable submit in *every* unresolved scope (`loading`/`error`/`empty`), where no picker renders — so a partner with zero orgs sees a permanently disabled button with nothing on screen explaining why. Correct behavior, opaque presentation; mostly transient.

Merging. This closes #3505 and unblocks #3506, which was waiting on exactly this — the explicit org selector, not the `currentOrgId` threading that would have looked fixed in a focused view and still failed in All-orgs.

PR #3571, opened 2026-08-15
fix(api,web,shared): record installer type for URL-created software packages

ToddHebebrand:
## The bug

Deploying an MSI by **download URL** could never work. Reported from OliveTech (DO US):

```
fork/exec C:\Windows\SystemTemp\breeze-sw-install-3609918658\package.exe:
This version of %1 is not compatible with the version of Windows you're running.
```

A software version created from a URL never recorded a `file_type`. Only the upload paths derived one (from the uploaded filename) — the JSON create-version route `POST /software/catalog/:id/versions` had no `fileType` field in `createVersionSchema` and never set the column.

`services/softwareDeployment.ts:482-484` then hit its fallbacks at dispatch:

```ts
fileName: versionRecord.originalFileName ?? `package.${versionRecord.fileType ?? 'exe'}`,
fileType: versionRecord.fileType ?? 'exe',
```

The agent received `fileType: "exe"`, saved the MSI as `package.exe`, and took the EXE branch of `executeInstaller` (`agent/internal/remote/tools/software_install.go:557-564`), which execs the downloaded file **directly**:

```go
cmd = exec.CommandContext(ctx, localPath, parts...)
```

Windows rejects an MSI at CreateProcess with `ERROR_BAD_EXE_FORMAT`. The operator's `msiexec /i "{file}" /qn /norestart` was passed as argv **to the MSI** rather than to msiexec — it was never used. Had `fileType` been `msi`, line 554 would have run msiexec correctly.

There was no workaround short of re-adding the package as an upload.

## The fix

| Area | Change |
|---|---|
| `packages/shared` | New `softwareFileType` util: `deriveSoftwareFileTypeFromUrl` + `defaultSilentArgsForFileType`, shared so the browser preview and the authoritative server derivation can't drift |
| `routes/software.ts` | `createVersionSchema` accepts optional `fileType`; handler prefers it, else infers from the URL, and applies the MSI silent-arg defaults the upload path already had |
| `routes/software.ts`, `routes/softwareUploads.ts` | Both upload paths take the msiexec strings from the shared helper instead of duplicating literals |
| `AddPackageModal` | Installer-type selector on the URL source (auto-detect default, shows what was detected), MSI arg prefill on URL entry matching the existing file-source behavior, explicit warning when the URL has no recognizable extension. Strings in all 7 locales |
| Migration | Backfills `file_type` for existing URL-created versions |

### Design notes

**Derivation returns `null` rather than guessing.** A URL like `https://vendor.example/download.php?product=foo` yields null and keeps the historical `'exe'` dispatch — an over-confident guess just re-creates this bug pointed the other way. That is what the explicit selector and the amber warning in the UI are for.

**Query/fragment stripped without `new URL()`.** Presigned vendor URLs put the signature after the extension, and a stored URL may still carry unresolved `{{org.name}}` / `{{var.site_key}}` deploy-time tokens that make the parser throw or percent-encode the braces.

**MSI is the only type with prefilled args.** An EXE's silent switch is vendor-specific (`/S`, `/silent`, `/quiet`, `-q`); guessing one installs interactively on an unattended machine.

**`silent_install_args` is deliberately not rewritten by the migration.** With the type corrected, `buildMSIExecArgs` already defaults empty args to `/i <path> /qn /norestart` and strips a leading `msiexec` token from a supplied one. Editing an operator's stored command is out of scope for a data repair.

## Schema / contracts

No new tables and **no new columns** — `file_type` and `original_file_name` already exist on `software_versions`. No RLS, cascade, or export-policy registration applies.

Migration is dated `2026-08-23` rather than today: existing migrations already run to `2026-08-22`, so today's date would not have sorted last. `check-migration-naming.sh` passes.

## Verification

**Migration applied against live Postgres** inside a rolled-back transaction:

```
WARNING:  backfilled file_type on 3 url-created software_versions row(s)
 version |                   url                   | file_type
---------+-----------------------------------------+-----------
 1       | https://cdn.example/acme.msi            | msi
 2       | https://cdn.example/acme.MSI?sig=x      | msi
 3       | https://vendor.example/download.php?p=1 | NULL     <- unusable ext, untouched
 4       | https://cdn.example/setup.exe           | exe
 5       | https://cdn.example/other.msi           | exe      <- already set, untouched
 6       | (null)                                  | NULL
=== second apply === (no warning — idempotent)
```

**Tests** — new assertions were mutation-tested, not just run:

- `packages/shared` — 1938 passed (12 new cases)
- `apps/api` `software.test.ts` + `softwareUploads.test.ts` — 137 passed (7 new). Reverting the fix fails 5.
- `apps/web` `AddPackageModal.test.tsx` — 19 passed (4 new). Reverting the fix fails 2.
- `apps/web` i18n coverage — 98 passed; `autoMigrate.test.ts` — 60 passed
- eslint clean on all changed sources

## Follow-up (not in this PR)

URL-created versions still send `originalFileName: null`, so the agent saves `package.<type>`. That is correct and always matches `file_type` — deriving a filename from the URL risks an extension/type mismatch that `validateInstallFileName` rejects outright. Worth revisiting only if a vendor installer turns out to care about its own filename.

### DIFF
--- a/apps/api/migrations/2026-08-23-software-version-url-file-type.sql
+++ b/apps/api/migrations/2026-08-23-software-version-url-file-type.sql
@@ -1,45 +1,65 @@
--- Backfill software_versions.file_type for versions created from a download URL.
+-- Backfill software_versions.file_type for versions that never recorded one.
 --
 -- The JSON create-version route (POST /software/catalog/:id/versions) never
 -- accepted or derived a file_type -- only the upload paths did, from the
 -- uploaded filename. services/softwareDeployment.ts therefore dispatched
 -- `fileType: 'exe'` / `fileName: 'package.exe'` for every URL-created version,
--- and the agent's EXE branch execs the downloaded file DIRECTLY
--- (exec.CommandContext(ctx, localPath, ...) in software_install.go). For an MSI
--- that fails at CreateProcess with ERROR_BAD_EXE_FORMAT ("This version of %1 is
--- not compatible with the version of Windows you're running") and the
--- operator's `msiexec /i "{file}" ...` command is discarded unused.
+-- and the agent's Windows EXE branch execs the downloaded file DIRECTLY
+-- (exec.CommandContext(ctx, localPath, parts...) in software_install.go). For an
+-- MSI that fails at CreateProcess with ERROR_BAD_EXE_FORMAT ("This version of %1
+-- is not compatible with the version of Windows you're running"). The operator's
+-- `msiexec /i "{file}" ...` was not discarded but MISROUTED -- passed as argv to
+-- the MSI itself. On macOS/Linux the same 'exe' fallback failed earlier still,
+-- as `unsupported file type "exe" on <os>`, which is why dmg/deb/pkg rows are
+-- in scope here too.
 --
--- Only rows where the URL carries an unambiguous installer extension are
--- touched. A URL with no usable extension keeps file_type NULL and the
+-- Derivation order is original_file_name FIRST, then download_url. This is not
+-- arbitrary: the agent's validateInstallFileName rejects the command outright
+-- unless the filename's extension equals '.' || file_type, and
+-- softwareDeployment.ts sends original_file_name verbatim as the filename when
+-- it is set. So for those rows the filename is the ONLY source that yields a
+-- dispatchable pair -- deriving from the URL instead could stamp a file_type
+-- that contradicts the filename and convert a working install into a hard
+-- validation failure.
+--
+-- Only rows whose source carries an unambiguous installer extension are
+-- touched. A row with no usable extension keeps file_type NULL and the
 -- historical 'exe' dispatch behavior -- guessing there would trade one wrong
--- answer for another.
+-- answer for another. Those rows are counted and reported below, because THAT
+-- is the actionable number: they are the packages that will still fail their
+-- next deploy.
 --
 -- silent_install_args is deliberately NOT rewritten: with file_type corrected,
 -- buildMSIExecArgs already defaults an empty args string to
--- `/i <path> /qn /norestart` and strips a leading `msiexec` token from a
--- supplied one. Editing an operator's stored install command is out of scope
--- for a data repair.
+-- `/i <path> /qn /norestart`, strips a leading `msiexec` token from a supplied
+-- one, and prepends `/i <path>` when the supplied args don't already reference
+-- the file -- so stored bare switches like `/qn /norestart` keep working.
+-- Editing an operator's stored install command is out of scope for a data repair.
 
 DO $$
 DECLARE
   updated_count integer;
+  skipped_count integer;
 BEGIN
   WITH derived AS (
     SELECT
       id,
       lower(
-        substring(
+        coalesce(
+          -- The filename the agent will validate against, when we have one.
+          substring(original_file_name FROM '\.([A-Za-z0-9]+)$'),
           -- Path portion only: a presigned or tokenized URL puts the signature
           -- in the query string, and `{{org.name}}`-style deploy-time tokens can
           -- appear anywhere in it.
-          split_part(split_part(download_url, '#', 1), '?', 1)
-          FROM '\.([A-Za-z]+)$'
+          substring(
+            split_part(split_part(download_url, '#', 1), '?', 1)
+            FROM '\.([A-Za-z0-9]+)$'
+          )
         )
       ) AS ext
     FROM software_versions
     WHERE file_type IS NULL
-      AND download_url IS NOT NULL
+      AND (download_url IS NOT NULL OR original_file_name IS NOT NULL)
   )
   UPDATE software_versions sv
   SET file_type = derived.ext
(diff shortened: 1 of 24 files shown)


ToddHebebrand:
## Review pass applied — 5 agents (code · tests · comments · errors · types)

Everything below is a confirmed finding acted on, not a suggestion logged. Two findings changed the *scope* of the fix; four corrected claims I had made incorrectly.

### The fix was incomplete

`SoftwareVersionManager.tsx` — the **add-version-to-an-existing-package** flow — POSTs to the same `/software/catalog/:id/versions` route with `downloadUrl` and no `fileType`, no selector, no warning, no prefill. That is the flow an MSP hits far more often over a package's life than the create-package modal, so the original bug still reproduced there. It now has the same treatment, with its own test file.

### The first commit introduced a silent misconfiguration

The MSI prefill was one-directional — fill when empty, never clear. Type a `.msi` URL, correct it to `.exe`, and the msiexec command stays. The agent then runs:

```
setup.exe msiexec /i C:\...\package.exe /qn /norestart
```

Most NSIS/InstallShield installers ignore unrecognized switches and **open an interactive UI on an unattended machine** — the install hangs rather than failing loudly. Nothing catches it agent-side either: `validateMSIInstallArgs` only runs when `fileType == "msi"`. Same failure family this PR exists to close.

New `lib/installerArgsPrefill.ts` makes the prefill **owned** rather than sticky: a field still holding a string the module could have produced is revised or withdrawn when the type changes; anything hand-typed is never touched. Ownership is decided by value, which is also what makes it correct across a URL→file source switch, where the previous type isn't otherwise recoverable.

### The bridge the bug actually lived on had zero tests

`softwareDeployment.ts` builds the `fileName`/`fileType` pair the agent receives. **No test in the repo asserted it** — every dispatch fixture set `originalFileName`, so the `` `package.${fileType}` `` branch that *every* URL-created version takes was never evaluated. You could revert that line to a hardcoded `'exe'` and the whole suite stayed green.

Added four tests, including the agent's real invariant — `fileName` must end in `.` + `fileType`, per `validateInstallFileName`. The two fields are coupled, not independent, and the coupling was enforced only on the device.

Also narrowed on read via `isSoftwareFileType`: the column is a bare nullable `varchar` with no CHECK, so any value that ever landed there reached the device verbatim and was rejected only *after* the download.

### Two reviewers disagreed about the migration; resolved on the merits

One wanted to backfill **from** `original_file_name`; the other wanted to **exclude** those rows to avoid creating a `fileName`/`fileType` mismatch. Both are half-right. The agent validates `fileName`'s extension against `file_type`, so when `original_file_name` is set it is the *only* source that yields a dispatchable pair. Derivation is now **`original_file_name` first, URL second** — which fixes those rows instead of either skipping them or corrupting them.

The migration now also reports the **skipped** count, unconditionally including zero. The skipped count is the actionable number — it's the set of packages that will still fail their next deploy. `"backfilled 12 rows"` alone reads as *done* when 40 may remain broken.

### My comments were wrong about the mechanism (the fix was not)

I wrote that the operator's msiexec command was **"discarded unused"** in four places. Wrong. The EXE branch *does* use `silentInstallArgs` — it substitutes `{file}`, tokenizes, and passes the result as argv **to the downloaded file**. Real argv was `[<the MSI>, "msiexec", "/i", <path>, "/qn", "/norestart"]`. **Misrouted, not dropped.** Corrected everywhere.

Also corrected: my claim that MSI is the only type with a "universally correct" silent-args pair implied deb/pkg/dmg switches merely aren't guessable. The agent **never reads `silentInstallArgs` for those at all** — `dpkg -i`, `installer -pkg`, and mount-and-install are hardcoded, so anything stored there is silently ignored. That misled in the operator-harmful direction. The header also cited a consumer file that imports none of these helpers, and every "execs directly" claim was Windows-only (macOS/Linux failed earlier as `unsupported file type` — which is why dmg/deb/pkg rows are in the backfill's scope).

### Drift and types

`ALLOWED_EXTENSIONS` is now **derived** from `SOFTWARE_FILE_TYPES` rather than the second hand-maintained copy my first commit created. `defaultSilentArgsForFileType` normalizes case — a stored `'MSI'` previously yielded no defaults silently while every other layer is case-insensitive. Return type named `SilentInstallDefaults`, with the `{file}` token contract documented.

Verified there is **no drift today** between the Go agent's `isSupportedInstallFileType`, `ALLOWED_EXTENSIONS`, and `SOFTWARE_FILE_TYPES`. The DB column was never a member of that agreement (`varchar(20)`, no CHECK) — which is what the read-side narrowing above now compensates for.

### Observability

The route records a warning when it stores a NULL `file_type` for a URL it couldn't classify. Not a 400 — an extensionless URL serving a real EXE is legitimate, and rejecting would break scripted clients. But the web forms warn in the UI while an API/MCP/script client got **no signal at all**, and nobody could count how often this happens in production. Host only, never the URL: managed-software URLs carry presigned capability query strings.

### Verification

Every new assertion was mutation-tested, not merely run:

| Mutation | Result |
|---|---|
| Remove migration's filename precedence | integration test fails — `expected null to be 'msi'` |
| Revert dispatch read-narrowing | 2/4 bridge tests fail |
| Revert `SoftwareVersionManager` wiring | 3/6 tests fail |
| Revert `AddPackageModal` wiring | 2 tests fail |
| Revert API `fileType` derivation | 5 tests fail |

The migration integration test was run against **real Postgres on an isolated stack** — the shared test DB carried a sibling branch's migration and refused the run, which is the documented false-result trap. 244 API + 415 web + shared suites green; eslint clean.

### Deliberately not done

A `warnings` array in the 201 response body. It is the only mechanism that reaches a non-web client, and the case for it is good — but it changes the public response envelope, so it deserves its own decision rather than riding along in a bugfix. Same for a CI guard asserting the Go↔TS installer-type lists stay in sync; today they agree, and nothing but a comment says they must.

PR #3577, opened 2026-08-15
fix(roles): let a partner admin create and manage org-scoped roles (#3524)

### LINKED ISSUE #3524: [Auth][SSO] Org-axis SSO auto-provisioning is unconfigurable: the default-role picker can never be populated by a partner admin
Reported by @myrfnrz in discussion #3464 (self-hosted, Google SSO, organization-axis provider).

## Symptom

Every JIT SSO login fails with `/login?error=default_role_required`. The admin cannot fix it from the UI, because the field the error is asking them to fill can never be populated.

## The dead end

Three correct-in-isolation rules intersect into an unsatisfiable requirement for a partner admin:

**1. An org-axis provider with auto-provisioning needs a default role.** `apps/api/src/routes/sso.ts:2588` redirects `default_role_required` when `autoProvision` is set and `validatedDefaultRoleId` is absent.

**2. That role must be a custom, org-scoped role.** `apps/web/src/components/settings/SsoProviderForm.tsx:154-159` filters the dropdown to `scope !== 'partner'` and `!isSystem`. Excluding system roles is deliberate and correct (SR2-10 Fix 1) — a built-in role has NULL `org_id`/`partner_id`, so JIT could never resolve it and `sso.ts:465` would 400 it. So the built-in Org Admin / Org Technician / Org Viewer are all unselectable.

**3. A partner admin cannot create an org-scoped role.** `apps/api/src/routes/roles.ts:92-98` — `getScopeContext` derives scope solely from the auth token: `auth.scope === 'partner'` always yields `{ scope: 'partner' }`. `POST /roles` therefore always writes a partner-scoped role. There is no target-org parameter and no other creation surface.

Net effect: the dropdown is empty and stays empty. Clearing the field to save the form is what produces the login failure.

## Why it isn't caught today

Each rule is individually defensible and individually tested. Nothing asserts the composite property — that for every provider axis, at least one role satisfying the picker's own filter is reachable by an admin who can configure that provider.

## Suggested fix

Let a partner admin create a role scoped to a chosen organization: accept an optional target `orgId` on `POST /roles` for partner-scoped callers, validated against `breeze_has_org_access` so it cannot escape the partner's tree, and surface an org selector in Settings → Roles. The dropdown filter in `SsoProviderForm.tsx` is correct as written and should not be relaxed — offering a system role would only move the failure to the JIT callback.

A regression test should assert the composite property rather than the mechanism: for an org-axis provider, a partner admin can reach a selectable `defaultRoleId`.

## Not established

Whether the partner axis has an equivalent trap. Partner-wide providers do not JIT at all (`sso.ts:2576` returns `invite_required`), so `default_role_required` looks unreachable there, but I did not test it.

## Workarounds for anyone hitting this now

- Invite a user into the organization, give them the built-in Org Admin role, sign in as them, and create a custom organization role. It will then appear in the partner admin's SSO picker.
- Or disable auto-provisioning and pre-create accounts. `sso.ts:2582` returns `user_not_found` instead, and existing passwordless accounts get linked (`sso.ts:2560-2567`).

bdunncompany:
## What

Fixes #3524 — an org-axis SSO provider's default-role picker could never be populated by a partner admin, so auto-provisioning was unconfigurable and every JIT login failed with `default_role_required`.

Three correct-in-isolation rules intersected into a dead end: (1) an org-axis provider needs a default role; (2) that role must be a custom **org-scoped** role (`SsoProviderForm` filters `scope !== 'partner' && !isSystem`, correctly); (3) `getScopeContext` derived scope solely from the token, so a partner admin's `POST /roles` always wrote a *partner*-scoped role. The dropdown was therefore always empty.

## Change

A partner admin can now create **and fully manage** roles scoped to an organization **inside their own tree**, which populates (and maintains) the picker. The tenant boundary is centralized in one helper.

**API (`apps/api/src/routes/roles.ts`)**
- `resolvePartnerFocusedOrgId(auth, orgId)` — the single gate: returns the org only when the caller is partner-scope, `auth.canAccessOrg(orgId)` is true (for a partner that resolves to exactly their own orgs), and the org is **not** the hidden `quick_support` org. The `roles` RLS policies (`breeze_has_org_access`) backstop every query at the DB.
- `resolveActingScope(auth, orgId)` — org scope when the helper yields one, else the caller's token scope.
- `POST /roles` accepts an optional `orgId`; a provided-but-invalid org is a hard `403`.
- `GET /roles` merges the focused org's org-scoped roles (+ correct per-axis user counts) for a partner supplying `?orgId=`.
- The six detail/mutation endpoints (`GET/:id`, `PATCH`, `DELETE`, `clone`, `users`, `effective-permissions`) resolve scope through `resolveActingScope`, so the whole lifecycle is consistent — no more "created a role you can't edit". Each still enforces `role.orgId === scopeContext.orgId`.

**Web**
- `RolesPage` sends `orgId` when focused on one org.
- `SsoProvidersPage` re-fetches roles/providers when the org scope resolves, fixing a cold-load race that left the picker empty until a manual reload.

## Tests

- `roles.test.ts`: org-scoped create; `403` on a foreign org; `403` on `quick_support`; `GET` merge of the focused org's roles + counts; foreign `?orgId` yields no merge; a focused partner can fetch an org-scoped role via `?orgId`. (80 pass.)
- `RolesPage.test.tsx`: create body carries `orgId` only when focused on an org.
- Composite property (per the issue): for an org-axis provider, a partner admin can now reach a selectable `defaultRoleId`.

The `SsoProviderForm` filter is intentionally unchanged — offering a system role would only move the failure to the JIT callback, as the issue notes.

### DIFF
--- a/apps/api/src/routes/roles.test.ts
+++ b/apps/api/src/routes/roles.test.ts
@@ -99,6 +99,17 @@ const { ASSIGNABLE_PERMISSIONS: REAL_ASSIGNABLE_PERMISSIONS } = await vi.importA
   typeof import('../services/permissions')
 >('../services/permissions');
 
+// #3524: resolvePartnerFocusedOrgId does a `select({type}).from(organizations)
+// .where().limit(1)` to exclude the hidden quick_support org. Queue this FIRST on
+// db.select for any request that supplies a valid focused/target orgId.
+function mockOrgTypeLookup(type: 'customer' | 'quick_support' = 'customer') {
+  vi.mocked(db.select).mockReturnValueOnce({
+    from: vi.fn().mockReturnValue({
+      where: vi.fn().mockReturnValue({ limit: vi.fn().mockResolvedValue([{ type }]) })
+    })
+  } as any);
+}
+
 describe('role routes', () => {
   let app: Hono;
 
@@ -188,6 +199,108 @@ describe('role routes', () => {
       expect(body.data).toHaveLength(2);
       expect(body.data[1].parentRoleName).toBe('Admin');
       expect(body.data[1].userCount).toBe(3);
+    });
+
+    // #3524: a partner focused on one org (web client injects ?orgId=) must ALSO
+    // see that org's org-scoped roles, or the role they just created is invisible
+    // and the SSO org-provider picker stays empty (the fix would be inert).
+    it('partner with a valid ?orgId merges the focused org’s org-scoped roles + counts', async () => {
+      const ORG = '22222222-2222-4222-8222-222222222222';
+      vi.mocked(authMiddleware).mockImplementation((c: any, next: any) => {
+        c.set('auth', {
+          scope: 'partner',
+          partnerId: 'partner-123',
+          partnerOrgAccess: 'all',
+          orgId: null,
+          user: { id: 'user-123', email: 'test@example.com' },
+          canAccessOrg: (id: string) => id === ORG
+        });
+        return next();
+      });
+      const now = new Date();
+      mockOrgTypeLookup('customer'); // resolvePartnerFocusedOrgId quick_support check
+      vi.mocked(db.select)
+        // 1) partner-scoped roles
+        .mockReturnValueOnce({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockResolvedValue([
+              { id: 'p1', name: 'Partner Admin', description: null, scope: 'partner', isSystem: true, parentRoleId: null, createdAt: now, updatedAt: now }
+            ])
+          })
+        } as any)
+        // 2) focused org's org-scoped roles
+        .mockReturnValueOnce({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockResolvedValue([
+              { id: 'o1', name: 'Org Tech', description: null, scope: 'organization', isSystem: false, parentRoleId: null, createdAt: now, updatedAt: now }
+            ])
+          })
+        } as any)
+        // 3) partner_users counts
+        .mockReturnValueOnce({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({ groupBy: vi.fn().mockResolvedValue([{ roleId: 'p1', count: 2 }]) })
+          })
+        } as any)
+        // 4) organization_users counts for the focused org
+        .mockReturnValueOnce({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({ groupBy: vi.fn().mockResolvedValue([{ roleId: 'o1', count: 5 }]) })
+          })
+        } as any);
+
+      const res = await app.request(`/roles?orgId=${ORG}`, {
+        method: 'GET',
+        headers: { Authorization: 'Bearer token' }
+      });
+
+      expect(res.status).toBe(200);
+      const body = await res.json();
+      expect(body.data).toHaveLength(2);
+      const org = body.data.find((r: any) => r.id === 'o1');
+      expect(org).toMatchObject({ scope: 'organization', userCount: 5 });
+      const partner = body.data.find((r: any) => r.id === 'p1');
+      expect(partner).toMatchObject({ scope: 'partner', userCount: 2 });
+    });
+
+    it('partner with an ?orgId OUTSIDE its tree gets only partner roles (no merge)', async () => {
+      vi.mocked(authMiddleware).mockImplementation((c: any, next: any) => {
+        c.set('auth', {
+          scope: 'partner',
+          partnerId: 'partner-123',
+          partnerOrgAccess: 'all',
+          orgId: null,
+          user: { id: 'user-123', email: 'test@example.com' },
+          canAccessOrg: () => false
+        });
+        return next();
+      });
+      const now = new Date();
+      // Only the partner-roles select + partner_users counts run — no org query,
+      // because canAccessOrg rejects the injected orgId.
+      vi.mocked(db.select)
+        .mockReturnValueOnce({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockResolvedValue([
+              { id: 'p1', name: 'Partner Admin', description: null, scope: 'partner', isSystem: true, parentRoleId: null, createdAt: now, updatedAt: now }
+            ])
+          })
+        } as any)
+        .mockReturnValueOnce({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({ groupBy: vi.fn().mockResolvedValue([]) })
+          })
+        } as any);
+
+      const res = await app.request('/roles?orgId=99999999-9999-4999-8999-999999999999', {
+        method: 'GET',
+        headers: { Authorization: 'Bearer token' }
+      });
+
+      expect(res.status).toBe(200);
+      const body = await res.json();
+      expect(body.data).toHaveLength(1);
+      expect(body.data[0].scope).toBe('partner');
     });
 
     it('should reject missing partner/org context', async () => {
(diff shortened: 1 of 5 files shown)


ToddHebebrand (changes requested):
**Requesting changes — the authorization design is sound, but two scope regressions fall out of `fetchWithAuth`'s automatic `?orgId=` injection.**

First the good part, because it's the load-bearing half: `resolvePartnerFocusedOrgId` (`apps/api/src/routes/roles.ts:118`) gates the target org on `auth.canAccessOrg(targetOrgId)`, excludes the hidden `quick_support` org, and hard-403s a provided-but-invalid org rather than falling through to a partner-scoped role. The `breeze_role_isolation_insert` policy backstops it at the DB. I went looking for an escalation path and there isn't one — a partner admin cannot mint a role outside their own tree. No schema change, so no cascade or export-policy obligation. That part is right.

**Must-fix 1 — a focused partner admin 404s on every partner-scoped role.** `apps/web/src/stores/auth.ts:638`: `fetchWithAuth` auto-injects `?orgId=` into *every* request unless the caller opts out with `skipOrgIdInjection`, and `orgStore` auto-selects the first org on load. So a normal partner session always sends an orgId. `resolveActingScope` (`roles.ts:140`) then returns `{ scope: 'organization' }`, and the `role.scope !== scopeContext.scope` guard — `roles.ts:743` on GET `/:id`, `:867` on PATCH, and the parallel copies on DELETE, clone, users and effective-permissions — returns 404. `GET /roles` still *lists* those partner-scoped roles, so the UI shows a role that 404s the moment you open it. Resolve the acting scope from the role's own axis rather than from the query param alone. Nothing in `roles.test.ts` pins this; a "focused partner can still PATCH a partner-scoped role" case would have caught it.

**Must-fix 2 — org-scope admins can no longer create roles at all.** `apps/web/src/components/settings/RolesPage.tsx:204-206` attaches `orgId` whenever `getOrgScope().scope === 'org'`, which is always true for an org-token user. The POST handler at `roles.ts:577` treats any `body.orgId !== undefined` as a partner-focus request, and `resolvePartnerFocusedOrgId` returns `null` for a non-partner caller by its first condition, so it 403s "You do not have access to the specified organization" — on the caller's *own* org. Treating `auth.orgId === body.orgId` as a no-op re-scope fixes it. The new `RolesPage.test.tsx` asserts only the request body shape, never the resulting API contract, which is why CI is green.

**Nice-to-have:** the four-line `#3524` explanation is duplicated verbatim at six call sites — one copy on the helper is enough. And now that `?orgId=` is present on every roles request, each one pays an extra `organizations` lookup in `resolvePartnerFocusedOrgId`.

Worth flagging separately: **#3579 contains this entire diff** — the same `roles.ts` +186/-51 and `roles.test.ts` +327 — under a `fix(web): surface swallowed backup-vault failures` title that never mentions the RBAC change. I've asked there for the roles half to be dropped so this PR stays the single reviewable home for it.

Status: changes requested — two regressions above, then re-review.

bdunncompany:
Both regressions fixed, rebased onto current main. Your diagnosis was exact on both, and both were things my tests couldn't see because they asserted the request shape rather than the API contract.

**Must-fix 1 — focused partner 404s on partner-scoped roles.** Confirmed the mechanism you described: `stores/auth.ts:638` injects `?orgId=` into every request and `orgStore` auto-selects, so a partner session always sends one. The acting scope for a single-role request now comes from the **role's own axis** (`resolveRoleActingScope`), not the query param. A partner-scoped role resolves on the partner axis regardless of any focused org.

**Must-fix 2 — org admins 403'd on their own org.** `RolesPage.tsx:204` attaches `orgId` for every org-token user, and the POST handler treated any `body.orgId` as a partner-focus request. Naming your own org is now a no-op re-scope. Test asserts the API contract this time: control-run against the old code it gets **403**, with the fix **201**.

**A third defect, found while attacking my own fix — and this one is worse than either of the above.**

Deriving the org from `role.org_id` alone lets a role id silently override the org the request *names*. A partner has umbrella access to every org in its tree, so RLS happily returns org B's role to a partner focused on org A — and `PATCH`/`DELETE`/`clone` would then operate against **B** while the UI and the audit entry both say **A**. That is the wrong customer, not merely the wrong scope, and it contradicts the contract the comments in this file promise ("a partner acts on an org by naming it").

An organization role now requires the request to name a validated org **and** for that org to equal the role's own. Partner-scoped roles still ignore the focused org, which is what fixes must-fix 1. Control-run: without the match check the request sails past the guard into the transaction; with it, 404 and no transaction opened.

Relatedly, an **org token** now resolves against its own token org rather than the query string. An org caller can't change organization context, so a stale or hand-supplied `?orgId=` should never be able to deny it a role it legitimately owns.

Net effect: strictly narrower than the previous revision of this PR, and no wider than main.

**On the cross-partner `is_system` question** — I checked rather than assuming RLS covers it. `partnerCreate.ts:94` does create a system role with a real `partner_id`, and the helper skips the ownership check for system roles. What saves it is that `breeze_role_isolation_select`'s system-role carve-out requires **both** `partner_id` and `org_id` to be null, so a partner-owned system role is only ever visible via `breeze_has_partner_access` — its own partner. The row never reaches the helper. Worth a hostile integration test at some point so this doesn't rest on RLS alone.

**One follow-up I deliberately left out** rather than widen this PR again: cloning a **partner-scoped** template while focused on an org produces another partner-scoped role, which `SsoProviderForm.tsx:154` filters out of the org default-role picker — so that particular path still doesn't populate the picker. Clone infers its destination from the source role instead of taking an explicit target. That shape predates this PR (main's clone used `getScopeContext` too), and fixing it properly means giving clone an explicit destination org plus a web change. Say the word and I'll do it as its own PR. The ordinary create path does target the selected org correctly, so #3524's main workflow works.

Also took both smaller notes: the four-line comment is down to one copy on the helper, and the `organizations` lookup is no longer universal — it runs only when a role is actually org-scoped, rather than on every roles request carrying an `?orgId=`.

**Verification** — node 22.23.2 (CI `.nvmrc`), pnpm 10.34.5, on current main:

- `api` `src/routes/` + `roleAssignment` — 479 files, **8631/8631**
- `api` `roles.test.ts` — **84/84**
- `web` RolesPage + SsoProvidersPage — **10/10**
- `web` eslint on both changed files — clean

All three new tests were control-run against the unfixed code first and fail there.

ToddHebebrand (approved):
**Approved — both of my must-fixes verified fixed, and the third defect you found on your own was the important one.**

I traced all three rather than taking the writeup at its word.

**Must-fix 1 — resolved.** `resolveRoleActingScope` (`apps/api/src/routes/roles.ts:159-201`) branches on `role.scope` rather than the injected `?orgId=`, and the partner branch still asserts `role.partnerId === tokenScope.partnerId`. What matters more than the helper is that it's applied at **all six** single-role call sites — GET `/:id` (:791), PATCH (:904), DELETE (:1076), clone (:1176), users (:1306), effective-permissions (:1384) — and every one of those selects pulls `partnerId`/`orgId`/`scope`/`isSystem`, so the helper actually has the fields it branches on. A correct helper missing from one call site would have been the realistic failure here; there isn't one, and `/permissions/available` is genuinely scope-agnostic.

**Must-fix 2 — resolved.** `roles.ts:638` skips the focus branch when `auth.scope === 'organization' && auth.orgId === body.orgId`, so naming your own org is the no-op you describe. I checked the adjacent hole too: an org token naming a *foreign* org still 403s, because `resolvePartnerFocusedOrgId` returns null for non-partner scope. And `body.orgId` never reaches the insert — `roleData` is assembled field-by-field at :667-687.

**Your third finding is the one that would have hurt.** A partner has umbrella access across its tree, so RLS would happily hand back org B's role to a partner focused on org A, and PATCH/DELETE/clone would mutate B while the UI and the audit row both said A. Wrong-customer, not wrong-scope — and it would have looked correct in every test written against a single-org fixture. `roles.ts:193-196` now requires a validated `?orgId=` **and** `role.orgId === resolved`, so `resolveAuditOrgId` provably equals the named org. Finding that by attacking your own fix, after the review had already passed, is the right instinct.

No schema change, so no cascade or export-policy obligation — correctly untouched.

**Nice-to-haves for a follow-up, none blocking:**
1. `roles.test.ts` covers the focused-partner branch only via GET `/:id` (:665). A PATCH 200 asserting the *mutated row's* `orgId` would pin fix 3 against regression — that's the case where a silent re-introduction costs a customer.
2. No test for an org token naming a foreign org; correct today, unguarded.
3. `resolvePartnerFocusedOrgId` re-queries `organizations` on each of :193/:498/:638 — memoize on the Hono context.
4. `roles.ts:526` now surfaces org-scoped *system* roles to a focused partner; cosmetic, since the SSO picker filters `isSystem`.

Merging — this unblocks #3524.

Issue #3531, opened 2026-08-14
Destructive confirmations hide failures (bare fetch / error behind modal / console-only) — operator sees success on 403/500


bdunncompany:
From an exhaustive web UI/UX audit (Codex, gpt-5.6-sol, `origin/main`). A recurring pattern: **destructive confirmations run a bare `fetch` (no `runAction`, no `response.ok` check), store any failure in page-level state that renders *behind* the still-open modal, or swallow it to the console** — so a 403/500/network failure looks identical to success. The operator believes the action happened.

- **API-key revocation** — `ApiKeysPage.tsx:146` bare DELETE, error at `:212` behind the `z-50` modal at `:352`. An admin can believe a compromised key was revoked while it stays active.
- **Automation delete** — `AutomationsPage.tsx:303`, error at `:367` behind overlay `:383`.
- **Automation-policy delete** — `PoliciesPage.tsx:92`, banner `:170` behind modal `:203`.
- **Quarantined-device denial** — `QuarantinedDevices.tsx:90`, banner `:161` behind modal `:250`. A rogue enrollment can look denied while still pending.
- **Vault sync** — `DeviceVaultStatus.tsx:59` bare POST, never checks `response.ok`, empty catch `:65`. A failed sync looks complete; an unavailable vault looks unconfigured.
- **Remote Tools** (processes/services/event-logs/scheduled-tasks) — `RemoteToolsPage.tsx:488,525,611,655` catch to console only; failed refreshes leave stale data displayed as current and first-load failures look like empty lists. Dangerous when stopping services / killing processes.

These are exactly the `no-silent-mutations` contract (CLAUDE.md) applied to destructive confirm flows. Fix: wrap in `runAction` (or at minimum check `response.ok`, keep the modal open on failure, and surface the error). The `no-silent-mutations` guard is not repo-wide, so these are unguarded.

ToddHebebrand:
**One of the six sites is fixed; leaving this open for the other five.**

#3579 shipped the **Vault sync** item — `DeviceVaultStatus` now checks `response.ok`, routes through `runAction`, and renders a distinct error card with a Retry instead of collapsing a load failure into the "no vault configured" empty state.

[...]

PR #3579, opened 2026-08-15
fix(web): surface swallowed backup-vault failures and show the real status (#3531)

bdunncompany:
## What

The per-device **Local Vault** panel (`DeviceVaultStatus`) hid failures and misreported state. Two problems, one of them found in review:

1. **Swallowed failures (#3531).** `handleSync` did a bare `POST /backup/vault/:id/sync` with no `response.ok` check and an empty `catch`, and `fetchVault` swallowed load errors. A failed **backup** sync looked complete, and a failed status load looked like "no vault configured". Since `fetchWithAuth` resolves on non-2xx, the missing checks silently proceeded.

2. **Wrong field mapping (caught by adversarial review).** The component read `vault.type` and `vault.status`, but the API (`toVaultResponse` in `routes/backup/vault.ts`) returns `vaultType` and `lastSyncStatus`. So the panel **always** rendered "Never synced" regardless of the real state.

## Change

- `handleSync` and `fetchVault` check `response.ok` and surface an `error`. A load failure with no vault yet renders a distinct error card (not the "no vault configured" empty state), with a **Retry** so a transient failure isn't a dead end.
- Read the real API fields: `vaultType`, and `normalizeVaultStatus(lastSyncStatus)` (`completed/ok/success`→Completed, `failed/error`→Failed, `pending/running`→Syncing, null→Never synced).
- The post-sync refresh is **silent** (no loading blank, so a successful sync doesn't flash the card away).
- A live `deviceIdRef` + request-id guard prevents a slow or late load — including a post-sync refresh fired after the `deviceId` prop changed (this component is not remounted per device) — from writing the **wrong** device's vault into the panel.

`snapshotCount` is intentionally left as-is: this endpoint returns no count, so wiring a real one is a separate change (the fabricated `0` is pre-existing, not introduced here).

## Tests

`DeviceVaultStatus.test.tsx` (new): sync failure surfaces an error (not silent success); a failed load shows the error state, not "no vault"; `lastSyncStatus` maps to the badge (a completed vault is **not** "Never synced" — control-verified); `failed`/`null` normalization; Retry recovery; and device-switch-during-load / during-sync regression coverage. Locale keys added across all 7 locales.

This is a targeted fix for the vault-sync instance of #3531; the other sites the original audit cited (ApiKeys/Automations/Policies/Quarantined) already gained `response.ok` checks upstream since that audit.


ToddHebebrand (changes requested):
**Requesting changes — the backup-vault half is good; it's carrying an undisclosed RBAC change that needs to leave.**

**Must-fix 1 — scope bleed.** This PR's file list includes `apps/api/src/routes/roles.ts` (+186/-51), `apps/api/src/routes/roles.test.ts` (+327), `apps/web/src/components/settings/RolesPage.tsx` and `SsoProvidersPage.tsx`. That is the *entire* diff of #3577 — I compared the two file lists and the roles changes are byte-identical. Neither the title (`fix(web): surface swallowed backup-vault failures`) nor the body mentions it. An authorization-surface change should not ride into main inside a web-toast PR; it needs its own title, its own body, and its own review, which is exactly what #3577 already is. Please rebase this onto main and drop the roles half, leaving `DeviceVaultStatus.*` and the `backup.json` locale keys.

**Must-fix 2 and 3 — inherited from that half.** I've written both up in full on #3577, so just the headlines here: a focused partner admin 404s on every partner-scoped role (`fetchWithAuth` auto-injects `?orgId=`, `resolveActingScope` flips to organization scope, the `role.scope !== scopeContext.scope` guard rejects), and org-scope admins get a hard 403 creating any role (`RolesPage.tsx:204-206` sends `orgId` for org-token users; the POST handler 403s a non-partner caller's own org). Both land on #3577 once the split is done.

**The backup-vault work itself checks out.** I verified the badge mapping is real rather than fabricated — `apps/api/src/services/backup/vaultSyncPersistence.ts:172` does write `completed`/`failed`, so the status normalization reflects actual persisted state. The request-id and deviceId guards are sound and the i18n coverage is complete.

**Nice-to-have:** `DeviceVaultStatus.handleSync` hand-rolls its error surfacing instead of going through `runAction` (`apps/web/src/lib/runAction.ts`). The file sits outside the `no-silent-mutations` target globs so nothing fails, but a PR whose whole thesis is "swallowed mutation failures" is the one place that should adopt the sanctioned helper rather than route around it. See the catch pattern in CLAUDE.md.

**One ordering note for the rebase:** #3497 (Turkish `tr-TR`) merged a few minutes ago, so there are now 8 locales, not 7. Your `backup.json` keys will need a `tr-TR` entry too or you'll break the locale parity the merge just established.

Status: changes requested — split out the roles half, rebase, then this side is ready.

bdunncompany:
Rebased onto current main and split — the roles half is gone. This branch is now `DeviceVaultStatus.*`, the `backup.json` keys, and one API-side line. The RBAC work stays in #3577, which is where it should have been. You were right to bounce it; I branched this off my own #3577 worktree instead of a fresh `origin/main` one, so an authorization change rode inside a web-toast PR with a title and body that never mentioned it.

`tr-TR` keys are in — all 8 locales, real Turkish, not English placeholders.

**An adversarial pass over the split turned up something neither of us caught, and it changes what this PR is.**

`Sync Now` could not work from the UI at all. Both callers — `DeviceVaultStatus.handleSync` and `VaultDashboard.handleSync:124` — post with `{ method: 'POST' }` and no body, while `fetchWithAuth` sets `Content-Type: application/json` unconditionally (`stores/auth.ts:697`). The route used `zValidator('json', vaultSyncSchema)`, and Hono's json validator calls `c.req.json()` whenever the content-type is JSON, throwing `400 Malformed JSON in request body` **before the handler runs**. So `queueCommandForExecution` was never reached from either screen.

Nobody saw it because both callers swallowed it — one with `// Silently fail`, the other into `console.error` behind a generic message. That is the exact failure mode #3531 is about, sitting under the thing #3531 asked me to fix.

`lib/validation.ts:155-170` documents this precise case and names the `fetchWithAuth` interaction by name, so the fix is the sanctioned one: `optionalJsonValidator(vaultSyncSchema)`. `snapshotId` was already `.optional()`. Same schema, still 400s on malformed JSON and schema violations; the deliberate difference is that an absent/empty/whitespace-only body becomes `{}`.

Control-verified rather than asserted: with the strict validator the new API test gets **400**, with `optionalJsonValidator` it gets **200** and queues `VAULT_SYNC`.

To be precise about scope — the *route* was never universally broken. An API client sending `{}` or a real `snapshotId` always worked. It was unreachable from the two UI paths specifically.

**Three more real defects in the same card, same class:**

- A failed load for device B left device A's vault on screen, including a live Sync button still bound to A's `vault.id`. Once the 400 above is fixed, that clicks a sync at the **wrong device**. The parent doesn't key this component per device, so the effect now clears the previous vault first.
- Guarding an in-flight sync on the deviceId *string* is not enough: A → B → A returns to the same id but is a different visit, so a stale failure could post its error over a live sync and its `finally` could clear the running spinner. Now a monotonic visit counter.
- The card claimed `0 snapshots stored` for every vault, because `toVaultResponse` never returns `snapshotCount`. Replaced that fabricated number with the vault's real `lastSyncError`, which the API does return and nothing was displaying.

I also took the `runAction` note — it wasn't really optional, `CLAUDE.md:127` makes it the rule, and it's what catches a 200 `{success:false}` body that `response.ok` passes as success.

**Deliberately not in this PR**, raising separately rather than widening it again: the vault list neither filters `isActive` nor orders rows, so a device with two vaults shows an arbitrary one; there is no refresh after the agent reports, so the badge sits on `Syncing` until remount; and the four status labels are hard-coded English, bypassing i18n.

**Verification** — node 22.23.2 (CI `.nvmrc`), pnpm 10.34.5, on current main:

- `api` `src/routes/backup/` — 19 files, **229/229**
- `web` `DeviceVaultStatus.test.tsx` — **15/15**
- `web` `src/lib/i18n/` — **105/105**
- `web` `astro check` — 0 errors, 0 warnings
- `web` eslint on both changed files — clean

Every new test was control-run against the unfixed code first and fails there.

ToddHebebrand (approved):
**Approved — the split is genuinely done, and the thing you found while splitting is the most valuable part of this PR.**

**Scope bleed is gone.** File list is now 12: `apps/api/src/routes/backup/vault.{ts,test.ts}`, `DeviceVaultStatus.{tsx,test.tsx}`, and 8 `backup.json` locales. Zero `roles.ts`, `roles.test.ts`, `RolesPage.tsx`, `SsoProvidersPage.tsx`. The authorization surface is back in #3577 where it gets its own review. Worth stating plainly: branching from your own feature worktree instead of a fresh `origin/main` is how an RBAC change ends up inside a web-toast PR, and it's a cheap habit to fix.

**The Sync Now defect is real and I verified the whole chain.** `vault.ts:210` used `zValidator('json', vaultSyncSchema)`; `stores/auth.ts:697` sets a JSON content-type when the caller sets none; both callers POST bodiless. Hono's json validator calls `c.req.json()` on that content-type and throws `400 Malformed JSON in request body` before the handler — so `queueCommandForExecution` was never reached from either screen. Sync Now has never worked from the UI.

**And the fix is the right shape rather than a local patch.** You used the existing `optionalJsonValidator` (`apps/api/src/lib/validation.ts:155-170`, already in `routes/devices/core.ts` and `software.ts`) instead of hand-rolling one. Because it's server-side it repairs **both** callers — `DeviceVaultStatus.handleSync` and `VaultDashboard.tsx:124` — and a body that *is* sent stays schema-validated. That it went unnoticed for so long is exactly #3531's thesis: one caller swallowed it with `// Silently fail`, the other into `console.error` behind a generic message.

**Verified rather than assumed:** `toVaultResponse` (`vault.ts:323`) really returns `vaultType`/`lastSyncStatus`, and the only values ever persisted are `pending` (`vault.ts:243`), `failed` (`:260`) and `completed`/`failed` (`services/backup/vaultSyncPersistence.ts:172,189`) — all handled, no unhandled state. The `visitRef` monotonic counter closes the A→B→A case a bare `deviceIdRef` would miss. `runAction` adopted with the correct `ActionError`/401 catch. All 8 locales present with real tr-TR.

**Two nice-to-haves, neither blocking:**
1. `DeviceVaultStatus.tsx` isn't in `TARGET_GLOBS` in `apps/web/src/lib/__tests__/no-silent-mutations.test.ts`, so the `runAction` adoption isn't locked against regression — one line, and this PR is the reason to add it.
2. The body is now stale on its own diff: it says `snapshotCount` is "left as-is" (you removed it for `lastSyncError`), says 7 locales (it's 8), and never mentions the validator change — which is the most important fix in here.

Merging.

Issue #3578, opened 2026-08-15
[Agent] Software deployments show "Pending" with zero feedback for up to 45 minutes — no in-progress state reported


ToddHebebrand:
## Description

A software deployment to an online device shows **Pending** in the UI from dispatch until the agent posts its final result — which can be up to ~45 minutes later (15-min download timeout + 30-min install timeout in `InstallSoftware`). During that whole window there is no signal distinguishing "agent is downloading a 161 MB installer" / "installer is running" from "the command was lost." 

Observed in production (US, 2026-08-15): a 161 MB Printix installer deployed from the new DO Spaces upload storage sat on Pending for ~30 minutes before the result (`installer exited with code 1`) landed. Investigating required going all the way to prod DB rows and agent logs to establish that the pipeline was actually healthy — the UI gave no indication anything was in flight.

## Root Cause

The agent's `software_install` handler (`agent/internal/remote/tools/software_install.go`) is fully synchronous: download → checksum → execute → report once at the end. `deployment_results.status` goes straight from `pending` to `completed`/`failed`; `started_at` is only stamped retroactively from the final result payload. Neither transport (WS `sw-install-*` command results in `apps/api/src/routes/agentWs.ts`, HTTP result route in `apps/api/src/routes/agents/commands.ts`) carries an intermediate state.

## Proposed Fix

1. **Agent**: emit an in-progress result event when `InstallSoftware` begins (and optionally at the download→install transition), carrying the same `sw-install-<deployment>-<device>-<attempt>` id. Keep it fire-and-forget — losing a progress event must not affect the install.
2. **API**: extend `applySoftwareInstallResult` (`apps/api/src/services/softwareDeploymentResult.ts`) to accept a progress status that sets `deployment_results.started_at` (and a new `downloading`/`installing` status value or a `progress` column) without touching terminal-state reconciliation. The existing pending-only + retryCount guards keep late/duplicate progress events a no-op.
3. **UI**: render the intermediate state ("Downloading… / Installing…") instead of "Pending" once `started_at` is set.

Older agents that never send progress events keep today's behavior — the change must be backward compatible in both directions.

## Affected Files

- `agent/internal/remote/tools/software_install.go` (primary — progress reporting)
- `agent/internal/heartbeat/handlers_software_install.go`
- `apps/api/src/services/softwareDeploymentResult.ts`
- `apps/api/src/routes/agentWs.ts`, `apps/api/src/routes/agents/commands.ts`
[...]

Issue #2536, opened 2026-07-15
RDS: helpers fail to spawn on RD Session Host sessions (AssignProcessToJobObject access denied)


ToddHebebrand:
## Summary

On a real **RD Session Host** (Windows Server with the Remote Desktop Session Host role — i.e. an actual terminal server, the target environment for the helper-lifecycle work), the agent **fails to spawn per-session helpers for every RDS session**. This was found by running branch `ToddHebebrand/helper-sessions-in-remote-desktop` (PR #2520) as the real `BreezeAgent` service on a Server 2022 RDS host with 6 concurrent RDP sessions plus the console.

Every helper for an RDS session fails with, repeating each reconcile cycle:
```
lifecycle: failed to spawn helper  helperKey=2-system  error="assign helper to job: AssignProcessToJobObject: Access is denied."
lifecycle: failed to spawn helper  helperKey=3-system  ...
   ... 2..7, both system and user roles ...
```

The **console** session's helpers spawn and connect normally (with the correct session-scoped admission key `identity=windows:S-1-5-18:session:1`). Only sessions created under the RD Session Host role fail.

## Root cause

- The helper Job Object is created with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` (`agent/internal/sessionbroker/helper_job_windows.go:28`).
- Helpers are launched with `CREATE_SUSPENDED | CREATE_NO_WINDOW | CREATE_UNICODE_ENVIRONMENT` and **no** `CREATE_BREAKAWAY_FROM_JOB` (`agent/internal/sessionbroker/spawner_windows.go:390,442`).
- An RD Session Host places each session's processes in the **session's own job**. A process already in a (non-breakaway) job cannot be assigned to a second job, so `AssignProcessToJobObject` returns `ERROR_ACCESS_DENIED`.
- Job assignment is **mandatory / fail-closed** (`spawner_windows.go:266-276`): on assign failure the suspended helper is terminated and `Spawn` returns an error. The lifecycle manager retries every reconcile (~30s), so it loops forever and **no helper ever spawns** on RDS sessions.

## Impact

- **This is a regression introduced by this branch.** `main` has no Job Object at all (`helper_job_windows.go` does not exist on `main`); on `main`, helpers spawn on RDS sessions (and then accumulate via the old SID-cap bug). This branch adds the Job Object, which now blocks helper spawning on RDS entirely.
[...]

PR #3584, opened 2026-08-16
fix(agent): EXE installers hang until the 30-minute install timeout

ToddHebebrand:
## Problem

Every EXE software deployment sat on **Pending** for 30 minutes and then failed with `installer exited with code 1`. MSI deployments worked fine. Found while verifying the hosted DO Spaces upload rollout — a 161 MB installer uploaded, downloaded, and checksummed correctly, then stalled.

Root cause: `executeInstaller` used `cmd.CombinedOutput()` with no `cmd.WaitDelay`. With pipe-backed stdout/stderr, `Wait` returns on pipe EOF, not on child exit — so it blocks until **every process that inherited the handles** exits. EXE installers are typically wrappers: they extract, spawn the real setup, and exit. The grandchild holds the pipe, so the agent waited out the full `installTimeout` and killed the wrapper. The kill was then mislabeled: a killed process reports exit code 1 on Windows, so timeouts were indistinguishable from genuine installer failures.

MSI was never affected — `msiexec /qn` installs through the MSI service and leaves no pipe-holding descendants, exactly matching the field report.

Every other exec site in the repo (`scripts/runner.go`, `executor/executor.go`, `collectors/command_limits.go`) already sets `WaitDelay`. The installer path was the one that missed it.

## Changes

Extracted `runInstallerCommand` as the single path for installer child processes:

- **`cmd.WaitDelay` (10s)** so `Wait` returns once the installer itself exits, even when abandoned descendants hold the output pipes. `exec.ErrWaitDelay` is judged by the real exit code rather than reported as a failure.
- **Honest timeout labeling** — a genuine timeout now reports `installer timed out after 30m0s and was terminated`. Critically, the deadline having passed is *not* on its own a timeout: a wrapper can exit 0 and have `Wait` drain descendants past the deadline, where the deadline killed nothing. Only claim a timeout when the process did not finish on its own with a success code, otherwise a good install inside the last 10s of the window is reported as a 30-minute hang.
- **Detection settling** — with the early return, the wrapper has exited while the real installer may still be running, so a single-shot detection check produced false `detection rule was not satisfied` failures. `descendantsPending` now drives a bounded re-check (60s, 2s interval) that returns as soon as the rule is satisfied. The ordinary path stays single-shot and unchanged in timing; unsupported-on-platform still means "detection not performed" and returns immediately.
- **Process-tree kill on genuine timeout** — Windows via a Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` (following the existing `sessionbroker/helper_job_windows.go` pattern), unix via `Setpgid` + signaling the group. Reachable from exactly one branch: deadline fired *and* the process did not finish successfully. `release()` clears kill-on-close before closing the job handle — the reverse would kill the descendants a healthy wrapper legitimately left running, reintroducing the bug this PR fixes.
- **`installDMG` routed through the same runner**, deleting a divergent hand-rolled third copy of the exit-code logic. macOS DMG installs previously kept the original bug in full. `hdiutil`/`cp` route through with an empty `fileType` so they gain WaitDelay and timeout labeling without inheriting the 3010/1641 reboot-code semantics, which are meaningless for a mount and a copy; their `descendantsPending` is dropped, since a lingering `diskimages-helper` is normal mount state rather than an install still landing.

`Start`/`Wait` replaces `CombinedOutput` so the process can be adopted by the tree the instant it exists, with a mutex-guarded buffer for merged output — required because WaitDelay lets `Wait` return while an abandoned descendant's copy goroutine is still writing.

## Testing

Written test-first; every regression test was observed failing for the right reason before the fix:

- `runInstallerCommand blocked on descendant process for 8.02s` — the core hang.
- `expected timeout to be labeled as such, got err=installer exited with code -1` — the mislabel.
- `expected success for a wrapper that exited 0, got err=installer timed out after 30m0s` — the late-success inversion.
- `installDMG blocked on a descendant for 8.11s` / `expected a timeout label, got pkg installer exited with code -1` — the same two bugs surviving in the DMG path.
- `descendant survived the install timeout and kept running` — the missing tree kill.

Guard tests cover the cases that must **not** change: real non-zero exits still report their true exit code, the ordinary detection path stays single-shot, unsupported detection returns immediately, and a healthy wrapper's descendants are left alone on success.

Verification: `go test -race ./...` green across the whole agent; `go vet ./internal/...` clean; `GOOS=windows` vet (including test files) plus windows/linux/darwin builds all pass.

## Known limitations

- **The Windows job-object path was never executed** — this was developed on macOS, and both new test files skip on Windows. It was validated by `GOOS=windows go vet` (which compiles the tests) and by conforming structurally to the in-production `sessionbroker` job pattern. **Worth a real-device check before this reaches the fleet.**
- `os/exec` doesn't expose the child handle, so the process is reopened by pid after `Start`; a descendant spawned inside that microsecond window would be outside the job. Documented in the code.
- Job assignment degrades rather than fails (a process already in a job that forbids breakaway cannot join a second one, per the `#2536` precedent) — the cost is the tree kill, never the install.

## Related

- #3578 — deployments show "Pending" with no feedback for up to 45 minutes. This PR removes the most common *cause* of the long wait, but the missing in-progress reporting is still worth fixing separately.
- The original Printix failure that surfaced this is a separate, device-specific issue (the installer needs `api.printix.net` reachability, which fails on that machine).

### DIFF
--- a/agent/internal/remote/tools/software_install.go
+++ b/agent/internal/remote/tools/software_install.go
@@ -1,6 +1,7 @@
 package tools
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
@@ -16,6 +17,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -610,14 +612,33 @@ func executeInstaller(localPath, fileType, silentInstallArgs string) (int, strin
 
 	case fileType == "dmg" && runtime.GOOS == "darwin":
 		// Mount, find .app or .pkg, install, unmount
-		exitCode, output, err := installDMG(ctx, localPath)
-		return exitCode, output, false, err
+		return installDMG(ctx, localPath)
 
 	default:
 		return 1, "", false, fmt.Errorf("unsupported file type %q on %s", fileType, runtime.GOOS)
 	}
 
 	return runInstallerCommand(ctx, cmd, fileType)
+}
+
+// lockedBuffer collects an installer's merged stdout+stderr. CombinedOutput's
+// plain bytes.Buffer is not safe here: with WaitDelay, Wait can return while the
+// copy goroutine of a descendant that still holds the pipe is writing.
+type lockedBuffer struct {
+	mu  sync.Mutex
+	buf bytes.Buffer
+}
+
+func (b *lockedBuffer) Write(p []byte) (int, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.Write(p)
+}
+
+func (b *lockedBuffer) Bytes() []byte {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return append([]byte(nil), b.buf.Bytes()...)
 }
 
 // runInstallerCommand runs one installer command. Its third return says the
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
Follow-up filed: #3601 — the Windows process-tree kill added here has never been executed on Windows (developed on macOS; the tree-kill test skips there). CI proves it compiles and that `Test Agent (Windows)` is green, not that the job object contains or releases a real installer's descendants correctly.

Gated on the **agent fleet promote** rather than the tag, since that's the step that puts it on customer endpoints.

PR #3585, opened 2026-08-16
fix(api): back manual software-remediation with a durable single-use authorization (#3553)

### LINKED ISSUE #3553: [API] Remediation trigger:'manual' is authorization state carried in forgeable BullMQ job data
## Description

Residual hardening item from #3543 (PR #3548), out of scope there because it needs a new table.

After #3548, the software remediation worker re-checks policy arming (`enforceMode` + `autoUninstall`) as defense in depth, but jobs stamped `trigger: 'manual'` (the audited manual-remediate route on detect-only policies) bypass that re-check by design. The trigger field lives in BullMQ job data — anything with Redis write access can enqueue a job claiming `trigger: 'manual'` and drive uninstalls on an unarmed policy.

Redis write access is already a serious compromise, so this is hardening, not an open exploit — but authorization state should not ride in unauthenticated queue payloads.

## Proposed Fix

A durable one-time remediation-request record (DB row created by the authenticated manual route, consumed atomically by the worker; job data carries only the record id). Worker refuses `manual` jobs without a matching unconsumed record.

## Affected Files

- `apps/api/src/jobs/softwareRemediationWorker.ts`
- `apps/api/src/routes/softwarePolicies.ts` (manual remediate route)
- new table + migration (with RLS/cascade/export registration per CLAUDE.md)

## Reported By

PR review during #3548; relates to #3543.

bdunncompany:
## What

Fixes #3553. A software-remediation job's `trigger: 'manual'` is authorization state — the worker skips the arming gate (`enforceMode`/`autoUninstall`) when it's set. That flag rode in BullMQ job data (Redis, no authentication of its own), so anything with Redis write access could enqueue a forged `{trigger:'manual'}` job and drive uninstalls on an unarmed policy.

## Change

A durable, single-use authorization record now backs every manual claim. The MFA-gated remediate route mints one row per `(policy, device)` it schedules and passes the row id in the job data; the worker consumes it atomically and honors `manual` only against a matching, unconsumed, unexpired, **ownership-coherent** row — otherwise it downgrades to `auto` and applies the arming gate (fail-closed).

- **New table `software_remediation_requests`** (dual-axis tenancy, FKs `ON DELETE CASCADE`, owner CHECK, forced RLS with dual-axis policies), registered in `DUAL_AXIS_TENANT_TABLES`.
- **Commit-before-enqueue**: records are committed in a system-context transaction *outside* the request's long transaction, so a fast worker can't consume-before-commit and wrongly refuse a legitimate run.
- **Ownership-coherent, race-free consume**: a single atomic `UPDATE` folds the current device+policy ownership check into a SQL `EXISTS`, and the device row is `FOR UPDATE`-locked for the worker's transaction. This closes the device-move retenanting hole (the generic device-move trigger rewrites `org_id`) with no Read-Committed TOCTOU window.
- **Manual bypasses the auto-cooldown** (a deliberate operator override; also prevents burning the single-use token on a cooldown skip).
- **Never overrides `audit_mode`** — a policy flipped to audit mode after authorization is not bypassed.
- **Trusted audit attribution**: the `manual_override` audit uses the requester from the consumed row, not the forgeable job-data field.
- **Manual jobs key on their authorization id**, so a manual request can't dedupe into an older auto job and lose its authorization.

## Known limitation

The lock guarantees ordering through command creation + commit, not command *delivery*: a command authorized just before a device move can still be delivered after it. A "never execute after retenanting" invariant would need dispatch-time tenant binding across all command delivery — out of scope here.

## Tests

- Unit (`softwareRemediationWorker.test.ts`): verified-manual override; forged/unverified manual downgraded and refused on an unarmed policy; record-not-consumed refused; `audit_mode` never overridden; trusted-requester attribution (with a forged job-data value); manual bypasses cooldown; scheduler mints + manual-specific job id.
- Integration (`softwareRemediationRequestsRls.integration.test.ts`, PostgreSQL): cross-partner + partner-axis RLS forge, device-move refusal (real fresh read), single-use consume.

Adversarially reviewed across four Codex passes (design + three diff passes); the design pass caught a cross-tenant device-move hole and a transaction race before any migration was written.


bdunncompany:
Merge-order heads-up: #3587 touches the same two files this PR does — `apps/api/src/services/tenantCascade.ts` and `apps/api/src/services/tenantExportPolicyRegistry.ts` — so whichever lands second will conflict there.

Worth flagging that it is **not** a "keep both added lines" resolution, because #3587's changes to those files aren't flat additions:

- In `tenantCascade.ts` it adds an **order-dependent** pre-clear sequence (its own comment: *"Order below is load-bearing: results, then deployments, then versions"*). That ordering has to survive the merge intact.
- In `tenantExportPolicyRegistry.ts` it **modifies the existing** `software_deployments` entry (adds `install_method_id` to `included`) rather than adding a new key — easy to lose if the conflict is resolved by naively keeping both sides.

This PR's own edits to those files are single additive entries (`software_remediation_requests` in the org-cascade order and one `tablePolicy` line), and they don't participate in that FK ordering, so they can sit alongside without semantic interaction — the risk is purely that #3587's ordering and its amended entry get clobbered during resolution.

I'm happy to do the rebase and resolve it carefully whichever way you sequence them — just say the word once one of them lands.

PR #3605, opened 2026-08-16
fix(web,api): script test-run recovery, stale device pin, and AI tool descriptions

ToddHebebrand:
Follow-ups from the post-merge review of #3582. Four findings, ranked by consequence.

## 1. Terminal poll failures left a permanently spinning chip — and the only escape was a second run

Both give-up exits from `pollExecution` (permanent poll error, deadline fall-through) set `phase='idle'` but left `execution.status` at `pending`/`queued`/`running`. The animated `Loader2` chip spun forever, the output block stayed hidden, and the Run button was simultaneously re-enabled — so the only affordance left was **Test Run, which starts a second execution on a real customer device**.

Both exits now clear `execution` and stash the id for a **Retry** control that re-enters `pollExecution` with the *same* execution id. Recovery never means a second run.

## 2. 408/429 were treated as permanent

The "4xx is permanent" branch swallowed 429. `globalRateLimit()` is mounted app-wide and a 2s poll loop can plausibly trip it, which aborted the poll and showed a wrong "deleted or access revoked" message. 408 and 429 now fall through to the retry path alongside 5xx.

## 3. A pinned device survived an OS-target change

`compatibleDevices` recomputed on `osTypes` change but `selectedDeviceId` was never reconciled. Removing an OS target after pinning a device left the `<select>` rendering as unselected (no matching `<option>`) and suppressed the offline warning — while Run stayed enabled and `handleRun` still POSTed the hidden device id. Executing on a machine the operator believes is deselected is the wrong failure direction.

Now reconciled: selection, localStorage pin, and `onTestDeviceChange(null)`. The effect guards on the raw `devices` list being empty (not `compatibleDevices`) so it can't fight the restore-from-localStorage effect during the load window.

## 4. The executionId filter was unguarded — verified by mutation

Deleting `eq(scriptExecutions.id, input.executionId)` from the handler's `.where(and(...))` left **all 6 tests green**, while the tool would return an arbitrary execution row (`.limit(1)` over an unfiltered join).

The existing `flattenSql` helper was the right instinct and already catches a deleted *org* condition — it just wasn't applied to the id predicate. The happy-path test now asserts the execution id appears in the clause's **bound params**, adjacent to an `id` column.

Mutation-verified twice (once by the implementing agent, once independently):
- predicate present → 6 passed
- predicate deleted → `AssertionError: expected [] to include '1111...'`, 1 failed / 5 passed
- restored → 6 passed

Also renamed a test whose name claimed `cross-org lookup` coverage it did not have (it mocks `[]` and only exercises the no-row branch).

## 5. Four AI-facing tool descriptions promised a recovery path that does not exist

`aiToolsScripts.ts`, `aiAgentSdkTools.ts`, `scriptBuilderTools.ts`, and `scriptBuilderPrompt.ts` all told the model to use `get_script_execution` to read a run that was still running when `run_script` returned. **That path does not exist.** Verified chain:

1. `run_script` waits via `waitForCommandResult(commandId, 60000)` (`aiToolsScripts.ts:328`)
2. at its deadline that terminalizes the command row to `status:'failed'` (`commandQueue.ts:588-605`)
3. the agent's real result then fails the `status IN ('pending','sent')` filter at the **lookup** in `agentWs.ts:1598-1610` — earlier than the CAS at `:1739` — and falls into `processOrphanedCommandResult`, which handles only SNMP/discovery/tunnel and returns
4. so `handleScriptResult` (the only writer of stdout/stderr/exitCode onto `script_executions`) never runs
5. `staleCommandReaper` later stamps status but explicitly does not persist output, and attributes it to the misleading #3097 "delivered but never recorded" message

Net: in exactly the scenario those strings named, the tool returns `status:'failed'`, `exitCode:null`, `stdout:null` permanently — instructing the model to attempt an impossible recovery **and to report a successful script as failed**. Since these are prompts the model acts on, this was a functional bug, not a wording nit.

The strings now describe the tool's real use: reading a run started outside the current tool call (the editor's Test Run button, or an id from `get_script_execution_history`).

## Out of scope — needs a separate issue

The underlying defect is untouched here: **a `run_script` call exceeding the 60s wait permanently loses the agent's real output.** Fix directions are either letting the WS path accept a late result for a `timeout`-flagged command, or making `handleScriptResult` reachable from the orphaned/late path. That touches the agent WS result path and deserves its own PR.

## Testing

- `apps/web` scripts components: 82 passed (7 files) — includes 4 new `ScriptTestRunner` cases covering the 404-stop + retry-same-id, 429-continues-polling, deadline, and OS-retarget-clears-selection paths
- `apps/api` script tools: 126 passed (12 files), `scriptBuilderTools.guard.test.ts` 36/36
- New polling tests use `vi.useFakeTimers()`, which is what makes the deadline path testable at all; the pre-existing polling tests were left on real timers (converting them is a cheap follow-up now that the helper exists)
- No new locale keys: the Retry button reuses `common:actions.retry`, already translated in all 7 locales. Seeding English-valued keys would have reddened `translationCoverage.test.ts`, which asserts an exact count of English-identical leaves per locale.

No schema, migration, or tenancy changes.

### DIFF
--- a/apps/api/src/services/aiAgentSdkTools.ts
+++ b/apps/api/src/services/aiAgentSdkTools.ts
@@ -1146,7 +1146,7 @@ export function createBreezeMcpServer(
 
     tool(
       'get_script_execution',
-      'Fetch one script execution by ID with status, exit code, stdout, and stderr. Use to read the result of a run that was still running when run_script returned.',
+      'Fetch one script execution by ID with status, exit code, stdout, and stderr. Use for runs started outside the current tool call (the script editor Test Run button, or an execution id from get_script_execution_history) — not to re-check a run_script call that already returned.',
       {
         executionId: uuid,
       },
--- a/apps/api/src/services/aiToolsScripts.getExecution.test.ts
+++ b/apps/api/src/services/aiToolsScripts.getExecution.test.ts
@@ -3,9 +3,13 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 /**
  * Coverage for the get_script_execution tier-1 tool (script-editor test loop):
  * single-execution lookup with the script-org condition applied and the site
- * axis enforced after the row loads. The tool exists so the AI can read a
- * run's output after the user's editor Test Run (or after a run outlives the
- * 60s tool window), so the shape of the returned JSON is part of the contract.
+ * axis enforced after the row loads. The tool exists so the AI can read the
+ * output of a run started outside the current tool call (the editor's Test Run
+ * button, or an id from get_script_execution_history), so the shape of the
+ * returned JSON is part of the contract. It is NOT a recovery path for a
+ * run_script that hit its 60s wait: that timeout terminalizes the command row,
+ * so the agent's late result loses the CAS in routes/agentWs.ts and never
+ * reaches handleScriptResult — stdout/exitCode stay null forever.
  */
 
 vi.mock('../db', () => ({
@@ -17,6 +21,7 @@ vi.mock('../db', () => ({
 
 import { eq } from 'drizzle-orm';
 import { db } from '../db';
+import { scriptExecutions } from '../db/schema';
 import { registerScriptTools } from './aiToolsScripts';
 import type { AiTool } from './aiTools';
 import type { AuthContext } from '../middleware/auth';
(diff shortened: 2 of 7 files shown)


ToddHebebrand:
Follow-ups from this review are now split out:

- **#3607** — the underlying defect this PR deliberately left unfixed: a `run_script` result arriving after the 60s wait deadline permanently loses the agent's real output. Includes the verified drop chain (it fails the row **lookup** at `agentWs.ts:1598-1610`, not the CAS at `:1739` — so a CAS-only patch would not fix it) and notes the same filter is duplicated on the HTTP path at `agents/commands.ts:44`.
- **#3610** — stacked on this branch: the device-picker 500-row cap, the silent `/devices` failure, regression tests for the two untested #3582 fixes (`ExecutionHistory`, `ScriptExecutionsPage` — neither had any test file), and the stale design-spec claims.

Worth flagging from #3610, since it lands on code from this PR: the pin-reconcile effect added here treats "the pin isn't in the fetched list" as proof the pin is incompatible, and `localStorage.removeItem`s it. On a fleet over 500 (the picker's page cap) or any failed fetch that's false, so a legitimate pin was being deleted for good. Mutation-testing also showed the effect read `devicesState` as stale `'loaded'` on the render where `osTypes` changes, clearing the pin against a device list belonging to the old targets. #3610 binds the fleet data to the OS-target key it was fetched for.

Issue #3607, opened 2026-08-16
run_script: a result arriving after the 60s wait deadline permanently loses the agent's real output


ToddHebebrand:
A `run_script` (or any dispatched command) whose agent result arrives after the
60s wait deadline **permanently loses the real output**. The execution is
reported to the operator and to the AI as `failed / exitCode:null / stdout:null`
even when the script succeeded on the device.

Split out of #3605, which only corrected the AI-facing tool descriptions that
told the model to recover such a run via `get_script_execution` — a path that
does not exist. This issue is the underlying defect.

## Verified chain

1. `run_script` waits on `waitForCommandResult(commandId, 60000)`
   (`apps/api/src/services/aiToolsScripts.ts:328`).
2. At its deadline that call **terminalizes the command row** to
   `status:'failed'`, `result:{status:'timeout'}`
   (`apps/api/src/services/commandQueue.ts:588-609`). The update itself is
   correctly guarded on `status IN ('pending','sent')`, so it only fires when the
   result genuinely has not landed yet.
3. The agent's real result then arrives over the WS. The **row lookup** filters
   on `inArray(deviceCommands.status, ACCEPTED_COMMAND_RESULT_STATUSES)` where
   that constant is `['pending','sent']`
   (`apps/api/src/routes/agentWs.ts:87`, filter at `:1598-1610`). The row is now
   `failed`, so the lookup returns **no row at all**.

   > This is one step earlier than it first appears. There is an identical
   > status filter on the CAS update at `agentWs.ts:1739`, and it is tempting to
   > blame the drop on a lost CAS. It never gets that far — `command` is
   > undefined and the code branches away before the CAS is reached. A fix
   > applied only at the CAS would not work.

4. With no command row, the handler falls into `processOrphanedCommandResult`
   (`agentWs.ts:858`, dispatched at `:1565` and `:1650`), which handles only
   SNMP / discovery / tunnel result types and returns for everything else.
5. So `handleScriptResult` — the **only** writer of `stdout`, `stderr` and
   `exitCode` onto `script_executions` — never runs.
6. `staleCommandReaper` later stamps a status on the execution but explicitly
   does not persist output, and attributes it to the misleading #3097
   "delivered but never recorded" message.

Net: the output is gone from the database. It is not late, not degraded —
it is never written, and there is nothing to re-read afterwards.

## Impact

- **AI**: `run_script` on any script slower than 60s reports a successful run as
  failed, with no output. The model then "fixes" a script that was already
[...]

PR #3610, opened 2026-08-16
fix(web,docs): device picker page cap, silent fleet-load failure, and missing regression tests

ToddHebebrand:
Second batch of follow-ups from the post-merge review of #3582 — the findings
#3605 deliberately left open.

> **Stacked on #3605** (`base: script-test-loop-followups`), because it isn't
> separable: the pin-reconcile effect this PR guards does not exist on `main`.
> Per CLAUDE.md a PR based on a sibling branch triggers **no CI**, so CI was
> hand-dispatched — see the Testing section. Rebase onto `main` after #3605
> merges.

## 1. The device picker silently truncated any fleet over 500

`ScriptTestRunner` fetched a bare `/devices`. That endpoint defaults to
`DEVICES_LIST_DEFAULT_LIMIT = 500` with a hard max of 1000
(`apps/api/src/routes/devices/cursor.ts:29,39`), and with no `page`/`cursor` it
returns the first page by `last_seen_at DESC`. On any fleet over 500 devices the
rest simply could not be picked for a test run — no message, no indication, just
an incomplete `<select>`.

It now issues **one server-filtered request per OS target**
(`?osType=…&limit=1000`). The cap applies per-OS to an already-compatible set
instead of fleet-wide to a set we then throw most of away, which in practice
takes the cliff well past any reachable fleet size.

## 2. A failed fleet fetch was indistinguishable from "no devices"

Both failure exits (`!response.ok` and the `catch`) returned silently, leaving
`devices` empty. The picker then rendered **"No compatible devices"** — which
reads as a settled, correct answer. The operator would go looking for a device
that was in fact right there, or conclude their OS targets were wrong.

The picker now says the list failed to load. One new locale key, genuinely
translated in all 7 catalogs (no new English-identical leaves, so
`translationCoverage.test.ts` stays green).

## 3. #3605's pin reconcile was destructive on exactly those paths — and had a latent staleness bug

This is the finding that made the two above worth more than cosmetic.

#3605 added an effect that clears a pinned device the current OS targets no
longer allow, including `localStorage.removeItem`. But "the pin isn't in the
fetched list" was being treated as proof the pin is incompatible. On a >500-device
fleet, or any failed fetch, it isn't — and the operator's pin was **deleted for
good**.

While writing the regression test for that, mutation-testing surfaced a second,
sharper bug: with `devices`/`devicesState`/`devicesTruncated` as three separate
`useState` values, the render where `osTypes` changes still reads `devicesState`
as `'loaded'` from the *previous* fetch. The reconcile therefore ran against a
device list belonging to the old OS targets and cleared the pin anyway — the
guard was there but read stale.

Fixed by holding the fleet in **one state object carrying the OS-target key it
was fetched for**:

```ts
{ key, status: 'loading' | 'loaded' | 'error', devices, truncated }
```

The reconcile now fires only when `fleet.key === osTypesKey`, `status === 'loaded'`
and `!truncated`. Correlated state that must be read together is no longer
splittable, so the staleness class is gone rather than patched.

On truncation the pin is deliberately **kept**: a stale pin stays visible and
selectable in the `<select>`, whereas deleting it destroys the operator's choice
permanently. Wrong-but-visible beats silently-erased.

## 4. Two #3582 fixes shipped with zero test coverage

`ExecutionHistory.tsx` and `ScriptExecutionsPage.tsx` had **no test files at
all**, so three merged fixes were unguarded. Adding `ExecutionHistory.test.tsx`
(7 tests) and `ScriptExecutionsPage.test.tsx` (5 tests):

| Fix | What a revert does to the user |
|---|---|
| null-safe search filter (`ExecutionHistory.tsx:113-114`) | typing in the search box throws on any row with a missing `scriptName`/`deviceHostname` and blanks the entire executions table |
| null-safe sort comparators (`:152`, `:155`) | clicking the Script or Device column header does the same |
| details-modal record fetch (`ScriptExecutionsPage.tsx:108-127`) | the list endpoint omits `stdout`/`stderr`, so the modal shows permanently empty output panels — looks like the script produced no output, not like a bug |

## 5. Design spec corrected

`docs/superpowers/specs/web-ui/2026-08-15-script-editor-test-loop-design.md` had
two claims that #3605 disproved:

- §C claimed `get_script_execution` "closes the *run outlived the 60s tool
  window and there is no recovery* gap". It does not and cannot as built — the
  output is never written, so there is nothing to read back.
- §D attributed the late-result drop to a lost CAS in `agentWs.ts`. It happens
  one step earlier, at the row **lookup** (`agentWs.ts:87`, `:1598-1610`), which
  matters because **patching only the CAS would not fix it**.

Both corrected in place with dated notes. The underlying defect is filed as
**#3607**.

## Testing

- `apps/web` scripts components: **98 passed** (9 files) — up from 82/7: two new
  files, +16 tests.
- i18n + `no-silent-mutations` guards: 203 passed. Lint clean. Typecheck clean
  (the lone `@tiptap/core` error is a stale-`node_modules` artifact of a local
  branch switch — the package is in `package.json`; CI installs fresh).
- **Every new assertion was mutation-verified**: each guard removed in turn, the
  corresponding test confirmed to fail, then restored. That process caught one of
  my own tests passing vacuously (the cached device was still compatible under
  the OS targets I'd chosen, so the guard never mattered) and the staleness bug in
  §3. It also found the pre-existing `clears a pinned device…` test was waiting on
  a signal that can't fail — a `<select>` whose value has no matching `<option>`
  already reports `''` in the DOM, which is the very bug it exists for. It now
  waits on the pin actually being dropped.

No schema, migration, or tenancy changes.


ToddHebebrand:
Filed **#3612** for the picker limitation this PR leaves open: the Test Run picker is still a native `<select>` with no search, so it stops being *usable* long before the page cap makes it *incorrect*.

Scoping that turned up a broader version of the same defect — `GET /devices` defaults to 500 and silently clamps anything over 1000, and ~13 `apps/web` call sites fetch device lists without accounting for it. Sharpest example: `ScriptsPage.tsx:95` requests `?limit=10000`, which the server rewrites to 1000 — the number itself shows the clamp already misled someone. #3612 has the full table.

Also noted there that the `fleet.truncated` guard added here must not be dropped as part of that work without re-reading why it exists — it's what stops an operator's pinned device being deleted from `localStorage`.

PR #3625, opened 2026-08-16
fix(i18n): translate the "/ea" unit in tr-TR and es-419 quote pricing

bdunncompany:
`contracts.contractPax8Drawer.priceEach` renders a per-unit price as `{{currency}} {{price}}/ea`. `/ea` is an English abbreviation for "each" — not a symbol, not an interpolation-only token — and two catalogs still ship it verbatim.

The other five locales all localize it, which is what makes this a gap rather than a house style:

| locale | value |
|---|---|
| fr-FR | `{{currency}} {{price}}/unité` |
| fr-CA | `{{currency}} {{price}}/unité` |
| de-DE | `{{currency}} {{price}}/Stück` |
| pt-BR | `{{currency}} {{price}}/un.` |
| it-IT | `{{currency}} {{price}}/cad` |
| **tr-TR** | **`{{currency}} {{price}}/ea`** |
| **es-419** | **`{{currency}} {{price}}/ea`** |

Now `/adet` (tr-TR) and `/u.` (es-419), the latter matching pt-BR's `/un.` abbreviation style.

### Why this also tightens the duplicate guard

Both were being counted as accepted exact-English duplicates, so the baseline was treating a real untranslated string as an approved literal. #3620 raised the tr-TR `billing.json` baseline to 18 on the reading that all 18 were intentional; this one was not. The baselines drop rather than rise:

- tr-TR `billing.json` 18 → 17
- es-419 `billing.json` 41 → 40

I went through all 18 tr-TR duplicates individually before concluding this was the only genuine miss. The rest are symbols (`✓`, `%`, `—`), interpolation-only format strings (`{{qty}} × {{price}}`, `{{succeeded}} {{verb}}`, `v{{number}}`), an example URL, and the cross-locale literals `MSRP`, `SKU`, `A4`, `Letter`, `PN`.

`PN` is the one I deliberately left alone: it is byte-identical in every locale including fr/de/pt/es/it, which reads as a deliberate cross-locale abbreviation rather than an omission. Happy to change it if the glossary says otherwise, but it did not look like the same class of defect as `/ea`.

Interpolation placeholders are untouched, so `localeParity`'s token check is unaffected.

### Verification

Local, node 22.23.2 (CI `.nvmrc`), pnpm 10.34.5:

- `apps/web` `src/lib/i18n/` — **105/105 pass** (localeParity, translationCoverage, terminologyQuality, extractionQuality, keyUsage)

Refs #3620, #3497

### DIFF
--- a/apps/web/src/lib/i18n/translationCoverage.test.ts
+++ b/apps/web/src/lib/i18n/translationCoverage.test.ts
@@ -80,7 +80,9 @@ const namespaceDuplicateBaselines = {
     // +1: order breakdown — "SKU" is a locale-invariant acronym.
     // +1: partnerBillingSettings.defaults.documentPageSizeA4 — "A4" is the
     // ISO 216 paper size code, identical in every catalog.
-    'billing.json': 41,
+    // 41 -> 40: `contracts.contractPax8Drawer.priceEach` is no longer a
+    // duplicate; its "/ea" was genuinely untranslated, not a literal.
+    'billing.json': 40,
     // +1: dashboard.vuln.kevCves_one — "{{count}} CVE" is a locale-invariant
     // acronym.
     // +8: PsaConnectionForm credential placeholders — literal token formats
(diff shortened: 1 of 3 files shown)


ToddHebebrand (approved):
**Approved — verified against `origin/main` rather than taking the table at its word, and it holds exactly.**

`contracts.contractPax8Drawer.priceEach` is `{{currency}} {{price}}/ea` in `en`, and `/ea` survives verbatim in precisely two catalogs — `apps/web/src/locales/tr-TR/billing.json:307` and `apps/web/src/locales/es-419/billing.json:307`. Every other locale localizes the unit (`/unité`, `/Stück`, `/un.`, `/cad`), so this is a gap and not house style. `/adet` and `/u.` are the right calls, and matching pt-BR's abbreviation style for es-419 is a nice touch.

**The part that matters more than the translation:** the baselines go *down* (tr-TR 18→17, es-419 41→40). A duplicate baseline that absorbs a genuinely untranslated string is worse than no baseline, because it converts a detectable gap into an approved literal — that's exactly what #3620 did here when it raised tr-TR to 18 on the reading that all 18 were intentional. Re-auditing the entries you inherited instead of trusting the number is the right instinct.

**One cosmetic nit, not worth a push on its own:** in `translationCoverage.test.ts` the tr-TR annotation now reads `+1:` above prose that still describes *two* bulk-result strings (`quotes.page.bulk.success` and `invoicesPage.bulk.success`, both `"{{succeeded}} {{verb}}"`). Both are still legitimately duplicates; the `-1` came from `priceEach`, which the comment does explain on its own line below. The enforced number 17 is correct — CI proves it — so only the `+2` → `+1` edit is off. Fold it into your next touch of that file.

Merging.

PR #3626, opened 2026-08-16
fix(web): make static device-group membership editable again (#3615)

### LINKED ISSUE #3615: [Web] Static device-group membership is no longer editable in the UI, though the API supports it (#3564 follow-up)
#3564 (merged) fixed the Device Groups edit 404 — the frontend was sending `PUT /device-groups/:id`, which has no handler; the route is `PATCH` (`apps/api/src/routes/devices/groups.ts:185`). It also removed the Description and Policy Assignment form controls, correctly: `device_groups` has no `description` column and no policy field (`apps/api/src/db/schema/devices.ts:410`), so the API was silently discarding both.

The same PR hid the static device chooser on edit, because `PATCH /:id` ignores `deviceIds` — offering the control there meant a "Save" that silently dropped membership changes. Hiding it is the right immediate call: a control that lies is worse than a missing one.

**But membership on an existing static group is now not editable from the UI at all**, and it should be. The capability exists on the API — membership is a separate pair of endpoints, `POST /:id/devices` (`groups.ts:339`) and `DELETE` (`groups.ts:426`) — it was simply never wired to the edit modal, which is why the create-time chooser was being reused and silently no-oping.

The fix is to wire the edit-mode chooser to those two endpoints rather than folding `deviceIds` into the PATCH body: diff the selection against current membership and issue the add/remove calls. Keeping membership on its own endpoints is the better shape anyway — it's the one part of a group that changes independently of the group's definition.

Low urgency, since the pre-#3564 behavior silently discarded these edits too, so nothing regressed. But it's a real gap in the UI against an API that already supports it.

**Also worth a cleanup pass:** the removed controls leave orphaned i18n keys across all 8 locales — `deviceGroupsPage.description`, `policyAssignment`, `noDescriptionProvided`, `noPolicyAssigned`, and `optionalDescriptionToHelpYourTeam`.

ToddHebebrand:
Closes #3615

## The gap

#3564 hid the static device chooser on edit, correctly: `PATCH /device-groups/:id` has no `deviceIds` field, so a "Save" that appeared to change membership silently dropped it. But the capability was on the API all along — membership lives on its own endpoints — and was simply never wired to the edit modal. Net effect: **static membership was not editable from the UI at all.**

## The fix

The chooser is offered on edit again, driven by the membership endpoints rather than folded into the PATCH.

- **Baseline** — `GET /device-groups/:id/devices` when the modal opens. Deliberately *not* the list row's `deviceIds`: the save diff turns "absent from the selection" into a `DELETE`, so a stale baseline would remove devices the user never deselected. Until that read lands — or if it fails — the chooser and the Save button are disabled and the error is shown. A control that misreports membership is exactly what #3564 was right to remove.
- **Save** — the selection is diffed against the baseline. Additions go to `POST /:id/devices` in one call; removals to `DELETE /:id/devices/:deviceId`, one per device (the API has no bulk-remove verb). The group definition is PATCHed first, so a dynamic→static conversion is committed before the membership calls — the API rejects those on a dynamic group.
- **Dynamic→static conversion** starts from an empty chooser. A dynamic group's members are the server's evaluation, not an authored list; re-POSTing them would forge manual memberships nobody asked for.
- The PATCH body is unchanged and still carries no `deviceIds`.

## Two adjacent defects found in the same path

Both were membership writes that never reached the server:

1. **Drag-and-drop between static groups** called `PUT /device-groups/:id` — the same no-handler 404 #3564 fixed for the form — carrying a `deviceIds` the update route ignores regardless. Only the optimistic local state ever changed. Now goes through the membership endpoints.
2. **The group list was fetched without `includeMemberships`**, so no static group ever carried `deviceIds`: every group card rendered as empty, and the drag handler's source-membership guard could never pass.

## i18n cleanup

Removes the five keys orphaned by #3564's control removal — `deviceGroupsPage.description`, `policyAssignment`, `noDescriptionProvided`, `noPolicyAssigned`, `optionalDescriptionToHelpYourTeam` — across all 8 locales, and adds the three new membership strings (`loadingGroupMembership`, `failedToLoadGroupMembership`, `failedToUpdateGroupMembership`). Verified unreferenced in source first; note `noDescriptionProvided` also exists under `longTail.snmp.SNMPTemplateEditor`, which is a distinct key and is untouched.

## Verification

- `DeviceGroupsPage.staticGroups.test.tsx` — 9 passed. The old "hides the chooser on edit" test is replaced by four covering: the chooser is seeded from the membership endpoint (not the list row), adds/removes go to the right endpoints with the PATCH carrying no `deviceIds`, an unchanged selection issues no membership calls at all, and a failed membership read disables the chooser and the Save.
- `DeviceGroupsPage.test.tsx` + `DeviceGroupsPage.dynamicGroups.test.tsx` — 10 passed.
- `localeParity` / `no-silent-mutations` / `no-envelope-fallthrough` — 165 passed.
- `tsc --noEmit` on `apps/web` — clean. No `.astro` files touched.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter.

**Findings:** 2 raised, both addressed in `552951375`; 0 outstanding.

- **Critical, flagged independently by both reviewers — cross-group race.** `fetchGroupMembership` was fire-and-forget with nothing tying its resolution to the group still being edited. Open Edit on group A → close → open Edit on group B, and A's late response installed *its* membership as B's baseline. No error, no loading state, nothing visibly wrong — and the next Save diffed A's device list against B and issued the adds/removes at B's endpoint. Devices moved between groups on request timing alone, with the save reporting success. Reads now carry a generation id; only the newest may write state.
- **Half-applied save.** Removals are one DELETE per device, so a mid-loop failure leaves the group partly changed. The error was surfaced, but the baseline was not — and `DELETE /:id/devices/:deviceId` 404s on a device that is no longer a member, so retrying the same selection failed forever on the already-removed device and never reached the legitimate remaining removals. A membership failure now re-reads the baseline (keeping the user's unsaved picks) and refreshes the list before surfacing the error; a failed re-read disables Save rather than leaving a stale baseline armed.

Both reviewers explicitly cleared the rest: partial-failure surfacing is honest (modal stays open, error shown, no `handleCloseModal`), `DeviceGroupsPage.tsx` is already on `runActionAllowlist.ts:18` so `no-silent-mutations` needs no new entry, `?includeMemberships=true` follows the existing `DevicesPage.tsx:343` precedent and the API's own "acceptable at current scale" note, the 8-locale i18n delta is consistent, and the 5 removed keys don't collide with the `longTail.snmp.SNMPTemplateEditor.*` namespace.

**Tests:** `DeviceGroupsPage.staticGroups` / `.dynamicGroups` / base + `localeParity` + `no-silent-mutations` — 180 passed, single run. `tsc --noEmit` on `apps/web` clean; no `.astro` touched. The new race regression test is **mutation-verified** — it fails when the generation guard is removed, so it isn't vacuous.

**Status:** review-clean, awaiting maintainer merge. Not merged and #3615 left open per the hand-off rule.

Issue #1318, opened 2026-06-13
[API/UI] Normalize all timezone fields to default to the partner timezone


ToddHebebrand:
## Background / goal

We have timezone (IANA tz string) fields scattered across many features — site settings, maintenance windows, automations, patch/backup/discovery/sensitive-data schedules, quiet hours, PAM time windows, ticketing, reporting. Each one defaults independently, almost always to a hardcoded `'UTC'`. Product ask (Todd): **normalize timezone defaulting so a tz field resolves to the partner's timezone when not explicitly set**, instead of UTC / org / browser / unset.

Goal: one canonical default-resolution behavior everywhere a tz exists, with the **partner timezone as the base default**, exposed via a single shared resolver rather than ~30 scattered `|| 'UTC'` fallbacks.

## Current state

**Where does the partner timezone live today?** It exists, but only as a free-form key inside the `partners.settings` JSONB blob — there is **no dedicated `timezone` column on `partners`**, and **no validation** on it. It is written/read by the UI via `PATCH /orgs/partners/me` (`partnerSettingsSchema`, `apps/web/src/components/settings/PartnerSettingsPage.tsx:117,157,216`). **Critically, nothing in the resolution chain ever consults the partner tz** — the only resolver in the codebase (`resolveDeviceTimezone`) goes site → org.settings → UTC and stops there.

| Entity | Has tz column? | Notes |
|---|---|---|
| `partners` | **NO** | tz only as a key in `partners.settings` JSONB (`apps/api/src/db/schema/orgs.ts:10,19`); no validation |
| `organizations` | **NO** | tz only as a key in `organizations.settings` JSONB (`orgs.ts:35,41`); read in `resolveDeviceTimezone` |
| `sites` | **YES** | `timezone varchar(50) NOT NULL DEFAULT 'UTC'` (`orgs.ts:58`); only entity with a real column + IANA validation on write (`routes/orgs.ts:127-158`) |
| `devices` | **NO** | reads `sites.timezone` for display (`routes/devices/core.ts:819` → `|| 'UTC'`) |
| `users` | **NO** | no per-user tz |

**DB columns that store a tz** (all default `'UTC'` except one nullable):
| table.column | file:line | default |
|---|---|---|
| `sites.timezone` | `db/schema/orgs.ts:58` | `'UTC'` |
| `maintenance_windows.timezone` | `db/schema/maintenance.ts:38` | `'UTC'` |
[...]

PR #3628, opened 2026-08-16
feat(web): add a Regional setup-wizard step for language, currency and time zone (#3204)

### LINKED ISSUE #3204: [UI] Setup wizard: add a regional step for language, currency and time zone
## Description

The first-run setup wizard collects none of language, currency, or time zone. The org and site created in step 2 silently default to `UTC`.

## Current state

Wizard lives at `/setup` (`apps/web/src/pages/setup.astro` -> `apps/web/src/components/setup/SetupWizard.tsx`), triggered by `users.setup_completed_at` being NULL via `requiresSetup` on `GET /users/me`.

Three steps today:

1. **Account** (`AccountSetupStep.tsx`) — email, current password, new password, confirm
2. **Organization** (`OrganizationSetupStep.tsx`) — organization name, site name
3. **Install Agent** (`EnrollDeviceStep.tsx`) — platform, device count, installer vs CLI, enrollment link

All three settings already exist elsewhere, so this is UI plumbing only — no API or schema work:

| Setting | Stored at | Existing UI |
|---|---|---|
| Time zone | `partners.timezone` (varchar 64, default 'UTC'), `sites.timezone` | Settings -> Partner -> Regional (`PartnerRegionalTab.tsx`) |
| Language | `partners.settings->>'language'` (default) and `users.preferences->>'locale'` (per-user) | Regional tab / `ThemingSettings.tsx` |
| Currency | `partners.currency_code` char(3) default 'USD' | Billing settings — a free-text 3-char input (`PartnerBillingSettings.tsx:160`) |

Note there is no per-user timezone; timezone is partner + site only.

## Proposed Fix

Add a **Regional** step between Organization and Install Agent:

- New `apps/web/src/components/setup/RegionalSetupStep.tsx`, modeled on `OrganizationSetupStep.tsx`. Reuse `apps/web/src/components/shared/TimezoneSelect.tsx` and the locale select markup from `PartnerRegionalTab.tsx`.
- `SetupWizard.tsx` — add the step label, bump `SETUP_STEP_COUNT` 3 -> 4, add the render branch.
- Persist via existing endpoints, no API changes: timezone -> `PATCH /orgs/partners/me` `{settings:{timezone}}` (column auto-syncs) and optionally `timezone` on the site created in step 2; language -> same PATCH `{settings:{language}}` and/or `PATCH /users/me` `{preferences:{locale}}`; currency -> `PATCH /partner/billing-settings` `{currencyCode}`.
- Nice touch: call `applyLocale()` / `writeLocalePreference()` (`apps/web/src/lib/appearance.ts`, `lib/i18n/index.ts`) on selection so the remaining wizard steps render in the chosen language.

While here: replace the free-text 3-char currency input on the billing settings page with a proper select.

## Gotcha

Locale keys must be added to **all seven** locale directories (`en`, `pt-BR`, `es-419`, `fr-FR`, `fr-CA`, `de-DE`, `it-IT`) under `setup.steps.*` / `setup.regional.*`. `apps/web/src/lib/i18n/localeParity.test.ts` enforces exact key parity and will fail otherwise.

## Cleanup

`apps/web/src/components/setup/ConfigReviewStep.tsx` and `SetupSummaryStep.tsx` exist but are referenced nowhere — dead code, safe to delete.

ToddHebebrand:
## What

The first-run setup wizard collected none of language, currency or time zone, so every fresh install landed on partner/site timezone `UTC`, currency `USD` and English — and the three settings that fix that live on three different pages a first-run user has no reason to visit (Settings → Partner → Regional, Theming, Billing settings).

Adds a **Regional** step between Organization and Install Agent. Pure UI plumbing over existing endpoints — no API or schema work.

## The step

`apps/web/src/components/setup/RegionalSetupStep.tsx`:

- Language `<select>` with self-named options reusing the existing `settings:language.*` labels, `TimezoneSelect` (the shared searchable IANA picker), and a currency `<select>`.
- Prefills from `GET /orgs/partners/me`, falling back to browser-detected zone and language. A stored `'UTC'` is treated as "never configured" (it is the column default) and loses to the detected zone; any other stored zone is a real choice and wins.
- Selecting a language **applies it immediately**, so the remaining wizard steps render in the chosen language.

### The four writes, and why each exists

| Endpoint | Body | Why |
|---|---|---|
| `PATCH /orgs/partners/me` | `{settings:{timezone,language}}` | Partner default. The route shallow-merges top-level settings keys and mirrors `settings.timezone` into the first-class `partners.timezone` column (#1318), so one call covers both. |
| `PATCH /orgs/sites/:id` | `{timezone}` | **The site created one step earlier defaults to `'UTC'` independently of the partner.** A partner-only write would leave the site — what device schedules actually resolve against — on UTC. |
| `PATCH /users/me` | `{preferences:{locale}}` | Per-user console language; server shallow-merges preferences. |
| `PATCH /partner/billing-settings` | `{currencyCode, invoiceNumberPrefix, invoiceTermsDays}` | Currency. `partnerBillingSettingsSchema` makes the other two **required** and `updatePartnerBillingSettings` writes both unconditionally, so they are **echoed back from the load** — a currency-only body would 400, and hardcoded `INV`/`30` would silently overwrite the partner's real values. |

Failure handling is deliberately non-advancing: a failed write names itself in an inline error and the user stays on the step, because advancing on a partial save would silently ship the exact defaults this step exists to replace and the wizard never returns here. **Skip** is the escape hatch for a role without `invoices:write` (the billing call is permission-gated).

## Also in this PR

- **Currency select on billing settings** (the issue's "while here"). The free-text `maxLength={3}` input happily accepted `XXX` or a typo'd `GPB` and only surfaced it once an invoice rendered. Replaced with a select over a curated ISO 4217 list (`apps/web/src/lib/currencies.ts`). Display names come from `Intl.DisplayNames` rather than the locale catalogs — 34 codes × 8 locales would otherwise be 272 hand-translated strings that are all parity/duplicate-baseline liabilities. An already-stored off-list code stays selectable, guarded by a regression test: a `<select>` whose value is absent from its options reads back `''` and would then be **saved** as `''`, wiping the partner's currency on any unrelated edit.
- **Dead code deleted**: `ConfigReviewStep.tsx` and `SetupSummaryStep.tsx` (referenced nowhere — confirmed by `git grep`), plus their now-orphaned `setup.config.*` / `setup.summary.*` / `setup.steps.configReview` keys in all eight bundles.

## i18n

New `setup.steps.regional` and `setup.regional.*` keys hand-translated in **all eight** locale directories including `tr-TR`. Deliberately avoided exact-English wordings (e.g. de-DE step label is `Region`, not `Regional`) so nothing lands in the duplicate baselines.

## Verification

- `src/lib/i18n/` contract suites — localeParity, translationCoverage, terminologyQuality, keyUsage, extractionQuality: **81 passed**.
- `src/components/setup/` — **12 passed** (7 new in `RegionalSetupStep.test.tsx`: partner prefill, partner+site timezone both written, invoice prefix/terms echo, immediate locale apply, no-advance-on-failure, site PATCH skipped when siteId is null, load-failure reported).
- `src/components/billing/PartnerBillingSettings.test.tsx` — **13 passed** (1 new).
- `src/lib/__tests__/no-silent-mutations.test.ts` — passed.
- `tsc --noEmit` clean; `eslint src` clean. No `.astro` files touched.

Closes #3204


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter`.

**Findings:** 3 raised, all real, all addressed. 0 outstanding.

1. **`void applyLocale(value)` discarded `usedFallback`** (silent-failure-hunter, HIGH). When the language chunk fails to load, `applyLocale` renders English instead — but the select still showed the chosen language and submit still persisted it, so what the user sees and what gets saved diverged with no signal. Now awaited with a toast, matching `ThemingSettings.tsx:161`. Uses a step-local string rather than `settings:themingSettings.languageLoadFailed`, which opens with "Preference saved" and would be untrue at select time. → `f85e624dd`
2. **The load effect listed `t` in its deps.** `useTranslation`'s `t` gets a new identity on every language change — which is exactly what this step's select does — so picking a language refetched the partner and reset the timezone and currency the user had just chosen. Effect now runs once. (Caught while reviewing, before the agents reported.) → `f85e624dd`
3. **A stale timer could yank the user forward after they left the step** (code-reviewer, confidence 88). The success banner delays `onNext` by 600ms, so the callback outlives an unmount, and Back had no `disabled={saving}` guard unlike Skip and Save. Click Save → Back went to Organization, then 600ms later the stale `onNext` fired `setCurrentStep(3)` and dropped the user on Install Agent. Back is now locked for the in-flight window and the advance is guarded by a mounted ref with the timer cleared on unmount. → `16b131395`

Non-blocking and deliberately not changed: the four writes use inline error/success state rather than `runAction` — consistent with the sibling wizard steps (`AccountSetupStep`, `OrganizationSetupStep`), none of which are in `TARGET_GLOBS`, and outcomes are surfaced either way. Flagging for the maintainer rather than deciding it here.

**Anti-vacuity:** the finding-3 regression test holds the first PATCH open, because with instant mocks `saving` flips back before the disabled state can be asserted — the obvious version of that test passes with both guards deleted. Verified load-bearing by reverting the guards: it fails.

**Tests:** `src/components/setup/`, `src/components/billing/PartnerBillingSettings.test.tsx`, `src/lib/i18n/`, `src/lib/__tests__/no-silent-mutations.test.ts` — **238 passed, 12 files** (11 new). `tsc --noEmit` clean, `eslint src` clean. No `.astro` files touched, so no `astro check` needed.

**Note for the reviewer:** commits `2c423e153`, `fd8aecf7e` and `edc1b1de1` are housekeeping, not product changes — a review pass left scratch files (`__t_identity.test.tsx`, `RegionalSetupStep.tsx.bak2`) and a NUL byte in `currencies.ts` (which made git treat the file as binary) in the worktree. All removed; the branch diff is 16 files.

**Status:** review-clean, awaiting maintainer merge. Issue stays open and assigned until you've verified the wizard end-to-end.

Issue #3614, opened 2026-08-16
[API] Manual remediation approval tokens: no sweeper, schema drift on the partial index, and three smaller follow-ups (#3585)


ToddHebebrand:
Follow-ups from the review of #3585 (merged), which backed manual software-remediation with a durable single-use approval token. The security core is sound — the consume is one atomic `UPDATE ... WHERE consumed_at IS NULL AND expires_at > now() ... RETURNING`, so there's no TOCTOU, and all four cascade lists plus RLS are registered. These are the loose ends around it.

**1. Unconsumed authorization rows accumulate with no sweeper.** The migration creates the table but nothing prunes it. A single remediate call mints up to 500 rows (one per target device), and any target that never reports back leaves its row behind until the device, policy, or org is deleted. On an active fleet that is unbounded growth on a table that only ever needs a short window of history. Add a prune to the existing cleanup job — anything past `expires_at` by a comfortable margin is dead weight.

**2. The partial index is missing from the Drizzle schema.** The migration creates `software_remediation_requests_expiry_idx` as a partial index, but `apps/api/src/db/schema/softwarePolicies.ts:344-349` doesn't declare it. Partial indexes are expressible in Drizzle here — `abuseSignals.ts:28` does it — so this will surface as an extra index in `pnpm db:check-drift`. Cheap to close and it keeps the drift check honest.

**3. A second pool connection is opened while the request transaction still holds one.** `createManualRemediationAuthorizations` in `softwareRemediationWorker.ts` calls `runOutsideDbContext`, which clears the ALS context but does not release the connection the request is already holding. That's two connections for the duration. Worth watching against the US region's 25-connection ceiling; not a bug today, but it's the shape that produced earlier pool-starvation incidents.

**4. Double-redemption is proven sequentially, not concurrently.** The tests confirm a second consume fails after the first completes. The atomic UPDATE means the concurrent case is almost certainly fine too, but a two-promise race test would lock the property in against a future refactor that reintroduces a read-then-write.

**5. Confirm the cooldown bypass is intended.** A manual remediation now skips the cooldown entirely. That reads like deliberate product behavior — a human explicitly asked for this — rather than a side effect of preserving the token, but it isn't stated anywhere and is worth writing down before someone "fixes" it.

PR #3630, opened 2026-08-16
fix(api): prune remediation authorization rows and close the #3585 follow-ups

bdunncompany:
Closes four of the five follow-ups from your #3585 review. The fifth I'm answering rather than changing, and one of the four comes with a correction to its premise.

### 1. Sweeper

Nothing pruned `software_remediation_requests`. One Remediate click mints up to 500 rows and every target that never reports back leaves its row until the device, policy, or org is deleted and the FK cascade takes it.

New hourly worker deletes rows past `expires_at` by a configurable grace (`SOFTWARE_REMEDIATION_REQUEST_RETENTION_HOURS`, default 72), batched, re-applying the same cutoff in the `DELETE` it selected on. Consumed rows go on the same schedule — the audit entry written at consume time is the durable record, not the row.

The sweep is **bounded at 50 batches**. Without a cap, a table receiving eligible rows as fast as they're deleted keeps every batch full and the loop never exits. The hourly schedule picks up the remainder, and the capped case is logged rather than silent.

It's its own worker rather than a graft onto an existing sweep because every retention concern in `jobs/` owns one small worker, and `softwareRemediationWorker` itself is job-driven with no repeatable component to hang this on. Happy to fold it elsewhere if you'd rather.

### 2. Partial index — declared, but a correction to the premise

Declared in Drizzle mirroring the migration, using the same `.where(sql\`...\`)` shape `abuseSignals.ts:28` uses.

**It does not change `pnpm db:check-drift`, though.** That script compares migration filenames against the `breeze_migrations` ledger, and `scripts/check-drift.ts:17-26` states that schema-vs-live-DB drift is *"intentionally NOT checked"* because the drizzle-kit round-trip produces thousands of false positives here. So no gate was failing and none starts passing.

Declaring it is still right — the schema file should describe what the database actually has, and it's what a reader reasons from — but I didn't want to ship a justification I'd checked and found wrong.

Related: the index serves the **consume** path (which filters `consumed_at IS NULL`). The sweep scans `expires_at` across consumed and unconsumed rows alike, so the partial index can't cover it. The comment says that rather than implying the sweep benefits.

### 4. Concurrent double-redemption

The existing test proved single-use only *after* the first consume committed. Added one that starts both together and asserts exactly one wins — never both (double uninstall dispatch), never neither (a burned token that queued nothing).

The unit tests now also **capture and assert the actual `WHERE` predicates** instead of counting calls. The first version I wrote would have passed against an implementation that selected the wrong column or dropped the cutoff recheck from the `DELETE`. Control-run with that recheck removed, it now fails on `expected [ 'id' ] to include 'expires_at'`.

### 5. Cooldown bypass — intended

Already documented at `softwareRemediationWorker.ts:400`; I've expanded it with the reasoning rather than the bare assertion. The cooldown exists to stop the *unattended* worker retrying a failing policy every tick, and a human who has passed MFA and burned a single-use row has cleared a stricter gate. The comment now also warns why turning it into a throttle is unsafe: consume-then-cooldown-skip burns the row for zero queued commands with no way to retry — that was the blocker while building #3585.

### 3. Second pool connection — deliberately not here

You called it "not a bug today" and I agree. Fixing it properly means touching `runOutsideDbContext` itself, which affects every caller. Wrong thing to slip into a follow-up PR.

### Also

A failed initialization now closes the worker and queue it already created. `initializeWorkers()` logs and keeps the process running, so a Redis error while scheduling would otherwise have left both live for the rest of the process.

### Verification

node 22.23.2 (CI `.nvmrc`), pnpm 10.34.5:

- `api` `src/jobs/` — 110 files, **1038/1038 pass**
- `api` `softwareRemediationRequestCleanup.test.ts` — **6/6**
- `api` eslint on all four changed files — clean

The concurrent-consume test is an integration test needing a real Postgres, so it runs in CI rather than locally.

Refs #3614, #3585

### DIFF
--- a/apps/api/src/__tests__/integration/softwareRemediationRequestsRls.integration.test.ts
+++ b/apps/api/src/__tests__/integration/softwareRemediationRequestsRls.integration.test.ts
@@ -16,6 +16,7 @@ import { and, eq } from 'drizzle-orm';
 import { db, withDbAccessContext, type DbAccessContext } from '../../db';
 import { devices, softwarePolicies, softwareRemediationRequests } from '../../db/schema';
 import { consumeManualRemediationAuthorization } from '../../jobs/softwareRemediationWorker';
+import { pruneExpiredRemediationRequests } from '../../jobs/softwareRemediationRequestCleanup';
 import { createOrganization, createPartner, createSite } from './db-utils';
 
 const createdRequests: string[] = [];
@@ -62,12 +63,12 @@ async function seedOrgPolicy(orgId: string): Promise<string> {
   return row!.id;
 }
 
-async function seedRequest(fields: { orgId: string | null; partnerId: string | null; policyId: string; deviceId: string; consumed?: boolean }): Promise<string> {
+async function seedRequest(fields: { orgId: string | null; partnerId: string | null; policyId: string; deviceId: string; consumed?: boolean; expiresAt?: Date }): Promise<string> {
   const [row] = await withDbAccessContext(SYSTEM_CTX, () =>
     db.insert(softwareRemediationRequests).values({
       orgId: fields.orgId, partnerId: fields.partnerId, policyId: fields.policyId, deviceId: fields.deviceId,
       requestedByUserId: null,
-      expiresAt: new Date(Date.now() + 60 * 60 * 1000),
+      expiresAt: fields.expiresAt ?? new Date(Date.now() + 60 * 60 * 1000),
       consumedAt: fields.consumed ? new Date() : null,
     }).returning({ id: softwareRemediationRequests.id }),
   );
@@ -146,6 +147,37 @@ describe('software_remediation_requests — RLS + device-move + consume (#3553)'
     expect(second.authorized).toBe(false); // already consumed
   });
 
+  // #3614 follow-up 4: the sequential case above proves single-use only AFTER
+  // the first consume has committed. The property that actually matters is that
+  // two consumes racing on the SAME row cannot both win. That holds because the
+  // consume is one atomic `UPDATE ... WHERE consumed_at IS NULL ... RETURNING`,
+  // so the second UPDATE blocks on the first's row lock and then matches zero
+  // rows. Pinning it here means a future refactor to read-then-write — which
+  // would still pass the sequential test — fails instead.
+  it('concurrent consumes of the same request: exactly one wins', async () => {
+    const partnerA = await createPartner();
+    const orgA = await createOrganization({ partnerId: partnerA.id });
+    const siteA = await createSite({ orgId: orgA.id });
+    const deviceD = await seedDevice(orgA.id, siteA.id);
+    const policyP = await seedOrgPolicy(orgA.id);
+    const requestId = await seedRequest({ orgId: orgA.id, partnerId: null, policyId: policyP, deviceId: deviceD });
+
+    // Started together, not awaited in sequence: both are in flight before
+    // either resolves.
+    const [a, b] = await Promise.all([
+      withDbAccessContext(SYSTEM_CTX, () =>
+        consumeManualRemediationAuthorization({ manualRequestId: requestId, policyId: policyP, deviceId: deviceD }),
+      ),
+      withDbAccessContext(SYSTEM_CTX, () =>
+        consumeManualRemediationAuthorization({ manualRequestId: requestId, policyId: policyP, deviceId: deviceD }),
+      ),
+    ]);
+
+    // Exactly one authorized — never both (double uninstall dispatch), never
+    // neither (a burned token that queued nothing).
+    expect([a.authorized, b.authorized].filter(Boolean)).toHaveLength(1);
+  });
+
   // #3553 finding 4: prove the PARTNER axis of the dual RLS + consume, not just
   // the org axis. A partner-wide policy's request is visible to its partner and
   // NOT to another partner, and consumes while the device stays under the partner.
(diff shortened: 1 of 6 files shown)


ToddHebebrand (changes requested):
**Requesting changes — one must-fix, and I confirmed it by running the mutation rather than reasoning about it.**

The prune logic itself is right, and I want to be specific about that because the fix here is narrow. The DELETE re-applies the cutoff instead of trusting the id list, so a row that stops being eligible between the SELECT and the DELETE survives — that race guard is the correct design. The work is batched and capped. The predicate is org-agnostic, so partner-wide (`org_id NULL`) rows are swept rather than silently skipped, which is the trap that usually catches these. The queue/init/shutdown wiring mirrors `softwareUploadSessionCleanup`, and cleaning up on init failure is a genuine improvement over the file you copied. The index declaration matches the shipped migration verbatim (`apps/api/migrations/2026-08-23-software-remediation-requests.sql:68`), and there's correctly no new migration, no new table and no new column — so `tenantCascade.ts` and `tenantExportPolicyRegistry.ts` (where `software_remediation_requests` is already registered at :329 and :300) need nothing.

**Must-fix — the test suite cannot detect an inverted comparison, on a surface where that mutation destroys live authorizations.**

`columnsIn` (`apps/api/src/jobs/softwareRemediationRequestCleanup.test.ts:51`) walks `queryChunks` and returns column *names* only — at :56 it returns `[node.name]` and stops. It never inspects the operator or the bound cutoff value. So `expect(columnsIn(capturedSelectWhere[0])).toEqual(['expires_at'])` at :86 asserts that the predicate *mentions* `expires_at`, not that it filters the correct direction.

I checked out the branch and inverted both comparisons at `softwareRemediationRequestCleanup.ts:104` and `:113` (`lt` → `gt`), which turns the job into one that deletes every **unexpired** authorization and leaves the expired ones. Result:

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```

Six for six, unchanged. The comment at test:23 states the suite would catch it if someone "dropped the cutoff recheck from the DELETE, or used `gt` instead of `lt`." Half of that is true — dropping the recheck changes the column set — but the `gt`/`lt` half is not, and a comment asserting coverage that doesn't exist is worse than no comment, because the next person reads it and doesn't look.

Two ways to close it, and I'd take the second:
- Assert the operator and the bound parameter, not just the column — walk to the `sql` operator chunk and the bound `cutoff` value.
- Better, since this is an approval surface: add a real-Postgres case to the existing `softwareRemediationRequestsRls.integration.test.ts`. Seed a live row, one expired an hour ago, and one expired 100 hours ago, across two orgs, then assert only the third is gone. Nothing in this PR currently exercises the prune against a real database, and the failure mode you're guarding is a row count, not a query shape.

**Nice-to-haves, fold in while you're here:**
1. `softwareRemediationRequestCleanup.ts:145` reaches through `__testOnly.MAX_BATCHES` at runtime — use the plain const; the `__testOnly` export is for the test, not the log line.
2. `pruneExpiredRemediationRequests` is exported but calls bare `db`; only the worker wraps it in `withSystemDbAccessContext`. A future caller inside a request path would silently inherit org scope. Either wrap internally or drop the export.
3. `hasMore` reports a false positive when a capped batch happened to take the last eligible rows — log-only, but it'll send someone chasing a backlog that isn't there.

Everything else here is solid; it's the one assertion.

bdunncompany:
Fixed — and the fix itself was wrong twice before it was right, so I want to show that rather than just claim it's done.

**The must-fix.** You were right, and I reproduced your mutation exactly: `lt` → `gt` on the cutoff, unit suite still 6/6. `columnsIn` returns column *names* only — it cannot see the operator or the bound value — and my comment claiming it caught `gt`/`lt` was false. As you said, a comment asserting coverage that doesn't exist is worse than none, because the next person stops looking. That comment now states its real scope.

Took your second option. `softwareRemediationRequestsRls.integration.test.ts` now seeds three rows across **two orgs** — live (+1h), expired 1h ago (inside the 72h grace), expired 100h ago in the other org — and asserts only the third is gone. **Control-run with your exact mutation it fails**: `expected false to be true`, the live authorization deleted. It's a row count, exactly as you said, so it can't be expressed as a query shape.

**Then an adversarial pass over my own fix caught two things I'd asserted without checking.**

First, I moved the prune behind `__testOnly` and wrote that it was now "unreachable" from request code. That's false — `__testOnly` is an ordinary runtime export; any module can import it and call straight through. I'd removed a named export, not reachability.

Second, I claimed self-protection was impossible, on the grounds that a nested `withSystemDbAccessContext` can't escalate. The nesting reading was right — `withDbAccessContext` short-circuits to `fn()` when a context exists — but the conclusion wasn't: **`runOutsideDbContext` exits both stores**, which is precisely how `createManualRemediationAuthorizations` already escalates in the sibling file. I'd argued from half the mechanism.

So the prune now establishes its own context **per batch** via `runOutsideDbContext(() => withSystemDbAccessContext(...))`. The integration test calls it with no ambient context at all, which is what proves it.

**That fixed a third thing the same pass found, and it undercut my own rationale.** The worker's single outer `withSystemDbAccessContext` opened **one transaction for the entire sweep** — so up to 50 batches' worth of row locks and dead tuples were held until the whole thing finished. My comment said batching stopped one sweep holding locks over a huge delete; the statements were batched but the transaction wasn't. Per-batch contexts commit independently. The sweep is no longer atomic, which is right for an idempotent retention job — a partial run just prunes fewer rows and the next hour finishes.

**Nice-to-haves, all three:** plain `MAX_BATCHES` in the log line; `hasMore` probes once before claiming a backlog — documented as a point-in-time observation rather than a guarantee, and noting that `LIMIT 1` bounds rows returned, not work done, since the partial `expires_at` index can't serve a predicate that spans consumed rows.

**Gate** — node 22.23.2, pnpm 10.34.5:

- `tsc --noEmit` — 0 errors
- `src/jobs/` — 110 files, **1038/1038**
- `softwareRemediationRequestsRls.integration` — **6/6 against real Postgres**
- eslint on all three changed files — clean
- control: the `lt` → `gt` mutation **fails** the integration test

Also confirmed main is green across all seven workflows at `ab7af17c8` after tonight's merges.

ToddHebebrand (approved):
**Approved — verified the fix against the diff, not the description.**

The integration case is exactly the shape I asked for: three rows across two orgs, survivors asserted by id against real Postgres, and the prune called with no ambient context so the per-batch `runOutsideDbContext(() => withSystemDbAccessContext(...))` is what's actually proven. The unit-test comment now states its real scope instead of overclaiming, `MAX_BATCHES` is read plainly in the log line, and the `hasMore` probe closes the false-positive.

The two things you found on your own pass were both worth more than my original nice-to-haves: the `__testOnly`-isn't-a-guard correction, and the single-transaction-for-50-batches lock hold. The per-batch commit note at `softwareRemediationRequestCleanup.ts` is the right call for an idempotent retention job, and the concurrent-consume race test is a genuine addition beyond the ask.

Merging.

PR #3631, opened 2026-08-16
feat(web): edit-in-place for persisted quote table/callout blocks (#3547)

### LINKED ISSUE #3547: Quote editor: edit-in-place for persisted table/callout blocks
The table/callout authoring shipped in #3539 supports create + read-only display in the editor canvas. Editing a persisted block requires delete-and-recreate — a papercut for anyone fixing a typo in a large table.

Add an edit affordance on `QuoteBlockCard` for `table`/`callout` reusing the existing create-form UI prefilled from the block content, saving via `updateBlock` (the contract-block edit pattern is the model).

ToddHebebrand:
## What

Persisted `table` and `callout` blocks in the quote editor were create-and-display only — fixing a typo in a large table meant deleting the block and rebuilding it. Both now have an inline **Edit** affordance on the canvas that reuses the existing create-form fields, prefilled from the block's current content, saving through the existing `updateBlock` path. The contract block's editor is the precedent followed.

## How

- **`QuoteBlockContentForms.tsx` (new)** — the table grid and callout authoring fields, lifted out of `QuoteEditor.tsx` and parameterised by an id/test-id prefix so the add form and a per-block edit form can be mounted at the same time. Sharing one implementation is the point: the "every row has exactly `columns.length` cells" invariant that `quoteTableContentSchema`'s exact-shape refinement enforces server-side now cannot drift between create and edit. Also exports `*FromContent` / `*ToContent` so both surfaces build identical request bodies.
- **`QuoteStructuredBlockEditor.tsx` (new)** — `TableBlockEditor` / `CalloutBlockEditor`. Read-only render at rest with an Edit button for writers; the form replaces it while editing, with Save / Cancel and the shared `SrSaved` flash. Deliberately different from `ContractBlockEditor`, which renders its form unconditionally: on the canvas a table should look like a table, so the grid is behind a toggle.
- **Prefill is defensive** — a legacy/hand-written row that disagrees with `columns.length` is padded or trimmed once, visibly, on open, rather than being PATCHed into a schema rejection the user can't act on.
- **Failure handling** — save goes through the parent's `editBlock` → `runAction`, so a rejected PATCH toasts and leaves the form open with the draft intact (no silent no-op).
- Read-only viewers keep the rendered block and get no edit affordance.

No API change: `PATCH /quotes/:id/blocks/:blockId` already validates `table` / `callout` content through `quoteBlockInputSchema`.

## Testing

- `QuoteEditor.tableblock.test.tsx` +6: read-only gating, prefill from persisted content, save PATCH body, add-column padding the persisted row (shape invariant), cancel discarding the draft, failed save keeping the draft.
- `QuoteEditor.calloutblock.test.tsx` +7: same plus emptied-body disabling save and a cleared title being omitted rather than sent as `''`.
- All 50 quote component suites green (420 tests), i18n locale parity green (8 locales), `tsc --noEmit` clean.

Closes #3547

### DIFF
--- a/apps/web/src/components/billing/quotes/QuoteBlockCard.tsx
+++ b/apps/web/src/components/billing/quotes/QuoteBlockCard.tsx
@@ -22,13 +22,12 @@ import {
   type QuoteBlock,
   type QuoteLine,
   type QuoteLineRecurrence,
-  type QuoteTableContent,
-  type QuoteCalloutContent,
   formatMoney,
 } from './quoteTypes';
 import { type LineUpdate, SrSaved, fieldRing, pendingKey, seamless } from './quoteEditorShared';
 import { GhostRow, EditableLineRow, ReadonlyLineRow, type LineRevealRequest } from './QuoteLineRows';
 import { ContractBlockEditor } from './QuoteContractBlockEditor';
+import { TableBlockEditor, CalloutBlockEditor } from './QuoteStructuredBlockEditor';
 
 // ── A single block, with an inline line builder when it is a pricing table ──
 export function BlockCard({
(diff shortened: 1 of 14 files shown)


ToddHebebrand:

**Findings:** 2 raised, both addressed in `686b1cc5e`; 0 outstanding.
- *silent-failure-hunter (high):* `tableFormFromContent` reshaped out-of-cap / ragged stored content and a subsequent Save overwrote the block with the reshaped version — silently, despite the doc comment claiming the reshape was "visible". Now returns `{ form, adjusted }`, and `TableBlockEditor` shows a warning above the grid stating that saving replaces the stored table. Covered by three new tests (ragged-row padding, >8-column truncation, and a negative case that does not warn).
- *pr-test-analyzer:* no vacuous assertions found in the new suites, but flagged the reshape path, the caps, and the saved cue as uncovered. All three now have tests. It also flagged `canSave` being a hardcoded `true` on the table toolbar — `canSave` now defaults to `true` and the table omits it with a comment on why a table has nothing to gate on (empty labels/cells are legal and the add form submits a blank grid too). Reusing the shared `useSavedFlash` also dropped a duplicated local copy.
- *code-reviewer:* 0 findings — confirmed the create-path POST body semantics are unchanged by the extraction, the edit builders can't violate `quoteTableContentSchema`, and all new i18n keys resolve in `en` + all 7 other locales.

The saved-cue assertion is deliberately on the live region's **text**, not its presence: `SrSaved` always mounts the span and only swaps its content, so a presence check would have passed vacuously.

**Tests:** `apps/web` quotes + i18n suites green single-fork — 57 files / 529 tests (`vitest run src/components/billing/quotes src/lib/i18n`), including 20 in `QuoteEditor.tableblock.test.tsx` and 14 in `QuoteEditor.calloutblock.test.tsx`. `tsc --noEmit` clean; `eslint` clean on the changed files.

**CI:** Type Check, Lint and Check Migrations pass; Test Web / Test API still running at time of writing with **no failing checks**.

**Status:** review-clean, awaiting maintainer merge. No API or schema change in this PR.

Issue #3432, opened 2026-08-11
[Web] QA polish batch: false unsaved-changes prompt after save, wrong Quick Support code copy, untranslated audit actions in pt-BR


ToddHebebrand:
## Description

Three small, unrelated papercuts from the v0.105.0 QA sweep. Batched because each is a few lines and none justifies its own issue — split out if any turns out to be deeper than it looks.

### 1. Spurious unsaved-changes prompt after a *successful* save

Settings → Organization → Remote Access: save successfully, then navigate away, and the browser still shows the `beforeunload` "leave site?" prompt. The dirty flag isn't cleared on a successful save, so the tab claims unsaved work that was in fact saved.

Worse than cosmetic: it trains people to dismiss that prompt, which is the one prompt you want them to read.

### 2. Quick Support code copy claims a code "can never be shown again", but it can

The technician-facing copy states the support code can never be shown again. The same session can be **reopened** — that was added deliberately in `9344a2b4e` — and the code is shown again when it is. The copy predates the reopen feature and now contradicts it.

Either soften the copy to match reopen semantics, or state the actual constraint (whatever the code's real lifetime is).

### 3. Audit-log action names untranslated in pt-BR

Switch locale to pt-BR and open the Audit Trail: the action names render in English while the surrounding chrome translates correctly. Other pages spot-checked in pt-BR (Dashboard, Devices, Discovery, Integrations) were clean, so this is specific to the audit action vocabulary rather than a locale-loading failure.

Likely these are raw enum values (`device_group.create` and friends) being displayed directly rather than mapped through a label catalog. If so, the fix is a label map, and the same map would improve the English rendering too — `device_group.create` is not a sentence.

## Affected files

- Org Remote Access settings tab (dirty-flag reset on successful save)
- Quick Support technician page (code copy)
- Audit Trail viewer + locale files (action-name labels)

Related: #2649 (i18n extraction damage class), #3424 (another codemod-damage rendering fix from the same sweep).

Reported By: internal Playwright UI QA sweep 2026-08-11, evidence in `docs/testing/FEATURE_TEST_LOG.md` (lines 3923-4033).

PR #3633, opened 2026-08-16
fix(web): UI polish batch from the 2026-08-11 pre-release sweep (#3452)

### LINKED ISSUE #3452: [UI] Small polish batch from 2026-08-11 pre-release sweep: icon-only pagination buttons, "1 device(s)", run-counter lag, bare ACTIVITY header
Four small papercuts from the same sweep, none release-blocking, batched because each is a one-liner:

1. **Scripts list pagination buttons are icon-only with no accessible name** — prev/next have no aria-label or text (`apps/web` scripts list pagination footer). Screen readers announce nothing.
2. **Migration banner pluralization** — "1 device(s) are running the hosted agent edition…" should use i18n plurals (`MigrationRequiredBanner`).
3. **Fleet remediation run counters lag the per-target rows** — until the first 30s poll tick the panel reads "0 succeeded, 0 failed, 0 skipped of 4 targeted" while all 4 rows already show Skipped (`RunProgressPanel`). Cosmetic but reads as contradictory.
4. **Device Overview ACTIVITY section renders a bare header** — no empty-state copy when there are no activity entries (`DeviceDetails`).

Also noting (no action proposed): GET /reliability/:deviceId 404s for devices with no snapshot — UI handles it, but it's console noise on every fresh device.

Reported By: internal Playwright UI QA sweep 2026-08-11 — evidence in docs/testing/FEATURE_TEST_LOG.md.

ToddHebebrand:
Closes #3452

Four papercuts from the same Playwright QA sweep. Two of the four turned out to have a different root cause than the issue assumed — noted below.

### 1. Scripts list pagination buttons had no accessible name
`ScriptList.tsx` — the prev/next arrows were icon-only with no `aria-label` or text. Labels reuse the shared `common:actions.previousPage` / `nextPage` pair already used by `ConfigPolicyList` rather than minting a scripts-namespace duplicate, so **no new locale keys**. Icons are now `aria-hidden`.

### 2. "1 device(s) are running…"
`common:migrationBanner.message` is now a real i18next plural (`message_one` / `message_other`) in **all 8 locale bundles**. The call site already passed `count`, so no component change was needed. tr-TR carries both forms with identical text — Turkish does not inflect the noun after a numeral.

### 3. Fleet run counters lag the per-target rows — real bug, server-side cause
Both the summary and the rows come from the *same* `getRun` response, so the lag isn't a fetch-ordering issue. The run row's `succeededCount`/`failedCount`/`skippedCount` are a **denormalised roll-up that only `pollRunProgress` recomputes**, while the dispatcher's own `markTargetSkipped`/`markTargetFailed` (`services/fleetFindings/dispatch.ts`) write the **target row alone**. Between dispatch and the first poll tick the panel therefore read "0 succeeded, 0 failed, 0 skipped of 4 targeted" directly above four rows already showing Skipped.

`RunProgressPanel` now counts the target rows in the same snapshot. It falls back to the roll-up when `targets.length !== targetCount`: a site-restricted caller gets a `targets` array filtered to their allowed sites (see `getRemediationRun`) while the roll-up still covers the whole run, and counting a subset there would under-report rather than fix anything.

Fixed client-side rather than by recomputing in `dispatchRunChunk`, which would put a run-wide aggregate write in the path of concurrently-racing chunks. The roll-up staying eventually-consistent is fine; the panel reading it as authoritative was not.

### 4. Bare ACTIVITY header — premise corrected
`DeviceActivityFeed` **does** have empty-state copy ("No recent actions on this device."), and has since it was created. It is simply unreachable on desktop: a device with nothing to show auto-collapses to the thin vertical rail (`onHasContentChange` → `setActivityCollapsed(!hasContent)`), and the expanded card is `lg:hidden` while collapsed. The rail rendered its badge only when `itemCount > 0`, so an empty device got a bare "ACTIVITY" and nothing else — indistinguishable from still-loading, which is exactly what QA saw.

Kept the deliberate space-saving auto-collapse and gave the rail the missing signal instead: a muted `0` badge, and the existing empty-state string appended to its `aria-label`. Reuses `deviceActivityFeed.noRecentActionsOnThisDevice`, so again **no new locale keys**.

### Not addressed
The `GET /reliability/:deviceId` 404 console noise was explicitly filed as "no action proposed" — left alone.

## Overlap with #3432
Checked before starting: there is currently **no open PR referencing #3432**, and no open PR touches any of the four files here. The nearest neighbours (#3626, #3563) are both in `DeviceGroupsPage.tsx`. The only shared surface with any in-flight work is `apps/web/src/locales/*/common.json`, where this PR adds one key pair inside the existing `migrationBanner` object — a textual conflict there would be trivial to resolve.

## Verification
- `apps/web/src/components/devices` — 56 files, 654 tests, green
- `apps/web/src/components/scripts` + `.../fleet` — 14 files, 159 tests, green
- i18n contract suites (`localeParity`, `translationCoverage`, `keyUsage`, `extractionQuality`) — 79 tests, green; locale parity holds across all 8 bundles
- `tsc --noEmit` on `apps/web` — clean

New regression coverage: `ScriptList.test.tsx` (new file, accessible-name assertions), banner singular/plural rendering against the real i18next resources, the derived-counts bug and the site-filtered fallback, and the collapsed empty rail. One pre-existing `RunProgressPanel` fixture was corrected — it set `succeededCount: 1, failedCount: 1` while leaving both target rows `queued`, a state the API cannot produce; the target statuses were made to agree and the assertion left as-is.

### DIFF
--- a/apps/web/src/components/devices/DeviceActivityFeed.test.tsx
+++ b/apps/web/src/components/devices/DeviceActivityFeed.test.tsx
@@ -149,6 +149,27 @@ describe('DeviceActivityFeed', () => {
     expect(bar).toHaveAttribute('aria-label', 'Activity, No recent actions on this device.');
   });
 
+  it('collapsed: announces the item count when there are events but no active alerts (#3452)', async () => {
+    // The badge is aria-hidden and an aria-label suppresses descendant text,
+    // so without the count in the label this everyday state announced a bare
+    // "Activity" and dropped the number entirely.
+    mockFeed([evt('e1'), evt('e2')], []);
+    render(<DeviceActivityFeed deviceId="dev-1" collapsed onToggleCollapse={() => {}} />);
+    const bar = await screen.findByTestId('activity-rail-collapsed');
+
+    await waitFor(() =>
+      expect(bar).toHaveAttribute('aria-label', 'Activity, 2 recent actions'),
+    );
+  });
+
+  it('collapsed: uses the singular form for a single item (#3452)', async () => {
+    mockFeed([evt('e1')], []);
+    render(<DeviceActivityFeed deviceId="dev-1" collapsed onToggleCollapse={() => {}} />);
+    const bar = await screen.findByTestId('activity-rail-collapsed');
+
+    await waitFor(() => expect(bar).toHaveAttribute('aria-label', 'Activity, 1 recent action'));
+  });
+
   it('expanded: the header chevron fires onToggleCollapse to collapse', async () => {
     mockFeed([evt('e1')], []);
     const onToggleCollapse = vi.fn();
--- a/apps/web/src/components/devices/DeviceActivityFeed.test.tsx
+++ b/apps/web/src/components/devices/DeviceActivityFeed.test.tsx
@@ -137,6 +137,18 @@ describe('DeviceActivityFeed', () => {
     expect(onToggleCollapse).toHaveBeenCalledTimes(1);
   });
 
+  it('collapsed: shows a "0" count badge and appends the empty-state copy to aria-label when there is nothing to show (#3452)', async () => {
+    // Without this, an empty device rendered a bare "Activity" rail —
+    // indistinguishable from still-loading.
+    mockFeed([], []);
+    render(<DeviceActivityFeed deviceId="dev-1" collapsed onToggleCollapse={() => {}} />);
+    const bar = await screen.findByTestId('activity-rail-collapsed');
+
+    await waitFor(() => expect(within(bar).getByTestId('activity-rail-count')).toBeInTheDocument());
+    expect(within(bar).getByTestId('activity-rail-count')).toHaveTextContent('0');
+    expect(bar).toHaveAttribute('aria-label', 'Activity, No recent actions on this device.');
+  });
+
   it('expanded: the header chevron fires onToggleCollapse to collapse', async () => {
     mockFeed([evt('e1')], []);
     const onToggleCollapse = vi.fn();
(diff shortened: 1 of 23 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` + `pr-test-analyzer`.

**Findings:** 2 raised, both addressed in `051f0cc94`; 0 outstanding.

- **Important — a11y gap in item 4.** The collapsed ACTIVITY rail's `aria-label` only carried extra detail for the active-alerts branch and the new empty branch, so the everyday "events but no active alerts" state announced a bare "Activity". The count badge is `aria-hidden` — and an `aria-label` suppresses descendant text regardless — so the number reached a screen reader nowhere. Now announced via a new `deviceActivityFeed.recentActions` plural added to all 8 bundles, with singular/plural tests.
- **Consistency — item 1.** `ScriptList` had introduced a third way to reach a second i18n namespace. Switched to the repo's combined-namespace `useTranslation(['scripts', 'common'])`; `scripts` stays the default so existing call sites are unaffected.
- **Minor, from the test analyzer.** Documented that the unmocking `describe` in `MigrationRequiredBanner.test.tsx` must stay last in the file, since the unmock is never restored.

Both reviewers independently cleared the item-3 fallback guard (`targets.length === targetCount`): a site-restricted caller's lengths can only agree when their allowed sites cover every target, in which case counting the visible set is correct rather than a coincidental under-count. The test analyzer also confirmed the one pre-existing fixture change was a genuine internal inconsistency the fix surfaced (`succeededCount: 1, failedCount: 1` against two `queued` target rows — a state the API cannot produce), not an assertion weakened to accommodate the new code, and that all four new tests fail against pre-fix source.

**Tests:** `apps/web` — devices 56 files/654 tests, scripts+fleet 14/159, i18n contract suites (`localeParity`, `translationCoverage`, `keyUsage`, `extractionQuality`) 79 tests, and the post-fix re-run 19 files/223 tests — all green. `tsc --noEmit` on `apps/web` clean. Locale parity holds across all 8 bundles with both new key pairs.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3452 — the reporter's verification and merge timing are yours.

PR #3635, opened 2026-08-16
fix(web): QA polish batch — false unsaved prompt, Quick Support code copy, pt-BR audit actions (#3432)

### LINKED ISSUE #3432: [Web] QA polish batch: false unsaved-changes prompt after save, wrong Quick Support code copy, untranslated audit actions in pt-BR

ToddHebebrand:
Three independent papercuts from the v0.105.0 QA sweep. Each is small; they share no files.

Closes #3432

## 1. False unsaved-changes prompt after a successful save

**Root cause was inverted from the report.** The dirty flag was not "failing to reset" — it was set on the *wrong side of the save*. In `OrgRemoteAccessSettings.tsx`, `addRule` / `updateRule` / `deleteRule` each persist through their own `POST`/`PUT`/`DELETE /tunnels/allowlist` and then called `onDirty()`, marking the page unsaved **after** the write succeeded. Nothing could ever clear it: the tab has no save button, and the parent only clears on `handleSaveSettings`, which this tab never calls. So one allowlist edit armed the `beforeunload` prompt permanently.

Those handlers no longer report dirty, and the tab no longer accepts an `onDirty` prop at all — it holds no draft state to warn about.

**The Source IP Restrictions control is removed rather than repaired.** `ipRestrictions` lived only in local component state, was never sent to any endpoint, appears nowhere else in the repo, and no API enforces it. It was the other producer of the unclearable flag, and a security control that silently does nothing is worse than its absence. Nothing ever persisted, so there is nothing to migrate. Implementing it for real needs an API + enforcement in the tunnel path — worth its own issue if wanted.

> Note: `event-logs` wires `onDirty` with no `onSave` too, so it may share the shape of this bug. Left alone — it was not reported and its editor does hold draft state, so it needs its own look.

## 2. Quick Support "can never be shown again"

The reporter was right, though not for the expected reason — worth reading before merging.

The server stores only a SHA-256 hash and no endpoint returns the code, so it genuinely cannot be recovered across page loads. **But `created` is never cleared**, so re-selecting the session minted in *this* page load re-renders the same code: click another history row and back, and there it is.

I kept that behaviour deliberately — same operator, same tab, already saw the code, and burning a live session on a stray click is the worse outcome — and corrected the copy instead to state the real constraint (Breeze keeps only a hash; leaving or reloading loses it for good). Updated in all 8 locales. Two code comments that asserted the stronger, false guarantee are corrected too.

**If you would rather enforce the original claim** (clear `created` on navigating away, so it truly is shown once), that is a ~3-line change and no locale churn — say the word and I will flip it.

## 3. Audit-log action names untranslated in pt-BR

`ACTION_DISPLAY` was a hardcoded English map in `lib/auditFormat.ts`, so action names rendered in English under every locale while the chrome around them translated correctly. The catalog moves into the locale files as `admin:audit.actions` (46 entries × 8 locales) behind a new `useAuditActionFormatter` hook. The in-code English map stays as the last-resort fallback — the API emits many hundreds of action codes and only the common ones are worth translating; unmapped ones still degrade to the prettifier rather than leaking the raw enum.

The catalog node is read **whole** via i18next `returnObjects` rather than key by key, because action codes contain dots that i18next treats as key separators, and a code that is both a leaf and a prefix of another (`agent.command` vs `agent.command.result.submit`) cannot be expressed as a nested tree at all.

`AuditLogDetail` rendered the raw dotted enum in two places (the list view already formatted); it now shows the label, with the raw code alongside it for support purposes.

## Verification

- `apps/web` affected suites: **79 files / 675 tests passed** (`src/components/{settings,audit,remote,dashboard}`, `src/lib/auditFormat.test.ts`).
- All 7 i18n contract suites: **105 passed**, including `localeParity` and `keyUsage`.
- `tsc --noEmit` clean; `astro check` 0 errors (2 pre-existing warnings in an unrelated file).
- **Mutation-tested the two guards that matter** — reintroducing each bug fails them:
  - restoring `onDirty={handleDirty}` on the tab → the new `OrgSettingsPage` prop-contract test fails.
  - making `formatAuditAction` ignore the locale catalog → the new pt-BR test fails.
- A drift test pins the English locale catalog to the in-code fallback map. Locale parity only compares locales *against English*, so it cannot catch English itself drifting from the code.

## Overlap with #3452

None. The sibling QA-sweep issue (#3452) had no open PR at the time of writing, and these changes touch only the Remote Access settings tab, the Quick Support page, and the audit/auditFormat surfaces. The one shared-ish area is the locale JSONs: this PR **removes** `orgRemoteAccessSettings.ipRestrictions` from `settings.json`, **edits** `quickSupport.code.shownOnce` in `remote.json`, and **adds** `audit.actions` to `admin.json`, in all 8 locales. If #3452 also edits those files, expect textual conflicts in the locale JSONs only — no logic overlap.


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` + `silent-failure-hunter`.
**Findings:** 0 raised by either reviewer; 0 outstanding. The code reviewer independently verified the one design risk I was unsure of — that `useMemo(..., [t])` recomputes on locale switch, since react-i18next rebinds `t` via `getFixedT` when the language changes — so the catalog does not go stale after a language change.
**Tests:** `apps/web` affected suites green — 79 files / 675 tests (`components/{settings,audit,remote,dashboard}` + `lib/auditFormat.test.ts`); all 7 i18n contract suites green (105 tests, incl. `localeParity` and `keyUsage`); `no-silent-mutations` green (105). `tsc --noEmit` clean; `astro check` 0 errors. **Full CI green** (Test Web, Type Check, Lint, Test API, all 4 Integration shards).
**Mutation-tested** the two guards that matter — restoring `onDirty={handleDirty}` on the Remote Access tab fails the new prop-contract test, and making `formatAuditAction` ignore the locale catalog fails the new pt-BR test. Neither passes vacuously.

**Two things for the maintainer, both called out in the PR body:**

1. **Item 2 is a judgement call I'd like confirmed.** The reporter was right that the code re-appears, but the mechanism isn't the reopen feature — `created` is never cleared, so re-selecting the session minted in the *same page load* re-renders it. I kept that (same operator, same tab, already saw the code; losing a live session to a stray click is worse) and corrected the copy to the true constraint instead. If you'd rather enforce the original "shown once" claim, that's ~3 lines and no locale churn.
2. **Removed UI:** the Source IP Restrictions control is deleted, not repaired — it was local-only state, never sent anywhere, enforced by no API. Nothing persisted, so nothing to migrate, but it is a visible control disappearing. Worth its own issue if you want it built for real.

**Pre-existing, not touched:** `OrgRemoteAccessSettings.tsx` uses bare local `error` state rather than `runAction`, and sits outside `TARGET_GLOBS` so the guard test is inert for it. Unchanged by this PR — a follow-up candidate.

Issue #2421, opened 2026-07-13
[Web] Hash-read-in-useState hydration mismatch — two live regressions from #2383 plus ~20 latent pages


ToddHebebrand:
## Description

The SSR hydration-mismatch class #2416 fixed on `IntegrationsPage` — reading `window.location.hash` inside a lazy `useState` initializer of a `client:load` island — is not contained:

**Introduced in this range by #2383 (live regressions):**
- `apps/web/src/components/configurationPolicies/ConfigPolicyDetailPage.tsx:139-153` — `tabFromHash()` as `useState` initializer.
- `apps/web/src/components/software/SoftwarePage.tsx:11-27` — same pattern.

Deep-linking to e.g. `/configuration-policies/<id>#backup` makes React discard the SSR tree on hydration: visible tab flash + hydration-mismatch console error on every deep link.

**Latent instances (pre-existing, same pattern):** `DevicesPage.tsx:107-131`, `OrganizationsPage.tsx:47-50`, `BackupDashboard.tsx:66-70`, `DRDashboard.tsx:84-87` (the four #2416 explicitly deferred), plus `AuthPage.tsx:17-24`, `DeviceDetails.tsx:179-199` (the CLAUDE.md reference example for the hash pattern), `NetworkDeviceDetailPage`, `VulnerabilitiesPage`, `EdrPage`, `PamPage`, `MonitoringPage`, `PatchesPage`, `DnsSecurityPage`, `SensitiveDataPage`, `AiForOfficePage`, `SiteDetailPage`, `VulnerabilityFleetPage`, `TicketsPage`, `AuditBaselinesPage`, `InvoicesPage`, `QuotesPage`, `ContractsList`, `TimesheetPage`.

## Proposed Fix

Phase 1 (pre-release): extract a shared `useHashState`/`useHashTab` hook that returns the SSR-safe default on first render and applies the hash in a post-mount effect (the #2416 fix pattern), and apply it to the two #2383 regressions.

Phase 2 (mechanical sweep): migrate the latent pages to the hook. One PR, no behavior change beyond eliminating the mismatch.

## Affected Files
- `apps/web/src/components/configurationPolicies/ConfigPolicyDetailPage.tsx:139-153`
- `apps/web/src/components/software/SoftwarePage.tsx:11-27`
- (sweep list above)

Found during pre-v0.95.0 whole-range review (v0.94.0..main); follow-up to #2416/#2383.

ToddHebebrand:
[...]

PR #2322, opened 2026-07-11
feat(onedrive-helper): Phase 4 — signed-in UPN reporting + Entra graph_group library targeting

ToddHebebrand:
## Summary

Phase 4 of the OneDrive/SharePoint Helper: per-user (Entra group) library targeting. Agents report which UPNs are signed in to OneDrive Business; the server resolves each UPN's transitive Entra group membership via Microsoft Graph and tags `graph_group` libraries with the UPNs allowed to mount them; the Windows agent then mounts those libraries **per session**, only when the session's signed-in UPN is allowlisted. Fail-closed at every step.

Stacked on #2318 (Phase 3 web UI); base branch is `feat/onedrive-helper-phase3`. Plan: `docs/superpowers/plans/2026-07-09-onedrive-helper-phase4-graph-group-tagging.md`.

## How it works

1. **Agent → server:** heartbeat device state gains `signedInUpns` (per-session `HKU\<SID>\...\Business1\UserEmail`, deduped case-insensitively, capped at 16 to match the wire schema) → stored in `onedrive_device_state.signed_in_upns` (jsonb).
2. **Server delivery tagging:** for each enabled `graph_group` library with a `groupId`, delivery resolves each reported UPN's transitive group membership (Graph `transitiveMemberOf`) through a 30-min TTL cache keyed `${orgId}:${upn.toLowerCase()}` (errors are never cached) and emits `allowedUpns` per library. Resolution failures log **device id + error code only — never UPNs** — and fail closed (`allowedUpns: []`).
3. **Agent apply:** `PartitionLibraries(rules, isLocalGroupMember, sessionUpn)` applies a `graph_group` rule iff the session UPN is in `AllowedUpns` (case-insensitive); otherwise the rule stays **pending** — not entitled, not mounted, and excluded from drift.

## Staleness bounds (documented in plan + ledger)

- **Sign-out:** 0s for new mounts — the live per-session UPN is read at Apply time.
- **Group removal:** worst case ≈ cache TTL 30 min + heartbeat 60 s (~31 min; up to ~90 min at max heartbeat config).
- Already-mounted libraries persist by design until Sub-project B (unmount/scrub).

## Review

Every task went through implement → review (→ fix → re-review) under SDD; implementation Tasks 2–5 were built by Codex (gpt-5.6-sol) with controller-run gates. Whole-branch final review verdict "with fixes"; all three fixes landed pre-merge and were re-reviewed **APPROVED**:

[...]

PR #3636, opened 2026-08-16
fix(web,api,agent): OneDrive Helper polish batch (#2336)

### LINKED ISSUE #2336: OneDrive Helper: polish batch from phase 2-4 reviews (UI minors + small agent/server edges)
Non-blocking minors accumulated across the OneDrive Helper review waves (#2311, #2318, #2322, #2332), batched for one polish pass:

**Web UI (from Phase 3 reviews):**
- Device OneDrive tab: value-based React keys in list rendering
- `kfmProtected` rollup caption mismatch (reads as "of signedIn" when the denominator is all reporting devices)
- `GET /onedrive/devices/:id/state` selects the full state row where the panel uses a subset (over-wide select)
- Policy tab: Override button not disabled while `saving`; `toPayload` should trim string inputs
- Library picker: async effects lack unmount guards (matches the SoftwareCatalogPicker idiom, but worth fixing when touched)
- `#onedrive_helper` URL hash does not preselect the feature tab for overflow tabs (lands on Overview)

**Agent:**
- AutoMount timer poke targets the Business1 slot only — for a sparse multi-account profile (Business1 unlinked, Business2+ present) convergence waits on OneDrive's own timer (latency-only edge)
- `LookupSID` result cache before broad `local_ad_group` rollout
- Site-qualified drift matching (tenant-cache paths are `<Site> - <Library>`; displayName-only matching can false-positive across sites — bit us once in live QA)

**Server:**
- Per-UPN Graph resolutions run sequentially (fine at 1-3 UPNs; parallelize with a small concurrency cap if multi-session hosts become common)

None are correctness/security issues; all were explicitly reviewed and deferred. Refs #2311 #2318 #2322 #2332.

ToddHebebrand:
Batches the non-blocking minors deferred across the OneDrive Helper review waves (#2311, #2318, #2322, #2332) into one pass. No new tables, columns, or migrations.

## Web

| Item | Fix |
|---|---|
| `kfmProtected` rollup caption mismatch | The tile captioned itself "of {signedIn} signed in", but the server derives `kfmProtected` from **every** reporting device — as does the tile's own `kfm-gap` row filter. Caption and warning threshold now both use `total`. |
| Override not disabled while `saving` | `FeatureTabShell`'s Override button was gated only on `saveDisabled`, so a double-click during the create round-trip fired `persist(null)` twice. Now gated on `saving` too — this is the shared shell, so it applies to **every** feature tab. |
| `toPayload` should trim string inputs | Every free-text field is trimmed on the way out; trimmed-to-empty optionals collapse to `null`. `libraryError` already validated against trimmed values, so an all-whitespace group name passed Save and persisted as a targeting value the agent can never match. |
| Value-based React keys | Library rows are keyed by a stable client-side `rowKey` rather than the array index, so removing a row no longer hands the next row's props to the removed row's DOM node. |
| Library picker unmount guards | Async effects are gated on a request-generation token, invalidated on unmount, on `orgId` change, and on Retry. |
| `#onedrive_helper` hash → overflow tab | **Already fixed** by the `useHashTab` work in #2421; verified and pinned with a regression test rather than re-fixed. |

## API

- `GET /onedrive/devices/:id/state` selected the whole state row; it now projects exactly the columns the panel consumes, so a future column cannot silently widen the response.
- Per-UPN Graph membership resolutions ran strictly serially under a 15s budget. On a multi-session host (RDS/VDI) the later users' lookups fell off the end of that budget and their libraries silently never mounted. They now run through a fixed **4-way pool** — capped deliberately, since these calls share one org's Graph token and rate limit. Results are applied in the original UPN order, so `allowedUpns` stays deterministic regardless of interleaving.

## Agent

- `pokeAutoMountTimer` poked the `Business1` slot only. Account slots go sparse after an unlink, so a Business2/3-only profile got a silent no-op and waited on OneDrive's own multi-hour timer. All populated slots are now poked.
- Group-name → SID lookups are memoized for 10 minutes ahead of a broader `local_ad_group` rollout — `LookupSID` is a DC round-trip for a domain group and ran once per rule per session per heartbeat. Only successes are cached, so a group created after a miss still converges.
- `ComputeDrift` is now **site-qualified and one-to-one**. Matching the display name anywhere in any mounted path could not distinguish `Contoso HR - Documents` from `Contoso Legal - Documents`, so with two sites entitling a same-named library ("Documents" is the default on every SharePoint site) a single mount suppressed the drift report for both — a false negative seen in live QA. Each mounted folder now satisfies at most one rule, resolved in three passes: site-confirmed pairing, name-only pairing, then the legacy substring scan so an unrecognized cache shape degrades to the old behaviour instead of reporting drift on everything.

## Verification

- `agent`: `go test -race ./internal/onedrivehelper/...` green; `GOOS=windows go build ./...` green (the poke/SID changes are Windows-only).
- `apps/web`: 240 tests green across `src/components/configurationPolicies/` + the OneDrive fleet/device suites; `tsc --noEmit` clean.
- `apps/api`: `onedrive.test.ts`, `onedrive.state.test.ts`, `agents/heartbeat.test.ts` (136) green; `tsc --noEmit` clean.
- Integration (live DB + Redis): `onedrive-helper-config-delivery` + `onedrive-helper-write-path`, 29 green.
- New tests were **mutation-checked** — each was re-run against the pre-fix behavior and confirmed to fail. Three first-draft assertions passed under mutation and were rewritten: the row-key test (controlled inputs mask index keys, so it now asserts DOM node identity), the picker test (React 19 no longer warns on setState-after-unmount, so it now asserts a superseded org's late response cannot clobber the current one), and the concurrency test (verified failing at `TAGGING_CONCURRENCY = 1`).

Closes #2336

### DIFF
--- a/agent/internal/onedrivehelper/onedrivehelper.go
+++ b/agent/internal/onedrivehelper/onedrivehelper.go
@@ -184,28 +184,126 @@ func TenantIDFromComposite(libraryID string) string {
 	return ""
 }
 
-// ComputeDrift flags applied libraries whose display name matches no mounted
-// local folder path. OneDrive's tenant cache stores mounted scopes as local
-// folder paths of the form "<Org> - <LibraryName>", so a case-insensitive
-// substring match on the display name is the practical detection (validated
-// against the live-spike cache shape, see the 2026-06-19 spike doc). A rule
-// the user previously stop-synced will never re-mount — that is exactly the
-// drift this surfaces (spec: report, don't rewrite forever).
+// pathBase returns the last path component of a mounted local folder path,
+// tolerating either separator (the values come from a Windows registry cache
+// but the logic is unit-tested on every OS).
+func pathBase(p string) string {
+	if i := strings.LastIndexAny(p, `\/`); i >= 0 {
+		return p[i+1:]
+	}
+	return p
+}
+
+// mountSiteForLibrary reports whether a mounted path is a mount OF the named
+// library, and if so returns the site-title portion of its folder name.
+//
+// OneDrive names these folders "<Site title> - <Library name>" (or bare
+// "<Library name>" for a library whose site title OneDrive omits). Matching on
+// the SUFFIX rather than splitting on " - " keeps library names that themselves
+// contain " - " intact, which a naive split would tear apart.
+func mountSiteForLibrary(p, displayName string) (site string, ok bool) {
+	base := pathBase(p)
+	if strings.EqualFold(base, displayName) {
+		return "", true
+	}
+	suffix := " - " + displayName
+	if len(base) > len(suffix) && strings.EqualFold(base[len(base)-len(suffix):], suffix) {
+		return base[:len(base)-len(suffix)], true
+	}
+	return "", false
+}
+
+// siteHintFromURL derives a comparable token from a SharePoint site URL — the
+// last non-empty path segment, e.g. ".../sites/Marketing" → "Marketing".
+//
+// It is only a HINT: the folder name carries the site TITLE, which is free text
+// and frequently differs from the URL slug ("Marketing Team" vs. /sites/mktg).
+// So it is used strictly to break a tie between several identically-named
+// libraries and never to turn a single unambiguous match into drift.
+func siteHintFromURL(siteURL string) string {
+	trimmed := strings.TrimRight(siteURL, "/")
+	if i := strings.LastIndex(trimmed, "/"); i >= 0 {
+		return trimmed[i+1:]
+	}
+	return ""
+}
+
+// ComputeDrift flags applied libraries that OneDrive did not actually mount. A
+// rule the user previously stop-synced will never re-mount — that is exactly
+// the drift this surfaces (spec: report, don't rewrite forever).
+//
+// Matching is site-qualified and one-to-one. The old check asked, per rule,
+// "does the display name appear anywhere in any mounted path?", which cannot
+// tell "Contoso HR - Documents" from "Contoso Legal - Documents": two sites
+// entitling a same-named library (and "Documents" is the default name on EVERY
+// SharePoint site) meant a single mount suppressed the drift report for both.
+// That false negative was seen in live QA.
+//
+// Treating it as an assignment fixes it: each mounted folder can satisfy at
+// most one rule, so two rules and one mount always leaves one rule drifted.
+// Which one is decided in three passes, strongest evidence first:
+//
+//  1. Folder name is "<site> - <library>" AND the site title contains the hint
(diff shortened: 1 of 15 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — code-reviewer + silent-failure-hunter, both against `origin/main...HEAD`.

**Findings:** 1 raised → addressed in `92231bb`; 0 outstanding.
- **code-reviewer: 0 findings.** Explicitly cleared the five risk areas — worker-pool fail-closed semantics and deterministic ordering, `ComputeDrift` bounds/determinism, `sidLookupCache` lock scope (no lock held across the DC round-trip) and per-slot registry key closing, the picker's token guard across effect re-run/unmount/Retry, and that `rowKey` cannot reach the persisted payload.
- **silent-failure-hunter: 1 real finding.** When several same-named libraries stay unresolved after the site-confirmed pass, `ComputeDrift`'s name-only pass assigns mounts by position — the *count* of unmounted libraries is right, but which library the entry names is a guess, and reporting it as plain `not_mounted` presented that coin-flip as fact (a tech would chase the wrong site). Fixed: those entries now report `not_mounted_ambiguous`; a site-confirmed pairing keeps `not_mounted`, so the common case is unchanged. The hunter separately confirmed the concurrency pool's fail-closed/`budgetExhausted` behavior, the success-only SID caching, and that no dropped response can strand the picker in a permanent spinner.

**Tests:** all green, and each new assertion was mutation-checked (re-run against the pre-fix behavior and confirmed to fail).
- `agent`: `go test -race ./internal/onedrivehelper/...`; `GOOS=windows go build ./...`; `gofmt` clean.
- `apps/web`: 240 tests across `src/components/configurationPolicies/` plus the OneDrive fleet/device suites; `tsc --noEmit` clean.
- `apps/api`: `onedrive.test.ts` / `onedrive.state.test.ts` / `agents/heartbeat.test.ts` (136); `tsc --noEmit` clean.
- Integration (live DB + Redis): `onedrive-helper-config-delivery` + `onedrive-helper-write-path`, 29 passed.
- **CI: 54/54 passing, 0 failing** — including Test API, Test Web, Test Agent (+race, Windows), Type Check, and all four Integration Tests shards.

Three first-draft assertions passed under mutation and were rewritten rather than kept: the row-key test (controlled inputs mask index keys — now asserts DOM node identity), the picker unmount test (React 19 no longer warns on setState-after-unmount — now asserts a superseded org's late response cannot clobber the current one), and the concurrency test (verified failing at `TAGGING_CONCURRENCY = 1`).

**Status:** review-clean, CI green, awaiting maintainer merge. Not merging or closing #2336 — the OneDrive UI/agent behavior is worth a human eye before close.

PR #2177, opened 2026-07-03
refactor(billing): shared list chrome, pill-role layering, plus follow-up tests (#2151)

ToddHebebrand:
Bundled the **well-scoped** items from the #2150 polishing round-2 punch list (#2151). None blocked the merge; each is small and independent.

## Done

**Small refactors**
- **Shared row-link className** — extracted the anchor class (duplicated 4× across QuotesPage/InvoicesPage/ContractsList) into `ROW_LINK_CLASS` in a new `billing/shared/listChrome.ts`.
- **Shared `writeHashFilters` helper** — the quotes/invoices bare-`#` clear-residue fix (strip the fragment via `replaceState` instead of leaving a dangling `#`) is now a shared helper. **ContractsList's `writeFilters` never got that fix** — it did, so contracts no longer leaves a bare `#` on clear. (Regression test added.)
- **`StatusPillRole` / `STATUS_PILL` moved** out of the invoice-specific `components/billing/invoiceTypes.ts` into `billing/shared/statusPillRoles.ts`, so `lib/api/contracts.ts` type-imports from the shared sub-layer rather than an api→components layering violation. `invoiceTypes` re-exports both, so existing consumers are unaffected. (Filename is `statusPillRoles.ts`, not `statusPill.ts`, to avoid a case-insensitive-FS collision with `StatusPill.tsx`.)

**Visual/a11y**
- Invoice notes/terms textareas now include `isPending(...)` in their `disabled` expression (busy cue, matching the quote editor's terms field). The inFlight guard already prevents a double-PATCH; this just makes the window visible.

**Tests added**
- Contracts sort comparators `org` / `status` / `estimate` / `start` (only `name` was covered) — org sorts by resolved name, estimate is numeric (90 < 100, not string order), start keeps null-start rows last in **both** directions. Plus the clear-residue regression, and a one-line comment on the intentional asc-first toggle (siblings are desc-first).
- Auto-renew checkbox in-flight-disable lock (the describe block covered auto-issue/timing/interval but not auto-renew).
- Invoice price resync guard — mid-edit price survives a server echo; a settled field still re-adopts the server value — plus the notes busy-cue disable.
- Mobile invoice-card em-dash draft link + accessible name (`invoices-card-link-*` draft case; only the desktop row was tested).

## Deferred (left for the issue / follow-ups)
These are larger, vaguer, or need a design call — intentionally **not** in this PR so #2151 can stay open or spawn follow-ups:
[...]

Issue #2151, opened 2026-07-02
Billing surfaces: follow-ups from polishing round 2 review


ToddHebebrand:
**Bundled minor findings from the PR #2150 review rounds** — none blocked the merge; each is small and independent.

**Tests to add**
- Contracts sort comparators `org`/`status`/`estimate`/`start` are untested (only `name` asc/desc is covered) — `apps/web/src/components/contracts/ContractsList.search-sort.test.tsx`. Also worth a one-line comment on the intentional asc-first toggle (siblings are desc-first).
- No in-flight-disable test for the auto-renew checkbox specifically (`ContractEditor.autosave.test.tsx` covers auto-issue/timing/interval with the same mechanism).
- Invoice editor price resync guard untested (qty path is tested; the code is symmetric) — `apps/web/src/components/billing/InvoiceEditor.tsx`.
- Mobile invoice-card em-dash draft link untested (`invoices-card-link-*` draft case).

**Small refactors**
- Extract the row-link anchor className string (duplicated 4× across QuotesPage/InvoicesPage/ContractsList) into a shared const.
- Extract a shared `writeHashFilters` helper: quotes/invoices carry a duplicated bare-`#` residue fix that contracts never received (`ContractsList.tsx` `writeFilters` still leaves `#` on clear).
- Move `StatusPillRole` out of `components/billing/invoiceTypes.ts` into `billing/shared/` — `lib/api/contracts.ts` currently type-imports from a component file (api→components layering).

**Visual/a11y odds and ends**
- Invoice notes/terms textareas don't include `isPending(...)` in their disabled expression (quote editor's terms field does) — visual busy-cue only; the inFlight guard already prevents double-PATCH.
- Duplicate `aria-label="Quantity"` across pick/manual/line inputs in both editors (pre-existing pattern) — worth one a11y pass.
- Quote editor's live-totals SR announcement fires once on mount (~800ms after load) before any edit; consider suppressing the first fire.
- Global toast container now lifts `lg:bottom-24` to clear the editor rail — fine today, but it's a shared-container smell if more pages grow right rails.
- Invoice PDF fallback wordmark is 'Invoice' while the preview's shared branding resolver falls back to 'Proposal' (unreachable under partner scope in practice) — align when convenient.
- Quote editor line grid at 1280px with the sidebar open still horizontally scrolls to reach the Total column (fits with sidebar collapsed); a real fix needs column-priority restructuring.

Refs #2150.

[...]

PR #3638, opened 2026-08-16
fix(web,api): billing punch-list a11y, toast anchor, and PDF wordmark (#2151)

ToddHebebrand:
Refs #2151 (line-grid column-priority item still open — see "Left out" below).

Four of the five items left on the #2151 punch list after the partial fix in #2177. The fifth is deliberately left out — see the bottom.

## 1. Duplicate `aria-label="Quantity"` (a11y pass)

Every qty and unit-price input in both editors answered to the same accessible name, so a screen reader's form-controls list was a run of indistinguishable entries with nothing to say which line each belonged to. Five collisions: quote ghost row + persisted row, invoice manual-add + catalog-pick + persisted row.

- Persisted rows: `Quantity for <item>` / `Unit price for <item>`. Deliberately the **persisted** title, not the live draft — an accessible name that changes on every keystroke is itself SR churn. Untitled quote lines reuse the existing `this line` phrase the bulk-select checkbox already uses; invoice lines fall back name → description → `this line`.
- New-line and catalog-pick rows name themselves (`New line quantity`, and the pick row is scoped to the picked item).
- New keys added across all 8 locale catalogs (parity suite green).

## 2. Quote live-totals SR announcement fired on mount

The rail's `role="status"` node announced ~800ms after load on a quote nobody had touched, despite an existing skip-the-first-render guard.

Root cause: the guard skipped only the **first run** of a debounce effect whose dep was the announcement **sentence**. That is not the same thing as "the totals changed" — any later render producing different sentence *text* while the figures sit still fires the effect a second time, past the guard. i18n resources arriving after first paint do exactly that (raw key → real copy on an untouched quote).

The effect now keys on a `srTotalsKey` built from the figures, so a copy-only change is a no-op while a real edit still announces (debounced, last-write-wins). The sentence is read at fire time via a ref so a burst that settles after the i18n swap still announces real copy.

`QuoteEditor.srtotals.test.tsx` is mutation-verified: reverting the fix turns the late-translation case red. Its `useTranslation` mock exists precisely to make that non-vacuous — without it, nothing changes the sentence text and the buggy version passes too.

## 3. Toast container `lg:bottom-24`

The issue flagged this as a shared-container smell, and it was: one page's sticky right rail set a global lift, so every page's toasts moved, and the next page to grow a rail would have re-litigated the single number.

The offset is now `--breeze-toast-bottom` (default `1.5rem`, rule in `globals.css`), and the page that has the rail opts in with `useToastRailOffset()` — refcounted, `lg`-only, cleaned up on unmount. Only `QuoteEditor` opts in today. Deliberately a separate module from `Toast.tsx`: that module is `vi.mock`ed wholesale by a large number of suites, and a layout hook exported from there would break every one of them.

## 4. PDF fallback wordmark

`invoicePdf` printed `Invoice` and the shared quote resolver `Proposal` in the wordmark slot when the partner row read back empty — a document-type word where a company name belongs, next to the header's own INVOICE/PROPOSAL eyebrow.

The realistic empty-partner case is the org-scoped RLS zero-row read `quoteBranding.ts`'s own header warns about, not a nameless partner — and the document still carries the seller name frozen onto it at send. Both sides now fall through `seller.name` before the literal, which is exactly what the two web previews (`InvoiceDocument` / `QuoteDocument`) already do, so PDF and preview degrade identically instead of diverging on the same missing row.

Unit-covered on the quote side (`quoteBranding.test.ts`, incl. live-partner-wins and null-seller-name cases). The invoice side is the same one-line change inside the unexported `loadInvoiceForRender`, reachable only through the DB path, so it is covered by the existing integration suite rather than a new unit test.

## Left out: quote-editor line-grid horizontal scroll

The punch list itself says this "needs column-priority restructuring", and that holds up. The grid already sizes Item flexibly and Qty/Price/Total to content, and at 1280px with the 300px rail open the content column is comfortably wider than the table's `min-w-[36rem]` floor — so the overflow is driven by *content* width (long item names plus the internal cost band), not by container arithmetic. Fixing it properly means deciding which columns yield first (truncation/stacking priority), which is a design change rather than a punch-list tidy. Left for a follow-up so this batch stays mechanical.

## Overlap with #3631

#3631 (block edit-in-place) also touches `QuoteEditor.tsx` and the locale `billing.json` files. Overlap here is small and in different regions: this PR adds an import + a one-line hook call near the top of `QuoteEditor` and rewrites the SR-debounce effect; locale changes are new keys in `quotes.editor.line` / `quotes.editor.ghost` / `invoiceEditor.fields`. No shared blocks, but whichever lands second should expect a trivial rebase on the locale files. `QuoteBlockCard.tsx` and `QuoteBlockContentForms.tsx` are untouched here.

## Verification

- `apps/web`: 99 files / 811 tests green (`src/components/billing`, `src/lib/i18n`, `src/components/shared`), including the 4 new suites.
- `apps/api`: `quoteBranding.test.ts` + `invoicePdf.test.ts`, 28 tests green.
- `astro check` clean (0 errors); `tsc --noEmit` clean in `apps/api`.
- No schema, migration, or tenancy surface touched.

### DIFF
--- a/apps/api/src/services/invoicePdf.ts
+++ b/apps/api/src/services/invoicePdf.ts
@@ -394,7 +394,9 @@ async function loadInvoiceForRender(invoiceId: string): Promise<{ invoice: Invoi
       // snapshot above still carries the company name — printing the word
       // "Invoice" in the wordmark slot (beside the header's own INVOICE
       // eyebrow) threw that away.
-      partnerName: partner?.name ?? (invoice.sellerSnapshot as SellerSnapshot | null)?.name ?? 'Invoice',
+      // `||`, not `??`: neither name column is constrained non-empty, and a
+      // blank wordmark is a worse document than the generic word.
+      partnerName: partner?.name || (invoice.sellerSnapshot as SellerSnapshot | null)?.name || 'Invoice',
       logoUrl: branding?.logoUrl ?? null,
       primaryColor: branding?.primaryColor ?? null,
       footerText: invoice.terms ?? partner?.invoiceFooter ?? branding?.footerText ?? null,
--- a/apps/api/src/services/invoicePdf.ts
+++ b/apps/api/src/services/invoicePdf.ts
@@ -386,7 +386,15 @@ async function loadInvoiceForRender(invoiceId: string): Promise<{ invoice: Invoi
     invoice,
     lines,
     branding: {
-      partnerName: partner?.name ?? 'Invoice',
+      // Seller-snapshot fallback before the document-type literal (#2151), so
+      // the invoice PDF degrades the same way the quote side does
+      // (resolveQuoteBranding) and the same way both web previews already do
+      // (`branding.partnerName || seller.name`). A partner row that reads back
+      // empty is an RLS/scope artifact, not a nameless partner, and the frozen
+      // snapshot above still carries the company name — printing the word
+      // "Invoice" in the wordmark slot (beside the header's own INVOICE
+      // eyebrow) threw that away.
+      partnerName: partner?.name ?? (invoice.sellerSnapshot as SellerSnapshot | null)?.name ?? 'Invoice',
       logoUrl: branding?.logoUrl ?? null,
       primaryColor: branding?.primaryColor ?? null,
       footerText: invoice.terms ?? partner?.invoiceFooter ?? branding?.footerText ?? null,
--- a/apps/api/src/services/quoteBranding.test.ts
+++ b/apps/api/src/services/quoteBranding.test.ts
@@ -96,6 +96,18 @@ describe('resolveQuoteBranding', () => {
     expect(b.partnerName).toBe('Proposal');
   });
 
+  it('blank names are skipped, not printed — a wordmark is never empty', async () => {
+    // Neither partners.name nor the snapshot's name is constrained non-empty,
+    // and `??` would happily hand an empty string to the renderer.
+    queue({ ...basePartner, name: '' }, baseBrand);
+    const frozen = { name: '', address: null, phone: null, email: null, website: null };
+    expect((await resolveQuoteBranding(source({ sellerSnapshot: frozen }))).partnerName).toBe('Proposal');
+
+    queue({ ...basePartner, name: '' }, baseBrand);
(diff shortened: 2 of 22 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter, both on the full diff.

**Findings:** 1 raised (both reviewers independently), addressed in `2352c8e72`; 0 outstanding.
- Both branding fallback chains used `??`, which only skips `null`/`undefined`. Neither `partners.name` nor the frozen snapshot's name is constrained non-empty, so an empty string would have been handed to the renderer as the wordmark. Switched to `||` in `quoteBranding.ts` + `invoicePdf.ts`, with a regression case in `quoteBranding.test.ts`.
- Everything else came back clean: refcounted `useToastRailOffset` verified correct under StrictMode double-invoke, remount, and concurrent holders (no leaked attribute path); `html[data-toast-rail]` (0,1,1) outranks `:root` (0,1,0) regardless of order and sits outside Tailwind's layers; no duplicate accessible name left in either editor and no path yielding `undefined`/`[object Object]`; key parity + `{{item}}` interpolation confirmed across all 8 locales; the seller extraction in `quoteBranding.ts` is behavior-preserving.
- Both reviewers checked the new suites for vacuous coverage and cleared them. `QuoteEditor.srtotals.test.tsx` was mutation-verified directly: restoring the old sentence-keyed effect turns the late-translation case red.

**Noted, not changed:** `srFirstKey` is a per-instance ref, so if `QuoteWorkspace` ever reused one `QuoteEditor` instance across a quote switch instead of remounting, the switch would read as an edit and announce. Pre-existing — the old `srMounted` ref had the identical shape — and there is no such code path today (no `key` juggling on `<QuoteEditor>`), so it is out of scope here.

**Tests:** `apps/web` 99 files / 811 tests green (`components/billing`, `components/shared`, `lib/i18n`); `apps/api` `quoteBranding` + `invoicePdf` 29 tests green. `astro check` 0 errors, `tsc --noEmit` clean in `apps/api`, `eslint src` clean in `apps/web`.

**Status:** review-clean, awaiting maintainer merge. Issue #2151 stays open — the line-grid column-priority item is still outstanding (see the PR body for why it was left out).

PR #3642, opened 2026-08-17
fix(api): partner-wide alert-template predicate leaked cross-org templates

ToddHebebrand:
Security review 2026-08-16 §1.5 (CRITICAL).

## Exploit

`templateScopeCondition()` (`apps/api/src/routes/alertTemplates/templates.ts`) built its visibility predicate as:

```ts
or(
  eq(alertTemplates.isBuiltIn, true),
  inArray(alertTemplates.orgId, auth.accessibleOrgIds),
  eq(alertTemplates.partnerId, auth.partnerId),   // <-- selects EVERY template under the partner
)
```

The third disjunct reads as "partner-wide rows", but it wasn't. The create route denormalized `partner_id` onto **org-owned** rows, and `alert_templates` had no one-owner CHECK (22 other tables have one), so rows with **both** `org_id` and `partner_id` set were legal — and are what is stored today. That disjunct therefore matched every template under the partner and voided the `inArray(orgId, accessibleOrgIds)` restriction sitting beside it in the same `OR`.

- **Partner scope with `orgAccess: 'selected'` / `'none'` → genuinely exploitable.** A partner user restricted to org A could list org B's alert templates (names, descriptions, conditions). Partner-axis RLS is flat, so the database did not catch it.
- **Org scope → app-layer bug**, currently saved by `breeze_org_isolation`. Defense in depth was load-bearing.

## Fix

1. **Predicate** — partner-wide now means `partner_id = X AND org_id IS NULL` on both the org-scope and partner-scope branches.

   Note on the org-scope branch: the generic repo rule is "partner-wide reads must be gated on `scope === 'partner'`". `alert_templates` is a documented exception — it is one of four catalog tables (with `scripts`, `script_categories`, `script_tags`) whose RLS carries an explicit `(org_id IS NULL AND partner_id = breeze_current_partner_id())` SELECT branch (`2026-06-13-catalog-partner-read-branch.sql`), and org-scope sessions are handed `currentPartnerId` precisely so org admins can read their MSP's shared templates read-only (`bearerTokenAuth.ts`, MCP-OAUTH-06). App layer and RLS agree; removing the branch would be a functional regression, so it was narrowed rather than deleted.

2. **Writers** — the create route now writes exactly one axis. Org-owned rows get `partner_id = NULL` (both the org-scope branch and the partner-scope "one specific org" branch).

3. **Same bug class on the other global disjunct** (found reviewing the fix). `is_built_in = true` is not an ownership marker: `policyAlertBridge.ensureTemplate()` auto-creates **org-owned** rows with `isBuiltIn: true`, so a bare `is_built_in` term handed every org's policy-compliance template to every other caller — including the unscoped `GET /alert-templates/templates/built-in` endpoint. Every global-visibility predicate now says `is_built_in = true AND org_id IS NULL`. Org-owned built-ins remain visible to their own org through the org disjunct, so nothing is lost.

## Migration

`apps/api/migrations/2026-08-16-alert-templates-one-owner.sql` (plain date form — no letter infix, and it does not touch the closed `2026-08-06` block):

1. `UPDATE alert_templates SET partner_id = NULL WHERE org_id IS NOT NULL AND partner_id IS NOT NULL`, wrapped in a `DO $$ … GET DIAGNOSTICS n = ROW_COUNT; IF n > 0 THEN RAISE WARNING 'cleaned % …' …` block. **The row count is deliberately logged** — those rows are exactly the ones that could have been read cross-org through the old predicate, so the count is forensic evidence, not noise.
2. A second `DO $$` that counts and `RAISE WARNING`s ownerless non-built-in rows (reported, not modified).
3. `alert_templates_one_owner_chk` added idempotently behind a `pg_constraint` existence check.
4. `alert_templates_partner_id_idx` (also added to the Drizzle schema so `db:check-drift` stays clean).

**The constraint is `CHECK (org_id IS NULL OR partner_id IS NULL)` — "never both" — not the usual strict XOR.** `alert_templates` has two legitimate ownerless shapes: the seeded global built-ins (`0018`, `0060`, `2026-08-12`: `org_id NULL`, `partner_id NULL`, `is_built_in true`) and rows a system-scope caller created with no `orgId` in the body. A strict XOR would reject both and make `ADD CONSTRAINT` **fail on a live database**. "Never both" is precisely the security-relevant invariant — the leak came from rows with both axes set, never from rows with neither. Re-applying the whole file is a true no-op.

## Sweep

Every `alertTemplates` reader/writer in `apps/api/src` was enumerated by hand (`grep -rn "alertTemplates" apps/api/src`, 70 files):

| Call site | Verdict |
|---|---|
| `routes/alertTemplates/templates.ts` (predicate, list, built-in list, get-by-id, create) | **fixed** — the finding, plus the `is_built_in` disjunct |
| `routes/alertTemplates/helpers.ts` `getAllTemplates` / `getTemplateById` | **fixed** — `is_built_in` disjunct narrowed |
| `routes/alertTemplates/rules.ts:154` (template-exists check) | **fixed** — `is_built_in` disjunct narrowed |
| `services/aiToolsFleet.ts:1804` (`list_templates`) | **fixed** — `is_built_in` disjunct narrowed; the org branch uses `auth.orgCondition` (org axis only, no partner leak) |
| `services/policyAlertBridge.ts` `ensureTemplate` | writes `orgId` only, `partner_id` NULL — OK (it is the *source* of org-owned built-ins) |
| `services/automationRuntime.ts:772` | reads and writes on `orgId` only — OK |
| `routes/alerts/helpers.ts` `resolveAlertTemplate` | writes both supplied axes, but both callers (`routes/alerts/rules.ts:362,537`) pass XOR ownership resolved from `alert_rules` — OK, no 23514 risk |
| `db/seed.ts` | built-ins, both axes NULL — OK |

## Contract lists

| List | Status |
|---|---|
| `CORE_ORG_CASCADE_DELETE_ORDER` (`tenantCascade.ts:77`) | already registered — no change |
| `CORE_TENANT_EXPORT_POLICY` (`tenantExportPolicyRegistry.ts:54`) | already registered, all columns classified; **no new column added**, so no change needed |
| `CORE_DEVICE_CASCADE_DELETE_TABLES` / `CORE_DEVICE_ORG_DENORMALIZED_TABLES` | N/A — no `device_id` column |
| `PARTNER_TENANT_TABLES` (`rls-coverage.integration.test.ts:199`) | already registered — no change |
| `DUAL_AXIS_TENANT_TABLES` | **added** — the table has both axes and was only registered as partner-axis, so nothing asserted the dual-axis shape |

## Tests

- **`apps/api/src/routes/alertTemplates/templates.scope.test.ts` (new, no-DB, runs in `test-api`).** Renders the predicate to real SQL through `PgDialect` and pins the exact text + bound params, so the partner-scope `orgAccess: 'selected'` case proves the non-accessible org is unreachable, plus a structural check that no top-level disjunct mentions `partner_id` (or `is_built_in`) without the `org_id IS NULL` conjunct. **Mutation-verified**: reverting the conjunct fails 3 of the 6 tests. A token/substring scan would have passed on the vulnerable code, so it is asserted structurally instead.
- **`apps/api/src/__tests__/integration/alert-templates-partner-wide.integration.test.ts` (extended).** Both-axes insert → `23514` / `alert_templates_one_owner_chk`; the three legal shapes accepted; ownerless row accepted (documenting the deliberate non-XOR); and a "no stored row carries both axes" assertion proving the backfill held. The existing cross-partner forge (42501) and org-isolation cases were already there.

## What was and was not run

- ✅ `pnpm --filter @breeze/api test` on the affected paths — **384 tests, 24 files, all passing** (`routes/alertTemplates`, `routes/alerts`, `aiToolsFleet`, `policyAlertBridge`, `db/autoMigrate`).
- ✅ `scripts/check-migration-naming.sh` — OK.
- ✅ `tsc --noEmit` on `@breeze/api`.
- ❌ **Integration/RLS suites were NOT run** — no Postgres was reachable in this environment. They will run in the blocking `integration-test` job.
- ❌ **`pnpm db:check-drift` was NOT run** — same reason. The new index was added to the Drizzle schema to keep drift clean, but that is unverified locally.


bdunncompany:
Not a finding against this PR — a note on how far the pattern reaches, since you're sweeping this class tonight.

**"Partner-axis RLS is flat" holds at the function level.** `breeze_has_partner_access` is `system → TRUE`, `NULL → FALSE`, else `target = ANY(breeze_accessible_partner_ids())`. Auth narrows `accessibleOrgIds` from `partner_users.orgAccess`, but computes partner access separately and hands every partner-scope user `[partnerId]`. So the partner disjunct is unaffected by an `orgAccess: 'selected'/'none'` restriction — which is exactly the property this PR's predicate was leaning on.

**Two other dual-owner tables have the same shape**, both with `has_partner_access(partner_id) OR has_org_access(org_id)` RLS and an at-least-one (not XOR) owner CHECK, so rows legitimately carry both axes:

- `software_policy_audit` — a device event stores the device's org *and* the policy's partner, deliberately (`2026-07-01-software-policies-partner-ownership.sql`).
- `software_remediation_requests` — mine, from #3585. For a partner-wide policy the writer sets `org_id` = the *device's* org and `partner_id` = the policy's partner.

**To be clear about what this is and isn't:** I went looking for a reachable disclosure and did not find one. Neither table has a partner-facing GET, list, AI tool, or service reader. `software_remediation_requests` is worker-only (atomic consume, mint, system-context cleanup) plus cascade metadata; tenant export is platform-admin + MFA and filters `WHERE org_id = requestedOrg`. The remediate route filters devices through `auth.orgCondition` before scheduling, so this can't be used to schedule against another org either. **Presently zero confidentiality impact through the supported API.**

**And for my table the behavior is pinned as intentional** — `softwareRemediationRequestsRls.integration.test.ts` builds a dual-axis row, constructs a partner context with `accessibleOrgIds: []`, and asserts the row stays visible specifically via `breeze_has_partner_access`. I wrote that test, so I'm not reporting my own design as a surprise.

**The actual risk is the future reader.** Someone adds a list endpoint or AI tool over one of these tables, trusts RLS the way this PR's predicate did, and inherits a cross-org read for restricted partner users. Worth noting that `software_policy_audit.details` can carry policy names, device ids, software names and error text — so unlike the remediation rows, that one isn't just ids and timestamps.

If it's useful I can turn this into a short "dual-owner tables need an explicit org filter on any partner-scope read" note in the RLS docs, or a coverage assertion. Happy to leave it here if #3643 already generalizes it — I couldn't read that PR's diff to check.

PR #3645, opened 2026-08-17
fix(api): stop agent-driven global patch-catalog poisoning and adopt SSRF guard on backup probes

ToddHebebrand:
Two verified HIGH findings from the 2026-08-16 security review (`internal/2026-08-16-security-review-findings.md` §1.6 and §1.7). Both are cross-tenant issues that the existing contract tests did not cover.

---

## §1.6 — a compromised agent could poison the GLOBAL patch catalog

`apps/api/src/db/schema/patches.ts:111` — `patches` is deliberately global: unique on `(source, external_id)`, no `org_id`, not in the org cascade. Every tenant's approval engine reads it, and `apps/api/src/routes/agents/patches.ts` upserts it directly from agent-submitted scan data.

PR #3556 (already merged into `main`) defended the **identity** columns — the `curated` flag gates `title`/`vendor`, `packageId` uses `fillIfNull`. This PR closes the remaining **volatile**-column gap, which #3556 did not touch:

| Column | Before | After |
|---|---|---|
| `severity` | agent value won whenever it wasn't `'unknown'` — an agent could **lower** it | fill-or-**raise** only (`raiseOnlySeverity`, `patches.ts:104`), applied at `patches.ts:247` |
| `version` | `updateIfReported` — agent rewrote it every scan | `fillIfNull` (`patches.ts:255`), matching what the installed path at `:360` already did |

**Why it mattered.** A global severity downgrade makes *other tenants'* auto-approval rules mis-classify a critical patch. `patches.version` is the target of the version pin / block rules in `patchApprovalEvaluator` (`comparePatchVersions` / `pinnedVersion`), so a poisoned value could defeat another tenant's pin. The inline comment already asserted the CVE worker "only ever raises severity" (`jobs/cveEnrichmentWorker.ts:112-114` — true), but the agent path could lower it; `raiseOnlySeverity` makes the code match the claim.

**Where the agent's observation went instead.** The available upgrade version genuinely *does* advance between scans, so it was not simply dropped — it moved to the tenant-scoped row:

- new column `device_patches.available_version` (`migrations/2026-08-25-device-patches-available-version.sql`, `db/schema/patches.ts`), written at `routes/agents/patches.ts:281`
- `services/patchApprovalEvaluator.ts:348` now selects `COALESCE(device_patches.available_version, patches.version)`, so a pin is evaluated against **this device's own** observation, with the catalog as fallback for pre-existing rows.

**Cascade registration.** `device_patches` is an org-cascade table, so the `ADD COLUMN` required a `CORE_TENANT_EXPORT_POLICY` entry — `available_version` added to the `included` bucket (`services/tenantExportPolicyRegistry.ts:138`). No other cascade list changes: this is a new column on a long-registered table, not a new table.

**Guard test** (`routes/agents/patches.test.ts`): asserts a *lower* incoming severity renders the ranked `CASE` whose fallthrough keeps the stored column; that `'unknown'`/`null` produce the COALESCE-keep form; that `'critical'` binds as a parameter in the raising form; that `patches.version` receives only a `COALESCE` fill while the reported version lands on `device_patches.available_version`; and that a scan omitting `version` keeps both stored values. Assertions walk the produced `sql` fragment's `strings`/`values` rather than substring-matching a template.

---

## §1.7 — the SSRF guard was strong but never adopted on the backup path

`services/urlSafety.ts` `safeFetch` was independently verified as genuinely solid — DNS-resolved and **IP-pinned** (rebinding closed), Host header discarded, redirects not followed, metadata/link-local/CGNAT blocked. But `grep -rn "safeFetch\|urlSafety" apps/api/src/routes/backup/ apps/api/src/services/backup*` returned **zero matches**, and `routes/backup/configs.ts:176` handed a tenant-controlled S3 endpoint to the AWS SDK from a *test connection* route — the most directly reachable SSRF primitive in that area.

**The finding was incomplete adoption, not absent defense**, which shapes the fix.

`urlSafety.ts` was **extended, never weakened**:
- `assertSafeUrl(url)` — scheme + resolve + block check, for actionable user-facing errors
- `createGuardedLookup()` — a `net.LookupFunction` that resolves normally but only ever hands back non-blocked addresses. It runs at **connect time** and is the only resolution the socket sees, so there is no validate-then-connect window at all
- `createGuardedHttpAgents()` — `http`/`https` agents wired to that lookup, for SDKs that build their own HTTP client
- `safeFetch` was refactored onto the same internal `resolveSafeRecords()` helper so the three entry points cannot drift. Its observable behaviour, messages and error `details` are unchanged (its existing tests pass untouched).

The AWS SDK can't use `safeFetch` directly, so `services/guardedS3Client.ts` (`createGuardedS3Client`) is the single sanctioned chokepoint: it passes the guarded agents via `requestHandler`, and a caller may tune timeouts but can never swap the agents out. No new dependency — SDK v3 accepts a plain `NodeHttpHandlerOptions` object.

All three tenant-endpoint S3 clients adopted it:
- `routes/backup/configs.ts:195` (`probeS3Config`, + `assertSafeUrl` at `:186` for a clear error before any request)
- `services/recoveryDownloadService.ts:58`
- `services/recoveryMediaService.ts:254` (also used by `backupSnapshotStorage.ts` and `recoveryBootMediaService.ts`)

`services/s3Storage.ts` is intentionally left alone: it builds from the server's own `S3_ENDPOINT` env var, which is operator- not tenant-controlled, so it is outside this threat model.

**Guard tests.**
- `services/backupSsrfAdoption.test.ts` (the durable half) — scans the source text of every non-test file under `routes/backup/` plus `services/backup*.ts` and the recovery services, and fails on `new S3Client(`, bare `fetch(`, `axios`, `http(s).request(`, `new http(s).Agent(` or `undici`. The route/service sets are **discovered**, so a new backup file is covered the day it lands. It also asserts the three known sites positively import `createGuardedS3Client`, and self-checks that the scanned set is non-empty and contains the expected files so a broken glob can't make it vacuous. The allowlist is **per-file-per-pattern**, not whole-file — `recoveryMediaService.ts` waives only `fetch(` (server-configured GitHub release URL, manifest-verified; `safeFetch` can't be used because it deliberately doesn't follow redirects and release assets redirect) and still fails if it grows an unguarded `S3Client`.
- `routes/backup/configs.test.ts` — the probe rejects an endpoint resolving to `169.254.169.254`, `127.0.0.1` and `10.0.0.5` **before** any S3 request is attempted (`s3SendMock` asserted not called), and does *not* reject a publicly-resolving endpoint.
- `services/urlSafety.test.ts` — `createGuardedLookup` calls back with `SsrfBlockedError` when every record is private, and returns only the safe record from a mixed public/private resolution.

---

## Verification

- `pnpm --filter @breeze/api exec tsc --noEmit` — **clean**
- `vitest run src/routes/agents/ src/services/backupSsrfAdoption.test.ts` — **45 files, 786 tests passed**
- `vitest run src/services/urlSafety.test.ts src/routes/backup/configs.test.ts src/services/backupSsrfAdoption.test.ts` — **69 passed**
- `src/db/autoMigrate.test.ts` — 58 passed; `scripts/check-migration-naming.sh` OK
- **Non-vacuity check**: reverting both fixes produces **6 failures** across the three suites (severity expression, version fill, the three blocked-IP probe cases, and the `assertSafeUrl` adoption assertion). Restored afterwards.

Integration/RLS suites need a live DB and were not run locally; the export-policy contract runs in CI's **Integration Tests** job.

## Deliberately deferred

- **Full tenant-scoping of the `patches` catalog.** Kept out of scope as instructed. The right long-term answer is probably splitting the global identity row from per-tenant observed state; this PR moves the one field that was actually load-bearing for enforcement (`version`) and makes the other (`severity`) monotonic. Worth a follow-up issue.
- **`routes/devices/software.ts:79` and `services/vulnerabilityRemediation.ts:140`** still read `patches.version` directly. Both join `device_patches` and are device-scoped, so they would be marginally more accurate reading `COALESCE(available_version, version)` too — but neither is an enforcement point (display and CVE-to-patch matching), so they were left alone to keep the change surgical.
- **No backfill** of `device_patches.available_version`; it fills in naturally on the next scan and the `COALESCE` fallback covers the gap meanwhile.

## Note on overlapping work

`chore/harden-agent-patch-ingestion` (PR #3556) is **already merged into `main`** and is what defended the identity columns — this PR builds on it rather than conflicting with it. No open in-flight branch touches these files.


ToddHebebrand:
Pushed `bd90d8842` — closes a self-hosted regression this PR would otherwise have shipped.

**The regression:** the SSRF guard adopted here blocks RFC1918/ULA and the backup adoption passed no opt-out. A self-hosted install whose S3 endpoint is MinIO or a NAS on the LAN (`http://10.0.0.5:9000`, `http://minio.local:9000`) would have failed the test-connection probe and every recovery download. Hard breaking change for self-hosters.

**The fix:** thread the repo's existing escape hatch, `selfHostAllowsPrivateNetwork()` (`config/env.ts`), through the guard — same pattern as `workers/webhookDelivery.ts` and `routes/sso.ts`.

- `services/guardedS3Client.ts` resolves the flag **itself** rather than taking it from callers. All three call sites (test-connection route, `recoveryDownloadService`, `recoveryMediaService`) want the same answer, a per-caller flag is precisely how the original incomplete-adoption finding happened, and a future fourth call site now inherits the correct behaviour for free.
- `routes/backup/configs.ts` passes the same value to the pre-flight `assertSafeUrl`, so the check can never be stricter than the connection it validates.

**Hosted is unchanged.** The opt-out opens only on an *affirmative* `IS_HOSTED` self-host signal (#570 fail-closed lesson — unset/garbage/truthy all stay strict), and `isAlwaysBlockedIp` still blocks cloud metadata, link-local, loopback and CGNAT in **both** modes.

**Tests:** new `services/guardedS3Client.test.ts` asserts at the connect-time `lookup` on the agents handed to the SDK (not on a flag being passed around): RFC1918 allowed on self-host, blocked on hosted, blocked when `IS_HOSTED` is unset. `routes/backup/configs.test.ts` covers the same both-ways through the probe route. Both files assert `169.254.169.254` and link-local stay rejected in both modes — the assertion that proves the opt-in did not become a blanket bypass. Verified non-vacuous: with the change reverted, exactly the two self-host cases fail.

`backupSsrfAdoption.test.ts` intent unchanged and still green. API typecheck clean; the five affected/SSRF suites pass locally (84 tests).

ToddHebebrand:
## Functional regression in this PR: recovery-media downloads broke on GitHub's 302

Pushed `a070a0c83` to fix a regression this PR introduced, plus the stale waiver that let it through.

### The regression

`recoveryMediaService.downloadFile()` was converted from raw `fetch` to `safeFetch`. `safeFetch` deliberately follows **no** redirects (`urlSafety.ts`: *"Follow no redirects by default — caller gets the raw response and can re-invoke safeFetch if they want to trust the Location header"*). The URL it receives comes from `getGithubBackupUrl()` → a `github.com/.../releases/download/...` asset URL, and those **302 to `objects.githubusercontent.com`**. `response.ok` is false for a 302, so the call throws `download failed with status 302` and the backup binary never downloads — recovery-media builds are dead in GitHub binary-source mode.

This was called out in advance in this very PR: the `recoveryMediaService.ts` / `global fetch(` ALLOWLIST entry in `backupSsrfAdoption.test.ts` says verbatim that safeFetch can't be used here because it doesn't follow redirects and GitHub asset URLs redirect. The conversion landed anyway and the waiver stayed, so the guard test had nothing to contradict.

### The fix — keep the guard, follow the chain explicitly

New exported `safeFetchFollowingRedirects()` in `services/urlSafety.ts`:

- bounded loop, `SAFE_FETCH_MAX_REDIRECTS = 5` (caller-overridable);
- each hop is a **fresh `safeFetch`**, so every intermediate URL is independently DNS-resolved, filtered and IP-pinned. That is exactly the difference between this and `redirect: 'follow'`, which validates only the first URL and is a textbook SSRF bypass;
- relative Locations resolve against the current URL; missing/unparseable Location, non-http(s) scheme, and hop-cap exhaustion are all errors, never silent successes;
- non-redirect failures keep the original `download failed with status N` message;
- spec-shaped hop handling: 303 (and 301/302 on a non-GET) degrades to GET with no body, and credential headers are dropped when the hop crosses origin.

**`safeFetch`'s own default is unchanged** — no existing caller silently starts following redirects. The helper lives in `urlSafety.ts` rather than as a local closure so a second caller (there will be one — the release-manifest fetches still use raw `fetch(url, { redirect: 'follow' })`) inherits the re-validation instead of reinventing it.

The stale ALLOWLIST waiver is removed, so a future raw `fetch` in `recoveryMediaService.ts` **is** now caught.

### Tests

New `recoveryMediaService.redirect.test.ts` runs the **real** guard (only DNS and the socket layer are stubbed — unlike the sibling suite it does not mock `./urlSafety`):

- 302 to a different host downloads the asset; both hosts dialed in order.
- **Security-critical:** a redirect to `169.254.169.254`, to an RFC1918 literal, and to a hostname that only *DNS* reveals as `192.168.1.10` are each rejected with `SsrfBlockedError`, and the request recorder asserts the second hop was **never dialed**.
- Hop-cap exhaustion → `too many redirects`, exactly 6 requests.
- Missing Location → error. 404 → the original `download failed with status 404`.

Plus direct helper cases in `urlSafety.test.ts`: scheme rejection, caller-supplied budget, cross-origin credential stripping, 303-vs-307 method handling.

**Non-vacuity confirmed:** reverting `downloadFile` to bare `safeFetch` fails 7 of the 8 new cases with `Error: download failed with status 302` — i.e. the suite reproduces the exact reported symptom.

### Sweep of the rest of the PR

The only other conversions are `new S3Client(` → `createGuardedS3Client` in `backup/configs.ts`, `recoveryDownloadService.ts` and `recoveryMediaService.ts`. Those swap in SSRF-guarded HTTP agents only; the AWS SDK's own redirect handling (S3 region `PermanentRedirect`, virtual-host routing) is untouched, and any redirect the SDK follows still resolves through the guarded lookup. `assertSafeUrl` in `configs.ts` is a pre-flight on the operator-entered endpoint, not on a redirect target. **No other redirect-dependent call site is affected.**

### Verification

- `pnpm --filter @breeze/api exec tsc --noEmit` — clean.
- `recoveryMediaService` (both suites), `urlSafety`, `backupSsrfAdoption`, `guardedS3Client`, `backup/configs`, `recoveryDownloadService`: **8 files, 105 tests passed**.

PR #3650, opened 2026-08-17
fix(web): What's New splash never auto-shows on the upgrade that introduces an entry (#3646)

### LINKED ISSUE #3646: What's New splash never auto-shows on the upgrade that introduces an entry (existing users baselined away)
## Summary

The What's New splash does not appear after an upgrade for users who were **already using Breeze before the splash feature shipped**. It's working exactly as written — the first-run baseline branch also catches existing users on their first post-upgrade load, silently marks the current version as "seen", and shows nothing.

Net effect: the `0.105.0` entry is never auto-shown to any pre-existing user. The earliest anything can auto-appear is the *next* release after a user's first visit.

## Root cause

`apps/web/src/lib/whatsNewState.ts:30-31`

```ts
const floor = storage.getItem(LAST_SEEN_KEY);
if (!floor) return { entry: null, baselineToSet: webVersion };   // first-ever load: show nothing
```

Suppressing the splash for a genuinely **new** user is correct — you don't want to greet someone on day one with "here's what changed in a version you never ran."

But an **existing** user upgrading from a build that predates the feature *also* has no `breeze.whatsNew.lastSeenVersion` key. They hit the same branch, and the consumer (`WhatsNewSplash.tsx:26-28`) writes the baseline and renders nothing:

```ts
const decision = decideWhatsNew(window.localStorage);
if (decision.baselineToSet) {
  window.localStorage.setItem(LAST_SEEN_KEY, decision.baselineToSet);
}
```

The two populations are indistinguishable by "has no key", but they want opposite behaviour.

## Reproduce

1. Ensure no `breeze.whatsNew.lastSeenVersion` in `localStorage` (simulates any user who last visited before the feature shipped).
2. Load the dashboard on a build where `PUBLIC_APP_VERSION` is `0.105.0` (must be a versioned build — `WEB_VERSION === 'dev'` short-circuits at `whatsNewState.ts:28`).
3. **Expected:** the `0.105.0` entry is shown once.
   **Actual:** nothing renders; `lastSeenVersion` is silently set to `0.105.0`, so the entry can never auto-appear.

## Secondary issue — `localStorage` is per-browser

`LAST_SEEN_KEY` lives in `localStorage`, so the baseline is per-browser and per-origin. A tech on a second machine, a fresh Chrome profile, or an incognito window is treated as first-ever and baselined away there too — even if they already saw the entry elsewhere.

## Mitigating detail

Not *totally* unreachable: `Sidebar.tsx:808` exposes a manual "What's New" link that fires `REOPEN_EVENT` and uses `latestApplicableEntry()`, which ignores the dismissal floor. So the content can be opened deliberately — it just never surfaces on its own, which defeats the point of a splash.

## Suggested fix

Distinguish "new user" from "existing user who predates the feature". Options, roughly in order of preference:

1. **Baseline to the oldest entry below the running build**, not to `webVersion`. On first run, set the floor low enough that the current release's entry still qualifies. New users still see at most one entry, and upgraders see the release they just moved to.
2. **Persist last-seen server-side** on the user record instead of `localStorage`. Fixes the per-browser problem at the same time, at the cost of a schema field and an endpoint.
3. Seed the baseline at account creation, so "no key" unambiguously means "pre-existing user".

Option 1 is the smallest change and fixes the reported symptom; option 2 is the more correct long-term shape if the splash is going to carry anything important.

## Notes

- Found while scoping an in-product security advisory banner. Worth stating explicitly: **this mechanism is not reusable for security notices** regardless of this fix — `WHATS_NEW_ENTRIES` is a compile-time constant in the web bundle, so a build can never warn about a vulnerability disclosed after that build was cut. That work needs a runtime source and is tracked separately.
- No security impact. Purely a product/release-comms bug.

ToddHebebrand:
## Problem

`decideWhatsNew` treated a missing `breeze.whatsNew.lastSeenVersion` key as "brand-new user": it wrote the running version as the baseline and rendered nothing. An **existing** user upgrading from a build that predates the splash has no key either, so they hit the same branch — the entry that shipped *with* the upgrade was silently baselined away and could never auto-show. The earliest anything could auto-appear was the release *after* a user's first post-upgrade visit, which defeats the point of the splash.

The two populations are indistinguishable by "has no key" but wanted opposite behaviour.

## Fix

Option 1 from the issue (smallest change that fixes the symptom): with no stored floor, don't baseline — show the newest entry the running build shipped, once.

- No floor → every entry at or below `webVersion` is a candidate, and `highest` picks one. A backlog still produces a **single** splash, and entries newer than the running build are still suppressed.
- `baselineToSet` is now written only when there is no floor **and** nothing applicable to show, so the next release still surfaces.
- An empty stored value is treated as absent rather than as a version.
- New users see at most one entry (the release they signed up on) — accepted trade-off, as noted in the issue.

## Not in scope

The secondary `localStorage` per-browser point (option 2, server-side last-seen) is a schema field + endpoint and is deliberately left out. After this fix, a second browser/profile shows at most one entry rather than none.

## Tests

- `apps/web/src/lib/whatsNewState.test.ts` — no-floor shows newest entry; respects the running web version; single entry despite a backlog; no-floor-with-nothing-applicable writes the baseline; empty floor treated as absent; no re-show after acknowledgement.
- `apps/web/src/components/whatsNew/WhatsNewSplash.test.tsx` — splash renders with cleared `localStorage` and writes nothing until "Got it"; renders nothing once acknowledged.

Verified the new tests are not vacuous: with the `whatsNewState.ts` change stashed, 6 of them fail.

Closes #3646

### DIFF
--- a/apps/web/src/components/whatsNew/WhatsNewSplash.test.tsx
+++ b/apps/web/src/components/whatsNew/WhatsNewSplash.test.tsx
@@ -50,10 +50,20 @@ describe('WhatsNewSplash', () => {
     });
   });
 
-  it('renders nothing when there is no applicable entry', async () => {
-    window.localStorage.clear(); // first-ever load => baseline only, no show
+  // #3646: with no stored floor (existing user upgrading from a pre-feature
+  // build) the newest shipped entry must still auto-show rather than be
+  // silently baselined away.
+  it('shows the newest entry when no floor is stored', async () => {
+    window.localStorage.clear();
+    render(<WhatsNewSplash />);
+    expect(await screen.findByText(/whatsNew\.title:/)).toBeInTheDocument();
+    // Nothing is written until the user acknowledges it.
+    expect(window.localStorage.getItem(LAST_SEEN_KEY)).toBeNull();
+  });
+
+  it('renders nothing once the newest entry has been acknowledged', async () => {
+    window.localStorage.setItem(LAST_SEEN_KEY, '99.0.0');
     const { container } = render(<WhatsNewSplash />);
-    // Give the mount effect a tick to run (it will write the baseline and stay closed).
     await waitFor(() => {
       expect(container).toBeEmptyDOMElement();
     });
--- a/apps/web/src/components/whatsNew/WhatsNewSplash.test.tsx
+++ b/apps/web/src/components/whatsNew/WhatsNewSplash.test.tsx
@@ -1,6 +1,6 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { render, screen, fireEvent, waitFor } from '@testing-library/react';
-import WhatsNewSplash from './WhatsNewSplash';
+import WhatsNewSplash, { REOPEN_EVENT } from './WhatsNewSplash';
 import { LAST_SEEN_KEY } from '../../lib/whatsNewState';
 
 // t returns key + interpolated version so assertions are stable without real i18n.
@@ -68,5 +68,16 @@ describe('WhatsNewSplash', () => {
       expect(container).toBeEmptyDOMElement();
     });
     expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
+  });
+
+  // The sidebar link reopens via a window event and ignores the dismissal floor.
+  it('reopens on the sidebar event even when the floor suppresses the splash', async () => {
+    window.localStorage.setItem(LAST_SEEN_KEY, '99.0.0');
+    render(<WhatsNewSplash />);
+    await waitFor(() => {
+      expect(screen.queryByText('whatsNew.gotIt')).not.toBeInTheDocument();
+    });
+    fireEvent(window, new Event(REOPEN_EVENT));
+    expect(await screen.findByText(/whatsNew\.title:/)).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit — `code-reviewer` + `pr-test-analyzer`.

**Findings:** 0 blocking. `code-reviewer` found no high-confidence issues and confirmed the unchanged `WhatsNewSplash.tsx` consumer still satisfies the revised `baselineToSet` contract. `pr-test-analyzer` confirmed the new tests are non-vacuous and that the two removed tests asserted the buggy behaviour itself (legitimate update, not a coverage loss), and raised three adjacent coverage gaps. Two were worth acting on and are addressed in `2f09250ed`:

- **Corrupted floor = permanent silent suppression.** An unparseable `breeze.whatsNew.lastSeenVersion` made every `semverCompare` return `null`, so nothing was ever applicable — and because the floor was non-null, the baseline-rewrite branch never fired either. The splash was dead forever in that browser with no path back. Pre-existing, but the same silent-disappearance shape as #3646, so fixed here: a floor that doesn't parse is treated like a missing one.
- **Untested paths:** empty `WHATS_NEW_ENTRIES` list, and the sidebar `REOPEN_EVENT` path at the component level.

The third ("Show me later" repeat-display proven by remount) was judged redundant — the no-write assertion plus the `decideWhatsNew` unit tests already cover it.

**Tests:** `apps/web` affected suites green single-fork — `vitest run src/lib/whatsNewState.test.ts src/components/whatsNew/WhatsNewSplash.test.tsx`, 22 passed. `tsc --noEmit` clean; `astro check` 0 errors / 0 warnings. Non-vacuity verified: with `whatsNewState.ts` stashed, 6 of the new tests fail.

PR #3651, opened 2026-08-17
fix(web): gate dashboard "all clear" claims on a non-zero device count (#3613)

### LINKED ISSUE #3613: [Web] Dashboard cards still claim "All clear" / "No open vulnerabilities" when zero devices are reporting
#3536 fixed the false "fleet is healthy" claim on the Alerts page — an empty result set was being rendered as an affirmative all-clear even when zero devices were reporting. The fix distinguishes all three states correctly there (`apps/web/src/components/alerts/AlertsPage.tsx:546/566/574`) and is merged.

**The same false claim survives on two dashboard cards, and one of them is the first thing a user sees after login.**

`apps/web/src/components/dashboard/AlertsFeed.tsx:115-121` — `rows.length === 0` renders "All clear / No active alerts across your fleet" with no device gate at all. Zero devices reporting is indistinguishable from a genuinely quiet fleet.

`apps/web/src/components/dashboard/VulnerabilitiesCard.tsx:51,66-72` — `totalFindings === 0` renders "No open vulnerabilities". Zero devices, or devices with no scan coverage, reads identically to scanned-and-clean. This one is the more dangerous of the two: "no vulnerabilities found" and "we never looked" are opposite facts.

**The pattern to copy already exists in the same directory.** `SecurityPostureCard.tsx:65` gates on `totalDevices === 0` before making its claim. That's the shape both of these need — the aggregate is only meaningful once you know the denominator is non-zero.

Worth doing as one sweep rather than two one-liners: the "no data yet" versus "checked and clean" distinction applies to every rollup card, and grepping for the affirmative copy strings is a cheap way to find any I've missed here.

Related: #3566 (merged) does the adjacent version of this for the security stat tiles — showing an em dash instead of fabricated zeros when the request 500s. Between that and #3536 the remaining gap is the dashboard cards above.

ToddHebebrand:
## Problem

Two dashboard cards rendered an affirmative health claim straight out of an empty result set, with no device gate:

- `AlertsFeed.tsx` — `rows.length === 0` → **"All clear / No active alerts across your fleet"**
- `VulnerabilitiesCard.tsx` — `totalFindings === 0` → **"No open vulnerabilities"**

With zero devices enrolled — what a brand-new partner sees immediately after login — those read identically to a scanned-and-clean fleet. "We looked and found nothing" and "we never looked" are opposite facts. The vulnerabilities one is the more dangerous of the two.

#3536 fixed the same class of bug on the Alerts page; `SecurityPostureCard.tsx:65` already gates on `totalDevices === 0`. This applies that shape to the two cards that still lacked it.

## Change

**`fleetPresence.ts` (new)** — one shared classifier over the existing `/devices/stats` dashboard query, so every rollup card resolves the denominator the same way rather than each re-deriving it:

| state | meaning |
|---|---|
| `present` | ≥1 device enrolled — an aggregate of zero is a real finding |
| `none` | zero devices — the aggregate means nothing |
| `loading` | count not settled; render neutral rather than flash a claim that may be retracted |
| `unknown` | `/devices/stats` failed or is 403/404-hidden; coverage unprovable |

A cached count wins over an in-flight *or* failed background poll — a transient poll failure shouldn't downgrade a known-good denominator to "unknown".

**`AlertsFeed`** and **`VulnerabilitiesCard`** now take the devices query and only render the green all-clear on `present`. The other three states render honestly: an enrollment-oriented empty state on `none`, a neutral skeleton on `loading`, and — on `unknown` — the one fact we actually have ("no active alerts" / "no open findings reported") next to an explicit note that coverage can't be confirmed. Non-empty rendering paths are untouched.

**Sweep** (the issue asked for one rather than two one-liners): the other rollup cards were checked and are already correct — `SecurityPostureCard` gates on `totalDevices`, `PatchComplianceCard` on `hasAnyPatchData`, `FleetStatusCard` returns null at `total === 0`, and `KpiStrip` already swaps in an enroll-your-first-device CTA. `RecentActivity`'s "No activity yet" is not a health claim. No other affirmative copy needs gating.

## Tests

`apps/web/src/components/dashboard/zeroDeviceClaims.test.tsx` — 14 cases: the classifier's five inputs plus the cached-count precedence rule, and for **each** card all four presence states, asserting the all-clear copy is *absent* in three of them and still present in the fourth. Plus a case per card proving the populated path is unaffected.

- `vitest run src/components/dashboard/` → 19 passed (2 files)
- i18n contract suites (localeParity, keyUsage, translationCoverage, extractionQuality, terminologyQuality) → 81 passed
- `tsc --noEmit` → clean

New keys are translated (not English placeholders) across all eight locales.

Closes #3613

### DIFF
--- a/apps/web/src/components/dashboard/AlertsFeed.tsx
+++ b/apps/web/src/components/dashboard/AlertsFeed.tsx
@@ -1,10 +1,11 @@
-import { AlertTriangle, AlertCircle, Info, XCircle, CheckCircle2, ChevronRight } from 'lucide-react';
+import { AlertTriangle, AlertCircle, Info, XCircle, CheckCircle2, ChevronRight, HardDriveDownload } from 'lucide-react';
 import { cn } from '@/lib/utils';
 import { getErrorMessage, getErrorTitle } from '@/lib/errorMessages';
 import { formatTimeAgo } from '@/lib/formatTime';
 import { useTranslation } from 'react-i18next';
+import { fleetPresence } from './fleetPresence';
 import type { DashboardQueryState } from '../../hooks/useDashboardQuery';
-import type { AlertRow, AlertsSummary } from './types';
+import type { AlertRow, AlertsSummary, DeviceStats } from './types';
 
 const severityConfig: Record<string, { icon: typeof Info; chipClass: string; iconClass: string }> = {
   critical: { icon: XCircle, chipClass: 'bg-destructive/10', iconClass: 'text-destructive' },
@@ -19,15 +20,20 @@ const severityConfig: Record<string, { icon: typeof Info; chipClass: string; ico
  * filters server-side); rows with a device deep-link to it. Hidden when the
  * caller can't read alerts — an empty success state on a 403 would be a
  * false health claim.
+ *
+ * An empty feed is only "all clear" once the fleet denominator is known to be
+ * non-zero — see `fleetPresence` (#3613).
  */
 export default function AlertsFeed({
   alerts,
   summary,
+  devices,
   showOrg,
   onRetry,
 }: {
   alerts: DashboardQueryState<AlertRow[]>;
   summary: DashboardQueryState<AlertsSummary>;
+  devices: DashboardQueryState<DeviceStats>;
   showOrg: boolean;
   onRetry: () => void;
 }) {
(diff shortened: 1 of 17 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (plus a targeted code-reviewer re-run on the follow-up commit, since the fixes reached into the shared `useDashboardQuery` hook).

**Findings:** 4 raised → 4 addressed, 0 outstanding.

1. **Cross-hook staleness could transiently reproduce the bug being fixed** (code-reviewer). Unlike `SecurityPostureCard`/`PatchComplianceCard`, which gate on a denominator embedded in their own payload, `fleetPresence` reads the *separate* `/devices/stats` query — and `useDashboardQuery` deliberately keeps the previous scope's `data` visible during a refetch. Switching from a populated org to an empty one, with the alerts/vulns query resolving first, would paint the old org's count over the new org's zeros and re-assert the all-clear. Fixed in `cae8045c6`: the hook now tracks which org scope produced the value it holds and exposes `staleScope`; `fleetPresence` treats it as `loading`. A failed post-switch refetch keeps the flag set — a failure doesn't make old-scope data current. Widgets rendering their own numbers ignore the flag.
2. **`FleetStatusCard` vanished entirely on a `/devices/stats` failure** (silent-failure-hunter) — `!stats` was folded in with `total === 0`, so an outage rendered as an absent card, the same false negative in a different costume, and contradicting `KpiStrip`'s documented rule for the same query. Fixed in `cae8045c6`; the 403/404 permission-hide still hides deliberately. No new copy — both keys already exist in all eight locales.
3. **`fleetPresence` collapses a 403 permission-hide and a 500 into one `unknown`** (silent-failure-hunter, medium). **Not changed, deliberately.** The reviewer agreed the copy is honest in both cases; the missing piece is observability of a real `/devices/stats` outage, which belongs in `useDashboardQuery`'s error path (it only `console.error`s non-`Response` errors today) rather than in a presentational classifier. Worth a separate issue.
4. **The follow-up commit clobbered `useDashboardQuery.test.tsx`** instead of extending it, dropping six pre-existing cases — including the sequence-guard regression test that pins the very cross-org data flash the new `staleScope` logic leans on. Caught by the re-review, restored verbatim in `5e892145a` with the four new cases appended to its existing harness. Good catch by the reviewer; the suite is now 10 tests.

The test analyzer mutation-tested the original suite (stubbing `fleetPresence` to always return `'present'` → 9/14 failed) and confirmed no vacuous assertions.

**Tests:** `vitest run src/hooks/useDashboardQuery.test.tsx src/components/dashboard/ src/lib/i18n/` → **137 passed** (10 files), including all five i18n contract suites. `tsc --noEmit` clean. Rebased onto current `main` (which had moved, and had touched the same locale files) and re-verified. CI green: Test Web, Type Check, Lint, Test API, Integration Tests all pass.

**Status:** review-clean, awaiting maintainer merge. Not merged and issue #3613 left open per handoff rules.

Issue #3599, opened 2026-08-16
Inbound email: "Contact your administrator" is a dead end for self-hosters — name the variable


ToddHebebrand:
## Summary

When the platform inbound domain is unset, the Inbound email card says:

> The platform inbound domain isn't configured yet. Contact your administrator.

On hosted Breeze that's correct — the operator is us. On a self-hosted instance the reader **is** the administrator, so the message tells them to contact themselves and names nothing they can act on.

This is exactly what happened to the Discord reporter that prompted #3586: they read the message, searched `.env` for anything resembling an inbound setting, found nothing (the variable was undocumented at the time), and concluded the feature was Microsoft 365-only.

## Current state

- String: `apps/web/src/locales/en/settings.json:596` (`domainNotConfigured`), translated across all 7 locale bundles.
- Rendered: `apps/web/src/components/settings/InboundEmailCard.tsx:279-283`, `data-testid="inbound-address-unconfigured"`.

## Suggested fix

Name the variable and link the docs. Something like:

> No platform inbound domain is configured, so an inbound address can't be generated. Your Breeze operator needs to set `TICKETS_INBOUND_DOMAIN` — see [Inbound email-to-ticket](https://docs.breezermm.com/deploy/environment/#inbound-email-to-ticket-mailgun). Alternatively, connect a Microsoft 365 shared mailbox below — that path needs no inbound domain.

Points worth keeping in whatever wording lands:

- Name `TICKETS_INBOUND_DOMAIN` explicitly. A self-hoster can act on a variable name; they cannot act on "contact your administrator".
- Link to the environment docs section added in #3586.
- Mention the M365 alternative, since it sidesteps the requirement entirely and is the cheaper path for most self-hosters.

Two nuances if we want to be precise rather than just clearer:

- The message is also shown to partner users on **hosted**, where "contact your administrator" is the correct advice and a variable name is noise. Ideally the copy branches on `IS_HOSTED`, or stays generic while a self-host-only hint carries the variable name.
- Once #3598 lands, this message should not appear at all when an M365 mailbox is connected — so this change is about the genuinely-unconfigured case only.

Needs the 7 locale bundles updated together; `en` alone will leave the other 6 on the old wording.

## Related

- #3586 — documents the self-hosted inbound setup (found during that work)
- #3597 — inbound `enabled` toggle not enforced server-side
- #3598 — same message shown when only the M365 path is configured

PR #3652, opened 2026-08-17
fix(web,api): don't report inbound email as unconfigured on the M365-only path (#3598)

### LINKED ISSUE #3598: Inbound email card shows "platform inbound domain isn't configured" even when only the M365 mailbox path is in use
## Summary

`domainConfigured` is derived purely from `TICKETS_INBOUND_DOMAIN` and never considers whether a Microsoft 365 shared mailbox is connected. An operator who deliberately runs **only** the M365 path — which needs no inbound domain, no MX records, and no Mailgun account — still sees an error on the Inbound email card telling them something is unconfigured, when in fact their setup is complete and working.

## Evidence

`apps/api/src/services/ticketConfigService.ts:374-377`:

```ts
const domain = getConfig().TICKETS_INBOUND_DOMAIN ?? '';
const domainConfigured = domain.length > 0;
const derived = domainConfigured && effectiveLocalPart ? `${effectiveLocalPart}@${domain}` : '';
```

The web card branches on that single boolean (`apps/web/src/components/settings/InboundEmailCard.tsx:251`) and falls through to `settings.json:596` — *"The platform inbound domain isn't configured yet. Contact your administrator."* — with `data-testid="inbound-address-unconfigured"`.

Nothing in that computation looks at connected mailboxes, so the message renders regardless of M365 state.

## Impact

Cosmetic, but actively misleading in the one place a self-hoster looks for confirmation their setup worked. It's how the original Discord reporter concluded email-to-ticket was M365-only and still broken: the docs pointed them at M365, and the card kept reporting an error afterward.

## Suggested fix

The card is really answering two questions that have been collapsed into one boolean — "is there a native inbound address?" and "can this partner receive email as tickets at all?".

Options, roughly in order of preference:

1. **Scope the message to the native-address section.** Keep `domainConfigured` as-is (it accurately describes the Mailgun path), but only surface the error where the native address would appear, and suppress it when at least one M365 mailbox is connected — showing the connected mailbox address as the effective inbound address instead.
2. **Add a separate `inboundReady` signal** to the ticket-config payload (`domainConfigured || connectedMailboxCount > 0`) and drive the card's top-level state from that.

Either way the copy should stop implying the whole feature is broken when one of two valid paths is fully configured.

Related: #3599 covers the wording of the message itself for the case where it *is* genuinely unconfigured.

## Related

- #3586 — documents the self-hosted inbound setup (found during that work)
- #3597 — inbound `enabled` toggle not enforced server-side

ToddHebebrand:
## Problem

`domainConfigured` is derived purely from `TICKETS_INBOUND_DOMAIN`, so the Inbound email card told every partner running **only** the Microsoft 365 shared-mailbox path that "the platform inbound domain isn't configured" — on a setup that is complete and actively ingesting mail. That path needs no inbound domain, no MX records and no Mailgun account. It's how the original Discord reporter concluded email-to-ticket was M365-only and still broken.

## Fix

The card was answering two questions with one boolean. Split them:

- **API** — `getTicketConfig` now also returns `connectedMailboxCount`, backed by a new partner-scoped `countConnectedMailboxes()` in `ticketMailbox/connectionService.ts`. It counts only `status = 'connected'` rows: a `pending_consent`/`error`/`reauth_required`/`disabled` mailbox is *not* a working inbound path and must not read as one. `domainConfigured` keeps its exact meaning (is there a native Mailgun address) — no behaviour change for the Mailgun path.
- **Web** — the "no native address" slot now renders a neutral note pointing at the connected-mailbox card directly below when `connectedMailboxCount > 0`, and keeps the existing amber error (both the hosted and the self-hosted #3599 variants) only when neither path is configured. Both paths can run at once: when the domain *is* configured the native address editor renders unchanged regardless of mailboxes.

Back-compat: the field is optional on the web side, so an older API that omits it reads as 0 and the card behaves exactly as before.

New i18n key `settings:inboundEmail.addressViaMailbox` added to all 8 locales.

## Tests

- `apps/api/src/services/ticketMailbox/connectionService.test.ts` — count filters on partner **and** `status='connected'` (asserts the actual where-predicate, not just the row count), and 0 when none.
- `apps/api/src/services/ticketConfigService.test.ts` — `connectedMailboxCount` surfaces in the inbound block alongside `domainConfigured=false`, and is 0 when no mailbox is connected.
- `apps/web/src/components/settings/InboundEmailCard.test.tsx` — error suppressed with a connected mailbox (hosted + self-hosted), error still shown with none, native editor untouched when the domain is configured.

Green: `vitest run` on the three API suites (131 passed) and the card suite (18 passed); `src/lib/i18n` locale parity/coverage suites 105 passed; `tsc --noEmit` clean in `apps/api` and `apps/web`.

## Not in scope

The wording of the genuinely-unconfigured message is #3599 (already merged); this PR only decides *when* that message is shown.

Closes #3598

### DIFF
--- a/apps/api/src/services/ticketConfigService.test.ts
+++ b/apps/api/src/services/ticketConfigService.test.ts
@@ -153,6 +153,12 @@ vi.mock('../db/schema/portal', () => ({
 const { configRef } = vi.hoisted(() => ({ configRef: { current: { TICKETS_INBOUND_DOMAIN: 'tickets.example.com' as string | undefined } } }));
 vi.mock('../config/validate', () => ({ getConfig: () => configRef.current }));
 
+// The M365 mailbox count is a separate partner-scoped query (#3598); mock it so
+// the inbound tests stay on ticketConfigService's own FIFO of select results.
+// countConnectedMailboxes' own filtering is covered in connectionService.test.ts.
+const { countConnectedMailboxesMock } = vi.hoisted(() => ({ countConnectedMailboxesMock: vi.fn(async () => 0) }));
+vi.mock('./ticketMailbox/connectionService', () => ({ countConnectedMailboxes: countConnectedMailboxesMock }));
+
 // createTicket lives in ./ticketService — mock it so convert doesn't hit the
 // real ticket-creation service (DB inserts, event emission, SLA resolution).
 const { createTicketMock } = vi.hoisted(() => ({ createTicketMock: vi.fn() }));
@@ -182,6 +188,8 @@ beforeEach(() => {
   dbMocks.insertErrors.length = 0;
   dbMocks.insertResult = [];
   dbMocks.updateResult = [];
+  countConnectedMailboxesMock.mockClear();
+  countConnectedMailboxesMock.mockResolvedValue(0);
 });
 
 // Recursively flatten a predicate sentinel (eq/inArray/and) into its leaf clauses.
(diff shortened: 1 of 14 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer + pr-test-analyzer (both on the full `origin/main...HEAD` diff).

**Findings:** 3 raised, 0 blocking.
- *Inaccurate comment* (both reviewers): the `countConnectedMailboxes` docblock framed an org-scoped read as a designed degrade path, but the only caller sits behind `requireScope('partner','system')`, so it's unreachable. Fixed in `a378c01b7` — the comment now states the real contract (partner/system context required; an org-scoped caller would understate the setup).
- *`getTicketConfig` has no try/catch around the new query* — deliberately not changed. Every other query in `getTicketConfig` fails the request the same way, and swallowing this one to `0` would reintroduce exactly the false "not configured" signal this PR removes. Failing loudly is the correct behaviour here.
- *No test asserting DB-side status filtering* — the mock returns queued rows regardless of the predicate, so the meaningful assertion is the where-clause shape, which is what the test checks (`and(eq(partnerId), eq(status,'connected'))`, sentinel-compared — it fails if the status filter is dropped). Real filtering is Postgres's job; noted rather than papered over.

Reviewers also confirmed `ticket_mailbox_connections` is already in `PARTNER_TENANT_TABLES` with a `breeze_has_partner_access` policy, and that no cascade/export-policy registration applies (no new table or column).

**Tests:** `vitest run` green locally — `ticketConfigService.test.ts` + `connectionService.test.ts` + `ticketConfig.test.ts` (131 passed), `InboundEmailCard.test.tsx` (18 passed), `src/lib/i18n` parity/coverage (105 passed), `no-silent-mutations` (105 passed). `tsc --noEmit` clean in `apps/api` and `apps/web`. Full CI on `a378c01b7`: 54 success, 1 skipped, 0 failing — including Test API, Test Web and all four Integration Tests shards.

PR #3655, opened 2026-08-17
fix(api): a result arriving after the 60s wait deadline no longer loses the agent's real output (#3607)

### LINKED ISSUE #3607: run_script: a result arriving after the 60s wait deadline permanently loses the agent's real output

ToddHebebrand:
Closes #3607

## The defect

A `run_script` (or any dispatched command) slower than the 60s wait deadline **permanently lost the agent's real output**. The run was reported to the operator and to the AI as `failed / exitCode:null / stdout:null` even when the script had succeeded on the device.

The failure is a **lookup**, not a race, and that distinction is the whole fix:

1. `waitForCommandResult(commandId, 60000)` gives up and terminalizes the `device_commands` row to `status:'failed'`, `result:{status:'timeout'}`.
2. The agent's real result then arrives. The ingest **lookup** filtered on `status IN ('pending','sent')`, so it matched **no row at all** — the handler never reached its compare-and-set. It is tempting to blame the CAS; the code branches away long before it.
3. With no command row, the handler fell into `processOrphanedCommandResult`, which handles only SNMP / discovery / tunnel and returns for everything else.
4. So `handleScriptResult` — the only writer of `stdout`/`stderr`/`exitCode` onto `script_executions` — never ran. The output was not late or degraded. It was never written, and there was nothing to re-read.

Long-running work (installs, updates, disk operations) is exactly what exceeds 60s, so this biased toward losing the results that matter most.

## The fix

A server-side timeout is now treated as **provisional**: the row stays terminal for every reader, but a genuine agent result may still overwrite it.

### 1. Command acceptance

**`services/commandResultAcceptance.ts` (new)** — one shared definition of "this row may still accept an agent result", replacing the `ACCEPTED_COMMAND_RESULT_STATUSES` constant that was duplicated on the WS path and its HTTP twin:

```
status IN ('pending','sent')  OR  (status = 'failed' AND result->>'status' = 'timeout')
```

Applied at **all four sites** the issue calls out — both WS lookups, the WS CAS, the REST pre-read short-circuit, and the REST CAS. Applying it only at a CAS would do nothing.

Why the discriminator is safe:

- `buildStoredCommandResult` stores the **agent's** status verbatim, and that is only ever `completed`/`failed`. An agent-reported failure is therefore never reopened — only the server's own `timeout` marker is. All three server-side timeout writers stamp it: the wait deadline, `staleCommandReaper`, and `markVerificationCommandTimedOut`.
- **Double-delivery is still rejected.** The predicate is re-evaluated inside the CAS, and the first late result rewrites `result.status` away from `'timeout'` — so a duplicate frame finds 0 rows and is ignored exactly as before.
- A cancellation stores `status:'cancelled'`, so a cancelled command is not resurrected. `result IS NULL` on a `failed` row also does not match (`NULL->>'status'` is NULL, falsy).

### 2. Execution recovery

`handleScriptResult` gained a second UPDATE for an execution row a server-side sweep already stamped terminal.

The predicate is **not** `status = 'timeout'`, and this is the subtle part: `reapStaleScriptExecutions` derives the execution status from the *command* row, and in exactly this scenario that row is `failed` with `result.status = 'timeout'` — so the reaper stamps the execution **`'failed'`**, not `'timeout'`. Keying on `'timeout'` would have left the dominant post-reaper path still losing output. The discriminator is instead "this execution never received the agent's output": `exit_code IS NULL AND stdout IS NULL`. A row already carrying real output is never overwritten.

This branch deliberately does **not** touch the batch counters — every writer of a terminal execution status already incremented one for this device, so bumping again would push `devicesCompleted + devicesFailed` past `devicesTargeted`.

If both updates match nothing, that is now logged with the row's actual current state and captured to Sentry. Widening upstream acceptance made that branch reachable, and it was the one remaining way output could vanish here silently — the #3162 lesson applied to the code that cites it.

### 3. Restore-job parity

`propagateTimedOutDeviceCommand` fails `restore_jobs` in lockstep with the command timeout, and `isMutableRestoreStatus` admitted only `pending`/`running`. Left alone, a recovered restore would read `completed` on `device_commands` and `failed` on `restore_jobs` — and the UI reads the latter. The same provisional carve-out now applies, keyed on the reaper's own `timedOutBy: 'server'` stamp, at both the guard and the compare-and-set. An agent-reported restore failure carries no such stamp and stays final. (Backup jobs and backup verification already had equivalent fallbacks.)

### 4. AI tool descriptions

#3605 told the model *not* to re-check a timed-out `run_script`, correctly, because that recovery did not exist. It exists now — `run_script` already returns the `executionId` — so both `get_script_execution` descriptions carve out exactly the `status: "timeout"` case and nothing else. **This is the one judgement call in the PR** rather than a mechanical consequence; flagging it explicitly in case you'd rather split it out.

## Tests

New `apps/api/src/__tests__/integration/lateCommandResultRecovery.integration.test.ts` — real Postgres, real handlers, assertions on what is actually in the tables. A mocked-db unit test cannot pin this: it would assert a where-clause shape and stay green with the fix deleted.

Ten cases: WS late success lands stdout/exitCode; WS late non-zero exit records a real failure (not a timeout); WS duplicate frame still ignored; WS cancelled command not reopened; HTTP fallback recovers identically; execution stamped `timeout` recovers; execution stamped `failed` by the reaper (the status it actually writes) recovers; an execution already carrying real output is not overwritten; a server-timed-out restore job accepts the real result; an agent-reported restore failure does not.

**Every behavioural change is mutation-verified.** Reverting the acceptance predicate turns 5 red; reverting only the execution-recovery branch turns exactly the reaper cases red; reverting only the restore carve-out turns exactly the restore case red. The negative cases stay green throughout, which is what makes them meaningful.

- Integration: 10/10 passed
- Unit (single-fork): 30 files, 528 passed / 2 skipped
- `tsc --noEmit` clean

No schema change, no migration, no tenancy surface touched.

### DIFF
--- a/apps/api/src/__tests__/integration/lateCommandResultRecovery.integration.test.ts
+++ b/apps/api/src/__tests__/integration/lateCommandResultRecovery.integration.test.ts
@@ -1,0 +1,396 @@
+/**
+ * Integration test — #3607: a result arriving after the 60s wait deadline must
+ * NOT lose the agent's real output.
+ *
+ * The defect this pins is a lookup, not a race:
+ *
+ *   1. `waitForCommandResult(commandId, 60000)` gives up and terminalizes the
+ *      `device_commands` row to `status:'failed'`, `result:{status:'timeout'}`.
+ *   2. The agent's REAL result lands over the WS (or the HTTP fallback). The
+ *      ingest lookup filtered on `status IN ('pending','sent')`, so it matched
+ *      NO ROW — the handler never reached its compare-and-set, it branched into
+ *      `processOrphanedCommandResult`, which handles only SNMP/discovery/tunnel.
+ *   3. `handleScriptResult` — the only writer of stdout/stderr/exitCode onto
+ *      `script_executions` — therefore never ran, and nothing ever re-read it.
+ *
+ * A mocked-db unit test cannot pin this: it would assert on a where-clause
+ * shape and stay green with the fix deleted. These cases run the real handlers
+ * against real Postgres and assert on what is actually IN the two tables.
+ *
+ * Covered:
+ *   1. WS path — timed-out command + late success → stdout/exitCode land.
+ *   2. WS path — a second copy of the same frame is still ignored (the CAS
+ *      protection the widened predicate must not break).
+ *   3. WS path — a genuinely cancelled command is NOT reopened.
+ *   4. HTTP fallback path — same recovery (the duplicated constant in
+ *      routes/agents/commands.ts).
+ *   5. A `script_executions` row already stamped `timeout` by the stale reaper
+ *      still receives the real output.
+ */
+import { describe, it, expect } from 'vitest';
+import { eq } from 'drizzle-orm';
+import { Hono } from 'hono';
+
+import './setup';
+import { getTestDb } from './setup';
+import { setupTestEnvironment } from './db-utils';
+import { withDbAccessContext } from '../../db';
+import { createAgentWsHandlers } from '../../routes/agentWs';
+import { commandsRoutes } from '../../routes/agents/commands';
+import { devices, deviceCommands, scripts, scriptExecutions } from '../../db/schema';
+
+const runDb = it.runIf(!!process.env.DATABASE_URL);
+
+/** Minimal WSContext stand-in — the command-result path only ever calls send(). */
+const fakeWs = { send: () => {} } as unknown as Parameters<
+  ReturnType<typeof createAgentWsHandlers>['onMessage']
+>[1];
+
+interface Fixture {
+  orgId: string;
+  siteId: string;
+  deviceId: string;
+  agentId: string;
+  userId: string;
+  scriptId: string;
+}
+
+async function makeFixture(): Promise<Fixture> {
+  const env = await setupTestEnvironment();
+  const tdb = getTestDb();
+  const agentId = `agent-3607-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+
+  const [device] = await tdb
+    .insert(devices)
+    .values({
+      orgId: env.organization.id,
+      siteId: env.site.id,
+      agentId,
+      hostname: `late-result-${agentId}`,
+      osType: 'windows',
+      osVersion: '11',
+      architecture: 'x86_64',
+      agentVersion: '0.0.0-test',
+      status: 'online',
+      enrolledAt: new Date(),
+    })
+    .returning({ id: devices.id });
+  if (!device) throw new Error('makeFixture: no device');
+
+  const [script] = await tdb
+    .insert(scripts)
+    .values({
+      orgId: env.organization.id,
+      name: `late-result-script-${agentId}`,
+      osTypes: ['windows'],
+      language: 'powershell',
+      content: 'Start-Sleep -Seconds 120; Write-Output "slow but fine"',
+    })
+    .returning({ id: scripts.id });
+  if (!script) throw new Error('makeFixture: no script');
+
+  return {
+    orgId: env.organization.id,
+    siteId: env.site.id,
+    deviceId: device.id,
+    agentId,
+    userId: env.user.id,
+    scriptId: script.id,
+  };
+}
+
+/**
+ * Seed the exact post-deadline state: a `script_executions` row still waiting
+ * for output, and a `device_commands` row already terminalized the way
+ * `waitForCommandResult` terminalizes it at 60s.
+ */
+async function seedTimedOutRun(
+  fx: Fixture,
+  opts: {
+    executionStatus?: 'running' | 'timeout';
+    commandStatus?: 'failed' | 'cancelled';
+    commandResult?: Record<string, unknown>;
+  } = {},
+): Promise<{ commandId: string; executionId: string }> {
+  const tdb = getTestDb();
+
+  const [execution] = await tdb
+    .insert(scriptExecutions)
+    .values({
+      scriptId: fx.scriptId,
+      deviceId: fx.deviceId,
+      orgId: fx.orgId,
+      triggeredBy: fx.userId,
+      triggerType: 'manual',
+      status: opts.executionStatus ?? 'running',
+      startedAt: new Date(Date.now() - 90_000),
+    })
+    .returning({ id: scriptExecutions.id });
+  if (!execution) throw new Error('seedTimedOutRun: no execution');
+
+  const [command] = await tdb
+    .insert(deviceCommands)
+    .values({
+      deviceId: fx.deviceId,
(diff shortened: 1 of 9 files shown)


ToddHebebrand:
**Review run:** two rounds of `/pr-review-toolkit:review-pr` (`code-reviewer` + `silent-failure-hunter`; round 2 was a focused re-review of the fix commit).

**Findings:** 4 raised → 4 addressed; 0 outstanding.

- **Round 1 — `silent-failure-hunter`:** widening upstream acceptance made a new 0-rows-matched branch in `handleScriptResult` reachable, and it was silent. Chasing it down found the bigger one: the recovery keyed on `status = 'timeout'`, but `reapStaleScriptExecutions` derives the execution status from the *command* row, which in exactly this scenario is `failed` + `result.status = 'timeout'` — so the reaper stamps the execution **`'failed'`**. The dominant post-reaper path was still losing output. Re-keyed on `exit_code IS NULL AND stdout IS NULL`, plus the missing log/Sentry. `036a3b46d`
- **Round 1 — `code-reviewer`:** `propagateTimedOutDeviceCommand` fails `restore_jobs` in lockstep with the command timeout, and `isMutableRestoreStatus` admitted only pending/running — so a recovered restore would have read `completed` on `device_commands` and `failed` on `restore_jobs`, and the UI reads the latter. Same provisional carve-out applied, keyed on the reaper's own `timedOutBy: 'server'` stamp, at both the guard and the CAS. `036a3b46d`
- **Round 2 — `code-reviewer`:** the new Sentry capture would fire on a *benign* race — `routes/scripts.ts`'s cancel handler stamps the execution `cancelled` but only cancels the command `WHERE status = 'pending'`, so a `sent` command survives and its real result later arrives against a cancelled execution. Dropping the output is correct there; it is not a defect and must not page. Log retained, capture skipped, behaviour pinned by a test. `57974d65a`
- Round 2 also confirmed clean: the raw `sql` jsonb predicate (operator precedence, NULL handling, identifier interpolation), and that a restore job cannot be recovered twice (`buildRestoreResultMetadata` never carries `timedOutBy` forward).

**Tests:** `lateCommandResultRecovery.integration.test.ts` — 11/11 green against real Postgres. Unit, single-fork: 30 files, 528 passed / 2 skipped. `tsc --noEmit` clean.

**Every behavioural change is mutation-verified** — reverting the acceptance predicate turns 5 red; reverting only the execution-recovery branch turns exactly the reaper cases red; reverting only the restore carve-out turns exactly the restore case red. The negative cases (duplicate frame, cancelled command, execution already carrying output) stay green throughout, which is what makes them worth having.

One thing worth your call rather than mine: the `get_script_execution` tool-description change. #3605 deliberately told the model *not* to re-check a timed-out `run_script` because that recovery did not exist; it does now, so the descriptions carve out exactly the `status: "timeout"` case. That is the one judgement call in the PR and is trivially separable if you'd rather land the data-path fix alone.

PR #3657, opened 2026-08-17
fix(api): make fleet_remediation_runs roll-up counts fresh at chunk end (#3637)

### LINKED ISSUE #3637: [API] fleet_remediation_runs roll-up counts go stale between dispatcher writes and pollRunProgress reconciliation
**Found while fixing #3452** (run-counter lag, PR #3633). The display symptom is fixed client-side, but the underlying data issue remains.

**Root cause:** the roll-up columns on `fleet_remediation_runs` (`succeededCount` / `failedCount` / `skippedCount`) are written by two paths that don't agree. The dispatcher's `markTargetSkipped` / `markTargetFailed` (`apps/api/src/services/fleetFindings/dispatch.ts:348`) update the target row only; only `pollRunProgress` recomputes the run-level roll-up later. Between a dispatch chunk completing and the next poll, the roll-up reads stale (typically 0s).

**Blast radius:** PR #3633 makes `RunProgressPanel` count target rows from the same snapshot, so the web UI is correct. Every OTHER consumer of the roll-up columns — API responses, AI tools, reports — still reads the stale values in that window.

**Why it wasn't fixed server-side in #3633:** recomputing the aggregate inside `dispatchRunChunk` would put a run-wide write in the path of concurrently-racing chunks. A proper fix needs either an atomic increment per target transition or moving the roll-up to a computed read.

Refs #3452, PR #3633.

ToddHebebrand:
Closes #3637.

## The bug

`fleet_remediation_runs`' roll-up columns (`succeeded_count` / `failed_count` /
`skipped_count`) were written by two paths that didn't agree.
`dispatchRunChunk`'s `markTargetSkipped` / `markTargetFailed` updated the
per-target row and nothing else; only `pollRunProgress`, on its 30s BullMQ
scheduler, ever recomputed the run-level roll-up.

Between a dispatch chunk finishing and the next poll tick, the roll-up read
stale — typically all zeros against a run that had already resolved. #3633
patched `RunProgressPanel` to count target rows from its own snapshot, so the
web UI was correct, but every other consumer still read the columns:

- `GET /fleet-findings/runs/:runId` (run detail, via `getRemediationRun`)
- `GET /fleet-findings/:id/runs` (run list, via `getFleetFinding`)
- AI tools and reports reading the same fields

A second, quieter defect sat underneath it: `pollRunProgress` computed its
counts from an in-memory target snapshot read at the *top* of the function and
then mutated. A dispatch chunk running concurrently was invisible to that
snapshot, so the poll could write a *lower* roll-up back over a fresher one.

## The fix

One shared writer, `recalculateRunRollup(runId, now?)`, is now the only thing
that writes those columns or the derived `status` / `completed_at`. It performs
an **absolute** recompute from the target rows in a single `UPDATE ... FROM
(SELECT COUNT(*) FILTER (...))`, and it is called from `dispatchRunChunk` (both
exits), `pollRunProgress`, and `markRunDispatchFailed`. Three duplicated
count-derivation blocks collapse into it.

**Why absolute, not the per-target atomic increment the issue floated.**
`markTargetFailed` can legitimately fire twice for one target under a BullMQ
chunk retry, so a relative `count = count + 1` double-counts, and any drift it
introduces is permanent. An absolute recompute is idempotent by construction,
self-correcting, and repairs rows historical drift already corrupted. It is
also *cheaper* than the design the issue rejected: the recompute runs **once
per chunk**, after the per-target loop — one run-wide write per chunk, not one
per target — so it never lands in the path of the per-target claim race.

**Why `SELECT ... FOR UPDATE` first.** Under READ COMMITTED the `UPDATE`'s
subquery reads a snapshot taken at statement start. Without the row lock two
recomputes can interleave and the one that started earlier overwrites the newer
result with older counts — worst case reopening a run the poll had already
terminalized and unscheduled, stranding it in `running` forever. Taking the
lock first forces the aggregate's snapshot to be taken after any competing
recompute has committed.

In-flight is counted directly from `status IN ('pending','queued')` rather than
derived as `target_count - succeeded - failed - skipped`. Deriving it from the
denormalized `target_count` would defeat a self-correcting recompute: too high
pins the run in `running` forever, too low terminalizes it while targets are
still live.

`status <> 'cancelled'` guards the write, because cancellation is an operator
decision about the run, not a fact about its targets — a late chunk or poll
must never resurrect a cancelled run.

## Also fixed in the same write path

These are all concurrency defects on the data the roll-up is derived *from*, so
fixing the roll-up without them would just make it faithfully report corrupted
targets:

- **`markTargetFailed` had no status guard.** A chunk retry could rewrite an
  already-`succeeded` target as `failed`. Now guarded on
  `status IN ('pending','queued')`, matching `markTargetSkipped`'s existing
  claim discipline.
- **The `queued -> running` update at the top of `dispatchRunChunk` matched on
  id alone.** Concurrent chunks each read `queued` in their own SELECT; a chunk
  that stalled between its read and its write could land `running` on top of a
  status the poll had since moved to terminal, un-finishing a completed run.
  Now a CAS on `status = 'queued'`.
- **`pollRunProgress`'s two flush statements classified from a stale
  snapshot** and wrote without a status predicate, so a concurrent chunk's real
  outcome could be overwritten with the poll's older verdict. Both now carry
  `status IN ('pending','queued')`.

## One deliberate behaviour change

When `markRunDispatchFailed` runs after some targets had *already succeeded*,
the run is now `partial` rather than a blanket `failed`. `partial` is the
truthful reading of those target rows and is what a run reaching the same
target state by any other path already reports. Its other two outcomes are
unchanged: targets still in flight hold the run at `running` with no
`completed_at` (the poll still owns convergence), and a run where nothing was
dispatched still lands on `failed`.

## Testing

- `apps/api/src/__tests__/integration/fleetFindings.integration.test.ts` —
  where the count/status derivation is now genuinely proven, against real
  Postgres, since it moved into SQL:
  - `(c)` strengthened: after `dispatchRunChunk`, with the target still in
    flight, the run has already left `queued` for `running` with no
    `completed_at` and no counts credited to an unresolved target.
  - `(c2)` **the #3637 regression** — two offline devices, so the whole chunk
    resolves inside dispatch as `skipped`/`unreachable` with no
    `device_commands`. `pollRunProgress` is deliberately never called: the run
    must read `skippedCount: 2`, `status: 'failed'`, `completedAt` set the
    moment the chunk ends. This fails on `main`.
  - `(c3)` replaying an identical chunk (as a BullMQ retry would) leaves the
    counts and `completed_at` untouched, and a late poll against a `cancelled`
    run leaves it `cancelled`.
- `apps/api/src/services/fleetFindings/dispatch.test.ts` — unit assertions
  moved onto the recompute: `FOR UPDATE` precedes the aggregate, the write runs
  inside a transaction, the `cancelled` guard and the `pending`/`queued`
  in-flight predicate are present, `target_count` is not referenced, `runId` is
  bound rather than interpolated, plus the new `markTargetFailed` and
  `queued -> running` guards.
- `tsc --noEmit` clean for `apps/api`.

No migration: this is a write-path change only, and the columns' meaning is
unchanged.

### DIFF
--- a/apps/api/src/__tests__/integration/fleetFindings.integration.test.ts
+++ b/apps/api/src/__tests__/integration/fleetFindings.integration.test.ts
@@ -525,6 +525,23 @@ describe('fleet findings — end-to-end (Task 11)', () => {
     expect(targetAfterDispatch?.status).toBe('queued');
     expect(targetAfterDispatch?.deviceCommandId).toBe(command.id);
 
+    // The chunk's own roll-up recompute (#3637) already ran — the run left
+    // `queued` without waiting for a poll tick. With the target still in
+    // flight the recompute must hold the run OPEN: `running`, no completedAt,
+    // and no count credited to a target that has not resolved. A recompute
+    // that derived in-flight from anything other than the live target rows
+    // would terminalize here and the poll scheduler would unschedule itself
+    // before ever reconciling this command.
+    const [runAfterDispatch] = await getTestDb()
+      .select()
+      .from(fleetRemediationRuns)
+      .where(eq(fleetRemediationRuns.id, runId));
+    expect(runAfterDispatch?.status).toBe('running');
+    expect(runAfterDispatch?.completedAt).toBeNull();
+    expect(runAfterDispatch?.succeededCount).toBe(0);
+    expect(runAfterDispatch?.failedCount).toBe(0);
+    expect(runAfterDispatch?.skippedCount).toBe(0);
+
     // Simulate the agent completing the command.
     await getTestDb()
       .update(deviceCommands)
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (correctness, RLS/tenancy, lock ordering) + `silent-failure-hunter` (suppressed errors, 0-row writes). An independent Codex read-only design review ran *before* implementation and shaped the final shape (absolute recompute over per-transition increments; `FOR UPDATE` before the aggregate; in-flight counted directly rather than derived from `target_count`).

**Findings:** 0 outstanding.
- `code-reviewer`: no consequential findings. Verified the status `CASE` ladder is a faithful, exhaustive translation of the JS it replaced (the five target statuses exactly partition the table's CHECK enum, so nothing falls through uncounted); `completed_at`'s `COALESCE` preserves stamp-once semantics; `runId` is parameter-bound, not interpolated; and — the one I most wanted a second pair of eyes on — `db.transaction` resolves through the ambient-context proxy to a **savepoint on the already-open context-scoped transaction**, so the raw `tx.execute` SQL still runs under the session's RLS GUCs. Not a new bypass. Lock ordering is consistent across both paths (target rows first, run row last), so no deadlock cycle.
- `silent-failure-hunter`: no consequential findings. The four new CAS-style predicates each have 0-rows as a documented, benign race outcome, so logging there would be noise. A `recalculateRunRollup` throw at the end of `dispatchRunChunk` fails the job loudly (BullMQ `on('failed')`), and the retry re-runs with `pendingTargets` empty — no double dispatch to devices.

**Not addressed (deliberate, flagged for your call):** `markRunDispatchFailed` has no run-not-found pre-check, unlike its two siblings which `console.warn` and return. The reviewer raised it as a consistency nit, not a defect — the window is a run being org-erased in the few hundred ms between `createRemediationRun` committing and the route's catch firing, and both writes are already benign no-ops there. Happy to add it if you'd rather the three paths look identical.

**Tests:** `apps/api` affected suites green single-fork — `vitest run src/services/fleetFindings/ src/routes/fleetFindings.test.ts` (164 passed), integration `fleetFindings.integration.test.ts` against real Postgres (7 passed, including the new `(c2)` #3637 regression and `(c3)` retry-idempotence/cancelled-guard cases). `tsc --noEmit` clean; eslint clean.

**Mutation-tested the new guards** rather than trusting a token scan — dropping the `cancelled` guard, the `FOR UPDATE` statement, the end-of-chunk recompute, the `markTargetFailed` status guard, the `queued -> running` CAS, or the direct in-flight count each turns the suite red. None of the new assertions are vacuous.

**Status:** review-clean, awaiting maintainer merge. 47/53 checks green with 0 failures at time of writing; the four blocking **Integration Tests** shards plus **Test API**/**Test Web** were still running — worth confirming those landed green before merging.

PR #3658, opened 2026-08-17
fix(web): send-failure banner names the Copy share link control (#3431)

### LINKED ISSUE #3431: [Web] Quote send-failure tells the sender to share the accept link manually, but no copy-link control exists
## Description

When sending a quote fails, the error message tells the sender to **"share the accept link manually"** — but there is no way to obtain that link anywhere in the UI.

The accept URL is returned as `data.acceptUrl` on `POST /quotes/:id/send`, and otherwise exists only in the email that just failed to send. There is no copy-link control on the quote, so the instruction is unfollowable: the user's only recourse is devtools.

## Why it matters

This is the failure path for a **revenue** action. A quote that can't be emailed (bad recipient domain, mail provider hiccup, the `2breeze.app` MX problem in #3363) is exactly when the sender most needs to get the link to the customer another way — and that is the moment the product tells them to do something it doesn't let them do.

## Proposed fix

Add a **Copy accept link** control on the quote — visible on the sent/failed state at minimum, arguably always once the quote has a public token. Then the error message's advice becomes actionable, and the message can point at the control by name.

Worth deciding whether the link should be copyable *before* a send is attempted too: "send me the link, I'll forward it myself" is a normal MSP workflow, and it makes the accept flow usable when email is untrusted or the customer's filter eats it.

## Affected files

- `apps/web/src/components/billing/quotes/QuoteActions.tsx` (the message)
- Quote detail / send flow — wherever the control lands

Related: #3363 (the transactional-email domain has no MX, so replies hard-bounce — same "email is not a reliable channel" theme).

ToddHebebrand:
Closes #3431

## What the issue found

When a quote send fails, the banner tells the sender to *"share the accept link manually"* — and at the time of filing there was no way to obtain that link in the UI.

## What actually shipped since

The control itself landed in #3501 (`Copy share link`, rail button + header kebab, backed by `GET /quotes/:id/share-link`). So the primary ask of the issue is already in `main`.

What was **not** done is the issue's second half: *"the message can point at the control by name."* The failure copy was never updated, so it still reads as advice with nothing named behind it — on the failure path of a revenue action.

## This PR

- The not-delivered banner and the matching post-flip toast now append a hint naming the control. The hint **interpolates the control's own label** (`quotes.actions.copyShareLink`) rather than hard-coding the words, so the name in the hint and the name on the button cannot drift apart in any locale.
- The hint is **withheld when the control is not on screen** — no `quotes:send`, or an expired quote (both refused by `assertLinkableQuote` server-side too). Naming a button the reader cannot see would be the original bug in a new form.
- To stop the banner and toolbar from disagreeing about that, the availability gate is extracted as `quoteShareLinkAvailable(quote, canSendQuotes)` and both call sites read it (previously the toolbar computed it inline).
- `sendEmailWarning.shareLinkHint` added to all eight locales. Existing strings are unchanged, so no translations were invalidated.

## Not changed (deliberate)

A **draft** whose scheduled send failed still has no copy-link control: `assertLinkableQuote` only resolves a link for an open, non-expired quote, because resolving one can *mint* a fresh 30-day accept credential. Extending that to drafts is a design decision, not a copy fix — flagging it rather than quietly widening the credential surface.

## Verification

- `apps/web` `QuoteDetail.send.test.tsx` — 30 passed (3 new: hint present + control rendered; hint withheld without `quotes:send`; hint withheld on an expired quote).
- Mutation-checked: forcing the hint off fails the positive test, so the new assertion is not vacuous.
- `QuoteActions.{test,resend,sendcopy,clone,margincheck}` — 41 passed.
- `src/lib/i18n/` (parity, coverage, terminology, extraction quality) + quote lifecycle/permissions/workspace suites — 121 passed.
- `tsc --noEmit` clean; `astro check` 0 errors / 0 warnings; `eslint` clean on touched files.

### DIFF
--- a/apps/web/src/components/billing/quotes/QuoteActions.tsx
+++ b/apps/web/src/components/billing/quotes/QuoteActions.tsx
@@ -25,11 +25,44 @@ const UNAUTHORIZED = () => void navigateTo('/login', { replace: true });
 
 type TFunction = ReturnType<typeof useTranslation>['t'];
 
+/** Whether the "Copy share link" control is actually offered for this quote.
+ *
+ *  Mirrors assertLinkableQuote (services/quoteLifecycle.ts), which both the
+ *  resend and share-link services call: an OPEN, non-expired quote only.
+ *  Settled outcomes (accepted/declined/converted) and expired quotes are
+ *  excluded because resolving a link can MINT one — manufacturing a fresh
+ *  30-day credential for a finished proposal. The action is quotes:send, not
+ *  quotes:read: it hands the customer a live accept credential.
+ *
+ *  Exported shape rather than an inline expression because the send-failure
+ *  banner has to agree with the toolbar about it: the banner names the control
+ *  by label, and naming a control the reader cannot see is the very bug this
+ *  copy is meant to fix (#3431). */
+export function quoteShareLinkAvailable(
+  quote: Pick<Quote, 'status' | 'expiryDate'>,
+  canSendQuotes: boolean,
+): boolean {
+  const isOpen = quote.status === 'sent' || quote.status === 'viewed';
+  const isExpired =
+    quote.expiryDate != null && new Date(`${quote.expiryDate}T23:59:59Z`).getTime() < Date.now();
+  return canSendQuotes && isOpen && !isExpired;
+}
+
 /** The honest "marked Sent but no email was delivered" copy for a persisted
  *  email-failure reason. Shared by the post-flip toast below and the
  *  persistent banner — unknown codes fall back to the generic send-failed
- *  copy so a new server-side reason never renders blank. */
-function sendEmailWarningMessage(t: TFunction, reason: QuoteSendEmailReason | string, orgName: string): string {
+ *  copy so a new server-side reason never renders blank.
+ *
+ *  `shareHint` is appended when the Copy share link control is on screen. The
+ *  base copy only says the link can be shared by hand; the hint is what makes
+ *  that instruction followable, and it is withheld when the control is not
+ *  offered rather than pointing at a button that isn't there (#3431). */
+function sendEmailWarningMessage(
+  t: TFunction,
+  reason: QuoteSendEmailReason | string,
+  orgName: string,
+  shareLinkAvailable = false,
+): string {
   const warnByReason: Record<QuoteSendEmailReason, string> = {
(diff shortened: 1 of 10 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter`, then a `silent-failure-hunter` verification pass on the fix commit.

**Findings:** 2 raised, both by the silent-failure pass, both confirmed against the code and fixed in `81830ab9d`; 0 outstanding. `code-reviewer` returned 0 findings (predicate extraction verified behavior-preserving vs. the inline logic it replaced; hook ordering and dep array sound; locale keys valid and correctly ordered; no `state.permissions` leak between tests).

1. **HIGH — the hint recommended a link that can be broken for the same reason.** `pdf_render_failed` is set when `loadContractPdfInputs` / `mergeUploadedContractPdfs` throws, and the public accept page re-runs that same contract-render path (`renderContractBlocksForClient` / `loadContractBlockRenderData`, `routes/quotesPublic.ts`) on every open — so "share the link instead" could aim the customer at a page failing for the identical cause. The hint is now withheld for that reason code; its own copy ("Check the attached contract files, then resend") is the remedy that works. Verification confirmed the exclusion set is complete: no other reason code touches contract rendering, and `schedule_failed` is draft-only so it never reaches the hint.

2. **MEDIUM — suppressing the hint left the original dead end in the base copy.** `noBillingContact` / `sendFailed` still said "share the accept link manually" on their own, so a viewer without `quotes:send` (or an expired quote) got the #3431 bug in narrowed form. The base strings no longer promise a manual share at all — they end in "then resend", which a real Re-send control backs — and the hint carries the manual-share instruction in full, appearing only when a control backs it.

**Tests:** `apps/web` quotes + i18n suites green — 59 files, 543 passed (`vitest run src/components/billing/quotes/ src/lib/i18n/`), including 5 new regression tests. The positive hint assertion was mutation-checked: forcing the hint off fails it, so it is not vacuous. `tsc --noEmit` clean; `astro check` 0 errors / 0 warnings; `eslint` clean on touched files.

**Status:** review-clean, awaiting maintainer merge. One judgement call is flagged in the PR body rather than decided here — a **draft** whose scheduled send failed still has no copy-link control, because resolving a link can mint a fresh 30-day accept credential and widening that to drafts is a design decision, not a copy fix.

PR #3659, opened 2026-08-17
fix(scripts): treat partner-wide name matches as bundle-import conflicts (#3450)

### LINKED ISSUE #3450: [UI] Script bundle import marks partner-wide name matches as "New" and duplicates them even in Skip mode
## Description
With the import modal's default availability ('org'), collision detection conflicts only against org-owned rows (`services/scriptBundle/index.ts` `scopeCondition`). Importing a bundle whose script names match existing **partner-wide** scripts therefore shows every entry as **New**, and "Skip it" mode imports them anyway — producing same-name duplicates in the one library list the operator sees (partner-wide + org rows are rendered together).

Repro (local stack, partner admin):
1. Export a bundle of 2 partner-wide scripts (e.g. "QA Script 001/002").
2. Scripts → Import bundle → same file → preview shows both as "New".
3. Collision mode "Skip it" → Import → toast "2 imported, 0 skipped"; /scripts now lists each name twice.

## Suspected root cause / discussion
The org-vs-partner-wide namespace split is deliberate (#3262) — org-target imports must not mutate partner rows. But the *preview annotation* and "skip existing" semantics should probably consider partner-wide rows the caller can see, or at least badge them ("exists partner-wide") so Skip mode does what the operator expects.

Affected files: `apps/api/src/services/scriptBundle/index.ts` (scopeCondition/previewBundle), `apps/web` import modal copy.

Related: #3276, #3245, #3262.

Reported By: internal Playwright UI QA sweep 2026-08-11 — evidence in docs/testing/FEATURE_TEST_LOG.md (pre-release gap-closure pass).

ToddHebebrand:
Closes #3450

## Problem

Script-bundle import collision detection (`scopeCondition`) only matched rows in the resolved **target** scope. With the import modal's default availability (`org`), importing a bundle whose script names match existing **partner-wide** scripts annotated every entry as **New**, and the "Skip it" collision mode imported them anyway — producing same-name duplicates in the one library list the operator sees, since partner-wide and org rows render together.

## Approach

The org-vs-partner-wide **write** split is deliberate (#3262) and is unchanged: an org-target import still never mutates or versions a partner-wide row. What changes is that a *visible but read-only* partner-wide row is now recognised as a collision rather than invisible.

- `findConflictByName` replaces `findExistingByName` and returns `{ row, kind }`. A `target-scope` (owned, writable) match always wins over a read-only `partner-wide` one, and the partner-wide query only runs when the owned lookup misses.
- The partner-wide lookup is gated on the **resolved scope** carrying both an `orgId` and a `partnerId` — deliberately **not** on `auth.scope === 'partner'`. Organization-scope users legitimately *see* their MSP's partner-wide scripts: the list route unions on `partner_id`, and RLS grants the read through the own-partner SELECT branch added in `2026-06-13-catalog-partner-read-branch.sql`. They just cannot write them, which is precisely why this is a separate read-only conflict class. Gating on `auth.scope` would have left org users — the majority of the affected callers — still seeing "New".
- Preview keeps `status: 'name-conflict'` and adds a `conflictKind` refinement rather than introducing a new status value, so a client older than this change renders a generic conflict badge instead of falling through to "Invalid" on an unrecognised enum member.

### Collision-mode semantics against a partner-wide match

| Mode | Before | After |
|---|---|---|
| Skip it | imported a duplicate | **skipped** |
| Rename | renamed only around org names (could land on a partner-wide name) | free name excludes **both** namespaces |
| New version | silently created a duplicate org row | **per-entry error** naming the modes that can proceed — except identical content, which stays a no-op skip, matching the existing idempotent-re-import behaviour |

A per-entry error (rather than silently creating an org-owned override) is the honest outcome: the operator asked to version an existing script, and that specific row cannot be versioned from this scope. The rest of the bundle still imports, consistent with every other per-entry failure mode in the importer.

## UI

The modal badges these entries **"Exists partner-wide"** in amber with a hint line explaining it is read-only here and to Skip or Rename. New keys `bundle.statusConflictPartnerWide` / `bundle.partnerWideConflictHint` added across all eight locales.

## Tests

Five new service tests (preview annotation, owned-wins-over-partner-wide with no second query, skip-mode skip, new-version refusal, new-version identical-content skip, rename across both namespaces), plus negative cases proving the extra lookup does **not** fire for a partner-wide target or a system-scope import. Mutation-checked: stubbing `partnerWideConflictCondition` to return `null` reddens all of them.

- `apps/api/src/services/scriptBundle/index.test.ts` — 38 passed
- `apps/api/src/routes/scriptBundle.test.ts`, `scripts.test.ts`, `scripts.execute-schema.test.ts` — 91 passed / 2 skipped
- `apps/web/src/lib/i18n/localeParity.test.ts`, `apps/web/src/lib/scriptBundle.test.ts` — 63 passed
- `tsc --noEmit` clean in both `apps/api` and `apps/web`

## Not addressed

The mirror case — a **partner-wide** import colliding with an org-owned name in some org under the partner — is out of scope here and not sanely fixable at import time (there is no single org to check against, and the conflict would be per-org). No migration; no schema change.

### DIFF
--- a/apps/api/src/services/scriptBundle/index.test.ts
+++ b/apps/api/src/services/scriptBundle/index.test.ts
@@ -485,6 +485,32 @@ describe('importBundle', () => {
     expect(h.state.updates).toHaveLength(0);
   });
 
+  it('isolates the partner-wide new-version refusal to its own entry — the rest of the bundle still imports', async () => {
+    const bundle = validBundle([baseEntry, { ...baseEntry, name: 'Unrelated script' }]);
+    h.state.selectQueue.push(
+      [], // entry 1: no org-owned match
+      [{ id: SCRIPT_ID, name: baseEntry.name, version: 4, content: 'DIFFERENT content' }],
+      [], // entry 2: no org-owned match
+      [] // entry 2: no partner-wide match either → imports normally
+    );
+    const result = await importBundle(makeAuth(), bundle, {
+      mode: 'new-version',
+      availability: 'org'
+    });
+    expect('error' in result).toBe(false);
+    if ('errors' in result) {
+      expect(result.errors).toHaveLength(1);
+      expect(result.errors[0]!.index).toBe(0);
+      // The refusal must `continue`, not abort the loop.
+      expect(result.imported).toBe(1);
+      expect(result.scripts).toHaveLength(1);
+      expect(result.scripts[0]).toMatchObject({ index: 1, name: 'Unrelated script', action: 'imported' });
+    }
+    const scriptInserts = h.state.inserts.filter((i) => i.table === scripts);
+    expect(scriptInserts).toHaveLength(1);
+    expect((scriptInserts[0]!.values as Record<string, unknown>).name).toBe('Unrelated script');
+  });
+
   it('new-version mode treats an identical-content partner-wide match as a no-op skip', async () => {
     h.state.selectQueue.push(
       [],
--- a/apps/api/src/services/scriptBundle/index.test.ts
+++ b/apps/api/src/services/scriptBundle/index.test.ts
@@ -392,7 +392,8 @@ describe('importBundle', () => {
   it('resolves tags by name in the target scope: reuses existing, creates missing, links both', async () => {
     const bundle = validBundle([{ ...baseEntry, tags: ['printing', 'windows'] }]);
     h.state.selectQueue.push(
-      [], // findExistingByName → none
+      [], // findConflictByName → no org-owned match
+      [], // findConflictByName → no partner-wide match (#3450)
       [{ id: TAG_ID, name: 'printing' }] // ensureTagIds → 'printing' exists, 'windows' missing
     );
     const result = await importBundle(makeAuth(), bundle, { mode: 'skip', availability: 'org' });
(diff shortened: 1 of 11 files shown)


ToddHebebrand:

**Findings:** 1 actionable → addressed in `a4eb434`; 0 outstanding.
- *pr-test-analyzer (5/10)*: the new-version refusal branch's comment promises "the rest of the bundle still imports", but every #3450 test used a single-entry bundle — turning the `continue` into a loop abort would have gone uncaught. Added a two-entry test; mutation-verified that `continue` → `break` reddens that test and only that test.
- *code-reviewer*: no issues. Confirmed the #3262 invariant holds — no path reaches a `scripts` update/insert against a `partner-wide` conflict — and that the `or()` inside `and(inArray(...), ...)` produces the intended SQL.
- *silent-failure-hunter*: no issues. Specifically ruled out the zero-row trap: `scope.partnerId` and the `breeze.current_partner_id` GUC backing the RLS read branch both derive from the same request-scoped `AuthContext.partnerId`, with no re-entry seam where they could diverge. Also confirmed the per-entry error reaches the modal's error list, and that an absent/unknown `conflictKind` falls back to the generic conflict badge rather than "Invalid".

**Tests:** `apps/api/src/services/scriptBundle/index.test.ts` 39 passed; `routes/scriptBundle.test.ts` + `scripts.test.ts` + `scripts.execute-schema.test.ts` 91 passed / 2 skipped; `apps/web` `localeParity.test.ts` + `scriptBundle.test.ts` 63 passed. `tsc --noEmit` clean in both `apps/api` and `apps/web`. CI: **54/54 checks green**, including all 4 integration shards.

**Noted, not fixed (pre-existing, both reviewers agreed not to block):** `apps/web/src/components/scripts/ScriptBundleImport.tsx` has no component test file on `main`, so the new badge branch is covered only at the service layer. Worth a follow-up issue rather than growing this PR.

Issue #3516, opened 2026-08-14
Agent software inventory silently stops syncing on hosts with >1MB package lists (413, error discarded)


ToddHebebrand:
Found while fixing #3482 (PR #3511). Not introduced by that PR — pre-existing, and separate enough to warrant its own fix.

## What happens

`PUT /api/v1/agents/:id/software` registers a route-level `bodyLimit` of 5 MB, but the global gate in `apps/api/src/index.ts:505` runs at `app.use('*')` before any route is mounted, so the effective cap is the 1 MB default from `bodyLimitForPath`. A route-level limit can only ever make a path tighter, never looser.

The same applies to the sibling agent ingest routes: `/hardware`, `/disks`, `/network` (`routes/agents/inventory.ts`), `/heartbeat` (`routes/agents/heartbeat.ts`), and `/connections` (`routes/agents/connections.ts`).

## Why it's silent

The failure is invisible from both ends:

- `agent/internal/heartbeat/heartbeat.go:1764-1780` sends software inventory as plain `json.Marshal` — no gzip, no chunking — even though the agent gzips request bodies elsewhere (`internal/logging/shipper.go:451`, `internal/workspaceindex/client.go:293`).
- `agent/internal/httputil/retry.go:37-43` — `isRetryableStatus` covers 429/500/502/503/504. **413 is not retryable.**
- `agent/internal/heartbeat/heartbeat.go:2060` — `sendSoftwareInventory`'s error is **discarded**. Same at `:2070` (disks) and `:2073+` (network). A 413 produces one `log.Warn` and nothing else: no failure counter, no health-monitor update, no server-side signal.
- The API side logs nothing either (see the companion issue on gate telemetry).

Net effect: the device re-sends the identical oversized body every 15 minutes forever, inventory goes permanently stale, and it is indistinguishable from "this host installed nothing new."

## Who it hits

Payloads run ~250-450 bytes/entry, so 1 MB is roughly 2,500-4,000 entries.

- **Linux: routine.** `software_linux.go` enumerates dpkg/rpm up to `collectorResultLimit = 5000` (`internal/collectors/command_limits.go:21`); a full desktop distro carries 2,000-4,000 packages.
- **Windows: edge case** (RDS/build hosts). `collectors/software_windows.go:28-49` has no item cap at all.

The server schema allows 10,000 items (`routes/agents/schemas.ts:555-565`), so the byte gate always binds first — the schema's limit is never the one that fires.

[...]

Issue #2401, opened 2026-07-12
[API] Expose capture_pprof as an MCP tool (capture_agent_pprof)


ToddHebebrand:
## Description

PR #2394 (#2389) adds the `capture_pprof` agent command (heap/goroutine profiles returned base64 via the command result, no listener). Making it AI-invokable was deliberately deferred because MCP tool registration requires a multi-site sweep, and partial registration is a known drift trap (the AI gate reads TWO maps — a tool registered in one but not the other silently misbehaves).

## Proposed Fix

Register `capture_agent_pprof` following the `set_agent_log_level` pattern (`apps/api/src/services/aiToolsAgentLogs.ts:150-230`), covering ALL registration sites in one PR:
- SDK tool map / registry hub
- guardrails permission map AND rate-limit map (both — dual-map drift)
- approval switch (approval-gated, Tier-2, like other device-touching diagnostics)
- tool schemas
- web tierConfig (×3 sites)

Result should surface profile metadata (sizes, capture time) and hand back the artifact in a way the chat UI can offer for download, not a giant inline base64 blob in the transcript.

## Blocked On
PR #2394 — the agent command and `CAPTURE_PPROF` command type only exist on that branch.

## Affected Files
- `apps/api/src/services/aiToolsAgentLogs.ts` (or a new aiToolsDiag file per the hub/per-domain pattern)
- guardrails + tierConfig registration sites

Refs #2389.

ToddHebebrand:
PR #2409 implements this: `capture_agent_pprof` registered at all sites in one pass (tool implementation, SDK tool map + registration, guardrails permission **and** rate-limit maps, approval description, tool schemas, system prompt, web tierConfig ×3 — the parity test passes with no new legacy-gap entries).

Two things the PR fixes beyond registration, found during review, that also affected the #2394 human path:
- Result-ingest secret redaction was statistically corrupting large base64 stdout — `capture_pprof` results are now stored byte-for-byte on both ingest legs.
- The single-command GET now passes profile stdout through (history redaction previously made the profiles unretrievable at all), and the heartbeat/REST result leg got a body-limit carve-out so multi-MB results don't 413 into fake timeouts.

The tool returns metadata only (sizes, capture time, runtime gauges incl. goroutine count) plus a retrieval pointer — no base64 in the chat transcript.

PR #3660, opened 2026-08-17
fix(api): give the global body-limit gate server-side telemetry (#3517)

### LINKED ISSUE #3517: Global body-limit gate 413s with zero server-side telemetry
Found while fixing #3482 (PR #3511). Pre-existing; deliberately left out of that PR because the fix has a design constraint (below) rather than being a one-liner.

## Problem

`apps/api/src/index.ts:510-511`:

```ts
const { maxSize, error } = bodyLimitForPath(c.req.path);
return bodyLimit({ maxSize, onError: (ctx) => ctx.json({ error }, 413) })(c, next);
```

`onError` returns JSON and logs nothing — no structured log, no error ID, no Sentry event. **Every 413 this middleware has ever produced is invisible server-side.**

That is precisely why this class of bug keeps reaching customers before it reaches us. #1377, #2401 and #3482 were all found by a user hitting a wall, because a route whose body limit was silently capped at 1 MB looked perfectly healthy from the operator side. #3516 is the same shape on the agent path, where nobody is watching at all.

## Why it isn't a one-liner

The obvious fix — log `c.req.path` — violates the repo's path-logging rule. `middleware/requestPathLogger.ts` never emits a raw request path; it uses `newRequestCorrelationId` before routing and the bounded `safeMatchedRouteLabel` template after. The body gate runs at `app.use('*')` **before routing**, so there is no matched-route label available at the point the 413 is raised, and the raw path is exactly what we're not allowed to log.

So this needs an actual decision, roughly one of:

1. Emit the `bodyLimitForPath` **carve-out identity** (a stable label for the branch that matched, e.g. `quote-image` / `default`) rather than the path. Bounded and safe by construction, and it is the dimension you actually want to group on.
2. Defer the log until after routing so `safeMatchedRouteLabel` is available — awkward, since the gate short-circuits.
3. Extend the safe-label helper to produce a bounded pre-routing label.

Option 1 looks right: the carve-out label is already a closed set, needs no new redaction machinery, and answers the operational question directly ("which limit is firing, and how often").

Also worth including: method, the configured `maxSize`, and `content-length`. Sample or rate-limit if agent-path volume is a concern — but it should not stay silent.

## Payoff

With this in place, the next shadowed upload route shows up as a Sentry/log cluster keyed on the carve-out label within hours, instead of surviving to a customer bug report. PR #3511 adds a drift guard that prevents *new* route-level limits from being silently shadowed; this issue covers detecting the ones already in the wild.

ToddHebebrand:
Closes #3517

## Problem

`apps/api/src/index.ts` mounted the global body-size gate as a bare
`bodyLimit({ maxSize, onError: (ctx) => ctx.json({ error }, 413) })`. `onError`
returned JSON and logged nothing — no structured log, no error ID, no Sentry
event. **Every 413 this middleware has ever produced was invisible
server-side.**

That is why this class of bug keeps reaching customers first: #1377, #2401 and
#3482 were all found by a user hitting a wall, because a route whose limit was
silently shadowed at 1MB looked perfectly healthy from the operator side.
#3516 is the same shape on the agent ingest path, where nobody is watching at
all.

## Approach — option 1 from the issue (carve-out identity, not the path)

Logging `c.req.path` is off the table: `middleware/requestPathLogger.ts` never
emits a raw request path, and this gate runs at `app.use('*')` **before**
routing, so `safeMatchedRouteLabel` has nothing to work with yet.

Instead every branch of `bodyLimitForPath` now returns a stable `rule` label
from a closed `BodyLimitRule` union. It is bounded by construction, needs no
new redaction machinery, and is the dimension operators actually want to group
on — *which limit is firing, and how often*. Adding a carve-out without a label
is a type error.

## What changed

- **`middleware/bodyLimit.ts`** — `bodyLimitForPath` returns `BodyLimitPolicy
  { rule, maxSize, error }`; each of the 10 carve-outs gets a distinct label.
- **`middleware/bodyLimitGate.ts`** (new) — the gate lifted out of `index.ts`,
  now emitting on rejection: `rule`, `method`, configured `max_size`, validated
  `content_length`, and the request correlation ID. `Content-Length` is
  caller-controlled, so it is only echoed when it is a plain decimal integer
  that round-trips; otherwise `unknown`.
- **`requestPathLogger` / `safeRequestLabel`** — the logger publishes its
  correlation ID on the context; `requestCorrelationId(c)` re-validates it as a
  canonical UUID or returns `unknown`, so a forged `X-Request-Id` can never
  reach a log line or a Sentry tag.
- **Sentry** — one event per rule per 5 minutes (a misconfigured limit fires on
  every request from every affected client; its *presence*, not its count,
  drives the fix). The console line is **never** throttled, so self-hosted
  operators keep the complete record. `body_limit_rule` / `body_limit_max_size`
  are allowlisted in `sentry.ts` — `scrubEvent` deletes `message`, `logentry`
  and `extra`, so without them the event would arrive contentless and
  ungroupable.
- **Drift guard (#3511)** — `bodyLimitGate.ts` added to `SCAN_EXEMPT`; it *is*
  the global gate, not a shadowed route-level limit.

## Behavior preserved

Status codes, error message bodies, every `maxSize`, and the `/oauth` bypass
(oidc-provider reads the raw Node stream itself) are unchanged. This PR adds
telemetry only.

## Tests

- `middleware/bodyLimitGate.test.ts` (new, 7 cases): asserts the emitted record
  never contains the raw path or the query string, that the matched carve-out
  rule is reported (not just `default`), Sentry throttling vs. complete console
  logging, `request_id=unknown` for a forged `X-Request-Id`, `content_length=unknown`
  for a streamed body, `/oauth` passthrough, and silence under the limit.
- `middleware/bodyLimit.test.ts`: existing assertions extended with `rule`, plus
  a new case proving every carve-out has a **distinct, non-`default`** label
  that is safe as a Sentry tag value.
- `services/sentry.test.ts`: the two new tags survive the allowlist.

`vitest run` (serial) on bodyLimit / bodyLimitGate / sentry / requestPathLogger:
**55 passed**. `tsc --noEmit` on `apps/api`: clean.

### DIFF
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -11,11 +11,10 @@ import { cors } from 'hono/cors';
 import { HTTPException } from 'hono/http-exception';
 import { prettyJSON } from 'hono/pretty-json';
 import { secureHeaders } from 'hono/secure-headers';
-import { bodyLimit } from 'hono/body-limit';
 
 import { securityMiddleware } from './middleware/security';
 import { requestPathLogger } from './middleware/requestPathLogger';
-import { bodyLimitForPath } from './middleware/bodyLimit';
+import { createGlobalBodyLimitMiddleware } from './middleware/bodyLimitGate';
 import { globalRateLimit } from './middleware/globalRateLimit';
 import { authRoutes } from './routes/auth';
 import { accountDeletionAdminRoutes } from './routes/auth/accountDeletion';
@@ -492,14 +491,12 @@ app.use(
   })
 );
 app.use('*', securityMiddleware());
-app.use('*', async (c, next) => {
-  // oidc-provider reads the raw Node IncomingMessage stream itself.
-  if (c.req.path === '/oauth' || c.req.path.startsWith('/oauth/')) {
-    return next();
-  }
-  const { maxSize, error } = bodyLimitForPath(c.req.path);
-  return bodyLimit({ maxSize, onError: (ctx) => ctx.json({ error }, 413) })(c, next);
-});
+app.use(
+  '*',
+  createGlobalBodyLimitMiddleware({
+    capture: (message, tags) => captureMessage(message, 'warning', undefined, tags),
+  })
+);
 app.use('*', globalRateLimit());
 app.use('*', prettyJSON());
 app.use(
--- a/apps/api/src/middleware/bodyLimit.test.ts
+++ b/apps/api/src/middleware/bodyLimit.test.ts
@@ -9,10 +9,12 @@ const KB = 1024;
 describe('bodyLimitForPath', () => {
   it('applies the tight 1MB default to ordinary routes', () => {
     expect(bodyLimitForPath('/api/v1/devices')).toEqual({
+      rule: 'default',
       maxSize: 1 * MB,
       error: 'Request body too large',
     });
     expect(bodyLimitForPath('/api/v1/software/catalog')).toEqual({
+      rule: 'default',
       maxSize: 1 * MB,
       error: 'Request body too large',
     });
@@ -20,10 +22,12 @@ describe('bodyLimitForPath', () => {
 
   it('carves out dev-push binary uploads at 150MB', () => {
     expect(bodyLimitForPath('/api/v1/dev/push')).toEqual({
+      rule: 'dev-push',
       maxSize: 150 * MB,
       error: 'Binary too large (max 150MB)',
     });
     expect(bodyLimitForPath('/api/v1/dev/push/anything')).toEqual({
+      rule: 'dev-push',
       maxSize: 150 * MB,
       error: 'Binary too large (max 150MB)',
     });
@@ -37,6 +41,7 @@ describe('bodyLimitForPath', () => {
     expect(
       bodyLimitForPath('/api/v1/agents/agent-1/commands/11111111-1111-4111-8111-111111111111/result'),
     ).toEqual({
+      rule: 'agent-command-result',
       maxSize: 12 * MB,
       error: 'Command result too large (max 12MB)',
     });
@@ -51,6 +56,7 @@ describe('bodyLimitForPath', () => {
   // the agent's WS read limit is derived from the same cap (issue #2399).
   it('carves out file-browser uploads at 8MB', () => {
     expect(bodyLimitForPath('/api/v1/system-tools/devices/dev-1/files/upload')).toEqual({
+      rule: 'file-upload',
       maxSize: 8 * MB,
       error: 'File too large (max 4MB)',
     });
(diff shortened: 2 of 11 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer; plus a focused silent-failure re-review of the fix commit (`c478151ce`) because the fixes touched two new route files.

**Findings:** 5 raised → 5 addressed; 0 outstanding.

- **code-reviewer — 0 findings.** Confirmed behavior preserved exactly (status codes, error bodies, every `maxSize`, the `/oauth` bypass), no caller-controlled string reaching a log line or Sentry tag, and that the `SCAN_EXEMPT` change is a 1:1 move of the `bodyLimit(` call out of the already-exempt `index.ts` — not a weakened drift guard.
- **silent-failure-hunter — 2 findings, both fixed in `c478151ce`:**
  - *Critical:* `agents/logs.ts` and `agents/processSample.ts` declare their **own** 256KB limits — **tighter** than the global 1MB default — so the newly-instrumented global gate never answers for them and their 413s stayed fully silent, on exactly the agent-authenticated paths the issue calls out as "nobody is watching". Extracted `reportBodyLimitRejection` / `bodyLimitOnError` and wired both routes; the Sentry throttle is now module state shared by both legs so a rule firing on each is throttled once.
  - *Medium:* `capture` was called unguarded on the error path, so a faulting Sentry SDK would have escalated a precise 413 into an opaque 500 — a regression **introduced** by adding telemetry. Both sinks are now `try`/`catch`'d; a broken sink loses the record, never the response.
- **pr-test-analyzer — 2 findings, both fixed in `c478151ce`:**
  - `safeContentLength`'s two guards were unmutation-tested (hono short-circuits on an unparseable `Content-Length`, so a forged header never reaches `onError` over HTTP). Exported it and tested 12 malformed inputs directly.
  - The throttle test exercised only one rule, so a *globally*-keyed throttle would have passed it identically. Now asserts two different rules in the same window both report.
- **Re-review of the fix — 1 finding, fixed in `208fb2c4f`:** my comment on the pre-existing `BodyLimitError` catch branch claimed hono throws on the streaming leg. It does not — under the pinned hono both legs `return onError(c)`, and `BodyLimitError` is defined nowhere in the tree, so the branch is unreachable today. Kept as explicit defense (a future hono reinstating a throwing path must not silently regress to a generic 400) but the rationale now says so honestly. Re-review also confirmed: no import cycle, no double-reporting, throttle keying bounded and correct, no behavior change beyond telemetry.

**Tests:** `vitest run` serial — 62 passed across `bodyLimit` / `bodyLimitGate` / `sentry` / `agents/logs`; the full `src/routes/agents` tree plus the affected middleware ran green at 867 passed / 49 files. `tsc --noEmit` on `apps/api` clean. (`bodyLimit.test.ts`'s directory-scan drift guard times out on a *cold* FS at the default 5s and passes at ~800ms warm — pre-existing, not introduced here.)

**Status:** review-clean, awaiting CI and maintainer merge.

PR #3661, opened 2026-08-17
fix(api,agent): fail loudly when a release is registered without a signed manifest (#3544)

### LINKED ISSUE #3544: [Binary Sync] One untrusted/rejected release asset aborts sync for the ENTIRE release, including unaffected platforms (macOS/Linux never registered when only the Windows asset fails trust validation)
### Description

After upgrading to v0.105.1, [binarySync] fails and aborts entirely on the first untrusted asset it
encounters (the unsigned Windows .exe, correctly rejected by the new edition/platformTrust
validation) — but this takes down registration for every OTHER asset in the same release too,
including macOS and Linux binaries that have no signing/trust issue at all and would otherwise
register cleanly.

Net effect: a single platform's trust-validation failure silently blocks agent version updates for
ALL platforms until the underlying asset is fixed (i.e., until the maintainers ship a signed
Windows build) — even though macOS/Linux users have no reason to be affected by a Windows-specific
signing gap.

### Steps to Reproduce

1. Self-hosted instance, BINARY_SOURCE=github, upgrade to a release where at least one platform's
   asset fails the new platformTrust validation (v0.105.1: Windows .exe is unsigned,
   platformTrust: "windows-authenticode-required" not met).
2. On API startup (or any breeze-api restart), binarySync runs and logs:
   [binarySync] BINARY_SOURCE=github, syncing from GitHub releases
   [startup] Binary sync failed (non-fatal in github mode): Error: Release asset
   breeze-agent-windows-amd64.exe requires platformTrust "windows-authenticode-required", got "none"
3. Check agent_versions for the new release version across ALL platforms:
   SELECT component, platform, version FROM agent_versions WHERE version = '0.105.1';
   -> 0 rows. Not just Windows — macOS and Linux never registered either.
4. Reproduced twice, via two independent breeze-api restarts, identical result both times — the
   sync deterministically fails at the same asset and aborts the same way every time, since it's
   only triggered at API startup (no retry/cron for a partial sync).

### Expected Behavior

A trust-validation failure on one platform's asset should be logged and that specific asset
skipped, while the sync continues to register every other asset in the release that passes
validation. macOS and Linux agent versions should be available in agent_versions and selectable
in "Update version targets" regardless of an unrelated Windows-specific signing gap.

### Actual Behavior

The entire sync operation aborts on the first failing asset, registering nothing from that release
for any platform. Confirmed via direct query and reproduced across two separate sync attempts (two
breeze-api restarts, ~15 minutes apart, identical failure and identical empty result both times).

### Component

API Server

### Version

Web/API 0.105.1, self-hosted via Docker Compose, BINARY_SOURCE=github

### Relevant Logs

```shell
[binarySync] BINARY_SOURCE=github, syncing from GitHub releases
[startup] Binary sync failed (non-fatal in github mode): Error: Release asset
breeze-agent-windows-amd64.exe requires platformTrust "windows-authenticode-required", got "none"
```

### Additional Context

This compounds with the intentional Windows-signing change in v0.105.0 (public Windows agent is
now unsigned by design, per that release's own notes) — meaning every self-hosted instance that
hasn't yet set up the new BYO signing pipeline will hit this on every future release until either
(a) they configure BYO signing, or (b) this sync is made per-asset fault-tolerant instead of
all-or-nothing. Suggest wrapping each asset's registration in its own try/catch inside the sync
loop, logging and skipping a failed asset rather than throwing out of the whole batch.

Practical impact tonight: macOS and Linux agent versions for v0.105.1 are completely unavailable in
"Update version targets" for our instance, blocking planned testing on those platforms too, despite
those binaries having no trust/signing issue of their own.

ToddHebebrand:
## What

Issue #3544's titled bug — one rejected asset aborting the whole binary sync — was fixed in #3538. The issue stayed open for the **two remaining halves** the maintainer identified in [this comment](https://github.com/LanternOps/breeze/issues/3544#issuecomment-5309260366): the failure is *silent and unbounded* on the agent side, and *nothing warns at write time*. This PR closes both.

## Root cause

`validateReleaseManifest` is mandatory on the **serving** path — `GET /agent-versions/:version/download` returns `409 {error, reason}` whenever a row's `release_manifest` / `manifest_signature` is missing or unverifiable (`agentVersions.ts:526`). But **no write path ever ran it**. The columns are nullable in the schema while the serving path treats them as required, so a row registers cleanly and then 409s on every single read, forever.

Reported symptom: every device in the fleet retrying every ~60s indefinitely, with `download info request failed with status 409` as the only signal — while the server was already returning a specific, actionable `reason` in the body that the agent threw away.

## API — reject at the door, not thousands of 409s later

- **`POST /agent-versions`** validates the submitted payload and returns **422** `{error, reason}` on failure. This runs **before** the `isLatest` demotion. Ordering is the point: demoting first and validating later strips the last *working* upgrade target off the fleet on behalf of a row we're about to refuse — turning a rejected request into an outage.
- **`POST /agent-versions/promote`** validates **every** target row before the demotion transaction, returning **409** `{error, reason, invalidTargets}` if **any** row fails. Promotion is the write that actually points the fleet at a version, so it's the one that causes the outage. Rejecting on *any* failure rather than *all* keeps promotion atomic — a partial promotion would split the fleet across platforms/components.

`binarySync` is unaffected: all four of its write sites derive the manifest from upstream signature verification, so normal registration and promotion still pass (its 50 tests stay green).

## Agent — surface the reason, stop retrying a doomed target

- `parseDownloadInfo` handles 409 explicitly, decoding the control plane's machine-readable `reason` and wrapping it in a new `ErrUntrustedRelease` sentinel. The reason is **sanitized** (lowercase snake_case only, length-bounded, body size-capped) rather than echoed verbatim — consistent with how the redirect branch and `SafeDownloadErrorFields` already treat server-supplied text that reaches agent logs.
- `doUpgrade` treats `ErrUntrustedRelease` as terminal for that target version and applies a **30-minute per-version cooldown**, mirroring the existing `watchdogUpgradeRetryCooldown`. Deliberately a cooldown rather than `setAutoUpdate(false)` (what `ErrReadOnlyFS` does): re-registering the version with a signed manifest must recover the fleet on its own. Promoting a *different* version bypasses the cooldown immediately, so the operator's fix takes effect on the next heartbeat. The check runs before `sendUpdateStatus` and before any download work, so a backed-off device also stops announcing "Updating".

The new log line names the actual problem and the fix, instead of a bare status code.

## Design note

The reject-vs-warn choice and the ordering hazards were put to an advisor quorum (independent Codex review) before implementing; it independently reached hard-reject on both endpoints, any-failure semantics for promote, and validate-before-demote — matching the implementation here.

## Testing

| Suite | Result |
|---|---|
| `apps/api` `agentVersions.test.ts` | **58 passed** (52 existing + 6 new) |
| `apps/api` binarySync / selftest / releaseTrustChain | **50 passed** (no regressions) |
| `agent` `go test -race ./internal/updater/...` | **ok** |
| `agent` `go test -race ./internal/heartbeat/...` | **ok** |
| `tsc --noEmit` (apps/api) | **clean** |
| `go build ./...` + `go vet` | **clean** |

New API tests cover: 422 on missing manifest, 422 on unparseable manifest, **a refused `isLatest:true` create never demoting the currently-promoted row**, promote rejecting a NULL-manifest target without running the transaction, a **mixed** valid/invalid promote being rejected whole, and the happy path still succeeding.

New Go tests cover: reason extraction, sanitization (uppercase / URLs / spaces / newline-injection / digits / overlong all dropped), body-size bounds, 409-without-reason still terminal, **non-409 statuses staying retryable**, and the cooldown's version-keying, expiry, and concurrency safety under `-race`.

9 existing API tests were updated to supply valid manifests — the guards were **not** weakened to accommodate them. Verified non-vacuous by mutation: neutering both guards fails 5 tests.

Closes #3544

### DIFF
--- a/agent/internal/heartbeat/heartbeat.go
+++ b/agent/internal/heartbeat/heartbeat.go
@@ -536,6 +536,13 @@ type Heartbeat struct {
 	// Tracks whether the read-only FS error has been logged (prevents log spam)
 	updateReadOnlyLogged bool
 
+	// Cooldown state for an upgrade target the server refused as untrusted
+	// (updater.ErrUntrustedRelease / HTTP 409). Guarded by untrustedReleaseMu
+	// because doUpgrade runs on a goroutine per heartbeat. Issue #3544.
+	untrustedReleaseMu  sync.Mutex
+	untrustedReleaseVer string
+	untrustedReleaseAt  time.Time
+
 	// Path to the agent state file, set by main after startup.
 	statePath string
 
@@ -6578,8 +6585,48 @@ func (h *Heartbeat) reconcileUserHelperFromExecutable() {
 	h.reconcileUserHelper(binaryPath)
 }
 
+// untrustedReleaseRetryCooldown bounds how often an upgrade target the server
+// has already refused to serve (HTTP 409, updater.ErrUntrustedRelease) is
+// retried. The condition is terminal for that version — no amount of retrying
+// on the device can produce a signed manifest — but it is not permanent: an
+// operator re-registering the version with a signed manifest must recover
+// automatically, so this is a cooldown rather than a hard disable (unlike
+// ErrReadOnlyFS, which calls setAutoUpdate(false)). Mirrors
+// watchdogUpgradeRetryCooldown. Issue #3544, where every device in a fleet
+// re-attempted the same doomed upgrade every ~60s indefinitely.
+const untrustedReleaseRetryCooldown = 30 * time.Minute
+
+// untrustedReleaseBackoffActive reports whether targetVersion was already
+// refused as untrusted within the cooldown window. Tracked per version so a
+// NEW upgrade target is always attempted immediately.
+func (h *Heartbeat) untrustedReleaseBackoffActive(targetVersion string) bool {
+	h.untrustedReleaseMu.Lock()
+	defer h.untrustedReleaseMu.Unlock()
+	return h.untrustedReleaseVer == targetVersion &&
+		time.Since(h.untrustedReleaseAt) < untrustedReleaseRetryCooldown
+}
+
+// noteUntrustedRelease starts (or restarts) the cooldown for targetVersion.
+func (h *Heartbeat) noteUntrustedRelease(targetVersion string) {
+	h.untrustedReleaseMu.Lock()
+	defer h.untrustedReleaseMu.Unlock()
+	h.untrustedReleaseVer = targetVersion
+	h.untrustedReleaseAt = time.Now()
+}
+
 // doUpgrade contains the actual upgrade logic, called by handleUpgrade.
 func (h *Heartbeat) doUpgrade(targetVersion string) {
+	// Checked before sendUpdateStatus and before any download work: the server
+	// re-sends the same upgradeTo on every heartbeat, so without this the
+	// device would keep announcing "Updating" and re-running the whole
+	// prefetch + download path for a version the server is guaranteed to
+	// refuse again.
+	if h.untrustedReleaseBackoffActive(targetVersion) {
+		log.Debug("upgrade skipped: server recently refused this version as untrusted; backing off",
+			"targetVersion", targetVersion)
+		return
+	}
+
 	log.Info("upgrade requested", "targetVersion", targetVersion)
 
 	h.sendUpdateStatus(targetVersion)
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter, plus a targeted re-review of the fix commit. An independent Codex read-only quorum ran on the design *before* implementation (reject-vs-warn, promote's any-vs-all semantics, validate-before-demote ordering).

**Findings:** 3 raised → all 3 addressed; 0 outstanding.

- **code-reviewer: 0 findings.** Confirmed no other caller POSTs to `/agent-versions` or `/promote` anywhere in `apps/`, `ee/`, `packages/`, `scripts/`, `.github/workflows/` — so the new 422/409 is not a breaking change; the two smoke workflows only `GET .../download`. Also cleared the `resp.Body` lifecycle, mutex lock-ordering, and `upgradeInProgress` cleanup on the new early return.
- **silent-failure-hunter: 1 HIGH** — `downloadInfoRejectionReason` returned a bare `""` both when the server sent *no* reason and when it sent one that failed sanitization, collapsing a future API/agent vocabulary drift into an invisible "server gave no reason". Fixed in `4ed67cf94`: now returns `(reason, malformed)` and reports the refused case distinctly. I deliberately did **not** take the reviewer's suggestion to echo the raw refused value — that would defeat the sanitization keeping server-supplied text out of agent logs (the `SafeDownloadErrorFields` convention). Naming the *shape* of the problem is enough to diagnose a drift without carrying the payload.
- **Re-review: 1 MEDIUM residual** — a `reason` large enough to be truncated by the 4KB `LimitReader` still collapsed into "no reason" via the parse-error path, i.e. the same bug through size rather than charset, with the test enshrining it. Reviewer called it follow-up-worthy, not blocking; closed anyway in `1f6a86a` since it was two lines.

**Tests:** `apps/api` 106 passed (`agentVersions` 58 · `binarySync` · `releaseTrustChain`, `--pool=forks --no-file-parallelism`); `agent` `go test -race ./internal/updater/... ./internal/heartbeat/...` ok; `tsc --noEmit` clean; `go build`/`go vet`/`gofmt` clean.

**Non-vacuous verification:** both API guards mutation-tested (neutering them fails 5 tests), and the truncation flag mutation-tested (pinning it false fails its test). 9 pre-existing API tests were updated to supply valid manifests — the guards were **not** weakened to accommodate them.

**Status:** review-clean, CI green so far (0 failures, run still completing), awaiting maintainer merge.

**One note for the maintainer:** this closes the *two remaining halves* of #3544 (silent/unbounded agent failure + no warning at write time). The titled sync-abort was already fixed in #3538. Worth confirming that's the scope you intended before merging, since the issue title still describes the shipped half.

Issue #3641, opened 2026-08-17
[API] avProducts is validated on every security-status post, used only for unreachable fallbacks, and never persisted


bdunncompany:
## Summary

The agent posts a per-product `avProducts[]` array with the security-status payload. The API validates it, reads it at exactly one place for a fallback that **cannot fire for a real agent payload**, and then discards it. Nothing persists it and nothing surfaces it, so after the fact there is no way to answer "which AV product claimed real-time protection was on?"

That question is currently blocking #3593.

## What the agent sends

`securityStatusIngestSchema` accepts up to 50 products, each with the fields that would identify the culprit (`apps/api/src/routes/agents/schemas.ts:420`):

```ts
avProducts: z.array(
  z.object({
    displayName: z.string().optional(),
    provider: z.string().optional(),
    realTimeProtection: z.boolean().optional(),
    definitionsUpToDate: z.boolean().optional(),
    productState: z.number().int().optional()
  })
).max(50).optional()
```

On Windows these come from a WMI SecurityCenter2 enumeration of **every** registered AV product, with `productState` decoded agent-side into the per-product `realTimeProtection`.

## What the API does with it

Every non-test reference in the repo:

```
apps/api/src/routes/agents/helpers.ts:487  const avProducts = Array.isArray(payload.avProducts) ? payload.avProducts : [];
apps/api/src/routes/agents/helpers.ts:488  const preferredProduct = avProducts.find((p) => p.realTimeProtection) ?? avProducts[0];
apps/api/src/routes/agents/schemas.ts:420  avProducts: z.array(
```

`preferredProduct` feeds only the nullish fallbacks in the upsert:

```ts
provider: normalizeProvider(payload.provider ?? preferredProduct?.provider)
realTimeProtection: payload.realTimeProtection ?? preferredProduct?.realTimeProtection ?? false
```

**Those fallbacks are unreachable from the agent.** `SecurityStatus.RealTimeProtection` and `.Provider` are plain non-pointer fields with no `omitempty` (`agent/internal/security/status.go`), so both are always marshalled — `false` and `""` included — and `false` is not nullish. For any payload the 0.105.x agent actually produces, `preferredProduct` is computed and then never consulted.

So the array is parsed and validated on every submission, used for dead code, and dropped.

## What's missing

`security_status` has no `av_products` column (`apps/api/src/db/schema/security.ts`), and `avProducts` appears nowhere in `apps/web`. The data is not stored, not returned by the API, and not shown in the UI.

[...]

PR #3662, opened 2026-08-17
fix(agent): Defender no longer claims a third-party AV's real-time protection (#3593)

### LINKED ISSUE #3593: [Security Status] real_time_protection reported as TRUE for Windows Defender despite Defender being fully disabled — confirmed via Get-MpComputerStatus and WMI SecurityCenter2 productState
### Description

security_status.real_time_protection is stored as true for Windows Defender on a device where
Defender is genuinely, completely disabled (superseded by a third-party AV). This is a false
positive in a security-relevant field, not just a display issue.

### Steps to Reproduce

1. Device with Defender fully disabled (real-time protection off) due to a third-party AV
   (CylancePROTECT + Elastic Security in this case) being the active endpoint protection.
2. On-device verification, two independent sources both confirm Defender is off:
   Get-MpComputerStatus -> RealTimeProtectionEnabled: False, AntivirusEnabled: False,
     AMRunningMode: "Not running"
   Get-CimInstance -Namespace root\SecurityCenter2 -ClassName AntiVirusProduct ->
     "Windows Defender" productState: 393472 (0x60010) — real-time-protection byte = 0x00 (off)
3. Query security_status for this device:
   SELECT * FROM security_status WHERE device_id = '<id>';
   -> provider: windows_defender, real_time_protection: t (TRUE)

### Expected Behavior

real_time_protection should reflect the actual, current state of the product — false here, since
Defender's real-time protection is genuinely off.

### Actual Behavior

Stored as true, a false positive. provider_version and definitions_version are both empty/null on
the same row despite definitions_date being populated and real_time_protection showing a
confident "true" — suggesting the collector may be misreading or defaulting this specific field
rather than parsing it correctly from the source data (possibly only checking product
registration/presence rather than the specific real-time-protection bit in Security Center's
productState bitmask).

### Component

Web Dashboard

### Version

Agent 0.105.1, self-hosted

### Relevant Logs

```shell

```

### Additional Context

Real operational risk: any Breeze security-posture scoring, compliance reporting, or alerting
that trusts this field could report a device as adequately AV-protected via Defender when it has
zero active Defender protection — masking a genuine gap if the actual active AV (third-party)
ever has its own issue.

ToddHebebrand:
## Problem

`security_status.real_time_protection` was stored as `true` on a `windows_defender` row for a device where Defender is genuinely, completely disabled (third-party AV active). Both on-device sources agreed Defender was off (`Get-MpComputerStatus` → `RealTimeProtectionEnabled: False`, `AMRunningMode: "Not running"`; SecurityCenter2 `productState` real-time byte `0x00`), so neither decoder was at fault — and prior investigation on the issue ruled out the API merge path, a stale row, and the WSC bitmask parser.

## Root cause

`CollectStatus` (`agent/internal/security/status.go`) used `status.Provider == "other"` as a *"no AV source has identified a provider yet"* sentinel. But `"other"` is also a **legitimate resolved provider**: `providerFromName` returns it for any registered Security Center product whose display name isn't in the mapping table — CylancePROTECT, in this report.

So on this device:

1. The Windows Security Center block correctly picks the active third-party product as primary (first product reporting real-time protection wins) and copies **its** `realTimeProtection: true` and provider `"other"` onto the status.
2. The Defender fallback then runs, sees `Provider == "other"`, reads it as "nothing found", and overwrites the label with `windows_defender` — leaving the third-party product's `true` untouched.

Result: a row that attributes another product's active protection to a disabled Microsoft Defender. That also explains the two other oddities in the report — `provider_version` is empty because the Windows `DefenderStatus` never populates it at all, and `definitions_date` is populated because it came from the WSC primary's timestamp, not from Defender.

## Fix

Track provider resolution in a dedicated `providerIdentified` flag instead of sentinel-comparing the value, and extract the two decisions into pure, table-tested helpers:

- `resolveWSCPrimary(products)` — picks the product that owns the provider label (first reporting real-time protection, else the first) and reports whether it is a genuine attribution (i.e. a registered product).
- `defenderOwnsProvider(providerIdentified, currentRealTime, defenderRealTime)` — Defender may claim the label only when **no** other source identified a provider, or when Defender is the **only** source reporting active real-time protection (in which case it really is what's protecting the device).

The macOS Defender and Linux Elastic Defend branches move to the same flag. On those platforms the sentinel was equivalent (the Security Center block never runs), so their behaviour is unchanged.

On the reporter's device the row will now read `provider: other`, `real_time_protection: true` — accurate, since an unmapped third-party product genuinely is providing real-time protection — instead of falsely crediting Defender.

## Tests

New table-driven `agent/internal/security/status_attribution_test.go`: 6 cases for `resolveWSCPrimary` (including the exact disabled-Defender + active-third-party ordering from this issue) and 6 for `defenderOwnsProvider` covering both directions of the ownership rule.

```
go test -race ./internal/security/...   ok
go vet ./internal/security/...          clean (GOOS=darwin, windows, linux)
GOOS=windows go build ./...             clean
```

## Out of scope

- Persisting `avProducts[]` so this class of report is diagnosable from the stored row — tracked separately as #3641.
- Adding a `cylance` provider value: the provider column is a Postgres enum (`security_provider`), so a new vendor needs a migration + shared/web mapping. Filed as a follow-up thought rather than smuggled into a bug fix.

Closes #3593

### DIFF
--- a/agent/internal/security/status.go
+++ b/agent/internal/security/status.go
@@ -130,6 +130,47 @@ func providerFromName(name string) string {
 	}
 }
 
+// resolveWSCPrimary picks the Windows Security Center product that owns the
+// device's provider label: the first product reporting real-time protection,
+// falling back to the first registered product. The second return value is
+// whether that product is a genuine attribution — a registered product resolves
+// the provider even when its display name maps to "other" (see
+// defenderOwnsProvider).
+func resolveWSCPrimary(products []AVProduct) (AVProduct, bool) {
+	if len(products) == 0 {
+		return AVProduct{}, false
+	}
+
+	primary := products[0]
+	for _, candidate := range products {
+		if candidate.RealTimeProtection {
+			primary = candidate
+			break
+		}
+	}
+
+	return primary, primary.Registered
+}
+
+// defenderOwnsProvider reports whether the Microsoft Defender fallback may claim
+// the `provider` label for this row.
+//
+// It may when no other AV source identified a provider, or when Defender is the
+// only source reporting active real-time protection. It must NOT when another
+// registered product already supplied both the provider and the
+// real-time-protection value: provider "other" is a real attribution for a
+// registered product whose display name we don't map (CylancePROTECT, for
+// example), not a "nothing found" sentinel. Treating it as one relabelled a
+// third-party product's real-time protection as Defender's, reporting
+// real_time_protection=true on a windows_defender row for a device where
+// Defender was fully disabled (#3593).
+func defenderOwnsProvider(providerIdentified, currentRealTime, defenderRealTime bool) bool {
+	if !providerIdentified {
+		return true
+	}
+	return defenderRealTime && !currentRealTime
+}
+
 func latestScanTime(defenderStatus DefenderStatus) string {
 	if defenderStatus.LastFullScan != "" {
 		return defenderStatus.LastFullScan
@@ -989,6 +1030,10 @@ func CollectStatus(cfg *config.Config) (SecurityStatus, error) {
 	status.Provider = "other"
 	status.EncryptionStatus = "unknown"
 
+	// "other" is both the default and a legitimate provider value, so track
+	// separately whether an AV source actually resolved the provider (#3593).
+	providerIdentified := false
+
 	// Windows Security Center AV products (workstations) first.
 	if runtime.GOOS == "windows" {
 		products, wscErr := GetWindowsSecurityCenterProducts()
@@ -999,17 +1044,11 @@ func CollectStatus(cfg *config.Config) (SecurityStatus, error) {
 		} else {
 			status.WindowsSecurityCenterAvailable = true
 			status.AVProducts = products
-			if len(products) > 0 {
-				primary := products[0]
-				for _, candidate := range products {
-					if candidate.RealTimeProtection {
-						primary = candidate
-						break
-					}
-				}
+			if primary, identified := resolveWSCPrimary(products); primary.Provider != "" {
 				status.Provider = primary.Provider
 				status.RealTimeProtection = primary.RealTimeProtection
 				status.DefinitionsUpdatedAt = primary.Timestamp
+				providerIdentified = identified
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer (Go-agent-only diff, so the silent-failure hunter wasn't relevant; no error-handling or fallback paths changed).

**Findings:** 0 blocking. code-reviewer found no issues at or above its confidence threshold and independently re-traced the #3593 scenario end-to-end (`[Defender(RTP=false), Cylance("other", RTP=true)]` → provider stays `other`, Defender does not relabel), plus confirmed the Windows Server / macOS / Linux paths are behaviourally unchanged and `primary.Provider != ""` is equivalent to the old `len(products) > 0` guard (`providerFromName` never returns `""`).

Two informational notes, neither actioned:

- **Untested integration wiring.** The two extracted helpers are well covered, but the `CollectStatus` call sites are `runtime.GOOS`-gated and `GetDefenderStatus` is build-tagged, so the glue itself never executes on the darwin/linux CI runners. Closing that would mean adding an injection seam across all three OS collectors — a refactor, not a bugfix. Flagging rather than smuggling it in.
- **`!providerIdentified` in the darwin branch** is dead-in-practice today (only the Windows block can set the flag). Kept for consistency across the three branches so a future source that resolves a provider doesn't have to re-derive the rule.

**Tests:** new table-driven `agent/internal/security/status_attribution_test.go` (6 + 6 cases). Both reviewers verified the tests fail if the fix is reverted — the `third-party owns an active row` case returns `true` under the old `status.Provider == "other"` sentinel and asserts `false` here. Locally: `go test -race ./internal/security/...` ok, `go vet` clean under GOOS darwin/windows/linux, `GOOS=windows go build ./...` clean. In CI: **Test Agent, Test Agent (race), Test Agent (Windows), Lint Agent (Go)** and all four `Build Agent` targets pass.

**Status:** review-clean, awaiting maintainer merge. Worth noting this changes what a real fleet reports — affected devices flip from `windows_defender`/`true` to their actual provider — so it wants an agent-fleet promote to take effect in the field, and the reporter (#3593) can confirm on the row afterwards.

Issue #3592, opened 2026-08-16
[Software Inventory] Stale/incorrect entries persist even after a fresh scan — confirmed via empty registry query after uninstall, and via agent version never updating post-self-update


cisspUser01:
### Description

software_inventory entries can remain wrong even when last_seen shows a recent, fresh scan —
this isn't simple staleness/lag, since the underlying system state has been directly confirmed
correct via registry query while the inventory entry still reports the old/wrong state.

### Steps to Reproduce

1. Uninstalled "Microsoft Teams Meeting Add-in for Microsoft Office" on LOUISVILLELP6 via
   Breeze's own uninstall action.
2. device_commands confirms real success:
   {"status":"completed","stdout":"{\"action\":\"uninstall\",...,\"success\":true}","exitCode":0}
   completed_at: 2026-08-16 13:52:59
3. software_inventory still shows the entry, with last_seen: 2026-08-16 14:11:48 — nearly 20
   minutes AFTER the uninstall completed, so the scan genuinely ran again since the uninstall.
4. Confirmed directly on the device the registry entry is genuinely gone:
   Get-ItemProperty HKLM:\...\Uninstall\* | Where DisplayName -like "*Teams Meeting*"
   -> empty, no results.
So: real state is correct (uninstalled), scan ran fresh after the change, but the inventory
record still incorrectly shows it installed.

"Breeze Agent" software_inventory entries remain frozen at the version from the original MSI
install (e.g. 0.100.0/0.101.0) despite devices.agent_version correctly showing the current,
self-updated version (0.104.0/0.105.1) for the same device, and despite last_seen on the
inventory row being current/recent.

### Expected Behavior

software_inventory should reflect genuinely current installed-software state after a scan runs,
matching what's actually present on the device.

### Actual Behavior

Confirmed stale/incorrect data persisting across scan cycles in two independent cases, one of
which was directly verified against the device's actual registry state.

### Component

Web Dashboard

### Version

0.105.1

### Relevant Logs

```shell

```

### Additional Context

Given Software Policies (allowlist/blocklist enforcement) reads from this same table, this is a
real operational risk beyond just a display annoyance — a policy evaluation could act on
incorrect inventory data (e.g. treating genuinely-uninstalled software as still present, or
flagging a fully-updated agent as an outdated-version violation).

PR #3663, opened 2026-08-17
fix(api): Breeze Agent inventory version tracks the running agent, not the MSI (#3591)

### LINKED ISSUE #3591: [Software Inventory] "Breeze Agent" entry shows stale version from original MSI install — never updates after agent self-update, diverging from the accurate devices.agent_version
### Description

Software Policies > Software > Inventory lists "Breeze Agent" with its version frozen at
whatever version the original MSI installer registered in Windows' Programs and Features — it
never reflects subsequent self-updates. devices.agent_version (heartbeat-reported) is correct and
current; software_inventory's version for the same device/software is stale, despite last_seen
timestamps confirming the inventory scan itself runs regularly and recently.

### Steps to Reproduce

1. A device originally enrolled at agent v0.100.0 or v0.101.0, since self-updated to a newer
   version (confirmed via devices.agent_version and live heartbeat activity).
2. Query:
   SELECT si.name, si.version, d.hostname, d.agent_version
   FROM software_inventory si JOIN devices d ON si.device_id = d.id
   WHERE si.name ILIKE '%breeze agent%';
3. software_inventory.version shows the ORIGINAL install version (0.100.0/0.101.0), while
   devices.agent_version correctly shows the current version (0.104.0/0.105.1) for the same device.
4. software_inventory.last_seen is current/recent (today), confirming the scan is actively running
   and re-reporting the same stale version, not simply failing to run at all.

### Expected Behavior

The "Breeze Agent" inventory entry's version should reflect the currently-running agent version,
matching devices.agent_version.

### Actual Behavior

Frozen at the version from the original MSI install. Confirmed via uninstall_string
(MsiExec.exe /I{...}) that this entry is sourced from the Windows registry's installed-programs
list, which the agent's self-update mechanism (a direct binary replace, not an MSI reinstall)
never updates.

### Component

Web Dashboard

### Version

0.105.1

### Relevant Logs

```shell

```

### Additional Context

Related to two other recently-reported inventory-freshness issues (device_patches installedAt
null/"Not reported" display bug) — possibly a shared root theme of inventory-collection paths not
staying in sync with the agent's own live-reported state, though this one has a distinct, specific
mechanical cause (MSI registry entry vs. direct binary self-update) worth noting on its own.

Practical impact: this makes "Breeze Agent" itself appear as a widespread outdated-software
violation risk in Software Policies allowlist/blocklist evaluation, for devices that are actually
fully current — worth being aware of if using Software Policies against this specific entry.

ToddHebebrand:
Closes #3591

## Problem

Software Policies > Software > Inventory listed **Breeze Agent** with the version the Windows MSI wrote into its Uninstall registry key at original install. The agent self-updates by replacing its binary in place — it never reinstalls the MSI — so `DisplayVersion` is frozen forever, while `devices.agent_version` (heartbeat-reported from the running binary, authoritative) moves on. `software_inventory.last_seen` stayed current the whole time: the scan runs fine, it just faithfully re-reports a stale registry value.

Practical damage: Software Policy allowlist/blocklist evaluation reads `software_inventory.version`, so the agent itself read as a fleet-wide outdated-software violation on devices that were fully current.

## Fix

Normalize at ingest, in `PUT /agents/:id/software` (`apps/api/src/routes/agents/inventory.ts`): rows whose name is one of the agent's own MSI product names get `devices.agent_version` instead of the reported registry version. New helper `apps/api/src/routes/agents/agentSelfInventory.ts` holds the name set and the resolution rule.

**Why API-side and not in the Go collector** (both options considered, second opinion sought):

- The API already holds the authoritative `devices.agent_version` at that point in the handler.
- The correction applies to every agent **already in the field**. An agent-side fix only reaches devices after a fleet promote, which is currently gated on an unrelated decision.
- One rule in one place; an agent-side copy of the product-name list would be a second rule free to drift.

Details:

- Matching is **exact** on the two `ProductName` literals from `agent/installer/build-msi.ps1` (`Breeze Agent`, `Breeze Agent (Self-Hosted)`), normalized for case/whitespace. A substring match would relabel neighbouring products — Breeze Installer, the watchdog — with the agent's version.
- When a device has no `agent_version` yet (mid-enrollment, before the first heartbeat) the reported version is kept. Normalization can never blank a version out.
- **No migration/backfill.** The software report wipes and reinserts every row for the device, so each device self-corrects on its next scan.

## Not in scope

Sibling issue #3592 (stale inventory entries generally) is deliberately left alone. Worth recording from this pass: this ingest path is a full wipe-and-reinsert per device inside one transaction, and it is the only writer of `software_inventory` from agents — so a stale row surviving a fresh scan cannot come from a missed upsert here. That points #3592 at the report itself (truncated or partial collection agent-side) or at a report that never reached this route, not at the ingest merge logic.

## Tests

- `apps/api/src/routes/agents/agentSelfInventory.test.ts` (new) — name matching incl. near-miss products, the override, the no-agent-version fallback, empty-version normalization.
- `apps/api/src/routes/agents/inventory.test.ts` — new route-level block asserting the rows actually handed to the INSERT. Mutation-checked: reverting the one-line change reds 2 of them.
- `vitest run src/routes/agents/agentSelfInventory.test.ts src/routes/agents/inventory.test.ts --no-file-parallelism` → 29 passed.
- `tsc --noEmit` on `apps/api` → clean.

### DIFF
--- a/apps/api/src/routes/agents/agentSelfInventory.test.ts
+++ b/apps/api/src/routes/agents/agentSelfInventory.test.ts
@@ -41,6 +41,21 @@ describe('resolveInventoryVersion (#3591)', () => {
     expect(resolveInventoryVersion('Breeze Agent', '0.100.0', '   ')).toBe('0.100.0');
   });
 
+  it('ignores the provisioning sentinel agent version', () => {
+    // devices.agent_version is NOT NULL and provisioning seeds '0.0.0' until
+    // the first heartbeat. The agent's software report is dispatched in a
+    // goroutine that is not ordered against that heartbeat, so a report can
+    // arrive while the sentinel is still stored. Copying it in would stamp the
+    // agent's row with a version that sorts below every real one — failing
+    // every min-version policy check, and worse than the bug being fixed.
+    expect(resolveInventoryVersion('Breeze Agent', '0.100.0', '0.0.0')).toBe('0.100.0');
+    expect(resolveInventoryVersion('Breeze Agent (Self-Hosted)', '0.101.0', '0.0.0')).toBe('0.101.0');
+    // ...and it must not leak in when there is nothing to fall back to either.
+    expect(resolveInventoryVersion('Breeze Agent', undefined, '0.0.0')).toBeNull();
+    // A real version that merely starts with a zero is untouched.
+    expect(resolveInventoryVersion('Breeze Agent', '0.100.0', '0.0.1')).toBe('0.0.1');
+  });
+
   it('normalizes empty/whitespace reported versions to null', () => {
     expect(resolveInventoryVersion('Google Chrome', '', '0.105.1')).toBeNull();
     expect(resolveInventoryVersion('Google Chrome', '  ', '0.105.1')).toBeNull();
--- a/apps/api/src/routes/agents/agentSelfInventory.test.ts
+++ b/apps/api/src/routes/agents/agentSelfInventory.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from 'vitest';
+import { isBreezeAgentProductName, resolveInventoryVersion } from './agentSelfInventory';
+
+describe('isBreezeAgentProductName', () => {
+  it('matches both MSI product names, case- and whitespace-insensitively', () => {
+    expect(isBreezeAgentProductName('Breeze Agent')).toBe(true);
+    expect(isBreezeAgentProductName('  breeze agent  ')).toBe(true);
+    expect(isBreezeAgentProductName('Breeze Agent (Self-Hosted)')).toBe(true);
+    expect(isBreezeAgentProductName('BREEZE AGENT (SELF-HOSTED)')).toBe(true);
+  });
+
+  it('does not match on substrings or neighbouring Breeze products', () => {
+    // A substring match would relabel these with the agent's version.
+    expect(isBreezeAgentProductName('Breeze Agent Helper')).toBe(false);
+    expect(isBreezeAgentProductName('Breeze RMM Watchdog')).toBe(false);
+    expect(isBreezeAgentProductName('Breeze Installer')).toBe(false);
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit — `code-reviewer` + `silent-failure-hunter` (both against `git diff origin/main...HEAD`).

**Findings:** 1 raised, addressed in `783a987`; 0 outstanding.

- `code-reviewer`: **0 findings.** Verified the two matched names against the actual `ProductName` literals in `agent/installer/build-msi.ps1:104,106`; confirmed `inventory.ts` is the *only* `insert(softwareInventory)` site in `apps/api/src`, so nothing bypasses the normalization; confirmed the downstream comparators (`softwarePolicyService.compareSoftwareVersions`, `softwareCurrency.compareVersions`) tokenize rather than string-match, so a differently-shaped version is safe.
- `silent-failure-hunter`: **1 finding, valid and consequential.** `devices.agent_version` is NOT NULL, and `routes/devices/provision.ts:272` seeds the literal `'0.0.0'` until the first heartbeat. The agent dispatches its software report from a goroutine that is *not* ordered against that heartbeat, so a report can arrive while the sentinel is still stored — and the original patch would have copied it in. That is strictly worse than the bug being fixed: `'0.0.0'` is chosen precisely to sort below every real version, so the agent's own row would have failed every min-version policy check, with the reported MSI version already discarded. Fixed in `783a987` — the sentinel is now treated as "no live version yet" and the reported version is kept.

**Tests:** `vitest run src/routes/agents/agentSelfInventory.test.ts src/routes/agents/inventory.test.ts --no-file-parallelism` → 30 passed. Mutation-checked non-vacuous: reverting the one-line ingest change reds 2 route tests. `tsc --noEmit` on `apps/api` → clean. CI on `783a987`: **Test API green, Type Check green, 0 failing checks.**

**Status:** review-clean, awaiting maintainer merge. Not merged and #3591 left open per hand-off rules.

PR #3664, opened 2026-08-17
fix(api): append a unique tiebreaker to every offset-paginated ORDER BY (#3462)

### LINKED ISSUE #3462: [API] Offset pagination orders by non-unique columns — page walks can silently skip and duplicate rows
## Summary

The offset-paginated list endpoints order by a **non-unique** column, so a client that walks
pages can receive the same row twice and never see another one. PostgreSQL gives no guarantee
about the relative order of rows with equal sort keys across two independent queries, so
`LIMIT/OFFSET` over `ORDER BY created_at` is only stable while every value is distinct.

This is the *ordering* half of the truncation work already merged in #3446 and #3301. Those
fixed "the client only ever saw page 1" by adding page walks. The walks are now in place, but
the endpoints they walk are not ordered deterministically, so the assembled "complete" list
can be silently wrong.

## Why this is reachable, not theoretical

`created_at`/`updated_at` default to `now()`, which is the **transaction** timestamp, so every
row written in one transaction shares an identical value. A bulk import, a seed, or a
migration produces exactly the ties required.

`main` currently ships four page-walking consumers. Three of them walk an endpoint with no
tiebreaker:

| walker | endpoint ordering | tiebreaker |
|---|---|---|
| `apps/web/src/lib/fetchAllOrganizations.ts:11` | `orgs.ts:1232` `orderBy(organizations.createdAt)`, `.offset()` at 1230 | **no** |
| `apps/web/src/lib/scriptsFetch.ts:72` | `scripts.ts:406` `orderBy(desc(scripts.updatedAt))`, `.offset()` at 408 | **no** |
| `apps/web/src/lib/devicesFetch.ts:194` (`fetchAllNetworkDevices`) | `devices/network.ts:183` `orderBy(desc(discoveredAssets.lastSeenAt))`, `.offset()` at 185 | **no** |
| `apps/web/src/lib/devicesFetch.ts:90` (`fetchAllDevices`) | `/devices` keyset cursor | n/a, keyset |

`scriptsFetch.ts` states its own contract in the file header: *"walks the offset-paginated
`/scripts` endpoint and accumulates every accessible row (#3301)"*. With more than one page of
scripts sharing an `updated_at` (a bundle import writes many rows in one transaction), that
accumulation can drop a script and duplicate another, with no error and no visible symptom.

`orgStore.ts:225` now walks organizations too, so the org switcher is fed by the same shape.

## The fix pattern already exists in this repo

Several endpoints already do this correctly, so this is an inconsistency rather than a new
design decision:

- `tickets/tickets.ts:345-348` — every sort branch ends in `tickets.id`, matching direction.
- `patches/list.ts:105-106` — appends `patches.id`; and `patches/list.ts:87` explicitly notes
  *"patches page (apps/web/.../PatchesPage.tsx) walks pages at `limit=200`"*, i.e. the walker
  and the tiebreaker were considered together.
- `mobile.ts:815` — keyset comparison on the `(triggeredAt, id)` tuple.
- `devices/core.ts` — the ordering helper appends the device id.

## Proposed fix

Append the table's primary key to the `ORDER BY` of offset-paginated list queries, matching
the direction of the leading column, exactly as `tickets.ts` and `patches/list.ts` already do.
It is a one-line change per query and does not alter the intended sort.

Priority order, if this is split up:

1. The three endpoints above that have a shipped page walker. These are actively wrong today.
2. Endpoints whose public contract exposes `page`/`offset` so a caller can request consecutive
   windows independently.

## Scope beyond the three

A repo-wide sweep of `apps/api/src` flagged roughly 90 further offset-paginated queries whose
`ORDER BY` has no unique component. I have **not** individually verified each of those, and
many have no page-walking consumer today, so their practical impact is a user paging manually
seeing an occasional repeat rather than silent data loss. I am listing them as scope rather
than asserting 90 separate defects.

Two cases genuinely need a maintainer's judgement rather than a mechanical fix: `devices/core.ts`
and `software.ts:1878` append a unique base-table id, but if a `LEFT JOIN` can fan out more
than one result row per device, duplicate rows still share a complete ordering key. I could not
establish from the query sites whether that fan-out occurs.

## Verification notes

Everything above about `main` was read from `origin/main` at `49e4bb522`. The four `fetchAll*`
walkers are the complete set matching `export async function fetchAll` under
`apps/web/src/lib`. `created_at` on `organizations` is
`timestamp('created_at').defaultNow().notNull()` with no unique constraint.

Happy to send a PR for the three walker-backed endpoints if that split is useful.

ToddHebebrand:
## What

`LIMIT/OFFSET` over an `ORDER BY` whose sort key is not unique is non-deterministic. PostgreSQL guarantees nothing about the relative order of rows with equal sort keys, and it makes that guarantee **per query** — two requests for consecutive pages are two independent queries and can even pick different plans. A client walking pages therefore receives some rows twice and never sees others, with no error and no visible symptom.

This is reachable, not theoretical: `created_at`/`updated_at` default to `now()`, which is the **transaction** timestamp, so every row written in one transaction — a seed, a bulk import, a script-bundle import, a discovery sweep, a provider sync — shares a byte-identical value.

`main` ships page-walking clients whose whole contract is "accumulate every accessible row": `apps/web/src/lib/fetchAllOrganizations.ts`, `scriptsFetch.ts`, and `fetchAllNetworkDevices` in `devicesFetch.ts`. All three walked endpoints with no tiebreaker, so the "complete" list they assembled could silently be wrong.

## The fix

Append the base table's primary key to the `ORDER BY` of every offset-paginated query, matching the direction of the leading column — the pattern `routes/patches/list.ts` (#3157) and `routes/tickets/tickets.ts` already use. **92 query sites across 71 files.** No change to the intended sort.

The three walker-backed endpoints called out in the issue carry an explanatory comment; the rest are a one-expression mechanical change.

## Cases that were not mechanical

| Site | Why | Tiebreaker used |
|---|---|---|
| `routes/incidents.helpers.ts` | `UNION ALL` over three sources — no single base-table PK | `(source, sourceId)`: `source` is constant per leg, `sourceId` unique within it |
| `services/reliabilityScoring.ts` | `device_reliability` has no `id` column | `device_id`, its primary key |
| `routes/cisHardening.ts` | base id is aliased inside a window-function subquery | `rankedResults.resultId` |
| `routes/software.ts` (deployment results) | ordered by joined `devices.hostname` then `deploymentResults.deviceId`, which is not schema-unique | `deploymentResults.id` |
| `routes/alerts/rules.ts` | the site-scoped branch never calls `.offset()` — it fetches the whole candidate set and slices it in JS, re-running the same unordered query per page request | same bug, fixed alongside |

Already correct and left alone: `routes/patches/list.ts`, `routes/tickets/tickets.ts`, `routes/devices/core.ts` (via `buildOrderBy`), `services/logSearch.ts` (via `resolveSearchOrder`), `routes/mobile.ts`, `routes/updateRings.ts`, `services/fleetFindings/dispatch.ts` (composite PK, `runId` pinned in the WHERE).

The issue asked for a maintainer call on whether `devices/core.ts` and `software.ts` could fan out more than one row per base id. They cannot: `deviceHardware` and `deviceReliability` are schema-enforced 1:1 with `devices` (`device_id` is their PK), and every other join in the sweep is many-to-one on an FK.

## The guard

`apps/api/src/db/offsetPaginationOrdering.test.ts` is a source-level contract test in the required **Test API** job: every statement that calls `.offset()` must order by something unique. Exceptions live in a named allowlist with a stated reason and a scoping substring, so an exemption cannot silently cover a second query in the same file; a companion assertion fails when an allowlist entry goes stale.

It earned its keep immediately — it found `services/reliabilityScoring.ts`, which **both the issue's repo-wide sweep and mine missed**, because that file contains a literal NUL byte (`const KEY_SEP = '\0'`) so `grep` classifies it as binary and skips it without `-a`. That site orders a whole fleet by a 0-100 integer score, where ties are the norm rather than the exception.

The guard was mutation-tested: reverting one fixed site turns it red with a pointed message.

## Verification

- `apps/api` typecheck (`tsc --noEmit`) — clean.
- 190 affected test files single-fork — **3,932 passed, 11 skipped, 0 failed** (every test file co-located with a changed source file, plus `routes/devices/`, `routes/alerts/`, `routes/portal/`).
- New guard test green; verified non-vacuous by reverting a fix.

## Notes for the reviewer

- No schema, migration, or tenancy change — this is `ORDER BY` only.
- Adding a PK to the sort key can change index selection on hot tables. Every affected query already carried a `LIMIT`, and the id column is indexed as a primary key, so the tiebreaker is a suffix on an existing sort rather than a new access path.
- `const KEY_SEP = '\0'` in `reliabilityScoring.ts` is left as-is (it looks deliberate), but it makes that file invisible to plain `grep` — worth knowing for any future repo-wide sweep.

Closes #3462

### DIFF
--- a/apps/api/src/db/offsetPaginationOrdering.test.ts
+++ b/apps/api/src/db/offsetPaginationOrdering.test.ts
@@ -34,6 +34,16 @@ const SRC_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..'
  * `orderBy(...)` of any statement that also calls `.offset(...)`. Anything else
  * — a composite key, an aliased subquery column, an ordering built by a shared
  * helper — must be justified by name in ALLOWED_WITHOUT_LITERAL_ID below.
+ *
+ * What it does NOT catch (know this before trusting a green run): the check is
+ * textual, so it cannot tell the BASE table's `id` from some JOINED table's
+ * `id`. `.from(alerts).innerJoin(devices, …).orderBy(devices.hostname,
+ * devices.id)` satisfies the regex while still being non-deterministic, because
+ * many alerts share one device. Proving otherwise needs the join cardinality,
+ * which is in the schema, not in the call site. So: when you review a new
+ * offset-paginated query, confirm by eye that the tiebreaker is the `.from()`
+ * table's own key. The guard catches the common case — no tiebreaker at all —
+ * not every possible wrong one.
  */
 
 /** Files whose `.offset()` statements order deterministically without a literal
@@ -151,7 +161,20 @@ describe('offset pagination ordering (#3462)', () => {
             (entry) => entry.file === rel && haystack.includes(entry.orderByContains)
           );
 
-        const orderByAt = statement.indexOf('.orderBy(');
+        // Pair the `.offset(` with the TEXTUALLY NEAREST `.orderBy(`, not the
+        // first one in the chunk. A chunk can hold a subquery's `.orderBy()`
+        // ahead of the outer paginated one, and taking the first would check
+        // the wrong clause — possibly masking a real violation by finding an
+        // unrelated `.id`. Nearest rather than nearest-preceding because both
+        // chain orders occur in this codebase (`routes/orgs.ts` writes
+        // `.limit().offset().orderBy()`).
+        const offsetAt = statement.indexOf('.offset(');
+        let orderByAt = -1;
+        for (let at = statement.indexOf('.orderBy('); at !== -1; at = statement.indexOf('.orderBy(', at + 1)) {
+          if (orderByAt === -1 || Math.abs(at - offsetAt) < Math.abs(orderByAt - offsetAt)) {
+            orderByAt = at;
+          }
+        }
         if (orderByAt === -1) {
           // Either genuinely unordered, or the `.offset()` was chained in a
           // later statement than the `.orderBy()` (allowlisted by name).
--- a/apps/api/src/db/offsetPaginationOrdering.test.ts
+++ b/apps/api/src/db/offsetPaginationOrdering.test.ts
@@ -1,0 +1,194 @@
+import { readFileSync, readdirSync, statSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+// apps/api/src/db -> apps/api/src is 1 level up.
+const SRC_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+
+/**
+ * Why this test exists
+ * --------------------
+ * `LIMIT/OFFSET` over an `ORDER BY` whose sort key is not unique is
+ * non-deterministic. PostgreSQL guarantees nothing about the relative order of
+ * rows with equal sort keys, and it makes that guarantee *per query* — two
+ * requests for consecutive pages are two independent queries and can even pick
+ * different plans. So a client walking pages over a tied sort key silently
+ * receives some rows twice and never sees others.
+ *
+ * This is not theoretical here. `created_at`/`updated_at` default to `now()`,
+ * which is the **transaction** timestamp, so every row written in one
+ * transaction — a seed, a bulk import, a patch-scan ingest, a provider sync —
+ * shares a byte-identical value. `main` ships several page-walking clients
+ * (`apps/web/src/lib/fetchAllOrganizations.ts`, `scriptsFetch.ts`,
+ * `devicesFetch.ts`) whose whole contract is "accumulate every accessible row",
+ * so the corruption lands in a list the user believes is complete.
+ *
+ * History: #3157 (patch catalog), then #3462 (repo-wide sweep). The fix is
+ * always one line — append the base table's primary key to the `ORDER BY`,
+ * matching the direction of the leading column — which is exactly why it keeps
+ * getting forgotten on the next endpoint. This guard lives in the required
+ * `test-api` job so the next offset-paginated query cannot ship without one.
+ *
+ * The guard is deliberately strict: it wants to see a literal `.id` inside the
+ * `orderBy(...)` of any statement that also calls `.offset(...)`. Anything else
+ * — a composite key, an aliased subquery column, an ordering built by a shared
+ * helper — must be justified by name in ALLOWED_WITHOUT_LITERAL_ID below.
+ */
+
+/** Files whose `.offset()` statements order deterministically without a literal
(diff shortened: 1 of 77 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer, both against `git diff origin/main...HEAD` with the schema in `apps/api/src/db/schema/` cross-checked.

**Findings:** 4 raised → all 4 addressed in `04c09f655`; 0 outstanding.

- *No behavioural coverage of the actual ORDER BY args* — the repo-wide contract test was textual and the only thing guarding the three walker-backed endpoints. Added one behavioural assertion each to `routes/scripts.test.ts`, `routes/orgs.test.ts` and `routes/devices/network.test.ts`, following the existing `selectPatchListCapture` pattern in `routes/patches/index.test.ts`. Each was proven non-vacuous by removing the `.id` term and confirming a red run.
- *Guard paired `.offset()` with the FIRST `.orderBy()` in its chunk* — a subquery's `.orderBy()` ahead of the outer paginated one would have been checked instead, potentially masking a real violation via an unrelated `.id`. Now pairs with the textually nearest, which also handles `routes/orgs.ts`'s `.limit().offset().orderBy()` chain order.
- *Guard cannot distinguish base-table `id` from a joined table's `id`* — confirmed real. Not fixable textually (it needs join cardinality from the schema), so it is now documented in the test's header instead of left implied. The guard covers "no tiebreaker at all", not "every possible wrong one". Both reviewers independently verified all 92 current sites use the correct base-table key.
- *Overclaiming comment in `incidents.helpers.ts`* — `(source, sourceId)` is a complete key only for the tracked leg. The EDR legs carry the vendor's id, whose unique index is per **integration** (`huntress_incidents_external_idx`), while the legs are scoped by `org_id` alone. Comment corrected to state the real constraint and record the fix if it ever bites (a fourth `integration_id` column on all three legs). Not changed in code: it needs identical rank *and* identical `detectedAt` *and* a colliding vendor id across two same-vendor integrations in one org.

No blocking findings on the 92 ORDER BY sites themselves. Both reviewers confirmed every appended column is the base table's real primary key, every direction matches its leading column, and no touched join can fan out and defeat the tiebreaker.

**Tests:** 190 affected test files single-fork — 3,932 passed, 11 skipped, 0 failed. Contract test green and mutation-verified (reverting one fixed site turns it red). `tsc --noEmit` on `apps/api` clean. CI: Type Check and Lint green; Test API and Integration Tests still running at time of writing.

**Status:** review-clean, awaiting maintainer merge. Not merged and issue #3462 left open per the handoff rule.

Issue #3493, opened 2026-08-12
Partner-wide Monitoring policies don't reach devices; deleted policy leaves stale watches running


ToddHebebrand:
Reported by **odobenusrosmarus_** in #bugs (Discord, message 1537162751183421630).

**Report (PII-redacted):**

I'm noticing an issue with Partner Wide Configuration Policies. Creating a Partner Wide Config Policy with just one Service & Monitoring item, went into assignments and assigned it to an organization, but the devices within the organization never got updated.
Verified that inside of the device it can see it was assigned the Partner Config Policy, but the monitoring options do not update.

This does not happen if I create a specific organization policy. When using a Organization Policy, it appears to apply successfully and the monitoring updates with the service item as expected. Upon deleting the Organization Policy, however, the service being monitored was not removed and remains being watched.

Server version 0.104.0
Device version 0.105.0
Device OS Windows 11 Pro
Server OS Debian 13

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

## Triage analysis

Two distinct real defects in the Monitoring config-policy feature, both verified in code:

1) Partner-wide policies never reach devices for the 'monitoring' feature. resolveDeviceMonitoringSettings() in apps/api/src/routes/agents/helpers.ts (lines 1928-1988) queries configPolicyAssignments joined to configurationPolicies with the filter `eq(configurationPolicies.orgId, device.orgId)` (line 1986). Partner-wide policies are stored with `orgId = NULL` and `partnerId = <partner>`, so this exact-match filter excludes them entirely — the innerJoin returns zero rows for any partner-wide policy no matter what assignment exists. This is the same class of bug tracked as #1724 and already fixed in the general effective-config resolver (apps/api/src/services/configurationPolicy.ts, lines 1763-1770), which explicitly OR's in `orgId IS NULL AND partnerId = org.partnerId` — but that fix was never applied to the monitoring-specific resolver. That explains why the Effective Configuration tab on the device shows the partner policy as assigned (it uses the fixed resolver) while the agent's actual monitor watches never update (it uses the unfixed one).

[...]

Issue #2949, opened 2026-07-30
Agent monitoring-config delivery gaps: cannot clear watches, partner-wide policies excluded, severity not delivered


ToddHebebrand:
Found while consolidating alert-rule ownership (PR #2946); all pre-existing in `apps/api/src/routes/agents/helpers.ts` (`resolveDeviceMonitoringSettings` / `buildMonitoringConfigUpdate`):

1. Zero enabled watches returns `null`, so the heartbeat omits `monitoring_settings` entirely and previously-delivered watches can never be cleared from an agent — even though the agent supports an empty config (`agent/internal/monitoring/monitor.go`).
2. The resolution predicate is org-only, so **partner-wide monitoring policies are never delivered** (Partner-Wide First violation, epic #2135), and assignment `role_filter`/`os_filter` are ignored on this path.
3. `alertSeverity` is not delivered to the agent, and `alertOnStop` / consecutive-failure settings don't currently drive alert creation — audit how watch reports become `service_stopped`/`process_stopped`/`process_cpu_high`/`process_memory_high` alerts and whether configured severity flows through.

Item 2 is the priority — it silently no-ops a documented MSP workflow.

ToddHebebrand:
**Item 1 is only half fixed on `main`. The deleted/unassigned-policy branch still can't clear watches.** Recording this here because #3493 (a Discord report) is being closed in favour of this issue for its stale-watch half.

Verified on `origin/main` (`d1cbf4fe27`):
- **Fixed by #4839 (v0.110.0):** a policy that resolves with zero enabled watches returns `{ watches: [] }`, which clears the agent.
- **Still open:**
  - When **no policy resolves** (the policy was deleted, unassigned, or deactivated), `resolvePolicyMonitoringSettings` returns `null` at `apps/api/src/routes/agents/helpers.ts:2417` (`eligibleRows.length === 0`).
  - `resolveDeviceMonitoringSettings` then returns `null` at `helpers.ts:2325` whenever there are also no monitor-derived watches.
  - `routes/agents/heartbeat.ts:2166-2167` only attaches `monitoring_settings` when it's truthy, so the key is omitted. The agent (`agent/internal/heartbeat/heartbeat.go:2960-2964`, `hasMon` check) treats an absent key as "no change" and keeps its previous watches indefinitely.
  - This is exactly #3493's repro: create an org policy with a service watch, then delete the policy, and the service stays watched.
[...]

Issue #2930, opened 2026-07-29
Config-policy resolvers ignore partner-owned policies (event_log, monitoring, pam, patch_source)


ToddHebebrand:
### Summary

While fixing the `helper` config-policy resolver to honor partner-owned (partner-wide) policies (PR #2911, Task 9 of the RDS per-session-helpers work), I found that **four sibling resolvers in the same file carry the identical org-only ownership bug**. Each joins policies on `configurationPolicies.orgId = device.orgId`, so a partner-owned policy (`org_id NULL`, `partner_id` set) silently never resolves for any device — the policy is authorable in the UI but never reaches an agent.

This violates the partner-wide-first config principle (epic #2135).

### Affected resolvers

All in `apps/api/src/routes/agents/helpers.ts` (line numbers approximate, drift over time):

- `event_log` settings resolver (~L1797)
- `monitoring` settings resolver (~L1986)
- `pam` settings resolver (~L2626)
- `patch_source` settings resolver (~L2797)

### Fix pattern (already applied to the `helper` resolver in PR #2911)

Replace the org-only `eq(configurationPolicies.orgId, device.orgId)` filter with the dual-axis condition used by `resolveEffectiveConfigWithExecutor` (`apps/api/src/services/configurationPolicy.ts`, ~L1794):

```ts
org?.partnerId
  ? sql`(${configurationPolicies.orgId} = ${device.orgId} OR (${configurationPolicies.orgId} IS NULL AND ${configurationPolicies.partnerId} = ${org.partnerId}))`
  : eq(configurationPolicies.orgId, device.orgId)
```

Two things to verify per resolver, both learned in Task 9:

1. **DB context.** The read must run where partner-owned rows are visible under RLS. If the resolver currently runs inside the request's org-scoped `withDbAccessContext` (`accessiblePartnerIds: []`), partner-wide rows are RLS-hidden even after the app-layer fix — move the call to `withSystemDbAccessContext` after the org transaction closes (the #1105 pattern already used for `policyProbeConfig`/`onedriveSettings` in `heartbeat.ts`).
2. **Disabled-vs-absent.** If any of these resolvers also treats "no policy / disabled" the way the helper resolver did (falling through to a legacy org-level fallback on `!enabled`), apply the same null-vs-disabled distinction so an explicitly-disabled policy wins.

### Not doing this in PR #2911

[...]

Issue #1724, opened 2026-06-21
Config Policies: 'Partner' scope errors — should mean 'Partner-Wide / All Orgs' with an optional device-type filter


ToddHebebrand:
## Summary

During a UI walkthrough of **Configuration Policies**, selecting the **"Partner"** assignment scope produces an error. The "Partner" target is currently implemented as *"pick a specific partner"* — the assignment UI tries to **list partners** to populate a target picker, which a normal MSP user cannot do under tenant isolation (a partner can't see other partners).

What's wanted instead: **"Partner"** should mean **"Partner-Wide / All Orgs"** — apply the policy across *all of the current partner's organizations* — without asking the user to pick a partner at all. Additionally, partner-wide (and ideally any-level) assignments should support an **optional narrowing filter** (e.g. by device type / role / OS / group) so a partner-wide policy can target, say, "all servers across all my orgs" rather than literally every device.

Good news from the code dig: the data model is already mostly there. The assignment table already carries `role_filter` / `os_filter` columns and the scope hierarchy already resolves a partner level against the device's `org.partner_id`. The gaps are (a) the UI's partner-target picker, and (b) those filters are persisted but **never applied during resolution**.

## Current behavior

**The "Partner" scope errors because the UI tries to list partners and gets a 403.**

- The assignment UI maps each level to a "list targets" endpoint; for partner it fetches `/orgs/partners?limit=200`:
  - `apps/web/src/components/configurationPolicies/AssignmentsTab.tsx:91-97` (`endpointMap`, `partner: '/orgs/partners?limit=200'`), called from `fetchTargetOptions` at `AssignmentsTab.tsx:88-125`.
- That endpoint requires **`system` scope**:
  - `apps/api/src/routes/orgs.ts:227` — `orgRoutes.get('/partners', requireScope('system'), requireOrgRead, ...)`
- A normal MSP user has `partner` or `organization` scope, not `system`, so the call returns **403**. The catch block (`AssignmentsTab.tsx:102-104,119-122`) surfaces "Failed to load targets" and leaves the target dropdown empty — so a partner assignment can't be created.
- The validator also requires a real partner UUID for partner level, with no relaxation:
  - `packages/shared/src/validators/index.ts:627-633` — `assignPolicySchema`, `targetId: z.string().guid()` is required for **all** levels including `partner`.
[...]

PR #3665, opened 2026-08-17
fix(api): partner-wide config policies invisible to org-scoped effective-config resolution (#3493)

ToddHebebrand:
Part of #3493 — partner-wide delivery half only; the stale-watch half is tracked in #2949, so the issue stays open until that lands.

## The bug

A partner-wide ("All orgs") Configuration Policy assigned to an organization never showed up as a device's effective configuration. The reporter hit it with Monitoring: the service watch was authored, assigned, and then simply never applied — while the same policy created org-scoped worked fine.

## Root cause

`resolveEffectiveConfigWithExecutor` (`services/configurationPolicy.ts`) has carried the correct **app-layer** dual-axis ownership predicate since #1724:

```ts
org?.partnerId
  ? sql`(${configurationPolicies.orgId} = ${device.orgId} OR (${configurationPolicies.orgId} IS NULL AND ${configurationPolicies.partnerId} = ${org.partnerId}))`
  : eq(configurationPolicies.orgId, device.orgId)
```

What it lacked is the **other half** of the partner-wide contract that `CLAUDE.md` spells out: **RLS is stricter than the app layer.** Partner-owned rows (`org_id NULL`, `partner_id` set) are gated by `breeze_has_partner_access`, and every org-scoped caller — which is exactly who opens a device's Effective Configuration tab — carries `accessiblePartnerIds: []`. Postgres dropped every partner-owned policy row **silently**: no error, just zero rows. The predicate read as correct while the join resolved nothing.

The agent-side resolvers hit the identical trap and fixed it in #2930 with `withPartnerWideVisibility`. That fix was never applied to this resolver.

## Why not just reuse `withPartnerWideVisibility`

It escapes by opening a fresh system context — a new transaction on a **second pooled connection**. `resolveEffectiveConfigWithExecutor` also runs on an open `tx` (the preview/diff path, which holds *uncommitted* proposed assignments). On that path a system escape both:

- cannot see the transaction's uncommitted rows, and
- does not even retarget a query already bound to `tx` — it becomes a no-op that merely double-holds a connection (the #1105 shape).

`current` and `proposed` would then disagree about whether a partner-wide policy exists — a worse bug than the original.

## The fix

New `withDevicePartnerPolicyVisibility` (`services/configPolicyOwnership.ts`) — the **same-connection** variant. It appends **exactly one** partner id (the device org's own, read under the caller's own RLS context) to `breeze.accessible_partner_ids` for the duration of the policy join, then restores the previous value.

- `breeze.scope` is never touched, so this is **strictly narrower than a system escape**: every other RLS axis keeps evaluating unchanged, and the only rows that become legible belong to that one partner.
- Steps 1-3 (device / org / group loads) stay in the caller's context on purpose — they are the tenancy boundary.
- The widening is skipped when it cannot help or cannot work: no partner, no ambient DB context (no held transaction ⇒ `SET LOCAL` would silently land on an arbitrary pooled connection), already system-scoped, or partner already allowlisted.

This also fixes `resolveRemoteAccessForDevice`, which called `resolveEffectiveConfig` with no escape of its own.

### Partner-wide reader sweep

Per the `CLAUDE.md` playbook step 7, swept the other readers for the same feature. `manage_service_monitors` (`services/aiToolsFleet.ts`) filtered monitoring policies with a bare `orgWhere(auth, configurationPolicies.orgId)`, so it omitted every partner-owned monitoring policy — including from the partner-scoped techs who authored them. Switched to `policyAccessCondition`.

## Scope note

Per the maintainer's triage comment on #3493, this PR covers **only** the partner-wide delivery half. The "deleted policy leaves stale watches running" half (the server structurally cannot send "you now have zero watches") is tracked separately in **#2949** and is untouched here.

## Testing

New real-Postgres integration suite `configurationPolicyOrgScopedPartnerWide.integration.test.ts`, driving resolution under a genuine **org-scoped** `withDbAccessContext` — the shape no existing suite covered:

- `configurationPolicyPartnerResolution.integration.test.ts` resolves under a **system-scope** AuthContext inside `withSystemDbAccessContext`, so `breeze_has_partner_access` short-circuits true and the RLS half is never exercised.
- `agentPolicyResolversPartnerWide.integration.test.ts` covers the four **agent-side** resolvers and only ever assigns at level `partner` — never the org-level assignment the reporter actually used.
- Every mocked unit suite is blind to this by construction: no RLS runs.

Cases: the reporter's org-level assignment; partner-level **fan-out across two orgs** of the same partner, each resolved under its own org context; the widening is exactly one partner id and never `*`; the GUC is restored after the join; and preview `current`/`proposed` stay consistent across the transaction boundary.

**Four of the five fail without this change** (verified by stashing the resolver edit).

Unit tests pin the widening value, append-not-replace, restore-on-throw, restore-failure-must-not-mask-the-real-error, and all four skip branches.

```
integration: configurationPolicyOrgScopedPartnerWide + configurationPolicyPartnerResolution
             + agentPolicyResolversPartnerWide          -> 18 passed
integration: configurationPoliciesPartnerRls            -> 13 passed
unit:        configPolicyOwnership + configurationPolicy -> 110 passed
unit:        aiToolsFleet                                -> 72 passed
unit:        routes/configurationPolicies + routes/remote + remoteAccessPolicy -> 226 passed
tsc --noEmit (apps/api)                                  -> clean
```

### DIFF
--- a/apps/api/src/__tests__/integration/configurationPolicyOrgScopedPartnerWide.integration.test.ts
+++ b/apps/api/src/__tests__/integration/configurationPolicyOrgScopedPartnerWide.integration.test.ts
@@ -1,0 +1,334 @@
+/**
+ * Effective-config resolution honours partner-wide policies for an
+ * ORG-SCOPED caller (#3493).
+ *
+ * `resolveEffectiveConfig` has carried the correct dual-axis APP-layer
+ * ownership predicate since #1724 (`org_id = device.org_id OR (org_id IS NULL
+ * AND partner_id = org.partner_id)`). What it did not have was the second half
+ * of the contract: RLS is STRICTER than the app layer. A partner-owned row is
+ * gated by `breeze_has_partner_access(partner_id)`, and EVERY org-scoped
+ * caller — which is exactly who opens a device's Effective Configuration tab —
+ * carries `accessiblePartnerIds: []`. Under that context Postgres silently
+ * drops every partner-owned policy row: no error, just zero rows. The
+ * predicate reads as correct while the join resolves nothing, and a
+ * partner-wide Monitoring policy appears never to reach the device.
+ *
+ * The fix is `withDevicePartnerPolicyVisibility` (services/configPolicyOwnership.ts)
+ * around ONLY the policy join. It is the SAME-CONNECTION variant of
+ * `withPartnerWideVisibility` on purpose: the preview/diff path passes an open
+ * transaction holding uncommitted proposed assignments, and a system-context
+ * escape would run on a second connection that cannot see them — so `current`
+ * and `proposed` would disagree about whether a partner-wide policy exists.
+ *
+ * Why this file has to exist alongside the two suites that look like they
+ * already cover it:
+ *   - `configurationPolicyPartnerResolution.integration.test.ts` resolves under
+ *     a SYSTEM-scope AuthContext AND inside `withSystemDbAccessContext`, so
+ *     `breeze_has_partner_access` short-circuits true and the RLS half is never
+ *     exercised at all.
+ *   - `agentPolicyResolversPartnerWide.integration.test.ts` covers the four
+ *     AGENT-side resolvers in routes/agents/helpers.ts, not this one, and only
+ *     ever assigns at level `partner` — never the org-level assignment the
+ *     reporter of #3493 actually used.
+ * Every mocked unit suite is blind to this by construction: no RLS runs.
+ */
+import './setup';
+import { afterEach, describe, expect, it } from 'vitest';
+import { randomUUID } from 'crypto';
+import { eq, inArray, sql } from 'drizzle-orm';
+import { db, withDbAccessContext, type DbAccessContext } from '../../db';
+import {
+  configurationPolicies,
+  configPolicyFeatureLinks,
+  configPolicyAssignments,
+  configPolicyMonitoringSettings,
+  configPolicyMonitoringWatches,
+  devices,
+} from '../../db/schema';
+import { resolveEffectiveConfig, previewEffectiveConfig } from '../../services/configurationPolicy';
+import { withDevicePartnerPolicyVisibility } from '../../services/configPolicyOwnership';
+import type { AuthContext } from '../../middleware/auth';
+import { createPartner, createOrganization, createSite, createUser } from './db-utils';
+
+const SYSTEM_CTX: DbAccessContext = {
+  scope: 'system',
+  orgId: null,
+  accessibleOrgIds: null,
+  accessiblePartnerIds: null,
+  userId: null,
+};
+
+/**
+ * The shape that reproduces the bug: an organization-scoped DB context.
+ * `accessiblePartnerIds: []` is not incidental — it is what every org token,
+ * portal session, and agent context carries, and it is what makes
+ * `breeze_has_partner_access` false.
+ */
+function orgDbContext(orgId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: [],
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter`, both against `git diff origin/main...HEAD`, with a design quorum (Codex, read-only, `xhigh`) on the one consequential design call.

**Findings:** 4 raised → 4 addressed in `c796a4b80`; 0 outstanding.

- **Missed partner-wide sweep** (`code-reviewer`) — `setup_auto_approval` in `aiToolsFleet.ts` carried the identical bare org-equality read of `configuration_policies`, one call away from the site this PR fixed. Swept it, **and** added the `canManagePartnerWidePolicies` gate the HTTP feature-link route enforces — making a partner-wide policy *reachable* must not make it *writable*. (That action early-returns as disabled, so both are defense-in-depth; called out in a comment, matching the disabled `manage_alert_rules` branch.)
- **Silent failure** (`silent-failure-hunter`) — the restore-path catch in `withDevicePartnerPolicyVisibility` swallowed with no telemetry. Its safety argument ("only an already-aborted transaction reaches here") is an assumption the code doesn't enforce, and the failure it assumes away — a widened partner allowlist outliving the resolve — is security-relevant. Now logs + `captureException`.
- **Test wouldn't catch a wiring regression** (`silent-failure-hunter`) — one integration case called the helper directly and would still pass if the wiring in `resolveEffectiveConfigWithExecutor` were reverted. Removed; the helper's widen-by-exactly-one-partner contract is already pinned by `configPolicyOwnership.test.ts`, and all four remaining integration cases now go through `resolveEffectiveConfig`/`previewEffectiveConfig`, so every one regresses on a wiring revert.
- **Coverage gap** (`code-reviewer`) — the `manage_service_monitors` fix had no test. Added one; verified it fails when the reader is reverted to `orgWhere(auth, configurationPolicies.orgId)`.

**Design quorum:** my first plan was to reuse `withPartnerWideVisibility` on the non-transactional path only. Codex rejected that (it ships a knowingly broken diff) *and* rejected system-scoping the preview transaction (that boundary needs far more than the ID checks I'd have added, and turns every future preview change into an RLS-bypass footgun). Its counter-proposal — widen the caller's own connection by exactly one derived partner id — is what shipped. Both reviewers independently traced the resulting primitive for tenant-isolation and found nothing.

**Also caught by CI, not by review:** `Test API` went red on `configurationPolicy.baseline.test.ts` — its hand-rolled `vi.mock('../db')` factory omits `getCurrentDbAccessContext`, which the new helper reads. That is the *designed* loud failure (the helper's named `db` imports exist precisely so a lying mock throws instead of silently skipping the visibility logic). Fixed in `e28ab4afd`, plus a sweep of every other suite with a hand-rolled `../db` mock touching this resolver — none reach the path.

**Tests:**
- Full `apps/api` unit suite, single-fork: `src/services` 9897 passed / 651 files; `src/routes` + `src/jobs` 9810 passed / 599 files.
- Integration (real Postgres): `configurationPolicyOrgScopedPartnerWide` + `configurationPolicyPartnerResolution` + `agentPolicyResolversPartnerWide` → 17 passed; `configurationPoliciesPartnerRls` → 13 passed. **4 of the 5 new integration cases fail without the fix** (verified by stashing the resolver edit).
- `eslint src` clean; `tsc --noEmit` clean. All 54 PR checks green, integration shards included.

**One thing to decide before closing #3493:** this PR is scoped to the **partner-wide delivery** half only, per your triage comment. The "deleted policy leaves stale watches running" half — the server structurally cannot say "you now have zero watches" — is **#2949** and is untouched here. So #3493 shouldn't be closed on the reporter's *full* original report without #2949 also landing.

Issue #2649, opened 2026-07-20
i18n: #2340 extraction replaced English copy with humanized key names across all five locales


ToddHebebrand:
### Summary

The i18n extraction in #2340 replaced user-facing English copy with **machine-humanized versions of the key names** in an unknown number of places. The four non-English locales were then translated from those placeholders, so the damage is mirrored across all five locales.

This was found while fixing #2614, where a user hit it directly: the "viewer didn't open" card rendered as a heading reading `Title` above body text reading `Viewer Description`.

### Confirmed and already fixed

PR #2645 restored **28 keys** across `ConnectDesktopButton` and `ConnectVncButton`, recovered from the pre-i18n components at `944bb8d18^`. Examples of what was live in production:

| key | shipped value | actual original |
|---|---|---|
| `connectDesktopButton.fallback.title` | `Title` | `Viewer didn't open?` |
| `connectDesktopButton.fallback.viewerDescription` | `Viewer Description` | `If the viewer opened, you can dismiss this. Otherwise, download it below.` |
| `connectDesktopButton.errors.deviceChanged` | `Failed to device changed` | `Remote launch failed: device record changed. Please refresh and try again.` |
| `connectDesktopButton.errors.securityCheckToast` | `Failed to security check toast` | `Remote launch unavailable: security check failed. Please contact your administrator.` |
| `connectVncButton.policyDisabledNamed` | `Policy Disabled Named` | `VNC relay is disabled by policy "{{name}}"` |
| `connectVncButton.vncDesktop` | `Vnc Desktop` | `VNC Desktop` |

The translations were literal renderings of the placeholder — `fr-FR` had `Description du spectateur` ("description of the spectator") for `viewerDescription`.

### Why I am not attaching a number to the rest

I tried three progressively stricter detectors against `en/*.json` and **all of them produce false positives**, because the extractor derived key names *from* the original English — so a humanized key frequently round-trips to something a human would plausibly have written.

- Naive "value equals humanized key, 2+ words": **3,869 hits**, but that includes legitimate copy like `noReasonGiven` → `No reason given`.
- "`Failed to <humanized key>`": **185 hits**, but `errors.loadContracts` → `Failed to load contracts` is exactly right. Only the ones whose key is a *noun* phrase are broken: `errors.handlerNotConfigured` → `Failed to handler not configured`.
[...]

PR #2340, opened 2026-07-11
feat(web): internationalize console for Brazilian Portuguese

ToddHebebrand:
## Summary

- add the web-console i18n runtime, locale preference primitives, per-user persistence, language selection, and partner-default locale precedence
- extract the console into domain-scoped English and Brazilian Portuguese namespaces with lazy non-English loading
- route dates, times, numbers, currency, relative time, metadata, navigation, CSV copy, and printable-report copy through locale-aware helpers
- add parity, translation-key usage, and extraction-quality enforcement tests
- fix live-only pt-BR residues and defer stored-locale switching until Astro islands finish initial hydration
- document Phase 2 completion plus Phase 3/4 follow-on plans

## Why

Breeze needs a maintainable localization foundation and a complete first non-English web-console locale. This work establishes English as the eager source locale, Brazilian Portuguese as a lazy-loaded locale, and contributor safeguards that prevent namespace drift or obvious untranslated UI from silently returning.

## User impact

Users can select Brazilian Portuguese and retain that preference across reloads and navigation. Partner defaults apply when a user has not made an explicit choice. The main web-console surfaces, interactive states, browser metadata, locale-sensitive formatting, exports, and printable content now render through the shared i18n system.

## Implementation notes

- locale preference precedence is explicit user choice → partner default → English
- structural enum/state values remain stable; only presentation strings are translated
- namespaces are split by domain and pt-BR resources are loaded lazily
- stored non-English locale application waits for Astro's SSR islands to hydrate, preventing English-SSR/pt-BR-client mismatches
- Phase 2 completed all 34 executable checklist items
- native-speaker linguistic QA is tracked separately in #2338

## Validation

- [x] Full web suite after merging current `main`: 430 files / 3,335 tests
- [x] API org/user route suites: 2 files / 214 tests
- [x] Astro check after merging current `main`: 1,346 files, 0 errors
- [x] Production web build
- [x] Translation namespace parity and static key-usage enforcement
- [x] Extraction-quality and visible-string residue sweeps
- [x] Isolated seeded OrbStack pt-BR walkthrough:
  - Dashboard
  - Profile
  - Devices
  - Alerts
  - Tickets
  - Remote Access
  - Scripts
  - Patches
  - Vulnerabilities
  - Reports
- [x] Fresh-tab browser check with zero hydration/runtime console errors

## Follow-up

- #2338 — native-speaker review of the machine-drafted Brazilian Portuguese copy before broad production rollout

PR #3666, opened 2026-08-17
fix(web,i18n): stop shipping API route paths as translatable locale values (#3426)

### LINKED ISSUE #3426: [Web][i18n] API route paths live in the locale files as translatable values — one translator edit silently breaks a feature in that locale
## Description

The #2340 extraction codemod swept **API route paths and one filesystem path** into the locale files as if they were display copy. They now sit in all seven locales as translatable values:

| Key | Value |
|---|---|
| `devices.json` → `deviceGroupsPage.deviceGroups` | `/device-groups` |
| `policies.json` → `configurationPolicies.configPolicyCreatePage.configurationPolicies2` | `/configuration-policies` |
| `policies.json` → `configurationPolicies.configPolicyDetailPage.configurationPolicies2` | `/configuration-policies` |
| `policies.json` → `configurationPolicies.featureTabs.automationTab.scriptsLimit200` | `/scripts?limit=200` |
| `policies.json` → `configurationPolicies.featureTabs.automationTab.alertsChannelsLimit200` | `/alerts/channels?limit=200` |
| `policies.json` → `configurationPolicies.featureTabs.automationTab.softwareCatalogLimit100` | `/software/catalog?limit=100` |
| `policies.json` → `software.builtinPackageDetail.integrations` | `/integrations` |
| `policies.json` → `configurationPolicies.featureTabs.complianceTab.etcSshSshdConfig` | `/etc/ssh/sshd_config` |

Enumerate them with:

```bash
python3 - <<'PY'
import json,glob,os
for f in sorted(glob.glob('apps/web/src/locales/en/*.json')):
    d=json.load(open(f))
    def walk(o,p=''):
        if isinstance(o,dict):
            for k,v in o.items(): walk(v,f"{p}.{k}" if p else k)
        elif isinstance(o,str) and o.startswith('/') and len(o)>4 and ' ' not in o:
            print(os.path.basename(f), p, repr(o))
    walk(d)
PY
```

## Why this is a bug and not just untidy

Nothing is broken **today**, because every locale happens to carry the verbatim English value. It works by luck. The failure mode is that a translator — human or machine — localizes one of these entries, and the feature breaks **only in that locale**, at runtime, with no test or type error to catch it. `/etc/ssh/sshd_config` is especially inviting to "translate".

`deviceGroupsPage.deviceGroups` was the live one: `DeviceGroupsPage.tsx` built its group-create `POST` URL from `t("deviceGroupsPage.deviceGroups")`. **That call site is already being fixed in #3423**, which inlines the literal and removes that key from all seven locales.

## Remaining work

The other seven entries have **no call sites** — the codemod left them behind as dead keys after the URLs were inlined back. So this is cleanup plus one judgement call:

1. Delete the six dead route/query-string keys from all seven locales (must be all seven, or a locale parity gate reds main — see #3220-class breakage).
2. `complianceTab.etcSshSshdConfig` **does** have a live call site (`ComplianceTab.tsx:948`) where it renders as display text — an example path in help copy. It should be a literal in the component, not a translatable string.
3. Consider a guard: a small test asserting no `en` locale value matches `^/[a-z0-9-]+` (excluding legitimately-translatable unit suffixes like `perMonth: "/mo"`, `perYear: "/yr"`). That makes this class mechanical instead of a future grep. This is the same shape of defect as #2649 and would have caught both.

## Affected files

- `apps/web/src/locales/*/policies.json` (7 locales)
- `apps/web/src/locales/*/devices.json` (7 locales — handled by #3423)
- `apps/web/src/components/configurationPolicies/featureTabs/ComplianceTab.tsx:948`

Related: #2649 (same codemod, different symptom — humanized key names replacing English copy), #3423 (fixes the one live route key), #3424 (same codemod, different symptom — a sentence split into four unspaced JSX expressions).

ToddHebebrand:
Closes #3426

## Problem

The #2340 extraction codemod swept API route paths, query-string endpoints, and one example filesystem path into the locale catalogs as translatable values. Nothing was broken **today** only because every locale happened to carry the verbatim English value — it worked by luck. A translator (human or machine) localizing one of these entries breaks that feature **in that locale only**, at runtime, with no type or test error to catch it.

The one live route key (`deviceGroupsPage.deviceGroups`) was already fixed in #3423. This PR handles the rest.

## Changes

**1. Deleted six dead keys** from all eight locales' `policies.json` (they have no call sites — the URLs were already inlined back into the components, leaving the keys orphaned):

- `configurationPolicies.configPolicyCreatePage.configurationPolicies2` → `/configuration-policies`
- `configurationPolicies.configPolicyDetailPage.configurationPolicies2` → `/configuration-policies`
- `configurationPolicies.featureTabs.automationTab.scriptsLimit200` → `/scripts?limit=200`
- `configurationPolicies.featureTabs.automationTab.alertsChannelsLimit200` → `/alerts/channels?limit=200`
- `configurationPolicies.featureTabs.automationTab.softwareCatalogLimit100` → `/software/catalog?limit=100`
- `software.builtinPackageDetail.integrations` → `/integrations`

**2. Inlined the one live path key.** `complianceTab.etcSshSshdConfig` renders as an input placeholder in `ComplianceTab.tsx` — an example filesystem path in help copy, not display copy. It is now the literal `"/etc/ssh/sshd_config"` with a comment saying why, and the key is deleted from all eight locales.

All deletions land in all eight locales (`en`, `de-DE`, `es-419`, `fr-CA`, `fr-FR`, `it-IT`, `pt-BR`, `tr-TR`) so the parity gate stays green — the #3220-class breakage.

**3. Added a guard** in `localeParity.test.ts` so this class is mechanical rather than a future grep: a leaf whose *entire* value is a bare slash-prefixed token (no whitespace) is rejected, **in every locale**.

Two design notes on the guard:

- **Exemptions are keyed by `namespace:key`, not by value.** The legitimate slash-prefixed copy is the per-unit price suffix and one separator glyph — and those values are themselves localized (`/mo` → `/mois` → `/ay` → `/Jahr`). A value allowlist like `new Set(['/mo','/yr'])` would only ever hold for `en` and reds every other locale. Five keys are exempt, enumerated explicitly so a new route path cannot slip in under a permissive regex.
- **Prose that merely mentions a path is skipped** (anything containing whitespace: `"/home and /etc with caches excluded."`, `"/tmp, /proc"`, the custom-CSS placeholders). Those are already pinned across locales by the existing `unix-path` protected-literal contract in the same file. The real hole this guard closes is bare route paths like `/configuration-policies`, which no existing protected-token pattern matches.

## Out of scope

`featureTabs.backupTab.varBackupsBreezeOrNasBackups`, `sensitiveDataTab.homeVarData`, and `sensitiveDataTab.tmpProc` are comma/`or`-joined *lists* of example paths mixed with connective words. They are already protected across locales by the existing `unix-path` token contract, so they carry no silent-breakage risk and left alone.

## The predicted failure already happened

While verifying the guard against the pre-fix catalogs, it turned out `tr-TR` on `main` had **already localized two of these route paths**:

| Key | `en` | `tr-TR` on `main` |
|---|---|---|
| `configPolicyCreatePage.configurationPolicies2` | `/configuration-policies` | `/yapılandırma-politikaları` |
| `software.builtinPackageDetail.integrations` | `/integrations` | `/entegrasyonlar` |

So this is not a hypothetical. The exact failure mode the issue predicts has already occurred in the catalog — it stayed harmless only because these particular keys were dead by then. Had the call sites still existed, Turkish users would have hit 404s with nothing red in CI. That is the case for the guard, not just the cleanup.

## Testing

- `vitest run src/lib/i18n/localeParity.test.ts` — **67 passed** (was 59 + 8 new: one per-locale guard assertion × 8 locales, plus 4 helper unit tests covering rejection, per-locale exemption, prose/non-string leaves, and a near-miss key that must NOT inherit an exemption).
- `vitest run src/lib/i18n/{translationCoverage,extractionQuality,terminologyQuality}.test.ts` — 22 passed.
- `tsc --noEmit -p apps/web/tsconfig.json` — clean.
- `astro check` — 0 errors, 0 warnings.

Verified the guard actually bites (not vacuous): checked out the pre-fix `origin/main` catalogs under the new test and it failed for **all 8 locales**, naming all seven offending keys with their values. Working tree restored afterwards.

Related: #2649 and #3424 (same codemod, different symptoms), #3423 (fixed the one live route key).


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer.

**Findings:** 3 raised → 1 addressed in 38964d06f, 2 declined with reasons; 0 outstanding.

- **Addressed — array leaves bypassed the guard.** `flattenValues` stops at arrays, so an array-typed leaf reached `routePathValueErrors` whole and its `typeof value === 'string'` check skipped it outright: a bare path listed inside a translatable array would have slipped past the very check added for it. No such leaf exists today, but it is the same seam the guard exists to close. Array leaves are now scanned element-wise and errors report the offending index (`key[1]`), with a unit test covering it.
- **Declined — an English path value containing a space** (`"/etc/ssh /config"`) evades both this guard and the cross-locale protected-literal contract. Reviewer's own confidence ~40. No real route or filesystem path contains a literal space, and widening the check to whitespace-bearing values would pull in genuine prose (`"/home and /etc with caches excluded."`). Left as-is deliberately.
- **Declined — no component test for the inlined `ComplianceTab.tsx` placeholder.** No `ComplianceTab.test.tsx` exists at all, and asserting a cosmetic placeholder string would pin nothing the guard doesn't already: the key cannot be reintroduced as a translatable value without redding the parity suite. Not worth the file.

**Independently verified by the reviewers** (not just asserted): all 7 deleted keys have zero remaining call sites including dynamic/template-literal construction and `.astro` files; the Python JSON round-trip changed exactly 7 lines per locale with no reordering or re-escaping; all 8 locales hold identical leaf counts (1686) post-deletion; and the guard is non-vacuous under mutation — neutering its core condition kills 2 helper tests, and re-adding a deleted key reds the per-locale assertion.

**Notable:** `tr-TR` on `main` had **already localized** two of these route paths (`/configuration-policies` → `/yapılandırma-politikaları`, `/integrations` → `/entegrasyonlar`). The failure mode #3426 predicts has already occurred in the catalog; it stayed harmless only because those keys were dead by then.

**Tests:** `vitest run src/lib/i18n/localeParity.test.ts` — 68 passed (59 pre-existing + 9 new). `translationCoverage` / `extractionQuality` / `terminologyQuality` — 22 passed. `tsc --noEmit -p apps/web/tsconfig.json` clean; `astro check` 0 errors, 0 warnings. Full CI running on the PR.

**Status:** review-clean, awaiting maintainer merge. Not merging or closing #3426 — the reporter/maintainer call.

Issue #3484, opened 2026-08-12
Quote rich-text sanitizer strips <table> markup silently


ToddHebebrand:
## Description
While recreating a large (21-page) external proposal as a Breeze quote, pasting/authoring an HTML table into a quote `rich_text` block collapsed into run-together inline text on save, with no warning or feedback that structure had been lost.

## Root Cause
The rich-text sanitizer's allowlist for `rich_text` blocks has no entries for table tags (`<table>`, `<tr>`, `<td>`, `<th>`, etc.), so they're stripped along with their structural role, leaving only the inline text content concatenated together. This happens silently — the editor doesn't warn the author that markup was removed.

## Proposed Fix
Either:
1. Add table tags to the sanitized allowlist and implement rendering support in the web editor, portal view, and PDF renderer, or
2. If table support is out of scope, reject or warn at input/paste time so authors know to reformat (e.g. as a list) instead of discovering the loss only after the fact in the rendered output.

## Affected Files
- Rich-text sanitizer allowlist (quote `rich_text` block content)
- Web editor / portal view / PDF renderer for `rich_text` blocks (if adding table support)

## Workaround
Render tabular data as a list instead of a table.

PR #3667, opened 2026-08-17
fix(api,web): report rich-text markup the sanitizer strips instead of a silent 200 (#3520)

### LINKED ISSUE #3520: [API] Rich-text sanitizer silently strips disallowed content — 200 response, data gone
## Description
Writing a rich-text quote block whose HTML contains tags outside the allowed subset (e.g. `<table>`, `<blockquote>`) returns HTTP 200, but the disallowed content is silently removed by the sanitizer. There is no warning at write time and no indication at read time — the author only discovers the loss by noticing their content is missing. For a `<table>`, the entire tabular structure disappears.

## Root Cause
`apps/api/src/services/richTextSanitize.ts` allows 11 tags (`RICH_TEXT_ALLOWED_TAGS`, line 6: `p, br, strong, em, u, h3, h4, ul, ol, li, a`) and sanitize-html drops everything else by design. The write path accepts the input, sanitizes, and persists the reduced HTML without comparing against what was submitted.

## Proposed Fix
Detect lossy sanitization at write time (e.g. compare normalized sanitized output against normalized input, or have the sanitizer report dropped tags) and either:
- reject with a 422 naming the disallowed tags, or
- accept but return a `warnings` field naming what was stripped, and surface it in the editor.

Rejecting is the safer default — a 200 that discards content is the worst of the options.

## Affected Files
- `apps/api/src/services/richTextSanitize.ts` (primary)
- quote block write routes in `apps/api/src/routes/quotes/` (wire the warning/rejection through)

Found during internal quote-authoring work.

ToddHebebrand:
Closes #3520

## The bug

The proposal rich-text subset allows 11 block tags (`p br strong em u h3 h4 ul ol li a`), and 5 inline marks in table cells. `sanitize-html` discards everything else by design. The write path accepted the input, sanitized it, persisted the reduced HTML, and answered a clean **200**.

So an author who pasted a `<table>`, `<blockquote>` or `<h1>` was told "saved" and only discovered the loss later by noticing their content was missing. For a `<table>`, the entire tabular structure vanished.

## The fix — fix the silence, not the subset

Writes now report what they removed.

**`richTextSanitize.ts`** gains reporting variants — `sanitizeRichTextHtmlWithReport`, `sanitizeInlineRichTextWithReport`, `sanitizePlainTextWithReport` — that produce **byte-identical HTML** plus the set of tags the profile discarded (a test asserts the byte-identity across both variants).

Detection uses `sanitize-html`'s public **`onOpenTag`** observer, which fires on the *source* tag before `transformTags` and before the `allowedTags` filter — so it sees discarded tags. The collector `Set` is created per call; module-level options (shared with the read/render paths) are never mutated. No wildcard `transformTags` (that would couple production sanitisation to transform ordering) and no second parse.

**Wire format.** Quote block writes (`rich_text` / `table` / `callout`) and authored contract template versions now answer:

```jsonc
{
  "data": { "id": "blk-1", "...": "..." },
  "warnings": [
    { "code": "UNSUPPORTED_HTML_TAGS_REMOVED", "field": "content.html", "removedTags": ["blockquote", "table", "td", "tr"] }
  ]
}
```

`warnings` is **always present** (usually empty) and lives beside `data`, never inside it. A table block's 48 lossy cells collapse into **one** warning per field family (`content.rows[].cells[]`) rather than 48 copies.

**Web.** `QuoteEditor` toasts a non-blocking warning naming the dropped tags on every block add/update. New `apps/web/src/lib/richTextWarnings.ts` (`strippedTagsFrom`) parses the envelope defensively — an API build with no `warnings` key returns `[]` rather than throwing, so it can be wired in unconditionally. i18n key added to all 8 locales.

**Observability.** A write-boundary `console.warn` carries the quote/block/version ids and the tag names — never raw HTML, and not inside the sanitizer, where the read and PDF-render paths would emit the same line on every page view.

## Design notes — what was deliberately NOT done

**No 422.** The issue floated rejecting as the safer default; a quorum (me + an independent Codex read) landed on warn-and-accept instead:

1. The **same sanitizer runs on the READ path and the template-render path** as defense in depth, so rejection could only ever cover a subset of call sites.
2. Real-world paste (Word, Google Docs, external proposals) is full of benign `div`/`span`/`font`/`meta`/`o:p` noise where dropping the tag costs **no author content**. A blanket 422 would make paste unusable and would be a hard behavioural regression for existing API clients and bulk template imports.
3. A "reject only on content-bearing structural tags, warn on the rest" hybrid was considered and rejected as an unprincipled allowlist — `h1` keeps its text, `table` loses structure, `img` loses media, `pre` loses whitespace semantics; the line isn't drawable.
4. The defect in the title is the **silence**, not the strip.

**Only TAG removal is reported.** Stripped attributes (`style`, `class`) and rejected `href` schemes (`javascript:`, protocol-relative) are not — nothing is lost in those cases, and flagging them would make every Word paste look alarming. Covered by an explicit test.

**The READ path stays silent.** A legacy row must render, not warn.

## Out of scope

Sibling issue #3484 (widen the allowlist so `<table>` actually survives) is deliberately **not** in this PR — that needs editor, browser-render *and* PDF-render support landing together. What this PR gives it: `#3484`'s users now at least get told the table was dropped, and `RICH_TEXT_ALLOWED_TAGS` remains the single place the three renderers agree on.

## Verification

- `apps/api`: 283 passed across `richTextSanitize`, `quoteService`, `contractTemplateService`, `aiToolsQuotes`, `routes/quotes/`, `routes/contracts/` (single-fork).
- `apps/web`: 447 passed across `components/billing/quotes/` + the new `richTextWarnings` unit test; i18n `localeParity` / `translationCoverage` / `terminologyQuality` / `extractionQuality` + `no-silent-mutations` green (182 passed).
- `tsc --noEmit` clean in both `apps/api` and `apps/web`.
- The new `QuoteEditor.strippedmarkup.test.tsx` was mutation-checked: commenting out the `notifyStrippedMarkup` calls turns it red.

### DIFF
--- a/apps/api/src/routes/contracts/templates.test.ts
+++ b/apps/api/src/routes/contracts/templates.test.ts
@@ -213,6 +213,32 @@ describe('contract template routes', () => {
     expect(svc.createDraftVersion).toHaveBeenCalledWith(expect.anything(), TEMPLATE_ID, { bodyHtml: '<p>New draft</p>' });
   });
 
+  // Issue #3520: the response envelope carries `warnings` beside `data` so a
+  // lossy save can't read as a clean success. Warnings never ride inside `data`.
+  it('POST /:id/versions surfaces stripped markup as top-level warnings, outside data', async () => {
+    const warnings = [{ code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'bodyHtml', removedTags: ['table'] }];
+    (svc.createDraftVersion as any).mockResolvedValue({
+      id: VERSION_ID,
+      versionNumber: 2,
+      sourceType: 'authored',
+      bodyHtml: '<p>New draft</p>',
+      status: 'draft',
+      orgId: ORG_ID,
+      partnerId: null,
+      warnings,
+    });
+    const res = await app().request(`${BASE}/${TEMPLATE_ID}/versions`, {
+      method: 'POST',
+      headers: { 'content-type': 'application/json' },
+      body: JSON.stringify({ bodyHtml: '<p>New draft</p><table><tr><td>x</td></tr></table>' }),
+    });
+    expect(res.status).toBe(200); // still a success — the version saved
+    const body = await res.json();
+    expect(body.warnings).toEqual(warnings);
+    expect(body.data.warnings).toBeUndefined();
+    expect(body.data.versionNumber).toBe(2);
+  });
+
   it('POST /:id/versions rejects an empty bodyHtml (400, no service call)', async () => {
     const res = await app().request(`${BASE}/${TEMPLATE_ID}/versions`, {
       method: 'POST',
--- a/apps/api/src/routes/contracts/templates.ts
+++ b/apps/api/src/routes/contracts/templates.ts
@@ -140,8 +140,10 @@ contractTemplateRoutes.post(
   async (c) => {
     try {
       const { id } = c.req.valid('param');
-      const version = await createDraftVersion(authFrom(c), id, c.req.valid('json'));
-      return c.json({ data: serializeVersion(version) });
+      // `warnings` names any markup the rich-text subset discarded from the
+      // submitted body — always present, usually empty (issue #3520).
+      const { warnings, ...version } = await createDraftVersion(authFrom(c), id, c.req.valid('json'));
+      return c.json({ data: serializeVersion(version), warnings });
     } catch (err) {
       return handleTemplateError(c, err);
     }
--- a/apps/api/src/routes/quotes/quotes.test.ts
+++ b/apps/api/src/routes/quotes/quotes.test.ts
@@ -395,6 +395,37 @@ describe('quote crud + lines routes', () => {
     );
   });
 
+  // Issue #3520: the sanitizer silently discarded out-of-subset markup and the
+  // route answered a clean 200. The service's `warnings` now ride the envelope
+  // beside `data` — never folded into the block itself.
+  it('PATCH /:id/blocks/:blockId surfaces stripped markup as top-level warnings, outside data', async () => {
+    const warnings = [{ code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'content.html', removedTags: ['blockquote', 'table'] }];
+    (svc.updateBlock as any).mockResolvedValue({ id: BLOCK_ID, blockType: 'rich_text', content: { html: '<p>kept</p>' }, warnings });
+    const res = await app().request(`/${QUOTE_ID}/blocks/${BLOCK_ID}`, {
+      method: 'PATCH',
+      headers: { 'content-type': 'application/json' },
+      body: JSON.stringify({ blockType: 'rich_text', content: { html: '<p>kept</p><table><tr><td>gone</td></tr></table>' } }),
+    });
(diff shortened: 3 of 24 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (correctness: onOpenTag detection, read/write path split, return-shape change across all callers, route envelope, web wiring) + `silent-failure-hunter` (whether the new feedback mechanism is itself silent anywhere).

**Findings:** 1 raised → addressed in `4b8b128`; 0 outstanding.

- `code-reviewer`: **0 findings.** Confirmed the per-call collector `Set` can't leak, the reporting variants emit byte-identical HTML, `sanitizeQuoteBlocksForRead` is behaviourally unchanged and never logs, `warnings` never reaches a `.values()`/`.set()` call (no JSONB persistence risk), and every repo-wide caller of `addBlock`/`updateBlock`/`createDraftVersion` still works.
- `silent-failure-hunter`: **1 finding, confirmed real.** The API half made `POST /contract-templates/:id/versions` answer `{ data, warnings }`, but `TemplateEditor.saveDraft` never read them — it toasted a flat "Saved" via `runAction`'s `successMessage`. An admin pasting a table into a contract body hit the exact silent loss this PR exists to fix, one file over, on content that becomes an executed legal document. Fixed in `4b8b128`: `saveDraft` now branches in `parseSuccess` (warn naming the tags, or success when clean), reusing `strippedTagsFrom`; i18n key added across all 8 locales; regression test mutation-checked. The hunter separately verified the quote-editor wiring has no bypass (bulk edit, tidy-all, autosave, cover page, `QuoteStructuredBlockEditor` all route through the covered `addBlock`/`editBlock` paths) and that every remaining silent `sanitizeRichTextHtml` call site is a read/render path where silence is correct.

**Tests:** `apps/api` 283 passed (`richTextSanitize`, `quoteService`, `contractTemplateService`, `aiToolsQuotes`, `routes/quotes/`, `routes/contracts/`, single-fork); `apps/web` 447 + 255 passed across the quote and contract editor suites plus the new `richTextWarnings` unit test; i18n `localeParity` / `translationCoverage` / `terminologyQuality` / `extractionQuality` and `no-silent-mutations` green. `tsc --noEmit` clean in both apps; `eslint` clean on every touched file. Both new component tests are mutation-checked — removing the notify wiring turns them red.

**CI:** all 55 checks green, including Test API, Test Web, Type Check and all 4 Integration Tests shards.

**Status:** review-clean, awaiting maintainer merge. Issue #3520 stays open and assigned until you've verified.

**Note for #3484** (widen the allowlist so `<table>` survives — deliberately not in this PR): `RICH_TEXT_ALLOWED_TAGS` / `INLINE_RICH_TEXT_ALLOWED_TAGS` remain the single place the TipTap editor, the browser render and `richTextPdf.ts` agree on, so that issue is still a one-list change plus three renderers. Two things this PR hands it: authors now at least get told the table was dropped rather than discovering it later, and `sanitizeInlineRichTextWithReport` gives the table-cell profile the same reporting — worth remembering that cells are inline-only *by design* (Spec A decision #4), so #3484 needs to decide whether nested tables in cells stay unsupported even after `<table>` is allowed at block level.

PR #3670, opened 2026-08-17
fix(agent): uninstall reported success without removing anything (#3592)

### LINKED ISSUE #3592: [Software Inventory] Stale/incorrect entries persist even after a fresh scan — confirmed via empty registry query after uninstall, and via agent version never updating post-self-update

ToddHebebrand:
## Problem

`software_inventory` kept listing a program that Breeze had "successfully" uninstalled, **even across a genuinely fresh scan** — the reporter confirmed a `device_commands` row with `{"action":"uninstall","success":true}`, `exitCode: 0`, and an inventory `last_seen` ~20 minutes *later*.

The API side was ruled out by the #3591 / #3663 investigation: `apps/api/src/routes/agents/inventory.ts` is the only `insert(softwareInventory)` site, and it wipes-and-reinserts the entire device list in one transaction with `lastSeen: now`. So a stale row carrying a fresh `last_seen` is not a merge bug — it means **the agent really did re-report the program**. It did, because the program was still installed, and the uninstall that claimed success never touched it.

## Root cause

`runUninstallAttempts` (`agent/internal/remote/tools/software.go`) had **two** paths that returned `nil` with no evidence of removal:

1. **A provider "not found" message ended the whole operation as "already absent."** winget is the *first* Windows attempt and answers `No installed package found matching input criteria.` for anything it does not index — under the SYSTEM service account that includes per-user installs and plenty of ordinary Add/Remove Programs entries. Breeze therefore returned success, **never ran the `wmic` fallback**, and never touched the machine.
2. **`err == nil` from the provider.** `wmic product where name='X' call uninstall` exits 0 even when the `WHERE` clause matches nothing, so the fallback attempt reported success unconditionally.

Both reproduce the reported evidence exactly.

## Fix

- A provider that cannot see the package **falls through to the next provider** instead of ending the run.
- The outcome is decided by a **post-condition check** against the same collector the agent reports inventory from (`softwareStillInstalled`, new `software_uninstall_verify.go`). "Still present" after the attempts is a failure, with a message that distinguishes the lying-provider case from the no-provider-could-find-it case.
- A collector error is **not** treated as proof of either state — it falls back to the provider signals rather than inventing a verdict.
- Matching is exact (case-insensitive, `.app`-normalized), not substring, so `"Microsoft Teams"` is not reported as still installed merely because `"Microsoft Teams Meeting Add-in"` remains.

### Two supporting fixes in the same path

- **`collectors/software_windows.go` called `isSystemComponent(subkey)` *after* `subkey.Close()`.** Every registry read on a closed handle fails, so the function always answered `false` — the `SystemComponent=1` / `"Update for …"` filter **has never filtered anything**, and Windows Update / system-component entries have been reported as ordinary installed software all along. That is the "incorrect entries" half of the title. The check now runs before the handle is closed and reuses the already-read `DisplayName`.
- **`handleSoftwareUninstall` now re-reports software inventory on success**, instead of leaving the dashboard stale for up to 15 minutes (the `sendInventory` cadence) — which was operationally indistinguishable from the silent-no-op bug above. Safe to fire immediately, because success now means *verified removed*.

## Tests

10 new tests in `agent/internal/remote/tools/software_uninstall_verify_test.go`, including the exact #3592 scenario (winget not-found + wmic exit-0 with the software still present must fail, and must attempt both providers). Idempotent re-runs still succeed when the software is verified absent.

```
go test -race ./internal/remote/tools/ ./internal/collectors/ ./internal/heartbeat/   → all ok
go vet (native + GOOS=windows)                                                        → clean
GOOS=windows/linux/darwin go build ./...                                              → clean
```

## Follow-ups (deliberately not in this PR)

- The Windows collector reads `registry.CURRENT_USER`, which under the SYSTEM service is the **SYSTEM profile's** hive — real users' per-user installs are invisible to inventory. Fixing needs `HKEY_USERS\<SID>` enumeration and would materially change inventory volume.
- `sendSoftwareInventory` drops the whole report when collection errors, leaving the server's previous list in place. Arguably correct (don't wipe inventory on a failed scan) but worth an explicit staleness signal.

Closes #3592

### DIFF
--- a/agent/internal/collectors/software_windows.go
+++ b/agent/internal/collectors/software_windows.go
@@ -28,11 +28,18 @@ var softwareRegistryPaths = []struct {
 func (c *SoftwareCollector) Collect() ([]SoftwareItem, error) {
 	var software []SoftwareItem
 	seen := make(map[string]bool)
+	var pathErrs []string
 
 	for _, regPath := range softwareRegistryPaths {
 		items, err := collectFromRegistry(regPath.root, regPath.path)
 		if err != nil {
-			// Continue on error - some paths may not exist or be accessible
+			// Continue on error - some paths may not exist or be accessible.
+			// A single unreadable hive is normal (WOW6432Node is absent on
+			// 32-bit Windows; HKCU under the SYSTEM service is the SYSTEM
+			// profile), so it must not fail the whole collection — but record
+			// it so a total failure below can be distinguished from a genuinely
+			// empty machine.
+			pathErrs = append(pathErrs, fmt.Sprintf("%s: %v", regPath.path, err))
 			continue
 		}
 
@@ -44,6 +51,15 @@ func (c *SoftwareCollector) Collect() ([]SoftwareItem, error) {
 				software = append(software, item)
 			}
 		}
+	}
+
+	// Every registry path failed: this is "we could not look", not "nothing is
+	// installed". Returning (nil, nil) here would let callers that treat an
+	// empty list as ground truth — notably the uninstall post-condition check in
+	// remote/tools — conclude that software is absent without ever having read
+	// the registry (#3592).
+	if len(pathErrs) == len(softwareRegistryPaths) {
+		return nil, fmt.Errorf("no installed-software registry path could be read: %s", strings.Join(pathErrs, "; "))
 	}
 
 	return software, nil
--- a/agent/internal/collectors/software_windows.go
+++ b/agent/internal/collectors/software_windows.go
@@ -70,6 +70,13 @@ func collectFromRegistry(rootKey registry.Key, path string) ([]SoftwareItem, err
 		}
 
 		item := readSoftwareFromKey(subkey)
+		// Both reads must happen BEFORE the handle is closed. isSystemComponent
+		// used to be called on the already-closed `subkey`, where every registry
+		// read fails and the function therefore always answered false — so the
+		// SystemComponent=1 / "Update for …" filter below has never actually
+		// filtered anything, and Windows Update and system-component entries have
+		// been reported as ordinary installed software all along (#3592).
+		systemComponent := isSystemComponent(subkey, item.Name)
 		subkey.Close()
 
 		// Skip items without a display name or system components
@@ -78,7 +85,7 @@ func collectFromRegistry(rootKey registry.Key, path string) ([]SoftwareItem, err
 		}
 
 		// Skip Windows updates and system components
-		if isSystemComponent(subkey) {
+		if systemComponent {
 			continue
 		}
 
@@ -160,7 +167,11 @@ func parseInstallDate(dateStr string) string {
 	return ""
 }
 
-func isSystemComponent(key registry.Key) bool {
+// isSystemComponent reports whether an Uninstall subkey describes a system
+// component or a Windows Update rather than user-facing installed software.
+// `key` must still be open; `displayName` is passed in because the caller has
+// already read it.
+func isSystemComponent(key registry.Key, displayName string) bool {
 	// Check SystemComponent flag
 	val, _, err := key.GetIntegerValue("SystemComponent")
 	if err == nil && val == 1 {
@@ -168,7 +179,7 @@ func isSystemComponent(key registry.Key) bool {
 	}
 
 	// Check for Windows Update entries
-	name, _ := readStringValue(key, "DisplayName")
+	name := displayName
 	if strings.HasPrefix(name, "Update for") ||
 		strings.HasPrefix(name, "Security Update for") ||
 		strings.HasPrefix(name, "Hotfix for") {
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter on `cd3931fa0`, then a targeted code-reviewer re-run on the follow-up `8d7282ec1`.

**Findings:** 5 raised, all real, all addressed.

- *Verification could launder a hard provider error into success.* A brew cask token (`google-chrome`) does not match the collector's display name (`Google Chrome`), so "absent from the collector" can mean "the collector cannot see this name" — which would have turned a genuine apt/brew/msiexec failure into `nil`. Verification may now only **downgrade** an unproven success, never upgrade an error. → `8d7282ec1`
- *"No provider could see the package" + unverifiable = silent success.* Reachable in production (a `system_profiler` hiccup during an uninstall of software Homebrew does not manage). Only a genuine removal claim is honoured when verification is unavailable. → `8d7282ec1`
- *wmic's exit-0 lie was the one signal still trusted when verification failed.* `wmic product where name='X' call uninstall` exits 0 and prints `No Instance(s) Available.` on no match, so the no-match markers are now checked on the success path too. → `8d7282ec1`
- *The collector was not a sound oracle.* `software_windows.go` returned an empty list with a nil error when every registry path failed, and the Linux collector does the same when neither dpkg-query nor rpm is usable. Both now read as "could not look", not "nothing installed". → `8d7282ec1`
- *Vacuous test.* Both cases in the re-report gate test were validation failures, so a handler that never fired the re-report would have passed it. Now drives both branches with an injected result. → `77c490f52`

Also corrected an overclaiming comment: the macOS direct-removal path returns on a successful `os.RemoveAll` without consulting the collector. The immediate re-report is still safe (every path is synchronous, and the report is idempotent — the API replaces the device's whole software list), but for the right reason.

**Re-review of the fixes:** 0 outstanding. The full decision table was re-walked cell by cell — {exit-0 real / exit-0 no-match / hard error / not-found / no provider} × {verified absent / present / unverifiable} — with no path returning `nil` without evidence and no path converting a previously-failing uninstall into a silent success. The all-registry-paths-failed guard was confirmed unreachable on a healthy endpoint (32-bit Windows loses only WOW6432Node; SYSTEM loses only HKCU — 2 of 3 still succeed).

**Tests:** 15 new (13 in `software_uninstall_verify_test.go`, 2 in `handlers_software_uninstall_test.go`). `go test -race ./internal/remote/tools/ ./internal/collectors/ ./internal/heartbeat/` green; `go vet` clean natively and under `GOOS=windows`; `go build ./...` clean for windows/linux/darwin. CI: Test Agent, Lint Agent, and all four Build Agent targets green on the prior push.

**Two follow-ups deliberately left out of this PR** (both pre-existing, neither is the reported bug):
1. The Windows collector reads `registry.CURRENT_USER`, which under the SYSTEM service is the **SYSTEM profile's** hive — real users' per-user installs are invisible to inventory entirely. Needs `HKEY_USERS\<SID>` enumeration and would materially change inventory volume.
2. The macOS direct-removal path bypasses the new post-condition check (`os.RemoveAll` success is taken as proof). Defensible, but worth routing through verification for consistency.

PR #3671, opened 2026-08-17
fix(api,web,portal): let <table> survive the rich-text sanitizer (#3484)

### LINKED ISSUE #3484: Quote rich-text sanitizer strips <table> markup silently

ToddHebebrand:
Closes #3484

## The bug

`RICH_TEXT_ALLOWED_TAGS` had no table entries, so pasting a real table into a quote/contract `rich_text` block dropped every structural tag and left the cell text concatenated. #3520 / #3667 made that loss *visible* (write responses now carry a `warnings` array, the editor toasts it) and explicitly deferred the other half — actually letting tables through — to this issue.

## What changed

**`richTextSanitize.ts` — the one list the editor, the browser render and the PDF renderer all agree on**

- `table thead tbody tfoot tr th td` are allowed, with **no attributes**.
- **`colspan`/`rowspan` are deliberately not allowed.** A span only pays off if all three renderers honour it, and `rowspan` has no representation in a row-at-a-time PDF renderer. A merged Word cell lands in one column instead of spanning; every cell's *text* still survives.
- **Cells stay inline-only**, keeping the contract the structured `table` block already has (Spec A decision #4) — the question the sibling PR flagged. Enforcement is a post-sanitize pass that re-runs each `<td>`/`<th>` body through the inline profile. `sanitize-html`'s `allowedTags` is flat, so without this a cell could hold a `<p>`/`<ul>`/nested `<table>` that the browser renders natively and the fixed-height PDF cell cannot — exactly the three-renderer drift this module exists to prevent.
- Block boundaries inside a cell become `<br />` rather than vanishing, so Word's `<td><p>a</p><p>b</p></td>` reads "a / b" instead of "ab", and a nested table's cells flatten one per line.
- Everything the cell pass removes is **merged into the `*WithReport` payload**, so #3520's toast still fires — the flattening is not a new silent strip. The scan runs on `sanitize-html` output (balanced, attribute-free) and re-emits only literal `<td>`/`<th>` delimiters copied from it or the output of another `sanitize-html` pass, so it cannot reintroduce markup.

**`richTextPdf.ts` — tables get drawn**

- `RichTextBlock` becomes a union; `parseRichText` emits a `table` block of rows of inline-run cells, collected through `thead`/`tbody`/`tfoot` (and straight off `<table>`, which is what a bare `<table><tr>` paste produces). A `<tr>`/`<td>` orphaned from its `<table>` folds into an implicit paragraph rather than disappearing from the PDF while still rendering in HTML.
- The grid draws with content-inferred column widths (HTML carries no weights; the weight is the longest plain-text cell, clamped so one prose-heavy column can't starve the rest), a tinted repeated header row, hairline borders, rows that never split, and an oversized-row degrade to stacked "header: value" paragraphs.
- Page breaks are detected from `doc.y` — the trick the block loop already used — so `renderRichTextIntoPdf`'s public `ensureRoom: (needed) => number` signature and all four call sites are unchanged.
- **Not `tablePdf.ts`.** That module renders the *structured* `table` block type, whose author-supplied weights/zebra/header-style have no counterpart in pasted HTML, and it imports this module — the dependency can only run one way. What the two share is the cell primitive (`drawRuns`/`measureRuns`, factored out of the existing inline exports).

**`RichTextEditor.tsx` — TipTap emits exactly what the sanitizer keeps**

TipTap's defaults would not round-trip: `<table style="min-width:…">` with a `<colgroup>`, cells carrying `colspan`/`rowspan`/`colwidth` even at defaults, and `<p>` inside every cell. A `getHTML()` that the sanitizer rewrites leaves the block permanently "unsaved" — it re-PATCHes on every blur and (since #3520) toasts a bogus "markup removed" warning each time, the same trap the `rel="nofollow"` override already documents. So: `renderHTML` overridden to a plain `<table><tbody>`, resizing off, cell attributes suppressed **at render time only** (they stay in the schema — prosemirror-tables' `TableMap` needs them), and cell content `inline*`.

Toolbar gains an insert-table button plus row/column/delete controls that appear only with the caret inside a table. The active-state reads move to `useEditorState`: TipTap 3's `useEditor` does not re-render on every transaction, so reading `editor.isActive(...)` in the render body left the table controls never appearing (and, latently, the existing `aria-pressed` states stale).

**Portal CSS** — `apps/portal` deliberately doesn't pull in Tailwind's typography plugin (web gets table styling free from `prose`), so `.quote-rich-text table/th/td` rules are hand-written to match the PDF's look. Without them a customer-facing proposal renders a borderless run of cells.

## Out of scope

Merged cells (`colspan`/`rowspan`), nested tables, per-column widths, and cell alignment. The structured `table` block remains the way to author a table with layout control; this is about a pasted table surviving intact.

## Verification

- `apps/api`: 345 passed across `richTextSanitize`, `richTextPdf`, `tablePdf`, `calloutPdf`, `quotePdf` (+ classic regression), `quoteService`, `contractTemplateService`, `contractDocumentService`, `contractTemplateRender`, `aiToolsQuotes`, `routes/quotes/`, `routes/contracts/` — single-fork.
- `apps/web`: 795 passed across `components/billing/quotes/`, `components/contracts/`, `components/common/`, `lib/i18n` (localeParity / translationCoverage / terminologyQuality / extractionQuality), `richTextWarnings`, `no-silent-mutations`.
- `tsc --noEmit` clean in `apps/api` and `apps/web`; `apps/portal` `astro build` clean (validates the new `@apply` rules).
- The three tests that used `<table>` as their canonical *stripped-tag* fixture now use `<pre>`/`<blockquote>`, which are still outside the subset — the assertions were re-pointed, not weakened.

### DIFF
--- a/apps/api/src/services/contractTemplateService.test.ts
+++ b/apps/api/src/services/contractTemplateService.test.ts
@@ -262,7 +262,7 @@ describe('createDraftVersion', () => {
     expect(persisted.bodyHtml).toContain('Hi');
   });
 
-  // Issue #3520: sanitizing away a pasted <table>/<blockquote> used to be
+  // Issue #3520: sanitizing away a pasted <pre>/<blockquote> used to be
   // invisible — the version saved and the author found out later.
   it('reports the tags it stripped from bodyHtml instead of saving silently', async () => {
     const auth = makeAuth({ scope: 'organization', canAccessOrg: () => true });
@@ -270,11 +270,11 @@ describe('createDraftVersion', () => {
     queueResult([{ maxVersion: null }]);
     queueResult([{ id: 'v1', versionNumber: 1 }]);
     const row = await svc.createDraftVersion(auth, ORG_TEMPLATE.id, {
-      bodyHtml: '<p>Terms</p><blockquote>note</blockquote><table><tr><td>c</td></tr></table>',
+      bodyHtml: '<p>Terms</p><blockquote>note</blockquote><pre>c</pre>',
     });
     expect(row.id).toBe('v1'); // still saves — warning, not rejection
     expect(row.warnings).toEqual([
-      { code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'bodyHtml', removedTags: ['blockquote', 'table', 'td', 'tr'] },
+      { code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'bodyHtml', removedTags: ['blockquote', 'pre'] },
     ]);
   });
 
--- a/apps/api/src/services/quoteService.test.ts
+++ b/apps/api/src/services/quoteService.test.ts
@@ -559,10 +559,10 @@ describe('quoteService deposits', () => {
     queueResult([{ blockType: 'rich_text' }]); // existing block type check
     queueResult([{ id: 'blk1', blockType: 'rich_text', content: { html: '<p>Updated</p>' } }]); // update returning
 
-    const row = await svc.updateBlock('q1', 'blk1', { blockType: 'rich_text', content: { html: '<p>Updated</p><table><tr><td>x</td></tr></table>' } }, actor);
+    const row = await svc.updateBlock('q1', 'blk1', { blockType: 'rich_text', content: { html: '<p>Updated</p><pre>x</pre>' } }, actor);
 
     expect(row.warnings).toEqual([
-      { code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'content.html', removedTags: ['table', 'td', 'tr'] },
+      { code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'content.html', removedTags: ['pre'] },
     ]);
   });
 
@@ -795,11 +795,11 @@ describe('sanitizeBlockContentForWrite loss reporting (#3520)', () => {
   it('reports the disallowed tags a rich_text write dropped', () => {
     const { content, warnings } = svc.sanitizeBlockContentForWrite({
       blockType: 'rich_text',
-      content: { html: '<p>Keep</p><blockquote>Quote</blockquote><table><tr><td>Cell</td></tr></table>' },
+      content: { html: '<p>Keep</p><blockquote>Quote</blockquote><pre>Cell</pre>' },
     } as never);
-    expect((content as { html: string }).html).not.toContain('<table');
+    expect((content as { html: string }).html).not.toContain('<pre');
     expect(warnings).toEqual([
-      { code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'content.html', removedTags: ['blockquote', 'table', 'td', 'tr'] },
+      { code: 'UNSUPPORTED_HTML_TAGS_REMOVED', field: 'content.html', removedTags: ['blockquote', 'pre'] },
     ]);
   });
 
--- a/apps/api/src/services/quoteService.test.ts
+++ b/apps/api/src/services/quoteService.test.ts
@@ -803,6 +803,30 @@ describe('sanitizeBlockContentForWrite loss reporting (#3520)', () => {
     ]);
   });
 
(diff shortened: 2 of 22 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (correctness/security focus on the sanitizer's hand-rolled cell pass, the PDF table draw/measure/page-break paths, and the editor's round-trip claim) and `pr-test-analyzer` (mutation-checked the load-bearing assertions).

**Findings:** 1 confirmed defect + 1 CI-surfaced issue → both fixed in `3ce9cfbc0`; 0 outstanding.

- **Pasted `colspan` desynced the ProseMirror model from the DOM.** TipTap's `TableCell` gives `colspan`/`rowspan` no `parseHTML` override, so a pasted `<td colspan="2">` — routine in the Word/Google-Docs tables this feature exists for — set a real span on the node while `renderHTML` suppressed the attribute. The row then drew *fewer* `<td>` elements than the model believed it spanned: the table misaligned **inside the editor** (not just after a save round-trip), and `tableEditing`'s DOM-based cell hit-testing (drag-select, Tab) desynced from the `TableMap`. Both spans are now pinned to `1` on parse (and `colwidth` to `null`), so a merged cell splits into ordinary cells that prosemirror-tables pads into a rectangular grid — the honest outcome given merges are unsupported end to end. New regression test is mutation-checked: removing the clamp turns it red.
- **CodeQL `js/redos` (high)** on `collapseBreaks`' repeated-group regex. Replaced with a split on a *single* break tag plus a rejoin — linear by construction. CodeQL is clean on the current head.
- Test-coverage gaps closed: empty cells, unbalanced cell tags, a wider-than-the-page table, and the first end-to-end coverage that a quote block **write** stores a table intact (and warns when a cell's block content was flattened).
- Reviewed and explicitly cleared: the cell-normalization scan's reliance on `sanitize-html` emitting balanced output (verified empirically, including mismatched close order), its idempotency on the read path, the `removedTags` merge against #3520's contract, and the degrade-path mutual recursion being bounded. The one noted measure/draw divergence (`measureTableBlock` not counting repeated headers) is unreachable today — `calloutPdf` is the only `measureRichText` caller and its `ensureRoom` never paginates — and is now documented in place rather than papered over.

**Tests:** `apps/api` 170 passed across `richTextSanitize`, `richTextPdf`, `quoteService`, `contractTemplateService` (plus 128 earlier across the PDF/quote render suites), single-fork; `apps/web` 12 passed in `RichTextEditor` and 795 across quotes/contracts/common/i18n. `tsc --noEmit` clean in `apps/api` and `apps/web`; `apps/portal` `astro build` clean. **All 55 CI checks green, including Test API, Test Web and all four Integration Tests shards.**

**One thing worth a human eye before merge:** none of this was exercised in a browser or against a real generated PDF — the table rendering is covered by unit tests and PDF byte assertions only. A quick pass authoring a table in a quote and downloading the PDF would be a cheap confirmation.

PR #3686, opened 2026-08-17
fix(workspace): reject local_profile sources that carry a crawl device (#3472)

### LINKED ISSUE #3472: workspace: sourcePatchSchema never applies validateSmbConfig
Migrated from the archived `LanternOps/breeze-workspace` repo (issue #12, filed 2026-07-20) as part of the ee/workspace merge (#3455). Paths below are updated to their post-merge locations.

---

`sourcePatchSchema` (`ee/workspace/src/routes/sources.ts`, formerly `src/routes/sources.ts:64-67`) uses `.partial().refine(...)` and never applies `validateSmbConfig` at all. `validateSmbConfig` itself also returns early for `local_profile`.

Consequences on PATCH:
- accepts `crawlDeviceId` on a `local_profile` source
- accepts an `smb_share` with no crawl device and no UNC root path

**Pre-existing** — not introduced by the PR it was found in. But it widens the precondition that the read-side device-summary guard defends against: the device-summary union relies on `device_id IS NULL` to avoid attributing *another* device's device-scoped rows to a `local_profile` source that happens to carry a `crawlDeviceId`. That guard is correct and mutation-verified, but the invariant would be better enforced at **write** time than depended on at read time.

Suggested fix: apply `validateSmbConfig` to the patch path, and reject or normalise `crawlDeviceId` for `local_profile` sources at the schema level.

bdunncompany:
Closes #3472.

## The gap

`sourcePatchSchema` is `sourceInputBaseSchema.partial().refine(...)`, so it never gets `validateSmbConfig`. The `.superRefine(validateSmbConfig)` on the adjacent line belongs to `smbConfigSchema`, not to the patch schema.

The smb_share half of the issue turns out to be **already closed**: the PATCH route runs a merged `smbConfigSchema.safeParse` on `{kind, rootPath, crawlDeviceId}` and 400s a source that would end up SMB with no crawl device or a non-UNC root.

The inverse half was genuinely open. Nothing anywhere rejected a `local_profile` holding a `crawl_device_id`:

- `validateSmbConfig` returns early for `local_profile`
- `validateCreate` returns early for `local_profile`
- `update()` did no state validation at all

That shape is not inert. `listForDevice` returns every active `local_profile` in the org **plus** rows assigned to the calling device, so a stranded crawl device hands one device the other device's device-scoped rows.

The service's own test fixture was itself the forbidden shape: `source()` defaulted to a `local_profile` carrying `DEVICE_ID`, so most of the suite exercised a row the service now refuses to write.

## The change

- `validateSmbConfig` rejects `crawlDeviceId` on `local_profile`. Covers POST and the route's merged PATCH check.
- `validateCreate` becomes `validateSourceState` and validates a **complete** state. `update()` reads the current row and validates the merged result, closing a direct-service bypass: `update(orgId, id, { crawlDeviceId: X })` against a `local_profile` previously wrote the forbidden shape with no route involved.
- **Reject, don't clear.** A flip to `local_profile` that would strand a device is rejected rather than silently nulled. A crawl-device assignment is operator configuration, and a body saying only `{kind:'local_profile'}` has not authorised deleting it. The web form always submits `crawlDeviceId` explicitly (`web/sourcesPage.ts:257`), so deliberate flips are unaffected. `credentialEnc` clearing is unchanged.
- PATCH maps `SourceValidationError` to 400, matching POST. Unreachable through this route today because the merged check rejects the same states first; it is there so a drift between the two independent validators cannot surface as an opaque 500.
- Fixture cleaned, plus a test that an `smb_share` assignment **survives** an unrelated update, since nothing covered that once the default went null.

## What I got wrong first

The first revision cleared `crawlDeviceId` on flip and had the route validate a synthetic post-update state. An adversarial review killed it on two counts I verified: that synthetic state is imaginary for a rootPath-only PATCH on an already-stale row (route fabricates null, service doesn't clear, stale device survives), and the direct-service bypass stayed wide open. Both are fixed here.

## Limits I am not fixing, deliberately

- The UNC check is `rootPath.startsWith('\\\\')`, so `\\` and `\\server` pass.
- GET/validate/UPDATE is not atomic and the update predicate carries no expected-kind condition, so two individually-valid concurrent PATCHes can compose into an invalid row.
- No DB CHECK constraint ties kind, crawl device and rootPath together, so raw SQL still writes the shape.
- `crawl_device_id` is `ON DELETE SET NULL`. Deleting a device orphans its `smb_share` into a state the merged validator rejects, so the source cannot be paused or renamed until a replacement device is assigned. This is **pre-existing on main** (that identical merged check is already in the PATCH route) and is not widened for HTTP callers here, though this change does extend the same rejection to the service layer. Worth its own issue and happy to take it.

Say the word if you'd rather have any of these in scope.

## Verification

`ee/workspace` vitest: **37 files / 593 tests pass**. `tsc --noEmit` and eslint clean on all four changed files.

Two control runs, because a passing suite proves nothing on its own:

- Neuter both rejection guards: exactly the **6 new rejection tests** fail (3 route, 3 service).
- Make `update()` clear `crawlDeviceId` unconditionally: only the **new preservation test** fails.

The 7th new test is a positive case (a deliberate flip that clears the device explicitly) and is not expected to move under either mutation.

### DIFF
--- a/ee/workspace/migrations/2026-08-17-workspace-local-profile-crawl-device-backfill.sql
+++ b/ee/workspace/migrations/2026-08-17-workspace-local-profile-crawl-device-backfill.sql
@@ -1,0 +1,31 @@
+-- #3472: `local_profile` sources must never carry a crawl device.
+--
+-- Nothing enforced that before: validateSmbConfig returned early for
+-- local_profile, validateCreate did the same, and sourcesService.update did no
+-- state validation at all. So existing rows can hold the forbidden shape — the
+-- service's own test fixture defaulted to exactly it.
+--
+-- The shape is not inert. listForDevice returns every active local_profile in
+-- the org PLUS rows assigned to the calling device, so a stranded crawl device
+-- hands one device the other device's device-scoped rows.
+--
+-- It also breaks editing once validation lands. The admin form seeds the crawl
+-- device from the row it is editing (web/sourcesPage.ts) and resubmits it, so
+-- without this backfill every PATCH against a legacy row would 400 and the
+-- operator could not clear it through the UI.
+--
+-- Idempotent: the predicate only matches rows still holding the forbidden
+-- shape, so re-applying is a no-op.
+DO $$
+DECLARE
+  n integer;
+BEGIN
+  UPDATE workspace_sources
+  SET crawl_device_id = NULL
+  WHERE kind = 'local_profile'
+    AND crawl_device_id IS NOT NULL;
+  GET DIAGNOSTICS n = ROW_COUNT;
+  IF n > 0 THEN
+    RAISE WARNING 'cleared crawl_device_id on % local_profile workspace_source(s) (#3472)', n;
+  END IF;
+END $$;
--- a/ee/workspace/src/__tests__/localProfileCrawlDeviceBackfill.integration.test.ts
+++ b/ee/workspace/src/__tests__/localProfileCrawlDeviceBackfill.integration.test.ts
@@ -1,0 +1,108 @@
+// #3472: the backfill that clears crawl_device_id from legacy local_profile
+// rows. Real Postgres, because the migration is plain SQL — a unit test cannot
+// tell a correct predicate from an inverted one, and getting this wrong either
+// leaves the forbidden shape in place or wipes crawl devices off smb_share
+// sources that legitimately need them.
+//
+// Bootstrap mirrors ingestJobs.integration.test.ts (:5433 stack, admin role,
+// idempotent migration re-apply).
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import postgres from 'postgres';
+import { randomUUID } from 'node:crypto';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+
+const ADMIN_URL = process.env.DATABASE_URL ?? 'postgresql://breeze_test:breeze_test@localhost:5433/breeze_test';
+if (new URL(ADMIN_URL).port === '5432') throw new Error('refusing to run against :5432 — use the test stack (:5433)');
+
+const MIGRATION = '2026-08-17-workspace-local-profile-crawl-device-backfill.sql';
+
+let admin: postgres.Sql;
+let partner: string, org: string, site: string, deviceId: string;
+let legacyLocal: string, cleanLocal: string, smbSource: string;
+
+async function applyBackfill() {
+  const here = dirname(fileURLToPath(import.meta.url));
+  const sql = readFileSync(join(here, '../../migrations/', MIGRATION), 'utf8');
+  await admin.begin(async (tx) => { await tx.unsafe(sql); });
+}
+
+beforeAll(async () => {
+  admin = postgres(ADMIN_URL, { max: 1 });
+
+  // The shared :5433 stack may not carry the workspace tables (it gets
+  // reprovisioned by other branches' runs), so apply the foundation
+  // idempotently first — same re-apply gotcha ingestJobs.integration.test.ts
+  // documents.
+  const here = dirname(fileURLToPath(import.meta.url));
+  const foundation = readFileSync(join(here, '../../migrations/2026-07-10-workspace-foundation.sql'), 'utf8');
+  await admin.begin(async (tx) => { await tx.unsafe(foundation); });
+
+  partner = randomUUID(); org = randomUUID(); site = randomUUID(); deviceId = randomUUID();
+  legacyLocal = randomUUID(); cleanLocal = randomUUID(); smbSource = randomUUID();
+  const sfx = randomUUID();
+
+  await admin`INSERT INTO partners (id, name, slug) VALUES (${partner}, 'wsp-3472', ${`wsp-3472-${sfx}`})`;
+  await admin`INSERT INTO organizations (id, partner_id, name, slug)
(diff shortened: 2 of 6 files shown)


ToddHebebrand (changes requested):
The fix itself is right — both paths are now covered (POST via `sourceInputSchema.superRefine(validateSmbConfig)`, PATCH via the merged `smbConfigSchema.safeParse` plus `validateSourceState` on the merged state in the service), no helper reinvented, both new queries org-scoped, and the rejection tests assert the right shapes including the explicit `crawlDeviceId: null` flip passing.

**One must-fix before merge: legacy rows have no backfill.** Main allowed `local_profile` + `crawlDeviceId` (the old test fixture was literally that shape), so existing rows can carry a crawl device. `sourcesPage.ts:217` seeds the edit form from `editing?.crawlDeviceId ?? ''` and resubmits it, so editing such a row now 400s every time — it doesn't self-heal the way the PR body implies, and the operator can't even see why (see the nit below). Please add an idempotent one-line migration: `UPDATE workspace_sources SET crawl_device_id = NULL WHERE kind = 'local_profile' AND crawl_device_id IS NOT NULL;` (with the repo's `DO $$ ... GET DIAGNOSTICS ... RAISE WARNING` row-count wrapper so the cleanup count lands in the logs).

**Two nits, take or leave:**
- `sources.ts:263` — the merged-check failure returns a generic `Invalid request body`, discarding the specific zod message; anyone hitting the legacy-row case gets nothing actionable.
- `sourcesService.ts:57` — `if (input.kind !== 'smb_share') return;` is dead now that kind is a two-value union.

**Status:** waiting on the backfill migration; everything else is mergeable as-is.

bdunncompany:
Backfill added, both nits taken. Head is now `c18d29e2f`, rebased onto current main.

**Migration** — `ee/workspace/migrations/2026-08-17-workspace-local-profile-crawl-device-backfill.sql`, the `DO $$` / `GET DIAGNOSTICS` / `RAISE WARNING` shape you asked for. Predicate only matches rows still holding the forbidden shape, so re-applying is a no-op.

I put it behind a real-Postgres test rather than shipping SQL nobody exercises — three rows: a legacy `local_profile` carrying a device, a clean `local_profile`, and an `smb_share` whose device must survive. Control-verified against three wrong migrations:

- drop the `kind` filter (also wipes SMB crawl devices) → fails
- invert it to `kind = 'smb_share'` → fails
- `DELETE` the rows instead of clearing the column → fails

That third one is worth flagging because **my first version of the test passed it.** The helper returned `row?.crawl_device_id ?? null`, which collapses "row was deleted" and "column is NULL" into the same answer. It now throws on a missing row and the suite pins the row count, since clearing a column is not the same as removing an operator's source.

**One correction to the rationale, though the backfill is still right.** The web form does not resubmit the seeded crawl device for a legacy row. `sourcesPage.ts:214` renders the `crawlDeviceId` input only when `#formKind === 'smb_share'`, and `#formKind` is set from `source.kind` when Edit is clicked (`:147`). `#readForm` reads an absent field as `''` and maps it through `|| null`, so editing a legacy `local_profile` row submits `crawlDeviceId: null` — which passes validation and clears the stranded device rather than 400ing.

So that path self-heals. What does not:

- Any API caller that PATCHes such a row **without** the field at all. The merge falls back to `existing.crawlDeviceId` and 400s, and there is no body they can send through a partial update to fix it other than an explicit `crawlDeviceId: null`.
- The actual harm needs nobody to edit anything. While the shape sits in the table, `listForDevice` returns those rows to a device that should not see them. That is the reason to backfill rather than wait for each row to be touched.

**Nits:**
- `sources.ts` merged-check 400 now returns the specific zod issue instead of `Invalid request body`.
- Dropped the dead `if (input.kind !== 'smb_share') return;`.

On that second one, an adversarial review raised a fair objection I'll leave to you: with the guard gone, anything reaching `validateSourceState` with a `kind` outside the union — a raw row cast, an `as SourceInput`, a future third enum value — is now silently treated as `smb_share` and either wrongly rejected for lacking a device/UNC path, or wrongly accepted if it happens to have both. The DB enum and the route schema both constrain it today, so nothing real can hit it. If you'd rather have an explicit unsupported-kind error than rely on exhaustiveness, say so and I'll add it; I didn't want to widen the diff past what you asked for.

Gate: `ee/workspace` 37 files / 593 tests, the new integration test 2/2 against the live stack, `tsc --noEmit` clean, eslint clean on every changed file, `check-migration-naming` and `check-migration-immutability` both OK, and `pnpm install --frozen-lockfile` clean after the rebase.

ToddHebebrand (approved):
**Re-review — all three items landed.**

- **Backfill (the must-fix):** `2026-08-17-workspace-local-profile-crawl-device-backfill.sql` is exactly the shape asked for — idempotent predicate, `DO $$ ... GET DIAGNOSTICS ... RAISE WARNING` so the cleanup count lands in the logs. The integration test is better than I'd have asked for: real Postgres, pins predicate *direction* (smb_share rows keep their device), idempotence, and the row-count-unchanged assertion that distinguishes clear-column from delete-row. Confirmed the runner picks it up automatically — `readDiskMigrations` (apps/api/src/extensions/builtinExtensions.ts:60) readdirs the package migrations dir, so no registration step exists to miss.
- **Error message nit:** `sources.ts` now surfaces `mergedConfig.error.issues[0]?.message` — the legacy-row case is actionable now.
- **Dead branch nit:** removed in `validateSourceState` with the two-value-union comment. (The route-side `if (input.kind !== 'smb_share') return;` survives, which is fine — kind can be absent there on a partial parse.)

CI green. Merging.

PR #3689, opened 2026-08-17
chore(deps): bump pg and @types/pg in /e2e-tests (supersedes #3675)

ToddHebebrand:
**Reduced scope:** now carries only dependabot #3675 — pg 8.23.0 + @types/pg 8.21.0 in `e2e-tests/` (its own npm lockfile, regenerated with `npm install --package-lock-only`, 24-line diff). #3675 went CONFLICTING after this morning's sibling e2e lockfile merges and dependabot ignored the rebase command.

**Why the root-workspace bumps (#3677/#3679/#3680) were dropped:** the root pnpm-lock has accumulated resolution drift from dependabot's serial incremental merges, so any pnpm re-resolution (full install or scoped `pnpm update`) reconciles it wholesale — including moving stripe 22.3.2 → 22.5.0, which breaks the pinned `apiVersion: '2026-06-24.dahlia'` literal in three files and would silently shift the billing SDK inside a deps batch. Not a trade worth making for patch bumps. Those three go back to dependabot via `@dependabot recreate`, which does minimal per-package lock edits.

Supersedes and will close #3675.


bdunncompany:
Independently verified your lockfile before your CI finished, since main is red right now and the regenerated-vs-surgical question is worth settling with measurements rather than waiting.

`d2f29aab4`:

- `hono@4.13.2` appears exactly once under `packages:` (line 7340) and once under `snapshots:` (line 17907). That cross-section pair is the healthy shape.
- Recursive scan for repeated scalar keys within any YAML mapping, covering `importers:`, peer-suffixed keys like `foo@1.0.0(bar@2.0.0)`, and nested dependency maps: **0 duplicates**.
- `pnpm install --frozen-lockfile` on a clean worktree of your head: **exit 0**, and `git status --porcelain pnpm-lock.yaml` is empty afterwards, so the install did not quietly rewrite it.

For contrast, current `main` at `b62a28eea` has 2 duplicates (`packages > hono@4.13.2` and `snapshots > hono@4.13.2`) and fails the same command with `duplicated mapping key (8037:3)`.

So the regeneration does clear the break as a side effect, and your PR is the better fix: it removes the stale incremental entries entirely rather than just deleting the one duplicated block, and it lands the four bumps at the same time. **#3688 is redundant if this merges** — I'll close it the moment this lands. I'm leaving it open only as a 6-line fallback in case this run hits the ongoing GitHub Actions trouble (archive downloads are at roughly a 50% error rate, which is what failed two of my runner steps earlier for reasons unrelated to any diff).

One data point for the prevention question, since this is the third time this exact break has landed (#3407 was `hono@4.13.1`). Counting `hono@4.13.x` keys per section at each commit in this morning's run pins it precisely:

| Commit | PR | `packages:` | `snapshots:` |
|---|---|---|---|
| `e5ae00036` | #3676 expo-sdk group | `4.13.1` x1 | `4.13.1` x1 |
| `75d0abbab` | #3681 hono group | `4.13.2` x1 | `4.13.2` x1 |
| `67e441adb` | #3682 user-event | `4.13.2` x1 | `4.13.2` x1 |
| `9138bad94` | **#3683 eslint / linting group** | `4.13.2` **x2** | `4.13.2` **x2** |

#3681 was correct. #3683 broke it, because its branch was cut while the lockfile still held `4.13.1`, so merging it after #3681 had already moved to `4.13.2` appended a second copy of the block instead of reconciling with the existing one. Requiring branches to be up to date before merging lockfile-touching PRs, or a merge queue, would have caught that specific merge rather than relying on a repair afterwards.

Also noticed while reading `ci.yml`: `test-mobile` and `test-agent-race` are not in the `ci-success` `needs:` list, so if branch protection requires only `CI Success`, a failing mobile or race job would not block a merge. Relevant here since this PR moves `react-native-worklets` and the mobile manifests. Happy to file it separately if useful.

bdunncompany:
**Correcting my comment above — it is now stale, and acting on it would leave main red.**

I verified `d2f29aab4`. You have since pushed `219adcb40` ("reduce to the e2e-tests pg bump only"), which reverted the `pnpm-lock.yaml` regeneration. That regeneration was the thing that cleared the break.

Measured on the current head `219adcb40`:

- `pnpm-lock.yaml` is now **byte-identical to `main`'s**: both SHA-256 `287f5120d3de66bdb4f28063…`. This PR no longer touches that file at all. It changes only `e2e-tests/package.json` and `e2e-tests/package-lock.json`.
- `hono@4.13.2` is back to twice under `packages:` (lines 8033, 8037) and twice under `snapshots:` (19767, 19769).
- `pnpm install --frozen-lockfile` on a clean worktree of `219adcb40`: **exit 1**, `ERR_PNPM_BROKEN_LOCKFILE ... duplicated mapping key (8037:3)` — the same failure main has.

So **merging this will not turn main green**, and this PR's own CI should fail on every JS job at `Install dependencies` for the same reason, independent of anything in the diff.

`e2e-tests` is outside the pnpm workspace and carries its own npm lock, so the two changes do not overlap: #3688 fixes the pnpm lockfile, this fixes the e2e-tests npm one. They can both land in either order.

Retracting what I said earlier: **#3688 is not redundant.** It is currently the only open change that makes `--frozen-lockfile` succeed — 6 deleted lines, both removed blocks asserted byte-identical to the kept copy before deletion, no regeneration, and CI on it completed green (41 success, 1 skipped, 0 failures).

Sorry for the churn on this thread; I should have re-read the head before asserting supersession rather than after.

PR #3703, opened 2026-08-17
fix(web): confirm-gate single-device reboot from the devices list (#3698)

### LINKED ISSUE #3698: [Web] Devices list kebab reboots a machine on one click — no confirmation, while the identical action on the device detail page is confirm-gated
## Description

On `/devices`, the row kebab (**Device actions** → **Reboot**) fires a reboot immediately on a single click. There is no confirmation step. The same action on the **device detail** page is gated behind a confirm modal.

So whether rebooting a production server asks "are you sure?" depends purely on which screen you happen to be on. The list view — the one where rows are dense, adjacent, and easy to mis-click, and where the kebab sits right next to `Run Script` and `Wake` — is the unguarded one.

Repro on a clean local stack on `main` (34d0234f5):

1. `/devices` → click the **Device actions** kebab on any row → **Reboot**.
2. A reboot is queued immediately. `POST /api/v1/devices/{id}/commands` → **201 Created**.
3. Confirmed in the DB:

```
 type   | status  | created_at
--------+---------+----------------------------
 reboot | pending | 2026-08-17 20:59:56.987491
 reboot | pending | 2026-08-17 20:59:43.973980
```

Both rows are from two stray clicks during a QA pass — nothing asked to confirm either one.

Worse for an **offline** device: the command sits `pending` and fires whenever the machine next reconnects, so the reboot can land minutes or hours later, on a box the tech has since stopped thinking about.

To be clear about what *does* work: the toast is fine and the copy is genuinely good — `Reboot queued — will run when e2e-macos.local reconnects`. The gap is purely the missing confirmation.

## Root Cause

The list-page menu item calls the action handler directly — `apps/web/src/components/devices/DeviceList.tsx:2449-2452`:

```tsx
onClick={() => {
  onAction?.("reboot", device);
  setRowMenuOpenId(null);
}}
```

`onAction` → `handleDeviceAction` (`apps/web/src/components/devices/DevicesPage.tsx:664-670`) → `sendDeviceCommand(device.id, action)`. The only guards are an in-flight check and `isCommandQueueable(device.status)` — neither is a user confirmation.

The detail page intercepts the same action into a modal first — `apps/web/src/components/devices/DeviceActions.tsx:227-240`:

```tsx
const handleAction = async (action: string) => {
  if (
    action === "reboot" ||
    action === "reboot_safe_mode" ||
    action === "shutdown" ||
    ...
  ) {
    setModalType(action);
    closeMenus();
    return;
  }
```

with copy already written at `DeviceActions.tsx:131-139` and a shared `ConfirmDialog` at `:421-427`. The list view simply never reaches it.

## Proposed Fix

Reuse the existing gate rather than inventing a second one: have the list-page kebab route destructive actions (`reboot`, `reboot_safe_mode`, `shutdown`, `decommission`, `clear-sessions`) through the same `ConfirmDialog` + `getModalConfig` copy the detail page uses, naming the target hostname in the prompt. That gets the two surfaces consistent and reuses the already-translated strings.

For an offline target the confirm should say the reboot will run on reconnect — the toast already words this well and the modal should match.

## Affected Files

- `apps/web/src/components/devices/DeviceList.tsx` (primary — unguarded menu item, ~2434-2456)
- `apps/web/src/components/devices/DevicesPage.tsx` (action handler, ~658-684)
- `apps/web/src/components/devices/DeviceActions.tsx` (the existing confirm gate + copy to reuse)

## Related

- #3616 — same class: a destructive dispatch reachable from one unlabelled/unguarded control.
- #2255 — step-up MFA before remote tools and Tier 3 destructive actions.

## Reported By

bdunncompany:
Closes #3698.

The row kebab on `/devices` fired a reboot on one click while the identical action on the device **detail** page has always been confirm-gated. Whether rebooting a production box asked "are you sure?" depended purely on which screen the tech was on — and the list is the wrong one to leave open, with dense adjacent rows and Reboot sitting beside Run Script and Wake.

It is worse offline: the command sits `pending` and fires whenever the machine next reconnects, so a stray click can land hours later on a box nobody is thinking about.

## Change

`handleDeviceAction` routes `reboot`, `reboot_safe_mode` and `shutdown` through a `ConfirmDialog` into a new `runDeviceAction`, which holds the original body unchanged.

Of those, **only `reboot` is currently emitted by the list kebab** — it emits `decommission, reboot, restore, run-script, settings, terminal, wake`. The other two are gated at the shared handler so a later menu addition inherits the guard instead of reintroducing this bug.

`lock` is deliberately not gated: reversible, does not disconnect the machine, and the detail page draws the same line.

The dialog reuses the existing `deviceActions.confirm.*` copy the detail page already uses, so the two screens read identically and **no new keys are needed in any of the eight locales**. It names the hostname, so a mis-click is recoverable rather than merely delayed. `shutdown` keeps `destructive`; the reboots stay `warning`, matching the detail page.

Confirming unmounts the dialog — that is what makes a double-click safe, for the reason the decommissioned-skip dialog documents a few lines above: kept mounted, both `ConfirmDialog`'s `disabled` and `actionInProgress` read a closure captured at render and are still false on the second click.

## Two things I want your call on

**1. `decommission` — knowingly left alone.** It is on the same kebab, it is the most destructive item there, and the detail page confirms it. But the list already guards it differently: a 5-second undo toast that defers the call. That reads as a deliberate product choice, not an oversight, and swapping an undo affordance for a modal is a UX decision I did not want to make inside a reboot fix. If you want true parity I will add it; if the undo pattern is intentional, this PR's "the list and the detail should agree" framing should probably say "except decommission, which agrees differently".

**2. A residual double-click concern I did not act on.** An adversarial review argued that unmount-on-confirm stops the *same* detached button firing twice but can let a genuine second physical click hit-test through to whatever sits underneath once the portal is removed, and that the safer shape is keeping the backdrop mounted with `isLoading` plus a synchronous ref guard. I did not change it, because the unmount behaviour here is the established pattern in this file with a written rationale, and the falsifier is a coordinate-level browser test rather than anything the unit suite can show. Worth its own issue if you think the concern is real.

## Tests

Six existing #2630 toast tests drove the row action directly and now click through the gate. I updated them rather than working around them, since they encoded exactly the unguarded behaviour this issue is about — they still distinguish sent-vs-queued after confirmation. Added a test that nothing reaches the API until the operator confirms and that the dialog names the machine.

Control run: with the gate removed, the new test fails and the six updated ones fail with it.

## Verification

Full `@breeze/web` suite **599 files / 5921 tests pass**. `astro check`: 0 errors, 0 warnings. eslint clean on both changed files. The 7 i18n suites (118 tests) pass, including locale parity and key usage — the dynamic key carries the `/* i18n-dynamic */` marker the key-usage gate expects.

### DIFF
--- a/apps/web/src/components/devices/DevicesPage.test.tsx
+++ b/apps/web/src/components/devices/DevicesPage.test.tsx
@@ -1158,9 +1158,32 @@ describe('DevicesPage — bulk agent commands gated on decommissioned only (#246
       render(<DevicesPage />);
       fireEvent.click(await screen.findByTestId(`row-reboot-${DEV_1}`));
 
+      // #3698: the row action is confirm-gated now, matching the device detail
+      // page. Nothing may reach the API until the operator confirms.
+      expect(vi.mocked(sendDeviceCommand)).not.toHaveBeenCalled();
+      fireEvent.click(await screen.findByTestId('confirm-device-action'));
+
       await waitFor(() => expect(vi.mocked(sendDeviceCommand)).toHaveBeenCalledTimes(1));
       return vi.mocked(showToast).mock.calls.map(c => c[0]);
     }
+
+    // #3698: the reason this issue existed — the list fired immediately while
+    // the detail page confirmed. Pin both halves of the gate.
+    it('does not queue anything until the confirm is accepted', async () => {
+      const { sendDeviceCommand } = await import('../../services/deviceActions');
+      vi.mocked(sendDeviceCommand).mockResolvedValue({ command: {} } as never);
+      vi.mocked(fetchAllDevices).mockResolvedValue({
+        data: [{ ...rawDevice(DEV_1, 'host-alpha'), status: 'online' }],
+      } as never);
+
+      render(<DevicesPage />);
+      fireEvent.click(await screen.findByTestId(`row-reboot-${DEV_1}`));
+
+      // The dialog names the machine, so a mis-click on a dense list is
+      // recoverable rather than merely delayed.
+      expect(await screen.findByText(/host-alpha/)).toBeTruthy();
+      expect(vi.mocked(sendDeviceCommand)).not.toHaveBeenCalled();
+    });
 
     it('online device: reports the command as sent, naming the device', async () => {
       const toasts = await rebootDeviceWithStatus('online');
--- a/apps/web/src/components/devices/DevicesPage.tsx
+++ b/apps/web/src/components/devices/DevicesPage.tsx
@@ -125,6 +125,12 @@ export default function DevicesPage() {
   // silently discarded.
   type PendingDecommissionedSkip = { action: string; devices: Device[]; skippedCount: number; totalCount: number };
   const [pendingDecommissionedSkip, setPendingDecommissionedSkip] = useState<PendingDecommissionedSkip | null>(null);
+  // #3698: single-device destructive actions from the row kebab. The device
+  // DETAIL page has always confirmed these (DeviceActions.tsx); the list did
+  // not, so whether rebooting a production box asked "are you sure?" depended
+  // on which screen you were on — and the list is the dense, easy-to-mis-click
+  // one, with Reboot sitting next to Run Script and Wake.
+  const [pendingDeviceAction, setPendingDeviceAction] = useState<{ action: string; device: Device } | null>(null);
   const [settingsDevice, setSettingsDevice] = useState<Device | null>(null);
   // v2 chip bar seeds its filter from the URL hash so a filtered view is
   // shareable; the legacy DeviceFilterBar owns its own state and ignores it.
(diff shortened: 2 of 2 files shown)


ToddHebebrand (approved):
**Approved.** Reviewed with a focus on the rewrite path, and the diff dodges the usual trap by being strictly additive: `runDeviceAction` is the old `handleDeviceAction` body verbatim — wake's `WakeCommandError` handling, the decommission undo toast, and the #2630 sent-vs-queued logic all survive untouched. Verified variant/copy parity against `DeviceActions.tsx:131-156` (reboots `warning`, shutdown `destructive`, `hostname` param), the keys exist in all eight locales under the `devices` namespace, and the six updated tests kept their sent-vs-queued assertions — only the confirm click was inserted. The new hostname assertion is non-vacuous (fails under mutation).

**On your two questions:**
- **decommission:** leave it. The 5-second undo toast is a genuine — arguably stronger — guard, so it "agrees differently" by design. No code change wanted.
- **double-click hit-through:** real but coordinate-level, and it applies equally to the two pre-existing dialogs in this file. File it as its own issue rather than changing the established unmount pattern here.

**Nits, take or leave (no re-review needed):**
- `DevicesPage.tsx:665-670` — `CONFIRM_REQUIRED_ACTIONS` / `confirmKeyFor` are re-created every render; module scope would do.
- The grid-card reboot path is gated too (nice bonus), but has no test and the PR body undersells it — worth one line in each.

Merging once CI completes.

Issue #3616, opened 2026-08-16
[Web][API] Remediate dispatches uninstalls to up to 500 devices from one unlabelled icon — no confirmation, no count, no preview


ToddHebebrand:
Split out from #3381, where this mechanism caused 259 unintended software uninstalls across a self-hosted fleet.

**The defect:** the Remediate control on the Compliance Dashboard dispatches uninstalls to up to 500 devices from a single click, with no confirmation, no device count, and no preview of what will be removed.

`apps/web/src/components/software/ComplianceDashboard.tsx:555-566` renders it as an 8x8 wrench icon button whose only text is a `title` tooltip, sitting in a row of icon buttons directly beside Edit. `onClick` goes straight to `handleRemediate(policy)` (line 348), which POSTs `/software-policies/:id/remediate` with `body: JSON.stringify({})`.

**The empty body is the dangerous part.** With no explicit `deviceIds`, `apps/api/src/routes/softwarePolicies.ts:786-806` selects every device holding a `violation` row for that policy, capped at 500. In allowlist mode with `allowUnknown: false` — which is the default, and which the inventory-approve flow auto-creates — every installed item not matching a rule is an `unauthorized` violation. So the implicit target set is effectively the whole fleet, and the operator has no way to see that before clicking.

**`enforceMode` does not gate this path, by design.** The comment at `softwarePolicies.ts:814-824` is explicit: "`enforceMode`/`autoUninstall` authorise UNATTENDED remediation rather than a deliberate operator action." That reasoning is defensible for the *authorization* question — the route is behind the execute permission and MFA. But it means the two checkboxes an operator reasonably reads as the safety interlock ("nothing gets uninstalled until I arm this") do not apply here at all, and nothing in the UI communicates that.

**What it needs:**

A confirmation step that states the blast radius in the operator's terms before anything is queued — the device count, the policy name and mode, and a sample of what will actually be uninstalled. The count is already computable server-side from the same query the route runs, so a dry-run/preview response is cheap.

The count should come from the server, not the client's current filter view. Part of what made #3381 so disorienting is that the operator was looking at one screen while the route independently resolved a 500-device set.

Give the control a text label. A destructive fleet action should not be an unlabelled icon adjacent to Edit.

Consider requiring typed confirmation (policy name, or the device count) above some threshold, the way the org-deletion flows already do.

[...]

PR #3709, opened 2026-08-18
fix(web): batch bulk vulnerability actions to the server's 200-id cap (#3694)

### LINKED ISSUE #3694: [Web] Bulk CVE remediation sends the whole selection in one POST — >200 findings always fails on the deviceVulnerabilityIds cap
**Reported by tim104979 in discussion #2598** (v0.105.1): "Remediate findings — 576 findings" fails with `deviceVulnerabilityIds: Too big: expected array to have <=200 items`.

**Root cause:** UI/API mismatch, no chunking anywhere.

- API caps the array at 200: `apps/api/src/routes/vulnerabilities.ts:97` — `deviceVulnerabilityIds: z.array(z.string().uuid()).min(1).max(200)` (same cap at :114, :120, :131 for accept-risk / mitigate / tickets).
- Web sends the entire selection in one POST with no chunking and no selection cap: `apps/web/src/lib/api/vulnerabilities.ts:240-247` (`remediateVuln`), invoked from `CveDrawer.tsx:373` and `SoftwareGroupDrawer.tsx:392` with raw `selectedIds`. Nothing in the drawers or `VulnBulkActionModal.tsx` limits or batches.

**The server cap is reasonable and should stay:** `remediateVulnerabilities` (`services/vulnerabilityRemediation.ts:283+`) loops per finding doing ~5 queries plus `queueCommandForExecution` inside one synchronous HTTP request — 200 is a request-size bound, not a product limit.

**Fix:** client-side chunking in the web layer (sequential 200-item batches with aggregate result reporting), applied to all four capped endpoints, not just remediate. Partial-success surfacing matters — if batch 2 of 3 fails, the operator must see which findings never scheduled.

bdunncompany:
Closes #3694. Fixes the 576-finding failure @tim104979 hit in #2598.

## Change

`runChunked` batches `deviceVulnerabilityIds` to the server's 200 cap, **sequentially** — parallel batches would multiply exactly the per-finding work (~5 queries + a command enqueue each) that the cap exists to bound.

At or under the cap it is a single request, byte-identical to before, with `runAction` still owning the toast. Above the cap `runAction` is called without `successMessage` so it stays quiet and one merged toast replaces N.

Applied to `remediateVuln`, `bulkAcceptVulnRisk`, `bulkMitigateVulns`.

## Two things I want your ruling on

**1. Remediation retries are not idempotent — this PR does not fix that.**

An adversarial review flagged it and I confirmed it: `remediateVulnerabilities` queues a fresh `install_patches` through `queueCommandForExecution` with no finding/action idempotency key. So if batch 2 of 3 fails and the operator retries the same selection, the 200 findings that already succeeded queue a **second install** on those devices. The drawers keep the original selection and re-enable submit on failure, so a blind retry is the natural next click.

Before this PR that hazard could not occur, because >200 failed atomically every time. It also meant the feature was unusable for the reported case. So this trades "always fails" for "works, with a partial-failure retry hazard", and I did not want to make that call for you.

What I did do client-side: the partial-failure copy names the confirmed lower bound and tells the operator to reload before retrying, rather than inviting the blind retry. Accept/mitigate are state-writes and tolerate retries; remediation is the one that doesn't.

A real fix is server-side — an idempotency key on `(deviceVulnerabilityId, action)`, or having the route skip findings with an active `install_patches`. Happy to take that as a follow-up if you want it before this lands.

**2. Ticket creation is deliberately NOT chunked**, which departs from the issue's "all four capped endpoints".

The route builds a per-request `byOrg` map and calls `createTicket` once per represented org, with no cross-request correlation or dedup. Three batches can therefore create up to three tickets for an org appearing in all three — trading a loud error for silently wrong data. It now fails fast with copy explaining the limit instead of the raw zod message. Raising the cap for that one route, or letting it append to an existing ticket, is your call.

## On the partial-failure message

The first version of this said "the remaining N were not scheduled". That was an overclaim and I've corrected it: `remediateVulnerabilities` queues each command **before** the awaited event publication, so a throw there leaves work queued that no response reported, and a lost HTTP response is ambiguous the same way. The merged total is now stated as a confirmed lower bound — "At least N scheduled" — never an exact applied count.

## Tests

11 cases, control-verified against three separate mutations, each of which fails the suite:

- disable chunking → 4 fail (including the 576 → 200/200/176 split)
- mute the single-batch toast → the parity test fails
- make the batches concurrent → the sequencing and stop-on-failure tests fail

That second and third mutation both passed my original suite; the review caught them and they're now covered. The success mock also echoes the real batch size — a fixed 200 previously let the 176-id final batch report 200, and the suite blessed an impossible total of 600.

## Verification

Full `@breeze/web` suite **601 files / 5943 tests pass**. `astro check`: 0 errors, 0 warnings. eslint clean on both changed files.

### DIFF
--- a/apps/web/src/lib/api/vulnerabilities.chunking.test.ts
+++ b/apps/web/src/lib/api/vulnerabilities.chunking.test.ts
@@ -1,0 +1,150 @@
+// #3694: the API caps deviceVulnerabilityIds at 200 and the web layer used to
+// POST the whole selection, so any selection over 200 failed with
+// `Too big: expected array to have <=200 items` — reported against a real
+// 576-finding remediation.
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const fetchWithAuth = vi.fn();
+vi.mock('../../stores/auth', () => ({ fetchWithAuth: (...a: unknown[]) => fetchWithAuth(...a) }));
+
+const showToast = vi.fn();
+vi.mock('../../components/shared/Toast', () => ({ showToast: (...a: unknown[]) => showToast(...a) }));
+
+import { remediateVuln, bulkAcceptVulnRisk, bulkMitigateVulns, createVulnTicket } from './vulnerabilities';
+
+const ids = (n: number) => Array.from({ length: n }, (_, i) => `id-${i}`);
+const sentBatches = () =>
+  fetchWithAuth.mock.calls.map((c) => JSON.parse((c[1] as { body: string }).body).deviceVulnerabilityIds as string[]);
+
+beforeEach(() => {
+  fetchWithAuth.mockReset();
+  showToast.mockReset();
+  // A Response body is single-use: mockResolvedValue would hand the SAME object
+  // to every batch and the second .json() would throw. Build a fresh one per call.
+  // Respond with the ACTUAL batch size: a fixed 200 let the 176-id final batch
+  // report 200 and the suite blessed an impossible total of 600.
+  fetchWithAuth.mockImplementation((_url: string, opts: { body: string }) => {
+    const n = (JSON.parse(opts.body).deviceVulnerabilityIds as string[]).length;
+    return Promise.resolve(new Response(JSON.stringify({ scheduled: n, skipped: [], success: true, succeeded: n })));
+  });
+});
+
+describe('#3694 bulk id chunking', () => {
+  it('sends 576 findings as three batches of 200/200/176, none over the cap', async () => {
+    await remediateVuln(ids(576));
+    const batches = sentBatches();
+    expect(batches.map((b) => b.length)).toEqual([200, 200, 176]);
+    expect(batches.flat()).toEqual(ids(576));            // every id sent exactly once, in order
+    expect(Math.max(...batches.map((b) => b.length))).toBeLessThanOrEqual(200);
+  });
+
+  it('at or under the cap it stays a SINGLE request — the unchanged path', async () => {
+    await remediateVuln(ids(200));
+    expect(fetchWithAuth).toHaveBeenCalledTimes(1);
+    expect(sentBatches()[0]).toHaveLength(200);
+  });
+
+  it('emits ONE aggregate success toast for a multi-batch run, not one per batch', async () => {
+    await remediateVuln(ids(576));
+    const successes = showToast.mock.calls.filter((c) => (c[0] as { type: string }).type === 'success');
+    expect(successes).toHaveLength(1);
+    expect((successes[0][0] as { message: string }).message).toContain('576'); // 200+200+176, the real total
+  });
+
+  it('reports partial progress when a later batch fails, instead of failing silently', async () => {
+    fetchWithAuth
+      .mockImplementationOnce(() => Promise.resolve(new Response(JSON.stringify({ scheduled: 200, skipped: [] }))))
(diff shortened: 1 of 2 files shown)


bdunncompany:
CI note — both red checks here are pre-existing and unrelated to this diff, which touches two files under `apps/web/src/lib/api/` and adds no dependencies.

**`Security Scanning` — a CVE published after main's last scan.** `CVE-2026-40345` (HIGH) in transitive `deepmerge-ts@7.1.6`, via `ee/workspace → mailparser → html-to-text@10.0.0`. Main's last run (`064d51cf8`, 22:31:17Z) was green; mine ran 04:56:14Z and Trivy pulls a fresh vuln DB each time. Filed as #3710 with the chain and a suggested override — it will red `main` on its next scheduled scan too.

**`Test Web` — the InvoiceWorkspace flake, recurring after #3219 was closed.**

```
FAIL src/components/billing/InvoiceWorkspace.test.tsx > InvoiceWorkspace >
  holds a header Issue click while the editor is dirty, then fires it when the save settles
Error: Test timed out in 5000ms.
```

600 of 601 files passed; that one test is the only failure. It is the exact test and title from #3219, which is currently closed. Locally this file is green — I ran the full web suite four times (601 files / 5943 tests) and it passed on runs 2, 3 and 4, with the same test failing once on run 1. So it is still intermittent rather than fixed.

I have not reopened #3219 since I do not know what closed it — if the fix was the file-wide one we discussed there, this is evidence it did not fully take; if it was closed as a duplicate, this belongs on whichever issue absorbed it. Happy to reopen with this run attached if that is the right home.

Re-running both jobs now. Nothing else to change here.

ToddHebebrand:
Reviewed and merging. The chunk boundary is right — server cap is `.max(200)` at all four routes (`apps/api/src/routes/vulnerabilities.ts:97,114,120,131`) and `BULK_ID_LIMIT = 200` matches exactly, so 200 stays a single request and 201 splits `[200, 1]`. No off-by-one at the boundary that caused #3694.

**On the idempotency question you raised — ship it, and here's the reasoning.**

You're right that a partial failure followed by a blind retry can double-queue installs on devices from an already-succeeded batch. I'm accepting that for now because the alternative today is strictly worse: before this PR, anything over 200 findings failed atomically, so the 576-finding case simply had no working path at all. Trading "always fails" for "works, with a retry hazard the copy warns about" is the right direction.

Two things make it tolerable rather than reckless. You stop on the first failed batch instead of pressing on, so the damage is bounded to what already succeeded. And the toast is worded as a lower bound — "At least N scheduled" — rather than an exact count, which is honest about what the client actually knows given `remediateVulnerabilities` queues commands before the awaited event publish.

The real fix is a server-side idempotency key so a replayed batch collapses instead of re-queueing, and that belongs in the API rather than bolted onto the client's chunk loop. Filed as a follow-up so it doesn't get lost.

Worth noting this lands next to two open blast-radius issues on the same surface — #3616 (Remediate dispatches to up to 500 devices from one unlabelled icon, no confirm) and #3381. Bulk remediation is accumulating enough sharp edges that it deserves a consolidated pass rather than more point fixes.

Nit, not blocking: a failed batch stacks two error toasts (runAction's per-batch failure plus runChunked's partial-progress one). Intentional per your comments and not misleading, just noisy.

PR #3726, opened 2026-08-19
fix(mobile): echo the server's CSRF token so the session can actually refresh (#3723)

### LINKED ISSUE #3723: [Mobile] Session cannot refresh: client sends a fixed CSRF header while the server sets a random CSRF cookie
## Summary

The mobile client sends a fixed CSRF header value while the server issues a random CSRF cookie, so `POST /auth/refresh` fails CSRF validation whenever the native cookie store actually works. With a 15-minute access token that means the app returns to the login screen roughly every 15 minutes.

Observed on a self-hosted deployment: after signing in, `/api/v1/mobile/*` requests begin returning **401** and the user is sent back to username/password + TOTP.

## The mechanism

Mobile sets the CSRF header to the literal `1` on every non-GET request:

```ts
// apps/mobile/src/services/api.ts
const CSRF_HEADER_NAME = 'x-breeze-csrf';
const CSRF_HEADER_VALUE = '1';
```

Login sets a **random** CSRF cookie alongside the refresh cookie (`setRefreshTokenCookie`, `randomBytes(32).toString('hex')`).

`validateCookieCsrfRequest` then has two paths:

```ts
const csrfCookie = getCookieValue(c.req.header('cookie'), CSRF_COOKIE_NAME);
if (!csrfCookie) {
  // Backward compatibility for non-browser clients that do not expose cookie APIs.
  if (csrfHeader === '1' && !isBrowserSignal) {
    return null;                       // ← passes only when the cookie is ABSENT
  }
  return 'Missing CSRF cookie';
}
if (!safeCompareTokens(csrfHeader, csrfCookie)) {
  return 'Invalid CSRF token';         // ← '1' vs random  →  refresh rejected
}
```

The `'1'` compatibility bypass is reachable **only when the CSRF cookie is missing**. React Native's fetch requests cookie handling explicitly (`credentials: 'include'`) and the native networking layer maintains a cookie store, so once that store returns the CSRF cookie the client takes the second branch and fails.

The bypass is therefore self-defeating: it exists for clients that "do not expose cookie APIs", but the client it targets *does* receive the cookie — it simply cannot read it to echo the value back.

## Why the session cannot recover

`resolveRefreshToken` accepts the refresh token from the cookie only:

```ts
export function resolveRefreshToken(c: Context): string | null {
  return getCookieValue(c.req.header('cookie'), REFRESH_COOKIE_NAME);
}
```

There is no body, `Authorization`, or header fallback, and `toPublicTokens()` deliberately omits the refresh token from the login JSON, so a native client has nothing to persist. The mobile app stores no refresh token in SecureStore (its keys are the auth token, user, biometric flag, server URL, onboarding flag, approver credential id, the approvals cache, and an install id).

So when refresh fails there is no second path back to a session — only a full re-login.

## Suggested fix

Either would work; the first is smaller:

1. **Let the client echo the CSRF cookie.** Drop `HttpOnly` on the CSRF cookie (it is already a non-secret double-submit token), and have mobile read and send its value instead of `1`.
2. **Accept a refresh token from the request body or an `Authorization` header** for non-browser clients, and persist it in SecureStore. This also removes the dependency on native cookie persistence across cold launches, which is documented as unstable in React Native.

## What would disprove this

A packet capture from a cold app launch showing `/auth/refresh` sent **without** the CSRF cookie (in which case the `'1'` bypass applies and the failure lies elsewhere), or evidence that the native store never returns that cookie on either platform.

I have not captured the request cookies on-device — this is a static reading of the two code paths plus the observed 401s. Happy to instrument a build and attach a capture if that would help.

## Environment

- Self-hosted, `JWT_EXPIRES_IN=15m`, `REFRESH_TOKEN_EXPIRES_IN=7d`
- iOS, app built from `main`

bdunncompany:
Fixes #3723.

The mobile client sends a fixed `x-breeze-csrf: 1` on every write, while the server issues a **random** `breeze_csrf_token` cookie at login. `validateCookieCsrfRequest` accepts the literal only when that cookie is **absent**:

```ts
if (!csrfCookie) {
  if (csrfHeader === '1' && !isBrowserSignal) return null;   // no-cookie path
  return 'Missing CSRF cookie';
}
if (!safeCompareTokens(csrfHeader, csrfCookie)) {
  return 'Invalid CSRF token';                               // '1' vs random
}
```

React Native's networking layer keeps its own cookie store and the client already asks for it with `credentials: 'include'`. So once login has happened the cookie **is** sent back, the fixed header fails the compare, and `/auth/refresh` is rejected. Because `resolveRefreshToken` reads the refresh token from the cookie only — no body or `Authorization` fallback, and `toPublicTokens()` deliberately keeps it out of the login JSON — there is nothing to retry with. The session ends at the access-token TTL and the user signs in again.

The compatibility path is self-defeating: it exists for clients that "do not expose cookie APIs", but the client it targets *receives* the cookie and simply never read it.

## The fix

The CSRF cookie is deliberately **not** HttpOnly (only the refresh cookie is), so the client can read and echo it. `services/csrfToken.ts` owns that token.

The parts worth reviewing:

**`cleared` is a distinct signal from `absent`.** Sign-out and a rejected refresh emit an empty `Max-Age=0` cookie. Treating that as "no information" would leave a stored token alive after the cookie jar dropped it — the client would then send a real-looking token with *no* cookie, which the no-cookie path rejects. That swaps one login loop for another, so a clear must drop our copy too.

**Last occurrence wins.** A clear and a re-set can be folded into one `set-cookie` value; resolving to the clear would discard the fresh token.

**Writes are serialised.** Two concurrent rotations could otherwise persist an older token than the cookie jar holds, which only shows up after a restart.

**Self-healing.** A response whose error mentions CSRF drops the stored token so the next request bootstraps. This covers the cases the client cannot otherwise detect — SecureStore survives an iOS reinstall while the cookie jar does not, so a stale token can outlive its cookie.

**`clearCsrfToken` does not swallow its keychain error**, so a failed wipe still surfaces through `clearAuthData`'s `SecureWipeError` rather than looking clean.

## What is not verified

Whether `response.headers.get('set-cookie')` is readable under Expo SDK 57's fetch. Expo installs `expo/fetch` as global fetch unless `EXPO_PUBLIC_USE_RN_FETCH=1`, and some fetch implementations do not expose `Set-Cookie`. I have not captured this on a device.

**The change is written so that does not matter for safety:** if the header is unreadable the client never learns a token, keeps sending `1`, and behaves exactly as it does today. The failure mode is "no improvement", not "worse". If a maintainer knows which fetch ships in SDK 57 and whether it surfaces that header, that would settle whether this is sufficient on its own.

## The alternative, if this is not sufficient

Accept the refresh token from a request body field or `Authorization` header for non-browser clients and persist it in SecureStore. That removes the dependency on native cookie persistence across cold launches entirely, and is the more robust fix — but it is a server-side auth change and felt like the maintainer's call rather than something to arrive unannounced in a PR. Happy to do it if preferred.

## Verification

Node 22.23.2 (matching `.nvmrc`), the two commands `test-mobile` runs:

```
pnpm test --filter=breeze-mobile   32 files, 339 passed | 3 skipped
pnpm exec tsc --noEmit             exit 0
```

17 tests cover the token module directly: cookie parsing (folded headers, percent-encoding, the near-miss `breeze_portal_csrf_token`), the cleared-vs-absent distinction, last-occurrence-wins, bootstrap fallback, cross-process persistence, concurrent-rotation ordering, self-heal, and the non-swallowed wipe failure.

## Review history

An adversarial review of the first revision found four defects, all fixed above. The most serious: the first version ignored the clearing cookie, which would have stranded a stale token after any 401 and produced a *different* permanent login loop. It also flagged the write-ordering race, the swallowed wipe error, and the iOS-reinstall case that motivated the self-heal.

### DIFF
--- a/apps/mobile/src/services/api.ts
+++ b/apps/mobile/src/services/api.ts
@@ -4,6 +4,12 @@ import * as Sentry from '@sentry/react-native';
 import { getServerUrl } from './serverConfig';
 import { getOrCreateInstallationId } from './installationId';
 import { fetchWithTimeout } from './fetchWithTimeout';
+import {
+  applyCsrfSignal,
+  forgetCsrfToken,
+  getCsrfHeaderValue,
+  readCsrfCookie,
+} from './csrfToken';
 
 export const FALLBACK_API_BASE_URL =
   process.env.EXPO_PUBLIC_API_URL || 'http://localhost:3001';
@@ -215,7 +221,10 @@ async function requestWithPrefix<T>(
   }
 
   if (!['GET', 'HEAD', 'OPTIONS'].includes(method)) {
-    (headers as Record<string, string>)[CSRF_HEADER_NAME] = CSRF_HEADER_VALUE;
+    // Echo the server's double-submit token. Sending the fixed bootstrap value
+    // once the CSRF cookie exists fails `safeCompareTokens` server-side and
+    // rejects every refresh, which ends the session at the access-token TTL.
+    (headers as Record<string, string>)[CSRF_HEADER_NAME] = await getCsrfHeaderValue();
   }
 
   // Always send the per-install id so the API can recognise this phone. This
@@ -247,6 +256,13 @@ async function requestWithPrefix<T>(
     timeoutMs
   );
 
+  // The server sets a fresh CSRF cookie on login and refresh, and clears it on
+  // sign-out or a rejected refresh. It is deliberately not HttpOnly so a native
+  // client can read and echo it. A `cleared` signal must drop our copy too —
+  // holding a token whose cookie is gone fails the server's no-cookie path,
+  // which accepts only the bootstrap literal.
+  await applyCsrfSignal(readCsrfCookie(response.headers.get('set-cookie')));
+
   if (!response.ok) {
     const body = await response.json().catch(() => ({} as Record<string, unknown>));
     const code = typeof body.code === 'string' ? body.code : undefined;
@@ -254,11 +270,22 @@ async function requestWithPrefix<T>(
       const reason = typeof body.reason === 'string' ? body.reason : null;
       notifyDeviceBlocked(reason);
     }
+    // Self-heal a token the server will not accept. The stored value can
+    // outlive its cookie — SecureStore survives an iOS reinstall while the
+    // cookie jar does not — and if `set-cookie` is not readable on this runtime
+    // we would never have learned a good one. Forgetting it makes the next
+    // request bootstrap with the literal the no-cookie path accepts, so the
+    // client recovers instead of looping.
+    const message =
+      (typeof body.error === 'string' && body.error)
+      || (typeof body.message === 'string' && body.message)
+      || 'An error occurred';
+    if (/csrf/i.test(message)) {
+      await forgetCsrfToken();
+    }
+
     const error: ApiError = {
-      message:
-        (typeof body.error === 'string' && body.error)
-        || (typeof body.message === 'string' && body.message)
-        || 'An error occurred',
+      message,
       code,
       statusCode: response.status
     };
--- a/apps/mobile/src/services/auth.test.ts
+++ b/apps/mobile/src/services/auth.test.ts
@@ -16,9 +16,23 @@ vi.mock('@sentry/react-native', () => ({
   captureException: (...a: unknown[]) => sentry.captureException(...a),
 }));
 
+// The CSRF wipe goes through its owning module, not SecureStore directly, so it
+// is invisible to the deleteItemAsync assertions above and needs its own mock.
+const csrf = {
+  clearCsrfToken: vi.fn(),
+};
+vi.mock('./csrfToken', () => ({
+  // auth.ts imports the canonical key from here too, so the factory must supply
+  // it — otherwise the deletions entry is keyed `undefined` and the wipe-failure
+  // report names nothing.
+  CSRF_TOKEN_KEY: 'breeze_csrf_token',
+  clearCsrfToken: (...a: unknown[]) => csrf.clearCsrfToken(...a),
+}));
+
 beforeEach(() => {
   secureStore.deleteItemAsync.mockReset().mockResolvedValue(undefined);
   sentry.captureException.mockReset();
+  csrf.clearCsrfToken.mockReset().mockResolvedValue(undefined);
 });
 afterEach(() => vi.restoreAllMocks());
 
@@ -31,6 +45,31 @@ describe('clearAuthData', () => {
 
     const keys = secureStore.deleteItemAsync.mock.calls.map((c) => c[0]);
     expect(keys).toContain('breeze.applock.state.v1');
+  });
+
+  it('removes the CSRF token, so the next account cannot inherit a stale double-submit value', async () => {
+    // The token is account-scoped: a survivor hands the next account a value
+    // that fails every write. This assertion is deliberately separate from the
+    // deleteItemAsync checks — the wipe runs through clearCsrfToken(), so it
+    // would not show up there even if it were working.
+    await clearAuthData();
+
+    expect(csrf.clearCsrfToken).toHaveBeenCalledTimes(1);
+  });
+
+  it('reports the real CSRF key when its wipe fails, and still runs every other delete', async () => {
+    // The failure path is the one that matters: a surviving CSRF token is only
+    // discoverable if SecureWipeError names the key that actually failed. This
+    // also pins the key to the module's exported constant, so renaming it there
+    // cannot silently leave this reporting a stale literal.
+    csrf.clearCsrfToken.mockRejectedValue(new Error('keychain locked'));
+
+    await expect(clearAuthData()).rejects.toBeInstanceOf(SecureWipeError);
+
+    const keys = secureStore.deleteItemAsync.mock.calls.map((c) => c[0]);
+    expect(keys).toContain('breeze_auth_token');
+    expect(keys).toContain('breeze.applock.state.v1');
+    expect(sentry.captureException).toHaveBeenCalled();
   });
 
   it('removes the auth token, the stored user, and the persistent approvals cache', async () => {
--- a/apps/mobile/src/services/auth.ts
+++ b/apps/mobile/src/services/auth.ts
@@ -3,6 +3,7 @@ import * as Sentry from '@sentry/react-native';
 import type { User } from './api';
 import { APPROVAL_CACHE_KEY, clearApprovalCacheOrThrow } from './approvalCache';
 import { APP_LOCK_STATE_KEY } from './appLockState';
+import { CSRF_TOKEN_KEY, clearCsrfToken } from './csrfToken';
 
 const TOKEN_KEY = 'breeze_auth_token';
 const USER_KEY = 'breeze_user';
@@ -120,6 +121,9 @@ export async function clearAuthData(): Promise<void> {
     // on the next launch and the leftover record waves it straight past the
     // lock. Deleting is the fail-secure direction — an absent record locks.
     { key: APP_LOCK_STATE_KEY, run: () => SecureStore.deleteItemAsync(APP_LOCK_STATE_KEY) },
+    // The CSRF token is account-scoped: leaving it behind would hand the next
+    // account a stale double-submit value that fails every write.
+    { key: CSRF_TOKEN_KEY, run: () => clearCsrfToken() },
   ];
 
   const results = await Promise.allSettled(deletions.map((d) => d.run()));
--- a/apps/mobile/src/services/csrfToken.test.ts
+++ b/apps/mobile/src/services/csrfToken.test.ts
@@ -1,6 +1,15 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 const store = new Map<string, string>();
+// csrfToken now reports read failures directly to Sentry (it cannot use
+// lib/errorReporting without an api -> csrfToken import cycle), and the real
+// @sentry/react-native is a Flow source this .ts-only suite cannot parse.
+const sentry = { captureException: vi.fn(), captureMessage: vi.fn() };
+vi.mock('@sentry/react-native', () => ({
+  captureException: (...a: unknown[]) => sentry.captureException(...a),
+  captureMessage: (...a: unknown[]) => sentry.captureMessage(...a),
+}));
+
 vi.mock('expo-secure-store', () => ({
   WHEN_UNLOCKED_THIS_DEVICE_ONLY: 'wutdo',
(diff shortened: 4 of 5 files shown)


bdunncompany:
Heads-up: this will conflict with #3746, and the resolution is additive.

Both branches start from the same `auth.ts` blob (`becd7670a0`) and both append an entry to the **same** `deletions` array in `clearAuthData`, at the same position right after `APPROVAL_CACHE_KEY` — #3746 adds `APP_LOCK_STATE_KEY`, this one adds the CSRF token. Both also add an import at the top of the file. Whichever lands second gets a conflict.

The two are complementary rather than competing: both are "register this key so it gets wiped on logout", and the merged result should keep both entries. Same reasoning in each case, that a survivor of a partial wipe is worse than a redundant delete.

Happy to rebase this one on top once #3746 lands, since that ordering leaves you nothing to resolve. Say the word if you'd rather take this first and rebase that instead.

No other overlap: I checked all six of my open mobile PRs against #3746 and #3747 and `apps/mobile/src/services/auth.ts` is the only shared file.

bdunncompany:
Rebased onto `b7776f55e` now that #3746 has landed, so there's nothing for you to resolve.

**The conflict resolved additively**, as expected — both branches appended to the same `deletions` array in `clearAuthData`. Kept both entries, app-lock first (main's order), CSRF after. Neither is conditional on the other and `Promise.allSettled` doesn't order them, so they're independent.

Two things came out of reviewing my own resolution that were worth fixing:

**1. I had duplicated the storage key.** My original carried a local `const CSRF_TOKEN_STORE_KEY = 'breeze_csrf_token'` in `auth.ts` purely to label wipe failures. That's a copy that can drift: rename the key in `csrfToken.ts` and `SecureWipeError`/Sentry would report a literal that no longer exists — and the report is the only way a surviving token is ever noticed. `csrfToken.ts` kept its real key private (`CSRF_TOKEN_KEY`, separate from the exported `CSRF_COOKIE_NAME` that happens to share the value), so I've exported it and `auth.ts` now imports the canonical constant instead of redeclaring it.

**2. The wipe had no failure-path test.** `toHaveBeenCalledTimes(1)` only proved wiring. Added a rejection test asserting `SecureWipeError` is thrown, the other deletes still run, and Sentry is notified — the failure path is the one that matters, since a surviving account-scoped token is only discoverable through that report.

**Verified, with controls rather than just a green run:**

- `pnpm test` 419 passed / 3 skipped / 35 files (417 before these two tests), `tsc --noEmit` clean
- removing the `{ key: CSRF_TOKEN_KEY, run: () => clearCsrfToken() }` line fails **exactly** the two new tests, exit 1
- removing `CSRF_TOKEN_KEY` from the test's module factory fails the whole `clearAuthData` suite, so the mock genuinely supplies it and isn't quietly `undefined`

**One thing I deliberately did not touch.** `Promise.allSettled(deletions.map((d) => d.run()))` builds its input eagerly, so a `run()` that threw *synchronously* would escape the aggregation and skip the SecureWipeError/Sentry path entirely. `clearCsrfToken` is `async`, so my entry can't trigger it, and it's pre-existing on main rather than something this PR introduces — flagging it rather than widening the diff. Happy to send it as a separate one-liner (`Promise.resolve().then(d.run)`) if you want it.

ToddHebebrand (commented):
The CSRF scheme here is right and I have no objection to the code. Cookie and header names line up with the server (`breeze_csrf_token` / `x-breeze-csrf`), the CSRF cookie is non-HttpOnly by design while the refresh cookie stays HttpOnly (`apps/api/src/routes/auth/helpers.ts:586,591`), and `safeCompareTokens`/`validateCookieCsrfRequest` are untouched — so double-submit semantics survive and nothing is weakened to make refresh work. The self-heal on a CSRF-rejected response is a good touch.

**I'm holding the merge on one question, because the body says `Fixes #3723` and I can't yet tell whether it does.**

Everything hinges on `response.headers.get('set-cookie')` (`apps/mobile/src/services/api.ts:261`) actually returning the CSRF cookie. The server appends the refresh cookie and the CSRF cookie as two separate `Set-Cookie` headers. Under the modern Fetch spec, `headers.get('set-cookie')` returns only the *first* value rather than a comma-joined string — and if RN's fetch behaves that way, the client would only ever see the refresh cookie and never learn the token. `readCsrfCookie` is written to scan a comma-joined string (`csrfToken.ts:52-68`), so it handles the joined case correctly, but it can't recover a header it was never handed.

You clearly thought about this — the comment at `api.ts:270-275` reasons about `set-cookie` being unreadable and the forget-then-bootstrap recovery. My concern is that the recovery path assumes the server's no-cookie branch accepts the bootstrap literal, but in this scenario the cookie *is* set server-side and merely unreadable client-side, so the server sees a real cookie against a bootstrap header. If that mismatches, we'd loop rather than recover — the same symptom #3723 reports, with more machinery behind it.

This is genuinely not answerable by reading code; it depends on RN's Headers implementation on each platform. Could you grab a device or emulator capture showing `set-cookie` visibility on a real refresh — ideally `getSetCookie()` alongside `get('set-cookie')`? If the header is readable, this merges as-is. If it isn't, the fix needs a different channel for the token and #3723 stays open.

No rush on my end, and nothing here needs code changes yet — I just don't want to auto-close #3723 on a fix we haven't seen work.

bdunncompany:
Good question to hold on, and I can answer most of it from source — the answer changes the shape of what still needs a capture.

**React Native's `fetch` is not the modern native Fetch.** `Libraries/Network/fetch.js` is four lines:

```js
require('whatwg-fetch');
export const fetch = global.fetch;
export const Headers = global.Headers;
```

so `Headers` is **`whatwg-fetch@3.6.20`**, and that polyfill's semantics are the *old* ones you were hoping for rather than the spec behaviour you feared:

```js
// whatwg-fetch 3.6.20, dist/fetch.umd.js:113-118
Headers.prototype.append = function(name, value) {
  var oldValue = this.map[name];
  this.map[name] = oldValue ? oldValue + ', ' + value : value;   // comma-joins
};
```

Response headers are built by `parseHeaders(xhr.getAllResponseHeaders())` (`:549`), which `append`s line by line. So **two `Set-Cookie` lines become one comma-joined value**, which is exactly the shape `readCsrfCookie` was written to scan. The "returns only the first value" behaviour doesn't apply here.

**Two corrections to the plan, though.** `getSetCookie()` **does not exist** in 3.6.20 — I grepped the dist bundle and there's no such method, so that half of the capture would just throw. And this is RN 0.87.0 in my worktree; main is now on 0.86.2 after #3747, so I'd want to confirm the polyfill version didn't move with it before treating this as settled for main.

**What I genuinely cannot answer from source, and where a capture is still the right call:** whether the *native* layer puts `Set-Cookie` into `responseHeaders` at all, and whether it collapses duplicates before JS ever sees them. RN's JS Network layer does no `set-cookie` filtering — I grepped `XMLHttpRequest.js` and all of `Libraries/Network/*.js` and there is none — but `getAllResponseHeaders()` only reformats whatever the platform handed up, and iOS `RCTHTTPRequestHandler` vs Android are separate implementations.

So the narrowed question is: **does `xhr.getAllResponseHeaders()` contain a `set-cookie` line on each platform, and does it contain one line or two?** That's a much smaller thing to capture than a full refresh trace, and it's the only remaining unknown.

I agree on not auto-closing #3723 until that's seen working. Want me to get that capture, or would you rather I drop the `Fixes #3723` from the body so this can merge on its own merits as a correct CSRF implementation, and leave #3723 open until it's demonstrated end to end? I'd lean toward the latter — the code is right either way, and the claim is what's unproven.

ToddHebebrand:
One more from a dedicated silent-failure pass, separate from the `set-cookie` question above.

**`getCsrfHeaderValue`'s bare catch can silently recreate the bug this PR fixes.** `csrfToken.ts:127-139`:

```ts
try {
  const stored = await SecureStore.getItemAsync(CSRF_TOKEN_KEY);
  if (stored) { cached = stored; return stored; }
} catch {
  // fall through to bootstrap
}
return CSRF_BOOTSTRAP_VALUE;
```

The catch treats *any* SecureStore failure the same as "no value stored" — a locked keychain, an OS-level error, a corrupted item. In that case we send the bootstrap literal while a real CSRF cookie exists on the request, which is exactly the mismatch that fails `safeCompareTokens`, so refresh gets rejected and the user is bounced. Silently, with no log, and looking identical to #3723's original symptom.

The `?? bootstrap` fallback is right for the genuine first-run case — the issue is only that an error is being folded into it. Distinguishing the two costs a line: report the error before falling through, so a keychain problem is diagnosable instead of presenting as a mystery logout. Worth pairing with a `forgetCsrfToken()` on that path if a read failure means the cached value can't be trusted either.

Not a blocker on its own — the merge is still held on the `set-cookie` question.

bdunncompany:
Fixed, and rebased onto current main so the stale `x/mod` Trivy red is gone too.

**The bare catch.** You're right that it recreates the bug this PR fixes. A read *failure* and "nothing stored" are different states, and folding them together means a real stored token coexists with us sending the bootstrap literal — genuine cookie against bootstrap header, `safeCompareTokens` fails, refresh rejected, user bounced. #3723's symptom with nothing in the logs to tell the causes apart. The failure is now reported; the `?? bootstrap` fallback is untouched for the genuine first-run case.

Test pins it: restoring the bare catch fails it. 518 pass, `tsc` clean.

**Two deliberate deviations from what you suggested, both with reasons:**

**Reported straight to Sentry, not via `lib/errorReporting`.** That module imports `services/api`, and `api.ts:12` imports this one — routing it there would create `csrfToken → errorReporting → api → csrfToken`. Same constraint I hit on #3727's bulk worker, where I solved it by carrying the errors out to the caller instead; that isn't available here because `getCsrfHeaderValue` returns a string on the request path.

**No `forgetCsrfToken()` on that path.** You suggested pairing it, and I think it's the wrong call here: `cached` is checked *before* the read, so there's no in-memory value to distrust, and a failed read tells us nothing about whether a stored token exists. A temporarily locked keychain would have us delete a perfectly good token and force an avoidable re-bootstrap. The self-heal on a CSRF-rejected response is the right place to discard, and it already exists. Happy to add it if you disagree — it's a one-liner either way.

**One cost worth flagging rather than hiding:** importing Sentry pulls react-native into a module that was previously RN-free (it only imported `expo-secure-store`), so `csrfToken.test.ts` now has to stub `@sentry/react-native` — the same stub `auth.test.ts` already carries. If you'd rather keep this module free of that dependency, the alternative is threading the failure out to `api.ts` and reporting there, at the cost of a more awkward signature.

The `set-cookie` question from your earlier comment is still the open item — my read of `whatwg-fetch@3.6.20` says the comma-join case is what actually applies, and the remaining unknown is whether the native layer surfaces `set-cookie` to `getAllResponseHeaders()` at all.

Issue #3724, opened 2026-08-19
[API][Mobile] Alert acknowledge takes 13-15s because the event publish awaits local handlers in the request path


bdunncompany:
## Summary

`POST /api/v1/mobile/alerts/:id/acknowledge` takes 13–15 seconds on a live deployment. The mobile client's request timeout is 15,000ms, so an acknowledge that **succeeds** server-side can abort client-side and be reported to the user as a failure — inviting a retry of an action that already completed.

## Measurements

From the API log on a self-hosted instance (216 devices, ~7,000 alerts), two consecutive acknowledges from the phone:

```
POST route=/api/v1/mobile/alerts/:id/acknowledge status=200 duration_ms=13361
POST route=/api/v1/mobile/alerts/:id/acknowledge status=200 duration_ms=15163
```

Both returned **200**. For comparison, on the same deployment in the same minutes:

```
GET /api/v1/mobile/alerts/inbox   duration_ms=654 … 1502
GET /api/v1/mobile/devices        duration_ms=134 … 789
```

So reads are sub-second-ish and this one write is 10–20× slower.

## What it is not

I checked the obvious candidates on that deployment before filing:

- **Not the database.** `EXPLAIN ANALYZE` of a PK read on `alerts` reports `Execution Time: 0.369 ms`. There are no non-internal triggers on `alerts`, and the table is small.
- **Not notification/webhook fan-out.** That tenant has **0** webhooks, **0** automations and **0** alert rules.
- **Not Redis.** `redis-cli --latency-history` reports 0–1ms.

## Where the time is spent

The handler awaits the event publish inside the request path, and `publish()` awaits local handlers serially:

- `apps/api/src/routes/mobile.ts` — the acknowledge handler awaits `publishEvent(...)`, then awaits `emitAlertStateFeedback(...)`
- `apps/api/src/services/eventBus.ts` — `publish()` awaits `xadd` + two `publish` calls, then unconditionally awaits `invokeLocalHandlers(event)`
- `invokeLocalHandlers` iterates handlers and `await`s each one in turn

It is neither conditional nor fire-and-forget, so any registered local handler's latency lands in the HTTP response.

The API also reports the symptom itself, repeatedly:

```
withDbAccessContext (scope=system) held a pooled connection in an open
transaction for 15439ms (>= 2000ms) — long enough that it likely did slow
non-DB work (Redis/HTTP/loops) or a slow query inside the context.
```

[...]

PR #3727, opened 2026-08-19
feat(mobile): swipe and multi-select to acknowledge alerts without waiting

bdunncompany:
Acknowledging from the phone was one alert at a time, each blocking on a write measured at **13,361ms and 15,163ms** on a real deployment (both returning 200 — see #3724). Clearing a screen of alerts was minutes of tapping and waiting.

Two ways through, sharing one optimistic path:

- **Swipe** an issue row to acknowledge it (`ReanimatedSwipeable`, `renderRightActions`, `rightThreshold={48}`).
- **Long-press to enter selection mode**, tick rows, acknowledge the batch. The long-press ticks the row you pressed, so the gesture is never a wasted step.

Rows disappear the moment the action starts and the request runs behind them.

## Why not `POST /alerts/bulk`

The obvious implementation — one bulk call for the selection — is wrong here, and this is the part I'd most like checked.

`/alerts/bulk` requires `alerts:write`. Partner Technician and Org Technician are seeded with `alerts:acknowledge` and **not** `alerts:write` (`db/seed.ts`), and `alerts.authz.test.ts` already asserts bulk returns 403 for exactly that case while single acknowledge succeeds. So a bulk implementation would 403 for the people most likely to be triaging from a phone, while working fine for an admin — the worst kind of bug to discover late.

This uses the per-alert **mobile** endpoint the app already uses, with bounded concurrency of **4**. Slower than one request, correct for every role, and the optimistic UI hides the latency either way.

The concurrency cap is deliberate rather than arbitrary: each acknowledge holds a pooled DB connection for its full duration (that is the substance of #3724), so firing a whole selection at once would pile long-lived connections onto the resource already under pressure. Four keeps a 20-alert batch to roughly a minute without acting as a load generator.

If bulk acknowledge were changed to accept `alerts:acknowledge`, this would collapse to a single request and I would happily follow up — the permission asymmetry looks unintentional given single acknowledge is deliberately available to Technicians.

## Honest outcomes

`acknowledgeAlerts` never rejects; it returns which ids landed and which did not, so:

- only the failures are restored
- the toast reports "Acknowledged 5, 2 failed" rather than claiming the full count
- successful ids stay hidden until the refetch completes, so the list cannot flash old rows back

`pendingAcks` **reference-counts** rather than using a set. Two overlapping operations naming the same id must both finish before the row returns — otherwise the first completion un-hides a row whose second request is still in flight.

The selection is **reconciled against currently visible rows** before submitting. Another operator may have acknowledged one, and a stale id would be silently skipped server-side while the UI claimed success.

## Interaction details

- Selection mode disables the swipe gesture: a half-swipe while ticking boxes would fire an action the user did not intend.
- In selection mode a tap toggles instead of navigating.
- The severity dot yields its slot to the checkbox so the control lands in the column the eye already tracks.
- A swipe commits without a confirmation step: acknowledging is reversible from the web UI, and the optimistic restore covers a refusal.

## Verification

Node 22.23.2 (matching `.nvmrc`), the two commands `test-mobile` runs:

```
pnpm test --filter=breeze-mobile   32 files, 335 passed | 3 skipped
pnpm exec tsc --noEmit             exit 0
```

`pendingAcks` is a pure module with direct coverage: optimistic hide/restore, per-id partial restore, reference counting across overlapping operations, non-mutation, selection reconciliation, and the label pluralisation.

Gesture wiring was checked against the installed package rather than assumed — `react-native-gesture-handler@3.1.0` ships `ReanimatedSwipeable`, and `App.tsx` wraps the navigator in `GestureHandlerRootView`, without which the gesture would silently never fire.

## Review history

An adversarial review of the first revision caught the permission problem above — the bulk-endpoint version passed every local test and would have failed only for Technician accounts in production. It also caught the unconditional success toast on a partial result, the set-vs-refcount race, and the stale-selection submission. All four are fixed here.


ToddHebebrand (changes requested):
Nice work on the rollback design — the reference-counted `pendingAcks` and restore-only-`failed` logic are right, and skipping `/alerts/bulk` for the permission mismatch is well argued. One blocker before this goes in.

**Must-fix: the ack request still uses the 15s default timeout, and this PR's own evidence says that isn't enough.**

`acknowledgeAlert` calls `request()` with no timeout override:

```
apps/mobile/src/services/api.ts:488
  export async function acknowledgeAlert(id: string): Promise<Alert> {
    const response = await request<MobileAlertRecord>(`/alerts/${id}/acknowledge`, {
      method: 'POST',
    });
```

so it falls back to `DEFAULT_FETCH_TIMEOUT_MS = 15000` (`fetchWithTimeout.ts:21`). The PR body cites measured writes of **13,361ms and 15,163ms, both returning 200**. The second one is past the client's abort deadline — so on exactly the latency profile that motivated this feature, a *successful* acknowledge aborts client-side, lands in `failed`, rolls back, and tells the operator "Could not acknowledge. Restored." while the server acknowledged it just fine.

That's the inverted-verdict failure mode: the operator believes an alert is unhandled when it is handled. `refresh()` papers over it on the next fetch, but the false toast is the thing users will act on.

What makes this clearly unintentional is that you already built the fix and didn't connect it — the `timeoutMs` parameter added to `request` (`api.ts:201`) and `coreRequest` (`api.ts:287`) is never passed by any caller. Threading it into `acknowledgeAlert` with something comfortably above the observed p99 (≥20s, maybe 30s given 15.1s was a real sample) closes it.

**Two nits, neither blocking:**

- `coreRequest` (`api.ts:284`) is exported but unused — leftover from the bulk-endpoint approach the body says you abandoned. Worth deleting so it doesn't read as a live path.
- There's no `isPending` guard before `onSwipeAcknowledge`/tap fires; you're relying on the row unmounting to prevent a double dispatch. Probably fine in practice since selection mode calls `exitSelection()` before awaiting, but a same-tick double gesture isn't structurally prevented.

Happy to merge once the timeout is wired through.

bdunncompany:
Blocker fixed, and you were right that I'd built the mechanism and not connected it.

**`ACK_TIMEOUT_MS = 30000`, passed by `acknowledgeAlert`.** That's the single choke point the bulk path fans out through, so both routes get it. 30s is ~2x the 15,163ms sample rather than a bare 20s, since 15.1s was a real observation and not a tail estimate.

Two tests pin it — one on the single path, one asserting all three writes in a bulk acknowledge carry the same deadline. **Control run: reverting `acknowledgeAlert` to the no-override form fails both, exit 1**; restoring returns to green. 337 pass (was 335), `tsc` clean.

**Both nits done.** `coreRequest` removed — `git log -S` confirms this PR introduced it, and it had exactly one reference (its own definition) against 15 for a control symbol; `API_CORE_PREFIX` still has live callers at `api.ts:336,369,383`.

**On the double-dispatch guard, my first attempt was wrong and I want to flag it rather than bury it.** I wrote a permanent per-row `dispatchedRef` latch. That breaks the rollback path this PR exists for: a failed acknowledge *restores* the row, and a latch that never resets would leave that restored row un-acknowledgeable by swipe for as long as React kept the instance alive — punishing the user for a failure that already cost them. Replaced with a 600ms window, which collapses the same-tick double you described without ever permanently disabling the gesture. `haptic.tap()` sits inside the guard so feedback only fires on a dispatch that actually happened.

**One thing I want to be straight about: 30s reduces this failure, it does not eliminate it.** A write that completes at 30,163ms, or completes server-side with the response lost, still rolls back an alert that was acknowledged. Making it actually safe needs an idempotent acknowledge plus reconciliation on ambiguous failure — refetch the alert before restoring it rather than trusting the client's verdict. That's a different and larger change than you asked for here, so I haven't smuggled it in; happy to file it separately if you think it's worth having. `refresh()` currently repairs the state, so what survives is the false toast, not corrupt data.

bdunncompany:
**Correcting my last comment — it described a fix that main had already made, and one of your nits would now break the build.** I rebased before pushing and found both.

**The blocker was already fixed on main, by my own other PR.** `ACKNOWLEDGE_TIMEOUT_MS = 45000` landed in #3720 at `29a3bd7d7`, merged **00:05:39Z — 41 seconds after your 00:04:58Z review**. You reviewed against a base that genuinely didn't have it. So my `ACK_TIMEOUT_MS = 30000` was a duplicate of an existing constant, and a weaker one: 45s with your #1105 reference and the server's own `held a pooled connection ... for 15439ms` log is better grounded than my 30s. **I dropped mine and kept main's.**

**The `coreRequest` nit is now stale and I'm glad I checked.** It was genuinely unused at your review time, but `tickets.ts:1` and `tickets.test.ts:13` import it as of #3716. Deleting it, which is exactly what my previous push did, **would have broken the build.** Kept, with the `timeoutMs` parameter threaded through so it matches `request`.

**What this PR actually contributes now:**

- The tests main was missing — one on `acknowledgeAlert`, one asserting all three writes in a bulk acknowledge carry the same deadline. Control: reverting the override fails both, exit 1. The constant existed with nothing pinning it.
- Your double-dispatch guard, as a 600ms window rather than a permanent per-row latch. A failed acknowledge *restores* the row, and a latch that never resets would leave that restored row un-acknowledgeable for as long as React kept the instance alive.

**Also worth flagging: the previous head never ran CI at all.** It showed zero check-runs and zero check-suites — `mergeable=CONFLICTING`, so GitHub couldn't build a merge ref and skipped the whole suite. It looked reviewed-and-passing when nothing had run. Now `MERGEABLE` at `19951dc17` with checks executing.

513 pass / 3 skipped / 44 files, `tsc` clean.

My earlier point stands unchanged: 45s reduces the false rollback, it doesn't remove it. A write completing at 45,163ms, or completing with the response lost, still rolls back an acknowledged alert. Idempotency plus reconciliation is the real fix, and still out of scope here.

ToddHebebrand:
Ran a dedicated silent-failure pass over this after the first review and it surfaced two more, both verified against the branch. Adding them to the existing bounce rather than opening a second round.

**The un-hide runs even when the refetch failed.** `SystemsScreen.tsx:207-210`:

```
setPendingAcks((p) => endAck(p, failed));
await refresh();
setPendingAcks((p) => endAck(p, acknowledged));
```

`refresh()` can't reject. `useSystemsData.fetchAll` catches its own error, reports it, sets `error: 'Failed to load systems data.'` and returns normally (`useSystemsData.ts:106-116`) — so the `await` always resolves and the third line un-hides the acknowledged ids unconditionally. On a flaky link the refetch fails, `data.alerts` stays stale, and the rows the user just acknowledged flash back looking unacknowledged — two lines under a comment that says "so the list cannot flash the old rows back."

The user has already seen "Acknowledged 6 alerts." Combined with the 15s timeout issue, the plausible field sequence is: success toast, rows return, then a generic "Couldn't refresh" banner. The operator can't tell which of the three states is true. #3719's banner means it isn't strictly silent, but it is genuinely misleading. Gate the third `endAck` on the refetch actually succeeding — have `fetchAll` signal outcome, or re-read `error` before un-hiding.

**The bulk worker discards the error entirely.** `api.ts:474-477`:

```
} catch {
  failed.push(id);
}
```

No `reportInternalError`, no log — just the id. So a 403, a 500, and a client-side abort are indistinguishable in the outcome, and the `reportInternalError('bulk-acknowledge')` in `SystemsScreen.tsx:213` never fires for them because they never propagate. That's also what makes the timeout bug hard to diagnose in the field: the telemetry that would show "these all aborted at exactly 15s" is thrown away. Capture the error before pushing to `failed`.

Same three-item list as before plus these two — the timeout wiring is still the blocker.

ToddHebebrand:
Last one, from a comment-accuracy pass. This one changes a UX decision rather than just a comment.

**`IssueRow.tsx:86` justifies skipping the confirm dialog on the grounds that acknowledging is "reversible from the web UI" — there is no unacknowledge route.** I grepped `apps/api/src` for one: the only `unacknowledge` hits are in `admin/abuse.ts` and the desktopWs files, both unrelated. `apps/api/src/routes/alerts.ts` has no un-acknowledge endpoint.

That matters because the rationale is load-bearing for the feature's core interaction. Swipe-to-acknowledge with no confirmation is a reasonable design *if* a misfire is undoable — on a phone, in a list, a stray swipe is not a rare event. If it isn't undoable, the calculus changes: either the gesture needs an undo affordance (a few-second "Undo" on the success toast is the standard answer and fits the optimistic model you already built), or the server needs the reverse route, or the gesture needs confirmation.

I'd suggest the undo-toast route — it preserves the "without waiting" feel that's the whole point of the PR, and it composes with the `pendingAcks` machinery already there.

**Two smaller comment issues on the same PR, both false rationales:**

- `SystemsScreen.tsx:228` / `pendingAcks.ts:69` say a stale ack is "silently skipped server-side." It isn't — the server returns HTTP 400 (`mobile.ts:904-906`), and the UI surfaces a failure toast. The comment describes a silent-skip that would justify not handling the case; the real behavior is already handled correctly.
- `IssueRow.tsx:59` says a double-open "double-decrements the refcount." The refcounting is symmetric, so it nets to zero — the actual cost is just a redundant request. Harmless, but it would send someone hunting a refcount bug that doesn't exist.

None of these are new blockers beyond the timeout item — but the "reversible" one is worth settling before this lands, since it's the premise the interaction rests on.

bdunncompany:
Both new findings confirmed. One is fixed; **the other I implemented, then backed out, because your suggested fix is worse than the bug** — and I'd rather show you why than quietly ship it.

**Fixed: the bulk worker discarding errors.** `BulkAckOutcome` now carries the causes and the caller reports them. Reported by the caller deliberately — `lib/errorReporting` imports from `services/api`, so reporting inline would make an `api → errorReporting → api` cycle. Test pins it; reverting the array fails it. You were right that this is what made the timeout invisible in the field.

**Not fixed, deliberately: gating the un-hide.** I built exactly what you asked for — `fetchAll` returning whether the refetch landed, `if (await refresh())` around the third `endAck` — and it introduces a worse failure:

`fetchAll` starts with `if (inFlight.current) return`. Acknowledging while any other refresh is in flight (push handler, realtime WS, tab-focus, pull-to-refresh) coalesces into it, so the gate reports no fresh read. **Nothing else releases those ids** — `endAck` for the acknowledged set only runs inside this `acknowledgeIds` call — so the rows stay hidden for the life of the screen. A briefly stale visible row is recoverable. A permanently concealed active alert is not, and this list is how the operator learns the alert exists.

There's a second hole in the same fix: `failedCount < 5` means "at least one of five slices arrived", not "alerts arrived". It returns true when devices and orgs land and `alerts` itself failed — un-hiding against stale alert data, which is the original bug with extra steps.

So the premise is wrong in a way that isn't obvious: **`refresh()` resolving isn't evidence of freshness, but neither is its return value.** The correct primitive is a fetch generation the `pendingAcks` map compares against, so a pending ack is released when a newer *alerts* snapshot lands regardless of which caller fetched it. That's a data-layer change rather than something at this call site. Happy to do it here or as a follow-up — say which.

I left both call sites commented with that reasoning so the obvious patch doesn't get re-attempted by the next person, including me.

**Comment corrections, all three confirmed:**
- "silently skipped server-side" — the per-alert route returns HTTP 400 (`mobile.ts:905`). The silent skip is real but belongs to `/alerts/bulk` (`alerts/alerts.ts:449-454`, `results.skipped++`), which this PR abandoned. The rationale outlived the approach — same root cause as the leftover `coreRequest` you flagged.
- "double-decrements the refcount" — symmetric, nets to zero; the cost is one redundant request.
- **"reversible from the web UI" — you're right, and I verified it.** There is no un-acknowledge route: every `unacknowledge` hit in `apps/api/src` is `admin/abuse.ts` or desktopWs, and `alerts.ts` has no reverse endpoint.

On the undo affordance: worth noting an undo toast can't work by reversing the acknowledge, for the reason you just gave. It works by **delaying the dispatch** — hide optimistically, hold the request for the toast's lifetime, and undo cancels before it fires. That needs no server route and composes with `pendingAcks`. It does mean the acknowledge is no longer instant, which trades against the "without waiting" goal. Your call whether that belongs in this PR.

514 pass, tsc clean.

ToddHebebrand:
Rulings on both, so you're unblocked.

**The un-hide gate: you were right to back it out, and right to say so rather than ship it quietly.** The failure you found is the worse one — a briefly stale visible row is recoverable, a permanently concealed active alert is not, and this list is the only place the operator learns the alert exists. The `inFlight` coalescing makes it not even an edge case: a push arriving while the operator swipes is an ordinary Tuesday. And your second hole is the sharper one — `failedCount < 5` answering "did anything arrive" rather than "did *alerts* arrive" would have un-hidden against stale alert data, which is the original bug wearing the fix's clothes.

Your read on the correct primitive is right too. `refresh()` resolving isn't evidence of freshness and neither is its return value; a generation stamped on the alerts snapshot and compared by `pendingAcks` is, because it makes coalescing harmless (the coalesced-into fetch still bumps it) rather than fatal.

**Filed as #3782 — do it there, not here.** It's a data-layer change to `useSystemsData`/`pendingAcks` and this PR shouldn't wait on it. Leave the call-site comments in; they're the reason the next person won't re-attempt the obvious patch.

**The undo: build it here, and your delay-dispatch shape is the right one.** You're correct that an undo toast can't work by reversing the acknowledge — there's no reverse route, which is the whole finding. Hide optimistically, hold the request for the toast's lifetime, cancel before it fires. No server route, composes with `pendingAcks`.

On the trade you flagged — that the acknowledge is no longer instant, against a PR whose title is "without waiting": I'll take it, and I don't think it costs much of the goal. "Without waiting" was about not making the operator tap through a detail screen and sit on a 15-second write, not about the row clearing in under a second. The gesture stays one swipe and the list stays responsive; the request just fires a few seconds later. What we'd be buying is that a stray swipe on a phone — not a rare event, in a list, one-handed — stops being unrecoverable.

Two things to get right in it:

- The pending window has to survive the operator leaving the screen. If unmount drops a held request, a swipe-then-navigate silently does nothing, which is a new silent failure in a PR that just closed several. Flush on unmount rather than cancel.
- Batch the multi-select case into one toast with one timer, not N. A per-row toast stack for a 30-alert selection is its own bug.

**Comment corrections: all three accepted, and thanks for chasing the `unacknowledge` question independently** instead of taking my grep for it. Your finding that the "silently skipped" rationale belongs to `/alerts/bulk` and outlived the approach is exactly the pattern worth naming — the code moved and the reasoning didn't.

Ship the undo and I'll merge.

bdunncompany:
Undo shipped in `399e2a9`, delay-dispatch shape as you called it.

`undoAck.ts` is the pure part, 15 tests. One window at a time, so a 30-alert selection is one toast with one timer — a second acknowledge commits the batch already waiting rather than stacking. Every batch carries a token, because a timer, an Undo tap and an unmount can all fire against a window that has already closed, and a stale one must release nothing or it commits a batch the operator undid.

**Unmount flushes, per your note.** Dropping a held request would make swipe-then-navigate silently do nothing, which is precisely the kind of silent failure this PR spent its life closing.

**Two wiring bugs worth naming, because this app's tests structurally cannot see them.** Mobile has 45 `.test.ts` and zero `.test.tsx` — no component-test infrastructure — so screen behaviour is covered by extracting pure modules, the convention `pendingAcks` / `heroCopy` / `mergeSystemsResults` already follow. Adversarial review caught both:

- The dispatch ran **inside a `useState` updater**. React may invoke an updater more than once and keep only the returned value; StrictMode does it deliberately. So the state stayed correct while the acknowledge went out **twice** — on an irreversible write. The ref is now the source of truth, every transition runs synchronously in the handler, and the state update afterwards is pure.

- The handlers read the **current** token rather than the one owned by the callback that fired, so batch A's expiry could commit batch B. That defeated the exact guards `undoAck` exists to provide. The token is now bound at the render that owns the batch.

I'd rather flag that than let it read as clean: the pure module was right the whole time, and the wiring quietly disarmed it.

**On the trade you accepted** — the acknowledge no longer being instant. Window is 5s. Success no longer raises its own toast, since the undo toast already said "acknowledged" when the rows vanished and a second one seconds later announces something the operator has moved on from. Failures still surface.

**Not addressed here, deliberately:** the un-hide still releases without proof of a fresh alerts snapshot. That's #3782, and both call sites keep the reasoning comments so the obvious patch doesn't get re-attempted.

Gate: mobile `tsc --noEmit` clean, full `vitest run` 45 files / 529 passed / 3 skipped.

One thing I can't claim: I have not seen the toast render on a device. The tokens are real (`theme.bg3`, `theme.textHi`, `type.metaCaps`, `theme.approve`) and it typechecks, but layout and contrast are unverified by eye. Happy to adjust if it looks wrong to you.

ToddHebebrand (changes requested):
**The undo is right, and both wiring fixes are genuinely fixed** — I traced each one. `flushAllUndo` clears the ref before it dispatches (`SystemsScreen.tsx:363-374`), so a StrictMode double-cleanup or a timer racing the unmount releases nothing the second time; `flushUndo`/`cancelUndo` compare against the live batch (`undoAck.ts:71-85`) and the handlers close over the render-bound token via `key={undoBatch.token}`, so a stale expiry can't commit a newer batch; a second acknowledge flushes the earlier batch rather than dropping it (`undoAck.ts:55` → dispatch at `:329`); Undo un-hides through `endAck` and leaves nothing in `pendingAcks`; a failed delayed request still toasts (`:237-245`). Good.

**One must-fix, and it's the hold itself creating a new silent failure:** there is no `AppState` handling anywhere in the PR. iOS suspends JS timers when the app backgrounds, so swipe → immediately switch apps means the 5s timer never fires; if the OS reclaims the app while suspended, the acknowledge is never sent — after the toast already told the operator "acknowledged." Before this PR an in-flight request survived the ~30s background grace; an *unstarted* timer does not, so delay-dispatch is strictly worse here unless the window closes on background. That's the exact class of bug this PR spent its life closing.

**Fix:** subscribe to `AppState` and on `change` to anything non-`active` run the same `flushAllUndo` + dispatch path the unmount effect uses. The precedent is already in the app — `screens/approvals/useApprovalQueueSync.ts:56`. `inactive` counts too (the app switcher, a phone call): treat it the same as `background`, don't wait for the second transition.

**Nits, not blocking:** `UndoToast.tsx:48` — `visible` is always literal `true` at the only call site, so the prop and its guard are dead. And both wiring bugs you found were *wiring*, which `undoAck.test.ts` structurally cannot see; I'll take your word that there's no component-test infra today rather than ask for one here.

Once the AppState flush lands this merges — nothing else is open. #3782 stays deferred as agreed.

bdunncompany:
I implemented the `AppState` flush, then reviewed it adversarially before pushing — and I'm holding it, because **it does not actually fix the failure you identified.** It relocates it. Two findings, both verified in source, plus a question about the `inactive` boundary where the evidence runs against your instruction.

### Why the flush as specified is not sufficient

`acknowledgeAlerts` is not one request. It's four workers draining an in-memory queue of **per-alert** calls (`api.ts`, `ACK_CONCURRENCY = 4`), and its own comment measures each request at **13–15s**, putting a 20-alert batch at "roughly a minute".

So on background with 20 selected:

1. `flushAllUndo` atomically removes **all 20** ids from `undoRef`.
2. Four requests start. Sixteen ids exist only in that in-memory queue.
3. The process is suspended or reclaimed.
4. The timer, later `AppState` events and the unmount cleanup all correctly see an empty ref and refuse to re-take the batch.

The tail is unrecoverable, and the toast already told the operator all 20 were acknowledged. That's the same silent failure the must-fix was about — the atomicity that makes the flush safe against double-send is exactly what makes the loss unrecoverable.

Worth connecting: the per-alert path exists because `POST /alerts/bulk` requires `alerts:write` while Technicians are seeded `alerts:acknowledge` only. So "just send one bulk request" — the obvious durable fix — is blocked by a **role-seeding** issue, not a client one.

### Second defect: resume leaves a lying UNDO button

The flush clears `undoRef` but not the rendered `undo` state, so after a transient `inactive` the toast is still mounted offering UNDO. Tapping it runs `cancelUndo` against an already-empty ref: it cannot retract the request or restore the row.

And `acknowledgeAlerts` **never rejects** for per-id failures — it returns them in `failed`. So the `.catch` I wrote is not the reconciliation path. A failed alert stays hidden behind `pendingAcks` with no error surfaced, and survives pull-to-refresh. My comment "there may be no UI left to update" conflated unmount with background; on `inactive` and after an ordinary resume there is very much UI to reconcile.

### The `inactive` question — evidence against your instruction

You asked for `inactive` to count, and not to wait for a second transition. I implemented that, but I think it's wrong, and this repo already contains the argument:

> `// The Face ID sheet itself flips the app to `inactive`. If that armed the lock, unlocking would re-arm it — an infinite biometric prompt loop...`
> — `services/appLockMachine.test.ts`, "inactive must never start the lock clock"

Apple's documented order for the case you're defending against — switch apps, Home — is `willResignActive` → `didEnterBackground` → snapshot → *then* possibly suspend. `background` always precedes suspension, so waiting for it doesn't miss the hazard. Meanwhile `inactive` alone fires for a call banner, Control Centre, Notification Centre, Face ID and system alerts, none of which suspend the app — and under my current implementation every one of them **irreversibly commits the acknowledge** while leaving a UNDO button on screen that can't undo it.

**My recommendation: commit on `background`, not `inactive`.** If you still want `inactive`, I'd want a device trace showing suspension or reclamation without a preceding `background` callback, because Apple documents that can't happen.

### What I'd like to land, if you agree

1. Commit boundary on `background`.
2. Split the mounted-lifecycle flush from the unmount flush: the mounted one must close the ref **and** the rendered undo state atomically, then reconcile `acknowledged` / `failed` / `pendingAcks` and surface errors.
3. Durability for large batches is the real fix and is architectural — a persisted outbox drained on resume, or making the bulk route usable by the `alerts:acknowledge` role. Both are bigger than this PR. **Until one exists, the honest options are to cap what a single undo batch may hold, or to dispatch immediately and drop the undo window for multi-select.**

Tell me which of those you want and I'll build it. I didn't want to push something that reads correct and still loses sixteen of twenty acknowledges.

Unrelated and already done: the dead `visible` prop on `UndoToast` is removed. Mobile typecheck clean, 559 mobile tests pass — though note `vitest.config.ts` includes only `.ts`, deliberately excluding `.tsx`, so none of the lifecycle behaviour above is reachable by the current test setup.

ToddHebebrand:
**Ruling — you're right on both counts, and thank you for holding it instead of pushing something that reads correct.**

**Commit boundary: `background`, not `inactive`.** I verified your evidence: the `appLockMachine.test.ts:183-186` precedent is exactly on point (Face ID itself fires `inactive`), and Apple documents that suspension never happens without `didEnterBackground` first. My `inactive` instruction was wrong — an irreversible commit on a Control Centre pull with a dead UNDO button still on screen is strictly worse than the hazard it was guarding against. Commit on `background`.

**Durability: build the real fix — make `POST /alerts/bulk` usable by `alerts:acknowledge`, and flush as ONE bulk request.** Of your three options, the role-seeding one is smaller than you think and it's the only one that isn't a workaround:

- The gate is per-action, not per-route: `alerts.ts:393` applies `requireAlertWrite` to the whole route, but a bulk *acknowledge* is just N single acknowledges, which the Technician role can already perform one at a time via `alerts.ts:598` (`requireAlertAcknowledge`). Granting bulk-acknowledge to `alerts:acknowledge` adds zero capability — it removes a rate limit on a permission the role already holds. `suppress`/`resolve`/`dismiss` stay behind `alerts:write`.
- Concretely: keep `requireScope`, drop the route-level `requireAlertWrite`, and gate inside the handler after `zValidator` — `action === 'acknowledge'` needs `alerts:acknowledge` OR `alerts:write`; everything else needs `alerts:write`. The route already does org scoping and `filterAlertsBySiteScope`, so there's no scope surface change. Extend `alerts.authz.test.ts` with the matrix: acknowledge-only role can bulk-acknowledge, gets 403 on bulk-resolve/suppress/dismiss.
- Then the client's flush (timer expiry, `background`, unmount) sends **one** request. A single dispatched fetch either lands whole or fails whole — no 16-of-20 tail stranded in an in-memory queue — and it deletes the reason the 4-worker queue exists. Keep `acknowledgeAlerts` as the fallback only if you find a mobile-session reason the core route is unreachable (verify the mobile token actually reaches `POST /alerts/bulk`; I don't expect a problem, but check rather than assume).

Do the API change in THIS PR, not a stacked one — stacked branches get no CI on this repo, and the server matrix test is the thing I most want CI to run.

**Your item 2 stands as specified:** the mounted-lifecycle flush must close `undoRef` and the rendered `undo` state atomically, then reconcile `acknowledged`/`failed`/`pendingAcks` and surface errors — no lying UNDO after a transient `inactive`, no rows stranded behind `pendingAcks`. With the bulk request this gets simpler: one response to reconcile.

**Out of scope, noted for later:** the 13–15s server-side acknowledge (event published inside the request path) is its own defect and I'll file it separately — don't let it shape more client architecture than it already has.

Status: waiting on the `background` boundary + bulk-permission change + reconcile split.

ToddHebebrand:
Correction to my last line: the 13–15s acknowledge already has an issue — **#3724**, closed as likely resolved by the #3299 Redis head-of-line fix in v0.105.0 (your own measurement + write-up). No new issue filed; if you still see 13–15s per acknowledge on a ≥v0.105.1 server while building this round, reopen #3724 with endpoint timings rather than tuning the client around it — the `ACK_CONCURRENCY` comment's numbers may already be stale.

bdunncompany:
Both halves of the ruling are in and green. Adversarial review then found **a security issue in the bulk route that neither of us had looked at**, and **one thing that undercuts the premise of the one-request fix** — that second one needs your call before this merges.

### Done

**API.** `POST /alerts/bulk` gates per action after `zValidator`: `acknowledge` accepts `ALERTS_ACKNOWLEDGE` or `ALERTS_WRITE`, the other three still require `ALERTS_WRITE`. The authz suite pins the matrix both ways; control run — with the old write-only gate, exactly one case fails (acknowledge-only can bulk-acknowledge) and the three still-403 cases keep passing.

I checked your equivalence rationale rather than assuming it, since this loosens RBAC. It holds: `getAlertWithOrgCheck` on the single route enforces org **and** site scope and refuses non-active with a 400; bulk gates on the same scope tiers, applies the same org predicate, filters via `filterAlertsBySiteScope`, and skips non-active. Bulk grants no capability the single route lacks. From the seed, exactly two roles gain anything — `Partner Technician` and `Org Technician`, both of which already hold `alerts:acknowledge`. `Partner Viewer` and `Org Viewer` don't hold it and gain nothing.

**Mobile.** One `POST /alerts/bulk` via the existing `coreRequest` (verified the mobile token reaches it — core alerts is at `/api/v1/alerts` and `coreRequest` carries auth, CSRF and install id). Commit on `background`. Mounted flush closes `undoRef` and the rendered state together and dispatches through `dispatchAcknowledge` so the result reconciles; unmount keeps the bare fire-and-forget.

Two things I had to add that your ruling assumed existed: the bulk response was **counts only**, so exact reconciliation was impossible — it now also returns `updatedIds`/`skippedIds`/`failedIds` (additive; the web toast reads only the counts, typed optional). And my first bulk call passed **no timeout**, silently falling back to the short default while the single path uses `ACKNOWLEDGE_TIMEOUT_MS`; a test caught it.

`ACK_CONCURRENCY` is gone — this made it dead, and per your note its 13–15s figure is probably stale.

### Security fix, pre-existing but this PR widens who can reach it

The bulk route wrote **one** audit record pinned to `accessible[0].orgId`, with `details.alertIds` listing **every** id in the batch. A partner-scope caller can legitimately submit alerts from several orgs — the predicate is `inArray(orgId, accessibleOrgIds)` — so **org B's alert UUIDs landed inside org A's audit row**. Audit list access is scoped by the row's `orgId`, not by what's inside `details`, and the sanitizer passes string arrays through, so an org-A-only audit reader can read them. Org B got no record of its own alerts at all.

Now one record per org with only that org's ids, per-org counts, and a `resourceId` naming an alert that record can actually account for rather than whichever sorted first. Granting bulk to technicians is what made this worth fixing here rather than filing.

### The open question — `background` releases ownership before the request is dispatched

This is the one I'd like your ruling on, because it contradicts what we both assumed.

`acknowledgeAlerts` doesn't reach `fetch` synchronously. It first awaits SecureStore token retrieval, the CSRF token, the installation id and the server URL. So on `background`: `flushAllUndo` empties the ref → `dispatchAcknowledge` yields on `await getToken()` → **if the process is suspended there, the batch is gone**. The timer can't resend, unmount can't resend, both correctly see an empty ref.

That's the same stranded-in-memory class we set out to close, moved from "unsent tail of a worker queue" to "whole batch taken before transport dispatch". One request removes the *partial* loss; it doesn't make the flush durable.

Options as I see them:

1. **Durable outbox** — persist the batch before releasing the ref, clear on a confirmed outcome, replay on next launch. The only one that actually survives process death. Needs dedupe on replay, since acknowledging is irreversible.
2. **Native background-execution assertion** — improves the odds, still not durable against termination.
3. **Accept it and shrink the window** — pre-warm credentials so the pre-dispatch await is short.

I'd take (1), but it's a bigger change than this PR and I didn't want to expand scope on my own judgement. Happy to do it here or split it.

Related sizing worry: the 45s client deadline now covers up to **100 serial** server-side acknowledges, each of which the code says has measured 13–15s. Even allowing that the figure is stale, a handful of alerts could cross it. That's why an ambiguous timeout now lands in a new `unknown` bucket rather than `failed` — `unknown` ids are **not** restored (restoring invites a second acknowledge of an irreversible action) and are released after the next authoritative fetch. The response arrays are validated before being trusted, so an older counts-only server reads as `unknown` rather than silently restoring rows it had acknowledged.

### Nit worth taking

RBAC now runs after validation, so a scoped caller without alert permissions can tell a malformed body (400) from a valid one (403) — a small schema oracle. Fix without parsing twice: a coarse pre-validation gate requiring *either* alert permission, then the action-specific check after. Say the word and I'll add it.

### Verification

`tsc` clean both packages. **24,512 API unit passed, 0 failed.** Integration **2,301 across 4 shards, fresh DB each**. Mobile **580**. Alerts suite **215**.

ToddHebebrand:
**Rulings on all three open points — and one logistical blocker: none of this is on the branch yet.** The head is still the 08-22 merge commit; the last push predates the 08-23 ruling. Push what you described and I'll review the code rather than the account of it.

**Ruling 1 — the background-durability question: take option 3 in this PR, and the durable outbox is now #3919.** You're right that the pre-dispatch awaits move the loss window rather than close it, and right that this contradicts what we both assumed. But the two windows are not morally equivalent: the residual failure is *honest* — the server never acknowledged, the rows un-hide on the next authoritative fetch, the operator re-acks. Nothing lies. A durable outbox replaying an **irreversible** action across process death is its own design problem — replay dedupe against the response arrays, staleness expiry (an alert acked hours later at next launch may have escalated since), and non-resurrection of `unknown`-bucket ids — and bolting that onto a PR in its seventh round is how careful work goes wrong. So: pre-warm the credentials to shrink the window here, and #3919 carries the rest with the design points written down. Your instinct not to expand scope on your own judgement was correct.

**Ruling 2 — the RBAC schema-oracle nit: yes, add it.** The coarse either-alert-permission pre-gate before validation, action-specific check after, no double parse — exactly as you proposed.

**The audit fix — good catch, and the shape is right.** One record per org with only that org's ids is the correct contract; cross-org alert UUIDs inside an org-A-scoped audit row is a real leak, and you're right that widening who can reach the route is what made it worth fixing in-PR rather than filing. I'll verify the per-org split and the `resourceId` choice when the commits land.

**On the `unknown` bucket:** not restoring ambiguous-timeout ids is the right call — restoring invites a second acknowledge of an irreversible action — and validating the response arrays so an older counts-only server degrades to `unknown` rather than silently restoring is exactly the backward-compatibility posture I'd have asked for.

Push it; assuming the code matches the description this is the last round.

bdunncompany:
Pushed — `f3087d65`. Your three rulings are implemented, plus one I'm declining with evidence rather than silently.

### Ruling 1 — background durability, option 3

Done. The flush sends one bulk request, commits on `background`, and the `unknown` bucket is not restored.

**The credential pre-warm is NOT in this PR, and I'd rather tell you why than ship a no-op.** It cannot help the case it was aimed at. `getToken` has no cache — it hits SecureStore on every call — so warming it does nothing for the dispatch and only adds a duplicate keychain read inside the window it was meant to protect. Caching it is not a local change either: `breeze_auth_token` is read directly by five service modules, each with its own copy of the key, and written from `auth.ts`, so a cache in `api.ts` goes stale against writes it cannot see.

That leaves CSRF and the installation id. The installation id is already warm from the screen's own list requests. Warming CSRF is the one case with real benefit — list GETs skip it, since it is mutation-only, so a restored process has a cold entry — but it makes a latent bug reachable: `getCsrfHeaderValue` assigns its module cache **after** an await, so a read that starts before a refresh installs a new token completes afterwards and writes the old one back over it. The next mutation then sends token A against cookie B. I wrote a generation guard, then reverted it: review found three interleavings it still missed, plus a fourth problem I had not touched — sign-out deletes directly while token writes go through `queueWrite`, so a delayed write can resurrect a token after a wipe reports success. A partial guard in shared auth code is worse than none, because it reads as solved.

A shared token accessor and the CSRF cache/serialisation are each their own change. Happy to file both if useful.

### Ruling 2 — the RBAC pre-gate

Implemented, but **not** as a permission-checking middleware ahead of the validator, which was the obvious shape and is wrong here. `middleware/auth.ts` warms the permission cache only for organization scope, so on the partner path a pre-body gate is a new route-level lookup — and since the client decides when the body arrives, the authorising read could then return the gate's own entry. A stalled body plus a revocation whose invalidation `INCR` failed keeps access alive for the cache TTL.

`zValidator` is gone from the route for a related reason: it answers its own 400 before any handler code runs, **even when given a hook**, so with it in the chain the schema oracle stays open. The handler reads the body once, does one route-level lookup, checks coarsely before any body verdict, then parses, then checks the action. The media-type verdict is preserved explicitly — `c.req.json()` parses valid JSON whatever the Content-Type says, so dropping the validator would otherwise have widened what reaches the mutation.

One limit worth stating: the no-body-verdict guarantee holds for a caller with **neither** alert permission. An acknowledge-only caller is unauthorised for resolve/suppress/dismiss and still distinguishes 400 from 403, because the action-specific check has to run after the parse.

### The audit fix

Per-org rows carrying only that org's ids, as you described.

### On the review

Eleven adversarial rounds. Worth recording what they caught, because most of it was tests agreeing with bugs:

- the media-type predicate I hand-rolled **accepted `application/json-bogus`**, which Hono rejects, and rejected `application/vnd.api+json`, which it accepts — and both of my tests passed, because they only covered cases where the two agree. It now copies Hono's `jsonRegex` behind a shared helper, with a drift guard that uses Hono itself as the oracle.
- the ordering test **counted** permission lookups, so moving the sole lookup above the body read kept it green. It now streams the body and records whether it was consumed at lookup time.
- the mocked `requirePermission` never called the tracked lookup, so re-introducing a middleware ahead of the handler would have passed.
- a structural assertion of mine could never fail (`closest()` matches the element it starts from).
- the prewarm test was **untracked** and would not have been in the PR at all.

Every fix here ships with a control that was observed failing against the broken version. The last three rounds found only comment-scope wording, no code defects, which is where I stopped.

**Verification:** `tsc` clean; API unit **24,538 passed / 0 failed**; mobile **580 passed / 0 failed**.

bdunncompany:
Rebased onto current main — head is now `c8cd1e60`, and the PR is mergeable again.

It went conflicting a few minutes after the previous push because main had moved. Worth noting how the conflict resolved, since it was not a real one: main's changes to `alerts.ts` are `withAlertActorNames` in the **list** and **get-by-id** handlers (#3966), which do not touch `/bulk` at all. Textually adjacent, semantically independent — both changesets are present.

I did not take git's automatic resolution. `rerere` replayed a stored resolution from an earlier rebase of this branch and produced an `alerts.ts` with **none** of this round's work in it — no `isJsonContentType`, no hand-parse, no route-level lookup — while reporting the file resolved and leaving no conflict markers. Silently correct-looking and wrong. I rebuilt the branch on `origin/main` instead and re-applied the `/bulk` hunks, then checked for both changesets explicitly before committing.

The test counts are the useful confirmation that the rebase took rather than clobbered:

|  | before rebase | after |
|---|---|---|
| API unit | 24,538 | **25,243** |
| Mobile | 580 | **612** |

Both went up by exactly main's new tests, with **0 failures** and `tsc` clean on api and mobile. A rebase that had dropped main's work would have shown the old numbers.

One incidental find while installing: main added `cron-parser`, so a worktree carrying an older `node_modules` fails `tsc` on `jobs/scheduleRegistry.contract.test.ts` with `Cannot find module 'cron-parser'` until you reinstall. Environmental, not a code problem, but it looks like a type error at first glance.

ToddHebebrand (approved):
**All three rulings verified in code, the decline is correct, and this is the last round — merging.**

I checked each item independently against `c8cd1e60` rather than against your account of it, since fix-up commits here have historically covered some of a review and not all of it. This one covers all of it.

**Ruling 1 — background durability.** `undoAck.ts` flushes one bulk request (`acknowledgeAlerts([...ids])`), `SystemsScreen.tsx:1200-1201` commits on `background` and not `inactive`, and the restore path takes only confirmed failures — `unknown` ids stay hidden, with the reasoning about a second acknowledge of an irreversible action written down at the call site. That is the shape I asked for.

**The pre-warm decline is right, and I verified it rather than taking it on trust.** `getToken` at `apps/mobile/src/services/api.ts:197-203` is a bare `SecureStore.getItemAsync` with no cache, so warming it genuinely does nothing but add a second keychain read inside the window it was meant to protect. And the CSRF race is real: `getCsrfHeaderValue` at `apps/mobile/src/services/csrfToken.ts:129-134` checks `cached`, awaits, *then* assigns — so a read that starts before a refresh installs a new token completes afterwards and writes the stale one back. Shipping a partial guard over that would have been worse than shipping nothing, and reverting the generation guard once review found interleavings it missed was the right instinct. Declining with evidence beats implementing a no-op; do this more.

**One correction: it is nine service modules, not five.** `breeze_auth_token` is read with its own local copy of the key in `aiChat.ts`, `api.ts`, `approvals.ts`, `approverDevice.ts`, `connectedApps.ts`, `mobileDevices.ts`, `search.ts`, `systems.ts` and `systemsRealtime.ts`, plus the write side in `auth.ts`. That cuts in your favour — it makes the shared-accessor argument stronger, not weaker — but the habit of checking a stated count against the tree is worth keeping, since a number in a comment outlives the comment.

**Yes, please file both follow-ups:** the shared token accessor, and the CSRF cache/serialisation including the sign-out-vs-`queueWrite` resurrection you spotted. Reference #3919 from them so the durability work reads as one thread.

**Ruling 2 — the RBAC pre-gate.** Implemented as specified and, more usefully, not as the obvious middleware shape. The handler does one `c.req.json()`, one route-level `getUserPermissions`, the coarse either-permission check before any body verdict, then media-type, then schema, then the action-specific check. Dropping `zValidator` while keeping an explicit `isJsonContentType` check is the correct trade — `c.req.json()` will parse a valid payload sent as `text/plain`, so without it you'd have widened what reaches the mutation. You also preserved the two things `requirePermission` would otherwise have done and that would have been silently lost: the `ai_agent` rejection and `c.set('permissions')`. That is the "parallel case" check I've had to ask for before; you ran it yourself.

Bounding the guarantee explicitly — an acknowledge-only caller still distinguishes 400 from 403 because the action check must follow the parse — is better than claiming a clean property that isn't true.

**Audit fix.** One `writeRouteAudit` per org carrying only that org's ids, per-org counts, and `resourceId` preferring an alert that was actually updated. Confirmed at `alerts.ts:724-748`. The cross-org leak was real: audit list access is scoped by the row's `orgId`, not by the ids inside `details`, so org B's alert UUIDs sitting in an org-A row were readable by an org-A-only auditor, and org B got no record at all.

**On the eleven rounds.** The five findings you listed are all the same species — tests agreeing with the bug — and two of them (a `closest()` assertion that can never fail; an untracked test file) are ones I've flagged on your PRs before. Finding them yourself, and shipping each fix with a control observed failing against the broken version, is the change I've been asking for across the last dozen rounds. The media-type case is the best of them: hand-rolled predicates that disagree with Hono at the edges, with both tests passing because they only covered agreement, is exactly the failure mode that survives review.

Merging.

PR #3728, opened 2026-08-19
fix(selfhost): stop .env.example shipping a localhost DATABASE_URL_APP

ToddHebebrand:
A self-hoster's from-scratch install dies at first boot. Migrations and role setup succeed, then the API crashes during the initial seed:

```
[auto-migrate] All migrations already applied
[ensure-app-role] breeze_app role ensured (NOSUPERUSER, NOBYPASSRLS)
[auto-migrate] No users found, running initial seed...
[CRITICAL] API startup failed: AggregateError [ECONNREFUSED]
  connect ECONNREFUSED ::1:5432
  connect ECONNREFUSED 127.0.0.1:5432
```

The reporter changed `BREEZE_DOMAIN` to a LAN IP, then to a real hostname, with no effect — `BREEZE_DOMAIN` was never involved. This is not user error.

## Cause

The bundled root `docker-compose.yml` treats the two DB URLs asymmetrically:

- `DATABASE_URL` is **rebuilt** as `...@postgres:5432` inside the api container.
- `DATABASE_URL_APP` is passed through **verbatim** (`${DATABASE_URL_APP:-}`).

`.env.example` shipped an *uncommented*:

```
DATABASE_URL_APP=postgresql://breeze_app:...@localhost:5432/breeze
```

so the documented quickstart — `cp .env.example .env && docker compose up -d` — puts a loopback URL into the api container's unprivileged request pool.

That explains the exact log ordering: `autoMigrate` and `ensureAppRole` use `DATABASE_URL` and reach `@postgres` fine; the **first** use of the request pool is the seed, which dials the api container's own loopback. Every new self-hoster on this path hits it.

## Fix

- **`.env.example`** — comment out `DATABASE_URL_APP`; explain that Compose derives `breeze_app@postgres` from the container's own `DATABASE_URL`. Also noted that the `localhost` in `DATABASE_URL` *is* correct and safe (host-side tooling only), so the next reader doesn't "fix" that one too.
- **Docs (`deploy/environment`)** — the table called `DATABASE_URL_APP` "Preferred", nudging Compose users straight into this. Scoped that to direct/non-Compose deployments and added a caution carrying the failure signature so the error is searchable.

## Guard

This is the same failure `check-guided-setup-db-urls.sh` already existed to prevent — its header literally documents "crash at seed with ECONNREFUSED". It only ever scanned `scripts/guided-setup.sh`, leaving the second and more commonly used path, `.env.example`, unguarded.

Renamed to `scripts/check-selfhost-db-urls.sh` (one CI reference updated) and broadened to both install paths:

1. `guided-setup.sh` — exactly one, empty, `set_env_value` assignment (unchanged).
2. `.env.example` — `DATABASE_URL_APP` must be commented out or empty.
3. `deploy/.env.example` — an explicit value is legitimate for managed Postgres, so it is checked only for loopback hosts.

## Verification

Each new branch was verified to fail on the regression it guards, not just to pass:

```
$ # reinstate the shipped line
$ bash scripts/check-selfhost-db-urls.sh
check-selfhost-db-urls: .env.example sets DATABASE_URL_APP to a non-empty value
(35:DATABASE_URL_APP=postgresql://breeze_app:...@localhost:5432/breeze). ...
exit=1

$ # loopback in deploy/.env.example
check-selfhost-db-urls: deploy/.env.example sets DATABASE_URL_APP to a loopback
host (342:...@127.0.0.1:5432/breeze); ...
exit=1

$ # empty assignment still allowed -> exit=0
```

- All five Lint-job self-host guards pass.
- `shellcheck` clean on the new script.
- Docs build completes (165 pages).

## Note for affected users

Existing broken installs need the line removed from their **already-generated** `.env` — this fix only corrects the template:

```
sed -i 's/^DATABASE_URL_APP=/#DATABASE_URL_APP=/' .env && docker compose up -d api
```

Follows #3717, which fixed the other first-boot blockers on this same path.


bdunncompany:
Heads up before you chase this PR's red check: the `Trivy Filesystem Scan` failure is not yours.

Its only two findings are `CVE-2026-56864` and `CVE-2026-56865` in `golang.org/x/mod v0.38.0` via `agent/go.mod`, and this branch touches no Go files. `pnpm-lock.yaml` scans 0, and your `.env.example` edits produced zero secret findings, so nothing here contributed to it.

`main` is red for the same reason at `10bb1af27`, and #3714 is exactly the `x/mod` `0.38.0 -> 0.40.0` bump that clears it. Every open PR is currently failing this job for the same reason, so it will follow anything branched off current `main`.

Worth knowing while you look at it: no commit caused this. Security Scanning passed on `064d51cf8` (2026-08-17) with the *same* `x/mod v0.38.0` already pinned, and neither #3717 nor #3718 touches `agent/go.mod`. Same `aquasecurity/trivy-action@ed142fd0`, same `trivy-db:2` in both runs. The advisories landed in Trivy's database between the two runs, so there is nothing to bisect and the timing against your two merges is coincidental.

One trap I walked into and backed out of, in case it saves you the same detour: #3714's own checks show `agent/go.mod gomod 0`, which reads like confirmation the bump works. It is not. #3713 leaves `x/mod` at `v0.38.0` and also scanned `0` in a run twelve minutes earlier the same day, because the 2026-08-18 database did not yet carry these advisories. #3714's checks predate the database update and need a rebase before they mean anything.

Details and the full evidence are in my comments on #3714.

PR #3729, opened 2026-08-19
fix(agent): bump golang.org/x/mod to v0.40.0 for CVE-2026-56864/56865

ToddHebebrand:
Unblocks the **blocking Trivy filesystem scan**, which has been failing on `main` since 2026-08-19 and reddens `security.yml` on every open PR.

```
agent/go.mod (gomod)
Total: 2 (HIGH: 2, CRITICAL: 0)
golang.org/x/mod  CVE-2026-56864  HIGH  v0.38.0 -> 0.40.0
                  CVE-2026-56865
```

## Not a regression

No Go dependency changed in the window — `security.yml` was green on 08-17 (`064d51cf8`) and red on 08-19 (`10bb1af27`). These are **newly published advisories landing on an unchanged pin**, so they surfaced on whichever PR happened to run next rather than being introduced by one.

- **CVE-2026-56864** — a malicious GOSUMDB could serve arbitrary module content.
- **CVE-2026-56865** — a malicious GOPROXY could forge up to two sumdb entries.

## Exposure

Both are **module-fetch-time** supply-chain issues. `golang.org/x/mod` is a *direct* dependency — `agent/internal/versionpolicy/semver.go` imports `golang.org/x/mod/semver` — but the agent only uses it for semver parsing at runtime and never fetches modules. So this is supply-chain hygiene plus a red-main fix, **not a live agent vulnerability**. Flagging that explicitly because this code ships to customer machines and the CVE severity alone would suggest otherwise.

## Change

`go get golang.org/x/mod@v0.40.0 && go mod tidy` — minimal diff, one line in `go.mod`, two in `go.sum`, nothing else pulled in.

## Verification

Run locally on the exact CI toolchain pin (`go1.26.6`, matching `GO_VERSION` in `ci.yml`):

```
$ go build ./...                                # clean
$ go test -race ./...                           # exit 0 — 69 packages, 0 failures
$ go test -race ./internal/versionpolicy/...    # ok — the x/mod consumer
```


bdunncompany:
`Integration Tests (shard 4/4)` here is a registry flake, not your change. It failed in `Start test-stack containers` after 47s:

```
postgres-test Error Head "https://registry-1.docker.io/v2/pgvector/pgvector/manifests/pg16":
Get "https://auth.docker.io/token?...&service=registry.docker.io":
net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
##[error]Process completed with exit code 1
```

Docker Hub timed out issuing the pull token for `pgvector/pgvector:pg16`, so the stack never came up and no test ran. The other three shards passed, and 38 of the 40 jobs in the run are green with only `Rust Check (Tauri apps)` still going. A rerun of the failed job should clear it. I will kick one once the run finishes if it is still sitting there.

More usefully, **this branch is the first evidence that the bump actually fixes the finding**. `Trivy Filesystem Scan` passes here in 37s, against today's database with `v0.40.0` in place.

That distinction matters because the obvious confirmation was misleading. #3714 is the same bump and shows `agent/go.mod gomod 0`, but that row proves nothing: #3713 leaves `x/mod` at `v0.38.0` and also scanned `0`, twelve minutes earlier the same day, because the 2026-08-18 database did not yet carry these advisories. Both dependabot PRs predate the database update, so their green rows are stale rather than meaningful. Your branch is the first run of the fixed version against a database that knows about the CVEs, which is the result that settles it.

Also confirms your exposure note: `Go Vulnerability Check` passes on both, consistent with the vulnerable symbols living in `x/mod/sumdb` and `sumdb/tlog` while the agent only reaches `x/mod/semver`.

bdunncompany:
Heads up on this PR's `Rust Check (Tauri apps)` state, and a disclosure: **the failed run there is one I triggered.**

I reran that job last night. The rerun stalled in `Install Tauri system dependencies` and was cancelled after about 35 minutes (job `96222993704`, 20:52:43Z → 21:28:15Z). The log is silent for ~33 minutes inside that step, so the existing three-attempt retry loop never got a chance to advance — a hung `apt` never returns, so the loop cannot reach attempt 2.

To be clear about what that does and does not mean: I caused that particular run to exist, so please don't read it as "every run of this PR fails this way." And a plain rerun may well succeed — my own evidence says so, below.

I opened #3731 to add per-command deadlines around the apt phases. Its own CI reproduced the same behaviour and then recovered: the first install attempt stalled 132s in, the 5-minute deadline terminated it, and attempt 2 succeeded 38 seconds later. `Rust Check` then passed in 6m52s. Total 5.8 minutes instead of a 35-minute job kill.

**#3731 is not a prerequisite for merging this.** A rerun here may pass on its own. What the change does is make this failure mode bounded so the retries you already have can actually work when the stall recurs. If you'd rather not carry a separate PR, the workflow change is small enough to fold directly into this one.

Worth knowing for prioritisation: `main`'s Security Scanning run is currently failing on exactly the `golang.org/x/mod` findings this PR updates (its only run for `10bb1af27` concluded `failure`), so getting this dependency update through is the thing that clears it.

bdunncompany:
Context on how this relates to #3714, since both are open and they look like duplicates.

**What main is actually red on.** At main head `10bb1af27`, the only failing check is `Trivy Filesystem Scan` (58 other checks pass). It reports two HIGH findings, both against `agent/go.mod`:

```
golang.org/x/mod  CVE-2026-56864  HIGH  fixed  v0.38.0 -> 0.40.0
golang.org/x/mod  CVE-2026-56865  HIGH  fixed  v0.38.0 -> 0.40.0
```

Every other target (both `Cargo.lock`s, `e2e-tests/package-lock.json`, `pnpm-lock.yaml`) is 0.

**The two PRs are byte-identical.** #3714 and this one produce the same blobs: `agent/go.mod` `f3e91a4a3..3fd231a3f` and `agent/go.sum` `4a96e03c3..fa9602194`. Two files, +1/-1 and +2/-2. There is no technical difference between them.

**Why #3714's CI is red anyway, and it is not this CVE.** Its Trivy run (2026-08-18) reports `agent/go.mod` = 0, and fails on something else entirely:

```
deepmerge-ts  CVE-2026-40345  HIGH  fixed  7.1.6 -> 8.0.0   (pnpm-lock.yaml)
```

That is a stale-base artifact. `merge-base(main, #3714)` is `064d51cf8` and main is 2 commits ahead of it; main now carries `deepmerge-ts@8.0.1` plus the override `deepmerge-ts@<8.0.0: '>=8.0.1 <9.0.0'`, while #3714's tree still has `7.1.6`. Rebasing #3714 onto main should clear the currently observed Trivy findings, subject to a fresh run.

**Status of this PR, stated precisely.** Both Trivy jobs pass here (`Trivy Filesystem Scan`, `Trivy Image Scan`), but I would not call it green. The aggregate `CI Success` check is failing, and the reason is visible in its log: every input is `success` except `RUST_CHECK_RESULT: cancelled` — the `Rust Check (Tauri apps)` job was cancelled, and the gate treats any non-success as failure. That is a cancellation, not a test failure. I have re-triggered it; if it comes back clean the aggregate should go with it.

**Neutral path,** since the dependency change is equivalent in both: either rebase and re-run #3714, or take this one once its aggregate check clears. I have no technical reason to prefer this PR over #3714 and am not arguing for one — whichever is easier to land removes main's red.

One correction to something I could easily have implied: #3714 showing 0 `x/mod` findings has nothing to do with scanner-database timing. That branch already contains the fixed `v0.40.0`, which fully accounts for the zero.

bdunncompany:
Follow-up: the aggregate check has cleared. Re-running the cancelled `Rust Check (Tauri apps)` was all it needed — this head is now **54 success, 1 skipped, 0 failures**, including all three Trivy jobs.

So this is green and removes main's only failing check. Same caveat as before on the choice between this and #3714: the two diffs are byte-identical, so if you would rather land the dependabot one, rebasing it onto current main should do it (its red is the already-fixed `deepmerge-ts@7.1.6` from a base 2 commits behind, not `x/mod`).

bdunncompany:
Closing this as redundant — **#3714 is the better PR now and is fully green.**

I opened this when #3714 looked stuck behind a failing Trivy scan. That turned out to be a stale base rather than anything wrong with its diff: it still carried `deepmerge-ts@7.1.6` while main had `8.0.1`, so the scan failed on deepmerge, not on `x/mod`. A rebase was all it needed.

Post-rebase, #3714 is at `352aad3d4`: **54 success / 1 skipped / 0 failures**, zero commits behind main, mergeable, and its entire diff is `agent/go.mod` +1/-1 and `agent/go.sum` +2/-2 — byte-identical to this one.

So there's no reason to weigh two identical diffs against each other. #3714 is dependabot-authored, current, and green, which makes it the cleaner merge for clearing main's `Trivy Filesystem Scan` red (`golang.org/x/mod` v0.38.0, CVE-2026-56864 / CVE-2026-56865, fixed in 0.40.0).

Reopen this if #3714 goes sideways for any reason — the branch stays put.

PR #3732, opened 2026-08-19
feat(mobile): let a self-hosted build add its own associated domain

### LINKED ISSUE #3725: [Mobile] Self-hosted domains cannot be added to associatedDomains, so password/2FA autofill never associates
## Summary

`ios.associatedDomains` is hard-coded to the two hosted regions, so a self-hosted deployment gets no strong password/2FA autofill association with the published iOS app. Signing in on a self-hosted server means filling the password from the password manager, then leaving the app to copy the TOTP code by hand.

## Current state

`apps/mobile/app.json`:

```json
"associatedDomains": [
  "webcredentials:us.2breeze.app",
  "webcredentials:eu.2breeze.app"
]
```

The server side is already in place for the hosted domains — `docker/Caddyfile.prod` serves `/.well-known/apple-app-site-association` (added in #3171). The gap is purely that the entitlement list is fixed at build time and a self-hoster's domain cannot be in it.

This is already acknowledged in `apps/mobile/STORE_SUBMISSION.md`, which notes the list is static at build time so self-hosted servers cannot be covered by the distributed app. This issue is a request to close that gap rather than a report that it is undocumented.

The login fields themselves are correct — `textContentType`/`autoComplete` are set for `username`, `password` and `oneTimeCode`, so generic autofill still appears. What is missing is the domain-to-app association that makes a saved entry match, which is what drives reliable TOTP suggestion.

## Why it matters beyond convenience

Self-hosting is a first-class deployment mode. On a self-hosted server the phone app currently has a materially worse sign-in than the web app, and with a 15-minute access token that friction is hit often (see the separate refresh issue).

## Possible directions

Only the first is likely worth the effort:

1. **Make the list build-configurable** — an `app.config.js` (or EAS profile) that reads a `BREEZE_ASSOCIATED_DOMAINS` env var, so a self-hoster building their own app can add their domain without patching `app.json`. Also useful for anyone shipping an internal build.
2. Ship a documented recipe for step 1 in `STORE_SUBMISSION.md` for self-hosters who already build from source.

Neither helps users of the published App Store build, which is inherent — Apple binds Associated Domains into the signed entitlement, so an arbitrary runtime domain cannot be added. Worth stating plainly in the docs so self-hosters do not chase it.

There is no `app.config.js`/`app.config.ts` and no `eas.json` in the tree today, so option 1 is new configuration rather than a change to existing plumbing.

## Environment

- Self-hosted server behind a reverse proxy; the app is pointed at it via the Custom server option
- iOS, app built from `main`

bdunncompany:
Closes #3725.

`ios.associatedDomains` is pinned in `app.json` to the two hosted regions, so a self-hosted deployment gets no domain-to-app association. A saved password entry then does not match the app and the TOTP code is never suggested, so signing in means leaving the app to copy it by hand — hit often against a 15-minute access token.

Apple binds Associated Domains into the signed entitlement, so this cannot be a runtime setting and **the published App Store build can never cover an arbitrary domain**. That part is a platform constraint and the docs now say so plainly, so self-hosters stop chasing it. What it *can* be is build-time: `app.config.js` merges `BREEZE_ASSOCIATED_DOMAINS` into the list from `app.json`, so a self-hoster building from source can cover their own server.

```bash
BREEZE_ASSOCIATED_DOMAINS=breeze.example.com npx expo prebuild --platform ios
```

## Two deliberate restrictions

The output lands in a signed entitlement, so this is more locked down than the issue asked for:

- **Only `webcredentials:` is ever emitted.** My first attempt also honoured `applinks:` and `activitycontinuation:` prefixes, which would let a build-time env var widen the entitlement into universal-link routing for no benefit to password autofill. Dropped.
- **A bad entry fails the build rather than being skipped.** A silently dropped entry ships an entitlement missing the domain, and the only symptom is autofill quietly not working, discovered much later. Wildcards, IP addresses, `localhost`, and a bare host carrying a port or `?mode=developer` are each rejected *by name* rather than trimmed:

```
$ BREEZE_ASSOCIATED_DOMAINS='*.example.com' npx expo config --type public
Error: Error reading Expo config at .../app.config.js:
BREEZE_ASSOCIATED_DOMAINS: "*.example.com" must not use a wildcard.
```

Silently rewriting an operator's input is how you end up with an entitlement that does not match what they asked for. `?mode=developer` is the sharp case: an earlier revision stripped it, which would have quietly changed behaviour.

With the variable unset the base list is returned verbatim — not merely equal, but uncopied and un-deduped — so the published build cannot be affected by this path at all. There is a test pinning that.

## Implementation note

The resolver is plain CommonJS, not TypeScript, and that is not a style choice. Expo transpiles `app.config.js` itself but **not** the modules it requires, so a `.ts` module fails at config evaluation:

```
Error: Cannot find module './src/config/associatedDomains'
```

That was the first thing I tried. The comment in the file records it so the next person does not repeat it.

## Verification

Through the real `expo config --type public` CLI, not a simulation:

| Input | `ios.associatedDomains` |
|---|---|
| *(unset)* | the two hosted regions, unchanged |
| `breeze.example.com` | hosted regions + `webcredentials:breeze.example.com` |
| `https://a.example.com/login US.2BREEZE.APP` | hosted regions + `webcredentials:a.example.com` (URL reduced, case-different duplicate dropped) |
| `*.example.com` | build fails with the message above |

Mobile CI gate, matching the `test-mobile` job:

```
pnpm test --filter=breeze-mobile     32 files, 350 passed, 3 skipped
pnpm exec tsc --noEmit               exit 0
```

The 28 new tests were checked against deliberately broken implementations rather than trusted because they were green — removing label validation fails 3, deduping the base list fails 1, and swapping the throw for a silent skip fails 17.

Independent adversarial review blocked the first revision of this and was right to: it caught the entitlement-widening prefixes, the destroyed `?mode=developer`, validation that amounted to "contains a dot", and silent skipping. A second pass then caught two false *rejections* — semicolon treated as a separator breaks a valid pasted URL containing one, and a bare `bücher.example.com` was rejected while `https://bücher.example.com` was accepted because `new URL()` punycodes it. Both fixed; bare hosts now go through the same parser so IDN behaves identically either way.

### DIFF
--- a/apps/mobile/STORE_SUBMISSION.md
+++ b/apps/mobile/STORE_SUBMISSION.md
@@ -143,11 +143,36 @@ Verify with:
 curl -sS -D- https://us.2breeze.app/.well-known/apple-app-site-association
 ```
 
-Three constraints worth recording:
+Four constraints worth recording:
 
-- The domain list is **static at build time** and can only name hosts we
-  control, so **self-hosted Breeze servers can never be covered** by it. Their
-  users get generic autofill, which is the same as today.
+- The domain list is **static at build time**, because Apple binds Associated
+  Domains into the signed entitlement. **The published App Store build therefore
+  cannot cover a self-hosted server** — that is a platform constraint, not
+  something a runtime setting can fix, and self-hosters should not chase it.
+  Those users get generic autofill.
+- A self-hoster **building the app themselves** can cover their own domain.
+  `app.config.js` merges `BREEZE_ASSOCIATED_DOMAINS` into the list from
+  `app.json`:
+
+  ```bash
+  BREEZE_ASSOCIATED_DOMAINS=breeze.example.com npx expo prebuild --platform ios
+  ```
+
+  Several entries may be separated by commas or whitespace, and a pasted URL is
+  accepted (`https://breeze.example.com/login` resolves to the host). The two
+  hosted regions above are always kept, so this can only add to the list and
+  cannot break autofill for hosted users. The self-hosted server still has to
+  serve the AASA file described above, with its own team ID and bundle
+  identifier if the build is signed under a different Apple account.
+
+  Only `webcredentials:` is emitted, and an entry that is not a usable hostname
+  **fails the build** rather than being skipped — wildcards, IP addresses,
+  `localhost`, and a bare host carrying a port or `?mode=developer` are all
+  rejected by name. An internationalised domain is punycoded automatically,
+  whether written bare or as a URL. Failing loudly is deliberate: a silently
+  dropped entry would ship an entitlement missing the domain, and the only
+  symptom would be autofill quietly not working. Edit `app.json` directly for
+  anything beyond a plain password-manager association.
 - `breezermm.com` was deliberately dropped: it is the marketing site behind
   Cloudflare, nobody signs into the app there, and claiming a domain that serves
   no AASA file just leaves an unanswered claim.
--- a/apps/mobile/app.config.js
+++ b/apps/mobile/app.config.js
@@ -1,0 +1,31 @@
+const { resolveAssociatedDomains } = require('./src/config/associatedDomains');
+
+/**
+ * Dynamic config layered over `app.json`.
+ *
+ * Everything still lives in `app.json`; this exists only so a self-hosted build
+ * can add its own domain to `ios.associatedDomains`. Apple binds Associated
+ * Domains into the signed entitlement, so the list is fixed at build time and
+ * cannot be a runtime setting — a self-hoster has to build the app themselves
+ * for their domain to associate. The published App Store build can never cover
+ * an arbitrary domain, which is a platform constraint rather than a gap here.
+ *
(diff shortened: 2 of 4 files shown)


ToddHebebrand:
Merging — this is careful work. I verified the validator end-to-end rather than taking the description's word for it: unset env leaves the hosted domains untouched, a wildcard like `*.example.com` fails the build loudly, and the resolver holds up against percent-encoding, userinfo, and separator-injection attempts. Only ever emitting `webcredentials:` and failing closed on bad input is the right shape for something that gates credential autofill.

One small correction for the record, since it's the kind of thing that misleads a future reader. The PR body says the base list is "returned verbatim — not merely equal, but **uncopied**," and the code comment agrees. It isn't: `associatedDomains.js:66-67` slices before checking `raw`, so the unset path returns a copy. Your own test asserts `expect(result).not.toBe(base)` — i.e. it proves the opposite of the prose. The behavior is fine (arguably safer, since the caller's array can never be mutated); it's the description that's wrong. Worth fixing the comment so nobody later "restores" the verbatim return to match it.

PR #3741, opened 2026-08-20
ci: dispatch-only hosted server-image builds (hosted-images.yml)

ToddHebebrand:
Adds a manual `workflow_dispatch` workflow that builds and pushes **only the server-side container images** (api, web, portal; m365 executors behind an opt-in input) for an already-pushed tag.

Why: the tag-triggered release pipeline fails closed when Windows signing is unavailable (as designed), which currently also blocks every GHCR image — so hosted prod cannot receive server-side fixes while the Trusted Signing profile is suspended (v0.105.2 and v0.106.0 are both tagged with no images). This workflow is the additive escape hatch scoped on 2026-08-17: it leaves `release.yml` and its integrity gate completely untouched.

Safety properties:
- No GitHub Release is created or modified — `/releases/latest` is untouched, so self-hosted binary sync, agent auto-promote, and the viewer's Tauri updater cannot act on these images.
- No binaries-init image — agent/viewer/helper binaries stay on the last fully-signed release; consumers must pin `BREEZE_BINARIES_IMAGE_REF` explicitly.
- Only the exact `<version>` tag is pushed; `latest`/`<major>`/`<major>.<minor>` never move. `release.yml` remains the only workflow that advertises floating tags.
- Tag input is regex-validated before any checkout; executor images keep the same push-by-digest → blocking trivy scan → imagetools promote flow as `release.yml`.

Verified locally: `actionlint` clean, `check-supply-chain-hardening.sh` passes.

### DIFF
--- a/.github/workflows/hosted-images.yml
+++ b/.github/workflows/hosted-images.yml
@@ -1,0 +1,230 @@
+name: Hosted Server Images
+
+# Manual escape hatch for when the tag-triggered release pipeline is blocked on
+# artifact signing (e.g. a Trusted Signing outage): builds and pushes ONLY the
+# server-side container images for an already-pushed tag so the hosted droplets
+# can still receive server-side fixes.
+#
+# This workflow deliberately does NOT:
+#   - create, edit, or publish any GitHub Release. `/releases/latest` is
+#     untouched, so nothing in the fleet (self-hosted binary sync, agent
+#     auto-promote, the viewer's Tauri updater) can act on these images.
+#   - build the binaries-init image. Agent/viewer/helper binaries stay on the
+#     last fully-signed release; deployments consuming these images must pin
+#     BREEZE_BINARIES_IMAGE_REF to that release explicitly.
+#   - move the `latest`, `<major>`, or `<major>.<minor>` image tags. Only the
+#     exact `<version>` tag is pushed. release.yml remains the only workflow
+#     that advertises floating tags, and its "no signed release -> no images"
+#     gate is intentionally left intact.
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Existing release tag to build from (e.g. v0.106.0)'
+        required: true
+        type: string
+      include_m365_executors:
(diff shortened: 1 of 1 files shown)


bdunncompany:
This is directly useful to me as a self-hoster: I'm pinned at 0.105.1 and confirmed `0.105.2` and `0.106.0` have no images at all (manifest probe returns 404 for api/web/binaries/portal at both versions, 200 at 0.105.1), so the escape hatch is the thing that unblocks server-side fixes. The safety properties as written match what I'd want, in particular not creating a Release and not moving `latest`/`<major>`/`<major>.<minor>`.

One small provenance point, offered as cheap hardening rather than a blocker.

`validate` resolves the supplied tag to an immutable SHA and exports it, but the build jobs check out `refs/tags/${{ inputs.tag }}` again rather than that SHA, so each checkout resolves the tag independently when its job runs. The SHA is currently used only for the label. If the tag moves during a run, an image can be built from the new target while `org.opencontainers.image.revision` still reports the SHA that `validate` resolved, so the label would assert a commit that wasn't what got built. Because `build-server-image` and `build-m365-executor-image` (and the individual api/web/portal matrix legs) each re-resolve on their own, one invocation could also produce an internally inconsistent set of images across those repositories.

Since the SHA is already exported, would `ref: ${{ needs.validate.outputs.sha }}` in the build checkouts work? That makes the built content and the revision label refer to the same commit by construction.

To be clear about the threat model: moving a published release tag takes repository authority and is abnormal, so I'd call this low risk rather than a security concern. The case for it is determinism, and guarding an accidental tag correction or concurrent operator activity mid-run, which seems worth a little extra care specifically because this is the emergency path used when the normal integrity gate is unavailable.

For what it's worth I did check whether the workflow drops a control by pushing api/web/portal with plain `push: true` and no Trivy step while the executors get push-by-digest → blocking scan → promote. It doesn't: `release.yml` on main has exactly the same split, so this mirrors established practice and I don't think it's a fair thing to raise against this PR.

PR #3757, opened 2026-08-21
fix(mobile): the device walk asks for a page size the server honours

### LINKED ISSUE #3755: [Mobile] Device walk silently caps at 2,000 (client asks 200/page, server clamps to 100) + three comments stating false rationale
## Description

A comment-accuracy pass over the merged mobile batch (#3716, #3719, #3720) found one real behavioral bug and three comments whose stated rationale is false in a way that would misdirect the next person to touch the code. All verified against `origin/main`.

## 1. The device walk truncates at 2,000, not 4,000 — the comment cites the wrong cap — BUG

`apps/mobile/src/services/api.ts:502,508,524`. The client requests `limit=200` per page across up to 20 pages, and the comment justifies 200 as the "server cap (patches/helpers.ts MAX_PAGE_LIMIT)".

That is the wrong route's constant. `/mobile/devices` caps at **100**, silently:

```js
// apps/api/src/routes/mobile.ts:57-59
const limit = Math.min(100, Math.max(1, Number.parseInt(query.limit ?? '50', 10) || 50));
```

`Math.min` clamps rather than rejecting, so the client asks for 200 and receives 100 with no indication. The effective ceiling is `100 x 20 = 2,000` devices, half the 4,000 the code reads as intending — and the walk stops there silently, which is the same class as #3612.

Fix is a one-liner (`DEVICE_PAGE_LIMIT = 100`, or raise `MAX_DEVICE_PAGES`), but the ceiling should be a deliberate number with a truncation affordance rather than an accident of a mismatched constant.

## 2. Stale comment describes single-page fetching in the PR that added pagination

`apps/mobile/src/screens/devices/deviceListFilters.ts:5-10` still says "the fleet arrives as one page from `/mobile/devices` (50 by default) ... rather than round-tripping." #3720 replaced exactly that with a cursor walk. Stale within its own PR.

## 3. `mergeSystemsResults` comments say "four independent fetches" — there are five

`apps/mobile/src/screens/systems/mergeSystemsResults.ts:4-6,~40` and `useSystemsData.ts:61`. The type carries five fields and the code sets `const total = 5`. Stale since `activeAlerts` was added. The risk is specific: someone reconciling the comment with the code could "correct" the threshold to 4, changing when the merge decides everything failed.

## 4. Ticket status-change comment claims a response shape the server doesn't return

`apps/mobile/src/screens/tickets/TicketDetailScreen.tsx:174-177` (and the matching doc in `services/tickets.ts`) says the status-change response "returns the custom label too," so omitting it would blank the row. `changeTicketStatus` returns a raw `tickets` row — `statusName`/`statusColor` exist only via the join in the list/detail GET routes, so that line always writes `null`. Correctness comes entirely from the follow-up `load()`.

Nothing is broken today, but the comment makes the refetch look redundant. Someone removing it as a wasted round trip would reintroduce blank status labels, and the comment would actively support that reasoning.

## Why file this

Individually these are small. The pattern is the point: a comment asserting a cap, a count, or a response shape gets trusted over the code by the next reader. #1 is already a live behavioral discrepancy that a reader checking the comment would not catch, because the comment agrees with the code and both are wrong about the server.

Found during the 2026-08-20 GH queue sweep. A parallel pass over the CI PRs (#3743/#3735/#3731) found their comments accurate apart from one arithmetic slip, so this is not a blanket problem with the batch.

bdunncompany:
Fixes #3755 — all four items. Each verified against `origin/main` before changing anything.

**The bug.** `DEVICE_PAGE_LIMIT = 200` cited `patches/helpers.ts MAX_PAGE_LIMIT`, which is a real constant but on a different route. `/mobile/devices` goes through `getPagination`, and `listDevicesSchema` declares `limit: z.string().optional()` with no bound, so validation passes 200 straight through to `Math.min(100, ...)` — a silent clamp, not a rejection. Client asks 200, gets 100, and 20 pages deliver 2,000 devices while the code reads as 4,000.

The part worth dwelling on: **the test asserted `limit=200` too.** Comment and test agreed with each other and both were wrong about the server, so nothing in the repo could catch it. That's the failure mode you described, with a test lending it false authority. Updated to the honoured value, with the reason recorded so it doesn't drift back.

**On your point that the ceiling should be deliberate with a truncation affordance** — agreed, and I've done both. 2,000 is now stated as the intended cap rather than a by-product of the page size, and exhausting the page budget with a live cursor reports rather than returning a short list silently. Without that, a caller can't distinguish 2,000 devices from "the first 2,000 of more", and any client-side filter over the truncated list looks authoritative.

**The three comments**, all confirmed:
- `deviceListFilters.ts` described the fleet arriving "as one page (50 by default)" — accurate before #3720 added the walk, stale the moment it landed. My own PR, my own stale comment.
- `mergeSystemsResults.ts` / `useSystemsData.ts` said "four independent fetches"; there are five and `const total = 5`. Your specific risk is the right one — someone reconciling comment to code could "correct" the threshold to 4 and silently change when the merge decides everything failed.
- `TicketDetailScreen.tsx` claimed the response "returns the custom label too". I traced it: `changeTicketStatus` in `services/ticketService.ts` ends in a bare `.returning()`, so it's the raw `tickets` row; `statusName`/`statusColor` exist only via the join on the list and detail GETs. That line always writes null, correctness comes entirely from the `load()` below, and the comment made that refetch look removable.

498 pass, `tsc` clean, and restoring the 200 fails the page-size test.

On the pattern: four of these were mine, and all four are the same shape — the code moved and the comment kept justifying the previous design. Worth me tightening that rather than treating them as one-offs.

### DIFF
--- a/apps/mobile/src/screens/devices/deviceListFilters.ts
+++ b/apps/mobile/src/screens/devices/deviceListFilters.ts
@@ -5,9 +5,12 @@ export type DeviceStatusFilter = 'all' | 'online' | 'offline';
 /**
  * Client-side list shaping for the devices browser.
  *
- * The fleet arrives as one page from `/mobile/devices` (50 by default), so
- * filtering and sorting happen here rather than round-tripping. Kept pure so
- * the ordering contract is testable without a renderer.
+ * `getDevices` walks the cursor to the end of the fleet before this runs, so
+ * filtering and sorting happen here over the whole in-memory list rather than
+ * round-tripping per keystroke. (This previously said the fleet "arrives as one
+ * page ... 50 by default" — true before #3720 added the walk, stale the moment
+ * it landed.) Kept pure so the ordering contract is testable without a
+ * renderer.
  */
 
 /** Offline-first: the machines that need attention sort to the top. */
--- a/apps/mobile/src/screens/systems/mergeSystemsResults.ts
+++ b/apps/mobile/src/screens/systems/mergeSystemsResults.ts
@@ -2,7 +2,7 @@ import type { Alert, Device } from '../../services/api';
 import type { MobileSummary, OrganizationSummary } from '../../services/systems';
 
 /**
- * The four independent fetches behind the Systems screen, in the order
+ * The five independent fetches behind the Systems screen, in the order
  * `fetchAll` issues them.
  */
 export interface SystemsSlices {
@@ -39,11 +39,11 @@ function take<T>(result: PromiseSettledResult<T>, previous: T): T {
 }
 
 /**
- * Merge the settled results of the four Systems fetches over the previously
+ * Merge the settled results of the five Systems fetches over the previously
  * rendered data.
  *
  * The screen used to issue these through `Promise.all`, so a single rejection
- * discarded ALL FOUR results — a transient failure on, say, the summary call
+ * discarded ALL FIVE results — a transient failure on, say, the summary call
  * blanked a fleet of devices that had loaded perfectly well, and the user saw an
  * empty screen with a generic error. Each slice now stands on its own: whatever
  * arrived is rendered, whatever failed keeps its last-known value, and the error
--- a/apps/mobile/src/screens/systems/useSystemsData.ts
+++ b/apps/mobile/src/screens/systems/useSystemsData.ts
@@ -58,7 +58,9 @@ function alertOrgId(a: Alert): string | undefined {
   return typeof id === 'string' ? id : undefined;
 }
 
-// Fetches summary, alerts, devices, and orgs in parallel. Owns the local
+// Fetches summary, alerts, activeAlerts, devices and orgs in parallel — five,
+// matching `const total = 5` in mergeSystemsResults. Miscounting here is not
+// cosmetic: that total is the threshold for "everything failed". Owns the local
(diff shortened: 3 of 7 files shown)


ToddHebebrand:
All four items landed, and two things you added weren't in the issue.

The Sentry warning when the walk exhausts its page budget is the right instinct — the issue asked for the ceiling to be *deliberate*, and you made it *observable*, which is strictly better. And setting `cursor = null` on clean exhaustion so the post-loop `if (cursor)` can't false-positive is a real subtlety I didn't call out; without it the warning would fire on every complete walk and become noise inside a week.

The test comment is the part worth keeping: *"the test agreed with the comment and both were wrong about the server, which is why neither caught it."* That's the actual lesson from this one. A self-consistent client — code, comment, and assertion all agreeing — is exactly the shape no amount of code review catches, because there's nothing internally inconsistent to notice. It only surfaces by reading the other side of the API. Worth remembering next time a constant cites a cap.

One follow-up, not blocking: the truncation is now visible to *us* via Sentry but still not to the *operator* on the device. At 2,000 that's academic today. If mobile ever targets fleets that size, the "showing N of more" affordance is the remaining half.

PR #3760, opened 2026-08-21
fix(api): lock the devices row before the cascade deletes its children

### LINKED ISSUE #3759: [API] deleteDeviceCascade takes the same inverted lock order that caused the #3739 deadlock — lock the devices row first
## Description

`deleteDeviceCascade` deletes every `device_id`-FK child table before the `devices` row, which is the same child-first lock order that caused the production deadlock fixed in #3739 (Sentry BREEZE-1S, Postgres `40P01`).

- `apps/api/src/services/deviceDeletion.ts` — deletes the child tables listed in `CORE_DEVICE_CASCADE_DELETE_TABLES` (`apps/api/src/routes/devices/core.ts:222`), then the `devices` row last (`deviceDeletion.ts:83`).
- Called inside a transaction from `apps/api/src/routes/devices/core.ts:1584-1587` (`DELETE /devices/:id/permanent`).

Every other writer spanning both levels takes the `devices` row first, then children — re-enrollment (`enrollment.ts` ~:831 then ~:929), the site-change path (`core.ts:1308`), and now `PUT /agents/:id/network` after #3739. A permanent delete racing any of those on the same device is a textbook AB-BA.

## Why this one can't just be reordered

Unlike #3739, the order here is forced: the FK constraints require children to go before the parent row. Deleting `devices` first is not an option.

The conventional fix is to take the parent lock explicitly at the top of the transaction, before touching any child table:

```sql
SELECT id FROM devices WHERE id = $1 FOR UPDATE
```

That establishes the same devices-first ordering every other writer uses, without violating FK delete order. It also gives the cascade a clean serialization point against concurrent agent writes.

## Severity: low, but real

The window is genuinely narrow — permanent deletion requires `status = 'decommissioned'` first, so a device being deleted is unlikely to have a live agent reporting inventory. Narrow is not zero: decommissioning and deletion can happen close together, an agent can still be mid-flight when the status flips, and re-enrollment of a device someone is deleting is exactly the kind of race that shows up once in production and then never reproduces locally.

Worth fixing while the lock-ordering convention is fresh and documented, rather than rediscovering it from another `40P01` in Sentry.

## How this was found

Swept during review of #3739: after fixing the one known inverted writer, I enumerated every transaction touching both `devices` and a device-child table to see whether any others had the same shape. This was the only remaining case. Confirmed NOT at risk: `core.ts:1308` (devices first, correct), `networkBaseline.ts:738` (its `FOR UPDATE` is on `network_baselines`, and its `devices`/`device_network` access is read-only), and `discoveryWorker.ts` / `dnsSyncJob.ts` / `s1Sync.ts` / `partnerApi/inventory.ts` / `partnerApi/relationships.ts` / `devices/hardware.ts` (no transaction blocks at all, so no lock is held across statements).

bdunncompany:
Fixes #3759. Your enumeration made this straightforward — thanks for listing the transactions you'd confirmed *not* at risk, that saved re-deriving it.

`SELECT ... FOR UPDATE` on the devices row as the first statement, inside `deleteDeviceCascade` rather than at the call sites. It's the one choke point both `DELETE /devices/:id/permanent` and the Quick Support reaper go through, so a future caller can't forget it.

**One thing I want to flag, because my first version claimed a guarantee it hadn't established.** Issuing the statement isn't the same as holding the lock. `FOR UPDATE` locks only rows it can *see*, so it locks nothing when:

- the device row is already gone — a re-run, or two reapers racing, and
- **RLS filters it.** This matters here specifically: the cascade runs inside the caller's tenant-scoped context, and this file's own comment notes that `abuse_endpoint_fingerprints` is deliberately invisible under that policy. If the parent is filtered while any child is reachable, the cascade proceeds under exactly the child-first ordering the lock exists to remove.

Aborting on zero rows would be wrong — deleting an absent device is legitimately idempotent, and two reapers racing would error instead of one finding nothing. So it reports instead. An unserialized cascade is now *visible* rather than silently assumed safe, and the comment states the precondition rather than implying the lock is unconditional.

**On the tests, honestly scoped.** They assert the ordering contract — lock issued before any child table, devices row deleted last — and both zero-row branches. They do **not** prove Postgres acquired a row lock, that RLS didn't filter it, or that two transactions serialize; that needs two real connections and belongs in an integration test. What they do cover is what a refactor can silently break, which is the failure mode that produced this issue.

deviceDeletion 4 pass, quickSupportReaper 14 pass, `src/routes/devices` 506 across 38 files, `tsc --noEmit` clean. Removing either the lock or the zero-row report fails its test.

**One question I deliberately did not decide for you.** This converts a fast `40P01` into a potentially unbounded lock wait on a request-path DELETE. There's no `lock_timeout` convention in the repo, though there is a transaction-local `statement_timeout` precedent (`softwareInventory.ts:378`, `filterEngine.ts:569`). Bounding it would make the conflict fast and retryable rather than a hang — but that changes failure semantics on a delete path, so it seemed like your call rather than something to slip into a lock-ordering fix. `SKIP LOCKED` would be wrong here: it recreates the zero-row-unlocked problem.

### DIFF
--- a/apps/api/src/db/rowCount.ts
+++ b/apps/api/src/db/rowCount.ts
@@ -1,0 +1,30 @@
+/**
+ * Row count out of a driver result, across the three shapes this codebase sees.
+ *
+ * `drizzle-orm/postgres-js` `execute()` resolves to a postgres-js `Result`,
+ * which extends Array and carries `.count` — NOT node-postgres' `.rowCount` /
+ * `.rows`. Reading the wrong one yields 0 on every call, which in a batched
+ * delete ends the loop early and silently leaves rows behind, and in a lock
+ * check reports the exact opposite of the truth.
+ *
+ * This lives in `db/` rather than in one of the retention jobs so that a
+ * REQUEST-PATH service can use it without importing from `jobs/`, which would
+ * invert services→jobs and pull BullMQ + ioredis into the module graph of a
+ * plain DELETE.
+ *
+ * Note there are still several private copies of this check in `jobs/` and
+ * other services. Consolidating them is worthwhile but is not this change.
+ */
+export function extractRowCount(result: unknown): number {
+  // Deliberately NOT null-safe. postgres-js never resolves a successful
+  // statement to null/undefined — a genuine zero-row result is an array-like
+  // `Result` with `.count === 0`. Mapping null to 0 would conflate "broken
+  // driver/adapter/mock" with "no rows", and 0 is load-bearing in both kinds of
+  // caller: here it means "the parent lock was not held", and in the batched
+  // retention loops it TERMINATES the loop and leaves old rows behind. Letting
+  // the property read throw keeps a broken contract loud.
+  const raw = result as { rowCount?: number; count?: number };
+  if (typeof raw.rowCount === 'number') return raw.rowCount;
+  if (typeof raw.count === 'number') return raw.count;
+  return Array.isArray(result) ? (result as unknown[]).length : 0;
+}
--- a/apps/api/src/routes/devices/cascadeDelete.test.ts
+++ b/apps/api/src/routes/devices/cascadeDelete.test.ts
@@ -94,6 +94,7 @@ import {
   DEVICE_LINKED_DEVICE_ID_TABLES,
 } from './core';
 import { db } from '../../db';
+import { isAgentConnected, sendCommandToAgent } from '../agentWs';
 
 const deviceCascadeDeleteTables = getDeviceCascadeDeleteTables();
 
@@ -406,23 +407,78 @@ describe('DELETE /devices/:id/permanent — tickets are detached, not destroyed'
    * 500 with no indication the uninstall had already gone out. Assert the
    * WRAPPED shape specifically — an unwrapped `{ code: '55P03' }` fixture would
    * pass against the broken code and prove nothing.
+   *
+   * Uses an ONLINE device with the uninstall actually dispatched. An earlier
+   * revision used the offline fixture, where `uninstallSent` can only ever be
+   * false, and merely asserted the property existed — code that hard-coded
+   * `false` would have passed it.
    */
-  it('maps a Drizzle-wrapped 55P03 to a retryable 409 that reports uninstallSent', async () => {
-    rigDeviceLookup(DEVICE);
+  // Still `decommissioned` — the route 400s anything else BEFORE it ever
+  // reaches the uninstall dispatch, so an 'online' status here would have
+  // tested the wrong branch entirely. What makes the uninstall fire is an
+  // agentId plus a live agent connection.
+  const CONNECTED_DEVICE = { ...DEVICE, agentId: 'agent-1' };
+
+  function rigUninstallDispatched() {
+    vi.mocked(isAgentConnected).mockReturnValue(true);
+    // Returns synchronously — `uninstallSent = sendCommandToAgent(...)`, not an
+    // awaited promise, so mockResolvedValue would make uninstallSent a Promise.
+    vi.mocked(sendCommandToAgent).mockReturnValue(true as never);
+  }
+
+  it('maps a Drizzle-wrapped 55P03 to a retryable 409 that reports uninstallSent: true', async () => {
+    rigDeviceLookup(CONNECTED_DEVICE);
+    rigUninstallDispatched();
     vi.mocked(db.transaction).mockImplementation(async () => {
       throw Object.assign(new Error('Failed query: SELECT id FROM devices ... FOR UPDATE'), {
         cause: Object.assign(new Error('canceling statement due to lock timeout'), { code: '55P03' }),
       });
     });
 
-    const res = await app.request(`/devices/${DEVICE.id}/permanent`, {
+    const res = await app.request(`/devices/${CONNECTED_DEVICE.id}/permanent`, {
       method: 'DELETE',
       headers: { Authorization: 'Bearer t' },
     });
 
     expect(res.status).toBe(409);
     const body = await res.json() as { error: string; uninstallSent: boolean };
-    expect(body).toHaveProperty('uninstallSent');
-    expect(body.error).toMatch(/busy/i);
+    expect(body.uninstallSent).toBe(true);
+    expect(body.error).toMatch(/already sent/i);
+    expect(body.error).toMatch(/retry/i);
+  });
+
+  /**
+   * The 23503 branch had NEVER executed before the unwrap — the top-level
+   * `.code` read meant it was unreachable. Switching it on is a behaviour
+   * change, so it gets the same online-agent coverage: a rolled-back cascade
+   * leaves the row present while the agent may already be uninstalling, and the
+   * web callers surface only `err.message`, so the disclosure must be in the
+   * text and not merely in the JSON field.
+   */
+  it('maps a Drizzle-wrapped 23503 to a 409 that discloses the already-sent uninstall', async () => {
+    rigDeviceLookup(CONNECTED_DEVICE);
+    rigUninstallDispatched();
+    vi.mocked(db.transaction).mockImplementation(async () => {
+      throw Object.assign(new Error('Failed query: DELETE FROM devices'), {
+        cause: Object.assign(new Error('update or delete violates foreign key constraint'), {
+          code: '23503',
+          detail: 'Key (id)=(...) is still referenced from table "some_child".',
+          table_name: 'some_child',
+        }),
+      });
+    });
+
+    const res = await app.request(`/devices/${CONNECTED_DEVICE.id}/permanent`, {
+      method: 'DELETE',
+      headers: { Authorization: 'Bearer t' },
+    });
+
+    expect(res.status).toBe(409);
+    const body = await res.json() as { error: string; uninstallSent: boolean };
+    expect(body.uninstallSent).toBe(true);
+    // table_name must come off the SAME node as the code, or this reads
+    // "related records in undefined".
+    expect(body.error).toContain('some_child');
+    expect(body.error).toMatch(/already sent/i);
   });
 });
--- a/apps/api/src/routes/devices/cascadeDelete.test.ts
+++ b/apps/api/src/routes/devices/cascadeDelete.test.ts
@@ -293,17 +293,28 @@ describe('DELETE /devices/:id/permanent — tickets are detached, not destroyed'
         execute: vi.fn().mockImplementation(async (q: any) => {
           const text = sqlToText(q);
           statements.push(text);
-          // The cascade reads the caller's lock_timeout so it can restore it
-          // after taking the parent row lock, and REFUSES to change the setting
-          // if it cannot read it back (a wrong-shaped result would otherwise
-          // silently widen a stricter caller). A bare [] here is not a result
-          // this driver can produce, and made the route 500.
-          if (text.includes('current_setting')) {
-            const row: Record<string, unknown>[] = [{ lock_timeout: '0' }];
-            Object.defineProperty(row, 'count', { value: 1, enumerable: false });
-            return row;
+          // postgres-js resolves to an array-like carrying a non-enumerable
+          // `.count`. A bare [] is not a result this driver can produce, and
+          // returning one made every statement look like it affected 0 rows.
+          const pgResult = (rows: Record<string, unknown>[], count = rows.length) => {
+            Object.defineProperty(rows, 'count', { value: count, enumerable: false });
+            return rows;
+          };
+          // The cascade tightens the lock bound and reads the caller's prior
+          // value in ONE pg_settings statement (milliseconds, as an integer).
+          // '0' is Postgres's "wait forever", the value that makes it actually
+          // apply its 3s bound, so this keeps the mock on the interesting path.
+          if (text.includes('pg_settings')) {
+            return pgResult([{ prior_ms: '0' }]);
           }
-          return [];
+          // The parent row lock must report ONE row. Returning [] here sent the
+          // whole route suite down the "ran without holding the lock" branch,
+          // so a regression that permanently lost the lock would have been
+          // invisible to these tests.
+          if (text.includes('FOR UPDATE')) {
+            return pgResult([{ id: DEVICE.id }]);
+          }
+          return pgResult([]);
         }),
         delete: vi.fn().mockReturnValue({
           where: vi.fn().mockResolvedValue(undefined),
--- a/apps/api/src/routes/devices/cascadeDelete.test.ts
+++ b/apps/api/src/routes/devices/cascadeDelete.test.ts
@@ -391,4 +391,38 @@ describe('DELETE /devices/:id/permanent — tickets are detached, not destroyed'
     expect(res.status).toBe(200);
     expect(dissolveLinkGroupIfBelowMinimum).not.toHaveBeenCalled();
   });
+
+  /**
+   * A lock timeout must surface as the retryable 409 that reports
+   * `uninstallSent`, because the SELF_UNINSTALL is dispatched BEFORE the
+   * transaction and is irreversible — a bounded lock failure is the one path
+   * that can leave an agent uninstalling itself while its device row survives.
+   *
+   * The shape here is not invented. It is what a REAL lock timeout produces:
+   * verified against live Postgres with two connections contending on one row,
+   * the error arrives as `{ code: undefined, cause: { code: '55P03' } }`,
+   * because Drizzle wraps the postgres-js PostgresError. The route used to read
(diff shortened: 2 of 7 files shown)


ToddHebebrand (changes requested):
The lock itself is right, and putting it inside `deleteDeviceCascade` rather than at the two call sites is the correct choice — a future caller can't forget it. The honesty about `FOR UPDATE` locking only visible rows is the part I'd have been most likely to get wrong myself, and the RLS branch is a real hazard, not a theoretical one.

**One blocker, and it's in the reporting rather than the lock.**

`deviceDeletion.ts` reads the result as:

```ts
const lockedRows = locked?.rowCount ?? locked?.rows?.length ?? 0;
```

Both of those properties are the **node-postgres** result shape. This API is on `drizzle-orm/postgres-js` (`apps/api/src/db/index.ts:8`), whose `execute()` resolves to an array-like carrying `.count` — not `.rowCount`, and not `.rows`. So both sides of the `??` chain are `undefined` and `lockedRows` is **0 on every call**, including the ordinary success path where the lock was genuinely acquired.

The effect is the inverse of what the code intends: instead of surfacing the rare unserialized cascade, it fires `device cascade ran without holding the devices row lock` on **every single device deletion**. The one signal that would tell us the RLS branch actually happened is buried under a permanent false positive — which is worse than not reporting at all, because the alert now reads as noise and gets muted.

The repo already settled this. There are **six** copies of `extractRowCount`, and every one checks three shapes in order:

```ts
if (typeof raw.rowCount === 'number') return raw.rowCount;
if (typeof raw.count === 'number') return raw.count;
return Array.isArray(result) ? result.length : 0;
```

(`services/tenantCascade.ts:803`, `jobs/auditRetention.ts:106`, `jobs/deviceMetricsRetention.ts:71`, `jobs/processSampleRetention.ts:37`, `jobs/serviceProcessCheckRetention.ts:73`, `services/metricRollupMaintenance.ts:127`.) `metricRollupMaintenance.ts:138` independently confirms the driver shape by reaching for `Array.isArray(result)` and indexing `result[0]`. Import one rather than adding a seventh copy — `deviceMetricsRetention` and `processSampleRetention` both export theirs.

**Worth naming: the tests can't see this, for the exact reason you identified over on #3769.** The mock returns `{ rowCount: lockedRowCount, rows: [] }` — a contract the driver has never had — so code, mock and assertion all agree with each other and are all wrong about the driver. That's the same shape as `devicesPaging.test.ts` returning a cursor on page one. Same lesson, opposite direction, landing in two PRs on the same day.

**Your `lock_timeout` question — yes, bound it, and I'd take it in this PR.**

You're right that this converts a fast `40P01` into a potentially unbounded wait, and on a request-path DELETE that's the worse failure for us specifically: a hung transaction holds its pooled connection, and connection-hold has been a recurring production incident here rather than a hypothetical (the US instance runs against a hard connection ceiling). A deadlock at least resolves itself in milliseconds; a lock wait behind a long-running site move does not.

Precedent is transaction-local `set_config`, exactly as you found — `routes/softwareInventory.ts:378` and `services/filterEngine.ts:569`. Same shape here: set a transaction-local `lock_timeout` before the `FOR UPDATE`, and let `55P03` surface as a retryable conflict rather than a hang. That makes the failure fast, explicit and diagnosable, which is the same thing the rest of this PR is doing.

**On the test scoping — your framing is correct and I don't want more here.** Asserting the ordering contract and both zero-row branches covers what a refactor silently breaks, which is what produced #3759. Proving actual lock acquisition needs two real connections; if you want it, it belongs in an integration test, and I'd take that as a follow-up rather than a condition of merging.

Fix the row-count read, add the `lock_timeout`, and I'll merge.

bdunncompany:
Both fixed in `0e877a4`. The row-count read was exactly as you described, and chasing the `lock_timeout` turned up a scope problem I had not accounted for.

**Row count.** Now imports `extractRowCount` from `deviceMetricsRetention` rather than adding a seventh copy. Your point about the mock was the more useful half: it returned `{ rowCount, rows: [] }`, a shape this driver has never produced, so the code, the mock and the assertion all agreed with each other and were all wrong. The mock now returns a faithful postgres-js result (array-like carrying a non-enumerable `.count`). I control-tested it — reverting the read to `.rowCount ?? .rows?.length` now fails `stays quiet on the normal path`, which it could not do before.

**`lock_timeout` — it needed more than adding the statement.** `set_config(..., true)` is transaction-local, not statement-local, and Postgres applies `lock_timeout` to every subsequent lock acquisition until the transaction ends. Both call paths already sit inside an outer transaction (`withDbAccessContext` opens one at `db/index.ts:465`), where a nested `db.transaction()` is only a savepoint — and releasing a savepoint does not undo a `SET LOCAL`.

So the naive version would have quietly governed every child `DELETE`, the route's own link-group dissolution at `core.ts:1593`, and — in the reaper, which wraps its whole pass in one context at `quickSupportReaper.ts:59` — every device purged after the first one. Not the "bounds the parent-row lock" the comment claimed.

It now captures `current_setting('lock_timeout')` and restores it as soon as the lock is held. The restore is deliberately **not** in a `finally`: a `55P03` aborts the subtransaction, so a statement issued there would fail with `25P02` and mask the error you actually need to see, and rollback to the preceding savepoint already undoes the `SET LOCAL`. If the prior value can't be read it aborts rather than guessing — `'0'` means *wait forever*, so substituting it would silently widen a stricter caller for the rest of the transaction.

**Mapped `55P03` to 409.** You said you wanted it surfaced as a retryable conflict; nothing was mapping it, so it would have become a generic 500 via `index.ts:1203`. The route only recognised `23503`.

**One thing I want your call on, because my change created the exposure.**

The best-effort `SELF_UNINSTALL` is dispatched at `core.ts:1564`, *before* the deletion transaction, and it is irreversible. Previously a contended delete just waited and eventually succeeded. With a bounded wait there is now a path where the agent uninstalls itself and the device row survives — an unmanageable orphan if the operator never retries.

I did not reorder the dispatch, since sending it first is presumably deliberate (catch the agent while it is known online) and that is a behavioural change beyond this fix. For now the 409 reports `uninstallSent` and tells the operator a retry is required rather than merely advisable. If you would rather the uninstall move to after commit, I am happy to do it here or as a follow-up — your call on which risk you prefer.

**Coverage, stated plainly.** The unit tests prove statement *ordering* and the row-count contract. They cannot prove the timeout actually fires, that the lock is held across the cascade, or that the setting is restored on a real connection — the mock has no transaction, savepoint or GUC semantics. That needs two real connections, and I would take it as the integration test you already offered to accept as a follow-up.

Gate: eslint clean, `tsc --noEmit` clean, 33 tests across `deviceDeletion`, `quickSupportReaper` and `deviceMetricsRetention`. Both new assertions were control-tested by reverting the fix and confirming they fail.

bdunncompany:
CI went red on `0e877a4` and I've fixed it in `0fe0117`. Owning what happened, because it's your review point landing on me twice.

`cascadeDelete.test.ts` mocks `tx.execute` to return `[]` for every statement. The cascade now reads `current_setting('lock_timeout')` to restore the caller's value after taking the lock, and it deliberately throws rather than guessing when it can't read that value back — substituting `'0'` would mean *wait forever* and silently widen a caller that had set something stricter. A bare `[]` isn't a result this driver can produce, so the throw fired and the route 500'd on three assertions.

The throw is right, and I'd rather have this evidence that it's reachable than not. The mock was the wrong half, and it's now returning a real row array.

That's the same defect you flagged one file over: a mock asserting a driver contract that has never existed. I fixed the two files I was looking at and didn't check who else mocked the same call, which is exactly the completeness miss worth naming.

**Gate this time was the full `apps/api` suite**, not the three files I'd changed: 23,155 tests, 0 failures. One file (`aiGuardrailsTierConfig.parity.test.ts`) needs `apps/web` installed to transform since it imports a web module for parity — verified passing separately, 4 tests. eslint and `tsc --noEmit` clean.

The substantive question from my last comment still stands whenever you get to it: whether the best-effort `SELF_UNINSTALL` at `core.ts:1564` should move to after the transaction commits, given the bounded wait now makes an orphaned device row reachable.

ToddHebebrand (changes requested):
**All three points from the bounce are correctly fixed, and I checked the driver claim against source rather than taking either of our words for it.** `drizzle-orm/postgres-js` `execute()` → `client.unsafe()` → a postgres-js `Result extends Array` with `.count` and no `rowCount`/`rows`; `db/index.ts:31` sets no `transform`, so `[0].lock_timeout` comes back under that name. `extractRowCount` hits the `.count` branch, both mocks now return that shape, and the only other cascade caller (`jobs/quickSupportReaper.ts:179`) mocks `services/deviceDeletion` wholesale, so there is no third stale mock. The four Integration shards exercise the real driver through `quickSupportChain.integration.test.ts:89`'s `afterEach` cascade, which is the independent signal I wanted.

**One must-fix, and it's the throw you defended.** `extractLockTimeout` at `services/deviceDeletion.ts:44-50` aborts the delete when it can't read the setting back. `current_setting` is a function, not a table, so RLS can't filter it and the throw is unreachable today — agreed. But *if* it ever fires, it fires as the first statement inside the transaction, i.e. **after** `SELF_UNINSTALL` has already gone out at `core.ts:1564`, and it carries no `code`, so `core.ts` rethrows to a 500 that says nothing about `uninstallSent`. "Refuse to widen a stricter caller" is fully satisfied by *not setting* the bound and proceeding with the pre-PR unbounded wait — which is strictly safer than aborting a delete whose uninstall is already in flight. Skip the `set_config` + report it (the `services/sentry.ts` wrapper, not raw `@sentry/node` — it carries the init guard and `scrubEvent`), don't throw. Keep the test, inverted.

While you're in that function: the 3000ms is set unconditionally, which *does* widen a caller that had set something stricter — the exact thing the doc comment says the function refuses to do. Skip the set when the prior value is non-`'0'` and smaller, and code and comment will agree.

**Nice-to-have, not blocking:** `deviceDeletion.ts:7` imports `extractRowCount` from `jobs/deviceMetricsRetention`, which drags BullMQ + ioredis into a request-path DELETE's module graph and inverts services→jobs — a shared `db/` util both can import is the right home. And `cascadeDelete.test.ts:294` still returns `[]` for the `FOR UPDATE`, so the route suite now always walks the "ran without the lock" branch; give it `pgResult(1)` so a regression that permanently loses the lock is visible there too.

**Your `SELF_UNINSTALL` question — you're right, and it's split out: #3817.** The dispatch at `core.ts:1561-1571` precedes `db.transaction` at `:1584`, so uninstall-sent-but-row-survives is reachable on 55P03. It is not new, though: the pre-existing `23503` branch at `:1599` has the identical hazard and doesn't even report `uninstallSent`. Moving the dispatch post-commit is mechanical (`device.agentId` is in scope) but changes three error paths and wants its own test, which is a different PR than a lock-ordering fix. Your 409 message is a fine interim mitigation.

Merges on the throw→skip change.

bdunncompany:
All four points addressed, plus a **P1 that your review indirectly uncovered and that invalidates a claim I made earlier in this PR**. Worth a careful read rather than a diff skim.

### The correction first

I said this PR's bound "fails fast into a retryable 409". **It did not, and could not.** The route read `(err as { code?: string })?.code`, but `drizzle-orm/postgres-js` rethrows the postgres-js `PostgresError` inside a `DrizzleQueryError` whose own `.code` is undefined — the SQLSTATE is on `.cause`.

I reproduced it rather than reasoning about it: live Postgres 16, two connections contending on one row, `lock_timeout='300ms'`, through the real Drizzle transaction path.

```
topLevelCode : null       <- what the route was reading
causeCode    : "55P03"
pgErrorCode  : "55P03"
```

So **both** branches in that catch were dead — the `55P03` one I added, and the pre-existing `23503` one. A bounded lock failure became a generic 500 with no `uninstallSent`, which is the one thing an operator needs when the irreversible `SELF_UNINSTALL` has already gone out. `utils/pgErrors.ts` documents this exact hazard and exports `pgErrorCode()` for it; the route hand-rolled the broken read anyway.

Switching `23503` on is a behaviour change — it had never executed — so it got the same treatment: it now reports `uninstallSent` and says so in the message **text**, since the web callers surface only `err.message`. I also added `pgErrorNode()`, because unwrapping just the code left `detail`/`table_name` on the outer error, which renders as "related records in **undefined**". And I softened "This table may need to be added to the cascade delete list" to "may be", since the same catch also covers `dissolveLinkGroupIfBelowMinimum` — a link-group FK violation is not necessarily a missing cascade entry.

### Your four points

**1. `extractLockTimeout` throw → skip.** Done, but not the way I first wrote it. Skipping the bound and proceeding leaves an *unbounded* `FOR UPDATE` on a request path holding a pooled connection — the outage class the bound exists to prevent, and arguably worse than the abort. So the tighten now happens **in SQL**:

```sql
WITH prior AS (SELECT setting::bigint AS ms FROM pg_settings WHERE name = 'lock_timeout')
SELECT prior.ms AS prior_ms,
       set_config('lock_timeout',
         (CASE WHEN prior.ms = 0 OR prior.ms > 3000 THEN 3000 ELSE prior.ms END)::text || 'ms',
         true) AS applied
FROM prior
```

`pg_settings.setting` for `lock_timeout` is a plain millisecond integer (`0`, `250`, `7000`), unlike `current_setting`, which renders `250ms`/`3s`/`2min` and needs a client-side unit parser. That parser is gone, and the abort-vs-hang choice with it: the timeout is always bounded before the lock runs, so an unreadable `prior_ms` costs only the restore — which is skipped, leaving the **stricter** value in force.

Verified live: caller at `0` → applied `3s`; caller at `500ms` → **preserved, not widened**; `'0ms'` round-trips to `0`; and `pg_settings` has exactly one row for `lock_timeout`, so the CTE cannot yield zero rows and silently skip the bound.

**2. Unconditional 3000ms widening a stricter caller.** Fixed by the `CASE` above, and the restore is skipped when nothing changed.

**3. `extractRowCount` importing from `jobs/`.** Moved to `db/rowCount.ts`, so a request-path DELETE no longer drags BullMQ + ioredis into its module graph. I left the other private copies alone — consolidating them is worthwhile but isn't this change.

**4. `cascadeDelete.test.ts` returning `[]` for the `FOR UPDATE`.** Fixed; the route suite no longer always walks the "ran without the lock" branch.

### One thing I got wrong in the reporting

My first attempt reported the fallback through the `services/sentry` wrapper with `{ area: 'device-deletion' }` — which produces a **blank event**. `captureMessage` does `void extra`, `area` isn't in `ALLOWED_TAG_NAMES`, and `scrubEvent` deletes `message`, `logentry` and `extra`. The `#3214` comment in that allowlist warns about precisely this. Now tagged `device_deletion_warning` with two closed values, allowlisted alongside the existing entries.

### Verification

`tsc` clean. **23,163 unit passed, 0 failed.** Integration **2,000 passed across 4 shards, each on a fresh database**. Both new route tests assert the **wrapped** shape and use a *connected* device with the uninstall actually dispatched — an earlier revision used the offline fixture, where `uninstallSent` can only be false and code hard-coding `false` would have passed. Controls run: reverting the unwrap turns both from 409 into 500; reading `table_name` off the outer error alone fails the `23503` case.

### Still open, your call

`emitCatalogEvent`-style concern applies here too: I did **not** move the `SELF_UNINSTALL` dispatch after commit (#3817), so the 409 remains the interim mitigation you called it. And there is no two-connection *integration* test driving the real route under contention — the live reproduction above is what stands in for it.

ToddHebebrand (approved):
**Approved — all four points verified in the diff, and the unwrap find is the best thing in this PR.**

I checked the load-bearing claims rather than taking the writeup on faith: `pg_settings.setting` for `lock_timeout` is indeed integer milliseconds, the CTE tightens-and-reads in one statement with the CASE preserving a stricter caller, the restore correctly runs only on the success path (a `finally` after 55P03 would 25P02 over the error the caller needs), and both new route fixtures assert the **wrapped** `{code: undefined, cause: {code}}` shape with diagnostics read off the same node — an unwrapped fixture would have passed against the broken code, and you built the control for exactly that. The Sentry tag rides the #3214 allowlist pattern correctly, and `deviceDeletion.test.ts` now documents both what the ordering mock proves and what it deliberately cannot.

**Two notes, neither blocking:**

- `db/rowCount.ts` is described as a move but is a seventh copy — the six private ones in `jobs/` and `services/` still exist (`deviceMetricsRetention.ts:71`, `tenantCascade.ts:803`, etc.). Right call not to consolidate here; it is now the canonical home, so the consolidation follow-up should point everything at it. I'll file that.
- There's a doubled comment block at the restore decision in `deviceDeletion.ts` (two consecutive "Restore only if…" paragraphs from successive revisions) — fold into one whenever you next touch the file; not worth a round trip.

The two open items stay where we put them: SELF_UNINSTALL post-commit is #3817, and the two-connection contention test is an integration follow-up, not a merge condition — your live reproduction is the evidence that matters for this round.

PR #3769, opened 2026-08-21
fix(mobile): stop Systems and Tickets presenting partial state as complete (#3753)

### LINKED ISSUE #3753: [Mobile] Systems and Tickets present failed/truncated state as complete — 4 silent failures in the #3716/#3719 batch
## Description

A silent-failure audit of the mobile Systems and Tickets screens (added in #3716 and #3719, both merged and **unreleased**) found four ways the UI presents an incomplete or failed state as a confident, complete one. Grouping them because they share a root cause — the screens distinguish "loading" from "loaded" but not "loaded" from "partially failed" — and fixing them together is cheaper than four passes.

Each was verified against `origin/main`, not inferred.

## 1. Tickets: "The open queue is clear" renders *alongside* the error — HIGH

`apps/mobile/src/screens/tickets/TicketsScreen.tsx:151-158`. `ListEmptyComponent` is gated on `loading` only:

```jsx
ListEmptyComponent={
  loading ? undefined : (<View style={styles.empty}>...</View>)
}
```

When `fetchTickets` rejects, `ticketsSlice` sets `error` but leaves `tickets` empty, so the screen renders the error line *and*, immediately below it, the reassuring empty-state copy — "No open tickets — The open queue is clear." Two contradictory messages in one viewport. A tech skimming past the error reads "queue is clear" and moves on.

`TicketDetailScreen.tsx` already gets this right with its `loadError && !ticket` branch and a "Try again" affordance — the list screen should mirror it. Fix: gate on `!loading && !error`, and give the error case its own retry-oriented state.

## 2. Active issues silently truncate at 50 — HIGH

`apps/mobile/src/services/api.ts:445-449`:

```js
export async function getAlerts(status: 'active' | 'all' = 'active'): Promise<Alert[]> {
  const path = status === 'active' ? '/alerts/inbox?status=active' : '/alerts/inbox';
  const response = await request<ListResponse<MobileAlertRecord>>(path);
  return response.data.map(mapAlert);
}
```

No `limit`, no `cursor`, and `response.pagination` is discarded entirely. The route defaults to `limit=50` (`apps/api/src/routes/mobile.ts:57-59`). So Active Issues and the org rollup counts only ever see the first 50 by recency, dropping any active alert outside that window — including criticals — with no "showing 50 of N" anywhere.

This matters most on exactly the tenants #3719 was written for: the PR cites a 7,023-alert tenant. `getDevices` in the same file already walks a cursor correctly — worth mirroring that, or at minimum surfacing `pagination.total` and showing a truncation notice.

This is the same class as #3612 (device pickers truncating silently).

## 3. `fetchAll` can leave the screen spinning forever — HIGH

`apps/mobile/src/screens/systems/useSystemsData.ts:80-132`. The rewrite to `Promise.allSettled` dropped the old catch-all, leaving only try/finally. `allSettled` cannot reject, but code between it and the `setData` call can still throw synchronously — the `rejectionReasons` loop calls `reportInternalError` → `Sentry.captureException`, and `mergeSystemsResults` runs inside the `setData` updater.

If either throws, `fetchAll` rejects with no catch anywhere in the chain: `useEffect(() => { fetchAll('initial') })`, `void fetchAll('refresh')`, and `onRefresh` (`SystemsScreen.tsx:122-124`) all discard the rejection. Because the throw precedes the state update that clears `loading`/`refreshing`, the skeleton or pull-to-refresh spinner spins indefinitely with no banner and no log.

Fix: restore a catch around the whole body that reports and sets `loading:false, refreshing:false` plus a generic error, in addition to the per-slice `allSettled` handling.

## 4. A failed `orgs` slice mislabels orgs as "Unknown organization" — MEDIUM

`useSystemsData.ts:216-252`. `orgRollups` and `filterOrgName` resolve names via `data.orgs.find(...)` and fall back to the literal `'Unknown organization'`, with no distinction between "this org genuinely isn't in the list" and "the orgs fetch failed and we're rendering under a partial-failure banner." Same root cause as the generic-banner problem below — a partial failure produces confidently-labelled UI.

Fix: when `orgs` is in the merge's failed set, flag or suppress org-name-dependent rows rather than rendering the placeholder identically to a real unknown.

## Related, already noted on the PRs

- The Systems error banner (`SystemsScreen.tsx:302-311`) hardcodes its text and never reads the computed `ALL_FAILED_MESSAGE` / `PARTIAL_FAILED_MESSAGE`, so "nothing loaded" and "some of this is stale" are indistinguishable. That directly feeds #4.
- #3727 (still open) has two related findings posted on the PR: the post-ack un-hide runs even when the refetch failed, and the bulk worker discards the error entirely.

Found during the 2026-08-20 GH queue sweep, running a dedicated silent-failure pass over the mobile batch. All four are in merged-but-unreleased code, so they can be fixed before this ships.

bdunncompany:
Fixes #3753.

All four, plus the banner you flagged as feeding #4. Every one is in
merged-but-unreleased code, so nothing here is user-facing yet.

## 1. Tickets empty state (HIGH)

Gated on `error` as well as `loading`, so "The open queue is clear." can no
longer render directly beneath the error line. The failure case gets its own
copy and a "Try again" affordance mirroring `TicketDetailScreen`.

Extracted `emptyStateKind(loading, error)` into `ticketCopy.ts` and tested it
there — that folder already tests its copy decisions as pure functions, and
mobile has no renderer harness.

## 2. The 50-alert truncation — but not fixed the way the issue suggests

`getAlerts` sent no `limit` and discarded `pagination` entirely. It now asks for
the server maximum and, more importantly, **reports whether that was
everything**.

The issue suggested mirroring `getDevices`, "which already walks a cursor
correctly". **It does not, and that turned out to be the bigger find:
`getDevices` was returning one page and calling it a fleet.**

`/mobile/devices` and `/mobile/alerts/inbox` both compute `nextCursor` only
inside `if (cursor)` (`mobile.ts:1156`). A cold-start caller has no cursor to
send, so its first response always carries `nextCursor: null` and it can never
obtain the token for page two. The walk treated that null as *clean exhaustion*,
so it set `cursor = null` and its own truncation warning could never fire. On a
216-device tenant the app showed 100 and looked authoritative.

`devicesPaging.test.ts` did not catch it because its mock returned a cursor on
page one — a contract the server has never had. Code, comment and test all
agreed with each other about the server, which is exactly why review could not
see it. Replaced with `listPaging.test.ts`, whose fixtures return
`nextCursor: null` like the real route.

**I tried both alternatives and rejected both:**

- *Making cursor mode reachable* (always over-fetch `limit + 1`). Passes, but it
  would ship silent row-skipping: `devices.last_seen_at` is nullable and DESC
  puts NULLs first while `encodeCursor(null, …)` returns null, so a page ending
  on a null row terminates the walk; `alerts.triggered_at` is Postgres
  microsecond precision while `encodeCursor` uses `Date.toISOString()`
  (milliseconds), so `< truncated_ts` skips rows inside the same millisecond;
  and `last_seen_at` is rewritten by every heartbeat, so the key mutates
  mid-walk. `routes/devices/core.ts:453` already solved all three on the other
  device route by keying cursor mode on the NOT NULL, immutable `hostname` plus
  explicit NULLS LAST. The mobile routes never got that.
- *An offset walk.* Reachable, but it reorders under the same mutable key: page
  two repeats rows page one returned and drops the ones that moved ahead of the
  offset. A row-count check cannot detect it, because the duplicates make the
  count come out right. It also turns one refresh into up to 62 requests against
  a shared 300/60s per-IP budget, on a screen that refreshes on push, websocket
  and focus.

So the walk is not the fix — **the honest claim is**. One request, plus the
server's exact `total`, so the UI can say it is partial. Unknown total never
reads as complete: only one direction of that flag is dangerous.

I'd rather not paper over the underlying keyset, so **the dead-and-hazardous
cursor mode on these two routes wants its own issue** — fix it the way
`core.ts` did, or delete it. Happy to file it; say which you'd prefer.

## 3. The permanent spinner

The guard is **inside the `setData` updater**, not only around the call. React
does not promise to run that updater synchronously — deferred to the render
phase, a throw escapes the outer frame entirely and the update that clears
`loading`/`refreshing` never commits, which is the exact spinner the catch is
meant to prevent. An outer catch alone would have been fake protection.

Same reasoning removed the `failedCount` read-back: it was assigned inside the
updater and read on the next line, so a deferred updater left it a stale `0`.
Now computed synchronously from `rejectionReasons(results).length`.

The outer catch clears state **before** reporting, since `reportInternalError`
calls into Sentry with no no-throw boundary, and the report itself is guarded so
a throwing reporter cannot convert a handled failure into an unhandled rejection.

## 4. "Unknown organization"

`resolveOrgName` distinguishes "not in a list we loaded" from "we never loaded
the list", and `OrgRollup` carries `nameUnavailable` so the row is not styled as
authoritative. `orgRollups` now lists `data.failed` in its deps — without it the
memo never recomputed, because a failed `orgs` fetch keeps the same `data.orgs`
reference.

## The banner

Renders the hook's computed message instead of one hardcoded string, so
`ALL_FAILED_MESSAGE` and `PARTIAL_FAILED_MESSAGE` stop reading identically. A
separate, non-error notice covers truncation, for alerts and for devices, since
the org counts are computed over those pages.

## Also corrected

Two comments asserting a walk that never happened, including
`deviceListFilters.ts`, which claimed `getDevices` "walks the cursor to the end
of the fleet" before filtering. It never did, so a filter there searches a
subset and "no match" means "not on this page".

## Not fixed here

`DevicesListScreen` still drops `truncated` (it calls the `getDevices` wrapper)
and can show a partial count and a false "No devices match". It is the Devices
screen rather than Systems or Tickets, so I left it rather than widen the diff —
flagging it rather than quietly half-doing it.

## Verification

- `apps/mobile` `tsc --noEmit` clean.
- `apps/mobile` vitest **508 passed / 3 skipped across 42 files**.
- Those two are exactly what the `test-mobile` job runs; there is no eslint
  config under `apps/mobile`.

Two adversarial review passes; the first killed the server-side approach and the
second killed the offset walk, so the shape here is the third design, not the
first.


bdunncompany:
Filed the cursor-mode follow-up as #3770 rather than leave it hanging on a question. It documents all three keyset hazards with the schema references, and points at `routes/devices/core.ts:453` + `routes/devices/cursor.ts` as the pattern that already solves them on the other device route. I left the fix-vs-delete choice open there — the porting option changes response behaviour for every client, so it's yours to call, not mine.

ToddHebebrand (approved):
Merging this. The headline fix is good, but the reason I'm calling it out is the part you weren't asked for.

**You found a server bug while fixing a client one, and you were right to reject the issue's own suggested fix.** I wrote "mirror `getDevices`, which already walks a cursor correctly" in #3753. It doesn't, and I verified your claim on `origin/main` before merging rather than taking it: `routes/mobile.ts` really does declare `let nextCursor: string | null = null` and only ever assign it inside `if (cursor)`. A cold-start caller has no cursor, so its first response carries `nextCursor: null` unconditionally and page two is unreachable. The old walk read that null as clean exhaustion — so it stopped at one page *and* suppressed its own truncation warning. A 216-device tenant saw 100 and looked authoritative.

That's a better find than the four silent failures the issue was filed for, and it invalidated the premise I handed you. Pushing back on it was the right call.

**Rejecting the walk entirely is also right.** Both alternatives you tried and discarded fail for reasons that would have been genuinely hard to see afterwards: the over-fetch route ships silent row-skipping (nullable `last_seen_at` with NULLs first and `encodeCursor(null, …)` returning null terminates the walk; microsecond `triggered_at` against a millisecond `toISOString()` skips rows inside the same millisecond; and `last_seen_at` is rewritten by every heartbeat, so the sort key mutates mid-walk), and offset mode reorders under you in a way a row-count check can't detect, because the duplicates make the count come out right. "One page plus the server's exact total, and say so" is the honest answer, and `truncated` being one-directional — never claiming complete when unknown — is the right default.

Replacing `devicesPaging.test.ts` with fixtures that return `nextCursor: null` like the real route matters more than the test count suggests. The old mock asserted a server contract that never existed, which is precisely why review couldn't see the bug: code, comment and test all agreed with each other and all disagreed with the server.

**One follow-up, not a blocker.** `fetchPage` fires a Sentry `captureMessage` whenever `truncated` is true, and `truncated` is true for any tenant over the page limit — which is the normal state, not an anomaly. `useSystemsData.fetchAll` runs on mount, tab focus, pull-to-refresh, push and WS, so a 216-device tenant generates a warning stream on every app interaction, forever. Legitimately-true warnings at that volume drown the ones that aren't; worth throttling to once per session (or per truncation-state change) rather than per fetch. Happy for that to be its own PR.

The `DEVICE_PAGE_LIMIT` comment rewrite is a good catch on its own — the old one cited `patches/helpers.ts MAX_PAGE_LIMIT`, a different route's constant, which is what made the walk read as 4,000 and deliver 2,000.

Filing the keyset fix separately was the right split. Merging.

PR #3795, opened 2026-08-22
fix(mobile): throttle truncation reports to a state change (#3783)

### LINKED ISSUE #3783: [Mobile] Truncation warnings fire on every fetch for any tenant over 100 rows — throttle to a state change
Follow-up from #3769 (merged as `b2bd606f5`); not a defect in that PR, which is why it didn't block it.

`fetchPage` in `apps/mobile/src/services/api.ts` fires a Sentry `captureMessage` whenever `truncated` is true:

- `device list is showing a partial fleet` (tag `device-list`)
- `alert inbox is showing a partial set` (tag `alert-inbox`)

`truncated` is `total === null || rows.length < total`, so it is true for **any tenant above the 100-row page limit** — which is the ordinary steady state for a real fleet, not an anomaly. `useSystemsData.fetchAll` runs on mount, tab focus, pull-to-refresh, push delivery and WS updates, so a 216-device tenant emits one of these on essentially every interaction, indefinitely.

The warnings are *true*, which is what makes this awkward: they aren't false positives to be deleted, they're correct statements at a volume that destroys their value. At that rate they crowd out the case worth seeing — `total === null`, i.e. a server that reported no count at all, which is the branch #3769's own comment says should not fire in practice.

**Suggested shape:** throttle to once per session, or better, fire only on a *change* in truncation state (and separately/always for the `total === null` branch, which is rare and diagnostic). A `Set` of already-reported keys in module scope is enough; it doesn't need to survive a cold start.

Worth doing before the next mobile release rather than after — this is unreleased code, so the noise hasn't reached Sentry yet.

bdunncompany:
Fixes #3783.

`truncated` is true for any tenant above the 100-row page limit, and `fetchAll` runs on mount, tab focus, pull-to-refresh, push delivery and WS updates — so a 216-device tenant emitted one of these on essentially every interaction, indefinitely.

As you said, the awkward part is that they're *true*. So this doesn't delete them, it reports on **entry** to a non-complete state.

**`unknown-total` still reports every time, as you asked.** I initially throttled it too, on the theory that if it's rare the behaviours are identical and if it isn't rare "always" rebuilds the same flood. I backed that out: a repeated missing count carries information a first report can't — whether it's ongoing, how many tenants or servers it spans, whether it came back after an account change — and it survives the first event being sampled or dropped. Your call was right.

**Three things the naive version got wrong.** All three suppress a report that should fire, which is the same class of bug as the noise:

- **`device-list` ignored `orgId`** even though it's sent to the server, so a scoped fetch and the unscoped fleet shared one state. It broke both ways: an org's first partial was swallowed by the fleet's state, and a small *complete* org reset the key so the unchanged fleet reported again on every return to Systems — rebuilding the flood. Now keyed `device-list:<orgId|all>`, which is what the alert inbox was already doing with `status`.

- **The tracker outlives a session.** It's module scope, so user A's `partial` would suppress the *first* report for user B — possibly on a different server, since the login screen can change it. Now reset alongside `Sentry.setUser(null)` / `analyticsReset()` in `RootNavigator`.

- **Clearing on sign-out isn't sufficient on its own.** A request issued by the old session can resolve *after* the reset and write its state back, re-creating the suppression. Each caller captures an epoch before its request; a result whose epoch has been superseded neither reports nor records. There's a test for exactly that sequence.

Only currently-partial keys are retained (`complete` needs no suppression, `unknown-total` always reports), so browsing many orgs can't grow the map without bound.

`listPaging.test.ts` needed a `resetTruncationTracking()` in `beforeEach` — the tracker's state is module scope, so without it the second case to fetch the same list sees "already reported" and goes quiet. Worth knowing that's now a property of these tests.

**Gate**, matching the `test-mobile` job: `tsc --noEmit` clean, full `vitest run` 44 files / 546 passed / 3 skipped. `pnpm audit` unchanged from main (5 pre-existing: 1 low, 2 moderate, 2 high). No Go, dependency or workflow changes, so those gates aren't affected by this diff. Each new assertion was control-tested by reverting the fix and confirming the test fails.

### DIFF
--- a/apps/mobile/src/navigation/RootNavigator.tsx
+++ b/apps/mobile/src/navigation/RootNavigator.tsx
@@ -14,6 +14,7 @@ import {
 } from '../store/authSlice';
 import { getStoredToken, getStoredUser, clearAuthData, SecureWipeError } from '../services/auth';
 import { getCurrentUser, onDeviceBlocked } from '../services/api';
+import { resetTruncationTracking } from '../services/truncationReporting';
 import { spacing, type } from '../theme';
 import { identify as analyticsIdentify, reset as analyticsReset } from '../lib/analytics';
 import {
@@ -91,6 +92,11 @@ export function RootNavigator() {
     } else {
       Sentry.setUser(null);
       analyticsReset();
+      // Same reason, one layer down: the truncation tracker is module scope and
+      // outlives the session, so a previous account's "already reported" state
+      // would silently swallow the FIRST report for the next one — and the
+      // login screen can point at a different server entirely.
+      resetTruncationTracking();
     }
   }, [user]);
 
--- a/apps/mobile/src/services/api.ts
+++ b/apps/mobile/src/services/api.ts
@@ -4,6 +4,7 @@ import * as Sentry from '@sentry/react-native';
 import { getServerUrl } from './serverConfig';
 import { getOrCreateInstallationId } from './installationId';
 import { fetchWithTimeout } from './fetchWithTimeout';
+import { currentTruncationEpoch, trackTruncation } from './truncationReporting';
 import {
   applyCsrfSignal,
   forgetCsrfToken,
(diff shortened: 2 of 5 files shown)


ToddHebebrand (approved):
**Verified, not just read.** Keys are stable per logical list at every call site (`getDevicesPaged` is only reached bare from `useSystemsData.ts:151` → `device-list:all`, or via `getDevices(orgId)`), the epoch is captured before the `await` at both `api.ts:489` and `:651`, and the reset sits at the one place `Sentry.setUser(null)` / `analyticsReset()` already live (`RootNavigator.tsx:95`) — every sign-out path funnels through `user` going null, and `setServerUrl` only runs from the signed-out `ServerSelectScreen`, so it's covered too. Inbox isn't org-scopable (`getAlertsPaged` sends only `status`), so `alert-inbox:<status>` is the right key. Retention-only-on-partial makes the growth claim hold.

**Nits, none blocking:** `shouldReportTruncation` is only ever called with `'partial' | undefined` as `previous` (state collapses into a `Set`), so the `'complete'`/`'unknown-total'` cases at `truncationReporting.test.ts:52-60` exercise unreachable inputs — fine as documentation. Nothing pins "epoch captured before the request": moving `currentTruncationEpoch()` after the `await` in `api.ts` would pass the suite; a `listPaging` case that resets mid-flight and asserts silence would close that. The `:108` doc says "The Map" — it's a Set.

Merging.

PR #3820, opened 2026-08-22
feat(api,web): activate secret tenant variables in scripts via tenantSecret parameters (#3409 PR4c-2)

ToddHebebrand:
Refs #3409 — PR4c-2, the last PR of the tenant-variables initiative. After this, secret tenant variables are usable from scripts end-to-end.

## What activates

A new script-parameter source **`tenantSecret`** ("From a secret variable"):

```json
{ "name": "api_token", "source": "tenantSecret", "variableKey": "vendor_token" }
```

The resolved value never enters the ordinary `parameters` map. It rides `payload.secretEnv[name]`, is sealed by the PR4a envelope (`secretEnvEnvelope`, AAD-bound to command id + device id), is opened only at delivery, and the PR4b agent exports it as **`BREEZE_VAR_API_TOKEN`**. The secret is never written into the script text.

**Declared, never inferred** (quorum ruling 2026-08-15): `source: 'tenantVariable'` keeps rejecting a secret target; `{{var.<secret>}}` content tokens stay permanently rejected. A `tenantSecret` binding whose target is missing or not actually a secret fails that device closed — no default, no caller override, no fallback.

## Gates (server)

At dispatch, before the `script_executions` insert (no orphan rows):
1. `runAs` — `system`/`elevated` only; `user` or any `targetSessionId` → `secrets_unsupported_run_as` (mirrors the agent's `runAsSupportsSecrets`).
2. Server key — no active secret-encryption key id → `secret_delivery_unavailable` (the envelope refuses to degrade to v1/no-AAD).
3. Agent capability — `devices.script_secret_env_version < 1` → `agent_upgrade_required`.

**At claim time, again** (an offline-queued command can be claimed after a downgrade): `decryptClaimedCommandsForDelivery` runs `failClaimedSecretCommandsForUnsupportedAgent` first and decrypts only the survivors. An unsupported agent never receives the frame; the command and its execution are driven `failed` with "Agent upgrade required…" and the envelope is erased. Two review findings are folded in:
- The heartbeat passes the capability **this beat reported** to the gate (not the stored column) — the device write is guarded on decommissioned/quarantined and can be skipped, so the stored value can be stale.
- The REST command poll (`routes/agents/commands.ts`, self-managed context) runs the gate inside the same system-context closure as the claim — it reads RLS'd `devices`.
- `scriptDispatch`'s immediate WebSocket send also runs the gate after its own claim (it never goes through the batch path); a withheld command reports `deliveryOutcome: 'agent_unsupported'`.

`commandQueue.ts:685/:954` (generic direct pushes) are not gated: no producer enqueues a `script` command with an envelope through them — `scriptDispatch` is the only producer and gates at enqueue.

## Save-time fix (pre-existing gap)

The API only ever checked `content` for secret references; the web warning claimed "the save will be rejected" for a `tenantVariable`→secret parameter but no 400 existed. Create, update and bundle import now reject both mismatch directions (`tenantVariable`→secret, `tenantSecret`→non-secret). Unknown keys still pass (partner-wide scripts resolve per org).

## Also
- Approval digest: `tenantSecret` references are pinned like `tenantVariable` ones (state/variableId/version/isSecret; never the value). No material version bump needed.
- Software deployment templates: a referenced secret now fails the device explicitly instead of being silently omitted.
- Web: source option, secret-mode variable picker (only secret rows selectable), `$BREEZE_VAR_…` hint, run-modal chip, Run-as-user warning; 8 locales; stale "a later release delivers it…" copy refreshed.
- Docs: `features/scripts.mdx` parameter-source table + "Secret variables in scripts" (env name, requirements, redaction limits); `software-inventory.mdx` note.
- The three "Inert until PR4c" comments are retired.

## Invariants checked
- `grep -rn "secretEnv\b" apps/api/src --include='*.ts' | grep -v test` — producers only in `sourcedParameters.ts`/`scriptDispatch.ts`; the only consumer seals via `sensitiveCommandPayload.ts` → `scriptSecretEnvelope.ts`.
- No schema or migration change: `git diff --stat origin/main...HEAD -- apps/api/migrations apps/api/src/db/schema` is empty.
- Secret values never reach `script_executions.parameters` (identity-only `$bindings`), `device_commands.result`, error strings, digest material, logs, or the AI tool's returned text — asserted by unit + integration tests.

## Tests
- shared 2013 · API unit 23,608 (full suite) · web 6,080 (one 5s load-timeout in `terminologyQuality.test.ts` during the concurrent run; green in isolation) · `astro check` 0 errors · API `tsc` 0 errors.
- Contract: `effectDigestCoverage.contract.test.ts`, `terminalPayloadErasure.coverage.test.ts`.
- Integration (real Postgres): new `scriptSecretDelivery.integration.test.ts` — sealed `enc:v3:` envelope stored with no plaintext anywhere in the row, `openSecretEnv` round-trip, identity-only bindings, claim-time downgrade → terminal failure + erasure, preflight refusals leave zero rows. Mutation-verified.
- Every new guard was mutation-verified (gate off → only its tests fail).

## Release note
Secret tenant variables are now usable from scripts via a **"From a secret variable"** parameter; the value is delivered only as `BREEZE_VAR_<NAME>`. Requires an agent with secure secret delivery (the release that ships PR4b — v0.107.0 or later); older agents fail the device with "Agent upgrade required" instead of running without the secret. User-context runs refuse secrets. Saving a script that binds a secret through the plain "From a variable" source is now rejected (previously only failed at run time).

Plan: `docs/superpowers/plans/2026-08-22-pr4c2-secret-activation.md`. Independent plan review (Codex, read-only) found 1 blocker + 4 majors, all addressed above.

### DIFF
--- a/apps/api/src/__tests__/integration/scriptSecretDelivery.integration.test.ts
+++ b/apps/api/src/__tests__/integration/scriptSecretDelivery.integration.test.ts
@@ -1,0 +1,291 @@
+/**
+ * Secret delivery end to end against real Postgres (#3409 PR4c-2).
+ *
+ * The unit suites (`services/scriptSecretDelivery.test.ts`,
+ * `services/scriptDispatch.test.ts`, `services/sourcedParameters.test.ts`)
+ * all mock `../db`, so every one of them asserts the SHAPE of a write that
+ * never happens. The properties this feature actually exists for only mean
+ * something once a row is really stored and read back:
+ *
+ *   1. ENQUEUE — a `tenantSecret` parameter is resolved, sealed, and STORED
+ *      as an AAD-bound `secretEnvEnvelope`; neither the persisted command row
+ *      nor the persisted execution row contains the plaintext anywhere, and
+ *      the envelope opens — with the stored row's own id as AAD — back to the
+ *      exact value the encrypted `tenant_variables` row holds. That round
+ *      trip crosses four independent encodings (tenant-variable AAD, jsonb
+ *      storage, envelope AAD, canonical envelope JSON); a mocked db proves
+ *      none of them agree.
+ *   2. CLAIM — after a real agent downgrade (`script_secret_env_version` back
+ *      to 0), the real claim batch + `decryptClaimedCommandsForDelivery`
+ *      withholds the command AND drives both it and its linked execution row
+ *      terminal, with the payload erased. The `status = 'sent'` guard on that
+ *      terminal UPDATE only has meaning against a row the claim actually
+ *      flipped to `sent`.
+ *   3-4. The two enqueue refusals that must leave NO orphan rows behind
+ *      (an incapable agent, and a user-context run) — "no orphan" is a
+ *      statement about the database, so it is asserted by counting rows.
+ *
+ * Fixture note: the `tenant_variables` row is sealed through the service's own
+ * `encryptTenantVariableValue` (as `tenantVariableResolution.integration.
+ * test.ts` does), because the ciphertext AAD binds the row id — a plaintext
+ * literal would fail to decrypt, drop out of the resolved scope, and turn
+ * every assertion below into a vacuous "unresolved parameter" failure.
+ *
+ * The device snapshot handed to `dispatchScriptToDevice` deliberately carries
+ * NO `agentId`: the immediate-send path would otherwise claim the command
+ * itself, and test 2 needs the row to still be `pending` for the real claim
+ * batch to take.
+ *
+ * Lives under `src/__tests__/integration/`, so the config's
+ * `src/__tests__/integration/**` include glob already covers it.
+ */
+import './setup';
+import { randomUUID } from 'crypto';
+import { beforeEach, describe, expect, it } from 'vitest';
+import { eq } from 'drizzle-orm';
+import type { ScriptParameterDefinition } from '@breeze/shared';
+
+import { withSystemDbAccessContext } from '../../db';
+import { getTestDb } from './setup';
+import { deviceCommands, devices, scriptExecutions, scripts, tenantVariables } from '../../db/schema';
+import { createOrganization, createPartner, createSite } from './db-utils';
+import { encryptTenantVariableValue } from '../../services/tenantVariables';
+import { loadTenantVariableScope } from '../../services/tenantVariableResolution';
+import { dispatchScriptToDevice, type DispatchScriptInput } from '../../services/scriptDispatch';
+import { claimPendingCommandsForDevice } from '../../services/commandDispatch';
+import { decryptClaimedCommandsForDelivery } from '../../services/commandDelivery';
+import { openSecretEnv } from '../../services/scriptSecretEnvelope';
+import { AGENT_UPGRADE_REQUIRED_MESSAGE } from '../../services/scriptSecretDelivery';
+
+const SECRET_KEY = 'vendor_token';
+/** Long enough to clear MIN_SECRET_TENANT_VARIABLE_VALUE_LENGTH and distinctive
+ * enough that a `JSON.stringify(row)` substring search is meaningful. */
+const SECRET_VALUE = 'sk-live-integration-9f3a2b7c4d1e';
+const PARAM_NAME = 'api_token';
+
+const SECRET_PARAMETERS: ScriptParameterDefinition[] = [
+  { name: PARAM_NAME, type: 'string', required: true, source: 'tenantSecret', variableKey: SECRET_KEY },
+];
+
+interface Scenario {
+  orgId: string;
+  deviceId: string;
+  variableId: string;
+  variableVersion: number;
+  script: typeof scripts.$inferSelect;
+  /**
+   * The snapshot dispatch receives. `agentId` is nulled on purpose (see the
+   * file header) — the column itself is NOT NULL, so this is a property of the
+   * caller's snapshot, not of the stored row.
+   */
+  deviceSnapshot: DispatchScriptInput['device'];
+}
+
+async function seedScenario(): Promise<Scenario> {
+  const partner = await createPartner();
+  const org = await createOrganization({ partnerId: partner.id });
+  const site = await createSite({ orgId: org.id });
+
+  const [device] = await getTestDb()
+    .insert(devices)
+    .values({
+      orgId: org.id,
+      siteId: site!.id,
+      agentId: randomUUID(),
+      hostname: `secret-delivery-${randomUUID().slice(0, 8)}`,
+      osType: 'linux',
+      osVersion: '24.04',
+      architecture: 'x86_64',
+      agentVersion: '0.0.0-test',
+      status: 'online',
+      enrolledAt: new Date(),
+      // The PR4b capability floor. Test 2 walks it back down to 0.
+      scriptSecretEnvVersion: 1,
+    })
+    .returning();
+
+  const variableId = randomUUID();
+  const [variable] = await getTestDb()
+    .insert(tenantVariables)
+    .values({
+      id: variableId,
+      orgId: org.id,
+      key: SECRET_KEY,
+      value: encryptTenantVariableValue(variableId, SECRET_VALUE),
+      isSecret: true,
+    })
+    .returning();
+
+  const [script] = await getTestDb()
+    .insert(scripts)
+    .values({
+      orgId: org.id,
+      name: `secret-delivery-${randomUUID().slice(0, 8)}`,
+      osTypes: ['linux'],
+      language: 'bash',
+      content: '#!/bin/bash\ncurl -H "Authorization: Bearer $BREEZE_VAR_API_TOKEN" https://vendor.example.test',
+      parameters: SECRET_PARAMETERS,
+      runAs: 'system',
+      timeoutSeconds: 300,
+    })
+    .returning();
+
+  return {
+    orgId: org.id,
+    deviceId: device!.id,
+    variableId: variable!.id,
+    variableVersion: variable!.version,
+    script: script!,
+    deviceSnapshot: {
+      id: device!.id,
+      orgId: device!.orgId,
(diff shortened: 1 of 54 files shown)


ToddHebebrand:
## Review round: 4 specialist lenses, 7 fixes pushed

Reviewed by four independent lenses (fail-closed/silent-failure, test-quality, security, completeness-sweep), each verifying claims against `origin/main` rather than the branch. Two findings were serious enough that I re-verified them by hand before acting. Everything below is now fixed and pushed.

### Fixed — would have shipped

**1. An org admin could exfiltrate a partner-wide secret** (`6cb421b48`) — the important one.
`tenantVariableReadCondition` deliberately widens partner-wide rows to organization-scope sessions, and `resolveScriptCreateScope` hands any org-scope caller with `scripts:write` a script. Combined with this PR's new delivery channel, an Org Admin of a **customer** org could bind the MSP's partner-wide secret and echo a transformed form of `$env:BREEZE_VAR_*` (both redactors are exact-substring, so base64 defeats them). Before this PR no secret's plaintext could leave the server at all — so this crosses the MSP→customer trust boundary, rather than merely restating the documented "running a script implies using the secrets it binds", which was written for *org-owned* secrets.
Binding a partner-owned secret now requires `canManagePartnerWidePolicies` — the repo's single gate for partner-wide writes, so system scope and full partner admins keep the entire workflow while org scope and limited-org-access partner techs fail closed. Enforced at **all four** write ingresses, including `POST /scripts/import/:id`, which ran neither this check nor the pre-existing content check. Dispatch cannot re-check it (automation/AI dispatch runs under system context with no author auth), which is exactly why every ingress carries it.

**2. Fleet remediation bypassed the entire gate chain** (`910b3d81b`).
A fourth script-dispatch path that never goes through `dispatchScriptToDevice`: it re-fetched only language/content/timeout/runAs and hand-built the command with `parameters: run.parameterSnapshot`. No sourced-parameter resolution, no variable scope, no preflight, no envelope — a `tenantSecret` script would have reached the agent with `BREEZE_VAR_*` **unset and no error**, the precise silent-wrong-run this PR exists to prevent. The snapshot is caller-supplied, so a value could also be injected under a bound parameter's name into the plain map the agent substitutes into script text. Now rejected at run creation and re-checked at dispatch (the script can be edited in between). All bound sources are refused, not just secrets.

**3. A terminal refusal reported as a successful queue** (`f12ffb39d`).
The immediate-send gate drove both rows to `failed`, then dispatch returned `ok:true`. Callers branch on `ok`, so the API answered "queued on N devices" and automation logged `Queued run_script action` / `success:true` — a scheduled credential rotation would silently no-op and stay green.

**4. A missing device row read as "incapable agent"** (`72ae6e926`) — and the consequence was irreversible: terminal write plus payload **erasure**, with no Sentry. An RLS/context regression, a device deleted mid-batch, or replica lag would have permanently failed secret runs fleet-wide while telling operators to upgrade a current agent. Now withheld + reported, row left `sent` for the reaper. Same commit: re-normalize `reportedVersion` inside the gate; enforce (not just document) the single-device contract; log the two paths that stranded an execution row silently; tighten the `runAs` mirror to match the agent's.

**5. Adding a secret parameter permanently disabled Test Run** (`274710d39`) — dead end on the primary authoring surface: the button demanded a default value the schema forbids. Also, the run-as advisory stated half the rule (the server refuses any targeted session too), and the device-detail quick-run picker showed nothing at all.

**6. Unreadable ≠ unset** (`4e919320f`) — a secret whose stored value fails to decrypt reported as "no value set", inviting a duplicate secret during a key rotation. Now a distinct message plus Sentry.

**7. 33+ secret parameters 500'd an entire fan-out** (in `6cb421b48`) — saved cleanly, then threw inside the sealed-payload build. Capped at 32 with a test pinning the constant to `MAX_SECRET_ENV_ENTRIES` so the two cannot drift.

### Cleared with no finding
Cross-tenant leakage (`variableScope` is never caller-supplied — five server-side construction sites, no route body); AAD/envelope replay (binds command id + device id; the re-arm path is hard-restricted to `desktop_stream_stop`); plaintext reaching execution history, audit, Sentry, digest material, or AI output; the complete gate enumeration for every path that can hand a `script` command to an agent; agent self-report spoofing (only ever affects that device's own commands). Test-quality audit found no vacuous Drizzle where-clause assertions and an identifiable killer test for every guard.

### Deferred, deliberately
- **Late-result redaction gap.** If the stale reaper terminalises a long-running secret script, it erases the envelope, and a late result (#3607) is then stored without the *server-side* exact pass. The agent-side redactor still runs, and both are exact-substring — so this is defence-in-depth, not a live leak. The fix touches `terminalPayloadErasureSet`, shared by 11 call sites; not worth that blast radius inside a security PR.
- **Routing fleet remediation through `dispatchScriptToDevice`** — the real fix; the guard above closes the hole meanwhile.
- **`parameterSnapshot` is not validated against declared parameters** on the remediation path (pre-existing).

No disclosure obligation: the feature has never shipped — `tenantSecret` does not exist on any released build, so finding 1 was never reachable in production.

### Verification
API **23,668** passed / 0 failed (full suite) · web **6,091** / 0 · shared **2,017** / 0 · API `tsc` 0 errors · `astro check` 0 errors · docs build clean. Every new guard mutation-verified (broken deliberately, confirmed only its own tests fail, restored). The 3 remaining `packages/shared` tsc errors are in `quotes.test.ts`, byte-identical to `origin/main`.

Two API tests (`mcpServer.orgKeyPartnerRole.test.ts`) failed on an interim run under load average 27 — a 5s timeout whose unconsumed `mockResolvedValueOnce` then leaked into the next test. Bisected against `origin/main` and the pre-fix commit, then re-run 3× on an idle box: green, and green in the full suite above. Load-induced flake, not a regression.

ToddHebebrand:
## Review round 2 — 4 more lenses, 1 blocker + 3 majors fixed

Ran the aspects the first round did not cover (comment-accuracy, type-design, general code-review) plus a re-run of the fail-closed lens against the **fix round itself**, which had never been reviewed. Three of the four serious findings were defects in that fix round.

### Blocker — the partner-secret gate I added was bypassable by the actor it targets

Variable-key uniqueness is **per scope**, so an org admin can create an org-owned secret that *shadows* a partner-wide key. The save-time lookup resolves the org row and passes; deleting that row then lets `resolveForOrg` inherit the partner-wide row at dispatch, where nothing re-checked ownership. Binding a not-yet-existing key is the same hole in time. Found independently by two lenses; I reproduced the mechanism against the code before acting.

Fixed by moving authority to **dispatch** and replacing the capability check with a tier-directional rule (owner-approved):

> **A script may resolve a secret at or below its own ownership tier, never above.**

| script | secret `ownerScope` | |
|---|---|---|
| partner-wide | `partner` | allow |
| partner-wide | `organization` | **allow** — one partner-wide script, each target org's own value resolved per device |
| org-owned | `organization` | allow |
| org-owned | `partner` | **deny** |

New `partnerSecretAboveScriptScope` denial arm, applied to the `tenantSecret` arm only — a partner-wide *non-secret* is already readable through the variables API, so there is nothing to escalate there. Save-time stays as a fast-fail and now derives from the script's own scope. One prior test that allowed a full-partner admin to bind a partner secret into an *org-scoped* script was rewritten to a 400: that script is afterwards editable and runnable by the customer org's own admins, which is the whole problem.

### Major — my "unreadable vs unset" commit was inert

`unreadableVariableKeys` was declared and consumed, but the sole production caller never passed it, so an undecryptable secret still reported *"no value for required parameter"* — the exact wording the change existed to remove. Its tests passed because they called the resolver directly. Now wired at dispatch, with an assertion that pins the wording.

### Major — my suppression set dropped the record of the *common* refusal

`agent_upgrade_required` came from two sites with opposite state: the enqueue preflight (refuses **before** the execution insert — the ordinary path for any pre-v0.106.0 agent) and the claim-time gate (rows already written). Suppressing by code silenced the first. Ten devices with three old agents produced seven execution rows, no record of the other three, and a `devicesFailed` count three short — so the batch could never finalise. Split into `agent_upgrade_required` (enqueue, writes its row) and `agent_upgrade_required_recorded` (claim-time, the sole suppressed code), plus `secret_gate_unavailable` so an infrastructure fault no longer tells an operator to upgrade a current agent.

### Major — the docs stated the wrong agent floor

The requirements table said v0.107.0. The agent side actually shipped in **v0.106.0** (`secretenv.go`, `secretredact.go`, `ScriptSecretEnvVersion: 1` — present at that tag, absent at v0.105.1). That is the one line an operator acts on when a device reports "Agent upgrade required", so a wrong number sends them chasing an upgrade that does not exist. Corrected, along with the tier rule and the org-over-partner precedence that makes the per-org pattern work.

### Cleared

The fail-closed invariant itself holds: every failure branch withholds before any write, both `catch` arms `continue`, and the multi-device throw fires before the loop so no partial terminal state is possible. No log line, error string, or Sentry capture carries a secret value. Registration-list contracts verified N/A (no table, no column). The `tenantSecret` zod arm rated 5/5 on invariant expression — illegal states genuinely unrepresentable.

### Verification

API **23,670** passed. Two files fail (`internal/synthetic.test.ts`, `mcpServer.partnerCreatorPermsNull.test.ts`) — both a `beforeAll`/test **timeout**, both untouched by this PR, and both reproduce identically on clean `origin/main` under the same load (average 19.5, 63 heavy processes from a concurrent session). Environmental, not a regression. Every new guard mutation-verified with its blast radius checked; web and shared untouched this round; docs build clean; `tsc` exit 0 package-wide.

### Deferred (unchanged)

Late-result redaction gap; routing fleet remediation through `dispatchScriptToDevice`; unvalidated `parameterSnapshot`. Plus type-design suggestions worth a follow-up: brand the `secretEnv` value type so `{...parameters, ...secretEnv}` cannot compile, make the `lookup.kind` chain exhaustiveness-checked, and drop the duplicate `ScriptRunAs`.

PR #3831, opened 2026-08-22
fix(api): serialise bundle composition on a globally ordered item lock (#3816)

### LINKED ISSUE #3816: Catalog bundle composition races: setBundleComponents/updateCatalogItem validate without locks (nested bundles possible)
Surfaced by the post-merge codex review of wave 3 (#3811) — pre-existing, not introduced by multi-currency.

`setBundleComponents` (`apps/api/src/services/catalogService.ts` ~:332-376) reads bundle/component flags without locking them; concurrently `updateCatalogItem` (~:476-509) can convert a component into a bundle after observing it is not yet referenced. Result: a nested bundle, or components inserted after the parent was converted to a plain item.

Fix: lock the parent catalog item first (`lockOwnedItemOr404`, now exists), then all component rows in deterministic id order, revalidate flags/ownership under those locks, and replace the component set in the same transaction. Add a two-client race test using the `pg_blocking_pids` holder pattern from `invoiceIssueRace.integration.test.ts`.

Refs #3772.

bdunncompany:
Fixes #3816.

`setBundleComponents` validated bundle and component flags with unlocked reads, so a concurrent `updateCatalogItem` could convert a component into a bundle after it had been observed as a plain item. Nothing in the schema forbids the resulting nested bundle.

Locks the deduplicated, globally sorted union of `{bundleId, ...componentIds}` as `catalog_items` rows, revalidates flags and ownership under those locks, and replaces the set in one transaction.

## Two departures from the fix as scoped, both deliberate

**Not parent-first-then-components — that order deadlocks.** `set(A,[B])` takes A then wants B while `set(B,[A])` takes B then wants A, and the cycle forms *before* the under-lock validation can reject either. Two requests that should both be a clean 400 become a `40P01` instead. Sorting each component list doesn't help, because the already-held parent sits outside that ordering. One global order over the union puts every caller on the same sequence, so the pair serialises and each is refused as `BUNDLE_NESTED`.

**Locking `catalog_items`, not the edge rows.** `updateCatalogItem`'s guard searches for the *absence* of an edge, and `FOR UPDATE` cannot lock a row that doesn't exist yet. It works because the component item must already exist (FK), so both writers contend on that item row. The parent lock alone is therefore sufficient in `updateCatalogItem` — no edge locks needed there.

One `WHERE id = ANY(...) ORDER BY id FOR UPDATE` rather than a per-id loop: the validator permits 200 components, so the loop was up to 201 serial round trips with every earlier lock held open. The `ORDER BY` is load-bearing — Postgres sorts qualifying tuples before the locking node acquires row locks, which is what supplies the common order.

## Correcting something I said on the issue

I claimed the old delete+insert were separate autocommit statements that could permanently empty a bundle. **That's false on the request path** — `db` is already the `withDbAccessContext` transaction there, so they were inside it and rolled back together. The explicit transaction stays because it *is* correct for a caller holding the raw handle, and because it scopes the locks. The comment now says that instead of describing a bug that doesn't exist.

Same misreading caught twice more, and it's a property of the `db` proxy rather than of this function: an inner `db.transaction` under `withDbAccessContext` is a **savepoint**, so a `.then()` after it still runs before the real commit. Event emission is unchanged from before and no longer claims otherwise. Worth knowing the row locks are held across that enqueue.

## The test is a regression guard, not a reproduction

Stated plainly in the file header too. Measured against the pre-fix code:

- Single race asserting "no nested bundle afterwards" — **passed unfixed**.
- Holder client pre-locking the component row, asserting both backends block — **also passed unfixed**. The INSERT's own FK check locks that row, so the unfixed path blocks too, just later; "both blocked" says nothing about whether validation happened under the lock.
- Repeated unsynchronised trials — failed once early, then passed 3/3 at 12 trials and 3/3 at 60. The scheduler here consistently favours one ordering.

So an earlier "the control fails correctly" claim of mine was a lucky hit I generalised from. Deterministic reproduction needs barriers pinning *validate-unlocked → conversion commits → insert*, which means a test-only seam in the service. I don't think that belongs in a lock fix, but say the word and I'll add it.

## Gate

Full `apps/api` unit suite **23,520 passed / 0 failures**; full integration suite **329/329 files, 2080 passed / 0 failures**; `tsc --noEmit` clean; eslint clean. Two adversarial review rounds on the diff — the first confirmed no cross-writer deadlock (every other catalog writer takes exactly one item lock) and that the missing `partnerId` predicate is RLS-covered, where adding one would have broken the established `BUNDLE_CROSS_PARTNER` assertion.


bdunncompany:
Heads-up before this surprises either of us: **#3832 edits `setBundleComponents` too**, so these two will conflict textually. Flagging it now rather than after a merge.

They're complementary, not competing. #3832 adds the `allocationCurrency` parameter, the `ALLOCATION_CURRENCY_REQUIRED` guard, and stamps `allocation_currency` on the inserted rows. This PR leaves all of that logic alone and changes the surrounding shape: the lock acquisition, the transaction boundary, and re-reading the flags from the locked set instead of the unlocked `metaRows` select at the top.

The overlap is concentrated in two spots — the function signature and preamble (your new param and guard sit where my `lockItemsInGlobalOrder` call and locked-parent re-read go), and the insert values object (your `allocationCurrency` line lands inside my `tx.insert`).

**#3832 should go first.** It's part of the wave series and has a lot more surface; I'll rebase this onto it and resolve, and the resolution is mechanical — your validation moves to after the locks are taken, so it revalidates against locked state rather than an unlocked read, which is strictly better for it too.

One thing worth your eye when I rebase: your guard reads `bundle.isBundle` from `getOwnedItemOr404`, an unlocked select. Under this PR that becomes the locked re-read, so `ALLOCATION_CURRENCY_REQUIRED` and `NOT_A_BUNDLE` both start evaluating under the row lock. No behaviour change for valid input; it just closes the same window for your check that this PR closes for the flags.

Say the word if you'd rather I rebase now against `#3832`'s head instead of waiting for it to land.

bdunncompany:
Rebased onto `835c11dc` (after #3832 and #3833 landed). The overlap in `setBundleComponents` resolved mechanically, but review of the rebase turned up three defects I'd introduced and one pre-existing one, so this is worth a re-read rather than a diff-skim.

### Resolving against #3832

Your `allocationCurrency` param, the `ALLOCATION_CURRENCY_REQUIRED` backstop and the `allocation_currency` insert column are all preserved. I moved the backstop **before** the transaction: it reads nothing from the database, so under the locks it meant locking every named `catalog_items` row only to reject on shape. Nothing observable changes over HTTP or through the AI tool, since both parse `setBundleComponentsSchema` first and its refine already rejects an allocation with no currency. For a direct service caller the precedence does change — `NOT_A_BUNDLE`, `ITEM_NOT_FOUND` and a raw invalid-uuid now all answer `ALLOCATION_CURRENCY_REQUIRED` first. Say the word if you'd rather keep your original ordering.

### Three things I got wrong, now fixed

**Locking foreign rows.** The first version locked every RLS-visible row in the request. Pre-fix, component metadata was a plain unlocked `SELECT` — only the parent went through `getOwnedItemOr404`. So foreign component rows went from unlocked to `FOR UPDATE`, letting an invalid request block another tenant's writer. Now query 1 locks only rows matching `partner_id` (ordered, `FOR NO KEY UPDATE`) and query 2 reads the rest unlocked, which is what keeps `BUNDLE_CROSS_PARTNER` reachable. Only the first query takes locks, so the global ordering still holds.

**Lock strength.** `FOR NO KEY UPDATE` is the weakest mode that still conflicts with itself and with `updateCatalogItem` (which pre-locks `FOR UPDATE`, and whose plain UPDATE would take `NO KEY UPDATE` anyway). An FK's `KEY SHARE` stays compatible, so inserting an edge that merely *references* one of these items is no longer blocked.

**UUID spelling.** This one is the reason I'd ask for a careful look. Moving the parent lookup out of SQL and into a `Map` made it case-sensitive where `eq` never was, so an upper-case parent 404'd a row this transaction had *already locked*. `manage_catalog` passes `String(input.catalogId)` with no GUID parse (its `setBundleComponentsSchema.parse` covers only `components`/`allocationCurrency`), so braced and hyphen-less ids reach the service too. Ids are now canonicalised once, before dedup/locking/validation/DML. The normaliser matches Postgres's grammar exactly and deliberately does **not** accept more than the server does — an over-permissive version would turn a string Postgres refuses into a valid id pointing at a real row.

### On the tests

The two race tests are **regression guards, not reproductions** — the header says so and I'd rather that stayed visible. Neither reliably fails against the unfixed service; forcing the interleaving needs a test-only seam in the service, which I left out of scope. I did tighten the conversion race from "some writer was refused" to *exactly one fulfilled, exactly one rejected, code is `BUNDLE_NESTED`*, since two unrelated rejections previously passed.

The UUID test is a real control: it fails against the unfixed helper. It asks Postgres itself via `pg_input_is_valid` rather than a `try/catch` around a cast (a catch-all scores a connection drop as "invalid uuid"), and it enumerates all 2^7 hyphen masks braced/unbraced/upper/lower — 512 accepted spellings — plus a negative set, asserting the helper agrees with the parser in both directions.

### Two I did not fix, deliberately

1. **`emitCatalogEvent` is awaited before the outer commit, with the row locks held.** The ordering predates this PR; what I add is the item locks across it. Dropping the `await` would fix the lock duration but changes delivery semantics for every caller, and a catalog outbox is a bigger change than a lock fix. Your call.
2. **Pre-existing, unrelated to this PR:** `requireOrgAccess` does `accessibleOrgIds.includes(orgId)`, a case-sensitive comparison against Postgres-rendered ids, so an upper-case `orgId` can yield a spurious `ORG_DENIED` in `setOrgPriceOverride`, `removeOrgPriceOverride`, `computeBundleEconomics` and `resolvePrice`. Happy to open a separate issue.

### Verification

Local, reproducing CI's shape: `tsc` clean; **24,038 unit tests passed, 0 failed**; integration **2,161 passed across 4 shards, each on a fresh database** (a single-database run of all 339 files exhausts the 5.9G tmpfs and fails ~18 files inside `cleanupDatabase` with `53100`, which is an artefact of not sharding, not a real failure). osv-scanner: 0 critical. Supply-chain guard: pass. Lockfile is byte-identical to main.

bdunncompany:
Rebased onto `8066aedb` now that #3874 has landed. **The rebase produced no conflict, and that turned out to be the problem** — worth flagging because the same trap is waiting for anyone else rebasing across wave 6.

### The silent merge

Wave 6 added the representability guard inside `setBundleComponents`:

```ts
if (allocCurrency) {
  for (const c of components) { ... assertRepresentable(...) }
}
```

This PR snapshots the caller's array once (`const items = components.map(...)`) because `components` is caller-owned, mutable, and read across awaits. Your loop and my snapshot don't textually overlap, so git merged both cleanly and produced a function where **one read used the live array and every other read used the snapshot**.

That isn't a stale read, it's an inversion: the guard would validate whatever the caller's array holds at that instant while the insert writes the snapshot, so an unrepresentable amount could be persisted having never been checked. Changed to read `items`.

### And then I got the position wrong

My first resolution hoisted your guard out of the transaction, next to the `ALLOCATION_CURRENCY_REQUIRED` backstop — it touches no database state, so it looked like free efficiency. It isn't: you shipped it *after* the parent lookup, so `ITEM_NOT_FOUND` and `NOT_A_BUNDLE` win over `PRICE_NOT_REPRESENTABLE`. Hoisting inverted that, and it's reachable over plain HTTP, since `money` accepts `100.50` and doesn't enforce per-currency minor units — a request naming a missing parent **and** a fractional-yen allocation carries both faults.

The guard now sits immediately after the locked parent checks. It doesn't need the locks; it needs that position.

The two problems interact in a way I'd have missed if I'd only fixed one: at the hoisted position there is no `await` between the snapshot and the loop, so reading `items` was merely defensive. After the lock query there *is* one — so `items` becomes load-bearing exactly where the guard belongs.

### Coverage

Your wave-6 test pairs a bad allocation with a **good** parent, so it passes in either position and can't see this. Added a precedence case: missing parent + fractional JPY must give `ITEM_NOT_FOUND`, non-bundle parent + fractional JPY must give `NOT_A_BUNDLE`, and a good parent + fractional JPY must still give `PRICE_NOT_REPRESENTABLE` so the first two are about ordering rather than a dead guard. Control run: with the guard hoisted it fails with `PRICE_NOT_REPRESENTABLE` where `ITEM_NOT_FOUND` is expected.

### Verification

`tsc` clean. **24,128 unit passed, 0 failed.** Integration **2,261 passed across 4 shards, each on a fresh database** — including your `setBundleComponents: a fractional allocation in a zero-decimal currency is refused` unchanged.

Nothing of wave 6's dropped or reordered inside this method other than the position fix described above.

ToddHebebrand (changes requested):
**The lock design is sound — verified, not just read.** One `WHERE id = ANY … ORDER BY id FOR NO KEY UPDATE` gives a genuinely global order (ids canonicalised before dedup/sort, so JS and uuid collation agree), the locks live inside the same transaction as the delete/insert, and I checked the other catalog writers: `updateCatalogItem` takes `lockOwnedItemOr404` FOR UPDATE *before* its absence-of-edge check and every other writer takes exactly one item lock, so there's no cross-writer cycle and both races #3816 names are closed, not a subset. The second-snapshot exclusion in `lockOwnedItemsInGlobalOrder` and the locked-set re-read of `is_bundle` are exactly right. Integration test reads real driver shapes throughout.

**One must-fix: the locking query has no `lock_timeout` bound.** `lockOwnedItemsInGlobalOrder` (catalogService.ts:685) can wait indefinitely for up to 201 row locks on a request path, holding a pooled connection the whole time — the recurring prod failure class here, and the same reason your #3760 round grew a bound. You already built the canonical pattern for this an hour ago: the one-statement tighten-and-read CTE in `services/deviceDeletion.ts` (merged `eb6dfa9be`) — `pg_settings` prior value + `set_config(…, true)` in one round trip, never widening a stricter caller, restore only on the success path, 55P03 mapped to a retryable 409 at the route. Reuse it here (extracting it into a small shared helper next to `db/rowCount.ts` is fine if you'd rather not copy it — your call).

**Optional, your judgment:** the comment notes `emitCatalogEvent` is awaited while the row locks are still held. With the bound above in place the exposure is capped, so I'm fine either way — move the enqueue after commit if it's cheap, or leave the comment as the documented trade.

Everything else is merge-ready; this is a one-round bounce.

bdunncompany:
Must-fix done — but it took three rounds to actually *be* a bound, and the first two versions would have looked right.

### The bound

Reused your `deviceDeletion` pattern by **extracting** it to `db/lockTimeout.ts` rather than copying, since "never widen a stricter caller, restore only on success" is exactly the invariant that rots when duplicated. Both call sites use it now.

Then adversarial review caught that **`lock_timeout` is not a statement bound at all** — it applies per lock *acquisition*. Measured on PG16 rather than argued: two rows, staggered holders, `lock_timeout='1000ms'` → the statement ran **1214ms and succeeded**. Scaled to 201 rows that is a connection held for minutes. My first "3 second bound" bounded nothing.

So the locking query now takes **both**, deliberately unequal:

- `lock_timeout` **3000ms** — the contention signal, and the only one mapped.
- `statement_timeout` **4000ms** — hard cap, unmapped.

They must not be equal: at the same value the statement deadline always fires first, `55P03` becomes unreachable, and the `ITEM_BUSY` mapping you asked for is dead code — ordinary single-row contention would surface as an uninformative cancellation. 4000 rather than something looser because the pool is 30 connections and `db/index.ts` already warns at a 2s held context; anything above 3000 keeps `55P03` reachable and 1s is ample for an indexed ≤201-key lookup. **Worth tuning from measured production waits — I picked it from those two constraints, not from data.**

### The mapping, and a gap it closed

Only **`55P03`** maps to `ITEM_BUSY`. `57014` is generic `query_canceled` with no evidence of its source — equally an admin cancel, a client disconnect, or a slow uncontended write under a stricter caller's own deadline — and the wrapper spans the whole transaction, so claiming it as contention would tell an operator to retry something that was never contended.

I also moved the mapping from the route to the **service boundary**. Mapping raw `55P03` in `handleServiceError` was too broad (it also serves `GET /:id/economics`) and missed a caller entirely: `aiToolsCatalog` converts `CatalogServiceError` only, so the AI path would still have surfaced a raw driver error.

### Where I got the tests wrong

Recording this because two of my own tests passed against broken code:

1. A mechanism test that installs the settings itself proves the Postgres semantics but **not the wiring** — deleting `tightenStatementTimeout` from production left it green. So did the single-holder contention test, since one indefinitely-held row is covered by `lock_timeout` alone.
2. The replacement asserted the settings were *restored afterwards*, which passes whether or not they were ever applied.
3. Matching on the GUC name alone found the **restore's** bare `set_config` rather than the tighten — control still passed with production code deleted.

It now keys on the tighten's `pg_settings` CTE. Control: deleting the production call fails with `statement_timeout was never tightened`.

### Also

The prior-value decoder accepted only postgres-js arrays, so a node-postgres `{rows:[...]}` — which carries the answer, and which the `deviceDeletion` suite already exercises — read as unrecoverable. Since the catalog path now throws on an unreadable prior value, that would have turned a parseable result into a 500. Handles both now.

That change then broke a `deviceDeletion` test whose fixture used `{rows}` *specifically to simulate an unreadable shape*; its premise was false, so it now uses a shape that genuinely is.

On your optional item: `emitCatalogEvent` is still awaited with the locks held. With the cap in place the exposure is bounded, so I left it as the documented trade — say the word if you'd rather it moved.

### Verification

`tsc` clean. **24,567 unit passed, 0 failed.** Integration **2,309 across 4 shards, fresh DB each**. The timing-sensitive test run 3× for flake — clean each time.

ToddHebebrand:
**Re-review — the bound is right, the extraction is faithful, and I verified the load-bearing claims independently. One CI blocker stands between this and merge.**

**Verified on a live PG16, not taken on trust:** all four grammar claims behind `canonicalUuid` reproduce exactly — leading/trailing whitespace rejected, a hyphen after any group of four accepted and canonicalised (`550e-8400-…`), braces + upper case + no hyphens accepted, a single unbalanced brace rejected. The regex is the server's grammar, not a looser one. And the exposure is real: `aiToolsCatalog.ts:528-534` parses only `components`/`allocationCurrency` through `setBundleComponentsSchema` and passes `String(input.catalogId)` with no GUID parse, so a non-canonical spelling reaches the Map exactly as you described.

**The extraction:** `deviceDeletion` keeps its 3000ms bound and the same never-widen CASE, and `readPriorMs` learning the `{rows}` shape is a strict improvement over the array-only decoder — turning a parseable node-pg result into "unrecoverable" was the worse behaviour, and the fixture update correctly moves the "unreadable" simulation to a shape that actually is.

**The dual timeout:** the unequal-bounds argument holds — equal values make 55P03 dead code and ordinary contention would surface as an uninformative 57014. The asymmetry on an unreadable prior is also right: an unrestorable `lock_timeout` left in force is merely stricter, but an unrestorable `statement_timeout` would govern the caller's remaining DML, so failing closed there is correct. Mapping only 55P03 through `pgErrorCode` (`.cause`-aware — the #3760/#3881 lesson applied before it bit) and leaving 57014 unmapped is exactly the discipline I wanted.

**Blocking — CI is red on your own new test:**

```
apps/api/src/__tests__/integration/catalogBundleCompositionRace.integration.test.ts(418,19): error TS2571: Object is of type 'unknown'.
```

Type Check fails, so Build/Integration/Smoke never ran. Fix that one line; on a green run this merges as-is.

bdunncompany:
**Correction to my previous comment.** It said the 3000/4000 split made the `55P03` mapping work. It doesn't, and I'd already pushed and commented before an adversarial review round caught it. What follows is what actually changed, and two things I'm deliberately leaving to you.

### The mapping was still broken, and my own comment described the bug as the fix

`lock_timeout` fires per **acquisition**. A statement locking N rows behind blockers that release at staggered times restarts the per-lock timer each time, so the **statement** deadline can land first even though every millisecond was spent waiting on locks. Measured on PG16 16.14 rather than argued — two rows, holders released 700ms apart, `lock_timeout=1000ms`/`statement_timeout=1400ms`:

```
cancelled at 1412ms with 57014 — 55P03 never raised
```

So mapping only `55P03` left ordinary multi-row contention surfacing as a 500. The 4000/3000 gap I described protects only the **first** blocked acquisition.

### What it does now

- **`55P03` → `ITEM_BUSY` (409).** Unambiguous: a lock this statement needed was held.
- **`57014` → `ITEM_LOCK_NOT_ACQUIRED` (409).** New code, and deliberately cause-neutral.

My first repair was wrong too, in a way worth recording: it decided whether a `57014` "was really contention" by checking whether this code had installed the deadline. That's unsound both directions — false positive on `pg_cancel_backend`, a slow plan, an I/O stall or expensive RLS (all reported as "another operation is modifying it" with nothing held), and false negative on a ROLE- or DATABASE-level `statement_timeout`, since `pg_settings` reports the effective value and cannot say who set it. It's gone. The catch wraps exactly the locking SELECT, which establishes *which* statement failed but never *why*, and nothing now claims otherwise. I renamed it away from `ITEM_LOCK_TIMEOUT` for the same reason: an admin cancel can arrive immediately, so "timeout" asserted a cause I can't prove.

### Two decisions I'm not making for you

**1. Status.** Strictly, 409 is wrong here — RFC 9110 scopes it to a conflict with the resource's current state, which a generic cancellation doesn't establish, so some 5xx is the stricter answer. I kept 409 anyway because status class drives alert severity: `HighErrorRate` is **critical at 5xx > 5% for 5m` against `High4xxRate` **warning at > 20% for 10m**, and since the route serialises the error rather than throwing, it never reaches Sentry — so that alert would fire with no exception behind it. Reclassifying ordinary contention into 5xx isn't a call I should make quietly on your repo. Happy to flip it to 503 plus a dedicated rule for this code if you'd rather.

I should flag that I can't verify the operational half from the repo: tracked `monitoring/alertmanager.yml` has its Slack, PagerDuty and email receivers commented out and routes critical alerts to `http://api:3001/webhooks/alerts`, which has no matching route under `apps/api/src`. If that's stale relative to production, my reasoning holds; if it's live, alert delivery is its own problem.

**2. Client copy.** `bundleFriendly` in `CatalogItemEditorDrawer.tsx` handles neither the new code nor the pre-existing `ITEM_BUSY`, and `CatalogServiceErrorCode` has no exhaustiveness check, so nothing fails when a member is added. Not user-visible today — `runAction` falls back to the server message — so I left it rather than widen the diff.

### On the tests

Sixteen mutations are each now verified to fail. Recording the ones that survived a version of this PR I'd have called done, since most are the "test passes against broken code" class:

| | mutation | why it survived |
|---|---|---|
| F | `.for('no key update')` → `.for('share')` | still conflicts with the tests' `FOR UPDATE` holders, so contention tests stayed green — but `FOR SHARE` doesn't self-conflict, so two concurrent replacements could both proceed and leave the **union** of two requests |
| J | drop the `priorMs === 0` arm | every restore test used non-zero values; `0` is the Postgres default, so this is the path production takes. Both restores silently stop running |
| L | tighten with `set_config(..., false)` | the transaction-local restore masks it for every in-transaction read; on commit the pooled backend keeps 3s/4s |
| M | delete `ORDER BY id` | green whenever Postgres happens to scan the seeded ids in the same physical order |
| N | filter by `partnerId` only | nothing recorded `.where`; locks the partner's **entire catalog** per bundle edit |
| O | `code === '57014'` → `if (code)` | nothing made the locking query fail for a non-lock reason, so 42501/40P01/08xxx became a friendly 409 with no Sentry event |
| P | drop `eq(partnerId)` from the lock | nothing named an item the partner doesn't own, so an invalid cross-partner request would take a row lock on the OTHER partner's item before rejecting — it returns `ITEM_BUSY` at 3.2s instead of `BUNDLE_CROSS_PARTNER` |

Two of my own tests also passed against broken code before I controlled them: one asserted the settings were *restored* (true whether or not they were ever applied), and one read the session leak back in a later transaction — which can't see it, because the restore writes the caller's original value, so at the default `0` a session-scoped write of `0` is indistinguishable from none, and it depended on the pool returning the same connection. That one is now an assertion on the emitted SQL.

The staggered test gates on `pg_blocking_pids` naming both waiter and blocker, zeroes the outer GUCs so a role default can't manufacture a first-row `57014`, and requires a **second** proven block before accepting the result — otherwise any `57014`, including one raised while still waiting on the first row, satisfied it.

### Verification

`tsc` clean. **24,567 unit passed, 0 failed.** Integration **2,315 across 4 shards, fresh DB per shard.** The timing-sensitive file run 3× for flake, clean each time. Every mutation above was applied, run, and reverted individually.

ToddHebebrand (approved):
**Approved — merging.** Re-reviewed 30cc3f344 as a delta (independent review round, all six of my focus areas verified against the code, not the comment):

- **57014 catch scope confirmed exact**: the `try` at `catalogService.ts:819-857` contains only the locking SELECT chain; the guard is `code === '55P03'` / `code === '57014'` with a bare `throw` fallthrough, and `pgErrorCode` walks `.cause`, so your mutation-O class is closed for real.
- **Restore semantics confirmed**: `priorMs === 0` handled on both the tighten CASE and `lockTimeoutWasChanged`; all four `set_config` calls are `is_local = true`, so nothing can leak to the pool.
- **partnerId scoping + `ORDER BY id` both survive and are now test-pinned.**
- `readPriorMs` checks `Array.isArray` first, so a postgres-js Result can never fall into the `{rows}` branch.

**Ruling 1 — status stays 409.** Your alerting argument is the operationally correct one: reclassifying ordinary contention into 5xx would make `HighErrorRate` page on traffic with no exception behind it. RFC purity loses to that. The cause-neutral rename (`ITEM_LOCK_NOT_ACQUIRED`) is the right call for the same reason you made it — we can't prove why the statement died, so the code shouldn't claim to. (And yes — `monitoring/alertmanager.yml` in the repo is stale relative to production; your reasoning holds.)

**Ruling 2 — client copy stays as-is.** `runAction`'s server-message fallback covers it; not worth widening this diff. If `CatalogServiceErrorCode` grows again we should add an exhaustiveness check then.

Two optional test-hardening follow-ups, non-blocking: `catalogBundleCompositionRace.integration.test.ts:509` assumes the CI role's prior GUCs are looser than 3s/4s — neutralise them first as the staggered test does; and the `:512` `toContain('true')` on the stringified SQL would be tighter as `not.toContain(', false')`.

The mutation table is the most convincing verification section I've had on this repo — F (FOR SHARE union-of-two-requests) and P (cross-partner lock-before-reject) especially. Thanks for correcting your own comment before I had to.

Issue #3796, opened 2026-08-22
Quote revisions — revise and resend a sent quote


ToddHebebrand:
Tracking issue for **Quote revisions — revise and resend a sent quote** (`quote-revisions`).

- Plan: `docs/superpowers/plans/2026-08-17-quote-revisions.md`
- Spec: `docs/superpowers/specs/billing/2026-08-17-quote-revisions-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All six waves merged to main on 2026-08-23 (W03 #3873, W04 #3875, W05 #3878, W06 #3879). Live acceptance run passed against a wt-stack stack: revision send retires the parent atomically, the old public link dies at every token route (410 QUOTE_SUPERSEDED, asset routes fail closed 404), accept on the new link invoices the revised price. Follow-ups tracked separately: #3905 (sendQuote holds the request transaction across PDF render + mail), #3906 (portal islands don't hydrate under wt-stack), #3880 (cascadeDeleteOrg e2e coverage). Unreleased — ships with the next tag.

ToddHebebrand:
**Live acceptance run (plan Task 11) — done, passing.** Run against a `wt-stack` stack on the W06 branch (W01–W05 on `main`), driving both the API and the real customer-facing pages in a browser.

**send → revise → change price → send revision**, quote `Q-2026-0006` @ $1,200.00 → `Q-2026-0006-R2` @ $1,740.00:

| Check | Result |
|---|---|
| Parent retired on send | `status=superseded`, `public_link_revoked_at` stamped in the same transaction |
| Lineage, both directions | parent detail links to `Q-2026-0006-R2`; revision banner names `Q-2026-0006` |
| Send dialog warning | "Sending this replaces Q-2026-0006 and disables the link the customer already has" |
| Old link — `GET` | **410**, branded sentence, no raw enum, no successor link |
| Old link — `POST accept` / `POST decline` | **410 `QUOTE_SUPERSEDED`** |
| Old link — assets | fail closed **404** (a retired link is deliberately indistinguishable from a bad id) |
| New link | live, renders the revised $145.00 unit / $1,740.00 total, accept form present |
| Accept on the new link | `converted`; **INV-2026-0002 total $1,740.00** — matches the revised price |
| Superseded parent | no invoice |

[...]

Issue #3801, opened 2026-08-22
W05: Web + portal UI — Revise action, revision/superseded banners, lineage links, locale parity


ToddHebebrand:
Wave W05 of #3796 — Web + portal UI — Revise action, revision/superseded banners, lineage links, locale parity
Branch: `feature/3796-quote-revisions/wave-3801`

ToddHebebrand:
Input for W05 from the five-lens review of #3873, filed so it is not lost.

W03 makes `status='superseded'` **reachable for the first time** — there was no writer of that status on `main` before this wave. That activates two UI gaps that were previously dead code paths:

**1. The customer portal renders the raw enum.** `apps/portal/src/components/portal/QuoteDetailView.tsx` and `QuoteList.tsx` have no `superseded` entry in their status label maps, so the `?? status` fallback prints the literal string **"superseded"** to the customer. `apps/web` does have localized labels; the portal is the gap. The portal also serves the complete quote body with the withdrawn pricing, where the public token route now refuses with a 410. Actions *are* correctly gated (`open = status === 'sent' || 'viewed'`), so there is no dead Accept button — that is the only reason this is not worse.

**2. A scheduled send refused because the parent was accepted reports as "email failed".** `sendQuote` throws the carefully-worded `PARENT_CONVERTED` ("The original quote was accepted while this revision was being drafted — it can no longer be sent"), but on the delayed path that message dies in Sentry: `quoteSendQueue.ts` stamps `sendEmailReason: 'schedule_failed'`, and `QuoteActions.tsx` maps `schedule_failed` → the generic email-transport warning string.

Sequence: tech schedules the revision, the customer accepts the original inside the undo window, the job fires and correctly refuses. The tech sees "we couldn't email this", presses Send again, gets a 409 they cannot reconcile with the banner, and never learns the original was accepted — while an invoice for the *old* pricing already exists. Needs a distinct `parent_settled` reason in the `SendQuoteEmailReason` union (a type change, not a migration) plus a real string.

**3. Also worth surfacing:** the send response already returns `superseded`, and `apps/web` has zero references to it. When a send both retires a parent and fails to email (`superseded && !emailed`), the banner should say the original was retired and its link revoked, and steer to *Copy share link* — otherwise the customer is left with a dead original and no replacement mail.

Issue #3800, opened 2026-08-22
W04: Public + portal API — 410 QUOTE_SUPERSEDED view, asset closure, portal successor pointer


ToddHebebrand:
Wave W04 of #3796 — Public + portal API — 410 QUOTE_SUPERSEDED view, asset closure, portal successor pointer
Branch: `feature/3796-quote-revisions/wave-3800`

ToddHebebrand:
Concrete input for the "asset closure" half of this wave, found during the W03 review (#3873) and deliberately deferred here rather than widened into that wave.

W03 makes a revision send retire its parent to `superseded` and stamp `publicLinkRevokedAt`. The main public view and the accept path both honour that and return 410. **The public asset routes do not.** They verify the signed token and that it resolves to the quote, but never inspect `status` or `publicLinkRevokedAt`:

- `/quotes/public/:token/images/:imageId`
- `/quotes/public/:token/line-image/:lineId`
- `/quotes/public/:token/contract-file/:blockId`

So after a parent is superseded, whoever already held the old link can still fetch that quote's assets — including uploaded contract PDFs — with a 200. Not a cross-tenant leak (the holder was legitimately sent that link), which is why it was not treated as a W03 blocker, but it does mean link revocation is not yet actually complete.

Suggested shape, so the next route added doesn't reintroduce it: put the active-link check in the shared public-token resolver rather than in each handler, rejecting `status='superseded'` or a non-null `publicLinkRevokedAt` for every token-authenticated route, and have the asset routes inherit it. Worth a test that enumerates the token routes and asserts each one 410s on a revoked link, so the check cannot be forgotten on a future route.

ToddHebebrand:
Second batch for W04, from the five-lens review of #3873. Same framing as the asset-closure note above: verified, deliberately not fixed in W03.

**Decline is less honest than accept for a superseded quote.** W03 gave `acceptQuote` a purpose-built `410 QUOTE_SUPERSEDED` ("This quote has been replaced by a newer version"). Neither decline path got the same treatment:

- `declineQuoteByActor` throws `409` with the raw internal status — `Cannot decline a quote in status superseded` — and that string reaches the customer portal verbatim via `routes/portal/quotes.ts`.
- The public decline route (`quotesPublic.ts`) falls into `'bad_state'` → a generic 409, *"This quote can no longer be declined"*.

[...]

PR #3873, opened 2026-08-23
feat(quotes): W03 — supersede-at-send + concurrency hardening

### LINKED ISSUE #3799: W03: Supersede-at-send + concurrency hardening — atomic parent retire, accept 410, row locks/CAS
Wave W03 of #3796 — Supersede-at-send + concurrency hardening — atomic parent retire, accept 410, row locks/CAS
Branch: `feature/3796-quote-revisions/wave-3799`

ToddHebebrand:
Closes #3799

Wave 3 of the quote-revisions feature (#3796). Makes sending a revision atomically retire its parent, and hardens every lifecycle write on the quote surface against the races that opens up.

## What changed

**Supersede-at-send.** When a revision draft is sent, its parent is flipped to `superseded` and its public link revoked (`publicLinkRevokedAt`) in the *same* transaction as the child's `draft→sent` claim. The parent is locked `FOR UPDATE` before validation, in the same lock order `acceptQuote` uses, so a concurrent accept and a revision send serialize rather than deadlock.

An `accepted` or `converted` parent is never superseded — that is a settled outcome with an invoice or contract behind it, so the send is refused with `PARENT_CONVERTED` (409) instead. `SUPERSEDABLE` is `sent | viewed | declined | expired`.

**Concurrency hardening across five lifecycle sites.** Each unlocked read followed by a write is now a real DB-level guard — a predicate in the `WHERE` clause, so a stale snapshot cannot stomp a concurrently-settled row — and every guarded write inspects its row count.

There is a deliberate asymmetry in how a zero-row result is treated:

- `markQuoteViewed` treats it as **success**. It is a cosmetic stamp, and losing to a concurrent settle is a legitimate outcome.
- `declineQuoteByActor` treats it as a **409**. Declining is a real customer-visible outcome that must never silently pretend to have happened.

**Draft edits serialize against send.** `loadDraft` takes `FOR UPDATE`, and `sendQuote` locks the child before reading its content — so an edit can no longer land between the content read and the send, which would have emailed a PDF that no longer matched the stored quote.

## Review

Codex ran a read-only adversarial pass over the diff at high effort, targeted at race correctness, transaction boundaries, error semantics, and test vacuity.

It **cleared the core design**: accept-vs-send is correctly serialized by the parent lock; duplicate child sends serialize on the child lock; revise-vs-accept and revise-vs-revise are covered by the parent lock plus the `quotes_revision_of_uq` unique index; and parent retire + child send share one Postgres transaction on the route, worker, bulk and AI call paths.

It raised five findings. Three were verified and fixed in `784205b58`:

1. The new `quote.superseded` audit was emitted *inside* the worker's transaction, so a later failure could roll back the send while leaving a success audit. Now emitted post-commit, success path only. The `try/catch` around it was also dead code — `writeAuditEvent` returns `void` and the underlying `createAuditLogAsync` never throws back to callers.
2. `markQuoteViewed` emitted `quote.viewed` off the stale pre-update snapshot even when the guarded write matched zero rows — a real business event for a quote that was never stamped viewed. Now gated on an actual match; caller semantics unchanged.
3. The parent-lock test assertion was **vacuous**. The chain mock is one shared object, so every `.for()` lands in one flat array and the child claim also locks `FOR UPDATE` — deleting the parent's lock left the test passing. `.for()` now records how many `.where()` calls preceded it, binding the lock to the parent's own query.

Deferred, with reasons stated in the fix commit: public **asset** routes (`images`, `line-image`, `contract-file`) check the token but not `status`/`publicLinkRevokedAt` — real, but it is W04's declared "asset closure" scope and not a cross-tenant leak; the BullMQ enqueue-before-commit window is pre-existing code this wave does not touch; and the route-path audit has only `c.json` after it, so its window is process death alone.

## Verification

- Typecheck clean.
- 24,049 API tests pass. The one failure in the full run (`ipAllowlistMode.config.test.ts`) passes 4/4 in isolation and is an unrelated env-ordering flake — this wave touches nothing under `config/`.
- The supersede suite was mutation-tested: replacing the guarded flip with a bare `WHERE id = ?`, letting an accepted parent through, dropping the `publicLinkRevokedAt` stamp, and removing the parent `FOR UPDATE` each fail a named test.

Rebased onto current `main` so integration tests run against a live base rather than a stale one.

### DIFF
--- a/apps/api/src/jobs/quoteSendQueue.test.ts
+++ b/apps/api/src/jobs/quoteSendQueue.test.ts
@@ -76,6 +76,12 @@ vi.mock('drizzle-orm', () => ({
   eq: (col: unknown, val: unknown) => ({ col, val }),
   and: (...args: unknown[]) => ({ and: args }),
   isNull: (col: unknown) => ({ isNull: col }),
+  // The worker now audits `quote.superseded`, which pulls auditEvents into this
+  // module graph, and that module uses sql`` at import time.
+  sql: Object.assign(
+    (strings: TemplateStringsArray, ...values: unknown[]) => ({ strings, values }),
+    { raw: (value: unknown) => ({ raw: value }) },
+  ),
 }));
 // Same constructor shape as the real class (services/quoteTypes.ts) — the
 // module under test both constructs and callers instanceof-check it.
--- a/apps/api/src/jobs/quoteSendQueue.ts
+++ b/apps/api/src/jobs/quoteSendQueue.ts
@@ -35,6 +35,7 @@ import { captureException } from '../services/sentry';
 import { db, withSystemDbAccessContext } from '../db';
 import { quotes } from '../db/schema/quotes';
 import { sendQuote, type SendQuoteEmailOptions } from '../services/quoteLifecycle';
+import { writeAuditEvent, requestLikeFromSnapshot } from '../services/auditEvents';
 import { QuoteServiceError, type QuoteActor } from '../services/quoteTypes';
 
 const QUOTE_SEND_QUEUE = 'quote-send';
@@ -181,6 +182,29 @@ export async function processQuoteSendJob(data: QuoteSendJobData): Promise<void>
     // out — the delayed path has no request to return `emailed:false` to.
     // (Success needs no write: the flip already cleared the marker.)
     const result = await sendQuote(quoteId, actor, emailOpts);
+    // Mirror the direct-send route's supersede audit: a scheduled send retires
+    // the parent just as permanently, and that must be traceable regardless of
+    // which path issued it. Wrapped because an audit failure must never fail a
+    // send that has already committed and emailed.
+    if (result.superseded) {
+      try {
+        await writeAuditEvent(requestLikeFromSnapshot({}), {
+          orgId: result.quote.orgId,
+          action: 'quote.superseded',
+          resourceType: 'quote',
+          resourceId: result.superseded.parentQuoteId,
+          result: 'success',
+          details: {
+            supersededByQuoteId: quoteId,
+            previousStatus: result.superseded.previousStatus,
+            revisionNumber: result.quote.revisionNumber,
+            emailed: result.emailed,
+          },
+        });
+      } catch (auditErr) {
+        console.error('[quotes] failed to audit supersede for quote', quoteId, auditErr);
+      }
+    }
     if (!result.emailed) {
       await db.update(quotes)
         .set({ sendEmailReason: result.emailReason ?? 'send_failed' })
--- a/apps/api/src/jobs/quoteSendQueue.ts
+++ b/apps/api/src/jobs/quoteSendQueue.ts
@@ -163,6 +163,13 @@ export async function processQuoteSendJob(data: QuoteSendJobData): Promise<void>
   // the real-DB integration suite: an in-transaction catch-and-stamp was
   // undone by its own rethrow, so the failure banner could never appear).
   let sendError: unknown;
+  let supersedeAudit: {
+    orgId: string;
+    parentQuoteId: string;
+    previousStatus: string;
+    revisionNumber: number;
+    emailed: boolean;
+  } | undefined;
   const failed = await withSystemDbAccessContext(async () => {
     // Atomic claim: fire only if this job is still the row's registered
     // schedule, and take that registration out from under any concurrent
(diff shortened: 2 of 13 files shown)


ToddHebebrand:
## Review round 2 — five specialist lenses

Ran five lenses over the diff (correctness, silent-failure, test-coverage, comment-accuracy, type-design) after the earlier Codex pass. They independently cleared the core design again — the hard invariant holds, the CAS guards are genuine DB predicates rather than read-then-write, and the zero-row swallow/throw asymmetry is correctly applied at every site. Everything below was verified by hand, usually by mutation, before being acted on. Fixed in `48b8381a7`.

### The finding that mattered most: a test that could not fail

The predicate walker flattened a Drizzle `WHERE` into a bag of `{column, value}` pairs — no operators, no tree shape. So changing the parent flip's `and(...)` to `or(...)` left **all ten supersede tests green**, while the predicate became:

```sql
WHERE id = 'q1' OR status IN ('sent','viewed','declined','expired')
```

That supersedes and revokes the public link of *every* sent/viewed/declined/expired quote reachable in that RLS context. `inArray`→`notInArray` (which supersedes exactly the accepted parents this wave exists to protect) and `eq`→`ne` were equally invisible. I confirmed all three mutations passed green.

Predicates are now compiled with `PgDialect` and asserted as exact SQL strings, so operator and structure are part of the assertion. Re-verified that each of the three mutations now fails.

This is the third variant of the same vacuous-assertion class in this feature — first the array-blind walker, then a globally-scoped `.for('update')` check, now operator-blindness.

### Also fixed

- **The supersede audit existed on only 2 of `sendQuote`'s 4 call paths.** Bulk-send and the AI tool dropped `result.superseded` entirely, so bulk-sending twelve revisions retired twelve customer-facing quotes and killed twelve public links with **zero audit rows** — on exactly the paths nobody watches. Payload now single-sourced, all four paths emit.

  Worth noting: the obvious refactor here is a helper that *writes* the audit. That silently anonymises every route-side row, because request paths need `writeRouteAudit` to attribute the acting user while the worker and AI paths have no request. The helper returns the payload instead and each caller picks its writer.

- **`freshShare?.status ?? quote.status`** in `getQuoteShareLink`/`resendQuote` — a zero-row locked read doesn't mean "nothing changed", it means the row is gone or invisible here, and the fallback then minted or re-mailed a live accept credential from the exact stale snapshot the locked read was added to eliminate. Both now 404, matching `reviseQuote`.

- **Parent lock and flip matched on `id` alone**, and RLS is inert in the worker's system context. Both now carry `org_id`, matching what the parent-recipients read already does 100 lines away for the same reason.

- **`previousStatus: string`** discarded an invariant on a field already on the wire via `c.json({data: result})`. `SUPERSEDABLE` and `REVISABLE_STATUSES` were the same four statuses in two representations with no type link — add a status to one only and you get either a revision draft that can never be sent, or an unreachable branch. Both now derive from one `satisfies`-checked tuple, with a type predicate replacing a cast.

- **Four comments wrong in load-bearing places** — most notably the lock order was documented "parent first" when the code locks the child first and contradicted itself ten lines earlier. (I repeated that error in the previous commit message; the real no-deadlock reason is that `acceptQuote` takes a single row lock and the revision chain is acyclic.)

### Not fixed here — owner's call

**The parent row lock is held across PDF render and the outbound SMTP round-trip**, and neither transport has a timeout (nodemailer defaults to a 10-minute socket; the Mailgun path calls `fetch` with no `AbortSignal`). While a send stalls, the *parent* quote's accept, share-link, resend and view-stamp all block on that lock, each pinning a pooled connection — the `#1105` connection-exhaustion shape. The child row was already effectively locked across the email pre-PR, so this is an extension of an existing pattern, but the parent (and therefore live customer accept) is newly implicated.

The fix is to commit the state transition and mail after commit, or hand the mail to the existing `quote-send` queue. That reshapes the send path rather than hardening it, so I'd rather it be a deliberate decision than a quiet change inside this wave.

### Filed, not dropped

This wave makes `status='superseded'` reachable for the first time, which activates several downstream gaps. All recorded on the wave issues rather than widened into here: asset routes + decline honesty on #3800, portal raw-enum rendering + the `schedule_failed` misreport on #3801. Real-Postgres proof of the interleavings is W06's declared scope — the mock cannot represent a transaction boundary and this PR doesn't pretend otherwise.

Typecheck clean; 24,051 API tests pass.

PR #3883, opened 2026-08-23
feat: multi-currency wave 7 — reporting-only FX, exchange rates and dashboard segmentation (#3779)

### LINKED ISSUE #3779: W07: FX + reporting: exchange_rates, ECB feed + manual rates, dashboard segmentation
Wave W07 of #3772 — FX + reporting: exchange_rates, ECB feed + manual rates, dashboard segmentation
Branch: `feature/3772-multi-currency/wave-3779`

ToddHebebrand:
Multi-currency wave 7 implements spec §8: a global FX reference table, a daily ECB feed, operator-managed manual rates, a staleness-bounded reporting conversion primitive, and an honest per-currency segmentation plus an optional "≈ approximate" line on every money rollup.

**No document, line, payment, PDF or accounting payload gained an FX column or a converted amount.** FX is reporting-only and is never persisted onto a document. Two greps prove it (`grep -rn "exchange_rate\|exchangeRate" apps/api/migrations` returns nothing outside the new migration; the only `convertForReporting` consumer under `apps/api/src/services` is `reportingTotals.ts`), and a deny-by-default test enforces it in CI.

## `exchange_rates` — tenancy contract

`exchange_rates(rate_date, base_code, quote_code, rate numeric(18,8), source, fetched_at)`, PK `(rate_date, base_code, quote_code)`, both currency columns FK'd to `supported_currencies(code)`, `CHECK (base_code <> quote_code)`, `CHECK (rate > 0)`, `CHECK (source IN ('ecb','manual'))`.

It is a **global reference table**: no `org_id`, no `device_id`. RLS is `ENABLED` **and** `FORCED`, with the same policy pair `supported_currencies` carries — a permissive `FOR SELECT USING (true)` (tenant contexts read) and a system-only `FOR ALL` (`current_setting('breeze.scope') = 'system'`, both `USING` and `WITH CHECK`).

Of CLAUDE.md's four registration lists, **zero** apply (no `org_id`, no `device_id`): not `CORE_ORG_CASCADE_DELETE_ORDER`, not `CORE_TENANT_EXPORT_POLICY`, not the two device lists. The one registration is `INTENTIONAL_UNSCOPED` in `rls-coverage.integration.test.ts`, next to `supported_currencies`. Because that set is documentation (auto-discovery keys on the presence of an `org_id` column), the wave ships a **dedicated proof suite**, `exchangeRates.integration.test.ts`, which asserts the `pg_class` `relrowsecurity`/`relforcerowsecurity` flags and the exact `pg_policy` rows, then proves tenant-read-OK / tenant-write-denied / system-write-OK behaviourally. All four tenancy contract suites were re-run to prove the "no registration needed" conclusion rather than assume it.

Migration: `apps/api/migrations/2026-09-03-exchange-rates.sql` (plain slug — it localeCompares after the last shipped file on that date; an `-a-`/`-b-` infix would have sorted *before* it). Idempotency was proven by **executing** the file twice against the disposable test database and diffing the catalog, not by inspection.

## EUR pivot, cross rates, and why a missing pair is `unavailable`

The feed is explicitly ECB-backed, so **EUR is the fixed reporting pivot**. `from → to` resolves as `from → EUR → to` from the stored `EUR→from` and `EUR→to` legs. `from === to` is the only synthetic `1.00000000`. A pair the feed does not cover is **`unavailable`** — never 1:1, never a guess. Both legs are read in a single `DISTINCT ON` statement so a cross rate can never straddle two commits.

## Manual precedence — the self-host / air-gapped path

`source='manual'` rows are authoritative for their `(rate_date, base_code, quote_code)` cell. The daily job's upsert carries `WHERE exchange_rates.source <> 'manual'`, so **a feed run can never clobber an operator override in either commit ordering**. Proven in `exchangeRateService.integration.test.ts` under both orderings and two forced transaction-barrier interleavings (a `Promise.all` would prove nothing — it forces no commit order).

## Staleness

Lookups resolve the latest row with `rate_date <= requestedDate`; age is computed in UTC calendar days. `0..7` days is available (exactly 7 is still valid), `>= 8` is `unavailable` with `reason:'stale'`, and no prior row at all is `unavailable` with `reason:'missing'`. `DEFAULT_MAX_STALENESS_DAYS = 7`, and the HTTP endpoint never lets a client raise it.

## Daily ECB job + the air-gap switch

A daily BullMQ job plus a boot one-shot pulls Frankfurter v2 with an explicit `providers=ECB` filter (wire format verified against the live service on 2026-08-23: `/v2/rates` returns flat `{date, base, quote, rate}` rows; `/v2/latest` 404s). Coverage is read from the response — an uncovered quote currency is reported, never synthesized. Permanent failures raise `UnrecoverableError`; transient ones retry. All HTTP I/O happens outside any DB context (#1105 connection-hold rule).

Two new env vars, **both documented AND compose-mapped** — an unmapped var is inert, and the entire air-gap story depends on it (neither compose file uses `env_file:`; every variable is hand-threaded into a service `environment:` block, which is exactly why `envComposeParity.test.ts` exists):

- `EXCHANGE_RATE_SYNC_ENABLED=false` — the air-gapped / self-hosted kill switch.
- `FRANKFURTER_BASE_URL` — internal-mirror override.

Both are in `.env.example` and `deploy/.env.example` with commented-out defaults, and mapped into the `api` service of `docker-compose.yml` and `deploy/docker-compose.prod.yml`.

## Manual-rate API is platform-admin only

`GET/PUT/DELETE /admin/exchange-rates` sit under `platformAdminMiddleware`, with `requireMfa()` on the two mutating verbs and a `.strict()` body schema (a client cannot smuggle `source: 'ecb'` in and forge provenance). **A partner-scoped write would be a cross-tenant mutation by construction**: the table has no tenant axis and its `SELECT` policy is `USING (true)`, so one partner's rate would move every partner's dashboard. Adding a partner axis would contradict §8's "global table". The spec frames manual rates as the self-host / air-gapped path, which this satisfies. Flagged here for the owner to overrule if the product intent differs.

## All reporting conversion happens server-side

Spec §8 names `convertForReporting(amount, from, to, date)` as *the* reporting primitive, so it is the single implementation and **the browser never multiplies money**. `apps/api/src/services/reportingTotals.ts` converts and totals; `GET /billing/reporting-totals` returns exact strings the web only formats. Totals accumulate in **`bigint` minor units**, parsed textually from the fixed-scale strings and formatted back the same way — `toMinorUnits` is `Math.round(n * 100)` on a double, fine for one line, lossy for a portfolio; the totals are tested above `Number.MAX_SAFE_INTEGER`.

The endpoint **derives the target currency from the actor's partner**, so organization-scoped viewers keep their approximate line (`/orgs/partners/me` is partner-scope only). Reporting presentation helpers live under `apps/web/src/lib/reporting/`, never in `components/billing/shared/format.ts` — a converted value must not be one import away from a document editor.

## Real per-currency PartnerDashboard MRR

The dashboard's blind sum (a hardcoded `mrr: 0` behind a `'USD'` label) is replaced with real per-currency MRR from active contracts. MRR resolves catalog lines through the **wave-3 price book**, so the dashboard agrees with the invoices it predicts. The deprecated `mrr` field is kept as an always-zero value for **one release** so an already-loaded web bundle cannot break mid-deploy.

## Approximate-line rules

- Per-currency segmentation is **always** visible, in every state — it is the authoritative number.
- The approximate line is optional, labelled approximate, and **discloses the oldest contributing rate date**.
- **One unavailable leg suppresses the WHOLE line.** Never a partial approximate total, never a placeholder, never a zero.

Wired into all six rollups (invoices, quotes, contracts, ticket time billing, timesheets, PartnerDashboard MRR) across all eight locales. Every pre-existing single-currency assertion passes unchanged, because `ApproximateMoneyLine` renders `null` for a single-currency book.

## Deny-by-default document-boundary guard

`exchangeRateBoundary.test.ts` scans **every** production file under `apps/api/src`, `apps/web/src`, `apps/portal/src`, `packages/shared/src` and `ee`, and fails on any FX import not on a short, reasoned allowlist. No enumerated denylist is trustworthy here: the first draft of this plan missed `quoteToContract.ts`, `stripeSettle.ts`, `quoteToPax8Order.ts`, `quoteOrderService.ts`, `stripeWebhook.ts`, `contractRenewal.ts` and `quotePay.ts`, and named a `depositMath.ts` path that does not exist.

## One spec-wording resolution, flagged for the owner

§8 calls manual rates "separate rows" in the same sentence that fixes the PK as `(rate_date, base_code, quote_code)` — which makes two rows for one date/pair impossible. This wave keeps the spec's **literal PK** and enforces precedence with the conditional upsert, so the owner-fixed invariant (the daily job can never overwrite a manual rate) holds under both commit orderings.

The consequence, stated rather than buried: **deleting a manual override does not restore that day's replaced feed value.** Online, the next feed run repopulates the cell; air-gapped, the lookup falls back to an earlier eligible row. Adding `source` to the PK would instead force every lookup to carry an `ORDER BY (source='manual') DESC` tiebreak and let two contradictory rates coexist for one date/pair.

Closes #3779


ToddHebebrand:
Pushed three review fixes (fresh CI run triggered, which also re-runs the flaky `Test Web` job):

- `d4aaa14e3` — `buildGroupsParam` now quantizes each leg at the currency minor unit before serializing. `sumByCurrency` accumulates in JS `number`, so a rollup emits e.g. `24714.529999999995`; the server's `toMinorBigInt` rejects any residue past the minor unit (400 `INVALID_RATE`), so the approximate line silently failed to render on a large share of real multi-currency rollups. Same commit: an unusable leg (negative amount from a credit note, non-finite value) was *skipped*, which produced an `available` total over a basis that silently omitted a currency — a partial approximate total, forbidden by §8. Any unusable leg now suppresses the whole line.
- `2013986b9` — `toFixedScale` in `frankfurterClient` normalized the whole part via `Number()` despite documenting itself as float-free; now textual, matching `assertRate`.
- `167fd3eb9` — plan doc Self-Review bullet (e) corrected (stale pre-review text that contradicted Global Constraints on env-var documentation).

Not changed: `deriveCrossRate`'s per-currency round-trip (perf nit, N ≤ 34, no table touched) and `fromMinorBigInt`'s `"1000.00"` wire form for zero-decimal currencies (round-trips consistently through `toMinorBigInt`, formats correctly via `formatMoney`).

Issue #1733, opened 2026-06-21
Scheduled patching did not run (Sunday 2am) — orphaned patch_jobs row never enqueued to executor


ToddHebebrand:
## Summary

A scheduled patch run for org **Toothzone** (US instance, us.2breeze.app) appeared to not run during a product-owner UI walkthrough. The schedule is **weekly, Sunday 02:00**.

Investigation shows the scheduler **did fire correctly** (timezone handling is fine), but the created `patch_jobs` row is **stuck in `status='scheduled'` and never progressed to execution** — it was never observably picked up by the executor and there is no failure surfaced anywhere. This is a **silent create-then-enqueue gap**, not a timezone bug.

## Production findings (US, read-only)

Org **Toothzone** `7a4eab04-01df-461e-b730-fa77aeace3c9` (partner OliveTech `ef6dcd90-...`).

**Stored schedule** — `config_policy_patch_settings` for config policy `28ae57f1-fd18-43c6-9037-26a80dc3c443` (assigned at org level to Toothzone):
- `schedule_frequency='weekly'`, `schedule_time='02:00'`, `schedule_day_of_week='sun'`, `auto_approve=false`
- **No timezone column on this table.** TZ is resolved per-device at scan time. Toothzone has no org timezone set (`organizations.settings->>'timezone'` is empty), so it correctly fell back to the **partner OliveTech timezone `America/Denver`**.

**The schedule DID fire (timezone is NOT the problem):**
- One `patch_jobs` row exists: `bbc2a342-1c82-4e9f-be18-d663e2af9e31`, `name='Scheduled Patch Job - Windows Workstations'`, `config_policy_id=28ae57f1-...`, `devices_total=9`, `status='scheduled'`.
- `targets->>'scheduleOccurrenceKey' = '2026-06-21T02:00'`, `targets->>'resolvedTimezone' = 'America/Denver'`.
- `scheduled_at = 2026-06-21 08:00:13 UTC` = **02:00 America/Denver (MDT, UTC-6)**. 2026-06-21 is a Sunday. The window evaluation was exactly correct.
- API log: `[PatchScheduler] Created job bbc2a342-... for config policy 28ae57f1-... (9 devices, America/Denver, 2026-06-21T02:00)`.

**But it never executed:**
- `devices_completed=0`, `devices_failed=0`, `devices_pending=9`. **Zero `patch_job_results` rows.**
[...]

PR #3912, opened 2026-08-23
fix(api): stop reporting expected BullMQ retries as error-level Sentry events (BREEZE-1A, BREEZE-1J)

ToddHebebrand:
## Summary

Two long-running Sentry issues — **BREEZE-1A** (342 events, `enqueueScanResults`) and **BREEZE-1J** (12 events, `Worker.processDesktopSessionFinalizationJob`) — are the same shape: a BullMQ job handler `throw`ing to *ask for a retry* on an **expected** condition, surfaced by `attachWorkerObservability` as an error-level exception, with nothing left in the payload to say what happened.

`scrubEvent` (`services/sentry.ts`) deletes `message`, `logentry`, `extra` and `contexts` from every outbound event and rewrites the exception value to `[redacted]`. Both issues therefore arrive as `Error: [redacted]` with one app frame. The exception **type** does survive (`rebuildSafeException` keeps a structural class name), so this PR leans on named error classes rather than on new allowlisted tags.

## BREEZE-1A — patch scheduler orphan reconcile

The events are the `#1733` orphan-recovery **counter**, not a failure:

```ts
captureException(new Error(`[PatchScheduler] Recovered ${recovered} orphaned scheduled patch job(s) — #1733 race active`))
```

Proof it is that exact line: the long-silent **BREEZE-14** carries the unredacted message and the frames
`enqueueScanResults(index.cjs:381627:7)` / `Worker.processFn(index.cjs:381650:29)`; BREEZE-1A carries
`enqueueScanResults(index:476778:7)` / `(index:476801:29)` — identical 23-line delta and identical columns.
BREEZE-14 stops 2026-07-26; the message/context scrubber landed 2026-07-27 (#2843, a50769487); BREEZE-1A starts 2026-08-02. Same call site, regrouped by the scrubber.

So a *successful backstop* was being reported at error severity once per 60s scheduler tick. A healthy recovery is visible for exactly one sweep (`processExecutePatchJob` claims the row out of `scheduled`), so the sustained bursts mean the re-enqueue was **not taking effect** — and nothing in 342 identical events said so.

Changes:

- **`patchJobExecutor.resolveActiveQueueJob`** — the real defect. `queue.add({ jobId })` is a **silent no-op** in BullMQ when a job hash with that id still exists, so "this id is free" has to be true. Two holes broke that: a `remove()` rejection was swallowed into `console.error` and the helper still answered *absent*; and `getState()` also answers `'unknown'` (hash present, in no list), which matched neither branch and fell through to the same no-op add. Either one strands a `patch_jobs` row in `status='scheduled'` forever. Removal failures now raise a named `StaleQueueJobRemovalError` — unless the job raced back into a reusable state, which is a correct outcome and is reused. `filterOrphanedJobIds` catches that per id: a wedged id is reported and skipped (never handed back as recoverable), so it cannot cost every other orphan its recovery.
- **`patchSchedulerWorker.enqueueScanResults`** — a **new** orphan is a warning-level `PatchOrphanRecoveredNotice` (rate stays observable, nobody gets paged); the **same** job id recovered on 5 consecutive sweeps escalates once to `PatchReconcileStalledError` naming the job. The three capture sites in this function also get a `patch_reconcile_stage` tag so they stop being indistinguishable under the scrubber.

## BREEZE-1J — desktop session finalization

`processDesktopSessionFinalizationJob` had four bare `new Error(...)`s. In production all four render identically. The dominant one is `stop_pending`: `finalizeDesktopSessionOnce` fails closed until the agent acknowledges `desktop_stream_stop`, and the WS close path only enqueues this job **because** the inline finalize already returned `stop_pending` — i.e. the first queue attempt is expected to fail, and the 1s/2s/4s/8s ladder usually clears it. Recovery is owned by the 30s orphan scanner regardless.

Changes:

- Named classes per condition, so the scrubbed event says which one fired: `DesktopFinalizationIntentAbsentError`, `DesktopFinalizationIntentMismatchError`, `DesktopFinalizationStopPendingError`, `DesktopFinalizationIntentReleaseError`.
- An **already-released intent** is a lost race with the inline close path / orphan scanner that no retry can win, so the handler calls `job.discard()` before throwing instead of burning four more attempts.
  `discard()` and not `UnrecoverableError`: BullMQ decides unrecoverable by `instanceof UnrecoverableError || err.name === 'UnrecoverableError'`, and `err.name` is also the only field Sentry's scrubber lets through — inheriting the marker means either the retry stops or the event says which condition fired, never both. Importing an `UnrecoverableError` *value* here also reaches the ~92 suites that stub `bullmq` with a partial `vi.mock` (this module is pulled in transitively via `routes/desktopWs`) and breaks them at collection time; the first push of this branch did exactly that to `staleCommandReaper.test.ts`, `softwareComplianceWorker.test.ts` and Integration Tests shard 1/4. `discard()` is an instance method already on the job the handler holds, so it costs neither.
- `attachWorkerObservability` gains an optional `classifyFailure` hook (default behaviour unchanged for all 38 workers). The desktop worker maps `stop_pending` to *warning, report once on exhaustion* — every held attempt is still `console.warn`ed, so nothing becomes silent.

## Verification

- `pnpm exec tsc --noEmit --project apps/api/tsconfig.json` — clean (exit 0).
- `eslint` on the eight changed files — clean.
- `pnpm --filter @breeze/api test:run` — see PR comment for the exact counts. One pre-existing failure on this machine, reproduced on the clean base: `src/middleware/bodyLimit.test.ts > every file that registers its own bodyLimit is recorded here` (5s `testTimeout` on a filesystem scan).
- No live-DB suites were run (no database available in this worktree).

## Follow-up not done here

`finalizeDesktopSessionOnce` returns a bare `'stop_pending'`, discarding `DesktopStopProof.reason` (`agent_disconnected` / `delivery_unacknowledged` / `result_unavailable`) — the single most useful diagnostic. Threading it out changes the signature across `desktopWs`, orphan recovery and the reconcile fence, so it is left for its own change.


ToddHebebrand:
### Local verification (commit `10715b3cb`)

| command | result |
|---|---|
| `pnpm exec tsc --noEmit --project apps/api/tsconfig.json` (`NODE_OPTIONS=--max-old-space-size=8192`) | **exit 0**, no output |
| `eslint` on the 8 changed files | **exit 0**, no output |
| `pnpm --filter @breeze/api test:run` (full unit suite) | **1440 test files passed, 3 skipped, 0 failed** · **24 631 tests passed, 22 skipped, 0 failed** |
| targeted: `workerObservability` + `desktopSessionFinalizationWorker` + `patchSchedulerWorker` + `patchJobExecutor` + `staleCommandReaper` + `softwareComplianceWorker` | 6 files, **125 passed** |

The full run still exits 1 on `Errors 2` — two pre-existing unhandled rejections in `src/services/streamingUpload.test.ts` (`TypeError: Invalid state: ReadableStream is already closed`, undici teardown). That file shares no import path with anything in this PR (it pulls only `node:*` and `@fastify/busboy`), and the same two errors reproduce when it is run on its own.

### A regression this branch introduced and then fixed

The first push imported `UnrecoverableError` from `bullmq` as a **value**. `desktopSessionFinalizationWorker` is reachable from `routes/desktopWs`, and ~92 suites stub `bullmq` with a partial `vi.mock`, so three of them died at collection time: `staleCommandReaper.test.ts`, `softwareComplianceWorker.test.ts`, and **Integration Tests shard 1/4**. `10715b3cb` drops the value import in favour of `job.discard()`, which sets the same `discarded` flag `shouldRetryJob` already honours. All three now pass locally; CI is re-running.

### No live-DB suites were run
No database is available in this worktree, so `test:integration`, `test:rls` and the other contract configs were not run locally — CI's `Integration Tests` shards are the signal for those. This PR adds no tables, columns, or migrations, so no cascade/export-policy registration applies.

Issue #3880, opened 2026-08-23
cascadeDeleteOrg has no end-to-end integration coverage — only its order list is tested


ToddHebebrand:
Found while writing the W06 revision-lineage tests (#3879).

## What I hit

I tried to drive a real org erasure over a revision chain by calling `cascadeDeleteOrg(orgId, actorId)` from an integration test against a real Postgres with the full migration set. It aborted:

```
[tenantCascade] DELETE from "audit_policy_states" failed for org=…:
Failed query: DELETE FROM "audit_policy_states" WHERE org_id = $1
```

`cascadeDeleteOrg` does `SET LOCAL ROLE breeze_audit_admin` for tables in `AUDIT_ADMIN_REQUIRED_TABLES`, so this is a role/grant question rather than an ordering one. I did not chase it further — it was orthogonal to the FK property W06 needed, and I did not want a silent detour inside that wave.

## Why it matters

`tenantCascade.integration.test.ts` asserts five properties of `CORE_ORG_CASCADE_DELETE_ORDER` — alphabetisation, completeness, no phantom tables, no duplicates, FK children before parents — all by **reading the list and the Drizzle schema statically**. It calls `topologicalCascadeOrder()`, never `cascadeDeleteOrg()`.

So as far as I can tell, nothing anywhere executes the actual erasure against a database. The list can be perfectly ordered while the runtime path fails on the first append-only table — which is what I observed.

That is the gap CLAUDE.md's cascade section is most worried about: org erasure is a GDPR obligation, and its failure mode (strand rows under a dead tenant, or abort mid-cascade leaving a partially-deleted org) is exactly what a static list check cannot see. The file's own comment says a single table failure aborts the cascade because "partial deletion is worse than no deletion" — that behaviour is untested too.

## Suggested shape

One integration test that seeds a small but *broad* org (rows in several cascade tables, including at least one append-only/audit table and one self-referencing chain), calls `cascadeDeleteOrg`, and asserts it completes with zero rows left for that org. If it needs `breeze_audit_admin` provisioned in the test database, that provisioning is itself worth pinning — a role that exists in production but not in the test image means this path can only ever be proven in production.

Worth also deciding whether the abort-on-first-failure behaviour should have a test of its own, since it is a deliberate design choice with real consequences.

## Not urgent, but not cosmetic

No evidence of a live bug — this may simply be a test-database role gap. But "we cannot currently prove org erasure works" is a poor place to be for a contract this repo has broken five times before.

[...]

PR #3921, opened 2026-08-24
fix(api): unwrap DrizzleQueryError when reading SQLSTATE in softwareInstallMethods (#3881)

### LINKED ISSUE #3881: Two dead SQLSTATE branches in softwareInstallMethods: 409s replaced by 500s because Drizzle wraps the error
## What

Two error mappers in `routes/softwareInstallMethods.ts` read the SQLSTATE off the **top-level** error, but `drizzle-orm/postgres-js` rethrows the postgres-js `PostgresError` inside a `DrizzleQueryError` whose own `.code` is `undefined` — the SQLSTATE lives on `.cause`. Both branches are therefore unreachable, and both replace a deliberate, user-facing 409 with a bare 500.

**`softwareInstallMethods.ts:132`** — after `db.insert(softwareInstallMethods).values(...).returning()`:

```ts
if ((err as { code?: string }).code === '23505') {
  return c.json({ error: 'An install method for this platform and kind already exists' }, 409);
}
```

**`softwareInstallMethods.ts:286`** — after `db.delete(softwareInstallMethods)`:

```ts
if ((err as { code?: string }).code === '23503') {
  return c.json({ error: 'This install method is referenced by past deployments — disable it instead of deleting it' }, 409);
}
```

So today: creating a duplicate install method 500s instead of saying it already exists, and deleting one that past deployments reference 500s instead of telling the operator to disable it. The messages were written deliberately and have never been shown.

## Evidence that this is the real driver shape

Measured against live Postgres 16 with two connections contending on one row, through the actual Drizzle transaction path (done while fixing the same defect in `routes/devices/core.ts`, #3760):

```
topLevelCode : null
causeCode    : "55P03"
pgErrorCode  : "55P03"
```

`utils/pgErrors.ts` documents this exact hazard and exports `pgErrorCode()` for it. Its header comment says checks reading only top-level `err.code` "miss every Drizzle-issued insert/update and leak a raw 500 instead of mapping the conflict to a friendly error" — which is precisely what these two do.

## Suggested fix

```ts
import { pgErrorCode } from '../utils/pgErrors';
// ...
if (pgErrorCode(err) === '23505') { ... }
```

For the `23503` case, `pgErrorNode()` (added in #3760) returns the node carrying `code`, `detail` and `table_name` together, if the message should ever name the referencing table.

**Tests must assert the WRAPPED shape** — `{ cause: { code: '23505' } }`. A flat `{ code: '23505' }` fixture passes against the broken code and proves nothing.

## Scope of the sweep

I grepped `apps/api/src` for the pattern and checked each hit rather than assuming:

| Site | Verdict |
|---|---|
| `softwareInstallMethods.ts:132` | **dead** — Drizzle insert |
| `softwareInstallMethods.ts:286` | **dead** — Drizzle delete |
| `middleware/bearerTokenAuth.ts:346` | fine — `jose` `ERR_*`, not Postgres |
| `services/cfAccessJwt.ts:136` | fine — `jose` |
| `services/clientAiEntraJwt.ts:99` | fine — `jose` |

Related: #3880 is the same class from a third direction — `tenantCascade.ts:723` rethrows a plain `new Error` interpolating only `err.message` and drops the cause, which is why an org-erasure failure reports a query string and no SQLSTATE.

Happy to send the fix + wrapped-shape tests if you want it as a PR.

ToddHebebrand:
## What

Two error mappers in `apps/api/src/routes/softwareInstallMethods.ts` read the SQLSTATE off the **top-level** error. `drizzle-orm/postgres-js` rethrows the postgres-js `PostgresError` inside a `DrizzleQueryError` whose own `.code` is `undefined` — the SQLSTATE lives on `.cause`. Both branches were unreachable, and each replaced a deliberate, user-facing 409 with a bare 500.

| Handler | SQLSTATE | Message that had never been shown |
|---|---|---|
| `POST …/install-methods` (create) | `23505` | "An install method for this platform and kind already exists" |
| `DELETE …/install-methods/:methodId` | `23503` | "This install method is referenced by past deployments — disable it instead of deleting it" |

## The fix

The repo already has the canonical helper — `pgErrorCode()` in `apps/api/src/utils/pgErrors.ts` (added in #3760), which walks the `.cause` chain and handles the raw postgres-js shape too. **No new helper was introduced**; both call sites now read through it.

```ts
if (pgErrorCode(err) === '23505') { … }
```

## Tests

The two pre-existing 409 tests used a **flat** `{ code: '23505' }` fixture — which passes against the broken top-level check and proves nothing. They are now parametrised over both shapes via `it.each`:

- `drizzleWrapped()` — the real driver shape: outer error with **no** `.code`, `PostgresError` on `.cause`
- `flatPgError()` — raw postgres-js, no Drizzle wrapper

**Verified red before the fix.** With the source stashed and only the tests applied, exactly the two wrapped-shape cases fail (`expected 500 to be 409`) while the flat ones pass — so the fixtures are non-vacuous and the bug reproduces:

```
Tests  2 failed | 27 passed (29)
```

With the fix: `Test Files 2 passed (2) · Tests 49 passed (49)` (`softwareInstallMethods.test.ts` + `pgErrors.test.ts`). `tsc --noEmit` on `apps/api` exits 0.

Also added negative cases so an unrelated SQLSTATE (`23502` on create, `40P01` on delete) is **not** swallowed as a 409.

## Out-of-scope sweep — other sites with the same latent bug

I swept `apps/api/src` for every Postgres SQLSTATE literal comparison and traced each hit to where its error originates. **None of these are fixed here** — listed so they can be filed/triaged separately:

| Site | Impact |
|---|---|
| `services/tenantCascade.ts:814` (`isUndefinedTable`) | **Worst of the four.** Top-level `42P01` check on `db.execute(sql…)` results. A legitimately-missing optional table now hits the "abort the whole erasure" branch instead of being tolerated — a real GDPR org-erasure regression, not just a silent no-op. Adjacent to #3880. |
| `routes/aiAgents.ts:95` (`isUniqueViolation`) | `23505` from `db.insert/update(aiAgents)` never maps → 500 instead of the intended 409. |
| `routes/networkBaselines.ts:542` | `23503` from a Drizzle `db.transaction`/`db.delete` never maps → 500. |
| `routes/agents/connections.ts:101` | Diagnostic-only: reads `code`/`detail`/`table_name`/`column_name`/`constraint_name` off the top level, so every logged `pgCode`/`pgDetail`/… field is always `undefined`. The diagnostics that comment promises never fire. |

Confirmed **fine** (already unwrap `.cause`, or read non-Postgres codes): `partnerApi/provisioning.ts`, `services/tenantVariables.ts`, `services/pax8SyncService.ts`, `services/pax8OrderService.ts`, `services/m365ControlPlane/connectionService.ts`, `routes/partnerServicePrincipals.ts`, `routes/clientAi/admin.ts`, the five `jobs/*Worker.ts` `42P01` helpers, `aiToolsPolicyPrereqs.ts` / `aiToolsFleet.ts` (already on `pgErrorCode`), and the `jose` / Stripe / Twilio / custom-app-code sites (`bearerTokenAuth.ts`, `cfAccessJwt.ts`, `clientAiEntraJwt.ts`, `m365ConsentCallback.ts`, `m365CustomerGraph{Read,Actions}.ts`, `accountingConnectionService.ts`).

No schema, migration, tenancy, or cascade surface is touched.

Closes #3881

### DIFF
--- a/apps/api/src/routes/softwareInstallMethods.test.ts
+++ b/apps/api/src/routes/softwareInstallMethods.test.ts
@@ -100,8 +100,10 @@ function ownedCatalogItem(overrides: Record<string, unknown> = {}) {
  */
 function drizzleWrapped(code: string, message: string, extra: Record<string, unknown> = {}) {
   const pgError = Object.assign(new Error(message), { code, severity: 'ERROR', ...extra });
+  // NB: no `name` override — the real DrizzleQueryError never sets `this.name`,
+  // so a genuine instance reports `name === 'Error'`. Keeping the fixture
+  // faithful here matters: it is the whole point of this builder.
   const wrapper = Object.assign(new Error('Failed query: insert into "software_install_methods" ...'), {
-    name: 'DrizzleQueryError',
     cause: pgError,
   });
   return wrapper;
--- a/apps/api/src/routes/softwareInstallMethods.test.ts
+++ b/apps/api/src/routes/softwareInstallMethods.test.ts
@@ -91,6 +91,32 @@ function ownedCatalogItem(overrides: Record<string, unknown> = {}) {
   };
 }
 
+/**
+ * The shape drizzle-orm/postgres-js actually throws: a DrizzleQueryError whose
+ * OWN `.code` is undefined, wrapping the postgres-js PostgresError on `.cause`.
+ * A flat `{ code }` fixture passes against a broken top-level `err.code` check
+ * and proves nothing (issue #3881), so every SQLSTATE mapping is exercised
+ * through this builder as well as the flat one below.
+ */
+function drizzleWrapped(code: string, message: string, extra: Record<string, unknown> = {}) {
+  const pgError = Object.assign(new Error(message), { code, severity: 'ERROR', ...extra });
+  const wrapper = Object.assign(new Error('Failed query: insert into "software_install_methods" ...'), {
+    name: 'DrizzleQueryError',
+    cause: pgError,
+  });
+  return wrapper;
+}
+
+/** Raw postgres-js shape (no Drizzle wrapper) — still must map. */
+function flatPgError(code: string, message: string, extra: Record<string, unknown> = {}) {
+  return Object.assign(new Error(message), { code, severity: 'ERROR', ...extra });
+}
+
+const SQLSTATE_SHAPES = [
+  ['DrizzleQueryError-wrapped (real driver shape)', drizzleWrapped] as const,
+  ['flat postgres-js error', flatPgError] as const,
+];
+
 function installedMethod(overrides: Record<string, unknown> = {}) {
   return {
     id: METHOD_ID,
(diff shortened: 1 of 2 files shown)


ToddHebebrand:

**Findings:** 1 raised → addressed in 82eaee81f; 0 outstanding.
- code-reviewer: **0 findings.** Confirmed no raw `.code` checks remain in the file and that `pgErrors.ts` is reused unmodified rather than reimplemented.
- silent-failure-hunter: **0 findings.** Verified the fix introduces no new over-matching risk: both `try` blocks wrap exactly one statement (`writeRouteAudit` is fire-and-forget and outside the delete's try), `software_install_methods` has exactly one unique index and exactly one inbound FK, so a `23505`/`23503` from those statements can only mean the one thing. Non-matching paths still `throw err` and reach `app.onError` → Sentry.
- pr-test-analyzer: **1 finding (cosmetic).** Independently re-derived the wrapped shape from `drizzle-orm@0.45.2` source and confirmed the fixture is structurally faithful, but flagged that it set `name: 'DrizzleQueryError'` while the real class never assigns `this.name`. Also independently reproduced the red-before-green result. Fixed in 82eaee81f.
- comment-analyzer: **0 critical.** Verified every new comment against the `drizzle-orm` / `postgres` source; raised the same `name` nit, now fixed.

**Tests:** `apps/api` affected suite green single-fork — `vitest run src/routes/softwareInstallMethods.test.ts src/utils/pgErrors.test.ts --no-file-parallelism`, **49 passed**. `tsc --noEmit` on `apps/api` exits 0; `eslint` on both changed files clean.

**Red-before-green proof:** with the source change stashed and only the tests applied, exactly the two `DrizzleQueryError`-wrapped cases fail (`expected 500 to be 409`) while the flat-shape cases pass — `Tests 2 failed | 27 passed (29)`. That is the point of the PR: the pre-existing flat `{ code }` fixtures passed against the broken code and proved nothing.

**CI:** `Test API` **pass**, `Type Check` **pass**, `Lint` **pass** on d7eb425e5. The follow-up commit 82eaee81f is a test-fixture comment/field change only.

**Note for the reviewer:** the PR body lists four *other* call sites with the same latent bug, deliberately left unfixed to keep this PR in scope. `services/tenantCascade.ts:814` is the one worth filing promptly — its top-level `42P01` check means a legitimately-missing optional table now hits the "abort the whole erasure" branch instead of being tolerated, which is a GDPR org-erasure behavior change rather than a cosmetic 500.

**Status:** review-clean, awaiting maintainer merge. Issue #3881 left open and assigned.

Issue #3697, opened 2026-08-17
[Web] Notification channel Test reports a failed test as success — HTTP 200 + testResult.success:false shows only "Last test: Just Now"


ToddHebebrand:
## Description

Clicking **Test** on a notification channel (`/alerts/channels`) always looks like it worked. The card flips from `Never Tested` to `Last test: Just Now` — with no pass/fail verdict — even when the send definitively failed.

Repro on a clean local stack on `main` (34d0234f5):

1. `/alerts/channels` → **New Channel** → type Email, name `QA Sweep Email Channel`, recipient `qa-sweep@example.com` → **Create Channel**.
2. Click **Test** on the new channel.
3. UI shows `Last test: Just Now`. No error, no toast, no red state.

The API said the opposite. `POST /api/v1/alerts/channels/{id}/test` returns **HTTP 200** with:

```json
{
  "testResult": {
    "success": false,
    "message": "Resend error: Invalid `to` field. Please use our testing email address instead of domains like `example.com`. See our documentation for more information.",
    "details": { "recipients": ["qa-sweep@example.com"] }
  },
  "testedAt": "2026-08-17T21:02:03.233Z"
}
```

So the operator is told the channel is fine when nothing was delivered. The whole point of a Test button is to find out that alerts will not arrive — this one actively hides it. An MSP would reasonably conclude their on-call email routing is verified.

## Root Cause

The handler treats the HTTP status as the outcome and ignores the `testResult.success` discriminator in the body, then renders only the `testedAt` timestamp.

This is exactly the case `runAction` already exists to prevent — per `CLAUDE.md`, `runAction` "treats HTTP-200 `{success:false}` / `{testResult:{success:false}}` response bodies as failures (not silent no-ops)". The channel Test path is not going through it.

Note the failure text is also already user-ready — it names the exact problem and the fix. It just never reaches the screen.

## Proposed Fix

- Route the Test click through `runAction` (`apps/web/src/lib/runAction.ts`) so an HTTP-200 `testResult.success:false` surfaces as a failure toast carrying `testResult.message`.
- Render a persistent pass/fail state on the channel card, not just a timestamp — e.g. `Last test: failed 2m ago` with the reason on hover/expand. `Last test: Just Now` should never be shown for a failed test.
- Consider persisting the last test outcome so the card still shows "last test failed" after a reload.

## Affected Files

- `apps/web/src/components/alerts/` — the notification-channel list/card + Test handler (primary)
[...]

PR #3923, opened 2026-08-24
fix(web,api): alert rule "Test" reports the real evaluation verdict (#3752)

### LINKED ISSUE #3752: [Web] Alert rule "Test" always reports "Test Passed" — the UI reads success/message fields the API never returns
## Description

"Test rule" on the Alert Rules page reports **"Test Passed — Test completed successfully"** for every 200 response, including when the rule would not have fired on the tested device. There is no code path that can render a failure for a successful request.

`apps/web/src/components/alerts/AlertRulesPage.tsx:80-84`:

```js
const data = await response.json();
setTestResult({
  success: data.success ?? true,
  message: data.message ?? 'Test completed successfully'
});
```

The server (`apps/api/src/routes/alerts/rules.ts`, `POST /alerts/rules/:id/test`) returns:

```
{ rule, device, targetMatch, conditionResults, wouldTrigger, testedAt }
```

There is no `success` key and no `message` key. So `data.success` is always `undefined`, `?? true` fires on every call, and `data.message` always falls back to the hardcoded string. The fields carrying the actual verdict — `wouldTrigger` and `conditionResults` — are not read anywhere in the file.

## Impact

An operator tests an alert rule against a device to confirm it will fire. The rule does not match (`targetMatch` false, or a condition evaluates false), and the UI shows a green "Test Passed". They ship the rule believing it is verified, and it never alerts. The failure is invisible precisely at the moment someone is trying to be careful.

This is the same shape as #3697 (fixed in #3733): an HTTP 200 whose body carries the real verdict while the UI renders a hardcoded positive. It is worse here — #3697 at least had a `testResult.success` field that was being ignored, so the data was present. Here the client reads keys the server has never sent, so no amount of correct server behavior can produce a failure state.

## Suggested fix

Read the real verdict. `wouldTrigger` is the boolean the UI should gate on, and `conditionResults` (`{condition, result, reason}[]`) is what makes a negative result actionable — it says *which* condition failed and why, which is what an operator debugging a rule actually needs.

Distinguish three states rather than two: the request failed; the rule was evaluated and would fire; the rule was evaluated and would not fire (with the failing conditions listed). Today the second and third are collapsed into a green box.

## Note on runAction

`AlertRulesPage.tsx` is on `RUN_ACTION_MIGRATION_BACKLOG`, but that is orthogonal — this is a body-shape mismatch, not a missed toast, so wrapping the call in `runAction` would not have caught it. Worth remembering when triaging the rest of that backlog: the guard covers unsurfaced *failures*, not fabricated *successes*.

Found during the 2026-08-20 GH queue sweep by running a silent-failure pass for siblings of the #3697/#3733 pattern. The sweep found this one and cleared TdSynnex, PsaConnectionsPage, ScriptTestRunner, the M365 retest cards, and the Pax8/Monitoring test handlers as correctly gating on the body's real verdict.

ToddHebebrand:
Closes #3752

## The bug

`AlertRulesPage.tsx` read `data.success` and `data.message` from `POST /alerts/rules/:id/test`. The endpoint returns neither, so `data.success ?? true` fired on every 200 and the modal rendered a green **"Test Passed — Test completed successfully"** whatever the evaluation concluded. `wouldTrigger` and `conditionResults` — the fields carrying the actual verdict — were read nowhere in the file.

## Why this could not be fixed client-side

Picking up @bdunncompany's analysis in the issue thread: gating the UI on `wouldTrigger` as written would have replaced a fabricated green with a fabricated red, because the endpoint did not compute a verdict. Confirmed against `origin/main` and fixed here. Five defects in the same handler:

| Defect | Effect |
|---|---|
| Condition evaluation was simulated — `result: false` pushed for every top-level key of `template.conditions` | The verdict was a function of key count, not device state. No rule carrying a condition could ever report that it would fire |
| Read `template.conditions` directly | The firing path resolves `overrideSettings.conditions ?? template.conditions` (`alertService.ts`), so an override-conditioned rule was tested against conditions it does not use |
| `targetMatch` only evaluated for `targetType === 'device'` | Every other target type started `true` and stayed `true` — an unevaluated match reported as a match |
| `wouldTrigger = targetMatch && conditionResults.every(r => r.result)` | Wrong verdict function for an OR group, which can trigger with conditions unmet |
| `alertRules.isActive` ignored | A disabled rule — filtered out by `getApplicableRules`, so it never fires — could report that it would |

And two reachability defects that meant the feature had never worked end to end:

- The client sent **no body**, while the route validates `{ deviceId: z.string().guid() }` as required — so every test 400'd before reaching any logic. What users saw was a red failure on every test, not the green one described in the issue; the fabrication became reachable the moment a body was supplied.
- `AlertRuleList` accepted an `onTest` prop it **never rendered**. There was no Test button in the UI at all, so the modal was unreachable.

## The fix

**API** (`apps/api/src/routes/alerts/rules.ts`) — compute a real verdict:

- Run the production `evaluateConditions()` against the **effective** conditions (`overrideSettings.conditions ?? template.conditions`), mirroring `getApplicableRules`.
- Match targets the way the firing path does: `all` / `org` / `site` / `group` (real `device_group_memberships` lookup) / `device`.
- `wouldTrigger = rule.isActive && targetMatch && evaluation.triggered`.
- Response gains `targetReason`, real `conditionResults`, `evaluationContext` (metric / actual value / threshold, so a negative is actionable), effective severity, and `rule.enabled`. No `success` / `message` keys are introduced — the client now reads what the server actually computes.

**Web** (`AlertRulesPage.tsx`, `AlertRuleList.tsx`):

- Test modal asks which device to evaluate against and sends `{ deviceId }`.
- Three distinct states instead of two: request failed / **Rule would fire** / **Rule would not fire**, the latter listing the unmet conditions and the targeting reason.
- A 200 whose body carries no `wouldTrigger` boolean is surfaced as a **failure**, not assumed to be a pass — that defaulting is the whole bug.
- `AlertRuleList` renders a Test action wired to the existing `onTest` prop.
- 17 new `alerts` i18n keys, translated across all 8 locales (parity suite green).

## Notes for review

- **`runAction` not wired here, deliberately.** `AlertRulesPage.tsx` stays on `RUN_ACTION_MIGRATION_BACKLOG`: the verdict is a three-state inline render in a modal, not a toast, and its other mutation handlers are unmigrated so the file cannot leave the backlog in this PR. Failures are surfaced inline, never silently. `runAction.ts` / `runActionAllowlist.ts` are untouched to avoid conflicting with the in-flight #3697 fix.
- **Locale-file conflict risk:** the sibling #3697 PR also edits `apps/web/src/locales/*/alerts.json`. Whichever merges second will need a trivial JSON conflict resolution.
- **Pre-existing divergence, deliberately not changed:** `formatAlertRuleResponse` exposes multi-target rules via `overrides.targets.ids`, but `getApplicableRules` evaluates only the persisted `targetType`/`targetId` columns. This PR mirrors the evaluation path, so a rule with extra override target ids will honestly report "not targeted" for devices under those extra ids. That surfaces the real product behaviour rather than hiding it; worth a separate issue.

## Verification

- `apps/api` alerts routes: **15 files, 186 tests passed** (`vitest run --no-file-parallelism src/routes/alerts/`)
- `apps/web` alerts + locale parity + no-silent-mutations: **17 files, 295 tests passed**
- New `rules.testVerdict.test.ts` (13 tests): **9 fail against the pre-fix handler**, all pass after.
- New `AlertRulesPage.testVerdict.test.tsx` (8 tests): **all 8 fail against the pre-fix components**, all pass after.
- `tsc --noEmit` clean in `apps/api`; `astro check` **0 errors, 0 warnings** in `apps/web`.

### DIFF
--- a/apps/api/src/routes/alerts/rules.siteScope.test.ts
+++ b/apps/api/src/routes/alerts/rules.siteScope.test.ts
@@ -24,6 +24,7 @@ vi.mock('../../db/schema', () => ({
   alerts: { ruleId: 'alert.ruleId', status: 'alert.status' },
   devices: { id: 'device.id', orgId: 'device.orgId', siteId: 'device.siteId' },
   deviceGroups: { id: 'group.id', orgId: 'group.orgId', siteId: 'group.siteId' },
+  deviceGroupMemberships: { deviceId: 'membership.deviceId', groupId: 'membership.groupId' },
   sites: { id: 'site.id', orgId: 'site.orgId' },
   organizations: { id: 'org.id', partnerId: 'org.partnerId' },
 }));
--- a/apps/api/src/routes/alerts/rules.testVerdict.test.ts
+++ b/apps/api/src/routes/alerts/rules.testVerdict.test.ts
@@ -1,0 +1,324 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Hono } from 'hono';
+
+// #3752: POST /alerts/rules/:id/test never evaluated anything. It pushed
+// `result: false` for every top-level key of `template.conditions`, so
+// `wouldTrigger` was a function of key count rather than of device state — no
+// rule carrying a condition could report that it would fire, and no rule
+// without one could report that it would not. It also read the template
+// conditions directly (the firing path prefers overrideSettings.conditions),
+// and evaluated `targetMatch` only for targetType 'device' so every other
+// target type reported a match it had never checked.
+
+const { authRef, selectQueue, ruleRef, evaluateConditionsMock } = vi.hoisted(() => ({
+  authRef: { current: {} as any },
+  selectQueue: [] as unknown[][],
+  ruleRef: { current: null as Record<string, unknown> | null },
+  evaluateConditionsMock: vi.fn(),
+}));
+
+vi.mock('../../middleware/auth', () => ({
+  requireScope: () => async (c: any, next: any) => { c.set('auth', authRef.current); await next(); },
+  requirePermission: () => async (_c: any, next: any) => next(),
+  requireMfa: () => async (_c: any, next: any) => next(),
+  siteAccessCheck: () => () => true,
+}));
+
+vi.mock('../../db/schema', () => ({
+  alertRules: { id: 'rule.id', orgId: 'rule.orgId', partnerId: 'rule.partnerId', templateId: 'rule.templateId', targetType: 'rule.targetType', targetId: 'rule.targetId', isActive: 'rule.isActive', createdAt: 'rule.createdAt' },
+  alertTemplates: { id: 'template.id' },
+  alerts: { ruleId: 'alert.ruleId', status: 'alert.status' },
+  devices: { id: 'device.id', orgId: 'device.orgId', siteId: 'device.siteId' },
+  deviceGroups: { id: 'group.id', orgId: 'group.orgId', siteId: 'group.siteId' },
+  deviceGroupMemberships: { deviceId: 'membership.deviceId', groupId: 'membership.groupId' },
+  sites: { id: 'site.id', orgId: 'site.orgId' },
+  organizations: { id: 'org.id', partnerId: 'org.partnerId' },
+}));
+
+vi.mock('../../db', () => {
+  const makeSelect = () => {
+    const chain: any = {
+      from: () => chain,
+      leftJoin: () => chain,
+      where: () => chain,
+      orderBy: () => chain,
+      limit: () => chain,
+      offset: () => chain,
+      then: (resolve: (value: unknown) => unknown) => Promise.resolve(selectQueue.shift() ?? []).then(resolve),
+    };
+    return chain;
+  };
+  return {
+    db: {
+      select: vi.fn(() => makeSelect()),
+      insert: vi.fn(),
+      update: vi.fn(),
+      delete: vi.fn(),
+    },
+  };
+});
+
+vi.mock('../../services/auditEvents', () => ({ writeRouteAudit: vi.fn() }));
+vi.mock('../../services/partnerWideAccess', () => ({
+  canManagePartnerWidePolicies: vi.fn(() => true),
+  PARTNER_WIDE_WRITE_DENIED_MESSAGE: 'Partner-wide denied',
+}));
+vi.mock('../../services/alertConditions', () => ({
+  conditionPayloadsFrom: vi.fn(() => []),
+  evaluateConditions: evaluateConditionsMock,
+  retiredConditionTypeError: vi.fn(() => null),
+}));
+vi.mock('./helpers', () => ({
+  getPagination: vi.fn(() => ({ page: 1, limit: 1, offset: 0 })),
+  ensureOrgAccess: vi.fn(() => true),
+  getAlertRuleWithOrgCheck: vi.fn(async () => ruleRef.current),
+  isRecord: (value: unknown) => !!value && typeof value === 'object' && !Array.isArray(value),
+  getOverrides: (value: unknown) => (value && typeof value === 'object' ? { ...(value as Record<string, unknown>) } : {}),
+  normalizeTargetsForRule: vi.fn(),
+  getNotificationChannelIds: vi.fn(() => []),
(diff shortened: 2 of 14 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 lenses: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `comment-analyzer`.

**Findings:** 12 raised → all addressed in 71655967f; 0 outstanding.

- **code-reviewer — 1 important, fixed.** A verdict survived changing the device selection: the `<select>` `onChange` only set the device id, so the previous device's "Rule would fire" stayed on screen beside a new selection — the same misattributed-verdict shape this PR exists to fix. The verdict is now cleared on selection change, and the panel names the device it was computed for (`Evaluated against <hostname>`). No other high-confidence issues; it confirmed the target-matching switch matches `getApplicableRules()` for all five target types, that `wouldTrigger` cannot be fabricated in either direction, and that org/partner/site scoping is untouched.
- **comment-analyzer — 1 critical, fixed.** The doc comment claimed `wouldTrigger` is "the same boolean the firing path computes". It is not: `createAlert()` applies cooldown, open-alert dedup and flapping suppression *after* the condition + target check this endpoint replicates. Overstating that is the same class of bug as #3752 at smaller scale. Corrected on both sides and the caveat is now shown in the UI alongside a positive verdict, so a green result no longer implies an alert row necessarily appears. Also qualified the `getApplicableRules` mirror comment (it is the standalone-`alertRules` firing path; config-policy-linked rules resolve targets separately and are not exercised here) and documented why `evaluationContext` is returned but not yet rendered. Five other cross-file claims were verified accurate.
- **silent-failure-hunter — 0 correctness findings.** Confirmed no fabricated positive (the `typeof data.wouldTrigger !== 'boolean'` guard throws rather than defaulting) and no fabricated negative (a malformed 200 routes to the distinct error state, not to "would not fire"); errors from `evaluateConditions` propagate to a real 500 and a visible banner rather than a misleading verdict. Its one actionable item was the untested device-fetch failure path — added.
- **pr-test-analyzer — 8 coverage gaps, addressed.** Added: device-list fetch failure, empty device list, both 401 redirect paths, the stale-verdict case, an exact-path assertion on the test POST (the old assertion checked the body only, so a stale rule id would still have matched the mock), plus API assertions for `evaluationContext` passthrough, the template-conditions call args in the non-override base case, device/severity row content, and the `targetType: 'all'` branch directly.

**Tests:** `apps/api` alerts routes **15 files / 187 passed**; `apps/web` alerts + locale parity + no-silent-mutations **17 files / 300 passed** (both `vitest run --no-file-parallelism`). Regression value verified by reverting: **9/13** API tests and **8/8** web tests fail against the pre-fix code. `tsc --noEmit` clean in `apps/api`; `astro check` **0 errors, 0 warnings**; eslint clean on all changed files. First push was green on **Test API**, **Test Web**, **Lint** and **Test Agent**; CI is re-running on the review-fix commit.

**Deliberately not changed (follow-ups, not blockers):**
- `evaluateConditions` treats an empty AND group as triggered (`[].every()` is vacuously `true`), so a rule with empty conditions reports "would fire". That is pre-existing behaviour on the live firing path — this endpoint now reports it faithfully rather than hiding it. Changing it would alter production alerting and belongs in its own PR.
- `formatAlertRuleResponse` exposes multi-target rules via `overrides.targets.ids`, but `getApplicableRules` evaluates only the persisted `targetType`/`targetId`. This PR mirrors the evaluation path, so extra override target ids honestly report "not targeted" — surfacing the divergence rather than papering over it.
- `runAction` is not wired here and `runAction.ts` / `runActionAllowlist.ts` are untouched, to avoid conflicting with the in-flight #3697 fix. `AlertRulesPage.tsx` stays on `RUN_ACTION_MIGRATION_BACKLOG`; the silent-failure lens confirmed the manual try/catch achieves the same always-show-the-outcome property.

**Status:** review-clean, awaiting maintainer merge. Note the sibling #3697 PR also edits `apps/web/src/locales/*/alerts.json` — whichever merges second needs a trivial JSON conflict resolution.

Issue #2014, opened 2026-06-28
[dev-infra] Toasts swallowed on Astro 7 / Vite 8 dev server (module duplication across islands)


ToddHebebrand:
## Description
On the **dev server** (`astro dev`, Astro 7 / Vite 8), `runAction` success/error toasts never render app-wide. Confirmed on Alerts acknowledge, Invoices payment/reverse, Site save, and device actions. This breaks manual QA and any e2e test that asserts on toast feedback.

**Important: this is dev-only — verified NOT present in a production build** (see verification below), so it is not a release blocker for the Astro 7/Vite 8 upgrade (#2008). Filing because it silently breaks the dev/QA/e2e feedback loop.

## Verification
- Dev (`astro dev`): click Alerts → Acknowledge → no `[data-testid=toast]` appears within 4s; alert still flips to "Acknowledged".
- Prod (`pnpm build:prod && astro preview`, same origin via caddy): same action → `"Alert acknowledged"` toast renders (`[data-testid=toast]` present). Toasts work.

## Root Cause
`apps/web/src/components/shared/Toast.tsx` delivers toasts via a **module-level singleton** (`addToastFn` / `pendingToasts`, lines 22-23). `showToast` (`apps/web/src/lib/runAction.ts`) and `ToastContainer` (`apps/web/src/layouts/DashboardLayout.astro:53`) must resolve to the same `Toast.tsx` module instance. Under Astro 7 / Vite 8 dev, each `astro-island` gets its own module graph, so the page-content island's `showToast` and the `ToastContainer` island register/read on **different** copies of the module — the emitter is never shared, so toasts queue into a `pendingToasts` that nothing drains. A production `astro build` dedupes `Toast.tsx` into one shared chunk, restoring the singleton. (Same dev-vs-prod pattern as the #1301 duplicate-toast quirk.)

## Proposed Fix Options
- Hoist the emitter off the module scope onto a stable shared singleton (e.g. `window`/`globalThis` keyed) so it survives island module duplication, **or**
- Use a custom DOM event (`window.dispatchEvent`) that `ToastContainer` listens for, decoupling delivery from module identity.

## Affected Files
- `apps/web/src/components/shared/Toast.tsx` (primary)
- `apps/web/src/lib/runAction.ts`

## Reported By
Internal Playwright UI QA sweep, 2026-06-27. Evidence in `docs/testing/FEATURE_TEST_LOG.md` (Triage addendum, with prod-build A/B).

PR #3926, opened 2026-08-24
fix(web,api): persist and show WHY a notification channel test failed (#3697)

### LINKED ISSUE #3697: [Web] Notification channel Test reports a failed test as success — HTTP 200 + testResult.success:false shows only "Last test: Just Now"

ToddHebebrand:
Closes #3697.

## What was actually left

#3733 (merged) fixed the half of #3697 that the issue titled: a failed channel test rendered byte-identically to a passing one, so the verdict now says `Failed` in text rather than living in a 12px `aria-hidden` icon.

It deliberately left the issue open for the rest of the ask — *"the reason on hover/expand"* and *"persisting the last test outcome"*. That is this PR. Today the card says **Failed** and nothing else. The provider message that says exactly what to fix —

> Invalid \`to\` field. Please use our testing email address instead of domains like \`example.com\`.

— is returned in the body, toasted for five seconds, and then gone forever. Reload the page and the operator is staring at `Failed` with no way to learn what broke.

## Not the root cause the issue proposed — confirming #3733's finding

The issue asserts the Test click does not go through `runAction`. It does, and I re-verified each layer on `ea27f9afc`:

- `NotificationChannelsPage.tsx:25` — `runChannelTest` calls `runAction`; `onTest={handleTest}` at `:540` is the only production wiring.
- `apiError.ts:115-124` — `isApiFailure` already special-cases `testResult.success === false`; `extractApiError` already falls back to `testResult.message`.
- `DashboardLayout.astro:62` — `<ToastContainer client:load>` is mounted app-wide, and the module-duplication bug that once ate toasts under `astro dev` was fixed in #2014.

So **`runAction.ts` and `runActionAllowlist.ts` are untouched by this PR** — worth stating explicitly, since a parallel PR for the analogous Alert Rules "Test rule" bug (#3752) was asked to stay out of that shared plumbing. There is no overlap; these two can merge in either order.

The defect was never in surfacing the failure. It was in what survived the request.

## The change

**`notification_channels.last_test_error`** (`2026-09-07-…sql`) — mirrors the shape the TD SYNNEX integration panels already use (`last_test_status` / `last_test_at` / `last_test_error`), so the card can render the reason the way those do. Idempotent `ADD COLUMN IF NOT EXISTS`; no RLS work, since the table's existing dual-axis policies cover every command and a new column inherits them.

**Registered in `CORE_TENANT_EXPORT_POLICY`** as `included`, alongside the existing `last_test_status`. Per CLAUDE.md this is the one registration that fires on a new *column* rather than a new table, and it is the one that gets missed. Precedent for the bucket: `network_monitors.last_error` and `pax8_orders.error`. The other three cascade lists need nothing — `notification_channels` is already in `CORE_ORG_CASCADE_DELETE_ORDER:239` and has no `device_id`.

**Written on failure, explicitly `NULL`ed on success.** A column only ever written on failure is the same bug wearing a different hat: fix the channel, retest, and a green `Success` sits above last week's error. There is a test for exactly that.

**Scrubbed before it is stored** — `scrubChannelTestError` in `notificationChannelSecrets.ts`, which is where the knowledge of *which config keys are secret* already lives, so a future addition to `secretKeysForType` is covered automatically. Two passes, in this order:

1. **Literally, against the channel's own decrypted secret values.** This is the load-bearing pass. For slack/teams/webhook the destination URL **is** the credential — that is why `webhookUrl` and `url` are in `secretKeysForType` — and no pattern-matcher can spot one, because a Slack incoming-webhook URL is an ordinary `https://` string. Knowing the value is the only reliable way to remove it. Uses `split`/`join`, not `RegExp`: secrets contain regex metacharacters as a matter of course (`?`, `.`, `+` are all ordinary in a URL). Secrets under 8 chars are skipped so a short Pushover `user` cannot carve holes through unrelated words.
2. **`redactSecretsFromOutput`**, the repo's existing scrubber for persisted `lastError`-style columns. `webhookSender`, `pagerDutySender` and `pushoverSender` all splice up to 500 characters of the **destination's own response body** into the error — untrusted third-party text that can carry a bearer token or JWT the channel config never held.

Plus URL-userinfo stripping (for a destination the sender normalised past a literal match) and a 500-char cap.

**Deliberately not stripped:** the SMS recipient number that `smsSender.ts:151` concatenates into every failed send. It is not secret here — `secretKeysForType` does not list `phoneNumbers`, so the recipient list is *already* returned verbatim to every `alerts:read` caller — and `+1555…: unverified trial number` is the single most useful part of the reason.

**On the card:** clamped red text under the verdict, full string on `title`. Not the hover/expand the issue floated — a tooltip is unreachable by touch and by keyboard, and this is the one line an operator most needs. One new i18n key (`Reason: {{reason}}`), translated in all eight locales.

## Verification

| | |
|---|---|
| `apps/api` affected | 79 passed — `channels.test.ts`, `channels.testOutcomePersist.test.ts` (new), `notificationChannelSecrets.test.ts` (new), `autoMigrate.test.ts` |
| `apps/api` regression | `alerts.test.ts` 34 passed (it already asserts the `lastTestStatus` persist), `aiToolsAlerts.channelSecrets.test.ts` 12 passed |
| `apps/web` | `src/components/alerts` 14 files / 113 passed; `localeParity` + `i18n` + `no-silent-mutations` 194 passed |
| typecheck | `tsc --noEmit` clean in both `apps/api` and `apps/web` |
| migration guard | `check-migration-naming.sh: OK` (also ran on the commit hook) |

**Controls, because a green test proves nothing on its own.** Reverting `lastTestError` out of the route's `.set()` fails all 3 of the new route tests. Removing the reason block from the card fails 2 of the 4 new component tests — the other two assert *absence* of a reason (passing test, never-tested channel) and correctly pass either way.

Not runnable locally: `tenant-export-policy.integration.test.ts` and `tenantExportErasureRoundtrip.integration.test.ts` need a live DB and the only local Postgres containers belong to other agents' stacks. They run in the blocking **Integration Tests** job. Statically, the registry entry and the Drizzle model now agree on all 15 columns, and `last_test_error` matches no `SUSPICIOUS_NAME_PARTS` entry.

One note for review: a fabricated Slack webhook URL in the test fixtures tripped GitHub push protection, so both are assembled at runtime by a small helper rather than written as literals. The value the code under test receives is byte-identical.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (parallel).

**Findings:** 3 raised → all 3 addressed in `fcb705da0`; 1 accepted-and-declined; 0 outstanding.

- **code-reviewer — 0 findings at confidence ≥80.** Explicitly verified the four things most likely to be wrong here: migration idempotency/ordering (sorts after `2026-09-06-ai-billing-source.sql`, clear of the closed `2026-08-06` block), the registration lists (`notification_channels` already in `CORE_ORG_CASCADE_DELETE_ORDER:239`; the column-level `CORE_TENANT_EXPORT_POLICY` row correctly updated; no `device_id` so the device lists don't apply), RLS (no-op — policies are row-level, a new column inherits them), and that `channels.ts:685` is the only write site for these columns.
- **silent-failure-hunter — 1 HIGH.** The persist's `console.error`-only catch was fine when it carried a timestamp, but this PR routes the operator-facing *reason* through it. A dropped write leaves the card showing its **previous** verdict — potentially a stale green `Success` under a toast that just said the test failed, which is #3697 reintroduced one layer down. Fixed: `captureException` on the catch, plus an `extractRowCount` zero-row check (a zero-row UPDATE does not throw, and the channel was fetched moments earlier under the same auth, so matching nothing means a concurrent org move or a policy divergence). The request still succeeds either way — the outbound send already happened and a DB hiccup must not retro-fail it. **This also answers @bdunncompany's open question on the issue** ("decide whether a zero-row persist deserves a warning log"): yes, and now it gets one.
- **pr-test-analyzer — 2 gaps rated 6/10, both closed.** (1) A sender that *throws* takes the catch-all `Failed to test channel: …` path, whose message comes from an `Error` rather than the hand-crafted strings — network clients routinely embed the URL they were called with, and for a slack channel that URL is the credential. Now tested with a thrown error carrying the channel's own webhook URL. (2) The unsupported-channel-type `501` branch's "persists nothing" invariant was stated only in a comment; now asserted. Also added coverage for the two new logging paths. The reviewer independently traced the Drizzle mock and confirmed `updateSetRef` is not vacuous.
- **Declined:** partner-wide (`orgId: null`) persist coverage, rated 4/10 by the reviewer itself — the `.set()` does not branch on ownership axis, and the partner-wide gate is already covered at `channels.test.ts:242`.

**Tests:** `apps/api` affected 4 files / 59 passed single-fork (`channels.testOutcomePersist.test.ts` now 7, `notificationChannelSecrets.test.ts` 11, plus `channels.test.ts` and `alerts.test.ts` as regression). `apps/web` `src/components/alerts` 14 files / 113 passed; locale-parity + i18n + `no-silent-mutations` 194 passed. `tsc --noEmit` clean in both apps.

**CI:** Test API, Test Web and **all four Integration Tests shards green** — that last one is the real check on the `CORE_TENANT_EXPORT_POLICY` entry, which needs a live DB and could not be run locally.

**One red check, unrelated:** `Test Agent (Windows)` — `internal/sessionbroker` hangs in teardown on go-winio's named-pipe `Close()` and burns the full 10m timeout under `-race` only. This PR changes **zero** Go files (`git diff --name-only origin/main...HEAD | grep ^agent/` → 0). It is the known intermittent flake tracked in **#3018**, which has recurred on unrelated `apps/api`-only PRs before (e.g. #3224) with the identical stack. Main is green on this job across its last five runs. Needs a re-run of that job, not a code change — worth reopening #3018.

**Status:** review-clean, awaiting maintainer merge. Not merged and #3697 left open per the hand-off rules.

ToddHebebrand:
**CI now fully green** — every required check passes, `CI Success` included.

`Test Agent (Windows)` passed on re-run in **3m31s**, against **13m05s** (the full 10-minute Go timeout plus setup) on the failed attempt. Same commit, same code, opposite outcome in a quarter of the time — which confirms the #3018 diagnosis: a teardown deadlock in go-winio's named-pipe `Close()`, not anything in this diff. Nothing was changed between the two attempts.

Worth reopening **#3018** with this occurrence attached; it has now recurred on at least two `apps/*`-only PRs (#3224 and this one) since it was last closed for going quiet.

PR #3929, opened 2026-08-24
fix(web): stop a double-click on a confirm dialog hit-testing through to the list (#3705)

### LINKED ISSUE #3705: Confirm dialogs unmount on confirm — a real double-click can hit-test through to the list underneath
Split out of the #3703 review, at the maintainer's request. It is not specific to the dialog that PR added — it applies to all three `ConfirmDialog` call sites in `DevicesPage.tsx` and to the shared `Dialog` itself.

## The shape

All three confirm dialogs in `apps/web/src/components/devices/DevicesPage.tsx` (script-run, decommissioned-skip, and the new device-action gate at `:1392`, `:1419`, `:1453`) clear their pending state inside `onConfirm`, which unmounts the dialog. That is deliberate and documented in the file — the decommissioned-skip comment explains it well:

> Confirming UNMOUNTS this dialog, which is what makes a double-click safe — a second click has no button to hit. No `isLoading`/spinner here on purpose: that would mean keeping the dialog mounted, and then neither guard holds — `ConfirmDialog`'s `disabled` and `runBulkAction`'s `actionInProgress` both read a closure captured at render, still `false` on the second click, so the fleet reboots twice.

That reasoning is correct for the failure it was written against: the *same* button cannot fire twice, because after the state update it is detached.

## What it does not cover

`Dialog` renders through `createPortal` and returns `null` when `!open` (`apps/web/src/components/shared/Dialog.tsx:115-117`), so confirming removes the backdrop and the panel together. On a real double-click the second physical press is hit-tested **after** that removal, against whatever now occupies those coordinates — the devices list or grid underneath.

The confirm button sits centred over a dense row/card layout, and the row kebab, `Run Script`, `Wake` and `Decommission` are all live targets there. So the risk is not a duplicate of the confirmed action; it is an *unrelated* action firing from the second half of a double-click.

## Why the current tests cannot see it

`DevicesPage.test.tsx` dispatches twice at a cached, detached element. jsdom has no layout or hit-testing, so a synthetic `fireEvent.click` on a removed node cannot reach anything underneath. The existing coverage proves the same-button case and is silent on this one, which is why I did not treat it as a regression in #3703.

## Suggested direction

The safer shape is to keep the backdrop mounted through the action and guard re-entry synchronously rather than relying on unmount:

- keep the dialog open with `isLoading` while the action runs, so the backdrop keeps absorbing clicks
- add a synchronous `useRef` re-entry guard, which does not suffer the stale-closure problem the file's comment describes for `disabled` and `actionInProgress` (both read a value captured at render; a ref reads current)
- close on settle

That fixes the original double-fire *and* the click-through, without depending on the removal timing.

## Verifying it

This needs a coordinate-level browser test — Playwright `dblclick()` on the confirm button in both list and grid layouts, asserting no other command was queued. A jsdom test cannot establish or refute it, so I would rather it be checked properly than argued from reading.

I have not confirmed the click-through actually reproduces; the mechanism is legible from the portal teardown but the observation is missing. Happy to write the Playwright case if that is the shape you want, or to close this if the timing turns out not to be exploitable in practice.

ToddHebebrand:
Closes #3705.

You framed this against all three `ConfirmDialog` call sites in `DevicesPage.tsx` rather than the one #3703 added, and said you had not confirmed the click-through actually reproduces.

**It reproduces.** With the fix reverted and the new tests kept, a double-click on the device-reboot confirm queues a **fleet Wake** — an action the operator never aimed at.

## Files touched (5)

| File | What |
|---|---|
| `apps/web/src/components/shared/Dialog.tsx` | gesture-tail guard (new) |
| `apps/web/src/components/shared/ConfirmDialog.tsx` | synchronous single-fire latch (new) |
| `apps/web/src/components/shared/ConfirmDialog.test.tsx` | **new file**, 10 tests |
| `apps/web/src/components/devices/DevicesPage.tsx` | comments only — the two that documented the old "the unmount is what saves us" reasoning |
| `apps/web/src/components/devices/DevicesPage.test.tsx` | 1 new test (the reproduction) |

## The bug

`Dialog` returns `null` when `!open`, so confirming removes the backdrop and the panel together. The first press lands on Confirm, React flushes, the portal is gone — and the second press of the same physical double-click is hit-tested against whatever now occupies those coordinates. The `DevicesPage` confirms sit centred over the bulk action bar and the device rows.

The existing coverage could not see this: it dispatches twice at a cached, detached node, which proves only that the *same* button cannot fire twice.

---

## ⚠️ I diverged from your suggested direction — please accept or reject this on the merits

Your suggestion in the issue was: **keep the dialog open with `isLoading` while the action runs so the backdrop keeps absorbing clicks, add a `useRef` re-entry guard, close on settle.**

I took the ref guard as you described it. I did **not** take the hold-the-dialog-open half, and instead put the click-through fix inside `Dialog`. Two reasons:

1. **Holding the dialog open only moves the window; it does not close it.** When the action settles and the dialog finally does close, a double-click *at that moment* has exactly the same click-through. So does a double-click on **Cancel**, and so does one on the **backdrop** — neither of which an `isLoading` spinner covers at all, because neither runs an action. The teardown is the hazard, and every exit path has one.
2. **It would be a behavioural contract change at 48 call sites.** Every `ConfirmDialog` caller would have to make `onConfirm` awaitable to get any benefit, and until it did it would stay exposed. Fixing it in the shared component covers all of them without touching any.

**What your direction is better at, and what I have therefore left on the table:** for a slow network-backed action, keeping the backdrop up genuinely does absorb the *whole* original gesture, communicates progress, and blocks unrelated interaction while something destructive is pending. That is better UX than what is here. The latch below is what makes that pattern *safe* to adopt — the trap documented in `DevicesPage.tsx` ("no `isLoading` here on purpose") is now gone. I deliberately did not convert the three call sites myself: that reshapes three destructive-action flows and is your call, not a side effect of a bug fix. Happy to do it in a follow-up if you want it.

If you would rather have your original shape, say so and I will rework it.

---

### 1. `ConfirmDialog` — synchronous single-fire latch (your suggestion)

A `useRef` latch on the confirm handler. `disabled={isLoading}` and a call site's own `actionInProgress` both read a value captured at render and are still `false` when the second click lands; a ref reads current. It releases when `isLoading` settles back to false so a failed action stays retryable, and on a synchronous throw — which never reaches the close or the `isLoading` toggle — so the button cannot be permanently killed.

### 2. `Dialog` — gesture-tail guard

On teardown, a capture-phase document listener swallows any `mousedown`/`mouseup`/`click`/`dblclick` carrying `detail >= 2`.

`MouseEvent.detail` is the platform click counter. Chromium, Gecko and WebKit all derive it from the button, the time and the distance between presses, never from the DOM target — so the second press still reports `detail === 2` even though it now lands on a different element. That identifies "tail of the gesture that ended on the dialog" with no coordinate maths. (This is consistent de-facto behaviour across the three engines, not a normative guarantee; the latch above is the belt to this pair of braces.)

Three things stop it over-blocking:

- **A new gesture disarms it.** A deliberate double-click after the dialog closes opens with a press whose `detail` is 1, standing the guard down before its own second press arrives. That also bounds its lifetime — there is no timer, because OS double-click intervals are user-configurable (Windows allows up to 5s) and any fixed ceiling would either expire early or linger.
- **It only arms on a teardown a completed *click* inside the dialog could have caused** (within 150ms). Keying this on `mousedown` — which is what the first push did — armed on orderings with no gesture in flight, most concretely "start a text selection inside the dialog, dismiss with Escape", which then ate the user's next click. This also makes StrictMode's extra cleanup cycle inert. It is a time window, not causation: an action that settles and closes in under 150ms still arms, which is the conservative direction.
- **Keyboard and AT clicks report `detail === 0`** and pass through untouched.

---

## ⚠️ This guard is on the shared `Dialog`, so it affects every dialog in the app

Not just the three `DevicesPage` confirms. Swept:

- **Shared `Dialog`: 57 render sites across 50 non-test files.**
- **Shared `ConfirmDialog`: 48 render sites across 38 non-test files.**

(`admin/AccountDeletionRequestDetail.tsx` defines its own unrelated local `ConfirmDialog` and is excluded. `remote/DeleteConfirmDialog.tsx` wraps the shared `Dialog` directly and is counted in the `Dialog` totals.)

All 48 `ConfirmDialog` sites were checked against the **latch** — the only change with call-site-visible behaviour — and **none need changes**. Each is either sync-close-on-confirm (latch moot), or keeps the dialog open behind an `isLoading` that toggles in a `try/finally` (which is exactly what unlatches it for retry), or renders conditionally so the ref resets on unmount.

### Where a wanted click could now be swallowed

The guard cannot distinguish "second half of an accidental double-click" from "deliberate fast click at the same spot right after confirming" — both arrive as `detail >= 2`. Because the counter only increments for presses within a few pixels of the previous one, the cost is bounded to a repeated click at the coordinates the confirm button just vacated. Erring that way is the right side for a destructive-action confirm, but it is a real trade-off and it is documented in the code.

It also survives a client-side route change — the platform counter is not reset by SPA navigation — so if `onConfirm` navigates and the new screen puts a control where the confirm button was, the user's first click on it can be counted as the second of a pair and swallowed. Scoping the guard by coordinates would not help: a `detail >= 2` event is already, by construction, within a few pixels of the click that armed it. Clicking again works, which is the recovery in every case.

**One concrete place it is likely to be felt: `apps/web/src/components/remote/FileManager.tsx`.** Its table rows are `onDoubleClick`-to-open (`:1484` → `initiateDownload`), and both the delete confirm (`:1692`) and the disk-cleanup confirm (`:1698`) are centred modals over that table. Delete files → confirm → the table reflows → immediately double-click a row to open the next file, and that double-click can land in the vacated confirm-button spot inside the counter window and be ignored. The user clicks again and it works; nothing is lost. It is one of only three `onDoubleClick` handlers in the whole web app.

Explicitly checked and **ruled out**:

- **Nested dialogs.** All 12 files importing both were checked. Only `DRExecutionView.tsx` and `PatchApprovalModal.tsx` are genuinely nested (inner opens while outer stays mounted); in both, the outer dialog's remaining content is a sparse form / static status list with nothing clickable under the confirm's position. The other 10 are page-level siblings that cannot be open simultaneously.
- **Steppers / +/- / repeat-click controls.** None exist — quantity in the quote/invoice/contract line-item editors is a plain `type="number"` input throughout.
- **`onMouseDown`/`onPointerDown` action handlers.** Four app-wide, none near a dialog teardown: a drag-start in `DashboardCustomizer.tsx:830`, an autocomplete select in `AuditFilters.tsx:224`, and two `preventDefault`-only focus-preservers in rich-text toolbars.

**Not verified:** actual pixel overlap between a closing confirm button and the row that reflows under it. That needs a running browser, not static analysis — the FileManager risk above is a layout-plausibility judgement, not a measurement.

## Verification

**Tests.** `ConfirmDialog.test.tsx` (new, 8) + 1 new `DevicesPage.test.tsx` case. jsdom has no layout, so it cannot pick the hit-test target itself — the tests hand the second press to the element underneath, which is what a browser does. What *is* faithfully modelled is the part that is ours: the press arrives carrying `detail === 2` and the guard must swallow it. A Playwright `dblclick()` in both list and grid layouts remains the stronger proof and is worth adding to the e2e suite separately.

**Control run** (both guards reverted, tests kept):

| Test | Against `origin/main` |
|---|---|
| `a double-click on confirm cannot fire an unrelated action underneath` (DevicesPage) | **FAILS** — `sendBulkWakeCommand` called once. This is #3705. |
| `swallows the second press after the dialog is torn down` | FAILS |
| `guards a Cancel press too` | FAILS |
| `ignores the second click on a dialog that stays mounted` | FAILS |
| `survives two dialogs closing back to back` | FAILS |
| the over-blocking guards (fresh click, deliberate dblclick, keyboard click, press-then-Escape) | pass — correctly unaffected |

`does NOT arm after a press inside the dialog that ends in Escape` passes against `origin/main` (no guard, nothing to swallow) and **fails against the first commit of this PR**, which armed on `mousedown`. That is the regression it exists to catch. `releases the latch when onConfirm throws` likewise passes against `origin/main` (no latch to stick) and fails if the `catch` is stripped from the latched version.

**Local runs** (`--pool=forks --maxWorkers=1`), every directory that renders a dialog:

| Scope | Result |
|---|---|
| `components/shared` | 10 files / 73 tests |
| `components/devices` | 56 files / 671 tests |
| `components/dr` + `patches` | 11 files / 114 tests |
| `components/remote` + `contracts` | 38 files / 257 tests |
| `components/billing` + `settings` | 154 files / 1391 tests |
| `tickets`, `snmp`, `pam`, `clientAi`, `software`, `monitors`, `discovery`, `sensitiveData`, `account` | 60 files / 630 tests |
| **Total** | **329 files / 3136 tests, all passing** |

- `astro check`: **0 errors, 0 warnings** (1672 files)
- `eslint`: clean on all five touched files
- **CI: `Test Web`, `Test API` and `Lint` all green** on the first commit — that is the full web suite, and it covers what I could not finish locally.
- **Full `@breeze/web` suite: NOT run to completion locally** — concurrent runs on this host kept getting killed for resource contention. Relying on CI's **Test Web** job for the remainder. Everything above is the dialog-consuming surface, but it is not the whole suite and I am not claiming it is.

## Not addressed

- **Keyboard key-repeat through focus restoration.** `Dialog` returns focus to the trigger on close, so a held Enter could in principle re-activate it. Different mechanism from the coordinate hit-test this issue is about; out of scope here.
- **`PossibleReplacementBanner.handleDecommission`** resets its busy/close state by hand on both paths instead of `try/finally` — the one outlier among ~20 similar handlers. Functionally safe today, more fragile to future edits. Pre-existing, untouched here; worth a separate tidy.
- Converting the three `DevicesPage` dialogs to the stay-mounted spinner shape — now safe, but your call (see the divergence section).

### DIFF
--- a/apps/web/src/components/devices/DevicesPage.test.tsx
+++ b/apps/web/src/components/devices/DevicesPage.test.tsx
@@ -1292,11 +1292,15 @@ describe('DevicesPage — bulk agent commands gated on decommissioned only (#246
 
     // An impatient double-click on a bulk REBOOT must not reboot the fleet twice.
     //
-    // Two things fence this now (#3705), and the unmount is no longer either of
-    // them: ConfirmDialog holds a synchronous ref latch, so converting this to a
-    // stay-mounted + spinner pattern no longer lets the second click through
-    // (`disabled` and runBulkAction's `actionInProgress` both read a closure
-    // captured at render and still cannot help). It does NOT discriminate the
+    // Be precise about what THIS test fences, because it is narrower than it
+    // looks: it re-dispatches at a stale reference to a button already detached
+    // by the first click, so the second event never reaches React's delegated
+    // listener. The unmount is still the only thing it exercises. It stays
+    // green with the #3705 latch reverted.
+    //
+    // In production the latch is the guard that matters, because it also holds
+    // when the dialog STAYS mounted — see ConfirmDialog.test.tsx, which drives
+    // that case directly. What neither test discriminates is the
     // set-state-before-dispatch ORDER: under React 18 both updates flush
     // together at the end of the handler, so swapping them would still pass.
     fireEvent.click(confirm);
--- a/apps/web/src/components/devices/DevicesPage.test.tsx
+++ b/apps/web/src/components/devices/DevicesPage.test.tsx
@@ -1292,18 +1292,59 @@ describe('DevicesPage — bulk agent commands gated on decommissioned only (#246
 
     // An impatient double-click on a bulk REBOOT must not reboot the fleet twice.
     //
-    // What this actually fences: converting the dialog to a stay-mounted +
-    // spinner pattern (leaning on ConfirmDialog's `disabled`, or on the
-    // actionInProgress guard — which reads a stale `false` closure and cannot
-    // help) lets the second click through and fires sendBulkCommand twice.
-    // It does NOT discriminate the set-state-before-dispatch ORDER: under React
-    // 18 both updates flush together at the end of the handler, so swapping them
-    // would still pass. The unmount is the guard; the ordering is not load-bearing.
+    // Two things fence this now (#3705), and the unmount is no longer either of
+    // them: ConfirmDialog holds a synchronous ref latch, so converting this to a
+    // stay-mounted + spinner pattern no longer lets the second click through
+    // (`disabled` and runBulkAction's `actionInProgress` both read a closure
+    // captured at render and still cannot help). It does NOT discriminate the
+    // set-state-before-dispatch ORDER: under React 18 both updates flush
+    // together at the end of the handler, so swapping them would still pass.
     fireEvent.click(confirm);
     fireEvent.click(confirm);
 
     await waitFor(() => expect(vi.mocked(sendBulkCommand)).toHaveBeenCalledTimes(1));
     expect(vi.mocked(sendBulkCommand)).toHaveBeenCalledTimes(1);
+  });
+
+  // #3705, the OTHER half of the double-click. The one above proves the same
+  // button cannot fire twice. This proves the second press cannot fire
+  // something ELSE: confirming tears the portal out, and in a browser that
+  // press is hit-tested afterwards against whatever now occupies those
+  // coordinates — the bulk bar and the device rows the dialog was centred over.
+  //
+  // jsdom has no layout, so it cannot pick that target itself; the test hands
+  // the press to the bulk Wake button directly, which is what a browser does.
+  // What is faithfully modelled is the part that matters: the press arrives
+  // carrying detail === 2, because the platform click counter comes from the
+  // time and distance between presses, never from the hit-test target.
+  it('a double-click on confirm cannot fire an unrelated action underneath', async () => {
+    const { sendDeviceCommand, sendBulkWakeCommand } = await import('../../services/deviceActions');
+    vi.mocked(sendDeviceCommand).mockResolvedValue({ command: {} } as never);
+    vi.mocked(sendBulkWakeCommand).mockResolvedValue({ sent: [], failed: [] } as never);
+    vi.mocked(fetchAllDevices).mockResolvedValue({
+      data: [{ ...rawDevice(DEV_1, 'host-alpha'), status: 'online' }],
+    } as never);
+
+    render(<DevicesPage />);
+    fireEvent.click(await screen.findByTestId(`row-reboot-${DEV_1}`));
+
+    // Press 1 of the double-click: lands on Confirm, unmounts the dialog.
+    const confirmBtn = await screen.findByTestId('confirm-device-action');
+    fireEvent.mouseDown(confirmBtn, { detail: 1 });
+    fireEvent.mouseUp(confirmBtn, { detail: 1 });
+    fireEvent.click(confirmBtn, { detail: 1 });
+    await waitFor(() => expect(vi.mocked(sendDeviceCommand)).toHaveBeenCalledTimes(1));
+
+    // Press 2, now hit-testing through to the fleet Wake the operator never
+    // aimed at. Wake is deliberately UNGATED, so nothing else would stop it.
+    const wake = screen.getByTestId('bulk-wake');
+    fireEvent.mouseDown(wake, { detail: 2 });
+    fireEvent.mouseUp(wake, { detail: 2 });
+    fireEvent.click(wake, { detail: 2 });
+
+    expect(vi.mocked(sendBulkWakeCommand)).not.toHaveBeenCalled();
+    expect(vi.mocked(sendDeviceCommand)).toHaveBeenCalledTimes(1);
+    expect(vi.mocked(sendDeviceCommand)).toHaveBeenCalledWith(DEV_1, 'reboot');
   });
 });
 
--- a/apps/web/src/components/devices/DevicesPage.tsx
+++ b/apps/web/src/components/devices/DevicesPage.tsx
@@ -1429,7 +1429,9 @@ export default function DevicesPage() {
           // row underneath. Previously only the unmount protected this, and it
           // protected only the first of those two failures.
           // (The set-state/dispatch ORDER below is not load-bearing either:
-          // React batches both.) DevicesPage.test.tsx pins both halves.
+          // React batches both.) The click-through half is pinned in
+          // DevicesPage.test.tsx; the latch half in ConfirmDialog.test.tsx,
+          // which can hold the dialog mounted the way this call site does not.
           onConfirm={() => {
             const p = pendingDecommissionedSkip;
             setPendingDecommissionedSkip(null);
--- a/apps/web/src/components/devices/DevicesPage.tsx
+++ b/apps/web/src/components/devices/DevicesPage.tsx
@@ -1421,14 +1421,15 @@ export default function DevicesPage() {
         <ConfirmDialog
           open={true}
           onClose={() => setPendingDecommissionedSkip(null)}
-          // Confirming UNMOUNTS this dialog, which is what makes a double-click
-          // safe — a second click has no button to hit. No isLoading/spinner
-          // here on purpose: that would mean keeping the dialog mounted, and
-          // then neither guard holds — ConfirmDialog's `disabled` and
-          // runBulkAction's `actionInProgress` both read a closure captured at
-          // render, still `false` on the second click, so the fleet reboots
-          // twice. (The set-state/dispatch ORDER below is not what saves us:
-          // React batches both. The unmount is.) DevicesPage.test.tsx pins it.
+          // Double-click safety lives in the shared components now (#3705), not
+          // in this call site. ConfirmDialog holds a synchronous ref latch, so
+          // the fleet cannot reboot twice even if the dialog stayed mounted;
+          // and Dialog swallows the tail of the gesture after its portal is
+          // torn out, so the second press cannot hit-test through to a device
+          // row underneath. Previously only the unmount protected this, and it
+          // protected only the first of those two failures.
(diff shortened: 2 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer. (A first pass died mid-flight on an API limit and was re-run in full; these are the complete results.)

**Findings:** 9 raised → 8 addressed in 80bfb02c3, 1 deferred; 0 outstanding.

**Fixed — the two that were real bugs:**

- **The guard armed too widely.** It keyed on *any* `mousedown` inside the dialog within 1s, not on the press that caused the close. pr-test-analyzer reproduced the consequence: start a text selection inside a dialog, dismiss with Escape, and the guard armed with no gesture in flight and ate the next legitimate click. Now keyed on a completed **click** inside the dialog, window 1000ms → 150ms. The comment no longer claims causation — it says plainly this is a time window, so a sub-150ms async settle still arms, which is the conservative direction.
- **A vacuous test.** `does NOT arm after an Escape close` never pressed inside the dialog first, so `lastPointerDownAtRef` was still `0` and it passed under either rule — it proved nothing. It now presses first, and fails against the previous commit of this PR.

**Fixed — coverage and honesty:**

- Added the throwing-`onConfirm` test. My "React 19 makes this untestable" note was wrong: the throw surfaces as a window `error` event, so swallowing that event exercises the real path without destabilising the worker. pr-test-analyzer verified it fails when the `catch` is stripped.
- Added a stacked-dialog test — `disarmActiveGestureTailGuard` was the only stated design intent in the module with no coverage.
- The latch's release contract now has a **failure signal**. It depends on the call site either closing the dialog or driving `isLoading`; all 48 sites do, but nothing enforced it, and a site that did neither would leave Confirm permanently inert in silence. A blocked click while `isLoading` is false is exactly that violation, and now warns in dev.
- Three comments corrected for overclaiming: the `catch` protects the *button*, not the user, and is inert for the `() => void asyncFn()` shape every call site uses; the bulk double-click test in `DevicesPage.test.tsx` re-dispatches at a detached node, so it still exercises only the unmount and stays green with the latch reverted; and the browser-engine claims now grade confidence per engine — **Chromium and WebKit verified in source, Gecko asserted from observed behaviour, none of it normative** — with `pointerdown`/`pointerup` hedged the same way rather than stated flatly, the Windows 5s figure described as the Win32 API ceiling rather than a control-panel setting, and the SPA-navigation case owned explicitly.

**Deferred:** `PossibleReplacementBanner.handleDecommission` resets its busy/close state by hand on both paths instead of `try/finally` — the one outlier among ~20 similar handlers. Safe today, fragile to future edits. Pre-existing and untouched here.

**Call-site sweep independently confirmed:** 48 `ConfirmDialog` render sites across 38 files, all 48 `onConfirm` handlers read. Zero can hit the stuck-latch failure — every one either closes synchronously or resets `isLoading` in a `finally`. `admin/AccountDeletionRequestDetail.tsx` defines its own local component of the same name and is correctly out of scope. Shared `Dialog`: 57 sites across 50 files.

**Tests:** `shared/` + `DevicesPage.test.tsx` green (`vitest run --pool=forks --maxWorkers=1`, 11 files / 130 tests). Control against `origin/main`: 5 tests fail, including the `DevicesPage` case where the double-click queues a fleet Wake. `astro check` 0 errors / 0 warnings (1672 files); eslint clean on all five files. Full `@breeze/web` suite was not run to completion locally — CI `Test Web` is the authority and was green on the first commit; it is re-running for 80bfb02c3.

**Status:** review-clean, awaiting maintainer merge. The divergence from your suggested `isLoading` direction is called out in the PR body under its own heading — that is the one thing I would like a yes/no on before this lands.

Issue #3951, opened 2026-08-24
Windows Remote Desktop startup fails when DXGI returns no frame during initial probe


sxwxrslvvt:
### Description

On Windows, Breeze Remote Desktop consistently fails during the initial screen-capture probe even though the endpoint has an active physical console session and an active connected display.

The Breeze Viewer launches, stays at 0 FPS, and the session terminates with “Connection Error — This remote session has ended. Relaunch it from the Breeze dashboard.”

The agent/helper-side failure is:

screen capture produced no frame after 5 attempts (session may have no capturable desktop)

This happens before an SDP answer is created and before ICE negotiation begins, so TURN, NAT, HAProxy and Caddy are not involved in this specific failure.

The endpoint was explicitly awakened and visible desktop activity was generated immediately before retrying, but the failure remained fully reproducible.

### Steps to Reproduce

1. Enroll a Windows endpoint in Breeze v0.107.0.

2. Log a user into the physical console session.

3. Verify that:
   - the console session is Active;
   - explorer.exe is running;
   - dwm.exe is running;
   - a physical monitor is connected and active.

4. Open the device in Breeze and click "Connect Desktop".

5. Launch Breeze Viewer.

6. Observe that the Viewer stays at 0 FPS and the session terminates.

7. Wake the physical display and generate visible desktop activity:
   - move the mouse;
   - open/close a window;
   - interact with the desktop.

8. Immediately retry "Connect Desktop".

9. The same failure occurs:
   screen capture produced no frame after 5 attempts
   (session may have no capturable desktop)

### Expected Behavior

Remote Desktop should proceed when the Windows endpoint has an active console session and an active physical display.

A temporary DXGI condition where no new frame is available yet should not be treated as proof that the desktop is unavailable.

The session should continue to SDP/WebRTC negotiation unless screen capture is actually unavailable or a hard capture error occurs.

### Actual Behavior

The Remote Desktop session is terminated during the initial screen-capture probe.

The helper returns five consecutive nil-frame results and Breeze treats this as a fatal "no capturable desktop" condition.

No SDP answer is generated and ICE negotiation never begins.

The Viewer then displays:
Connection Error — This remote session has ended.

### Component

Remote Access / Terminal

### Version

v0.107.0

### Relevant Logs

```shell
Active console session:

User2: Active console session 12
WTSGetActiveConsoleSessionId() = 12

The normal Windows desktop processes are running in the same session:
- explorer.exe
- dwm.exe

[...]

PR #3960, opened 2026-08-24
fix(agent): force a desktop repaint before the remote-desktop startup probe (#3951)

ToddHebebrand:
## Root cause

The Windows Remote Desktop startup probe aborts healthy sessions on an **idle desktop**.

`StartSession` (`agent/internal/remote/desktop/session_webrtc.go:268`) creates a fresh DXGI duplication and immediately runs:

```go
probeCapture(capturer.Capture, 5, 200*time.Millisecond)
```

DXGI's `AcquireNextFrame` only delivers a frame when desktop content **changes**. On a static desktop it returns `DXGI_ERROR_WAIT_TIMEOUT`, which `dxgi_capture_windows.go` correctly maps to "no frame yet" — `(nil, nil)`. The probe treats a persistent `(nil, nil)` as proof the desktop is uncapturable. Five attempts × (50 ms acquire timeout + 200 ms sleep) is ~1.25 s of budget; if nothing repaints in that window the session dies before an SDP answer is ever created, with:

```
screen capture produced no frame after 5 attempts (session may have no capturable desktop)
```

TURN, ICE, HAProxy and Caddy are never reached — which matches the report exactly.

### The asymmetry

The **streaming loop already knows about this failure mode.** `session_capture.go` calls `forceDesktopRepaint()` in **eleven** places (thirteen repo-wide, counting `session.go` and `session_control.go`) — post-monitor-switch, startup warm-up, secure-desktop, idle cached-frame resend, GPU-encode-error fallback. Its own comment says it out loud:

> Critical for static displays where DXGI produces zero dirty rects naturally.

The **startup probe called it zero times.** That's the bug. (The maintainer's triage comment on the issue said "five places" — the distinct *scenarios*; the literal call-site count is eleven.)

This also explains why the reporter's step 7/8 — wake the display, generate activity, retry — still failed: by the time the new session's duplication is created and probed, the desktop has gone static again.

## The fix

`probeCapture` takes a `repaint func()` and invokes it **immediately before every capture attempt, including the first**. Repaint-before-capture (not after) is deliberate: `RedrawWindow` with `RDW_UPDATENOW` processes `WM_PAINT` synchronously, so the dirty rects are already queued by the time `AcquireNextFrame` runs.

On Windows the callback comes from a new `newProbeRepainter()`, which **composes the existing helpers rather than inventing a mechanism** — it runs the same pair the capture loop's static-display warm-up uses:

```go
func newProbeRepainter() (repaint func(), release func()) {
	runtime.LockOSThread()
	switchThreadToInputDesktop()
	return func() {
		nudgeSecureDesktop()
		forceDesktopRepaint()
	}, runtime.UnlockOSThread
}
```

Both halves of the repaint are needed, per the existing comment in `session_capture.go`: `InvalidateRect` has no HWND to target on a windowless desktop, and the `SendInput` jiggle is what makes the DWM compositor emit dirty rects there. The `+1/-1` relative move nets to zero and the streaming loop already injects it during the first ~2 s of every session.

The `LockOSThread` + `switchThreadToInputDesktop()` prologue is load-bearing: unlike the capture loop, the probe does **not** run on a `prepareCaptureThread()` goroutine — `newPlatformCapturer` releases its OS-thread lock once `initDXGI` succeeds, so Go may have migrated `StartSession` onto a thread that isn't on the input desktop, and `SendInput` / `InvalidateRect` / `RedrawWindow` all act on the *calling thread's* desktop.

It is hoisted to **once per probe, not per attempt** (second commit, a self-review catch): `switchThreadToInputDesktop` deliberately leaves its `HDESK` open — closing it would detach the thread (`dxgi_desktop_windows.go:322`) — so calling it per attempt would have leaked five desktop handles per session. `StartSession` calls `release()` unconditionally right after `probeCapture`, **before** the error check, so the pin never outlives the probe on either path. On macOS/Linux `newProbeRepainter` returns two no-ops alongside the existing stubs.

### Explicitly unchanged

- **Retry ceiling** — still `5 × 200ms`.
- **Genuine-failure path** — a hard capture error (device lost, no output, access denied) still aborts on the **first** attempt with the underlying error. This removes a false negative on idle desktops; it does not disable the probe.

The exhausted-attempts message now reads `...after %d attempts **with forced desktop repaints** (session may have no capturable desktop)`, so field logs distinguish a pre-fix from a post-fix agent at a glance.

## Test coverage

`TestProbeCaptureRepaintSequence` is table-driven and asserts on a recorded `repaint`/`capture` **event sequence** — a `capture` not immediately preceded by a `repaint` is the bug. Cases:

| Case | Asserts |
|---|---|
| frame on first attempt | repaint still fires first (`[repaint capture]`) |
| idle desktop, frame on 3rd try | `[repaint capture ×3]` — one repaint per retry |
| genuinely dead desktop | one repaint per attempt, **ceiling still 3/5** |
| hard capture error | aborts after `[repaint capture]` — no extra attempts |

Plus `TestProbeCaptureNilRepaintIsSafe` for the nil-callback shape. Verified non-vacuous by mutation: deleting the `repaint()` call reds all four sub-cases (`event sequence = [capture], want [repaint capture]`).

**Field verification only** (not unit-testable — Win32 syscalls with no seam): that `forceProbeRepaint` actually shakes a DXGI frame loose on real hardware, and that the `switchThreadToInputDesktop` prologue matters for a SYSTEM-spawned helper. The reporter's repro is the verification.

## Verification

```
$ cd agent && go test -race ./...
exit=0 — 69 packages ok, 5 with no test files, 0 FAIL, 0 data races
ok  github.com/breeze-rmm/agent/internal/remote/desktop  1.729s   (-count=1)
```

`gofmt -l` clean; `go vet` clean on darwin; `GOOS=windows go build ./...` and `go vet` clean (the `possible misuse of unsafe.Pointer` notes are pre-existing in untouched `mft_*`/`dxgi_*`/`encoder_*` files).

Refs #3951

### DIFF
--- a/agent/internal/remote/desktop/desktop_repaint_other.go
+++ b/agent/internal/remote/desktop/desktop_repaint_other.go
@@ -8,3 +8,8 @@ func forceDesktopRepaint() {}
 
 // nudgeSecureDesktop is a no-op on non-Windows platforms.
 func nudgeSecureDesktop() {}
+
+// forceProbeRepaint is a no-op on non-Windows platforms.
+// macOS CGDisplayStream / Linux X11 capture return a frame for an idle desktop,
+// so the startup probe has nothing to shake loose.
+func forceProbeRepaint() {}
--- a/agent/internal/remote/desktop/desktop_repaint_other.go
+++ b/agent/internal/remote/desktop/desktop_repaint_other.go
@@ -9,7 +9,9 @@ func forceDesktopRepaint() {}
 // nudgeSecureDesktop is a no-op on non-Windows platforms.
 func nudgeSecureDesktop() {}
 
-// forceProbeRepaint is a no-op on non-Windows platforms.
-// macOS CGDisplayStream / Linux X11 capture return a frame for an idle desktop,
-// so the startup probe has nothing to shake loose.
-func forceProbeRepaint() {}
+// newProbeRepainter is a no-op on non-Windows platforms: macOS CGDisplayStream
+// and Linux X11 capture return a frame for an idle desktop, so the startup
+// probe has nothing to shake loose and no thread desktop to attach to.
+func newProbeRepainter() (repaint func(), release func()) {
+	return func() {}, func() {}
+}
--- a/agent/internal/remote/desktop/desktop_repaint_windows.go
+++ b/agent/internal/remote/desktop/desktop_repaint_windows.go
@@ -4,6 +4,7 @@ package desktop
 
 import (
 	"log/slog"
+	"runtime"
 	"unsafe"
 )
 
@@ -59,4 +60,25 @@ func nudgeSecureDesktop() {
 	if ret != uintptr(len(nudges)) {
 		slog.Debug("nudgeSecureDesktop: SendInput did not inject full nudge sequence", "sent", ret)
 	}
+}
+
+// forceProbeRepaint shakes a frame out of an idle desktop for the startup
+// capture probe (see probeCapture). It runs the same warm-up pair the capture
+// loop uses on a static display — nudgeSecureDesktop followed by
+// forceDesktopRepaint (session_capture.go). Both are needed: InvalidateRect has
+// no HWND to target on a windowless desktop, and the SendInput jiggle is what
+// makes the DWM compositor produce dirty rects there.
+//
+// It pins and re-attaches the calling thread to the input desktop first.
+// Unlike the capture loop, the probe does not run on a prepareCaptureThread
+// goroutine: newPlatformCapturer releases its OS-thread lock once DXGI is
+// initialized, so Go may have migrated StartSession onto a thread that is not
+// on the input desktop — and SendInput, InvalidateRect and RedrawWindow all
+// act on the calling thread's desktop.
+func forceProbeRepaint() {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+	switchThreadToInputDesktop()
+	nudgeSecureDesktop()
+	forceDesktopRepaint()
 }
--- a/agent/internal/remote/desktop/desktop_repaint_windows.go
+++ b/agent/internal/remote/desktop/desktop_repaint_windows.go
@@ -62,23 +62,30 @@ func nudgeSecureDesktop() {
 	}
 }
 
-// forceProbeRepaint shakes a frame out of an idle desktop for the startup
-// capture probe (see probeCapture). It runs the same warm-up pair the capture
-// loop uses on a static display — nudgeSecureDesktop followed by
-// forceDesktopRepaint (session_capture.go). Both are needed: InvalidateRect has
-// no HWND to target on a windowless desktop, and the SendInput jiggle is what
-// makes the DWM compositor produce dirty rects there.
+// newProbeRepainter returns the repaint callback for the startup capture probe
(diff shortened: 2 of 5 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (full-rigor pass, agent code shipping to customer Windows machines) + `silent-failure-hunter`. Both re-read the branch at `df1e7d901` after a mid-review push.

**Findings:** 0 at reportable confidence from either reviewer. Three items surfaced and were handled:

- **Desktop-handle leak (found in self-review, fixed pre-review-completion in `df1e7d901`)** — `switchThreadToInputDesktop` deliberately leaves its `HDESK` open, since closing it would detach the thread (`dxgi_desktop_windows.go:322`). The first cut called it inside the per-attempt repaint, leaking five desktop handles per session and re-returning a desktop-mutated OS thread to Go's scheduler pool five times during the highest-goroutine-churn moment of session start. Replaced `forceProbeRepaint()` with `newProbeRepainter() (repaint, release)`: lock + attach **once** around the whole probe, `release()` called unconditionally right after `probeCapture` and **before** the error check. `code-reviewer` independently reached the same conclusion and confirmed the fix.
- **Stale comment** — `probe_test.go` still named `forceProbeRepaint` after the rename. Fixed in `494c76c79`.
- **Noted, not actioned (reviewer said don't block):** the repaint closure no longer re-attaches per attempt, so if the input desktop changes *mid-probe* (a UAC dialog dismissing between attempt 1 and 3), later nudges target the desktop attached at probe start. Deliberate trade-off against the handle leak; `Capture()` still runs its own `checkDesktopSwitch` / GDI-fallback recovery independently.

Reviewers explicitly confirmed: the retry ceiling and the hard-error-aborts-immediately path are preserved; repaint-before-capture is the correct ordering given `RDW_UPDATENOW`'s synchronous `WM_PAINT` vs. `AcquireNextFrame`'s 50 ms poll; no raw `err` reaches `slog` (would serialize as `{}`); the pre-answer `nudgeSecureDesktop()` crosses no consent boundary — the identical jiggle already fires unconditionally a few hundred ms later in the streaming loop's warm-up.

**Tests:** `cd agent && go test -race ./...` → **exit 0, 69 packages ok, 5 with no test files, 0 FAIL, 0 data races.** `go test -race -count=1 ./internal/remote/desktop/...` green uncached. `gofmt -l` clean; `go vet` clean on darwin; `GOOS=windows go build ./...` clean (the `possible misuse of unsafe.Pointer` vet notes are pre-existing in untouched `mft_*`/`dxgi_*`/`encoder_*` files).

New `TestProbeCaptureRepaintSequence` asserts the recorded `repaint`/`capture` **event interleaving**, not call counts. Verified non-vacuous by mutation — deleting the `repaint()` call reds all four sub-cases (`event sequence = [capture], want [repaint capture]`).

**Field-verification only** (Win32 syscalls, no unit-test seam): that the repaint actually shakes a DXGI frame loose on real hardware, and that the `switchThreadToInputDesktop` prologue matters for a SYSTEM-spawned helper. The reporter's repro on the affected endpoint is the verification.

**Status:** review-clean, awaiting maintainer merge. Body says `Refs #3951`, not `Closes` — @sxwxrslvvt verifies and closes.

Issue #3932, opened 2026-08-24
Windows File Browser delete fails because nested paths are treated as top-level paths


sxwxrslvvt:
### Description

On Windows, deleting a normal file from **Remote Tools → File Browser** fails consistently.

Browse, upload and download work correctly, and the same file can be deleted from Breeze Remote Terminal running as `NT AUTHORITY\SYSTEM`, so this does not appear to be a filesystem permission or service account issue.

I traced the failure through the API and agent logs. The delete request reaches the API, a `FILE_DELETE` command is created, and the Windows agent receives it, but the agent rejects the path with:

```text
recursive delete denied on top-level path:
C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt
```

From looking at the v0.107.0 agent code, it looks like the recursive delete safety check splits the cleaned path only on `/`. On Windows, `filepath.Clean()` returns backslash-separated paths, so the full `C:\...` path appears to be treated as a single path component.

This makes ordinary nested Windows paths look like a top-level path whenever the delete command is sent with `recursive: true`.

### Steps to Reproduce

1. Enroll a Windows device in Breeze.
2. Create a normal test file, for example:

```powershell
New-Item -ItemType Directory -Path 'C:\ProgramData\SOTIKS\BreezePilot' -Force
'DELETE TEST' | Set-Content 'C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt'
```

3. Open **Remote Tools → File Browser**.
4. Browse to:

```text
C:\ProgramData\SOTIKS\BreezePilot
```

5. Select `delete-test.txt`.

6. Click **Delete** and try either:

   * Move to Trash
   * Delete Permanently

7. The file is not deleted.

For comparison, open **Remote Tools → Terminal** on the same device and run:

```powershell
Remove-Item 'C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt' -Force -ErrorAction Stop
```

The file is deleted successfully from the terminal.

### Expected Behavior

A normal file selected in File Browser should be moved to Trash or permanently deleted.

The delete safety checks should still prevent dangerous operations such as deleting a drive root or other protected top-level directories.

### Actual Behavior

The File Browser delete operation fails.

The API successfully creates the file delete command, but the Windows agent rejects the normal nested path as if it were a top-level path.

Agent result:

```text
recursive delete denied on top-level path:
C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt
```

The API then returns a failed result.

[...]

PR #3961, opened 2026-08-24
fix(agent): volume-aware recursive-delete boundary so Windows nested paths delete (#3932)

ToddHebebrand:
Refs #3932

## Root cause

`DeleteFile` (`agent/internal/remote/tools/fileops.go:631-637`) gated recursive deletes with:

```go
parts := strings.Split(strings.TrimPrefix(cleanPath, "/"), "/")
if len(parts) <= 1 { /* recursive delete denied on top-level path */ }
```

`filepath.Clean` returns **backslash**-separated paths on Windows, so `C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt` survives the split as a **single** component and trips a check written for `/home`-style paths. Every Windows File Browser delete sent with `recursive: true` was refused regardless of depth — exactly as @sxwxrslvvt traced. Confirmed still present on current `main` before this change.

## The fix

Replace the slash split with `isRecursiveDeleteBoundary` (new `fileops_delete_boundary.go`): strip the volume/root specifier, then count the components **below** it.

A drive specifier (`C:`, `C:\`), a UNC share root (`\\server\share`), an extended-length prefix (`\\?\C:\`, `\\?\UNC\server\share`) and a POSIX `/` all contribute **zero** depth.

## Boundary semantics — the judgement call

**Depth `<= 1` is refused.** That is the existing POSIX rule (`/` and `/home` refused, `/home/user` allowed) carried across to Windows unchanged:

| path | before | after |
|---|---|---|
| `C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt` | **denied (bug)** | allowed |
| `C:\Temp\build`, `C:/Temp/build`, `C:\Temp/build\out` | denied (bug) | allowed |
| `C:\`, `C:/`, `C:` | denied | denied |
| `C:\Windows`, `C:\Users`, `C:\ProgramData` | denied | denied |
| `\\server\share`, `\\server\share\sub` | denied | denied |
| `\\server\share\sub\data` | denied (bug) | allowed |
| `/`, `/home`, `/var` | denied | denied |
| `/home/user`, `/home/user/projects/x` | allowed | allowed |

This is **stricter than the direction sketched in the issue comment** (\"`C:\Temp` recursive delete stays possible\"), and that divergence is deliberate — worth a maintainer sanity-check. `deniedSystemPaths` (`fileops.go:34`) is entirely POSIX (`/`, `/boot`, `/proc`, `/sys`, `/dev`, `/bin`, `/sbin`, `/usr`) and has **no Windows entries at all**, so on Windows this guard is the only thing between a recursive delete and `C:\Windows`, `C:\Users` or `C:\ProgramData`. Refusing only the bare volume root would have made all three recursively deletable while `/usr` stayed refused on Linux. The reported scenario is unaffected either way — the reporter's path is four levels deep.

### Three shapes fail closed rather than being counted

- **Unrecognised device namespaces** — `\\?\GLOBALROOT\Device\HarddiskVolume1\Windows`, `\\?\Volume{GUID}\Windows`, `\\.\PhysicalDrive0`, `\??\` without a drive or `UNC\` marker. Their components are object-manager nodes, not filesystem depth, so counting them would make a volume-root-adjacent target look arbitrarily deep. `\\?\C:\…`, `\\.\C:\…`, `\??\C:\…` and `\\?\UNC\server\share\…` are recognised and parsed normally.
- **Unresolved `..`** — real depth depends on the agent's working directory (`..\Windows` counts two components but can resolve to a volume-root child). Refused rather than guessed. This is a *tightening* vs. the old rule, which allowed `../etc`; the file browser always sends absolute paths.
- **Backslash on POSIX** — still an ordinary, legal filename character, so `/data\evil` stays **one** top-level component there. Normalising separators unconditionally would have over-counted depth and turned a currently-refused top-level directory into an allowed recursive delete.

## Tests

`agent/internal/remote/tools/fileops_delete_boundary_test.go` — 55 table-driven cases across two suites plus three call-site tests.

The platform grammar is a **parameter** (`isRecursiveDeleteBoundaryFor(path, windows bool)`), not `runtime.GOOS`, so the Windows table actually executes on Linux CI and macOS dev machines instead of being GOOS-skipped everywhere the agent is built. The POSIX suite pins the *original* guard's verdicts, so the fix is provably a no-op off Windows.

```
$ cd agent && go build ./... && go test -race ./...
EXIT=0 — 69 packages ok, 0 FAIL, 0 data races (5 packages have no test files)

$ go test -race -v -run DeleteBoundary ./internal/remote/tools/
58 PASS lines (55 subtests + 3 parents)
```

## Not in scope

The reporter also noted the delete failure is only surfaced in the File Browser **Activity** panel, which is closed by default, so a failed delete looks like the button did nothing. That is a real UX gap but a separate web change — flagging it rather than bundling it here.

### DIFF
--- a/agent/internal/remote/tools/fileops.go
+++ b/agent/internal/remote/tools/fileops.go
@@ -628,12 +628,13 @@ func DeleteFile(payload map[string]any) CommandResult {
 		return NewErrorResult(err, time.Since(start).Milliseconds())
 	}
 
-	// Block recursive deletes on any top-level directory (e.g. /home, /var, /opt)
-	if recursive {
-		parts := strings.Split(strings.TrimPrefix(cleanPath, "/"), "/")
-		if len(parts) <= 1 {
-			return NewErrorResult(fmt.Errorf("recursive delete denied on top-level path: %s", cleanPath), time.Since(start).Milliseconds())
-		}
+	// Block recursive deletes on any filesystem root or top-level directory
+	// (e.g. /, /home, /var, C:\, C:\Windows, \\server\share). See
+	// isRecursiveDeleteBoundary (fileops_delete_boundary.go) for the boundary
+	// semantics and why a plain slash-separated component count was wrong on
+	// Windows (#3932).
+	if recursive && isRecursiveDeleteBoundary(cleanPath) {
+		return NewErrorResult(fmt.Errorf("recursive delete denied on top-level path: %s", cleanPath), time.Since(start).Milliseconds())
 	}
 
 	// Check if path exists
--- a/agent/internal/remote/tools/fileops_delete_boundary.go
+++ b/agent/internal/remote/tools/fileops_delete_boundary.go
@@ -70,6 +70,18 @@ func isRecursiveDeleteBoundaryFor(cleanPath string, windows bool) bool {
 			// \\server\share\... — the host and share are the volume.
 			volumeComponents = 2
 		}
+
+		// Past the volume specifier a colon is not legal in a Windows path
+		// component, and it is not a separator either — so it would survive as a
+		// component of its own and inflate the depth. "C::\Windows" would
+		// otherwise strip "C:", split ":\Windows" into [":", "Windows"] and read
+		// as depth 2, allowing a recursive delete of a path one stray colon away
+		// from a volume-root child. Alternate-data-stream suffixes
+		// ("C:\Temp\x:stream") land here too and are equally refused: the file
+		// browser has no reason to address either shape.
+		if strings.ContainsRune(rest, ':') {
+			return true
+		}
 	}
 
 	depth := 0
--- a/agent/internal/remote/tools/fileops_delete_boundary.go
+++ b/agent/internal/remote/tools/fileops_delete_boundary.go
@@ -1,0 +1,142 @@
+package tools
+
+import (
+	"runtime"
+	"strings"
+)
+
+// isRecursiveDeleteBoundary reports whether cleanPath sits at — or immediately
+// below — the root of its filesystem, and so must not be recursively deleted.
+//
+// #3932: this replaces `strings.Split(strings.TrimPrefix(p, "/"), "/")` plus a
+// `len(parts) <= 1` test. That form is correct only for POSIX paths:
+// filepath.Clean returns BACKSLASH-separated paths on Windows, so every Windows
+// path collapsed into a single component and the guard denied every recursive
+// delete the file browser issued, however deeply nested — the reporter's
+// `C:\ProgramData\SOTIKS\BreezePilot\delete-test.txt` among them. The failure
+// mode being fixed is a false DENY, so nothing here may turn into a false ALLOW
+// near a root.
+//
+// The boundary is expressed as depth below the filesystem root rather than as a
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — code-reviewer (adversarial false-ALLOW hunt + differential fuzz of ~4,765 `filepath.Clean`-reachable POSIX shapes against the *old* rule) and pr-test-analyzer (vacuousness + branch coverage). Plus a `codex exec` read-only second opinion on the boundary semantics before implementation.

**Findings round 1:** 1 real — a **false ALLOW**. `hasDriveSpecifier` accepted `C:` without checking what followed, so `C::\Windows` stripped the drive, left `:\Windows`, and the colon survived as its own component (it is not a separator) → depth 2 → allowed. One stray colon away from a volume-root child. Fixed in 342fb63 by failing closed on any colon past the volume specifier (Windows branch only — a colon is a legal POSIX filename character, and `/home/user/back:up` must stay deletable). Same commit adds the `\Go\src\pkg` case that was the one uncovered line in `deviceNamespacePrefixLen`.

Codex also caught three false-ALLOW vectors *before* the first commit, which is why they are handled in the original implementation rather than as review fixes: `\??\`, `\\?\GLOBALROOT\Device\…` / `\\?\Volume{GUID}\…` (namespace nodes counted as filesystem depth), and a leading `..` (depth depends on the agent's working directory).

**Findings round 2** (re-reviewed because the fix touched the safety surface itself): **0 findings.** Confirmed the colon check cannot be bypassed — the only early returns before it are `true` (deny) — and confirmed each new case genuinely flips verdict when the fix is removed, i.e. non-vacuous. One nit declined: the `/back:up` POSIX case would pass without the fix too (it is denied by the depth rule anyway); its sibling `/home/user/back:up` is the one that actually pins "no leak into POSIX".

**Tests:** `cd agent && go build ./... && go test -race ./...` → exit 0, 69 packages ok, 0 FAIL, 0 data races. `go test -race -v -run DeleteBoundary` → 65 PASS lines (62 subtests + 3 parents), 17 of them **allow** assertions, so an always-deny implementation fails immediately and the pre-fix implementation fails all 13 Windows-allow cases. `go vet` and `gofmt` clean on the touched files. No TypeScript changed, so no `tsc` run.

**CI:** 54/54 green — including **Test Agent (Windows)**, so the guard is exercised on a real Windows runner and not only through the parameterized table.

**Two things worth a maintainer sanity-check before merge:**

1. **The depth boundary is `<= 1`, not `<= 0`** — stricter than the `C:\Temp` example in the issue comment. `C:\Windows`, `C:\Users` and `C:\ProgramData` stay refused because `deniedSystemPaths` has no Windows entries at all, making this guard the sole protection there. Rationale and the full before/after table are in the PR description. The reported scenario is unaffected either way (the reporter's path is four levels deep) — but if you want `C:\Temp` recursively deletable, that is a one-character change and a handful of table rows.
2. **Alternate data streams now fail closed** (`C:\Temp\x:stream`), a side effect of the colon rule. Believed to have no legitimate File Browser use.

PR #3962, opened 2026-08-24
fix(api): count partner-level ring assignments via organizations.partner_id (#3954)

### LINKED ISSUE #3954: Update ring DEVICES count shows 0 for partner-level policy assignments (targetId queried as org_id)
**Symptom:** an update ring linked from a configuration policy assigned at partner level ("all organizations") shows **DEVICES 0** and COMPLIANCE "—" in the Update Rings list, even though the ring is genuinely applied to devices. Reported in the wild via discussion #2598 (screenshot of the rings list with 0 devices while the policy is visibly assigned to the device).

**Root cause:** `apps/api/src/routes/updateRingsHelpers.ts` (`resolveRingAssignedDeviceIds`) lumps partner-level assignments in with org-level:

```ts
else if (a.level === 'organization' || a.level === 'partner') orgIds.push(a.targetId);
```

and then queries `devices.orgId IN (orgIds)`. A partner-level assignment's `targetId` is a **partner** id, so it matches zero devices — the standard org/partner fan-out failure class (see CLAUDE.md partner-wide step 5). The helper predates partner-level assignments (#1724/#2280) and was never updated.

**Blast radius:** display-only. The patch scheduler resolves devices through its own correct partner branch (`apps/api/src/jobs/patchSchedulerWorker.ts:210-232`), so scheduling honors the ring; only the count/compliance surface lies. But the lie is damaging: it reads as "your ring is not being used," which is exactly how the #2598 reporter interpreted it.

**Fix:** split `partner` into its own bucket and expand via the devices-join on the org's `partner_id` (orgs of that partner → devices), mirroring the scheduler's partner branch, with the same `isEphemeral=false` exclusion.

ToddHebebrand:
## Root cause

`resolveRingAssignedDeviceIds` (`apps/api/src/routes/updateRingsHelpers.ts`) treated a partner-level config-policy assignment's `targetId` as if it were an org id:

```ts
else if (a.level === 'organization' || a.level === 'partner') orgIds.push(a.targetId);
// ... then
.where(and(inArray(devices.orgId, orgIds), eq(devices.isEphemeral, false)))
```

`config_policy_assignments.target_id` is **polymorphic** — its referent depends on `level`. For `level='partner'` it is a **partner** id, so `devices.org_id IN (<partner uuid>)` matches nothing. The Update Rings list rendered **DEVICES 0** / COMPLIANCE "—" for a ring that was genuinely applied (reported via discussion #2598).

The helper predates partner-level assignments (#1724/#2280) and was never updated. This is the standard org/partner fan-out failure class called out in `CLAUDE.md` ("Partner-Wide First", step 5).

**Blast radius: display-only.** `patchSchedulerWorker.resolveDeviceIdsForAssignment` has its own correct partner branch, so scheduling always honored the ring. Only the count/compliance surface lied — but the lie reads as "your ring is not being used", which is exactly how the reporter interpreted it.

## The fix

**1. Partner-level assignments get their own bucket**, expanded through the device's org `partner_id` rather than `devices.org_id`:

```ts
.from(devices)
.innerJoin(organizations, eq(devices.orgId, organizations.id))
.where(and(or(...partnerConditions), eq(devices.isEphemeral, false)))
```

Faithful to the scheduler on the details: the **`policyOrgId` clamp** (a legacy *org-owned* policy carrying a partner-level assignment still clamps to its own org, so the count can never overstate what gets patched — the DB integrity trigger permits that shape, so it is reachable), and **`isEphemeral = false`** so Quick Support machines stay out.

**2. A uniform ring-partner clamp on every level** (added in response to review — see below). Every branch now inner-joins `organizations` and applies `eq(organizations.partnerId, ringPartnerId)`.

**Drive-by:** the `device_group` branch now inner-joins `devices` so the ephemeral exclusion applies there too (previously it read memberships alone), and the `device` branch now verifies the target exists under the ring's partner instead of adding the assignment row's id blindly. Both match the scheduler.

### Why clamp to the *ring's* partner rather than thread the scheduler's two owner fields

The scheduler threads `policyOrgId` + `policyPartnerId` and re-clamps per level. Clamping to the ring's partner is **one value applied uniformly to all five levels**, and is equivalent rather than weaker: `2026-07-27-a-feature-policy-reference-ownership.sql` constrains a `feature_type='patch'` link so the referenced `patch_policies.partner_id` **equals** `COALESCE(policy.partner_id, policy_org.partner_id)`. So "org is under the ring's partner" and "org is under the policy's partner" are the same predicate here.

The clamp is unconditional SQL rather than an RLS assumption, which is what makes it scope-independent — see the review finding.

## Review finding (round 1) — a real bug, fixed in 396f0a6

The first review round flagged that the subset branches (device / device_group / site / organization) had **no partner clamp at all** and leaned on the caller's RLS context to exclude a target org reparented to another partner. That argument holds for `scope='partner'` but **not for `scope='system'`**: `serializeAccessibleIds` treats a null `accessibleOrgIds` as unrestricted, so a system-scope caller of either endpoint got no org clamp whatsoever and could see another partner's devices counted toward the ring.

I reproduced it against real Postgres before fixing — forging the reparented state and resolving under system scope returned **1** where it must return **0** — then fixed it with the uniform clamp above and added a regression test that forges the same state.

Worth noting the DB *does* independently guard the reparent path (`a_config_policy_assignment_target_update` on `organizations` re-runs `breeze_validate_config_policy_assignment_target` and raises on a partner mismatch), so this was defense-in-depth rather than a live leak. The clamp removes the dependency on that reasoning entirely.

## Call-site sweep

Swept every reader of `config_policy_assignments` that resolves `level`/`targetId` to devices.

**Fixed here:** `apps/api/src/routes/updateRingsHelpers.ts`. Both consumers — `resolveRingDeviceCounts` (rings list) and `resolveRingDeviceIds` (ring compliance) — go through it, so both surfaces are fixed by the one change.

**Already correct (verified, left alone):** `services/featureConfigResolver.ts` (`resolveDeviceIdsForSoftwarePolicy`, `resolveAllVulnerabilityEnabledDevices`) · `jobs/automationWorker.ts` · `services/policyEvaluationService.ts` · `jobs/patchSchedulerWorker.ts` (the reference) · `routes/partnerApi/configuration.ts` (`assignmentEffectiveOrgSql`, pre-scoped to the principal's partner). Device→policy-direction resolvers are correct by construction (they derive the device's org's `partnerId` and match `level='partner'` rows): `featureConfigResolver.loadDeviceHierarchy`, `configurationPolicy.resolveEffectiveConfigWithExecutor`, `helperPermissions.resolveHelperPermissionLevelForDevice`, and four resolvers in `routes/agents/helpers.ts`. `services/patchJobService.ts` `scanScheduledPatchSettings` is dead code (`FIXME: not yet wired to a worker`, zero callers). CRUD/validation/display only: `routes/configurationPolicies/assignments.ts`, `configurationPolicy.listAssignments`/`validateAssignmentTarget`, `services/aiToolsConfigPolicy.ts`, `routes/softwareInventory.ts`, `scripts/migrateToConfigPolicies.ts`.

**Found but NOT fixed here — two adjacent bugs, verified firsthand, different subsystems:**

1. **`services/warrantyAlertEvaluator.ts:71`** — `targetIds = [deviceId, ...groupIds, device.siteId, device.orgId]` never includes the device's `organizations.partner_id`, yet the comment directly above ranks priority as `device > device_group > site > organization > partner`. A partner-wide warranty policy is never queried, so it silently falls through to `DISABLED_SETTINGS`.
2. **`routes/agents/helpers.ts:1961`** (`getOrgEventLogRetentionDays`, driven per-org by `jobs/eventLogRetention.ts`) — filters `eq(configPolicyAssignments.level, 'organization')` with no partner branch, so a partner-level event-log retention policy is ignored and every org silently falls back to the 30-day default.

Different mechanism from this bug (partner rows never queried, rather than a mis-typed id) but the same customer-visible symptom: partner-wide config silently doesn't apply. Out of scope for a patching display fix — happy to file these as separate issues.

## Tests

Every assertion was verified to **fail against the code it guards** — none are vacuous:

- `apps/api/src/routes/updateRingsHelpers.partnerFanout.test.ts` (7 tests) — renders the real condition through `PgDialect` and pins the predicate text. Mutating `organizations.partnerId` back to `devices.orgId` turns **4 of 7 red**. This matters because the repo's usual chainable Drizzle mock never inspects the argument to `.where()` and would pass against both the broken and the fixed code.
- `apps/api/src/__tests__/integration/updateRingPartnerDeviceCount.integration.test.ts` (6 tests) — real Postgres, unprivileged `breeze_app`. Covers the cross-org count, ephemeral exclusion, cross-partner isolation, the legacy org-owned clamp, and the system-scope clamp. Restoring the original bucketing yields `expected +0 to be 2` (the reported symptom); removing the ring-partner clamp yields `expected 1 to be +0`.

Local runs on a private test-stack: **90 unit passed** (6 files, incl. `patchSchedulerWorker.test.ts`), **20 integration passed** (4 files), `tsc --noEmit` clean, `eslint` clean.

No new table, column, or migration — the cascade / export-policy / RLS-allowlist registration steps in `CLAUDE.md` do not apply.

Closes #3954

### DIFF
--- a/apps/api/src/__tests__/integration/updateRingPartnerDeviceCount.integration.test.ts
+++ b/apps/api/src/__tests__/integration/updateRingPartnerDeviceCount.integration.test.ts
@@ -31,7 +31,8 @@
 import './setup';
 import { describe, expect, it } from 'vitest';
 import { randomUUID } from 'node:crypto';
-import { inArray } from 'drizzle-orm';
+import { inArray, sql } from 'drizzle-orm';
+import { getTestDb } from './setup';
 import { db, withDbAccessContext, withSystemDbAccessContext, type DbAccessContext } from '../../db';
 import {
   configurationPolicies,
@@ -241,6 +242,47 @@ describe('update-ring device count — partner-level assignment fan-out (#3954)'
 
     // orgA1's single non-ephemeral device only — NOT orgA2's.
     expect(counts.get(f.ringA.id)).toBe(1);
+  });
+
+  runDb('SYSTEM scope: a target org under another partner is clamped out', async () => {
+    // The ring-partner clamp is defense-in-depth against a subset assignment
+    // whose target org has been reparented to a different partner. Normally the
+    // DB forbids reaching that state (config_policy_assignment_target_integrity
+    // rejects the write, and a_config_policy_assignment_target_update rejects
+    // the reparent), so we FORGE it with triggers disabled — the same idiom the
+    // cross-tenant forge suites use.
+    //
+    // This runs under SYSTEM scope deliberately: accessible_org_ids is
+    // unrestricted there, so RLS contributes NO org clamp. If this passes, the
+    // SQL clamp is doing the work, not the caller's RLS context.
+    const f = await seedFixture();
+    const policyId = await linkPolicyToRing({
+      ringId: f.ringA.id,
+      partnerId: f.partnerA.id,
+      level: 'organization',
+      targetId: f.orgA2.id, // legal: orgA2 belongs to partner A
+    });
+
+    const admin = getTestDb();
+    // DISABLE TRIGGER USER rather than naming one: the integrity trigger has
+    // already been renamed once (2026-07-29-serialize-config-policy-assignment-integrity.sql
+    // split it into a_config_policy_assignment_integrity_{insert,update,delete}),
+    // and a name-specific ALTER would silently rot into a 42704 on the next rename.
+    await admin.execute(sql`ALTER TABLE config_policy_assignments DISABLE TRIGGER USER`);
+    try {
+      // Repoint at partner B's org — the state a reparent would have produced.
+      await admin.execute(
+        sql`UPDATE config_policy_assignments SET target_id = ${f.orgB1.id}
+            WHERE config_policy_id = ${policyId} AND level = 'organization'`
+      );
+    } finally {
+      await admin.execute(sql`ALTER TABLE config_policy_assignments ENABLE TRIGGER USER`);
+    }
+
+    const counts = await withSystemDbAccessContext(() => resolveRingDeviceCounts([f.ringA.id]));
+
+    // partnerB's device must NOT be attributed to partnerA's ring.
+    expect(counts.get(f.ringA.id) ?? 0).toBe(0);
   });
 
   runDb('non-vacuity probe: system scope sees the seeded partner devices', async () => {
--- a/apps/api/src/__tests__/integration/updateRingPartnerDeviceCount.integration.test.ts
+++ b/apps/api/src/__tests__/integration/updateRingPartnerDeviceCount.integration.test.ts
@@ -1,0 +1,259 @@
+/**
+ * Update-ring DEVICES count for PARTNER-level policy assignments (#3954).
+ *
+ * `resolveRingAssignedDeviceIds` (apps/api/src/routes/updateRingsHelpers.ts)
+ * used to lump partner-level assignments in with org-level ones:
+ *
+ *     else if (a.level === 'organization' || a.level === 'partner') orgIds.push(a.targetId);
+ *     ... where(inArray(devices.orgId, orgIds))
+ *
+ * A partner-level assignment's `targetId` is a PARTNER id, so that matched zero
+ * rows and the Update Rings list rendered "DEVICES 0" for a ring that was
+ * genuinely applied (the patch scheduler has its own correct partner branch).
+ *
+ * WHY REAL POSTGRES: the failure was entirely inside a `.where()` predicate and
+ * depends on the devices→organizations join actually existing. A chainable
+ * Drizzle mock that ignores `.where()` passes against both the broken and the
+ * fixed code (see memory: vacuous_drizzle_where_clause_assertions). The
+ * compiled-SQL unit suite (updateRingsHelpers.partnerFanout.test.ts) pins the
+ * predicate text; this proves the resolved COUNT against real rows, under a
+ * real partner-scope RLS context on the unprivileged `breeze_app` role.
+ *
+ * Coverage:
+ *   - partner-wide policy + partner-level assignment counts devices across ALL
+ *     of the partner's orgs (the regression: was 0, must be 2)
+ *   - ephemeral Quick Support devices stay excluded from the partner fan-out
+ *   - another partner's devices are never counted (cross-partner isolation)
+ *   - a partner B context resolves 0 for partner A's ring (RLS hides the chain)
+ *   - a legacy ORG-owned policy carrying a partner-level assignment clamps to
+ *     its own org, mirroring patchSchedulerWorker.resolveDeviceIdsForAssignment
+ */
+import './setup';
+import { describe, expect, it } from 'vitest';
+import { randomUUID } from 'node:crypto';
+import { inArray } from 'drizzle-orm';
+import { db, withDbAccessContext, withSystemDbAccessContext, type DbAccessContext } from '../../db';
+import {
+  configurationPolicies,
+  configPolicyFeatureLinks,
+  configPolicyAssignments,
+  devices,
+  patchPolicies,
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` — two rounds. Round 1 on the full PR (correctness vs. the scheduler reference, RLS/tenancy, test non-vacuity, CLAUDE.md contract registration). Round 2 scoped to the round-1 fix commit, because that fix touched the tenancy surface.

**Findings:** Round 1 raised **1 MUST-FIX** (0 nice-to-have) → fixed in 396f0a6. Round 2: **0 findings**, all six checked items clean.

**What round 1 caught (a real bug, not a nit):** the subset branches (device / device_group / site / organization) had no partner clamp and leaned on the caller's RLS context to exclude a target org reparented to another partner. That holds for `scope='partner'` but **not `scope='system'`** — `serializeAccessibleIds` treats a null `accessibleOrgIds` as unrestricted, so a system-scope caller of either endpoint got no org clamp at all. I reproduced it against real Postgres before fixing (forged the reparented state, resolved under system scope, got **1** where **0** is required), fixed it with a uniform ring-partner clamp on all five levels, and added that forge as a regression test.

Round 2 independently verified the load-bearing claim behind that fix — that clamping to the *ring's* partner is equivalent to the scheduler's *policy*-partner clamp. `2026-07-27-a-feature-policy-reference-ownership.sql` is not just a one-time preflight: it installs live triggers (`config_policy_feature_links_reference_integrity`, plus AFTER UPDATE/DELETE triggers on `configuration_policies`, `organizations`, and `patch_policies`) routing through `breeze_validate_config_policy_feature_reference`, which raises 23503 on a partner mismatch. The invariant is DB-enforced on writes, so the equivalence holds.

**Tests — every assertion verified to fail against the code it guards (no vacuous mocks):**

| Suite | Result | Mutation check |
|---|---|---|
| `updateRingsHelpers.partnerFanout.test.ts` (compiled SQL via `PgDialect`) | 7 passed | reverting to `devices.org_id` → **4 of 7 red** |
| `updateRingPartnerDeviceCount.integration.test.ts` (real Postgres, `breeze_app`) | 6 passed | original bucketing → `expected +0 to be 2` (the reported symptom); clamp removed → `expected 1 to be +0` |
| `updateRings_*` + `patchSchedulerWorker.test.ts` | 90 passed (6 files) | — |
| related integration suites (`update-rings-partner-scope`, `patchSchedulerPartnerResolution`, `helperPolicyResolution`) | 20 passed (4 files) | — |

Local runs on a private `test-stack` (the shared `:5433` DB had drifted — it carried `2026-09-08-managed-by-agent.sql`, which this base does not). `tsc --noEmit` clean, `eslint` clean on all three files.

**CI:** Type Check, Lint, Check Migrations (both), Security Audit, CodeQL and ~35 others green; **0 failures**. `Test API` and `Test Web` were still running when I posted this — worth a glance before merging, though `Test API` runs the same vitest files I ran locally above.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3954 — both are your call.

**Two things for you to rule on:**
1. The sweep found **two adjacent bugs I did not fix** (different subsystems, same "partner-wide config silently doesn't apply" symptom): `services/warrantyAlertEvaluator.ts:71` never puts the device's partner id in `targetIds` despite its own comment ranking `partner` in the priority chain, and `routes/agents/helpers.ts:1961` (`getOrgEventLogRetentionDays`) filters `level='organization'` only, so partner-level retention silently falls back to 30 days. Both verified by reading the code. Want them filed as issues?
2. A pre-existing asymmetry this PR does not change: the display count is RLS-clamped to `active`/`trial` orgs and to a `orgAccess='selected'` technician's curated list, while the scheduler deliberately spans **all** the partner's orgs regardless of billing status. So for a suspended org, or a restricted tech, the count is a floor rather than an exact match. The common case (partner admin, active orgs) is now exact.

PR #3971, opened 2026-08-25
fix(web): render angle brackets and HTML entities in locale strings (#3964)

### LINKED ISSUE #3964: [UI] Angle brackets and HTML entities inside locale strings render literally — Enrollment Keys shows "&lt;key>", Suppress dialog shows "&ldquo;"
## Description

Translation values in `apps/web/src/locales/**` that contain markup-ish characters reach the user as **literal text** instead of the character they encode. Two variants, both confirmed in a real browser on `main` (`6ec30b06a`).

**Variant A — a literal `<key>` is parsed as a tag by `<Trans>`.**

`/settings/enrollment-keys` renders its subtitle as:

> Create and manage keys for agent enrollment. Use these keys with breeze-agent enroll **&lt;key>**

DOM proof:

```
innerText: "…breeze-agent enroll &lt;key>"
innerHTML: "…breeze-agent enroll &amp;lt;key&gt;"
```

The en value is plain text and is correct as authored:

```json
"description": "Create and manage keys for agent enrollment. Use these keys with breeze-agent enroll <key>"
```

It breaks at the render site, `apps/web/src/components/settings/EnrollmentKeyManager.tsx:497`:

```tsx
<Trans i18nKey="enrollmentKeys.description" t={t} components={{ code: <code className="…" /> }} />
```

`<Trans>` parses the value as markup, finds a `<key>` element with no entry in `components`, and escapes it. All 8 locales carry the same plain `<key>`, so every language is affected.

**There is a second bug on that same line:** the `components={{ code: … }}` map is **dead**. No locale variant of `enrollmentKeys.description` contains a `<code>` tag, so the styled `<code>` wrapper it exists to provide is never applied. The string was presumably meant to be `… with <code>breeze-agent enroll &lt;key&gt;</code>`.

**Variant B — raw HTML entities sitting in locale JSON.**

The Suppress-alert dialog (`/alerts` → row `Mute`) renders:

> How long should **&ldquo;**E2E fixture: high CPU**&rdquo;** stay suppressed?

DOM proof:

```html
<p class="mt-1 text-sm text-muted-foreground">How long should &amp;ldquo;E2E fixture: high CPU&amp;rdquo; stay suppressed?</p>
```

React text-escapes the `&`, so an entity written into a JSON value can never decode. Scanning `locales/en/*.json` for `&[a-zA-Z]{2,10};` / `&#\d+;` finds **9 affected strings**:

| File | Key | Entity |
|---|---|---|
| `alerts.json` | `suppressAlertDialog.howLongShould` | `&ldquo;` |
| `alerts.json` | `suppressAlertDialog.staySuppressed` | `&rdquo;` |
| `alerts.json` | `createTicketFromAlertDialog.couldnTCheckForExistingTicketsOn` | `&apos;` |
| `alerts.json` | `createTicketFromAlertDialog.categoriesCouldnTBeLoaded` | `&apos;` |
| `backup.json` | `recoveryBootstrapTab.breezeBackupBmrRecoverTokenLtTokenGt` | `&lt;token&gt;`, `&lt;api-server&gt;` |
| `integrations.json` | `unifiIntegration.registerEachSelfHostedUniFiNetworkControllerA` | `&nbsp;` |
| `integrations.json` | `unifiIntegration.assignABreezeAgentAtTheSiteTo` | `&nbsp;` |
| `policies.json` | `configurationPolicies.featureTabs.patchTab.times` | `&times;` |
| `security.json` | `sensitiveDataCreateScanModal.form.deviceIdsPlaceholder` | `&#10;` |

`policies.json`'s `&times;` is a close-button glyph that will render as the literal string `&times;`, and `deviceIdsPlaceholder` wants a newline between two example UUIDs and instead prints `&#10;`.

## Root Cause

Both variants come from the same place: the i18n extraction pass moved copy out of JSX into JSON without accounting for the two contexts that treat the value as more than plain text — `<Trans>`'s markup parser, and the fact that a JSON value is inserted as a React text node (escaped) rather than as HTML.

Almost certainly the same extraction that produced #2649.

## Proposed Fix

- **Variant A:** author the string with the wrapper the render site expects and escape the placeholder for `Trans`'s parser — e.g. `"… with <code>breeze-agent enroll &lt;key&gt;</code>"` — so `components={{ code }}` becomes live and the brackets survive. Alternatively drop `<Trans>` for a plain `t()` here, and then delete the unused `components` prop.
- **Variant B:** replace each entity with the literal character in the JSON (`&ldquo;` → `“`, `&nbsp;` → U+00A0, `&times;` → `×`, `&#10;` → a real `\n`, `&apos;` → `'`, and the `backup.json` `&lt;token&gt;` → `<token>`), across **all** locales, not just `en`.
- **Guard:** a unit test over `locales/**/*.json` asserting no value matches `/&(?:[a-zA-Z]{2,10}|#\d{2,5});/` would make both variants un-reintroducible. `localeParity.test.ts` is the natural home.

## Affected Files

- `apps/web/src/components/settings/EnrollmentKeyManager.tsx` (primary — line 497)
- `apps/web/src/locales/*/settings.json` (`enrollmentKeys.description`)
- `apps/web/src/locales/*/alerts.json`, `backup.json`, `integrations.json`, `policies.json`, `security.json` (the 9 entity strings)
- `apps/web/src/lib/i18n/localeParity.test.ts` (guard)

## Related

- #2649 — same extraction pass, different artifact (humanized key names replacing English copy)

## Reported By

Internal ego-browser UI QA sweep, 2026-08-24, verified in-browser on `main` (`6ec30b06a`). Evidence in `docs/testing/FEATURE_TEST_LOG.md`.

ToddHebebrand:
Closes #3964

## Problem

Locale values carrying markup-ish characters reached the user as **literal source text**. Two variants, one root cause: the i18n extraction pass moved copy into JSON without accounting for the two contexts that treat a value as more than plain text — `<Trans>`'s markup parser, and the fact that a JSON value is inserted as a React *text node* (escaped), never as HTML.

## Variant A — `<Trans>` escaped the literal `<key>`, and its `components` map was dead

`/settings/enrollment-keys` rendered `… breeze-agent enroll &lt;key>`. `<Trans>` parsed the plain `<key>` in `enrollmentKeys.description` as an element, found no entry for it in `components`, and escaped it. All 8 locales carry the same value, so every language was affected.

The same line's `components={{ code: <code className="…" /> }}` was simultaneously **dead** — no locale variant contained a `<code>` tag, so the styled chip never applied.

**The issue's proposed fix does not work.** It suggested authoring `<code>breeze-agent enroll &lt;key&gt;</code>`. I probed that against the pinned `react-i18next@17` / `i18next@26` before implementing:

| Authored value | Rendered `innerHTML` | |
|---|---|---|
| `<code>… enroll &lt;key&gt;</code>` | `<code>… enroll &amp;lt;key&amp;gt;</code>` | ❌ entity not decoded |
| `<code>… enroll {{ph}}</code>` + `values={{ ph: '<key>' }}` | `<code>… enroll &amp;lt;key&gt;</code>` | ❌ escaped |
| `<code>… enroll <key/></code>` + `components={{ key }}` | `<code>… enroll &lt;key&gt;</code>` | ✅ literal `<key>` |

`<Trans>` does **not** decode HTML entities, so the proposed fix would have swapped one literal artifact for another.

**Implemented instead:** the value carries `<code>breeze-agent enroll <key/></code>`, and the render site binds `key` to the literal text. This:

- makes the previously-dead `components={{ code }}` map live (fixes the second bug on that line),
- renders real angle brackets,
- keeps the sentence translatable as **one unit** — `extractionQuality.test.ts` explicitly forbids rebuilding translated sentences from English fragments, which the "drop `<Trans>` and concatenate" alternative would have required,
- is enforced across locales for free by `localeParity`'s existing `richTextTags` contract,
- and leaves **no HTML entity in the JSON**, so the new guard below needs no allowlist carve-out.

## Variant B — 9 strings held raw HTML entities

React text-escapes the leading `&`, so an entity written into a JSON value can never decode in *any* render path. Every one of the 9 keys is rendered through a plain `t()` call (verified — none goes through `<Trans>`), so replacing each with the character itself is safe:

| File | Key | Was | Now |
|---|---|---|---|
| `alerts.json` | `suppressAlertDialog.howLongShould` / `.staySuppressed` | `&ldquo;` `&rdquo;` | `“` `”` |
| `alerts.json` | `createTicketFromAlertDialog.couldnT…` ×2 | `&apos;` | `'` (en, it-IT) / `’` (fr-FR, fr-CA — matches the ~130:5 convention already in those files) |
| `backup.json` | `recoveryBootstrapTab.breezeBackupBmrRecover…` | `&lt;token&gt;` `&lt;api-server&gt;` | `<token>` `<api-server>` |
| `integrations.json` | `unifiIntegration.*` ×2 | `&nbsp;` | U+00A0 |
| `policies.json` | `configurationPolicies.featureTabs.patchTab.times` | `&times;` | `×` |
| `security.json` | `sensitiveDataCreateScanModal.form.deviceIdsPlaceholder` | `&#10;` | a real newline |

Applied across **all 8 locales**, not just `en`. `CreateScanModal.tsx`'s inline `defaultValue` fallback carried the same `&#10;` and is fixed with it.

## Guard

`localeParity.test.ts` now fails any locale value containing an HTML entity — one `it` per locale, naming the offending key and entity. Plus two helper unit tests, including one proving a bare `&` (`Research & Development`) is not flagged.

## Verification

- `pnpm exec tsc --noEmit` — clean.
- `localeParity` + `extractionQuality` + `i18n` + `keyUsage` + `terminologyQuality` + `format` — 113 passed.
- `translationCoverage` — 15 passed.
- `EnrollmentKeyManager` + `SuppressAlertDialog` — 29 passed (3 new).
- `CreateTicketFromAlertDialog`, `RecoveryBootstrapTab`, `PatchTab`, `UnifiIntegration`, `FindingsTab` — 61 passed.
- **Non-vacuity proven:** stashing the source fix and re-running leaves the 11 new assertions **red** (8 locale-guard failures, `expected null to be truthy` for the dead `code` map, `expected … to contain 'breeze-agent enroll <key>'`, and the curly-quote assertion) — then green with the fix restored.

## Note for the reviewer

`policies.json`'s `patchTab.times` (`×`) is a decorative close-button glyph whose button already carries `aria-label={t('common:actions.close')}`. Arguably it should be inlined in the component rather than translated at all, but that is an extraction-scope change beyond this fix — it now at least renders correctly.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer (silent-failure-hunter and type-design-analyzer skipped: the diff adds no error handling and no new types).

**Findings:** 4 raised → all addressed in d10f8eb79; 0 outstanding.

1. **Regression this PR introduced** (found independently by *both* code-reviewer and pr-test-analyzer, `localeParity.test.ts:177`): the `protectedCommands` fixture still held the pre-fix, entity-encoded spelling of the bmr-recover command. Once the locale values switched to real angle brackets, that literal no longer occurred anywhere, its expected occurrence count fell to `0` in every namespace, and the `<locale> preserves source-aligned technical literals` check passed **vacuously** — a translator mangling `--token <token>` into `--jeton <jeton>` would have shipped undetected. Fixed, and proven live again by injecting exactly that mistranslation into `fr-FR/backup.json`: green before the fix commit, red after (`protected literal occurrence mismatch (cli-flag:--token expected 1, received 0; ...)`).
2. `htmlEntityValueErrors` walked array elements but not *into* them — an entity inside an array-of-objects leaf would have been skipped (latent, no such leaf exists today). Now recurses through containers, with a helper test pinning it.
3. Both copies of the "react-i18next does not decode entities" note stated a library-wide fact; it is actually contingent on this app never enabling `shouldUnescape`. Hedged in `EnrollmentKeyManager.tsx` and `localeParity.test.ts`.
4. comment-analyzer verified every other factual claim in the new comments against `react-i18next@17.0.10`'s source and the pre-fix locale JSON — all accurate, no removals recommended.

**Tests:** `apps/web` affected suite green — `vitest run src/lib/i18n/ src/components/settings/EnrollmentKeyManager.test.tsx src/components/alerts/SuppressAlertDialog.test.tsx`, **158 passed (9 files)**; plus `CreateTicketFromAlertDialog`, `RecoveryBootstrapTab`, `PatchTab`, `UnifiIntegration`, `FindingsTab` — 61 passed. `tsc --noEmit` clean. Full PR CI green (Test Web, Test API, all 4 Integration shards).

**Non-vacuity:** stashing the source fix leaves the 11 new assertions red (8 locale-guard failures, the dead-`code`-map `expected null to be truthy`, the escaped-`<key>` text, and the curly-quote assertion), green with it restored. pr-test-analyzer reproduced this independently.

**Deliberately out of scope** (noted by reviewers, not defects of this PR — worth follow-up issues):
- ~35 components use the inline `t(key, { defaultValue: '...' })` fallback pattern. Those literals live in `.tsx`, so the new locale guard is structurally blind to them. `CreateScanModal.tsx:155` was the only one actually carrying an entity (fixed here); a lint rule or grep check over `defaultValue:` literals would close the class.
- Two *pre-existing* dead `components` maps: `LoginBrandingCard.tsx:163` (`code`) and `OrgDefaultsEditor.tsx:498` (`strong`). Same dead-map shape this PR fixed for `enrollmentKeys.description`, but harmless — no tag present means no escaping bug, just inert code. All 24 other `<Trans>` sites were enumerated and their locale values matched against their maps: no other unmapped-tag bug exists.
- `PatchTab`'s ring-editor modal and `UnifiIntegration`'s self-hosted cards are render paths no existing test exercises at all — pre-existing coverage gap, untouched by this PR.

**Status:** review-clean, CI green, awaiting maintainer merge. Issue #3964 stays open and assigned until you verify in-browser.

PR #3972, opened 2026-08-25
fix(web): restore the dropped space between i18n labels and their values (#3965)

### LINKED ISSUE #3965: [UI] i18n extraction dropped the space between a label and its value — 14 sites render glued text ("0 of0 providers", "Step1 of3")
## Description

The i18n extraction split label copy into a `t()` call and left the following JSX expression immediately adjacent, with nothing supplying the space that used to be part of the literal. The result is visibly glued text.

Confirmed in a browser on `main` (`6ec30b06a`) at `/settings/sso`:

> SSO Providers
> **0 of0 providers**

Source, `apps/web/src/components/settings/SsoProviderList.tsx:87`:

```tsx
{filteredProviders.length} {t('ssoProviderList.of')}{providers.length} {t('ssoProviderList.providers')}
```

`ssoProviderList.of` is `"of"` — no trailing space — so `of` and `0` collide. Note the *first* gap is fine (there's a literal `{' '}`-equivalent space in the JSX); only the gap after the `t()` call is missing.

## Scope

Method: match `{t('KEY')}{expr` in `apps/web/src/**/*.tsx` excluding `{' '}`, resolve `KEY` against `locales/en/*.json`, and keep only values whose text does not already end in a space or opening punctuation. 21 raw matches, 7 of them false positives (the following template literal starts with its own space, e.g. the `Tax` totals rows). The remaining **14 are real**:

| File:line | Renders as |
|---|---|
| `components/settings/SsoProviderList.tsx:87` | `0 of0 providers` |
| `components/settings/ApiKeyList.tsx:103` | `N of0 keys` |
| `components/settings/ApiKeyList.tsx:237` | `Page1 of3` |
| `components/settings/AccessReviewPage.tsx:668` | `Due8/24/2026` |
| `components/settings/AccessReviewPage.tsx:701` | `N of0 reviewed` |
| `components/settings/AccessReviewPage.tsx:756` | `Showing12 of40 users` |
| `components/settings/AccessReviewForm.tsx:135` | `Step1 of3` |
| `components/settings/AccessReviewList.tsx:138` | `N of0` |
| `components/settings/CatalogItemsTab.tsx:539` | `Showing the first100 items…` |
| `components/settings/RoleManager.tsx:395` | `Permissions forPartner Billing` |
| `components/settings/TdSynnexEcExpressPanel.tsx:428` | `SYNNEX SKUABC-123` |
| `components/settings/SsoProviderForm.tsx:277` | `Client Secret` + adjacent conditional node |
| `components/settings/MFASettings.tsx:584` | `(phone verified: …1234)` |
| `components/remote/FileManager.tsx:1218` | `Activity7` |

The `AccessReview*` cluster is the worst — a user setting up an access review sees `Step1 of3`, `Showing12 of40 users` and `Due8/24/2026` on the same screen.

## Root Cause

Mechanical artifact of the extraction codemod: it replaced `of ` (with trailing space) with `{t('…of')}` and dropped the space, because the space was consumed as part of the matched literal rather than preserved as JSX whitespace. The same codemod produced #2649 and the entity/markup damage in #3964.

## Proposed Fix

Per site, add the separator back in JSX — `{t('…of')}{' '}{count}` — rather than baking a trailing space into the translation value. Trailing whitespace in a JSON value is invisible to translators, easy to strip in a round-trip through a TMS, and breaks languages that order the number differently.

Better where the string is a real sentence: use interpolation so the whole phrase is one translatable unit, e.g. `t('ssoProviderList.count', { shown, total })` → `"{{shown}} of {{total}} providers"`. That also fixes the fact that several of these are currently unlocalizable — `"of"` as a standalone token cannot be translated correctly for languages that don't use an X-of-Y construction.

A lint rule or test asserting no `{t\('[^']+'\)\}\{` adjacency (other than `{' '}`) would keep it from recurring.

## Affected Files

The 14 files listed above; `components/settings/AccessReview*.tsx` is the densest cluster.

## Related

- #3964 — same extraction pass, entities and angle brackets in locale values
- #2649 — same extraction pass, humanized key names replacing English copy

## Reported By

Internal ego-browser UI QA sweep, 2026-08-24, `/settings/sso` confirmed in-browser on `main` (`6ec30b06a`); the other 13 identified by source scan. Evidence in `docs/testing/FEATURE_TEST_LOG.md`.

ToddHebebrand:
Closes #3965

## What was wrong

The i18n extraction codemod consumed the separating space into the matched English literal and never emitted it back as JSX whitespace. `{t('…of')}{count}` concatenates its two JSX children with nothing between them, so the UI rendered glued text:

| Screen | Before | After |
|---|---|---|
| `/settings/sso` | `0 of0 providers` | `0 of 0 providers` |
| API keys | `3 of0 keys` · `Page1 of3` | `3 of 0 keys` · `Page 1 of 3` |
| Access review | `Step1 of3` · `Showing12 of40 users` · `Due8/24/2026` · `7 of40 reviewed` | all spaced |
| Roles | `Permissions forPartner Billing` | `Permissions for Partner Billing` |
| TD SYNNEX | `SYNNEX SKUABC-123` | `SYNNEX SKU ABC-123` |
| Passkeys | `Last used:8/24/2026` | `Last used: 8/24/2026` |

## Why interpolation, not a trailing space in the JSON

A trailing space baked into a translation value is invisible to translators, trivially stripped by a TMS round-trip, and useless to a language that orders the value before the label. Each composed phrase instead becomes **one interpolated translatable unit**:

```diff
-{filteredProviders.length} {t('ssoProviderList.of')}{providers.length} {t('ssoProviderList.providers')}
+{t('ssoProviderList.countSummary', { shown: filteredProviders.length, total: providers.length })}
```
```jsonc
"countSummary": "{{shown}} of {{total}} providers"
```

**This makes several of these strings translatable for the first time.** A standalone `"of"` cannot be rendered correctly in a language with no X-of-Y construction. tr-TR had translated it as the postposition `arasında`, and `roleManager.permissionsFor` as `için izinler` — which composed **backwards** (`için izinler Partner Billing`). The new tr-TR values put the placeholder where Turkish needs it:

```jsonc
"permissionsForRole": "{{role}} için izinler",
"countSummary":      "{{total}} sağlayıcıdan {{shown}} tanesi"
```

All 8 locales (en, de-DE, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR) were updated in step so `localeParity` stays exact. `apiKeyList.keys` also picked up a correction along the way: tr-TR had `tuşları` (keyboard keys) → now `anahtar` (API key).

Two sites keep an explicit `{' '}` separator instead, because the two parts are genuinely independent nodes — `SsoProviderForm`'s Client Secret label restores the JSX newline whitespace the pre-extraction source had. The span's `ml-2` styles the gap, but the `<label>`'s **accessible name concatenates its text nodes verbatim**, so the hint ran straight into the label with no gap.

## Deviations from the issue's list of 14

- **`FileManager.tsx:1218` is a false positive** — the following template literal already starts with its own space (`` ` (${n})` ``), so it renders `Activity (7)`. Left alone; recorded in the guard's reviewed list.
- **`MFASettings.tsx:584` renders correctly today** (`(phone verified: ...1234)`), but the closing paren lived in JSX where no locale could move it. Folded into the value as `"(phone verified: ...{{last4}})"` — zero rendering change, one fewer fragment.
- **`ProfilePage.tsx:802` is a 14th real site the issue's scan missed** — `profilePage.lastUsed` is `"Last used:"` and rendered `Last used:8/24/2026`. The issue's filter dropped it because the value ends in punctuation.

## Review round 1 — a regression this PR introduced, and two pre-existing siblings

Review caught a **real user-visible regression in the first cut**: the new interpolated badge key was named `accessReviewPage.dueDate`, but that key **already existed** as the completed-campaigns table column header (`AccessReviewPage.tsx:572`, a bare `t('accessReviewPage.dueDate')`). Overwriting its value with `"Due {{date}}"` made the header render the literal text `Due {{date}}` in all eight locales — i18next has no `missingInterpolationHandler` configured, so an unfilled token passes through verbatim. Fixed by restoring `dueDate` to its original header string everywhere and moving the badge to a new `dueOn` key.

Scanning for the same shape turned up **two pre-existing instances of the identical bug**, both one call site from code this PR touched:

- `securitySecurityScanManager.started` is `"Started {{time}}"`, used both as an interpolated cell (`SecurityScanManager.tsx:526`, correct) and as a bare column header (`:562`) that rendered `Started {{time}}`. Split out `startedColumn`.
- `roleManager.cloneRole` is `"Clone Role: {{name}}"`, used bare as a button `title=` (`:360`), rendering the raw token as a tooltip. It now passes `{ name: role.name }`, as its other call site already did.

Also from review: **de-DE and tr-TR `manufacturerPart` were missing their leading separator space**, so those two locales rendered `SYNNEX SKU 12345· MFG ABC-123` — the exact glued-text symptom this PR exists to remove, on the line next to one it fixed. Restored the space and localised the abbreviation (`Hrst.`, `Üretici`) the way fr-FR already localises it to `FAB`, rather than making the value byte-identical to English and burning translation-coverage headroom.

## Recurrence guards

`apps/web/src/lib/i18n/extractionQuality.test.ts` gains two repo-wide scans, each with a unit test for the scanner itself:

1. **Glued adjacency** — every `{t('key')}{expr}` whose en value does not already end in whitespace must carry an explicit `{' '}`/`{" "}` or be listed in `REVIEWED_ADJACENCIES` with a written reason. The 8 survivors are all sites where the following template literal supplies its own leading separator (the `Tax (12%)` totals rows, `DeviceReliabilityPanel`, `FileManager`) or where the value deliberately ends mid-token (`SuppressAlertDialog`'s opening curly quote — the raw entity there is #3964's problem). Matches both quote styles: the repo has `t("key")` call sites, so a single-quote-only regex would have been blind to half of them.
2. **Unfilled placeholder** — every bare `t('key')` whose en value still contains a `{{token}}` fails, with a reviewed allowlist for the four keys whose placeholders are shown *on purpose* (form `placeholder=` hints demonstrating template syntax the user is meant to type). **This is the guard that would have caught the `dueDate` collision above**, which is why it exists.

Both were proven non-vacuous by re-introducing the original bugs and confirming the tests fail — the adjacency guard also caught a first attempt at the `SsoProviderForm` fix that nested `{' '}` one level too deep.

Also swept: the now-dead `settings.json:accessReviewPage.of` exemption in `localeParity.test.ts`'s `pathLikeValueExemptKeys` (this PR deletes that key).

## Verification

- `vitest run src/lib/i18n/` — **7 files, 123 passed** (localeParity, translationCoverage, keyUsage, extractionQuality, i18n runtime, format, terminologyQuality)
- `vitest run src/components/settings/ src/components/security/` plus i18n — **87 files, 873 passed**
- `tsc --noEmit -p apps/web/tsconfig.json` — clean
- `eslint` on every changed source file — clean
- New render-level assertions cover the fixed summaries end-to-end, which nothing did before: `0 of 0 providers`, `Step 1 of 3`, `Showing 12 of 40 users`, `Permissions for Partner Billing`, `Last used: 8/24/2026`.
- No `.astro` files touched, so `astro check` is not implicated.

## Not in scope

`SuppressAlertDialog`'s `&ldquo;`/`&rdquo;` entities are the same codemod's damage but belong to #3964.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 0: code-reviewer, pr-test-analyzer, comment-analyzer (parallel); round 1: focused code-reviewer re-review of the fix commit.

**Findings:** 9 raised across both rounds → all 9 addressed in `686782439` and `952256ab7`; 0 outstanding.

Two were genuine user-visible bugs, and **both were introduced or missed by me, then caught by review** — worth calling out because they are the same failure mode this PR exists to fix:

1. **Critical (round 0)** — I named the new interpolated badge key `accessReviewPage.dueDate`, which **already existed** as the completed-campaigns table column header. Overwriting its value with `"Due {{date}}"` made that header render the literal token `Due {{date}}` in all 8 locales (no `missingInterpolationHandler` is configured). Restored the header string everywhere; badge moved to `dueOn`.
2. **Important (round 1)** — my new guards were blind to the `i18n.t(...)` calling convention (`ADJACENCY` required a literal `{t(`; `BARE_CALL`'s `(?<![\w.])` lookbehind rejected the preceding dot). The repo uses hook-`t()` and singleton-`i18n.t()` side by side, ~1,600 sites on the latter. Widening the scan immediately exposed a live `RevokeMyLaptop?` heading in the **revoke-device confirmation dialog** — the exact #3965 symptom in a destructive-action flow.

Scanning for the shape of #1 also turned up **two pre-existing instances of the same class** (`securitySecurityScanManager.started` rendering `Started {{time}}` as a column header; `roleManager.cloneRole` rendering `Clone Role: {{name}}` as a tooltip), plus **de-DE/tr-TR `manufacturerPart` missing its leading space** (`SYNNEX SKU 12345· MFG ABC-123`). All fixed here.

Remaining findings addressed: dead `accessReviewPage.of` exemption swept from `localeParity`'s `pathLikeValueExemptKeys`; adjacency regex widened to both quote styles; a JSX comment that quoted a screen-reader announcement not matching the actual locale value corrected; guard scope and the unresolved-key skip documented; render-level assertions added for the fixed summaries (nothing covered them before).

**Guards:** both scans were proven non-vacuous by re-introducing the original bugs and confirming the tests fail — the `dueDate` collision fails 2 tests, the `RoleManager` glue fails the adjacency test, and the adjacency guard also caught a first attempt at the `SsoProviderForm` fix that nested `{' '}` one level too deep.

**Tests:** `vitest run src/lib/i18n/ src/components/settings/ src/components/security/` — **87 files, 874 passed**; `src/lib/i18n/` alone 124 passed. `tsc --noEmit -p apps/web/tsconfig.json` clean; `eslint` on every changed source file clean. No `.astro` touched.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3965 — the reporter should confirm the screens first.

**One call for the maintainer:** this PR grew past the issue's stated scope to fix 3 pre-existing bugs of the same class (`started` header, `cloneRole` tooltip, `manufacturerPart` spacing) and 1 more the widened guard exposed (`revoke` dialog heading). They are each one line plus locale values, and leaving them would have meant allowlisting known-broken strings in the new guards. Say the word if you'd rather split them into a follow-up.

ToddHebebrand:
**CI:** all checks complete on `952256ab7` — **53/53 green, 0 failing** (Test Web, Test API, Type Check, Lint, and all 4 Integration Tests shards included). Still awaiting maintainer merge; #3965 stays open and assigned until the reporter confirms the screens.

PR #3973, opened 2026-08-25
fix(alerts): resolve acknowledged/resolved-by user ids to display names (#3966)

### LINKED ISSUE #3966: [UI] Alert detail drawer shows a raw user UUID for "Acknowledged by" instead of the technician's name
## Description

Acknowledge an alert, then open its detail drawer. The Device Information block reads:

> Acknowledged &nbsp;&nbsp; 8/24/2026, 5:00:19 PM by **9cea2f85-2da1-445d-88cc-7c404d7504c4**

That UUID is the acknowledging user's id. The name — "Breeze Admin" — is what the operator needs, and it's what the same page shows everywhere else (the Recent Activity and Audit Trail tables both resolve the actor to a display name).

Reproduced in-browser on `main` (`6ec30b06a`): `/alerts` → `Ack` on any active alert → click the row to open the drawer.

## Root Cause

The API returns `acknowledgedBy` as a bare user id and the components print it verbatim:

`apps/web/src/components/alerts/AlertDetails.tsx:258-259`

```tsx
{alert.acknowledgedBy && (
  <span className="text-muted-foreground"> {t('alertDetails.by')} {alert.acknowledgedBy}</span>
)}
```

`apps/web/src/components/alerts/AlertDetailPage.tsx:432-435` has the same shape. Both type the field as `acknowledgedBy?: string` with no accompanying name field, so there is nothing to render *but* the id.

## Proposed Fix

Preferred: resolve it server-side. Have the alert read path join `users` and return `acknowledgedByName` (or a small `{ id, name, email }` actor object) alongside the id, the way the audit endpoints already do. Both components then render the name and keep the id available for tooling.

Fallback if the join is unattractive on the alerts hot path: resolve client-side from the already-loaded user list and fall back to the id when the user isn't found (e.g. deleted user). Avoid rendering a bare id as the primary text in either case.

Worth checking `resolvedBy` / `suppressedBy` on the same records while in here — they are likely the same shape.

## Affected Files

- `apps/web/src/components/alerts/AlertDetails.tsx` (primary — line 259)
- `apps/web/src/components/alerts/AlertDetailPage.tsx` (primary — line 435)
- `apps/web/src/components/alerts/AlertList.tsx` (type definition, line 49)
- alert read route in `apps/api/src/routes/alerts*` (if fixing server-side)

## Reported By

Internal ego-browser UI QA sweep, 2026-08-24. Evidence in `docs/testing/FEATURE_TEST_LOG.md`.

ToddHebebrand:
## Problem

Acknowledging an alert and opening its detail drawer rendered the acknowledging user's raw UUID:

> Acknowledged &nbsp; 8/24/2026, 5:00:19 PM by **9cea2f85-2da1-445d-88cc-7c404d7504c4**

The API returned `acknowledgedBy` as a bare user id with no accompanying name, and both `AlertDetails.tsx` (drawer) and `AlertDetailPage.tsx` printed it verbatim. `resolvedBy` had the same shape.

## Fix

Resolved **server-side in the alert read path** (the issue's preferred option) rather than with an extra web fetch. Both read paths are covered — the drawer's data comes from `GET /alerts`, the standalone page from `GET /alerts/:id`.

New `apps/api/src/routes/alerts/actorNames.ts`:

- `withAlertActorNames(rows)` batches **one** lookup of the `acknowledgedBy` / `resolvedBy` / `dismissedBy` ids across a whole page of alerts and attaches `acknowledgedByName` / `resolvedByName` / `dismissedByName`.
- It skips the query entirely when no row carries an actor id, so an untouched page pays nothing.
- A field the caller didn't select gets no fabricated name key.

### Why the lookup runs in a system DB context

`users` RLS grants visibility via `breeze_has_partner_access(partner_id)` **OR** org access to `users.org_id` **OR** self-read, and MSP staff rows carry `org_id = NULL`. An **org-scoped** caller therefore cannot see the technician who acknowledged their alert, so a plain `leftJoin(users)` would silently null the name out for exactly the common MSP case — the bug would persist for customer users while looking fixed for partner admins.

The lookup is wrapped in `runOutsideDbContext` + `withSystemDbAccessContext`, the same shape (and same justification) as the partner-wide notification-channel join already in this handler. Exposure is bounded:

- ids only ever come off alert rows the caller was **already access-checked** for (`getAlertWithOrgCheck` / the list's org + site conditions);
- only the display **name** leaves the block — no email, no tenancy columns;
- this is strictly **less** than the Audit Trail already shows those same callers for the same actions (`resolveActorName` falls back to `actorEmail`).

No schema change, no migration, no new table.

### UI

- Renders the name; keeps the raw id in a `title` tooltip for tooling.
- A deleted user yields `null` → generic `Unknown user` label, never a bare UUID.
- The "by" clause is omitted entirely when there's no actor id at all, so the optimistic post-acknowledge render (which sets `acknowledgedAt` but no actor) doesn't claim an unknown user.
- `unknownUser` added to all 9 locales; `localeParity` green.

`suppressedBy` doesn't exist as a column (suppression uses `suppressedUntil`), so there was nothing to fix there. `dismissedBy` is handled by the same helper.

## Verification

| Check | Result |
|---|---|
| `apps/api` — `src/routes/alerts.test.ts` + `src/routes/alerts/` + `src/routes/mobile*` (single-fork) | 315 passed, 3 skipped |
| `apps/web` — `src/components/alerts/` + `src/lib/i18n/` (single-fork) | 249 passed |
| `apps/web` `tsc --noEmit` | clean |

New tests:

- `apps/api/src/routes/alerts/actorNames.test.ts` — compiled-SQL assertion on the `"users"."id" in (...)` predicate against the **real** schema (not a mock-only `where` assertion), system-context wrapping, id dedupe, columns limited to `{id, name}`, deleted-user → `null`, no query when there's nothing to resolve.
- `apps/api/src/routes/alerts.test.ts` — list route returns both names; no users round trip when no alert has an actor.
- `apps/api/src/routes/alerts/byid.sitescope.test.ts` — by-id route returns the name, and `null` for a deleted user.
- `apps/web/src/components/alerts/AlertDetails.actorName.test.tsx` — name rendered, UUID absent from the DOM text, tooltip present, generic fallback, no "by" clause without an actor.

Closes #3966


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter (type-design-analyzer skipped: no new domain types beyond two response fields; comment-analyzer skipped: no doc surface changed).

**Findings:** 1 HIGH + 3 coverage gaps raised → all addressed in `8928d82e2`; 1 knowingly deferred; 0 outstanding.

- **HIGH — silent-failure-hunter:** the actor-name enrichment was unguarded, so a transient `users` failure (pool exhaustion, PG blip, a fault establishing the system context) would discard an already-fetched, already-access-checked alerts payload and 500 the whole Alerts page for a cosmetic name. `withAlertActorNames` now catches, reports via `captureException` + `console.error`, and returns the rows unenriched — clients already render a missing name as the generic "Unknown user" label. Loud for engineers, graceful for the operator: without the report the degradation would itself be silent, indistinguishable from "that technician was deleted".
- **pr-test-analyzer:** batching was never proven end-to-end (every test used a single row) → added three alerts acknowledged by one technician, asserting exactly one query with deduped params; added coverage for the new throw path; the by-id route now asserts `resolvedByName` alongside `acknowledgedByName`; and `AlertDetailPage.tsx` — the second render site this PR touches, which had **no test file at all** — now has one mirroring the drawer's.
- **code-reviewer:** 0 findings. Confirmed the system-context read is bounded (ids only ever come off alert rows the caller was already access-checked for; only `id`/`name` selected) and that no other alert reader still hands a bare id to the UI.
- **Deferred (noted, not fixed):** there is no live-Postgres proof that an org-scoped caller can read an `org_id = NULL` staff row through the system context. That is the pre-existing shape of the partner-wide notification-channel join in the same handler, which likewise has no integration coverage — adding an RLS suite for it is out of scope for this fix.

**Tests:** `apps/api` `src/routes/alerts.test.ts` + `src/routes/alerts/` single-fork → 242 passed. `apps/web` `src/components/alerts/` single-fork → 135 passed (incl. `localeParity`, run separately, green). `tsc --noEmit` clean in both `apps/api` and `apps/web`; eslint clean on all changed files.

**CI:** 54/54 required checks green, including **Test API**, **Test Web** and all four **Integration Tests** shards.

**Status:** review-clean, awaiting maintainer merge. Nothing here needs a product decision — the one judgement call was resolving server-side in a system DB context rather than with a client-side join, and the reasoning is in the PR body and the helper's docstring.

PR #3974, opened 2026-08-25
fix(api): enforce organizations.slug uniqueness per partner in Postgres (#3967)

### LINKED ISSUE #3967: [API] organizations.slug is declared UNIQUE in the Drizzle model but has no unique index in Postgres — duplicate slugs are accepted
## Description

Two organizations under the same partner can be created with the **same slug**, via the normal Settings → Organizations UI. The API returns `201`.

Repro on `main` (`6ec30b06a`), local stack, `E2E_MODE` unset:

1. Settings → Organizations → Add organization → name `QA Sweep Org`, slug auto-fills `qa-sweep-org` → Create. → `POST /api/v1/orgs/organizations` **201**.
2. Add organization again → name `QA Sweep Org Dup`, slug typed as `qa-sweep-org` → Create. → **201**, no error, no warning.
3. `GET /api/v1/orgs/organizations`:

```
20a0e7f2-… | 'Default Organization' | slug='default-organization'
02125729-… | 'QA Sweep Org'         | slug='qa-sweep-org'
d13ed06b-… | 'QA Sweep Org Dup'     | slug='qa-sweep-org'   <-- duplicate
```

## Root Cause

The Drizzle model asserts an invariant the database does not enforce.

**Model** — `apps/api/src/db/schema/orgs.ts:21`:

```ts
slug: varchar('slug', { length: 100 }).notNull().unique(),
```

**Database** — `\d organizations` lists no index on `slug` at all:

```
"organizations_pkey" PRIMARY KEY, btree (id)
"organizations_id_partner_id_unique" UNIQUE, btree (id, partner_id)
"organizations_id_partner_uq" UNIQUE CONSTRAINT, btree (id, partner_id)
"organizations_partner_quick_support_uniq" UNIQUE, btree (partner_id) WHERE type = 'quick_support'
```

No migration under `apps/api/migrations/` ever creates one — the `.unique()` was added to the model without a matching DDL migration.

**No app-layer guard either.** `routes/orgs.ts` writes the value straight through on create (line 1417) and update (line 1766) with no pre-existence check.

The contrast with `partners` makes this look like an oversight rather than a decision: the partner routes *do* guard their slug, in three places —

```ts
or(eq(partners.inboundLocalPart, data.slug), eq(partners.slug, data.slug))
```

(`routes/orgs.ts:414`, `:882`, `:969`). Organizations got the model annotation but neither the index nor the check.

`pnpm db:check-drift` does not catch this class — it compares the schema against the migrations, not against the live database — so a model-only `.unique()` passes clean.

## Blast Radius

Limited **today**, but worth fixing before it isn't. `organizations.slug` is currently only ever projected into responses, never used as a lookup key:

- `routes/orgs.ts:1174`, `routes/partnerApi/organizations.ts:268`, `services/aiToolsOrgs.ts:147,254,273,352` — all `SELECT`-side
- `db/seed.ts:1095` — the only `WHERE slug = …`, and it's dev seeding

So this is a data-integrity / API-contract bug rather than a live tenant-isolation break. But the **partner API hands slugs to external consumers** as if they were stable identifiers, and any future slug-based lookup silently becomes cross-tenant. A duplicate today is a wrong-tenant read the first time someone writes `WHERE slug = ?`.

## Proposed Fix

1. Decide the intended scope. `.unique()` in the model means **globally** unique; given the tenancy model, unique **per partner** (`UNIQUE (partner_id, slug)`) is more likely what's wanted and is the weaker, safer constraint.
2. Migration: report and resolve existing duplicates first (per the repo's cleanup-statements-must-report-row-counts rule), then add the index — `CREATE UNIQUE INDEX IF NOT EXISTS organizations_partner_slug_uniq ON organizations (partner_id, slug)`.
3. Correct the model to match whichever scope is chosen — a bare `.unique()` is wrong if the constraint is composite.
4. Add the app-layer pre-check on create/update so the user gets "that slug is already taken" instead of a `23505`, mirroring the partner path.

## Affected Files

- `apps/api/src/db/schema/orgs.ts` (line 21 — the model claim)
- `apps/api/migrations/` (new migration — the missing index)
- `apps/api/src/routes/orgs.ts` (lines ~1417, ~1766 — missing guard; lines 414/882/969 are the pattern to copy)

## Reported By

Internal ego-browser UI QA sweep, 2026-08-24. Evidence in `docs/testing/FEATURE_TEST_LOG.md`.

ToddHebebrand:
Closes #3967

## The bug

`organizations.slug` has **no unique index in Postgres** and no app-layer guard, so two orgs under the same partner can both be created with slug `qa-sweep-org` and both return `201`. Reproduced through the ordinary Settings → Organizations UI.

The issue reported the Drizzle model as declaring `.unique()`; on current `main` it does not — `apps/api/src/db/schema/orgs.ts:119` is a bare `notNull()`. (The quoted line 21 is `partners.slug`, which *is* globally unique and *does* have a constraint.) The underlying defect is the same and worse than described: **nothing anywhere enforced it**, and three service paths were already written as if something did.

- `services/partnerCreate.ts` and `services/quickSupportOrg.ts` derive slugs that are unique *by construction*.
- `services/aiToolsOrgs.ts` and `services/orgImport/index.ts` de-duplicate against a **per-partner** `taken` set.
- `routes/partnerApi/provisioning.ts:282` maps a `23505` on this insert to `409 partner_provisioning_slug_conflict` — **a branch that could never fire**, because no constraint existed to raise it. It becomes live with this PR.
- `routes/orgs.ts` (the human-facing API) assumed nothing at all.

`pnpm db:check-drift` cannot catch this class: it reconciles the migration files against the `breeze_migrations` ledger, not the model against a live database. Its own output says so.

## Uniqueness scope: per-partner, case-insensitive, lifetime

```sql
CREATE UNIQUE INDEX IF NOT EXISTS organizations_partner_slug_uniq
  ON organizations (partner_id, lower(slug));
```

Each of the three properties was traced, not guessed. Both an independent trace of every generator/lookup/projection and a second read-only design review (Codex, `xhigh`) landed on the same shape.

**Per-partner, not global.** The only equality lookup on an org slug in the repo is `db/seed.ts:1093`, and it is already `AND partner_id = …`. Nothing routes, links, or authenticates by org slug: `apps/portal` never references it, no web/astro route or subdomain uses it, and the Partner API addresses organizations by `orgId` with `slug` merely a payload field. A global index would stop two unrelated MSPs from both onboarding an "acme" and would leak cross-tenant existence through the duplicate `409`. `partners.slug` is global for a concrete reason that does not apply here — it resolves a globally addressed inbound-email local part (`services/inboundEmail/resolvePartner.ts`).

**Case-insensitive.** Every first-party generator lowercases (`orgImport/slug.ts`, `aiToolsOrgs.slugifyOrgName`, and both web `slugify` helpers), and the web form enforces `^[a-z0-9-]+$`. But `createOrganizationSchema` and the Partner API body schema both accept `z.string().min(1).max(100)` with no format constraint, so without `lower()` `Acme` and `acme` remain creatable duplicates — the same ambiguity the issue is about, just narrower.

**Lifetime, NOT `WHERE deleted_at IS NULL`.** `DELETE /organizations/:id` soft-deletes. `services/orgImport/index.ts` deliberately reserves "every slug under the partner (**incl. soft-deleted**)" — its own comment — precisely so its `matched-soft-deleted` **reactivate** path cannot find the slug re-claimed by a replacement row. A live-rows-only index would allow exactly that and turn reactivation into a `23505`. The cost is that a slug stays reserved after a delete; the API says so explicitly rather than pointing the user at an org they cannot see (see the second error message below).

## Changes

| File | What |
|---|---|
| `apps/api/migrations/2026-09-08-organizations-partner-slug-unique.sql` | Report-and-rename pre-existing duplicates, then create the index |
| `apps/api/src/db/schema/orgs.ts` | Matching `uniqueIndex(...).on(partnerId, sql\`lower(slug)\`)` |
| `apps/api/src/routes/orgs.ts` | 409 on create and update, plus `23505` → 409 |
| `apps/api/src/services/aiToolsOrgs.ts`, `services/orgImport/index.ts` | Lower-case the reserved-slug sets |
| `apps/api/src/services/quickSupportOrg.ts` | Comment only — it claimed org slugs were *globally* unique |

### Migration

**The dedupe UPDATE runs against an RLS-FORCED table, and that nearly shipped broken.** `organizations` is `ENABLE` + `FORCE ROW LEVEL SECURITY`, and FORCE applies to the table *owner* too. Local and CI migrations run as a superuser, which is exempt; production is DigitalOcean managed Postgres where the admin role is **not** one — the exact trap `2026-07-16-td-synnex-sftp-price-file.sql` already documents in its own header. Without an ambient context `breeze_has_org_access(id)` is false for every row, so the rename would have no-op'd, the count-reporting `RAISE WARNING` would never have fired, and `CREATE UNIQUE INDEX` would have failed the deploy instead of healing it.

The DO block therefore opens with `PERFORM set_config('breeze.scope', 'system', true)` (precedent: `2026-04-13-fix-uuid-hostnames.sql`), transaction-local so it unwinds with autoMigrate's per-file transaction. Verified both directions against real Postgres as the unprivileged `breeze_app` role — **0 rows renamed without it, 1 with** — and that is now an integration test which replays the DO block (sliced out of the shipped file, not retyped) through `getAppDb()`. Removing the `set_config` line makes that test fail; I checked.

Pre-existing duplicates are **renamed, not deleted** — a slug clash is never a reason to drop a tenant's row. Canonical row per `(partner_id, lower(slug))` is the oldest by `created_at` (`id` as a deterministic tie-break); the rest become `<slug truncated to 91>-<first 8 chars of uuid>`, inside `varchar(100)`. Per the repo's cleanup rule the count lands in the Postgres log:

```
RAISE WARNING '#3967: renamed % duplicate organizations.slug row(s) ...'
```

The rename is deliberately *not* proven collision-free. If a survivor remains, `CREATE UNIQUE INDEX` aborts the whole migration loudly rather than the block silently mangling data a second time. Dedupe and index share one transaction. Idempotent; `2026-09-08` sorts after every shipped migration (the newest on `main` is `2026-09-07-*`), and the `2026-08-06` block is untouched.

### API

The index is the guarantee; the pre-check only buys a good error message, and it is racy on its own — so **both** write paths also map the violation to the same 409.

- `That organization slug is already in use`
- `That organization slug is still reserved by a deleted organization` — when the holder is soft-deleted, so the caller is not chasing an org they cannot find.

Two details worth review attention:

- The clash probe runs under a **system** DB context. `organizations` is policed by `breeze_has_org_access(id)`, so a partner-scope caller whose accessible org set excludes the clashing org would read zero rows and fall through to a raw 23505 → 500.
- On update the partner is resolved **from the org row**, not from `auth.partnerId` — which is `null` for a system-scope caller and would have scoped the probe to the wrong tenant.

## Verification

- **New integration suite** `organizationSlugUnique.integration.test.ts` (8 tests, real Postgres): the index exists with the expected definition and *no* partial predicate; same-partner duplicate rejected; case-variant rejected; **same slug under a different partner accepted**; soft-deleted holder still reserves its slug; and the migration replayed against hand-made duplicates renames the right row and is a no-op on re-run. This suite exists because nothing but a real `INSERT` could have caught the original bug.
- Verified in Postgres directly: `"organizations_partner_slug_uniq" UNIQUE, btree (partner_id, lower(slug::text))`.
- 526 unit tests green across 16 files (`orgs.test.ts`, `organizations.test.ts`, `aiToolsOrgs`, `quickSupportOrg`, `provisioning`, `autoMigrate`, `orgImport`, `utils`).
- Contract suites: `test:rls-coverage` 75 passed; `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip` green. No new table or column, so no cascade/export-policy registration is due.
- 9 org-related integration files, 67 tests, green. `tsc --noEmit` clean. `db:check-drift` clean. `check-migration-naming.sh` OK.
- Swept all 86 `insert(organizations)` call sites: every test fixture uses a dynamic slug, so no suite starts colliding under the new index.

Two pre-existing red tests in `orgs.test.ts` (`should create an organization`, `should allow system scope create with explicit partnerId`) only passed via a `db.select` mock leaking in from an earlier test — they 404'd in isolation on `main`. They now queue their own selects and stand alone.

## Review round

`/pr-review-toolkit:review-pr` ran four reviewers (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer). One HIGH (the migration/RLS interaction above), four MEDIUM, and a set of comment-precision fixes — all addressed in `f759bddea`:

- Dropped a locally-written unique-violation detector for the canonical `isPgUniqueViolation(err, constraint)` in `utils/pgErrors.ts`. The local copy matched the constraint name in *any* message regardless of SQLSTATE, so an unrelated error merely mentioning the index could have been reported as a routine 409 and hidden from Sentry.
- `services/aiToolsOrgs.ts` `create_org` had the pre-check but no 23505 mapping, so a lost slug race reached the AI operator as `Operation failed. Check server logs.` It now returns a retryable `ORG_SLUG_CONFLICT`.
- `services/quickSupportOrg.ts`'s `onConflictDoNothing` swallows *which* index fired; the resulting throw now names the org squatting the slug instead of a bare "provisioning failed".
- The 23505 → 409 race path now logs — a spike there means the pre-check has stopped working, and a bare 409 is indistinguishable from ordinary user error.
- Added the coverage the test reviewer asked for: legacy mixed-case slug reserved in both generators, the AI conflict mapping (and non-mapping for a different index), and the soft-deleted-holder message on the UPDATE path.

The test reviewer specifically checked the mocked route assertions against this repo's vacuous-Drizzle-assertion history and confirmed they are not vacuous: because `../db/schema` is mocked with per-column sentinels, column identity survives into `compiled.params`, so a case-sensitive, global, live-rows-only, or self-inclusive query each fail a distinct assertion.

## Known edge, not fixed here

`getOrCreateQuickSupportOrg` slugs the hidden org `quick-support-<partnerId>`. A partner who deliberately slugs one of their *own* orgs with their own partner UUID now wedges their own Quick Support (loud `throw`, own tenant only, no cross-tenant effect). Reserving the prefix would need a reserved-slug concept the API does not have; flagging rather than smuggling it in.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 reviewers on `5efb0f4cf` (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer), then a focused code-reviewer re-run on the fix commit `f759bddea` because the fix touched a migration.

**Findings:** 1 HIGH, 4 MEDIUM, 2 LOW, 5 comment-precision → all addressed in `f759bddea` and `4b9342aa2`; 0 outstanding. Re-review of `f759bddea` returned 0 Critical / 0 Important and one LOW (an untested error branch), fixed in `4b9342aa2`.

The HIGH is worth reading before merging, because it would have shipped silently:

> The migration's dedupe `UPDATE` runs against `organizations`, which is `ENABLE` + **`FORCE`** ROW LEVEL SECURITY — and FORCE applies to the table owner too. Local and CI migrations run as a superuser (exempt); production is DigitalOcean managed Postgres where the admin role is **not** one. With no ambient context `breeze_has_org_access(id)` is false for every row, so the rename would have matched **zero** rows, the count-reporting `RAISE WARNING` would never have fired, and `CREATE UNIQUE INDEX` would have failed the deploy instead of healing it. `2026-07-16-td-synnex-sftp-price-file.sql` documents this exact trap in its own header.

Fixed with `PERFORM set_config('breeze.scope', 'system', true)` in the DO block (precedent: `2026-04-13-fix-uuid-hostnames.sql`), plus a post-rename survivor check that raises with a count instead of leaving a bare 23505. Verified against real Postgres as the unprivileged `breeze_app` role (`NOSUPERUSER, NOBYPASSRLS`): **0 rows renamed without it, 1 with**. That is now an integration test replaying the DO block through `getAppDb()` — and I mutation-checked it: delete the `set_config` line and the test fails.

Other fixes: swapped a locally-written unique-violation detector (which matched the constraint name in *any* message regardless of SQLSTATE) for the canonical `isPgUniqueViolation` in `utils/pgErrors.ts`; added the missing 23505 → `ORG_SLUG_CONFLICT` mapping to the AI `create_org` tool; made `quickSupportOrg`'s provisioning throw name the org squatting the slug; logged the 409 race path.

**Tests:** 527 unit tests green across 16 files (`vitest run --no-file-parallelism`, single fork). Integration against real Postgres: `organizationSlugUnique` 8/8, plus `rls-coverage` 75, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, and 19 org-touching integration files — all green. `tsc --noEmit` clean. `db:check-drift` clean. `check-migration-naming.sh` OK. Index confirmed present in Postgres: `organizations_partner_slug_uniq UNIQUE, btree (partner_id, lower(slug::text))`.

**The test reviewer specifically checked the mocked route assertions** against this repo's vacuous-Drizzle-assertion history and confirmed they are not vacuous: because `../db/schema` is mocked with per-column sentinels, column identity survives into `compiled.params`, so a case-sensitive, global, live-rows-only, or self-inclusive query each trips a distinct assertion.

**For the maintainer:** the only judgement call left is the uniqueness scope itself — per-partner, case-insensitive, **lifetime** (a soft-deleted org keeps its slug reserved). The lifetime part is the debatable one and is deliberate: `services/orgImport/index.ts` already reserves slugs from "every slug under the partner (incl. soft-deleted)" so its reactivate path cannot find the slug re-claimed. The user-visible cost is that deleting "Acme Corp" and re-onboarding them needs a different slug; the API says so explicitly rather than pointing at an invisible org.

**CI (all green on `4b9342aa2`, 0 failures):** Test API, Type Check, Lint, Smoke Test, both Check Migrations jobs, and all four blocking Integration Tests shards. Verified the new suite actually **ran** rather than being silently skipped by `it.runIf` — `organizationSlugUnique.integration.test.ts` appears in the shard 1/4 log.

Issue #3704, opened 2026-08-17
[Web] Failed SSO exchange shows "Your session expired" — the sso_exchange_failed notice is overwritten by a racing refresh eviction


ToddHebebrand:
## Description

Follow-up to #3702 (which fixed the actual SSO lockout — the success path is not in question here). On the **failure** path, the SSO-specific error notice never reaches the user: it is issued, then immediately clobbered by the generic session-expired redirect.

An admin whose SSO config is broken is told "Your session expired. Please sign in again to continue." — which points away from SSO and invites an infinite retry loop, since signing in again just re-runs the same broken SSO round trip.

Repro on `main` (8c64a060d), clean browser with no session:

1. Navigate to `/#ssoCode=any-invalid-code`.
2. Land on `/login?reason=session-expired` showing "Your session expired. Please sign in again to continue."
3. Expected: `/login?error=sso_exchange_failed` showing "Single sign-on was almost complete, but the last sign-in step failed. Please start the sign-in again."

Deterministic — reproduced on two consecutive attempts with different codes.

## Root Cause

The network trace shows the intended path executing correctly, then losing a race:

```
POST /api/v1/sso/exchange              => 400 Bad Request      (correct rejection)
GET  /login?error=sso_exchange_failed  => FAILED net::ERR_ABORTED   <-- intended redirect, aborted
POST /api/v1/auth/refresh              => 401 Unauthorized      (racing eviction wins)
```

`AuthOverlay.tsx:130` (and the sibling at `:157`) does the right thing:

```ts
void navigateTo('/login?error=sso_exchange_failed', { replace: true });
```

but a concurrent `fetchWithAuth`/refresh path fails its 401 and hard-evicts to `/login?reason=session-expired`, aborting the in-flight navigation and replacing the URL.

The copy itself is fully implemented and unreachable on this path — `LoginPage.tsx:84` maps `sso_exchange_failed`, and all 8 locales carry `login.ssoErrors.ssoExchangeFailed`.

Note this is the mirror image of the race #3702 already fixed for the *success* path: the SSO login gate holds `requestTokenRefreshShared` callers until the exchange settles. On a **rejected** exchange the gate settles and releases the refresh callers, which then evict with the generic reason and outrun the error redirect.

## Proposed Fix

[...]

PR #3975, opened 2026-08-25
fix(auth): treat a throttled /auth/refresh as a wait, not a logout (#3696)

### LINKED ISSUE #3696: [Web][API] Navigating ~10 pages in a minute force-logs-out the user — every page load spends one of 10 allowed /auth/refresh calls
## Description

An authenticated user who clicks through the sidebar at a normal triage pace is **forcibly signed out and dumped on `/login`**, losing whatever they were doing. No warning, no toast — the app just lands on the sign-in screen.

Reproduced on a clean local stack running `main` (34d0234f5), with plain top-level navigations (no tabs, no automation tricks):

1. Log in as `admin@breeze.local`.
2. Click through 11 sidebar destinations (`/devices`, `/alerts`, `/scripts`, `/patches`, `/monitoring`, `/security`, `/reports`, `/audit`, `/settings/users`, `/settings/organizations`, `/integrations`, `/backup`) — about 3–4 s per page.
3. Around the 10th–11th page load the app redirects to `/login`.

Elapsed time in the repro: **11 page loads in ~43 seconds**.

API log for that burst — 9 refreshes succeed, then the limiter trips:

```
9 status=200
5 status=429   <- POST /api/v1/auth/refresh
```

## Root Cause

Two correct-in-isolation decisions multiply into a user-facing logout:

1. **The access token is held in memory only** (nothing in `localStorage`/`sessionStorage`; only `breeze_refresh_token` + `breeze_csrf_token` are cookies). Every full page load starts a fresh JS realm with no access token, so the client *must* `POST /api/v1/auth/refresh` to get one. Verified empirically: the refresh counter in the API log went **73 → 76 across exactly 3 navigations** — one refresh per page load, all `200`.

2. **`/auth/refresh` is capped at 10 requests / 60 s per user** — `apps/api/src/routes/auth/login.ts:751-766`, key `refresh:<userId>`, `rateLimiter(redis, refreshRateKey, 10, 60)`.

The web app is an Astro MPA, so **every navigation is a full document load**. That makes the per-user refresh budget a hard cap of ~10 page views per minute, which is well inside normal RMM triage behaviour.

Two aggravating details:

- On a `429` the client **retries with backoff** (`MAX_TRANSIENT_REFRESH_RETRIES = 2`, `apps/web/src/stores/auth.ts:409-410,436-441`), so a rate-limited refresh burns *more* of the same budget instead of backing off cleanly. That is why the burst shows 5 `429`s rather than 1–2.
- A `429` (a transient, retryable throttle) is ultimately surfaced as `AuthSessionExpiredError: Session expired` and treated as a hard logout, rather than as "back off and retry after `retryAfter`".

Note the client-side refresh dedup is **not** the problem — refresh is properly single-flighted per realm via a shared promise (`apps/web/src/stores/auth.ts:445-457`) plus a cross-tab Web Lock. The issue is one refresh *per realm*, and every navigation creates a new realm.

This is not an E2E/dev artifact: the limiter is skipped only when `E2E_MODE` is set, and we observed real `429`s, so production behaves the same way.

## Proposed Fix

Options, roughly in order of preference:

1. **Never hard-logout on a `429`.** Treat a rate-limited refresh as transient: honour the returned `retryAfter`, keep the session, and surface a non-destructive retry. A throttle is not an expired session. (Smallest, safest change; fixes the visible symptom.)
2. **Stop burning budget on retries** — don't count/retry a `429` against the same window; back off once using `retryAfter` instead of the transient-retry loop.
3. **Raise or re-scope the per-user refresh cap** so it reflects "one refresh per page load" reality (e.g. 60/min per user). The limiter runs *after* the refresh token is verified, so it isn't protecting against unauthenticated brute force — it's throttling already-authenticated holders.
4. Longer term, avoid a network round-trip per navigation (client-side routing, or a short-lived access token cached for the tab's lifetime) — but 1–3 fix the defect without changing the security posture.

## Affected Files

- `apps/api/src/routes/auth/login.ts` (primary — refresh rate limit, lines ~751-766)
- `apps/web/src/stores/auth.ts` (primary — 429 handling / transient retry, lines ~409-441, 729-741)

## Reported By

ToddHebebrand:
Closes #3696

## The bug, reproduced

An authenticated user clicking through the sidebar at normal triage pace was forcibly signed out. Reproduced live on `origin/main` with a real stack (`E2E_MODE` unset, limiter live) — **10 navigations in 24 seconds**:

```
NAV  9 /settings/users  @t+22s -> /settings/users        REFRESH 200
REFRESH 429 on /settings/roles
REFRESH 429 on /settings/roles      <- the retry loop, burning more budget
REFRESH 429 on /settings/roles
NAV 10 /settings/roles   @t+24s -> /login?next=%2Fsettings%2Froles&reason=session-expired
```

The session was healthy. Nothing expired.

## Root cause

`apps/web` is an Astro **MPA** and access tokens are memory-only by design, so **every** full-page navigation starts a fresh JS realm that must spend one `POST /auth/refresh`. That endpoint was capped at **10 req / 60 s per user**, which turned the refresh budget into a hard ceiling of ~10 page views per minute.

Three defects compounded it:

1. **The client retried *into* the throttle.** A 429 was classified `transient` and fed to the backoff ladder, which spends its entire budget in ~0.9 s — always *inside* the server's 60 s window. Each throttled load therefore fired 3 refreshes instead of 1, and `rateLimiter` ZADDs rejected requests too, so every retry consumed another slot and pushed recovery further out. The backoff could not help here even in principle.
2. **`transient` was handled identically to `auth-failed`.** Both `fetchWithAuth` eviction sites called `handleSessionExpired('session-expired')` → hard `location.replace('/login?…&reason=session-expired')`. A throttle became a logout, with copy that lied about the cause.
3. **The silent variant, which the reporter called worse.** `DashboardLayout` mounts `AuthOverlay`, `Sidebar`, `Header` and page content as *ungated siblings* — nothing blocks page content on auth recovery. AuthOverlay's own recovery used the boolean `restoreAccessTokenFromCookie()` and collapsed every non-success into a bare soft `navigateTo('/login')`. Mid-burst, pages painted their full chrome with every data call 401'd and no error, no toast, no retry affordance — indistinguishable from "you have no integrations configured".

### API — `routes/auth/login.ts`, `services/rate-limit.ts`

- Budget raised to **60/60 s** and re-keyed **per refresh-token family** (`refresh:fam:<fam>`) instead of per user, so one wedged browser can't starve the same person's other devices. `fam` is preserved across rotation, so a family is one browser profile's session chain — the granularity that actually matches "one navigation, one refresh".
- New opt-in `refundOnReject` limiter option, used here: **rejected attempts no longer consume window capacity.** Without it the advertised `Retry-After` is unhonourable — it derives from the *oldest* entry, so a client that waits exactly that long is still rejected by its own queued-up rejections, and pushed further out each time. Verified live: `Retry-After` now *decreases* across repeated rejections (59 → 58 → 58) instead of extending.
- `Retry-After` response header alongside the JSON field, clamped to `>= 1` so a `Retry-After: 0` (i.e. "hammer me again immediately") can never be advertised.
- Operator knobs `AUTH_REFRESH_RATE_LIMIT` / `AUTH_REFRESH_RATE_WINDOW_SECONDS`, read at call time, **mapped through both compose files** — a value present only in `.env` is inert.

The limiter's rationale comment was also corrected: it sits *before* reuse-detection and the user/epoch checks, so it is a cheap volume guard, not an authorization control. Credential-guessing is not its job (an attacker without a signed refresh token never reaches it), and raw volume is still capped by the global 300/min/IP limiter.

### Web — `stores/auth.ts`, `AuthOverlay.tsx`, `AuthGuard.tsx`, `DashboardWrapper.tsx`

- A 429 is now its own `{ kind: 'throttled', retryAfterMs }` outcome, returned **immediately** without spending transient retries.
- The shared refresh **waits the server-supplied window out once** (bounded; the parsed `Retry-After` is clamped to 1–90 s) and retries. Neither `fetchWithAuth` path evicts on a throttle. A definitive 401/403 still evicts exactly as before.
- New `authThrottledUntil` store field drives a **non-destructive AuthOverlay mask** with a countdown, auto-retry, manual retry, and a sign-out escape hatch — so a throttled page can never render as loaded-but-empty. The 10 s safety-net redirect is suspended while it is set (a rate-limit window is legitimately longer than 10 s, and leaving it armed would reinstate the very logout this removes).
- `AuthThrottledError` is deliberately **not** a subclass of `AuthSessionExpiredError`: dozens of callers do `if (err instanceof AuthSessionExpiredError) return;`, and inheriting it would make all of them swallow a throttle silently — exactly the silent variant. For the same reason the 401 path throws rather than returning the bare 401, which `lib/errorMessages.ts` would have labelled "Session expired".
- Copy says **"You're still signed in"**, never "expired" — because it hadn't.

## Verification

**Reporter's exact repro (11 navigations, ~20 s), re-run against the fix** — zero 429s, zero logouts:

```
NAV  1 /devices  @t+2s  -> /devices        …
NAV 11 /integrations @t+20s -> /integrations
```

**Forced-throttle scenario** (`AUTH_REFRESH_RATE_LIMIT=4`) to prove the *handling*, not just the raised ceiling:

```
NAV  5 /reports      @t+8s  -> /reports       [THROTTLE MASK VISIBLE]   REFRESH 429
NAV  6 /analytics    @t+10s -> /analytics     [THROTTLE MASK VISIBLE]   REFRESH 429
…
NAV 11 /integrations @t+18s -> /integrations  [THROTTLE MASK VISIBLE]   REFRESH 429
```

Seven consecutive throttled navigations: session survived every one, the mask was shown on each (never a silent blank page), and **exactly one refresh per page load** instead of three.

- `apps/api`: 92 tests green (`rate-limit.test.ts`, `login.test.ts`).
- `apps/web`: 220 tests green (auth store, auth components, locale parity).
- `tsc --noEmit` clean for both apps; `astro check` 0 errors.

## Notes for review

- New strings are translated across all 8 locales (`localeParity` green).
- `refundOnReject` **defaults to false**, so every other limiter (login, per-account lockout) keeps its punitive behaviour unchanged — there's an explicit non-regression test for that.
- The ungated-sibling-island architecture in `DashboardLayout.astro` is left alone deliberately; making the overlay authoritative fixes this bug, and the layout redesign is a much broader change.
- Out of scope, flagged: `handleSessionExpired` still says "Your session expired" on a genuine `transient` (5xx) eviction. That overlaps #3704.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (all three in parallel), plus an independent design review from Codex (`xhigh`, read-only) before implementation.

**Findings:** 6 raised → all 6 addressed in `7a6c8263`; 0 outstanding. Plus one CI regression the reviewers didn't catch but the full local suite did.

| # | Source | Finding | Resolution |
|---|---|---|---|
| 0 | full local API suite | **Test API was red.** `routes/auth.test.ts` is a *second* file exercising `POST /auth/refresh`; its `vi.mock('../services')` didn't stub the new getters, so the handler threw and 10 refresh tests 500'd. | Getters stubbed. Test API now green. |
| 1 | pr-test-analyzer | **Mutation-proven test gap.** Moving AuthOverlay's throttle branch below the `fadeState === 'hidden'` return kept all 20 tests green — yet that ordering is exactly what prevents symptom (b), the fully-painted-but-empty page. Every throttle test started with `tokens: null`, so the fade lifecycle never ran. | Added a test that drives the overlay through `fadeOverlayOut()` first. **I re-ran the reviewer's mutation and confirmed the new test fails under it** and passes when restored. |
| 2 | all three | `AuthGuard`/`DashboardWrapper` left `isRecovering` true forever on a throttle, and AuthGuard's pre-existing 10s safety timer cleared it and fell through to `navigateTo('/login')` — re-creating this very bug on a wait that can legitimately run to 90s. | Both clear `isRecovering`, gate the redirect on `authThrottledUntil`, and the safety timer leaves recovery armed while throttled (matching AuthOverlay). Dead code today, but the PR touched it — shouldn't ship as a half-fixed trap. |
| 3 | code-reviewer + silent-failure-hunter | `ForcedMfaSetupPage` labelled a throttle "Network error" — on a security-sensitive enrollment page, for something that resolves itself. AuthLayout mounts no AuthOverlay, so no mask there either. | Handles `AuthThrottledError` explicitly with honest copy (`common.refreshThrottledError`, all 8 locales). |
| 4 | silent-failure-hunter | **My `AuthThrottledError` doc comment over-claimed** that the mask makes the state "never invisible". False for the bare `Layout.astro`/`AuthLayout.astro` shells. | Comment now states the real scope and names the untreated pages. |
| 5 | silent-failure-hunter | `parseRetryAfterMs` silently fell back to 60s, hiding a proxy stripping `Retry-After` or an operator-retuned window. | Warns on fallback. |

**Explicitly cleared by review:** the limiter's security posture (`refundOnReject` is opt-in at exactly one call site; every other limiter keeps punitive behaviour, with a non-regression test); no unbounded wait or permanently-stuck mask; and my own top concern — **the up-to-90s throttle wait does *not* hold the cross-tab Web Lock**, because the throttled result returns from inside `withRefreshLock` before the sleep. I verified that one from the source myself as well as via the reviewer.

**Deliberately out of scope** (flagged, not fixed): mounting `AuthOverlay` on the four `pages/remote/**` viewers and the `catch {}` in `ProxyTunnelPage.pollStatus` / `RemoteToolsPage.fetchDevice`. Putting a `fixed inset-0` mask over a live video/terminal surface needs its own design pass. Also unchanged: `handleSessionExpired` still says "Your session expired" on a genuine 5xx eviction — overlaps #3704.

**Tests:** `apps/web` 304 green (auth store, auth + layout components, locale parity); `apps/api` full unit suite clean of auth failures, `routes/auth.test.ts` 83 green; `tsc --noEmit` clean both apps; `astro check` 0 errors. **CI: 54 passing, 0 failing**, including all 4 Integration Test shards.

**Live re-verification against the final commit** — the reporter's exact repro (11 navigations, ~20s) completes with zero 429s and zero logouts; and with the budget forced down (`AUTH_REFRESH_RATE_LIMIT=4`) seven consecutive throttled navigations kept the session alive, showed the mask every time, and fired **one** refresh per page load instead of three.

PR #3998, opened 2026-08-25
fix(api): unwrap the SQLSTATE in tenantCascade's isUndefinedTable (#3927)

### LINKED ISSUE #3927: tenantCascade isUndefinedTable reads the SQLSTATE off the wrapper — a missing optional table aborts org erasure and partner purge instead of being tolerated
## Description

Same root cause as #3881 (fixed in #3921), but in the org-erasure path, where the consequence is behavioural rather than cosmetic.

`isUndefinedTable` in `apps/api/src/services/tenantCascade.ts:813` reads the SQLSTATE off the **top-level** error:

```ts
function isUndefinedTable(err: unknown): boolean {
  const code = (err as { code?: string })?.code;
  // Postgres SQLSTATE 42P01 = undefined_table
  return code === '42P01';
}
```

The errors it inspects come from `dbModule.db.execute(...)`, and `drizzle-orm/postgres-js` rethrows the postgres-js `PostgresError` wrapped in a `DrizzleQueryError` whose own `.code` is `undefined` — the SQLSTATE lives on `.cause`. So `isUndefinedTable` returns `false` for a genuinely missing table.

## Why it matters

Both call sites invert a deliberate tolerate-and-continue into abort-the-whole-operation:

- **`:687`** — org erasure, `ASSOCIATED_SYSTEM_SCOPED_TABLES` pre-clear loop. The comment states the intent explicitly:
  > Tolerate missing tables (e.g. a deployment that doesn't have every optional table). Anything else aborts the erasure — record it forensically first (#2195)

  With the wrapped error, a deployment missing an optional table now takes the `!isUndefinedTable` branch: it writes an **erasure-failed audit** via `writeErasureFailedAudit` and throws, aborting the GDPR org erasure partway through.

- **`:1007`** — the same shape in the partner purge pre-clear loop (`writePurgeFailedAudit` + throw).

So on any deployment where one of these optional tables is absent, org erasure and partner purge fail — and fail loudly, with a forensic audit row claiming a real failure.

## Fix

Route it through the canonical helper `pgErrorCode()` in `apps/api/src/utils/pgErrors.ts` (added in #3760, and what #3921 used for the `softwareInstallMethods` mappers):

```ts
function isUndefinedTable(err: unknown): boolean {
  return pgErrorCode(err) === '42P01';
}
```

Add a test using the real wrapped shape — a flat `{ code: '42P01' }` fixture passes against the broken check and proves nothing (that is exactly how the two dead branches in #3881 survived their own tests).

## Related latent sites (cosmetic, lower priority)

The #3881 sweep found three more top-level SQLSTATE reads, all leaking a 500 where a mapped response was intended:

- `apps/api/src/routes/aiAgents.ts:95` — `23505` → 500 instead of 409
- `apps/api/src/routes/networkBaselines.ts:542` — `23503` → 500
- `apps/api/src/routes/agents/connections.ts:101` — diagnostic-only: every logged `pgCode` / `pgDetail` / `pgTable` field is permanently `undefined`, so the diagnostics the comment promises never fire

Worth a follow-up sweep for `(err as { code?: string }).code` repo-wide, plus a lint rule or a shared-helper convention so new call sites can't reintroduce it.

Refs #3881, #3921, #3760. Adjacent to #3880 (cascadeDeleteOrg has no end-to-end integration coverage), which is why this survived.

bdunncompany:
Closes #3927.

`isUndefinedTable` read the SQLSTATE off the **top-level** error. These errors come from `db.execute(...)`, and `drizzle-orm/postgres-js` rethrows the driver error wrapped in a `DrizzleQueryError` whose own `.code` is `undefined` — the SQLSTATE is on `.cause`. So it returned `false` for a genuinely missing table, which **inverts the helper's purpose**: both call sites use it to tolerate an optional table a deployment does not have, so a `false` turns *skip it and carry on* into *write an erasure-failed forensic audit and throw*, aborting a GDPR org erasure or a partner purge partway through.

Routed through the canonical `pgErrorCode()`, as the issue proposed.

### The test that existed proved nothing

There was already a test named `tolerates a missing associated system-scoped table (42P01)`. It used a **flat** `err.code = '42P01'` fixture, which passes whether the helper unwraps or not — exactly what the issue predicted. It is kept (renamed to say `FLAT shape`) and a second test added using the shape production actually throws: own `.code` undefined, `cause.code = '42P01'`.

Control, run both ways:

```
# with the top-level read restored
× tolerates a missing associated table when the SQLSTATE is WRAPPED (DrizzleQueryError)
  Error: [tenantCascade] DELETE from "device_commands" failed for org=…
  Tests  1 failed | 16 passed        <- the FLAT test still passed
```

That error text is the production failure itself: the erasure aborting on a table that is merely absent.

### On the three "latent sites" in the issue — one is included, two are not, and the reason matters

**Included:** `routes/agents/connections.ts`. It reads diagnostics off the top-level error, so every logged `pgCode` / `pgDetail` / `pgTable` / `pgColumn` / `pgConstraint` was permanently `undefined` — the diagnostics that block promises never fired. Now uses `pgErrorNode()` so all of them come from the same node. (Logging is not transactional, so this one genuinely works. Note it now reports the inner Postgres message rather than the outer query-and-params text; `message` itself was never missing.)

**Deliberately NOT included: `aiAgents.ts` (23505 → 409) and `networkBaselines.ts` (23503 → 409).** Unwrapping the SQLSTATE there does **not** deliver the 409, because a second, independent defect sits underneath: those routes run inside the request-wide `withDbAccessContext` transaction, and postgres.js re-throws the rejected statement at commit even after JavaScript has caught it and returned a response. The mapping is not wrong — it is unreachable. This repo already documents exactly that, twice:

> `routes/networkBaselines.ts` — *"ON CONFLICT DO NOTHING instead of catch-and-map: the request runs inside the withDbAccessContext transaction, and postgres.js re-throws a raised unique violation at commit time even after it's caught here, turning the mapped 409 back into a raw 500."*

> `services/aiAgents/agentService.ts` — *"Letting the insert trip 23505 is not an option here: the whole request runs inside one withDbAccessContext transaction, so an in-statement error poisons it and the COMMIT 500s."*

So those two need `ON CONFLICT` / a pre-check / a savepoint, not an unwrap. I started both changes and reverted them rather than ship something that reads as a fix and alters nothing. Worth noting a mocked route test would happily confirm the 409, because a mock has no transaction to poison.

### On the suggested repo-wide sweep

`grep "as { code?: string }"` outside tests returns four more hits, and **three of them must not be changed**: `middleware/bearerTokenAuth.ts`, `services/cfAccessJwt.ts` and `services/clientAiEntraJwt.ts` each wrap `jwtVerify` and read **jose** error codes (`ERR_JWS_*` / `ERR_JWT_*`), not Postgres. `pgErrorCode` is the wrong abstraction there. A blind sweep or a lint rule keyed on the cast shape would break them — worth scoping any future rule to files that actually touch the driver.

### Verification

`tsc --noEmit` clean. Full API unit suite **25,210 passed, 0 failed**. Both controls above run and observed failing before the fix.

### DIFF
--- a/apps/api/src/routes/agents/connections.ts
+++ b/apps/api/src/routes/agents/connections.ts
@@ -7,6 +7,7 @@ import { devices, deviceConnections } from '../../db/schema';
 import { captureException } from '../../services/sentry';
 import { requireAgentRole } from '../../middleware/requireAgentRole';
 import { submitConnectionsSchema } from './schemas';
+import { pgErrorNode } from '../../utils/pgErrors';
 
 // Hard caps that match the device_connections column widths. Agent
 // collectors emit kernel-bound values (TCP states, /proc comm names) that
@@ -98,7 +99,15 @@ connectionsRoutes.put(
     // Global onError returns a generic 500; re-log with pg error fields
     // (code/constraint/column) so server logs retain diagnostic context,
     // and capture to Sentry for durability.
-    const pg = err as { code?: string; detail?: string; table_name?: string; column_name?: string; constraint_name?: string; message?: string };
+    // `pgErrorNode` returns the error that actually CARRIES the SQLSTATE, so
+    // code/detail/table_name all come from the same object. Reading them off
+    // the top-level error left every POSTGRES DIAGNOSTIC field below
+    // permanently `undefined` under drizzle-orm/postgres-js — code, detail,
+    // table_name, column_name, constraint_name. (`message` was never missing:
+    // the outer DrizzleQueryError has its own. Note it now reports the inner
+    // Postgres message rather than the outer query-and-params text, which is
+    // the more useful half for these diagnostics.)
+    const pg = (pgErrorNode(err) ?? {}) as { code?: string; detail?: string; table_name?: string; column_name?: string; constraint_name?: string; message?: string };
     console.error('connections-inventory insert failed', {
       agentId,
       deviceId: device.id,
--- a/apps/api/src/services/tenantCascade.test.ts
+++ b/apps/api/src/services/tenantCascade.test.ts
@@ -207,7 +207,11 @@ describe('cascadeDeleteOrg', () => {
     expect(stats.totalRowsDeleted).toBe(5 + 3 * cascadeOrder.length);
   });
 
-  it('tolerates a missing associated system-scoped table (42P01)', async () => {
+  it('tolerates a missing associated system-scoped table (42P01, FLAT shape)', async () => {
+    // The flat shape — SQLSTATE on the error itself. Kept, but note it does
+    // NOT exercise what production actually throws: see the wrapped-shape test
+    // below, which is the one that fails against a top-level `.code` read.
+    //
     // Override the default mock to make ONLY the device_commands DELETE
     // throw 42P01; everything else returns 0.
     vi.mocked(db.execute).mockImplementation(((q: unknown) => {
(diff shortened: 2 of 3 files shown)


ToddHebebrand (approved):
**Correct fix, and the right helper.** Verified the mechanism rather than taking the description: `pgErrorNode`/`pgErrorCode` in `apps/api/src/utils/pgErrors.ts` already document this exact wrapping (`DrizzleQueryError.code` undefined, SQLSTATE on `.cause`), so reaching for the existing helper instead of hand-rolling another unwrap is the right call.

**The wrapped-shape test is the valuable half.** You called out that the flat fixture passes whether `isUndefinedTable` unwraps or not — that is exactly right, and it is the reason the original bug survived having a test at all. Keeping both, with the comment explaining that the flat one proves nothing on its own, is better than replacing it.

**One thing your sweep missed — the same broken read is live in two more places on `main`.** Not a blocker for this PR; filing separately. Both bypass `pgErrors.ts` with a top-level `.code`:

- `apps/api/src/routes/networkBaselines.ts:542-543` — `const pgError = error as { code?: string }; if (!deleteChanges && pgError.code === '23503')`. The error comes from `db.delete(...)`, so `.code` is undefined and the branch never fires: the intended 409 with the `?deleteChanges=true` hint degrades to a generic 500.
- `apps/api/src/routes/aiAgents.ts:103` — `isUniqueViolation` reads the top-level `.code === '23505'` for the create race the pre-check cannot win. Same miss, and notably it is a re-implementation of `isPgUniqueViolation`, which handles the cause chain correctly.

Merging.

PR #3999, opened 2026-08-25
fix(web): un-nest the AutomationTab delete button from the row expander (#3990)

### LINKED ISSUE #3990: [UI] Nested <button> in AutomationTab causes a React hydration error and makes delete also toggle the row
## Description

Opening **Configuration Policies → (a policy) → Automations → Add Automation** logs a React DOM-nesting error:

```
In HTML, <button> cannot be a descendant of <button>.
This will cause a hydration error.

  <ConfigPolicyDetailPage>
    <AutomationTab>
      <FeatureTabShell title="Automations" ...>
>       <button type="button" onClick={...} className="flex w-full items-center justify-between px-4 py-3 text-left">
          <div>
>         <button type="button" onClick={...} className="flex h-7 w-7 items-center justify-center rounded-md text-destructive ...">
```

The row's **delete (trash)** button is nested inside the row's **expand/collapse** button.

## Impact

1. Invalid HTML that React explicitly flags as a hydration hazard.
2. Real interaction bug: a click on the inner delete button also bubbles to the outer expand handler, so deleting an automation row simultaneously toggles the row. Any `stopPropagation` currently masking this is load-bearing and fragile.
3. Accessibility: nested interactive controls are not reliably reachable or announced by screen readers and keyboard navigation.

## Steps to reproduce

1. Create a configuration policy.
2. Open it → **Automations** tab → **Add Automation**.
3. Observe the console error above.

Verified against `main` @ `c2f9db1ce`.

## Proposed fix

Restructure the row so the expand control and the delete control are siblings, not nested — e.g. make the row a `<div>` with an explicit expand `<button>` and a separate delete `<button>`, or render the expand affordance as the row header with the delete button positioned outside it.

## Affected files

- `apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.tsx`

bdunncompany:
Closes #3990.

The row's **delete** control was a `<button>` nested inside the row's **expand** `<button>`. Three consequences, only one of them cosmetic:

1. Invalid HTML that React explicitly flags as a hydration hazard.
2. A click on delete bubbled to the expand handler, so deleting a row also toggled it.
3. Nested interactive controls are not reliably reachable or announced by keyboard and screen readers.

The expand and delete controls are now **siblings** inside a plain `<div>`, as the issue proposed.

### The `stopPropagation` was load-bearing

The bubbling in (2) was masked by an `e.stopPropagation()` in the delete handler. That is why this survived: the symptom was invisible while the invalid markup stayed. With siblings there is nothing to stop, so the call is gone rather than kept "just in case" — leaving it would have preserved the thing that hid the bug.

### Keeping the full-row hit target

Worth calling out because the obvious restructure silently regresses it. Moving the old button's `px-4 py-3` onto the new wrapper `<div>` leaves the disclosure control at roughly text height inside a ~46px header, with the surrounding padding dead space — visually identical, materially worse to click. The padding therefore lives on the **expand button**, with `self-stretch` so it fills the header height; the wrapper keeps only the right padding for the delete control.

### Accessibility

`aria-expanded` on the expand control, and the icon-only delete button is named **`Delete: <automation name>`**, following the existing `ConfigPolicyList` precedent — a bare "Delete" leaves every row's button identical in a screen-reader button list. It reuses the existing `common:actions.delete` key plus the automation's own name, so there are no new locale entries and nothing for `localeParity.test.ts` to catch across the 7 translated locales.

### Tests, and one that was a false positive

Two tests: **structural** (the delete button has no `<button>` ancestor) and **behavioural** (deleting a collapsed row leaves the expanded row expanded, with its form still mounted).

The behavioural one is deliberately sequenced. An earlier version expanded row 0 implicitly, added a second row, deleted row 0 and counted survivors — and that **passes with the bug present**: `deleteItem` shifts `expandedIndex` from 1 to 0 when the row below is removed, and a leaked row-0 expand handler also sets 0, so the two outcomes are indistinguishable. It only went red in the control because the old markup had no `aria-expanded` for the query to find, i.e. for the wrong reason entirely.

Expanding row 0 and deleting the **collapsed** row 1 separates them: correct code leaves `expandedIndex` at 0, whereas bubbling toggles row 1 open first and `deleteItem(1)` then clears it to null, collapsing the survivor.

Control, with the leak reintroduced and the DOM otherwise untouched, so the red is about bubbling and not a missing query target:

```
× deleting a collapsed row leaves the expanded row expanded
  Tests  1 failed | 7 passed
```

Also worth recording: my first structural assertion was `expect(del.closest('button')).toBe(del)`, which **can never fail** — `closest()` matches the element it starts from, so it passed with the nesting fully restored. It now walks up from `parentElement`.

### Verification

Full web suite green, `eslint src` clean, node 22.23.2 per the repo `.nvmrc`. Both controls above were run and observed failing before the fix.

### DIFF
--- a/apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.test.tsx
+++ b/apps/web/src/components/configurationPolicies/featureTabs/AutomationTab.test.tsx
@@ -165,6 +165,68 @@ describe('AutomationTab — schedule timezone picker', () => {
     }
   });
 
+  /**
+   * #3990 — the row's delete control used to be a <button> NESTED inside the
+   * row's expand <button>.
+   *
+   * Three things were wrong and only one of them was cosmetic: invalid HTML
+   * React flags as a hydration hazard; a click on delete bubbling to the expand
+   * handler so removing a row also toggled it; and nested interactive controls
+   * that keyboard and screen-reader users cannot reach reliably.
+   *
+   * The bubbling was masked by an `e.stopPropagation()` in the delete handler,
+   * which made the symptom invisible while leaving the invalid markup — so this
+   * asserts the STRUCTURE, not just the behaviour. A test that only clicked
+   * delete and checked the row vanished would pass with the nesting restored
+   * and stopPropagation back in place.
+   */
+  it('renders the expand and delete controls as siblings, not nested', () => {
+    addAutomation();
+
+    const del = screen.getAllByRole('button', { name: /^Delete:/i })[0];
+    // Walk up from the PARENT. `del.closest('button')` matches the delete
+    // button itself, so it can never fail — that version of this assertion
+    // passed with the nesting fully restored.
+    expect(del.parentElement?.closest('button')).toBeNull();
+  });
+
+  /**
+   * Deleting a COLLAPSED row must not disturb the expanded one.
+   *
+   * The sequence matters, and an earlier version of this test got it wrong in a
+   * way that made it a false positive: it expanded row 0 implicitly, added a
+   * second row, deleted row 0, and counted the survivors. That passes with the
+   * nesting restored — `deleteItem` shifts `expandedIndex` from 1 to 0 when the
+   * row below is removed, and a LEAKED row-0 expand handler also sets 0, so the
+   * two outcomes are indistinguishable. It only went red in the control because
+   * the old markup had no `aria-expanded` for the query to find, i.e. for the
+   * wrong reason entirely.
+   *
+   * Expanding row 0 and deleting the COLLAPSED row 1 separates them: with
+   * correct code `expandedIndex` stays 0 and row 0 keeps its form mounted; with
+   * bubbling, the delete click first toggles row 1 open (setting the index to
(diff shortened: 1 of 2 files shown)


ToddHebebrand (approved):
**Good fix, and the tests are the reason I'm merging it without a round.** The structural assertion is the part that matters — walking up from `del.parentElement` rather than `del.closest('button')` is the difference between an assertion that can fail and one that cannot, and you flagged that yourself.

The second test is the one I'd have asked for: deleting the *collapsed* row separates "expandedIndex shifted because a row below was removed" from "a leaked expand handler fired," which the obvious version of that test conflates. Documenting why the earlier version was a false positive is more useful than the test itself.

**Verified the i18n keys exist** rather than assuming: `policies:...automationTab.untitledAutomation` is present in all eight locales on `main`, and `common:actions.delete` resolves — so no locale JSON was needed in this diff and parity stays intact.

Keeping the padding on the button rather than the wrapper (`self-stretch py-3 pl-4`) preserves the full-row hit target that the old nesting gave for free. Easy detail to lose in this kind of un-nesting; you didn't.

Merging.

PR #4003, opened 2026-08-25
fix(web): surface the API's own error on the Organizations page mutations (#3989)

### LINKED ISSUE #3989: [UI] Org and notification-channel create discard the API's specific error and show a generic message
## Description

Two independent create flows throw away the specific error the API returns and show a generic string instead. The user is told *that* something failed but never *what*, even though the API said exactly what.

Verified in a browser against `main` @ `c2f9db1ce`.

### 1. Organization create — 409 message discarded

Creating an org with a slug that already exists:

- **API actual:** `POST /api/v1/orgs/organizations` → `409` `{"error":"That organization slug is already in use"}`
- **UI actual:** `Failed to save organization`

Root cause is exact — `apps/web/src/components/settings/OrganizationsPage.tsx:324`:

```ts
if (!response.ok) {
  throw new Error(t('organizationsPage.errors.saveOrganization'));
}
```

The response body is never read on failure, so **409 / 400 / 403 / 500 all collapse to the same sentence**. `OrganizationsPage.tsx` contains zero `runAction` calls (`grep -c runAction` = 0) and is not listed in `runActionAllowlist.ts`, despite the CLAUDE.md contract "Mutation handlers must surface outcome via `runAction`". `runAction` already has `extractApiError()`, which would have surfaced the real message.

### 2. Notification channel create — `details[]` discarded

- **API actual:** `POST /api/v1/alerts/channels` → `400` `{"error":"Invalid webhook channel configuration","details":["Webhook URL must use HTTPS","Webhook URL hostname is required"]}`
- and for a loopback URL → `400` `{"error":"Invalid webhook channel configuration","details":["Webhook URL cannot target loopback, link-local, metadata, or CGNAT addresses"]}`
- **UI actual (both cases):** `Invalid webhook channel configuration`

The `details[]` array — which names the offending field and rule — is dropped entirely. The user sees the same sentence whether the scheme is wrong, the hostname is missing, or SSRF protection rejected the target.

(The underlying validation is correct and fails closed in both cases — `javascript:alert(1)` was rejected and nothing persisted. This issue is purely about surfacing the reason.)

## Impact

Both are dead ends: the form stays open with no indication of which field to fix. The org case is worse because the fix ("pick a different slug") is trivial and completely knowable — the API already said it.

## Proposed fix

- Route both handlers through `runAction` so `extractApiError()` surfaces the API message.
- Where the API returns `details[]`, render them as field-level errors rather than only the top-level `error`.

## Affected files

- `apps/web/src/components/settings/OrganizationsPage.tsx` (line ~324, plus the sibling `catch` blocks at ~127/~362/~389/~458/~476 which use the same generic-fallback pattern)
- the notification-channel create/update handler under `apps/web/src/components/alerts/`
- `apps/web/src/lib/runAction.ts` (`extractApiError`) — already correct, just not used here

Related: #3531 (destructive confirmations hiding failures) is the same "bare fetch instead of `runAction`" family, but a distinct symptom — that one hides failure entirely, this one reports failure without the reason.

bdunncompany:
Closes #3989.

Two flows discarded the specific error the API returned. Investigating turned up that **each half of the issue was right about a different thing, and I got the first attempt wrong** — recording that here because it shaped the diff.

### Part 1 — `runAction` alone would not have fixed the channel case

The issue notes that `runAction` "already has `extractApiError()`, which would have surfaced the real message". True for the org 409. **Not** true for the channel 400:

```jsonc
{ "error": "Invalid webhook channel configuration",
  "details": ["Webhook URL must use HTTPS", "Webhook URL hostname is required"] }
```

`extractApiError` returned only the generic sentence. `detailsToString` tries a string, then `joinZodIssues` — which requires objects carrying `.message` and returns `null` for strings — then `joinZodFlatten` (`{formErrors, fieldErrors}`). A plain `string[]` matches none, so `details` was dropped before any caller saw it.

Added a `joinStringArray` branch **after** the zod branches, so an issues array still renders through its own path. Control:

```
× renders a plain string[] details alongside the top-level error
× ignores non-string entries rather than printing [object Object]
× surfaces the details[] reasons in the error toast, not just the generic error
  Tests  3 failed | 51 passed
```

That third one is end-to-end through the channel save path. The existing channel test asserted only `{ type: 'error' }`, so the whole regression could return while it stayed green.

### Part 2 — my first attempt made the message correct and still invisible

I initially read the body into `setError` and argued against converting to `runAction`, on the grounds that it would duplicate an inline form error. **That was wrong, and I had not checked it.** In this file:

- the error banner renders at `OrganizationsPage.tsx:579`, **outside** the modals
- every modal is `fixed inset-0 z-50` (lines 791/811/841/895) and covers it
- **nothing renders inside any dialog**
- the modal closes only on success

So a 409 landed underneath the form the user is still looking at. The pre-existing site-save handler has the same shape — it reads the body correctly and puts the result somewhere invisible.

Toasts do clear it: `DashboardLayout.astro` renders `<slot />` at line 56 and `<ToastContainer>` at line 62, so at equal `z-50` the toast is later in DOM order and paints above the overlay.

**So all five mutations now go through `runAction`:** org create, org delete, site save, site delete — and the drag-reorder `PATCH`, which was **entirely silent**. That one set the page error and then called `fetchOrganizations()` to revert the optimistic order; that function calls `setError(undefined)` before its first `await`, so React batches the two updates and renders no failure at all.

`OrganizationsPage.tsx` is added to the guarded set in `no-silent-mutations.test.ts` so it cannot regress, which required bumping that suite's file count 96 → 97.

This also settles the contract point: the file previously had zero `runAction` calls, was not in `runActionAllowlist.ts`, and was not in the guarded target set — so CI blessed the `AGENTS.md` violation by omission.

### Two further fixes from review

**`onUnauthorized` on every call.** `runAction` deliberately does not toast a 401 — it relies on that callback for navigation. Omitting it meant a 401 fell into the catch and back into the page banner behind the modal: silent again, the exact bug this PR exists to remove. All five now pass `handleSessionExpired` — **not** a bare `navigateTo('/login')`, which is what
the rest of this file and ~44 other call sites do. A bare navigate races and clobbers a richer
destination: `fetchWithAuth` funnels its own expiry 401s through `handleSessionExpired`, which has
already called `window.location.replace(loginPathWithNext() + '&reason=...')`, so a second bare
navigation drops both the `next` and the `reason`. Passing the function itself is correct in both
directions — it early-returns on `sessionExpiryInFlight` when an expiry is already redirecting, and
on a 401 that is *not* session expiry it performs the real logout the bare navigate only implied.
**Changing the five callbacks did not close every 401 path on the page.** `fetchOrganizations` had
its own bare navigate, and create/delete call it immediately after a successful mutation through
`refreshOrgs` — so a 401 on that follow-up GET produced exactly the clobber the callbacks had just
been fixed to avoid. There is a second variant with no redirect at all behind it: when a token
refresh *succeeds* but the replayed GET still returns 401, `fetchWithAuth` returns that response
without calling `handleSessionExpired`, so the bare navigation was the only redirect and `next` and
`reason` were simply lost. That branch now calls `handleSessionExpired` too.

The `BulkOrgImport` prop passed a bare navigate too, and it is reached from this page's own bulk
mutations, so it now passes `handleSessionExpired` as well. The four modal catches now only `setError` for a **non**-`ActionError` escape, since `runAction` has already toasted anything else — otherwise every failure had a second, invisible copy. `fetchOrganizations`' catch is untouched: it is a GET, and its banner is not behind a modal.

**The failed reorder reconciles with the server, and getting there took three attempts.** The
reorder is optimistic, so a failure leaves the user looking at an order the server may not have.

1. I first *skipped* reconciliation for `ActionError` status 0, to dodge a race between the revert
   GET and the login redirect. That was worse: `runAction` collapses any request-side throw to 0, so
   an ordinary dropped connection took the skip path and left a wrong order on screen indefinitely,
   corrected only by a page load nothing guarantees.
2. I then captured the pre-drag array and restored it locally. That fixes the visible symptom and is
   still wrong, for a reason worth stating plainly: **status 0 does not mean the server rejected the
   write.** A `PATCH` that committed and then lost its response is indistinguishable, client-side,
   from one that never arrived. Restoring locally would show an order the server does not have —
   the same lie in the other direction. It also races an in-flight authoritative GET and can clobber
   newer data with an older snapshot, and it needed a sequence guard for overlapping drags that
   still could not produce a correct answer when both requests failed.
3. Only the server knows what persisted, so the catch asks it: `void fetchOrganizations()`.

The refetch was never the problem. It was unsafe **only** because `fetchOrganizations` answered its
own 401 with a bare `navigateTo('/login')` — the very bug fixed above. With that gone, a second GET
is harmless, and reconciliation needs no status special-casing and no sequence guard: whichever
refetch lands last reports the server's real order, whatever combination of requests committed.

**Reorders are serialized, which is what actually removes the stale-order races.** Every one of them
needs a *second* drag to begin while the first request or its reconciling GET is still in flight, so
dragging is disabled from the moment a reorder is issued until it settles (`reorderPending`).

I reached that by discarding two worse answers. A client generation counter (bump per load and per
committed reorder, discard stale loads) was implemented and **reverted**: it threw away genuinely
current create/delete/import refreshes — a newly created org could stay invisible with no
replacement fetch — while still not covering a GET that resolves during an in-flight PATCH. And the
overlap had previously been blocked only as an *accident* of the full-page loading spinner
unmounting the draggable rows, which the silent reconciliation above deliberately stops doing; so
"this race predates the PR" was not quite true, and serializing is what makes it true again.

What is still not fixed, and is written at the call site: two reorders from **separate tabs** can
interleave, because the endpoint fully replaces the stored order with no revision or
compare-and-swap.

**Pagination.** `GET /orgs/organizations` pages by `created_at, id` (`LIMIT`/`OFFSET`) and applies
the partner's `organizationOrder` **only within the page it just selected**, so past the first page
it cannot express a cross-page order at all. That affects the ordinary list render, not just
reconciliation. Pre-existing, not fixed here — filed as #4004.

**Reconciliation is silent.** `loading` is an early return that swaps the whole page for a spinner —
right for the first load, wrong for a background correction. Because the refetch now runs on *every*
reorder failure rather than rarely, it blanked the list at the exact moment the failure toast
appeared, taking the user's scroll position and selected org with it. `fetchOrganizations` takes a
`{ silent }` option that skips only the loading flag. `tsc` caught a related hazard while adding it:
the retry button passed `fetchOrganizations` directly as `onClick`, so a click event would have
arrived as the options object — fixed at the call site rather than by widening the parameter type.

**The `saveSite` fallback string.** `organizationsPage.errors.saveSite` interpolates `{{status}}`, which `runAction` does not expose when building a fallback — passing `0` renders "Failed to save site (0)". It now uses the existing status-free sibling `siteDetailPage.errors.saveSite`, which is present in all 8 locales, so no new keys and nothing for locale parity to catch.

**The reorder's revert refetch on 401.** `persistOrganizationOrder` re-fetches the authoritative order after a failure. On a 401 that second authenticated GET races the `onUnauthorized` redirect and can start its own bare `/login` navigation from the fetch's own 401 branch, dropping the destination the session-expiry redirect intended to carry. It no longer needs a status guard at all: reconciliation is unconditional and authoritative.

### Two more fixes from review

**Mixed `details` arrays lost half their messages.** With a zod-issues branch tried before a
string branch, `details: [{ message: 'name is required' }, 'Webhook URL must use HTTPS']` rendered
only the object's message and silently dropped the string — the same class of loss this PR exists
to remove, reintroduced by my own fix's branch ordering. Nothing guarantees a route emits a
homogeneous array. The two branches are now a single pass that accepts both representations, which
is also less code. Equivalent for a pure-object array, because `joinZodIssues` reads only
`.message` and ignores `path`. Control: restoring the ordered branches reds the new mixed test and
nothing else (`1 failed | 38 passed`).

**Unexpected modal errors went to the known-invisible banner.** The four modal catches fell back to
`setError` for a non-`ActionError` — the same banner the comment two lines above calls invisible.
Reachable in practice: `runAction` invokes `onUnauthorized` *outside* its request try/catch, so a
throw from `handleSessionExpired`'s `logout()` or `location.replace` arrives there. They toast now.

### One red the fix caused, and the control on the repair

Adding the `handleSessionExpired` import broke four tests in
`OrganizationsPage.firstSite.test.tsx` — a **pre-existing** file this diff otherwise does not touch.
It mocks `../../stores/auth` with a factory returning only `fetchWithAuth`, and Vitest throws on any
export the factory omits. So merely *building* the `runAction` options object threw
`No "handleSessionExpired" export is defined on the mock` before a request was ever made, and the
create flow appeared to silently do nothing — which looks nothing like a missing mock entry. The
factory now returns it, with a comment saying why.

The repair is test-infrastructure, so the useful question is whether the test still discriminates.
Control — break only the nag decision (`existingSites?.length === 0` → `=== -1`), smallest possible
mutation, query surface untouched:

```
Tests  1 failed | 3 passed (4)
```

Exactly the nag test reds and the other three stay green. That rules out the failure mode where a
control reds because it removed the thing the test queries by, which would have killed all four.

### Verification

- `tsc --noEmit` clean
- `eslint src` clean
- full web suite green (see the run cited below)
- node v22.23.2, matching `apps/web/.nvmrc`

`OrganizationsPage.mutationFeedback.test.tsx` is new. Its reorder tests drive reconciliation
through a mock server whose order the test controls, so they prove the list came from the SERVER
rather than a client snapshot. The important one simulates the ambiguous case — the `PATCH`
commits, then the response is lost — and asserts the UI ends on the server's NEW order; a local
rollback fails it. Each also asserts the optimistic order appeared FIRST, so a regression making
the drag a no-op cannot satisfy the final assertion for free. Each was controlled:

|---|---|---|
| revert to a local snapshot rollback | all three reorder tests red, 401 test green | `3 failed \| 1 passed` |
| restore the bare `navigateTo('/login')` | only the 401 test red | `1 failed \| 2 passed` |
| dedup on the whole joined string | only the dedup test red | `1 failed \| 39 passed` |
| restore the ordered zod/string branches | only the mixed-array test red | `1 failed \| 38 passed` |
| make reconciliation non-silent | only the spinner test red | `1 failed \| 4 passed` |
| remove the `reorderPending` drag gate | only the serialization test red | `1 failed \| 4 passed` |
| per-message dedup only | only the aggregate test red | `1 failed \| 40 passed` |
| whole-string dedup only | only the mixed-array test red | `1 failed \| 40 passed` |

The overlap test waits on the reconciling GET rather than on the PATCH count — the count was
already satisfied before the late rejection, so waiting on it could have finished the test before a
broken rollback ever ran.

**Dedup needs both rules; each alone regresses a real shape.** Per-message matching is required for
`{ error: 'name is required', details: [{message:'name is required'}, 'Webhook URL must use
HTTPS'] }`, where no comparison against the joined string ever matches and the first message echoes.
Whole-string matching is required for `{ error: 'a; b', details: [{message:'a'}, {message:'b'}] }`,
which is the same text split up — no individual message equals the aggregate, so per-message alone
prints all of it twice. Applying both, and removing at most one occurrence per excluded part rather
than every match, satisfies both. The two controls below show each half is load-bearing.

One of these controls initially did **not** discriminate, which is worth recording. The first
version of the spinner test passed with the fix reverted: the spinner exists only *while* the
reconciling GET is in flight, and the assertions ran after it resolved, so the test was checking a
window that had already closed. It now holds the GET open and asserts inside it. A green test whose
control stays green is not evidence of anything.

Every control above was run and observed failing against the broken version before the fix. Gate
artifacts were confirmed newer than every changed source file, so none of the results are stale.

### DIFF
--- a/apps/web/src/components/alerts/NotificationChannelsPage.test.tsx
+++ b/apps/web/src/components/alerts/NotificationChannelsPage.test.tsx
@@ -73,6 +73,45 @@ describe('runChannelSave', () => {
     expect(ON_UNAUTHORIZED).not.toHaveBeenCalled();
   });
 
+  // #3989 — the exact reported symptom. The route answers with the reason in a
+  // plain `details` string array, and the user used to see only the generic
+  // top-level sentence: identical whether the scheme was wrong, the hostname
+  // was missing, or SSRF protection rejected the target.
+  //
+  // Asserts the toast TEXT, not merely that an error toast happened. The test
+  // above checks `{ type: 'error' }` only, so the whole regression could return
+  // while it stayed green.
+  it('surfaces the details[] reasons in the error toast, not just the generic error', async () => {
+    fetchWithAuthMock.mockResolvedValue(
+      makeJsonResponse(
+        {
+          error: 'Invalid webhook channel configuration',
+          details: ['Webhook URL must use HTTPS', 'Webhook URL hostname is required'],
+        },
+        false,
+        400
+      )
+    );
+
+    await expect(
+      runChannelSave(
+        { url: '/alerts/channels', method: 'POST', payload: {}, channelName: '', isCreate: true },
+        { onUnauthorized: ON_UNAUTHORIZED }
+      )
+    ).rejects.toThrow();
+
+    // ONE exact error-toast call, not a join across every call. Joining passes
+    // if the two reasons arrive in separate toasts, or in a success toast —
+    // neither of which is the contract. The extractor's own unit test asserts
+    // the exact string; this asserts the integration actually delivers it.
+    expect(showToastMock).toHaveBeenCalledTimes(1);
+    expect(showToastMock).toHaveBeenCalledWith({
+      message:
+        'Invalid webhook channel configuration: Webhook URL must use HTTPS; Webhook URL hostname is required',
+      type: 'error',
+    });
+  });
+
   it('calls onUnauthorized and does not show an error toast on 401', async () => {
     fetchWithAuthMock.mockResolvedValue(makeJsonResponse({}, false, 401));
     const onUnauthorized = vi.fn();
--- a/apps/web/src/components/settings/OrganizationsPage.firstSite.test.tsx
+++ b/apps/web/src/components/settings/OrganizationsPage.firstSite.test.tsx
@@ -4,8 +4,14 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import OrganizationsPage from './OrganizationsPage';
 import { fetchWithAuth } from '../../stores/auth';
 
+// handleSessionExpired is the `onUnauthorized` callback for every mutation on
+// this page. Mocking the module replaces ALL of its exports, so omitting it
+// makes the mere ACT of building the runAction options throw
+// ("No export is defined on the mock") before any request is made — which
+// reads as the create flow silently doing nothing.
 vi.mock('../../stores/auth', () => ({
-  fetchWithAuth: vi.fn()
+  fetchWithAuth: vi.fn(),
+  handleSessionExpired: vi.fn()
 }));
 
 const navigateTo = vi.fn();
--- a/apps/web/src/components/settings/OrganizationsPage.mutationFeedback.test.tsx
+++ b/apps/web/src/components/settings/OrganizationsPage.mutationFeedback.test.tsx
@@ -1,0 +1,209 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import OrganizationsPage from './OrganizationsPage';
+import { fetchWithAuth, handleSessionExpired } from '../../stores/auth';
+import { showToast } from '../shared/Toast';
+
+vi.mock('../../stores/auth', () => ({
+  fetchWithAuth: vi.fn(),
+  handleSessionExpired: vi.fn()
+}));
+
+const navigateTo = vi.fn();
+vi.mock('@/lib/navigation', () => ({ navigateTo: (...args: unknown[]) => navigateTo(...args) }));
+
+vi.mock('../shared/Toast', () => ({ showToast: vi.fn() }));
+
+const storeFetchOrganizations = vi.fn().mockResolvedValue(undefined);
+vi.mock('../../stores/orgStore', () => ({
+  useOrgStore: { getState: () => ({ fetchOrganizations: storeFetchOrganizations }) }
+}));
+
+const fetchMock = vi.mocked(fetchWithAuth);
+const sessionExpiredMock = vi.mocked(handleSessionExpired);
+const toastMock = vi.mocked(showToast);
+
+const ORG_A = { id: 'aaaaaaaa-1111-4111-8111-111111111111', name: 'Alpha Ltd', status: 'active', deviceCount: 0, createdAt: '2026-01-01T00:00:00Z' };
+const ORG_B = { id: 'bbbbbbbb-2222-4222-8222-222222222222', name: 'Beta Ltd', status: 'active', deviceCount: 0, createdAt: '2026-01-02T00:00:00Z' };
+const ORG_C = { id: 'cccccccc-3333-4333-8333-333333333333', name: 'Gamma Ltd', status: 'active', deviceCount: 0, createdAt: '2026-01-03T00:00:00Z' };
+
+/** Drag the row with `sourceId` onto the row with `targetId`. */
+function drag(sourceId: string, targetId: string) {
+  const source = screen.getByTestId(`org-row-${sourceId}`);
+  const target = screen.getByTestId(`org-row-${targetId}`);
+  const dataTransfer = { effectAllowed: '', setData: vi.fn(), dropEffect: '' };
+  fireEvent.dragStart(source, { dataTransfer });
+  fireEvent.dragOver(target, { dataTransfer });
+  fireEvent.drop(target, { dataTransfer });
+}
+
+const jsonResponse = (payload: unknown, ok = true, status = ok ? 200 : 500): Response =>
+  ({ ok, status, statusText: ok ? 'OK' : 'ERROR', json: vi.fn().mockResolvedValue(payload) }) as unknown as Response;
+
+/** Rendered order of the org rows, top to bottom. */
+function renderedOrgIds(): string[] {
+  return Array.from(document.querySelectorAll('[data-testid^="org-row-"]')).map(
+    el => (el.getAttribute('data-testid') ?? '').replace('org-row-', '')
+  );
+}
+
+beforeEach(() => {
+  fetchMock.mockReset();
+  navigateTo.mockReset();
+  sessionExpiredMock.mockReset();
+  toastMock.mockReset();
+  storeFetchOrganizations.mockClear();
+  window.location.hash = '';
+});
+
+describe('OrganizationsPage — the list GET is not a second, poorer redirect', () => {
+  it('routes a 401 through handleSessionExpired instead of a bare /login navigation', async () => {
+    // fetchWithAuth funnels a real session expiry through handleSessionExpired,
+    // which redirects carrying `next` and `reason`. A bare navigateTo('/login')
+    // here would race that and replace it with a destination-less /login.
+    fetchMock.mockImplementation(async () => jsonResponse({ error: 'Unauthorized' }, false, 401));
+
+    render(<OrganizationsPage />);
+
+    await waitFor(() => expect(sessionExpiredMock).toHaveBeenCalled());
+    expect(navigateTo).not.toHaveBeenCalledWith('/login', expect.anything());
+  });
+});
+
+describe('OrganizationsPage — a failed reorder reconciles with the server', () => {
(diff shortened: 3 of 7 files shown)


ToddHebebrand (approved):
**Approved — and the guard change is the part I want to call out.** You added `OrganizationsPage.tsx` to `TARGET_GLOBS` and moved the count 96 → 97 in `apps/web/src/lib/__tests__/no-silent-mutations.test.ts`. That widens the guarded set rather than carving out an exception, which is the opposite of how a page usually gets "adopted." Stating plainly what the guard does *not* enforce — it checks lexical containment in a `runAction` call, it does not inspect catch blocks — is the kind of scope honesty worth rewarding, because the next person to read it won't over-trust it.

**The `detailsToString` rework is right for a reason that is easy to get wrong.** Trying `joinZodIssues` first and falling back to a string join drops half of a mixed array, since the first branch returns as soon as any entry has a `.message`. One pass over both representations is correct, and the per-message *and* whole-string dedup rules are both load-bearing — either alone regresses a real shape, and you documented which shape each one protects.

**Catch pattern matches the contract** in every handler I checked: 401 handled by `onUnauthorized`, `ActionError` already toasted by `runAction`, and only a non-`ActionError` escape surfaced with `showToast` — a toast rather than `setError`, since the banner renders behind the still-open modal. That last point is the actual user-visible bug and it's fixed consistently across org create, org delete, site save and site delete, not just the one the issue named.

Merging.

PR #4007, opened 2026-08-25
fix: gate admission on readiness and drain offline sweeps

ToddHebebrand:
API admission now checks live consumers required by the process role. Offline sweeps retain their continuation in BullMQ, use stable observation identities, and guard each online-to-offline transition against a concurrent heartbeat. This branch integrates main `a52504b9ee` and preserves the earlier readiness gate, Redis command/blocking connection, pause/shutdown, webhook replacement, and failed-job recovery corrections.

The post-transition crash gap is now addressed: the successful device status change commits pending event and alert work in the same PostgreSQL transaction. The existing offline queue drains it with bounded recovery scans and fenced leases, independently of whether the device is still online. Deterministic alert admission commits child effects atomically. Redis publication, cooldown/flapping updates and correlation enqueue run after database transactions close.

Stable event IDs preserve logical identity through retries; physical publication is at least once. Existing best-effort subscriber processing is not promoted to exactly-once delivery. Reconnects and ownership changes suppress obsolete pending alerts while historical event payloads retain the original organization. Durable cooldown receipts preserve suppression through delayed Redis work, adaptive resolution, and long cooldowns.

The new `2026-10-09-000800-offline-transition-effects.sql` migration adds forced RLS and system-only writes, source immutability, tenant export/erasure registration, and the historical-table exclusion from device-move discovery. Shipped migrations are unchanged. This explicitly extends the original no-migration port.

Validation at `ae83348e2fe499f6a5c9f4a7503318c405404acc`:

- Full API: **34,695 tests passed, 21 skipped; 1,870 files passed, 3 skipped**, exit 0.
- Real PostgreSQL/Redis recovery: **26 tests passed**, including rollback, publication/queue failures, lease expiry, duplicate delivery, reconnects, legacy/policy alerts, ownership moves, erasure, and export boundaries. Existing export/erasure round-trip: 3 passed.
- Organization-merge checks: 14 passed; tenant-cascade: 5; forced-RLS catalog: 100; integration discovery: 5.
- API lint/typecheck, migration naming/immutability, ledger drift (675 files), and repeated migration application passed.
- Removing lease fencing or durable cooldown protection made the targeted database tests fail; restored source passed both controls.
- Independent design and final code/error/test/type/tenancy review found no remaining actionable findings. The original post-CAS lost-effects engineering blocker is addressed.

GitHub verification for `ae83348e2fe499f6a5c9f4a7503318c405404acc`: **68 checks passed, one expected Main Red Alert skipped, no failures**. The actual CI workflow [run 33987131421](https://github.com/LanternOps/breeze/actions/runs/33987131421) and its [CI Success gate](https://github.com/LanternOps/breeze/actions/runs/33987131421/job/101366685375) succeeded, including API/web tests, all four integration shards, migrations, builds and smoke checks. Earlier live readiness evidence at `acbf4dbc75` remains historical verification of the retained corrections.

The PR remains open and draft. `/health` remains liveness; `/ready` and `/health/ready` use the aggregate readiness contract. Existing failed jobs retain their stored options and require targeted cleanup before reused-queue rehearsals. Exact-candidate 10,000-device drain, pool/headroom and readiness failure/recovery measurements remain outstanding. No merge, deployment, candidate nomination or formal QA closure is claimed.


bdunncompany:
Took the independent review you asked for, scoped to the parts that can take the fleet offline rather than all 80 files. Two findings at head `3f11e249a`. The first was reproduced against a live edge; the second is a reading of the source, its own tests and current Compose behaviour — I did not reproduce the cold-start failure on a candidate stack, which is the same gap your fixed-unverified note describes.

### 1. The external smoke check stops being diagnostic behind an auth edge

`scripts/prod/deploy.sh` moves the post-deploy check from `/health` to `/ready`. Measured against a live instance behind Cloudflare Access:

```
/health  -> HTTP 200   {"status":"ok","version":"0.105.1",...}
/ready   -> HTTP 302   (redirect to the identity provider)
```

`curl --fail` only fails at ≥400, so the 302 exits **0**. The retry loop does not help — the first 302 satisfies the `if` and breaks it (`:391-397`), and the final request is identical (`:399-402`). Control: a deliberately bogus path behaves the same, exiting 0, which is what shows the edge is answering rather than Breeze.

Scope, stated precisely: this is the tested Access policy and proxies configured similarly — anything that answers `/ready` with a sub-400 response before reaching Breeze. A proxy that bypasses `/ready`, authenticates the probe, or returns 401/403 does not produce it, and a bare origin is unaffected (`readiness.ts` returns 200 only when ready, 503 otherwise). Notably this is not Breeze's own routing: Caddy sends both paths to the API and exempts both from the same middleware. The difference lives in the external policy, which covers `/health` and not `/ready`.

It is also not true that the deploy would sail through with the API dead — `:344-358` requires at least one running API container on the expected digest and `:369-382` fails closed without that proof, and the whole script is `set -e`. What is genuinely lost is the one check that proved traffic traversed the edge and reached Breeze. It can still miss a failure that appears after the Compose gate, or a container that is running but not serviceable through the edge.

Worth noting `--fail-with-body` would not fix it (same ≥400 threshold), and adding `-L` is worse — it can land on a 200 login page. The reliable forms are asserting the unredirected status is exactly 200, validating the readiness JSON, or probing `/ready` inside the Compose network.

### 2. Readiness now gates first-boot startup for three dependents

Three files switch the API healthcheck to `/ready` — both production definitions and `docker-compose.override.yml.dev`. A required consumer that is stopped, disconnected, initialization-failed or never attached makes `requiredConsumersRunnable()` false (`workerReadinessRegistry.ts:286-290`), which lands in the `ready` conjunction (`readiness.ts:191-212`) and returns 503 (`routes/readiness.ts:70-79`). Your own tests already cover those states as not-ready, so this is the intended behaviour, not a bug in it.

The consequence is on the dependency graph: `caddy` (`:55`), `web` (`:489`) and `portal` (`:521`) all declare `depends_on: api: condition: service_healthy`. On a first boot or a full-stack recreation, once the API exhausts its grace (`start_period 40s`, `interval 30s`, `retries 3`) and goes unhealthy, Compose aborts that operation and leaves those three unstarted — it does not wait forever, and it does not start them anyway. Caddy unstarted is the whole ingress, so an internal worker problem becomes loss of agent connectivity rather than degraded background work.

Two things I initially got wrong and want to correct rather than leave implied:

- **An ordinary reboot is not affected.** Docker Engine restarts existing containers under their restart policy without re-applying the Compose dependency gate. This needs a first boot, a full recreation, or a Compose operation where the dependent is absent or stopped — not any restart.
- **Your own deploy path largely sidesteps it**, because `deploy.sh` force-recreates Caddy with `--no-deps` at `:310` and again at `:388`, so Caddy is already up before the full-stack command runs.

So the exposure is narrower than "this can lock out the fleet": it is first boot and full recreation, on a stack where a required consumer cannot attach. Still worth a deliberate decision — whether ingress should be gated on worker readiness at all, or whether the container healthcheck stays on liveness while `/ready` serves the load balancer and the deploy check. The evaluator itself is bounded (probes raced against a timeout clamped to ≤5s), so this is about the gate's placement, not a hang.

### Not findings

The liveness/readiness split reads correct, and keeping the public payload to aggregate counts is right. I did not review the offline-sweep compare-and-set in depth.

ToddHebebrand:
Both findings verified against the branch and both fixed in `5e3ee6ace`. Thank you for scoping it to the fleet-offline surface rather than all 80 files — that is what made it reviewable.

**Finding 1 accepted, and the fix was already sitting in the file.** `curl --fail` really does only fail at >= 400, so the 302 exits 0 and the retry loop breaks on the first one. What makes this worse than it looks: `expect_barrier_status` at `scripts/prod/deploy.sh:238-244` already captures `%{http_code}` and compares it exactly, and it is used for all nine remote-access barrier probes immediately above. **The readiness smoke check was the only probe in the script not using the script's own idiom** — so this was not a missing technique, it was one call site that skipped it. `readiness_ok` now asserts an unredirected 200 *and* `"ready":true` in the body. Not `-L` and not `--fail-with-body`, for the reasons you gave.

**Finding 2 accepted, and I am taking it further than the fix you implied.** You framed it as a question of where the gate belongs. I think the sharper statement is that **`healthcheck:` is the wrong place for readiness at all**, and the reason is structural rather than about tuning: Compose exposes exactly one dependency condition, `service_healthy`, and it gates *startup*. There is no `service_ready`. So putting an admission probe in `healthcheck:` does not merely add a gate — it silently promotes admission into a hard startup dependency for everything downstream, and `caddy` (`:50-56`), `web` (`:488-490`) and `portal` (`:520-522`) are all downstream.

That inverts the tradeoff the readiness work was meant to make. Readiness exists to say "do not send this process traffic yet"; used here it says "do not let ingress exist yet." Losing background work is a degradation; losing Caddy is the fleet. So the three healthchecks go back to `/health`, and `/ready` keeps the deploy admission gate and any external load balancer — both of which can withhold traffic without preventing a process from starting.

Your two self-corrections both hold, and I checked them rather than taking them: `deploy.sh` force-recreates Caddy with `--no-deps` at `:310` and `:388`, and an ordinary reboot restarts existing containers without re-applying the dependency gate. So the exposure was first boot and full recreation only. I still treated it as must-fix, because within that window the failure is total and silent — Compose abandons the operation, and nothing about an unhealthy API says "your ingress will not come up."

`docs/operations/health-probes.md` said "Compose and deployment admission checks use `/ready`" — one clause conflating the two consumers. It now has a "Which probe goes where" section that names the Compose constraint explicitly, so the next person changing it sees why before they see the rule.

**On the test.** `productionReadinessWiring.test.ts` asserted `/ready` in the healthcheck, so it would have gone red on the fix — it is inverted to pin liveness, and it now also asserts the `depends_on` coupling that makes the rule load-bearing rather than stylistic. Four mutation controls, each confirmed to red only its intended named case: `/ready` restored to the healthcheck; caddy losing its `api` dependency; the bare `curl --fail` restored; `-L` added. **The caddy control did not land on the first attempt** — a naive slice for the next service key matched the nested `api:` inside caddy's own `depends_on`, so the mutation silently removed nothing and the suite stayed green. Caught only because the script asserted the mutation landed before running. That is your #4034 lesson arriving three hours later in my own work; the assertion is the reason it was not recorded as a passing control.

Noted that you did not review the offline-sweep compare-and-set. That half is still uncovered by an outside read, and it is the other place this PR can take the fleet offline — worth a look if you have the appetite, otherwise I will get an independent pass on it before this leaves draft.

Issue #3987, opened 2026-08-25
[Web] Replace Decommission / Permanent Delete with "Remove" (agent-uninstall choice) + "Delete permanently" — and fix menu parity on removed devices


ToddHebebrand:
## Description

Device offboarding is a two-step flow whose second step is nearly undiscoverable, whose confirm copy describes something the action doesn't do, and whose menus disagree with each other about what a decommissioned device can do next.

The API half of this change is #3986. This issue is the UX.

### Problem 1 — the menus don't agree, and one of them offers a 400

| Surface | Menu on an already-decommissioned device |
|---|---|
| Grid card (`DeviceCard.tsx:302`) | switches to **Restore** — never offers delete |
| List row menu (`DeviceList.tsx:2486`) | switches to **Restore** — never offers delete |
| Detail page / compact (`DeviceActions.tsx:410`, `:623`) | **no switch at all** — still offers "Decommission", which the API rejects with `400 Device is already decommissioned` |
| Device Settings modal (`DeviceSettingsModal.tsx:209`) | Restore **and** Permanently Delete — the only place either exists |

So permanent delete is reachable only by: unhide decommissioned rows via the status filter (or the `DecommissionedHiddenHint` "show" link) → open the kebab → Device Settings → scroll to the "Decommissioned Device" panel. From the detail page there is no route to it at all, and the one destructive item still on offer there fails.

### Problem 2 — the confirm copy is wrong in both directions

`apps/web/src/locales/en/devices.json:187`:

> Are you sure you want to decommission {{hostname}}? This will **permanently remove the device from your fleet**. The agent will stop reporting and the device will no longer be monitored.

Neither half holds. It is **not permanent** — `POST /devices/:id/restore` and the "Restore Device" button are two clicks away. And "the agent will stop reporting" reads as though the endpoint is now clean; it isn't. The agent stays installed and running, heartbeating into a `403` forever (`middleware/agentAuth.ts:419`). Nothing on screen says so.

`deviceSettings.decommissionWarning` (`devices.json:1542`) is accurate; it's the confirm string that's wrong.

### Problem 3 — "decommission" and "permanent delete" are the wrong axis

The two actions differ on *two* independent things at once: whether the agent comes off the machine, and whether the device's history is destroyed. Coupling them makes the most common MSP case unreachable — *"this laptop is being wiped and returned; take the agent off but keep the history for billing and compliance."*

## Proposed Fix

[...]

Issue #1322, opened 2026-06-13
[UI][API] Unified Devices list: surface network/discovered devices alongside agent endpoints + type-specific overview pages


ToddHebebrand:
## Description

Today the main **Devices** list shows only agent-managed endpoints (machines running the Go agent). Network-discovered devices — printers, routers, switches, firewalls, NAS, access points, cameras, IoT — live in a completely separate surface (the **Discovery** page). An MSP tech has to context-switch between two pages to see "everything on this site."

**Product owner (Todd) request:** Surface network devices in the **main Devices list** so there's one place to manage/monitor everything, and clicking a network device should open a **type-specific overview/detail page** that resembles the normal device detail page but is tailored — no agent-only sections (remote desktop, terminal, scripts, patching), instead showing the monitoring data that *does* exist for that device (SNMP, ping/uptime, interfaces, open ports, alerts).

This is a larger design/feature request. Filing as **one umbrella issue** with a design-forward plan and proposed sub-tasks (see end).

---

### Two separate data models

**Agent endpoints — `devices` table**
`apps/api/src/db/schema/devices.ts`
- Agent presence inferred from `agentTokenHash` (line 18) + `agentId` (line 17); `agentVersion` (line 55); `watchdogStatus` (line 73).
- **No explicit `type`/`kind`/`deviceClass` discriminator.** "Agent-ness" is implicit (token hash present). `deviceRole` (line 50) classifies *role* (server/workstation/etc.), not whether an agent exists.
- Scoped by `org_id` + `site_id` (lines 15-16).

**Network/discovered devices — `discovered_assets` table** (SEPARATE table)
`apps/api/src/db/schema/discovery.ts:106-139`
- `assetType` enum (`discovered_asset_type`): `workstation`, `server`, `printer`, `router`, `switch`, `firewall`, `access_point`, `phone`, `iot`, `camera`, `nas`, `unknown` (line 115).
- `approvalStatus`: `pending` / `approved` / `dismissed` (line 116) — discovery has a triage gate that the Devices list does not.
- `ipAddress`/`macAddress`/`hostname`/`label`, `manufacturer`/`model`, `openPorts`, `osFingerprint`, `snmpData`, `responseTimeMs`, `isOnline`, `lastSeenAt`.
- **`linkedDeviceId` → `devices.id` (line 128)**: the bridge. A discovered asset MAY link to an enrolled endpoint. A network-only device (printer/router) has `linkedDeviceId = NULL`.
[...]

PR #4011, opened 2026-08-25
fix(web): confirm-gate decommission from the device list/grid kebab (#4009)

### LINKED ISSUE #4009: [Web] Decommission from the device list/grid kebab skips the confirm dialog its own copy is already written for
## Description

Decommissioning a device from the **device list row kebab** or the **grid card kebab** executes immediately on a single click. No dialog, no "are you sure" — the toast says "Decommissioned" and the row leaves the list.

The device **detail** page confirms the same action. So whether decommissioning a production machine asks for confirmation depends entirely on which screen you happen to be on — which is exactly the inconsistency #3698 set out to remove.

Verified in a browser against `main` @ `c2f9db1ce`: Devices → row kebab → Decommission on `e2e-windows.local` decommissioned it outright. (Restored afterwards via the kebab's Restore, which also fires with no confirmation.)

## Root cause

The dialog and its copy already exist and are already wired. Only the gate omits the action.

`apps/web/src/components/devices/DevicesPage.tsx:98`

```ts
const CONFIRM_REQUIRED_ACTIONS = new Set(['reboot', 'reboot_safe_mode', 'shutdown']);
```

`DevicesPage.tsx:674`

```ts
const handleDeviceAction = async (action: string, device: Device) => {
  if (actionInProgress) return;
  if (CONFIRM_REQUIRED_ACTIONS.has(action)) {
    setPendingDeviceAction({ action, device });
    return;
  }
  await runDeviceAction(action, device);   // <- decommission lands here
};
```

Both kebabs dispatch straight into that handler:

- `DeviceList.tsx:2502` → `onAction?.("decommission", device)`
- `DeviceCard.tsx:320` → `onAction?.("decommission", device)`

And the confirm renderer at `DevicesPage.tsx:1456-1470` builds its strings dynamically from `deviceActions.confirm.${action}.{title,message,confirm}` — keys that **already exist** for decommission at `apps/web/src/locales/en/devices.json:185-189`. Adding the action to the set is enough to light the dialog up; nothing else has to be built.

The comment above the set explains the exclusion of `lock` — "it is reversible and does not disconnect the machine, which is where DeviceActions.tsx draws the same line". Decommission is neither of those things. The detail page's own gate (`DeviceActions.tsx:176-182`) treats it as destructive.

## Impact

The dense, easy-to-mis-click list is the screen the confirm gate was written for, and it is the one screen where the most consequential single-device action has no gate. Recovery exists (Restore), so this is not data loss — but on a real fleet a stray click silently pulls a machine out of monitoring, and the operator gets a past-tense toast rather than a question.

## Proposed fix

Add `decommission` to `CONFIRM_REQUIRED_ACTIONS`. The existing dynamic dialog and locale keys cover the rest.

Worth deciding at the same time whether `restore` should confirm; it is non-destructive, so probably not.

Note the confirm **copy** is separately wrong — see #3987, which documents that `devices.json:187` claims the action is permanent and that the agent stops reporting, neither of which holds. That is a wording fix in the same strings this issue turns on; the two should land together or in a deliberate order.

## Affected files

- `apps/web/src/components/devices/DevicesPage.tsx` (`:98` the set, `:674` the gate, `:1456` the renderer)
- `apps/web/src/components/devices/DeviceList.tsx` (`:2502`)
- `apps/web/src/components/devices/DeviceCard.tsx` (`:320`)
- `apps/web/src/locales/en/devices.json` (`:185-189`, already present)

Related: #3987 (offboarding redesign — assumes this confirm exists and criticises its wording), #3698 (added the list confirm gate this action was left out of), #3531 (destructive confirmations hiding failures — different symptom, same family).

Reported By: internal Playwright UI QA sweep 2026-08-25 (post-v0.107.1 merge range).

ToddHebebrand:
Closes #4009

## What was wrong

Decommissioning from the **device list row kebab** or the **grid card kebab** executed on a single click — no dialog, a past-tense "Decommissioned" toast, and the row gone. The device **detail** page has always confirmed the same action, so whether retiring a production machine asked a question depended on which screen you happened to be on. That is exactly the inconsistency #3698 set out to remove; decommission was simply left out of the set.

## The fix

One token in `apps/web/src/components/devices/DevicesPage.tsx`:

```ts
const CONFIRM_REQUIRED_ACTIONS = new Set(['reboot', 'reboot_safe_mode', 'shutdown', 'decommission']);
```

Verified end-to-end rather than assumed:

- The dialog renderer (`:1456-1470`) already composes `deviceActions.confirm.${action}.{title,message,confirm}` dynamically, so it lights up with no new UI.
- Those decommission keys already exist **in all eight locales** (`de-DE`, `en`, `es-419`, `fr-CA`, `fr-FR`, `it-IT`, `pt-BR`, `tr-TR`) — the detail page has been rendering them for a while. **No locale files are touched by this PR**, so there is no parity exposure.
- Both kebabs (`DeviceList.tsx:2502`, `DeviceCard.tsx:320`) dispatch into the shared `handleDeviceAction`, so both surfaces inherit the gate from the one change.

One extra line, deliberately: the dialog's `variant` was `action === 'shutdown' ? 'destructive' : 'warning'`, which would have rendered decommission as a caution-triangle while `DeviceActions.tsx:176-182` renders it as a stop-octagon. `ConfirmDialog` encodes severity by **shape** as well as colour (for glance-reading and colourblind users), so that would have been a visible severity downgrade on the denser of the two surfaces. Replaced with a small `DESTRUCTIVE_CONFIRM_ACTIONS` set that mirrors the detail page's own grading.

## Decision: `restore` stays ungated

The issue asked this to be decided deliberately. **It does not belong in the set.** `restore` is the *undo* of a decommission — the recovery path. Gating it would put a dialog in front of the thing you reach for after a mis-click, which makes the recovery slower without protecting anything: nothing is destroyed, and the action is itself reversible (decommission again). `lock` is excluded on the same reasoning today. There is now a test pinning this, so a future "confirm every lifecycle action" sweep has to argue with a test rather than quietly change the answer.

## What this PR does NOT do

- **The confirm copy is separately wrong and is out of scope.** `devices.json:187` claims the action is permanent and that the agent stops reporting; neither holds (Restore exists, and the agent keeps checking in). That is tracked in **#3987** (the offboarding Remove-vs-Delete-permanently redesign). **This PR lands first and turns those strings on for the list/grid**, so whoever takes #3987 is now rewording copy that is live on three surfaces (detail page, list kebab, grid kebab) instead of one. No new strings were added here, so #3987 rewords exactly the keys it already planned to.
- **The 5s undo toast in `runDeviceAction` is untouched.** It is a *post-hoc recovery window*, not a confirmation — it was already there while this bug was live, and a user who has looked away never sees it. Confirm-before plus undo-after is belt-and-braces on the most consequential single-device action in the kebab; collapsing the two is a UX call that belongs with #3987, not here.

## Tests

Five cases in `DevicesPage.test.tsx`, alongside the existing #3698/#3929 reboot coverage:

1. Row kebab opens the dialog instead of decommissioning — asserts the API is untouched, that *no undo toast fires either* (the pre-fix path raised one synchronously), that the dialog carries the detail page's own copy, and that the confirm button is graded `destructive`.
2. Cancelling decommissions nothing.
3. Confirming still performs the action — drives the 5s undo window with faked `setTimeout` and asserts `decommissionDevice(DEV_1)` fires exactly once after it elapses.
4. Grid card is gated on the same terms (the gate lives on the shared handler; this pins that it actually reaches the card).
5. `restore` stays ungated — the scope decision above, as an assertion.

**Revert probe:** with the one-token change reverted, 4 of the 5 fail (the 5th is the `restore` boundary, which is a pin rather than a regression test). Reverting *only* the `variant` line fails the severity assertion on its own, so that assertion is not vacuous either.

## Verification

- `vitest run src/components/devices` — 56 files, 676 tests, green.
- `vitest run src/components/devices/DevicesPage.test.tsx` — 60 passed.
- `vitest run src/lib/i18n src/components/shared/ConfirmDialog.test.tsx src/lib/__tests__/no-silent-mutations.test.ts` — 255 passed (the i18n dynamic-key guard covers the `/* i18n-dynamic */` call sites this touches).
- `tsc --noEmit -p apps/web/tsconfig.json` — clean.
- `eslint` on both changed files — clean.
- No `e2e-tests` spec references decommission, so nothing there needs updating.

## Review follow-up — new issue #4014 (not fixed here)

The `pr-test-analyzer` pass surfaced a real, **pre-existing** gap that this PR deliberately does not fix: `DeviceCard.tsx` has no `deviceClass` branch anywhere, where `DeviceList.tsx:2308-2321` replaces its whole actions cell with a "View" button for `deviceClass === 'network'` rows (their id is a `discovered_assets.id`, not a `devices.id` — #1322). So the **grid** kebab still offers decommission, reboot and terminal for a network-discovered asset. Filed as **#4014**; the confirm added here at least puts a dialog in front of it. The grid-card test now carries a comment naming that limit, so it is not misread as vetting the grid kebab.

### DIFF
--- a/apps/web/src/components/devices/DevicesPage.test.tsx
+++ b/apps/web/src/components/devices/DevicesPage.test.tsx
@@ -1,6 +1,6 @@
 import '@/lib/i18n';
 
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 import DevicesPage from './DevicesPage';
@@ -118,9 +118,10 @@ vi.mock('./DeviceFilterToolbar', () => ({ DeviceFilterToolbar: () => null }));
 vi.mock('../shared/ProgressBar', () => ({ default: () => null }));
 
 // DeviceCard stub renders the hostname so the grid contents are assertable, and
-// re-emits reboot so the GRID path through handleDeviceAction is covered too —
-// the real DeviceCard emits it from its own kebab (DeviceCard.tsx), and #3698's
-// gate lives on the shared handler, so both surfaces inherit it.
+// re-emits reboot + decommission so the GRID path through handleDeviceAction is
+// covered too — the real DeviceCard emits both from its own kebab
+// (DeviceCard.tsx:278 reboot, :320 decommission), and the #3698/#4009 gate
+// lives on the shared handler, so both surfaces inherit it.
 vi.mock('./DeviceCard', () => ({
   default: ({ device, onAction }: { device: { id: string; hostname: string }; onAction?: (action: string, device: unknown) => void }) => (
     <div data-testid={`device-card-${device.id}`}>
@@ -132,6 +133,12 @@ vi.mock('./DeviceCard', () => ({
         aria-label="card reboot"
         data-testid={`card-reboot-${device.id}`}
         onClick={() => onAction?.('reboot', device)}
+      />
+      <button
+        type="button"
+        aria-label="card decommission"
+        data-testid={`card-decommission-${device.id}`}
+        onClick={() => onAction?.('decommission', device)}
       />
     </div>
   ),
@@ -176,7 +183,11 @@ vi.mock('./DeviceList', () => ({
         </button>
       ))}
       {/* Drives DevicesPage.handleDeviceAction for ONE device — the row-menu /
-          grid-card path, as opposed to the bulk buttons above. */}
+          grid-card path, as opposed to the bulk buttons above. The real row
+          kebab emits these from DeviceList.tsx (:2450 reboot, :2490 restore,
+          :2502 decommission). The decommission/restore buttons are deliberately
+          text-free so they cannot collide with the sibling getByText assertions
+          that match on the word "decommissioned". */}
       {devices.map(d => (
         <button
           key={`row-reboot-${d.id}`}
@@ -186,6 +197,24 @@ vi.mock('./DeviceList', () => ({
         >
           row reboot {d.id}
         </button>
+      ))}
+      {devices.map(d => (
+        <button
+          key={`row-decommission-${d.id}`}
+          type="button"
+          aria-label={`row decommission ${d.id}`}
+          data-testid={`row-decommission-${d.id}`}
+          onClick={() => onAction?.('decommission', d)}
+        />
+      ))}
+      {devices.map(d => (
+        <button
+          key={`row-restore-${d.id}`}
+          type="button"
+          aria-label={`row restore ${d.id}`}
+          data-testid={`row-restore-${d.id}`}
+          onClick={() => onAction?.('restore', d)}
+        />
       ))}
       {onShowDecommissioned && (
         <button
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three lenses in parallel (code-reviewer, pr-test-analyzer, comment-analyzer). `type-design-analyzer` and `silent-failure-hunter` were skipped deliberately: this diff adds no types and no error-handling code (the `decommission` try/catch in `runDeviceAction` is untouched).

**Findings:** 0 critical, 1 important, 4 suggestions. Acted on the important one and 2 suggestions in `38790f8`; 2 deliberately left.

**Important — acted on (as scope, not as a code change).** `pr-test-analyzer` found that `DeviceCard.tsx` has **no `deviceClass` branch anywhere**, where `DeviceList.tsx:2308-2321` swaps its entire actions cell for a "View" button on `deviceClass === 'network'` rows — their id is a `discovered_assets.id`, not a `devices.id` (#1322). So the **grid** kebab still offers decommission, reboot and terminal for a network-discovered asset. I verified it independently rather than taking the agent's word: zero `deviceClass` references in `DeviceCard.tsx`, and `DevicesPage.tsx:301-309` → `:1364-1371` puts network rows into `<DeviceCard>` with the full `onAction` wiring whenever the class filter is `All`.

This **predates #4009 and is not decommission-specific**, so fixing it here would have been scope creep on a one-token gate fix. Filed as **#4014** instead, and the grid-card test now carries a comment naming the limit so it is not misread as vetting the grid kebab. The confirm added here does at least put a dialog in front of that call.

**Suggestions — acted on:**
- Dropped hard-coded `DeviceCard.tsx:278/:320` and `DeviceList.tsx:2450/:2490/:2502` line refs from the test-stub comments. They were verified accurate, but they rot silently with nothing to catch it; the `onAction?.(...)` call sites are cited instead.
- Softened an unfalsifiable superlative ("the densest, easiest-to-mis-click surface in the product") in the `CONFIRM_REQUIRED_ACTIONS` comment.
- Also documented what the stub deliberately does *not* model: the real menus render `restore` only for a decommissioned device and `decommission` only for one that is not, so this file proves `handleDeviceAction`'s dispatch, never that the two are mutually exclusive on screen.

**Suggestions — deliberately left:**
- *Severity grading lives in two places* (`DESTRUCTIVE_CONFIRM_ACTIONS` here vs `getModalConfig`'s per-case `variant` in `DeviceActions.tsx`) and can drift. Real, but pre-existing in kind — this file already had a hard-coded `action === 'shutdown' ? …` ternary, and the two files compose confirms differently (generic i18n-key lookup vs a switch-per-action config), so `getModalConfig` is not a drop-in here. Unifying them is a refactor, not part of a gate fix.
- *No decommission-specific double-click or Escape test.* The #3705 double-click tests already cover the identical `ConfirmDialog` + unmount-on-confirm path with the same `confirm-device-action` testid, and Escape is `Dialog`'s own wiring, exercised by the cancel test. Decommission adds itself to two Sets and no new confirm-path code, so both would re-verify proven shared mechanics.

**Verified clean by the review:** no modal-on-modal risk (`DeviceSettingsModal.tsx:250-252` calls `onClose()` *before* `onAction`); bulk decommission untouched (neither Set is read outside `handleDeviceAction` and the dialog `variant`); the new stub buttons are `aria-label`-only so they cannot collide with sibling `getByText(/decommissioned/)` assertions; and every factual claim in the added comments checked out against source, including that `shutdown` and `decommission` are the *only* `destructive` cases in `DeviceActions.tsx`.

**Tests:** `vitest run src/components/devices` — 56 files / 676 tests green locally; `DevicesPage.test.tsx` 60/60. Revert probe: 4 of the 5 new cases fail with `'decommission'` removed from the Set, and the `bg-destructive` assertion fails on its own if only the `variant` line is reverted — so neither is vacuous. `tsc --noEmit` and `eslint` clean.

**CI on `38790f8`:** Test Web **pass**, Type Check **pass**, Lint **pass**, Integration Tests shard 2/4 **pass**, Smoke Test **pass**. Test API and integration shards 1/3/4 still queued; this PR touches only `apps/web`. No red checks.

**Status:** review-clean, awaiting maintainer merge. Not merging, and #4009 stays open and assigned until the reporter confirms.

PR #4012, opened 2026-08-25
fix(web): keep the #rings deep link alive until the JWT scope is known (#4010)

### LINKED ISSUE #4010: [Web] /patches#rings never survives a page load: the scope guard downgrades and clears the hash before the JWT scope exists
## Description

`/patches#rings` is written into the URL when you click the **Update Rings** tab, but it can never be read back. Loading or refreshing that URL renders the **Compliance** tab and rewrites the address bar to a bare `/patches`, discarding the hash.

This is not a race that sometimes loses — it fails on **every** cold load, for **every** user, because the value the guard depends on is structurally unavailable at first paint.

Verified in a browser against `main` @ `c2f9db1ce`, logged in as a partner-scope admin:

1. Click **Update Rings** → URL becomes `/patches#rings`. Correct.
2. Reload (or paste that URL fresh) → active tab is **Compliance**, `location.hash` is `""`, `location.href` is `http://localhost:32812/patches`.
3. The **Update Rings** tab is present in the tab bar the whole time — so the user *is* allowed to see it. It just isn't selected.

## Root cause

`apps/web/src/components/patches/PatchesPage.tsx`

The tab is guarded by JWT scope:

```ts
const { scope } = getJwtClaims();
const canManageRings = scope === 'partner' || scope === 'system';   // :84
```

```ts
function resolveTab(tab: TabKey, canManageRings: boolean): TabKey {   // :53
  return tab === 'rings' && !canManageRings ? 'compliance' : tab;
}
```

```ts
useEffect(() => {                                                     // :114
  const syncFromHash = () => {
    const raw = getTabFromHash();
    const resolved = resolveTab(raw, canManageRings);
    if (resolved !== raw) {
      setActiveTab(resolved); // downgrade — also clears the stale hash
    } else {
      setActiveTabState(resolved);
    }
  };
  syncFromHash();
  ...
}, [canManageRings, setActiveTab]);
```

`setActiveTab` calls `setTabInHash`, which for `compliance` does `url.hash = ''` + `history.replaceState`. That is the only writer of the hash on this page, so the observed clear proves the downgrade branch ran, which proves `canManageRings` was `false` on the first `syncFromHash()`.

**Why it is always false at that point.** `getJwtClaims()` (`apps/web/src/lib/authScope.ts:21`) reads `useAuthStore.getState().tokens?.accessToken`. Access tokens are deliberately **not persisted** — the `breeze-auth` localStorage blob holds only `user` and `isAuthenticated`:

```json
{"state":{"user":{...,"canManagePartnerWide":true},"isAuthenticated":true},"version":2}
```

So on any fresh page load there is no token in the store yet, `getJwtClaims()` returns `NO_CLAIMS`, `scope` is `null`, and `canManageRings` is `false` — regardless of who the user actually is.

The effect's own comment anticipates late-arriving scope: *"Depending on `canManageRings` also re-runs the guard when scope/permissions arrive after the first render (defense-in-depth)."* The re-run does happen — but it cannot recover anything, because the first run already deleted the `#rings` it would have needed to read. `getTabFromHash()` now returns `compliance`.

## Impact

Small but squarely against the project's own URL-state convention (CLAUDE.md: hash for selected tab, which the file's header comment cites). A bookmarked or shared Update Rings view silently lands somewhere else, and refreshing while working in rings kicks you back to Compliance. The same shape would bite any other scope-gated tab that adopts this pattern.

## Proposed fix

Do not let an unresolved scope destroy the hash. Options, roughly in order of preference:

1. **Only downgrade once scope is actually known.** Track a "claims resolved" signal (token present, or an explicit `scopeLoaded` flag) and skip `syncFromHash`'s downgrade branch until then. The effect already re-runs on `canManageRings`, so the correct tab lands as soon as the token arrives.
2. **Downgrade the view without rewriting the URL.** Render compliance but leave `#rings` in place, so the re-run can still recover it. Clear the hash only on a downgrade taken against a *known* scope.
3. **Derive the guard from something available at first paint.** `user.canManagePartnerWide` is already persisted in `breeze-auth` and would be true here; it is not identical to JWT scope, so this needs a deliberate call rather than a swap.

Whichever is chosen, a regression test that loads the page with `#rings` and a not-yet-populated auth store would pin it.

## Affected files

- `apps/web/src/components/patches/PatchesPage.tsx` (`:53` `resolveTab`, `:84` `canManageRings`, `:114-127` the sync effect, `:43` `setTabInHash`)
- `apps/web/src/lib/authScope.ts` (`:21` `getJwtClaims` — reads a store that is empty at first paint)

Worth a sweep of the sibling pages the comments name as precedent — `DiscoveryPage`, `DeviceDetails` — for the same guard-before-scope shape.

Reported By: internal Playwright UI QA sweep 2026-08-25 (post-v0.107.1 merge range).

ToddHebebrand:
## Problem

`/patches#rings` is written into the URL when you click **Update Rings**, but it can never be read back. Every cold load of that URL renders **Compliance** and rewrites the address bar to a bare `/patches`.

The tab guard derived `canManageRings` from `getJwtClaims().scope`, and access tokens are deliberately never persisted — `partialize` in `stores/auth.ts` keeps only `user` and `isAuthenticated`. So at first paint the auth store is empty and **every** user decodes as `scope: null`. The mount-time sync effect took the downgrade branch, which routes through `setActiveTab` → `setTabInHash` and deletes the hash. The effect's `canManageRings` dependency *did* re-run once the token landed — but by then `getTabFromHash()` had nothing left to read. Deterministic, not a race.

## Fix — option 1 from the issue (with a slice of option 2)

Option 1 asked for an explicit "claims resolved" signal instead of inferring one from a falsy `canManageRings`; option 2 asked to downgrade the view without rewriting the URL. Neither alone is quite right — option 1 still needs to decide what to render during the unresolved window, and option 2 alone leaves `canManageRings` non-reactive. Doing both, in the two places they belong:

**`lib/authScope.ts` — new `useJwtClaims()`**, a reactive companion to `getJwtClaims()`:
- subscribes to the access token, so the component genuinely re-renders when it arrives rather than relying on an incidental re-render from some unrelated fetch;
- returns a **discriminated union**, `JwtClaimsState = { status: 'unresolved' } | { status: 'resolved'; claims: JwtClaims }`, so `.claims.scope` is unreachable until the caller narrows on `status`. That is deliberate: a flat `{ ...claims, resolved: boolean }` would let the next caller destructure `{ scope }`, ignore the flag, and reproduce #4010 with the compiler's blessing. The check is structural, not a comment.
- a present-but-undecodable token is `'resolved'` with all-null claims, so we still fail closed. (Client-side UX only — the server rejects such a token outright with a 401 rather than degrading the page.)

`getJwtClaims()` is unchanged for its existing callers; the decoder is factored into a shared pure function so there is still one implementation.

**`PatchesPage.tsx`** keeps all three access states as `ringAccess: 'unresolved' | 'allowed' | 'denied'` rather than collapsing to a boolean. `'denied'` and `'unresolved'` both hide the rings UI, but **only `'denied'` may clear the hash** — the destructive branch keys on a positive `=== 'denied'`, not on the absence of a flag. Everything that merely hides or disables UI still reads the fail-closed `canManageRings = ringAccess === 'allowed'`.

The **resolved → unresolved** transition (a token going *away*) defers too, and that is deliberate rather than an oversight: while `/auth/refresh` is rate limited the session is perfectly valid and yet tokenless for up to 90 seconds (#3696), so wiping the deep link there would be #4010 again in slow motion. Logout lands on the same branch and leaves `#rings` behind, which is harmless because every logout path navigates away from the page. Pinned by a test.

**Option 3 was not taken** — deliberately. Swapping the guard to `user.canManagePartnerWide` changes *what authorizes a partner-scoped tab* from JWT scope to a different predicate (#3262: false for partner users with `org_access = 'selected'`). That is a product decision about who may manage update rings, not a bug fix, so it stays out of this PR.

### The security property is unchanged

An org user on `#rings` still lands on Compliance, and the rings tab and body are never rendered while the scope is unknown — an unknown scope is treated as *not yet a grant*, never as a grant. Their downgrade (and hash clear) is merely deferred to the moment it is real. Both directions are pinned by tests.

## Tests

`PatchesPage.scopeResolution.test.tsx` is new and mocks **neither** `authScope` nor the auth store — it drives the real `useJwtClaims` against the real store so the token genuinely arrives after mount, which is the only way to reproduce this bug. (The existing `PatchesPage.test.tsx` mocks `authScope` with an already-known scope, i.e. the one state a cold load never starts in; that is why the bug shipped past a suite that already had four `#rings` guard tests.)

Coverage:
- partner **and** system scope arriving after first paint → `#rings` survives the unresolved window and the Update Rings tab becomes active;
- org scope arriving after first paint → still downgrades to Compliance **and** clears the stale `#rings`;
- undecodable token → resolved-and-denied, hash cleared (fails closed);
- a `hashchange` back to `#rings` during the unresolved window → also deferred, not wiped;
- a token going *away* again (throttled refresh / logout) → rings hidden, hash deliberately kept, and rings restored when the token returns;
- a token landing while the user is on `#patches` → the Update Rings nav chrome appears without yanking them off their tab;
- `useJwtClaims` unresolved/resolved/undecodable/logout semantics and return-value identity stability in `authScope.test.ts`.

**Revert probe:** restoring the old effect body fails **6 of the 7** cases in the new suite with the bug's signature, `AssertionError: expected '' to be '#rings'`. The one that passes either way is the `#patches` case, which has no rings hash to destroy.

Two of the original five cases initially passed against the reverted code — the old premature clear and the new deliberate clear reach the same *final* state for a denied user, and those tests only asserted the final state. They now assert the hash is intact *before* the token lands, which is the property that actually differs. That was a review finding, and it is the difference between 3/5 and 6/7 discriminating.

## Review

Four reviewers ran (`code-reviewer`, `pr-test-analyzer`, `comment-analyzer`, `type-design-analyzer`), plus an independent second opinion from Codex on the type-shape question.

`code-reviewer` returned **0 Critical / 0 Important**, having traced that no ordering — SSR, hydration, `hashchange` during the unresolved window, logout mid-session — can turn "unresolved" into a grant. The rest of the findings are addressed in the follow-up commit: the union refactor above (raised independently by `type-design-analyzer` and Codex, which is why it was adopted rather than deferred), the two non-discriminating tests, two missing sequences, and three stale comments — the worst of which asserted that "canManageRings itself stays accurate everywhere else (the hash-resolution effect, ring actions) so the #rings deep-link and scope guards are unaffected", a claim #4010 disproved, sitting fifteen lines above its own fix.

## Sibling pages — swept, and the sweep found something else

The issue asked whether `DiscoveryPage` and `DeviceDetails`, cited in the code comments as precedent for this hash pattern, share the shape. **Neither does**, and the chain breaks in a different place in each:

- **`DiscoveryPage.tsx`** — the mount effect's `getTabFromHash()` (`:247-254`, called at `:286-291`) only checks membership in `DISCOVERY_TABS`. There is no permission gate on the hash value at all. The page's `useOrgScope()` (`:281`) gates *rendering* (`:658-660`) and never touches `location.hash`; hash writes happen only in the user-driven `navigateToTab` (`:601-604`).
- **`DeviceDetails.tsx`** — reads through `useHashState` (`:318`), whose contract is read-only on mount (`lib/useHashState.ts:19-20`). Its one gate (`:370-371`) is derived from `device.linkGroupId`, not a permission, and only overrides a local render variable — it never writes the URL. `getJwtClaims` is not referenced in the file.

A repo-wide cross-reference (every file calling `getJwtClaims` × every file touching `location.hash`/`replaceState`/`pushState`) leaves five candidates: this file plus `PartnerSettingsPage`, `TicketingSettingsTabs`, `QuickbooksIntegration` and `IntegrationsPage`. None of the other four reproduces the full shape — each either has no permission gate on the hash value, or gates render only and never writes the URL. **`PatchesPage` was the only instance.**

The sweep did turn up a **different** latent bug with the same root cause, filed as #4013 rather than widened into this PR:

> `TicketingSettingsTabs.tsx:80` — `const canManageInbound = useMemo(() => getJwtClaims().scope === 'partner', [])`. Empty deps freeze a value read from a store that is empty at first paint, so on a cold load it is pinned `false` for the life of the mount. It gates render only (never a URL write), so it is not #4010's shape — but it hides Intake Forms / Inbound Email / Canned Responses from partner users who land directly on `/settings/partner#ticketing`, which is exactly where the permanent 301 alias at `pages/settings/ticketing.astro` sends every old bookmark. `useJwtClaims()` from this PR is the ready-made fix.

## Verification

- `apps/web` — `vitest run src/lib/authScope.test.ts src/components/patches/` → **11 files, 130 tests passed** (single-fork).
- Wider authScope-consumer sweep (settings + tickets suites that mock `authScope`) → **21 files, 381 tests passed**.
- `astro check` → **0 errors, 0 warnings** (1679 files).
- `eslint` on all five changed files → clean.

Closes #4010

### DIFF
--- a/apps/web/src/components/patches/PatchesPage.scopeResolution.test.tsx
+++ b/apps/web/src/components/patches/PatchesPage.scopeResolution.test.tsx
@@ -117,6 +117,13 @@ describe('PatchesPage #rings deep link vs. late-arriving JWT scope (#4010)', () 
     expect(await screen.findByRole('button', { name: /Compliance/i })).toBeInTheDocument();
     // Deferred, not granted: the rings tab is not offered while the scope is unknown.
     expect(screen.queryByRole('button', { name: 'Update Rings' })).toBeNull();
+    // ...and the hash is still intact at this point. The guard cannot know in
+    // advance which way an unresolved user will go, so it must defer for
+    // EVERYONE — including the ones who will turn out to be denied. Without this
+    // assertion the test passes against the buggy code too, since the old
+    // premature clear and the new deferred-then-real clear reach the same final
+    // state for an org user.
+    expect(window.location.hash).toBe('#rings');
 
     tokenArrives({ scope: 'organization', orgId: 'org-1' });
 
@@ -134,6 +141,8 @@ describe('PatchesPage #rings deep link vs. late-arriving JWT scope (#4010)', () 
     // "scope pending" state that never clears a bookmark an org user can't use.
     render(<PatchesPage />);
     expect(await screen.findByRole('button', { name: /Compliance/i })).toBeInTheDocument();
+    // Same discriminating assertion as above: intact until the token exists.
+    expect(window.location.hash).toBe('#rings');
 
     act(() => {
       useAuthStore.setState({ tokens: { accessToken: 'not-a-jwt', expiresInSeconds: 900 } });
@@ -141,6 +150,47 @@ describe('PatchesPage #rings deep link vs. late-arriving JWT scope (#4010)', () 
 
     await waitFor(() => expect(window.location.hash).toBe(''));
     expect(screen.queryByRole('button', { name: 'Update Rings' })).toBeNull();
+  });
+
+  it('a token going away again (throttled refresh / logout) hides rings but does NOT wipe the hash', async () => {
+    // The resolved -> unresolved transition is a real sequence: while
+    // /auth/refresh is rate limited the session is valid and yet tokenless for
+    // up to 90s (#3696). Wiping the deep link there would be #4010 in slow
+    // motion, so the guard defers exactly as it does before the first token.
+    render(<PatchesPage />);
+    tokenArrives({ scope: 'partner', partnerId: 'p-1' });
+    await waitFor(() => expect(navTab('Update Rings')).toHaveClass('border-primary'));
+
+    act(() => {
+      useAuthStore.setState({ tokens: null });
+    });
+
+    // Falls closed on the view...
+    await waitFor(() => expect(screen.queryByRole('button', { name: 'Update Rings' })).toBeNull());
+    expect(screen.queryByRole('button', { name: /New Ring/i })).toBeNull();
+    // ...but the deep link survives, so the same session recovers on re-auth.
+    expect(window.location.hash).toBe('#rings');
+
+    tokenArrives({ scope: 'partner', partnerId: 'p-1' });
+    await waitFor(() => expect(navTab('Update Rings')).toHaveClass('border-primary'));
+  });
+
+  it('a token landing while the user is on #patches reveals the Update Rings tab without moving them', async () => {
+    // The nav chrome is gated on `mounted && canManageRings`, a separate path
+    // from the hash guard: a partner browsing Patches during the unresolved
+    // window must gain the tab when the token lands, and must not be yanked off
+    // the tab they are on.
+    window.history.replaceState({}, '', '/#patches');
+
+    render(<PatchesPage />);
+    await waitFor(() => expect(navTab('Patches')).toHaveClass('border-primary'));
+    expect(screen.queryByRole('button', { name: 'Update Rings' })).toBeNull();
+
+    tokenArrives({ scope: 'partner', partnerId: 'p-1' });
+
+    await waitFor(() => expect(screen.getByRole('button', { name: 'Update Rings' })).toBeInTheDocument());
+    expect(navTab('Patches')).toHaveClass('border-primary');
+    expect(window.location.hash).toBe('#patches');
   });
 
   it('a hashchange back to #rings while the scope is still unknown does not wipe the hash', async () => {
--- a/apps/web/src/components/patches/PatchesPage.scopeResolution.test.tsx
+++ b/apps/web/src/components/patches/PatchesPage.scopeResolution.test.tsx
@@ -1,0 +1,166 @@
+import { act, render, screen, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import '@/lib/i18n';
+
+// #4010. PatchesPage.test.tsx mocks `lib/authScope` with a scope that is already
+// known on the first render, which is exactly the state a cold page load does NOT
+// start in: access tokens are never persisted (`partialize` in stores/auth.ts
+// keeps only `user` + `isAuthenticated`), so at first paint the store is empty and
+// EVERY user decodes as `scope: null`. This suite therefore mocks neither
+// authScope nor the auth store — it drives the real `useJwtClaims` against the
+// real store so the token genuinely arrives after mount, which is the only way to
+// pin the bug: the mount-time guard used to downgrade #rings AND clear the hash,
+// destroying the deep link before the scope it depends on existed.
+
+vi.mock('../../components/shared/Toast', () => ({ showToast: vi.fn() }));
+
+// Real auth store (the subject), stubbed network helper.
+vi.mock('../../stores/auth', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../../stores/auth')>()),
+  fetchWithAuth: vi.fn(),
+}));
+
+const orgState = vi.hoisted(() => ({ currentOrgId: null as string | null }));
+
+vi.mock('../../stores/orgStore', () => {
+  const organizations = [{ id: 'org-1', name: 'Acme Corp' }];
+  const read = () => ({ currentOrgId: orgState.currentOrgId, organizations });
+  return { useOrgStore: Object.assign(read, { getState: read }) };
+});
+
+import PatchesPage from './PatchesPage';
+import { fetchWithAuth, useAuthStore } from '../../stores/auth';
+
+const fetchMock = vi.mocked(fetchWithAuth);
+
+const makeJsonResponse = (payload: unknown, ok = true, status = ok ? 200 : 500): Response =>
+  ({
+    ok,
+    status,
+    statusText: ok ? 'OK' : 'ERROR',
+    json: vi.fn().mockResolvedValue(payload),
+  }) as unknown as Response;
+
+const emptyDataFetch = async (input: unknown): Promise<Response> => {
+  const url = String(input);
+  if (url === '/update-rings') return makeJsonResponse({ data: [] });
+  if (url === '/patches?limit=200') return makeJsonResponse({ data: [] });
+  if (url === '/patches/compliance') {
+    return makeJsonResponse({ data: { totalDevices: 0, compliantDevices: 0, devicesNeedingPatches: [] } });
+  }
+  if (url === '/devices?limit=200') return makeJsonResponse({ devices: [] });
+  return makeJsonResponse({}, false, 404);
+};
+
+function makeToken(payload: Record<string, unknown>): string {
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1: code-reviewer, pr-test-analyzer, comment-analyzer, type-design-analyzer (plus an independent Codex second opinion on the type-shape question). Round 2: code-reviewer, scoped to the refactor delta `66a97308e`.

**Findings:** round 1 — **0 Critical, 0 Important from code-reviewer**, which traced that no ordering (SSR, hydration, `hashchange` during the unresolved window, logout mid-session) can turn "scope unresolved" into a grant. 8 substantive findings raised across the other three lenses; all 8 addressed in `66a97308e` and `079a797dc`. Round 2 — **0 Critical, 0 Important, 2 Suggestions**, both taken in `079a797dc`. 0 outstanding.

What actually changed as a result of review:

- **The type now enforces its invariant instead of documenting it.** The original `ResolvedJwtClaims extends JwtClaims { resolved: boolean }` let a caller destructure `{ scope }`, ignore the flag, and reproduce #4010 with the compiler's blessing — the same bug, one call site later. `useJwtClaims()` now returns a discriminated union `{ status: 'unresolved' } | { status: 'resolved'; claims: JwtClaims }`, so `.claims.scope` is unreachable until `status` is narrowed. `type-design-analyzer` and Codex reached this independently, which is why it was adopted here rather than deferred. `PatchesPage` correspondingly keeps three states (`ringAccess`), and the destructive branch keys on a positive `=== 'denied'`.
- **Two of the five original regression cases did not discriminate.** The old premature hash-clear and the new deliberate one reach the same *final* state for a denied user, and those tests only asserted the final state. They now assert the hash is intact *before* the token lands. The revert probe went from 3/5 to **6/7** failing.
- **Pinned the resolved → unresolved transition.** A token can vanish while the session is fine — `/auth/refresh` throttling (#3696) leaves a valid session tokenless for up to 90s. Deferring rather than clearing there is deliberate; it was previously only an inference from reading the code.
- Plus two missing sequences, hook return-value identity stability, and three stale comments — the worst of which asserted that "canManageRings itself stays accurate everywhere else (the hash-resolution effect, ring actions) so the #rings deep-link and scope guards are unaffected", a claim #4010 disproved, sitting fifteen lines above its own fix.

**Tests:** `apps/web` affected suites green — `vitest run src/lib/authScope.test.ts src/components/patches/` → **11 files, 133 tests passed** (single-fork). Wider sweep of every suite that mocks `authScope` (settings + tickets) → 21 files, 381 passed. `astro check` → **0 errors, 0 warnings** (1679 files). `eslint` on all changed files clean.

**CI:** green on `079a797dc` — `Test Web` ✅, `Type Check` ✅, `Test API` ✅, `Lint` ✅. 51 passed, 0 failures.

**Sibling sweep:** `DiscoveryPage` and `DeviceDetails`, named in the code comments as precedent, are **not** affected — neither gates its hash-derived tab on a permission. A repo-wide cross-reference of `getJwtClaims` callers against `location.hash`/`replaceState` writers leaves five candidates, and `PatchesPage` was the only one with the full shape. The sweep did surface a different latent bug with the same root cause (a frozen `useMemo` over JWT scope hiding three ticketing tabs on a cold load) — filed as #4013 rather than widened into this PR.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4010 — both are yours to call.

PR #4015, opened 2026-08-25
fix(api): redact channel credentials before composing the test-failure reason (#3992)

### LINKED ISSUE #3992: [UI] Zero-data empty states say "adjust your search"; channel test reason dumps raw HTML into the card
## Description

Two papercuts found while verifying the post-v0.107.0 merge range. Both are small and localized.

### 1. Zero-data empty states use "adjust your search" copy

When a list has **no rows at all** and no search or filter has been entered, the UI shows the no-search-results message:

- `/configuration-policies` → *"No policies found. Try adjusting your search."*
- `/alerts/channels` → *"No notification channels found. Try adjusting your search or filters."*

This tells a first-time user to adjust a search they never made, and hides the actual next step (create one).

The codebase already gets this right elsewhere, so this is an inconsistency rather than a house style:

- `/settings/enrollment-keys` → *"No enrollment keys found. Create one to get started."*
- `/contracts` → *"No contracts yet — Contracts bill an organization on a repeating cadence and auto-generate the invoices for you."* + a **Create your first contract** button
- `/billing/quotes` → *"No quotes yet"* + explanatory copy + **New quote**

**Fix:** branch on `totalCount === 0 && !hasActiveSearch/Filter` and render a first-run empty state with the create action, matching the contracts/quotes pattern.

### 2. Notification-channel test failure dumps the raw HTML response body

The #3926 persisted-test-reason feature works correctly — status and reason are stored and shown. But the reason echoes the remote response body verbatim into the channel card:

> **Failed** · Last test: Just Now
> Reason: `HTTP 405: <!doctype html><html lang="en"><head><title>Example Domain</title><link rel="icon" href="data:,"><meta name="viewport" content="width=device-width, initial-scale=1"><style>body{background:#eee;width:60vw;margin:15vh auto;font-family:system-ui,sans-serif}h1{font-size:1.5em}...`

Storage is bounded (`length(last_test_error)` = 500, truncated server-side), so this is not unbounded growth — but 500 characters of HTML markup in a card makes the card unreadable and buries the one useful token (`HTTP 405`).

**Fix:** store/render the status line plus a short body excerpt with markup stripped, or show `HTTP 405` in the card and put the body behind a "details" disclosure.

## Affected files

- `apps/web/src/components/configurationPolicies/` (policy list empty state)
- `apps/web/src/components/alerts/NotificationChannelsPage.tsx` (empty state + test-reason rendering)
- the channel-test handler that composes `last_test_error`

Verified against `main` @ `c2f9db1ce`.

ToddHebebrand:
Part 2 of #3992. **Part 1 shipped in #4032** (@bdunncompany) — this branch was rebuilt as a single commit on top of it and no longer touches the empty states, `ConfigPolicyList`, or `policies.json`.

Closes #3992

## The reported bug

A webhook channel test that failed showed the operator 500 characters of the destination's HTML error page, burying the only useful token. `formatHttpFailure` now reduces the body to one readable sentence — markup stripped, entities decoded, whitespace collapsed, capped at 160 characters — and the unshortened form is still kept in a log line or the delivery record on every path.

## The security-relevant part: ordering

`scrubChannelTestError` runs at the route (`apps/api/src/routes/alerts/channels.ts:689`) and removes the channel's own credentials by **literal substring**. Composing first and scrubbing afterwards therefore publishes any secret the transform rewrote. Two reachable shapes, both found by @bdunncompany on #3992:

- a configured value containing a repeated space is collapsed before the scrub looks for it;
- one containing `&amp;` is decoded before the scrub looks for it.

Both reach the channel card **and** the test toast in the clear. These are admin-typed values validated as `z.record(z.string(), z.unknown())` with no charset restriction, so neither shape is theoretical.

## Why not "also match the normalised secret"

That was my first design and it is **unsound** — normalisation is not substring-preserving. A configured `abcd&amp` echoed inside `rejected abcd&amp;` decodes to `rejected abcd&`, because the terminating `;` came from *outside* the secret. The transformed text contains neither the raw secret nor any normalisation of it, so no variant-matching scheme can find it. Tag stripping splits a secret the same way (`alpha<b>beta` → `alpha beta`), and truncation can cut a whitespace-containing secret in half — the residual the module's own `truncate` header already documents.

Redacting first is immune to all three, because the secret is gone before any of them run.

## What changed

- Senders pass their own secrets into the composer (`collectChannelSecretStrings(type, config)`), which masks them **before the first lossy step**. The route scrub stays as the second half of the bracket, for messages that never pass through the composer — a thrown `Error.message`, a provider SDK string.
- Masking runs on a window widened by the longest secret, so the 8 KB scan bound cannot bisect an occurrence; and sorts longest-first across *all* secrets, so a shorter secret that is a prefix of a longer one cannot carve it into a fragment that no longer matches.
- `formatHttpFailureDetail` masks too. It is deliberately un-normalised so debugging loses nothing, but nothing downstream scrubs it — and the webhook worker's HMAC signing secret rides that path.
- Also fixes the relative timestamp on the same card, which read "Last test: **Hours Ago**": the private formatter had been extracted into title-cased strings that dropped the `{{count}}` placeholder. It now delegates to the shared `alerts:relativeTime.*` catalog the alerts list already uses. This half was orphaned by the split — #4032 did not carry it.

## Tests

Seven leak cases, each with a **control asserting the leak is reachable without the fix**. All seven fail against a neutered mask (verified, not assumed). Locally: 347 API tests and 131 web tests green, including #4032's empty-state suites.

## Not in scope

The two pre-existing scrub bypasses @bdunncompany found — the case-sensitive host match and the numeric character reference — reproduce on `main` unchanged and are getting their own issue rather than being folded in here.

### DIFF
--- a/apps/api/src/routes/alerts/channels.testOutcomePersist.test.ts
+++ b/apps/api/src/routes/alerts/channels.testOutcomePersist.test.ts
@@ -296,3 +296,65 @@ describe('POST /alerts/channels/:id/test — persisted test outcome (#3697)', ()
     expect(stored).toContain('no_service');
   });
 });
+
+// The reason reaches the operator on TWO surfaces: the persisted card (above)
+// and the toast fired at test time, which renders `testResult.message` from
+// this response. The response carried the RAW provider text, so the card was
+// scrubbed while the toast next to it showed the credential (#3992).
+describe('POST /alerts/channels/:id/test — the toast surface (#3992)', () => {
+  beforeEach(() => {
+    updateSetRef.current = undefined;
+    updateRowCountRef.current = 1;
+    capturedExceptionsRef.current = [];
+    channelRowRef.current = emailChannel();
+  });
+
+  it('does not echo a slack webhook URL back in the response the toast renders', async () => {
+    const webhookUrl = fakeSlackWebhookUrl('YYYYYYYYYYYYYYYYYYYYYYYY');
+    channelRowRef.current = emailChannel({ type: 'slack', config: { webhookUrl } });
+    senderResultRef.current = { success: false, error: `HTTP 404: no_service for ${webhookUrl}` };
+
+    const res = await makeApp().request(`/alerts/channels/${CHANNEL_ID}/test`, { method: 'POST' });
+    const body = await res.json() as { testResult: { success: boolean; message: string } };
+
+    expect(body.testResult.success).toBe(false);
+    expect(body.testResult.message).not.toContain(webhookUrl);
+    expect(body.testResult.message).not.toContain('YYYYYYYYYYYYYYYYYYYYYYYY');
+    expect(body.testResult.message).toContain('no_service');
+  });
+
+  // Card and toast must not diverge: an operator who dismisses the toast and
+  // reads the card should see the same sentence.
+  it('renders the same reason on both surfaces', async () => {
+    senderResultRef.current = { success: false, error: RESEND_ERROR };
+
+    const res = await makeApp().request(`/alerts/channels/${CHANNEL_ID}/test`, { method: 'POST' });
+    const body = await res.json() as { testResult: { message: string } };
+
+    expect(body.testResult.message).toBe(updateSetRef.current?.lastTestError);
+    expect(body.testResult.message).toBe(RESEND_ERROR);
+  });
+
+  it('leaves a successful test message untouched', async () => {
+    senderResultRef.current = { success: true };
+
+    const res = await makeApp().request(`/alerts/channels/${CHANNEL_ID}/test`, { method: 'POST' });
+    const body = await res.json() as { testResult: { success: boolean; message: string } };
+
+    expect(body.testResult.success).toBe(true);
+    expect(body.testResult.message).toBe('Test email sent successfully');
+  });
+
+  // scrubChannelTestError returns null for an empty/whitespace-only message
+  // (never as a result of scrubbing itself — redaction substitutes a non-empty
+  // placeholder). The toast must not go blank when a sender hands us one.
+  it('falls back to a generic reason when scrubbing leaves nothing', async () => {
+    senderResultRef.current = { success: false, error: '   ' };
+
+    const res = await makeApp().request(`/alerts/channels/${CHANNEL_ID}/test`, { method: 'POST' });
+    const body = await res.json() as { testResult: { message: string } };
+
+    expect(body.testResult.message.trim().length).toBeGreaterThan(0);
+    expect(body.testResult.message).toContain('check the channel configuration');
+  });
+});
--- a/apps/api/src/routes/alerts/channels.ts
+++ b/apps/api/src/routes/alerts/channels.ts
@@ -688,6 +688,22 @@ channelsRoutes.post(
       ? null
       : scrubChannelTestError(channel.type, channelConfig, testResult.message);
 
+    // The reason reaches the operator TWICE: the persisted card, and the toast
+    // fired at test time — and the toast renders `testResult.message`, which
+    // until now was the raw provider text. So the card was scrubbed while the
+    // toast still showed the destination URL that, for slack/teams/webhook, IS
+    // the credential (#3992). Both surfaces render the same scrubbed string
+    // from here on. `scrubChannelTestError` returns null only for a non-string
+    // or an empty/whitespace-only message — never as a RESULT of scrubbing,
+    // since every redaction pass substitutes a non-empty placeholder — but the
+    // fallback keeps the toast from going blank when a sender hands us one.
+    if (!testResult.success) {
+      testResult = {
+        ...testResult,
+        message: lastTestError ?? 'Test failed — check the channel configuration.',
+      };
+    }
+
     try {
       const persistResult = await withAuthDbAccessContext(auth, () =>
         db.update(notificationChannels)
--- a/apps/api/src/services/httpFailureMessage.test.ts
+++ b/apps/api/src/services/httpFailureMessage.test.ts
@@ -1,0 +1,405 @@
+import { describe, it, expect } from 'vitest';
+import {
+  formatHttpFailure,
+  formatHttpFailureDetail,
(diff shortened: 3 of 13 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer; plus a **focused re-review** of the truncation rule after that rule changed (it is a security boundary, so it earned a second pass).

**Findings:** 10 raised → 10 addressed across commits `75afdbb61`, `d86121d8b`, `58a6c1bae`, `26bccd4a2`, `548189203`; **0 outstanding.**

Two were self-caught by probing my own fixes rather than waiting to be told, and the sharpest ones were all in code this PR introduced:

| # | finding | source |
|---|---|---|
| 1 | PagerDuty/Pushover lost the raw body on the **live alert path** — their `error` is persisted to `alert_notifications.error_message`, not just shown in a toast | silent-failure-hunter + comment-analyzer |
| 2 | truncation leaked a **17-char** token prefix (60%-of-budget threshold) | self |
| 3 | truncation leaked a **38-char** token prefix (hard-cut fallback) | code-reviewer |
| 4 | truncation leaked a **`TOKENPART1`** fragment (entity-injected boundary inside a secret) | self |
| 5 | truncation leaked a **44-char** `apiKeyValue` prefix, and my "not a live exposure" comment was **factually wrong** — `secretKeysForType('webhook')` covers `authToken`/`authPassword`/`apiKeyValue`, which are admin-typed with no charset restriction | re-review |
| 6 | **quadratic regex** on an unbounded hostile body — measured **45 s** on the main API thread, reachable by anyone who can press "Test channel" | code-reviewer + self |
| 7 | body cut mid-`<!doctype` leaked the fragment | pr-test-analyzer |
| 8 | `webhookDelivery.ts`'s changed line had **zero** coverage | pr-test-analyzer |
| 9 | the boundary test was **vacuous** (URL sat entirely past the cap) | self |
| 10 | two comments asserted things the code does not do | comment-analyzer |

Findings 2–5 are all the same root cause and worth stating plainly: **the 160-char cap runs *before* `scrubChannelTestError`**, which removes credentials by *literal substring* replacement — so any cut through a secret defeats the scrubber and publishes a prefix. Five successive rules were needed; the shipped one cuts **only at whitespace** and drops the window entirely when there is none. Finding 5 was also a genuine **regression**, not an inherited risk: for a 150–500 char body the pre-#3992 `slice(0, 500)` kept the URL whole, so scrubbing removed it completely.

Reported clean by the reviewers: web `isFirstRun` derivation, `onCreate` wiring, locale completeness across all 8, the route's scrub-before-reassign ordering, audit-log integrity, `HTML_TAG` correctness, and a11y.

**Tests:** every new test **revert-probed** — with the source change undone, it fails.

| suite | reverted | with the fix |
|---|---|---|
| `notificationSenders/failureMessage.test.ts` | 3 fail (297 chars > 160) + 2 fail (no log line) | 4 pass |
| `routes/alerts/channels.testOutcomePersist.test.ts` | 2 fail (webhook URL echoed in the response) | 11 pass |
| `services/httpFailureMessage.test.ts` boundary rules | 3 fail (token fragment survives) | 23 pass |
| `services/httpFailureMessage.test.ts` ReDoS bound | fails at **45,236 ms** | passes < 1 ms |
| 3 web suites (empty states + timestamp) | 14 fail | green |

- API affected: **26 files / 319 tests** green (single-fork)
- Web `components/alerts/` + `components/configurationPolicies/` + `lib/i18n/`: **52 files / 528 tests** green (locale parity + duplicate baselines hold)
- `tsc --noEmit` (api) clean · `astro check` (web) 0 errors across 1680 files · `eslint` clean
- **CI: 54/54 green**, including all 4 Integration Tests shards, Test API, Test Web and Type Check

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3992 — the issue came from an internal Playwright QA sweep, so a quick look at the two cards in a running stack before closing is worth it.

ToddHebebrand:
Re-scoping this PR, and recording a confirmed defect in it before I do.

**Part 1 is gone — @bdunncompany's #4032 shipped it** (`2aaa38495`). He split #3992 himself and asked first, and his half is the better-scoped version: the same derived-boolean approach, tests that render a real row before searching so the control is genuine, and the create action wired through rather than left dangling. This PR drops its `1020eb786` commit and keeps only part 2 (the readable channel-test failure reason).

**His normalise-before-scrub warning is correct, and I verified it against this branch rather than taking it on faith.** `summarizeResponseBody` (`apps/api/src/services/httpFailureMessage.ts:186-206`) runs `decodeEntities` and then `.replace(/\s+/g, ' ')`, and the composed string only reaches `scrubChannelTestError` afterwards at `apps/api/src/routes/alerts/channels.ts:689`. That scrub removes credentials by literal substring, so any transform in between can make a configured secret stop matching itself:

- a secret containing a repeated space is collapsed to a single space before the scrub looks for it — no match, published;
- a secret whose literal text contains `&amp;` is decoded to `&` before the scrub looks for it — no match, published.

`secretKeysForType('webhook')` covers `authToken`, `authPassword` and `apiKeyValue`, all admin-typed against `z.record(z.string(), z.unknown())` with no charset restriction, so both shapes are reachable configurations rather than theoretical ones.

**This is not the case the module already reasons about.** The `truncate` header goes four rounds on keeping the scrubber's literal match intact and states its residual as "a secret that itself contains whitespace could still be split." That covers *cutting*. It does not cover *rewriting* — and rewriting is what the decode and collapse steps do to every message, at full length, whether or not truncation ever engages. The existing reasoning reads as though the surface were closed, which is what made this easy to miss.

**Fix is his:** bracket the transform — scrub the raw body, normalise, then scrub again — so a secret that survives in either form is caught. Cheap, and it makes the ordering non-load-bearing instead of a standing invariant someone has to remember.

Not merging this until that lands. The two pre-existing bypasses he found — the case-sensitive host match and the numeric character reference — are separate from this and are getting their own issue rather than being folded in here.

PR #4016, opened 2026-08-25
fix(web): put the alert-rule Test verdict on the live editor (#3988)

### LINKED ISSUE #3988: [UI] Alert rule "Test" verdict fix (#3923) shipped to an unroutable component — no Test action exists in the live rule editor
## Description

PR #3923 ("alert rule 'Test' reports the real evaluation verdict") landed a correct, well-tested fix in `apps/web/src/components/alerts/AlertRulesPage.tsx` and `AlertRuleList.tsx` — but **neither component is reachable from any route**, so no user can see the fix.

Verified in a browser against `main` @ `c2f9db1ce`:

1. All three alert-rule pages are pure redirects:
   - `apps/web/src/pages/alerts/rules/index.astro` → `return Astro.redirect('/configuration-policies', 301)`
   - `apps/web/src/pages/alerts/rules/[id].astro` → same
   - `apps/web/src/pages/alerts/rules/new.astro` → same

   Clicking **Alerts → Rules** in the UI lands on `/configuration-policies`.

2. `AlertRuleList` is imported only by `AlertRulesPage.tsx` and by the barrel `apps/web/src/components/alerts/index.ts`. That barrel is imported by **nothing** (`grep -rn "from '.*components/alerts'" apps/web/src` outside the folder returns no hits).

3. The **live** alert-rule editor is `apps/web/src/components/configurationPolicies/featureTabs/AlertRuleTab.tsx`. Confirmed in-browser (created a policy → Alerts → Add Alert Rule): the editor offers Metric / Operator / Value / Cooldown / Auto-Resolve and **no Test action anywhere**.

The API half of #3923 (`apps/api/src/routes/alerts/rules.ts` + `rules.testVerdict.test.ts`) is real and does return a true verdict — it just has no UI caller.

## Impact

The user-facing benefit of #3923 is zero: there is no button in the product that runs an alert-rule test. The web tests (`AlertRulesPage.testVerdict.test.tsx`) pass against a component that is never mounted, so CI cannot detect this.

## Proposed fix

Decide which is intended:

- **(a)** Port the verdict UI into `AlertRuleTab.tsx` (the live editor) and delete `AlertRulesPage.tsx` / `AlertRuleList.tsx` + the barrel entry.
- **(b)** Restore a real route for `AlertRulesPage` if the standalone rules page is still wanted.

Either way, adding a guard that fails when a component has tests but no routed importer would prevent the class.

## Affected files

- `apps/web/src/pages/alerts/rules/{index,[id],new}.astro`
- `apps/web/src/components/alerts/AlertRulesPage.tsx`, `AlertRuleList.tsx`, `index.ts`
- `apps/web/src/components/configurationPolicies/featureTabs/AlertRuleTab.tsx`

ToddHebebrand:
Closes #3988

## The problem

PR #3923 fixed the alert-rule "Test" modal so it reports the real evaluation verdict instead of a fabricated green — and shipped it into a component no route can reach.

All three `apps/web/src/pages/alerts/rules/{index,new,[id]}.astro` have been a bare `Astro.redirect('/configuration-policies', 301)` since `d8a6bc833` (**2026-02-22**) — *six months* before #3923 merged — and `components/alerts/index.ts` is imported by nothing. So `AlertRulesPage.tsx` was never mounted, its 346 lines of tests passed against a component no user could see, and CI had no way to notice.

The live alert-rule editor is the Configuration Policy **Alerts** tab (`configurationPolicies/featureTabs/AlertRuleTab.tsx`). It had no Test action at all.

## Why this wasn't a one-line port

The tab's rules are **not** `alert_rules` rows. They are `config_policy_alert_rules` rows, and:

- `assembleInlineSettings()` strips `r.id` from every item the read path returns — the editor never sees a row id;
- `decomposeInlineSettings()` + `deleteNormalizedRows()` **delete and recreate** the whole set on every save, so the id isn't stable anyway;
- targeting doesn't live on the rule — it comes from the policy's assignments.

`POST /alerts/rules/:id/test` addresses a rule *by id* and reads `targetType`/`targetId` off it, so it is structurally unreachable from this editor. A new endpoint was required.

## What this adds

**`POST /configuration-policies/:id/alert-rules/test`** — body `{ deviceId, conditions }`, returning the same verdict shape as the standalone endpoint so the #3923 rendering and its `alerts:alertRulesPage.*` locale keys are reused verbatim.

It takes the **draft conditions**, not a row id. That is the point twice over: there is no id to send, and a tech tuning a threshold wants a verdict on the number they just typed, not on what was last saved.

**Targeting** — a new `resolveGoverningAlertRulePolicyForDevice(deviceId, candidatePolicyId)`. It deliberately does not reuse `resolveAlertRulesForDevice`, and is not a link-only resolver either — both answer the wrong question here:

| Approach | Failure |
|---|---|
| `resolveAlertRulesForDevice` (inner-joins rule rows) | A policy whose first rule is unsaved has no rows, so it can never win — the draft's own policy would always report "does not govern this device". |
| Resolve on the `alert_rule` feature link alone | A policy holding an **empty** link would outrank one that actually has rules — which is not what happens at runtime. |

So the candidate is overlaid onto real runtime behaviour: the candidate policy competes as though its draft were already saved, every *other* policy competes only if it currently holds at least one persisted alert rule, and the ordinary hierarchy sort (level → priority → age, with role/OS filters and the ownership predicate unchanged) picks the winner.

`wouldTrigger = targetMatch && evaluation.triggered`, and each negative names its own remedy — "not assigned to this device" and "another policy takes precedence" are different reasons, because they need different fixes.

**UI** — a `FlaskConical` **Test rule** button on each expanded rule card opens `AlertRuleTestModal.tsx` (extracted next to the tab rather than growing a 1256-line file). It is disabled while the rule holds a retired condition type or unsupported metric: the server would refuse the test body for the same reason it refuses the save, and the existing amber banner already says what to fix.

## Tenancy

The governance resolver runs under `withPartnerWideVisibility` — a **system RLS context**, because partner-wide policies are invisible under an org-scoped one — so it can never be the tenancy boundary. The route authorizes both inputs itself, before anything is evaluated:

1. the policy, via `getConfigPolicy(id, auth)` (org/partner access condition);
2. the device, on the org axis (`auth.canAccessOrg`) **and** the site axis (`siteAccessCheck` — RLS does not defend sites).

A denied org returns the byte-identical 404 body as a missing device, so the endpoint is not a tenant-existence oracle. `targetReason` never identifies the policy that won instead — it may belong to another org under the partner.

No schema change, no migration, no new tenant-scoped table, so no cascade/export-policy registration applies.

## Accuracy fix carried along

The suppression caveat shown beside a positive verdict listed cooldown, open-alert dedup and flapping. For the config-policy path, `evaluateDeviceAlertsFromPolicy()` skips the device entirely while a maintenance window with `suppressAlerts` is active — *before* any of those. The note now leads with it, in all 8 locales.

## Deleted

`AlertRulesPage.tsx` (485 lines), `AlertRuleList.tsx` (403), their barrel entries, the orphaned `AlertRulesPage.testVerdict.test.tsx` (346), and the `runActionAllowlist` migration-backlog entry. Every importer was grepped first; `AlertsTabStrip` stays (still used by AlertsPage / NotificationChannelsPage / correlations.astro), and the 301 redirect pages stay — they serve bookmarks and the deliberate "managed in Configuration Policies" story.

## Tests

Every meaningful case from the deleted suite is ported onto the new call site as `AlertRuleTab.testVerdict.test.tsx` (14 tests), plus new coverage for testing the unsaved draft and the retired-condition gate. Two cases did not carry over, deliberately: config-policy rules have no per-rule `enabled` column, and there is no rule id to pin in the request path.

**Revert probes, both run:**

- Web — reintroduced the #3752 bug (`data.success ?? true`) in the modal: **4 tests fail** (would-not-fire, targeting negative, missing verdict, positive-vs-negative divergence). Restored, green.
- API — made `wouldTrigger` ignore `targetMatch`: **2 tests fail** (both targeting negatives). Restored, green.

Local runs:

- `apps/api` `routes/configurationPolicies/` + `routes/alerts/` — 24 files, **353 passed**
- `apps/api` `services/featureConfigResolver*` — 3 files, **45 passed**
- `apps/web` `components/configurationPolicies/` + `components/alerts/` — 43 files, **380 passed**
- `apps/web` locale parity + i18n + `no-silent-mutations` — **205 passed**
- `tsc --noEmit -p apps/api/tsconfig.json` — clean
- `astro check` — **0 errors, 0 warnings**
- eslint on every changed source file — clean

## Notes for the reviewer

- **Permission choice.** The endpoint uses `requireConfigPolicyRead` (`devices:read`), matching the sibling feature-link routes that already create and read these exact rules — not `alerts:read` as the standalone endpoint does. Requiring `alerts:read` would let a user author a rule but not test it. Happy to add it if you'd rather have both.
- **Still-dead siblings, left alone on purpose.** `AlertRuleEditPage.tsx`, `AlertRuleForm.tsx` and `AlertRuleEditor.tsx` are also barrel-only, but they carry their own tests and ~1,500 lines, and removing them is a separate sweep rather than blast radius on this fix. Say the word and I'll file a follow-up.
- **Review round 1 is in `41a28ceef`.** Findings acted on: the verdict outcome became a 3-variant union (a 2x2 product made an impossible-but-renderable "another policy takes precedence, naming no policy" state), the resolver gained 9 direct unit tests with a revert probe, the delete-while-testing invariant gained coverage, `conditions: unknown[]` was narrowed, and two comments were factually wrong — the suppression caveat named `createAlert()` (the standalone path; config-policy rules fire through `evaluateDeviceAlertsFromPolicy`, which duplicates those checks inline), and the redirect date/commit above was wrong, now corrected.
- **Declined, with reasons.** A shared Zod schema for the *response* (so the client's `RuleTestVerdict` can't drift) is a real improvement but strictly over-strict parsing risk on a brand-new endpoint whose two sides ship together — worth a follow-up, not this PR. The 401-branch "idle-looking modal" is a pre-existing app-wide convention (a dozen-plus call sites) that this PR inherits, not a regression. Folding `submitting` into `TestState` is cosmetic.
- **Residual locale keys.** The `alerts:alertRulesPage.*` block keeps a handful of now-unused page-chrome strings (`newRule`, `deletingRule`, …). Pruning them is 8 files of churn for no functional gain; flagged rather than done.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer (5 lenses, run in parallel against `git diff origin/main...HEAD`).

**Findings:** 9 raised → 6 addressed in `41a28ceef`, 3 declined with reasons. 0 Critical from the code/security lenses; the one Critical came from comment-analyzer and was a factual error in a comment, now corrected.

Addressed:
- **Verdict outcome type** (type-design, High) — `{ winningPolicyId: string | null; candidateAssigned: boolean }` is a 2x2 product whose fourth combination is unreachable today but, if ever produced, would render *"Another configuration policy takes precedence"* naming a policy that does not exist. That is the exact fabricated-verdict class this endpoint closes, so it is now a 3-variant union (`governs` / `outranked` / `unassigned`) and the route branches on the discriminant instead of re-deriving the match.
- **`createAlert()` cited for the wrong path** (comment-analyzer, Critical) — verified: `evaluateDeviceAlertsFromPolicy` never calls it. It duplicates cooldown (`isConfigPolicyRuleCooling`), open-alert dedup (an `alerts.configPolicyId` lookup) and flapping inline, then inserts directly. The sentence was copy-pasted from the sibling standalone endpoint, where it is true. Fixed in the route and the modal.
- **Redirect date was wrong** (comment-analyzer) — verified against the real `main` history: the 301 landed in `d8a6bc833` on **2026-02-22**, not `4d0cc04bb`/2026-07-23 (that commit is unrelated API work). `AlertRulesPage` was unroutable for *six* months before #3923 shipped into it. Corrected in three tombstone comments and in the PR body.
- **Resolver had no direct tests** (code-reviewer + pr-test-analyzer, independently) — added 9 cases to `featureConfigResolver.test.ts` on that file's existing mocked-drizzle harness. Load-bearing case: a ruleless competitor assigned at a *closer* level must not outrank the candidate.
- **`testingIndex` shift-on-delete untested** (pr-test-analyzer) — the tab comment claimed an invariant nothing exercised. Two cases added.
- **`conditions: unknown[]`** (type-design) — narrowed to the editor's exported `Condition[]`; also added the missing `.orderBy` for tie determinism parity with `resolveAlertRulesForDevice` (code-reviewer suggestion).

Declined, with reasons: a shared Zod schema for the *response* (real improvement, but over-strict-parse risk on a new endpoint whose two sides ship together — follow-up, not this PR); the 401-branch "idle-looking modal" (pre-existing app-wide convention at a dozen-plus call sites, not a regression here); folding `submitting` into `TestState` (cosmetic).

**Revert probes — 3 run, all had teeth:**
| Mutation | Result |
|---|---|
| Modal reads `data.success ?? true` (the #3752 bug) | 4 web tests fail |
| `wouldTrigger` ignores `targetMatch` | 2 API tests fail |
| `contenders` drops the `haveRules` filter | the ruleless-competitor case fails |

All restored; suites green after each.

**Tests:** apps/api `routes/configurationPolicies` + `routes/alerts` + `services/featureConfigResolver` — 27 files, 406 passed. apps/web `components/configurationPolicies` + `components/alerts` + `lib/i18n` + `no-silent-mutations` — 51 files, 627 passed. `tsc --noEmit -p apps/api/tsconfig.json` clean; `astro check` 0 errors / 0 warnings; eslint clean on every changed source file.

**CI:** all required checks green — Type Check, Lint, Test API, Test Web, Build Web, Smoke Test, and all 4 Integration Tests shards. 0 failures.

**Status:** review-clean, awaiting maintainer merge. Two open calls for you, both in the PR body: whether the endpoint should also require `alerts:read` on top of `devices:read`, and whether to sweep the three remaining barrel-only alert-rule components (`AlertRuleEditPage`/`AlertRuleForm`/`AlertRuleEditor`) in a follow-up.

Issue #4024, opened 2026-08-25
File Browser errors are blanked on hosted: agent messages returned as 502/504 are replaced by Cloudflare's error page


ToddHebebrand:
### Description

On hosted deployments, **every** File Browser error message is replaced with a generic "Failed to download" / "Failed to copy" string. The technician never sees the reason the device actually gave.

Confirmed end-to-end in US production today.

A user tried to download an 8.9 MB PDF from `C:\temp\New instal\Install\Manuals\` via Remote Tools → File Browser. The Transfers panel showed only **"Failed to download"**, twice.

What actually happened, traced through prod:

1. **The agent refused the file — correctly.** `maxFileReadSize` is 1 MB (`agent/internal/remote/tools/fileops.go:16`, enforced `:506`). `device_commands` in prod:

   ```
   file_read | failed | {"error": "file too large: 8981850 bytes (max 1048576 bytes)"}
   ```

2. **The API returned that message with HTTP 502.** `mapCommandFailure()` in `apps/api/src/routes/systemTools/fileBrowserHelpers.ts` uses 502 as its catch-all for agent failures and 504 for timeouts. Prod API log confirms `status=502` on both attempts.

3. **Cloudflare threw the body away.** Per [Cloudflare's docs](https://developers.cloudflare.com/support/troubleshooting/http-status-codes/cloudflare-5xx-errors/error-502-504/): *"Cloudflare returns a Cloudflare-branded HTTP 502 or 504 error when your origin web server responds with a standard HTTP 502 bad gateway or 504 gateway timeout error."* Origin-error pass-thru is Enterprise-only and is not enabled on this zone.

4. **The client fell back to the generic string.** `FileManager.tsx` does `await res.json().catch(() => ({ error: t('fileManager.errors.download') }))`. Cloudflare's HTML body fails `.json()`, so the fallback wins.

### Blast radius

502 is `mapCommandFailure`'s catch-all and 504 its timeout branch — **exactly the two statuses Cloudflare intercepts**. So this silently blanks the reason for every File Browser failure on hosted: size caps, containment denials, missing paths, permission errors, timeouts.

Bulk copy/move/delete are hit too: they answer 502 when every item fails, so the whole per-item `results` array is discarded at the edge and the UI can only say "Copy failed".

Self-hosted deployments have no Cloudflare in front and have seen the real messages all along — which is why this survived undetected.

### Expected Behavior

[...]

PR #4028, opened 2026-08-25
test(web): update DevicesPage assertions to the Remove copy that #3994 shipped

ToddHebebrand:
## Why

`main` is currently red on **Test Web**.

#3994 renamed the device *Decommission* action to *Remove*, including `deviceActions.confirm.decommission.{title,message,confirm}` in `en/devices.json`. Three assertions in `DevicesPage.test.tsx` still expect the pre-rename strings, so #4009's confirm-gate tests fail:

```
TestingLibraryElementError: Unable to find an element with the text: /decommission host-alpha\?/i
Unable to find an element with the text: Decommission Device
```

## What

Copy-only. The behaviour under test — #4009's confirm gate on the row/grid kebab — is unchanged and still asserted.

| assertion | was | now |
|---|---|---|
| dialog title (×2) | `Decommission Device` | `Remove Device` |
| dialog message | `/decommission host-alpha\?/i` | `/remove host-alpha\?/i` |
| confirm button | `Decommission` | `Remove` |

## Verification

`pnpm vitest run src/components/devices/DevicesPage.test.tsx` → **63 passed** (was 2 failed | 61 passed).

Found while investigating an unrelated File Browser bug (#4024) — the full web suite on a fresh branch off `main` surfaced it.

### DIFF
--- a/apps/web/src/components/devices/DevicesPage.test.tsx
+++ b/apps/web/src/components/devices/DevicesPage.test.tsx
@@ -1506,11 +1506,11 @@ describe('DevicesPage — decommission from the row/grid kebab is confirm-gated 
 
     // The SAME keys DeviceActions.tsx renders — proving no new copy was needed
     // and that the two screens still read identically.
-    expect(await screen.findByText('Decommission Device')).toBeTruthy();
-    expect(screen.getByText(/decommission host-alpha\?/i)).toBeTruthy();
+    expect(await screen.findByText('Remove Device')).toBeTruthy();
+    expect(screen.getByText(/remove host-alpha\?/i)).toBeTruthy();
 
     const confirmBtn = await screen.findByTestId('confirm-device-action');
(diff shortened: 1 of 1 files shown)


ToddHebebrand:
**Superseded by #4023, which merged as `1fd01dbb9` — closing.** Same diagnosis, same four assertions, same replacement values; we were both fixing main's `Test Web` break from #4011 + #3994 landing back to back, and #4023 got there first. Main is verified green on that SHA by a full CI run, so this branch is now a no-op against `main`.

Nothing wrong with the change — for the record, the analysis matched: the `decommission` action id never moved, only the rendered copy (`deviceActions.confirm.decommission.title` → "Remove Device", `.confirm` → "Remove", and the message body).

No action needed. Flagging the collision itself, since it will recur: the two source PRs had **no textual conflict** — their `DevicesPage.tsx` hunks genuinely don't overlap — so git merged them clean and only the shared test file caught it.

Issue #1327, opened 2026-06-13
[UI] Surface all ticketing options on the Partner settings page (match Org page layout)


ToddHebebrand:
## Description

Product request (Todd): *"The Partner page needs all the ticketing options. Maybe it can match the layout of the Org pages."*

The **Organization settings page** surfaces ticketing as a first-class, integrated tab inside its settings hub. The **Partner settings page** has no ticketing tab at all — partner-wide ticketing config (statuses, priorities, categories, billing export) lives on a *separate, orphaned page* at `/settings/ticketing` that isn't wired into the partner settings hub or any nav alongside the other partner tabs. The request is to bring all ticketing options onto the Partner page and match the Org page's tabbed layout.

Good news from investigation: **the backend already scopes the core ticketing config at the partner level** — this is mostly a web information-architecture / wiring task, not a schema migration. See the design question below for the one place a decision (and possibly schema work) is needed.

## Current State

**Org settings hub** — `apps/web/src/components/settings/OrgSettingsPage.tsx:30-79` — 8 hash-routed tabs, ticketing integrated as one of them:

| Org tab | Component |
|---|---|
| General / Branding / Customer Portal / Notifications / Security / Event Logs / Remote Access | (various) |
| **Ticketing** (`#ticketing`) → SLA overrides + billing defaults | `OrgTicketSettingsEditor.tsx:35-263` (`OrgSettingsPage.tsx:442-444`) |

Org-level ticketing is intentionally narrow: **per-org SLA overrides** (`OrgTicketSettingsEditor.tsx:162-211`) and **billing defaults** — default hourly rate + billable (`OrgTicketSettingsEditor.tsx:213-248`). These *override* partner-wide defaults; blank cells inherit.

**Partner settings hub** — `apps/web/src/components/settings/PartnerSettingsPage.tsx` (`TABS` at lines 52-62) — 9 tabs: Company, Regional, Security, Notifications, Event Logs, Defaults, Branding, AI Budgets, Remote. **Zero ticketing tabs.** (`apps/web/src/pages/settings/partner.astro`.)

**Partner-wide ticketing config that EXISTS but lives on an orphaned page** — `apps/web/src/components/settings/TicketingSettingsPage.tsx:33-99`, served at `apps/web/src/pages/settings/ticketing.astro`. Four tabs, all already partner-scoped and fully reusable (no org context):

| Tab | Component | What it does |
|---|---|---|
[...]

PR #4030, opened 2026-08-25
fix(web): react to late-arriving JWT scope in ticketing tabs + patch approval (#4013)

### LINKED ISSUE #4013: [Web] Ticketing settings: Intake Forms / Inbound Email / Canned Responses tabs are hidden from partner users on a cold load (frozen useMemo over JWT scope)
## Description

`TicketingSettingsTabs` decides whether to offer the **Intake Forms**, **Inbound Email** and **Canned Responses** tabs by reading the JWT scope into a `useMemo` with an **empty dependency array**. The value is therefore computed exactly once, on the component's first render — and on a cold page load that render happens before an access token exists, so it is pinned `false` for the entire life of the mount.

`apps/web/src/components/settings/TicketingSettingsTabs.tsx:80`

```ts
const canManageInbound = useMemo(() => getJwtClaims().scope === 'partner', []);
```

Same root cause as #4010: `getJwtClaims()` (`apps/web/src/lib/authScope.ts`) reads `useAuthStore.getState().tokens?.accessToken`, and access tokens are deliberately **not** persisted — the `breeze-auth` localStorage blob holds only `user` and `isAuthenticated` (see `partialize` in `apps/web/src/stores/auth.ts`). At first paint the store is empty, so `scope` is `null` for every user regardless of who they are.

It is a **different shape** from #4010, though, and that is why the fix there does not cover it. #4010 was a guard that *wrote to the URL* before the scope existed. This one never touches the URL — it gates render only. What makes it worse in one respect is the empty dep array: #4010's effect at least re-ran when scope arrived (it just had nothing left to read), whereas `useMemo(..., [])` never recomputes at all.

## Impact

A partner-scoped user who lands **directly** on the ticketing settings surface — rather than clicking into it after the app has already authenticated — sees three tabs missing, with no error and nothing to retry. A reload does not help; only navigating in from an already-warm session does.

That direct-landing path is not exotic. `apps/web/src/pages/settings/ticketing.astro` is a permanent 301 alias to `/settings/partner#ticketing` (kept for old bookmarks, #1327), and `PartnerSettingsPage.tsx:684` mounts the component as soon as the hash-restored tab resolves to `ticketing`. So every old `/settings/ticketing` bookmark and every `/settings/partner#ticketing` link lands on exactly the cold-load path.

The three gated surfaces:

- `activeTab === 'forms' && canManageInbound` — Intake Forms
- `activeTab === 'inbound' && canManageInbound` — Inbound Email (+ the M365 mailbox card)
- `activeTab === 'canned' && canManageInbound` — Canned Responses

The tab list itself is built from the same flag (`TABS`, `:87-98`), so the tabs are absent from the tab bar, not merely inert.

## Not a security hole

The gate is a UX hint only — the comment at `:77-79` says so and the server re-checks every request. Failing closed means a partner user is under-served, never that an org user gains access. This is a usability/regression issue, not an authorization one.

## Suggested fix

#4010 shipped the tool for this: `useJwtClaims()` in `apps/web/src/lib/authScope.ts` is a reactive form of `getJwtClaims()` that subscribes to the access token, so the component re-renders when it lands. The change is roughly:

```ts
const { scope } = useJwtClaims();
const canManageInbound = scope === 'partner';
```

which also deletes the frozen memo. Worth confirming at the same time that nothing else in the file depends on `canManageInbound` being stable across the mount.

## Worth checking in the same pass

Other one-shot `getJwtClaims()` reads that are memoized or captured and then relied on after mount. A repo sweep for #4010 found no other instance of *that* issue's URL-writing shape, but it was scoped to hash-writing code and would not have flagged render-only gates like this one. `grep -rn "getJwtClaims" apps/web/src` is the starting point; the ones to look at are any inside `useMemo`/`useState`/`useRef` initializers or `[]`-dep effects.

## Affected files

- `apps/web/src/components/settings/TicketingSettingsTabs.tsx` (`:80` the frozen memo; `:87-98` the tab list; `:149`, `:157`, `:165` the gated panels)
- `apps/web/src/components/settings/PartnerSettingsPage.tsx` (`:684` the mount site)
- `apps/web/src/pages/settings/ticketing.astro` (the 301 that routes bookmarks straight into the cold-load path)

Found while sweeping sibling pages for #4010 (PR #4012). Filed rather than folded into that PR to keep the fix scoped.

ToddHebebrand:
## Problem

Two components read the JWT scope into a `useMemo` with an **empty dependency array**, so the value is computed once, on the component's first render. Access tokens are deliberately never persisted (`partialize` in `stores/auth.ts` keeps only `user` and `isAuthenticated`), so on every cold load that first render happens with an empty store and **every** user decodes as `scope: null`. The memo never recomputes, so the empty-store answer is pinned for the life of the mount.

`grep -rn "useMemo(() => getJwtClaims" apps/web/src` returns exactly these two, across 31 `getJwtClaims()` call sites. Both are fixed here.

### 1. `TicketingSettingsTabs.tsx:80` — confirmed live, and it fails in the user's face

`canManageInbound` pinned `false` hides **Intake Forms**, **Inbound Email** and **Canned Responses** from partner users — not just the panels, the tab bar entries too. A reload does not help; only navigating in from an already-warm session does.

Two entry points land there cold by design:

- `pages/settings/ticketing.astro` is a permanent 301 to `/settings/partner#ticketing` (#1327), so every old bookmark is a cold load;
- `PartnerSettingsPage.tsx:684` passes `initialTab='inbound'` on the M365 consent return (`?ticketMailbox=…`), which is a full-page navigation back from Microsoft — i.e. always cold. That user landed on the Inbound sub-tab with an **empty body and no error**.

To be precise about *when* the user sees this: `AuthOverlay` masks the whole window while the access token is absent, so the unresolved window itself is generally hidden. The harm is what is on screen **after** the overlay lifts — the memo has already frozen by then and never recomputes, so the tabs are missing for the rest of the mount. Review round 1 corrected an earlier version of this section that implied users watch a blank body during the token round trip.

### 2. `PatchApprovalModal.tsx:87` — the issue graded this "probably not reachable". It is reachable.

The comment on #4013 reasoned that a modal only opens after interaction, by which point the token has arrived. That is true of the *open*, but not of the *mount*: `PatchesPage.tsx:902` renders `<PatchApprovalModal>` **unconditionally** — it sits in the page body, not behind `{modalOpen && ...}` — and PatchesPage has no early return in front of it (one `return (` at `:737`). `patches/index.astro` hydrates it `client:load`, and `DashboardLayout` puts `AuthOverlay` beside the slot rather than wrapping it. `if (!patch) return null` sits *after* the hooks, so they run anyway.

So the modal's first render is the page's cold load, and that is where the memo froze. For an org-scoped user who cold-loads `/patches`, `isOrgScope` stays `false` for the life of the page: `canSubmit` (`:90`) enables Approve, the partner-level explainer (`:240`) never renders, and the disabled-button tooltip (`:266`) is absent. They are offered an Approve button, click through the scope-naming confirm dialog, and only then get told no.

**It is not an authorization hole.** `handleSubmit` re-reads the scope live and blocks the submit, and the server re-checks every request. The damage is that the refusal arrives as an error after the click instead of as an explanation before it.

This is demonstrated end-to-end, not argued: the new `PatchesPage.scopeResolution.test.tsx` case drives the real page with the real auth store and **fails against the current code**.

## Fix

Both sites move to `useJwtClaims()` — added by #4012 (`lib/authScope.ts`) for exactly this failure mode. It subscribes to the access token and returns a discriminated `{ status: 'unresolved' } | { status: 'resolved'; claims }`. `getJwtClaims()` is untouched (~30 other callers), and no second reactive-claims abstraction was introduced.

Neither site collapses the three states back to two, and they resolve the unknown window **differently**, because what they gate is different:

**`TicketingSettingsTabs`** keeps `inboundAccess: 'unresolved' | 'allowed' | 'denied'`, and the partner-only tab list is now *built* from one `PARTNER_ONLY_TABS` constant that the placeholder gate also reads. Both non-`'allowed'` states fail closed for *rendering* — an unknown scope is never a grant, so the partner-only tabs stay out of the tab bar either way. What they do not share is the *explanation*: when the active sub-tab is partner-only and the scope is still unresolved, a small "Checking your access..." placeholder renders instead of a blank body. Given the AuthOverlay note above, that placeholder is mostly a backstop rather than a common sight — it is kept because it is the only DOM difference between `'unresolved'` and `'denied'`, without which nothing would notice a future collapse back to a boolean. A settled `'denied'` renders nothing, exactly as before. Nothing in this component writes the URL or discards state, so there is no #4010-style destructive branch that needs deferring — the hash survives the unresolved window untouched, which the tests pin.

**`PatchApprovalModal`** keeps `approvalAccess: 'unresolved' | 'allowed' | 'denied'` (named for parity with `ringAccess` and `inboundAccess`).

- `'unresolved'` **disables submit** — an unknown scope is not a grant, so we do not offer an Approve the API would refuse;
- `'unresolved'` **does not** show the "managed at the partner level" banner. It gets a neutral "Checking your access..." one in the same slot instead — we have not been told no, and asserting a denial we have not received is the same class of mistake as #4010 in the other direction. Both non-`'allowed'` states get a *visible* reason rather than a `title` tooltip alone, since a disabled button is out of the tab order and has no hover on touch (review round 1).
- Only a *resolved* org scope is a denial. That keeps the render gate byte-identical to the predicate `handleSubmit` and the server enforce (`scope === 'organization'`).

A present-but-undecodable token is `'resolved'` with all-null claims, so it stays `'allowed'` here — deliberately unchanged, and deliberately the opposite of what `ringAccess` does with the same input (a cross-reference comment now records why, so the two are not "harmonized" by mistake). Telling someone whose token cannot be decoded that "approvals are managed at the partner level" would be the wrong explanation for what is really a 401 on their next request; the server refuses it either way. Pinned by a test so the choice is visible rather than incidental.

`getJwtClaims()` stays in `handleSubmit`, which is where a one-shot read is the *correct* shape — an event handler wants the scope as of the click and the answer cannot outlive the call.

### Not widened

The sweep also turns up render-time reads that are not memoized (`CatalogItemsTab.tsx:66`, `CatalogSettingsPage.tsx:13`, `SsoProvidersPage.tsx:42`, …). Those recompute on every render, so they are a weaker shape than this one — they self-heal on any incidental re-render rather than freezing forever. They are worth a separate pass, not a silent widening of this PR.

Also left alone: a **pre-existing** gap where a settled `'denied'` on a partner-only sub-tab renders a blank body (an org user who hand-crafts `#tab=inbound`). Raised in review, and real — but it is not what #4013 broke, and it needs its own copy decision. Recorded in a code comment.

## Tests

Two new suites plus two cases on an existing one. All three mock **neither** `authScope` nor the auth store — they drive the real `useJwtClaims` against the real store so the token genuinely arrives after mount, which is the only way to reproduce this. (The pre-existing suites for both components mock `authScope` with a scope that is *already known on the first render*, i.e. the one state a cold load never starts in. That is why the bug shipped past them.)

- **`TicketingSettingsTabs.coldLoad.test.tsx`** (9 cases) — partner scope arriving after first paint reveals all three tabs; `#tab=forms|inbound|canned` deep links show the pending placeholder and then the real panel; the M365 `initialTab='inbound'` path; org scope arriving after first paint keeps the tabs hidden *and* stops saying "pending"; a `hashchange` onto a partner-only sub-tab during the unresolved window defers rather than denies; a token going away again re-hides rather than keeping a stale grant; an undecodable token is a settled denial rather than a permanent spinner.
- **`PatchApprovalModal.coldLoad.test.tsx`** (5 cases) — each mounts the modal **closed with `patch={null}` first** and opens it later, which is the sequence the page actually produces. Covers org-after-mount, partner-after-mount, the unresolved window (disabled but not denied, with a *visible* reason), a token going away again, and the undecodable token.
- **`PatchesPage.scopeResolution.test.tsx`** (+2 cases) — the reachability proof: cold-load the real page, let the token land, click Review, assert what the modal offers.

**Revert probe** (restore both `useMemo(..., [])` lines, keep the tests):

| Suite | Result against reverted code |
|---|---|
| `TicketingSettingsTabs.coldLoad` | **9 of 9 fail** |
| `PatchApprovalModal.coldLoad` | **4 of 5 fail** |
| `PatchesPage.scopeResolution` (new cases) | **1 of 2 fails** |

The one that passes either way in each of those suites is the partner-scope-arrives case, and that is expected rather than a gap: a frozen `false` and a correctly-resolved partner scope reach the *same* final state (Approve enabled, no banner). Their job is the other direction — proving the fix does not over-correct into blocking legitimate partner users. The unresolved window itself, which is the only place those two differ, is covered by its own discriminating case.

## Verification

- full `apps/web` suite → **640 files, 6506 tests passed**
- `astro check` → **0 errors** (1684 files)
- `eslint` on all changed/added `.tsx` files → clean

Rebased onto current `main` (`1fd01dbb9`). The branch was originally cut at `9ff8b0f1a`, which predates #4023's fix for `DevicesPage.test.tsx` — the two failures that file showed locally reproduce with this PR's changes reverted, so they were inherited from the stale base, not caused here. Green across the board after the rebase.

Locale parity: the new `checkingAccess` key is added to `settings.json` and `patches.json` in all 8 locales (en, de-DE, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR); `localeParity.test.ts` passes.

Closes #4013

### DIFF
--- a/apps/web/src/components/patches/PatchApprovalModal.coldLoad.test.tsx
+++ b/apps/web/src/components/patches/PatchApprovalModal.coldLoad.test.tsx
@@ -1,0 +1,129 @@
+import { act, render, screen, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import '@/lib/i18n';
+
+// #4013. PatchApprovalModal.test.tsx mocks `lib/authScope` with a scope that is
+// already known on the first render — the one state a cold page load never
+// starts in. Access tokens are deliberately never persisted (`partialize` in
+// stores/auth.ts keeps only `user` + `isAuthenticated`), so at first paint the
+// store is empty and EVERY user decodes as `scope: null`.
+//
+// It matters HERE and not only in principle because PatchesPage renders this
+// modal UNCONDITIONALLY — `<PatchApprovalModal open={modalOpen} patch={...} />`
+// sits in the page body, not behind `{modalOpen && ...}`, and PatchesPage has no
+// early return in front of it. So the modal's first render happens during the
+// page's cold load, with `patch={null}`, long before anyone opens it. Hooks run
+// on that render, which is why the old
+// `useMemo(() => getJwtClaims().scope === 'organization', [])` captured the
+// empty-store answer and pinned `isOrgScope=false` for the life of the PAGE.
+//
+// Every case below therefore mounts the modal CLOSED first and opens it later,
+// which is the sequence the app actually produces, and mocks neither `authScope`
+// nor the auth store.
+
+vi.mock('../../stores/auth', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../../stores/auth')>()),
+  fetchWithAuth: vi.fn(),
+}));
+
+import PatchApprovalModal from './PatchApprovalModal';
+import { useAuthStore } from '../../stores/auth';
+
+const PATCH = {
+  id: 'patch-1',
+  title: 'Security Update',
+  severity: 'critical' as const,
+  source: 'Microsoft',
+  os: 'Windows',
+  releaseDate: '2026-04-01T00:00:00.000Z',
+  approvalStatus: 'pending' as const,
+};
+
+function makeToken(payload: Record<string, unknown>): string {
+  const header = btoa(JSON.stringify({ alg: 'HS256', typ: 'JWT' })).replace(/=/g, '');
+  const body = btoa(JSON.stringify(payload)).replace(/=/g, '');
+  return `${header}.${body}.sig`;
+}
+
+/** Simulate /auth/refresh returning an access token some time after first paint. */
+function tokenArrives(payload: Record<string, unknown>) {
+  act(() => {
+    useAuthStore.setState({ tokens: { accessToken: makeToken(payload), expiresInSeconds: 900 } });
+  });
+}
+
+const PARTNER_LEVEL = /patch approvals are managed at the partner level/i;
+
+describe('PatchApprovalModal vs. late-arriving JWT scope (#4013)', () => {
+  beforeEach(() => {
+    useAuthStore.setState({ tokens: null });
+  });
+
+  afterEach(() => {
+    useAuthStore.setState({ tokens: null });
+  });
+
+  it('org scope arriving after the page mounted the closed modal: explains the denial and disables Approve', async () => {
+    // The exact sequence a cold load of /patches produces for an org-scoped
+    // user. THE REGRESSION: with the frozen memo the scope was captured here,
+    // from an empty store, and stayed `not organization` for the whole page —
+    // so this user was handed an enabled Approve button and no explanation, and
+    // only found out it was refused after clicking through the confirm dialog.
+    const { rerender } = render(<PatchApprovalModal open={false} patch={null} onClose={() => {}} />);
+
+    tokenArrives({ scope: 'organization', orgId: 'org-1' });
+
+    rerender(<PatchApprovalModal open patch={PATCH} ringId={null} onClose={() => {}} />);
+
+    expect(await screen.findByText(PARTNER_LEVEL)).toBeInTheDocument();
+    const submit = screen.getByTestId('patch-approval-submit');
+    expect(submit).toBeDisabled();
+    expect(submit).toHaveAttribute('title', 'Patch approvals are managed at the partner level');
+  });
+
+  it('partner scope arriving after the closed mount: Approve becomes available, with no false denial', async () => {
+    const { rerender } = render(<PatchApprovalModal open={false} patch={null} onClose={() => {}} />);
+
+    tokenArrives({ scope: 'partner', partnerId: 'partner-1' });
+
+    rerender(<PatchApprovalModal open patch={PATCH} ringId={null} onClose={() => {}} />);
+
+    await waitFor(() => expect(screen.getByTestId('patch-approval-submit')).not.toBeDisabled());
+    expect(screen.queryByText(PARTNER_LEVEL)).toBeNull();
+    expect(screen.getByTestId('patch-approval-submit')).not.toHaveAttribute('title');
+  });
+
+  it('unresolved scope: Approve is withheld, but NOT as a partner-level denial', async () => {
+    // The third state has to render as its own thing. Disabling submit is the
+    // fail-closed half (an unknown scope is not a grant); withholding the
+    // "managed at the partner level" banner is the honest half — we have not
+    // been told no, we have not been told anything.
+    render(<PatchApprovalModal open patch={PATCH} ringId={null} onClose={() => {}} />);
+
+    const submit = await screen.findByTestId('patch-approval-submit');
+    expect(submit).toBeDisabled();
+    expect(submit).toHaveAttribute('title', 'Checking your access...');
+    expect(screen.queryByText(PARTNER_LEVEL)).toBeNull();
+
+    tokenArrives({ scope: 'partner', partnerId: 'partner-1' });
+
+    await waitFor(() => expect(screen.getByTestId('patch-approval-submit')).not.toBeDisabled());
+  });
+
+  it('a present-but-undecodable token resolves without claiming a partner-level denial', async () => {
+    render(<PatchApprovalModal open patch={PATCH} ringId={null} onClose={() => {}} />);
+    expect(await screen.findByTestId('patch-approval-submit')).toBeDisabled();
+
+    act(() => {
+      useAuthStore.setState({ tokens: { accessToken: 'not-a-jwt', expiresInSeconds: 900 } });
+    });
+
+    // Resolved, and NOT org-scoped — so the gate matches what `handleSubmit` and
+    // the server enforce, both of which key on `scope === 'organization'`. This
+    // is deliberately unchanged by the fix: telling someone whose token cannot
+    // be decoded that "approvals are managed at the partner level" would be a
+    // wrong explanation for what is really a 401 on the next request.
+    await waitFor(() => expect(screen.getByTestId('patch-approval-submit')).not.toBeDisabled());
+    expect(screen.queryByText(PARTNER_LEVEL)).toBeNull();
+  });
+});
--- a/apps/web/src/components/patches/PatchApprovalModal.coldLoad.test.tsx
+++ b/apps/web/src/components/patches/PatchApprovalModal.coldLoad.test.tsx
@@ -76,6 +76,7 @@ describe('PatchApprovalModal vs. late-arriving JWT scope (#4013)', () => {
     rerender(<PatchApprovalModal open patch={PATCH} ringId={null} onClose={() => {}} />);
 
     expect(await screen.findByText(PARTNER_LEVEL)).toBeInTheDocument();
+    expect(screen.getByTestId('patch-approval-scope-notice')).toHaveTextContent(PARTNER_LEVEL);
     const submit = screen.getByTestId('patch-approval-submit');
     expect(submit).toBeDisabled();
     expect(submit).toHaveAttribute('title', 'Patch approvals are managed at the partner level');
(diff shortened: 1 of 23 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — five reviewers: `code-reviewer`, `pr-test-analyzer`, `comment-analyzer`, `type-design-analyzer`, `silent-failure-hunter`.

**Findings:** 12 raised (0 Critical) → 8 addressed in a2bebb7a5, 4 declined with reasons; 0 outstanding.

Addressed:

- **`silent-failure-hunter`** — the new `'unresolved'` state explained itself only through the submit button's `title`. A disabled button is out of the tab order and has no hover on touch, so keyboard and touch users got a dead Approve button with no on-screen reason — strictly weaker than the treatment `'denied'` already had. The existing warning banner now covers both non-`'allowed'` states with the right copy for each.
- **`silent-failure-hunter`** — established that `AuthOverlay` masks the whole window while the access token is absent (`shouldHide` requires `tokens?.accessToken`), so the pending placeholder is rarely what a user sees on an ordinary cold load. **This does not touch the bug** — the frozen memo's wrong answer outlived the overlay, which is exactly when the user saw missing tabs — but the comments and the PR body said "staring at a blank body", which overstated it. Both corrected. The placeholder is kept, and the comment now records why: it is the only DOM difference between `'unresolved'` and `'denied'`, so without it nothing would notice a future collapse back to a boolean.
- **`comment-analyzer` + `type-design-analyzer`** (raised independently, which is why it was fixed rather than reworded) — the `PARTNER_ONLY_TABS` comment claimed the constant kept the tab list, the panel gates and the placeholder from drifting, but only the placeholder read it. Made the claim true: the tab list is now built from the constant, and the comment states plainly that the three panel bodies remain individual and share the one `canManageInbound` gate.
- **`comment-analyzer`** — two overlapping comments disagreed on whether one or three surfaces were gated. Folded.
- **`type-design-analyzer`** — `approvalScope: 'unresolved'|'orgScoped'|'eligible'` broke parity with the `ringAccess`/`inboundAccess` precedent and collided conceptually with `JwtClaims.scope`, a different 4-valued domain. Renamed to `approvalAccess: 'unresolved'|'allowed'|'denied'`, dropping the intermediate `isOrgScope` boolean. Predicate polarity is a property of the computation, not of what the states are called.
- **`type-design-analyzer`** — added the cross-reference it asked for: `ringAccess` treats a resolved-but-undecodable token the *opposite* way, deliberately, and the comment now says why so the two are not "harmonized" by mistake.
- **`pr-test-analyzer`** — added the `resolved → unresolved` transition to both suites and a `hashchange`-during-the-unresolved-window case: three properties the comments asserted but nothing pinned.

Declined:

- **`silent-failure-hunter`** — a settled `'denied'` on a partner-only sub-tab renders a blank body (an org user who hand-crafts `#tab=inbound`). Real, and **pre-existing** — not what #4013 broke, and it needs its own copy decision. Recorded in a code comment; worth a follow-up issue if you want it.
- **`pr-test-analyzer`** — `waitFor` → direct `expect` style nit (the reviewer verified none of them mask anything), and a remount case it rated low value itself.

`code-reviewer` returned **0 findings** across all six things it was asked to check, including that no ordering — hydration, remount, `hashchange`, logout, token arriving or going away — can turn `'unresolved'` into a grant at either site.

**Revert probe (re-run after the fixes):** restoring the two `useMemo(..., [])` lines fails **14 of 23** cases. The 9 that pass either way are the 7 pre-existing #4010 cases plus the two partner-scope-arrives cases, which exist to prove the fix does not over-correct into blocking legitimate partner users.

**Tests:** full `apps/web` suite — **640 files, 6506 tests passed**. `astro check` → 0 errors (1684 files). `eslint` on every changed file → clean. Locale parity green with the new `checkingAccess` key in all 8 locales.

**Rebased onto current `main`.** The full-suite run initially showed 2 failures in `DevicesPage.test.tsx` — a file this PR never touches. They reproduce with this PR's changes reverted, so they were inherited from a stale base: the branch was cut at `9ff8b0f1a`, and `1fd01dbb9` ("update #4009's confirm-dialog assertions to #3994's renamed copy — main is red", #4023) landed after it. Rebased onto `1fd01dbb9` and the whole suite is green. Flagging the mechanism rather than just the outcome, since a stale base is the failure mode `CLAUDE.md` calls out as having bitten this repo repeatedly — here it was the harmless direction (my base missing a fix), not the dangerous one.

**Reachability of the second site (the issue graded it "probably not reachable"):** it is reachable. `PatchesPage.tsx:902` renders the modal unconditionally, so it mounts during the page's cold load and the memo froze then — not when the user opened it. Demonstrated end-to-end, not argued: the new `PatchesPage.scopeResolution.test.tsx` case drives the real page with the real auth store and fails against the pre-fix code.

**CI on `e182c5a5e`:** **Test Web pass**, **Test API pass**, **Type Check pass**; 54 of 55 checks green, 0 red, 1 skipped (CodeQL). Integration Tests (all 4 shards) green.

**Status:** review-clean, CI green, awaiting maintainer merge. Not merging and not closing #4013 — both are yours.

PR #4031, opened 2026-08-25
fix(web): stop the grid DeviceCard offering agent actions for network assets (#4014)

### LINKED ISSUE #4014: [Web] Grid-view DeviceCard kebab offers agent actions for network-discovered assets — no deviceClass guard
## Description

`DeviceCard.tsx` (grid view) has **no `deviceClass` handling anywhere in the file**. Its kebab unconditionally offers Terminal, Run Script, Reboot, Settings, and Decommission — including for `deviceClass === 'network'` rows, which are discovered assets with no agent.

The list view already handles this. `DeviceList.tsx:2308-2321` replaces the entire actions cell with a single "View" button for network rows:

```tsx
{(device.deviceClass ?? "agent") === "network" ? (
  // Network devices have no agent — none of the remote
  // actions (desktop/terminal/scripts/reboot) apply.
  // Phase 1 routes to the existing Discovery view.
  <button data-testid={`device-${device.id}-open-network`} …>{t("deviceList.view")}</button>
) : ( … full action set … )}
```

The grid has no equivalent branch, so the same device gets a "View" button in list view and a full agent action menu one toggle away in grid view.

## Why it matters

A network row's `id` is a **`discovered_assets.id`, not a `devices.id`** — the exact distinction that #1322 was filed for on the bulk path. Firing an agent action from the grid card therefore sends a foreign id at a `/devices/:id` endpoint. Decommission is the worst case: `decommissionDevice(device.id)` with a discovered_assets id, from a kebab that looks identical to the agent one.

## Reachability

Any time the network arm is on and the class segment filter is `All`:

- `DevicesPage.tsx:301-309` — `gridDevices` derives from `classFilteredDevices`, which includes network rows under the `all` filter.
- `DevicesPage.tsx:1364-1371` — every one of those is rendered as a `<DeviceCard>` with the full `onAction={handleDeviceAction}` wiring.

## Scope

This predates #4009 and is **not** decommission-specific — reboot, terminal and run-script are exposed the same way. The natural fix mirrors the list: a `deviceClass === 'network'` branch in `DeviceCard`'s kebab that collapses to View/open-in-Discovery, plus a test pinning it in `DeviceCard.test.tsx`.

## Related

- #1322 — the same id-class hazard, fixed on the bulk-action path only.
- #4009 / PR #4011 — added the decommission confirm dialog to both kebabs. That does **not** fix this; it only means the bad call now asks first. The PR's grid-card test is explicitly scoped in a comment so it is not misread as vetting the grid kebab.

Found by: PR #4011 review (pr-test-analyzer), 2026-08-25.

ToddHebebrand:
## Why

`DeviceCard.tsx` (grid view) had **no `deviceClass` handling anywhere in the file**. Its kebab unconditionally offered Remote Terminal, Run Script, Reboot, Settings and Remove — including for `deviceClass === 'network'` rows, which are discovered assets with no agent. A network row's `id` is a **`discovered_assets.id`, not a `devices.id`** (#1322), so every one of those actions posts a foreign id at a `/devices/:id` endpoint.

The list view has handled this since #1322: `DeviceList.tsx` replaces the entire actions cell with a single **View** button for network rows. The grid had no equivalent branch, so the same printer showed a View button in list view and a full agent action menu one toggle away in grid view. `DevicesPage.tsx` routes network rows into `<DeviceCard>` with full `onAction` wiring whenever the class segment filter is **All**, so this is reachable in normal use.

## What the API actually does — verified per path, not assumed

The issue called decommission "the worst case". It is, but it is a dead end, not a destruction. Every affected path was traced to its handler:

| Kebab action | Endpoint | Result with a `discovered_assets.id` |
|---|---|---|
| Reboot | `POST /devices/:id/commands` | **404** `Device not found` — no `device_commands` row inserted |
| Run Script | `POST /scripts/:id/execute` | **400** `No valid devices found` — validated before enqueue, nothing queued |
| Remote Terminal | client nav → `GET /devices/:id` | **404** — dead-end "device not found" page, session component never mounts |
| Remove (decommission) | `DELETE /devices/:id` | **404** — no row touched |
| Settings | modal open is client-only; Save → `PATCH /devices/:id` | **404** on save |
| *(card mount)* metrics | `GET /devices/:id/metrics?range=1h` | **404** on every network card mount, swallowed into `historyState='error'` |

**Could a `discovered_assets.id` collide with a `devices.id` and hit an unrelated real device? No.** Both primary keys are independent `uuid('id').primaryKey().defaultRandom()` (`db/schema/devices.ts:14`, `db/schema/discovery.ts:129`), neither `insert(devices)` call site supplies an explicit `id`, and there is no promotion/conversion path that copies an asset id into a device row — the only cross-reference is the separate nullable `discoveredAssets.linkedDeviceId` FK. Every handler re-resolves the id against `devices` before any side effect, so all six paths fail closed.

**So this is a UX/noise bug, not a tenant-safety or data-loss one.** Worth stating plainly rather than leaving implied.

## What changed

**1. `DeviceCard` mirrors the list rather than inventing a new treatment.** The kebab collapses to the same **View** button `DeviceList` renders, reusing its `deviceList.view` copy and `device-<id>-open-network` test id, wired to `onClick` (which `DevicesPage.handleSelectDevice` already routes to `/devices/network/:id`). **No new strings, so no locale parity work** — `deviceList.view` and `deviceList.text` already exist in all 8 locales.

**2. Same file, same root cause:** the card no longer fires `GET /devices/:id/metrics` with a discovered-asset id on mount, and renders CPU/RAM as `—` rather than the placeholder `0%` `DevicesPage` fills in for network rows. That is the dash the list already shows for agent-only columns (`DeviceList.tsx` `agentCell`).

**3. A backstop in `handleDeviceAction`.** `handleBulkAction` has filtered network rows since #1322; the single-device funnel — shared by the list kebab, the grid card and `DeviceSettingsModal` — had no such guard. Enforcing the invariant once there (reusing the existing `devicesPage.toasts.agentOnlyAction` copy) means a future surface that wires `onAction` cannot silently re-open the hole. The UI branch is the gate; this is the backstop.

## Tests

New `DeviceCard.networkClass.test.tsx` drives the **real** component. Worth noting: `DeviceCard` is *stubbed* in `DevicesPage.test.tsx`, which is exactly why no test there could ever have caught this — #4011's grid-card test was scoped in a comment for that reason.

**Revert probe** (implementation reverted to the base commit, tests kept): **8 of 11** new `DeviceCard` tests fail, and **5 of 6** of the new `DevicesPage` guard tests fail. The 4 that pass are the deliberate regression guards (agent card keeps its kebab, is not badged, still fetches metrics; a row with no `deviceClass` still defaults to agent) and are meant to hold in both states. One assertion was deliberately reshaped after the first probe — asserting the *count and identity* of controls rather than the absence of menu items, because the menu is closed by default and "no agent buttons in the DOM" passed vacuously against the unfixed code.

Six tests in `DevicesPage.test.tsx` pin the handler backstop from both surfaces and across **both** the confirm-gated actions (reboot, decommission) and the ungated ones (run-script, settings) — a guard that only held on the confirm-gated path would otherwise have looked green. One asserts an agent row still runs its action, so the guard is not a blanket block.

**Verification:** `apps/web` suite green — devices 58 files / 702 tests, and every other `src/**` directory run in batches; `tsc --noEmit` clean; `eslint` clean on all changed files.

## Review round 1

`/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter. **6 findings, 5 acted on, 2 declined.** The follow-up commit carries the detail; the substantive ones:

- **The card did not explain itself** (silent-failure-hunter, MEDIUM). With the kebab and metrics gone, a printer rendered the generic monitor glyph, a blank OS line and two dashes — indistinguishable from a broken agent. The list answers that with a Class badge; it is now ported, same copy/testid/palette, plus the list's Network glyph. Still no new strings.
- **A comment said "Decommission"** (comment-analyzer, CRITICAL) — #3994 renamed that label to "Remove" before this branch existed.
- **A comment asserted a snapshot of today's UI** rather than the invariant. Reframed: this handler must refuse network rows on its own, because nothing guarantees every future caller hides the actions first — which is precisely how #4014 happened.
- **The guard was proven only for confirm-gated actions** (pr-test-analyzer, 6/10) — widened, see above.
- **The metrics effect's new `isNetwork` dependency was unpinned** (3/10) — now tested.

Declined: reusing the bulk `agentOnlyAction` copy for a single device (plural phrasing, but the alternative is a new string in 8 locales for no gain), and a cosmetic trailing-space difference from the list.

**Also caught by review:** this branch originally carried its own fix for the same `main`-red that #4023 fixed 25 minutes after this branch's base was cut. That duplicate commit has been dropped and the branch rebased onto current `origin/main`, so only the #4014 work remains.

Closes #4014

### DIFF
--- a/apps/web/src/components/devices/DeviceCard.networkClass.test.tsx
+++ b/apps/web/src/components/devices/DeviceCard.networkClass.test.tsx
@@ -1,0 +1,163 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import DeviceCard from './DeviceCard';
+import type { Device } from './DeviceList';
+import { fetchWithAuth } from '../../stores/auth';
+
+vi.mock('../../stores/auth', () => ({
+  fetchWithAuth: vi.fn()
+}));
+
+const fetchWithAuthMock = vi.mocked(fetchWithAuth);
+
+const makeJsonResponse = (payload: unknown): Response =>
+  ({
+    ok: true,
+    status: 200,
+    statusText: 'OK',
+    json: vi.fn().mockResolvedValue(payload)
+  }) as unknown as Response;
+
+const agentDevice: Device = {
+  id: '11111111-1111-1111-1111-111111111111',
+  deviceClass: 'agent',
+  hostname: 'edge-01',
+  os: 'windows',
+  osVersion: '11',
+  status: 'online',
+  cpuPercent: 58,
+  ramPercent: 71,
+  lastSeen: '2026-02-09T10:00:00.000Z',
+  orgId: 'org-1',
+  orgName: 'Org One',
+  siteId: 'site-1',
+  siteName: 'HQ',
+  agentVersion: '1.0.0',
+  tags: []
+};
+
+// Shaped exactly like DevicesPage's `transformedNetworkDevices` mapping: the
+// `id` is a `discovered_assets.id`, os/agentVersion are blank, and cpu/ram are
+// filled with a placeholder 0 that must never be presented as a reading.
+const networkPrinter: Device = {
+  id: '22222222-2222-2222-2222-222222222222',
+  deviceClass: 'network',
+  assetType: 'printer',
+  hostname: 'Lobby Printer',
+  os: '' as Device['os'],
+  osVersion: '',
+  status: 'online',
+  cpuPercent: 0,
+  ramPercent: 0,
+  lastSeen: '2026-02-09T10:00:00.000Z',
+  orgId: 'org-1',
+  orgName: 'Org One',
+  siteId: 'site-1',
+  siteName: 'HQ',
+  agentVersion: '',
+  tags: [],
+  manufacturer: 'HP',
+  model: 'LaserJet',
+  monitoringEnabled: true
+};
+
+// #4014: DeviceCard had no `deviceClass` handling at all, so the grid offered
+// Terminal / Run Script / Reboot / Settings / Remove for network-discovered
+// assets. A network row's `id` is a `discovered_assets.id`, NOT a `devices.id`
+// (#1322) — every one of those actions posts a foreign id at a /devices/:id
+// endpoint. The list row already collapses to a single "View" button
+// (DeviceList.tsx), and the grid card must match it.
+describe('DeviceCard network-class guard (#4014)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    fetchWithAuthMock.mockResolvedValue(makeJsonResponse({ metrics: [] }));
+  });
+
+  it('renders a View button instead of the actions kebab for a network row', () => {
+    render(<DeviceCard device={networkPrinter} />);
+
+    expect(screen.getByTestId(`device-${networkPrinter.id}-open-network`)).toBeInTheDocument();
+    // The kebab trigger itself must be gone — not merely emptied — so there is
+    // no path to the agent menu at all.
+    expect(screen.queryByTestId(`device-${networkPrinter.id}-actions-menu`)).toBeNull();
+    expect(screen.queryByLabelText(`Actions for ${networkPrinter.hostname}`)).toBeNull();
+  });
+
+  it('reuses the list row\'s "View" copy (deviceList.view), not a new string', () => {
+    render(<DeviceCard device={networkPrinter} />);
+
+    expect(screen.getByTestId(`device-${networkPrinter.id}-open-network`).textContent?.trim()).toBe('View');
+  });
+
+  it('routes the View button to onClick (the network detail page) with the device', () => {
+    const onClick = vi.fn();
+    render(<DeviceCard device={networkPrinter} onClick={onClick} />);
+
+    fireEvent.click(screen.getByTestId(`device-${networkPrinter.id}-open-network`));
+
+    expect(onClick).toHaveBeenCalledTimes(1);
+    expect(onClick).toHaveBeenCalledWith(networkPrinter);
+  });
+
+  it('exposes exactly one control on a network card, and it is View — never an agent action', () => {
+    const onAction = vi.fn();
+    render(<DeviceCard device={networkPrinter} onAction={onAction} />);
+
+    // Asserting the *count and identity* of controls (rather than the absence
+    // of the menu items) keeps this non-vacuous: before the fix there is also
+    // exactly one button, but it is the kebab trigger, not View.
+    const controls = screen.getAllByRole('button');
+    expect(controls).toHaveLength(1);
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter (4 lenses, one round).

**Findings:** 6 raised → 5 addressed in `df0350ece`, 2 declined with reasons, 0 outstanding.
- *silent-failure-hunter, MEDIUM* — the card no longer explained itself once the kebab and metrics were gone (generic monitor glyph, blank OS line, two dashes = looks like a broken agent). Ported the list's Class badge + Network glyph; same copy, testid and palette, still no new locale strings.
- *comment-analyzer, CRITICAL* — a comment said the kebab offered "Decommission"; #3994 renamed that label to "Remove" before this branch existed.
- *comment-analyzer, suggestion* — the handler-guard comment asserted a snapshot of today's UI. Reframed as the invariant, since "the handler trusted its callers" is exactly how #4014 happened.
- *pr-test-analyzer, 6/10* — the guard was proven only for confirm-gated actions (reboot, decommission); a guard that only held on that path would have looked green. Widened the stub and pinned `run-script` and `settings` too.
- *pr-test-analyzer, 3/10* — the metrics effect's new `isNetwork` dependency was unpinned; now tested.
- **Declined:** reusing the bulk `agentOnlyAction` copy for a single device (plural phrasing, but the alternative is a new string across 8 locales for no gain), and a cosmetic trailing-space difference from the list.

**Also caught by review:** this branch originally carried its own fix for the same `main`-red that #4023 fixed 25 minutes after this branch's base was cut. That duplicate commit is dropped and the branch is rebased onto current `origin/main` — only the #4014 work remains.

**Tests:** `apps/web` devices suite green (`vitest run src/components/devices`, 58 files / 707 passed); every other `src/**` directory run in batches, all green; `tsc --noEmit` clean; `eslint` clean on all changed files.

**Revert probe** (implementation reverted to the base commit, tests kept): **8 of 11** new `DeviceCard` tests fail and **5 of 6** new `DevicesPage` guard tests fail. The 4 that pass are the intentional regression guards (agent card keeps its kebab, is not badged, still fetches metrics; a row with no `deviceClass` defaults to agent).

**CI:** 54/54 required checks pass, CodeQL skipped. Notably **Test Web** (green, 10m21s) is the job that would have caught this.

**One thing worth your judgement:** the API turned out to fail closed on every affected path (404, or 400 for run-script) because both primary keys are independent `uuid().defaultRandom()` and nothing promotes an asset id into a device row — so this was a dead-end UX bug, not the destructive one the issue's decommission framing implied. If you'd rather the `handleDeviceAction` backstop were its own PR, say so and I'll split it.

**Status:** review-clean, CI green, awaiting maintainer merge. Not merged, issue not closed.

Issue #4008, opened 2026-08-25
[Web] Paginated lists keep a stale currentPage when rows shrink — empty page, wrong empty-state copy, and no pager to recover


bdunncompany:
## Description

Both list components keep `currentPage` in local state and never reconcile it when the row count shrinks underneath them. Deleting the only row on the last page leaves the user on a page that no longer exists: an empty table, the wrong empty-state copy, and **no pagination control to get back**.

## Reproduction

Confirmed by rendering the components directly (11 rows, `pageSize={10}`, advance to page 2, rerender with 10 rows — which is what the parent does after a delete + refetch):

```
NotificationChannelList  onPage2=true  adjustSearch=true  anyRow=false  buttons=0
ConfigPolicyList         onPage2=true  adjustSearch=true  anyRow=false  buttons=0
```

`buttons=0` is the part that makes it a dead end. `totalPages` drops to 1, so the pager stops rendering entirely, and with no rows there are no row actions either. The only ways out are typing in the search box, changing a filter, or reloading.

## Why it happens

- `apps/web/src/components/alerts/NotificationChannelList.tsx` — `currentPage` is `useState(1)` and `paginatedChannels` slices from it. Nothing clamps it to `totalPages` when `channels` shrinks.
- `apps/web/src/components/configurationPolicies/ConfigPolicyList.tsx` — identical state machine, identical result.

Changing the search or the status/type filter *does* reset the page, so only the "rows disappeared underneath me" path is affected.

The empty state renders the no-search-results message here, because the unfiltered list is not empty and the filters are at rest. That copy is misleading in this state, but it is a symptom — the page index is the defect.

## Suggested fix

Clamp the page when the filtered count changes, e.g. `useEffect(() => setCurrentPage(p => Math.min(p, Math.max(1, totalPages))), [totalPages])`, in both components.

## Second, smaller finding in the same markup

The pagination buttons have **no accessible name**. They contain only a `ChevronLeft` / `ChevronRight`, and lucide-react stamps `aria-hidden="true"` on an icon with no children and no `aria-*`/`role`/`title` prop — the same mechanism recorded in #3697. So the previous/next controls are absent from the accessibility tree and unreachable by name. `getByRole('button', { name: /next/i })` finds nothing; the probe above had to select them positionally.

## What I checked, and what I did not

[...]

PR #4032, opened 2026-08-25
fix(web): distinguish a zero-data list from a search that matched nothing (#3992)

bdunncompany:
Part 1 of #3992 (the empty states). **Part 2 is deliberately not here** — see the note at the bottom.

### Zero-data empty states no longer say "adjust your search"

`/configuration-policies` and `/alerts/channels` showed the no-search-results copy to a tenant with
**no rows at all that had never searched**, telling a first-time user to change a search they never
made and hiding the only useful next step.

Both components already held what was needed to tell the states apart — the unfiltered list plus the
control values — so the branch is a derived boolean, not new state:

```ts
const hasNoPoliciesAtAll =
  policies.length === 0 && query.trim().length === 0 && statusFilter === "all";
```

Keyed off the **unfiltered** list *and* the controls being at rest, so a search that genuinely
matches nothing still gets the adjust-your-search message.

The create action reuses what each page already has: policies link to `/configuration-policies/new`
with the existing `newPolicy` label, channels call the page's existing `handleCreate` through a new
optional `onCreate` prop with the existing `newChannel` label. **No new button strings** — the only
new copy is one sentence per list.

### Terminology, checked by usage rather than by my own translation

Locale parity checks that keys exist, not that the words are right, so I diffed each new sentence
against how its own catalogue actually talks. **Three of eight were wrong and every test stayed
green.**

My first correction was wrong too. I aligned the French to the sibling string (`stratégie`) on a
match-your-neighbour heuristic, which anchored on one string instead of the page: `fr-FR` uses
**politique 95 times against stratégie 16**, the page reads *Politiques de configuration* and
*Rechercher des politiques...*, and the CTA directly beneath this empty state is *Nouvelle
politique*. `fr-CA` is 97:14. The sibling is the anomaly, and `TERMINOLOGY.md` agrees — so the new
strings use **politique**. Italian genuinely is **criterio/criteri** (110:84, glossary concurs).

The copy also had to stop overpromising: it first said creating a policy starts *enforcing settings
across your devices*, but a policy is created unassigned — this catalogue's own string is *"No
assignments yet. Assign this policy to targets above."* It now says create, then assign.

Two more came out of review. German said *Warnungen an E-Mail* where its own channel catalogue uses
*Benachrichtigungen per E-Mail*; fixed. And `notificationChannelsPage.newChannel` was *Nouvelle
chaîne* while the notification copy is overwhelmingly *canal* (30:9 `fr-FR`, 31:7 `fr-CA`) — I had
intended to flag rather than migrate, but reusing that label as the empty-state CTA renders it
directly beside a sentence saying *canal*, so shipping the mismatched pair would be this change's
defect rather than a pre-existing one. Both French values are now *Nouveau canal*, which also
corrects the page header.

Still flagged and **not** changed: `noPoliciesFoundTryAdjustingYourSearch` says *stratégie* in both
French catalogues, against the dominant *politique* and the glossary. Nothing in this PR renders it
next to the new copy, so it stays a maintainer's call.

### One thing found and deliberately not fixed

Both lists keep a stale `currentPage` when the row count shrinks. Delete the only row on page 2 and
`totalPages` drops to 1, so the pager stops rendering and there is no control left to get back:

```
NotificationChannelList  onPage2=true  adjustSearch=true  anyRow=false  buttons=0
ConfigPolicyList         onPage2=true  adjustSearch=true  anyRow=false  buttons=0
```

That state also shows the no-search-results copy, but the page index is the defect and it pre-dates
this change. Filed as #4008, along with a second finding from the same markup: the pager buttons
have no accessible name.

There is a second state the predicate cannot see: a malformed HTTP 200 is coerced to `[]` upstream
(`asList.ts` for channels, the parent loader for policies), so "no rows" is not proof that none
exist. Distinguishing that needs the loaders to reject an unrecognised success shape, which is
outside this change. Both predicates now carry a comment saying explicitly that they establish
"the list is empty and the filters are untouched" and **not** that the tenant has none — the copy is
a suggestion, not a claim.

### Verification

- `tsc --noEmit` clean, **6440 passed / 0 failed**, `eslint src` clean; artifacts confirmed newer
  than every changed source.

Each new test controlled — broken on purpose, observed failing, and only its own test failing:

|---|---|---|
| force `hasNoPoliciesAtAll` false | only the zero-data policies test red | `1 failed \| 1 passed` |
| force `hasNoChannelsAtAll` false | only the zero-data channels tests red | `2 failed \| 1 passed` |
| force `hasNoChannelsAtAll` **true** | only the *search* test red | `1 failed \| 2 passed` |
| blank both new sentence values | only the two sentence assertions red | `2 failed \| 3 passed` |

The last row is the one that matters: forcing the branch **on** reds only the search case, which
proves it discriminates in both directions rather than just rendering some empty message.

### Why part 2 is not in this PR

I had implemented the channel-test markup fix centrally in `scrubChannelTestError`. Your comment on
the issue redirects it to the composition point, and you are right on two things I had wrong: there
are **four** call sites (I had missed `workers/webhookDelivery.ts:167`), and the reason reaches the
user twice — the route returns `testResult` raw and the page toasts `testResult.message`, so a
scrub-time fix never touches the toast at all. Rather than ship the inferior shape, I have dropped
it here and will follow up separately. Findings that came out of building it are on the issue.


ToddHebebrand (approved):
Approving and merging. The split was the right call, and this half is clean.

**Verified rather than taken on the body's word:** the filter defaults the predicate keys off are genuinely at rest by default (`NotificationChannelList.tsx:180` `useState<string>('all')`, `ConfigPolicyList.tsx:90` `useState<string>("all")`), `apps/web/src/pages/configuration-policies/new.astro` exists so the anchor is not a dead link, and both new keys are present in all eight locales — tr-TR included, which is the one that usually gets dropped.

**Two things I check on your PRs by default, and neither fired.** `onCreate` is actually passed at `NotificationChannelsPage.tsx:538` rather than added and left unconnected. And the comments are self-limiting instead of overclaiming — stating outright that the predicate does not establish the tenant is new, naming the retained-`currentPage` case as uncovered and pre-existing, and pointing at #4008 for it. That is the right shape for a comment: what it proves, and what it deliberately does not.

**The tests earn their assertions.** Rendering a real row first, then typing `zzzznomatch`, is a genuine control — it proves the search is what emptied the list rather than asserting into an empty render. And asserting the sentence instead of only the absence of the old one closes exactly the hole your own comment names.

**On the two mask bypasses you found on #3992** — the case-sensitive host match and the numeric character reference. Both are real and both are worth their own issue; file them separately rather than folding them into part 2, since they pre-date this work and have a different fix. The normalise-before-scrub landmine you flagged lands on my side: my part-2 branch composes at the sender, which is precisely the ordering you warned about, so I am reworking it against your bracketing point.

Merging as part 1 of #3992. Part 2 stays open on my branch.

Issue #3996, opened 2026-08-25
[API] Aborting an org/partner offboarding commits the status flip before cancelling the queued uninstalls — a fleet-wide window where agents claim them as ordinary commands


ToddHebebrand:
## Description

Aborting an organization or partner offboarding un-drains the tenant **before** it cancels the `self_uninstall` commands the drain queued. Between those two commits, every agent under the tenant authenticates on the normal path — where the queued uninstall is an ordinary, claimable command.

## The window

While a tenant is offboarding, `getAgentTenantState` (`services/tenantStatus.ts:203-240`) returns `'draining'`, which narrows each agent's command claim to `['self_uninstall']` and restricts its route surface. That narrowing is keyed on `organizations.status === 'offboarding'`.

On abort (`routes/orgs.ts`), two things happen in **separate transactions**:

1. The org-status UPDATE. On the `suspendedLifecycleOverride` branch it runs in its own fresh system transaction — `runOutsideDbContext(() => withSystemDbAccessContext(runUpdate))` (`routes/orgs.ts:~1931`) — and commits on its own.
2. `abortOrganizationOffboarding(organization.id)` (`routes/orgs.ts:1958`, `:1963`, `:2016`), which cancels the outstanding `self_uninstall` rows. Per its own doc comment (`services/tenantOffboarding.ts:33-92`) it may itself fall back to a second independent system transaction when the ambient context cannot see the tenant row.

So the sequence at runtime is: **status flips to `active` and commits** → tenant is no longer draining → agents re-authenticate on the ordinary path, where `claimPendingCommandsForDevice` is called with **no** type allowlist → the still-`pending` `self_uninstall` rows are claimable as ordinary commands → *then* the cancel commits.

Any agent heartbeat landing in that window collects the uninstall and removes itself from the machine. Agents beat on a ~60s cycle, so for a tenant of any size the window is very likely to be hit by at least one device, and the blast radius is the tenant's whole fleet rather than a single endpoint.

## Why this is being filed now

This is the **same shape** as a device-level bug just fixed in #3986 (PR pending): `POST /devices/:id/restore` flipped `devices.status` to `'offline'` and released the queued uninstall in a separate round-trip. `isDeviceUninstallDraining` requires `status='decommissioned'`, so the flip ended the drain while the row was still `pending`, and `'offline'` is an ordinary status — the agent re-authenticated normally and could claim the uninstall as an ordinary command. Fixed by releasing **before** the flip and running both writes in one transaction.

[...]

Issue #2877, opened 2026-07-28
Offboarding entry (#2781) self-deadlocks: request-txn org-row lock vs nested system-context UPDATE on the same row


ToddHebebrand:
Found + code-verified in the 2026-07-27 pre-release QA sweep (live stack + real Windows agent). **Release blocker for the offboarding drain feature.**

## Symptom
`PATCH /organizations/:id {status:'offboarding'}` wedges the request indefinitely (idle-in-transaction; no lock/idle timeout catches it — it's an application-level cycle Postgres can't detect). On manual backend termination the route 500s and the status update **rolls back while the side effects commit** (offboardingStartedAt set, suspension cleared, self_uninstall queued) — a torn entry.

## Root cause
In `routes/orgs.ts` the PATCH handler runs, inside the request DB context/transaction:
```ts
const [organization] = await db.update(organizations).set(updates).where(conditions).returning();  // row-locks the org row
...
if (data.status === 'offboarding') {
  await beginOrganizationOffboarding(organization.id, ...);   // orgs.ts:~1488
}
```
`beginOrganizationOffboarding` (`services/tenantOffboarding.ts:155`) does `runOutsideDbContext(() => withSystemDbAccessContext(async () => { await db.update(organizations).set({offboardingStartedAt...}).where(...) }))` — a **second UPDATE on the same org row from a fresh pooled connection**. The request transaction still holds the row lock (runOutsideDbContext detaches the ALS context but does NOT close/commit the outer transaction — see the known runOutsideDbContext-does-not-close-outer-txn / #1105 txn-pool-poison family), so the system-context UPDATE blocks on the lock, and the request can't commit to release it. Deadlock.

## Fix direction
Do the offboarding side effects AFTER the request transaction commits (move `beginOrganizationOffboarding` out of the row-locking path), or perform the whole status transition + side effects in ONE context instead of nesting a fresh system connection that re-locks the row. Whatever the fix, it must be atomic: the torn-entry outcome (status rolled back but self_uninstall queued) must not be reachable.

Same structural class as the #1105 request-long connection-hold incidents.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ToddHebebrand:
Independently reproduced by a second QA agent (confirmed via `pg_stat_activity`: the request backend sits `idle in transaction` holding the org row lock while a second pooled connection blocks on the same row). Two sharper consequences worth capturing:

[...]

PR #4034, opened 2026-08-26
fix(api): recheck the tenant precondition under the repair's own lock (#4022)

### LINKED ISSUE #4022: [API] repairIncompleteEntry never rechecks status under its lock — the offboarding sweep can queue self_uninstall against a tenant that was already aborted back to active
**`repairIncompleteEntry` locks the tenant row by id and never rechecks its status, so it can queue fresh `self_uninstall` commands against a tenant that was aborted out of offboarding while the sweep was running.** This does not merely fail to backstop #3996 — it manufactures the state #3996 describes.

Traced by @bdunncompany on #3996; every link independently verified against `origin/main` below.

**The interleaving.**

1. `sweepOffboardingTenants` (`services/tenantOffboarding.ts:713`) selects candidates with `eq(organizations.status, 'offboarding')` inside one `runOutsideDbContext(() => withSystemDbAccessContext(...))` block, capturing `{ id, startedAt }`.
2. Each tenant is then processed in a **separate** `runOutsideDbContext(() => withSystemDbAccessContext(...))` block (`:737-742`), and the repair branch keys on the **snapshotted** value — `if (!org.startedAt) { await repairIncompleteEntry(...) }`. Nothing re-reads the status.
3. Meanwhile an operator aborts the offboarding. With a null stamp, `abortOrganizationOffboarding` correctly does nothing: its UPDATE guards on `isNotNull(organizations.offboardingStartedAt)` (`:341`), matches no row, and returns `{ aborted: false, uninstallsCancelled: 0 }` at `:345` **before reaching the cancel**. The org status flips to `active`/`trial`/`suspended` and commits.
4. `repairIncompleteEntry` (`:636`) then takes its lock — `select({id}).from(organizations).where(eq(organizations.id, scopeId)).for('update')` (`:657-662`). **`eq(id)` only. No status predicate, and none afterwards.**
5. It proceeds unconditionally to `prepareAgentDrainForOrgIds(orgIds)` and `queueDrainUninstalls(orgIds, null)` (`:669-670`), then writes the stamp guarded only on `isNull(offboardingStartedAt)` (`:678-685`) — which still matches, because the stamp is null.

**Result: an `active` or `trial` tenant with a live `offboardingStartedAt` and freshly queued `self_uninstall` rows.** Because the tenant is no longer `offboarding`, `getAgentTenantState` does not return `'draining'`, so `claimPendingCommandsForDevice` runs with **no type allowlist** and the uninstalls are claimable as ordinary commands on the next heartbeat. For `suspended` it produces the suspended-plus-live-stamp-plus-queued-uninstalls combination that the `suspendedLifecycleOverride` branch exists to handle — a state previously assumed to arise only from an already-torn tenant.

**Why this is worse than #3996.** #3996 is a window between two commits: real, but narrow and timing-dependent. This one is a durable wrong state written by a scheduled job, against a tenant an operator has explicitly rescued. The audit row it writes says `organization.offboarding_entry_repaired` with a success result and a device count — so the trail records it as a repair, not as a fleet-wide uninstall queued against a live customer.

**Fix.** Recheck under the lock. The `.for('update')` select should return the status and stamp, and the repair must abandon unless `status = 'offboarding' AND offboarding_started_at IS NULL` still holds at that point. The lock is already acquired in the right order (`:646-656` documents the #2877 deadlock rationale) — it just isn't being used to validate the precondition it was taken for.

Worth applying the same reasoning to the partner branch, which has the identical shape.

**Test.** The deterministic pin is an integration test that snapshots a candidate, aborts the tenant, then runs the repair and asserts no `self_uninstall` rows exist and no stamp was written. @bdunncompany offered to write this and it is the right next step — it is more reproducible than #3996's delivery race.

**Not verified:** no concurrency test was run for this; it is a static trace, though every cited line was confirmed. The partner-side paths (`abortPartnerOffboarding`, system-only partner PATCH/DELETE) and `DELETE /organizations/:id` were not traced to the same depth.

Related: #3996 (the abort-ordering window this can create), #3995 (agent-side pre-teardown fence — the residual once the API-side ordering is fixed).

bdunncompany:
Closes #4022. Traced on #3996, filed by you, and this is the fix plus the deterministic test you asked for.

### The defect

`repairIncompleteEntry` took its row lock on `eq(id)` alone and never rechecked what it had locked. The sweep selects candidates in one system context and repairs each in a **separate** one, keying on the snapshotted `startedAt`, so an operator can abort the tenant in between — and that abort is a legitimate no-op when the stamp is null, because its UPDATE guards on `isNotNull(offboardingStartedAt)` and returns before cancelling anything.

The repair then arrived at a rescued tenant, queued fresh `self_uninstall` rows and stamped it. With the tenant no longer `offboarding`, `getAgentTenantState` stops returning `'draining'` and the claim runs with no type allowlist, so the fleet collects those uninstalls as ordinary commands on the next heartbeat.

### The fix

The `.for('update')` select now returns `status` and `offboardingStartedAt`, and the repair abandons unless `status = 'offboarding' AND offboarding_started_at IS NULL` still holds under the lock. Both the organization and partner branches. As you noted, the lock was already taken in the right order — it just was not being used to validate the precondition it was taken for.

Abandoning is deliberately not an error: losing that race is the correct outcome. The `warn` moved **after** the recheck, because the old message said "completing the entry" before the row was owned or the precondition confirmed, directly above the branch that now abandons. Its comment also names repair-vs-repair, not just operator-abort — another repair winning the row first leaves `offboarding` + stamped and lands in the same branch.

### The test

`offboardingRepairRecheck.integration.test.ts`, six cases against real Postgres. Each was controlled by breaking the fix and confirming **only** the intended case reds:

|---|---|---|
| baseline, fix in place | all pass | `6 passed` |
| drop the stamp conjunct (both branches) | only the org already-stamped case red | `1 failed \| 4 passed` |
| partner branch never repairs | only the partner positive control red | `1 failed \| 4 passed` |
| move the early return **below** `prepareAgentDrainForOrgIds` | only the two enrollment-key cases red | `2 failed \| 4 passed` |
| drop the stamp conjunct on the **partner branch only** | only the partner already-stamped case red | `1 failed \| 5 passed` |

The last two rows exist because review caught the first pair proving less than I claimed.

**The audit assertion did not prove drain prep was skipped.** `offboarding_entry_repaired` is written only *after* prep, queue and stamp, and `writeAuditEvent` is fire-and-forget, so its absence proves neither. Moving the early return below `prepareAgentDrainForOrgIds` left every test green while the abandoned tenant still had its enrollment keys expired and its sockets dropped. Each abandon case now seeds a key with a far-future expiry and asserts it is untouched, since prep sets `expiresAt = now` on every unexpired key for the org.

**The partner branch was unpinned.** The original mutation removed the stamp conjunct from *both* branches, so it demonstrated org sensitivity and nothing about the partner path — removing it from the partner side alone went undetected. There is now a partner stamp-negative case, and the control targets that branch specifically.

Worth recording how the mis-aimed control was caught: it red the *org*-named test while claiming to mutate the partner branch. The first `.from(partners)` in this file is at `:527`, well above `repairIncompleteEntry`, so a naive index search landed on the org predicate at `:685` instead of the partner one at `:692`. The rerun targets the line and asserts a `.from(partners)` appears above it before running. A red control proves nothing until you have confirmed the mutation landed where you meant.

An earlier version of the suite also failed all three cases at once, which looked like a successful control and was not — a setup bug of mine (`createOrganization` takes an options object; I had passed a bare id, so partner rows went into `organizations`). The positive controls are what caught it.

### Two limits, stated rather than implied

**The test pins the recheck, not the lock.** Every abort in it commits before the repair starts, so removing `.for('update')` leaves all five green. Pinning the lock needs a backend observably blocked before the abort commits — the `pg_stat_activity` barrier at `orgCurrencyCreationBarrier.integration.test.ts:102` is the pattern for it. I did not build that here because you said this needs no concurrency harness, and it is a bigger change than the guard; happy to add it if you want the lock pinned too.

**I exported `repairIncompleteEntry`** to stage the race. The sweep's own select would no longer return the tenant once the abort has landed, so the interleaving is unreachable through the public entry point. That barrier approach would also remove the need for the export, which is the other argument for it.

### Verification

Full API gate, local:

- `tsc --noEmit` clean
- unit **25263 passed / 19 skipped / 0 failed**
- `eslint src` clean
- integration **2418 passed / 3 skipped**, with 2 failures in `patchCanonicalExportParity.integration.test.ts`

Those two are **not from this change**: they fail identically with this diff stashed on clean `origin/main` (four `schema defaults` assertions, e.g. `expected { …(2) } to deeply equal { autoApproveDeferralDays: +0, …(1) }`). That looks like an existing red on main worth its own issue — say the word and I will file it with the baseline evidence.

Two existing unit fixtures in `tenantOffboarding.test.ts` needed updating: the mocked lock select is now read rather than ignored, so `{ id }` had to become `{ status, startedAt }`. Intent unchanged.

### DIFF
--- a/apps/api/src/__tests__/integration/offboardingRepairRecheck.integration.test.ts
+++ b/apps/api/src/__tests__/integration/offboardingRepairRecheck.integration.test.ts
@@ -1,0 +1,326 @@
+import './setup';
+
+import { describe, expect, it } from 'vitest';
+import { and, eq } from 'drizzle-orm';
+import { withSystemDbAccessContext } from '../../db';
+import { auditLogs, deviceCommands, devices, enrollmentKeys, organizations, partners } from '../../db/schema';
+import {
+  abortOrganizationOffboarding,
+  repairIncompleteEntry,
+} from '../../services/tenantOffboarding';
+import { createOrganization, createPartner, createSite } from './db-utils';
+import { getTestDb } from './setup';
+
+/**
+ * #4022 — the offboarding sweep selects candidates in one system context and
+ * repairs each in a separate one, keying on the SNAPSHOTTED `startedAt`. An
+ * operator can abort the tenant in between, and that abort is a legitimate
+ * no-op when the stamp is null: its UPDATE guards on
+ * `isNotNull(offboardingStartedAt)` and returns before cancelling anything.
+ *
+ * So the repair could arrive at a tenant that is back to `active`, take its
+ * lock on `eq(id)` alone, and queue fresh `self_uninstall` rows plus a live
+ * stamp. Because the tenant is no longer `offboarding`, `getAgentTenantState`
+ * stops returning `'draining'` and the claim runs with no type allowlist — the
+ * fleet collects those uninstalls as ordinary commands on the next heartbeat.
+ * The audit row records `offboarding_entry_repaired` with a success result, so
+ * nothing about it looks wrong afterwards.
+ *
+ * This drives the real service against real Postgres, staging the race by
+ * calling the repair with a snapshotted candidate — the sweep's own select
+ * would not return the tenant once the abort has committed.
+ *
+ * Scope, deliberately: this pins the RECHECK, not the lock. Every abort here
+ * commits before the repair starts, so removing `.for('update')` leaves all of
+ * these green. Holding an uncommitted `active` UPDATE open in one transaction
+ * while the repair runs in another WOULD pin the lock — see the
+ * `pg_stat_activity` barrier in `orgCurrencyCreationBarrier.integration.test.ts`
+ * — and would also remove the need to export the function.
+ */
+const runDb = it.runIf(!!process.env.DATABASE_URL);
+
+async function seedDevice(orgId: string, siteId: string, label: string): Promise<string> {
+  const [row] = await getTestDb()
+    .insert(devices)
+    .values({
+      orgId,
+      siteId,
+      agentId: `agent-${label}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`,
+      hostname: `host-${label}`,
+      osType: 'windows',
+      osVersion: '11',
+      architecture: 'x86_64',
+      agentVersion: '0.0.0-test',
+      status: 'offline',
+    })
+    .returning({ id: devices.id });
+  return row!.id;
+}
+
+/**
+ * `prepareAgentDrainForOrgIds` expires every unexpired enrollment key for the
+ * org (`expiresAt = now`). Seeding one with a far-future expiry gives the
+ * abandon path a DB-observable probe: if the early return were placed AFTER
+ * drain prep instead of before it, this key would come back stamped to ~now
+ * even though nothing was queued and no stamp was written. The audit row alone
+ * cannot cover this — it is emitted only after prep, queue and stamp, and
+ * `writeAuditEvent` is fire-and-forget, so its absence proves neither.
+ */
+const FAR_FUTURE = new Date('2099-01-01T00:00:00Z');
+
+async function seedEnrollmentKey(orgId: string, label: string): Promise<string> {
+  const [row] = await getTestDb()
+    .insert(enrollmentKeys)
+    .values({
+      orgId,
+      name: `key-${label}`,
+      key: `k-${label}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`,
+      expiresAt: FAR_FUTURE,
+    })
+    .returning({ id: enrollmentKeys.id });
+  return row!.id;
+}
+
+async function keyStillUnexpired(keyId: string): Promise<boolean> {
+  const [row] = await getTestDb()
+    .select({ expiresAt: enrollmentKeys.expiresAt })
+    .from(enrollmentKeys)
+    .where(eq(enrollmentKeys.id, keyId));
+  return row?.expiresAt?.getTime() === FAR_FUTURE.getTime();
+}
+
+async function repairAudits(orgId: string): Promise<number> {
+  const rows = await getTestDb()
+    .select({ id: auditLogs.id })
+    .from(auditLogs)
+    .where(
+      and(
+        eq(auditLogs.orgId, orgId),
+        eq(auditLogs.action, 'organization.offboarding_entry_repaired')
+      )
+    );
+  return rows.length;
+}
+
+async function pendingUninstalls(deviceId: string): Promise<number> {
+  const rows = await getTestDb()
+    .select({ id: deviceCommands.id })
+    .from(deviceCommands)
+    .where(and(eq(deviceCommands.deviceId, deviceId), eq(deviceCommands.type, 'self_uninstall')));
+  return rows.length;
+}
+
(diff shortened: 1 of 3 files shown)


ToddHebebrand (approved):
Verified against `origin/main` (`879b98890`) rather than the body. Approving — but the check that mattered was not one you listed, so recording it.

**The central claim depends on something neither the PR nor #4022 states: `repairIncompleteEntry` opens no transaction of its own.** A `SELECT ... FOR UPDATE` in autocommit releases its lock at statement end, which would have made the recheck a plain read and the title's "under the repair's own lock" false — the recheck would still close the *observed* race (abort commits before the repair starts), but nothing would hold the row across drain prep, queue and stamp.

It holds. `withSystemDbAccessContext` → `withDbAccessContext` → `baseDb.transaction(...)` at `apps/api/src/db/index.ts:465`, and `:213-217` says outright the transaction MUST stay open for the full callback because the RLS context is set with `SET LOCAL`. The sweep calls the repair inside that callback (`tenantOffboarding.ts:934-937` org, `:967-976` partner), so the lock spans the recheck through the stamp. The RLS mechanism is what makes the lock load-bearing — worth knowing, because a future move to per-statement context would silently un-fix this.

**Both supporting claims independently confirmed.** The abort's `isNotNull(offboardingStartedAt)` guard is real at `:438` (org) and `:457` (partner), so a null-stamp abort genuinely cancels nothing and leaves no trace for the repair to notice. `getAgentTenantState` is `services/tenantStatus.ts:203`.

**The six tests really run in CI.** `it.runIf(!!process.env.DATABASE_URL)` silently skips the whole file when that variable is absent, and a skipped integration file reads as a green shard. `DATABASE_URL` is set in the integration job env at `.github/workflows/ci.yml:1782`, and all four shards are green — so this is executed coverage, not a passing no-op. Your two local `patchCanonicalExportParity` failures did not reproduce on any shard; treat them as local.

**On your two stated limits — both are the right call, and the first earns a follow-up rather than a change here.** The lock is now *more* load-bearing than it was before this PR: previously it was a lock-order guard for #2877, now it is also what makes the recheck trustworthy. It is untested in exactly that new role — dropping `.for('update')` leaves all six green, as you say. That is a refactor hazard, not a defect in this diff, so I do not want the `pg_stat_activity` barrier bolted on here. Filing it separately, and the export goes away when it lands.

The self-corrections are the most useful part of the write-up. The audit-row assertion proving nothing because `writeAuditEvent` is fire-and-forget, and the partner branch being unpinned because the mutation hit `:685` instead of `:692` — both are the kind of thing that normally ships as false confidence. The far-future enrollment-key probe is the right fix for the first: it is DB-observable and it fails for the right reason. Only two of the four abandon cases carry it, which is sufficient rather than an omission — the early return is one shared block, so a third probe would pin nothing new.

Merging.

ToddHebebrand:
Follow-up for the unpinned lock is #4036 — includes the reason it matters more than it looks (the lock's lifetime is a side effect of how `withSystemDbAccessContext` establishes RLS context, so it can be revoked from a long way away), and notes that a barrier test lets `repairIncompleteEntry` go back to module-private.

PR #4041, opened 2026-08-26
fix(auth): let passwordless SSO accounts enroll a first MFA factor via IdP re-authentication (#4018)

### LINKED ISSUE #4018: [Auth][SSO] SSO sessions can never satisfy requireMfa() — 509 MFA-gated routes unreachable, and every in-product remedy is a dead end
## Description

A user who signs in via SSO gets `403 MFA_REQUIRED` on every `requireMfa()` route, and **none of the remedies the product offers can clear it**. Reported by a hosted partner on 2026-08-25 who had just set up partner-wide SSO: "Generate Link" in Add Device returns *"Multi-factor authentication is required to generate links. Set up MFA in your profile settings and sign in again, then retry."* Following that instruction cannot work.

`requireMfa()` gates on the JWT claim, not on factor enrollment (`apps/api/src/middleware/auth.ts:860`):

```ts
export function hasSatisfiedMfa(auth) {
  if (!ENABLE_2FA) return true;
  return auth.token?.mfa === true;
}
```

## The loop

**1. SSO mints `mfa: false` by default.** `apps/api/src/routes/sso.ts:2706`:

```ts
const idpMfa = provider.trustsIdpMfa === true && idpAssertedMfa(idClaims);
const ssoMfa = idpMfa && (user.mfaEnabled === true || !ssoPolicy.required);
```

`sso_providers.trusts_idp_mfa` defaults to `false` (`apps/api/src/db/schema/sso.ts:56`), so `ssoMfa` is `false` for every provider nobody has explicitly opted in.

**2. "Set up MFA in your profile settings" is impossible for an SSO-provisioned user.** JIT provisioning writes `passwordHash: null` (`apps/api/src/routes/sso.ts:2660`, *"SSO users don't have passwords"*). `POST /auth/mfa/setup` requires the current password, and `requireCurrentPasswordStepUp` returns **401 "Invalid credentials"** on a null hash (`apps/api/src/routes/auth/helpers.ts:173`). The user sees a wrong-password error for a password they never had.

**3. "…and sign in again" changes nothing.** Even for a user who *does* have a factor, `ssoMfa` is `idpMfa && (…)` — with `trustsIdpMfa` off, `idpMfa` is `false`, so enrollment is irrelevant. The claim stays `false` through refresh rotation (`apps/api/src/routes/auth/login.ts:924`).

**4. The one real remedy has no UI and is itself MFA-gated.** `trustsIdpMfa` appears in `createProviderSchema`/`updateProviderSchema` (`apps/api/src/routes/sso.ts:180`) but has **zero references in `apps/web/src`** and **zero in `apps/docs/src`** — there is no toggle and no documentation. And `PATCH /sso/providers/:id` sits behind `requireMfa()` (`apps/api/src/routes/sso.ts:1080`), so an admin whose only session is SSO cannot flip the flag that would unblock their session. Circular.

Net: a partner who adopts SSO — especially with `enforceSSO` on, which removes the password escape hatch — loses **509 `requireMfa()`-gated handlers** (software, deployments, discovery, orgs, ai, snmp, pam, tunnels, sso itself, …) with no in-product way out.

## Reproduction

1. Configure a partner SSO provider, leave `trustsIdpMfa` at its default.
2. Sign in via `GET /api/v1/sso/login/partner/:partnerId`.
3. Add Device → Generate Link → `403 MFA_REQUIRED`.
4. Follow the banner to Profile → Set up MFA → 401 "Invalid credentials" (no password on the account).

Audit trail confirms the claim: the `user.login` row for the session records `method: 'sso-partner'`, `mfa: false` (`apps/api/src/routes/sso.ts:2952`).

## Proposed fixes

1. **Surface `trustsIdpMfa`** as a toggle in `SsoProvidersPage.tsx` and document it in `apps/docs/src/content/docs/reference/sso.mdx`, with the `amr: ["mfa"]` requirement stated plainly (`idpAssertedMfa`, `apps/api/src/services/sso.ts:255`) — many IdPs need an optional-claims change to emit it.
2. **Let passwordless accounts enroll a factor.** `POST /auth/mfa/setup` needs an alternative step-up when `passwordHash IS NULL` — a recent-SSO-auth assertion or an emailed confirmation — mirroring the L4 passwordless fallback that already exists for accounts that *have* a factor (`apps/api/src/routes/auth/helpers.ts:185`). Today that fallback is chicken-and-egg.
3. **Break the circularity**: `PATCH /sso/providers/:id` behind `requireMfa()` means SSO admins can't self-serve the fix. At minimum the SSO provider write routes should accept a fresh step-up rather than a session claim.
4. **Make the copy SSO-aware.** When the session was minted by SSO, "Set up MFA in your profile settings and sign in again" is actively misleading — it should say the IdP's MFA assertion isn't trusted for this provider and point the admin at the setting.

## Affected files

- `apps/api/src/middleware/auth.ts:830,860` (gate)
- `apps/api/src/routes/sso.ts:2660,2706,1080` (claim derivation, null password, circular gate)
- `apps/api/src/routes/auth/helpers.ts:173` (401 on passwordless step-up)
- `apps/web/src/components/devices/AddDeviceModal.tsx:959,986,1064` (the misleading copy)
- `apps/web/src/components/settings/SsoProvidersPage.tsx` (missing toggle)
- `apps/docs/src/content/docs/reference/sso.mdx` (undocumented)

ToddHebebrand:
## Summary

A passwordless, SSO-provisioned user can now prove identity through a forced IdP re-authentication and use that proof — instead of a password — to enroll a first MFA factor. This also exposes `trustsIdpMfa`, which existed in the API but had no UI and no docs.

Before this, such accounts could enroll **nothing**. Every enrollment surface (`/auth/mfa/setup`, `/auth/passkeys/register/*`) requires the current password, and JIT-provisioned SSO users have `password_hash IS NULL`, so `requireCurrentPasswordStepUp` returned `401 Invalid credentials` for a password they never had. That locked them out of ~509 `requireMfa()`-gated handlers with no in-product remedy — the reported symptom was "Generate Link" in Add Device returning 403 and the UI advising a profile-settings flow that cannot work.

Closes #4018.

## How it works

```
POST /sso/reauth/start          authMiddleware only, NOT requireMfa()
  → provider resolved from the caller's OWN user_sso_identities row
  → sso_sessions row in reauth mode, snapshotting {authEpoch, mfaEpoch, sid, providerVersion}
  → IdP with prompt=login & max_age=0

GET /sso/callback  (reauth branch)
  → auth_time must exist, and be >= the transaction's own created_at
  → (provider_id, sub) must ALREADY belong to this user
  → account must still be passwordless
  → mints a single-use Redis grant, operation: 'enroll_first_factor'
  → 302 /settings/profile#ssoReauthGrant=<id>          (mints NO tokens)

/mfa/setup · /mfa/enable · /passkeys/register/*
  → resolveEnrollmentStepUp accepts that grant in place of currentPassword,
    ONLY for password_hash IS NULL AND zero existing factors
```

## Reviewer notes — please read these five

1. **New column + CHECK on `sso_sessions`** (`apps/api/migrations/2026-08-25-sso-sessions-reauth-user-id.sql`): `reauth_user_id uuid REFERENCES users(id) ON DELETE CASCADE`, plus `sso_sessions_single_mode_chk` making link and reauth mutually exclusive. **No cascade or export-policy registration is owed** — `sso_sessions` has no `org_id`, so it is not in `CORE_ORG_CASCADE_DELETE_ORDER` and needs no `CORE_TENANT_EXPORT_POLICY` entry. `tenantCascade.ts` already clears it provider-keyed on both axes; only its comment changed. Flagging this explicitly so nobody re-derives it.

2. **`enroll_first_factor` is deliberately absent from `STEP_UP_OPERATIONS`** (`routes/auth/schemas.ts`). Adding it would let anyone who can satisfy an ordinary `/auth/mfa/step-up` mint an enrollment grant with **no** IdP re-authentication. A test locks the exclusion. Note the neighbouring `satisfies readonly StepUpOperation[]` comment previously claimed the type would catch divergence — it does not, and that comment has been corrected.

3. **13 machine-drafted locale keys across 8 locales need native review — including `tr-TR`**, which no commit body names. The commit bodies carry review notes for pt-BR, es-419, fr-FR, fr-CA, de-DE and it-IT only. Key parity itself is clean (0 missing / 0 extra across all 8 catalogs).

4. **Sessions already signed in when this deploys keep `hasPassword: undefined`** until their next `/users/me` refresh, and degrade to the previous AddDeviceModal copy. Correct behaviour (unknown → old copy), but the SSO-only user who most needs the fix is also least likely to re-login soon.

5. **`/auth/me`'s new `hasPassword` currently has no consumer** — ProfilePage reads `/users/me`. It is implemented and tested for API symmetry; drop it if you would rather not carry it.

## The bug this nearly shipped with

The freshness check originally bounded `auth_time` using `session.createdAt.getTime()`. `sso_sessions.created_at` is `timestamp without time zone`; postgres.js parses OID 1114 with a bare `new Date(x)`, so V8 reads the offsetless string as **local** time, while `auth_time` is a true UTC epoch. Drizzle *has* a UTC-correct path, but it is guarded on `typeof value === 'string'` and postgres.js hands it a pre-parsed `Date`, so it falls through — which is why the codebase looks like it handles this.

- Host behind UTC (any US dev box or self-hoster): bound lands hours in the future → `auth_time_stale` on **100%** of attempts. Feature dead on arrival.
- Host ahead of UTC: window *loosens* by the offset → a cached IdP session hours old passes as "fresh", weakening the only control this feature rests on.
- Containers run UTC, so hosted production was unaffected — which is exactly why it survived 25,500 passing tests.

Fixed via `utcMsFromOffsetlessTimestamp` (`services/sso.ts`), applied at the single call site. Measured against a live pg16: the same instant returned **+6h** under `America/Denver`, **−2h** under `Europe/Berlin`, exact under UTC. The route fixture previously fabricated `createdAt` as `new Date(Date.now() - n)` — a true instant, not what the driver produces — which is why the tests were blind; it now simulates the real driver path, and 14 route tests fail on revert.

## Verification

| Suite | Result |
|---|---|
| API unit (full) | 1,473 files / 25,530 tests, exit 0 |
| Web unit (full) | 644 files / 6,573 tests, exit 0 |
| `vitest.config.rls.ts` | 22 passed, exit 0 |
| `vitest.config.rls-coverage.ts` | 75 passed, exit 0 |
| `tenantCascade` + `tenant-export-policy` integration | 7 passed, exit 0 |
| `tsc --noEmit` (api, web) | exit 0 |
| Targeted suites under `TZ=America/Denver` | 92 passed — the host TZ that would have failed before |
| Migration | applies from zero on an empty DB, and re-applies as a no-op |
| Targeted suites under `TZ=UTC` **and** `TZ=America/Denver` | 99 each — the guard now bites in both |

Every task went through an independent task review, then a five-lens PR review (tests, silent failures, code, comments, types) which found two further Criticals and nine Importants — all fixed and re-reviewed. Reviews — not tests — caught every serious defect on this branch, including: a double-charged password step-up on `/mfa/enable` (two rate-limit slots and two argon2 verifies per call, with a possible 429 landing *after* `consumeMFAToken` burned the TOTP time-step), a security fix that had no test guarding it, a fail-open probe chain where a missing DB context made `userIsMfaProtected` answer "unprotected" for an account that holds factors, and — most instructively — **the timezone guard itself was vacuous on CI**: its assertion collapsed to `0 === 0` under UTC, no `TZ` is pinned anywhere, and GitHub runners are UTC, so deleting the timezone fix kept CI green. It only ever bit on a non-UTC developer machine. The assertions are now spy-based and simulate offsets from −345 to +420.

## Known gaps — follow-ups, not blockers

- **Three silent dead ends, newly reachable now that these accounts can enroll:** `/mfa/recovery-codes`, `DELETE /auth/passkeys/:id` and `POST /auth/mfa/disable` are still password-only, and the client greys the control rather than issuing a request — so there is no error, just a button that never works. These are removal/rotation operations, so an `enroll_first_factor` grant must **not** authorize them; they need the SR2-20 existing-factor road, which such a user can now take because they finally hold a factor.
- Post-IdP intent is not routed: the passkey card's re-verify CTA returns the user to the TOTP QR view and leaves a stale pending TOTP setup in Redis. Not a dead end (the grant is still held), but a confusing detour.
- The offsetless-timestamp trap is neutralised at this call site only. Every other `timestamp without time zone` column still parses as local time, and the next external-epoch comparison reintroduces it invisibly to UTC-based CI. A lint rule or contract test would close the class.
- Grant TTL is 300s, which spans a scan-QR-then-type `/mfa/setup` flow; expiry surfaces as `Invalid credentials`.
- No real-Postgres integration coverage for the reauth callback branch (same pre-existing gap as link mode).
- `mfaDisableSchema` inherits `ssoReauthGrantId` via `mfaVerifySchema.extend`; `/mfa/disable` never reads it. Not exploitable, but a disable endpoint silently accepting an enrollment-grant field is a poor shape.

## History note

Two `Revert` commits neutralize an unsanctioned commit from a concurrent session that briefly wrote into this worktree. `efc797048`'s message says fixture fixes are "addressed in the next commit" — they are in that same commit; the message is stale, the code is correct.

### DIFF
--- a/apps/api/migrations/2026-08-25-sso-sessions-reauth-user-id.sql
+++ b/apps/api/migrations/2026-08-25-sso-sessions-reauth-user-id.sql
@@ -1,0 +1,26 @@
+-- #4018: third sso_sessions mode. A "reauth" transaction proves a passwordless
+-- SSO account's identity through a fresh IdP round-trip so the user can enroll
+-- a FIRST MFA factor. Mirrors link mode (link_user_id + the three initiating_*
+-- binding columns), which this reuses as-is.
+--
+-- ON DELETE CASCADE for the same reason link_user_id has it: sso_sessions has
+-- no org_id/partner_id, so the tenant-erasure sweep never reaches it by tenancy
+-- and an abandoned transaction must not block a hard user delete.
+ALTER TABLE sso_sessions
+  ADD COLUMN IF NOT EXISTS reauth_user_id uuid REFERENCES users(id) ON DELETE CASCADE;
+
+-- A session is exactly one of: login (both NULL), link, or reauth. Both set is
+-- a programming error that would make the callback's mode discriminator
+-- ambiguous, so refuse it in the database rather than ordering the checks.
+DO $$
+BEGIN
+  ALTER TABLE sso_sessions
+    ADD CONSTRAINT sso_sessions_single_mode_chk
+    CHECK (link_user_id IS NULL OR reauth_user_id IS NULL);
+EXCEPTION
+  WHEN duplicate_object THEN NULL;
+END $$;
+
+CREATE INDEX IF NOT EXISTS sso_sessions_reauth_user_id_idx
+  ON sso_sessions (reauth_user_id)
+  WHERE reauth_user_id IS NOT NULL;
--- a/apps/api/src/db/schema/sso.ts
+++ b/apps/api/src/db/schema/sso.ts
@@ -125,6 +125,12 @@ export const ssoSessions = pgTable('sso_sessions', {
   // has no partner_id/org_id, so the tenant-cascade sweep never reaches it.
   linkUserId: uuid('link_user_id').references(() => users.id, { onDelete: 'cascade' }),
 
+  // Reauth-mode marker (#4018): when set, the callback mints a single-use
+  // enrollment step-up grant for this user instead of logging in or linking.
+  // Same ON DELETE CASCADE rationale as linkUserId above. Mutually exclusive
+  // with linkUserId (sso_sessions_single_mode_chk).
+  reauthUserId: uuid('reauth_user_id').references(() => users.id, { onDelete: 'cascade' }),
+
   // SR2-11 pending-transaction binding.
   //
   // providerVersion: sso_providers.config_version at creation. NULL only for
--- a/apps/api/src/db/schema/sso.ts
+++ b/apps/api/src/db/schema/sso.ts
@@ -137,12 +137,18 @@ export const ssoSessions = pgTable('sso_sessions', {
   // rows written before this column existed — the callback REJECTS those.
   providerVersion: integer('provider_version'),
 
-  // The three initiating_* columns are LINK-MODE ONLY (set by
-  // POST /sso/link/start alongside linkUserId). A login session has no
+  // The three initiating_* columns are set by BOTH user-initiated modes — link
+  // (POST /sso/link/start, alongside linkUserId) and reauth (POST
+  // /sso/reauth/start, alongside reauthUserId). A login session has no
   // initiating user, so they stay NULL there — which is why they are nullable.
-  // The link callback requires all three to be present AND to still match live
-  // state, which is what makes logout / password reset / MFA reset / suspension
-  // / global revocation invalidate a pending link.
+  // Both callbacks require all three to be present AND to still match live
+  // state (validateSessionBinding), which is what makes logout / password reset
+  // / MFA reset / suspension / global revocation invalidate a pending
+  // transaction in either mode.
+  //
+  // Reauth leans on them twice: once for that binding check, and again to stamp
+  // the minted step-up grant with the epochs and sid the transaction STARTED
+  // under, so a grant cannot outlive the session that authorized it.
   initiatingAuthEpoch: integer('initiating_auth_epoch'),
   initiatingMfaEpoch: integer('initiating_mfa_epoch'),
   // = refresh_token_families.family_id == the initiating access token's `sid`.
--- a/apps/api/src/middleware/auth.test.ts
+++ b/apps/api/src/middleware/auth.test.ts
@@ -98,7 +98,7 @@ vi.mock('../db/schema', () => ({
 }));
 
 import { Hono } from 'hono';
-import { authMiddleware, requireScope, requirePermission, requireMfa, requireOrg, requirePartner, requireOrgAccess, resolveOrgAccess, AuthContext } from './auth';
+import { authMiddleware, requireScope, requirePermission, requireMfa, requireOrg, requirePartner, requireOrgAccess, resolveOrgAccess, isMfaEnrollmentExemptPath, AuthContext } from './auth';
 import { verifyToken } from '../services/jwt';
 import { isTokenIssuedBeforePasswordChange, isUserTokenRevoked } from '../services/tokenRevocation';
 import { db, withDbAccessContext } from '../db';
@@ -547,6 +547,19 @@ describe('authMiddleware', () => {
     expect(res.status).not.toBe(428);
     expect(res.status).toBe(200);
     expect(vi.mocked(getEffectiveMfaPolicy)).not.toHaveBeenCalled();
+  });
+
+  it('exempts /sso/reauth/* from forced MFA enrollment', () => {
+    // A policy-required, unenrolled, passwordless SSO user is the entire target
+    // population. Without this exemption authMiddleware 428s them before the
+    // handler runs and the enrollment flow can never start.
+    expect(isMfaEnrollmentExemptPath('/api/v1/sso/reauth/start')).toBe(true);
+    expect(isMfaEnrollmentExemptPath('/sso/reauth/start')).toBe(true);
+  });
+
+  it('does not exempt other /sso paths', () => {
+    expect(isMfaEnrollmentExemptPath('/api/v1/sso/providers')).toBe(false);
+    expect(isMfaEnrollmentExemptPath('/api/v1/sso/link/start/abc')).toBe(false);
   });
 
   it('permits an enrolled user without consulting the resolver at all', async () => {
--- a/apps/api/src/middleware/auth.ts
+++ b/apps/api/src/middleware/auth.ts
@@ -197,7 +197,7 @@ export function siteAccessCheck(
  *
  * Path is the API path *after* the `/api/v1` mount, e.g. `/auth/mfa/setup`.
  */
-function isMfaEnrollmentExemptPath(path: string): boolean {
+export function isMfaEnrollmentExemptPath(path: string): boolean {
   // Strip the /api/v1 prefix if present so the check works whether Hono
   // gives us the absolute path or a sub-app path.
   const rel = path.startsWith('/api/v1') ? path.slice('/api/v1'.length) : path;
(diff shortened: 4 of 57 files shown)


ToddHebebrand:
Reviewed against `origin/main`. Approving and merging — it closes a p1 that currently leaves SSO accounts unable to satisfy `requireMfa()` at all, so shipping it is itself the security improvement.

**The identity check is the right one, and it is the one this design lives or dies on.** Requiring that `(provider, sub)` **already** maps to `reauth_user_id` — rather than comparing emails the way link mode does — closes the attack that would otherwise make the whole feature unsafe: on an IdP where users control their own email address, an email match would let one user re-authenticate as another. The comment says exactly that, and the code does exactly that.

**Migration is clean and correctly reasoned.** Idempotent throughout (`ADD COLUMN IF NOT EXISTS`, `DO $$ ... EXCEPTION WHEN duplicate_object`, partial index `IF NOT EXISTS`), no inner `BEGIN`/`COMMIT`. The `sso_sessions_single_mode_chk` CHECK is the right call — making the mode discriminator unambiguous in the database rather than ordering the checks in the callback is what stops a future reordering from silently changing which branch runs. And the header states why no cascade or export-policy registration is needed (`sso_sessions` carries no `org_id`/`partner_id`), which is the registration step that has shipped broken five times when left implicit.

**Defence in depth is layered rather than piled up.** Rate limit 5/15min per user; 10-minute session expiry; `prompt=login&max_age=0` plus a fresh `auth_time` requirement; live re-validation of `{authEpoch, mfaEpoch, sid}` against the initiating session and refresh family; a re-check that the account is *still* passwordless in case one was set mid-flow; grant scoped to a new `enroll_first_factor` operation rather than blanket step-up; public error codes kept coarse with the precise binding reason audit-only. Taking the narrowed values off `binding` instead of re-reading the session row with `!` is a real correctness point, not style — the comment explaining why is worth keeping.

---

**Verified:** the migration text and its idempotency; the `(provider, sub) → reauth_user_id` check; the passwordless re-check; the `tenantCascade.ts` change is comment-only apart from covering the new column; `sso_sessions` has no `org_id`, so no cascade/export registration applies; CI green at 57 SUCCESS / 1 skipped, which means the Integration Tests job — the only place the cascade and export-policy contracts can fail — actually ran.

**Inferred:** that callback replay is bounded by the HMAC-bound single-use state plus the epoch binding rather than by row consumption. The `sso_sessions` row is **not** deleted on consumption — that is pre-existing behaviour shared with login and link mode, not introduced here — so the session survives until its 10-minute expiry. A second grant minted inside that window should be dead on arrival because enrolling the first factor bumps `mfaEpoch`, but I reasoned that from the binding rather than observing it.

**Not checked:** the web components (`ProfilePage`, `MFASettings`, `AddDeviceModal`) beyond confirming tests were added alongside; the `pgOffsetlessTimestamp` helper; whether the CHECK constraint and the `ON DELETE CASCADE` behave as intended against real Postgres — which is exactly what #4049 says is missing.

**#4049 is the follow-up that matters** and should not sit. The DB-level invariants this PR adds — the single-mode CHECK and the FK cascade — are the parts with no real-Postgres coverage, and they are also the parts a unit test structurally cannot exercise. Everything else in the deferred list (#4050 through #4055) reads as genuine cleanup.

Two of those deserve calling out as good self-reporting rather than backlog padding: #4051/#4052 state plainly that a passkey alone still cannot reach L4 and that no client sends the L4 fields, so the tier is honestly documented as incomplete rather than implied to work. #4054 catches a *stale comment* claiming `SameSite=Lax` omits the cookie on the callback — it does not, and the real protection is the HMAC. Correcting that matters more than it looks, because the wrong explanation is an argument for deleting the thing actually doing the work.

PR #4043, opened 2026-08-26
fix(ci): reconcile the drift detector with side-branch hotfix releases (#4037)

### LINKED ISSUE #4037: [CI] Drift detector red since 08-24: the side-branch hotfix release process collides with its ancestry invariant, and its runbook link is dead on main
## Description

`Main History Drift Detector` has been red on `main` since 2026-08-24 (it passed on 08-22 and 08-23). The alarm is **correct** — the tag genuinely is unreachable — but the diagnosis it prints is wrong for this shape, and the runbook it points at does not exist on `main`.

```
main has lost ancestry to latest release tag v0.107.1
main can no longer reach v0.107.1 via parent links — main was probably force-pushed.
See docs/runbooks/2026-05-02-main-orphan-and-userhelper-flag.md.
```

## What is actually true

- `v0.107.1` peels to `70d4a326c`, which is contained only by `origin/hotfix/0.107.1`.
- `git merge-base --is-ancestor v0.107.1 origin/main` is false, so the check fires correctly.
- The same change is on `main` as `3946b27bf`. `git cherry origin/main v0.107.1` returns `- 70d4a326c`, and the **patch delta is identical** — all seven changed paths have matching pre/post modes and blob IDs.
- `main` lacks exactly that one commit and is well ahead of the tag.

Worth stating precisely: identical patch delta is **not** proof that history was not rewritten. `git patch-id` ignores whitespace, hunk offsets, file ordering, and all commit metadata, and a force-push can preserve every patch. What supports "not force-pushed" here is separate — this clone's `origin/main` reflog shows only fast-forwards across the interval. Authoritative ref history on the GitHub side would be stronger evidence than anything I can see locally.

## This is recurring, not a one-off

`v0.105.1` and `v0.105.2` are unreachable in exactly the same way, each contained only by its own `origin/hotfix/*` branch, each with an equivalent patch on `main`. The detector predates them.

So this is not an edge case that appeared with `v0.107.1` — it is a standing collision between two approved things: **the side-branch hotfix release process, and the detector's topology invariant.** The detector selects `LATEST_TAG=$(git tag --sort=-creatordate -l 'v*' | head -1)` (`drift-detector.yml:57`) — newest by creation date, prereleases included — so any hotfix tag cut this way becomes the selected tag and pins the job red until a newer `v*` supersedes it. On a daily `0 14 * * *` cron, that is a red every day in between.

## Two things worth separating

**1. The diagnosis, which is the cheap fix.** "Unreachable tag" is observed; "main was probably force-pushed" is an inference that does not hold for a cherry-picked hotfix tag. Those should not be one sentence. And `docs/runbooks/2026-05-02-main-orphan-and-userhelper-flag.md` **is not on `main`** — it exists only on the unmerged `origin/docs/incident-runbook-orphan-2026-05-02`. Anyone following the error today lands on nothing.

**2. The invariant, which is yours to decide.** I do not think a generic patch-equivalence fallback is the right answer, and I want to argue against my own first instinct here: making the check pass whenever `git cherry` reports `-` would defeat what it exists for. A force-push that preserved the genesis root and rebased history into patch-equivalent commits would satisfy that fallback and report green — which is precisely the severed-parent-link case the detector was built to catch (`drift-detector.yml:3-14`). It would also mis-handle squash-merges, since `git cherry` matches individual patches rather than cumulative state, and it is blind to metadata, signatures and trailers.

A narrower reconciliation seems better than a bypass — for example recording hotfix provenance explicitly so the check can distinguish "tag cut from an approved side branch" from "ancestry lost", and giving those two cases different diagnostics. But which way to reconcile depends on what the release process intends for hotfix tags, which is your call.

## What I checked, and what I did not

Verified against `origin/main`: the ancestry results and containing branches for `v0.107.1`, `v0.105.2`, `v0.105.1`, `v0.106.0` and `v0.107.0`; the identical patch delta (modes and blob IDs per path); the detector's tag selection, ancestry assertion and cron; and that the referenced runbook is absent from `main` but present on the unmerged docs branch. I read the failing run's log for the exact error text.

**Not verified:** I have no authoritative view of GitHub-side ref history, so "not force-pushed" rests on this clone's reflog. I also have not traced what the release workflow intends to do about merging hotfix commits back, and I did not re-run the two failing jobs myself — the conclusions come from the run list and one log.

ToddHebebrand:
Closes #4037. Reported and root-caused by @bdunncompany.

`Main History Drift Detector` has been red on `main` since 2026-08-24 (green 08-20 → 08-23). Verified independently before fixing: `v0.107.1` peels to `70d4a326c`, contained only by `origin/hotfix/0.107.1`; `v0.105.2` and `v0.105.1` are unreachable in exactly the same way, each contained only by its own hotfix branch. The alarm is correct. The diagnosis was not.

### Three separable defects

**The message asserted a cause it had not established.** "main can no longer reach the tag" is observed; "main was probably force-pushed" is an inference that is false for a tag cut on a hotfix branch. The two causes have opposite remediations, and the message only named one.

**The runbook link was dead on main.** `docs/runbooks/2026-05-02-main-orphan-and-userhelper-flag.md` was referenced three times and existed only on the unmerged `docs/incident-runbook-orphan-2026-05-02` branch. Anyone following the error landed on nothing. Brought onto main unchanged, 222 lines, single file.

**The invariant genuinely collides with the release process.** Tag selection is newest-by-`creatordate`, so a side-branch hotfix tag becomes the selected tag and pins the job red until a newer `v*` supersedes it — every day, on a daily cron. This is not specific to `v0.107.1`; the detector predates `v0.105.1`.

### Why the exemption is a committed file rather than an inference

The obvious fix is to pass whenever `git cherry` or `git patch-id` finds an equivalent patch on main. **That would defeat the detector.** A force-push that rebased history into patch-equivalent commits would satisfy it and report green — which is precisely the severed-parent-link case the check was built for after the 2026-04-25 incident. `patch-id` is additionally blind to commit metadata, signatures and trailers, and mismatches squash-merges.

Billy made that argument against his own first instinct in #4037, and it is the reason this fix has the shape it does.

So `.github/release-provenance/side-branch-tags.tsv` records `tag → main-side equivalent commit` for the three affected releases. Two properties make it safe:

- a force-push cannot add a row without that addition being a **visible commit on main**, which is the thing an inference cannot give you;
- the recorded commit is **re-verified as reachable on every run**, so a stale row cannot mask a later genuine loss of history.

The unreachable-and-unrecorded path still fails. It now describes both possible causes and what to check, and explicitly warns against "fixing" it with patch-equivalence — the failure message is where the next person will look, so the reasoning lives there rather than only in a PR body.

### Verification

Ran the step's script against real repository state, four cases:

| case | expected | observed |
|---|---|---|
| reachable tag (`v9.9.9` on main HEAD) | fast path, exit 0 | `OK: ... and reaches v9.9.9-drifttest` |
| unreachable + recorded + equivalent reachable | exit 0, distinct message | side-branch message, exit 0 |
| unreachable + **unrecorded** | exit 1, two-cause diagnosis | exit 1, correct text |
| unreachable + recorded but equivalent **unreachable** | exit 1, stale-exemption | exit 1, correct text |

The third and fourth are the ones that matter — they confirm the exemption is narrow rather than a bypass.

### Note on process

The preferred remedy for future hotfixes is still to **merge the hotfix branch into main** so the tag is reachable and no row is needed. A row here is the fallback for releases already cut that way. That is stated in the file header so the next hotfix does not reach for the file first.

Billy also flagged one limit worth preserving: "not force-pushed" here rests on a local clone's reflog, not on authoritative GitHub-side ref history. This change does not alter that — it only stops the detector from firing on a case that is not a force-push.

### DIFF
--- a/.github/release-provenance/side-branch-tags.tsv
+++ b/.github/release-provenance/side-branch-tags.tsv
@@ -1,0 +1,31 @@
+# Release tags cut from a side branch, whose commit is therefore NOT reachable
+# from main by parent links.
+#
+# WHY THIS FILE EXISTS
+# The Main History Drift Detector asserts that main can still reach the latest
+# v* tag. That assertion is how a force-push severing main's history is caught
+# within 24h (see the 2026-04-25 incident and docs/runbooks/2026-05-02-main-
+# orphan-and-userhelper-flag.md). A hotfix released from an `origin/hotfix/*`
+# branch violates the same assertion for an entirely legitimate reason: the
+# fix reaches main as an EQUIVALENT commit, never as that commit.
+#
+# WHY IT IS A FILE AND NOT AN INFERENCE
+# The tempting shortcut is to let the detector pass whenever `git cherry` or
+# `git patch-id` finds an equivalent patch on main. That would defeat the
+# detector: a force-push that rebased history into patch-equivalent commits
+# would satisfy it and report green, which is precisely the severed-parent-link
+# case the check exists to catch. patch-id is also blind to commit metadata,
+# signatures and trailers, and mismatches squash-merges. So the exemption has
+# to be an explicit, reviewed, committed fact — a force-push cannot add a row
+# here without that addition itself being a visible commit on main.
+#
+# The recorded main-side commit is re-verified as reachable on every run, so an
+# entry cannot mask a later loss of that commit either.
+#
+# FORMAT  <tag><TAB><main-side equivalent commit><TAB><note>
+# Add a row when, and only when, a release is tagged on a branch that is not
+# merged back into main. Prefer merging the hotfix branch into main instead;
+# a row here is the fallback for releases already cut that way.
+v0.105.1	01f71e2547381e6158f4f9cc7cbd9a6bc8454b0b	quote PDF image block overlap (#3481)
+v0.105.2	8582acb9dc69e15d6b172e2f30139c7719ea8eaa	quote PDF contract placeholder + cover background (#3489)
+v0.107.1	3946b27bf05cb9803b284eb34a668851b014f99f	hosted agent update 409 (#3836, #3872)
--- a/.github/workflows/drift-detector.yml
+++ b/.github/workflows/drift-detector.yml
@@ -13,6 +13,15 @@
 # Triggered by the 2026-04-25 main-orphan incident, when ~6 days of
 # parallel-line releases shipped before the topology was noticed.
 #
+# Releases tagged on an `origin/hotfix/*` branch are NOT reachable from main —
+# the fix lands on main as an equivalent commit, never as that commit. Those
+# tags are exempted by an explicit committed row in
+# .github/release-provenance/side-branch-tags.tsv, whose recorded main-side
+# commit is re-verified as reachable on every run. The exemption is deliberately
+# a reviewed file rather than a patch-id inference: a force-push that rebased
+# history into patch-equivalent commits would satisfy an inference and report
+# green, which is the exact case this detector exists to catch (#4037).
+#
 # This is a monitor, not a required status check. Failures surface via
 # GitHub-native error markup and the standard workflow-failure email.
 # See docs/runbooks/2026-05-02-main-orphan-and-userhelper-flag.md.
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
End-to-end proof, not just the local script run: dispatched `Main History Drift Detector` against this branch — [run 32926924390](https://github.com/LanternOps/breeze/actions/runs/32926924390) **succeeded**, against `failure` on main for 08-24 and 08-25. The job resolves `origin/main` regardless of the ref it runs on, so this is the real ancestry check passing with the provenance row, not a branch-local artifact.

PR #4056, opened 2026-08-26
fix(api,ci): pin a non-UTC TZ for auth/SSO test runs (#4046)

### LINKED ISSUE #4046: CI never pins a non-UTC TZ, so timezone-correctness tests are structurally vacuous
### Description

Neither `apps/api/vitest.config.ts` nor `.github/workflows/ci.yml` sets a `TZ` environment variable. GitHub-hosted runners default to UTC. Any test written to assert timezone-correcting behavior — comparing an offsetless-timestamp-derived value against an epoch, applying a UTC offset correction, etc. — runs with the host offset pinned at exactly 0, where the bug class described in the companion issue ("offsetless-timestamp-vs-epoch comparisons") is dormant by construction. Such a test can be 100% green in CI while asserting nothing about the behavior it claims to cover.

This is exactly how `#4018`'s TZ bug (fixed in PR #4041; `sso_sessions.created_at` compared against `auth_time` without correcting for local-time parsing) survived the existing suite: the ledger for that branch notes 25,508 API tests were green at `HEAD` under default (UTC) `TZ`, and the bug was only caught by a human final-review pass that reasoned about the code, then confirmed by re-running under `TZ=America/Denver`.

### Why it matters

Every future test of TZ-sensitive logic is at risk of the same false confidence unless the author remembers to manually run it under a non-UTC zone locally — which the repo's documented workflow doesn't prompt anyone to do, and which CI cannot verify by re-running. "Passes in CI" currently does not imply "is timezone-correct" for any of this code, silently.

### Affected files

- `apps/api/vitest.config.ts` — no `TZ` set for the unit config.
- `.github/workflows/ci.yml` — `test-api` (and other Node-based jobs) inherit the runner's default UTC with nothing overriding it.
- `apps/api/src/services/sso.test.ts:640-658` — `offsetlessTimestampFromHostAt()`, the pattern this branch used to get real teeth under both `TZ=UTC` and `TZ=America/Denver` by simulating the driver parse and overriding `getTimezoneOffset()` per-test — a per-test workaround for the absence of a runner-level pin, not a substitute for one.

### Suggested approach

Pin `TZ` for at least the auth/SSO-adjacent suites to a non-UTC value in CI — either:
- Add `TZ=America/Denver` (or similar) at the job/step level in `ci.yml` for `test-api` (simplest, but changes the ambient timezone for every test in the job — verify nothing else in the suite assumes UTC in a way that would newly break), or
- Run those suites twice in a matrix (`TZ: [UTC, America/Denver]`), which is more CI time but isolates which zone a regression shows up under and avoids risking an unrelated UTC-assuming test breaking silently under a blanket pin.

**Important constraint:** setting `process.env.TZ` from inside a test file does **not** work — Node/V8 cache the resolved timezone at first use (of `Date`, `Intl`, etc.) within a process, so a mid-file assignment has no effect on already-initialized `Date` behavior in that worker. The pin has to happen before the Node process starts (runner env, `vitest.config.ts`'s `env` block passed to workers, or a `globalSetup` that runs in a separate process) — not inside a test or `beforeEach`.

ToddHebebrand:
## Summary

GitHub-hosted runners default to UTC, and neither `.github/workflows/ci.yml` nor `apps/api/vitest.config.ts` ever pins a different zone. A test written to assert timezone-correcting behavior (e.g. comparing an offsetless-timestamp-derived value against a UTC epoch) is dormant by construction under UTC — the assertion can collapse to `0 === 0` and stay green while the underlying conversion is missing or wrong.

**Verified the issue's premise directly**, rather than taking it on faith: this is exactly how #4018's TZ bug shipped past 25,000+ green tests in PR #4041 (`sso_sessions.created_at` compared against `auth_time` without correcting for local-time parsing) — that branch had to build a per-test `getTimezoneOffset()` override shim (`offsetlessTimestampFromHostAt` in `apps/api/src/services/sso.test.ts`) purely to get real teeth under UTC, because the runner itself never varies the zone.

Closes #4046.

## What changed

- **`apps/api/vitest.config.tz.ts`** (new) — a narrow vitest config scoped to the auth/SSO/MFA/passkey unit suites where this bug class lives (33 files, 677 tests).
- **`apps/api/package.json`** — new `test:tz` script (`vitest run --config vitest.config.tz.ts`), following the existing `test:rls` / `test:integration` convention of one script per dedicated config.
- **`.github/workflows/ci.yml`** — new step in the existing `test-api` job: `TZ=America/Denver pnpm --filter=@breeze/api test:tz`, right after the normal `Run API tests` step.

## Why this scope, not a bigger one

- **Not a blanket TZ pin on the whole job.** Changing the ambient TZ for every test in `test-api` risks silently breaking an unrelated test that assumes UTC, and the failure would be indirect (some unrelated suite goes red with no obvious link to a TZ change).
- **Not a UTC/non-UTC matrix for the whole suite.** That doubles the cost of every unrelated test for zero coverage gain — only timestamp/timezone arithmetic is offset-sensitive.
- **This is a second, targeted run of just the suites where the bug class lives**, under a fixed non-UTC zone (`America/Denver`, deliberately a >0 UTC offset west and a non-half-hour DST transition, matching the zone #4018 was caught under). It's additive (~12s locally) and isolates any regression to a code area that's actually plausible.

## Why the TZ pin lives in the CI step, not the vitest config

Node/V8 cache the resolved timezone at first use of `Date`/`Intl` within a process. Setting `process.env.TZ` from inside a test file (or `beforeEach`) has no effect on already-initialized `Date` behavior in that worker. The pin has to happen **before the Node process starts** — so it's set in the GitHub Actions step's `env:` block, which is in place before `vitest` (and its worker processes) ever launch.

**Verified locally**, not assumed: comparing `TZ=UTC` vs `TZ=America/Denver` against a throwaway probe test showed `new Date(...).getTimezoneOffset()` reading `0` vs `360` respectively inside the actual vitest worker — confirming the env var propagates to where test code runs, not just the parent CLI process.

## Testing

- `TZ=America/Denver pnpm --filter=@breeze/api test:tz` → 33 files, 677 tests, all green.
- Same targeted files also verified to run unaffected under the existing default `vitest.config.ts` (no double-run conflict between the two CI steps).
- `TZ=UTC` vs `TZ=America/Denver` probe test confirms the pin reaches the vitest worker process before any `Date`/`Intl` use.
- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` — clean.
- No migrations, schema, or tenancy-scoped code touched; no RLS/cascade/export-policy registration applies.

## Non-conflict with PR #4041

This PR does not touch any file PR #4041 modifies (confirmed via `gh pr view 4041 --json files`), and #4041 does not touch `ci.yml`, `apps/api/package.json`, or `apps/api/vitest.config.ts`/`vitest.config.tz.ts`. No expected conflict.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,6 +174,26 @@ jobs:
       - name: Run API tests
         run: pnpm test --filter=@breeze/api
 
+      # Timezone-correctness regression gate (#4046): GitHub-hosted runners
+      # default to UTC, and nothing else in this job or in vitest.config.ts
+      # overrides that. A test asserting timezone-correcting behavior (e.g.
+      # comparing an offsetless-timestamp-derived value against a UTC epoch)
+      # is dormant by construction under UTC — the assertion can collapse to
+      # `0 === 0` and stay green forever while the underlying conversion is
+      # missing or wrong. That is exactly how #4018's TZ bug shipped past
+      # 25,000+ green tests. Re-run just the auth/SSO/MFA/passkey suites
+      # (vitest.config.tz.ts) under a fixed non-UTC zone so a regression of
+      # that shape fails here instead of passing vacuously — not the full
+      # suite a second time, since only timestamp/timezone arithmetic is
+      # offset-sensitive. TZ is set here, at the step level, before the Node
+      # process starts: assigning process.env.TZ from inside a test file has
+      # no effect, because Node/V8 cache the resolved zone at first Date/Intl
+      # use within a process.
+      - name: Run API tests (auth/SSO, pinned non-UTC TZ)
+        env:
+          TZ: America/Denver
+        run: pnpm --filter=@breeze/api test:tz
+
       - name: Run load-test static contract tests
         run: pnpm test:load-static
 
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -33,6 +33,7 @@
     "test:rls-coverage": "vitest run --config vitest.config.rls-coverage.ts",
     "test:request-db-role": "vitest run --config vitest.config.request-db-role.ts",
     "test:site-scope-coverage": "vitest run --config vitest.config.site-scope-coverage.ts",
+    "test:tz": "vitest run --config vitest.config.tz.ts",
     "test:run": "vitest run"
   },
   "dependencies": {
--- a/apps/api/src/__tests__/tzPinCanary.test.ts
+++ b/apps/api/src/__tests__/tzPinCanary.test.ts
@@ -1,0 +1,22 @@
+import { describe, it, expect } from 'vitest';
+
+/**
+ * Canary for issue #4046: this file must ONLY run as part of the pinned
+ * non-UTC pass (`apps/api/vitest.config.tz.ts`, wired into CI's `test-api`
+ * job via `pnpm test:tz` with `TZ=America/Denver` set at the step level) —
+ * it is excluded from the default `vitest.config.ts` unit run (see its
+ * `exclude` list) specifically so this assertion has teeth in both places:
+ * green here confirms the pin reached this worker process; if it ever ran
+ * under the plain UTC job instead, it would fail there.
+ *
+ * If a future refactor of ci.yml or vitest.config.tz.ts silently drops the
+ * TZ pin (the step's `env:` block gets deleted, the include entry below
+ * gets removed, etc.), this test fails loudly instead of the whole
+ * tz-pinned job quietly degrading back into a vacuous UTC run — which is
+ * exactly the failure mode #4046 exists to prevent.
+ */
+describe('non-UTC TZ pin (issue #4046)', () => {
+  it('the process actually observes a non-UTC offset', () => {
+    expect(new Date().getTimezoneOffset()).not.toBe(0);
+  });
+});
--- a/apps/api/vitest.config.ts
+++ b/apps/api/vitest.config.ts
@@ -174,6 +174,13 @@ export default defineConfig({
       // Real-DB suites owned by vitest.integration.config.ts (#3778).
       'src/services/invoiceService.issue.integration.test.ts',
       'src/services/invoicePdf.integration.test.ts',
+      // Canary for issue #4046: asserts the process observes a non-UTC
+      // offset. It must ONLY run under the pinned non-UTC pass
+      // (vitest.config.tz.ts, TZ=America/Denver), where it belongs — it is
+      // excluded here deliberately, not because it doesn't apply to the
+      // unit runner: it WOULD fail on this (UTC) runner, since that is
+      // exactly the point of the check.
+      'src/__tests__/tzPinCanary.test.ts',
     ],
     setupFiles: ['src/__tests__/setup.ts'],
     coverage: {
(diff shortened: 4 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, pr-test-analyzer)

**Findings:** 5 raised → 2 addressed in `ef07c1281`, 2 deferred to a filed follow-up (#4059), 1 low-urgency suggestion left as-is.

- **Addressed:** no canary asserted the TZ pin itself stays active (a future `ci.yml`/`vitest.config.tz.ts` refactor could silently drop it and nothing would flag the regression) → added `apps/api/src/__tests__/tzPinCanary.test.ts` (`getTimezoneOffset() !== 0`), included only in `vitest.config.tz.ts`, excluded from the main config. Verified locally: passes under `vitest.config.tz.ts` + `TZ=America/Denver`, fails under `vitest.config.tz.ts` + `TZ=UTC`, and is correctly excluded (not discovered) under the main unit config.
- **Addressed:** the include list in `vitest.config.tz.ts` is a manually-curated allowlist with no forcing function, and PR #4041 (still open) will add two new TZ-sensitive test files not yet on `main` → left an explicit maintenance comment naming both files so they aren't missed when #4041 merges.
- **Deferred to #4059:** `tokenRevocation.ts`'s `isTokenIssuedBeforePasswordChange` compares a JWT `iat` against the legacy timestamp-without-tz `passwordChangedAt` column — same bug shape, live on every authenticated request — but its test fixtures use literal `Date` instants and would need reshaping (not just inclusion) to get real teeth under the pin. Out of scope for a config-only CI fix.
- **Deferred to #4059:** the pin is scoped to auth/SSO/MFA/passkey suites only; ~806 timestamp-without-tz columns exist repo-wide, with concrete TZ-sensitive logic also in `automationRuntime.ts`, `reportScheduleWorker.ts`, `patchSchedulerWorker.ts`, `discoveryWorker.ts`, and `pamRuleEngine.ts`. Kept incremental per the PR's own proportionality rationale rather than expanding into a repo-wide audit here.
- **Not acted on (low urgency):** `apps/api/src/routes/sso.test.ts` (largest SSO-adjacent file) is omitted from the include list — not currently TZ-sensitive on `main` (uses `Date.now() + offset`, not offsetless-timestamp comparisons), worth a second look once #4041 lands.

**Tests:** `TZ=America/Denver pnpm --filter=@breeze/api test:tz` — 34 files, 678 tests, all green. `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` clean. Confirmed no file overlap with PR #4041's diff.

ToddHebebrand:
Tier-1, merging. The canary is the part that makes this more than a config change: `tzPin.test.ts` asserts the process actually observes a non-UTC offset, and it is excluded from the default UTC run so it has teeth in both directions — if the `env:` block or the include entry is ever dropped, the job fails loudly instead of quietly degrading back into the vacuous UTC pass it exists to prevent. `America/Denver` is the right pick since it observes DST and so is never offset 0. Scoping it to the auth/SSO suites rather than re-running everything is also right — only timestamp arithmetic is offset-sensitive.

PR #4057, opened 2026-08-26
fix(config,api): pass DB password vars through turbo + fail fast on skipped breeze_app bootstrap (#4048)

### LINKED ISSUE #4048: `turbo.json` has no `globalEnv`, so `pnpm db:migrate` silently drops `BREEZE_APP_DB_PASSWORD`/`POSTGRES_PASSWORD` and fails 135 files later with a misleading error
### Description

`turbo.json` declares `globalPassThroughEnv: [DATABASE_URL, REDIS_URL, NODE_ENV]` but no `globalEnv`, and the `db:migrate` task entry has no per-task `env` either. Turbo strips any environment variable not declared in one of those lists before invoking the underlying script. `BREEZE_APP_DB_PASSWORD` and `POSTGRES_PASSWORD` are in neither list, so when `pnpm db:migrate` is run through turbo, those variables never reach the migration process — even if they're correctly set in the calling shell.

Inside the migration process, `ensureAppRole()` (`apps/api/src/db/ensureAppRole.ts:14-28`) checks for a password via `selectAppRolePassword()`; when both env vars are absent it logs a warning and returns early:

```
[ensure-app-role] Neither BREEZE_APP_DB_PASSWORD nor POSTGRES_PASSWORD is set — skipping breeze_app role setup. RLS will NOT be enforced against the admin connection.
```

Migrations then proceed to run as the admin/superuser connection (since `breeze_app` was never created/ensured), and typically get well into the migration file list — 135 files, per the branch's ledger — before hitting one that explicitly references the `breeze_app` role (an RLS policy grant, most commonly), which fails with Postgres error `42704: role "breeze_app" does not exist`.

**`DATABASE_URL` does get through** (it's on the pass-through list), which is what makes this deceptive: the connection itself visibly works, the failure surfaces far downstream, and the error text names a missing database role — nothing about it points back to "an environment variable was silently stripped before the process started."

This was hit directly on the #4018 branch (PR #4041; ledger: "Env note: turbo strips undeclared env vars ... `pnpm db:migrate` silently skips ensureAppRole -> migrations fail 42704 on breeze_app.") and worked around by bypassing turbo entirely: `cd apps/api && DATABASE_URL=... BREEZE_APP_DB_PASSWORD=... npx tsx scripts/migrate.ts`.

### Why it matters

Affects any database that Docker Compose didn't already provision with `breeze_app` pre-created: a fresh local dev DB, a scratch/throwaway container (exactly the kind spun up per-worktree for isolated development, as this branch did), or a restored dump being migrated forward. `pnpm db:migrate` is the command CLAUDE.md documents for this — so the standard, documented path fails with an error that gives no hint about turbo's env-stripping as the cause, sending anyone who hits it down a role-permissions debugging path instead of an env-var one.

### Affected files

- `turbo.json:3-7` — `globalPassThroughEnv` list, missing `BREEZE_APP_DB_PASSWORD` / `POSTGRES_PASSWORD`.
- `turbo.json:29-31` — the `db:migrate` task entry, which has no task-level `env` override either.
- `apps/api/src/db/ensureAppRole.ts:14-28` — where the silent skip happens (the warning it does log only appears if you're watching stdout for a `pnpm db:migrate` run through turbo, and doesn't propagate as an error).

### Suggested approach

Add `BREEZE_APP_DB_PASSWORD` and `POSTGRES_PASSWORD` to `turbo.json`'s `globalPassThroughEnv` (or a task-scoped `env`/`passThroughEnv` on `db:migrate` specifically, if the global list is meant to stay minimal). Consider also hardening `ensureAppRole`'s warning path — since a silent `console.warn` followed by a failure 135 files later is a bad debugging experience even for someone who *did* set the vars but hit a different pass-through gap in the future — e.g. by having `autoMigrate` fail fast with a pointed error when `ensureAppRole` skipped and the subsequent migration run then hits `42704` on `breeze_app`, rather than letting Postgres's generic role-not-found error be the only signal.

ToddHebebrand:
## Summary

`turbo.json` declared `globalPassThroughEnv: [DATABASE_URL, REDIS_URL, NODE_ENV]` but not `BREEZE_APP_DB_PASSWORD`/`POSTGRES_PASSWORD`. Turbo strips any env var not in that list (or a task's own `env`) before invoking the script, so `pnpm db:migrate` (run through turbo) never received either password var — even when correctly set in the calling shell.

Because `DATABASE_URL` *was* pass-through-listed, the connection still worked, which made the failure deceptive: `ensureAppRole()` silently warned and skipped `breeze_app` bootstrap, migrations then ran ~135 files deep as the admin/superuser, and finally died on Postgres `42704: role "breeze_app" does not exist` — an error naming an RLS grant, not the missing env var. Affects any DB compose didn't already provision with `breeze_app` (fresh dev DBs, scratch containers, restored dumps).

## Fix decision (per the issue's ask)

- **turbo.json fix — done.** Added `BREEZE_APP_DB_PASSWORD` and `POSTGRES_PASSWORD` to `globalPassThroughEnv`, matching the existing `DATABASE_URL`/`REDIS_URL` pattern.
- **Cache-invalidation check — done, no concern.** Verified empirically with `turbo run <task> --dry-run=json`: the hash for a cacheable task (`lint`) was byte-identical whether the two vars were unset, or set to two different values. `passThroughEnv`/`globalPassThroughEnv` vars are excluded from turbo's cache key by design (unlike `globalEnv`, which would have poisoned every cached task's key on every local password rotation). `db:migrate` itself also already has `cache: false`, so this was moot for that task specifically, but the global list applies to every task.
- **CLAUDE.md doc change — not done, and not needed.** The documented `pnpm db:migrate` command becomes correct again once the env var actually reaches the process; there's no remaining gap to caveat or a workaround to record.
- **Extra hardening (from the issue's "suggested approach"), done:** `ensureAppRole()` now returns whether it actually ran (`false` when both password vars are absent). `autoMigrate()` checks that against whether `breeze_app` already exists — a compose-provisioned DB with the role pre-created still skips harmlessly — and fails fast with a message naming the missing env vars and turbo's `globalPassThroughEnv`, instead of letting the migration loop run ~135 files deep into an opaque `42704`. This closes the failure mode for any *future* pass-through gap too, not just this one.

## Files changed

- `turbo.json` — added the two vars to `globalPassThroughEnv`.
- `apps/api/src/db/ensureAppRole.ts` — `ensureAppRole()` now returns `Promise<boolean>` (ran vs. skipped) instead of `Promise<void>`.
- `apps/api/src/db/autoMigrate.ts` — checks the return value before applying RLS-policy migrations; fails fast with a pointed error if `breeze_app` doesn't exist and bootstrap was skipped.

## Verification

- `npx tsc --noEmit` (apps/api, with increased heap — the default OOMs on this codebase) — clean.
- `vitest run src/db/autoMigrate.test.ts` — 58/58 passed.
- **Real end-to-end run against a disposable Postgres 16 container** with no pre-existing `breeze_app` role (verifying, not just inspecting config):
  - **Pre-fix repro** (`git stash` on `turbo.json` only): `pnpm db:migrate` via turbo reproduces the silent skip → now hits the new fail-fast error immediately, naming the real cause, instead of a downstream `42704`.
  - **Post-fix**: `pnpm db:migrate` via turbo bootstraps `breeze_app` (confirmed `rolsuper=f`, `rolbypassrls=f`) and completes all 567 migrations + seed successfully.

Closes #4048

### DIFF
--- a/apps/api/src/db/autoMigrate.test.ts
+++ b/apps/api/src/db/autoMigrate.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, describe, it, expect } from 'vitest';
 import {
   detectState,
+  assertAppRoleBootstrapped,
   hashSql,
   hasNoTransactionDirective,
   splitSqlStatements,
@@ -85,6 +86,23 @@ describe('autoMigrate', () => {
 
     it('should return "normal" when both users and breeze_migrations exist', () => {
       expect(detectState(true, true)).toBe('normal');
+    });
+  });
+
+  describe('assertAppRoleBootstrapped', () => {
+    it('does not throw when ensureAppRole succeeded', () => {
+      expect(() => assertAppRoleBootstrapped(true, false)).not.toThrow();
+      expect(() => assertAppRoleBootstrapped(true, true)).not.toThrow();
+    });
+
+    it('does not throw when ensureAppRole was skipped but breeze_app already exists (e.g. compose-provisioned dev DB)', () => {
+      expect(() => assertAppRoleBootstrapped(false, true)).not.toThrow();
+    });
+
+    it('throws a pointed error when ensureAppRole was skipped AND breeze_app does not exist (#4048)', () => {
+      expect(() => assertAppRoleBootstrapped(false, false)).toThrow(
+        /BREEZE_APP_DB_PASSWORD.*POSTGRES_PASSWORD.*globalPassThroughEnv/s,
+      );
     });
   });
 
--- a/apps/api/src/db/autoMigrate.ts
+++ b/apps/api/src/db/autoMigrate.ts
@@ -551,7 +551,33 @@ export async function autoMigrate(): Promise<void> {
     //        role isn't created first. Idempotent — safe on every run. We
     //        still call ensureAppRole again at step 7b so any tables created
     //        in this loop receive the privilege grants.
-    await ensureAppRole();
+    //
+    //        If ensureAppRole was skipped (no password env var reached this
+    //        process — e.g. turbo stripped it) AND breeze_app doesn't already
+    //        exist (e.g. a fresh/scratch DB that compose didn't pre-provision),
+    //        fail fast here with a pointed error instead of letting the
+    //        migration loop run ~135 files deep and die on the opaque
+    //        Postgres 42704 `role "breeze_app" does not exist` (#4048).
+    const roleEnsured = await ensureAppRole();
+    if (!roleEnsured) {
+      const roleRow = await client`
+        SELECT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'breeze_app') AS exists
+      `;
+      const breezeAppRoleExists = Boolean(roleRow[0]?.exists);
+      if (!breezeAppRoleExists) {
+        throw new Error(
+          '[auto-migrate] Refusing to proceed: the `breeze_app` role does not exist and ' +
+            'ensureAppRole() could not create it because neither BREEZE_APP_DB_PASSWORD nor ' +
+            'POSTGRES_PASSWORD is set in this process\'s environment. Set one of those two ' +
+            'variables. If you invoked this via `pnpm db:migrate` (which runs through turbo), ' +
+            'confirm the variable is listed in turbo.json\'s `globalPassThroughEnv` — turbo ' +
+            'silently strips any env var not declared there or in a task-level `env`/' +
+            '`passThroughEnv`, and a plain DATABASE_URL connection will otherwise proceed as ' +
+            'the admin/superuser and fail confusingly, many migration files later, with ' +
+            'Postgres error 42704.',
+        );
+      }
+    }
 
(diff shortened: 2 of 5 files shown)


ToddHebebrand:
Tier-1, merging — fully green (54 SUCCESS / 1 skipped).

The fix is in the right place and the error message is the valuable part: it names `globalPassThroughEnv` explicitly, which is the non-obvious step. Turbo silently stripping any var not declared there is precisely the shape that makes the original failure unreadable — a plain DATABASE_URL connection proceeds as superuser and dies ~135 migration files later on an opaque 42704.

Extracting `assertAppRoleBootstrapped` as a pure function is what makes the decision testable without a live database, and the `pg_roles` existence check correctly keeps the compose-provisioned dev DB working rather than failing everyone who already has the role.

One note, not a blocker: the guard is only reached when `!roleEnsured`, so the `roleEnsured=true` cases in the unit test are unreachable in production. Harmless as defensive coverage — worth knowing they are not exercising a live path.

PR #4058, opened 2026-08-26
docs(claude-md): fix broken pnpm vitest filter guidance (#4047)

### LINKED ISSUE #4047: `pnpm --filter <pkg> test -- --run <path>` silently runs the entire suite instead of the targeted file — CLAUDE.md documents the broken form
### Description

CLAUDE.md's "Running Tests Locally" section recommends filtered test runs via `pnpm --filter @breeze/api test`, and the broader "targeted tests while developing" guidance implies appending a path to scope a run. In practice, the `--` passthrough form does not scope anything:

```
pnpm --filter @breeze/api test -- --run <path>   # ran 1,470 files / 25,380 tests
cd apps/api && npx vitest --run <path>            # ran 1 file, as intended
```

This was independently rediscovered during the #4018 branch (PR #4041; noted by the task-3 implementer as "matches known repo trap") and is a repeat of a previously logged issue (`integration_test_dashdash_runs_whole_suite` in this project's memory) — except this instance is in the **unit** test path, not integration, and is the form CLAUDE.md itself teaches, so it isn't just an easy mistake to make, it's the documented instruction.

The mechanism: `pnpm --filter <pkg> <script> -- <args>` passes `<args>` to the underlying `package.json` script, but the `test` script in `apps/api/package.json` (and siblings) is generally something like `vitest run` or similar with `turbo` in front — depending on the exact wiring, the appended flags either get swallowed by an intermediate layer or land in a position `vitest` doesn't interpret as a path filter, and it falls back to running everything.

A sibling trap compounds this: even the correct direct form only matches by filename, not by directory prefix in the way one might expect —

```
npx vitest --run src/routes/auth/          # matches NEITHER of:
src/routes/auth.test.ts
src/routes/auth.passkeys.test.ts
```

— because vitest's path filter is a substring/glob match against file paths, and `src/routes/auth/` as a directory-shaped string doesn't match files that live at `src/routes/auth.*.ts` (sibling files, not children of a directory). This exact trap surfaced during #4018 task 5's review, which specifically had to verify both `auth.test.ts` and `auth.passkeys.test.ts` were run as siblings rather than assuming a directory-style filter covered both.

### Why it matters

CLAUDE.md's own stated efficiency guidance — "Run targeted tests while developing; the full suite ... only before PR" — does not work as written for `apps/api`. Anyone following the documented command believes they ran one file and actually ran the whole suite (slow, and also gives false confidence if they're specifically checking that unrelated tests didn't regress, since a full-suite run masks whether their targeted file was even executed vs. skipped). This has already cost review cycles on at least this branch and the sibling-file trap it names.

### Affected files

- `CLAUDE.md` — "Running Tests Locally" section, which documents `pnpm --filter @breeze/api test` without warning that appending `-- --run <path>` doesn't scope the run.
- `apps/api/package.json` — the `test` script definition that the `--` passthrough ultimately reaches.
- `turbo.json` — the `test` task (`"dependsOn": ["build"]`), which sits between `pnpm --filter` and the underlying script and may itself be part of why the passthrough doesn't reach vitest as expected.

### Suggested approach

1. **Fix the docs immediately** (cheap, no risk): amend CLAUDE.md's "Running Tests Locally" section to state the working form explicitly — `cd apps/api && npx vitest --run <path>` — and call out the sibling-file trap (`--run src/routes/auth/` matches neither `auth.test.ts` nor `auth.passkeys.test.ts`; use `--run 'src/routes/auth*'` or list both files).
2. **Investigate the script/turbo wiring** to see if `pnpm --filter <pkg> test -- --run <path>` can be made to actually work — if the fix is small (e.g. the script needs `vitest run "$@"` instead of a bare `vitest run`), fixing the plumbing is preferable to permanently working around it in docs, since the documented form is the more discoverable one for anyone not already carrying this trap in their memory.

ToddHebebrand:
## Summary

- `pnpm --filter <pkg> test -- --run <path>` silently runs the **entire** suite instead of the one targeted file. Reproduced directly: the exact documented-workaround form hung well past 2 minutes scanning `apps/api`'s full suite (1,470 files) instead of returning in seconds for one file.
- Root cause, confirmed via `NODE_OPTIONS=--require` argv logging: pnpm forwards the **literal `--` token** into the script's argv (vitest actually receives `["--", "--run", "<path>"]`). vitest's CLI parser (cac) stops parsing recognized flags once it hits that `--`, so `--run` is swallowed as a raw positional filter string instead of the flag that disables watch mode — vitest stays in watch mode and falls back to scanning the whole project.
- **Fix is docs-only.** Dropping the `--` (`pnpm --filter <pkg> test --run <path>`) works correctly — verified: 1 file, exits in seconds — and is exactly the form `ci.yml`'s `compatibility.test.ts` step already uses (line ~184). No package script or CI changes were needed or made.
- Also documents the adjacent sibling-file trap called out in the issue: vitest's path filter is a **substring match**, not a glob and not a directory prefix. `vitest run src/routes/auth/` matches only files physically inside the `auth/` directory and silently skips the sibling files `src/routes/auth.test.ts` and `src/routes/auth.passkeys.test.ts`. I also verified the issue's own suggested workaround (`--run 'src/routes/auth*'`) does **not** actually work — vitest does not glob-expand CLI filters, so that literal string matches zero files ("No test files found"). The doc instead recommends listing files explicitly or using a bare substring (`vitest run src/routes/auth`), with a caution that a bare substring can over-match (it also pulls in `authenticator.test.ts`).

## Why docs-only (no script/turbo rewiring)

Investigated whether the package `test` scripts should be rewired so the documented `-- --run` form works as-is. It can't be made to work that way: the breakage is inherent to how pnpm forwards `--` combined with how vitest's CLI parser treats it — even rewriting the script to `"vitest run"` doesn't help, since `vitest run -- <path>` reproduces the same hang (verified). The only real fix is to stop putting `--` in the invocation, which is already possible today with zero code changes, and is already what CI does. Rewiring scripts would add churn without fixing anything additional.

## Test plan

- [x] Reproduced the broken form (`pnpm --filter @breeze/api test -- --run src/routes/auth.test.ts`) — confirmed it does not scope and hangs past 2 minutes.
- [x] Confirmed the fixed form (`pnpm --filter @breeze/api test --run src/routes/auth.test.ts` and `cd apps/api && npx vitest run src/routes/auth.test.ts`) — both scope correctly to 1 file / 83 tests in seconds.
- [x] Confirmed the sibling-file trap (`vitest run src/routes/auth/` → 15 files, misses both `auth.test.ts` and `auth.passkeys.test.ts`) and that the issue's suggested glob workaround does not work (`vitest run 'src/routes/auth*'` → 0 files).
- [x] Confirmed via `gh pr diff 4041 --name-only` that PR #4041 does not touch `CLAUDE.md`, so no overlap.
- [x] Doc-only change — no `package.json` or `ci.yml` changes were made.

Closes #4047


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, comment-analyzer)
**Findings:** 1 raised (independently, by both agents) → fixed across 2 follow-up commits; 0 outstanding.
- Both agents flagged the same overclaim: the original text said "every workspace package's `test` script is a bare `\"vitest\"\"`, which is false — 6 of 17 packages (`apps/portal`, `apps/helper`, `apps/mobile`, `packages/extension-sdk`, `packages/extension-web-sdk`, `packages/extension-testkit`) already use `"test": "vitest run"`.
- Verified independently and fixed in a first pass, then code-reviewer's own repro against `apps/portal` (`vitest run` script) surfaced that my first fix *understated* the impact — I'd claimed the `--` trap was "less harmful" on `vitest run` packages, but direct repro showed it still runs the FULL suite there too (26 files instead of 1), it just exits afterward instead of hanging in watch mode. Corrected in a second follow-up commit and re-verified against both `apps/api` (bare `vitest`) and `apps/portal` (`vitest run`) — the fix (dropping the `--`) scopes correctly on both.
**Tests:** N/A — docs-only change (CLAUDE.md), no source/test/script files touched. Confirmed via `gh pr diff 4058 --repo LanternOps/breeze --name-only` that only `CLAUDE.md` changed, and via `gh pr diff 4041 --repo LanternOps/breeze --name-only` that PR #4041 does not touch it (no overlap). `apps/api/package.json`, `turbo.json`, and `.github/workflows/ci.yml` were inspected but not modified — CI already uses the correct form (`ci.yml:184`), so no script rewiring was needed.

ToddHebebrand:
Tier-1, merging. Verified the load-bearing claim rather than the prose: `ci.yml:184` really is `pnpm --filter=@breeze/api test --run src/extensions/compatibility.test.ts`, so the recommended form is the one CI already proves works. Both traps match what has actually bitten us — the `--` passthrough and the substring-not-glob sibling skip — and the repro evidence (vitest receiving `["--", "--run", "<path>"]` via argv logging, 1,470 files vs 1) is what makes this durable rather than folklore.

Issue #2183, opened 2026-07-03
Partner-level SSO + login page branding for self-hosted MSP deployments


win-wxx:
### Problem

We run a self-hosted Breeze deployment (v0.88.0, Docker Compose, on-prem)

Current SSO (/api/v1/sso/*) is fully org-scoped: every provider requires an org_id, defaultRoleId must resolve to an org-scoped role, and /sso/check/:orgId needs an orgId just to know whether to show a login button. This works for end-customers logging into their own org, but there's no way for our own internal team (20 technicians) to SSO into the Partner-level dashboard with orgAccess: all — which is what we actually need day to day.

We noticed partners.ssoConfig (jsonb) is documented as a field but has no endpoint, no login flow, and isn't referenced anywhere in the SSO reference doc — looks like a placeholder rather than something wired up end-to-end.

Related: because SSO buttons only resolve per-org, the top-level login page has no way to show partner branding or a partner SSO entry point before the user's org is known. It also hardcodes generic marketing copy ("10,000+ endpoints" etc.), which reads oddly on a single-partner self-hosted instance.

### Proposed Solution

A partner-scoped SSO provider + login entry point (e.g. /sso/partner-login/:partnerId) that accepts partner-scoped roles (Partner Admin/Technician/etc.) as the default role instead of requiring an org-scoped role.

Same safety model as org SSO: testing > active status, break-glass password login preserved unless explicitly enforced.

A basic partner_branding config (logo, accent color, headline override, optional /login/:partnerSlug) so the login page can reflect the partner and surface the partner SSO button.

### Component

Web Dashboard

### Scope

Medium (new feature within existing module)

ToddHebebrand:
**Acknowledged — and your read of the code is right.** `partners.ssoConfig` exists in the schema (`apps/api/src/db/schema/orgs.ts`) and is persisted through the partner routes, but nothing reads it back for authentication — there is no partner login flow and no `/sso/partner-*` endpoint surface. It is storage-only today; all live SSO is org-scoped exactly as you described, including `defaultRoleId` requiring an org-scoped role and `/sso/check/:orgId` needing an org up front.

[...]

PR #4071, opened 2026-08-26
feat(api,web): lockout preflight + confirm-through before Enforce SSO goes live (#4068)

### LINKED ISSUE #4068: [Auth][SSO] Enforce SSO toggle needs a lockout preflight — show which active users have no linked identity before enabling
## Description

Enabling **Enforce SSO** on a provider immediately blocks password login for every user on the tenant axis (`assertPasswordAuthAllowedBySso`). Any active user without a `user_sso_identities` link for that provider is locked out entirely: SSO bounces them with `sso_link_required` (password-holding accounts are never auto-linked, per the #2183 guard) and the password road is gone.

The form's amber warning today only says passwords stop working — it does not say *unlinked users lose all access with no self-recovery*. This has produced two full-tenant lockouts on hosted (2026-08-17 and 2026-08-26, same partner: admin enabled enforcement after linking only themselves; the other 4 staff were locked out; remediation both times was a manual prod UPDATE).

## Proposed Fix

Preflight at the point of enabling:

- **API**: when a create/update sets `enforceSSO: true` (and it wasn't already true), compute the affected population — active users on the provider's axis (partner: `users.partner_id = provider.partner_id AND org_id IS NULL`; org: members of `provider.org_id`) with no `user_sso_identities` row for this provider — and return it (count + emails) in the response or via a dry-run/preflight endpoint.
- **UI** (`SsoProviderForm.tsx` / `SsoProvidersPage.tsx`): before saving with enforcement newly on, show the list: *"3 of 5 users have not connected SSO and will be unable to sign in: a@…, b@…, c@…"* and require explicit confirmation (type-to-confirm or a distinct checked acknowledgement). Zero unlinked users → save proceeds with the existing warning.
- Decide: hard-block vs confirm-through. Recommend **confirm-through** (an MSP may intentionally cut off a departed user), but the default answer is No.
- The admin's own account should be called out separately if unlinked — that's the guaranteed-total-lockout case.

Complementary to #4067 (link-on-first-SSO-login), which removes the lockout mechanism itself; this preflight is the cheap near-term guard and remains useful afterward as rollout visibility.

## Affected Files

- `apps/api/src/routes/sso.ts` (provider create/patch, or a new `/providers/:id/enforcement-preflight`)
- `apps/web/src/components/settings/SsoProviderForm.tsx`, `SsoProvidersPage.tsx`
- Tests: `apps/api/src/routes/sso.test.ts` (axis-correct population), web form tests

## Reported By

Internal, from the 08-17 / 08-26 partner lockouts. Related: #4067, #3229, #2183.

ToddHebebrand:
Closes #4068

## Problem

Enabling **Enforce SSO** blocks password login for every user on the provider's axis (`assertPasswordAuthAllowedBySso`), and any active user without a linked identity bounces off SSO with `sso_link_required` (password accounts are never auto-linked, per the #2183 guard) — total lockout with no self-recovery. Two full-tenant lockouts on hosted (2026-08-17 and 2026-08-26) were remediated with manual prod UPDATEs. The form's amber warning never said *who* would lose access.

### API

**`POST /sso/providers/enforcement-preflight`** — computes the lockout population before the admin commits:

- **Membership**: active members of the provider's axis via `organization_users` / `partner_users` — **partner membership wins the axis** (mirrors `resolveCurrentUserTokenContext`), so an org-axis preflight excludes users who also hold a partner membership.
- **Linkage**: a user is safe only if linked to the **effective login provider** — the oldest `(createdAt, id)` provider that will be active once the change is live, because the pre-auth login entries (`GET /sso/login/:orgId`, `/sso/login/partner/:partnerId`) route through exactly that one. Links to a newer active provider are unusable at login and do not count. (This deliberately deviates from the issue's per-target-provider sketch — that shape both over- and under-reports.)
- A **SAML** effective provider counts as *no path* (the login entry 400s on it rather than falling through).
- Passkey holders keep a sign-in path (passkey routes never consult the SSO gate) — annotated `hasPasskey`, not excluded; disabled passkeys don't count.
- Same auth surface as the writes it fronts: `sso:admin` + MFA + `canWriteProviderRow` / partner-wide gates. Population read runs in a system DB context (identities/passkeys are user-id-scoped under RLS — an admin context would read 0 rows and report everyone unlinked).
- Response: exact counts, capped list (200) with self sorted first, `selfLockedOut`, `loginProvider`, `truncated`.

**Server-side confirm-through** on the two transitions that actually make enforcement live:

1. `PATCH /providers/:id` setting `enforceSSO: true` on an **active** provider
2. `POST /providers/:id/status` activating an **already-enforcing** provider

Either returns **409** `sso_enforcement_lockout_confirmation_required` *with the preflight payload* unless the body carries `acknowledgeLockout: true` — raw API callers get the same guard as the UI. Zero-lockout transitions proceed unchanged, so existing automation only breaks where it was about to lock a tenant out. `acknowledgeLockout` is stripped before the update spread (the `preset`-omit trap). Both refusals are audited.

### Web

`SsoProvidersPage` preflights on save-with-enforcement-newly-on and on activate-of-an-enforcing-provider, then shows a type-checked acknowledgement dialog listing affected users (self called out separately — the guaranteed self-lockout case, per the issue) and resends with `acknowledgeLockout`. Rendered as an overlay above the form modal so form state survives cancel. A failed preflight read falls through to the save — the server 409 is the backstop, so a broken preflight can neither skip the protection nor wedge saving. New `enforceLockout*` strings in all 8 locales.

## Design notes

- **Confirm-through over hard-block** (issue's recommendation): an MSP may intentionally cut off a departed user; default answer is No (checkbox-gated).
- Create flow shows the warning too, but the server guard sits on the *live* transitions only — providers are born `inactive`, so create can't lock anyone out; activation is guarded.
- Design cross-checked with an independent codex consult (xhigh): it independently flagged the oldest-active-provider login selection and the status-activation path, both incorporated.

## Tests

- `sso.test.ts` (+14): route contract — authority checks (403/404, partner-wide gate), 409/ack/zero-lockout matrix on PATCH and status, `acknowledgeLockout` never reaches `set()`, SAML/no-provider shapes.
- `ssoEnforcementPreflight.integration.test.ts` (new, real Postgres): axis membership, partner-wins exclusion, linked-to-newer-provider still locked out, invited/disabled excluded, inactive target as future login provider, SAML, passkey/disabled-passkey annotation.
- `SsoProvidersPage.test.tsx` (+4): confirm flow gating, straight-through on zero, no preflight when enforcement already on, activate path.
- `tsc --noEmit` clean (api, web); i18n parity/key-usage/no-silent-mutations suites green.

### DIFF
--- a/apps/api/src/__tests__/integration/ssoEnforcementPreflight.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ssoEnforcementPreflight.integration.test.ts
@@ -77,12 +77,30 @@ async function linkIdentity(userId: string, providerId: string) {
 }
 
 describe('enforce-SSO lockout preflight population (#4068)', () => {
-  it('org axis: effective-provider linkage, partner-membership exclusion, status filtering', async () => {
+  it('org axis: effective-provider linkage, partner-membership exclusion, status filtering, tenant isolation', async () => {
     const db = getTestDb();
     const partner = await createPartner();
     const org = await createOrganization({ partnerId: partner.id });
     const orgRole = await createRole({ scope: 'organization', orgId: org.id, partnerId: partner.id });
     const partnerRole = await createRole({ scope: 'partner', partnerId: partner.id });
+
+    // Tenant-isolation discriminator: a SIBLING org under the same partner
+    // with its own active unlinked member and its own provider. If any of the
+    // org_id filters were dropped, this user/provider would leak into the
+    // result — this is a system-context read that returns email addresses.
+    const otherOrg = await createOrganization({ partnerId: partner.id });
+    const otherOrgRole = await createRole({ scope: 'organization', orgId: otherOrg.id, partnerId: partner.id });
+    const outsider = await createUser({
+      partnerId: partner.id,
+      orgId: otherOrg.id,
+      email: `outsider-${uniq()}@example.com`,
+      name: 'outsider',
+      status: 'active'
+    });
+    await assignUserToOrganization(outsider.id, otherOrg.id, otherOrgRole.id);
+    // Older than every provider in the org under test: if the provider-axis
+    // filter leaked, THIS would be picked as the effective login provider.
+    await seedProvider({ orgId: otherOrg.id, status: 'active', createdAt: new Date(Date.now() - 120_000) });
 
     // Two ACTIVE providers: pOld is the one the login entry actually uses.
     const pOld = await seedProvider({ orgId: org.id, status: 'active', createdAt: new Date(Date.now() - 60_000) });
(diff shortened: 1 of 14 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter, pr-test-analyzer) + an independent codex design consult (xhigh) pre-implementation.

**Findings:** 2 HIGH (both web mutation paths mishandled the server 409 — toggle path swallowed the payload, save path toasted API-caller instructions), 2 Important (acknowledged lockouts were unauditable; no test discriminated any tenant filter in the population SQL), plus create-flow false-lockout warning fatigue, 401/MFA gate coverage on the email-disclosing preflight route, truncation, cancel-path, and self-sort vacuity findings — **all addressed in f12e12af3; 0 outstanding.**

**Deliberately not acted on** (noted, fail-safe direction or out of scope): passwordless auto-link over-report; the `allowedDomains`-removal under-report (a linked user whose email domain was later removed from the allowlist is locked out but reported safe — narrow, pre-existing gate ordering in the callback); no loading affordance on the list's Activate button.

**Tests:** `sso.test.ts` 158 passed; `ssoEnforcementPreflight.integration.test.ts` 4 passed against real Postgres (tenant filters now discriminating — sibling-org and sibling-partner fixtures); `SsoProvidersPage.test.tsx` 17 passed; i18n parity/key-usage/no-silent-mutations suites green (255); `tsc --noEmit` clean on api and web.

ToddHebebrand:
**Review run (round 2):** /pr-review-toolkit:review-pr code-reviewer, focused on the round-1 fix commit f12e12af3.
**Findings:** 1 raised (confidence 82: the client lockout gate lacked a status term, so editing an *inactive* provider with Enforce SSO ticked still fired the false lockout warning — same defect class the commit fixed for create) → fixed in 5f02e28e1; everything else verified clean, including the reworked guard control flow, the acknowledged-path audit stamp, DB-context safety of the extra preflight query on both routes, the org RLS probe (one intentional tightening: a partner token with an out-of-selected-list focused org now 403s instead of receiving emails), and the web 409 interception (no loop, no state leak, no double body read).
**Tests:** SsoProvidersPage.test.tsx 18 passed (new inactive-edit no-warn test); web `tsc --noEmit` clean.

Issue #1104, opened 2026-06-05
[agent] Watchdog heartbeat response omits agent upgradeTo, so a live watchdog can't recover a dead main agent


bdunncompany:
## Summary
A live watchdog has no way to recover a dead/wedged main agent, because the watchdog heartbeat response never carries the agent upgrade directive.

## Mechanism
- The watchdog branch of `POST /api/v1/agents/:id/heartbeat` (`role=watchdog`) returns `commands` and `watchdogUpgradeTo`, but **never an agent `upgradeTo`** (`apps/api/src/routes/agents/heartbeat.ts`) — even when the device's recorded `agentVersion` is behind latest.
- The watchdog's `processHeartbeatResponse()` already handles `resp.UpgradeTo` → `doUpdateAgent()`, so the recovery mechanism exists; it is simply never populated for the watchdog role.
- The server already detects the failure condition via the `#800` main-agent-silent asymmetry detector (watchdog keeps reporting while the main agent is silent), but nothing acts on it.

## Impact
When a main agent is wedged (process alive but not heartbeating), there is no server-side path to recover it: the watchdog can update *itself* but is never told to update/restart the *agent*. Recovery currently requires out-of-band access to each host.

## Suggested fix
When the watchdog heartbeats and the device's `agentVersion` is behind latest (same `compareAgentVersions` logic already used for `watchdogUpgradeTo`), include the agent `upgradeTo` in the response so the watchdog's existing `doUpdateAgent()` path can recover the agent.

ToddHebebrand:
**Confirmed — the recovery gap is real and the suggested fix is sound.**

**The asymmetry is in the heartbeat response.** The watchdog branch of `heartbeat.ts` (75-201) returns `commands` and `watchdogUpgradeTo` only; the main-agent branch (527-543) additionally returns `upgradeTo`. The watchdog branch computes `watchdogUpgradeTo` for the watchdog itself but never computes/returns an agent `upgradeTo`, even when the device's recorded `agentVersion` is behind latest.

**The recovery machinery already exists on the watchdog side.** `breeze-watchdog/main.go:600-605` branches on `resp.UpgradeTo` and calls `doUpdateAgent(...)` (`:723-748`). The field is wired through `HeartbeatResponse` (`failover.go`) — it's simply never populated by the server for the watchdog role.

[...]

PR #4091, opened 2026-08-26
fix(agent,api): self-host→hosted update transition + edition-aware offer gating (#4072)

### LINKED ISSUE #4072: Agents on 0.105.0-0.106.x can never auto-update to hosted-edition artifacts — permanent 60s retry loop, no transition path
## Summary

An agent built as the **self-host** edition that also carries the manifest edition check can never auto-update to a **hosted**-edition artifact. It rejects the download, stays at `status='updating'`, and retries roughly every 60 seconds indefinitely. There is no transition path and no server-side escape hatch.

The affected band is narrow and non-obvious: **0.105.0 – 0.106.x**. Agents both *older* and *newer* update normally.

## Observed failure

On a device running 0.105.1, offered 0.108.0:

```
warn   user-helper download failed; proceeding with agent-only upgrade
         error: download info request failed with status 404
         targetVersion: 0.108.0   currentVersion: 0.105.1

error  failed to update
         error: failed to download binary:
                update rejected: artifact edition "hosted" does not match this build
         targetVersion: 0.108.0
```

The device keeps heartbeating normally throughout. It never converges and never gives up — the loop repeats every heartbeat until the offer is withdrawn.

## Why this band specifically

The edition check was introduced in **v0.105.0** (#3349), at `agent/internal/updater/updater.go:628` and `:1214`. Combined with which edition a given release's binaries were built as, that yields three bands:

| Installed agent | Has edition check | Built as | Hosted update |
|---|---|---|---|
| `< 0.105.0` | no | self-host | works — no check to fail |
| **`0.105.0` – `0.106.x`** | **yes** | **self-host** | **permanently refused** |
| `>= 0.107.0` | yes | hosted | works |

## Why this is easy to miss

A staged canary does not catch it unless it deliberately samples the middle band. We ran three canary waves against 0.108.0:

1. agents on `0.107.0` → upgraded cleanly
2. an agent on `0.104.0` → upgraded cleanly
3. an agent on `0.105.1` → **wedged**

Waves 1 and 2 were green and looked like sufficient evidence to promote. Because the working versions bracket the broken one on *both* sides, "test an old one and a new one" is actively misleading here.

## Impact

Any deployment that has cut over to serving hosted-edition artifacts strands **every enrolled agent in the 0.105.0–0.106.x band simultaneously** the moment the new version is promoted. Those agents:

- never reach the new version,
- sit permanently in `updating`,
- generate a download attempt every ~60s each, indefinitely.

Pinning does **not** rescue them. The version pin is resolved and offered correctly server-side; the refusal happens agent-side after download begins, so no pin value produces a working outcome.

Recovery today is a manual reinstall from a hosted-edition installer, per device.

## Suggested direction

Options, roughly in increasing order of invasiveness:

1. **Allow a one-way self-host → hosted transition** in the updater when the manifest signature verifies under the embedded official key, treating edition as informational for that specific transition rather than a hard gate.
2. **Server-side: withhold the offer** from agents whose build edition cannot accept it, so they idle quietly on their current version instead of hot-looping. This contains the damage but does not fix the upgrade.
3. **Ship a targeted remediation build** for the band.

(1) is the only option that actually unsticks the fleet without touching every device by hand. (2) is worth doing regardless — the indefinite 60s retry with no backoff and no give-up is its own defect.

## Also worth fixing alongside

The first warning in the log excerpt is a separate issue: `user-helper download failed … status 404`. The hosted artifact set registers no `user-helper` (or `helper`) component rows, so that lookup 404s on every hosted upgrade. The agent correctly proceeds without it, so it is not the cause of this bug — but it is noise on every single hosted update and hides real failures.

## Environment

- Reproduced on Windows `amd64`, agent `0.105.1`, target `0.108.0`.
- Reproduced against two independent deployments/regions.
- Reverting the offer returns the device to `online` within ~2 minutes with no lasting damage.

ToddHebebrand:
Closes #4072

## Problem

An agent built as the **self-host** edition that also carries the manifest edition check (field band **0.105.0–0.106.x**) can never auto-update to a **hosted**-edition artifact. `updater.editionAllowed` rejects the artifact *after* download, the device wedges at `status='updating'`, and retries every ~60s indefinitely. Pinning doesn't help (the refusal is agent-side), and agents on both sides of the band update fine, so a two-sample canary reads green.

## What this changes

**Agent (`agent/internal/…`)**
- `updater/edition.go` — one-way **self-host → hosted** acceptance. Both call sites run strictly after Ed25519 manifest verification against the trust roots, and a hosted build only *adds* enforcement (the hostpolicy allowlist), so the transition grants nothing an attacker could not already get from a signed self-host manifest. The reverse (hosted build accepting a self-host artifact) stays a hard refusal — that direction would strip the allowlist. The cross-edition acceptance is logged.
- `heartbeat.go` `migrationSignal` — self-host builds now report `agentEdition:"self-host"` (previously silent). Because the reporting and the allowance ship in the same binary, a reported edition — either value — doubles as the server's "this build can accept hosted artifacts" capability signal. The zod schema already accepted both values.
- `watchdog/failover.go` — the failover heartbeat reports `agentEdition` too (the watchdog is the binary that downloads during failover).

**API (`apps/api/src/routes/agents/…`)**
- `helpers.ts` — `resolvePinnedUpgradeTarget` (both queries) and the heartbeat helper-latest query are now **edition-scoped** (`eq(agentVersions.edition, getBinaryEdition())`), matching the download/register/promote paths; previously a row of the other edition could be resolved and offered.
- `helpers.ts` — new `agentAcceptsServedEdition` predicate: reported edition ⇒ capable; silent + core version `< 0.105.0` ⇒ predates the check; silent + `≥ 0.105.0` ⇒ would refuse hosted (**fail closed**, incl. `0.105.0-rc.*` and unparseable versions). Serving self-host: only a `'hosted'` reporter is refused (symmetric protection for hosted builds pointed at a self-host server).
- `heartbeat.ts` — every offer path (agent / helper / watchdog on the main branch, bootstrap included; watchdog self-update + agent recovery on the failover branch) consults the predicate with **this beat's payload** and withholds offers the build would refuse, with a once-per-device deduped warn. Wedged devices drop back to `online` on the next beat after the offer disappears (verified in the field by reverting the pin: ~2 min).

## What this does NOT do

- **It does not recover the already-stranded band by itself.** Those 35 devices run old code and can't receive this fix OTA; this PR stops the wedge loop (they idle at their current version) and gives every *future* self-host build a transition path. Recovery for the existing band = per-device reinstall with a hosted installer, or a one-off bridge release decision (discussed in the issue).
- The user-helper 404 noise on hosted upgrades (hosted lane registers no `user-helper` rows) is a separate defect and untouched.
- AI/manual one-shot update commands (`aiToolsAgentMgmt`, WS) don't loop and are not gated here — follow-up candidate to share the same predicate.

## Accepted risk (documented in `edition.go`)

A malicious/compromised control plane could substitute a signed strict-hostpolicy hosted artifact and cut a self-hoster's agents off from their own server. Such a control plane can already run arbitrary scripts on enrolled devices, so no new attacker capability is added; server-side, offers are edition-scoped to the deployment's own edition, so a healthy self-host server never offers one. An independent design review (codex, xhigh-adjacent) confirmed no signature bypass and this residual risk.

## Tests

- Agent: full `go test -race ./...` green. Flipped edition-matrix tests (3 sites), new wire-contract tests for `agentEdition` (main heartbeat + watchdog failover).
- API: `helpers.agentUpdatePolicy.test.ts` — predicate matrix (12 cases incl. prerelease + fail-closed) and PgDialect compiled-SQL assertions pinning the edition filter params in both resolver queries. `heartbeat.test.ts` — 8 wiring tests: withhold per offer path (bootstrap included), payload-not-column sourcing, failover-branch gating on the watchdog's own payload, once-per-device warn dedupe.
- `tsc --noEmit` clean; all 47 `src/routes/agents/` test files (847 tests) + `aiToolsAgentMgmt.test.ts` green.

https://claude.ai/code/session_01Thjb1KChG8kqpt2fUFT71J

### DIFF
--- a/agent/internal/heartbeat/heartbeat.go
+++ b/agent/internal/heartbeat/heartbeat.go
@@ -137,9 +137,11 @@ type HeartbeatPayload struct {
 	// written unconditionally server-side (the outboundNetworkPolicyVersion
 	// self-healing pattern, NOT the sticky isVirtual pattern) so a resolved
 	// condition clears a dashboard migration banner on the next beat.
-	// omitempty on both: a self-host build (the only build in this repo
-	// today) reports "" / false — both omitempty fields drop out — so the
-	// wire payload is byte-identical to pre-Task-8 agents.
+	// Since #4072 BOTH build editions report a value ("self-host"/"hosted") —
+	// a reported edition doubles as the server's signal that this build can
+	// accept hosted-edition update artifacts (see migrationSignal). omitempty
+	// is retained so MigrationRequired=false stays off the wire; AgentEdition
+	// is never empty from this build.
 	AgentEdition      string `json:"agentEdition,omitempty"`
 	MigrationRequired bool   `json:"migrationRequired,omitempty"`
 }
@@ -150,12 +152,19 @@ type HeartbeatPayload struct {
 // nothing is persisted to violate the allowlist when there is no backup.
 // Pure; independent of hostpolicy.Strict() — reporting is telemetry, not
 // enforcement, so it fires the same in gap and strict hosted builds.
-// Self-host returns ("", false) — NOT "self-host" — because the
-// AgentEdition field is omitempty: only the empty string drops out of the
-// wire payload, preserving byte-identity with pre-Task-8 agents.
+//
+// Self-host builds report "self-host" explicitly (#4072). This is the
+// server's edition-transition capability signal: a build that reports an
+// edition — either value — also carries the one-way self-host → hosted
+// allowance in updater.editionAllowed (both shipped in the same binary),
+// while a silent build ≥0.105.0 is a self-host build that hard-refuses
+// hosted-edition artifacts, so the server withholds those offers rather
+// than wedging it in a permanent retry loop. Regressing this to the empty
+// string would re-strand every future self-host agent that migrates to a
+// hosted control plane.
 func migrationSignal(server, backup string) (edition string, migrationRequired bool) {
 	if !hostpolicy.Enforced() {
-		return "", false
+		return "self-host", false
 	}
 	if hostpolicy.AllowedURL(server) != nil {
 		return "hosted", true
--- a/agent/internal/heartbeat/migration_signal_test.go
+++ b/agent/internal/heartbeat/migration_signal_test.go
@@ -7,11 +7,13 @@ import (
 )
 
 func TestMigrationSignal(t *testing.T) {
-	// Self-host build: empty edition (omitempty drops it from the wire
-	// payload, keeping byte-identity with pre-Task-8 agents), never
-	// migration-needed.
-	if e, m := migrationSignal("https://anything.example", ""); e != "" || m {
-		t.Fatalf("self-host: want (\"\",false), got (%s,%v)", e, m)
+	// Self-host build: reports its edition explicitly (#4072 — the server
+	// withholds hosted-edition update offers from agents that never report an
+	// edition, because a silent self-host build ≥0.105.0 would hard-refuse
+	// them; reporting "self-host" marks this build as transition-capable),
+	// never migration-needed.
+	if e, m := migrationSignal("https://anything.example", ""); e != "self-host" || m {
+		t.Fatalf("self-host: want (\"self-host\",false), got (%s,%v)", e, m)
 	}
 
 	restore := hostpolicy.SetAllowedHostsForTest("hosted-a.example")
--- a/agent/internal/heartbeat/migration_signal_wire_test.go
+++ b/agent/internal/heartbeat/migration_signal_wire_test.go
@@ -8,17 +8,19 @@ import (
 	"github.com/breeze-rmm/agent/internal/hostpolicy"
 )
 
-// TestHeartbeatPayloadWireByteIdentity pins the omitempty contract on the
+// TestHeartbeatPayloadWireSelfHostSignal pins the wire contract on the
 // build-edition telemetry fields at the JSON layer, which is the only layer
 // where it actually matters.
 //
-// TestMigrationSignal asserts migrationSignal() returns ("", false) for a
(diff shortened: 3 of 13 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer, pr-test-analyzer, silent-failure-hunter in parallel on the full diff, plus a focused code-reviewer pass on the fix commit; independent codex design review (read-only, high) on the security direction before implementation.
**Findings:** 1 critical (deployed watchdogs predate edition reporting → failover gate would have withheld the #1104 recovery path from the whole hosted fleet; fixed with the stored-`agentEdition` fallback in 45a87793c), 7 observability findings (recovery-withhold trace, Sentry parity, Info→Warn shipper level, cause-accurate messages, dedupe re-arm/keying) → fixed across 45a87793c + abae4eb8c; 2 probe-verified test gaps (unscoped helper-latest query, zod passthrough) → closed; 1 cross-surface finding deferred to #4093 (manual/AI dispatch bypasses the gate — one-shot, no wedge loop). 0 outstanding.
**Tests:** agent `go test -race ./...` full suite green; API: 47 files / 856 tests under `src/routes/agents/` + `aiToolsAgentMgmt.test.ts` green; `tsc --noEmit` clean. Test-analyzer ran revert-probes: stripping the five offer gates reddens exactly the four withhold tests — no vacuous coverage.
**Status:** review-clean, awaiting maintainer merge. Note for rollout: this PR stops the wedge loop and opens the transition path for future self-host builds; the already-stranded 0.105.0–0.106.x devices (35, 11 online) cannot receive this code OTA — they need per-device reinstall with a hosted installer, or a one-off bridge-release decision.

Issue #4085, opened 2026-08-26
W09: Wave 3.5c — durable event dispatch: BullMQ ingress + independent per-subscriber delivery jobs, flag-gated and proven inside the existing all-in-one process


ToddHebebrand:
Wave 3.5c of #3821. Split out of #3825 on 2026-08-26 — see the plan doc
`docs/superpowers/plans/ai-mcp/2026-08-26-ai-agents-wave3.5a-idempotency.md`
(landed in #4083).

## The problem

All event delivery today is in-process. `EventBus.publish()` writes a Redis
stream plus two pub/sub channels, then calls `invokeLocalHandlers`, which awaits
each handler and **swallows every failure** (logs `EVENT_BUS_LOCAL_HANDLER_FAILED`,
never rethrows). Subscribers therefore only ever see events published by their
own process.

That is wave 3d's written-down accepted risk: a trigger can drop silently
between `createAlert` and the BullMQ enqueue.

## Decision already made — do NOT relitigate

**Do not activate `startConsuming()`** (`apps/api/src/services/eventBus.ts:442`).
Two independent reviewers (Claude + codex `xhigh`) reached this separately on
2026-08-26. It has no caller and is not a nearly-finished implementation:

- creates consumer groups at id `0`, replaying up to ~10k retained events per org
  on first enablement (`eventBus.ts:173,448`);
- builds `XREADGROUP` multi-stream arguments in the wrong keys-then-IDs order
  (`eventBus.ts:471`);
- reads only `>` and never reclaims abandoned pending entries (`eventBus.ts:474`);
- **ACKs even when a handler threw**, then writes a monolithic DLQ record
  (`eventBus.ts:541`);
- snapshots org ids once at startup, so new orgs are never consumed.

## The shape to build instead

BullMQ ingress plus **independent per-subscriber delivery jobs**:

- the API publishes one versioned `route-event` job, id `event-route-<eventId>`;
- the worker routes it into separate `deliver-event` jobs, e.g.
  `event-handler-automation-<eventId>`;
- each subscriber gets a stable id, its own retry policy, metrics, DLQ, and a
  durable receipt keyed `(event_id, subscriber_id)`.

**Never retry the whole fan-out because one handler failed** — that replays
already-successful webhooks, notifications and automations.

Keep live WebSocket pub/sub separate, and `publishUserEvent()` must stay private
and continue bypassing global subscribers (`eventBus.ts:323`).

[...]

PR #4098, opened 2026-08-26
fix(api): recover webhook deliveries recorded but never enqueued (#4095)

### LINKED ISSUE #4095: [API] Webhook deliveries recorded but never enqueued have no recovery path, and the dedupe skip is silent
Found during the code review of #4083 (AI agents wave 3.5a). Latent today; becomes
a permanent silent drop once wave 3.5c (#4085) makes event delivery at-least-once.

## The gap

#4083 added a unique index on `webhook_deliveries (webhook_id, event_id)` and made
the `'*'` subscriber skip queueing when the delivery row already exists:

```ts
// apps/api/src/workers/webhookDelivery.ts
if (createDeliveryRecord && deliveryId === null) continue;
```

The delivery row is committed to Postgres **before** `queueDelivery`, which is a bare
`redis.lpush` onto a non-durable list (`webhookDelivery.ts:262-281`). Two states leave
a row recorded with no job ever enqueued:

1. the LPUSH throws (Redis down), or the process dies between the commit and the LPUSH;
2. the worker `BRPOP`s the job and dies before delivering.

Both leave `status = 'pending'` with no job. **There is no reaper.** I grepped: nothing
re-queues `webhook_deliveries` rows stuck in `pending`. `routes/webhooks.ts:764` exposes
only a manual per-delivery retry, and only for `status === 'failed'`.

Today that costs one lost attempt. Once redelivery exists, the redelivered event hits
the unique index, `createDeliveryRecord` returns null, the subscriber skips — and that
`(webhook, event)` is suppressed **forever**. The customer's endpoint is never called,
and nothing logs it.

Also note the `try/catch` sits **outside** the `for (const webhook of webhooks)` loop
(`webhookDelivery.ts:474`), so one webhook's failure aborts delivery for webhooks
N+1… of that event. Combined with the new dedupe, the retry then skips the webhooks
that already got a row while the rest proceed — so the failing webhook is the one
permanently dropped.

## Suggested fix

1. **Make the skip observable.** It currently emits nothing, so in production you
   cannot distinguish "dedupe working as designed" from "dedupe eating deliveries":

   ```ts
   console.log('[WebhookDelivery] duplicate-delivery-skipped', JSON.stringify({
     errorId: 'WEBHOOK_DELIVERY_DUPLICATE_SKIPPED',
     webhookId: webhook.id, orgId: event.orgId, eventId: event.id,
     eventType: event.type, existingStatus,
   }));
   ```

2. **Recover orphans.** Have `createDeliveryRecord` return the existing row's status,
   and re-queue rather than skip when it is `pending` and older than a threshold —
   or add a sweep for stale `pending` rows.

3. **Move the `try/catch` inside the loop** so one webhook cannot deny delivery to the
   others.

## Gate

Worth doing **before** #4085 flips delivery to at-least-once — that is what turns this
from "one lost attempt" into "permanently suppressed".

ToddHebebrand:
Closes #4095

## The gap

`webhook_deliveries` rows are committed to Postgres **before** `queueDelivery`, which is a bare `redis.lpush` onto a non-durable list. Three states leave a row recorded with no job:

1. the LPUSH throws (Redis down), or the process dies between the commit and the LPUSH;
2. the worker `BRPOP`s the job and dies before delivering;
3. one webhook's failure aborts the fan-out for webhooks N+1… of the same event (the `try/catch` sat **outside** the loop).

All leave `status = 'pending'` with no job. **There was no reaper** — the manual retry endpoint only accepts `status = 'failed'`. And the dedupe skip added by #4083 emitted *nothing*, so you could not tell "dedupe working as designed" from "dedupe eating deliveries".

Latent today. Once #4085 makes delivery at-least-once, the redelivered event hits the `(webhook_id, event_id)` unique index, the subscriber skips — and that pair is suppressed **forever**, silently.

**No #4085 scope is implemented here**: no BullMQ ingress for webhook delivery, no `startConsuming()` activation.

## What changed

**1. A recovery sweep** (`jobs/webhookDeliveryRecovery.ts`) — every 5 min, batch-bounded at 200, backed by a **partial** index over the unresolved statuses only. Each row is taken with a CAS that re-asserts the *whole* candidate scope (not just the id), so N API instances sweep and exactly one drives a given row.

**2. An execution claim on the delivery worker** — the part that makes recovery *safe*. The queue is a plain Redis list with no job identity, so the sweep cannot distinguish a lost job from a merely **backlogged** one and will re-queue both. `BRPOP`'s 5s is only its empty-queue wait; the FIFO backlog ahead of a job is unbounded, so **no stale threshold alone is safe**. Without a claim this fix would have traded a missed POST for a **duplicate POST to the customer's endpoint**. Now one popped copy flips `pending` → `retrying` and delivers; every other copy loses and is dropped (loudly). It also bounds the sweep, since a claimed row leaves `pending` immediately.

Retries (`attempts > 0`) skip the claim — their row already left `pending`, so claiming would kill the backoff ladder. DLQ replays skip it too (`skipExecutionClaim`), since they mint a fresh id with no backing row.

**3. The dedupe skip is now reported**, carrying the status/age of the row it deferred to. Rows the sweep will not drive again become terminal `failed` — the one status the existing per-delivery retry endpoint accepts, so they stay hand-drivable from the UI. A `retrying` row whose worker died is resolved terminally and **never re-sent**: its outcome is genuinely unknown, and repeating a possibly-delivered POST is worse than surfacing one unknown outcome.

**4. The subscriber's `try/catch` moved inside the loop**, so one webhook cannot deny delivery to the others.

## Design calls worth flagging

- **Recovery attempts get their own `recovery_attempts` column** rather than borrowing `attempts`. The delivery callback overwrites `attempts` with the HTTP attempt count and the UI renders it as such, so a counter kept there would reset on the first completed callback and misreport enqueue recoveries as delivery attempts.
- **The sweep re-queues the SAME row**, under the original `event_id` — that id goes out as `X-Breeze-Event-Id`, the customer's own idempotency key, and is what the unique index dedupes on. (The manual retry endpoint mints a fresh UUID; that is not a good precedent for recovery.)
- The decrypt path is extracted to `services/webhookConfig.ts`, and the two `index.ts` CAS closures to `services/webhookDeliveryRecord.ts`, so the sweep and the subscriber cannot drift.

## Evidence

**The partial index is the load-bearing claim** — `webhook_deliveries` is empty in both regions today and has **no retention job anywhere**, so it grows unbounded while the sweep ticks forever. Measured against a 50,000-row fixture on real Postgres 16:

```
 table_size | partial_idx_size
------------+------------------
 6160 kB    | 16 kB

 ->  Index Scan using webhook_deliveries_unresolved_idx on webhook_deliveries
       Index Cond: (created_at < (now() - '00:15:00'::interval))
       Filter: ((next_retry_at IS NULL) OR (next_retry_at <= now()))
 Execution Time: 0.079 ms
```

Index scan, not a seq scan, and the index stays ~0 because `pending`/`retrying` are transient. The migration was applied twice to confirm re-application is a true no-op.

**Predicates are pinned as compiled SQL via `PgDialect`** in files that import the real drizzle-orm — a mocked-drizzle suite can only substring-match column names and cannot tell an `and` from an `or`. `NOT (...)` over the nullable `next_retry_at` is avoided explicitly and asserted against.

**A real-Postgres suite proves what mocks cannot**: that the `IS NULL` arm actually matches a never-leased row (this repo has shipped a *correct* compiled-SQL assertion while only real Postgres caught the bug), that both CAS predicates genuinely serialise two concurrent callers, that the read-back does not confuse a sibling webhook's delivery for this one, and that the partial index exists and is partial.

**21 revert probes**, each mutating the fix and re-running the suites — all 21 red. Two initially passed **vacuously**; the tests that missed them were fixed, not the probes (one assertion was satisfied by a downstream `TypeError` rather than the guard; the other by a success path that returns before the retry ladder it claimed to test).

```
vitest run <unit suites>                    -> 7 files, 49 tests passed
vitest run --config vitest.integration      -> 2 files, 15 tests passed (real Postgres)
tsc --noEmit                                -> clean
eslint <changed files>                      -> clean
scripts/check-migration-naming.sh           -> OK
```

## Review

Three reviewers ran (`code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`) plus an independent design review before implementation. The second commit is entirely review findings — several were cases where this fix would have introduced the class of bug it removes:

- the catch now covering a real customer POST had no Sentry, and could not distinguish a failure before the POST (recoverable) from one after it (not);
- the outcome write was keyed on `id` alone, which stopped being safe the moment the sweep became a second writer;
- `retrying` was classified as a benign dedupe while the sweep treats it as unresolved;
- `retryFromDLQ` would have been eaten by the new claim and reported as a "duplicate";
- all four terminal writes discarded `markUnrecoverable`'s result, so a lost lease still logged "abandoned";
- "never enqueued" claimed more than the code can prove.

Reviewers independently confirmed the migration ordering, the Drizzle/SQL index parity, the sub-hourly `every:` interval, and that `webhook_deliveries` (no `org_id`) needs no cascade or export-policy registration.

## Out of scope, worth filing

- After a *failed* attempt the row is already `failed`; if the retry LPUSH then dies, the retry is lost and a sweep scoped to unresolved rows never sees it. Same shape as this bug, one hop later — needs #4085's durable retry state.
- `webhook_deliveries` still has no retention job, so the `(webhook_id, event_id)` dedupe index grows without bound with no defined horizon.
- Pre-existing, noticed in review: `webhook_deliveries.webhook_id → webhooks.id` has no `ON DELETE` and the table is in none of the cascade lists, while `webhooks` is a cascade-registered org table. Looks like a latent org-erasure gap; untouched here.

### DIFF
--- a/apps/api/migrations/2026-09-11-d-webhook-delivery-recovery.sql
+++ b/apps/api/migrations/2026-09-11-d-webhook-delivery-recovery.sql
@@ -1,0 +1,39 @@
+-- #4095: recovery path for webhook deliveries recorded but never enqueued.
+--
+-- `webhook_deliveries` rows are committed to Postgres BEFORE `queueDelivery`,
+-- which is a bare `redis.lpush` onto a non-durable list. If that LPUSH throws,
+-- or the process dies between the commit and the LPUSH, the row sits at
+-- `status = 'pending'` with no job anywhere and nothing ever looks at it again.
+-- Latent while delivery is at-most-once; once wave 3.5c (#4085) makes it
+-- at-least-once, the redelivered event hits the (webhook_id, event_id) unique
+-- index from 2026-09-11-a, the '*' subscriber skips, and that pair is
+-- suppressed forever.
+--
+-- Two additions, both driving `jobs/webhookDeliveryRecovery.ts`.
+
+-- 1. A dedicated recovery counter.
+--
+-- The sweep needs a bounded number of attempts per row, and it cannot borrow
+-- `attempts`: that column is the HTTP attempt count the delivery callback
+-- overwrites (`SET attempts = result.attempts`) and the UI renders as such.
+-- Counting enqueue recoveries in it would both misreport the delivery history
+-- and lose the count on the first completed callback.
+ALTER TABLE webhook_deliveries
+  ADD COLUMN IF NOT EXISTS recovery_attempts integer NOT NULL DEFAULT 0;
+
+-- 2. A PARTIAL index over the candidate set.
+--
+-- The sweep runs every five minutes forever, and `webhook_deliveries` has no
+-- retention job anywhere — it grows without bound. A plain index on
+-- (status, created_at) would grow with the table; this one only ever holds
+-- rows in a TRANSIENT state, so in a healthy fleet it stays at approximately
+-- zero entries no matter how large the table gets, and the sweep is an index
+-- scan over nothing rather than a sequential scan over everything.
+--
+-- `pending` and `retrying` are both indexed because both are unresolved:
+-- `pending` = never claimed by a worker (safe to re-drive), `retrying` =
+-- claimed but never completed (outcome unknown, resolved terminally rather
+-- than re-POSTed). `delivered` and `failed` are terminal and excluded.
+CREATE INDEX IF NOT EXISTS webhook_deliveries_unresolved_idx
+  ON webhook_deliveries (status, created_at)
+  WHERE status IN ('pending', 'retrying');
--- a/apps/api/migrations/2026-09-11-d-webhook-delivery-recovery.sql
+++ b/apps/api/migrations/2026-09-11-d-webhook-delivery-recovery.sql
@@ -34,6 +34,13 @@ ALTER TABLE webhook_deliveries
 -- `pending` = never claimed by a worker (safe to re-drive), `retrying` =
 -- claimed but never completed (outcome unknown, resolved terminally rather
 -- than re-POSTed). `delivered` and `failed` are terminal and excluded.
+-- Keyed on created_at ALONE, not (status, created_at). The partial predicate
+-- already restricts the index to the two unresolved statuses, so carrying
+-- `status` as the leading key buys nothing and actively costs: with a 2-value
+-- IN, a (status, created_at) index yields two separately-ordered runs, so the
+-- sweep's `ORDER BY created_at` needs a sort (or a merge) on top. Keyed on
+-- created_at the index IS the order, oldest-first, which is exactly how the
+-- sweep drains its bounded batch. Same number of entries either way.
 CREATE INDEX IF NOT EXISTS webhook_deliveries_unresolved_idx
-  ON webhook_deliveries (status, created_at)
+  ON webhook_deliveries (created_at)
   WHERE status IN ('pending', 'retrying');
--- a/apps/api/src/__tests__/integration/webhookDeliveryRecovery.integration.test.ts
+++ b/apps/api/src/__tests__/integration/webhookDeliveryRecovery.integration.test.ts
@@ -250,8 +250,19 @@ describe('webhook delivery recovery against real Postgres', () => {
     // so the same event legitimately gets one row per webhook.
     expect(mine.created).toBe(true);
     expect(theirs.created).toBe(true);
-    expect(mine.created === true && theirs.created === true
-      && mine.deliveryId !== theirs.deliveryId).toBe(true);
+    const myId = mine.created === true ? mine.deliveryId : null;
+    const theirId = theirs.created === true ? theirs.deliveryId : null;
+    expect(myId).not.toBe(theirId);
+
+    // Now force the READ-BACK, which the two winning inserts above never reach.
+    // With a sibling row present under the SAME event_id, a lookup keyed only
+    // on event_id could return either row — and Postgres is free to hand back
+    // the sibling's. It must name THIS webhook's delivery.
+    const again = await recordWebhookDelivery({ id: webhookId } as never, event as never);
+
+    expect(again.created).toBe(false);
+    expect(again.created === false && again.existing?.id).toBe(myId);
+    expect(again.created === false && again.existing?.id).not.toBe(theirId);
   });
 
   it('the execution claim lets exactly ONE popped copy of a job deliver', async () => {
@@ -265,7 +276,11 @@ describe('webhook delivery recovery against real Postgres', () => {
       claimDeliveryForExecution(job)
     ]);
 
-    expect([a, b].filter(Boolean)).toHaveLength(1);
+    expect([a, b].filter((r) => r.claimed)).toHaveLength(1);
+    // The loser must be able to say WHY: the winner moved the row to
+    // `retrying`, which is "already claimed", not "no row" or "already done".
+    const loser = [a, b].find((r) => !r.claimed);
+    expect(loser).toMatchObject({ claimed: false, observedStatus: 'retrying' });
 
     const [row] = await withSystemDbAccessContext(() =>
       db
@@ -282,7 +297,16 @@ describe('webhook delivery recovery against real Postgres', () => {
   it('the execution claim refuses a row that has already resolved', async () => {
     const deliveryId = await insertDelivery(webhookId, { status: 'delivered' });
 
-    expect(await claimDeliveryForExecution({ id: deliveryId } as never)).toBe(false);
+    expect(await claimDeliveryForExecution({ id: deliveryId } as never))
+      .toMatchObject({ claimed: false, observedStatus: 'delivered' });
+  });
+
+  it('the execution claim reports a MISSING row distinctly from a lost race', async () => {
+    // A DLQ replay's minted id has no row at all. Collapsing that to the same
+    // "duplicate" verdict as a lost race sends triage after a race that never
+    // happened.
+    expect(await claimDeliveryForExecution({ id: randomUUID() } as never))
+      .toMatchObject({ claimed: false, observedStatus: null });
   });
 
   it('the partial index the sweep depends on exists and is partial', async () => {
--- a/apps/api/src/__tests__/integration/webhookDeliveryRecovery.integration.test.ts
+++ b/apps/api/src/__tests__/integration/webhookDeliveryRecovery.integration.test.ts
@@ -1,0 +1,305 @@
+/**
+ * #4095, against real Postgres: prove the recovery predicates actually select
+ * and actually serialise.
+ *
+ * WHY THIS FILE EXISTS
+ * --------------------
+ * The sweep is pinned by a mock-based unit suite and by compiled-SQL assertions
+ * (`jobs/webhookDeliveryRecovery.sql.test.ts`), and neither can prove the two
+ * things that decide whether this fix works in production:
+ *
+ *   1. that `next_retry_at IS NULL OR next_retry_at <= now` really matches a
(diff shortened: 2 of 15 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` (plus an independent design review via `codex exec` *before* implementation, which is what caught the duplicate-POST hazard in the original design).

**Findings:** 15 raised → 14 addressed in `644da5157`; 1 accepted as out of scope (below). 0 outstanding.

The consequential ones — several were cases where this fix would have introduced the class of bug it removes:

- **The sweep could have double-POSTed a customer.** It cannot distinguish a lost job from one merely backlogged on the Redis list, and `BRPOP`'s 5s is only its empty-queue wait — the FIFO backlog ahead of a job is unbounded, so *no* staleness threshold is safe on its own. Fixed by an execution claim (`pending` → `retrying` CAS) on the delivery worker; that also bounds the sweep, since a claimed row leaves the candidate set at once.
- **The catch now wrapping a real customer POST had no Sentry**, and could not tell a failure *before* the POST (recoverable — the sweep reclaims it) from one *after* (not — the endpoint was already called and that fact is unrecorded). Split, instrumented, and rethrown so the retry ladder cannot run on an outcome we failed to record.
- **The outcome write was keyed on `id` alone**, which stopped being safe the moment the sweep became a second writer — and `deliverWebhook` clears its abort timeout once response *headers* arrive, so a slow body can land after the lease expired. Now a CAS that will not overwrite a recorded success.
- **`retrying` was classified as a benign dedupe** while the sweep treats it as unresolved — burying the exact case this issue is about.
- **`retryFromDLQ` would have been eaten** by the new claim and reported as a "duplicate" (it mints a fresh id with no backing row). Unreachable today, but a landmine this PR armed.
- **All four terminal writes discarded `markUnrecoverable`'s result**, so a lost lease still logged "abandoned" and incremented a resolution counter for a row that was still unresolved.
- **"Never enqueued" claimed more than the code can prove** — an unclaimed row proves only that no *worker* took it. Reworded so on-call isn't pointed at the wrong subsystem.

Reviewers independently confirmed the migration ordering, Drizzle↔SQL index parity, the sub-hourly `every:` interval, and that `webhook_deliveries` (no `org_id`) needs no cascade or export-policy registration.

**Tests:** 7 unit files / 49 tests green; 2 integration files / 15 tests green against real Postgres. **21 revert probes, all red** — two initially passed *vacuously* and the tests that missed them were fixed, not the probes (one assertion was satisfied by a downstream `TypeError` rather than the guard it named; the other by a success path that returns before the retry ladder it claimed to test). Predicates are pinned as compiled SQL via `PgDialect`, not column-name substrings.

**CI:** 56 pass / 0 fail on `644da5157`, including all 4 integration shards. Verified the new suite actually *executed* rather than being silently skipped — `✓ webhookDeliveryRecovery.integration.test.ts (10 tests) 3227ms` in shard 2/4. `tsc --noEmit` and `eslint` clean locally.

**Not fixed, deliberately** (needs #4085's durable retry state): after a *failed* attempt the row is already `failed`, so if the retry LPUSH then dies the retry is lost and a sweep scoped to unresolved rows never sees it — same shape as this bug, one hop later in the ladder. Also noted in review and left alone: `webhook_deliveries` has no retention job (so the dedupe index has no horizon), and its FK to `webhooks` has no `ON DELETE` while appearing in none of the cascade lists — a pre-existing latent org-erasure gap worth its own issue.

**Status:** review-clean, CI green, awaiting maintainer merge. Not merged and #4095 left open per the hand-off rules.

ToddHebebrand:
**Review run 2:** independent 2-lens review (code + silent-failure) at `644da5157`. Design confirmed — no duplicate-send hole across six interleavings, no stranding, migration correct — with 12 defects to fix. All addressed in **`5da86cffc`**.

**Behaviour (6):**

| # | Finding | Fix |
|---|---|---|
| 1 | The whole-event routing catch was **the one drop with no recovery path** — the lookup fails, so no rows exist and the sweep can't see the event either; the entire fan-out is gone for every subscribed webhook. No Sentry, no event identity, and the transient branch claimed it "will retry on next event" — false | `captureException` + structured `WEBHOOK_EVENT_ROUTING_FAILED` with `eventId`/`orgId`/`eventType` and an explicit statement that the fan-out is lost |
| 2 | DLQ replays **poisoned the new zero-row warning** — `retryFromDLQ` mints an id with no row, so the outcome write matched nothing and fired `OUTCOME_WRITE_SKIPPED` on *every* routine replay with an untrue cause, voiding it for the real race | `hasDeliveryRow` on the result; the writer skips the row UPDATE and logs `WEBHOOK_DLQ_REPLAY_COMPLETED` at info, still moving aggregates since the POST happened |
| 3a | The claim charged a recovery attempt **before** the enqueue and never refunded it — 6 consecutive failures (~75 min of Redis downtime) would terminally fail a delivery never attempted, charging infra downtime against the customer | Attempt returned under the lease CAS; lease deliberately kept so the next window waits out the cooldown rather than hot-looping |
| 3b | `claimDeliveryForExecution` returned a bare boolean, so the log couldn't tell "lost race" from "already resolved" from "no row" | Returns the observed status; the log names the cause |
| 10 | `initializeWebhookDeliveryRecovery` leaked a live BullMQ Worker if queue setup threw, with no handle left to stop it | close + null first |
| 12 | No upper age bound — on first deploy the sweep could POST months-old payloads | 7d ceiling, as a *terminal, loudly logged resolution* rather than a scan filter (filtering would leave them unresolved forever, which is this bug) |

**Index (11):** you were right, and it's measurable. Re-keyed to `(created_at)` alone — the partial predicate already restricts to the two unresolved statuses, so a leading `status` bought nothing and cost a sort. Verified on 50k rows: **the `Sort` node is gone**, plan is a bare `Index Scan`, same 16 kB. Edited in place since the migration is unmerged.

**Comment (4):** documented that neither the status test nor the id defends the SELECT→CAS gap — against a rival sweeper *or* a delivery worker, it's the **lease predicate** that serialises. It read as if the status test were the guard.

**Tests (5, 6, 7, 8, 9)** — each was green under the mutation it should have killed:
- `retryFromDLQ`'s bypass flag **and** the worker's `hasDeliveryRow` stamp had zero producer-side coverage; deleting either was green repo-wide. Both pinned. (Your item 2 had this second, unlisted half — I found it by probing.)
- The outcome-write CAS is extracted out of coverage-excluded `index.ts`; compiled-SQL + zero-row + DLQ-branch coverage added.
- `webhookConfig.test.ts` mocked `decryptWebhookHeaders` as identity, so dropping the decrypt wrap — **shipping encrypted header values to customer endpoints as live headers** — stayed green. Mock now transforms.
- The integration sibling case never reached the read-back (both inserts won); it now forces a third conflicting insert and asserts the returned row is this webhook's.
- `BENIGN_DUPLICATE_STATUSES` fully guarded, plus the `failed` case.

**Verification:** 14 unit files / 164 tests, 2 integration files / 16 tests (real Postgres), `tsc` + `eslint` + migration guard clean. **14 new revert probes, all red** — one initially passed vacuously and the test that missed it was fixed, not the probe.

**CI on `5da86cffc`** (the new head, not the old one): **56 pass / 0 fail**, all 4 integration shards. Verified the suite actually executed — `✓ webhookDeliveryRecovery.integration.test.ts (11 tests) 4372ms` in shard 2/4.

**Not done, per your instruction:** decrypt-failure-terminal-on-transient-KMS and re-checking `webhook.events` before re-POST — follow-up material.

**Status:** review-clean, CI green on the current head, awaiting maintainer merge. Still not merged; #4095 still open.

PR #4099, opened 2026-08-26
fix(api): route every alert-resolve path through the compare-and-swap (#4094)

### LINKED ISSUE #4094: [API] Two user-facing resolve routes bypass the resolveAlert compare-and-swap
Found during the code review of #4083 (AI agents wave 3.5a). Not a regression from
that PR — pre-existing, and it undercuts the guarantee 3.5a establishes.

## The gap

`resolveAlert()` (`apps/api/src/services/alertService.ts`) was changed in #4083 to a
compare-and-swap:

```ts
.where(and(eq(alerts.id, alertId), inArray(alerts.status, RESOLVABLE_ALERT_STATUSES)))
.returning()
```

so only one caller can transition an alert, and only the winner publishes
`alert.resolved` (which cancels pending escalations and fans out to the `'*'`
automation subscriber).

**The two user-facing resolve routes bypass it entirely** and still do
read-status → unconditional `UPDATE ... WHERE eq(alerts.id, alertId)` → `publishEvent`:

- `apps/api/src/routes/alerts/alerts.ts:852-869`
- `apps/api/src/routes/mobile.ts:981-999`

Textbook check-then-act. These are also the **most likely** concurrent resolvers in
production: two techs on the same alert, or a tech racing the auto-resolve sweep
(`checkAutoResolve` / `checkAutoResolveFromConfigPolicy`, which DO go through the CAS).

Concrete double-fire: two techs resolve the same alert within the same second. Both
UPDATEs succeed, both publish `alert.resolved`, escalation cancellation runs twice,
and the automation subscriber sees the event twice.

Note the repo already has the right shape one file over — the bulk-alert route does
a CAS and surfaces `results.skipped` + `skippedIds` to the caller
(`routes/alerts/alerts.ts:636-642`).

## Suggested fix

Route both through `resolveAlert()` and use its boolean:

- winner → 200 as today;
- loser → 409 (or 200 with an `alreadyResolved` flag, matching whatever the bulk
  route's `skipped` shape does) rather than reporting a resolution that did not happen.

Add a route test asserting the losing call does not publish.

## Related

- The boolean `resolveAlert` now returns is also **unconsumed** at three internal
  sites, where a losing caller still writes a cooldown and inflates its count:
  `alertService.ts:238,246` (returns `true` regardless) and `alertService.ts:811-813,
  820-822` (`resolvedCount++` unconditionally). Worth fixing in the same pass.
- Wave 3.5c (#4085) makes event delivery at-least-once, which raises the odds of the
  concurrent case considerably.

ToddHebebrand:
Closes #4094

## The gap

#4083 made `resolveAlert()` a compare-and-swap so only one caller can transition an alert, and only the winner publishes `alert.resolved` (which cancels pending escalations and fans out to the `'*'` automation subscriber).

**Five other paths that stamp an alert `resolved` never went through it.** Each read the status, then ran `UPDATE ... WHERE eq(alerts.id, alertId)` unconditionally, then published — textbook check-then-act. Two techs resolving the same alert within the same second both "won": both UPDATEs succeeded, both published, escalation cancellation ran twice, and the automation subscriber saw the event twice.

## What changed

Every path now uses the one exported `buildResolveAlertCas(alertId)` predicate.

| Path | Source |
|---|---|
| `routes/alerts/alerts.ts` — `POST /alerts/:id/resolve` | named in the issue |
| `routes/mobile.ts` — `POST /mobile/alerts/:id/resolve` | named in the issue |
| `services/aiToolsAlerts.ts` — `manage_alerts` `action:"resolve"` | **found by the repo-wide sweep** |
| `routes/alerts/correlations.ts` — group resolve/acknowledge | **found by the repo-wide sweep** |
| `services/warrantyAlertEvaluator.ts` — auto-resolve loop | **found by the repo-wide sweep** |

The three extra sites are the same defect and the same blast radius; the AI-tool one matters more than most, since #4085 makes event delivery at-least-once and an agent racing a technician is exactly the case that gets likelier.

### `resolveAlert` was building a second copy of its own predicate

`buildResolveAlertCas` was exported by 8763a3239 specifically so the CAS could be asserted as compiled SQL, but `resolveAlert` kept an inline duplicate. The compiled-SQL test was therefore one step removed from production and the two could drift apart silently. Demonstrated: narrowing the *inline* list to `['active']` left both pre-existing helper tests green. `resolveAlert` now calls the helper.

### The unconsumed boolean (second half of the issue)

- `alertService.ts:238,246` — `checkAutoResolve` returned a hard-coded `true`, so `checkAllAutoResolve`'s count included every caller that lost the race. Now returns what the CAS did.
- `alertService.ts:811-813,820-822` — `checkAutoResolveFromConfigPolicy` incremented `resolvedCount` **and wrote the config-policy cooldown** unconditionally. A losing sweep reported a resolution it never performed and stamped a cooldown that suppresses the next legitimate alert for that rule. Both now gate on the winner.
  The explicit cooldown write is **removed** rather than gated: `resolveAlert`'s winning path already calls `markConfigPolicyRuleCooldown` with the same rule id, device and `cooldownMinutes`, so it was a duplicate of a write the winner performs anyway.

## API surface change: "already resolved" is 409, not 400

The pre-read check and losing the CAS are the **same outcome** — the alert is terminal and this request did not transition it. Leaving the pre-read at 400 while the race returned 409 would make one user action return two status codes depending purely on timing. Both are now 409. **Dismissed stays 400** — that is a genuine "you can't do that", not a race.

`openapi.ts` previously declared only a `200` for `resolveAlert`; the 400 and 409 are now documented. The only HTTP consumers are two web components (below); `/mobile/alerts/:id/resolve` has **no** client caller today — `apps/mobile` never shipped a resolve affordance.

## Web

Both callers discard the response body and re-fetch, so a 409 needs no body parsing, and `extractApiError` already surfaces the server's `error` sentence into the toast — no new i18n keys.

- `AlertDetailPage` no longer latches its **persistent red banner** on a 409 (framing a benign race as a broken page) and re-fetches so the header stops showing a stale open status.
- `AlertsPage` refreshes the list on a 409 for the same reason — previously it did nothing, leaving the row offering *Resolve* on an already-finished alert.

Feedback is deliberately **not** suppressed: a resolve that silently does nothing is the failure mode `runAction` exists to prevent, and there's a test pinning that the error toast still fires.

## Testing

Per the issue's warning and main 8763a3239: **no test here mocks drizzle-orm wholesale or substring-matches column names.** Every suite keeps `drizzle-orm` and `db/schema` real, mocks only the DB *connection*, captures the predicate the code actually passed to `.where()`, and asserts its **compiled SQL** through `PgDialect` — exact string and exact params.

**Mutation probes run (each confirmed red, then restored):**

| Mutation | Result |
|---|---|
| `buildResolveAlertCas` `and()` → `or()` | **6 tests red across all 5 files** |
| `resolveAlert`'s inline list → `['active']` (pre-fix shape) | 2 new tests red; both pre-existing helper tests stayed **green** — this is the gap being closed |
| `aiToolsAlerts` predicate → `eq(alerts.id, …)` alone | 1 red |
| `correlations` `and(...)` → `or(...)` | 2 red (resolve + acknowledge shapes) |
| `warrantyAlertEvaluator` predicate → `eq(alerts.id, …)` alone | 1 red |
| Revert `alertService.ts` entirely | 3/6 of the new auto-resolve outcome tests red |
| Revert both routes | 6/12 of the new route tests red |
| Revert the `AlertDetailPage` 409 branch | 1 red |

**Two pre-existing suites needed mock updates** (not assertion changes): `aiToolsAlerts.feedback.test.ts` and `warrantyAlertEvaluator.test.ts` stubbed `.where()` as terminal, and the CAS now chains `.returning()`. Both helpers default to a one-row winner so every existing assertion keeps its original meaning, and both can now model a loser. `mobile.test.ts`'s "should reject already resolved alerts" moved 400 → 409 with a comment recording why.

**Commands run:**
```
apps/api  vitest run --pool=forks --fileParallelism=false  (8 service files)   34 passed
apps/api  vitest run --pool=forks --fileParallelism=false  (9 route files)    191 passed, 3 skipped
apps/web  vitest run --fileParallelism=false               (4 files)          140 passed
apps/api  tsc --noEmit                                                        clean
apps/web  tsc --noEmit                                                        clean
eslint    (8 changed production files)                                        clean
```

## Design note — one deliberate divergence from the issue's suggestion

The issue suggested routing both routes *through* `resolveAlert()`. An independent Codex `xhigh` review agreed. **I did not**, for two reasons that only surface once you read the call graph:

1. **It would silently change flapping behaviour.** `resolveAlert` calls `recordStateTransition(...)`; the routes do not. `isFlapping` "counts all recorded events, not only alternations" and, when it trips, **suppresses new alert creation** (`alertService.ts:116-122`). Routing manual resolves through it would push rules toward that threshold and silently stop alerting — a customer-visible behaviour change riding inside a concurrency bugfix.
2. **It cannot cover all five sites.** `correlations.ts` resolves a group in one bulk UPDATE; per-alert `resolveAlert` calls would turn that into N round-trips.

There are also two contract deltas neither function subsumes: the routes publish `resolvedBy` with `{ userId }` metadata, while `resolveAlert` publishes neither but attaches `{ siteId }` — which gates site-restricted WebSocket delivery (`eventWs.ts:419`).

So the fix shares the thing that actually governs correctness — **the predicate** — and leaves the five side-effect pipelines alone. This also matches the repo's own reference shape one file over: the bulk-alert route the issue itself cites does an inline CAS.

## Out of scope (noted, not fixed)

The single-alert **acknowledge** handlers have the same check-then-act shape. Left alone: `alert.acknowledged` carries no escalation cancellation and no `'*'` automation fan-out, so a duplicate is far less harmful. Worth a follow-up issue if you want the symmetry.

https://claude.ai/code/session_012o7QB15EFjEvetDAXMxmae

### DIFF
--- a/apps/api/src/openapi.ts
+++ b/apps/api/src/openapi.ts
@@ -2806,6 +2806,26 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
                 schema: { $ref: '#/components/schemas/Alert' }
               }
             }
+          },
+          '400': {
+            description: 'Alert is dismissed and cannot be resolved',
+            content: {
+              'application/json': {
+                schema: { $ref: '#/components/schemas/Error' }
+              }
+            }
+          },
+          '409': {
+            description:
+              'Alert already reached a terminal status — either before this request, ' +
+              'or because a concurrent resolver won the compare-and-swap. The alert is ' +
+              'resolved; this request simply did not perform the transition, so no ' +
+              'alert.resolved event was published for it.',
+            content: {
+              'application/json': {
+                schema: { $ref: '#/components/schemas/Error' }
+              }
+            }
           }
         }
       }
--- a/apps/api/src/openapi.ts
+++ b/apps/api/src/openapi.ts
@@ -2817,10 +2817,11 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
           },
           '409': {
             description:
-              'Alert already reached a terminal status — either before this request, ' +
-              'or because a concurrent resolver won the compare-and-swap. The alert is ' +
-              'resolved; this request simply did not perform the transition, so no ' +
-              'alert.resolved event was published for it.',
+              'The alert already reached a terminal status (resolved or dismissed) — ' +
+              'either before this request, or because a concurrent caller won the ' +
+              'compare-and-swap in between. This request did not perform the transition, ' +
+              'so no alert.resolved event was published on its behalf. Re-read the alert ' +
+              'to see which terminal status it landed in.',
             content: {
               'application/json': {
                 schema: { $ref: '#/components/schemas/Error' }
--- a/apps/api/src/routes/alerts/alerts.resolveCas.test.ts
+++ b/apps/api/src/routes/alerts/alerts.resolveCas.test.ts
@@ -1,0 +1,219 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Hono } from 'hono';
+import { PgDialect } from 'drizzle-orm/pg-core';
+import type { SQL } from 'drizzle-orm';
+
+/**
+ * POST /alerts/:id/resolve is a compare-and-swap (#4094).
+ *
+ * It used to read the alert's status and then UPDATE by id unconditionally —
+ * textbook check-then-act. Two techs resolving the same alert within the same
+ * second both "won": both stamped the row, both published `alert.resolved`
+ * (cancelling pending escalations twice and handing the `'*'` automation
+ * subscriber the same event twice), and both wrote a cooldown.
+ *
+ * Two kinds of assertion here, deliberately:
+ *
+ *  - BEHAVIOUR — an empty `RETURNING` (exactly what Postgres gives a caller whose
+ *    CAS matched nothing) must produce a 409 and NO fan-out.
+ *  - COMPILED SQL — the predicate the handler actually passed to `.where()`,
+ *    compiled through `PgDialect`. drizzle-orm and ../../db/schema are REAL in this
+ *    file for that reason: a mocked-drizzle `where` assertion can only substring-
+ *    match column NAMES, which cannot tell `and` from `or` and cannot see the
+ *    status list gaining a terminal value. Both of those mutations were verified
+ *    to pass green against exactly that style of test (main 8763a3239).
+ */
+const { dbMock, updateWheres, updateReturns, alertRow } = vi.hoisted(() => {
+  const updateWheres: unknown[] = [];
+  const updateReturns: unknown[][] = [];
+  const alertRow = {
+    id: '11111111-1111-4111-8111-111111111111',
+    orgId: 'org-1',
+    deviceId: 'device-1',
+    ruleId: null as string | null,
+    configPolicyId: null as string | null,
+    context: null as unknown,
+    status: 'active',
+    title: 'Disk almost full',
+  };
+  const dbMock = {
+    select: () => ({
+      from: () => ({ where: () => ({ limit: () => Promise.resolve([]) }) }),
+    }),
+    update: () => ({
+      set: () => ({
+        where: (predicate: unknown) => {
+          updateWheres.push(predicate);
+          return { returning: () => Promise.resolve(updateReturns.shift() ?? []) };
+        },
+      }),
+    }),
+  };
+  return { dbMock, updateWheres, updateReturns, alertRow };
+});
+
+const publishEvent = vi.fn((..._args: unknown[]) => Promise.resolve('evt'));
+const emitAlertStateFeedback = vi.fn((..._args: unknown[]) => Promise.resolve());
+const writeRouteAudit = vi.fn();
+const setCooldown = vi.fn((..._args: unknown[]) => Promise.resolve());
+const markConfigPolicyRuleCooldown = vi.fn((..._args: unknown[]) => Promise.resolve());
+const getAlertWithOrgCheck = vi.fn();
+
+vi.mock('../../db', () => ({
+  db: dbMock,
+  runOutsideDbContext: (fn: () => unknown) => fn(),
+  withSystemDbAccessContext: (fn: () => unknown) => fn(),
+}));
+vi.mock('../../middleware/auth', () => ({
+  authMiddleware: async (_c: unknown, next: () => Promise<void>) => next(),
+  requireScope: () => async (c: { set: (k: string, v: unknown) => void }, next: () => Promise<void>) => {
+    c.set('auth', {
+      scope: 'organization',
+      user: { id: 'user-1', name: 'Tech One', email: 'tech@org.example' },
+      partnerId: null,
+      orgId: 'org-1',
+      accessibleOrgIds: null,
+      canAccessOrg: () => true,
+    });
+    await next();
+  },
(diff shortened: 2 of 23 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, run in parallel against `66d40a75a`).

**Findings:** 9 raised (2 critical, 4 important, 3 suggestion) → all 9 addressed in `9072dc4fc`; 0 outstanding.

- **critical** *silent-failure-hunter* — adding the status predicate to `correlations.ts` gave the `writtenIds` shortfall a second, benign cause, so its `captureException` would now fire on ordinary contention and bury a real cross-tenant write in noise. `reportUnwrittenAlerts` re-reads the unwritten rows in the same request context and pages only for the half a concurrent transition cannot explain.
- **critical** *comment-analyzer* — four test files cited `8763a3239` as a main commit. Verified with `git merge-base --is-ancestor`: it is **not** on `origin/main` (unmerged wave-3825 branch only). Citations corrected.
- **important** *pr-test-analyzer* — every `checkAutoResolveFromConfigPolicy` fixture pinned `autoResolveConditions: null`, so the CAS gate in the other arm could regress green. Both arms now parameterised + two guards proving the fixtures really select different branches.
- **important** *pr-test-analyzer* — both routes' post-CAS cooldown branches sat behind a uniform `ruleId/configPolicyId: null` fixture. Covered, asserting the override chain's **value** (42, not the template's 15) so it cannot pass vacuously.
- **important** *pr-test-analyzer* — `AlertsPage`'s 409 handler had no coverage. Added, with a 500 contrast case proving the branch keys on the status.
- **important** *silent-failure-hunter* — the warranty sweep's bare `continue` left no trace; it now warns once per invocation when it transitions 0 of N (never per loss).
- **suggestion** ×3 — cause-neutral 409 body (shared `ALERT_CAS_LOST_MESSAGE`), self-contradicting OpenAPI 409 description, and the one unguarded recovery path (`AlertDetailPage`'s post-409 `fetchAlert`).

**Vacuity check:** the reviewers ran ~40 independent mutation probes against the new suites (`and`→`or`, dropped id equality, terminal status admitted, each converted site reverted, the 409 branch deleted). Every probe went red on its intended target; none passed green. Separately, narrowing `resolveAlert`'s inline list left both pre-existing helper tests green while the two new production-path tests failed — that gap is what this PR closes.

**A CI regression the review caught that local runs had not:** `aiToolsAlerts.siteScope.test.ts` — a *third* sibling suite whose mock stubbed `.where()` as terminal — broke on the new `.returning()` (1 of 25,715 tests). Fixed, and the sweep for stale update-mocks extended across every suite touching the six changed modules.

**Tests:** `apps/api` 13 service files (79 passed) + 9 route files (198 passed, 3 skipped), single-fork; `apps/web` 5 files (142 passed). `tsc --noEmit` clean in both packages; eslint clean on all 8 changed production files.

**CI:** 54 passing, 0 failing — including **Test API** (the job that failed on the first push) and all four **Integration Tests** shards.

**Status:** review-clean, awaiting maintainer merge. Two calls are yours, both written up in the PR body: the deliberate divergence from the issue's "route through `resolveAlert()`" suggestion (it would newly feed `recordStateTransition` → `isFlapping`, which *suppresses* alert creation), and the 400 → 409 change on "already resolved".

Issue #3828, opened 2026-08-22
W07: Wave 6 — supervision surfaces, fix-held watch, ticket-triggered helpdesk agent, anomaly sources


ToddHebebrand:
Wave W07 of #3821 — Wave 6 — supervision surfaces, fix-held watch, ticket-triggered helpdesk agent, anomaly sources
Branch: `feature/3821-ai-agents/wave-3828`

ToddHebebrand:
## Inbound from wave 3 (#3824): the remediation/software/service event families land here

Wave 3 planning found that spec §7's "wave-3 prerequisite PR" for lifecycle event publishing is misfiled. Nothing in wave 3 consumes those events — the triage agent triggers on `alert.triggered`, which already exists and is published. They are **anomaly sources**, which this wave already scopes.

**Work that moves here:**

1. **Add the event types.** `policy.remediation.completed`, `service.restart_exhausted` and the `software.*` family are not `EventType` members at all today — `services/eventBus.ts:7-117` has only `policy.remediation.triggered`. This is add-the-types *and* write-the-publishers, not "move existing publishing".

2. **Design an honest vocabulary first.** `softwareRemediationWorker.ts:520,544` often only queues uninstall commands and leaves the remediation row `pending`. Publishing `completed` when that BullMQ job returns would be a lie. The events need to distinguish requested / command queued / command completed / failed / compliant-on-recheck.

3. **`service.restart_exhausted` needs a Go agent wire-format change.** Exhaustion is known only inside the Go monitor (`agent/internal/monitoring/monitor.go:206`); the transmitted result carries only attempted/succeeded (`agent/internal/monitoring/types.go:46`). So this one carries agent-shipped-code rigor and a **fleet promote** before it is useful — worth sequencing early in the wave.

Full reasoning: `docs/superpowers/plans/ai-mcp/2026-08-23-ai-agents-wave3.md` §0.2.0.

ToddHebebrand:
Completed via PR #4203

Wave 6 landed as four vertical PRs, all merged 2026-08-28/29:
- #4157 — supervision: execution-trace runs UI + exposure readout
- #4168 — safety controls: fix-held watch (ai_agent_fix_watches), per-org circuit breaker (ai_agent_circuit_state) with a single terminalization chokepoint (source-scan contract test), kill-state admin API + docs/deploy/ai-kill-switch.md
[...]

PR #4171, opened 2026-08-28
feat(api,web): wave 6.2a — fix-held watch, with the circuit ledger inert (#3828)

ToddHebebrand:
**PR 2 of 5 for #3828 — do NOT close the wave issue.**

Wave 6.1 let operators *see* what an agent did. This answers the question that follows — **did the fix hold?** — and lays the durable counter that 6.2b turns into an enforcing circuit breaker.

## The inertness contract

`ai_agent_circuits` is **written but read by no gate** in this PR. Counters accumulate and a circuit can reach `open`; nothing consults that state to make a decision. Verified mechanically: outside the writer module, its retention sweep, and tests, `grep aiAgentCircuits` returns nothing, and `actRevalidation.ts` / `services/actionIntents/` are untouched.

## Why enforcement is a separate PR

**"Downgrade to propose" does not mean "a human must approve."** `recordProposal` calls `createActionIntent` for tier-3 (`runLoop.ts`), `intentService.ts:628` then fires `attemptPolicyDecision`, and `manage_services:restart` — the exact op this wave watches — is in `POLICY_DECIDABLE_TIER3` (`policyDecidable.ts`). A circuit-open restart could therefore be "downgraded" to a proposal and then authorized unattended, routing the breaker around itself.

So enforcement needs **three** gate sites, not one:

1. `actRevalidation.ts` — after target normalization, before the exposure reservation (so a blocked op never burns a ledger row)
2. `attemptPolicyDecision` — one gate covers both its callers (the creation-time trigger and `intentReleaseWorker`)
3. `intentReleaseWorker`'s release path — the circuit-opens-after-authorization race

That is a safety change deserving its own review rather than a rider on a telemetry PR. The bypass is dark in production today (the policy-decide flag is off), which is what makes the split safe.

## Two lanes, and why

| Lane | What it asks | Device I/O |
|---|---|---|
| `alert_recurrence` | Did an alert of the same **identity** re-trigger on this device after the remediation? | None |
| `postcondition` | Re-run the op's own `verifySpec` read-back | Yes |

`service_running` is the **only** postcondition kind v1 watches. `process_absent` isn't in the act manifest at all, and re-checking a kill *by name* would be a broader claim than the postcondition rather than a weaker one (legitimate respawn, multiple instances, unrelated binaries sharing a name). `disk_usage_improved`'s verification only reads the cleanup command's own `status`/`failedCount` and never performs a disk read, so there is no baseline to compare against.

That leaves 1 of 5 live ops — which is precisely why the op-agnostic `alert_recurrence` lane exists. It is I/O-free, so it resolves while the device is offline, and it is the only thing that says anything at all about `run_script` or `execute_playbook`.

`readServiceRunningState` was factored out of `actVerify` and is **shared** with the delayed check. That is load-bearing: "the fix held" is only meaningful measured against the predicate that produced the original "passed", and two parsers drifting apart would let a watch report `held` for a state the immediate verify would have called `failed`.

## Sweeper design

DB-backed, not a delayed job per watch. The deciding failure mode: a watch row commits but its Redis job is lost (enqueue failure, flush/restore, eviction, operational reset), and there is no atomic Postgres+Redis transaction — so the row must be rediscoverable.

Lease-claim due rows with `FOR UPDATE SKIP LOCKED` → run the check **outside every transaction** (holding a connection across a device read is what exhausted the pool on 2026-05-21) → finalize guarded on the circuit `epoch`, which is exactly the window where a human could clear the breaker mid-check.

`inconclusive` is retried with backoff and **never** moves a counter — an offline device must not open a breaker.

## Tenancy

Both tables are org-owned **operational** records, not config, so neither takes the Partner-Wide First dual-ownership shape: each describes one action on one device in one target org, as `ai_agent_runs` does while its `agent_id` may name a partner-wide agent. Configurable thresholds belong in the already dual-owned agent policy surface.

All six registration lists updated. On a cross-org device move the rows are **DELETED, never re-stamped** — the first two entries of the previously-empty `CORE_DEVICE_ORG_MOVE_DELETE_TABLES`. A re-stamped pending watch would fire a device command against a device now owned by a different tenant, and would carry the source tenant's target data across the boundary.

## Verification

Run against an **isolated** Postgres — the shared test DB on `:5433` carries another worktree's migration ledger and fails the checksum guard on an unrelated file (`2026-09-11-g-drop-event-bus-events.sql`).

- `aiAgentFixWatch.integration.test.ts` — 11/11 (RLS forge 42501 on both tables, cross-org run refused 23503, lease exclusivity under real concurrency, expired-lease reclaim, live-lease skip, epoch guard, threshold)
- `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` — 9/9
- `rls-coverage` contract — 75/75
- `pg_catalog`: both tables `relrowsecurity` **and** `relforcerowsecurity` = `t`, isolation policies present
- Migration replays twice with no diff; `db:check-drift` clean (587 files match the ledger)
- API unit: 484 aiAgents + route/contract suites; typecheck clean
- Web: 88/88 including `localeParity` across all 8 locales

### Defects the suites caught that review would not have

1. **`recordCircuitFailure` threw against every real database.** A bare `${date}` inside a `sql` template reaches the postgres driver unserialized (Drizzle only handles Dates flowing through typed columns) → `ERR_INVALID_ARG_TYPE`. All three unit suites passed regardless — the mocks never bind a parameter.
2. **`verify_failed` was defined and incremented by nothing.** Half the circuit's evidence was missing: a remediation failing its immediate read-back every time would never have moved a counter.
3. **`2,32 * * * *` was an invalid sweep slot** — a 30-minute gap is not a *coarse* schedule, so the registry that exists to make epoch-collision slots visible does not account for it. Now hourly on minute 2 (never minute 0 — that co-fire was the #3793 pool hold), with `POSTCONDITION_WINDOW_MS` moved 15m → 30m so the window is sized against the cadence that actually exists.

## Design authority

Two-advisor quorum 2026-08-28 (Claude Fable 5 proposing, Codex gpt-5.6-sol @ `xhigh` independent). Agreed on all six questions; five Codex corrections accepted and recorded in the plan doc, including the enforcement-bypass finding above (verified against the code before acceptance) and the device-move deletion rule, which reversed the proposing advisor's lean.

## Known gaps

- The postcondition lane is **never exercised against a real device** in these tests — the integration suite covers lease exclusivity and the epoch guard, not a `list_services` round trip. That path is covered only through the shared helper's existing `actVerify` coverage.
- Recovery is **manual reset only**; there is no automatic half-open probe. A tripped breaker stays tripped until a human clears it. The quorum's position was that auto half-open is defensible only as a single leased on-demand probe, and that manual-reset-only is the correct v1 otherwise. The reset UI lands in 6.2b along with the gate.

Plan: `docs/superpowers/plans/ai-mcp/2026-08-28-ai-agents-wave6-2a-fix-watch.md`


ToddHebebrand:
**Closing as superseded.** Wave 6.2's safety controls landed via #4168 (`fa8d374dd`), and the remainder of this branch's scope is re-planned under the AI agents phase 2 feature — #4187, waves #4188–#4193 (P2-1 plan done). Nothing here merges as-is; the fix-held watch shape carries forward into the phase-2 waves.

PR #4105, opened 2026-08-26
fix(pam): make actuation lifecycle durable and restart-safe

ToddHebebrand:
## Summary

- Implements Track E Tasks 1–8: durable PAM actuation/result state, atomic desired-state lifecycle, capability-gated exact v2 dispatch, verified Windows lifecycle primitives, restart-safe reconciliation, device no-transfer protection, ownership-preserving `received` transport, a two-organization failure matrix, durable-ledger state-machine/fuzz tests, and the private fail-closed Windows evidence harness.
- The rc.3 signed-Windows matrix executed 25/25 cases with zero invariant failures and zero missing `received` receipts. Independent process/account/token inspection was clean after every accepted cleanup; 21 cleanup samples measured p95 442.75 ms and maximum 612.007 ms.
- Preserves exact command/device/agent/organization/tenant ownership, generation/desired-state checks, independent cleanup evidence, the frozen PAM ledger, the ordinary result transaction input, and WebSocket behavior. No command-ID bypass, alternate unauthenticated path, or ownership relaxation was introduced.
- Remains stacked on Track D PR #4035. Keep this stack until Track D merges; then retarget #4105 to `main` and re-evaluate its exact diff and CI.

## Linked issues and decisions

- Refs #4060, the canonical S0 tracker.
- Closes #4196 when this PR merges; the rc.3 lab verified crash cleanup and subsequent admission on the same endpoint.
- Closure decisions: `docs/superpowers/specs/2026-08-29-s0-track-e-pam-closure-decisions.md`.

## Exact state

- Documentation closure HEAD: `ee7df0b1b`, pushed and synchronized.
- Exact shipping-artifact candidate: `a11d432a6e68288eceef050933483c231f5d40dc` (`v0.109.0-rc.3`).
- The candidate-to-HEAD delta is exactly two documentation commits (`638551d33`, `ee7df0b1b`) touching only the Track E plan and closure decision record; no shipping artifact changed.
- Exact-candidate core workflow: [33231300371](https://github.com/LanternOps/breeze/actions/runs/33231300371), 41 successful jobs and only the expected skipped Main Red Alert.
- The three checks attached directly to the current PR HEAD are two binary-source smoke checks and Cloudflare Pages. They are not the core workflow.

## Exact-candidate evidence

- Signed hosted-gap Windows agent SHA-256: `9cf3bcb4781ac8319b7d2018b610743932b1d321e5d16f75f6f4bafb09c99be7`.
[...]

PR #4197, opened 2026-08-29
fix(agent): let the pinned config directory accept atomic renames (backport #4184)

### LINKED ISSUE #4184: [Agent] Windows: pinned config-directory handle (FILE_SHARE_READ) blocks every atomic rename in C:\ProgramData\Breeze — agent.state never persists, PAM v2 ledger fails closed
## Description
On Windows the agent pins its main config directory for its whole lifetime via `MainAgentLockDirectory` (`openMainAgentDirectoryPinned`, `agent/internal/config/permissions_windows.go` ~L199) with share mode **`FILE_SHARE_READ` only**, and pins the `run` child the same way (`openMainAgentChildDirectory`, ~L219). While that handle is held, every rename of a file *into* `C:\ProgramData\Breeze` fails with `ERROR_SHARING_VIOLATION` ("The process cannot access the file because it is being used by another process"). File creation still works, so the failure only surfaces in atomic write-temp-then-rename paths.

Observed on two Server 2022 lab endpoints running the signed `0.109.0-rc.1` agent (`d473815e`):
- `agent.state` is **never written**: 43 `rename state file after 4 attempts` warnings in 25 minutes (startup, heartbeat, stopping); the file does not exist on disk. The watchdog survives only because it falls back to IPC `state_sync`.
- PAM v2 (`pamlifetime.Store.persistLocked` → `replaceFile`) cannot persist `pam-lifetime-ledger.json`, so `PrepareApply`/`PrepareCleanup` return an error and the manager reports **`invalid_command`** for every `pam_apply_v2` and `pam_cleanup_v2`. RMM-QA-445 physical evidence cannot be collected on this candidate as-is.

## Root Cause (verified)
Bisection with a diagnostic that reproduces `Store.PrepareApply` and a raw temp-create + `os.Rename`:
- agent running → 0/5 renames succeed in `C:\ProgramData\Breeze`; agent stopped → 5/5. Watchdog irrelevant. Sibling dirs (`C:\ProgramData`, `C:\ProgramData\Breeze\data`) unaffected. Not Defender (exclusion made no difference; reverted).
- `handle64 -a C:\ProgramData\Breeze` shows `breeze-agent.exe` holding directory handles on `C:\ProgramData\Breeze` and `...\run` (plus `agent.lock`, `agent.log`, `audit.jsonl`).
- Holding a directory handle with the agent's exact access mask and varying only the share mode, child renames: `0x1` (READ) → 0/5, `0x3` (READ|WRITE) → 5/5, `0x5` (READ|DELETE) → 0/5, `0x7` → 5/5. **`FILE_SHARE_WRITE` on the pinned parent is what child renames need; `FILE_SHARE_DELETE` is not.**

Introduced by #2520 (`e4989eab7`, 2026-07-15, "Windows helper lifecycle durability"), present in every release from **v0.96.0 through v0.108.0** (git tag containment; runtime verification done on 0.109.0-rc.1).

## Proposed Fix
Open the pinned handles with `FILE_SHARE_READ|FILE_SHARE_WRITE` (main directory and `run`). Leaving `FILE_SHARE_DELETE` out preserves the anti-tamper intent (the directory itself still cannot be deleted/renamed while pinned) while allowing the agent's own atomic renames. Add a Windows test that pins the directory the way the agent does and asserts a temp-create + rename in it succeeds; and a `pamlifetime` store test on Windows for the same. Consider surfacing the underlying persist error in the PAM failure code instead of `invalid_command`.

## Affected Files
- `agent/internal/config/permissions_windows.go` (primary: `openMainAgentDirectoryPinned`, `openMainAgentChildDirectory`)
- `agent/internal/state/state.go` (symptom: `renameReplace` retries cannot succeed)
- `agent/internal/pamlifetime/store.go`, `store_replace_windows.go` (symptom: ledger persist)
- `agent/internal/config/config.go` `SaveTo` (`os.Rename` into the same dir — not exercised in the lab; likely affected)

## Reported By
Internal — found during RMM-QA-445 exact-candidate lab work, see #4060.

ToddHebebrand:
## What

Backport of the issue #4184 fix to `main`, restricted to `agent/internal/config`.

`openMainAgentDirectoryPinned` and `openMainAgentChildDirectory` pinned `C:\ProgramData\Breeze` and its `run` child with `FILE_SHARE_READ` only. This grants `FILE_SHARE_WRITE` on both pins via a new named constant `MainAgentPinnedShareMode`, and keeps `FILE_SHARE_DELETE` withheld.

## Why

Windows checks the **parent directory's** share mode when a file is renamed **into** it. With the pin held and `FILE_SHARE_WRITE` withheld, every temp-file-plus-rename the agent performs inside its own config directory failed with `ERROR_SHARING_VIOLATION`, while plain file creation kept working — which is why the breakage was silent:

- `agent.state` was never written (`state.Write` → `renameReplace`)
- `config.SaveTo` could not replace `agent.yaml`

**Shipped range: v0.96.0 through v0.108.0.**

Withholding `FILE_SHARE_WRITE` bought no security: anyone with write access to the directory could already create files there, and content trust rests on the PROTECTED `SYSTEM`+`Administrators` DACL these handles verify on every open — not on the share mode. `FILE_SHARE_DELETE` stays withheld, which is the bit that actually carries the anti-tamper intent (#2520): while pinned, neither directory can be renamed or deleted, and a hostile pre-existing delete/rename handle still makes `OpenPreparedMainAgentLockDir` fail closed.

## Windows RED → GREEN control

Package test binary cross-compiled `GOOS=windows GOARCH=amd64` and run on a Windows Server 2022 lab VM (Go test exe only; no agent service touched, nothing written under `C:\ProgramData\Breeze`).

**RED** — built from `main` with the new tests but the pre-fix `FILE_SHARE_READ`-only pin (constant introduced with the old value so the share-mode assertion compiles, call sites unchanged). Reproduced identically on two runs:

| Test | RED | GREEN |
|---|---|---|
| `TestMainAgentPinnedShareModeAllowsWriteAndRefusesDelete` | **FAIL** | PASS |
| `TestPinnedMainAgentDirectoriesAllowAtomicRenamesInside` | **FAIL** | PASS |
| `TestPinnedMainAgentDirectoriesRefuseNamespaceMutation` | PASS | PASS |

RED failure lines:

```
--- FAIL: TestMainAgentPinnedShareModeAllowsWriteAndRefusesDelete (0.00s)
    permissions_windows_test.go:327: pin share mode 0x1 must grant FILE_SHARE_WRITE or renames into the pinned directory fail

--- FAIL: TestPinnedMainAgentDirectoriesAllowAtomicRenamesInside (0.00s)
    permissions_windows_test.go:372: pinned config directory: rename into pinned directory: rename ...\.pin-rename-700320239 ...\agent.state: The process cannot access the file because it is being used by another process.
```

`TestPinnedMainAgentDirectoriesRefuseNamespaceMutation` passes in both states by design — it is the security guard, not the discriminator. The two tests that flip are the ones that discriminate.

**GREEN** — full `internal/config` package binary run on the same VM: `PASS`.

## Verified vs not

**Verified (read on `main`, not assumed from the source branch):**

- `agent.state` really is written by rename-into-`ConfigDir`. `agentapp/main.go:546,719,1005` call `state.Write(state.PathInDir(config.ConfigDir()), …)`; `state.Write` writes `path + ".tmp"` then `renameStateFile(tmp, path)`, which on Windows is `renameReplace`. The pin is held for the agent's lifetime by `fileMainAgentGuard` via `agentapp/main_instance_windows.go:145` → `config.OpenPreparedMainAgentLockDir()`. So the defect does affect `main`.
- **The real rename mechanism, not just the one the test uses.** The committed regression test exercises `os.Rename` (MoveFileEx), but `agent.state` actually uses `renameReplace` (`SetFileInformationByHandle` / `FileRenameInfoEx` with POSIX semantics) — a different call. A throwaway probe (built, run on the VM, then deleted — not part of this PR) pinned a temp directory both ways and ran both primitives:

  ```
  PRE-FIX  pin=FILE_SHARE_READ      : create temp (plain create): ok
  PRE-FIX  pin=FILE_SHARE_READ      : renameReplace (FileRenameInfoEx, the real agent.state path): FAILED ... used by another process
  PRE-FIX  pin=FILE_SHARE_READ      : os.Rename (MoveFileEx, what the committed test uses):     FAILED ... used by another process
  POST-FIX pin=FILE_SHARE_READ|WRITE: create temp (plain create): ok
  POST-FIX pin=FILE_SHARE_READ|WRITE: renameReplace (FileRenameInfoEx, the real agent.state path): ok
  POST-FIX pin=FILE_SHARE_READ|WRITE: os.Rename (MoveFileEx, what the committed test uses):     ok
  ```

  Both primitives are gated by the parent pin and both are fixed. Plain create succeeds in both states, confirming the "silent" failure mode.

- `GOOS=windows go vet ./internal/config/` — clean (rc=0).
- `go test ./internal/config/...` on the host — `ok` (25.7s).

**Not verified / out of scope:**

- No end-to-end run of a real agent against a real `C:\ProgramData\Breeze`. The lab VM runs a live signed agent service that was deliberately left untouched; all tests pin temp directories only. The evidence is mechanism-level, not a field repro of `agent.state` reappearing.
- `config.SaveTo` replacing `agent.yaml` is **inferred** from the shared rename-into-`ConfigDir` mechanism; it was not separately exercised.
- `GOOS=windows go vet ./internal/agentapp/` reports `service_windows.go:263:46: possible misuse of unsafe.Pointer`. **Pre-existing** — byte-identical output from unmodified `main`, and this PR touches no `agentapp` file.
- `TestPersistedServerURLProviderFollowsPromotion` (`serverurl_test.go:137`, unrelated to this change) is a **pre-existing flake**: it failed 1 of 8 full-suite runs of the *pristine `main`* binary on this VM. Flagging it, not fixing it here.

## Relationship to #4105

This is `fb1731e97` from `fix/s0-pam-actuation-lifecycle` with only its `agent/internal/config` half applied. The `agent/internal/pamlifetime` changes (ledger `ledger_persist_failed` / `ledger_unavailable` failure codes) stay on **PR #4105** — that package does not exist on `main`.

Both files here are **byte-identical** to their `fb1731e97` versions, so the branches converge without conflict when #4105 lands. One consequence worth flagging for review: a comment in the test file still names the PAM v2 ledger as a third caller of this write pattern. That is accurate history for the shipped defect rather than a reference to code on `main`, and it was left as-is to preserve byte-identity.

Fixes #4184

https://claude.ai/code/session_01L4i4e4EXsMqdZ2QsSv4BVt


ToddHebebrand:
Synchronized with current `main` for review at `d5e2da79ceeaf10a6938290b7443628a2034d80c`.

- Base observed during sync: `b73a9ad794c3b4983d4e54818abdba804cbf2fc8`
- Scope diff remains exactly `agent/internal/config/permissions_windows.go` and `permissions_windows_test.go`.
- Fresh local gates: host `go test ./internal/config` passed; Windows amd64 package-test cross-compile passed; `git diff --check` passed.
- The attached `Check Migrations` failure is inherited from `main`, not introduced by this two-file diff. Triage found that `v0.109.0-rc.3` is a non-ancestor Track E tag containing eight Track D/E migrations absent from main; the unchanged guard passes against main's latest ancestral release `v0.108.0`. The release-lineage decision is tracked separately in #4227.

This PR is ready for code review but is **not authorized to merge** while #4227 remains unresolved and main's migration gate remains red.

ToddHebebrand:
Independent boundary review completed clean at `d5e2da79ceeaf10a6938290b7443628a2034d80c`; no findings.

Reviewed the complete two-file scope for Windows share-mode correctness, namespace anti-tamper preservation, call-site consistency, and regression-test adequacy. The reviewer independently confirmed:

- `FILE_SHARE_WRITE` enables child rename operations while `FILE_SHARE_DELETE` remains withheld from pinned directories;
- both absolute and handle-relative directory opens use the same frozen share-mode constant;
- tests cover config/run atomic renames and blocked config/run namespace mutation;
- Windows amd64 test-binary cross-compile, Windows `go vet ./internal/config`, host `go test -race ./internal/config/...`, and `git diff --check` pass;
- the base/head scope remains exactly the two intended `agent/internal/config` files.

This consumes the implementation's single independent review. The inherited migration gate remains separately blocked by #4227; no merge is authorized here.

Issue #3206, opened 2026-08-07
Ticketing and time entry in the mobile app


ToddHebebrand:
## Description

Field technicians cannot work tickets or log time from the mobile app. The backend for both is mature; the gap is entirely client-side.

## Current state

**Mobile ships two tabs** (`apps/mobile/src/navigation/MainNavigator.tsx`): Home (AI chat with streaming SSE, rich blocks, voice input) and Systems (fleet overview -> AlertDetail, DeviceDetail). Plus auth stack, onboarding, biometric app-lock, and hardware-signed approvals. v1.0.0, iOS `com.breeze.rmm`, APNs production live, TestFlight stage, Android push not wired.

`apps/api/src/routes/mobile.ts` is thin — 12 routes (notification register/unregister, devices list/get/settings/delete, device actions, alert inbox/ack/resolve, summary, search) plus `/mobile/approvals` and core `/ai/sessions`.

**Ticketing backend is complete.** `tickets`, `ticket_comments`, `ticket_categories`, `ticket_alert_links`, `ticket_statuses`/`ticket_priorities`, `ticket_forms`, `ticket_response_templates`, `ticket_mailbox_*`. Routes under `apps/api/src/routes/tickets/`, services `ticketService.ts`, `ticketSla.ts`, `ticketTriage.ts`, `ticketEvents.ts`, `aiTicketDraft.ts`. No new backend needed for the core flow.

**Time entry exists and reaches invoices.** `time_entries` (`2026-06-12-a-ticketing-time-parts.sql`) has `started_at`/`ended_at`/`duration_minutes`/`is_billable`/`hourly_rate`/`billing_status` (`not_billed|billed|no_charge|contract`)/`is_approved`, with a **DB-enforced single running timer per user** (partial unique index). Endpoints: `GET /time-entries/running`, `POST /start`, `POST /stop`, `POST /bulk-approve`, `GET /timesheet`, plus `GET /tickets/:id/time-entries`. `services/invoiceAssembly.ts` pulls unbilled billable time and parts into `invoice_lines` (`source_type` = `time_entry|part`), and contracts consume `billing_status='contract'`. It is not an isolated silo.

[...]

PR #4235, opened 2026-08-30
feat(mobile): Time-entry API client and running-timer state (#3206 W03)

### LINKED ISSUE #3897: W03: Time-entry API client and running-timer state
Wave W03 of #3206 — Time-entry API client and running-timer state
Branch: `feature/3206-mobile-ticketing-time-entry/wave-3897`

ToddHebebrand:
## What / why

Wave W03 of the mobile time-entry plan (`docs/superpowers/plans/mobile/2026-08-23-mobile-time-entry.md`). Client-side only — no API, schema or migration changes. The phone calls the existing core `/api/v1/time-entries` routes with the token it already holds, the same way `services/tickets.ts` does for tickets.

## Tasks covered

- **Task 2 — `apps/mobile/src/services/timeEntries.ts`**: typed client (`getRunningTimer`, `startTimer`, `stopTimer`, `createTimeEntry`, `getTimesheet`) plus `TimeEntryError` carrying the server's `code`/`status`. Responses are narrowed to the fields the app consumes; undefined keys are dropped before send (Zod `.optional()` rejects explicit null); `currency` and `hourlyRate` are never sent.
- **Task 4 — `apps/mobile/src/store/timeSlice.ts`**: pure reducers (`timerStarted`, `timerStopped`, `queueDepthChanged`, `timerErrored`) and `elapsedSeconds` (clamped at zero for clock skew). `queueDepthChanged` takes a plain number — Task 3's `QueuedWrite` lands in the next wave and is not imported here.

Deviation from the plan snippet (verified): `services/api.ts` rejects with a plain `ApiError = { message, code, statusCode }` object, not the `{ status, body: { code } }` shape the plan's test used. `asTimeEntryError` handles both and a test pins the real shape.

## Tests

- verified: `pnpm --filter breeze-mobile test` — 54 files, 651 passed / 3 skipped (19 new tests across the two suites, written red-first).
- verified: `pnpm --filter breeze-mobile typecheck` — clean.

## Not checked

- not-checked: on-device / TestFlight behaviour against a live API.
- not-checked: role grants — per the W02 gate, the default Partner Technician role has `time_entries:read` but not `time_entries:write`, so `startTimer`/`stopTimer` succeed only for Partner Admin or custom roles until a grant decision is made. This wave does not depend on the AI path.

Closes #3897


ToddHebebrand:

No confirmed defects (0 findings reached 2+ independent confirmations). Suggestions were applied only where trivial and behaviour-preserving.

### Fixed (verified: typecheck clean, 653 passed / 3 skipped)
- `getRunningTimer` now uses `== null`, so an empty 2xx body (`coreRequest` resolves `{}`) reads as "no timer running" instead of a `TypeError` re-labelled as `TimeEntryError`. New test pins it.
- `TimeEntryError` now also exposes `statusCode` (alias of `status`), matching the `ApiError` / `mobileDevices` / `connectedApps` idiom. Test asserts both.
- ENTRY_BILLED test now asserts `{ code: 'ENTRY_BILLED', status: 409 }`, not just `instanceOf`.
- Added a sparse-payload test pinning `narrowTimeEntry` defaults (`billingStatus: 'not_billed'`, `isBillable: false`, `isApproved: false`, `endedAt`/`durationMinutes` null) so the fallback branches are exercised.

### Declined (behaviour-changing or design-level; deferred to the UI waves)
- Vacuous `not.toHaveProperty('currency')` assertion / `compact` vs `JSON.stringify`: making `compact` an allowlist is a design change; the type already forbids `currency`/`hourlyRate`. Left as-is.
- Throwing / Sentry-capturing on missing `isBillable`/`isApproved`/`billingStatus`: behaviour change; server always emits these today (inferred from `entrySelection` in timeEntryService.ts). Revisit if a screen needs loud contract-drift detection.
- `elapsedSeconds` logging an unparseable `startedAt`: behaviour change, needs Sentry wiring that the mobile app does not have yet (not-checked beyond memory note).
- `ServerTimeEntry` typed as `Record<string, unknown>`: type-only but requires casts through `narrow*`; not trivial. Noted for follow-up.
- `timerStarted` accepting only `RunningTimer` / refusing `endedAt !== null`: reducer matches the plan verbatim; changing the contract belongs in Wave 5 when the caller exists.
- Timesheet UTC day-buckets: not introduced by this PR; consistent with web. Note for the UI wave only.

### Refuted
- **timeSlice escapes `withLogoutReset`**: the slice is pure functions + `initialTimeState`; no store, singleton, or consumer exists in this PR (verified by grep), so no `TimeState.running` survives logout. Valid design note for Wave 5/6.
- **`asTimeEntryError` erases transport signal the offline queue needs**: plan Task 5 enqueues only when offline before sending; drain (Task 3) keys on `status` 4xx, which the wrapper preserves. No double-send path exists.
- **`weekStart` date-only contract unguarded**: no caller outside the test; pass-through matches the plan and web.
- **Catch blocks re-label in-try programming errors**: trigger required the server to omit `data`, which the route never does (verified `c.json({ data: entry })`); the empty-body path is now handled anyway (see Fixed).
- **No test pins transport / `session_superseded` mapping**: wrapper preserves `code`; nothing in this PR consumes `status` for retry decisions. Coverage preference, not a defect.
- **`TimeEntryError.code` is a bare string**: no literal comparisons exist in this PR; hypothetical Task 5 typo.
- **Org-scoped tokens 403 on every call**: deliberate server boundary (partner/system scope only, spec D4) shared with web; backend changes are out of scope per the plan. Client already maps the 403 to `TimeEntryError`.

### Strengths
- API contract is accurate against the server: paths (`/time-entries/running`, `/start`, `/stop`, `/`, `/timesheet`), the `{ data: ... }` envelope, the 201 on start/create, and the error codes named in the doc comment all match `apps/api/src/routes/timeEntries/timeEntries.ts:134-216` and `TimeEntryServiceErrorCode` (`timeEntryService.ts:10-28`). (verified)
- Honours the multi-currency contract: `currency` is never sent on any write, and `hourlyRate` is deliberately absent from `CreateTimeEntryInput` (`timeEntries.ts:11-13, 37-44`), with a test asserting the absence. (verified)
- No backend, schema or migration changes; nothing touches tenancy, RLS or any cascade/export-policy registration list. (verified)
- Tests mock the transport (`vi.mock('./api')`) per the plan and `tickets.test.ts` convention, and cover the real `{message, code, statusCode}` shape `coreRequest` actually throws, not only the plan draft's `{body:{code}}` shape. (verified)
- Test files are co-located with sources per CLAUDE.md; `pnpm --filter breeze-mobile typecheck` is clean and the suite passes (653 passed, 3 skipped on the fix commit). The earlier `hardwareSigner.test.ts` 5s-timeout failure did not reproduce on this run; it is a pre-existing flake in a file this diff does not touch (inferred, not bisected).

Issue #4124, opened 2026-08-27
Register the event-dispatch maintenance repeatables unconditionally


ToddHebebrand:
Follow-up from wave 3.5c (PR #4117, final review recommendation).

The receipt-retention and shadow-comparison repeatables (queue `event-dispatch-maintenance`) register only when the main event-dispatch worker starts (mode != off, or backlog > 0). Documented consequence: after a rollout ends (`EVENT_DISPATCH_MODE` back to `off`, queue drained), residual `event_delivery_receipts` rows only age out the next time the worker starts.

Simpler shape rejected during the wave for scope reasons but recommended by the final review: register the maintenance worker + repeatables **unconditionally** at boot. Both job bodies are cheap and self-gating (the shadow sweep re-checks mode at run time; retention on an empty table is three fast index scans), and the registration failure domain is already isolated so it cannot pin `/ready`.

🤖 Filed by Claude Code from the wave-3.5c review.

PR #4238, opened 2026-08-30
fix(api): surface per-command Redis errors from the shadow-mode multi() pipeline (#4125)

### LINKED ISSUE #4125: Shadow-mode multi() pipeline swallows per-command Redis errors
Follow-up from wave 3.5c (PR #4117, final fix-wave re-review).

`recordShadowLocalInvocation` (`services/eventDispatchQueue.ts`) coalesces its HINCRBY + HSET + EXPIRE into one `multi()`. With ioredis, `pipeline.exec()` only rejects on transport-level failure; an individual command's error (e.g. WRONGTYPE on one key) resolves as an `[err, result]` tuple and never reaches the caller's `.catch` → `console.warn` at the eventBus call sites. Before the coalescing, each command was awaited discretely and any failure warned.

Blast radius: shadow-mode comparison bookkeeping only (best-effort by design; delivery unaffected). Fix: inspect `exec()`'s result tuples and warn on any per-command error. A few lines + one test planting a per-command error tuple.

ToddHebebrand:
## Problem

`recordShadowLocalInvocation` (`apps/api/src/services/eventDispatchQueue.ts`) coalesces its HINCRBY + HSET + EXPIRE into one `multi()` — a cost trim from the wave-3.5c final review (#4085), since it runs once per local subscriber invocation at full event volume.

With ioredis, `pipeline.exec()` only **rejects** on a transport-level failure. An individual command's error (WRONGTYPE against a shadow key, an OOM eviction refusal, …) resolves as an `[error, result]` tuple, so it never reached the caller's `.catch()` → `console.warn` at the eventBus call sites. Before the coalescing each command was awaited discretely and any failure surfaced there; coalescing silently dropped that signal.

Net effect: shadow-comparison counters could be quietly wrong — and the shadow-comparison job would then report mismatches with no upstream breadcrumb explaining them.

## Fix

Inspect `exec()`'s result tuples:

- **Per-command errors** → one structured `console.warn` line, `EVENT_DISPATCH_SHADOW_COMMAND_FAILED`, carrying `eventId` / `eventType` / `orgId` / `subscriberId` / `outcome` and **every** failing command index in a single line (not one line per failure).
- **`exec()` resolving `null`** (the MULTI was discarded — nothing ran, a strictly worse outcome than one failed command) → its own `EVENT_DISPATCH_SHADOW_PIPELINE_DISCARDED` line.

Deliberately **log-only and non-throwing**: this is shadow-comparison bookkeeping, best-effort by design, and must never break the delivery path it observes. Also deliberately **no `captureException`** — at one call per local subscriber invocation, a persistently-broken key would be one Sentry issue per published event; the greppable structured lines are the intended signal. The transport-failure rejection path is unchanged and still propagates to the eventBus `.catch()`.

Scope is two files, both under `services/eventDispatchQueue*`. Nothing in the worker-registration path is touched, so this does not collide with the sibling issue #4124.

## Tests

Four new cases in `apps/api/src/services/eventDispatchQueue.test.ts`, written red-first against the unfixed code (3 failed / 15 passed before the fix):

- a planted per-command error tuple warns, and the call still **resolves** (that resolution is the bug — it proves the warning has to come from inside this module)
- multiple failing commands are reported as indices `[0, 2]` in **one** line
- an all-success pipeline stays silent (control)
- `exec()` resolving `null` warns with the discarded errorId

## Verification

- `vitest run src/services/eventDispatchQueue.test.ts src/services/eventBus.test.ts src/services/eventSubscriberRegistry.test.ts src/jobs/eventDispatchWorker.test.ts` → **4 files, 89 tests passed**
- `pnpm exec tsc --noEmit --project apps/api/tsconfig.json` → clean (exit 0)
- `eslint` on both changed files → clean

Closes #4125


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — four reviewers in parallel: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `comment-analyzer`.

**Findings:** 8 raised (0 critical / 3 important / 5 suggestions) → 6 addressed in 42f4b76, 2 consciously declined; 0 outstanding.

Important, all three fixed:
- **Dropped stack trace** (`code-reviewer`) — error serialization used a flattened `` `${name}: ${message}` `` string, deviating from this file's own `enqueue-failed` log and eventBus.ts's `local-handler-failed`, which both emit `{ name, message, stack }`. For the WRONGTYPE-on-a-shadow-key case this fix exists to catch, the stack is the only evidence of *where* the bad key type was introduced. Restored to the file's convention.
- **Comment overclaimed ioredis semantics** (`comment-analyzer`) — "`exec()` only REJECTS on transport-level failure" is not the whole contract: an **EXECABORT** (a command Redis refuses at queue time) also rejects. Inert for these three fixed, well-formed calls, but it was stated as a general fact a future maintainer would rely on to rule out "one of our commands is malformed". Reworded to name both rejection causes and contrast them against the runtime errors that resolve as tuples.
- **Self-contradicting rationale** (`comment-analyzer`) — the no-`captureException` note said "once per local subscriber invocation" and then "one Sentry event per published event". An event with N local subscribers calls this N times, so the volume claim understated itself. Now tracks subscriber invocations.

Suggestions taken:
- **Failures now name the command**, not just the index (`silent-failure-hunter`) — `hincrby:count` / `hset:detail` / `expire:detail`. The pipeline is 1 **or** 3 commands long depending on sampling, so a bare index was ambiguous on an on-call pager and required reading this function's source to decode.
- **New test for the non-sampled, single-command pipeline** (`silent-failure-hunter`) — every prior new case used `alert.triggered` (100% sampled, 3 commands), leaving the shorter shape unasserted.
- **Pinned the deliberate no-`captureException` choice with an assertion** (`pr-test-analyzer`) so a future "helpful" fix has to argue with a failing test rather than an absence. This needed an explicit `mockClear()`: the sentry mock is module-scoped and is *not* reset between tests in this file — worth knowing before adding further `captureException` assertions here.
- **Documented that the `null`-discard branch is unreachable today** (`comment-analyzer`) — this pipeline never `WATCH`es — so a reader doesn't go hunting for how it fires.

Declined, with reasons: testing the non-`Error` serialization branch (real ioredis always rejects tuple errors as `Error` instances — the reviewer rated it low value itself), and widening scope to the other `redis.multi()` call sites in the repo (`rate-limit.ts`, `tokenRevocation.ts`, …) — those read result tuples for their values and are a separate concern from #4125.

Worth recording from `pr-test-analyzer`: the "stays quiet when every queued command succeeded" case is **not** discriminating for this bug (the old code emitted no warning either way). It is kept deliberately as a guard against the opposite regression — over-firing warnings on success — not as proof the fix works. The three red-first cases are the ones that prove it.

**Tests:** `vitest run src/services/eventDispatchQueue.test.ts src/services/eventBus.test.ts src/services/eventSubscriberRegistry.test.ts src/jobs/eventDispatchWorker.test.ts` → **4 files, 90 passed**. `tsc --noEmit --project apps/api/tsconfig.json` → clean (exit 0). `eslint` on both changed files → clean. Red-first was verified before the fix: 3 failed / 15 passed against the unfixed source.

**CI (first commit a8a6523):** `Test API`, `Type Check`, `Lint`, all four `Integration Tests` shards, and `CI Success` green. Two red jobs, **both inherited from main, neither caused by this PR** — this branch changes zero migration files and zero Dockerfiles:
- `Check Migrations` — the immutability guard diffs against tag `v0.109.0-rc.3`, an off-main rc whose eight `2026-09-1x-*.sql` migrations don't exist on `main`, so it reports them as deleted. **`main` is red on this same job at this PR's exact base commit `355b3746`.**
- `Trivy Image Scan` — Alpine base-image CVE-2026-14456 (`libcrypto3`/`libssl3` 3.5.7-r0, needs 3.5.8-r0) in the M365 Graph Read Executor image. The security workflow is failing on all three most recent `main` runs.

CI is re-running on `42f4b76`; the local evidence above is for that SHA.

**Status:** review-clean, awaiting maintainer merge. Not merged and issue #4125 left open by design — both are the maintainer's call.

Issue #3261, opened 2026-08-08
[UI] Discovered-asset linking: manual unlink self-destructs on next scan, link state illegible, picker unscoped


ToddHebebrand:
## Description

Discovered-asset ↔ managed-device linking works (identity assertion, dedupe, alert attribution) but its lifecycle and display are confusing enough that users read it as part of proxy access. Worst: **manual unlink silently undoes itself on the next discovery scan and destroys its own reversibility.**

Full design + recommended scope: `docs/superpowers/specs/monitoring/2026-08-08-asset-link-lifecycle-design.md` (companion to #3199, which fixes the proxy-side conflation).

## Defects

1. **Unlink self-destructs.** `DELETE /discovery/assets/:id/link` clears both `linked_device_id` and `link_source` (`apps/api/src/routes/discovery.ts:1542-1547`). The next scan sees the asset as never-linked, re-matches the same MAC/IP, and re-links it as `auto` (`apps/api/src/jobs/discoveryWorker.ts:875-922`) — and auto-links cannot be unlinked (403 at `discovery.ts:1533-1535`). Net effect: Unlink works once, then the button disappears forever, and the device flickers in/out of the Devices list on scan cadence with no explanation. Nothing records "user said no"; `discoveryWorker.test.ts` has zero coverage of unlink-then-rescan.
2. **Auto-links can't be unlinked at all** — the 2026-06-27 design's manual-only rule made sense without suppression (unlink would just be undone), but with defect 1 fixed it's an arbitrary restriction, and auto-link (MAC/IP match) is the common case.
3. **Link state is illegible.** Discovery list shows a bare green check + device name with no label (`DiscoveredAssetList.tsx:508-513`) — reads as a status, not "linked to X". The asset modal fetches `linkedDeviceName` and renders it zero times (being fixed in #3199's spec). In the Devices list (flag on), linking is expressed purely as row *absence*.
4. **Cross-linking is one-way.** `/devices/network/:id` links to the managed device, but `DeviceDetails.tsx` never mentions linked network assets — no way to discover the relationship from the device side.
[...]

PR #4241, opened 2026-08-30
fix(api): clear discovered_assets.link_source when a device hard-delete detaches the link (#3952)

### LINKED ISSUE #3952: Permanent device delete 500s when discovered_assets was auto-linked (link_source constraint)
Reported by **dunnbeetle** in #bugs (Discord, message 1541490649025482792).

**Report (PII-redacted):**

> Not sure if this is a known bug but I wanted to report just in case (AI helped me write the report and discover)
> "Hey—found a bug on my self-hosted Breeze instance running 0.107.0.
> 
> Permanently deleting a decommissioned device fails with a 500 when it has an auto-linked discovered asset. Breeze tries to clear `discovered_assets.linked_device_id`, but leaves `link_source=auto`, which violates the `discovered_assets_link_source_requires_link` database constraint.
> 
> Expected behavior would be to either clear both fields or remove the linked discovered-asset record during the permanent-delete cascade.
> 
> Relevant log:
> 
> ```text
> DELETE /api/v1/devices/:id/permanent → 500
> 
> UPDATE "discovered_assets"
> SET linked_device_id = NULL
> WHERE linked_device_id = $1
> 
> PostgresError: new row for relation "discovered_assets" violates check constraint
> "discovered_assets_link_source_requires_link"
> ```
> 
> The affected device was linked automatically by discovery, not manually. Repro: let discovery auto-link an asset to a device, decommission it, then attempt permanent deletion."

## Triage analysis

Reproduced in code. DELETE /devices/:id/permanent (apps/api/src/routes/devices/core.ts) cascades deletion inside a transaction. For tables in DEVICE_LINKED_DEVICE_ID_TABLES (which includes discovered_assets), it runs `UPDATE <table> SET linked_device_id = NULL WHERE linked_device_id = :deviceId` (core.ts lines 1443-1445) but never touches the `link_source` column. Migration apps/api/migrations/2026-06-27-discovered-asset-link-source.sql adds a CHECK constraint `discovered_assets_link_source_requires_link` requiring `link_source IS NULL OR linked_device_id IS NOT NULL`. When a discovered asset was auto-linked (link_source='auto'), nulling only linked_device_id violates that constraint, and the transaction throws a Postgres check-constraint error (code 23514). The route's catch block (lines 1465-1476) only special-cases FK violations (code '23503') and returns 409; any other error, including 23514, falls through to `throw err`, surfacing as an unhandled 500 — exactly matching the reported log. The cascade-delete coverage test (cascadeDelete.test.ts) verifies table membership in DEVICE_LINKED_DEVICE_ID_TABLES but has no assertion about link_source, so this gap has no test coverage.

## Affected code

- `apps/api/src/routes/devices/core.ts:1443-1445` — Cascade loop nulls linked_device_id for discovered_assets/network_change_events but never clears link_source
- `apps/api/src/routes/devices/core.ts:1465-1476` — Catch block only handles FK violation code 23503; check-constraint violations (23514) fall through to throw err -> 500
- `apps/api/migrations/2026-06-27-discovered-asset-link-source.sql:19-26` — Adds discovered_assets_link_source_requires_link CHECK (link_source IS NULL OR linked_device_id IS NOT NULL)
- `apps/api/src/db/schema/discovery.ts:45-48,140` — link_source column/enum on discoveredAssets schema
- `apps/api/src/routes/devices/cascadeDelete.test.ts:88-152` — Coverage test checks DEVICE_LINKED_DEVICE_ID_TABLES membership but has no assertion on link_source handling

## Potential fix

In the DEVICE_LINKED_DEVICE_ID_TABLES loop in core.ts, when clearing linked_device_id on discovered_assets also set link_source = NULL in the same UPDATE (e.g. add a table-specific clause, or generalize the loop to null any accompanying provenance column), so the row can never end up with link_source set but linked_device_id NULL.

ToddHebebrand:
## The bug

Permanently deleting a device returned **500** whenever discovery had auto-linked an asset to it. Reported by a self-hosted user on 0.107.0.

The cascade detached the asset with:

```sql
UPDATE discovered_assets SET linked_device_id = NULL WHERE linked_device_id = $1
```

but `apps/api/migrations/2026-06-27-discovered-asset-link-source.sql` adds:

```sql
CHECK (link_source IS NULL OR linked_device_id IS NOT NULL)
```

An auto-linked row carries `link_source = 'auto'`, so nulling the pointer alone leaves *a source without a link*. Postgres raises **23514**, the whole cascade transaction rolls back, and the route — whose catch special-cases only `23503` and `55P03` — rethrows into an unhandled 500.

Manual links were **not** immune — the constraint draws no manual/auto distinction and a manually-linked asset failed identically. `'auto'` is simply what the bug report carried; why no manual-link report arrived is **not established**, so it should not be read as the bug's boundary. (It is specifically *not* an API asymmetry: #3261/#3295 removed the manual-only rule from the unlink route on 2026-08-11, ten days before 0.107.0 was tagged.) The integration test covers both link sources for that reason.

## The fix

Clear both columns in **one** statement, via a per-table registry:

```ts
export const DEVICE_LINK_DEPENDENT_COLUMNS: Readonly<Record<string, readonly string[]>> = {
  discovered_assets: ['link_source'],
};
```

```sql
UPDATE discovered_assets SET linked_device_id = NULL, link_source = NULL WHERE linked_device_id = $1
```

One statement, not two: a `CHECK` is evaluated per row at the end of each statement, so *"null the pointer, then null the source"* would still transit the forbidden state and fail identically.

### Decision: clear `link_source`, do **not** delete the row

Both satisfy the constraint. Detaching is right, for three reasons:

1. **The row is not a child of the device.** A discovered asset describes a network endpoint that exists whether or not Breeze manages it. It carries operator-curated state — `label`, `notes`, `tags`, `approval_status`/`dismissed_at`, type classification, `first_seen_at`. Deleting it would erase that, and the next discovery scan would resurrect the endpoint as a fresh `pending` row with all curation lost.
2. **The list membership already decided this.** `discovered_assets` sits in `DEVICE_LINKED_DEVICE_ID_TABLES` (detach), not in the cascade-delete list. The bug was an *incomplete detach*, not a table in the wrong list. Fixing it by deleting rows would silently reverse a prior design decision.
3. **Repo precedent — every other unlink path already clears both.** `jobs/discoveryWorker.ts` (same-site re-link guard) and `DELETE /discovery/assets/:id/link` both `.set({ linkedDeviceId: null, linkSource: null })`. The cascade was the one outlier.

`auto_link_suppressed_at` is deliberately **untouched**: it records a human's "stop re-linking this" (#3261), not a property of the link, and the constraint does not cover it. Deleting a device says nothing about that preference.

The registry is **per-table**, not a flat column list, because `network_change_events` is also a `linked_device_id` table but has no `link_source` column — blanket-appending the assignment would trade 23514 for **42703** (`undefined_column`), a 500 either way. There is an explicit test for that.

### Not changed, deliberately

No new SQLSTATE branch in the route's catch. A check-constraint violation during the cascade means the cascade is internally inconsistent — a server bug. Mapping it to a retryable 409 would be a lie, since retrying fails identically forever. The global `onError` already Sentry-reports it. The fix removes the cause.

## Cascade-registration audit (`discovered_assets`)

Reported for all four lists per `CLAUDE.md`, including the ones I did not change.

| List | Location | State | Verdict |
|---|---|---|---|
| `CORE_DEVICE_CASCADE_DELETE_TABLES` | `routes/devices/core.ts` | **Not a member** | Correct, unchanged. That loop is `DELETE FROM <t> WHERE device_id = ...`; `discovered_assets` has **no `device_id` column** — its FK is `linked_device_id`. It belongs to `DEVICE_LINKED_DEVICE_ID_TABLES`, which is what this PR fixes. |
| `CORE_DEVICE_ORG_DENORMALIZED_TABLES` | `routes/devices/core.ts` | **Not a member** | Correct, unchanged. That list is a subset of the cascade ∪ detach tables and moveOrg's rewrite loop keys on `WHERE device_id = ...`, which cannot reach a `linked_device_id` table. An asset's `org_id`/`site_id` record where it was **discovered**, not wherever a linked device later moves. `moveOrg.coverage.test.ts` passes unchanged. |
| `CORE_ORG_CASCADE_DELETE_ORDER` | `services/tenantCascade.ts:212` | **Member** | Present, alphabetically placed. Unchanged. |
| `CORE_TENANT_EXPORT_POLICY` | `services/tenantExportPolicyRegistry.ts:163` | **Member** | Present; `link_source` already classified `included`. This PR adds **no column**, so the column-level trigger for this list does not fire. Unchanged. |
| RLS coverage | `rls-coverage.integration.test.ts` | Shape 1 (direct `org_id`), auto-discovered | No allowlist entry needed. Unchanged. |

**No migration.** The constraint is correct; the writer was wrong.

## Tests

Red-first: each assertion below was watched failing before the fix, and re-verified red by emptying `DEVICE_LINK_DEPENDENT_COLUMNS` (3 failures, then green on restore).

- **`services/deviceDeletion.test.ts`** — the detach statement for `discovered_assets` clears `link_source`; the one for `network_change_events` does **not** (the assertion that distinguishes "clear this table's columns" from "clear `link_source` everywhere").
- **`routes/devices/cascadeDelete.test.ts`** —
  - Route-level assertion on the **compiled SQL**: exactly `UPDATE discovered_assets SET linked_device_id = NULL, link_source = NULL WHERE linked_device_id = <id>`, plus proof no `DELETE FROM discovered_assets` is issued.
  - A **migration-derived contract**: any column a `CHECK` in `apps/api/migrations` ties to `linked_device_id` must be registered in `DEVICE_LINK_DEPENDENT_COLUMNS`, or CI fails. This has to read the SQL — a `CHECK` constraint exists **only** in migrations, so no schema-derived test can see it, which is precisely why this shipped. Carries a vacuity guard so a parser that silently matches nothing cannot pass as green, and a documented escape hatch for a future constraint that genuinely does not forbid the detached state.
  - A registry sanity check (every named table/column exists), so a typo becomes a test failure instead of a 42703 raised from inside the delete transaction.
- **`__tests__/integration/deviceDeleteDiscoveredAssetLink.integration.test.ts`** (new, real Postgres) — the reporter's scenario end to end, with a **control test first**: the old statement really is rejected with 23514 against the test database. Without that control, the two regression cases could pass on a schema where the constraint was never applied. Then: an auto-linked asset no longer blocks the delete, the device row is gone, and the asset survives with `linked_device_id`/`link_source` NULL, its curated fields intact, and `auto_link_suppressed_at` still NULL. Same for a manual link.

Closes #3952


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 lenses, parallel).

**Findings:** 4 Important raised → all 4 addressed in `1de5bc84`; 0 outstanding. code-reviewer returned **0 Critical / 0 Important**.

- **comment-analyzer ×2 — an unsupported causal claim, and a stale one.** I had written that manual links "survive only because the manual-link route is also the manual-UNLINK route, and users unlink before deleting." That asserts a cause I never established, and the mechanism it leans on had already been removed: #3261/#3295 dropped the manual-only rule from the unlink route on 2026-08-11, **ten days before 0.107.0 was tagged**. Corrected in `core.ts`, the integration-test header, and the PR body to what is actually known — the constraint draws no manual/auto distinction, `'auto'` is what the report carried, and why no manual-link report arrived is not established.
- **silent-failure-hunter — the generic rethrow dropped every breadcrumb.** It stays a **500** (a cascade defect is not user-retryable, so a 409 would advertise a retry that fails identically forever), but the global `onError` logs a bare `Error:` with no `deviceId` and returns a sanitized body in production — so the fact that an irreversible `SELF_UNINSTALL` had already gone out was unrecoverable server-side. That missing breadcrumb is precisely what the original 0.107.0 report lacked. Now logs `deviceId` + `uninstallSent` + SQLSTATE, then rethrows unchanged. New route test asserts 500 (not 409) and all three logged facts.
- **pr-test-analyzer — the migration-CHECK derivation was under-pinned.** Only the one constraint that exists today held it up, so a parser that stopped recognising a future SQL shape would take the whole contract quietly vacuous with it. Added direct tests for `checkConstraints`/`owningTable` over synthetic SQL: `ALTER TABLE ADD CONSTRAINT`, an inline `CHECK` in `CREATE TABLE` with nested parens and a `public.` qualifier, nearest-table attribution across two tables in one file, and lower-case/quoted identifiers. Its one known blind spot (intersects against the Drizzle schema, so SQL-only column drift escapes it — `db:check-drift`'s job) is now documented in-place rather than assumed impossible.

Accepted-and-declined, stated for the record: I did **not** add a `23514` branch to the route's catch. Mapping a server-side cascade defect to a 409 would misreport it as user-retryable. The fix removes the cause; the log line above closes the diagnosis gap without touching status-code semantics.

**Tests:** `deviceDeletion.test.ts` + `cascadeDelete.test.ts` — **29 passed** (`vitest run --pool=threads`). New real-Postgres suite `deviceDeleteDiscoveredAssetLink.integration.test.ts` — **3 passed** against a live DB, including the control proving the pre-fix statement really is rejected with 23514. Red-first verified: emptying `DEVICE_LINK_DEPENDENT_COLUMNS` turns 3 assertions red across both unit suites, green on restore. `eslint` clean on all five changed files.

**Typecheck:** CI **Type Check: pass**. (Local `tsc --noEmit` for `@breeze/api` OOM'd and then timed out on this loaded host — the known API-tsc memory issue, not a result. CI is the signal here.)

**CI:** on the previous head `c2adb02f`, Test API, Type Check, Lint and **all four Integration Tests shards** passed; re-running for `1de5bc84`. `Check Migrations` is red — **pre-existing and repo-wide**: it is failing identically on `main` itself (base commit `355b3746`) and on every other open PR (#4237, #4238, #4240). This PR adds no migration.

**Status:** review-clean, awaiting maintainer merge. Not merged, issue #3952 left open and assigned.

Issue #2950, opened 2026-07-30
Config policy list: Features column always renders em-dash; icon-only action buttons unlabeled


ToddHebebrand:
Two pre-existing UI issues on the configuration-policies list page, found during E2E QA of PR #2946:

1. The "Features" column always renders `—` — the list endpoint returns no feature-link data, so the column is dead weight. Either populate it (feature badges are already rendered elsewhere from links) or drop it.
2. The row action buttons are icon-only with no `title`/`aria-label`/`data-testid` — inaccessible, and unaddressable by the repo's data-testid E2E convention.

ToddHebebrand:
Fixed in #2978 (`d2a832301`) — the config-policy Features column is populated and the icon-only row actions are labelled. Closing; the PR used `Refs` rather than `Closes`.

PR #4242, opened 2026-08-30
fix(web): clamp the paginated list page when rows shrink underneath it (#4008)

### LINKED ISSUE #4008: [Web] Paginated lists keep a stale currentPage when rows shrink — empty page, wrong empty-state copy, and no pager to recover

ToddHebebrand:
## Problem

`NotificationChannelList` and `ConfigPolicyList` both keep `currentPage` in local state and never reconciled it with the row count. Deleting the only row on the last page — the parent refetches and hands down a shorter array — left the user on a page that no longer exists:

- an empty list,
- the "try adjusting your search" copy over an **untouched** search box,
- and, because `totalPages` had dropped below the stored page, **no pager rendered at all**.

With no rows there are no row actions either, so the only ways out were typing in the search box, changing a filter, or reloading. Search and filter changes already reset the page, so only the "rows disappeared underneath me" path was affected.

## Fix

Both components render from a clamped `safePage` instead of the stored one, and floor `totalPages` at 1 so an empty list is page 1 of 1 — never page 1 of 0, and never a page index of 0 slicing from a negative offset.

Two deliberate choices worth a reviewer's eye:

- **Clamping during render, not in an effect.** The issue suggested `useEffect(... setCurrentPage(Math.min(...)))`. That works, but it commits one frame of the dead page before correcting, so the user sees the empty state flash. Deriving `safePage` during render means the bad page never paints.
- **A companion effect still retires the out-of-range value.** Without it the clamp is cosmetic: `currentPage` stays at 3, and a later create + refetch that grows the list back past the stored page would teleport the user forward to a page they had already been bounced off. It renders the same output either way, so it costs a state write and no visible frame.

Clamping lands the user on the **last page that still exists** rather than resetting to page 1, which keeps them as close to where they were as the new row count allows.

## Second finding in the same markup

`NotificationChannelList`'s pager buttons had no accessible name. They held only a `ChevronLeft`/`ChevronRight`, and lucide-react stamps `aria-hidden="true"` on an icon with no children and no `aria-`, `role` or `title` prop — the mechanism recorded in #3697 — so both controls were absent from the accessibility tree entirely and unreachable by name. Now labelled with the existing `common:actions.previousPage`/`nextPage` keys (pager nouns, not the navigation verbs in `actions.back`/`next`), which are present and translated in all eight locales, so **no new strings**. `ConfigPolicyList` already had these names from #2950.

The stale comments in both files that acknowledged this case as uncovered are replaced, since it is now covered.

## Verification

Seven new assertions across two co-located suites, **each confirmed red against the unfixed components** (component edits reverted, suites re-run):

- shrink cases fail with the reported empty page and the literal `No policies found. Try adjusting your search.` copy;
- the a11y case fails with `Unable to find an accessible element with the role "button" and name /next page/i`.

Green after the fix, with the pre-existing suites for both components named explicitly (a scoped directory path would miss a sibling `Foo.test.tsx`):

```
NotificationChannelList.pagination / .emptyState / .redactedConfig / .testVerdict
NotificationChannelsPage
ConfigPolicyList.pagination / ConfigPolicyList / .emptyState
ConfigurationPoliciesPage
→ 9 files, 60 tests passed
```

`tsc --noEmit` on `@breeze/web`: exit 0. `eslint` on the four touched files: exit 0.

## Not in scope

Roughly twenty other components share this `currentPage` state machine (`PatchList`, `ScriptList`, `AutomationList`, `DeviceList`, `AuditLogViewer`, …). The issue flags the sweep as follow-up; this PR fixes the two components it names. `DeviceIpHistoryTab` already clamps via an effect, but `setCurrentPage(totalPages)` there can write a page of 0 on an empty list — worth folding into that sweep.

Closes #4008


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter (4 agents, parallel). type-design-analyzer skipped: no new types.

**Findings:** 4 raised → 3 addressed in a8a5b45, 1 declined with reasoning; 0 outstanding.

- **comment-analyzer (critical) — addressed.** The a11y comment claimed the unnamed pager buttons were "absent from the accessibility tree". Wrong and self-contradictory: a native `<button>` stays in the tree regardless of its child icon's `aria-hidden`; what the hidden icon removed was the button's only *naming* source, leaving it present but nameless. Reworded in the component and the test. Also softened the floor-of-1 comment, which billed the negative-index slice as the danger — it is harmless against an empty array; the wrong page **count** is the real defect.
- **pr-test-analyzer (sev 7) — addressed.** The zero-row trough was asserted by comment only. Worth noting the shape of the test that closes it: the empty-state copy alone does **not** discriminate, since an unfixed component renders the same thing (a page-2 slice of `[]` is empty for the same reason a page-1 slice is). The test therefore continues through a *refill*, where an unfloored `safePage` of 0 slices from a negative index and the list comes back empty. That assertion is what earns it.
- **pr-test-analyzer (sev 5) — addressed.** The clamp was only exercised against the raw prop length. Both suites now shrink the **filtered** count with a type/status filter engaged, proving `totalPages` follows the filtered rows rather than being propped up by rows the filter excludes.
- **silent-failure-hunter — declined, deliberately.** Unvalidated `pageSize`: `pageSize={0}` yields `totalPages = Infinity`, `NaN` renders "page NaN of NaN". Real, but currently unreachable — the prop is optional with a hardcoded default of 10 and neither production caller passes it. A guard would be untested hardening for a dead path. Worth revisiting if a page-size selector ever lands.
- **code-reviewer — 0 findings.** It also flagged concurrent edits to the worktree mid-review; that was me applying the fixes above, not a foreign process.

Also caught and fixed a vacuous assertion in my own first draft of the zero-row test: `queryByText(/of 0/i)` was matching the header summary ("0 of 0 channels"), not the pager.

**Tests:** all **11** assertions across the two new suites re-confirmed **red against `origin/main`'s components**, then green with the fix — the shrink cases fail with the reported empty page and the literal "No policies found. Try adjusting your search." copy, the a11y case with an unfindable button. Full affected set green locally, files named explicitly (a scoped directory path would miss a sibling `Foo.test.tsx`): 9 files / 64 tests passed. `tsc --noEmit` on `@breeze/web` exit 0; `eslint` on the four touched files exit 0.

**CI on head a8a5b45:** Test Web **pass** (13m25s), Type Check **pass**, Lint **pass**. Test API and the 4 integration shards were still running at the time of writing — none of them touch this diff (web-only, no DB/tenancy/migration surface).

**Status:** review-clean, awaiting maintainer merge. Issue #4008 stays open and assigned until you verify.

ToddHebebrand:
**CI final on head a8a5b45 — every test job green.**

Test Web **pass** (13m25s) · Test API **pass** (17m33s) · Integration Tests **4/4 shards pass** · Type Check **pass** · Lint **pass** · Lint Agent (Go) **pass**.

**Two red checks, both pre-existing and repo-wide — not from this PR.** This diff touches four files, all under `apps/web/src/components`, and no migration:

- **Check Migrations** — red on the last five `main` runs (including `355b3746`, the base this branch was cut from) and on every currently-open PR I sampled: #4240, #4241, #4243, #4244. Note `Check Migrations (non-superuser)` **passes** on all of them, including here.
- **Trivy Image Scan** — likewise red on #4240, #4241, #4243, #4244.

Flagging rather than fixing: both belong to whoever owns the current main breakage, and neither gates the correctness of a web-only render fix. Worth confirming they are already tracked before this merges, since a `--admin` merge would go in over two red checks.

PR #4243, opened 2026-08-30
fix(web): make in-page hash links switch device tabs without a refresh (#4229)

### LINKED ISSUE #4229: "View all activity" updates URL hash but does not navigate until page refresh
### Description

Clicking **View all activity** on a device updates the URL to:

`/devices/<device-id>#activities`

but the page does not switch to the Activity view.

If the page is then refreshed, the Activity view opens correctly.

This suggests the device page does not react to hash changes after initial render/navigation.

### Steps to Reproduce

1. Open any device.
2. Click **View all activity** in the Activity card.
3. Observe that the URL changes to `#activities`.
4. Observe that the visible page does not change.
5. Refresh the browser.
6. The Activity view now opens correctly.

### Expected Behavior

The Activity view should open immediately after clicking **View all activity**, without requiring a page refresh.

### Actual Behavior

Page does not switch to the Activity view.

### Component

Web Dashboard

### Version

_No response_

### Relevant Logs

```shell

```

### Additional Context

_No response_

ToddHebebrand:
Closes #4229

## The bug

Clicking **View all activity** on a device moved the URL to `/devices/<id>#activities` and nothing else happened. Refreshing the page then opened the Activity view correctly.

## Root cause

Not `useHashState` — it already subscribes to `hashchange`, and tab *buttons* work because `switchTab` assigns `window.location.hash` (which fires the event).

It is Astro's `<ClientRouter />`, rendered by `Layout.astro`. Its document-level click listener intercepts **every** same-origin anchor:

```js
ev.preventDefault();
navigate(href);   // astro:transitions/client
```

and `navigate()` short-circuits a link whose path and query match the current URL and which only changes the fragment:

```js
if (samePage(from, to) && !options.formData) {
  if (direction !== "back" && to.hash || ...) {
    moveToLocation(to, from, options, document.title, historyState);
    return;                       // -> history.pushState(..., to.href)
  }
}
```

`history.pushState` fires **neither `hashchange` nor `popstate`**. So the address bar moved to `#activities` while every hash subscriber in the app kept rendering the old value, until a reload re-read the hash on mount and made it look fixed.

## The fix

A small `HashLink` component (`apps/web/src/components/shared/HashLink.tsx`) that renders the same real `<a href="#…">` but, on a plain left click, calls `preventDefault()` and assigns `window.location.hash` itself. That fires a genuine `hashchange`, and the `preventDefault()` is what keeps Astro out of it — its listener skips events whose default is already prevented, and React 18 dispatches at the island root, which is inside `document`, so our handler always runs first.

Modified clicks (cmd/ctrl/shift/alt, non-primary button) are deliberately left alone so open-in-new-tab still works — that is why this stays an anchor with a real `href` rather than becoming a `<button>`.

`window.location.hash` remains the URL-state mechanism per CLAUDE.md. **No migration to query params.**

### Why a shared component rather than a one-line in-component patch

All three in-page hash anchors in `apps/web/src` had the identical defect, so the fix was needed in three places regardless:

| Component | Link | Target tab |
|---|---|---|
| `DeviceActivityFeed` | "View all activity" | `#activities` (the reported bug) |
| `DeviceActivityFeed` | pinned "N active alerts" | `#alerts` |
| `DeviceAlertHistory` | "View all N alerts" | `#alerts` |

The component keeps the modified-click handling and the ClientRouter explanation in one place instead of three, and gives the next hash link somewhere correct to start from. It adds no new state mechanism and no new dependency.

## Tests

New helper `apps/web/src/__tests__/astroClientRouterStandIn.ts` reproduces the ClientRouter interception in jsdom. This matters: jsdom follows anchor defaults, so a naive `fireEvent.click` on the **broken** code fires `hashchange` anyway and the test passes vacuously. With the stand-in installed, the two new navigation cases in `DeviceDetails.hashNavigation.test.tsx` **fail against the pre-fix tree** (verified) and pass after.

- `apps/web/src/components/shared/HashLink.test.tsx` — 6 cases: real `href`, `#`-prefix normalisation, `hashchange` fires under ClientRouter, cmd-click delegated to the browser, caller `onClick` may veto, prop forwarding.
- `apps/web/src/components/devices/DeviceDetails.hashNavigation.test.tsx` — 3 cases: clicking "View all activity" mounts the Activity panel; clicking the pinned alert link mounts the Alerts panel; both links keep their fragment `href`.

### Verification

- `vitest run` over the 9 affected files (`DeviceDetails.*`, `DeviceActivityFeed`, `HashLink`, `useHashState`): **56 passed**.
- Pre-fix control run: the 2 navigation cases fail with "Unable to find an element by: [data-testid=...]" — the tab never switches.
- `eslint` on all touched files: clean.
- Scoped `tsc --noEmit` over the touched files and their import graph: **no errors in any changed file**. (The full `apps/web` project has pre-existing `import.meta.env` / `astro:transitions/client` resolution errors in untouched files; web has no `typecheck` script and CI's Test Web job runs vitest only.)

## Risk

Web-only, no API/schema/tenancy surface. Behaviour change is confined to the three links above.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` on `fca27fbb0` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 lenses, parallel).

**Findings:** 5 raised → all addressed in `acd7c328b`; 0 outstanding.

- **code-reviewer** — 0 critical, 0 important. Independently re-derived the root cause against the installed `astro@7.2.4` source and confirmed the handler, the prop-spread ordering (`href`/`onClick` cannot be overridden), accessibility/i18n parity of the tag swaps, and no scroll-jump/back-forward/SSR regression. Its one action item was procedural (uncommitted work in the reviewer's worktree) — that was the in-flight fix commit, now pushed.
- **comment-analyzer** — 2 critical, both factual errors in my own doc comment, both fixed:
  - said *"React 18"*; `apps/web` is on `react ^19.2.8`, and root-container delegation landed in React 17 — wrong on two counts. Version dropped; the mechanism is now stated as the bubble-path fact it actually is.
  - said Astro intercepts *"EVERY"* same-origin anchor click; it skips `data-astro-reload`, `download`, non-`_self` targets, missing `href`, and modified clicks. Reworded, version-anchored to `astro@7.2.4`, flagged as an unexported internal to re-check on bump, and the `popstate` claim tightened (`moveToLocation` *can* dispatch one, but only when `history.state` is falsy — which its own `pushState` rules out). The same explanation had been duplicated across three files; it now lives once in `HashLink.tsx`.
- **silent-failure-hunter** — 1 finding, currently unreachable: a caller-supplied `onClick` that calls `preventDefault()` returns without navigating. No caller passes `onClick` today. Documented as an explicit prop contract rather than removed — the veto is the standard anchor idiom, and a caller that vetoes owns the feedback. Confirmed 0 findings on same-hash no-op, invalid-tab fallback, SSR safety, and empty/error target tabs.
- **pr-test-analyzer** — 0 critical, 2 important, both closed:
  - `DeviceAlertHistory`'s converted link had no coverage (the DeviceDetails suite mocks the whole component). Added `DeviceAlertHistory.hashNavigation.test.tsx`. **Control run:** reverting that one anchor to a raw `<a>` turns it red — the assertion discriminates.
  - keyboard activation was untested even though "stays a real anchor" is the stated design reason. Added an Enter-key case.

**Red-first evidence:** before the fix, the two navigation cases in `DeviceDetails.hashNavigation.test.tsx` fail with `Unable to find an element by: [data-testid="event-log-viewer-stub"]` — the tab never switches. The stand-in is load-bearing: without it jsdom follows the anchor default and the test would pass against the broken code.

**Tests:** 33 passed across 5 affected files, single-fork (`HashLink`, `DeviceDetails.hashNavigation`, `DeviceAlertHistory.hashNavigation`, `DeviceActivityFeed`, `useHashState`). `eslint` clean on all 7 touched files. Scoped `tsc --noEmit` over the touched files and their import graph: no errors in any changed file.

**CI on `acd7c328b`:** Test Web, Type Check, Lint, Test API and all 4 Integration Tests shards **pass**. Two red checks are **pre-existing on `main`**, verified at this PR's exact base `355b3746`: `Check Migrations` / "Shipped-migration immutability guard" (run 33273323158) and `Trivy Image Scan` (run 33273323133). This PR touches zero migrations and zero dependencies.

**Note for the reviewer:** the third converted link — `DeviceAlertHistory`'s "View all N alerts" — renders only when a caller passes `limit`, and its sole call site (`DeviceDetails.tsx:808`) does not. It is currently unreachable in the product; it carried the identical defect and is fixed and tested pre-emptively, but it is not a user-visible part of this fix. The two user-visible links are both in `DeviceActivityFeed`.

Issue #3946, opened 2026-08-24
metricRollupMaintenance registers attachWorkerObservability twice


ToddHebebrand:
Found while auditing worker-name tag cardinality for PR #3913.

Counting `attachWorkerObservability(...)` call sites across `apps/api/src` gives **44 call sites but only 43 distinct worker names** — `metricRollupMaintenance` is registered twice.

Consequences are mild but real: the worker gets two `failed` listeners, so a single job failure produces two Sentry events once the `worker` tag starts surviving the scrubber (PR #3913 allowlists it). Worth confirming whether the duplicate registration is intentional.

Filed separately rather than folded into #3913, which was already carrying nine review fixes.

ToddHebebrand:
Re-verified 2026-08-30: this is a **false positive**. There is exactly one production registration — `apps/api/src/jobs/metricRollupMaintenance.ts:126`. The second `attachWorkerObservability(worker, 'metricRollupMaintenance')` occurrence is in `apps/api/src/jobs/workerObservability.test.ts:132` (a test constructing its own worker). Across the repo: 43 non-test call sites, 43 distinct worker names — no duplicates. The counting method that produced this issue included test files (it would equally have flagged `desktopSessionFinalizationWorker`, which also appears in that test file). Recommend closing as not-a-bug — leaving that to Todd.

ToddHebebrand:
**Re-verified 2026-08-31** at `origin/main` `3e01e7dcb` — after today's ~25 merges, including #4285 (metric-rollup DB contexts). The false-positive verdict still holds; nothing changed the picture.

**Verified:**
- Exactly **one** production registration: `apps/api/src/jobs/metricRollupMaintenance.ts:126`, inside `initializeMetricRollupMaintenanceWorker()`.
- Repo-wide non-test count is now **53 call sites / 53 distinct worker names — zero duplicates** for any worker, not just this one.
- `initializeMetricRollupMaintenanceWorker` has exactly one production caller: the single `WORKER_REGISTRY` entry at `services/workerRegistry.ts:106`. `index.ts` and `worker.ts` are role-guarded mutually-exclusive entrypoints (each `process.exit(78)` on the wrong `BREEZE_ROLE`), each bootstrapping once, so the registry is iterated once per process.
- It was never a duplicate: at `e31fc7337` (main when this was filed) the non-test count was also 1.

[...]

Issue #4227, opened 2026-08-29
ci(migrations): define release baseline when latest tag is not on branch ancestry


ToddHebebrand:
## Problem

`Check Migrations` is red on `main` because the migration immutability guard selects the highest semantic-version tag even when that tag is not an ancestor of the commit being checked.

The current highest tag, `v0.109.0-rc.3` (`a11d432a6e68288eceef050933483c231f5d40dc`), was cut from the stacked Track E candidate line. It is not an ancestor of current `main` (`b73a9ad794c3b4983d4e54818abdba804cbf2fc8`). The tag contains eight Track D/E migration filenames that do not exist on main, so `scripts/check-migration-immutability.sh` interprets those feature-line migrations as deletions from main.

This is a release-lineage/provenance problem. It is not evidence that main edited the bytes of a migration in its own ancestry, and it must not be normalized by renaming migrations, copying candidate migrations to main out of order, or weakening ownership/immutability checks.

## Reproduction

- Main CI run: `33267501194`
- Failing job: `Check Migrations`
- Candidate tag: `v0.109.0-rc.3` = `a11d432a6e68288eceef050933483c231f5d40dc`
- Current main during triage: `b73a9ad794c3b4983d4e54818abdba804cbf2fc8`
- Merge-base ancestry test: `git merge-base --is-ancestor v0.109.0-rc.3 origin/main` returns non-zero
- Merge base: `92947ea74cec9b42286ac4230723f36b90333ed2`
- Latest release tag that is an ancestor of main: `v0.108.0` (`fe6b01cb9884b550a3c630dafd93f1c45cd216dc`)
- The unchanged guard passes when explicitly run against that ancestral tag:

  ```text
  check-migration-immutability: comparing apps/api/migrations against v0.108.0
  check-migration-immutability: OK
  ```

## Required decision and acceptance criteria

- Define the authoritative migration baseline when the highest release/prerelease tag is not an ancestor of the checked branch.
- Preserve detection of genuinely modified or deleted shipped migrations.
- Fail explicitly on invalid release provenance when appropriate; do not silently choose a convenient baseline if doing so could hide a shipped candidate migration.
- Decide whether draft/prerelease tags cut from stacked feature lines are valid immutable-migration baselines for unrelated branches.
- Add tests for both an ancestral release tag and a higher non-ancestor candidate tag.
- Restore `Check Migrations` on main without changing or casually renumbering existing migration files.

## Scope boundary

[...]

PR #4244, opened 2026-08-30
refactor(api): consolidate the thirteen hand-rolled row-count readers onto db/rowCount (#3894)

### LINKED ISSUE #3894: Consolidate the seven private extractRowCount copies onto db/rowCount.ts
**What:** #3760 created `apps/api/src/db/rowCount.ts` as the canonical postgres-js/drizzle row-count reader (three-shape check, deliberately not null-safe — see its doc comment for why). Six private copies of the same check remain and should be re-pointed at it:

- `apps/api/src/jobs/auditRetention.ts:106`
- `apps/api/src/jobs/deviceMetricsRetention.ts:71`
- `apps/api/src/jobs/processSampleRetention.ts:37`
- `apps/api/src/jobs/serviceProcessCheckRetention.ts:73`
- `apps/api/src/services/metricRollupMaintenance.ts:127`
- `apps/api/src/services/tenantCascade.ts:803`

Plus one inline wrong-shape read worth folding in while there: `apps/api/src/jobs/reportScheduleWorker.ts:506` does `result.rowCount ?? rows.length` — the node-postgres shape the canonical helper's doc comment warns yields 0 on every call under `drizzle-orm/postgres-js`. Verify what that call site actually receives before assuming it's benign; if it is broken, its batched loop terminates early the same way the #3760 lock check reported the opposite of the truth.

**Why it matters:** the copies have already diverged once in a way that shipped a bug (the #3760 lock check read `rowCount ?? rows.length` and got 0 on every call). One helper, one contract, one place to fix.

**Context:** #3760 review round 1 (bounced on a seventh hand-rolled copy), #3769 (same mock-shape failure mode, opposite direction).

ToddHebebrand:
Closes #3894

## What

`apps/api/src/db/rowCount.ts` has been the canonical postgres-js/drizzle row-count reader since #3760, but **fifteen** private re-implementations of the same three-shape check had accumulated. Four were under a *different name*, and two more cast to `Record<string, unknown>` rather than an inline object type — which is why the issue (and a `grep extractRowCount`) only found six.

> **Review round 1** produced two follow-up commits:
> - `f24d0e2` — added the last two copies (`agentLogRetention`, `changeLogRetention`). My first-pass sweep *and* the structural guard I shipped with it both keyed on the inline cast, so both missed them. The guard now keys on the discriminating `typeof x.rowCount === 'number'` test, which is cast-style agnostic.
> - `9b5d834` — corrected three historical claims in the new comments. I had taken them from the issue body without checking them against `git log`; two were wrong and one was unsubstantiated. Details below.

All thirteen now import the canonical helper. Net **−263 / +228**, and `db/rowCount.ts`'s function body is unchanged — only its stale "consolidating them is worthwhile but is not this change" note was rewritten.

## Every call site converted

| # | Call site | Symbol | Origin |
|---|---|---|---|
| 1 | `jobs/auditRetention.ts` | `extractRowCount` (private) | named in #3894 |
| 2 | `jobs/deviceMetricsRetention.ts` | `extractRowCount` | named in #3894 |
| 3 | `jobs/processSampleRetention.ts` | `extractRowCount` | named in #3894 |
| 4 | `jobs/serviceProcessCheckRetention.ts` | `extractRowCount` | named in #3894 |
| 5 | `services/metricRollupMaintenance.ts` | `extractRowCount` | named in #3894 |
| 6 | `services/tenantCascade.ts` | `extractRowCount` (private) | named in #3894 |
| 7 | `jobs/aiUnattendedExposureRetention.ts` | `extractRowCount` | **found by sweep** |
| 8 | `jobs/eventDispatchWorker.ts` | `extractAffectedCount` | **found by sweep** (renamed) |
| 9 | `jobs/mlOutputRetention.ts` | `extractMlOutputRetentionRowCount` | **found by sweep** (renamed) |
| 10 | `jobs/reliabilityRetention.ts` | `extractReliabilityRetentionRowCount` | **found by sweep** (renamed) |
| 11 | `jobs/userRiskRetention.ts` | `extractUserRiskRetentionRowCount` | **found by sweep** (renamed) |
| 12 | `services/orgMergeCustomExecutors.ts` | `extractRowCount` (re-exported into `orgMerge.ts`) | **found by sweep** |
| 13 | `jobs/ipHistoryRetention.ts` | inline `(result as unknown as { count: number }).count ?? result.length ?? 0` | **found by sweep** (not a function) |
| 14 | `jobs/agentLogRetention.ts` | inline, cast to `Record<string, unknown>` | **found by PR review** |
| 15 | `jobs/changeLogRetention.ts` | inline, cast to `Record<string, unknown>` | **found by PR review** |

**Each really was the same check.** #1–#10 were byte-identical to the canonical body. The five that were not identical, and what differed:

- **#12 `orgMergeCustomExecutors`** and **#6 `tenantCascade`** had drifted to `raw?.rowCount` / `raw?.count` — null-*safe*, mapping a nullish result to `0`. `tenantCascade`'s cast also carried a dead `length?: number` member it never read.
- **#13 `ipHistoryRetention`** read `.count` but never `.rowCount`.
- **#14 `agentLogRetention`** and **#15 `changeLogRetention`** were the same `rowCount → count → isArray?length` ladder written as a nested ternary over a `Record<string, unknown>` cast, with a final fallback to the string literal `'unknown'`. Their `deletedCount` therefore narrows from `number | string` to `number`; it only ever reached a `console.log` and neither worker has other consumers, so nothing downstream changes.

The sweep is mechanical, not judgement: every copy performs the same discriminating test, so ``git grep -E "typeof [A-Za-z_$][A-Za-z0-9_$]*\\??\\.rowCount === 'number'"`` enumerates them regardless of cast style. That grep is now a test (below). Repo-wide across `apps`, `ee` and `packages` it matches exactly one file: `db/rowCount.ts` itself.

## Two deliberate behaviour changes

**1. `orgMergeCustomExecutors` + `tenantCascade` lose their null-guard.** The canonical helper is deliberately NOT null-safe, and its doc comment explains why: postgres-js never resolves a successful statement to null/undefined, so a nullish result means a broken driver, adapter or mock. Mapping that to `0` conflates it with "no rows" — and `0` is load-bearing. It terminates the batched retention loops, and in `tenantCascade` it is the **per-table row count written into the GDPR erasure audit**. A silent `0` there is precisely the failure this issue exists to stop. Converging on the strict helper is the point; `tenantCascade`/`orgMerge` suites are green (110 tests).

**2. `ipHistoryRetention` now also handles the node-postgres shape.** No change under postgres-js — that result has no `.rowCount`, so the same `.count` is read.

## `reportScheduleWorker.ts` — verified, NOT a bug, correctly out of scope

The issue asked me to verify rather than assume. **Verdict: it is not the driver-shape bug, and there is no early loop termination.** Evidence, read against this branch:

- The line is now `reportScheduleWorker.ts:552`: `const rowCount = result.rowCount ?? rows.length;`
- `result` is the return of `generateReport(...)`, declared `): Promise<ReportResult>` at `reportGenerationService.ts:577`. `ReportResult` is a **domain type** (`reportGenerationService.ts:26`): `{ rows?: unknown[]; rowCount?: number; summary?; generatedAt?; previous? }`. No postgres-js `Result` ever reaches this line, so `??` is an ordinary optional-field fallback, not a wrong-shape read.
- Every tabular generator populates `rowCount` itself — eleven sites in `reportGenerationService.ts` (`:105 :221 :256 :283 :319 :340 :385 :414 :457 :610 :614`). Executive summary is summary-only and deliberately has none, which is what the `rows.length` fallback records as 0.
- **There is no batched loop.** `rowCount` occurs exactly twice in the worker: computed at `:552`, stored on `report_runs` at `:560`. Nothing branches or terminates on it.

Folding it onto `extractRowCount` would be an active **regression**: the helper reads a top-level `.rowCount`/`.count` and never a nested `result.rows`, so a future generator returning `{ rows: [a, b] }` with no `rowCount` would go from 2 to 0. Same for the identical read at `routes/reports/runs.ts:160`. Both left alone. This matches the conclusion already reached in the issue thread.

## Tests: one contract suite replaces eight duplicated blocks

New: `apps/api/src/db/rowCount.test.ts` (17 tests).

**The deleted assertions were duplicated *and* vacuous.** All 27 removed `expect(...)` calls used object/array literals — `{ count: 5 }`, `{ rowCount: 7, count: 3 }`, `[{}, {}]`, `{}`. I evaluated all 8 distinct fixtures against both the correct helper and the array-before-count ordering that produces the 0-on-every-call failure: **8/8 pass under both.** They could not detect the bug class they were named for.

The new fixture is a faithful postgres-js `Result`, verified against `postgres@3.4.9` source: `class Result extends Array` (`src/result.js`) whose `count` is installed via `Object.defineProperties` — **non-enumerable** — and set from the CommandComplete tag (`src/connection.js:595`). `drizzle-orm/postgres-js` `session.cjs` returns `client.unsafe(...)` verbatim for a field-mapper-less `db.execute`, so that raw `Result` reaches callers with no `.rowCount` at all. A `DELETE` without `RETURNING` therefore has `length === 0` *and* `count === n` — putting the two branches in direct conflict, which an object literal can never do.

Proven discriminating by mutation:

| Mutation | Result |
|---|---|
| reorder helper to check `Array.isArray` before `.count` | **5 tests fail** (all 8 old fixtures would have passed) |
| add a 14th hand-rolled private copy | **guard fails, naming the file** |

The suite also pins the contracts the copies had drifted from: `.rowCount` priority, `count: null` falling through to length (postgres-js's default), full-batch values surviving so loops continue, and `null`/`undefined` **throwing** rather than reporting 0.

### Deleted test file — what it held and where that coverage lives

`jobs/processSampleRetention.test.ts` is **deleted** (31 lines). It contained three `vi.mock` calls and a single `describe` whose only subject was the now-removed local `extractRowCount` export — it tested nothing else about the module, so there was no other coverage to preserve. Its five assertions map 1:1 onto strictly stronger ones:

| Deleted assertion | Now covered by |
|---|---|
| prefers `rowCount` over `count` | `prefers .rowCount over .count when a shape carries both` |
| falls back to `count` (postgres-js DELETE) | `reads .count off a postgres-js DELETE result whose .length is 0` — real `Result`, not `{count:5}` |
| falls back to array length | `falls back to length for a plain array of mapped rows` |
| returns 0 for unrecognised object | `returns 0 for an unrecognised object shape` |
| full batch reported as batch size | `reports a full batch of %i as itself` (1 / 5000 / 10000) |

The other seven test files kept every unrelated test; only their duplicated `extractRowCount` blocks were removed. Confirmed mechanically — **every** removed `expect(` line in this diff is an `extractRowCount`-family assertion, zero unrelated assertions removed.

## Comment accuracy (`9b5d834`)

Comment review checked the new comments' *historical* claims against `git log` rather than against the issue body I lifted them from. Three did not hold. All are corrected or cut; every third-party technical claim (postgres-js `Result` shape, drizzle `execute()` passthrough) was checked and stands.

| Claim as first written | What the history actually shows | Action |
|---|---|---|
| "#3769 was the same failure mode in the other direction" | The only commit carrying that number is `b2bd606f5`, `fix(mobile): stop Systems and Tickets presenting partial state as complete`. Issues and PRs share one namespace, so #3769 is not a row-count incident. | **Removed** — I could not substantiate it, so it does not belong in a code comment. |
| "one of those copies shipped a lock check that reported the exact opposite of the truth (#3760)" | `eb6dfa9be` (#3760) **created** `db/rowCount.ts` and the `SELECT … FOR UPDATE` check in `services/deviceDeletion.ts`; it touches none of the fifteen consolidated files. | **Corrected** — reworded to describe the failure the check would have had (0 on every call fires "lock not held" on every *successful* delete), not a bug shipped by a copy. |
| "copied past eleven more times" | Note added 2026-08-23. Dating each copy: **twelve predate it**, exactly **three** followed (`eventDispatchWorker` 08-27, `orgMergeCustomExecutors` 08-27, `aiUnattendedExposureRetention` 08-28). | **Corrected to three** — which still makes the point that a comment is not a control. |

## Regression guard

`db/rowCount.test.ts` shells out to `git grep` and fails on any re-introduced private copy. It carries a **positive control** (the canonical file must be found by the guard's own pattern), so a typo'd regex or broken invocation fails loudly instead of passing vacuously — the exact way fifteen copies accumulated behind a doc comment that merely *asked* for consolidation.

The guard keys on the **discriminating test**, not the cast:

```
typeof [A-Za-z_$][A-Za-z0-9_$]*\??\.rowCount === 'number'
```

Proof it is stronger than the cast-based version review rejected — run against the two files as they stood at `bb461f2a5`:

| file (pre-fix) | old cast pattern | new logic pattern |
|---|---|---|
| `jobs/agentLogRetention.ts` | 0 matches | **1 match** |
| `jobs/changeLogRetention.ts` | 0 matches | **1 match** |

A second pattern still keys on `as unknown as { count: number }`, because `ipHistoryRetention`'s old read never tested `.rowCount` and would slip past a `.rowCount`-keyed pattern. `git grep` exit codes are handled precisely: only status 1 (no matches) counts as clean; 128 (bad regex / not a repo) and a missing `git` binary rethrow rather than reporting "no copies found".

## Verification

- `vitest run` — **192 passed / 15 files** (plus 17/17 re-run after the review-round-1 amendment), single-run `--pool=threads --maxWorkers=2`: `db/rowCount` (17), `services/{tenantCascade, tenantCascade.partner, orgMerge, orgMergeExecutors, metricRollupMaintenance}` (110), `jobs/{aiUnattendedExposure, audit, deviceMetrics, serviceProcessCheck, userRisk, eventDispatchWorker, eventDispatchWorker.sql, mlOutput, reliability}Retention` (82).
- `eslint` on all 23 changed files — **clean (exit 0)**.
- **Typecheck: CI's `Type Check` job passed on `bb461f2a5`** — that is the authoritative full `tsc --noEmit --project apps/api/tsconfig.json`. Locally the same command could not complete inside the 10-minute tool cap on this host (load average 25; attempts killed at `exit=124`), so local evidence is a scoped project over the changed files **and their full import closure** — which for `tenantCascade`, `orgMerge` and `eventDispatchWorker` pulls in most of the API — **exit 0, zero diagnostics**, re-run after the amendment.
- No migrations, no schema changes, no tenancy/cascade list changes — this PR only moves an existing pure function. `db/rowCount.ts`'s function body is byte-identical to `origin/main`; only its doc comment changed.

## Not fixed here (flagged by review, pre-existing)

- `services/tenantCascade.ts`'s final `partners` DELETE is the one delete in that function with no surrounding `try/catch`, so it skips the `writePurgeFailedAudit` breadcrumb every sibling writes. Pre-existing on `main`, reachable only via the internal synthetic-canary route, and only if `extractRowCount` throws — which requires a nullish driver result. Out of scope for a consolidation PR; worth its own issue.
- `metricRollupMaintenance.ts`'s duplicate `attachWorkerObservability` registration is issue #3946 and is deliberately untouched.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 reviewers in parallel: `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter`, `comment-analyzer`. (Skipped `type-design-analyzer` — no production types added — and `code-simplifier`, which mutates code on a pure-refactor PR.)

**Findings:** 6 raised → 4 addressed in `f24d0e2` + `9b5d834`, 2 recorded as out-of-scope. 0 outstanding.

| # | Reviewer | Severity | Finding | Resolution |
|---|---|---|---|---|
| 1 | code-reviewer | important | Sweep **and** the guard shipped with it both missed `agentLogRetention.ts` + `changeLogRetention.ts` — they cast to `Record<string, unknown>`, not the inline object type the regex keyed on. "ONLY row-count reader" was false as shipped. | **Fixed** `f24d0e2` — both converted; guard re-keyed on the discriminating `typeof x.rowCount === 'number'` test, which is cast-style agnostic. Verified against those files at `bb461f2a5`: old pattern **0** matches each, new pattern **1** each. Repo-wide (`apps`/`ee`/`packages`) it now matches exactly one file — `db/rowCount.ts`, which is also the positive control. |
| 2 | comment-analyzer | critical | `#3769` cited as "the same failure mode in the other direction". The only commit with that number is `b2bd606f5`, `fix(mobile): stop Systems and Tickets presenting partial state as complete`. | **Removed** `9b5d834`. I took it from the issue body without checking; unsubstantiated claims don't belong in code comments. |
| 3 | comment-analyzer | critical | "one of those copies shipped a lock check that reported the opposite of the truth (#3760)". `eb6dfa9be` (#3760) **created** `db/rowCount.ts` and the `SELECT … FOR UPDATE` check in `deviceDeletion.ts`; it touches none of the 15 consolidated files. | **Corrected** `9b5d834` — reworded to the failure that check *would* have had (0 on every call fires "lock not held" on every **successful** delete), not a bug shipped by a copy. |
| 4 | comment-analyzer | important | "copied past eleven more times" didn't match history. | **Corrected to three** `9b5d834`. Note added 2026-08-23; dating each copy: **12 predate it**, exactly **3** followed (`eventDispatchWorker` 08-27, `orgMergeCustomExecutors` 08-27, `aiUnattendedExposureRetention` 08-28). |
| 5 | silent-failure-hunter | suggestion | `tenantCascade.ts`'s final `partners` DELETE is the only delete in that function with no `try/catch`, so it skips the `writePurgeFailedAudit` breadcrumb its siblings write. | **Not fixed — pre-existing on `main`**, reachable only via the internal synthetic-canary route and only if `extractRowCount` throws. Recorded in the PR body; worth its own issue. |
| 6 | code-reviewer | suggestion | `vulnerabilityJobs.ts:1046` reads `result.length` off a `db.execute(… RETURNING)`. | **Not fixed — correct as written.** `RETURNING` makes the postgres-js `Result` a real array, so `.length` is accurate. Not a shape-check duplicate. |

All three third-party technical claims in the comments were independently checked by `comment-analyzer` and **stand**: the postgres-js `Result` shape (`Array` subclass, non-enumerable `count`, `null` default), and drizzle's `execute()` returning `client.unsafe(...)` verbatim.

---

### Scope accounting — every call site converted

The issue named six; the sweep found **fifteen**. Four were under a different name and two behind a `Record<string, unknown>` cast, which is why `grep extractRowCount` only found six.

| # | Call site | Symbol | Origin |
|---|---|---|---|
| 1 | `jobs/auditRetention.ts` | `extractRowCount` (private) | named in #3894 |
| 2 | `jobs/deviceMetricsRetention.ts` | `extractRowCount` | named in #3894 |
| 3 | `jobs/processSampleRetention.ts` | `extractRowCount` | named in #3894 |
| 4 | `jobs/serviceProcessCheckRetention.ts` | `extractRowCount` | named in #3894 |
| 5 | `services/metricRollupMaintenance.ts` | `extractRowCount` | named in #3894 |
| 6 | `services/tenantCascade.ts` | `extractRowCount` (private) | named in #3894 |
| 7 | `jobs/aiUnattendedExposureRetention.ts` | `extractRowCount` | found by sweep |
| 8 | `jobs/eventDispatchWorker.ts` | `extractAffectedCount` | found by sweep (renamed) |
| 9 | `jobs/mlOutputRetention.ts` | `extractMlOutputRetentionRowCount` | found by sweep (renamed) |
| 10 | `jobs/reliabilityRetention.ts` | `extractReliabilityRetentionRowCount` | found by sweep (renamed) |
| 11 | `jobs/userRiskRetention.ts` | `extractUserRiskRetentionRowCount` | found by sweep (renamed) |
| 12 | `services/orgMergeCustomExecutors.ts` | `extractRowCount` (re-exported into `orgMerge.ts`) | found by sweep |
| 13 | `jobs/ipHistoryRetention.ts` | inline `.count ?? .length ?? 0` | found by sweep |
| 14 | `jobs/agentLogRetention.ts` | inline, `Record<string, unknown>` cast | found by PR review |
| 15 | `jobs/changeLogRetention.ts` | inline, `Record<string, unknown>` cast | found by PR review |

#1–10 were byte-identical to the canonical body. #6 and #12 had drifted to null-**safe** `raw?.rowCount` and now throw instead — deliberate, since 0 is load-bearing (it's the per-table figure in the GDPR erasure audit). `silent-failure-hunter` verified the other 11 sites already had no `?.`, so only those two changed behaviour, and traced both: the throw surfaces through `writeErasureFailedAudit` / `org.merge.failed` and is rethrown, never swallowed.

### Deleted test file — resolution

`jobs/processSampleRetention.test.ts` is **deleted, not restored**, and no coverage was lost. It was 31 lines: three `vi.mock` calls plus one `describe` whose only subject was the now-removed local `extractRowCount` export. `pr-test-analyzer` independently confirmed it asserted nothing else about the module.

Its five assertions map 1:1 onto stronger ones in `db/rowCount.test.ts` (rowCount-over-count priority; postgres-js DELETE; array length; unrecognised shape → 0; full batch preserved).

**The deleted assertions were vacuous against the bug they were named for.** All 27 removed `expect(...)` calls used object/array literals. I evaluated their 8 distinct fixtures against both the correct helper and the array-before-count ordering that produces the 0-on-every-call failure: **8/8 pass under both**. `pr-test-analyzer` reproduced this independently, then mutation-tested the new suite — **5 of 17 fail** under that same ordering — and restored the worktree clean. Every removed `expect(` line in the diff is an `extractRowCount`-family assertion; zero unrelated coverage removed.

### `reportScheduleWorker.ts` — verdict: **benign, not a bug. Correctly out of scope.**

The issue asked me to verify rather than assume. Evidence, read against this branch:

- The line is `reportScheduleWorker.ts:552`: `const rowCount = result.rowCount ?? rows.length;`
- `result` is the return of `generateReport(...)`, declared `): Promise<ReportResult>` at `reportGenerationService.ts:577`. `ReportResult` is a **domain type** (`reportGenerationService.ts:26`) — `{ rows?: unknown[]; rowCount?: number; summary?; generatedAt?; previous? }`. **No postgres-js `Result` ever reaches this line**, so `??` is an ordinary optional-field fallback, not a wrong-shape read.
- Every tabular generator populates `rowCount` itself — 11 sites in `reportGenerationService.ts` (`:105 :221 :256 :283 :319 :340 :385 :414 :457 :610 :614`). Executive summary is summary-only and deliberately has none, which the `rows.length` fallback records as 0.
- **There is no batched loop and no early termination.** `rowCount` occurs exactly twice in the worker: computed at `:552`, stored on `report_runs` at `:560`. Nothing branches or terminates on it.

Folding it onto `extractRowCount` would be an active **regression**: the helper reads a top-level `.rowCount`/`.count` and never a nested `result.rows`, so a future generator returning `{ rows: [a, b] }` with no `rowCount` would go 2 → 0. Same for the identical read at `routes/reports/runs.ts:160`. Both left alone. This matches the conclusion already reached in the issue thread.

### Tests / typecheck

- `vitest run` — **192 passed / 15 files** single-run (`--pool=threads --maxWorkers=2`), plus 17/17 re-run after each of the two follow-up commits.
- **Typecheck: CI's `Type Check` job passed on `bb461f2a5`** (the authoritative full `tsc --noEmit --project apps/api/tsconfig.json`); `Lint` passed there too. Locally that command could not finish inside the 10-min tool cap (host load average 25, killed at `exit=124`), so local evidence is a scoped project over the changed files **and their full import closure** — **exit 0, zero diagnostics**, re-run after both follow-ups.

### CI as-of this comment (head `9b5d834d7`)

GitHub's queue is heavily backed up; recording status as-of rather than blocking.

- **20 green:** incl. Test Mobile, Test Portal, Test Excel/Word/Outlook/PowerPoint Add-ins, Office Add-in Core, all three M365 executors, Gitleaks, NPM Audit, Trivy, Trivy Filesystem Scan, Go Vulnerability Check, Check Migrations (non-superuser), Workflow Lint.
- **14 still queued:** Test API, Type Check, Lint, Test Web, Test Agent, Test Agent (race), Security Audit, both CodeQL Analyze jobs, Cargo Audit, Rust Check (Tauri apps), Lint Agent (Go), Check Migrations, Trivy Image Scan.
- **0 failures so far.**
- ⚠️ **`Check Migrations` and `Trivy Image Scan` are red repo-wide** — they fail on `main` and on unrelated PRs #4232 / #4236 / #4237. #4227 tracks the migrations one and #4236 fixes it. **Not attributable to this change**: this PR adds no migrations and touches no Dockerfile or dependency.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3894 — both are the maintainer's call.

PR #4252, opened 2026-08-30
feat(mobile): Offline queue for time-entry writes (#3206 W04)

### LINKED ISSUE #3898: W04: Offline queue for time-entry writes
Wave W04 of #3206 — Offline queue for time-entry writes
Branch: `feature/3206-mobile-ticketing-time-entry/wave-3898`

ToddHebebrand:
## What / why

Wave W04 of mobile ticketing + time entry (#3206). Implements plan **Task 3** — the offline queue that makes the field-technician case work: a tech in a basement starts a timer, stops it 40 minutes later, and must not lose the entry.

`apps/mobile/src/services/timeEntryQueue.ts` is an AsyncStorage-backed queue with durable append, oldest-first drain, and replay-outcome classification. No React, no direct network — `drain(send)` takes the transport as a callback, so Task 6 owns connectivity and Task 5 owns the API calls.

Semantics:

- **Transient failure stops the drain** rather than skipping past it. A `stop` replayed before its `start` would 404, so ordering is worth more than throughput. The head write's `attempts` is incremented and the remainder stays queued.
- **A 4xx is dropped**, not retried. It is a verdict rather than a delay; retrying forever would wedge every later entry behind a poison write. Dropped writes are returned in `DrainResult.dropped` so Task 6 can surface them — a silently discarded time entry is unbillable work. Both `status` and the `statusCode` alias are recognised.
- **Corrupt storage reads as an empty queue.** Losing unsynced entries is bad; an unopenable Time tab is worse.

### Two durability holes fixed on top of the plan text

Both were found reviewing the draft against the plan, and both were written red-first (test added, watched fail, then fixed):

1. **A write enqueued during an in-flight drain was clobbered.** `drain` snapshotted the queue, then persisted `snapshot.slice(index)` at the end — so a tech tapping Stop mid-replay had that write overwritten. `drain` now re-reads storage before persisting and appends anything not in its snapshot, preserving order.
2. **Overlapping drains sent the same write twice.** Task 6 calls `drain` on every `false -> true` network transition, so a flapping connection double-sent; the replayed `start` 409s `ENTRY_RUNNING` and is then discarded as a 4xx — silently losing billable work. `drain` calls are now serialised through a rejection-safe promise chain.

`enqueue` deliberately does **not** wait on the drain chain (it is called from inside `send` callbacks), and the storage lock never encloses the send loop, so the two cannot deadlock.

**Deliberately not added:** no cap on queue length and no cap on `attempts`. The plan forbids a queued time entry vanishing quietly (Task 6 Step 2), and a technician offline for days must not have work discarded by a retry limit. Raising this as a decision rather than a gap.

## Tasks covered

- Plan Task 3 — `apps/mobile/src/services/timeEntryQueue.ts` + `.test.ts`.
- Plan Task 4 `queueDepthChanged` contract — **no change needed** (verified): the plan specifies `queueDepthChanged(state, n)` taking a plain number, and W03 shipped exactly `queueDepthChanged(state, pendingCount: number): TimeState` in `apps/mobile/src/store/timeSlice.ts:43`, covered by `timeSlice.test.ts:61`. The slice stays a pure reducer; the caller passes the depth in Task 6.

## Test summary

`pnpm --filter breeze-mobile test` — **665 passed, 3 skipped, 55 files** (verified). `pnpm --filter breeze-mobile typecheck` — **clean** (verified).

12 queue tests, covering every plan-listed case plus the two new ones: ordered drain; `enqueue` stamps id/queuedAt/attempts and survives a fresh `readQueue`; transient failure retains order; `attempts` increments on the failed head only; 5xx retained as transient; 4xx dropped; `statusCode` alias; a poison entry dropped once with later entries draining; corrupt-JSON and non-array storage recovery; enqueue-during-drain preserved; overlapping drains serialised.

The two new tests were confirmed **discriminating** — against the pre-fix implementation they failed with the exact predicted symptoms (`['stop']` vs `[]`, and `['start','start','stop','stop']` vs `['start','stop']`).

## Not checked

- **No on-device verification.** Plan Task 6 Step 3 (airplane mode -> start -> stop -> reconnect) is that wave's step; nothing here has run on a phone.
- **No replay against a live API.** Unit tests mock the transport, per the plan's Global Constraints.
- `enqueue` ids use `Date.now()` + `Math.random()`. Tests assert only shape and distinctness, never a literal clock value, so they are not time-flaky — but collision resistance is argued, not measured.
- Concurrent `enqueue` calls from different call sites are serialised by the storage lock; not stress-tested.

## CI

Base is `feature/3206-mobile-ticketing-time-entry/wave-3897`, not `main`, so `ci.yml` (`pull_request: branches: [main]`) does **not** run on this PR. CI dispatched per branch: `gh workflow run CI --ref feature/3206-mobile-ticketing-time-entry/wave-3898`.

**Check Migrations** and **Trivy Image Scan** are red on `origin/main` itself (355b37465, step "Shipped-migration immutability guard") — pre-existing, not caused by this work, which touches no migrations and no images.

Closes #3898

### DIFF
--- a/apps/mobile/src/services/timeEntryQueue.test.ts
+++ b/apps/mobile/src/services/timeEntryQueue.test.ts
@@ -14,7 +14,7 @@ vi.mock('@react-native-async-storage/async-storage', () => ({
   default: {
     getItem: async (k: string) => {
       // Literal, not the imported const: this factory is hoisted above imports.
-      if (failNextQueueRead && k === 'breeze.timeEntryQueue.v1') {
+      if (failNextQueueRead && k === 'breeze.timeEntryQueue.v2') {
         failNextQueueRead = false;
         throw new Error('SQLite busy');
       }
@@ -45,8 +45,12 @@ import {
   enqueue,
   readQueue,
   drain,
+  readNeedsAttention,
+  clearNeedsAttention,
   QUEUE_KEY,
   QUEUE_CORRUPT_KEY,
+  QUEUE_NEEDS_ATTENTION_KEY,
+  QUEUED_KINDS,
 } from './timeEntryQueue';
 import type { QueuedWrite } from './timeEntryQueue';
 
@@ -69,21 +73,67 @@ function lastSentryExtra(): Record<string, unknown> {
   return context?.extra ?? {};
 }
 
+const SPAN = { startedAt: '2026-08-30T09:00:00.000Z', endedAt: '2026-08-30T09:40:00.000Z' };
+
+describe('the queue can only hold self-timestamped writes', () => {
+  // The whole offline-span design rests on this: `/time-entries/start` and
+  // `/time-entries/stop` are SERVER-STAMPED verbs, so queueing one defers its
+  // stamping to an arbitrary later moment and silently rewrites the customer's
+  // bill. Making the union runtime-inspectable is what lets a test pin it.
+  it('QUEUED_KINDS contains neither start nor stop', () => {
+    expect([...QUEUED_KINDS].sort()).toEqual(['closeEntry', 'create']);
+    expect(QUEUED_KINDS).not.toContain('start');
+    expect(QUEUED_KINDS).not.toContain('stop');
+  });
+
+  it('enqueue never produces a dependsOn key on any write', async () => {
+    // `dependsOn` existed only to pair a queued start with the stop that closed
+    // it. With no start to pair, a surviving field is a heuristic waiting to be
+    // reintroduced — so assert on the KEY, not on a falsy value.
+    const created = await enqueue({ kind: 'create', payload: { ...SPAN } });
+    const closed = await enqueue({ kind: 'closeEntry', payload: { id: 'E', endedAt: SPAN.endedAt } });
+    for (const write of [created, closed]) {
+      expect(Object.prototype.hasOwnProperty.call(write, 'dependsOn')).toBe(false);
+    }
+    for (const write of await readQueue()) {
+      expect(Object.prototype.hasOwnProperty.call(write, 'dependsOn')).toBe(false);
+    }
+    const raw = JSON.parse(store.get(QUEUE_KEY) ?? '[]') as Array<Record<string, unknown>>;
+    for (const row of raw) {
+      expect(Object.prototype.hasOwnProperty.call(row, 'dependsOn')).toBe(false);
+    }
+  });
+
+  it('refuses a persisted row whose kind is a server-stamped verb', async () => {
+    // A row written by the pre-v2 build (or a future one) must not resurrect a
+    // timer verb through the untyped storage boundary.
+    store.set(
+      QUEUE_KEY,
+      JSON.stringify([
+        { id: 'a', kind: 'start', payload: {}, queuedAt: '2026-08-30T09:00:00.000Z', attempts: 0 },
+        { id: 'b', kind: 'create', payload: SPAN, queuedAt: '2026-08-30T09:00:00.000Z', attempts: 0 },
+      ])
+    );
+    const kept = await readQueue();
+    expect(kept.map((w) => w.kind)).toEqual(['create']);
+  });
+});
+
 describe('timeEntryQueue', () => {
   it('replays writes in the order they were made', async () => {
-    await enqueue({ kind: 'start', payload: { ticketId: 'k1' } });
-    await enqueue({ kind: 'stop', payload: { isBillable: true } });
+    await enqueue({ kind: 'create', payload: { ...SPAN, ticketId: 'k1' } });
+    await enqueue({ kind: 'closeEntry', payload: { id: 'E', endedAt: SPAN.endedAt } });
     const seen: string[] = [];
     const result = await drain(async (w) => {
       seen.push(w.kind);
     });
-    expect(seen).toEqual(['start', 'stop']);
-    expect(result).toMatchObject({ sent: 2, remaining: 0, dropped: [] });
+    expect(seen).toEqual(['create', 'closeEntry']);
+    expect(result).toMatchObject({ sent: 2, remaining: 0, needsAttention: [] });
     await expect(readQueue()).resolves.toEqual([]);
   });
 
   it('enqueue stamps id, queuedAt and attempts=0 and is durable across a fresh readQueue', async () => {
-    const item = await enqueue({ kind: 'create', payload: { durationMinutes: 30 } });
+    const item = await enqueue({ kind: 'create', payload: { ...SPAN } });
     expect(item.id).toEqual(expect.any(String));
     expect(item.id.length).toBeGreaterThan(0);
     expect(Number.isNaN(Date.parse(item.queuedAt))).toBe(false);
@@ -95,26 +145,27 @@ describe('timeEntryQueue', () => {
     expect(store.has(QUEUE_KEY)).toBe(true);
 
     // Two enqueues get distinct ids.
-    const second = await enqueue({ kind: 'start', payload: {} });
+    const second = await enqueue({ kind: 'closeEntry', payload: { id: 'E', endedAt: SPAN.endedAt } });
     expect(second.id).not.toBe(item.id);
     await expect(readQueue()).resolves.toHaveLength(2);
   });
 
   it('stops draining on a transport failure and keeps the rest queued', async () => {
-    await enqueue({ kind: 'start', payload: {} });
-    await enqueue({ kind: 'stop', payload: {} });
+    await enqueue({ kind: 'create', payload: { ...SPAN } });
+    await enqueue({ kind: 'closeEntry', payload: { id: 'E', endedAt: SPAN.endedAt } });
     const result = await drain(async (w) => {
-      if (w.kind === 'start') throw Object.assign(new Error('offline'), { status: undefined });
+      if (w.kind === 'create') throw Object.assign(new Error('offline'), { status: undefined });
     });
-    // Order matters more than throughput: a stop must never land before its start.
-    expect(result).toMatchObject({ sent: 0, remaining: 2, dropped: [] });
+    // Order matters more than throughput: a correction must never land before
+    // the entry it corrects.
+    expect(result).toMatchObject({ sent: 0, remaining: 2, needsAttention: [] });
     const queue = await readQueue();
-    expect(queue.map((q) => q.kind)).toEqual(['start', 'stop']);
+    expect(queue.map((q) => q.kind)).toEqual(['create', 'closeEntry']);
   });
 
   it('increments attempts on the write that failed transiently, and persists it', async () => {
-    await enqueue({ kind: 'start', payload: {} });
-    await enqueue({ kind: 'stop', payload: {} });
+    await enqueue({ kind: 'create', payload: { ...SPAN } });
+    await enqueue({ kind: 'closeEntry', payload: { id: 'E', endedAt: SPAN.endedAt } });
     await drain(async () => {
       throw new Error('offline');
     });
@@ -132,43 +183,43 @@ describe('timeEntryQueue', () => {
   });
 
   it('treats a 5xx as transient (retained), not a verdict', async () => {
-    await enqueue({ kind: 'start', payload: {} });
+    await enqueue({ kind: 'create', payload: { ...SPAN } });
     const result = await drain(async () => {
       throw Object.assign(new Error('boom'), { status: 500 });
     });
-    expect(result).toMatchObject({ sent: 0, remaining: 1, dropped: [] });
+    expect(result).toMatchObject({ sent: 0, remaining: 1, needsAttention: [] });
   });
 
-  it('drops a permanently-rejected write instead of blocking the queue forever', async () => {
-    await enqueue({ kind: 'stop', payload: {} });
-    await enqueue({ kind: 'start', payload: { ticketId: 'k2' } });
+  it('parks a permanently-rejected write instead of blocking the queue forever', async () => {
+    await enqueue({ kind: 'closeEntry', payload: { id: 'gone', endedAt: SPAN.endedAt } });
+    await enqueue({ kind: 'create', payload: { ...SPAN, ticketId: 'k2' } });
     const result = await drain(async (w) => {
-      if (w.kind === 'stop')
-        throw Object.assign(new Error('no timer'), { status: 404, code: 'NO_RUNNING_TIMER' });
+      if (w.kind === 'closeEntry')
+        throw Object.assign(new Error('no entry'), { status: 404, code: 'ENTRY_NOT_FOUND' });
     });
-    // A 4xx will never succeed on retry — dropping it lets the queue progress.
-    expect(result.dropped.map((d) => d.kind)).toEqual(['stop']);
-    expect(result).toMatchObject({ sent: 1, remaining: 0 });
+    // A 4xx will never succeed on retry — parking it lets the queue progress.
+    expect(result.needsAttention.map((d) => d.kind)).toEqual(['closeEntry']);
+    expect(result).toMatchObject({ sent: 1, remaining: 0, needsAttentionTotal: 1 });
     await expect(readQueue()).resolves.toEqual([]);
   });
 
   it('recognises the statusCode alias used by TimeEntryError as a permanent failure', async () => {
-    await enqueue({ kind: 'stop', payload: {} });
-    await enqueue({ kind: 'start', payload: {} });
+    await enqueue({ kind: 'closeEntry', payload: { id: 'E', endedAt: SPAN.endedAt } });
+    await enqueue({ kind: 'create', payload: { ...SPAN } });
     const result = await drain(async (w) => {
-      if (w.kind === 'stop')
+      if (w.kind === 'closeEntry')
         throw Object.assign(new Error('running'), { statusCode: 409, code: 'ENTRY_RUNNING' });
     });
-    expect(result.dropped.map((d) => d.kind)).toEqual(['stop']);
+    expect(result.needsAttention.map((d) => d.kind)).toEqual(['closeEntry']);
     expect(result).toMatchObject({ sent: 1, remaining: 0 });
   });
 
-  it('drops a 422 as a payload verdict', async () => {
-    await enqueue({ kind: 'create', payload: {} });
+  it('parks a 422 as a payload verdict', async () => {
+    await enqueue({ kind: 'create', payload: { ...SPAN } });
     const result = await drain(async () => {
       throw Object.assign(new Error('unprocessable'), { status: 422 });
     });
-    expect(result.dropped.map((d) => d.kind)).toEqual(['create']);
+    expect(result.needsAttention.map((d) => d.kind)).toEqual(['create']);
     expect(result).toMatchObject({ sent: 0, remaining: 0 });
   });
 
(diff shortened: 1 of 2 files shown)


ToddHebebrand:

Every finding below was run past three independent verifiers with a mandate to refute it. Four survived unanimously; seven were refuted and are **not** applied. Claims are labelled **verified** (I read the code or ran it), **inferred**, or **not-checked**.

Head commit: `fc3a93d1ec14d4c70b79bd017a5dbf6a607824b7`. Each fix has a test that was **red before the change** (verified: 10 new tests failed against `b3b402bd4`, exactly the 10 intended, with the other 15 green).

### Fixed

**1. Blanket 4xx drop destroyed recoverable billable writes** — `timeEntryQueue.ts:83`
`status >= 400 && status < 500` treated the whole 4xx range as a verdict and deleted the write from AsyncStorage. Three of those codes succeed on retry, and two are the most likely responses on the reconnect path this queue exists to serve (verified): **401** — `requestWithPrefix`/`coreRequest` (`api.ts:220-345`) has no refresh-and-retry, only the 428 native-binding bootstrap, so a phone offline past the access-token TTL gets a 401 on the first replay; **403** — the W02 gate recorded in #4234: the default seeded `Partner Technician` holds `time_entries:read` but not `time_entries:write`, so a default-role technician 403s on *every* start/stop until an admin grants the permission; **429** — `app.use('*', globalRateLimit())` (`apps/api/src/index.ts:399`) covers `/api/v1/time-entries`, and a drain replays a backlog back-to-back with no pacing; **408** — request timeout (weakest leg: the client's own `fetchWithTimeout` abort carries no status, so only an intermediary produces one).
The drop set is now an explicit verdict list — `PERMANENT_STATUSES = {400, 404, 409, 422}` — and everything else is retained and retried. New tests assert a 401, 403, 429 and 408 each leave `remaining: 1, dropped: []` and the write still in `readQueue()`.
*Note on the 403 choice:* finding #1 proposed 403 as permanent while finding #3 proved it must be retained. Retaining wins: a missing permission is grantable, so dropping destroys work that a one-line role change would have saved. The cost is a head-of-line wedge until the grant lands, which is an availability problem, not a data-loss one — and `headAttempts` (below) is how a caller sees it.

**2. Dropping a start but replaying its paired stop hit an unrelated running entry** — `timeEntryQueue.ts:84`
Verified in `apps/api/src/services/timeEntryService.ts:425-446, 543-546`: `stopTimer` calls `stopRunningEntry(actor)`, a CAS on `user_id = actor AND ended_at IS NULL`. It takes **no entry id** from the client (the `stopTimerSchema` carries only `description`/`isBillable`), so it stops whatever entry that user currently has running and stamps `ended_at = now` plus the mobile payload's description/isBillable. An orphaned stop is therefore not a harmless 404 — it inflates and overwrites a timer the phone never started (e.g. one started that morning from the web dashboard). The drop path violated the module's own stated invariant at line 107 ("a stop can never reach the server before its start"), which the transient path `break`s to preserve.
`enqueue` now links a `stop` to the queued `start` it closes (`dependsOn`, set automatically by scanning back to the most recent unclosed start), and a dropped write takes its dependants with it — unsent. A stop for a timer started **online** has no queued start, stays unlinked, and still replays (test added, green).
*Correction carried from verification (verified):* a closed/deleted ticket is **not** a trigger — the time-tracking ticket read selects no status and no deleted flag, and ticket deletion is soft. The real triggers are org-access revocation (`resolveTicketLink` → 404 `TICKET_ORG_DENIED`), `TICKET_WRONG_PARTNER`, the rate-representability 400, and 409 `ENTRY_RUNNING`.

**3. No test pinned 401/403/429; the blanket drop silently destroyed recoverable offline work** — `timeEntryQueue.test.ts:93`
The suite exercised only 400, 404 and 409 on the drop branch — all genuine verdicts — so nothing discriminated the retriable codes. Verified by probe against byte-identical source before the fix: two queued writes drained against 403 gave `{sent:0, dropped:[start,stop], remaining:0}` with storage `[]`. Covered now by the four retention tests in #1, plus a 422-is-a-verdict test so the permanent set is pinned from both sides.

**4. `readQueue` swallowed every storage failure to `[]`, and the next enqueue clobbered the backlog** — `timeEntryQueue.ts:38`
The bare `catch { return [] }` covered `AsyncStorage.getItem` rejecting as well as `JSON.parse`, and emitted nothing — no console, no Sentry — in a module whose siblings file Sentry events for far less (`api.ts:265` for a missing installation id, `auth.ts:157` for a partial keychain wipe). Because `enqueue` does read-modify-write through the same `readQueue`, one transient read failure plus one Stop tap persisted `[stopWrite]` over the real blob, which made the design's own rationale ("the rows are recoverable by hand") false.
Now: a **read** failure throws `QueueStorageError`, so `enqueue` and `drain` abort rather than persist a stale view (tests assert the bytes on disk are unchanged and the drain sent nothing); a **parse** failure parks the blob at `breeze.timeEntryQueue.v1.corrupt` before returning `[]`, so the next write cannot overwrite it; both paths `Sentry.captureException` with `tags: { area: 'time-entry-queue' }`.
*Correction carried from verification (verified):* a failure of `drainQueue`'s **initial** read was not destructive under the old code — `snapshotIds` is empty, so every stored row is reclassified as an appended write and rewritten. Only `enqueue`, and a failure of the drain's **final** merge read, destroyed data. The fix covers all three.

### Suggestions applied

- **`readQueue` validated only the top-level shape** — a well-formed array with a `null` element threw a `TypeError` out of `queue.map(w => w.id)` and, because nothing rewrites storage on that path, never self-healed. Elements are now validated (string `id`, known `kind`, object `payload`) and malformed ones discarded with a quarantine + Sentry record. Test added.
- **`exports the QueuedWrite shape` was vacuous** — it asserted a property of a literal the test itself wrote, exercised no module code, and could not fail. Replaced with a payload deep round-trip through storage, which is the assertion nothing previously covered (the old durability test compared against the same object reference `enqueue` returned).
- **`attempts` was written and read by nothing** — `DrainResult` now carries `headAttempts`, so a caller can tell "the phone is offline" (1) from "the head write keeps failing and nothing behind it can move" (large), which is the signal the retained-403 case needs.

### Declined

- **Cap `attempts` and drop or dead-letter a write past N failures.** Declined: it reintroduces exactly the silent destruction finding #1 and #3 establish as the critical defect — a 403 that is retried 20 times is still recoverable the moment the permission is granted, and a threshold would delete it just before that. The wedge is now *observable* via `headAttempts` instead; the policy for surfacing it belongs to the wiring wave that owns the UI (plan Task 6). Not-checked: whether that wave will want a user-triggered discard.

### Refuted (not applied)

- **`attempts` has no cap, so a permanently-failing non-4xx wedges the queue forever.** Refuted: the named trigger is wrong (verified — `time_entries.description` is Postgres `text`, and every write schema caps it at `z.string().max(10_000)`, so an over-long description is a 400, not a 500), and `TimeEntryServiceError.status` is typed `400 | 403 | 404 | 409` — every domain rejection is 4xx by construction. No permanent non-4xx was demonstrated to exist. The unbounded-growth half is also wrong: retries mutate a number in place, they do not append.
- **Replay is at-least-once with no dedupe key, so a lost response double-bills a `create`.** Refuted as a defect in this diff: the mechanism is real (verified — no idempotency key anywhere in `routes/timeEntries/` or `timeEntryService.ts`), but a write is enqueued *only* when the device is already known offline, so a write that reached the API never entered the queue; and no `kind: 'create'` is enqueued anywhere in the plan. An idempotency key is a backend change, which this wave's "no API, schema, or migration changes" constraint excludes.
- **A corrupt array of malformed elements bricks the drain permanently.** Refuted **as filed** — verified by probe that only a literal `null` element throws (`[{}]` and primitives map cleanly), and no writer can produce one: `QUEUE_KEY` has exactly two writers, both `JSON.stringify` of fully-constructed literals, and a truncated write yields unparseable JSON, not a parseable array. The hardening it asked for was nevertheless cheap and is applied above.
- **`attempts` is asserted but the real wedge goes untested.** Refuted: the named error (`coreRequest` throwing a statusless "Failed to refresh token") does not exist on this path — verified, `refreshToken()` has exactly three references and `coreRequest` is not one of them. The genuine auth failure is an HTTP 401, which is now classified and retained.
- **No test for an AsyncStorage write failure at the final persist; re-sent writes are then dropped as 409.** Refuted: `startTimer` does **not** 409 when a timer is already running — verified at `timeEntryService.ts:474-528`, it auto-stops the previous timer and inserts `ON CONFLICT DO NOTHING`; `ENTRY_RUNNING` is raised only after two failed attempts on a genuine concurrent race. The claimed "one entry lost" outcome does not follow. The residual send-then-persist window is inherent at-least-once and cannot be closed without server-side idempotency.
- **The blanket 4xx drop is a design decision of the approved plan, not a defect.** Refuted as a defence: the plan does specify it (lines 429-431), but the plan's premise — that a 4xx is always a verdict, and that an orphaned stop would harmlessly 404 — is false for 401/403/408/429 and false for `stopRunningEntry`. Findings #1-#3 are plan-level defects faithfully implemented; fixing them here is correct.
- **`QUEUE_KEY` is not wiped on logout, so one account's writes replay under the next account's token.** Refuted for this PR: verified that `clearAuthData()` (`auth.ts:116-134`) does not clear the queue key and that the queue module has **zero** production call sites, so the key is never populated by a real user. The shape of the concern is real and belongs as a requirement on the wave that wires `enqueue`/`drain` — it should land there with a regression test.
- **Dropped writes are deleted before any caller surfaces them; there is no dead-letter record.** Refuted: the scenario is self-contradictory (`ENTRY_BILLED` is raised only from `updateTimeEntry`, and there is no `update` kind in the queue), `dropped` hands the caller the complete `QueuedWrite`, and the plan assigns the surfacing to the later wave (Task 6, "surface a Toast naming how many entries could not be saved").

### Strengths

- Storage serialisation is correct: `withStorageLock` (lines 22-29) chains on both fulfilment and rejection, so a failed operation cannot wedge the chain, and `enqueue`'s read-modify-write is genuinely atomic against a concurrent drain persist.
- The in-flight-enqueue merge (lines 94-100) is the right design — snapshotting ids and re-reading under the lock means a Stop tapped mid-replay is preserved rather than clobbered by the drain's final write, and the test at line 144 proves it with the real failure framing ("unbillable work if the drain's final persist clobbers it").
- `drain` claims the chain synchronously before yielding (line 120), which is what actually makes the overlapping-drain test at line 156 meaningful; the comment correctly explains why a double-sent start would 409 and then be lost as a 4xx.
- Stopping the drain on a transient failure to preserve start-before-stop ordering is the right trade (throughput sacrificed for correctness), and the docstring states the reasoning rather than just the behaviour.
- Tests are non-vacuous: they assert persisted state via a fresh `readQueue` rather than a return value, cover the poison-entry sequence with an explicit send-order assertion (line 134), and exercise both the `status` and `statusCode` aliases that `TimeEntryError` actually emits.

### Verification

- `pnpm --filter breeze-mobile typecheck` — **clean** (verified).
- `pnpm --filter breeze-mobile test` — **55 files, 678 passed, 3 skipped** (verified).
- CI re-dispatched on the branch: https://github.com/LanternOps/breeze/actions/runs/33322525511
- Pre-existing and unrelated: **Check Migrations** and **Trivy Image Scan** fail on `origin/main` itself (`355b37465`, step "Shipped-migration immutability guard"). Not caused by this branch — do not chase them here.

ToddHebebrand:
## Round-two fixes

All four confirmed findings are fixed on `e5b0634fc`. Every fix was written red-first: the test was added against the pre-fix code and watched to fail with the predicted symptom (failure output quoted per item below), then the code changed. **verified**: `pnpm --filter breeze-mobile typecheck` clean, `pnpm --filter breeze-mobile test` 686 passed / 3 skipped across 55 files.

### Q1 (critical) — the drop decision is now durable, not drain-local

The "this start was permanently rejected" decision lived only in the drain's in-memory `droppedIds`. Two ways an orphaned stop escaped it:

- the drain broke on a transient failure *before* reaching the stop, so the stop was persisted with a `dependsOn` pointing at a write that no longer existed, and the next drain (empty `droppedIds`) sent it;
- a stop enqueued *during* the drain was re-appended verbatim by the final merge, which never consulted the drop set.

Because the server stop is a CAS on `(user_id, ended_at IS NULL)` with no entry id, either path closes whatever unrelated timer is running and stamps it with the replay time.

**Fix** — the decision is written back to storage rather than remembered:
- `resolveDependants()` (new, fixpoint) strips dropped ids from every surviving write's candidate list and yields the writes left with nothing to depend on;
- the drain's final merge runs it over `snapshot-remainder ++ writes-appended-during-the-drain` and persists the result, so the orphan is *gone from the blob*, not merely skipped;
- when storage fails mid-reconcile, the outcome is recorded under a new `QUEUE_TOMBSTONE_KEY` (`{sent, dropped}`) which `readQueue` applies on every subsequent read — so even the "storage broke at the worst moment" path cannot resurrect the orphan. The tombstone is written under the storage lock (an `enqueue` interleaving there would otherwise clear a tombstone it never saw) and is retired by the next successful persist.

**Tests**: `never replays an orphaned stop on a LATER drain when its start was dropped` (red: `dropped` was `['start']`, the stop was still in storage and the second drain sent it) · `drops a stop enqueued DURING the drain that dropped its start` (red: same, `['start']`).

### Q2 — a failed final re-read no longer re-sends transmitted writes

If the re-read at the end of `drainQueue` threw, the whole drain rejected with the sent rows still in storage; the next drain replayed them — a duplicate billable `create` on the customer's invoice.

**Fix** — the merge is wrapped; on any failure the drain records the tombstone of what it actually sent and dropped and returns normally (the sends really happened, so rejecting was the wrong report). `readQueue` filters those ids out on the next read. Writes appended mid-drain are *not* lost, because the tombstone removes ids rather than overwriting the blob with a stale snapshot. If the tombstone write itself fails, the Sentry event carries `recorded: false` — the one case we cannot make safe is at least not silent.

**Tests**: `does not re-send transmitted writes when the drain final re-read fails` (red: the drain rejected with `QueueStorageError`; both creates were still queued) · `a write enqueued during a drain whose re-read fails is still replayed` (pins that the fix does not trade a duplicate for a loss).

### Q3 — `quarantine()` cannot report a park that did not happen

The empty `catch` meant a failed park still returned normally: `readQueue` returned `[]`, the next `enqueue` overwrote `QUEUE_KEY`, and the Sentry event positively claimed the bytes were quarantined. The forensic trail said "recoverable by hand" about bytes that no longer existed.

**Fix** — `quarantine()` returns whether the copy landed and reports `parked` (plus `parkError`) in the Sentry `extra`. When the park fails, `readQueue` throws `QueueStorageError` instead of returning `[]`, which is what stops the next write from destroying the only copy — the same "a read failure is not an empty queue" rule the module already applies elsewhere. Applied to all three corruption paths (unparseable, not-an-array, malformed elements).

**Tests**: `refuses to report a quarantine that did not happen, and keeps the bytes` (red: `readQueue` resolved `[]`) · `reports parked:true when the corrupt blob really was copied aside` (red: no `parked` key at all — this is the control that proves the flag is not hard-coded).

### Q4 — `dependsOn` identifies the specific start, not "the most recent one"

`enqueue` linked a stop to the newest queued start only. With `[startA, startB, stop]`, rejecting `startB` discarded the stop — but `startTimer` auto-stops the previous timer, so after the replay `A` is the running entry and the stop legitimately closes it. The old behaviour left `A` open forever.

**Fix** — `dependsOn` is now `string[]`: the ordered candidate starts back to the previous queued stop. A candidate is struck from the list when it is dropped; the stop is discarded only when the list empties. `asQueuedWrite` normalises a legacy bare-string `dependsOn` to a one-element list so rows already sitting on a TestFlight device keep their link across the upgrade.

**Tests**: `keeps a stop whose EARLIER start landed when a later start was rejected` (red: the stop was never sent) · `honours a dependsOn persisted by the first build as a bare string` — **verified by mutation**: deleting the string branch of `asDependsOn` turns that test red, so it discriminates rather than confirms.

Round one is untouched and still pinned by its own tests: 401/403/408/429 **retained**, 400/404/409/422 **permanent**.

### Deferred to W05 (#4259)

- **The ~0-minute offline span** the reviewers also raised against this branch is **not** fixed here. It is a property of how the caller stamps and collapses an offline start/stop pair, which lives in W05 (`timeEntryCollapse.ts` / `timerActions.ts`), not in the queue's replay logic. Fixing it in two places was explicitly off the table, and the quorum has decided its handling; it is being applied on W05.
- **W05 needs a one-line follow-up when it rebases onto this commit** — *inferred, not compile-checked* (W05 has not been rebased yet): `apps/mobile/src/services/timeEntryCollapse.ts:88` does `byId.get(stop.dependsOn)`, which no longer typechecks against `string[]`. The correct pairing there is the **last** candidate (`stop.dependsOn[stop.dependsOn.length - 1]`) — the newest start is the one actually running when the stop is taken. Its test fixtures at `timeEntryCollapse.test.ts:27,51,65,92,106,116` pass `dependsOn: 'a'` and want `['a']`.

### Residual risk (not a regression, and not fixed here)

A stop enqueued **after** the drain that dropped its start has no candidate to link to — the start row is already gone from the queue — so it is treated as "a timer started while online" and sent. Closing that needs the caller to pass an explicit `dependsOn` from the local running-timer state; `enqueue` already accepts one, so the hook exists. That is a W05 wiring question, and it is unchanged from the behaviour on this branch before these fixes.

ToddHebebrand:
## Round-three review — the two rewrite commits

Round two ended by *replacing* the offline design rather than repairing it, and its two replacement commits (`76cb5de23`, `af7576c77`) landed 20 minutes after the round-two summary was posted. That code had therefore never been reviewed. This round covers exactly that gap: the diff `e5b0634fc..af7576c77` (`timeEntryQueue.ts` + its suite).

Also merged `main` into the branch (via W03) — `#4236` on main is what fixes the `Check Migrations` and `Trivy Image Scan` reds these branches were carrying, so those failures were a stale base rather than anything in the PR.

### Confirmed defect (1)

**`readNeedsAttentionUnlocked` returned `[]` for two different situations — an empty store, and a store it could not read — and both writers then persisted over it.**

The queue side of this module already refuses to make that conflation (`readQueue`: *"A read failure is NOT an empty queue"*, and it throws `QueueStorageError` rather than return `[]`). The parked store, added by these two commits, did the opposite:

- **`parkNeedsAttention`** pushed onto that `[]` and wrote **one row over every previously parked one**, then returned `true`. The drain treats `true` as "the park landed" and its next step is to remove the row from the queue — so a transient `AsyncStorage` read failure destroyed both the earlier parked records *and* the queued original, and reported success. This is reachable from two callers: the drain's permanent-rejection path, and `timerActions.ts`'s `CLOCK_WENT_BACKWARDS` park.
- **`clearNeedsAttention`** computed `kept` from the same `[]` and persisted it, so dismissing one row during a storage hiccup cleared all of them.

Severity is the module's own worst case: silent deletion of billable work that nothing else holds a copy of. The parked store is not an exotic path either: every 400/404/409/422 verdict lands in it, and `timerActions.ts` parks a backwards-clock span there directly.

**Fixed in `cf55c29a5`.** The reader returns `null` for a storage failure and the writers branch on it: the park returns `false` (the drain leaves the write queued and the queue wedges visibly — the module's stated preference over a deletion) and the clear makes no write at all. `readNeedsAttention()` keeps its lenient public contract (`?? []`), since its two UI callers already `.catch(() => null)` and only render.

Unparseable content is deliberately handled the *opposite* way. It is deterministic, so refusing it forever would wedge every later park behind one bad blob and a poison entry would block the queue for good. Those bytes are copied to a new `QUEUE_NEEDS_ATTENTION_CORRUPT_KEY` — separate from `QUEUE_CORRUPT_KEY` so a corrupt queue and a corrupt parked store cannot overwrite each other's only surviving copy — and the read succeeds as empty.

### Knock-on fix in W05 (#4259, `ead4d41c2`)

`parkAndEmptyQueue` is the W05 call site of the same reader, and its last act is `setItem(QUEUE_NEEDS_ATTENTION_KEY, '[]')`. Rows it could not read were never copied into the orphan batch, so it would have deleted them. It now throws `QueueParkError`, matching what it already does when the orphan park itself fails. Surfaced by `tsc`, not by the suite — see that PR for the detail.

### Not defects (checked, no change)

- **Drain reads outside the storage lock.** `readTombstone()`/`readQueue()` at the top of `drainQueue` are unlocked while `enqueue` holds the lock across read+persist. Traced both interleavings: the reconcile step re-reads under the lock and rebuilds `kept` from `snapshotIds`, so a write enqueued at any point during a drain survives and a sent one is still removed. Already covered by *"a write enqueued during a drain whose re-read fails is still replayed"*.
- **`sentAt` / `attempts` mutated in place.** The persisted rows are the same object references from the snapshot, so the rewritten bounds and the attempt count do reach storage on the retained path.
- **`PERMANENT_STATUSES` excluding 401/403/408/429.** Deliberate and correct; each succeeds on a later attempt.

### Lesser finding (not fixed)

`quarantine` writes to a single slot, so a **second** corruption overwrites the first quarantined blob — the same single-slot shape round two fixed for the orphan park by making it an appended list. Rare enough (corruption twice, before anyone extracts the bytes) that I left it rather than widen this diff; say the word and it becomes an append.

### Verification

`apps/mobile`: **689 passed / 3 skipped, 55 files**; `tsc --noEmit` exit 0 (both verified on `cf55c29a5`). Both fixes written red-first — the park test failed with `landed === true` while the earlier row vanished, and the clear test left `[]` where two rows had been.

**Still not checked:** no on-device run. The airplane-mode → start → stop → reconnect path from the plan's Task 6 has never been executed on a phone.

PR #4253, opened 2026-08-30
docs(api): attribute SSO state-cookie protection to the HMAC binding, not SameSite (#4054)

### LINKED ISSUE #4054: Stale comment at `apps/api/src/routes/sso.ts` misattributes the SSO state-cookie protection to `SameSite=Lax`
### Description

Re-checked at HEAD — the comment has shifted a few lines from the plan doc's cited `:69` (this branch has added substantial code above it) but is still present, now at `apps/api/src/routes/sso.ts:72-74`, quoted exactly:

```
// proven M365 admin-consent flow (`routes/c2c/m365Auth.ts`). SameSite=Lax means
// the cookie is NOT attached on a cross-site top-level GET navigation to
// /callback, so the forged-login delivery fails the cookie/state match.
```

(Full surrounding block: `apps/api/src/routes/sso.ts:61-74`, describing the login-CSRF browser-binding cookie added to fix a forced-login vulnerability. Predates PR #4041, but PR #4041 is the branch on which this stale comment was flagged during review.)

This is factually wrong. `SameSite=Lax` cookies **are** sent on cross-site top-level GET navigations — that is the defining difference between `Lax` and `Strict`, and precisely why `Lax` is the recommended default for auth flows that need to survive a top-level redirect back from an IdP. A GET to `/callback?code=...&state=...` following a cross-site redirect (exactly the login-CSRF delivery vector this comment is describing) **would** carry a `Lax` cookie.

The actual protection is sound and is described correctly earlier in the same comment block: `state` is bound via the cookie's HMAC to the specific, unpredictable, single-use value generated at `/sso/login/:orgId`. An attacker who captures a `/callback?code=...&state=...` URL from their own IdP session and feeds it to a victim cannot forge a *matching* cookie — the cookie's HMAC binds to a `state` the attacker's browser never received. That mechanism works regardless of `SameSite`; the comment's closing sentence just names the wrong reason.

### Why it matters

Not a security bug — the actual defense (HMAC-bound single-use state) is intact and unaffected by this comment being wrong. But a wrong rationale left in place is a hazard for whoever reads it next: someone auditing or refactoring this code, reasoning from the comment, could conclude the `SameSite=Lax` attribute is load-bearing and "simplify" by relaxing the state-binding check on the (incorrect) belief that the cookie attribute alone already blocks cross-site delivery. Fixing the comment removes that trap without touching any code.

### Affected files

- `apps/api/src/routes/sso.ts:61-74` — the full comment block; the incorrect sentence is at `:72-74` specifically.

### Suggested approach

Rewrite the closing sentence to attribute the protection correctly, e.g.: "The protection is the HMAC binding between this cookie and the single-use, unpredictable `state` value generated at initiation — not `SameSite`, which does not block this delivery vector (`Lax` cookies are sent on cross-site top-level GET navigations)." Comment-only change; no code or test impact.

ToddHebebrand:
## What

Comment-only fix. Three comments in the SSO login-CSRF browser-binding code claimed `SameSite=Lax` is what blocks forced-login delivery, on the stated basis that a Lax cookie *"is NOT attached on a cross-site top-level GET navigation to /callback"*.

That is backwards. **`SameSite=Lax` cookies ARE sent on cross-site top-level GET navigations** — that is the defining difference between `Lax` and `Strict`, and it is precisely why `Lax` (not `Strict`) is the correct attribute for an OIDC flow that must survive the IdP's top-level redirect back to `/callback`. The forced-login delivery vector *is* a cross-site top-level GET, so a `Lax` cookie **would** be attached. SameSite is not the control here.

## What actually protects the flow (verified against the code)

Unchanged and intact — this PR only makes the comments say the true thing:

| Step | Where |
|---|---|
| `/sso/login/:orgId` mints an unpredictable, single-use `state`, then sets `breeze_sso_state` = `HMAC-SHA256(APP_ENCRYPTION_KEY \|\| SECRET_ENCRYPTION_KEY, "sso-login-state:" + state)` | `apps/api/src/routes/sso.ts:151-171` (`buildSsoStateCookieValue` / `buildSsoStateCookie`), set at `apps/api/src/routes/sso.ts:2510-2520` |
| `/sso/callback` constant-time compares the presented cookie against the HMAC recomputed from the URL `state`, **before** the session is consumed, and fails closed when the signing secret is unset | `apps/api/src/routes/sso.ts:2556` calling `isValidSsoStateCookie` (`apps/api/src/routes/sso.ts:197+`) |
| Net effect | A replayed `/callback?code=...&state=A` dies because the victim's browser holds no cookie bound to state `A` (it holds none, or one bound to its own state `B`), and the attacker cannot mint one without the server-side secret |

`HttpOnly` + `Path=/api/v1/sso/callback` + `Secure` (prod) + 10-minute `Max-Age` + `SameSite=Lax` remain ordinary cookie hygiene that narrows exposure. They are not the control that defeats forced-login.

**This is not a security fix** — no defense was missing. It removes a documentation trap: a future reader reasoning from the old comment could have concluded the `SameSite` attribute was load-bearing and "simplified" away the state-binding check.

## Scope note — three sites, not one

The issue cited one comment block. A sweep found the identical misattribution in two more places, all three fixed here:

1. `apps/api/src/routes/sso.ts:123-134` — the banner comment (the one the issue cited).
2. `apps/api/src/routes/sso.ts:2546-2552` — the same claim restated at the `/callback` binding check.
3. `apps/api/src/routes/sso.test.ts:1822-1826` — the `rejects a callback with NO binding cookie` case explained its own no-cookie setup as *"simulates the cross-site top-level navigation a SameSite=Lax cookie would not be attached to"* — the same wrong rationale, sitting on the test that proves the real control.

`apps/api/src/routes/sso.ts:2510` was checked and left alone: it names the attributes but already credits the HMAC, so it was never wrong. `apps/api/src/routes/c2c/m365Auth.ts` (the flow the banner cites as precedent) carries no equivalent claim.

## Verification

- `pnpm exec vitest run src/routes/sso.test.ts src/routes/sso.reauth.test.ts` — **214 passed / 214**, 2 files.
- `pnpm --filter @breeze/api exec tsc --noEmit` — **exit 0, zero output** (no OOM).
- `git diff` is comment-only across both files — machine-checked: every changed line begins with `//`, no cookie attribute, string literal, or executable statement touched.

## CI note

`Check Migrations` and `Trivy Image Scan` are red repo-wide right now — they fail on `main` and on unrelated PRs (#4227 tracks the first; PR #4236 fixes it). Not attributable to this change, which touches no migration and no dependency.

Closes #4054


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — comment-analyzer, code-reviewer, pr-test-analyzer. `silent-failure-hunter` and `type-design-analyzer` were deliberately skipped: the diff changes no error-handling path and introduces no types.

**Findings:** 3 reviewers → **0 findings** (0 critical, 0 important, 0 suggestions requiring change in this PR); 0 outstanding.

The load-bearing check here was comment-analyzer, since the failure mode this PR exists to prevent is a comment that is confidently wrong in a *new* way. It re-derived every claim from the code rather than from the PR body, and confirmed all four:

- `buildSsoStateCookieValue` (`apps/api/src/routes/sso.ts:152-161`) is `HMAC-SHA256(APP_ENCRYPTION_KEY || SECRET_ENCRYPTION_KEY, "sso-login-state:" + state)`; `isValidSsoStateCookie` (`:201-214`) length-checks then `timingSafeEqual` — genuinely constant-time.
- Ordering: binding check at `:2553` runs strictly before session consumption via `claimSsoCallbackIssuance` at `:2561`. "BEFORE" is accurate.
- Fail-closed in both directions: `/callback` returns false whenever the expected value is `null`; all four `buildSsoStateCookie` call sites (`:2136, :2261, :2399, :2514`) 500 rather than start an unbindable flow.
- The web-platform claim is correct per RFC 6265bis — `Lax` withholds only on cross-site *subresource*/non-safe requests, not on top-level GET navigation, so `Lax` is the only viable choice for an IdP redirect-back and would indeed be attached to a forced-login replay.

It also ruled the "cookie hygiene, not the forced-login defense" framing fair rather than an underclaim: `Lax` still does real work against non-navigational vectors, and the comment scopes its claim to *this* vector only. An independent repo-wide sweep found no remaining instances of the misattribution, and confirmed `sso.ts:2508-2511` was correctly left untouched (it lists the attributes but already credits the state match).

**Tests:** `apps/api` affected suites green — `vitest run src/routes/sso.test.ts src/routes/sso.reauth.test.ts` → **214 passed / 214**, 2 files. `pnpm --filter @breeze/api exec tsc --noEmit` → **exit 0, zero output** (no OOM).

pr-test-analyzer confirmed no new test is warranted (no behavioral diff) and verified the three existing binding tests are non-vacuous — the test helper `ssoStateCookieHeader` independently reimplements the HMAC instead of calling `buildSsoStateCookieValue`, so it is a real oracle rather than a tautology, and the wrong-state case exercises the `timingSafeEqual` compare rather than short-circuiting on length. One out-of-scope follow-up it flagged, **not** for this PR: the "signing secret unconfigured" fail-closed branch has no test in either direction (`/login` 500 or `/callback` reject) — `APP_ENCRYPTION_KEY` is set once in the global `beforeEach` (`sso.test.ts:623`) and never unset. Pre-existing; worth a small follow-up issue.

**CI as-of this comment:** 33 green, 7 still queued (Test API, Type Check, Test Web, Test Agent (race), Test Agent (Windows), Analyze (javascript-typescript), Trivy Image Scan), CodeQL skipping.

`Check Migrations` is red — **repo-wide, not attributable to this change.** Verified failing on `main` itself (runs `33319775782`, `33273323158`); #4227 tracks it and PR #4236 fixes it. `Trivy Image Scan` is red repo-wide for the same reason. This PR touches no migration and no dependency — the diff is comment text in two files.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4054 — both are the maintainer's call.

Issue #3139, opened 2026-08-04
[CI] No test-mobile job — the entire apps/mobile vitest suite never runs in CI


ToddHebebrand:
## Description

`ci.yml` has required `test-api`, `test-web`, and `test-agent` jobs, but no job runs the `apps/mobile` vitest suite. All 28 mobile test files — including ~30 tests just added by the fixes for #3114, #3115, #3117, #3118, and #3119 — run local-only. A regression in mobile code merges green.

Flagged independently by the reviewers on three separate PRs (#3124, #3125, #3131).

## Proposed Fix

Add a `test-mobile` job to `.github/workflows/ci.yml` mirroring `test-web` (pnpm install, `pnpm --filter breeze-mobile test`, plus `tsc --noEmit` in `apps/mobile`). Consider making it required once it proves stable.

## Affected Files

- `.github/workflows/ci.yml`

## Reported By

Review findings on #3124, #3125, #3131 (follow-up from the #3114–#3119 batch)

PR #4254, opened 2026-08-30
fix(ci): make test-mobile a blocking check in ci-success (#3941)

### LINKED ISSUE #3941: test-mobile is not a required CI job, so mobile test failures cannot block a merge
Surfaced during the Sentry triage sweep (PR #3910).

`.github/workflows/ci.yml` runs `test-mobile` (vitest + `tsc`) but it is **not a required check**, so a red mobile suite does not block a merge. PR #3910 adds 25+ new tests guarding the release-build Sentry DSN enforcement — those tests can go red without anyone being stopped.

This is the same class of gap that let mobile ship with zero crash telemetry for 90 days: the guard existed, nothing enforced it.

Related: #3139.

**Suggested fix:** add `test-mobile` to the required checks for `ci-success`.

ToddHebebrand:
Closes #3941

## Problem

`.github/workflows/ci.yml` defines a `test-mobile` job (vitest + `tsc --noEmit` for `apps/mobile`) that runs on every PR — but it was wired into **nothing**. Before this PR, `git grep test-mobile -- .github/` on `origin/main` returned exactly one hit: the job definition itself.

A red mobile suite therefore could not block a merge. PR #3910 added 25+ tests guarding the release-build Sentry DSN enforcement; those could all go red silently. Same class of gap that let mobile ship with zero crash telemetry for 90 days: the guard existed, nothing enforced it.

## The mechanism that actually gates (verified, not assumed)

`ci-success` is the aggregate check. **Verified against the live branch ruleset:** `main` is protected by ruleset `Protect main branch` (active), whose `required_status_checks` list contains exactly one context — **`CI Success`**, the `name:` of the `ci-success` job. So wiring into `ci-success` is the entire merge gate, and nothing else needed changing. (The classic `branches/main/protection` API returns 404 for this repo — protection is via rulesets, not the legacy endpoint. Worth knowing before concluding "main isn't protected".)

A job is only truly blocking when wired in **three** coordinated places:

1. the `needs:` array of `ci-success`;
2. a `<JOB>_RESULT: ${{ needs.<job>.result }}` entry in the `env:` block of the `Check all jobs passed` step;
3. an assertion `[[ "${<JOB>_RESULT}" != "success" ]] || \` inside that step's **unconditional** bash if-condition.

**(3) is what fails the job.** Adding a job to `needs:` alone is a vacuous fix that looks correct — the step's `run:` body only inspects the named `*_RESULT` variables. This PR adds all three, positioned after `test-web` so file order matches job-definition order.

`test-mobile` is unconditional in its own right (no `if:`, no `paths:` filter, no `continue-on-error:`), so it is wired exactly like the sibling `test-web` / `test-portal` jobs — **not** like `smoke-test`, which is deliberately PR-exempt via the separate `IS_PR` guard further down the script.

Also removes a now-false comment above the job ("Not yet wired into ci-success / branch protection — see #3139. Promote to required once the job proves stable."). Its stated promotion condition is met — see the green-on-main evidence below.

## Verification of "which mechanism" — evidence

I did not take the diff at face value. I extracted the `run:` script from both the pre- and post-change workflow and executed the real gate logic with every `*_RESULT` set to `success` except mobile:

| script | `TEST_MOBILE_RESULT` | exit | meaning |
|---|---|---|---|
| pre-change | `failure` | **0** | the bug — mobile red, gate passes |
| post-change | `failure` | **1** | blocked |
| post-change | `success` | 0 | passes |
| post-change | `skipped` | **1** | blocked |

A structural parse of both files (needs array / env map / blocking condition, resolved back to job names) confirms:

- blocking jobs: **28 → 29**, the only addition being `test-mobile`;
- **no job removed** from `needs:`, the env block, or the blocking condition;
- **no other job's gating status changed** (each job's `(blocking, conditional)` pair is identical before and after);
- `smoke-test` remains in its `IS_PR`-guarded branch, unchanged.

`bash -n` passes on the extracted script in both versions, and the repo's own **Workflow Lint** (actionlint) job is green on this branch.

## Is `test-mobile` currently green on `main`?

**Yes.** Making a red job blocking would immediately block every PR, so I checked before proceeding. `Test Mobile` concluded `success` in the last four `ci.yml` runs on `main`: runs `33273323158`, `33267501194`, `33266598965`, `33235607846` — and again on this PR.

## Regression guard

`apps/api/src/config/ciSuccessGatingContract.test.ts` (new, runs in the required **Test API** job) enforces the contract for **every** job, not just this one:

- every job in `needs:` has a `*_RESULT` env var;
- every job except the documented exemption (`smoke-test`) is asserted in the unconditional blocking condition;
- the `smoke-test` exemption is asserted as its **exact** guard expression, so an operator inversion (`==` for `!=`, `||` for `&&`) is caught;
- **no blocking job may carry job-level `continue-on-error`** — that makes `needs.<job>.result` report `success` even when the job fails, which is precisely how `smoke-test` is non-blocking on PRs today. Step-level `continue-on-error` (the legitimate `db:check-drift` one in `lint`) is correctly not flagged;
- **every top-level job in `ci.yml` must be in `needs:` or in a documented `UNGATED_JOBS` set** — without this, a job missing from `needs:` *entirely*, which was test-mobile's actual bug, stays invisible;
- an explicit `test-mobile` regression assertion for #3941;
- parse-sanity assertions so a formatting change to `ci.yml` fails loudly instead of emptying the parsed sets and passing vacuously.

**Every assertion was proven to fail against a deliberately broken workflow, not merely to pass against the good one:**

| injected defect | result |
|---|---|
| `ci.yml` reverted to its pre-PR state | **2 tests fail** — incl. the completeness test, so this suite *would* have caught #3941 |
| `test-mobile` added to `needs:` only (the vacuous fix) | **3 tests fail**, incl. *"Being in needs: alone does not fail the gate."* |
| `continue-on-error: true` added to `test-mobile` | **1 test fails** |
| a brand-new job added to `ci.yml` | **1 test fails** |
| smoke-test's `IS_PR` guard inverted (`!=` → `==`) | **1 test fails** |
| unmodified workflow | **7 pass** |

## Pre-existing gaps this PR deliberately does NOT close

The completeness test surfaced four jobs that run on every PR but are absent from `ci-success`'s `needs:` — the same defect as #3941, and they cannot block a merge today: **`check-migrations`**, **`lint-agent`**, **`test-agent-race`**, **`agent-windows-manifest-guard`**. (`rust-check-windows` is a fifth, but is deliberately excluded with a comment at its definition: path-filtered and usually skipped.)

They are pinned in `UNGATED_JOBS` with a comment so the gap is **visible and documented** rather than silent, and so any *new* job forces a conscious decision. Promoting them is out of scope here — notably `check-migrations` is red repo-wide right now, so making it blocking would block every PR. Worth a follow-up issue.

## Test / typecheck evidence

- `apps/api` `src/config/` suite: **14 files, 488 tests passed** (`vitest run --pool=threads --maxWorkers=2 src/config/`)
- `pnpm exec tsc --noEmit` in `apps/api`: **exit 0, clean**
- `pnpm exec eslint src/config/ciSuccessGatingContract.test.ts`: **exit 0, clean**
- Rebased onto current `origin/main` (`e906cee89`) and re-verified, so this is not a stale-base green.

## Note on unrelated red checks

`Check Migrations` and `Trivy Image Scan` are red repo-wide right now — they fail on `main` and on unrelated PRs. #4227 tracks the first and PR #4236 fixes it. Neither is attributable to this change; this PR touches only the `ci-success` aggregate wiring, one comment, and one new test file.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter (3 agents, parallel).

**Findings:** 6 raised → 4 addressed in `ada96010b`, 1 rejected with evidence, 1 declined as out of scope; 0 outstanding.

Addressed:
1. **Stale comment (code-reviewer, Important).** `ci.yml` still read *"Not yet wired into ci-success / branch protection — see #3139. Promote to required once the job proves stable"* directly above the job this PR promotes. Replaced. Its stated condition is met — `Test Mobile` was green on the last four `main` runs.
2. **Weak smoke-test exemption assertion (all three agents).** It only checked that `SMOKE_TEST_RESULT` and `IS_PR` each appeared *somewhere*; an operator inversion kept both substrings and passed. Now asserts the exact guard expression.
3. **Job-level `continue-on-error` defeats the gate (pr-test-analyzer, Critical).** It makes `needs.<job>.result` report `success` even when the job fails — live in this workflow, since that is exactly how `smoke-test` is non-blocking on PRs. No blocking job may now carry it; step-level (`db:check-drift` in `lint`) is correctly not flagged.
4. **A job missing from `needs:` entirely was invisible (silent-failure-hunter, Critical).** Every assertion only iterated jobs already inside `needs:` — so the suite would *not* have caught this very issue. Every top-level job must now be in `needs:` or in a documented `UNGATED_JOBS` set.

Rejected: **"branch protection has no required checks — `ci-success` gates nothing" (silent-failure-hunter, Important).** Not correct. The legacy `branches/main/protection` endpoint 404s for this repo, but ruleset `Protect main branch` is **active** and its `required_status_checks` contains exactly one context: **`CI Success`**. Wiring into `ci-success` is the whole gate.

Declined (out of scope, flagged for follow-up): four jobs — `check-migrations`, `lint-agent`, `test-agent-race`, `agent-windows-manifest-guard` — run on every PR but are absent from `needs:`, the same defect as #3941. They are now pinned in `UNGATED_JOBS` so the gap is documented rather than silent. Promoting them is a separate decision; `check-migrations` is red repo-wide today, so making it blocking would block every PR.

**Tests:** every new assertion was verified to **fail** against a deliberately broken workflow, not merely to pass against the good one — reverting `ci.yml` to its pre-PR state fails 2 tests; the vacuous `needs:`-only fix fails 3; injected `continue-on-error`, a new ungated job, and an inverted `IS_PR` guard each fail 1; unmodified workflow passes 7. `apps/api` `src/config/`: 14 files / 488 tests green. `tsc --noEmit` exit 0. `eslint` exit 0. Rebased onto current `origin/main` (`e906cee89`) and re-verified, so this is not a stale-base green.

**CI:** 53 pass / 2 fail / 1 skipped. **`CI Success` — the only context the `main` ruleset requires — is green**, as are all 4 integration shards and Smoke Test. The two failures are `Check Migrations` and `Trivy Image Scan`, both red repo-wide: `Check Migrations` also fails on the latest `main` run (#4227 tracks it, PR #4236 fixes it). Neither is attributable to this change.

Issue #3922, opened 2026-08-24
Hosted BYO-LLM phase 2: third-party endpoints behind an approved-provider allowlist + egress controls


ToddHebebrand:
Phase 2 of per-partner LLM BYOK (#3228, shipped 2026-08-23 as Anthropic-key-only with the endpoint fixed to api.anthropic.com). Let hosted partners route their AI traffic to third-party or self-hosted LLM endpoints (OpenRouter, LiteLLM, vLLM with the Anthropic-compatible /v1/messages surface).

**Decided direction (owner, 2026-08-23): an approved-URL allowlist, not free-form base URLs.** A curated list of vetted provider endpoints (platform-maintained; per-partner selection from the list), so the SSRF surface is bounded by review rather than by runtime validation alone. Free-form URLs, if ever offered, come later and only on top of the full egress-control stack.

Required security work regardless of allowlist:
- Private-network blocking: resolve and refuse RFC-1918/loopback/link-local and cloud metadata (169.254.169.254) targets.
- DNS-rebinding pinning: connect to the vetted IP, not a re-resolved one.
- Redirect discipline: refuse or re-validate HTTP redirects.
- Per-partner egress audit log (which partner's AI traffic went to which host).
- The hosted ANTHROPIC_BASE_URL fail-closed guard (#1412) stays intact for the platform path; partner endpoint selection is a separate, allowlist-bound mechanism.

Product/support work:
- Tool-call fidelity matrix: the agent loop is tool-heavy; each allowlisted provider needs a verified tool-calling test before listing.
- Per-partner pricing overrides for cost tracking (MODEL_PRICING covers Anthropic models only; the MCP_LLM_PRICE_* precedent exists).
- Schema is ready: partner_llm_configs already carries provider/base_url columns behind phase-1 CHECK constraints — phase 2 relaxes them in a new migration.

Origin: Discord request in #3228; discussion #505; spec `docs/superpowers/specs/ai-mcp/2026-08-23-per-partner-llm-byok-design.md` (Non-goals + §8).

---

Tracking issue for **Hosted BYO-LLM phase 2: third-party endpoints behind an approved-provider allowlist + egress controls** (`byo-llm-phase2`).

- Plan: `docs/superpowers/plans/ai-mcp/2026-08-26-byo-llm-phase2-allowlist-egress.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
[...]

PR #4255, opened 2026-08-30
fix(api): route release-manifest fetches through the SSRF-guarded redirect helper (#3649)

### LINKED ISSUE #3649: releaseArtifactManifest fetches manifest + signature with naive redirect:'follow' and no SSRF guard
Found while fixing the SSRF-guard adoption in #3645 (security review §1.7).

## What

`apps/api/src/services/releaseArtifactManifest.ts:419`:

```ts
async function fetchSmallBuffer(url: string, label: string): Promise<Buffer> {
  const resp = await fetch(url, { redirect: "follow" });
```

This is a raw global `fetch` with **naive redirect following**, used to retrieve the release **manifest and its signature**. It does not go through `services/urlSafety.ts`.

Two distinct concerns:

1. **No SSRF guard.** Unlike `safeFetch`, there is no DNS-resolve-and-pin, no metadata/link-local/RFC1918 blocking, and the Host header is not discarded. `redirect: 'follow'` in particular is the classic SSRF bypass — even if the initial URL is validated, the redirect target is not.
2. **Supply-chain adjacency.** These are the manifest and signature used to verify release artifacts. An attacker-influenced redirect chain here targets the verification input itself.

Mitigating context (why this is a follow-up and not a hotfix): the URL is derived from server configuration rather than tenant input, and the fetched manifest is Ed25519-verified downstream. So this is defense-in-depth, not a known live exploit path.

## Suggested fix

#3645 added `safeFetchFollowingRedirects()` to `services/urlSafety.ts` — a bounded (`SAFE_FETCH_MAX_REDIRECTS = 5`) redirect loop that re-invokes `safeFetch` per hop, so **every** hop is independently DNS-resolved and IP-pinned, with relative-Location resolution, non-http(s) scheme rejection, and cross-origin credential stripping. It exists precisely so this pattern is not reimplemented naively per caller.

Swapping `fetchSmallBuffer` onto it should be close to a drop-in, since GitHub asset URLs redirect (which is presumably why `redirect: 'follow'` was set here in the first place).

Worth also considering whether the `backupSsrfAdoption.test.ts` source-scanning guard should be generalized beyond the backup surface to catch raw `fetch(` on other outbound paths like this one.

ToddHebebrand:
Closes #3649

## The hole

`apps/api/src/services/releaseArtifactManifest.ts` fetched the release artifact **manifest and its Ed25519 signature** with a raw global fetch:

```ts
const resp = await fetch(url, { redirect: "follow" });
```

`redirect: 'follow'` is the classic SSRF bypass — the initial URL may be trusted, but the redirect target is dialed with no validation at all: no DNS-resolve-and-pin, no RFC1918/loopback/link-local/metadata blocking, Host header not discarded. And these two responses are the *inputs to release artifact verification*, so the exposure is supply-chain adjacent rather than merely outbound egress.

Not a known live exploit path (the URL comes from server config, and the manifest is Ed25519-verified downstream) — this is the defense-in-depth follow-up #3645 flagged.

## The fix

Route `fetchSmallBuffer` through the **existing** `safeFetchFollowingRedirects` in `services/urlSafety.ts`. That helper was added by #3645 precisely so this pattern is not reimplemented naively per caller — **no new helper or module was written for this PR.** Each hop is a fresh `safeFetch`, so every intermediate URL is independently DNS-resolved, filtered, and IP-pinned before the socket is dialed; hop budget is 5; relative `Location` is resolved; credential headers are stripped cross-origin; exhausting the budget throws rather than silently returning a 3xx.

Also passes `maxBytes: MAX_MANIFEST_BYTES`, upgrading the manifest size limit from a post-hoc buffer check to a streaming ceiling. **Both** pre-existing size checks are retained as a backstop (demonstrated live, below), and every existing error string is byte-identical — callers and tests key on them.

`fetchSmallBuffer` is this module's only outbound HTTP call, invoked twice from `verifyGithubReleaseArtifactBuffer` — once for the manifest, once for the signature — so converting the one function closes both. A source-scan test now enforces that it stays closed.

Production diff is **8 lines**.

## Tests

New `releaseArtifactManifest.redirect.test.ts`, modelled on the existing `recoveryMediaService.redirect.test.ts`. It deliberately does **not** mock `./urlSafety` — the guard runs for real; only DNS (`__setLookupForTests`) and the socket layer are stubbed, so every dialed host is recorded and assertable.

1. **The legitimate chain still works** — a normal 302 `github.com` → `objects.githubusercontent.com` is followed and the manifest + signature verify end to end. This case is the point: GitHub release assets *require* following redirects, and an over-tight guard here would break agent installer downloads rather than secure them.
2. **Redirect to cloud metadata is refused** — `Location: http://169.254.169.254/…` raises `SsrfBlockedError`, and the recorded requests prove the metadata host was never dialed.
3. **Redirect to a host that DNS-resolves into RFC1918 is refused** — the deeper bypass, where the `Location` looks public and only resolution reveals `192.168.1.10`.
4. **Oversized manifest aborts at the streaming ceiling** — 1MiB+1 raises `ResponseTooLargeError` rather than handing a truncated buffer onward to signature verification.
5. **Anti-regrowth source scan** — no raw `fetch(`, no `globalThis/window.fetch(`, and `safeFetchFollowingRedirects` must actually be *called*, not merely imported.

### Red-first controls (both mutations verified, not assumed)

- **Whole fix reverted, tests kept:** 4/4 failed, both security cases specifically on `expected Error: Failed to fetch… to be an instance of SsrfBlockedError`. Restored → green.
- **Only the `maxBytes` argument removed:** the ceiling test failed with `expected Error: release artifact manifest is too large to be an instance of ResponseTooLargeError`. That result is doubly useful — it proves the new test discriminates, *and* it proves the retained post-hoc size checks are a live backstop rather than dead code.

Honest caveat: under the first control the raw global `fetch` uses undici and so escapes the `http/https.request` stub entirely, meaning it errors out rather than demonstrating a completed fetch of metadata. The control is a genuine discriminator; it is not a live exploit reproduction.

## The interesting failure this surfaced

CI's **Test API** caught something local affected-file runs did not. Five `fetchRegularMsi` cases in `installerBuilder.test.ts` intercept the network with `vi.stubGlobal('fetch', …)`, which worked only because `fetchSmallBuffer` used the global fetch. `safeFetchFollowingRedirects` deliberately does not — it dials Node's `http`/`https` directly so each hop can be pinned. So the stub stopped intercepting, and those tests **fell through to the live internet**, failing on a real 404 from github.com.

Worth stating plainly: they were not merely red, they were *unhooked* — a harness that had quietly stopped covering anything. Fixed by routing the guarded helper back through the global stub (the same idiom `recoveryMediaService.test.ts` already uses), leaving the five cases otherwise untouched. One assertion pinned `{ redirect: 'follow' }` — the exact argument this issue removes — so rather than deleting it, it now pins the stronger property: the signature goes through the guarded helper, under the byte ceiling.

## Verification

- 11 affected suites, **288 tests green** (installerBuilder, binarySync, binarySyncManifestRefusal, releaseTrustChain.e2e, recoveryMediaService ×2, urlSafety, backupSsrfAdoption, agentVersions, and both manifest suites).
- `tsc --noEmit` on `@breeze/api` — clean. Note: it OOMs at the default heap on a loaded host (known repo condition); passes under `--max-old-space-size=12288`. CI **Type Check** is green.
- `eslint` on all four changed files — clean.
- All four **Integration Tests** shards green.

## Two things a reviewer should weigh

**1. A pre-existing `#1105` tripwire becomes visible (no behavior change).** `safeFetch` calls `assertOutsideHeldDbContext`. Four caller paths reach this fetch inside a held DB context — the authenticated installer routes and both public installer-download routes in `routes/enrollmentKeys.ts` (which wrap `serveInstaller` in `withSystemDbAccessContext`), via `installerBuilder.ts`. The raw `fetch` did the identical network round-trip inside the identical held transaction; the tripwire simply could not see it. **This PR surfaces an existing problem rather than creating one.**

It is warn-only: `assertOutsideHeldDbContext` throws only under `DB_CONTEXT_TRIPWIRE_STRICT`, which no CI job sets, and the Sentry report is deduped per call site per process.

The repo does have a sanctioned mechanism — `SELF_MANAGED_DB_CONTEXT_ROUTES` in `middleware/selfManagedDbContextRoutes.ts` — and #3922 is a direct precedent that registered routes there in the same PR as its `safeFetch` adoption. **I deliberately did not do that here**, because for these routes it is not a mechanical registration: it requires restructuring `serveInstaller`'s transaction boundaries on a *public, unauthenticated* download path where the current `withSystemDbAccessContext` wrap deliberately keeps the enrollment-key lookup and usage bump in one RLS transaction. That is a high-blast-radius change that deserves its own PR and its own review, not a bolt-on to a security fix. Flagging it so the call is conscious rather than implicit — happy to do it as a follow-up.

**2. Sibling holes remain, and one is closer than it looks.** The same naive pattern exists in `installerBuilder.ts` (4 raw fetches, 3 with `redirect: 'follow'`, fetching the actual MSI/pkg bytes) and `binarySync.ts` (4).

The `binarySync.ts` one deserves emphasis: `fetchReleaseAssetBuffer` → `fetchTrustedReleaseManifest` is a **structurally parallel, independently unguarded implementation that fetches the very same two files** (`release-artifact-manifest.json` and `.ed25519`), on a path that runs at boot under `BINARY_SOURCE=github`. So #3649's *module* is fixed, but the same two artifacts are still fetched unguarded by a different code path. Converting those is a larger change touching agent binary delivery, and #3649 scoped itself to this module — but a reviewer should not read this PR as "the manifest can no longer be fetched unguarded anywhere." Recommend a follow-up issue.

## CI note

`Check Migrations` and `Trivy Image Scan` are red repo-wide right now — they fail on `main` and on unrelated PRs (#4227 tracks the first; PR #4236 fixes it). Not attributable to this change, which touches no migrations and no images.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, parallel). Skipped type-design-analyzer: no new types.

**Findings:** 0 critical. 2 important + 6 suggestions raised → 4 addressed in `03bf897f6`, 2 consciously declined and documented in the PR body.

Addressed:
- *pr-test-analyzer* — the new `maxBytes` streaming ceiling had no test, and it pre-empts rather than is backed by the post-hoc length checks. Added an overrun case asserting `ResponseTooLargeError`.
- *pr-test-analyzer* — the source-scan guard's negative lookbehind exempted `globalThis.fetch(`, and matching the bare identifier would accept an orphaned import. Both tightened.
- *comment-analyzer* — this file cites the originating issue on every other justification comment; the new one now cites #3649.
- *silent-failure-hunter* — added the message-content assertion the RFC1918 case was missing versus its metadata-IP sibling.

Declined, with reasoning in the PR body: registering the installer routes in `SELF_MANAGED_DB_CONTEXT_ROUTES` (it requires restructuring `serveInstaller`'s transaction boundaries on a public unauthenticated download path — high blast radius, deserves its own PR), and converting the sibling raw fetches in `installerBuilder.ts` / `binarySync.ts` (out of #3649's scope; `binarySync` notably has a parallel unguarded fetch of the *same two artifacts*, so this PR should not be read as closing that path).

**Separately, CI caught a real miss my local runs did not:** five `fetchRegularMsi` cases in `installerBuilder.test.ts` stub the global `fetch`, which the guarded helper no longer routes through — so they were making live requests to github.com and failing on a real 404. They were unhooked, not merely red. Fixed in `464cbb58f`.

**Red-first controls:** whole fix reverted → 4/4 new tests red, both security cases on `expected … to be an instance of SsrfBlockedError`; `maxBytes` argument alone removed → ceiling test red, which also proves the retained post-hoc checks are a live backstop rather than dead code.

**Tests:** 288 passed across the 11 affected suites (`installerBuilder`, `binarySync`, `binarySyncManifestRefusal`, `releaseTrustChain.e2e`, `recoveryMediaService` ×2, `urlSafety`, `backupSsrfAdoption`, `agentVersions`, both manifest suites), single-fork. `tsc --noEmit` clean (needs `--max-old-space-size=12288`; OOMs at default heap on a loaded host). `eslint` clean.

**CI as of this comment:** `CI Success`, `Test API`, `Type Check`, `Lint`, all 4 `Integration Tests` shards, `Smoke Test` and `BINARY_SOURCE=github download path` all green; nothing queued. `Check Migrations` and `Trivy Image Scan` are red **repo-wide** — they fail on `main` and on unrelated PRs (#4227 tracks the migrations one, PR #4236 fixes it) — not attributable to this change, which touches no migrations and no images.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3649 — both are the maintainer's call.

PR #4256, opened 2026-08-30
fix(api): return 404, not 403, for an unusable Mcp-Session-Id (#3744)

### LINKED ISSUE #3744: MCP Streamable HTTP returns 403 instead of 404 for expired/unknown Mcp-Session-Id, blocking the spec's client recovery path
## Summary

The non-initialize `POST /api/v1/mcp/sse` handler treats two different conditions identically:

1. **No Redis entry exists** for the supplied `Mcp-Session-Id` (including after normal TTL expiry).
2. **An entry exists but belongs to a different authenticated principal.**

Both return HTTP **403** with `{"code":-32001,"message":"Mcp-Session-Id principal mismatch"}`.

Per the Streamable HTTP session-management spec, case 1 must be **404**:

> **3.** The server **MAY** terminate the session at any time, after which it **MUST** respond to requests containing that session ID with HTTP 404 Not Found.
> **4.** When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached.

(<https://modelcontextprotocol.io/specification/2025-06-18/basic/transports>, Session Management)

Returning 403 removes the spec's recovery path. A compliant client that receives 404 transparently re-initializes and self-heals; one that receives 403 has no defined recovery and stays broken until the client process is restarted. In my case a long-running MCP client went from working to returning this error on **every** subsequent tool call for hours, while `GET /api/v1/mcp/sse` continued to return 200 — so it looks connected while every call fails.

## Why the branch is reachable in normal operation

`apps/api/src/routes/mcpServer.ts`. The Redis entry is written with a TTL **only** on `initialize`:

```ts
const MCP_SESSION_TTL_SECONDS = envInt('MCP_SESSION_TTL_SECONDS', 11 * 60);
...
if (isInitialize) {
  trustedSessionId = mintMcpSessionId();
  await redis.setex(`${MCP_SESSION_REDIS_PREFIX}${trustedSessionId}`, MCP_SESSION_TTL_SECONDS, principalKey);
}
```

Searching `apps/api/src` for `MCP_SESSION_REDIS_PREFIX` returns exactly four hits: the constant (`:645`), two `redis.get` reads (`:586`, `:713`), and that single `setex` (`:676`). There is no `expire`/`pexpire`/second `setex`, so **the TTL is never refreshed on subsequent requests**. Ordinary expiry ~11 minutes after `initialize` is therefore sufficient to reach the `!storedPrincipal` branch, regardless of how actively the session is being used.

To be precise about causation: expiry is *sufficient* to produce this, but Redis eviction, restart, flush, or a storage misconfiguration reach the same branch. I have not instrumented a single session's key lifecycle end-to-end, so I'm reporting the protocol-handling defect rather than asserting TTL expiry as the unique cause of my incident.

## Proposed change

```ts
if (!storedPrincipal) {
  // expired / unknown / terminated session -> 404 so the client re-initializes
  return c.json({ jsonrpc: '2.0', id: pre.body.id ?? null,
    error: { code: -32001, message: 'Mcp-Session-Id not found' } }, 404);
}
if (storedPrincipal !== principalKey) {
  return c.json({ jsonrpc: '2.0', id: pre.body.id ?? null,
    error: { code: -32001, message: 'Mcp-Session-Id principal mismatch' } }, 403);
}
```

This keeps the MED-1 property from the comment above the constant: genuine cross-principal session reuse still 403s and can still be logged as an audit event.

**Tradeoff worth your call, not mine.** Splitting the responses introduces a small existence oracle — 404 means "no such session", 403 means "exists, wrong principal". With 16 random bytes from `mintMcpSessionId()` blind probing isn't practical, but it does reveal liveness for an ID that leaked some other way (logs, telemetry, a proxy). If you'd rather not expose that at all, returning **404 for both** while logging the mismatch internally also satisfies the spec's recovery contract and conceals existence entirely. I don't know which you'd prefer, hence an issue rather than a PR.

## On the TTL itself

I'm **not** proposing changing `MCP_SESSION_TTL_SECONDS` or making it sliding. The comment says it "matches OAuth access-token lifetime plus a small buffer", and refreshing on every request would convert a deliberately bounded lifetime into an indefinitely renewable one. The spec explicitly permits terminating a session whenever the server likes. The defect is the *status code* for an already-terminated session, not its lifetime.

## Environment

Self-hosted, v0.105.1, remote MCP over Streamable HTTP with an API key. `MCP_SESSION_TTL_SECONDS` unset (so the 660s default). Redis reachable and healthy — an authenticated `INFO keyspace` shows a populated `db0`, with zero `mcp-session:*` keys present at the time every call was 403ing.

ToddHebebrand:
## Summary

`POST /api/v1/mcp/sse` (Streamable HTTP) treated two different conditions as one 403:

1. no Redis entry for the presented `Mcp-Session-Id` — unknown, terminated, or aged out via TTL;
2. an entry that exists but belongs to a different authenticated principal.

Per [MCP Streamable HTTP 2025-06-18, Session Management](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports) rules 3–4, case 1 must be **404** — that is the signal that tells a compliant client to re-initialize. **403** reads as "authenticated but forbidden", has no defined recovery, and dead-ends the client until its process restarts. The reporter's symptom matches exactly: every tool call 403'd for hours while `GET /sse` kept returning 200, so the connection looked healthy while nothing worked.

The branch is reachable in **ordinary operation**, not just under attack. `mcp-session:*` is written with `MCP_SESSION_TTL_SECONDS` (default 660s) **only on `initialize`**, and nothing ever refreshes it — so ~11 minutes after `initialize`, every subsequent call hit this branch.

## The one design decision, and why

The issue explicitly left this to us: split the responses (404 unknown / 403 mismatch), or return **404 for both**?

**This PR returns 404 for both**, with an identical body.

Splitting them would create a session-existence oracle — a distinct status confirms "this session id is live" for an id that leaked through logs, telemetry, or a proxy. `mintMcpSessionId()` uses `randomBytes(16)`, so blind probing is infeasible, but 128 bits of entropy does not mitigate *disclosure*. And a caller presenting someone else's session has no legitimate recovery other than re-initializing under its own principal — which is exactly what 404 tells it to do. This also matches how this repo already handles object-level authorization on unguessable ids elsewhere (inaccessible orgs in `orgTicketSettings`, scoped saved filters, cross-tenant configuration policies all collapse "absent" and "not yours" into 404).

Both responses are built from **one expression**, not two copies of a literal — the no-oracle property holds only while they are indistinguishable, and two literals are two things a later edit can drift apart.

**MED-1's property is unchanged.** The request is still rejected before dispatch, and the mismatch case is still reported server-side.

*(Design decision was taken to a two-model quorum per CLAUDE.md; both independently landed on uniform 404, and the repo-precedent argument above came out of that review.)*

## The cost of 404, and what this PR does about it

404 is the spec's self-heal signal — which is precisely what makes a **mass session-store loss** invisible. A Redis flush, an eviction under memory pressure, or a failover to an empty replica does **not** throw, so none of them reach the 503 paths; they present as ordinary unknown sessions. Under the old 403 such an event was *loud* (every client stayed broken until restarted, someone filed a ticket). Under 404 every client silently re-initializes and the incident passes unnoticed.

The masking vector predates this PR — `main` had the same blind spot with zero logging — but the *consequence* of hitting it is what this change alters, so it is this PR's to close:

- **Unknown/expired 404s** are counted per process; **one** Sentry event (`mcp_session_unknown_rate_high`) is raised per 60s window once the rate exceeds what ordinary expiry can produce (`MCP_UNKNOWN_SESSION_ALERT_THRESHOLD`, default 50). Per-request telemetry is not an option — every session ages out this way by design, so a log line per 404 is noise with no operator action attached. This is a "something ate the session store" tripwire, not a metric.
- **Principal mismatch** now goes through `captureMessage` with a registered event code (`mcp_session_principal_mismatch`) rather than `console.warn` alone — nothing in this app forwards console output to Sentry, so a candidate session-probe signal was previously grep-only. Unbounded detail stays on the console because `scrubEvent` strips message/extra from outbound events and principal ids are too high-cardinality to be tags.

Both codes are registered in `sentryEventCodes.ts`, per that file's closed-set contract.

## Deliberately NOT changed

These were checked explicitly, because turning any of them into a 404 would be the real regression here:

| Case | Status | Why unchanged |
|---|---|---|
| API key missing `ai:read` | **403** | Genuine authz failure. Lives in `preflightMcpRequest`, still runs *before* session validation. |
| Missing / non-`mcp-`-prefixed session header | **400** | Client sent something the transport can never accept; re-initializing wouldn't help. |
| Session store unreachable (Redis null **or** `get` throwing) | **503** | A Redis outage must **not** make every client re-initialize in a loop. |
| `MCP_SESSION_TTL_SECONDS`, non-sliding | unchanged | The reporter explicitly did not ask for a lifetime change, and making it sliding would convert a deliberately bounded lifetime into an indefinitely renewable one. The defect was the status code for an already-terminated session, not its lifetime. |
| `POST /message`, legacy `GET /sse` | unchanged | Different transport; no `Mcp-Session-Id` header semantics. |

## Tests

11 new cases in `apps/api/src/routes/mcpServer.streamable.test.ts`, plus one updated MED-1 assertion in `mcpServer.test.ts`:

- 404 for an unknown session id
- **404 after the session ages out of Redis (TTL lapse) — the core #3744 regression**
- 404 for a different principal's session, asserting no `Mcp-Session-Id` header is echoed back and the body does not say "mismatch"
- 404 is *recoverable*: expired → 404 → re-`initialize` → new id → `tools/list` 200, i.e. the spec's recovery path end-to-end
- a notification (no JSON-RPC `id`) on an expired session → 404, **not** a silent 202 — this locks the validate-before-notification-dispatch ordering
- `redis.get` **rejecting** → 503 (distinct branch from `getRedis() === null`)
- `redis.setex` rejecting on `initialize` → id still minted, next call 404
- the mismatch raises the operator signal; an ordinary expiry does not
- guard: missing `ai:read` still **403**, not 404 — sent *with* a valid session id so it proves the scope check still wins
- guard: missing / client-supplied session header still **400**
- guard: `getRedis() === null` still **503**, not 404

### Red-first control

Run against the **unfixed** implementation (impl stashed, tests kept) to confirm the tests discriminate rather than merely confirm:

```
× returns 404 for an unknown Mcp-Session-Id so the client re-initializes
× returns 404 when the session expired out of Redis (TTL lapse)
× returns 404 (not 403) when the session belongs to a different principal...
× a 404 is recoverable: the client can re-initialize and the new session works
  AssertionError: expected 403 to be 404
Tests  4 failed | 12 passed (16)
```

Exactly the 4 behaviour tests went red; the 3 guard tests (403 / 400 / 503) stayed **green** on the old code — which is the point: the fix demonstrably did not collapse genuine authz or outage cases into 404.

A second control was run for the new `redis.get`-throws test, because review flagged that catch block as never previously exercised: mutating it to return 404 makes the new test fail (`expected 404 to be 503`), so it genuinely reaches the catch rather than an earlier branch.

**One honest correction to an earlier version of this description:** the body deep-equal in the mismatch test would *also* have passed against the old code (both cases shared a single branch there, so their bodies were trivially equal). It is a forward regression guard, not proof of the fix. The status assertions and the "does not say mismatch" assertion are what discriminate. The test comment now says so.

## Verification

- All 15 `mcpServer*.test.ts` files + the two Sentry suites → **179 passed**. Two files (`creatorAuthz`, `partnerCreatorPermsNull`) time out at 5s under a 17-file parallel run on a loaded host and pass in isolation (10/10) — pre-existing parallel flake, unrelated to this change.
- `tsc --noEmit` (`@breeze/api`) → **exit 0, clean**. Note: it OOMs on a loaded host at the default heap; it passes under `NODE_OPTIONS=--max-old-space-size=12288`. That is the known pre-existing `apps/api` tsc memory issue, not something this PR introduces.
- `eslint` on all four changed files → clean
- CI on the first commit was green including **CI Success**, **Test API**, **Type Check** and all four **Integration Tests** shards.
- No migrations, no schema, no tenant tables touched — no RLS/cascade/export-policy registration applies.

> **CI note:** `Check Migrations` and `Trivy Image Scan` are red repo-wide right now — they fail on `main` and on unrelated PRs. #4227 tracks the first and PR #4236 fixes it. Please don't attribute those two to this change. (`Check Migrations (non-superuser)` passes here.)

Closes #3744


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, parallel). type-design-analyzer skipped: no new types.

**Findings:** 7 raised → all 7 addressed in `cc445a2`; 0 outstanding.

- **code-reviewer: 0 findings.** Traced control flow by hand and independently confirmed the fix does not turn any genuine authz or availability failure into a 404 — scope-403 still precedes session validation, header-400 and store-503 intact, no other call site or doc assumes the old 403.
- **silent-failure-hunter (MEDIUM/HIGH):** 404 is the spec's self-heal signal, which is exactly what makes a mass session-store loss invisible — a Redis flush/eviction/empty-replica failover does not throw, so it never reaches 503 and now presents as routine expiry that every client silently recovers from. Under the old 403 the same event was loud and got escalated. The blind spot predates this PR; the *consequence* is what changed, so it was ours to close → added a throttled rate tripwire (one Sentry event per 60s window past `MCP_UNKNOWN_SESSION_ALERT_THRESHOLD`, default 50).
- **silent-failure-hunter (MEDIUM):** the MED-1 mismatch signal was `console.warn`-only, and nothing in this app forwards console output to Sentry → now also `captureMessage` with a registered event code. Both new codes registered in `sentryEventCodes.ts`.
- **comment-analyzer + silent-failure-hunter (both, independently):** the `redis.setex` catch comment claimed a failed write means "404 and the client re-initializes" — true only for a one-off write failure on a healthy connection; a real outage sends the next call to 503. Comment now states both outcomes.
- **pr-test-analyzer (6/10):** the `redis.get`-rejecting → 503 catch was dead code as far as the suite was concerned → covered, and **mutation-checked**: flipping that block to 404 makes the new test fail (`expected 404 to be 503`), so it genuinely reaches the catch.
- **pr-test-analyzer (4/10):** a notification (no JSON-RPC `id`) on an expired session was untested → covered, locking the validate-before-notification-dispatch ordering so an invalid-session notification can never be silently 202'd.
- **pr-test-analyzer (2/10):** no assertion that a rejection omits the `Mcp-Session-Id` header → added; the no-oracle property extends to headers.

**Correction carried into the PR body:** the body deep-equal in the mismatch test would *also* have passed against the old code (both cases shared one branch there), so it is a forward regression guard, not proof of the fix. The status and "does not say mismatch" assertions are what discriminate. Flagged by pr-test-analyzer and now stated in the test itself — recorded here rather than quietly dropped.

**Red-first controls (2):** the 4 behaviour tests fail `expected 403 to be 404` against the unfixed implementation while the 3 guard tests (403/400/503) stay green; separately, the new Redis-throw test fails when its catch block is mutated. Both were run, not assumed.

**Tests:** all 15 `mcpServer*.test.ts` + both Sentry suites → 179 passed. `tsc --noEmit` (`@breeze/api`) exit 0, clean (needs a raised heap on a loaded host — known pre-existing `apps/api` tsc OOM). `eslint` clean on all four changed files. CI on the head commit `9d7dbac`: **53 checks passing**, including **CI Success**, **Test API**, **Type Check**, **Lint** and all four **Integration Tests** shards.

**Not attributable to this PR:** `Check Migrations` and `Trivy Image Scan` are red repo-wide (they fail on `main` and unrelated PRs; #4227 tracks the first, #4236 fixes it). `Check Migrations (non-superuser)` passes here.

**Status:** review-clean, awaiting maintainer merge. One judgment call worth a maintainer's eye before merging — noted in the PR body: the issue explicitly left the split-vs-uniform response to us, and this PR chose **uniform 404** (no session-existence oracle) over 404/403. Flipping it back to a split is a one-line change if you'd rather keep the distinct signal for the caller.

PR #4257, opened 2026-08-30
fix(docker): patch CVE-2026-14456 in the images allowed to carry the fix (#4246)

### LINKED ISSUE #4246: [CI][Security] Trivy Image Scan red on main: M365 executor images miss the libcrypto3/libssl3 apk upgrade the other images have (CVE-2026-14456)
`Trivy Image Scan` (workflow `security.yml`) has been red on `main` for its last several runs, which means every open PR inherits a red check that is unrelated to its own diff. Six PRs opened tonight (#4238, #4240, #4241, #4242, #4243, #4244) each had to record it as "not attributable" — that is the cost of leaving it: a permanently-red security job stops being a signal and starts masking real findings.

### The finding

From run [33273323133](https://github.com/LanternOps/breeze/actions/runs/33273323133) on `main`:

```
Total: 2 (HIGH: 2, CRITICAL: 0)
│ libcrypto3 │ CVE-2026-14456 │ HIGH │ fixed │ 3.5.7-r0 │ 3.5.8-r0 │ openssl: OpenSSL: Denial of Service via unbounded memory
│ libssl3    │ CVE-2026-14456 │ HIGH │ fixed │ 3.5.7-r0 │ 3.5.8-r0 │
```

A fixed version exists upstream (`3.5.8-r0`), so this is a base-image refresh, not a wait-on-vendor.

### Root cause

`docker/Dockerfile.api`, `docker/Dockerfile.web` and `apps/portal/Dockerfile` already carry the mitigation:

```dockerfile
RUN apk upgrade --no-cache libcrypto3 libssl3 && ...
```

The three M365 executor images do **not**, and they are what the scan trips on:

- `apps/m365-graph-read-executor/Dockerfile`
- `apps/m365-graph-actions-executor/Dockerfile`
- `apps/m365-communications-executor/Dockerfile`

So the fix already exists in this repo; it just was never applied to the executors when they were added. (Verified by grepping every tracked Dockerfile for `apk upgrade`.)

### Also worth a look while there

`apps/api/Dockerfile` and `apps/web/Dockerfile` likewise lack the line. They are not what the security workflow builds (it uses `docker/Dockerfile.api` / `docker/Dockerfile.web`), so they are not causing this failure — but if they are live build paths anywhere else they carry the same exposure, and if they are dead they should go.

### Suggested fix

1. Add the `apk upgrade --no-cache libcrypto3 libssl3` step to the three executor Dockerfiles, matching the existing images.
2. Consider a guard so a new service image cannot ship without it — the same shape as `composeBindMounts.test.ts`: a test that parses every tracked Dockerfile and asserts the upgrade step is present. This is the third class of "new app/package ships without the wiring the others have" in this repo, and a mechanical check is what has actually caught those.

🤖 Filed by Claude Code while working an unrelated batch of issues; CVE, affected images and the missing-directive grep all verified against `main`.

ToddHebebrand:
## Summary

- patches CVE-2026-14456 in every digest-pinned shipping Node/Alpine image, including the two customer-Graph credential-boundary executors
- preserves the credential boundary by allowing only `apk upgrade --no-cache libcrypto3 libssl3`; `apk add`, general upgrades, extra/reordered packages, alternate flags, and continuation-line bypasses remain rejected
- removes the temporary OpenSSL-coverage exemptions so all in-scope images are governed uniformly
- adds fixture-driven tests against the real shell policy

Closes #4246.

## Owner-approved security tradeoff

The repository owner approved the narrow exception on 2026-08-30. The two credential-boundary builds become time-dependent only for the two already-installed OpenSSL packages. The base digest remains pinned; no dependency, credential, workflow-permission, deployment, or runtime-identity boundary changes.

## Exact state

- Security implementation commit: `d31a936a02c4ca96a39d9434098d9e17ffb4b85c`
- Combined stacked HEAD: `355f49bd8f564feaed01dd2d1e63d76145be02ed`
- Base: `fix/4227-release-lineage` (PR #4236)
- This stack deliberately combines the migration-lineage and Trivy repairs so neither required check needs an administrative bypass. After this PR is green, it can merge into #4236; #4236 then becomes the single green mainline integration PR.

## TDD and verification

RED first: the OpenSSL coverage suite failed four cases for the two missing runtime upgrades and missing hardening-policy calls. A real-policy fixture then failed because `--check-apk` did not yet reject an `apk add` fixture.

Fresh local evidence:

- OpenSSL/hardening contract: 16/16 passed
- API config suite: 497/497 passed
- release-lineage suite: 53/53 passed
- workflow-security suite: 113/113 passed on the updated #4236 base
- automatic migration guard selected `v0.108.0`, excluded the three registered v0.109.0 candidates, and passed
- supply-chain hardening guard, ESLint, Bash syntax, TypeScript API typecheck, and `git diff --check` passed
- both credential executor images built successfully
- both runtime images contain `libcrypto3-3.5.8-r0` and `libssl3-3.5.8-r0`
- local Trivy HIGH/CRITICAL scans: 0 findings for each image

The policy fixture rejects `apk add`, bare/general upgrade, package reordering, extra packages, alternate flags, `/sbin/apk`, and continuation-line invocation. Comments remain allowed.

## Safety boundaries

- no image was published
- no deployment or rollout occurred
- no production, hosted, or customer-device state was mutated
- no migration, tag, or release was changed


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer (parallel).

**Findings:** 3 raised → all addressed in `fd8aa2596`; 0 outstanding.
- *pr-test-analyzer (real):* `UPGRADE_RE` required `libcrypto3` **before** `libssl3` on one line, so the functionally identical `apk upgrade --no-cache libssl3 libcrypto3` read as unpatched. Failed safe, but would have misled the next author. Replaced with an order-independent `upgradesOpenssl()` that still rejects a single package or an `apk add` mention.
- *self-found, pre-empted:* `FROM --platform=$BUILDPLATFORM …` parsed the flag as the base ref, silently dropping a multi-arch image out of scope — the one failure mode this guard can't afford, since out-of-scope looks identical to passing. Fixed + regression test.
- *pr-test-analyzer (minor):* documented the two assumptions the guard rests on (last stage == shipped stage; the hardening coupling proves rule *text*, not that the script is still wired into CI).
- *code-reviewer:* 0 findings at confidence ≥80. *comment-analyzer:* 0 inaccuracies; confirmed the two untouched executors' original comments are still true for them.

**Tests:** `vitest run src/config/` — **494 passed** (14 files). `tsc --noEmit` exit 0 · `eslint` exit 0 · `check-supply-chain-hardening.sh` passes.

**Evidence the fix actually works** (verified, not inferred):
- Built `apps/m365-graph-actions-executor/Dockerfile` locally — build exit 0, and the **final shipped image** contains `libcrypto3-3.5.8-r0` / `libssl3-3.5.8-r0`.
- On the repo's exact pinned digest `d32cdf61…`: before `3.5.7-r0` → after `3.5.8-r0`.
- Guard is **red-first and still discriminating**: reverting the three patched Dockerfiles fails on exactly those three. The `HARDENING_BLOCKED` coupling was mutation-tested — deleting the hardening rule turns it red.

**Trivy Image Scan:** still **fail** — and the log confirms it is failing for the expected reason only. The single failing step is `Scan M365 Graph Read Executor image` (`Total: 2 (HIGH: 2, CRITICAL: 0)`), one of the two images deliberately left unpatched. The API, Web and Portal scans pass. Note the job aborts at that step, so the Graph Actions and Communications scans never execute — which is why I verified the Actions fix with a local build instead.

**This PR cannot turn Trivy green on its own.** It reduces the failing images from 5 to 2; the last 2 need the owner decision below.

**Awaiting owner decision:** patching `m365-graph-read-executor` and `m365-communications-executor` requires relaxing `check-supply-chain-hardening.sh:153,327`, which rejects any `apk upgrade`/`apk add` in those credential-boundary images. I did not do that unilaterally — it trades build reproducibility for CVE exposure on the customer credential boundary. Recommended: narrow the rule to keep rejecting `apk add` and general `apk upgrade` while allowing the exact audited `apk upgrade --no-cache libcrypto3 libssl3`. ~3 lines plus the two Dockerfiles; happy to do it as a follow-up on your word.

**Status:** review-clean, awaiting maintainer merge. `Check Migrations` is red repo-wide (#4227, fixed by #4236) and is not attributable to this diff — no migrations touched.

Issue #4251, opened 2026-08-30
Mobile time entry is unusable for the default Partner Technician role (time_entries:write never granted)


ToddHebebrand:
Recording two blockers found by the W02 permission gate of #3206 (mobile ticketing / time entry), so they survive the merge of PR #4234 (the W02 findings doc). Every claim below is labelled **verified** / **inferred** / **not-checked**.

---

## 1. Permission gap: default Partner Technician can read but never write time entries

**Verified** — `packages/shared/src/constants/permissions.ts:80-81` defines both perms:

```ts
TIME_ENTRIES_READ:  { resource: 'time_entries', action: 'read' },
TIME_ENTRIES_WRITE: { resource: 'time_entries', action: 'write' },
```

**Verified** — the *only* grant path is `apps/api/migrations/2026-06-12-a-ticketing-time-parts.sql:102-134`, which inserts the two permission rows and then propagates them off the matching `tickets` perm: `time_entries:read` to every role holding `tickets:read`, and `time_entries:write` to every role holding `tickets:write`.

**Verified** — the seeded `'Partner Technician'` role (`apps/api/src/db/seed.ts:231-246`) lists `tickets:read` and **not** `tickets:write`. So on a database seeded before that migration, Technician gets `time_entries:read` and never `time_entries:write`: reads the timesheet, 403s on start/stop.

**Verified** — `'Partner Admin'` is unaffected only because it carries a `*:*` wildcard, not because anything grants it the perm explicitly.

**Verified** — roles seeded *after* that migration get **neither** perm: `seed.ts:98-106` deliberately omits `time_entries` from `DEFAULT_PERMISSIONS` (the comment states the registry may define permissions "no system role grants yet (e.g. `time_entries:*`, `automations:*`)"). A fresh self-hosted install is therefore worse off than an upgraded one — the backfill migration has no rows to propagate from at seed time.

**Consequence (inferred, but directly from the above):** #3206 waves W03-W05 ship a mobile timer that a default Partner Technician — the exact persona the feature targets — receives a 403 from on start/stop. W03 (PR #4235) already handles this client-side, but the underlying grant is still missing.

### Contradiction to resolve before choosing a fix

**Verified** — `apps/api/migrations/2026-06-19-billing-roles.sql:5-6` asserts the opposite of the seed, in its own header comment:

> "the feature backfill migrations granted them because Technician holds `tickets:write`"

[...]

PR #4259, opened 2026-08-30
feat(mobile): Timer bar, ticket-detail start/stop, and weekly timesheet (#3206 W05)

### LINKED ISSUE #3899: W05: Timer bar, ticket-detail start/stop, and weekly timesheet
Wave W05 of #3206 — Timer bar, ticket-detail start/stop, and weekly timesheet
Branch: `feature/3206-mobile-ticketing-time-entry/wave-3899`

ToddHebebrand:
## What

Wave W05 of the mobile time-entry feature (#3206): the technician-facing surfaces on top of the client, queue and state slice landed in W03/W04.

- **Ticket detail** gains a TIME section with Start/Stop, driven by a new pure `timerActions.ts` (`startForTicket`, `stopRunningTimer`).
- **TimerBar** rides above the tab bar — `HH:MM:SS` on a 1s interval, the ticket reference, a pending-writes badge, and a Stop action. It is also the app's single queue-replay owner: a false→true `useNetworkConnected` transition drains the offline queue from whatever tab the technician is on.
- **Time tab / weekly timesheet** — `GET /time-entries/timesheet?weekStart=YYYY-MM-DD` grouped by day with daily and weekly totals, inline correction via `PATCH /time-entries/:id`.

## Why these particular shapes

- **`dropped.length > 0` raises a toast naming the count.** A silently discarded time entry is unbillable work; it must never vanish quietly.
- **`weekStart` is date-only, derived from local calendar fields.** Both obvious implementations are wrong: `new Date('2026-08-24')` parses as UTC midnight (the previous day west of Greenwich), and `toISOString()` shifts the day boundary by the phone's offset. Either silently moves entries between weeks. `timesheetWeek.ts` exists to make that testable, and its weekday/month names are fixed strings because Hermes ships a reduced ICU.
- **Approved / billed rows render read-only** with the reason on them, rather than accepting an edit the server answers with `APPROVED_IMMUTABLE` / `ENTRY_BILLED`.
- **403 is never a crash or an empty timesheet.** New `timeEntryAccess.ts` classifies the two different 403s the route emits, because the W02 gate (PR #4234) found both are reachable in production:
  - `requireScope('partner','system')` → `"Insufficient permissions"` → the account is **organization-scoped**; `time_entries` has no org-axis RLS policy, so this can never be retried. Copy says so.
  - `requirePermission` → `"Permission denied"` / `"No permissions found"` → the **role** lacks `time_entries:write`. The default seeded *Partner Technician* role is exactly this case. Copy says an administrator can grant it.
  - Plus `NOT_OWN_ENTRY` (ownership) and `ADMIN_REQUIRED`. An unclassifiable 403 still quotes the server's own message — never a bare "An error occurred".
  A recorded denial is sticky for the session and withdraws the Start/Stop control, so a technician is not offered a button that always fails.
- **`timeSlice.ts` gains a Redux binding** over its existing pure reducers and is registered in the store, so `withLogoutReset` wipes a running timer on sign-out instead of leaking it into the next account's session. The pure functions and their W04 tests are untouched.
- **Transport failures queue; verdicts do not.** NetInfo reports "connected" on a captive portal, so a start/stop that fails with no HTTP status is enqueued rather than lost. A 403/409 is not — a queued 409 would be dropped by the queue's `PERMANENT_STATUSES` rule anyway, and a queued 403 would wedge the head of the queue.
- **`timeEntryReplay.ts`** maps a queued write back to its endpoint. An unknown `kind` or a `create` missing its time bounds throws with status 400 *on purpose*: `drain` treats that as a verdict and drops the row, where a bare throw is retried forever and wedges every entry behind it.

## Tasks covered

- Plan Task 5 — `screens/tickets/timerActions.ts` (+ test) and start/stop wired into `TicketDetailScreen.tsx`.
- Plan Task 6 — `components/TimerBar.tsx` mounted above the tab bar in `navigation/MainNavigator.tsx`, with queue replay on reconnect.
- Plan Task 7 — `screens/time/TimesheetScreen.tsx` and the `TimeTab` route.

Plan: `docs/superpowers/plans/mobile/2026-08-23-mobile-time-entry.md`.

## Tests (verified)

`pnpm --filter breeze-mobile test` — **60 files, 725 passed, 3 skipped**.
`pnpm --filter breeze-mobile typecheck` — **clean**.

New suites, all written red-first and confirmed failing before implementation:

| Suite | Covers |
|---|---|
| `services/timeEntryAccess.test.ts` | 403 classification: scope vs. role vs. ownership vs. unclassifiable; never a generic message; accepts the bare `ApiError` shape |
| `screens/tickets/timerActions.test.ts` | success / offline-queue / `ENTRY_RUNNING` / `NO_RUNNING_TIMER` / denial pass-through; verdicts are not queued; a statusless transport failure is |
| `services/timeEntryReplay.test.ts` | kind→endpoint routing; server error propagated untouched so `drain` can read its status; unknown kind and a bounds-less `create` reject as permanent 400 |
| `lib/timeFormat.test.ts` | `HH:MM:SS` incl. past 24h and negative/NaN clamping; `formatMinutes` renders `null` as an em dash, not `0m` |
| `screens/time/timesheetWeek.test.ts` | Monday-anchored week start; date-only output; local-vs-UTC day boundary; month/year-crossing week steps; deterministic labels |
| `store/timeSlice.test.ts` (extended) | the new Redux binding: adoption of a server-reported timer, sticky denial, queue depth, error raise/clear |

The React Native components (`TimerBar.tsx`, `TimesheetScreen.tsx`, the two modified screens) carry no unit tests, matching this app's convention — `apps/mobile/vitest.config.ts` deliberately includes only `.ts` so component imports never pull the RN/Expo runtime into Vitest. All decision logic was extracted into the pure modules above precisely so it is covered.

## Not checked

- **Plan Task 6 Step 3 — the airplane-mode device walkthrough (airplane mode → start → pending badge → stop → reconnect → both writes land, badge clears) was NOT run.** No device or simulator is available in this environment. The replay path is covered by unit tests on `drain` (W04) and `makeReplaySender` (here), but the end-to-end NetInfo transition on a real handset is unverified.
- Not rendered on a device or simulator at all: layout, the timer bar's fit above the tab bar, and the timesheet's appearance are **not-checked**.
- The 403 states are exercised against synthesised errors, not a live organization-scoped login or a real default *Partner Technician* account — the classification input strings are taken from `apps/api/src/middleware/auth.ts` (verified by reading), not from a live response.

## CI

This PR is based on `feature/3206-mobile-ticketing-time-entry/wave-3898`, not `main`, so `ci.yml` (`pull_request: branches: [main]`) does not fire for it. CI has been dispatched manually against this branch.

Separately: **"Check Migrations" and "Trivy Image Scan" are already red on `origin/main` itself** (355b37465, step "Shipped-migration immutability guard"). Those two failures are pre-existing and are not caused by this work. This PR contains no migration and no Dockerfile change.

Closes #3899

### DIFF
--- a/apps/mobile/src/components/TabIcons.tsx
+++ b/apps/mobile/src/components/TabIcons.tsx
@@ -74,3 +74,22 @@ export function SettingsIcon({ color, size }: Props) {
   );
 }
 
+
+// Clock face for the Time tab: a ring with hour and minute hands. Deliberately
+// not a stopwatch — the tab is the weekly timesheet, and the running-timer
+// affordance is the TimerBar sitting directly above it.
+export function TimeIcon({ color, size }: Props) {
+  return (
+    <Svg width={size} height={size} viewBox="0 0 24 24">
+      <Circle cx={12} cy={12} r={8.5} stroke={color} strokeWidth={1.75} fill="none" />
+      <Path
+        d="M12 7 L12 12 L15.5 14"
+        stroke={color}
+        strokeWidth={1.75}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        fill="none"
+      />
+    </Svg>
+  );
+}
--- a/apps/mobile/src/components/TimerBar.tsx
+++ b/apps/mobile/src/components/TimerBar.tsx
@@ -1,0 +1,532 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { Pressable, StyleSheet, Text, View } from 'react-native';
+
+import { palette, radii, spacing, type } from '../theme';
+import { useAppDispatch, useAppSelector } from '../store';
+import {
+  elapsedSeconds,
+  needsAttentionChanged,
+  pendingWritesChanged,
+  runningTimerAdopted,
+  stoppedTimer,
+  timeAccessDenied,
+} from '../store/timeSlice';
+import {
+  createTimeEntry,
+  getRunningTimer,
+  getTimesheet,
+  stopTimer,
+  updateTimeEntry,
+} from '../services/timeEntries';
+import {
+  drain,
+  enqueue,
+  parkNeedsAttention,
+  readNeedsAttention,
+  readQueue,
+  reconcileQueueOwner,
+  type QueuedWrite,
+} from '../services/timeEntryQueue';
+import { makeReplaySender } from '../services/timeEntryReplay';
+import { serverNowMs } from '../services/serverClock';
+import {
+  clearLocalTimer,
+  readLocalTimer,
+  stampNow,
+  writeLocalTimer,
+} from '../services/localTimer';
+import {
+  makeReconciledSender,
+  needsReconciliation,
+  planReconciliation,
+  type ReconcilePlan,
+} from '../services/timerReconcile';
+import { weekStartFor } from '../screens/time/timesheetWeek';
+import { classifyTimeEntryDenial, isAccountLevelDenial } from '../services/timeEntryAccess';
+import { stopRunningTimer } from '../screens/tickets/timerActions';
+import { stopOutcomeEffects } from '../screens/tickets/timerOutcomeEffects';
+import { isQueueWedged, isTimerBarVisible, shouldReplayNow } from './timerBarLogic';
+import { useNetworkConnected } from '../lib/useNetworkConnected';
+import { formatElapsed } from '../lib/timeFormat';
+import { ticketRef } from '../screens/tickets/ticketCopy';
+import { Toast } from './Toast';
+
+/**
+ * Persistent running-timer affordance, mounted above the tab bar.
+ *
+ * It is also the app's single queue-replay owner: it is the one component
+ * alive for the whole signed-in session, so subscribing here means a
+ * reconnection replays the backlog no matter which tab the technician is on.
+ */
+export function TimerBar({ onOpenTimesheet }: { onOpenTimesheet?: () => void } = {}) {
+  const dispatch = useAppDispatch();
+  const connected = useNetworkConnected();
+  const running = useAppSelector((state) => state.time.running);
+  const pendingCount = useAppSelector((state) => state.time.pendingCount);
+  const denial = useAppSelector((state) => state.time.denial);
+  const tickets = useAppSelector((state) => state.tickets.tickets);
+  const userId = useAppSelector((state) => state.auth.user?.id ?? null);
+  /**
+   * Standing count, not a toast. A toast for work that could not be saved is
+   * gone in four seconds and takes the only notice of unbilled minutes with it.
+   * It lives in the store because the ticket screen can park a row too.
+   */
+  const needsAttentionCount = useAppSelector((state) => state.time.needsAttentionCount);
+
+  const [now, setNow] = useState(() => new Date());
+  const [busy, setBusy] = useState(false);
+  const [toast, setToast] = useState<{ kind: 'success' | 'error'; text: string } | null>(null);
+  /**
+   * The head write keeps failing on a status the queue (correctly) retains —
+   * usually a 403, which issue #4251 makes the ORDINARY state for a default
+   * Partner Technician, not an exotic one. `headAttempts` was added as this
+   * signal and had no consumer, so a wedged queue replayed silently forever.
+   */
+  const [wedged, setWedged] = useState(false);
+  /** A local timer whose start request may also have landed on the server. */
+  const [startUnconfirmed, setStartUnconfirmed] = useState(false);
+
+  const mounted = useRef(true);
+  const stopInFlight = useRef(false);
+  const wasConnected = useRef(connected);
+  // Read by the cold-start effect, which must not re-run on every reconnection.
+  const connectedRef = useRef(connected);
+  /**
+   * No drain may run before the queue's owner has been established. Otherwise a
+   * reconnection landing first would replay the PREVIOUS technician's unsent
+   * writes under this session's bearer token — the leak that keeping the queue
+   * across an involuntary session loss would otherwise open.
+   */
+  const ownerReady = useRef(false);
+
+  useEffect(() => {
+    mounted.current = true;
+    return () => {
+      mounted.current = false;
+    };
+  }, []);
+
+  /** Returns the depth, or null when storage could not be read. */
+  const refreshQueueDepth = useCallback(async (): Promise<number | null> => {
+    try {
+      const queued = await readQueue();
+      if (mounted.current) dispatch(pendingWritesChanged(queued.length));
+      return queued.length;
+    } catch {
+      // QueueStorageError — keep the last known depth. Reporting zero would
+      // tell the technician their unsent minutes are already saved.
+      return null;
+    }
+  }, [dispatch]);
+
+  // 1s tick, only while something is actually running.
+  useEffect(() => {
+    if (running === null) return;
+    setNow(new Date());
+    const interval = setInterval(() => setNow(new Date()), 1000);
+    return () => clearInterval(interval);
+  }, [running]);
+
+  const refreshNeedsAttention = useCallback(async (): Promise<void> => {
+    const rows = await readNeedsAttention().catch(() => null);
+    if (rows !== null && mounted.current) dispatch(needsAttentionChanged(rows.length));
+  }, [dispatch]);
+
+  /**
+   * Resolves what the server actually has before replaying anything that could
+   * duplicate it.
+   *
+   * Only runs when this device is holding something ambiguous — an unconfirmed
+   * local start, or a queued `create` flagged `unconfirmedStart` — so the
+   * ordinary reconnect still costs exactly one drain.
+   */
+  const reconcile = useCallback(async (queue: QueuedWrite[]): Promise<ReconcilePlan> => {
+    const empty: ReconcilePlan = { adopt: null, substitute: null };
+    const localTimer = await readLocalTimer().catch(() => null);
+    if (mounted.current) setStartUnconfirmed(localTimer !== null && !localTimer.startConfirmed);
+    if (!needsReconciliation(localTimer, queue)) return empty;
+
+    let server;
+    try {
+      server = await getRunningTimer();
+    } catch {
+      // Never block a drain on the reconciliation probe: the writes are safe to
+      // send, and the worst case is a duplicate the technician can delete.
+      return empty;
+    }
+
+    const plan = planReconciliation({ localTimer, queue, server });
+    if (plan.adopt !== null && localTimer !== null) {
+      const adopted = {
+        ...localTimer,
+        serverEntryId: plan.adopt.serverEntryId,
+        startedAtWall: plan.adopt.startedAt,
+        startConfirmed: true,
+      };
+      await writeLocalTimer(adopted).catch(() => undefined);
+      if (mounted.current) {
+        setStartUnconfirmed(false);
+        dispatch(
+          runningTimerAdopted({
+            id: plan.adopt.serverEntryId,
+            localId: null,
+            ticketId: adopted.ticketId,
+            startedAt: plan.adopt.startedAt,
+            description: adopted.description,
+          })
+        );
+      }
+    }
+    return plan;
+  }, [dispatch]);
+
+  const replay = useCallback(async () => {
+    let queue: QueuedWrite[];
+    try {
+      queue = await readQueue();
+    } catch {
+      // QueueStorageError: the queue is intact and untouched, so the next
+      // reconnect tries again rather than draining a view we could not read.
+      return;
+    }
+    const plan = await reconcile(queue);
+
+    const base = makeReplaySender({
+      createTimeEntry,
+      updateTimeEntry,
+      serverNow: () => serverNowMs().ms,
+    });
+    const send = makeReconciledSender({
+      base,
+      plan,
+      updateTimeEntry,
+      lookupWeekEntries: async (write) => {
+        const startedAt = write.payload.startedAt;
+        if (typeof startedAt !== 'string') return null;
+        const week = await getTimesheet(weekStartFor(new Date(startedAt))).catch(() => null);
+        return week === null ? null : week.days.flatMap((day) => day.entries);
+      },
+    });
+
+    try {
+      const result = await drain(send);
+      if (!mounted.current) return;
+      dispatch(pendingWritesChanged(result.remaining));
+      dispatch(needsAttentionChanged(result.needsAttentionTotal));
+      // A retained failure that has burned several attempts is not "offline":
+      // nothing behind it can move until somebody acts.
+      setWedged(isQueueWedged(result));
+      if (result.needsAttention.length > 0) {
+        setToast({
+          kind: 'error',
+          text:
+            result.needsAttention.length === 1
+              ? '1 time entry needs attention. Re-enter it from the timesheet.'
+              : `${result.needsAttention.length} time entries need attention. Re-enter them from the timesheet.`,
+        });
+      } else if (result.sent > 0) {
+        // Only writes that actually reached the server are reported as synced —
+        // never ones that merely moved to needs-attention.
+        setToast({ kind: 'success', text: `Synced ${result.sent} offline time ${result.sent === 1 ? 'entry' : 'entries'}` });
+      }
+      // A reconciled or externally started entry can be running without the
+      // store knowing, so the server stays the authority on what is live.
+      try {
+        const timer = await getRunningTimer();
+        if (mounted.current && timer !== null) dispatch(runningTimerAdopted(timer));
+        else if (mounted.current && timer === null) {
+          const stillLocal = await readLocalTimer().catch(() => null);
+          // Do NOT clear a device-only timer just because the server has none:
+          // it is not supposed to be there yet.
+          if (stillLocal === null) dispatch(runningTimerAdopted(null));
+        }
+      } catch {
+        // Leave the local view alone rather than clearing a real timer.
+      }
+    } catch {
+      await refreshQueueDepth();
+      await refreshNeedsAttention();
+    }
+  }, [dispatch, reconcile, refreshQueueDepth, refreshNeedsAttention]);
+
+  useEffect(() => {
+    const previous = wasConnected.current;
+    wasConnected.current = connected;
+    connectedRef.current = connected;
+    if (
+      ownerReady.current &&
+      shouldReplayNow({ coldStart: false, previousConnected: previous, connected, pendingCount })
+    ) {
+      void replay();
+    }
+  }, [connected, pendingCount, replay]);
+
+  // Cold start: the authority on "is a timer running" is the server, not this
+  // process. A timer started from the web dashboard, or before a reinstall,
+  // must appear here rather than being invisible until the next start 409s.
+  //
+  // It is also the only chance a backlog from a PREVIOUS launch gets to drain:
+  // `useNetworkConnected` seeds `true`, so an app relaunched at the office on
+  // strong WiFi never sees the false -> true edge the effect above waits for,
+  // and the unsent entries would sit behind their badge indefinitely. Declared
+  // after that effect so `connectedRef` already holds this render's value.
+  useEffect(() => {
+    let cancelled = false;
+    ownerReady.current = false;
+    void (async () => {
+      // Claim the queue for THIS account before anything can replay it. An
+      // unowned queue is adopted; one left by a different technician is parked
+      // and cleared. This — not the sign-out hook — is what makes a
+      // cross-account replay impossible, because it runs on every path into a
+      // session, including a re-login after a token expiry.
+      if (userId !== null) await reconcileQueueOwner(userId).catch(() => null);
+      ownerReady.current = true;
+      const depth = await refreshQueueDepth();
+      await refreshNeedsAttention();
+      // A device-only timer survives a relaunch, and nothing on the server can
+      // reveal it — without this the bar would be blank until the next tap.
+      const localTimer = await readLocalTimer().catch(() => null);
+      if (localTimer !== null && localTimer.serverEntryId === null && !cancelled && mounted.current) {
+        setStartUnconfirmed(!localTimer.startConfirmed);
+        dispatch(
+          runningTimerAdopted({
+            id: null,
+            localId: localTimer.localId,
+            ticketId: localTimer.ticketId,
+            startedAt: localTimer.startedAtWall,
+            description: localTimer.description,
+          })
+        );
+      }
+      try {
+        const timer = await getRunningTimer();
+        // A null answer must not clear a device-only timer: the server is not
+        // supposed to know about one yet.
+        if (!cancelled && mounted.current && (timer !== null || localTimer === null)) {
+          dispatch(runningTimerAdopted(timer));
+        }
+      } catch (error) {
+        const denied = classifyTimeEntryDenial(error);
+        // Only an account-shaped verdict is a wall. Recording a per-row 403
+        // here would withdraw time tracking app-wide until sign-out.
+        if (denied !== null && isAccountLevelDenial(denied) && !cancelled && mounted.current) {
+          // Record it once so every time surface reports the same explicit
+          // reason instead of a blank screen.
+          dispatch(timeAccessDenied(denied));
+        }
+        // Any other failure is transient; the bar simply stays hidden.
+      }
+      if (
+        !cancelled &&
+        mounted.current &&
+        shouldReplayNow({
+          coldStart: true,
+          previousConnected: wasConnected.current,
+          connected: connectedRef.current,
+          pendingCount: depth ?? 0,
+        })
+      ) {
+        await replay();
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, [dispatch, refreshQueueDepth, refreshNeedsAttention, replay, userId]);
+
+  const onStop = useCallback(async () => {
+    if (stopInFlight.current) return;
+    stopInFlight.current = true;
+    setBusy(true);
+    try {
+      const outcome = await stopRunningTimer(
+        { running },
+        {
+          stopTimer,
+          enqueue,
+          readLocalTimer,
+          clearLocalTimer,
+          parkNeedsAttention,
+          isConnected: () => connected,
+          stamp: stampNow,
+        }
+      );
+      if (!mounted.current) return;
+      // One decision table, shared with TicketDetailScreen — see
+      // screens/tickets/timerOutcomeEffects.ts.
+      const effects = stopOutcomeEffects(outcome);
+      if (effects.clearRunning) {
+        dispatch(stoppedTimer());
+        setStartUnconfirmed(false);
+      }
+      if (effects.accountDenial !== null) dispatch(timeAccessDenied(effects.accountDenial));
+      if (effects.refreshQueueDepth) await refreshQueueDepth();
+      if (effects.refreshNeedsAttention) await refreshNeedsAttention();
+      if (mounted.current) setToast(effects.toast);
+    } finally {
+      stopInFlight.current = false;
+      if (mounted.current) setBusy(false);
+    }
+  }, [connected, dispatch, refreshQueueDepth, refreshNeedsAttention, running]);
+
+  // The toast is a child of this bar, not a portal: unmounting on an emptied
+  // queue would swallow the "N offline time entries could not be saved"
+  // warning in exactly the case it exists for.
+  const visible = isTimerBarVisible({
+    hasRunningTimer: running !== null,
+    // A standing "needs attention" count keeps the bar mounted on its own: it
+    // is the only place unbilled work is reported once the toast has gone.
+    pendingCount: pendingCount + needsAttentionCount,
+    hasToast: toast !== null,
+  });
+  if (!visible) return null;
+
+  const ticket = running?.ticketId
+    ? tickets.find((candidate) => candidate.id === running.ticketId)
+    : undefined;
+  const label = running?.ticketId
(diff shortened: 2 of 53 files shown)


ToddHebebrand:

Every finding below was independently verified by three voters against this branch; one candidate was refuted and is listed with its reason. All fixes are on `124ad9be`. Claims are labelled **verified** (read or executed in this worktree), **inferred** (follows from verified code paths but not executed), or **not-checked**.

The app has no React Native test runtime (`vitest.config.ts` includes `src/**/*.test.ts` only; there is no `*.test.tsx` and no testing-library dependency), so each component-level fix was made testable by extracting its decision into a pure module and pinning it red-first there. The rendered components themselves remain untested — **inferred**, not verified, that the wiring is correct.

### Fixed

**1. Offline start + stop replayed as a ~0-minute entry — the whole offline span was lost** (`services/timeEntryReplay.ts`)
Neither timer endpoint accepts a client timestamp (`startTimerSchema` / `stopTimerSchema` have no `startedAt`/`endedAt`), and the service stamps both ends at request time, so `drain` replaying the pair back to back produced `durationMinutes: 0` — with a green "Synced 2 offline time entries" toast. **verified.**
Fix: new `services/timeEntryCollapse.ts` merges a queued start and the stop that closes it (via the `dependsOn` link `enqueue` already makes) into one `create` carrying the real `queuedAt` bounds, which `createTimeEntrySchema` accepts — no backend change. It declines to collapse a backwards or legacy-epoch pair rather than inventing a span, and widens a same-instant pair by one second so the `endedAt > startedAt` refinement cannot turn it into a dropped 400. A lone queued start still replays as a start: the create endpoint needs both bounds, so an unfinished offline timer remains stamped at reconnect time — a known limit, documented in the module. Tests: `timeEntryCollapse.test.ts` (9 cases) plus two drain-level cases in `timeEntryQueue.test.ts`.

**2. A per-row 403 became the sticky account-level denial and disabled time tracking app-wide** (`screens/time/TimesheetScreen.tsx`)
`classifyTimeEntryDenial` returned non-null for *every* 403, and the caller treated any non-null result as an account wall — so one `APPROVED_IMMUTABLE` on one row replaced the whole Time tab, removed Stop from a running timer, and stayed until sign-out. The author's own per-row `APPROVED_IMMUTABLE` copy was dead code behind the early return, which is what confirmed the intent. **verified.**
Fix: new `entry` denial reason for `APPROVED_IMMUTABLE`, and `isAccountLevelDenial` — only `scope` and `permission` are account-shaped and sticky. All four dispatch sites (timesheet edit, timesheet load, timer bar cold start, both timer surfaces) now gate on it; a per-row verdict toasts and re-reads the stale row instead. Tests in `timeEntryAccess.test.ts` and `timerOutcomeEffects.test.ts`.

**3. A queued (offline) stop never cleared `running`** (`components/TimerBar.tsx`, `screens/tickets/TicketDetailScreen.tsx`) — *reported twice, one fix*
The `ok === 'queued'` branch toasted "Saved offline" but left the bar ticking and Stop live, so a second tap enqueued a duplicate that replayed as 404 `NO_RUNNING_TIMER` → dropped → "1 offline time entry could not be saved" for work that saved. **verified.**
Fix: both screens now share one decision table, `screens/tickets/timerOutcomeEffects.ts`, which clears the local timer on a queued stop (safe: the queued write is what makes the server agree) while deliberately *not* inventing a running timer for a queued start. Tests: `timerOutcomeEffects.test.ts`.

**4. Offline `enqueue` was unguarded — the tap was a silent no-op** (`screens/tickets/timerActions.ts`) — *reported twice, one fix*
`await deps.enqueue(...)` on the offline branch sat outside any try, unlike the online transport branch 30 lines below, and every caller uses `try { … } finally { … }` with no catch invoked as `void onStartTimer()`. A `QueueStorageError` therefore reached no handler: the button returned from "Working…" with no toast and nothing queued. **verified.**
Fix: same try/catch as the online branch, returning `{ ok: false, reason: 'unknown' }` with copy that tells the technician to note the time. Tests: two rejecting-enqueue cases in `timerActions.test.ts`.

**5. The replay-result toast could never render** (`components/TimerBar.tsx`)
`visible = running !== null || pendingCount > 0` short-circuits above the `<Toast>`, which is a plain child rather than a portal. After a drain that empties the queue by *dropping* writes, `remaining === 0` — so the warning was swallowed in exactly the case it was written for. **verified.**
Fix: `isTimerBarVisible` in the new `components/timerBarLogic.ts` keeps the bar mounted while a toast is pending. Tests: `timerBarLogic.test.ts`.

**6. Queued writes survived sign-out and replayed under the next account** (`services/timeEntryQueue.ts`, `services/auth.ts`)
`breeze.timeEntryQueue.v1` is a single un-namespaced key, a queued write carries no owner, and the replay sends with whatever token the current session holds. On a shared field phone, technician A's queued start is created under B — and because `startTimer` auto-stops the caller's running timer first, it also silently ends B's own entry. **verified**; the cross-account server-side attribution is **inferred** from the token/identity path (no live API run).
Fix: `clearQueueForSignOut()` joins the `clearAuthData` teardown alongside the approvals cache, which closes the same class of leak. The rows are parked under an `.orphaned` key and reported to Sentry rather than silently deleted — discarding unsent billable minutes quietly would be its own defect. A failed wipe names the key in `SecureWipeError` instead of reporting a clean sign-out. Tests in `timeEntryQueue.test.ts` and `auth.test.ts`.

**7. A backlog from a previous launch never replayed if the app started online** (`components/TimerBar.tsx`)
`useNetworkConnected` seeds `true` and `wasConnected` seeds from it, so an app relaunched at the office on strong WiFi produced no false → true edge; `drain` has exactly one call site. The badge sat there until connectivity happened to drop and return. **verified.**
Fix: `shouldReplayNow` (in `timerBarLogic.ts`) adds an explicit cold-start pass, gated on a non-empty queue so an ordinary launch spends no round trip; the transition rule is otherwise unchanged. Tests: `timerBarLogic.test.ts` (6 cases covering both triggers and the three non-triggers).

**8. `lockReason` over-locked billed rows the server allows editing** (`screens/time/TimesheetScreen.tsx`)
`BILLED_LOCKED_ENTRY_FIELDS` deliberately excludes `description` and the 409 says "only its description can change", yet the UI hid both Edit and the billable toggle for any billed row and told the technician to ask an administrator. `billed && !isApproved` is an ordinary state — invoice assembly has no approval filter. **verified.**
Fix: new `screens/time/entryLock.ts` returns `{ canEditDescription, canToggleBillable, badge, note }`; billed rows keep Edit and lose only the billable toggle, approved rows stay fully read-only (`assertCanMutate` rejects every field). Tests: `entryLock.test.ts`, table-driven over the four billing statuses × approval.

**9. Day grouping assumed server day keys were local dates** (`screens/time/TimesheetScreen.tsx`)
The client derives a local week and sends it date-only; `z.coerce.date()` makes it UTC midnight, so `getTimesheet` windows on `[weekStartUTC, +7d)` and buckets by `toISOString()`. The seven key *strings* coincide, which is why nothing looked broken — but in America/New_York a Sunday 20:00 callout is `00:00Z` Monday: absent from its own week's rows *and* total, reappearing the next week under Monday with a "20:00" label. **verified** (arithmetic reproduced under a pinned TZ).
Fix: new `screens/time/timesheetLocalDays.ts`. The screen now fetches the one overlapping neighbour week (chosen by `getTimezoneOffset` sign; a failed neighbour fetch degrades to the boundary entries, never the whole week), re-files every entry by its own local day, de-duplicates the overlap, and recomputes day and week totals from exactly the rows on screen. The endpoint takes no timezone parameter and this wave adds no API changes, so the correction is client-side. Tests: `timesheetLocalDays.test.ts` with `process.env.TZ` pinned and an explicit non-vacuity assertion that the zone took effect.

### Applied from suggestions

- **`ENTRY_RUNNING` copy and the ticket-detail note misdescribed the server.** `startTimer` auto-stops the caller's previous timer before inserting, so the 409 is only ever a lost insert race. "A timer is already running. Stop it first." sent the technician after a phantom, and "a start here would 409" was simply wrong — a start on ticket B succeeds and silently closes ticket A's entry. Both strings corrected. **verified** against `timeEntryService.ts`.
- **Billed rows read-only** — same as fix 8 above.

### Declined

- **Queued payloads as a union discriminated by `kind`.** Real improvement, but not behaviour-preserving or trivial: it changes `QueuedWrite`, both `enqueue` dep signatures in `timerActions.ts`, the replay sender's runtime re-parsing, and several tests that construct payload bags directly. Worth a follow-up; the risk today is bounded because only `start` and `stop` are ever enqueued (**verified** by grep).
- **`StartOutcome`/`StopOutcome` redesign (`status: 'sent' | 'queued' | 'failed'`, non-optional `denial` on the denied branch).** Both points are correct — the illegal state is representable and the `true | 'queued' | false` discriminant is a trap. Declined here as a cross-file type refactor touching every caller; note the defensive `outcome.reason === 'denied' && outcome.denial` checks it targets are now gone anyway, absorbed into `timerOutcomeEffects.ts`.
- **`TimeEntryError.code` typed as the server's closed union.** Declined as not behaviour-preserving: `asTimeEntryError` builds `code` from an untyped payload, so narrowing it to a union either needs a runtime allowlist (which would silently drop unknown codes) or a cast that buys no real checking.

### Refuted (not applied)

- **"A queued (offline) start leaves the button reading 'Start timer', inviting a duplicate that is later dropped."** The load-bearing step does not happen. `startTimer` is auto-stop-then-insert, and `drainQueue` replays strictly sequentially, so a second queued start is not concurrent with the first and does not 409 — nothing is dropped and the misleading "could not be saved" toast never fires. **verified** in `timeEntryService.ts` (the 409 needs a genuinely concurrent competing start) and in `drainQueue`'s serialised loop. The accurate parts stand — a second offline tap does enqueue a second start, with no dedupe — but the consequence is one extra ~0-minute closed entry on replay, cosmetic and correctable, not lost work. The remaining substance is an affordance preference, out of scope for a confirmed finding.

### Strengths

- The pure-module split matches the mobile app's existing convention exactly (no react-native imports in `timeFormat.ts`, `timesheetWeek.ts`, `timerActions.ts`, `timeEntryAccess.ts`, `timeEntryReplay.ts`), and each of those five modules ships a co-located Vitest suite. The repo has no component-test precedent in `apps/mobile` (no `*.test.tsx`, no testing-library dependency), so extracting the testable logic was the right way to get coverage here — this review extended the same pattern rather than fighting it.
- No use of `runAction` anywhere — the mobile surfaces correctly go through `coreRequest` plus the typed `TimeEntryError` / `TimeEntryDenial` shapes, as the brief requires.
- The W02 gate outcome is honoured concretely: `classifyTimeEntryDenial` distinguishes the org-scope wall (`requireScope`) from a missing `time_entries:write` grant with distinct, non-generic copy, and both `TimesheetScreen` and `TicketDetailScreen` render an explicit "not available for your account" state rather than an empty screen or a generic error.
- `timesheetWeek.ts` gets the timezone reasoning right for the request side: local-calendar arithmetic and a date-only `weekStart`, with `fromDateOnly` avoiding the `new Date('YYYY-MM-DD')` UTC-midnight trap, and fixed English weekday/month names instead of `toLocaleDateString` on Hermes' reduced ICU. (Finding 9 is the response side of that same seam, which the module could not reach alone.)
- `UnreplayableWriteError` carrying status 400 so `drain` drops rather than wedges on an unknown kind is a genuinely good call, and the comment explains why with the right trade-off (one unsendable row vs. the whole backlog behind it).

### Verification

- `pnpm --filter breeze-mobile test` — **65 files, 784 passed, 3 skipped** (**verified**, run locally on `124ad9be`). Every fix was red first: 12 assertions failed across 9 files before the implementation.
- `pnpm --filter breeze-mobile typecheck` — clean (**verified**).
- Six pre-existing `timeEntryQueue` tests used a start/stop pair as a generic "two writes" fixture and had to be re-pointed at non-collapsing writes; the two `dependsOn` drop tests now seed an *uncollapsible* pair (legacy epoch `queuedAt`), which is exactly where that rule still applies. No property those tests pinned was weakened — **verified** by reading each one.
- **Test-infra note:** `vitest.config.ts` now aliases `@react-native-async-storage/async-storage` to an in-memory stub. Its ESM build imports `./createAsyncStorage` with no extension, which Node cannot resolve, so the new `auth.ts → timeEntryQueue` edge broke eight unrelated suites at import time. Suites that assert on storage still declare their own `vi.mock`, which takes precedence.
- CI re-dispatched for this branch (`gh workflow run CI --ref feature/3206-mobile-ticketing-time-entry/wave-3899`). **Note:** "Check Migrations" and "Trivy Image Scan" fail on `origin/main` itself (`355b3746`, step "Shipped-migration immutability guard") — pre-existing and unrelated to this work, which touches no migrations.

ToddHebebrand:
## Round-two fixes

Head: `f7a99e743`. W05 is rebuilt on the new W04 head (`af7576c77`, PR #4252) — the queue narrowing lives there, and nothing in it is re-fixed here.

### The offline-span design changed, and why

The round-one design queued `start` and `stop` and collapsed the pair at replay time. That is a category error, not four separate bugs. **VERIFIED:** `POST /time-entries/start` sets `startedAt = new Date()` and `POST /time-entries/stop` sets `endedAt = new Date()` (`apps/api/src/services/timeEntryService.ts:446-506, 425-441`); neither accepts a client timestamp. Queueing a verb whose meaning is "now" into a structure that replays it later rewrites when the work happened, every time.

Every symptom followed from that:

- **VERIFIED (a):** a timer started offline could not be stopped offline. `startOutcomeEffects` returned `startRunning: null` for a queued start, so `state.time.running` stayed null, `TicketDetailScreen` kept rendering "Start timer" and `TimerBar` rendered no Stop control. No UI path could produce a queued start *and* a queued stop, so `collapseTimerPairs` was dead code and the 40-minute basement job was lost.
- **VERIFIED (b):** `enqueue` paired a stop to a start by scanning backwards — a heuristic, not an identity.
- **VERIFIED (c):** an unpaired queued start replayed through `startTimer`, creating a reconnect-time phantom.
- **VERIFIED (d):** a fast device clock produced a span rejected by `notFarFuture` (400), which is in `PERMANENT_STATUSES`, so the technician's real work was deleted.
- **VERIFIED (T3):** a stop queued at 15:00 and replayed at 18:00 billed the customer three extra hours.

**Advisor quorum (Claude + Codex, agreed): replace with client timestamps.** Repairing the collapse cannot remove the root cause — to make the collapse fire at all it must first give the technician a locally-ticking timer with a real client start time and an offline Stop, and once that exists the queued pair is redundant, because the closed span can be emitted directly as a `create`. A repaired collapse converges on this design with an extra reconstruction layer and a heuristic still in the middle.

**Net invariant:** the queue only ever holds writes carrying their own explicit timestamps, and `drain` cannot call `/start` or `/stop`. Mobile-originated over-billing and phantom timers are now unrepresentable rather than defended against.

**VERIFIED that this needs no backend change:** `PATCH /time-entries/:id` accepts `endedAt` on a still-running entry, validates `endedAt > startedAt` and recomputes `durationMinutes` (`timeEntryService.ts:615-671`); `updateTimeEntrySchema.endedAt` carries no `notFarFuture` refine (`packages/shared/src/validators/timeEntries.ts:29-36`). So a timer the server started is closed at its true tap time by id. That is also *more* robust than the old queued stop on an ambiguous transport failure: a targeted PATCH re-applying the same timestamp is effect-idempotent, where a retried `stop` returns 404 `NO_RUNNING_TIMER`, is dropped, and is reported to the technician as work that could not be saved — when in fact it was.

---

### Defects and the tests that pin them

| # | Defect | Fix | Test |
|---|---|---|---|
| T1 | `clearQueueForSignOut` ran on automatic session invalidation, so an expired token cost a technician a full offline day | `clearAuthData({ deliberate })`, threaded through `beginSessionInvalidation` and `logoutAsync`; **defaults to false** so a new call site cannot destroy unsent work by omission. The three user-facing Sign out buttons opt in; the device-blocked listener and the sync `logout()` reducer do not. What actually makes keeping the queue safe is `reconcileQueueOwner(userId)` at sign-in, which parks a queue left by a *different* technician and adopts an unowned one — TimerBar gates all draining behind it | `auth.test.ts`: "discards … on a DELIBERATE sign-out" / "KEEPS the queue on an involuntary session loss"; `timeEntryQueue.test.ts`: owner adopt / switch / same-user cases |
| T2 | An in-flight drain rewrote the queue after sign-out wiped it | A generation counter, bumped by every deliberate emptying. The drain captures it before its first read and refuses to persist — on both the normal path and the tombstone path — if it changed | "an in-flight drain cannot write the queue back after a sign-out cleared it"; "…does not leave a tombstone behind a sign-out either"; "a drain that started before an owner switch cannot restore the old queue". **Mutation-verified:** removing the fence reddens two of them |
| T5 | Neighbour-week fetch used TODAY's UTC offset, not the displayed week's | `neighbourWeekOffsets(weekStart)` derives it from that week's own local boundaries, and returns a list | London/New-York DST cases with an injectable local-midnight resolver (deterministic, no ambient zone) plus one case tying the resolver to the real calendar. **Mutation-verified:** restoring the today-derived form reddens four |
| T6 | A failed neighbour fetch silently reintroduced the local-day bug | `reportInternalError` + a standing banner naming the consequence ("the total may be low") | not-checked by unit test — the fetch loop is inside the screen; the telemetry call and banner are code-reviewable only |
| T7 | A failed week load after a successful one rendered an empty week with a 0m total | Loaded entries are tagged with the week they were fetched for; `timesheetPhase` decides spinner/error/week from that. A failed *refresh* of a week already on screen keeps its rows and shows a banner | `timesheetLoadState.test.ts`. **Mutation-verified:** dropping the week tag reddens "refuses entries fetched for a DIFFERENT week" |
| T8 | A queue wedged on a retained 403 replayed silently forever; `headAttempts` had no consumer | `isQueueWedged` in `timerBarLogic.ts` (pure, therefore testable) drives a standing row naming the likely cause. Nothing is dropped to unwedge it. Weighted for issue #4251: a default Partner Technician genuinely lacks `time_entries:write`, so a permanent 403 is the **expected** state for many users | `timerBarLogic.test.ts`, including the state that must NOT alarm (one failed attempt is just being offline) |
| T9 | The orphaned park was not a backup | It is now an APPENDED list of batches, each stamped with its owner and carrying the needs-attention rows too. A park that does not land raises `QueueParkError`, the rows are **left in place**, and `clearAuthData` reports a partial wipe | "APPENDS to the orphan park…"; "refuses to delete the queue when the park did not happen". **Mutation-verified:** letting a failed park proceed to the delete reddens it |
| T10 | Guards with no discriminating test | The MAX_SPAN clamp is gone with `timeEntryCollapse`. The two `enqueue` try/catch guards and the three `timeSpan` guards each now have a test that fails when the guard is removed | **Mutation-verified, five separate runs:** removing either enqueue guard, the monotonic comparison, the both-bounds shift, or the 1s floor each reddens exactly one test |
| T11 | `process.env.TZ` in a test file does not apply under the threads pool | Pinned in `vitest.config.ts` before any worker spawns. The existing non-vacuous offset assertion stays and now actually guards something | **VERIFIED both ways:** `--pool=threads --maxWorkers=2` was red on the old code (`expected 360 to be 240`) and is green now; the full suite passes under both pools |

---

### What the offline path now does

- **Start** persists a LOCAL timer (`services/localTimer.ts`) *before* any network call, and nothing is ever enqueued at start time. Offline it is `startConfirmed: true` (no request was made, so no server entry can exist); a transport failure with no status leaves it `startConfirmed: false` and the reconciler resolves the ambiguity against one `GET /time-entries/running` rather than guessing.
- **Stop** branches on **identity**, not connectivity. A device-only timer becomes one `create` with its real bounds, online or offline — `/stop` cannot close an entry the server never created. A server-owned timer is closed with `closeEntry` at the tap time.
- **`services/timeSpan.ts`** holds the money logic, pure: `closeLocalSpan` measures from `performance.now()` anchors so a mid-job wall-clock correction cannot invent billable hours, refuses to invent a span when the clock ran backwards (parked for manual entry, not silently billed), and floors a mis-tap to 1s. `shiftIntoPast` translates **both** bounds, so duration is invariant.
- **`services/serverClock.ts`** anchors an offset from every response `Date` header — three lines in `api.ts`, client-side only, no backend change — so the shift still works while offline.
- **Needs-attention** replaces deletion: a permanently-rejected write is parked and readable, surfaced as a persistent tappable row rather than a four-second toast. "Synced N" now counts only writes that actually reached the server.

**End-to-end proof** (`services/offlineSpan.test.ts`, driving the real modules): start offline at 09:00, stop at 09:40, drain at 12:00 → exactly one `create` with those exact bounds and neither timer verb called. Removing the `ok: 'local'` running timer turns it red, which is what pins claim (a). Same suite covers the 15:00→18:00 PATCH, the fast-clock shift with its 40 minutes exact, a 400 readable from `readNeedsAttention()`, and both reconciliation outcomes.

**Structural guard** (`timeEntryReplay.test.ts`): a Proxy recorder proves `makeReplaySender` touches only `createTimeEntry` / `updateTimeEntry` / `serverNow` for every member of `QUEUED_KINDS`, plus a compile-time assertion that `ReplaySenders` declares neither verb — so reintroducing one is a type error, not a convention somebody re-reads.

---

### Verification

- **VERIFIED:** `pnpm --filter breeze-mobile typecheck` clean on both branches.
- **VERIFIED:** `pnpm --filter breeze-mobile test` — 860 passed, 3 skipped, 70 files (W05); 686 passed, 3 skipped (W04). Green under both the default forks pool and `--pool=threads --maxWorkers=2`.
- CI dispatched per branch (neither PR targets `main`, so `ci.yml` never fires on the PR itself). "Check Migrations" and "Trivy Image Scan" are red on `origin/main` itself and are not chased.

### Risks worth a reviewer's eye

- **not-checked:** no React Native component test runtime exists, so `TimerBar`, `TicketDetailScreen` and `TimesheetScreen` are reviewed, not tested. Every decision they make was pushed into a pure module precisely so it *is* tested; what remains untested is the wiring.
- **inferred:** `reconcileQueueOwner` keys on `state.auth.user.id`. If that id is ever null for a signed-in session, the queue stays unowned and the cross-account guard does not fire — the sign-out path still covers the deliberate case.
- **VERIFIED:** `clearNeedsAttention` and `readOrphanedQueues` are exercised by tests but have no app-code consumer yet — they are the recover-by-hand API for a future timesheet affordance. Worth a reviewer's call on whether that affordance belongs in this wave.
- **inferred:** the duplicate-create guard costs one `getTimesheet` per retried `create`. It is best-effort by design — when the week cannot be read we send, because a duplicate is correctable from the timesheet while a lost entry is simply unbilled.

ToddHebebrand:
## Round-two fixes — offline design replaced, not repaired

A second review round (5 lenses, 3-voter adversarial verification) ran over the code that round one's *fixes* introduced. It confirmed **20 defects**, and the largest finding was that round one's headline fix did not work.

### The design decision (advisor quorum: Claude + Codex, agreed)

`POST /time-entries/start` and `/stop` are **server-stamped verbs**. Queueing one defers its stamping to reconnect time, and every symptom followed from that single category error:

- a 40-minute basement job billed as **0 minutes**
- a 15:00 stop replayed at 18:00 and billed to **18:00** — silently over-billing the customer
- a reconnect-time **phantom timer** nothing closes
- a stop paired to a start by a **backwards scan** heuristic

`repair-collapse` was rejected on the merits: to make the collapse fire at all it must first give the technician a locally-ticking timer with a real client start time and an offline Stop — and once that exists, the queued start/stop pair is redundant, because the closed span can be emitted directly as a `create`. A repaired collapse converges on this design with an extra reconstruction layer and a heuristic still in the middle. `revert-collapse` abandons the plan's Task-3 goal and leaves the over-billing in place.

**Verified the replacement needs no backend change:** `PATCH /time-entries/:id` accepts `endedAt` on a still-running entry, validates `endedAt > startedAt`, and recomputes `durationMinutes` (`timeEntryService.ts:615-671`); `updateTimeEntrySchema.endedAt` carries no `notFarFuture` refine. So a server-started timer can be closed at the **true tap time by id**.

**Net invariant:** the queue only ever holds writes carrying their own explicit timestamps, and `drain` can never call `/start` or `/stop`. Over-billing and mobile-originated phantom timers are now *unrepresentable* rather than defended against. `timeEntryCollapse.ts` is deleted.

### Fixed

| # | Defect | Fix |
|---|---|---|
| T3/T4 | Offline span lost; delayed stop over-bills; phantom timers; heuristic pairing | Local ticking timer (`localTimer.ts`); stop branches on **identity, not connectivity**; `timeSpan.ts` measures from `performance.now()` anchors so a wall-clock correction cannot invent billable hours; `serverClock.ts` anchors an offset from every response `Date` header; `timerReconcile.ts` resolves a lost start against `GET /time-entries/running` — adopting the server identity or closing the phantom, never posting a second entry |
| T1 | `clearQueueForSignOut` ran on 401/403 expiry, unreadable keychain and device block — a full offline day lost on relaunch | `clearAuthData` takes `deliberate`, defaulting to **false** so a new call site cannot destroy unsent work by omission; safety comes from ownership (`reconcileQueueOwner`) rather than lifecycle guessing |
| T2 | An in-flight drain wrote the previous technician's queue back after the wipe | Generation counter; the drain refuses to persist a snapshot taken under a different generation |
| T9 | The orphan park was not a backup — a second sign-out overwrote it, a failed park was swallowed | Appended list of owner-stamped batches; a park that does not land raises `QueueParkError` and the rows are **left in place** |
| T5 | Neighbour-week fetch used *today's* UTC offset, not the displayed week's — boundary entries vanished across DST | `neighbourWeekOffsets(weekStart)` derives from that week's own local boundaries |
| T6 | A failed neighbour fetch silently restored the local-day bug | Reports to `reportInternalError` + standing banner naming the consequence |
| T7 | A failed week load rendered a confident **empty week with a 0m total** | Entries tagged with the week fetched; a failed refresh keeps its rows and shows a banner |
| T8 | A queue wedged on a retained 403 replayed silently forever; `headAttempts` had no consumer | `isQueueWedged` + a standing row naming the likely cause. Matters more than it sounds: per #4251 a default Partner Technician genuinely lacks `time_entries:write`, so a permanent 403 is the **expected** state for many users. Nothing is dropped to unwedge it |
| T10 | Guards with no discriminating test | Tests that redden when the guard is removed |
| T11 | `process.env.TZ` does not apply under `--pool=threads`, so the timezone suite was red on the documented local invocation | Zone pinned in `vitest.config.ts` before any worker spawns; verified green **with** and red **without** |

Needs-attention count lives in `state.time`, kept deliberately separate from `pendingCount`: a pending write will still be sent, a parked one never will, and collapsing them turns the notice back into a badge that looks about to clear itself.

### Verification

`pnpm --filter breeze-mobile test` — **860 passed / 3 skipped, 70 files** (verified). `typecheck` clean (verified). Every fix was written red-first; the load-bearing guards are mutation-verified (removing the drain generation fence reddens both in-flight-drain tests; letting a failed park proceed to the delete reddens the park-failure test; restoring the today-derived week offset reddens four DST cases).

### Not checked

- **No on-device run.** Plan Task 6 Step 3 (airplane mode → start → stop → reconnect) has not been executed on a phone; the new `performance.now()` anchoring and the app-kill path especially want a real device.
- No replay against a live API — unit tests mock the transport per the plan's Global Constraints.
- One residual accepted by the quorum: keeping `POST /start` on the genuinely-online path leaves a narrow window where a **lost response** can strand a server timer during an outage. Drain-time reconciliation closes it; making it literally impossible would require mobile never to call `/start`, which would cost cross-device visibility of a technician on the clock. Documented in the UI rather than designed away.
- `Check Migrations` and `Trivy Image Scan` are red on `origin/main` itself (`355b37465`) — pre-existing and unrelated; this PR touches no migrations.

ToddHebebrand:
## Base refreshed + one knock-on fix from W04's round-three review

**Base.** Merged `main` through the stack (`main` → W03 → W04 → W05). `#4236` on main is what fixes the `Check Migrations` and `Trivy Image Scan` reds these branches carried — they were a stale base, not anything in the PR. Zero conflicts; merge commits, not a force-push, so the stack stays reversible and yours to sequence.

**Fix (`ead4d41c2`).** #4252's round-three review found that `readNeedsAttentionUnlocked` conflated "empty store" with "store I could not read", and both writers persisted over it. Fixing that in W04 (`cf55c29a5`) made the reader return `null` on a storage failure — and `parkAndEmptyQueue` here is the W05 call site that has to branch on it.

Its last act is `setItem(QUEUE_NEEDS_ATTENTION_KEY, '[]')`. Rows it could not read were never copied into the orphan batch, so it would have deleted them outright — the same destruction it already refuses when the orphan park itself fails (*"refuses to delete the queue when the park did not happen"*). It now throws `QueueParkError` and reports `parkError: 'needs-attention-unreadable'`. A storage failure is transient, so the next sign-out or `reconcileQueueOwner` retries the whole park; the cost is the already-documented narrow cross-account window, not lost work.

Honest provenance: **`tsc` surfaced this, not the suite.** Both `parkedAttention.length` and the orphan-batch field type went red on the merge, and the pre-fix behaviour was `Cannot read properties of null (reading 'length')`. The test asserting the intended contract — the sign-out rejects and the parked row survives — was watched to fail with exactly that symptom before the fix. It needed a new mock flag (`failNextNeedsAttentionRead`), because the existing `getItemFailures` counter cannot target the third `getItem` of the sign-out path.

### Verification

`apps/mobile`: **863 passed / 3 skipped, 70 files**; `tsc --noEmit` exit 0 (both verified on `ead4d41c2`).

### Unchanged and still open

- **No CI has ever run on this PR.** Its base is a sibling branch, so `ci.yml`'s `pull_request: branches: [main]` trigger never fires and `gh pr checks` reads green off one Cloudflare Pages check. Same for #4252. Retargeting is part of the merge sequence, which is yours.
- **No on-device run.** Plan Task 6 Step 3 (airplane mode → start → stop → reconnect) has still never been executed on a phone; the `performance.now()` anchoring and the app-kill path especially want one.
- **#4251** — the default Partner Technician role lacks `time_entries:write`, so the timer 403s for exactly the user it was built for.

Issue #4261, opened 2026-08-30
[CI] Four jobs still cannot block a merge: check-migrations, lint-agent, test-agent-race, agent-windows-manifest-guard


ToddHebebrand:
Follow-up from #3941 / PR #4254, which made `test-mobile` blocking.

`main` is protected by ruleset `Protect main branch`, whose `required_status_checks` contains exactly one context: **`CI Success`**. A job therefore blocks a merge only when it appears in all three of:
1. the `needs:` array of `ci-success`,
2. a `<JOB>_RESULT` env var,
3. an assertion in that job's **unconditional** bash if-condition.

Being in `needs:` alone is not enough — that is the vacuous-fix trap PR #4254's contract test now catches.

**Four jobs run on every PR but are in none of the three, so they cannot block a merge:**

- `check-migrations`
- `lint-agent`
- `test-agent-race`
- `agent-windows-manifest-guard`

PR #4254 pinned these in a documented `UNGATED_JOBS` set in `apps/api/src/config/ciSuccessGatingContract.test.ts`, so the gap is visible and any *new* job forces a conscious decision. It deliberately did not promote them.

### Why they were not promoted in #4254

`check-migrations` is **red repo-wide right now** — the immutability guard diffs against tag `v0.109.0-rc.3`, which is not on `main` (it points at `origin/fix/s0-pam-actuation-lifecycle`) and carries eight `2026-09-1x-*.sql` migrations that do not exist on `main`, so they read as deleted. Making it blocking today would block every PR. **#4227 tracks this and PR #4236 fixes it.**

### Suggested sequence

1. Land PR #4236 so `check-migrations` goes green.
2. Confirm each of the four is green on `main` over several consecutive runs.
3. Promote them one at a time, removing each from `UNGATED_JOBS` in the same PR — the contract test then enforces the promotion.

`test-agent-race` deserves particular attention: a race detector that cannot block a merge is a race detector that will eventually be ignored.

🤖 Filed by Claude Code from the #3941 fix; the three-place gating mechanism was verified by extracting and executing the real `ci-success` script against both the pre- and post-change workflow.

Issue #4260, opened 2026-08-30
[CI][Security] Trivy scans docker/Dockerfile.api+web, but release and hosted-images ship apps/api+web/Dockerfile — the deployed images are unscanned


ToddHebebrand:
**The container images that ship to customers are not the images Trivy scans.**

Found while fixing #4246. Verified against `origin/main`:

`security.yml` builds and scans:
```
docker/Dockerfile.api        -t breeze-api:security-scan
docker/Dockerfile.web        -t breeze-web:security-scan
apps/portal/Dockerfile       -t breeze-portal:security-scan
apps/m365-graph-read-executor/Dockerfile
apps/m365-graph-actions-executor/Dockerfile
apps/m365-communications-executor/Dockerfile
```

But the images actually pushed to the registry come from different files for api and web:
- `release.yml:2783` → `file: apps/api/Dockerfile` (`push: true`)
- `release.yml:3135` → `file: apps/web/Dockerfile` (`push: true`)
- `hosted-images.yml:88` → `dockerfile: apps/api/Dockerfile`
- `hosted-images.yml:92` → `dockerfile: apps/web/Dockerfile`

So `apps/api/Dockerfile` and `apps/web/Dockerfile` — the two most widely deployed images in the product — **have never been scanned**. `docker/Dockerfile.api` / `docker/Dockerfile.web` are scanned in their place.

This is not hypothetical: both shipped a known HIGH CVE. They used the same vulnerable `node:24-alpine` digest (libcrypto3/libssl3 3.5.7-r0, CVE-2026-14456) and lacked the `apk upgrade --no-cache libcrypto3 libssl3` step that the scanned `docker/Dockerfile.*` variants carry. PR #4257 patches them, but the *coverage gap* is the durable defect: any future base-image CVE in the shipped api/web images will again be invisible.

Portal and the three M365 executors are fine — same file scanned as shipped.

### Suggested fix

1. Point the Trivy image scan at the Dockerfiles that are actually published (`apps/api/Dockerfile`, `apps/web/Dockerfile`), or scan both sets.
2. Better: derive the scan matrix from the same source of truth the release/hosted-images workflows use, so the two cannot drift again. A hand-maintained parallel list is what failed here — the same failure mode as the hand-maintained deploy service list that left `portal` five releases stale.
3. Consider whether `docker/Dockerfile.api` and `docker/Dockerfile.web` should still exist as separate files at all, or whether the duplication is itself the bug.

[...]

PR #4263, opened 2026-08-30
fix(security): narrow the executor apk rule to the audited OpenSSL upgrade (#4246)

ToddHebebrand:
## Summary

Owner-approved follow-up to #4257. Todd chose option A: **narrow** the credential-boundary apk rule rather than leave two images shipping a known HIGH CVE.

**Stacked on #4257** (base is `fix/4246-m365-executor-openssl-upgrade`, not `main`) — it edits the guard test that PR introduces, so it has to land after it. Merge #4257 first.

## The problem

`apps/m365-graph-read-executor` and `apps/m365-communications-executor` rejected **every** `apk` invocation:

> `executor runtime must not resolve mutable Alpine packages during the image build`

That made CVE-2026-14456 unfixable in them. Verified with docker in #4257: the current upstream `node:24-alpine` tag *still* ships the vulnerable `3.5.7-r0`, so refreshing the pinned digest fixes nothing. `apk upgrade` is the only mitigation that exists today.

## What changed

The control is narrowed, not dropped. Exactly one literal form is permitted:

```
apk upgrade --no-cache libcrypto3 libssl3
```

Everything else — `apk add`, a bare `apk upgrade`, upgrading a different package, `apk del` — stays rejected.

**The allowance is subtraction over a character-literal form**: enumerate every apk invocation, drop the single permitted line, fail on whatever remains. This is the part that matters. An over-broad "allow `apk upgrade`" pattern would silently reopen the whole policy while the check still reported green — a worse outcome than the CVE. So the flags, the package names, and their order are all pinned, and the rule is fail-closed against any form it has not seen.

Narrowing was also the moment to close a hole in the **old** rule: it matched only `^RUN .*apk`, so an apk on a `\`-continuation line — or behind an absolute path like `/sbin/apk` — walked straight through it. Both are now caught.

## Accepted tradeoff

These two builds become **time-dependent** rather than fully reproducible from the pinned base digest, bounded to those two named packages. That is the property the original rule bought, and it is the one being traded. Everything else the rule protected (no new packages, no capability expansion) is preserved.

## Proving it still REJECTS

A rule change that only proves the happy path proves nothing, so:

- **`--self-test`** — a 14-case accept/reject matrix run through the real function.
- **`--check-apk <file>`** — applies the rule to one arbitrary file, so `executorApkPolicy.test.ts` drives it against its **own** fixtures rather than trusting the script's case list. Deleting cases from the script cannot quietly reduce coverage.
- **Mutation-tested four ways** (all confirmed to fail loudly):

| Mutation | Result |
|---|---|
| `AUDITED_APK_UPGRADE` → `apk upgrade` | 5 self-test cases fail |
| `ANY_APK_INVOCATION` → `^RUN .*apk` | continuation-line case fails |
| plant `apk add --no-cache curl` in read-executor | real script fails, names the line |
| plant bare `apk upgrade` in comms-executor | real script fails, names the line |

Loosening the rule in the vitest layer turns 5 of 8 tests red — the suite is not vacuous.

## Also updated from #4257

`dockerfileOpensslUpgrade.test.ts`: both executors are now **covered** rather than exempt, so `HARDENING_BLOCKED` and its self-invalidating coupling are gone. That suite would stay green if this policy were loosened, so its docstring now points at `executorApkPolicy.test.ts` as the separate proof — the two are not interchangeable.

## Flagged, deliberately not fixed here

`apps/m365-graph-actions-executor` has **no hardening block at all**, despite handling Graph **write** actions — arguably higher risk than the read boundary. The script itself acknowledges this in a comment ("The actions executor never got one — an acknowledged inconsistency"). Out of scope for a policy-narrowing PR; worth its own issue.

## Verification

- `check-supply-chain-hardening.sh` full run: exit 0 · `--self-test`: 14/14
- `vitest run src/config/` — **500 passed** (15 files)
- `tsc --noEmit` exit 0 · `eslint` exit 0 · `shellcheck -S error` clean · `bash -n` OK
- Both upgrade steps verified to land in the **final `runner` stage** mechanically (computed each step's owning stage; confirmed nothing `COPY --from`s it)

**CI caveat:** `security.yml` triggers only on `pull_request: branches: [main]` and has no `workflow_dispatch`, so a stacked PR gets **no Trivy run at all** — `gh pr checks` reading green here means little. I am verifying Trivy by building and scanning all six images locally with the workflow's exact settings (`--severity HIGH,CRITICAL --exit-code 1`) and will report the result in a follow-up comment, stating which images were proved by CI versus locally.

Context from this work: #4260 (Trivy scans `docker/Dockerfile.api|web` but release ships `apps/api|web/Dockerfile`), #4261 (four CI jobs cannot block a merge).

Refs #4246

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -59,9 +59,11 @@ RUN pnpm build
 FROM node:24-alpine@sha256:d32cdf619f63fe0471182d08996dd516c6275bb5fd31ae06e55a570bd9e1ad43 AS runner
 WORKDIR /app
 
-# Install wget for healthcheck; remove the bundled npm (and its vendored undici,
-# CVE-2026-12151) since the runtime entrypoint is `node`, not npm.
-RUN apk add --no-cache wget && \
+# Refresh libcrypto3/libssl3 because the pinned base digest predates the fix for
+# CVE-2026-14456; install wget for healthcheck; remove the bundled npm (and its
+# vendored undici, CVE-2026-12151) since the runtime entrypoint is `node`, not npm.
+RUN apk upgrade --no-cache libcrypto3 libssl3 && \
+    apk add --no-cache wget && \
     rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx
 
 # Set production environment
--- a/apps/api/src/config/dockerfileOpensslUpgrade.test.ts
+++ b/apps/api/src/config/dockerfileOpensslUpgrade.test.ts
@@ -51,13 +51,48 @@ import { describe, expect, it } from 'vitest';
  * new image appears, or a base ref changes shape, the snapshot fails and forces
  * a human to decide which side of the line it belongs on, rather than letting an
  * image drop silently out of coverage.
+ *
+ * Two assumptions this guard rests on, recorded so they are not rediscovered
+ * the hard way:
+ *
+ *  - **"Last stage in the file" == "the stage that ships."** True for every
+ *    Dockerfile here today, but a `docker build --target <stage>` in a workflow
+ *    would ship an earlier stage and quietly defeat the whole check. Nothing in
+ *    the repo does that at present.
+ *  - The `HARDENING_BLOCKED` coupling below proves the hardening *rule text*
+ *    still exists; it cannot prove the hardening *script* is still wired into
+ *    CI. If that script were dropped from its workflow the control would stop
+ *    running while this suite stayed green.
  */
 
 // apps/api/src/config -> repo root is 4 levels up.
 const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../../..');
 
-/** The exact mitigation the shipped images already use. */
-const UPGRADE_RE = /apk\s+upgrade\b[^\n]*?\blibcrypto3\b[^\n]*?\blibssl3\b/;
+/** The packages CVE-2026-14456 is fixed in. */
+const REQUIRED_PACKAGES = ['libcrypto3', 'libssl3'] as const;
+
+/**
+ * True when `bodies` collectively `apk upgrade` both OpenSSL packages.
+ *
+ * Deliberately order-independent, and tolerant of the two being split across
+ * separate `apk upgrade` commands: `apk` does not care about argument order, so
+ * a guard that did would reject a Dockerfile that is genuinely patched and send
+ * the next author hunting for a bug that is not there. What it still requires is
+ * that each package be named on an actual `apk upgrade` line — a bare mention in
+ * a comment or an `apk add` does not count.
+ */
+function upgradesOpenssl(bodies: string[]): boolean {
+  const upgraded = new Set<string>();
+  for (const body of bodies) {
+    for (const line of body.split('\n')) {
+      if (!/\bapk\s+upgrade\b/.test(line)) continue;
+      for (const pkg of REQUIRED_PACKAGES) {
+        if (new RegExp(`\\b${pkg}\\b`).test(line)) upgraded.add(pkg);
+      }
+    }
+  }
+  return REQUIRED_PACKAGES.every((pkg) => upgraded.has(pkg));
+}
 
 /**
  * A digest-pinned node Alpine base, e.g.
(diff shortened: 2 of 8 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer (security-focused, adversarial) + pr-test-analyzer, in parallel.

**Findings:** 4 raised → all addressed; 0 outstanding.

1. **Mid-line `#` bypass (real, security-relevant).** Found by my own adversarial probing and independently confirmed by the security review against `db16eee8e`: `reject_unaudited_apk` stripped comments with `sed 's/#.*$//'`, so `RUN echo "#" && apk add curl` collapsed to `RUN echo "` and an arbitrary install **passed the control** (exit 0, "is clean"). The premise was wrong as well as unsafe — in a Dockerfile `#` only opens a comment at line start. Fixed in `012da19a4` (`grep -v '^[[:space:]]*#'`), with regression cases in both proof layers. Reviewer re-ran all four bypass fixtures against the fix: all now correctly rejected.
2. **Two tests silently lost (real).** My `HARDENING_BLOCKED` removal deleted a contiguous range, and two *unrelated* tests sat inside it — the only coverage of `upgradesOpenssl()` order-independence and of `--platform` base-ref parsing, neither covered elsewhere. Restored verbatim in `f304a2da0`.
3. **Missing `--no-cache` case untested.** `apk upgrade libcrypto3 libssl3` was correctly rejected but unproven — the likeliest fat-finger of the audited line. Added to both layers.
4. **Literal-token scan limitation.** A command that never spells `apk` (base64 into `sh`, shell-variable assembly) is invisible to the rule. Reviewer confirmed this is **not a regression** — the previous rule required the same literal token — and is inherent to grep-based Dockerfile scanning. Documented in the script so the check isn't mistaken for a defence against a hostile author: it guards against drift, not evasion.

**What the adversarial review could NOT break** (each probed with real fixtures and observed exit codes): chaining after the audited form (`&& apk add`, `; apk add`), `busybox apk`, `/sbin/apk`, `sh -c "apk add"`, `$(which apk)`, JSON exec-form `RUN ["apk","add"]`, BuildKit heredocs, version-pinned `libssl3=3.5.7-r0`, CRLF endings, and separate-RUN placement. All correctly rejected. The self-test harness was verified sound (`fail`'s `exit 1` inside `( )` is correctly observed by the `if`).

**Non-vacuity, verified by execution:** stubbing `reject_unaudited_apk` to always-accept turns **8 of 11** tests red. Deleting a case from the bash `--self-test` matrix *and* loosening the rule for it still fails the independent vitest fixture — the two layers genuinely don't trust each other, as designed.

**Trivy — all six images green. Which was proved how:**

| Image | Result | Proved by |
|---|---|---|
| `m365-graph-read-executor` | `trivy_exit=0`, 0 hits, `libcrypto3/libssl3-3.5.8-r0` | **locally** (built + scanned, `--severity HIGH,CRITICAL --exit-code 1`) |
| `m365-communications-executor` | `trivy_exit=0`, 0 hits, `3.5.8-r0` | **locally**, same settings |
| `m365-graph-actions-executor` | `trivy_exit=0`, 0 hits, `3.5.8-r0` | **locally**, same settings |
| `breeze-api` (`docker/Dockerfile.api`) | pass | **CI**, #4257 run 33321719827 |
| `breeze-web` (`docker/Dockerfile.web`) | pass | **CI**, same run |
| `breeze-portal` | pass | **CI**, same run |

The three executors are the only images this PR can affect, and they were exactly the ones CI could not report on: `security.yml` triggers only on `pull_request: branches: [main]` with no `workflow_dispatch`, so this stacked PR gets **no Trivy run at all** — `gh pr checks` here is not evidence. In #4257's run the job aborted at the read-executor scan (`Total: 2 HIGH`), so the two executors after it never scanned; that is now resolved. Caveat: local builds are arm64, CI is amd64; the Alpine package version (`3.5.8-r0`) is identical across both.

**Tests:** `vitest run src/config/` — **505 passed** (15 files). `--self-test` 18/18 · full `check-supply-chain-hardening.sh` exit 0 · `tsc --noEmit` 0 · `eslint` 0 · `shellcheck -S error` clean · `bash -n` OK.

**Status:** review-clean, awaiting maintainer merge. **Merge #4257 first** — this is stacked on it.

**Flagged, not fixed here:** `apps/m365-graph-actions-executor` has no hardening block at all despite handling Graph **write** actions; the script itself calls this an acknowledged inconsistency. Worth its own issue.

ToddHebebrand:
Closing as superseded. Merged #4236 independently landed everything this PR does:

- all three M365 executor Dockerfiles now carry `apk upgrade --no-cache libcrypto3 libssl3` on `main` (verified: `m365-graph-read`, `m365-graph-actions`, `m365-communications`)
- `check-supply-chain-hardening.sh` already has the narrowed `AUDITED_APK_UPGRADE` rule plus `reject_unaudited_apk`, with both executor call sites wired through `require_audited_openssl_upgrade`
- `dockerfileOpensslUpgrade.test.ts` is present on main

The only artifact here that main lacked was `executorApkPolicy.test.ts` — the suite that proves the narrowed rule still *rejects*. That has been salvaged and adapted to main's factoring in #4328, with both mutation controls run rather than assumed.

This branch also conflicted with #4319 on `check-supply-chain-hardening.sh`, so rebasing it would have been wasted work.

PR #4265, opened 2026-08-30
fix(api): honor the reboot policy when a patch job partially fails (#4228)

### LINKED ISSUE #4228: Required reboot is skipped when a patch job partially fails
### Description

A scheduled patch job can install most patches successfully, report `Reboot Required: Yes`, but skip the reboot entirely if one unrelated patch in the same job fails.

Example observed behavior:

- 13 patches selected
- 12 installed successfully
- 1 failed
- Patch Operation History reports `Reboot Required: Yes`
- Overall operation status is `Failed`
- Reboot policy is `If required`
- Device is not rebooted

If one of the successfully installed patches requires a reboot, the reboot policy should still be honored even if another patch in the same job fails.

### Steps to Reproduce

1. Configure a Windows device with automated patching.
2. Set reboot policy to `If required`.
3. Run a patch job containing multiple patches.
4. Allow most patches to install successfully while one patch fails.
5. Confirm that the operation reports `Reboot Required: Yes`.
6. Observe whether Breeze reboots the device.

In my case, 12 of 13 patches installed successfully, one patch failed, the operation reported that a reboot was required, but Breeze did not reboot the device.

### Expected Behavior

Reboot requirements should be evaluated independently from the overall patch job success state.

If at least one successfully installed patch requires a reboot and the configured reboot policy allows it, Breeze should either:

- reboot the device as configured, or
- explicitly surface why the reboot was skipped.

A partial patch failure should not silently suppress a reboot that is still required by successfully installed patches.

### Actual Behavior

The overall patch operation becomes `Failed` because one patch failed.

Although the operation reports `Reboot Required: Yes`, the device is not rebooted when the reboot policy is set to `If required`.

### Component

Agent (Windows)

### Version

_No response_

### Relevant Logs

```shell

```

### Additional Context

Image

ToddHebebrand:
## The bug

A scheduled patch job installs 12 of 13 patches, one fails, the operation reports `Reboot Required: Yes` — and the device never reboots, even with the reboot policy set to `If required`. Nothing in the logs explains it.

## Root cause

The Windows agent is correct and is **not** changed here. `executePatchInstallCommand` (`agent/internal/heartbeat/heartbeat.go`) walks the whole batch, ORs `rebootRequired` across every *successful* install, and emits the full summary — `{ success: false, installedCount: 12, failedCount: 1, rebootRequired: true, results: [...] }` — on stdout. But because `failedCount > 0` it returns `Status: "failed"` / `ExitCode: 1`.

Server-side, `recordDeviceExecution` (`apps/api/src/jobs/patchJobExecutor.ts`) computes `anyRebootRequired` correctly from that payload — and then gated the **entire** reboot-policy evaluation behind `if (overallSuccess)`:

```ts
const overallSuccess = finalCommand?.status === 'completed' && ...   // false: agent said "failed"

if (overallSuccess) {
  const rebootEval = await evaluateRebootPolicy(deviceId, rebootPolicy, anyRebootRequired);
  ...
}
```

One failed patch out of thirteen makes `overallSuccess` false, so `evaluateRebootPolicy` was never called at all. The `If required` policy was never consulted, the reboot the other twelve installs needed was dropped, and — because the guard simply skipped the block — **not one log line** recorded that a reboot had been suppressed. That silence is why the UI kept showing `Reboot Required: Yes` next to a device that never restarted.

## The fix

Whether a reboot is needed is a property of **what actually installed**, not of the job's aggregate status.

- **Evaluate on what landed, not on overall success.** A new `anyPatchInstalled` — the agent's own `installedCount`, falling back to a per-patch entry with `status: 'installed' | 'rolled_back'` — gates the policy call. A partially failed job whose successful patches require a reboot now honors `if_required` (and every other policy).
- **A run that installed NOTHING still skips.** There is nothing to finalize, and evaluating unconditionally would let an `always` policy reboot a device the job never touched. This is logged, not silent.
- **Guard the static-flag fallback.** `anyRebootRequired` fell back to `approvedPatches.some(p => p.requiresReboot)` when the result was unparsable. Those flags describe what *would* be needed had every approved patch installed — an assumption that only holds for a success-shaped run. On a failed or timed-out command we have no idea which patches landed, so they must not manufacture a reboot. The fallback is now conditioned on `overallSuccess`, which leaves fully successful runs behaving exactly as before.
- **Log the decision every way it can go** — dispatched (explicitly naming the partial failure, so an operator reading "the job failed but the box rebooted" can tell intent from accident), declined by the policy (carrying the policy's own reason, including the deferred maintenance-window case), and not evaluated because nothing installed.

### On surfacing the skip reason

Reboot *dispatches* are already durably audited: `schedule_reboot` is in `commandQueue`'s `AUDITED_COMMANDS`, so `executeReboot` writes an `agent.command.schedule_reboot` audit row with the reason and delay. There is no equivalent persisted per-device surface for a *skip* — `patch_job_results` rows are per-patch, and `patch_job_results.rebooted_at` has no writer anywhere in the API today — so the skip reason lands in the structured log rather than inventing a schema surface for it. Worth noting separately: the UI's `Reboot Required: Yes` is read from the agent payload on `device_commands`, not from `patch_job_results`, so it was reporting the truth all along; only the dispatch was missing.

## Tests

New suite in `apps/api/src/jobs/patchJobExecutor.test.ts`, driving the real per-device worker processor end to end (prepare → poll → record) against the exact command shape the agent produces:

| Test | Guards |
|---|---|
| dispatches the reboot when one patch fails but an installed patch requires one | the reported bug — **red before the fix** (`evaluateRebootPolicy`: 0 calls) |
| records that the dispatched reboot followed a partially failed job | the dispatch log — **red before** |
| records why the reboot was skipped when the policy declines it | the declined-by-policy log — **red before** |
| does not reboot when every patch failed and nothing was installed | no reboot on a total failure, and it says so — **red before** (log assertion) |
| does not manufacture a reboot from static flags when a failed run is unparsable | the fallback guard |
| does not reboot when the install command never came back | timeout path |
| still falls back to static requiresReboot flags on a fully successful unparsable run | **behavior preserved** — green before *and* after |

Red-then-green confirmed: 4 failing → `43 passed (43)`.

## Verification

- `npx vitest run src/jobs/patchJobExecutor.test.ts` — 43 passed
- `npx vitest run src/services/patchRebootHandler.test.ts src/jobs/patchSchedulerWorker.test.ts src/jobs/patchSchedulerWorker.dbcontext.test.ts` — 55 passed
- `npx tsc --noEmit` (apps/api) — clean
- `npx eslint` on both changed files — clean

Server-side only; no schema, migration, or agent changes.

Closes #4228

### DIFF
--- a/apps/api/src/jobs/patchJobExecutor.test.ts
+++ b/apps/api/src/jobs/patchJobExecutor.test.ts
@@ -1546,4 +1546,96 @@ describe('post-patch reboot policy is independent of overall job success (#4228)
     expect(evaluateRebootPolicy).toHaveBeenCalledWith('device-1', 'if_required', true);
     expect(executeReboot).toHaveBeenCalled();
   });
+
+  it('honors a reported rebootRequired: false over the static patch flags', async () => {
+    // The approved set statically claims patch-1 requires a reboot, but the agent
+    // installed it and reported `rebootRequired: false` — Windows does not always
+    // need the restart the catalog advertises. The agent observed the real
+    // machine, so its `false` wins. Drop the `??` and read the static flags
+    // directly and this device gets an unnecessary reboot.
+    vi.mocked(evaluateRebootPolicy).mockResolvedValue({
+      shouldReboot: false,
+      reason: 'No installed patch requires reboot',
+      deferred: false,
+    });
+
+    await runDeviceExecution({
+      approvedPatches: TWO_PATCHES, // patch-1 has requiresReboot: true
+      rebootPolicy: 'if_required',
+      command: agentCommandRow({
+        success: true,
+        installedCount: 2,
+        failedCount: 0,
+        rebootRequired: false,
+        results: [
+          { id: 'patch-1', externalId: 'KB5000001', status: 'installed', rebootRequired: false },
+          { id: 'patch-2', externalId: 'KB5000002', status: 'installed', rebootRequired: false },
+        ],
+      }),
+    });
+
+    expect(evaluateRebootPolicy).toHaveBeenCalledWith('device-1', 'if_required', false);
+    expect(executeReboot).not.toHaveBeenCalled();
+  });
+
+  it('treats a per-patch installed entry as an install even when installedCount is 0', async () => {
+    // The two signals disagree: the summary counter says nothing installed, the
+    // per-patch array says patch-1 did. `anyPatchInstalled` ORs them, so the
+    // reboot still gets evaluated. Reading only `installedCount` would silently
+    // reinstate #4228 for any agent whose counter is absent, stale, or wrong.
+    vi.mocked(evaluateRebootPolicy).mockResolvedValue({
+      shouldReboot: true,
+      reason: 'Installed patch requires reboot',
+      deferred: false,
+    });
+
+    await runDeviceExecution({
+      approvedPatches: TWO_PATCHES,
+      rebootPolicy: 'if_required',
+      command: agentCommandRow({
+        success: false,
+        installedCount: 0,
+        failedCount: 1,
+        rebootRequired: true,
+        results: [
+          { id: 'patch-1', externalId: 'KB5000001', status: 'installed', rebootRequired: true },
+          { id: 'patch-2', externalId: 'KB5000002', status: 'failed', error: '0x80070005' },
+        ],
+      }),
+    });
+
+    expect(evaluateRebootPolicy).toHaveBeenCalledWith('device-1', 'if_required', true);
+    expect(executeReboot).toHaveBeenCalled();
+  });
+
+  it('evaluates the reboot policy for a rollback-only run with no net installs', async () => {
+    // Deliberate: `rolled_back` counts as "the device changed". Uninstalling a
+    // patch can require a restart just as installing one can, so a rollback that
+    // reports rebootRequired must reach the policy even though nothing was
+    // installed. Pinned because it is the least obvious arm of anyPatchInstalled
+    // and reads like a copy-paste at a glance.
+    vi.mocked(evaluateRebootPolicy).mockResolvedValue({
+      shouldReboot: true,
+      reason: 'Installed patch requires reboot',
+      deferred: false,
+    });
+
+    await runDeviceExecution({
+      approvedPatches: TWO_PATCHES,
+      rebootPolicy: 'if_required',
+      command: agentCommandRow({
+        success: false,
+        installedCount: 0,
+        failedCount: 1,
+        rebootRequired: true,
+        results: [
+          { id: 'patch-1', externalId: 'KB5000001', status: 'rolled_back', rebootRequired: true },
+          { id: 'patch-2', externalId: 'KB5000002', status: 'failed', error: '0x80070005' },
+        ],
+      }),
+    });
+
+    expect(evaluateRebootPolicy).toHaveBeenCalledWith('device-1', 'if_required', true);
+    expect(executeReboot).toHaveBeenCalled();
+  });
 });
--- a/apps/api/src/jobs/patchJobExecutor.test.ts
+++ b/apps/api/src/jobs/patchJobExecutor.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 const shared = vi.hoisted(() => ({
   getJobMock: vi.fn(),
@@ -105,6 +105,7 @@ import {
 } from './patchJobExecutor';
 import { resolveApprovedPatchesForDevice } from '../services/patchApprovalEvaluator';
 import { queueCommandForExecution } from '../services/commandQueue';
+import { evaluateRebootPolicy, executeReboot } from '../services/patchRebootHandler';
 
 function createSelectChain(rows: any[] = []) {
   const chain: any = {};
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer`, `silent-failure-hunter`, then a post-hoc `pr-test-analyzer` and `comment-analyzer` pass.

**Findings:** 1 correctness/observability finding + 3 coverage gaps → all addressed; 0 outstanding.

- `code-reviewer`: **0 findings** in all four categories. It verified the red-then-green empirically rather than by inspection — reverting `patchJobExecutor.ts` to `origin/main` and rerunning made exactly the 4 partial-failure tests fail, while the three tests covering intentionally-unchanged behavior stayed green. It also confirmed the five `db.select` `mockImplementationOnce` stubs match the real call sequence, and that the new describe block is last in the file so its `vi.resetAllMocks()` + fake timers cannot leak into the pre-existing suites.
- `silent-failure-hunter`: **1 actionable finding.** The reboot decision is now read out of the agent's JSON summary, which made the pre-existing bare `catch {}` around `JSON.parse` load-bearing — a corrupted or truncated stdout became the one remaining way a genuine partial success still reads as "nothing installed", i.e. the exact shape of #4228 reached through the one path the fix had not touched. Worse, the skip log stated `no patch installed successfully` as fact when the true state was "we could not read the result". Fixed in `449a9cb`: an unparsable result now reports to Sentry with a `console.warn` breadcrumb, and the two skip reasons read differently. A test asserts the new `captureException` **and** that the log does *not* claim the stronger wording; it fails against the previous commit, so it discriminates.
- Its other two answers were "leave it alone", with reasons worth recording: routine `declined`/`not evaluated` outcomes should stay `console.log` rather than Sentry (they are ordinary fleet operation at 10k agents and this file already reserves `captureException` for infra anomalies), and the `maintenance_window` deferred case is **not** a drop — `apps/api/src/jobs/maintenanceRebootWorker.ts` sweeps `devices.pending_reboot` independently. Added a test pinning the `(deferred)` log so the "every branch logs" claim is enforced rather than asserted.

**Coverage gaps closed in `127dd976`** (`pr-test-analyzer`, test-only — no source change). Every test up to that point drove `rebootRequired` and `installedCount` in *agreement*, leaving the code paths that exist precisely to handle them disagreeing unguarded:

| New test | Regression it catches |
|---|---|
| `honors a reported rebootRequired: false over the static patch flags` | dropping the `??` and reading `approvedPatches.some(...)` directly — an unnecessary reboot when the agent installed the patch and observed no restart was needed |
| `treats a per-patch installed entry as an install even when installedCount is 0` | reading only `installedCount` in `anyPatchInstalled`, which silently reinstates #4228 for any agent whose counter is absent, stale, or wrong |
| `evaluates the reboot policy for a rollback-only run with no net installs` | pins `rolled_back` counting as "the device changed" — deliberate, since an uninstall can require a restart just as an install can, and it is the least obvious arm of the predicate |

Verified by regression, not by inspection: dropping the `??` reddens **exactly** the first test and nothing else; dropping the `results.some(...)` OR reddens **exactly** the other two. That each one is the sole detector of its own regression is the evidence these were real gaps rather than restatements of existing coverage.

`comment-analyzer`: **clean bill** — no comment-accuracy or comment-rot findings.

**Tests:** `apps/api` affected suites green — `vitest run src/jobs/patchJobExecutor.test.ts` (**47 passed**, 11 new) and `vitest run src/services/patchRebootHandler.test.ts src/jobs/patchSchedulerWorker.test.ts src/jobs/patchSchedulerWorker.dbcontext.test.ts` (55 passed). `tsc --noEmit` clean; `eslint` clean on both changed files.

**CI (head SHA `127dd976`):** Test API, Lint, Lint Agent (Go), Test Agent, Test Agent (race), Security Audit, CodeQL and **all four Integration Tests shards** pass. Three reds, none attributable to this branch — the full diff vs `origin/main` is two files under `apps/api/src/jobs/`, with **zero** Go, schema, migration, dependency or Dockerfile changes:
- `Check Migrations` — the `v0.109.0-rc.3` tag was pushed off `fix/s0-pam-actuation-lifecycle` and is not an ancestor of `main`, so the immutability guard sees 8 "deleted" migrations on every branch. Red on `main` and on unrelated PR #4266.
- `Trivy Image Scan` — OpenSSL CVE-2026-14456 in the M365 Graph Read Executor image; fix in flight as open PR #4263.
- `Test Agent (Windows)` — a 600s panic-timeout in `agent/internal/sessionbroker`: `TestNamedPipeListenAndAccept` hangs in `Broker.StopAcceptingAndWait` (`broker.go:859`), a named-pipe accept-loop shutdown deadlock. It **passed on the two earlier SHAs of this same branch** against byte-identical agent code, so it is a pre-existing flake/deadlock in the agent, not a consequence of a TypeScript-test-only commit. Worth its own issue if it recurs.

**Follow-up filed:** #4267 — `patch_job_results` records **every** patch as failed when one fails, because `perPatchResult?.success` reads a field the agent never emits (it sends `status: 'installed' | 'failed' | 'rolled_back'`). Same root-cause family as #4228, but it changes DB-recorded statuses that feed compliance reporting, so it wants its own tests and review rather than being folded in here. Note it does **not** affect the reporter's Patch Operation History view, which reads the agent payload straight off `device_commands`.

**Status:** review-clean across all four analyzers, awaiting maintainer merge. Not merged, and #4228 left open pending your call.

PR #4266, opened 2026-08-30
fix(web): respect browser hour-cycle fallthrough + migrate ad-hoc time formatting (#4231)

### LINKED ISSUE #4231: Add 24-hour time format option / respect user locale in charts and timestamps
### Problem

Breeze currently uses 12-hour time formatting with AM/PM in parts of the UI, including performance charts.

In European environments, including Poland, a 24-hour clock is generally more natural and easier to read. For example, the 24h performance graph currently shows labels such as `02:05 AM` and `01:15 PM` instead of `02:05` and `13:15`.

The same issue can affect readability across patch schedules, maintenance windows, event history and other timestamp-heavy views.

A consistent time-format preference would improve localization and make the UI feel more natural outside regions that primarily use the 12-hour clock.

### Proposed Solution

Add a user or account preference for time formatting:

- `Auto / Browser locale`
- `12-hour`
- `24-hour`

Ideally, `Auto` would use the user's browser/account locale via standard locale-aware date formatting.

The selected preference should be applied consistently across:

- performance charts
- patch schedules
- maintenance windows
- event/activity history
- reports
- other timestamps displayed in the Web Dashboard

For example:

`01:15 PM` → `13:15`

This would avoid hardcoding one time format for all users and provide better localization for international deployments.

### Alternatives Considered

No real workaround considered.

Also, despite the number of GitHub issues I’ve opened recently, I genuinely enjoy using BreezeRMM. Most of these reports are coming from actively testing it in real-world scenarios and wanting to help improve the product.

### Component

Web Dashboard

### Scope

Small (UI tweak, config option)

ToddHebebrand:
## Summary

Re-scoped per maintainer comment on the issue (https://github.com/LanternOps/breeze/issues/4231#issuecomment-5470743158): the 12h/24h preference already exists (Profile → Appearance, and a partner-level Regional default), and honors correctly through the central formatting helper for most views. Two real bugs remained:

**Part A — browser-detect fallthrough (the bug).** `getEffectiveTimeFormat` in `apps/web/src/lib/dateTimeFormat.ts` resolved `explicit ?? readTimeFormatPreference()`, returning `undefined` whenever nothing was stored. With no explicit override, `withUserTimeFormatOptions` then left the formatter's hour-cycle options untouched, so the hour cycle silently fell back to whatever the app UI locale's Intl default is (English UI = 12-hour AM/PM) — even when the user's browser/OS reports a 24-hour clock. A browser-hour-cycle detector (`detectBrowserTimeFormat`, `appearance.ts`) and a preference-then-browser chain (`readResolvedTimeFormatPreference`) already existed but were never wired into the formatting path. Fixed by having `getEffectiveTimeFormat` fall through to `readResolvedTimeFormatPreference()` instead of the bare stored-preference reader, so the chain is now: explicit per-call override → stored preference → browser/OS detection.

  Side effect noted in the review: `getEffectiveTimeFormat` now always resolves to a concrete `'12h' | '24h'` (browser detection never returns `undefined`), so `withUserTimeFormatOptions`'s `!timeFormat` early-return is effectively dead for callers going through the standard helpers. Checked all six existing tests in `dateTimeFormat.test.ts` — none depended on the previous "leave hour-cycle untouched" behavior for a time-bearing format, so no assertions needed to change.

**Part B — migrate ad-hoc call sites.** Re-grepped `apps/web/src` for every raw `toLocaleString`/`toLocaleTimeString`/`toLocaleDateString` call site rather than trusting the issue's illustrative list. The real in-scope set is call sites that render a **time** component (and therefore ignore the 12h/24h preference entirely) — pure `toLocaleDateString` (date-only, never shows AM/PM) call sites are unaffected by this preference and were left alone as out of scope for this bug. Migrated onto the central `formatDateTime` helper:

- `components/organizations/Pax8OrgTab.tsx` (2 call sites — subscription `lastSeenAt`, order `updatedAt`)
- `components/settings/OfficeAddinBindingsPage.tsx` (local `formatDate` wrapper)
- `components/software/DeploymentProgress.tsx` (local `formatDateTime` wrapper — aliased the central import to avoid a name collision)
- `components/devices/DeviceMonitoringTab.tsx` (local `formatTimestamp`)
- `components/devices/DeviceComplianceTab.tsx` (local `formatTimestamp`)
- `components/devices/DevicePatchStatusTab.tsx` (local `formatDateTime` wrapper — aliased import; the file's separate date-only `formatDate` helper was left untouched, out of scope)
- `components/configurationPolicies/ComplianceStatusTab.tsx` (local `formatTimestamp`)
- `components/settings/TdSynnexSftpPanel.tsx` (inline call)
- `components/settings/AiAgentSchedulesSection.tsx` (inline call)

**Exclusions (out of scope, left as-is):**
- Every pure `toLocaleDateString` call site found in the sweep (~30 files: `PatchList.tsx`, `AutomationList.tsx`, `ScriptList.tsx`, `DeviceList.tsx`, etc.) — date-only output never shows AM/PM, so the 12h/24h preference doesn't apply to them. Migrating these to the central `formatDate` would be a separate locale-consistency improvement, not part of this bug.
- `lib/hydration-safe.ts` — deliberately pins `'en-US'` explicitly; the filename and the pinned locale indicate it exists specifically to avoid SSR/CSR hydration mismatches from locale-dependent formatting.
- `components/devices/ProcessDrilldownPanel.test.tsx` — uses raw `toLocaleString()` only to compute its own expected-value string for an assertion; the component under test (`ProcessDrilldownPanel.tsx`) already calls the central `formatDateTime`.
- Bare `Number.prototype.toLocaleString()` calls (e.g. `pax8OrderUi.ts`, quantity/byte-size formatting) — not dates at all.

## Testing

- **Part A (red → green):** Added a test to `dateTimeFormat.test.ts` that mocks `Intl.DateTimeFormat`'s locale-undefined resolution path to simulate a 24h browser/OS with no stored preference. Confirmed it fails against pre-fix code (`03:45 PM` instead of `15:45`), then passes after the fix.
- `cd apps/web && npx vitest run src/lib/dateTimeFormat.test.ts` — 7/7 passed.
- Full affected-component suite (all 9 migrated components' co-located tests): `npx vitest run <9 test files>` — 94/94 passed. One test (`OfficeAddinBindingsPage.test.tsx`) needed its `@/lib/i18n` mock extended with `getFallbackFormattingLocale`/`i18n` — routing through the central helper now exercises `resolvedFormattingLocale()`, which the old ad-hoc `toLocaleString()` call never touched.
- `npx vitest run src/lib/i18n/translationCoverage.test.ts` and `src/lib/i18n/extractionQuality.test.ts` — both green; no user-visible strings changed.
- `npx astro check` — 0 errors (pre-existing unrelated warnings only).

Closes #4231


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer + pr-test-analyzer (Sonnet), both independent passes against the pushed branch.

**Findings:** 0 confirmed consequential findings from either agent.
- code-reviewer: verified all 9 migrated call sites preserve prior fallback/timezone/date-style behavior exactly; confirmed `getEffectiveTimeFormat` always resolving to a concrete `'12h'|'24h'` (vs. previously possibly `undefined`) has no other callers depending on the `undefined` case; confirmed the `formatDateTime as formatDateTimeCentral` import alias in `DevicePatchStatusTab.tsx`/`DeploymentProgress.tsx` (needed to avoid colliding with each file's pre-existing local `formatDateTime` wrapper) is applied consistently with no call-site mix-ups. No CLAUDE.md violations.
- pr-test-analyzer: independently reverted the Part A fix locally and confirmed the new `dateTimeFormat.test.ts` case is genuinely discriminating (fails with `03:45 PM` vs expected `15:45` pre-fix, passes post-fix — not vacuous). Confirmed skipping new test files for the Part B formatting-swap call sites is reasonable (existing co-located suites all green, underlying helper already unit-tested); flagged as a minor, non-blocking observation (4/10 severity) that none of the 9 migrated components' own tests assert on rendered timestamp text, so a future wiring regression there would be visually obvious rather than caught by an automated test — not worth blocking on.

**Tests:** `dateTimeFormat.test.ts` — 7/7 passed (red confirmed pre-fix, green post-fix). Full affected-component suite (9 migrated components' co-located tests) — 94/94 passed. `translationCoverage.test.ts` and `extractionQuality.test.ts` — green, no user-visible strings changed. `npx astro check` — 0 errors.

**CI:** All checks green except two known pre-existing, unrelated failures — `Check Migrations` (off-main `v0.109.0-rc.3` tag reds this check repo-wide right now; this PR touches zero migration files) and `Trivy Image Scan` (CVE-2026-14456, pre-existing base-image finding). `CI Success` passed. `Test Web` (13m56s), `Test API` (17m46s), all 4 `Integration Tests` shards, `Build Web`, `Type Check`, `Lint`, and `Smoke Test` all passed.

PR #1059, opened 2026-06-02
feat(users): avatar file upload (POST/GET/DELETE /users/me/avatar)

bdunncompany:
## What

Replaces the URL-only avatar field on **Settings → Profile** with a real file upload. Three new endpoints back it; files are stored on the existing `api_data` volume at `/data/avatars/<userId>.<ext>`, mirroring the `fileStorage.ts` pattern used by remote transfers. **No DB migration** — reuses the existing `users.avatar_url` column (now holds an internal `/api/v1/users/<id>/avatar` URL instead of an external one).

## API

- **POST `/users/me/avatar`** — multipart upload (single `file` field)
  - MIME allowlist: `image/png`, `image/jpeg`, `image/webp` (no SVG)
  - **Magic-byte verification** — does not trust the `Content-Type` header
  - 5 MB cap via `bodyLimit` middleware (413 on overflow)
  - Atomic write (tmp + rename), unlinks a prior file at a different extension
  - Sets `users.avatar_url = /api/v1/users/<id>/avatar`; audit `user.avatar.upload`
- **GET `/users/:id/avatar`** — auth-required, streams bytes, `Cache-Control: private, max-age=300`, weak ETag + 304 support
- **DELETE `/users/me/avatar`** — unlinks file, clears `avatar_url`; audit `user.avatar.delete`

## Behavior change (please confirm the call)

`avatarUrl` is **removed from the `PATCH /users/me` schema**. Avatars now flow exclusively through the upload endpoints; the strict schema makes any client still sending `avatarUrl` get a `400` (rather than a silent drop). This is the deliberate intent — switching from "paste a URL" to "upload a file." Consequences for the existing SOC2 audit tests, all updated in this PR:

- The audit-attribution test that used `avatarUrl` as its example non-email field now uses `name` (same intent: partner-scope self-change → `user.profile.update`).
- The `rejects avatarUrl with javascript: scheme` test is removed — the `avatarUrl` URL-scheme refinement no longer exists, and "unknown field rejected" is already covered by the strict-schema test.

If you'd rather **keep** the URL field alongside upload, say so and I'll rework to additive.

## Web

`ProfilePage.tsx`: 96px preview, file picker, drag-drop onto the preview, and a Remove button (shown only when an avatar is set). Helper: "PNG, JPG, or WebP. Max 5 MB." `updateUser` fires on success/delete so the topbar avatar refreshes immediately. The URL text field is removed.

[...]

PR #4268, opened 2026-08-30
docs(security): document the api_data volume and Redis on the backup page (#3958)

### LINKED ISSUE #3958: Backup docs omit the api_data volume and Redis entirely — self-hosters without S3 have no documented uploads backup
**Gap:** `apps/docs/src/content/docs/security/backup.mdx` documents backup for exactly three things — Postgres (`pg_dump`/`pg_restore`), object storage (S3/MinIO via `aws s3 sync` / `mc mirror`), and config (`.env`, certs). It never tells a self-hoster to back up the **`api_data` volume** mounted at `/data`.

**Why it matters:** a self-hosted deployment with no S3/MinIO configured has *no documented backup path for uploaded or generated files at all*. `PATCH_REPORT_STORAGE_PATH` defaults under `/data` (`/data/patch-reports`), so patch reports — and anything else written to the API data volume — are silently outside the documented backup surface. `scripts/backup.sh` has no flag for it either.

**Second, related gap: Redis.** `docs/registers/internal-trust-register.md:7-20` lists rate-limit buckets, access/refresh-token revocation lists, BullMQ queue + worker coordination, and event-bus streams as Redis-resident, each annotated "Yes, assuming Redis persists". `DISASTER_RECOVERY.md:258` says jobs are durable in Redis and resume after restart. But `backup.mdx` has no Redis section and `scripts/backup.sh` has no Redis flag, so nothing tells an operator that losing the `redis_data` volume drops in-flight and scheduled jobs across ~30 queues and lets revoked JWTs survive to natural expiry.

**Fix:**
- Add an `api_data` / `/data` section to `backup.mdx` covering the volume itself, and call out that the S3 section does not substitute for it when object storage is unconfigured.
- Add a Redis section stating plainly what is lost and recommending a snapshot taken alongside the Postgres dump.
- Consider adding both to `scripts/backup.sh` so the documented tool matches the documented policy.

**Found via** #3930 — a self-hoster's TrueNAS SCALE deployment review, where `PATCH_REPORT_STORAGE_PATH` sits in `api_data` with no S3 configured.

ToddHebebrand:
Closes #3958

## The gap

`apps/docs/src/content/docs/security/backup.mdx` documented exactly three things -- Postgres, object storage, and config -- and never mentioned that a Compose deployment keeps state in **six** named Docker volumes. Nothing on the page hinted `api_data` existed, so its omission was silent. Three other pages carried the same incomplete list; all are fixed here (see "Contradicting pages" below).

## What the facts turned out to be

The facts were established from the compose files and API source, not from the issue body. Two of them contradict the issue's premise, and the docs now say so explicitly.

**`api_data` holds patch compliance reports, and nothing else.** `PATCH_REPORT_STORAGE_PATH: /data/patch-reports` is the only env var in either shipped compose file that points into the volume (`docker-compose.yml:122`, `deploy/docker-compose.prod.yml:94`). The other four `/data` env vars address `/data/binaries`, which is the *separate* read-only `binaries` volume mounted over that subpath (`docker-compose.yml:463`). `patchComplianceReportWorker.ts:191-201` writes the CSVs unconditionally with no S3 branch, and records `outputPath` on the Postgres row, so losing the volume leaves report rows whose download fails (`routes/patches/compliance.ts:533`).

**The issue's "uploads" premise is wrong, and correcting it is part of the fix.** No user uploads live on `api_data`:

- Avatars are Postgres `bytea` on `users.avatar_data`. `avatarStorage.ts:7-15` records that this *replaced* a `/data/avatars/<id>.<ext>` implementation precisely because depending on the `api_data` volume was unreliable (EACCES → 500, #1059).
- Catalog images and quote/proposal images are likewise Postgres `bytea`.
- Software package uploads **require** object storage and fail closed without it: `softwareUploads.ts:205,603` and `software.ts:1031` return `503`, and the web UI grays the control out up front via `routes/config.ts:43`. Upload chunks buffer in `/tmp`, which both compose files mount as a 64 MB `tmpfs`.

An operator who believes uploads sit on that volume builds the wrong backup job, so the page now says this outright.

**Redis should be documented, not backed up.** The issue suggested "recommending a snapshot taken alongside the Postgres dump". This PR deliberately does not, and says why on the page itself:

- Redis already runs `appendonly yes` (`docker-compose.yml:615`), so it survives restarts. The requirement is to *preserve* the volume (never `docker compose down -v`), which is not the same as backing it up.
- There is no consistent-snapshot mechanism between Postgres and Redis. Restoring a stale Redis into a live stack re-injects job state for work Postgres already recorded as complete.
- Recurring schedules are **not** lost: every scheduler re-registers its repeatables at boot (`agentLogRetention.ts:90-104`, `alertWorker.ts:354-378`), and the schedules themselves live in Postgres (`patchSchedulerWorker.ts:4` is a 60-second tick that scans config-policy schedules and writes `patch_jobs` rows).
- Refresh-token family revocation is **not** lost: it is dual-written to `refresh_token_families.revoked_at`, which `tokenRevocation.ts:288-293` documents as the durable side of the split, "surviving Redis flushes".
- What *is* lost is bounded and now stated precisely: queued and in-flight jobs, and access-token revocation entries -- so a revoked but unexpired access token is honoured again for at most 15 minutes, or 2 hours for a viewer token (`jwt.ts:5,14`).

## Changes

**Published docs**

- **`security/backup.mdx`** -- "What Gets Backed Up" becomes "What Persists, and What to Back Up": a complete inventory of all six volumes plus object storage and config, each with a verdict and whether a script flag exists. New sections for `api_data` (backup/restore recipe writing into `$BACKUP_DIR`, with the caveats below), `redis_data`, and `binaries`.
- **`security/overview.mdx`** -- the Backup and Recovery components line carried the same incomplete three-item list.
- **`deploy/upgrades.mdx`** -- the Compose upgrade `<Steps>` had no backup step. Adds one first, because migrations are forward-only and rolling the tag back does not roll the schema back.

**Contradicting pages found by a repo-wide sweep** -- a docs fix that leaves a contradicting page is half a fix:

- **`docs/security/SECURITY_PRACTICES.md`** carried a *verbatim twin* of the `security/overview.mdx` line fixed here, and is linked from `README.md`.
- **`docs/operations/BACKUP_RESTORE.md`** is the internal runbook for these very scripts and told the reader `--all` covers everything.
- **`docs/operations/DISASTER_RECOVERY.md`** (cited in the issue itself) omitted the volume from "What to Back Up" and said nothing about not restoring Redis.

## Review findings addressed

A review round raised three points on the first commit, all fixed in `affcdd7`:

1. The `tar` recipe wrote to `$(pwd)` and sat right above a "Retention" section that does not apply to it. It now writes into `$BACKUP_DIR` with an explicit caveat that `BACKUP_RETENTION_DAYS` does not prune it and that cron does not inherit a working directory.
2. `tar` extraction *overlays* a volume rather than replacing it, so a report written after the backup survives a restore and mixes in. Now stated, with an optional clear step for an exact restore.
3. The page did not explain *why* there is no script flag for the volume, only that there isn't one. Now explained (the script shells out to `pg_dump`/S3/`openssl`; reading a Docker volume needs the Docker socket).

## Dead code removed

The sweep found an orphaned runbook doing exactly what this PR documents as unsafe, so it is deleted here (`d709a41`):

- `docs/operations/UPGRADE_v0.1_to_v0.2.md`
- `scripts/ops/backup.sh`
- `scripts/ops/restore.sh`

The runbook told operators to snapshot Redis (`redis-cli --rdb`) and restore that snapshot over a live database. Leaving an executable runbook that contradicts the documented policy is worse than the original gap. None of it could run regardless: all three set `COMPOSE_FILE="${REPO_ROOT}/docker/docker-compose.prod.yml"`, which does not exist (the real file is `deploy/docker-compose.prod.yml`).

Verified orphaned against `origin/main` before deleting: nothing references the runbook by path or title, and the only referrers of the two scripts were the runbook's own lines 19 and 62. The surviving `scripts/ops/offsite-backup.sh:74` and `restore-test.sh:154` call the **root-level** `scripts/backup.sh` / `scripts/restore.sh`, which both remain; both survivors still parse under `bash -n`. The docs-site index builder only globs `.mdx` under `apps/docs`.

**`scripts/ops/verify-monitoring.sh` was proposed for deletion in the same batch and is deliberately kept.** It shares the same broken `COMPOSE_FILE` path, but it is not orphaned: `docs/operations/DEPLOY_PRODUCTION.md:113` instructs operators to run it, and `docs/notes/SOC_A1.1_CAPACITY_NOTES.md:146` cites its output as SOC A1.1 validation evidence. That makes it a broken script to fix, not dead code to remove. Its broken compose path is a pre-existing bug this PR does not touch -- worth a follow-up issue.

## Not extending `scripts/backup.sh`, deliberately

No `--volumes`/`--redis` flag was added. The issue listed it as "consider". A Redis flag would encode advice this PR argues against, and an `api_data` flag needs Docker socket access from a script that only shells out to `pg_dump`/`aws`/`openssl`. The documented recipe gives the capability without expanding the script's surface.

## Verification

Docs-only, so `ci.yml` does not run (its `paths-ignore` excludes `apps/docs/**` and `**/*.mdx`). The gating workflow is **Docs CI**, and all three of its steps were run locally on this branch:

- `pnpm --filter @breeze/docs check` -- 0 errors, 0 warnings, 0 hints
- `pnpm --filter @breeze/docs build` -- 167 pages built, complete
- `pnpm test:docs-automation` -- 5/5 pass

The new in-page anchor was verified against the built HTML (`id="the-api_data-volume"` exists and resolves), and `https://docs.breezermm.com/security/backup/` was curled for a 200 before being cited from the internal runbook.

Note: `Check Migrations` and `Trivy Image Scan` are red repo-wide right now (#4227 tracks the first, PR #4236 fixes it) and are not attributable to this PR.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, comment-analyzer (run as a claim-by-claim fact check against the compose files and API source), plus a repo-wide sweep for pages contradicting the new policy.

**Findings:** 8 raised → 7 addressed, 1 escalated to you below; 0 outstanding.

- **1 Critical (fact check)** — the page said the relative `PATCH_REPORT_STORAGE_PATH` default resolves to `/app/data/patch-reports`. It does not. That came from reading `docker/Dockerfile.api` (WORKDIR `/app`), which is only built by the security-scan job; the shipped image is built from `apps/api/Dockerfile`, whose final WORKDIR is `/app/apps/api`. Fixed in `9addbb6`, reworded to lead with the durable fact (the default is relative, so it lands in the container's writable layer) instead of a brittle absolute path.
- **3 Important (code review)** — the `tar` recipe wrote to `$(pwd)` directly above a Retention section that does not cover it; `tar` extraction overlays a volume rather than replacing it, so post-backup reports survive a restore; and the page never said *why* there is no script flag. All fixed in `affcdd7`.
- **3 contradicting pages (sweep)** — all fixed in `affcdd7`. `docs/security/SECURITY_PRACTICES.md` carried a verbatim twin of the `security/overview.mdx` line fixed here and is linked from the README; `docs/operations/BACKUP_RESTORE.md` told the reader `--all` covers everything; `docs/operations/DISASTER_RECOVERY.md` (cited in the issue itself) omitted the volume and said nothing about not restoring Redis.
- **1 escalated, not touched** — `scripts/ops/backup.sh:38` snapshots Redis and `scripts/ops/restore.sh:63-66` restores it over a live database, driven by `docs/operations/UPGRADE_v0.1_to_v0.2.md`. That is executable code doing precisely what this PR documents as unsafe. It is already dead (those scripts reference `docker/docker-compose.prod.yml`, which does not exist, and nothing links to that runbook), so I recommend deleting all three as a follow-up rather than rewriting a v0.1→v0.2 guide. Deleting shipped files is your call, so I left them.

Two review notes deliberately not actioned: the unused `Tabs, TabItem` import in `backup.mdx:9` predates this PR, and `docker compose down` in the restore recipe is safe as written (it never removes named volumes without `-v`).

**Tests:** Docs CI green on `9addbb6` — and all three of its steps also run locally: `pnpm --filter @breeze/docs check` (0 errors/warnings/hints), `pnpm --filter @breeze/docs build` (167 pages), `pnpm test:docs-automation` (5/5). `ci.yml` does not run for this PR by design: its `paths-ignore` excludes `apps/docs/**` and `**/*.mdx`. The new in-page anchor was verified against the built HTML, and `https://docs.breezermm.com/security/backup/` returned 200 before being cited from the internal runbook.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3958 — the one open decision is the orphaned `scripts/ops` Redis snapshot/restore path above.

ToddHebebrand:
**Review run 2** — scoped to the deletion commit `d709a41` only, since a deletion is where a reviewer earns its keep.

**Findings:** 0 requiring changes. The reviewer independently reproduced the orphan analysis rather than taking the commit message's word for it.

- **Dangling references:** none. Verified by full path, by basename, and by the runbook's title text across docs, `.github/`, `package.json`, Makefiles and Dockerfiles.
- **Surviving scripts intact:** `scripts/ops/offsite-backup.sh:74` and `restore-test.sh:154` call the **root-level** `scripts/backup.sh` / `scripts/restore.sh`, not the deleted `ops/` pair. Both roots still exist; all survivors pass `bash -n`. This was the highest-risk item and it is clear.
- **Nothing unique lost:** the deleted `ops/restore.sh` was the only Redis-restore path in the repo, and removing it *is* the intended policy change — it is the exact procedure this PR documents as unsafe.
- **"Unrunnable" claim confirmed:** `docker/docker-compose.prod.yml` does not exist; the real file is `deploy/docker-compose.prod.yml`, so every `compose` call in all three scripts failed immediately.

**One file was pulled from the deletion batch.** `scripts/ops/verify-monitoring.sh` was proposed for removal alongside the others and is **kept**. It shares the same broken `COMPOSE_FILE` path, but it is not orphaned: `docs/operations/DEPLOY_PRODUCTION.md:113` tells operators to run it during a production deploy, and `docs/notes/SOC_A1.1_CAPACITY_NOTES.md:146` cites its output as SOC A1.1 validation evidence. Deleting it would have broken a live runbook and dropped cited compliance evidence.

That does surface a **pre-existing bug worth a follow-up issue**: because of the same bad path, `DEPLOY_PRODUCTION.md:113` currently points operators at a command that cannot run. `git log --follow` puts that line in `3b723dc87`, long before this PR, and this commit does not touch the file — so it is out of scope here, but it should be fixed (the path should be `deploy/docker-compose.prod.yml`). Flagging rather than filing, since that is your call.

**CI (final).** The `.sh` deletions push this PR outside `ci.yml`'s `paths-ignore` for the first time, so the full suite ran. **54 checks green, 2 red**, both confirmed **not attributable**:

- `Check Migrations` — this PR touches zero migration files, and the same job fails on `main` at `23001934d`, which is this branch's exact base commit. Tracked by #4227 (closed), fixed by #4236 (open).
- `Trivy Image Scan` — this PR touches no Dockerfile, lockfile or source file, and `Security Scanning` fails on the same three `main` commits.

All four Integration Test shards, Test API, Test Web, Type Check, the agent builds and Smoke Test passed. `Docs CI` green on `d709a41`.

**Status:** review-clean, awaiting maintainer merge. Not merging, not closing #3958.

Issue #4245, opened 2026-08-30
[ee/workspace] Two more top-level err.code SQLSTATE checks are dead under Drizzle — plus a lint rule to end the bug class


ToddHebebrand:
Follow-up from #4020 / PR #4240, which fixed the two `apps/api` instances of this bug class.

`drizzle-orm/postgres-js` catches the postgres-js `PostgresError` and rethrows a `DrizzleQueryError` whose own `.code` is `undefined` — the SQLSTATE lives on `.cause`. Any check reading `err.code` directly returns false for every Drizzle-issued statement. `apps/api/src/utils/pgErrors.ts` already documents this and carries the canonical helpers.

**Two instances remain in `ee/workspace`** (verified on `origin/main`):

1. `ee/workspace/src/services/ingestJobsService.ts:102`
   ```ts
   function isUniqueViolation(err: unknown): boolean {
     return typeof err === 'object' && err !== null && (err as { code?: string }).code === '23505';
   }
   ```
   Consequence: the concurrent-insert retry in `ensureJob` never fires — a genuine 23505 race surfaces as a 500 instead of being retried.

2. `ee/workspace/src/routes/sources.ts:146`
   ```ts
   function isForeignKeyViolation(error: unknown): boolean {
     return typeof error === 'object' && error !== null &&
       'code' in error && (error as { code?: unknown }).code === '23503';
   }
   ```
   Consequence: a bad `crawlDeviceId` returns 500 instead of the intended 400.

**These are in scope, not false positives:** `ee/workspace/src/hostTypes.ts:44` declares `export type WorkspaceDatabase = PostgresJsDatabase;` — the extension runs on the host's org-scoped Drizzle connection, so both handlers see the wrapped `DrizzleQueryError`, not a raw `PostgresError`.

**Why this wasn't folded into PR #4240:** `@breeze/ext-workspace` has no dependency on `apps/api`, so reaching `pgErrors.ts` requires moving the helpers into a shared package (`packages/shared`?) or duplicating them behind a documented contract. That is a design call, not a one-liner.

### Suggested scope

1. Decide where the canonical SQLSTATE helpers live so `ee/*` can import them, then fix both call sites.
[...]

PR #4269, opened 2026-08-30
fix(api): pin the partner-API provisioning 23505 to the org slug index (#3982)

### LINKED ISSUE #3982: [API/Web] Partner-API provisioning mislabels any 23505 as a slug conflict; org UI discards the specific slug-conflict message
Two follow-ups from the review of #3974 (org slug uniqueness), neither merge-blocking, but the first is a live correctness bug that #3974 *activates*.

**1. `partnerApi/provisioning.ts` maps any 23505 to a slug conflict.** At `apps/api/src/routes/partnerApi/provisioning.ts:127` and `:282` the handler uses an unconstrained `isUniqueViolation(error)` check and reports `partner_provisioning_slug_conflict`. Before #3974 there was no unique index on `organizations (partner_id, lower(slug))`, so that branch was effectively dead for slugs; now it is live — and it will also catch a unique violation raised by *any other* insert in the same transaction and mislabel it as a slug conflict. A partner-API caller then gets a confident, wrong diagnosis for an unrelated constraint.

Fix: constrain it to the specific index, e.g. `isPgUniqueViolation(error, 'organizations_partner_slug_uniq')`, and let anything else fall through to the generic error path.

**2. The web UI throws away the server message #3974 exists to produce.** `apps/web/src/components/settings/OrganizationsPage.tsx:325` throws a generic i18n string on failure, discarding the API's specific body — which now distinguishes "slug is already in use" from "reserved by a deleted organization". The second case is the non-obvious one (the slug is held by a soft-deleted org, deliberately, so import reactivation cannot be broken), and it is exactly the case a user cannot diagnose on their own. Surface the server's message.

Related: `apps/api/src/routes/orgs.ts:1922` — on the non-override update path the 23505 aborts the request's outer transaction. Benign today because nothing runs after the 409, but it becomes a real bug the moment a write is added below it. Worth a comment at minimum.

ToddHebebrand:
Closes #3982

## The bug (issue item 1)

`POST /partner-api/v1/provisioning/organizations` reported **every** unique violation raised by its create transaction as `partner_provisioning_slug_conflict`:

```ts
} else if (isUniqueViolation(error)) {          // any 23505, from any index
  return c.json({ error: 'An organization with this slug already exists.',
                  code: 'partner_provisioning_slug_conflict' }, 409);
```

Before #3967 there was no unique index on `organizations (partner_id, lower(slug))`, so that branch was effectively dead for slugs and nothing exercised how wrong it was. #3967 made it live.

The slug index is not the only unique index that INSERT is subject to: `organizations` also carries `organizations_id_partner_id_unique` on `(id, partner_id)`. That one is practically unreachable today (`id` defaults to a random UUID), and — corrected during review — the table's AFTER triggers are **not** a source either: the only one that writes another table inserts into `partner_export_configuration_org_state` with `ON CONFLICT DO NOTHING`, so it structurally cannot raise 23505.

So the argument for naming the index is about **construction, not an imminent failure**. A check that answers `partner_provisioning_slug_conflict` for a constraint it never verified hands an unattended provisioning client a confident, **wrong** diagnosis — go fix a slug that was never the problem — which is worse than no diagnosis, because it sends the retry loop somewhere it can never succeed. The set of indexes on this table is not fixed, and the next one added must not silently re-arm the bug.

**Fix:** pin the check to the slug index *by name*.

```ts
} else if (isPgUniqueViolation(error, ORG_SLUG_UNIQUE_INDEX)) {
```

Anything else falls through to the pre-existing generic error path, which `app.onError` logs and captures to Sentry before answering 500 — the honest answer for a unique violation this route does not model. Not a silent failure: the full error still reaches the logs and Sentry, and (added in review) the route now emits one `console.warn` naming the constraint that actually fired, so triage does not have to unwrap `.cause` off a raw `DrizzleQueryError` by hand.

### Why the canonical helper, not a fixed-up local one

The deleted local `isUniqueViolation` is exactly how this bug class keeps shipping. `drizzle-orm/postgres-js` rethrows a `DrizzleQueryError` whose **own `.code` is `undefined`** — the SQLSTATE and `constraint_name` live on `.cause`. A check reading the top-level `err.code` is dead for every Drizzle-issued statement. Found five times in this repo already (#3998, #4020, and two in `ee/workspace` filed as #4245). `utils/pgErrors.ts` walks the `.cause` chain and is now the only implementation this route uses.

## One name, one place

`ORG_SLUG_UNIQUE_INDEX` moves into `apps/api/src/db/schema/orgs.ts` and is used by the `uniqueIndex()` declaration itself, so the string is written exactly once. `routes/orgs.ts`, `services/aiToolsOrgs.ts` and the new provisioning check all import it. Every mapping of this 23505 to a 409 has to name the index *exactly*; a copy-pasted name is a silent way for the mapping to stop firing, which is the same failure mode as the bug above.

Imported from the **concrete** schema module (`../db/schema/orgs`), not the `db/schema` barrel — several suites mock that barrel with a non-partial factory, where a newly added export throws `No "X" export is defined on the mock` at the exact moment the 409 mapping runs. (Caught locally: adding it to the barrel reddened two `orgs.test.ts` cases.)

## Issue item 3 — comment on the org-update path

`routes/orgs.ts` now records the asymmetry the issue flagged. The create path runs its insert in its **own** transaction (`runOutsideDbContext(() => withSystemDbAccessContext(...))`), so a 23505 poisons only that inner tx. The non-override update branch does not: `runUpdate()` executes on the request's ambient context and `withDbAccessContext` is a real `baseDb.transaction(...)`, so the 23505 aborts the **request's** transaction and every later statement in it would fail with 25P02. Benign today only because the catch returns the 409 and nothing after it touches the DB — and a hazard for the next write added there. Comment, not a behaviour change, per the issue.

## Issue item 2 (web) — already shipped, no change needed

The org UI no longer discards the server's message. #3989 covered exactly this and shipped as #4003 (`ae1fb4749`); every mutating call to an organizations endpoint now routes through `runAction`, so `extractApiError` surfaces the API's own 409 text — including the non-obvious *"That organization slug is still reserved by a deleted organization"*. Verified by enumerating every POST/PUT/PATCH to `/orgs/organizations*` across `apps/web/src`: all on `runAction`, except the setup-wizard's two call sites (`OrganizationSetupStep.tsx`), which hand-roll but still call `extractApiError` — so they surface the API body too, and are not the discard bug. The four remaining generic-string throws in `OrganizationsPage.tsx` are all on **GET** reads. No org edit/rename UI sends `slug` at all today.

The issue's line reference (`OrganizationsPage.tsx:325`) now points at the archived-orgs fetch — the file was reworked by #4003 after the issue was filed.

## Tests

The existing conflict case used a flat `{ code: '23505' }` fixture — a shape the driver **never** produces for a Drizzle-issued statement, and precisely why a top-level-`.code` check can pass its own test while being dead in production. Every fixture now goes through a `drizzleUniqueViolation()` helper that builds the real `DrizzleQueryError` wrapper with SQLSTATE + `constraint_name` on `.cause`.

Three cases, and they **discriminate** rather than merely exercise the branch:

| Case | Old code | New code |
|---|---|---|
| slug index 23505 | 409 slug conflict | 409 slug conflict |
| **non-slug 23505** (`organizations_id_partner_id_unique`) | **409 slug conflict (wrong)** | **500, no slug claim** |
| slug index named only in the driver message | 409 | 409 |

The middle row is red against `main` and green after; the other two pass against both, which proves the fixture actually reaches the branch.

## Verification

- `vitest run src/routes/partnerApi/provisioning.test.ts src/routes/orgs.test.ts` — 292 passed
- `vitest run src/services/aiToolsOrgs.test.ts src/routes/partnerApi/writeSurface.test.ts src/services/quickSupportOrg.test.ts src/services/orgImport/orgImport.test.ts src/utils/pgErrors.test.ts src/db/autoMigrate.test.ts` — all green
- `eslint` on all changed files — clean
- No migration, no schema DDL change (the index name string is byte-identical), no web change.

## Review round

`/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer.

- **code-reviewer: 0 findings.** Independently confirmed reachability of the new check, no drift risk from moving the index name into a const, and no missed call site repo-wide.
- **pr-test-analyzer:** verdict *genuinely discriminating* — it ran all three fixtures through both the old and new implementations and confirmed only the non-slug case flips. Two accuracy findings on the fixture/comment, both fixed in `11f34e157`. One residual gap accepted, see below.
- **silent-failure-hunter:** no silent failure introduced; the targeted constraint log was its actionable finding and is now in. Declined its parity suggestion to log the *matched* slug conflict — the sibling handlers in `orgs.ts` log because a 409 there means their pre-check lost a race; provisioning has no pre-check, so a 409 there is ordinary client error and logging it would be spam.
- **comment-analyzer: 1 Critical** — the AFTER-trigger claim above was wrong. Verified against the migration myself and rewritten. Also dropped an unverifiable "found five times, two in ee/workspace" tally (a grep finds one `ee/` site).

**Accepted residual gap:** no integration test drives this route against real Postgres. `organizationSlugUnique.integration.test.ts` already proves the index exists and that `isPgUniqueViolation` matches a real 23505 from it; combined with a driver-faithful unit fixture, a new authenticated service-principal integration suite is not worth its cost for an error-mapping change.

**CI note:** `Check Migrations` and `Trivy Image Scan` are red repo-wide right now (#4227 tracks the first, PR #4236 fixes it) — not attributable to this PR.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (type-design-analyzer skipped: no new types).

**Findings:** 4 raised → all 4 addressed in `11f34e157`; 2 declined with reasons; 0 outstanding.

- **CRITICAL (comment-analyzer)** — my comment claimed the org INSERT's AFTER triggers were a live source of a foreign 23505. **Wrong**, and I verified it myself before rewriting: the only trigger on `organizations` that writes another table inserts into `partner_export_configuration_org_state` with `ON CONFLICT (resource, org_id) DO NOTHING` (`2026-07-24-partner-export-configuration-material-state.sql:354-360`), so it structurally cannot raise 23505. The real second unique index is `organizations_id_partner_id_unique` on `(id, partner_id)` — which is exactly the constraint the regression test already used. Comment rewritten to cite the mechanism that exists and to say plainly that it is practically unreachable today: the case for naming the index is **construction, not an imminent failure**. PR body corrected too.
- **silent-failure-hunter** — the unmodelled-23505 rethrow reached Sentry as a raw `DrizzleQueryError`, leaving triage to unwrap `.cause` by hand. The route now emits one `console.warn` naming the constraint that actually fired before rethrowing.
- **pr-test-analyzer** — the fixture stamped `.name = 'DrizzleQueryError'`, which the real class never sets; dropped, and the docstring now states what the fixture is faithful *about* rather than claiming to be an instance. The test comment no longer claims this bare-Hono app exercises the real Sentry-wired `onError`.
- **comment-analyzer** — dropped an unverifiable "found five times, two in ee/workspace" tally (a grep finds one `ee/` site) and a third restatement of the DrizzleQueryError mechanism; both defer to the canonical JSDoc on `utils/pgErrors.ts`.

**Declined, with reasons:**
- *Log the matched slug conflict too, for parity with `orgs.ts`.* The sibling handlers log because a 409 there means their slug pre-check lost a race — a signal the pre-check broke. Provisioning has no pre-check, so a 409 there is ordinary client error; logging it would be spam.
- *Add an integration test driving this route against real Postgres.* `organizationSlugUnique.integration.test.ts` already proves the index exists and that `isPgUniqueViolation` matches a real 23505 from it. With a driver-faithful unit fixture on top, a new authenticated service-principal integration suite isn't worth its cost for an error-mapping change. Recorded as accepted residual risk.

**code-reviewer returned 0 findings** — it independently traced reachability of the new check, confirmed no drift risk from moving the index name into a const, and found no missed call site repo-wide.

**Tests:** the regression case is genuinely discriminating — pr-test-analyzer ran all three fixtures through both the old and new implementations and confirmed only the non-slug case flips (old `true` → 409, new `false` → 500) while the two slug cases pass against both. The new log assertion was mutation-checked locally: disabling the `console.warn` turns that test red.

Local: `provisioning.test.ts` 24/24; `orgs.test.ts` + `aiToolsOrgs.test.ts` + `writeSurface.test.ts` 301/301; `quickSupportOrg` / `orgImport` / `pgErrors` / `autoMigrate` green; eslint clean.

CI on `11f34e157`: **Type Check pass, Lint pass, Test API pass, Test Web pass, Integration Tests 4/4 shards pass.**

**Not attributable:** `Check Migrations` and `Trivy Image Scan` are red. Verified repo-wide, not caused here — `main` itself is red on `Check Migrations`, and the failing step is the shipped-migration immutability guard diffing `apps/api/migrations/` against `v0.109.0-rc.3`, a tag that is not an ancestor of `main` (#4227; fix in draft PR #4236). This PR adds no migration and the guard never reads `db/schema/`. Trivy is an Alpine `libcrypto3` CVE in an unrelated executor base image, also failing on `main`.

**Status:** review-clean, awaiting maintainer merge. Issue #3982 stays open and assigned.

Issue #3632, opened 2026-08-16
[Web] "t" as a useEffect dependency re-runs effects on locale change — one drops a live VNC session (38 sites, 12 doing network work)


bdunncompany:
## Summary

`t` from `useTranslation` is used as a `useEffect` dependency in 38 places. It gets a new identity on `languageChanged`, so those effects re-run when the locale changes. Most are redundant reads, but one tears down a **live remote-control session**, and it is reachable on an ordinary page load for any user with a saved non-English locale.

This is the same class as the one you fixed in #3628 ("the load effect listed `t` in its deps... picking a language refetched the partner and reset the timezone and currency the user had just chosen"), and the same class an adversarial pass found in #3579. Filing the inventory rather than fixing piecemeal, since the interesting question is whether this wants a lint rule.

## The mechanism, precisely

react-i18next 17.0.10 defaults to `bindI18n: 'languageChanged'` and `bindI18nStore: ''` (`node_modules/react-i18next/src/defaults.js:3`), and this repo sets no `react` overrides (`apps/web/src/lib/i18n/index.ts:284`). So:

- `addResourceBundle` alone does **not** re-render `useTranslation` — store events aren't bound.
- `applyLocale` calls `changeLanguage`, which emits `languageChanged`, and `useTranslation` then hands back a fresh function (it caches per language and recreates on change, `useTranslation.js:143`).

Worth noting because the existing comment at `CisHardeningPage.tsx:75-76` says `t` "changes identity when i18n resources finish loading" — directionally right, but it's the language change that follows the load which actually does it. I nearly repeated that imprecision.

**Why this is not just a manual-language-switch problem:** English namespaces are eager and init is synchronous, so an English user never sees the re-run. But a user with a persisted non-English locale hydrates in English and *then* has the saved locale applied after hydration (`i18n/index.ts:126`, `:246`, `:320`). For them, `changeLanguage` fires on every normal page load, after the component has mounted and connected.

## The one that matters: a live VNC session is dropped

`apps/web/src/components/remote/VncViewerPage.tsx:32` has `t` in its deps (`:61`). Each run POSTs `/tunnels/:id/ws-ticket` and calls `setWsUrl()`.

[...]

PR #4274, opened 2026-08-30
fix(web): scope the devices list to the selected org on first load after login (#4147)

### LINKED ISSUE #4147: Devices list ignores selected org on first load after login (race vs org-context resolution)
Reported by **semotech** in #bugs (Discord, message 1542677837776167014).

**Report (PII-redacted):**

> Small UI bug on v0.107.0 - After login one is taken to the Devices Section and even tho the last Organization is selected all devices from all organizations appear in the list.  One needs to select another organization before they list aligns with the allocations.

## Triage analysis

On the very first Devices-page load after login, the device fetch races the org-context resolution and never self-corrects. `DevicesPage.fetchDevices` calls `fetchAllDevices()` -> `fetchWithAuth('/devices?...')`, which auto-injects `orgId` by reading `useOrgStore.getState().currentOrgId` synchronously at call time (auth.ts). `useOrgStore` (orgStore.ts) only resolves the real org (either via persisted-storage rehydration or the async `/orgs/partners` + `/orgs/organizations` calls kicked off by `OrgSwitcher`'s own mount effect) slightly after the DevicesPage mount effect fires `fetchDevices(controller.signal)`. If that resolution hasn't landed yet, `currentOrgId` is still null, so no `orgId` param is sent and the API returns devices across every organization the partner user can see. Crucially, `fetchDevices` is memoized with `useCallback(..., [t])` and its mount effect depends only on `[fetchDevices]` — neither depends on `currentOrgId` — so once the org store *does* resolve moments later (and the OrgSwitcher pill correctly shows the last-used org), the device list is never re-fetched. The only thing that fixes it is picking a different org in OrgSwitcher, but that 'fix' is actually just `applyOrgSwitch` doing a `window.location.reload()` (orgSwitch.ts) — a full reload re-runs boot with the org context already resolved, so the next fetch happens to be scoped correctly. This matches the report exactly: correct org shown in the selector, unscoped device list, fixed only by forcing a reload via re-selecting an org.

## Affected code

- `apps/web/src/components/devices/DevicesPage.tsx:291-527` — fetchDevices() calls fetchAllDevices() with no orgId; useCallback deps are [t] only, and the mount effect deps are [fetchDevices] — never re-runs when currentOrgId resolves/changes.
- `apps/web/src/lib/devicesFetch.ts:90-128` — fetchAllDevices() hits '/devices' via fetchWithAuth with no explicit orgId param, relying entirely on the auto-injection in fetchWithAuth.
- `apps/web/src/stores/auth.ts:519-526` — fetchWithAuth reads _getOrgId() (a snapshot of the org store) at call time; if the org store hasn't resolved yet, no orgId is appended and the API returns all accessible orgs' devices.
- `apps/web/src/stores/orgStore.ts:98-261` — currentOrgId is resolved asynchronously — either via zustand persist rehydration or the fetchPartners/fetchOrganizations network calls kicked off elsewhere (OrgSwitcher) — so it isn't guaranteed to be set before DevicesPage's first fetch.
- `apps/web/src/lib/orgSwitch.ts:59-71` — Switching org calls window.location.reload() (or redirects); this full reload, not any reactivity, is why manually reselecting an org appears to 'fix' the list.

## Potential fix

Gate DevicesPage's initial fetch on the org store having resolved (e.g. wait for orgStore.organizationsLoaded or persist hydration) and/or add currentOrgId to fetchDevices's dependency array so the mount effect re-fires and refetches once the org context settles, instead of only being corrected by OrgSwitcher's full-page reload.

ToddHebebrand:
## The race, as actually established

Not "the caller forgot to pass an orgId" — the devices list never passes one.
`fetchWithAuth` auto-injects `?orgId=` by reading `useOrgStore.getState().currentOrgId`
**synchronously at call time** (`apps/web/src/stores/auth.ts:993`, provider registered at
`apps/web/src/stores/orgStore.ts:368`). The bug is that the fetch fires before that value
exists.

The missing link the issue body didn't have: **logout wipes the persisted org key.**
`evictLocalAuthState` (`apps/web/src/stores/auth.ts:1496`) does
`localStorage.removeItem('breeze-org')`. So:

1. Logout clears `breeze-org`. On the next login the org store hydrates with
   `currentOrgId: null` — zustand's `persist` rehydrates *synchronously* from
   localStorage, so a plain F5 has never reproduced this; only a login does.
2. The user lands on `/devices`. `DevicesPage`'s mount effect fires
   `fetchDevices()` → `fetchAllDevices()` → `fetchWithAuth('/devices?…')`.
   `_getOrgId()` returns `null`, no `orgId` is appended, and the API reads an
   absent orgId as *"every accessible org"*.
3. Moments later `OrgSwitcher`'s own mount effect (`OrgSwitcher.tsx:85`) finishes
   `fetchOrganizations()`, which auto-selects `organizations[0]`. The pill now shows
   an org — matching the report's "the last Organization is selected".
4. Nothing refetches. `fetchDevices` is `useCallback(…, [t])` and its effect is
   `[fetchDevices]`; neither observes `currentOrgId`. The list stays fleet-wide.
5. Re-picking an org "fixes" it only because `applyOrgSwitch` ends in
   `window.location.reload()` (`apps/web/src/lib/orgSwitch.ts`) — by then the org
   *is* persisted, so it rehydrates synchronously and the next fetch is scoped.

Proven, not inferred: the red run of the new test recorded the literal request
`/api/v1/devices?includeDecommissioned=true&limit=200&includeTotal=true` — no `orgId=`.

## The fix — re-fetch on scope, not suppress the first render

Keyed on the resolved org scope via the repo's existing `useOrgScope` hook:

```ts
const orgScope = useOrgScope();
const orgScopeResolving = orgScope.status === 'loading';
const orgContextFailed  = orgScope.status === 'error';
const orgScopeKey = /* 'status:…' | 'scope:all' | `org:${id}` */;

useEffect(() => {
  if (orgScopeResolving || orgContextFailed) return;
  const controller = new AbortController();
  fetchDevices(controller.signal);
  return () => controller.abort();
}, [fetchDevices, orgScopeResolving, orgContextFailed, orgScopeKey]);
```

**Re-fetching is the mechanism** (the effect is keyed on the scope, so any later
resolution or change re-issues the request); the early-return only keeps it from firing
while the key is not yet meaningful. Both are needed:

- **Keying alone** would leave the first, unscoped, fleet-wide cursor walk in place —
  a real cross-tenant query cost on a 10k-device fleet, plus a visible flash of other
  orgs' devices before it corrects.
- **Gating alone** would not survive a scope change that lands after the gate opens.
  There is now a test for exactly that (org A → org B while mounted), because a boolean
  `ready` gate would pass every other case in the suite.

`empty` (org list loaded, partner genuinely has zero orgs) is terminal, not transient, so
it fetches and settles rather than holding the skeleton.

**No new `t`/i18n dependency trap (#3632).** The added deps are booleans and a string
derived from the org store. The pre-existing `useCallback(…, [t])` on `fetchDevices` is
untouched (it already refetched on locale change; changing it is out of scope).

## Review round — three fixes folded in

`/pr-review-toolkit:review-pr` (code-reviewer, pr-test-analyzer, silent-failure-hunter,
comment-analyzer). Two reviewers independently flagged the same gap:

1. **The event-stream refetch bypassed the gate** (code-reviewer IMPORTANT,
   pr-test-analyzer CRITICAL). `useEventStream` connects on auth alone, so a
   `device.enrolled` / `device.decommissioned` push can land inside the pre-resolution
   window and fire the very unscoped request the gate exists to prevent — possibly
   landing *after* the scoped one. Now guarded on the same condition; nothing is lost,
   since the mount effect fetches as soon as the scope resolves. Covered by a test.
2. **`status: 'error'` silently rendered a cross-tenant list** (silent-failure-hunter
   IMPORTANT). With no selection to fall back on, the old behaviour fetched unscoped and
   rendered a fleet-wide list indistinguishable from a real org-scoped one —
   `ContextScopeLine` deliberately does *not* let the error state read as fleet view, so
   nothing on screen said otherwise. Now renders the shared `OrgLoadFailedState`, matching
   `DiscoveryPage` and `MonitoringAssetsDashboard`; its Retry re-runs the org resolution,
   which doubles as the escape hatch if the context is stuck. Only reachable with no
   selection at all — `deriveOrgScope` gives a concrete `currentOrgId` precedence over a
   later refetch failure, so a warm session is unaffected. No new i18n strings.
3. **Test gaps** (pr-test-analyzer): added org-switch-while-mounted, abort-of-the-
   superseded-request, and the `empty` scope; `settle()` now flushes several macrotasks
   plus a timer, since one tick cannot tell "never fetches" from "fetches after a delay"
   and the negative assertions are this suite's whole point.

One reported finding was **not** actioned, after verifying it is unreachable: a claim that
`resetSelection()` can strand the page on `status: 'loading'` forever. That branch
(`orgStore.ts:263`) is only reached when `organizations.length === 0`, which
`deriveOrgScope` maps to `empty` — terminal, and the page fetches. `resetSelection` has
exactly one caller.

## Tests

`DevicesPage.orgScope.test.tsx` deliberately runs end-to-end through the **real**
`orgStore` and **real** `fetchWithAuth`, stubbing only `globalThis.fetch`. DevicesPage
never passes an orgId itself, so mocking either side would have made "the request carried
the selected org" an assertion about the mock.

| Case | Asserts |
|---|---|
| unresolved → resolved | no `/devices` request while loading; then exactly one, carrying `orgId=org-a` |
| warm reload (resolved at mount) | exactly one request, scoped |
| explicit All-organizations | exactly one request, **no** `orgId=` |
| org list failed | `OrgLoadFailedState` renders; **no** device request at all |
| failure → recovery | one scoped request after the context recovers |
| enroll event mid-window | event refetch does not escape the gate |
| org A → org B while mounted | second request carries `orgId=org-b` |
| org change mid-flight | the superseded `AbortController` is aborted |
| `empty` (zero orgs) | fetches and settles rather than holding the skeleton |

Confirmed red before the fix: the unresolved→resolved and org-list-failed cases failed with
the unscoped URL above; the warm-reload and All-orgs cases passed on the unfixed component,
and are kept as regression guards on *this* diff (they catch a double-fetch or a fleet-view
hang that the bug-reproduction cases cannot see).

Also fixed a **vacuous mock** in the existing `DevicesPage.test.tsx`: it stubbed
`useOrgStore` with a function that ignored its selector, so every
`useOrgStore(s => s.currentOrgId)` read back the whole state object (truthy), and the
derived scope came out "resolved" by accident. Now selector-aware, defaulting to the
explicit All-organizations scope so the 69 existing tests keep the exact values they relied on.

## Verification

- `DevicesPage.orgScope.test.tsx` — 9 passed (red-first confirmed against the unfixed component)
- `DevicesPage.test.tsx` — 69 passed · `useOrgScope.test.ts` — 8 passed (86 total, one run)
- `orgStore.test.ts` + `orgStore.scope.test.ts` — 28 passed
- `eslint` on all three changed files — clean
- CI **Type Check** green (that job is what runs `astro check` for `apps/web`)

## Scope notes

- Touches only `apps/web/src/components/devices/DevicesPage.tsx` and its two test files.
  No overlap with PR #4243's device-**detail** files.
- `useAdvancedFilterIds` races the same way, but only when an advanced filter is seeded
  from the URL hash (never on a fresh login), and its ids are intersected with the
  now-correctly-scoped list — so the visible result stays right. Left alone rather than
  widening this PR.
- `Check Migrations` and `Trivy Image Scan` are red repo-wide right now (#4227 tracks the
  first; PR #4236 fixes it). Not attributable to this change.

Closes #4147


ToddHebebrand:

**Findings:** 5 raised → 4 addressed in `cd5b4a52e` / `47f03f5a8` / `098479849`; 1 rejected on verification; 0 outstanding.

- **Event-stream refetch bypassed the new gate** (code-reviewer IMPORTANT + pr-test-analyzer CRITICAL — flagged independently by both). `useEventStream` connects on auth alone, so a `device.enrolled` / `device.decommissioned` push can land inside the pre-resolution window and fire the very unscoped request the gate exists to prevent — possibly landing *after* the scoped one. Guarded on the same condition; nothing is lost, since the mount effect fetches as soon as the scope resolves. Regression test added.
- **`useOrgScope` status `'error'` silently rendered a cross-tenant list** (silent-failure-hunter IMPORTANT). With no selection to fall back on, the old path fetched unscoped and rendered a fleet-wide list indistinguishable from a real org-scoped one — `ContextScopeLine` deliberately does *not* let the error state read as fleet view, so nothing on screen said otherwise. Now renders the shared `OrgLoadFailedState`, matching `DiscoveryPage` and `MonitoringAssetsDashboard`; its Retry re-runs the org resolution, which doubles as the escape hatch if the context is stuck. No new i18n strings.
- **Test gaps** (pr-test-analyzer IMPORTANT ×2 + SUGGESTION). Added org-switch-while-mounted (the case that distinguishes *keyed* from merely *gated* — a boolean `ready` gate passes every other test in the suite), abort-of-the-superseded-request, and the `empty` scope. `settle()` now flushes several macrotasks plus a timer: one tick cannot tell "never fetches" from "fetches after a delay", and the negative assertions are this suite's whole point.
- **Not actioned, verified unreachable:** a claim that `resetSelection()` can strand the page on `status: 'loading'` forever. That branch (`orgStore.ts:263`) is only reached when `organizations.length === 0`, which `deriveOrgScope` maps to `'empty'` — terminal, and the page fetches. `resetSelection` has exactly one caller. Reported rather than silently dropped.

comment-analyzer verified every factual claim in the added comments against source (the `fetchWithAuth` injection point, the `breeze-org` eviction on logout, `OrgSwitcher`'s resolution effect, `applyOrgSwitch`'s reload) — 0 inaccuracies.

**Tests:** `DevicesPage.orgScope.test.tsx` 9 passed; `DevicesPage.test.tsx` 69 passed; `useOrgScope.test.ts` 8 passed (86 in one run, `--pool=forks --no-file-parallelism`). `orgStore.test.ts` + `orgStore.scope.test.ts` 28 passed. `eslint` clean on all three changed files. Red-first confirmed: against the unfixed component the suite recorded the literal unscoped request `/api/v1/devices?includeDecommissioned=true&limit=200&includeTotal=true` — no `orgId=`.

**CI as-of this comment:** 20 green (incl. **Lint**, Trivy Filesystem Scan, Gitleaks, Workflow Lint); 20 still queued — **Type Check**, **Test Web**, **Test API**, Integration Tests and the agent/build jobs have not reported on this head SHA yet. `Type Check` was green on the previous head; `Test Web` has not yet run against the review commits, so treat the CI picture as incomplete until those land.

**Known-red, not attributable:** `Check Migrations` and `Trivy Image Scan` fail repo-wide right now — they are red on `main` and on unrelated PRs. #4227 tracks the migrations one; PR #4236 fixes it. This branch adds no migrations.

**Status:** review-clean, awaiting the remaining checks and maintainer merge. Not merging and not closing #4147 — both are the maintainer's call.

Issue #4033, opened 2026-08-25
[CI] InvoiceWorkspace failure-nonce test is flaky — same commit passed and failed across runs


ToddHebebrand:
`src/components/billing/InvoiceWorkspace.test.tsx > InvoiceWorkspace > does NOT fire a queued Issue when the pending save fails — the whole point of the failure nonce` failed once in `Test Web` and then passed on a re-run of the same branch with no change to that code.

## Evidence it is a flake, not a real failure

- **Failed:** PR #4029, run [32889642724](https://github.com/LanternOps/breeze/actions/runs/32889642724) (job 97938271956), `Test Web` reported `2 failed | 636 passed`.
- **Did not fail on the equivalent main run.** The same base commit's main run [32887380424](https://github.com/LanternOps/breeze/actions/runs/32887380424) reported `1 failed | 637 passed` — only `DevicesPage.test.tsx`, which was the known #4009/#3994 copy collision fixed by #4023. `InvoiceWorkspace` passed there.
- **Passed on the rebase.** #4029 rebased onto green main with no source change beyond the rebase, and `Test Web` went fully green.
- **#4029 touches no billing code.** Its diff is `systemTools`, `devices/filesystem`, `remote/sessions`, `FileManager.tsx`, `remote.json` locales and `packages/shared/src/constants/agentFileTransfer.ts`. There is no path by which it affects `InvoiceWorkspace`.

So the same commit-under-test produced both outcomes, which is the definition.

## Why it is worth chasing rather than ignoring

The assertion is specifically about **not** firing a queued side effect when a pending save fails — a negative assertion guarding a nonce. Flaky negative assertions are the expensive kind: this one failing intermittently means it can also *pass* intermittently while the guard is broken, so it is unreliable in the direction that matters. A test that cannot be trusted to fail is not coverage.

Most likely shapes, in order: a timer or microtask not awaited (the "pending save" implies an in-flight promise the test may not be settling deterministically), or shared module state leaking between tests in the same file — note `vi.clearAllMocks()` does not clear implementations, which has bitten this repo before.

## Suggested next step

Run the file in isolation with `--repeat` to see whether it reproduces standalone or only under full-suite ordering. If it only fails in the full suite, it is cross-test state rather than the test itself.

Found while triaging the PR queue; not blocking anything today, since the branch that surfaced it is merged and green.

bdunncompany:
[...]

Issue #3980, opened 2026-08-25
[Web] InvoiceWorkspace failure-nonce test times out at 5s and reds main


ToddHebebrand:
## Description

`Test Web` failed on `main` (commit `fe6407861`, run [32764359539](https://github.com/LanternOps/breeze/actions/runs/32764359539)) with a single test timing out. That also failed the `CI Success` gate, so main was red.

```
FAIL src/components/billing/InvoiceWorkspace.test.tsx > InvoiceWorkspace >
  does NOT fire a queued Issue when the pending save fails — the whole point of the failure nonce
Error: Test timed out in 5000ms.
  src/components/billing/InvoiceWorkspace.test.tsx:214:3
```

Test Files 1 failed | 625 passed (626).

## Assessment

Most likely a load-sensitive flake rather than a real regression:

- The commit is a dependabot bump (`node-stream-zip`, `@react-native-async-storage/async-storage`) that does not touch billing.
- The same suite passed on the AI-agent wave PRs merged either side of it (#3953, #3976 — 6357 and 6393 web tests green).
- Timeouts in this repo's suites have repeatedly turned out to be machine-load artifacts that pass on isolated re-run.

Not yet confirmed, which is why this is filed rather than dismissed — a 5s default timeout on a test that awaits a deliberately-failing save is exactly the shape that becomes a real intermittent failure under CI contention.

## Proposed Fix

1. Re-run the job to confirm it is a flake and get main green.
2. If it recurs, give this test an explicit timeout or replace the implicit wait with a deterministic assertion on the failure nonce, rather than raising the global `testTimeout`.

## Affected Files

- `apps/web/src/components/billing/InvoiceWorkspace.test.tsx:214`

## Reported By

Internal — observed while auditing main's CI state before a release, 2026-08-24.

PR #4275, opened 2026-08-30
fix(api): exclude decommissioned and ephemeral devices from backup fan-out and the low-readiness gauge (#3968, #3969)

### LINKED ISSUE #3968: [API] Backup policy fan-out targets decommissioned devices — daily failing jobs and phantom low-readiness rows
## Description

Found while investigating the `BreezeBackupLowReadinessDevices` warning on US prod (2026-08-24).

A device that was decommissioned on 2026-08-08 (`CO-SPACE.decom-289a2245`, org `Comtodo`) is **still enrolled in its backup policy**. It has been getting a fresh `backup_jobs` row every day at 04:00 UTC ever since, each failing with `Agent not connected` — 14 jobs and counting. It also carries a `recovery_readiness` row scoring 0, and is one of the 4 devices pinning the low-readiness alert.

Decommissioning a device should stop backup dispatch to it.

## Root Cause

`resolveAllBackupAssignedDevices` resolves each assignment level to device IDs, and **not one of the five branches filters on `devices.status` or `devices.isEphemeral`**:

- `device` branch — `apps/api/src/services/featureConfigResolver.ts:1568-1576`
- `device_group` branch — `:1578-1592`
- `site` branch — `:1594-1601`
- `organization` branch — `:1603-1616`
- `partner` branch — `:1618-1632`

Every branch correctly re-tenants to `orgId` (that contract is well documented in the surrounding comments), but nothing excludes decommissioned or ephemeral records. So an org-level backup assignment sweeps up every row in `devices` for that org, dead ones included.

Downstream consequences:
- `backup_jobs` accumulates guaranteed-failing rows forever.
- `recovery_readiness` gets a row for a device that can never be verified, which feeds `breeze_backup_low_readiness_devices` (see the sibling gauge issue).
- Backup commands are dispatched toward a decommissioned agent.

## Proposed Fix

Add `status != 'decommissioned'` and `isEphemeral = false` to all five branches — ideally as one shared predicate so a future sixth branch can't miss it. Note the existing device-status exclusions elsewhere (`GET /metrics/`, `readFleetGauges`) use exactly this pair, so the semantics are already established.

Also worth deciding: should existing `recovery_readiness` rows for decommissioned devices be pruned at decommission time, or just filtered at read time? Decommission is not a delete, so `CORE_DEVICE_CASCADE_DELETE_TABLES` does not fire.

## Affected Files

- `apps/api/src/services/featureConfigResolver.ts` (primary — `resolveAllBackupAssignedDevices`, lines 1480-1640)
- `apps/api/src/routes/backup/readinessCalculator.ts` (secondary — `addMissingAssignedDevicesToReadiness` synthesises rows from the same resolver output)

## Evidence

US prod, org `Comtodo`:

```
        hostname         |     status     | agent_version |      last_seen_at       | n_backup_jobs | error
-------------------------+----------------+---------------+-------------------------+---------------+-------------------
 CO-SPACE.decom-289a2245 | decommissioned | 0.104.0       | 2026-08-08 18:49:03     |            14 | Agent not connected
```

Latest job for that device: `2026-08-24 04:00:00`, 16 days after decommission.

### LINKED ISSUE #3969: [API] breeze_backup_low_readiness_devices gauge counts decommissioned/ephemeral devices, contradicting its own doc comment
## Description

Found while investigating the `BreezeBackupLowReadinessDevices` warning on US prod (2026-08-24). One of the 4 devices inflating the gauge is a device decommissioned 16 days earlier.

`readFleetGauges` carries this doc comment:

> Ephemeral Quick Support devices and decommissioned records are excluded, the same exclusions `GET /metrics/` applies, so the gauge and the dashboard count the same fleet.

— `apps/api/src/routes/metrics.ts:1293-1295`

That is true of the `devices` query immediately below it, which does filter `isEphemeral = false` and `status != 'decommissioned'` (`:1303-1315`). It is **not** true of the readiness query directly underneath, which is a bare count over `recovery_readiness` with no join to `devices` and no status predicate:

```ts
const [readinessRow] = await db
  .select({ count: sql<number>`count(*)` })
  .from(recoveryReadinessTable)
  .where(sql`${recoveryReadinessTable.readinessScore} < ${BACKUP_LOW_READINESS_THRESHOLD}`);
```
— `apps/api/src/routes/metrics.ts:1322-1326`

So the gauge and the dashboard do *not* count the same fleet, and the comment actively misleads anyone triaging the alert.

## Root Cause

The exclusion pair was applied to the fleet query and never extended to the readiness query added in the same function.

## Proposed Fix

Join `recovery_readiness` to `devices` and apply the same two predicates, so the gauge matches both its comment and `GET /metrics/`.

Related: #3968 — the fan-out bug that produces readiness rows for decommissioned devices in the first place. Fixing that one removes the source; this one makes the gauge correct regardless of what is already in the table (decommission is not a delete, so historical rows persist).

## Affected Files

- `apps/api/src/routes/metrics.ts` (primary — `readFleetGauges`, lines 1297-1334)
- `apps/api/src/routes/metrics.test.ts` (coverage — currently asserts the raw count only, `:893`)

## Notes

Separately worth a look: the alert rule is `max by (region, host) (breeze_backup_low_readiness_devices) > 0`, a fleet-wide count with no org dimension. One customer with broken backups pins the warning indefinitely and the metric cannot say who. A ratio, or a bounded-cardinality per-partner breakdown, would make the page actionable.

ToddHebebrand:
Two halves of the same bug, fixed together because they are the same two exclusions applied on either side of one data flow: the fan-out that *creates* readiness rows, and the gauge that *counts* them. Splitting them would have put conflicting edits in adjacent code.

## #3968 — backup policy fan-out targeted decommissioned devices

`resolveAllBackupAssignedDevices` resolves each policy assignment to device IDs. **I verified the issue's count of five branches** — `device`, `device_group`, `site`, `organization`, `partner` — and **none** of them filtered on `devices.status` or `devices.isEphemeral`. Every branch correctly re-tenanted to `orgId`; the liveness of the device was simply never considered.

Consequence on US prod: a device decommissioned on 2026-08-08 kept collecting a fresh `backup_jobs` row every day at 04:00 UTC, each failing `Agent not connected`, plus a `recovery_readiness` row scoring 0.

`checkSchedules` (`apps/api/src/jobs/backupWorker.ts`) draws its entire device set from this resolver, so the fix stops the daily failing jobs at the source. `addMissingAssignedDevicesToReadiness` (`readinessCalculator.ts`) synthesises its phantom rows from the same resolver output, so it is fixed transitively — no separate edit needed there.

Fixed with **one** predicate bound once and reused by all five branches, so a sixth branch has to reach for the same name rather than re-derive it:

```ts
function backupTargetableDeviceCondition(): SQL {
  return sql`${devices.status} <> 'decommissioned' AND ${devices.isEphemeral} = false`;
}
```

Semantics established against this schema before filtering, not assumed: `decommissioned` is a member of the `device_status` pgEnum and is the schema's **soft delete** (`DELETE /devices/:id` only flips `status`; there is no `decommissioned_at` column), and `isEphemeral` marks Quick Support session devices the reaper purges ~6h after the session — already documented in `devices.ts` as excluded from license counts, listings, billing rollups and alert evaluation. This is the same pair `GET /metrics/`, `readFleetGauges`, and five fleet workers already use.

## #3969 — the low-readiness gauge counted them too

**Source of truth: the doc comment.** `readFleetGauges` already promised "Ephemeral Quick Support devices and decommissioned records are excluded, the same exclusions `GET /metrics/` applies, so the gauge and the dashboard count the same fleet." That was true of the devices query and false of the readiness query directly beneath it, which was a bare `count(*)` over `recovery_readiness` with no join to `devices` at all. The code was made to match the comment — not the other way round — because the promised behaviour is the correct one and the comment is what a responder reads while triaging the alert.

The readiness query now joins `devices` and carries the same two predicates. `INNER` is the right join: `recovery_readiness.device_id` is `NOT NULL` with an FK to `devices.id` and the row is cascade-deleted with the device, so the join can never drop a readiness row that should have been counted.

The doc comment was also tightened to say *both* queries carry the exclusions, so the next reader is not told a half-truth.

## Red-first evidence

Both fixes were driven by a failing assertion, watched fail before any source edit:

- **#3968** — 15 new assertions across all five branches (`featureConfigResolver.backupAssignments.test.ts`). Red: **10 failed / 9 passed** — exactly the 5 branches × 2 exclusions. The 5 control assertions (org re-tenanting still present) passed in the same run, which is what proves the harness captured a real predicate rather than nothing. Green after the fix: **19/19**.
- **#3969** — 4 new assertions (`metrics.test.ts`). Red: **3 failed / 1 passed**, the passing one being the control that the threshold filter is still there. Green after: **61/61**.

### On vacuous Drizzle assertions

This repo's known trap is asserting against a mock-shaped result that the real driver never produces. Neither new suite does that:

- The metrics assertions compile the **captured predicate** with the real `PgDialect` and match the resulting SQL text (`"devices"."status" <> 'decommissioned'`, `"devices"."is_ephemeral" = false`). A mock can be told to return any count; only the compiled text can show the filter reached Postgres. The pre-existing `count: 3` assertion passes just as happily with no filter at all, which is why it never caught this.
- The resolver assertions render the predicate tree the resolver handed to Drizzle (columns spliced in from the real call) and match on that. Delete the exclusion from the source and the text stops containing it.

## Verification

| Check | Result |
|---|---|
| `vitest run src/services/featureConfigResolver{,.timezone,.backupAssignments}.test.ts` | 68 passed |
| `vitest run src/routes/metrics.test.ts` | 61 passed |
| `vitest run src/routes/backup.test.ts src/routes/backup/{verificationService,dashboard,sla}.test.ts` | 51 passed |
| `vitest run src/jobs/backup{Worker.checkSchedules,SlaWorker}.test.ts src/routes/backup/{jobs,hyperv,mssql}.test.ts` | 38 passed |

All run single-fork (`--pool=threads --maxWorkers=2`). No schema, migration, table or column changes, so no cascade / export-policy / RLS-allowlist registration applies.

## One decision left for the maintainer

#3968 asks whether existing `recovery_readiness` rows for decommissioned devices should be **pruned at decommission time** or **filtered at read time**. This PR does neither, deliberately:

- The gauge (#3969) is now correct regardless of what is already in the table, which is the part that was paging anyone.
- The fan-out no longer creates new ones.
- What remains is historical rows surfacing in the per-org backup dashboard (`listRecoveryReadiness` / `getBackupHealthSummary().readiness.lowReadinessCount`). Pruning is destructive and decommission is explicitly not a delete; read-time filtering is a behaviour change to a customer-facing list. Both are judgement calls that belong to you, not to a bug fix.

**Recommendation:** read-time filtering, in a follow-up — it is non-destructive and keeps the org dashboard consistent with the now-corrected gauge. Say the word and I will open it.

## CI note

`Check Migrations` and `Trivy Image Scan` are red repo-wide right now — they fail on `main` and on unrelated PRs (#4227 tracks the first; PR #4236 fixes it). Neither is attributable to this change, which touches no migrations and no dependencies.

Closes #3968
Closes #3969


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, all on the full `origin/main...HEAD` diff).

**Findings:** 2 raised → 1 fixed in ba23ce95b, 1 deliberately declined; 0 outstanding.

1. **comment-analyzer + code-reviewer (both, independently) — comment inaccuracy.** The new INNER-join comment said the readiness row is "cascade-deleted with the device". The FK on `recovery_readiness.device_id` is actually `ON DELETE no action` (`2026-03-28-backup-verification.sql:65-70`); the delete happens app-side via `CORE_DEVICE_CASCADE_DELETE_TABLES` / `deleteDeviceCascade`. Substance was right, mechanism was not stated — and a comment promising something the DDL does not say is precisely the failure mode #3969 exists to correct. Fixed in ba23ce95b.

2. **silent-failure-hunter — coverage shrinks with no labelled signal (declined, deliberately).** A device auto-decommissioned by the uninstall-intent reaper (`offlineDetector.ts:807`) now leaves the backup-assigned set silently: `coveragePercent` loses it from numerator *and* denominator at once, and `missed_backup` SLA breaches stop firing for it. Not acted on here, because (a) the reaper only fires 24h after the agent signals uninstall **and** stops heartbeating, so the agent is already gone — the "signal" removed is a guaranteed-failing job row, i.e. noise; (b) a decommissioned device already vanishes from `GET /metrics/`, the fleet gauges and every fleet worker, so backup coverage matching that is consistent, whereas an `excludedDevices` counter on only the backup dashboard would not be; (c) the proposed mitigation is a new field on two API responses plus UI to consume it — a product surface with no repo precedent. Recorded as a follow-up for @ToddHebebrand rather than scope-crept into a bug fix.

**Verification of the tests themselves (the point that mattered most here):** pr-test-analyzer did not take the suites on trust — it **mutation-tested** them. Deleting `targetableDevice` from *just* the `device_group` branch reddened exactly those two tests and left the other 17 green; reverting the readiness query to its pre-fix bare-count shape reddened exactly 3 of the 4 new metrics tests, with the threshold-control test correctly staying green. Both suites assert on real SQL — `PgDialect().sqlToQuery()` compiled text in `metrics.test.ts`, and the literal predicate the resolver handed Drizzle in the resolver suite — not on hand-shaped mock fixtures. Both also confirmed the `NOT(...)`-over-nullable trap does not apply: `devices.status` and `devices.isEphemeral` are both `.notNull()`.

**Branch count independently confirmed:** code-reviewer checked the `config_assignment_level` enum (`configurationPolicies.ts:51-57`) — exactly five values, exactly five switch cases, all five now carrying the predicate.

**Tests (local, single-fork `--pool=threads --maxWorkers=2`):** 68 passed (featureConfigResolver ×3) · 61 passed (metrics.test.ts) · 51 passed (backup routes) · 38 passed (backup workers/routes). `eslint` clean on all four changed files.

**CI as of this comment (run 33338745621, head ba23ce95b):**
- Green on the previous run at da0c9fcb8: **Type Check**, **Test API**, **Integration Tests all 4 shards**, Lint, Smoke Test.
- **Check Migrations** and **Trivy Image Scan** — red repo-wide, not attributable (this PR touches no migrations and no dependencies; #4227 tracks the first, #4236 fixes it).
- **Test Web** — failed once on the first run (job 99327866556). **Not attributable to this diff, which changes zero web files** (`apps/api` only, 4 files). The failure was `src/components/billing/InvoiceWorkspace.test.tsx > does NOT fire a queued Issue when the pending save fails` — `TestingLibraryElementError: Unable to find an element by: [data-testid="invoice-issue-saving-hint"]`, 1 failed / 6872 passed.

  **This is the known flake tracked in #3277** ("[Flaky] InvoiceWorkspace queued-Issue tests fail intermittently in CI (Test Web)"), previously filed and closed four more times as #4033, #3980, #3219 and #2925 — same test, same `waitFor` timeout on that hint under CI load. That hint element only exists while a save is in flight, which is why it loses the race on a loaded runner. **Test Web** is green on `main` at this PR's exact base commit 23001934 (job 99290200193) and on the three most recent `main` runs. Nothing here to fix on this PR; #3277 remains the place for it.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing either issue — both stay open and assigned.

PR #4277, opened 2026-08-30
fix(extensions): lazy BUILTINS manifest resolution so a disabled built-in cannot fail boot (#3470)

### LINKED ISSUE #3470: extensions: lazy BUILTINS loading — disabled deployment should tolerate a missing/corrupt built-in manifest
Follow-up from #3455 (workspace ee-merge), from the post-review hardening pass.

`apps/api/src/extensions/builtinRegistry.ts` parses `ee/workspace/manifest.json` at **import time**. A deployment with `BREEZE_WORKSPACE_ENABLED` unset therefore still requires the manifest to exist and parse — a missing or corrupt manifest fails boot even though the built-in is disabled and nothing else about it is needed.

Make `BUILTINS` resolution lazy so the disabled path never touches the manifest: a deployment that has opted out should boot even if `ee/workspace` files are absent or damaged (e.g. a slimmed custom image). Keep the current fail-hard behavior whenever the flag is on.

ToddHebebrand:
Closes #3470

## Problem

`apps/api/src/extensions/builtinRegistry.ts` resolved `ee/workspace/manifest.json` from disk **at module-import time** — the `BUILTINS` object literal called `loadBuiltinManifest('ee/workspace')`, and `BUILTIN_EXTENSION_NAMES` was a module-scope `Set` derived from `builtin.manifest.name`. `apps/api/src/index.ts` statically imports `./extensions/builtinExtensions` (which imports the registry), so a deployment with `BREEZE_WORKSPACE_ENABLED` unset **still** required the manifest to exist and parse. A slimmed custom image that drops `ee/workspace` failed boot for a built-in it had explicitly opted out of.

## What changed

**1. Lazy, memoised manifest resolution** (`builtinRegistry.ts`)

New exported `defineBuiltin()` builds a registry entry whose `manifest` is a **memoised lazy getter**. Both outcomes are cached and neither re-reads disk: success returns the same manifest object forever; failure rethrows the *identical* error instance, so a retrying caller can't hammer the disk and the operator sees one stable diagnostic.

`BuiltinExtension` gains a static `name` field — the identity the disabled path can use when the manifest is unavailable. Resolution enforces `manifest.name === spec.name` so the registry entry and the shipped manifest can't silently disagree. `BUILTIN_EXTENSION_NAMES` now derives from that static field, so computing it touches no filesystem either.

**2. A disabled built-in tolerates an unreadable manifest — but only when it can prove it's safe** (`builtinExtensions.ts`)

Laziness alone does *not* deliver the issue's promise: `skipDisabledBuiltin` also needs the manifest, to publish a tenancy declaration narrowed to whichever of the built-in's tables actually exist. That publication is what keeps tables created by an *earlier enabled boot* inside org-deletion cascades and the GDPR tenant-export path. Skipping it unconditionally would be a silent data-erasure hole.

So the disabled path now resolves the manifest inside a `try`, and on failure consults a new `builtinEverMigrated(name)` port — "has this built-in ever applied even one migration on this database?", answered from the core `breeze_migrations` ledger via the namespaced `'<name>/%'` filenames extension migrations are recorded under:

- **No ledger row** → the built-in never ran here, so it can own no tables and there is no tenancy to declare. One structured `builtin_extension_manifest_unavailable` warning, boot continues. *This is the slimmed-image case the issue is about.*
- **A ledger row exists** → fail hard, with `cause` set to the manifest error and a message naming the cascade/export consequence and the remedy.
- **The probe itself throws** → fail hard. Never proceed on an unknown answer.

The gate is sound because each extension migration's DDL and its ledger insert **commit in the same transaction** (`migrator.ts`), so "no ledger row" really does mean "created no tables" — including for a partially-applied multi-file run, which leaves both the committed tables and their rows.

The **enabled** path is untouched and still fails hard: it reads `builtin.manifest` directly, with no ledger check and no catch.

## Does laziness move a failure from boot time to first-request time?

**No — verified, not assumed.** `loadBuiltinExtensions` is `await`ed at `apps/api/src/index.ts:1599`, inside `bootstrap()`, **before** `runStartupChecks()` (1610) and before `serve({...})` (1708). `bootstrap()`'s rejection handler logs `[CRITICAL] API startup failed` and `process.exit(1)`. A manifest failure therefore still aborts the process before the HTTP listener ever exists — it cannot surface as a `/ready`, `/health`, or request-time error.

What *does* move: the failure now surfaces from the extension-loading phase rather than from ESM evaluation, i.e. after core DB init and the first log lines instead of before them. That is strictly better diagnostics, and the `loadBuiltinManifest` doc comment that claimed the opposite has been corrected. `worker.ts` dynamically imports the pipeline and is fatal through its own outer catch either way.

No production code reads `BUILTIN_EXTENSION_NAMES` or `builtinTenancyDeclarations()`; the only production consumer of `BUILTINS` is `loadBuiltinExtensions` via `ports.builtins`. Verified by `git grep`.

## Stale comments corrected

`loader.ts` was deleted (#3488), but three comments in `builtinRegistry.ts` still described it — the file header's import-cycle rationale, and `builtinTenancyDeclarations`' "this exists for the legacy source loader's sweep". Both are now accurate; the accessor is documented as retained for the tenancy/tenant-export contract tests. The `loadBuiltinManifest` "runs at IMPORT time" comment is corrected for the new timing.

## Tests

Red-first: the import-time test was written and confirmed **failing** against the eager version (it recorded a real read of `.../ee/workspace/manifest.json` during module evaluation) before any implementation existed.

- `builtinRegistry.test.ts` — no manifest read at module import; resolution at first access, not at definition; success memoised (the fixture tree is deleted between the two reads); failure memoised to the identical error instance; static/manifest name mismatch rejected.
- `builtinExtensions.test.ts` — the four disabled-path branches (ledger clean → boots and publishes nothing; ledger dirty → rejects with the manifest error as `cause`; probe throws → rejects, doesn't swallow; **enabled** → manifest error propagates with the ledger never consulted), plus a drift pin that the static `name` equals the real shipped `manifest.name`.
- `builtinTableProbe.integration.test.ts` — the real ledger SQL against live Postgres, including that `workspace/%` does not match a `workspace-extra/…` lookalike.

## Review round

`/pr-review-toolkit:review-pr` ran four specialists (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer) plus an independent docs sweep. code-reviewer found nothing. The other three raised findings that were addressed in `52c54da`:

- The continue-on-inference branch logged at `console.warn`, level-equal to the routine "this built-in is switched off" line, despite being the only branch here that proceeds without proof. Raised to `console.error`, and the payload now names the assumption it rests on.
- "The ledger **proves** no tables exist" was overstated: it is an inference about intact bookkeeping, not a survey of the schema, and it does not hold on a database whose `breeze_migrations` rows were dropped or restored separately from their tables. The port doc, `skipDisabledBuiltin`'s doc and the log line now say so. A suggested `<name>_%` table-prefix fallback sweep was **declined** — table naming is convention, not contract, so it would manufacture false confidence.
- `BuiltinExtension.name` is now documented as immutable once migrations have shipped under it (ledger rows are namespaced by it, so a bare rename would make the probe answer "never ran here" about a database that has the tables), and the deliberate never-retry of a cached manifest failure is documented.
- `builtinTenancyDeclarations`' doc contradicted itself after the first rewrite — one paragraph said the accessor has no production caller while the next still described "the sweep" in the present tense. Both halves now agree.
- The static-name drift pin asserted `name === manifest.name`, which the getter already guarantees, so it read as discriminating when it structurally could not fail. Rewritten as `not.toThrow()` with a comment saying what it actually contributes: it is the only place that forces resolution of the real `ee/workspace/manifest.json`.

Two findings were noted and **not** actioned, to keep the diff to the issue: `skipDisabledBuiltin`'s pre-existing doc and its `existingDeclaredTables` catch string both describe `assertNoUnaccountedPublicTables` as an active boot-time gate, but it has no production caller (only `extensionTenancyRls.integration.test.ts`). Both sentences predate this PR. Three `docs/superpowers/` plan/spec files also still describe the eager shape; those are point-in-time execution records for #3455, not living docs.

## Verification

- `vitest run src/extensions/ src/worker.boot.test.ts src/services/workerEntrypointClosure.contract.test.ts` — green (extensions: 57 passed across the two directly-affected suites).
- `pnpm exec eslint src/extensions` — clean.
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit --project apps/api/tsconfig.json` (the exact CI command) — exit 0. CI's own **Type Check** job also passed.

`aiToolNames.test.ts` times out on this machine under concurrent load; a **control run with these changes stashed on pristine `origin/main` failed it worse (4/5 vs 2/5)**, so it is host load, not this PR.

`Check Migrations` and `Trivy Image Scan` are red repo-wide right now and are not attributable to this change: `Check Migrations` fails identically on `origin/main` at this branch's exact base SHA (`23001934d`), and this PR adds no migration files at all. PR #4236 (still open) is the fix.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, plus an independent docs/prose sweep. One round.

**Findings:** 8 raised → 5 addressed in `52c54da`, 3 declined with reasons; 0 outstanding.

- **code-reviewer — 0 findings.** Independently verified the ledger-atomicity claim against `migrator.ts`, confirmed no `{...builtin}` / `Object.keys(builtin)` / `JSON.stringify(builtin)` anywhere in the repo (which would eagerly trip the new getter), and confirmed the cascade/export/RLS registration lists correctly need no entry — this PR adds no tables, columns or migrations.
- **silent-failure-hunter — 4.** The continue-on-inference branch logged at `console.warn`, level-equal to the routine skip line, despite being the only branch here that proceeds without proof → raised to `console.error` with the assumption named in the payload. "The ledger *proves* no tables exist" was overstated → reworded across the port doc, `skipDisabledBuiltin`'s doc and the log line to state the intact-ledger assumption explicitly. Added the invariant that `BuiltinExtension.name` is immutable once migrations have shipped under it, and documented the deliberate never-retry of a cached manifest failure. Its two "correctly loud" verdicts (the `everMigrated === true` throw and the probe-throws rethrow, both preserving `cause`) were confirmed, not assumed.
- **pr-test-analyzer — 1 actioned.** The static-name drift pin asserted `name === manifest.name`, which `defineBuiltin`'s getter already guarantees — it could never observe a mismatch, so it read as discriminating when it structurally was not. Rewritten as `not.toThrow()` with a comment stating what it actually contributes. Its three LOW items (a multi-built-in loop case, the `DATABASE_URL` guard, worker mode) were declined as low-value; it confirmed by control-flow reading that the disabled path is mode-agnostic, so no worker-mode test is owed.
- **comment-analyzer — 1 actioned.** `builtinTenancyDeclarations`' doc contradicted itself after the first rewrite — one paragraph declared the accessor test-only while the next still described "the sweep" in the present tense. Both halves now agree.

**Declined (with reason):** a `<name>_%` table-prefix fallback sweep for the manifest-unavailable path — table naming is convention, not contract, so it would manufacture false confidence rather than remove the caveat. A Sentry hook on the skip branch — this is a leaf boot module with deliberately no such dependency. And two pre-existing prose inaccuracies (`skipDisabledBuiltin`'s doc and its `existingDeclaredTables` catch string both describe `assertNoUnaccountedPublicTables` as an active boot-time gate when it has no production caller); both predate this PR, so they are left for a follow-up rather than widened into this diff.

**Red-first:** the import-time test was written and confirmed **failing** against the eager version — it recorded a real read of `…/ee/workspace/manifest.json` during module evaluation — before any implementation existed.

**Tests:** after the review fixes, `vitest run src/extensions/` → **16 files, 231 passed**; the two directly-affected suites are 57 of those. `pnpm exec eslint src/extensions` clean. Typecheck: `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit --project apps/api/tsconfig.json` (the exact CI command) exit 0 on the first commit, and CI's **Type Check** job passed on the review-fix commit `52c54da`.

**CI at time of writing:** 40 pass, 2 fail, rest in flight. Both failures are the known repo-wide reds and neither is attributable — `Check Migrations` fails identically on `origin/main` at this branch's exact base SHA (`23001934d`) and this PR adds no migration files at all (PR #4236, still open, is the fix); `Trivy Image Scan` is red repo-wide. Still running when this was posted: `Test API`, `Test Web`, the four `Integration Tests` shards and `Smoke Test` — the integration shards are the ones that exercise the new ledger SQL against real Postgres, so check them before merging.

**Status:** review-clean, awaiting maintainer merge. Not merged and issue #3470 not closed — both are the maintainer's call; the issue stays open and assigned.

PR #4281, opened 2026-08-30
feat: Ticket push categories — API (#3206 W07-A)

### LINKED ISSUE #3901: W07: Push categories for ticket assignment and SLA breach
Wave W07 of #3206 — Push categories for ticket assignment and SLA breach
Branch: `feature/3206-mobile-ticketing-time-entry/wave-3901-api`

ToddHebebrand:
`Refs #3901` — this PR ships the **backend half** of W07 only: the `ticket_push_preferences` table and its RLS, the preference endpoints, the push-sender generalisation, and the worker fan-out with dedupe.

**The mobile half has not shipped.** Tap routing (`PushTapRouter`, `navigationRef`), the Settings-sheet controls, the mobile API client/slice, and the docs updates (plan Tasks 13–18) are all still to come in a stacked PR, and **that** PR closes #3901. Nothing user-visible on the phone changes from this PR alone: with no mobile build that registers for the `tickets` category, the fan-out simply pushes to whatever APNs tokens already exist.

Plan: `docs/superpowers/plans/mobile/2026-08-29-ticket-push-categories.md` (on `docs/3206-w06-w08-specs-plans`, PR #4258). Tasks 2–12.

## What a technician gets

An iOS push when a ticket is assigned to them, and when a ticket they own — or, by opt-in, any ticket in their partner — breaches its response or resolution SLA. Preferences are per-user: `assignedEnabled` (default true) and `slaScope` of `off` / `owned` (default) / `any`.

## Tasks completed

An earlier session had already landed Tasks 2–10 as nine commits and left Task 11 uncommitted mid-flight. This run audited all of it (including mutation-testing the controls) rather than trusting it, then finished the wave.

**Already built by the earlier run (audited, kept):**

| Task | What |
|---|---|
| 2 | `packages/shared/.../ticketPushPreferences.ts` — Zod schemas, defaults, total `resolveTicketPushPrefs` |
| 3 | Migration `2026-09-25-ticket-push-preferences.sql`, Drizzle schema, RLS registration + behavioural RLS suite |
| 4 | `eventId` stamped on the ticket-event envelope |
| 5 | `isInQuietHours` extracted to `services/quietHours.ts` |
| 6 | APNs `thread-id` / `category`, emitted only when set |
| 7 | `PushSpec`, `dispatchPushToTokens`, `buildTicketPush`; approval dispatch became a wrapper |
| 8 | `GET`/`PATCH /api/v1/users/me/ticket-push-preferences` |
| 9 | Widened the `/users` router's self-service exemption so the route is reachable |
| 10 | `services/ticketPush.ts` — recipient discovery, authorisation, per-user push collection |

**Completed in this run:**

| Task | What |
|---|---|
| 11 | Worker fan-out: `createNotification({ dedupeKey })` as the idempotency anchor, per-recipient re-authorisation, push collection inside the system context, `dispatchPushToTokens` **after** it exits (#1105). Committed the previous run's work after fixing three type errors and two dead schema imports it had left. |
| 12 | `ticketPushFanout.integration.test.ts` — six real-Postgres tests: cross-partner and cross-org isolation with a no-permission negative control, an isolated `listAnySlaSubscribers` partner-filter assertion, dedupe replay, A→B→A reassignment, and cap truncation. |
| — | `ticketEventsContract.test.ts` updated: the producer→consumer seam test asserted the old direct `db.insert` into `user_notifications`. Not named by the plan (it lives under `services/`), and it is a **required Test API** job, so it would have reddened CI. |

Tasks 13–18 (mobile) and Task 19 Step 2 (manual device checklist) are **out of scope** for this PR. The device checklist has **not been run** — there is no mobile build for this wave yet.

## Tenancy

`ticket_push_preferences` is **Shape 6** (user-id scoped): `user_id` PK, no `org_id`, no `partner_id`, no `device_id`, not append-only. It is a personal preference in the same category as `mobile_devices` — this sentence is the CLAUDE.md-required justification for it not being a Partner-Wide-First config table.

RLS is `ENABLE` + `FORCE` + one policy in the creating migration: `breeze_current_scope() = 'system' OR user_id = breeze_current_user_id()`, on both `USING` and `WITH CHECK`. The `system` branch is required — the notify worker reads this table inside `withSystemDbAccessContext`.

### Registration lists

| List | File | Edited? |
|---|---|---|
| `USER_ID_SCOPED_TABLES` | `apps/api/src/__tests__/integration/rls-coverage.integration.test.ts` | **YES** — the only entry this table needs |
| `CORE_ORG_CASCADE_DELETE_ORDER` | `services/tenantCascade.ts` | no — no `org_id` column |
| `CORE_TENANT_EXPORT_POLICY` | `services/tenantExportPolicyRegistry.ts` | no — not an org-cascade table, and **no existing org-cascade table gains a column in this wave** |
| `CORE_DEVICE_CASCADE_DELETE_TABLES` / `CORE_DEVICE_ORG_DENORMALIZED_TABLES` | `routes/devices/core.ts` | no — no `device_id` column |
| `AUDIT_ADMIN_REQUIRED_TABLES` | `services/tenantCascade.ts` | no — not append-only |
| `orgMergeRegistry` | `services/orgMergeRegistry.ts` | no — verified: the registry's required set is `getOrgCascadeDeleteOrder()` + a hardcoded `EXTRA_REQUIRED`; adding this table would fail its "no policy names an unrequired table" assertion |

Proved by grep, not judgement: `grep -rn 'ticket_push_preferences' apps/api/src/services/tenantCascade.ts` is empty, and `mobile_devices` — the closest existing analogue, also user-PK — appears in none of these lists either. All six suites were run (below).

## The two risky surfaces, and how they are held

**Task 9 widened an authorisation predicate on `userRoutes.use('*')`, so its blast radius is every route on the `/users` router.** The exemption went from `/\/me(\/avatar)?$/` to `/\/me(\/avatar|\/ticket-push-preferences)?$/` — an allowlist, deliberately not `/\/me(\/.*)?$/`, which would auto-exempt every future `/me/*` route. It newly exempts exactly one path. `users.test.ts` covers both new exemptions plus a **control** that a `'selected'` partner-admin is still 403'd on partner-wide user management, so widening the regex too far cannot pass silently. Both handlers derive the subject from `auth.user.id`; the PATCH schema is `.strict()`, so a smuggled `userId` is a 400.

**`services/ticketPush.ts` runs under `withSystemDbAccessContext` with RLS bypassed — it is the only tenant boundary for pushes.** Three independent layers: the partner filter compiled into `anySlaSubscribersQuery`, `assertSamePartner` (warn + `captureException` on mismatch), and `isAuthorisedForTicket` (`getUserPermissions` → `tickets:read` + `canAccessOrg`). Active status is checked too.

Every control was **mutation-tested**, and this found something worth reporting. The plan's mutation table expected "`assertSamePartner` returns true" and "drop the SQL partner filter" each to redden the end-to-end fan-out test. **Neither does on its own** — the three layers are individually redundant, so removing any one leaves the fan-out correct. Rather than accept an unpinned clause, the test file gained a direct `listAnySlaSubscribers` assertion that reddens on the dropped filter (verified), and `assertSamePartner` is pinned by its own unit tests plus the worker's foreign-partner case (verified: mutating it reddens three tests). Confirmed-discriminating mutations:

| Mutation | Test that went RED |
|---|---|
| `isAuthorisedForTicket` returns `true` unconditionally | fan-out isolation test — `anyB` and `noPerm` appear |
| drop `eq(users.partnerId, partnerId)` from `anySlaSubscribersQuery` | `listAnySlaSubscribers never returns a subscriber from another partner` |
| `createNotification` called without `dedupeKey` | integration replay test (two owner rows) + two worker unit tests |
| `assertSamePartner` returns `true` unconditionally | 3 unit tests across `ticketPush.test.ts` and `ticketNotifyWorker.test.ts` |
| push dispatch moved inside the system context | 4 worker unit tests, incl. the explicit ordering assertion |
| `truncated` forced to `false` | cap-truncation integration test |

**#1105:** all network I/O happens after the DB context exits. The push loop sits deliberately **before** the email early-return, because an `any` SLA subscriber produces zero email payloads.

Run against a live per-worktree Postgres (`pnpm test-stack up`).

| Suite | Outcome | Note |
|---|---|---|
| `@breeze/shared` unit | **passed** | 84 files, 2208 tests |
| `@breeze/api` unit (full) | **passed** | after fixing `ticketEventsContract`; see "pre-existing" below |
| `pnpm lint` | **passed** | 6/6 tasks |
| `tsc --noEmit` (api) | **passed** | needs `NODE_OPTIONS=--max-old-space-size=8192` locally — OOMs at the default heap |
| `pnpm db:check-drift` | **passed** | 595 migration files match the ledger |
| `test:rls` (`vitest.config.rls.ts`) | **passed** | 26 passed, 5 skipped |
| `rls-coverage.integration.test.ts` | **passed** | 75 tests — has its own runner (`vitest.config.rls-coverage.ts`); it is *excluded* from the integration config, so passing it as a path there silently runs nothing |
| `ticketPushPreferencesRls.integration.test.ts` | **passed** | 4 tests |
| `ticketPushFanout.integration.test.ts` | **passed** | 6 tests, none skipped |
| `userNotificationsRls.integration.test.ts` | **passed** | 12 tests |
| `tenantCascade.integration.test.ts` + `tenant-export-policy.integration.test.ts` | **passed** | 7 tests — no diff expected, none needed |
| `tenantExportErasureRoundtrip.integration.test.ts` + `orgMergeRegistry.integration.test.ts` | **passed** | 13 tests |

**Pre-existing local failure, not from this PR:** `src/db/offsetPaginationOrdering.test.ts` times out (5s `testTimeout`, ~80s of repo-wide AST scanning on this machine). Verified by running the same file on a checkout that does not contain these changes — it fails identically there. `tenantVariables.test.ts` and `workerEntrypointClosure.contract.test.ts` each failed once under `--maxWorkers=2` contention and pass in isolation.

## Deviations from the plan

1. **`getUserPushTargets` re-selects devices** instead of reusing `getUserPushTokens`, because the latter drops `quiet_hours`. (Anticipated by the plan.)
2. **Task 12's cap test was rewritten.** The plan seeded `ANY_SUBSCRIBER_CAP + 5` users, each with a role and a permission grant — ~2000 serial statements, which blew the 30s integration timeout twice. The plan's own fallback applies: `listAnySlaSubscribers`/`anySlaSubscribersQuery` now take an optional `cap` defaulting to `ANY_SUBSCRIBER_CAP`. The worker never passes it, production is unchanged, and `ticketPush.sql.test.ts` still pins `ANY_SUBSCRIBER_CAP + 1` into the compiled `LIMIT`. The rewritten tests prove truncation, the lowest-user-id ordering that makes truncation deterministic, and that the owner is reached through the assignee branch rather than the capped list.
3. **Task 12's mutation table was wrong about two of its four rows** (see above). Fixed by adding a control, not by relaxing an expectation.
4. **`const dispatch = vi.fn(...)` in the plan's Task 12 snippet cannot work** — `vi.mock` factories hoist above top-level bindings, so it dies at import with `Cannot access 'dispatch' before initialization`. Uses `vi.hoisted`.
5. **`ticketEventsContract.test.ts` needed updating** and the plan does not mention it. It is in the required **Test API** job.
6. **Branch prefix.** `feature/…` per CLAUDE.md's Feature Lifecycle rule, not `feat/` per `AGENTS.md:309` — the same prefix waves W01–W03 of this feature already shipped under. Flagging, not "fixing".
7. **Migration date.** `2026-09-25`, not the plan's `2026-09-23`: `2026-09-23` and `2026-09-24-a/-b` shipped since the plan was written. Re-verified at push time that it still sorts after every migration on `origin/main` (`23001934`).
8. `slaScope='off'` suppresses the owner's **push** only; their in-app row and email are still written (spec D6). Resolved in the plan, restated here because it is the kind of thing a reviewer will read as a bug.

## CI

Two checks are **already red on `origin/main` itself** (`355b37465`) and are unrelated to this PR — do not chase them:
- **Check Migrations** — step "Shipped-migration immutability guard"
- **Trivy Image Scan**

This PR targets `main`, so CI runs normally, including the blocking integration shards. No hand-dispatch needed.

### DIFF
--- a/apps/api/migrations/2026-09-25-ticket-push-preferences.sql
+++ b/apps/api/migrations/2026-09-25-ticket-push-preferences.sql
@@ -1,0 +1,40 @@
+-- apps/api/migrations/2026-09-25-ticket-push-preferences.sql
+-- W07 (#3901): per-user mobile push preferences for ticket events.
+-- Shape 6 (user-scoped). No org_id/partner_id by design: a personal preference,
+-- same category as mobile_devices — NOT a Partner-Wide-First config table.
+-- Consequently it appears in NO cascade list (no org_id, no device_id) and in
+-- no export-policy entry; USER_ID_SCOPED_TABLES is the only registration.
+-- Idempotent; no inner BEGIN/COMMIT (autoMigrate wraps each file).
+
+DO $$ BEGIN
+  CREATE TYPE ticket_sla_push_scope AS ENUM ('off', 'owned', 'any');
+EXCEPTION WHEN duplicate_object THEN NULL; END $$;
+
+CREATE TABLE IF NOT EXISTS ticket_push_preferences (
+  user_id          uuid PRIMARY KEY REFERENCES users(id) ON DELETE CASCADE,
+  assigned_enabled boolean NOT NULL DEFAULT true,
+  sla_scope        ticket_sla_push_scope NOT NULL DEFAULT 'owned',
+  created_at       timestamptz NOT NULL DEFAULT now(),
+  updated_at       timestamptz NOT NULL DEFAULT now()
+);
+
+-- The 'any' fan-out query (services/ticketPush.ts) filters on sla_scope = 'any'
+-- then joins users on partner_id; keep the opted-in set cheap to enumerate.
+CREATE INDEX IF NOT EXISTS ticket_push_preferences_sla_any_idx
+  ON ticket_push_preferences (user_id) WHERE sla_scope = 'any';
+
+ALTER TABLE ticket_push_preferences ENABLE ROW LEVEL SECURITY;
+ALTER TABLE ticket_push_preferences FORCE ROW LEVEL SECURITY;
+
+-- Policy spelling mirrors user_notifications_user_isolation
+-- (2026-09-04-ai-agent-notifications.sql). The `system` branch is required:
+-- the ticket notify worker reads this table inside withSystemDbAccessContext.
+-- No partner/org admin branch is ORed in — nobody but the user and system
+-- jobs needs a push preference.
+DROP POLICY IF EXISTS ticket_push_preferences_user_isolation ON ticket_push_preferences;
+CREATE POLICY ticket_push_preferences_user_isolation ON ticket_push_preferences
+  FOR ALL
+  USING      (public.breeze_current_scope() = 'system' OR user_id = public.breeze_current_user_id())
+  WITH CHECK (public.breeze_current_scope() = 'system' OR user_id = public.breeze_current_user_id());
+
+GRANT SELECT, INSERT, UPDATE, DELETE ON ticket_push_preferences TO breeze_app;
--- a/apps/api/src/__tests__/integration/rls-coverage.integration.test.ts
+++ b/apps/api/src/__tests__/integration/rls-coverage.integration.test.ts
@@ -660,6 +660,11 @@ const USER_ID_SCOPED_TABLES: ReadonlySet<string> = new Set<string>([
   'user_sso_identities',
   'push_notifications',
   'mobile_devices',
+  // ticket_push_preferences: W07 (#3901) per-user ticket push preferences.
+  // Pure Shape 6 — user_id PK, no org/partner axis. Behavioural proof is
+  // ticketPushPreferencesRls.integration.test.ts; this entry only pins that
+  // the policy references breeze_current_user_id.
+  'ticket_push_preferences',
   // ticket_comments: Shape 6 on the author axis, PLUS an extra permissive
   // SELECT policy (breeze_ticket_parent_select, 2026-06-10-a migration)
   // that ORs in visibility when the parent ticket is org-accessible —
--- a/apps/api/src/__tests__/integration/ticketPushFanout.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ticketPushFanout.integration.test.ts
@@ -276,6 +276,52 @@ describe('ticket push fan-out — tenant boundary', () => {
     expect(allTruncated).toBe(false);
   });
 
+  /**
+   * REGRESSION (#4281 review): `getUserPushTargets`' device filter
+   * (notifications_enabled = true AND status = 'active') was pinned by NOTHING
+   * — the unit suite's chainable db mock ignores the WHERE and every fixture
+   * here took the column defaults. Tokens are deliberately left populated on
+   * the blocked/opted-out rows so this pins the QUERY, not the revoke route's
+   * token clearing.
+   */
+  runDb('a blocked device and a notifications-off device never receive a ticket push', async () => {
+    const fx = await seed();
+    await withSystemDbAccessContext(async () => {
+      await db.insert(mobileDevices).values({
+        userId: fx.owner.id,
+        deviceId: uniq('dev-blocked'),
+        platform: 'ios',
+        apnsToken: `tok-blocked-${fx.owner.id}`,
+        status: 'blocked',
+      });
+      await db.insert(mobileDevices).values({
+        userId: fx.owner.id,
+        deviceId: uniq('dev-muted'),
+        platform: 'ios',
+        apnsToken: `tok-muted-${fx.owner.id}`,
+        notificationsEnabled: false,
+      });
+    });
+
+    await handleTicketEvent({
+      type: 'ticket.assigned',
+      ticketId: fx.ticket.id,
+      orgId: fx.orgA.id,
+      partnerId: fx.p1.id,
+      actorUserId: fx.anyB.id,
+      eventId: 'e-devfilter',
+      payload: { assigneeId: fx.owner.id },
+    });
+
+    const calls = dispatch.mock.calls as unknown as Array<[{ token: string }[]]>;
+    const sent = calls.flatMap((c) => c[0].map((t) => t.token));
+    // Positive control: the active, notifications-enabled handset DID receive it,
+    // so an empty `sent` can never read as "the filter works".
+    expect(sent).toContain(`tok-${fx.owner.id}`);
+    expect(sent).not.toContain(`tok-blocked-${fx.owner.id}`);
+    expect(sent).not.toContain(`tok-muted-${fx.owner.id}`);
+  });
+
   runDb('the owner is notified from the assignee branch, never via the capped list', async () => {
     const fx = await seed();
     await handleTicketEvent({
--- a/apps/api/src/__tests__/integration/ticketPushFanout.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ticketPushFanout.integration.test.ts
@@ -1,0 +1,275 @@
+/**
+ * W07 (#3901): the push fan-out's tenant boundary, proven against real Postgres.
+ *
+ * The worker runs inside withSystemDbAccessContext (RLS bypassed), so isolation
+ * here is entirely app-layer: the partner filter compiled into
+ * anySlaSubscribersQuery plus the per-user permission re-check in
+ * isAuthorisedForTicket. Push transport is mocked (no APNs in CI); the durable
+ * observable is user_notifications.
+ *
+ * Prerequisites: a live test database (`pnpm test-stack up`, or
+ * docker compose -f docker-compose.test.yml up -d).
+ */
+import './setup';
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { eq } from 'drizzle-orm';
+
+vi.mock('../../services/apns', async (orig) => ({
+  ...(await orig<typeof import('../../services/apns')>()),
+  isApnsConfigured: () => true,
+}));
+// vi.hoisted, not a plain const: vi.mock factories are hoisted above every
+// top-level binding, so a bare `const dispatch` here dies with
+// "Cannot access 'dispatch' before initialization" the moment the worker
+// imports expoPush.
+const dispatch = vi.hoisted(() => vi.fn(async () => ({ tokensFound: 1, dispatched: 1, errors: 0 })));
+vi.mock('../../services/expoPush', async (orig) => ({
+  ...(await orig<typeof import('../../services/expoPush')>()),
+  dispatchPushToTokens: dispatch,
+}));
+vi.mock('../../services/email', () => ({ getEmailService: () => null }));
+
+import { db, withSystemDbAccessContext } from '../../db';
+import { mobileDevices, ticketPushPreferences, tickets, userNotifications } from '../../db/schema';
+import { handleTicketEvent } from '../../jobs/ticketNotifyWorker';
+import { ANY_SUBSCRIBER_CAP, listAnySlaSubscribers } from '../../services/ticketPush';
+import {
+  assignUserToOrganization,
+  assignUserToPartner,
+  createOrganization,
+  createPartner,
+  createRole,
+  createUser,
+  grantRolePermissions,
+} from './db-utils';
+
+const runDb = it.runIf(!!process.env.DATABASE_URL);
+
+let seq = 0;
+const uniq = (p: string) => `${p}-${Date.now()}-${seq++}`;
+
+/**
+ * A user who actually passes the fan-out gate: a membership AND a role that
+ * really holds `tickets:read` through the permissions catalog that
+ * getUserPermissions resolves at runtime.
+ *
+ * `createUser({ withMembership: true })` is NOT enough — the role it builds has
+ * zero permissions (db-utils createRole only inserts a `roles` row); grants live
+ * in the separate grantRolePermissions. A withMembership user is therefore
+ * filtered out of every fan-out, and the "obvious fix" (relaxing the expectation
+ * to `[]`) would be a vacuously green test. Pass `perms: []` to build the
+ * negative control: identical shape, no tickets:read.
+ */
+async function makeUser(opts: {
+  partnerId: string;
+  orgId?: string | null;
+  perms?: Array<{ resource: string; action: string }>;
+  orgAccess?: 'all' | 'selected' | 'none';
+}) {
+  const perms = opts.perms ?? [{ resource: 'tickets', action: 'read' }];
+  const user = await createUser({
+    partnerId: opts.partnerId,
+    orgId: opts.orgId ?? null,
+    email: `${uniq('fanout')}@example.com`,
+  });
+  if (opts.orgId) {
+    const role = await createRole({ scope: 'organization', orgId: opts.orgId, partnerId: opts.partnerId });
+    if (perms.length) await grantRolePermissions(role.id, perms);
+    await assignUserToOrganization(user.id, opts.orgId, role.id);
+  } else {
+    const role = await createRole({ scope: 'partner', partnerId: opts.partnerId });
+    if (perms.length) await grantRolePermissions(role.id, perms);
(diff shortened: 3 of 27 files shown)


ToddHebebrand:

Round produced **11 raw findings**; each was put to two independent refuters, and only findings **both** refuters failed to refute were treated as real. **6 confirmed** (deduplicating to **4 distinct defects** — the security and code lenses independently found the same collapse-id and same email-ordering bug), **5 refuted**.

Head after fixes: `08d7fe45f`. Every fix is pinned by a regression test **proven red against the pre-fix code before the fix was restored** — I reverted each fix in place, ran the suites, and confirmed exactly the intended tests failed (8 red) while 81 unrelated tests in the same files stayed green, so the assertions discriminate rather than merely confirm.

### Fixed

**1. (critical) SLA `apns-collapse-id` exceeded APNs' 64-byte limit — every native-APNs SLA push was rejected** — `apps/api/src/services/expoPush.ts`
`buildTicketPush` emitted `ticket:${ticketId}:sla_breached:${target}`. *[verified — measured with `Buffer.byteLength`]* with a real 36-char uuid (`tickets.id` is `uuid(...).defaultRandom()`) that is **65 bytes** for `response` and **67** for `resolution`; `assigned` at 52 bytes was fine, so the assignment half of the feature worked and only the SLA half was dead. `dispatchPushToTokens` forwarded it verbatim and `buildApnsRequest` set it as the header with no length guard. *[inferred — Apple's documented contract, not exercised against live APNs]* APNs caps the header at 64 bytes and answers 400 `BadCollapseId`. `BadCollapseId` is not in `UNREGISTERED_REASONS`, so the token was never purged and the failure surfaced as one `console.warn` — no Sentry event, no retry, no token cleanup. The technician saw the inbox row and the email but never the push, making it read as a phone-side problem.
Fix: shortened to `ticket:<uuid>:sla:<target>` (58 bytes max, and the two SLA targets provably stay distinct), plus `clampCollapseId` in `buildApnsRequest` as a **wire-layer backstop** so no future caller can lose an entire push category the same way. The clamp is deterministic (coalescing still works) and injective (distinct notifications never merge). Every existing fixture used the 3-char id `t-1`, which is exactly why the length was never exercised — the new test uses a real uuid.

**2. SLA owner email was queued before the dedupe check, so a redelivered job re-emailed** — `apps/api/src/jobs/ticketNotifyWorker.ts`
The owner's email payload was pushed onto `emails` *before* `notify()` ran `createNotification({ dedupeKey })`, and `notify` had no way to withdraw it. The SLA dedupe key carries no eventId, so it is stable across redeliveries: a stalled or killed job replayed to a suppressed row and a suppressed push but a **re-sent email**, once per redelivery. The assigned branch already did this correctly (`if (id === null) return none;` before building the email), so the two branches implemented different contracts for the same stated invariant — and the PR body's "a retry re-pushes nothing and re-emails nobody" was false on the SLA path.
Fix: `notify()` returns whether it wrote; the email is queued only on a non-replay. *[verified]* the previous integration replay test mocked `getEmailService: () => null`, so it structurally could not observe an email — the new unit test asserts `sendEmail` is not called on replay.

**3. `status !== 'active'` was a terminal gate, dropping the in-app row AND email for invited users** — `apps/api/src/jobs/ticketNotifyWorker.ts`
The check sat *before* `createNotification` and before the email payload was built, so it short-circuited the whole collector. *[verified]* `users.status` defaults to `'invited'`, and `assertAssigneeInPartner` checks only existence and partner match — no status filter — so an invited technician is assignable through both UI and API. On `main` both paths only required the user row to exist, so this was a **behaviour regression**: assign a ticket to a newly-invited tech before they accept, and the assignment became invisible. Not listed in the PR's Deviations section.
Fix: account status now gates the **push only** — a device cannot be registered without a login, so a non-active user has nothing to push to, but they still need to be told. Same restructuring applied to the SLA owner branch.

**4. `assertSamePartner(candidate, null)` made legacy null-partner tickets terminal** — `apps/api/src/jobs/ticketNotifyWorker.ts`
`assertSamePartner` returns false when the event partner is null — it did not merely skip the push, it made the whole recipient terminal, *and* fired `captureException` framing a legacy row as a forgery signal. *[verified]* `tickets.partner_id` is deliberately nullable; the creating migration says so in as many words ("old API code may still insert tickets without it during a rolling deploy"), and both emitters propagate the null verbatim. Every worker fixture hard-coded `partnerId: 'p-1'`; the only null coverage anywhere was a pure helper assertion pinning the return value, which said nothing about the notification loss it caused — a uniform fixture hiding a branch.
Fix: a missing event partner withholds the **push** (already conditional on `event.partnerId`), never the row, the email, or a clean Sentry log.

### Also folded in

Two changes that the review's own refutations were written *against* — both refuters read the working tree, which already carried them, rather than the pushed commit, so they refuted code that was not yet on the PR. They are behaviour-improving and now committed:

- **Push path split into two phases.** `admitPush` (Redis only — D8 + throttle) runs with **no DB context open**; `resolvePushJobs` does **one batched device read** in its own short system context; the APNs/Expo send happens after that closes. Previously the Redis `multi()` and an N-query-per-recipient device read ran inside the open fan-out transaction — the #1105 shape `alertWorker` was refactored away from. Pinned by an ordering assertion (`ctx:enter → ctx:exit → admit → ctx:enter → tokens → ctx:exit → dispatch`).
- **`userPushTargetsQuery` exposed as an unexecuted builder.** The device filter (`notifications_enabled = true AND status = 'active'`) is security-relevant — it is the only guard for a muted handset and for a soft-blocked one (lost-phone revocation) — and was pinned by nothing, because the unit suite's chainable db mock ignores the `WHERE`. Now asserted via `.toSQL()` in a file that deliberately does **not** mock `../db`, plus a real-DB integration case that seeds a `blocked` and a `notifications_enabled = false` device **with tokens left populated** (so it pins the query, not the revoke route's token clearing) and carries a positive control so an empty dispatch can never read as a pass.

**One defect I introduced and caught:** the two-phase refactor widened `DeviceRow.platform` to `string`, which is not assignable to `TaggedPushToken.platform` (`'ios' | 'android'`). `tsc` caught it; narrowed to the `device_platform` enum. This is why typecheck ran before the commit rather than after.

### Declined

None. All six confirmed findings were fixed in code.

### Refuted (not applied)

- **`status !== 'active'` regression (duplicate report)** — refuted only because it was raised a second time against a working tree where the fix had already been applied. The underlying defect was real and **is** fixed above; this was a stale-read, not a false positive.
- **"Redis round-trips run inside the open system DB transaction" (×2, two separate framings)** — both refuters read the already-refactored tree and correctly observed `admitPush` outside the context; one further argued the proposed remedy (hoisting permission resolution out) would be *counterproductive*, since `getUserPermissions` on a cache miss with no ambient context opens a fresh pooled transaction per recipient per axis — strictly worse pool pressure. That reasoning stands and was not applied. The genuine part of the concern is addressed by the two-phase split above.
- **"`getUserPushTargets` device filter is pinned by nothing"** — accurate about the pre-fix state, refuted because the compiled-SQL test and the real-DB case were already written. Now committed.
- **"The spec-D5 `users.status = 'active'` re-check is never exercised"** — refuted on the merits: the suggested test encoded the *opposite* contract (asserting the row and email are withheld for a non-active assignee), which is precisely the bug fixed in finding 3. The check is genuinely non-redundant — `getUserPermissions` reads membership and role tables only and never consults `users.status` — and is now covered by two named regression tests.

### Strengths

- **RLS is correct and complete in the creating migration** *[verified]* — `apps/api/migrations/2026-09-25-ticket-push-preferences.sql` has `ENABLE` + `FORCE ROW LEVEL SECURITY`, one `FOR ALL` policy with the same predicate on both `USING` and `WITH CHECK` (`breeze_current_scope() = 'system' OR user_id = breeze_current_user_id()`), and `GRANT SELECT, INSERT, UPDATE, DELETE ... TO breeze_app`. Spelling and GRANT pattern match the reference `user_notifications` policy in `2026-09-04-ai-agent-notifications.sql:51,52,95`.
- **The Shape 6 classification is right, and "no other registration list applies" holds against the code, not just the prose** *[verified — re-derived by grep, not taken on trust]* — `ticket_push_preferences` has no `org_id`, no `partner_id`, no `device_id`, no `site_id`, and is not append-only. It appears in `tenantCascade.ts` **0** times (correctly), `routes/devices/core.ts` **0**, `tenantExportPolicyRegistry.ts` **0**. `mobile_devices` and `push_notifications` — the two closest user-PK analogues — are likewise registered **only** in `USER_ID_SCOPED_TABLES`, and the new entry sits directly beside them. No existing org-cascade table gains a **column** in this wave, so the export-policy trip-wire (the one that fires on a new column, not just a new table) is genuinely not armed.
- **The orgMergeRegistry exclusion argument is correct** *[verified — re-derived]* — `orgMergeRegistry.integration.test.ts:237` builds `required = new Set([...getOrgCascadeDeleteOrder(), ...EXTRA_REQUIRED])`; `EXTRA_REQUIRED` is a hardcoded 10-name list this table is not in, and the "no policy names an unrequired table" assertion would go **red** if it were added. Adding it would break CI; leaving it out is right.
- **Org erasure is genuinely covered without a cascade-list entry** *[verified]* — the FK is `user_id uuid PRIMARY KEY REFERENCES users(id) ON DELETE CASCADE`, `users` is already in `CORE_ORG_CASCADE_DELETE_ORDER`, and Postgres RI cascade actions bypass RLS: no orphan rows, no FK abort. `ticketPushPreferencesRls.integration.test.ts` proves it *behaviourally* against real Postgres rather than by argument. The per-user removal path is a soft-disable (`neutralizeUserIfOrphaned` sets `status='disabled'`), so there is no per-user hard-delete enumeration to register either.
- **Migration hygiene is clean** *[verified — re-diffed against `origin/main` after fetching]* — `2026-09-25-ticket-push-preferences.sql` sorts after every migration on main (last is `2026-09-24-b-ai-agents-org-narrative.sql`), it is not in the closed `2026-08-06` block, it is fully idempotent (`DO $$ ... EXCEPTION WHEN duplicate_object` for the enum, `CREATE TABLE/INDEX IF NOT EXISTS`, `DROP POLICY IF EXISTS` before `CREATE POLICY`), and it has no inner `BEGIN;`/`COMMIT;`. The Drizzle partial index matches the SQL one exactly. The branch is behind main by six unrelated migrations, all dated earlier, so none of them reorder against this one.

### Contract-suite outcomes

Each count below is the **actual number of tests reported as run** — not an exit code. `rls-coverage` is excluded from `vitest.integration.config.ts` and has its own runner, so passing it as a path to the integration config silently runs **zero** tests and still prints green; it was run through `vitest.config.rls-coverage.ts` and the count confirms it executed.

| Suite | Runner | Result |
|---|---|---|
| API unit (full) | `vitest.config.ts` | **28990 passed**, 22 skipped, 0 failed (1635 files) |
| `rls-coverage` | `vitest.config.rls-coverage.ts` | **75 passed** |
| `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` + `orgMergeRegistry` | `vitest.integration.config.ts` | **20 passed** |
| `ticketPushFanout` + `ticketPushPreferencesRls` | `vitest.integration.config.ts` | **11 passed** |
| Typecheck (`tsc --noEmit`) | — | clean (one error found and fixed) |
| ESLint (changed files) | — | clean |

Integration suites ran against real Postgres; the fan-out suite was re-run with `--reporter=verbose` to confirm the new device-filter case executed rather than being skipped by the `runDb` guard.

Issue #3902, opened 2026-08-23
W08: Photo and attachment upload on ticket comments


ToddHebebrand:
Wave W08 of #3206 — Photo and attachment upload on ticket comments
Branch: `feature/3206-mobile-ticketing-time-entry/wave-3902-api`

ToddHebebrand:
Out of scope for the current plan (`docs/superpowers/plans/mobile/2026-08-23-mobile-time-entry.md`) — a separate subsystem that needs its own plan before it starts. Tracked here so it is not lost.

ToddHebebrand:
## W08 design spec + implementation plan are up

Docs PR: **https://github.com/LanternOps/breeze/pull/4258** (docs-only, no code)

- **Spec:** `docs/superpowers/specs/mobile/2026-08-29-ticket-attachments-design.md`
- **Plan:** `docs/superpowers/plans/mobile/2026-08-29-ticket-attachments.md` (26 tasks)

### Summary

W08 adds photo and file attachments to ticket comments across mobile, web and the customer portal. A new Shape-1 tenant table `ticket_attachments` holds metadata plus either an S3 key or an inline bytea blob, chosen once at upload by `isS3Configured()`, so hosted installs keep 10 MiB photos out of Postgres while self-hosters without a bucket get identical behaviour rather than a 503. Upload is two-step — a multipart `POST /tickets/:id/attachments` returns a pending row, and the existing JSON comment POST carries `attachmentIds[]` and claims those rows inside a newly-introduced comment transaction — which puts the object write in its own request where put-before-insert plus a compensating delete is natural, and keeps mobile on the single authenticated `coreRequest` transport. Bytes are always served through an authenticated API route with ETag/304 and `nosniff`, never a public or presigned URL; visibility is computed from the parent comment so a soft-deleted or internal comment's photos are neither listed nor served, an hourly slot-allocated reaper clears abandoned pending uploads, and org erasure deletes S3 objects before rows and aborts rerunnably on a storage fault so customer bytes are never left in the bucket with no row to find them by.

### PR split

[...]

PR #4282, opened 2026-08-30
feat: Ticket comment attachments — API, web, portal and docs (#3206 W08-A)

ToddHebebrand:
Part of #3902 (wave W08 of #3206). **PR A only — the API, web, portal and docs half.**

Technicians can attach photos and PDFs to a ticket comment from the web, and they render in the ticket feed and (public comments only) in the customer portal. Works identically on hosted (S3) and self-hosted (bytes inline in Postgres) installs.

> **The mobile half has NOT shipped.** Tasks 21–26 (Expo pickers, image resize, the composer chips and the `AttachmentViewer` route) need new native Expo modules and therefore a prebuild plus its own TestFlight cycle. They land as PR B, branched from `main` **after** this merges — not stacked, because a PR based on a sibling branch triggers no CI at all.

## What this PR does

Two-step upload: `POST /tickets/:id/attachments` (multipart, one file) returns a *pending* row with `comment_id NULL`; the existing `POST /tickets/:id/comments` carries `attachmentIds[]` and claims those rows **inside the comment transaction**, so a rejected attachment rolls the comment back. Bytes are only ever served from an authenticated route — no public URL, no presigned link. An hourly reaper deletes pending rows older than 24 h, and org erasure deletes the S3 objects *before* the rows.

## Task ledger

**Already built by an earlier run on this branch** (killed mid-execution by a session limit; every commit re-verified here, and three of them needed fixing — see *Audit findings*):

| # | Task |
|---|---|
| 1 | Shared limits, `TicketAttachmentMeta`, `attachmentIds` on the comment schema |
| 2 | Migration + Drizzle schema + cascade and export-policy registration |
| 3 | Org re-stamp on ticket move and device move |
| 4 | Magic-byte sniffer |
| 5 | `s3Storage` `putObjectBuffer` / `getObjectStream` / `deleteObjects` |
| 6 | `ticketAttachmentStorage` backend selection and byte lifecycle |
| 7 | `bodyLimit` carve-out |
| 8 | `POST /tickets/:id/attachments` |
| 9 | Claim inside the comment transaction |
| 10 | Content + delete routes |
| 11 | `GET /tickets/:id` returns `attachments` per comment |
| 12 | Portal read path |
| 13 | Permission/scope contract test |
| 14 | Pending-upload reaper (`'32 * * * *'`, allocated via `jobSchedule`) |

**Completed in this run:**

| # | Task |
|---|---|
| 15 | Org-erasure S3 object pre-clear — was **uncommitted and red** (4 failing tests) |
| 16 | `ticketAttachmentsRls.integration.test.ts` — 9 real-Postgres properties |
| 17 | Web: `TicketAttachmentList` + feed rendering |
| 18 | Web: composer upload, chips, and all eight locale dirs |
| 19 | Docs: three endpoint rows + limits + storage-backend section |
| 20 | This sweep and PR |

**Not in this PR:** Tasks 21–26 (mobile, PR B).

## Registration lists edited

`ticket_attachments` is Shape 1 (direct `org_id`) and needed **four** registrations, not one:

| List | File | Note |
|---|---|---|
| `CORE_ORG_CASCADE_DELETE_ORDER` | `services/tenantCascade.ts` | between `ticket_alert_links` and `ticket_email_links` (`'al' < 'at' < 'em'`), and before its FK parent `tickets` |
| `CORE_TENANT_EXPORT_POLICY` | `services/tenantExportPolicyRegistry.ts` | `data` (bytea) → `excludedOpen`; all 13 other columns `included` |
| `TICKET_ORG_DENORMALIZED_TABLES` | `services/ticketService.ts` | appended **last**, so both movers keep one lock order |
| `CUSTOM_ORG_REWRITE_TABLES` | `routes/devices/core.ts` (+ the UPDATE in `moveOrg.ts`) | rewrite placed **after** `ticket_parts`, not next to `ticket_alert_links`, to extend rather than invert the documented lock order |

Not applicable: the device cascade lists (no `device_id`), `AUDIT_ADMIN_REQUIRED_TABLES` (not append-only), and the `rls-coverage` allowlists (Shape 1 is auto-discovered).

Migration: `apps/api/migrations/2026-09-25-ticket-attachments.sql` — re-checked against `origin/main` at push time; it still sorts after the newest shipped file (`2026-09-24-b-ai-agents-org-narrative.sql`). `scripts/check-migration-naming.sh`: OK.

All run against a live Postgres in this worktree.

| Suite | Runner | Outcome |
|---|---|---|
| `rls-coverage.integration.test.ts` | `test:rls-coverage` | **75 passed** — `ticket_attachments` auto-discovered with all four `breeze_has_org_access` policies |
| `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` + the new `ticketAttachmentsRls` | `vitest run --config vitest.integration.config.ts <4 paths>` | **18 passed / 4 files** |
| `rls.integration.test.ts` | `test:rls` | **26 passed, 5 skipped** |
| `site-scope-coverage` | `test:site-scope-coverage` | **7 passed** |
| API unit — every suite this branch touches | `vitest run` | **passed** (610 in the largest batch) |
| Web — `components/tickets/`, `localeParity`, `no-silent-mutations` | `vitest run` | **367 + 79 passed** |
| Shared | `pnpm --filter @breeze/shared test` | **2203 passed**; `queryParams.test.ts` timed out once in the parallel run and passes alone in 0.7 s — machine load, untouched by this branch |
| Docs build | `pnpm --filter @breeze/docs build` | **167 pages, no errors** |
| Typecheck (api, web) | `tsc --noEmit` | **clean** |
| Lint | `pnpm lint` | **clean** (after the fix below) |
| Full integration suite (all 376 files) | — | **not run locally** — it is sharded ×4 in CI and does not fit a local session; the four suites the plan names were run individually and are listed above |

**Cross-org forge.** The plan asked for a manual `psql` transcript. The automated mirror is `ticketAttachmentsRls.integration.test.ts` → *"rejects a cross-org forge with 42501 while the same-org control succeeds"*, which performs **both** controls inside one test as `breeze_app` (`rolbypassrls = false`): the same-org INSERT returns one row, the org-B INSERT raises `42501 new row violates row-level security policy for table "ticket_attachments"`. A separate hand-typed transcript was **not** captured — the worktree's test database is tmpfs-backed and was recycled mid-session; the automated version proves the same property with the same two controls and runs on every PR.

## Audit findings on the earlier run's work

Three defects in already-committed code, all fixed here:

1. **`pnpm lint` was red.** `routes/tickets/attachments.ts` imported `zValidator` from `@hono/zod-validator`, which `no-restricted-imports` forbids.
2. **`routes/portal.test.ts` threw `attachmentRows is not iterable`.** The earlier run checked `routes/portal/tickets.test.ts` but not its sibling, whose `db.select` mock queue ran out when the detail handler gained its attachments SELECT.
3. **`attachments.permissions.test.ts` silently skipped all nine assertions.** Its `beforeAll` imports `./tickets`; the default 10 s hook timeout is not enough on a loaded runner. Raised to 60 s.

And one from the security-review round (Codex timed out on this repo, so this is a self-review finding, with a red test first):

4. **A non-ASCII filename 500'd the byte route.** A Node header value must be latin-1, and `contentDispositionFor` put the stored filename straight into the quoted-string form, so `写真.png` threw `ERR_INVALID_CHAR` — the file uploaded fine and was then permanently unreadable, on the technician *and* portal routes. Now an ASCII fallback plus an RFC 5987 `filename*` parameter. The header-injection guard is unchanged: control characters, quotes and backslashes are stripped on the way in **and** on the way out.

The rest of the security pass found nothing: the four GET visibility rungs all return a bare 404, the DELETE ownership matrix holds, the body-limit carve-out is anchored to `^/api/v1/tickets/[^/]+/attachments$` and does not widen the comments route, all five claim predicates are present, and the served `Content-Type` is always the sniffed-at-upload value.

## Deviations from the plan

- **Migration date.** Plan said `2026-09-23-`; `2026-09-23` and `2026-09-24-{a,b}` had already shipped, so the file is `2026-09-25-ticket-attachments.sql`.
- **Erasure pre-clear placement.** The plan put the S3 pre-clear *between* the `ASSOCIATED_SYSTEM_SCOPED_TABLES` loop and the cascade walk. That contradicts the step's own abort semantics — the associated loop already issues DELETEs, so a storage fault there would abort *after* rows were gone. It now runs **first**, which makes "aborted before anything was removed" literally true.
- **Composer chips are cleared on a successful send**, including a failed one. A leftover error chip would otherwise ride along on the next comment.
- **Test edits.** Three pre-existing tests were deliberately changed and each still fails on the intended cause, not a typo: the 4→5 table count in `ticketService.test.ts`, the two `onSend` arity assertions in `TicketComposer.test.tsx`, and the two exact-string `Content-Disposition` assertions.

## CI note

**"Check Migrations" (step *Shipped-migration immutability guard*) and "Trivy Image Scan" are already red on `origin/main` itself** (verified 2026-08-30 at `355b37465`). They are pre-existing and unrelated to this PR — do not chase them.

This PR targets `main`, so CI including the blocking integration shards runs automatically; no hand-dispatch is needed.


ToddHebebrand:

Four lenses (security · code · tests · tenancy-contracts) produced **14 raw findings**. Every finding was then handed to **two independent refuters** whose job was to kill it; **6 survived both** and are fixed below. Claims are labelled **verified** (I ran it or read it), **inferred** (follows from code I read but I did not execute), or **not-checked**.

All fixes are in `18ede93b3`.

---

### Fixed

**1. `ticket_attachments` was missing from `orgMergeRegistry` — the sixth registration list.** *(critical · security + code + tests lenses, 3 findings, one defect)* — **verified**

The table was registered in `CORE_ORG_CASCADE_DELETE_ORDER` (`tenantCascade.ts:392`) but had no merge policy. `getOrgMergePolicies()` is exactly `SPECIAL ∪ REPOINT_TABLES ∪ follows-parent` with **no default**, and `orgMerge.ts:866-873` throws unconditionally on a missing entry — reached in the `resolve` phase, before any row work. So once `2026-09-25-ticket-attachments.sql` landed, **every org merge would abort**, including for orgs owning zero attachments. `previewOrgMerge` (`orgMerge.ts:1109`) does `if (!policy) continue`, so the operator would have been shown a clean, lossless plan and the execute would then throw.

This is the failure class `CLAUDE.md` records as shipped-or-blocked five times, with code review catching it **0/5** and the contract tests **5/5** — and it is what reddened this PR's CI: integration shards **2, 3 and 4** plus one of the three `Test API` failures, all with the same root cause.

Fixed as a plain `repoint` in `REPOINT_TABLES`, localeCompare-correct between `ticket_alert_links` and `ticket_email_links`. Plain repoint is right: shape 1 (own `org_id`), the only unique index is the pkey on `id` so there is no dedupe key, and pending rows (`comment_id IS NULL`) repoint too because `addTicketComment`'s claim predicate matches on the ticket's `org_id`.

**Red-first (verified):** with the entry reverted, `orgMergeRegistry.integration.test.ts` fails with `expected [ 'ticket_attachments' ] to deeply equal []` (1 failed | 10 passed). With it, 11/11 pass.

**2. The pending reaper deleted S3 objects for rows it then refused to delete.** *(important · security + code lenses, 2 findings, one defect)* — **verified**

The old shape was: unlocked `SELECT` of candidates → `deleteObjectKeys(keys)` (a network round trip for up to 500 keys) → `DELETE ... AND comment_id IS NULL`. The re-check on the DELETE protects the **row** but not the **object** — the objects were already destroyed unconditionally for every id in the batch. `withSystemDbAccessContext` is READ COMMITTED (no isolation override exists anywhere in `db/index.ts`), so an `addTicketComment` claim committing during the S3 call is visible to the DELETE, which then matches zero rows.

The claim `UPDATE` in `ticketService.ts:1052-1065` has **no age predicate**, so a >24h-old pending row is fully claimable exactly when the reaper is mid-sweep. The result: a live comment renders an attachment whose bytes are gone, `openBytes` returns `{body: null}`, and `GET /tickets/:id/attachments/:attachmentId/content` serves a permanent 404. Unrecoverable, and nothing ever cleans the row up.

The claim is now **one atomic statement** — `DELETE ... WHERE id IN (SELECT ... FOR UPDATE SKIP LOCKED) RETURNING id, storage_backend, storage_key` — and objects are deleted only for rows the DELETE actually won. This is a **deliberate revision of spec D9's "objects before rows"**, documented in the module header: D9's rationale (never orphan bytes, since the row is the only index to the key) is the lesser harm. A storage fault now orphans *abandoned* bytes loudly, naming every key in the log and rethrowing, rather than destroying *live* customer data.

**3. That same S3 round trip ran inside a held `withSystemDbAccessContext` (#1105).** *(important · code lens)* — **verified**

`withSystemDbAccessContext` → `withDbAccessContext` → `baseDb.transaction(...)`, which pins a pooled connection for the whole callback; `createWorker` wrapped the entire sweep in it. `assertOutsideHeldDbContext` is wired only into the Redis/HTTP primitives, never the S3 client, so nothing warned — the tripwire's silence was not evidence of correctness. This PR's own org-erasure pre-clear (`tenantCascade.ts:757-772`) already had the right shape: read keys inside the context, close it, then delete objects. The reaper now matches: a short context around the claim, closed before the object delete, and the worker no longer wraps the sweep (wrapping would nest and the inner context would short-circuit).

*Scope note (inferred, per the refuters):* the reviewer's "loops forever" consequence is **overstated** — `DeleteObjects` is idempotent for absent keys, so a later healthy run does clear the backlog. And `idle_in_transaction_session_timeout` is **not-checked** against the production droplets; the connection-kill limb rests on the #1105 incident record, not on repo config. The connection-pinning defect itself does not depend on that timeout existing.

**Red-first for 2 and 3 (verified):** with `ticketAttachmentReaper.ts` and `portal/tickets.ts` reverted to the pre-fix sources and only the new tests applied, **5 tests fail** across the two files — including a context-depth probe asserting the claim runs at depth ≥ 1 and the object delete at depth 0. Two new real-Postgres cases in `ticketAttachmentsRls.integration.test.ts` pin the locking semantics directly (a row another transaction holds `FOR UPDATE` is skipped and its object is never touched), because a mocked db cannot express them.

**4. Portal byte route left `openBytes` unguarded.** *(suggestion, applied)* — **verified**

`routes/portal/tickets.ts:574` called `await openBytes(att)` bare. `openBytes` wraps any S3 transport fault in `AttachmentStorageError` and throws, so a blip surfaced to the **customer** as a generic 500 and was never attributed to this route in Sentry — while the technician twin (`routes/tickets/attachments.ts:326-332`) already returned a retryable 503 with `captureException`. Now mirrored, with a test.

**5. `'deletes the OBJECT before the row'` asserted an ordering the harness could not observe.** *(suggestion, applied)* — **verified**

The test pushed `'object'` from inside `deleteBytesMock` (during the request) but `'row'` from the test body *after* `await del()` resolved, so `['object','row']` held for any implementation that called `deleteBytes` at all. Both markers are now recorded from inside the request (the `'row'` marker moved into the `db.delete` stub). **Proven discriminating:** swapping the two statements in the route makes this assertion fail — along with its sibling `'does NOT delete the row when the object delete fails'`.

**6. Duplicate `data-testid` on the attachment strip.** *(suggestion, applied)* — **verified**

`TicketAttachmentList` rendered `ticket-attachments-${ticketId}` once per **comment**, so a ticket with photos on two comments emitted the same test id twice; `e2e-tests/README.md` mandates test-id-only selection, and a duplicate makes the strip unaddressable. Now keyed on the comment id (falling back to the ticket id). Nothing referenced the old id — **verified** by grep across `apps/web/src` and `e2e-tests`.

---

### Also fixed — three pre-existing CI failures this branch was already carrying

These were **not** review findings; they were red on the PR before the review round and are fixed in the same commit. All three are **verified** from the CI logs of run `33339068951` and reproduced locally.

- **`ticketEventsContract.test.ts` — `TypeError: db.transaction is not a function`.** This branch moved `addTicketComment`'s comment insert, `firstResponseAt` stamp and attachment claim into one transaction (correctly — the claim must roll back with the comment). That contract suite's `db` mock predates the change and exposes no `transaction`. The mock now hands the callback the same object, so every queue stays shared between the two handles; the suite asserts the emitted **event**, not which handle issued the write.
- **`workerEntrypointClosure.contract.test.ts` — `expected 114 to be 113`.** A hardcoded registry-count sanity check missed the new `ticketAttachmentReaper` worker. Bumped to 114.
- **…and a second failure in that file that CI never showed.** The count assertion sits in a `describe` **body**, so it threw at collection and aborted the file — hiding that `EXPECTED_NAMES` was *also* missing `ticketAttachmentReaper` (`registry losslessness` would have failed next). Both are fixed; worth noting because a describe-body assertion masks everything after it.

The PR body reports the API unit result as *"every suite this branch touches — passed (610 in the largest batch)"*. That is how all three slipped through: they are cross-cutting registry/contract suites that the branch does not "touch" by filename but does invalidate. **Verified**: the full run is 1,640 files / ~28,966 tests.

---

### Declined

- **304 short-circuit runs after the bytea is already loaded** (`routes/tickets/attachments.ts:318`). The finding is **verified and correct**: `loadAttachmentRow` selects `data` before the `If-None-Match` check, so on the `db` backend (the default whenever S3 is unconfigured — i.e. most self-hosted installs) up to 10 MB crosses the Postgres wire and lands in the Node heap before the route returns an empty 304. The comment on that check ("a 304 that still fetched from the object store is a silent egress bill") only holds for the s3 backend. **Declined here** because the fix is not behaviour-preserving-trivial: it means either splitting the byte path into a meta-only SELECT plus a second targeted fetch, or changing `openBytes`' contract to do its own `data` read. That is a real change to the hot read path and deserves its own change with its own tests, not a drive-by in a review-fix commit.
- **`onDeleteAttachment` is never passed by `TicketWorkbench`**, so `canDelete` is always false and `ticket-attachment-delete-*` never renders — leaving `DELETE /tickets/:id/attachments/:attachmentId` with no web caller. **Verified.** But this is the plan's scope, not an oversight: Task 17 specifies the control and its gate at the **component** level only (`TicketAttachmentList` with `canDelete`/`onDelete`, covered by `TicketAttachments.test.tsx`), and no task in `docs/superpowers/plans/mobile/2026-08-29-ticket-attachments.md` wires a workbench-level delete handler. Wiring one needs a `runAction` mutation plus a refetch — feature work, not a review fix. Applied the reviewer's fallback instead: a comment on the prop recording why it is unpassed, so the next reader does not re-derive this.

---

### Refuted (raised, investigated, **not** applied)

- **"`ticket_attachments` IS registered in `orgMergeRegistry`"** — this refutation was itself wrong, and instructively so. The worktree carried an **uncommitted** working-tree edit adding the entry, so a naive `grep` of the checked-out file found it at line 555 and read as "already fixed". `git show HEAD:` had zero occurrences, and `git log origin/main..HEAD -- .../orgMergeRegistry.ts` returned nothing. Three of the six confirmed findings flagged this trap explicitly. **Lesson worth keeping: verify a registration claim against `git show HEAD:` / `origin/main`, never the working tree.**
- **"Nothing discriminates that the comment INSERT is inside the transaction — the tx and db mocks share the same recorders."** The *test-quality* observation is accurate (`db.insert` and `tx.insert` resolve to the same `valuesMock`/`insertReturning`; only `tx.execute` has its own recorder), but the claimed production consequence does not follow. `db` is a Proxy resolving through the AsyncLocalStorage-stored transaction, and `authMiddleware` wraps ticket-comment routes in `withDbAccessContext` (they are not in `SELF_MANAGED_DB_CONTEXT_ROUTES`), so `db.transaction` here is a **nested** transaction — drizzle implements it as a savepoint on the same reserved connection, and a `db.insert` inside the callback is undone by `ROLLBACK TO` when the claim's row-count mismatch throws. The "photo-less comment already committed, technician double-posts" outcome cannot occur. Residual is a lower-severity coverage note, not a defect.
- **"Portal attachment-content authz ladder has no discriminating test."** True of the committed state, already remedied by this round: the hand-mirrored `portalVisible()` helper is gone, replaced by `describe('portal attachment read path (REAL route, real Postgres)')`, which mounts `routes/portal/tickets.ts` itself behind a stubbed `portalAuth` and drives the byte route inside an org RLS context. Deleting `eq(ticketComments.isPublic, true)` from the route now turns the internal-comment case green→red; deleting `isNull(ticketComments.deletedAt)` reddens the deleted case. A **positive control** (200 + byte equality) runs first, so the four 404s cannot be a broken rig. The mocked unit suite now carries a NOTE saying its three 404 cases prove only the empty-result path and pointing at the integration suite.

---

### Strengths

Migration `apps/api/migrations/2026-09-25-ticket-attachments.sql` is correct on every contract I checked (**verified**): `ENABLE` + `FORCE ROW LEVEL SECURITY` with all four DML policies on `public.breeze_has_org_access(org_id)` in the **creating** migration; idempotent throughout (`CREATE TABLE IF NOT EXISTS`, `DO $$ … EXCEPTION WHEN duplicate_object`, `CREATE INDEX IF NOT EXISTS`, `DROP POLICY IF EXISTS` before each `CREATE POLICY`); no inner `BEGIN`/`COMMIT`; `GRANT` to `breeze_app`; and the filename sorts after every shipped migration on `origin/main` (newest there is `2026-09-24-b-ai-agents-org-narrative.sql`) without touching the closed `2026-08-06` block. Re-diffed against `origin/main` after the review — nothing has landed since.

Cascade placement is right on both axes (**verified**): `ticket_attachments` sits between `ticket_alert_links` and `ticket_email_links` in `CORE_ORG_CASCADE_DELETE_ORDER` (localeCompare `'al' < 'at' < 'em'`), and precedes both of its in-list FK parents — `ticket_comments` (`'at' < 'co'`) and `tickets` — so children are deleted before parents. `tenantCascade.test.ts` pins all four of those orderings explicitly.

Export policy is complete and correctly bucketed (**verified**): all 14 physical columns are classified, `data` (bytea) is `excludedOpen` per the open-container rule, and nothing else trips `SUSPICIOUS_NAME_PARTS` — I checked `sha256` and `storage_key` against the actual list in `tenantExportPolicy.ts:35-54` (`'hash'` does not substring-match `'sha256'`; `'private_key'`/`'encryption_key'`/`'access_key'`/`'client_key'` do not match `'storage_key'`), so plain `included` is right rather than `reviewedIncluded`.

Both org-move re-stamp paths are present and shaped exactly like their siblings (**verified**): `ticketService.ts:1439` appends the table to `TICKET_ORG_DENORMALIZED_TABLES` (restamped `WHERE ticket_id = ticketId`, which also catches PENDING rows so a claim after a move still passes the `org_id = ticket.orgId` predicate), and `moveOrg.ts:343` adds the tickets-join UPDATE for the device path, registered in `CUSTOM_ORG_REWRITE_TABLES` (`devices/core.ts:219`) with a matching assertion in `moveOrg.coverage.test.ts`. Both are covered by real-Postgres cases in `ticketAttachmentsRls.integration.test.ts`.

The org-erasure object pre-clear is genuinely fail-closed (**verified**): `tenantCascade.ts:757-784` reads the s3 keys under `withSystemDbAccessContext` and deletes objects BEFORE any row anywhere, and `s3Storage.deleteObjects` throws on a non-empty `Errors` array rather than swallowing a partial failure — so a bucket fault writes the `tenant.erasure.failed` audit with `failedTable: 'ticket_attachments_objects'` and aborts with nothing deleted, leaving the erasure genuinely rerunnable. Four mocked cases plus two real-Postgres cases pin the ordering, the org+backend scoping, the zero-call db-only path and the rerunnable abort.

---

### Contract-suite outcomes (local, real Postgres unless noted)

| Suite | Runner | Outcome |
|---|---|---|
| `orgMergeRegistry.integration.test.ts` | `--config vitest.integration.config.ts <path>` | **11 passed / 1 file** (red before the fix: `missing = ['ticket_attachments']`) |
| `orgMerge` + `orgMergeCustomExecutors` integration | same | **8 passed / 2 files** (both were red in CI on the prior head) |
| `ticketAttachmentsRls.integration.test.ts` | same | **14 passed / 1 file** (was 9 — this round added 5) |
| `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` | same | **9 passed / 3 files** |
| `rls-coverage.integration.test.ts` | `test:rls-coverage` (**its own config** — it is EXCLUDED from `vitest.integration.config.ts`, so passing it as a path there silently runs ZERO tests and still prints green) | **75 passed / 1 file** |
| `rls.integration.test.ts` | `test:rls` | **26 passed, 5 skipped / 2 files** |
| `site-scope-coverage` | `test:site-scope-coverage` | **7 passed / 1 file** |
| API unit — the files this round touched | `vitest run <paths>` | **146 passed / 6 files**, plus **122 passed / 3 files** for the worker-registry contracts and **4 passed** for `ticketEventsContract` |
| Web — `components/tickets/` | `vitest run` | **250 passed / 15 files** |
| `apps/api` typecheck | `tsc --noEmit -p apps/api/tsconfig.json` | **clean** (one error found and fixed en route: `app.request` is `Response \| Promise<Response>`) |
| `apps/web` typecheck | `tsc --noEmit` | **clean** |
| Lint (all changed files, api + web) | `eslint` | **clean** |

Test counts were read from each runner's summary line rather than assumed — a suite that runs zero tests still prints green.

### CI on `18ede93b3`

**Fully green — 58 pass, 1 skipping, 0 fail** (verified via `gh pr checks 4282`; the one skip is `Main Red Alert`, a main-only job). The four previously-red jobs all pass: **Test API** (19m37s), **Integration Tests shards 1-4** (all four), and **Test Web**. `CI Success` is green.

Worth flagging: **Check Migrations** and **Trivy Image Scan** also pass on this head, despite both being red on `origin/main` at the time this round started. Nothing in this commit touches either surface, so treat that as those jobs having recovered upstream rather than as anything this PR did — **inferred**, not established.

Local full-suite figures for the record: `apps/api` **1,636 files / 29,043 tests passed** (the one failure in that run was a stale `workerEntrypointClosure` file from before the `EXPECTED_NAMES` fix, verified passing since); `apps/web` **250 passed** across `components/tickets/` plus **117** for locale parity and `no-silent-mutations`; `packages/shared` **2,204 passed / 84 files**; `apps/api/src/services/` alone **13,692 passed / 806 files**; `routes/tickets` + `routes/portal` + `jobs` **2,099 passed / 158 files**.

*Method note:* three other agents were running vitest concurrently in sibling worktrees on this machine, which timed out several whole-source-tree scan suites locally (`autoMigrate`, `partner-wide-write-coverage`, `noRawEnvNumberCoercion`, `offsetPaginationOrdering`, `tenantVariables`, `bodyLimit`). Each re-ran green in isolation — `bodyLimit` and the worker contracts verified directly. CI on `18ede93b3` is the authoritative full-suite signal.

PR #4284, opened 2026-08-31
fix(web): stop a created quote block from being invisible in the editor (#3519)

### LINKED ISSUE #3519: [Web] Quote image upload can create blocks server-side while the editor shows nothing
## Description
While authoring a quote with several image blocks (US prod, v0.105.1), four image uploads created their blocks server-side while the editor UI showed nothing: the submit button stayed latched disabled at "Upload & add image", no toast fired, and no block appeared in the list. This led to repeated re-uploads and ~10 duplicate image blocks that then needed manual cleanup. Reloading the page showed the blocks had existed all along.

This is the silent-mutation class the `runAction` convention exists to prevent — the mutation succeeded but the user was told nothing.

## Notes from a first look
The current image path in `apps/web/src/components/billing/quotes/QuoteEditor.tsx` (`submitBlock`, ~1123–1170) does wrap both steps in `runAction` inside `runScoped('add-block', ...)`, so the failure is likely not a missing toast but state handling after the upload: a throw between `addBlock` succeeding and `refresh()` (e.g. in `positionNewBlock`) would leave the block created server-side, the pending flag latched, and the list stale. The block list only reflecting reality after a full reload suggests `refresh()` was never reached.

## Proposed Fix
- Ensure the pending/disabled state and list refresh happen in a `finally`-style path so a late-step failure can't strand a successfully created block invisibly.
- If any step after block creation fails, surface an error that says the block WAS created (prompting a refresh), not a generic failure that invites re-submitting.
- Regression test: simulate `positionNewBlock`/`refresh` failure after a successful `addBlock` and assert the user sees an outcome and the button unlatches.

## Affected Files
- `apps/web/src/components/billing/quotes/QuoteEditor.tsx` (primary)

Found during internal quote-authoring work.

ToddHebebrand:
Closes #3519

## The stranding mechanism (established, not assumed)

The issue's first-look note guessed at "a throw between `addBlock` succeeding and `refresh()` leaving the pending flag latched". I traced it and that is **not** what happens on `main`: `runScoped` already clears its pending key in a `finally`, and `positionNewBlock` already swallows its own reorder failure. Neither can latch the button.

The real mechanism is one link further out:

1. Add-block deliberately has **no success toast** — every branch carries the comment *"No success toast — the new section visibly appears"*. The block appearing in the canvas **is** the success signal.
2. That canvas only moves when the workspace's **quiet** refetch lands.
3. `QuoteWorkspace.fetchDetail` swallowed every quiet failure, justified in-code as *"the inline action's own runAction toast already surfaced the failure."* That rationale is sound for a failed mutation — but this is the refresh **after a success**, where `runAction` toasted nothing on purpose.

So a failed quiet refetch turned the whole interaction silent: the block existed server-side, the canvas stayed empty, and the submit button went disabled again — not because it was latched, but because the form reset cleared `imageFile` while the **uncontrolled** `<input type="file">` kept displaying the filename. Disabled button + visible filename + no toast + no block is visually identical to "still uploading". Hence the re-uploads and the ~10 duplicates.

A second, independent vector in the same function: two quiet reloads overlap routinely (a slow multi-megabyte upload finishes while an earlier edit's refetch is still open) and `fetchDetail` had no request sequencing, so the **older** response could resolve last and re-render the editor from pre-mutation data.

## The fix

- **`QuoteWorkspace.fetchDetail`** now returns whether the view actually caught up. A quiet failure still stays silent *at this layer* — a mid-edit refetch must not remount the form or flash a page-level error — because reporting it is the caller's job: only the caller knows what the user was promised. It also carries a monotonic request/apply sequence so a stale overlapping response can never call `setDetail`.
- **`QuoteEditor.finishBlockCreate`** is the tail every add-block branch (image, contract, table, callout, heading/rich-text/line-items) runs *after* the POST has already created the block. Two rules:
  1. **It never throws.** A throw escaped into `runScoped`'s catch and toasted *"Could not add the image section"* over a block that **does** exist — precisely the copy that invites a re-submit.
  2. **If the canvas does not catch up, it says so**, with copy that names the real state and discourages retrying.
  The form reset sits in a `finally` so a failed reposition can't leave the picked file and the open insert-gap looking like work still in flight.
- The **uncontrolled file input** is cleared on reset (React state alone can't clear it), so the filename no longer sits on screen beside a silently-disabled button.

### On the `runAction` convention

`QuoteEditor.tsx` is in the `no-silent-mutations` **targeted set** and is **not** in `runActionAllowlist.ts` — no allowlist entry was added or needed. Every mutation here already goes through `runAction`, and all of them still do; the guard passes unchanged. This bug was never a missing `runAction` wrapper. It was the deliberate *absence of a success toast* (a legitimate choice) resting on a delivery channel — the refetch — that could fail without telling anyone. `runAction` cannot see that: it correctly reported "the POST succeeded". The fix restores the invariant `runAction` exists to protect, one layer out.

## Both directions covered

- **Success direction:** create lands, refetch does not → the user is told the section exists.
- **Failure direction:** anything after the create fails (including a refetch that *throws*) → the user still gets the honest outcome, never *"could not add"*.
- **Pre-create failure** (guard against over-reporting): a non-2xx `addBlock` still toasts the server's error, adds nothing, says nothing about "added", and leaves the button live.
- **Happy path** (guard against toast noise): a landed refetch stays completely quiet.

## Verification

**Red-first, proven.** Both new suites were run against the *unfixed* `QuoteEditor.tsx` / `QuoteWorkspace.tsx` (sources reverted, tests kept). 3 controls failed, all for the same reason — **total silence**:

```
FAIL QuoteEditor.blockcreatestrand.test.tsx > tells the user the section was added when the list refetch never lands
     AssertionError: expected "vi.fn()" to be called at least once
FAIL QuoteEditor.blockcreatestrand.test.tsx > does not report "could not add the image section" when the create succeeded and only the resync failed
     AssertionError: expected "vi.fn()" to be called at least once
FAIL QuoteWorkspace.test.tsx > warns that the section was added when the post-create reload cannot land
     AssertionError: expected "vi.fn()" to be called with arguments: [ ObjectContaining{…} ]
Test Files  2 failed (2) | Tests  3 failed | 9 passed (12)
```

The first control also asserts `quote-blocks-empty` is still rendered *before* it asserts the toast — so "the editor shows no block" is pinned as the premise, not assumed. The 9 that pass in both states are the guards.

With the fix: `2 passed (2) | 13 passed (13)`.

**Review round 2 added a third red-first proof.** The sequence guard now has a dedicated regression test, and it discriminates precisely: deleting *only* the two guard lines (`seq < appliedSeq.current` / `appliedSeq.current = seq`) turns that one test red while the other eight workspace tests stay green.

**One assertion was removed for being vacuous.** Review suggested asserting the uncontrolled file input's `.value` is cleared. Measured: with the DOM reset deleted, `.value` is *still* `''` — jsdom never populates `value` from a programmatic `files` assignment, and a file input's value cannot be set to a non-empty string. The assertion passed either way, so it was removed rather than shipped as fake coverage. **The file-input DOM reset is therefore covered by inspection only; proving it needs a real browser (Playwright).** Labelling that rather than implying coverage.

| Check | Result |
|---|---|
| All 59 `apps/web/src/components/billing/quotes/*.test.tsx` | **59 files, 708 tests passed** (2 foreground batches, re-run after the review fixes) |
| `no-silent-mutations` + i18n guards (`localeParity`, `translationCoverage`, `keyUsage`, `terminologyQuality`, `extractionQuality`) | **passed** |
| `tsc --noEmit` (`@breeze/web`, foreground) | **exit 0, zero output** |
| `eslint` on the 4 changed `.ts(x)` files | **exit 0** |

No `.astro` files changed, so `astro check` adds nothing here.

### i18n

One new key, `quotes.editor.errors.sectionAddedListStale`, added to **all 8 locales** — `en`, `de-DE`, `es-419`, `fr-CA`, `fr-FR`, `it-IT`, `pt-BR`, `tr-TR`. Each was verified present and **genuinely translated** (no locale kept the English string); `fr-CA` and `fr-FR` intentionally share one French sentence, matching how the neighbouring `imageAddedSectionFailed` / `addSection` keys are already written in those two files. `localeParity` and `translationCoverage` pass.

## Known gap this PR does NOT close (found in review, deliberately out of scope)

**Line creation has the exact same defect.** `doAddCatalog`, `addManualLine`, and the Pax8/distributor imports all rest on the same premise the block path just stopped trusting — *"No success toast: the new row visibly appears and the totals move"* — and all still end in the fire-and-forget `refresh()` rather than the honest tail. A line POST that succeeds while its quiet refetch fails is still completely silent, and a tech on a flaky link can re-add the same line the way the reporter re-uploaded the same image.

It is out of scope here because it needs its own copy, its own 8 locales and its own tests, and folding it in would double a diff that is currently provable end to end. It is documented at the call site (`QuoteEditor.tsx`, above `doAddCatalog`) so it cannot be lost, and the `finishBlockCreate` docstring no longer overstates its own reach. **Worth filing as a follow-up.**

Relatedly: the two rules in `finishBlockCreate` only hold for branches that route through it, and nothing enforces that routing — a future sixth block type could re-inline the old tail and reopen the bug. The docstring now says so explicitly rather than claiming the class is structurally impossible.

## CI notes

- `Check Migrations` and `Trivy Image Scan` are red repo-wide right now — they fail on `main` at this PR's exact base (`23001934d`) and on unrelated PRs. #4227 tracks the first; PR #4236 fixes it. **Not attributable to this change**, which touches no migrations and no image contents.
- If `Test Web` reports `InvoiceWorkspace.test.tsx` / `invoice-issue-saving-hint`, that is the known flake #3277 (dupes #4033, #3980, #3219, #2925) — unrelated to this diff.

### DIFF
--- a/apps/web/src/components/billing/quotes/QuoteEditor.blockcreatestrand.test.tsx
+++ b/apps/web/src/components/billing/quotes/QuoteEditor.blockcreatestrand.test.tsx
@@ -1,0 +1,183 @@
+/**
+ * Regression suite for #3519 — "Quote image upload can create blocks server-side
+ * while the editor shows nothing".
+ *
+ * The stranding mechanism: an image block's ONLY success signal is the block
+ * appearing in the list ("No success toast — the image block visibly appears"),
+ * and that list only moves when the parent's *quiet* refetch lands. A quiet
+ * refetch that fails is swallowed by design, so the user saw: block created
+ * server-side, zero toast, an empty list, and a submit button that had gone
+ * disabled again because the form reset — indistinguishable from "still
+ * uploading". They re-uploaded, four times, into ~10 duplicate blocks.
+ *
+ * These tests pin BOTH directions of the post-create tail:
+ *  - success direction: the create landed but the list never resynced → the
+ *    user must still be told the section exists;
+ *  - failure direction: something after the create failed → the user must NOT
+ *    be told "could not add the section" (the copy that invites a re-submit).
+ */
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import QuoteEditor from './QuoteEditor';
+import type { QuoteDetail as QuoteDetailData } from './quoteTypes';
+import { addBlock, uploadQuoteImage } from '../../../lib/api/quotes';
+
+vi.mock('../../../stores/auth', () => ({
+  registerOrgIdProvider: vi.fn(),
+  fetchWithAuth: vi.fn().mockResolvedValue(
+    { ok: true, status: 200, statusText: 'OK', json: vi.fn().mockResolvedValue({ data: {} }) } as unknown as Response,
+  ),
+  useAuthStore: Object.assign(
+    (selector: (s: { user: { permissions: { resource: string; action: string }[] } }) => unknown) =>
+      selector({ user: { permissions: [{ resource: '*', action: '*' }] } }),
+    { getState: () => ({ tokens: null }) },
+  ),
+}));
+vi.mock('@/lib/navigation', () => ({ navigateTo: vi.fn() }));
+const showToast = vi.fn();
+vi.mock('../../shared/Toast', () => ({ showToast: (a: unknown) => showToast(a) }));
+
+vi.mock('../../../lib/api/catalog', () => ({
+  listCatalog: vi.fn().mockResolvedValue(
+    { ok: true, status: 200, statusText: 'OK', json: vi.fn().mockResolvedValue({ data: [] }) } as unknown as Response,
+  ),
+  createCatalogItem: vi.fn(),
+  polishTextRequest: vi.fn(),
+}));
+
+vi.mock('../../../lib/api/quotes', () => ({
+  addBlock: vi.fn(),
+  deleteBlock: vi.fn(),
+  updateBlock: vi.fn(),
+  addManualLine: vi.fn(),
+  addCatalogLine: vi.fn(),
+  updateLine: vi.fn(),
+  removeLine: vi.fn(),
+  moveLine: vi.fn(),
+  reorderBlocks: vi.fn(),
+  uploadQuoteImage: vi.fn(),
+  addQuoteImageFromUrl: vi.fn(),
+  quoteImageUrl: vi.fn().mockReturnValue('/quotes/q-1/images/img-1'),
+}));
+
+const okRes = (data: unknown) =>
+  ({ ok: true, status: 200, statusText: 'OK', json: vi.fn().mockResolvedValue({ data }) } as unknown as Response);
+const errRes = () =>
+  ({ ok: false, status: 502, statusText: 'Bad Gateway', json: vi.fn().mockResolvedValue({ error: 'upstream exploded' }) } as unknown as Response);
+
+const detail: QuoteDetailData = {
+  quote: {
+    id: 'q-1', quoteNumber: null, partnerId: 'p-1', orgId: 'org-1', siteId: null, status: 'draft',
+    currencyCode: 'USD', issueDate: null, expiryDate: null, subtotal: '0.00', taxRate: null,
+    taxTotal: '0.00', total: '0.00', oneTimeTotal: '0.00', monthlyRecurringTotal: '0.00',
+    annualRecurringTotal: '0.00', billToName: null, introNotes: null, terms: null,
+    termsAndConditions: null, sellerSnapshot: null, acceptedAt: null, declinedAt: null,
+    convertedAt: null, convertedInvoiceId: null, sentAt: null, viewedAt: null, createdBy: null,
+    createdAt: '2026-06-01T00:00:00Z', updatedAt: '2026-06-01T00:00:00Z',
+  },
+  blocks: [],
+  lines: [],
+};
+
+const addBlockMock = vi.mocked(addBlock);
+const uploadMock = vi.mocked(uploadQuoteImage);
+
+/** Mount the editor with a parent whose refetch behaviour the test controls,
+ *  pick the image block type, and attach a file — i.e. the exact state the
+ *  reporter was in when they hit "Upload & add image". */
+async function armImageUpload(onChanged: () => unknown) {
+  render(<QuoteEditor detail={detail} onChanged={onChanged as () => void} />);
+  await waitFor(() => expect(screen.getByTestId('quote-editor')).toBeInTheDocument());
+  fireEvent.click(screen.getByTestId('quote-add-block-type-image'));
+  const file = new File(['bytes'], 'diagram.png', { type: 'image/png' });
+  fireEvent.change(screen.getByTestId('quote-block-image-file'), { target: { files: [file] } });
+  return file;
+}
+
+const toastMessages = () =>
+  showToast.mock.calls.map((c) => (c[0] as { message: string }).message);
+
+describe('QuoteEditor — an image block created server-side is never invisible (#3519)', () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  it('tells the user the section was added when the list refetch never lands', async () => {
+    // Both mutations succeed — the block EXISTS server-side from here on.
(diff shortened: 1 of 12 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 agents in parallel: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `comment-analyzer`.

**Findings:** 11 raised → 8 acted on in `8afea6a`, 3 recorded and not acted on; 0 outstanding that block merge.

Acted on:
- **`pr-test-analyzer` (sev 8) — my scope note in the PR body was factually wrong.** I had claimed the request/apply sequence guard wasn't cheaply testable because `runScoped` serialises the only driveable overlap. `runScoped` only serialises calls sharing a *key*, and `'add-block'` is not the key a terms blur-save uses — so blurring terms then adding a section fires two genuinely overlapping GETs. Regression test added, and **verified to discriminate**: deleting only the two guard lines turns it red while the other 8 workspace tests stay green. The false claim is gone from the body.
- **`pr-test-analyzer` (sev 5) — the file-input reset had no real coverage.** The suggested fix (assert the input's `.value`) turned out to be vacuous too: with the DOM reset deleted, `.value` is *still* `''`, because jsdom never populates `value` from a programmatic `files` assignment and it can't be set to a non-empty string. Removed rather than shipped as fake coverage; the reset is now labelled inspection-only, provable in a real browser.
- **`silent-failure-hunter` (medium)** — `resync()`'s catch erased the reason a refetch threw; added the breadcrumb it was inconsistent about.
- **`comment-analyzer` + `silent-failure-hunter`** — the `finishBlockCreate` docstring claimed the bug class was "structurally impossible from here on". It isn't: nothing enforces that a new block type routes through the helper. Softened, with an explicit instruction for future branches. Also corrected "one extra GET" (per-add, not aggregate), noted the 401 exception to `fetchDetail`'s return contract, noted that the sequence guard covers the success path only, and de-duplicated a rationale that was stated three times.

Recorded, not acted on:
- **`silent-failure-hunter` (HIGH) — line creation has the same defect** (`doAddCatalog`, `addManualLine`, Pax8/distributor imports: same "the appended row IS the feedback" premise, same fire-and-forget `refresh()`). Real, and the same root cause. Out of scope here — own copy, own 8 locales, own tests — so it is documented at the call site and called out in the PR body. **Worth a follow-up issue.**
- `console.error` doesn't reach Sentry (pre-existing convention, matches `runAction.ts`).
- A `resetForm()` throw would be attributed to the resync in the toast copy (`resetForm` is a setState batch; accepted).

`code-reviewer` returned **0 findings at or above its confidence bar**.

**Tests:** all 59 `apps/web/src/components/billing/quotes/*.test.tsx` green (**708 passed**, 2 foreground batches, re-run after the review fixes); `no-silent-mutations` + the i18n guards (`localeParity`, `translationCoverage`, `keyUsage`, `terminologyQuality`, `extractionQuality`) green. `tsc --noEmit` on `@breeze/web` clean (exit 0, zero output); `eslint` on the 4 changed files clean.

**Red-first:** 3 controls proven red against the unfixed sources (all failing with *total silence* — the reported symptom), plus a 4th proven red against the guard-removed `fetchDetail`.

**CI:** `Check Migrations` and `Trivy Image Scan` are red repo-wide — they fail on `main` at this PR's base `23001934d` and on unrelated PRs (#4227 tracks the first, PR #4236 fixes it). Not attributable: this change touches no migrations and no image contents.

**Status:** review-clean, awaiting maintainer merge. Not merged, #3519 left open and assigned.

ToddHebebrand:
**Correction to the CI note.** The PR body pre-excused `Check Migrations` and `Trivy Image Scan` as red repo-wide. That was true of the first run on `9f1c4b3` (Check Migrations failed), but both **pass** on the current head `8afea6a` — so the caveat no longer applies and should not be read as an excuse for anything.

**All checks are green on `8afea6a`**, with no failures and none pending: Test Web, Test API, Type Check, Lint, all four Integration Tests shards, both Check Migrations jobs, and Trivy Image Scan.

Still not merged, and #3519 is still open — both are the maintainer's call.

Issue #4276, opened 2026-08-30
metricRollupsWorker holds a system DB context for 2s+ every 5 minutes — day-level derived rollups re-scan the whole day inside one 26-statement transaction


ToddHebebrand:
Sentry BREEZE-18 (`event_code:db_context_held_too_long`, post-#3913 tagging) shows `worker: metricRollupsWorker` as the **top hold offender: 983 events/7d** — ahead of agentAuthMiddleware (399) and the old #1105 offender alertWorker (117). Since captures are throttled per scope (5 min, `apps/api/src/db/index.ts:244-270`) and this context is unlabeled system scope, that's ~half of the entire system-scope capture budget, crowding out other system-scope holders.

## Mechanism (verified in code)

`rollup-org-range` jobs (fanned out per org every 5 min, `apps/api/src/jobs/metricRollups.ts:85-101`) wrap **all of `rollupDeviceMetricsRange`** in one `withSystemDbAccessContext` (`apps/api/src/jobs/metricRollups.ts:122-124`). Inside that single pooled-connection transaction (`apps/api/src/services/metricRollups.ts:512-558`):

- 2 reads for ML flag gating (`services/mlFeatureFlags.ts:179-193`)
- **24 sequential heavy `INSERT … SELECT … ON CONFLICT` statements**: 10 raw device-metric rollups (percentile_cont per generate_series bucket), 7 process-sample rollups (jsonb_array_elements expansion), 1 SNMP rollup (4-way join + md5 per row), 6 derived rollups.

The amplifier: `expandRangeToBucketBounds` (`services/metricRollups.ts:139-145`, applied at `:456`) widens the 15-min window to full target-bucket bounds, so the **day-level derived passes (`:540`, `:544`, `:548`) re-scan and re-upsert the org's entire day of hourly rollups on every 5-minute run** — cost grows through the day, resets at midnight UTC. Blowing the 2s `DB_CONTEXT_HELD_WARN_MS` default is the expected outcome.

This is pure DB work (no Redis/WS/HTTP inside the context), but long single-connection holds are exactly what elongates recovery when the pool degrades (#3225 storms, latest recurrence 2026-08-30).

## Fix directions

1. Bound day-level derived rollups to the buckets actually touched, or move them to a low-frequency schedule instead of the 5-minute path.
2. Split the 26-statement transaction into per-statement or per-source-table contexts — every statement is an idempotent upsert; nothing needs the single atomic transaction.
[...]

Issue #3225, opened 2026-08-07
postgres.js 3.4.9: teardown clears nextWriteTimer without nulling it — poisoned connection can never flush again (upstream patch)


ToddHebebrand:
Split out of #3214, which shipped detection only (PR #3224). This issue tracks the **actual repair** — a ~2-line fix in the driver's write scheduler.

**Root cause.** In `postgres` 3.4.9, `src/connection.js` defers small writes behind a `setImmediate`, and arms that timer only when it is currently unarmed:

```js
// src/connection.js:246
function write(x, fn) {
  chunk = chunk ? Buffer.concat([chunk, x]) : Buffer.from(x)
  if (fn || chunk.length >= 1024)
    return nextWrite(fn)
  nextWriteTimer === null && (nextWriteTimer = setImmediate(nextWrite))
  return true
}

function nextWrite(fn) {
  const x = socket.write(chunk, fn)
  nextWriteTimer !== null && clearImmediate(nextWriteTimer)
  chunk = nextWriteTimer = null      // <-- the ONLY reset of either variable
  return x
}
```

`nextWrite()` running at `:257` is the only place `nextWriteTimer` is ever set back to `null`. Both teardown paths clear the timer **without** nulling it, and neither resets `chunk`:

- `terminate()` — `src/connection.js:427`, `clearImmediate(nextWriteTimer)`
- `closed()` — `src/connection.js:440`, `clearImmediate(nextWriteTimer)`

**Consequence.** A pooled connection whose socket dies while a write is buffered ends up with a stale non-null `nextWriteTimer` and a stale `chunk`, and the connection object is then reused for reconnect. The `StartupMessage` on every subsequent reconnect is appended to `chunk` and the arming guard at `:250` sees non-null, so no flush is ever scheduled. Nothing reaches the socket, `connect_timeout` expires, the connection is destroyed and reconnects — forever. That pool slot is dead for the life of the process.

Worth noting the failure is slightly worse than "never flushes": `:248` still force-flushes once `chunk.length >= 1024`, so a sufficiently poisoned slot eventually writes a **corrupt** protocol stream with stale leading bytes rather than staying silent.

[...]

PR #4285, opened 2026-08-31
fix(api): metric rollups run one short-lived labeled DB context per statement (#4276)

ToddHebebrand:
Part of #4276 (fix directions 2 and 3; direction 1 — bounding the day-level derived ranges — stays open on the issue).

## Problem
`metricRollupsWorker` is the top `db_context_held_too_long` offender (983 events/7d, ~half the system-scope Sentry capture budget): `rollup-org-range` wrapped all of `rollupDeviceMetricsRange` in ONE system context — 24 heavy `INSERT…ON CONFLICT` upserts plus the ML feature-flag gate on one pooled connection, 2s+ per org, every 5 minutes. Long single-connection holds are exactly what elongates recovery when the pool degrades (#3225, recurred in prod 2026-08-30).

## Change
- Each statement now runs in its own `runOutsideDbContext(() => withSystemDbAccessContext(fn, label))` pair. No atomicity was lost that mattered — every statement is an idempotent upsert; a half-finished pass is re-upserted by the next 5-minute run.
- The escape is load-bearing: `withDbAccessContext` short-circuits into an ambient context, so a wrapping caller (the integration suite does this) would otherwise silently rebuild the single long transaction — the alertWorker/#3216 trap.
- The ML gate keeps its own labeled context: it reads RLS-forced `organizations`/`partners` and would see zero rows contextless (= every org silently disabled).
- `withSystemDbAccessContext` gains an optional low-cardinality `label`; under the tsup single-file bundle `parseOpenerFrame` collapses every anonymous-arrow opener to bare `index`, so labels are the only Sentry attribution (#3218).
- Worker level: `scan-orgs` labeled `metricRollups.scanOrgs`; the wrap around `rollup-org-range` removed with a comment explaining why re-adding it restores the bug.

## Tests
Trace-shape pins in `services/metricRollups.test.ts`: exactly 25 labeled contexts, nesting depth never exceeds 1, all 24 executes inside a context, an escape precedes every open, and the exact label set is asserted. Worker tests pin the scan-orgs label and that no worker-level context wraps rollup-org-range. `tsc --noEmit` clean; 19/19 targeted tests green.

## Provenance
Started by an issue-fixer agent that hit the session spend limit mid-service-edit; recovered and completed in the main session — the agent's db/index.ts + test work retained, the service/worker implementation finished to match its test specs.

https://claude.ai/code/session_01Bn3WFTjEBmK3nPqmUdQinb

### DIFF
--- a/apps/api/scripts/metric-rollup-backfill.ts
+++ b/apps/api/scripts/metric-rollup-backfill.ts
@@ -1,5 +1,5 @@
 #!/usr/bin/env tsx
-import { closeDb, withSystemDbAccessContext } from '../src/db';
+import { closeDb } from '../src/db';
 import { rollupDeviceMetricsRange } from '../src/services/metricRollups';
 import { parseMetricRollupBackfillArgs } from './metric-rollup-backfill.lib';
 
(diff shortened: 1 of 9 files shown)


ToddHebebrand:
Review round (high) complete — 4 confirmed findings, all fixed in 81701a64c:
1. Backfill script's outer context wrap was defeated by the per-statement escape — it pinned an idle-in-transaction connection for the whole (up to 31-day) pass and could fail the script on COMMIT after all inner statements committed. Dropped.
2. Integration suite had the same wrap; dropped, so the suite now exercises the exact production call shape — removing the escape or per-statement contexts would now fail loudly against forced RLS instead of being silently absorbed.
3. The escape's doc comments described the wrong hazard (claimed a wrap would be collapsed into; it is defeated). Rewritten.
4. Added the promised systemDbAccessContext unit test (blank-label fallthrough to the shared constant; trim-on-copy).

Verified clean by the review: escape semantics against db/index.ts AsyncLocalStorage stores, the accepted torn-read tradeoff (re-covered by the 15-min lookback), label cardinality. 22/22 targeted tests, tsc clean.

Issue #4002, opened 2026-08-25
getServerUrl cannot distinguish 'not configured' from 'read failed', so a keychain error silently routes a self-hosted tenant to the hosted default host


bdunncompany:
## Description

`getServerUrl()` cannot distinguish **"no server is configured"** from **"I could not read which server is configured"**, and every caller treats both as permission to fall back to the compiled-in host.

`apps/mobile/src/services/serverConfig.ts`:

```ts
export async function getServerUrl(): Promise<string | null> {
  try {
    return await SecureStore.getItemAsync(SERVER_URL_KEY);
  } catch (error) {
    console.error('Error retrieving server URL:', error);
    return null;
  }
}
```

A successful read of an unset key returns `null`. A **failed** read — a transiently locked keychain, a Keychain error, SecureStore unavailable — also returns `null`, after logging.

Every consumer then does:

```ts
const baseUrl = (await getServerUrl()) || FALLBACK_API_BASE_URL;
```

and `FALLBACK_API_BASE_URL` is `process.env.EXPO_PUBLIC_API_URL || 'http://localhost:3001'`, with `eas.json` setting `EXPO_PUBLIC_API_URL` to the hosted US host in all three build profiles.

## Impact

**A self-hosted or non-US tenant whose keychain read fails transiently sends its requests to the hosted US host instead.** That includes the authentication path — login and MFA request bodies go to a server the tenant never configured. The user sees a failed sign-in, not a misrouted one.

This is the same "a successful read and a failed read must not collapse to the same value" class as the CSRF handling in `csrfToken.ts`, which documents it explicitly:

> *"A read FAILURE is not the same as 'nothing stored', and folding the two together recreates the bug this module exists to fix."*

`getServerUrl` folds them together.

## Scope

Ten call sites use the `|| FALLBACK_API_BASE_URL` idiom (17 references to `getServerUrl` overall):

```
services/api.ts:252              services/aiChat.ts:96, :351
services/approvals.ts:56         services/approverDevice.ts:37
services/connectedApps.ts:28     services/mobileDevices.ts:42
services/search.ts:86            services/systems.ts:29
services/systemsRealtime.ts:93
```

None can currently tell the two cases apart, so none can fail closed.

## Suggested direction

Make the two outcomes distinguishable and let the destination fail closed when it is unknown — a successful read returning `null` legitimately means "not configured, fallback is fine"; a rejected read means the destination is unknown and falling back guesses at it.

Options, not a recommendation:

[...]

PR #4287, opened 2026-08-31
fix(mobile): fail the build when EXPO_PUBLIC_API_URL cannot reach a server (#3943)

### LINKED ISSUE #3943: EXPO_PUBLIC_API_URL pointing at localhost is silently fatal in a mobile release build
Found while fixing the mobile Sentry DSN gap (PR #3910).

`EXPO_PUBLIC_API_URL` has exactly the failure mode #3910 just closed for the Sentry DSN: a release build with a `localhost` (or missing) value produces an app that cannot reach the API, with **no build-time error**. The only check lives in `apps/mobile/scripts/preflight.mjs`, which nothing in the real build path invokes — the actual path is `expo prebuild` + Xcode Archive, and ⌘B skips the script entirely.

PR #3910 established the fix pattern: enforcement moved into `apps/mobile/app.config.js`, which the `expo-constants` Xcode build phase (`execution_position => :before_compile`, `always_out_of_date = "1"`) evaluates on **every** iOS build, and which `expo export:embed` evaluates on the Android release bundle path. The same hook covers this variable for roughly the same cost.

**Suggested fix:** extend the `app.config.js` guard to reject a release build whose `EXPO_PUBLIC_API_URL` is absent, `localhost`, or a private-range host.

ToddHebebrand:
Closes #3943

## The failure

`EXPO_PUBLIC_API_URL` is the base URL every mobile service module falls back to before the user has picked a server — nine files carry the same fallback:

```ts
const FALLBACK_API_BASE_URL = process.env.EXPO_PUBLIC_API_URL || 'http://localhost:3001';
```

Ship a release build with that variable unset or pointing at `localhost` and the literal string `http://localhost:3001` is compiled into the IPA. On a phone that is the phone, so every request dies with a transport error. Nothing throws at build time, nothing looks wrong in review, and there is no console to read in TestFlight — the same silent shape as the missing Sentry DSN that cost this app 90 days of zero telemetry (#3910).

Two paths reach that fallback beyond first run, both worth naming:

- `getServerUrl()` returns `null` when the SecureStore read **throws**, not only when nothing is stored — that is exactly the ambiguity **#4002** is about. A keychain hiccup on a configured device drops it onto this fallback.
- `SettingsSheet.tsx` passes `FALLBACK_API_BASE_URL` to `getAccountDeletionUrl()`, so a localhost fallback also breaks the account-deletion route Apple requires.

The only existing check lived in `apps/mobile/scripts/preflight.mjs`, which nothing in the repo invokes — the real release path is a human pressing Product → Archive.

## Build-time, not runtime — and why

**Build-time.** `EXPO_PUBLIC_*` values are inlined by the Metro transform, so this is a compile-time constant. A runtime check could only report a defect that is already inside a signed binary, and replacing that binary is an App Store review cycle. A build that never ships beats a shipped build that complains. The value is fully known at config-evaluation time, so there is no reason to defer it.

`app.config.js` is the one hook that cannot be skipped: `expo-constants` installs an Xcode script build phase (`execution_position => :before_compile`, `always_out_of_date = "1"`) that evaluates it on **every** build, and `expo export:embed` evaluates it on the Android release bundle. Same mechanism PR #3910 chose for the DSN, and why `preflight.mjs` was never enough.

Android local builds are covered too, and this was checked rather than assumed: `@expo/cli`'s `exportEmbedAsync` calls `setNodeEnv(options.dev ? 'development' : 'production')` at its top, *before* the `getConfig` call that evaluates this config — so `./gradlew bundleRelease` trips the `NODE_ENV="production"` signal. `expo run:android --variant release` and `expo run:ios --configuration Release` set it in their own entry points.

## What lands

- **`src/config/apiUrl.js`** (new) — `describeApiUrlProblem` + `assertReleaseApiUrl`. On a release build it rejects: unset, unparseable/non-http(s), placeholder hostnames, loopback (`localhost`, `*.localhost`, `127/8`, `0.0.0.0/8`, `::1`, and IPv4-mapped forms), private ranges (`10/8`, `172.16/12`, `192.168/16`, `169.254/16`, CGNAT `100.64/10`, IPv6 ULA/link-local, `.local`/`.lan`/`.internal`), and plaintext `http` to a **public** host.
- **`src/config/releaseBuild.js`** (new) — the "is this a release build?" logic and its `BREEZE_MOBILE_DEV` escape hatch, lifted out of `sentryDsn.js` now that a second guard needs it. `sentryDsn.js` re-exports every name it used to own, so **its existing 28-test suite is the regression net for the extraction** (it passes untouched). The suppression warning now names *which* check was silenced instead of always saying "Sentry DSN".
- **`app.config.js`** — calls the gate.
- **`scripts/preflight.mjs`** — now calls `describeApiUrlProblem` directly instead of keeping its own weaker copy of the rules, so the manual script and the build gate can never disagree.
- **`.env.example` / `STORE_SUBMISSION.md`** — both previously stated the API URL had *no* build-time guard. Corrected.

### Self-hosters are not blocked

Breeze is self-hostable, so this deliberately does not demand a `2breeze.app` host. Loopback is unconditional (no phone can reach the build machine's loopback; a Release-configuration simulator run is what `BREEZE_MOBILE_DEV=1` is for). A private address **is** legitimate for an MSP building an internal-distribution app against a LAN-hosted Breeze, so it gets a named opt-in rather than a veto:

```
BREEZE_MOBILE_ALLOW_PRIVATE_API_URL=1
```

That accepts a private-network host, **plaintext http included** (a LAN box often has no publicly-issued certificate), and **warns on stderr on every build where it rescued a value**. It does *not* accept loopback, a placeholder, or plaintext to a **public** host, and it takes the literal `1` so a near-miss spelling fails safe. Deliberately narrower than `BREEZE_MOBILE_DEV=1`, which would also switch off the Sentry guard.

## Verification

**Red first.** `appConfigApiUrl.test.ts` was written before the fix and failed with `AssertionError: expected [Function] to throw an error` — with a valid DSN stubbed, so the Sentry guard could not manufacture a false pass.

**End-to-end through the real Expo path**, not just a vitest import of `app.config.js`:

| `expo config --json --type public` | result |
|---|---|
| `CONFIGURATION=Release` + `http://localhost:3001` | **exit 1** — `points at "http://localhost:3001", which on a phone is the phone itself` |
| `CONFIGURATION=Release` + `http://us.2breeze.app` + `ALLOW_PRIVATE=1` | **exit 1** — `uses plaintext http` (the opt-in must not cover a public host) |
| `CONFIGURATION=Release` + `192.168.1.50` + `ALLOW_PRIVATE=1` | exit 0, **with the `[breeze] WARNING:` line on stderr** |
| `NODE_ENV=production` + `http://localhost:3001` (Android bundle path) | **exit 1** — `This is a release build (NODE_ENV="production")` |
| `CONFIGURATION=Release` + `https://us.2breeze.app` | exit 0, silent |
| `CONFIGURATION=Debug` + `http://localhost:3001` | exit 0 — the dev loop is untouched |

**Suites** (`apps/mobile`): 55 files, 681 passed / 3 skipped — 45 new assertions across `apiUrl.test.ts`, `appConfigApiUrl.test.ts` and `releaseBuild.test.ts`. `tsc --noEmit` clean, `pnpm lint` clean, `pnpm preflight` exercised across unset / localhost / private+opt-in / public-plaintext / good.

**Not vacuous.** Eight mutations were each confirmed to turn the suite red and then reverted — disabling loopback detection, widening `172.16/12` to `172/8`, matching placeholders against the whole URL, re-introducing the plaintext gating bug, removing the opt-in warning, disabling IPv4-mapped folding, reverting label matching to substrings, and routing `warnFn` to `console.warn` instead of stderr. For each, the mutation diff was verified as applied before trusting the red.

## Review round

`/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer. Three real defects found and fixed in `e40ef156a`; details in the review-summary comment below.

## Relationship to #4002

No overlap in code. #4002 is about `getServerUrl()` in `src/services/serverConfig.ts` conflating "not configured" with "read failed"; this PR adds build-time modules with no import relationship to the runtime services layer. They compose rather than conflict: #4002 narrows *when* the fallback is reached, this makes the fallback itself safe to reach. No other open PR touches `apps/mobile/src/config/` or `app.config.js`.

## Deliberate scope boundaries

- **The validated URL is not republished on `expo.extra`.** The DSN is, because App.tsx needs a second delivery path. Here every consumer already reads `process.env.EXPO_PUBLIC_API_URL` directly, and rewiring those nine modules is the same code path #4002 is changing — and `services/api.ts` is already being edited by open PRs #4259 and #4237. So this stays a pure gate. The residual gap is honest and narrow: config-phase env and Metro-transform env come from the same `apps/mobile/.env` / EAS profile, but they are technically separate phases, so a guard pass does not *prove* the bundled value. Consolidating the nine duplicated `FALLBACK_API_BASE_URL` constants behind one resolver is the right follow-up, after #4002 lands.
- **The two gates report the first failure, not both.** A release build missing the DSN *and* the API URL surfaces the DSN first. Now covered by an explicit test documenting that as intent rather than accident; aggregating would mean changing `resolveSentryDsn`'s shipped contract for a case that costs one rebuild.

## Note on CI

`test-mobile` runs on every PR but is **not yet a required check** (PR #4254 makes it blocking), so a green badge alone would not prove these tests ran. Confirmed from the job log instead — see the review-summary comment.


ToddHebebrand:

**Findings:** 6 raised → all 6 addressed in `e40ef156a`; 0 outstanding.

Three were real defects, all in code I wrote in the first commit:

1. **`silent-failure-hunter` + `pr-test-analyzer` (independently):** the plaintext-http check was gated on `!allowPrivate`, but every private host returns before reaching it — so that branch only ever ran for **public** hosts, and `BREEZE_MOBILE_ALLOW_PRIVATE_API_URL=1` silently permitted `http://` to a public host. A flag left in `.env` after LAN testing would have shipped a build ATS blocks on the device, with no console — the exact failure class this PR exists to close. Fixed: the flag now means "a private host is fine", never "plaintext to the open internet is fine".
2. **`silent-failure-hunter`:** that same opt-in suppressed its check in **total silence** — the only escape hatch in these files that said nothing, and a direct inconsistency with `BREEZE_MOBILE_ALLOW_NO_SENTRY` next door. `assertReleaseApiUrl` now computes the strict problem first, establishes the release reason, then applies the opt-in and **warns** when it rescued a value. Same "compute before suppress" ordering `releaseBuildReason` documents as the bug shape to avoid.
3. **`code-reviewer` (raised as a low-confidence suggestion; I'd found it independently while probing edge cases):** `normalizeHost` missed IPv4-mapped IPv6 — the URL parser rewrites `[::ffff:127.0.0.1]` to `[::ffff:7f00:1]`, matching neither rule, so loopback written that way passed. Folded back to a dotted quad.

The other three were `comment-analyzer` accuracy fixes: softened two `(verified — …)` claims about Expo internals to point-in-time SDK 57 observations, corrected the `0.0.0.0/8` scope wording, and stopped describing `api.ts`'s exported two-line fallback as "identical" to the eight module-private one-liners.

**Also fixed unprompted:** placeholder hostnames now match reserved domains by suffix and marker words by whole **label**, so `forexample.com` and `changemedia.com` are no longer rejected by a substring hit.

**Resolved rather than fixed:** `silent-failure-hunter` flagged (explicitly as *inferred, not verified* — it had no `node_modules`) that a local `./gradlew bundleRelease` might set none of the three release signals. Checked against the installed `@expo/cli`: `exportEmbedAsync` calls `setNodeEnv(options.dev ? 'development' : 'production')` at its top, before the `getConfig` that evaluates `app.config.js`, so the Android Gradle path trips `NODE_ENV="production"`. Confirmed end-to-end (`NODE_ENV=production` + localhost → exit 1) and documented in `releaseBuild.js`.

**`code-reviewer` verdict:** 0 critical, 0 important — including a semantic diff of `origin/main`'s `sentryDsn.js` against the new `releaseBuild.js` + `sentryDsn.js` pair confirming the extraction is behaviour-preserving, and independent verification of every CIDR boundary.

**Tests:** `apps/mobile` 55 files, 681 passed / 3 skipped (`pnpm exec vitest run`); +20 assertions in this round, including a new `releaseBuild.test.ts` covering `warnFn`'s default **stderr** branch — every prior test injected a spy while `app.config.js` passes no `io` at all, so 100% of real build-warning traffic went through the one untested path. `tsc --noEmit` clean; `pnpm lint` clean.

**Mutation-checked:** all eight behaviours (three from the first commit, five from this round) were each confirmed to turn the suite red, then reverted. One mutation initially appeared to pass — the `perl` substitution had not matched — so it was re-run with a line-targeted edit and the mutation diff verified as applied before trusting the red.

**CI:** `Test Mobile` is green **and confirmed to have actually run these tests** — the job log shows `✓ src/config/apiUrl.test.ts (25 tests)` and `✓ src/config/appConfigApiUrl.test.ts (4 tests)`. Worth stating explicitly because `test-mobile` is not yet a required check until #4254 lands, so a green badge alone would not have proven it. All 55 checks on the first commit settled with 0 failures; this round's push is re-running.

**Status:** review-clean, awaiting maintainer merge. Issue #3943 stays open and assigned.

ToddHebebrand:
**CI update — review-fix commit `e40ef156a` has settled: 56 pass, 0 fail, 0 pending.** (The one non-pass is `Main Red Alert`, skipped because it is a main-branch-only workflow.)

`Test Mobile` on this commit reports **55 files / 681 tests passed**, matching the local run exactly — so `releaseBuild.test.ts`, added in this commit, ran in CI too. Together with the earlier log line for `apiUrl.test.ts (25 tests)` and `appConfigApiUrl.test.ts (4 tests)`, every new test file in this PR is confirmed executed by CI rather than merely assumed from a green badge — worth stating because `test-mobile` is not a required check until #4254 lands.

Nothing outstanding. Ready for a maintainer merge decision; not merging, and #3943 stays open and assigned.

PR #4288, opened 2026-08-31
fix(api): close the check-then-act race in every single-alert acknowledge and suppress path (#4101)

### LINKED ISSUE #4101: [API] Single-alert acknowledge handlers have the same check-then-act race the resolve paths had
Carve-out from PR #4099 (issue #4094), where it's noted in the PR body as a deliberate non-scope-creep follow-up.

The single-alert **acknowledge** handlers use the identical read-check-write shape that made the resolve paths racy before #4083/#4099 introduced the `buildResolveAlertCas` compare-and-swap. Two concurrent acknowledges (or acknowledge racing a resolve) can both pass the status check and both write.

Severity is much lower than the resolve case — acknowledgement triggers no escalation cancellation and no `'*'` automation fan-out — which is why #4099 left it alone. The fix is mechanical now that the pattern exists: an acknowledge CAS predicate built the same way as `buildResolveAlertCas`, asserted as compiled SQL via `PgDialect` (not column-name substrings — see the vacuous-assertion trap that hit #4083's first draft).

ToddHebebrand:
Closes #4101

## The bug

The single-alert **acknowledge** and **suppress** handlers carried the identical read-check-write shape the resolve paths had before #4083/#4099 introduced `buildResolveAlertCas`: read the alert's status, then `UPDATE ... WHERE id = $1` unconditionally. Two concurrent callers both clear the check and both write.

Per the issue's own review comment, the damage is worse than a double-acknowledge:

> Tech A resolves (CAS wins, `alert.resolved` published, escalation cancelled); tech B's stale list still shows it active and B clicks Acknowledge — B's id-only UPDATE stamps `status='acknowledged'` over the resolution, producing a reopened alert with `resolvedAt`/`resolvedBy` populated and its escalation already cancelled.

Suppress has the same shape: a stale suppress landing on a just-resolved alert un-resolves it.

## The fix — reusing the resolve-side pattern, not inventing a second one

Two new status lists and two new predicate builders sit beside `buildResolveAlertCas` in `apps/api/src/services/alertService.ts`. All three now funnel through one private `buildAlertStatusCas`, so the CAS **shape** has a single definition and cannot change on one transition while missing the other two — the exact drift the existing `buildResolveAlertCas` docblock warns about.

| Constant | Value | Why |
|---|---|---|
| `ACKNOWLEDGEABLE_ALERT_STATUSES` | `['active']` | Every ack call site already refuses a non-active alert at its pre-read; the CAS is that same rule expressed where it is enforceable |
| `SUPPRESSIBLE_ALERT_STATUSES` | `['active','acknowledged','suppressed']` | Everything except the two terminal statuses. `suppressed` stays in the set on purpose — re-timing an existing mute is legitimate, unlike re-acknowledging |

`SUPPRESSIBLE_*` is textually identical to `RESOLVABLE_*` today and is kept a **separate** constant anyway: the two express different invariants and must be free to diverge when a status is added. Aliasing them would make a future non-terminal status silently inherit both answers.

## Call sites — enumerated and verified, not trusted

The issue said five. I enumerated `status: 'acknowledged'` / `status: 'suppressed'` writes across `apps/api/src` and `ee/` and confirmed it: **five single-alert sites were racy, two multi-row sites were already correct.**

**Fixed (5):**

| # | File | Handler | Was |
|---|---|---|---|
| 1 | `apps/api/src/routes/alerts/alerts.ts` | `POST /alerts/:id/acknowledge` | `.where(eq(alerts.id, alertId))` |
| 2 | `apps/api/src/routes/alerts/alerts.ts` | `POST /alerts/:id/suppress` | `.where(eq(alerts.id, alertId))` |
| 3 | `apps/api/src/routes/mobile.ts` | `POST /mobile/alerts/:id/acknowledge` | `.where(eq(alerts.id, alertId))`, **and never read `RETURNING`** |
| 4 | `apps/api/src/services/aiToolsAlerts.ts` | `manage_alerts { action: 'acknowledge' }` | `.where(eq(alerts.id, ...))`, **no `RETURNING` at all** |
| 5 | `apps/api/src/services/aiToolsAlerts.ts` | `manage_alerts { action: 'suppress' }` | `.where(eq(alerts.id, ...))`, **no `RETURNING` at all** |

**Already correct — verified, left unchanged (2):** both write many rows and need `inArray(alerts.id, ...)` rather than an id equality, so they compose the status predicate directly:

- `routes/alerts/alerts.ts` bulk action — stricter still, pins each row to `eq(alerts.status, alert.status)` from its snapshot.
- `routes/alerts/correlations.ts` `mutateAlerts` — `eq(alerts.status, 'active')` for acknowledge.

## Silent failures fixed along the way

- **`routes/mobile.ts`** never looked at the UPDATE's `RETURNING` (`updated?.id ?? alertId`). A write matching zero rows — a lost race, *or a row invisible to this tenant context, which raises no error under `breeze_app` RLS* — still published `alert.acknowledged`, still fed the ML feedback loop, still wrote an audit entry, and still answered **200 with a `null` body**.
- **Both AI-tool branches** requested no `RETURNING` and returned `success: true` unconditionally.

## Response codes

- CAS loser → **409** with a message that states only what the code can verify (never "another request acknowledged it" — an empty `RETURNING` is also what an RLS-invisible write produces).
- `acknowledged` at the pre-read → **409** (was 400). Losing at the read and losing at the write are the same real-world event, so they must not return two codes purely on timing. **#4099 made exactly this change for resolve** (`git log -S` confirms the 400 → 409 flip landed in `1c59bb687`).
- `resolved` / `dismissed` at the pre-read → stays **400** (illegal transition, not a race), matching resolve's treatment of `dismissed`.
- No client depends on the old codes: `AlertsPage.tsx` / `AlertDetailPage.tsx` apply their optimistic update *after* `runAction` resolves, so a 409 skips it and toasts. The removed `'Failed to acknowledge alert'` / `'Failed to suppress alert'` 500 strings had zero references repo-wide.

## Tests — written red first

**20 of 31 new assertions failed against the unpatched source**, including the ones that matter most: the mobile route publishing `alert.acknowledged` on a zero-row write, both AI-tool branches reporting `success: true`, and every compiled-SQL assertion (main's predicate has no status term at all).

Four suites, mirroring the four established `*.resolveCas.test.ts` files. Each asserts **both**:

- **Behaviour** — an empty `RETURNING` (exactly what Postgres gives a caller whose CAS matched nothing) must produce the conflict and **no** fan-out: no event, no ML feedback, no audit.
- **Compiled SQL** — the predicate the handler *actually passed* to `.where()`, compiled through `PgDialect`. A mocked-drizzle assertion that substring-matches column names cannot tell `and` from `or` (which would stamp every matching alert in **every** tenant from one request) and cannot see the status list gaining a value that makes the CAS a no-op.

> **Concurrency disclosure:** these are compiled-SQL + lost-race-behaviour assertions, **not** two genuinely concurrent transactions against real Postgres. The unit job has no database. The predicate is what makes the write atomic in Postgres, and these tests pin both the predicate and the handler's response to losing.

`alertService.ackCasSql.test.ts` additionally derives its expectations from the `alert_status` **enum**, so adding a status forces a deliberate classification instead of silently inheriting one list's answer.

New files:
- `apps/api/src/services/alertService.ackCasSql.test.ts`
- `apps/api/src/routes/alerts/alerts.ackSuppressCas.test.ts`
- `apps/api/src/routes/mobile.ackCas.test.ts`
- `apps/api/src/services/aiToolsAlerts.ackSuppressCas.test.ts`

## Verification

- New suites: **31 passed** (4 files).
- Regression, single-fork: `alerts.test.ts`, `alerts.resolveCas.test.ts`, `alerts.authz.test.ts`, `correlations.test.ts`, `correlations.resolveCas.test.ts` → **98 passed**; `mobile.test.ts`, `mobile.resolveCas.test.ts`, `mobile.cursor.test.ts` → **80 passed / 3 skipped**; nine `alertService.*` / `aiToolsAlerts.*` / `warrantyAlertEvaluator.resolveCas` suites → **55 passed**.
- `eslint` on all eight touched files: clean.
- No schema, migration, or tenancy-table change — no RLS or cascade-registration surface touched.

## Docs

`apps/docs/src/content/docs/reference/api.mdx` now records the 409-on-conflict contract for acknowledge/resolve/suppress (web + mobile), which #4099 introduced for resolve but never documented.

### DIFF
--- a/apps/api/src/routes/alerts/alerts.ackSuppressCas.test.ts
+++ b/apps/api/src/routes/alerts/alerts.ackSuppressCas.test.ts
@@ -1,0 +1,286 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Hono } from 'hono';
+import { PgDialect } from 'drizzle-orm/pg-core';
+import type { SQL } from 'drizzle-orm';
+
+/**
+ * POST /alerts/:id/acknowledge and POST /alerts/:id/suppress are compare-and-swaps
+ * (#4101), for the same reason POST /alerts/:id/resolve became one in #4094/#4099.
+ *
+ * Both used to read the alert's status and then UPDATE by id unconditionally —
+ * textbook check-then-act. The damage is not merely a double-acknowledge: tech A
+ * resolves (the resolve CAS wins, `alert.resolved` publishes, the escalation is
+ * cancelled), tech B's stale list still shows the alert active and B clicks
+ * Acknowledge. B's id-only UPDATE stamps `status='acknowledged'` over the
+ * resolution, leaving a reopened alert that carries `resolvedAt`/`resolvedBy` and
+ * whose escalation is already gone. Suppress has the identical shape.
+ *
+ * Two kinds of assertion here, deliberately (see alerts.resolveCas.test.ts):
+ *
+ *  - BEHAVIOUR — an empty `RETURNING` (exactly what Postgres gives a caller whose
+ *    CAS matched nothing) must produce a 409 and NO fan-out.
+ *  - COMPILED SQL — the predicate the handler actually passed to `.where()`,
+ *    compiled through `PgDialect`. drizzle-orm and ../../db/schema are REAL in this
+ *    file for that reason: a mocked-drizzle `where` assertion can only substring-
+ *    match column NAMES, which cannot tell `and` from `or` and cannot see the
+ *    status list gaining a status that makes the CAS a no-op.
+ */
+const { dbMock, updateWheres, updateReturns, selectRows, alertRow } = vi.hoisted(() => {
+  const updateWheres: unknown[] = [];
+  const updateReturns: unknown[][] = [];
+  const selectRows: unknown[][] = [];
+  const alertRow = {
+    id: '33333333-3333-4333-8333-333333333333',
+    orgId: 'org-1',
+    deviceId: 'device-1',
+    ruleId: null as string | null,
+    configPolicyId: null as string | null,
+    context: null as unknown,
+    status: 'active',
+    title: 'Disk almost full',
+    triggeredAt: new Date('2026-01-01T00:00:00.000Z'),
+  };
+  const dbMock = {
+    select: () => ({
+      from: () => ({ where: () => ({ limit: () => Promise.resolve(selectRows.shift() ?? []) }) }),
+    }),
+    update: () => ({
+      set: () => ({
+        where: (predicate: unknown) => {
+          updateWheres.push(predicate);
+          return { returning: () => Promise.resolve(updateReturns.shift() ?? []) };
+        },
+      }),
+    }),
+  };
+  return { dbMock, updateWheres, updateReturns, selectRows, alertRow };
+});
+
+const publishEvent = vi.fn((..._args: unknown[]) => Promise.resolve('evt'));
+const emitAlertStateFeedback = vi.fn((..._args: unknown[]) => Promise.resolve());
+const writeRouteAudit = vi.fn();
+const getAlertWithOrgCheck = vi.fn();
+
+vi.mock('../../db', () => ({
+  db: dbMock,
+  runOutsideDbContext: (fn: () => unknown) => fn(),
+  withSystemDbAccessContext: (fn: () => unknown) => fn(),
+}));
+vi.mock('../../middleware/auth', () => ({
+  authMiddleware: async (_c: unknown, next: () => Promise<void>) => next(),
+  requireScope: () => async (c: { set: (k: string, v: unknown) => void }, next: () => Promise<void>) => {
+    c.set('auth', {
+      scope: 'organization',
+      user: { id: 'user-1', name: 'Tech One', email: 'tech@org.example' },
+      partnerId: null,
+      orgId: 'org-1',
+      accessibleOrgIds: null,
+      canAccessOrg: () => true,
+    });
+    await next();
+  },
+  requirePermission: () => async (c: { set: (k: string, v: unknown) => void }, next: () => Promise<void>) => {
+    c.set('permissions', { granted: new Set<string>() });
+    await next();
+  },
+  requireMfa: () => async (_c: unknown, next: () => Promise<void>) => next(),
+  siteAccessCheck: () => () => true,
+}));
+vi.mock('../../services/alertCooldown', () => ({
+  setCooldown: vi.fn(),
+  markConfigPolicyRuleCooldown: vi.fn(),
+}));
+vi.mock('../../services/auditEvents', () => ({ writeRouteAudit: (...a: unknown[]) => writeRouteAudit(...a) }));
+vi.mock('../../services/eventBus', () => ({ publishEvent: (...a: unknown[]) => publishEvent(...a) }));
+vi.mock('../../services/mlFeedbackEmitters', () => ({
+  emitAlertStateFeedback: (...a: unknown[]) => emitAlertStateFeedback(...a),
+  emitCorrelationFeedback: vi.fn(),
+}));
+vi.mock('../../services/ticketService', () => ({
+  createTicketFromAlert: vi.fn(),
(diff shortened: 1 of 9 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (all three on Sonnet, against `git diff origin/main...HEAD`).

**Findings:** 0 raised across all three lenses; 0 outstanding. No commits were needed after review.

Highlights the reviewers independently verified rather than took on trust:

- **Call-site enumeration re-done from scratch** — grepped every `.update(alerts)` that sets `status` to `'acknowledged'`/`'suppressed'` across `apps/api/src` and `ee/`. Confirms **five** single-alert sites (all fixed) and that the two multi-row sites are already correct and rightly untouched: the bulk action pins `eq(alerts.status, alert.status)` per row from its snapshot, and `mutateAlerts` in `routes/alerts/correlations.ts` pins `eq(alerts.status, 'active')`. `mobile.ts` has no suppress route at all, so nothing is missing there.
- **The `buildResolveAlertCas` refactor is SQL-neutral** — the pre-existing, untouched `alertService.resolveCasSql.test.ts` still asserts the identical compiled string and passes.
- **Mutation testing of the new suites** (pr-test-analyzer built an isolated worktree and did this empirically): reverting the four source files to `origin/main` reproduced exactly **20 failed / 11 passed of 31**. Each of the four mutations was then applied individually and caught — `and`→`or` (7 tests), dropping the status predicate (7), admitting a terminal status into `SUPPRESSIBLE_ALERT_STATUSES` (5, including the enum-derived drift guard, proving it is not vacuous), and removing the `if (!updated) return 409` gate (both the `[updated]`-destructure and `.length === 0` shapes). The 11 pre-fix passes were audited and are all legitimate happy-path / pre-read-guard assertions — none is mislabeled as CAS-discriminating.
- **No fan-out survives a zero-row CAS** — all six single-alert acknowledge/suppress paths return before `publishEvent` / `emitAlertStateFeedback` / `writeRouteAudit`. No `?.` or `?? alertId` fallback remains.
- **400 → 409 has no dependents** — web, mobile app, e2e and docs checked; `runAction.ts` special-cases only 401 and treats 400/409 identically, and no caller branches on the old 400.

**Tests:** all affected suites run single-fork on Node 22.23.2. New suites `alertService.ackCasSql.test.ts` + `alerts.ackSuppressCas.test.ts` + `mobile.ackCas.test.ts` + `aiToolsAlerts.ackSuppressCas.test.ts` → **31 passed**. Regression: `alerts.test.ts` / `alerts.resolveCas.test.ts` / `alerts.authz.test.ts` / `correlations.test.ts` / `correlations.resolveCas.test.ts` → 98 passed; `mobile.test.ts` / `mobile.resolveCas.test.ts` / `mobile.cursor.test.ts` → 80 passed, 3 skipped; nine `alertService.*` / `aiToolsAlerts.*` / `warrantyAlertEvaluator.resolveCas` suites → 55 passed; the sibling `src/routes/alerts.test.ts` plus `byid.sitescope` / `actorNames` / `alertTickets` / `hideAiNoise.sql` → 70 passed.

**Typecheck:** `tsc --noEmit --project apps/api/tsconfig.json` exit 0. `eslint` clean on all eight touched files.

**CI:** green on the head SHA — **56 pass, 0 fail**, 1 skipped (`Main Red Alert`, a main-only job). Includes `Test API`, `Test Web`, `Typecheck`, `Lint`, all four `Integration Tests` shards, and both `Check Migrations` jobs.

**Concurrency disclosure (stated plainly):** these are compiled-SQL assertions plus lost-race *behaviour* assertions — **not** two genuinely concurrent transactions against real Postgres. The unit job has no database. The status predicate is what makes the write atomic in Postgres; these tests pin both the predicate the handler actually ships and the handler's response to losing.

**Status:** review-clean, CI green, awaiting maintainer merge. No schema, migration, or tenancy-table surface touched — no RLS or cascade-registration obligations apply.

Issue #3900, opened 2026-08-23
W06: Auto-suggested time entries from remote_sessions durations


ToddHebebrand:
Wave W06 of #3206 — Auto-suggested time entries from remote_sessions durations
Branch: `feature/3206-mobile-ticketing-time-entry/wave-3900-api`

ToddHebebrand:
Out of scope for the current plan (`docs/superpowers/plans/mobile/2026-08-23-mobile-time-entry.md`) — needs its own plan before it starts.

The product angle from the parent issue: durations are already recorded and unbilled. `remote_sessions.duration_seconds` is the strongest signal (a remote support session is literally billable labor), with `support_sessions`, `script_executions`, `device_commands` and ticket status transitions behind it. End-of-day prompt: "You ran a 38-minute remote session on ACME-DC01 and closed TKT-1041 — log it?"

Depends on W03 (the client) and W02 (the permission answer).

ToddHebebrand:
## W06 design spec + implementation plan are up

Docs PR: **https://github.com/LanternOps/breeze/pull/4258** (docs-only, no code)

- **Spec:** `docs/superpowers/specs/mobile/2026-08-29-auto-suggested-time-entries-design.md`
- **Plan:** `docs/superpowers/plans/mobile/2026-08-29-auto-suggested-time-entries.md` (18 tasks)

### Summary

W06 lets a technician turn a recorded remote session into a time entry with one tap, without ever writing one on their behalf. Suggestions are computed at read time from `remote_sessions` rows the caller can already see under RLS — there is no suggestions table — and the only new table is `time_suggestion_decisions`, a partner-axis ledger whose `UNIQUE (user_id, signal_kind, signal_id)` index doubles as the offline-replay idempotency key and whose `ON DELETE SET NULL` link to `time_entries` keeps a deleted entry from silently re-suggesting the session. A new server-stamped `time_entries.source` column records provenance (`manual | timer | location | remote_session | support_session`); no public schema accepts it, so it cannot be forged, and the confirm route re-reads every signal under the caller's own DB context before creating an ordinary entry through the existing ticket/org currency path. The wave is off by default behind a partner setting, dispatches no push (it ships only the payload builder and the reserved dedupe keys for W07), and deliberately skips any session the technician has already logged by hand or with the timer, so the feature can never put a duplicate billable row one tap away.

### PR split

[...]

Issue #3778, opened 2026-08-21
W06: Enable org currency selection/change (release-gated on vertical slices)


ToddHebebrand:
Wave W06 of #3772 — Enable org currency selection/change (release-gated on vertical slices)
Branch: `feature/3772-multi-currency/wave-3778`

ToddHebebrand:
Completed via PR #3874

Squash-merged 2026-08-23 with full CI green (57/57) after a codex plan review (13 findings, 3 blockers) and a code review round (7 findings, all fixed). The §13 release gate was run FIRST and found 6 real defects in already-merged waves 3-5 code — chiefly a missing isRepresentableInCurrency guard on unitPrice/rate across five services, and addManualLine storing a 2dp unit_price against a line_total derived from the raw input. Ships: org currency selector + preflight/locked change transaction, org FOR SHARE creation barrier, active-contract escape hatch keyed on a new durable invoice_lines.source_contract_id lineage column, and the contract-vs-org currency mismatch report (the owner-approved escape-hatch + report decision).

PR #4290, opened 2026-08-31
feat: Auto-suggested time entries — backend, web parity and docs (#3206 W06-A)

ToddHebebrand:
Part of #3900 (`#3206` W06). **The mobile half has not shipped.** This PR is W06-A only — Tasks 1–13 of `docs/superpowers/plans/mobile/2026-08-29-auto-suggested-time-entries.md` (backend, web parity, docs). Tasks 14–18 (the React Native client, the queued write kinds, the screens) are a separate PR that depends on the still-unmerged W03–W05, so `#3900` stays open.

## What this is

At the end of the day a technician sees *"You ran a 38-minute desktop session on ACME-DC01 and closed T-2026-1041 — log it?"* and one tap creates an ordinary `time_entries` row with server-stamped provenance. **Nothing is written without the tap**, and the whole feature is invisible until a partner admin turns it on (off by default).

Suggestions are computed **at read time** from `remote_sessions` rows the caller can already see under RLS — there is no suggestions table. The only new table is a partner-axis decisions ledger that records confirmed/dismissed per signal and doubles as the offline-replay idempotency key.

## Tasks completed (1–13)

| # | Task | Commit |
|---|---|---|
| 1 | Migration + Drizzle schema (`time_entries.source`, signal index, decisions ledger) | `908143f71` |
| 2 | Registration lists + contract suites | `b5d6a705a` |
| 3 | Shared validators + `source` vocabulary | `e1fb864c1` |
| 4 | Partner `sessionSuggestions` setting (default off) + `PATCH /partners/me` | `ef051d55f` |
| 5 | Pure suggestion rules — day window, precision, merge, ticket ranking, F19 overlap | `6d103c354` |
| 6 | `timeEntryService` provenance: `source` stamps, internal `provenance`, `resolveAndLockOrgLink` | `0f49b7575` |
| 7 | `timeSuggestionService` — signal query, list, count | `ced4fb83b` |
| 8 | Confirm / dismiss / undismiss with the decisions ledger | `ea69dff2a` |
| 9 | The four routes (`routes/timeEntries/suggestions.ts`) | `b21354d60` |
| 10 | `time_suggestion_decisions` RLS + tenancy integration suite | `7db18caa5` |
| 11 | W07 push hook — `buildTimeSuggestionPush`, reserved event/dedupe keys (**no dispatch in this wave**) | `7b670bae8` |
| 12 | Web parity — `source` badge + Settings → Ticketing → Time Tracking (8 locales) | `1a883d3d2` |
| 13 | Docs + release notes | `b24c3a15b` |
| — | Two gaps this session found and closed (below) | `1092f4d4d`, `3f29e4d57` |

**No push is dispatched in this wave.** Task 11 ships the payload builder and reserves the `push_notifications.event_type` string and the `user_notifications.dedupe_key` shape so W07 and this wave cannot pick different spellings. Nothing calls `buildTimeSuggestionPush`.

## Migration

`apps/api/migrations/2026-09-25-time-entry-source-and-suggestion-decisions.sql` — sorts last against `origin/main` (whose last file is `2026-09-24-b-…`); `scripts/check-migration-naming.sh` OK. The plan named `2026-09-23-…`, which had since been taken on main; dated forward rather than wedged into a shipped date block.

⚠️ **It takes a brief blocking lock on `remote_sessions`.** `remote_sessions_user_ended_idx` is built with a plain `CREATE INDEX` — `CONCURRENTLY` is impossible because `autoMigrate` wraps each file in one transaction. On a self-hosted instance with millions of `remote_sessions` rows, session start/end writes block for the build. Called out in `docs/release-notes/next-release-draft.md`. Verified idempotent by replaying the file against an already-migrated database (all `IF NOT EXISTS` paths took the skip branch, exit 0).

## Registration lists — including the ones deliberately NOT edited

| List | Edited? | Reason |
|---|---|---|
| `PARTNER_TENANT_TABLES` (`rls-coverage.integration.test.ts`) | ✅ added `time_suggestion_decisions` → `partner_id` | RLS Shape 3, same policy shape as `time_entries` |
| `CORE_TENANT_EXPORT_POLICY` (`tenantExportPolicyRegistry.ts`) | ✅ added `source` to `time_entries`.`included` | **A new COLUMN on a long-registered org-cascade table breaks this suite** — the row this repo's contract most often misses |
| `ALLOWED_WITHOUT_CAPABILITY_CHECK` (`partner-wide-write-coverage.test.ts`) | ✅ added `services/timeSuggestionService.ts` | Rows are **per-technician** (`user_id = actor`); `partner_id` exists only to give the Shape-3 policy an axis. Gating on `canManagePartnerWidePolicies()` would mean only a partner admin could dismiss their own suggestion |
| `CORE_ORG_CASCADE_DELETE_ORDER` (`tenantCascade.ts`) | ❌ | `time_suggestion_decisions` has **no `org_id` column**; that list enumerates org-axis tables only. Verified against `tenantCascade.ts`, not taken on the plan's word |
| `CORE_DEVICE_CASCADE_DELETE_TABLES` / `CORE_DEVICE_ORG_DENORMALIZED_TABLES` (`routes/devices/core.ts`) | ❌ | No `device_id` column. `signal_id` is deliberately FK-less and un-indexed by device — a purged Quick Support device must leave an inert orphan, not cascade |
| `AUDIT_ADMIN_REQUIRED_TABLES` (`tenantCascade.ts`) | ❌ | Not append-only: no `REVOKE DELETE`, no immutability trigger. Undismiss genuinely deletes rows |
| `orgMergeRegistry.ts` (the 6th list) | ❌ | Its contract scope is `getOrgCascadeDeleteOrder() + EXTRA_REQUIRED`. `time_suggestion_decisions` is in neither, and adding it would **fail** `orgMergeRegistry.integration.test.ts`'s "no policy names an unrequired table". Verified by reading the suite, not inferred |
| `CORE_TENANT_EXPORT_POLICY` entry *for the new table* | ❌ | A table with no `org_id` needs no entry |
| Partner cascade | ❌ (nothing to edit) | `cascadeDeletePartner` **auto-discovers** `partner_id` tables from the catalog and topo-sorts them; the `partner_id` FK is `ON DELETE CASCADE`. Proved by an integration case that deletes the partner and asserts the rows are gone |
| `translationCoverage.test.ts` duplicate baselines | ✅ de-DE 104→105, it-IT 105→106 | `longTail.time.sourceBadge.timer` is "Timer" in both catalogs — already the word each uses for the running-timer widget |

## Tenancy shape

`time_suggestion_decisions` is **Shape 3 (partner-axis)**, one `FOR ALL` policy `system OR breeze_has_partner_access(partner_id)`, `ENABLE` + `FORCE` in the creating migration. Confirmed against real Postgres (`\d+` shows *"Policies (forced row security enabled)"*). The cross-tenant forge is an integration case, not a manual note: *"forging a decision for ANOTHER partner fails with 42501"* passes.

## Where the risk is, and how it is pinned

- **Task 6 changes `createTimeEntry`'s signature.** A third `provenance` argument, defaulting to `{ source: 'manual' }`, so every existing caller is unchanged. A wrong org link here writes a **billable row in the wrong currency**, so: a ticket link always wins over an org link (the ticket path holds the ticket + org locks, creation barrier #3778); `resolveAndLockOrgLink` re-checks `entryOrgAllowed` **and** the org's `partner_id` before taking the same `organizations FOR SHARE` the ticket path takes; and the partner-currency fallback is now gated on `currencyCode == null` so an org-linked entry can never be re-denominated in the partner's currency. Five service tests cover exactly those branches.
- **Task 7's isolation *is* the SQL.** The tests compile the query with `new PgDialect().sqlToQuery(...)` and assert on the emitted string and bound parameters — `rs.user_id = $n`, `o.partner_id = $n`, `rs.org_id = ANY(...)`, and that an empty allowlist renders `ARRAY[]::uuid[]` (matches nothing) rather than dropping the predicate. No assertion is made against a Drizzle builder object. The `uuidArray` helper binds one parameter per element on purpose — interpolating a JS array is the #2655 `malformed array literal` class that mocked tests cannot see.
- **Task 8 writes the money row and the ledger rows together.** Both happen inside the request transaction that `withDbAccessContext` already holds open for `SET LOCAL`, so `pg_advisory_xact_lock(hashtext(user:kind:signal))` — taken per signal, in sorted id order to avoid deadlocking overlapping merged suggestions — is held for the whole confirm. A double tap yields one entry; the ledger's `ON CONFLICT DO NOTHING` is only a backstop. Replay returns `200 {data, replay:true}` re-read through the same selection `createTimeEntry` returns, so the body is shape-identical to the `201`.

Run locally against a live Postgres (the worktree's `.env.test` database), **test counts read from the log**, because `rls-coverage.integration.test.ts` is excluded from `vitest.integration.config.ts` and passing it as a path there silently runs zero tests and still prints green.

| Suite | Result |
|---|---|
| `vitest.config.rls-coverage.ts` (**rls-coverage**) | ✅ 1 file, **75 tests** |
| `vitest.config.rls.ts` | ✅ 2 files, **26 passed / 5 skipped** |
| `timeSuggestionDecisionsRls.integration.test.ts` (new) | ✅ 1 file, **18 tests** |
| `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` + `orgMergeRegistry` (integration) | ✅ 4 files, **20 tests** |
| Full `vitest.integration.config.ts` | ⚠️ **not completed** — the local run was killed part-way. See below |
| API unit (`apps/api`) | ✅ 1636 files, **29071 passed / 22 skipped** |
| Web unit (`apps/web`) | ✅ 663 files, **6882 passed** |
| Shared (`packages/shared`) | ✅ 83 files, **2209 passed** |
| `tsc --noEmit -p apps/api/tsconfig.json` | ✅ clean |
| `astro check` (web) | ✅ 1721 files, 0 errors |
| `eslint src` (api, web) | ✅ clean |
| `@breeze/docs check` + `build` | ✅ 0 errors, 167 pages |
| `scripts/check-migration-naming.sh` | ✅ OK |

### About the full integration run

The whole-suite local run did not finish (killed part-way), so it is reported as **not-checked**, not as a pass. Two files had failed before it stopped, and both were chased down rather than waved through:

- `builtinCatalogVersionsRoute.integration.test.ts` (2 failures, one at 30s) — **passes 7/7 in isolation.** Cross-file DB state / timeout in a long sequential local run, not a defect.
- `patchCanonicalExportParity.integration.test.ts` (1 failure, `apps`/`autoApproveDeferralDays` come back `undefined` instead of the schema defaults) — **pre-existing.** Reproduced with `origin/main`'s `apps/api/src/routes/orgs.ts` checked back in over this branch's version; that file is the only one in this diff anywhere near the partner-settings export path. Unrelated to this wave.

The five contract suites the plan names were each run individually against the live database and each reported a real, non-zero test count (above). The blocking `integration-test` shards on this PR are the authoritative full run.

## Two gaps this session found

The branch arrived with 13 commits from earlier runs and one uncommitted file. Both were audited rather than trusted:

1. The uncommitted `partner-wide-write-coverage.test.ts` edit was **finished work, not a half-edit** — removing it and re-running the suite reproduces `services/timeSuggestionService.ts mutates timeSuggestionDecisions without calling canManagePartnerWidePolicies()` (1 failed / 4 passed); restoring it gives 5 passed. Committed with that evidence.
2. The **full** web suite was red: `common.json: 105 exact-English duplicates exceeds baseline 104` (de-DE, and +1 for it-IT). The earlier passes had only run the touched component files. Fixed by bumping the two reviewed baselines with the reason, rather than mistranslating a loanword.

A third, smaller correction: the release note claimed a client-sent `source` is *"rejected"*. It is not — `createTimeEntrySchema`/`updateTimeEntrySchema` are plain (non-strict) zod objects, so an unknown key is **stripped**. Reworded to say what actually happens.

## Deviations from the plan

- **Migration date.** Plan said `2026-09-23-…`; main had already shipped that date plus `2026-09-24-{a,b}-…`, so the file is `2026-09-25-…`. This is what the plan's own D8 instructs, and the alternative would have replayed in a different order on a fresh database than on a migrated one.
- **A fifth `source` value.** The location spec defined four; `support_session` was added for a confirmed Quick Support entry (which lands `org_id NULL`). The location spec has been amended in place with a dated reconciliation note.
- **Tasks 14–18 not implemented**, per the scope of this PR.
- Manual TestFlight checks (plan Task 18 Step 5) are **not-checked** — they require the mobile client, which is the other half.

## CI note

**"Check Migrations" (step *Shipped-migration immutability guard*) and "Trivy Image Scan" are already red on `origin/main` itself** (verified 2026-08-30 at `355b37465`) — pre-existing and unrelated to this branch. This PR targets `main`, so the blocking `integration-test` shards run normally; no hand-dispatch is needed.


ToddHebebrand:

Lenses: **code**, **security**, **tests**, **tenancy/contracts**. Every candidate finding was handed to two independent refuters with the instruction to *kill* it; only findings both refuters failed to refute are treated as confirmed. Round produced **15 raw findings, 3 confirmed, 3 refuted** (the rest folded into the suggestion list below).

All claims below are labelled **[verified]** (I read the code or ran the command), **[inferred]** (reasoned from verified facts) or **[not-checked]**.

---

### Fixed

**1. Published docs and release notes described a timesheet suggestion UI that ships in no surface in this PR** — *security lens, important*
`apps/docs/src/content/docs/features/ticketing.mdx`

[verified] The docs section and the appended `/timesheet` sentence described a suggestion list, a one-tap confirm and a dismiss/restore control on **/timesheet** — i.e. `apps/web/src/components/time/TimesheetPage.tsx`, whose entire diff in this PR is a read-only `source` badge. `git diff origin/main...HEAD --stat -- apps/mobile` is **empty**, so the documented UI exists in *no* surface here. The settings card *does* ship and *is* mountable, so an admin could enable the toggle, follow the doc to `/timesheet`, and find nothing.

Fixed by scoping the copy to what actually ships:
- A `<Aside type="caution" title="Mobile app only">` stating plainly that the suggestion list, confirm and dismiss/restore live in the Breeze mobile app and that **turning the setting on changes nothing a technician sees on /timesheet**.
- The `/timesheet` sentence rewritten to say the badge is what web shows and the suggestion list is a mobile screen.
- `docs/release-notes/next-release-draft.md` retitled to "**API and partner setting**" and rewritten to say there is no web suggestions panel.
- The settings-card `subheading` string retranslated across all 8 locales so the in-product copy makes the same promise as the docs.

**2. `mergeSignals` compared against the last-appended signal's `endedAt`, so a nested session split into a second overlapping suggestion** — *code lens, important*
`apps/api/src/services/timeSuggestionRules.ts:79`

[verified by executing the committed function in node] With `mergeGapMinutes=10` and three ended sessions on one device — A terminal 10:00–12:00, B desktop 10:05–10:10, C desktop 10:30–10:40 (a concurrent terminal + desktop pair on one device is ordinary) — the committed grouping returned `[[a,b],[c]]`. C lies *entirely inside* A's envelope yet opened a second group, so the endpoint emitted two **overlapping** suggestions and `unloggedCount: 2`.

[verified] Nothing downstream caught it: `dropAlreadyLogged` consults only existing `time_entries`, never sibling suggestions in the same response; and `confirmTimeSuggestion`'s guards are keyed on **signal id**, so `{A,B}` and `{C}` — disjoint id sets — never trip the partial-ledger 409. Confirming both from one loaded list (exactly the W07/W08 offline-queue drain, which replays queued confirms without refetching) wrote two billable entries both covering 10:30–10:40.

Fixed by carrying a running `groupEnd = max(groupEnd, row.endedAt)` and measuring the gap against that, never against the tail row.

*Red-first, [verified]:* against the pre-fix function the two new cases fail (`expected [[a,b],[c]] to deeply equal [[a,b,c]]`); with the fix, `46 passed (46)`. A third case pins that the group still *splits* when the true group end is beyond the gap, so the fix cannot silently over-merge.

**3. No test discriminated the `currencyCode == null` gate that stops an org-linked entry being re-denominated in the partner currency** — *tests lens, important*
`apps/api/src/services/timeEntryService.ts:445`

[verified] The production gate is correct — **this was a missing-test finding, not a live bug** — but the PR body's claim that "five service tests cover exactly those branches" was false. Every `orgLink` create test passed **no `hourlyRate`** (so `input.hourlyRate != null` was false and the guard was never evaluated), and every test that *did* pass an `hourlyRate` had no ticket and no orgLink (so `currencyCode` was already null and the new clause was a no-op). [verified] The branch is reachable in production: `confirmSuggestionSchema` accepts `hourlyRate` and `confirmTimeSuggestion` forwards it to `createTimeEntry` alongside `provenance.orgLink`.

Fixed by adding a case that passes `hourlyRate: 90` with `orgLink { orgId:'o1', currencyCode:'EUR' }`, queues a partners row returning `USD`, and asserts both that the row lands `currencyCode: 'EUR'` **and** that `getPartnerCurrency` was never consulted.

*Red-first, [verified by mutation:* deleting `currencyCode == null` from line 445 fails **only** the new test, with exactly the money symptom — `- "currencyCode": "EUR" / + "currencyCode": "USD"` — while the five pre-existing tests stay green, which is the finding restated as a measurement.]

#### Suggestions also applied

- **A distinct code for the partial-ledger 409.** The `decisions.length > 0` fall-through reused `SUGGESTION_DISMISSED`, pointing a client at a "restore it first" remediation that cannot work. Added `SUGGESTION_PARTIALLY_LOGGED` to `TimeEntryServiceErrorCode`, the throw site and the route status-map test, and tightened the existing partial-ledger service test from `{status: 409}` to assert the **code**. [verified] No other consumer references the old code (`grep` across `apps/`, `packages/`, `docs/`).
- **Merged dismiss/undismiss no longer writes a concatenated id.** `entryId: signals.map(s => s.id).join('+')` produced `resource_id = NULL` (the audit writer stores `resource_id` only when the value parses as a uuid) plus a `details.entryIds` array whose one element was not an id. Now one audit mutation **per signal**. *[verified by mutation:* restoring the joined form fails the new test — `expected "vi.fn()" to be called 2 times, but got 1 times`.]
- **The `/suggestions` registration-order guard could not fail.** It asserted `status !== 404` and "`updateTimeEntry` not called" — but there is no `GET /:id` to swallow it, `updateTimeEntry` is a PATCH-only mock, and the unmocked service returned 500, which also satisfies `not.toBe(404)`. Replaced with two positive assertions (`GET /suggestions` → 200 + `listTimeSuggestions` called; `POST /suggestions/confirm` → 201 + `confirmTimeSuggestion` called and `createTimeEntry` **not**). *[verified by mutation:* deleting the `route('/suggestions', …)` mount reddens both.] The source comment was corrected too — it claimed a hazard (`PATCH /:id` swallowing `/suggestions/confirm`) that is unreachable at today's route shape, since `/:id` is one segment and `/suggestions/confirm` is two.
- **The deadlock-avoiding sorted lock order is now tested.** Both existing multi-signal fixtures were *already sorted*, so no assertion on them could discriminate `[...ids].sort()` — a uniform fixture hiding the branch. Added a case with a deliberately unsorted `[s2, s1]`. *[verified by mutation:* replacing `[...ids].sort()` with `ids` fails exactly that test.]
- **The Quick-Support-with-a-ticket confirm branch now has a test**, pinning both halves that were incidental: the `support_session` stamp, and the *deliberate* absence of the org-agreement check (a QS session has no real org, so `createTimeEntry`'s own ticket path is what authorises it).
- **Two comments that asserted guarantees the code does not make** were corrected rather than left to mislead a future reader:
  - `SuggestionActor.scope` — the JSDoc said the service uses it to decide DB-context visibility. [verified: `grep` finds no read of `actor.scope` in the service] The narrowing is entirely `requireScope('partner','system')` at the router plus explicit `o.partner_id = $n` / `rs.user_id = $n` predicates. Comment now says so.
  - `countUnloggedSuggestions` — the docstring justified `withSystemDbAccessContext` on predicates that hold for `loadSignals` but **not** for `loadLoggedRanges`, whose only predicate is `te.user_id`. It is safe today only because `users` is single-partner — an invariant of another table. The docstring now names that, plus the `runOutsideDbContext` requirement (the system wrap is a **no-op inside a request**, where `accessibleOrgIds: null` would then run under the caller's narrower policy). No live defect: [verified] nothing calls this function in this wave; it is a W07 trap, now labelled.
- **`suggestionsQuerySchema` `.strict()` vs `fetchWithAuth`'s automatic `?orgId=`.** Documented at the schema that a web caller **must** pass `skipOrgIdInjection`, since suggestions are user-scoped and there is no `orgId` to honour.
- **Release note now names `time_entries` as a second lock.** The validated `time_entries_source_chk` holds ACCESS EXCLUSIVE and scans every row (and the `DROP … IF EXISTS` + re-add means a re-run repeats the scan). The note previously called out only the `remote_sessions` index build and said "no data migration is required", which read as though that were the only lock to plan around.

---

### Declined

- **Splitting the CHECK into `NOT VALID` + a follow-up `VALIDATE CONSTRAINT`.** Correct in general, but it trades a documented one-time scan for a two-migration dance on a table that is small on every instance we run, and it would leave the constraint unvalidated between releases. Took the other option the finding offered: extend the release note. Reconsider if a self-hoster reports a long upgrade.
- **Adding `orgId: z.string().guid().optional()` to `suggestionsQuerySchema`.** `.strict()` is deliberate and pinned by a test (a typo'd param must 400, never silently widen the day). Accepting-and-ignoring an `orgId` on a user-scoped endpoint would be a small lie in the contract. Documented the `skipOrgIdInjection` requirement instead — no caller exists yet, so the note lands before the first one.

---

### Refuted (raised, investigated, **not** applied)

- **"`deleting the partner CASCADEs the decisions away` never deletes a partner."** The test-name inaccuracy is real, but the claimed failure cannot occur. [verified] `cascadeDeletePartner` does not rely on `ON DELETE CASCADE` at all: it discovers partner-axis tables from `information_schema.columns WHERE column_name='partner_id'`, topologically orders them, and issues an explicit `DELETE FROM "<table>" WHERE partner_id = $1` per table *before* `DELETE FROM partners`. Flipping the FK to `NO ACTION` would raise no 23503 and strand no tenant. The load-bearing coverage is `tenantCascadePartner.integration.test.ts`, which drives the real function end-to-end. Wording nit, not a defect.
- **"The entire write half has zero real-database coverage."** The coverage observation is accurate, but the failure scenario is counterfactual — it is premised on the advisory lock *not* serialising. [verified] `withDbAccessContext` runs its callback inside `baseDb.transaction(...)`, so `pg_advisory_xact_lock` in `lockSignals` holds to COMMIT and the second concurrent confirm blocks, then takes the `{replay:true}` 200 branch. And the invariant the finding says is untested at DB level **is** tested at DB level: `timeSuggestionDecisionsRls.integration.test.ts` asserts a duplicate `(user_id, signal_kind, signal_id)` raises 23505 with exactly one row surviving. Hardening, not a defect.
- **"The cross-technician test never asserts the user id on the F19 already-logged query."** [verified] Production code is correct — `listTimeSuggestions` resolves `const userId = opts.userId ?? actor.userId` and passes *that* to `loadLoggedRanges`, so a dispatcher's own logged day can never subtract from the technician's sessions. The claimed empty-list/unbilled-day outcome does not occur. It was a genuine mutation-coverage gap, so the two missing assertions on `compiled(1).params` were added anyway — as hardening, filed here rather than under Fixed because nothing was broken.

---

### Strengths

Worth recording, because these are the rows this repo gets wrong most often and every one of them checks out **against the code, not against the plan**:

- **Registration lists are complete, and every "does not apply" argument in the PR body holds up.** [verified] `time_suggestion_decisions` has no `org_id` (migration lines 28–40; `apps/api/src/db/schema/timeTracking.ts:88-98`), and `CORE_ORG_CASCADE_DELETE_ORDER` (`tenantCascade.ts:65`) enumerates `org_id`-columned tables only — adding it would make `cascadeDeleteOrg` issue `DELETE … WHERE org_id` against a table with no such column. No `device_id`, so the two lists in `routes/devices/core.ts` correctly stay untouched. Not append-only (no `REVOKE DELETE`, no immutability trigger), so `AUDIT_ADMIN_REQUIRED_TABLES` correctly stays untouched. The orgMergeRegistry claim is exactly right: `orgMergeRegistry.integration.test.ts:237` builds `required` from `getOrgCascadeDeleteOrder() + EXTRA_REQUIRED` and asserts "no policy names an unrequired table" (lines 245–247) — adding an entry **would have failed** the suite.
- **The new COLUMN on a long-registered table is classified — the row this repo misses most.** [verified] `source` is in the `time_entries` `included` bucket (`tenantExportPolicyRegistry.ts:367`). Correct bucket: a `varchar` provenance value, not a credential, not a `json`/`jsonb`/`bytea` open container. I also confirmed the checker is **set-based, not order-sensitive** (`apps/api/scripts/check-tenant-export-policy.ts:38-45`), so placing `source` before `is_approved` is fine even though `ALTER TABLE ADD COLUMN` appends it last by `ordinal_position` on a freshly migrated DB — and that the export-policy suite keys its required-table set off `getOrgCascadeDeleteOrder()`, which is why the new *table* correctly needs no entry.
- **Partner cascade needs no edit, and the stated reason is the real one.** [verified] `cascadeDeletePartner` discovers partner-axis tables from `information_schema` and topo-sorts them (`tenantCascade.ts:1095-1108`) — there is no hand-maintained partner list to forget. The `partner_id` FK is `ON DELETE CASCADE` and every other FK on the new table carries an explicit `ON DELETE` (`users` CASCADE, `time_entries` SET NULL), so neither the topological sweep nor the org cascade's `time_entries` DELETE can raise a NO-ACTION FK violation against it.
- **RLS is enabled, forced and policied in the creating migration, in the same shape as the sibling table.** [verified] `2026-09-25-…sql:48-62` is byte-for-byte the same policy predicate as `time_entries_partner_access` (`2026-06-12-a-ticketing-time-parts.sql:60-64`): `FOR ALL TO breeze_app`, `breeze_current_scope() = 'system' OR breeze_has_partner_access(partner_id)` on **both** `USING` and `WITH CHECK`, wrapped in a `pg_policies` existence check, with an explicit GRANT for `breeze_app`.
- **Migration hygiene is correct on every axis I checked.** [verified] Re-diffed against `origin/main` after the review (main has since added migrations through `2026-09-24-b`): `2026-09-25-…` still sorts last, it does not touch the closed `2026-08-06` block, there is no inner `BEGIN`/`COMMIT`, and every statement is idempotent (`ADD COLUMN IF NOT EXISTS`, `DROP CONSTRAINT IF EXISTS` then re-add, `CREATE TABLE/INDEX IF NOT EXISTS`, `pg_policies` guard).

---

### Contract-suite outcomes

Verified each suite **actually ran** by reading its test count — a green print with zero tests is not a pass.

| Suite | Result |
|---|---|
| `tsc --noEmit` (apps/api) | **clean**, exit 0, zero diagnostics (needed `--max-old-space-size=12288`; the default heap OOMs — known repo trap) |
| API unit suite (`vitest run`, full) | **29079 passed \| 22 skipped (29101)**, 1636 files, exit 0 |
| `scripts/check-migration-naming.sh` | `check-migration-naming: OK` |
| `autoMigrate` + `composeBindMounts` + `cascadeDelete` + `moveOrg.coverage` | **91 passed (91)**, 4 files |
| `timeSuggestionRules.test.ts` | **46 passed (46)** |
| `timeSuggestionService.test.ts` | **50 passed (50)** |
| `timeEntryService.test.ts` | **115 passed (115)** |
| Web i18n parity + `TimeTrackingSettingsCard` + `time/` | **155 passed (155)**, 10 files |
| `packages/shared` | **2209 passed (2209)**, 83 files |
| `astro check` (apps/docs) | **0 errors, 0 warnings, 0 hints** |
| `pnpm lint` | **6 successful, 6 total** |
| RLS + integration contract suites | **[not-checked] locally** (no live Postgres in this worktree) — **[verified] green in CI** on `5ac0d8796`, and I read the shard logs for the counts rather than trusting the green: `timeSuggestionDecisionsRls.integration.test.ts` **18 tests** (shard 3), `tenantCascade.integration.test.ts` **5 tests** (shard 4), `tenant-export-policy.integration.test.ts` **2 tests** (shard 1); **105 test files passed** per shard. |

**[verified] Full CI on `5ac0d8796`: 58 pass, 0 fail, 1 skipping** (`Main Red Alert`, a main-only job). All four blocking `Integration Tests` shards SUCCESS, plus `Test API`, `Test Web`, `Type Check` and `CI Success`. Worth noting explicitly: **`Check Migrations` and `Trivy Image Scan` — the two jobs currently red on `main` itself — both PASS on this PR**, so nothing here needs the usual "known main-red, ignore it" caveat.

One more [verified] note for anyone reading a red local run on this branch: an API full-suite run executed *concurrently with `tsc`* reported six failures (`migrationOrdering`, `offsetPaginationOrdering`, `revocationCallsites`, `aiToolNames`, `ipAllowlistMode.config`, `quotePdf.classicRegression`). All six are **timeout artifacts of CPU contention**, not defects — re-run in isolation they pass (`3 passed (3) / 11 tests` and `3 passed (3) / 6 tests`), and the clean full run above is green. Worth knowing because `migrationOrdering` failing on a branch that adds a migration looks alarming and is not.

**[verified] Trap worth recording for the next agent on this feature:** `rls-coverage.integration.test.ts` is *excluded* from `vitest.integration.config.ts` and has its own runner (`vitest.config.rls-coverage.ts`). Passing it as a path to the integration config runs **zero tests and still prints green**. Always read the count.

All of the above is at **`5ac0d8796`** (`fix: address W06A review findings (#3206)`), pushed to this PR.

PR #4292, opened 2026-08-31
fix(api): guard created_by at the queueCommand insert site (#3978)

### LINKED ISSUE #3978: [API] queueCommand stamps created_by unguarded — agent-released intents will fail with 23503 after approval
## Description

`executeCommand` was hardened in wave 3b so that a synthetic-auth id which is not a `users` row degrades `device_commands.created_by` to NULL. Its sibling insert site, `queueCommand` / `queueCommandForExecution`, still stamps `created_by` verbatim — so any tool that dispatches through it will hit the same FK violation the 3b fix was written to prevent.

This is the nastiest failure shape in the approval flow: it fires **after** a human has already approved the intent, at execution time, not at proposal time. The operator sees an approved action that then dies with a raw Postgres error.

The wave-3b author flagged this explicitly in a code comment as a follow-up for PR 3c, but 3c and 3d both shipped without it, so it is now unowned. Filing it so it does not get lost.

## Root Cause

`ai_agent` auth sets `auth.user.id` to the agent's `ai_agents` id, which is not a `users` row. Tool handlers pass `auth.user.id` through verbatim.

`apps/api/src/services/commandQueue.ts:927` (guarded, wave 3b):

```ts
createdBy: safeUserId,   // users-existence probe; non-user ids degrade to NULL
```

`apps/api/src/services/commandQueue.ts:492` (unguarded):

```ts
createdBy: userId || null,
```

Tools dispatching through the unguarded path, per the in-code note: hyperv, backup, vault, mssql, incident, agent-logs.

## Impact

Latent today — no agent can run in production while `BREEZE_AI_AGENTS_ENABLED` is off, and the wave-3 triage allowlist does not include these tools. It becomes live the moment an agent is allowed to propose one of them, which is squarely in scope for wave 4 (#3826, act mode).

## Proposed Fix

Apply the same treatment as `executeCommand`: resolve the id through the users probe at the `queueCommand` insert, degrading to NULL when it does not resolve. Better still, extract the probe into one helper both insert sites call, so a third insert site cannot drift again.

Add a test that dispatches one of these tools under an `ai_agent` AuthContext and asserts the row lands with `created_by IS NULL` rather than raising 23503.

## Affected Files

- `apps/api/src/services/commandQueue.ts:492` (primary), `:927` (existing guarded site to factor against)

## Reported By

Internal — flagged in-code during wave 3b (#3931), still outstanding after 3c (#3953) and 3d (#3976).

ToddHebebrand:
## Problem

`queueCommand` stamped `createdBy: userId || null` verbatim. Several synthetic-auth classes reach it with an `auth.user.id` that is **not** a `users` row, so the insert died on `device_commands_created_by_users_id_fk` with **SQLSTATE 23503**.

The worst case is `ai_agent`: `buildAgentAuthContext` sets `user.id` to the agent's `ai_agents` id, and the intent release worker runs **approved** agent intents through the same tool handlers the human paths use. So the FK violation fires **after a human has already approved the action**, at execution time, surfacing as a raw Postgres error.

`executeCommand` was hardened in wave 3b (#3931) and its in-code comment explicitly named `queueCommand`/`queueCommandForExecution` as the sibling site needing the same treatment "before the runner releases agent intents" for the hyperv/backup/vault/mssql/incident/agent-logs tools. 3c (#3953) and 3d (#3976) both shipped without it.

## What `created_by` holds per principal

Established by reading each `AuthContext` builder, not assumed:

| Principal | `auth.user.id` | Real `users` row? | `created_by` after this PR |
|---|---|---|---|
| `user_session` | `users.id` (live lookup) | yes | that user id |
| `api_key` | `apiKeys.createdBy` (NOT NULL FK to `users.id`) | yes | that user id |
| `oauth_grant` | JWT `sub`, live-validated against `users` | yes | that user id |
| `ai_agent` | `ai_agents.id` | **no** | `NULL` ← the reported bug |
| `helper` | `devices.id` | **no** | `NULL` (short-circuit, no DB read) |
| `client_user` | `portal_users.id` | **no** | `NULL` |
| `system` | `'system'` literal / nil-UUID | **no** | `NULL` |

So yes — `created_by` **must** be nullable for non-human origins. It already is: `created_by uuid` with no `NOT NULL` (baseline `0001-baseline.sql`), and an FK always admits NULL. **No migration needed**, and no new table or column, so no cascade/export-policy registration applies.

## Fix

Extract the probe into one exported helper, `resolveCommandCreatedBy`, called from **both** insert sites so a third cannot drift again (the issue's explicit ask). `queueCommandForExecution` has no insert of its own — it delegates to `queueCommand` — so this single change covers all ~50 call sites, including two agent-reachable paths the issue did not name: `aiToolsVulnerability` → `vulnerabilityRemediation.ts:409` and `aiToolsDR` → `drExecutionService.ts:392`.

### The load-bearing detail: the probe opens its own system context

`users` is RLS-protected:

```sql
breeze_has_partner_access(partner_id)
OR (org_id IS NOT NULL AND breeze_has_org_access(org_id))
OR id = breeze_current_user_id()
```

`withSystemDbAccessContext` alone is **not** enough — `withDbAccessContext` short-circuits when a context store already exists (`if (dbContextStorage.getStore()) return fn()`), so inside a caller's context the probe would silently run under the **caller's** scope. Every BullMQ worker and agent run dispatches under an org-scoped context with `userId: null` (`agentDbAccessContext`), where a partner-level user (`org_id IS NULL`) matches no branch of that policy — and a contextless worker call matches none at all.

A probe there reads zero rows and degrades a **real human** to `created_by NULL`, silently destroying attribution *while still passing an agent-only test suite*. So the helper exits the caller's context first via `runOutsideDbContext` (which clears both stores), letting the nested system context be genuinely fresh. This mirrors the audit block directly below it, which escapes the caller's context for the same reason.

The dispatch metric and audit row now key on the **resolved** id rather than the raw one, matching `executeCommand` — previously a non-user id wrote an audit row claiming `actorType: 'user'`.

## Tests

**Red first.** `commandQueueCreatedBy.integration.test.ts` (new, real Postgres) reproduced the exact bug before the change — 4 failed / 5 passed, with `code: '23503'`, `constraint_name: 'device_commands_created_by_users_id_fk'`, `Key (created_by)=(…) is not present in table "users"`. All 9 pass after.

It deliberately covers both halves. The human-attribution cases are the discriminating ones — they are green before **and** after, and a naive context-less fix turns them red:

- ai_agent id → `NULL`, no 23503 (bare, and inside the agent run's own org-scoped context)
- helper session / unknown uuid / no userId → `NULL`
- org-scoped human → keeps their id
- **partner-level human (`org_id IS NULL`) from a contextless worker dispatch** → keeps their id
- **partner-level human inside an org-scoped, user-less context** → keeps their id

Unit tests in `commandQueue.test.ts` pin the contract in the fast job: what actually lands in `.values()` (captured, not a hand-shaped fixture), that the probe predicate is real (compiled via `PgDialect().sqlToQuery` — asserting `"users"."id"` and the bound param, so a helper probing a constant cannot pass), and that the probe nests `runOutsideDbContext` → `withSystemDbAccessContext` → read.

### Verification

- New integration suite: 9/9 green (was 4 red pre-fix).
- `agentIntentLifecycle.integration.test.ts` (covers the refactored `executeCommand` probe): green.
- 60 affected unit files, 1092 tests: green (batched, threads pool).
- `tsc --noEmit` on `apps/api`: clean. `eslint` on changed files: clean.

## Follow-up (not in this PR)

`services/scriptDispatch.ts:355-368` is a **third**, independently-written copy of this probe. It is already FK-safe so it is unaffected by the reported crash, but it shares the nested-context gap described above (it calls `withSystemDbAccessContext` without escaping the caller's context first). Its dual-column/`degradedActorId` semantics differ enough that folding it into the shared helper is its own change with its own test surface — flagging rather than bundling it.

Closes #3978

### DIFF
--- a/apps/api/src/__tests__/integration/commandQueueCreatedBy.integration.test.ts
+++ b/apps/api/src/__tests__/integration/commandQueueCreatedBy.integration.test.ts
@@ -1,0 +1,242 @@
+/**
+ * Real-Postgres coverage for `device_commands.created_by` attribution at the
+ * `queueCommand` insert site (#3978).
+ *
+ * `queueCommand` used to stamp `createdBy: userId || null` verbatim. Two
+ * synthetic-auth classes reach it with a `userId` that is NOT a `users` row —
+ * `ai_agent` principals (`buildAgentAuthContext` sets `auth.user.id` to the
+ * agent's `ai_agents` id) and helper sessions (`auth.user.id` IS the device
+ * id) — so the insert died on the `device_commands_created_by_users_id_fk`
+ * foreign key with SQLSTATE 23503. For an agent-released intent that lands
+ * AFTER a human approved the action, at execution time.
+ *
+ * Only a real database can prove this: the FK lives in Postgres, so a mocked
+ * unit suite can assert the value handed to `.values()` but can never observe
+ * the 23503 the guard exists to prevent. Hence this suite.
+ *
+ * The `human attribution` tests are not padding — they are the discriminating
+ * half. The probe reads `users`, which is RLS-protected:
+ *
+ *   breeze_has_partner_access(partner_id)
+ *   OR (org_id IS NOT NULL AND breeze_has_org_access(org_id))
+ *   OR id = breeze_current_user_id()
+ *
+ * A guard that probes inside the CALLER's context (the obvious fix —
+ * `withSystemDbAccessContext` alone is a no-op when a context is already open,
+ * because `withDbAccessContext` short-circuits on an existing store) sees zero
+ * rows for a partner-level user (`org_id IS NULL`) whenever the caller context
+ * is org-scoped with `userId: null` — exactly the shape every BullMQ worker and
+ * agent run uses. That fix would silently degrade a REAL human to
+ * `created_by NULL` and pass an agent-only test suite.
+ *
+ * Run:
+ *   pnpm test-stack up            # from repo root
+ *   cd apps/api && pnpm exec vitest run --config vitest.integration.config.ts \
+ *     src/__tests__/integration/commandQueueCreatedBy.integration.test.ts
+ */
+import './setup';
+
+import { randomUUID } from 'node:crypto';
+import { beforeEach, describe, expect, it } from 'vitest';
+import { eq } from 'drizzle-orm';
+
+import { db, withDbAccessContext, withSystemDbAccessContext } from '../../db';
+import { aiAgents, deviceCommands, devices } from '../../db/schema';
+import { CommandTypes, queueCommand } from '../../services/commandQueue';
+import { createOrganization, createPartner, createSite, createUser } from './db-utils';
+
+// `list_processes` is deliberately NOT in AUDITED_COMMANDS, so queueCommand
+// takes its short path: the users probe is the only SELECT and no audit row is
+// written. That keeps each assertion about created_by and nothing else.
+const UNAUDITED_TYPE = CommandTypes.LIST_PROCESSES;
+
+interface Fixture {
+  partnerId: string;
+  orgId: string;
+  /** A conventional org-scoped human: `users.org_id` is set. */
+  orgUserId: string;
+  /**
+   * A partner-level human (`users.org_id IS NULL`) — an MSP tech operating
+   * inside a customer org. Invisible to an org-scoped RLS probe that has no
+   * `breeze.user_id` set, which is what makes it the discriminating fixture.
+   */
+  partnerUserId: string;
+  deviceId: string;
+  /** An `ai_agents` row id — a valid uuid that is NOT a `users` row. */
+  aiAgentId: string;
+}
+
+async function seed(): Promise<Fixture> {
+  const partner = await createPartner();
+  const org = await createOrganization({ partnerId: partner.id });
+  const site = await createSite({ orgId: org.id });
+
+  const orgUser = await createUser({
+    partnerId: partner.id,
+    orgId: org.id,
+    email: `org-user-${randomUUID()}@example.test`,
+  });
+  const partnerUser = await createUser({
+    partnerId: partner.id,
+    orgId: null,
+    email: `partner-user-${randomUUID()}@example.test`,
+  });
+
+  const unique = randomUUID().slice(0, 8);
+  const [device] = await withSystemDbAccessContext(() =>
+    db
+      .insert(devices)
+      .values({
+        orgId: org.id,
+        siteId: site.id,
+        agentId: `test-agent-${unique}`,
+        hostname: `test-host-${unique}`,
+        osType: 'linux',
+        osVersion: '22.04',
+        architecture: 'x86_64',
+        agentVersion: '0.0.0-test',
+        status: 'online',
+      })
+      .returning(),
+  );
+
+  const [agent] = await withSystemDbAccessContext(() =>
+    db
+      .insert(aiAgents)
+      .values({
+        orgId: org.id,
+        partnerId: null,
+        kind: 'triage',
+        name: `Test Agent ${unique}`,
+        createdBy: orgUser.id,
+      })
+      .returning(),
+  );
+
+  if (!device || !agent) {
+    throw new Error('fixture seeding failed: device and ai_agent rows are required');
+  }
+
+  return {
+    partnerId: partner.id,
+    orgId: org.id,
+    orgUserId: orgUser.id,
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — two rounds.
Round 1 on `d5b6e14` (code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer). Round 2 on `95104da` (code-reviewer, pr-test-analyzer, silent-failure-hunter), because a third commit landed that round 1 never saw.

**Findings:** 6 raised across both rounds → all 6 addressed; 0 outstanding.

Round 1 → fixed in `b58a17e`:
1. *Flaky test* (code-reviewer, reproduced twice). The fire-and-forget audit block shares the single-use wrapper mocks, but the drain was a fixed 3 microtasks — an early return both failed the assertion and leaked unconsumed `mockImplementationOnce` entries into later tests. Replaced with `vi.waitFor` on the full call order.
2. *Inaccurate comment* (comment-analyzer). "Every BullMQ worker dispatches under an org-scoped context with `userId: null`" was false — those workers wrap in `withSystemDbAccessContext`, and a contextless call opens a genuine system context, so both are incidentally safe. The genuinely vulnerable shape is an already-open **org-scoped** context, which is what the AI-tool handlers use. Also re-pointed the citation: `agentDbAccessContext` has no production callers; the real builder is `dbAccessContextFromAuth`.
3. *"Single source of truth" overstated.* `services/scriptDispatch.ts` carries an independent copy of this probe (FK-safe, so unaffected by this crash, but not wired to the helper). Now stated in the docstring.
4. *Degrade was invisible* (silent-failure-hunter) → added a warn.

Round 2 → fixed in `71fa02b`:
5. *That warn was the wrong instrument* (silent-failure-hunter). For an `ai_agent` principal the degrade is the **designed** outcome, so it fired on every agent-issued command as the normal case — the cry-wolf shape that buried `db/index.ts`'s contextless-write reporter, and `console.warn` never reaches Sentry. Distinguishing expected from anomalous needs the caller's principal kind, which `queueCommand` does not receive. Removed: the degrade is already observable because `dispatchActor` keys on the resolved id, so a degraded dispatch counts as `actor="system"` rather than `actor="user"` on the existing `commandsDispatchedTotal`.
6. *Test-mock hygiene* (pr-test-analyzer) — confirmed via `@vitest/spy` source that `clearAllMocks` does not drain once-queues; the construction self-drains via `vi.waitFor`, and the one permanent `db.select` impl is now explicitly re-mocked per test.

Round 2 verdicts on the rest: code-reviewer **no Critical or Important findings**; pr-test-analyzer confirmed the 14-step ordering assertion catches an unwrapped insert in the **fast unit job** rather than only in an integration shard.

**CI caught a real bug mid-review**, worth calling out: Integration shard 2 failed with `DB write .insert() ran with no RLS access context (#1375)`. `queueCommand`'s `device_commands` insert had always been a bare-pool write, while `executeCommand`'s sibling insert was wrapped for exactly this. It passed locally only because the tripwire throws solely when `DB_CONTEXTLESS_WRITE_STRICT=true`, which `ci.yml` sets for the integration job. Fixed in `95104da`; re-verified locally under that flag.

**Tests:**
- New real-Postgres suite `commandQueueCreatedBy.integration.test.ts` — **red first**: 4/9 failed pre-fix with `code: '23503'`, `constraint device_commands_created_by_users_id_fk`. Now 9/9.
- `agentIntentLifecycle.integration.test.ts` (covers the refactored `executeCommand` probe) green; 16/16 with both strict tripwires on.
- `commandQueue.test.ts` 35/35, stable over 6+ consecutive runs.
- ~60 affected unit files / ~1090 tests green, batched.
- `tsc --noEmit` clean; `eslint` clean.

**CI:** 55/55 pass on `71fa02b` (1 skipped is `Main Red Alert`, main-only). All four integration shards green.

**Status:** review-clean, awaiting maintainer merge. Not merged, and #3978 left open per the runbook.

**One follow-up deliberately not bundled:** `services/scriptDispatch.ts:355-368` holds a third, independently-written copy of this probe. It is FK-safe today so it is unaffected by the reported crash, but it calls `withSystemDbAccessContext` **without** escaping the caller's context first, so it carries the same latent attribution bug this PR fixes here. Its dual-column/`degradedActorId` semantics differ enough that converging it needs its own change and test surface.

PR #4294, opened 2026-08-31
fix(web): stop a deferred prop-sync effect from discarding an invoice notes edit (#3277)

### LINKED ISSUE #3277: [Flaky] InvoiceWorkspace queued-Issue tests fail intermittently in CI (Test Web)

ToddHebebrand:
Closes #3277.

## The flake, and why four previous attempts didn't hold

`InvoiceWorkspace.test.tsx`'s two queued-Issue tests have reddened at least six PRs and been filed five times since 2026-07-29 — #2925, #3219, #3277, #3980, #4033 — always on PRs touching no billing code, always green on a same-commit rerun. Two fixes shipped and neither held:

| | what it did | why it didn't hold |
|---|---|---|
| #3284 | raised `asyncUtilTimeout` 1000 → 5000ms | diagnosed "the propagation chain outruns the budget under load". It doesn't. |
| #3956 | pinned a 15000ms per-test ceiling above it | correctly restored a *legible* failure, and said so — it explicitly did not claim to fix the race. |

Both were built on the theory that these tests await the end of a multi-hop chain (PATCH settles → editor reports → workspace clears `savePending` → header un-gates → queued Issue fires) and that a loaded runner outran the budget. The ~1038ms failures looked like it.

That theory is wrong, and #3956's own evidence is what disproves it: the recurrence exhausted the **full 5000ms**. A chain that is merely slow finishes in five seconds. This one never started.

## Deterministic repro

A throwaway harness ran the failing scenario 200× in one file (startup dominates cost) under ambient load ~22–46 on a 14-core box, with the component instrumented. On a failing iteration:

```
--- editor visible; about to fireEvent.change ---
NOTES-ONCHANGE v="Edited"        <- change handler set notes='Edited', notesDirty=true
SYNC-EFFECT-RUN notes=""         <- the MOUNT-time prop-sync effect flushed AFTER, wiping both
--- change dispatched; about to fireEvent.blur ---
SAVE-NOTES-CALLED dirty=false notes=""   <- short-circuits; NO PATCH is ever dispatched
```

`afterChange=""` — the textarea is already empty *immediately after* `fireEvent.change({value:'Edited'})`, and `patchCalls=0`.

## Root cause

`InvoiceEditor.tsx` re-synced its free-text drafts from the server value in an effect:

```ts
useEffect(() => { setNotes(invoice.notes ?? ''); setNotesDirty(false); }, [invoice.notes]);
```

That effect also runs **on mount**, where it is normally a redundant no-op — the `useState` initialisers already hold the server value. But a passive effect is **deferred**: React commits the editor and only schedules the flush for a later task. Testing Library's `waitFor` resolves on the commit (MutationObserver), so it can return while the mount effect is still pending. The next `act()` — inside `fireEvent.change` — applies the keystroke *first* and then flushes the stale mount effect, which overwrites it with the pre-edit value and clears `notesDirty` with it.

`saveNotes()` short-circuits on `!notesDirty`, so the blur dispatches **no PATCH at all**. `savePending` never goes true, so `invoice-issue-saving-hint` never renders.

**No timeout is large enough for a condition that can never become true.** That is the whole reason both attempts at enlarging the budget failed.

This is not purely a test artifact: the same deferred write can land after a real keystroke and discard an edit with no request, no error and no "unsaved" cue. The window is one scheduler task, so a human rarely hits it — but it is the same defect.

## The fix

Evaluate the same condition **during render** (React's documented "adjusting state when a prop changes" pattern) instead of from an effect. There is then no deferred write that can land after a local edit, and the mount pass is a genuine no-op because the synced value starts equal to the server value. Behaviour is otherwise preserved: a server value that genuinely changes still replaces the local draft.

One deliberate improvement: the comparison is on the **normalised** strings, so a `null → ''` round-trip no longer re-runs the sync and clobbers a draft. The old raw-prop dependency did.

### Timeouts reverted, deliberately

Both inflations are removed and the file is back on Testing Library's defaults. With the race gone, the saving hint is present **synchronously inside the blur's own `act()` flush** — measured present-before-any-poll on 200/200 runs — so the default budget is not merely sufficient, it is unused. Keeping 5000ms/15000ms would only buy a slower, less legible failure for every genuine future breakage here. This is the opposite of widening a timeout, and it is backed by measurement rather than assumption.

The two queued-Issue tests now assert their preconditions directly (the edit stuck; a PATCH was dispatched) before the wait that depends on them, so a regression reports *"the edit was discarded"* by name and instantly, rather than exhausting a wait on a hint that could never have rendered.

## Evidence

**A/B under identical ambient load (~22–46), Testing Library default 1000ms budget, 200 iterations each:**

| arm | result |
|---|---|
| pre-fix `InvoiceEditor.tsx` | **2 failed / 200** |
| fixed `InvoiceEditor.tsx` | **0 failed / 200** |

Arm B's file was confirmed by md5 against the saved fixed copy, not assumed.

**Stability after the fix — 50 consecutive process runs, all green:**

```
# 25 runs, 7/7 each
pnpm exec vitest run --pool=threads --maxWorkers=2 src/components/billing/InvoiceWorkspace.test.tsx
#   -> pass=25 fail=0 stall=0

# 25 runs, 32/32 each
pnpm exec vitest run --pool=threads --maxWorkers=2 \
  src/components/billing/InvoiceWorkspace.test.tsx src/components/billing/InvoiceEditor.test.tsx
#   -> pass=25 fail=0 stall=0
```

(`--pool=threads` because the default forks pool cannot start workers at all on this loaded host — it reports `no tests`, which is a stall, not a pass. CI is unaffected.)

**Regression tests — red-first, verified against the pre-fix implementation:**

Three cases added to `InvoiceEditor.test.tsx`. The `null → ''` pair is the deterministic discriminator; both fail on the old code and pass on the new:

```
x keeps a dirty notes draft when a refetch reports the same value under a different raw form (null -> "")
x keeps a dirty terms draft across the same null -> "" round-trip
```

Each of those two blurs the field after the re-sync and asserts the **PATCH actually goes out carrying the drafted text** — not merely that the text is still on screen. That distinction came out of review: the visible value is only a proxy, and the damage in #3277 was the silently-cleared dirty flag, so a regression that cleared the flag without touching the value would have passed a text-only assertion.

The third (`DOES replace the draft when the server value genuinely changes`) passes on both — it pins the behaviour that is deliberately retained.

**On reverting the timeouts specifically:** the 50 consecutive runs above execute *all seven* tests in `InvoiceWorkspace.test.tsx` at the default budget, not just the two queued-Issue cases — including `updates the header from "Draft invoice"…`, which has its own multi-hop async chain. That test also ran on the 1000ms default from its creation until #3284 (2026-08-10) and was never implicated in any of the five filings.

**Gate:** `src/components/billing` — 88 files / 773 tests, 0 failures · `tsc --noEmit` exit 0 · `eslint` exit 0 on all three changed files. No `.astro` touched.

## Not fixed here

- **The same anti-pattern exists in two quote-side files** — `billing/quotes/QuoteEditor.tsx:408` and `billing/quotes/QuoteHeaderMeta.tsx:54` carry the identical deferred prop-sync shape and are presumably vulnerable to the same silent edit-discard. Left alone **on purpose: open PR #4284 is editing `billing/quotes/*`**, so touching them here would collide. Worth a follow-up issue once #4284 lands.
- **A separate, orthogonal hazard I did not change → filed as #4296.** After a successful notes save the quiet refetch returns the new value, which re-syncs and discards anything typed while the PATCH was in flight — silently. Two independent review agents confirmed the trace and both agreed it is out of scope here, since closing it needs a product decision about when the server may overwrite what someone is typing. The quote-side occurrences above are recorded there too.

### DIFF
--- a/apps/web/src/components/billing/InvoiceEditor.test.tsx
+++ b/apps/web/src/components/billing/InvoiceEditor.test.tsx
@@ -472,12 +472,14 @@ describe('InvoiceEditor', () => {
   // never saw and overwrite it with the pre-edit value, silently clearing the
   // dirty flag; `saveNotes`/`saveTerms` then short-circuit on `!dirty` and no
   // PATCH is sent at all. That is what made the InvoiceWorkspace queued-Issue
-  // tests flaky for six weeks across #2925/#3219/#3277/#3980/#4033.
+  // tests flaky, filed five times from 2026-07-29 on, across
+  // #2925/#3219/#3277/#3980/#4033.
   //
   // The ordering hazard itself is only reachable by out-racing React's
   // scheduler, so these cases pin the observable contract instead: a re-render
   // whose server value is UNCHANGED after normalisation must never touch the
-  // draft, and one whose server value genuinely changed must replace it. The
+  // draft — and must leave it SAVEABLE, which is the part that actually broke —
+  // while one whose server value genuinely changed must replace it. The
   // null → '' case is the deterministic discriminator — the old
   // `useEffect(..., [invoice.notes])` compared the RAW prop, so that round-trip
   // re-ran the effect and discarded the draft.
@@ -493,6 +495,18 @@ describe('InvoiceEditor', () => {
       // Nothing the user cares about changed, so the draft must survive.
       rerender(<InvoiceEditor detail={draft([manualLine], { notes: '' })} onChanged={vi.fn()} />);
       expect(screen.getByTestId('invoice-notes')).toHaveValue('Half-typed note');
+
+      // …and it must still be SAVEABLE. The visible text is only a proxy: the
+      // damage in #3277 was the silently-cleared `notesDirty`, which makes
+      // `saveNotes()` short-circuit so the blur sends nothing at all. Asserting
+      // the text alone would pass for a regression that clears the flag without
+      // touching the value, which is the same silent data loss wearing a
+      // different hat — so assert the PATCH actually goes out.
+      fireEvent.blur(screen.getByTestId('invoice-notes'));
+      await waitFor(() => expect(fetchMock.mock.calls.some(
+        (c) => (c[1] as RequestInit)?.method === 'PATCH'
+          && String((c[1] as RequestInit)?.body).includes('Half-typed note'),
+      )).toBe(true));
     });
 
     it('keeps a dirty terms draft across the same null → "" round-trip', async () => {
@@ -504,6 +518,14 @@ describe('InvoiceEditor', () => {
 
       rerender(<InvoiceEditor detail={draft([manualLine], { termsAndConditions: '' })} onChanged={vi.fn()} />);
       expect(screen.getByTestId('invoice-terms')).toHaveValue('Net 15');
+
+      // Same reason as the notes case: prove the edit is still saveable, not
+      // merely still visible.
+      fireEvent.blur(screen.getByTestId('invoice-terms'));
+      await waitFor(() => expect(fetchMock.mock.calls.some(
+        (c) => (c[1] as RequestInit)?.method === 'PATCH'
+          && String((c[1] as RequestInit)?.body).includes('Net 15'),
+      )).toBe(true));
     });
 
     it('DOES replace the draft when the server value genuinely changes', async () => {
--- a/apps/web/src/components/billing/InvoiceEditor.test.tsx
+++ b/apps/web/src/components/billing/InvoiceEditor.test.tsx
@@ -466,4 +466,58 @@ describe('InvoiceEditor', () => {
     expect(price).toHaveTextContent('420.00');
     expect(price).not.toHaveTextContent('500');
   });
+
+  // #3277 — the free-text fields re-sync from the server value DURING RENDER,
+  // not from a `useEffect`. A deferred effect can flush AFTER a keystroke it
+  // never saw and overwrite it with the pre-edit value, silently clearing the
+  // dirty flag; `saveNotes`/`saveTerms` then short-circuit on `!dirty` and no
+  // PATCH is sent at all. That is what made the InvoiceWorkspace queued-Issue
+  // tests flaky for six weeks across #2925/#3219/#3277/#3980/#4033.
+  //
+  // The ordering hazard itself is only reachable by out-racing React's
+  // scheduler, so these cases pin the observable contract instead: a re-render
+  // whose server value is UNCHANGED after normalisation must never touch the
+  // draft, and one whose server value genuinely changed must replace it. The
+  // null → '' case is the deterministic discriminator — the old
+  // `useEffect(..., [invoice.notes])` compared the RAW prop, so that round-trip
+  // re-ran the effect and discarded the draft.
+  describe('server value re-sync', () => {
+    it('keeps a dirty notes draft when a refetch reports the same value under a different raw form (null → "")', async () => {
+      const { rerender } = render(<InvoiceEditor detail={draft([manualLine], { notes: null })} onChanged={vi.fn()} />);
+      await waitFor(() => expect(screen.getByTestId('invoice-editor')).toBeInTheDocument());
+
+      fireEvent.change(screen.getByTestId('invoice-notes'), { target: { value: 'Half-typed note' } });
+      expect(screen.getByTestId('invoice-notes')).toHaveValue('Half-typed note');
+
+      // A quiet refetch lands carrying '' where the previous payload had null.
+      // Nothing the user cares about changed, so the draft must survive.
+      rerender(<InvoiceEditor detail={draft([manualLine], { notes: '' })} onChanged={vi.fn()} />);
+      expect(screen.getByTestId('invoice-notes')).toHaveValue('Half-typed note');
+    });
+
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 agents in parallel: `code-reviewer`, `pr-test-analyzer`, `comment-analyzer`, `silent-failure-hunter`. (`type-design-analyzer` skipped — no types added or changed.)

**Findings:** 5 raised, all addressed in `1b59f20a4`; 0 outstanding.

| agent | severity | finding | disposition |
|---|---|---|---|
| pr-test-analyzer | CRITICAL | The two `null → ''` regression cases asserted only that the visible text survived. The real damage in #3277 was the silently-cleared dirty flag, so a regression that cleared it without touching the value would have passed. | **Fixed.** Both cases now blur after the re-sync and assert the PATCH goes out carrying the drafted text. Re-verified red-first: still fail against the pre-fix implementation, pass after. |
| comment-analyzer | CRITICAL | "flaky for six weeks" was wrong by date math (#2925 filed 2026-07-29 → ~4.4 weeks), and the same wrong figure was baked into three separate files. | **Fixed.** Replaced with the non-decaying fact: filed five times from 2026-07-29 on. |
| comment-analyzer | CRITICAL | "three successive rounds of raising the `waitFor` budget" — there were two changes, and only #3284 raised the async budget; #3956 added the per-test ceiling above it. | **Fixed.** Each PR is now named for what it actually did. |
| comment-analyzer | IMPORTANT | "measured present-before-any-poll on 200/200 runs" asserted a precise count with no artifact a reader could check, because the harness was never committed. | **Fixed.** The comment now leads with the mechanism you can verify by reading the code (`runScoped` marks the key pending *before* it awaits, and `fetchWithAuth` is a bare `vi.fn()` in that file, so nothing suspends before the commit), and points here for the measurement rather than asserting a number in the tree. |
| pr-test-analyzer | IMPORTANT | Reverting the file-wide timeout removes headroom from `updates the header from "Draft invoice"…`, which has its own multi-hop chain and was never implicated in the flake. | **Answered, no code change.** The 50 consecutive runs execute *all seven* tests in the file at the default budget, not just the two queued-Issue cases. That test also ran on the 1000ms default from creation until #3284 (2026-08-10) without ever flaking — reverting restores its own status quo. |

**Declined, with reason:**

- `pr-test-analyzer` (SUGGESTION) proposed a lint/source-grep guard banning `useEffect` around these setters. Declined: source-grep guards are brittle (a multi-line reintroduction slips straight through) and the reviewer itself offered it as "cheap, otherwise accept as documented residual risk". The strengthened behavioural test is the better guard, and both sites carry an explicit "must NOT move back into a `useEffect`" comment.

**Escalated rather than absorbed:**

- `silent-failure-hunter` (CRITICAL) and `pr-test-analyzer` independently confirmed a **second, orthogonal silent-loss window** on this surface: text typed during a save round-trip is discarded by the quiet-refetch resync, with no cue. Both agreed it is out of scope here, since closing it needs a product decision about when the server may overwrite what someone is typing. **Filed as #4296**, along with the two unfixed quote-side occurrences of the same anti-pattern (`quotes/QuoteEditor.tsx:408`, `quotes/QuoteHeaderMeta.tsx:54` — left alone only because open PR #4284 is editing that directory).

**`code-reviewer`: 0 findings.** It independently traced the render-phase `setState` for termination and hooks-order safety, confirmed the normalisation change is behaviour-preserving everywhere `invoice.notes`/`termsAndConditions` are read, and confirmed the two new precondition assertions are synchronous and non-vacuous rather than flake-prone.

**Tests:**
- `src/components/billing` — 88 files / 773 tests, 0 failures (re-run after the review fixes).
- Flake stability: 50 consecutive process runs green pre-review + 20 more after the review fixes, all at Testing Library's **default** 1000ms budget.
- A/B on the root cause, 200 iterations each under ambient load ~22–46: **pre-fix 2 failed / 200, fixed 0 failed / 200** (arm B's file confirmed by md5, not assumed).
- `tsc --noEmit` exit 0 · `eslint` exit 0 on all three changed files · `no-silent-mutations` guard 117/117.

**CI on `1b59f20a4`:** the three checks this PR can actually affect are green — **Test Web: success**, **Type Check: success**, **Lint: pass**. 50 checks passing, 0 failures; only `Test API` and the three `Integration Tests` shards are still running, and this PR touches no API, migration or tenancy surface.

**Status:** review-clean, awaiting maintainer merge. Issue #3277 deliberately left **open and assigned** — closing it is the maintainer's call once this lands and CI stays green on the two tests.

PR #4295, opened 2026-08-31
fix(api): gate manual/AI agent-update dispatch on the artifact edition (#4093)

### LINKED ISSUE #4093: Manual/AI agent-update dispatch bypasses the artifact-edition gate (#4072 follow-up)
Follow-up to #4072 / PR #4091, flagged independently by two review passes.

`aiToolsAgentMgmt.ts` (`trigger_agent_upgrade`, ~line 302) resolves a target through the now edition-scoped `resolvePinnedUpgradeTarget` but never consults `agentAcceptsServedEdition` before adding a device to `targetByDevice`. On a hosted control plane, an operator or AI running "update these agents" over a stranded silent 0.105.0–0.106.x self-host device still dispatches an update command whose artifact the agent refuses after download.

Severity is lower than the heartbeat path #4091 fixed — a dispatched command is one-shot, not a ~60s offer loop — but it is the same failure class as "manual Remediate ignores enforceMode" (#3381): the manual door bypasses the gate the automatic door honors. Explicit `targetVersion` input also bypasses edition/platform/arch resolution entirely.

Suggested fix: before adding a device to `targetByDevice`, evaluate `agentAcceptsServedEdition({ reportedEdition: device.agentEdition, agentVersion: device.agentVersion })` and skip (with a per-device result message naming the reason) devices that cannot accept the served edition. Sweep any other manual dispatch call sites (WS `update_status` is agent-initiated status, not a dispatch — already checked).

ToddHebebrand:
## What

`trigger_agent_upgrade` (AI chat + MCP) resolved a target through the edition-scoped `resolvePinnedUpgradeTarget` and then queued `update_agent` **unconditionally** — never consulting `agentAcceptsServedEdition`, the gate #4072/PR #4091 added to the heartbeat offer path. On a hosted control plane, "update these agents" over a stranded 0.105.0–0.106.x self-host device dispatched a command whose artifact the agent refuses after download. Same failure class as "manual Remediate ignores enforceMode" (#3381): the manual door bypasses the gate the automatic door honors.

## Where the gate went, and why

**`services/commandQueue.ts` `executeCommand`** — the single point every agent-binary update dispatch funnels through — not the one known caller. Gating a call site is how the third call site ships ungated; this repo has that scar five times over. `executeCommand` already loads the device row and already carries command-type (`INTERACTIVE_COMMAND_TYPES`) and target-role (watchdog liveness) policy, so the gate costs three extra selected columns plus a pure predicate. It runs **before** the liveness gates: an edition mismatch is a permanent property of the installed build, so reporting the transient "watchdog is not reporting" first would send the operator back to retry a dispatch that can never succeed.

Two further holes closed at the same layer:

- **`queueCommand` refuses these types outright.** It is the sibling `device_commands` insert site (used by `queueCommandForExecution` and BullMQ workers). It cannot set `target_role` — the row would default to `'agent'`, which has no handler — and has no device row to evaluate the gate against, so it can never be a correct dispatcher for them.
- **`update_agent` / `update_watchdog` must target the watchdog.** `agent/cmd/breeze-watchdog` `handleFailoverCommand` is the only consumer; an agent-targeted row goes to the agent WebSocket and is never picked up.

The AI tool needed no gate of its own — it already maps `executeCommand`'s `{status:'failed', error}` into its per-device `errors` map, so the refusal reaches the operator per device, naming the observed edition and version.

## Dispatch paths audited

| Path | Creates an agent-binary update? | Status |
|---|---|---|
| `routes/agents/heartbeat.ts` main branch (`upgradeTo`, `helperUpgradeTo`, `watchdogUpgradeTo`) | offer, not a command | already gated (#4072) |
| `routes/agents/heartbeat.ts` watchdog-failover branch (`watchdogUpgradeTo`, recovery `upgradeTo`) | offer, not a command | already gated (#4072) |
| `services/aiToolsAgentMgmt.ts` `trigger_agent_upgrade` — AI chat **and** `routes/mcpServer.ts` / `routes/mcpExecutionOrg.ts` (same shared `aiTools` registry, not a separate implementation) | `update_agent`, `targetRole:'watchdog'` | **the bug — now gated** |
| `POST /devices/:id/commands`, `POST /devices/bulk/commands`, `routes/mobile.ts` device action | zod enums that cannot express these types | n/a (the stray `'update'` enum member has no consumer server- or agent-side) |
| `routes/agentVersions.ts` promote ("fleet promote") | flips `isLatest` only; no dispatch | n/a — feeds the gated resolvers |
| `routes/devPush.ts` `dev_update` | explicit URL + checksum, never a signed release manifest, so `updater.editionAllowed` is never consulted | n/a by design — deliberately excluded from the gate |
| `scripts/recover-stuck-agents.ts` | `dev_update` for 0.65.5–0.65.9 only, all below the check band | n/a |
| deployments, workers/jobs, `ee/`, `packages/` | no agent-binary update dispatch exists | n/a |

`update_agent` is named in exactly one non-test API file today, and a static contract test now pins that set — a future hand-rolled `db.insert(deviceCommands)` dispatcher trips a test instead of shipping ungated.

## Module move

`parseComparableVersion` / `compareAgentVersions` / `AGENT_EDITION_CHECK_INTRODUCED` / `agentAcceptsServedEdition` move from `routes/agents/helpers.ts` to a new leaf `services/agentEditionCompat.ts`. `helpers.ts` imports `services/commandQueue`, so importing the predicate back would have closed an import cycle; a lazy import would have hidden the layering problem rather than fixed it. `helpers.ts` re-exports every moved symbol, so all existing import sites — and the suites that `vi.mock('./helpers')` — are untouched. The withhold wording (`editionWithheldDetail`) is now shared, so both doors describe the same condition in the same words.

## Downloader identity

The watchdog performs the download for these commands, so:

- **Version** reads `watchdogVersion`, written alongside `watchdogLastSeen` on every watchdog beat (`heartbeat.ts`), falling back to `agentVersion` only for pre-telemetry rows — agent and watchdog install and upgrade from the same lane.
- **Edition** uses `agentEdition` as the watchdog's proxy. There is no `watchdogEdition` column; this is exactly the stand-in, and the same documented imprecision, that `heartbeat.ts`'s failover branch already accepts. Persisting the watchdog's own reported edition would sharpen it and is a reasonable follow-up — it needs a migration plus the `CORE_TENANT_EXPORT_POLICY` column registration, so it is deliberately out of scope here.

## Second hole named by the issue

An explicit `targetVersion` bypassed edition/platform/arch resolution entirely — it only checked the version string existed in `agent_versions` at all, so a version with no build for a device dispatched an `update_agent` that could only 404 on the box while the tool reported `queued`. It now runs through the same per-device `resolvePinnedUpgradeTarget` call the pin path uses, behind a cheap component+edition-scoped existence probe so a typo'd or model-invented version fails fast instead of tripping the resolver's fleet-freeze Sentry capture.

## Verification

- Red first: 6 new `commandQueue` tests failed with `expected 'completed' to be 'failed'` (the ungated dispatch) before the change; the explicit-version test failed on the old resolution path. The static contract test was proven to discriminate by planting a probe file that names `update_agent`.
- `vitest run` (single fork) green: `agentEditionCompat.test.ts`, `commandQueue.test.ts`, `aiToolsAgentMgmt.test.ts`, `heartbeat.test.ts`, `helpers.agentUpdatePolicy.test.ts` — **268 passed**. Plus `commands.test.ts`, `agentWs.test.ts`, `devices/commands.test.ts`, all `routes/agents/helpers.*.test.ts`, `__tests__/helpers.test.ts` — 321 passed.
- `tsc --noEmit` clean; `eslint src` clean.
- No schema, migration, or tenancy surface touched.

Closes #4093

### DIFF
--- a/apps/api/src/routes/agents/heartbeat.ts
+++ b/apps/api/src/routes/agents/heartbeat.ts
@@ -50,8 +50,11 @@ import { decryptClaimedCommandsForDelivery } from '../../services/commandDeliver
 import { normalizeReportedScriptSecretEnvVersion } from '../../services/scriptSecretDelivery';
 import { redactSecretsDeep } from '../../services/secretRedaction';
 import { recordAgentHeartbeat, resolveResponseStatus } from '../metrics';
-import { getBinaryEdition } from '../../services/binaryEdition';
 import { ingestRollbackObservation } from '../../services/agentRollbackResult';
+import {
+  editionWithheldDetail,
+  type EditionWithheldContext as SharedEditionWithheldContext,
+} from '../../services/agentEditionCompat';
 
 /**
  * #1121 — pure collapse detector for the watchdogState tolerance gap.
@@ -100,28 +103,11 @@ export function __resetEditionWithheldWarnCacheForTests(): void {
   editionWithheldCaptured = false;
 }
 
-type EditionWithheldContext = {
-  deviceId: string;
-  reportedEdition: string | null | undefined;
-  agentVersion: string | null | undefined;
-};
-
-// The message states the OBSERVED facts and keeps every remediation
-// conditional — agentAcceptsServedEdition returns false for several distinct
-// states (silent post-cutover self-host build, hosted build on a self-host
-// server, unusable version) and asserting one cause for all of them sends an
-// operator down the wrong path.
-function editionWithheldDetail(args: EditionWithheldContext): string {
-  return (
-    `this server serves ${getBinaryEdition()}-edition artifacts and the downloading build ` +
-    `(reported edition=${args.reportedEdition ?? 'none'}, version=${args.agentVersion ?? 'unknown'}) ` +
-    `cannot be confirmed to accept them — the agent-side edition check refuses a mismatched ` +
-    `artifact AFTER download and retries every heartbeat forever, so the server withholds instead. ` +
-    `If the build is a silent ≥0.105.0 self-host agent, recover it with a ` +
-    `${getBinaryEdition()}-edition installer or an agent release that reports its edition; ` +
-    `if it reports the other edition, it is enrolled against a server of the wrong edition.`
-  );
-}
+// The withhold explanation is shared with the DISPATCH gate (#4093) so both
+// doors describe the same condition in the same words. Imported straight from
+// the leaf module rather than through `./helpers` (which suites mock) so the
+// real text is always what an operator reads.
+type EditionWithheldContext = SharedEditionWithheldContext & { deviceId: string };
 
 // Dedupe entries are keyed per (device, reporting role): the main-agent and
 // watchdog branches gate on DIFFERENT binaries' capabilities, and a device
--- a/apps/api/src/routes/agents/helpers.ts
+++ b/apps/api/src/routes/agents/helpers.ts
@@ -291,111 +291,22 @@ export function inferPatchOsType(source: string, deviceOs: unknown): 'windows' |
 }
 
 // ============================================
-// Version Comparison
+// Version Comparison / artifact-edition compatibility
 // ============================================
-
-export function parseComparableVersion(raw: string): { core: number[]; prerelease: string | null } | null {
-  const trimmed = raw.trim().replace(/^v/i, '');
-  if (!trimmed) return null;
-
-  const [rawCorePart, prereleasePart] = trimmed.split('-', 2);
-  const corePart = rawCorePart ?? '';
-  if (!corePart) return null;
-  const coreTokens = corePart.split('.');
-  if (coreTokens.length === 0) return null;
-
-  const core: number[] = [];
-  for (const token of coreTokens) {
-    if (!/^\d+$/.test(token)) return null;
-    core.push(Number.parseInt(token, 10));
-  }
-
-  return {
-    core,
-    prerelease: prereleasePart ?? null,
-  };
-}
-
-export function compareAgentVersions(leftRaw: string, rightRaw: string): number {
-  const left = parseComparableVersion(leftRaw);
-  const right = parseComparableVersion(rightRaw);
-  if (!left || !right) return 0;
-
-  const maxLen = Math.max(left.core.length, right.core.length);
-  for (let i = 0; i < maxLen; i += 1) {
-    const leftPart = left.core[i] ?? 0;
-    const rightPart = right.core[i] ?? 0;
-    if (leftPart !== rightPart) {
-      return leftPart > rightPart ? 1 : -1;
-    }
-  }
-
-  if (left.prerelease === right.prerelease) return 0;
-  if (!left.prerelease) return 1;
-  if (!right.prerelease) return -1;
-  return left.prerelease.localeCompare(right.prerelease);
-}
-
-/**
- * The agent release that introduced the client-side artifact-edition check
- * (updater.editionAllowed, #3349). Builds older than this never consult the
- * manifest's edition field and can apply artifacts of either edition; builds
- * from this version on refuse a mismatched edition AFTER download, so the
- * server must not offer them one (#4072).
- */
-export const AGENT_EDITION_CHECK_INTRODUCED = '0.105.0';
-
-/**
- * Can the binary that would perform a download apply an artifact of THIS
- * server's edition (#4072)?
- *
- * The updater's edition check runs agent-side after the download, so offering
- * a version the build will refuse does not fail once — it wedges the device in
- * a permanent ~60s retry loop (`status='updating'`, never converges, no
- * server-side escape hatch). This predicate is the server-side gate: withhold
- * the offer instead, and the device idles quietly on its current version.
- *
- * Inference, in order:
- *  - A REPORTED agentEdition (either value) means the build both knows its
- *    edition and — because the heartbeat reporting and the one-way
- *    self-host → hosted allowance in updater.editionAllowed shipped in the
- *    same agent release — can apply a hosted artifact. So when serving hosted,
- *    any reporter is accepted; when serving self-host, a 'hosted' reporter is
- *    refused (hosted builds hard-refuse self-host artifacts by design — that
- *    direction would strip the host-policy allowlist and is never relaxed).
- *  - A SILENT agent (no reported edition) is either a pre-0.105.0 build (no
- *    edition check at all → accepts anything) or a self-host build without
- *    the transition allowance (the stranded 0.105.0–0.106.x band, plus any
- *    unfixed self-host build pointed at a hosted control plane → refuses
- *    hosted). Split on the version that introduced the check. A missing or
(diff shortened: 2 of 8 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter (all three on the full `origin/main...HEAD` diff).

**Findings:** 5 raised → all addressed in `0631418`; 0 outstanding.
- *code-reviewer:* 0 findings. Independently verified the no-import-cycle claim (`agentEditionCompat` → `binaryEdition` → nothing), that no `queueCommand` call site can pass the newly-refused types, and that an agent-targeted `update_agent` was already a silent dead command before this PR — so the new role refusal converts silence into a loud error rather than blocking anything that used to work.
- *silent-failure-hunter:* 0 new silent failures. Confirmed the refusal reaches the operator per device (`executeCommand` → `errors[deviceId]`) and that the `queueCommand` throw has no swallowing caller today.
- *pr-test-analyzer:* 4 real gaps, all fixed — (1) the explicit-version probe's `component`/`edition` scoping was **vacuously covered**: the mocked query chain returns its canned rows whatever it is asked, so deleting either `eq()` left every test green. The helper now captures each `.where()` argument and walks the drizzle SQL for the columns actually filtered on. (2) nothing asserted the refusal reaches the logs. (3) nothing asserted the gate runs *before* the watchdog-liveness check, an ordering the code comments call deliberate. (4) the static dispatch-site scan now covers `ee/` and states its own limitation (literal-string match; the runtime guards are the real enforcement).

**Also fixed, found while acting on the review:** a device that passed the access check but was deleted before the device-row SELECT produced no per-device reason — and when the whole batch vanished the `errors` map was omitted entirely, leaving a batch-level message that named the wrong cause. Each such device now gets its own reason. The two resolver `catch` blocks in that loop now log the swallowed cause.

**Control checks (each new guard proven to discriminate, not just to pass):**
- Reverting `commandQueue.ts` → 6 gate tests fail with `expected 'completed' to be 'failed'`, i.e. the ungated dispatch.
- Reverting the AI-tool resolution → the explicit-version test fails.
- Dropping the `component`/`edition` scoping → `expected [ 'version' ] to deeply equal [ 'component', 'edition', 'version' ]`.
- Moving the gate after the liveness checks → `expected 'Watchdog is not reporting…' to match /edition/i`.
- Planting a file that names `update_agent` → the dispatch-site scan fails and names it.

**Tests:** `vitest run` single-fork, 447 passed across `agentEditionCompat`, `commandQueue`, `aiToolsAgentMgmt`, `heartbeat`, `helpers.agentUpdatePolicy`, `agents/commands`, `agentWs`, `devices/commands`. `tsc --noEmit` clean (`NODE_OPTIONS=--max-old-space-size=12288`); `eslint src` clean.

**CI:** 55/55 required checks pass on head `0631418`, 1 skipped (CodeQL). Includes all four **Integration Tests** shards, **Test API**, **Type Check**, **Check Migrations**, and **Smoke Test**.

**Status:** review-clean, awaiting maintainer merge. No schema, migration, or tenancy surface touched, so no RLS/cascade/export-policy registration applies.

**One thing for the maintainer to weigh:** the gate uses `devices.agent_edition` — written from main-agent beats — as a stand-in for the **watchdog's** edition, because no `watchdog_edition` column exists. That is the same proxy, with the same documented imprecision, that `heartbeat.ts`'s failover branch already accepts. Persisting the watchdog's own reported edition (it does send one) would sharpen it, but needs a migration plus the `CORE_TENANT_EXPORT_POLICY` column registration — deliberately left out of a bugfix PR and worth a follow-up issue if you want it.

ToddHebebrand:
### Rebased onto `main` — how the `commandQueue.ts` overlap with #4292 was resolved

Rebased (not merged) onto `main` and force-pushed with `--force-with-lease`; new head `d252514f1`. Two content conflicts, both resolved to keep **both** sides:

**1. `services/commandQueue.ts` — `queueCommand` (the real one).** #4292 ("guard `created_by` at the `queueCommand` insert site", #3978) rewrote the exact lines this PR touched: it routes `createdBy` through the shared `resolveCommandCreatedBy` probe and wraps the `device_commands` insert in `withSystemDbAccessContext`. Both of those properties are load-bearing and are kept verbatim — the probe still runs `runOutsideDbContextSafe(() => withSystemDbAccessContext(...))` (it has to exit the caller's context first, because `withDbAccessContext` short-circuits on an existing store), and the insert stays context-scoped rather than reverting to a bare-pool write. Dropping either would have reintroduced a #1375-class contextless write that only reddens under `DB_CONTEXTLESS_WRITE_STRICT=true` in the integration job.

This PR's contribution to that function — the `AGENT_BINARY_UPDATE_COMMAND_TYPES` refusal — is layered *on top*, placed **before** `resolveCommandCreatedBy`: it is a pure `Set` lookup that throws, so a refused type no longer pays for the `users` probe round-trip. Resulting order is `#4093 refusal → resolveCommandCreatedBy → withSystemDbAccessContext(insert)`, and `dispatchActor` / the audit row keep keying on #4292's resolved `safeUserId`.

`commandQueue.test.ts` auto-merged; the `created_by FK guard (#3978)` describe block and this PR's `agent-binary update edition gate (#4093)` block coexist untouched.

**2. `routes/agents/heartbeat.ts` — import block.** #4035 added `import { ingestRollbackObservation } from '../../services/agentRollbackResult'` immediately after the `getBinaryEdition` import that this PR deletes. Kept the new `agentRollbackResult` import, kept this PR's `agentEditionCompat` import, dropped `getBinaryEdition` — it has no remaining reference in the file once `editionWithheldDetail` moves out.

**Verification on the rebased head**

- `tsc --noEmit --project apps/api/tsconfig.json` — clean.
- `eslint` on the five changed source files — clean.
- `vitest run` on `commandQueue.test.ts`, `agentEditionCompat.test.ts`, `aiToolsAgentMgmt.test.ts`, `heartbeat.test.ts`, `helpers.agentUpdatePolicy.test.ts`, `agents/agentUpdatePolicy.test.ts` — **303 passed / 6 files**.
- Adjacent suites over the surfaces `main` moved (`commandQueueTransitions`, `agents/commands`, `agentWs`, `agentRollback`, `staleCommandReaper`, `heartbeatMetricsGate`, all `agents/helpers.*`) — **289 passed / 15 files**.
- The `#4093` static contract test still passes with its allowlist unchanged: none of the new S0 device-control code on `main` names `update_agent` / `update_watchdog`.

**One thing worth a reviewer's eye:** #4035 added a *third* `device_commands` insert site on `main`, `insertQueuedCommandInTransaction`, which is neither `queueCommand` (refuses these types) nor `executeCommand` (evaluates the gate). It cannot dispatch an agent-binary update today without naming the command type, so the static contract test would catch that regression — but it is a new bypass surface that did not exist when this PR was written.

Issue #3471, opened 2026-08-12
ci: discriminate ESRCH vs EPERM in the ee boot-step kill handling


ToddHebebrand:
Follow-up from #3455 (workspace ee-merge), from the post-review hardening pass. Smallest of the follow-up set.

The "Boot API once to apply built-in extension migrations (ee)" step in `.github/workflows/ci.yml` (shard 1) now fails closed on PGID handling and treats zombie-only process groups as dead, but the kill/poll logic does not distinguish **ESRCH** (group already gone — fine) from **EPERM** (group exists but is not signalable — a real problem that should fail the step loudly rather than being folded into the same branch).

Discriminate the two `kill` error cases so an EPERM surfaces as its own failure with the PGID in the message.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4319, opened 2026-08-31
chore(ci): repin Swatinem/rust-cache to the v2.9.2 release commit (#3748)

### LINKED ISSUE #3748: [CI] swatinem/rust-cache is pinned to master HEAD, not a release tag — the only action in the repo whose `# v2` comment doesn't match its SHA
## Description

`swatinem/rust-cache` is pinned to an **untagged `master` HEAD commit** in six workflow call sites, four of which are in `release.yml` — the workflow that builds and signs customer-facing binaries. It is the only action in the repo that does this; every other pinned action resolves exactly to the tag its trailing comment claims.

**Current state on `main`:**

```
uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
```

- `f0d9c388…` is `Swatinem/rust-cache` **`master` HEAD** (`Merge pull request #380 from Swatinem/dependabot/…`, 2026-08-17). Confirmed: `GET /repos/Swatinem/rust-cache/git/ref/heads/master` returns exactly this SHA.
- The `v2` tag points at `6323deb102c322ba6fcbdcafc7e3dddab59af2b6` — the `2.9.2` release commit, 2026-08-06.
- So the `# v2` comment is inaccurate at all six sites: the pin resolves to eleven days of untagged, unreleased `master` drift past what `v2` actually means.

## Why it's the only one

I audited every SHA-pinned action against its commented tag. Eleven of twelve are exact matches — both the precise-version style (`trivy-action # v0.36.0`, `action-gh-release # v3.0.2`, `pnpm/action-setup # v6.0.10`, `github-script # v9.0.0`, `paths-filter # v4.0.3`, `artifact-signing-action # v2.0.0`) and the major-only style (`actions/checkout # v7`, `setup-node # v7`, `setup-go # v7`, `upload-artifact # v7`, `docker/build-push-action # v7`). Only `rust-cache` mismatches.

The drift looks self-perpetuating rather than newly introduced. The *previous* pin `a45951ff…` (2026-08-10) was also an untagged `master` merge commit, also past `2.9.2`. Once a pin sits off-tag, Dependabot appears to fall back to tracking the default branch and keeps proposing the next `master` HEAD, so each weekly `github-actions` group PR carries the drift forward. #3722 (merged 2026-08-20) was one such step — it preserved the existing off-tag state rather than causing it, which is why I merged it rather than blocking on it.

## Impact

Low likelihood, non-trivial blast radius. A build-cache action can influence compiled output through the cache it restores, and four of the six call sites are in `release.yml` at `apps/viewer/src-tauri` and `apps/helper/src-tauri` — code that ships signed to customer machines. SHA-pinning is supposed to buy us "this is a specific reviewed release"; pinning to a moving branch head gives up that property while still *looking* pinned, which is the part worth fixing. Nothing here indicates the current commit is malicious.

## Suggested fix

Re-pin all six sites to the `2.9.2` release SHA and switch to the precise-version comment style already used elsewhere in these files:

```
uses: swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
```

Call sites:

- `.github/workflows/ci.yml:1576`
- `.github/workflows/ci.yml:1639`
- `.github/workflows/release.yml:1052`
- `.github/workflows/release.yml:1210`
- `.github/workflows/release.yml:1295`
- `.github/workflows/release.yml:1432`

A precise `# vX.Y.Z` comment should also stop the branch-tracking behaviour, since it gives Dependabot an unambiguous tag to resolve against.

Two smaller follow-ons worth folding in:

1. The org is `Swatinem` (capital S); all six sites write `swatinem`. GitHub resolves this case-insensitively so it is not a bug, but normalizing it makes the pin greppable against upstream.
2. Consider extending `scripts/check-supply-chain-hardening.sh` to assert that every `uses: …@<sha> # <tag>` actually resolves to `<tag>` upstream. That is the check that would have caught this drift on the PR that introduced it, instead of during a manual queue sweep. It needs network access, so it may fit better as a scheduled workflow than a pre-commit hook.

## Verification

```bash
gh api repos/Swatinem/rust-cache/git/ref/heads/master --jq '.object.sha'
# f0d9c3887740aee45f6153b24b3a6b815192ec16   <- what we pin today

gh api repos/Swatinem/rust-cache/git/ref/tags/v2 --jq '.object.sha'          # 49a0bdc7… (annotated tag object)
gh api repos/Swatinem/rust-cache/git/tags/49a0bdc70d2e1b713ca9e2869b211fcce03d3c1c --jq '.object.sha'
# 6323deb102c322ba6fcbdcafc7e3dddab59af2b6   <- what "v2" actually means
```

Found during the 2026-08-20 GH queue sweep while reviewing #3722.

ToddHebebrand:
## What

Repins all **6** `swatinem/rust-cache` call sites from an untagged `master` HEAD to the **v2.9.2 release commit**, and adds a guard so the drift cannot silently return.

## Why

`f0d9c3887740aee45f6153b24b3a6b815192ec16` is `Swatinem/rust-cache` **`master` HEAD** — a Dependabot merge commit dated 2026-08-17 — not a release. The trailing `# v2` comment claimed otherwise, so the pin carried eleven days of untagged branch drift while still *looking* pinned. It was the only action in the repo whose comment didn't resolve to its SHA.

The drift is self-perpetuating: a bare major-version comment gives Dependabot no tag to resolve against, so it tracks the default branch and each weekly group PR carries the next branch head forward. The previous pin (`a45951ff…`) was also an untagged `master` merge commit past 2.9.2.

Four of the six sites are `release.yml` jobs building **signed customer binaries** (`apps/viewer/src-tauri`, `apps/helper/src-tauri`), where a restored cache can influence compiled output. SHA-pinning is supposed to buy "this is a specific reviewed release"; pinning to a moving branch head gives that up while still looking pinned. Nothing indicates the current commit is malicious.

## Tag → SHA mapping (resolved from upstream, not copied)

Both annotated tags were dereferenced to their target commit:

| Ref | Tag object | Target commit | Notes |
|---|---|---|---|
| `tags/v2` | `49a0bdc70d2e1b713ca9e2869b211fcce03d3c1c` | **`6323deb102c322ba6fcbdcafc7e3dddab59af2b6`** | commit message `2.9.2`, 2026-08-06 |
| `tags/v2.9.2` | `63fed3e2fecf6f7b51dc6f043341b79ef82a9ae7` | **`6323deb102c322ba6fcbdcafc7e3dddab59af2b6`** | same commit — agrees |
| `heads/master` | — | `f0d9c3887740aee45f6153b24b3a6b815192ec16` | what we pinned *before*; 2026-08-17 Dependabot merge |

`v2.9.2` is the newest release (2026-08-06). Every site is now:

```yaml
uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
```

## Call sites — 6 confirmed by grep

Verified against `origin/main` (`a48b840bb`); line numbers shifted from those in the issue.

- `.github/workflows/ci.yml:1659`, `:1722`
- `.github/workflows/release.yml:1057`, `:1215`, `:1300`, `:1437`

The only other repo-wide `rust-cache` hit is a prose comment in `.github/zizmor.yml` — no pin, unchanged.

## Also folded in

**Owner casing** — normalized `swatinem` → `Swatinem` (upstream casing). GitHub resolves case-insensitively so this was never a bug, but it makes the pin greppable against upstream. Same lines already changing, so zero extra diff surface.

**Regression guard** (`scripts/security/check-supply-chain-hardening.sh`) — this is the one piece of scope beyond the pins; **say the word and I'll drop it.** It lives in a different file from the workflows, so it can't collide with #4254 or #3471. Rationale: without it, next week's Dependabot group PR re-drifts the pin and we are back here — that is exactly how this state arose. `zizmor`'s `unpinned-uses: hash-pin` policy only proves *a* SHA is present, not that it is a **released** one, which is why the drift survived every prior check.

The guard is deliberately the local, network-free version of follow-on #2 from the issue (the full "every action's comment resolves upstream" check needs network and fits better as a scheduled workflow — not attempted here). It follows the file's existing hardcoded-pin convention (`pnpm@10.34.5`) and asserts two things:

1. Each workflow contains the exact expected pin line (catches deletion/rename).
2. **No** `rust-cache@` line anywhere in `.github/workflows/` deviates from it — because `require_grep` alone only proves one good line exists, which would miss a single drifted site.

## Verification

Red-first, and both directions of the control were proven:

- **RED** (guard added, pins still drifted) → `ERROR: .github/workflows/ci.yml must pin Swatinem/rust-cache to the v2.9.2 release commit`
- **GREEN** (after repin) → `Supply-chain hardening checks passed.`
- **Mutation test** — reverted *one* of six sites to the old SHA while leaving the `# v2.9.2` comment intact (the exact Dependabot failure shape, which assertion #1 alone would miss). Guard failed and named the site: `.github/workflows/ci.yml:1722`. Restored → passes.
- **YAML validity** — both workflows parse: `ci.yml` 37 jobs, `release.yml` 26 jobs.
- `git grep f0d9c388…` → 0 hits repo-wide.

No application code touched, so no test/typecheck surface beyond the above; the changed script is itself the test and it runs in CI at `ci.yml:745` and in `scripts/security/preflight.sh`.

Closes #3748

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1656,7 +1656,7 @@ jobs:
           toolchain: stable
 
       - name: Rust cache
-        uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           workspaces: |
             apps/helper/src-tauri -> target
@@ -1719,7 +1719,7 @@ jobs:
 
       - name: Rust cache
         if: steps.changes.outputs.cargo == 'true'
-        uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           workspaces: |
             apps/helper/src-tauri -> target
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1054,7 +1054,7 @@ jobs:
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
 
       - name: Rust cache
-        uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           workspaces: apps/viewer/src-tauri -> target
 
@@ -1212,7 +1212,7 @@ jobs:
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
 
       - name: Rust cache
-        uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           workspaces: apps/helper/src-tauri -> target
 
@@ -1297,7 +1297,7 @@ jobs:
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
 
       - name: Rust cache
-        uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           workspaces: apps/viewer/src-tauri -> target
 
@@ -1434,7 +1434,7 @@ jobs:
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
 
       - name: Rust cache
-        uses: swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           workspaces: apps/helper/src-tauri -> target
 
--- a/scripts/security/check-supply-chain-hardening.sh
+++ b/scripts/security/check-supply-chain-hardening.sh
@@ -198,9 +198,19 @@ for workflow in .github/workflows/ci.yml .github/workflows/release.yml; do
     "$workflow must pin Swatinem/rust-cache to the v2.9.2 release commit"
 done
 # require_grep only proves one good line exists; this proves no site deviates.
-rust_cache_offenders="$(
-  grep -rn -i -- 'rust-cache@' .github/workflows/ | grep -vF -- "$RUST_CACHE_PIN" || true
-)"
+# Capture the scan's own status rather than letting a trailing `|| true` absorb
+# it: grep exits 1 for "no matches" but 2 for "could not read", and a security
+# predicate must never answer "clean" about input it could not read (the rule
+# reject_unaudited_apk above is written to). Piping straight into the filter
+# would hide that, because pipefail reports the rightmost status and the filter
+# legitimately exits 1 once it removes every compliant line.
+rust_cache_status=0
+rust_cache_lines="$(grep -rn -i -- 'rust-cache@' .github/workflows/)" || rust_cache_status=$?
+((rust_cache_status <= 1)) || fail \
+  "rust-cache scan of .github/workflows/ failed (grep status $rust_cache_status)"
+# The filter is deliberately case-sensitive: the owner is `Swatinem` upstream,
+# so a lowercase re-drift is reported instead of silently accepted.
+rust_cache_offenders="$(printf '%s\n' "$rust_cache_lines" | grep -vF -- "$RUST_CACHE_PIN" || true)"
 [[ -z "$rust_cache_offenders" ]] || fail \
   "every rust-cache pin must be ${RUST_CACHE_PIN}, found:"$'\n'"$rust_cache_offenders"
 
--- a/scripts/security/check-supply-chain-hardening.sh
+++ b/scripts/security/check-supply-chain-hardening.sh
@@ -184,6 +184,26 @@ for dockerfile in apps/api/Dockerfile apps/web/Dockerfile docker/Dockerfile.api 
     "$dockerfile must pin pnpm to 10.34.5"
 done
 
+# Swatinem/rust-cache sat on an untagged master HEAD for months while a trailing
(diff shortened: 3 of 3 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` + `silent-failure-hunter` (both on the diff at `aa0001089`).

**Findings:** 1 important raised → addressed in `61dc5036f`; 0 outstanding.

Both reviewers independently converged on the same defect, and neither found anything critical. The new guard's offender scan absorbed its own exit status with a trailing `|| true`. `grep` exits 1 for "no matches" but **2** for "could not read", and piping straight into the filter hid the difference — `pipefail` reports the rightmost status, and the filter legitimately exits 1 once it removes every compliant line. So a directory that could not be opened was indistinguishable from "no offenders" and the guard passed silently. That is precisely the anti-pattern `reject_unaudited_apk` is written against a few lines above ("a security predicate must never answer 'clean' about input it could not read"), and only `require_grep`'s hardcoded paths were incidentally catching it.

Fixed by capturing the scan's status and failing on anything above 1, mirroring that existing convention. Verified: a scan against a missing directory now fails with `grep status 2` instead of passing.

Both reviewers also confirmed independently — via `git ls-remote` against upstream — that `refs/tags/v2^{}` and `refs/tags/v2.9.2^{}` both dereference to `6323deb102c322ba6fcbdcafc7e3dddab59af2b6`, that exactly 6 sites changed with nothing else touched, and that the diff is disjoint from #4254 (its `ci.yml` edits sit at ~484 and ~2378-2430, far from 1656/1719). #3471 is still an issue with no PR, so nothing to collide with.

Two review notes accepted as intended behaviour rather than changed, both now documented in-script:

- A legitimate future rust-cache bump will trip the guard until `RUST_CACHE_PIN` is updated in the same commit. That is the point — it matches the file's existing `pnpm@10.34.5` contract and is what forces the bumper to state a real tag.
- The filter is case-sensitive, so a lowercase `swatinem/` re-drift is reported. Deliberate: the issue asked for the casing normalization so the pin greps against upstream.

**Tests:** no application code touched — the changed script *is* the test, and it runs in CI at `ci.yml:745` and via `scripts/security/preflight.sh`.

Red-first, with the control mutated in both directions:
- Guard added **before** the repin → `ERROR: .github/workflows/ci.yml must pin Swatinem/rust-cache to the v2.9.2 release commit`. After repin → `Supply-chain hardening checks passed.`
- Four drift mutations, each caught and named by `file:line`: drifted SHA with the comment left as `# v2.9.2` (the exact Dependabot shape, which the existence check alone would miss), comment regressed to bare `# v2`, lowercase owner, and unreadable scan target.
- Both workflows parse as valid YAML (`ci.yml` 37 jobs, `release.yml` 26 jobs); `git grep f0d9c388…` → 0 hits repo-wide.

**CI:** 35 passed, 0 failed, 7 still queued. Notably **Security Audit** (which executes the hardening script) and **Workflow Lint** (actionlint/zizmor over the edited workflows) both pass on the real runner.

**Status:** review-clean, awaiting maintainer merge. Flagging one scope call for you: the regression guard in `scripts/security/check-supply-chain-hardening.sh` is the one piece beyond the pins themselves — happy to strip it to a pins-only diff if you'd rather keep this minimal.

PR #4320, opened 2026-08-31
fix(api): register the event-dispatch maintenance repeatables unconditionally (#4124)

### LINKED ISSUE #4124: Register the event-dispatch maintenance repeatables unconditionally

ToddHebebrand:
Closes #4124

## Root cause

`initializeEventDispatchWorker` (`apps/api/src/jobs/eventDispatchWorker.ts`) returned early when `EVENT_DISPATCH_MODE=off` **and** the queue was empty:

```ts
if (backlog === 0) {
  console.log('[EventDispatchWorker] mode=off and queue empty — worker not started');
  return;   // <- skipped the maintenance block below
}
```

That `return` sat *above* the block that constructs the maintenance worker and registers the two repeatables, so registration was coupled to the **main** worker starting.

For `shadow-compare` that coupling is harmless. For `receipt-retention` it is backwards: the moment retention matters most is exactly the mode=`off`, queue-drained state **after** a rollout ends, when residual `event_delivery_receipts` rows have nothing left to prune them. As shipped in wave 3.5c they only aged out the next time the worker happened to start again (mode re-enabled, or a fresh backlog appeared) — the "Accepted gap" the original docstring recorded.

## Change

**Commit 1 — split the two concerns** inside `initializeEventDispatchWorker`:

- The **main** worker keeps its existing gate unchanged (`mode !== 'off'`, or a backlog to drain). The mode=`off` branch now sets a `startMainWorker` flag instead of returning.
- The **maintenance** worker + both repeatables register on **every** boot.

Neither job body needs a registration-time gate — they are both self-gating, which is what the issue's proposed shape rests on:

- `runShadowComparisonSweep` re-checks `eventDispatchMode() === 'shadow'` at **run** time and returns `{ skipped: true }` otherwise. A 5-minute tick under mode=`off` is one function call.
- `runReceiptRetentionSweep` on an empty table is three index-driven `DELETE` passes plus one `COUNT`, once a day (`3 15 * * *` via the scheduleRegistry cron lane).

The maintenance registration **failure domain is unchanged and still isolated** — swallowed, closes only what it opened, never rethrown. `index.ts` sets `workerStatus['eventDispatch'] = true` only if this promise resolves, so a transient Redis blip during boot must degrade to "repeatables didn't get scheduled this boot" rather than pinning `/ready` to not-ready for the process lifetime over a housekeeping job.

Docstrings that recorded the now-closed gap (module header + `registerEventDispatchMaintenanceRepeatables` + `initializeEventDispatchWorker`) are rewritten to state the new contract.

**No new registration site.** Process-role gating is untouched: `index.ts` calls this only when `breezeRole() !== 'api'`, and `src/worker.ts` has its own call — so exactly one process still registers the repeatables.

**Commit 2 — observability proportional to the widened reach** (from review, see below). That deliberate suppression now runs on the *default* mode=`off` boot, i.e. on every instance in the fleet rather than only the opt-in rollout population — so the Sentry report became the only operator-visible trace that retention never got scheduled, and it was a bare untagged `captureException`. It now carries:

- a structured log line with `errorId: 'EVENT_DISPATCH_MAINTENANCE_REGISTRATION_FAILED'` plus the mode and whether the main worker started — the same shape as this file's existing `EVENT_DISPATCH_RECEIPTS_ABANDONED` / `EVENT_DISPATCH_SHADOW_MISMATCH` lines;
- a `worker: 'eventDispatchMaintenance'` Sentry tag, the triage axis `attachWorkerObservability` puts on every job-level failure from this worker. `worker` is on `sentry.ts`'s `ALLOWED_TAG_NAMES`, so it survives the scrubber (BREEZE-18).

## Deliberate non-fix (reviewer suggestion, declined with reason)

The review also suggested an informational `workerStatus['eventDispatchMaintenance']` key. **Not doing that**, and it is not merely a scope call: `apps/api/src/services/readiness.ts:106-107` computes

```ts
const outcomes = Object.values(workerStatus);
return outcomes.length > 0 && outcomes.every(Boolean);
```

so there is no such thing as an "informational" entry in that map — any `false` pins `/ready` to not-ready, which is precisely what this catch block exists to prevent. A non-readiness-gating health signal for maintenance repeatables is a genuine gap, but it needs its own mechanism and is a separate change from this issue.

## Scope note (#3946)

Issue #3946 — `metricRollupMaintenance` registering `attachWorkerObservability` twice — is adjacent but **not** fixed here and does not collide: it is a different worker in `metricRollupMaintenance.ts`. This PR adds **no** new `attachWorkerObservability` call site; the existing `eventDispatchMaintenance` registration simply now also runs on the mode=`off` path. Worker-name cardinality is unchanged.

## Tests

Red-first, both commits. In `apps/api/src/jobs/eventDispatchWorker.test.ts`:

1. `mode off, empty queue` — main worker NOT started, but exactly one observability registration (`eventDispatchMaintenance`), the Worker constructed against `event-dispatch-maintenance`, and both repeatables added in order. This is the assertion the issue is about.
2. `mode off, backlog probe throws` — same, and still exactly one `captureException` (the probe failure), proving maintenance registration succeeded on that path too.
3. `mode off, empty queue, maintenance registration fails` — resolves rather than rejecting, closes only the maintenance worker + queue, and (commit 2) emits the `errorId` log line and the `worker`-tagged capture. Covers the isolated failure domain on the newly-reachable mode=`off` path.

Cases 1-3 were confirmed failing against the pre-fix implementation before the change landed (`3 failed | 34 passed`); the commit-2 assertion was likewise watched fail first.

Also added an `afterEach` shutdown to that describe block: the module holds `worker` / `maintenanceWorker` / `maintenanceQueue` in module-level state, and now that maintenance registers on paths that previously created nothing, a test that left it running would leak close()/construct() counts into the next case.

**Local evidence** (foreground, single-fork `--pool=threads --maxWorkers=2`):

| Check | Result |
|---|---|
| `vitest run src/jobs/eventDispatchWorker.test.ts src/jobs/eventDispatchWorker.sql.test.ts src/worker.boot.test.ts src/jobs/workerObservability.test.ts src/services/sentry.test.ts` | **98 passed (5 files)** |
| `tsc --noEmit --project apps/api/tsconfig.json` (the CI `typecheck` job's exact command) | exit 0 |
| `eslint src/jobs/eventDispatchWorker.ts src/jobs/eventDispatchWorker.test.ts` | exit 0 |

No schema, migration, tenancy, cascade, or export-policy surface is touched — no new table and no new column, so none of the registration lists apply.

## Operational note for the reviewer

The one real cost of this change: a worker-role process now constructs a BullMQ `Worker` + `Queue` for `event-dispatch-maintenance` even when the feature is entirely disabled (a couple of extra Redis connections), and runs a daily no-op retention sweep. That is the tradeoff the issue explicitly proposes, and it is what makes retention actually reach the rows it exists to prune.

### DIFF
--- a/apps/api/src/jobs/eventDispatchWorker.test.ts
+++ b/apps/api/src/jobs/eventDispatchWorker.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 /**
  * Behaviour of the event-dispatch worker (route-event + deliver-event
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `silent-failure-hunter` (both on the full diff vs `a48b840b`).

**Findings:** 2 raised → 1 fixed in `7fb89e5`, 1 declined with reason, 0 outstanding.

- **`code-reviewer`: 0 findings.** It independently verified the tests are discriminating rather than vacuous — checked the pre-fix commit into a scratch worktree, copied only the new test file onto it, and confirmed the three new assertions fail there (`observedWorkerNames()` returns `[]` instead of `['eventDispatchMaintenance']`). It also hand-traced all four reachable `(main worker started | not) × (maintenance ok | failed)` combinations through `shutdownEventDispatchWorker` and confirmed the `startMainWorker` refactor is behaviour-preserving for every pre-existing path.
- **`silent-failure-hunter`: 1 MAJOR — fixed in `7fb89e5`.** The maintenance-registration `catch` swallows by design, but this PR widens its reach from the opt-in rollout population to every instance's default `EVENT_DISPATCH_MODE=off` boot — making that Sentry report the only operator-visible trace that retention never got scheduled, while it was a bare untagged `captureException` with an unstructured log line. Now emits `errorId: 'EVENT_DISPATCH_MAINTENANCE_REGISTRATION_FAILED'` (plus mode and whether the main worker started) and a `worker: 'eventDispatchMaintenance'` Sentry tag, which is on `sentry.ts`'s `ALLOWED_TAG_NAMES` so it survives the scrubber.
- **`silent-failure-hunter`: 1 MINOR — noted, not changed.** `close().catch(() => {})` on the maintenance worker/queue reaches more boots than before. Pre-existing, intentional (a close failure during cleanup-after-failure is low-value to surface loudly), and the reviewer explicitly said no fix was required.

**Declined suggestion (with reason):** the reviewer proposed an "informational" `workerStatus['eventDispatchMaintenance']` key. There is no such thing — `apps/api/src/services/readiness.ts:106-107` computes `outcomes.every(Boolean)` over that map, so any `false` entry pins `/ready` to not-ready, which is exactly what this catch block exists to prevent. A non-readiness-gating health signal for maintenance repeatables is a real gap but needs its own mechanism; it is out of scope here and worth a follow-up issue.

**Tests:** `apps/api/src/jobs/eventDispatchWorker.test.ts` **37 passed**; the wider affected set (`eventDispatchWorker.test.ts`, `eventDispatchWorker.sql.test.ts`, `worker.boot.test.ts`, `workerObservability.test.ts`, `sentry.test.ts`) **98 passed**, single-fork `--pool=threads --maxWorkers=2`. Red-first on both commits: the issue's three assertions were watched fail against the pre-fix implementation (`3 failed | 34 passed`), and the observability assertion was likewise watched fail before `7fb89e5`.

**Typecheck / lint:** CI's **Type Check** job (`tsc --noEmit --project apps/api/tsconfig.json`) and **Lint** both passed on `7fb89e5`; **Test API** passed on `7927f42`. `eslint` on both changed files is clean locally. `60adb6be` on top is comment-only (removes a stray `//` line) and is re-running CI now.

**Status:** review-clean, awaiting maintainer merge. Not merged and issue not closed — both are the maintainer's call.

PR #4321, opened 2026-08-31
fix(web): return an IdP re-verification to the card it started from (#4055)

### LINKED ISSUE #4055: Returning from a PR #4041 IdP re-verification always lands on the TOTP QR screen, even when the user started from the passkey card
### Description

Verified at HEAD of `fix/4018-sso-reauth-enrollment` (branch moved after a rebase since the plan was written; line numbers below are current, not the plan's).

PR #4041 lets a passwordless account start an IdP re-verification round-trip from either of two places: the TOTP setup card (`MFASettings.tsx`) or the "Add a passkey" card (`ProfilePage.tsx:1121-1130`, `data-testid="passkey-sso-reauth"`, calling `handleSsoReauthStart`). Both call the same `POST /sso/reauth/start` (`ProfilePage.tsx:1135-1139`) and land back on `/settings/profile#ssoReauthGrant=<id>`.

The return path does not remember which card the user left from. `ProfilePage.tsx`'s mount effect that consumes the returned grant runs unconditionally:

```ts
useEffect(() => {
  const grantId = takeSsoReauthGrantFromHash();
  if (!grantId) return;
  setSsoReauthGrantId(grantId);
  void requestMfaSetup({ ssoReauthGrantId: grantId }).then((ok) => {
    if (ok) setSsoSetupReady(true);
  });
}, []);
```

(`ProfilePage.tsx:495-509`) — it always calls `requestMfaSetup`, which POSTs `/auth/mfa/setup` (the TOTP endpoint), regardless of whether the user actually wanted TOTP or came back to add a passkey. `setSsoSetupReady(true)` then flows into `MFASettings.tsx`'s own mount effect:

```ts
useEffect(() => {
  if (ssoSetupReady) setView('setup');
}, [ssoSetupReady]);
```

(`MFASettings.tsx:134-136`, `view` state declared at `:94`) — which only ever *sets* `view` to `'setup'` (the TOTP QR screen) and is never reset to anything else. So a user who clicked "Verify with your Identity Provider" on the **passkey** card, completed the IdP round-trip, and returned is dropped onto the **TOTP QR/enrollment view** instead — a different card, a different factor, and a `/mfa/setup` call the user never asked for.

### Why it matters

- **Confusing detour, not a dead end.** The `ssoReauthGrantId` is still held in state (`ProfilePage.tsx:162-163`), so the user isn't stranded — they can back out of the TOTP view and, if `hasSsoReauthGrant` is still true, the passkey card should still recognize the grant. But this is a UI-level bait-and-switch: someone who wanted a passkey lands on a QR code screen with no explanation why.
- **A stale pending TOTP setup is left in Redis.** `/mfa/setup`'s handler stashes a candidate TOTP secret under `mfa:setup:<userId>` (`apps/api/src/routes/auth/mfa.ts:136`), which is only cleaned up when the corresponding confirm step runs (`redis.del` at `:522` and `:771`) — even when the user never intended to set up TOTP at all. This is silently abandoned if the user goes and registers a passkey instead, sitting until its TTL expires.
- **This is the same path that makes `needsSsoReVerify` reachable at all** (`MFASettings.tsx:128`, `needsSsoReVerify = isPasswordless && !ssoReauthGrantAvailable`) — meaning the "grant already spent / lost" guard PR #4041 added is exercised by a real, everyday UI route (mis-routed intent forcing a second round-trip) and not merely a defensive/theoretical case. A user bounced from the passkey card to the TOTP view, who then backs out without completing TOTP setup, can burn their `ssoReauthGrantId` (single-use across `/mfa/setup` register/options and register/verify by design — `ProfilePage.tsx:487-490`) and land on exactly that guard on their next attempt.

### Affected files

- `apps/web/src/components/settings/ProfilePage.tsx:495-509` — mount effect that unconditionally calls `requestMfaSetup` on grant return.
- `apps/web/src/components/settings/ProfilePage.tsx:1121-1139` — the passkey card's `handleSsoReauthStart`, the other entry point into the same round-trip.
- `apps/web/src/components/settings/MFASettings.tsx:94, 128, 134-136` — `view` state, `needsSsoReVerify`, and the effect that force-opens the TOTP `'setup'` view whenever `ssoSetupReady` is true.
- `apps/api/src/routes/auth/mfa.ts:136, 522, 771` — the `mfa:setup:<userId>` Redis key: stashed at `:136`, only cleaned up by the confirm paths at `:522`/`:771` — left stale when the detour aborts a TOTP setup nobody wanted.

### Suggested approach

Carry the user's original intent across the IdP round-trip so the return path routes back to the card they left, not always to TOTP. A `sessionStorage` flag (e.g. `pendingSsoReauthIntent: 'totp' | 'passkey'`) set immediately before `handleSsoReauthStart`'s navigation, read back by the mount effect on return, is simplest — it survives the full-page navigation to the IdP and back without needing a server round-trip, and can be cleared as soon as it's consumed (mirroring how the `#ssoReauthGrant=` hash is stripped as it's read). A query param on the reauth-start/callback redirect is the alternative if there's a reason to prefer that over `sessionStorage` (e.g. if the callback redirect URL is already being used to carry state). Either way, gate the `requestMfaSetup` call on intent `=== 'totp'` and, when intent is `'passkey'`, skip straight to the passkey registration flow that already knows how to consume `hasSsoReauthGrant`.

ToddHebebrand:
Closes #4055

## Root cause

A passwordless (SSO-provisioned) account can start the **same** `POST /sso/reauth/start` round-trip from two cards on the profile page: the TOTP setup card (`MFASettings`) and "Add a passkey" (`ProfilePage`, `data-testid="passkey-sso-reauth"`). Both leave the origin entirely — `window.location.assign(<IdP url>)` — and the SSO callback returns everyone to the same server-minted `/settings/profile#ssoReauthGrant=<id>`.

Nothing recorded **which** card the user left from, so `ProfilePage`'s mount effect ran `requestMfaSetup` unconditionally and set `ssoSetupReady`, which `MFASettings`' own effect turns into `setView('setup')` — and never resets. A user who clicked "Verify with your identity provider" on the **passkey** card came back to the **TOTP QR screen**: a different card, a different factor, and a `/auth/mfa/setup` call they never asked for. That call's candidate secret is stashed in Redis under `mfa:setup:<userId>` and is only deleted by the TOTP *confirm* step, so an abandoned detour leaves it sitting there until its TTL expires.

## The fix

Carry the originating card across the round-trip and route the return leg on it.

- **New `apps/web/src/lib/ssoReauthIntent.ts`** — stashes `'totp' | 'passkey'` and consumes it on return.
- **`handleSsoReauthStart(intent)`** — both call sites pass their own card's intent. The stash happens only once `runAction` has produced an `authUrl`, so a failed start (which never leaves the page) can't leave a stale value behind.
- **The mount effect** consumes the intent **unconditionally** — an error return (`?ssoReauthError=`) has nothing to route, and a value left behind would misroute the *next* trip — then skips `/auth/mfa/setup` entirely when the intent is `'passkey'`. The passkey card already knows how to spend the same grant on `register/options` + `register/verify` once `hasSsoReauthGrant` is true.
- **A passkey return scrolls the passkey card into view and focuses the name field.** The trip is a full-page navigation, so the browser drops the user at the *top* of a long settings page with the card they left from below the fold; "return the user to the card" has to mean actually taking them there. Focus does the same job for keyboard and screen-reader users.
- **The card's instruction line changes once the grant is in hand** ("Your identity is verified. Name this passkey and continue.") — telling someone to "re-verify before adding a passkey" seconds after they did exactly that reads as a failed round-trip.

## Why `sessionStorage` and not the repo's `window.location.hash` convention

The hash convention (`DeviceDetails.tsx`, `OrganizationsPage.tsx`) covers **in-page UI state on a URL we control**. Neither half holds here:

1. The trip leaves the origin for the IdP, so all React state — and any hash we set — is gone by the time we come back.
2. The return URL is minted **server-side** by the SSO callback. The client cannot hang anything off it.

The alternative was round-tripping the intent through the API's re-auth state row. Rejected deliberately: that row is security-relevant (epochs, `sid`, the single-use grant), and a purely presentational "which card do I go back to" preference has no business on it — it would also mean a server contract change and a migration for a UI routing fix.

`sessionStorage` is the shape this repo already uses for exactly this problem — `apps/web/src/lib/orgSwitch.ts` stashes a confirmation toast across a full-page reload — including the same private-mode caveat: every access is wrapped in try/catch, and a `null` read falls back to the **historical TOTP road** so a storage-blocked browser behaves as it did before this PR rather than landing on a page where neither card does anything.

No API, schema, migration, or tenancy surface is touched; the stale-Redis-key symptom in the issue is fixed by no longer making the call, not by changing `mfa.ts`.

## Verification

- `apps/web` affected suites, single-fork on a base rebased onto current `main`:
  `ssoReauthIntent.test.ts` (5), `ProfilePage.test.tsx` (32),
  `ProfilePage.passkeys.test.tsx` (13) — **50 passed**.
- Regression batch: `src/lib/i18n` (locale parity + translation coverage) and
  `src/lib/__tests__/no-silent-mutations.test.ts` — **252 passed (8 files)**;
  plus `MFASettings.test.tsx`, `ConnectSsoCard`, `ApproverDevicesSection`,
  `ChangePasswordForm` green on the pre-rebase run.
- `tsc --noEmit -p apps/web/tsconfig.json` — clean. `pnpm --filter @breeze/web lint`
  — clean. No `.astro`, API, schema, or migration files changed.

**Red-first.** The 6 passkey-road and 3 TOTP-road assertions were written against
unmodified `ProfilePage` and watched fail (8 red; the 9th — "falls back to the TOTP
road when no intent was recorded" — passes before and after by design, it pins the
fallback). One of those initial reds was a *test* defect, not a product one: a
`window.location` stub restored after the assertions instead of in a `finally`
leaked an empty hash into every later test in the file. Fixed before implementing,
so the remaining reds are honest.

**Second red, found in review — the landing never fired in production.** The first
commit hung the scroll+focus off an object ref plus an effect keyed on the return
flag. But `profile.astro` renders `<ProfilePage client:load />` with **no**
`initialUser`, so `isLoadingUser` starts true and the component's first commits
return the "Loading profile…" early return — the passkey card is not in the DOM at
the moment the mount effect sets the flag. The effect found a null ref, no-opped
through the optional chain, and never ran again once the card mounted. Every
existing test passes `initialUser` and so skips that branch entirely, which is
exactly how a landing that never happens stays green. The second commit replaces it
with a **callback ref** latched by `passkeyLandingDone`, which fires on whichever
render actually attaches the node, and adds a regression test that renders
`<ProfilePage />` with no `initialUser` — watched failing on the old effect first.

`runAction`: `handleSsoReauthStart` already routes `POST /sso/reauth/start` through
`runAction`; this PR changes its signature, not its error path.

## i18n

One new key, `profilePage.yourIdentityIsVerifiedNameThisPasskeyAndContinue`, added to all 8 locales.

pt-BR strings are machine-drafted pending native review

es-419, fr-FR, fr-CA, de-DE, and it-IT strings are machine-drafted pending native review

### DIFF
--- a/apps/web/src/components/settings/ProfilePage.passkeys.test.tsx
+++ b/apps/web/src/components/settings/ProfilePage.passkeys.test.tsx
@@ -35,8 +35,9 @@ vi.mock('./ConnectSsoCard', () => ({
 }));
 
 import ProfilePage from './ProfilePage';
+import { SSO_REAUTH_INTENT_KEY, stashSsoReauthIntent } from '@/lib/ssoReauthIntent';
 
-const makeJsonResponse = (payload: unknown, ok = true, status = ok ? 200 : 500): Response =>
+const makeJsonResponse =(payload: unknown, ok = true, status = ok ? 200 : 500): Response =>
   ({
     ok,
     status,
@@ -47,6 +48,9 @@ const makeJsonResponse = (payload: unknown, ok = true, status = ok ? 200 : 500):
 describe('ProfilePage passkey management', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    // #4055 records the originating card here; a value bleeding between tests
+    // would silently reroute the one after it.
+    sessionStorage.clear();
     globalThis.URL.createObjectURL = vi.fn(() => 'blob:fake');
     globalThis.URL.revokeObjectURL = vi.fn();
   });
@@ -188,6 +192,9 @@ describe('ProfilePage passkey management', () => {
     it('sends the SAME grant to BOTH register/options and register/verify', async () => {
       // Arriving back from the IdP: the callback hands the grant over in the
       // fragment, and ProfilePage holds it for whichever road spends it first.
+      // #4055: the trip started from the passkey card, so that is what is
+      // recorded — the return must come back here, not to the TOTP QR screen.
+      stashSsoReauthIntent('passkey');
       window.history.replaceState(null, '', `/settings/profile#ssoReauthGrant=${GRANT}`);
 
       fetchWithAuthMock.mockImplementation(async (url: string) => {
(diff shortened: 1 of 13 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (correctness: React effect/ref semantics, the full `hasPassword` × intent × grant routing matrix, sessionStorage lifetime, #4041 single-use grant semantics, repo conventions) and `pr-test-analyzer` (vacuous-assertion audit, uncovered branches, test hygiene).

**Findings:** 3 raised → all 3 addressed in `f019c6e6`; 0 outstanding.

1. **Mount race defeated the scroll+focus landing in production** *(correctness — the one consequential finding)*. `profile.astro` renders `<ProfilePage client:load />` with no `initialUser`, so the component sits behind the `isLoadingUser` early return while `GET /users/me` is in flight. The passkey card is not in the DOM when the mount effect sets `passkeyReauthReturn`, so the effect found a null ref, no-opped through the optional chain, and never ran again once the card mounted. Every existing test passes `initialUser` and skips that branch — a landing that never fires stayed green. Replaced the object ref + effect with a **callback ref** latched by `passkeyLandingDone`, which fires on whichever render actually attaches the node. Regression test renders `<ProfilePage />` with **no** `initialUser` and was watched failing on the old effect before the change.
2. **Untested false branch of the instruction ternary** *(coverage)*. Only the grant-present copy was asserted; the branches could have been swapped — telling a user with no grant that their identity is already verified — without a red test. Added text-level assertions for both branches.
3. **Pre-existing `window.location` stub leak** *(test hygiene)*. `ProfilePage.passkeys.test.tsx` restored the stub after its assertions rather than in a `finally`, so one failure leaked an empty hash into every later test in the file. Now restored unconditionally in `beforeEach`, matching the guard `ProfilePage.test.tsx` already carries.

Two lower-severity notes were reviewed and **not** actioned, deliberately:
- *"`still opens the QR screen…` is a regression pin, not a fix-discriminator."* Correct, and intended — only its trailing `takeSsoReauthIntent` assertion discriminates. Kept as-is; the TOTP road must not change, and that is what it pins.
- *"Cross-card grant reuse (grant from a TOTP-originated trip, then spent on a passkey) is now untested."* Real but pre-existing, and orthogonal to this fix: `setSsoReauthGrantId` runs **before** the intent branch, so `hasSsoReauthGrant` — not `passkeyReauthReturn` — still gates that CTA. Better as a fast-follow than as scope on a routing fix.

**Tests:** `ssoReauthIntent.test.ts` (5) + `ProfilePage.test.tsx` (32) + `ProfilePage.passkeys.test.tsx` (13) = **50 passed**, single-fork, on a base rebased onto current `main`. Regression batch `src/lib/i18n` + `no-silent-mutations` — **252 passed (8 files)**. `tsc --noEmit` clean; `pnpm --filter @breeze/web lint` clean. No API, schema, migration, or tenancy surface touched.

**Status:** review-clean, rebased onto current `main`, awaiting maintainer merge. Not merged and issue #4055 left open per handoff rules.

PR #4323, opened 2026-08-31
fix(web,api): give recurring maintenance windows a start time (#4224)

### LINKED ISSUE #4224: Configuration Policy Maintenance Window has no start time field and saves window_start as NULL
### Description

When configuring Maintenance for a Configuration Policy, the UI allows selecting:

- recurrence
- duration
- timezone
- suppression options
- reboot behavior
- notifications

However, there is no field to specify the maintenance window start time.

After saving, the backend stores the configuration with:

`window_start = NULL`

Example:

- recurrence: `daily`
- duration_hours: `2`
- timezone: `Europe/Warsaw`
- window_start: `NULL`

The `maintenance_windows` table also contains no generated window/occurrence for this policy.

This makes it unclear when the configured maintenance window is expected to begin.

### Steps to Reproduce

1. Open a Configuration Policy in Breeze.
2. Go to the **Maintenance** section.
3. Enable/configure Maintenance.
4. Set recurrence to `Daily`.
5. Set duration to `2 hours`.
6. Set timezone to `Europe/Warsaw`.
7. Configure any suppression/reboot/notification options.
8. Observe that there is no field in the UI to specify the maintenance window start time.
9. Save the policy.
10. Check the stored configuration.

The resulting database record contains:

- `recurrence = daily`
- `duration_hours = 2`
- `timezone = Europe/Warsaw`
- `window_start = NULL`

Additionally, no corresponding record is created in `maintenance_windows`.

As a result, the policy defines how long and how often the maintenance window should run, but not when it should start.

### Expected Behavior

The Configuration Policy Maintenance UI should provide a start time field.

For example:

Start time: `01:50`
Duration: `2 hours`
Timezone: `Europe/Warsaw`

The configured start time should be persisted in `config_policy_maintenance_settings.window_start` and used when evaluating recurring maintenance windows.

### Actual Behavior

There is no start time field in the UI.

Saving the policy results in `window_start = NULL`, so the recurring maintenance configuration has no explicit start time.

### Component

Web Dashboard

### Version

_No response_

### Relevant Logs

```shell
name       | window_start | recurrence | duration_hours |   timezone    
-----------------+--------------+------------+----------------+---------------
 Standard        |              | weekly     |              2 | Europe/Warsaw
 Windows Servers |              | daily      |              2 | Europe/Warsaw
(2 rows)

breeze=#
```

### Additional Context

This was observed while configuring a daily maintenance window for a server patching policy.

Database verification:

`config_policy_maintenance_settings.window_start` is NULL after saving the policy.

No records were present in `maintenance_windows`.

ToddHebebrand:
## Problem

Configuring Maintenance on a Configuration Policy offered recurrence, duration, timezone, suppression, reboot and notification controls — but **no way to say when the window opens** for `daily` / `weekly` / `monthly`. The start-time control existed, gated on `recurrence === 'once'`, so recurring policies saved `window_start` empty.

The evaluator matched that gap. `isInMaintenanceWindow` read `windowStart` only for `once` and hardcoded every recurring window to **local midnight** (`setHours(0, 0, 0, 0)`). So the reporter's "daily, 2h, Europe/Warsaw" policy was silently enforcing **00:00–02:00 Warsaw every day** — a window the admin never chose and the UI never displayed. The policy read as configured while running on an invisible default.

(The empty `maintenance_windows` table in the report is expected, not part of this bug: that table backs the separate manually-created windows feature. Config-policy maintenance is computed in-memory and never materialises rows there.)

## Fix

`config_policy_maintenance_settings.window_start` becomes recurrence-discriminated:

| recurrence | `window_start` | anchor |
|---|---|---|
| `once` | ISO-8601 local datetime (unchanged) | that datetime |
| `daily` | `"HH:MM"` | that time of day, every day |
| `weekly` | `"HH:MM"` | that time of day on Sunday |
| `monthly` | `"HH:MM"` | that time of day on the 1st |

**No migration.** The column is already `varchar(30)`, and **NULL keeps meaning midnight**, so every policy in the field keeps running on exactly the schedule it runs on today. A full datetime is also accepted for the recurring cadences (its time component is used), so a policy switched from `once` never silently reverts to midnight.

### Most-recent-occurrence evaluation

A configurable start time can sit late in its period, which the midnight-only anchor never had to handle. The evaluator now resolves the **most recent occurrence at or before now** rather than the current period's, so a 23:00 daily window is still open at 00:30 the next morning (and a Sunday-night weekly window stays open into Monday). One occurrence is sufficient in every case: the previous one is always at least a full period older, and durations are capped at 72h.

An unparseable `window_start` on a recurring cadence warns and falls back to midnight rather than treating the window as never open — a fail-open that matches the pre-existing invalid-timezone handling and keeps suppression working on bad data.

### UI

`MaintenanceTab` renders `<input type="time">` for the recurring cadences and keeps the datetime picker for `once`. Switching cadence converts the value between the two shapes, because both input types **silently discard** a value they cannot parse — a carried-over `"HH:MM"` in a `datetime-local` would render blank and then save blank, reproducing the exact defect. A blank field normalises to `00:00` **on save, not on keystroke**, so it does not fight a half-entered time. Hydration normalises too, which makes the midnight anchor the API has been applying all along finally visible in the form.

New i18n keys added across all 8 locales (`localeParity` is a hard gate).

## Verification

- `apps/api` — `featureConfigResolver.test.ts` **58 passed** (24 in `isInMaintenanceWindow`, 14 of them new and each watched fail before the fix); affected batch with the maintenance consumers (`maintenanceRebootWorker`, `patchSchedulerWorker`, `patchJobService`, `featureConfigResolver.timezone`) **128 passed**; `aiToolsConfigPolicy.test.ts` **26 passed**.
- `apps/web` — `MaintenanceTab.test.tsx` **18 passed** (11 new, each watched fail first); `localeParity`, `i18n`, `no-silent-mutations`, `no-translated-comparisons`, `featureTypeParity`, `ConfigPolicyDetailPage` green.
- **CI green on the first commit**: Type Check, Test API, Test Web, all 4 Integration Tests shards, Lint, CI Success. (The local full `tsc`/`astro check` runs were abandoned — this host was at load 40+ and CI's run is authoritative.)
- `eslint` clean on every changed file.
- One test that passed vacuously was tightened: the tab echoes unknown `inlineSettings` straight back, so a `windowStart` round-trip proves nothing on its own — it now also asserts the control displays the stored value.

No mutation handlers were added or changed, so `runAction` adoption is untouched — `useFeatureLink` remains the typed service layer it was, with its error surfaced inline by `FeatureTabShell`.

## Review findings addressed

A `pr-review-toolkit` code review found two real bugs in the new parsing, both fixed in `e1fc313`:

1. **Offset-bearing instants read as local wall-clock time.** `migrateToConfigPolicies` writes migrated `once` windows as `toISOString()`, so a policy can hold a `Z`-suffixed value. Carrying its digits across a cadence switch shifted the window by the zone's UTC offset, invisibly. Both parsers now reject a trailing `Z` / `±HH:MM` and fall back to the visible midnight default instead of guessing.
2. **Web/API parser divergence on `"HH:MM:SS"`.** The API accepted seconds; the web regex did not, so opening the tab and saving silently downgraded such a policy to `00:00`. The patterns are now identical and carry a comment saying they must stay that way.

The review also flagged the `manage_policy_feature_link` AI-tool description as stale — it still advertised `windowStart` as "ISO-8601 (for once)", so an agent had no way to know a recurring window could take a time of day. Updated.

Nothing was left unaddressed. The review separately confirmed backwards compatibility (NULL is byte-for-byte equivalent to the old hardcoded midnight anchor in all three branches), the most-recent-occurrence arithmetic including the >24h-duration overlap case, and the fail-open-to-midnight choice for unparseable values.

## Left for a follow-up

Weekly is still hardcoded to **Sunday** and monthly to the **1st**. Making those configurable needs new columns, which ripples into the shipped partner-export canonicalisation functions — a separate change, deliberately not bundled here.

Closes #4224


ToddHebebrand:
**Review run:** `pr-review-toolkit` code-reviewer, scoped to the five risk areas this change actually carries — backwards compatibility of the midnight fallback, the most-recent-occurrence arithmetic, parser agreement between the API and the web form, the React state flow, and fail-open vs fail-closed on bad data.

**Findings:** 3 raised (2 confirmed bugs + 1 documentation gap) → all 3 addressed in `e1fc313`; 0 outstanding.

1. **Offset-bearing instants read as local wall-clock time** — `migrateToConfigPolicies` writes migrated `once` windows as `toISOString()`, so a policy can hold a `Z`-suffixed value. Carrying its digits across a cadence switch shifted the window by the zone's UTC offset with nothing in the UI to show for it. Both parsers now reject a trailing `Z` / `±HH:MM` and fall back to the visible midnight default rather than guessing a conversion the rest of this (deliberately wall-clock-naive) evaluator does not do.
2. **Web/API parser divergence on `"HH:MM:SS"`** — the API accepted an optional seconds group, the web regex did not, so opening the tab on such a policy and saving silently downgraded it to `00:00`. The two patterns are now identical and carry a comment saying they must stay that way.
3. **Stale AI-tool contract** — `manage_policy_feature_link` still advertised `windowStart` as `"ISO-8601 (for once)"`, so an agent had no way to know a recurring window could take a time of day at all. Updated with the recurrence-discriminated shape and an explicit warning against passing an instant.

The review separately **confirmed and found nothing wrong in** three areas worth recording: NULL `window_start` is byte-for-byte equivalent to the old hardcoded midnight anchor in all three recurring branches (no existing policy's schedule moves); the step-back arithmetic is correct at the month, week and midnight boundaries, and checking a single occurrence is provably sufficient even at the 72h duration cap; and the fail-open-to-midnight choice for an unparseable value is the right one — failing closed would silently *stop* suppressing alerts and patching during a real maintenance window.

**Tests:** every new assertion was watched fail before the fix. `apps/api` `featureConfigResolver.test.ts` **58 passed** (24 in `isInMaintenanceWindow`, 14 new); affected-consumer batch (`maintenanceRebootWorker`, `patchSchedulerWorker`, `patchJobService`, `featureConfigResolver.timezone`) **128 passed**; `aiToolsConfigPolicy.test.ts` **26 passed**. `apps/web` `MaintenanceTab.test.tsx` **18 passed** (11 new), plus `localeParity`, `i18n`, `no-silent-mutations`, `no-translated-comparisons`, `featureTypeParity`, `ConfigPolicyDetailPage`. One test that passed vacuously was tightened — the tab echoes unknown `inlineSettings` straight back, so a `windowStart` round-trip proves nothing on its own; it now also asserts the control displays the stored value. `eslint` clean on every changed file.

**CI:** green on head `e1fc313` — 55 pass, 0 fail. Type Check, Test API, Test Web, all four Integration Tests shards, Lint, CI Success.

**Status:** review-clean, awaiting maintainer merge. One deliberate scope boundary for the merger to weigh: weekly is still hardcoded to **Sunday** and monthly to the **1st**. Making those configurable needs new columns, which ripples into the shipped partner-export canonicalisation functions — left for a separate change rather than bundled here.

PR #4324, opened 2026-08-31
fix(api): route openaiCompatibleProvider egress through guarded safeFetch (#4121)

### LINKED ISSUE #4121: openaiCompatibleProvider.ts: swap raw fetch for guarded/safe fetch
Deferred follow-up from #3922 (Hosted BYO-LLM phase 2), item 3 of 4 filed at W4 close-out.

**Ask:** `openaiCompatibleProvider.ts:86` calls raw `fetch(` instead of going through the guarded fetch adapter (`safeFetch` / `services/urlSafety.ts`) that W02 built for catalog LLM clients (DNS-pinning, private-network blocking, `redirect:'error'`).

**Why deferred, not built in phase 2:** phase 2's architecture is Anthropic-dialect-only (Agent SDK subprocess + one-shot `Anthropic` clients); `openaiCompatibleProvider.ts` is dead code on the hosted path — it's reachable only in self-host configurations. Not a phase-2 blocker, but worth closing the gap so the file doesn't become a silent SSRF/egress-bypass exception the next time someone extends it.

**Scope when picked up:** swap the raw `fetch(` call for `safeFetch` (or the guarded adapter from `apps/api/src/services/llm/guardedLlmFetch.ts`) with `redirect: 'error'`, matching the pattern already used by the catalog LLM clients. Bundleable cleanup — can ride any future wave/PR that touches this file.

Ref: `docs/superpowers/plans/ai-mcp/2026-08-26-byo-llm-phase2-allowlist-egress.md`, "Deferred / follow-ups" item 3.

ToddHebebrand:
Closes #4121

## The problem

`openaiCompatibleProvider.chatStream` dialed its endpoint with raw global `fetch`, bypassing three things every other outbound caller gets from `safeFetch`: connect-time DNS pinning / private-network filtering, redirect discipline, and the #1105 held-DB-context tripwire (whose own comment notes that "modules that call global `fetch()` directly bypass safeFetch and are NOT covered").

The URL comes from operator config (`MCP_LLM_BASE_URL`), not from a tenant, so this was a standing exception rather than a live exploit — which is exactly how the issue framed it: worth closing "so the file doesn't become a silent SSRF/egress-bypass exception the next time someone extends it."

## Why it wasn't a one-line swap

The issue and the plan doc both say "swap `fetch(` for `safeFetch` with `redirect:'error'`". Taken literally that silently breaks the feature:

- **`safeFetch` fully buffers.** It accumulates `chunks.push(c)` and only resolves on `'end'` with `new Response(Buffer.concat(chunks))`. `chatStream` does incremental SSE parsing over `response.body.getReader()`, so buffering would collapse token-by-token streaming into one burst delivered after the turn finished, and hold the whole turn in memory across the 6-minute timeout. (`guardedLlmFetch.ts` documents this explicitly — it is fine there because those are one-shot JSON clients.)
- **`redirect: 'error'` is inert.** `safeFetch` never reads `init.redirect`; it follows nothing and hands back the raw 3xx. The refusal has to be explicit at the call site.
- **`safeFetchFollowingRedirects` is the wrong helper.** It also buffers, and it *follows* redirects — the opposite of what a POST carrying `Authorization: Bearer` wants.

So `safeFetch` gains an **opt-in streaming mode**. Default behavior is byte-identical for every existing caller (SSO JWKS, Huntress, Pax8, DNS, SentinelOne, webhooks, PSA, log forwarding, backup).

## Approach

Considered and rejected: adding `undici` as a direct dependency to pass a guarded `dispatcher` to native `fetch` (it is override-only today, and `createGuardedHttpAgents` returns Node-core `http.Agent`s that global `fetch` cannot accept); and a separate streaming fetch module under `services/llm/` (would duplicate Host derivation, header sanitization, the cleartext-vs-pinned-record check, timeout, abort and the tripwire — the drift the shared helper exists to prevent).

### `apps/api/src/services/urlSafety.ts`

- **`SafeFetchInit.streamResponse`** — resolve as soon as headers arrive with a `Response` whose body is the live socket. SSRF resolution, filtering and IP pinning all still happen before the socket is dialed; only body delivery changes.
- **`maxBytes` enforced as bytes flow** — exactly the cap passes, the first byte past it destroys the socket and errors the body stream (rather than rejecting an already-settled promise).
- **Backpressure** via a single pause/resume bridge; `cancel()` tears the socket down, so a consumer that walks away does not leave the model generating into a dead connection.
- **Post-settle transport errors surface on the body.** Once streaming, `reject` is inert — routing a socket failure there would leave the consumer awaiting a stream that never ends. This is also the path an abort/timeout `req.destroy(err)` takes mid-stream.
- **Premature close is an error, not EOF.** A peer that drops mid-body emits `'close'` with `res.complete === false` and possibly no `'error'`; treating that as a clean end would hand back a silently truncated body.
- `toResponseHeaders` extracted and shared by both paths.

### `apps/api/src/services/llm/openaiCompatibleProvider.ts`

- `safeFetch` with `streamResponse: true`, a 32 MiB turn ceiling, and the same `allowPrivateNetwork: selfHostAllowsPrivateNetwork()` + `requirePrivateForCleartext: true` pair `webhookDelivery.ts` uses.
- **3xx refused explicitly**, with an actionable message. Following one would be an SSRF bypass and would replay the bearer token at the `Location`.
- **The turn timeout stays armed across the non-2xx body read.** With a live body, an endpoint that sends error headers and then stalls would otherwise hang the turn forever — a hazard that did not exist while the body was pre-buffered.
- An SSRF refusal yields a message naming what is actually dialable, since the operator set the URL themselves.

## Self-hoster note (behavior change)

`MCP_LLM_PROVIDER=openai-compatible` is a self-host-only path, and `allowPrivateNetwork` permits RFC1918/ULA but **never loopback** — that contract is deliberately shared across many callers and is not broadened here. Containerized deployments resolve the model host to a service name or LAN IP (RFC1918) and are unaffected. A bare-metal install pointing at `http://localhost:8000/v1` must switch to the model host's service name or LAN address; the error message says so.

## Testing

- **`openaiCompatibleProvider.test.ts` (new, 9 tests)** — the file had zero coverage before. Written red-first: all 9 failed against `main` because a stubbed global `fetch` threw. Covers no-raw-`fetch`, the guard options, streamed delivery, SSRF refusal surfacing as an `error` event, and 3xx refusal.
- **`urlSafety.test.ts` (+10 tests, 56 total)** — headers resolve before the body ends; chunks arrive incrementally; exactly-`maxBytes` passes and cap+1 errors while destroying the socket; a mid-stream transport error reaches the reader instead of hanging; cancel destroys the request; premature close errors; null-body statuses; SSRF policy still refuses before any socket; and the default path stays buffered.
- **Regression sweep across every `safeFetch` consumer suite** — ~492 tests green (sso, webhookDelivery, webhookSender, guardedS3Client, backupSsrfAdoption, dnsProviders, recoveryMedia, huntress, logForwarding, psa, sentinelOne, guardedLlmFetch, llmEgressProxy, providerFidelityHarness, llmConfigResolver, backup/configs, llmProviderCatalog, tripwire).
- `eslint` clean on all four files.

### DIFF
--- a/apps/api/src/services/llm/__scripts__/openai-smoke.ts
+++ b/apps/api/src/services/llm/__scripts__/openai-smoke.ts
@@ -18,6 +18,10 @@ import type { ChatMessage } from '../types';
 import { envFloat } from '../../../utils/envFloat';
 import { envStr } from '../../../utils/envStr';
 
+// The provider dials through the SSRF-guarded `safeFetch` (#4121), which
+// refuses loopback even with the self-host private-network opt-in. So this
+// default no longer connects to anything: set MCP_LLM_BASE_URL to the model
+// host's LAN address or container service name (RFC1918/ULA) before running.
 const BASE_URL = envStr('MCP_LLM_BASE_URL', 'http://localhost:8000/v1');
 const API_KEY = envStr('MCP_LLM_API_KEY', 'changeme');
 const MODEL = envStr('MCP_LLM_MODEL', 'qwen3.6-27b');
--- a/apps/api/src/services/llm/openaiCompatibleProvider.test.ts
+++ b/apps/api/src/services/llm/openaiCompatibleProvider.test.ts
@@ -13,14 +13,20 @@
  *    finished, and hold a whole LLM turn in memory across the 6-minute timeout.
  */
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import http from 'http';
+import { EventEmitter } from 'events';
 
-const { safeFetchMock, selfHostAllowsPrivateNetworkMock } = vi.hoisted(() => ({
+const { safeFetchMock, selfHostAllowsPrivateNetworkMock, realUrlSafety } = vi.hoisted(() => ({
   safeFetchMock: vi.fn(async (_url: string, _init?: unknown) => new Response('', { status: 200 })),
   selfHostAllowsPrivateNetworkMock: vi.fn(() => false),
+  // Stashed so the seam suite below can drive the provider through the REAL
+  // safeFetch (and the real module's DNS test hook) instead of a stand-in.
+  realUrlSafety: { mod: null as typeof import('../urlSafety') | null },
 }));
 
 vi.mock('../urlSafety', async () => {
   const actual = await vi.importActual<typeof import('../urlSafety')>('../urlSafety');
+  realUrlSafety.mod = actual;
   return { ...actual, safeFetch: safeFetchMock };
 });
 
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** three `pr-review-toolkit` lenses — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`. A `codex exec` design consultation ran first, before implementation, on whether to extend `safeFetch` vs. duplicate it.

**Findings:** 9 raised → 9 addressed in `9fa0f1ba6`; 0 outstanding. 1 further finding accepted as out-of-scope (below).

- **`code-reviewer` — 1 major, fixed.** The `NULL_BODY_STATUSES` (204/205/304) branch was the only streamed response path with no `res.on('error')` listener. An `'error'` on an EventEmitter with no listener throws synchronously and takes the **whole API process** down, and that branch has neither a body stream nor a live `reject` to absorb it. Both the response's and the request's late errors now route to a handler that logs. Everything else it checked came back clean against the pre-image: the default buffered path is byte-identical (`toResponseHeaders` matches the loop it replaced), no double-settle, no `failStream` assignment window, no backpressure deadlock, `maxBytes` off-by-one correct.
- **`silent-failure-hunter` — 4 in-scope, all fixed.** The new error listener was a pure black hole (now logs); the same branch left `failStream` unassigned so a late `req` error hit an inert `reject` (now routed); `response.body.cancel()` on the 3xx path floated (now `.catch`ed); a failed read of a non-2xx error body was swallowed unlogged (now says why). It confirmed the highest-value question is clean and tested: **no path lets a truncated stream present as a complete answer** — a mid-stream failure rejects `reader.read()`, which always yields an `error` event and returns, never reaching `message_end`.
- **`pr-test-analyzer` — 4 consequential coverage gaps, all closed (+7 tests).** It confirmed the core regression guard is non-vacuous (`safeFetch` being mocked doesn't weaken the global-`fetch` assertion, which is checked against a separate throwing stub). Gaps closed: the non-2xx live-body read, mid-stream stall/abort at the provider layer, backpressure (`res.pause`/`res.resume` were inert stubs no test asserted — deleting the backpressure code would have failed nothing), and a **seam suite** driving the provider through the *real* `safeFetch` with only `http.request` faked, since every other suite mocks `safeFetch` and so proves intent rather than behavior.

**Controls verified by mutation, not assumed** — the two fixes whose failure mode is a hang rather than an assertion:

- Restoring `clearTimeout` to its pre-#4121 position makes the timeout test **hang** (15s timeout), not fail.
- Flipping `requirePrivateForCleartext` to `false` fails the seam suite's cleartext test.
- Removing the new `res.on('error')` listener fails the crash regression test.

**Tests:** 73 green on the changed files (`urlSafety.test.ts` 58, `openaiCompatibleProvider.test.ts` 15 — the provider had **zero** coverage before this PR). ~492 green across every `safeFetch` consumer suite (sso, webhookDelivery, webhookSender, guardedS3Client, backupSsrfAdoption, dnsProviders, recoveryMedia, huntress, logForwarding, psa, sentinelOne, guardedLlmFetch, llmEgressProxy, providerFidelityHarness, llmConfigResolver, backup/configs, llmProviderCatalog, tripwire). `eslint` clean.

**CI on `9fa0f1ba6`:** 57 checks, 57 completed, **0 failures** — including Type Check, Test API, Lint, Check Migrations, and all 4 Integration Tests shards.

**Not fixed here, deliberately — recommend a follow-up issue:** `openaiSessionManager.ts` calls `captureException` in its `catch` around the stream loop, but not in the `case 'error':` branch inside it. Provider-yielded failures therefore reach the user without reaching Sentry. Pre-existing, and in a file this PR does not touch — but this PR routes materially more real transport failures (SSRF refusals, `maxBytes` overruns, truncation, timeouts) through that exact branch, so the gap starts mattering.

**Status:** review-clean, CI green, awaiting maintainer merge. Note the self-hoster behavior change called out in the PR body — loopback `MCP_LLM_BASE_URL` is now refused.

PR #4326, opened 2026-08-31
fix(api): apply organizationOrder in SQL so it survives pagination (#4004)

### LINKED ISSUE #4004: [API] Organization list applies organizationOrder only within a page, so a cross-page order cannot be expressed
## Description

`GET /orgs/organizations` selects each page by `created_at, id` with `LIMIT`/`OFFSET`, and applies the partner's `organizationOrder` **only to the rows already selected into that page**. Past the first page the endpoint therefore cannot express the stored order at all.

## The gap

In `apps/api/src/routes/orgs.ts` the list query is:

```ts
const data = await db
  .select()
  .from(organizations)
  .where(conditions)
  .limit(limit)
  .offset(offset)
  .orderBy(organizations.createdAt, organizations.id);
```

and only afterwards:

```ts
ordered = applyOrganizationOrder(data, preferredOrder);
```

`applyOrganizationOrder` reorders `data` — the current page — so page membership is decided by `created_at, id` before the preferred order is consulted. An organization that `organizationOrder` puts first, but whose `created_at, id` places it on page 2, can never be returned at the top of page 1.

`apps/web/src/lib/fetchAllOrganizations.ts` walks pages of 100 and concatenates them, so the assembled client list is page-partitioned: correctly ordered within each 100-row block, wrong across block boundaries.

## Impact

Any partner with more than one page of organizations sees a drag-reorder that appears to work and then does not survive a reload once the move crosses a page boundary. This is the ordinary list render, not an error path.

Below one page (the common case today) the behaviour is correct, which is likely why it has not been noticed.

## What I verified, and what I did not

Verified by reading `origin/main`: the `.limit().offset().orderBy(createdAt, id)` chain, the `applyOrganizationOrder(data, preferredOrder)` call operating on the page, and the client's page-walk-and-concatenate.

**Not verified:** I did not reproduce this against a live instance with more than 100 organizations, and I have not checked whether any deployment currently has one. Severity depends on that.

## Possible fix

The preferred order has to participate in page selection rather than be applied afterwards — e.g. order by the position of `organizations.id` within `organizationOrder` (nulls last, then `created_at, id` as the tiebreaker) inside the SQL, so `LIMIT`/`OFFSET` walks the intended sequence. The existing `created_at, id` tiebreaker should stay: it is load-bearing for stable pagination (#3462), since `created_at` is `defaultNow()` and ties across a seed or bulk import.

## Context

Found while reviewing #3989, where a failed drag-reorder reconciles by re-fetching the list. That made the limit visible: the refetch is only authoritative up to the first page. #3989 documents the limit at its call site and does not attempt to fix it here, since this is a server-side query change.

ToddHebebrand:
Closes #4004

## The bug

`GET /orgs/organizations` selected each page with `ORDER BY created_at, id` +
`LIMIT`/`OFFSET`, and only afterwards handed **that page** to
`applyOrganizationOrder`. Page membership was decided before the partner's
stored `organizationOrder` was consulted, so the in-JS sort could only permute
rows a different sort had already chosen — an org the partner dragged to the top
could never leave page 2.

`apps/web/src/lib/fetchAllOrganizations.ts` walks pages of 100 and concatenates,
so the assembled client list was correctly ordered inside each 100-row block and
wrong across every block boundary. Below one page the behaviour was already
correct, which is why it went unnoticed.

Re-verified against `origin/main` (a48b840bb) before starting: still present.

## The fix

The preferred order is now the **leading ORDER BY term of the paginated query
itself**, so Postgres sorts the whole result set before slicing the page:

```sql
order by array_position(ARRAY[$2, $3]::uuid[], "organizations"."id") asc nulls last,
         "organizations"."created_at", "organizations"."id"
limit $4 offset $5
```

`array_position` returns the 1-based index or `NULL`, and an ASC sort puts NULLs
last — so orgs the partner never placed trail the ones it did, in registration
order. That is exactly what `applyOrganizationOrder` computed, now applied to the
whole result set instead of one page. The `created_at, id` tiebreaker stays and
`id` remains mandatory (#3462).

The query moved into `routes/orgs.listQuery.ts` (the `routes/incidents.helpers.ts`
pattern) so the ORDER BY / LIMIT relationship can be asserted on the **compiled**
statement without a database.

### Details worth a reviewer's eye

- **The partner-settings read moved ahead of the page query**, since it now
  supplies the sort key. Its soft-fail is unchanged in kind: a failed settings
  read logs, captures to Sentry, and degrades to `created_at, id` with the list
  still rendering. `orgs.test.ts`'s soft-fail case was updated to queue the
  throwing read in its new position and now additionally proves the list query
  still runs after it.
- **Stored ids are filtered to UUID shape before binding.** `partners.settings`
  is an unconstrained jsonb blob and the old in-JS sort simply never matched a
  junk entry; a `::uuid[]` cast over one raises 22P02 and would take the whole
  endpoint down. Covered by both a compiled-SQL case and an integration case.
- **Ids bind as N parameters inside `ARRAY[...]::uuid[]`**, not as
  `${ids}::uuid[]` — Drizzle's `sql` template spreads a JS array into positional
  parameters, so the natural-looking form fails at runtime with `cannot cast type
  record to uuid[]` (the trap documented at `routes/devices/core.ts:771`). This is
  precisely the class of bug a Drizzle-mock assertion cannot see, hence the
  compiled-SQL suite.
- **`applyOrganizationOrder` is deleted**, not left unused. Keeping a
  post-pagination sort helper around is how this bug comes back; the module
  header now says so.
- **Tenancy unchanged.** Same `conditions` predicate, same request DB access
  context, no bare-pool query, no schema change and no migration — so no
  cascade/export-policy registration applies. Only the ORDER BY moved.
- **Performance.** `organizations` has no `created_at` index, so the pre-fix
  query was already a full sort — the new sort key costs nothing index-wise. The
  array is bounded by the partner's own org count (`PATCH /organizations/order`
  sanitizes against it, cap 10k), and the term only appears when a partner order
  exists, i.e. for a query already scoped to one partner.
- **Stale comment corrected** in `OrganizationsPage.tsx`: #3989 documented this
  limit at the reorder-reconciliation call site ("authoritative only up to the
  first page"). That is no longer true, and a stale caveat is worse than none.

## Tests

- `apps/api/src/routes/orgs.listQuery.test.ts` — **new**, DB-less compiled-SQL
  guard via `.toSQL()`. Asserts the emitted statement, not a Drizzle mock's call
  shape: the ordering term sits inside the ORDER BY and precedes both LIMIT and
  OFFSET; `created_at, id` follow it; ids bind as parameters rather than being
  interpolated; junk/duplicate/empty stored orders degrade correctly. Compiles at
  `offset: 2` on purpose — Drizzle omits `offset 0`, which would make half the
  assertions vacuous.
- `apps/api/src/__tests__/integration/organizationOrderPagination.integration.test.ts`
  — **new**, real Postgres. Five orgs, explicit `created_at` stamps, page size 2,
  preferred order reversed against `created_at` so **every** org has to cross a
  page boundary. Walks every page the way `fetchAllOrganizations` does and
  asserts the reassembled sequence, plus the unlisted-orgs-trail and junk-id
  cases. Single-page assertions pass against the broken code, so the fixture
  spans more than one page deliberately.

Red-first, verified rather than asserted: with `organizationOrderBy` stubbed back to `[created_at, id]`, the compiled-SQL suite failed 4/8 and the integration suite failed **4 of 5** — the one case that stayed green is the "no stored order" fallback, which correctly does not depend on the feature. Both go green with the fix (8/8 and 5/5).

### DIFF
--- a/apps/api/src/__tests__/integration/organizationOrderPagination.integration.test.ts
+++ b/apps/api/src/__tests__/integration/organizationOrderPagination.integration.test.ts
@@ -1,0 +1,185 @@
+/**
+ * Cross-page organization ordering (#4004) — integration.
+ *
+ * `GET /orgs/organizations` used to select each page by `created_at, id` with
+ * LIMIT/OFFSET and only then re-sort THAT PAGE by the partner's stored
+ * `organizationOrder`. Page membership was therefore decided before the
+ * preferred order was consulted, so an org the partner dragged to the top could
+ * never leave page 2. `apps/web/src/lib/fetchAllOrganizations.ts` walks pages of
+ * 100 and concatenates, which makes the assembled list correctly ordered inside
+ * each block and wrong across every block boundary.
+ *
+ * These cases only discriminate because the fixture spans MORE THAN ONE PAGE
+ * and the preferred order disagrees with `created_at` across that boundary. A
+ * single-page assertion passes against the broken code.
+ *
+ * The compiled-SQL companion (`routes/orgs.listQuery.test.ts`) pins the
+ * statement shape without a database; this file proves the shape actually
+ * executes on real Postgres and returns the sequence claimed — including that
+ * the `ARRAY[...]::uuid[]` parameter form is the one Postgres accepts (a
+ * `${jsArray}::uuid[]` spread raises "cannot cast type record to uuid[]" at
+ * runtime and no mock can see that).
+ *
+ * Prerequisites:
+ *   docker compose -f docker-compose.test.yml up -d
+ *
+ * Run:
+ *   pnpm test:integration -- src/__tests__/integration/organizationOrderPagination.integration.test.ts
+ */
+import './setup';
+
+import { Hono } from 'hono';
+import { eq, sql } from 'drizzle-orm';
+import { describe, expect, it } from 'vitest';
+import { organizations, partners } from '../../db/schema';
+import { orgRoutes } from '../../routes/orgs';
+import { createIntegrationTestClient, createOrganization } from './db-utils';
+import { getTestDb } from './setup';
+
+function buildApp(): Hono {
+  const app = new Hono();
+  app.route('/api/v1/orgs', orgRoutes);
+  return app;
+}
+
+interface ListResponse {
+  data: Array<{ id: string; name: string }>;
+  pagination: { page: number; limit: number; total: number };
+}
+
+/**
+ * Stamp an explicit `created_at` per org so the "registration order" the
+ * endpoint falls back to is deterministic. `created_at` is `defaultNow()` and
+ * `now()` is the TRANSACTION timestamp, so fixture rows can otherwise tie —
+ * which is exactly why `id` is a mandatory tiebreaker (#3462) and exactly what
+ * would make an ordering assertion flaky rather than wrong.
+ */
+async function stampCreatedAt(orgIds: string[]): Promise<void> {
+  const base = new Date('2026-01-01T00:00:00.000Z').getTime();
+  for (const [index, orgId] of orgIds.entries()) {
+    await getTestDb()
+      .update(organizations)
+      .set({ createdAt: new Date(base + index * 60_000) })
+      .where(eq(organizations.id, orgId));
+  }
+}
+
+async function setPreferredOrder(partnerId: string, orderedIds: string[]): Promise<void> {
+  // `organizationOrder` is plaintext inside the settings blob (the encrypted
+  // column registry only wraps nested remote-access credentials), so seeding it
+  // directly is faithful to what the PATCH /organizations/order route persists.
+  await getTestDb()
+    .update(partners)
+    .set({
+      settings: sql`COALESCE(${partners.settings}, '{}'::jsonb) || ${JSON.stringify({
+        organizationOrder: orderedIds,
+      })}::jsonb`,
+    })
+    .where(eq(partners.id, partnerId));
+}
+
+/** Walk every page the way `fetchAllOrganizations.ts` does and concatenate. */
+async function walkPages(
+  client: { get: (path: string) => Promise<Response> },
+  limit: number,
+): Promise<string[]> {
+  const ids: string[] = [];
+  let page = 1;
+  for (;;) {
+    const res = await client.get(`/api/v1/orgs/organizations?page=${page}&limit=${limit}`);
+    expect(res.status).toBe(200);
+    const body = (await res.json()) as ListResponse;
+    ids.push(...body.data.map((org) => org.id));
+    if (body.data.length < limit || ids.length >= body.pagination.total) break;
+    page += 1;
+    if (page > 20) throw new Error('page walk did not terminate');
+  }
+  return ids;
+}
+
+/**
+ * Five orgs under one partner, `created_at`-ascending, all reachable by the
+ * returned partner-scoped client.
+ */
+async function seedFiveOrgs() {
+  const app = buildApp();
+  const client = await createIntegrationTestClient(app, { scope: 'partner' });
+  const partnerId = client.env.partner.id;
+
+  const extra = [];
+  for (let i = 1; i < 5; i += 1) {
+    extra.push(await createOrganization({ partnerId, name: `Ordering Customer ${i}` }));
+  }
+  const orgIds = [client.env.organization.id, ...extra.map((o) => o.id)];
+  await stampCreatedAt(orgIds);
+  return { client, partnerId, orgIds };
+}
+
+describe('GET /orgs/organizations — organizationOrder across page boundaries (#4004)', () => {
+  it('returns the preferred order across pages, not merely within each page', async () => {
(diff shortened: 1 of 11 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (correctness / tenancy / param-binding) + `pr-test-analyzer` (coverage quality). Both ran against the branch worktree, not a summary of the diff.

**Findings:** 2 raised, both real, both addressed in b64dcd0c2; 0 outstanding.

1. **Vacuous assertion** (`orgs.listQuery.test.ts`, "keeps created_at and id as tiebreakers AFTER the preferred-order term"). It compared `indexOf` results without first asserting the term was present — `indexOf` returns `-1` when absent, so `createdAtAt > -1` held for *any* ORDER BY and the case passed with the ordering feature removed entirely. In a file whose entire purpose is to not be vacuous, that mattered. Fixed by asserting presence first. The analyst found this by running the stub, not by reading.
2. **Untested system-scope branch** (`else if (auth.scope === 'system' && queryPartnerId)`). Attempted an HTTP-level integration case for it; it is **not reachable from the integration harness**, and the suite now records why instead of leaving a silent hole: `requireOrgRead` → `requirePermission(ORGS_READ)` → `getUserPermissions` resolves permissions through `organization_users` / `partner_users` keyed by the request's `partnerId`/`orgId`, and a system-scope token carries neither by design — so the caller is 403'd "No permissions found" before any ordering code runs. Promoting the fixture user to `is_platform_admin` clears the separate SR2-02 live-binding check (`middleware/auth.ts:591`) but not this one; there is no platform-admin bypass in `getUserPermissions`. Closing that branch needs a harness change (a system-reachable role), not a change to this suite. Flagged rather than faked.

`code-reviewer` returned **no findings** in all five categories it was asked for: SQL-vs-JS ordering semantics (duplicates, missing/stale ids, NULL placement, tiebreaker order), tenancy (`conditions` and DB access context unchanged, no bare-pool query), the read-reordering's effect on the soft-fail and `noLiveOrgs`/archived branches, `ARRAY[...]::uuid[]` parameter-binding safety, and whether the reordered unit-test mock queues now pass vacuously.

**Red control (verified, not asserted):** stubbing `organizationOrderBy` back to the bare `created_at, id` tiebreaker fails the compiled-SQL suite **5/8** (was 4/8 before finding 1 was fixed — the delta *is* that fix) and the integration suite **4/5**. The cases that stay green in each are the genuine no-stored-order fallback paths.

**Tests** (all foreground, this branch):
- `routes/orgs.listQuery.test.ts` — 8 passed.
- `__tests__/integration/organizationOrderPagination.integration.test.ts` — 5 passed, real Postgres.
- Affected route/service suites (`orgs`, `organizations`, `orgOrdering`, `orgArchive`, `orgMerge`, `partner_multi_org_orgid`, `orgs.listQuery`) — **386 passed**, 0 failed.
- `tsc --noEmit --project apps/api/tsconfig.json` — exit 0, no output. (It caught a real `readonly []` error in the first draft, fixed in 55087f784.)
- `eslint` clean on every changed API and web file.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4004 — reporter/maintainer call.

ToddHebebrand:
**CI follow-up:** the first full run went red on **Test API** — 1 failed / 29655 passed. Fixed in 4cd29aaab.

`src/db/offsetPaginationOrdering.test.ts` (the #3462 guard) scans for a literal `.id` inside the `orderBy(...)` of any statement that also calls `.offset(...)`. Moving the ordering behind `...organizationOrderBy(params.preferredOrder)` hides that token from a textual scan, so the guard flagged this query — correctly on its own terms, since it cannot see through a helper. The tiebreaker was never actually removed.

Registered in `ALLOWED_WITHOUT_LITERAL_ID`, which is the path the guard's own header prescribes for "an ordering built by a shared helper", matching the existing `routes/devices/core.ts` and `routes/tickets/tickets.ts` entries. Chose that over padding the ORDER BY with a duplicate trailing `id`.

The exemption is not taken on prose alone — that is the failure mode an allowlist invites. **Verified by regression:** dropping `organizations.id` from *both* branches of `organizationOrderBy` fails 3 cases in `routes/orgs.listQuery.test.ts`, which asserts the tiebreaker on the compiled SQL. So the exemption moves enforcement from a textual scan to a stronger check rather than removing it.

Worth noting for anyone reviewing: my local runs could not have caught this. The guard lives in `src/db/` and only fires on a repo-wide scan, so an affected-files run is blind to it — the same blind spot as the cascade-list contracts in CLAUDE.md. I have since run `src/db/` (66 files, 2309 tests) green locally.

**Also green on that run, before the failure:** all four `Integration Tests` shards passed, which is the real CI-side proof of the new cross-page suite, and `Type Check` passed (4m3s). Re-verified locally after this commit: `src/db/` + affected route/service suites 305 passed, integration suite 5 passed, `tsc --noEmit` exit 0.

PR #4333, opened 2026-08-31
fix(ci): discriminate ESRCH from EPERM in the ee boot-step kill handler (#3471)

### LINKED ISSUE #3471: ci: discriminate ESRCH vs EPERM in the ee boot-step kill handling

ToddHebebrand:
## Problem

The `Boot API once to apply built-in extension migrations (ee)` step (integration-test shard 1) folded both `kill` failure modes into a single branch:

```bash
if ! kill -- "-$PGID" 2>/dev/null; then
  echo "API process group had already exited after logging the built-in load line — dumping its output:"
  ...
  echo "::warning::API exited on its own after loading built-in extensions ...; migrations were applied, continuing."
fi
```

`2>/dev/null` discards the errno entirely, so two opposite situations get the same handling:

- **ESRCH** (`No such process`) — the group is gone. Benign: the API exited on its own between logging the built-in load line and this kill, and its migrations already landed.
- **EPERM** (`Operation not permitted`) — the group is **alive** and refuses the signal. Not benign at all: a live API keeps writing to this shard's DB and poisons the workspace integration suite that runs next, which is the exact flake this teardown exists to prevent.

**Observed behaviour on the EPERM path** (reproduced with the branch neutralised — see Testing):

```
API process group had already exited after logging the built-in load line — dumping its output:
::warning::API exited on its own after loading built-in extensions (no startup failure logged); migrations were applied, continuing.
API process group 4242 survived SIGTERM and SIGKILL. ...
```

Both the `already exited` line and the `migrations were applied, continuing` warning are **false** for a process that is still running, and the step then burns the full 30 s + 10 s settle budget before failing with a diagnosis — "survived SIGTERM and SIGKILL" — naming signals it never successfully delivered. Anyone reading that log is sent down the wrong path.

Worse, the downstream survivor assert is the *only* thing that catches it, and it depends on `ps` being able to enumerate the group. Where it cannot — hidepid, a PID namespace, a group owned by another user — the assert sees nothing and **the step passes outright with a live API against the integration shard's DB**. That is the fail-open this issue is about.

## Fix

EPERM now fails immediately, naming the PGID and quoting the errno text:

```bash
if [ "$(live_in_group "$pgid")" -gt 0 ] || printf '%s' "$kill_err" | grep -qi 'not permitted'; then
```

Design notes:

- **Ground truth is the process table, not `kill`'s stderr.** `strerror()` output is locale-dependent, so the errno string is quoted for the human reading the log but never trusted as the sole discriminator.
- **The text match is a second, independent trigger**, so an EPERM still fails loudly on exactly those hosts where `ps` is blind. The two conditions can only ever agree *toward failing* — neither can wave a live group through, and neither can turn a genuine ESRCH into a failure (no survivors + no "not permitted" text ⇒ benign path, unchanged).
- **No new silent-failure path in the other direction:** the ESRCH branch keeps its existing behaviour, including the fatal-sentinel check that still fails the step when an already-exited group also logged `[CRITICAL] API startup failed`.

### Why the teardown moved into a script

The teardown (SIGTERM → bounded grace → SIGKILL → fail closed) now lives in `scripts/ci/stop-ee-boot-group.sh`; `ci.yml` calls it and keeps only `wait "$BOOT_PID"` inline (`wait` only works in the shell that spawned the job).

While this logic lived in the `run:` block, the only thing a test could assert was the workflow's *text*. That is how the ESRCH/EPERM conflation survived two earlier hardening passes on this same step. The script is both executable and sourceable, so tests can call `stop_ee_boot_group` directly with `kill`/`ps`/`sleep` shadowed; sourcing runs nothing and does not mutate the caller's shell options. All existing comments and semantics (non-zombie liveness counting, group-not-leader polling, bounded grace, fail-closed assert) are carried over verbatim.

## Testing

`apps/api/src/config/eeBootStopGroup.test.ts` — 9 tests, in the blocking **Test API** job. The two errno branches are asserted **separately**:

| Case | Expected |
|---|---|
| EPERM, group live | exit 1, names the PGID and the errno, **no** `::warning::`, **no** "already exited" |
| ESRCH, group gone | exit 0, benign `::warning::`, **no** EPERM wording |
| ESRCH + fatal sentinel in log | exit 1 |
| Signals land, group survives | exit 1, "survived SIGTERM and SIGKILL" |
| Signals land, group gone | exit 0, quiet |
| Zombie-only group | exit 0 (treated as gone, not a survivor) |
| **Real** live process group (`set -m`) | exit 0, actually killed |
| **Real** already-exited group | exit 0, benign ESRCH path |
| `ci.yml` wiring | step delegates to the script; teardown cannot drift back inline |

EPERM cannot be provoked without a process owned by another user, so it is driven with a shadowed `kill`; ESRCH and the success path are additionally exercised against **genuine** process groups so the real `kill`/`ps`/awk plumbing is covered too.

**Red-first evidence.** With only the discrimination neutralised (`if false; then`, mutation confirmed present in the file), exactly one test fails:

```
× fails loudly and names the PGID when the group exists but is not signalable (EPERM)
  → expected 'API process group had already exited …' to match /not permitted|permission|EPERM/i
Tests  1 failed | 8 passed (9)
```

The ESRCH control stays green throughout, proving the harness discriminates the two branches rather than failing vacuously. Restoring the fix: 9/9 green.

**Verification run** (all foreground, on the rebased branch):

- `vitest run src/config/ src/__tests__/tzPinCanary.test.ts` — **16 files, 524 tests passed** (the broader sweep of every `apps/api` contract test that parses `ci.yml`, not just the new file)
- `node --test .github/scripts/check-workflow-security.test.mjs` — 100/100 pass
- `node --test .github/scripts/docs-automation-retirement.test.mjs` — 5/5; `scripts/security/check-release-trust-allowlist.test.mjs` — 10/10
- `actionlint -shellcheck= .github/workflows/ci.yml` (the CI-equivalent invocation) — clean
- `shellcheck -S warning scripts/ci/stop-ee-boot-group.sh` — clean
- `scripts/security/check-supply-chain-hardening.sh` — pass
- `tsc --noEmit` and `eslint` — clean

## Notes

- The step's two **earlier** `kill` sites (the timeout and fatal-boot branches) are deliberately untouched: both are `|| true` cleanup immediately followed by `exit 1`, so the errno cannot change the outcome there.
- No collision with #4319 (merged while this was in flight): its `ci.yml` hunks are at lines ~1656/1719, this one is at ~1893. Branch is rebased onto current `origin/main`.
- Base: `origin/main` @ `e17e7b3a4`.

Closes #3471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1893,62 +1893,23 @@ jobs:
             exit 1
           fi
           PGID="$(cat /tmp/ee-boot.pgid 2>/dev/null || echo "$BOOT_PID")"
-          # Guarded like the two failure branches above: under this step's
-          # default `bash -e`, an unguarded `kill` on an already-exited
-          # group (e.g. the process crashed on its own between logging the
-          # built-in load line and this kill) would abort the step on a bare
-          # "process not found" with no log content, making a correct
-          # migration load look like an unexplained CI failure. Dump the log
-          # so it's diagnosable either way — and then decide: an already-dead
-          # group that ALSO logged the fatal sentinel is a genuine boot
-          # failure that merely happened to log the built-in line first, and
-          # must fail the step rather than be waved through.
-          if ! kill -- "-$PGID" 2>/dev/null; then
-            echo "API process group had already exited after logging the built-in load line — dumping its output:"
-            cat /tmp/ee-workspace-boot.log
-            if grep -q "$FATAL" /tmp/ee-workspace-boot.log; then
-              echo "The API logged the built-in load line and then FAILED startup — the schema this step is"
-              echo "supposed to guarantee may be incomplete, so the workspace integration suite below would"
-              echo "fail confusingly. Failing here instead."
-              exit 1
-            fi
-            echo "::warning::API exited on its own after loading built-in extensions (no startup failure logged); migrations were applied, continuing."
-          fi
-          # Bound the wait for graceful shutdown instead of an unbounded
-          # `wait "$BOOT_PID"`, which could otherwise eat the whole job
-          # ceiling if the process ignores SIGTERM or hangs mid-shutdown.
-          # Poll the GROUP, not $BOOT_PID: the leader (pnpm's wrapper) can
-          # exit seconds before the node/tsx descendants finish shutting
-          # down, and watching only the leader cuts the grace period short
-          # (observed on the first live run of this step). Count only
-          # NON-ZOMBIE members: `kill -0` "succeeds" against a zombie, but a
-          # zombie is dead — unreaped only because its parent died with it —
-          # and cannot touch the DB. The step's shell cannot reap
-          # grandchildren; init does so asynchronously after SIGKILL, so a
-          # zombie-only group must be treated as gone, not as a survivor
-          # (also observed live: `[node]`/`[esbuild] <defunct>` tripping a
-          # naive `kill -0 -- -PGID` assert).
-          live_in_group() {
-            ps -eo pgid=,stat= | awk -v p="$PGID" '$1 == p && $2 !~ /^Z/' | wc -l
-          }
-          for _ in $(seq 1 30); do
-            [ "$(live_in_group)" -gt 0 ] || break
-            sleep 1
-          done
-          kill -9 -- "-$PGID" 2>/dev/null || true
+          # Teardown (SIGTERM → bounded grace → SIGKILL → fail closed on a
+          # survivor) lives in scripts/ci/stop-ee-boot-group.sh rather than
+          # inline here, so its branches are reachable by a real test:
+          # apps/api/src/config/eeBootStopGroup.test.ts drives them with
+          # stubbed `kill`/`ps` and with genuine process groups. While this was
+          # inline YAML the only thing a test could assert was the workflow's
+          # text, which is how the ESRCH-vs-EPERM conflation (#3471) survived
+          # two earlier hardening passes on this same step. The script
+          # discriminates a group that is already gone (ESRCH — benign, its
+          # migrations landed) from one that is alive but refuses the signal
+          # (EPERM — a live API that would keep writing to this DB shard, so a
+          # hard failure naming the PGID).
+          bash scripts/ci/stop-ee-boot-group.sh "$PGID" /tmp/ee-workspace-boot.log "$FATAL"
+          # Reap the setsid leader. `wait` only works in the shell that spawned
+          # the job, so it cannot move into the script; it runs after the
+          # teardown has already proved the group is gone, so it cannot hang.
           wait "$BOOT_PID" 2>/dev/null || true
-          # Fail CLOSED on a genuinely surviving process group — but give
-          # SIGKILL a moment to land before judging (delivery is async).
-          for _ in $(seq 1 10); do
-            [ "$(live_in_group)" -gt 0 ] || break
-            sleep 1
-          done
-          if [ "$(live_in_group)" -gt 0 ]; then
-            echo "API process group $PGID survived SIGTERM and SIGKILL. A live API would keep writing to"
-            echo "this step's DB shard and poison the workspace integration suite that runs next."
-            ps -eo pid,pgid,stat,cmd | awk -v pgid="$PGID" '$2 == pgid' || true
-            exit 1
-          fi
 
       - name: Integration-test workspace (ee)
         if: matrix.shard == 1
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three lenses in parallel: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`.

**Findings:** 1 actionable raised → addressed in c2305aa64; 1 declined with reason; 0 outstanding.

**Addressed — test coverage (`pr-test-analyzer`, 7/10).** The EPERM guard is an OR of two deliberately independent signals:

```bash
if [ "$(live_in_group "$pgid")" -gt 0 ] || printf '%s' "$kill_err" | grep -qi 'not permitted'; then
```

My original EPERM test set **both** operands true, so it could not tell which one fired — the reviewer demonstrated that dropping the `live_in_group` clause left all 9 tests green. The script's whole design claim (process table is ground truth; the stderr text is an independent fallback for hosts where `ps` is blind) was therefore unverified. Fixed by adding one test per operand:

- ps-visible survivor + an **unrecognised** errno text → only `live_in_group` can catch it
- EPERM stderr + a `ps` that reports **nothing** → only the text match can catch it (the hidepid / PID-namespace host where the pre-fix code passed outright with a live API still on the DB)

Verified by mutating each operand out **separately**: each mutation now fails exactly its own test (1 of 14), never the other. Also added three subprocess tests, since every previous case ran the script *sourced* and none invoked it the way `ci.yml` actually does — this also settles the entrypoint anomaly the reviewer flagged as "not conclusively diagnosed" (it invokes and propagates correctly; 14/14 green).

**Declined — `ps`-blind survivor after SIGKILL (`silent-failure-hunter`, informational).** The post-SIGKILL survivor assert still trusts `ps` alone, so a host where `ps` cannot enumerate the group *and* the group survives SIGKILL would be waved through. Declining in this PR because:

1. It is **pre-existing and unchanged** — I diffed that block against `origin/main`; it is identical apart from mechanical `$PGID`→`$pgid` and `exit 1`→`return 1` renames. Not a regression introduced here.
2. The obvious fallback (`kill -0` as a second operand) is **not safe to add blind**: this step's own history records a naive `kill -0 -- -PGID` assert being tripped by `[node]`/`[esbuild] <defunct>` members on the Linux runner, which is precisely why the non-zombie `live_in_group` counting exists. My local probe on a genuinely zombie-only group did *not* reproduce that (macOS), so the behaviour looks platform-dependent — which is a reason to treat it as its own change with its own evidence, not to bolt it on here.

Happy to file a follow-up issue if you want that closed.

**Other lenses:** `code-reviewer` — 0 findings at or above its confidence bar; confirmed line-by-line behaviour preservation of the extracted teardown, that the `wait "$BOOT_PID"` reordering is inert (`live_in_group` already excludes zombies), correct `$FATAL` quoting, and correct sourceable/executable dual-mode. `silent-failure-hunter` — 0 CRITICAL/HIGH; confirmed no benign ESRCH can become a hard failure, and that the remaining `|| true` suppressions are justified.

**Tests:** `apps/api/src/config/eeBootStopGroup.test.ts` — **14 passed**. Broader sweep (every `apps/api` contract test that parses `ci.yml`, not just the touched file): `vitest run src/config/ src/__tests__/tzPinCanary.test.ts` — **16 files, 529 passed**. Plus `check-workflow-security.test.mjs` 100/100, `docs-automation-retirement.test.mjs` 5/5, `check-release-trust-allowlist.test.mjs` 10/10. `tsc --noEmit` clean, `eslint` clean, `shellcheck -S warning` clean, `actionlint -shellcheck=` (the CI-equivalent invocation) clean.

**Red-first evidence:** with only the discrimination neutralised, the EPERM test failed while the ESRCH control stayed green — the two errno branches are asserted separately, not vacuously.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #3471 — those are yours.

PR #4348, opened 2026-09-01
feat(api): export retention and rollup job outcomes as Prometheus metrics (#4345)

### LINKED ISSUE #4345: [API] Export retention and rollup job outcomes as Prometheus metrics
## Description
The API's /metrics endpoint (already scraped per region) exposes no visibility into the ~20 retention jobs or the metric-rollup pipeline. DB-side monitoring added 2026-09-01 (postgres-exporter custom queries: table sizes, statement exec totals, txn age) can show *that* the database grows, but not *which* job fell behind or silently stopped. Several retention jobs can fail open: batched jobs with a `MAX_BATCHES` cap return `hasMore` and nothing re-enqueues (see #4343), and a BullMQ repeatable that stops firing is invisible today.

## Proposed Fix
Add counters/gauges via the existing metrics registry, labeled by job/table:
- `breeze_retention_rows_deleted_total{job_name}`
- `breeze_retention_last_run_timestamp_seconds{job_name}` (staleness alert: no run in >26h)
- `breeze_retention_backlog_incomplete{job_name}` (1 when a capped run returned hasMore)
- `breeze_rollup_runs_total{status}` / `breeze_rollup_duration_seconds`

Wire an alert rule on staleness + backlog once exported.

## Affected Files
- `apps/api/src/jobs/*Retention.ts`, `apps/api/src/jobs/metricRollupMaintenance.ts`, `apps/api/src/services/metricRollups.ts`
- metrics registry (`apps/api/src/services/metrics*`)

Related: #4342, #4343

ToddHebebrand:
Closes #4345

## What

The API's `/metrics` endpoint exposed no visibility into the retention job family or the metric-rollup pipeline. A retention worker that silently stopped deleting — or one permanently pinned at its `MAX_BATCHES` ceiling while its table kept growing — was unobservable until someone noticed the disk.

This adds five series, all through the metrics registry that `routes/metrics.ts` already owns:

| Series | Type | Labels |
|---|---|---|
| `breeze_retention_rows_deleted_total` | counter | `job_name` |
| `breeze_retention_last_run_timestamp_seconds` | gauge | `job_name` |
| `breeze_retention_backlog_incomplete` | gauge | `job_name` |
| `breeze_rollup_runs_total` | counter | `status` |
| `breeze_rollup_duration_seconds` | histogram | `status` |

`breeze_rollup_duration_seconds` is a Histogram with the same bucket ladder as the existing `breeze_s1_sync_duration_seconds`, per the issue's "match whatever pattern the registry already uses".

## How

New `apps/api/src/services/retentionMetrics.ts`, following the established recorder pattern (`services/actionIntents/metrics.ts`, `services/anomalyMetrics.ts`): a settable recorder so `jobs/*` can emit without importing `routes/metrics` — which pulls the whole metrics + backup graph and would close an import cycle — plus `registerRetentionPrometheusMetrics(registry)`, called once from `routes/metrics.ts` alongside the other `register*` calls.

Two helpers, `recordRetentionRun(jobName, { rowsDeleted, incomplete })` and `recordRollupRun(status, durationSeconds)`, replace what would otherwise be ~15 duplicated registration blocks. `job_name` is a closed union (`RETENTION_JOB_NAMES`), so the label set stays bounded.

Each job gets a one-to-two-line insertion into its existing success path. **No job control flow was changed.**

Both helpers swallow their own failures and log. This is load-bearing, not habit: every retention call site sits *after* the DELETE committed and immediately before the job's `return`, so a throw would turn a completed purge into a failed BullMQ job — and `rollupDeviceMetricsRange` calls the recorder from inside a `catch`, where a throw would replace the in-flight rollup error before the rethrow ran.

## Instrumentation coverage

| Job | rows-deleted | last-run | backlog |
|---|---|---|---|
| `deviceMetricsRetention` | yes | yes | — |
| `processSampleRetention` | yes | yes | — |
| `serviceProcessCheckRetention` | yes | yes | — |
| `agentLogRetention` | yes | yes | — |
| `changeLogRetention` | yes | yes | — |
| `ipHistoryRetention` | yes | yes | — |
| `aiUnattendedExposureRetention` | yes | yes | — |
| `auditRetention` | yes (`stats.rowsDeleted`) | yes | **yes** (any policy errored) |
| `mlOutputRetention` | yes | yes | **yes** (`hasMore`) |
| `userRiskRetention` | yes | yes | **yes** (`hasMore`) |
| `reliabilityRetention` | yes | yes | **yes** (`hasMore`) |
| `metricRollupMaintenance` | yes (summed per-tier) | yes | **yes** (any tier `hasMore`) |
| `eventLogRetention` | — | yes | **yes** (any org skipped) |
| `playbookRetention` | — | yes | **yes** (either half failed) |
| `snmpRetention` | — | yes | — |

`breeze_retention_backlog_incomplete` means "the run did not clear everything it was supposed to, so rows remain past the cutoff" — the issue's literal "did not finish". Two causes, identical to an operator: a batch cap was hit, or errored items were skipped. That second group matters because three jobs swallow per-item failures and continue; without it, a run in which *every* item failed still published a fresh last-run stamp and a 0-row counter, and a staleness alert built on those would never fire.

The rollup metrics wrap `rollupDeviceMetricsRange` in `services/metricRollups.ts` — the real per-org "one run" boundary that `jobs/metricRollups.ts` fans out every 5 minutes — rather than the BullMQ processor, which would also count the unrelated `scan-orgs` fan-out job. Status is `success` / `skipped` (ML feature gate off) / `failure` (rethrown, not swallowed).

## Deliberate omissions

Each of these avoids publishing a claim the code cannot back:

- **Jobs that discard their DELETE row counts** (`eventLogRetention`, `snmpRetention`, `playbookRetention`) record no rows-deleted series. A permanent `0` on the counter reads as "deleted nothing", which is a different claim from "not measured". The issue explicitly allowed this ("still record the `last_run_timestamp` gauge for it").
- **Nothing is seeded at boot.** Retention workers are `global` / `socket-owner` placements in `workerRegistry.ts`, so exactly one instance runs each. Seeding `last_run = 0` (the `fleetGaugeLastRefreshGauge` pattern) would publish a permanent "never ran" series on every replica that legitimately never runs the job, and any staleness alert built on it would fire forever. Alert with `absent()` or `max by (job_name)` across the fleet instead.
- **A lock-contended `metric_rollup_maintenance` run is not recorded.** It returns `skipped` with empty arrays; recording it would claim "ran, deleted 0, backlog clear" for work that never happened.
- **`jobs/backupRetention.ts` is out of scope** — it is a library of two separately-invoked sweeps (row-level snapshot cleanup and object-storage GC) with two independent totals driven from `backupWorker.ts`, and backup already has its own instrument set in `services/backupMetrics.ts`. Instrumenting it is a design call beyond this issue's stated list.

## Testing

Red-first: `apps/api/src/services/retentionMetrics.test.ts` was written and run failing before the implementation existed. 14 tests, asserting on real Prometheus samples read back out of a fresh `Registry` (not internal state).

The two guard tests added from review were **mutation-checked**: removing the try/catch makes both fail, confirming they discriminate rather than confirm.

- 12 affected files (retention/rollup job suites + `routes/metrics.test.ts`) — **148 passed, 0 failed**
- `pnpm --filter @breeze/api lint` — clean

No schema, migration, or tenancy surface is touched; this is pure instrumentation.

## Note on CI

Every job on this PR fails at `pnpm install`, and **it is not this change** — `pnpm-lock.yaml` on `main` has a duplicated `hono@4.13.5:` mapping key (lines 8397/8401 and 20508/20510), so CI rejects it with `ERR_PNPM_BROKEN_LOCKFILE`. This PR does not touch the lockfile. Main needs a lockfile regen before anything can go green.

https://claude.ai/code/session_01T62pzFYZhVXCEuo5gYEoUu

### DIFF
--- a/apps/api/src/jobs/agentLogRetention.ts
+++ b/apps/api/src/jobs/agentLogRetention.ts
@@ -15,6 +15,7 @@
 import { Queue, Worker, Job } from 'bullmq';
 import { sql } from 'drizzle-orm';
 import { getBullMQConnection } from '../services/redis';
+import { recordRetentionRun } from '../services/retentionMetrics';
 import { attachWorkerObservability } from './workerObservability';
 import { jobSchedule } from './scheduleRegistry';
 import {
@@ -72,6 +73,7 @@ export function createAgentLogRetentionWorker(): Worker<RetentionJobData> {
       const durationMs = Date.now() - startTime;
       console.log(`${LOG_PREFIX} Pruned ${deletedCount} agent logs older than ${retentionDays} days (batches=${batches}) in ${durationMs}ms`);
       warnOnRetentionBacklog(LOG_PREFIX, 'agent_logs', { deleted: deletedCount, batches, hasMore });
+      recordRetentionRun('agent_log_retention', { rowsDeleted: deletedCount, incomplete: hasMore });
 
       return { durationMs, deletedCount, retentionDays, batches, hasMore };
     },
--- a/apps/api/src/jobs/aiUnattendedExposureRetention.ts
+++ b/apps/api/src/jobs/aiUnattendedExposureRetention.ts
@@ -23,6 +23,7 @@ import { sql } from 'drizzle-orm';
 import * as dbModule from '../db';
 import { extractRowCount } from '../db/rowCount';
 import { getBullMQConnection } from '../services/redis';
+import { recordRetentionRun } from '../services/retentionMetrics';
 import { attachWorkerObservability } from './workerObservability';
 import { jobSchedule } from './scheduleRegistry';
 
@@ -57,6 +58,7 @@ export async function pruneAiUnattendedExposure(): Promise<RetentionJobResult> {
   const durationMs = Date.now() - startedAt;
 
   console.log(`[AiUnattendedExposureRetention] Pruned ${deletedCount} exposure-ledger rows older than ${RETENTION_HOURS}h in ${durationMs}ms`);
+  recordRetentionRun('ai_unattended_exposure_retention', { rowsDeleted: deletedCount });
   return { deletedCount, durationMs };
 }
 
--- a/apps/api/src/jobs/auditRetention.ts
+++ b/apps/api/src/jobs/auditRetention.ts
@@ -71,6 +71,7 @@ import {
 import { extractRowCount } from '../db/rowCount';
 import { captureException } from '../services/sentry';
 import { getBullMQConnection } from '../services/redis';
+import { recordRetentionRun } from '../services/retentionMetrics';
 import { jobSchedule } from './scheduleRegistry';
 
 const QUEUE_NAME = 'audit-log-retention';
@@ -635,6 +636,17 @@ export async function pruneExpiredAuditLogs(
       reason: 'backlog_ceiling',
     });
   }
+
+  // `stats.errors` counts policies whose DELETE threw and were skipped, so their
+  // rows are still past the cutoff; `stats.orgsWithBacklog` counts policies that
+  // hit the batch ceiling with more rows below the cutoff. Either means rows
+  // remain — without this the degenerate run (every policy failed, rowsDeleted
+  // 0, fresh last-run stamp) looks perfectly healthy on the dashboard, and only
+  // Sentry knows otherwise.
+  recordRetentionRun('audit_retention', {
+    rowsDeleted: stats.rowsDeleted,
+    incomplete: stats.errors > 0 || stats.orgsWithBacklog > 0,
+  });
   return stats;
 }
 
--- a/apps/api/src/jobs/changeLogRetention.ts
+++ b/apps/api/src/jobs/changeLogRetention.ts
@@ -14,6 +14,7 @@ import { Queue, Worker, Job } from 'bullmq';
 import { sql } from 'drizzle-orm';
 import { captureException } from '../services/sentry';
 import { getBullMQConnection } from '../services/redis';
+import { recordRetentionRun } from '../services/retentionMetrics';
 import { jobSchedule } from './scheduleRegistry';
 import {
   parsePositiveIntEnv,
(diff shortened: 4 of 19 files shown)


ToddHebebrand:
**Review run:** `code-reviewer` + `silent-failure-hunter` (one round, per the rigor calibration for instrumentation-only work).

**Findings:** 5 raised → all 5 addressed in 6b27f7d3a; 0 outstanding.

- `code-reviewer` — **0 findings.** It independently verified the things most likely to be wrong here: every `recordRetentionRun` sits on the single unconditional success-return path; each core prune function (`pruneExpiredAuditLogs`, `pruneMlOutputs`, `compactUserRiskSnapshots`, `pruneReliabilityHistory`, `pruneAiUnattendedExposure`) has exactly one caller, so no double-recording path exists; and the `job_name` literals across the 15 job files are an exact bijection with `RETENTION_JOB_NAMES`.
- `silent-failure-hunter` — **4 findings, 2 of them real bugs in code this PR introduced:**
  1. **(HIGH)** `recordRollupRun` was not exception-safe, and `rollupDeviceMetricsRange` calls it *from inside a `catch`*. A throw there would have replaced the in-flight rollup error before `throw error` ran — discarding the actual failure and leaving only a metrics-library stack trace. Fixed.
  2. **(HIGH)** The same unenforced contract at all 15 retention call sites, each sitting *after* the DELETE committed and immediately before the job's `return` — a throw would have flipped a completed purge into a failed BullMQ job. Guarded once at the source in `recordRetentionRun`/`recordRollupRun` rather than at 16 call sites.
  3. **(HIGH)** `auditRetention` swallows per-policy failures and continues, then recorded a run that looked healthy. The degenerate "every policy failed" case published a fresh last-run stamp and a 0-row counter, so a staleness alert would never fire and only Sentry knew. Fixed.
  4. **(MEDIUM)** Same shape in `eventLogRetention` (skips orgs whose retention lookup or DELETE threw) and `playbookRetention` (two independently-failing halves). Fixed.

Findings 3 and 4 were resolved **without adding a sixth series**: `breeze_retention_backlog_incomplete` now carries the issue's literal "did not finish" meaning — rows remain past the cutoff, whether from a batch cap or from errored items being skipped. Help text updated to match.

**Tests:** 12 affected files green — 148 passed, 0 failed (`vitest run --pool=threads --maxWorkers=2`, apps/api). `tsc --noEmit` clean; `pnpm --filter @breeze/api lint` clean. The helper suite was written red-first, and the two guard tests added from this review were **mutation-checked** — removing the try/catch makes both fail, so they discriminate rather than merely confirm.

**Status:** review-clean.

⚠️ **CI cannot go green on this PR, and not because of this change.** Every job fails at `pnpm install` with `ERR_PNPM_BROKEN_LOCKFILE`: `pnpm-lock.yaml` on `main` has a duplicated `hono@4.13.5:` mapping key (lines 8397/8401, again at 20508/20510) from the recent dependabot bump series (#4306, #4307, #4308, #4310, #4312). This branch does not touch the lockfile. **Main needs a lockfile regen before this or any other PR can pass** — filing/fixing that is outside this issue's scope, flagging it here so it isn't mistaken for a defect in this PR.

PR #4352, opened 2026-09-01
fix(viewer,agent): deliver Paste Text as literal Unicode, not US-layout keystrokes (#4089)

### LINKED ISSUE #4089: Viewer Paste Text mangles shell/commands (spaces and punctuation corrupted)
## Summary

Viewer **Paste Text** into a remote macOS session corrupts the pasted string so shell commands are unusable (spaces removed, `/` and `=` altered, casing mangled).

## Repro

1. BRD into a Mac (observed on **Breeze Viewer** → agent **0.108.0** darwin-arm64, self-host API **0.108.0**).
2. Open TextEdit (or Terminal) on the remote.
3. Use Viewer **Paste Text** with a multi-line shell block, e.g. inventory commands containing paths like `/usr/local/bin/...`, `$(id -u)`, `egrep` pipes, etc.

## Result

Pasted text arrives mangled, for example conceptually:

- spaces stripped (`echo "===== HOST ==="` → `echo"=====HOST=====`)
- `/` turned into `?` or similar (`/usr/local/bin` → `?USR?LOCAL?BIN?`)
- `=` → `+`
- aggressive uppercasing

Cannot reliably paste admin commands via this feature.

## Expected

Paste Text should deliver the exact Unicode/ASCII string the tech submitted (or document intentional transforms). Shell paths and operators must survive unchanged.

## Related

- Surfaced during #3380 canary validation (used Paste Text only after BRD was confirmed working).
- Separate but adjacent: Remote Terminal multi-line paste/input also interleaved/garbled in the same session window (will note on a terminal issue / reopen track if needed). Historical: #3163 / #2870 for PTY garble classes.

ToddHebebrand:
Fixes #4089

## The bug

Viewer **Paste Text** into a remote macOS session corrupted the pasted string — spaces stripped, `/` arriving as `?`, `=` as `+`, aggressive uppercasing — so shell commands were unusable.

Paste was never a text transfer. The viewer walked the clipboard and emitted one synthetic `key_press` **per character** (`textToKeyEvents`, `apps/viewer/src/lib/paste.ts`), and the macOS agent replayed each one through a hardcoded **US-ANSI QWERTY virtual-keycode table** (`keyNameToKeycode`, `agent/internal/remote/desktop/input_darwin.go`). Two defects compound there:

1. **No layout adaptation.** A keycode that means `/` on US ANSI renders as whatever the remote machine's *active* input source maps to that physical key.
2. **Additive-only modifier flags.** `CGEventSetFlags` was called *only* when a modifier was requested and never explicitly cleared, so a modifier still held from an earlier event bled into later keystrokes.

## The fix

Paste stops round-tripping through keycodes. The viewer sends the **literal string** and the agent injects it with the platform's Unicode primitive — `CGEventKeyboardSetUnicodeString` on macOS, `KEYEVENTF_UNICODE` on Windows — which is layout-independent by construction. **Live keystroke synthesis is untouched**; only the paste path changes.

### Protocol

New `type_text` control message plus an `input_capabilities` request/response handshake, mirroring the existing `list_monitors` pattern.

Two deliberate choices:

- **It rides the `control` channel, not `input`.** The input channel is created with `maxRetransmits: 0` (`apps/viewer/src/lib/webrtc.ts:162`) — the right trade for a mouse move, the wrong one for a paste, where one dropped datagram silently truncates a shell command. The `type_text` handler is therefore the one control message that explicitly stamps the idle watchdog and wakes the capture loop, since a paste *is* operator presence.
- **The old path stays as the fallback.** The Tauri viewer auto-updates independently of the agent fleet, so a new viewer routinely talks to a 0.108.0-and-older agent. Those agents have no case for `input_capabilities` and no `default:` branch, so they never answer, the viewer's flag stays false, and Paste Text behaves exactly as it does today. Without that, this change would turn "mangled paste" into "silently no paste at all" for every un-upgraded agent.

The viewer waits up to 500 ms for the handshake reply before deciding, so a paste fired the instant a session opens doesn't lose the race and drop to the fallback.

### Agent

- **`input_text.go` (new, platform-agnostic, unit-tested on any OS)** — `InjectText` plus two pure helpers:
  - `splitTextSegments` keeps newlines and tabs as **real key presses**. A U+000A inside a Unicode-string event does not reliably submit a line in Terminal.app; a Return keystroke does, which is what makes a pasted multi-line shell block actually execute.
  - `chunkUTF16` is surrogate-pair-safe, since `CGEventKeyboardSetUnicodeString` takes a `UniChar` buffer and a split pair would arrive as two replacement characters.
  - Linux/XTEST has **no** Unicode injection primitive (`x11.Injector.KeyBySym` only resolves keysyms already on the current layout), so it falls back to today's US-layout key synthesis — unchanged behaviour, except it now *reports* characters it could not map instead of dropping them silently.
- **macOS** — `TypeText`/`TypeChar` via `CGEventKeyboardSetUnicodeString` with flags explicitly zeroed, and `CGEventSetFlags` is now called **unconditionally** on the `SendKeyPress` path. That fixes defect (2) for live typing too, and specifically stops a stale Shift landing on the Return between two pasted lines (Shift+Return inserts a line break rather than executing). `SendKeyDown`/`SendKeyUp` deliberately keep the old conditional — the viewer uses them to *hold* a modifier for Shift+Click, and zeroing flags there would drop the modifier it just asked to hold.
- **Windows** — added a batched `TypeText`, and fixed two latent defects in the existing `TypeChar` that this PR would otherwise have started exercising on every pasted character:
  - it called `sendInput.Call(2, unsafe.Pointer(&down), …)` while `up` was a *separate local variable*. `SendInput` reads `cInputs` **contiguous** `INPUT` structs, so the second one was whatever memory followed `down` — the key-up was never reliably sent, and it was an out-of-bounds read of Go memory.
  - `ki.wScan = uint16(ch)` truncated any rune above U+FFFF instead of sending its UTF-16 surrogate pair.
- **AI computer-use** — the `type` action now shares `InjectText`, so macOS AI typing also stops going through the US-ANSI table. Behaviour change: a `\n` in a `type` action is now a real Return press.

### Drive-by

`apps/viewer/src/components/DesktopViewer.tsx` contained a **raw NUL byte** (a separator in a template literal), now written as the equivalent escape. That single byte made git classify the whole file as binary, so every diff on it rendered as "Binary files differ" and `grep` silently matched nothing on it. **This PR's diff for that file still shows as binary because the base blob is** — from the next change onward it will render normally. To review it: `git diff --text origin/main -- apps/viewer/src/components/DesktopViewer.tsx`.

### Failure reporting

The viewer's progress indicator completes on *send*, not on delivery, so a paste the agent dropped would otherwise look exactly like one that worked. The agent replies `type_text_result` **on failure only** — input unavailable at the login window, a malformed message, a partial `SendInput`, or characters the Linux fallback cannot map — and the viewer surfaces it as a transient toolbar notice. (`errorMessage` was deliberately not reused: it renders only behind the full-screen connection-error overlay, so a failed paste would have torn down a working session.)

Agent-side errors report **counts, never the characters themselves** — a paste can carry a password, and those errors reach the agent's log.

## Verification

| Check | Result |
|---|---|
| `go test -race ./internal/remote/desktop/...` | ok (incl. the new `input_text_test.go` + `session_type_text_test.go`) |
| `go test ./internal/remote/tools/` | ok |
| `go build` for darwin/arm64 (cgo), windows/amd64, linux/amd64 | all clean |
| `go vet` on the desktop package (all three GOOS) | no new findings |
| viewer `vitest run` | 20 files, 220 tests passed |
| viewer `tsc --noEmit` | clean |

The new Go and viewer tests were **mutation-checked** — each was re-run against a deliberately broken implementation (CRLF collapse removed, surrogate pair split, `type_text` path disabled, `InputAvailable` guard and activity stamps removed, `InjectText`'s dispatch priority reversed, `typeText: true` flipped, the capability gate's re-arm reset dropped, its timeout disabled) and confirmed to fail, so none of them passes vacuously.

Two pre-existing local failures were confirmed identical on `origin/main` and are unrelated: `internal/heartbeat` `TestHandleHttpRequest` (self-signed TLS in this sandbox) and a `go-m1cpu` cgo init crash in `internal/remote/tools` under Go 1.27 on arm64 (passes with `CGO_ENABLED=0`).

## Not in scope (flagged, not fixed)

- **`key_press` is missing from the desktopWs zod enum** (`apps/api/src/routes/desktopWs.ts:64`), so paste-as-keystrokes is *already* fully dropped on the WebSocket fallback transport today — a separate pre-existing bug. The WS and VNC transports have no control channel, so they keep today's behaviour here and `type_text` was deliberately not added to the WS/heartbeat input path.
- **macOS Secure Input** (a `sudo` password prompt in Terminal.app) blocks all CGEvent injection, the Unicode path included. Not a regression — the old keycode path was blocked identically — but worth knowing given the motivating scenario.

## Testing needed on real hardware

The CGEvent calls cannot be exercised in CI. Worth a canary pass on the arm64 Mac from the original report: paste a multi-line shell block into Terminal and TextEdit and confirm paths, `=`, quotes and casing survive, on both a US and a non-US input source.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer. A separate read-only design critique ran before implementation (Codex was unavailable in this environment — 401 — so the independent opinion came from a subagent instead).

**Findings:** 6 raised → 6 addressed in e5c5ae1; 0 outstanding.

- *code-reviewer:* no confirmed bug at ≥80 confidence. It independently verified the cgo pointer passing, the `setFlags` modifier-hold contract (`SendKeyDown`/`SendKeyUp` unchanged), the Windows `SendInput` rewrite, and the reconnect-safe capability handshake. Two minor notes — the unread `unicodeText` field (**removed**) and pre-existing tab indentation in the touched block (left as-is, it matches the surrounding lines).
- *silent-failure-hunter (critical):* every agent-side paste failure ended in a local `slog.Warn` and nothing else, while the viewer's progress indicator completes on *send* — so a dropped paste looked identical to a successful one. **Fixed:** the agent now replies `type_text_result` on failure and the viewer shows a transient toolbar notice. It also flagged reusing `errorMessage`; that turned out to be wrong for this case (it only renders behind the full-screen connection-error overlay, which would tear down a working session), so the notice went in the toolbar instead.
- *pr-test-analyzer:* three real gaps. `sendInputCapabilities`'s payload survived flipping `typeText` to `false`; `InjectText`'s dispatch priority survived reversing the switch because no test double implemented both `TextTyper` and `TypeCharHandler`; and the viewer's handshake wiring had no coverage at all. **Fixed:** extracted `buildInputCapabilities`/`typeTextResult` and pinned their wire shapes, added a both-interfaces test double, and extracted the handshake into `lib/inputCapabilities.ts` with unit tests for the reconnect ordering and the timeout. Its suspected reconnect race was traced and refuted — React runs an effect's cleanup before the next setup — but the extraction stands on testability grounds.

The design critique also caught the one thing the original patch got wrong: `InjectText`'s Return/Tab presses still went through the `CGEventSetFlags`-only-when-nonzero path, so a stale Shift on the Return between two pasted lines would insert a line break instead of executing the command. That's fixed in the first commit.

Also tightened while in there: agent errors reported the skipped characters themselves and named the character a `TypeChar` call failed on. A paste can carry a password and those errors reach the agent's log, so they now report counts only.

**Tests:** `go test -race ./internal/remote/desktop/...` ok; `go test ./internal/remote/tools/` ok; `go build` clean for darwin/arm64 (cgo), windows/amd64 and linux/amd64; `go vet` no new findings; viewer 20 files / 220 tests passed; `tsc --noEmit` clean. Every new assertion was mutation-checked against a deliberately broken implementation and confirmed to go red. All 55 CI checks green on e5c5ae1.

**Needs a human before merge:** the CGEvent and `SendInput` calls cannot run in CI, so this wants a canary pass on real hardware — paste a multi-line shell block into Terminal and TextEdit on the arm64 Mac from the report, with the remote input source set to a **non-US** layout, and confirm paths, `=`, quotes and casing survive and each line actually executes. On Windows, paste something over 64 UTF-16 units containing an astral character to exercise the chunking and the `SendInput` fix.

Issue #387, opened 2026-04-11
Agent: desktop helper reconnect loop storms logs on headless Windows Server (Session 0 only)


ToddHebebrand:
## Summary

On a **headless Windows Server** (no interactive user logged in, Session 0 only — typical VPS / cloud-hosted Windows Server), the Breeze desktop helper enters a **tight reconnect loop at ~15 attempts/minute**, logs the same warning 1,000+ times per day, and the parent agent stops heartbeating — marking the device **offline** even though the agent process is still alive and shipping logs.

Reproduced on a production customer's Contabo-style Windows Server 2016 Datacenter VPS running agent `0.62.11`. The device went online at enrollment, stayed online for ~8 minutes, then offline ever since while generating **1,300+ "helper disconnected, reconnecting" warnings in 24h**.

## Symptoms

Warning distribution over a 24h window on the affected device:

| component | level | count | sample |
|---|---|---|---|
| `main` | warn | **1,542** | `helper disconnected, reconnecting` |
| `sessionbroker` | warn | 36 | `binary path verification failed`, `connection rate limited`, `auth missing SID on Windows` |
| `heartbeat` | warn | 9 | `patch inventory collection warning`, `security status collection warning` |

Sample log messages (consecutive, 165-200ms apart) from `main`:

```
16:28:53.780  helper disconnected, reconnecting
16:28:53.589  helper disconnected, reconnecting
16:28:53.424  helper disconnected, reconnecting
16:28:43.486  helper disconnected, reconnecting
16:28:43.479  helper disconnected, reconnecting
```

Device state:
- enrolled at T+0, last heartbeat at T+8min, offline ever since
- agent process still alive (still shipping logs) but no heartbeat
- `management_posture` field never populated (posture scanner never completed)

## Customer impact

- VPS appears **offline** in the Breeze UI even though it's healthy.
- Remote desktop, scripts, and commands cannot be targeted at the device (UI gates on online status).
- Breaks a core MSP use case — managing headless cloud servers — which is a major target workload.

## Root cause

The log line is emitted from [`runHelperProcess()` in `agent/cmd/breeze-agent/main.go`](../blob/main/agent/cmd/breeze-agent/main.go) around line 811, inside a reconnect loop that runs in the desktop-helper subprocess (`breeze-agent desktop-helper --context user_session`):

```go
const (
    helperMinBackoff = 2 * time.Second
    helperMaxBackoff = 30 * time.Second
)

backoff := helperMinBackoff
for {
[...]

PR #4353, opened 2026-09-01
fix(agent,macos): desktop helper self-reconnects instead of exiting on IPC failure (#4194)

### LINKED ISSUE #4194: [Agent] macOS desktop helper: exit on any IPC failure with no self-reconnect — availability rides entirely on launchd + an Aqua session
**Found while working #4092's lid-closed report.** The main agent and the desktop helper have opposite failure behavior:

- **Main agent** (root LaunchDaemon, no session limit): reconnect loop with exponential backoff (`agent/internal/websocket/client.go:333-391`) — survives sleep/wake and network blips, so the device flips back to online almost immediately after any wake.
- **Desktop helper** (`LimitLoadToSessionType: Aqua` LaunchAgent, `agent/service/launchd/com.breeze.desktop-helper-user.plist`): `Run()` connects once and enters the command loop (`agent/internal/userhelper/client.go:112-159`); any IPC failure returns an error and the process exits 1 (`agent/cmd/breeze-desktop-helper/main.go:146-149`). Recovery depends entirely on launchd KeepAlive/ThrottleInterval, which requires a live Aqua session to respawn into.

**Consequence:** any transient IPC hiccup (agent restart, socket blip, sleep/wake ordering) can strand `desktopAccess.reason = "helper_not_connected"` (`agent/internal/heartbeat/desktop_access_darwin.go:42,50,77,90`) while the device shows online — the #4092 symptom shape, minus the clamshell-sleep half (tracked there).

**Fix direction:** give the helper's IPC client the same reconnect/backoff loop the agent WS client has, instead of exit-on-error; keep hard-exit only for auth-rejection so a bad build doesn't spin against the socket.

Refs: #4092, #4162, #3380.

ToddHebebrand:
Fixes #4194

## The bug

The macOS desktop helper connected once and killed the process on any IPC error:

```go
if err := client.Run(); err != nil {
    log.Error("desktop helper error", "error", err)
    os.Exit(1)
}
```

Recovery therefore rode entirely on launchd respawning the LaunchAgent into a live Aqua session. Any transient gap — the agent restarting and recreating the socket, a sleep/wake ordering hiccup — stranded `desktopAccess.reason = "helper_not_connected"` while the device still showed **Online** and Desktop Access sat greyed out.

Meanwhile the main agent, and the Windows/Linux user-helper (`agentapp.runHelperProcess`), already had the right pattern: a reconnect loop with exponential backoff, jitter, and a stability-gated reset. **`cmd/breeze-desktop-helper` was a drifted single-shot copy of that same startup code with the loop left out** — the drift *is* the bug.

## The fix

Extract the proven loop into `internal/userhelper.Supervisor` (policy-parameterised, `os.Exit`-free, table-tested) and run **both** paths on it, so there is one implementation instead of two that can drift again.

Exit-code policy stays with each caller, because it genuinely differs per platform. The Windows/Linux semantics are preserved exactly: `not_desired` → exit 0, other permanent rejects and `ErrSIDLookupFailed` → exit 2.

**macOS gets its own tuning.** The Windows 30s floor is deliberately conservative (most of its disconnects are permanent identity problems, #387). The macOS failure mode is the opposite — a transient gap that clears in seconds, where every second costs the user a greyed-out button:

| | Windows/Linux user-helper | macOS desktop helper |
|---|---|---|
| Min backoff | 30s | **1s** |
| Max backoff | 5m | **60s** |
| Stable-reset threshold | 60s | 60s |

Dialling a unix socket that does not exist fails instantly and costs nothing, so the eager floor is cheap.

### Not fighting launchd

`com.breeze.desktop-helper-user.plist` sets `KeepAlive: true` (unconditional) with `ThrottleInterval: 10`, so launchd respawns after **any** exit regardless of code. Two consequences shaped the design:

1. **No plist change.** The XML exists in **three** hand-synced copies — `agent/service/launchd/*.plist` (copied by `build-pkg.sh`), Go string constants in `service_cmd_darwin.go` (used by `service install` / `healLaunchdPlists`), and a third set in `heartbeat/handlers_desktop_helper.go` (runtime self-repair). Editing one silently diverges from the others, and a plist change only reaches machines that reinstall. A binary-only fix ships with normal agent auto-update. Worth a separate cleanup issue.
2. **Fatal errors get a 10-minute in-process cooldown before exiting.** Exiting immediately on a permanent rejection would turn one unfixable problem into a respawn every 10s — and since every start runs the startup capture probe, that is also a **permission-prompt storm** (the reporter notes each probe re-fires the Screen Recording dialog). Staying alive is what suppresses it: launchd only respawns a job that has exited. After the cooldown we exit and launchd grants one fresh attempt, so a corrected config is still picked up without a manual `kickstart`. The cooldown is interruptible by SIGTERM.

Keeping the process resident also means the startup probe now runs **once per login session instead of once per IPC blip** — a direct reduction in TCC prompts.

## Three hazards that only appear once the process stays resident

Found by auditing what the old `os.Exit` was silently cleaning up. All three are load-bearing: without them the reconnect loop does not actually work.

1. **`user.Current()` memoization would permanently poison auth.** The stdlib caches that call's result *including its error* in a process-wide `sync.Once`. A respawn used to clear it; now one early failed lookup would be replayed forever and every reconnect would fail auth. Now leads with `os.Getuid()` + `user.LookupId()` (neither cached), with `user.Current()` only as a fallback. Windows already bypassed `user.Current` for the same class of reason.
2. **Orphaned TCC check loops.** `RunTCCCheckLoop` watched `c.stopChan`, which the supervisor never closes — it discards the client rather than calling `Stop()`. The goroutine was stranded holding a dead `*ipc.Conn` for up to three 60-minute intervals, with a fresh orphan per reconnect. Per-`Run` goroutines now key off a channel `Run` closes on every return path.
3. **Data race on `c.conn`** between `connect()` and `Stop()`, which the supervisor's shutdown goroutine makes real. Both guarded.

## Other platforms (checked, not changed)

Per the issue's scope, darwin only — but I checked the rest:

- **Windows** runs `breeze-user-helper.exe` → `agentapp.Main` → `runHelperProcess`, so it was already looped and stays so. `cmd/breeze-desktop-helper` is macOS-only (`agent/Makefile:338`; `release.yml` gates the build on `matrix.goos == 'darwin'`).
- **Linux** has no desktop-capture helper yet (`spawnHelperForDesktop` returns `ErrLinuxDesktopHelperUnsupported`); its `breeze-agent user-helper` systemd unit already goes through the looped path *and* has `Restart=on-failure`.
- **Noted, not fixed** (pre-existing, unchanged by this PR): the package-level banner handle in `banner_windows.go` has no teardown tied to a dropped session — harmless on darwin, where `banner_other.go` stubs it out. And a script started before a reconnect becomes uncancellable on the new `Executor`; today's `os.Exit` orphans it just the same, so this is not a regression.

## Verification

- `go test -race` clean for `internal/userhelper`, `internal/agentapp`, `cmd/breeze-desktop-helper` — **0 data races**. Run in a `golang:1.26.6` Linux container because local cgo is broken by an unrelated `go-m1cpu` segfault that also reproduces on untouched packages at `origin/main`.
- Full `go test -race ./...` for the agent: only 4 failures, all pre-existing container-environment artifacts, confirmed by re-running them on clean `origin/main` and by `go list -deps` showing zero dependency on the touched packages.
- Builds: darwin/arm64 **with cgo** (the shipped config), plus darwin/amd64, windows/amd64, linux/amd64.
- `golangci-lint` v2.12.2: no new issues.
- **Mutation-checked**, not just green: reverting the TCC loop to `c.stopChan` turns the teardown test red, and putting `user.Current()` back in front turns the ordering test red. The backoff test also caught a real off-by-one in my own jitter seam.

The 9 `helperWarnLimiter` tests moved to `internal/userhelper` alongside the type. Net effect on coverage: the reconnect loop previously had **no direct test at all** on any platform — it now has 20.

---

## Review round (added in `65f438e5`)

Two reviewers (`code-reviewer`, `silent-failure-hunter`) raised 6 findings; all 6 addressed. The substantive one: **this change trades away a signal.** A permanently-broken helper used to die on every attempt, and the crash-loop *was* the alarm — a resident process retrying quietly looks healthy to launchd/systemd. Added a rate-limited ERROR escalation once a run has gone 15 minutes without **ever** authenticating.

Two suggestions were **declined**, deliberately:

- **Promoting a stuck helper to fatal** (rather than just escalating the log). A helper can legitimately fail to connect for a while at boot or during an agent upgrade; a false fatal costs a 10-minute dead window. The customer-facing signal is already correct throughout — `desktopAccess.reason = "helper_not_connected"` is derived from the broker's live session state, so the UI never lies during either a retry storm or the cooldown.
- **Broadening `IsFatalHelperError`** to cover more error shapes (empty username, invalid SID, `Accepted:false && !Permanent`). The `Permanent` flag is the broker's contract for exactly this decision; second-guessing it client-side would strand helpers the broker intends to have retry. Worth revisiting on the broker side if the duplicate-session-ID reject at `broker.go:2098` should be marked permanent — that one is bounded by the 45s keepalive reap today.
- **Replacing the `user.Current` fallback with a raw `getpwuid`.** The fallback is only reachable when `LookupId` fails on the process's own uid, and a stale username fails auth loudly rather than silently. The uid never comes from the fallback, so the broker's uid-based authorization cannot widen. Documented as accepted rather than left implied.

## Authorship note

Commit `4a3c882f` (`agent/internal/userhelper/client_stoprace_test.go`) was authored by a concurrent session sharing this worktree, not by the session that opened this PR. It is additive and closes a real gap — the `connMu` fix had no regression coverage — so it was kept, and its claim was re-verified independently: stripping the guard does produce `DATA RACE` under `-race`.


ToddHebebrand:
**Review run:** `/pr-review-toolkit` — `code-reviewer` (behaviour-preservation of the refactored Windows/Linux path) and `silent-failure-hunter` (the specific risk that a retry loop swallows failures that used to be loud). Codex was unavailable for the design quorum (CLI returned 401), so the second opinion came from the review agents instead.

**Findings:** 6 raised → 6 addressed in `65f438e5`; 0 outstanding.

Both reviewers independently flagged the same top item, which is the one worth calling out:

1. **The reconnect loop trades away a signal.** A permanently-broken helper used to die on every attempt, and the launchd/systemd crash-loop *was* the alarm. A resident process retrying quietly forever looks healthy to every service manager. Added a rate-limited ERROR escalation that fires only once a run has gone `StuckThreshold` (15m) without **ever** authenticating. Deliberately an escalation and **not** the fatal promotion the reviewer suggested — a helper can legitimately fail to connect for a while at boot or during an agent upgrade, and a false fatal costs a 10-minute dead window. Flagging the judgement call explicitly since it is a partial decline.
2. `agentapp`'s fatal path hardcoded `code="sid_lookup_failed"` for anything that was not a `*PermanentRejectError`, so any future addition to `IsFatalHelperError` would be mislabelled in production logs. Now checks `errors.Is` explicitly with an `"unknown"` fallback, matching `supervise.go`.
3. Restored the `"helper stopped after error"` log distinction the refactor collapsed.
4. The macOS fatal cooldown is now slept in slices so the helper keeps saying why it is idle — during the hold `launchctl print` shows a live PID and no exit code, which reads as healthy.
5. `errcheck`: `Stop()`'s conn calls moved lines and so lost their grandfathered status under `--new-from-rev`. This is what reddened **Lint Agent (Go)** on the first push; now green.
6. Documented the accepted residual risk in `resolveUnixIdentity`'s `user.Current` fallback rather than leaving it implied.

**Declined, with reasons in the PR body:** broadening `IsFatalHelperError` to cover more error shapes (the `Permanent` flag is the broker's contract; second-guessing it here would strand helpers the broker intends to retry), and replacing the `user.Current` fallback with a raw `getpwuid` (the fallback is only reachable when `LookupId` fails on the process's own uid, and a stale username fails auth loudly rather than silently).

**Tests:** `go test -race` green for `internal/userhelper`, `internal/agentapp`, `cmd/breeze-desktop-helper` — 0 data races. Full `go test -race ./...` for the agent: only pre-existing container-environment failures, confirmed by re-running them on clean `origin/main` and by `go list -deps` showing zero dependency on the touched packages. Builds: darwin/arm64 **with cgo** (the shipped config) plus darwin/amd64, windows/amd64, linux/amd64. `golangci-lint` v2.12.2: no findings in any file this PR adds or changes.

Three of the four new contract tests were **mutation-checked**, not just observed green — reverting the TCC loop to `c.stopChan`, putting `user.Current()` back in front, and stripping `connMu` each turn the corresponding test red. The backoff test also caught a real off-by-one in my own jitter seam during development.

**CI:** all agent jobs green on `65f438e5` — Test Agent, Test Agent (Windows), Test Agent (race), Lint Agent (Go). 40 checks passing, 0 failing; the rest are web/API/integration jobs untouched by this Go-only diff.

**Note on authorship:** commit `4a3c882f` (`client_stoprace_test.go`) was authored by a concurrent session sharing this worktree, not by me. I kept it because it is additive and closes a genuine gap — the `connMu` fix had no regression coverage. I re-verified its claim independently rather than taking it on trust: stripping the guard does produce `DATA RACE` under `-race`.

**Status:** review-clean, CI green on the agent jobs, awaiting maintainer merge.

PR #4354, opened 2026-09-01
fix(ops): repoint verify-monitoring.sh at the real prod compose files (#4278)

### LINKED ISSUE #4278: [Ops] DEPLOY_PRODUCTION.md step 113 is broken: verify-monitoring.sh points at a nonexistent docker/docker-compose.prod.yml
`docs/operations/DEPLOY_PRODUCTION.md:113` sends operators to a command that cannot run:

```
./scripts/ops/verify-monitoring.sh .env.prod
```

`scripts/ops/verify-monitoring.sh:7` sets

```bash
COMPOSE_FILE="${REPO_ROOT}/docker/docker-compose.prod.yml"
```

but that file does not exist on `origin/main`. The production compose file is `deploy/docker-compose.prod.yml`. With `set -euo pipefail` at the top, the script fails as soon as it uses `$COMPOSE_FILE`.

**Impact:** the monitoring-verification step of the production deploy runbook has been broken since `3b723dc87`. `docs/notes/SOC_A1.1_CAPACITY_NOTES.md:146` additionally cites this script's output as SOC A1.1 baseline-health validation evidence — evidence that cannot currently be produced.

**Fix:** repoint `COMPOSE_FILE` at `deploy/docker-compose.prod.yml`, then actually run it against a deployment to confirm the rest of the script still matches reality (the wrong path means nothing downstream of it has executed in a long time, so there may be further drift behind it).

Found while deleting the orphaned `scripts/ops/backup.sh` / `restore.sh` and the `UPGRADE_v0.1_to_v0.2.md` runbook in PR #4268. Those three were genuinely dead. `verify-monitoring.sh` shares their broken compose path but is **not** dead — it has two live referrers — so it was deliberately kept and needs fixing rather than removing.

Worth checking as part of this: whether any other script under `scripts/ops/` or `scripts/` carries the same stale `docker/docker-compose.prod.yml` path.

🤖 Filed by Claude Code from the #3958 work; the missing compose file and both referrers verified against `origin/main`.

ToddHebebrand:
## Summary

`scripts/ops/verify-monitoring.sh` referenced a nonexistent `docker/docker-compose.prod.yml`, breaking step 3 of the production deploy runbook (`docs/operations/DEPLOY_PRODUCTION.md:113`) since `3b723dc87`, and the monitoring evidence cited by `docs/notes/SOC_A1.1_CAPACITY_NOTES.md:146`.

- **Root cause goes one step past the issue's suggested fix.** The real prod compose file is `deploy/docker-compose.prod.yml`, but repointing `COMPOSE_FILE` there alone is not sufficient: the script's `compose exec -T grafana ...` check needs the Grafana/Prometheus/Loki services, and those are defined in a separate overlay file, `docker-compose.monitoring.yml` (repo root) — not in `deploy/docker-compose.prod.yml` at all. `scripts/prod/deploy.sh` (the canonical deploy script) already composes both files together with `-f deploy/docker-compose.prod.yml -f docker-compose.monitoring.yml` for exactly this reason. This PR mirrors that same two-file pattern in `verify-monitoring.sh`.
- Also fixed the same stale `docker/docker-compose.prod.yml` path in three docs pages (`security/mtls.mdx`, `reference/schema.mdx`, `reference/troubleshooting.mdx`) — a repo-wide grep for the exact wrong path turned up no other occurrences (checked `scripts/`, `docs/`, and the docs site source).
- `scripts/docs-review/mapping.json` already references the correct `deploy/docker-compose.prod.yml` path, so no change needed there.

## Test plan

- [x] `bash -n scripts/ops/verify-monitoring.sh` — syntax OK
- [x] `grep -rn "docker/docker-compose\.prod\.yml"` repo-wide (excluding `.git`/`node_modules`) — zero remaining occurrences after the fix
- [x] Confirmed `deploy/docker-compose.prod.yml` exists and `docker/docker-compose.prod.yml` does not, on `origin/main`
- [x] Confirmed `docker-compose.monitoring.yml` defines `grafana`, `prometheus`, `loki` services (and `GRAFANA_PORT`/`PROMETHEUS_PORT`/`LOKI_PORT` env vars) that `deploy/docker-compose.prod.yml` does not
- [x] Confirmed the two-file `-f deploy/docker-compose.prod.yml -f docker-compose.monitoring.yml` pattern matches `scripts/prod/deploy.sh`'s existing, working `compose()` wrapper
- [ ] Not run against a live deployment (no prod/staging environment available in this sandbox) — the issue itself asks for this as a follow-up ("run it against a deployment to confirm the rest of the script still matches reality")

Closes #4278

### DIFF
--- a/apps/docs/src/content/docs/reference/schema.mdx
+++ b/apps/docs/src/content/docs/reference/schema.mdx
@@ -123,9 +123,9 @@ pnpm db:studio
 ## Inspecting the Database
 
 ```bash
-# Connect to the database
-docker compose -f deploy/docker-compose.prod.yml exec postgres \
-  psql -U breeze -d breeze
+# Connect to the database (prod uses a managed Postgres instance, not a
+# local container — connect with the DATABASE_URL from .env.prod)
+psql "$DATABASE_URL"
 
 # List tables
 \dt
--- a/apps/docs/src/content/docs/reference/schema.mdx
+++ b/apps/docs/src/content/docs/reference/schema.mdx
@@ -124,7 +124,7 @@ pnpm db:studio
 
 ```bash
 # Connect to the database (prod uses a managed Postgres instance, not a
-# local container — connect with the DATABASE_URL from .env.prod)
+# local container; run with .env.prod sourced: set -a; source .env.prod; set +a)
 psql "$DATABASE_URL"
 
 # List tables
--- a/apps/docs/src/content/docs/reference/schema.mdx
+++ b/apps/docs/src/content/docs/reference/schema.mdx
@@ -124,7 +124,7 @@ pnpm db:studio
 
 ```bash
 # Connect to the database
-docker compose -f docker/docker-compose.prod.yml exec postgres \
+docker compose -f deploy/docker-compose.prod.yml exec postgres \
   psql -U breeze -d breeze
 
 # List tables
--- a/apps/docs/src/content/docs/reference/troubleshooting.mdx
+++ b/apps/docs/src/content/docs/reference/troubleshooting.mdx
@@ -50,14 +50,14 @@ This is a **public** key (also embedded in the agent and CI), so it is safe to c
 
 **Cause**: PostgreSQL not ready or connection string incorrect.
 
-**Fix**:
+**Fix**: run from the host with `.env.prod` sourced (`set -a; source .env.prod; set +a`):
 ```bash
 # Check PostgreSQL is running (prod uses a managed Postgres instance, not a
 # local container)
 pg_isready -d "$DATABASE_URL"
 
 # Check connection string
-docker compose -f deploy/docker-compose.prod.yml exec api env | grep DATABASE_URL
+docker compose -f deploy/docker-compose.prod.yml --env-file .env.prod exec api env | grep DATABASE_URL
 ```
 
 ## Agent Issues
@@ -109,7 +109,7 @@ curl -H "Authorization: Bearer $TOKEN" \
   http://localhost:9090/api/v1/query?query=histogram_quantile(0.95,rate(http_request_duration_seconds_bucket[5m]))
 
 # Database slow queries (prod uses a managed Postgres instance, not a
-# local container)
+# local container; run with .env.prod sourced)
 psql "$DATABASE_URL" -c "SELECT * FROM pg_stat_activity WHERE state = 'active';"
 ```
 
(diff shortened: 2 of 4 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, 3 rounds — each round's findings addressed before re-review)

**Findings:**
- Round 1 (initial fix): Critical — the two-file `docker compose -f deploy/docker-compose.prod.yml -f docker-compose.monitoring.yml` overlay is an **invalid compose project**: `docker-compose.monitoring.yml`'s `postgres-exporter` hard-depends on a `postgres` service that only exists in the root (dev) `docker-compose.yml`, not in the prod stack (managed DB, no local postgres). Verified locally with `docker compose config` against a fully-populated dummy env — reproduced the exact failure. Fixed in `dcad5177d`/`664c07c8d` by dropping the compose overlay entirely and `docker exec`-ing into the pinned `breeze-grafana` container name instead.
- Round 2: Important — every `docker compose -f deploy/docker-compose.prod.yml ...` doc command was missing `--env-file .env.prod` (compose auto-loads `.env`, not `.env.prod`), and the Redis memory-check command would fail `NOAUTH` since prod Redis requires a password from a mounted secret. Fixed in `3a8c9bfdd`, matching `scripts/prod/deploy.sh`'s actual invocation pattern for both.
- Round 3 (final): 0 outstanding Critical/Important. One out-of-scope pre-existing bug found and filed separately as **#4362** (`docker-compose.monitoring.yml` is fundamentally not composable with `deploy/docker-compose.prod.yml` today — breaks `scripts/prod/deploy.sh` itself whenever `ENABLE_MONITORING=true`, its default; needs a change to the monitoring stack definition plus live-deployment verification, so it's scoped separately rather than folded into this PR). A second, smaller pre-existing issue was also surfaced (`security/secrets.mdx` has the same `--env-file`-missing and wrong-service-name patterns in unrelated, untouched command blocks) — flagging here rather than opening a third issue, since it's a one-line-per-command doc fix a maintainer may want to just take care of directly.

**Tests:** No test suite covers these files (shell script + docs). Evidence gathered instead: `bash -n scripts/ops/verify-monitoring.sh` (syntax OK); repo-wide grep confirms zero remaining `docker/docker-compose.prod.yml` references; reproduced the round-1 invalid-compose-project failure locally and confirmed the final `docker exec breeze-grafana` approach doesn't depend on that compose overlay at all; cross-checked the redis auth fix and `--env-file` placement against `scripts/prod/deploy.sh`'s actual working invocation pattern.

**Status:** review-clean (0 outstanding), awaiting maintainer merge. Not run against a live deployment — the original issue (#4278) itself asks for that as a follow-up once merged.

PR #4357, opened 2026-09-01
fix(api): dispatch SELF_UNINSTALL only after the device delete commits (#3817)

### LINKED ISSUE #3817: Device delete dispatches SELF_UNINSTALL before the transaction — agent uninstalls while the device row survives on 23503/55P03
**Where:** `apps/api/src/routes/devices/core.ts:1561-1571` sends the best-effort `SELF_UNINSTALL` command, and only then opens `db.transaction` at `:1584`. If the cascade aborts, the agent has already been told to remove itself while the device row survives.

**Reachable paths today:** the pre-existing `23503` branch (`:1599`) — FK violation, and it doesn't even report `uninstallSent` in its response — and, once #3760 merges, the new `55P03` lock-timeout branch (`:1622`), which does report it and whose 409 message is surfaced verbatim by `apps/web/src/services/deviceActions.ts:465`. Neither is new with #3760; the lock PR just makes the second one honest.

**Fix:** move the dispatch to after the transaction commits. `device.agentId` is already in scope post-transaction, so the move is mechanical, but it changes behaviour on three error paths and deserves its own test (uninstall NOT sent on 23503 / 55P03 / generic rethrow; sent exactly once on success). Also add `uninstallSent` to the `23503` response so the operator message is consistent across both failure branches.

Split out of #3760 at review so the lock-ordering fix doesn't carry a dispatch-ordering change. Low severity: deletion requires `decommissioned`, and the operator is told to retry.

ToddHebebrand:
Closes #3817

## The bug

`DELETE /devices/:id/permanent` sent the **irreversible** `SELF_UNINSTALL` command to the agent *before* opening the cascade transaction (`core.ts:1755` dispatch, `:1777` `db.transaction`). Every rollback path therefore left the agent removing itself while its device row survived:

| Path | Reachable | Reported it? |
|---|---|---|
| `23503` FK violation | pre-existing | yes (`uninstallSent` in the 409) |
| `55P03` bounded lock timeout | since #3760 | yes |
| generic rethrow (e.g. the `23514` from #3952) | pre-existing | log only — sanitized 500 to the caller |

The result in each case is an endpoint that is gone from the operator's fleet in practice, but still present, still billed, and no longer reachable to fix.

## The fix

Move the dispatch below the `try`/`catch` so it runs only on a **committed** cascade. `device.agentId` is already in scope post-transaction, so the move itself is mechanical.

The inverse race — the agent disconnecting between commit and send — degrades to the `warning` already in the success response plus a manual uninstall. That is recoverable; the other direction is not.

**Message changes.** With the hazard gone, the two 409s drop the "the uninstall command was already sent to the agent" disclosure that existed *for* it — keeping that text would now tell an operator their agent may be gone when it demonstrably is not. `uninstallSent` stays in both response bodies (now pinned to `false`) so a caller cannot confuse "not sent" with a shape change, and stays in the generic-rethrow log as the standing assertion that the command did not escape a failed delete.

## Why the reorder is safe

- `sendCommandToAgent` (`agentWs.ts:3123`) is a pure in-memory WS send — no DB write, no dependency on the device row existing.
- `self_uninstall` is not tracked by `recordOrphanedResultExpectation`, so the send registers no state that the cascade could have deleted.
- The permanent-delete path never calls `disconnectAgent` (only decommission does, at `:1606`), and nothing in `deleteDeviceCascade` touches `agentWs` — so the socket's connected/disconnected state is unaffected by the transaction.
- No other consumer reads `uninstallSent` / `agentUninstallSent`: the web caller (`deviceActions.ts:permanentDeleteDevice`) surfaces only `err.message`.

## Tests

The ordering is the *only* thing that distinguishes fixed from broken — response fields and call counts are identical either way, and both orderings send exactly one command on the happy path. So the new test records **when** the send happens relative to the commit:

```
expect(order).toEqual(['tx-commit', 'dispatch']);
```

**Verified red against the pre-fix route**, which produced `['dispatch', 'tx-commit']`. It wraps the real transaction rig rather than stubbing it, so the cascade still runs its actual statements.

The three rollback tests now assert `sendCommandToAgent` was **never called** — on the `CONNECTED_DEVICE` fixture, where the dispatch is one ordering mistake away from firing. On the offline fixture `uninstallSent` can only ever be `false` and those tests would pass against the very regression they exist to catch.

**Evidence:** `vitest run src/routes/devices/ src/routes/devices.test.ts src/services/deviceDeletion.test.ts` → 41 files, 566 passed. `tsc --noEmit` clean. `eslint` clean on both changed files.

Behaviour is otherwise unchanged: same success contract (`agentUninstallSent`, the unreachable-agent `warning`), same audit `uninstallCommandSent` field, same status codes.

### DIFF
--- a/apps/api/src/routes/devices/cascadeDelete.test.ts
+++ b/apps/api/src/routes/devices/cascadeDelete.test.ts
@@ -678,26 +678,6 @@ describe('DELETE /devices/:id/permanent — tickets are detached, not destroyed'
     expect(dissolveLinkGroupIfBelowMinimum).not.toHaveBeenCalled();
   });
 
-  /**
-   * A lock timeout must surface as the retryable 409 that reports
-   * `uninstallSent`, because the SELF_UNINSTALL is dispatched BEFORE the
-   * transaction and is irreversible — a bounded lock failure is the one path
-   * that can leave an agent uninstalling itself while its device row survives.
-   *
-   * The shape here is not invented. It is what a REAL lock timeout produces:
-   * verified against live Postgres with two connections contending on one row,
-   * the error arrives as `{ code: undefined, cause: { code: '55P03' } }`,
-   * because Drizzle wraps the postgres-js PostgresError. The route used to read
-   * the top-level `.code`, so this branch was dead and the operator got a bare
-   * 500 with no indication the uninstall had already gone out. Assert the
-   * WRAPPED shape specifically — an unwrapped `{ code: '55P03' }` fixture would
-   * pass against the broken code and prove nothing.
-   *
-   * Uses an ONLINE device with the uninstall actually dispatched. An earlier
-   * revision used the offline fixture, where `uninstallSent` can only ever be
-   * false, and merely asserted the property existed — code that hard-coded
-   * `false` would have passed it.
-   */
   // Still `decommissioned` — the route 400s anything else BEFORE it ever
   // reaches the uninstall dispatch, so an 'online' status here would have
   // tested the wrong branch entirely. What makes the uninstall fire is an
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1 (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer) on `ec91f7ea5`; round 2 (silent-failure-hunter, code-reviewer) on the delta `c2f06d0dd`.

**Findings:** 9 raised → 9 addressed across `c2f06d0dd` and `97ada74d3`; 0 outstanding.

Round 1 (must-fix):
- `isAgentConnected` was evaluated **outside** the try that the comment claimed protects an already-committed delete. It is not a bare Map read — it asserts the process role and throws in the worker role — so a throw would have 500'd a committed delete, losing the audit entry and the device-count invalidation while the row was permanently gone. Flagged independently by three of the four agents. Guard moved inside the try, matching the `getRedis()` block 30 lines down.
- Three mutations of the moved block **survived** the original suite (agent-offline-at-commit, the `warning` spread, and the `device.permanent_delete` audit entry — the last had no coverage anywhere in the repo). Now covered.
- Three comment claims were false as written, including one asserting a caller distinction that no caller makes, and one that invited a future maintainer to delete the three rollback tests as redundant.

Round 2 (must-fix):
- `captureException(err instanceof Error ? err : new Error(String(err)), c)` — it takes `unknown` deliberately and derives its Sentry tags by inspecting the value, so the wrap strips `pg_code` / CONNECT_TIMEOUT off non-Error throws, and `String()` on a hostile object can throw out of the very catch meant to contain it. Now passes `err` raw, matching both sibling call sites.

**On the test that was meant to pin that last one:** it was vacuous *twice*, and mutation testing is the only reason that is known. `toHaveBeenCalled()` stayed green with the wrap; so did message-matching (`String(err)` on an Error still contains the original message); so did object identity (with an `Error` input the ternary is a no-op). It is only observable on a **non-Error** throw, which is the fixture the final test uses.

**Mutation evidence** — every one of these now fails the suite, and each was verified individually rather than assumed: connectivity guard dropped · `agentUninstallSent` hard-coded / `warning` deleted · audit `uninstallCommandSent` hard-coded · guard moved back outside the try · `captureException` removed · its request-context arg dropped · the `Error(String(err))` wrap reintroduced.

**Tests:** `vitest run src/routes/devices/ src/routes/devices.test.ts src/services/deviceDeletion.test.ts` → 41 files, **570 passed**. `tsc --noEmit` clean; `eslint` clean on all three changed files.

**CI on `97ada74d3`:** Test API, Test Web, Type Check, Lint, Security Audit, Check Migrations (+ non-superuser) all green. Builds and integration shards still running at time of writing.

**Deliberately out of scope:** the web caller discards this route's success-path `warning`, so on the (dominant) path where the agent is already gone the operator sees a green toast while the agent is still installed. Pre-existing and not worsened by this PR — filed as #4368 rather than folded in, and the code comment says so rather than claiming a guarantee it does not deliver.

**Status:** review-clean, awaiting maintainer merge. Issue #3817 left open and assigned.

PR #4359, opened 2026-09-01
fix(security): Trivy-scan the api/web images we actually ship (#4273)

### LINKED ISSUE #4273: Shipped apps/api and apps/web images are never Trivy-scanned (CI scans docker/Dockerfile.* proxies that release never ships)
## Summary

The `Trivy Image Scan` job builds and scans `docker/Dockerfile.api` and `docker/Dockerfile.web`, but **those Dockerfiles are never shipped**. The images that actually ship to customers are built from `apps/api/Dockerfile` and `apps/web/Dockerfile`, and **neither is Trivy-scanned** — not in CI, and not at release time. So the two highest-traffic shipped images have no image-level vulnerability gate.

This is very likely the root cause of why `apps/api/Dockerfile` and `apps/web/Dockerfile` silently lacked the `libcrypto3/libssl3` upgrade that the other images carried (fixed in #4257): the scan that would have flagged them never looked at them.

## Evidence (VERIFIED)

**What CI scans** — `.github/workflows/security.yml:170,173`:
```
docker build -f docker/Dockerfile.api  -t breeze-api:security-scan .
docker build -f docker/Dockerfile.web  -t breeze-web:security-scan .
```
Neither `apps/api/Dockerfile` nor `apps/web/Dockerfile` appears anywhere in `security.yml`.

**What release ships** — `.github/workflows/release.yml`:
- `file: apps/api/Dockerfile` (`:2819`) and `file: apps/web/Dockerfile` (`:3171`) are the pushed images (also `hosted-images.yml`).
- `docker/Dockerfile.api` / `docker/Dockerfile.web` appear in **no** `release.yml` or `hosted-images.yml` `file:` input.

**Release-time Trivy** — only the three M365 executors get an `image-ref:` digest scan (`release.yml:2885, 2981, 3077`). The api/web release jobs (`:2819`, `:3171`) have **no** scan step.

Net: `docker/Dockerfile.api|web` are scan-only proxies that ship nothing; `apps/api|web/Dockerfile` ship but are scanned nowhere. `docker/Dockerfile.binaries` is the inverse — it *is* pushed (`release.yml:2757`) but is out of the guard's scope.

## Why it matters

The `dockerfileOpensslUpgrade.test.ts` docblock asserts the in-scope set is "exactly the set that both ships to users and freezes its package set." That is inverted for api/web, and the inverted mental model is plausibly what let the two shipped images sit unpatched. A green `Trivy Image Scan` currently says nothing about the images customers run.

## Proposed fix (options)

1. Point the CI Trivy job at the shipped Dockerfiles (`apps/api/Dockerfile`, `apps/web/Dockerfile`) instead of the `docker/*` proxies — or, if the proxies must stay for build-context reasons, add the shipped ones alongside.
2. Add an `image-ref:` digest Trivy scan to the api and web **release** jobs, matching the executor pattern (build → scan exact digest → promote), so a HIGH/CRITICAL blocks promotion.
3. Reconcile the two `docker/Dockerfile.api|web` proxies vs the shipped `apps/*/Dockerfile` — ideally one Dockerfile per image, scanned and shipped, so they cannot drift again.

## Blast radius

Supply-chain scan coverage for the primary shipped images. Priority: **high** — this is a standing blind spot, not a one-off; it already caused one missed CVE.

ToddHebebrand:
Closes #4273

Also addresses #4260 (same defect, filed independently from the #4246 fix — leaving it open for the reporter to confirm).

## The defect

`security.yml`'s `Trivy Image Scan` job built and scanned `docker/Dockerfile.api` and `docker/Dockerfile.web`. Neither file is ever published. The images customers actually run are built from `apps/api/Dockerfile` and `apps/web/Dockerfile` — and those were scanned **nowhere**: not in CI, not at release time, not in the hosted lane.

So a green `Trivy Image Scan` said nothing about the two most widely deployed images in the product. That is very likely how both came to ship a known HIGH CVE (CVE-2026-14456, libcrypto3/libssl3) that the *scanned* proxies had already been patched for (#4246 / #4257).

Verified coverage before this PR:

| Dockerfile | Published by | CI scan | Release-time scan |
|---|---|---|---|
| `apps/api/Dockerfile` | release.yml, hosted-images.yml | **no** | no |
| `apps/web/Dockerfile` | release.yml, hosted-images.yml | **no** | no |
| `apps/portal/Dockerfile` | release.yml, hosted-images.yml | yes | no |
| `apps/m365-*-executor/Dockerfile` (x3) | release.yml, hosted-images.yml | yes | yes (digest→scan→promote) |
| `docker/Dockerfile.binaries` | release.yml | no | no |
| `docker/Dockerfile.api` / `.web` | **nothing** | yes | n/a |

## The fix

The durable defect was the hand-maintained parallel list: a scan roster kept by hand beside a publish roster kept by hand, with nothing tying them together. Retargeting the two paths alone would just reset the clock, so:

**`.github/workflows/security.yml`** — `trivy-image-scan` becomes a matrix over `{image, dockerfile}` covering all six published Dockerfiles plus the two `docker/Dockerfile.*` compose variants (still built from source by `ci.yml`'s smoke test via `docker-compose.override.yml.ci`, and by self-hosters via `docker-compose.override.yml.local-build`, so they stay gated). Running as a matrix also parallelises what were six sequential image builds in one job.

**`apps/api/src/config/dockerfileImageScanCoverage.test.ts`** (new) — the anti-drift contract. It hand-lists nothing; it walks every workflow in `.github/workflows` and derives:

- **published** = the `file:` input of every `docker/build-push-action` step that actually pushes (`push: true`, or the `push=` field of `outputs:` for the push-by-digest lanes), with `${{ matrix.* }}` expanded against that job's own matrix;
- **scanned** = the `dockerfile:` values of the security workflow's scan matrix;

…and asserts published ⊆ scanned. Add a published image without scanning it, or retarget a publish step at a different Dockerfile, and this fails. Anything the derivation cannot resolve **throws** rather than resolving to nothing — a silently shrinking derivation would make every set comparison vacuously green, which is the same failure this file exists to prevent one level up.

`docker/Dockerfile.binaries` is the single recorded exemption, with its reason stored next to it: release builds it from `context: staging/`, a release-time artifact directory that does not exist in a repo checkout, so CI cannot reproduce the image (its base is the floating `alpine:3.24` tag and its payload is our own release binaries, already covered by the Trivy filesystem scan).

**`scripts/security/preflight.sh`** — built the same unshipped proxies, so a developer running the local mirror saw the same false green CI did. Now builds the shipped api/web Dockerfiles.

**`scripts/security/check-supply-chain-hardening.sh`** — its three executor scan assertions keyed off image tag strings (`breeze-m365-…:security-scan`) that the matrix no longer spells out literally; they now assert the matrix entry. Added the same assertion for the shipped api/web images so the regression is visible to the shell check too.

## Verification

**Live CI is the headline:** all 8 `Trivy Image Scan (…)` legs went green on the first push, including the two that never existed before — `Trivy Image Scan (api)` and `Trivy Image Scan (web)`. The shipped images are now scanned, and they are clean, so this does not arrive red on `main`.

- Real `docker build` of both `apps/api/Dockerfile` and `apps/web/Dockerfile` from a plain repo-root context succeeds locally — the new matrix entries are buildable, not just well-formed YAML.
- `dockerfileImageScanCoverage.test.ts` 16 passed; 555 passed across `apps/api/src/config/` (the two sibling Dockerfile guards included).
- **Mutation-checked, twice.** The first review ran a 12-mutation battery and found three mutations that broke the control while leaving the suite green. All three now fail (see the second commit): a build step carrying the matrix values in `env:` while hardcoding `-f apps/api/Dockerfile`; a job-level `if: github.event_name == 'schedule'`; `continue-on-error: true` on the Trivy step. Deleting the `apps/api/Dockerfile` matrix entry also still fails, naming the exact publishers.
- `tsc --noEmit` (apps/api) clean; `eslint` clean on the new file.
- `bash scripts/security/check-supply-chain-hardening.sh` passes.
- `actionlint 1.7.12 -shellcheck=` (the pinned CI version) clean over all workflows; `check-workflow-security.mjs` policy + its 113 self-tests pass.

## Two things worth a maintainer decision

**1. This makes the images visible, not merge-blocking.** `main`'s ruleset requires exactly one status check, `CI Success`, whose `needs:` list contains no job from `security.yml`. So a red `Trivy Image Scan` shows on the PR and on `main` but does not block a merge — that was true before this PR too. (Visibility is still what surfaced CVE-2026-14456.) If enforcement is wanted, the eight new per-image contexts would need adding to the ruleset, or folding into a `CI Success`-style aggregator. Related: the matrix renamed the check from `Trivy Image Scan` to `Trivy Image Scan (<image>)`, which is safe *only* because nothing requires the old context. Both facts are recorded in the new test's docblock so a future ruleset edit does not silently re-break it.

**2. Release-time digest scanning is still missing for api/web/portal/binaries.** `release.yml` and `hosted-images.yml` push those tags with no scan; only the three M365 executors do build → scan the exact manifest → promote. Extending that pattern means hand-rolling the multi-tag promotion (`{{version}}` / `{{major}}.{{minor}}` / `{{major}}` / `latest`) that `docker/metadata-action` does for those jobs today — `latest` gates every CLI install — so it belongs in its own change. The new test snapshots which publishing jobs have release-time scanning and which do not, so the gap now carries a marker instead of being invisible.

## CI cost

Eight concurrent runners per PR instead of one runner doing six sequential builds. Wall-clock should improve; total runner-minutes go up. The two new legs use plain `docker build` with no `cache-from: type=gha` (matching the existing legs); adding GHA cache is an easy follow-up if the minutes matter.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -157,77 +157,61 @@ jobs:
           format: 'table'
           exit-code: '1'
 
+  # Builds and scans every container image this repo publishes, using the SAME
+  # Dockerfile the release lanes ship.
+  #
+  # This list used to be hand-maintained beside the publish lists in release.yml
+  # and hosted-images.yml, and the two drifted: the job built the
+  # `docker/Dockerfile.api|web` compose variants while release/hosted shipped
+  # `apps/api|web/Dockerfile`, so the two most widely deployed images in the
+  # product were gated by nothing and shipped a known HIGH CVE
+  # (issues #4273 / #4260). `apps/api/src/config/dockerfileImageScanCoverage.test.ts`
+  # now derives the published set from those workflows and fails if anything
+  # published is missing from the matrix below.
   trivy-image-scan:
-    name: Trivy Image Scan
+    name: Trivy Image Scan (${{ matrix.image }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # --- Published to GHCR by release.yml and hosted-images.yml ---
+          - image: api
+            dockerfile: apps/api/Dockerfile
+          - image: web
+            dockerfile: apps/web/Dockerfile
+          - image: portal
+            dockerfile: apps/portal/Dockerfile
+          - image: m365-graph-read-executor
+            dockerfile: apps/m365-graph-read-executor/Dockerfile
+          - image: m365-graph-actions-executor
+            dockerfile: apps/m365-graph-actions-executor/Dockerfile
+          - image: m365-communications-executor
+            dockerfile: apps/m365-communications-executor/Dockerfile
+          # --- Never published, but built from source by CI and by users ---
+          # ci.yml's smoke test builds these via docker-compose.override.yml.ci,
+          # and self-hosters build them via docker-compose.override.yml.local-build,
+          # so they stay gated too.
+          - image: api-compose
+            dockerfile: docker/Dockerfile.api
+          - image: web-compose
+            dockerfile: docker/Dockerfile.web
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           persist-credentials: false
 
-      - name: Build API image
-        run: docker build -f docker/Dockerfile.api -t breeze-api:security-scan .
+      - name: Build image
+        env:
+          DOCKERFILE: ${{ matrix.dockerfile }}
+          IMAGE: ${{ matrix.image }}
+        run: docker build -f "$DOCKERFILE" -t "breeze-${IMAGE}:security-scan" .
 
-      - name: Build Web image
-        run: docker build -f docker/Dockerfile.web -t breeze-web:security-scan .
-
-      - name: Build Portal image
-        run: docker build -f apps/portal/Dockerfile -t breeze-portal:security-scan .
-
-      - name: Build M365 Graph Read Executor image
-        run: docker build -f apps/m365-graph-read-executor/Dockerfile -t breeze-m365-graph-read-executor:security-scan .
-
-      - name: Build M365 Graph Actions Executor image
-        run: docker build -f apps/m365-graph-actions-executor/Dockerfile -t breeze-m365-graph-actions-executor:security-scan .
-
-      - name: Build M365 Communications Executor image
-        run: docker build -f apps/m365-communications-executor/Dockerfile -t breeze-m365-communications-executor:security-scan .
-
-      - name: Scan API image
+      - name: Scan image
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
-          image-ref: 'breeze-api:security-scan'
-          severity: 'HIGH,CRITICAL'
-          format: 'table'
-          exit-code: '1'
-
-      - name: Scan Web image
-        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
-        with:
-          image-ref: 'breeze-web:security-scan'
-          severity: 'HIGH,CRITICAL'
-          format: 'table'
-          exit-code: '1'
-
-      - name: Scan Portal image
-        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
-        with:
-          image-ref: 'breeze-portal:security-scan'
-          severity: 'HIGH,CRITICAL'
-          format: 'table'
-          exit-code: '1'
-
-      - name: Scan M365 Graph Read Executor image
-        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
-        with:
-          image-ref: 'breeze-m365-graph-read-executor:security-scan'
-          severity: 'HIGH,CRITICAL'
-          format: 'table'
-          exit-code: '1'
-
-      - name: Scan M365 Graph Actions Executor image
-        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
-        with:
-          image-ref: 'breeze-m365-graph-actions-executor:security-scan'
-          severity: 'HIGH,CRITICAL'
-          format: 'table'
-          exit-code: '1'
-
-      - name: Scan M365 Communications Executor image
-        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
-        with:
-          image-ref: 'breeze-m365-communications-executor:security-scan'
+          image-ref: 'breeze-${{ matrix.image }}:security-scan'
           severity: 'HIGH,CRITICAL'
           format: 'table'
           exit-code: '1'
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `pr-test-analyzer`, two rounds, each driven by an actual mutation battery rather than a read-through. A `code-reviewer` agent was launched alongside it but never returned a report — so the findings below are entirely the test-analyst's, and nothing in this PR rests on a second independent reviewer. Flagging that rather than implying broader coverage than was obtained.

**Round 1 — 12 mutations, 3 survivors (all breaks of the control that left the suite green):**
- build step carrying the matrix values in `env:` while hardcoding `-f apps/api/Dockerfile` — eight legs building the API image, reporting green `Trivy Image Scan (web)`. Verbatim the failure the assertion's own comment claimed to catch.
- job-level `if: github.event_name == 'schedule'` — roster correct, job never runs.
- `continue-on-error: true` on the Trivy step — finding reported green.

Plus 8 lower-rated items: `stepPushes` failing open on `push: ${{ … }}` (a published image would be invisible), the hand-listed `PUBLISHING_WORKFLOWS` reinstating the very parallel-list defect this guard exists to abolish, a too-narrow Dockerfile walk, and a "mutation test" that asserted `Array.prototype.filter` filters without ever invoking the real predicate. All addressed in c96fa21f0.

**Round 2 — re-ran the battery against the fix. All six named survivors now fail.** One narrow survivor and three new observations remained; all addressed in c3a41c5c9:
- The build-step assertion was still evadable (`echo "building $DOCKERFILE"` satisfied the positive check; `-f ./apps/api/Dockerfile` dodged the negative one). Collapsed into one claim about one token: `-f` must *be* the matrix variable.
- **The one that mattered most:** the release-time snapshot wrote the label `scans the pushed digest` from a predicate that only checks "has a trivy-action step somewhere" — an unrelated `scan-type: fs` step would earn a job that label. The test went red on such a change, but the natural review action is to accept the new label, writing a false coverage claim into the file that the docblock then cites. Labels now say what is actually checked.
- `the compose variants are still built from source` derived only half its rationale — repointing every `.ci` reference in ci.yml left it green. Now asserted.
- Lazy matrix expansion was over-strict; whether a step pushes needs no matrix, so that decision now comes first.

**Findings:** 11 raised round 1 → all addressed in c96fa21f0; 4 raised round 2 → all addressed in c3a41c5c9; 0 outstanding. Two knowingly-accepted limits are recorded in the test docblock rather than silently left: the snapshot reads the git index so it is blind to *untracked* Dockerfiles, and the ruleset fact below lives in repo settings that no test here can read (source and date cited in-file).

**Tests:** `apps/api/src/config/` 555 passed (18 files), 16 in the new guard. `tsc --noEmit` clean, `eslint` clean, `check-supply-chain-hardening.sh` passes, `actionlint 1.7.12` (the pinned CI version) clean, `check-workflow-security.mjs` + its 113 self-tests pass. Real `docker build` of both `apps/api/Dockerfile` and `apps/web/Dockerfile` from a plain repo-root context succeeds locally.

**CI on the head SHA:** 24 passed, 0 failed, rest still running. `Trivy Image Scan (api)` and `Trivy Image Scan (web)` — the two legs that did not exist before this PR — are **green**, so the shipped images are now scanned and are clean. This does not arrive red on `main`.

**Status:** review-clean, awaiting maintainer merge. Two things left to a maintainer decision, both written up in the PR body: (1) this makes the shipped images *visible*, not merge-blocking — `main`'s ruleset requires only `CI Success`, which needs no `security.yml` job (read via `gh api repos/LanternOps/breeze/rulesets`, 2026-09-01); (2) release-*time* digest scanning is still absent for api/web/portal/binaries, deliberately deferred because it means hand-rolling the `latest`/major/minor tag promotion that gates every CLI install.

PR #4360, opened 2026-09-01
fix(api): route binarySync's release fetches through the SSRF-guarded redirect helper (#4262)

### LINKED ISSUE #4262: [API][Security] binarySync.ts still fetches the release manifest + signature unguarded — #3649 closed one path, not the artifact
Follow-up from #3649 / PR #4255, which routed `releaseArtifactManifest`'s outbound fetches through the SSRF-guarded `safeFetchFollowingRedirects` (`apps/api/src/services/urlSafety.ts`).

**The same two artifacts are still fetched unguarded by a second path**, so #3649's module is closed but the manifest is not. Verified on `origin/main` — `apps/api/src/services/binarySync.ts` contains no reference to `safeFetch` or `urlSafety`, and reaches the network with bare `fetch` at five sites:

```
:159   fetch(asset.browser_download_url, …)   ← inside fetchReleaseAssetBuffer
:224   fetch(checksumAsset.browser_download_url, …)
:1090  fetch(ghUrl, { headers: ghHeaders })
:1426  fetch(ghUrl, { headers: ghHeaders })
```

`fetchReleaseAssetBuffer` (:155) is what `fetchTrustedReleaseManifest` (:168) uses to pull **`release-artifact-manifest.json` and its `.ed25519` signature** at :201-202 — the exact pair #3649 was filed about. `fetchTrustedReleaseManifest` is then called at :1102 and :1437.

This runs at boot under `BINARY_SOURCE=github`, so it is on the startup path of any instance configured to sync binaries from GitHub.

### Why it matters

`browser_download_url` and `ghUrl` come from the GitHub API response, not from local config. A bare `fetch` follows redirects with no hop re-validation and no private-range check, which is the same shape #3649 identified — an attacker-influenced or compromised redirect can reach loopback, link-local (`169.254.169.254`) or RFC1918 addresses from inside the API container.

### Suggested fix

Route all five sites through the same canonical `safeFetchFollowingRedirects`, with the byte ceiling PR #4255 uses. Do **not** write a second guard — reuse the canonical helper.

Note the tripwire interaction PR #4255 surfaced: `safeFetch` calls `assertOutsideHeldDbContext`, so converting a call site that runs inside a held DB context will start warning (see #1105). Check whether the binarySync paths hold a context before converting, and decide deliberately rather than suppressing the warning.

Worth adding a source-scan test of the kind PR #4255 added, extended to cover `binarySync.ts`, so a future bare `fetch` in either module fails CI.

🤖 Filed by Claude Code from the #3649 fix; all line numbers and the absence of any `safeFetch`/`urlSafety` import verified against `origin/main`.

ToddHebebrand:
Closes #4262

## The hole

PR #4255 routed `releaseArtifactManifest.ts` through the SSRF-guarded `safeFetchFollowingRedirects`, closing #3649. But it flagged in its own body that `binarySync.ts` is a **structurally parallel, independently unguarded implementation that fetches the very same two files** — `release-artifact-manifest.json` and its `.ed25519` signature. So #3649's *module* was closed and the *artifact* was not.

Four bare `fetch` sites, verified against `origin/main`:

| Site | Function | What it fetches |
|---|---|---|
| `:159` | `fetchReleaseAssetBuffer` | the manifest **and its signature** (via `fetchTrustedReleaseManifest`) |
| `:224` | `parseChecksumsFallback` | `checksums.txt` — the integrity fallback |
| `:1090` | `syncFromGitHub` | GitHub Releases API JSON |
| `:1426` | `backfillBackupRowsForVersion` | GitHub Releases API JSON |

`browser_download_url` comes from the GitHub API response, not from local config. A bare `fetch` follows redirects with no hop re-validation and no private-range check, so an attacker-influenced or compromised redirect could reach loopback, `169.254.169.254` or RFC1918 from inside the API container. This runs **at boot** under `BINARY_SOURCE=github`.

## The fix

All four sites go through the **existing** `safeFetchFollowingRedirects` in `services/urlSafety.ts`. **No new helper or module** — that helper exists precisely so this is not reimplemented per caller. Each hop is a fresh `safeFetch`: independently DNS-resolved, filtered and IP-pinned before the socket is dialed; hop budget 5; relative `Location` resolved; exhausting the budget throws rather than silently returning a 3xx.

Each site also gains a `maxBytes` streaming ceiling. The bare `fetch` had **no ceiling at all** — it buffered whatever the remote sent. Ceilings are sized to be unreachable by any legitimate response (1 MiB manifest, 1 MiB checksums, 8 MiB release JSON) rather than tight: this is the boot path, and a ceiling a real release could trip would turn a security control into an availability bug.

**A correction to an earlier version of this description.** It claimed the guard closed a `GITHUB_TOKEN` leak, because bare `fetch` "replays headers across redirects". That is false, and review caught it. undici implements the fetch spec's step 13 and deletes `authorization`/`proxy-authorization`/`cookie`/`host` on a cross-origin redirect — I verified it in this worktree's `undici@8.10.0`, `lib/web/fetch/index.js:1348-1358`. So no token leak existed and none is fixed. The real win at those two sites is the per-hop DNS resolution and IP pinning; the header handling merely keeps parity. The code comments have been corrected too.

Production diff is 4 call sites, 1 import, 3 constants.

## The interesting failure this surfaced

Three suites intercepted the network with `vi.stubGlobal('fetch', …)`, which worked **only because `binarySync` used the global fetch**. `safeFetchFollowingRedirects` deliberately does not — it dials Node's `http`/`https` directly so each hop can be pinned. So the stubs stopped intercepting.

Worth stating plainly: those tests were not merely red, they were **unhooked**. I measured it — removing the bridge and running `binarySync.test.ts` fails **33 cases** with `Error: GitHub API error: 404`, i.e. they were reaching the live internet. Fixed by routing the guarded helper back through the global stub, the idiom `recoveryMediaService.test.ts` already uses:

```ts
vi.mock("./urlSafety", async (importOriginal) => ({
  ...(await importOriginal<typeof import("./urlSafety")>()),
  safeFetchFollowingRedirects: (url: string, init?: RequestInit) =>
    globalThis.fetch(url, init),
}));
```

Three files needed it (`binarySync.test.ts`, `releaseTrustChain.e2e.test.ts`, `agentVersionsLocalModeRoundtrip.test.ts`), plus `assertOutsideHeldDbContext: vi.fn()` on their `../db` mocks, since `safeFetch` calls it.

## Tests

New `binarySync.redirect.test.ts`, modelled on `releaseArtifactManifest.redirect.test.ts`. It deliberately does **not** mock `./urlSafety` — the guard runs for real; only DNS (`__setLookupForTests`) and the socket layer are stubbed, so every host that would be dialed is recorded and assertable.

1. **The legitimate chain still works** — a normal 302 `github.com` → `objects.githubusercontent.com` is followed and both artifacts are retrieved. This case is the point: an over-tight guard here would break every `BINARY_SOURCE=github` boot rather than secure it.
2. **GitHub API redirect to cloud metadata refused** (site `:1090`) — `SsrfBlockedError`, and the recorded requests prove `169.254.169.254` was never dialed.
3. **Manifest-asset redirect to cloud metadata refused** (site `:159`) — the exact artifact pair this issue is about.
4. **checksums.txt redirect to a host that DNS-resolves into RFC1918 refused** (site `:224`) — the deeper bypass, where the `Location` looks public and only resolution reveals `192.168.1.10`.
5. **Oversized manifest aborts at the streaming ceiling** — `ResponseTooLargeError` rather than a truncated buffer handed onward to signature verification.
6. **Anti-regrowth source scan** — no raw `fetch(`, no `globalThis/window/global.fetch(`, `safeFetchFollowingRedirects` must actually be *called* (exactly 4 sites), and **every** call site must pass `maxBytes`.

### Red-first controls (all mutations verified, not assumed)

- **Whole fix reverted, tests kept:** 6/6 failed, the security cases specifically on `expected Error: … to be an instance of SsrfBlockedError`, and the source scan on `expected '…' not to match /(?<![.\w])fetch\s*\(/`. Restored → green.
- **Only the `maxBytes` argument removed from the manifest site:** exactly 2 failed, 4 stayed green. The ceiling case failed with `expected ReleaseManifestSignatureError to be an instance of ResponseTooLargeError` — which is doubly useful, because it shows the *unbounded* failure mode concretely: without the ceiling the oversized body is buffered whole and fed to signature verification. The structural scan failed independently with `call site at offset 7422 must pass maxBytes`.
- **Test bridge removed from `binarySync.test.ts`:** 33 cases failed on live-network 404s, proving the bridge is load-bearing rather than decorative.

Honest caveat, same as #4255's: under the first control the reverted bare `fetch` uses undici and so escapes the `http/https.request` stub entirely, meaning it errors out rather than demonstrating a *completed* fetch of metadata. The control is a genuine discriminator; it is not a live exploit reproduction.

## Verification

- **21 affected suites, 418 tests green** (binarySync ×4, releaseTrustChain.e2e, releaseArtifactManifest ×2, releaseAssetTrust, urlSafety ×2, backupSsrfAdoption, recoveryMediaService ×2, installerBuilder, fleetFindings/reconcile, agentVersions, agentVersionsLocalModeRoundtrip).
- `tsc --noEmit` on `@breeze/api` — clean (needs `--max-old-space-size=12288` on a loaded host; known repo condition).
- `eslint` on all five changed files — clean.

## One thing a reviewer should weigh: the #1105 tripwire

`safeFetch` calls `assertOutsideHeldDbContext`. I traced all four sites before converting rather than after, and **all four run inside a held DB access context**:

- **Boot:** `index.ts:1669-1671` wraps `syncBinaries()` in `runWithSystemDbAccess` → `withSystemDbAccessContext` → `withDbAccessContext`, which opens a real `baseDb.transaction`. Every site inherits it.
- **Route:** `POST /api/v1/agent-versions/sync-github` takes the `withDbAccessContext(...)` branch in `middleware/auth.ts:746`; it is **not** in `SELF_MANAGED_DB_CONTEXT_ROUTES`.

**This PR surfaces an existing problem rather than creating one.** The bare `fetch` did the identical network round-trip inside the identical held transaction — the tripwire simply could not see it, because it only instruments `safeFetch`. The consequence of making it visible is a `console.warn` on every boot and every `/sync-github`, plus a handful of deduped Sentry `db_operation_inside_held_context` events per process.

It is **warn-only**: `assertOutsideHeldDbContext` throws only under `DB_CONTEXT_TRIPWIRE_STRICT`, which no CI job sets (verified — it appears only in `dbContextTripwire.integration.test.ts`, set inline for its own cases, and that suite does not touch binarySync).

**I deliberately did not suppress it**, per the issue's instruction to decide rather than silence. Two things I explicitly rejected:

- Wrapping the fetches in `runOutsideDbContext` — that clears the AsyncLocalStorage store but does **not** release the held transaction. It would silence the warning while leaving the connection pinned, which is worse than the warning.
- Restructuring the boot path — the correct fix is to drop the outer `runWithSystemDbAccess` and let `binarySync` open its own short contexts around its four `db.transaction` blocks (`:365`, `:624`, `:1584`, `:1662`), plus register the route in `SELF_MANAGED_DB_CONTEXT_ROUTES`. Both halves must land together. That is a boot-critical, high-blast-radius change — a contextless `db` write is a DENY, not a bypass, so getting it wrong means agent binaries silently fail to register at boot. It deserves its own PR and its own review, not a bolt-on to a security fix.

Happy to do that as a follow-up; flagging it so the call is conscious rather than implicit. This mirrors how #4255 handled the same tripwire interaction on the installer routes.

## Review round (three reviewers, all findings addressed)

`code-reviewer`, `silent-failure-hunter` and `pr-test-analyzer`. **1 blocking, 6 should-fix, 11 nits.** The two failure modes I most wanted checked came back clean and verified: no fail-open downgrade (a *blocked* manifest fetch can never be mistaken for an *absent* one — the `Promise.all` has no try/catch, so the refusal propagates and aborts the sync), and no vacuous assertions.

**BLOCKING — my own Sentry escalation was a no-op.** `captureException`'s tags pass through `ALLOWED_TAG_NAMES`, and neither `eventCode` nor `releaseSyncContext` was in it. `eventCode` is a `captureMessage` option besides, validated against a registry my codes were not in — and `scrubEvent` deletes `message` from every event. So the escalation I added to make fail-open refusals visible would have arrived **contentless and untriageable**: precisely the BREEZE-18 failure the allowlist's own doc comment warns about. Fixed by registering `release_sync_failure_reason` (closed 2-value set derived from the error *class*, never its message) and `release_sync_context` (hardcoded call-site literal), and asserting both through *both* gates in `sentry.test.ts`. Verified by mutation: removing them from the allowlist turns that test red, naming the two dropped tags.

**Two false claims in my comments**, both corrected against source rather than swapped for another guess — the credential-replay claim above, and a claim that `SsrfBlockedError` keeps the refused host out of `.message` (it doesn't; what it withholds is the resolved-IP *list*, which is the fact an operator actually needs).

**Other fixes:**
- **`timeoutMs` on all four sites.** `safeFetch` applies no timeout unless asked. Bytes without time still lets a hung socket pin a pooled Postgres connection idle-in-transaction — and per the tripwire note below, all four of these run inside a held context. The source scan now requires both ceilings at every call site.
- **`describeFetchFailure`** at the three fail-open catch sites, so a refusal is diagnosable rather than rendered as an indistinguishable `${err.message}`.
- **Route-level classification** of both new error types: 502 with a fixed operator-facing message, `resolvedIps` deliberately kept server-side. Three tests, including one pinning that the generic 422 path is unchanged.
- **`MAX_MANIFEST_BYTES` exported and imported** rather than copied — the "matches releaseArtifactManifest.ts" comment was a promise nothing enforced.
- **Test coverage for call site 4** (`backfillBackupRowsForVersion`), which was scan-only. It is also the site whose failure is *swallowed* by its caller, so a regression there is silent by construction — which inverts the usual priority.
- **The ceiling proven across a redirect hop.** It previously fired on hop 1; in production every asset fetch redirects, and *no test in this repo* pinned that `maxBytes` survives a hop.
- **Source scan broadened** to the `backupSsrfAdoption.test.ts` `FORBIDDEN` list plus uncapped bare `safeFetch(`, and the `maxBytes` check tightened — `maxBytes: undefined` used to satisfy it while disabling the ceiling. It now scans code with comments stripped, since it bans bare words (`undici`, `axios`) the comments legitimately discuss; the strip is deliberately conservative (block + whole-line comments only) so it can never delete real code from the scan.

Every fix above was mutation-verified, not assumed.

## Remaining sibling hole

`installerBuilder.ts` still has 4 raw fetches — all four with `redirect: 'follow'` (`:279`, `:317`, `:344`, `:383`), and it imports no guard at all — fetching the actual MSI/pkg bytes. (#4255's body said 3 of 4; I re-counted against `origin/main`, it is 4.) Out of scope here — #4262 scoped itself to `binarySync.ts` — but a reviewer should not read this PR as "release artifacts can no longer be fetched unguarded anywhere."

### DIFF
--- a/apps/api/src/routes/agentVersions.test.ts
+++ b/apps/api/src/routes/agentVersions.test.ts
@@ -39,6 +39,15 @@ vi.mock("../services/manifestSigning", () => ({
   signManifest: vi.fn().mockResolvedValue("test-signature"),
 }));
 
+// #4262: the sync-github route now classifies SSRF-guard refusals. Mock the
+// service so the route's error branches can be driven directly, and the Sentry
+// capture so it can be asserted without a DSN.
+vi.mock("../services/binarySync", () => ({
+  syncFromGitHub: vi.fn(),
+}));
+vi.mock("../services/sentry", () => ({
+  captureException: vi.fn(),
+}));
 vi.mock("../services/auditEvents", () => ({
   writeRouteAudit: vi.fn(),
 }));
@@ -76,6 +85,9 @@ import { db } from "../db";
 import { agentVersions } from "../db/schema";
 import * as manifestSigning from "../services/manifestSigning";
 import { writeRouteAudit } from "../services/auditEvents";
+import { syncFromGitHub } from "../services/binarySync";
+import { captureException } from "../services/sentry";
+import { ResponseTooLargeError, SsrfBlockedError } from "../services/urlSafety";
 import { requiredPlatformTrustFor } from "../services/releaseAssetTrust";
 
 // Recursively searches an and()/eq() spy tree (see drizzleSpies above) for an
@@ -244,6 +256,72 @@ describe("agentVersions routes", () => {
       }),
     };
   }
+
+  describe("POST /agent-versions/sync-github — guard-refusal classification (#4262)", () => {
+    it("answers 502 and does NOT leak the resolved internal address", async () => {
+      // The refusal message and .resolvedIps both name internal addresses.
+      // Pre-#4262 the raw err.message was echoed straight into the body; the
+      // route must now return a fixed operator-facing string instead. This is
+      // the assertion most at risk of being silently refactored away, because
+      // "just echo the error" reads like an improvement.
+      vi.mocked(syncFromGitHub).mockRejectedValue(
+        new SsrfBlockedError(
+          "all resolved IPs for api.github.com are private/loopback/link-local",
+          { hostname: "api.github.com", resolvedIps: ["10.1.2.3", "169.254.169.254"] },
+        ),
+      );
+
+      const res = await app.request("/agent-versions/sync-github", {
+        method: "POST",
+      });
+
+      expect(res.status).toBe(502);
+      const body = (await res.json()) as { error: string };
+      expect(body.error).toMatch(/private or link-local/i);
+      // The whole point: no internal address reaches the client.
+      expect(body.error).not.toMatch(/10\.1\.2\.3|169\.254\.169\.254/);
+      // …but it IS escalated server-side, since the catch otherwise loses it.
+      expect(vi.mocked(captureException)).toHaveBeenCalledTimes(1);
+      expect(vi.mocked(captureException).mock.calls[0]?.[2]).toMatchObject({
+        release_sync_failure_reason: "ssrf-blocked",
+      });
+    });
+
+    it("answers 502 on a body-ceiling abort", async () => {
+      vi.mocked(syncFromGitHub).mockRejectedValue(
+        new ResponseTooLargeError(1024 * 1024),
+      );
+
+      const res = await app.request("/agent-versions/sync-github", {
+        method: "POST",
+      });
+
+      expect(res.status).toBe(502);
+      const body = (await res.json()) as { error: string };
+      expect(body.error).toMatch(/1048576-byte limit/);
+      expect(vi.mocked(captureException).mock.calls[0]?.[2]).toMatchObject({
+        release_sync_failure_reason: "response-too-large",
+      });
+    });
+
+    it("leaves the pre-existing generic classification alone", async () => {
+      // Guards against over-reach: a plain failure must still take the original
+      // 422 path, not be swept into the new 502 branches.
+      vi.mocked(syncFromGitHub).mockRejectedValue(
+        new Error("something mundane broke"),
+      );
+
+      const res = await app.request("/agent-versions/sync-github", {
+        method: "POST",
+      });
+
+      expect(res.status).toBe(422);
+      expect((await res.json() as { error: string }).error).toBe(
+        "something mundane broke",
+      );
+      expect(vi.mocked(captureException)).not.toHaveBeenCalled();
+    });
+  });
 
   describe("GET /agent-versions (platform-admin list)", () => {
     it("non-platform-admin → 403", async () => {
--- a/apps/api/src/routes/agentVersions.ts
+++ b/apps/api/src/routes/agentVersions.ts
@@ -14,6 +14,8 @@ import {
 import { platformAdminMiddleware } from "../middleware/platformAdmin";
 import { writeRouteAudit } from "../services/auditEvents";
 import { syncFromGitHub } from "../services/binarySync";
+import { captureException } from "../services/sentry";
+import { ResponseTooLargeError, SsrfBlockedError } from "../services/urlSafety";
 import { getBinaryEdition } from "../services/binaryEdition";
 import { getGithubReleaseVersion } from "../services/binarySource";
 import { PERMISSIONS } from "../services/permissions";
@@ -985,8 +987,48 @@ agentVersionRoutes.post(
       // short registration instead of assuming a clean sync.
       return c.json(result);
     } catch (err) {
+      // A guard refusal (#4262) is not a routine sync failure: it means the
+      // release host resolved to a private/loopback/link-local address, which
+      // is a DNS-hijack signal. Left in the generic branch it became a 422 with
+      // no server-side trace at all — no log, no Sentry (writeRouteAudit only
+      // runs on success), so nobody could reconstruct it later. Log and
+      // escalate, and answer 502: the fault is upstream, not in the request.
+      // The response deliberately does NOT echo err.resolvedIps — internal
+      // addresses stay in the server-side log.
+      if (err instanceof SsrfBlockedError) {
+        console.error(
+          `[agentVersions] sync-github REFUSED by the SSRF guard (host=${err.hostname ?? "?"}, resolved=${err.resolvedIps?.join(", ") ?? "n/a"})`,
+        );
+        captureException(err, c, {
+          release_sync_failure_reason: "ssrf-blocked",
+          release_sync_context: "sync-github-route",
+        });
+        return c.json(
+          {
+            error:
+              "Release sync refused: the release host resolved to a private or link-local address. Check DNS and egress on the API host.",
+          },
+          502,
+        );
+      }
+      if (err instanceof ResponseTooLargeError) {
+        console.error(
+          `[agentVersions] sync-github aborted: response exceeded the ${err.maxBytes}-byte ceiling`,
+        );
+        captureException(err, c, {
+          release_sync_failure_reason: "response-too-large",
+          release_sync_context: "sync-github-route",
+        });
+        return c.json(
+          {
+            error: `Release sync aborted: the release host returned a body over the ${err.maxBytes}-byte limit.`,
+          },
+          502,
+        );
+      }
       const msg = err instanceof Error ? err.message : String(err);
       const status = msg.includes("GitHub API error") ? 502 : 422;
+      console.error(`[agentVersions] sync-github failed: ${msg}`);
       return c.json({ error: msg }, status);
     }
   },
--- a/apps/api/src/routes/agentVersionsLocalModeRoundtrip.test.ts
+++ b/apps/api/src/routes/agentVersionsLocalModeRoundtrip.test.ts
@@ -48,8 +48,13 @@ vi.mock("../db", () => ({
 // and this suite makes real network calls.
 vi.mock("../services/urlSafety", async (importOriginal) => ({
   ...(await importOriginal<typeof import("../services/urlSafety")>()),
-  safeFetchFollowingRedirects: (url: string, init?: RequestInit) =>
-    globalThis.fetch(url, init),
+  // Typed against SafeFetchInit (not RequestInit) so a call site's `maxBytes` /
+  // `timeoutMs` survive the bridge instead of being silently dropped, and a
+  // vi.fn() so a suite CAN assert on what binarySync passed the helper.
+  safeFetchFollowingRedirects: vi.fn(
+    (url: string, init?: import("../services/urlSafety").SafeFetchInit) =>
+      globalThis.fetch(url, init as RequestInit),
+  ),
 }));
 
 // agentVersions.ts imports these middlewares/services at module scope; stub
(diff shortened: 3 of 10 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` (3 agents, independent).

**Findings:** 18 raised (1 blocking, 6 should-fix, 11 nits) → all addressed in `8cb0fbcb8`; 0 outstanding.

The blocking one was mine and worth naming, because it failed in the direction that looks fine: the Sentry escalation I added in `859ebe41b` to make fail-open SSRF refusals visible was **itself silently discarded**. `captureException`'s tags pass through `ALLOWED_TAG_NAMES`, and neither key I used was in it — `eventCode` is a `captureMessage` option besides, validated against a registry my codes were not in, and `scrubEvent` deletes `message` from every event. The capture would have arrived contentless and untriageable: exactly the BREEZE-18 failure the allowlist's own doc comment describes. Fixed by registering `release_sync_failure_reason` (closed 2-value set, derived from the error *class*, never its message) and `release_sync_context` (hardcoded call-site literal), and asserting both through **both** gates in `sentry.test.ts`.

Review also caught **two false claims in my comments**, both corrected against source rather than replaced with another guess:
- I claimed bare `fetch` replays credentials across redirects, so the guard closed a `GITHUB_TOKEN` leak. It does not and it did not — undici implements the fetch spec's step 13 (verified in this worktree: `undici@8.10.0`, `lib/web/fetch/index.js:1348-1358`). The PR description carried that claim prominently and has been corrected.
- I claimed `SsrfBlockedError` keeps the refused host out of `.message`. It doesn't; what it withholds is the resolved-IP *list*.

Two things the reviewers confirmed clean, both verified by reading control flow rather than assumed: **no fail-open downgrade** (a *blocked* manifest fetch cannot be mistaken for an *absent* one — the `Promise.all` has no try/catch, so the refusal propagates and aborts the sync), and **no vacuous assertions**.

Substantive non-blocking fixes: `timeoutMs` on all four sites (bytes without time still lets a hung socket pin a pooled DB connection — and all four run inside a held context); `describeFetchFailure` so a refusal is diagnosable at the three fail-open catch sites; route-level 502 classification with `resolvedIps` kept server-side; `MAX_MANIFEST_BYTES` exported and reused instead of copied; behavioral coverage for call site 4 (the one whose failure is *swallowed*, so a regression there is silent by construction); the byte ceiling proven across a redirect hop (nothing in the repo pinned that); and the source scan broadened to the `backupSsrfAdoption` `FORBIDDEN` list plus uncapped bare `safeFetch(` — `maxBytes: undefined` previously satisfied it while disabling the ceiling.

**Tests:** 21 affected suites, **418 passed** (`vitest run --pool=threads` over binarySync ×4, releaseTrustChain.e2e, releaseArtifactManifest ×2, urlSafety ×2, sentry ×4, agentVersions ×2, backupSsrfAdoption, recoveryMediaService ×2, installerBuilder, releaseAssetTrust, fleetFindings/reconcile). `tsc --noEmit` on `@breeze/api` exit 0; `eslint` on all 10 changed files exit 0.

**Mutation controls** (every fix verified to discriminate, not assumed):

| Mutation | Result |
|---|---|
| Whole fix reverted | 6/6 new tests red |
| `maxBytes` removed from one site | exactly 2 red (behavioral + structural), 4 stayed green |
| `maxBytes: undefined` | now red — previously a **false pass** |
| One `timeoutMs` removed | scan red, naming the offset |
| Both tags un-allowlisted | `sentry.test.ts` red, naming the dropped tags |
| Route branches removed | 2 red, generic-422 case stayed green (no over-reach) |
| Bare `fetch` restored in code | scan still red after comment-stripping |
| urlSafety bridge removed | **33** cases red on live-network 404s |

That last row is the one I would not have predicted: those suites had been intercepting with `vi.stubGlobal('fetch')`, which worked only because binarySync used the global fetch. The guarded helper dials `http`/`https` directly, so without the bridge they were not merely red — they were **unhooked**, reaching the real internet.

**Status:** review-clean, awaiting maintainer merge. Two things for the merger, both in the PR body: the #1105 tripwire is now *visible* on the boot path (pre-existing, warn-only, but latently fatal if `DB_CONTEXT_TRIPWIRE_STRICT` is ever set — deliberately not papered over with `runOutsideDbContext`, which would silence it without releasing the connection), and `installerBuilder.ts` remains an unguarded sibling (4 raw fetches, all with `redirect: 'follow'`, no guard imported) — out of scope here, worth its own issue.

PR #4361, opened 2026-09-01
fix(api): stop System Tools routes reporting an agent timeout as success (#4025)

### LINKED ISSUE #4025: System Tools routes treat a command timeout as success — process kill reports "terminated successfully" when the agent never answered
### Description

Every System Tools route except the File Browser tests `result.status === 'failed'` and nothing else. A command that comes back `status: 'timeout'` therefore falls straight through into the success path.

Found while fixing #4024. The correct predicate already exists — `isCommandFailure()` in `fileBrowserHelpers.ts` handles both `'failed'` and `'timeout'` — the sibling routes just never adopted it.

### Actual Behavior

**Process kill reports a success that did not happen.** `apps/api/src/routes/systemTools/processes.ts:156`:

```ts
if (result.status === 'failed') {
  return c.json({ error: result.error || 'Failed to kill process' }, 500);
}
try {
  const data = JSON.parse(result.stdout || '{}');
  return c.json({ success: true, message: `Process ${pid} (...) terminated successfully` });
```

On a timeout, `result.stdout` is `undefined` → `JSON.parse('{}')` → the route answers **`success: true, "Process N terminated successfully"`**. The API has no idea whether the process died; the agent never answered. The audit row is written as `result: 'failure'` at line 152, so the audit trail and the API response actively contradict each other.

**Listings silently render empty.** `processes.ts:48` has the same shape: a timed-out `list_processes` parses to `{}` and returns `data: []` with HTTP 200. The technician sees "no processes running" on a device that simply did not answer in time.

The same pattern repeats in `services.ts`, `registry.ts`, `eventLogs.ts`, and `scheduledTasks.ts` — roughly 20 `if (result.status === 'failed')` sites.

### Expected Behavior

A timeout is not a success. These routes should use `isCommandFailure()` and return the classifier's `agent_timeout` (503) — and for state-changing commands, `unverified: true`, since on a timeout we genuinely cannot tell whether the device carried the operation out.

### Why it is filed separately from #4024

Different failure mode (a false success, not a blanked message), a different set of files, and each of the ~20 sites needs its own test. #4024's PR fixes the classifier those routes should call; this issue is adopting it.

### Component

API

ToddHebebrand:
Closes #4025

## The bug

Every System Tools route outside the File Browser tested `result.status === 'failed'` and nothing else. The command queue's timeout result is `{ status: 'timeout', error: 'Command timed out after ${timeoutMs}ms' }` with **no `stdout`** (`commandQueue.ts:753-762`), so it fell straight through into the success path.

**Process kill reported a success that did not happen.** `JSON.parse(undefined || '{}')` succeeds, so `processes.ts` answered:

```
200 { success: true, message: "Process 3456 (unknown) terminated successfully" }
```

…for a command the agent never acknowledged — while the audit row it had written moments earlier said `result: 'failure'`. The audit trail and the API response actively contradicted each other about whether a device had been mutated. Pinned by a test that now asserts both agree.

**Listings silently rendered empty.** The same shape made every list route return `200 { data: [] }`, so a technician read *"no processes running"* off a device that simply had not answered in time.

## The fix

All **25** call sites across `processes`, `services`, `registry`, `eventLogs` and `scheduledTasks` now classify through the same `isCommandFailure()` + `classifyCommandFailure()` that #4029 built for the File Browser and explicitly left these routes out of.

State-changing routes pass `mutating: true`, so a timeout answers **503 `agent_timeout` with `unverified: true`** and copy that tells the user to verify before retrying, rather than "please try again". Re-running a kill, a registry delete or a service stop against a half-completed state can compound the damage — the API genuinely cannot tell whether the device carried the operation out.

| | routes | timeout now answers |
|---|---:|---|
| read | 14 | 503 `agent_timeout`, no `unverified` |
| state-changing | 11 | 503 `agent_timeout` + `unverified: true` |

### One builder instead of 25 hand-rolled bodies

The `{ error, code, unverified? }` + status shape moves into `buildCommandFailureResponse()`. Hand-rolling a conditional spread at 25 sites is precisely how one of them ends up missing the `unverified` flag — the same class of copy-paste divergence that produced this bug. It returns `body` and `status` separately rather than one flat object because several routes already bind a local `status` (the service-list query filter), which a destructured `{ status, ...payload }` would collide with.

### Audit trail

Mutating routes switch `errorMessage: result.error` → `auditErrorMessage(result)`, so a timeout is tagged `[unverified]` in the audit trail. The trail and the response can no longer disagree about whether a device's final state was confirmed.

### The `not found` substring tests

The `error.toLowerCase().includes('not found') ? 404 : 500` checks are replaced by the classifier's not-found patterns. I verified this is equivalent rather than assuming it — every "not found" string these five tool families actually emit still answers 404:

| agent source | string |
|---|---|
| `processes.go:172,197` | `process not found: …` |
| `services_windows.go:89,123,144`, `services_linux.go:81`, `services_darwin.go:82` | `service not found: …` |
| `registry_windows.go:143` | `value not found: …` |
| `eventlogs_windows.go:108` | `event not found` |
| `tasks_windows.go:267,276` | `task not found: …` |

I also grepped these tool files for strings that could collide with the classifier's *other* prose branches (`is unknown`, `is offline`, `cannot execute command`, `timed out`, `did not complete`) — **none match**. Those branches only ever fire on queue-generated errors (`commandQueue.ts:832,1013`), which is where they are meant to fire, so widening the classifier's reach to five more route families introduces no misclassification.

## Verification

| suite | result |
|---|---|
| `systemTools.test.ts` + `systemTools/` (4 files) | **160 passed** |
| plus `bodyLimit` + `aiToolsFilesystem` ×2 (every suite touching systemTools) | **186 passed** |
| `tsc --noEmit -p apps/api` | clean |
| `eslint src` | clean |

**Red first.** Before the change the new tests failed 41/95, with `kill_process` returning **200** where the test demanded 503 — the reported bug reproduced exactly.

**Both guards revert-probed:**

- Reverting a single call site (`tasks:disable`) to `status === 'failed'` fails exactly its 2 table rows and nothing else — the table discriminates per site, so a future route added without a row can't hide.
- Dropping the `unverified` spread from the shared builder turns **13** tests red.

The table asserts its own length (25) so a new call site added without a row is a test failure, not a silent gap.

Tests also cover the paths that must *not* change: an unclassified agent failure still answers 500 with the agent's message, `not found` still answers 404, an offline device answers 503 `device_offline` without `unverified`, and a **prose-only** timeout (`status: 'failed'`, `error: 'command timed out after 30s'`) is still treated as unverified — the trap that was fixed in `buildBulkItemFailure` and would have been reintroduced by testing `status === 'timeout'` literally in the routes.

## Compatibility — checked, nothing breaks

Every web consumer (`RemoteToolsPage.tsx`, `ProcessDrilldownPanel.tsx`) branches on `res.ok` only, never on a specific status, so 503/404/422 are handled exactly as 500 was. The added `code` / `unverified` fields are inert to code that reads `json.error`. No mobile or e2e code calls these endpoints. No existing test asserts a status or body shape that changes.

`apps/docs/.../features/system-tools.mdx:616-635` already documents this classifier table generically for *"System Tools commands"* — before this PR that documentation was aspirational for these five route families. It is now accurate, and needs no edit.

## Found during review — a regression this PR introduced, now fixed

Routing these 25 routes through the shared classifier initially **regressed** the device-state message, and review caught it. The queue refuses every non-online device with one sentence — `Device is ${device.status}, cannot execute command` (`commandQueue.ts:1013`) — and `deviceStatusEnum` has seven values (`db/schema/devices.ts:7`). The classifier discarded that text and answered a flat `"The device is offline."`, so five of the six non-online states were reported wrongly. Before this PR these routes surfaced the real string.

It matters most for **`updating`**, which is set on every agent self-update (`agentWs.ts:2156`) — a technician restarting a service mid-upgrade was told the device was offline and sent to chase a network fault — and for **`quarantined`**, a security-containment state that must not read as a connectivity problem. The genuinely-offline case keeps its clean copy; the rest now name their real state. `code` stays `device_offline`, which is what callers branch on.

Two test guards were also found to be weaker than their comments claimed, and both were rebuilt:

- `expect(routes).toHaveLength(25)` promised to catch a new route added without a table row. It could not — the array is declared in the same file, so it only fails when a row is *deleted*. Proven by adding a 26th real call site and watching the suite stay green. It now reads the five route files and pairs each `await executeCommand(` with the gate that must follow it. Probed against the subtle version too: removing `registry:keys`' gate while adding a redundant one to `registry:values` keeps every count identical, and the rebuilt guard still reds.
- The `[unverified]` audit tag was asserted at **1 of 11** mutating routes; reverting the other ten reddened nothing. Now asserted at all eleven.

`isCommandFailure` is additionally an exhaustive switch, so adding a fourth `CommandResult` status is a compile error rather than a silent new fall-through — the precise shape of this bug. Re-probed: a fourth status fails with `TS2322 … not assignable to type 'never'`. Its default arm fails closed.

## Not in this PR

- **The web UI drops the `unverified` signal.** The mutating handlers do surface the warning text (they throw `json.error`, which is the "may have completed — refresh to verify before retrying" message), so the fix's user-visible benefit lands. But the four list-fetchers (`fetchProcesses`, `fetchServices`, `fetchEventLogs`, `fetchTasks`) catch and only `console.error`, so a timed-out listing now leaves the table unchanged instead of positively asserting "empty" — strictly better than the lie it used to tell, but still silent. Surfacing it properly needs an error-state design (toast vs inline empty-state), new locale keys across 8 catalogs, and E2E proof, so it wants its own issue rather than a rider here.
- **The BreezeAgent self-protection refusal stays a 500.** `"Cannot stop the Breeze agent service: operation blocked by self-protection"` is a deterministic refusal and arguably a 422, but it isn't in `REFUSAL_PATTERNS`, and the classifier's documented policy is that unrecognised refusals degrade to 500 (the safe direction). Changing it would re-pin an assertion #4029 just shipped, for a reason unrelated to timeouts.
- **The same bug class is live in `apps/api/src/routes/backup/`.** Verified, not inferred: `vss.ts:57`, `hyperv.ts` (5 sites) and `mssql.ts` (4 sites) all still test `result.status === 'failed'` only. `vss.ts` even does `result.stdout ? JSON.parse(result.stdout) : null`, so a timed-out VSS writer list answers **200 with `data: null`**. Ten sites, outside #4025's stated scope (System Tools) — worth its own issue.
- **Platform refusals still answer 500.** `registry is only supported on Windows` / `event logs are only supported on Windows` / `scheduled tasks are only supported on Windows` (`registry_other.go`, `eventlogs_other.go`, `tasks_other.go`) are deterministic refusals that degrade to `agent_execution_failed`. This is *unchanged* behaviour, not a regression — those paths answered 500 before too — but it is the single most common real failure for three of these tabs (any macOS or Linux device), and 422-vs-500 is a monitoring-visible policy call that deserves its own decision rather than a rider here.
- **`fileBrowserHelpers.ts` keeps its name** despite now serving all six route families. A rename churns a file that merged hours ago and would split it from its co-located test, which owns the Cloudflare property test. I added a module header saying to treat it as shared.

### DIFF
--- a/apps/api/src/routes/systemTools.test.ts
+++ b/apps/api/src/routes/systemTools.test.ts
@@ -1137,4 +1137,203 @@ describe('system tools routes', () => {
       expect(body.purged).toBe(3);
     });
   });
+
+  // Issue #4025. Every systemTools route outside the File Browser tested
+  // `result.status === 'failed'` and nothing else, so a `status: 'timeout'`
+  // result fell straight through into the success path:
+  //
+  //   - `kill_process` answered `{ success: true, "Process N terminated
+  //     successfully" }` when the agent never replied — while the audit row it
+  //     had just written said `result: 'failure'`. The trail and the response
+  //     actively contradicted each other about whether a device was mutated.
+  //   - Every listing parsed `undefined` stdout as `{}` and returned HTTP 200
+  //     with an empty array, so a technician read "no processes running" off a
+  //     device that simply did not answer in time.
+  //
+  // These routes now share the File Browser's classifier. The table below
+  // covers every one of the 25 sites so a new route cannot regress silently.
+  describe('agent timeout is never reported as success (#4025)', () => {
+    const taskPath = encodeURIComponent('\\Backup\\Daily Backup');
+    const registryQs = 'hive=HKEY_LOCAL_MACHINE&path=SOFTWARE';
+    const jsonPost = (body: unknown) => ({
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(body),
+    });
+
+    type RouteCase = {
+      label: string;
+      path: string;
+      init?: RequestInit;
+      /** Route changes device state → response must warn the result is unverified. */
+      mutating: boolean;
+    };
+
+    const routes: RouteCase[] = [
+      // processes.ts
+      { label: 'processes:list', path: '/processes', mutating: false },
+      { label: 'processes:get', path: '/processes/2048', mutating: false },
+      { label: 'processes:kill', path: '/processes/3456/kill?force=true', init: { method: 'POST' }, mutating: true },
+      // services.ts
+      { label: 'services:list', path: '/services', mutating: false },
+      { label: 'services:get', path: '/services/WinRM', mutating: false },
+      { label: 'services:start', path: '/services/WinRM/start', init: { method: 'POST' }, mutating: true },
+      { label: 'services:stop', path: '/services/WinRM/stop', init: { method: 'POST' }, mutating: true },
+      { label: 'services:restart', path: '/services/WinRM/restart', init: { method: 'POST' }, mutating: true },
+      // registry.ts
+      { label: 'registry:keys', path: `/registry/keys?${registryQs}`, mutating: false },
+      { label: 'registry:values', path: `/registry/values?${registryQs}`, mutating: false },
+      { label: 'registry:value:get', path: `/registry/value?${registryQs}&name=ProductName`, mutating: false },
+      {
+        label: 'registry:value:set',
+        path: '/registry/value',
+        init: {
+          method: 'PUT',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ hive: 'HKEY_LOCAL_MACHINE', path: 'SOFTWARE', name: 'TestValue', type: 'REG_SZ', data: 'Hello' }),
+        },
+        mutating: true,
+      },
+      { label: 'registry:value:delete', path: `/registry/value?${registryQs}&name=TestValue`, init: { method: 'DELETE' }, mutating: true },
+      { label: 'registry:key:create', path: '/registry/key', init: jsonPost({ hive: 'HKEY_LOCAL_MACHINE', path: 'SOFTWARE\\Breeze' }), mutating: true },
+      { label: 'registry:key:delete', path: `/registry/key?${registryQs}\\Breeze`, init: { method: 'DELETE' }, mutating: true },
+      // eventLogs.ts
+      { label: 'eventLogs:list', path: '/eventlogs', mutating: false },
+      { label: 'eventLogs:info', path: '/eventlogs/System', mutating: false },
+      { label: 'eventLogs:query', path: '/eventlogs/System/events', mutating: false },
+      { label: 'eventLogs:get', path: '/eventlogs/System/events/15234', mutating: false },
+      // scheduledTasks.ts
+      { label: 'tasks:list', path: '/tasks', mutating: false },
+      { label: 'tasks:get', path: `/tasks/${taskPath}`, mutating: false },
+      { label: 'tasks:history', path: `/tasks/${taskPath}/history?limit=10`, mutating: false },
+      { label: 'tasks:run', path: `/tasks/${taskPath}/run`, init: { method: 'POST' }, mutating: true },
+      { label: 'tasks:enable', path: `/tasks/${taskPath}/enable`, init: { method: 'POST' }, mutating: true },
+      { label: 'tasks:disable', path: `/tasks/${taskPath}/disable`, init: { method: 'POST' }, mutating: true },
+    ];
+
+    // The table must stay exhaustive: 25 call sites were converted, and a new
+    // route added without a row here would be untested.
+    it('covers every converted call site', () => {
+      expect(routes).toHaveLength(25);
+    });
+
+    it.each(routes)('$label answers 503 agent_timeout, never a success', async ({ path, init, mutating }) => {
+      mockDeviceSelect();
+      // `stdout` is absent on a timeout — exactly what used to parse to `{}`
+      // and become an empty listing or a "terminated successfully" message.
+      mockExecuteCommand.mockResolvedValue({ status: 'timeout', error: 'Command timed out' });
+
+      const res = await app.request(`/system-tools/devices/${deviceId}${path}`, init);
+
+      // 503, not 502/504: Cloudflare replaces an origin 502/504 body with its
+      // own branded page, which would blank this message on hosted.
+      expect(res.status).toBe(503);
+      const body = await res.json();
+      expect(body.code).toBe('agent_timeout');
+      expect(body.error).toContain("didn't respond in time");
+      // The regression being pinned: no success shape, and no empty listing.
+      expect(body.success).toBeUndefined();
+      expect(body.data).toBeUndefined();
+      expect(body.message).toBeUndefined();
+      // Only a state-changing route can leave the device in an unknown state.
+      expect(body.unverified).toBe(mutating ? true : undefined);
+    });
+
+    it.each(routes.filter((route) => route.mutating))(
+      '$label warns the operation may have completed before a retry',
+      async ({ path, init }) => {
+        mockDeviceSelect();
+        mockExecuteCommand.mockResolvedValue({ status: 'timeout', error: 'Command timed out' });
+
+        const res = await app.request(`/system-tools/devices/${deviceId}${path}`, init);
+        const body = await res.json();
+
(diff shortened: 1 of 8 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1: code-reviewer, silent-failure-hunter, pr-test-analyzer (all three on the full diff). Round 2: code-reviewer on the round-1 fix commit only, because that fix touched `classifyCommandFailure`, which `fileBrowser.ts` also calls.

**Findings:** 5 raised → all 5 addressed in `4c83ef098` + `bd1e177b4`; 0 outstanding.

| # | Round | Finding | Resolution |
|---|---|---|---|
| 1 | 1 | **Regression this PR introduced** — the classifier flattened all six non-online `device_status` values to "The device is offline.", where these routes previously showed the real string. Live impact: `updating` is set on every agent self-update (`agentWs.ts:2156`), `quarantined` is a containment state. | Parses the state out of the queue sentence and names it; genuinely-offline keeps its clean copy. `4c83ef098` |
| 2 | 1 | `expect(routes).toHaveLength(25)` did not guard what its comment claimed — reviewer **proved** it by adding a 26th real call site and watching the suite stay green. | Replaced with a source scan pairing each `await executeCommand(` with its gate. `4c83ef098` |
| 3 | 1 | The `[unverified]` audit tag was pinned at **1 of 11** mutating routes; reverting the other ten reddened nothing. | Asserted at all eleven. `4c83ef098` |
| 4 | 2 | The round-1 fix echoed unrecognised internal text back with a confident 503 for two guard shapes the queue sentence doesn't cover. | Those keep the generic message; unclassified text belongs on the 500 fallback. `bd1e177b4` |
| 5 | 2 | The new coverage guard compared per-file **totals**, so one route could gain a second gate while another lost its only one and the counts would still balance. | Now pairs by position. Probed: that exact balanced edit reds it, where the count version passed. `bd1e177b4` |

Two further review points were **verified and deliberately not acted on**, both recorded in the PR body: platform refusals (`registry is only supported on Windows`) still answer 500, which is *unchanged* behaviour rather than a regression and is a monitoring-visible 422-vs-500 policy call; and the web list-fetchers still swallow the error, which needs an error-state design plus locale keys across 8 catalogs. The dangerous half of #4025 — the false `"Process N terminated successfully"` — is fixed end to end, because the mutating handlers surface `json.error`.

**Also found, verified, out of scope:** the identical predicate is live at 10 sites in `apps/api/src/routes/backup/` (`vss.ts:57`, `hyperv.ts` ×5, `mssql.ts` ×4). `vss.ts` answers **200 with `data: null`** on a timeout. Not touched here — flagged for its own issue.

**Tests:** 197 passed across the 7 suites that touch systemTools (`systemTools.test.ts`, `systemTools/` ×3, `bodyLimit`, `aiToolsFilesystem` ×2). `tsc --noEmit` clean, `eslint src` clean.

**Evidence:** red-first — before the change the new tests failed 41/95 with `kill_process` returning **200** where 503 was required, reproducing the report exactly. Six revert-probes, each confirming the guard it targets: single call site reverted → only its 2 rows red; `unverified` spread dropped → 13 red; device-state fix reverted → 11 red; a 26th ungated dispatch → red; `auditErrorMessage` reverted at a previously-untested site → red; a 4th `CommandResult` status → `TS2322 … not assignable to type 'never'`. A dispatch probe also confirmed all 25 table rows reach distinct handlers and none is short-circuited by middleware.

**Status:** review-clean. CI still running on `bd1e177b4` at time of writing — **awaiting green checks, then maintainer merge.**

ToddHebebrand:
**CI update:** green on the current head `bd1e177b4` — **55 success, 1 skipped, 0 failing** of 56 checks. Verified by SHA (`check-runs` queried against the head oid, not `pr checks` on a possibly-stale run); local HEAD, PR head and the check-run SHA all match.

Supersedes the "CI still running" line in the review summary above. **Status: review-clean, CI-green, awaiting maintainer merge.** Not merged and issue #4025 not closed — both left to you.

PR #4363, opened 2026-09-01
fix(viewer): fall back gracefully when the WebView has no WebRTC (#3410)

### LINKED ISSUE #3410: Linux Breeze Viewer (webkit2gtk WebView) can lack RTCPeerConnection, breaking WebRTC
Reported by **rgadagno7123** in #agent-windows (Discord, message 1536413550325670009).

**Report (PII-redacted):**

I did inspect elements in the breeze viewer and saw this WebRTC connection failed:
ReferenceError: Can't find variable: RTCPeerConnection
ut — index-BX25cdYu.js:10:62218

## Triage analysis

The reported console line 'WebRTC connection failed: ReferenceError: Can't find variable: RTCPeerConnection' is not a browser message — it's the exact string logged in apps/viewer/src/lib/transports/webrtc.ts's catch block (console.warn('WebRTC connection failed:', err)) when createWebRTCSession's unguarded `new RTCPeerConnection(...)` (apps/viewer/src/lib/webrtc.ts:101) throws. The phrasing 'Can't find variable: X' is JavaScriptCore's ReferenceError format, not V8 (Chrome) or SpiderMonkey (Firefox), which both say 'X is not defined'. That points to the error actually occurring inside the native Breeze Viewer desktop app's embedded WebView, not the poster's browser tab. apps/viewer/src-tauri/Cargo.lock confirms the Linux build of that Tauri app links webkit2gtk (with javascriptcore-rs as its JS engine) — this is the WebView Tauri uses on Linux (an AppImage per the docs' Viewer Installers table). Depending on the distro's webkit2gtk/GStreamer build, RTCPeerConnection can be entirely absent as a global, so WebRTC can never start. The code does catch this and returns null so DesktopViewer.tsx's connect() (lines 913-919) falls back to the WebSocket/JPEG transport, but nothing detects/warns about missing WebRTC support up front, and there's no documentation of this Linux WebView limitation. The user's cross-platform pattern (Windows+Chrome/WebView2 works, Debian+Firefox/Chrome via the same native Viewer app fails) is consistent with this being a WebView-engine gap on Linux rather than a TURN/network issue.

## Affected code

- `apps/viewer/src/lib/webrtc.ts:101` — new RTCPeerConnection(...) called with no feature-detection guard
- `apps/viewer/src/lib/transports/webrtc.ts:186-201` — catch block logs 'WebRTC connection failed:' + err — matches the reported console line verbatim, and swallows non-Agent/SessionEnded errors, returning null
- `apps/viewer/src/components/DesktopViewer.tsx:913-919` — connect() falls back to connectWebSocket only after webrtcOk is falsy, i.e. after this exact ReferenceError is caught
- `apps/viewer/src-tauri/Cargo.lock:5302-5324` — webkit2gtk (+ javascriptcore-rs) is a dependency of the Linux Viewer build — the WebView/engine that produces 'Can't find variable' style ReferenceErrors

## Potential fix

Feature-detect RTCPeerConnection before calling createWebRTCSession (e.g. `if (typeof RTCPeerConnection === 'undefined') return null` early in webrtc.ts) so the viewer skips straight to the WebSocket fallback with a clear log/UI hint ('WebRTC unsupported on this system, using WebSocket fallback') instead of relying on catching a ReferenceError, and document the webkit2gtk/GStreamer WebRTC dependency for the Linux AppImage build.

ToddHebebrand:
Closes #3410

## The bug

The Linux Viewer is a Tauri app rendered by **webkit2gtk**, whose WebRTC support depends on how the distribution compiled it and whether the matching GStreamer plugins are installed. Where `RTCPeerConnection` is not a global, `apps/viewer/src/lib/webrtc.ts` constructed it unguarded:

```ts
const pc = new RTCPeerConnection({ iceServers });
```

That threw a bare `ReferenceError` from deep inside the signalling path. JavaScriptCore words it `Can't find variable: RTCPeerConnection` — the exact string the reporter pasted, and why the console line looked like a browser error rather than ours.

**Reproduced before fixing** (jsdom, where the identifier is genuinely absent, as on the affected webkit2gtk builds):

```
typeof RTCPeerConnection = undefined
THROWN: ReferenceError | RTCPeerConnection is not defined
```

The session was not dead — the transport caught the throw and returned `null`, so the viewer fell through to WebSocket/JPEG. But the fallback was accidental rather than designed: a full ICE-servers round trip was spent on every connect and reconnect for a peer connection that could never exist; the log said the generic `WebRTC connection failed:`, which reads as transient; nothing told the operator why quality had dropped; and reconnect could burn its whole 30s window retrying the impossible.

## The fix

- **`isWebRTCSupported()` + `WebRTCUnsupportedError`** (`lib/webrtc.ts`), with the probe at the very top of `createWebRTCSession` — before the ICE-servers request, so an unsupported WebView costs no network at all. It is a `typeof` check against the global object; reading the bare identifier would itself throw.
- **`createPeerConnection()` helper** wrapping construction, `addTransceiver` and both `createDataChannel` calls. A webkit2gtk build missing its GStreamer plugins commonly *exposes* `RTCPeerConnection` and then throws on use — the `typeof` probe alone would call that supported and let the failure resurface as a generic "connection failed". Any failure now maps to `WebRTCUnsupportedError` with the underlying cause preserved. This also closes a **pre-existing leak**: `close()` was not defined until after the channels were built, so a throw from `createDataChannel` stranded an open peer connection.
- **`transports/webrtc.ts`** distinguishes "this WebView has none" from "this attempt failed", logging an actionable cause while still returning `null` so the caller falls through.
- **`lib/transportAvailability.ts`** (new) — pure predicates gating every WebRTC affordance, so the dropdown option, the switch pill and the `switchTransport` backstop cannot drift apart. Same shape as the existing `autoHandoff.ts`.
- **A dismissible `role="status"` notice** explaining the compatibility transport and what it costs.
- **Docs** — a webkit2gtk/GStreamer section with per-distro install commands plus a troubleshooting entry on `features/remote-access`.

## Review rounds 2 and 3 — four real defects in my own earlier commits

The first commit stopped the crash but was not a complete fix. These were caught in review, not shipped:

1. **Silent dead controls.** The toolbar transport dropdown and the "user logged in — switch to WebRTC" pill stayed enabled. Clicking either hit the guard and silently returned; the pill even *self-dismissed*, signalling success. That traded a bad error message for **no feedback at all** — strictly worse than the crash on the feedback axis. Both are now gated, the macOS auto-handoff is gated at its call site (it polls every 2s), and the `switchTransport` check is an explicit backstop logging at `error` level, since reaching it means a UI gate has drifted.
2. **The probe missed the issue's own stated root cause** — see `createPeerConnection()` above. Worse, `webrtcSupported` would have been `true` in that case, actively *suppressing* the new notice.
3. **The notice was gated on `transport === 'websocket'`**, hiding it from every VNC session — including VNC deep links, which never attempt WebRTC at all — and making it vanish mid-session on the macOS websocket→VNC handoff. Now gated on the capability itself.
4. **The classification never reached the UI** (round 2). `webrtcSupported` came only from the up-front `typeof` probe and was never written again, so on a WebView that *exposes* `RTCPeerConnection` and throws on use — the case the issue actually describes — every gate stayed off and the notice never appeared. The net effect of `createPeerConnection` was a different console string. The transport now reports it via `onWebRTCUnsupported` and the component latches it, split across a ref (effects) and state (render) so the latch cannot re-run the connect effect and re-exchange the connect code.
5. **The STUN-only retry degraded silently** (round 3). It bound the original error and never logged it, so a malformed ICE/TURN config produced zero diagnostics while the session quietly ran without its TURN relay — the exact failure class the retry was added to avoid.

Also corrected: the docs table and notice **overstated what survives the fallback**. Ground truth is `capabilitiesFor()` in `lib/transports/types.ts` — websocket has `clipboardChannel: false`, so listing clipboard as available contradicted both the code and this page's own existing text. Clipboard sync, session switching and cursor streaming moved out of the WebSocket column, and quality/scale/FPS moved into it (only *bitrate* is WebRTC-only). The column is now titled "Not available on WebSocket" rather than "WebRTC only", because VNC keeps clipboard sync — and for the same reason the in-app notice lists only the intersection of what both fallbacks lose.

## Two adjacent problems this surfaced

**1. The Viewer's TypeScript tests ran in zero CI jobs.** Only the Rust half (`apps/viewer/src-tauri`) was covered. The 20 test files here — including this PR's regression tests — would never have guarded anything. Adds a `test-viewer` job (tests + `tsc --noEmit`) wired into `ci-success` on all four legs, so it fails closed. Confirmed the pre-existing suite was already green before wiring it. (Covers tests and types only: `@breeze/viewer` still has no `lint` script.)

**2. `DesktopViewer.tsx` was a "binary" file to git.** It held a raw NUL byte — a cache-key delimiter written as a literal instead of an escape — so git rendered *no diff* for this 2000-line component, in any PR. Replaced with the equivalent Unicode escape, verified byte-identical at runtime (`charCodeAt(4) === 0`, strings compare equal).

> ⚠️ Because the *base* blob still has the NUL, GitHub shows `DesktopViewer.tsx` as binary **in this PR only**. From the next PR on it diffs normally. To review it here:
> `git diff --text origin/main...fix/3410-viewer-webrtc-unsupported -- apps/viewer/src/components/DesktopViewer.tsx`

## Verification

| Check | Result |
|---|---|
| `vitest run` (whole viewer suite) | **20 files, 211 tests passed** (was 18/189 on main) |
| `tsc --noEmit` (apps/viewer) | clean |
| `pnpm test:workflow-security` | 113/113 pass |
| `pnpm --filter=@breeze/docs build` | 167 pages, clean; anchor verified in built HTML |
| `pnpm install --frozen-lockfile` | clean |

Tests were written **red first**, and the guards are **mutation-verified** rather than assumed:

| Mutation | Result |
|---|---|
| Remove the `createWebRTCSession` guard | 4 tests red |
| Remove the transport-layer `WebRTCUnsupportedError` branch | 1 test red |
| `isWebRTCSupported()` → `return true` | 5 tests red |
| Drop `webrtcSupported` from the pill predicate | 1 test red |
| Drop `webrtcSupported` from the dropdown predicate | 2 tests red |
| Drop the `onWebRTCUnsupported` call | 1 test red |
| Drop the ICE-rejection warning | 1 test red |

Both test files remove the global outright rather than stubbing it `undefined` — an assigned-but-undefined global yields a `TypeError`, and only a genuinely absent identifier reproduces the `ReferenceError` this issue is about.

One assertion was **deleted** during review: "does not fire the failure lifecycle callbacks" passed under every mutation (those callbacks only fire from a live connection-state handler), so it guarded nothing and inflated the apparent guard count.

## Known gaps

- **`apps/viewer` has no component-test infrastructure** (no `@testing-library/react`, no `.test.tsx`), so the `ViewerToolbar` wiring rests on `tsc` plus the tested predicates rather than a render test. Adding that infra is a follow-up, not something I wanted to smuggle into this PR as a lockfile change.
- **No telemetry.** `console.warn` is invisible in a shipped Tauri build (no devtools, no Sentry in the viewer), so nobody can answer "how many Linux viewers are silently on the JPEG transport?" — the question that would have surfaced this years earlier. Worth its own issue.
- No agent-side or API changes; this is entirely viewer-local capability detection. Not exercised on a real webkit2gtk host without WebRTC.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -547,6 +547,40 @@ jobs:
       - name: Run Portal tests
         run: pnpm test --filter=@breeze/portal
 
+  # The Viewer's TypeScript tests (transports, keymap, protocol, paste, …) ran
+  # in no CI job at all until issue #3410 added a regression test here and found
+  # the gap — only the Rust half (apps/viewer/src-tauri) was covered, by
+  # rust-check. Uses test:run because the package's `test` script is watch mode.
+  test-viewer:
+    name: Test Viewer
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
+        with:
+          version: ${{ env.PNPM_VERSION }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version-file: .node-version
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run Viewer tests
+        run: pnpm --filter=@breeze/viewer test:run
+
+      - name: Run Viewer type check
+        working-directory: apps/viewer
+        run: pnpm exec tsc --noEmit
+
   test-office-addin-core:
     name: Test Office Add-in Core
     runs-on: ubuntu-latest
@@ -2360,7 +2394,7 @@ jobs:
   ci-success:
     name: CI Success
     runs-on: ubuntu-latest
-    needs: [lint, typecheck, test-api, test-m365-graph-read-executor, test-m365-graph-actions-executor, test-m365-communications-executor, test-web, test-mobile, test-portal, test-office-addin-core, test-excel-addin, test-word-addin, test-powerpoint-addin, test-outlook-addin, security-audit, build-api, build-m365-graph-read-executor, build-m365-graph-actions-executor, build-m365-communications-executor, build-web, build-portal, build-agent, test-agent, test-agent-windows, windows-runtime-smoke, integration-test, check-migrations-nonsuperuser, rust-check, smoke-test, auth-browser-transition-browser-contract]
+    needs: [lint, typecheck, test-api, test-m365-graph-read-executor, test-m365-graph-actions-executor, test-m365-communications-executor, test-web, test-mobile, test-portal, test-viewer, test-office-addin-core, test-excel-addin, test-word-addin, test-powerpoint-addin, test-outlook-addin, security-audit, build-api, build-m365-graph-read-executor, build-m365-graph-actions-executor, build-m365-communications-executor, build-web, build-portal, build-agent, test-agent, test-agent-windows, windows-runtime-smoke, integration-test, check-migrations-nonsuperuser, rust-check, smoke-test, auth-browser-transition-browser-contract]
     if: ${{ !cancelled() }}
     steps:
       - name: Check all jobs passed
@@ -2374,6 +2408,7 @@ jobs:
           TEST_WEB_RESULT: ${{ needs.test-web.result }}
           TEST_MOBILE_RESULT: ${{ needs.test-mobile.result }}
           TEST_PORTAL_RESULT: ${{ needs.test-portal.result }}
+          TEST_VIEWER_RESULT: ${{ needs.test-viewer.result }}
           TEST_OFFICE_ADDIN_CORE_RESULT: ${{ needs.test-office-addin-core.result }}
           TEST_EXCEL_ADDIN_RESULT: ${{ needs.test-excel-addin.result }}
           TEST_WORD_ADDIN_RESULT: ${{ needs.test-word-addin.result }}
@@ -2406,6 +2441,7 @@ jobs:
              [[ "${TEST_WEB_RESULT}" != "success" ]] || \
              [[ "${TEST_MOBILE_RESULT}" != "success" ]] || \
              [[ "${TEST_PORTAL_RESULT}" != "success" ]] || \
+             [[ "${TEST_VIEWER_RESULT}" != "success" ]] || \
              [[ "${TEST_OFFICE_ADDIN_CORE_RESULT}" != "success" ]] || \
              [[ "${TEST_EXCEL_ADDIN_RESULT}" != "success" ]] || \
              [[ "${TEST_WORD_ADDIN_RESULT}" != "success" ]] || \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -551,6 +551,8 @@ jobs:
   # in no CI job at all until issue #3410 added a regression test here and found
   # the gap — only the Rust half (apps/viewer/src-tauri) was covered, by
   # rust-check. Uses test:run because the package's `test` script is watch mode.
+  # Covers tests + types only: @breeze/viewer still defines no `lint` script, so
+  # `turbo run lint` skips it and the repo-wide Lint job does not reach it.
   test-viewer:
     name: Test Viewer
     runs-on: ubuntu-latest
--- a/apps/docs/src/content/docs/features/remote-access.mdx
+++ b/apps/docs/src/content/docs/features/remote-access.mdx
@@ -522,6 +522,38 @@ export APPIMAGE_EXTRACT_AND_RUN=1
 
 Extract-and-run unpacks the whole bundle on every launch, so sessions start noticeably slower and each launch writes to `$TMPDIR`. Installing FUSE 2 is the better outcome where it is allowed.
 
+### WebRTC needs webkit2gtk built with the GStreamer plugins
+
+The Linux viewer renders in webkit2gtk, the system WebView, rather than bundling its own browser engine. Whether WebRTC is available therefore depends on how your distribution compiled webkit2gtk and which GStreamer plugins are installed — on a number of builds `RTCPeerConnection` is not exposed at all.
+
+When that is the case the viewer detects it before connecting, skips WebRTC, and runs the session on the WebSocket (JPEG) transport instead. A notice in the session window explains the switch, and the toolbar badge reads **WS** rather than **WebRTC**. The session still works; you lose the WebRTC-only features:
+
+| Available on WebSocket | WebRTC only |
+|---|---|
+| Screen view, keyboard, mouse, clipboard | Remote audio |
+| Reconnect, session end | Multi-monitor switching |
+| | Bitrate and quality controls |
+| | Secure attention sequence (Ctrl+Alt+Del) |
+
+To get WebRTC, install the GStreamer plugin sets webkit2gtk uses for it, then relaunch the viewer:
+
+```bash
+# Debian / Ubuntu
+sudo apt install gstreamer1.0-plugins-base gstreamer1.0-plugins-good \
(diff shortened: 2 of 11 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three rounds. Round 1: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` (parallel) on `f5bcddbc8`. Round 2: `code-reviewer` on the `9c9f92fc8` delta. Round 3: `code-reviewer` on the `f5086a64c` delta.

**Findings:** 3 rounds → 6 substantive findings raised, all 6 addressed; 0 outstanding. Three of them were real defects in my own earlier commits, not polish:

1. **Silent dead controls** (round 1, all three reviewers independently). Skipping WebRTC stopped the crash but left the toolbar dropdown and the "switch to WebRTC" pill enabled. Clicking either hit the guard and silently returned — the pill even self-dismissed, signalling success. That traded a bad error message for *no feedback at all*. Fixed in `9c9f92fc8` via pure predicates in `lib/transportAvailability.ts` so the dropdown, the pill and the `switchTransport` backstop cannot drift apart.
2. **The fix missed its own stated root cause** (round 2). `isWebRTCSupported()` is a `typeof` probe, but a webkit2gtk build without its GStreamer plugins *exposes* `RTCPeerConnection` and throws on use. In that case `webrtcSupported` stayed `true`, so every gate — including the notice — stayed off. Net effect of the classification was a different console string. Fixed in `f5086a64c`: the transport now reports the discovery via `onWebRTCUnsupported` and the component latches it.
3. **Silent config degradation** (round 3). The STUN-only retry I added bound the original error and never logged it, so a malformed ICE/TURN config produced zero diagnostics and the session silently ran without its TURN relay. Fixed in `00dd8d4f6`.

Also corrected along the way: the docs table and in-app notice **overstated what survives the fallback** (clipboard is not available on WebSocket, per `capabilitiesFor()`; quality/scale/FPS *are*), a `title` on a `disabled` button that could never render, and several comments that had drifted from the code.

**Verified clean in round 3:** the two-stage capability state machine does not re-run the connect effect on latch (every dep resolves to a stable root; `markWebRTCUnusable` is a `[]`-dep `useCallback`), the ref cannot go stale, and the STUN retry cannot leak a peer connection.

**Tests:** `apps/viewer` suite green — **20 files, 211 tests** (main has 18/189). `tsc --noEmit` clean. `pnpm test:workflow-security` 113/113. `@breeze/docs` build clean, 167 pages. `pnpm install --frozen-lockfile` clean.

Guards are **mutation-verified**, not assumed — each of these was run and the expected tests went red:

| Mutation | Tests red |
|---|---|
| Remove the `createWebRTCSession` guard | 4 |
| Remove the transport `WebRTCUnsupportedError` branch | 1 |
| `isWebRTCSupported()` → `return true` | 5 |
| Drop `webrtcSupported` from the pill predicate | 1 |
| Drop `webrtcSupported` from the dropdown predicate | 2 |
| Drop the `onWebRTCUnsupported` call | 1 |
| Drop the ICE-rejection warning | 1 |

One assertion was **deleted** rather than kept: "does not fire the failure lifecycle callbacks" passed under every mutation, so it guarded nothing.

**CI:** the new `test-viewer` job **passed** on an earlier commit's run, confirming the job works end-to-end. At the time of writing, checks for the final commit `00dd8d4f6` are still queued behind a repo-wide runner backlog — **please confirm CI is green on the head SHA before merging.**

**Status:** review-clean, awaiting maintainer merge. Two things worth a decision:

- `apps/viewer` has **no component-test infrastructure** (no `@testing-library/react`, no `.test.tsx`), so the `ViewerToolbar` wiring rests on `tsc` plus the tested predicates rather than a render test. I did not add the dependency, to keep a lockfile change out of a bugfix PR — worth a follow-up.
- **No telemetry.** `console.warn` is invisible in a shipped Tauri build (no devtools, no Sentry in the viewer), so nobody can answer "how many Linux viewers are silently on the JPEG transport?" — the question that would have surfaced this years earlier. Also worth its own issue.

Not verified on a real webkit2gtk host lacking WebRTC; behaviour is covered by unit tests rather than a live Linux repro.

Issue #3742, opened 2026-08-20
Self-hosted binary sync auto-promotes GitHub /releases/latest past the server's own version


ToddHebebrand:
Found while scoping the internal-only 0.106.0 server-image rollout.

## Problem 1: "manual agent updates" is a hosted-prod-only property, not a product default

On API boot, `syncBinaries()` (github mode) fetches GitHub **`/releases/latest`** with no version argument (`apps/api/src/services/binarySync.ts` — `syncFromGitHub()`), and `upsertVersion` promotes what it finds because `AGENT_AUTO_PROMOTE` defaults to **true** when unset (`apps/api/src/services/binarySource.ts`, `getAgentAutoPromote()`). `agent_versions.isLatest` is exactly what the heartbeat hands out as `upgradeTo` (`apps/api/src/routes/agents/helpers.ts`, `resolvePinnedUpgradeTarget`).

So for a default self-hosted install, publishing a GitHub Release rolls their entire agent fleet on their next API restart. The manual-promote gate (`POST /agent-versions/promote`) only exists for operators who opted into `AGENT_AUTO_PROMOTE=false` — which is our droplets, not self-hosters.

## Problem 2: isLatest can run AHEAD of the bytes the API can serve → failed-update loop

The download redirect is pinned to the deployed image: `getGithubReleaseVersion() = BINARY_VERSION || BREEZE_VERSION || 'latest'` (`binarySource.ts`), used by every `/api/v1/agents/download/*` 302. A self-hoster on 0.105.x images who merely **restarts** their API after a 0.106.0 release publishes gets:

1. boot sync promotes `isLatest = 0.106.0` (demoting 0.105.x),
2. heartbeat emits `upgradeTo=0.106.0`,
3. agent fetches 0.106.0's manifest/checksums, but the redirect serves the **0.105.x** asset,
4. checksum verification fails → repeated failed update attempts (fails safe, but loops forever and spams logs).

`apps/api/src/routes/agentVersions.ts` has an explicit guard comment for the *inverse* case (isLatest behind the server version, backup component); this direction has no guard.

## Proposed fix

Cap boot-sync auto-promotion at the server's own version: when `AGENT_AUTO_PROMOTE` is unset/true, sync `releases/tags/v${BREEZE_VERSION}` (or clamp `isLatest` to `compareAgentVersions(version, APP_VERSION) <= 0`) instead of unconditionally promoting `/releases/latest`. That fixes both problems in one change:
[...]

PR #4364, opened 2026-09-01
fix(api): serve component binaries from the promoted agent_versions row (#3499)

### LINKED ISSUE #3499: Install script checksum mismatch: /download and /agent-versions/latest resolve version independently
Reported by **rgadagno7123** in #agent-linux (Discord, message 1537460735683072031).

**Report (PII-redacted):**

Updated to 105.1 and when adding a linux device i'm getting a checksum error [ERROR] Checksum verification failed for downloaded agent binary. Expected 5ef6108d8993e62bfaa7f72906bbd60008394b98df173be346da9aedf2072d67, got 00eea294b90f47e5fbc6031d34928f24f46d63e83cd4979162a3224bf003da73.  Is this over the signing of agents?

## Triage analysis

The install.sh flow for Linux (agent/api/src/routes/agents/download.ts generateInstallScript) fetches the expected SHA-256 from GET /agent-versions/latest, then downloads the binary from GET /agents/download/:os/:arch, and compares them. These two endpoints resolve 'which release is current' through two independent mechanisms that are not guaranteed to agree: (1) /agent-versions/latest reads the checksum from the agent_versions DB row where isLatest=true — written by binarySync.syncBinaries()/ensureCurrentVersionRegistered(), keyed off APP_VERSION/BREEZE_VERSION and the AGENT_AUTO_PROMOTE/promote flow; (2) /agents/download/:os/:arch (BINARY_SOURCE=github, the default) 302-redirects to getGithubAgentUrl(), which resolves the release tag from process.env.BINARY_VERSION || BREEZE_VERSION || 'latest' at request time in whichever API process/pod answers that specific request. If a request for the metadata endpoint and a request for the binary endpoint are served by processes/pods with different resolved versions (e.g. mid-rolling-upgrade to 105.1 where old and new API instances briefly coexist behind a load balancer, or any deployment where BINARY_VERSION/BREEZE_VERSION drifts from what's actually registered as isLatest in the DB), the client gets a checksum for one release and bytes for another — exactly the symptom reported: 'Checksum verification failed for downloaded agent binary' right after upgrading to 105.1 while adding a new Linux device. This is not a signing problem (verify_sha256 in the generated script is a plain SHA-256 compare, unrelated to code-signing) — the checksum check did its job and correctly refused a binary that doesn't match the metadata it was told to expect. The underlying defect is that the binary-serving route and the checksum-metadata route can disagree about which release is 'latest'.

## Affected code

- `apps/api/src/routes/agents/download.ts:14-44` — GET /download/:os/:arch resolves the GitHub redirect target via getGithubAgentUrl() at request time, independent of the DB's isLatest row
- `apps/api/src/routes/agents/download.ts:812-842` — Generated install.sh verify_sha256(): plain SHA-256 compare against the checksum fetched from /agent-versions/latest, unrelated to binary signing
- `apps/api/src/routes/agents/download.ts:924-963` — Linux install path: fetches checksum from VERSION_METADATA_URL, then downloads from DOWNLOAD_URL, and fails hard on any mismatch
- `apps/api/src/services/binarySource.ts:33-51` — getGithubReleaseVersion()/githubDownloadBase()/getGithubAgentUrl(): resolves release tag from BINARY_VERSION||BREEZE_VERSION||'latest' per-process env, used by the /download redirect
- `apps/api/src/routes/agentVersions.ts:394-455` — GET /agent-versions/latest returns the checksum from the agent_versions DB row where isLatest=true, a completely separate source of truth from the download route's env resolution
- `apps/api/src/services/binarySync.ts:383-396` — syncBinaries(): unconditional syncFromGitHub() (hits GitHub's literal /releases/latest) followed by ensureCurrentVersionRegistered() (fetches the specific APP_VERSION tag and re-promotes it), showing isLatest is written by a background process disconnected from the download route's request-time env lookup
- `apps/api/src/services/binarySync.ts:774-817` — ensureCurrentVersionRegistered(): writes/promotes the agent_versions row for APP_VERSION/BREEZE_VERSION, the DB-side half of the split source of truth

## Potential fix

Make the binary-serving route derive its redirect target from the same agent_versions row that /agent-versions/latest just served (e.g. have the install script pass the specific version it got back from /agent-versions/latest to a versioned download path, or have /agents/download/:os/:arch look up the current isLatest DB row instead of re-resolving BINARY_VERSION/BREEZE_VERSION independently), so the checksum and the bytes always come from one query.

ToddHebebrand:
Closes #3499

## The defect

`install.sh` fetches the expected SHA-256 from `GET /agent-versions/latest` (the `agent_versions` row with `isLatest=true`) and then downloads the bytes from `GET /agents/download/:os/:arch`. Those two endpoints resolved "which release is current" through **two independent mechanisms**:

- **checksum** — the promoted `agent_versions` row, written by `binarySync`.
- **bytes** — `getGithubReleaseVersion()` (`services/binarySource.ts:33`), reading `BINARY_VERSION || BREEZE_VERSION` from **per-process env at request time**.

So the download path was structurally incapable of agreeing with the checksum path except by coincidence. When the GitHub sync stalled, `agent_versions` stayed frozen at 0.104.0 while the route happily served 0.105.1 — one full release apart — and the install aborted with `Checksum verification failed for downloaded agent binary`.

To be explicit about what was *not* wrong: the checksum check behaved correctly. It refused a binary that genuinely did not match the metadata it was handed, and this has nothing to do with code signing. The defect is that the two halves could disagree at all.

## The fix

Both halves now go through one query. `services/promotedAgentVersion.ts` resolves the promoted (`isLatest`) row — same platform/arch/component/**edition** scoping *and the same `created_at DESC` tiebreak* as `/agent-versions/latest` — and the download routes pin the GitHub release tag to it. A stale sync now yields a *consistent old version* instead of an inconsistent pair.

Applied to all five components pulled with a checksum comparison, not just the agent: `install.sh` checksum-verifies **backup** the same way it does the agent, and watchdog / user-helper / helper are fetched by the agent's verified updater against the same metadata.

**Two distinct failure modes, deliberately handled differently:**

| condition | behavior | why |
|---|---|---|
| No promoted row at all | Fall back to the env-resolved URL, log every time, capture to Sentry once per tuple | The cold-start state of a deployment that has never synced. Must keep working exactly as before. |
| Lookup **faults** | Throw → route returns **503 + Retry-After** | Falling back here *is* the pre-#3499 bug: it hands the client bytes that fail the checksum it already holds, reporting a server-side DB fault to an end user as "Checksum verification failed" — the exact string this fix eliminates. |

## Coupled change worth reviewing closely

`/agent-versions/:version/download` gates its `component=backup` URL rewrite on "is this row what the versionless route will serve". It used to hardcode *the env version*, correct only while the route resolved env.

Rather than restate the route's new rule here (a plain `isLatest` test would have been wrong too — for a never-synced deployment nothing is promoted and the route legitimately falls back to env, which would have silently removed backup self-heal for exactly the deployments the fallback protects), **the guard now asks the route's own resolver what will be served and compares.** The two cannot drift.

## Verification

- **Mutation-tested**, because a reviewer found the original tests could not detect a revert. Each of these previously survived and now fails:
  - builders accepting the promoted version and **ignoring** it — a complete revert of the fix — passed all ~30k API tests before; now fails 5.
  - `and(...)` → `or(...)` and `eq(isLatest,true)` → `ne(...)` in the resolver — both passed before; now fail.
  - the `darwin`→`macos` platform mapping.
- The regression test was red before the fix with exactly the reported symptom (`Expected v0.104.0, Received v0.105.1`).
- Affected suites green single-run: 12 files / ~540 tests across `promotedAgentVersion`, `binarySource`, `binarySync`, `agents/download`, `agentVersions`, `agentVersionsLocalModeRoundtrip`, `supportPublic`, `viewers/download`, `helpers.agentUpdatePolicy`, `heartbeat`, `sentry*`.
- `tsc --noEmit` clean; `eslint` clean on all touched files.
- CI's `BINARY_SOURCE=github download path` smoke job (real API + real DB) passed on the first commit — end-to-end evidence the DB lookup works outside the mocks.

## Notes for the reviewer

- **`/agent-versions/latest` had no `ORDER BY`.** Nothing enforces one promoted row per `(component, platform, arch, edition)` — the invariant is demote-then-insert, not a unique constraint — so adding the tiebreak to only the new resolver would have made the two halves select *different* rows on a duplicate. Both now share it, and a test pins it. **A partial unique index on `isLatest` is the real fix and is worth a follow-up.**
- **`#3742`'s operator guidance was made false by this change** and is updated: it told operators to set `BINARY_VERSION` "so boot sync and the download redirect agree", but the redirect follows the promoted row now.
- **macOS `.pkg` is deliberately not pinned.** Its install path verifies xar magic bytes and Gatekeeper notarization, never a SHA-256 against `/agent-versions/latest`, so there is no checksum/bytes pair to keep consistent. Pinned as a test so it reads as intentional. Consequence: a macOS install lands on the env release while Linux lands on the promoted one, reconciling on first heartbeat.
- **Not fixed here:** for non-backup components the rewrite stays unconditional, so an agent pinned to a *non-promoted* version still gets the promoted bytes and fails safe without healing. Pre-existing and unchanged (it previously got the env version, equally mismatched); gating every component looked like a deliberate design decision to leave alone.
- **`BINARY_SOURCE=local` is still unpinned** — this fix reconciles the github branch only.
- Adds one indexed read to a public route that already 302-redirects. A short TTL cache would make it free if that ever matters.

No schema or migration changes — `agent_versions` is a global, non-RLS catalog table.

### DIFF
--- a/apps/api/src/routes/agentVersions.test.ts
+++ b/apps/api/src/routes/agentVersions.test.ts
@@ -229,6 +229,11 @@ describe("agentVersions routes", () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
+    // clearAllMocks clears calls but KEEPS implementations, so a
+    // mockResolvedValue set by one backup-guard test would leak into every
+    // later one and silently stop it exercising the branch its name claims.
+    vi.mocked(getPromotedComponentVersion).mockReset();
+    vi.mocked(getPromotedComponentVersion).mockResolvedValue(null);
     platformAdminState.allow = true;
     delete process.env.AGENT_UPDATE_MANIFEST_PUBLIC_KEYS;
     delete process.env.BREEZE_UPDATE_MANIFEST_PUBLIC_KEYS;
@@ -1558,6 +1563,72 @@ describe("agentVersions routes", () => {
         expect(body.url).toBe(canonical);
         expect(body.checksum).toBe(checksum);
       } finally {
+        delete process.env.PUBLIC_API_URL;
+        delete process.env.BINARY_VERSION;
+      }
+    });
+
+    // The guard must mirror the download route's BINARY_SOURCE branch, not
+    // just its github half. In local mode that route streams ONE unversioned
+    // file from disk/S3 whose version is the binaries-volume build — the env
+    // version — and never consults the promoted row. Deriving the promoted row
+    // here would withhold the rewrite (and cost a pointless query) whenever the
+    // promoted row differs from the disk build, e.g. AGENT_AUTO_PROMOTE=false
+    // or after a rollback via POST /agent-versions/promote, silently ending
+    // backup self-heal for those deployments.
+    it("compares against the env version in local mode, without consulting the promoted row", async () => {
+      const canonical =
+        "https://github.com/LanternOps/breeze/releases/download/v1.0.0/breeze-backup-linux-amd64";
+      const checksum = "c".repeat(64);
+      const signed = makeSignedReleaseManifest({
+        component: "backup",
+        platform: "linux",
+        arch: "amd64",
+        url: canonical,
+        checksum,
+        size: 2048,
+      });
+
+      vi.mocked(db.select).mockReturnValue({
+        from: vi.fn().mockReturnValue({
+          where: vi.fn().mockReturnValue({
+            limit: vi.fn().mockResolvedValue([
+              {
+                version: "1.0.0",
+                platform: "linux",
+                architecture: "amd64",
+                component: "backup",
+                downloadUrl: canonical,
+                checksum,
+                fileSize: BigInt(2048),
+                releaseManifest: signed.manifest,
+                manifestSignature: signed.signature,
+                signingKeyId: "test-key",
+                // Not promoted — 1.1.0 is. In github mode that would withhold
+                // the rewrite; in local mode the disk build is what matters.
+                isLatest: false,
+              },
+            ]),
+          }),
+        }),
+      } as any);
+      vi.mocked(getPromotedComponentVersion).mockResolvedValue("1.1.0");
+
+      process.env.BINARY_SOURCE = "local";
+      process.env.PUBLIC_API_URL = "https://us.example.com";
+      process.env.BINARY_VERSION = "1.0.0";
+      try {
+        const res = await app.request(
+          "/agent-versions/1.0.0/download?platform=linux&arch=amd64&component=backup",
+        );
+        expect(res.status).toBe(200);
+        const body = await res.json();
+        expect(body.url).toBe(
+          "https://us.example.com/api/v1/agents/download/backup/linux/amd64",
+        );
+        expect(getPromotedComponentVersion).not.toHaveBeenCalled();
+      } finally {
+        delete process.env.BINARY_SOURCE;
         delete process.env.PUBLIC_API_URL;
         delete process.env.BINARY_VERSION;
       }
--- a/apps/api/src/routes/agentVersions.test.ts
+++ b/apps/api/src/routes/agentVersions.test.ts
@@ -30,6 +30,14 @@ vi.mock("../db", () => ({
   },
 }));
 
+// #3499: the component=backup rewrite guard asks this resolver what the
+// versionless download route will actually serve. Default to "no promoted
+// row", which makes the guard fall back to comparing against the env version —
+// the behavior every pre-existing test in this file was written against.
+vi.mock("../services/promotedAgentVersion", () => ({
+  getPromotedComponentVersion: vi.fn(async () => null),
+}));
+
 vi.mock("../services/manifestSigning", () => ({
   // Simulate no DB-provisioned deployment keys by default so tests that
   // don't set env vars still get a soft-pass (no env + no DB = empty keyset).
@@ -88,6 +96,7 @@ import { writeRouteAudit } from "../services/auditEvents";
 import { syncFromGitHub } from "../services/binarySync";
 import { captureException } from "../services/sentry";
 import { ResponseTooLargeError, SsrfBlockedError } from "../services/urlSafety";
+import { getPromotedComponentVersion } from "../services/promotedAgentVersion";
 import { requiredPlatformTrustFor } from "../services/releaseAssetTrust";
 
 // Recursively searches an and()/eq() spy tree (see drizzleSpies above) for an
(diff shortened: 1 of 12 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three agents on the first commit (`code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`), then a focused `code-reviewer` re-review of the second commit's delta, since that commit changed failure semantics and re-cut the backup guard.

**Findings:** 2 rounds, all addressed; 0 outstanding.

Round 1 → `9b4bac8b2`. The one both the code-reviewer and the silent-failure-hunter independently called blocking: **`GET /agent-versions/latest` had no `ORDER BY` while the new resolver added `created_at DESC`.** Nothing enforces one promoted row per `(component, platform, arch, edition)` — the invariant is demote-then-insert, not a unique constraint — so a duplicate row would make the checksum half and the bytes half select *different* rows. That is #3499 reproduced inside its own fix, and silent. Also: the backup guard now derives what the route serves instead of restating its rule (a plain `isLatest` test would have broken backup self-heal on never-synced deployments); a lookup **fault** now 503s instead of falling back to the env version (that fallback *is* the bug — it reports a server-side DB fault to the end user as "Checksum verification failed"); missing rows reach Sentry under a registered event code.

Round 2 → `b4f2d5f23`. Two real defects in my own round-1 work: the backup guard derived the promoted row **unconditionally**, but the route only does so in the `github` branch — local mode serves one unversioned file off disk, so the guard silently ended backup self-heal there under `AGENT_AUTO_PROMOTE=false`. And the malformed-tag throw escaped the 503 handler as a bare 500.

**Tests:** 349 affected tests green single-run across 10 files (`vitest run` on `promotedAgentVersion`, `binarySource`, `binarySync`, `agents/download`, `agentVersions`, `agentVersionsLocalModeRoundtrip`, `supportPublic`, `viewers/download`, `helpers.agentUpdatePolicy`, `sentryEventCodes`); `tsc --noEmit` clean; `eslint` clean on all 12 touched files.

**Mutation-verified**, because `pr-test-analyzer` found the original tests could not detect a revert. Each of these previously survived and now fails: builders accepting the promoted version and *ignoring* it (a complete revert — passed all ~30k API tests before, now fails 5); `and(...)`→`or(...)`; `eq(isLatest,true)`→`ne(...)`; the `darwin`→`macos` mapping; and the local-mode guard branch.

**CI:** in flight on `b4f2d5f23` — please confirm green before merging. The two prior commits were fully green, including both `BINARY_SOURCE=github download path` and `BINARY_SOURCE=local download path`, which boot a real API against a live Postgres and are the end-to-end evidence that the DB lookup behaves outside the mocks.

**Status:** review-clean, awaiting maintainer merge. Two follow-ups worth separate issues, deliberately not done here: a **partial unique index on `isLatest`** (the shared `ORDER BY` papers over a missing constraint), and `services/aiToolsAgentMgmt.ts:111`, which reads `isLatest` with no component/platform/arch/edition filter at all and can return a helper or backup row as "the latest agent version" — pre-existing, unrelated to this path.

Issue #4239, opened 2026-08-30
Audit retention chooses per-policy full-table scans of audit_logs at production scale


bdunncompany:
## Summary
At ~22M audit rows, `pruneExpiredAuditLogs()` (apps/api/src/jobs/auditRetention.ts) processes each
retention **policy** serially, and for each one the cutoff computation plans as a **full Seq Scan of
`audit_logs`** hash-joined to a full scan of `audit_log_chain`. Because that cost is a function of
total table size rather than of how much the policy actually deletes, per-policy time is roughly
**constant regardless of the org's size**, and the whole pass scales with `policies × table size` on
the observed plan.

## Environment
Self-hosted single-node, Postgres 16, ~22M rows each in `audit_logs` / `audit_log_chain`, 14 active
retention policies. Default retention worker.

## Observed
A single manual run took **~13-22 minutes per policy, and the time did not track org size** — a
50-device org and a 4-device org each took ~18 min. 14 policies × that per-policy cost came to
~3.4 h total; the box ran IO-saturated (load ~7 on 4 vCPU) throughout. Prod stayed healthy.

## Evidence (plan shape)
Isolated cutoff subquery from `deleteChainPrefix`:
```sql
SELECT MIN(c2.chain_seq)
FROM audit_log_chain c2 JOIN audit_logs a2 ON a2.id = c2.audit_id
WHERE c2.org_id = <ORG> AND a2.timestamp >= (now() - (retention_days * interval '1 day'));
```
```
Finalize Aggregate
 -> Parallel Hash Join (a2.id = c2.audit_id)
    -> Parallel Seq Scan on audit_logs a2   (Filter: timestamp >= now()-'60 days')   -- whole table
    -> Parallel Seq Scan on audit_log_chain c2 (Filter: org_id = <ORG>)
```
The outer delete's plan also Seq Scans `audit_logs` to hash-join the deletable set.

`EXPLAIN (ANALYZE, BUFFERS)` of that cutoff SELECT for one org (non-parallel, post a 60-day prune, so
the tables are ~17.8M rows each):
```
Aggregate (actual time=200532..200532 rows=1)
  Buffers: shared read=1943843 (~14.8 GiB of block reads; some may be OS-cache hits), temp read/written=80840
  -> Hash Join (a2.id = c2.audit_id) (actual 84339..200220, rows=2312312)
[...]

PR #4365, opened 2026-09-01
fix(api): batch five unbounded retention DELETEs (#4343)

### LINKED ISSUE #4343: [API] Five retention jobs run unbounded DELETEs (agent/event/change/ip/snmp logs)
## Description
`agentLogRetention`, `eventLogRetention`, `changeLogRetention`, `ipHistoryRetention`, and `snmpRetention` each issue a single DELETE with no LIMIT/batch loop, unlike the well-built batched siblings (`deviceMetricsRetention`, `processSampleRetention`, ML/reliability/user-risk jobs). On a large fleet or after a backlog these hold a pooled connection for the whole delete — exactly the failure mode `scheduleRegistry.ts`'s header documents.

Additional issues found in the same sweep:
- `eventLogRetention` does a `selectDistinct(orgId)` over the full `device_event_logs` table on every run before deleting anything
- `snmpRetention` is hardcoded to 7 days with no env override, unlike every sibling
- The batched jobs with a `MAX_BATCHES` cap (ML/user-risk/reliability, 50 batches) silently never catch up if daily volume exceeds `batchSize × maxBatches` — `hasMore` is returned but nothing re-enqueues

## Proposed Fix
Convert the five jobs to the standard `LIMIT`-loop batch pattern; derive org lists from `organizations` instead of a distinct scan; add `SNMP_METRICS_RETENTION_DAYS`; re-enqueue (or log loudly) when a capped job reports `hasMore`.

## Affected Files
- `apps/api/src/jobs/agentLogRetention.ts`, `eventLogRetention.ts`, `changeLogRetention.ts`, `ipHistoryRetention.ts`, `snmpRetention.ts`
- `apps/api/src/jobs/mlOutputRetention.ts`, `userRiskRetention.ts`, `reliabilityRetention.ts` (catch-up behavior)

ToddHebebrand:
## Problem

`agentLogRetention`, `eventLogRetention`, `changeLogRetention`, `ipHistoryRetention` and `snmpRetention` each issued a **single DELETE with no LIMIT or batch loop**, unlike their well-built siblings (`deviceMetricsRetention`, `processSampleRetention`, ML/reliability/user-risk). On a large fleet or after a backlog each holds a pooled connection — and row locks on tables the agent write path is concurrently inserting into — for the entire statement. That is exactly the failure mode `jobs/scheduleRegistry.ts`'s header documents.

Three smaller defects from the same sweep:

- `eventLogRetention` ran `SELECT DISTINCT org_id` across the whole of `device_event_logs` on **every** run, just to learn a list `organizations` already holds.
- `snmpRetention` was hardcoded to 7 days with no env override, unlike every sibling.
- The capped jobs (ML / user-risk / reliability) returned `hasMore` and then said nothing actionable — a job that never caught up looked identical to one that did.

## Fix

New **`apps/api/src/jobs/retentionBatch.ts`** holds the shared primitives rather than a sixth hand-rolled copy of the loop (same consolidation rationale as `db/rowCount.ts` in #3894):

| Helper | Job |
|---|---|
| `pruneInCtidBatches` | the house `DELETE ... WHERE ctid IN (SELECT ctid ... LIMIT n)` loop, capped by `maxBatches`, reporting `hasMore` |
| `resolveRetentionDays` / `parsePositiveIntEnv` | non-positive and unparseable config **falls back** rather than clamping |
| `warnOnRetentionBacklog` | announces a capped sweep on `console.warn` |

The fall-back-not-clamp rule is load-bearing in two ways: `Math.max(1, NaN)` is NaN, which reaches `new Date(NaN).toISOString()` and throws `RangeError` on every run; and clamping a configured `0` to the 1-day floor would prune nearly the whole table on the next pass.

`warnOnRetentionBacklog` deliberately does **not** re-enqueue — the remainder is left for the next scheduled run, matching the contract `softwareRemediationRequestCleanup` already uses. Re-enqueueing a capped sweep is how one job monopolises a connection.

### Per job

All five now batch, accept `*_RETENTION_BATCH_SIZE` / `*_RETENTION_MAX_BATCHES` knobs (defaults 10000 / 50), and warn on backlog:

- **`agentLogRetention`** — `agent_logs`, rides `agent_logs_timestamp_idx`.
- **`changeLogRetention`** — `device_change_log`; still prunes on `created_at` (ingest time), *not* the `timestamp` column, preserving the pre-batching behaviour.
- **`ipHistoryRetention`** — `device_ip_history`; bound stays `<=` and only `is_active = false` rows are eligible.
- **`snmpRetention`** — `snmp_metrics`; **gains `SNMP_METRICS_RETENTION_DAYS`** and now reports a deleted count (it previously reported none).
- **`eventLogRetention`** — batches **per org**, and reads its org list from `organizations`. That is *safe*, not merely cheaper: `device_event_logs.org_id` is `NOT NULL REFERENCES organizations(id)`, so no event log can exist for an org absent from that table. Orgs are **not** filtered by status — an archived org's logs still need pruning. A corrupt per-org policy value is clamped rather than read as "delete this org's entire history" (the validator bounds the field to 7..365).

`mlOutputRetention`, `reliabilityRetention`, `userRiskRetention` gain only the loud backlog warning; their batching was already correct.

### On the cutoff type change

Cutoffs move from Drizzle `Date` params to ISO strings (postgres-js does not coerce `Date` in template-literal params). This is semantically identical, not a behaviour change: for `timestamp without time zone` Postgres parses the ISO text and drops the `Z`, yielding the same naive-UTC value Drizzle's encoder writes. `device_event_logs.timestamp` is the one `timestamptz` here and is cast explicitly. Column types were checked against `0001-baseline.sql` rather than assumed.

## Verification

- **167 tests green** across the 11 affected files, run serially (`--pool=threads --maxWorkers=1`) — includes the `workerEntrypointClosure` contract (97) and the `rowCount` no-private-copies guard (17).
- `tsc --noEmit` clean; `eslint` clean on all 18 changed files.
- **Every new assertion was mutation-checked.** Breaking the `hasMore` cap logic, the batch-loop break condition, and the per-org retention clamp each turned the corresponding test red before the source was restored — so the greens discriminate rather than confirm.

New test files: `retentionBatch.test.ts`, `agentLogRetention.test.ts`, `changeLogRetention.test.ts`, `snmpRetention.test.ts`, `ipHistoryRetention.test.ts`, `eventLogRetention.test.ts`.

No migrations, no schema changes, no RLS surface change — every job runs in the same `withSystemDbAccessContext` it did before.

## Notes for the reviewer

- **`auditRetention.ts` is deliberately untouched** — it is covered by the concurrent #4239 PR. This PR does not conflict with it.
- New env knobs are not added to `.env.example`; that file documents **no** retention knob today (`grep -ic retention` → 0), so the siblings' knobs are undocumented too. Worth a separate docs pass rather than a one-off here.
- Defaults are conservative: 50 batches × 10000 rows = 500k rows per run per table. A table with a deeper backlog drains across several runs and says so on `console.warn` each time.

Closes #4343

### DIFF
--- a/apps/api/src/jobs/agentLogRetention.test.ts
+++ b/apps/api/src/jobs/agentLogRetention.test.ts
@@ -1,4 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { SQL } from 'drizzle-orm';
+import { PgDialect } from 'drizzle-orm/pg-core';
 
 const {
   addMock,
@@ -16,7 +18,7 @@ const {
   removeRepeatableByKeyMock: vi.fn(),
   queueCloseMock: vi.fn(),
   workerCloseMock: vi.fn(),
-  withSystemDbAccessContextMock: vi.fn(async (fn: () => Promise<unknown>) => fn()),
+  withSystemDbAccessContextMock: vi.fn(async (fn: () => Promise<unknown>, _label?: string) => fn()),
   dbExecuteMock: vi.fn(),
   attachWorkerObservabilityMock: vi.fn(),
   capturedWorkerProcessor: { current: null as null | ((job: { data: Record<string, unknown> }) => Promise<unknown>) },
@@ -43,11 +45,17 @@ vi.mock('../db', () => ({
   db: {
     execute: (...args: unknown[]) => dbExecuteMock(...(args as [])),
   },
-  withSystemDbAccessContext: (fn: () => Promise<unknown>) => withSystemDbAccessContextMock(fn),
+  withSystemDbAccessContext: (fn: () => Promise<unknown>, label?: string) => withSystemDbAccessContextMock(fn, label),
+  runOutsideDbContext: (fn: () => unknown) => fn(),
 }));
 
 vi.mock('../services/redis', () => ({
   getBullMQConnection: vi.fn(() => ({ host: 'localhost', port: 6379 })),
+}));
+
+vi.mock('../services/sentry', () => ({
+  captureMessage: vi.fn(),
+  captureException: vi.fn(),
 }));
 
 vi.mock('./workerObservability', () => ({
@@ -110,7 +118,14 @@ describe('Agent log retention worker', () => {
       data: { retentionDays: 7, batchSize: 4, maxBatches: 5 },
     });
 
-    expect(withSystemDbAccessContextMock).toHaveBeenCalledTimes(1);
+    // ONE short context per batch, not one spanning the whole sweep:
+    // withDbAccessContext opens a transaction, so a single outer context would
+    // hold every lock until the last batch committed (worse than the unbounded
+    // DELETE this replaced).
+    expect(withSystemDbAccessContextMock).toHaveBeenCalledTimes(3);
+    expect(withSystemDbAccessContextMock.mock.calls.map((c) => c[1])).toEqual(
+      Array(3).fill('agentLogRetention.prune'),
+    );
     expect(dbExecuteMock).toHaveBeenCalledTimes(3);
     const sqlDump = JSON.stringify(dbExecuteMock.mock.calls);
     expect(sqlDump).toContain('DELETE FROM');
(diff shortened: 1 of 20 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — round 1: `code-reviewer` + `pr-test-analyzer`; round 2: `code-reviewer` on the delta (the round-1 fix restructured transaction semantics on a data-deletion path, so a second round was warranted).

**Findings:** 13 raised → 13 addressed across `fc770541d` and `09a685122`; 0 outstanding.

Round 1 — the finding that mattered most invalidated the original approach:

- **Batching delivered none of its stated benefit.** `withSystemDbAccessContext` → `withDbAccessContext` opens a real Postgres transaction and early-returns inside an existing context, so wrapping the loop in one outer context ran all batches in a *single* transaction — every lock held until the last commit. Strictly worse than the unbounded DELETE it replaced. Now each batch opens a fresh escaped context (`runOutsideDbContext(() => withSystemDbAccessContext(fn, label))`), the `softwareRemediationRequestCleanup` pattern. Round 2 verified this against `dbContextStorage`/`AsyncLocalStorage.exit()` semantics and confirmed it genuinely commits per batch with no context leak.
- Also fixed: unlabelled DB contexts (arrive unattributed in Sentry and crowd the per-scope throttle), backlog signal was `console.warn`-only, `resolveRetentionDays` silently shortening a configured window, `MAX_BATCHES` 50 → 200 (safe once batches commit independently), and `eventLogRetention` reporting a clean drain after a mid-sweep failure.

Round 1 test analysis proved **8 regressions passed the original suite**. The three worst are now covered and each was mutation-verified:

- Flipping the cutoff to `Date.now() + days` — *delete the entire table* — passed in 4 of 5 jobs. Now caught.
- Replacing `LIMIT ${batchSize}` with `LIMIT 1000000` — restoring the unbounded delete — passed 39/39. Tests now compile via `PgDialect().sqlToQuery` and assert `$n` placeholders plus bound values, so an inlined literal fails too.
- The module-default batch branch was never executed, yet it is the rolling-deploy path (jobs already in Redis carry no `batchSize`). Now covered.

Round 2 — three findings against the fix, all addressed:

- `retentionTarget`/`batches` weren't in `ALLOWED_TAG_NAMES`, so every backlog event reached Sentry carrying only `event_code` — all nine call sites collapsing into one useless issue. `retentionTarget` allowlisted with the required justification; `batches` dropped (it equals `maxBatches` whenever `hasMore`).
- That tag would have shipped **tenant UUIDs** — `eventLogRetention` passed `device_event_logs org=<uuid>` as the label. Split into a bounded `target` (tagged) and `detail` (console only), plus the per-target hourly throttle that per-org loop needed.
- `hasMore` counted failed but not *skipped* orgs; a skipped org is never pruned, so a run where every policy lookup threw deleted nothing and reported a clean drain.

**Tests:** 220 green across 13 affected files, run serially (`vitest --pool=threads --maxWorkers=1`) — includes the `workerEntrypointClosure` contract (97) and the `rowCount` no-private-copies guard. `tsc --noEmit` and `eslint` clean. Every new assertion was mutation-checked: the cutoff sign flip, the `LIMIT` decoupling, the batch-default branch, the context escape, the `hasMore` cap logic, and the Sentry tag boundedness each turned a test red before the source was restored.

**CI:** 0 failures. `Test API`, `Lint`, `Check Migrations`, CodeQL and the security/audit jobs all pass; `Test Web`, `Smoke Test` and the 4 integration shards were still running at the time of writing.

**Not in this PR:** `auditRetention.ts` is deliberately untouched — covered by the #4239 PR, no conflict.

**Status:** review-clean, awaiting maintainer merge. Worth a look before merging: `MAX_BATCHES` is now **per org** for `eventLogRetention`, so its per-run ceiling scales with org count (each batch still commits separately, so the lock-duration goal holds — only wall-clock per nightly run grows).

Issue #1002, opened 2026-05-30
[API] Audit hash-chain forks under concurrent writes — verify_chain false positives on busy instances


ToddHebebrand:
## Description

On busy instances, `audit_log_verify_chain()` reports breaks even with zero tampering. Surfaced on the **US droplet** immediately after the v0.68.2 deploy activated the audit hash chain for the first time on a high-traffic instance: ~20 false breaks accumulated within the first few minutes, **all** from live agent telemetry rows (`agent.sessions.submit`, `agent.security_status.submit`, `agent.patches.submit`, …), and the count grows continuously under load. The 89k historical (backfilled) rows verify clean — only concurrently-inserted live rows break.

This is a **pre-existing design bug** in the chain (PR #900), not a regression from the `convert_to` bytea fix (#994). The deploy merely ran the chain at write-concurrency for the first time. EU (low traffic, ~5k rows) shows 0 breaks but has the same latent bug.

**No availability or data impact:** every audit row is written and individually checksummed correctly; only the chain *linkage* forks, which degrades the tamper-detection signal (false positives make real tampering indistinguishable).

## Root Cause

The BEFORE INSERT trigger `audit_log_compute_checksum()` selects its predecessor with no serialization:

```sql
SELECT checksum INTO prev
FROM audit_logs
WHERE org_id IS NOT DISTINCT FROM NEW.org_id
  AND id <> NEW.id
ORDER BY timestamp DESC, id DESC
LIMIT 1;
```

When N transactions insert audit rows for the same org concurrently, each reads the *same* latest committed checksum as `prev` and chains off it → the chain **forks**. Verified on US: 6 distinct `prev_checksum` values were each used by 2–5 rows (a clean serial chain uses each exactly once), all within one 4-minute window of concurrent agent telemetry. `audit_log_verify_chain` walks `timestamp, id` order and flags every forked row as a break.

The same-timestamp variant of this was already noted as known future-work in the test suite:
`apps/api/src/__tests__/integration/audit-checksum.integration.test.ts:110-111` — *"requires a chain_seq bigserial + per-org advisory lock — tracked as future-task hardening."* In practice it manifests under plain write concurrency, not just same-transaction batches.

## Proposed Fix

1. **Serialize per-org inserts** in the trigger with a transaction-scoped advisory lock before the predecessor SELECT, so concurrent same-org inserts each see the prior committed checksum:
   ```sql
[...]

Issue #915, opened 2026-05-26
[security] Audit retention: breeze_app inherits breeze_audit_admin (single-process bypass possible)


ToddHebebrand:
**Severity:** MEDIUM (forensic-integrity defense-in-depth)
**Found by:** /security-review sweep of merged changes since v0.67.0

## Root cause

`apps/api/migrations/2026-05-25-i-audit-retention-role.sql:51`:

```sql
GRANT SELECT, DELETE ON TABLE audit_logs TO breeze_audit_admin;
GRANT breeze_audit_admin TO breeze_app;
```

The two-gate retention design — `SET LOCAL ROLE breeze_audit_admin` AND `SET LOCAL breeze.allow_audit_retention='1'` — is robust against SQL accidents, but both gates are reachable from the same connection. An attacker with SQL injection or runtime code execution in the API process can replicate the bypass in-process and `DELETE FROM audit_logs WHERE ...`.

Append-only triggers + REVOKE close the simple cases; this is the sophisticated-attacker case.

## Fix

Run retention in a separate Postgres connection pool that connects as `breeze_audit_admin` directly (no `SET ROLE` from `breeze_app`). Then `REVOKE breeze_audit_admin FROM breeze_app`. The API process holds two pools:

- `breeze_app` pool (most code) — cannot delete audit rows.
- `breeze_audit_admin` pool (retention worker only) — used solely by `jobs/auditRetention.ts`.

This makes audit deletion require both code-execution in a specific subsystem and access to a distinct credential.

## Related: external anchor (separate issue follows)

In-tree integrity is internally consistent but not anchored externally. After the privilege split, also export per-org HEAD checksums nightly to immutable external storage (S3 Object Lock or a write-only Postgres) so forged-chain-after-deletion is detectable post-hoc.

## Test

- Integration: connect as `breeze_app`, attempt `SET ROLE breeze_audit_admin` — should fail after the REVOKE.
- Confirm retention worker (new pool) still runs end-to-end.

PR #4366, opened 2026-09-01
perf(api): index-driven audit retention cutoff + batched deletes (#4239)

### LINKED ISSUE #4239: Audit retention chooses per-policy full-table scans of audit_logs at production scale

ToddHebebrand:
Closes #4239

## What was wrong

`pruneExpiredAuditLogs()` computed each policy's chain-prefix bound with `MIN(chain_seq)` over a full `audit_log_chain ⋈ audit_logs` join. `MIN` has to consume the whole join before it can answer, and the arm never constrains `a2.org_id`, so `audit_logs_org_timestamp_idx (org_id, timestamp DESC)` was unusable. At the reporter's ~22M rows the planner picked a Seq Scan of `audit_logs` hash joined to a Seq Scan of `audit_log_chain` — a cost that tracks **total table size, not the org** — repeated once per policy. The outer DELETE Seq-Scanned `audit_logs` a second time, unbounded.

## What changed

Both moves were pre-approved in the issue thread.

**1. Ordered early-stop cutoff (direction 1).** `ORDER BY chain_seq LIMIT 1` in place of `MIN(chain_seq)`. Identical value over an identical filtered set — a pure plan change, and it does not depend on timestamps being monotonic in chain order. The bound is also **hoisted out of the DELETE** so it is computed once per policy instead of re-planned inside every batch. The `COALESCE(..., MAX+1)` all-old fallback stays (it is an index lookup, not a scan).

**2. Batched deletes (direction 3).** The prefix cut and the unsealed-row sweep now run in `LIMIT`-bounded batches with a per-org ceiling, matching the sibling retention jobs (`mlOutputRetention`, `deviceMetricsRetention`, `eventDispatchWorker`): delete a bounded slice, stop when a batch comes back short. Tunable via `AUDIT_RETENTION_BATCH_SIZE` (default 5000) / `AUDIT_RETENTION_MAX_BATCHES` (default 200), and overridable per run from the job payload so an operator draining a backlog can widen the ceiling without a redeploy.

No migration, no new index, no schema change. Privilege separation, both bypass layers, per-policy transaction isolation and the "retention never UPDATEs the audit tables" guarantee are all untouched.

### The invariant, and why `ORDER BY` is load-bearing

`ORDER BY c.chain_seq` inside the batch subquery is **not cosmetic**. Each batch must take the LOWEST remaining `chain_seq` so that what has been deleted is always a contiguous prefix. Without it, `LIMIT` picks an arbitrary subset and stopping at `maxBatches` leaves holes mid-chain — exactly the permanent linkage break the #1002 prefix-cut design exists to prevent.

This is not theoretical. Deleting the `ORDER BY` and re-running the integration suite: with 6 expired rows, `batchSize: 2, maxBatches: 2`, Postgres deleted `old-3..old-6` and left `old-1, old-2` — a real mid-chain hole.

## Evidence

**Plan shape, measured on 270k seeded rows across 6 orgs (pg16, `EXPLAIN (ANALYZE, BUFFERS)`).** This reproduces the reported shape at a scale I can reach; it is not 22M-row evidence.

Cutoff SELECT, one org:

| | old (`MIN`) | new (ordered early-stop) |
|---|---|---|
| plan | Parallel Hash Join over **Seq Scan `audit_logs`** + **Seq Scan `audit_log_chain`** | `Limit` → Nested Loop → Index Scan + PK probes |
| rows discarded | 76,500 + 74,999 | 6,668 |
| buffers | **15,717** | **263** |
| exec time | 30.7 ms | 1.3 ms |

Equivalence checked on that data: both forms return `chain_seq = 6720`.

The DELETE plan changes from `Seq Scan on audit_logs` **twice** (once for the cutoff InitPlan, once for the outer hash join) to fully index-driven — `Limit` → Index Scan on `audit_log_chain` → Index Scan on `audit_logs_pkey`, with no Seq Scan of `audit_logs` at all.

**What I could not verify:** `EXPLAIN` from the reporter's real 22M-row tables. The issue asked for before/after at that scale and it is still the useful confirmation — the steady-state claim (walk ≈ prefix length) is a plan-shape argument, not a measurement at production size.

## Tests

- `apps/api/src/jobs/auditRetention.test.ts` — 22 passed. New coverage: cutoff uses an ordered early-stop and contains no `MIN(`; the bound is hoisted out of the DELETE; the prefix DELETE carries `ORDER BY c.chain_seq` + `LIMIT`; the loop continues until a short batch; the `maxBatches` ceiling stops it and reports the org as backlogged; a null cutoff skips the prefix DELETE but still sweeps; limits clamp to ≥1 so a `0` can never disable pruning. Ordinal mocking was replaced with statement-kind dispatch — batch loops have a data-dependent call count that ordinals cannot express.
- `apps/api/src/__tests__/integration/audit-retention.integration.test.ts` — 10 passed on real Postgres. New: the `old → young → old` chain-order invariant (deletes only the leading old prefix, keeps the trailing old row, `audit_log_verify_chain()` returns 0 breaks); a multi-batch prefix drains completely; **stopping at the ceiling leaves a contiguous suffix, not holes**; an all-young org is untouched.

Both of the load-bearing controls were **mutation-verified** — I broke the implementation each way and confirmed the specific test went red, then restored:
- remove `ORDER BY` from the batch DELETE → `deletes the prefix in ordered, LIMIT-bounded batches` fails, and the integration `contiguous suffix` test fails on real data.
- revert the cutoff to `MIN()` → `computes the prefix bound with an ordered early-stop` fails.

**Status:** `tsc --noEmit` clean; `eslint` clean on all three files; 202 unit tests green across the 8 affected files; 12 integration tests green across both audit-retention suites.

## Note for the reviewer

`RetentionStats` gains `unsealedRowsDeleted`, `batches` and `orgsWithBacklog`. The unsealed sweep previously deleted rows that were never counted anywhere, so a manual prune could report `0 rows deleted` while rows vanished; it is now reported separately rather than folded into `rowsDeleted`, which keeps that field's existing meaning.

### DIFF
--- a/apps/api/src/__tests__/integration/audit-retention.integration.test.ts
+++ b/apps/api/src/__tests__/integration/audit-retention.integration.test.ts
@@ -151,6 +151,136 @@ describe('audit-log retention pruning', () => {
     expect(postBreaks[0]?.n).toBe(0);
   });
 
+  /**
+   * Issue #4239 rewrote the prefix bound from `MIN(chain_seq)` over the full
+   * `audit_log_chain ⋈ audit_logs` join to an ordered early-stop
+   * (`ORDER BY chain_seq LIMIT 1`), and split the DELETE into `LIMIT` batches.
+   * The invariant those must not break is the one the prefix-cut design
+   * (#1002) exists for: chain_seq is COMMIT order, so timestamps can run
+   * old → young → old along the chain, and retention must delete ONLY the
+   * leading old prefix.
+   */
+  describe('chain-prefix invariant under the #4239 rewrite', () => {
+    // Each execute() is its own transaction, so statement order == chain_seq order.
+    const seed = async (rows: Array<{ action: string; ageDays: number }>) => {
+      for (const row of rows) {
+        await getTestDb().execute(sql`
+          INSERT INTO audit_logs (org_id, actor_type, actor_id, action, resource_type, result, timestamp)
+          VALUES (${orgId}, 'system', gen_random_uuid(), ${row.action}, 'test', 'success',
+                  now() - (${row.ageDays}::int * interval '1 day'))
+        `);
+      }
+    };
+
+    const survivors = async () =>
+      (await getTestDb().execute(sql`
+        SELECT a.action, c.chain_seq::text AS chain_seq
+        FROM audit_logs a
+        JOIN audit_log_chain c ON c.audit_id = a.id
+        WHERE a.org_id = ${orgId}
+        ORDER BY c.chain_seq
+      `)) as unknown as Array<{ action: string; chain_seq: string }>;
+
+    const chainBreaks = async () => {
+      const rows = (await getTestDb().execute(
+        sql`SELECT count(*)::int AS n FROM audit_log_verify_chain(${orgId})`,
+      )) as unknown as Array<{ n: number }>;
+      return rows[0]?.n;
+    };
+
+    beforeEach(async () => {
+      await getTestDb().execute(sql`
+        INSERT INTO audit_retention_policies (org_id, retention_days) VALUES (${orgId}, 30)
+      `);
+    });
+
+    it('old → young → old in chain order: deletes only the leading old prefix and leaves the chain verifiable', async () => {
+      await seed([
+        { action: 'old-leading', ageDays: 90 },
+        { action: 'young-blocker', ageDays: 5 },
+        { action: 'old-straggler', ageDays: 60 },
+      ]);
+      expect(await chainBreaks()).toBe(0);
+
+      const stats = await pruneExpiredAuditLogs();
+      expect(stats.errors).toBe(0);
+      // Only the row BELOW the first young entry goes. The 60-day straggler
+      // sits behind it in chain order and must survive this cycle — deleting
+      // it would punch a permanent hole in the linkage.
+      expect(stats.rowsDeleted).toBe(1);
+
+      expect((await survivors()).map((r) => r.action)).toEqual([
+        'young-blocker',
+        'old-straggler',
+      ]);
+      expect(await chainBreaks()).toBe(0);
+    });
+
+    it('drains a multi-batch prefix completely and stays verifiable', async () => {
+      await seed([
+        { action: 'old-1', ageDays: 95 },
+        { action: 'old-2', ageDays: 94 },
+        { action: 'old-3', ageDays: 93 },
+        { action: 'old-4', ageDays: 92 },
+        { action: 'old-5', ageDays: 91 },
+        { action: 'young', ageDays: 1 },
+      ]);
+
+      const stats = await pruneExpiredAuditLogs({ batchSize: 2, maxBatches: 50 });
+
+      expect(stats.rowsDeleted).toBe(5);
+      expect(stats.orgsWithBacklog).toBe(0);
+      expect((await survivors()).map((r) => r.action)).toEqual(['young']);
+      expect(await chainBreaks()).toBe(0);
+    });
+
+    it('stopping at the maxBatches ceiling leaves a contiguous suffix, not holes', async () => {
+      await seed([
+        { action: 'old-1', ageDays: 95 },
+        { action: 'old-2', ageDays: 94 },
+        { action: 'old-3', ageDays: 93 },
+        { action: 'old-4', ageDays: 92 },
+        { action: 'old-5', ageDays: 91 },
+        { action: 'old-6', ageDays: 90 },
+        { action: 'young', ageDays: 1 },
+      ]);
+
+      // 2 batches x 2 rows = 4 of the 6 expired rows; the ceiling stops the loop
+      // while full batches are still coming back.
+      const capped = await pruneExpiredAuditLogs({ batchSize: 2, maxBatches: 2 });
+      expect(capped.rowsDeleted).toBe(4);
+      expect(capped.orgsWithBacklog).toBe(1);
+      expect(capped.errors).toBe(0);
+
+      // The batches must have taken the LOWEST chain_seq values. Anything else
+      // (an unordered LIMIT) would leave a gap mid-chain here.
+      const remaining = await survivors();
+      expect(remaining.map((r) => r.action)).toEqual(['old-5', 'old-6', 'young']);
+      expect(await chainBreaks()).toBe(0);
+
+      // The next run continues the prefix rather than being stuck.
+      const rest = await pruneExpiredAuditLogs({ batchSize: 2, maxBatches: 50 });
+      expect(rest.rowsDeleted).toBe(2);
+      expect(rest.orgsWithBacklog).toBe(0);
+      expect((await survivors()).map((r) => r.action)).toEqual(['young']);
+      expect(await chainBreaks()).toBe(0);
+    });
+
+    it('an org whose rows are all inside the window is untouched and reports no backlog', async () => {
+      await seed([
+        { action: 'young-1', ageDays: 3 },
+        { action: 'young-2', ageDays: 2 },
+      ]);
+
+      const stats = await pruneExpiredAuditLogs({ batchSize: 1, maxBatches: 5 });
+
+      expect(stats.rowsDeleted).toBe(0);
+      expect(stats.orgsWithBacklog).toBe(0);
+      expect((await survivors()).map((r) => r.action)).toEqual(['young-1', 'young-2']);
+      expect(await chainBreaks()).toBe(0);
+    });
+  });
+
   // Regression guard: the bypass GUC must default to off. Without
   // setting it, `breeze_app` (even with the audit_admin role membership)
   // must still see the trigger fire on DELETE.
--- a/apps/api/src/__tests__/integration/audit-retention.integration.test.ts
+++ b/apps/api/src/__tests__/integration/audit-retention.integration.test.ts
@@ -108,11 +108,18 @@ describe('audit-log retention pruning', () => {
     expect(second.errors).toBe(0);
   });
 
-  // Regression guard: pruning must re-anchor the surviving chain so the
-  // verifier still passes. Without the re-anchor, the new oldest row's
-  // prev_checksum still references a deleted row and the verifier flags
-  // it as a tamper.
-  it('re-anchors the hash chain after pruning so audit_log_verify_chain passes', async () => {
+  // Regression guard for the #1002 trusted-anchor rule. NOTE: the name/comment
+  // here used to say retention "re-anchors" the chain — it does not, and cannot:
+  // migration 2026-06-11-g REVOKEs UPDATE from breeze_audit_admin and
+  // audit_log_chain_block_update raises on every UPDATE. Retention deletes a
+  // PREFIX and rewrites nothing; the verifier treats the first surviving entry's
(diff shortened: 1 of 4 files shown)


ToddHebebrand:

**Findings:** 24 raised → 23 addressed in `bbec891` + `920de5a`; 1 deliberately deferred (below). code-reviewer's verdict on the fix itself was **no must-fix bugs** — every critical invariant (chain-prefix under batching, no-UPDATE, both bypass layers armed before the new cutoff SELECT, RLS scope) verified to hold.

**The three that mattered most:**

1. **The batch `LIMIT` was pinned by nothing.** Number params are invisible in drizzle's flattened SQL text, so a statement binding `maxBatches` instead of `batchSize` read *byte-identically* and survived the entire suite — shipping a job that deletes 200 rows/org/night while reporting success. Added a `sqlParams()` control and confirmed it kills that mutant.
2. **`extractRowCount` returns 0 for any unrecognised-but-non-null shape**, and a 0 *ends* the batch loop. A driver or adapter change would have silently retained expired audit rows and reported a clean success — the same failure mode the cutoff guard already refuses. Now guarded symmetrically at the DELETE call site.
3. **Non-numeric job-payload limits produced `NaN`**, and `batches < NaN` is false, so an operator-enqueued drain could issue **zero** DELETEs and still report `orgsPruned: 1, errors: 0`. Validated rather than clamped.

Also: the unsealed sweep now repeats `org_id` on the outer DELETE rather than relying on ctid alone (ctid is unique per *partition*, and that statement runs in system scope where RLS would not catch it), and that path — which had never deleted a row in any test — now has a two-tenant integration case.

**Corrections to my own claims,** since they were wrong in the direction of overselling:
- The `~3.3 min / 14.8 GiB` numbers are the **reporter's**, measured at **~17.8M rows, not 22M**, with an OS-cache caveat I had dropped. Now attributed and corrected in code and above.
- My comment predicted the planner would prefer `audit_log_chain_org_seq_idx` at scale. #4239 measured the planner **rejecting** it (not covering — `audit_id` absent). Reworded to state plainly that the plan is unverified at production scale and an early stop is **not** guaranteed; if this recurs, the reporter's direction 2 (denormalised `logged_at` + covering index) is the follow-up.
- Batching bounds each **statement**, not the transaction — it buys **no** lock-duration relief. The cutoff hoist is the actual fix; I'd implied otherwise.

**Documented a load-bearing dependency:** hoisting the cutoff out of the DELETE is only safe because `audit_log_seal_one` takes a per-org advisory lock *before* allocating `chain_seq` and holds it through commit, making allocation order equal commit-visibility order. If that lock is ever relaxed, the hoist becomes unsound. That is now written down at the call site.

**Deferred (needs your call, not mine):** `stats.errors > 0` never fails the BullMQ job, so a total retention failure completes green and `removeOnComplete: {count: 10}` discards the evidence. Pre-existing, and making the job throw changes retry/alerting semantics fleet-wide. I added a loud `console.error` + tagged Sentry events instead. Worth its own issue.

**Tests:** 33 unit (`vitest run src/jobs/auditRetention.test.ts`), 213 across the 8 affected unit files, 14 integration on real Postgres (both audit-retention suites, including a new batched/ceiling case on the **secure dedicated-pool path** — the production path, previously single-row only). `tsc --noEmit` clean; `eslint` clean on all four files.

**Mutation-verified controls** (broke the impl, confirmed the specific test went red, restored) — 5 total: drop the batch `ORDER BY` (unit + integration both red; Postgres really deleted `old-3..old-6` and left a mid-chain hole), revert the cutoff to `MIN()`, restore `?? null` on the cutoff read, bind `LIMIT` to `maxBatches`, and drop both `org_id` predicates from the sweep (cross-tenant delete: 2 rows instead of 1).

**Heads-up on #4365** (issue #4343): it adds a shared `apps/api/src/jobs/retentionBatch.ts`. No file overlap with this PR, so no conflict either way. Consolidating is **not** a straight swap though — that helper is hardwired to the shared `db` handle, while this job must run on either the dedicated `breeze_audit_admin` pool or the `breeze_app` tx (#915), and the prefix cut is an ordered join-driven delete rather than a `ctid ... WHERE ts < cutoff` sweep. The unsealed sweep could adopt it; the prefix cut would need the helper to take an executor.

ToddHebebrand:
**CI note — `Lint Agent (Go)` failure was a stale-base artifact, now resolved by rebase.**

The job flagged two `errcheck` issues in `agent/internal/userhelper/client.go:172-173`. This PR changes **four TypeScript files and zero Go files**, so I traced it rather than touching that file.

Cause: those exact lines were unchecked (`c.conn.SendTyped(...)` / `c.conn.Close()`) at this branch's base `3ab97b8`. #4353 (`7941ec1`) then landed on main and fixed them to `_ = conn.SendTyped(...)` / `_ = conn.Close()`. Because the branch was three commits behind, `golangci-lint --new-from-rev` read my branch as *reintroducing* the unchecked calls and attributed them to this PR.

Rebased onto current `origin/main` (`6467a45`). The branch now carries #4353's fix, and the diff is still exactly the same four files. Main was never broken and I changed nothing in the agent.

Re-verified on the new base: `tsc --noEmit` clean, 77 unit tests green across the affected files. Integration suites were verified pre-rebase against real Postgres (14 passed) and no Go/DB-facing code moved in the rebase.

**Status:** unchanged — review-clean, awaiting maintainer merge.

PR #4369, opened 2026-09-01
fix(api): make POST /alerts/:id/dismiss a compare-and-swap (#4293)

### LINKED ISSUE #4293: [API] POST /alerts/:id/dismiss is the last single-alert transition without a CAS predicate — lost race silently clobbers dismissedAt/dismissedBy
Flagged independently by two reviewers during the #4101 work (PR #4288) and deliberately scoped out there.

`POST /alerts/:id/dismiss` still updates by id alone. Verified on `origin/main`, `apps/api/src/routes/alerts/alerts.ts:1139-1147`:

```ts
const [updated] = await db
  .update(alerts)
  .set({ status: 'dismissed', dismissedAt, dismissedBy: auth.user.id })
  .where(eq(alerts.id, alertId))     // ← no status predicate
  .returning();
```

Every other single-alert transition now carries a compare-and-set predicate: resolve via `buildResolveAlertCas` (`services/alertService.ts:296`, from #4083/#4099), and acknowledge/suppress via `buildAcknowledgeAlertCas` / `buildSuppressAlertCas` added in PR #4288. Dismiss is the last one updating unconditionally.

### Severity: lower than #4101, but not nil

Dismiss is deliberately valid from **any** status (the docblock at `:1113-1118` says so explicitly), so an unguarded write cannot reopen a resolved alert the way an unguarded acknowledge could. What a lost race does clobber is **provenance**: two concurrent dismissals, or a dismiss racing another transition, leave `dismissedAt`/`dismissedBy` reflecting whichever write landed last, with no signal that anything was overwritten. For an action described in its own comment as the terminal "make this go away for good" operation, "who dismissed this and when" is the field most likely to be asked about later.

The `if (!updated) return 500` gate is already present, so a zero-row outcome is at least not silent — it just cannot currently happen, because the predicate matches unconditionally.

### Suggested fix

Route it through the same private `buildAlertStatusCas` helper PR #4288 introduces, so all four transitions share one CAS shape and none can drift. Since dismiss is legal from any status, the predicate is about detecting concurrent modification rather than restricting source states — decide deliberately what it should pin (e.g. the status read a moment earlier, as the bulk path already does with `eq(alerts.status, alert.status)`) and what a lost race should return.

Note the two multi-row paths already do this correctly and are worth copying: the bulk action in `routes/alerts/alerts.ts` pins `eq(alerts.status, alert.status)`, and `mutateAlerts` in `routes/alerts/correlations.ts` pins `eq(alerts.status, 'active')`.

Best done after PR #4288 merges, since it introduces the helper this should reuse.

🤖 Filed by Claude Code from the #4101 work; the unguarded where-clause and the existing CAS helpers verified against `origin/main`.

ToddHebebrand:
Closes #4293

`POST /alerts/:id/dismiss` was the last single-alert transition still updating by id alone:

```ts
.where(eq(alerts.id, alertId))   // no status predicate
```

Resolve got a CAS in #4094/#4099, acknowledge and suppress in #4101/#4288 — all three route through a shared private `buildAlertStatusCas`. Dismiss did not.

## What the unguarded write actually cost

Narrower than #4101, and **not** a reopened alert: dismiss is deliberately legal from every other status, so a dismiss landing on a just-resolved alert is the intended outcome, not a bug.

The casualty is **provenance**. Two techs dismiss the same alert; both id-only UPDATEs match, so `dismissedAt`/`dismissedBy` end up describing whichever write landed *second* — while **both** callers get a 200, an ML feedback emit, and an audit row claiming they performed the transition. For the terminal "make this go away for good" action, "who dismissed this and when" is the field most likely to be asked about later.

## The fix

Adds `DISMISSIBLE_ALERT_STATUSES` (every enum value except `dismissed`) and `buildDismissAlertCas` through the **same** shared helper, so all four transitions now share one CAS shape that cannot drift apart.

**Deliberately not the snapshot-equality pin** (`eq(alerts.status, alert.status)`) the issue floated as an alternative. Dismiss is legal from any non-dismissed status, so snapshot equality would refuse a *legitimate* dismiss whenever someone concurrently acknowledged or suppressed the alert — a false conflict, not a real one. The bulk path keeps snapshot equality because a miss there degrades to a `skipped` count rather than a refused operation (already documented in `alertService.ts`).

## Two response codes change — both close a split, not tighten a rule

| Case | Before | After | Why |
|---|---|---|---|
| CAS loss (`!updated`) | `500` | `409` | The gate was written as a server fault back when an id-only UPDATE made it unreachable. It is now reachable, and what it reports is a client-visible conflict. |
| Pre-read "already dismissed" | `400` | `409` | #4099 (resolve) and #4288 (acknowledge) established that losing at the pre-read and losing at the CAS are the same real-world event and must not return two codes purely on timing. Leaving the `400` would have made the response a coin-flip on which side of the read the other dismissal landed. |

**Blast radius checked — nothing branches on it.** The web UI dismisses via `POST /alerts/bulk`, not this route (`AlertsPage.tsx` `handleDismiss` → `executeBulkAction`), and its error handling is generic ok/not-ok. Mobile, portal and e2e have no dismiss call site at all. No test outside `apps/api` asserts a dismiss status code. The endpoint is absent from `openapi.ts` (it always was); the docs table in `reference/api.mdx` is updated in this PR.

## Tests — both suites proven red against the pre-fix code before being committed

- **`alerts.dismissCas.test.ts`** — behaviour (409 + no fan-out for the loser; 200 with exactly one feedback/audit for the winner; `resolved` still dismissible) plus the compiled-SQL predicate. Reverting the route to its pre-fix shape reds 4 of these, including the compiled SQL collapsing back to a bare `"alerts"."id" = $1` — the bug itself.
- **`alertService.dismissCasSql.test.ts`** — compiled SQL and set semantics.
- **The cross-list drift guard** in `alertService.ackCasSql.test.ts` is extended to a fourth list **without weakening the existing one**: the "never admits a terminal status" loop is re-scoped to the three non-dismiss lists and still rejects *both* terminals for each, while dismiss gets its own explicit assertion (must contain `resolved` — it is the one transition legal from a terminal status; must never contain `dismissed`) plus enum-exhaustiveness. Adding `dismissed` to the set — a CAS that looks present but guards nothing — reds 6 tests.

## Verification

- `vitest run src/routes/alerts src/services/alertService src/services/aiToolsAlerts src/services/warrantyAlertEvaluator` — **37 files, 397 tests, all green**
- `tsc --noEmit` — clean (exit 0)
- `pnpm --filter @breeze/api lint` — clean (exit 0)
- No schema, migration or tenancy surface touched; no cascade/export-policy registration applies.

## Follow-up noted, not fixed here

`previousStatus` in the audit row and ML feedback is read from the pre-CAS snapshot, so under a race it can name a stale status. This is **pre-existing and shared with resolve and suppress**, not introduced here — but dismiss now has the widest CAS admission set of the four, so it has the largest window. Worth a tracked follow-up; fixing it would touch all four handlers and belongs in its own change.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, parallel). `type-design-analyzer` and `code-simplifier` were not applicable: no new types beyond a `readonly` status tuple, and the docblocks are deliberate rather than incidental.

**Findings:** 2 raised → both addressed in `1032b6cf`; 0 outstanding.

- **silent-failure-hunter (HIGH)** — `ALERT_DISMISS_CAS_LOST_MESSAGE` contradicted its own docblock. The docblock states the message must not claim "another request dismissed it" (an empty `RETURNING` is equally what an RLS-invisible or cascade-deleted row produces), then the string said *"it has already been dismissed"* — exactly that actor claim, quietly foreclosing the non-benign hypothesis. Reworded to *"it already reached the dismissed status"*, the same eligibility-fact shape `ALERT_CAS_LOST_MESSAGE` already uses.
- **comment-analyzer (Critical)** — the `buildDismissAlertCas` docblock claimed *"the two multi-row paths"*, copied from the resolve/acknowledge docblocks where two genuinely exist. Dismiss has only **one** (`/alerts/bulk`); there is no correlation-group dismiss, since `mutateAlerts` in `routes/alerts/correlations.ts` is typed `action: 'acknowledge' | 'resolve'`. Corrected, and the absence is now stated explicitly.

**Clean:** code-reviewer found no Critical or Important issues — it hand-traced every status-pair race and independently re-ran the touched suites (60/60) and `tsc`. pr-test-analyzer rated coverage adequate to merge with no blockers, and confirmed the compiled-SQL assertions are load-bearing rather than vacuous.

**Tests:** 30 files / 350 tests green across `src/routes/alerts` + `src/services/alertService` (397 across the wider alert domain incl. `aiToolsAlerts`, `warrantyAlertEvaluator`); `tsc --noEmit` clean; `pnpm --filter @breeze/api lint` clean.

**Red-first evidence** — both new suites were proven to fail against the pre-fix code before being committed, not merely written after it:
- Reverting the route to `.where(eq(alerts.id, alertId))` reds **4** tests, including the compiled SQL collapsing to a bare `"alerts"."id" = $1` — the bug itself.
- Adding `'dismissed'` into `DISMISSIBLE_ALERT_STATUSES` — a CAS that is present in the code but guards nothing — reds **6** tests.

**Two follow-ups deliberately not fixed here** (both pre-existing, neither a regression):
1. The CAS-lost branch logs nothing beyond the generic access line, so a genuine race and a vanished/RLS-invisible row are indistinguishable in logs. True of **all four** handlers — fixing it only for dismiss would recreate the inconsistency this issue exists to remove.
2. `/alerts/{id}/dismiss` has never had an `openapi.ts` entry, though acknowledge/resolve/suppress do. The docs table in `reference/api.mdx` is updated in this PR.

**Status:** review-clean, awaiting maintainer merge. Note the `409` on the already-dismissed pre-read is a deliberate public-API change from `400` — rationale and the client blast-radius check are in the PR body.

ToddHebebrand:
**Note on the `Lint Agent (Go)` failure at `1032b6cf` — it was a stale-base artifact, not a defect in this PR.**

This PR changes **zero Go files** (`git diff --name-only origin/main...HEAD` has no `.go`/`go.mod`/`go.sum` entries). The job runs:

```
golangci-lint run --new-from-rev="origin/${BASE_REF}" ./...
```

`--new-from-rev` diffs the branch against **current** `origin/main`. This branch was cut at `1a383416`, and main then advanced to `7941ec16` with a single commit touching 13 files under `agent/` (`userhelper`, `agentapp`, `breeze-desktop-helper`). Because those Go changes were absent from this branch, golangci-lint saw their lines as differing and attributed findings inside them to this PR.

Corroborating evidence: `Lint Agent (Go)` is **green** on the three most recent `main` runs, so the job itself is healthy — the failure only appeared against a branch whose base predated that Go commit.

**Fix:** merged current `origin/main` into the branch (`d931243a`) rather than force-pushing, so the phantom Go diff is gone and the base is fresh. A merge commit is harmless here since the repo squash-merges. Re-verified after the merge: all 7 changed files intact, the CAS still wired, and 30 files / 350 tests green across `src/routes/alerts` + `src/services/alertService`.

CI is re-running on `d931243a`. Review status is unchanged: 2 findings raised, both addressed, 0 outstanding.

PR #4382, opened 2026-09-01
feat(api): auto-dispatch edition migration for stranded self-host agents (#4072 follow-up)

ToddHebebrand:
## Summary

Closes the #4072 remediation loop. PR #4091 made the server withhold update offers a stranded self-host-build agent would refuse, leaving the device idling on its old version; PR #4102 added the identity-preserving 'Migrate Agent Edition (Windows)' system script but required a manual per-device operator run with hand-supplied presigned MSI URL + SHA. This PR dispatches that script automatically, server-side, at the exact moment the withhold fires.

Live motivation: on 2026-09-01 a partner reported two devices wedged in the `updating` retry loop (reboot/remote access blocked); a prod sweep found 13 live-wedged devices across 10 orgs in both regions, all needing manual intervention. The stranded band also turned out to include public v0.107.0 self-host installs, not just 0.105.x–0.106.x.

## How it works

- **Hook** (`routes/agents/heartbeat.ts`): in the main-agent #4072 withhold branch, fire-and-forget call into the new service. The heartbeat response never waits on it and a failure can never fail the beat.
- **Service** (`services/agentEditionAutoMigrate.ts`), in evaluation order:
  1. `AGENT_EDITION_AUTO_MIGRATE_ENABLED === 'true'` (default off; documented in `.env.example`)
  2. hosted-serving deployments only (hosted→self-host is refused by design and never auto-migrated)
  3. Windows/amd64 only (what the script + staged MSI support)
  4. org effective update policy + maintenance window (`updateGateAllows`) must allow an update now
  5. pin-honoring target resolution via the same `resolvePinnedUpgradeTarget` as the offer path — **a pin at/below the installed version holds auto-migration too**, so the org pins currently un-wedging prod act as the operator hold until deliberately lifted
  6. environment preconditions (base URL, staged MSI readable, system script row present) checked **before** the claim so a transient gap never consumes a device's attempt
  7. **atomic once-per-device claim**: `UPDATE devices SET edition_migration_dispatched_at = now() WHERE ... IS NULL RETURNING` — concurrent heartbeats race safely; a dispatched-but-failed MSI dance is never auto-retried into an uninstall/reinstall loop (Sentry has it; operator follows up)
  8. dispatch through `dispatchScriptToDevice` (`triggerType: 'policy'`) with `msi_url`, `msi_sha256`, `target_edition`
- **Raw MSI route** (`GET /api/v1/agents/download/windows/amd64/msi`): serves the staged installer verbatim — unlike the enrollment installer routes it embeds no per-download bootstrap token, so the bytes have a stable sha256 matching the pin the dispatcher computes from the same file. Deliberately disk-only (no S3 presign / github redirect) so route bytes and pinned hash can't diverge; `BINARY_SOURCE=github` deployments 404 here and the feature stays inert. Public/no-auth like every other binary route — a raw MSI enrolls nothing without a token.
- **Schema**: `devices.edition_migration_dispatched_at` (idempotent migration `2026-09-28-…`, sorts after newest committed `2026-09-27-…`) + `included` classification in `CORE_TENANT_EXPORT_POLICY`.

## Testing

- `agentEditionAutoMigrate.test.ts` (15): dispatch wiring incl. exact parameters, claim-before-dispatch ordering, lost-race stand-down, the full skip ladder (flag/edition/OS/arch/policy/already-dispatched/no-target/holdback-pin/missing MSI/missing script/missing base URL — all with zero DB writes), claim release + in-process no-retry on dispatch refusal, never-throws contract.
- `download.test.ts` (+2): MSI served verbatim; 404 does not disclose `AGENT_BINARY_DIR`.
- `heartbeat.test.ts` (+3): hook fires exactly on withhold with the right args, not on accept; a rejected hook promise never fails the heartbeat.
- All 7 affected unit suites green (374 tests); `tsc --noEmit` clean; eslint clean; migration-naming guard + `autoMigrate.test.ts` green.
- Not run: RLS/integration contract suites (no schema shape change — a nullable column on an already-registered table; export-policy suites will validate the registry entry in CI's Integration Tests job).

## Rollout notes

- Ships default-off. Enable per-droplet with `AGENT_EDITION_AUTO_MIGRATE_ENABLED=true` once this is deployed and the org tourniquet pins are lifted.
- The command is expected to report no result on success (the reporting agent is the one replaced); verification is the device returning online on a hosted build — the script's own description says this and progress is logged to `C:\ProgramData\BreezeMigration\migration.log` on the device.
- macOS/Linux stranded devices remain manual (no migration script exists for them).

https://claude.ai/code/session_01Kp14ziNayoMHL6mPfszsaE


ToddHebebrand:
Review round complete (internal /code-review high — Codex quorum unavailable, auth expired). 10 findings; applied in 61cf6f0:

**Fixed**
- **Detached-context bug (the real one):** the fire-and-forget hook inherited the heartbeat's org-scoped `withDbAccessContext` *transaction* via ALS; after the handler returned and the tx committed, the service's queries would hit a dead handle and the feature would be nondeterministically inert (invisible to unit tests that mock `db`). Now wrapped in `runOutsideDbContext + withSystemDbAccessContext`, same as the drain-commands path at `heartbeat.ts:334`.
- Streaming sha256 (no more ~30MB `readFileSync` + digest blocking the event loop post-deploy); `Readable.toWeb` on the MSI route for backpressure.
- 60s TTL cache on target resolution keyed `platform:arch:pin` (held-back stranded fleets no longer re-query `agent_versions` every beat).
- `claimed` reset immediately after a successful dispatch — a throw in the informational logging can no longer release the one-attempt claim.
- `envFlag()` vocabulary + boot-time boolean typo-guard in `config/validate.ts`; informational Sentry signal moved to `captureMessage` with registered event code `agent_edition_auto_migration_dispatched` (BREEZE-18); success log demoted from `warn`.

**Accepted / deferred with rationale**
- *At-parity edition swap* (stranded device already at the hosted target's version) stays parked until the next release moves the target — deliberate, because it preserves pin-as-hold semantics the rollout depends on; documented at the gate.
- *Restart re-arms a failed device* — acknowledged in the docstring; durable failure sentinel deferred (one Sentry event per persistent failure per deploy is acceptable signal, arguably desirable).
- *Name-based system-script lookup* — fails safe (never dispatches a wrong script) and self-heals at next boot via `ensureSystemLibraryScripts`; a stable slug is a `systemScriptLibrary` refactor for another PR.

ToddHebebrand:
Codex quorum round (gpt-5.6-sol, read-only) came back **no-merge** with 4 findings — all addressed in a770468:

1. **Pin-integrity gate (valid blocker):** the raw MSI route serves the deployment's single staged installer, but the resolved target could be an org pin below it — dispatch would have installed a version the tenant didn't select, and a staged release ahead of the promoted `isLatest` could leak past controlled promotion. Now dispatch requires `target === getGithubReleaseVersion()` (the release binaries-init staged), failing closed with a deduped warn on mismatch or when the deployment release is unknown.
2. **Compose parity (valid blocker):** the flag wasn't mapped in `docker-compose.yml`'s api environment — `envComposeParity.test.ts` was red and `.env` settings were a silent no-op. Mapped.
3. **Script secrets exposure (valid, pre-existing in #4102 but worse when automated):** the identity backup staged `secrets.yaml` (bearer token + mTLS key) in a ProgramData dir any local user could pre-create or junction. The script now refuses reparse points, forces a SYSTEM/Administrators-only ACL via `icacls` before staging anything, recreates the backup dir fresh each run, and deletes the staged copies after a successful install (`ensureSystemLibraryScripts` version-bumps the row on boot).
4. **Detached-launch structure (partially valid):** cheap non-DB precheck now runs in the heartbeat before any system context opens (flag-off = comparisons only); the claim additionally binds `org_id` + non-decommissioned status (device-move race); a THROWN dispatch keeps the claim since queue state is indeterminate — only typed `ok:false` refusals (provably nothing queued) release it; cold-cache MSI hashing got singleflight. Not taken: moving the launch after the org block closes — the connection hold is bounded to actual candidates (rare, once per device) and the restructuring cost isn't justified.

All affected suites green (701 tests incl. the previously-red parity contract), tsc + eslint clean.

PR #4383, opened 2026-09-01
feat(mobile): ticket push routing, notification prefs and Settings controls (#4336)

### LINKED ISSUE #3901: W07: Push categories for ticket assignment and SLA breach

### LINKED ISSUE #4336: W10: W10: Mobile half of W07 — ticket push routing, notification prefs and Settings controls
Mobile half of W07 (#3901). The backend half merged as #4281; #3901 auto-closed on that merge even though PR B was never built.

**Plan:** `docs/superpowers/plans/mobile/2026-08-29-ticket-push-categories.md`, **Tasks 13–17** (+18 docs, 19 verify).

- Task 13 — `parseTicketNotification`, `tickets` channel, badge-aware handler
- Task 14 — `pushRouting.ts` and `navigationRef.ts`
- Task 15 — `PushTapRouter` and `RootNavigator` wiring
- Task 16 — API client + `notificationPrefsSlice`
- Task 17 — Settings sheet controls

Note: the plan's Task 1 Step 3 says to stack the mobile branch on the API branch. Do NOT — that predates the API half merging, and a stacked PR runs zero CI. Branch from `main`.

ToddHebebrand:
Closes #4336

W10 of the mobile ticketing feature (#3206): the **mobile half** of W07 (#3901). The backend half shipped in #4281, so this PR consumes a contract that is already on `main` — it does not add or change any server code.

Plan: `docs/superpowers/plans/mobile/2026-08-29-ticket-push-categories.md`, **Tasks 13–18**. Tasks 1–12 (API) were done in #4281 and are untouched here.

## What ships

| Task | Change |
|---|---|
| 13 | `parseTicketData` / `parseTicketNotification` / `shouldSetBadgeFor`, the `tickets` Android channel, and a badge-aware foreground handler |
| 14 | `navigation/pushRouting.ts` (pure route resolver + cold-start replay guard) and `navigation/navigationRef.ts` (module-level ref with a one-slot buffer) |
| 15 | `PushTapRouter` + `RootNavigator` wiring (`ref`, `onReady`) |
| 16 | `services/ticketPushPrefs.ts` client + `store/notificationPrefsSlice.ts` with optimistic save and rollback |
| 17 | Two preference controls in the Settings sheet |
| 18 | Docs: ticket push categories, the preference endpoints, and two stale claims retired |

A technician now gets a push when a ticket is assigned to them or breaches its response/resolution SLA, tapping it opens `TicketDetail` (background **and** cold start), and the two categories are controllable from Settings.

## Contract verified against the shipped API, not the plan

The plan predates #4281, so the wire shape was re-read from `main` before writing the parser. `buildTicketPush` (`apps/api/src/services/expoPush.ts`) emits:

```ts
{ type: 'ticket', ticketId, reason: 'assigned' | 'sla_breached', target?: 'response' | 'resolution', internalNumber? }
```

`internalNumber` is a presentation extra the plan did not mention. The parser deliberately ignores it and returns only the routing-relevant fields, with a test pinning that a change to the presentation extras cannot break tap handling.

## Decisions worth a reviewer's attention

**`PushTapRouter` is a SIBLING of `ApprovalGate`, never a child.** `ApprovalGate` returns `<ApprovalScreen />` *instead of* its children while an approval is focused, so nesting the router would tear down its notification subscriptions for exactly the lifetime the spec requires taps to survive. If a later refactor moves it inside, the symptom is silent: pushes still arrive, taps just stop routing while an approval is on screen.

**Ticket pushes never touch the badge.** `reconcileApprovalNotifications` *sets* the badge to the pending-approval count, so the badge means "approvals waiting on you". A ticket push that bumped it would read as a phantom approval until the next reconcile silently corrected it. The test asserts the registered handler actually consults `shouldSetBadgeFor` — asserting the pure function alone would have stayed green against the old hard-coded `shouldSetBadge: true`.

**The SLA control suppresses PUSH only.** The in-app inbox row and the email are written on every breach regardless of the preference (spec D6, implemented server-side in #4281). The row copy says so explicitly: *"Your in-app inbox still records every breach."* A control that quietly means more than it says is how support tickets get filed against the notification system.

**`navigateToTicket` buffers one pending id.** react-navigation silently drops `navigate()` before the container is ready, which is precisely the cold-start tap. One slot, not a queue: a technician who taps two pushes during a cold start wants the last one they touched.

**Cold-start replay is identifier-guarded and persisted.** `getLastNotificationResponseAsync()` returns the same response for the life of the process, so replaying it unguarded yanks the technician back to a ticket on every navigator remount. The handled identifier is written to AsyncStorage so a relaunch does not replay the launching tap either.

## Deviations from the plan

1. **Branched from `main`, not stacked on the API branch.** The plan's Task 1 Step 3 predates #4281 merging. A stacked PR whose base is not `main` runs **zero** CI here (`ci.yml` triggers on `pull_request: branches: [main]`), and the API half is already on `main` anyway.
2. **`Closes #4336`, not `Closes #3901`.** #3901 auto-closed on the #4281 merge even though the mobile half was never built; #4336 is the tracker sub-issue that actually represents this work.
3. **The API client lives in a new `services/ticketPushPrefs.ts`, not appended to `services/api.ts`.** This follows the `services/tickets.ts` precedent (feature service over `coreRequest`), keeps `api.ts` from growing past 1000 lines, and — the reason that matters — makes the endpoint path unit-testable. `coreRequest` prefixes `/api/v1` while the default `request` helper prefixes `/api/v1/mobile`, where no such route exists; the wrong helper is a silent 404 indistinguishable from "nothing saved yet". There is now a test pinning it.
4. **`MainTabParamList.TicketsTab` widened to `NavigatorScreenParams<TicketsStackParamList> | undefined`** rather than casting the `navigate` call, as the plan's `[not-checked]` note allowed. A cast would let a rename inside `TicketsStackParamList` compile fine and fail only on a device.
5. **The preference rows read the slice directly** instead of taking six more props through `SheetBody`, which is already at twenty props and none of which the rest of the sheet uses.
6. **One extra doc fix, in scope.** `mobile.mdx` still claimed the Settings screen has "Push Notifications" / "Critical Alerts Only" toggles persisted via AsyncStorage. #3118 replaced those with a status readout. Corrected in the same section this PR was already editing.

## Verification

| Check | Result |
|---|---|
| `npx vitest run` (apps/mobile, full suite) | **81 files, 1040 passed, 3 skipped** |
| `npx tsc --noEmit` (apps/mobile) | **clean** |
| `pnpm --filter @breeze/docs build` | **167 pages, Complete** (Starlight fails on malformed MDX / broken links) |
| Migrations touched | **none** — no tenancy, cascade, export-policy or RLS surface in this diff |

Every new module was written red-first: the parser tests, the routing/nav-ref tests, the client tests and the slice tests were each run and observed failing before the implementation existed.

`test-mobile` in CI runs exactly these two commands, so a green job here is the same evidence.

## Not verified — needs a device pass before merge

There is **no iOS device build in this loop**, so the plan's Task 19 Step 2 checklist is outstanding. Everything below is covered by unit tests at the logic level but unproven end-to-end on hardware:

- assign from web → push arrives → tap from terminated / background / foreground
- force-quit then tap → opens `TicketDetail` after boot; relaunch normally → no navigation (replay guard)
- approval overlay focused, then a ticket tap → ticket revealed after the decision clears (this is the one that catches a regression moving `PushTapRouter` under `ApprovalGate`)
- badge unchanged after ticket pushes
- Settings: rows appear only when push is registered; airplane mode disables them; API down → toast + value rolls back

`buildTicketPush` sets APNs `category: 'BREEZE_TICKET'`, and `apns.ts` notes "No client category is registered in W07". This PR does not call `setNotificationCategoryAsync` — an unregistered category identifier is inert on iOS (default presentation, no custom actions), and W07 ships no notification actions for it to expose. Registering it is only worth doing when someone wants Acknowledge / Assign-to-me buttons on the lock screen. Flagging rather than deciding.

### DIFF
--- a/apps/docs/src/content/docs/features/mobile.mdx
+++ b/apps/docs/src/content/docs/features/mobile.mdx
@@ -143,6 +143,27 @@ The notification service exposes listeners for:
 - `addNotificationResponseReceivedListener` -- fires when the user taps a notification.
 - `getLastNotificationResponse` -- retrieves the notification that launched the app from a terminated state.
 
+### Ticket Notifications
+
+Technicians receive two categories of ticket push on iOS:
+
+| Category | Sent when | Governed by |
+|----------|-----------|-------------|
+| **Assigned to me** | A ticket is assigned to you by someone else (self-assignment never pushes) | `assignedEnabled` |
+| **SLA breach** | A ticket misses its response or resolution SLA | `slaScope` |
+
+Tapping either notification opens that ticket's detail screen, from a cold start as well as from the background.
+
+Notification bodies are lock-screen safe: they carry the ticket number and the customer organization only, never the ticket subject. The subject loads after you unlock and the app authenticates.
+
+**What the preferences do and do not suppress.** These settings govern **push delivery only**. Your in-app notification inbox and any configured email still record every assignment and breach, so turning pushes off never hides work from you -- it only stops your phone buzzing.
+
+Additional delivery rules:
+
+- **Quiet hours** configured on the device suppress ticket pushes outright; they are dropped, not deferred.
+- A **per-user burst cap** (20 pushes per 5 minutes) protects against bulk reassignment. Capped pushes are dropped; the in-app rows are still written.
+- `All tickets` fans out to at most 500 opted-in technicians per breach event.
+
 ### Notification Settings
 
 Users can configure notification preferences per mobile device through `PATCH /devices/:id/settings`:
@@ -153,7 +174,21 @@ Users can configure notification preferences per mobile device through `PATCH /d
 | `severities` | `string[]` | Filter notifications by severity: `critical`, `high`, `medium`, `low`, `info` |
 | `quietHours` | `object \| null` | Suppress notifications during specified hours with `start`, `end`, and optional `timezone` |
 
-The Settings screen in the app also provides local toggles for "Push Notifications" and "Critical Alerts Only" that are persisted via AsyncStorage.
+The Settings screen in the app shows a **Notifications** row that reports the real push-registration state for this phone, and links out to system Settings when the OS permission is the actual control. It is a status readout, not a toggle -- the app has no code path that could honour an in-app on/off for OS-presented pushes, and the earlier "Push Notifications" / "Critical Alerts Only" switches only wrote AsyncStorage keys that nothing consumed.
+
+Ticket push categories are a **per-user** preference (they follow you across every phone you sign in on), configured from **Settings** in the app or through the API:
+
+| Method | Path | Description |
+|--------|------|-------------|
+| GET | `/users/me/ticket-push-preferences` | Read the caller's ticket push preferences. Returns defaults when nothing has been saved |
+| PATCH | `/users/me/ticket-push-preferences` | Update one or both settings |
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `assignedEnabled` | `boolean` | `true` | Push when a ticket is assigned to you |
+| `slaScope` | `"off" \| "owned" \| "any"` | `"owned"` | `off` = no SLA pushes; `owned` = tickets assigned to you; `any` = every ticket in your MSP you have access to |
+
+Both endpoints act on the authenticated caller only -- there is no way to read or write another user's preferences, and a `userId` in the request body is rejected.
 
 ---
 
--- a/apps/docs/src/content/docs/features/notifications.mdx
+++ b/apps/docs/src/content/docs/features/notifications.mdx
@@ -341,7 +341,10 @@ Breeze also supports push notifications to registered mobile devices via Firebas
 Push notifications subscribe to the same `alert.triggered` event bus and run in parallel with the notification dispatcher. Delivery status is tracked in the `push_notifications` table with states: `pending`, `sent`, `stubbed`, `failed`.
 
 <Aside type="note">
-  APNS (iOS) push delivery is currently stubbed and not yet fully implemented. Android push via FCM is functional when `FIREBASE_SERVICE_ACCOUNT` is configured.
+  APNs (iOS) delivery is live: it powers approval requests and ticket push
+  categories. Android push via FCM is functional when `FIREBASE_SERVICE_ACCOUNT`
+  is configured; ticket push categories are iOS-only today — the server logs and
+  skips FCM tokens for them.
 </Aside>
 
 ---
--- a/apps/mobile/src/navigation/MainNavigator.tsx
+++ b/apps/mobile/src/navigation/MainNavigator.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import { View } from 'react-native';
 import type { NavigatorScreenParams } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
@@ -16,6 +17,7 @@ import { TicketsScreen } from '../screens/tickets/TicketsScreen';
 import { TicketDetailScreen } from '../screens/tickets/TicketDetailScreen';
 import { TimesheetScreen } from '../screens/time/TimesheetScreen';
 import { TimeSuggestionsScreen } from '../screens/time/TimeSuggestionsScreen';
+import { flushPendingNavigation } from './navigationRef';
 import { HomeIcon, SystemsIcon, TicketsIcon, TimeIcon } from '../components/TabIcons';
 import { TimerBar } from '../components/TimerBar';
 import { palette, fontFamily } from '../theme';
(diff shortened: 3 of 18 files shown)


ToddHebebrand:
**Review run:** two independent review passes over `git diff origin/main...HEAD` — (1) correctness / silent-failure, focused on the push-tap lifecycle, the optimistic-save reducer and test vacuity; (2) integration consistency against the already-merged API half (#4281), plus repo-contract compliance (store registration, navigation types, registration lists, doc-claim accuracy).

**Findings:** 8 raised → 8 addressed in `cae497f`; 0 outstanding.

1 blocker, 5 should-fix, 2 nits. The blocker is worth reading:

**Ticket taps taken during an approval takeover were stranded permanently.** `NavigationContainer`'s `onReady` fires **at most once per container instance** — react-navigation guards it with an `onReadyCalledRef` it never resets (`@react-navigation/core@7.21.5`, `BaseNavigationContainer.tsx:323-331`, read in `node_modules`). Container readiness is `listeners.focus[0] != null`, i.e. *a root navigator is mounted*. `ApprovalGate` renders `<ApprovalScreen />` **instead of** its children, so `MainNavigator` unmounts and readiness genuinely cycles `true → false → true` inside a single session. A tap arriving in that window buffered correctly and then had no second `onReady` to release it — silently dropped, no error, no log. That is exactly the behaviour the sibling-mount design exists to guarantee ("a ticket tap while an approval is focused navigates underneath and is revealed when the decision clears"), so the feature was broken in its own headline case. Fixed by flushing from `MainNavigator`'s mount effect, which is the event that actually corresponds to readiness returning.

Other findings addressed: expo delivers the app-launching response to the tap listener *as well as* through `getLastNotificationResponseAsync`, so the cold-start branch and the listener both acted on the same tap (`shouldHandleTap`, with the deliberately opposite no-identifier default to `shouldReplayResponse`); cold-start replay failures were `console.warn`-only and now reach Sentry; `handleTap` re-checks `cancelled` after its await; the toast said "Could not save" for a *load* failure on sheet open; and `UpdateTicketPushPreferences` was a `Partial<>` that let `{}` typecheck against a server schema whose `.refine()` requires at least one key.

**Concurrent-save rollback** deserves a note. The single `rollback` slot is only truthful while at most one save is outstanding. Two overlapping saves let the second snapshot the first's *unconfirmed optimistic* value as "last known good"; a double rejection then left a never-persisted setting on screen permanently, silently, until the next load. The controls already disable while saving, but a tap landing inside that render tick got through — so the invariant is now enforced in the slice via `createAsyncThunk`'s `condition` (with `dispatchConditionRejection` at its default, so a refused save dispatches nothing at all) rather than trusted to the component.

**Tests:** `apps/mobile` full suite green — **81 files, 1047 passed, 3 skipped** (`npx vitest run`); `npx tsc --noEmit` clean; `pnpm --filter @breeze/docs build` → 167 pages, Complete. Every new behaviour was written red-first and observed failing: the 6 review-fix assertions (`shouldHandleTap` ×3, `errorKind`, save serialisation ×2) were confirmed red before the fixes landed.

**Coverage gap, stated plainly:** the blocker fix itself is **not** covered by a unit test. `MainNavigator.tsx` is `.tsx`, which this repo deliberately excludes from the vitest include pattern. `navigationRef.test.ts` now contains a `ready → not-ready → ready` case that documents the contract at the ref level, but it passed *before* the fix too — the buffer was never the broken half; the missing flush caller was. Catching a regression here needs the device check below, so please do not drop that line from the checklist.

**CI:** `Test Mobile`, `Type Check`, `Lint`, CodeQL and the security jobs green on `cae497f`; `Test API` and the four Integration shards were still running at the time of writing — this diff touches no API, migration or tenancy surface, so they are unaffected by it, but they do gate `ci-success`.

**Status:** review-clean, **not merge-ready without a device pass.** The plan's Task 19 Step 2 checklist is outstanding — there is no iOS build in this loop, so tap-from-terminated, the cold-start replay guard, and above all *approval overlay focused → ticket tap → revealed after the decision* are unproven on hardware. That last one is the regression check for the blocker fixed here.

**Open question for the maintainer:** the API sets APNs `category: 'BREEZE_TICKET'` and `apns.ts` notes no client category is registered. This PR does not register one — an unknown category identifier is inert on iOS and W07 ships no notification actions to expose. Only worth doing if you want Acknowledge / Assign-to-me buttons on the lock screen. Flagging, not deciding.

ToddHebebrand:
**Update — rebased onto current `main`, full CI now green.**

`main` moved twice while this was in review. Rebased onto `660bc6f`; head is now `f999a6b`.

**One conflict, in `RootNavigator.tsx`, resolved semantically rather than mechanically.** #4237 (MFA clients and enrollment) landed a `mfaEnrollmentRequired` branch on the same JSX this PR edits. Both changes are kept: the container gets `ref` + `onReady`, and the MFA gate stays the outermost branch — so `PushTapRouter` is **not** mounted while enrollment is required, and a buffered ticket tap waits instead of navigating past a mandatory gate. That ordering is deliberate and commented in the file.

**CI on `f999a6b`: 0 failures, 0 pending** — including `Test API`, all four `Integration Tests` shards, `Type Check`, `Lint`, `Test Mobile` and the security jobs. This supersedes the "still running" line in my previous comment.

Local re-verification after the rebase: `apps/mobile` full suite **83 files, 1063 passed, 3 skipped** (file count rose because the rebase pulled in `main`'s new suites), `tsc --noEmit` clean.

`mergeStateStatus` is `BLOCKED` only for the required approving review. **Still not merge-ready without the iOS device pass** listed in the previous comment — in particular *approval overlay focused → ticket tap → revealed after the decision*, which is the on-hardware regression check for the blocker fixed in `cae497f`.

PR #4385, opened 2026-09-01
feat(mobile): ticket comment attachments (#4337 W11)

### LINKED ISSUE #4337: W11: W11: Mobile half of W08 — ticket comment attachments (FormData transport, native prebuild, composer, viewer)
Mobile half of W08 (#3902). The backend half merged as #4282; #3902 auto-closed on that merge even though PR B was never built.

**Plan:** `docs/superpowers/plans/mobile/2026-08-29-ticket-attachments.md`, **Tasks 21–26**.

- Task 21 — `FormData` bodies on `coreRequest`
- Task 22 — mobile packages, permissions and native prebuild
- Task 23 — mobile attachment service
- Task 24 — composer chips and feed rendering
- Task 25 — `AttachmentViewer` modal route
- Task 26 — PR finish

Heaviest of the three: new Expo modules mean a native prebuild, so this rides its own TestFlight cycle. Branch from `main`, never stack.

ToddHebebrand:
Closes #4337

Mobile half of W08 (#3902) — ticket comment attachments. The backend half
merged as #4282; this is PR B of that plan
(`docs/superpowers/plans/mobile/2026-08-29-ticket-attachments.md`, Tasks 21–26).
Branched from `main`, not stacked.

## What a technician gets

Attach photos (camera or library) and PDFs to a ticket comment from the phone.
Each pick uploads immediately and shows a chip; Send is blocked while anything
is still uploading; the feed renders images as a 3-column grid and PDFs as named
rows, with a full-screen viewer for images.

## Tasks

**21 — `FormData` bodies on `coreRequest`.** `requestWithPrefix` hardcoded
`Content-Type: application/json`. A multipart body needs that header **omitted**
so the runtime supplies its own boundary; a hand-set value has none, so the
server parses zero parts and the upload fails with a confusing 400. A
caller-supplied `Content-Type` is stripped case-insensitively for the same
reason. Also adds `getAuthImageHeaders()` (Bearer + device id) for components
that fetch bytes themselves, and exports `API_CORE_PREFIX` so raw-URL builders
cannot drift from the prefix `coreRequest` uses.

**22 — packages, permissions, native config.** `expo-image-picker`,
`expo-image-manipulator`, `expo-document-picker`, `expo-image`,
`expo-file-system` and `expo-sharing` on the `~57.x` line. Camera + photo-library
usage strings, Android `CAMERA`. The stale `NSCameraUsageDescription` ("Camera
access for QR code scanning" — the app has no QR scanner) is rewritten to the
real reason; App Review rejects usage strings that do not match observed
behaviour.

**23 — attachment service.** Pickers (always `exif: false`, D15), `prepareImage`
(long edge ≤ 2048, re-encoded to JPEG), one-file-per-request upload with a 120 s
deadline, and a typed `AttachmentUploadError` mapping every server code to its
own message. Size and type are checked client-side **before** the request, so a
12 MB photo does not cost a minute of cellular upload to learn it is a 413.

**24 — composer chips and feed rendering.** See the commit message for the
detail; the load-bearing rules live in `attachmentComposer.ts` and are tested.

**25 — `AttachmentViewer` modal route.** Images render inline via `expo-image`
with `cachePolicy: 'memory-disk'` and the auth headers. PDFs are downloaded to
the cache and handed to the OS share sheet — the content route is authenticated,
so `Linking.openURL` would 401, and React Native has no PDF view.

## Deviations from the plan (all deliberate)

1. **`services/ticketAttachmentContract.ts` is new** — not in the plan's file
   structure. The plan assumed mobile could import `TICKET_ATTACHMENT_LIMITS`
   from `@breeze/shared`; **`apps/mobile` has no `@breeze/shared` dependency**
   (no entry in `package.json`, no tsconfig path, zero imports — Metro resolves
   no workspace packages here). Limits are therefore mirrored locally, exactly
   as `services/tickets.ts` already mirrors `TICKET_STATUS_TRANSITIONS`. The
   contract module also exists because the native-facing service transitively
   loads Flow-typed `react-native` source that Vitest cannot parse — splitting
   the pure half out is what makes the composer/feed logic testable at all.

2. **Tests are pure-logic modules, not component renders.** The plan's Tasks
   24/25 ask for tests that open action sheets and assert navigation.
   `apps/mobile/vitest.config.ts` deliberately includes `src/**/*.test.ts` and
   **not** `.tsx` — "the mobile app has no React Native test runtime
   configured". So the rules live in `attachmentComposer.ts` (tested) and the
   `.tsx` files are thin renderers, following the existing
   `timerActions.ts` / `timerOutcomeEffects.ts` / `timerBarLogic.ts` pattern.
   The `AttachmentViewer` param-list contract is enforced by `tsc` instead.

3. **Attach row instead of an action sheet.** `ActionSheetIOS` is iOS-only and
   `Alert.alert` degrades past three buttons on Android — which is exactly this
   count plus Cancel. A visible three-button row works identically on both
   platforms with no new dependency and costs one tap instead of two.

4. **`expo-file-system` and `expo-sharing` added** beyond the plan's Task 22
   list. Task 25's "PDFs download to cache and open the share sheet" cannot be
   built without them.

5. **Non-deprecated image-manipulator API.** The plan predates it; `manipulateAsync`
   is deprecated in `expo-image-manipulator@57`, so this uses the contextual
   `ImageManipulator.manipulate(uri).resize(...).renderAsync().saveAsync(...)`.

## Precondition checked (Task 21 asked for this explicitly)

**Mobile tokens are not scope-limited for tickets.** `tickets:write` is not a
JWT claim at all — `requirePermission` (`apps/api/src/middleware/auth.ts:818-848`)
resolves permissions per-request from the user's role via `getUserPermissions`,
identically for mobile and web sessions. Mobile login goes through the same
`routes/auth/login.ts` flow. Not a blocker.

Also verified the upload route's `ORG_CONTEXT_REQUIRED` guard fires only for
`auth.scope === 'organization'` without an `orgId`, so partner-scoped mobile
tokens pass through — mobile's `coreRequest` does no `?orgId=` injection the way
web's `fetchWithAuth` does, and does not need to.

## Verification

- `apps/mobile` full suite: **80 files, 1085 passed, 3 skipped** (`vitest run --pool=threads --maxWorkers=2`).
- `tsc --noEmit` in `apps/mobile`: clean.
- `pnpm install --frozen-lockfile` succeeds (CI's install mode).
- `apps/mobile` defines no `lint` script, so `turbo run lint` does not cover it.
- Rebased onto current `origin/main` (`51e458b72`) and re-verified green.

## ⚠️ This build is native, not OTA — and is UNTESTED on a device

Six new Expo modules mean `npx expo prebuild` and a **new native binary**. This
cannot ship as an OTA update and **rides its own TestFlight cycle.**

**I could not run a native build or a device pass in this environment**, so the
plan's Task 26 Step 2 device transcript is *not* included. The following are
**unverified** and must be exercised on a real device before merge-to-release:

- [ ] `npx expo prebuild --clean` succeeds (known hazard: Ruby 4 / CocoaPods gem breakage — an environment issue, not a code bug)
- [ ] iOS camera capture and library pick; HEIC selected → transcoded → accepted
- [ ] Served bytes carry no EXIF/GPS
- [ ] PDF pick, and PDF open via the share sheet
- [ ] Airplane mode mid-upload → failed chip offers Retry
- [ ] Android 13 media-permission grant
- [ ] Portal-visible public comment vs invisible internal comment

Everything above is covered by unit tests at the logic level; what is untested is
the native layer those tests mock.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (correctness, server-contract and native-API conformance) and `silent-failure-hunter` (error handling).

**Findings:** 5 raised → 4 fixed in 9db79110d, 1 judged not-a-bug; 0 outstanding.

- `code-reviewer`: **0 correctness findings.** It verified the client's URL paths, multipart field name, response shape, `attachmentIds` contract and every error-code string against `apps/api/src/routes/tickets/attachments.ts` and `routes/tickets/tickets.ts`, and checked every Expo call against the installed v57 `.d.ts` files rather than from memory (`ImageManipulator.manipulate`, `File.downloadFileAsync`, `Sharing.shareAsync`, `DocumentPicker.getDocumentAsync`, `ImagePicker.launch*Async`). It also confirmed the `chipsRef`/`chips` mirror cannot diverge and that the JSON path through `coreRequest` is genuinely unchanged.
- `silent-failure-hunter`: 5 findings, all triaged on the merits.

**Fixed (9db79110d):**
1. **The pickers could reject, not just return a failure outcome.** A native picker can throw (picker already open, an iCloud file that will not download, an unlinked module) and the tap site dispatches with `void` — so it was an unhandled rejection: no toast, no Sentry, no chip, and a technician who tapped Camera and saw *nothing happen at all*. Pickers are now total by contract via `runPicker`, with a new `failed` outcome the composer surfaces.
2. **`toAttachmentError` ignored `statusCode`.** A reverse proxy rejects an oversized body before the API sees it and answers with HTML, so there is no JSON `code` — only a 413. That collapsed to a *retryable* "check your connection", pointing at the wrong cause and inviting a retry that can never succeed. Added a 413/415/429/503 status fallback; body `code` still wins, and 409 is deliberately excluded (it is either `TICKET_DELETED` or `ATTACHMENT_NOT_CLAIMABLE`, and only the body says which).
3. **Neither `<Image>` handled `onError`.** Resolving the URL says nothing about whether the bytes load, so a stale token or a deleted attachment left the viewer spinning forever with no way out but backing out of the modal. Both the viewer and the feed thumbnail now render a failed state; the viewer offers a retry that remounts the `Image` rather than re-serving the cached failure.
4. **`ATTACHMENT_NOT_CLAIMABLE` copy was unreachable** — `submitComment`'s catch never routed through `toAttachmentError`. It does now for that code.

Also corrected two comments that claimed more than the code did (the `getAuthImageHeaders` Sentry-parity claim, and `useAttachmentSource`'s "the tap path reports its own failures", which is untrue for images).

**Not fixed — deliberate, not a defect:** `prepareImage` re-encodes PNGs to JPEG, losing alpha. The re-encode *is* the EXIF/GPS strip (spec D15) and JPEG q0.8 is what the plan specifies, so this is the plan's call rather than an oversight. Flagging it so it is a conscious choice.

**Tests:** `apps/mobile` full suite green — **80 files, 1094 passed, 3 skipped** (`vitest run --pool=threads --maxWorkers=2`); +9 tests came from the review fixes, each written red first. `tsc --noEmit` clean. `pnpm install --frozen-lockfile` succeeds.

**CI (head `9db79110d`):** 53 pass, 0 fail, 7 long-running jobs still going (Test API, Test Web, the four Integration shards, CodeQL) — none of which touch `apps/mobile`. **Test Mobile** and **Lint** both pass.

**Status:** review-clean. **Not ready to merge on CI alone** — see the native-build caveat in the PR body: six new Expo modules mean this is a native build, not OTA, and the device pass (camera, HEIC transcode, EXIF absence, PDF share, mid-upload airplane mode, Android 13 permissions) is **unverified** and needs its own TestFlight cycle.

PR #4393, opened 2026-09-01
fix(mobile): getServerUrl throws on a failed read instead of returning null (#4002)

### LINKED ISSUE #4002: getServerUrl cannot distinguish 'not configured' from 'read failed', so a keychain error silently routes a self-hosted tenant to the hosted default host

ToddHebebrand:
## Summary

`getServerUrl()` folded a SecureStore read **failure** (locked keychain, SecureStore unavailable) into the same `null` return as a genuine "nothing configured" absence. Every one of its ~11 call sites does:

```ts
const baseUrl = (await getServerUrl()) || FALLBACK_API_BASE_URL;
```

`FALLBACK_API_BASE_URL` is the hosted US host in all build profiles, so a transient keychain read failure silently rerouted a self-hosted or non-US tenant to the hosted default — including on the **auth path** (login/MFA request bodies going to a server the tenant never configured).

## Fix

`getServerUrl()` now:
- Returns `null` **only** when SecureStore genuinely has nothing stored (a successful read of an absent key) — callers are free to fall back in that case.
- **Throws `ServerUrlReadError`** when the read itself fails, instead of catching it into `null`.

Because the throw happens inside the same `await getServerUrl()` expression every caller already uses, the failure now propagates through that `await` and never reaches the `|| FALLBACK_API_BASE_URL` — **every existing call site fails closed with no per-caller change needed.** Verified by reading each of the ~11 call sites: none wraps the `getServerUrl()` call in a try/catch that would re-swallow it (the one call site inside a try/catch, `aiChat.ts`'s streaming path, already routes the caught error to `opts.onError(...)`, which is the correct surfaced-failure behavior).

`getAccountDeletionUrl` (which calls `getServerUrl` directly) now propagates the same error. Its one caller that awaited it outside a try/catch (`SettingsSheet.tsx`'s delete-account flow) is updated to catch it and show an error toast instead of leaving an unhandled promise rejection.

`ServerUrlReadError` follows the existing `SecureWipeError` convention in `services/auth.ts`: sets `.name` explicitly (RN/Hermes `instanceof` across bundles is unreliable) and preserves the original error as `.cause`.

## Testing

- `apps/mobile/src/services/serverConfig.test.ts` — new coverage for both states: a successful read of an absent key resolves `null`; a failed read throws `ServerUrlReadError` (with `.name` and `.cause` assertions); `getAccountDeletionUrl` propagates rather than silently falling back.
- `apps/mobile/src/services/api.serverUrlFailure.test.ts` (new) — caller-level regression test through the real `requestWithPrefix`/`login` path: a server-URL read failure rejects `login()` with `ServerUrlReadError` and **the request is never sent** (`fetchWithTimeout` not called), vs. a genuine `null` (unconfigured) still reaching `FALLBACK_API_BASE_URL` as before.
- All 12 other test files touching `getServerUrl` (aiChat, search, approverDevice, ticketAttachments, api.mfa, api.logout, api.formdata, etc.) re-run clean — no regressions from the changed contract.

```
pnpm exec vitest run src/services/serverConfig.test.ts src/services/api.serverUrlFailure.test.ts --pool=threads --maxWorkers=2
# 2 files, 15 tests passed

pnpm exec vitest run <12 other affected files> --pool=threads --maxWorkers=2
# 12 files, 132 tests passed

pnpm exec tsc --noEmit
# clean
```

Closes #4002

### DIFF
--- a/apps/mobile/src/screens/chat/components/SettingsSheet.tsx
+++ b/apps/mobile/src/screens/chat/components/SettingsSheet.tsx
@@ -262,11 +262,23 @@ export function SettingsSheet({ visible, onCancel }: Props) {
             // We track the user-intent click, not the actual server-side
             // deletion request (that happens on the web flow).
             track('account_deletion_requested');
-            // The deletion page is served on the server the user selected at
-            // sign-in (e.g. https://us.2breeze.app/account/delete), not a
-            // hardcoded marketing domain. Resolve it at tap time.
-            const url = await getAccountDeletionUrl(FALLBACK_API_BASE_URL);
-            await safeOpen(url);
+            try {
+              // The deletion page is served on the server the user selected
+              // at sign-in (e.g. https://us.2breeze.app/account/delete), not
+              // a hardcoded marketing domain. Resolve it at tap time.
+              //
+              // getAccountDeletionUrl throws ServerUrlReadError when the
+              // stored server can't be read (rather than opening the wrong
+              // tenant's deletion page) — surface that instead of letting the
+              // promise reject silently.
+              const url = await getAccountDeletionUrl(FALLBACK_API_BASE_URL);
+              await safeOpen(url);
+            } catch {
+              setToast({
+                kind: 'error',
+                text: 'Could not open the deletion page. Please try again.',
+              });
+            }
           },
         },
       ],
--- a/apps/mobile/src/services/api.serverUrlFailure.test.ts
+++ b/apps/mobile/src/services/api.serverUrlFailure.test.ts
@@ -1,0 +1,86 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Caller-level regression test for #4002: a SecureStore read failure for the
+// configured server URL must fail the request, not silently reroute it to
+// FALLBACK_API_BASE_URL (the hosted US host). This exercises the real
+// `requestWithPrefix` code path (via the exported `login`), not just the
+// `serverConfig` unit in isolation, since the bug was in how a caller
+// combined `getServerUrl()` with `|| FALLBACK_API_BASE_URL`.
+
+const secureValues = new Map<string, string>();
+const fetchWithTimeout = vi.fn();
+const serverConfig = vi.hoisted(() => ({
+  getServerUrl: vi.fn(async (): Promise<string | null> => 'https://api.example.test'),
+}));
+const secureStore = vi.hoisted(() => ({
+  getItemAsync: vi.fn(async (key: string) => secureValues.get(key) ?? null),
+  setItemAsync: vi.fn(async (key: string, value: string) => { secureValues.set(key, value); }),
+  deleteItemAsync: vi.fn(async (key: string) => { secureValues.delete(key); }),
+}));
+
+vi.mock('expo-secure-store', () => ({
+  WHEN_UNLOCKED_THIS_DEVICE_ONLY: 'device-only',
+  getItemAsync: (...args: Parameters<typeof secureStore.getItemAsync>) => secureStore.getItemAsync(...args),
+  setItemAsync: (...args: Parameters<typeof secureStore.setItemAsync>) => secureStore.setItemAsync(...args),
+  deleteItemAsync: (...args: Parameters<typeof secureStore.deleteItemAsync>) => secureStore.deleteItemAsync(...args),
+}));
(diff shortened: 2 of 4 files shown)


ToddHebebrand:
**Review run:** Manual code-reviewer + silent-failure-hunter + pr-test-analyzer pass (subagent pool was saturated by concurrent sibling issue-fixers, so I ran the same checks directly rather than dispatching agents).

**Findings:** 0 blocking findings. Traced all 11 `getServerUrl()` call sites (`api.ts`, `aiChat.ts` x2, `approvals.ts`, `approverDevice.ts`, `connectedApps.ts`, `mobileDevices.ts`, `search.ts`, `systems.ts`, `systemsRealtime.ts`, `ticketAttachments.ts`) to confirm none re-swallows the new `ServerUrlReadError` throw back into a silent fallback — 10 have no enclosing try/catch (propagates straight through), and the one that does (`aiChat.ts`'s streaming IIFE) already surfaces via `opts.onError(...)`. `approverDevice.ts`'s `runAttempt` catches broadly by pre-existing, documented design ("FAILS OPEN — never throws, never blocks login") and records `exception:${err.name}` rather than losing it silently — not a regression. `SettingsSheet.tsx`'s `getAccountDeletionUrl` caller (the one call site outside a try/catch) now shows an error toast instead of an unhandled rejection.

**Tests:** `apps/mobile` affected suite green — `serverConfig.test.ts` + new `api.serverUrlFailure.test.ts`: 15 passed (`vitest run src/services/serverConfig.test.ts src/services/api.serverUrlFailure.test.ts --pool=threads --maxWorkers=2`). All 12 other test files touching `getServerUrl` re-run clean: 132 passed, no regressions from the changed contract. `tsc --noEmit` clean.

PR #4394, opened 2026-09-01
fix(web): surface permanent-delete's agent-unreachable warning (#4368)

### LINKED ISSUE #4368: Permanent delete drops the API's "agent could not be reached" warning — operator sees a green success toast while the agent is still installed
**Where:** `apps/web/src/services/deviceActions.ts` — `permanentDeleteDevice` is typed `Promise<{ success: boolean }>` and returns `data.data ?? data`, so `agentUninstallSent` and `warning` are discarded. Both call sites — `apps/web/src/components/devices/DeviceDetailPage.tsx` and `apps/web/src/components/devices/DevicesPage.tsx` — `await` it and then show a hard-coded green success toast without inspecting the body.

**The gap:** `DELETE /devices/:id/permanent` returns, on success:

```json
{ "success": true, "agentUninstallSent": false,
  "warning": "The agent could not be reached for remote uninstall. You may need to manually remove it from the endpoint." }
```

That warning is never shown. The operator sees `"<hostname> has been permanently deleted"` in green while the device row is gone, the agent is still installed and running on the endpoint, and Breeze can no longer reach it to clean up.

**Why it matters more now:** #3817 (PR #4357) moved the `SELF_UNINSTALL` dispatch to after the cascade commits, which is correct — it removes the "agent uninstalls itself while the device row survives" hazard on every rollback path. The accepted cost of that ordering is the inverse race: the agent dropping between commit and dispatch. The route already reports that honestly in the 200 body; the web layer is where it gets dropped.

Worth noting this is likely the **common** case, not a rare race: permanent delete requires `decommissioned`, and decommissioning force-closes the agent socket (`core.ts` `disconnectAgent(..., 4041, 'Device decommissioned')`) with the WS handshake rejecting reconnects for decommissioned devices. So `isAgentConnected` is false for most devices reaching this route, both before and after #3817. Pre-existing — #3817 neither created nor worsened it — but it is the last mile of that fix's safety story.

**Suggested fix:** widen the return type and branch the toast. `showToast` already supports `type: 'warning'`.

```ts
export async function permanentDeleteDevice(
  deviceId: string
): Promise<{ success: boolean; agentUninstallSent?: boolean; warning?: string }> { … }

const result = await permanentDeleteDevice(device.id);
showToast(result.warning
  ? { type: 'warning', message: `${device.hostname} was deleted, but ${result.warning}`, duration: 10000 }
  : { type: 'success', message: `${device.hostname} has been permanently deleted` });
```

Needs a web test pinning the warning branch — per `quiet_failure_ui_needs_e2e_proof`, a hide-on-failure UI is exactly the shape that regresses silently.

Split out of the review of PR #4357 so the ordering fix does not carry a web change.

ToddHebebrand:
## Summary

`permanentDeleteDevice()` in `apps/web/src/services/deviceActions.ts` was typed `Promise<{ success: boolean }>` and returned `data.data ?? data`, discarding the `agentUninstallSent` / `warning` fields that `DELETE /devices/:id/permanent` returns when the agent could not be reached for remote uninstall. Both call sites awaited it and always showed a hard-coded green success toast, so the operator saw `"<hostname> has been permanently deleted"` while the agent was still installed and running on the endpoint, with Breeze no longer able to reach it to clean up.

As the issue notes, this is likely the **common** case rather than a rare race: permanent delete requires `decommissioned`, and decommissioning force-closes the agent's WS handshake (rejecting reconnects for decommissioned devices), so `isAgentConnected` is false for most devices reaching this route.

## Changes

- Widened `permanentDeleteDevice`'s return type to `{ success: boolean; agentUninstallSent?: boolean; warning?: string }`.
- Branched the toast at both call sites:
  - `apps/web/src/components/devices/DeviceDetailPage.tsx` (~line 429)
  - `apps/web/src/components/devices/DevicesPage.tsx` (~line 933)
  - A `warning` in the response now shows a 10s `warning`-type toast with the API's message; no `warning` still shows the existing success toast.
- Added the new wrapping copy (`"{{hostname}} was deleted, but {{warning}}"`) as i18n keys across every locale the web app ships: en, de-DE, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR. The API-generated `warning` text itself stays English (it isn't localized at the source); only the surrounding sentence is translated.
- Added unit tests pinning the warning branch on both callers (`DeviceDetailPage.permanentDelete.test.tsx`, new `describe` block in `DevicesPage.test.tsx`), plus a companion assertion that the success toast still fires when there's no warning — per the `quiet_failure_ui_needs_e2e_proof` lesson, a hide-on-failure UI is exactly the shape that regresses silently without a pinning test.

## Test plan

- [x] `pnpm exec vitest run src/components/devices/DeviceDetailPage.permanentDelete.test.tsx --pool=threads --maxWorkers=2` — 2 passed
- [x] `pnpm exec vitest run src/components/devices/DevicesPage.test.tsx --pool=threads --maxWorkers=2` — 72 passed
- [x] `pnpm exec vitest run src/components/devices/DeviceDetailPage.scriptAdmission.test.tsx src/lib/i18n/localeParity.test.ts src/lib/i18n/i18n.test.ts --pool=threads --maxWorkers=2` — 96 passed (sibling file + locale-parity/i18n guards)
- [x] `pnpm exec vitest run src/lib/__tests__/no-silent-mutations.test.ts --pool=threads --maxWorkers=2` — 118 passed
- [x] `npx tsc --noEmit` in `apps/web` — clean
- [x] `npx eslint` on all touched files — clean

Closes #4368


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer, pr-test-analyzer, and silent-failure-hunter lenses applied manually against the diff (the toolkit's subagent dispatch was saturated by ~38 concurrent sibling issue-fixer agents running elsewhere on this machine; rather than retry against a known-saturated pool, I applied the same three review rubrics directly to `git diff origin/main...HEAD`).

**Findings:** 0.
- **Correctness:** both call sites branch cleanly on `result.warning` (truthy check), preserve the existing catch-block error handling unchanged, and still run `navigateTo`/`fetchDevices` on both the warning and success paths — the device row still leaves the list either way. The widened return type in `deviceActions.ts` matches the API's actual 200 body from the issue.
- **CLAUDE.md compliance:** these call sites use the pre-existing try/catch + `showToast` pattern already established in both files for this action (not `runAction`); the diff doesn't touch that convention, so no new violation introduced.
- **i18n:** new `permanentlyDeletedWithWarning` key added to `deviceDetailPage` and `devicesPage.toasts` in all 8 shipped locales (en, de-DE, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR), verified JSON-valid and single-occurrence per locale. The API-generated warning text itself intentionally stays English (not localized at the source).
- **Silent-failure risk:** no new swallow path — this change *removes* a silent-failure (a discarded API warning), and doesn't touch the existing error branches.
- **Test coverage:** both callers get a real behavioral pin — `DeviceDetailPage.permanentDelete.test.tsx` (new file) and a new `describe` block in `DevicesPage.test.tsx`, each asserting a `warning`-type toast (not `success`) when the API response carries a warning, and that `success` still fires when it doesn't. Confirmed these fail if the branch is reverted (caught and fixed a fake-timer install-order bug during development — timers must be armed before the click that schedules the 5s undo-window `setTimeout`, not after). Minor untested edge case: an empty-string `warning` would fall through to the success branch; low severity, not expected from the API.

**Tests:** `apps/web` affected suites, single-fork (`--pool=threads --maxWorkers=2`):
- `DeviceDetailPage.permanentDelete.test.tsx` — 2/2 passed
- `DevicesPage.test.tsx` — 72/72 passed
- `DeviceDetailPage.scriptAdmission.test.tsx` + `lib/i18n/localeParity.test.ts` + `lib/i18n/i18n.test.ts` — 96/96 passed
- `lib/__tests__/no-silent-mutations.test.ts` — 118/118 passed
- `npx tsc --noEmit` — clean
- `npx eslint` on all touched files — clean

PR #4395, opened 2026-09-01
fix(api): escape the request context in the scriptDispatch actor probe (#4299)

### LINKED ISSUE #4299: [API] scriptDispatch actor probe runs withSystemDbAccessContext without escaping the request context — partner-level humans silently lose script attribution
Follow-up from #3978 / PR #4292, deliberately not bundled there.

`apps/api/src/services/scriptDispatch.ts:355-368` holds a third, independently-written copy of the "is this actor a real `users` row?" probe:

```ts
const actorCandidateId = input.createdBy ?? input.triggeredBy ?? null;
const actorIsRealUser = actorCandidateId
  ? await withSystemDbAccessContext(async () => {
      const [userRow] = await db.select({ id: users.id }).from(users)
        .where(eq(users.id, actorCandidateId)).limit(1);
      return Boolean(userRow);
    })
  : true;
```

**It calls `withSystemDbAccessContext` without first escaping the caller's context.** Verified on `origin/main`: `runOutsideDbContext` appears **zero** times in this file.

`withDbAccessContext` short-circuits when a store already exists, so when this runs inside a request-scoped, org-scoped context the "system" probe is not actually system-scoped — it inherits the caller's scope. A partner-level human (`org_id IS NULL`) probed from inside an org-scoped context is then invisible to the SELECT, `actorIsRealUser` comes back `false`, and a **real human's attribution is silently degraded to NULL** (with the id parked in `degradedActorId` and the `parameters` sidecar instead of the FK column).

PR #4292 fixed exactly this class at the `queueCommand` site, and its test analyst confirmed empirically that the partner-level-human case is the *only* test that fails against a `withSystemDbAccessContext`-without-`runOutsideDbContext` fix — i.e. this is the discriminating case, and nothing here currently covers it.

### Severity

Lower than #3978's: this path is FK-safe, so it does **not** produce the 23503 crash that issue reported. The damage is quieter — script/automation executions attributed to nobody when a partner-level human triggers them from an org-scoped request. Wrong attribution on an audit-adjacent record, with no error.

### Why it wasn't folded into PR #4292

Its dual-column semantics differ: it probes one id to settle both `triggeredBy` and `createdBy`, and it has a `degradedActorId` sidecar concept `queueCommand` has no equivalent for. Reusing #4292's helper verbatim would change behaviour beyond the context fix.

### Suggested scope

1. Wrap the probe in `runOutsideDbContext` before `withSystemDbAccessContext`, per the documented rule in CLAUDE.md ("call `runOutsideDbContext` first if inside a request").
2. Add the discriminating test: a partner-level human (`org_id IS NULL`) inside an org-scoped context must keep real attribution on both columns. An agent-only or single-org suite passes either way — that is the vacuous version.
3. While there, consider whether these three probe copies (`commandQueue.ts`, `scriptDispatch.ts`, and any other) should share one helper. Three independent implementations of the same tenancy-sensitive check is how the first two drifted.

🤖 Filed by Claude Code from the #3978 work; the missing `runOutsideDbContext` verified by grep against `origin/main`.

ToddHebebrand:
Closes #4299

## What was wrong

`apps/api/src/services/scriptDispatch.ts` held a third, independently-written copy of the "is this actor a real `users` row?" probe, and it called `withSystemDbAccessContext` **without first escaping the caller's context**. `runOutsideDbContext` appeared zero times in the file.

`withDbAccessContext` short-circuits when a store already exists, so inside a request-scoped, org-scoped context the "system" probe was not system-scoped at all — it inherited the caller's scope. From the org-scoped, user-less context `dbAccessContextFromAuth` builds for an `ai_agent` principal, a partner-level human (`users.org_id IS NULL`) matches **no** branch of the RLS SELECT policy on `users`:

- `breeze_has_partner_access(partner_id)` — partner access is not granted to an org-scoped caller
- `org_id IS NOT NULL AND breeze_has_org_access(org_id)` — skipped on a NULL `org_id`
- `id = breeze_current_user_id()` — null, the context forces `userId: null`

The probe read zero rows, `actorIsRealUser` came back `false`, and a **real human's attribution was silently degraded to NULL** on both `script_executions.triggered_by` and `device_commands.created_by`, with the id parked in the `$actor` sidecar.

Unlike #3978, this path is FK-safe — there is no 23503 crash. The damage is quieter: script/automation executions attributed to nobody, on an audit-adjacent record, with no error anywhere.

## The fix

One function. Wrap the probe in `runOutsideDbContext` before `withSystemDbAccessContext`, per the documented rule in CLAUDE.md ("call `runOutsideDbContext` first if inside a request"). This mirrors the fix shipped for `resolveCommandCreatedBy` in `commandQueue.ts` (PR #4292).

The directly-imported `runOutsideDbContext` is used, never `db.runOutsideDbContext` — the `db` proxy delegates property lookups to the active transaction, which has no such method (the trap `commandQueue.ts` documents at its `runOutsideDbContextSafe` alias).

Everything else is unchanged: same single probe settling both columns, same degrade semantics, same `degradedActorId` / `$actor` sidecar. No behaviour beyond the context escape moved.

## Test

Added `runs the actor probe OUTSIDE the caller DB context, in a system context` to `scriptDispatch.test.ts` — a nesting-order assertion that the probe opens `runOutsideDbContext`, then `withSystemDbAccessContext`, and lands the `users` read between them, plus an assertion that a real user still survives on both columns.

**Verified red before the fix**, with exactly the regression shape:

```
- "enter-outside",
    "enter-system",
    "users-probe",
    "exit-system",
- "exit-outside",
```

## Scope note (issue item 3, deliberately not done here)

The issue suggested considering whether the three probe copies should converge on one helper. Not done — `commandQueue.ts`'s `resolveCommandCreatedBy` returns a single resolved id, while this site probes one id to settle **two** columns and carries a `degradedActorId` sidecar the command site has no equivalent for. Reusing that helper verbatim would change behaviour beyond the context fix, which is exactly what a tenancy-sensitive change should not do. Convergence is worth its own PR.

## Coverage honesty

`db` is mocked in `scriptDispatch.test.ts`, so this suite proves the **escape happens**, not that Postgres RLS then behaves as described — no policy is evaluated. The real-Postgres proof of the identical escape (a partner-level human kept through an org-scoped, user-less caller context) is the load-bearing test in `__tests__/integration/commandQueueCreatedBy.integration.test.ts`, covering the sibling probe. An equivalent integration suite for the dispatch path was **not** added here: it needs the truncating shared test DB, and this machine currently has many agents running against it.

## Verification

- `pnpm exec vitest run` over 14 affected files (`scriptDispatch`, `scriptExecution`, `sourcedParameters`, all `automationRuntime.*`, `aiToolsScripts.runScript.orgEquality`, `fleetFindings/dispatch`, `commandQueue`, `agentEditionAutoMigrate`) — **432 passed**. The wider sweep matters because adding the `runOutsideDbContext` import breaks any sibling suite whose `../db` mock omits it; none did.
- `eslint` on both touched files — clean.

### DIFF
--- a/apps/api/src/services/scriptDispatch.test.ts
+++ b/apps/api/src/services/scriptDispatch.test.ts
@@ -3,6 +3,7 @@ import { SQL, Param } from 'drizzle-orm';
 
 vi.mock('../db', () => ({
   db: { select: vi.fn(), insert: vi.fn(), update: vi.fn(), delete: vi.fn() },
+  runOutsideDbContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
   withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
 }));
 vi.mock('./commandQueue', () => ({ queueCommand: vi.fn() }));
@@ -484,6 +485,72 @@ describe('dispatchScriptToDevice — users-FK probe-and-degrade (#3826)', () => 
     expect(db.insert).not.toHaveBeenCalled();
     const [, , , userId] = vi.mocked(queueCommand).mock.calls[0]!;
     expect(userId).toBeUndefined();
+  });
+
+  // #4299: the load-bearing half of the probe. `users` is an RLS-forced
+  // dual-axis table and `withSystemDbAccessContext` short-circuits to the
+  // caller's store when one is already open (`withDbAccessContext` returns
+  // early on an existing store), so a probe that does not FIRST escape the
+  // request context runs org-scoped. A partner-level human (`users.org_id IS
+  // NULL`) matches no branch of the users SELECT policy from an org-scoped,
+  // user-less context, so the probe reads zero rows and degrades a REAL human
+  // to NULL on both `triggered_by` and `created_by` — silently, since this
+  // path is FK-safe. Nesting order is the observable proof this suite can
+  // make: `db` is mocked here, so no RLS policy is actually evaluated. The
+  // real-Postgres proof of the identical escape (a partner-level human kept
+  // through an org-scoped, user-less caller context) is the load-bearing test
+  // in __tests__/integration/commandQueueCreatedBy.integration.test.ts, which
+  // covers the sibling probe in commandQueue.ts. This assertion fails against
+  // the pre-#4299 code — verified, not assumed.
+  it('runs the actor probe OUTSIDE the caller DB context, in a system context', async () => {
+    const dbModule = await import('../db');
+    const order: string[] = [];
+    vi.mocked(dbModule.runOutsideDbContext).mockImplementationOnce(async (fn: () => unknown) => {
+      order.push('enter-outside');
+      const result = await fn();
+      order.push('exit-outside');
+      return result;
+    });
+    vi.mocked(dbModule.withSystemDbAccessContext).mockImplementationOnce(async (fn: () => unknown) => {
+      order.push('enter-system');
+      const result = await fn();
+      order.push('exit-system');
+      return result;
+    });
+    vi.mocked(db.select).mockReturnValueOnce({
+      from: vi.fn().mockReturnValue({
+        where: vi.fn().mockReturnValue({
+          limit: vi.fn().mockImplementation(() => {
+            order.push('users-probe');
+            return Promise.resolve([{ id: 'probed' }]);
+          }),
+        }),
+      }),
+    } as any);
+
+    const r = await dispatchScriptToDevice({
+      device: device(),
+      source: { kind: 'saved', script: savedScript() },
+      triggeredBy: 'user-1',
+      createdBy: 'user-1',
+    });
+
+    expect(r.ok).toBe(true);
+    // Nesting matters: outside must open before system, and the read must land
+    // between them. `withSystemDbAccessContext` alone (no escape) yields
+    // ['enter-system', 'users-probe', 'exit-system'] and is the regression.
+    expect(order).toEqual([
+      'enter-outside',
+      'enter-system',
+      'users-probe',
+      'exit-system',
+      'exit-outside',
+    ]);
+    // And the real user still survives the guard on BOTH columns.
+    const execValues = vi.mocked(db.insert).mock.results[0]!.value.values.mock.calls[0]![0];
+    expect(execValues.triggeredBy).toBe('user-1');
+    const [, , , userId] = vi.mocked(queueCommand).mock.calls[0]!;
+    expect(userId).toBe('user-1');
   });
 
   it('does not add an $actor key when the script has no bound parameters and the actor IS real', async () => {
--- a/apps/api/src/services/scriptDispatch.ts
+++ b/apps/api/src/services/scriptDispatch.ts
@@ -2,7 +2,7 @@ import { randomUUID } from 'node:crypto';
 import { and, eq } from 'drizzle-orm';
 import { canonicalizeScriptParameters, hasVariableTokens } from '@breeze/shared';
 
-import { db, withSystemDbAccessContext } from '../db';
+import { db, runOutsideDbContext, withSystemDbAccessContext } from '../db';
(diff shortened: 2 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` could **not** dispatch — the machine-wide subagent pool was saturated (20/20 in use by sibling issue-fixers), so `code-reviewer`, `silent-failure-hunter` and `pr-test-analyzer` never ran. Reviewed inline instead across four lenses: context-escape correctness, test discriminating-power, import-blast-radius, and tenancy. Stating that plainly so the label matches what actually executed.

**Findings:** 4 raised, all resolved without a code change; 0 outstanding.

1. *Does escaping the caller context break the surrounding dispatch?* — No. `runOutsideDbContext` is `dbContextStorage.exit(() => dbContextMetaStorage.exit(fn))`; the exit is scoped to the callback, so the `script_executions` insert and the `queueCommand` call that follow still run in the caller's context. The escaped read is a read-only PK lookup on `users` returning a boolean, so nothing crosses the transaction boundary.
2. *Could the new test pass against a broken implementation?* — No. Both wrappers are instrumented with `mockImplementationOnce`, so the assertion is on the full nesting sequence, not on call counts. Dropping `withSystemDbAccessContext` yields `['enter-outside','users-probe','exit-outside']`; a `runOutsideDbContext` call that does not *wrap* the probe yields `['enter-outside','exit-outside','enter-system',…]`. Both fail. The pre-existing `single probe covers BOTH columns` test proves the probe is the only `db.select` on this path, so the recorded `users-probe` marker cannot be a different query.
3. *Import blast radius.* Adding `runOutsideDbContext` to the `../db` import breaks any suite whose `vi.mock('../db')` factory omits it **and** that actually executes the probe. Enumerated every test file referencing `scriptDispatch`: 10 mock `./scriptDispatch` outright, 2 reference it only in comments (`scriptSecretDelivery.test.ts`, `workerEntrypointClosure.contract.test.ts`), 1 is the integration suite that uses the real helper. All were run anyway — see below. None needed a mock change.
4. *Tenancy: does the escape widen an existence oracle?* — This was the one finding worth chasing. Post-fix the probe can observe a `users` row in **any** tenant, where the org-scoped version could not, so a caller-supplied uuid would leak "this id is a real user" via the presence/absence of the `$actor` sidecar. Checked all five production call sites: `aiToolsScripts.ts:450` and `scriptExecution.ts:252` pass `auth.user.id`, `automationRuntime.ts:1216/1319` pass the stored `automations.created_by`, `agentEditionAutoMigrate.ts:300` passes `null`. No request-body value reaches the probe, so there is no oracle. Not a defect — recorded because a future call site that accepts a client-supplied actor id would create one.

**Tests:** 16 affected suites green single-run (`vitest run --pool=threads --maxWorkers=2`), **596 passed / 0 failed** — `scriptDispatch`, `scriptExecution`, `scriptExecution.admission`, `sourcedParameters`, `scriptSecretDelivery`, `workerEntrypointClosure.contract`, all five `automationRuntime.*`, `aiToolsScripts.runScript.orgEquality`, `fleetFindings/dispatch`, `commandQueue`, `agentEditionAutoMigrate`. The new test was **verified red against the pre-fix code** before the fix landed, failing with exactly `['enter-system','users-probe','exit-system']` — the regression shape, not a generic failure. `tsc --noEmit` on `apps/api` clean; `eslint` on both touched files clean.

**Not covered (deliberate):** no real-Postgres suite for this path — `db` is mocked here, so this proves the escape happens, not that RLS then behaves as described. The real-RLS proof for the identical bug class is the load-bearing test in `commandQueueCreatedBy.integration.test.ts` (#4292) against the sibling probe. An equivalent integration suite here would need the shared truncating test DB, which many agents are running against on this machine right now.

**Status:** review-clean, awaiting maintainer merge. Issue #4299 left open and assigned.

Issue #3919, opened 2026-08-24
[Mobile] Bulk acknowledge flush is not durable: the batch leaves the undo ref before transport dispatch, so process suspension mid-await loses it


ToddHebebrand:
Split out of #3727 (per the maintainer ruling there) — identified by @bdunncompany during that PR's adversarial review.

**The window:** on `background`, `flushAllUndo` empties the undo ref and hands the batch to `dispatchAcknowledge`, which awaits SecureStore token retrieval, the CSRF token, the installation id, and the server URL **before** it reaches `fetch`. If iOS suspends or terminates the process during those awaits, the batch is gone: the timer cannot resend, unmount cannot resend, and both correctly see an empty ref. This is the same stranded-in-memory class #3727 set out to close, relocated from "unsent tail of a worker queue" to "whole batch taken before transport dispatch".

**What #3727 ships instead (accepted residual):** pre-warmed credentials to shrink the pre-dispatch window (option 3 from the PR thread). The residual failure mode is honest — the server never acknowledged, so the rows un-hide on the next authoritative fetch and the operator re-acks. The UI never claims an acknowledge that did not happen.

**The durable fix (this issue):** a persisted outbox — write the batch to durable storage *before* releasing the undo ref, clear it only on a confirmed outcome, replay on next launch. Design points that make this its own change rather than a PR-seven addition to #3727:

- **Replay dedupe is mandatory.** Acknowledging is irreversible and there is no unacknowledge route, so a replay must not re-fire on ids the server already processed. The bulk response's `updatedIds`/`skippedIds`/`failedIds` arrays (added in #3727) are the confirmation signal to clear against.
- **Replay staleness.** A batch replayed at next launch may be hours old; an alert that has since escalated or resolved should probably not be silently acknowledged. Needs an expiry or a re-validate-before-replay rule.
- **Interaction with the `unknown` bucket** (#3727): ambiguous-timeout ids are deliberately not restored and are released by the next authoritative fetch; the outbox must not resurrect them.

Related: #3782 (post-acknowledge un-hide needs a fetch generation, not a refresh() gate) — same surface, different invariant.

PR #4396, opened 2026-09-01
fix(mobile): gate ack un-hide on activeAlerts fetch generation (#3782)

### LINKED ISSUE #3782: [Mobile] Post-acknowledge un-hide releases rows without proof of a fresh alerts snapshot — needs a fetch generation, not a refresh() gate
Split out of #3727 so that PR isn't held on a data-layer change. Raised and diagnosed by @bdunncompany, who built the obvious fix, found it was worse than the bug, and backed it out rather than shipping it.

**Current behaviour.** `acknowledgeIds` hides the acknowledged rows optimistically, fires the request, then calls `refresh()` and un-hides via `endAck` unconditionally. If the refetch didn't actually land, the rows un-hide against stale data — an acknowledged alert flickers back as unacknowledged, or an alert that failed to acknowledge presents as handled.

**Why the obvious fix is wrong.** Gating the un-hide on `refresh()`'s success (`if (await refresh()) endAck(...)`) produces a strictly worse failure:

- `fetchAll` opens with `if (inFlight.current) return`. An acknowledge issued while any other refresh is in flight — push handler, realtime WS, tab-focus, pull-to-refresh — coalesces into that one and the gate reports no fresh read.
- Nothing else releases those ids. `endAck` for the acknowledged set only runs inside that `acknowledgeIds` call, so the rows stay hidden for the life of the screen.
- A briefly stale visible row is recoverable. A permanently concealed active alert is not, and this list is how the operator learns the alert exists.

There is a second hole in the same patch: `failedCount < 5` means "at least one of five slices arrived", not "alerts arrived". It returns true when devices and orgs land and `alerts` itself failed — un-hiding against stale alert data, which is the original bug with extra steps.

**The actual fix.** `refresh()` resolving is not evidence of freshness, and neither is its return value. The primitive that is: a monotonic fetch generation stamped on each *alerts* snapshot, which `pendingAcks` records at ack time and compares against. A pending ack releases when a newer alerts snapshot lands, regardless of which caller fetched it — so coalescing is harmless (the coalesced-into fetch still bumps the generation) and a partial failure that didn't include `alerts` doesn't release anything.

That is a change in the data layer (`useSystemsData` / `pendingAcks`), not at the `acknowledgeIds` call site, which is why it isn't in #3727.

Both call sites in #3727 are commented with this reasoning so the obvious patch doesn't get re-attempted.

ToddHebebrand:
## Summary

Split out of #3727 per that issue's request (data-layer change, kept off the call-site PR).

`dispatchAcknowledge` in `SystemsScreen.tsx` hid acknowledged rows optimistically, then un-hid them unconditionally after `await refresh()`. `refresh()` resolving is not proof of freshness:

- `fetchAll` coalesces into any already-in-flight request (`if (inFlight.current) return false`), so an ack's own refresh call can report nothing new even though it "resolved."
- Nothing else released those ids, so a coalesced refresh left an acknowledged (or attempted-and-failed) alert permanently hidden for the life of the screen.
- The tempting fix — gate the un-hide on `refresh()`'s return value — is strictly worse for the reasons above, and is explicitly called out as wrong in the existing code comments (this issue was split out precisely so that patch wouldn't get re-attempted).

## Fix

- `useSystemsData.ts`: adds a monotonic `activeAlertsGeneration` counter, bumped whenever the `activeAlerts` slice (not `alerts` — that's the slice `activeIssues`, and therefore every row `pendingAcks` hides, is actually built from) resolves fulfilled inside `fetchAll`, regardless of which caller triggered the fetch. Exposed both as state (`activeAlertsGeneration`, for a consumer to react to changes) and via a ref-backed `getActiveAlertsGeneration()` getter (for reading the *current* value synchronously, since acknowledging can take 13-15s and a value captured in a stale closure would be wrong).
- `pendingAcks.ts`: adds `markAckConfirmed(state, ids, generation)` (records the generation observed when the server confirmed the ack, without releasing anything) and `releaseStaleAcks(state, generation)` (releases ids whose recorded generation has been superseded by a strictly newer one). `beginAck`/`endAck` keep their existing reference-counting semantics for the failure/undo/unmount paths, which don't need to wait on a fetch.
- `SystemsScreen.tsx`: `dispatchAcknowledge` now stamps confirmed + unknown-outcome ids with the generation read via `getActiveAlertsGeneration()` right after the server call resolves, then fires `refresh()` without awaiting its result. A new effect watches `activeAlertsGeneration` and releases any pending ack whose bar has been passed — so release happens off *any* fetch that lands a fresh `activeAlerts` snapshot, coalesced or not.

Does **not** touch the bulk-ack undo durability work (#3919) — out of scope per the issue.

## Test plan

- [x] `pendingAcks.test.ts` — 8 new cases covering: stale/same-generation fetch does not release; a strictly newer generation does; a fetch landing before ack confirmation is even recorded releases nothing; overlapping acks on the same id merge to the stricter (later) bar and release conservatively; no-ops on unknown ids. 18/18 passing.
- [x] Existing `pendingAcks.test.ts`, `mergeSystemsResults.test.ts`, `undoAck.test.ts` still green (45/45) — reference-counting and merge behavior unchanged.
- [x] `npx tsc --noEmit -p apps/mobile` clean.

```
pnpm exec vitest run src/screens/systems/pendingAcks.test.ts src/screens/systems/mergeSystemsResults.test.ts src/screens/systems/undoAck.test.ts --pool=threads --maxWorkers=2
# Test Files  3 passed (3) | Tests  45 passed (45)
```

Closes #3782

### DIFF
--- a/apps/mobile/src/screens/systems/SystemsScreen.tsx
+++ b/apps/mobile/src/screens/systems/SystemsScreen.tsx
@@ -66,6 +66,17 @@ type Nav = NativeStackNavigationProp<SystemsStackParamList, 'Systems'>;
  */
 const UNDO_WINDOW_MS = 5000;
 
+/**
+ * Bounded retry schedule for the post-ack refresh that satisfies the
+ * fetch-generation release condition (#3782). A single `refresh()` call can
+ * itself fail to land a fresh `activeAlerts` snapshot (transient rejection,
+ * or coalescing into an in-flight fetch that also rejects) — this retries a
+ * few times before giving up and deferring to the ambient triggers
+ * (push/WS/focus/pull), rather than leaving a confirmed-acknowledged row
+ * hidden with no time bound.
+ */
+const ACK_REFRESH_RETRY_DELAYS_MS = [0, 2000, 5000];
+
 // Inline magnifying glass — see SearchSheet for the input-decorating sibling.
 // 16px sizing here matches the right-edge of the Hero copy block.
 function HeaderSearchIcon({ color, size = 16 }: { color: string; size?: number }) {
@@ -286,12 +297,28 @@ export function SystemsScreen() {
         setPendingAcks((p) =>
           markAckConfirmed(p, [...acknowledged, ...unknown], generationAtConfirm)
         );
-        // Kick a refresh so the release condition above gets satisfied
-        // promptly. Its return value is deliberately unused for un-hiding —
-        // that happens automatically via the generation-watching effect,
-        // whichever fetch actually lands the fresh snapshot, coalesced or
-        // not, this one included.
-        void refresh();
+        // Kick a refresh so the release condition above gets satisfied.
+        // Un-hiding itself happens automatically via the generation-watching
+        // effect, off whichever fetch actually lands the fresh snapshot,
+        // coalesced or not — this call's return value is never used for
+        // that. But THIS specific fetch can itself fail to advance the
+        // generation (a transient rejection on `activeAlerts`, or coalescing
+        // into an in-flight fetch that also rejects it), and nothing else is
+        // guaranteed to retry soon: WS/push only fire on unrelated activity,
+        // and focus-refresh is behind a 60s debounce. Left unbounded, that
+        // is a permanently concealed active alert — the exact outcome #3782
+        // exists to prevent, just moved one step later. So retry with a
+        // short bounded backoff until the generation clears the bar just
+        // recorded, then give up and defer to those ambient triggers.
+        void (async () => {
+          for (const delayMs of ACK_REFRESH_RETRY_DELAYS_MS) {
+            if (delayMs > 0) {
+              await new Promise((resolve) => setTimeout(resolve, delayMs));
+            }
+            await refresh();
+            if (getActiveAlertsGeneration() > generationAtConfirm) return;
+          }
+        })();
         return;
       } catch (err) {
         reportInternalError(err, 'bulk-acknowledge');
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer agent (a second agent's concurrent review was throttled by the machine's parallel-agent capacity; the code-reviewer pass is the one that landed).

**Findings:** 1 Critical, 1 Important, 3 Suggestions raised → addressed in c898150ae:
- **Critical** — the post-ack `refresh()` that satisfies the release condition could itself fail to land a fresh `activeAlerts` snapshot (transient rejection, or coalescing into an in-flight fetch that also rejects), with no other trigger guaranteed to retry soon (WS/push only fire on unrelated activity; focus-refresh is behind a 60s debounce) — leaving a confirmed-acknowledged row hidden with no time bound. **Fixed:** bounded retry schedule (`ACK_REFRESH_RETRY_DELAYS_MS`, 3 attempts over ~7s) in `dispatchAcknowledge`.
- **Important** — a coalesced fetch's response can technically predate the exact ack-commit instant in rare cases, so "strictly newer generation" isn't a perfect causal proof. Clarified in `markAckConfirmed`'s doc comment: the actual guarantee is "at least one fetch has landed since confirmation," bounding staleness to one residual cycle rather than proving perfect ordering — consistent with the issue's own stated tradeoff ("briefly stale is recoverable, permanently concealed is not"). No behavior change; this was already the issue's explicit design intent for the coalescing case.
- **Suggestions** — `releaseStaleAcks` now releases a pending id's whole entry once its merged bar is passed, instead of draining the reference count one generation-tick per call (since `markAckConfirmed` already merges overlapping same-id operations to a single, stricter bar via `Math.max`, clearing it is sufficient proof for all of them at once). Fixed a stale comment in `flushHeldAcknowledgesMounted` describing the old refresh()-gated release. Not addressed: no hook-level test exercises `useSystemsData`'s generation-bump condition directly (fulfilled bumps / rejected doesn't) — the codebase has no existing `renderHook`-based test infra for any hook, and the bump condition is a single obvious conditional; covered indirectly by the `pendingAcks.ts` pure-function tests, which is where the issue's stated test requirement ("stale fetch resolving after ack does not un-hide; fresh fetch does") lives instead.

**Tests:** `apps/mobile/src/screens/systems/pendingAcks.test.ts` (18 tests, including 8 new covering: same-generation does not release; strictly-newer releases; unconfirmed ack never auto-releases; overlapping same-id acks merge to the stricter bar and release together; no-ops on unknown ids) + `mergeSystemsResults.test.ts` + `undoAck.test.ts` — 45/45 passing (`pnpm exec vitest run src/screens/systems/pendingAcks.test.ts src/screens/systems/mergeSystemsResults.test.ts src/screens/systems/undoAck.test.ts --pool=threads --maxWorkers=2`). `npx tsc --noEmit -p apps/mobile` clean.

PR #4397, opened 2026-09-01
fix(web): stop a failed quiet refetch from stranding new quote lines (#4286)

### LINKED ISSUE #4286: [Web][Quotes] Line creation has the same silent-strand defect as image blocks — a failed quiet refetch duplicates billable lines
Follow-up from #3519 / PR #4284, which fixed the same defect for quote **image blocks**. Scoped out there deliberately: it needs its own copy, 8 locales and tests, and folding it in would have doubled a diff that is currently provable end to end.

### The shared premise

Several quote-editor mutations deliberately have **no success toast**, on the reasoning that the newly appended row *is* the feedback. That is a legitimate UX choice, but it makes the visible result the delivery channel — and the canvas only moves when `QuoteWorkspace`'s **quiet** refetch lands. Before PR #4284, `fetchDetail` swallowed every quiet failure, justified in-code as "the inline action's own `runAction` toast already surfaced the failure." That is sound for a failed mutation and **false for the refresh after a success**, where `runAction` toasted nothing on purpose.

Net effect: the write succeeds server-side, the refresh fails, the operator sees nothing, retries, and duplicates the row.

### Call sites still carrying it

- `doAddCatalog`
- `addManualLine`
- the Pax8 / distributor import paths

All three end in a fire-and-forget `refresh()`. A technician on a flaky link can re-add the same line exactly as the reporter of #3519 re-uploaded the same image — except a duplicated billable line is a worse outcome than a duplicated image block.

PR #4284 documents this at the call site so it cannot be lost, but does not fix it.

### Suggested fix

Apply the same shape PR #4284 established for image blocks:
1. Surface quiet-refetch failure to the operator rather than swallowing it, so the "the row appearing is the signal" premise stops resting on a channel that can fail silently.
2. Carry over the **request-sequencing guard** too. #4284 found a second vector in the same function: overlapping quiet reloads with no sequencing, so an *older* GET can resolve last and re-render pre-mutation data. Note that `runScoped` only serialises calls sharing a key — a terms blur-save and an add-line use different keys, so they genuinely overlap.
3. New user-facing strings need all 8 locales.

Worth reading PR #4284's tests first: it proves the fix red-first four times, and it removed one suggested assertion (checking the file input's `.value` is cleared) after showing it passes with or without the fix, because jsdom never populates `value` from `files`. The equivalent trap likely exists on the line-creation path.

🤖 Filed by Claude Code from the #3519 work.

ToddHebebrand:
## Summary

Follow-up from #3519 / PR #4284, which fixed the same defect for quote **image blocks**. This closes the deliberately-scoped-out twin for **line creation**.

`doAddCatalog` and `addManualLine` in `QuoteEditor.tsx` ended in a fire-and-forget `refresh()` after a successful line POST. When the quiet refetch failed, the write had already succeeded server-side but the canvas never moved and no toast fired — these mutations deliberately have no success toast of their own ("the appended row and moving totals ARE the feedback"). A technician on a flaky link reads that silence as "nothing happened" and re-adds the same line, duplicating a billable charge — a worse outcome than #3519's duplicated image block.

## Changes

- Ports the exact pattern PR #4284 established for image blocks: added a `finishLineCreate()` helper that routes the post-create tail through the one-shot `resync()` (not the coalescing `refresh()` — adding a line is a one-shot action, not the tab-through burst the throttle exists to cap) and warns the user honestly (`lineAddedListStale` toast) when the canvas didn't catch up.
- Both call sites (`doAddCatalog` — which also covers the Pax8/distributor import paths that route through it — and `addManual`) now call `finishLineCreate()` instead of the bare `refresh()`.
- `QuoteWorkspace.fetchDetail` already carries the success/failure return value and the request-sequencing guard from #4284, so no `QuoteWorkspace.tsx` changes are needed — this PR is a pure `QuoteEditor.tsx` + locale change.
- New `quotes.editor.errors.lineAddedListStale` copy added to all 8 web locales (de-DE, en, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR), following the exact phrasing pattern of #4284's `sectionAddedListStale`.

## Tests

New `QuoteEditor.linecreatestrand.test.tsx` (7 tests, mirrors `QuoteEditor.blockcreatestrand.test.tsx`'s structure) covering both call sites:
- catalog add: warns (not "could not add") when the create lands but the resync returns `false`
- catalog add: same warning (not a generic add-failure) when the resync throws outright
- catalog add: a genuine create failure still reports the real error, never "added" (guards against over-reporting)
- catalog add: stays quiet on the happy path
- manual add: warns when the create lands but the resync fails
- manual add: a genuine create failure still reports the real error
- manual add: stays quiet on the happy path

Verified red-first: stashed the `QuoteEditor.tsx` fix and confirmed exactly the 3 "warns on failed resync" tests go red (the 4 failure/happy-path tests correctly stay green, since those paths were never broken) — then restored the fix and confirmed all 7 pass.

## Verification

- `pnpm exec vitest run src/components/billing/quotes/QuoteEditor.linecreatestrand.test.tsx --pool=threads --maxWorkers=2` — 7/7 passed
- `pnpm exec vitest run src/components/billing/quotes/QuoteEditor.test.tsx src/components/billing/quotes/QuoteEditor.costmarkup.test.tsx src/components/billing/quotes/QuoteEditor.blockcreatestrand.test.tsx src/components/billing/quotes/QuoteWorkspace.test.tsx src/components/billing/quotes/QuoteEditor.distributor.test.tsx --pool=threads --maxWorkers=2` — 33/33 passed (no regressions in sibling suites)
- `npx tsc --noEmit` in `apps/web` — clean, exit 0
- `npx eslint` on both changed files — clean
- All 8 locale JSON files validated with `JSON.parse`

Closes #4286

### DIFF
--- a/apps/web/src/components/billing/quotes/QuoteEditor.linecreatestrand.test.tsx
+++ b/apps/web/src/components/billing/quotes/QuoteEditor.linecreatestrand.test.tsx
@@ -24,6 +24,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import QuoteEditor from './QuoteEditor';
 import type { QuoteDetail as QuoteDetailData } from './quoteTypes';
 import { addCatalogLine, addManualLine } from '../../../lib/api/quotes';
+import { createCatalogItem } from '../../../lib/api/catalog';
 
 vi.mock('../../../stores/auth', () => ({
   registerOrgIdProvider: vi.fn(),
@@ -101,6 +102,7 @@ const detail: QuoteDetailData = {
 
 const addCatalogLineMock = vi.mocked(addCatalogLine);
 const addManualLineMock = vi.mocked(addManualLine);
+const createCatalogItemMock = vi.mocked(createCatalogItem);
 
 const toastMessages = () =>
   showToast.mock.calls.map((c) => (c[0] as { message: string }).message);
@@ -137,10 +139,14 @@ describe('QuoteEditor — a catalog/manual line created server-side is never sil
     await submitCatalogLine();
 
     await waitFor(() => expect(addCatalogLineMock).toHaveBeenCalledTimes(1));
-    await waitFor(() => expect(showToast).toHaveBeenCalled());
-    expect(toastMessages().join(' | ')).toMatch(/added/i);
-    // Must not be the copy that invites a re-submit.
-    expect(toastMessages().join(' | ')).not.toMatch(/could not add/i);
+    // Assert the exact resolved copy, not just /added/i — the block-strand
+    // fix's `sectionAddedListStale` string also matches that loose regex, and
+    // a copy-paste that toasted the wrong locale key would still pass a loose
+    // "contains added" check (verified: swapping in `sectionAddedListStale`
+    // here left the loose assertion green).
+    await waitFor(() => expect(showToast).toHaveBeenCalledWith(
+      expect.objectContaining({ message: 'Line added, but the list could not refresh. Reload the page to see it.', type: 'warning' }),
+    ));
   });
 
   it('catalog add: still reports the stale-list warning (not a generic add failure) when the resync throws outright', async () => {
@@ -151,10 +157,9 @@ describe('QuoteEditor — a catalog/manual line created server-side is never sil
     await submitCatalogLine();
 
     await waitFor(() => expect(addCatalogLineMock).toHaveBeenCalledTimes(1));
-    await waitFor(() => expect(showToast).toHaveBeenCalled());
-    const joined = toastMessages().join(' | ');
-    expect(joined).toMatch(/added/i);
-    expect(joined).not.toMatch(/could not add the catalog item/i);
+    await waitFor(() => expect(showToast).toHaveBeenCalledWith(
+      expect.objectContaining({ message: 'Line added, but the list could not refresh. Reload the page to see it.', type: 'warning' }),
+    ));
   });
 
   it('catalog add: a genuine create failure still reports the real error, not "added"', async () => {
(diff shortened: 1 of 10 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer and pr-test-analyzer (silent-failure-hunter could not be scheduled twice due to a concurrent-subagent limit from ~38 sibling issue-fixers running on this machine; its concerns were substantively covered by code-reviewer's explicit "no exception escape / no swallowed toast" verification below).

**Findings:** 2 important raised by code-reviewer → both addressed in `1989d7c94`; 0 outstanding.
1. Test assertions used a loose `/added/i` regex that also matches the block-strand fix's "Section added…" copy, so a copy-pasted wrong locale key would have passed silently. Verified by mutation (swap in `sectionAddedListStale`): before the fix, 0/9 tests caught it; after tightening to assert the exact `lineAddedListStale` string, 4/9 correctly go red. Tightened all three "warns on failed resync" assertions to exact-string `toHaveBeenCalledWith`.
2. `addManual`'s optional "save to catalog" follow-up sat outside a `try/finally`, so a failing `createCatalogItem` call threw past the `finishLineCreate()` tail entirely — the manual line already existed server-side, but the canvas was left completely un-resynced with no stale-list warning (worse than the `refresh()` this PR replaces). Wrapped the optional catalog-save in `try/finally` so the resync always runs once the line is known to exist; added a dedicated regression test.

pr-test-analyzer found 0 critical/important gaps; its 3 low-priority suggestions (manual-add "resync throws" parity test, explicit Pax8/distributor-path test, saveToCatalog+failed-resync interaction test) were all addressed in the same follow-up commit except the Pax8/distributor one, which it rated not-needed (both import paths route through the already-covered `doAddCatalog`, verified by grep — no third line-create call site exists).

**Tests:** `apps/web/src/components/billing/quotes/QuoteEditor.linecreatestrand.test.tsx` — 9/9 passed (single-fork, `--pool=threads --maxWorkers=2`). Sibling suites re-run with no regressions: `QuoteEditor.test.tsx`, `QuoteEditor.costmarkup.test.tsx`, `QuoteEditor.blockcreatestrand.test.tsx`, `QuoteWorkspace.test.tsx`, `QuoteEditor.distributor.test.tsx` — 42/42 passed. `no-silent-mutations.test.ts`, `localeParity.test.ts`, `keyUsage.test.ts` — 201/201 passed. `npx tsc --noEmit` in `apps/web` — clean, exit 0. `npx eslint` on both changed files — clean. Red-first verified twice: (a) stashing the whole `QuoteEditor.tsx` fix reds exactly the 3 original "warns on failed resync" tests; (b) mutating the toast copy to the wrong locale key reds exactly the 4 tests now asserting the exact string.

PR #4398, opened 2026-09-01
fix(api): desktopWs input schema omits key_press, drops paste-as-keystrokes on WS fallback (#4370)

### LINKED ISSUE #4370: [API] desktopWs input schema omits 'key_press' — Viewer paste-as-keystrokes fully dropped on the WebSocket fallback transport
**Found during:** PR #4352 review work on #4089.

**Verified:** `apps/api/src/routes/desktopWs.ts:64` — the input-event zod enum is `['mousemove','mousedown','mouseup','keydown','keyup','wheel','click','dblclick','mouse_move','mouse_down','mouse_up','key_down','key_up']`; `key_press` is not in it.

**Inferred impact (not yet exercised end-to-end):** the Viewer's Paste Text path emits `key_press` events (`apps/viewer/src/lib/paste.ts` `textToKeyEvents`), so on the WebSocket fallback transport every pasted character fails schema validation and is dropped — paste silently does nothing there, independent of the layout-mangling fixed by #4352. Live typed keys (`key_down`/`key_up`) pass.

**Fix shape:** add `key_press` (and the new `type_text` control message from #4352, if the WS fallback should carry it) to the schema, plus a test asserting the enum covers every event type the Viewer can emit.

Status: unassigned; surfaced by automated review, needs a quick end-to-end confirm on the WS fallback before fixing.

ToddHebebrand:
Closes #4370

## The bug

`apps/api/src/routes/desktopWs.ts`'s `desktopInputEvent` zod enum omitted `key_press`, which the Viewer's Paste Text path (`apps/viewer/src/lib/paste.ts` `textToKeyEvents`) emits for every pasted character. On the WebSocket fallback transport (used whenever WebRTC isn't available), any input event that fails schema validation is silently dropped — so pasting into a remote session over the WS fallback did nothing at all, independent of the layout-mangling bug that #4352 already fixed for the WebRTC path.

## The fix

Added `key_press` to the enum. While building the schema-coverage regression test the issue asked for, I found `mouse_scroll` was *also* missing — same root cause, silently dropping every scroll-wheel event on the WS fallback (`DesktopViewer.tsx`'s wheel handler calls `sendInputFn({ type: 'mouse_scroll', ... })` regardless of transport). Both `key_press` and `mouse_scroll` are already in the agent's own accepted-input allowlist (`agent/internal/heartbeat/handlers_desktop.go` `desktopInputTypes`), so this brings the API's WS validation back in line with what the Viewer actually sends and the agent actually accepts.

`desktopInputEvent` is now exported so the new test can import it directly.

## Test

`apps/api/src/routes/desktopWs_inputSchema.test.ts` reads `DesktopViewer.tsx`'s source directly and regex-extracts every `sendInputFn({ type: '<kind>', ... })` call site, then asserts the WS schema's `safeParse` accepts each one — a real coverage test (not hardcoded against the fix) that fails loudly if the Viewer ever emits an event kind this schema doesn't know about. A sanity-floor assertion guards against the extraction itself silently returning zero and making the parametrized checks vacuous, plus one negative test confirming the schema still rejects garbage.

## Verification

- Red-first: reverted the schema change, ran the new test → 8/9 failed. Restored the fix → 9/9 passed.
- `apps/api/src/routes/desktopWs*.test.ts` (8 files, 88 tests): all pass — `pnpm exec vitest run src/routes/desktopWs_inputSchema.test.ts src/routes/desktopWs.test.ts src/routes/desktopWs_onmessage.test.ts src/routes/desktopWs_lifecycle.test.ts src/routes/desktopWs_multitenant.test.ts src/routes/desktopWs_onopen.test.ts src/routes/desktopWs_rate_limit_cleanup.test.ts src/routes/desktopWs_utils_http.test.ts --pool=threads --maxWorkers=2`
- `tsc --noEmit -p apps/api` (matches CI's `typecheck` job step): clean.
- `eslint` on both touched files: clean.

### DIFF
--- a/apps/api/src/routes/desktopWs.ts
+++ b/apps/api/src/routes/desktopWs.ts
@@ -62,9 +62,12 @@ import {
 // Zod validation for desktop user messages.
 // Exported for desktopWs_inputSchema.test.ts, which asserts this enum covers
 // every input kind the Viewer's sendInputFn (apps/viewer/src/components/
-// DesktopViewer.tsx) can emit on the WebSocket fallback transport — the same
-// events the agent's WebRTC-side desktopInputTypes allowlist
-// (agent/internal/heartbeat/handlers_desktop.go) accepts.
+// DesktopViewer.tsx) can emit. Note WebRTC sessions inject input via the
+// user helper's data channel and never reach this schema at all (this route
+// only carries the WebSocket *fallback* transport) — but the agent's own
+// command-relay handler for that fallback path, handleDesktopInput /
+// desktopInputTypes in agent/internal/heartbeat/handlers_desktop.go, accepts
+// the same event kinds, so this enum should stay a superset of that map.
 export const desktopInputEvent = z.object({
   type: z.enum(['mousemove', 'mousedown', 'mouseup', 'keydown', 'keyup', 'wheel', 'click', 'dblclick', 'mouse_move', 'mouse_down', 'mouse_up', 'mouse_scroll', 'key_down', 'key_up', 'key_press']),
   x: z.number().min(-10000).max(100000).optional(),
--- a/apps/api/src/routes/desktopWs.ts
+++ b/apps/api/src/routes/desktopWs.ts
@@ -59,9 +59,14 @@ import {
   type RemoteWsUpgradeContext,
 } from '../services/remoteWsUpgrade';
 
-// Zod validation for desktop user messages
-const desktopInputEvent = z.object({
-  type: z.enum(['mousemove', 'mousedown', 'mouseup', 'keydown', 'keyup', 'wheel', 'click', 'dblclick', 'mouse_move', 'mouse_down', 'mouse_up', 'key_down', 'key_up']),
+// Zod validation for desktop user messages.
+// Exported for desktopWs_inputSchema.test.ts, which asserts this enum covers
+// every input kind the Viewer's sendInputFn (apps/viewer/src/components/
+// DesktopViewer.tsx) can emit on the WebSocket fallback transport — the same
+// events the agent's WebRTC-side desktopInputTypes allowlist
+// (agent/internal/heartbeat/handlers_desktop.go) accepts.
+export const desktopInputEvent = z.object({
+  type: z.enum(['mousemove', 'mousedown', 'mouseup', 'keydown', 'keyup', 'wheel', 'click', 'dblclick', 'mouse_move', 'mouse_down', 'mouse_up', 'mouse_scroll', 'key_down', 'key_up', 'key_press']),
   x: z.number().min(-10000).max(100000).optional(),
   y: z.number().min(-10000).max(100000).optional(),
   button: z.union([z.string().max(20), z.number().int().min(0).max(4)]).optional(),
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer (agent), plus inline test-coverage and comment-accuracy passes (the pool was saturated by ~38 concurrent sibling issue-fixer sessions, so those two lenses ran inline rather than as separate agents).

**Findings:** 3 raised → all addressed in cddd0d91b / 392ebf773; 0 outstanding blocking.
- **Important (code-reviewer):** the source-scraper only matched literal `sendInputFn({ type: '...' })` call sites in `DesktopViewer.tsx`, so it never actually saw the Paste Text path this issue is about — Paste forwards events built in `paste.ts` through an indirect `sendInput: event => sendInputFn({ ...event })` spread. It caught `key_press` only by coincidence, via unrelated live-keystroke call sites; if those ever stopped using `key_press`, the guard would stay green while paste stayed silently broken (the exact bug #4370 reports, unguarded). **Fixed in cddd0d91b:** the scraper now scans `paste.ts` directly for the `type:` literals it constructs (verified independently: paste.ts alone yields `key_press` with no dependency on DesktopViewer.tsx's own call sites), and the DesktopViewer.tsx scan is now quote/property-order-agnostic. Re-verified the direct scan still yields exactly the same 7 kinds, with no WebRTC-only control-channel messages (`list_monitors`, `set_cursor_stream`, etc.) leaked in.
- **Important (inline comment-accuracy check):** my own doc comment on `desktopInputEvent` mislabeled the agent's `desktopInputTypes` allowlist (`agent/internal/heartbeat/handlers_desktop.go`) as "WebRTC-side" — it's actually the agent's counterpart of the *WS fallback* command-relay path (`handleDesktopInput`'s own comment there states "WebRTC sessions handle input via the data channel in the user helper"). **Fixed in 392ebf773.**
- **Suggestions (not addressed, out of scope for this PR):** (1) 8 legacy camelCase enum values (`mousemove`, `click`, `dblclick`, etc.) that the agent's `desktopInputTypes` never accepts — pre-existing dead surface, not introduced by this change, flagged as a follow-up cleanup candidate. (2) ~110 lines of mock boilerplate to import one zod schema in the new test file — matches the existing `desktopWs.test.ts` convention in this repo, not a regression.

**Tests:** `apps/api/src/routes/desktopWs*.test.ts` (8 files, 88 tests) green — `pnpm exec vitest run src/routes/desktopWs_inputSchema.test.ts src/routes/desktopWs.test.ts src/routes/desktopWs_onmessage.test.ts src/routes/desktopWs_lifecycle.test.ts src/routes/desktopWs_multitenant.test.ts src/routes/desktopWs_onopen.test.ts src/routes/desktopWs_rate_limit_cleanup.test.ts src/routes/desktopWs_utils_http.test.ts --pool=threads --maxWorkers=2`. Red-first verified: reverting the enum change made the new test fail 8/9 (schema-coverage cases + rejection case); restoring it passed 9/9. `tsc --noEmit -p apps/api` (matches CI's `typecheck` job) clean. `eslint` on both touched files clean.

PR #4399, opened 2026-09-01
fix(web,api): jitter throttled-refresh retries, single retry owner, correct rate-limit remaining (#3984)

### LINKED ISSUE #3984: [Web] Throttled-refresh recovery has no jitter and two competing retries — synchronized page-reload burst under load
Follow-ups from the review of #3975 (treat a throttled `/auth/refresh` as a wait, not a logout). The core fix is correct; these are about what happens when *many* clients are throttled at once, which is the condition that produces the throttle in the first place.

**1. No jitter — every client wakes in lockstep, and each does a full page reload.** Both the store's retry (`apps/web/src/stores/auth.ts:631`) and the mask's countdown (`apps/web/src/components/auth/AuthOverlay.tsx:347`) derive their deadline from the server-supplied `resetAt`. Every tab and every client that was throttled in the same window therefore retries at the same instant. Worse, the mask's recovery is `window.location.reload()` — heavier than the single refresh request it replaces. A rate limiter that fired because the endpoint was under load is answered with a synchronized burst of full page loads. Add jitter to both deadlines.

**2. Two independent retries for one throttle.** The store arms `setTimeout(waitMs)` to retry the refresh, and the mask separately counts down to a reload. They fire at the same moment; the reload usually preempts the store's retry, so the store's bounded-wait logic is partly wasted work. Pick one owner for recovery — the store is the better one, since it already bounds itself via `MAX_THROTTLE_WAITS`.

**3. `rate-limit.ts:88` reports `remaining` pre-refund.** With `refundOnReject: true`, a rejected request has its slot returned, but the `remaining` value sent to the client is computed before the refund, so it reports `0` when capacity was actually restored. Cosmetic, but it is the number a client would use to decide when to retry.

**4. `pages/remote/**` mounts no overlay**, so an `AuthThrottledError` there is unhandled. Acknowledged in the PR body; recording it here so it is not lost.

ToddHebebrand:
## Summary

Follow-ups from the review of #3975 (treat a throttled `/auth/refresh` as a wait, not a logout). The core fix there was correct; this addresses what happens when *many* clients are throttled at once — the condition that produces the throttle in the first place.

- **Jitter on the retry wait.** Both the store's own bounded wait (`apps/web/src/stores/auth.ts`) and the deadline `AuthThrottledMask` displays derive from the same server-supplied `Retry-After`. Every tab/client throttled in the same window previously retried at the exact same instant. `parseRetryAfterMs` now applies up to 25% additive jitter (never subtractive — retrying early just earns another 429) so a throttled fleet's own recovery isn't itself a synchronized burst.
- **Single retry owner.** The store's internal `setTimeout(waitMs)` retry and the mask's own countdown-to-`window.location.reload()` used to race the same deadline; the mask's reload usually preempted the store's own in-flight retry, wasting it. Now only the store schedules the one automatic recovery reload (`scheduleThrottleReload`), fired only once its bounded in-memory wait (`MAX_THROTTLE_WAITS`) is exhausted, and only if the access token is still actually missing (a keepalive throttle with a valid token never reloads). `AuthThrottledMask` is now pure display — its countdown never independently acts. The manual "Retry now" button is unchanged (an explicit user action, not a competing automatic retry).
- **Correct `remaining` under `refundOnReject`.** `apps/api/src/services/rate-limit.ts` reported `remaining` from the pre-refund count, so a client that had its rejected slot refunded (the whole point of `refundOnReject: true`) would still be told `remaining: 0` — exactly the number a well-behaved client reads to decide when to retry. Now reflects the post-refund count when the refund actually landed.
- **Mount AuthOverlay on remote pages.** `pages/remote/vnc/[tunnelId].astro`, `pages/remote/proxy/[tunnelId].astro`, and `pages/remote/terminal/[deviceId].astro` used bare `Layout.astro` with no `AuthOverlay`, so an `AuthThrottledError` there went unhandled. Also found and fixed `pages/remote/tools.astro`, which had the identical gap (same `RemoteToolsPage` + bare `Layout.astro` pattern) but wasn't named in the issue.

## Test plan

- [x] `apps/api/src/services/rate-limit.test.ts` — added 2 tests for post-refund `remaining` (success + refund-failure path); 50/50 passed.
- [x] `apps/web/src/stores/auth.ts` — added jitter-bounds test (random=0 → exact wait; random≈1 → capped at +25%) and 2 single-retry-owner tests (exactly one scheduled reload once the bounded wait is exhausted; no reload if the session was restored before the deadline). Updated the pre-existing exact-timing test to pin jitter to 0. 93/93 passed.
- [x] `apps/web/src/components/auth/AuthOverlay.test.tsx` — updated the "retries automatically" test: the mask's own countdown no longer reloads on its own (ownership moved to the store, covered above). 22/22 passed.
- [x] `npx astro check` (apps/web) — 0 errors, 0 warnings (pre-existing hints only), covering the 4 edited `.astro` files.
- [x] `npx eslint` on all touched `.ts`/`.tsx` files — clean.

Closes #3984

### DIFF
--- a/apps/api/src/services/rate-limit.test.ts
+++ b/apps/api/src/services/rate-limit.test.ts
@@ -214,7 +214,12 @@ describe('rate-limit service', () => {
 
       // #3984: `remaining` used to be computed from the pre-refund `count`, so
       // a well-behaved client honouring Retry-After would read `remaining: 0`
-      // even though its own rejected attempt's slot was just refunded.
+      // even though its own rejected attempt's slot was just refunded. Uses
+      // cost=2 deliberately: at cost=1 (both current refundOnReject callers)
+      // a rejection always means `remaining` is 0 either way, since refunding
+      // only this call's own 1 entry can never bring the count below `limit`
+      // — see the NOTE in rate-limit.ts. This proves the general contract for
+      // any weighted-cost caller, present or future.
       it('reports remaining from the post-refund count, not the pre-refund count', async () => {
         mockRejectedExec(2); // count=6, limit=5, cost=2 -> post-refund count=4
 
--- a/apps/api/src/services/rate-limit.test.ts
+++ b/apps/api/src/services/rate-limit.test.ts
@@ -212,6 +212,35 @@ describe('rate-limit service', () => {
         expect(mockRedis.zrem).toHaveBeenCalledWith('test-key', ...expectedMembers);
       });
 
+      // #3984: `remaining` used to be computed from the pre-refund `count`, so
+      // a well-behaved client honouring Retry-After would read `remaining: 0`
+      // even though its own rejected attempt's slot was just refunded.
+      it('reports remaining from the post-refund count, not the pre-refund count', async () => {
+        mockRejectedExec(2); // count=6, limit=5, cost=2 -> post-refund count=4
+
+        const result = await rateLimiter(mockRedis as Redis, 'test-key', 5, 60, 2, { refundOnReject: true });
+
+        expect(result.allowed).toBe(false);
+        // Pre-refund this would be Math.max(0, 5 - 6) = 0; post-refund it's
+        // Math.max(0, 5 - 4) = 1 — the capacity the refund actually restored.
+        expect(result.remaining).toBe(1);
+      });
+
+      it('reports remaining from the pre-refund count when the refund itself fails', async () => {
+        mockRejectedExec(2);
+        mockRedis.zrem = vi.fn().mockRejectedValue(new Error('redis down'));
+        const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+        const result = await rateLimiter(mockRedis as Redis, 'test-key', 5, 60, 2, { refundOnReject: true });
+
+        // The zrem never landed, so the entries are still in Redis — reporting
+        // restored capacity here would be a lie in the other direction.
+        expect(result.allowed).toBe(false);
+        expect(result.remaining).toBe(0);
+
+        consoleErrorSpy.mockRestore();
+      });
+
       it('does NOT call zrem when refundOnReject is true but the request is allowed', async () => {
         const now = Date.now();
         mockMulti.exec.mockResolvedValue([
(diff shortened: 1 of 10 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit code-reviewer + pr-test-analyzer (fresh, run separately due to sibling-agent subagent-concurrency limits at dispatch time; silent-failure-hunter round skipped given time budget — code-reviewer's pass already covered the error/silent-failure-shaped surface area: timer leaks, guard correctness, double-scheduling).

**Findings:** 9 raised across both agents → all addressed in `7abaaa84a`:
- Important: jitter applied after the 90s clamp, letting waits exceed `MAX_REFRESH_RETRY_AFTER_MS` — fixed by clamping to a pre-jitter base so base+jitter never exceeds the ceiling, and moving jitter to the single call site (`parseRetryAfterMs` stays pure parse+clamp).
- Important: a reload timer armed by one throttle cycle could still fire mid a *later* cycle's own in-flight wait — added a `tokenRefreshInFlight` guard so a stale timer defers to whichever cycle is actually in progress.
- Important: the store-owned reload fired unconditionally, including on pages that handle `AuthThrottledError` without ever mounting `AuthOverlay` (`ForcedMfaSetupPage` on `AuthLayout`, which deliberately shows a non-reloading "please wait" message) — added a `throttleMaskMounted` gate toggled by the mask's own mount/unmount effect.
- Important: stale `AuthThrottledError` docstring still said the remote viewers mount no `AuthOverlay`, after this PR mounted it on all four — updated.
- Important (framing): the `rate-limit.ts` `remaining` fix has no observable effect on either current `refundOnReject` caller (both use cost=1, where `remaining` is 0 on rejection with or without the refund) — corrected the comment/PR description to say this fixes the general `rateLimiter()` contract for a future weighted-cost caller, not an active client-visible bug today.
- Suggestion: adopted (jitter moved out of the parser for testability).
- 0 outstanding.

**Tests:** `apps/web` affected suites green — `src/stores/auth.test.ts` (97 passed) and `src/components/auth/AuthOverlay.test.tsx` (23 passed), including new coverage for the jitter ceiling boundary, the mask-mounted gate from both directions (mounted/unmounted, plus mid-flight unmount), the `tokenRefreshInFlight` race guard, `login()`/`logout()` cancelling a pending scheduled reload, and the `AuthOverlay`→store mount/unmount wiring itself. `apps/api` `src/services/rate-limit.test.ts` green (50 passed). `npx astro check` clean (0 errors, covering all 4 edited `.astro` files). `npx eslint` clean on every touched `.ts`/`.tsx` file. All single-forked via `--pool=threads --maxWorkers=2`, not the full suite.

PR #4400, opened 2026-09-01
feat(api): retention/prune job for ticket_outbox, intent_outbox and metric_anomaly_incidents (#4210)

### LINKED ISSUE #4210: Retention/prune job for ticket_outbox, intent_outbox and metric_anomaly_incidents
Follow-up to #3828 (PRs #4195, #4203). `ticket_outbox` (~6 rows per ticket lifecycle), `intent_outbox`, and `metric_anomaly_incidents` accumulate published/dispatched rows forever and are walked by org erasure. Add a `published_at/dispatched_at < now() - N days` retention worker.

ToddHebebrand:
## Summary

Follow-up to #3828 (PRs #4195, #4203). `ticket_outbox` (~6 rows per ticket lifecycle), `intent_outbox`, and `metric_anomaly_incidents` are transactional outbox tables — their own publisher workers (`ticketOutboxPublisher.ts`, `intentOutboxPublisher.ts`, `metricAnomalyIncidentPublisher.ts`) drain rows within seconds, but nothing ever deleted the drained rows, so all three accumulated every event forever and org erasure had to walk an ever-larger table for no benefit.

Adds three new daily BullMQ retention jobs following the batched-ctid-DELETE pattern established in PR #4365 (`retentionBatch.ts`'s `pruneInCtidBatches`) and instrumented per PR #4348 (`retentionMetrics.ts`):

- `apps/api/src/jobs/ticketOutboxRetention.ts`
- `apps/api/src/jobs/intentOutboxRetention.ts`
- `apps/api/src/jobs/metricAnomalyIncidentRetention.ts`

## Terminal-state rule

A row is only pruned once it is terminal, mirroring each table's own publisher's claim query:

- **delivered/dispatched** — `published_at`/`dispatched_at IS NOT NULL`. Cutoff measured from that timestamp.
- **permanently failed** — still unpublished, but `publish_attempts`/`dispatch_attempts` has exceeded the publisher's own stuck-row threshold (`MAX_PUBLISH_ATTEMPTS`, now exported from each publisher file so the retention cutoff can never drift out of sync with the value the publisher itself uses to give up retrying). Cutoff measured from `created_at`, since there's no delivery timestamp.

A row still in normal retry (unpublished, attempts within budget) is **never** touched regardless of age — it may still be delivered on the next publisher pass.

## Wiring

- Registered in `workerRegistry.ts` (`global` placement — no agent-socket dependency, same as the sibling publishers).
- Scheduled daily at `7:13`/`7:33`/`7:53` UTC via `scheduleRegistry.ts`'s allocated daily-tier slots (mod-5 lane, verified collision-free by `scheduleRegistry.contract.test.ts`).
- New `RETENTION_JOB_NAMES` entries (`ticket_outbox_retention`, `intent_outbox_retention`, `metric_anomaly_incident_retention`) wired into `retentionMetrics.ts` for the `breeze_retention_*` Prometheus series.
- Env knobs documented in `.env.example` (all optional, default 14-day retention / 5000-row batches / 50-batch cap per run).
- Updated the two registry drift-guard tests (`workerRegistry.test.ts`, `workerEntrypointClosure.contract.test.ts`) that enumerate `WORKER_REGISTRY` by name/count.

## Test plan

- [x] `pnpm exec vitest run` on the three new job test files (21 tests) plus every touched sibling (publishers, `scheduleRegistry.contract.test.ts`, `retentionMetrics.test.ts`, `workerRegistry.test.ts`, `workerEntrypointClosure.contract.test.ts`) — 189 tests, all green, single-fork (`--pool=threads --maxWorkers=2`).
- [x] `tsc --noEmit` clean across `apps/api`.
- [x] `eslint` clean on every touched/added file.
- [x] No schema/migration changes — no `db:check-drift` impact.

Closes #4210


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter` (parallel dispatch).

**Findings:** 2 real, independently-corroborated issues raised → both addressed in `433059ab`; 0 outstanding.

1. **Migration lock safety** (code-reviewer + silent-failure-hunter, converged independently): the new `2026-09-29-outbox-retention-indexes.sql` originally used a plain `CREATE INDEX IF NOT EXISTS` inside `autoMigrate`'s default per-file transaction, which takes a SHARE lock on three hot write-path tables (`ticket_outbox`/`intent_outbox`/`metric_anomaly_incidents`) for the whole index build — these tables have never been pruned before this PR, so their production size at deploy time is unknown. Fixed: `-- @no-transaction` + `CREATE INDEX CONCURRENTLY IF NOT EXISTS`, matching the existing `2026-05-17-g-agent-logs-composite-index.sql` precedent.
2. **Stuck-row purge loses its own forensic trail** (code-reviewer): folding the delivered/dispatched and permanently-failed branches into one OR'd DELETE reported a single aggregate count, so an operator couldn't tell "N drained rows" from "N events that never delivered" — the same concern CLAUDE.md codifies for migration cleanup DELETEs. Fixed: each job now runs the two branches as separate `pruneInCtidBatches` passes and warns distinctly when the stuck count is non-zero.

**Additional finding acted on (pr-test-analyzer, Critical):** the original unit tests only asserted the *rendered SQL text* mentions the right columns/operator against a mocked `db.execute` — that would not have caught an AND/OR flip or a `>` vs `>=` regression that deletes a row still within its publisher's retry budget (the exact bug class this PR exists to guard against). Added `apps/api/src/__tests__/integration/outboxRetention.integration.test.ts` (mirrors `mlOutputRetention.integration.test.ts`): seeds real rows in every state the WHERE clause distinguishes (delivered/dispatched-old, delivered/dispatched-recent, stuck-old, **mid-retry-old**) against a real Postgres and asserts by id exactly which survive — the mid-retry-old row surviving despite its age is the single most safety-critical assertion in the PR. Also covers `metric_anomaly_incidents`' documented invariant that `lastSeenAt` must never gate the cutoff, only `dispatchedAt`.

**Tests:** 281 unit tests green across the three new jobs + every touched sibling (publishers, `scheduleRegistry.contract.test.ts`, `retentionMetrics.test.ts`, `workerRegistry.test.ts`, `workerEntrypointClosure.contract.test.ts`, schema tests) — `vitest run --pool=threads --maxWorkers=2`. 7 new integration tests green against a private per-worktree Postgres/Redis stack (`pnpm test-stack up/down`) — `vitest run --config vitest.integration.config.ts`. `tsc --noEmit` clean. `eslint` clean on every touched/added file. No schema/migration drift beyond the new index migration (idempotent, correctly ordered, no-transaction).

PR #4401, opened 2026-09-01
fix(extensions): record activeVersion only after built-in activation succeeds (#3468)

### LINKED ISSUE #3468: extensions: built-in installed_extensions lifecycle polish + mid-pipeline-failure coverage
Follow-up from #3455 (workspace ee-merge), flagged in the final whole-branch review and the post-review hardening pass. Not blocking, no behavior bug today.

- Record `activeVersion` in `installed_extensions` only **after** activation succeeds (today it is recorded before, so a mid-pipeline failure can leave a row claiming an active version that never activated).
- Drop the redundant first-boot `setEnabled` call in the built-in registration path.
- Add a **two-builtin mid-pipeline-failure test**: with two built-ins registered, make the second fail partway through the staged pipeline and assert the first stays correctly activated and the failure aborts boot with accurate `installed_extensions` state.

Relevant code: `apps/api/src/extensions/builtinExtensions.ts` and the staged v1 pipeline it shares with signed runtime bundles.

ToddHebebrand:
## Summary

Follow-up from #3455 (workspace ee-merge), flagged in the final whole-branch review and post-review hardening pass. No behavior bug in practice today — this is lifecycle-state hardening for the staged built-in extension pipeline.

- `installed_extensions.activeVersion` is now written **only after** `registry.activate()` succeeds. Previously it was set in the same `upsertObserved()` call that seeds `configuredVersion`, before `activate()` ran — so a mid-pipeline activation failure could leave a row claiming an active version that was never actually activated. `recordActive()` (already called only after a successful `activate()`) is now the sole writer of `activeVersion`.
- Dropped the redundant first-boot `setEnabled(name, true)` call in `loadBuiltinExtensions`. The `installed_extensions.enabled` column already defaults to `true` on `INSERT` (`db/schema/extensions.ts`), and `upsertObserved`'s `UPDATE` path never touches `enabled` — so the explicit call was a no-op duplicating the DB default (and the `stateStore.get(name)` lookup that gated it is no longer needed either).

## Test plan

- [x] Added a two-builtin mid-pipeline-failure regression test in `apps/api/src/extensions/builtinExtensions.test.ts`: two real built-ins are registered where the second shares the first's `routeNamespace`, so `ExtensionContributionRegistry.activate()` genuinely throws (a real route-namespace-collision error, not a mock) partway through the second builtin's staged pipeline. Asserts the first builtin stays correctly activated (`activeVersion` = its version, `lifecycleState` = `'active'`), and the second's row shows `activeVersion: null` / `lifecycleState: 'discovered'` — i.e. `upsertObserved` seeded it but `recordActive` never ran.
- [x] `pnpm exec vitest run src/extensions/builtinExtensions.test.ts --pool=threads --maxWorkers=2` — 45 passed (all existing tests + the new one).
- [x] `npx tsc --noEmit` in `apps/api` — clean.
- [x] `npx eslint src/extensions/builtinExtensions.ts src/extensions/builtinExtensions.test.ts` — clean.

Closes #3468

### DIFF
--- a/apps/api/src/extensions/builtinExtensions.test.ts
+++ b/apps/api/src/extensions/builtinExtensions.test.ts
@@ -350,6 +350,64 @@ describe('loadBuiltinExtensions', () => {
     expect((await stateStore.get(NAME))?.enabled).toBe(false);
   });
 
+  // Regression for #3468: `activeVersion` must be written to
+  // `installed_extensions` only AFTER `registry.activate()` succeeds, not
+  // observed alongside `configuredVersion` before it runs. Two real built-ins
+  // registered, the second sharing the first's `routeNamespace` so
+  // `ExtensionContributionRegistry.activate` genuinely throws (a real
+  // namespace-collision failure, not a mocked one) partway through the second
+  // builtin's staged pipeline.
+  it('a mid-pipeline activation failure on the second built-in leaves the first correctly active and records no phantom activeVersion for the second', async () => {
+    const NAME_2 = 'demo-builtin-2';
+    const VERSION_2 = '2.0.0';
+    const ENABLE_ENV_VAR_2 = 'BREEZE_DEMO_BUILTIN_2_ENABLED';
+    const builtin2 = fixtureBuiltin({
+      name: NAME_2,
+      enableEnvVar: ENABLE_ENV_VAR_2,
+      // Same NAMESPACE as the first builtin (fixtureManifest's default) —
+      // ExtensionContributionRegistry.activate() throws a real "route
+      // namespace already owned" error for this, once the first builtin is
+      // already active.
+      manifest: fixtureManifest({ name: NAME_2, version: VERSION_2 }),
+    });
+
+    process.env[ENABLE_ENV_VAR_2] = 'true';
+    try {
+      const h = createHarness({ builtins: [fixtureBuiltin(), builtin2] });
+
+      await expect(h.load()).rejects.toThrow(/route namespace .* is already owned by extension "demo-builtin"/i);
+
+      // Full pipeline ran for builtin 1, and got as far as 'activate' (where
+      // it throws) for builtin 2 — nothing downstream of that ran for it.
+      expect(h.calls).toEqual([
+        'migration', 'tenancy', 'stage', 'validate', 'activate', 'web',
+        'migration', 'tenancy', 'stage', 'validate', 'activate',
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` was invoked, but subagent dispatch (code-reviewer, pr-test-analyzer, silent-failure-hunter) was unavailable for the duration of this session — the concurrent-subagent limit was saturated by ~38 sibling issue-fixer agents running in parallel on this machine, and repeated attempts (including after a wait) all returned "Concurrent subagent limit reached." Substituted a direct manual review covering the same three lenses, including an empirical check rather than just reading the diff.

**Findings:** 0 outstanding.
- Correctness: confirmed `upsertObserved` is called from exactly one place in `apps/api/src` (this function), so removing `activeVersion` from that call and making `recordActive` (already gated on a successful `registry.activate()`) the sole writer of `active_version` closes the gap with no new regression — including on a later boot where a row already has a persisted `activeVersion`, since `upsertObserved` only updates fields it's explicitly given.
- Safety of the dropped `setEnabled(true)` call: `installed_extensions.enabled` defaults to `true` on `INSERT` (`db/schema/extensions.ts:46`), and `upsertObserved`'s `UPDATE` path never includes `enabled` in its `SET` — confirmed against both the Drizzle backend and the in-memory test backend, which mirror each other. The `skipDisabledBuiltin` path never touched this call at all, so it's unaffected.
- Silent failures: no new catch/swallow introduced; the function's existing "any phase failure aborts boot" policy is unchanged — the removed code was pure state-store bookkeeping, not error handling.
- Test discrimination proven empirically, not just read: swapped the pre-fix `builtinExtensions.ts` (from `origin/main`) back in with the new test still in place and reran it — it failed exactly as expected (`expected '2.0.0' to be null`, i.e. the second built-in's row showed the phantom `activeVersion` the issue describes). Restored the fix and reran the full file: green again.

**Tests:** `apps/api` affected suite green — `pnpm exec vitest run src/extensions/builtinExtensions.test.ts --pool=threads --maxWorkers=2`, 45/45 passed (44 pre-existing + 1 new regression test, confirmed red pre-fix / green post-fix as above). `npx tsc --noEmit` in `apps/api` clean (exit 0). `npx eslint src/extensions/builtinExtensions.ts src/extensions/builtinExtensions.test.ts` clean.

PR #4403, opened 2026-09-01
fix(api): split remote-access launcher availability from issuance (#3402)

### LINKED ISSUE #3402: [API] Device detail decrypts a remote-access provider password just to compute a boolean

ToddHebebrand:
## Summary

GET /devices/:id resolved the *full* launcher — decrypting the provider password and building the substituted launch URL via `resolveRemoteAccessLauncherForDevice` — purely to compute the `hasRemoteAccessLauncher` boolean, then discarded the URL (`apps/api/src/routes/devices/core.ts:1125-1136`, pre-fix). That ran a decrypt (`decryptForColumn`) on every device-detail page load for no reason.

This splits `apps/api/src/services/remoteAccessLauncher.ts` into two entry points that share one internal pre-decrypt implementation:

- `checkRemoteAccessLaunchAvailability` — provider selection, enabled check, device-identifier presence, template non-empty. **No decrypt, no substitution.** This is now the *only* launcher call GET `/devices/:id` makes.
- `resolveRemoteAccessLaunch` — unchanged issuance path (decrypts the provider password, substitutes the template, re-checks the substituted scheme). Still the only thing POST `/devices/:id/remote-access-launch` calls.

Both call a shared internal `evaluateRemoteAccessLaunchAvailability` (provider selection + enabled/identifier/template checks) so availability and issuance can never disagree on skip-reason vocabulary or on *which* provider they're evaluating (the skew case called out in the issue: availability must pick the same provider issuance would, including the per-technician preference from #3391).

The one asymmetry, by design: `scheme_not_allowed` can only ever be reported by issuance, since detecting a template that resolves to a disallowed scheme requires substituting the real (decrypted) password — work availability deliberately skips. Availability's `skipReason` type therefore never includes it.

`apps/api/src/routes/devices/core.ts` now has three helpers instead of one: `loadRemoteAccessLauncherContext` (shared partner-settings + preference lookup, still wrapped in `withSystemDbAccessContext` for the partner-axis RLS reason documented inline), `checkRemoteAccessLauncherAvailabilityForDevice` (GET), and `resolveRemoteAccessLauncherForDevice` (POST, unchanged behavior).

## Test plan

- [x] `apps/api/src/services/remoteAccessLauncher.test.ts` — new `checkRemoteAccessLaunchAvailability` parity suite: a table-driven test asserting availability and issuance agree on `skipReason`/`providerId`/available-vs-launchUrl across every skip-reason scenario (no provider, unknown default, disabled provider, missing identifier, empty template, happy path, preference selection/fallback), plus an explicit test that `scheme_not_allowed` is issuance-only.
- [x] `apps/api/src/routes/devices/core.remoteAccessLaunch.test.ts` — new suite asserting GET `/devices/:id` resolves `hasRemoteAccessLauncher`/`remoteAccessLaunchSkipReason` correctly **without ever invoking `decryptForColumn`**, while POST `/devices/:id/remote-access-launch` still does (wrapped, not stubbed, so real crypto behavior is unchanged). Also rebuilt the file's DB mock to be a table-aware chainable stub so the GET handler actually reaches 200 in tests instead of previously crashing partway through on an unmocked query shape (the old test tolerated a 500).
- [x] `pnpm exec vitest run src/services/remoteAccessLauncher.test.ts src/routes/devices/core.remoteAccessLaunch.test.ts --pool=threads --maxWorkers=2` — 46/46 passed.
- [x] `npx tsc --noEmit` (apps/api) — clean.
- [x] `npx eslint` on all four touched files — clean.

Closes #3402

### DIFF
--- a/apps/api/src/routes/devices/core.remoteAccessLaunch.test.ts
+++ b/apps/api/src/routes/devices/core.remoteAccessLaunch.test.ts
@@ -1,33 +1,66 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { Hono } from 'hono';
+import { getTableName, is, Table } from 'drizzle-orm';
 
 // Hold a configurable partner-settings value the mocked select chain returns.
 // Each test mutates this before calling app.request().
 let mockPartnerSettings: unknown = {};
+
+// A fully chainable, table-aware query-builder stub. Every chain method
+// (from/innerJoin/where/orderBy/limit) returns the same node so any call
+// shape core.ts's GET /devices/:id handler exercises (hardware, network
+// interfaces, metrics, memberships, site, org, partner settings) resolves
+// without throwing, instead of crashing partway through on an unmocked
+// chain shape (which previously meant the GET handler never reached the
+// launcher-availability code at all -- see #3402). The only query that
+// needs a real answer is the partner-settings lookup keyed off `.from(partners)`;
+// everything else safely resolves to an empty array.
+function makeSelectMock() {
+  return vi.fn(() => {
+    let fromTableName: string | null = null;
+    const resolveRows = (): unknown[] =>
+      fromTableName === 'partners' ? [{ settings: mockPartnerSettings }] : [];
+    const node: any = {
+      from: vi.fn((table: unknown) => {
+        try {
+          fromTableName = is(table, Table) ? getTableName(table as Table) : null;
+        } catch {
+          fromTableName = null;
+        }
+        return node;
+      }),
+      innerJoin: vi.fn(() => node),
+      where: vi.fn(() => node),
+      orderBy: vi.fn(() => node),
+      limit: vi.fn(async () => resolveRows()),
+      then: (resolve: (v: unknown[]) => unknown, reject?: (e: unknown) => unknown) =>
+        Promise.resolve(resolveRows()).then(resolve, reject),
+    };
+    return node;
+  });
+}
 
 vi.mock('../../db', () => ({
   runOutsideDbContext: vi.fn((fn) => fn()),
   withDbAccessContext: vi.fn(async (_ctx: unknown, fn: () => Promise<unknown>) => fn()),
   withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
   db: {
-    select: vi.fn(() => ({
-      from: vi.fn(() => ({
-        innerJoin: vi.fn(() => ({
-          where: vi.fn(() => ({
-            limit: vi.fn(async () => [{ settings: mockPartnerSettings }])
-          }))
-        })),
-        where: vi.fn(() => ({
-          limit: vi.fn(async () => [])
-        }))
-      }))
-    })),
+    select: makeSelectMock(),
   }
 }));
 
 vi.mock('../../db/schema', async (importOriginal) => {
   const actual = (await importOriginal()) as Record<string, unknown>;
   return { ...actual };
+});
+
+// Wraps the REAL decryptForColumn so existing behavior (encrypt-at-rest
+// no-op for plaintext) is unchanged, while letting tests assert on whether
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` was invoked, but the `code-reviewer` and `pr-test-analyzer` subagents could not be dispatched — concurrent subagent capacity was exhausted machine-wide (38 sibling issue-fixer sessions running in parallel at the time). Rather than skip review, I performed the equivalent checks manually:

- **Behavior-preservation walk:** verified every branch of the original `resolveRemoteAccessLaunch` (`no_provider_configured`, `provider_disabled`, `missing_device_identifier`, `empty_url_template`, `scheme_not_allowed`, success) is reproduced exactly by the new shared `evaluateRemoteAccessLaunchAvailability`/`selectRemoteAccessProvider` helpers, including which `providerId` is reported on each failure path (matches the original's pattern of setting `providerId` even when unavailable).
- **Decrypt-boundary check:** confirmed `decryptForColumn` is referenced only inside `resolveRemoteAccessLaunch` (post the shared pre-decrypt check), never in `checkRemoteAccessLaunchAvailability` — the GET path (`checkRemoteAccessLauncherAvailabilityForDevice`) cannot reach it.
- **Consumer sweep:** `grep`'d the whole repo for callers of the launcher exports — only `apps/api/src/routes/devices/core.ts` consumes them, no other route or web code needed updating.
- **Discriminating-power check on the new tests:** deliberately mutated `remoteAccessLauncher.ts` (disabled the `empty_url_template` check) and re-ran the suite — the new `checkRemoteAccessLaunchAvailability — parity with issuance` table-driven test caught it immediately (the pre-existing `buildRemoteAccessLaunchUrl` test did not, since it only asserts the URL is null, not *why*). Reverted after confirming. This validates the parity suite genuinely discriminates divergence bugs between availability and issuance, not just passing vacuously.

**Findings:** 0 — no correctness, type-safety, or dead-code issues found in manual review.
**Tests:** `pnpm exec vitest run src/services/remoteAccessLauncher.test.ts src/routes/devices/core.remoteAccessLaunch.test.ts --pool=threads --maxWorkers=2` — 46/46 passed (12 new availability/issuance parity cases + 4 new GET-never-decrypts / POST-does-decrypt cases). `npx tsc --noEmit` (apps/api) clean. `npx eslint` on all 4 touched files clean.
**Status:** review-clean by manual equivalent review, awaiting maintainer merge. If subagent capacity frees up, a follow-up automated `code-reviewer`/`pr-test-analyzer` pass would still be reasonable but is not expected to surface anything new given the above.

PR #4404, opened 2026-09-01
fix(api): un-export resolveUserDisplayNames RLS-bypassing name oracle (#3983)

### LINKED ISSUE #3983: [API] resolveUserDisplayNames is an exported RLS-bypassing name oracle — constrain or rename it
Follow-up from the review of #3973 (resolve alert actor ids to display names).

`resolveUserDisplayNames` in `apps/api/src/routes/alerts/actorNames.ts:453` is **exported**, accepts an arbitrary list of user ids, and deliberately runs outside the request DB context (`runOutsideDbContext` → `withSystemDbAccessContext` at `:463`) so it can resolve names regardless of the caller's tenancy scope.

That is correct *as used today*: every id reaching it originates from an alert row that has already passed the org and site gates, and every writer sets the actor to `auth.user.id`, so an id can only ever be a user who legitimately acted on that org's alert. The review confirmed there is no cross-tenant path through the current call sites.

The problem is the shape, not the behaviour. As an exported helper it reads like a general-purpose "give me names for these ids" utility while actually being an **RLS-bypassing name oracle** — one careless future caller (an endpoint that echoes ids from a request body, a batch/list route that resolves ids it did not itself authorize) turns it into user enumeration across tenants. Nothing at the call site would look wrong.

Two cheap options, either is fine:

- un-export it and keep the batching internal to this module; or
- keep it exported but rename to something that cannot be adopted casually — `resolveUserDisplayNamesUnscoped` — with the tenancy precondition stated in its doc comment as a requirement on the caller.

Also worth noting from the same review, minor: `apps/api/src/openapi.ts:417` documents the list-route fields but not `dismissedByName`, which the by-id route produces.

ToddHebebrand:
## Summary

`resolveUserDisplayNames` in `apps/api/src/routes/alerts/actorNames.ts` deliberately runs outside request DB context (`runOutsideDbContext` → `withSystemDbAccessContext`) so it can resolve technician names regardless of the caller's tenancy scope. That's correct for its one real use — every id it receives already came off an access-checked alert row — but as an **exported** helper it reads like a general-purpose "resolve any user ids" utility. A future careless caller (an endpoint echoing ids from a request body, a batch route resolving un-authorized ids) could turn it into cross-tenant user enumeration with nothing at the call site looking wrong.

This PR:
- Un-exports `resolveUserDisplayNames`, making it module-private to `actorNames.ts`. `withAlertActorNames` (the sole legitimate caller) stays exported and unchanged in behavior.
- Adds a doc comment on the function stating the tenancy precondition and why it's kept private.
- Re-routes the three `resolveUserDisplayNames`-specific unit tests (dedup + selected columns, compiled SQL predicate, system-DB-context escape) through the public `withAlertActorNames` entry point instead of importing the internal function directly.

`git grep -n "resolveUserDisplayNames" apps/api/src` confirms the only references left are the function definition, its internal call site, and the test file's `describe` label/comments — no import of it remains anywhere.

Closes #3983

## Test plan
- [x] `pnpm exec vitest run src/routes/alerts/actorNames.test.ts --pool=threads --maxWorkers=2` — 10/10 passed
- [x] `npx tsc --noEmit` (apps/api) — clean
- [x] `git grep -n "resolveUserDisplayNames" apps/api/src` — no remaining import outside the module itself

### DIFF
--- a/apps/api/src/routes/alerts/actorNames.test.ts
+++ b/apps/api/src/routes/alerts/actorNames.test.ts
@@ -87,7 +87,10 @@ describe('resolveUserDisplayNames (via withAlertActorNames)', () => {
       { id: 'alert-2', acknowledgedBy: ADMIN_ID, resolvedBy: ADMIN_ID },
     ]);
 
-    expect(enriched.every((row) => row.acknowledgedByName === 'Breeze Admin')).toBe(true);
+    expect(enriched.map((row) => row.acknowledgedByName)).toEqual([
+      'Breeze Admin',
+      'Breeze Admin',
+    ]);
 
     // Assert the compiled SQL, not just that `where` was called: a mock-only
     // assertion would pass on an empty or wrong-column predicate.
--- a/apps/api/src/routes/alerts/actorNames.test.ts
+++ b/apps/api/src/routes/alerts/actorNames.test.ts
@@ -22,7 +22,7 @@ vi.mock('../../db', () => ({
 
 vi.mock('../../services/sentry', () => ({ captureException: captureExceptionMock }));
 
-import { resolveUserDisplayNames, withAlertActorNames } from './actorNames';
+import { withAlertActorNames } from './actorNames';
 
 const ADMIN_ID = '9cea2f85-2da1-445d-88cc-7c404d7504c4';
 const TECH_ID = '1f0e3f2c-9a2b-4c7d-9f10-8f6a2b3c4d5e';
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit code-reviewer (subagent), inline — checked correctness, residual export/import surface, and doc-comment accuracy.

**Findings:** 0 blocking. 1 optional nit (confidence ~60): the dedup test's `enriched.every(...)` assertion passes vacuously on an empty array — addressed in c7387a0ff by switching to the non-vacuous `.map(...).toEqual([...])` form already used elsewhere in the file. 0 findings outstanding.

Reviewer verified by mutation (removing the dedup, dropping the tenancy-escape wrapper, relaxing the empty-string id filter) that all three re-routed tests actually catch the corresponding regression — not a mock-only rewrite. Repo-wide `grep -rn "resolveUserDisplayNames"` confirms zero importers remain outside `actorNames.ts` and its own test file.

**Tests:** `apps/api/src/routes/alerts/actorNames.test.ts` — 10/10 passed (`pnpm exec vitest run --pool=threads --maxWorkers=2`). Reviewer additionally ran the full `apps/api/src/routes/alerts/` suite (24 files / 313 tests) green. `npx tsc --noEmit` (apps/api) clean.

Issue #4223, opened 2026-08-29
Patch job fails on battery preflight, but UI shows patches as Pending approval and hides failure reason


arch3rek:
### Description

When a Windows laptop is running on battery power, Breeze correctly blocks patch installation during the preflight check.

However, the failure reason is not surfaced in the UI. The affected patches continue to appear as `Pending approval`, even though a scheduled patch job was created and actually failed during preflight.

The real failure reason was only visible by querying the PostgreSQL database directly:

`preflight check "battery" failed: running on battery power (battery: 76%)`

This makes troubleshooting misleading, because the UI suggests an approval problem while the actual blocker is the battery preflight check.

### Steps to Reproduce

1. Configure a Windows device with Patch Management enabled.
2. Link an Update Ring with automatic approval enabled.
3. Configure a scheduled installation time.
4. Leave the laptop running on battery power.
5. Wait for the scheduled patch job to run.
6. Open Patch Management in the Breeze UI.

Observed backend state:
- A patch job is created.
- The job reaches the device preflight stage.
- The job fails because the device is on battery power.
- `patch_job_results.error_message` contains:
  `preflight check "battery" failed: running on battery power (battery: 76%)`

However, the UI continues to display affected patches as `Pending approval`.

### Expected Behavior

The UI should clearly indicate that patch deployment was blocked or failed because the device was running on battery power.

For example:

`Patching skipped — device is running on battery power. Connect AC power and retry.`

The patch/job status shown in the UI should reflect the actual deployment state rather than continuing to show `Pending approval`.

The battery preflight requirement should also be documented in the Patch Management documentation.

Optionally, Breeze could notify the logged-in user when a scheduled patch job is blocked by battery power, asking them to connect the device to AC power.

For example:

"Updates are ready to install. Please connect your device to AC power so scheduled patching can continue."

This would be especially useful for laptops that are frequently used on battery power and would help make patching truly unattended.

### Actual Behavior

The scheduled patch job fails during the battery preflight check, but the UI does not display the failure reason.

Affected patches remain displayed as `Pending approval`.

The actual cause can only be discovered by querying the Breeze PostgreSQL database and inspecting `patch_job_results.error_message`.

Example:

`preflight check "battery" failed: running on battery power (battery: 76%)`

### Component

Agent (Windows)

### Version

Breeze Web/API: 0.108.0 Agent: 0.108.0

### Relevant Logs

```shell
Patch job status: failed
Devices total: 1
Devices completed: 0
[...]

PR #4405, opened 2026-09-01
fix(api): stop stamping agent.command.* audit rows as success at dispatch (#4225)

### LINKED ISSUE #4225: Audit log records agent.command.* as result=success at dispatch; install_patches renders as "Patches installed" even when the job later fails
**Split out of #4223 (reported by arch3rek).**

**What happens:** the audit row for every queued agent command is written at dispatch time with `result: 'success'` hardcoded — `apps/api/src/services/commandQueue.ts:532-542` (and the second write path at `:944-955`). Nothing ever updates that row after the command completes (no `update(auditLogs)` exists anywhere in the patch execution path). The device events view then maps `agent.command.install_patches` to the label **"Patches installed"** (`apps/api/src/routes/devices/events.ts:353`), and `formatActionMessage` (`events.ts:383-396`) only appends "(failed)" when `result === 'failure'` — which can never occur.

**Observed consequence (#4223):** a patch job that failed its battery preflight — every `patch_job_results` row `status = failed` — shows in the audit log as "Patches installed", result success, listing all 13 patches. The audit trail asserts an outcome that never happened.

**Why it's shaped this way:** the dispatch audit is deliberately fire-and-forget (`commandQueue.ts:496-509`) so audit failures can't block command delivery. That explains the write-at-dispatch, not the success stamp or the completed-tense label.

**Fix direction:**
- Minimum: rename the labels to dispatch semantics ("Patch install command sent") and stamp `result` as something neutral (`dispatched`), so the row stops asserting completion.
- Better: emit a second audit event on command completion (success/failure with the error message), keeping the dispatch row as-is. Distinguishing dispatched / completed / failed is what the reporter asked for and matches how `patch_job_results` already models it.

**Scope note:** this is generic to `agent.command.*`, not just patches — every command type audit-logs success at enqueue.

ToddHebebrand:
## Summary

`services/commandQueue.ts` writes the `agent.command.*` audit row at
**dispatch time** (fire-and-forget, by design) but hardcoded
`result: 'success'` at both write sites (`queueCommand` and `executeCommand`).
Nothing ever updates that row after the command completes, so a command that
later fails still reads as a successful completion in the audit trail — e.g.
a patch job that failed its preflight showed as **"Patches installed"**,
`result: success`, listing all patches (#4223).

## Fix (minimal, per issue scope)

- Added a neutral `'dispatched'` value to the `audit_result` enum:
  `packages/shared/src/constants/index.ts` (the single source of truth,
  widens the Drizzle enum / Zod validators / OpenAPI spec for free) +
  `apps/api/migrations/2026-09-29-audit-result-dispatched.sql`
  (`ALTER TYPE ... ADD VALUE IF NOT EXISTS`, isolated in its own file per the
  existing `2026-09-05-b-audit-actor-type-ai-agent.sql` pattern — Postgres
  forbids using a value added by `ALTER TYPE` in the same transaction).
- Both `result: 'success'` audit-insert sites in `commandQueue.ts` now write
  `result: 'dispatched'`.
- Updated the `agent.command.*` labels in `routes/devices/events.ts` from
  completed-tense ("Patches installed") to dispatch-tense ("Patch install
  command sent") so the rendered message doesn't assert an outcome that
  hasn't happened. `formatActionMessage`'s failure/denied suffix logic is
  unaffected — 'dispatched' is neutral and gets no suffix, same as before.
- Updated `audit.enums.test.ts`'s pinned enum-parity assertion to include the
  new value.

## Out of scope (explicitly, per issue)

A second completion-time audit event (dispatched → completed/failed) is
**not** built here — that's the "better" fix the issue calls out as a
separate, larger change. This PR only stops the dispatch-time row from
falsely claiming success.

## Tests

- `apps/api/src/services/commandQueue.test.ts`: both audit-insert call sites
  (`queueCommand`'s fire-and-forget block, and `executeCommand`'s org-scoped
  insert) now assert `result: 'dispatched'` instead of `'success'`.
- `apps/api/src/routes/devices/events.test.ts`: `formatActionMessage` tests
  updated to dispatch-tense copy, plus a new assertion that the neutral
  `'dispatched'` result gets no `(failed)`-style suffix.
- `apps/api/src/db/schema/audit.enums.test.ts`: enum parity assertion
  updated to include `'dispatched'`.

Affected suites run single-fork, all green:
```
apps/api: commandQueue.test.ts, devices/events.test.ts, audit.enums.test.ts
  Test Files  3 passed (3)
       Tests  76 passed (76)
```
`tsc --noEmit` clean on `apps/api` (pre-existing unrelated errors in
`packages/shared/src/validators/quotes.test.ts` confirmed present on
`origin/main` before this change, untouched here).

Closes #4225

### DIFF
--- a/apps/api/migrations/2026-09-29-audit-result-dispatched.sql
+++ b/apps/api/migrations/2026-09-29-audit-result-dispatched.sql
@@ -1,0 +1,13 @@
+-- #4225: agent.command.* audit rows are written at dispatch time
+-- (services/commandQueue.ts) and previously hardcoded result='success', so a
+-- command that later fails still reads as a successful completion in the
+-- audit trail. Add a neutral 'dispatched' outcome to the audit_result enum
+-- so the dispatch-time row can stop asserting an outcome it doesn't know yet.
+--
+-- This file contains ONLY the ALTER TYPE. Postgres forbids USING a value
+-- added by ALTER TYPE ... ADD VALUE inside the same transaction, and
+-- autoMigrate wraps each file in one — so any statement referencing
+-- 'dispatched' must live in a later file, not here (mirrors
+-- 2026-09-05-b-audit-actor-type-ai-agent.sql).
+
+ALTER TYPE public.audit_result ADD VALUE IF NOT EXISTS 'dispatched';
--- a/apps/api/src/db/schema/audit.enums.test.ts
+++ b/apps/api/src/db/schema/audit.enums.test.ts
@@ -114,7 +114,7 @@ describe('actor_type enum parity (shared ↔ DB schema ↔ validators ↔ OpenAP
   });
 
   it('audit_result parity stays intact alongside', () => {
-    expect([...auditResultEnum.enumValues].sort()).toEqual(['denied', 'failure', 'success']);
+    expect([...auditResultEnum.enumValues].sort()).toEqual(['denied', 'dispatched', 'failure', 'success']);
     expect([...AUDIT_RESULTS].sort()).toEqual([...auditResultEnum.enumValues].sort());
     const zodOptions = auditQuerySchema.shape.result.unwrap().options;
     expect([...zodOptions].sort()).toEqual([...auditResultEnum.enumValues].sort());
--- a/apps/api/src/routes/devices/events.test.ts
+++ b/apps/api/src/routes/devices/events.test.ts
@@ -109,14 +109,9 @@ describe('formatActionMessage (automated command labels)', () => {
   // #4225: these rows are written at DISPATCH time (commandQueue.ts), before
   // the agent has reported back, so the label must not claim completion and
   // the audit row's `result` must not claim 'success'.
-  it('labels automated patch installs in dispatch tense, not completed tense', () => {
+  it('labels automated patch installs in dispatch tense, not completed tense, with no suffix for the neutral result', () => {
     expect(formatActionMessage('agent.command.install_patches', 'host-1', 'dispatched'))
       .toBe('Patch install command sent — host-1');
-  });
-
-  it('does not append a suffix for the neutral dispatched result', () => {
-    expect(formatActionMessage('agent.command.install_patches', 'host-1', 'dispatched'))
-      .not.toContain('(failed)');
   });
 
   it('labels automated script runs', () => {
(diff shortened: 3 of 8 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (verified locally: affected API tests, `autoMigrate.test.ts`, `tsc --noEmit`, and the touched web test all pass). Test-coverage assessment done directly (subagent pool was saturated by parallel sibling issue-fixers).

**Findings:** 3 raised → all addressed in `be78d568f`; 0 outstanding.
- Important: a comment in `commandQueue.ts` claimed no completion-time audit event exists anywhere. That's wrong — `agent.command.result.submit` is written on both transports (`agentWs.ts`, `routes/agents/commands.ts`) with a real success/failure result. The actual gap is narrower: that event is keyed on `resourceId = commandId` with no `deviceId` in `details`, so it can't join a *device's* activity feed (which matches `resourceId = deviceId OR details->>'deviceId'`). Comment corrected to name the real event and the real gap, so it doesn't mislead a future reader into either duplicating the event or concluding the audit log holds no completion data.
- Suggestion: dropped a near-vacuous test in `events.test.ts` whose assertion was strictly weaker than (and could never fail independently of) the test directly above it.
- Suggestion: refreshed two stale fixtures in `DeviceActivityFeed.test.tsx` that still encoded pre-fix completed-tense copy (`'Patches installed'`) and `result: 'success'` for `agent.command.*` rows.

Verified clean by the reviewer: migration isolation (only the bare `ALTER TYPE`, mirrors `2026-09-05-b-audit-actor-type-ai-agent.sql`, correctly dated after the last `2026-09-28-*` file), enum widening reaches every guarded surface (Drizzle, shared type, Zod, OpenAPI ×2) since they all derive from `AUDIT_RESULTS`, no missed `'success'`-tense assumption elsewhere in `formatActionMessage`/`DeviceActivityFeed.tsx`, and no i18n surface to follow (none of the renamed labels has a translated counterpart).

**Tests:** `apps/api` affected suite green (`commandQueue.test.ts` + `devices/events.test.ts` + `audit.enums.test.ts`, 75 passed); `apps/web` `DeviceActivityFeed.test.tsx` green (13 passed); `tsc --noEmit` clean on `apps/api` (pre-existing, unrelated errors in `packages/shared/src/validators/quotes.test.ts` confirmed present on `origin/main` before this change).

PR #4406, opened 2026-09-01
fix(ai): don't evict an actively-processing OpenAI session; expire the row on every eviction path (#4384)

### LINKED ISSUE #4384: [API/AI] OpenAI session eviction can abort an actively-processing turn, and idle/LRU evictions leave ai_sessions.status='active' forever
## Description

Two related defects in the eviction paths of `apps/api/src/services/llm/openaiSessionManager.ts` (verified against `origin/main`).

**1. LRU eviction can kill a session mid-stream.** `session.lastActivityAt` is refreshed only inside `getOrCreate()` (~L72, L82). Neither `tryTransitionToProcessing()`, `startTurn()`, nor `runTurn()` touches it, and neither `evictStaleSessions()` (~L341) nor `evictLeastRecentlyActive()` (~L370) checks `state === 'processing'` before calling `remove()`. Under cap pressure (`MAX_ACTIVE_SESSIONS = 200`, ~L89) the least-recently-*created-or-resumed* session can be the one currently streaming a response: `remove()` aborts its controller and closes its event bus mid-turn. The provider swallows the user-kind abort (`openaiCompatibleProvider.ts` ~L144/L291 returns on `abortKind === 'user'`), so the turn ends "cleanly" — the partial `assistantText` is persisted as a complete assistant message and the terminal `done` publish lands on a closed bus. The idle branch (2h) can't realistically hit a processing session, but LRU can.

**2. Idle-timeout and LRU evictions never mark the DB row expired.** Only the age branch (`age > SESSION_MAX_AGE_MS`, ~L353) writes `status: 'expired'` to `aiSessions`. The idle branch and `evictLeastRecentlyActive()` delete the in-memory session but leave `aiSessions.status = 'active'` permanently — nothing else in the file corrects it, so anything keyed on `status = 'active'` overcounts, and it worsens under exactly the load that triggers LRU.

## Proposed Fix

- Skip `state === 'processing'` sessions in both eviction paths (or refresh `lastActivityAt` when a turn starts/streams so a live session can't be the LRU victim); if a processing session must be evicted at hard cap, prefer the oldest *idle* one first.
- Perform the `status: 'expired'` (or a distinct terminal status) DB update on all three eviction paths, not just the 24h-age branch.

## Affected Files

- `apps/api/src/services/llm/openaiSessionManager.ts` (`evictStaleSessions`, `evictLeastRecentlyActive`, `getOrCreate`)

## Notes

Found during a blind model-review benchmark over this file (same exercise that produced #4329); findings independently reported by two reviewers and verified by hand against `origin/main`. Not checked: whether any UI or report currently queries `ai_sessions.status = 'active'`.

ToddHebebrand:
Closes #4384

## What was wrong

**1. Eviction could kill a session mid-stream.** `session.lastActivityAt` was refreshed only in `getOrCreate()`, and neither `evictStaleSessions()` nor `evictLeastRecentlyActive()` checked `state === 'processing'`. Under cap pressure the least-recently-*created-or-resumed* session could be the one currently streaming. `remove()` aborts its controller and closes its event bus mid-turn; because `openaiCompatibleProvider` returns on a user-kind abort without yielding an error, `hadError` stays false and the partial `assistantText` is persisted as a **complete** assistant message while the terminal `done` lands on a closed bus.

**2. Idle and LRU evictions never retired the DB row.** Only the 24h-age branch wrote `status: 'expired'`.

## The fix

**In-flight turns are protected.** Both eviction paths skip a session with a turn in flight. Liveness is `state` **and** recent stream progress, never `state` alone: `lastActivityAt` is refreshed at `tryTransitionToProcessing` and on every content delta, and a processing session with no progress for `PROCESSING_STALL_TIMEOUT_MS` (10 min) is a dead turn and stays reclaimable. Without that, `runTurn` throwing before its trailing `state = 'idle'` — its outer `try` has a `finally` but no `catch` — would pin a session forever. When *every* session is mid-turn, LRU evicts nothing and raises a throttled capacity alarm; overshooting the soft cap is self-correcting, corrupting a live turn is not.

**Staleness evictions retire the row; capacity eviction does not.** The idle and 24h paths now write `'expired'` (an existing enum value, no migration), matching what `runPreFlightChecks` would have written lazily on the next request. LRU deliberately does **not** — see below.

## Two judgement calls a maintainer should confirm

**a. LRU eviction leaves the row `active`, contrary to the issue's letter.** The issue asked for the expire write on all three paths, with the rationale that "anything keyed on `status = 'active'` overcounts" — and noted that whether anything actually queries it was *not checked*. It was checked here: every consumer (`services/aiAgent.ts:186`, `routes/mobile.ts`, `routes/clientAi/*`, `routes/helper/index.ts`, `services/aiCostTracker.ts`) is a list/display select. No quota, capacity, or billing gate keys on it.

Meanwhile an LRU victim is a *healthy* conversation dropped for our own capacity reasons: history lives in `ai_messages`, `buildMessagesFromHistory` rebuilds it, and pre-fix the user's next message transparently recreated the session. Expiring it turns a transient server condition into a hard 410 for a conversation minutes old, since preflight rejects on status before `getOrCreate` runs. That is the same recoverable in-memory drop as a deploy — which `shutdown()` is already careful not to expire, now pinned by a test. So the row stays truthful: `active` means resumable, and preflight still expires it lazily once it genuinely idles out. **Say the word and it's a one-line revert to the literal reading.**

**b. `'expired'` now maps to 410, not 400.** `runPreFlightChecks` checks `status !== 'active'` *before* the age branches and returned `'Session is not active'`, which lacks the word "expired" that `routes/ai.ts` matches on for 410. Expiring eagerly therefore moved the idle path from 410 to 400 — a regression — and the 24h path already had it. The status branch now reports an expired session as expired, so one contract holds however the row got there. `'closed'` is unchanged, and no caller string-matches the old message (verified across `routes/ai.ts`, `routes/scriptAi.ts`, helper routes).

## The tenancy trap, and the bug it hid

`withDbAccessContext` **joins** an already-open context instead of replacing it (`apps/api/src/db/index.ts:529-531`), so a write nested in a request runs under the *requester's* GUCs and, for another tenant, matches zero rows under RLS while reporting success.

The first draft hoisted `runOutsideDbContext` around the whole expire loop. That is correct only for the first iteration: `AsyncLocalStorage.exit()` does not cover iterations resuming after an `await`. A test evicting a three-org cohort under an open requester context returned `['org-a', 'org-requester', 'org-requester']` — orgs 2..N silently stranded. The escape is now re-entered per statement, and `extractRowCount` + `captureMessage('db_write_expecting_rows_zero')` makes a future regression *visible* rather than silent, since a zero-row batch is the RLS signature.

Expire writes are also grouped one statement per org (`inArray`) and serialized: previously this could open one transaction per session — up to 200 — against a `DB_POOL_MAX` of 30 shared with live traffic, and idle cohorts expire together after a spike.

## Verification

- `openaiSessionManager.eviction.test.ts` — 26 tests, confirmed **red first** (9 failed / 4 passed against unfixed code), plus a co-located preflight test. Every load-bearing assertion was mutation-tested rather than assumed:

  | Mutation | Result |
  |---|---|
  | Hoist the context escape out of the expire loop | 2 cross-tenant tests fail |
  | Drop the in-flight skip in LRU | 3 tests fail |
  | Remove the `content_delta` keepalive | 2 tests fail |
  | Flip `<=` to `<` at the stall boundary | 1 test fails |
  | Collapse the expired→410 ternary | 1 test fails |

  Three tests that passed with their own fix reverted were **fixed or removed rather than shipped**: a timer-context assertion (deleted — the per-statement escape means no write can observe it), a zero-row alarm control (was waiting on a synchronously-issued write, so it resolved before the check it asserted), and the stall boundary (LRU takes one victim, so the at-edge session survived by position, not protection).
- `pnpm exec vitest run src/services/llm/ src/services/aiAgentSdk.test.ts src/services/sentryEventCodes.test.ts` — **11 files, 302 passed**. AI route suites (`ai_sessions_crud`, `ai_sessions_actions`, `ai.ticket`, `ai.m365session`, `scriptAi_*`) green.
- `tsc --noEmit` exit 0; `eslint` clean on all touched files.

The db mock uses a real `AsyncLocalStorage` rather than a save/restore flag, so it cannot drift from the join-if-open contract it is standing in for.

## Follow-up, not fixed here

`streamingSessionManager` (the **default** Anthropic path) carries both identical defects — no in-flight check on either eviction path, and only its 24h branch writes `'expired'`. Its 24h write also lacks the context escape, and is safe only because that singleton is module-level. Tracked as #4514 rather than doubling this PR into the primary AI path.

### DIFF
--- a/apps/api/src/services/aiAgentSdk.test.ts
+++ b/apps/api/src/services/aiAgentSdk.test.ts
@@ -449,6 +449,20 @@ describe('runPreFlightChecks', () => {
     expect(result).toEqual({ ok: false, error: 'Session is not active' });
   });
 
+  it('words an already-expired session as expired, so routes map it to 410', async () => {
+    // This branch runs BEFORE the age checks below, so once eviction retires a
+    // row eagerly it becomes the common path for expired sessions. Collapsing
+    // it back to one string silently downgrades every evicted session from 410
+    // to 400, while the lazy age branches keep producing the right wording —
+    // which is exactly what makes the regression invisible.
+    mockGetSession.mockResolvedValue(makeSession({ status: 'expired' }));
+    const result = await runPreFlightChecks('session-1', 'hello', auth);
+    expect(result).toMatchObject({
+      ok: false,
+      error: expect.stringContaining('expired'),
+    });
+  });
+
   // --- Turn limit ---
 
   it('returns error when turn limit is reached', async () => {
--- a/apps/api/src/services/aiAgentSdk.ts
+++ b/apps/api/src/services/aiAgentSdk.ts
@@ -387,7 +387,18 @@ export async function runPreFlightChecks(
   }
 
   if (session.status !== 'active') {
-    return { ok: false, error: 'Session is not active' };
+    // 'expired' must read as expired to the caller: routes map on the word to
+    // return 410, and a session retired eagerly by openaiSessionManager's
+    // eviction reaches this branch BEFORE the age checks below would have
+    // produced that wording lazily. Without this, the same terminal state
+    // surfaced as 410 or 400 depending purely on which path got there first.
+    return {
+      ok: false,
+      error:
+        session.status === 'expired'
+          ? 'Session has expired. Please start a new session.'
+          : 'Session is not active',
+    };
   }
 
   if (session.turnCount >= session.maxTurns) {
--- a/apps/api/src/services/llm/openaiSessionManager.eviction.test.ts
+++ b/apps/api/src/services/llm/openaiSessionManager.eviction.test.ts
@@ -31,7 +31,14 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { PgDialect } from 'drizzle-orm/pg-core';
 import type { SQL } from 'drizzle-orm';
 
-const { dbUpdateMock, capturedExpires, mockState, contextStore, captureExceptionMock } =
+const {
+  dbUpdateMock,
+  capturedExpires,
+  mockState,
+  contextStore,
+  captureExceptionMock,
+  captureMessageMock,
+} =
   vi.hoisted(() => {
     const { AsyncLocalStorage } = require('node:async_hooks') as typeof import('node:async_hooks');
     return {
@@ -44,6 +51,8 @@ const { dbUpdateMock, capturedExpires, mockState, contextStore, captureException
       mockState: {
         /** Context in force at the moment the UPDATE is issued. */
         effectiveContext: null as unknown,
+        /** Row count the next UPDATE resolves with (0 = the RLS signature). */
+        nextRowCount: 1,
       },
       /**
        * The REAL primitive the db module uses, not a hand-rolled stand-in. A
@@ -54,6 +63,7 @@ const { dbUpdateMock, capturedExpires, mockState, contextStore, captureException
        */
       contextStore: new AsyncLocalStorage<unknown>(),
       captureExceptionMock: vi.fn(),
+      captureMessageMock: vi.fn(),
     };
   });
 
@@ -78,7 +88,7 @@ vi.mock('../../db', () => ({
 
 vi.mock('../sentry', () => ({
   captureException: captureExceptionMock,
-  captureMessage: vi.fn(),
+  captureMessage: captureMessageMock,
 }));
 
 import { OpenAISessionManager } from './openaiSessionManager';
@@ -132,6 +142,8 @@ describe('OpenAISessionManager eviction (#4384)', () => {
   beforeEach(() => {
     capturedExpires.length = 0;
     mockState.effectiveContext = null;
+    mockState.nextRowCount = 1;
+    captureMessageMock.mockClear();
     captureExceptionMock.mockClear();
     dbUpdateMock.mockReset();
     dbUpdateMock.mockImplementation(() => ({
@@ -142,7 +154,7 @@ describe('OpenAISessionManager eviction (#4384)', () => {
             set: values,
             where: clause,
           });
-          return Promise.resolve([]);
+          return Promise.resolve({ count: mockState.nextRowCount });
         },
       }),
     }));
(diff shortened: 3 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (three separate rounds; each re-run against the code the previous round produced).

**Findings:** 16 raised → 15 addressed, 1 declined with reasons. 0 outstanding.

Round 1 (code-reviewer) — no blockers, 3 should-fix:
- Terminal status surfaced as **400, not 410**. `runPreFlightChecks` checks `status !== 'active'` before the age branches, and its message lacked the word the route matches on — so expiring eagerly silently downgraded the idle path from 410 to 400. Fixed in `5a89732`.
- **Unbounded transaction fan-out**: up to 200 concurrent transactions against a `DB_POOL_MAX` of 30, on the newly-common idle path. Now one grouped statement per org, serialized.
- `shutdown()`'s non-expiry was unpinned while every `afterEach` called it. Test added.

Round 2 (silent-failure-hunter) — verified the RLS escape empirically (it rebuilt the mock semantics standalone and ran all three variants), then found 5 more:
- **The eviction timer inherits the constructing request's DB context.** This manager is a *lazy* singleton built inside a request, so `setInterval` carries that ALS scope on every tick for the process lifetime — my comment claimed the opposite. Escaped at construction.
- A **zero-row UPDATE was unobservable**, which is precisely the RLS-denial signature the escape exists to prevent. Now `extractRowCount` + `captureMessage('db_write_expecting_rows_zero')`.
- Capacity alarm reached stdout only; now throttled to Sentry. Plus `.catch()` on the fire-and-forget promise, `finally` around the sweep, and session ids in the failure log (**not** as Sentry tags — the scrubber allowlist voids tenant-scoped tags, so they would have silently vanished).

Round 3 (pr-test-analyzer) — proved a second false control by experiment and found the 410 change and the `content_delta` keepalive had **zero** coverage. All closed in `f51c560`.

**Declined:** porting the fix to `streamingSessionManager` (the default Anthropic path), which carries both identical defects. Filed as **#4514** rather than doubling this PR into the primary AI path.

**Self-caught, worth flagging:** batching the writes made the loop `await`, and hoisting `runOutsideDbContext` around it was correct only for the first iteration — `AsyncLocalStorage.exit()` does not cover continuations resuming after an `await`. A three-org cohort under an open requester context returned `['org-a', 'org-requester', 'org-requester']`; orgs 2..N would have matched zero rows under RLS and reported success. Fixed in `4268d3c` and pinned.

**Tests:** `vitest run src/services/llm/ src/services/aiAgentSdk.test.ts src/services/sentryEventCodes.test.ts` → **11 files, 302 passed**. Eviction suite confirmed red-first (9 failed / 4 passed unfixed) and every load-bearing assertion mutation-tested — the table in the PR body lists the five mutations and what each reddens. `tsc --noEmit` exit 0; `eslint` clean.

**Two calls for you, both one-liners to reverse:**
1. **LRU eviction no longer expires the row**, contrary to the issue's letter. The issue's rationale was that something keys on `status='active'` and noted that was unchecked — it is now checked, and every consumer is a list/display select. Meanwhile expiring turned a capacity blip into a hard 410 for a resumable conversation. Details in the PR body.
2. `'expired'` now maps to **410 instead of 400** across both AI routes.

**Status:** review-clean, CI running, awaiting maintainer merge. Not merged and #4384 left open, per the runbook.

PR #4407, opened 2026-09-01
fix(api): record per-patch status instead of collapsing to batch status (#4267)

### LINKED ISSUE #4267: patch_job_results records every patch as failed when one patch in the batch fails
### Description

Found while fixing #4228 (PR #4265) — same root cause family, different symptom, so it is filed separately rather than folded in.

`recordDeviceExecution` in `apps/api/src/jobs/patchJobExecutor.ts` decides each per-patch row's status with:

```ts
const perPatchResult = parsedResult?.results?.find(
  (r) => r.patchId === patch.patchId || r.externalId === patch.externalId
);
const patchSuccess = perPatchResult?.success ?? overallSuccess;
```

The Windows agent never emits a per-patch `success` field. `patchCommandResultFields` (`agent/internal/heartbeat/heartbeat.go`) builds each entry from `id`, `source`, `externalId`, `packageId`, `title`, `installId`, and the caller then sets **`status`** — `'installed'`, `'failed'`, or `'rolled_back'` — plus `error` / `rebootRequired` / `message`. There is no `success` key anywhere in the payload.

So `perPatchResult?.success` is always `undefined`, and `patchSuccess` collapses to `overallSuccess` for **every** patch in the batch.

### Impact

In the 13-patch/1-failure scenario from #4228, all 13 `patch_job_results` rows are written with `status: 'failed'` even though 12 installed cleanly. The `output` and `errorMessage` columns are populated from the same lookup, so per-patch error detail is lost too.

Note the device *Patch Operation History* view is **not** affected — `GET /devices/:id/patches/history` (`apps/api/src/routes/devices/patches.ts`) reads the agent payload straight off `device_commands`, so that screen shows the truth. This is specifically about the `patch_job_results` table and anything downstream of it (patch job detail, compliance reporting).

### Suggested fix

Read the field the agent actually sends, keeping `success` as a defensive fallback:

```ts
const patchSuccess =
  perPatchResult?.success
  ?? (perPatchResult?.status
    ? perPatchResult.status === 'installed' || perPatchResult.status === 'rolled_back'
    : overallSuccess);
```

PR #4265 already added the same `status`-aware predicate for its `anyPatchInstalled` gate, so the two should be factored into one shared helper.

Worth checking in the same pass whether the `r.patchId === patch.patchId` half of the `find` ever matches — the agent sends the patch id as `id`, not `patchId`, so the lookup appears to succeed only via the `externalId` branch.

### Component

API

### Additional Context

Confirmed by code reading against `origin/main` (agent payload shape in `agent/internal/heartbeat/heartbeat.go:executePatchInstallCommand` / `patchCommandResultFields`). Not reproduced against a live device — labelling as inferred-from-source, not field-observed.

No test covers `patch_job_results` row status today, so a fix should add one.

ToddHebebrand:
## The bug

`recordDeviceExecution` in `apps/api/src/jobs/patchJobExecutor.ts` decided each per-patch `patch_job_results` row's status with:

```ts
const perPatchResult = parsedResult?.results?.find(
  (r) => r.patchId === patch.patchId || r.externalId === patch.externalId
);
const patchSuccess = perPatchResult?.success ?? overallSuccess;
```

The Windows agent never emits a per-patch `success` field. `patchCommandResultFields` (`agent/internal/heartbeat/heartbeat.go`) builds each entry from `id`, `source`, `externalId`, `packageId`, `title`, `installId`, plus a `status: 'installed' | 'failed' | 'rolled_back'` the caller sets. There is no `success` key anywhere in the payload, so `perPatchResult?.success` was always `undefined` and `patchSuccess` collapsed to `overallSuccess` for **every** patch in the batch.

In a 13-patch/1-failure job, all 13 `patch_job_results` rows were written `status: 'failed'` even though 12 installed cleanly, and the `output`/`errorMessage` columns lost the per-patch error detail too.

The per-patch lookup was also broken independently: the agent sends the patch's server-side id as `id`, not `patchId` — so `r.patchId === patch.patchId` never matched, and the lookup only ever succeeded via the `externalId` branch.

The device *Patch Operation History* view (`GET /devices/:id/patches/history`) is unaffected — it reads the agent payload straight off `device_commands`.

## The fix

- Added `isPatchResultSuccessful()` — a shared per-patch success predicate: checks a boolean `success` first (defensive, for a future/alternate agent shape), then the agent's actual `status` field, then falls back to a caller-supplied default only when no per-patch entry matched at all.
- Reused this predicate for **both** the `anyPatchInstalled` reboot gate (added in #4265/#4228) and the new per-patch `patchSuccess`, per the issue's suggestion to factor the two together instead of duplicating the status-aware logic.
- Fixed the per-patch lookup to match on the agent's actual `id` field (kept `patchId` as a defensive fallback), alongside `externalId`.

## Tests

New suite in `apps/api/src/jobs/patchJobExecutor.test.ts`, `per-patch patch_job_results status is not collapsed to the batch status (#4267)`, driving the real per-device worker processor end to end (reusing the #4228 `runDeviceExecution` harness, extended to capture `patchJobResults.values()` calls):

| Test | Guards |
|---|---|
| records a mixed batch (one failed, two installed) with per-patch status, not all-failed | the reported bug — **red before the fix** |
| matches each per-patch result on the agent-reported `id` field, not a `patchId` field the agent never sends | the lookup-key bug — **red before the fix** |
| falls back to the batch status only when no per-patch entry matches at all | fallback behavior preserved |

Red-then-green confirmed by reverting just the source fix and re-running: 2 of the 3 new tests failed against the old code (as expected — the third exercises unchanged fallback behavior), then all 50 tests in the file passed after restoring the fix.

## Verification

- `npx vitest run src/jobs/patchJobExecutor.test.ts --pool=threads --maxWorkers=2` — 50 passed
- `npx tsc --noEmit` (apps/api) — clean
- `npx eslint src/jobs/patchJobExecutor.ts src/jobs/patchJobExecutor.test.ts` — clean

Server-side only; no schema, migration, or agent changes.

Closes #4267


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — pr-test-analyzer (async, subagent pool saturated ~38 parallel sibling issue-fixers), code-reviewer + silent-failure-hunter run inline (pool at concurrency limit for those two).

**Findings:**
- pr-test-analyzer: independently re-verified the red-before-fix claim (swapped in the pre-fix source, confirmed the same 2 of 3 new tests failed, all pass post-fix). Flagged 3 important coverage gaps (no direct test for a `rolled_back` per-patch entry, a matched entry with neither `success` nor `status`, and the per-patch `error` vs command-level error/output fallback) and 2 suggestions (multi-failure batch, a stale comment in the id-lookup test). All 3 important gaps + both suggestions addressed in `b472dd9a3` — 4 new tests added (rolled_back, malformed-entry fallback, error/output fallback, multi-failure batch) and the stale comment fixed.
- code-reviewer + silent-failure-hunter (inline): no critical or important issues. The fix is scoped correctly to the root cause (agent sends `id`/`status`, never `patchId`/`success`); the new `isPatchResultSuccessful` fallback ordering doesn't introduce any way a real failure could read as success or vice versa. The kept `r.patchId === patch.patchId` branch in the lookup is dead code today (agent never sends `patchId`) but is explicitly commented as a defensive fallback — not flagged as a blocker.

**Tests:** `apps/api/src/jobs/patchJobExecutor.test.ts` — 54/54 passed (`npx vitest run src/jobs/patchJobExecutor.test.ts --pool=threads --maxWorkers=2`). `npx tsc --noEmit` (apps/api) — clean. `npx eslint` on both changed files — clean.

PR #4408, opened 2026-09-01
fix(quotes): reject WebP quote images at upload (#3483)

### LINKED ISSUE #3483: WebP quote images silently vanish from the generated PDF
## Description
While recreating a large (21-page) external proposal as a Breeze quote, an image uploaded as WebP rendered correctly in the web and portal quote views but was missing entirely from the generated PDF, with no error surfaced anywhere in the UI.

## Root Cause
The quote image upload surface accepts WebP (per its own "PNG, JPEG, or WebP" copy), but the PDF renderer uses pdfkit, which can only embed PNG and JPEG. When the renderer calls `doc.image()` on a WebP source, it throws; the surrounding code appears to catch-and-continue, logging only a `console.error` and silently dropping the image from the output PDF. Since web/portal rendering doesn't go through pdfkit, the mismatch is invisible until a customer actually opens the PDF — by which point the quote has already been sent.

## Proposed Fix
One of:
1. Transcode WebP → PNG server-side at upload time (before it's ever handed to pdfkit), so all downstream consumers get a supported format, or
2. Reject WebP uploads for quote images at the API boundary (and update the UI copy/accept filter to match), or
3. If an image fails to embed during PDF render, fail the render visibly (error surfaced to the author) rather than silently continuing without it.

Option 1 is likely least disruptive since WebP is already advertised as accepted.

## Affected Files
- PDF render path that calls `doc.image()` for quote image blocks (pdfkit)
- Quote image upload route / accepted MIME types
- Related: PR #3481 (image cursor fix) touches nearby quote-image code but does **not** address this — noting only as related context, not a fix.

## Evidence
Reproduced 2026-08-12 on hosted US: WebP image block renders in web/portal quote view; same quote's PDF export omits the image with a blank/collapsed space where it should appear.

ToddHebebrand:
## Summary

Quote PDFs are rendered with pdfkit, which can only embed PNG/JPEG. Calling `doc.image()` on WebP bytes throws, and the render loop's catch-and-continue was silently dropping the image from the exported PDF — with no error surfaced anywhere — even though the upload surface advertised (and accepted) WebP. A customer could open a sent quote PDF with a blank gap where an image should be.

`apps/api` has no image-transcoding library in its dependencies (`sharp` is only a `devDependency` of `apps/docs`, not used by the API), so per the issue's own fallback guidance this implements **option 2: reject WebP at the upload boundary**, rather than adding a new dependency to transcode it.

## Changes

- `apps/api/src/services/quoteImageStorage.ts` — `fetchRemoteImage` (URL-import path) now rejects a sniffed `image/webp` mime with a clear `QUOTE_IMAGE_WEBP_REJECTED_MESSAGE`, distinct from the generic "not an image" case.
- `apps/api/src/routes/quotes/lifecycle.ts` — the multipart `POST /:id/images` route applies the same WebP rejection (415) for direct file uploads.
- `apps/web/src/components/billing/quotes/QuoteEditor.tsx` and `QuoteLineRows.tsx` — the file-picker `accept` attributes drop `image/webp` to match what the API now actually accepts.
- `apps/api/src/services/quotePdf.ts` — defense-in-depth: the two render-time `doc.image()` catch blocks that only did `console.error` (cover image, inline image block) now also call `captureException`, matching the third call site (line thumbnails) that already reported to Sentry. This covers any WebP bytes stored before this fix shipped — they still degrade gracefully instead of aborting the PDF, but the failure is no longer invisible.

## Tests

- `apps/api/src/services/quoteImageStorage.test.ts` — new test proving `fetchRemoteImage` rejects a real WebP buffer with the clear message (not just "not an image").
- `apps/api/src/routes/quotes/lifecycle.test.ts` — new `POST /:id/images — multipart file upload` describe block proving PNG is accepted, WebP is rejected with 415 + the clear message (and never persisted), and unrecognized bytes are rejected too. Also added the URL-import-path equivalent 415 test and fixed the now-stale "PNG, JPEG, or WebP" message in an existing test.
- `apps/api/src/services/quotePdf.test.ts` — replaced the old assertion that only proved `imageIntrinsicSize` can't *parse* WebP with a real `renderQuotePdf` test proving the render path actually degrades gracefully (no throw, surrounding content still renders) when WebP bytes reach `doc.image()` — the scenario that previously shipped as a silent gap.

## Test plan

- [x] `pnpm exec vitest run apps/api/src/services/quotePdf.test.ts apps/api/src/services/quotePdf.classicRegression.test.ts apps/api/src/services/quoteImageStorage.test.ts apps/api/src/routes/quotes/lifecycle.test.ts --pool=threads --maxWorkers=2` — 103/103 passed
- [x] `tsc --noEmit` clean in `apps/api` and `apps/web`
- [x] `eslint` clean on all touched files

Closes #3483

### DIFF
--- a/apps/api/src/middleware/bodyLimit.test.ts
+++ b/apps/api/src/middleware/bodyLimit.test.ts
@@ -148,9 +148,10 @@ describe('bodyLimitForPath', () => {
   });
 
   // Regression for #3482: quote block/cover image uploads. The UI advertises
-  // "PNG, JPEG, or WebP, up to 5 MB" and the route registers its own 5MB+64KB
-  // bodyLimit, but the global gate ran first and 413'd anything over 1MB with
-  // the generic "Request body too large".
+  // "PNG or JPEG, up to 5 MB" (WebP dropped in #3483 — pdfkit can't embed it)
+  // and the route registers its own 5MB+64KB bodyLimit, but the global gate
+  // ran first and 413'd anything over 1MB with the generic "Request body too
+  // large".
   it('carves out quote image uploads at the route 5MB cap (#3482)', () => {
     expect(bodyLimitForPath('/api/v1/quotes/11111111-1111-4111-8111-111111111111/images')).toEqual({
       rule: 'image-upload',
--- a/apps/api/src/routes/quotes/lifecycle.test.ts
+++ b/apps/api/src/routes/quotes/lifecycle.test.ts
@@ -38,14 +38,21 @@ vi.mock('../../jobs/quoteSendQueue', () => ({
   scheduleQuoteSend: vi.fn(),
   cancelQuoteSend: vi.fn(),
 }));
-vi.mock('../../services/quoteImageStorage', () => ({
-  writeQuoteImage: vi.fn(), readQuoteImage: vi.fn(), sniffImageMime: vi.fn(), MAX_QUOTE_IMAGE_SIZE_BYTES: 5 * 1024 * 1024,
-  fetchRemoteImage: vi.fn(),
-  QUOTE_IMAGE_WEBP_REJECTED_MESSAGE: "WebP images can't be used in quote PDFs — please upload a PNG or JPEG image instead.",
-  RemoteImageError: class RemoteImageError extends Error {
-    constructor(public reason: string, msg: string) { super(msg); this.name = 'RemoteImageError'; }
-  },
-}));
+// QUOTE_IMAGE_WEBP_REJECTED_MESSAGE comes from the REAL module (importActual)
+// rather than a hardcoded copy here — a hardcoded duplicate would let the
+// mock and the real constant drift apart silently, since every assertion
+// below compares the route's response against this same import.
+vi.mock('../../services/quoteImageStorage', async (importActual) => {
+  const actual = await importActual<typeof import('../../services/quoteImageStorage')>();
+  return {
+    writeQuoteImage: vi.fn(), readQuoteImage: vi.fn(), sniffImageMime: vi.fn(), MAX_QUOTE_IMAGE_SIZE_BYTES: 5 * 1024 * 1024,
+    fetchRemoteImage: vi.fn(),
+    QUOTE_IMAGE_WEBP_REJECTED_MESSAGE: actual.QUOTE_IMAGE_WEBP_REJECTED_MESSAGE,
+    RemoteImageError: class RemoteImageError extends Error {
+      constructor(public reason: string, msg: string) { super(msg); this.name = 'RemoteImageError'; }
+    },
+  };
+});
 vi.mock('./quotes', () => ({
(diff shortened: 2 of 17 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (subagent), plus a self-review pass covering the `pr-test-analyzer`/`silent-failure-hunter` lenses after concurrent-subagent capacity was exhausted (38 sibling issue-fixers running).

**Findings:** 4 raised → all addressed in `f40976676`.
- Self-review (before the code-reviewer ran): the `imageHelp` copy under `quotes.editor.addSection` and `quotes.editor.coverPage` still advertised "PNG, JPEG, or WebP" in **all 8 locales** even after the accept filters and API were fixed — directly contradicting the new upload behavior. Fixed in `b1ed858a8`.
- code-reviewer (important): `quotePdf.ts` cover-image draw called `doc.restore()` inside the `try`, so a draw failure (the exact WebP case this PR handles) left an unmatched `doc.save()` graphics-state push, corrupting the content stream for every subsequent draw. Moved `doc.restore()` into a `finally`.
- code-reviewer (important): the two new `captureException()` calls (the "no longer silent" half of the fix) were untested — `quotePdf.test.ts` had no `./sentry` mock at all. Added a spied mock and assertions that `captureException` fires exactly once for both the inline-image and cover-image WebP draw failures; the cover-image test also exercises the save/restore fix by checking content on both sides of the failed draw survives.
- code-reviewer (suggestion): `lifecycle.test.ts`'s `quoteImageStorage` mock hardcoded a second copy of `QUOTE_IMAGE_WEBP_REJECTED_MESSAGE`, so message-drift could never be caught. Switched to `importActual`.
- code-reviewer (suggestion): stale "PNG, JPEG, or WebP" comment in `bodyLimit.test.ts` fixed.

Noted, not addressed (correctly out of scope per the issue's own "Affected Files" list): catalog item images (`apps/api/src/routes/catalog/catalog.ts`, `catalogImageStorage.ts`, `CatalogItemEditorDrawer.tsx`) still accept WebP and can also be embedded into quote PDFs as line thumbnails via `loadCatalogImage`. They now at least get a `captureException` report via the same shared `quotePdf.ts` call site instead of vanishing invisibly, but the upload-time acceptance itself is unchanged. Worth a follow-up issue if it matters in practice.

**Tests:** affected suite green — `quotePdf.test.ts`, `quotePdf.classicRegression.test.ts`, `quoteImageStorage.test.ts`, `lifecycle.test.ts`, `bodyLimit.test.ts` = **123/123 passed** (`vitest run ... --pool=threads --maxWorkers=2`). `tsc --noEmit` clean in `apps/api` and `apps/web`. `eslint` clean on all touched files. `apps/web`'s `localeParity.test.ts` + `i18n.test.ts` (95/95) re-run clean after the 8-locale copy edits.

PR #4410, opened 2026-09-01
fix(api/ai): report LLM stream 'error' events to Sentry (#4329)

### LINKED ISSUE #4329: [API/AI] Stream 'error' events never reach Sentry — only the thrown-error path calls captureException
## Summary

In `apps/api/src/services/llm/openaiSessionManager.ts`, the stream loop's `case 'error':` branch publishes the failure to the client but never calls `captureException`. Provider-yielded stream errors therefore reach the user and are **invisible in Sentry**.

**Verified** against `origin/main`:

```ts
case 'error':
  hadError = true;
  session.eventBus.publish({ type: 'error', message: event.message });
  break;                                    // ~openaiSessionManager.ts:220
...
} catch (err) {
  hadError = true;
  captureException(err);                    // ~openaiSessionManager.ts:231
  session.eventBus.publish({ type: 'error', message: sanitizeErrorForClient(err) });
}
```

The two paths are the same failure surface reported two different ways. The `catch` — for errors *thrown* out of the iterator — reports to Sentry. The `case 'error'` — for errors the provider *yields* as a stream event — does not. Every other failure site in the file (lines ~145, ~169, ~249, ~267, ~288, ~362) calls `captureException`; this branch is the lone exception.

## Why now

The gap is pre-existing, but #4324 (merged: route `openaiCompatibleProvider` through the guarded `safeFetch`) materially increases what flows through this branch. Guarded fetch turns transport-layer conditions that previously surfaced as thrown exceptions — DNS pinning refusals, private-network/loopback blocks, redirect refusals, timeouts — into provider-yielded `error` events. Exactly the failures an operator most needs to see after a BYO-LLM misconfiguration are the ones that now go unreported.

That matters concretely for the self-hoster behavior change #4324 ships: a bare-metal install pointing `MCP_LLM_BASE_URL` at `http://localhost:8000/v1` is now refused, and with this gap the refusal produces no Sentry signal at all — only a message in the user's chat window.

## Suggested fix

Report the yielded error the same way the `catch` does, preserving the already-sanitized client message:

```ts
case 'error':
  hadError = true;
  captureException(new Error(`LLM stream error: ${event.message}`));
  session.eventBus.publish({ type: 'error', message: event.message });
  break;
```

Worth checking whether `event.message` can carry provider response text before it reaches Sentry — the `catch` path runs `sanitizeErrorForClient` for the client but sends the raw `err` to `captureException`, so the conventions here should be made deliberate rather than inherited. See also the Sentry scrubber's allowlist behavior when adding tags.

## Not checked

Whether any existing alert or dashboard already counts these via the event bus rather than Sentry.

ToddHebebrand:
## Summary

`apps/api/src/services/llm/openaiSessionManager.ts`'s stream loop had two
paths for the same failure surface reported two different ways: the `catch`
block (errors *thrown* out of the provider iterator) called `captureException`,
but the `case 'error':` branch (errors the provider *yields* as a stream
event) only published to the client's eventBus — never reaching Sentry.

`#4324` (merged) routed `openaiCompatibleProvider` through the guarded
`safeFetch`, which turns transport-layer conditions that previously surfaced
as thrown exceptions (DNS pinning refusals, private-network/loopback blocks,
redirect refusals, timeouts) into provider-yielded `error` events instead.
That materially increases what flows through the unreported branch — exactly
the failures an operator most needs to see after a BYO-LLM misconfiguration
(e.g. a bare-metal install pointing `MCP_LLM_BASE_URL` at
`http://localhost:8000/v1`, now refused by the egress guard) produced no
Sentry signal at all, only a message in the user's chat window.

## Fix

```ts
case 'error':
  hadError = true;
  captureException(new Error(`LLM stream error: ${event.message}`));
  session.eventBus.publish({ type: 'error', message: event.message });
  break;
```

Mirrors the sibling `catch` block's convention: that block calls
`captureException(err)` with no extra tags, so there's nothing to mirror onto
the Sentry scrubber's tag allowlist here either — kept intentionally plain
rather than inventing new tags that would silently get voided.

`event.message` is already the exact string the very same line publishes to
the client, so capturing it in Sentry doesn't expose anything beyond what the
client already sees — no separate sanitization step (like the `catch`
block's `sanitizeErrorForClient`) is needed.

Diff is surgical to the `case 'error':` branch — #4384 (session eviction) is
being fixed in the same file in parallel.

## Test plan

- Added a regression test asserting `captureException` is called when the
  provider yields a stream `error` event (previously would fail with
  "called 0 times" — verified red against the pre-fix code before applying
  the fix).
- `apps/api/src/services/llm/openaiSessionManager.test.ts` — 4/4 passed
  (`pnpm exec vitest run src/services/llm/openaiSessionManager.test.ts --pool=threads --maxWorkers=2`)
- `pnpm exec tsc --noEmit --project apps/api/tsconfig.json` — clean
- `eslint` on both touched files — clean

Closes #4329

### DIFF
--- a/apps/api/src/services/llm/openaiSessionManager.test.ts
+++ b/apps/api/src/services/llm/openaiSessionManager.test.ts
@@ -1,8 +1,36 @@
-import { describe, it, expect, afterEach } from 'vitest';
+import { describe, it, expect, afterEach, vi } from 'vitest';
+
+// #4329: the runTurn() background turn touches withDbAccessContext (turnCount
+// increment in its `finally`) even on the stream-error path under test here.
+// Same db-mock shape as aiKillState.test.ts: pass calls straight through so
+// runTurn's DB write is a no-op instead of hitting a real pool.
+vi.mock('../../db', () => ({
+  db: {
+    update: () => ({ set: () => ({ where: () => Promise.resolve() }) }),
+  },
+  runOutsideDbContext: vi.fn(<T,>(fn: () => T): T => fn()),
+  withDbAccessContext: vi.fn(async (_ctx: unknown, fn: () => Promise<unknown> | unknown) => fn()),
+}));
+
+vi.mock('../sentry', () => ({
+  captureException: vi.fn(),
+}));
+
+vi.mock('./historyBuilder', () => ({
+  buildMessagesFromHistory: vi.fn(async () => []),
+  ToolUseInHistoryError: class ToolUseInHistoryError extends Error {},
+}));
+
+vi.mock('../../config/validate', () => ({
+  getConfig: vi.fn(() => ({ MCP_LLM_MODEL: 'test-model' })),
+}));
+
 import { OpenAISessionManager } from './openaiSessionManager';
+import { captureException } from '../sentry';
 import type { RequestLike } from '../auditEvents';
 import type { OpenAICompatibleProvider } from './openaiCompatibleProvider';
 import type { AuthContext } from '../../middleware/auth';
+import type { LLMStreamEvent } from './types';
 
 // Mirrors the canonical shim in services/clientIp.test.ts.
 function makeContext(headers: Record<string, string | undefined>, remoteAddress?: string): RequestLike {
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** Manual self-review across the code-reviewer / silent-failure-hunter / pr-test-analyzer lenses (the `pr-review-toolkit` subagents were unavailable — concurrent subagent limit was saturated by the ~38 sibling issue-fixer agents running on this machine; retry is disallowed under that limit, so this pass was done directly against the diff instead).

**Findings:** 0 outstanding.
- `captureException` call matches the sibling `catch` block's convention exactly (no tags — nothing to mirror onto the Sentry scrubber allowlist since the catch block passes none either).
- `event.message` sent to Sentry is the identical string already published to the client one line below via `eventBus.publish(...)`, so this introduces no new data exposure.
- Confirmed the other `case 'error':`-shaped eventBus publishes in this file (session TTL eviction ~line 344, LRU capacity eviction ~line 381) are session-lifecycle events, not provider failures — correctly out of scope here (that's #4384's territory, being fixed in parallel in the same file).
- New test is not vacuous: verified red-first by stashing the source fix and re-running — the new assertion (`captureException` called 0 times) failed exactly as expected while the pre-existing client-facing "error still publishes" assertion kept passing, then restored the fix and confirmed green.

**Tests:** `apps/api/src/services/llm/openaiSessionManager.test.ts` — 4/4 passed (`pnpm exec vitest run src/services/llm/openaiSessionManager.test.ts --pool=threads --maxWorkers=2`). `pnpm exec tsc --noEmit --project apps/api/tsconfig.json` clean. `eslint` on both touched files clean.

PR #4411, opened 2026-09-01
fix(api): bound the agent software-inventory ingest's lock wait (#3925)

### LINKED ISSUE #3925: Bound the agent software-inventory ingest's lock wait (lock_timeout + 55P03 retry)
Follow-up carved out of #3911 (BREEZE-1N deadlock fix), filed so the residual risk that PR introduces is tracked rather than lost.

## Context

#3911 fixes a real `40P01` between the vulnerability correlation pass and `PUT /api/v1/agents/:id/software` by having `correlateOrg` pre-lock its `device_vulnerabilities` rows in ascending id order before taking any `software_inventory` FK lock.

That is the correct fix for the cycle, but it changes the failure mode of an overlap:

| | before #3911 | after #3911 |
|---|---|---|
| overlap outcome | `40P01`, one side killed | the ingest **waits** for the pass |
| observability | SQLSTATE -> Sentry + `retryOnTransientLockError` log | nothing, if the wait outlives the client |

## The residual risk

There is no `lock_timeout` or `statement_timeout` configured anywhere in `apps/api/src/db/` or `apps/api/src/config/`. (`idle_in_transaction_session_timeout` does not cover this — a backend blocked on a lock is not idle.) So the ingest waits unbounded, holding a request-long context and a pooled connection.

The agent's budget is **30s**: `sendInventoryData` in `agent/internal/heartbeat/heartbeat.go` wraps the PUT in `context.WithTimeout(..., 30*time.Second)` and its `http.Client` carries `Timeout: 30 * time.Second`.

So: if a correlation pass holds its locks longer than 30s, the agent gives up, the report is dropped for that cycle, and **the server logs no SQLSTATE** — no Sentry event, no retry log. A quiet version of the BREEZE-3 outcome.

Partial mitigation already in place: `withDbAccessContext`'s held-context tripwire warns and captures to Sentry when a context is held past `getHeldContextWarnMs()`, so an over-long correlation pass is visible from the *correlation* side even though the dropped report is not visible from the ingest side.

## Proposed fix

1. `SET LOCAL lock_timeout` at the top of the ingest transaction in `apps/api/src/routes/agents/inventory.ts` (~5s, so the existing 3 `retryOnTransientLockError` attempts fit inside the agent's 30s budget with room for the real work).
2. Add `55P03` (`lock_not_available`) to `isTransientLockError` in `apps/api/src/utils/pgErrors.ts`. Semantically identical to the `40P01`/`40001` it already covers — "you lost a lock race, the work was not applied, try again" — and it is a no-op for every existing caller, since `55P03` can only be raised where a `lock_timeout` is set and none is set today.

Net effect: the wait becomes bounded and, when it is exceeded, **loud** (retry log + Sentry) instead of silent.

## Why it was not done in #3911

It touches a different, customer-facing code path (the agent ingest) and changes the semantics of a shared error helper, so it wants its own review and its own test rather than riding along on the deadlock fix.

ToddHebebrand:
## Summary

Follow-up carved out of #3911 (BREEZE-1N deadlock fix). That PR fixed the `40P01` between the vulnerability correlation pass and `PUT /api/v1/agents/:id/software` by having `correlateOrg` pre-lock its `device_vulnerabilities` rows before taking any `software_inventory` FK lock — but that changes an overlap's failure mode from a `40P01` (loud) into an unbounded wait (silent, since nothing configures `lock_timeout`/`statement_timeout` anywhere in `apps/api/src/db/` or `apps/api/src/config/`). If the wait outlives the agent's 30s HTTP budget (`sendInventoryData` in `agent/internal/heartbeat/heartbeat.go`), the report is dropped for that cycle with no SQLSTATE, no Sentry event, and no retry log.

This closes that gap:

- `apps/api/src/services/softwareInventoryObservations.ts`: sets a 5s `lock_timeout` (`INVENTORY_LOCK_TIMEOUT_MS`) at the top of the ingest transaction, via the existing `tightenLockTimeout` helper (`apps/api/src/db/lockTimeout.ts`) already used by `deviceDeletion.ts` and `catalogService.ts`. 5s leaves room for the existing `retryOnTransientLockError`'s 3 attempts (worst case 15s) inside the agent's 30s budget. Re-applied inside `db.transaction`'s callback since each retry opens a fresh savepoint and `ROLLBACK TO SAVEPOINT` undoes a `SET LOCAL` issued after it.
- `apps/api/src/utils/pgErrors.ts`: teaches `isTransientLockError` about SQLSTATE `55P03` (`lock_not_available`) — semantically identical to the `40P01`/`40001` it already covers ("you lost a lock race, the work was not applied, try again"), and a no-op for every existing caller since `55P03` can only be raised where a `lock_timeout` is set, which none was before this PR.
- `apps/api/src/services/vulnerabilityCorrelation.ts`: updates the comment that tracked this as "the fuller fix... tracked in #3925" to reflect that it's now implemented.

## Test plan

- [x] `apps/api/src/utils/pgErrors.test.ts` (unit, single-fork): added coverage for `55P03` classification (bare + Drizzle-`.cause`-wrapped) and confirmed `57014` (`statement_timeout` cancellation) stays non-transient; added `retryOnTransientLockError` cases for a `55P03` that clears on retry and one that exhausts all attempts and rethrows. 23/23 passed.
- [x] `apps/api/src/services/softwareInventoryObservations.test.ts` (unit, single-fork): unchanged, 15/15 still passing.
- [x] `npx tsc --noEmit` on `apps/api` — clean, no errors.
- [x] `pnpm exec eslint` on all 4 changed files — clean.
- Not run: DB-backed integration coverage of the actual `55P03` firing under real Postgres (would need a real overlapping lock scenario, similar in spirit to `lockTimeoutBounds.integration.test.ts`) — left for a maintainer/CI run since this worktree shares Docker infra with ~38 other parallel agents right now and I didn't want to add contention or risk cross-talk on the shared test Postgres containers.

Closes #3925

### DIFF
--- a/apps/api/src/services/softwareInventoryObservations.test.ts
+++ b/apps/api/src/services/softwareInventoryObservations.test.ts
@@ -1,8 +1,31 @@
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import type { LegacySoftwareInventoryReport, SoftwareInventoryObservationV2 } from '@breeze/shared';
+
+// Hoisted so the `vi.mock` factories below (themselves hoisted above these
+// imports by Vitest) can close over them. Only `ingestSoftwareInventoryReport`
+// touches these — `decideSoftwareInventoryAcceptance` and
+// `replaceSoftwareInventoryProjection` take a `tx` argument directly and never
+// reach the module-level `db` import, so mocking it here doesn't affect them.
+const { transactionMock, tightenLockTimeoutMock } = vi.hoisted(() => ({
+  transactionMock: vi.fn(),
+  tightenLockTimeoutMock: vi.fn(async () => null),
+}));
+
+vi.mock('../db', () => ({
+  db: { transaction: transactionMock },
+  runOutsideDbContext: (fn: () => unknown) => fn(),
+  withSystemDbAccessContext: (fn: () => unknown) => fn(),
+}));
+
+vi.mock('../db/lockTimeout', () => ({
+  tightenLockTimeout: tightenLockTimeoutMock,
+}));
+
 import {
   decideSoftwareInventoryAcceptance,
   replaceSoftwareInventoryProjection,
+  ingestSoftwareInventoryReport,
+  INVENTORY_LOCK_TIMEOUT_MS,
 } from './softwareInventoryObservations';
 
 const item = { name: 'Google Chrome', version: '127', vendor: 'Google LLC' };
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, silent-failure-hunter, pr-test-analyzer)

**Findings:** 5 raised → all addressed in c77e04464; 0 outstanding.
- code-reviewer: factually inaccurate claim in `pgErrors.ts`'s `isTransientLockError` comment (other transactions already set `lock_timeout` before #3925, just not through `retryOnTransientLockError`) — corrected.
- code-reviewer + silent-failure-hunter (raised independently): the "worst case 3×5s=15s" framing overclaimed — `lock_timeout` bounds each lock *acquisition*, not a whole multi-row statement, and `replaceSoftwareInventoryProjection`'s `FOR UPDATE` can lock several `device_vulnerabilities` rows in one statement. Comment corrected to state the guarantee that actually holds (no single acquisition waits forever → always eventually bounded and loud, not silently unbounded) and to flag `tightenStatementTimeout` pairing as a deliberate, noted follow-up rather than a silent gap.
- silent-failure-hunter: minor comment inaccuracy about when the outer (`withDbAccessContext`-opened) transaction actually commits relative to retry attempts — corrected.
- pr-test-analyzer: no test proved `tightenLockTimeout` was actually wired into `ingestSoftwareInventoryReport` (position + bound value) — added two unit tests in `softwareInventoryObservations.test.ts` that mock `db.transaction`/`tightenLockTimeout` to assert the call happens first, with the right bound, and re-applies on every retry attempt (SAVEPOINT rollback undoes `SET LOCAL`).
- silent-failure-hunter also flagged that a *successful* retry only reaches Sentry as a breadcrumb via `console.warn`, not a captured event — this is `retryOnTransientLockError`'s pre-existing, shared behavior (already true for 40P01/40001 since #3911), not something this PR introduces; left out of scope as a broader observability change affecting other callers (`vulnerabilityJobs.ts`, `invoiceService.ts`) that wants its own review.

**Tests:** `apps/api/src/utils/pgErrors.test.ts` (23/23), `apps/api/src/services/softwareInventoryObservations.test.ts` (17/17, +2 new), `apps/api/src/routes/agents/inventory.test.ts` (13/13) — all single-fork via `vitest run --pool=threads --maxWorkers=2`, all green. `tsc --noEmit` on `apps/api` clean. `eslint` on all 5 changed files clean.

PR #4419, opened 2026-09-01
perf(api): roll up each metric source in one pass instead of once per metric (#4341)

### LINKED ISSUE #4341: [API] Process-sample rollup rescans the same window 7× per org — dominates DB CPU
## Description
`pg_stat_statements` on US production (31.5h window, 2026-09-01): the process-sample rollup CTE (`rollupRawProcessSampleMetric`) ran ~107k times at 1.5–2.0s mean, ~168,600s total execution — i.e. ~1.5 query-seconds per wall-second, continuously occupying ~150% of the managed DB's single vCPU, with 900+ GB of shared buffer reads. It is the top DB load by a wide margin and the main reason interactive dashboard queries contend for buffers (observed `BufferMapping` waits; cache hit ratio 79.7%).

## Root Cause
`rollupMetricsForOrg` (`apps/api/src/services/metricRollups.ts:525-531`) loops `PROCESS_SAMPLE_ROLLUP_SOURCES` (7 metrics) and issues one full CTE per metric, every 5 minutes per org. Each statement rescans the identical `device_process_samples` window (the `process_devices` DISTINCT scan + `sample_values` scan), so the same rows are read ~7× per cycle. Same pattern applies to the 10-metric `DEVICE_METRIC_ROLLUP_SOURCES` loop over `device_metrics`, though at lower measured cost.

## Proposed Fix
Compute all metrics for a source table in a single pass: one scan of the window producing all 7 (resp. 10) aggregate rows via `unnest`/lateral values or conditional aggregation, then one `INSERT ... ON CONFLICT`. Expected ~7× reduction in rollup DB time. Batching multiple orgs per statement is a further option but the single-pass change is the big win.

## Affected Files
- `apps/api/src/services/metricRollups.ts` (`rollupRawProcessSampleMetric`, `rollupRawDeviceMetric`, the source loops at 525-531)

ToddHebebrand:
## Problem

`rollupMetricsForOrg` looped its source lists and issued **one full CTE per metric** — 10 statements rescanning the `device_metrics` window and 7 rescanning the `device_process_samples` window, per org, every 5 minutes.

On US production, `pg_stat_statements` over a 31.5h window (2026-09-01) showed the process-sample CTE alone at ~107k calls at 1.5–2.0s mean, ~168,600s total execution — roughly **1.5 query-seconds per wall-second**, continuously occupying ~150% of the managed DB's single vCPU, with 900+GB of shared buffer reads. It was the top DB load by a wide margin and the main reason interactive dashboard queries contend for buffers (`BufferMapping` waits; cache hit ratio 79.7%).

`device_process_samples` has no `(org_id, timestamp)` index, so each of those 7 statements re-reads the same window from a table holding weeks of retention.

## Change

Each source table is now read **once per pass**. Both raw passes aggregate every series in ONE grouped pass over the bucket grid and unpivot the aggregates afterwards via `CROSS JOIN LATERAL (VALUES …)`. The process pass additionally expands each sample's `top_processes` array once, through a shared LATERAL aggregate, instead of once per array-derived series.

**24 statements per pass → 9.** (10 + 7 + 1 raw + 6 derived → 1 + 1 + 1 + 6.) The SNMP and derived passes are untouched.

Distinct from #4276 — that PR (#4285) split the pass into per-statement DB contexts. This one reduces how many statements there are. The per-statement context/escape/label contract is preserved unchanged.

## Why output is identical

- **Aggregates are unpivoted, not inputs.** A row-wise unpivot needs one common `double precision` value column, which would turn `sum(real)` — Postgres accumulates that in `real` — into a float8 sum. It also multiplies the bucket grid by the series count; the inflated `generate_series` row estimate then trips `jit_above_cost` and adds ~1.1s of JIT compilation to a sub-100ms query (measured, which is why the shape you see here is the one that shipped).
- **`metric_devices` keeps its per-metric `IS NOT NULL` device filter.** Half the `device_metrics` columns are nullable, and a device that never reports (say) `process_count` must still produce **no** `process_count` rollups, not a window's worth of `isGap` rows. The shared grid is built from the union of those device sets and re-narrowed per series by the closing join.
- **The per-column `IS NOT NULL` predicate is dropped from the bucket join**, because every aggregate (including `percentile_cont`) already skips NULL inputs — so each series' numbers are untouched while all ten share one join.
- The shared `top_processes` aggregate is value-identical to the old per-series subqueries: float8 addition of `0` is exact, so `sum(CASE … ELSE 0)` and `coalesce(sum(x), 0)` agree; the old `max` CASE already produced NULL for non-numbers; and an empty array aggregates over zero rows to NULL → `0` in both forms.

## Verification

**Equivalence (real Postgres 16).** Replayed the old 24-statement pass and the new 9-statement pass over identical randomized fixtures in a throwaway database and diffed every `metric_rollups` row (row set via `EXCEPT ALL` both ways, plus a keyed `FULL OUTER JOIN` comparing all value columns with `IS NOT DISTINCT FROM`). Fixtures covered empty and non-numeric `top_processes` payloads, optional `diskBps`/`netBps` absent, all-NULL nullable columns, mixed-NULL devices, sparse gap buckets, out-of-window rows on both sides, foreign-org rows, and a second pass to exercise the `ON CONFLICT DO UPDATE` replay path. **Identical row sets and bit-identical values across 5 seeds × 2 session timezones (10 runs).**

**Performance (same harness, with production's index set and a table far larger than the window).** Buffer blocks reading the source window, per pass:

| source | old | new | |
|---|---|---|---|
| `device_metrics` | 19730 blocks (154MB), 10 statements | 2057 blocks (16MB), 1 statement | **9.6× fewer** |
| `device_process_samples` | 16380 blocks (128MB), 7 statements | 2340 blocks (18MB), 1 statement | **7.0× fewer** |

Wall time on that local fixture is 2.5×/2.2× because the `metric_rollups` upsert side is unchanged (same rows written) and dominates locally; on prod the read side is what is saturating the vCPU.

**Tests.**
- `apps/api/src/services/metricRollups.test.ts` — new `#4341 single-pass raw rollups` block (5 tests) written red-first against the old code: exactly one statement per raw source table, all 10 device series with their own `metric_type` from that one scan, the per-metric non-null device filter still present, all 7 process series from one scan, and `jsonb_array_elements(dps.top_processes)` appearing exactly once. Existing SQL assertions were re-anchored from fragile positional `mock.calls[n]` indices to context labels.
- `apps/api/src/__tests__/integration/metricRollups.integration.test.ts` — 2 new tests against a real DB: every device-metric and process-sample series is emitted from one pass with its own `metric_type` and its own values, and a column that is NULL across the whole window produces no rows at all (not gap rows). Mutation-checked: deleting the per-series narrowing from the closing join turns both red.
- 25 unit tests and 8 integration tests green; `tsc --noEmit` and `eslint` clean.

Closes #4341

### DIFF
--- a/apps/api/src/__tests__/integration/metricRollups.integration.test.ts
+++ b/apps/api/src/__tests__/integration/metricRollups.integration.test.ts
@@ -50,7 +50,18 @@ async function insertMetric(options: {
   deviceId: string;
   timestamp: Date;
   cpuPercent: number;
+  /** #4341: leave every nullable column NULL, as an agent that reports no disk/network/process counters does. */
+  nullableColumnsNull?: boolean;
 }): Promise<void> {
+  const nullable = options.nullableColumnsNull
+    ? {
+      diskReadBps: null, diskWriteBps: null,
+      bandwidthInBps: null, bandwidthOutBps: null, processCount: null,
+    }
+    : {
+      diskReadBps: 100, diskWriteBps: 200,
+      bandwidthInBps: 300, bandwidthOutBps: 400, processCount: 50,
+    };
   await (getTestDb() as any).insert(deviceMetrics).values({
     orgId: options.orgId,
     deviceId: options.deviceId,
@@ -60,11 +71,7 @@ async function insertMetric(options: {
     ramUsedMb: 2048,
     diskPercent: 40,
     diskUsedGb: 120,
-    diskReadBps: 100,
-    diskWriteBps: 200,
-    bandwidthInBps: 300,
-    bandwidthOutBps: 400,
-    processCount: 50,
+    ...nullable,
   });
 }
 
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** checklist run inline (the `pr-review-toolkit:code-reviewer` subagent was killed mid-run by an API rate limit and its output was lost, so the same checklist — SQL equivalence, NULL semantics, `sql.raw` injection surface, unpivot cardinality, planner/memory risk, assertion vacuity — was executed directly rather than delegated).

**Findings:** 1 raised → addressed in `9acdcfce8`; 0 outstanding.

- **Vacuous assertion (fixed).** `expect(deviceSql).toContain(metricType)` in `metricRollups.test.ts` was substring-matching the serialized SQL, and 3 of the 5 metric types are substrings of their own metric name — `cpu` ⊂ `cpu_percent`, `disk` ⊂ `disk_percent`, `process` ⊂ `process_count`. It passed even with `metric_type` dropped from the unpivot entirely. Replaced with an assertion that the compiled parameter list binds each series as `(metric_type, metric_name)` adjacently. Mutation-checked: binding a constant `'cpu'` as every series' `metric_type` now fails the test; under the old assertion it did not.

**Checks that came back clean:**

- **`sql.raw` injection surface** — every `sql.raw` argument traces to a module-level `as const` source list, a literal separator, or `PROCESS_SAMPLE_PROCESS_AGGREGATE_SQL`. `options.orgId` and the window bounds are bound parameters at all 9 call sites; nothing caller-controlled reaches `sql.raw`.
- **Shared `top_processes` LATERAL ≡ the old per-series scalar subqueries** — verified on Postgres 16 against adversarial payloads: empty array, absent keys, and `cpu` as every non-number jsonb type (`null`, `true`, array, object, string), plus negatives/zero/1e10. All six aggregates `IS NOT DISTINCT FROM` the old form on every row.
- **Dropping the per-column `IS NOT NULL` from the bucket join** — verified with the shipped grid shape (grid LEFT JOIN rows): `avg`/`min`/`max`/`percentile_cont`/`sum`/`count`/`gap_seconds`/`isGap` all match, including a bucket where the column is NULL on every row (`sample_count` 0, `isGap` true) and a bucket with no rows at all.
- **Aggregate-unpivot vs input-unpivot** — confirmed the design rationale rather than assuming it: `pg_typeof(sum(real))` is `real`, and over 300 values `sum(x)::float8` = 4515 vs `sum(x::float8)` = 4515.000000022352. An input-unpivot would have silently changed `sum_value` for the four `real` columns.
- **Unpivot cardinality** — `CROSS JOIN LATERAL (VALUES …)` emits exactly one row per series per grouped bucket; the closing `JOIN metric_devices` re-narrows per series. Mutation-checked earlier: deleting `AND md.metric_name = series.metric_name` turns 5 tests red, both new ones included.
- **Planner/memory** — the grid keeps its pre-change cardinality (that was the point of unpivoting aggregates rather than inputs); only the aggregate list widens. Measured: no JIT on the `device_metrics` statement, 25ms on the process statement.

**Tests:** 25 unit green (`vitest run src/services/metricRollups.test.ts src/jobs/metricRollups.test.ts`); 8 integration green locally against a private stack **and in CI** — shard 1/4 logs `✓ src/__tests__/integration/metricRollups.integration.test.ts (8 tests)`, so the new cases really ran rather than being silently skipped. `tsc --noEmit` and `eslint` clean.

**CI:** 64 SUCCESS, 1 SKIPPED (`Main Red Alert`, main-only). An earlier run had one `Test Web` failure in `TicketWorkbench.test.tsx` — this PR touches only `apps/api/**`, that file passes locally on this branch (99/99), and it passed on re-run. Flake, not caused by this change.

PR #3627, opened 2026-08-16
fix(web,api): name TICKETS_INBOUND_DOMAIN in the inbound-email hint for self-hosters (#3599)

ToddHebebrand:
## What

When no platform inbound domain is configured, the Inbound email card said:

> The platform inbound domain isn't configured yet. Contact your administrator.

On hosted Breeze that's correct — the operator is us. On a self-hosted instance the reader **is** the administrator, so it told them to contact themselves and named nothing actionable. That's exactly how the Discord reporter behind #3586 concluded email-to-ticket was Microsoft 365-only.

## How

`apps/web` had no way to tell hosted from self-hosted — `IS_HOSTED` is API-only and was never surfaced to the client. So `GET /ticket-config` now returns `inbound.isHosted` alongside the existing `inbound.domainConfigured`, and the card branches on it:

- **Hosted (or older API):** unchanged copy. "Contact your administrator" is the right advice and a variable name would be noise.
- **Self-hosted:** a new string that names `TICKETS_INBOUND_DOMAIN`, links the [environment docs section](https://docs.breezermm.com/deploy/environment/#inbound-email-to-ticket-mailgun) added in #3586, and points at the Microsoft 365 shared-mailbox path as the alternative that needs no inbound domain.

The two defaults deliberately run in opposite directions:

- **Client:** `isHosted` is optional and only an explicit `false` takes the self-hosted branch, so an older API never shows a self-host variable name to a hosted partner.
- **Server:** `IS_HOSTED` unset parses as `false` → self-hosted → the actionable message. Naming the variable to a hosted partner is noise; withholding it from a self-hoster is the dead end this fixes.

## Locales

`domainNotConfiguredSelfHosted` is added to **all 8 bundles** in this PR — locale parity enumerates the locales directory, so an `en`-only change reds the suite. Each translation preserves the `<var>`/`<docs>` placeholder tags and the literal `TICKETS_INBOUND_DOMAIN` / `Microsoft 365` names the parity guard pins.

## Docs

Both pages that quoted the old wording verbatim are updated: `deploy/environment.mdx` and `features/ticketing.mdx`.

## Scope note

#3598 (same message shown when only the M365 path is configured) is **not** addressed here — this PR is the genuinely-unconfigured case only, as the issue asks.

## Tests

- `apps/api/src/services/ticketConfigService.test.ts` — `isHosted` reported for `IS_HOSTED` true / false / unset.
[...]

PR #4420, opened 2026-09-01
fix(api): default new partners' inbound email-to-ticket to off (#3608)

### LINKED ISSUE #3608: Decide the inbound email-to-ticket default for NEW partners
Follow-up from #3597 / PR #3606, which made `settings.ticketing.inbound.enabled` an enforced gate. That PR deliberately preserved existing behavior and left the *product* default for new partners undecided.

## Current state (after #3606)

`partnerCreate.ts:66-81` inserts a partner with no `settings.ticketing.inbound` object at all. Both readers default an absent flag to permissive:

- `loadPartnerInboundPolicy` (`apps/api/src/services/inboundEmail/resolveOrg.ts:154`) — `enabled: inbound.enabled !== false`
- `getTicketConfig` (`apps/api/src/services/ticketConfigService.ts:396`) — same rule

So a brand-new partner is **on**: the moment the platform inbound domain exists, `<slug>@<domain>` accepts mail for them and converts it to tickets. #3597 called this out directly:

> once the platform inbound domain is configured, `<slug>@<domain>` accepts mail for every partner immediately, since the derived address exists regardless of the toggle ... it is a surprising default-on surface.

The default-on is not a data-exposure issue — unknown senders still default to `quarantine` — but it is a surface nobody opted into.

## The decision

**Should a newly created partner start with inbound email-to-ticket off?**

- **A — Keep default-on (status quo).** Pro: zero-config; an MSP that publishes the address gets tickets immediately, no hidden switch to discover. Con: every partner has a live mail-ingestion surface they never enabled, and the "surprising default-on" complaint stands.
- **B — Default new partners off.** `partnerCreate` writes an explicit `enabled: false`; the absent-flag default stays `true` purely as the upgrade path for pre-#3606 partners. Pro: clean opt-in semantics, matches what the card used to *display*. Con: an operator who wires up a Mailgun route before finding the toggle has their mail **silently discarded** (an `ignored` audit row, no bounce) — a worse failure than the one being fixed, unless it's paired with UI that makes the off-state obvious on the Inbound email card.

If B is chosen, the absent → `true` fallback must stay for back-compat, which means the loader's default and the new-partner default deliberately disagree. That is fine but should be commented, or a later reader will "fix" it back.

An advisor-quorum read (codex, `xhigh`) favored B with an explicit write at creation time, on the grounds that upgrade compatibility and the steady-state default are separate concerns and shouldn't share one fallback.

## Recommendation

**B, but only bundled with the card work in #3598/#3599.** The discard-silently failure mode is only acceptable if the Inbound email card states plainly that ingestion is off — which is exactly the copy those two issues are already reworking. Shipping B on its own would trade a surprising default-on for a silent default-drop.

## Related

- #3597 — the enforcement gate (fixed in #3606)
- #3598 — card shows "platform inbound domain isn't configured" when only the M365 path is in use
- #3599 — "Contact your administrator" is a dead end for self-hosters

ToddHebebrand:
## Summary

- Implements the Option B decision recorded on #3608 (2026-08-16): `createPartner` now writes an explicit `settings.ticketing.inbound.enabled: false` at partner-creation time, so new partners must opt in to inbound email-to-ticket instead of silently inheriting the readers' absent-flag-means-true default.
- The readers (`loadPartnerInboundPolicy` in `apps/api/src/services/inboundEmail/resolveOrg.ts` and `getTicketConfig` in `apps/api/src/services/ticketConfigService.ts`, both `enabled !== false`) are left unchanged and now serve solely as the upgrade path for partners created before this default existed — commented in `partnerCreate.ts` so a future reader doesn't "fix" it back into agreement, per the issue's explicit warning.
- No backfill of existing partners' `settings` — scope is new-partner creation only, as specified in the issue.
- The prerequisite card-visibility work (#3598, closed) and copy fix (#3599 / PR #3627, merged) that make the off-state visible to operators are already shipped, satisfying the issue's "only bundled with the card work" condition.

## Test plan

- [x] New unit test in `apps/api/src/services/partnerCreate.test.ts`: asserts the partner insert's `settings` matches `{ ticketing: { inbound: { enabled: false } } }`. Confirmed red before the fix, green after.
- [x] `pnpm --filter @breeze/api exec vitest run src/services/partnerCreate.test.ts --pool=threads --maxWorkers=2` — 6/6 passed.
- [x] `pnpm --filter @breeze/api exec vitest run src/services/ticketConfigService.test.ts src/services/inboundEmail/resolvePartner.test.ts --pool=threads --maxWorkers=2` — 66/66 passed (no regression in the readers this decision deliberately leaves unchanged).
- [x] `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api` — clean.

Closes #3608

### DIFF
--- a/apps/api/src/services/partnerCreate.test.ts
+++ b/apps/api/src/services/partnerCreate.test.ts
@@ -177,6 +177,25 @@ describe('createPartner', () => {
     expect(userCall.values.passwordHash).toBe('hashed');
   });
 
+  // Issue #3608: new partners must opt IN to inbound email-to-ticket rather
+  // than inheriting the readers' absent-flag-means-true fallback (which
+  // exists solely as the pre-#3606 upgrade path for existing partners).
+  it('writes an explicit settings.ticketing.inbound.enabled=false for new partners', async () => {
+    await createPartner({
+      orgName: 'Acme',
+      adminEmail: 'alex@acme.com',
+      adminName: 'Alex',
+      passwordHash: 'hashed',
+      origin: { mcp: false },
+      status: 'active',
+    });
+
+    const partnerCall = insertCalls.find((c) => (c.table as any).__t === 'partners')!;
+    expect(partnerCall.values.settings).toMatchObject({
+      ticketing: { inbound: { enabled: false } },
+    });
+  });
+
   it('inserts six system ticket_statuses rows inside the transaction', async () => {
     await createPartner({
       orgName: 'Acme',
--- a/apps/api/src/services/partnerCreate.ts
+++ b/apps/api/src/services/partnerCreate.ts
@@ -83,6 +83,16 @@ export async function createPartner(
         mcpOriginUserAgent: mcpOrigin ? (input.origin as { userAgent?: string }).userAgent ?? null : null,
         signupIp: !mcpOrigin ? (input.origin as { ip?: string }).ip ?? null : null,
         signupUserAgent: !mcpOrigin ? (input.origin as { userAgent?: string }).userAgent ?? null : null,
+        // Issue #3608 (decision: Option B, 2026-08-16): new partners opt IN to
+        // inbound email-to-ticket rather than inheriting the readers'
+        // "absent flag reads as true" fallback. That fallback
+        // (`loadPartnerInboundPolicy` in inboundEmail/resolveOrg.ts and
+        // `getTicketConfig` in ticketConfigService.ts, both `enabled !== false`)
+        // stays as-is — it exists solely as the pre-#3606 upgrade path for
+        // partners created before this default existed. Do NOT "fix" the
+        // readers to match this new-partner default; the two are deliberately
+        // different concerns. Do NOT backfill existing partners' settings.
+        settings: { ticketing: { inbound: { enabled: false } } },
       })
       .returning();
 


ToddHebebrand:
**Review run:** manual inline checklist (the \`pr-review-toolkit:code-reviewer\` subagent could not be spawned — concurrent subagent pool was at its cap from sibling issue-fixers running in parallel this session; ran the equivalent checklist by hand instead: diff read, call-site audit for both \`createPartner\` callers, jsonb-override safety check, lint, and the test-quality check below).

**Findings:** 0 blocking. 1 informational (no change made, out of scope per the issue and task instructions):
- \`apps/api/src/routes/orgs.ts:474\` (\`POST /partners\`, \`requireScope('system')\`) is a separate platform-admin partner-creation path that does **not** go through \`createPartner()\` — it inserts with whatever \`settings\` the caller supplies (\`z.any().optional()\`), or the jsonb column default \`{}\` if omitted. A partner created via that admin endpoint without explicit settings will still fall back to the readers' absent-flag-means-true default. The issue (#3608) and task scope explicitly target \`partnerCreate.ts\` only (shared by \`/register-partner\` and the MCP \`create_tenant\` bootstrap — both covered by this fix), so this is left as-is; flagging in case a follow-up is wanted for that admin surface.

**Tests:** \`vitest run src/services/partnerCreate.test.ts --pool=threads --maxWorkers=2\` — 6/6 passed (new test confirmed red before the fix, green after). \`vitest run src/services/ticketConfigService.test.ts src/services/inboundEmail/resolvePartner.test.ts --pool=threads --maxWorkers=2\` — 66/66 passed (no regression in the readers this decision deliberately leaves unchanged). \`NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api\` — clean. \`eslint\` on both changed files — clean.

**Status:** review-clean (0 blocking findings), awaiting maintainer merge.

PR #4421, opened 2026-09-01
fix(api): unique index + atomic upserts for automation_policy_compliance (#4122)

### LINKED ISSUE #4122: Add unique constraint + dedupe for automation_policy_compliance (policy_id, device_id)
Follow-up from wave 3.5c (PR #4117, final whole-branch review).

`automation_policy_compliance` has no unique constraint on `(policy_id, device_id)`, and `policyEvaluationService.ts` upserts via select-then-insert (not atomic), so concurrent evaluations of the same policy can create duplicate rows for one device.

Consequences today:
- The wave-3.5c reconcile guard in `policyAlertBridge.handlePolicyViolation` reads this table; with duplicates it could read a stale row. PR #4117 mitigates with `.orderBy(desc(updatedAt))`, but that is a mitigation, not a guarantee.
- The evaluation upsert itself can keep writing to an arbitrary one of the duplicates.

Work:
1. Dedupe migration keeping the freshest row per `(policy_id, device_id)` (also consider the config-policy shape keyed `(config_policy_id, config_item_name, device_id)` — same select-then-insert pattern, same exposure). Follow the forensic-trail rule: `GET DIAGNOSTICS` + `RAISE WARNING` with the count.
2. Unique index per shape (partial indexes since `policy_id` XOR `config_policy_id` is nullable).
3. Convert the upserts to `onConflictDoUpdate`.

ToddHebebrand:
## The bug

`automation_policy_compliance` had **no uniqueness on either of the two shapes it stores**, and `policyEvaluationService` wrote it with a non-atomic select-then-insert. Two concurrent evaluations of the same policy — scheduled sweep + manual "Evaluate now", or two workers on one policy — both saw "no row" and both inserted, so duplicates accumulated per key.

Readers then picked an arbitrary one:
- `policyAlertBridge.handlePolicyViolation`'s reconcile guard could read a stale row. #4117 mitigated this with `ORDER BY updated_at DESC`, which is a tiebreak, not a guarantee.
- The evaluation upsert itself kept writing to whichever duplicate its unordered `SELECT` returned, so status could ping-pong between rows.

## The fix

**Migration `2026-09-29-100000-automation-policy-compliance-unique.sql`** — dedupes both shapes keeping the freshest row per key (`updated_at DESC, id DESC`), then adds two **partial** unique indexes:

| Index | Key | Predicate |
|---|---|---|
| `apc_policy_device_uq` | `(policy_id, device_id)` | `policy_id IS NOT NULL` |
| `apc_config_policy_item_device_uq` | `(config_policy_id, config_item_name, device_id)` | both `IS NOT NULL` |

Partial because the key columns are nullable and exactly one axis is populated per row.

Notes on the migration:
- **Sets `breeze.scope` to `system` first.** The table is FORCE RLS and `breeze_current_scope()` defaults to `'none'` (0012 superseded 0008's `'system'` default), while autoMigrate sets no scope. On the containerised deployment `DATABASE_URL` is a superuser so RLS is bypassed — but on managed Postgres (DigitalOcean/RDS) the admin role is not, and there the DELETEs would match zero rows and `CREATE UNIQUE INDEX` would then abort on the survivors, refusing the API's boot. Same one-liner `2026-04-13-fix-uuid-hostnames.sql` and ~10 later migrations use. (Caught in review; see the verification section.)
- Each dedupe reports its count via `GET DIAGNOSTICS` + `RAISE WARNING` **unconditionally** (no `IF n > 0` guard) — these rows drive compliance state and alert reconciliation, so a zero is evidence too.
- The config-shape dedupe is scoped to rows where **both** key columns are populated. NULLs never collide in a btree unique index, so a row without a `config_item_name` cannot be keyed and must not be deleted on a key that cannot be formed.
- Dedupe and index build share one transaction, so no racing writer can slip a duplicate in between. That also rules out `CONCURRENTLY`, which cannot run inside `autoMigrate`'s wrapper and would finish `INVALID` if a duplicate landed mid-build.
- New file only; no shipped migration edited.

**Service** — both write sites become `ON CONFLICT DO UPDATE` upserts, extracted into `buildPolicyComplianceUpsert` / `buildConfigPolicyComplianceUpsert`.
- `remediation_attempts` stays out of every `SET` list, matching the `UPDATE` these replace.
- The automation-policy site **keeps** its existing `SELECT` — it still supplies `previousStatus` for the emitted events; it just no longer decides insert-vs-update.
- Each `targetWhere` reproduces its index predicate verbatim. Postgres only infers a partial index as an `ON CONFLICT` arbiter when the statement's predicate implies the index's, so the three copies (migration, Drizzle schema, service) must move together.

## Verification

**Unit:** 181 tests green; `tsc --noEmit -p apps/api` clean; eslint clean.

Because the failure mode is "wrong arbiter" rather than "wrong function called", the unit test asserts the **compiled SQL** (`.toSQL()`, real Drizzle builder, `../db` deliberately not mocked) and diffs the arbiter against both the migration DDL and the Drizzle schema predicate. Every assertion is mutation-checked: dropping `targetWhere`, dropping a conflict-target column, swapping the `policy_id`/`device_id` bindings, and drifting the schema predicate each turn it red.

**Real Postgres:** brought up an isolated per-worktree stack, applied all 618 migrations, and confirmed both indexes exist with exactly the intended predicates. `pnpm db:check-drift` clean.

Two cases added to the existing `evaluatePolicy` integration harness (10 tests green, run 3x, no flake): a re-evaluation that takes the `DO UPDATE` branch against the **live** index, and a blind duplicate INSERT that must be rejected — asserted on SQLSTATE `23505` plus the constraint name, since Drizzle wraps the driver error and a message regex would also pass for an RLS denial. Dropping the live index reds these with the real `42P10 infer_arbiter_indexes`, which is the failure the unit test can only assert by proxy.

The RLS-scope bug above was reproduced both ways against a non-superuser table owner under FORCE RLS: without the fix, `removed 0` (an RLS artifact that reads as "there were none") followed by `ERROR: could not create unique index`; with it, `removed 1` and a clean build.

> One thing I got wrong and want visible: I first wrote the concurrency case as two racing `evaluatePolicy` calls. Reverting the writer to the old select-then-insert left it **green** — nothing forces both SELECTs to land before either INSERT — so it was not a test, and I replaced it with the deterministic index-level assertion above.

## Follow-up (not in this PR)

Migrations doing DML on FORCE-RLS tables without setting `breeze.scope` is a repo-wide pattern, not unique to this file (`2026-09-22-ai-alert-verdicts-live-unique.sql` has the same gap). Worth a separate issue.

Closes #4122

### DIFF
--- a/apps/api/migrations/2026-09-29-100000-automation-policy-compliance-unique.sql
+++ b/apps/api/migrations/2026-09-29-100000-automation-policy-compliance-unique.sql
@@ -1,0 +1,101 @@
+-- automation_policy_compliance: dedupe + unique indexes (#4122)
+--
+-- The table has always been an "upsert" target but never had a uniqueness
+-- constraint to upsert against, and `policyEvaluationService.ts` wrote it with a
+-- non-atomic select-then-insert. Two concurrent evaluations of the same policy
+-- (scheduled sweep + manual "Evaluate now", or two workers on the same policy)
+-- both saw "no row" and both inserted, so duplicates accumulate for a single
+-- (policy, device) pair. Readers then pick an arbitrary one:
+--   * `policyAlertBridge.handlePolicyViolation`'s reconcile guard reads this
+--     table to decide whether a violation is still current (mitigated in #4117
+--     with `ORDER BY updated_at DESC`, which is a tiebreak, not a guarantee).
+--   * The evaluation upsert itself kept writing to whichever duplicate its
+--     unordered SELECT happened to return, so status could ping-pong.
+--
+-- Two shapes live in this table and exactly one axis is populated per row:
+--   1. automation policy   -> (policy_id, device_id)
+--   2. config policy item  -> (config_policy_id, config_item_name, device_id)
+-- Both key columns are nullable, so uniqueness is expressed as two PARTIAL
+-- unique indexes rather than table constraints.
+--
+-- Index creation is deliberately NOT `CONCURRENTLY`: autoMigrate wraps each file
+-- in a transaction, and splitting the dedupe from the index build would leave a
+-- window in which a racing writer re-creates a duplicate and the concurrent
+-- build finishes INVALID. Doing both in one transaction is the atomic option.
+
+-- ---------------------------------------------------------------------------
+-- 1. Dedupe the automation-policy shape, keeping the freshest row per pair.
+--    "Freshest" = highest updated_at, id DESC as a deterministic tiebreak.
+--    The surviving row carries its own remediation_attempts; counters on the
+--    discarded duplicates are intentionally not merged — the freshest row is
+--    the one every reader was already trying to read.
+-- ---------------------------------------------------------------------------
+DO $$
+DECLARE
+  removed_policy_dupes INTEGER;
+BEGIN
+  WITH ranked AS (
+    SELECT
+      id,
+      ROW_NUMBER() OVER (
+        PARTITION BY policy_id, device_id
+        ORDER BY updated_at DESC, id DESC
+      ) AS ordinal
+    FROM automation_policy_compliance
+    WHERE policy_id IS NOT NULL
+  )
+  DELETE FROM automation_policy_compliance
+   WHERE id IN (SELECT id FROM ranked WHERE ordinal > 1);
+  GET DIAGNOSTICS removed_policy_dupes = ROW_COUNT;
+
+  -- Deliberately always-report (no `IF n > 0` guard): a zero is evidence too.
+  -- These rows drive compliance state and alert reconciliation, so the count of
+  -- what this cleanup discarded belongs in the Postgres log unconditionally.
+  RAISE WARNING 'automation_policy_compliance dedupe: removed % duplicate (policy_id, device_id) row(s)',
+    removed_policy_dupes;
+END $$;
+
+-- ---------------------------------------------------------------------------
+-- 2. Dedupe the config-policy-item shape.
+--    Scoped to rows where BOTH key columns are populated: NULLs never collide
+--    in a btree unique index, so a row missing config_item_name cannot be keyed
+--    and must not be deleted on a key we cannot actually form.
+-- ---------------------------------------------------------------------------
+DO $$
+DECLARE
+  removed_config_dupes INTEGER;
+BEGIN
+  WITH ranked AS (
+    SELECT
+      id,
+      ROW_NUMBER() OVER (
+        PARTITION BY config_policy_id, config_item_name, device_id
+        ORDER BY updated_at DESC, id DESC
+      ) AS ordinal
+    FROM automation_policy_compliance
+    WHERE config_policy_id IS NOT NULL
+      AND config_item_name IS NOT NULL
+  )
+  DELETE FROM automation_policy_compliance
+   WHERE id IN (SELECT id FROM ranked WHERE ordinal > 1);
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (correctness/migration/RLS) + `pr-test-analyzer` (test quality). Both ran as subagents; a first dispatch bounced off the concurrent-subagent cap and was retried, so no checklist was run inline.

**Findings:** 6 raised → all 6 addressed in `d62ef708`; 0 outstanding.

1. **[Important — would have broken boot]** The dedupe `DELETE`s run against a FORCE-RLS table, and `breeze_current_scope()` defaults to `'none'`, not `'system'` — `0012-tenant-rls-deny-default.sql` superseded the `0008` definition I had reasoned from, and autoMigrate sets no scope. On the containerised deployment `DATABASE_URL` is a superuser so RLS is bypassed; on **managed Postgres (DigitalOcean/RDS) the admin role is not**, and there the DELETEs matched zero rows and `CREATE UNIQUE INDEX` then aborted on the survivors, refusing the API's boot. Fixed with the `set_config('breeze.scope','system',true)` line ~10 existing migrations already use. My original clearance of this was wrong; the reviewer was right.
2. **[Test — proven green-on-broken]** `expect.arrayContaining` for the bound params is order-insensitive: swapping the `policy_id`/`device_id` bindings passed 9/9. Now positional.
3. **[Test — proven green-on-broken]** The schema's index predicate was only `toBeDefined()`, so pointing `apc_policy_device_uq` at a different column passed 9/9. Now compiled via `PgDialect` and diffed against the migration.
4. **[Test]** `normalizePredicate` stripped wrapping parens before trimming (false fail on a reformatted migration); dropped a decorative `toMatch(/"policy_id"/)` that was trivially true since Drizzle names every column in the INSERT list.
5. **[Coverage]** No real-DB proof that Postgres can *infer* the partial index. Added two cases to the existing `evaluatePolicy` integration harness.
6. **[Stale comment]** `policyAlertBridge` still claimed duplicate `(policy, device)` rows "exist in the wild". Updated.

Cleared without change (reviewer-confirmed): the dedupe can neither over-delete nor miss a duplicate (`device_id` is `notNull`, so no partition column is nullable within either scope); `ON CONFLICT DO UPDATE` is safe under the device-join RLS policy because `device_id` is in both conflict keys, so a conflicting row is exactly as visible as the inserted one; no writer was left unconverted; no behaviour dropped at either call site.

**Tests:** 181 unit green (`vitest run` on the 11 affected files); 10 integration green against a real migrated DB (run 3x, no flake); `tsc --noEmit -p apps/api` clean; eslint clean; `pnpm db:check-drift` clean (618 migrations match the ledger).

Every assertion in the new unit file is mutation-checked — dropping `targetWhere`, dropping a conflict-target column, swapping the column bindings, and drifting the schema predicate each turn it red. Dropping the live index reds the integration cases with the real `42P10 infer_arbiter_indexes`.

One correction worth recording: I first wrote the concurrency case as two racing `evaluatePolicy` calls. Reverting the writer to the old select-then-insert left it **green** — nothing forces both SELECTs to land before either INSERT — so it was not a test. It was replaced with a deterministic index-level assertion (blind duplicate INSERT must fail `23505` on `apc_policy_device_uq`).

**Status:** review-clean, awaiting maintainer merge. Not merging and leaving #4122 open — both are yours to call. CI was still running at hand-off, so please gate the merge on green checks for `d62ef708`.

Issue #4264, opened 2026-08-30
[Security] m365-graph-actions-executor has no Dockerfile hardening block — the Graph WRITE boundary is less guarded than the read one


ToddHebebrand:
Flagged during the #4246 / PR #4263 work and deliberately left out of scope.

`scripts/security/check-supply-chain-hardening.sh` asserts **Dockerfile-content** hardening for two of the three M365 executors but not the third. Verified on `origin/main`:

- **`m365-graph-read-executor`** — full block at `:141-158`: digest-pinned base (build + runtime stages), non-root `node` user, bounded health endpoint, compiled-runtime-only entrypoint, **no mutable Alpine package resolution**, no env/cert/key/secret copies, explicit build-context allowlist. Plus Dependabot and CI-gating assertions.
- **`m365-communications-executor`** — equivalent coverage (10 references).
- **`m365-graph-actions-executor`** — **no Dockerfile-content block at all.** Its only references (`:180-186`, `:250`, `:256-259`, `:350`, `:386`) cover CI gating, the Trivy scan name, release digest-pinning, compose rejection and the deployment template. Nothing asserts anything about the contents of `apps/m365-graph-actions-executor/Dockerfile`.

The script itself acknowledges the asymmetry in a comment.

### Why this is the wrong way round

The actions executor handles Microsoft Graph **write** actions. The read executor — which has the strictest content rules — handles reads. The component with the higher blast radius has the weaker structural guarantee, so a future edit to the actions-executor Dockerfile that drops the non-root user, un-pins the base image, or adds an `apk add` would pass CI silently, while the same edit to its read sibling fails.

This is a latent gap rather than a live vulnerability: the actions-executor Dockerfile appears to follow the same conventions today. Nothing enforces that it keeps doing so.

### Suggested fix

Give `m365-graph-actions-executor` the same content block the read executor has. PR #4263 restructures the shared `apk` rule already, so the mechanics are established — the remaining work is extending the per-image assertions and confirming the current Dockerfile actually satisfies each one before pinning it.

Worth considering as part of it: the three executor blocks are near-identical copy-paste. A loop over an executor list would make the next executor's coverage automatic instead of a thing someone has to remember — the same "hand-maintained parallel list drifts" failure behind #4260 and the stale deploy service list.

🤖 Filed by Claude Code from the #4246 fix; the per-executor coverage asymmetry was verified line-by-line against `origin/main`.

ToddHebebrand:
[...]

PR #4422, opened 2026-09-01
fix(security): enforce credential-boundary Dockerfile policy on actions executor (#4272)

### LINKED ISSUE #4272: m365-graph-actions-executor Dockerfile has no credential-boundary policy in check-supply-chain-hardening.sh
## Summary

`scripts/security/check-supply-chain-hardening.sh` enforces a credential-boundary Dockerfile policy for two of the three M365 customer-Graph executors but **not** the third. `apps/m365-graph-actions-executor/Dockerfile` — the executor that holds *mutation* credentials — has no Dockerfile-content policy at all.

## Detail (VERIFIED)

The script defines exactly two Dockerfile path variables and runs `require_audited_openssl_upgrade` on each:

- `EXECUTOR_DOCKERFILE=apps/m365-graph-read-executor/Dockerfile` (`:190`, checked at `:202`)
- `COMMS_EXECUTOR_DOCKERFILE=apps/m365-communications-executor/Dockerfile` (`:363`, checked at `:375`)

There is no `ACTIONS_EXECUTOR_DOCKERFILE`. Grepping `m365-graph-actions-executor` in the script returns only workflow / release / compose assertions — its **Dockerfile contents are never inspected**: no `USER node` check, no `HEALTHCHECK`, no `CMD` pin, no secret-COPY rejection, and no apk policy.

Verified: appending `RUN apk add --no-cache sudo openssh netcat-openbsd` to `apps/m365-graph-actions-executor/Dockerfile` and running the full guard exits **0** ("Supply-chain hardening checks passed").

## Why it matters now

This gap **predates** #4257 (the removed blanket `reject_grep` covered only the same two files), so it is not a regression. But #4257 makes it load-bearing:

1. It adds `RUN apk upgrade --no-cache libcrypto3 libssl3` to the actions executor with nothing constraining what else can ride in on that line.
2. The new test's `CREDENTIAL_BOUNDARY_DOCKERFILES` (`apps/api/src/config/dockerfileOpensslUpgrade.test.ts:108-111`) lists only the read and comms executors — so the omission now looks deliberate and reviewed.
3. `docs/deploy/m365-customer-graph-actions-executor.md:135` documents this image as the same private-identity credential-boundary workload class as the read executor — and it is currently the *least* constrained of the three.

## Proposed fix

Add `ACTIONS_EXECUTOR_DOCKERFILE=apps/m365-graph-actions-executor/Dockerfile` with a `require_audited_openssl_upgrade` call (and ideally the same `USER`/`HEALTHCHECK`/`CMD`/secret-COPY checks the read executor gets, for parity), and add the path to `CREDENTIAL_BOUNDARY_DOCKERFILES` in the test.

## Blast radius

Security-coverage gap at a mutation-credential boundary. Priority: elevated — closing it is a few lines and removes an unguarded apk surface on the highest-privilege of the three executors.

ToddHebebrand:
## Summary

`scripts/security/check-supply-chain-hardening.sh` enforced a credential-boundary Dockerfile policy for the read and communications M365 executors but not the third — `apps/m365-graph-actions-executor/Dockerfile`, the executor that holds *mutation* credentials, had no Dockerfile-content policy at all. This meant an unaudited `apk add`, a dropped non-root `USER`, an un-pinned base, or a copied secret could land in the highest-privilege of the three executor images and the guard would still exit 0.

Also covers #4264 (duplicate report of the same gap).

## Changes

- `scripts/security/check-supply-chain-hardening.sh`: add `ACTIONS_EXECUTOR_DOCKERFILE` and register it against the same checks the read executor gets — digest-pinned base (build + runner), `USER node`, `HEALTHCHECK .../healthz`, `CMD ["node", "dist/index.cjs"]`, `require_audited_openssl_upgrade`, secret-COPY rejection, and the explicit-build-context-allowlist check — plus a Dependabot-coverage assertion (mirroring the read executor's).
- `apps/api/src/config/dockerfileOpensslUpgrade.test.ts`: add the actions executor to `CREDENTIAL_BOUNDARY_DOCKERFILES` so `it.each` proves it permits only the exact audited OpenSSL upgrade, same as its siblings.
- `apps/api/src/config/executorApkPolicy.test.ts`: extend the wiring test to all three executors, and add a new regression test that reproduces the issue's own verification — append `RUN apk add --no-cache sudo openssh netcat-openbsd` to the real actions Dockerfile content and assert the policy now rejects it (this exact mutation was verified in the issue to pass the guard with exit 0 before this fix).

`apps/m365-graph-actions-executor/Dockerfile` itself is unchanged — it already conforms to the policy (digest-pinned base, `apk upgrade` CVE repair, non-root `USER node`, healthcheck). This PR closes the CI-enforcement gap, not the Dockerfile.

## Verification

- Full guard passes on `main`'s current actions Dockerfile: `bash scripts/security/check-supply-chain-hardening.sh` → `Supply-chain hardening checks passed.`
- Reproduced the issue's exact regression: appending `RUN apk add --no-cache sudo openssh netcat-openbsd` to `apps/m365-graph-actions-executor/Dockerfile` now makes the full guard fail with `ERROR: apps/m365-graph-actions-executor/Dockerfile contains an unaudited apk invocation: ...` (previously exited 0). File was reverted after the manual check; `git diff` on that file is empty.
- `vitest run src/config/dockerfileOpensslUpgrade.test.ts src/config/executorApkPolicy.test.ts --pool=threads --maxWorkers=2` (from `apps/api`): 28/28 passed.
- `tsc --noEmit -p apps/api`: clean.

## Test plan

- [x] Affected vitest files pass
- [x] `tsc --noEmit -p apps/api` clean
- [x] Full `check-supply-chain-hardening.sh` passes on `main`'s actions Dockerfile
- [x] Full guard rejects the issue's reported regression when reproduced

Closes #4272

### DIFF
--- a/apps/api/src/config/dockerfileOpensslUpgrade.test.ts
+++ b/apps/api/src/config/dockerfileOpensslUpgrade.test.ts
@@ -114,8 +114,16 @@ const PINNED_NODE_ALPINE_RE = /^node:[^\s@]*-alpine@sha256:[0-9a-f]{64}$/;
 /** Customer-Graph credential-boundary images with the narrow CVE exception. */
 const CREDENTIAL_BOUNDARY_DOCKERFILES = [
   'apps/m365-graph-read-executor/Dockerfile',
+  'apps/m365-graph-actions-executor/Dockerfile',
   'apps/m365-communications-executor/Dockerfile',
 ] as const;
+
+/** Hardening-script shell variable each credential-boundary Dockerfile is checked through. */
+const CREDENTIAL_BOUNDARY_SCRIPT_VARIABLE: Record<(typeof CREDENTIAL_BOUNDARY_DOCKERFILES)[number], string> = {
+  'apps/m365-graph-read-executor/Dockerfile': 'EXECUTOR_DOCKERFILE',
+  'apps/m365-graph-actions-executor/Dockerfile': 'ACTIONS_EXECUTOR_DOCKERFILE',
+  'apps/m365-communications-executor/Dockerfile': 'COMMS_EXECUTOR_DOCKERFILE',
+};
 
 const HARDENING_SCRIPT = 'scripts/security/check-supply-chain-hardening.sh';
 
@@ -273,10 +281,7 @@ describe('Dockerfile OpenSSL upgrade coverage', () => {
     (dockerfile) => {
       const script = readFileSync(path.join(REPO_ROOT, HARDENING_SCRIPT), 'utf8');
       const source = SOURCES.get(dockerfile)!;
-      const variable =
-        dockerfile === 'apps/m365-graph-read-executor/Dockerfile'
-          ? 'EXECUTOR_DOCKERFILE'
-          : 'COMMS_EXECUTOR_DOCKERFILE';
+      const variable = CREDENTIAL_BOUNDARY_SCRIPT_VARIABLE[dockerfile];
       const apkMutationLines = source
         .split('\n')
         .map((line) => line.trim())
--- a/apps/api/src/config/executorApkPolicy.test.ts
+++ b/apps/api/src/config/executorApkPolicy.test.ts
@@ -1,5 +1,5 @@
 import { execFileSync } from 'node:child_process';
-import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` was invoked, but both the code-reviewer and pr-test-analyzer subagent spawns failed with "Concurrent subagent limit reached" (many sibling issue-fixers running in parallel on this machine). Per the fallback instruction, I ran the review checklist inline myself instead.

**Findings:** 0 outstanding. Inline checks performed:
- Diffed the full PR change set (`git diff origin/main...fix/4272-actions-executor-hardening-check`) line by line.
- Confirmed the new `ACTIONS_EXECUTOR_DOCKERFILE` block in `check-supply-chain-hardening.sh` is a structural mirror of the existing `EXECUTOR_DOCKERFILE` (read executor) block, register-for-register.
- Confirmed the stale comment above the communications-executor block (previously claiming "the actions executor never got one") was updated to reflect the new state.
- Confirmed `apps/m365-graph-actions-executor/Dockerfile` itself is untouched — it already conforms to the policy; this PR only closes the enforcement gap.
- Confirmed the new fixture regression test in `executorApkPolicy.test.ts` reproduces the issue's exact verification (`RUN apk add --no-cache sudo openssh netcat-openbsd` appended to the real actions Dockerfile) and asserts it's now rejected.
- Confirmed the broader Dockerfile-content checks (USER/HEALTHCHECK/CMD/digest-pin/secret-COPY) are proven only via `require_grep`/`reject_grep` against the real file (exercised by CI running the full script directly, per `.github/workflows/ci.yml:782`) rather than fixture-mutated in vitest — this matches the pre-existing, unchanged test rigor for the read and communications executor siblings, not a gap introduced here.
- `eslint` on both changed test files: clean.
- `bash -n` syntax check on the script: clean.

**Tests:** `vitest run src/config/dockerfileOpensslUpgrade.test.ts src/config/executorApkPolicy.test.ts --pool=threads --maxWorkers=2` (from `apps/api`) — 28/28 passed. `tsc --noEmit -p apps/api` — clean. Full `bash scripts/security/check-supply-chain-hardening.sh` — passes on `main`'s current tree. Manually reproduced the issue's exact regression (appended the unaudited `apk add` line to the real actions Dockerfile, ran the full guard, got `ERROR: ... contains an unaudited apk invocation`, then reverted — `git diff` on that file confirmed clean afterward).

**Status:** review-clean (inline), awaiting maintainer merge.

Issue #4086, opened 2026-08-26
W10: Wave 3.5d — BREEZE_ROLE split: separate entrypoints, compose worker service, setInterval to repeatables for the genuinely-global timers, droplet deploy + version parity


ToddHebebrand:
Wave 3.5d of #3821 — the actual role split. Split out of #3825 on 2026-08-26;
see `docs/superpowers/plans/ai-mcp/2026-08-26-ai-agents-wave3.5a-idempotency.md`
(landed in #4083).

**Blocked by #4084 (socket affinity) and #4085 (durable event dispatch).**
Shipping this before both is what silently breaks agent command dispatch and
every event-driven subscriber.

## Scope

`BREEZE_ROLE=api|worker|all`, defaulting to `all` so self-hosters are unaffected.
Hosted gets a `worker` container.

## `setInterval` → repeatables: most sites must NOT be converted

Of ~23 `setInterval` call sites in `apps/api/src`, the majority sweep
**process-local** state and must keep one timer per process. Converting them to a
shared repeatable would break them:

- `services/remoteAccessPolicy.ts:96`, `routes/agentWs.ts:2781` — sweep
  process-local `Map`s. **API-only, leave as-is.**
- `index.ts:2100` — drains the **module-local** audit retry queue
  (`services/auditService.ts:43`). One per role that calls `createAuditLogAsync`;
  never a global singleton.
- `db/dbPoolHealthMonitor.ts:588`, `services/eventLoopMonitor.ts:238`,
  `services/streamingSessionManager.ts:500`,
  `services/llm/openaiSessionManager.ts:50` — per-process by definition.
- Per-connection timers (`agentWs.ts:2005`, `tunnelWs.ts:825,878`,
  `desktopWs.ts:793,961`, `terminalWs.ts:738,833`, `mcpServer.ts:371`,
  `clientAi/sessions.ts:748`) — excluded entirely.

Genuinely global, still to handle here:

- `jobs/incidentJobs.ts:206` correlation — read/count/log only; duplicate load,
  not duplicate data. Worker-only placement.
- `jobs/mtlsCertificateRevocation.ts:177`, `jobs/oauthRevocationRetryWorker.ts:77`,
  `services/desktopSessionOrphanRecovery.ts:349` — already guarded (stable job id
  / `FOR UPDATE SKIP LOCKED` / Redis Lua lease claim). Placement only.
- `jobs/reportScheduleWorker.ts:638` — **not yet fixed.** Two processes can select
[...]

PR #4423, opened 2026-09-01
feat(worker): /metrics endpoint, event-loop + DB-pool monitors, breeze_role Sentry tag (#4143)

### LINKED ISSUE #4143: Worker container observability: /metrics registry, event-loop + DB-pool monitors, breeze_role Sentry tag
Follow-up to #4086 (PR #4140). The worker container currently exposes no /metrics (the prom registry is private to routes/metrics.ts and only mounted on the Hono app — BullMQ counters for the 76 relocated workers vanish from scraping once a droplet flips to split mode); it also lacks startEventLoopMonitor + setConnectTimeoutClassifier (#3022) and startDbPoolHealthMonitor (#3214), which matter MOST where the heavy jobs run; and neither entrypoint sets a breeze_role Sentry tag so the two containers' events are indistinguishable. Extract the metrics registry to a leaf module served by the worker's slim HTTP server, add the monitors (+ their stops in the shutdown preamble), tag Sentry scope with the role.

ToddHebebrand:
## The problem

In split-role mode (#4086), the `worker` container is the process running the 76 relocated BullMQ workers, the retention sweeps, and the heavy scheduled jobs. It was also completely unobservable — three separate gaps:

1. **No `/metrics`.** The Prometheus `Registry` was a module-private `const register = new Registry()` inside `routes/metrics.ts` — the one file a worker-role process must never import (`services/workerEntrypointClosure.contract.test.ts` enforces this). The only way to render a scrape was to load the entire route graph. So the worker published *nothing*: not stale series, not zeros — **absent**, with no `up` target either, which no alert rule can see.
2. **No `startEventLoopMonitor` (#3022) / `startDbPoolHealthMonitor` (#3214).** Both were wired only in `index.ts`. The loudest signature in #3022's original incident was the patch scheduler — which in split mode runs in *this* container.
3. **No `breeze_role` Sentry tag.** Both containers run the same image, same DSN, same release, same environment. Their events were indistinguishable.

## The fix

Three new **leaf** modules, each held to the worker's import-closure contract:

| Module | Closure | Role |
|---|---|---|
| `services/metricsRegistry.ts` | `prom-client` only | the shared `Registry` singleton |
| `services/metricsRuntime.ts` | `prom-client` + 3 leaves + `db/dbPoolHealthMonitor` | role-agnostic runtime series |
| `services/metricsScrapeAuth.ts` | `node:crypto` only | the scrape gate both roles enforce |

**`routes/metrics.ts`** now aliases the shared registry as `register` — every existing series declaration is unchanged. The runtime gauges (process identity, event-loop lag ×5, connect-timeout ×2, pool-health ×4) moved to `metricsRuntime` verbatim, along with their seeds and scrape-time refresh.

`metricsRuntime` also owns the `setDbConnectTimeoutMetricsRecorder` binding. That recorder is a **single global slot**: bound from the route file, a worker-role process counted no CONNECT_TIMEOUTs at all while `db/dbPoolHealthMonitor` — running in that same process — alerted on the rate derived from them. It is still re-called from `bindMetricsRecorders()`, because `__resetDbConnectTimeoutStatsForTests()` nulls the slot and `resetMetricsForTesting()` is what tests rely on to restore it.

**`worker.ts`** serves `/metrics` from its slim raw-node:http server behind the shared gate, starts the event-loop monitor + SAFE CONNECT_TIMEOUT classifier right after `initSentry()`, starts the pool-health watchdog after `validateConfig()` (both of index.ts's documented ordering constraints preserved), and **stops both in the shutdown preamble ahead of the phases** — so the watchdog cannot open a probe connection while the `db` phase drains the pool and report `database-unreachable` about a process that is simply shutting down.

**`services/sentry.ts`** sets `breeze_role` (`api`|`worker`|`all`) on the global scope in `initSentry()`, **and** adds it to `ALLOWED_TAG_NAMES`. Both halves are required: `scrubEvent` rebuilds `tags` from that allowlist, so a set-but-unlisted tag arrives as nothing — exactly how the `worker` tag regressed for two days (documented in that file). The value is re-derived locally rather than imported from `config/env`, which is *not* a leaf (~40 module-scope `process.env` reads) and would have grown the graph of a module ~120 others import, including `db/index.ts`. An anti-drift test pins the two normalisations together across every input class.

### One deliberate behavioural difference

The api role's `/api/metrics/scrape` resolves the client IP via `getTrustedClientIpOrUndefined`, which consults forwarded headers against that process's trusted-proxy config. The worker's port has **no** trusted-proxy configuration, so its allowlist matches the **direct peer** and ignores `X-Forwarded-For` — otherwise any caller could name its own source IP and walk through the allowlist. Documented at the call site, in `docs/deploy/worker-split.md`, and covered by a test.

## Verification

- **Contract test (the hard constraint):** `workerEntrypointClosure.contract.test.ts` — **97 passed**. Mutation-proved live: adding `import { metricsRoutes } from '../routes/metrics'` to the new leaf produced **2 failures**, then restored.
- **Affected suites, single-fork** (`--pool=threads --maxWorkers=2`): 18 files, **621 passed / 0 failed** — worker boot, routes/metrics, sentry ×2, event-loop monitor + wiring, dbPoolHealth monitor + wiring, connect-timeout stats, metricsMiddleware, heartbeat, softwareRemediationWorker, retention/actionIntents/m365/agentCertificateBinding metrics.
- **Typecheck:** `tsc --noEmit --project apps/api/tsconfig.json` clean.
- **Lint:** `eslint` clean on all 9 changed/added files.
- **New tests mutation-checked** (not vacuous) — each mutation below produced real failures, then was reverted:

  | Mutation | Result |
  |---|---|
  | drop the `/metrics` route from worker.ts | 8 failed |
  | drop `stopEventLoopMonitor()` from the shutdown preamble | 1 failed |
  | drop `Sentry.setTag('breeze_role', …)` | 4 failed |
  | drop `breeze_role` from `ALLOWED_TAG_NAMES` | 4 failed |

The `/metrics` tests run against the **real** slim HTTP server and register a canary gauge on the registry singleton the freshly-imported worker binds to — so they distinguish "serves the SHARED registry" from "serves some empty registry of its own", which would otherwise satisfy every auth assertion while publishing nothing.

## Operator note

`docs/deploy/worker-split.md` gains an Observability section: the worker is a **separate Prometheus scrape target** (`http://<worker>:3001/metrics`) that must be added when adopting the split. It publishes the runtime set only — expect the HTTP/business/fleet series from the api target and do not alert on their absence here.

Closes #4143

### DIFF
--- a/apps/api/src/routes/metrics.ts
+++ b/apps/api/src/routes/metrics.ts
@@ -7,25 +7,23 @@
 import { Hono } from 'hono';
 import { routePath as honoRoutePath } from 'hono/route';
 import { avg, and, eq, gte, inArray, sql } from 'drizzle-orm';
-import { Counter, Gauge, Histogram, Registry } from 'prom-client';
-import { createHash, timingSafeEqual } from 'crypto';
+import { Counter, Gauge, Histogram } from 'prom-client';
 
 import { db, runOutsideDbContext, withSystemDbAccessContext } from '../db';
 import { deviceMetrics, devices, metricRollups, recoveryReadiness as recoveryReadinessTable, remoteSessions } from '../db/schema';
 import { authMiddleware, requirePermission, requireScope } from '../middleware/auth';
 import { getTrustedClientIpOrUndefined } from '../services/clientIp';
-import { getEventLoopLagStats, readLatestEventLoopLag } from '../services/eventLoopMonitor';
+import { metricsRegistry } from '../services/metricsRegistry';
 import {
-  getDbConnectTimeoutStats,
-  setDbConnectTimeoutMetricsRecorder,
-} from '../services/dbConnectTimeoutStats';
+  bindRuntimeMetricsRecorders,
+  initializeRuntimeMetricDefaults,
+  updateRuntimeMetrics,
+} from '../services/metricsRuntime';
 import {
-  DB_POOL_HEALTH_VERDICTS,
-  getDbPoolHealthCheckFailures,
-  getDbPoolHealthProbeCloseFailures,
-  getDbPoolHealthWindowMs,
-  getLastDbPoolHealthAssessment,
-} from '../db/dbPoolHealthMonitor';
+  evaluateMetricsScrapeAuth,
+  parseMetricsScrapeIpAllowlist,
+  resolveMetricsScrapeToken,
+} from '../services/metricsScrapeAuth';
 import { PERMISSIONS, type UserPermissions } from '../services/permissions';
 import { BACKUP_LOW_READINESS_THRESHOLD } from './backup/constants';
 import {
@@ -71,14 +69,6 @@ export {
 export const metricsRoutes = new Hono();
 const requireMetricsRead = requirePermission(PERMISSIONS.DEVICES_READ.resource, PERMISSIONS.DEVICES_READ.action);
 
-function resolveMetricsScrapeToken(): string | undefined {
-  const rawToken = process.env.METRICS_SCRAPE_TOKEN?.trim();
-  // Production hardening: refuse to run with obvious placeholder tokens.
-  return (process.env.NODE_ENV ?? 'development') === 'production' && (!rawToken || rawToken === 'REDACTED_DEV_TOKEN')
-    ? undefined
-    : rawToken;
-}
-
 let METRICS_SCRAPE_TOKEN = resolveMetricsScrapeToken();
 
 function envFlag(name: string, fallback: boolean): boolean {
@@ -86,17 +76,6 @@ function envFlag(name: string, fallback: boolean): boolean {
   if (!raw) return fallback;
   const normalized = raw.trim().toLowerCase();
   return normalized === '1' || normalized === 'true' || normalized === 'yes' || normalized === 'on';
-}
-
-function parseCsvSet(raw: string | undefined): Set<string> {
-  if (!raw) return new Set();
-  return new Set(raw.split(',').map((s) => s.trim()).filter((s) => s.length > 0));
-}
-
-function safeEqual(a: string, b: string): boolean {
-  const ha = createHash('sha256').update(a).digest();
-  const hb = createHash('sha256').update(b).digest();
-  return timingSafeEqual(ha, hb);
 }
 
 // Default: hide org IDs in Prometheus labels in production (they can leak tenant identifiers).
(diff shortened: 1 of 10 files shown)


ToddHebebrand:
**Review run:** inline checklist by the author — the `pr-review-toolkit` subagents could not be dispatched (concurrent-subagent cap reached; ~20 sibling issue-fixers running on this machine). Recorded here so the gap is explicit rather than implied: **no independent reviewer has looked at this yet.** A second pass with `code-reviewer` / `silent-failure-hunter` / `pr-test-analyzer` is worth running before merge if you want one.

**What the inline review covered** (six risk areas, each verified mechanically rather than by reading):

| # | Risk | Method | Result |
|---|---|---|---|
| 1 | worker.ts closure reaching `routes/` | `workerEntrypointClosure.contract.test.ts`, **mutation-proved live** (added a `routes/metrics` import to the new leaf → 2 failures, reverted) | clean; all 5 new static imports are leaves |
| 2 | `services/sentry.ts` losing leaf status | diffed its import list vs `origin/main` | unchanged — `config/env` deliberately not imported; anti-drift test pins the two `breezeRole` normalisations across 9 input classes |
| 3 | `breeze_role` voided by the scrubber | grep + a test asserting it survives `scrubEvent` | set (sentry.ts:517) **and** allowlisted (sentry.ts:273); both mutations fail the tests |
| 4 | the ~141-line gauge move dropping/duplicating a series | byte-compared the moved block, diffed the full metric-name set, diffed every seed statement and both update-function bodies against `origin/main` | **block byte-identical**; 56 series in, 56 out, no duplicates; 0 seeds lost; `updateDbPoolHealthMetrics`/`updateEventLoopMetrics` identical |
| 5 | double-binding the connect-timeout recorder | traced the slot | it is a single-slot **assignment** (not a list), and there is now exactly **one** `setDbConnectTimeoutMetricsRecorder(...)` call site; re-invoking it is idempotent |
| 6 | auth bypass / timing / info disclosure on the new endpoint | read the gate + 20 tests | order 503→403→401 enforced (403 asserted with a *correct* token, which is what makes it a real ordering test); `safeEqual` hashes first so no length leak; error bodies are fixed strings |

**Findings:** 1 raised, 1 fixed, 0 outstanding.

- *(medium, fixed in b8f43a16e)* the `/metrics` 500 path could double-write: if the throw came from `res.end(body)` rather than `metricsRegistry.metrics()`, the 200 header is already on the wire and the catch block's `writeJson` raises `ERR_HTTP_HEADERS_SENT` **from inside the catch** — rejecting the promise and reporting a secondary, misleading error instead of the real render fault. Now guarded on `res.headersSent`.

**Tests:** 18 affected files, **621 passed / 0 failed**, single-fork (`vitest run --pool=threads --maxWorkers=2`) — re-run after the fix commit. `tsc --noEmit --project apps/api/tsconfig.json` clean; `eslint` clean on all 9 changed/added files. Full suite deliberately not run (flaky in parallel).

**Anti-vacuity:** every new assertion group was mutation-checked — dropping the `/metrics` route (8 fail), dropping `stopEventLoopMonitor()` from the shutdown preamble (1 fail), dropping the `setTag` (4 fail), dropping the allowlist entry (4 fail). The `/metrics` tests register a canary gauge on the registry singleton *the freshly-imported worker binds to*, so they distinguish "renders the SHARED registry" from "renders an empty one of its own" — which would otherwise pass every auth assertion while publishing nothing.

**Not covered by this PR** (deliberate, flagging rather than hiding): the worker still does not publish the api's HTTP/business/fleet series or the extension/backup/S1 counters — those recorders are bound in `routes/metrics.ts`, and moving them is a much larger extraction. The worker publishes the role-agnostic runtime set only. `docs/deploy/worker-split.md` says so explicitly so nobody alerts on their absence.

**Status:** review-clean by the above, but **single-reviewer** — awaiting maintainer merge.

PR #4424, opened 2026-09-01
fix(clientAi): use isPgUniqueViolation helper for tenant-mapping conflict (#4325)

### LINKED ISSUE #4325: [API] clientAi/admin.ts hand-rolls a single-level 23505 unwrap and ignores the constraint name
## Summary

`apps/api/src/routes/clientAi/admin.ts:136` hand-rolls a SQLSTATE 23505 check that unwraps exactly one `.cause` level and does not check the constraint name, instead of using the canonical `isPgUniqueViolation` helper in `apps/api/src/utils/pgErrors.ts`. Same class as #4245 and #4020.

```ts
} catch (err) {
  // client_ai_tenant_mappings_tenant_uniq: the tenant is already mapped to
  // a DIFFERENT org. Deliberately opaque — do not reveal which org.
  if ((err as { cause?: { code?: string } }).cause?.code === '23505') {
    return c.json({ error: 'tenant_already_mapped' }, 409);
  }
  throw err;
}
```

## Two defects

**1. Single-level unwrap.** `isPgUniqueViolation` walks the `.cause` chain up to 5 deep precisely because "postgres.js raises a `PostgresError` with `.code === '23505'`, but Drizzle wraps it in a `DrizzleQueryError` whose own `.code`/`.constraint` are undefined." This call site handles only the one-wrap shape. It happens to work against today's Drizzle wrapping, so this is fragility rather than a live break — but it is exactly the shape that silently stopped working in #4020, and a flat `PostgresError` (any path that reaches the driver without the Drizzle wrapper) already falls through to a raw 500.

**2. The constraint name is never checked.** The comment names `client_ai_tenant_mappings_tenant_uniq` as the intended trigger, but the code matches *any* 23505 from that statement and reports it as `tenant_already_mapped`. The insert uses `.onConflictDoUpdate({ target: clientAiTenantMappings.orgId })`, so the `orgId` conflict is absorbed — but any other unique index on the table (now or later) would be misreported to the caller as a tenant-mapping collision. `isPgUniqueViolation(err, 'client_ai_tenant_mappings_tenant_uniq')` fixes both defects in one call and preserves the deliberate opacity of the response.

## Suggested fix

```ts
import { isPgUniqueViolation } from '../../utils/pgErrors';
...
if (isPgUniqueViolation(err, 'client_ai_tenant_mappings_tenant_uniq')) {
  return c.json({ error: 'tenant_already_mapped' }, 409);
}
```

A regression test should assert both wrapper shapes (flat `PostgresError` and `DrizzleQueryError`-wrapped) map to 409, and that a 23505 naming a *different* constraint does not.

## Scope

Found while rebasing PR #4240. **Verified** by reading `origin/main`: this is the only remaining hand-rolled single-level unwrap of its kind in the AI/client-AI routes. `partnerServicePrincipals.ts`, `pax8OrderService.ts` and `pax8SyncService.ts` each walk the `.cause` chain themselves and are fine. **Not checked:** whether a flat `PostgresError` can actually reach this catch in production — the fix is worth making either way, since it is a one-line swap onto the shared helper.

ToddHebebrand:
## Summary

`apps/api/src/routes/clientAi/admin.ts` hand-rolled a single-level `.cause?.code === '23505'` unwrap to detect the tenant-mapping unique-conflict case, instead of using the canonical `isPgUniqueViolation` helper in `apps/api/src/utils/pgErrors.ts`. Two defects:

1. **Single-level unwrap.** The helper walks the `.cause` chain up to 5 deep because a `DrizzleQueryError` wraps a `PostgresError`, and other wrapper shapes exist too. The hand-rolled check only handled the one-wrap shape.
2. **No constraint-name check.** The comment names `client_ai_tenant_mappings_tenant_uniq` as the intended trigger, but the code matched *any* 23505 on the insert statement and reported it as `tenant_already_mapped`, which would misreport an unrelated unique violation on the same table.

## Fix

Swap to `isPgUniqueViolation(err, 'client_ai_tenant_mappings_tenant_uniq')`, which fixes both defects in one call and preserves the deliberately-opaque 409 response.

## Tests

`apps/api/src/routes/clientAi/admin.test.ts`:
- Replaced the old single-level-unwrap test with one asserting a `DrizzleQueryError`-wrapped 23505 (outer generic error, `PostgresError` with `code` + `constraint_name` on `.cause`) on the correct constraint still maps to 409 `tenant_already_mapped`.
- Added a test asserting a 23505 naming a *different* constraint on the same table does NOT get mapped to 409 — it propagates (500).

Closes #4325

## Test plan

- [x] `pnpm exec vitest run src/routes/clientAi/admin.test.ts --pool=threads --maxWorkers=2` — 16/16 passed
- [x] `pnpm exec vitest run src/utils/pgErrors.test.ts --pool=threads --maxWorkers=2` — 20/20 passed (helper unchanged, sanity check)
- [x] `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api` — clean
- [x] `npx eslint` on both changed files — clean

### DIFF
--- a/apps/api/src/routes/clientAi/admin.test.ts
+++ b/apps/api/src/routes/clientAi/admin.test.ts
@@ -165,13 +165,21 @@ describe('client-ai admin — tenant mapping', () => {
     );
   });
 
-  it('PUT maps a tenant-uniqueness violation to 409 tenant_already_mapped', async () => {
+  it('PUT maps a DrizzleQueryError-wrapped tenant-uniqueness violation to 409 tenant_already_mapped', async () => {
+    // DrizzleQueryError shape: generic outer error, real PostgresError (with
+    // code + constraint_name) on .cause — mirrors what postgres.js + Drizzle
+    // actually produce, not a flattened one-level stub.
+    const pgErr = Object.assign(new Error('duplicate key value violates unique constraint'), {
+      code: '23505',
+      constraint_name: 'client_ai_tenant_mappings_tenant_uniq',
+    });
+    const drizzleErr = Object.assign(new Error('Failed query: insert into "client_ai_tenant_mappings" ...'), {
+      cause: pgErr,
+    });
     dbInsertMock.mockImplementation(() => ({
       values: vi.fn(() => ({
         onConflictDoUpdate: vi.fn(() => ({
-          returning: vi.fn(() =>
-            Promise.reject(Object.assign(new Error('duplicate'), { cause: { code: '23505' } }))
-          ),
+          returning: vi.fn(() => Promise.reject(drizzleErr)),
         })),
       })),
     }));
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit code-reviewer subagent (diff review against `origin/main`)
**Findings:** 1 raised (Important, comment-accuracy only — no functional defect) → addressed in `311ef8558`; 0 outstanding.
- The reviewer verified correctness by mutation testing: reverting `admin.ts` to the old hand-rolled `.cause?.code === '23505'` check, and separately dropping the constraint-name argument, each broke exactly one new test (`expected 409 to be 500`) — confirming the two new tests genuinely discriminate the fix from the old behavior.
- The one finding was that the first test's comment overclaimed its fixture depth ("not a flattened one-level stub") when it's actually the same depth the old check handled; the constraint-name discrimination is what the second test guards. Fixed by correcting the comment, not the test logic.
**Tests:** `apps/api/src/routes/clientAi/admin.test.ts` affected suite green (16/16, `vitest run --pool=threads --maxWorkers=2`); `apps/api/src/utils/pgErrors.test.ts` sanity-checked green (20/20, helper unchanged). `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api` clean. `eslint` clean on both changed files.

PR #4425, opened 2026-09-01
fix(docker): route portal dev island module URLs to portal in Caddy (#3906)

### LINKED ISSUE #3906: wt-stack: no portal island hydrates — Astro dev module URLs miss the /portal Caddy route, so portal e2e asserts dead markup
In a `pnpm wt-stack` stack, **no portal React island ever hydrates**. Pages render their server HTML and then fail silently — the markup looks right, so a browser-level check passes while nothing on the page is actually interactive.

Found while doing the live acceptance run for quote revisions (#3796): the customer proposal page showed its "Accept & sign" form, but the form was inert because `PublicQuoteView` never mounted.

### Reproduce

```
pnpm wt-stack up
# open <baseUrl>/portal/quote/<any live quote token>
```

Console:

```
GET /src/components/portal/PublicQuoteView.tsx  404 (Not Found)
[astro-island] Error hydrating /src/components/portal/PublicQuoteView.tsx
  TypeError: Failed to fetch dynamically imported module
```

### Cause

The portal's Astro dev server emits island module URLs **without** the `/portal` base prefix, and Caddy only routes that prefix to the portal container:

```
# Caddyfile
@portal path /portal /portal/*
handle @portal {
  reverse_proxy portal:4322
}
```

So `/src/components/portal/PublicQuoteView.tsx` is proxied to `web:4321`, which has no such file. Confirmed directly against a running stack:

| URL through Caddy | Status |
|---|---|
| `/src/components/portal/PublicQuoteView.tsx` | **404** |
| `/portal/src/components/portal/PublicQuoteView.tsx` | **200** |

`apps/portal/astro.config.mjs` does set `base: PORTAL_BASE`, so this is specific to how the dev server serves island modules — a production build serves bundled assets under the portal's own base path, which is why this has never been seen in prod.

### Why it matters more than a dev annoyance

`e2e-tests/playwright.config.ts` uses `stack?.baseUrl` — the same Caddy front door. Any Playwright spec that drives a portal island in a worktree stack is therefore asserting against **server-rendered markup only**: the element is present, the click does nothing, and the test can pass without the feature working. That is the same "test that cannot fail" shape as the vacuous assertions we've been finding.

The failure is also easy to misread: the page *looks* correct, and the only signal is a console 404 nobody reads.

### Fix options

- **A — route the dev module paths.** Add a dev-only Caddy matcher sending Astro's unprefixed dev paths (`/src/*`, `/@fs/*`, `/@vite/*`, `/node_modules/.vite/*`) to `portal:4322` when the referring page is under `/portal`. Confined to the worktree/dev override; production config untouched.
- **B — publish the portal port** in `docker-compose.override.yml.worktree` (it currently sets `ports: !reset []`) and let portal e2e/QA hit the portal origin directly, where its own base path resolves.
- **C — make the portal dev server emit base-prefixed island URLs**, so the existing `/portal/*` route already covers them. Cleanest if Astro/Vite can be configured for it; needs a look at whether `base` is meant to apply to dev module URLs at all.

**Recommend A** — it fixes it for every consumer of the stack (manual QA and e2e alike) without changing what URL anything is served from. B is a fine short-term unblock for a single QA session.

Whichever lands, add an e2e guard that fails when a portal island did not hydrate — otherwise the next regression is invisible again. `e2e-tests/pages/hydration.ts` looks like the right home.

ToddHebebrand:
## Summary

In a `pnpm wt-stack` (or plain `docker-compose.override.yml.dev`) stack, no portal React island ever hydrated. Pages rendered their server HTML and then failed silently — the markup looked right, so a browser-level check could pass while nothing on the page was actually interactive.

**Root cause:** `apps/portal/astro.config.mjs` sets `base: /portal`, but Astro's dev server (Vite) does not honor that `base` for the raw module-graph URLs the `astro-island` runtime fetches to hydrate a `client:load` island — e.g. `/src/components/portal/PublicQuoteView.tsx`, `/@fs/*`, `/@vite/*`, `/node_modules/.vite/*`. `docker/Caddyfile.prod` only matched `/portal` and `/portal/*`, so those unprefixed dev requests fell through to the web catch-all (`web:4321`), which has no matching file — a silent 404. A production build serves portal's own bundled, base-prefixed assets, so this has never shown up in prod.

**Fix (Option A from the issue):** a dev-only `@portalDevAssets` Caddy matcher, gated on a new `CADDY_PORTAL_DEV_ASSETS` env var set only by `docker-compose.override.yml.dev` (which `docker-compose.override.yml.worktree` layers on top of for `wt-stack`). It's unset in production, so `{env.CADDY_PORTAL_DEV_ASSETS}` is empty there and the matcher can never fire — production routing is untouched. Since `apps/web` is *also* Astro + React islands and serves the same class of raw dev paths for its own islands, the matcher additionally gates on the `Referer` header so only requests that came from a `/portal` page get routed to the portal dev server, rather than stealing web's own dev assets.

- `docker/Caddyfile.prod` — new `@portalDevAssets` matcher block (env + Referer + path gated), placed before the web catch-all.
- `docker-compose.override.yml.dev` — sets `CADDY_PORTAL_DEV_ASSETS: "1"` on the `caddy` service.
- `e2e-tests/tests/quote-contract-proposal.spec.ts` / `multi-currency.spec.ts` — add a `waitForHydration` guard on the public portal accept island (previously these specs worked around the dead-hydration bug by calling the accept API directly instead of clicking) so a routing regression fails loud instead of only showing up as a console 404 nobody reads.

## Test plan

- [x] `caddy validate --config docker/Caddyfile.prod --adapter caddyfile` (via pinned `caddy:2-alpine` docker image) — valid, no new warnings vs. the pre-change file.
- [x] `docker compose config` rendered for both the prod-only file set and the `.dev` + `.worktree` layered set (with a dummy env file covering the compose file's required vars) — confirmed `CADDY_PORTAL_DEV_ASSETS=1` merges through for the caddy service in dev/worktree, and is absent entirely in the prod-only render.
- [x] Existing Caddyfile contract test: `apps/api/src/config/proxyTrustCompose.test.ts` + `composeBindMounts.test.ts` — 9/9 passed.
- [x] `e2e-tests` typecheck (`npx tsc --noEmit -p tsconfig.json`) — clean.
- [x] `scripts/security/check-supply-chain-hardening.sh` — passed (no Watchtower/compose hardening regressions).
- [ ] Not run: the new `waitForHydration` e2e assertions against a live `wt-stack` (guardrail for this session — the shared stack ports were in use by sibling in-flight work). These should be exercised the next time this branch runs through `pnpm wt-stack up && pnpm wt-stack test -- tests/quote-contract-proposal.spec.ts tests/multi-currency.spec.ts` before merge.

Closes #3906

### DIFF
--- a/docker-compose.override.yml.dev
+++ b/docker-compose.override.yml.dev
@@ -14,6 +14,13 @@ services:
   caddy:
     # Uses canonical docker/Caddyfile.prod mounted by the base compose.
     # No command override needed — Caddy reads the volume-mounted config.
+    environment:
+      # Enables the @portalDevAssets matcher (#3906) so unprefixed Astro/Vite
+      # dev module URLs for portal islands (/src/*, /@fs/*, /@vite/*,
+      # /node_modules/.vite/*) route to the dev portal service instead of
+      # 404ing against web:4321. Unset in production — see the matcher's
+      # comment in docker/Caddyfile.prod.
+      CADDY_PORTAL_DEV_ASSETS: "1"
 
   binaries-init:
     image: alpine:3.19
--- a/docker/Caddyfile.prod
+++ b/docker/Caddyfile.prod
@@ -312,6 +312,43 @@
     reverse_proxy portal:4322
   }
 
+  # --- Dev-only: portal client-island module URLs (#3906) ---
+  # apps/portal/astro.config.mjs sets `base: /portal`, but that only prefixes
+  # *page* routes. Astro's dev server (Vite) does NOT honor `base` for the
+  # raw module-graph URLs the `astro-island` runtime fetches to hydrate a
+  # `client:load` island — e.g. `/src/components/portal/PublicQuoteView.tsx`,
+  # `/@fs/*`, `/@vite/*`, `/node_modules/.vite/*`. In dev those unprefixed
+  # requests fell through to the web catch-all below (web:4321), which has no
+  # such file: a silent 404, so the island never hydrated even though the
+  # SSR'd markup still looked correct (portal e2e specs were asserting
+  # against dead markup — see project memory
+  # portal_dev_island_hydration_404.md). A production build serves portal's
+  # own bundled assets already prefixed under /portal/*, so this has never
+  # been a prod issue.
+  #
+  # The web app is ALSO Astro + React islands and serves the SAME class of
+  # raw dev paths for its OWN islands, so this can't be routed by path alone
+  # — that would steal web's dev assets. Gate on the Referer (the browser
+  # sends it on these module-graph fetches, pointing at the page that
+  # requested them) so only requests that came from a /portal page reach the
+  # portal dev server.
+  #
+  # CADDY_PORTAL_DEV_ASSETS is unset in production — docker-compose.yml sets
+  # no such var for the caddy service, so `{env.CADDY_PORTAL_DEV_ASSETS}` is
+  # "" there and this matcher can never fire (same env-gate pattern as
+  # @remoteAccessClosed above). Only docker-compose.override.yml.dev sets it
+  # to "1", for the code-mounted dev portal (Dockerfile.portal.dev) used by
+  # both plain dev-mode and wt-stack/worktree stacks (which layer on top of
+  # .dev — see scripts/dev/wt-stack/compose.ts).
+  @portalDevAssets {
+    expression {env.CADDY_PORTAL_DEV_ASSETS} == "1"
+    header_regexp Referer ^https?://[^/]+(?::\d+)?/portal(?:/|$)
+    path /src/* /@fs/* /@vite/* /node_modules/.vite/*
+  }
+  handle @portalDevAssets {
+    reverse_proxy portal:4322
+  }
+
   # --- Web frontend (catch-all, with compression) ---
   handle {
     encode zstd gzip
--- a/e2e-tests/tests/multi-currency.spec.ts
+++ b/e2e-tests/tests/multi-currency.spec.ts
@@ -1,7 +1,7 @@
 import type { APIRequestContext, Page, Request } from '@playwright/test';
 import { test, expect } from '../fixtures';
 import { clearRefreshState } from '../test-helpers';
-import { waitForAppReady } from '../pages/hydration';
+import { waitForAppReady, waitForHydration } from '../pages/hydration';
 import { OrgBillingSettingsPage } from '../pages/OrgBillingSettingsPage';
 import { InvoicesPage } from '../pages/InvoicesPage';
 import { QuotesPage } from '../pages/QuotesPage';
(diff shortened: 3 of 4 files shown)


ToddHebebrand:
**Review run:** subagent pool was at its concurrency cap, so the pr-review-toolkit review was run inline (manually) rather than dispatching the code-reviewer/silent-failure-hunter/pr-test-analyzer subagents, per the fallback in the task guardrails.

**Checklist covered:**
1. **Caddyfile matcher correctness** — confirmed `expression {env.CADDY_PORTAL_DEV_ASSETS} == "1"` follows the exact established pattern of the existing `@remoteAccessClosed` matcher in the same file (unset env var → empty string → never matches). Confirmed via Caddy docs that combining `expression` + `header_regexp` + `path` directives inside one named matcher block ANDs them (all must match), matching intent. `header_regexp Referer <pattern>` without an explicit name is valid syntax (name is optional, taken from the enclosing matcher).
2. **Validated with `caddy validate`** (pinned `caddy:2-alpine` via docker) — valid configuration, zero new errors/warnings vs. the pre-change file (pre-existing `not formatted` warning on line 23 confirmed unrelated/pre-existing by diffing against unmodified `origin/main`).
3. **Prod isolation verified two ways**: rendered `docker compose config` for (a) `docker-compose.yml` alone and (b) `.dev` + `.worktree` layered, both with a dummy env file covering all `:?`-required vars — confirmed `CADDY_PORTAL_DEV_ASSETS=1` is present only in the dev/worktree render and entirely absent from the prod-only render.
4. **Security** — the whole matcher block is inert unless `CADDY_PORTAL_DEV_ASSETS=1`, which no production/CI compose file sets; the added Referer gate additionally prevents it from stealing `apps/web`'s own dev-mode Vite asset requests (same raw path shapes) even when the env gate is on.
5. **e2e spec changes** — `waitForHydration` is correctly imported (`multi-currency.spec.ts`) / already locally defined (`quote-contract-proposal.spec.ts`); both new calls sit inside the existing `try/finally` so `publicContext.close()` still runs on a hydration-guard failure; a regression will surface as a normal 20s Playwright timeout, not a hang.
6. **Repo-wide search** — no other tests/docs reference `docker/Caddyfile.prod` or `docker-compose.override.yml.dev` beyond the two already-run tests and the `worktree-stack` skill (which doesn't describe Caddy routing internals, so no doc update needed there).

**Findings:** 0 blocking, 0 non-blocking. One follow-up noted in the PR body's test plan: the new `waitForHydration` e2e assertions haven't been exercised against a live `wt-stack` (guardrail — shared stack ports were in use by sibling in-flight sessions during this run); recommend running `pnpm wt-stack test -- tests/quote-contract-proposal.spec.ts tests/multi-currency.spec.ts` once before merge to confirm the fix actually resolves hydration end-to-end.

**Tests:** `apps/api/src/config/proxyTrustCompose.test.ts` + `composeBindMounts.test.ts` (`vitest run ... --pool=threads --maxWorkers=2`) — 9/9 passed. `e2e-tests` typecheck (`npx tsc --noEmit -p tsconfig.json`) — clean. `caddy validate` — valid. `scripts/security/check-supply-chain-hardening.sh` — passed.

**Status:** review-clean (manual inline review), awaiting maintainer merge — including the live-stack e2e confirmation noted above.

PR #4426, opened 2026-09-01
fix(ci): warn on dpkg recovery failure, surface main-red-alert write failures (#3754)

### LINKED ISSUE #3754: [CI] Tauri apt deadlines can relocate the job timeout, and main-red-alert reports success when it fails to file a ci-red issue
## Description

Two follow-ups from a silent-failure audit of #3731 and #3735 (both merged). Neither is a regression — both PRs improved on what they replaced — but each leaves a path where CI reports a clean or misleading result.

## 1. The apt deadline can relocate the timeout instead of removing it — HIGH

`#3731` bounds each apt phase so a mirror stall fails fast rather than eating the job. The deadlines are sized against the job budget, but `timeout-minutes: 35` is **job-level** (`.github/workflows/ci.yml:1534`, job `rust-check` / "Rust Check (Tauri apps)"), not step-level — and the apt step is followed in the same job by Rust cache restore, Check helper, Check viewer, and Test viewer.

Worst-case apt is 27.75m:

```
per attempt   = update (2m + 30s kill grace) + install (5m + 30s)  = 8m
3 attempts                                                          = 24m
dpkg --configure -a recovery after attempts 1 and 2, 2 x (60s+30s)  =  3m
sleeps 15s + 30s                                                    = 0.75m
                                                                    -------
                                                                      27.75m
```

That leaves ~7.25m for Rust toolchain setup plus a cold `cargo check` on two webkit2gtk-linked Tauri crates plus `cargo test`. On a cache miss that is not enough. The job then dies on GitHub's generic "job timed out" with no diagnostic — which is the exact failure mode #3730 described and #3731 set out to eliminate, just moved one step later.

Options: give the apt step its own `timeout-minutes` well below the job budget; raise the job budget; or reduce the retry count now that each attempt is bounded (three bounded attempts may be more than the stall case needs).

Related nit already noted on the PR: the in-code comment states this worst case as "under 27m" — it is 27.75m. Harmless against 35m, but the number should match if someone later tunes against it.

## 2. `dpkg --configure -a || true` masks a genuinely broken runner — MEDIUM

`.github/workflows/ci.yml:1593`. The recovery exists so a `timeout`-killed apt doesn't leave the dpkg lock held, and `|| true` is right for the common case. But it also swallows failure when the runner is actually broken — corrupted dpkg status DB, disk full. That run then fails every attempt and ends at the same generic "Failed to install Tauri system dependencies after 3 attempts" as a slow mirror, misattributing root cause after ~24 wasted minutes. Emitting a `::warning::` when the recovery itself fails would separate the two without changing control flow.

## 3. main-red-alert reports success on every write failure — HIGH

`#3735` deliberately drops `set -e` so a transient API error can't redden a green run (#3692), and handles each call explicitly. That tradeoff is right and the read path is well built. But all three write paths end in `|| echo "::error::..."`, so the step's exit code is always 0 and the job shows a green check regardless.

`::error::` annotations don't change job conclusion and don't appear in the checks list — they're only visible by opening that run's Annotations tab. So the bad case is: main goes red, `gh issue create` fails, no ci-red issue exists, and nothing anywhere indicates that. The mitigation is "self-heals on the next push to main," which has no time bound — over a quiet weekend main stays red and untracked.

The no-retry-on-writes reasoning is sound (a retried create after a lost response opens a duplicate issue) so this isn't asking for retries. But the create-failure path should be visible without digging: writing to `$GITHUB_STEP_SUMMARY` would surface it on the run page, and is the cheapest option that doesn't reintroduce the redness #3735 removed.

Found during the 2026-08-20 GH queue sweep.

ToddHebebrand:
## Summary

Follow-up from the silent-failure audit in #3754. **Item 1 (job-level `timeout-minutes: 35` swallowing the step-level apt deadline) was already fixed by #3756** — this PR covers the two remaining items only.

- **Item 2 (MEDIUM):** `.github/workflows/ci.yml` — `dpkg --configure -a || true` between Tauri apt retry attempts swallowed all failures identically, so a genuinely broken runner (corrupted dpkg state, disk full) looked exactly like a slow mirror stall in the logs. Now emits an `::warning::` annotation when the recovery command fails, without changing control flow (the retry loop still proceeds).

- **Item 3 (HIGH):** `main-red-alert` job — all three write paths (`gh issue close`, `gh issue comment`, `gh issue create` against the `ci-red` tracking issue) ended in `|| echo "::error::..."`, so the step always exits 0 regardless of whether the write succeeded. `::error::` annotations don't affect job conclusion and aren't visible in the checks list — only on the run's Annotations tab. Each failure path now also writes to `$GITHUB_STEP_SUMMARY`, so a failed write (especially a failed `gh issue create`, which leaves main red and completely untracked) is visible on the run page itself. Exit code is deliberately left at 0, consistent with #3735's reasoning that this job is a notifier, not a gate, and should not reintroduce the redness that PR removed.

## Test plan

- [x] `actionlint .github/workflows/ci.yml` — no new findings (one pre-existing SC2155 warning at line 1178, confirmed present on `origin/main` before this change, unrelated to the edited sections)
- [x] `apps/api/src/config/ciSuccessGatingContract.test.ts` (parses `ci.yml` job structure) — 7/7 passed, single-fork
- [x] Manual read-through of both diffs for shell correctness (`|| { ... }` blocks under `set -uo pipefail`, no `-e`, so multi-statement failure handlers execute correctly without a stray exit)

No app code touched, so no `tsc`/unit-test surface beyond the config-contract test above.

Closes #3754

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1699,7 +1699,8 @@ jobs:
             # Bounded too: dpkg runs maintainer scripts and can block on its
             # own, and an unbounded recovery would reintroduce the very stall
             # this step exists to prevent.
-            sudo timeout --kill-after=30s 60s dpkg --configure -a || true
+            sudo timeout --kill-after=30s 60s dpkg --configure -a || \
+              echo "::warning::dpkg --configure -a recovery failed after attempt ${attempt}; this can mean a genuinely broken runner (corrupted dpkg state, disk full) rather than a slow mirror — check this run's logs if all ${attempts} attempts fail."
             sleep $((attempt * 15))
           done
           echo "::error::Failed to install Tauri system dependencies after ${attempts} attempts."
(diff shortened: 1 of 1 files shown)


ToddHebebrand:
**Review run:** subagent pool was at its concurrency cap (many sibling issue-fixers running in parallel), so I ran the review checklist inline myself instead of dispatching `pr-review-toolkit:code-reviewer`.

**Checked:**
- Bash correctness of the `cmd || { ...; { ...; } >> "$GITHUB_STEP_SUMMARY"; }` compound statements under `set -uo pipefail` (no `-e`) — valid; every statement in each block executes on failure, and the nested brace group correctly scopes the `>>` redirect to only the summary lines while the `::error::` annotation still goes to the step's stdout.
- Quoting of `$short_sha` / `$RUN_URL` / `$existing` inside the new `echo` lines — consistent with the pre-existing usage elsewhere in the same step; backticks are backslash-escaped literals for Markdown, not command substitution.
- Injection surface — none of the interpolated values are attacker-controlled (`$short_sha` from `github.sha` on a `push` to `main`, `$RUN_URL` built from static context fields, `$existing` a numeric issue number from `gh issue list --jq`). No new surface introduced.
- `$GITHUB_STEP_SUMMARY` usage — standard default-env-var + `>>` append pattern.
- Intent match against the issue text — item 2's `::warning::` doesn't change control flow (retry loop still proceeds either way); item 3 keeps the deliberate exit-0 "notifier not gate" behavior from #3735 while making a `gh issue create` failure (main red + untracked) visible on the run page instead of only in the Annotations tab. I extended the summary-write to all three write paths (close/comment/create), not just create, since the issue's own problem statement calls out "all three write paths."

**Findings:** 0 — clean.

**Tests:**
- `actionlint .github/workflows/ci.yml` (installed via brew, includes shellcheck backend) — no new findings. One pre-existing `SC2155` warning at line 1178 confirmed present on `origin/main` before this change (unrelated section).
- `apps/api/src/config/ciSuccessGatingContract.test.ts` (parses `ci.yml` job structure) — 7/7 passed, single-fork (`vitest run --pool=forks --no-file-parallelism`).
- No app code touched, so no `tsc --noEmit` surface beyond the config-contract test above.

PR #4427, opened 2026-09-01
fix(web): resolve typed lib/api mutation wrappers in no-silent-mutations guard (#3503)

### LINKED ISSUE #3503: no-silent-mutations guard can't see mutations routed through typed lib/api wrappers
## What happens

`no-silent-mutations` (`apps/web/src/lib/__tests__/no-silent-mutations.test.ts`) enforces the CLAUDE.md rule that web mutation handlers surface their outcome via `runAction`. It works by scanning component source for **lexical `fetchWithAuth` calls** with a mutating method.

The quotes web layer doesn't call `fetchWithAuth` in components. It goes through typed wrappers in `apps/web/src/lib/api/quotes.ts` (`sendQuote`, `resendQuote`, `cloneQuote`, `deleteQuote`, …), which is the established convention here and for invoices/contracts/catalog. Those wrappers are not scanned by anything, so the guard sees no mutations to check.

Demonstrated during review of #3501: replacing both `runAction(...)` blocks in `QuoteActions.tsx` with bare `await`s leaves the entire web suite green, including this guard.

## Impact

`QuoteActions.tsx` is in `TARGET_GLOBS` and absent from `runActionAllowlist.ts`, so it reads as covered — but the coverage is vacuous. Same for every other component whose mutations go through a typed API module. A regression that drops `runAction` (and with it the failure toast and the HTTP-200 `{success:false}` handling) ships green.

This is a pre-existing gap in the guard's reach, not specific to quotes — #3501 just made it visible because both new mutations land in it.

## Possible directions

1. Extend the scan to resolve imports from `lib/api/*` and treat a call to an exported wrapper whose body issues a mutating `fetchWithAuth` as a mutation at the call site. Most faithful; most work.
2. Maintain a list of known-mutating wrapper names per API module and scan for those identifiers at call sites. Cheaper, needs upkeep.
3. Invert it: assert inside `lib/api/*` that mutating wrappers are only ever *called* from within a `runAction` argument.

Option 1 is the real fix; 2 would close most of the hole today.

ToddHebebrand:
## What

`no-silent-mutations` (`apps/web/src/lib/__tests__/no-silent-mutations.test.ts`) only recognized *lexical* `fetchWithAuth(...)` calls with a mutating method. Mutations routed through typed `lib/api/*` wrappers (the established convention for quotes/invoices/contracts/catalog) were invisible to the guard — demonstrated in #3501's review, where replacing both `runAction(...)` blocks in `QuoteActions.tsx` with bare `await`s left the whole web suite, including this guard, green.

## Fix

The guard now builds a real `ts.Program` over the targeted set, resolves each `fetchWithAuth` call site's callee through the TypeScript checker (following aliased imports and re-exports), and — for any call that resolves to a declaration under `src/lib/api/` — inspects that wrapper's body for an unwrapped mutating `fetchWithAuth`. If found, the *call site* is flagged exactly as a direct mutation would be, subject to the same `runAction`/`runaction-exempt`/allowlist rules as before.

A new fixture-based self-check (`fixtures/no-silent-mutations/src/...`) proves the resolution path: a component importing a renamed re-export of a mutating wrapper is correctly flagged.

**Extra fix beyond the original patch:** the type-aware `ts.Program` only understands `.ts`/`.tsx`, but `TARGET_GLOBS` grew to include `src/pages/approvals.astro` after this fix was first drafted (#4380 review, Sept 1). Feeding an `.astro` path into `ts.createProgram` silently drops it, so `program.getSourceFile()` returned `undefined` and the guard threw. Fixed by filtering the program's `rootNames` to native TS extensions and falling back to the legacy standalone-parse scan (direct `fetchWithAuth` only, no wrapper resolution) for anything the program doesn't cover.

## Verification

- `apps/web`: `pnpm exec vitest run src/lib/__tests__/no-silent-mutations.test.ts --pool=threads --maxWorkers=2` — 119/119 passed.
- Manually reproduced the exact regression from #3501's review: replaced the `runAction(...)` wrapper around `sendQuote(...)` in the real `QuoteActions.tsx` with a bare `await` — guard went **red**, flagging the call site. Reverted — guard is **green** again.
- `npx tsc --noEmit -p apps/web` — clean.
- `npx astro check` (apps/web) — 0 errors (pre-existing unrelated warnings only).
- `runActionAllowlist.ts` untouched — allowlist semantics preserved.

Closes #3503

### DIFF
--- a/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/components/QuoteActions.tsx
+++ b/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/components/QuoteActions.tsx
@@ -1,0 +1,5 @@
+import { sendQuote as issueQuote } from '../lib/api';
+
+export async function issue(id: string): Promise<void> {
+  await issueQuote(id);
+}
--- a/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/lib/api/index.ts
+++ b/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/lib/api/index.ts
@@ -1,0 +1,1 @@
+export { getQuote, sendQuote } from './quotes';
--- a/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/lib/api/quotes.ts
+++ b/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/lib/api/quotes.ts
@@ -1,0 +1,9 @@
+import { fetchWithAuth } from '../../stores/auth';
+
+export function getQuote(id: string): Promise<Response> {
+  return fetchWithAuth(`/quotes/${id}`);
+}
+
+export function sendQuote(id: string): Promise<Response> {
+  return fetchWithAuth(`/quotes/${id}/send`, { method: 'POST' });
+}
--- a/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/stores/auth.ts
+++ b/apps/web/src/lib/__tests__/fixtures/no-silent-mutations/src/stores/auth.ts
@@ -1,0 +1,3 @@
+export function fetchWithAuth(_url: string, _options?: RequestInit): Promise<Response> {
+  throw new Error('fixture transport is never executed');
+}
--- a/apps/web/src/lib/__tests__/no-silent-mutations.test.ts
+++ b/apps/web/src/lib/__tests__/no-silent-mutations.test.ts
@@ -1,6 +1,7 @@
 /**
- * Guard: in the targeted set, every *mutating* fetchWithAuth call site must be
- * lexically wrapped by `runAction(...)` OR carry an explicit, reasoned
+ * Guard: in the targeted set, every *mutating* fetchWithAuth call site — direct
+ * or reached through an imported `src/lib/api/*` wrapper — must be lexically
+ * wrapped by `runAction(...)` OR carry an explicit, reasoned
  * `// runaction-exempt:` marker (for the legitimate aggregate / inline-feedback
  * handlers). Whole-file allowlist entries (typed service layers, transport
  * stores) are still skipped via RUN_ACTION_ALLOWLIST.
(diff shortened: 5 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, pr-test-analyzer) — both subagent spawns hit the concurrent-subagent-pool cap (many sibling issue-fixers running in parallel right now), so I ran the review checklist inline myself instead.

**Findings:** 0 blocking. 1 note for future work (not addressed, non-blocking):
- `isMutatingApiWrapper` (`no-silent-mutations.test.ts`) resolves one hop: it inspects the wrapper's own body for a direct `fetchWithAuth` call, but doesn't recursively resolve a wrapper that itself calls *another* `lib/api/*` wrapper. Checked the real convention in `apps/web/src/lib/api/quotes.ts` — every wrapper there calls `fetchWithAuth` directly, no nesting, so this isn't a live gap today, just worth knowing if the wrapper pattern ever grows a shared-helper layer.

**Inline checks performed:**
- Symbol/alias resolution (`resolvedSymbolAt`) correctly follows import → re-export chains (verified via the new fixture test, which specifically exercises a renamed import through a `lib/api/index.ts` re-export).
- `.astro` fallback (2nd commit, `99f0f4546`) is exactly the pre-existing legacy scan behavior for any file the TS program can't parse — no regression there, just restores what crashed.
- `npx eslint` clean on both the guard test file and the new fixture tree.
- Confirmed `runActionAllowlist.ts` untouched (semantics preserved per issue spec).

**Tests:** `apps/web`: `pnpm exec vitest run src/lib/__tests__/no-silent-mutations.test.ts --pool=threads --maxWorkers=2` — 119/119 passed. Manually reproduced the #3501 regression scenario (bare `await sendQuote(...)` in real `QuoteActions.tsx` in place of `runAction(...)`) — guard went red, flagging the exact call site; reverted, guard green again. `npx tsc --noEmit -p apps/web` clean. `npx astro check` — 0 errors (pre-existing unrelated warnings only).

PR #4429, opened 2026-09-01
docs: correct self-hosted server sizing table with real baseline (#3959)

### LINKED ISSUE #3959: No documented resource sizing for self-hosted deployments — guided-setup thresholds are the only figures and ignore agent count
**Gap:** there is no documented CPU/RAM/disk sizing guidance for a self-hosted Breeze install anywhere in `apps/docs/` or `README.md`.

The only figures that exist in the repo are *warning thresholds* in `scripts/guided-setup.sh:22-25`:

- `MIN_CPU_CORES=2`
- `MIN_RAM_MB=4096`
- `RECOMMENDED_RAM_MB=8192`
- `MIN_DISK_GB=20` (described at `:442` as "for images, volumes, logs, and backups")

Those are floors that trigger a warning during guided setup. They are not a sizing model, they are not surfaced in the docs (`README.md:159` mentions the check but states no numbers), and critically **they say nothing about agent count** — which is the variable a prospective self-hoster actually needs to plan against.

**Why it matters:** "how big a box do I need for N agents?" is a first-question-before-evaluating item, and today the honest answer is "undocumented." It also affects disk planning specifically: metric/telemetry retention scales with fleet size and there's no stated growth rate.

**Suggested shape:**
- A sizing table keyed on agent count (e.g. <50 / 50-500 / 500-2000 / 2000+) giving CPU, RAM, and disk.
- Call out which component is the binding constraint at each tier (Postgres connections, Redis `maxmemory`, API workers).
- Note the default Redis `--maxmemory 256mb` with `--maxmemory-policy noeviction`: at the ceiling, writes fail rather than data being silently dropped, so it needs raising as the fleet grows.
- State the disk growth rate for metric retention at a given fleet size.

**Found via** #3930, where a self-hoster explicitly asked for "recommended resource allocation for a small MSP deployment" and the accurate answer was that no such documentation exists. Real-world numbers from that deployment may be a useful first data point.

ToddHebebrand:
## Summary

- The `getting-started/prerequisites.mdx` "Server Requirements" table gave `40 GB` min / `100 GB` recommended disk figures gated on a "500+ agents" tier that was never sourced from `scripts/guided-setup.sh` or any measured deployment — it silently invented a per-agent sizing formula, which is exactly what this issue asks us not to do.
- Replace it with the two sources of truth we actually have:
  - **Floor** — `scripts/guided-setup.sh:22-25`'s warning thresholds (2 vCPU / 4 GiB RAM / 20 GB disk), explicitly labeled as a warning threshold the script enforces, not a tested capacity limit.
  - **Baseline** — the real small-MSP data point from [#3930](https://github.com/LanternOps/breeze/issues/3930) that ToddHebebrand supplied on this issue (4 vCPU / 8 GiB RAM / 60 GB SSD + snapshot headroom).
- Adds an explicit "no per-agent sizing formula" callout (per the issue's instruction not to invent one), including the Redis `256mb maxmemory` / `noeviction` note and a link to `REDIS_MAXMEMORY` in `deploy/environment.mdx`.
- Adds a caveat that the table covers the server only, not agent-side footprint.
- Maps `scripts/guided-setup.sh` → `getting-started/prerequisites.mdx` in `scripts/docs-review/mapping.json` so future threshold changes flag this doc for review.

Docs-only change. No code, no per-agent formula invented — per the issue's explicit instruction.

Closes #3959

## Test plan

- [x] `cd apps/docs && npx astro build` — 167 pages built clean, `getting-started/prerequisites/index.html` generated with no MDX errors.
- [x] Manually reviewed the diff for factual accuracy against `scripts/guided-setup.sh:22-25` and the issue's 2026-08-25 comment.

### DIFF
--- a/apps/docs/src/content/docs/getting-started/prerequisites.mdx
+++ b/apps/docs/src/content/docs/getting-started/prerequisites.mdx
@@ -18,20 +18,20 @@ This table covers the Breeze **server** (API, web, PostgreSQL, Redis, and the re
 |---|---|---|
 | **CPU** | 2 x86-64 vCPU | 4 vCPU |
 | **RAM** | 4 GiB | 8 GiB |
-| **Disk** | 20 GB SSD | 60 GB SSD, plus snapshot headroom |
+| **Disk** | 20 GiB free in the install directory | 60 GiB SSD-backed, plus snapshot headroom |
 | **OS** | Ubuntu 22.04+ / Debian 12+ | Ubuntu 24.04 LTS |
-| **Network** | Public IP, ports 80/443 open | Static IP with DNS A record |
+| **Network** | Public IP, ports 80/443 open, NTP | Static IP with DNS A record, TLS reverse proxy, NTP |
 
-The **Floor** column is what [`guided-setup.sh`](https://github.com/LanternOps/breeze/blob/main/scripts/guided-setup.sh) checks before install and warns below — it's a warning threshold the script enforces, not a tested capacity limit. Breeze will start below it, but expect resource pressure as your fleet grows.
+The **CPU / RAM / Disk Floor** values are what [`guided-setup.sh`](https://github.com/LanternOps/breeze/blob/main/scripts/guided-setup.sh) checks before install: below them it warns and, without an explicit "continue anyway" (defaulted to no, including in unattended `--yes` runs), the install stops. They're a warning threshold, not a tested capacity limit. The OS and Network rows are general recommendations — the script doesn't check them.
 
-The **Baseline** column is a real-world small-MSP data point, contributed via [#3930](https://github.com/LanternOps/breeze/issues/3930): a two-person MSP running Breeze on TrueNAS SCALE settled on 4 vCPU / 8 GiB RAM / 60 GB SSD, plus extra headroom for snapshots. The jump from the 20 GB disk floor reflects PostgreSQL, Redis's append-only file, container images, logs, upgrade image churn, and (on ZFS) snapshots all competing for the same volume — 20 GB leaves little margin for any of that.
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (subagent), verified against `scripts/guided-setup.sh`, `docker-compose.yml`/`deploy/docker-compose.prod.yml`, `apps/docs/src/content/docs/deploy/environment.mdx`, and the #3959/#3930 source comments.

**Findings:** 5 important + 3 minor raised → all addressed in f9372f52d; 0 outstanding.
- Fixed the `/deploy/environment/#redis` link — `REDIS_MAXMEMORY` actually lives under the `## Docker Deployment` heading in that file, not `## Redis`; corrected the anchor on both the Redis and arm64 cross-links.
- Corrected "no arm64 server build" (overreached the source and contradicted `environment.mdx`'s documented `docker-compose.override.yml.local-build` path) to match the issue comment's actual wording: "we do not publish arm64 server images."
- Scoped the Floor/Baseline sourcing sentences to the CPU/RAM/Disk rows only — the table's OS/Network rows were carried over unchanged from the old table and aren't part of the guided-setup.sh check or the #3930 data point; the prose no longer implies otherwise.
- Reworded the Disk floor as "20 GiB free in the install directory" (matches `detect_disk_gb()`'s actual semantics — free space in `WORK_DIR`, not a flat provisioned-disk claim) and standardized on GiB throughout (was mixing GB/GiB).
- Fixed "Breeze will start below it" — `ask_yes_no` defaults to "no" and aborts on both an unattended `--yes` run and an interactive decline, so an undersized install actually stops.
- Restored the dropped NTP prerequisite in the Network row, trimmed a maintainer-facing justification clause, and mapped `docker-compose*.yml` to this doc in `scripts/docs-review/mapping.json` since the Redis maxmemory/noeviction figures now cited here come from that file.

**Build:** `cd apps/docs && npx astro build` — 167 pages built clean both before and after the fixes, `getting-started/prerequisites/index.html` generated with no MDX errors; confirmed `id="docker-deployment"` present in the built `deploy/environment/index.html` so the corrected anchor resolves.

PR #4430, opened 2026-09-01
fix(agent/ipc): replace flaky wall-clock margins with deterministic write observation (#4322)

### LINKED ISSUE #4322: [Agent/CI] TestConnSendStalledDoesNotStarveOtherWriter has a 20ms margin and flakes under -race
## Summary

`TestConnSendStalledDoesNotStarveOtherWriter` (`agent/internal/ipc/protocol_test.go`) has a wall-clock assertion whose passing margin is only **20 ms out of a 100 ms budget**, so it fails under `-race` on a loaded runner even when the code under test behaves correctly.

**Observed (verified):** `Test Agent (race)` red on PR #4254 — a `.github/workflows/ci.yml`-only diff that cannot affect agent code.

```
--- FAIL: TestConnSendStalledDoesNotStarveOtherWriter (0.18s)
    protocol_test.go:575: writer 2 took 156.374042ms (>= writeTimeout 100ms) —
    it attempted its own write rather than failing fast on the poison re-check
FAIL	github.com/breeze-rmm/agent/internal/ipc	3.057s
```

`Test Agent (race)` is green on `main` at the same period, and a re-run of the same PR SHA was dispatched — this is a flake, not a regression.

## Root cause (verified by reading the test)

The clock starts before writer 2 can possibly make progress:

```go
writeTimeout = 100 * time.Millisecond          // :554
go func() { w1 <- client.Send(...) }()          // writer 1 takes c.mu at ~t0
time.Sleep(20 * time.Millisecond)               // :569
start := time.Now()                             // :561  == t0 + 20ms
go func() { w2 <- client.Send(...) }()          // blocks on c.mu
...
if elapsed := time.Since(start); elapsed >= writeTimeout { t.Fatalf(...) }  // :574
```

Writer 2 **cannot** return before writer 1's deadline fires at `t0 + writeTimeout`. So on a correct run:

```
elapsed ≈ (t0 + 100ms) − (t0 + 20ms) = 80ms      # assertion needs < 100ms
```

The entire margin is the 20 ms sleep. Any scheduling jitter that stretches writer 1's stall past 100 ms — routine under `-race` on a shared runner — pushes `elapsed` over the threshold. The observed 156 ms is a *correct* run measured against the wrong origin, not writer 2 burning its own timeout.

The assertion also fails to discriminate the bug it was written for. The #3007 regression (writer 2 skips the under-lock poison re-check and attempts its own write) costs roughly `2 × writeTimeout ≈ 180ms` from `start`. 156 ms sits inside that band, so a real regression and a slow-but-correct run are not separable by this threshold.

## Suggested fix

Measure writer 2 from the moment the lock can actually be acquired, not from before writer 1's deadline. The loop already reads `w1` first (`i == 0`), so:

```go
var w1Done time.Time
for i, ch := range []chan error{w1, w2} {
    select {
    case err := <-ch:
        ...
        if i == 0 { w1Done = time.Now() }
        if i == 1 {
            // writer 2 must fail on the under-lock poison re-check the instant
            // writer 1 releases c.mu — not burn a writeTimeout of its own.
            if elapsed := time.Since(w1Done); elapsed >= writeTimeout {
                t.Fatalf(...)
            }
        }
```

That keeps the assertion discriminating (a #3007 regression still burns a full `writeTimeout` *after* the lock frees) while removing the dependence on how long writer 1's deadline actually took. The `strings.Contains(err, "poisoned")` check above it is unaffected and remains the primary guard.

Note `w2` may legitimately fail *before* the test observes `w1`'s channel send, making `time.Since(w1Done)` ~0 or slightly negative — that still passes, which is correct.

`TestConnSendPoisonedFailsFast` (:399) uses the same `time.Since(start) >= writeTimeout` shape and is worth checking for the same margin problem while in there.

## Not checked

Whether this has failed on other PRs before #4254 — I only have the one occurrence plus the structural analysis above.

ToddHebebrand:
Closes #4322

## Problem

`TestConnSendStalledDoesNotStarveOtherWriter` started its clock **20 ms before writer 1's 100 ms deadline could possibly fire**, so a correct run measured ~80 ms against a 100 ms threshold. The entire passing margin was the sleep. Any scheduling jitter under `-race` on a loaded runner pushed it over — observed red on PR #4254, a `.github/workflows/ci.yml`-only diff that cannot affect agent code.

The threshold also **failed to discriminate the bug it guards**. The #3007 regression (writer 2 skips the under-lock poison re-check and attempts its own write) costs roughly `2 × writeTimeout ≈ 180 ms` from `start`; a slow-but-correct run and a regressed run overlap around the 100 ms line. The observed 156 ms was a *correct* run measured against the wrong origin.

## Fix

Both problems come from using **duration as a proxy** for what the test actually cares about. This replaces the proxy with direct observation, rather than widening the sleep (which would only move the flake threshold).

A small `writeObserverConn` wraps the client's `net.Conn` and exposes two things:

| Old (timing) | New (deterministic) |
|---|---|
| `time.Sleep(20 * time.Millisecond)` — guess at when writer 1 holds `c.mu` | receive on a channel closed when writer 1 **enters** `conn.Write`. `Send` calls `Write` under `c.mu` with the deadline already armed, so the receive is *proof* writer 1 holds the mutex and is stalled. |
| `time.Since(start) >= writeTimeout` — duration as a stand-in for "this Send did not stall on the socket again" | exact **socket-write count**. No threshold to tune. |

The write-count form is strictly *more* discriminating than what it replaces: it separates the #3007 regression from a merely slow run, which the duration threshold could not.

### Scope

Three tests in `agent/internal/ipc/protocol_test.go` shared the `elapsed >= writeTimeout` shape and are all converted:

- **`TestConnSendStalledDoesNotStarveOtherWriter`** — the actual flake. Both the sleep and the margin removed.
- **`TestConnSendPoisonedAfterWriteError`** and **`TestConnSendPoisonsOnPayloadWriteError`** — flagged in the issue as sharing the shape. **Their margins were not actually tight** (a fast-path return of a few microseconds against a 100 ms / 150 ms line), so they were not flake risks. They are converted because "reached the socket zero more times" is precisely the property the duration was standing in for, stated directly.

A sweep of the whole `agent/` tree for `time.Since(start) > d` found no other test with this defect — the remaining hits (`sessionbroker`, `agentapp`, `pamactuator`, `elevaccount`, `heartbeat`, `workspaceindex`, `remote/tools`) are prompt-return checks with generous absolute bounds (the tightest expects ~2 ms against 100 ms), not a margin equal to a preceding sleep. Left alone.

`agent/internal/ipc/protocol.go` is **untouched** — this is a test-only change.

The remaining `time.After` guards are liveness bounds, not margins: a correct run settles in one `writeTimeout`, while a starved writer never returns at all. Comment added to say so.

## Verification

**Mutation-tested first.** Each new assertion was verified to fire *independently* when the behaviour it pins is broken:

1. Removed the under-lock poison re-check from `Send()` (reproduces #3007). The poison-string check and the write-count check were confirmed to fire separately — the string check was temporarily downgraded to `t.Logf` in a scratch run so the count assertion could be observed alone:
   ```
   protocol_test.go:613: writer 2 wrote into a poisoned stream instead of failing on the under-lock re-check (issue #3007): ipc: write header: write pipe: i/o timeout
   protocol_test.go:632: expected exactly 1 socket write (writer 1's stalled header), got 2 — writer 2 attempted its own write rather than failing fast on the poison re-check
   ```
   That run also confirms writer 2 genuinely queues on `c.mu` after clearing the pre-lock fast path (its error is its *own* `write header` timeout), so the strong path is the one being exercised.

2. Removed `poison()` from the payload-write branch:
   ```
   protocol_test.go:521: second send performed 1 additional socket write(s) — payload branch did not poison
   ```

With `protocol.go` reverted to pristine (confirmed via `git diff --stat`):

| Check | Result |
|---|---|
| `go test -race -count=20 ./internal/ipc/ -run TestConnSendStalledDoesNotStarveOtherWriter` | ok, 3.437s |
| `go test -race -count=3 ./internal/ipc/` (full package) | ok, 5.555s |
| `GOMAXPROCS=1 go test -race -count=20` on all three converted tests (constrained-runner proxy) | ok, 8.359s |
| `go vet ./internal/ipc/` | clean |
| `gofmt -l internal/ipc/` | clean |

### DIFF
--- a/agent/internal/ipc/protocol_test.go
+++ b/agent/internal/ipc/protocol_test.go
@@ -8,9 +8,40 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 )
+
+// writeObserverConn wraps a net.Conn so a test can observe two things about
+// Send() without measuring wall-clock time (issue #4322):
+//
+//   - firstWrite closes the instant a Send enters its socket write. Send calls
+//     conn.Write under c.mu with the deadline already armed, so a receive on
+//     that channel is proof the writer holds the mutex and is stalled on the
+//     peer — replacing a slept guess at how long that takes.
+//   - writes counts every write that actually reached the socket, which is the
+//     exact property the old `elapsed < writeTimeout` assertions were using
+//     duration as a proxy for ("this Send did not stall on the socket again").
+//     A count has no margin to tune, so it cannot flake under -race on a loaded
+//     runner the way a 20ms-of-100ms budget did.
+type writeObserverConn struct {
+	net.Conn
+	writes     atomic.Int64
+	firstWrite chan struct{}
+	once       sync.Once
+}
+
+func newWriteObserverConn(c net.Conn) *writeObserverConn {
+	return &writeObserverConn{Conn: c, firstWrite: make(chan struct{})}
+}
+
+func (c *writeObserverConn) Write(b []byte) (int, error) {
+	c.writes.Add(1)
+	c.once.Do(func() { close(c.firstWrite) })
+	return c.Conn.Write(b)
+}
 
 func TestConnSendRecv(t *testing.T) {
 	// Create a pair of connected Unix sockets (or TCP for portability)
@@ -381,23 +412,26 @@ func TestConnSendPoisonedAfterWriteError(t *testing.T) {
 	defer serverConn.Close()
 	// Never read from serverConn so the first write stalls out.
 
-	client := NewConn(clientConn)
+	observed := newWriteObserverConn(clientConn)
+	client := NewConn(observed)
 	payload, _ := json.Marshal("x")
 
 	// First send stalls and errors at the deadline, poisoning the Conn.
 	if err := client.Send(&Envelope{ID: "1", Type: TypePing, Payload: payload}); err == nil {
 		t.Fatal("expected first send to fail at the write deadline")
 	}
+	writesAfterFirstSend := observed.writes.Load()
 
-	// Second send must fail fast (poison fast-path), well under a fresh
-	// writeTimeout — i.e. it does not stall on the socket again.
-	start := time.Now()
+	// Second send must fail fast (poison fast-path) — i.e. it does not stall on
+	// the socket again. Stated as "reached the socket zero more times" rather
+	// than "returned in under a writeTimeout": the count is what the duration
+	// was standing in for, and it carries no margin to flake on (issue #4322).
 	err := client.Send(&Envelope{ID: "2", Type: TypePing, Payload: payload})
 	if err == nil {
 		t.Fatal("expected second send to fail fast on a poisoned Conn, got nil")
 	}
-	if elapsed := time.Since(start); elapsed >= writeTimeout {
-		t.Fatalf("second send took %v — did not use the poison fast-path", elapsed)
+	if extra := observed.writes.Load() - writesAfterFirstSend; extra != 0 {
+		t.Fatalf("second send performed %d additional socket write(s) — did not use the poison fast-path", extra)
 	}
 	// Pin the reason to the poison fast-path (not some other incidental error)
 	// and confirm the original write cause is surfaced through it.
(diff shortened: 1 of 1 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (two passes — initial diff, then a follow-up pass on the third test conversion added mid-review). The `pr-test-analyzer` lane could **not** be spawned (concurrent subagent cap reached), so that checklist was run **inline** instead; both lanes' scope is recorded below.

**Findings:** 0 outstanding.
- `code-reviewer`, pass 1 (2 tests): **0 findings ≥ confidence 80**. Independently verified `gofmt`/`go vet` clean, `-race -count=50` and `-cpu=1,2,4 -count=30` green, and confirmed production code untouched.
- `code-reviewer`, pass 2 (third test): **0 findings ≥ confidence 80**. Reproduced the mutation test independently, confirmed `protocol.go` restored pristine, and validated the dynamic write-count baseline.
- Inline test-analysis lane: **1 real gap found and fixed during review** — `TestConnSendPoisonsOnPayloadWriteError` was a *third* instance of the `elapsed >= writeTimeout` shape in the same file that the first commit missed. Converted in the amended commit (`0d527481c`). No other findings.

**Key correctness point (raised by the reviewer, worth recording):** the exact-write-count assertion is sound **only because all three tests dial `net.Pipe()`** — synchronous and unbuffered, so a `Write` cannot return until a matching `Read`. Had they used the TCP `createSocketPair` helper used elsewhere in this file, a small payload would fit in the kernel send buffer, both writes could succeed instantly, and the exact count *would* have become a new flake source in the other direction. Anyone porting this pattern to another test needs to check that first.

**Evidence — mutation-tested before trusting any green.** Each new assertion was proven to fire independently when the behaviour it pins is broken:
- Removed the under-lock poison re-check from `Send()` (reproduces #3007) → both the poison-string check and the write-count check fire separately (string check temporarily downgraded to `t.Logf` in a scratch run to observe the count assertion alone). That run also showed writer 2's error was its *own* `write header: i/o timeout`, confirming it genuinely queues on `c.mu` after clearing the pre-lock fast path — the strong path is the one being exercised, not the fast path.
- Removed `poison()` from the payload-write branch → `second send performed 1 additional socket write(s) — payload branch did not poison`.

**Tests** (all foreground, `protocol.go` confirmed pristine via `git diff --stat` first):

| Check | Result |
|---|---|
| `go test -race -count=20 ./internal/ipc/ -run TestConnSendStalledDoesNotStarveOtherWriter` | ok, 3.437s |
| `go test -race -count=3 ./internal/ipc/` (full package) | ok, 5.555s |
| `GOMAXPROCS=1 go test -race -count=20`, all three converted tests | ok, 8.359s |
| Reviewer's independent `-race -count=50` + `-cpu=1,2,4 -count=30` | pass |
| `go vet ./internal/ipc/` / `gofmt -l internal/ipc/` | clean |

**Scope sweep:** grepped the whole `agent/` tree for `time.Since(start) > d`. The remaining hits (`sessionbroker`, `agentapp`, `pamactuator`, `elevaccount`, `heartbeat`, `workspaceindex`, `remote/tools`) are prompt-return checks with generous absolute bounds — the tightest expects ~2ms against a 100ms line — not a margin equal to a preceding sleep. **Not** the #4322 defect; deliberately left alone.

**Status:** review-clean. CI still in flight at time of writing (no failures so far); `Test Agent (race)` is the job that was flaking, so it's worth eyeballing that one specifically before merge. Awaiting maintainer merge — not merging, and leaving #4322 open.

PR #4409, opened 2026-09-01
fix(ops): make docker-compose.monitoring.yml composable with prod stack (#4362)

ToddHebebrand:
## Summary

`docker-compose.monitoring.yml`'s `postgres-exporter` service hard-depended
on a `postgres` service (`depends_on: postgres: condition: service_healthy`)
that only exists in the root `docker-compose.yml` (dev stack). It does not
exist in `deploy/docker-compose.prod.yml`, which intentionally has no local
Postgres — production uses a managed database.

`docker compose -f deploy/docker-compose.prod.yml -f docker-compose.monitoring.yml config`
therefore failed immediately with `service "postgres-exporter" depends on
undefined service "postgres": invalid compose project`. This is exactly the
overlay `scripts/prod/deploy.sh` applies when `ENABLE_MONITORING=true` (its
default), so any strict-prod deploy with monitoring enabled broke at the
`compose config` validation step (`scripts/prod/deploy.sh:285`), and the
monitoring stack never actually started via that path.

## Changes

- `docker-compose.monitoring.yml`: drop the hard `depends_on: postgres`
  (the exporter now retries via `restart: unless-stopped` until its DSN is
  reachable — no ordering dependency needed). `DATA_SOURCE_NAME` now reads
  from a new `POSTGRES_EXPORTER_DSN` override, falling back to the existing
  local dev DSN unchanged, since prod has no `postgres` hostname to resolve
  against.
- `scripts/prod/deploy.sh`: require `POSTGRES_EXPORTER_DSN` alongside the
  existing `METRICS_SCRAPE_TOKEN`/`GRAFANA_ADMIN_PASSWORD` checks when
  `ENABLE_MONITORING=true`, so a missing value fails fast with a clear
  message instead of surfacing as a buried Compose interpolation error.
- `deploy/.env.example` + `docs/operations/DEPLOY_PRODUCTION.md`: document
  the new var (and the previously-undocumented `GRAFANA_ADMIN_PASSWORD`
  requirement, which `deploy.sh` already enforced but `.env.example` never
  mentioned).
- `scripts/security/check-supply-chain-hardening.sh`: extend the existing
  monitoring-compose assertions with a behavioral proof — `docker compose
  config` must resolve for **both** base-file combinations (root
  `docker-compose.yml` + overlay, and `deploy/docker-compose.prod.yml` +
  overlay), gated on Docker availability like the existing coturn probe in
  `check-relay-edge-hardening.sh`. Backed by two new placeholders in
[...]

Issue #4362, opened 2026-09-01
[Ops] docker-compose.monitoring.yml is not composable with deploy/docker-compose.prod.yml — breaks deploy.sh with ENABLE_MONITORING=true


ToddHebebrand:
Found while fixing #4278 (verify-monitoring.sh's wrong compose path), via PR #4354 review.

`docker-compose.monitoring.yml`'s `postgres-exporter` service has a hard `depends_on: postgres: condition: service_healthy` (line ~144). The `postgres` service it depends on only exists in the **root** `docker-compose.yml` (the dev stack) — it does not exist in `deploy/docker-compose.prod.yml`, which intentionally has no local Postgres (prod uses a managed database; see `deploy/docker-compose.prod.yml:374`).

**Impact:** `docker compose -f deploy/docker-compose.prod.yml -f docker-compose.monitoring.yml ...` is an invalid compose project — `docker compose config` fails immediately with:

```
service "postgres-exporter" depends on undefined service "postgres": invalid compose project
```

This is exactly the two-file overlay `scripts/prod/deploy.sh` uses when `ENABLE_MONITORING=true` (its default — see `scripts/prod/deploy.sh:69-71,193-199`). Concretely:

- `scripts/prod/deploy.sh:285` runs `compose config >/dev/null` as part of its validation step, which hits this error on any strict-prod run with monitoring enabled.
- The monitoring stack (Grafana/Prometheus/Loki/postgres-exporter) never actually starts via `deploy.sh` in that configuration.
- Docs (`docs/operations/DEPLOY_PRODUCTION.md`) describe monitoring as part of the standard deploy ("Starts the full stack (edge, app, billing, monitoring, Loki/Promtail)"), so this is a real gap between documented and actual behavior.

Likely unnoticed in practice because droplets deploy via `/opt/breeze`'s own compose setup rather than `scripts/prod/deploy.sh` directly (see the "Known drift" note in root `CLAUDE.md` re: `deploy/docker-compose.prod.yml` vs the deployed compose file).

[...]

PR #4432, opened 2026-09-01
test(security): exercise the 'upgrade present' half of the apk hardening guard (#4271)

### LINKED ISSUE #4271: apk hardening guard: test suite doesn't exercise the 'upgrade present' half; rejection asserts lack specificity
## Summary

The guard added with the CVE-2026-14456 repair (`scripts/security/check-supply-chain-hardening.sh` + `apps/api/src/config/dockerfileOpensslUpgrade.test.ts`) has two test-side gaps. Neither is a live bypass, but both let a real regression in the control ship green. Surfaced during review of #4257 (the C1/C2 script bugs from that review are already fixed on the PR branch in `801bdc45`; this issue tracks the remaining test hardening).

## Gap 1 — the "upgrade must be present" half is never exercised behaviorally

`--check-apk` mode calls `reject_unaudited_apk` **directly** (`check-supply-chain-hardening.sh:73-76`), not `require_audited_openssl_upgrade` (`:65`). The `require_grep` half — "the audited upgrade line is actually present" — is therefore never reached by any test.

Consequence (VERIFIED by the reviewers via mutation): deleting the `require_grep` call from `require_audited_openssl_upgrade` leaves the whole vitest suite green. A read-executor Dockerfile that silently loses its OpenSSL upgrade — i.e. the original #4246 bug returning — would pass the full hardening script, caught only by the TS reimplementation in `it.each(COVERED)`. The two-layer defence collapses to one without a red.

## Gap 2 — rejection assertions carry no discriminating power

`dockerfileOpensslUpgrade.test.ts:320` asserts only `.not.toBe(0)`. A `spawnSync` failure yields `status = null` and a missing script yields `127`; both satisfy `!== 0`. Nothing asserts *why* the exit was non-zero, so a future change that makes the script fail rejected inputs for an unrelated reason (a `set -u` trip, an arg-handling regression, a bash-version difference in CI) keeps all rejection cases green.

(Correction to one review claim: the test case does *not* fully pass with the script deleted — the `accepted` loop runs first, asserts `toBe(0)`, and goes red at `127`. So total script loss is caught; the narrower "rejection asserts nothing specific" gap is what stands.)

## Proposed fix

1. Add a `--require-apk` mode (or have the test invoke the script without `--check-apk` against a temp file) so `require_audited_openssl_upgrade` — the present-half — is exercised, with fixtures for missing / commented-out / half-present upgrades.
2. Tighten the rejection assertion to be specific:
   ```ts
   expect(result.error, 'harness failed to run the policy').toBeUndefined();
   expect(result.status, `unexpectedly accepted: ${source}`).toBe(1);
   expect(result.stderr).toMatch(/contains an unaudited apk invocation/);
   ```
   The reviewer confirmed `stderr` carries that string for all rejected fixtures under the real script and none under a missing one, so it discriminates today.

## Blast radius

Test-only; the shipped control is sound after `801bdc45`. Priority: normal — it protects the control from silent future regression.

ToddHebebrand:
## Summary

The apk credential-boundary guard (`scripts/security/check-supply-chain-hardening.sh`) has two test-side gaps flagged in #4271:

1. **`--check-apk` mode never exercises the "upgrade present" half.** It calls `reject_unaudited_apk` directly, never `require_audited_openssl_upgrade`'s `require_grep` call. Deleting that `require_grep` call left the whole vitest suite green (verified by mutation before/after this fix — see test plan).
2. **Rejection assertions carried no discriminating power.** `dockerfileOpensslUpgrade.test.ts` asserted only `.not.toBe(0)`, which is satisfied by a `spawnSync` harness failure (`status: null`) or a missing-script exit (`127`) just as much as by the intended policy rejection.

## Changes

- `scripts/security/check-supply-chain-hardening.sh`: add a `--require-openssl-upgrade` mode that runs the full `require_audited_openssl_upgrade` rule (both `reject_unaudited_apk` and the `require_grep` presence check), parallel to the existing `--check-apk` mode.
- `apps/api/src/config/dockerfileOpensslUpgrade.test.ts`:
  - New `checkOpensslUpgradePolicy` helper invoking `--require-openssl-upgrade`.
  - New test exercising missing / commented-out / half-present / correct upgrade fixtures against the full rule, asserting both exit code and the specific stderr reason.
  - Tightened `'rejects every apk mutation except the exact two-package repair'` to assert `result.error` is undefined, `result.status === 1`, and `result.stderr` matches `/contains an unaudited apk invocation/`, instead of the previous bare `.not.toBe(0)`.

Test-only change; the shipped control itself is unmodified in behavior (the new script mode is additive — `--check-apk` and `require_audited_openssl_upgrade` are untouched).

## Coordination with in-flight PRs

Two other open PRs touch the same two files:
- #4422 (#4272): adds the actions executor to `CREDENTIAL_BOUNDARY_DOCKERFILES` + a `require_audited_openssl_upgrade "$ACTIONS_EXECUTOR_DOCKERFILE"` script block, around `check-supply-chain-hardening.sh:353+` (M365 actions-executor section).
- #4409 (#4362): adds a compose-config behavioral check around `check-supply-chain-hardening.sh:679+` (monitoring overlay section).

This PR's script edit is scoped to the top of the file (the `--check-apk`/`--require-openssl-upgrade` mode dispatch, right after `require_audited_openssl_upgrade`'s definition, before `extract_yaml_job`), and its test edit only adds a new helper/test and tightens one existing assertion — no shared lines with either PR's diff. Should merge cleanly after either lands.

## Test plan

- `pnpm exec vitest run src/config/dockerfileOpensslUpgrade.test.ts --pool=threads --maxWorkers=2` — 17/17 passing.
- Mutation check: manually deleted the `require_grep` call from `require_audited_openssl_upgrade` in the script, re-ran the suite — the new test went red (`expected +0 to be 1`) exactly as the issue predicted, confirming it closes the gap. Restored the script and re-ran — 17/17 green again.
- `shellcheck scripts/security/check-supply-chain-hardening.sh` — clean.
- `npx eslint apps/api/src/config/dockerfileOpensslUpgrade.test.ts` — clean.
- `node --max-old-space-size=8192 ./node_modules/typescript/bin/tsc --noEmit -p .` (apps/api) — clean (plain `tsc --noEmit` OOMs on this monorepo's default heap regardless of this change; increasing heap resolves it).

Closes #4271

### DIFF
--- a/apps/api/src/config/dockerfileOpensslUpgrade.test.ts
+++ b/apps/api/src/config/dockerfileOpensslUpgrade.test.ts
@@ -134,6 +134,27 @@ function checkApkPolicy(source: string) {
   }
 }
 
+/**
+ * Unlike checkApkPolicy (--check-apk, which only calls reject_unaudited_apk
+ * and therefore never reaches the require_grep "upgrade must be present" half
+ * of require_audited_openssl_upgrade), this invokes --require-openssl-upgrade
+ * so both halves of the credential-boundary rule run. See issue #4271.
+ */
+function checkOpensslUpgradePolicy(source: string) {
+  const directory = mkdtempSync(path.join(tmpdir(), 'breeze-openssl-upgrade-policy-'));
+  const dockerfile = path.join(directory, 'Dockerfile');
+  writeFileSync(dockerfile, source);
+  try {
+    return spawnSync(
+      'bash',
+      [path.join(REPO_ROOT, HARDENING_SCRIPT), '--require-openssl-upgrade', dockerfile],
+      { encoding: 'utf8' },
+    );
+  } finally {
+    rmSync(directory, { recursive: true, force: true });
+  }
+}
+
 interface Stage {
   /** Lowercased `AS` alias, or undefined for an unnamed stage. */
   alias?: string;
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:code-reviewer` (async subagent) + inline silent-failure-path trace (silent-failure-hunter subagent failed to spawn — concurrent subagent pool at cap from sibling issue-fixer runs — so I ran that checklist manually, see below).

**Findings:** 0 blocking. code-reviewer independently verified via mutation testing (not just inspection):
- Deleted `require_grep` from `require_audited_openssl_upgrade` → full `src/config` suite (556 tests/18 files): only the new test failed, confirming the gap was real repo-wide.
- Dropped the `^` anchor from `AUDITED_APK_UPGRADE` → new test caught it, proving the `commentedOut` fixture isn't redundant with `missing`.
- Changed `fail()` to `exit 2` → both the new test and the tightened rejection-assertion loop caught it; under the old bare `.not.toBe(0)` the loop would have stayed green.
- `git merge-tree` against both #4422 and #4409 head SHAs: zero conflicts with either (nearest hunk gap 11 lines vs #4422, no overlap at all vs #4409).
- One non-blocking cosmetic nit noted (near-duplicate helper function in the test file) — left as-is, doesn't warrant a diff for a test-only PR.

Inline silent-failure trace: `fail()` unconditionally calls `exit 1`, so every rejection path (via `reject_unaudited_apk` or `require_grep`) inside `require_audited_openssl_upgrade` terminates with status 1 deterministically — no ambiguous/swallowed exit codes, and `set -euo pipefail` doesn't interact adversarially with the new `--require-openssl-upgrade` dispatch block (it mirrors the existing `--check-apk` block structurally).

**Tests:** `apps/api` affected suite green — `pnpm exec vitest run src/config/dockerfileOpensslUpgrade.test.ts --pool=threads --maxWorkers=2`, 17/17 passed. Confirmed the new test is a real regression-catcher by manually deleting the `require_grep` call from the script pre-PR and re-running: it failed as predicted (`expected +0 to be 1`); restored and re-verified green. `tsc --noEmit` clean on `apps/api` (plain invocation OOMs on this monorepo regardless of this change — resolved with `--max-old-space-size=8192`, unrelated to this diff). `shellcheck` and `eslint` clean on the touched files.

PR #4434, opened 2026-09-01
test(tickets): real-Postgres integration test for ticket_outbox rollback atomicity (#4219)

### LINKED ISSUE #4219: AI agents: hostile-ticket integration test — outbox atomicity on rollback
Follow-up to #3828 (PR #4195). The branch review noted the 'rollback leaves no outbox row' claim is proven structurally (ambient transaction handle) but only asserted by mocked call-ordering tests. Add an integration test that rolls back a ticket mutation and asserts zero `ticket_outbox` rows.

ToddHebebrand:
## Summary

Follow-up to #3828 (PR #4195). The branch review that landed the transactional
outbox (`ticket_outbox`) noted the "rollback leaves no outbox row" claim was
proven only **structurally**: `ticketService.test.ts`'s
`ticket_outbox — same-transaction write (source guarantee)` describe block
greps `writeTicketOutbox`'s source for the *absence* of
`withSystemDbAccessContext`/`runOutsideDbContext` wrapping. That mocked unit
harness stubs `db.transaction`/`withDbAccessContext` as identity
passthroughs and can never behaviorally distinguish "ran in the ambient
request tx" from "ran in a separately-opened system tx" against a real
Postgres transaction.

This PR adds the missing real-Postgres integration test:
`apps/api/src/__tests__/integration/ticketOutboxRollbackAtomicity.integration.test.ts`

- **Hostile-ticket case**: drives `changeTicketStatus` through the actual
  `withDbAccessContext` wrapper (the same one `authMiddleware` puts around
  every request in `apps/api/src/middleware/auth.ts`'s `dispatch` closure).
  The status change and its `ticket_outbox` write both run to completion, then
  the same request-handler callback throws — simulating a hostile/malformed
  ticket payload causing a later step in the same handler to fail. Asserts
  **zero** `ticket_outbox` rows and the ticket's status change both roll back
  together.
- **Positive control**: the identical call *without* the injected failure
  commits exactly one outbox row and the status change — so the rollback
  assertion can't pass vacuously (a broken fixture/query that always finds
  zero rows would fail this case).
- **Fixture sanity check**: confirms the seeded org is actually visible under
  the RLS-scoped context used, so the scenario is exercising real RLS, not
  bypassing it.

### Proof the test discriminates

I temporarily reintroduced the exact regression this test guards against —
wrapping `writeTicketOutbox`'s insert in
`runOutsideDbContext(() => withSystemDbAccessContext(...))`, which escapes
the ambient request transaction (mirroring what `createAuditLogAsync`
deliberately does, and what `writeTicketOutbox`'s doc comment explicitly says
it must never do). Both non-control test cases failed immediately — not with
a subtle assertion mismatch, but with a foreign-key violation, because the
escaped transaction could no longer see the not-yet-committed `tickets` row:

```
PostgresError: insert or update on table "ticket_outbox" violates foreign key constraint "ticket_outbox_ticket_id_fkey"
```

I then reverted the mutation (`git diff` on `ticketService.ts` shows no
changes) and reran to confirm green.

## Test plan

- [x] New integration test, single file, run directly against an isolated
      per-worktree Postgres/Redis stack (`pnpm test-stack up`):
      `npx vitest run --config vitest.integration.config.ts --pool=threads --maxWorkers=1 src/__tests__/integration/ticketOutboxRollbackAtomicity.integration.test.ts`
      → 3 passed
- [x] Discrimination proof: temporarily reintroduced the regression, watched
      both non-control cases fail (FK violation), reverted, confirmed green
      again.
- [x] Sibling mocked suite still green:
      `npx vitest run --pool=forks --maxWorkers=1 src/services/ticketService.test.ts`
      → 183 passed
- [x] `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api` → clean
- [x] File placement confirmed to match the integration config's include glob
      (`src/__tests__/integration/**/*.test.ts` in
      `apps/api/vitest.integration.config.ts`), so it actually runs in CI.

Closes #4219

### DIFF
--- a/apps/api/src/__tests__/integration/ticketOutboxRollbackAtomicity.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ticketOutboxRollbackAtomicity.integration.test.ts
@@ -1,0 +1,172 @@
+/**
+ * Real-Postgres proof for #4219 (follow-up to #3828 / PR #4195): the
+ * `ticket_outbox` row a ticket mutation writes must roll back WITH the
+ * mutation itself when the surrounding request transaction later fails —
+ * "hostile ticket" scenario: a ticket write succeeds and its outbox row is
+ * inserted, but something else in the SAME request handler then throws, and
+ * the whole thing must roll back as one unit.
+ *
+ * The branch review that opened this issue found the "rollback leaves no
+ * outbox row" claim proven only STRUCTURALLY, by `ticketService.test.ts`'s
+ * `ticket_outbox — same-transaction write (source guarantee)` describe block,
+ * which greps `writeTicketOutbox`'s source for the ABSENCE of
+ * `withSystemDbAccessContext`/`runOutsideDbContext` wrapping. That is a real
+ * but indirect guarantee — it can't behaviorally distinguish "ran in the
+ * ambient request tx" from "ran in a separately-opened system tx" because the
+ * mocked unit harness stubs `db.transaction`/`withDbAccessContext` as identity
+ * passthroughs and never touches a real Postgres transaction. This test
+ * drives the REAL `withDbAccessContext` (the same wrapper `authMiddleware`
+ * puts around every request — see `apps/api/src/middleware/auth.ts`'s
+ * `dispatch` closure) around a real `changeTicketStatus` call, and forces the
+ * SAME ambient transaction to fail immediately after that call returns —
+ * exactly what a route handler with a hostile/malformed payload does when a
+ * later step in the same handler throws.
+ *
+ * Two cases:
+ *  1. Hostile rollback: `changeTicketStatus` writes its `ticket_outbox` row
+ *     and returns successfully, then the request handler throws. Zero
+ *     `ticket_outbox` rows and zero committed status changes must survive —
+ *     proving both the outbox row and the ticket mutation it announces rolled
+ *     back TOGETHER, not just structurally-argued to.
+ *  2. Positive control (discriminates the test itself): the identical call
+ *     WITHOUT the injected failure commits normally and leaves exactly one
+ *     `ticket_outbox` row and the updated ticket status. Without this, case 1
+ *     would pass vacuously if the fixture/query were wrong in a way that
+ *     always finds zero rows regardless of what actually happened.
+ *
+ * What would make this test FAIL (and did, pre-#3828-fix if `writeTicketOutbox`
+ * were ever changed to wrap its insert in `withSystemDbAccessContext` /
+ * `runOutsideDbContext`, escaping the ambient request transaction): case 1's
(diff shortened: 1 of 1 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (single async agent — the review checklist was run via the standard toolkit reviewer; a broader multi-lens dispatch wasn't needed given this PR is a single additive test file with no production-code changes)
**Findings:** 1 should-fix + 2 nits raised → should-fix and one nit addressed in c36da1ae2; the remaining nit (whether to add a "for discoverability only" entry in `vitest.integration.config.ts`'s include list) left as-is — the reviewer confirmed this convention is applied inconsistently across the ~20 files in `src/__tests__/integration/` already and called it genuinely optional; the file is already covered by the directory's shared glob so it runs in CI either way.
  - Should-fix (addressed): the file header overstated which regression the test behaviorally catches. `withDbAccessContext`/`withSystemDbAccessContext` early-return into the ambient transaction when a context is already open, so wrapping `writeTicketOutbox`'s insert in `withSystemDbAccessContext` ALONE would just nest and still roll back normally — that mis-wrap stays caught only by the structural grep in `ticketService.test.ts`, not behaviorally by this test. Only the `runOutsideDbContext` escape form (which genuinely exits both AsyncLocalStorage stores) is what this test catches. Header now states this precisely.
  - Nit (addressed): moved the in-transaction-callback status assertion out of the `withDbAccessContext` callback so a real assertion failure surfaces as itself instead of being masked as "wrong error thrown" by `.rejects.toThrow(injectedError)`.
  - Nit (not addressed, optional): discoverability-listing convention inconsistency noted above.
  - Reviewer confirmed: no vacuous assertions, no contextless RLS-deny traps (every read wrapped in `withSystemDbAccessContext`/`withDbAccessContext`), correct test isolation via TRUNCATE-CASCADE reachability, correct CI placement (matches `vitest.integration.config.ts`'s include glob, `DATABASE_URL` set in the `integration-test` CI job so `it.runIf` actually executes rather than silently skipping), and no flakiness risk.
**Tests:** New integration test green (`vitest run --config vitest.integration.config.ts --pool=threads --maxWorkers=1 src/__tests__/integration/ticketOutboxRollbackAtomicity.integration.test.ts`, 3 passed) against an isolated per-worktree Postgres/Redis stack; sibling mocked suite unaffected (`vitest run --pool=forks --maxWorkers=1 src/services/ticketService.test.ts`, 183 passed); `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api` clean. Discrimination independently verified pre-review by temporarily reintroducing the exact regression the test guards against (escaping the ambient transaction via `runOutsideDbContext`) — both non-control cases failed with an FK violation — then reverted.

PR #4435, opened 2026-09-01
fix(quotes): theme PDF header fonts, tighten docTheme type, fix vacuous test (#3546)

### LINKED ISSUE #3546: Proposal presentation follow-ups: condensed-theme font remnants + small test gaps
Follow-ups from the proposal presentation system final review (#3526 / #3535 / #3539 / #3545), none blocking, all small:

- **Condensed theme: PDF header labels still Helvetica.** `FROM` / `PREPARED FOR` / `PREPARED BY` labels and `renderLineTable` typography are hard-coded `Helvetica-Bold`, so a condensed document renders them next to Barlow Condensed headings (mixed-font header block). Swap to the theme font table (`fonts.heading/body`) — classic resolves to the same strings, so the classic regression harness must stay green.
- **Vacuous 9-column validator test.** `packages/shared/src/validators/quotes.test.ts` (cols9 case) uses `rows: []`, which fails `rows.min(1)` before the column cap is ever checked — the test passes even if `columns.max(8)` is deleted. Give it one valid 9-cell row.
- **PDF table: multi-run non-left-aligned cells untested.** Bold/link runs inside centered/right-aligned cells rely on pdfkit continued-text behavior with no covering test.
- **`documentShell.tsx` `docTheme` typed as bare `string`** — tighten to `DocumentThemeId | null` so typos fail the build.

ToddHebebrand:
## Summary

Three small follow-ups from the proposal presentation system final review (#3526 / #3535 / #3539 / #3545):

- **Condensed theme: PDF header/terms labels still Helvetica.** `apps/api/src/services/quotePdf.ts` — the FROM / PREPARED FOR / PREPARED BY header block (cover-page renderer + main render's first-page header), plus the intro notes, image caption, and TERMS & CONDITIONS block, hard-coded `Helvetica`/`Helvetica-Bold` instead of the resolved theme fonts, so a condensed document rendered these next to Barlow Condensed headings (mixed-font blocks). Swapped to `fonts.heading.bold` / `fonts.body.regular`, matching the convention already used for the title and PROPOSAL eyebrow elsewhere in the same function. `classic` resolves to the same Helvetica strings, so the classic regression harness (`quotePdf.classicRegression.test.ts`) is byte-for-byte unchanged. `renderLineTable`/`renderRecurringSummary`'s own hard-coded fonts (line-item table rendering) are a separate, larger change — `renderLineTable` doesn't currently take a `fonts` param — tracked in #4438.
- **`documentShell.tsx` `docTheme` typed as bare `string`.** Tightened `DocumentPaper`'s `docTheme` prop to `DocumentThemeId | null` (from `@breeze/shared`, mirroring `apps/api/src/services/documentThemes.ts`'s server-side source of truth) so a typo'd theme id fails the build instead of silently falling through to `classic`. This surfaced that `PublicInvoiceDetail.branding.theme` (and, for consistency, its sibling `pageSize`) in `apps/portal/src/lib/api.ts` were also bare `string` — tightened both to `DocumentThemeId`/`DocumentPageSize`.
- **Vacuous 9-column validator test.** `packages/shared/src/validators/quotes.test.ts` — the `cols9` case (rejects >8 columns) used `rows: []`, which fails the schema's `rows.min(1)` before `columns.max(8)` is ever evaluated, so the assertion passed even with the column cap deleted. Replaced with a real row of 9 cells matching the 9 columns so the only remaining violation is the column cap. Also added the same-technique `rows.max(100)` case the test's own name promised but never covered.

Scoped tightly to avoid conflicting with open PR #4408 (also touching `quotePdf.ts`, in the `doc.image()` catch blocks / cover-image save-restore `finally` — a disjoint region from this diff; verified with a three-way `git merge-file` — zero conflicts).

Not included (filed as follow-up #4438): the `renderLineTable`/`renderRecurringSummary` typography and the multi-run non-left-aligned-cell PDF table test gap — both larger changes than the items above.

Closes #3546
Refs #4438

## Test plan

- [x] `packages/shared`: `vitest run src/validators/quotes.test.ts` — 56 passed. Verified non-vacuous by temporarily removing `columns.max(8)` from the schema and confirming the `cols9` assertion goes red, then restoring it.
- [x] `apps/api`: `vitest run src/services/quotePdf.test.ts src/services/quotePdf.classicRegression.test.ts` — 51 passed (classic regression green).
- [x] `apps/portal`: `vitest run src/components/portal/PublicQuoteView.test.tsx src/components/portal/PublicQuoteView.superseded.test.tsx` — 11 passed.
- [x] `tsc --noEmit -p apps/api` — clean.
- [x] `tsc --noEmit -p apps/portal` — clean.
- [x] `eslint` on all four changed files — clean.
- [x] `tsc --noEmit -p packages/shared` reports 3 pre-existing errors in `quotes.test.ts` unrelated to this change (verified identical on unmodified `origin/main` via `git stash`).

## Review

Reviewed via `pr-review-toolkit:code-reviewer`. Two "Important" findings, both addressed in the second commit (extended the font swap to the terms/intro/caption block; filed #4438 for the genuinely out-of-scope remainder rather than leaving it unlinked under `Closes #3546`). Two "Minor" findings, both addressed (rows>100 test case; `pageSize` type tightening).

### DIFF
--- a/apps/api/src/services/quotePdf.ts
+++ b/apps/api/src/services/quotePdf.ts
@@ -751,13 +751,13 @@ async function renderCoverPage(
 
   const preparedForName = cp.preparedForName ?? quote.billToName ?? null;
   if (preparedForName) {
-    doc.fillColor('#9ca3af').fontSize(9).font('Helvetica-Bold').text('PREPARED FOR', c.left, rowY);
-    doc.fillColor('#111827').fontSize(12).font('Helvetica-Bold').text(preparedForName, c.left, rowY + 14, { width: colW });
+    doc.fillColor('#9ca3af').fontSize(9).font(fonts.heading.bold).text('PREPARED FOR', c.left, rowY);
+    doc.fillColor('#111827').fontSize(12).font(fonts.heading.bold).text(preparedForName, c.left, rowY + 14, { width: colW });
     // Start the address at the name's real bottom edge (doc.y) — a name long
     // enough to wrap painted the address on top of its second line when this
     // assumed a fixed one-line name height.
     let addrY = Math.max(rowY + 30, doc.y + 4);
-    doc.fillColor('#4b5563').fontSize(9).font('Helvetica');
+    doc.fillColor('#4b5563').fontSize(9).font(fonts.body.regular);
     for (const line of addressLines(quote.billToAddress as BillToAddress | null)) {
       doc.text(line, c.left, addrY, { width: colW });
       addrY += 12;
@@ -768,10 +768,10 @@ async function renderCoverPage(
   // explicit `false` as "show" so a legacy/loosely-typed value degrades safely.
   if (cp.showPreparedBy !== false) {
     const seller = (quote.sellerSnapshot as SellerSnapshot | null) ?? null;
-    doc.fillColor('#9ca3af').fontSize(9).font('Helvetica-Bold').text('PREPARED BY', rightX, rowY);
-    doc.fillColor('#111827').fontSize(12).font('Helvetica-Bold').text(seller?.name ?? partnerName, rightX, rowY + 14, { width: colW });
+    doc.fillColor('#9ca3af').fontSize(9).font(fonts.heading.bold).text('PREPARED BY', rightX, rowY);
+    doc.fillColor('#111827').fontSize(12).font(fonts.heading.bold).text(seller?.name ?? partnerName, rightX, rowY + 14, { width: colW });
     let addrY = Math.max(rowY + 30, doc.y + 4); // same wrap-safe start as PREPARED FOR
-    doc.fillColor('#4b5563').fontSize(9).font('Helvetica');
+    doc.fillColor('#4b5563').fontSize(9).font(fonts.body.regular);
     for (const line of sellerAddressLines(seller)) {
       doc.text(line, rightX, addrY, { width: colW });
       addrY += 12;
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (single subagent — pool was near capacity, ran this one review inline rather than fanning out silent-failure-hunter/pr-test-analyzer as well)
**Findings:** 4 raised (2 important, 2 minor) → all 4 addressed in commit 114080a38:
- Important: font swap stopped short of intro/caption/terms lines in the same function → extended (5 more `Helvetica`/`Helvetica-Bold` → `fonts.body.regular`/`fonts.heading.bold`).
- Important: `Closes #3546` would auto-close the issue with 2 of its 4 bullets (`renderLineTable` typography, multi-run cell test) still unaddressed and unlinked → filed #4438 to track them, kept `Closes #3546` since everything else is now done or moved.
- Minor: test name promised `rows.max(100)` coverage it didn't have → added.
- Minor: `PublicInvoiceDetail.branding.pageSize` left as bare `string` next to the newly-tightened `theme` → tightened to `DocumentPageSize`.

Reviewer also verified (not just asserted): font swap is a genuine no-op for `classic` (same Helvetica strings), all 4 `DocumentPaper` call sites still typecheck, the `cols9`/`rows101` fixtures trip only the intended cap, and — via a three-way `git merge-file` against open PR #4408's head — zero conflicts with #4408's disjoint `doc.image()` catch-block changes in the same file.

**Tests:** `packages/shared` `vitest run src/validators/quotes.test.ts` 56 passed · `apps/api` `vitest run src/services/quotePdf.test.ts src/services/quotePdf.classicRegression.test.ts` 51 passed (classic regression green) · `apps/portal` `vitest run src/components/portal/PublicQuoteView.test.tsx src/components/portal/PublicQuoteView.superseded.test.tsx` 11 passed · `tsc --noEmit -p apps/api` clean · `tsc --noEmit -p apps/portal` clean · `eslint` on all 4 changed files clean · `tsc --noEmit -p packages/shared` shows 3 pre-existing errors in `quotes.test.ts` unrelated to this change (confirmed identical on unmodified `origin/main`).

PR #4436, opened 2026-09-01
fix(devices): detach ai_agent_runs.ticket_id on device org-move (#4215)

### LINKED ISSUE #4215: moveOrg: detach ai_agent_runs.ticket_id on device org-move (ticket runs are device-less)
Follow-up to #3828 (PR #4203, pre-existing from PR #4195). Tickets are re-stamped to the destination org on device move, but ticket runs have no `device_id`, so the existing `UPDATE ai_agent_runs … WHERE device_id = <moved>` detach cannot reach them and a source-org run can point at a now-foreign ticket. Recorded as `KNOWN_UNDETACHED_RUN_LINEAGE_COLUMNS` in `moveOrg.coverage.test.ts`. Fix: `WHERE ticket_id IN (SELECT id FROM tickets WHERE device_id = <moved>)` in both `moveOrg.ts` and `breeze_cascade_device_org_id()`, then remove the allowlist entry.

ToddHebebrand:
Closes #4215.

## The gap

On a cross-org device move, `ai_agent_runs` history deliberately stays with the **source** org (owner decision 2026-08-23), so every FK pointing at a row that moves *with* the device has to be severed instead. Four were: `device_id`, `alert_id`, `session_id`, `anomaly_incident_id`.

`ticket_id` is the fifth, and it is **unreachable from `WHERE device_id = <moved>`** — a ticket-triggered run carries `ticket_id` with a **NULL `device_id`** (`trigger_kind = 'ticket'`). Meanwhile `tickets` *is* in `getDeviceOrgDenormalizedTables()`, so a ticket bound to the moved device is re-stamped to the destination org by the generic loop, and the retained source-org run is left naming a now-foreign ticket.

Exactly the cross-tenant-pointer class #3828 fixed for `anomaly_incident_id`; it has been carried since then as `KNOWN_UNDETACHED_RUN_LINEAGE_COLUMNS` in `moveOrg.coverage.test.ts`. That allowlist is now gone.

## The fix

A second, **ticket-keyed** statement at both detach sites, using the same `ticket_id IN (SELECT id FROM tickets WHERE device_id = ...)` join the `ticket_attachments` / `time_entries` / `ticket_parts` org rewrites already use. Keying off the *ticket's* `device_id` reaches both the device-less ticket runs and device runs on the same ticket, and touches nothing whose ticket stays behind in the source org.

| Site | Change |
|---|---|
| `apps/api/src/routes/devices/moveOrg.ts` | New UPDATE, placed **before** the denormalized-table loop so both sides are still read under the source org (same ordering rationale as the `metric_anomaly_incidents` reverse pointer). |
| `apps/api/migrations/2026-09-29-detach-ticket-runs-on-device-org-move.sql` | `CREATE OR REPLACE FUNCTION public.breeze_cascade_device_org_id()` — the DB-side path for direct-SQL/non-route callers such as `orgMerge`. Body copied verbatim from `2026-09-20-ai-agents-anomaly-pilot.sql` (the newest definition) with only the new statement added. Fix-forward; no shipped migration edited. |

**No historical backfill, deliberately.** `ai_agent_runs` is `FORCE ROW LEVEL SECURITY`, and the repo has a *Check Migrations (non-superuser)* CI job — a bare in-migration `UPDATE ... WHERE r.org_id IS DISTINCT FROM t.org_id` would be subject to policy with no `breeze_*` access context set, match zero rows, and report a misleading "0 cleaned". If a sweep is wanted it belongs in a scripted backfill under a known access context. Rationale is recorded in the migration itself.

## Test contract

The `ai_agent_runs run-lineage detach coverage` block keeps the two sites in sync by deriving the expected column set from the schema's **own FK columns** (not a hand list). Two changes were needed because the columns no longer live in a single statement:

- It now **unions the SET clauses of every `UPDATE ai_agent_runs`** at each site, rather than matching one statement.
- It **resolves the newest migration defining `breeze_cascade_device_org_id()` dynamically** (filename `localeCompare` order, the same way `autoMigrate` applies files) instead of by hardcoded filename — so a later redefinition can't leave the contract silently asserting a superseded body. That was a live footgun: the old test pointed at `2026-09-20-...` by name.

Plus targeted assertions at both sites that `ticket_id` is severed *through the tickets join*, not by a blanket `WHERE device_id`.

## Verification

Every claim below was run, not inferred.

**Red first.** With the tests written and no implementation, `moveOrg.coverage.test.ts` + `moveOrg.test.ts` were **5 failed / 42 passed**; after the fix, **47 passed**.

**Real Postgres, DB-side function** (throwaway container, minimal fixture, trigger fired via `UPDATE devices SET org_id`):

| Run | device-less ticket run | device run on same ticket | run whose ticket stays |
|---|---|---|---|
| pre-fix function | `ticket_id` **kept** | `ticket_id` **kept** | kept (correct) |
| this PR's function | `ticket_id` **NULL** | `ticket_id` **NULL** | **kept** (correct) |

Assertion `count(*) FROM ai_agent_runs r JOIN tickets t ON t.id = r.ticket_id WHERE r.org_id IS DISTINCT FROM t.org_id` → **2 pre-fix, 0 post-fix**.

**Integration test** (`agentRunMoveSemantics.integration.test.ts`, real DB, `breeze_app` NOSUPERUSER/NOBYPASSRLS so RLS is genuinely exercised) — drives `POST /devices/:id/move-org` end to end:

- New case green (5/5 in the file).
- Control: pre-fix trigger **and** route statement removed → the new case fails on exactly the `ticket_id` assertion, other 4 pass. The test discriminates.
- Second control: route statement restored, trigger still pre-fix → green, confirming the route-side statement is load-bearing on its own for the route path (the migration covers the non-route path).

**Migration applied in-set:** `pnpm db:migrate` against a fresh DB applied it **last** (`[auto-migrate] Applying: 2026-09-29-detach-ticket-runs-on-device-org-move.sql`), and re-applying the file is a no-op (`CREATE OR REPLACE`). `pnpm db:check-drift` → *"No drift detected — all 618 migration files match the breeze_migrations ledger."*

**Other:** `tsc --noEmit -p apps/api` clean; `eslint` on all four TS files clean; `autoMigrate.test.ts` 63 passed.

### DIFF
--- a/apps/api/migrations/2026-09-29-detach-ticket-runs-on-device-org-move.sql
+++ b/apps/api/migrations/2026-09-29-detach-ticket-runs-on-device-org-move.sql
@@ -1,0 +1,80 @@
+-- #4215 — detach ai_agent_runs.ticket_id on a device org-move.
+--
+-- breeze_cascade_device_org_id() (last replaced in
+-- 2026-09-20-ai-agents-anomaly-pilot.sql) severs the moved device's
+-- ai_agent_runs lineage pointers with a single `WHERE device_id = NEW.id`
+-- statement. That WHERE can never reach `ticket_id`: ticket-triggered runs are
+-- device-less (trigger_kind 'ticket' stamps ticket_id and leaves device_id
+-- NULL), yet `tickets` IS returned by breeze_device_child_orgid_tables(), so a
+-- ticket bound to the moved device is re-stamped to the destination org by the
+-- generic loop below while the run — whose org_id deliberately stays with the
+-- SOURCE org (owner decision 2026-08-23) — keeps pointing at a now-foreign
+-- ticket. Same cross-tenant-pointer class that #3828 fixed for
+-- anomaly_incident_id, recorded since then as a known gap in
+-- moveOrg.coverage.test.ts.
+--
+-- Fix: a second, ticket-keyed detach using the same
+-- `ticket_id IN (SELECT id FROM tickets WHERE device_id = ...)` join the
+-- route's ticket_attachments / time_entries / ticket_parts org rewrites
+-- already use. It reaches both the device-less ticket runs and device runs on
+-- the same ticket, and touches nothing whose ticket stays behind in the source
+-- org. Placed BEFORE the generic loop so both sides of the statement are still
+-- read pre-restamp, matching the metric_anomaly_incidents reverse pointer.
+--
+-- Mirrors the identical pair of statements in
+-- apps/api/src/routes/devices/moveOrg.ts (the route path); this function is
+-- the DB-side path for direct-SQL/non-route callers such as orgMerge. The two
+-- sites are held in sync by moveOrg.coverage.test.ts's "ai_agent_runs
+-- run-lineage detach coverage" block, which derives the expected column set
+-- from the ai_agent_runs schema's own FK columns and resolves the newest
+-- migration defining this function dynamically.
+--
+-- Full function body copied verbatim from 2026-09-20-ai-agents-anomaly-pilot
+-- (the newest definition; no later migration replaces this function) with only
+-- the ticket-keyed UPDATE added. CREATE OR REPLACE is idempotent by
+-- construction — re-applying this file re-installs the same body.
+CREATE OR REPLACE FUNCTION public.breeze_cascade_device_org_id()
+  RETURNS trigger
+  LANGUAGE plpgsql
+  SECURITY DEFINER
+  SET search_path = public, pg_catalog
+  AS $$
+DECLARE
+  child_table text;
+BEGIN
+  -- Agent-run history stays with the SOURCE org (owner decision 2026-08-23):
+  -- sever the moved device's lineage links instead of re-stamping org_id.
+  UPDATE public.ai_agent_runs
+    SET device_id = NULL, alert_id = NULL, session_id = NULL, anomaly_incident_id = NULL
+    WHERE device_id = NEW.id;
+  -- ticket_id is device-lineage too, but unreachable from `WHERE device_id`:
+  -- ticket-triggered runs carry a ticket_id with a NULL device_id. Key off the
+  -- ticket's device_id instead (#4215).
+  UPDATE public.ai_agent_runs
+    SET ticket_id = NULL
+    WHERE ticket_id IN (SELECT id FROM public.tickets WHERE device_id = NEW.id);
+  -- Reverse pointer: the incident's back-link to the (now-detached) run must
+  -- not keep naming a source-org run once the incident itself is re-stamped
+  -- to the destination org by the generic loop below.
+  UPDATE public.metric_anomaly_incidents
+    SET agent_run_id = NULL
+    WHERE device_id = NEW.id;
+  FOR child_table IN SELECT public.breeze_device_child_orgid_tables() LOOP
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer (both Opus, read-only). Completed **after** this PR was merged, so the findings could not be folded in here.

**Findings:** 0 blockers. 2 important + 5 nits, all addressed in **#4515** (tests and comments only; no behaviour change to the fix that shipped here).

The two important ones, both about the contract rather than the fix:

- **The coverage contract lost its reachability binding.** Unioning SET clauses across the two detach statements pins *which* columns are severed but not *which predicate* severs them — folding `ticket_id = NULL` back into the device-keyed statement re-introduces #4215 exactly (it nulls nothing on a device-less run) while still passing set-equality. #4515 splits each statement into SET/WHERE and asserts every derived column is severed exactly once under a predicate that can reach it. Mutation-verified: that regression now fails, where before it passed.
- **Nothing attributed the effect to the migration.** `breeze_cascade_device_org_id()` is an `AFTER ... FOR EACH ROW` trigger on the `devices` UPDATE, so it fires first and this route's own statements then match nothing — the route test would have passed with the migration reverted. #4515 adds a direct `UPDATE devices SET org_id` case (no route) that isolates the trigger; red against the pre-fix function, green against the shipped one.

Nits, also in #4515: a factually wrong ordering rationale in `moveOrg.ts` / `moveOrg.test.ts` (the trigger has already restamped `tickets.org_id` by the time the route's copy runs); two weak discriminators in the integration fixture; and the migration resolver reading all 618 `.sql` files three times per run.

**Verified clean by the code reviewer:** function-body fidelity (diffed against `2026-09-20-ai-agents-anomaly-pilot.sql` — identical except the 6 intended added lines, no dropped statement), migration rules (sort order, idempotency, no inner `BEGIN`/`COMMIT`, no shipped migration edited), `SECURITY DEFINER` / `search_path`, SQL predicate NULL semantics and index coverage, RLS reachability of the tickets subselect at both sites, and the derived column set.

**Tests at merge time:** `moveOrg.coverage.test.ts` + `moveOrg.test.ts` 47 passed (5 failed red-first before the fix); `autoMigrate.test.ts` 63 passed; `agentRunMoveSemantics.integration.test.ts` green on real Postgres under `breeze_app` NOSUPERUSER/NOBYPASSRLS; `rls-coverage` 80 passed; `orgMerge` 10 passed; `db:check-drift` → *"No drift detected — all 618 migration files match the breeze_migrations ledger"*, with this migration applying **last**. `tsc --noEmit -p apps/api` and `eslint` clean.

**One thing worth its own issue:** `moveTicketOrg` (`services/ticketService.ts`) moves a **ticket** between orgs and does **not** sever `ai_agent_runs.ticket_id` — same cross-tenant-pointer class on the ticket axis. Unlike `ticket_drafts` / `action_intents` there, it will not 23503 to announce itself, because `ai_agent_runs.ticket_id` is a plain single-column FK. This PR deleted `KNOWN_UNDETACHED_RUN_LINEAGE_COLUMNS`, the only written record of that hazard; #4515 restores a home for it in the coverage block header, but the gap itself is still open.

**Status:** merged. Follow-up #4515 open and awaiting review.

PR #4439, opened 2026-09-01
fix(web): surface MFA enrollment failures and stop 'View codes' regenerating recovery codes (#4413, #4414)

### LINKED ISSUE #4413: [UI][MFA] Wrong TOTP code during enrollment fails silently — panel collapses, secret discarded, no error shown
**Verified.** `POST /auth/mfa/enable` with a wrong TOTP returns 401 `Invalid MFA code`, but the UI shows nothing: a MutationObserver over document.body recorded only the enrollment panel collapsing back to 'Enable' (sampled 300ms–2.5s, no toast, no inline error). The QR/secret are discarded and re-enrolling issues a NEW secret — so a user who mistyped once has a dead authenticator entry and no idea why. The login challenge renders the identical 401 correctly, so this is one UI branch, not the API.

**Suspected:** the mfa/enable failure branch in `apps/web/src/components/settings/MFASettings.tsx` bypasses `runAction` (the no-silent-mutations contract).

**Severity:** release-blocking for v0.109.0 — #4237 merged today with zero human eyes on this path.

Reported By: internal Playwright UI QA sweep 2026-09-01, evidence in docs/testing/FEATURE_TEST_LOG.md (pre-0.109.0 targeted sweep).

### LINKED ISSUE #4414: [UI][MFA] 'View codes' silently REGENERATES recovery codes — reading your codes invalidates the ones you saved
**Verified via DB.** The 'View codes' control fires `POST /auth/mfa/recovery-codes`, which returns 10 brand-new codes: stored count went 10 → 9 (one consumed at login) → 10 again, sharing nothing with the enrollment set. The 'this invalidates previous codes' warning renders only AFTER the mutation, and a separate 'Regenerate codes' button exists in the same panel — so the obvious read action is destructive.

**Fix shape:** codes are presumably stored hashed and cannot be re-shown — so 'View codes' should not exist as a separate control; fold into a single 'Regenerate codes' action gated behind an explicit confirm that states the invalidation BEFORE mutating.

Reported By: internal Playwright UI QA sweep 2026-09-01, evidence in docs/testing/FEATURE_TEST_LOG.md (pre-0.109.0 targeted sweep).

### LINKED ISSUE #4471: [UI][MFA] 'Copy recovery codes' has no success feedback and no clipboard failure handling — codes are shown exactly once
**Verified.** `handleCopyRecoveryCodes` (`apps/web/src/components/settings/MFASettings.tsx:355`) fires `navigator.clipboard.writeText` with no `.catch()` and no visible "Copied" confirmation. Pre-existing, but after #4439 recovery codes are displayed exactly once (enrollment, or a confirmed regenerate) — a clipboard denial (non-secure context, permissions policy, Safari gesture rules) silently leaves the user with nothing copied and no way back to the codes.

**Fix shape:** await the write; on success show a transient "Copied" state on the button; on failure show an inline error telling the user to copy/download manually while the codes are still on screen. A download-as-text fallback would remove the clipboard dependency entirely.

Surfaced by the #4413/#4414 fix (PR #4439); reported by internal UI QA follow-up 2026-09-01.

ToddHebebrand:
Fixes #4413
Fixes #4414

Both bugs live in the same component pair (`MFASettings.tsx` + its parent `ProfilePage.tsx`), so they land together.

## #4413 — wrong TOTP during enrollment failed silently (release-blocking)

Two independent causes, both verified by reading the code and reproduced in jsdom:

1. **The panel collapsed on every outcome.** `MFASettings.handleEnableSubmit` called `setView('status')` unconditionally after `await onEnable?.(...)`, and `ProfilePage.handleMfaEnable` swallowed the rejection (it only set `mfaError` and returned `undefined`) — so the awaited promise resolved exactly like a success. The QR/secret was left behind on a view the user could not get back to, along with the password collected at the gate, so a single mistyped digit cost a full re-enrollment against a **new** secret and stranded the authenticator entry the user had just scanned.

2. **A rejected proof was mistaken for an expired session.** `POST /auth/mfa/enable` answers **401** for "that code is wrong" — the same status the bearer guard uses. `fetchWithAuth` (`stores/auth.ts:1132`) hands *any* 401 to refresh-and-replay: on success it replays the same single-use code, and when the refresh does not restore it calls `handleSessionExpired` → `logout()` + `window.location.replace('/login…')`, mid-enrollment. That is the path on which nothing at all renders, which matches the QA sweep's MutationObserver evidence.

**Fix:** `onEnable` / `onDisable` / `onGenerateRecoveryCodes` now report a verdict — `false` means rejected, `undefined` still means success so the prop stays back-compatible (same convention as the existing `onRequestSetup`). On rejection the component **stays on the QR view**, keeps `currentPassword`, clears only the digits, focuses the first input, and renders an explicit "your QR code is still valid, wait for the next code" hint next to the server's message. All three MFA proof endpoints now pass `skipUnauthorizedRetry: true`, so a rejected proof can never be replayed or read as a dead session.

The same one-line flag is applied to `apiEnableTotpMfa` in `stores/auth.ts`, which the **forced-enrollment** page uses — an eviction there strands the user with no way back in. That path already surfaced its error through a typed result, so only the misclassification needed fixing.

> **Follow-up worth filing (not attempted here):** the durable fix is API-side — return 400/422 with a stable error code for a rejected factor proof, and reserve 401 for invalid bearer auth. Then the generic refresh path can stay on for these endpoints.

## #4414 — 'View codes' silently regenerated

`POST /auth/mfa/recovery-codes` regenerates: it returns ten new codes and invalidates every code the user had saved. It was wired to a **"View codes"** button — the plainest possible read affordance — with the invalidation warning rendered only *after* the write, and a second "Regenerate" button in the same panel. On top of that, a *failed* regeneration still ran `setShowCodes(true)`, re-displaying the **previous** set as though it were new.

**Fix:**

- The read affordance is gone. The status card now offers **"Regenerate recovery codes"** and says plainly that codes are shown once and cannot be displayed again.
- The recovery panel has exactly one action, behind a `ConfirmDialog` (`variant="destructive"`, `confirmTestId="confirm-regenerate-recovery-codes"`) that states the invalidation **before** anything is sent. Nothing is POSTed until confirm is accepted.
- Codes are revealed only on a confirmed success.
- **Necessary complement:** enrollment now shows the codes `/mfa/enable` already returns. They were previously fetched into state and never displayed — which is precisely why regenerating was the only way to ever see any. Removing the read path without this would have made the first set unreachable forever.

Dead catalog keys (`viewCodes`, `showRecoveryCodes`, `clickTheButtonBelowToViewOrGenerateNewRecoveryCodes`) are removed from all 8 locales; 6 new `mFASettings.*` keys added to all 8 with per-language translations (non-English are machine-drafted pending native review). Locale parity + translation-coverage suites pass.

## Tests (red first)

New: `MFASettings.recoveryCodes.test.tsx` (5), `ProfilePage.mfaEnrollment.test.tsx` (5), plus one case in `stores/auth.test.ts`. All were written against the old code and watched fail — 9 of 10 red on the first run, the one green case being the deliberate success control ("still completes enrollment on a correct code"), and the auth-store case failing on `expected 1 call, got 2` (the refresh).

They pin: the 401 is visible; the enrollment panel survives it; a retry reuses the same secret and still carries the password (no second `/auth/mfa/setup`); `skipUnauthorizedRetry` is set; enrollment displays its recovery codes without a second request; no POST before confirm; dismissing confirm mutates nothing; a failed regeneration never shows stale codes.

**Verification**

- `vitest run` (full `apps/web` suite): **688 files / 7094 tests passed**, including `src/lib/i18n/` (locale parity, translation coverage, key usage) and `no-silent-mutations`.
- `astro check`: 1754 files, **0 errors**.
- `eslint` on all touched files: clean.

**Note on `runAction`:** `ProfilePage.tsx` is not in the `no-silent-mutations` target set, and its MFA handlers use the sanctioned alternative — explicit inline error UI, now actually reachable. Migrating the whole page to `runAction` (which would also need `treatUnauthorizedAsError` for these 401s) is a larger change than these two bugs justify.

Design was cross-checked with an independent Codex read-only review, which agreed on all four decisions (collapse cause, `skipUnauthorizedRetry`, boolean-verdict over rethrow, and which state to preserve on failure).

Fixes #4471 (folded in: copy-recovery-codes now confirms or fails inline — 3 red-first tests, 8 locales).


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three lenses in parallel: `code-reviewer` (verdict contract, view state machine, `skipUnauthorizedRetry` blast radius), `silent-failure-hunter` (does the diff introduce or leave a silent path), `pr-test-analyzer` (vacuity + coverage).

**Findings:** 4 acted on in `b02fada8a`; 0 outstanding blockers.

- **Real bug, fixed** — `displayCodes` is `smsRecoveryCodes || recoveryCodes`, and a regeneration only refreshes the `recoveryCodes` prop. A user who enrolled via SMS and then regenerated kept looking at the SMS set the regeneration had just invalidated: the exact failure #4414 exists to prevent, reached through the control this PR introduces. `handleRegenerateCodes` now clears `smsRecoveryCodes` on success.
- **Coverage gaps, closed** — a rejected `/mfa/disable` now has a test proving the panel stays open (it used to collapse to a status card still reading "Enabled"); `skipUnauthorizedRetry` is asserted on `/mfa/disable` and `/mfa/recovery-codes`, not only `/mfa/enable`; the enrollment retry hint is pinned appearing on rejection and clearing when the user cancels out and re-enters setup.
- **Confirmed clean** — no swallowed errors on any touched handler (every catch traced to a rendered node); `ProfilePage` is outside the `no-silent-mutations` target set both before and after, and the handlers meet the contract's intent (every outcome renders on whichever view the user is left on); locale key sets are identical across all 8 locales.
- **Independently corroborated** — the reviewer's read of the API confirmed `/auth/mfa/recovery-codes` returns `{error}` with **no** `message` field on the wrong-password path, so this PR's `errorData.error ?? errorData.message` change fixes a second, previously unreported bug: that failure used to fall through to generic copy instead of "Invalid credentials".

**Not fixed here (pre-existing, out of scope):** `handleCopyRecoveryCodes` (`MFASettings.tsx:355`) calls `navigator.clipboard.writeText` with no `.catch()` and no "Copied!" state, so a denied-permission rejection is silent. Untouched by this diff. Worth its own issue — it matters more now that codes are shown exactly once.

**Follow-up worth filing:** the durable fix for #4413 is API-side — `/auth/mfa/{enable,disable}` and the step-up helper should answer **400/422** with a stable code for a rejected proof and reserve 401 for invalid bearer auth. `skipUnauthorizedRetry` is the correct client-side stopgap, not the end state.

**Tests:** red-first, and each new case verified to *discriminate* — reverting its own fix turns it red (checked by mutating the implementation and re-running, 3/3 flipped). `apps/web` settings + i18n + auth + no-silent-mutations suites: **102 files / 1296 tests passed**; full suite was green at the first commit (688 files / 7094 tests). `astro check`: 1754 files, **0 errors**. `eslint` on all touched files: clean.

**Status:** review-clean, awaiting maintainer merge. Not merged, and #4413/#4414 left open per hand-off rules.

Issue #3772, opened 2026-08-21
Multi-currency (billing, quotes, contracts, ticketing, catalog, reporting)


ToddHebebrand:
Tracking issue for **Multi-currency (billing, quotes, contracts, ticketing, catalog, reporting)** (`multi-currency`).

- Plan: `docs/superpowers/specs/billing/2026-08-21-multi-currency-design.md`
- Spec: `docs/superpowers/specs/billing/2026-08-21-multi-currency-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All 8 waves (#3773–#3780) merged by 2026-08-23 (PRs #3794, #3804, #3811, #3810, #3809, #3874, #3883, #3882 plus review follow-ups #3832, #3833). Shipped in v0.107.0+. Closing the parent; the final close_feature step was never run after the last wave landed.

PR #4440, opened 2026-09-01
fix(web): make the FX approximate line report failure instead of vanishing (#4415)

### LINKED ISSUE #4415: [Web][FX] ApproximateMoneyLine renders null for loading, failed AND unavailable — rich API failure detail discarded on all seven surfaces
**Verified.** The reporting API returns an explicit failure shape (`status:"unavailable"`, `unavailableCurrencyCodes:["NGN"]`, `reason:"missing"`), but `apps/web/src/components/billing/ApproximateMoneyLine.tsx:24-26` returns `null` for loading, error, and unavailable alike. Seven surfaces share the component, so self-hosters without exchange-rate provider access never see the FX line or learn why.

This is the fourth instance of the exact hide-on-failure defect class this component's surface produced during the multi-currency program (see [[quiet_failure_ui_needs_e2e_proof]] / #3772 W07 history — three prior bugs each made this line render nothing while unit-green).

**Fix shape:** render an explicit muted 'FX unavailable (\<codes\>)' state for `unavailable`/error; keep null only for loading. Follow-up of #3772.

Reported By: internal Playwright UI QA sweep 2026-09-01, evidence in docs/testing/FEATURE_TEST_LOG.md (pre-0.109.0 targeted sweep).

ToddHebebrand:
Fixes #4415. Follow-up of #3772 (multi-currency W07).

## The bug

`ApproximateMoneyLine` returned `null` for loading, a failed request **and** the server's `status: "unavailable"` alike:

```tsx
if (loading || failed) return null;
const view = selectApproxTotal(response);
if (view.status !== 'available') return null;
```

The endpoint answers with real detail (`status:"unavailable"`, `unavailableCurrencyCodes:["NGN"]`, per-group `reason:"missing"|"stale"`) and all of it was discarded. Seven surfaces share the component — invoices, quotes, contracts, timesheets, partner dashboard, ticket labor, ticket parts — so a self-hoster with no exchange-rate coverage saw a line that simply never rendered and had no way to learn why.

This is the **fourth** hide-on-failure defect on this exact surface during the multi-currency program, so the fix removes the mechanism, not the instance.

## The fix

`ApproxTotalView` is now **exhaustive** over everything the endpoint can answer, with exactly ONE silent member (`hidden`) reserved for the two states that carry no information: no request was made, and `not-needed` (the book is already entirely in the reporting currency). `unavailable` is its own member carrying the codes and the reason — re-hiding a failure now requires deleting a union member and reddening the component.

| State | Renders |
|---|---|
| loading | nothing — a flash of "unavailable" would be its own lie |
| `not-needed` / empty book | nothing |
| `unavailable` | `≈ total unavailable — no CAD exchange rate for NGN` |
| request failed | `≈ total unavailable — could not load exchange rates` |

Same `text-xs text-muted-foreground` styling as the total it replaces and the same `data-testid`, plus a `data-approx-state` attribute so a page suite cannot pass on the wrong state. **No figure of any kind appears in a failure state**, so the existing no-silent-1:1 guard is preserved (the suite now asserts `not.toMatch(/\d/)` on the line, which is stricter than the old blanket "no ≈ anywhere").

## Three accuracy points

These came out of an independent read of the design (Codex, read-only) and are worth calling out because two of them are things the naive fix gets wrong:

- **The copy names the PAIR, not the leg.** `unavailableCurrencyCodes` flags the *source* group that could not convert. When the **target** leg fails, `reportingTotals.ts` makes every group inherit that reason — so "NGN's rate is missing" would assert a cause we have not established. "no CAD exchange rate for NGN" is true either way.
- **The code list is capped at four and the remainder is COUNTED** (`… and 2 more`). A many-currency partner must not turn a one-line companion into a paragraph on seven surfaces, and a CSS truncation would hide the overflow silently — the same defect class again.
- **`unavailableReason` accepts only `missing`/`stale`.** The body is unvalidated JSON at that point; a drifted server cannot make the copy assert a cause the contract has no word for. Disagreement stays `mixed` rather than promoting the first leg's reason.

## One adjacent hole closed

`buildGroupsParam` returned `''` for both an empty book **and** an unusable one (a credit note pushing a currency negative, an unroundable currency code) — the identical defect one layer down: the line vanished with no request and no trace. It returns `{ kind: 'empty' | 'invalid' | 'query' }` now, and `invalid` surfaces as the visible failure state.

## Deliberately NOT done

409 `NO_REPORTING_CURRENCY` reads as the generic failure rather than getting its own message. `partners.currency_code` is `NOT NULL DEFAULT 'USD'`, so the branch is effectively unreachable for a real partner- or org-scoped viewer; threading an error code through `loadApproximateTotal` and its cache would buy a sentence nobody sees. Noted in the hook's contract comment so the next reader does not re-litigate it.

## Strings

Five new keys in `money.*`, landed in **all 8 locale catalogs** (en, de-DE, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR). The keys are referenced as **literals**, not through a lookup table, so `keyUsage.test.ts` can still prove they exist in the catalogs. The seven-surface contract suite resolves them through the *shipped* `en` file rather than a mocked template, so the wiring is proven end to end.

## Verification

- `apps/web` full suite: **686 files / 7096 tests passed** (`vitest run --pool=threads --maxWorkers=2`)
- `tsc --noEmit` on `apps/web`: clean
- `eslint` on all touched files: clean
- `apps/api/src/services/exchangeRateBoundary.test.ts` (FX reporting-only architectural guard): 7 passed — no new file references an FX module
- Locale parity suite: 79 passed

Red-first: the new assertions were written and watched fail (16 failures, all "element not found") before the implementation landed.

Three existing suites asserted the *old* hide-on-failure contract and were inverted: `ApproximateMoneyLine.test.tsx`, the seven-surface `moneyRollupApproximate.test.tsx` (states 3/4/5), and a sibling copy in `PartnerDashboard.test.tsx` that a `src/components/billing/` path glob would have missed.

### DIFF
--- a/apps/web/src/components/billing/shared/ApproximateMoneyLine.test.tsx
+++ b/apps/web/src/components/billing/shared/ApproximateMoneyLine.test.tsx
@@ -167,6 +167,27 @@ describe('ApproximateMoneyLine', () => {
     expect(fetchWithAuth).not.toHaveBeenCalled();
   });
 
+  // Both of these bodies PASS `useApproximateTotal.validate` — an empty array is
+  // still an array, a whitespace code is still a non-empty string — so they
+  // reach the component for real. The line must fall back to what it can prove
+  // rather than rendering "no  exchange rate for " (or nothing at all).
+  it('falls back to the generic failure when the server names NO codes', async () => {
+    fetchWithAuth.mockResolvedValue(jsonRes(unavailableBody([], [])));
+    render(<ApproximateMoneyLine byCurrency={BOOK} date="2026-08-21" />);
+    const line = await screen.findByTestId('approximate-money-line');
+    expect(line.textContent).toBe('≈ total unavailable — could not load exchange rates');
+    expect(line.dataset.approxState).toBe('unavailable');
+  });
+
+  it('falls back to the generic failure when the target currency is unusable', async () => {
+    const body = unavailableBody([{ currencyCode: 'NGN', reason: 'missing' }], ['NGN']);
+    fetchWithAuth.mockResolvedValue(jsonRes({ data: { ...body.data, targetCurrencyCode: '   ' } }));
+    render(<ApproximateMoneyLine byCurrency={BOOK} date="2026-08-21" />);
+    const line = await screen.findByTestId('approximate-money-line');
+    // Never "no  exchange rate for NGN": half a pair is not a fact.
+    expect(line.textContent).toBe('≈ total unavailable — could not load exchange rates');
+  });
+
   it('still shows NO figure in the unavailable state — no partial total, no 1:1', async () => {
     fetchWithAuth.mockResolvedValue(jsonRes(unavailableBody([{ currencyCode: 'EUR', reason: 'missing' }], ['EUR'])));
     const { container } = render(<ApproximateMoneyLine byCurrency={BOOK} date="2026-08-21" />);
--- a/apps/web/src/components/billing/shared/ApproximateMoneyLine.test.tsx
+++ b/apps/web/src/components/billing/shared/ApproximateMoneyLine.test.tsx
@@ -5,11 +5,17 @@ const fetchWithAuth = vi.fn();
 vi.mock('@/stores/auth', () => ({ fetchWithAuth: (...a: unknown[]) => fetchWithAuth(...a) }));
 vi.mock('../../../stores/auth', () => ({ fetchWithAuth: (...a: unknown[]) => fetchWithAuth(...a) }));
 
-// The en catalog entry lands with the rest of the locales in the wiring task;
-// pinning the template HERE proves the component passes exactly the
-// interpolations that entry consumes ({{amount}} and {{rateDate}}).
+// Pinning the templates HERE proves the component passes exactly the
+// interpolations each catalog entry consumes. The catalogs themselves are
+// exercised for real by `moneyRollupApproximate.test.tsx`, which renders the
+// seven surfaces against the shipped `en` file, and by the locale parity suite.
 const TEMPLATES: Record<string, string> = {
   'money.approximateTotal': '≈ {{amount}} approximate · rates as of {{rateDate}}',
+  'money.approximateUnavailableMissing': '≈ total unavailable — no {{target}} exchange rate for {{codes}}',
+  'money.approximateUnavailableStale': '≈ total unavailable — {{target}} exchange rate too old for {{codes}}',
+  'money.approximateUnavailableMixed': '≈ total unavailable — {{codes}} could not be converted to {{target}}',
+  'money.approximateUnavailableCodesOverflow': '{{codes}} and {{more}} more',
+  'money.approximateFailed': '≈ total unavailable — could not load exchange rates',
 };
 vi.mock('react-i18next', () => ({
   useTranslation: () => ({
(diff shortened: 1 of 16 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` and `pr-test-analyzer` returned; `silent-failure-hunter` was dispatched but never returned findings (pinged three times). Its scope is not unreviewed: the code-reviewer ran an explicit silent-render audit of every `return null` path, and I enumerated the remaining swallow points by hand (see "Known, not fixed" below). A design question that predated implementation — whether a failed request should surface on all seven surfaces, whether 409 `NO_REPORTING_CURRENCY` deserves its own message, and whether `not-needed` stays silent — also went to an independent read-only Codex pass, which agreed on all three and independently proposed the exhaustive-union shape.

**Findings:** 3 raised → 3 addressed in `2d3dcf77c`; 0 outstanding.

- `code-reviewer` — **0 findings.** Verified rather than read: ran 105 tests across the 5 affected files, parsed all 8 locale JSONs, confirmed `buildGroupsParam` has exactly one call site, and swept all 7 consumer surfaces plus `e2e-tests/` for the `approximate-money-line` testId to prove nothing outside this PR relied on the old "renders nothing on failure" behaviour.
- `pr-test-analyzer` — **0 vacuous assertions**, and confirmed the rewritten state-4 guard is *stronger* than the `not.toContain('≈')` it replaced (exact full-line `.toBe()` match, plus the retained `TARGET_SYMBOL`/`NAIVE_SUM` body checks, plus a digit check on the line). It raised two real gaps, both now closed:
  1. The `unavailable`-with-no-codes / unusable-target fallback had no coverage. It called this unreachable; that is true of the *server* but not of the *client* — `validate()` accepts an empty `unavailableCurrencyCodes` array and a whitespace-only `targetCurrencyCode`, so contract drift reaches the component for real. Two tests now drive it through the real hook.
  2. `useApproximateTotal.test.tsx`'s "does not fetch when every entry is malformed" asserted only `loading === false` and no fetch — it passed identically before and after the fix, locking in nothing, and its title contradicted the new code. It now asserts `failed === true`, with a negative-leg case and a recovery round trip.

**Every assertion added in the follow-up commit was mutation-verified**, not just observed green: reverting the fallback branch to `if (false)` reddens both new component tests, and collapsing `unbuildable` back into `IDLE` reddens all three new hook tests.

**Tests:** `apps/web` full suite green at `eaa2f5e1c` (686 files / 7096 tests). At `2d3dcf77c`, all affected suites green — 172 files / 2059 tests (`vitest run --pool=threads --maxWorkers=2` over `src/lib/`, `src/components/billing/`, `moneyRollupApproximate`, and the five other consuming surfaces). `tsc --noEmit` clean; `eslint` clean on every touched file. `apps/api` FX reporting-only architectural guard (`exchangeRateBoundary.test.ts`) green — no new file references an FX module. Locale parity + i18n key-usage guards green.

**Known, not fixed — pre-existing, deliberately out of scope:** `approximateTotalCacheKey` bumps the cache generation as a side effect, but the generation is not in the hook's effect dependency array (`useApproximateTotal.ts:191`), so a **mounted** approximate line does not refetch after a partner reporting-currency change — it keeps whatever state it already had. That gap predates this PR and is about stale *successful* totals, which is the thing the cache module's own comment says it exists to prevent. This PR makes one corner of it newly visible (a request in flight across the reset now renders "could not load exchange rates" instead of rendering nothing), which is consistent with the PR's thesis but does mean a visible message where there was previously invisible breakage. Reachability is low — the money rollups are normally unmounted while an admin is on the partner billing settings page. Worth its own issue; I did not expand this PR to cover it.

**Status:** review-clean, awaiting maintainer merge. CI was still queued at the time of writing (42 checks, runners backed up) — please confirm green before merging. Issue #4415 stays open and assigned until you close it.

ToddHebebrand:
**silent-failure-hunter verdict (arrived out of band):** the fix closes the defect class — every path in the diff renders a visible, distinct message or is genuinely informationless (loading / hidden). Two pre-existing LOW notes (bare `catch {}` house style; generation-discard edge) are not regressions; the generation/effect-deps gap is now tracked separately. No blockers.

Issue #4201, opened 2026-08-29
[Selfhost] guided-setup reboot-startup service fails to install: Type=oneshot + Restart=on-failure is a bad unit file setting


ToddHebebrand:
Requested by **mafiso9682**; +1 from theangrypenguin69. In #general (Discord, message 1542903196060618873).

**Discussion (PII-redacted):**

**mafiso9682** (2026-08-28T14:10:20.394Z):
> 

**theangrypenguin69** (2026-08-28T14:26:27.696Z):
> OK -- back at it (had a few other projects to focus on...)!
> 
> Ran guided-setup.sh with all of its defaults (except for the bootstrap admin stuff). Set up SSH tunnelling and made it through the initial setup, including changing the bootstrap admin's password.  Once that was complete, I returned to the SSH session and proceeded with the .env cleanup.
> 
> 1st issue:
> 
> After signing in, finish the initial admin setup in Breeze.
> 
> Have you signed in successfully and finished the initial setup? [y/N] y
> 
> ============================================================
> == Reboot Startup ==
> ============================================================
> 
> After a host reboot, Docker may restart individual containers before dependencies are ready.
> Setup can install a small systemd service that gives Breeze a cleaner shutdown and startup path:
>   - on shutdown, it asks Docker Compose to stop the Breeze stack before Docker itself stops
>   - on startup, it reruns Docker Compose after Docker and networking are online
> This helps Breeze stop cleanly and then start Postgres/Redis, API/Web, and optional services in the intended order.
> Install the Breeze reboot startup service now? [Y/n]
> [sudo: authenticate] Password:
> Created symlink '/etc/systemd/system/multi-user.target.wants/breeze-rmm.service' → '/etc/systemd/system/breeze-rmm.service'.
> Failed to start breeze-rmm.service: Unit breeze-rmm.service has a bad unit file setting.
> See system logs and 'systemctl status breeze-rmm.service' for details.
> breeze@breez
> …[truncated]

_Filed automatically by @breeze-discord triage. Verify and add labels/repro as needed._

ToddHebebrand:
[...]

PR #4478, opened 2026-09-01
ci(selfhost): run guided-setup.sh end to end on a runner — systemd unit, packaged Caddy, bootstrap login, SSH-tunnel refresh

ToddHebebrand:
## Why

Nothing executed `scripts/guided-setup.sh` before release, so its defects only ever surfaced on self-hosters' machines: the reboot-startup unit was rejected by systemd twice under #4201 (#4387, then #4473), and the quickstart's own SSH-tunnel recipe produced an endless "session expired" loop after login (#4475). This job runs the real installer on a real Linux host with Docker and systemd — what the runner is, and what a customer's box is.

## What runs

New job **Guided Setup Smoke** (`guided-setup-smoke`), same policy as the existing Smoke Test: non-blocking on PRs, required on `main` via `ci-success`. It builds this checkout's api/web/portal images under the `ci-smoke` tag and runs `scripts/smoke-guided-setup.sh`, which:

1. Stages this PR's `docker-compose.yml`, `.env.example`, `docker/Caddyfile.prod` and the installer in a work dir; seeds only what a self-hoster has to type (version, bootstrap admin).
2. Runs `guided-setup.sh --no-download --yes` end to end with `BREEZE_SETUP_INSTALL_SYSTEMD=true`: generate, pull (local-only tags fail non-fatally, the documented bring-your-own-images path), `up -d`, health wait, and the real `sudo systemctl enable --now breeze-rmm.service`.
3. Asserts the generated `.env` defaults (`BREEZE_DOMAIN=localhost`, `CORS_ALLOWED_ORIGINS=https://localhost`, `FORCE_HTTPS=true`), that the unit is `enabled` + `active` and passes `systemd-analyze verify`, and that `/health` and `/login` are 200 through the packaged Caddy over HTTPS.
4. Signs in as the bootstrap admin at the configured origin and refreshes.
5. Stands up a `socat` forward on 8443 (the `ssh -L 8443:127.0.0.1:443` stand-in), signs in at `https://localhost:8443`, and proves a session survives `POST /auth/refresh` through the tunnel both with `Sec-Fetch-Site: same-origin` and with `Origin == Host` alone. **This is the regression test for the login loop and is expected to fail until #4475 merges** — re-run after that lands.
6. Stops the stack through `systemctl stop`, checks no `breeze-*` container survives, starts it again through the unit, and waits for health.

`scripts/smoke-guided-setup.sh logs` / `teardown` feed the job's failure diagnostics and always-run cleanup; the script is runnable on any Linux box with the three images present.

`guided-setup.sh` gains `BREEZE_SETUP_VERSION`, which preselects the release and skips the GitHub/GHCR lookups (needed to pin the locally built tag; also useful for pinned, air-gapped or self-built installs).

## Verified locally

- Generation half of the flow with the exact CI flags on macOS (`--no-download --no-up --yes`, `BREEZE_SETUP_VERSION=ci-smoke`, GitHub API pointed at a closed port): installer exits 0, `.env` pins `BREEZE_VERSION=ci-smoke`, `docker compose config` resolves `ghcr.io/lanternops/breeze/{api,web,portal}:ci-smoke`.
- `bash -n`, `shellcheck -S warning` clean on both scripts; workflow YAML parses; `actionlint` reports nothing in the new job.
- The systemd/`socat`/`systemctl` half needs the runner — see the job's first run on this PR.

Follows #4473 and #4475.

https://claude.ai/code/session_01PiZHjEw4swqE7HvYP5cEpo


ToddHebebrand:
First run of the new job on this branch (which is `main` without #4473) reproduced #4201 verbatim on a real ubuntu runner — step 5 of `scripts/smoke-guided-setup.sh`:

```
FAIL: breeze-rmm.service is-active: inactive
  Loaded: bad-setting (Reason: Unit breeze-rmm.service has a bad unit file setting.)
  systemd[1]: /etc/systemd/system/breeze-rmm.service:10: WorkingDirectory= path is not absolute: '/home/runner/breeze-guided-smoke'
```

It also surfaced a third defect the customer log already contained: `systemctl enable --now` returned 0 (the enable succeeded), so the installer printed `Installed … and enabled breeze-rmm.service.` and `Guided setup complete.` right after systemd's `Failed to start …` line. Pushed a post-enable `systemctl is-active` check to #4473 so the installer fails loudly instead.

Expected sequence for this job to go green: merge #4473 (unit) and #4475 (tunnel refresh), then re-run.

PR #4505, opened 2026-09-02
chore(mobile): bump iOS buildNumber to 8

ToddHebebrand:
Build 7 already exists in App Store Connect (uploaded 2026-08-20), and `app.json` still declared `"buildNumber": "7"` with `appVersionSource: "local"`.

Apple rejects a duplicate build number within a version. If Xcode Cloud passes the project's `CFBundleVersion` through rather than assigning its own, the next upload is rejected and the build minutes are wasted — so this is cheap insurance: harmless if Xcode Cloud overrides the value, and the difference between a delivered build and a wasted run if it doesn't.

One-line change, no code paths touched.

https://claude.ai/code/session_017eEp7QaCXwJNioNR6vZ6eP

### DIFF
--- a/apps/mobile/app.json
+++ b/apps/mobile/app.json
@@ -9,7 +9,7 @@
     "ios": {
       "supportsTablet": true,
       "bundleIdentifier": "com.breeze.rmm",
-      "buildNumber": "7",
+      "buildNumber": "8",
       "associatedDomains": [
         "webcredentials:us.2breeze.app",
         "webcredentials:eu.2breeze.app"


ToddHebebrand:
Not needed. Confirmed in practice: Xcode Cloud assigns its own build number and does not pass the project's `CFBundleVersion` through, so `app.json`'s `buildNumber` never reaches App Store Connect via the cloud path — the duplicate-7 collision this guarded against cannot happen there.

Closing rather than merging, since the workflow's Files and Folders condition is now scoped to `apps/mobile/**` and merging this would spend a cloud build to change a value nothing reads.

`app.json`'s `buildNumber` still matters for a local Product → Archive, where 7 would collide with the 2026-08-20 upload. Worth bumping alongside other mobile work if anyone returns to that path, rather than on its own.

Issue #4036, opened 2026-08-26
[API/Test] The offboarding repair's FOR UPDATE lock is untested — deleting it leaves #4022's suite green


ToddHebebrand:
`#4034` (merged `a03f27c98`) fixed #4022 by re-reading `status` + `offboarding_started_at` under the repair's row lock and abandoning if the precondition has moved. The guard is correct and covered by six real-Postgres cases in `apps/api/src/__tests__/integration/offboardingRepairRecheck.integration.test.ts`. This issue is about the half of it that is **not** covered.

**The lock is now load-bearing in a way nothing tests.** Before #4034, `.for('update')` in `repairIncompleteEntry` was purely a lock-ordering guard for the #2877 cross-connection cycle. After it, the same statement is also what makes the recheck trustworthy — it is the reason the values read cannot change between the check and the drain prep, queue and stamp that follow.

**Nothing fails if it is deleted.** Every abort in the new suite commits *before* the repair starts, so the recheck alone catches all six cases; removing `.for('update')` leaves the file green. @bdunncompany states this limit explicitly in the PR body rather than leaving it implied.

**Why that is worth closing.** The lock only holds because `withSystemDbAccessContext` → `withDbAccessContext` opens a real transaction (`apps/api/src/db/index.ts:465`) and holds it open for the whole callback to keep the `SET LOCAL` RLS GUCs alive (`:213-217`). The sweep calls the repair inside that callback (`services/tenantOffboarding.ts:934-937` org, `:967-976` partner). So the lifetime of a lock in the offboarding repair is a side effect of how RLS context is established. Any future change to that mechanism — per-statement context, a connection-scoped GUC, a pool-level `SET` — silently reverts #4022 with a fully green suite, and the failure mode is the original one: `self_uninstall` rows queued against a tenant an operator has rescued, collected by the fleet as ordinary commands because `getAgentTenantState` no longer reports `'draining'`.

**Shape of the fix.** Hold an uncommitted `UPDATE ... SET status = 'active'` open in one transaction while the repair runs in another, and assert the repair blocks rather than proceeding. The `pg_stat_activity` barrier at `apps/api/src/__tests__/integration/orgCurrencyCreationBarrier.integration.test.ts:102` is the established pattern.

[...]

Issue #2774, opened 2026-07-25
[API][Agent] Offboarding severs agent credentials before self_uninstall can be delivered — 78 of 80 uninstalls never reached the endpoint


ToddHebebrand:
**Symptom:** Remote agent uninstall does not work on offboarded tenants, and fails silently. Across US production all-time, `self_uninstall` has **8 completed and 80 failed** — and of those 80 failures, **78 were never delivered at all** (`executed_at IS NULL`, expired by the 30-minute server-side timeout). Only 2 ever reached an endpoint and failed there. Effective success rate is roughly 9%.

The operational consequence is that agents remain installed and running on former customers' machines indefinitely, polling forever against a 401. In the current fleet, 26 agents belonging to inactive tenants are generating ~14,500 requests every 3 hours (~116k/day) that can never succeed.

## Root cause — ordering, and two independent gates

Tenant offboarding severs the agent's authenticated channel *before* the uninstall is ever queued, so the command is delivered over a path that has already been destroyed by design.

`severAgentCredentialsForOrgIds` (`tenantLifecycle.ts:44-80`) invalidates the tenant cache, sets `agent_token_suspended_at` on every device in the org, expires the enrollment keys, and severs live WS sockets. From that instant the agent 401s on every authenticated route — including the command poll it would need in order to receive `self_uninstall`.

Sampled evidence from six devices across two churned tenants: every uninstall was created **days after** the tokens were suspended (5 days in one case, 2.5 days in the others), every one has `executed_at IS NULL`, and every one carries the same result — `Command expired: agent never received the command (30 min timeout)`. In each case `last_seen_at` predates the suspension: the agents went dark the moment the tenant was severed.

Worth noting the lockout is enforced **twice**, which matters for any fix:

1. `agent_token_suspended_at` at the device level (`tenantLifecycle.ts:58`)
2. `isAgentTenantActive(orgId)` → `getActiveOrgTenant` → `isUsableOrgStatus`, which accepts **`'active'` or `'trial'` only** (`tenantStatus.ts:13-14`), checked at `agentAuth.ts:477`

Simply deferring the token suspension is therefore not sufficient — flipping org status to `suspended`/`churned` independently 401s the whole fleet via gate 2.

## This was partly known

[...]

PR #4516, opened 2026-09-02
test(api): address review findings on the offboarding lock barrier (#4036)

ToddHebebrand:
Follow-up to #4433 (merged as `0fb5af40d`). That PR landed before this review round finished, so these fixes ship separately. Refs #4036 — leaving the issue open for the maintainer to close.

## Why a second PR

The review on #4433 raised three Major findings. The PR merged while they were still being addressed, so they are applied here against fresh `main` rather than lost.

## Findings addressed

**1. Orphaned sweep promise on every failure path (Major).** The harness never held a reference to the in-flight `sweepOffboardingTenants` promise. If a barrier assertion threw, the `finally` released the holder and returned *while the sweep was still blocked mid-transaction* — a writer touching `enrollment_keys`, `devices`, `device_commands`, `organizations` and Redis, left in flight to run straight into the **next file's** `beforeEach TRUNCATE ... CASCADE`. Integration files run sequentially and the setup retries deadlocks only three times, so an orphan here could red an unrelated suite, in a way that would look like flake in that suite rather than a bug here. The promise is now hoisted and drained with `allSettled` before the client closes.

**2. The barrier could only time out, never diagnose (Major).** The poll is now raced against the sweep itself:

```ts
const blockedQuery = await Promise.race([
  waitForStatementBlockedBy(holderPid, what),
  running.then(() => { throw new Error(`${what}: the sweep ran to completion while another transaction held the tenant row ...`); }),
]);
```

A sweep that runs to **completion** while another transaction holds the tenant row *proves* the repair never queued behind it. That is a fact, established in milliseconds; the old 15s timeout could only guess, and its message named a cause ("the repair read the tenant WITHOUT its row lock") it could not distinguish from a loaded CI box that had not reached the repair yet. The timeout message no longer asserts a cause.

**3. The other allowed outcome was untested (Major).** Blocking is only half the contract. A repair that queued correctly and then abandoned anyway would still be a silent no-op of the #2774 class, and nothing asserted otherwise: the two abort cases expect abandon by construction, and the positive control never contends for the row. Adds a fourth case running the same barrier with a **benign** uncommitted write (a rename — same `FOR NO KEY UPDATE` lock, but nothing the recheck predicate reads has moved), where the repair must wait, re-read, and **proceed**.

**4. Comment accuracy.** #4433 falsified two comments that are still on `main`: the export rationale in `tenantOffboarding.ts` describes the barrier test as hypothetical ("this export is the cost of not doing that here"), and the recheck suite's scope note says pinning the lock "WOULD" require one. Both now point at the suite that exists, and the recheck note keeps its still-true warning — a green run of *that* file is not evidence the lock is covered.

> The `tenantOffboarding.ts` change is **comment-only**. Verified mechanically: every `+`/`-` line in that file's diff is inside a block comment.

Also folded in the smaller review notes: the `orgsFinalized` / `partnersFinalized` assertions are labelled as shape checks rather than discriminators (`failures` is the one that discriminates), the `pg_stat_activity` privilege requirement is documented (a non-superuser poller would see NULL `state`/`query` and time out rather than fail loudly), and the harness's helper duplication + the `repairAudits` name collision with the sibling suite are called out.

## Re-verified by mutation

Both mutations applied to `repairIncompleteEntry` and restored afterwards.

**1. `.for('update')` deleted from both selects — all 3 barrier cases red:**

```
× organization: ... abandons once it commits
× partner: ... abandons once it commits
× organization: a repair that waits on the row and still finds it torn PROCEEDS

AssertionError: blocked on "update "organizations" set "updated_at" = $1,
  "offboarding_started_at" = now() where (...)" — expected the repair's locking SELECT
AssertionError: expected 1 to be +0     <-- one self_uninstall queued against
                                            the rescued tenant. #4022, exactly.
Tests  3 failed | 1 passed (4)
```

The positive control stays green, as it must — it never contends for the row.

**2. `.for('update', { skipLocked: true })`** — a plausible "optimization" that silently turns contention into a zero-row read. All 3 barrier cases red in ~200ms each, on the **new** completion arm:

```
Error: organization repair: the sweep ran to completion while another transaction
held the tenant row — the repair never queued behind it, so it read status/stamp
WITHOUT the row lock and a tenant rescued mid-sweep can still be drained (#4036)
```

**Green with both restored:** `Test Files 2 passed (2) / Tests 10 passed (10)` across this suite and the #4034 recheck suite.

### One claim I walked back

An earlier draft of the header said the benign case would be the *sole* catcher of an over-strict recheck. Both mutations above red it together with the abort cases, so that was not established and the comment now says so explicitly. Its actual justification is narrower and stands on its own: it covers the second of the two allowed outcomes, which nothing else asserts.

## Verification

- `vitest run --config vitest.integration.config.ts` on both offboarding suites — **10 passed**.
- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api` — exit 0, no output.
- `npx eslint` on all three touched files — exit 0.
- Real Postgres via a private per-worktree `pnpm test-stack`, not the shared `:5433`, so the `beforeEach` TRUNCATE could not disturb concurrent sessions.

### DIFF
--- a/apps/api/src/__tests__/integration/offboardingRepairLockBarrier.integration.test.ts
+++ b/apps/api/src/__tests__/integration/offboardingRepairLockBarrier.integration.test.ts
@@ -29,6 +29,17 @@ import { getTestDb } from './setup';
  * mechanism — per-statement context, a connection-scoped GUC, a pool-level
  * `SET` — silently reverts #4022 with a fully green suite.
  *
+ * WHAT THIS FILE DOES AND DOES NOT PIN — read before trusting a green run.
+ * It pins lock ACQUISITION (the repair queues behind the tenant row before it
+ * rechecks) and, via the stamp-time assertion in the benign case, that the
+ * stamp is written by the SAME transaction that was blocked. It does NOT prove
+ * the row is still held throughout drain prep and the queue: a change that took
+ * the lock and dropped it early, keeping the rest in the same transaction,
+ * would still pass everything here. Proving that needs a third client probing
+ * `SELECT ... FOR UPDATE NOWAIT` for 55P03 during the repair's post-barrier
+ * work, which is a harder timing shape and is deliberately not attempted.
+ * Treat the residual gap as known, not as covered.
+ *
  * THE RACE STAGED HERE is the one the committed-abort cases cannot reach:
  *
  *     sweep:    SELECT candidates -> {org, startedAt: null}   (committed read)
@@ -85,9 +96,16 @@ import { getTestDb } from './setup';
  * The layer-1 assertions are `expect.soft` precisely so a future regression
  * reports both halves in one run instead of stopping at the first.
  *
- * FOUR CASES. Two abort barriers (organization and partner — `.for('update')`
- * appears at two separate call sites in `repairIncompleteEntry`, so pinning
- * only one leaves the other uncovered), then:
+ * FIVE CASES. Two abort barriers and two benign barriers, one of each per
+ * scope — `.for('update')` appears at two separate call sites in
+ * `repairIncompleteEntry` and the recheck predicate is written out twice, so
+ * pinning only the organization branch leaves the partner branch uncovered —
+ * then a positive control. Expect three #1105 held-context warnings per run:
+ * every barrier case parks the repair's context for the length of the wait,
+ * well past the 2s `DB_CONTEXT_HELD_WARN_MS` default. They are this file
+ * working, not a leak.
+ *
+ * The roles:
  *
  *  - THE OTHER ALLOWED OUTCOME: the same barrier with a BENIGN uncommitted
  *    write, where the repair waits, re-reads, finds the entry still torn and
@@ -135,6 +153,17 @@ function deferred<T>(): Deferred<T> {
   return { promise, resolve, reject };
 }
 
+/** A rejecting deadline arm for a `Promise.race`; clears its timer when it loses. */
+function lockTimeout(ms: number, message: string): Promise<never> {
+  let timer: NodeJS.Timeout;
+  const promise = new Promise<never>((_, reject) => {
+    timer = setTimeout(() => reject(new Error(message)), ms);
+  });
+  // Unref'd so a lost race cannot hold the event loop open at teardown.
+  timer!.unref?.();
+  return promise;
+}
+
 async function closeRaceClients(...clients: Sql[]): Promise<void> {
   const results = await Promise.allSettled(clients.map((c) => c.end({ timeout: 1 })));
   const failures = results.flatMap((r) => (r.status === 'rejected' ? [r.reason] : []));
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` + `pr-test-analyzer` (2 agents, both on this PR's diff). Round 1 ran on #4433 and its findings are what this PR is; this is round 2, on this PR's own code.

**Findings:** 12 raised (0 Critical, 2 Major, 8 Minor, 4 Nit) → all addressed in `7773bc73e`; 0 outstanding.

The Major worth reading, because it changed what this file claims:

- **The suite did not pin what its header said it did.** The header named "RLS context goes per-statement" as the motivating regression. That change still *acquires* the lock, so every case would have stayed green — the file pinned lock **acquisition**, not lock **lifetime**. Fixed both ways: the header now states the scope plainly and names the residual gap (a lock dropped early with the rest of the work in the same transaction is still invisible; proving that needs a `FOR UPDATE NOWAIT` probe, deliberately not attempted), and the benign cases now assert the stamp predates a `clock_timestamp()` captured while the repair was provably still blocked. The repair stamps with `now()` = `transaction_timestamp()`, so that catches a stamp written by a transaction that *began* after the barrier lifted.
- **Partner scope was covered more shallowly than organization.** The recheck predicate is written out twice, once per branch, so an over-strict recheck in the partner branch was invisible to every org case. Added the partner benign case — which also proves the partner candidate reaches the repair at all, so a partner sweep that stopped selecting candidates can no longer surface as a *lock* accusation.

Minor/Nit, briefly: cancellable block poller (it kept querying for the rest of its deadline after the race settled — the same orphan class round 1 fixed for the sweep); deadline on the holder-lock race (a hang there skipped the `finally` and leaked a connection idle-in-transaction on a row lock); the completion-arm message now reports the sweep's `failures` count so a Redis blip is not misreported as a lock fault; poll deadline 15s → 10s so it cannot lose its diagnostic to the 30s `testTimeout` under load; holder `UPDATE` asserts it matched one row; `tearOrg` pins `offboarding_target = 'churn'` (the enrollment-key probe only discriminates on that path, so a flipped default would have made it vacuously true); audit-absence assertions relabelled as unable to red; "unsatisfiable by unrelated lock waits" corrected; the un-drained fire-and-forget audit insert and the three expected #1105 warnings per run are now documented rather than implied away.

**Mutation evidence** (both applied to `repairIncompleteEntry` and restored; production file verified byte-identical to `main` afterwards):

| Mutation | Result |
|---|---|
| `.for('update')` deleted from both selects | all **4** barrier cases red; positive control green |
| `DB_NOW` `now()` → `clock_timestamp()` — proxy for a stamp written after the barrier lifted | **only the 2 benign cases** red, on the new stamp-time assertion; abort cases + positive control green |

The second one is the isolating evidence round 1 could not produce: it reds exactly the assertion added for it and nothing else.

**Tests:** `vitest run --config vitest.integration.config.ts` on this suite + the #4034 recheck suite — **11 passed**. `tsc --noEmit -p apps/api` exit 0. `eslint` on all three touched files exit 0. Real Postgres via a private per-worktree `test-stack`, not the shared `:5433`.

**One claim walked back:** an earlier draft said the benign case would be the sole catcher of an over-strict recheck. Mutations A and B both red it alongside the abort cases, so that was never established and the comment says so. Mutation C above is what it *does* uniquely catch.

**Note on `tenantOffboarding.ts`:** the change there is **comment-only** — verified mechanically (every `+`/`-` line in that file's diff falls inside a block comment). It corrects two comments that #4433 falsified and that are still wrong on `main`.

**Status:** review-clean, awaiting maintainer merge. Issue #4036 left **open** for you to close.

Issue #4503, opened 2026-09-02
[API] patchCanonicalExportParity: policy normalisation returns {} / 0 where undefined is expected


ToddHebebrand:
## Description
On current `main`, `patchCanonicalExportParity.integration.test.ts` fails in 15 sub-cases of the same shape:

```
AssertionError: rebootDelayMinutes rejects values below one: expected { apps: [], autoApproveDeferralDays: 0 }
to deeply equal { apps: undefined, autoApproveDeferralDays: undefined }
```

The canonical export of a patch policy now fills defaults (`apps: []`, `autoApproveDeferralDays: 0`) where the parity test expects the fields to be absent.

## Root Cause (not yet traced)
Either the canonical exporter started defaulting optional fields, or the schema gained `.default()`s that the parity fixture predates. One of the two contracts needs to win.

## Affected Files
- `apps/api/src/__tests__/integration/patchCanonicalExportParity.integration.test.ts`
- patch policy canonical export / validator

## Reported By
Internal, pre-PR integration sweep on `feat/quickbooks-invoice-push` (PR #4492). Not caused by that branch.

Issue #4500, opened 2026-09-02
[API] action_intents immutability trigger deny-list misses scope_kind / scope_device_id / scope_ticket_id


ToddHebebrand:
## Description
`actionIntentsImmutabilityTrigger.integration.test.ts` fails on current `main` (observed on a branch that does not touch it, full integration sweep 2026-09-01):

```
expected [ 'action_name', …(23) ] to deeply equal [ 'action_name', …(26) ]
- scope_device_id / scope_kind / scope_ticket_id missing from expected deny-list
...
expected 'insert or update on table "action_int…' to match /action_intents content is immutable/
Received: violates foreign key constraint "action_intents_scope_device_id_fkey" / "..._scope_ticket_org_fk"
```

Two overlapping failures in `aiAgentSchedulesPartnerRls.integration.test.ts` have the same origin.

## Root Cause (inferred, not yet traced)
The `scope_*` columns were added to `action_intents` by later AI-agent migrations (2026-09-23..25 range) without extending the immutability trigger's protected-column list, and the test fixture inserts rows that now violate the new FKs before the trigger fires.

## Proposed Fix
New migration extending the trigger's deny-list to the three scope columns; fix the test fixture to satisfy the scope FKs (or set `scope_kind` to a value that needs no FK target).

## Affected Files
- the `action_intents` immutability trigger migration (grep `action_intents content is immutable` under `apps/api/migrations/`)
- `apps/api/src/__tests__/integration/actionIntentsImmutabilityTrigger.integration.test.ts`
- `apps/api/src/__tests__/integration/aiAgentSchedulesPartnerRls.integration.test.ts`

## Reported By
Internal, pre-PR integration sweep on `feat/quickbooks-invoice-push` (PR #4492). Not caused by that branch.

PR #4531, opened 2026-09-02
QuickBooks Phase D: payment pull-back (QBO to Breeze)

ToddHebebrand:
## Summary

QuickBooks Phase D — payment pull-back. QBO-origin payments applied against pushed invoices now mirror into Breeze automatically, via an Intuit webhook (single-region latency optimisation) and a guaranteed 15-minute CDC sweep. This closes out the multi-currency §11 contract in `accountingCurrency.ts` (items 1-4, all delivered) and adds the sandbox checklist / feature docs for the slice.

## What's in it

**API**
- Migration `apps/api/migrations/2026-10-01-quickbooks-payment-pullback.sql`: four new columns on `accounting_connections` (realm fingerprint, `pull_payments`, CDC cursor, webhook verifier state) — partner-axis table, no new tenancy shape, no cascade/export-policy/RLS-allowlist registrations required.
- `accountingProvider`'s CDC `reconcileChanges` (Task 2): widened `ChangeSet`, deletions, split-payment lines, 30-day floor, deterministic window-halving paging (no real `/cdc` cursor exists — decision 3, sandbox item 22 will confirm/replace).
- `accountingPaymentPull.ts` (Task 3): the applier. One transaction, invoice-first lock order:
  1. unlocked discovery read of the invoice mapping (which invoice to lock; not authoritative)
  2. `SELECT ... FROM invoices ... FOR UPDATE`
  3. the **payment** mapping re-read under that lock — the authoritative at-most-once claim
  4. `normalizeAccountingPayment` against the locked invoice's stamped currency (equality asserted before any minor-unit conversion)
  5. `invoice_payments` insert + `recomputeInvoiceStatus` in the same transaction
  Also mirrors QBO-side deletions (`reverseAccountingPayment`, `markInvoiceDeletedRemotely`) and owns Breeze-side mapping cleanup (`clearPaymentMappingForInvoicePayment`).
- `accountingReconcileWorker.ts` (Task 4): the `accounting-reconcile` BullMQ queue, 15-minute sweep, clean-run cursor advance. Registered in `WORKER_REGISTRY` (`EXPECTED_119_NAMES`, up from 118) and `workerEntrypointClosure.contract.test.ts`.
- `POST /api/v1/webhooks/quickbooks` (Task 5): unauthenticated, HMAC-verified against `QBO_WEBHOOK_VERIFIER_TOKEN`, mounted outside the auth chain. Status matrix:
  - `429` rate limiter denies (fails closed, Redis outage included) → Intuit retries
  - `503` verifier token unset → **never** 200, so Intuit keeps retrying instead of an unconfigured region silently swallowing events
  - `401` missing/bad `intuit-signature` → permanent, no retry
  - `400` non-JSON body or no `eventNotifications` → permanent, no retry
  - `503` every queue add failed, or the realm lookup itself threw → retries
  - `202` accepted, including "all realms unknown" or "some dropped past the 50-realm cap"
  Verify → enqueue → 202; the webhook is a doorbell, `reconcileChanges` is the only thing that decides what actually changed.
- `pullPayments` setting + `POST /:provider/reconcile` ("Sync now") + reconcile status on the connection `GET` (Task 6).
- `partner-wide-write-coverage.test.ts` exemption for `accountingPaymentPull.ts` (QBO-signed webhook / system-context CDC writer; no tenant caller to gate, connection resolved by id/realm fingerprint, every write carries that connection's `(integration_id, partner_id)` via the composite FK).

**Web**
- `QuickbooksIntegration.tsx`: pull-payments toggle, "Sync now" action, QuickBooks payment badge, void-suppression, i18n across all 8 locales. Added to `no-silent-mutations.test.ts`'s `TARGET_GLOBS` (closing pre-existing gap — the file already had run-action mutations that nothing was guarding).

**Docs**
- `accountingCurrency.ts`: item 4 marked delivered, recording the actual lock order and the composite-key refinement.
- `docs/integrations/quickbooks-sandbox-verification.md`: Phase D checklist (items 17-26), status PENDING — no live sandbox run in this task.
- `apps/docs/src/content/docs/features/accounting-integrations.mdx`: payment pull-back feature docs.

## Decisions locked

1. **Composite payment remote id.** `accounting_entity_mappings.remote_entity_id` for a payment mapping is `'<PaymentId>/<remoteInvoiceId>'`, never the bare QBO Payment id. One QBO `Payment` legitimately settles several Breeze invoices (a split payment carries one `Line` per invoice); a bare payment id would let only the first split line claim `accounting_entity_mappings_remote_uniq` and the rest would collide. `reverseAccountingPayment` recovers the whole set with a `<PaymentId>/%` prefix match.
2. **Mirror-delete reversal.** A payment deleted in QBO deletes only the mirrored Breeze row and recomputes the invoice; a manually-recorded Breeze payment on the same invoice is untouched. Over-payment is allowed on the pull path (unlike the manual `recordPayment` throw) because QuickBooks is reporting money that already moved — refusing it would leave Breeze permanently disagreeing with the ledger.
3. **Webhook enqueues + CDC applies.** The webhook route never reads or acts on payload entity ids — it verifies, resolves realm → connection, and enqueues a reconcile job. `reconcileChanges` (CDC) is the only source of truth for what changed.
4. **Single webhook URL / sweep is the guarantee.** Intuit allows one webhook URL per app; the hosted contract uses one Intuit app for both regions, so only the US deployment gets webhook notifications. The 15-minute sweep reaches the same end state everywhere and is the guaranteed path; the webhook is a latency optimisation only.
5. **Realm fingerprint.** Connections are looked up by an HMAC fingerprint of the realm id (`hmacFingerprint`), never the raw realm id, so the webhook route never persists or logs a bare realm id.
6. **`pull_payments`.** Off by default reasoning: pull and (future) push both writing the same invoice would double-count until the push slice ships. Off ⇒ the sweep no-ops for that connection with no QBO call at all; the backlog catches up on the next sweep after re-enabling.

## Review fixes landed during execution

- Task 2: CDC test network-leak fix; server-time cursor (not a synthetic clock) so a stale local clock can't skew the cursor.
- Task 3: void-invoice guard (payment against a Breeze-voided invoice is a clean `invoice_void` marker, not a retried failure) + phantom-claim throw (a concurrent writer's mapping row vanishing between commit and re-read fails the item rather than reporting a false `replayed`, which would otherwise silently advance the cursor past a payment that exists nowhere).
- Task 5: Sentry capture throttling on the webhook route, 50-realm fan-out cap + chunking, mount-order guard (route sits outside the auth middleware chain, verified by `webhooks.mountOrder.test.ts`).

## Verification

All foreground, this branch merged with `origin/main` (`4274644ac`) first — clean merge, no conflicts.

| Step | Result |
|---|---|
| `apps/api` full unit suite (`npx vitest run`) | 1744 files / 31516 tests passed, 3 files / 22 tests skipped |
| Integration shard 1/4 | 114 files / 761 tests passed, 3 skipped |
| Integration shard 2/4 | 114 files / 733 tests passed |
| Integration shard 3/4 | 113 files / 801 tests passed |
| Integration shard 4/4 | 113 files / 656 tests passed |
| `vitest.config.rls-coverage.ts` | 1 file / 100 tests passed |
| `vitest.config.rls.ts` | 2 files / 31 tests passed |
| `pnpm db:check-drift` | clean — 626 migration files match the ledger |
| `apps/web` full unit suite | 692 files / 7199 tests passed |
| `pnpm lint` | clean (6/6 tasks) |
| `tsc --noEmit --project apps/api/tsconfig.json` (8 GB heap) | 0 errors |
| `astro check` (apps/web, 8 GB heap) | 0 errors, 0 warnings (248 pre-existing hints) |
| `docker compose ls -a \| grep payment-pullback` post-teardown | empty |

No branch-caused failures; none of the four known pre-existing main reds (#4500-#4503) surfaced in this run.

## Not in this PR

- Payment **push** (Breeze → QuickBooks) — the next slice; this is why `pull_payments` exists as an independent switch.
- Foreign-currency payments.
- Credit memos and QuickBooks-originated refunds.
- Cross-region webhook forwarding.
- The sandbox walkthrough (checklist items 17-26 in `quickbooks-sandbox-verification.md`) — status PENDING, requires a live Intuit sandbox + tunnel.

https://claude.ai/code/session_01BdvjnGttKy7i5oPuGz5HnJ


ToddHebebrand:
## Final-review fix wave — all 9 findings addressed (`ac953ec26..608d487d8`)

Each fix was driven red-first; RED output and GREEN commands per finding are in `.superpowers/sdd/2026-09-02-quickbooks-phase-d-payment-pullback/final-fix-report.md`.

**A — CDC overflow lost data.** Window-halving is gone: QBO's `/cdc` takes only `changedSince`, so the "left half" re-issued a byte-identical request and a truncated window was returned as complete while the cursor advanced past everything withheld. An overflowing entity is now re-enumerated through `/query` (`... where MetaData.LastUpdatedTime >= '<since>' orderby ... startposition n maxresults 1000`) until a short page, capped at 50 pages. Query rows are authoritative for the ids they cover, so a payment's CDC lines are rebuilt rather than merged; the CDC deletion lists survive verbatim. `ChangeSet.overflowed` flags "could not fully enumerate"; the worker applies what QBO did return, holds the cursor, captures and rethrows.

**B — payment edit that drops an allocation.** A QBO Payment edited to settle a different invoice set emits the Payment, never a deletion, so the stale Breeze payment row and the balance it paid down stood forever. The worker now reconciles per Payment and calls the new `reverseStaleAllocations`, reversing every `<PaymentId>/<invoice>` mapping not in the current line set through the same row-level path a full reversal uses.

**C — reconnect to a different realm.** Three guards: the OAuth callback wipes the mappings and nulls `cdc_cursor`/`last_reconcile_at` when the prior realm is positively known to differ (audited; never on an unreadable read); `advanceReconcileCursor` is a compare-and-set on the realm fingerprint captured at job start, returning a boolean and logging+skipping on a miss; and all four applier entry points re-read the connection row inside their own transaction, returning the new clean `realm_changed` outcome without writing.

**D — settings permission.** `PATCH /:provider/settings` now runs `requireInvoicePush` when the body carries `pushMode` or `pullPayments` (account-refs stay ungated, system scope still bypasses). The web panel hides both controls without `invoices:write`, keeping the org-scope gate.

**E — fingerprint backfill.** One transaction per row instead of one for the sweep: a caught unique violation used to leave the transaction aborted, 25P02-ing every later row and rolling back every earlier one. Proven against real Postgres.

**F — webhook partial failure.** Any failed enqueue now answers 503 rather than only "all failed" (the deterministic jobId makes Intuit's retry harmless), and the realm lookups within a chunk run sequentially — they share one transaction, so a concurrent rejection 25P02'd the rest.

**G — payment errors clobbering push errors.** Payment-originated markers carry the `Payment pull: ` prefix; a later `applied`/`updated` clears only a prefixed marker, scoped by the prefix in SQL. "Deleted in QuickBooks" is unprefixed and unclearable from this path. Docs copy updated.

**H — operator visibility.** A failed or truncated run stamps a sanitized one-liner (counts only) on `accounting_connections.last_error`, cleared on the next clean run and prefix-scoped so it cannot wipe a reauth message. An empty `TxnDate` is refused with an explicit message before any write instead of surfacing as a driver error on `received_at`. A cursor older than the 30-day CDC floor logs a warning and raises one Sentry event per run naming the gap; the docs claim that the sweep "catches up on anything missed" is corrected.

**I — env plumbing.** All five `QBO_*` vars are documented in `.env.example` and mapped as `${VAR:-}` in the `x-api-env` block of `docker-compose.yml` and `deploy/docker-compose.prod.yml`, pinned on all three axes by `envComposeParity.test.ts`.

**Verified:** full `apps/api` unit suite 1744 files / 31584 passed; 36 accounting + 17 RLS/cascade integration tests on real Postgres; 276 web tests; api + web `tsc` clean; `pnpm lint` 6/6; `db:check-drift` clean. Note for reviewers: `advanceReconcileCursor` now returns a boolean instead of throwing on a zero-row match (the ruled behaviour — a realm change must not fail the job), the applier entry points take a trailing `expectedRealmFingerprint`, and `ChangeSet` gains `overflowed`.

PR #4540, opened 2026-09-02
fix(portal): serve the dev module graph under /portal so islands hydrate (#3906)

### LINKED ISSUE #3906: wt-stack: no portal island hydrates — Astro dev module URLs miss the /portal Caddy route, so portal e2e asserts dead markup

ToddHebebrand:
Closes #3906

## The problem

No portal React island hydrated in a `pnpm wt-stack` / dev stack. Pages SSR'd correctly and then died silently — the markup looked right, so a browser-level check passed while nothing on the page was interactive. Portal e2e specs were asserting against dead markup.

`astro dev` deliberately keeps Vite base-agnostic: it strips the configured `base` off every incoming request in its own front-of-stack middleware (`astro/dist/vite-plugin-astro-server/base.js`) and **never sets Vite's `base`**. So the dev server emits its module URLs from the server root — `/src/*`, `/@fs/*`, `/@id/*`, `/@vite/client`, `/node_modules/.vite/deps/*`. Caddy routes the portal **by path** (`/portal` and `/portal/*`), so those fell through to the web catch-all: a portal-only file 404'd, and a path both Astro apps happen to share was silently served from the **wrong app**.

### Why #4425 could not work

#4425 tried to route the unprefixed paths by sniffing a `/portal` `Referer`. Three independent reasons that gate can never fire on the requests that actually hydrate an island:

1. The `astro-island` dynamic `import()` sends **no `Referer` at all**.
2. Nested/transitive imports send the **importing module's own** unprefixed `/@fs/...` URL as `Referer`.
3. `apps/portal/src/middleware.ts` sets `Referrer-Policy: no-referrer` on exactly the `/quote/` and `/invoice/` pages the gate was meant to serve.

## The fix (option C — fix the emission side)

Make the portal dev server emit base-prefixed module URLs, so the existing path rule already covers them. No header sniffing, no dev-only proxy rules.

- **`breeze:portal-dev-base`** (`apps/portal/astro.config.mjs`) sets Vite's `base` in dev. That prefixes the entire module graph: import analysis, pre-bundled deps, HMR client, react-refresh. It also removes Vite's own base-stripping middleware — Astro has already stripped `/portal` by the time it runs, so Vite's would double-strip and 404 **every** request with *"The server is configured with a public base URL of /portal/…"* (verified live, see below). The plugin throws a loud, explanatory error if that middleware ever stops being findable rather than silently reding out dev. `apply: 'serve'` keeps `astro build` untouched.
- Vite's `base` never reaches the handful of entry-point URLs **Astro itself hardcodes** into the HTML (`component-url`, `renderer-url`, `before-hydration-url`, `/@vite/client`, `/src/styles/*`) — and those are precisely what starts hydration. `src/middleware.ts` prefixes them in dev via the new pure helper `src/lib/devAssetBase.ts` (13 unit tests).
- **Caddy cleanup:** the `@portalDevAssets` Referer matcher and its `CADDY_PORTAL_DEV_ASSETS` env gate are deleted. The single `@portal` path rule now covers dev and prod alike.

### Production is unaffected

`PORTAL_BASE_PATH` is baked in at build time and prod already serves the portal under `/portal` on the main domain (`docker/Caddyfile.prod`, `deploy/docker-compose.prod.yml`), so dev and prod are now consistent. Verified by building:

- `astro build` still emits `"/portal/_astro/…"` asset URLs — **no double prefix** anywhere in `dist/`.
- `grep -rl "prefixDevAssetUrls\|@react-refresh" dist/server` → **empty**: the dev rewrite is dead-stripped from the production bundle.

## Live verification (wt-stack, Playwright)

Stack: `pnpm wt-stack up` on this branch → `http://localhost:32806`.

**`/portal/login`** — island hydrated:
```
component-url = "/portal/src/components/portal/LoginForm.tsx"
ssr attribute  = dropped (island.hasAttribute('ssr') === false)
React fiber elements on page = 10
console errors/warnings = 0
[vite] connected.   ← HMR websocket works through Caddy too
```
All 60+ module requests were `/portal/…` and **200** — zero unprefixed, zero 404s.

**Public quote page** (`/portal/quote/<token>` — the `no-referrer` page #4425's gate could never serve):
```
component-url = "/portal/src/components/portal/PublicQuoteView.tsx"
ssr attribute  = dropped
public-quote-accept has __reactFiber$ = true
React fiber elements on page = 63
169 network requests, all /portal/…, all 200/304, zero 404s
console errors/warnings = 0
```
Interactivity proven, not just presence — typing a signer name and ticking the agreement box flipped the accept button through React state:
```
aria-disabled: "true"  →  "false"      (disabled prop: false)
```
That is the exact symptom from the issue ("accept button stays `aria-disabled`, zero non-GET requests").

**Regression check — the web app is unchanged.** Removing the Caddy block returns those root paths to web, which is correct since web is the catch-all. `/login` on the web app: 2 islands, **0** still-`ssr`, 58 React fiber elements, 0 console errors. Web's dev URLs stay unprefixed at the root; the two apps no longer overlap at all.

**Edge paths through Caddy:** `/portal` → 302 `/portal/login`, `/portal/` → 302, `/portal/quotes` → 302 (auth), `/portal/nope-404` → 404, `/` → 200 from web. `caddy validate` → `Valid configuration`.

## Tests

| Check | Result |
|---|---|
| `e2e-tests/tests/multi-currency.spec.ts` (live stack) | **1 passed (48.8s)** — clears `waitForHydration(publicPage, 'public-quote-accept')` |
| `apps/portal` vitest (full suite, post-merge with main) | **27 files / 272 tests passed** |
| `apps/portal/src/lib/devAssetBase.test.ts` (new) | 13 passed; mutation-checked (breaking the regex reds it) |
| `astro check` (portal) | **0 errors**, 0 warnings |
| `tsc --noEmit` (e2e-tests) | clean |
| `pnpm lint` | 6/6 tasks successful |
| `pnpm test:dev-scripts` | 6 files / 23 tests passed |
| `caddy validate` | Valid configuration |

`quote-contract-proposal.spec.ts` has an unrelated pre-existing failure earlier in the **web** flow (AuthOverlay / session-expired on the contracts tab) and was not chased here — only its stale `@portalDevAssets` comment is updated.

### DIFF
--- a/apps/portal/astro.config.mjs
+++ b/apps/portal/astro.config.mjs
@@ -59,6 +59,15 @@ function portalDevBase(basePath) {
         // of /portal/ — did you mean to visit ...". Astro's is the one that has to
         // stay: it also handles `/portal` exactly, trailing slashes and the
         // outside-the-base redirect. Drop the redundant one.
+        //
+        // Loudness caveat on the throw below: from a cold `astro dev` it
+        // propagates uncaught out of `vite.createServer` and kills the process,
+        // which is what we want (verified — the container logs the full message
+        // and the portal never comes up). On a *live restart* (editing this file
+        // while the dev server runs) Vite's own `restartServer` catches it, logs
+        // "server restart failed" and keeps serving the previous config, so the
+        // signal is two easy-to-miss log lines plus a stale server. Restart the
+        // process if a base-path change does not seem to take.
         const stack = server.middlewares.stack;
         const index = stack.findIndex((layer) => layer.handle?.name === 'viteBaseMiddleware');
         if (index === -1) {
--- a/apps/portal/astro.config.mjs
+++ b/apps/portal/astro.config.mjs
@@ -9,6 +9,72 @@ import node from '@astrojs/node';
 // to serve at the root. See docker/Caddyfile.prod for the matching `/portal` carve-out.
 const PORTAL_BASE = process.env.PORTAL_BASE_PATH || '/portal';
 
+/**
+ * Dev-only: mount the Vite dev server's module graph under the portal's base path.
+ *
+ * `astro dev` deliberately keeps Vite base-agnostic. It strips `base` off every
+ * incoming request in its own front-of-stack middleware
+ * (astro/dist/vite-plugin-astro-server/base.js) and never sets Vite's `base`, so
+ * the dev server emits its module URLs from the server root: `/src/*`, `/@fs/*`,
+ * `/@id/*`, `/@vite/client`, `/node_modules/.vite/deps/*`.
+ *
+ * That is fine when the browser talks to the dev server directly, but the
+ * worktree/dev stack puts Caddy in front of both Astro apps and routes the portal
+ * **by path** (`/portal` and `/portal/*` — docker/Caddyfile.prod). Unprefixed
+ * module URLs therefore fell through to the web app: a portal-only file 404'd, and
+ * a path both apps happen to share (`/src/lib/utils.ts`, `/node_modules/...`) was
+ * silently served from the *wrong app*. Either way no portal island hydrated,
+ * while the SSR'd markup still looked correct — so portal e2e specs asserted
+ * against dead markup (#3906).
+ *
+ * Setting Vite's `base` makes Vite emit `/portal/`-prefixed URLs across the whole
+ * module graph (import analysis, pre-bundled deps, HMR client, react-refresh), so
+ * the existing path-based Caddy route already covers them — no Referer sniffing
+ * and no dev-only proxy rules. Astro's own middleware strips the prefix back off on
+ * the way in, exactly as it already does for page routes. The handful of entry-point
+ * URLs Astro hardcodes into the HTML itself are prefixed in src/middleware.ts
+ * (see src/lib/devAssetBase.ts) — Vite's `base` does not reach those.
+ *
+ * `apply: 'serve'` keeps this out of `astro build`: a production build already emits
+ * base-prefixed bundled assets via Astro's own `base`, and forcing Vite's `base`
+ * there would double-prefix them.
+ */
+function portalDevBase(basePath) {
+  // Vite wants a trailing slash; the `base` convention here does not carry one.
+  const viteBase = basePath === '/' ? '/' : `${basePath.replace(/\/+$/, '')}/`;
+  const enabled = viteBase !== '/';
+
+  return {
+    name: 'breeze:portal-dev-base',
+    apply: 'serve',
+    config: () => (enabled ? { base: viteBase } : {}),
+    configureServer(server) {
+      if (!enabled) return undefined;
+      // Post hook — runs after Vite has installed its own middleware stack.
+      return () => {
+        // Vite installs its own base-stripping middleware whenever `base !== '/'`.
+        // Astro has already stripped `/portal` off `req.url` by the time it runs
+        // (Astro unshifts its middleware to the front of the stack), so Vite's would
+        // reject every request with "The server is configured with a public base URL
+        // of /portal/ — did you mean to visit ...". Astro's is the one that has to
+        // stay: it also handles `/portal` exactly, trailing slashes and the
+        // outside-the-base redirect. Drop the redundant one.
+        const stack = server.middlewares.stack;
+        const index = stack.findIndex((layer) => layer.handle?.name === 'viteBaseMiddleware');
+        if (index === -1) {
+          throw new Error(
+            "[breeze:portal-dev-base] Vite's base middleware ('viteBaseMiddleware') is not in the " +
+              'dev middleware stack. Either Vite stopped installing it or it was renamed — without ' +
+              'removing it Astro and Vite both strip the base path and every dev request 404s. ' +
+              'Re-check this plugin against the installed Vite version.'
+          );
+        }
+        stack.splice(index, 1);
+      };
+    }
+  };
+}
+
 export default defineConfig({
   output: 'server',
   base: PORTAL_BASE,
@@ -42,6 +108,6 @@ export default defineConfig({
     port: 4322
   },
   vite: {
-    plugins: [tailwindcss()]
+    plugins: [tailwindcss(), portalDevBase(PORTAL_BASE)]
   }
 });
--- a/apps/portal/src/lib/devAssetBase.test.ts
+++ b/apps/portal/src/lib/devAssetBase.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, it } from 'vitest';
+import { isHtmlContentType, prefixDevAssetUrlsFor } from './devAssetBase';
+
+/**
+ * Regression cover for #3906: the portal dev server emits its module URLs from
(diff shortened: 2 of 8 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (production-safety + regex correctness lens) and `silent-failure-hunter` (this bug *is* a silent-failure bug, so that lens was the priority).

**Findings:** 4 raised → 3 addressed in `160cc543d` + `f6aa4df62`, 1 documented-not-fixed (below). `code-reviewer` returned **0 findings** at critical or important, having independently re-verified `apply: 'serve'`, the `import.meta.env.DEV` gate, root-deploy (`PORTAL_BASE_PATH=/`) no-ops, regex idempotency/backtracking/`lastIndex`, and the `viteBaseMiddleware` name against the installed `vite@8.2.2` source.

| # | Finding | Disposition |
|---|---|---|
| 1 | Rewrite enumerated the four `/@…` namespaces we happened to observe; a new Vite namespace would miss **silently** — SSR fine, nothing hydrates, i.e. #3906 again | **Fixed** in `f6aa4df62`. Match the whole reserved `/@` prefix instead of a hand-maintained list. New test: a made-up `/@some-future-vite-namespace/` is prefixed. No portal route begins with `/@`. |
| 2 | Null-body status (204/304) carrying `text/html` would make `new Response('', {status:304})` **throw** → a 500 | **Fixed** in `160cc543d` (self-caught before the review landed). Conditions moved into a pure `shouldPrefixDevAssetUrls`, mirroring `resolvePortalCspHeader`; the test asserts the constructor really does throw, so the guard is load-bearing, not decorative. |
| 3 | The plugin's throw is loud on a cold `astro dev`, but Vite's `restartServer` **catches** it on a live restart and keeps serving the previous config | **Documented** in `f6aa4df62` — the loudness envelope is now written down at the throw site. Verified live: on a cold start the container logs the full message and the portal never comes up. |
| 4 | The `waitForHydration` guards are non-vacuous but **are not wired into any required CI check** | **Not fixed — flagged deliberately.** Pre-existing and repo-wide: the only Playwright invocation in CI is the narrow `playwright.auth-browser-transition.config.ts`, and `playwright.config.ts` has no `webServer` (it reads a `.breeze-stack.json` that only exists once a stack is up by hand). Wiring a dev-mode e2e job into required checks is an infra/runtime-cost decision for you, not something to smuggle into this PR. **Consequence to be aware of: a future regression here would pass every required check silently.** |

**Verification (live, re-run against the final commit):**
- Public quote page: `component-url="/portal/src/components/portal/PublicQuoteView.tsx"`, `ssr` attribute dropped, accept button carries a `__reactFiber$`, 63 fiber elements, 0 console errors. 169 network requests, all `/portal/…`, all 200/304, **zero 404s**.
- Interactivity, not just presence: typing a signer name + ticking the box flipped the accept button `aria-disabled` `"true"` → `"false"` — the exact symptom from the issue.
- `e2e-tests/tests/multi-currency.spec.ts` — **1 passed (34.5s)**, clearing `waitForHydration(publicPage, 'public-quote-accept')`.
- Crawled every portal page type (login, forgot-password, quote, invoice-bad-token, 404) for leftover unprefixed dev URLs: **none**. The only root-relative attribute left is Vite's `data-vite-dev-id`, which holds an absolute *filesystem* path rather than a URL and is correctly untouched.
- Regression: web app unchanged — `/login` has 2 islands, **0** still-`ssr`, 58 fiber elements, 0 console errors.

**Tests:** `apps/portal` **27 files / 278 tests passed**; `devAssetBase.test.ts` mutation-checked twice (breaking the regex, and dropping an entry from `DEV_ASSET_NAMESPACES`, each red a test — so the const is genuinely wired to the regex, not decorative). `astro check` 0 errors; `tsc --noEmit` (e2e-tests) clean; `pnpm lint` 6/6; `pnpm test:dev-scripts` 23 passed; `caddy validate` Valid configuration.

**Prod safety:** `astro build` still emits `/portal/_astro/*` with no double prefix, and `grep -rl "prefixDevAssetUrls" dist/server` is empty — the dev rewrite is dead-stripped from the production bundle.

**Status:** review-clean, awaiting maintainer merge. Not merged, and #3906 left open per the hand-off rule.

PR #4592, opened 2026-09-02
fix(agent): treat a never-provisioned elevation account as clean in PAM cleanup (#4587)

### LINKED ISSUE #4587: [Agent] PAM v2 0.109.0: 'PAM disable cleanup unverified: account:account_cleanup_failed' logged at ERROR every heartbeat on deployments that never enabled PAM
## Summary

On a Windows deployment that has **never** enabled PAM (`pam_enabled: false`, the default), a 0.109.0 agent logs at ERROR on **every heartbeat**:

```
PAM disable cleanup could not be verified error="PAM disable cleanup unverified: account:account_cleanup_failed"
```

Expected: no-op success. PAM was never enabled, there is nothing to clean up.

Reported by a community user on Discord while troubleshooting an unrelated Defender/offline issue. Not present in v0.108.0; introduced by #4105 (`fix(pam): make actuation lifecycle durable and restart-safe`).

## Root cause (verified in source)

1. The dormant `~breeze_elev` account is only provisioned when `cfg.PAMEnabled` is true (`agent/internal/agentapp/main.go:890`). Default is `false`, so on a never-PAM deployment the account does not exist.
2. The PAM lifecycle manager is constructed unconditionally in `Heartbeat.SetStatePath` (`agent/internal/heartbeat/heartbeat.go:1094`), with initial state `enabled: true, admissionOpen: true, available: true` (`pamlifetime/manager.go:111`).
3. Every heartbeat response calls `handleUACInterception(response.UacInterceptionEnabled)` (`heartbeat.go:4445`). A nil/false policy calls `pamLifetimeManager.SetEnabled(ctx, false)`. Because `m.enabled` starts `true`, the early-return at `manager.go:495` is skipped and, with an empty ledger, `verifyAccountClean` runs (`manager.go:508`).
4. `verifyAccountClean` -> `account.Deprovision` -> `setPassword("~breeze_elev", ...)` -> `NetUserSetInfo` returns `NERR_UserNotFound` (2221). `setPassword` does not tolerate that (unlike `removeFromAdministrators`, which already maps 2221 to nil), so the manager reports `account_cleanup_failed`.
5. On failure `m.enabled` is never cleared (only set `false` after success at `manager.go:535`), so the next heartbeat repeats steps 3-4. Hence the ERROR on every heartbeat.

The reporter's theory that `pam_lifetime_protocol_version = 0` causes this is backwards: the failure stores `pamVerificationAvailable=false`, and `pamLifetimeProtocolVersion()` (`heartbeat.go:1131`) returns 0 whenever that flag is false. The server is not involved.

## Impact

- ERROR-level log spam every heartbeat on every Windows agent in never-PAM fleets (the majority of the fleet).
- The device also permanently refuses to enable UAC interception (`refusing to enable UAC interception before PAM reconciliation is verified`) until restart, if the customer later turns PAM on.
- No security impact: the failure is fail-closed.

## Fix shape

Two layers, both small:

1. **`agent/internal/elevaccount/elevaccount_windows.go`**: in `Deprovision` and `VerifyClean`, treat `NERR_UserNotFound` (2221) from `NetUserSetInfo` / `NetUserGetInfo` as "account absent = clean" and return `AccountEvidence{Enabled: false, InAdministrators: false}, nil`. This matches the existing handling in `removeFromAdministrators`. An absent account is provably not enabled and not in Administrators.
2. **`agent/internal/pamlifetime/manager.go` `SetEnabled(false)`**: the disable path must be idempotent. After the empty-ledger account verification succeeds, `m.enabled` is already set `false` at the end; the loop only exists because the verification fails. With fix 1 the first `SetEnabled(false)` succeeds and subsequent calls hit the early return. Add a regression test in `pamlifetime` with a fake `accountLifecycle` whose `Deprovision`/`VerifyClean` model an absent account, asserting `SetEnabled(false)` returns nil, and a second call does not invoke the account manager again.
3. Optional hardening: gate manager construction on `cfg.PAMEnabled` in `SetStatePath` so never-PAM agents skip the machinery entirely. Consider whether that breaks the reconciliation / `pamReceivedObservationReady` readiness contract before doing it.

## Acceptance

- Never-PAM Windows agent: zero PAM ERROR lines across heartbeats; `pamLifetimeProtocolVersion` advertised per existing rules (0 is fine when unavailable, but not because of a spurious failure).
- Existing `pamlifetime` and `heartbeat` PAM tests still pass (`cd agent && go test -race ./internal/pamlifetime/... ./internal/heartbeat/... ./internal/elevaccount/...`).
- A genuinely failing `NetUserSetInfo` (any errno other than 2221) still reports `account_cleanup_failed`.

ToddHebebrand:
## Root cause

The dormant `~breeze_elev` account is only created when `cfg.PAMEnabled` is set, so on the majority of the fleet it does not exist. Every heartbeat without a UAC-interception policy calls `pamLifetimeManager.SetEnabled(ctx, false)`; with an empty ledger that runs `verifyAccountClean` against that missing account, which failed. Because `m.enabled` is only cleared on success, the next heartbeat repeated the whole cleanup — hence `PAM disable cleanup could not be verified` at ERROR on every heartbeat, forever, on every never-PAM Windows agent.

There were **two independent failure points**, not the one in the issue. `elevaccount`'s `Deprovision`/`VerifyClean` returned `NERR_UserNotFound` (2221) from `NetUserSetInfo`, `NetUserGetInfo` and `NetUserGetLocalGroups` as a real error — but `verifyAccountClean` then calls `VerifyNoPrivilegedToken`, whose first statement resolves the account name to a SID and fails with `ERROR_NONE_MAPPED` (1332) when no such account exists. Fixing only `elevaccount` would have swapped `account_cleanup_failed` for `privileged_token_verification_failed` and left the loop running. Confirmed by an independent audit of the call chain before writing the fix.

## What changed

- **`agent/internal/elevaccount/absent.go`** (new, untagged): one predicate, `IsAccountAbsent`, matching `NERR_UserNotFound` and `ERROR_NONE_MAPPED` as wrapped `syscall.Errno` values. It is deliberately not behind `//go:build windows` so the classification that decides whether the agent reports a cleanup failure is compiled and unit-tested on every platform.
- **`elevaccount_windows.go`**: `Deprovision` and `VerifyClean` return `AccountEvidence{Enabled: false, InAdministrators: false}, nil` when the account is reported absent — an account that does not exist is provably not enabled and provably not in Administrators. Every other status stays a real error.
- **Absence is decided by the probe that names the account**, not by a status code seen at the aggregate level. `accountInAdministrators` and `localGroupMembersCall` both resolve the builtin Administrators alias (`S-1-5-32-544`) *before* they touch the elevation account, and that lookup fails with the very same `ERROR_NONE_MAPPED` — so a machine that cannot resolve its own Administrators group would have read as "elevation account absent, all clean". A package sentinel `errAccountAbsent` is attached only by the syscall that named the account (`NetUserSetInfo`/`NetUserGetInfo`/`NetUserGetLocalGroups` 2221, `LookupSID(username)` 1332); `cleanIfAbsent` and `removeFromAdministrators` match that and nothing else. `IsAccountAbsent` stays as a documented single-call classifier. A measured `enabled == true` followed by an absent report is a contradiction that now fails closed instead of returning clean.
- **`pamlifetime/store.go`**: `Store.Entries()` ignored `loadErr`, so an unreadable ledger was indistinguishable from a device that never actuated. `LoadError()` is now part of the `lifetimeStore` contract, `SetEnabled(false)` fails closed with `account:ledger_unavailable` before the empty-ledger branch, and the never-actuated condition is derived inside `verifyAccountClean` from both signals rather than trusting a caller-supplied bool.
- **`pamlifetime/manager.go`**: the token scan is **not** softened at the primitive. `VerifyNoPrivilegedToken` is also called while reconciling a recorded actuation, and a Windows process keeps its elevated SID after the account behind it is deleted — so an unresolvable name there could be hiding exactly the orphaned elevated process the scan exists to catch. Instead `verifyAccountClean` takes an explicit `neverActuated` flag, true only on the empty-ledger disable path, and tolerates an absent-account lookup failure only there. `reconcileActiveLocked` passes `false`. Every other error, and every error at all once an actuation is on record, still fails closed.
- **Evidence and observability**: a skipped token scan leaves `PrivilegedTokenPresent` nil rather than `boolPtr(false)` — not measured is not the same claim as measured absent — and the skip emits one INFO line, once per agent lifetime (the settled disable returns at the early return on every later call).
- **`.github/workflows/ci.yml`**: `./internal/elevaccount` joins the Windows Go test job. The package is entirely `//go:build windows` netapi32 code whose only Windows-tagged test was a compile-time interface assertion that no CI job ever ran — which is why a mishandled status code in the *default* configuration shipped. Verified the same way the job's comment requires for `logging`/`ipc`: under `GOOS=windows GOARCH=amd64 CGO_ENABLED=0` the package vets, its test binary compiles, and its Windows dependency graph contains no `remote/desktop` and no cgo.

## Tests

Written first and watched fail:

- `elevaccount/absent_test.go` — 20-case table over `IsAccountAbsent`. Runs on **every** platform (so it executes in the existing Linux `Test Agent` job). Covers both wrap shapes the production wrappers use, double wrapping, and negatives: access denied, `NERR_UserExists`, `NERR_UserNotInGroup`, off-by-one on both statuses, a non-errno error, and an error carrying `2221` only in its message text (proves it is not a string match). **Red before the fix**: `undefined: IsAccountAbsent`.
- `pamlifetime/manager_never_enabled_test.go` — `TestSetEnabledFalseToleratesUnresolvableAccountNameInTokenScan` is the genuine regression test. **Verified red before the fix** by flipping `neverActuated` to `false`: `SetEnabled(false) = PAM disable cleanup unverified: account:privileged_token_verification_failed, want nil`. Alongside it: the disable path is a quiet idempotent no-op and a second call does not re-enter the account manager (`Deprovision`/`VerifyClean` call counts pinned at 1); genuine `account_cleanup_failed` and `account_verification_failed` still surface, stay fail-closed and retry; five other token-scan failures (access denied, invalid parameter, non-errno, status-in-text, and an actually-found privileged token) still fail; and `TestReconcileDoesNotTolerateUnresolvableAccountNameInTokenScan` pins the safety boundary — with an actuation on record the same error must still fail.
- `elevaccount/absent_windows_test.go` (Windows-only) — exercises the real netapi32 path: `Deprovision` and `VerifyClean` against the never-provisioned `~breeze_elev`. They skip by design if the account exists or local-account queries are refused, so `ci.yml` also gets a counted step (the shape the `collectors` and `backup` steps already use) that runs exactly these two with `-v` and fails on a SKIP or on a PASS count other than two. The shared Windows step runs without `-v`, where a skip is indistinguishable from a pass — and these are the only live-syscall coverage anywhere of this fix. Also asserts the untagged `errorNoneMapped` constant still equals `windows.ERROR_NONE_MAPPED`.
- `elevaccount/absent_test.go` — `TestCleanIfAbsentConvertsOnlyAbsenceIntoCleanEvidence` covers the shared `cleanIfAbsent` decision all four probe sites delegate to, including the misuse case (handed a nil error it returns an error rather than manufacturing clean evidence for an account nobody probed).

Commands run locally, foreground:

```
cd agent && go build ./... && GOOS=windows GOARCH=amd64 go build ./...
  -> both OK (the elevaccount change is Windows-only code; it cross-compiles)

cd agent && CGO_ENABLED=0 go test -count=1 ./internal/elevaccount/ ./internal/pamlifetime/ ./internal/heartbeat/
  ok  internal/elevaccount   0.238s   (30 PASS, incl. subtests)
  ok  internal/pamlifetime   1.221s   (99 PASS, incl. subtests)
  ok  internal/heartbeat    34.483s

cd agent && go vet ./internal/elevaccount/... ./internal/pamlifetime/...
cd agent && GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go vet ./internal/elevaccount/... ./internal/pamlifetime/...
  -> both clean

cd agent && GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/elevaccount/
  -> Windows test binary compiles
```

`go test -race` was also run and is green for `elevaccount` and `pamlifetime`. It cannot be used for `heartbeat` on this machine: `-race` forces cgo on darwin/arm64 and `github.com/shoenig/go-m1cpu` SIGSEGVs in its package `init`, before any test runs. That is pre-existing and environment-specific — CI runs the agent suite with `CGO_ENABLED=0`, which is the command reported above.

## Review round

Three lenses ran (`silent-failure-hunter`, `code-reviewer`, `pr-test-analyzer`). `code-reviewer`: 0 findings. The other two raised three, all addressed in `47a4965` and `5464dc7`:

1. **The fail-closed guarantee was implicit.** For an agent with an actuation on record it rested on `VerifyNoPrivilegedToken` happening to resolve the account name as its first statement — and because the pamlifetime tests drive a *fake* primitive, a future refactor that special-cased "no such account" there would void the contract without reddening a single test. There is now a comment at that `LookupSID` call stating why absence must not be interpreted locally, pointing at `verifyAccountClean`'s `neverActuated` parameter.
2. **Two of the four absence guards shipped unexercised.** Only the first probe in each of `Deprovision`/`VerifyClean` is reachable on a never-provisioned host, so per-site copies of the decision were dead in context. They now share one untagged, unit-tested helper.
3. **A test doc comment overclaimed.** `TestSetEnabledFalseOnNeverProvisionedAgentIsAQuietIdempotentNoOp` called itself "the manager half of issue #4587" but passes with and without the fix — it pins the idempotence contract that turns one failure into an unbounded ERROR loop, not the tolerance. Corrected, and it now names the test that does detect the regression. The reviewer confirmed by reverting `manager.go` wholesale that exactly one test fails: `TestSetEnabledFalseToleratesUnresolvableAccountNameInTokenScan`.

One finding was considered and **not** taken: threading `neverActuated` down into `elevaccount.Deprovision`/`VerifyClean` themselves. Ledger state is manager policy, not something the account layer should know about; scoping it at the caller keeps `elevaccount` reporting facts about the account and leaves one place that decides what they mean.

### Second round (five lenses)

A second independent round found that round one's shared classifier was applied at the wrong level. All findings fixed in `c085c5e`, each red-first:

1. **HIGH — absence inferred from the wrong principal's failure.** Details in "What changed" above. Red-first proof: with `cleanIfAbsent` back on the aggregate errno match, `TestCleanIfAbsentIgnoresAbsenceStatusesFromOtherPrincipals` fails with `cleanIfAbsent(errno 1332) returned evidence {false false} and no error`. This was a defect introduced by round one's own widening of the predicate.
2. **MEDIUM — an unreadable ledger read as "never actuated".** Red-first: without the guard, `TestSetEnabledFalseFailsClosedWhenTheLedgerCannotBeRead` gets `privileged_token_verification_failed` instead of `ledger_unavailable`, and the account manager gets called on a device whose history is unknown.
3. **MEDIUM — unmeasured token evidence reported as measured-false.** Red-first: `TestToleratedTokenScanLeavesPrivilegedTokenEvidenceUnmeasured/skipped_scan` fails with `PrivilegedTokenPresent = false, want nil`.
4. **MEDIUM — the tolerance was invisible.** Now one INFO line on the skipped-scan branch.
5. **Comment accuracy** (`comment-analyzer`): an errno may arrive as *or unwrap to* a `syscall.Errno`; `NERR_InvalidComputer` is an invalid computer name, not an unreachable one; dropped a literal probe-site count that would rot.

**Follow-up, deliberately not in this PR:** surfacing `AccountNeverProvisioned` on `ResultEvidence` so the server can see the tolerated path. The evidence value is discarded on that code path today, so it would need a separate change to be observable end to end.

## Deliberately not done

The issue's optional hardening — gating PAM lifecycle manager construction on `cfg.PAMEnabled` in `Heartbeat.SetStatePath` — is **out of scope for this PR**. It interacts with the reconciliation and `pamReceivedObservationReady` readiness contract and deserves its own change; this PR fixes the reported bug without touching that wiring.

Fixes #4587

https://claude.ai/code/session_01KbAcVYJNziZHQtGoGYph9N

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1320,11 +1320,26 @@ jobs:
       # sessionbroker. Note the #3007 bug itself was NOT a data race — sendSeq
       # was already atomic — so -race would not have caught it; the regression
       # test asserts receiver-visible ordering instead.
+      #
+      # ./internal/elevaccount is the fourth instance of the eventlog/logging
+      # gap, found by #4587: the whole package is netapi32 account lifecycle
+      # behind //go:build windows, and the only Windows-tagged test it had was
+      # a compile-time interface assertion that no job ran. The bug it shipped
+      # was precisely a netapi32 status code the Windows-only code mishandled
+      # (NERR_UserNotFound on an account that was never provisioned, which is
+      # the DEFAULT state of every agent that never enabled PAM), so a gate
+      # that never executes on Windows could not have caught it and cannot
+      # catch the next one. Verified the same way as logging and ipc: with
+      # GOOS=windows GOARCH=amd64 CGO_ENABLED=0 the package vets, its test
+      # binary compiles, and its Windows dependency graph contains no
+      # remote/desktop and no cgo. Its two behavioural tests assert against the
+      # real netapi32 and skip loudly, rather than pass, on a host where
+      # ~breeze_elev exists or local-account queries are not permitted.
       - name: Run Windows Go tests
         working-directory: agent
         env:
           CGO_ENABLED: "0"
-        run: go test ./internal/sessionbroker ./internal/eventlog ./internal/ipc ./internal/logging ./internal/state ./internal/watchdog ./internal/backup/vss ./cmd/breeze-watchdog
+        run: go test ./internal/sessionbroker ./internal/eventlog ./internal/elevaccount ./internal/ipc ./internal/logging ./internal/state ./internal/watchdog ./internal/backup/vss ./cmd/breeze-watchdog
 
       # Software inventory has platform-specific registry behavior. Keep this
       # filtered until the inherited Windows package backlog in #2523 is green,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1341,6 +1341,30 @@ jobs:
           CGO_ENABLED: "0"
         run: go test ./internal/sessionbroker ./internal/eventlog ./internal/elevaccount ./internal/ipc ./internal/logging ./internal/state ./internal/watchdog ./internal/backup/vss ./cmd/breeze-watchdog
 
+      # The step above runs without -v, so a t.Skip is indistinguishable from a
+      # pass in its output. That matters here more than usual: elevaccount's two
+      # behavioural tests are the ONLY live-syscall coverage anywhere of the
+      # #4587 fix, and they skip by design when ~breeze_elev already exists or
+      # when local-account queries are refused. A silent skip would leave the
+      # real netapi32 path — the actual blast radius of the bug — with zero
+      # coverage while CI stayed green, which is the same vacuous-gate failure
+      # the collectors and backup steps below already guard against. Assert both
+      # actually PASSED.
+      - name: Run Windows elevation-account absence tests (must not skip)
+        working-directory: agent
+        env:
+          CGO_ENABLED: "0"
+        shell: pwsh
+        run: |
+          $out = go test ./internal/elevaccount -run '^Test(VerifyClean|Deprovision)TreatsAbsentAccountAsClean$' -v 2>&1
+          $exit = $LASTEXITCODE
+          $out | ForEach-Object { Write-Host $_ }
+          if ($exit -ne 0) { throw "go test failed (exit $exit)" }
+          $passes = @($out | Select-String -Pattern '^--- PASS: Test').Count
+          $skips = @($out | Select-String -Pattern '^--- SKIP: Test').Count
+          if ($skips -ne 0) { throw "$skips of the absent-account tests SKIPPED - the runner could not reach the never-provisioned state they assert against, so the #4587 netapi32 fix has no live coverage here" }
+          if ($passes -ne 2) { throw "expected exactly 2 top-level PASS lines, got $passes - a test in the -run filter was renamed, removed, or added without updating this count" }
+
       # Software inventory has platform-specific registry behavior. Keep this
       # filtered until the inherited Windows package backlog in #2523 is green,
       # and count PASS lines so a rename cannot turn the gate vacuously green.
--- a/agent/internal/elevaccount/absent.go
+++ b/agent/internal/elevaccount/absent.go
@@ -2,6 +2,7 @@ package elevaccount
 
 import (
 	"errors"
+	"fmt"
 	"syscall"
 )
 
@@ -36,22 +37,50 @@ const (
 // agent re-ran the same cleanup and logged the same ERROR on every heartbeat.
 //
 // Only the two "no such account" statuses match. Every other status — access
-// denied, a locked SAM, an unreachable computer — stays a real error so
-// genuine cleanup failures keep failing closed. The status has to arrive as a
-// wrapped syscall.Errno, which is how the netapi32 and LSA wrappers report it;
-// it is never scraped out of the message text.
+// denied, a locked SAM, an invalid computer name — stays a real error so
+// genuine cleanup failures keep failing closed. The status has to arrive as,
+// or unwrap to, a syscall.Errno, which is how the netapi32 and LSA wrappers
+// report it; it is never scraped out of the message text.
+//
+// This is a classifier for ONE call's status, and callers must treat it that
+// way. It says nothing about which account the call was about, so it is only
+// meaningful applied to a call that names the elevation account. Aggregate
+// callers, which see a single error out of a sequence of syscalls, must use
+// errAccountAbsent instead — see cleanIfAbsent.
 func IsAccountAbsent(err error) bool {
 	return errors.Is(err, syscall.Errno(nerrUserNotFound)) ||
 		errors.Is(err, syscall.Errno(errorNoneMapped))
 }
 
+// errAccountAbsent marks an error as proof that the ELEVATION ACCOUNT itself
+// does not exist. Only a probe may attach it, and only from the one syscall in
+// that probe which names the account.
+//
+// The distinction is not pedantic. A probe runs more than one lookup, and the
+// others are about different principals entirely: accountInAdministrators and
+// localGroupMembersCall both resolve the builtin Administrators alias
+// (S-1-5-32-544) first, and that lookup can fail with the very same
+// ERROR_NONE_MAPPED. Matching the status code at the aggregate level would
+// read "the machine cannot resolve its own Administrators group" as "the
+// elevation account is absent, everything is clean" — inverting a security
+// check on the strength of an unrelated failure.
+var errAccountAbsent = errors.New("elevation account does not exist")
+
+// absentAccountErr tags a probe failure as proof of absence. Call it ONLY for
+// the status of a syscall that named the elevation account.
+func absentAccountErr(err error) error {
+	return fmt.Errorf("%w: %w", errAccountAbsent, err)
+}
+
 // cleanIfAbsent decides what a FAILED account probe means for the caller.
 //
-// When the failure means the account does not exist, that is positive proof of
-// a clean state rather than a cleanup failure, so it yields clean evidence and
-// no error. Every other failure is passed through untouched.
+// When the failure is tagged as the account being absent, that is positive
+// proof of a clean state rather than a cleanup failure, so it yields clean
+// evidence and no error. Every other failure is passed through untouched —
+// including a raw ERROR_NONE_MAPPED from some other principal's lookup, which
+// is exactly what this must not mistake for absence.
 //
-// All four probe sites in Deprovision and VerifyClean route through this one
+// Each probe site in Deprovision and VerifyClean routes through this one
 // function so the tolerance is defined — and tested — in a single place. Only
 // the first probe in each of those functions is reachable in the
 // never-provisioned case that motivated it, so per-site copies of this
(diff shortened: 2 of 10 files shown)


ToddHebebrand:
**Review runs:** two independent rounds. Round 1 — `code-reviewer` (0 findings), `silent-failure-hunter`, `pr-test-analyzer`. Round 2 — five lenses including `comment-analyzer`.

**Findings:** 3 raised in round 1 → all addressed in `47a4965` / `5464dc7`. 5 raised in round 2 → all addressed in `c085c5e` / `4f88e61`; 0 outstanding.

The round-2 HIGH was a defect **introduced by round 1's own fix**: the shared absence classifier was consulted at the aggregate level, but `accountInAdministrators` and `localGroupMembersCall` both resolve the builtin Administrators alias (`S-1-5-32-544`) before they touch the elevation account, and that lookup fails with the identical `ERROR_NONE_MAPPED`. A host that could not resolve its own Administrators group would have been reported as "elevation account absent, everything clean" — and in `VerifyClean` that also discarded an already-measured `enabled == true`. Absence is now carried by a sentinel that only the account-naming syscall may attach.

**Red-first proof** (each fix was reverted and the named test observed failing, then restored):

| Fix | Test that fails without it |
|---|---|
| absence tolerance | `TestSetEnabledFalseToleratesUnresolvableAccountNameInTokenScan` |
| sentinel vs. aggregate errno | `TestCleanIfAbsentIgnoresAbsenceStatusesFromOtherPrincipals` (`evidence {false false} and no error`) |
| alias-lookup passthrough | `TestEvidenceAfterAdminProbeNeverDiscardsAMeasuredEnabledAccount/alias_lookup_failure_propagates` |
| ledger load error | `TestSetEnabledFalseFailsClosedWhenTheLedgerCannotBeRead` (got `privileged_token_verification_failed`, want `ledger_unavailable`) |
| unmeasured token evidence | `TestToleratedTokenScanLeavesPrivilegedTokenEvidenceUnmeasured/skipped_scan` |

**Tests** (local, foreground, on `4f88e61`):

```
CGO_ENABLED=0 go test -count=1 ./internal/elevaccount/ ./internal/pamlifetime/ ./internal/heartbeat/
  ok internal/elevaccount  ok internal/pamlifetime  ok internal/heartbeat
  PASS counts incl. subtests: elevaccount 41, pamlifetime 103, heartbeat 1017
  1 SKIP, pre-existing: TestWindowsNetapiRoundTripManual (manual destructive VM test)
go test -race -count=1 ./internal/elevaccount/ ./internal/pamlifetime/   -> ok
go vet (linux + GOOS=windows) on both packages                          -> clean
go build ./... and GOOS=windows GOARCH=amd64 go build ./...             -> OK
```

**Windows coverage confirmed, not assumed.** Both reviewers flagged that the shared Windows step runs without `-v`, where a `t.Skip` is indistinguishable from a pass — and these two tests are the only live-netapi32 coverage of the fix. `5464dc7` adds a counted step that fails on a SKIP or on a PASS count other than two. It ran green on `windows-latest`, so the real-syscall tests genuinely execute there.

**Status:** review-clean, CI running on `4f88e61`, awaiting maintainer merge. Not merged and issue #4587 left open by design.

PR #4490, opened 2026-09-02
feat(api,shared): P2-5a1 — graduation evidence ledger, exactly-once evidence writers, intent-anchored fix watches (#4187)

ToddHebebrand:
## Summary

An agent earns narrower human supervision by evidence. Every terminal outcome an agent produces — a released action intent, an act-mode manifest execution, a fix-watch verdict, a 👍/👎 on an alert verdict — writes exactly ONE immutable row into a new `ai_agent_op_evidence` ledger keyed by `(source_kind, source_id, metric)`, so BullMQ redelivery can never double-count.

This PR (A1) ships the evidence substrate that PR A2's graduation state machine will read: the migration for `ai_agent_op_evidence` + `ai_agent_graduation` (RLS, grants, fix-watch column/constraint additions), the four registration lists (org cascade, export policy, org-merge, RLS auto-discovery), the canonical `tool:action` key resolver, the exactly-once evidence writer, intent-release terminalization evidence, intent-anchored fix watches with durable enqueue + recovery, watch-verdict and act-execution evidence, verdict-feedback evidence, an `updateAgent` row lock, and a 400-day evidence retention worker. No promotion/demotion logic and no UI ship in this PR — those are A2 and B.

## Scope calls / rulings

Verbatim `Ruling:` entries from `progress.md`:

- **Spec/plan amendment conflict**: rewrote the `Stats.`/`Promote suggestion.`/`Auto-demote.` paragraphs in spec §4.5 (lines 145-162) that still described the dropped `ai_agent_op_stats_daily` design, contradicting the amendment paragraph inserted immediately above them. Followed the doc's own "superseded by the amendment above/below" convention (precedent: §4.3's `Trigger.` paragraph) rather than a silent rewrite. Also extended the fix to the `Verdict precision.` paragraph (line 164) one line past the finding's stated range, which had the identical contradiction (referenced dropped `ai_agent_op_stats`) — leaving it would have defeated the finding's own stated purpose of §4.5 no longer containing two incompatible designs. No table/CHECK/FK/§5 text touched (already correct); no code files touched.

[...]

Issue #4187, opened 2026-08-28
AI agents phase 2 — intelligence layer (alert verdicts, sweeps, narrative, ticket triage, graduation, impact)


ToddHebebrand:
Tracking issue for **AI agents phase 2 — intelligence layer (alert verdicts, sweeps, narrative, ticket triage, graduation, impact)** (`ai-agents-p2`).

- Plan: `docs/superpowers/plans/ai-mcp/2026-08-28-ai-agents-p2-1-alert-verdicts.md`
- Spec: `docs/superpowers/specs/ai-mcp/2026-08-28-ai-agents-phase2-intelligence-layer-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All six waves merged (unreleased as of 2026-09-02):

- W01 P2-1 alert verdicts — #4220, #4222
- W02 P2-2 scheduled sweeps — #4232, #4233
- W03 P2-3 weekly org narrative — #4249, #4250
- W04 P2-4 ticket triage (act) — #4300, #4301
- W05 P2-5 feedback + graduation — #4490 (evidence ledger, intent-anchored fix watches), #4533 (partner-key ceiling, eligibility, four-eyes promote, auto-demote, graduation API), #4565 (graduation panel, "Approve and always allow", partner-ceiling hint, impact promote-eligible tile)
- W06 P2-6 value accounting — #4486 (rollup, weights, impact API), #4504 (impact page)

Deploy notes for the next release: new daily jobs (impact rollup `33 18`, evidence retention `48 18`, graduation evaluation); no new env vars; the impact rollup is gated by `BREEZE_AI_AGENTS_ENABLED`; promotion writes are gated by `BREEZE_AI_AGENTS_POLICY_DECIDE_ENABLED` (still dark in production); **behaviour change** — partner-level `supervisedActionKeys` are now a ceiling and no longer apply to an org without its own override row (each org must be granted, via graduation or a manual edit); the graduation read returns 404 for an org-only agent with no partner-wide baseline (by design).

Deferred to the roadmap: #4442 (act-mode auto-execution of sweep proposals), #4173, #4174, #4175, #4176, #4177, #4179, #4180, #4182, #4183, #4230, #4247, #4248. Follow-ups from the earlier waves filed as #4444–#4469; flake #4495 found during merges. Quorum rulings are on #4192 / #4193; per-task controller rulings are in each PR body.

[...]

PR #4595, opened 2026-09-02
fix(db): make the AI-agent graduation composite FKs deferrable; scope the test-side FK repair helper so CI can catch it (#4187 follow-up)

ToddHebebrand:
## Cause

`apps/api/migrations/2026-10-01-100000-ai-agents-graduation-evidence.sql` (`02716187d`, #4490) added three composite `(x, org_id)` foreign keys with no `DEFERRABLE` clause, so Postgres created them `NOT DEFERRABLE`:

| Constraint | Table | References | ON DELETE |
|---|---|---|---|
| `ai_agent_op_evidence_run_org_fk` | `ai_agent_op_evidence (run_id, org_id)` | `ai_agent_runs (id, org_id)` | `SET NULL (run_id)` |
| `ai_agent_graduation_intent_org_fk` | `ai_agent_graduation (promoted_intent_id, org_id)` | `action_intents (id, org_id)` | `SET NULL (promoted_intent_id)` |
| `ai_agent_fix_watches_intent_org_fk` | `ai_agent_fix_watches (intent_id, org_id)` | `action_intents (id, org_id)` | `CASCADE` |

The org-merge contract requires every composite FK referencing an `org_id` column to be `DEFERRABLE INITIALLY IMMEDIATE`: `services/orgMerge.ts` runs `SET CONSTRAINTS ALL DEFERRED` and re-points the parent's and the child's `org_id` in separate statements, which a non-deferrable composite FK aborts mid-merge. `2026-09-12-100001-org-lifecycle-foundations.sql` Section 2 swept the 19 FKs that existed when it shipped; these three were created after it and were never covered.

**Verified** on a fresh local stack (real Postgres 16, full `autoMigrate`), as the superuser and as `breeze_app` — identical results, so catalog visibility is not a factor:

```
              conname               | condeferrable | condeferred
------------------------------------+---------------+-------------
 ai_agent_fix_watches_intent_org_fk | f             | f
 ai_agent_graduation_intent_org_fk  | f             | f
 ai_agent_op_evidence_run_org_fk    | f             | f
```

### Why CI reported this green

`orgLifecycleFoundations.integration.test.ts:39` reads live `pg_constraint` state in a database shared with every other file in its shard. Three suites replay shipped migrations raw and then called `reapplyOrgIdFkDeferrability()` (`db-utils.ts`) to undo what the replay broke — but that helper ran the org-lifecycle migration's **whole-database sweep**, repairing every non-deferrable composite `org_id` FK in the database, not only the one its own replay damaged.

**Verified** from the CI log of run `33668113575`, Integration Tests shard 2 — same shard, same database, ordered by timestamp:

```
18:44:39  [auto-migrate] Applying: 2026-10-01-100000-ai-agents-graduation-evidence.sql   (applied cleanly; no swallowed error)
18:46:49  ✓ m365ConnectionsRls.integration.test.ts          → sweeps every non-deferrable composite org_id FK
18:47:39  ✓ agentIntentConstraints.integration.test.ts      → sweeps again
18:53:12  ✓ orgLifecycleFoundations.integration.test.ts (4 tests)   ← reads an already-repaired catalog
```

Every other candidate was ruled out: the migration applied with no `NOTICE`/`ERROR` around the `DO` blocks (nothing was swallowed by `EXCEPTION WHEN duplicate_object`); `DATABASE_URL` and `DATABASE_URL_APP` in `ci.yml` point at the same `localhost:5433/breeze_test`; `pg_constraint` returns identical rows to both roles; and the log names all four tests, including the deferrable one.

**Verified by direct local reproduction** (not inference): on a fresh stack with this migration absent and the *original* helper, running only those three replay suites flips all three constraints to `condeferrable = t`, after which the contract test passes 4/4 — CI's exact result.

## What changed

- **`apps/api/migrations/2026-10-04-100001-ai-agents-graduation-fks-deferrable.sql`** (new) — `ALTER CONSTRAINT ... DEFERRABLE INITIALLY IMMEDIATE` for exactly those three constraint names. Same statement the org-lifecycle sweep uses, so no `DROP` + re-`ADD`: no re-validation of existing rows, and no restating the column list, referenced table or `ON DELETE` action by hand. Idempotent — the loop selects only `condeferrable = false`. Reports a row count via `RAISE WARNING`, per the migration-authoring rules. Scoped to three names deliberately: a blanket sweep would rebuild the same blind spot at migration level.
- **`reapplyOrgIdFkDeferrability`** now takes a **required** constraint-name list and repairs only those, throwing when a named constraint does not exist so a rename fails loudly rather than silently leaving the contract un-restored. Each caller names what its own replay un-defers. **Verified** against a real database by replaying each migration and diffing the offender set — each un-defers exactly one constraint:
  - `m365ConnectionsRls` → `m365_consent_sessions_connection_identity_fkey`
  - `agentIntentConstraints` → `action_intents_requesting_agent_run_id_org_id_fkey`
  - `partnerApiReconstructionWatermark` → `devices_site_org_fk`
- **`orgLifecycleFoundations.integration.test.ts`** — comment recording why no test may repair schema state this assertion reads.

No table or column is added, so no cascade / export-policy registration changes are required; all three tables were already registered by #4490.

## How verified

Fresh per-worktree stack (`pnpm test-stack`, Postgres 16), torn down afterwards.

1. **Repro (fresh DB, `main` as-is):** contract RED, offenders exactly the three FKs above.
2. **Control (fresh DB, migration absent, original helper):** three replay suites → all three constraints `condeferrable = t` → contract **4/4 green**. Reproduces the false CI green locally.
3. **CI-side fix would have caught it (fresh DB, migration absent, scoped helper):** three replay suites 35/35 green and they restore only their own three constraints (`m365_consent_sessions_connection_identity_fkey`, `action_intents_requesting_agent_run_id_org_id_fkey`, `devices_site_org_fk` → `t`; the three AI-agent FKs stay `f`) → contract **RED**, naming all three.
4. **Fix (fresh DB, migration present):** `autoMigrate` applies it once — `WARNING: ai-agents graduation: made 3 composite org_id FKs deferrable` — contract **4/4 green**.
5. **Idempotency, twice on a fresh DB:** re-applying the file raw is a silent no-op; all three end `condeferrable = t, condeferred = f`. A second `autoMigrate` pass is a ledger skip and the contract stays green.
6. **Together:** the three replay suites + the contract in one run → **39/39 green**.
7. `tsc --noEmit` clean, `eslint` clean on all five touched TS files, `bash scripts/check-migration-naming.sh` OK (whole-dir and `--staged`), `vitest run src/db/autoMigrate.test.ts` **63/63**.

## Risk

Low. The migration touches only FK *deferrability* — no data, no schema shape, nothing modelled in Drizzle. `ALTER CONSTRAINT` takes a brief `ACCESS EXCLUSIVE` lock on the three tables but performs no table scan or row re-validation. `INITIALLY IMMEDIATE` means normal transactions still check these FKs at statement time exactly as before; only an explicit `SET CONSTRAINTS ... DEFERRED` (i.e. org merge) changes behaviour. The helper change is test-only, and its failure direction is safe: a future replay that damages an unlisted constraint makes the contract test go red rather than silently pass.

Refs #4187

https://claude.ai/code/session_015tHmJsqrTy5iHcqgpadtMp

### DIFF
--- a/apps/api/migrations/2026-10-04-100001-ai-agents-graduation-fks-deferrable.sql
+++ b/apps/api/migrations/2026-10-04-100001-ai-agents-graduation-fks-deferrable.sql
@@ -1,0 +1,59 @@
+-- Fix-forward for 2026-10-01-100000-ai-agents-graduation-evidence.sql (#4187 P2-5).
+--
+-- That migration added three composite (x, org_id) foreign keys without a
+-- DEFERRABLE clause, so Postgres created them NOT DEFERRABLE:
+--
+--   ai_agent_op_evidence.ai_agent_op_evidence_run_org_fk
+--   ai_agent_graduation.ai_agent_graduation_intent_org_fk
+--   ai_agent_fix_watches.ai_agent_fix_watches_intent_org_fk
+--
+-- The org-merge contract requires every composite FK that references an
+-- `org_id` column to be DEFERRABLE INITIALLY IMMEDIATE: orgMerge.ts runs
+-- `SET CONSTRAINTS ALL DEFERRED` and re-points the parent's and the child's
+-- `org_id` in separate statements, which a non-deferrable composite FK aborts
+-- mid-merge. 2026-09-12-100001-org-lifecycle-foundations.sql Section 2 swept
+-- the 19 FKs that existed when it shipped; these three were created after it
+-- and so were never covered.
+--
+-- The shipped file is content-hash immutable, so this fixes forward. The
+-- deferrability of an existing FK is changed in place with ALTER CONSTRAINT
+-- (the same statement Section 2 of the org-lifecycle migration uses) — no
+-- DROP + re-ADD, which would take the same lock but additionally re-validate
+-- every existing row and require the column list, referenced table and
+-- ON DELETE action to be restated by hand.
+--
+-- Scoped to these three constraint names ON PURPOSE. A blanket sweep would
+-- also silently repair any future non-deferrable composite org_id FK that
+-- happened to be in the database when it ran, which is precisely the failure
+-- this migration exists to correct: the contract test must stay the thing that
+-- catches a new offender.
+--
+-- Idempotent: the loop selects only `condeferrable = false`, so a second run
+-- matches nothing and does nothing.
+
+DO $$
+DECLARE
+  fk record;
+  n integer := 0;
+BEGIN
+  FOR fk IN
+    SELECT con.conname, con.conrelid::regclass AS child_table
+    FROM pg_constraint con
+    WHERE con.contype = 'f'
+      AND con.condeferrable = false
+      AND con.connamespace = 'public'::regnamespace
+      AND con.conname IN (
+        'ai_agent_op_evidence_run_org_fk',
+        'ai_agent_graduation_intent_org_fk',
+        'ai_agent_fix_watches_intent_org_fk'
+      )
+    ORDER BY con.conrelid::regclass::text, con.conname
+  LOOP
+    EXECUTE format('ALTER TABLE %s ALTER CONSTRAINT %I DEFERRABLE INITIALLY IMMEDIATE',
+                   fk.child_table, fk.conname);
+    n := n + 1;
+  END LOOP;
+  IF n > 0 THEN
+    RAISE WARNING 'ai-agents graduation: made % composite org_id FKs deferrable', n;
+  END IF;
+END $$;
--- a/apps/api/migrations/2026-10-04-100001-ai-agents-graduation-fks-deferrable.sql
+++ b/apps/api/migrations/2026-10-04-100001-ai-agents-graduation-fks-deferrable.sql
@@ -12,8 +12,8 @@
 -- `SET CONSTRAINTS ALL DEFERRED` and re-points the parent's and the child's
 -- `org_id` in separate statements, which a non-deferrable composite FK aborts
 -- mid-merge. 2026-09-12-100001-org-lifecycle-foundations.sql Section 2 swept
--- the 19 FKs that existed when it shipped; these three were created after it
--- and so were never covered.
+-- the composite org_id FKs that existed when it shipped; these three were
+-- created after it and so were never covered.
 --
 -- The shipped file is content-hash immutable, so this fixes forward. The
 -- deferrability of an existing FK is changed in place with ALTER CONSTRAINT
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
Review follow-ups applied in `b313d2c36`.

**1. The helper contract is now executable, not prose.** New suite `apps/api/src/__tests__/integration/orgIdFkDeferrabilityHelper.integration.test.ts` asserts against real Postgres that an empty name list rejects, an unknown name rejects, and that repairing one named constraint leaves an unrelated non-deferrable composite `org_id` FK untouched — the property that keeps `orgLifecycleFoundations` able to catch a genuinely broken FK. It builds its own probe tables and drops them in `afterAll`, so no shipped constraint is mutated and a failed run cannot leak a non-deferrable `org_id` FK into the contract test's or rls-coverage's view of the schema (verified: 0 rows matching `fk_deferrability_probe%` in `information_schema` after the run).

**2. The initial mode is read from the catalog instead of hard-coded.** The helper forced `INITIALLY IMMEDIATE`; several later migrations declare a composite `org_id` FK `DEFERRABLE INITIALLY DEFERRED` on purpose (`2026-09-13-agent-rollback-lifecycle.sql`, `2026-09-28-100002-software-inventory-observations.sql`), so a future caller naming one would have silently downgraded it and changed when Postgres checks that FK for the rest of the shard. It now reads `pg_constraint.condeferred` and restores the same mode. A constraint a replay knocked to `NOT DEFERRABLE` also has `condeferred = false`, so those still come back `INITIALLY IMMEDIATE` exactly as the org-lifecycle sweep intends — the three existing callers are unchanged.

Red first, against the previous helper:

```
× preserves INITIALLY DEFERRED instead of downgrading it to IMMEDIATE
AssertionError: expected { deferrable: true, deferred: false }
             to deeply equal { deferrable: true, deferred: true }
Tests  1 failed | 3 passed (4)
```

The other three cases pin behaviour this PR had already introduced, so they were green from the start — stating that plainly rather than implying all four were red.

**3. Dropped the disputed constraint count.** The migration header said the org-lifecycle sweep converted "19 FKs". Its own header (`2026-09-12-100001-org-lifecycle-foundations.sql:16-17`) and `partnerApiReconstructionWatermark.integration.test.ts:53-54` both say 16; 19 was what I observed in the 2026-09-02 CI replay, several migrations later. The number is derived and drifts, so it is gone rather than restated.

**Re-verified** on a fresh per-worktree stack (real Postgres 16), torn down afterwards: the migration applies once via `autoMigrate` (`WARNING: ai-agents graduation: made 3 composite org_id FKs deferrable`), and the five touched suites — the new helper contract, the three migration-replay suites, and `orgLifecycleFoundations` — run **43/43 green** together. All three AI-agent FKs end `condeferrable = t / condeferred = f`. `tsc` (exit 0), `eslint`, `check-migration-naming.sh` (whole-dir and `--staged`) and `autoMigrate.test.ts` (**63/63**) all pass.

PR #4634, opened 2026-09-02
fix(web): refetch the mounted approximate-total line on a partner currency reset (#4472)

### LINKED ISSUE #4472: [Web][FX] Mounted approximate-total line never refetches after a partner reporting-currency change — cache generation is not in the hook's effect deps
**Verified (pre-existing; surfaced while fixing #4415 in PR #4440).** `apps/web/src/lib/useApproximateTotal.ts:191` — the effect's dependency list omits the cache generation, so when `resetPartnerCurrencyCache()` bumps it (same-tab partner reporting-currency save in `PartnerBillingSettings.tsx:158-160`), an already-mounted line keeps its existing state — stale total included — and only a remount or a query-key change refreshes it. That is exactly what the cache module's own comment claims to prevent. Compounding: the cache key (`:156`) doesn't encode generation either, so a request discarded by the generation guard (`:142`) can leave a mounted line stuck on "could not load exchange rates" without self-healing.

**Reachability:** low — rollups are normally unmounted while the user is on billing settings — but #4440 makes one corner of it newly *visible*, so it will read as a bug when hit.

**Fix shape:** include the generation in the effect deps (or subscribe to the cache's reset), and fold generation into the cache key so a discarded in-flight response can't pin a stale status.

Reported by: internal review follow-up 2026-09-01 (PR #4440), #3772 follow-up.

ToddHebebrand:
## Summary

- The approximate-total cache's generation was only synced **lazily** inside `approximateTotalCacheKey()`, which runs during render — so an already-mounted `useApproximateTotal` line with no other reason to re-render never noticed `resetPartnerCurrencyCache()` (a same-tab partner reporting-currency save in `PartnerBillingSettings.tsx`) and kept its stale total. Every existing test happened to unmount + remount around the reset, which sidesteps this: a fresh mount always computes a fresh key.
- The cache key also didn't encode generation, so a stale in-flight response discarded by the existing generation guard (`loadApproximateTotal`) could leave a mounted line pinned on "could not load exchange rates" with nothing to self-heal it.

## Fix

- `apps/web/src/lib/partnerCurrencyCache.ts`: add `subscribePartnerCurrencyCache()` so a reset notifies listeners synchronously the instant it happens, instead of only being observable lazily on the next unrelated render.
- `apps/web/src/lib/approximateTotalCache.ts`: subscribe to that eagerly at module load (clearing this cache and bumping its own generation the moment the partner-currency cache resets), fold the generation into `approximateTotalCacheKey()`'s return value, and expose `subscribeApproximateTotalCache()` for consumers. The old lazy check is kept as an idempotent fallback.
- `apps/web/src/lib/useApproximateTotal.ts`: use `useSyncExternalStore` on the new subscription so a mounted line force-renders on a reset (a module-level mutation React otherwise has no way to notice), and list the generation explicitly in the effect's own dependency array alongside the now generation-scoped `key`.

## Test plan

- [x] New tests in `useApproximateTotal.test.tsx` cover the case none of the existing tests did: a **single mounted instance, never unmounted**, that must refetch in place after `resetPartnerCurrencyCache()`, self-heal from a stale in-flight response discarded by the generation guard, and survive a plain `resetApproximateTotalCache()` (logout) too.
- [x] `cd apps/web && npx vitest run src/lib/__tests__/useApproximateTotal.test.tsx` — 29 passed
- [x] `cd apps/web && npx vitest run src/lib/__tests__/usePartnerCurrency.test.tsx src/components/billing/PartnerBillingSettings.test.tsx` — 26 passed
- [x] `cd apps/web && npx tsc --noEmit` — clean
- [x] `cd apps/web && npx eslint` on the four changed files — clean

Closes #4472

### DIFF
--- a/apps/web/src/lib/__tests__/useApproximateTotal.test.tsx
+++ b/apps/web/src/lib/__tests__/useApproximateTotal.test.tsx
@@ -368,3 +368,64 @@ describe('useApproximateTotal — a partner reporting-currency change invalidate
     expect(fetchWithAuth).toHaveBeenCalledTimes(1);
   });
 });
+
+// ---------------------------------------------------------------------------
+// #4472: the cache generation was not in the effect's dependency list, so an
+// already-mounted line — never unmounted, no other prop/state change — kept
+// its stale state after `resetPartnerCurrencyCache()` and only picked up the
+// new currency on a later remount. Every test above unmounts and remounts
+// (`first.unmount()` then a fresh `render()`), which happens to work because
+// a brand-new mount always computes a fresh key — it never exercised the
+// actually-broken path. These do.
+// ---------------------------------------------------------------------------
+describe('useApproximateTotal — an ALREADY-MOUNTED line reacts to a reset in place (#4472)', () => {
+  it('refetches without unmounting when a same-tab reporting-currency save resets the cache', async () => {
+    fetchWithAuth.mockResolvedValueOnce(jsonRes({ data: AVAILABLE }));
+    render(<Probe id="a" date="2026-08-21" />);
+    await waitFor(() => expect(screen.getByTestId('a').textContent).toBe('available:22940.00:CAD'));
+
+    // What PartnerBillingSettings.tsx does after a reporting-currency save —
+    // the mounted Probe is never unmounted or given new props.
+    fetchWithAuth.mockResolvedValueOnce(
+      jsonRes({ data: { ...AVAILABLE, targetCurrencyCode: 'EUR', total: '17000.00' } }),
+    );
+    resetPartnerCurrencyCache();
+
+    await waitFor(() => expect(screen.getByTestId('a').textContent).toBe('available:17000.00:EUR'));
+    expect(fetchWithAuth).toHaveBeenCalledTimes(2);
+  });
+
+  it('a stale response discarded by the generation guard self-heals instead of pinning `failed`', async () => {
+    let resolveFirst: (r: Response) => void = () => {};
+    fetchWithAuth.mockReturnValueOnce(new Promise<Response>((r) => { resolveFirst = r; }));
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer, silent-failure-hunter, pr-test-analyzer (all sonnet)
**Findings:** 3 Important raised (0 Critical) → all 3 addressed in 79651c7f0; 0 outstanding.
- silent-failure-hunter: race between the `cancelled` closure flag and the generation-discard guard in the effect's `.then()` — not structurally guaranteed, only true by timing. Fixed by checking the started generation directly at the callback.
- silent-failure-hunter: unguarded listener `forEach` in both cache modules — one throwing listener would abort the loop and silently skip later subscribers. Fixed with try/catch + `Sentry.captureException` per listener.
- pr-test-analyzer (6/10): no test covered the fix's actual fan-out mechanism (a shared listener `Set` feeding multiple mounted hook instances). Added a two-simultaneously-mounted-lines-different-keys test.
- code-reviewer: 0 findings — verified the 3 downstream consumer suites (`moneyRollupApproximate.test.tsx`, `ApproximateMoneyLine.test.tsx`, `PartnerBillingSettings.test.tsx`) pass unmodified, no key-format regression.
- pr-test-analyzer also empirically confirmed (applying only the new test file to pre-fix source) that the 3 original regression tests fail on the old code and pass on the fix — genuine red-first coverage, not vacuous.

**Tests:** `apps/web` — `useApproximateTotal.test.tsx` (30), `usePartnerCurrency.test.tsx` + `PartnerBillingSettings.test.tsx` (26), `ApproximateMoneyLine.test.tsx` + `moneyRollupApproximate.test.tsx` (54) — 110 total, all passing. `npx tsc --noEmit` clean. `npx eslint` on all 4 changed files clean.

PR #4635, opened 2026-09-02
fix(ops): stop passing DATABASE_URL on pg_dump/pg_restore/psql argv (#4497)

### LINKED ISSUE #4497: [Ops] scripts/backup.sh passes DATABASE_URL (with password) on pg_dump's argv — visible in ps for the whole dump
**Verified on a prod droplet during the v0.109.0 rollout.** `scripts/backup.sh:151` runs `pg_dump "${DATABASE_URL}" -Fc …`, so for the full duration of the dump (30+ min on US) the managed-Postgres admin URL, password included, is readable by any process that can list processes (`ps -o args`). Root-only host today, but it also leaks into anyone's terminal who inspects the box mid-backup (it did into mine).

**Why not a one-liner:** libpq only expands a connection URI when it's the positional/`--dbname` argument — `PGDATABASE=postgresql://…` is *not* parsed (confirmed: pg_dump fell back to the local socket). The fix is to split the URL into `PGHOST/PGPORT/PGUSER/PGPASSWORD/PGDATABASE` (+ `PGSSLMODE` from the query string) in the script and call `pg_dump -Fc -Z 6 -f "$dest"` with no argv secret — or write a mode-600 temp `PGPASSFILE` and pass a password-stripped URL. Same treatment for any `psql "$DATABASE_URL"` calls in `/opt/breeze-backup/*` (restore-test) and the release runbook's promote step.

Reported by: v0.109.0 rollout, 2026-09-02.

ToddHebebrand:
## Summary

`scripts/backup.sh` and `scripts/restore.sh` passed `DATABASE_URL` (with the
password embedded) as a positional argument to `pg_dump` / `pg_restore` /
`psql`. That argument stays visible to any local user via `ps -o args` for
the whole duration of the dump/restore — 30+ min on the US droplet, per the
issue's prod verification.

- Adds `scripts/lib/pg-connect-env.sh`: `pg_url_to_env` splits a
  `postgres://`/`postgresql://` URL into `PGHOST`/`PGPORT`/`PGUSER`/
  `PGPASSWORD`/`PGDATABASE` (+ `PGSSLMODE` from the query string), handling
  percent-decoded user/password/dbname. `pg_url_unset_env` clears them
  afterward.
- `backup.sh`'s `backup_database()` now calls `pg_dump -Fc -Z 6 -f "$dest"`
  with no connection string on argv (dbname resolves from `PGDATABASE`).
- `restore.sh`'s `restore_database()` now calls `pg_restore --clean
  --if-exists -d "${PGDATABASE}" "$DB_FILE"` and `psql -d "${PGDATABASE}" ...`
  — `-d` still needs a value to actually connect/restore (an omitted `-d`
  makes `pg_restore` write to stdout instead), but it's just the plain
  database name now, not a secret; host/user/password resolve from the
  environment.
- Existing behavior is otherwise unchanged (same flags, same exit-code
  handling, same log messages).

## Why not restore-test.sh / the release runbook

`scripts/ops/restore-test.sh` already passes credentials via `-e
PGPASSWORD=...` to its `docker exec psql` call and only sets `DATABASE_URL`
as an env var when invoking `restore.sh` (never on its own argv), so it
inherits the fix transitively and needed no direct change. The release
runbook's promote step and `/opt/breeze-backup/*` mentioned in the issue live
outside this repo (droplet-side / gitignored `internal/`) — flagging that as
a manual follow-up rather than guessing at their content.

## Test plan

- [x] `node --test scripts/lib/pg-connect-env.test.mjs` — 7/7 pass: URL
      parsing (full URL w/ encoded password + sslmode, no-password, no-user,
      malformed, missing-dbname) + a static regression guard on both scripts
      against reintroducing `DATABASE_URL` on `pg_dump`/`pg_restore`/`psql`
      argv.
- [x] Confirmed the regression guard fails against the pre-fix
      `scripts/backup.sh` (checked out from `origin/main`) before it passes
      against the fixed version.
- [x] `bash -n` on all three changed/added scripts.
- [x] `shellcheck` clean (only an info-level SC1091 for the `source`, not a
      CI-gated check).
- [x] End-to-end functional test against a fake `pg_dump`/`pg_restore`/`psql`
      on `PATH`: confirmed `PGPASSWORD` reaches them via environment while
      their recorded argv contains no password/host/user — only flags and
      (for restore) the plain database name.
- [x] `scripts/security/scan-confidential.sh --all` — clean.
- [x] Wired the new test into CI as `pnpm test:pg-connect-env`
      (`security-audit` job), following the existing `scripts/*.test.mjs` +
      root `package.json` alias convention (`test:release-lineage`,
      `test:agent-binary-signatures`, etc).

Closes #4497

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -829,6 +829,14 @@ jobs:
       - name: Run agent mTLS edge policy guard
         run: pnpm check:agent-mtls-edge-policy
 
+      # #4497: scripts/backup.sh and scripts/restore.sh used to pass
+      # DATABASE_URL (password included) directly on pg_dump's/pg_restore's/
+      # psql's argv, visible to any local user via `ps` for the whole
+      # dump/restore. Covers the PG* URL-splitting helper's parsing and
+      # statically guards both scripts against reintroducing the pattern.
+      - name: Test pg connection env helper (#4497)
+        run: pnpm test:pg-connect-env
+
   build-api:
     name: Build API
     runs-on: ubuntu-latest
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "check:agent-mtls-edge-policy": "bash scripts/check-agent-mtls-edge-policy.sh",
     "check:agent-binary-signatures": "bash scripts/security/check-agent-binary-signatures.sh --source",
     "test:agent-binary-signatures": "node --test scripts/security/check-agent-binary-signatures.test.mjs",
+    "test:pg-connect-env": "node --test scripts/lib/pg-connect-env.test.mjs",
     "db:migrate": "turbo run db:migrate --filter=@breeze/api",
     "db:check-drift": "turbo run db:check-drift --filter=@breeze/api",
     "db:seed": "turbo run db:seed --filter=@breeze/api",
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -27,6 +27,10 @@
 
 set -euo pipefail
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=lib/pg-connect-env.sh
+source "${SCRIPT_DIR}/lib/pg-connect-env.sh"
+
 # ---------------------------------------------------------------------------
 # Configuration
 # ---------------------------------------------------------------------------
@@ -148,12 +152,22 @@ backup_database() {
   log "Starting database backup..."
   local dest="${BACKUP_DIR}/db_${TIMESTAMP}.dump"
 
-  if pg_dump "${DATABASE_URL}" -Fc -Z 6 -f "${dest}" 2>&1; then
+  # Split DATABASE_URL into PG* environment variables rather than passing it
+  # as an argument — pg_dump's argv is visible to any local user via `ps` for
+  # the whole duration of the dump (#4497).
+  if ! pg_url_to_env "${DATABASE_URL}"; then
+    log "ERROR: Database backup failed"
+    return 1
+  fi
+
+  if pg_dump -Fc -Z 6 -f "${dest}" 2>&1; then
+    pg_url_unset_env
     local size
     size=$(du -h "${dest}" | cut -f1)
     log "Database backup complete: ${dest} (${size})"
     TASKS_SUCCEEDED=$((TASKS_SUCCEEDED + 1))
   else
+    pg_url_unset_env
     log "ERROR: Database backup failed"
     rm -f "${dest}"
     return 1
--- a/scripts/lib/pg-connect-env.sh
+++ b/scripts/lib/pg-connect-env.sh
@@ -111,12 +111,17 @@ pg_url_to_env() {
     return 1
   fi
 
-  local user_raw="${BASH_REMATCH[3]}"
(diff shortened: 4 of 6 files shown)


ToddHebebrand:
**Update:** pushed a follow-up commit (\`df605bf10\`) hardening \`scripts/lib/pg-connect-env.sh\` in response to the first review round (code-reviewer + silent-failure-hunter, run in parallel):

- Bracketed IPv6 host support (\`[::1]\`) — the original host regex excluded \`:\`, so any IPv6 \`DATABASE_URL\` failed to parse at all (a real regression vs. the old code, which just handed the URL straight to libpq).
- Percent-encoding is now validated (exactly two hex digits after each \`%\`) before decoding; a malformed escape now fails loudly (\`return 1\`) instead of potentially producing a silently-corrupted \`PGPASSWORD\`/\`PGUSER\`/\`PGDATABASE\` value while still reporting success.
- Everything is decoded/validated before anything is exported, so a rejected URL never leaves a partial PG* environment behind.
- Any query-string parameter besides \`sslmode\` now gets a stderr warning instead of being silently dropped (the original code handed the whole URL, query string included, to libpq — this fix only translates \`sslmode\`, so anything else is a real behavioral narrowing that's now at least visible).
- Fixed two bash-3.2-incompatible constructs the hardening itself introduced (negative-length substring expansion, unguarded empty-array expansion under \`set -u\`).

\`scripts/lib/pg-connect-env.test.mjs\` expanded from 7 to 13 tests covering all of the above; full suite still green, functional end-to-end test against fake \`pg_dump\`/\`pg_restore\`/\`psql\` binaries re-verified (no secret on argv).

ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter, pr-test-analyzer — all model: sonnet, run in parallel; pr-test-analyzer re-run once against the final commit after its first run didn't return)
**Findings:** 3 real gaps raised in round 1 (both code-reviewer and silent-failure-hunter converged on the same set): bracketed-IPv6 hosts unparseable, malformed percent-encoding silently corrupting a decoded value without erroring, and non-`sslmode` query params (`connect_timeout`, `sslrootcert`, etc.) silently dropped. All 3 fixed in `df605bf10` (which also fixed two bash-3.2-incompatible constructs the fix itself introduced — negative-length substring expansion and an unguarded empty-array expansion under `set -u`). Round 2 (pr-test-analyzer against the hardened code) found 0 bugs, rated coverage adequate to merge, and suggested 4 non-blocking test additions — added inline in `6125fad2d` rather than deferred.
**Tests:** `node --test scripts/lib/pg-connect-env.test.mjs` — 18/18 passing (URL-parsing correctness incl. IPv6, percent-decoding on user/password/dbname, malformed-encoding rejection, sslmode + extra-query-param warning, static regression guard on both scripts). `bash -n` clean on all three shell files; `shellcheck` clean (info-level SC1091 only, not CI-gated). Functional end-to-end test against fake `pg_dump`/`pg_restore`/`psql` binaries confirms `PGPASSWORD`/`PGHOST`/`PGUSER` reach them via environment while their recorded argv contains no secret. `scripts/security/scan-confidential.sh --all` clean. No TypeScript touched (pure shell/CI/package.json change), so no `tsc` run.

PR #4638, opened 2026-09-02
fix(api): apply the new-partner inbound default on every creation path (#4520)

### LINKED ISSUE #4520: [API] Platform-admin POST /partners bypasses createPartner() and misses the inbound-email default
**Problem:** PR #4420 (#3608) made `createPartner()` in `apps/api/src/services/partnerCreate.ts` write an explicit `settings.ticketing.inbound.enabled = false` for new partners. The platform-admin creation path at `apps/api/src/routes/orgs.ts:474` (`POST /partners`, `requireScope('system')`) inserts partners directly and does not go through `createPartner()`, so partners created that way still get `{}` settings and fall back to the readers' absent-means-enabled behavior.

**Fix:** route the admin endpoint through `createPartner()` (or extract the default-settings builder and call it from both), plus a test on the admin route asserting the default lands.

Related: #3608, PR #4420.

ToddHebebrand:
## Problem

PR #4420 (#3608) made new partners opt **in** to inbound email-to-ticket by writing an explicit `settings.ticketing.inbound.enabled = false` at creation — but the default lived **inline** in `createPartner()`.

The platform-admin `POST /orgs/partners` handler (`requireScope('system')`, `apps/api/src/routes/orgs.ts`) inserts partners directly and never calls `createPartner()`. Partners minted that way still landed with `{}` settings and fell through to the readers' `enabled !== false` fallback — i.e. inbound email-to-ticket silently **on**, the exact behaviour #3608 set out to stop.

## Fix

Routing the admin endpoint through `createPartner()` isn't viable — that function also mints an admin user, role, permission set, default org and default site from a required `adminEmail`/`adminName`/`passwordHash`, none of which the admin endpoint has or wants. So this takes the issue's second option: **extract the default-settings builder**.

New dependency-free module `apps/api/src/services/partnerDefaultSettings.ts` exporting `applyNewPartnerDefaultSettings()`, called from **all three** partner-insert sites (`grep -rn 'insert(partners)' apps ee packages --include='*.ts'` — these are the only three outside tests; no raw-SQL partner inserts in migrations):

| Site | Before | After |
|---|---|---|
| `services/partnerCreate.ts` | inline literal | calls the builder |
| `routes/orgs.ts` — admin `POST /partners` | **missing** (the bug) | calls the builder |
| `db/seed.ts` — dev "Default Partner" | **missing** | calls the builder |

The seed is included so local dev matches production behaviour rather than quietly testing a config prod never has.

### Merge semantics

The admin route accepts caller-supplied `settings` (`z.any()`), so the builder **merges** instead of overwriting:

- An explicit `ticketing.inbound.enabled` from the caller — `true` **or** `false` — is preserved. The default only fills the flag in where it is absent.
- Unrelated keys, and siblings under `ticketing` / `ticketing.inbound`, pass through untouched.
- The caller's object is never mutated; a fresh object is returned. The copy depth (three levels) matches exactly the three levels written.
- Values that *cannot* hold the flag (a non-object at `settings`, `ticketing`, or `inbound`) are normalized rather than preserved. The readers traverse that path and treat anything untraversable as absent → enabled, so preserving a malformed branch would fail **open**. Nothing usable is lost — no reader can interpret those shapes — and the handler's `.returning(partnerPublicColumns())` projects the persisted row (settings included) into the 201 body, so a caller sees exactly what landed. That echo is now asserted end-to-end by a route-level test, not just argued.

Ordering with the existing `foldAllowedMfaMethodsAlias` canonicalization is preserved and covered by a test.

The readers (`loadPartnerInboundPolicy` in `inboundEmail/resolveOrg.ts`, `getTicketConfig` in `ticketConfigService.ts`) are **unchanged** and still serve as the pre-#3606 upgrade path. No backfill of existing partners' settings — same scope decision as #4420.

## Test plan

Red-first: the new admin-route assertions and the helper suite were confirmed failing against the unfixed handler before implementing. The added normalization/echo test was separately re-verified by reverting the handler call and watching it fail.

- [x] `vitest run src/routes/orgs.test.ts src/routes/orgs.listQuery.test.ts src/services/partnerDefaultSettings.test.ts src/services/partnerCreate.test.ts` — **304/304 passed**. 9 new helper cases (absent/null/non-object settings, sibling preservation at both levels, explicit true and false, no-mutation, malformed-branch normalization) + 5 new admin-route cases (default lands with no settings; merges alongside caller settings; respects explicit `enabled: true`; still folds the MFA alias; normalizes a non-object value and echoes it in the 201 body).
- [x] `vitest run src/services/ticketConfigService.test.ts src/services/inboundEmail` — 197/197 passed (the readers this decision deliberately leaves unchanged).
- [x] `vitest run src/routes/auth/register.test.ts src/__tests__/partner-wide-write-coverage.test.ts` — 33/33 passed.
- [x] `npx tsc --noEmit -p apps/api` — clean.
- [x] `npx eslint` on every touched file — clean.

No schema or migration changes, so no RLS/cascade/export-policy registration applies.

**Coverage caveat:** the `db/seed.ts` call site has no direct test — it is a local dev-seed path whose only obligation is calling the builder with no arguments, which the helper suite already characterizes fully. Verified by inspection and typecheck, not by a test.

## Reviewed

Three independent reviews (code-reviewer, silent-failure-hunter, pr-test-analyzer): **0 MUST-FIX**. The test-coverage finding above was addressed in 9d9d702. Two suggestions were consciously declined:

- **Narrow `settings: z.any()` to a record schema so malformed input 400s instead of normalizing.** Reasonable, but it tightens a public API's accepted input domain beyond this issue, and `z.any()` for `settings` is systemic in this file (`createPartnerSchema`, `updatePartnerSchema`, and the org schemas) — changing only the create path would leave the update path inconsistent. Wants its own change that sweeps all of them.
- **Log when a malformed `settings` value is discarded.** The 201 body already carries the ground truth, the branch is only reachable by an MFA-gated platform admin, and `partnerDefaultSettings.ts` is deliberately import-free so any creation path (including the seed and suites that mock `../db/schema`) can use it.

## Noted, deliberately not in scope

`PATCH /orgs/partners/:id` writes `settings` **wholesale**. A platform-admin update that omits `ticketing` therefore drops the opt-out flag and re-enables inbound at the readers — the same fail-open class as the `ipAllowlist` case that `preserveIpAllowlistOnOmit` already guards on that handler. Whether an admin clearing settings *intends* to re-enable inbound is a product call, so it wants its own decision rather than a guess bundled here.

Closes #4520

### DIFF
--- a/apps/api/src/db/seed.ts
+++ b/apps/api/src/db/seed.ts
@@ -3,6 +3,7 @@
 import '../config/normalizeNodeEnv';
 import { db, withSystemDbAccessContext } from './index';
 import { roles, permissions, rolePermissions, scripts, alertTemplates, partners, organizations, sites, users, partnerUsers } from './schema';
+import { applyNewPartnerDefaultSettings } from '../services/partnerDefaultSettings';
 import { seedSystemTicketStatuses } from '../services/ticketConfigService';
 import { eq, and } from 'drizzle-orm';
 import { hashPassword } from '../services/password';
@@ -1091,7 +1092,10 @@ export async function seedDefaultAdmin() {
           name: 'Default Partner',
           slug: 'default-partner',
           type: 'msp',
-          plan: 'enterprise'
+          plan: 'enterprise',
+          // #4520: keep the seeded dev partner on the same inbound opt-out
+          // default real partners get, so local behaviour matches production.
+          settings: applyNewPartnerDefaultSettings()
         })
         .returning();
       await seedSystemTicketStatuses(tx, newPartner!.id);
--- a/apps/api/src/routes/orgs.test.ts
+++ b/apps/api/src/routes/orgs.test.ts
@@ -469,6 +469,107 @@ describe('org routes', () => {
         'partner-1'
       );
     });
+
+    // Issue #4520: this platform-admin path inserts partners directly instead of
+    // going through createPartner(), so it used to miss the #3608 opt-out default
+    // and fall back to the readers' absent-means-enabled behaviour.
+    describe('new-partner default settings (#4520)', () => {
+      const captureInsertedValues = () => {
+        const captured: Record<string, unknown>[] = [];
+        vi.mocked(db.select).mockReturnValue({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({
+              limit: vi.fn().mockResolvedValue([])
+            })
+          })
+        } as any);
+        vi.mocked(db.transaction).mockImplementation(async (fn: (tx: any) => any) => {
+          const tx = {
+            insert: vi.fn(() => ({
+              values: vi.fn((vals: Record<string, unknown>) => {
+                captured.push(vals);
+                return { returning: vi.fn().mockResolvedValue([{ id: 'partner-1' }]) };
+              })
+            }))
+          };
+          return fn(tx);
+        });
+        return captured;
+      };
+
+      it('writes settings.ticketing.inbound.enabled=false when the caller sends no settings', async () => {
+        const captured = captureInsertedValues();
+
+        const res = await app.request('/orgs/partners', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ name: 'Partner', slug: 'partner' })
+        });
+
+        expect(res.status).toBe(201);
+        expect(captured[0]?.settings).toEqual({ ticketing: { inbound: { enabled: false } } });
+      });
+
+      it('adds the default alongside caller-supplied settings without clobbering them', async () => {
+        const captured = captureInsertedValues();
+
+        const res = await app.request('/orgs/partners', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({
+            name: 'Partner',
+            slug: 'partner',
+            settings: {
+              security: { ipAllowlist: ['10.0.0.0/8'] },
+              ticketing: { inbound: { unknownSenderMode: 'triage' } }
+            }
+          })
+        });
+
+        expect(res.status).toBe(201);
+        expect(captured[0]?.settings).toEqual({
+          security: { ipAllowlist: ['10.0.0.0/8'] },
+          ticketing: { inbound: { unknownSenderMode: 'triage', enabled: false } }
+        });
+      });
+
+      it('respects an explicit ticketing.inbound.enabled=true from the caller', async () => {
+        const captured = captureInsertedValues();
+
+        const res = await app.request('/orgs/partners', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({
+            name: 'Partner',
+            slug: 'partner',
+            settings: { ticketing: { inbound: { enabled: true } } }
+          })
+        });
+
+        expect(res.status).toBe(201);
+        expect(captured[0]?.settings).toEqual({ ticketing: { inbound: { enabled: true } } });
+      });
+
+      it('still folds the legacy allowedMfaMethods alias while applying the default', async () => {
+        const captured = captureInsertedValues();
+
+        const res = await app.request('/orgs/partners', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three independent reviewers on 3556157 (code-reviewer, silent-failure-hunter, pr-test-analyzer).

**Findings:** **0 MUST-FIX** from all three. 1 test-coverage finding addressed in 9d9d702; 2 suggestions consciously declined (rationale in the PR body).

- **code-reviewer — 0 findings.** Independently verified the call-site sweep (3/3 production `insert(partners)` sites, no raw-SQL partner inserts in migrations), that the builder's copy depth matches the three levels it writes so the caller's object is never mutated, that `{...spread}` over `JSON.parse` output carries no prototype-pollution hazard (define semantics, not the `__proto__` setter), that the new call is scoped to `POST /partners` only and runs before the insert with nothing downstream re-reading a pre-default `data.settings`, and that both readers (`loadPartnerInboundPolicy`, `getTicketConfig`) are genuinely closed for the admin-create path.
- **silent-failure-hunter — 0 MUST-FIX.** Confirmed the "201 echoes the persisted settings" justification is real rather than aspirational, by tracing `partnerPublicColumns()` → `.returning(...)` → `c.json(partner, 201)`. Confirmed the fail-open argument against the actual reader source, and that the dev-seed change cannot silently alter an existing dev DB (`seedDefaultAdmin` skips the insert when the `default-partner` slug exists). Raised the coverage gap addressed below, plus two suggestions declined in the PR body (narrowing `settings: z.any()`, and logging on the discard branch).
- **pr-test-analyzer — 0 MUST-FIX.** Verified discrimination empirically rather than by reading assertions: built a throwaway worktree, reverted the fix, and reran. Removing the handler's `applyNewPartnerDefaultSettings` call fails exactly the route tests that should catch it; breaking the merge logic fails the helper suite. Confirmed `captureInsertedValues` captures the real object handed to `.values()` rather than fabricating one. Noted `db/seed.ts` has no direct test — acknowledged as a caveat in the PR body (dev-only path; its sole obligation is a zero-arg call the helper suite already characterizes).

**Fix applied (9d9d702):** added a route-level case driving a non-object `settings` through the handler and asserting both the persisted value and the 201 echo, and made the insert mock project the persisted settings so that echo is asserted rather than assumed. Verified discriminating by reverting the handler call and watching it fail.

**Tests:** 304/304 across `orgs.test.ts`, `orgs.listQuery.test.ts`, `partnerDefaultSettings.test.ts`, `partnerCreate.test.ts`; 197/197 across the untouched inbound readers; 33/33 on `register.test.ts` + `partner-wide-write-coverage.test.ts`. `tsc --noEmit -p apps/api` clean; `eslint` clean on every touched file. Full CI is running on 9d9d702.

**Status:** review-clean, awaiting maintainer merge — not merging, and leaving #4520 open and assigned pending your verification.

ToddHebebrand:
**CI note — the 3 red checks are pre-existing on `main`, not from this PR.**

Failing: `Trivy Filesystem Scan`, `Trivy Image Scan (portal)`, `Trivy Image Scan (web)`.

The identical Trivy jobs fail on **dede9c178** — this PR's base commit on `main` — in run [33695832467](https://github.com/LanternOps/breeze/actions/runs/33695832467), which reports `Trivy Filesystem Scan`, `Trivy Image Scan (portal)`, `Trivy Image Scan (api-compose)`, `Trivy Image Scan (web)` and `Trivy Image Scan (api)` all failing. Security Scanning has been red on `main` across several consecutive commits (33695213120, 33692106418, 33682426347).

This PR adds no dependencies and does not touch `pnpm-lock.yaml`, `package.json`, or any Dockerfile — the diff is six TypeScript files under `apps/api` plus one new service module. Nothing in it can move a container-image or filesystem CVE scan, and the two failing image scans are for the **portal** and **web** images, which this PR does not touch at all.

Everything else is green or still running: 10 passed, 30 pending, 0 non-Trivy failures at the time of writing. The blocking jobs (`Test API`, `Test Web`, `Lint`, `Check Migrations`, Integration Tests) had not reported yet — worth a look before merging.

The Trivy breakage on `main` looks like it wants its own issue; flagging rather than folding it into this PR.

PR #4640, opened 2026-09-02
fix(api/ai): always notify on auto-demote, even with no readable run (#4582)

### LINKED ISSUE #4582: [AI agents] auto-demote via fix-watch recurrence can revoke a supervised key silently when no run id resolves
Found during the #4187 docs pass (PR #4581).

`apps/api/src/services/aiAgents/supervisedKeyDemote.ts:317-382` only sends the high-priority demotion notification when a `runId` resolves (dedupe key `graduation-demote-<orgAgentId>-<opKey>-<runId>`). A watch-recurrence demotion whose watch has no resolvable run only logs — the org's supervised key is removed with no recipient notified.

Spec §4.5 / P2-5 plan (Task 16) require notifying `recipients` at high priority on every auto-demote. Fix: derive the dedupe key from `(orgAgentId, opKey, watchId)` when no run is available and always notify; add a unit test for the watch-only path.

ToddHebebrand:
Closes #4582

## The bug

`notifyDemotion` (`apps/api/src/services/aiAgents/supervisedKeyDemote.ts`) stood down whenever the run's policy snapshot could not be read — either no `runId` on the input, or the run row not readable in the org:

```ts
if (!runId) { console.warn('… — skipping demote notify', …); return; }
…
if (!runRow) { console.warn('… — skipping demote notify', …); return; }
```

The auto-demote transaction had already **committed** by then: the key was removed from the org row, the audit entry written, the graduation row stamped `demoted`. So an organization lost a supervised action key — its unattended authority for that operation — with **no recipient notified**.

Spec §4.5 / P2-5 plan Task 16 require notifying `recipients` at high priority on every auto-demote. The plan also specified the dedupe key as `graduation-demote-<orgAgentId>-<opKey>-<runId ?? watchId>`; the `?? watchId` branch had been dropped as "unreachable", justified by the very stand-down that is the bug.

### Reachability — stated precisely

Because "this branch is unreachable" is exactly the reasoning that dropped the notification, this PR does not repeat it loosely:

- **`runId === null` is NOT reachable from either caller today.** `ai_agent_fix_watches.run_id` is `NOT NULL`, and `intentReleaseWorker` returns before building a `NotifyDemotionInput` when `requesting_agent_run_id` is absent. That arm is a contract guarantee for this exported function's `string | null` signature, not a live regression path. (verified against the schema and both call sites)
- **A run row that no longer resolves IS reachable.** `ai_agent_runs` is org-cascade registered, so a tenant erasure racing a fix-watch phase-2 job lands on exactly this path. (verified — `ai_agent_runs` is in `CORE_ORG_CASCADE_DELETE_ORDER`)

So: one live path, one contract path, both closed.

## The fix

**The run snapshot stays the preferred recipient source, but is no longer a precondition.** It is the policy the run actually executed under, so it names exactly the people who were on the hook for the operation being graded — that reasoning is unchanged when it *is* readable.

When it is not, the fallback is the org's **effective** recipients: the partner baseline's `recipients` unioned with the org override's, resolved through the canonical `mergeAgentPolicies` so the union rule cannot drift from what every other reader sees. The baseline row's own `recipients` column alone would **not** do — it is the PARTNER's list and silently drops everyone the organization added through its override, which is the objection the original docstring already recorded.

`resolveRecipientUserIds` re-derives live membership against **this** org either way, so the fallback cannot widen the audience beyond people who currently have access to the org the notice describes. An empty recipient set is still a legitimate outcome; an *unattempted resolution* is not.

Three smaller consequences:

- **Dedupe key** is restored to the planned `<runId ?? watchId>`. With neither id present the notice carries **no dedupe key at all** rather than a literal placeholder — `(user_id, dedupe_key)` is a partial unique index with no time bound, so a fixed suffix would collapse every *future* demotion of that tuple into the first one. With no episode there are no siblings to collapse, and `runId`/`watchId` ride in `metadata` regardless.
- **Link** degrades with the recipient source: `/settings/ai-agents` (the page whose state actually changed, and where `AiAgentGraduationPanel` lives) instead of `/ai-agents/runs/<id>`, which would 404 without a readable run.
- The baseline agent row is now read in full rather than three columns — `kind` pairs it with the org override (there is no FK between them) and its policy columns are the partner half of the merge. One read either way on the hot path; the fallback adds exactly one more.

### Every remaining way this goes quiet is now loud

Review raised that closing one silent path while leaving its siblings on `console.warn` — invisible in production — is not much of a fix. `supervisedKeyDemote.ts` now Sentry-captures all three:

| Path | Behaviour |
|---|---|
| No agent row | Cannot build a notice at all (no name, no recipients) — still returns, now **captured**. Read by primary key, so a miss is a genuine anomaly. |
| No readable run | Notifies from the effective policy — **captured**, since the recipient list differs from the run's snapshot and the missing row is worth a look. |
| Recipients resolve to nobody | Notifies no one — **captured**. An act-mode agent may only start with a resolvable recipient (`hasResolvableAgentRecipient`), so zero on a granted-then-revoked key is worth attention. |

Identifiers only in every message (per the module's leak rules); `org_id` is the only tag, since `services/sentry.ts` scrubs anything outside its allowlist.

`fixWatch.ts`'s two post-commit notification catches logged to `console.error` with no Sentry capture, so a throw out of `notifyDemotion` — the last remaining route to silence — was invisible. Both now capture. (This also closes the `sendRecurrenceNotifications` gap noted as "not fixed" in the first revision of this description.)

## Tests

`supervisedKeyDemote.test.ts` — the two tests that pinned the stand-down are inverted, `../sentry` is now mocked so "loud, not silent" is an asserted contract rather than a claim, and eight tests cover the new behaviour:

- no run id → notifies the **union** of partner-baseline and org-override recipients (proves it is not just the baseline column), dedupe key keyed by `watchId`, link `/settings/ai-agents`;
- run row gone → same fallback, dedupe key still keyed by `runId` (the episode identity does not depend on the row being readable);
- neither run nor watch → still notifies, with a **null** dedupe key;
- run row present but `policySnapshot` null → falls through to empty recipients rather than throwing;
- effective policy names nobody → sends nothing, asserts the resolution **ran** and that both Sentry reports fired;
- org override row gone → notifies from the partner baseline alone;
- agent row gone → sends nothing, asserts the Sentry report;
- **the new org-override read's predicate asserted as compiled SQL** (`org_id`, `kind`, `disabled_at IS NULL`) — the capture-and-replay mock hands back the queued row whatever the query shape is, so without this a dropped `org_id` filter would pass every other test in the block. **Verified by mutation**: removing `eq(aiAgents.orgId, orgId)` from the implementation fails this test and only this test.

Verification (all run locally, single-fork):

- `npx vitest run src/services/aiAgents/supervisedKeyDemote.test.ts` — 22 passed (red first: 4 failed before the implementation).
- `npx vitest run` over `supervisedKeyDemote` + `fixWatch` + `fixWatch.sql` + `supervisedKeyGrant` + `effectivePolicy` + `intentReleaseWorker` + `workerEntrypointClosure.contract` — **360 passed**. The closure contract matters: `supervisedKeyDemote` is in `fixWatchWorker`'s import graph, and both new imports (`mergeAgentPolicies` from `effectivePolicy.ts`, `captureException` from `services/sentry.ts`) were already in that closure — no new edge.
- `npx tsc --noEmit` — clean. `npx eslint` on all three changed files — clean.

No schema, migration, tenancy or cascade surface is touched.

### DIFF
--- a/apps/api/src/services/aiAgents/fixWatch.ts
+++ b/apps/api/src/services/aiAgents/fixWatch.ts
@@ -880,6 +880,14 @@ export async function checkFixWatchPhase2(watchId: string): Promise<FixWatchPhas
     console.error('[fixWatch] failed to notify a recurrence (non-fatal — the watch state is already committed)', {
       watchId, error,
     });
+    // #4582 — a console line is invisible in production, and both catches
+    // below/above swallow the only signal a human would ever get that a
+    // committed verdict went unannounced. Identifiers only in the message.
+    captureException(
+      new Error(`recurrence notification failed for fix watch ${watchId}; the verdict is committed`, {
+        cause: error,
+      }),
+    );
   }
   // Separately caught: a failing recurrence notification must not suppress
   // the revoke notice, which is the more consequential of the two (an
@@ -892,6 +900,16 @@ export async function checkFixWatchPhase2(watchId: string): Promise<FixWatchPhas
       console.error('[fixWatch] failed to notify a supervised-key revoke (non-fatal — the revoke is already committed)', {
         watchId, opKey: demotion.opKey, error,
       });
+      // #4582 — the revoke is COMMITTED: an org just lost unattended
+      // authority. `notifyDemotion` reports its own no-recipient outcomes, so
+      // a throw reaching here is the last remaining way that becomes silent.
+      captureException(
+        new Error(
+          `supervised-key revoke notification failed for fix watch ${watchId} `
+          + `(op key "${demotion.opKey}"); the revoke is committed`,
+          { cause: error },
+        ),
+      );
     }
   }
   return { action: 'recurred' };
--- a/apps/api/src/services/aiAgents/supervisedKeyDemote.test.ts
+++ b/apps/api/src/services/aiAgents/supervisedKeyDemote.test.ts
@@ -22,6 +22,10 @@ const RUN_ID = '00000000-0000-4000-8000-0000000000b5';
 const WATCH_ID = '00000000-0000-4000-8000-0000000000b6';
 const INTENT_ID = '00000000-0000-4000-8000-0000000000b7';
 const USER_ID = '00000000-0000-4000-8000-0000000000b8';
+/** Named on the PARTNER baseline row's `recipients`. */
+const PARTNER_USER_ID = '00000000-0000-4000-8000-0000000000b9';
+/** Named ONLY on the ORG override's `recipients` — the half the baseline column drops. */
+const ORG_USER_ID = '00000000-0000-4000-8000-0000000000ba';
 const SCRIPT_ID = '00000000-0000-4000-8000-0000000000bf';
 const OP_KEY = 'manage_services:restart';
 const OTHER_KEY = 'manage_alerts:acknowledge';
@@ -457,8 +461,22 @@ describe('notifyDemotion', () => {
     ...overrides,
   });
 
+  /** The PARTNER baseline row `agentId` names — read in full so the run-less
+   *  fallback can merge its policy with the org override's. */
+  function baselineRow(overrides: Record<string, unknown> = {}) {
+    return {
+      ...orgRow([OP_KEY]),
+      id: AGENT_ID,
+      orgId: null,
+      partnerId: PARTNER_ID,
+      name: 'Disk Cleaner',
+      recipients: { userIds: [PARTNER_USER_ID], roleIds: [] },
+      ...overrides,
+    };
+  }
+
   function queueNotify(opts: { agent?: unknown[]; run?: unknown[] } = {}): void {
-    state.selectQueue.push(opts.agent ?? [{ name: 'Disk Cleaner', orgId: null, partnerId: PARTNER_ID }]);
+    state.selectQueue.push(opts.agent ?? [baselineRow()]);
     state.selectQueue.push(
       opts.run ?? [{ policySnapshot: { effective: { recipients: { userIds: [USER_ID] } } } }],
     );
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three lenses in parallel (`code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`), all on Sonnet.

**Findings:** 7 raised → 7 addressed in `65b30e61b`; 0 outstanding.

- **CRITICAL (silent-failure)** — the remaining `!agentRow` stand-down was the same silent shape #4582 was filed against, and the new docstring's "every auto-demote notifies" was false for it. It genuinely cannot build a notice (no name, no recipients), so it stays a `return` — but it is now `captureException`, not a console line no operator sees. Same treatment for the effective-policy fallback and for a recipient set that resolves to nobody.
- **IMPORTANT (silent-failure)** — `console.warn` is invisible in production and out of step with every sibling in `services/aiAgents/`. Three Sentry captures added; identifiers only in the message, `org_id` the only tag (`services/sentry.ts` scrubs anything outside its allowlist).
- **IMPORTANT (silent-failure)** — `normalizeAgentPolicy`'s zod `.parse` is a new throw surface, and `fixWatch.ts`'s two post-commit notification catches logged to console with no capture. Both now capture, so a throw out of `notifyDemotion` is no longer the last silent route.
- **IMPORTANT (test-analyzer)** — reachability was overstated. `runId === null` is **not** producible by either caller (`ai_agent_fix_watches.run_id` is `NOT NULL`; the release worker returns before building an input without a run); the reachable arm is a run row that no longer resolves, via org-erasure cascade racing a fix-watch phase-2 job. Corrected in the docstring and the PR body — "this branch is unreachable" is precisely the reasoning that dropped the notification in the first place.
- **IMPORTANT (all three lenses)** — the new org-override read had no compiled-SQL assertion, against this suite's own stated bar; the capture-and-replay mock returns the queued row whatever the predicate is. Now pinned like its `demoteSupervisedKey` sibling. **Verified by mutation**: removing `eq(aiAgents.orgId, orgId)` fails that test and only that test.
- **MINOR** — zero resolved recipients logged nothing at all; now captured (it is the outcome #4582 is ultimately about).
- **MINOR** — `policySnapshot: null` branch uncovered; test added.

Confirmed clean by review and re-checked: `dedupeKey: null` is handled correctly by `userNotifications.ts` (the partial unique index excludes NULL), the fallback cannot widen the notified audience (`resolveRecipientUserIds` re-derives live membership against this org in both branches), and `effectivePolicy.ts` is unmocked in the suite so the union assertion exercises the real merge rather than echoing a stub.

**Tests:** `supervisedKeyDemote.test.ts` 22 passed (red first — 4 failed before the implementation). Full affected set single-fork — `supervisedKeyDemote` + `fixWatch` + `fixWatch.sql` + `supervisedKeyGrant` + `effectivePolicy` + `intentReleaseWorker` + `workerEntrypointClosure.contract` — **360 passed**. `tsc --noEmit` clean; `eslint` clean on all three changed files.

**Status:** review-clean, awaiting maintainer merge. CI is still running on `65b30e61b` at time of writing — no failures reported yet, but please confirm green before merging.

PR #4641, opened 2026-09-02
fix(web): persist Remote Tools tab in the URL hash and let the terminal fill the viewport (#4512, #4510)

### LINKED ISSUE #4510: Remote Tools Terminal: grow pane to fill browser viewport height when output overflows
## Ask
In **Remote Tools → Terminal**, the terminal pane is a fixed-height box. When output grows (or even when idle), a large empty region remains below it in the browser viewport. Please make the terminal pane **flex/grow to use available vertical space** (match browser content height), with internal scroll inside xterm as needed.

## Why
Long paste/diagnostic sessions (multi-line scripts, heredocs) are harder to read; techs resize the browser but the pane does not follow. Small UX win for self-host MSP workflows.

## Context
Noticed while verifying Terminal paste behaviour on self-host **v0.109.0** web (#4090 / #4152). Separate from connect/paste bugs.

### LINKED ISSUE #4512: Remote Tools: browser refresh while on Terminal jumps to Processes tab
## Summary
While a Remote Tools **Terminal** session is open, a normal browser refresh (reload the page) does not stay on / restore the Terminal tab. It lands on the **Processes** tab instead.

## Expected
Refresh (or reconnect) should return to the same Remote Tools surface the user was on — ideally **Terminal** with session state restored, or at least re-open Terminal rather than silently switching to Processes.

## Actual
Refresh → Processes tab.

## Context
Seen on self-host canary while debugging Desktop Access / helper on an Intel Mac (RoNY18-Jing). Easy to hit when operators reload after a paste/mangle or to clear a stuck Terminal pane.

## Related
- #4152 (Terminal connect lifecycle / File Browser warmup)
- #4510 (Terminal pane height)

ToddHebebrand:
## Summary

**#4512** — `RemoteToolsPage.tsx` kept the active tab in plain `useState(initialTab)` with nothing syncing it to `window.location.hash`, and `tools.astro` never passed an `initialTab`, so a browser refresh while on the Terminal tab always landed back on Processes. Switched to the repo's existing `useHashTab` hook (`apps/web/src/lib/useHashState.ts`) — the same pattern already used by `DnsSecurityPage.tsx` / `DeviceDetails.tsx` (CLAUDE.md "URL State in Components": hash, not query params). The hash is read post-mount (SSR-safe) and written on every tab click via a new `switchTab` callback.

**#4510** — `RemoteTerminal.tsx`'s root wrapper had no `flex-1`/`min-h-0`, so it never grew to fill `RemoteToolsPage`'s `flex-1 min-h-0` tab content area — it only ever sized to its header plus the inner container's 400px floor (`u-min-h-px-400`). Added `flex-1 min-h-0` to the root wrapper and the inner "Terminal Container"/xterm-mount divs so the pane fills whatever height its flex parent gives it, while keeping the 400px floor. The component already had a `ResizeObserver` that calls `fitAddon.fit()` on resize, and xterm's `onResize` already forwards the new cols/rows to the server — no protocol change was needed, this was purely a CSS/layout bug.

## Test plan
- [x] Red-first: added `RemoteToolsPage.test.tsx` (hash-driven tab activation on mount, hash written on tab click, hashchange re-sync, default/unknown-hash fallback) and two new cases in `RemoteTerminal.test.tsx` (wrapper carries `flex-1`/`min-h-0`/`flex-col`; xterm container keeps `flex-1` + the 400px floor) — confirmed all 4 new assertions failed against pre-fix code, then passed after the fix.
- [x] `cd apps/web && npx vitest run src/components/remote` — 20 files / 151 tests passed.
- [x] `cd apps/web && npx tsc --noEmit -p .` — clean.
- [x] `npx eslint` scoped to the touched files — clean.

Closes #4512
Closes #4510

https://claude.ai/code/session_01Xr3ZNTFRkCM1t4gEjPttTg

### DIFF
--- a/apps/web/src/components/remote/RemoteTerminal.test.tsx
+++ b/apps/web/src/components/remote/RemoteTerminal.test.tsx
@@ -582,3 +582,22 @@ describe('RemoteTerminal hostname prop lifecycle (#4152)', () => {
     expect(MockWebSocket.instances).toHaveLength(1);
   });
 });
+
+describe('RemoteTerminal layout fills its flex parent (#4510)', () => {
+  it('gives the root wrapper flex-1 min-h-0 so it grows inside a flex column parent', () => {
+    const { container } = renderTerminal();
+
+    const root = container.firstElementChild as HTMLElement;
+    expect(root.className).toMatch(/\bflex-1\b/);
+    expect(root.className).toMatch(/\bmin-h-0\b/);
+    expect(root.className).toMatch(/\bflex-col\b/);
+  });
+
+  it('keeps the xterm container growing to fill the wrapper while retaining the 400px floor', () => {
+    const { container } = renderTerminal();
+
+    const xtermContainer = container.querySelector('.u-min-h-px-400') as HTMLElement | null;
+    expect(xtermContainer).not.toBeNull();
+    expect(xtermContainer!.className).toMatch(/\bflex-1\b/);
+  });
+});
--- a/apps/web/src/components/remote/RemoteTerminal.tsx
+++ b/apps/web/src/components/remote/RemoteTerminal.tsx
@@ -677,7 +677,12 @@ export default function RemoteTerminal({
   return (
     <div
       className={cn(
-        'flex flex-col rounded-lg border bg-card shadow-xs overflow-hidden',
+        // flex-1 min-h-0 (#4510): lets the pane grow to fill a flex-column
+        // parent's available height (RemoteToolsPage's terminal tab panel)
+        // instead of shrinking to the header + the inner 400px floor below.
+        // Inert when the parent isn't a flex container (e.g. RemoteTerminalPage,
+        // which sizes this via the `className` prop instead).
+        'flex flex-1 min-h-0 flex-col rounded-lg border bg-card shadow-xs overflow-hidden',
         isFullscreen && 'fixed inset-4 z-50',
         className
       )}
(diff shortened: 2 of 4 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (Sonnet subagent)
**Findings:** 1 raised (worth-fixing) → addressed in 0170cf1b5; 0 outstanding.
- Worth-fixing: `VALID_TABS` (used to validate the deep-linked hash) included windows-only tab ids regardless of the current device's OS, so a stale/bookmarked/shared `#services`-style hash on a non-Windows device could land `activeTab` on a tab whose button is filtered out of `availableTabs` and whose content is OS-gated — no active tab highlighted, empty content pane. Fixed by resetting to the Processes tab (and correcting the hash) whenever the active tab is windows-only but the resolved device isn't Windows, with a new regression test.
- Everything else reviewed clean: hash read/write mechanics match the established `useHashTab` pattern (`DnsSecurityPage.tsx`), the `isFullscreen` conditional removal on the terminal wrapper is a no-op in fullscreen mode (governed by `position:fixed`) and a correct fix for the embedded case, and `RemoteTerminal`'s other consumer (`RemoteTerminalPage.tsx`, non-flex parent) is unaffected since `flex-1` is inert there.

**Tests:** `apps/web` remote suite green — `cd apps/web && npx vitest run src/components/remote` → 20 files / 152 tests passed. `cd apps/web && npx tsc --noEmit -p .` clean. `npx eslint` scoped to the 4 touched files clean.

PR #4642, opened 2026-09-02
fix(api/tenancy): sever ai_agent_runs.ticket_id on a ticket org-move (#4524)

### LINKED ISSUE #4524: [API][Tenancy] moveTicketOrg does not sever ai_agent_runs.ticket_id — silent cross-org pointer on ticket org-move
**Problem:** PR #4436 (#4215) severs `ai_agent_runs.ticket_id` when a *device* moves org. The ticket axis has the same gap: `moveTicketOrg` in `apps/api/src/services/ticketService.ts` moves a ticket between orgs and does not null out `ai_agent_runs.ticket_id` for runs attached to it. After the move, a run scoped to the old org points at a ticket now owned by the new org.

**Why it will not announce itself:** `ticket_drafts` and `action_intents` in the same function fail with 23503 on a bad cross-org pointer, but `ai_agent_runs.ticket_id` is a plain single-column FK, so the stale pointer survives silently. #4436 also deleted `KNOWN_UNDETACHED_RUN_LINEAGE_COLUMNS`, which was the only written record of this hazard; PR #4515 restores a comment.

**Fix:** in `moveTicketOrg`, `UPDATE ai_agent_runs SET ticket_id = NULL WHERE ticket_id = $ticketId` inside the same transaction (mirroring #4436's device-side statement), a real-Postgres integration test with a cross-org fixture, and a note in the org-lifecycle registry if tickets are ever merged rather than moved.

Related: #4215, PR #4436, PR #4515.

ToddHebebrand:
Closes #4524

## The gap

`moveTicketOrg` (`apps/api/src/services/ticketService.ts`) moves a ticket between two organizations of the same partner. Agent-run history deliberately **stays with the SOURCE org** (owner decision 2026-08-23 — `ai_agent_runs.org_id` is trigger-immutable via `ai_agent_runs_immutable_guard`, and the table is intentionally absent from `TICKET_ORG_DENORMALIZED_TABLES`). But nothing severed `ai_agent_runs.ticket_id`, so after a move a source-org run kept pointing at a ticket the **target** org now owns, and `/ai-agents/:id/runs` would serve that foreign ticket id back to the source org.

**Why it never announced itself:** `ticket_drafts` and `action_intents` in the same function carry composite `(x, org_id) -> tickets(id, org_id)` tenant FKs and raise 23503 on a cross-org pointer — which is why they already have tombstone/delete statements. `ai_agent_runs.ticket_id` is a **plain single-column FK** with `ON DELETE SET NULL`, so the ticket UPDATE completed happily and the stale pointer survived in silence.

This is the ticket-axis twin of the device-axis detach shipped in #4215 / #4436. There is **no** Postgres trigger on `UPDATE OF org_id ON tickets` — the device axis has `breeze_cascade_device_org_id()`, the ticket axis has no equivalent — so this service is the only place the contract can be enforced.

## Changes

- **`moveTicketOrg` nulls `ai_agent_runs.ticket_id`** for the moved ticket, inside the same transaction. Placed after the ticket UPDATE but **before** `assertTicketMoveCurrencyCompatible`: that guard takes row locks on `time_entries`/`ticket_parts`, and the device-move path locks `ai_agent_runs` immediately after the parent row flip and before any ticket-child table — so the reverse order would give a concurrent ticket-move and device-move an inverted lock pair, the exact deadlock shape the function's global lock-order comment exists to prevent.
- **Also nulls the reverse pointer `ticket_comments.agent_run_id`.** `ticket_comments` has no `org_id` (child-via-parent tenancy), so every comment travels into the target org while the run that authored it stays behind — the same reverse-pointer class #3828 fixed for `metric_anomaly_incidents.agent_run_id` on the device axis, just the other end of this ticket's link. `origin_principal_kind` is untouched, and that (not `agent_run_id`) is what the helpdesk loop guard keys on, so severing cannot open a feedback loop.
- **Note beside `orgMergeRegistry`'s `tickets` entry** (as the issue asked) recording why the merge path needs no equivalent: a merge repoints `tickets.org_id` by direct SQL without going through `moveTicketOrg`, but Phase A fences the loser org to `status='merging'` **before** the repoint runs — dropping it out of `accessibleOrgIds` and every ingress gate — and `ai_agent_runs` is `leave-for-erasure`, so those rows are unreachable and then destroyed with the loser shell rather than outliving it. `ticket_id` being a plain FK also means it can never 23503 the repoint the way `ticket_drafts` does.

No migration: this is service logic only, and the schema is unchanged.

## Tests

Three real-Postgres cases in `apps/api/src/__tests__/integration/ticket-move-org.integration.test.ts`:

1. **Detach + controls** — nulls `ticket_id` on a device-less `trigger_kind='ticket'` run *and* on a run that names both the ticket and a device (rejects a `WHERE device_id IS NULL` narrowing), asserts both runs stay in the source org, and keeps a run on a **sibling ticket that never moved** intact (rejects a blanket "null every ticket_id").
2. **Real RLS context** — drives the move under a genuine partner-scoped `withDbAccessContext`, not `withSystemDbAccessContext`. `ai_agent_runs` is `FORCE ROW LEVEL SECURITY` with a `breeze_has_org_access(org_id)` policy, so a detach that only worked under `breeze.scope=system` would be a silent zero-row no-op on the one path that actually runs in production. Every other case in that file uses system scope and could not catch it.
3. **Reverse pointer** — an AI-authored comment's `agent_run_id` is nulled while its `ticket_id`, content and `origin_principal_kind` survive.

All three were confirmed **red before the fix** (`expected '<uuid>' to be null`) and green after.

### Verification

| Check | Result |
|---|---|
| `ticket-move-org.integration.test.ts` | 14 passed (11 pre-existing + 3 new) |
| `ticketAttachmentsRls` + `orgCurrencyCreationBarrier` + `agentRunMoveSemantics` integration | 29 passed |
| Unit: `ticketService`, `tickets/moveOrg`, `devices/moveOrg.coverage`, `aiToolsTicketing` ×2, `orgMerge` ×3, `ticketService.aiExecutors` | 349 passed |

`ticketService.test.ts`'s `../db/schema` mock gained `aiAgentRuns` and the two `ticketComments` columns — without them the new statements threw `No "aiAgentRuns" export is defined on the mock`, which is incidentally a nice proof that the mocked path actually reaches them.

## Adjacent gaps found by the sweep (filed separately, not fixed here)

Sweeping every pointer between `tickets` and `ai_agent_runs` turned up three more stale-cross-org-state defects. Each spans **both** move axes (ticket and device) or a different subsystem, so fixing one axis here would reproduce exactly the asymmetry that created this issue:

- **#4643** — `ticket_email_links.org_id` is never re-stamped on a ticket move (absent from both `TICKET_ORG_DENORMALIZED_TABLES` and the device path's `CUSTOM_ORG_REWRITE_TABLES`). Live today. A `NOT`-here comment now marks it as a tracked gap rather than a ruling.
- **#4644** — `ticket_comments.agent_run_id` has the same reverse-pointer gap on the **device** axis; closing it means a new migration replacing `breeze_cascade_device_org_id()`. Nothing writes the column yet, so neither axis leaks today.
- **#4645** — `device_vulnerabilities.ticket_id` keeps naming a departed ticket. Consequence today is a dead link under RLS, and nulling it would discard a remediation linkage — a product call, not a mechanical one.

### DIFF
--- a/apps/api/src/__tests__/integration/ticket-move-org.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ticket-move-org.integration.test.ts
@@ -887,15 +887,70 @@ describe('moveTicketOrg — ai_agent_runs.ticket_id detach (#4524)', () => {
       .returning();
     expect(aiComment.agentRunId).toBe(run.id);
 
+    // Control: an AI comment on a DIFFERENT ticket in the SAME source org.
+    // Rejects a detach scoped by org rather than by the moved ticket — e.g.
+    // `WHERE ticket_id IN (SELECT id FROM tickets WHERE org_id = source)`,
+    // which the single-comment case alone would pass.
+    const sibling = await seedSiblingTicket(orgA.id, partner.id);
+    const siblingRun = await seedRunOnTicket(orgA.id, agent.id, sibling.id, null);
+    const [siblingComment] = await adminDb
+      .insert(ticketComments)
+      .values({
+        ticketId: sibling.id,
+        authorType: 'internal',
+        commentType: 'comment',
+        content: 'Triage note on a ticket that never moves.',
+        isPublic: false,
+        originPrincipalKind: 'ai_agent',
+        agentRunId: siblingRun.id,
+      })
+      .returning();
+
     await withSystemDbAccessContext(() => moveTicketOrg(ticket.id, orgB.id, actor));
 
     const after = await readComment(aiComment.id);
     expect(after?.agentRunId).toBeNull();
+    // The sibling ticket never moved, so its comment keeps its run link.
+    expect((await readComment(siblingComment.id))?.agentRunId).toBe(siblingRun.id);
     // Only the cross-org link is dropped. The comment itself, its content and
     // its non-user provenance (which is what the helpdesk loop guard actually
     // keys on) all survive the move.
     expect(after?.ticketId).toBe(ticket.id);
     expect(after?.content).toBe('Proposed next step from triage.');
     expect(after?.originPrincipalKind).toBe('ai_agent');
+  });
+
+  it('rolls the detach back with the rest of the transaction when the currency guard blocks the move', async () => {
+    // The detach runs BEFORE assertTicketMoveCurrencyCompatible (deliberately —
+    // see the lock-order note at the statement), so a blocked move must unwind
+    // it along with everything else. Verified discriminating by mutation: hoist
+    // the detach out of db.transaction (its own withSystemDbAccessContext, as a
+    // "tidier" refactor might) and ONLY this case fails — every other test in
+    // this block still passes while the pointer is severed on a move that never
+    // happened. Note that swapping `tx` for `db` is NOT that mutation: `db` is
+    // an AsyncLocalStorage-bound proxy (db/index.ts) that resolves to the same
+    // pinned connection inside the callback, so it stays transactional.
+    const adminDb = getTestDb() as any;
+    const { orgA, orgB, partner, actor, ticket, timeEntry, ticketPart } = await seedMoveOrgFixture();
+    const agent = await seedTriageAgent(orgA.id, partner.id);
+    const run = await seedRunOnTicket(orgA.id, agent.id, ticket.id, null);
+
+    await setOrgCurrency(orgB.id, 'EUR');
+    await adminDb.update(timeEntries)
+      .set({ hourlyRate: '100.00', isBillable: false, billingStatus: 'not_billed' })
+      .where(eq(timeEntries.id, timeEntry.id));
+    await adminDb.delete(ticketParts).where(eq(ticketParts.id, ticketPart.id));
+
+    const err = await withSystemDbAccessContext(() =>
+      moveTicketOrg(ticket.id, orgB.id, actor)
+    ).catch((e) => e);
+    expect(err).toBeInstanceOf(TicketMoveCurrencyBlockedError);
+
+    // Nothing moved — and the pointer is still there, because the ticket is
+    // still the source org's.
+    expect((await readTicket(ticket.id))?.orgId).toBe(orgA.id);
+    const after = await readRun(run.id);
+    expect(after?.ticketId).toBe(ticket.id);
+    expect(after?.orgId).toBe(orgA.id);
   });
 });
--- a/apps/api/src/__tests__/integration/ticket-move-org.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ticket-move-org.integration.test.ts
@@ -730,3 +730,172 @@ describe('moveTicketOrg — requester contact detach (#3258 W03)', () => {
     expect(after.orgId).toBe(orgA.id);
   });
 });
+
+// ── #4524: the ai_agent_runs ↔ ticket pointer pair must be severed on a move ──
+
+/**
+ * Seeds one `ai_agent_runs` row pointing at `ticketId`. Unlike
+ * `seedTicketScopedIntent` (which needs a run only as an intent's requester
+ * and leaves `ticket_id` NULL), this is the shape #4524 is about: a
+ * trigger_kind='ticket' run whose `ticket_id` names the ticket.
+ *
+ * `deviceId` is parameterised because ticket-triggered runs are normally
+ * device-LESS (the ticket axis stamps ticket_id and leaves device_id NULL),
+ * while a device-triggered run can also carry a ticket_id — the ticket-keyed
+ * detach must reach both, and a `WHERE device_id IS NULL` narrowing must not
+ * pass.
+ */
+async function seedRunOnTicket(
+  orgId: string,
+  agentId: string,
+  ticketId: string,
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` on `393845cb0`, then a second focused `code-reviewer` pass on the lock-order fix (`a3e4e2a78`).

**Findings:** 5 raised → 5 addressed, 0 outstanding.

1. **Lock-order inversion (`code-reviewer`, important).** The detach originally sat *after* `tx.update(tickets)`, but `breeze_cascade_device_org_id()` severs `ai_agent_runs` in its first two statements and only then runs the loop that re-stamps its child tables — and `tickets` is one of them (`CORE_DEVICE_ORG_DENORMALIZED_TABLES`, `routes/devices/core.ts:243`). A device-move and a ticket-move racing over the same device-linked ticket could each hold the lock the other wanted → `40P01`. My original comment argued the placement was safe but only compared `ai_agent_runs` against `time_entries`/`ticket_parts`, never against `tickets` itself. Verified against the trigger source and fixed in `a3e4e2a78` by moving the statement above the ticket UPDATE. Re-review confirms the inversion is gone with no regression to RLS, FK, or rollback behaviour.
2. **Rollback untested (`pr-test-analyzer`, critical).** Nothing pinned that a currency-blocked move unwinds the detach. Added in `b10cc9af7` and **proved discriminating by mutation**: hoist the detach out of `db.transaction` and only that case goes red. (Swapping `tx` for `db` is *not* that mutation — `db` is an AsyncLocalStorage-bound proxy resolving to the same pinned connection, so it stays transactional. Recorded in the test comment so nobody re-derives it.)
3. **Missing sibling control on the comment detach (`pr-test-analyzer`).** A detach scoped by org rather than by the moved ticket would have passed. Control added in `b10cc9af7`; mutation-verified to fail on the over-broad WHERE.
4. **Lock-order comment understated the real sequence (re-review).** It read as though `ai_agent_runs` is taken right after the org locks; it is actually fourth (`action_intents` and `ticket_drafts` precede it). Fixed in `421eb1fb4`, which now also records *why* the `action_intents` inversion is safe (the mutually-exclusive `scope_kind` CHECKs make the two axes' row sets disjoint) and why that argument does not extend to `ticket_alert_links`.
5. **`ticket_comments` placement overclaimed (re-review).** Running it before the ticket UPDATE would work too — every caller holds both orgs for the whole transaction. Reworded to "placement for readability, not a constraint."

`silent-failure-hunter` found **no active silent no-op**: it traced both callers and confirmed each gates on source *and* target org access before entering, using the same `accessibleOrgIds` the `DbAccessContext` is built from, so app-layer and RLS cannot drift here.

**Tests:** `ticket-move-org.integration.test.ts` 15 passed (11 pre-existing + 4 new); `ticketAttachmentsRls` + `orgCurrencyCreationBarrier` + `agentRunMoveSemantics` 29 passed; 349 unit passed across `ticketService`, `tickets/moveOrg`, `devices/moveOrg.coverage`, `aiToolsTicketing` ×2, `orgMerge` ×3. `tsc --noEmit` clean, `eslint` clean. All four new cases were confirmed **red before the fix**, and the two controls added in review were confirmed discriminating by mutating the implementation.

**CI:** green so far on the head commit with **0 failures**. Note that `Trivy Filesystem Scan` and `Trivy Image Scan (api / api-compose / web / portal)` are **already failing on `main`** (run `33697627853`) — pre-existing and unrelated to this diff, which is 4 statements of TypeScript.

**Follow-ups filed, not fixed here** — each spans both move axes or another subsystem, so half-fixing one axis would reproduce the very asymmetry that created #4524: **#4643** (`ticket_email_links.org_id` never re-stamped — live today), **#4644** (the `ticket_comments.agent_run_id` reverse pointer on the device axis), **#4645** (`device_vulnerabilities.ticket_id`), **#4657** (`ticket_alert_links` locked in opposite order by the two paths — the remaining instance of the deadlock class fixed here), and **#4650** (the AI `move_org` tool is structurally unreachable: both release-time actor contexts pin `accessibleOrgIds` to one org, so it can never pass its own cross-org gate). Each is cited from the code comment that notes the gap.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4524 — both are yours.

ToddHebebrand:
**CI settled on head `0ef5b1a46`: 65 success, 1 skipped (`Main Red Alert`, main-only), 0 failures.**

Both blocking jobs that actually exercise this change are green — `Test API`, and all four `Integration Tests` shards (which is where the four new cases in `ticket-move-org.integration.test.ts` run). `Type Check` green too.

Correcting the caveat in my previous comment: the Trivy jobs are **green on this head**, all ten of them. They were red on `main` (run `33697627853`) and red on an earlier push of this branch, so I flagged them as an expected pre-existing failure — that turned out not to hold, and nothing here is blocked on them.

Nothing outstanding from my side. Merge and the closure of #4524 are yours.

PR #4646, opened 2026-09-03
fix(api): re-issue the caller's session when MFA recovery codes are rotated (#4480)

### LINKED ISSUE #4480: [API][MFA] Regenerating recovery codes bumps mfa_epoch and evicts the current session — the new codes are never displayed
**Verified in browser (2026-09-01 re-verify, post-#4439).** The regenerate flow now confirms before mutating (#4414) and promises to show the new codes once. But `POST /auth/mfa/recovery-codes` bumps `mfa_epoch` (`apps/api/src/routes/auth/mfa.ts:1246`), which invalidates the caller's own session — the 200 response carrying the codes lands on a page that is already being evicted to /login, so the user never sees them. Net effect: every regenerate silently destroys the old codes and hides the new ones — the failure #4414 was meant to end, now via the API instead of the UI.

**Decision needed (not a one-liner):** regenerating recovery codes is not a factor change. Options: (a) don't bump `mfa_epoch` on recovery-code regeneration (other sessions keep working — codes are single-use secrets, not session authority); (b) bump the epoch but exempt the issuing session (stamp the new epoch onto it in the same response); (c) keep the eviction and make the web render the codes before the eviction fires. (a) is the simplest and matches how enrollment-time codes are shown. Whichever is chosen, add a browser-level test: regenerate → codes visible → next page load still authenticated.

**Not release-blocking for 0.109.0** (enrollment shows codes correctly; regenerate is the rarer path) but it should be called out as a known issue in the release notes.

Reported by: internal Playwright re-verify 2026-09-01, evidence in docs/testing/FEATURE_TEST_LOG.md.

ToddHebebrand:
## The bug

`POST /auth/mfa/recovery-codes` routed its write through `invalidateMfaAssuranceAfterFactorChange`, which advances `mfa_epoch` **and** revokes every refresh-token family for the user.

That blast radius is correct — a stale recovery-code set must not remain usable from another live session (SR2-19). What was wrong is that it also evicted the **actor**. `middleware/auth.ts:578` rejects the caller's next request on the stale `mep` claim, and the refresh cookie it would retry with belongs to a family revoked in the same transaction, so `fetchWithAuth`'s refresh fails and `handleSessionExpired` hard-redirects to `/login` — while the one-time plaintext codes are still on screen. Net effect: every regenerate destroyed the old codes and hid the new ones.

## The fix

Rotation now takes the road `/mfa/enable` already takes for the codes **it** shows at enrollment (`completeInitialMfaEnrollment`): one transaction advances the epoch, revokes every family, mints a **replacement** session bound to the post-bump epochs, and writes the new hashes. Every other session still dies; the caller leaves holding a token that works, plus rotated refresh/CSRF cookies and `tokens.accessToken` in the body.

Of the three options in the issue this is **(b)**. Not (a): dropping the epoch bump gives up a real control. Not (c): timing-dependent, and browser-only.

### Changes

- **`services/mfaEnrollmentSession.ts`** — extract `replaceSessionOnMfaFactorWrite`, with `expectedMfaEnabled` folded into the epoch bump's `WHERE` so the precondition is checked under the row lock rather than in a racy pre-read (a concurrent `/mfa/disable` now loses with a 409 instead of writing codes onto an unprotected account). `completeInitialMfaEnrollment` becomes the `false` specialization and keeps its MFA-assured-identity invariant — enrollment behaviour is unchanged.
- **`routes/auth/mfa.ts`** — the rotation route mirrors `/mfa/enable`: `beginAuthIssuance` → `replaceSessionOnMfaFactorWrite` → `bindIssuedUserSession` + `installAuthorizedUserSessionCookies`, with `cancelAuthIssuance` on every failure path. The replacement identity carries the caller's **own** `mfa` claim (a password step-up must never upgrade a session's assurance) and the **signed** `mdid` via `carryForwardBinding`, never the forgeable `x-breeze-mobile-device-id` header — SR-001, same rule `/auth/refresh` follows.
- **`services/tokenRevocation.ts` / `authLifecycle.ts`** — `revokeAllUserTokens` takes an optional `preserveTokensIssuedAtOrAfter`, and the replacement flow passes its pre-issuance instant. The default cutoff is `now - 1s`, so any stall over a second between minting the replacement and post-commit cleanup made the fresh **refresh** token satisfy `iat <= revokedAfter` and get rejected at `/auth/refresh` — the same eviction, just delayed to the first refresh. This latent race predates this PR (it applies equally to enrollment); the option only ever *narrows* the cutoff, so every other caller is byte-for-byte unchanged.
- **Web** — `commitReissuedSessionIfCurrent` on the auth store (tokens only; unlike `commitMfaEnrollmentIfCurrent` it asserts nothing about the user record), and `ProfilePage.handleGenerateRecoveryCodes` adopts `data.tokens` under a session-generation fence *before* revealing the codes. A refused commit (a logout raced the request) still reveals them — the old set is already gone, so hiding them would strand the user with no working set at all.
- **`openapi.ts`** — documents the replacement `tokens`, and the new 409 / 428.

## Verification

- `src/services/mfaEnrollmentSession.test.ts`, `src/services/tokenRevocation.test.ts`, `src/routes/auth.test.ts` — **red first**, including a mutation check that the signed-binding assertion actually discriminates (reverting to `readMobileDeviceId(c)` fails it).
- Real-Postgres: `__tests__/integration/mfaEnrollmentSession.integration.test.ts` gains two rotation cases — one proving every other family is revoked while the returned family is the only live one, one proving the `expectedMfaEnabled: true` precondition refuses to rotate onto an account whose MFA was disabled underneath it (no epoch bump, no revoke, no code write).
- Web: new `ProfilePage.recoveryCodeRotation.test.tsx` (adopt / stale-generation / older-API-without-tokens) plus store-level coverage in `stores/auth.test.ts`.
- API affected suites 243 + 260 passed; web ProfilePage/MFASettings/auth-store 195 passed; `tsc --noEmit` clean in `apps/api` and `apps/web`; eslint clean on every changed file.

## Notes for review

- This changes the endpoint's client contract: a caller with no auth binding now gets 428/409 where it previously got a 200. That is already true of `/mfa/enable`, and the web app is the only client of this endpoint (`grep` across mobile/portal/helper/e2e finds none).
- `passkeys.ts:350` (registering a **second** passkey) still uses the eviction-only path. It shows no one-time secret so it is not this bug, but it does sign the user out — worth a separate look.

Closes #4480


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (full-rigor auth pass over the extraction, the route diffed line-by-line against `/mfa/enable`, the revocation-cutoff monotonicity, the web generation fence, and authority-elevation) + `pr-test-analyzer` (vacuity check, mock drift, integration sufficiency). Design was quorum-checked against `codex exec --sandbox read-only` at `xhigh` before implementation — it endorsed option **(b)** over (a)/(c) and the shared-core refactor, and its findings on the signed `mdid` carry-forward and the Redis cutoff race are both fixed here.

**Findings:** code-reviewer 0. pr-test-analyzer 4 → 3 addressed in `984760b2a`, 1 declined:

- **Post-commit failure could destroy the codes (rated 8) — fixed.** `bindIssuedUserSession` / cookie install / audit-org lookup all run after the new hashes are durable and the old codes are gone; a throw returned 500 and took the only copy of the replacements with it. Both blocks are now reported and stepped over, the audit records `sessionInstalled`, and `tokens` is withheld on a failed install rather than selling the caller a few minutes on an unbindable refresh JTI. **Mutation-checked** — without the guard the request 500s.
- **Redis cutoff never proven against real Redis (rated 7) — fixed.** The integration test now asserts, against the suite's real Redis, that the cutoff is armed and the replacement token clears it. The comment is explicit that this does **not** discriminate the `preserveTokensIssuedAtOrAfter` clamp: same-second runs clear the default cutoff too, so the clamp's arithmetic stays pinned by `tokenRevocation.test.ts`.
- **Mock drift on plaintext↔hash correspondence — fixed.** The route test now asserts the hashes handed to the write are neither the plaintext codes nor a different count.
- **No full-HTTP-through-real-infra test (rated 5) — declined.** No such test exists for `/mfa/enable` either; standing one up needs password step-up + binding bootstrap + cookie plumbing and is its own piece of work. The three suites triangulate it: route plumbing (mocked service), service+DB+Redis (real), and the web adoption path.
- The "user disappeared" defensive throw (rated 6) is inside the transaction — it rolls everything back and commits nothing, and the 409 test already proves an error path returns no codes. Left as is.

Also mutation-checked the signed-binding control: reverting `carryForwardBinding(auth.token)` to `readMobileDeviceId(c)` fails the forged-header test.

**Tests:** API affected suites green — 240 passed across `auth.test.ts`, `mfaEnrollmentSession`, `tokenRevocation`, `authLifecycle`, `auth.passkeys`, `auth/phone` (`vitest run`, single-fork). Integration run against a **real** Postgres + Redis test stack: `mfaEnrollmentSession` 5 passed (3 pre-existing + 2 new rotation cases), plus `mfaAssurance` / `pendingMfaEpoch` / `auth-browser-transition` 14 passed. Web: 195 passed across ProfilePage / MFASettings / auth-store, plus the new `ProfilePage.recoveryCodeRotation.test.tsx`. `tsc --noEmit` clean in `apps/api` and `apps/web`; eslint clean on every changed file. Stack torn down.

**Status:** review-clean, awaiting maintainer merge. Repo CI still running at time of writing — check it before merging.

PR #4649, opened 2026-09-03
test(api): guard integration-suite include/exclude coverage (#4522)

### LINKED ISSUE #4522: [CI] invoiceService.issue + invoicePdf integration tests are in no CI job
**Problem:** `apps/api/src/__tests__/integration/invoiceService.issue.integration.test.ts` and `invoicePdf.integration.test.ts` are not matched by any include list in `vitest.integration.config.ts`, so they run in no CI job. Surfaced in a follow-up comment on #3818 and confirmed out of scope for PR #4428.

**Fix:** add both to the integration config allowlist, confirm they appear in the shard log of the next CI run, and check whether any other `*.integration.test.ts` under that directory is similarly orphaned (a glob-vs-allowlist contract test would prevent recurrence).

Related: #3818, PR #4428.

ToddHebebrand:
## Summary

Issue #4522 reported that `invoiceService.issue.integration.test.ts` and `invoicePdf.integration.test.ts` were unmatched by any include list in `vitest.integration.config.ts`, so they'd run in no CI job.

**Investigation finding:** on current `main`, both files are already covered — the entries were added in the same PR that created them (`8066aedb0ba`, "Wave 6 (#3778)", merged 2026-08-23, well before this issue was filed). Verified with a real run against Postgres/Redis using the exact CI-equivalent command (`vitest run --config vitest.integration.config.ts`):

```
Test Files  2 passed (2)
     Tests  17 passed (17)
```

I also did the second half of the issue's ask — a full static sweep of every `*.integration.test.ts` file under `apps/api/src` against the config's actual `include`/`exclude` arrays. No orphans turned up beyond the already-documented, already-covered exceptions (dedicated runners for `rls`, `rls-coverage`, `site-scope-coverage`, `request-db-role`; the deliberately-disabled `auth.integration.test.ts` pending a separate follow-up audit; and `manifestSigning.integration.test.ts`, a mocked unit test despite its name that intentionally runs under the unit config).

## What this PR actually adds

The one thing genuinely missing, per the issue's own "Fix" section: *"a glob-vs-allowlist contract test would prevent recurrence."* Nothing previously asserted that the integration config's include/exclude lists stay in sync with the files that exist on disk — a new `*.integration.test.ts` file added under a path no include pattern matches could silently run in **no CI job, forever**, exactly as this issue alleged (even though, as it turns out, these two files weren't actually in that state).

- `apps/api/src/__tests__/integration/integration-suite-coverage.integration.test.ts` — pure static analysis. Imports `vitest.integration.config.ts`'s own `test.include`/`test.exclude` arrays (not a hand-copied duplicate — can't drift from the real config), walks `src/**/*.integration.test.ts` from disk, and asserts every file is either matched by the config or present in a documented `KNOWN_OUTSIDE_INTEGRATION_SUITE` allowlist with a justifying comment. A second assertion catches stale allowlist entries (files that no longer exist).
- `apps/api/vitest.config.integration-suite-coverage.ts` — dedicated runner (no `setupFiles`, no DB), following the exact precedent of `vitest.config.site-scope-coverage.ts` / `vitest.config.rls-coverage.ts`.
- `apps/api/vitest.integration.config.ts` — excludes the new coverage test itself from the general integration glob (so it doesn't get hooked to `setup.ts`'s real-postgres-pool TRUNCATE cycle).
- `apps/api/package.json` — new `test:integration-suite-coverage` script.
- `.github/workflows/ci.yml` — wired into the fast, blocking `test-api` job right next to the sibling `site-scope-coverage` step (pure static analysis, no DB/Redis needed, so it doesn't belong in the slower `integration-test` job).

**Verified red-first:** manually dropped a throwaway orphan file under `src/services/`, confirmed the test fails with the exact orphan path named in the assertion message, then removed it and confirmed green again.

## Test plan

- [x] `vitest run --config vitest.integration.config.ts src/services/invoiceService.issue.integration.test.ts src/services/invoicePdf.integration.test.ts` against real Postgres/Redis — 2 files / 17 tests passed.
- [x] `vitest run --config vitest.config.integration-suite-coverage.ts` — 2/2 passed.
- [x] Red-first proof: temporarily added an orphan `*.integration.test.ts` file, confirmed the contract test fails and names it; removed it, confirmed green.
- [x] `tsc --noEmit` clean.
- [x] `eslint` clean on all changed/new files.

Closes #4522

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,6 +220,20 @@ jobs:
       - name: Run site-scope coverage contract test
         run: pnpm --filter=@breeze/api test:site-scope-coverage
 
+      # #4522 — invoiceService.issue.integration.test.ts and
+      # invoicePdf.integration.test.ts were reported as unmatched by any
+      # include list in vitest.integration.config.ts (they run in the
+      # blocking `integration-test` job today, but nothing asserted that
+      # the config's include/exclude lists stay in sync with the files
+      # that actually exist on disk — a newly-added *.integration.test.ts
+      # file could silently fall out of coverage forever). This contract
+      # test is pure static analysis (reads the integration config's own
+      # include/exclude arrays, walks `src/**/*.integration.test.ts` from
+      # disk, no DB/Redis), so it runs here in the fast, blocking job
+      # rather than `integration-test`.
+      - name: Run integration-suite coverage contract test
+        run: pnpm --filter=@breeze/api test:integration-suite-coverage
+
       # RLS session-context contract test (vitest.config.rls.ts). Gates the
       # APPLICATION half of the tenant-isolation contract: that
       # withDbAccessContext stamps exactly the `breeze.*` GUCs the RLS helpers
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -29,6 +29,7 @@
     "test:docker:up": "docker compose -f ../../docker-compose.test.yml up -d --wait",
     "test:integration": "vitest run --config vitest.integration.config.ts",
     "test:integration:watch": "vitest --config vitest.integration.config.ts",
+    "test:integration-suite-coverage": "vitest run --config vitest.config.integration-suite-coverage.ts",
     "test:rls": "vitest run --config vitest.config.rls.ts",
     "test:rls-coverage": "vitest run --config vitest.config.rls-coverage.ts",
     "test:request-db-role": "vitest run --config vitest.config.request-db-role.ts",
--- a/apps/api/src/__tests__/integration/integration-suite-coverage.integration.test.ts
+++ b/apps/api/src/__tests__/integration/integration-suite-coverage.integration.test.ts
@@ -1,0 +1,149 @@
+import { describe, it, expect } from 'vitest';
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import integrationConfig from '../../../vitest.integration.config';
+
+/**
+ * Contract test for #4522: every `*.integration.test.ts` file under `src/`
+ * must be reachable by SOME CI job, or explicitly and visibly documented as
+ * intentionally excluded.
+ *
+ * `invoiceService.issue.integration.test.ts` and `invoicePdf.integration.test.ts`
+ * were briefly reported as unmatched by `vitest.integration.config.ts`'s
+ * include list (they were in fact added in the same PR that created them —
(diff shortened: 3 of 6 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer), both `model: sonnet`

**Findings:** 2 Important raised (both independently confirmed, both addressed in `ffd605805`), 1 Suggestion addressed, 1 Suggestion left as-is (documented below); 0 outstanding.

- **Important (code-reviewer):** the contract test only checked disk→config, never the reverse. Confirmed real, not hypothetical: `src/extensions/twoReplicaReconcile.integration.test.ts` was still listed in both `vitest.integration.config.ts`'s include list and `vitest.config.ts`'s exclude list months after the file itself was deleted (#3488). Fixed: added a third assertion (every literal, non-glob include entry must resolve to an existing file), removed the two dead entries, verified red-first by temporarily reintroducing the stale entry (test failed, naming it) then removing it again (green).
- **Important (pr-test-analyzer):** `globToRegExp`'s two supported pattern shapes had no assertions locking down their own behavior in isolation. Fixed: added direct matcher unit cases for both the literal-path and `dir/**/*.ext` shapes.
- **Suggestion (code-reviewer):** undisclosed dotenv side effect from importing the real config module. Fixed: added a comment explaining the trade-off.
- **Suggestion (pr-test-analyzer, not addressed):** the contract test checks the vitest config's include/exclude arrays, not `.github/workflows/ci.yml` wiring itself — a future removal of the `test:integration-suite-coverage` CI step or drift between the package.json script and the config file would stay invisible to this test. Left as-is: this matches the existing precedent of its sibling contract tests (`site-scope-coverage`, `rls-coverage`), which have the same scope limit, and closing it would require parsing GitHub Actions YAML — disproportionate to this test's narrow job. Noting it here for visibility rather than silently dropping it.

**Tests:** `vitest run --config vitest.integration.config.ts` for the two originally-reported files against real Postgres/Redis — 2 files / 17 tests passed (proves they were already covered on `main`, unrelated to this PR). `vitest run --config vitest.config.integration-suite-coverage.ts` — 5/5 passed. Red-first verified twice (a synthetic orphan file, and the real stale `twoReplicaReconcile` entry). `tsc --noEmit` clean. `eslint` clean on all changed files.

PR #4651, opened 2026-09-03
fix(api/auth): a rejected MFA proof is 400 + a stable code, not 401 (#4470)

### LINKED ISSUE #4470: [API][Auth] MFA proof endpoints return 401 for a rejected code — indistinguishable from bearer expiry, so fetchWithAuth logs the user out mid-enrollment
**Verified.** `POST /auth/mfa/enable` (and the sibling proof endpoints) answer a wrong TOTP with **401** (`apps/api/src/routes/auth/mfa.ts:978`), the same status the bearer guard uses. `fetchWithAuth` (`apps/web/src/stores/auth.ts:1132`) hands any 401 to refresh-and-replay → `handleSessionExpired` → logout + redirect. That is the mechanism behind the silent enrollment collapse in #4413.

PR #4439 stops the bleeding client-side with `skipUnauthorizedRetry` on every MFA proof call. That is a stopgap: any new caller (mobile, portal, a future settings surface) that forgets the flag re-inherits the logout-on-typo behaviour.

**Durable fix:** a rejected factor proof is a request-validation failure, not an auth failure — return **400/422 with a stable error code** (e.g. `mfa_code_invalid`) and reserve 401 for bearer problems. Then drop the `skipUnauthorizedRetry` flags in the web client. Touches the login challenge path too (it currently *relies* on the 401 rendering correctly), so change both ends in one PR with tests on each client.

Surfaced by the #4413/#4414 fix (PR #4439); reported by internal UI QA follow-up 2026-09-01.

ToddHebebrand:
Closes #4470

## The bug

`POST /auth/mfa/enable` — and every sibling proof endpoint — answered a wrong TOTP with **401**, the same status the bearer guard uses. `fetchWithAuth` (`apps/web/src/stores/auth.ts`) hands any 401 to refresh-and-replay → `handleSessionExpired` → logout + redirect. One mistyped digit signed the user out mid-enrollment (the mechanism behind #4413/#4414).

PR #4439 stopped the bleeding client-side with `skipUnauthorizedRetry` on the MFA calls. That flag is a trap: **any new caller that forgets it re-inherits the logout.** This is the server-side fix it was waiting for.

## The contract

On `/auth/mfa/*` and `/auth/phone/*`, a **rejected proof** — the TOTP/SMS/recovery code or the step-up password the user typed *into the request body* — now answers **400** with `{ error, message, code }`:

| `code` | meaning |
|---|---|
| `mfa_code_invalid` | wrong TOTP / SMS / recovery code |
| `mfa_proof_invalid` | rejected `POST /auth/mfa/step-up` factor proof (TOTP, SMS, passkey) |
| `invalid_credentials` | rejected step-up password or enrollment grant |

**401 is reserved** for "the credential authenticating *this* request is dead":

- the bearer guard (`authMiddleware`),
- the login challenge's `tempToken` — `{ error: 'Invalid or expired MFA session' }`,
- `{ error: 'sso_link_expired' }`.

Those bodies are **unchanged**, so the login page can still tell "retype your code" from "start the login over". A new test pins that: a dead `tempToken` still 401s.

Clients must branch on `code`, never on the human `error`/`message` text.

## Scope control (why this isn't a 7-route breaking change)

`requireCurrentPasswordStepUp` / `requireFreshMfaStepUp` / `resolveEnrollmentStepUp` are shared by 7 route files. They now take an opt-in `rejectionStatus` that **defaults to 401**, so `approvals`, `pam`, `users` (email change) and `authenticator` keep the contract their clients already key on — notably mobile `approvals.ts`, which maps a 401 decision to `STEP_UP_FAILED`. Only the **factor-management** routes opt in: `/auth/mfa/*`, `/auth/phone/*`, and the authenticated passkey routes.

Those unchanged callers **do** gain `message` + `code` in the body at the same 401. That is the one wire change outside the MFA routes.

Rejections stay **uniform per call site**: `enrollment_proof_required` follows the same status as its siblings, because a status that differed there would re-open the password-vs-passwordless oracle the #4018 opacity rule exists to close. It is the uniformity that closes the oracle, not the particular number.

## Client changes

- Dropped `skipUnauthorizedRetry` from the three MFA calls in the web app. A 401 from those endpoints now means *only* "your bearer expired" — which **should** refresh; the forced-enrollment page has no other way to get a token. A test pins both halves: a 400 never refreshes, a 401 still does.
- `ApproverDevicesSection` branched on `status === 401` + a string match on `'Invalid credentials'` to pick its tier-specific copy ("Incorrect code." / "Incorrect password." / passkey). `/auth/mfa/step-up` moving to 400 would have silently degraded that to the raw server string. It now branches on the stable `code` (plumbed through `RegisterStepError`) and keeps the 401 branch for `/authenticator/register-grant`, which has **not** migrated.
- `authenticator.ts` keeps `skipUnauthorizedRetry` on `/auth/mfa/step-up` — justified now by the reason the flag actually documents (the passkey body carries a single-use WebAuthn assertion), not by the overloaded status.

Also fixes the same latent logout on **`/auth/mfa/setup`, `/auth/phone/verify`, `/auth/phone/confirm`, `/auth/mfa/sms/enable`, `/auth/passkeys/register/options`, `/auth/passkeys/register/verify` and `DELETE /auth/passkeys/:id`** — none of which ever carried the client-side flag, so a wrong password there logs users out on `main` today.

The **login-time** passkey routes (`/auth/mfa/passkey/options`, `/auth/mfa/passkey/verify`) deliberately keep their 401s: pre-session, `tempToken`-scoped, and reached with a plain `fetch` that has no refresh path to mislead.

## Design note

Consequential API-surface change, so it went through the advisor quorum (CLAUDE.md): independent read-only Codex review agreed on 400-over-422 (matches `zValidator`'s existing 400), on the `{error, message, code}` body, on keeping the login-path `tempToken` rejection at 401, and on the opt-in default for the shared helpers.

Codex also ranked a transport-level discriminator higher long-term — `authMiddleware` emitting `WWW-Authenticate: Bearer` and `fetchWithAuth`/`runAction` refreshing only on *that*, which would immunise every remaining application-level 401 in the app (approvals, PAM, passkey enrollment). **Deliberately not in this PR**: it inverts the refresh path app-wide, and if the header is ever lost (CORS `Access-Control-Expose-Headers`, a proxy) token refresh silently stops working everywhere — a worse failure than the bug being fixed. Worth its own issue.

## Audit-trail note

The global fallback-audit middleware (`apps/api/src/index.ts`) classifies 401/403 as `denied` and other 4xx as `failure`, so these rejections shift `denied` → `failure` on the *fallback* row. The explicit `writeAuthAudit` rows these handlers already emit (`auth.mfa.setup.failed` / `auth.mfa.disable.failed`, reason `invalid_mfa_code`) were and remain `failure`, and nothing alerts on the fallback `denied` (the only `result = 'denied'` query in the repo is scoped to `action = 'agent.enroll'`).

## Docs

`apps/docs/.../users-and-roles.mdx` carried a public `<Aside type="caution">` telling integrators that a wrong code and an expired session **share** a 401 and must be told apart by message text. That is now false and is rewritten to document the split plus the `code` values.

## Tests

- **New #4470 contract block** (`auth.test.ts`, 8 tests): wrong TOTP on enable, wrong password on enable/setup/recovery-codes, wrong code on disable, rejected step-up proof — each asserting the exact `code` — plus the negative control that a dead `tempToken` still answers 401.
- **New login-path test** for `/auth/mfa/verify` with a wrong TOTP. That branch had **no coverage at all** before. It also asserts the pending challenge **survives** the typo and the same `tempToken` completes on the correct code.
- **New `ApproverDevicesSection` case** for the 400 → tier-specific-copy mapping.
- **New web-store tests**: a 400 never refreshes and never sets `sessionExpiredReason`; a genuine 401 from the same endpoint still refreshes and replays.
- `recoveryCode.integration.test.ts` identical-code race updated (loser is now 400 `mfa_code_invalid`).
- 26 pre-existing assertions moved to the new statuses/bodies. **No assertion was weakened, skipped, or deleted** — every `toEqual` stayed a `toEqual` and gained the new fields explicitly.

## Review round (findings addressed in `e6244a7`)

Two independent Sonnet reviews ran — one adversarial security lens (enumeration/opacity, skipped side effects, rate limiting, disclosure, replay safety) and one correctness/silent-failure lens.

Security review: **no HIGH findings.** It confirmed the #4018 opacity rule survives (every `resolveEnrollmentStepUp` exit shares one status per call site, `enrollment_proof_required` included), the C1 swapped-phone defense on `/mfa/step-up` still returns an identical response for "no active SMS factor" and "wrong code", every converted site still returns early / still audits / still refuses to burn a grant, TOTP time-step, recovery code or pending-MFA key, rate-limiter ordering is unchanged, and no unmigrated call site gained a new discriminator.

Correctness review raised **one HIGH, now fixed**: `passkeys.ts` still 401'd rejected proofs on the three authenticated passkey routes, which `ProfilePage` calls through `fetchWithAuth` completely unguarded — the exact bug this PR exists to remove. Migrated, plus three LOWs (uniform `requireFreshMfaStepUp` body, the now-unnecessary `/auth/mfa/step-up` client opt-out, and stale `openapi.ts` response lists).

One review note deliberately **not** actioned: `ApproverDevicesSection`'s catch-all for "400 with no `code`" surfaces the server string rather than localized copy. No live path reaches it (every 400 from those two mint endpoints sets `code`), and for the ones that could — a zod validation 400, the `agent_rollback` binding 400 — showing the server's message is the correct behaviour, not a degradation.

**Verification (after the review round):** `apps/api` auth-adjacent suites **947/947** pass; `apps/web` stores + settings + lib **2321/2321** pass; `tsc --noEmit` clean in both `apps/api` and `apps/web`; `pnpm lint` clean. Verified on a base with current `origin/main` (7d5ebb4f3) merged in.

## Overlap with #4480

A sibling fix is in flight on #4480 (recovery-code regeneration + `mfa_epoch`). Contact points, both small:

- `POST /auth/mfa/recovery-codes` — this PR touches only its `requireCurrentPasswordStepUp` call (adds the `rejectionStatus` option); the rotation body is untouched.
- `mfa.ts` recovery-code rejection sites (`RecoveryCodeInvalidError`) — status/body only, no logic change.

Whichever lands second should expect a trivial conflict in those two spots.

### DIFF
--- a/apps/api/src/__tests__/integration/recoveryCode.integration.test.ts
+++ b/apps/api/src/__tests__/integration/recoveryCode.integration.test.ts
@@ -101,7 +101,7 @@ describe('POST /auth/mfa/verify (method: recovery) — real-PG single-use concur
     tempTokens.length = 0;
   });
 
-  it('identical-code race: exactly one 200, one 401, persisted array has exactly 2 hashes (no double-spend)', async () => {
+  it('identical-code race: exactly one 200, one 400, persisted array has exactly 2 hashes (no double-spend)', async () => {
     const userId = await seedUserWithRecoveryCodes();
     const { getRedis } = await import('../../services');
     const redis = getRedis();
@@ -118,8 +118,12 @@ describe('POST /auth/mfa/verify (method: recovery) — real-PG single-use concur
 
     const [r1, r2] = await Promise.all([verify(app, tokenX, CODE_A), verify(app, tokenY, CODE_A)]);
 
+    // #4470: the loser's rejection is a REJECTED PROOF (its code is gone), so
+    // it answers 400 `mfa_code_invalid` — not the bearer guard's 401.
     const statuses = [r1.status, r2.status].sort((a, b) => a - b);
-    expect(statuses).toEqual([200, 401]);
+    expect(statuses).toEqual([200, 400]);
+    const loser = r1.status === 400 ? r1 : r2;
+    expect(await loser.json()).toMatchObject({ code: 'mfa_code_invalid' });
 
     const remaining = await readRecoveryCodes(userId);
     expect(remaining).toHaveLength(2);
--- a/apps/api/src/openapi.ts
+++ b/apps/api/src/openapi.ts
@@ -1044,6 +1044,10 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
               }
             }
           },
+          // #4470: a rejected proof (wrong code) is 400 with a stable `code`
+          // (`mfa_code_invalid`); 401 is reserved for a dead credential —
+          // an invalid/expired tempToken or bearer.
+          '400': { $ref: '#/components/responses/BadRequest' },
           '401': { $ref: '#/components/responses/Unauthorized' },
           '429': { $ref: '#/components/responses/TooManyRequests' }
         }
@@ -1356,6 +1360,10 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
               }
             }
           },
+          // #4470: a rejected proof (wrong code) is 400 with a stable `code`
+          // (`mfa_code_invalid`); 401 is reserved for a dead credential —
+          // an invalid/expired tempToken or bearer.
+          '400': { $ref: '#/components/responses/BadRequest' },
           '401': { $ref: '#/components/responses/Unauthorized' },
           '429': { $ref: '#/components/responses/TooManyRequests' }
         }
@@ -1451,6 +1459,10 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
               }
             }
           },
+          // #4470: a rejected proof (wrong code) is 400 with a stable `code`
+          // (`mfa_code_invalid`); 401 is reserved for a dead credential —
+          // an invalid/expired tempToken or bearer.
+          '400': { $ref: '#/components/responses/BadRequest' },
           '401': { $ref: '#/components/responses/Unauthorized' },
           '429': { $ref: '#/components/responses/TooManyRequests' }
         }
--- a/apps/api/src/routes/auth.passkeys.test.ts
+++ b/apps/api/src/routes/auth.passkeys.test.ts
@@ -561,7 +561,37 @@ describe('passkey MFA auth routes', () => {
       body: JSON.stringify({ currentPassword: 'wrong-password' }),
     });
 
-    expect(res.status).toBe(401);
+    expect(res.status).toBe(400);
+    expect(passkeyMocks.generatePasskeyRegistrationOptions).not.toHaveBeenCalled();
+  });
+
+  // #4470: the two statuses above are the whole point, so pin them apart.
+  // `ProfilePage.handleAddPasskey` calls this endpoint through `fetchWithAuth`
+  // with no 401 opt-out, so while a wrong step-up password answered 401 it was
+  // fed to refresh-and-replay and could sign the user out of the settings page
+  // for a typo. A MISSING bearer must still be the 401 that refreshes.
+  it('#4470: a wrong step-up password is 400 invalid_credentials, a missing bearer is still 401', async () => {
+    vi.mocked(verifyPassword).mockResolvedValueOnce(false);
+    dbState.selectQueue.push([{ passwordHash: '$argon2id$hash' }]);
+
+    const rejected = await app.request('/auth/passkeys/register/options', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json', Authorization: 'Bearer access-token' },
+      body: JSON.stringify({ currentPassword: 'wrong-password' }),
+    });
+    expect(rejected.status).toBe(400);
+    expect(await rejected.json()).toEqual({
+      error: 'Invalid credentials',
+      message: 'Invalid credentials',
+      code: 'invalid_credentials',
+    });
+
+    const noBearer = await app.request('/auth/passkeys/register/options', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ currentPassword: 'correct-password' }),
+    });
+    expect(noBearer.status).toBe(401);
     expect(passkeyMocks.generatePasskeyRegistrationOptions).not.toHaveBeenCalled();
   });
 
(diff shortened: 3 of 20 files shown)


ToddHebebrand:
**Review run:** two independent reviewers on the first commit — an adversarial **security** lens (enumeration/opacity, skipped side effects, rate limiting, information disclosure, replay safety) and a **correctness / silent-failure** lens — then a third targeted **re-review** of the follow-up commit.

**Findings:** 4 raised, all addressed.

- **HIGH — `passkeys.ts` reproduced the very bug this PR removes.** `/passkeys/register/options`, `/passkeys/register/verify` and `DELETE /passkeys/:id` still 401'd a rejected step-up password / stale challenge / failed attestation, and `ProfilePage` calls all three through `fetchWithAuth` with no opt-out and no status branch. My original scope note claimed `passkeys.ts` was excluded because "its clients key on 401" — false for these three. Migrated in `e6244a7`; the PR body is corrected.
- **LOW — `requireFreshMfaStepUp` body shape.** It was the last step-up helper emitting a bare `{ error }`, so `approvals`/`pam` answered two different shapes for the same gate depending on the road taken. Now uniform at the unchanged 401.
- **LOW — the `/auth/mfa/step-up` client opt-out was no longer justified.** Dropped in `e6244a7`. The re-review specifically verified the load-bearing claim: every 401 that endpoint can emit originates in `authMiddleware` *before* the handler runs, so the WebAuthn assertion in the body is never already-burned and refresh-and-replay is correct. Keeping the flag would have signed the user out for an access token that merely aged out mid-ceremony.
- **LOW — `openapi.ts`** listed only 401 for three operations that now also answer 400.

Security review returned **no HIGH findings**, and confirmed positively: the #4018 opacity rule survives (every `resolveEnrollmentStepUp` exit shares one status per call site, `enrollment_proof_required` included); the C1 swapped-phone defence on `/mfa/step-up` still answers identically for "no active SMS factor" and "wrong code"; every converted site still returns early, still audits, and still refuses to burn a grant, TOTP time-step, recovery code or pending-MFA key; rate-limiter ordering is unchanged; and no unmigrated call site gained a new discriminator.

**Not actioned, deliberately:**
- Two more endpoints 401 a rejected body-supplied password with the same unguarded `fetchWithAuth` callers — `POST /auth/change-password` and `POST /auth/account-deletion-request`. Same bug class, but they are not MFA proof endpoints and changing password-change / account-deletion semantics deserves its own review. Filed as **#4660** with the fix sketch, plus the `users.ts` / `authenticator.ts` stragglers and a `PasskeyChallengeError` accuracy bug (it conflates "Redis unavailable" with "challenge expired" — pre-existing, both shared the old 401).
- `ApproverDevicesSection`'s catch-all for "400 with no `code`" surfaces the server string rather than localized copy. No live path reaches it, and for the ones that could (a zod 400, the `agent_rollback` binding 400) showing the server's message is correct, not a degradation.

**Tests:** `apps/api` auth-adjacent suites **947/947**; `apps/web` stores + settings + lib **2321/2321**; `tsc --noEmit` clean in `apps/api` and `apps/web`; `pnpm lint` clean. 26 pre-existing assertions moved to the new statuses/bodies — every reviewer independently confirmed **none was weakened, skipped, or deleted**.

**CI:** all **67** checks green on `9ca5489`, including **Test API**, **Test Web**, **Lint**, **Security Audit** and all four **Integration Tests** shards.

**Status:** review-clean, awaiting maintainer merge. Not merged and #4470 left open per the hand-off rule.

PR #4658, opened 2026-09-03
test(web): scope AiUsagePage test queries to their own render container (#4601)

### LINKED ISSUE #4601: flaky: AiUsagePage alert threshold ladder fails in full Test Web run, passes in isolation
**What happened**
`Test Web` failed on PR #4593 (API-only migration diff, no web changes) with exactly one failure:

```
FAIL src/components/settings/AiUsagePage.test.tsx > AiUsagePage alert threshold ladder > sends alertThresholdPercents: nu…
Test Files 1 failed | 698 passed (699)
```
Run: https://github.com/LanternOps/breeze/actions/runs/33677130178/job/100404821679

**Evidence it is a flake, not a regression**
- The same test passes in isolation on current main (3d7dcfd72): `cd apps/web && npx vitest run src/components/settings/AiUsagePage` → 14/14.
- `gh run rerun --failed` on the same merge commit went green (67 pass).
- Earlier today a separate full `apps/web` run showed a one-off failure in `VulnerabilityTable.test.tsx` that also did not reproduce in isolation, so this looks like a suite-order / shared-state interaction in the full web run rather than this test alone.

**Suspect**
The test was added with the AI budget alert thresholds UI (#4388 follow-ups, merged 2026-09-02, ce3f8f1db). Likely candidates: an un-reset mock queue (`vi.clearAllMocks` does not drop queued `mockResolvedValueOnce` implementations) or a shared `showToast`/fetch mock leaking from a neighbouring file.

**Ask**
Make the assertion independent of suite order (reset queued mock implementations in `beforeEach`, or isolate the module), and confirm by running the full `apps/web` suite three times.

Refs #4388.

ToddHebebrand:
## Summary

Fixes the flaky `AiUsagePage alert threshold ladder` test reported in #4601 (`Test Web` failed once on PR #4593 with 1 of 699 files red; the same test always passed in isolation).

**Root cause (proven, not guessed):** every test in `AiUsagePage.test.tsx` queried the shared `screen` object, which binds to `document.body`, not to any one render. Under real CPU contention — the state a full `Test Web` run puts the host in — a preceding test's RTL `cleanup()` can still be settling when the next test's async `it()` body starts. When that happens, a `screen.findByTestId(...)` in the "next" test can transiently match the **prior** test's still-mounted, already-edited component instance instead of its own freshly-rendered one.

I reproduced this directly (not just theorized it): pinned the host CPU with several busy-loops on top of an already-loaded box, added temporary debug logging to `AiBudgetThresholdsInput`'s `commit()` and `AiUsagePage`'s `handleSaveBudget`, and caught the exact mechanism in the logs — the "sends the edited ladder" test's own `commit()` fired with `text=""` (the *previous* test's already-cleared value) instead of the `"60, 90"` it had just typed, because its `findByTestId` had matched the previous, not-yet-unmounted instance. The console output also showed describe blocks interleaving out of file declaration order under pressure, confirming genuine cross-test overlap rather than a same-test logic bug.

**Fix:** every test now scopes its queries to its own render's container via `within(container)` (wrapped in a small `renderPage()` helper) instead of the global `screen`. A query can then never match another test's DOM, regardless of how fast (or slow) a neighbouring test's teardown runs — independent of suite order, as the issue asked for.

Per repo precedent (#3277), I deliberately did **not** raise any `waitFor`/test timeout — that was tried twice for a similar-looking flake in `InvoiceWorkspace.test.tsx` and never held, because the underlying condition never became true (see the comment on that file). This fix instead removes the actual cross-instance query ambiguity at its source.

**Honesty on residual risk:** under the same injected CPU pressure, the failure rate dropped from readily reproducible (pre-fix) to 3/120 runs post-fix — one was the same PUT-body-mismatch symptom, two were a different "element not found" flavor. That pressure (6 saturating busy-loops stacked on an already-heavily-loaded 18-core dev box) is more extreme than a real CI runner sees, so this is a substantial, evidence-based improvement rather than a mathematically-proven elimination of all possible timing windows.

## Test plan
- [x] `AiUsagePage.test.tsx` in isolation: 14/14 passed
- [x] `AiUsagePage.test.tsx` under injected CPU pressure, 120 runs: reliably reproduced the original failure pre-fix; 3/120 residual failures post-fix (down from readily reproducible)
- [x] Full `apps/web` suite run 1 (normal load, `--pool=threads --maxWorkers=2`): 700/700 files, 7379/7379 tests passed
- [x] Full `apps/web` suite run 2 (same config): 700/700 files, 7379/7379 tests passed
- [x] `npx tsc --noEmit` on `apps/web`: clean
- [x] `npx eslint` on the changed file: clean

Closes #4601

### DIFF
--- a/apps/web/src/components/settings/AiUsagePage.test.tsx
+++ b/apps/web/src/components/settings/AiUsagePage.test.tsx
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { fireEvent, render, screen, waitFor, act } from '@testing-library/react';
+import { fireEvent, render, within, waitFor, act } from '@testing-library/react';
 
 const fetchWithAuth = vi.fn();
 vi.mock('../../stores/auth', () => ({ fetchWithAuth: (...a: unknown[]) => fetchWithAuth(...a) }));
@@ -7,6 +7,20 @@ const orgState: { currentOrgId: string | null } = { currentOrgId: null };
 vi.mock('../../stores/orgStore', () => ({ useOrgStore: () => ({ currentOrgId: orgState.currentOrgId }) }));
 
 import AiUsagePage from './AiUsagePage';
+
+// Every test below scopes its queries to its OWN render's container via
+// `within(container)` instead of the shared `screen` (which binds to
+// `document.body`). Under real suite-order/CPU-load conditions a previous
+// test's `cleanup()` can still be settling when this test's body starts (both
+// are async and RTL's afterEach isn't guaranteed to have unmounted yet), so a
+// global `screen` query can silently match the PRIOR test's still-mounted
+// instance instead of this one. Scoping to `container` makes every query
+// order-independent regardless of how fast a neighbouring test's teardown
+// runs (#4601).
+function renderPage() {
+  const { container, ...utils } = render(<AiUsagePage />);
+  return { container, ...utils, ...within(container) };
+}
 
 function jsonRes(body: unknown, status = 200) {
   return {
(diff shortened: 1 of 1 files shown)


ToddHebebrand:

**Findings:**
- code-reviewer: 0 critical/important. 2 sub-threshold stylistic suggestions (missing one-line JSDoc summary on `renderPage()`; could cite `Toast.test.tsx`'s existing `within(container)` precedent) — neither actioned, below the reporting bar. Verdict: **clean**, verified the fix live (checked out `pull/4658/head`, ran the file, 14/14 pass) and confirmed by construction that container-scoping eliminates the cross-test DOM race structurally (not just narrows its probability).
- pr-test-analyzer: 0 critical/important. Flagged that of the 3/120 residual failures observed under injected CPU pressure, 2 carried a **different** signature (`ai-budget-thresholds-error` not found) than the one this PR fixes, and recommended filing a follow-up rather than treating #4601 as fully closed. Addressed: filed https://github.com/LanternOps/breeze/issues/4659 with the repro conditions and stack trace.

**Tests:** `AiUsagePage.test.tsx` in isolation 14/14 passed. Full `apps/web` suite run twice (`--pool=threads --maxWorkers=2`, normal/non-injected load): 700/700 files, 7379/7379 tests, both green. `npx tsc --noEmit` clean. `npx eslint` on the changed file clean.

**Status:** review-clean, follow-up filed for the distinct residual signature (#4659), awaiting maintainer merge.

PR #4683, opened 2026-09-03
fix(tickets): scope AI draft card action lock to its own draft id (#4469)

### LINKED ISSUE #4469: AI agents: ticket draft cards share one in-flight lock
On the ticket workbench (wave P2-4), the AI-draft cards (reply draft, resolution-note draft) share a single in-flight action lock, so acting on one draft (send/discard) briefly disables the other even though they're independent operations.

Source: PR #4301 body ("Follow-ups").

ToddHebebrand:
## Summary
- The ticket workbench's reply-draft and resolution-note-draft AI cards are independent (`ticket_drafts_active_uq`: at most one active `reply` + one active `resolution_note` per ticket), but `sendAiDraft`/`discardAiDraft` guarded on "any in-flight action" (`if (sendingDraftId || discardingDraftId) return;`) instead of the specific draft being acted on.
- Effect: clicking discard on the resolution_note card while the reply card's send was in flight silently no-opped — the button wasn't visually disabled (its `disabled` prop was already correctly scoped per draft id), but the handler's own guard clause returned before the fetch ever fired.
- Fix: scope both guards to `draft.id` (`sendingDraftId === draft.id || discardingDraftId === draft.id`) so the two cards can be acted on concurrently, matching the already-correct `disabled` UI state.

Closes #4469

## Test plan
- [x] Added a regression test (`#4469: discarding one draft while the other is mid-send still fires the discard request`) that holds the reply card's send request open and asserts the resolution_note card's discard POST still fires — confirmed **red** against the pre-fix code (discard fetch never called), then **green** after the fix.
- [x] `cd apps/web && npx vitest run src/components/tickets/TicketWorkbench.test.tsx` — 102 passed.
- [x] `cd apps/web && npx tsc --noEmit -p .` — clean.

### DIFF
--- a/apps/web/src/components/tickets/TicketWorkbench.test.tsx
+++ b/apps/web/src/components/tickets/TicketWorkbench.test.tsx
@@ -935,6 +935,54 @@ describe('TicketWorkbench AI drafts (#4191, Task 11)', () => {
     resolveTk2Drafts(makeJsonResponse({ data: [resolutionDraft] }));
     await screen.findByTestId('ticket-ai-draft-resolution_note');
   });
+
+  // #4469 — the two draft cards are independent operations (different draft
+  // ids); acting on one must not block the other. Holds the reply's send
+  // request open (never resolved during the assertion) so the resolution_note
+  // card's discard click happens while sendingDraftId is still set to the
+  // OTHER draft's id — a regression that guards on "any in-flight action"
+  // rather than "this draft's own in-flight action" would silently swallow
+  // the discard click here (the button isn't visually disabled, but the
+  // handler's guard clause no-ops before the fetch fires).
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (Sonnet)
**Findings:** 0 — guard-scoping fix verified correct (no new double-submit or stale-closure risk; matches the already-correct `disabled` UI scoping), and the new regression test confirmed to be a genuine red-before/green-after check (deferred-promise mock keeps the reply send in flight while asserting the resolution_note discard POST still fires).
**Tests:** `cd apps/web && npx vitest run src/components/tickets/TicketWorkbench.test.tsx` — 102 passed. `npx tsc --noEmit -p .` — clean.

PR #4585, opened 2026-09-02
feat(billing): contract lines billed by device role (#3205)

ToddHebebrand:
Closes #4600 (wave W01 of feature #3205; the parent closes via feature-lifecycle when its waves are done). Refs #3205. Device groups are tracked in #4584.

Spec: `docs/superpowers/specs/billing/2026-09-02-contract-lines-per-device-role-design.md`
Plan: `docs/superpowers/plans/billing/2026-09-02-contract-lines-per-device-role.md`

## What

- New `per_device_role` contract line type that bills a **set** of device roles (`contract_lines.device_roles text[]`, CHECK-enforced against the eleven billable roles; `unknown` is never billable).
- All device-counted quantities on a contract (existing `per_device` too) now come from **one grouped snapshot per org**, so a role change mid-generation cannot double-bill or skip a device. The estimate's quantities and its coverage warning are computed from the same snapshot by construction.
- The estimate endpoint and invoice generation report **uncovered devices** (devices no line bills, by role). Editor and detail pages show it; the generate dialog toasts it; the billing worker warns on it. Auto-issue still proceeds.
- **Site ownership** on `contract_lines.site_id` is now enforced: composite FK to `sites(id, org_id)` with `ON DELETE SET NULL (site_id)`, plus a `SITE_NOT_IN_ORG` 400 from both writers. Previously a foreign-org site was accepted and silently counted zero.
- `DEVICE_ROLES` moved to `packages/shared/src/validators/deviceRoles.ts` (package exports unchanged) with `BILLABLE_DEVICE_ROLES` beside it; the web mirror gains the same constant.
- AI `manage_contracts` tool description documents the new line shape. Docs table updated.

## Migrations

- `2026-10-05-100000-contract-line-type-per-device-role.sql`: enum value only (own transaction).
- `2026-10-05-100100-contract-lines-device-roles.sql`: column, CHECK, foreign-org site cleanup (row count logged), composite FK replacing `contract_lines_site_fkey`.

Both idempotent; named past main's newest (`2026-10-04-100000-ticket-requester-contact.sql`).

## Verification

- Typecheck: api 0, web 0, shared 0 (excluding three pre-existing errors in `quotes.test.ts` that are unchanged on main). `pnpm lint` clean.
- Unit: shared 2306, api 423 (contract/coverage/worker/aiTools/routes/autoMigrate), web 266 (contracts components + locale parity).
[...]

Issue #3207, opened 2026-08-07
End-user reboot prompts with deferral on Windows, macOS, and Linux


ToddHebebrand:
## Description

There is no end-user reboot notification with a deferral option. Users cannot postpone a patch reboot, and on macOS and Linux they are not notified at all.

Depends on the silent-reboot defect being fixed first — today the warnings that do exist never fire.

## Current state

**Agent can prompt (Windows only).** `agent/internal/patching/reboot_windows.go` has a `RebootManager` with staged warnings at 60/15/5 minutes plus a "Rebooting Now" toast, wired via `heartbeat.go:785` -> `sessionBroker.BroadcastNotification` -> `broker.go:1281` -> IPC `notify` -> `agent/internal/userhelper/notify_windows.go` (PowerShell WinRT toast, AUMID `Breeze.Agent`).

`agent/internal/patching/reboot_other.go` is a no-op stub — macOS and Linux get nothing; Linux relies on the `shutdown -r +15` wall message.

**No deferral anywhere.** `ipc.NotifyRequest.Actions` and `ipc.NotifyResult.ActionClicked` exist (`agent/internal/ipc/message.go:245,251`) but **no platform notifier implements actions** — the Windows toast XML is `ToastText02` with no `<actions>` block, so buttons are impossible today. `BroadcastNotification` is also fire-and-forget, with no response path. The only existing protection is a circuit breaker (`maxRebootsPerDay`, default 3, persisted to `reboot_history.json`).

**Policy has reboot behavior but no notification settings.** `config_policy_patch_settings.reboot_policy` (`never|if_required|always|maintenance_window`) at `apps/api/src/db/schema/configurationPolicies.ts:201`. Delay is hardcoded. The web UI (`PatchTab.tsx:678-712`) renders only the four reboot-policy radios.

`config_policy_maintenance_settings.notifyBeforeMinutes` / `notifyOnStart` exist in DB and UI with no runtime consumer.

## Reuse vehicles

Three existing paths can show UI to the logged-in user:

- **Toast:** `sessionbroker.BroadcastNotification` / heartbeat `notify_user` (`handlers_user.go`)
- **Blocking native dialog with countdown and buttons:** `agent/internal/userhelper/consent.go` + `consent_dialog_windows.go` (MessageBoxTimeoutW), `_darwin.go` (osascript with `giving up after`), `_linux.go` (zenity). **This is the closest analogue** — notify + user choice + timeout default, already cross-platform.
[...]

PR #4686, opened 2026-09-03
feat(patching): reboot deferral policy settings (#3207 W1)

### LINKED ISSUE #4661: W01: Policy surface: reboot deferral settings on patch policy, migration, export re-emit, PatchTab, docs (default off)
Wave W01 of #3207 — Policy surface: reboot deferral settings on patch policy, migration, export re-emit, PatchTab, docs (default off)
Branch: `feature/3207-reboot-deferral/wave-4661`

ToddHebebrand:
Closes #4661.

Wave W01 of #3207 — the **policy surface only**, wired end to end and **off by default**. No agent honours the new settings yet (that is W2/W3), so production behaviour is unchanged.

## What ships

| Layer | Change |
|---|---|
| Migration | `2026-10-05-100200-patch-reboot-deferral-settings.sql` — 3 columns + 2 CHECKs on `config_policy_patch_settings`, plus a full re-emit of `breeze_partner_export_policy_settings_pre_patch` **and its ACLs** |
| Schema | `rebootAllowDeferral` / `rebootMaxDeferrals` / `rebootDeferralMinutes` Drizzle columns |
| Validator | `patchInlineSettingsSchema` fields + two coherence refinements |
| Services | write path, read projection, patch inventory, backfill insert |
| Partner API | `PATCH_NORMALIZED_MATERIAL_KEYS` |
| Dispatch | `schedule_reboot` payload gains `deadline`, `allowDeferral`, `maxDeferrals`, `deferralMinutes` (both the patch path and the maintenance-window worker) |
| Web | `PatchTab` toggle + budget fields, i18n in all 8 locales |
| Docs | patch-policy settings table + the `schedule_reboot` payload contract |

## Defaults preserve today's behaviour, verified

`reboot_allow_deferral` defaults to `false`. With it off:

- `allowDeferral: false, maxDeferrals: 0, deferralMinutes: 0` go out on the wire and mean "not postponable".
- `deadline` is **exactly the scheduled reboot time** — which is what the agent already computes for itself when the key is absent (`handlers_patch.go`: `deadline := time.Now().Add(delay)`), so nothing changes for any shipped agent.
- **Verified, not assumed**, that extra payload keys are harmless: `Command.Payload` is `map[string]any` (`agent/internal/heartbeat/heartbeat.go:276`) and `handleScheduleReboot` reads named keys off it. `DisallowUnknownFields` is used by the peripheral / rollback / actuate / PAM handlers — **not** this one. No agent code was modified.
- The absent-key direction is safe by construction too: `GetPayloadBool(..., false)` — an absent `allowDeferral` can never read as enabled.

## Why the contract suites were touched (they are supposed to red)

The canonical partner export is a hand-enumerated `jsonb_build_object`, and `PATCH_NORMALIZED_MATERIAL_KEYS` **fails closed on an exact count mismatch** — SQL-without-TS or TS-without-SQL makes every `patch` partner export silently `blocked`. So the new columns necessarily red three key-list assertions, all updated here:

- `partnerApiConfigurationWatermark.integration.test.ts` — the `toEqual` settings shape.
- `routes/partnerApi/configuration.test.ts` — the `NORMALIZED_PATCH_FACTS` material fixture.
- `patchCanonicalExportParity.integration.test.ts` — mirrored whole-document parity cases for the new bounds and both new refinements.

The ACL re-emit at the end of the migration is likewise deliberate: `CREATE OR REPLACE` resets function ACLs, and the watermark suite asserts `patchPreMaterializerPublic: false, patchPreMaterializerApp: true`.

## Tenancy

`config_policy_patch_settings` has **no `org_id` and no `partner_id`** (confirmed against the live schema, not just the Drizzle file). Tenancy is transitive: `feature_link_id → config_policy_feature_links → configuration_policies`, which already carries `org_id XOR partner_id`, and `'patch'` is already in `PARTNER_LINKABLE_FEATURE_TYPES`. Consequently:

- Partner-Wide First is satisfied by inheritance — no new `one_owner_chk`, no new RLS policy, no `DUAL_AXIS_TENANT_TABLES` entry.
- No cascade or export-policy registration: the table appears in **none** of `tenantCascade.ts`, `tenantExportPolicyRegistry.ts`, `routes/devices/core.ts`, `orgMergeRegistry.ts` (grep-verified), because all of those key on `org_id`.

Rather than assert that, the parity suite now **proves** it: a new case builds a partner-wide policy (`org_id NULL`, `partner_id` set), assigns it to an org, and round-trips all three settings out through the partner export — asserting `blocked` is empty, so a missed allowlist entry can't hide behind a row that never arrived.

## Deviations from the plan

1. **Migration filename** is `2026-10-05-100200-…`, not the planned `…-100100-…`: open PR #4585 already claims `2026-10-05-100100-contract-lines-device-roles.sql`. W5 must re-check its own name at execution time (the plan's `…-100200-…` reservation is now taken).
2. **One policy walk, not two.** The plan had `executeReboot` call `resolveRebootDelayMinutes` and `resolveRebootDeferralSettings` separately — two hierarchy walks per dispatch. Both are still exported (W2/W3 use them), but they now delegate to a shared `resolveRebootPlan`, and `executeReboot` / the maintenance worker each resolve once. The existing "an explicit delay skips the policy lookup entirely" assertion became "exactly one lookup answers both", plus a new case proving zero lookups when the caller supplies both.
3. **No `deps` parameter on `executeReboot`.** The plan added one for test injection; the test file already injects through `vi.mock('./featureConfigResolver')`, so the parameter would be production-dead surface. `options` did gain `deferral` and `windowEndsAt`, which the maintenance path really does pass.
4. **`MaintenanceWindowStatus` gains `windowEndsAt: Date | null`** so a maintenance-window reboot's deadline is capped at the close of the window. Required rather than optional, which is why three unrelated test files gained a `windowEndsAt: null` fixture line.
5. **`computeRebootDeadline` never returns a deadline earlier than the scheduled reboot** — a window closing inside the warning period would otherwise describe a reboot that is late before it is due. Covered by its own test.
6. **`agents/commands.mdx` also updated** (the plan scheduled it for W5): the `schedule_reboot` payload contract changes in *this* wave, so leaving it would ship stale wire docs.
7. **The 7-day ceiling counts `rebootDelayMinutes`.** The plan's refinement bounded only `maxDeferrals x deferralMinutes`; `computeRebootDeadline` actually emits `delay + maxDeferrals x deferralMinutes`, so the check now sums the delay in and the docs say so. Found in review — see the review comment.
8. **`evaluateRebootPolicy` returns `windowEndsAt`.** The plan wired the maintenance-window deadline cap only into the maintenance sweep; the post-patch reboot path needed it too, or a `maintenance_window` patch policy shipped an uncapped deadline. Found in review — see the review comment.

## Test evidence

**Verified — ran locally, green:**

| Suite | Result |
|---|---|
| `packages/shared` full suite | 92 files / **2314 passed** |
| `apps/api` affected unit sweep (patchRebootHandler, maintenanceRebootWorker, patchJobExecutor, featureConfigResolver, partnerApi/configuration, patchJobService, scriptExecution ×2, configurationPolicy, configPolicyPatching, autoMigrate) | 24 files / **714 passed** |
| `apps/web` — configurationPolicies components + localeParity | 30 files / **364 passed** |
| Integration (private per-worktree stack) — `partnerApiConfigurationWatermark` + `patchCanonicalExportParity` | **20 passed** |
| Integration — `tenantCascade` + `tenant-export-policy` | 2 files / **7 passed** |
| RLS coverage (`vitest.config.rls-coverage.ts`) | **100 passed** |
| `tsc --noEmit` — `apps/api`, `apps/web` | clean |
| `pnpm db:check-drift` | no drift, 636 files match the ledger |
| Migration idempotency | applied the file twice by hand via `psql -v ON_ERROR_STOP=1`; second pass is a clean no-op |

One review round ran (three parallel reviewers: code-reviewer, silent-failure-hunter, pr-test-analyzer). 4 confirmed findings, all fixed in `1c618d99a`; 2 nits declined with reasons. Full write-up in the review comment on this PR. Counts above are the post-fix re-run.

**Not run:** the rest of the `apps/api` unit suite and the remaining integration shards (CI covers both); `packages/shared` `tsc` reports 3 pre-existing errors in `validators/quotes.test.ts`, untouched by this PR.


ToddHebebrand:
**Review run:** three independent reviewers in parallel — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` (one round).

**Findings:** 4 confirmed (1 must-fix, 3 should-fix) → all 4 fixed in `1c618d99a`; 2 nits declined with reasons; 0 outstanding.

**1 — must-fix: the maintenance-window deferral cap never applied on the patch path.** `evaluateRebootPolicy` computed the window close and discarded it, and `patchJobExecutor` had nothing to pass, so a `maintenance_window` patch policy with deferral enabled shipped an **uncapped** deadline — a user could postpone a patch restart well past the window. Silent: no log, no Sentry, and the *maintenance sweep* path capped it correctly, so the asymmetry would not have shown up in either place's tests. `RebootEvaluation` now carries a **required** `windowEndsAt: Date | null` (required, not optional, so a future branch that forgets it fails to compile rather than quietly uncapping), and `patchJobExecutor` threads it into `executeReboot`. Covered at both layers: `patchRebootHandler.test.ts` (evaluate returns it; `executeReboot` caps on it; null is ignored) and `patchJobExecutor.test.ts` (the executor actually forwards it).

**2 — should-fix: the 7-day ceiling did not bound what its comment claimed.** The refinement checked only `maxDeferrals × deferralMinutes ≤ 10080`, but `computeRebootDeadline` emits `delay + maxDeferrals × deferralMinutes`. `rebootDelayMinutes: 1440` + a 10080 budget put the real deadline 1440 minutes past the agent's ceiling. Not exploitable today (the agent range-checks only `delayMinutes`, and deadline enforcement is W2), but it becomes live the moment W2 lands. The check now sums the delay in; message and docs updated to match. New boundary cases pin exactly-10080 (accepted) and 10081 (rejected).

**3 — should-fix: `windowEndsAt`'s projection shipped untested.** It is the source of truth for the feature's headline safety invariant and the projection is subtle (wall-clock-rendered-as-UTC values whose *difference* is the only meaningful quantity). Added six cases to `featureConfigResolver.test.ts`. **Mutation-verified:** replacing the projection with a naive `new Date(windowEnd)` turns 3 of the 6 red, including the non-UTC case that would have been 5 hours early — i.e. it would have silently truncated deferral time.

**4 — should-fix: the opt-in guard was not pinned.** The code deliberately uses `rebootAllowDeferral !== true`, not a truthy check, but every existing case (`false` → OFF, `true` → ON) passed under either. **Mutation-verified:** swapping in `!settings.rebootAllowDeferral` now fails.

**Declined (2 nits, both raised and both judged not worth a change):**
- `Number(e.target.value) || <default>` in the two new `PatchTab` number inputs. It matches the two adjacent pre-existing fields in the same file, and an out-of-range value is rejected by the server and surfaced through `useFeatureLink`'s error state — not silently persisted. Changing only the new fields would make the file internally inconsistent; changing all four is scope creep for this wave.
- The new `patchCanonicalExportParity` cases verify DB/jsonb round-trip and route wiring rather than independent schema correctness (both sides call the same normalizer). That is the existing pattern for every case in that array; independent correctness of the two refinements is carried by `packages/shared/src/validators/index_inline_settings.test.ts`.

**Reviewers found no other must-fix or should-fix issues.** Independently verified by the reviewers, not taken on faith: the migration's copied function body is byte-identical to `2026-08-21-patch-reboot-delay-minutes.sql:54-230` apart from the three added keys (ACL block included); the `plan!` non-null assertion in `executeReboot` is unreachable-when-null by construction; the SQL/zod/UI/docs bounds agree; `config_policy_patch_settings` genuinely has no `org_id`/`partner_id` and appears in no cascade or export-policy list; and every new test file lands in exactly one CI job with the two integration files confirmed un-gated in the blocking `integration-test` job.

**Tests (all re-run after the fixes, verified green):**

| Suite | Result |
|---|---|
| `packages/shared` full | 92 files / **2314 passed** |
| `apps/api` affected unit sweep (patchRebootHandler, maintenanceRebootWorker, patchJobExecutor, featureConfigResolver, partnerApi/configuration, patchJobService, scriptExecution ×2, configurationPolicy, configPolicyPatching, autoMigrate) | 24 files / **714 passed** |
| `apps/web` configurationPolicies + localeParity | 30 files / **364 passed** |
| Integration (private per-worktree stack): `partnerApiConfigurationWatermark` + `patchCanonicalExportParity` | **20 passed** |
| Integration: `tenantCascade` + `tenant-export-policy` | **7 passed** |
| RLS coverage | **100 passed** |
| `tsc --noEmit` (`apps/api`, `apps/web`) | clean |
| `pnpm lint` | 6/6 |
| `pnpm db:check-drift` | no drift, 636 files match the ledger |
| Migration idempotency | applied twice by hand under `ON_ERROR_STOP=1`; second pass a clean no-op |

Not run locally: the remaining `apps/api` unit files and the other integration shards — CI covers both. `packages/shared` `tsc` reports 3 pre-existing errors in `validators/quotes.test.ts`, untouched here.

**Status:** review-clean, awaiting maintainer merge. Not merged, and #4661 left open — both are yours to call.

ToddHebebrand:
**CI: green on `1c618d99a`** — 68 pass, 1 skipping, 0 fail. `CI Success`, all four `Integration Tests` shards, `Test API`, `Test Web`, `Type Check`, and both `Check Migrations` jobs pass.

**One flake worth recording, unrelated to this PR but relevant to a later wave of #3207.** `Test Agent (Windows)` failed on the first attempt and passed on retry. The cause was **not** a test failure — `internal/sessionbroker` blew the 10-minute package timeout:

```
FAIL  github.com/breeze-rmm/agent/internal/sessionbroker  600.048s

goroutine 590 [chan receive, 9 minutes]:
github.com/Microsoft/go-winio.(*win32PipeListener).Close(...)
    go-winio@v0.6.2/pipe.go:578
sessionbroker.(*Broker).StopAcceptingAndWait(...)
    agent/internal/sessionbroker/broker.go:859
sessionbroker.(*Broker).Close(...)
    agent/internal/sessionbroker/broker.go:558
sessionbroker.TestNamedPipeListenAndAccept(...)
    agent/internal/sessionbroker/broker_windows_test.go:123
```

`TestNamedPipeListenAndAccept` hung for 9 minutes inside `Broker.Close()` → `StopAcceptingAndWait` → go-winio's named-pipe listener `Close`, which never returned; every other Windows package (`eventlog`, `ipc`, `logging`, `state`, `watchdog`, `backup/vss`, `cmd/breeze-watchdog`) was `ok`. That is a non-deterministic close/accept race in the broker's named-pipe shutdown path, not a product of this diff — this PR changes **zero** files under `agent/`, the same job passed on the previous head `35ca509bd` against a byte-identical `agent/` tree, and neither commit `main` gained in between (#4585, #4604) touches `agent/` either.

Flagging it rather than letting the retry bury it, because **W3 of this feature modifies exactly that file** — the plan adds a correlated `RequestNotificationDecision` fan-out to `sessionbroker`. Whoever picks up W3 should expect this hang to recur and may want to fix the shutdown race first. Probably worth its own issue; I have not filed one.

Issue #4685, opened 2026-09-03
[Ops] Extract HTTP/business/fleet metric recorders so the worker can publish full series parity


ToddHebebrand:
**Context:** #4523 (fixed by wiring the worker's `/metrics` target into the checked-in Prometheus config) noted that series parity between the `api` and `worker` roles is still partial by design — the worker only publishes the role-agnostic runtime set (event-loop lag, DB pool health) added in #4423. It does not publish the HTTP/business/fleet series or the extension/backup/S1 counters, because those recorders are bound in `routes/metrics.ts`, which a worker-role process must never import (`services/workerEntrypointClosure.contract.test.ts`).

**Ask:** if we want those series reported from the worker container too (e.g. for background-job-specific business counters), extract the relevant recorders out of `routes/metrics.ts` into their own leaf module(s) — following the `metricsRegistry.ts` / `metricsRuntime.ts` / `metricsScrapeAuth.ts` pattern from #4423 — so `worker.ts` can bind them without pulling in the full route graph.

Not urgent: dashboards already know to expect the HTTP/business/fleet series only from `job="breeze-api"` (documented in `docs/deploy/worker-split.md` and `monitoring/prometheus.yml`). This is a "nice to have" for anyone who wants worker-side visibility into those specific counters, not a currently-missing alert path.

Related: #4143, #4423, #4523.

PR #4689, opened 2026-09-03
fix(ops): wire the worker /metrics endpoint into Prometheus scrape config (#4523)

### LINKED ISSUE #4523: [Ops] Worker /metrics is a new Prometheus target nothing scrapes; series parity with API is partial
**Problem:** PR #4423 (#4143) gave the worker container its own `/metrics` endpoint on port 3001, event-loop and DB-pool monitors, and a `breeze_role` Sentry tag. Two operational gaps remain:

1. **Nothing scrapes it.** The worker is a new, separate Prometheus target. `docker-compose.monitoring.yml` / the prod Prometheus config still only scrape the API. Until a scrape job is added, the new series never reach Grafana and no alert can fire on worker pool exhaustion.
2. **Series parity is partial.** The worker does not publish the API's HTTP/business/fleet series or the extension/backup/S1 counters, because those recorders bind in `routes/metrics.ts`. Dashboards that assume a single `breeze_*` source should filter on `breeze_role` or they will double-count once both roles report.

**Fix:** add the worker scrape job (allowlisted by the direct-socket IP gate documented in the PR), update the deploy doc, and file or fold in the recorder extraction if the remaining series are wanted on the worker.

Note: #4423 was reviewed inline by its author because the subagent pool was saturated at the time. A second-reviewer pass on the auth gate and the `metricsRegistry` leaf-module boundary would be reasonable.

Related: #4143, PR #4423.

ToddHebebrand:
## The problem

#4423 gave the worker container its own `/metrics` Prometheus target on port 3001 (event-loop lag + DB pool health monitors, `breeze_role` Sentry tag). Two operational gaps remained (#4523):

1. **Nothing scrapes it.** `docker-compose.monitoring.yml` / the checked-in Prometheus config only had a `breeze-api` job. Until a scrape job is added, the worker's series never reach Grafana and no alert can fire on worker pool exhaustion — the container running the heavy scheduled jobs stayed invisible.
2. **Series parity is partial by design.** The worker only publishes the role-agnostic runtime set (event-loop lag, DB pool health) — not the HTTP/business/fleet series or extension/backup/S1 counters, which are bound in `routes/metrics.ts` (a file a worker-role process must never import). Dashboards assuming a single `breeze_*` source need to filter by job or risk double-counting once both roles report.

## The fix

- **`monitoring/prometheus.yml`**: add a `breeze-worker` scrape job (the local/optional `docker-compose.monitoring.yml` stack), reusing the same `METRICS_SCRAPE_TOKEN` bearer auth and `credentials_file` as `breeze-api`. Resolves `worker:3001` over the shared `breeze` Docker network — no host port publish needed locally, and it's simply "down" if the worker-split profile isn't enabled (same optional-target pattern as the redis/postgres/node exporters already in this file).
- **`deploy/prometheus-remote-scrape.example.yml`**: add the equivalent multi-region Tailscale template block ("(A2) WORKER TARGET") for droplets that adopt the worker-split profile, including the distinct host-port bind-publish needed (the worker also listens on 3001 internally, so it can't share api's `:3001` bind on the same droplet). Also documents that the public-edge fallback path does **not** extend to the worker: it isn't proxied through Caddy at all, and its scrape gate deliberately ignores `X-Forwarded-For` (direct-peer auth only, per `metricsScrapeAuth.ts`).
- **`docs/deploy/worker-split.md`**: clarifies what's now wired automatically (local stack) vs what still needs manual droplet-side setup (remote multi-region Prometheus).
- No internal hostnames/IPs added — the remote-scrape example already used placeholder tokens (`<EU_TAILSCALE_IP>`, etc.) and the new content follows the same convention.

**Filed #4685** for the separate, larger ask (recorder extraction so the worker can publish full series parity) — the issue itself calls this "file or fold in," and it's a bigger leaf-module refactor than scrape wiring, so it gets its own PR.

## Verification

- **New regression test** (`apps/api/src/config/monitoringPrometheusScrapeConfig.test.ts`), red-first: parses `monitoring/prometheus.yml` with `js-yaml` and asserts the `breeze-worker` job exists, targets `worker:3001`, serves `/metrics`, and shares the same bearer `credentials_file` as `breeze-api`. Confirmed it failed before the fix (job missing) and passes after.
  ```
  cd apps/api && npx vitest run src/config/monitoringPrometheusScrapeConfig.test.ts src/config/composeBindMounts.test.ts src/config/envComposeParity.test.ts
  # 3 files passed, 40 tests passed
  ```
- **Typecheck:** `NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit --project apps/api/tsconfig.json` — clean.
- **Lint:** `npx eslint src/config/monitoringPrometheusScrapeConfig.test.ts` — clean.
- Manually parsed both YAML files with `js-yaml` to confirm they're well-formed after editing (the worker jobs in the remote-scrape example are template blocks, commented like the existing public-edge fallback block, by design).

Closes #4523

### DIFF
--- a/apps/api/src/config/monitoringPrometheusScrapeConfig.test.ts
+++ b/apps/api/src/config/monitoringPrometheusScrapeConfig.test.ts
@@ -30,6 +30,7 @@ interface ScrapeConfig {
   scheme?: string;
   static_configs?: Array<{ targets?: string[] }>;
   authorization?: { type?: string; credentials_file?: string };
+  relabel_configs?: Array<{ source_labels?: string[]; target_label?: string }>;
 }
 
 interface PrometheusConfig {
@@ -51,6 +52,12 @@ describe('monitoring/prometheus.yml worker scrape target (#4523)', () => {
   });
 
   it('scrapes the worker container on its own job, distinct from breeze-api', () => {
+    // Exactly one — a duplicate job_name (e.g. from a bad merge) would have
+    // Prometheus scrape the same target twice under identical labels, which
+    // silently double-counts the worker's series the same way an unfiltered
+    // dashboard would double-count across roles.
+    expect(jobs.filter((j) => j.job_name === 'breeze-worker')).toHaveLength(1);
+
     const workerJob = jobs.find((j) => j.job_name === 'breeze-worker');
     expect(workerJob).toBeDefined();
     expect(workerJob!.job_name).not.toBe('breeze-api');
@@ -58,10 +65,15 @@ describe('monitoring/prometheus.yml worker scrape target (#4523)', () => {
     const targets = workerJob!.static_configs?.flatMap((sc) => sc.targets ?? []) ?? [];
     expect(targets).toContain('worker:3001');
 
-    // The worker exposes `/metrics` (unauthenticated path shape differs from
-    // the api role's `/api/metrics/scrape`) — see worker.ts and
-    // docs/deploy/worker-split.md.
+    // The worker's Bearer-authenticated scrape endpoint is `/metrics` — a
+    // different path shape than the api role's `/api/metrics/scrape` — see
+    // worker.ts and docs/deploy/worker-split.md.
     expect(workerJob!.metrics_path).toBe('/metrics');
+
+    // Same instance-relabel rule as breeze-api, so `instance` carries the
+    // bare hostname (not `worker:3001`) and dashboard/alert queries that
+    // join on `instance` across both jobs still line up.
+    expect(workerJob!.relabel_configs?.[0]?.target_label).toBe('instance');
   });
 
   it('authenticates the worker scrape with the same bearer token as breeze-api', () => {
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, comment-analyzer, pr-test-analyzer (3 parallel Sonnet subagents), one round.

**Findings:** 3 real issues raised → all fixed in 4c8d75e6b; 0 outstanding.
- code-reviewer: `deploy/prometheus-remote-scrape.example.yml` documented `docker compose up -d worker` without `--profile worker-split` — the worker service is profile-gated in both `docker-compose.yml` and the real droplet file `deploy/docker-compose.prod.yml`, so the command as written wouldn't start it. Fixed.
- comment-analyzer: two comments mischaracterized the worker's auth — one called `/metrics` "unauthenticated" (it's Bearer-gated, identical gate to the api role via `metricsScrapeAuth.ts`), one implied `monitoring/prometheus.yml`'s worker job had "no request-route auth" (same gate, just raw-`node:http` instead of the Hono route graph), and one misattributed the shared `credentials_file` to the `x-api-env` compose anchor (that anchor feeds the *servers'* own token check; the shared file is Prometheus's own `metrics_scrape_token` Docker secret in `docker-compose.monitoring.yml`). All three reworded.
- pr-test-analyzer: rated the new regression test adequate but flagged two cheap gaps — no duplicate-job-name guard, no assertion on the `instance` relabel rule. Both added.

**Tests:** `cd apps/api && npx vitest run src/config/monitoringPrometheusScrapeConfig.test.ts src/config/composeBindMounts.test.ts` — 2 files, 6 passed. `eslint` on the touched test file — clean.
**Typecheck:** `NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit --project apps/api/tsconfig.json` — clean.

Issue #4673, opened 2026-09-03
Partner-wide reads via breeze_current_partner_id() SELECT branch (#2468)


ToddHebebrand:
Tracking issue for **Partner-wide reads via breeze_current_partner_id() SELECT branch (#2468)** (`partner-current-partner-id-select`).

- Plan: `docs/superpowers/plans/tenancy-rls/2026-08-16-2468-partner-current-partner-id-select-branch.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All four waves merged: W01 #4690, W02 #4725, W03 #4803, W04 #4965 (517f6b133d). Unreleased — ships with the next tag.

Not verified: the 23 tables still lacking a breeze_current_partner_id() SELECT branch are tracked as follow-ups #4942–#4963 and #4970 and gated by the shrink-only ratchet in rls-coverage.integration.test.ts.

Issue #2468, opened 2026-07-13
Partner-wide reads: replace nested system-context escalation with a breeze_current_partner_id() RLS read branch


ToddHebebrand:
## Problem

A partner-wide row (`org_id NULL, partner_id = P`) is invisible to an ORG-scoped token: `breeze_has_org_access(NULL)` is false, and `breeze_has_partner_access(P)` is false because org scope carries an empty `accessible_partner_ids`. Verified against a live DB — `configuration_policies` has **no** own-partner read branch, so **org tokens cannot see partner-wide config policies at all**.

Every request-path reader of a partner-linkable feature is therefore blind to partner-wide state. In #2417 this meant manual "Back up now" fell back to a legacy single-mode job and dashboards reported partner-protected devices as unprotected.

## Current workaround (and its cost)

CLAUDE.md prescribes the #1105 pattern — `runOutsideDbContext(() => withSystemDbAccessContext(...))`. It works, and #2417 uses it, as do `vulnerabilityFleetQueries.ts`, `ticketConfigService.ts`, `commandQueue.ts` and others on request paths today.

But the **entire request handler already runs inside a `withDbAccessContext` transaction** (`middleware/auth.ts:613`), so a nested system context **acquires a second pooled connection while the first is still held**. At concurrency >= pool size, every in-flight request holds one connection and waits for a second — postgres-js queues acquisitions with no timeout, so that is a hang, not just contention. The US DB ceiling is ~25 connections.

It also **bypasses RLS**, so every escalated query has to self-tenant or it becomes a cross-tenant hole. (#2417 shipped exactly such a hole in an adjacent code path — a partner-wide policy assigned to org B fanned out org B's devices when resolving org A — caught in review and fixed, but the class is latent wherever we escalate.)

This is a pre-existing, repo-wide characteristic of the sanctioned pattern, not specific to backup.

## Proposed fix

Follow the precedent already in the repo: `apps/api/migrations/2026-06-13-catalog-partner-read-branch.sql`. Add a **SELECT-only** own-partner branch to the RLS policies:

```sql
OR (org_id IS NULL AND partner_id = public.breeze_current_partner_id())
```

`breeze_current_partner_id()` reads the `breeze.current_partner_id` GUC — the caller's **own** partner, set for every scope, distinct from `accessible_partner_ids` which governs partner-axis *writes*. INSERT/UPDATE/DELETE policies stay unchanged, so writes remain locked to partner/system scope. No extra connection, no RLS bypass.

[...]

PR #2417, opened 2026-07-13
feat(backup): backup profiles — multi-source selection profiles, partner-wide backup, per-selection job fan-out

ToddHebebrand:
## Summary

Implements Backup Profiles per `docs/superpowers/specs/2026-07-13-backup-profiles-design.md` (phases 1+2 together, as confirmed): Cove-style **selection profiles** answering "what to protect" per device class, **partner-wide backup**, and **per-selection job fan-out** — fixing the silent multi-mode shadowing where a server assigned file + System State + SQL policies got exactly one working backup. Builds on the tab redesign in #2415.

### Data model + RLS
- `backup_profiles`: dual-ownership (org XOR partner, CHECK + dual-axis RLS in the creating migration), `selections` jsonb validated by a shared Zod schema (any subset of file / system_image / mssql / hyperv with per-source options).
- `config_policy_backup_settings` goes dual-axis (denormalized `partner_id`, `org_id` now nullable) so partner-wide policies can carry backup links — backup **leaves `ORG_SCOPED_ONLY_FEATURE_TYPES`** and joins `PARTNER_LINKABLE_FEATURE_TYPES`. New columns: `backup_profile_id` (RESTRICT), `destination_config_id` (SET NULL).
- `backup_configs.is_default` + partial unique index: the **org default destination**, resolved at job time for partner-wide policies and profile links without a pinned destination.
- `backup_jobs.backup_mode` / `mode_targets`: each fan-out job carries its own selection.

### Resolution + worker (semantics preserved)
- Winner-take-all hierarchy is untouched — the winning link's profile expands to N selections, and the scheduler/run-now create one job per (device, selection), deduped and idle-checked per device+mode. Dispatch prefers the job's own selection; legacy NULL-mode jobs read the settings row exactly as before.
- Destination chain: explicit link destination → legacy `featurePolicyId` destination (pre-profile links keep working, no data migration) → org default. Missing destination = loud error, never a silent skip.
- The scheduler enumerates orgs covered by partner-wide policies (partner fan-out playbook step 5); `effectiveBackupModes()` replaces raw `settings.backupMode` filters in the Hyper-V/SQL discovery-target readers.
- Cross-tenant guard: a link's pinned destination must belong to the policy's own org (FK proves existence, not tenancy); partner policies must use the org-default sentinel.

### API + AI
[...]

PR #4690, opened 2026-09-03
feat(api): partner-wide SELECT branch on the config-policy RLS chain (#2468)

### LINKED ISSUE #4674: W01: Migration + RLS integration proof (SELECT branch on breeze_current_partner_id)
Wave W01 of #4673 — Migration + RLS integration proof (SELECT branch on breeze_current_partner_id)
Branch: `feature/4673-partner-current-partner-id-select/wave-4674`

ToddHebebrand:
Wave 1 of #4673 (origin issue #2468). Plan: `docs/superpowers/plans/tenancy-rls/2026-08-16-2468-partner-current-partner-id-select-branch.md` (brought onto this branch from `plan/2468-partner-rls-select` with `tracking_issue` frontmatter added).

## Problem

A partner-wide config row is `org_id NULL, partner_id = P`. An **org-scoped** session cannot see it: `breeze_has_org_access(NULL)` is false, and `breeze_has_partner_access(P)` is false because org scope carries an empty `accessible_partner_ids` (that GUC governs partner-axis **writes**; an org token never holds it).

Every request-path reader of a partner-linkable feature therefore escalates via the sanctioned #1105 pattern, `runOutsideDbContext(() => withSystemDbAccessContext(...))`, which:

1. **Double-holds pooled connections.** The whole request already runs inside `withDbAccessContext`, so the nested system context opens a SECOND transaction on a SECOND connection while the first is still held. postgres-js queues acquisitions with no timeout, so at concurrency ≥ pool size (~25 on the hosted DBs) that is a hang, not just contention.
2. **Bypasses RLS entirely**, so every escalated query has to self-tenant or it becomes a cross-tenant hole (#2417 shipped exactly that class in an adjacent path).

## What this PR does

Adds a **SELECT-only** own-partner branch to the config-policy chain, keyed on `public.breeze_current_partner_id()` — the caller's **own** partner, read from the `breeze.current_partner_id` GUC, which `buildDbAccessContext` populates for *every* scope including org tokens. No extra connection, no RLS bypass, writes untouched.

Migration: `apps/api/migrations/2026-10-05-110000-config-policy-partner-wide-select.sql` (idempotent, no inner `BEGIN`/`COMMIT`, sorts after every committed migration incl. the two that landed on main from #4585 mid-flight).

**15 tables**, each getting a new `<table>_partner_wide_select` policy:

| Shape | Tables |
|---|---|
| Direct column (`org_id IS NULL AND partner_id = breeze_current_partner_id()`) | `configuration_policies`, `config_policy_backup_settings`, `backup_profiles` |
| One-hop EXISTS join (`config_policy_id`) | `config_policy_feature_links`, `config_policy_assignments` |
| Two-hop JOIN (`feature_link_id`) | `config_policy_alert_rules`, `config_policy_automations`, `config_policy_compliance_rules`, `config_policy_patch_settings`, `config_policy_maintenance_settings`, `config_policy_event_log_settings` |
| Scalar-subquery chain (matches existing per-command policy shape) | `config_policy_sensitive_data_settings`, `config_policy_monitoring_settings`, `config_policy_monitoring_watches` (3 hops), `config_policy_remote_access_settings` |

### Design decisions worth reviewing

- **Separate `FOR SELECT` policy per table, never an edit to the existing FOR ALL policy.** Appending the branch to a FOR ALL `USING` would also widen UPDATE/DELETE row targeting — an org admin could delete their MSP's shared policy. Postgres never consults FOR SELECT policies when computing write target rows, so a separate permissive policy ORs into reads only. This also makes the migration create-only, hence immune to the policy-name drift several of these tables have accumulated. Same mechanism as `cis_baselines_partner_wide_select` (2026-08-10) and `2026-06-13-catalog-partner-read-branch.sql`.
- **`=`, not `IS NOT DISTINCT FROM`.** Agent sessions set `currentPartnerId: null` (`middleware/agentAuth.ts`), so `partner_id = NULL` is NULL and never true. **Agents are deliberately unaffected by this wave** — wave 2 populates that GUC on purpose. `IS NOT DISTINCT FROM` would have matched NULL-partner rows; there is an explicit test for this.
- **`config_policy_onedrive_settings` is excluded.** It carries `org_id NOT NULL` and `breeze_validate_config_policy_onedrive_settings` (2026-07-27-b) RAISEs when the parent policy has `partner_id IS NOT NULL`. A partner-wide OneDrive settings row is not representable, so a branch there would be unreachable code. Documented in the migration header.
- **No registration-list changes.** All 15 tables are already in `DUAL_AXIS_TENANT_TABLES` / `PARENT_FK_JOIN_POLICY_TABLES` / the cascade + export-policy lists. This PR changes policies only — no new table, no new column — so cascade and export-policy registration are untouched.

### Three existing suites had to be flipped — and were passing vacuously

`configurationPoliciesPartnerRls`, `backupProfilesPartnerRls` and `remote-access-settings-rls` each asserted that an org-scoped caller sees **zero** partner-wide rows of its own partner.

All three built their org context **without `currentPartnerId`** — i.e. the AGENT context shape, not an org token's. They were therefore not testing what their names and comments claimed, and they still pass unchanged after this migration. Rather than leave a false invariant in place, each `orgContext` helper now takes the partner id (defaulting to `null`, so untouched call sites are unchanged), and the three assertions are flipped to the new contract, each paired with:

- a **rowCount-based** no-write assertion (RLS filters an UPDATE's target rows silently; "it didn't throw" would be satisfied by a successful hijack), and
- a **different-partner** negative proving the branch is keyed on the caller's partner, not the org.

## Verification (all against real Postgres, `breeze_app`, FORCE RLS)

New suite `apps/api/src/__tests__/integration/configPolicyPartnerWideSelect.integration.test.ts` (11 tests): seeds a full partner-wide chain (policy + 10 feature links + 2 assignments + every per-feature child + backup profile/settings) and probes **all 16 relations per assertion**, so a table omitted from the migration fails loudly instead of silently reading zero rows forever.

| Check | Result |
|---|---|
| **Red first** — new suite before the migration | ✅ RED: all 16 relations invisible to an org session of the owning partner |
| New suite after the migration | ✅ 11 passed |
| **Mutation control** — drop the 3 policies, re-run the flipped suites | ✅ exactly the 3 flipped assertions re-red; every cross-partner negative stays green |
| Migration re-applied over itself | ✅ no-op (17 `*_partner_wide_select` policies, incl. the 2 pre-existing) |
| `configPolicyPartnerWideSelect` + `configurationPoliciesPartnerRls` + `backupProfilesPartnerRls` + `remote-access-settings-rls` + `normalizedConfigPolicyRls` + `agentPolicyResolversPartnerWide` | ✅ 78 passed (6 files) — on a **freshly recycled DB**, so migration ordering is proven on a from-scratch replay |
| `vitest.config.rls-coverage.ts` | ✅ 100 passed |
| `vitest.config.rls.ts` | ✅ 26 passed, 5 skipped |
| `configurationPolicyPartnerResolution`, `sensitiveDataPoliciesPartnerRls`, `backupFeatureLinkCascade`, `configPolicyFeatureReferenceIntegrity`, `configPolicyAutomationRunRls`, `partnerAxisSystemContext`, `configurationPolicyOrgScopedPartnerWide` | ✅ green |
| `src/db/autoMigrate.test.ts` | ✅ 67 passed |
| `scripts/check-migration-naming.sh` and `--against-ref origin/main` | ✅ OK |
| `tsc --noEmit` (apps/api), eslint on changed files | ✅ clean |

Branch has `origin/main` merged in, so the above ran against the same tree CI will build.

## Risk

This is the repo-wide read-visibility change #2417 deliberately deferred: org-token reads now widen, **at the DB layer**, to their own partner's partner-wide rows. That is the intended product semantic (org users should see the MSP's shared policies — the app layer already emulates it via escalation) and it is strictly *narrower* than what the system-context escape shows those same callers today, which is every tenant's rows. Reviewed as a standalone security PR per the plan.

## Not in this wave

- **Wave 2** (#4673): populate `currentPartnerId` on agent + hand-built org contexts.
- **Wave 3**: remove `withPartnerWideVisibility` and the `configurationPolicy.ts` system-context wraps. Must merge after wave 2 — unwrapping an agent path before the GUC is populated silently delivers zero partner-wide config to agents, with no error.
- **Wave 4**: CLAUDE.md playbook step 3 + a contract assertion in `rls-coverage` so the *next* partner-wide table fails loud.

Closes #4674

---

## Downstream behavior changes found by the review sweep (read this before merging)

The review audited all 36 production files touching the 15 tables. **No leak, and no code changed in this PR** — but three org-scoped request paths read these tables with *no* tenant predicate and *no* system wrapper, so they were silently getting zero rows for partner-wide policies and now get the real ones. All three move in the "silently broken → works as configured" direction, which is the point of #2468, but they are behavior changes and none is yet pinned by a test.

**Scope note:** the newly-affected population is narrower than "all org sessions". `configuration_policies_org_isolation` already ORs in `breeze_has_partner_access`, so any context with a non-empty `accessiblePartnerIds` (api-key auth, helper auth) already saw these rows. Newly affected: standard user JWTs, org-scoped OAuth bearers, and AI-agent runs. Unaffected: agent/WS/heartbeat, portal, client-AI (all `currentPartnerId: null`).

1. **`jobs/backupRetention.ts:104-108` (`resolveGfsConfigForJob`) — backup retention.** For a job whose `featureLinkId` belongs to a partner-wide policy, the settings row was invisible → `gfsConfig = null` → `computeExpiresAt(...) === null` → **the snapshot never expired.** It now resolves and writes GFS tags + a finite `expiresAt`. Verified by reading the function: this runs at result-persistence time per job, so **existing rows with `expires_at NULL` are untouched — there is no retroactive deletion.** New snapshots under partner-wide backup policies will start expiring on the retention the MSP actually configured. Correct, but it is a data-retention change and worth a conscious merge decision.
2. **`services/backupResultPersistence.ts:429-436` — legal hold / immutability.** Same shape. Previously `legalHold: false` and no `immutableUntil`; now stamped (which also engages provider object-lock). Strictly more protection. Narrow trigger window.
3. **`services/configPolicyPatching.ts:296,308` via `routes/configurationPolicies/patchJobs.ts:370` — user-visible fix.** `GET /configuration-policies/:id/resolve-patch-config/:deviceId` returned `autoApproveDeferralDays: 0`, `apps: []`, `approvalRing: null` whenever the winning policy was partner-wide — the UI showed "no approval ring" and dropped every app block/pin rule the scheduler was in fact enforcing. It now returns the real values.

Two smaller ones, both error-path→success-path and intent-matching: an org caller passing a partner-wide backup-profile id to the `manage_policy_feature_link` AI tool no longer gets a spurious "Backup destination not found in this organization" (`services/configurationPolicy.ts:638-642`), and the `sensitive_data` whole-policy-reference branch at `:2293-2299` — which already had the right app-layer predicate but ran in a context that could not see the rows — now works.

**Follow-up owed (wave 3, the sweep wave):** regression tests pinning 1-3, and a re-check of `services/helperPermissions.ts:86-95`, the one config-policy resolver with neither an ownership predicate nor a visibility wrapper (safe today only because helper auth already grants partner access — it becomes load-bearing when wave 3 removes the escalations).

## Review round

`code-reviewer` + `silent-failure-hunter` (Sonnet), plus a full production-reader audit of all 36 files touching the 15 tables. **0 correctness findings; 0 tenant-isolation findings.** Three documentation findings, all fixed in this PR:

- `services/configurationPolicy.ts` — `policyAccessCondition`'s comment claimed RLS was stricter and cited the very test this PR flips. It is now the *app* layer that is stricter for reads; rewritten so nobody relaxes that gate believing RLS backstops it.
- `middleware/bearerTokenAuth.ts` and `middleware/agentAuth.ts` — both said `currentPartnerId` feeds "only the catalog branch"; already stale, and more so now.


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` + `silent-failure-hunter` (Sonnet), plus a production-reader audit of all 36 files touching the 15 affected tables.

**Findings:** 0 correctness, 0 tenant-isolation. 3 documentation findings, all fixed in-PR (`67027aa76`, `3a5170492`): `policyAccessCondition`'s comment asserted RLS was the stricter layer and cited the very test this PR flips (it is now the *app* layer that is stricter for reads); `bearerTokenAuth.ts` and `agentAuth.ts` both claimed `currentPartnerId` feeds "only the catalog branch". 0 outstanding.

The sweep also surfaced 3 **downstream behavior changes** on org-scoped request paths that read these tables with no tenant predicate and no system wrapper — all "silently broken → works as configured", none a leak, none a code change in this PR. They are enumerated with file:line in the PR body. The one worth a conscious merge decision is `jobs/backupRetention.ts:104-108`: snapshots under a partner-wide backup policy were getting `expiresAt = null` (never expire) and will now expire on the retention the MSP configured. Verified by reading the call path that this runs per-job at result-persistence time, so **existing `expires_at NULL` rows are untouched — no retroactive deletion.**

**Tests:**
- Red-first: the new suite was RED before the migration (all 16 relations invisible to an org session of the owning partner), green after.
- Mutation control: dropping the 3 policies re-reds exactly the 3 flipped assertions and nothing else — the flipped tests are not vacuous.
- Also caught and fixed: the 3 pre-existing suites that asserted the old invisibility were all building their org context **without `currentPartnerId`**, i.e. the agent context shape — they were passing vacuously and would have kept passing after this migration. Their helpers now take the partner id.
- Local, against real Postgres on a freshly recycled DB (so migration ordering is proven on a from-scratch replay): 78 passed across the 6 config-policy/partner suites; `vitest.config.rls-coverage.ts` 100 passed; `vitest.config.rls.ts` 26 passed / 5 skipped; `autoMigrate.test.ts` 67 passed; `tsc --noEmit` clean; eslint clean.
- CI on head `3a5170492`: **all green** — Integration Tests 4/4 shards, Test API, Check Migrations (+ non-superuser), Lint, CI Success.

**Status:** review-clean, CI green, awaiting maintainer merge. Wave 2 (agent + hand-built org contexts) must land before wave 3 unwraps the escalations — an agent path unwrapped early delivers zero partner-wide config with no error.

PR #4691, opened 2026-09-03
feat(ai-agents): surface ticket-triage skip reasons + deselect drifted batch cards (#4462, #4459)

### LINKED ISSUE #4459: AI agents: use offending ids to deselect drifted cards in batch approve
When a batch approve/decline (wave P2-2) partially fails because some cards drifted out of the batchable set, the response's `offending` ids aren't used to deselect just those cards in the inbox UI — currently the whole batch selection has to be redone.

Source: PR #4233 body ("Follow-ups").

### LINKED ISSUE #4462: AI agents: surface per-field skip reasons on the ticket-triage run DTO
The ticket-triage finalizer (wave P2-4) computes per-field skip reasons (e.g. confidence below threshold, provenance pre-filter) when deciding which proposed field writes to apply, but they're only logged to the console today — the plan asked for them to be surfaced on the run DTO so techs can see why a field wasn't written.

Source: #4191 rulings comment ("Deferred to follow-ups") + PR #4301 body ("Follow-ups").

ToddHebebrand:
Closes #4462
Closes #4459

Two small P2-4/P2-2 follow-ups from the #4187 sweep, combined into one PR per the 2026-09-02 triage cluster note (todd's comment on #4462: "both small same-wave fast-track leftovers ... one combined PR").

## #4462 — surface per-field skip reasons on the ticket-triage run DTO

`persistTicketTriage` (`ticketTriageFindings.ts`) already computed a fully-typed `skipped: TicketTriageSkip[]` list (why a slot — fields/link/note/draft-reply/draft-resolution — did not become a write), but `finalizeTicketTriage` (`runLoop.ts`) only `console.info`'d it and threw it away.

- `TicketTriageSlot` / `TicketTriageSkipReason` / `TicketTriageSkip` moved from the API-internal `ticketTriageFindings.ts` into `@breeze/shared`'s `aiAgentRuns.ts` (same pattern as `AlertVerdictSuggestionReason` right above them), so both the API and the DTO can reference the same union.
- `finalizeTicketTriage` now attaches `skipped` onto `result.outcome.ticketTriageSkipped` (mutates in place, same pattern as `outcome.alertVerdictIntent`), so `finishRun` persists it in the run's `outcome` jsonb.
- `runTrace.ts`'s `mapTicketProposal` projects it onto `AiAgentRunTicketProposalDto.skipped` (undefined-when-empty, matching `intentIds`/`draftsWritten`).
- `RunDetailPage.tsx`'s `TicketProposalSection` renders a new "Not applied" block listing each skipped item + its reason (literal-switch label helpers, not dynamic i18n keys — matches `draftKindLabel`'s convention so the key-usage scanner can see every key). New locale keys added to all 8 locales (7 non-English are unreviewed machine translations, matching the project's established convention).

## #4459 — use `offending` ids to deselect drifted cards in batch approve

When a batch approve/decline 422s `batch_not_homogeneous`, the server's `offending` ids named exactly which cards drifted out of the batchable set — but the web inbox discarded them and froze the *whole* group behind one banner, forcing the approver to redo the selection.

- `ApprovalsInbox.tsx`: `decideGroup`'s `batch_not_homogeneous` handling now marks each `offending` id with a new per-row `driftedFromBatch` error (reusing the existing per-row error/banner machinery) instead of a whole-group banner. `buildSections` now excludes a drifted id from grouping, so on the next render: the offending card renders standalone (single-card decide), and its former groupmates re-group — possibly as a smaller batch — immediately re-batchable with **no redone selection**. Falls back to the old whole-group banner only if the server names no offending ids (defensive).
- Fixed a real gap in the plumbing this depended on: `AssertionChallengeError` (`stores/authenticator.ts`) discarded the `offending` field from a refused batch-challenge body, keeping only the `token`. The batch **challenge** route (`/batch/assertion-challenge`) re-validates the set *before* minting a proof, so on an **approve** this is usually where drift is caught — not the later `/batch/decide` 422, which is only reached on **deny** (skips the ceremony). Without this fix, only deny could ever surface real offending ids; approve always got `offending: []`. Now `AssertionChallengeError` carries `offending` through and `decideIntentApprovalBatch` uses it instead of a hardcoded `[]`.
- New locale key `errors.driftedFromBatch` added to all 8 locales.

## Verification

- API: `runLoop.ticketTriage.test.ts` (9/9, new case asserts `skipped` lands in the persisted `outcome`), `runTrace.test.ts` (30/30, new cases for `ticketProposal.skipped`), `ticketTriageFindings.test.ts` (10/10, unaffected by the type move) — all green.
- Web: `RunDetailPage.test.tsx` (38/38, new skipped-section cases), `ApprovalsInbox.test.tsx` (43/43 — updated the existing `batch_not_homogeneous` test for the new deselect behavior, added a partial-drift regroup case and a no-offending-ids fallback case), `intentApprovals.test.ts` + `authenticator.test.ts` (new cases for the `offending` plumbing fix), `localeParity`/`translationCoverage`/`keyUsage`/`no-silent-mutations` — all green (350 web tests total across the affected files).
- `tsc --noEmit` clean in `apps/api`, `apps/web`, and `packages/shared`. `astro check`: 0 errors. `eslint` clean on every touched file.


ToddHebebrand:
**Review run:** `/code-review` (medium) — 4 parallel angles: cross-file tracer, removed-behavior auditor, reuse/simplification/efficiency, altitude/conventions, plus a line-by-line diff scan.

**Findings:** 1 confirmed and fixed, several considered and consciously left as-is:
- **Fixed (`8083d2876`):** three of the four review passes independently flagged the same spot — `driftedIds` was being derived inline via `Object.entries(rowErrors).filter(...)` in JSX on every render instead of being its own state, unlike the component's established `decidingIds` idiom. Promoted it to a real `useState` kept in sync with `rowErrors` by `clearRowErrors` and `decideGroup`'s `batch_not_homogeneous` branch.
- **Considered, not changed:**
  - A drifted card never automatically rejoins grouping on the 30s poll refresh even if the underlying drift was transient — it stays a standalone row until the approver decides it (or reloads). This matches the app's existing convention that row errors persist until an explicit decide, not until the next poll (true for `alreadyDecided`/`expired`/etc. too) — not a regression.
  - `fieldsSkipReason`'s reason-collapsing (a mixed `human_set` + `below_confidence_floor` run reports only the coarser reason) is pre-existing, already-documented behavior in `ticketTriageFindings.ts`, unchanged by this PR — just relocated into `@breeze/shared`.
  - `driftedFromBatch` sharing the `DecisionErrorKind` union with "a decide attempt failed" kinds is a minor semantic looseness; splitting it out would require duplicating the row's error-box render logic for one extra kind — not worth the added surface for a cosmetic type-purity gain.

**Tests:** API — `runLoop.ticketTriage.test.ts` (9/9), `runTrace.test.ts` (30/30), `ticketTriageFindings.test.ts` (10/10). Web — `RunDetailPage.test.tsx` (38/38), `ApprovalsInbox.test.tsx` (43/43, re-verified after the driftedIds fix), `intentApprovals.test.ts` + `authenticator.test.ts` (42/42), `localeParity`/`translationCoverage`/`keyUsage`/`no-silent-mutations` (227/227). `tsc --noEmit` clean in `apps/api`, `apps/web`, `packages/shared`; `astro check` 0 errors; `eslint` clean on every touched file.

Issue #4678, opened 2026-09-03
Script custom-field write-back (#2698)


ToddHebebrand:
Tracking issue for **Script custom-field write-back (#2698)** (`script-custom-field-writeback`).

- Plan: `docs/superpowers/plans/misc/2026-09-02-script-custom-field-writeback.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All three waves merged: W01 #4692 (API ingest + per-field script_write gate + audit), W02 #4722 (web toggle + run-detail panel + docs), W03 #4781 (Go agent structured channel, marker extraction pre-sanitizer). Unreleased; ships with the next release. Not verified: end-to-end marker round-trip on a real agent build (dev host cannot run cgo) — check during pre-release smoke.

Issue #2698, opened 2026-07-21
Feature request: allow scripts to write back custom field values on the executing device


DeusMaximus:
### Problem

There's currently no supported way for a script running on a device to update that device's own custom fields. The only write path is `PATCH /api/v1/devices/:id` with a `customFields` object, which requires an API token and the device's own UUID — neither of which a script has access to today. This makes it impossible to use Custom Fields for the thing they're most useful for: capturing data a script discovers about the device and persisting it as structured, queryable metadata.
 
Custom Fields are the mechanism for extending an RMM beyond its built-in inventory — tracking whatever your environment actually needs that the vendor didn't ship a column for. In practice, that almost always means "a script figures something out, and a custom field remembers it." A few concrete examples:
 
- **RAM slot inventory** — total DIMM slots, populated vs. free, and installed module type (DDR4/DDR5, speed) per device. Not part of standard hardware inventory, trivially discoverable with a script, genuinely useful for planning upgrades.
- **Secure Boot certificate expiry tracking** — flagging devices approaching relevant Secure Boot certificate expirations so they can be prioritized for a CA update before they start failing to boot with new UEFI/OS media.
- **Feature update readiness / upgrade experience indicators** — a script determines whether a device is eligible for or blocked from a Windows feature update (driver holds, compatibility holds, etc.) and needs somewhere to persist that per-device.
None of this works if the script can compute the fact but has no way to store it against the device.
 
There are actually two gaps here, and they compound each other:
 
1. **No write-back primitive.** There's no agent-injected device ID, no scoped self-write token, and no helper cmdlet/function exposed to scripts for setting a custom field on the device they're running on. The only door is the general device PATCH endpoint, which assumes an out-of-band API token and a UUID the script doesn't have.
2. **No custom-field-aware querying or targeting, even with the prior gap worked around.** `GET /api/v1/devices` doesn't support filtering by custom field value server-side (current docs say to pull the full device list and filter client-side, or query the `custom_fields` JSONB column directly in Postgres). Automation device targeting only supports Site / Group / OS / Tag conditions — there's no way to target "devices where `secure_boot_cert_expiry` < 30 days" as a native automation trigger.
Put together: even a determined workaround (script calls the device PATCH endpoint itself with an embedded token) only closes half the loop. There's still no way to act on the resulting data from inside Breeze — you'd need an external poller hitting the API and driving remediation from outside the platform entirely, which defeats a lot of the value of having this live in the RMM.
 
[...]

PR #4692, opened 2026-09-03
feat(scripts): write device custom fields from a script — API ingest (#4679)

### LINKED ISSUE #4679: W01: API ingest of versioned stdout marker + per-field script_write gate + audit
Wave W01 of #4678 — API ingest of versioned stdout marker + per-field script_write gate + audit
Branch: `feature/4678-script-custom-field-writeback/wave-4679`

ToddHebebrand:
Wave 1 of **Script custom-field write-back** (#4678, origin issue #2698): a script running on a device can persist values into that device's own custom fields — no API key, no device UUID, no new network call.

**API-only, so it works against every agent version already in the fleet.**

Closes #4679

## How it works

A script emits a marker line on stdout:

```
::breeze:custom-fields:: {"ram_slot_type":"DDR5-5600","free_dimm_slots":2}
```

`handleScriptResult` — the single handler both the WebSocket and REST result transports dispatch — extracts it, validates each key against `custom_field_definitions`, coerces by declared type, merges into `devices.custom_fields`, audits, and records a per-execution summary.

Channel B (`result.customFieldWrites`, `schemaVersion: 1`) is accepted from day one and takes precedence; the agent starts producing it in Wave 3. It is namespaced *and* versioned deliberately — `toWSCommandResult` reparses whole-JSON stdout into `result`, so an unnamespaced `customFields` key would turn any script printing such a document into an unintended write-back.

## Authorization

Two independent gates:

1. **Device scope is structural, not checked.** `deviceId` comes from the transport that already authorized the command row. Neither wire channel has a field that can name a device, so "a script may only write fields of the device it runs on" holds by construction. The UPDATE still carries a redundant `org_id` predicate, and an integration test names a foreign `deviceId` directly to prove RLS wins if the two ever disagree.
2. **Field scope is opt-in per definition.** New `custom_field_definitions.script_write`, `DEFAULT false` — no existing field becomes script-writable on deploy. An admin turns it on per field.

No dispatcher-permission check at ingest: that path has no user. The gate that matters is at definition time, already behind `devices:write` + `canManagePartnerWidePolicies`.

## Two things that would have been silent bugs

**Partner-wide definitions.** `custom_field_definitions` is dual-axis, and the caller runs under `runWithAgentOrgDbAccess` (`accessiblePartnerIds: []`), where the RLS policy hides every `org_id IS NULL` row. A partner defining one field for all its orgs would have had *no* script-writable fields at all. The definitions read therefore runs `runOutsideDbContext(() => withSystemDbAccessContext(...))` — the only form that genuinely opens a second context. The device UPDATE deliberately stays in the ambient org context so `devices` RLS remains a real backstop.

Verified non-vacuous by mutation: replacing that wrapper with a plain call turns exactly the three partner-wide-dependent integration tests red, and restoring it turns them green.

**The per-org advisory lock.** Every `devices` UPDATE that actually changes `custom_fields` fires `breeze_partner_export_z_custom_values_update` → `pg_advisory_xact_lock(1000201, hashtext(org_id))`, an EXCLUSIVE per-org lock held to COMMIT. A fleet-wide script on 5,000 devices in one org would serialize behind it. Two mitigations, both load-bearing: the marker parse is pure and runs *first* (a result with no marker costs zero DB work), and an unchanged merge skips the UPDATE entirely.

## Cascade registration

No new tables, so no cascade-list changes. But **two new columns land on tables already in `CORE_ORG_CASCADE_DELETE_ORDER`, and that fires `CORE_TENANT_EXPORT_POLICY`** — registered in the same commit as the migration:

| Column | Bucket | Why |
|---|---|---|
| `custom_field_definitions.script_write` | `included` | Ordinary boolean flag |
| `script_executions.custom_field_result` | `excludedOpen` | jsonb — every `json`/`jsonb`/`bytea` column is, no exceptions |

## Caps and failure reporting

At most 20 marker lines, 50 distinct keys, 8192 bytes of JSON per line. A line over a cap or failing `JSON.parse` is reported as a `rejected` entry, **never silently dropped**.

Known limitation, documented in the parser's file comment: the agent's `SanitizeOutput` and the server's `redactSecretsFromOutput` rewrite `(api_key|token|secret|password|…)\s*[=:]…` pairs to `[REDACTED]` *before* the marker reaches the server, so such a marker fails to parse. Wave 3 fixes it (extraction from raw stdout, pre-sanitizer); until then it surfaces loudly as `marker_unparseable` rather than mysteriously.

**`custom_fields` is not a secrets store.** The marker rides stdout, which is persisted to `script_executions.stdout` and visible to any `scripts:read` user. Stated in the parser's file comment; Wave 2 documents it for users.

## Not in this PR

Wave 2 (web toggle + docs), Wave 3 (agent structured channel), Wave 4 (output binding). The `script_write` toggle is exposed on the API create/update routes here — otherwise Wave 1 would ship a gate nobody can turn on — but has no UI until Wave 2.

## Verification

| Check | Result |
|---|---|
| `apps/api` unit (customFields, commandResultHandlers, routes/customFields, autoMigrate, tenantExportPolicy) | 186 passed / 13 files |
| `agentWs` regression suites | 137 passed / 4 files |
| Integration, real Postgres (`scriptCustomFieldWriteBack`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade`) | 16 passed / 4 files |
| RLS contract (`vitest.config.rls.ts`, incl. `rls-coverage`) | 26 passed, 5 skipped (pre-existing, unrelated suite) |
| `tsc --noEmit` — `apps/api`, `apps/web` | clean |
| `pnpm db:check-drift` | no drift, 636 migrations match ledger |

All re-run after merging `origin/main` (which had gained #4585 and #4604), since CI tests the merge commit.


ToddHebebrand:
**Review run:** three Sonnet reviewers in parallel — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`.

**Findings:** 3 substantive → all fixed in d842c16e5; 3 accepted-and-documented; 0 outstanding.

`code-reviewer` returned **0 findings**, having explicitly verified the export-policy classification of both new columns, the `runOutsideDbContext(() => withSystemDbAccessContext(...))` form, the provenance of `resolvedDeviceId` back to the authenticated WS connection, and the migration's idempotency and ordering.

**Fixed:**

- **Envelope failures were silent** (`scriptWriteMarkers.ts`). `readEnvelope` returned a bare `null` for a wrong `schemaVersion` or a malformed payload, so a version-skewed agent emitting no stdout fallback was indistinguishable from a script with nothing to say — the module's own "never silent" contract, broken on its own second channel. It now separates *absent* (normal, stays silent) from *failed*, reporting `envelope_unsupported_version` / `envelope_unparseable`. It still falls through to the stdout marker: during a rollout an agent may emit both, and a channel the API understands beats refusing the write.
- **The summary was stored but unreachable.** `GET /scripts/executions/:id` uses an explicit column list that omitted `custom_field_result`, so nothing in the product could read what was applied or rejected. Added — Wave 2 renders it. (Caught by `silent-failure-hunter`; a good catch, since the storage layer looked complete in isolation.)
- **A dropped `__proto__`-class key left no record.** Now reported as `forbidden_key`.

**Test gaps closed** (from `pr-test-analyzer`, which found no vacuous assertions):

- Sibling-org isolation for an **org-scoped** definition — the case the fixture already set up but never attacked from org B's side. Mutation-verified: replacing the org predicate with `isNotNull(orgId)` reddens it, and restoring it greens it.
- `Number(true) === 1` and `Number('') === 0` coercion guards, plus non-finite input. The empty-string guard is mutation-verified.
- Both `device_not_found` paths (invisible row; UPDATE matching no row), asserting `applied` reverts to `[]` and no audit is written.
- The cross-org integration assertion now pins the rejection *reason*, so a regression reports which layer stopped refusing.

**Accepted and now documented in-code, rather than left ambiguous:**

- No audit row when nothing was applied — an audit event records a *change*, and there was none. The rejection lives on `script_executions.custom_field_result` (now readable) plus a `console.warn`. Auditing every rejected marker would add an agent-driven row per script run to a table already dominated by agent telemetry.
- Read-modify-write on `devices.custom_fields` with no optimistic concurrency. Both `code-reviewer` and `pr-test-analyzer` raised it independently; it mirrors the pre-existing PATCH value endpoint exactly, is self-healing on the next run, and a version column would be a device-wide contention point worse than the rare lost update.
- The summary is discarded (not partially persisted) when the write-back throws. A half-built summary would misreport what landed; Sentry still gets the error.

**Tests:**

| Check | Result |
|---|---|
| `apps/api` unit — customFields, commandResultHandlers, routes/customFields | 114 passed / 11 files |
| `apps/api` unit — + `routes/scripts.test.ts` | 164 passed, 2 skipped / 8 files |
| `agentWs` regression suites | 137 passed / 4 files |
| Integration, real Postgres — `scriptCustomFieldWriteBack` (8), `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade` | 17 passed / 4 files |
| RLS contract (`vitest.config.rls.ts`, incl. `rls-coverage`) | 26 passed, 5 skipped (pre-existing, unrelated suite) |
| `tsc --noEmit` — `apps/api`, `apps/web` | clean |
| `pnpm db:check-drift` | no drift |

Two controls were mutation-verified rather than assumed: removing the system-context wrapper reddens exactly the three partner-wide tests, and dropping the org predicate reddens the new isolation test.

**Status:** review-clean, awaiting maintainer merge. Not merged and #4679 left open by design.

PR #4698, opened 2026-09-03
fix(web): quote editor line grid no longer scrolls at 1280px with sidebar open (#4668)

### LINKED ISSUE #4668: [Web][Quotes] Quote editor line grid horizontally scrolls at 1280px with the sidebar open — needs column-priority restructuring
Carved out of #2151 (the only remaining item from that review round).

At 1280px viewport with the left sidebar open, the quote editor line grid still scrolls horizontally to reach the **Total** column; it fits with the sidebar collapsed. A real fix needs column-priority restructuring (drop/collapse low-priority columns first: e.g. SKU/description-secondary before Qty/Unit/Total), not a narrower padding tweak.

Acceptance: at 1280px with sidebar open, Total is visible without horizontal scroll; no regression at 1440/1920.

ToddHebebrand:
## Summary
- At 1280px viewport with the left nav sidebar expanded, the quote editor's pricing table horizontally scrolled to reach the **Total** column. Root cause: the table's blanket `min-w-[36rem]` (576px) plus the Item column's own `min-w-[12rem]` (192px) floor summed to exactly the ~576px actually available at that viewport/sidebar combination, so any rendering variance (scrollbar, border, subpixel rounding) tipped it into `overflow-x-auto`.
- Restructured column priority: Item is the lowest-priority column (it already degrades gracefully via its own text input and `title` tooltip on overflow) — its floor drops from `12rem` to `8rem`, and the table's blanket floor drops from `36rem` to `26rem`. Qty/Unit Price/Total/actions are untouched — they were never the columns that needed to shrink; they're already content-sized/fixed-width and were never at risk of squeezing.
- Updated the adjacent layout comments (`QuoteEditor.tsx`'s `xl:grid-cols-[1fr_300px]` rail-breakpoint note, `QuoteBlockCard.tsx`'s column doc, `QuoteLineRows.tsx`'s cell doc) so the numbers and reasoning stay accurate for the next person who touches this.

Closes #4668

## Test plan
- [x] New regression test `apps/web/src/components/billing/quotes/QuoteBlockCard.gridwidth.test.tsx` — red-first (confirmed failing against the old `min-w-[36rem]`/`min-w-[12rem]` classes), now green. jsdom has no real layout engine so it can't assert on-screen pixels/scroll state directly; it locks the column-priority contract in Tailwind-class form (table's blanket floor shrunk; Item's floor shrunk on both header `<th>` and body `<td>`, catching a header/body mismatch too).
- [x] Full quotes test suite: `cd apps/web && npx vitest run src/components/billing/quotes` — 58 files / 510 tests passed.
- [x] `npx tsc --noEmit` — clean.
- [x] `npx eslint` on all touched files — clean.
- [ ] **Manual verification (not run here — no browser/dev-server access in this session):** apps/web dev server, open a quote in the editor at exactly 1280px viewport with the left nav sidebar expanded — Total should be visible with no horizontal scrollbar on the pricing table. Re-check 1440px and 1920px for no regression (both already had comfortable margin before this change and gain more here).

### DIFF
--- a/apps/web/src/components/billing/quotes/QuoteBlockCard.gridwidth.test.tsx
+++ b/apps/web/src/components/billing/quotes/QuoteBlockCard.gridwidth.test.tsx
@@ -1,0 +1,126 @@
+// Regression guard for #4668: at 1280px viewport with the left nav sidebar
+// open, the quote editor's pricing table horizontally scrolled to reach the
+// Total column (it fit fine with the sidebar collapsed). Root cause: the
+// pricing table carried a blanket `min-w-[36rem]` floor plus a `min-w-[12rem]`
+// floor on the Item column — together wider than the ~576px actually
+// available for the table at that viewport/sidebar combination (1280 −
+// 256px sidebar − 48px page padding − 300px rail − 24px grid gap − 76px
+// canvas padding), so the table's declared minimum exceeded its container by
+// a hairline and `overflow-x-auto` kicked in on any rendering variance
+// (scrollbar, border, subpixel rounding).
+//
+// jsdom has no real layout engine, so this can't assert on-screen pixels or
+// scroll state directly. Instead it locks the column-priority contract in
+// Tailwind class form: Item (the lowest-priority column — it already
+// degrades via its own text input / title tooltip) carries the smaller
+// floor and shrinks first, while the table's blanket floor is reduced enough
+// that Qty/Unit Price/Total/actions are never squeezed below their natural
+// content width. See QuoteBlockCard.tsx / QuoteLineRows.tsx for the full
+// math. Manual verification: apps/web dev server, quote editor at 1280px
+// viewport with the sidebar expanded — Total is visible with no horizontal
+// scrollbar on the pricing table; re-check at 1440/1920 for no regression.
+import { render, screen, waitFor, within } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import QuoteEditor from './QuoteEditor';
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (Sonnet), verified the specific pinch-point claim against a real Chromium layout render (Playwright) at the container widths from the PR's own math (576px / 416px), not just jsdom class assertions.

**Findings:** 0 confirmed issues (checked: are the reduced 8rem/26rem floors safe or does Item get squeezed/overlap at the pinch-point width; does `ReadonlyLineRow`'s first `<td>` need the same treatment; does the new test prove anything real; any other correctness/regression risk). One minor test-quality observation noted but explicitly below the report bar (not acted on per single-review-round policy): the new test asserts the *old* `min-w-[36rem]`/`min-w-[12rem]` literals are gone rather than asserting the *new* values are small enough — a fix that shrank the floors only trivially would still pass. Not blocking; jsdom's lack of a layout engine limits how much stronger this assertion can get without a real browser.

**Tests:** `cd apps/web && npx vitest run src/components/billing/quotes` — 58 files / 510 tests passed (includes the new `QuoteBlockCard.gridwidth.test.tsx`, red-first confirmed against the old classes before the fix). `npx tsc --noEmit` clean. `npx eslint` on all touched files clean.

**Status:** review-clean, awaiting maintainer merge. Manual verification (1280px viewport, sidebar expanded, no browser access in this session) still needed before merge — noted in the PR description's test plan as unchecked.

Issue #3253, opened 2026-08-08
[Agent] `schedule_reboot` `deadline` is stored and reported but never enforced


ToddHebebrand:
## Description

`schedule_reboot` accepts and propagates a `deadline`, stores it in `RebootState`, returns it in the command result — and nothing ever reads it. There is no hard-deadline enforcement anywhere in the reboot path.

Found while investigating #3197.

## Current state

- `agent/internal/heartbeat/handlers_patch.go:212` computes `deadline := time.Now().Add(delay)` — i.e. **identical to the reboot time**, so the default carries no information.
- `handlers_patch.go:215-219` allows an RFC3339 `deadline` override from the payload. A parse failure is silently swallowed (the code only checks `err == nil`), so a malformed deadline degrades to the default with no error and no log.
- `RebootManager.Schedule` stores it at `agent/internal/patching/reboot_windows.go:87`.
- The only timer created is `time.AfterFunc(delay, ...)` at `reboot_windows.go:93`. **`state.Deadline` is never read** anywhere in the tree.
- It is surfaced to the server via `rebootStateToMap` (`handlers_patch.go:256-266`), so the API receives a field that looks authoritative and isn't.

No server-side caller sends `deadline` today (`apps/api/src/services/patchRebootHandler.ts:79-83`, `apps/api/src/jobs/maintenanceRebootWorker.ts:76-80`).

## Why it matters

A "deadline" in an RMM reboot context reads as *the reboot will not be postponed past this point* — which is exactly the guarantee #3207's deferral feature needs, and exactly what this field does not provide. Leaving a decorative field named `deadline` in the payload and the state is a trap for whoever builds deferral on top of it.

## Options

1. **Enforce it** — a second timer at `deadline` that fires the reboot regardless of deferrals, with the deferral logic in #3207 clamping to it. This is the version #3207 wants.
2. **Remove it** — drop from the payload, `RebootState`, and `rebootStateToMap`, and let #3207 add a real one when it needs it.

Either is fine; shipping #3207 on top of the current no-op is not. Whichever way it goes, the silent RFC3339 parse failure at `handlers_patch.go:216` should return an error rather than falling through.

## Affected files

- `agent/internal/heartbeat/handlers_patch.go` (:212-219, :256-266)
- `agent/internal/patching/reboot_windows.go` (:87, :93)
- `agent/internal/patching/reboot_other.go` (stub — see #3197, which makes this a real implementation)

ToddHebebrand:
[...]

PR #4699, opened 2026-09-03
feat(agent): reboot deferral state machine (#3207 W2)

### LINKED ISSUE #4662: W02: Agent deferral state machine in RebootManager (untagged core, persisted ledger, schedule_reboot payload)
Wave W02 of #3207 — Agent deferral state machine in RebootManager (untagged core, persisted ledger, schedule_reboot payload)
Branch: `feature/3207-reboot-deferral/wave-4662`

ToddHebebrand:
Closes #4662. Wave 2 of #3207 (end-user reboot prompts with deferral).

The agent's `RebootManager` gains a deferral state machine: it understands the
budget W1 started shipping on the `schedule_reboot` payload, can execute a
postponement, clamps every one against the hard deadline, and persists the count
across agent restarts. **Nobody can trigger a deferral yet** — the prompt is W3.
This wave exists so the state machine is provably inert when the policy is off
*before* any UI can drive it.

## What changed

| File | Why |
|---|---|
| `agent/internal/patching/reboot_deferral.go` *(new, untagged)* | `DeferralPolicy`, `ComputeDeferral` (pure clamp arithmetic), `DeferralLedger` persistence |
| `agent/internal/patching/reboot_manager.go` | `RebootOptions`, `ScheduleWithOptions`, `Defer()`, four deferral fields on `RebootState` |
| `agent/internal/heartbeat/handlers_patch.go` | parse `allowDeferral` / `maxDeferrals` / `deferralMinutes`; strict `deadline` parse |
| 3 new `_test.go` files | table-driven coverage of the clamp, the ledger, the manager and the payload |

### The deadline is the guarantee, the counter is an affordance

`RebootState.Deadline` has been stored and reported since #3421 but never
enforced (#3253). A postponement is now
`min(scheduledAt + deferralMinutes, deadline)`, and one that would leave less
than `MinRebootDelay` is refused outright — so a maintenance-window reboot
cannot be postponed past the close of its window no matter how much counter
budget is left.

**The window is added to the scheduled time, not to now.** Review caught this
independently twice: measured from `now`, a 120-minute reboot delay postponed by
a 60-minute window moved the restart an *hour earlier*, and every test in the
first draft masked it by using a window larger than the remaining countdown.
It is also the semantics the merged API already assumed — it derives the
deadline as `scheduledAt + maxDeferrals * deferralMinutes`, so spending the
whole budget lands exactly on the deadline rather than short of it.

A property test sweeps the deadline across an 8-hour range × six scheduled
times and asserts a granted postponement never lands past the deadline and
never lands earlier than the restart already scheduled.

### Backward compatibility, both directions

- **Old agent, new API** — the new keys ride a `map[string]any` payload and are
  ignored by construction. Pinned by `TestScheduleRebootIgnoresUnknownPayloadKeys`.
- **New agent, old API** — the keys are absent, `GetPayloadBool(..., false)`
  yields `false`, deferral is off, `deadline` falls back to `now + delay`:
  today's behaviour exactly. `allowDeferral` is the *only* switch, so a payload
  carrying a budget without the flag (stale or forged) also stays off.
- **`Schedule(...)` keeps its exact 4-arg signature and semantics** and delegates
  to `ScheduleWithOptions` with a zero `RebootOptions`. Every existing call site
  and every existing test keeps today's behaviour with **no edit** — the #3197
  invariant suites (`reboot_manager_test.go`, `reboot_plan_test.go`) are
  byte-for-byte unmodified on this branch, which is why the new manager tests
  live in their own file.

### Deferral does not touch the circuit breaker

`rebootHistory` is appended only inside `runOSReboot`, and deferring never
reaches it (D6). `TestDeferDoesNotConsumeCircuitBreakerBudget` runs three
deferrals against a `maxRebootsPerDay: 1` manager and asserts the history is
still empty and the eventual reboot still fires.

### Concurrency

`Defer()` holds `mu` from its first check through the ledger write. Splitting
that sequence — which the plan's shape does — opens two holes, both found and
then closed on this branch:

- **Releasing `mu` before re-entering `ScheduleWithOptions`** lets a concurrent
  `Cancel()` succeed in the gap, after which the re-schedule **resurrects the
  reboot the operator just cancelled** — the machine goes down after the console
  said it would not. Proven rather than argued: widening the gap with a
  one-millisecond sleep made `TestDeferNeverResurrectsACancelledReboot` fail on
  iteration 0 with `RebootScheduled: true` after a successful `Cancel`. The same
  probe moved inside the atomic section passes.
- **Writing the ledger outside `mu`** lets a concurrent schedule for the same
  campaign reload the stale on-disk count and hand back a postponement already
  spent. Measured at 141 disagreements in 300 trials;
  `TestConcurrentDefersLeaveTheLedgerAgreeingWithMemory` now pins it.

`TestConcurrentDefersRespectTheBudget` runs 8 goroutines against
`MaxDeferrals: 2` under `-race` and asserts exactly 2 grants.

### Refusals that matter

- After `OSInvokeAt` the countdown lives in the OS and macOS BSD `shutdown(8)`
  has no cancel flag, so `Defer()` is refused there rather than reporting a
  success the machine would contradict — the same class of lie `Cancel()`
  documents.
- After `Cancel()` or `Stop()`, and when nothing is scheduled.
- A later plain `Schedule` clears the budget: a fresh administrator decision
  carries no allowance from the previous campaign.

### Ledger

Persisted next to `reboot_history.json`, keyed by the deadline truncated to the
minute, so restarting the agent mid-countdown does not refill the budget. Sub-minute
jitter in the re-sent ISO timestamp is the same campaign; a different deadline is
a different administrator decision and legitimately starts fresh. Missing,
corrupt, expired or negative all mean **zero used** — best-effort, never an error
that blocks a reboot. Losing it costs the user one extra postponement; it can
never push past the deadline, which is enforced separately.

## Deviations from the plan

1. **`ComputeDeferral` takes the currently scheduled time** and adds the window
   to *it*, rather than the plan's `min(now + deferralMinutes, deadline)`. See
   above — the plan's arithmetic makes a postponement pull the restart earlier,
   and disagrees with the deadline W1 already ships.
2. **`Defer()` is one atomic `mu` section** covering the checks, the
   re-schedule, the increment *and* the ledger write (the plan released `mu`
   between the checks and the re-schedule, and wrote the ledger outside it). Two
   distinct bugs came out of that gap — see the concurrency section below. The
   cost is a small file write under the lock, which briefly blocks `State()` and
   the timer callbacks; that is the right trade against a budget the user can
   exceed.
3. **The ledger is written atomically** (temp file, `fsync`, `rename`, mode
   0600) rather than a plain `os.WriteFile`. A torn write reads back as
   corrupt — which means "zero used", quietly handing the user their whole
   budget again.
4. **`Cancel()` also clears the deferral budget** (plan was silent). Keeps the
   reported and internal copies from disagreeing if a prompt answer arrives
   after the cancellation.
5. **Manager tests live in `reboot_manager_deferral_test.go`**, not appended to
   `reboot_manager_test.go` — "with deferral disabled nothing changes" is only a
   real claim if the #3197 suite never had to be adjusted.
6. **The plan's sample heartbeat tests assert `res.Status == "success"`; the real
   constructor emits `"completed"`.** Tests use the actual value.
7. Payload range bounds are named constants mirroring
   `patchRebootHandler.ts`'s `MAX_REBOOT_DEFERRALS` / `MIN|MAX_REBOOT_DEFERRAL_MINUTES`
   rather than inline literals. Same values (1-10, 5-1440); deliberately
   duplicated, not derived, because the agent must reject a budget the API could
   not have produced whatever version sent it.
8. Extra coverage beyond the plan's list: ledger resume across process lifetimes,
   unknown-key tolerance, budget-without-flag, malformed `deadline`, `Cancel`/`Stop`
   refusals, and the `get_reboot_status` JSON surface.

Also hardens the `deadline` parse, which has swallowed malformed values since
#3421. That was cosmetic while nothing read `Deadline`; it now silently collapses
the deferral budget to nothing, so it fails the command instead.

## Test evidence

**verified** (all Go 1.26.6 = `GO_VERSION`, in a `golang:1.26.6` Linux container,
because this host has a pre-existing cgo crash — see below):

- `go test -race -count=1 ./internal/heartbeat/ ./internal/patching/` → **both ok**
  (heartbeat 29.7s, patching 1.8s), re-run after every review fix. 30 new test
  functions across the three new files.
- `CGO_ENABLED=0 go test -count=1 ./...` (the **test-agent** CI job's exact
  command) → green except `TestCollectHardwareOnly` (no CPU model in a container)
  and `TestConfigureRunAsElevatedWhenNotRoot` (runs as root, so no sudo path).
  **Both reproduce identically on the unmodified base checkout in the same
  container** — environment, not this change.
- `go vet ./...` → clean.
- `golangci-lint v2.12.2` (the CI pin) over `./internal/patching/... ./internal/heartbeat/...`
  → **0 issues in any of the six touched files**; the 48 remaining are the
  pre-existing backlog CI already excludes via `--new-from-rev`.
- `GOOS=windows|darwin|linux go build ./...` → all clean.
- Red-first: every test in this PR was run and observed failing (compile-undefined
  for the two new APIs; 10 assertion failures for the payload parsing) before the
  implementation landed.

**not-run:**

- Any macOS/Windows runtime behaviour — this wave adds no platform-specific code
  and every new file is untagged (`grep -L 'go:build'` returns all four).
- End-to-end deferral: there is no way to trigger one until W3 ships the prompt.
- `-race` on this dev host: `go-m1cpu`'s cgo `init` segfaults here on **both**
  Go 1.26.6 and 1.27 (reproduced on unmodified `main` with `-run XXXNoSuchTest`),
  so every race run above was done in the Linux container instead. CI's
  `test-agent-race` job runs on `macos-latest` and is unaffected.

Cross-refs #3253 without closing it — the deadline is now enforced for
deferrals; the broader "deadline should bound the schedule generally" question
stays open.

### DIFF
--- a/agent/internal/heartbeat/handlers_patch.go
+++ b/agent/internal/heartbeat/handlers_patch.go
@@ -11,6 +11,19 @@ import (
 	"github.com/breeze-rmm/agent/internal/remote/tools"
 )
 
+// Accepted bounds for the schedule_reboot deferral budget (#3207). They mirror
+// the API-side CHECK constraints and zod validator exactly
+// (apps/api/src/services/patchRebootHandler.ts: MAX_REBOOT_DEFERRALS,
+// MIN/MAX_REBOOT_DEFERRAL_MINUTES). Duplicated rather than derived because the
+// agent must reject a payload the API could not have produced, whatever version
+// of the API sent it.
+const (
+	minPayloadDeferrals       = 1
+	maxPayloadDeferrals       = 10
+	minPayloadDeferralMinutes = 5
+	maxPayloadDeferralMinutes = 1440
+)
+
 func init() {
 	handlerRegistry[tools.CmdPatchScan] = handlePatchScan
 	handlerRegistry[tools.CmdInstallPatches] = handleInstallPatches
@@ -228,14 +241,51 @@ func handleScheduleReboot(h *Heartbeat, cmd Command) tools.CommandResult {
 	delay := time.Duration(delayMinutes) * time.Minute
 	deadline := time.Now().Add(delay)
 
-	// Allow overriding deadline via payload
+	// Allow overriding deadline via payload. Strict since #3207: a malformed
+	// deadline used to be swallowed, leaving deadline = now+delay, which now
+	// silently collapses the deferral budget to nothing. Fail the command
+	// instead of quietly delivering a different policy than the API asked for.
 	if deadlineStr := tools.GetPayloadString(cmd.Payload, "deadline", ""); deadlineStr != "" {
-		if parsed, err := time.Parse(time.RFC3339, deadlineStr); err == nil {
-			deadline = parsed
+		parsed, err := time.Parse(time.RFC3339, deadlineStr)
+		if err != nil {
+			return tools.NewErrorResult(fmt.Errorf("invalid deadline %q: %w", deadlineStr, err), time.Since(start).Milliseconds())
+		}
+		deadline = parsed
+	}
+
+	// Deferral budget (#3207). Absent keys mean OFF: an old API never sends
+	// them, and "not mentioned" must never widen what the agent will do.
+	// Ranges mirror the API-side CHECK constraints so a forged or corrupted
+	// payload is REJECTED rather than clamped — a silent clamp is how #3373
+	// turned a malformed delayMinutes into a 60-minute reboot.
+	opts := patching.RebootOptions{}
+	if tools.GetPayloadBool(cmd.Payload, "allowDeferral", false) {
+		maxDeferrals, err := tools.ParsePayloadInt(cmd.Payload, "maxDeferrals", 0)
+		if err != nil {
+			return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+		}
+		deferralMinutes, err := tools.ParsePayloadInt(cmd.Payload, "deferralMinutes", 0)
+		if err != nil {
+			return tools.NewErrorResult(err, time.Since(start).Milliseconds())
+		}
+		if maxDeferrals < minPayloadDeferrals || maxDeferrals > maxPayloadDeferrals {
+			return tools.NewErrorResult(
+				fmt.Errorf("maxDeferrals must be %d-%d, got %d", minPayloadDeferrals, maxPayloadDeferrals, maxDeferrals),
+				time.Since(start).Milliseconds())
+		}
+		if deferralMinutes < minPayloadDeferralMinutes || deferralMinutes > maxPayloadDeferralMinutes {
+			return tools.NewErrorResult(
+				fmt.Errorf("deferralMinutes must be %d-%d, got %d", minPayloadDeferralMinutes, maxPayloadDeferralMinutes, deferralMinutes),
+				time.Since(start).Milliseconds())
+		}
+		opts.Deferral = patching.DeferralPolicy{
+			Allowed:         true,
+			MaxDeferrals:    maxDeferrals,
+			DeferralMinutes: deferralMinutes,
 		}
 	}
 
-	if err := h.rebootMgr.Schedule(delay, deadline, reason, source); err != nil {
+	if err := h.rebootMgr.ScheduleWithOptions(delay, deadline, reason, source, opts); err != nil {
 		return tools.NewErrorResult(err, time.Since(start).Milliseconds())
 	}
 
--- a/agent/internal/heartbeat/handlers_patch.go
+++ b/agent/internal/heartbeat/handlers_patch.go
@@ -259,6 +259,15 @@ func handleScheduleReboot(h *Heartbeat, cmd Command) tools.CommandResult {
 	// payload is REJECTED rather than clamped — a silent clamp is how #3373
 	// turned a malformed delayMinutes into a 60-minute reboot.
 	opts := patching.RebootOptions{}
+	// GetPayloadBool swallows a non-boolean, which fails safe (deferral stays
+	// off) but is otherwise invisible. Say so, or "why does this device not
+	// offer postponements" has no signal at all to start from.
+	if raw, present := cmd.Payload["allowDeferral"]; present {
+		if _, ok := raw.(bool); !ok {
+			log.Warn("schedule_reboot allowDeferral is not a boolean; treating deferral as disabled",
+				"got", fmt.Sprintf("%T", raw))
+		}
+	}
 	if tools.GetPayloadBool(cmd.Payload, "allowDeferral", false) {
 		maxDeferrals, err := tools.ParsePayloadInt(cmd.Payload, "maxDeferrals", 0)
 		if err != nil {
--- a/agent/internal/heartbeat/handlers_patch_test.go
+++ b/agent/internal/heartbeat/handlers_patch_test.go
@@ -1,0 +1,193 @@
+package heartbeat
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/breeze-rmm/agent/internal/patching"
+)
+
+func newTestHeartbeatWithRebootManager(t *testing.T) *Heartbeat {
+	t.Helper()
+	mgr := patching.NewRebootManager(func(string, string, string) {}, 3)
+	t.Cleanup(mgr.Stop)
+	return &Heartbeat{rebootMgr: mgr}
+}
+
+func scheduleRebootResultMap(t *testing.T, res any) map[string]any {
+	t.Helper()
+	stdout, ok := res.(string)
+	if !ok {
+		t.Fatalf("result stdout is %T, want string", res)
+	}
+	var got map[string]any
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** four independent lenses on `390397eb0` — `pr-review-toolkit:code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` (Sonnet), plus a `codex gpt-5.6-sol` read-only bug-hunt. Agent-shipped code, so this got the wide pass rather than a single reviewer.

**Findings:** 12 raised → **7 fixed** in `aae2900cd`, 2 rejected with reasons, 1 out of scope, 2 informational.

**Fixed**

1. **A postponement could pull the restart *earlier*.** Found independently by *both* codex and the test analyzer, and the most valuable finding in the round. `ComputeDeferral` measured the new delay from `now`, so a 120-minute reboot delay postponed by a 60-minute window moved the restart an hour *closer* — the user pressed Postpone and lost an hour. Every test in the first draft masked it by using a window larger than the remaining countdown. The window is now added to the currently scheduled time, floored at `now` for an overdue schedule, still clamped to the deadline, and a deadline sitting behind the schedule is refused rather than clamped. This is also the semantics the merged API already assumed (`deadline = scheduledAt + maxDeferrals * deferralMinutes`), so the whole budget now lands exactly on the deadline instead of short of it.
2. **Two clock samples per deferral** — `ComputeDeferral` sampled, then `scheduleLocked` sampled again, placing the reboot at (second sample + a delay computed against the first). Just past the deadline whenever the deferral was clamped to it; arbitrarily past it on a wall-clock step. One sample now, threaded through `scheduleLockedAt`.
3. **Ledger write moved inside the lock.** Outside it, a concurrent schedule for the same campaign could reload the stale on-disk count and hand back a postponement already spent. The test analyzer measured **141 disagreements in 300 trials** with the write outside. `deferMu` is gone; `mu` covers the whole sequence.
4. **`fsync` before the rename** — a power loss could otherwise leave the ledger present but empty, and an empty file reads back as corrupt, i.e. a silently refilled budget.
5. Ledger-save warning now names the path, deadline and count (it logged only `error`, below the bar every other log line in the file meets).
6. A non-boolean `allowDeferral` is now logged. It still fails safe to "off", but silently — leaving "why does this device not offer postponements" with no signal at all.
7. Coverage gaps the analyzer named: ledger/memory agreement under concurrent postponements, `Defer` surviving an unwritable ledger (the best-effort contract was asserted nowhere), inclusive payload bounds `1/10/5/1440` (only rejections were tested, so an off-by-one would have gone unnoticed), and `Cancel` clearing the reported budget.

**Rejected, with reasons**

- *"Fail closed when the ledger cannot be read or written."* No — the deadline is the guarantee, not the counter. A ledger failure costs at most one extra postponement and can never push past the deadline, whereas failing closed makes a transient disk hiccup strand a user with an un-postponable restart. Deliberate, documented, and now asserted by test.
- *"Key the ledger on the exact deadline, not truncated to the minute."* Truncation is the conservative direction: exact matching would refill the budget on every re-dispatch, since the API recomputes the deadline from a fresh `now`. Intent is pinned by a test.

**Out of scope (pre-existing, not touched by this PR)**

- codex flagged a race in `runOSReboot`/`Cancel`: `osInvoked` is set before `execOSReboot` runs, so an abort arriving in that window finds nothing to abort and the reboot proceeds. That code is unchanged from #3197/#3421 and fixing it here would balloon the blast radius of a feature wave. **Worth a follow-up issue — flagging for your call rather than filing unilaterally.**

**Two notes on the review itself, so the transcripts read straight**

- The silent-failure hunter saw `TestDeferNeverResurrectsACancelledReboot` fail once on its first run and never again across 9,000+ iterations, and honestly reported it as unexplained. It is explained: it ran while I had a deliberate one-millisecond sleep in the tree, widening the race window to *prove* the resurrection bug was real before fixing it. Its build cache picked that up. Not a flake.
- The test analyzer likewise saw the working tree change under it mid-review; same cause.

**Tests (verified):** `go test -race -count=1 ./internal/heartbeat/ ./internal/patching/` → both ok (29.7s / 1.8s), Go 1.26.6 in a `golang:1.26.6` Linux container. Full `CGO_ENABLED=0 go test ./...` matching the `test-agent` job → green apart from two failures that reproduce identically on the unmodified base checkout in the same container (no CPU model in a container; running as root so no sudo path). `go vet ./...` clean. `golangci-lint v2.12.2` → 0 issues in the six touched files. `GOOS=windows|darwin|linux go build ./...` clean. Every test was observed failing before its fix.

**Not run:** any macOS/Windows runtime path (no platform-specific code added; all four new files untagged), and end-to-end deferral — nothing can trigger one until W3 ships the prompt. `-race` cannot run on the dev host at all: `go-m1cpu`'s cgo init segfaults there on both Go 1.26.6 and 1.27, reproduced on unmodified `main`, which is why every race run above was containerised.

**Status:** review-clean, awaiting maintainer merge. Do not merge before CI is green on `aae2900cd` — the head moved after the review fixes.

PR #4725, opened 2026-09-03
feat(api): populate currentPartnerId on agent DB contexts (#4675)

### LINKED ISSUE #4675: W02: Populate currentPartnerId on agent + hand-built org contexts, with real-DB fan-out proof
Wave W02 of #4673 — Populate currentPartnerId on agent + hand-built org contexts, with real-DB fan-out proof
Branch: `feature/4673-partner-current-partner-id-select/wave-4675`

ToddHebebrand:
Wave 2 of #4673 (origin #2468). Plan: `docs/superpowers/plans/tenancy-rls/2026-08-16-2468-partner-current-partner-id-select-branch.md`

## The bug this closes

Wave 1 ([on main](https://github.com/LanternOps/breeze/blob/main/apps/api/migrations/2026-10-05-110000-config-policy-partner-wide-select.sql)) shipped SELECT-only RLS branches across the configuration-policy chain:

```sql
USING (org_id IS NULL AND partner_id = public.breeze_current_partner_id())
```

That branch reads the `breeze.current_partner_id` GUC, which `db/index.ts` `SET LOCAL`s from `DbAccessContext.currentPartnerId`. **Every agent context hard-coded that field to `null`**, so on agent paths the branch could never be true — a partner-wide config row read as **zero rows, with no error**.

That silent-zero is precisely why the agent-facing resolvers must escape to a system context (`withPartnerWideVisibility`), which double-holds a pooled connection and bypasses RLS wholesale (#1105, #2417). This wave resolves the device org's owning MSP and threads it through every agent-authenticated DB context, so **Wave 3 can delete those escapes**.

## Changes

| File | Change |
|---|---|
| `middleware/agentAuth.ts` | Device auth select inner-joins `organizations` for `partner_id`. `AgentAuthContext` gains a required `partnerId`. The request-long org wrap sets `currentPartnerId` from it. |
| `routes/agents/heartbeat.ts` | In `SELF_MANAGED_DB_CONTEXT_ACTIONS` — skips that wrap and hand-builds its own context, so it reads `agent.partnerId` explicitly. **This is THE agent config-delivery path.** |
| `routes/agents/reliability.ts` | Same self-managed shape. |
| `routes/agentWs.ts` | Same join on the socket auth select; `AgentDbContext` gains `partnerId`; threaded through `runWithAgentOrgDbAccess` and `processCommandResult`'s five short-lived contexts. |

**The join is total, not a narrowing.** `devices.org_id` is `NOT NULL` with an FK to `organizations`, and `organizations.partner_id` is `NOT NULL`, so an INNER join can never drop an authenticable device. A LEFT join would have been worse — it would hand an orphan device a NULL partner and silently degrade it to the pre-W02 blind behaviour.

## Read-only by construction — and proven, not asserted

`accessiblePartnerIds` stays `[]` on every agent context. That array gates `breeze_has_partner_access`, the write-capable partner **AXIS**; `currentPartnerId` feeds `FOR SELECT` policies only. The two axes are deliberately not merged, and there's a unit tripwire if someone later merges them.

The new suite proves the read/write split rather than assuming it:
- a partner-wide row an agent **can read** is still **un-targetable** by UPDATE/DELETE — asserted as **0 rows affected**, since RLS hides rows from write commands silently rather than raising;
- and **un-forgeable** by INSERT — asserted on the **unwrapped pg `cause.code === '42501'`**, because drizzle wraps the driver error and a bare `rejects.toThrow(/42501/)` would match the wrapper's "Failed query:" text and pass for *any* failure.

## Deliberately NOT in this wave

- **`portal/auth.ts` and `clientAiAuth`** — client end-users seeing MSP-wide config is a product decision, not a mechanical sweep. Left `null`.
- **`commandQueue.ts`'s audit-log context** — a write-only `audit_logs` INSERT that reads no partner-wide table; a lookup there would cost a query on a hot path for zero benefit.
- The ~40 AI-agent-SDK / streaming-session contexts, which all already have `session.auth.partnerId` in scope — a separate, non-agent surface.

## Blast radius

**Behaviour is inert until Wave 3.** Every agent-path config reader either already escalates (so already saw these rows) or filters on `org_id = <device org>` / `level = 'organization'`, which partner-wide rows (`org_id NULL`, level `partner`) can never match — verified by sweeping `configurationPolicies` readers in `routes/agents/helpers.ts`. Org/user tokens were already widened by Wave 1 via `buildDbAccessContext`, so this PR's blast radius is **agent paths only**.

## Tests

**New: `agentContextPartnerWideFanout.integration.test.ts`** — the real-DB fan-out proof, against the RLS-forced `breeze_app` role.

It deliberately does *not* reuse `agentPolicyResolversPartnerWide.integration.test.ts`, which passes today with `currentPartnerId: null` because the resolvers still escalate internally — so that suite structurally cannot detect whether this wave works. The new suite issues **plain, un-escalated** queries and lets RLS alone decide.

Two guards make it non-vacuous:
1. **BYPASSRLS non-vacuity guard, first test.** A worktree missing `.env.test` runs as a superuser, under which every negative assertion passes trivially *and* the positives pass for the wrong reason. Fails loudly instead.
2. **The negative control** (load-bearing): the identical query under a context differing **only** by `currentPartnerId: null` returns zero rows, plus a direct assertion that `breeze_current_partner_id()` returns the partner vs `NULL`. Without this, every positive assertion could be satisfied by some unrelated policy widening.

Also covers: the auth join pinned against the real schema, whole-chain visibility (feature link → settings child → assignment), end-to-end delivery through the real `buildEventLogConfigUpdate` resolver, foreign-MSP isolation, and sibling-org isolation *within* one MSP (guards against the branch ever being loosened to "same partner").

| Suite | Result |
|---|---|
| `agentContextPartnerWideFanout.integration` (new) | **13 passed** |
| Affected unit suites (9 files: agentAuth, agentWs, commands, heartbeat, reliability, token, elevationRequests, pamObservations, pamReconciliation) | **537 passed** |
| `rls-coverage` contract (`test:rls-coverage`) | **100 passed** |
| RLS contract (`vitest.config.rls.ts`) | **26 passed, 5 skipped** |
| Partner-wide regression (agentPolicyResolvers, configPolicyPartnerWideSelect, normalizedConfigPolicyRls, backupProfilesPartnerRls, partnerAxisSystemContext, configurationPolicyOrgScopedPartnerWide) | **72 passed** |
| Agent integration suites touched by the fixture change | **25 passed** |
| `tsc --noEmit` (whole `apps/api`) | **clean** |
| `eslint` on changed sources · `check-migration-naming` · `autoMigrate.test.ts` | **clean · OK · 67 passed** |

Red-first: the four new `agentAuth` unit tests were written against `currentPartnerId: null` / `partnerId: undefined` and observed failing before the implementation landed.

No migration in this wave.

## Rollback

Revert the PR — no schema involvement. Wave 1's policies are additive and stay correct with the field back at `null`.

Closes #4675

### DIFF
--- a/apps/api/src/__tests__/integration/agentContextPartnerWideFanout.integration.test.ts
+++ b/apps/api/src/__tests__/integration/agentContextPartnerWideFanout.integration.test.ts
@@ -1,0 +1,512 @@
+/**
+ * #4673 Wave 2 (#4675) — the agent DB context carries `currentPartnerId`, and
+ * that is what makes partner-wide config reachable on agent paths WITHOUT a
+ * system-context escape.
+ *
+ * Wave 1 shipped SELECT-only RLS branches across the configuration-policy
+ * chain:
+ *
+ *   USING (org_id IS NULL AND partner_id = public.breeze_current_partner_id())
+ *
+ * `breeze_current_partner_id()` reads the `breeze.current_partner_id` GUC,
+ * which `db/index.ts` SET LOCALs from `DbAccessContext.currentPartnerId`. Until
+ * Wave 2 the agent context hard-coded that field to `null`, so on agent paths
+ * the branch could never be true: a partner-wide policy read as ZERO ROWS with
+ * no error. That silent-zero is exactly why the agent-facing resolvers had to
+ * escape to a system context (`withPartnerWideVisibility`), which double-holds
+ * a pooled connection and bypasses RLS wholesale (#1105, #2417).
+ *
+ * WHY A SEPARATE FILE FROM agentPolicyResolversPartnerWide.integration.test.ts:
+ * that suite proves the RESOLVERS return partner-wide config, but it passes
+ * today with `currentPartnerId: null` because the resolvers still escalate
+ * internally. It therefore cannot detect whether Wave 2 works. This suite
+ * removes that confound by issuing PLAIN, UN-ESCALATED queries under the agent
+ * context and letting RLS alone decide — which is the precondition Wave 3
+ * depends on before it deletes the escapes.
+ *
+ * The load-bearing test in here is the NEGATIVE CONTROL
+ * ("the pre-Wave-2 agent context sees nothing"): it re-runs the identical query
+ * under an otherwise identical context whose only difference is
+ * `currentPartnerId: null`. Without it, every positive assertion below could be
+ * satisfied by some unrelated policy widening and nobody would notice.
+ *
+ * If this file were deleted: flipping `currentPartnerId` back to `null` in
+ * agentAuth.ts / heartbeat.ts / reliability.ts / agentWs.ts would compile,
+ * pass every unit test (they mock the DB and never evaluate RLS), and pass the
+ * resolver suite (which escalates) — while silently cutting every agent off
+ * from its MSP's shared configuration.
+ */
+import './setup';
+import { afterEach, describe, expect, it } from 'vitest';
+import { randomUUID } from 'crypto';
+import { eq, sql } from 'drizzle-orm';
+import { db, withDbAccessContext, type DbAccessContext } from '../../db';
+import {
+  configurationPolicies,
+  configPolicyFeatureLinks,
+  configPolicyAssignments,
+  configPolicyEventLogSettings,
+  devices,
+  organizations,
+} from '../../db/schema';
+import { buildEventLogConfigUpdate, EVENT_LOG_DEFAULTS } from '../../routes/agents/helpers';
+import { getRedis } from '../../services/redis';
+import { createPartner, createOrganization, createSite } from './db-utils';
+
+const SYSTEM_CTX: DbAccessContext = {
+  scope: 'system',
+  orgId: null,
+  accessibleOrgIds: null,
+  accessiblePartnerIds: null,
+  userId: null,
+};
+
+/**
+ * The POST-Wave-2 agent context, byte-for-byte the shape agentAuth.ts now
+ * builds (and that heartbeat.ts / reliability.ts / agentWs.ts hand-build from
+ * `agent.partnerId`).
+ *
+ * Note `accessiblePartnerIds: []`. That array gates `breeze_has_partner_access`
+ * — the partner-AXIS predicate that admits WRITES. Leaving it empty while
+ * setting `currentPartnerId` is the whole point: read visibility widens, write
+ * capability does not. If a future edit fills this array, the write assertions
+ * further down are what should start failing.
+ */
+function agentContext(orgId: string, partnerId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: [],
+    currentPartnerId: partnerId,
+    userId: null,
+  };
+}
+
+/** The PRE-Wave-2 agent context — identical except for the one field. */
+function legacyAgentContext(orgId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: [],
+    currentPartnerId: null,
+    userId: null,
+  };
+}
+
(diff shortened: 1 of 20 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + silent-failure-hunter + pr-test-analyzer (3 parallel Sonnet reviewers).

**Findings:** 3 raised → all 3 addressed in `d44851dc0` (+ `8d78dda10`); 0 outstanding.

1. **Missing regression coverage for this wave's own deliverable** (raised independently by *both* the test-analyzer and the silent-failure-hunter — the strongest signal in the run). The new integration suite proves the RLS **policy** given a context, but it *hand-rolls* that context: it never invokes `heartbeat.ts`, `reliability.ts` or `agentWs.ts`, so it could not prove those three actually **build** it. Confirmed by mutation: reverting `currentPartnerId` to `null` in any of the three left every suite green — the exact silent-zero bug this wave fixes, just scoped to fewer routes. Fixed by adding one assertion per site, reusing existing scaffolding. Each verified to **discriminate**:

   | Mutation | Result |
   |---|---|
   | `heartbeat.ts` → `null` | 1 failed / 176 passed |
   | `reliability.ts` → `null` | 1 failed / 10 passed |
   | `agentWs.ts` → `null` | 1 failed / 119 passed |
   | `agentWs.ts` `orgId`/`partnerId` **transposed** | 1 failed / 119 passed |

   The transposition case is why the agentWs assertion checks the **value**, not just presence: `runWithAgentOrgDbAccess(label, orgId, partnerId, fn)` takes two adjacent `string` params across five call sites, so swapping them typechecks cleanly — verified, `tsc --noEmit` reported **0 errors** on the transposed tree while the test caught it.

2. **`heartbeat.ts` comment overstated impact** — it claimed partner-wide policies "would silently vanish" without this field. Not true yet: every config read on that route still runs in its own `withSystemDbAccessContext`, which bypasses RLS. The field is **inert on this route until Wave 3** removes those escapes. Comment now says so, so nobody over-reads it.

3. **`agents/helpers.ts` stale justification** — its cis-baseline system escape was justified by "`breeze_current_partner_id()` is NULL", which this wave makes false. The escape stays (Wave 3 owns removing it) but no longer misstates why.

**Separately caught while verifying** (not from the reviewers): my `innerJoin` broke `wingetBootstrap.test.ts`'s `db.select` mock — 6 runtime failures that `tsc` could not see, because the mock is cast `as any`. Found by running the full `src/routes/agents` suite rather than only the files tsc flagged; fixed in `8d78dda10`.

**Explicitly cleared by review** (verified, not assumed): the INNER JOIN cannot drop an authenticable device (`devices.org_id` NOT NULL + FK, `organizations.partner_id` NOT NULL, both `NO ACTION`); `accessiblePartnerIds` stays `[]` at every touched site with no path merging the two axes; the empty-string GUC fails closed (`breeze_current_partner_id()` returns NULL on `''`); the excluded producers (`portal/auth.ts`, `clientAiAuth`, `commandQueue.ts` audit insert) leave no silent-zero gap — none touch a partner-wide-branched table; and the integration suite's BYPASSRLS guard, negative control, rowCount write assertions and `.cause.code` 42501 unwrap were each checked for vacuity and found sound.

**Tests:** `tsc --noEmit` clean (whole `apps/api`) · unit **1655 passed / 76 files** (`src/routes/agents` + `src/middleware`) · new fan-out proof **13 passed** · partner-wide regression suites **22 passed** · `rls-coverage` **100 passed** · RLS contract **26 passed / 5 skipped** · agent integration suites **25 passed** · eslint clean · `check-migration-naming` OK · `autoMigrate` **67 passed**.

Two pre-existing failures in `enrollmentCollision.integration.test.ts` (`403` vs `201` on `/agents/enroll`) were **verified pre-existing** — reproduced identically on the base commit `2296bb870` with all my changes reverted. Unrelated to this PR.

**Status:** review-clean, awaiting maintainer merge. No migration in this wave; rollback is a plain revert.

Issue #4547, opened 2026-09-02
Block hours: prepaid hour banks on contracts with drawdown from time entries and overage billing


ToddHebebrand:
## Description

MSPs sell blocks of prepaid support hours (10, 20, 40 hours per month or per term). Technician time draws the block down, the customer sees how much is left, and hours beyond the block bill at an overage rate. Unused hours either expire at period end or roll over, depending on the agreement.

Breeze cannot model this. Raised in an MSP demo alongside billing by device type (#3205); this is the second half of that conversation.

## Current state

There is no hour-allowance, bank, retainer, or overage concept anywhere in the schema or validators. A repo-wide grep for `hours|retainer|allowance|bank|block|prepaid|overage|rollover` across `apps/api/src/db/schema/` and `packages/shared/src/validators/` returns nothing relevant.

**Contracts** (`apps/api/src/db/schema/contracts.ts`): line types are `flat | per_device | per_seat | manual`. `generateDueInvoice` in `contractService.ts` resolves quantity per line by a switch on that enum; no case reads time entries. The billing sweep in `apps/api/src/jobs/contractWorker.ts` is date-driven only.

**Time entries** (`apps/api/src/db/schema/timeTracking.ts`): `time_entries` carries `is_billable`, `hourly_rate`, `currency_code`, `duration_minutes`, `ticket_id`, and a `billing_status` enum `not_billed | billed | no_charge | contract`. There is no `contract_id` on time entries or tickets. The `contract` billing status is defined but dead: every read filter in `invoiceAssembly.ts` excludes it and nothing ever writes it. `contractService.ts` says so in its own comment: contract-qualified time is not a queryable relation.

**Invoicing from time** (`apps/api/src/services/invoiceAssembly.ts`): `gatherOrgTimeEntries` pulls billable, `not_billed` entries in a date range; `timeEntryToLineSpec` turns each into an invoice line at `hours x hourly_rate`. The rate is stamped on the entry at creation from `org_ticket_settings.default_hourly_rate` or the ticket category default, never from a contract. Contract-sourced invoice lines (`sourceType: 'contract'`) and time-entry lines are separate code paths that never intersect.

**Catalog** (`apps/api/src/db/schema/catalog.ts`): item types are `hardware | software | service`; `unit_of_measure` is free text and nothing in billing reads it.

[...]

Issue #4191, opened 2026-08-28
W04: P2-4 Ticket triage (act) — autonomous metadata + private note, ticket_drafts, Send-as-me, ticket scope


ToddHebebrand:
Wave W04 of #4187 — P2-4 Ticket triage (act) — autonomous metadata + private note, ticket_drafts, Send-as-me, ticket scope
Branch: `feature/4187-ai-agents-p2/wave-4191`

ToddHebebrand:
**P2-4 executed — PRs open:** #4300 (API+shared, → main) and #4301 (web, stacked, Closes this issue). Merge order: #4300 first, then rebase/retarget #4301. Both CI-green except `Check Migrations`, which fails on every branch incl. main from the off-main `v0.109.0-rc.3` tag (pre-existing).

**Controller rulings made during execution** (full ledger + smoke screenshots in the worktree's `.superpowers/sdd/2026-08-30-ai-agents-p2-4-ticket-triage/`):
1. Tasks A4+A5 batched into one dispatch (schema↔dispatch parity tests would red between separate commits).
2. `'triage'` union compile-forward: placeholder arms allowed ahead of A6 (empty outcome tools, real profileCaps).
3. Org-merge: `ticket_drafts` = custom resolve-phase DELETE executor (drafts are ephemeral; plain repoint AND leave-for-erasure both 23503 — one leg always breaks).
4. `link_device` excludes `status='decommissioned'` (spec's `deleted_at` column doesn't exist on devices).
5. Private note is the triage deliverable → first in `maxActionsPerRun` cap order (spec §4.4 "exactly one per run").
6. RBAC for new actions: `tickets:read`/`write` (no `update` action exists in the grant catalog; guardrails use `tickets:update` deliberately AS unreachable-by-RBAC).
7. Autonomy decided inside the intent-creation transaction; `decidedVia: 'ticket_autonomy'`; fails closed (incl. on gate-evaluation exceptions) to human_required with a breadcrumb.
8. Resolution-note drafts consumed via the resolve flow only; **edited technician text always wins over draft content** (fixed on both resolve paths after final review).
9. Resolved-trigger lane skips the origin loop guard (every triage run posts an AI note; the guard would permanently dead the lane).
10. Ticket move-org tombstones ALL ticket-scoped intents + deletes drafts pre-UPDATE inside the move transaction (composite FKs are INITIALLY IMMEDIATE).
[...]

PR #4726, opened 2026-09-03
fix(tenancy): composite tenant-integrity FKs on time_entries / ticket_parts (#4596)

### LINKED ISSUE #4596: [API][RLS] time_entries / ticket_parts accept an org_id from another partner: partner-axis RLS never checks org, and the org FK is not composite
## Description

`time_entries` and `ticket_parts` are protected by a **partner-axis** RLS policy (`time_entries_partner_access`, `apps/api/migrations/2026-06-12-a-ticketing-time-parts.sql:60`) and carry an `org_id` whose FK points at `organizations(id)` alone (`apps/api/src/db/schema/timeTracking.ts:25` nullable on time_entries, `:61` NOT NULL on ticket_parts). Nothing at the database level ties that `org_id` to the row's `partner_id`, so a request authorised for partner A can write a row whose `org_id` belongs to partner B's organization. RLS passes (partner matches), the FK passes (the org exists), and the row then attributes billable labour to another tenant's customer.

Surfaced while drafting the block-hours spec (#4547), which will read these rows per org for drawdown. Verified against current `main`: policy is partner-only; no composite FK exists; the unique index the composite FK needs, `organizations_id_partner_uq`, already exists (`2026-04-11-users-rls.sql:66-74`, used by the `users` composite FK).

## Proposed fix

One migration, same shape as `users` (dual-axis precedent) and `contract_lines_site_org_fk` (#3205):

1. `SELECT set_config('breeze.scope','system',true);` (both tables are FORCE RLS; without it any cleanup is a silent 0-row no-op on managed Postgres).
2. Audit + clean mismatched rows with a logged count: `UPDATE ... SET org_id = NULL` (time_entries) / report and decide for ticket_parts (org_id NOT NULL there; likely `DELETE` is wrong, so raise a WARNING with the count and keep the FK `NOT VALID` until reviewed if any rows exist).
3. `ADD CONSTRAINT time_entries_org_partner_fk FOREIGN KEY (org_id, partner_id) REFERENCES organizations (id, partner_id)` and the same on `ticket_parts`; keep the existing single-column FKs.
4. Integration test forging a cross-partner `org_id` as `breeze_app` → `23503` with the new constraint name; export-policy registry unchanged (no new columns).

Also check the app layer: `timeEntryService` create/update paths should reject an org outside `auth.accessibleOrgIds` before the DB does.

## Affected files

- `apps/api/migrations/<new, sorts after the newest committed>`
- `apps/api/src/db/schema/timeTracking.ts` (comment only; FK is SQL-only)
- `apps/api/src/services/timeEntryService.ts`
- `apps/api/src/__tests__/integration/` (new forge test)

## Reported by

Found by the block-hours spec pass on #4547 (2026-09-02); verified by the pipeline session.

ToddHebebrand:
Closes #4596

## What was open

`time_entries` is RLS **Shape 3 (partner-axis)**: `time_entries_partner_access` checks `breeze_has_partner_access(partner_id)` and says **nothing about `org_id`**, whose FK pointed at `organizations(id)` alone. A caller authorised for partner A could write a row attributing billable labour to **partner B's customer** — RLS passed (the partner matched), the FK passed (the org existed), and every org-keyed reader downstream (`invoiceAssembly.ts`, `orgCurrencyService.ts`, the #4547 block-hours drawdown) would have counted it against the victim's org.

**No known reachable exploit today.** All three INSERT sites into these tables live in `timeEntryService.ts` behind `resolveTicketLink` / `resolveAndLockOrgLink`, and both already reject a foreign-partner org. #4547 adds new org-keyed readers and plausibly new writers, and the database was the missing wall.

## Corrections to the issue as filed

**`ticket_parts` is org-axis (Shape 1), not partner-axis** — it has four `breeze_has_org_access(org_id)` policies and **no `partner_id` column at all**, so the issue's proposed `FOREIGN KEY (org_id, partner_id)` is not implementable there. Verified live: `SELECT ... FROM information_schema.columns WHERE table_name='ticket_parts'` returns no `partner_id`.

Its real gap is narrower and is closed here instead: nothing tied `ticket_parts.org_id` to its parent `tickets.org_id`, so a part could hang off **another tenant's ticket** while carrying the writer's own `org_id` — and `invoiceAssembly.ts:170-176` bills parts by `org_id`.

Both facts are pinned as a discriminating **42501/23503 pair** in `ticketChildOrgFks.integration.test.ts`:

| Forge | Result | Why |
|---|---|---|
| part carrying **another partner's** `org_id` | **42501** | org-axis `WITH CHECK` already refuses it — this is why the proposed composite is unnecessary here |
| part carrying **its own** `org_id` on **another partner's ticket** | **23503** | RLS never looks at the ticket; only the new composite FK closes this |

**The app-layer ask is already satisfied on `main`** — `resolveAndLockOrgLink`'s "outside `accessibleOrgIds`" and "another partner's org" denials are both already pinned in `timeEntryService.test.ts:1699`. No fourth redundant check was added.

## Changes

Two migrations (`SET CONSTRAINTS`-aware, idempotent, no inner `BEGIN`/`COMMIT`):

- **`2026-10-06-110000-time-entries-org-partner-fk.sql`** — `time_entries_org_partner_fk (org_id, partner_id) -> organizations (id, partner_id)`, against the pre-existing `organizations_id_partner_uq`.
- **`2026-10-06-110100-ticket-child-org-fks.sql`** — `time_entries_ticket_org_fk` and `ticket_parts_ticket_org_fk`, both `(ticket_id, org_id) -> tickets (id, org_id)` against the pre-existing `tickets_id_org_uq`. `ON DELETE SET NULL (ticket_id)` uses the PG15 **column-list** form so deleting a ticket does not also wipe the org attribution of billable labour.

All three are `DEFERRABLE INITIALLY IMMEDIATE`. The single-column FKs are **kept**: composite FKs are `MATCH SIMPLE`, so a row with `ticket_id` set and `org_id NULL` is not a referencing row at all and its `ON DELETE SET NULL` would never fire, leaving a dangling `ticket_id`. Postgres evaluates FKs conjunctively, so the survivor is redundant, never permissive. (Deliberate departure from the `portal_users` precedent, which dropped its superseded FK.)

**Both org-move paths now defer the two ticket-keyed constraints BY NAME** (`services/ticketService.ts` `moveTicketOrg`, `routes/devices/moveOrg.ts`). Both rewrite `tickets.org_id` before the children follow, so without this the tickets UPDATE 23503s the instant it completes — the existing move suites went red exactly that way and are the proof. Never `ALL`: `tickets_requester_contact_org_fk`, `ticket_drafts_ticket_org_fk` and `action_intents_scope_ticket_org_fk` stay immediate as fail-fast guards, and the load-bearing "no `SET CONSTRAINTS` is issued in this transaction" claims in the C1 (#4191) comment were updated rather than left stale.

## Cleanup row counts (all zero)

Each migration `set_config('breeze.scope','system',true)` first — both tables are FORCE RLS, so an unscoped UPDATE is a silent 0-row no-op on managed Postgres — and reports its count unconditionally:

```
WARNING:  cleaned 0 cross-partner time_entries.org_id row(s)
WARNING:  re-derived 0 time_entries.org_id row(s) that disagreed with their ticket
WARNING:  unlinked 0 time_entries row(s) whose ticket belongs to another partner (cross-partner link; org attribution kept)
WARNING:  re-derived 0 ticket_parts.org_id row(s) that disagreed with their ticket
```

Wave 1 **nulls** `org_id` (the partner disagreement means we cannot know which org is correct; nulling cannot violate `time_entries_currency_required_when_org_chk`, which is one-directional). Wave 2 splits two cases, and the split is load-bearing:

- **Same-partner disagreement → re-derive** `org_id` from the parent ticket, because the ticket is the authority for a ticket-linked row.
- **Cross-partner ticket link → sever `ticket_id`, keep the org.** A `time_entries` row can be internally consistent (its `org_id` belongs to its own `partner_id`, so W1 accepts it and W1's cleanup leaves it alone) while its *ticket* belongs to another partner's org — `time_entries` RLS checks only `partner_id`, and the old single-column ticket FK checked no tenancy at all. Blindly re-deriving such a row writes another partner's org onto it and **violates `time_entries_org_partner_fk`**, which is already live and IMMEDIATE by then, aborting the entire W2 file with a 23503 on exactly the drifted data these migrations exist to remediate. The link is the corrupt field, not the org, so W2 nulls `ticket_id` and keeps the (real, self-consistent) billable attribution. The two counts are reported separately because a non-zero cross-partner count is a tenant-isolation signal, not drift.

> **Pre-release action:** run the same three audit queries against both prod regions before the release that carries these migrations, so a non-zero count is investigated as a possible isolation incident rather than discovered in a Postgres log.

## Registration-list audit — nothing to add

No new table, no new column, no new parent table. Verified by grep, not judgement:

| List | Status |
|---|---|
| `CORE_ORG_CASCADE_DELETE_ORDER` | both tables already present, alphabetised, `organizations` last |
| `CORE_TENANT_EXPORT_POLICY` | both classified; **no new columns**, so the one list that fires on `ADD COLUMN` is untouched |
| `CORE_DEVICE_CASCADE_DELETE_TABLES` / `..._ORG_DENORMALIZED_TABLES` | N/A — neither table has `device_id`; both are in `CUSTOM_ORG_REWRITE_TABLES` instead |
| `AUDIT_ADMIN_REQUIRED_TABLES` | N/A — neither is append-only |
| `orgMergeRegistry` REPOINT_TABLES | both present; merge works only because the new FKs are `DEFERRABLE` |
| `rls-coverage` | `time_entries` in `PARTNER_TENANT_TABLES`; `ticket_parts` auto-discovered Shape 1 |

**FK direction:** no new parent for either child, so `topologicalCascadeOrder()` is unchanged. **Partner-wide-first (#2135):** not applicable — these are transactional billing records, not config/policy tables.

## Verification

Real Postgres 16.15 via `pnpm test-stack up`, `breeze_app` role, RLS enforced.

**Red-first, with the constraints removed from the DB:** both `time_entries` forges and all three ticket forges **resolved** (the live vulnerability reproduced), and the `pg_constraint` reads returned 0 rows. The two 42501 controls passed throughout, so the suites discriminate between "RLS refused it" and "the FK refused it" rather than just failing.

| Suite | Result |
|---|---|
| `timeEntriesOrgPartnerFk` + `ticketChildOrgFks` + `timeEntriesTicketOrgFkCleanup` (new) | 25 passed |
| `ticket-move-org` + `time-entries-rls` + `deviceMoveOrgCurrency` | 22 passed (11 red before the deferral) |
| `orgLifecycleFoundations`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, `orgMerge`, `orgMergeRegistry` | 37 passed |
| `rls-coverage` (own runner) | 100 passed |
| `integration-suite-coverage` (own runner) | 5 passed |
| `moveOrg.test.ts`, `ticketService.test.ts`, `timeEntryService.test.ts`, `autoMigrate.test.ts` | 415 passed |
| `tsc --noEmit`, `eslint`, `db:check-drift` | clean / no drift |

**Mutation-tested, not assumed.** Removing the `SET CONSTRAINTS` from `moveTicketOrg` reddens 11 move tests; changing it to `SET CONSTRAINTS ALL` reddens the new unit test. Both restored.

Re-verified end to end on a **from-scratch** database (`test-stack down` + `up`, all 579 migrations replayed from zero). Migrations replay as no-ops. Renamed off `2026-10-06-100100-` mid-flight after `origin/main` landed a colliding migration; `check-migration-naming.sh --against-ref origin/main` passes and every reference was swept.

Plan: `docs/superpowers/plans/tenancy-rls/2026-09-02-time-entries-ticket-parts-tenant-fks.md`


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (Sonnet, in parallel, against `origin/main...HEAD`).

**Findings:** 3 real → all fixed in `475b579`; 0 outstanding.

1. **`code-reviewer` (critical) — the W2 cleanup could abort the whole migration.** A `time_entries` row can be internally consistent (`org_id` belongs to its own `partner_id`, so W1 accepts it and W1's cleanup leaves it alone) while its *ticket* belongs to another partner's org — `time_entries` RLS checks only `partner_id`, and the old single-column ticket FK checked no tenancy at all. The blind `SET org_id = t.org_id` re-derive then writes another partner's org onto that row and violates `time_entries_org_partner_fk`, which is already live and IMMEDIATE (110000 committed in its own transaction), aborting the entire 110100 file with a 23503 — **on exactly the drifted data these migrations exist to remediate.**

   Confirmed by direct repro against Postgres 16 before fixing:
   ```
   ERROR: insert or update on table "time_entries" violates foreign key constraint "time_entries_org_partner_fk"
   DETAIL: Key (org_id, partner_id)=(bbbb…b1, aaaa…01) is not present in table "organizations".
   CONTEXT: SQL statement "UPDATE time_entries te SET org_id = t.org_id FROM tickets t …"
   ```
   W2 now splits the cases: a same-partner disagreement re-derives from the ticket; a **cross-partner ticket link severs `ticket_id` and keeps the org attribution** — the link is the corrupt field, not the org, and the labour is real. The two counts are reported separately, because a non-zero cross-partner count is a tenant-isolation signal rather than drift.

2. **`silent-failure-hunter` (CI-blocking) — `routes/devices/moveOrg.test.ts` was left red.** Its positional `statements.slice(0, 4)` assertion shifted by one because of the new leading `SET CONSTRAINTS`. I had run `ticketService.test.ts` but not this sibling — a real miss. Fixed, and the device-move path now carries its own assertion that the statement is issued **BY NAME** and is never `SET CONSTRAINTS ALL`, which it previously had none of.

3. **`pr-test-analyzer` — the migration cleanups had never run against a drifted row.** Every CI run applies them to zero rows, so a reversed `IS DISTINCT FROM` or a missing partner guard would report a truthful `0` and pass everything in the PR. New suite `timeEntriesTicketOrgFkCleanup.integration.test.ts` forges the drift and replays each migration file verbatim — the replay's own `ADD CONSTRAINT` is the assertion surface. Following the `portalUserContactCompositeFk` precedent the reviewer pointed at.

   Also from this reviewer: every `23503` forge assertion now pins `constraint_name` (a future FK on these tables could otherwise satisfy them for the wrong reason), UPDATE forges added for both ticket-keyed FKs, and `ticketService.test.ts` now pins the total `tx.execute()` count so a duplicated or interposed statement cannot hide behind the `executedTableNames()` filter.

**Not accepted:** the suggestion to add a device-path integration test for `moveOrg × time_entries` — that combination is already covered by `time-entries-rls.integration.test.ts`'s "rewrites tickets, time_entries, and ticket_parts org_id atomically", which drives the `moveOrg` route and went red with `time_entries_ticket_org_fk` before the deferral landed.

**Mutation-tested, not assumed.** Each fix was verified to fail without it: reverting the W2 partner guard turns the new cleanup suite's third case red with the exact 23503 above; removing the `SET CONSTRAINTS` reddens 11 move tests; changing it to `SET CONSTRAINTS ALL` reddens both new unit tests.

**Tests:** re-verified end to end on a **from-scratch** database (`test-stack down` + `up`, all 579 migrations replayed from zero):

| Suite | Result |
|---|---|
| new forge + cleanup suites, `ticket-move-org`, `time-entries-rls`, `deviceMoveOrgCurrency` | 47 passed |
| `orgLifecycleFoundations`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, `orgMerge`, `orgMergeRegistry` | 37 passed |
| `rls-coverage` / `integration-suite-coverage` (own runners) | 100 / 5 passed |
| `moveOrg`, `ticketService`, `timeEntryService`, `autoMigrate` unit | 415 passed |
| `tsc --noEmit`, `eslint`, `db:check-drift`, `check-migration-naming --against-ref origin/main` | clean |

**CI:** 68/68 green on `475b579`, 0 failures. Verified the three new suites actually **executed** in the integration shards (1 and 2) rather than being silently skipped.

> **Pre-release action for the merger:** run the three audit queries in the PR body against both prod regions before the release that carries these migrations, so a non-zero count is investigated as a possible isolation incident rather than found in a Postgres log.

PR #4728, opened 2026-09-03
feat(agent): end-user reboot prompt with deferral, Windows + macOS (#3207 W3)

### LINKED ISSUE #4663: W03: Deferral prompt on Windows + macOS: correlated RequestNotificationDecision + helper native dialog
Wave W03 of #3207 — Deferral prompt on Windows + macOS: correlated RequestNotificationDecision + helper native dialog
Branch: `feature/3207-reboot-deferral/wave-4663`

ToddHebebrand:
Closes #4663. Wave W03 of #3207 — the deferral prompt on Windows + macOS.

## What ships

When policy enables deferral and a helper session is present, the reboot's early warning rungs become a **native modal dialog** with **Restart now** / **Postpone N**, and the answer drives `RebootManager.Defer()`. The feature becomes real; W1 sent the budget, W2 built the state machine, this wave puts a button in front of a person.

Four commits, each independently green:

1. **`fix(agent)`: bound the session broker's listener close.** Pre-existing shutdown deadlock — see below.
2. **`feat(agent)`: `Broker.RequestNotificationDecision`.** The response-bearing sibling of `BroadcastNotification`, which is untouched.
3. **`feat(agent)`: the helper-side prompt dialog.** `showNotifyPromptFn` seam plus windows/darwin/other implementations; `handleNotify` routes on `len(Actions) > 0`.
4. **`feat(agent)`: `promptFn` on the manager** + the heartbeat wiring, plus the `osInvoked` fix codex flagged in the W2 review.

## Two defects fixed inline

**`BroadcastNotification` sent an empty envelope id**, so a `NotifyResult` reply matched no pending command and `dispatchHelperMessage` dropped it as unsolicited. Every other piece of the contract already existed — `NotifyRequest.Actions`, `NotifyResult.ActionClicked`, `expectedResponseType`'s `TypeNotify -> TypeNotifyResult` mapping, the helper replying on `env.ID`. `RequestNotificationDecision` allocates a real per-session id; `BroadcastNotification` stays fire-and-forget, with a test pinning that it registers no pending response.

The id is a **monotonic counter**, not the `time.Now().UnixMilli()` convention the repo's other correlated requests use (`LaunchProcessViaUserHelperWithArgs`). `SendCommand` rejects a duplicate in-flight id outright, and a mutation run confirmed the millisecond form loses answers under 25 concurrent requests.

**The Windows CI hang** (`TestNamedPipeListenAndAccept`, closed issue #3018, recurred on PR #4686 at 2026-09-03 02:12Z, [job 100494963534](https://github.com/LanternOps/breeze/actions/runs/33704894539/job/100494963534)). Root cause established from the panic dump, not guessed:

> `goroutine 590 [chan receive, 9 minutes]` — `winio.(*win32PipeListener).Close` at `pipe.go:578`, i.e. `<-l.doneCh` **after a successful send** on `closeCh`.
> `goroutine 571 [select, 9 minutes]` — `listenerRoutine` at `pipe.go:462`, the **top-level** select.
> The broker's accept-loop goroutine is absent from the dump entirely.

Those three facts pin it. `Close()`'s single `closeCh` token was consumed by `makeConnectedServerPipe`'s abort branch (`pipe.go:447`), which rewrites only `nil` and `ErrFileClosed` into `ErrPipeListenerClosed`. A connect already failing on its own errno when `Close()` cancelled it — `ERROR_NO_DATA` from a client that dialled and hung up, which is exactly what this test does — leaks that errno through. `listenerRoutine` delivers it to the waiting `Accept()` (hence the accept goroutine exiting cleanly and vanishing from the dump), leaves `closed` false, and returns to its select at 462 to wait for a second token that will never be sent. `doneCh` is never closed, so `Close()` never returns.

**This is an upstream go-winio bug and v0.6.2 is the newest release**, so there is no upgrade to take. The fix is defensive: `StopAcceptingAndWait` closes the listener on its own goroutine and bounds the wait on the caller's context. Abandoning that close is safe where abandoning the pre-auth drain is not — `acceptStopped` is set under `acceptMu` before the close is even attempted, so nothing can be admitted whether or not the handle closes. Cost is one leaked goroutine and one handle for the life of the process; the alternative is agent shutdown never returning (on a device that turns a service restart into a kill).

The regression test drives it through `net.Listener` rather than a real pipe, so it runs in the ordinary **Test Agent** job on every platform. Verified red before the fix: it hangs for the full 10s guard on Linux, the same shape as the 10-minute CI hang.

**Third item, also fixed inline.** Codex flagged during the W2 review that `runOSReboot` set `osInvoked` *before* `execOSReboot` ran, so a `Cancel` in that window took the abort path, found nothing to abort, and reported success while the machine still went down. This wave widens it — "Restart now" re-enters `scheduleLockedAt`, which takes the same abort path — so it is fixed here rather than deferred. A separate `osInvoking` flag covers the interval: `Cancel`, `Defer` and a re-schedule are all **refused** there rather than handed a success that cannot be delivered, and `osInvoked` is set only once `exec` returns successfully.

## Contracts held

- **Cancel-safe.** The dialog blocks for up to two minutes, during which the operator can cancel from the console. `Defer()` already refused a cancelled reboot, but `ScheduleWithOptions` has no generation check — a "Restart now" answered after a cancellation would have armed a brand-new reboot the operator had just called off. `restartNow` checks the generation under the lock. A refused click tells the user why instead of looking like it worked.
- **Budget-safe.** `ComputeDeferral` is the single authority on whether a button appears, so the manager can never offer a postponement `Defer()` would then refuse. Budget spent, or a deadline with no room, means no button — and the warning says "This restart can no longer be postponed." in words.
- **Never on the closing rung.** The critical notice fires at `OSInvokeAt`, from which point the countdown is in the OS and `Defer()` is refused. Covered for the collapsed single-rung plan too.
- **Never blocks the manager.** `r.mu` is not held across `promptFn`; `State()` during an open dialog is tested.
- **The prompt never outlives the next rung**, property-tested across seven schedule lengths. A dialog still open then would make the helper refuse to stack a second modal, silently costing the ladder a warning.
- **Always-warn (#3197) is independent of the prompt.** A dialog that cannot render reports `shown=false` and the helper falls back to the plain toast. `promptFn == nil` (headless, no helper) still warns and still reboots on time.
- **Headless is the story, not a gap.** No helper means no prompt and a reboot on schedule. No switch added.
- **Backward compatible.** `Actions`/`TimeoutMs` are `omitempty`: an old helper ignores them and gets no correlated reply, which reads as no decision; a new helper on an old agent never receives actions and behaves exactly as today.
- **Scope filter unchanged (#3255).** The assist helper and watchdog are excluded from both paths — a modal prompt is strictly more intrusive than the toast that filter was written for.
- **Deferral off is byte-identical**, asserted: `promptFn` is never called and the copy never mentions postponement.

## Review round (all findings addressed or explicitly left)

Three reviewers: Sonnet `code-reviewer`, Sonnet `silent-failure-hunter`, and `codex gpt-5.6-sol` at medium on the concurrency. The first two findings were reached independently by more than one reviewer.

**Fixed in `6f210437a`:**

1. **The always-warn invariant was lost on any device with no signed-in helper.** `PromptFunc` returned only a label, so "a person saw the dialog and did nothing" and "nothing reached anybody" were the same empty string — and `emitNotification` treated both as "the prompt was the warning" and emitted nothing more. On a headless box, a Windows server at the logon screen, or a machine whose helper had crashed, **every deferrable rung warned nobody**; only the closing notice survived. That is #3197 reintroduced for exactly the machines the ladder was written for. My own headless test missed it because it used a `nil` `promptFn`, which production never has. `PromptFunc` now returns `(clicked, shown)`, the manager falls back to the ordinary notification when nothing rendered, and the tests use the production shape. The signal already existed and was being thrown away one hop up — the helper honestly reports `Delivered:false`.
2. **A "Postpone" answered after the schedule was *superseded* postponed the replacement and spent its budget.** `Defer()` asks only whether *some* reboot is scheduled, which stays true across a last-writer-wins re-dispatch. `restartNow` already checked the generation; `deferForGeneration` now does too. Mutation-verified: reverting to plain `Defer()` fails the new test on all three assertions.
3. **`execOSReboot` was an unbounded exec, and this wave's `osInvoking` flag made its return load-bearing** — a `shutdown` that never returned would leave Cancel, Defer and every re-schedule refused permanently. All six platform invocations now route through one untagged `runRebootCommand` with a 30s kill timeout, so the bound is covered by tests that run in CI.
4. **`Broker.Close` discarded `StopAcceptingAndWait`'s error**, so a stalled listener looked like a clean shutdown to any caller that had not called it directly.

The heartbeat's prompt closure became a named, tested `rebootPromptFunc` rather than an inline lambda — that hop is exactly where "the helper showed nothing" could be mistaken for "the user declined to answer", and the previous wiring test only pinned the type signature.

**Pre-existing, deliberately not fixed here** (both are W2/#3197 structure this wave does not touch — flagging rather than silently dropping):

- **`Cancel` runs `abortOSReboot` outside the lock.** It clears `osInvoked`, unlocks, then runs the abort. A concurrent `Schedule` in that gap sees a clean manager and arms a new reboot; the late `shutdown /a` then aborts the *new* countdown while `Cancel` has already reported success. If the abort instead fails, `osInvoked` stays false, so a retry is impossible and a live countdown is no longer represented. Fix: an `osAborting` state that refuses scheduling and cancellation until the abort resolves, clearing `osInvoked` only on success.
- **`Stop` can return while a committed reboot still fires.** `runOSReboot` sets `osInvoking`, unlocks, and does not re-check `stopped` before invoking; `Stop` meanwhile returns claiming to have cancelled pending work. `Stop` also never aborts an already-live countdown. Whether it *should* is a product question — the agent stopping is often *because* of the reboot — which is why this is flagged rather than guessed at.

## Deviations from the plan

1. **An osascript `-128` cancel maps to no decision, not to the last button.** The plan said map it to the last button — but that is "Restart now", so Escape would have *accelerated* the user's reboot. Leaving the countdown where it is matches the manager's own rule that silence never postpones and never accelerates.
2. **The seam returns `(clicked string, shown bool)`**, not just `clicked`. `Delivered: true` on a dialog that never rendered is a lie that inflates delivery accounting, and `shown=false` is what triggers the toast fallback that keeps #3197 true.
3. **Prompts are serialised per helper process.** Both vehicles are `MB_SYSTEMMODAL|MB_TOPMOST` / a standard macOS dialog; a second prompt would stack over the first and steal focus. One that cannot take the lock falls back to a toast rather than queueing behind a dialog with two minutes left to run.
4. **The prompt window is computed per rung**, not a flat `min(next rung, 2m)` at call time — it lives in `planRebootRungs` as pure data so it is testable, and a rung with less than 20s of room offers no button at all.
5. **The manager's copy states the remaining postponements.** The plan didn't ask; a user who cannot see their remaining budget cannot plan around it, and the exhausted case has to say so once the button disappears.
6. **`ComputeDeferral` takes five arguments, not four** (the plan's W3 sketch predates W2's `scheduledAt` parameter). Followed the code.

### Observation, not changed here

`ComputeDeferral` grants a postponement when `deadline == scheduledAt`: the clamp lands the "new" time exactly where the old one was, so a deferral is spent that moves nothing. That is W2 arithmetic and out of scope for this wave — flagging it rather than editing it.

## Test evidence

**Verified** (`golang:1.26.6` container; this dev host cannot run cgo — `go-m1cpu` segfaults on init):

| Command | Result |
|---|---|
| `go test -race ./internal/sessionbroker/` | ok, 14.3s |
| `go test -race ./internal/patching/` | ok — **132 tests pass**, 1 pre-existing skip |
| `go test -race ./internal/userhelper/` | ok, 2.0s |
| `go test -race ./internal/heartbeat/` | ok, 30.7s |
| `go test -race ./internal/ipc/` | ok, 2.6s |
| `CGO_ENABLED=0 go test ./...` (matches the **Test Agent** job) | 2 failures, both environmental — see below |
| `go vet` × `GOOS=windows,darwin,linux` | clean (this compiles the tagged files, incl. the new `notify_prompt_windows.go` / `_darwin.go`) |
| `CGO_ENABLED=0 go build ./...` × three GOOS | rc=0 each |
| `golangci-lint v2.12.2` on the five touched package trees | **0 issues in any file this PR touches** (61 pre-existing elsewhere, which CI's `--new-from-rev` mode skips) |
| `gofmt -l` on all 17 changed Go files | clean |

The two `./...` failures are `internal/backup/systemstate` `TestCollectHardwareOnly` (no CPU model/cores inside a container) and `internal/executor` `TestConfigureRunAsElevatedWhenNotRoot` (the container runs as root). Both are in packages this PR does not touch, and the executor one **passes when the same container runs as uid 1000** — confirmed, not assumed.

Red-first was observed for every new behaviour: the shutdown tests hang for their full guard without the fix, and both the empty-id and `UnixMilli`-id mutations were run and caught.

**Not verified — needs a real box.** Windows-tagged and darwin-tagged rendering code is executed nowhere in CI (#3019, #3046), so this is a claim I am explicitly not making:

- The `MessageBoxTimeoutW` dialog actually rendering under the user-helper session on Windows, its Yes/No buttons mapping to Restart now / Postpone, and the 32000 timeout return firing.
- The `osascript display dialog` rendering from the macOS desktop helper, its button ordering, and `gave up:true` / `-128` behaviour.
- End to end: a patch-triggered reboot with deferral on shows the dialog, Postpone moves the countdown, and `get_reboot_status` reports `deferralsUsed: 1`.

Everything that *decides* what those dialogs render and what their answers mean is untagged and covered — the timeout bound, the button-index mapping, the Yes/No legend, the AppleScript button ordering and its result parsing.

## Not in this PR

W4 (Linux parity) and W5 (console visibility) — deliberately out of scope.

### DIFF
--- a/agent/internal/heartbeat/heartbeat.go
+++ b/agent/internal/heartbeat/heartbeat.go
@@ -1011,26 +1011,12 @@ func NewWithVersion(cfg *config.Config, version string, token *secmem.SecureStri
 				h.sessionBroker.BroadcastNotification(title, body, urgency)
 			}
 		},
-		func(title, body, urgency string, actions []string, timeout time.Duration) string {
+		rebootPromptFunc(func(req ipc.NotifyRequest, timeout time.Duration) (ipc.NotifyResult, error) {
 			if h.sessionBroker == nil {
-				return ""
+				return ipc.NotifyResult{}, nil
 			}
-			res, err := h.sessionBroker.RequestNotificationDecision(ipc.NotifyRequest{
-				Title:     title,
-				Body:      body,
-				Urgency:   urgency,
-				Actions:   actions,
-				TimeoutMs: int(timeout.Milliseconds()),
-			}, timeout)
-			if err != nil {
-				// Not a failure: no helper, an old helper that ignores Actions, or
-				// a user who simply did not answer. All of them mean "no decision",
-				// and the reboot proceeds exactly as scheduled.
-				log.Debug("reboot prompt returned no decision; proceeding as scheduled", "error", err.Error())
-				return ""
-			}
-			return res.ActionClicked
-		},
+			return h.sessionBroker.RequestNotificationDecision(req, timeout)
+		}),
 		cfg.PatchRebootMaxPerDay,
 	)
 
--- a/agent/internal/heartbeat/heartbeat.go
+++ b/agent/internal/heartbeat/heartbeat.go
@@ -1003,12 +1003,36 @@ func NewWithVersion(cfg *config.Config, version string, token *secmem.SecureStri
 	// Register winget provider (SYSTEM/machine-scope; see winget_register_windows.go)
 	h.registerSystemWinget()
 
-	// Initialize reboot manager (uses session broker for user notifications)
-	h.rebootMgr = patching.NewRebootManager(func(title, body, urgency string) {
-		if h.sessionBroker != nil {
-			h.sessionBroker.BroadcastNotification(title, body, urgency)
-		}
-	}, cfg.PatchRebootMaxPerDay)
+	// Initialize reboot manager (uses session broker for user notifications, and
+	// for the interactive postponement prompt when policy enables deferral).
+	h.rebootMgr = patching.NewRebootManagerWithPrompt(
+		func(title, body, urgency string) {
+			if h.sessionBroker != nil {
+				h.sessionBroker.BroadcastNotification(title, body, urgency)
+			}
+		},
+		func(title, body, urgency string, actions []string, timeout time.Duration) string {
+			if h.sessionBroker == nil {
+				return ""
+			}
+			res, err := h.sessionBroker.RequestNotificationDecision(ipc.NotifyRequest{
+				Title:     title,
+				Body:      body,
+				Urgency:   urgency,
+				Actions:   actions,
+				TimeoutMs: int(timeout.Milliseconds()),
+			}, timeout)
+			if err != nil {
+				// Not a failure: no helper, an old helper that ignores Actions, or
+				// a user who simply did not answer. All of them mean "no decision",
+				// and the reboot proceeds exactly as scheduled.
+				log.Debug("reboot prompt returned no decision; proceeding as scheduled", "error", err.Error())
+				return ""
+			}
+			return res.ActionClicked
+		},
+		cfg.PatchRebootMaxPerDay,
+	)
 
 	// Set backup binary path for IPC forwarding to breeze-backup helper
 	h.backupBinaryPath = cfg.BackupBinaryPath
--- a/agent/internal/heartbeat/reboot_prompt.go
+++ b/agent/internal/heartbeat/reboot_prompt.go
@@ -1,0 +1,61 @@
+package heartbeat
+
+import (
+	"errors"
+	"time"
+
+	"github.com/breeze-rmm/agent/internal/ipc"
+	"github.com/breeze-rmm/agent/internal/patching"
+	"github.com/breeze-rmm/agent/internal/sessionbroker"
+)
+
+// notifyDecisionRequester is the one broker call the reboot prompt needs, taken
+// as a function rather than an interface so the nil-broker case stays a plain nil
+// check at the call site — a nil *sessionbroker.Broker stored in an interface is
+// not a nil interface, and that distinction has no business in this logic.
+type notifyDecisionRequester func(ipc.NotifyRequest, time.Duration) (ipc.NotifyResult, error)
+
+// rebootPromptFunc adapts the broker's correlated notification request to the
+// reboot manager's prompt seam.
+//
+// It exists as a named function rather than a closure inside the manager's
+// construction so this translation is directly testable: it is the exact hop at
+// which "the helper told us it showed nothing" could be mistaken for "the user
+// looked at a dialog and chose not to click", and those two demand opposite
+// behaviour from the manager.
+//
+//   - shown=false means nothing reached a person — no helper session, a dead
+//     transport, or a helper that could put neither a dialog nor a toast on
+//     screen. The manager falls back to the ordinary notification, which is what
+//     keeps the #3197 always-warn invariant true on a headless box.
+//   - shown=true with an empty label means a real person saw the prompt and did
+//     nothing. The reboot proceeds exactly as scheduled and no second warning is
+//     emitted, because the dialog already was the warning.
+func rebootPromptFunc(request notifyDecisionRequester) patching.PromptFunc {
+	return func(title, body, urgency string, actions []string, timeout time.Duration) (string, bool) {
+		res, err := request(ipc.NotifyRequest{
+			Title:     title,
+			Body:      body,
+			Urgency:   urgency,
+			Actions:   actions,
+			TimeoutMs: int(timeout.Milliseconds()),
+		}, timeout)
+		if err != nil {
+			if errors.Is(err, sessionbroker.ErrCommandTimeout) {
+				// The helper had the prompt and the user did not answer inside the
+				// window. Ordinary, and not worth a warning on every rung — but it
+				// is still "nothing was confirmed shown", so the manager falls back
+				// rather than assuming a dialog the helper never acknowledged.
+				log.Debug("reboot prompt went unanswered; proceeding as scheduled")
+				return "", false
+			}
+			// A transport or helper error. This one IS worth seeing: it means the
+			// interactive path is broken on this device, which is invisible from
+			// the console.
+			log.Warn("reboot prompt could not be delivered; falling back to a plain notification",
+				"error", err.Error())
+			return "", false
+		}
+		return res.ActionClicked, res.Delivered
+	}
+}
--- a/agent/internal/heartbeat/reboot_prompt_wiring_test.go
+++ b/agent/internal/heartbeat/reboot_prompt_wiring_test.go
@@ -1,0 +1,71 @@
+package heartbeat
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/breeze-rmm/agent/internal/ipc"
+	"github.com/breeze-rmm/agent/internal/patching"
+)
+
+// TestRebootManagerIsBuiltWithThePromptSeam is a source assertion, in the idiom
+// of TestBrokerLifecycleCleanupNeverReopensRecordedPID (sessionbroker).
+//
+// The construction it guards sits inside the agent's start-up path, which cannot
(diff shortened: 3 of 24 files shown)


ToddHebebrand:
**Review run:** Sonnet `pr-review-toolkit:code-reviewer` + Sonnet `pr-review-toolkit:silent-failure-hunter` + `codex exec` (gpt-5.6-sol, medium, read-only) scoped to the concurrency. Three reviewers, two of them converging independently on the first two findings.

**Findings:** 6 raised → **4 fixed in `6f210437a`**, 2 pre-existing and explicitly left (documented in the PR body with the concrete fix).

Fixed:
1. **The always-warn invariant (#3197) was lost on any device with no signed-in helper.** `PromptFunc` returned only a label, so "a person saw the dialog and did nothing" and "nothing reached anybody" were the same empty string — and every deferrable rung then warned *nobody* on a headless box, a server at the logon screen, or a machine whose helper had crashed. My own headless test missed it: it used a `nil` `promptFn`, which production never has. `PromptFunc` now returns `(clicked, shown)` and the manager falls back to the ordinary notification.
2. **A "Postpone" answered after the schedule was superseded postponed the replacement and spent its budget.** `restartNow` already checked the generation; `deferForGeneration` now does too. Mutation-verified.
3. **`execOSReboot` was unbounded**, and this wave's `osInvoking` flag made its return load-bearing — a hung `shutdown` would have refused Cancel/Defer/re-schedule permanently. Now one untagged `runRebootCommand` with a 30s kill timeout.
4. **`Broker.Close` discarded `StopAcceptingAndWait`'s error**, so a stalled listener read as a clean shutdown.

Left (pre-existing W2/#3197 structure, not touched by this wave): `Cancel` runs `abortOSReboot` outside the lock, so a late abort can kill a newer countdown; and `Stop` can return while a committed reboot still fires. Both are written up in the PR body — the second is partly a product question, so I did not guess at it.

**Tests:** `go test -race` green on all four touched packages — `patching` **132 pass** / 1 pre-existing skip, `sessionbroker`, `userhelper`, `heartbeat`, `ipc`. `go vet` and `go build` clean across `GOOS=windows|darwin|linux`. `golangci-lint v2.12.2`: **0 issues in any file this PR touches**. `gofmt` clean on all 18 changed files. The two failures under `CGO_ENABLED=0 go test ./...` are environmental (container has no CPU model; container runs as root) in packages this PR does not touch — the executor one passes when the same container runs as uid 1000.

**Not verified — needs a real box.** Windows- and darwin-tagged rendering runs nowhere in CI (#3019, #3046): the `MessageBoxTimeoutW` dialog under the user-helper session, the `osascript` dialog from the macOS desktop helper, and the end-to-end patch reboot showing `deferralsUsed: 1`. Everything that *decides* what those dialogs render and what their answers mean is untagged and covered.

PR #4730, opened 2026-09-03
fix(web): force TLS 1.2 in the generated Windows install one-liner

### LINKED ISSUE #4586: Windows `service install` not idempotent; PowerShell installer lacks explicit TLS 1.2
Requested by **andrepereira5963**. In #bugs (Discord, message 1544642338293555281).

**Discussion (PII-redacted):**

**andrepereira5963** (2026-09-02T09:37:02.068Z):
> Another one I've found.
> 
> I found another issue while testing the new signed agent on an older machine (Windows Server 2016).
> The PowerShell install command initially failed at the download stage with:
> Invoke-WebRequest : The request was aborted: Could not create SSL/TLS secure channel.
> 
> I checked:
> 
> [Net.ServicePointManager]::SecurityProtocol
> 
> and on this server it returned:
> 
> Ssl3, Tls
> 
> I then forced TLS 1.2:
> 
> [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
> 
> After doing that, Invoke-WebRequest successfully downloaded the agent from the new Breeze API URL, so adding TLS 1.2 explicitly to the generated PowerShell command may be needed for compatibility with older Windows/PowerShell/.NET environments.
> 
> After fixing TLS, the command progressed further but failed at:
> 
> breeze-agent.exe service install
> 
> with exit code 1.

**andrepereira5963** (2026-09-02T09:37:09.935Z):
> Running that command separately showed the actual problem:
> 
> failed to install service binary in protected Program Files location: replace existing protected executable: remove C:\Program Files\Breeze\breeze-agent.exe: Access is denied.
> 
> This machine already had BreezeAgent and BreezeWatchdog installed from a previous installation. I stopped both services/processes and was then able to replace the existing breeze-agent.exe.
> 
> Running service install again successfully copied the new binary into:
> 
> C:\Program Files\Breeze\breeze-agent.exe
> 
> but then failed with:
> 
> failed to create service: The specified service already exists.
> 
> The existing service was already correctly pointing to:
> 
> "C:\Program Files\Breeze\breeze-agent.exe" run
> 
> I also verified the replacement executable with Get-AuthenticodeSignature, and it showed Status: Valid with my Mear Technology Ltd signing certificate, so the correct BYO-signed agent was now installed.
> 
> Because the one-liner stops when service install returns exit code 1, it never reaches the enrollment stage. I therefore ran the enrollment manually.
> 
> The first valid enrollment was accepted by the Breeze server and an Agent ID was issued, but saving the local configuration failed with:
> 
> rename C:\ProgramData\Breeze\agent.yaml.partial C:\ProgramData\Breeze\agent.yaml: The process cannot access the file because it is being used by another process.

**andrepereira5963** (2026-09-02T09:37:11.577Z):
> After making sure the existing Breeze services/processes were stopped and retrying the enrollment with fresh enrollment credentials, it completed successfully. I started the services again and the device then appeared correctly in the Breeze dashboard.
> 
> So the full sequence on this Windows Server 2016 machine was:
> 
> TLS negotiation failure → force TLS 1.2 → existing agent executable locked → stop existing Breeze services → signed agent successfully replaces old binary → existing Windows service causes service install to return exit 1 → manual enrollment → existing agent.yaml locked → stop Breeze processes → retry enrollment → successful enrollment and device appears in dashboard.
> 
> There seem to be two improvements worth considering for the generated Windows installer:
> 
> Explicitly enable TLS 1.2 before Invoke-WebRequest for older PowerShell/.NET environments.
> Handle existing Breeze installations more gracefully/idempotently — stopping the existing services as required, updating the binary, not failing just because the service already exists, and avoiding the agent.yaml lock during re-enrollment.
> 
> I can confirm the signed agent itself works correctly once these issues are worked around.

## Triage analysis

Two distinct, verifiable code gaps behind the reported sequence on Windows Server 2016:

1) Windows `service install` is not idempotent, unlike Linux. In agent/internal/agentapp/service_cmd_windows.go, `serviceInstallCmd` calls `serviceinstall.InstallProtectedBinary` (which just tries to remove+rename the exe in place, agent/internal/serviceinstall/paths.go `copyExecutable`) without first stopping any existing BreezeAgent service, then unconditionally calls `mgr.CreateService(...)`, which fails with 'service already exists' if the service is already registered — there's no `OpenService`/update path. Compare this to agent/internal/agentapp/service_cmd_linux.go, whose `serviceInstallCmd` explicitly stops the existing systemd service before replacing the binary and uses an idempotent `WriteFile` for the unit (never fails just because it already exists). Docs (chunk 8, Advanced Install Methods) state 'service install is safe to re-run — it overwrites the binary at the install path, restarts the agent service' with no Windows caveat, so the Windows behavior is a real regression/asymmetry against both the docs and the other platforms' implementation.

2) The generated Windows one-liner in apps/web/src/lib/installCommands.ts never sets `[Net.ServicePointManager]::SecurityProtocol`, so on hosts whose default protocol list excludes TLS 1.2 (as seen on this Windows Server 2016 box), `Invoke-WebRequest` fails outright before the agent is even downloaded.

The third symptom (agent.yaml.partial rename 'being used by another process') is not a bug: agent/internal/config/config.go's `atomicWriteFile` (lines 877-921) documents this exact Windows tradeoff — MoveFileEx/os.Rename can hit ERROR_ACCESS_DENIED if another process (the still-running old agent) holds the destination open, and the comment explicitly says failing loudly here was chosen over the old in-place O_TRUNC write to avoid silent corruption (issue #642). It only surfaces because of bug (1) leaving an old agent process alive.

## Affected code

- `agent/internal/agentapp/service_cmd_windows.go:38-68` — serviceInstallCmd copies the binary and calls mgr.CreateService unconditionally; no check/stop for an already-registered BreezeAgent service, so re-running on a machine with an existing install fails with 'service already exists'.
- `agent/internal/serviceinstall/paths.go:73-115` — copyExecutable only removes+renames the target exe; if the running service holds it open, this is the 'Access is denied' the reporter hit — nothing here stops the existing service first.
- `agent/internal/agentapp/service_cmd_linux.go:127-155` — Reference implementation: stops the existing service before replacing the binary, and unit install is an idempotent overwrite — shows Windows path is inconsistent with the documented 'safe to re-run' behavior.
- `apps/web/src/lib/installCommands.ts:57-63` — Generated Windows one-liner has no explicit TLS 1.2 (SecurityProtocol) setting before Invoke-WebRequest.
- `agent/internal/config/config.go:877-921` — Documents the Windows ERROR_ACCESS_DENIED rename tradeoff for agent.yaml.partial — intended behavior (issue #642), not a new defect.

## Potential fix

On Windows, before creating the service: try mgr.OpenService(windowsServiceName); if it exists, stop it (like the Linux path does) and either update its binary path/config via ChangeConfig or delete+recreate it, instead of only calling CreateService. Also stop any existing BreezeAgent/Watchdog process before copyExecutable so the binary replace doesn't hit 'Access is denied'. Separately, prepend `[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12;` to the generated Windows command in installCommands.ts.

ToddHebebrand:
## Summary

Refs #4586 (TLS half only — see below).

The PowerShell one-liner generated by `buildInstallCommands` (`apps/web/src/lib/installCommands.ts`) runs `Invoke-WebRequest` without first setting `[Net.ServicePointManager]::SecurityProtocol = ... Tls12`. On older Windows PowerShell 5.1 hosts with legacy defaults (reported: Windows Server 2016, default `Ssl3, Tls`), this makes the download fail outright with:

```
Invoke-WebRequest : The request was aborted: Could not create SSL/TLS secure channel.
```

- Adds a TLS 1.2 line before the download step, **bitwise-OR'd into the existing `SecurityProtocol` value** (`-bor`) rather than replacing it, so `Tls13` stays enabled on hosts where it's already present.
- Swept the repo for other generated PowerShell install/bootstrap one-liners sharing the gap (`grep -i "Invoke-WebRequest|iwr"` under `apps/web/src`, `apps/api/src`, `apps/docs`, `scripts/`):
  - `apps/api/src/services/systemScriptLibrary.ts` (edition-migration script) — already sets `Tls12`, no gap.
  - `apps/docs/src/content/docs/migration/toolkit.mdx` (Recipe 3 push payload) — already sets `Tls12`, no gap.
  - `apps/docs/src/content/docs/agents/installation.mdx` (Windows tab) — **had the gap, fixed**.
  - `apps/docs/src/content/docs/agents/advanced-install.mdx` (Windows tab) — **had the gap, fixed**.
  - Nothing under `scripts/` matched.
- Extended `apps/web/src/lib/installCommands.test.ts` with a red-first test asserting the exact `-bor` expression is present and runs before `Invoke-WebRequest`.

## Explicitly out of scope

This PR does **not** touch the Windows `service install` idempotency half of #4586 (Go agent: stopping an existing service before replace, updating an already-registered service instead of failing with "service already exists", the `agent.yaml.partial` rename contention). That needs its own design — issue stays open, using `Refs #4586` rather than `Closes #4586`.

## Test plan

- [x] Red first: new test failed against the pre-fix `installCommands.ts` (`SecurityProtocol` substring absent).
- [x] `cd apps/web && npx vitest run src/lib/installCommands.test.ts` — 13/13 passed (single-fork, targeted).
- [x] `cd apps/web && npx vitest run src/components/setup/EnrollDeviceStep.test.tsx src/components/devices/AddDeviceModal.test.tsx` — 33/33 passed (consumers of `buildInstallCommands`, no coupling to the exact PS string).
- [x] `cd apps/web && npx tsc --noEmit -p .` — clean, exit 0.

### DIFF
--- a/apps/docs/src/content/docs/agents/advanced-install.mdx
+++ b/apps/docs/src/content/docs/agents/advanced-install.mdx
@@ -172,6 +172,11 @@ For MDM, configuration management (Ansible, Chef, Puppet, Salt), golden images, 
   </TabItem>
   <TabItem label="Windows">
     ```powershell
+    # Older PowerShell/.NET defaults (e.g. Windows Server 2016) can omit
+    # TLS 1.2, which makes Invoke-WebRequest fail outright.
+    [Net.ServicePointManager]::SecurityProtocol = `
+      [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
     Invoke-WebRequest `
       -Uri "https://breeze.yourdomain.com/api/v1/agents/download/windows/amd64" `
       -OutFile "C:\Program Files\Breeze\breeze-agent.exe"
--- a/apps/docs/src/content/docs/agents/installation.mdx
+++ b/apps/docs/src/content/docs/agents/installation.mdx
@@ -88,6 +88,11 @@ clear message instead of installing a binary that can't run.
       throw "Breeze: Windows 10 or Windows Server 2016 or later is required (detected $($osv.Major).$($osv.Minor))"
     }
 
+    # Older PowerShell/.NET defaults (e.g. Windows Server 2016) can omit
+    # TLS 1.2, which makes Invoke-WebRequest fail outright.
+    [Net.ServicePointManager]::SecurityProtocol = `
+      [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
     # Download the agent
     Invoke-WebRequest -Uri "https://breeze.yourdomain.com/api/v1/agents/download/windows/amd64" `
       -OutFile "breeze-agent.exe"
--- a/apps/web/src/lib/installCommands.test.ts
+++ b/apps/web/src/lib/installCommands.test.ts
@@ -64,6 +64,20 @@ describe('buildInstallCommands', () => {
       expect(windows).toContain('Invoke-WebRequest');
     });
 
+    it('forces TLS 1.2 before the download for older PowerShell/.NET defaults (#4586)', () => {
+      // Windows Server 2016 / PS 5.1 hosts can default SecurityProtocol to
+      // Ssl3, Tls (no Tls12), which makes Invoke-WebRequest fail outright
+      // with "Could not create SSL/TLS secure channel." Bitwise-OR the flag
+      // in rather than replacing the value, so Tls13 (where present) stays
+      // enabled alongside it.
+      const { windows } = buildInstallCommands(base);
+      expect(windows).toContain(
+        '[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12'
+      );
+      // Must run before the download, not after.
+      expect(windows.indexOf('SecurityProtocol')).toBeLessThan(windows.indexOf('Invoke-WebRequest'));
+    });
+
     it('downloads the agent from the server, not GitHub (#4441)', () => {
       // The server's download route is what serves BYO / self-hosted signed
       // binaries (BINARY_SOURCE=local, or a custom BINARY_GITHUB_REPOSITORY).
--- a/apps/web/src/lib/installCommands.ts
+++ b/apps/web/src/lib/installCommands.ts
@@ -68,9 +68,18 @@ export function buildInstallCommands(opts: InstallCommandOptions): InstallComman
     `$b=[IO.File]::ReadAllBytes("$pwd\\breeze-agent.exe"); ` +
     `if($b.Length -lt 2 -or $b[0] -ne 0x4D -or $b[1] -ne 0x5A)` +
     `{throw "Breeze: downloaded file is not a Windows executable - a captive portal or web filter may be intercepting this network"}`;
+  // Older Windows PowerShell 5.1 hosts (e.g. Windows Server 2016) can default
+  // SecurityProtocol to Ssl3, Tls with no Tls12, which makes
+  // Invoke-WebRequest fail before the agent is even downloaded ("Could not
+  // create SSL/TLS secure channel", #4586). OR the flag into the existing
+  // value rather than replacing it, so Tls13 stays enabled where present.
+  const winTlsCheck =
+    `[Net.ServicePointManager]::SecurityProtocol = ` +
+    `[Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12`;
   const windows =
     `$ErrorActionPreference='Stop'; ` +
     `${winOsFloorCheck}; ` +
+    `${winTlsCheck}; ` +
     `Invoke-WebRequest -Uri "${apiUrl}/api/v1/agents/download/windows/amd64" -OutFile breeze-agent.exe; ` +
     `${winMzCheck}; ` +
     `.\\breeze-agent.exe service install; ${winThrow('service install')}; ` +


ToddHebebrand:
## Review summary

**Automated review (code-review skill, medium effort):** 0 findings. Verified against the actual PR diff (fetched `refs/pull/4730/head` directly): TLS check runs after the OS-floor check and before `Invoke-WebRequest` in all three touched files; `-bor` bitwise-OR syntax and backtick line-continuations in the `.mdx` fences are valid PS 5.1; the two call sites (`EnrollDeviceStep.tsx`, `AddDeviceModal.tsx`) only display/copy the returned string, no coupling to the added line; confirmed `systemScriptLibrary.ts` and `migration/toolkit.mdx` genuinely already set `Tls12` pre-existing (not this PR's doing) and were correctly left untouched.

**Residual, non-blocking observation from the reviewer:** those two pre-existing TLS12 lines (`systemScriptLibrary.ts`, `migration/toolkit.mdx`) use plain assignment (`=`) rather than `-bor`, so on a host with `Tls13` already present they'd clobber it back off. That's pre-existing behavior, not something this diff touches or regresses, and outside the reported bug (they don't have "the gap" — TLS12 is set either way). Flagging for a possible future cleanup, not fixing here to keep this PR scoped to the reported issue.

**Tests/typecheck (foreground, this session):**
- Red-first: new assertion in `apps/web/src/lib/installCommands.test.ts` failed against pre-fix code, passed after the fix.
- `cd apps/web && npx vitest run src/lib/installCommands.test.ts` — 13/13 passed.
- `cd apps/web && npx vitest run src/components/setup/EnrollDeviceStep.test.tsx src/components/devices/AddDeviceModal.test.tsx` — 33/33 passed (consumers of `buildInstallCommands`).
- `cd apps/web && npx tsc --noEmit -p .` — clean, exit 0.

**CI:** all currently-reported checks green; a few (Test Web/API/Agent, Integration Tests) were still running as of this comment.

**Scope:** TLS half of #4586 only (`Refs #4586`, not `Closes`). The Windows `service install` idempotency half (Go agent) is untouched and needs its own design — tracked on the issue, not this PR.

PR #4733, opened 2026-09-03
feat(agent): Linux end-user reboot prompt with deferral (#3207 W4)

### LINKED ISSUE #3197: [Agent] Patch-triggered reboots never warn the user — warning thresholds are unreachable at the hardcoded 5-minute delay

### LINKED ISSUE #4664: W04: Linux parity: linuxsession package + daemon-side zenity/notify-send prompt as the session user
Wave W04 of #3207 — Linux parity: linuxsession package + daemon-side zenity/notify-send prompt as the session user
Branch: `feature/3207-reboot-deferral/wave-4664`

ToddHebebrand:
Closes #4664

Wave W04 of #3207 — **Linux parity**. Linux devices now get the same warning ladder and the same postponement prompt Windows and macOS got in W3.

## Why the daemon draws the dialog

No desktop helper binary ships for Linux — `release.yml` builds the helper for **darwin only** — so on Linux the session broker has no notify-scoped session, `RequestNotificationDecision` returns `delivered: false` immediately, and every interactive rung of the ladder was silently swallowed. There is nothing to hand the prompt to, so the daemon draws it itself and drops to the signed-in user to do it.

This also retroactively closes #3197's Linux gap: the warning ladder now reaches a Linux desktop user instead of relying solely on the `shutdown -r +N` wall message.

## What landed

**New package `agent/internal/patching/linuxsession`** — enumerates local graphical sessions via `loginctl list-sessions` / `show-session`, and builds the environment and credentials a process in one of them needs. No new dependency: `go.mod` has no D-Bus library, so this shells out to `loginctl` (the same vehicle `sessionbroker/detector_linux.go` and `remote/desktop/x11/resolve_linux.go` already use).

Parsing and environment construction are **untagged** and table-driven, so they run in the `test-agent` job on every platform — the same reason `reboot_plan.go` and `reboot_deferral.go` are untagged. Only the exec, which needs `syscall.Credential`, is behind `//go:build linux`.

Sessions are excluded when: `Type` is not x11/wayland, `State=closing`, `Remote=yes`, `Class` is present and not `user` (the display manager's greeter), the uid is unparseable or **zero**, or an x11 session reports no `Display`. `State=online` (a signed-in user on a background VT) is included — they will see the dialog when they switch back. On a Wayland session logind's `Display` property holds the *X11* display, so it is ignored and `wayland-0` assumed.

**`agent/internal/patching/reboot_prompt_desktop.go`** (untagged) holds every decision — the argv handed to `zenity` and `notify-send`, what each exit code means, and how a fan-out across several signed-in users resolves to one answer. `reboot_prompt_desktop_linux.go` is the exec; `reboot_prompt_desktop_other.go` returns `("", false)` everywhere else, so **Windows and macOS behaviour is unchanged**.

**`agent/internal/heartbeat/reboot_prompt.go`** chains the two vehicles: helper first (where it exists it *is* the user's session, and asking it second would draw two dialogs on one screen), daemon dialog second. `shown=false` — not the clicked label — is what advances the chain: it is the only value meaning "nothing reached a person".

## Privilege drop

The daemon runs as root; `zenity` must not. Three things enforce that, and the mechanism is **`SysProcAttr.Credential`** (setgid + setgroups + setuid performed by the kernel at fork time), never `su`/`sudo` — a wrapper would drag in PAM, a shell, and an interpolation surface for nothing:

1. Real and effective uid/gid are the session user's. A uid-0 session is refused outright (`ErrRefusedRootSession`), in the parser *and* again in `Command`.
2. Supplementary groups are set explicitly from the passwd/group database with **gid 0 filtered out**. An empty list would also be safe (Go then calls `setgroups(0, NULL)`), but a user's own groups are what a real login session has.
3. The environment is built from scratch (`CommandEnv`), never inherited — the daemon's `HOME`, `XDG_RUNTIME_DIR`, `DBUS_SESSION_BUS_ADDRESS` and `PATH` are all root's and would either break the dialog or aim it at the wrong bus. `HOME` is *omitted* when unknown rather than guessed, so glibc resolves it from passwd instead of pointing the toolkit at `/root`.

The binary is resolved from a fixed root-owned search path (`/usr/local/bin`, `/usr/bin`, `/bin`) rather than `exec.LookPath`, because `$PATH` is the one input a local user could aim at a binary of their choosing. `cmd.Dir` is `/`, and **nothing is written into the user's home** — this path creates no files at all.

## Why the reply cannot be spoofed

The answer is the **exit code and stdout of a process the daemon forked itself**. There is no socket, no file, no shared channel a third party could write to — only a pipe between the daemon and its own child, and that child runs as the very user whose decision it reports. A different local user cannot influence it without already being able to act as that user, at which point they could simply click the button.

The related trap is inside zenity, not outside it: `--question` gives the Cancel button, the window close box and the ESC key **the same exit code as an `--extra-button`** (1). Putting "Postpone" on Cancel would have turned every dismissal into a spent postponement. So the affirmative takes the OK button (exit 0) and the postponement is an `--extra-button`, which prints its own label on stdout; exit 1 with empty stdout is a dismissal and buys the user nothing. `--no-markup` stops Pango markup in the body from rendering, and no part of this path goes through a shell, so quotes and semicolons in dialog text are inert. (All dialog text is agent-generated from durations and constants — no API- or user-supplied string reaches it — the sanitiser is defence in depth.)

## `(clicked, shown)` on Linux

W3's review found that "nobody saw the dialog" must stay distinguishable from "the user ignored it". Preserved:

| Situation | Result | Manager behaviour |
|---|---|---|
| No graphical session (headless) | `("", false)` | falls back to the plain notification; reboot proceeds on schedule |
| `zenity` absent | `("", true)` after a `notify-send` warning | no duplicate notification; user is warned but cannot postpone |
| Dialog dismissed / ESC / zenity's own `--timeout` | `("", true)` | reboot proceeds; nothing further emitted |
| Exit 1 that looks like a **display failure** (stderr signature, or too fast for a human) | `("", false)` | manager re-warns — see review fixes |
| Killed for overrunning even its own `--timeout` | `("", false)` | it was wedged; nothing can be claimed |
| `zenity` crashed / unknown exit code | `("", false)` | manager still warns through the plain path |
| Button pressed | `(label, true)` | honoured — a printed label is positive proof it rendered |

Multiple signed-in users each get a dialog; the **first answer wins** and the losing dialogs are cancelled, matching W3's helper fan-out.

## Three review findings fixed inline

All three are in `agent/internal/patching/`, which W5 does not touch. Each has a test that fails on the previous code.

1. **`Cancel` ran `abortOSReboot` outside the lock**, so a schedule armed inside that window would reach `OSInvokeAt`, register a new OS countdown, and have the still-pending `shutdown -c` silently kill it — the manager would go on reporting a reboot the OS had been told to forget. The abort deliberately stays outside the lock (it is an exec with a 30s ceiling; holding `mu` across it would stall `State()` and every timer callback), so the window is closed by **refusal**: a new `osAborting` flag makes `scheduleLockedAt` and `Defer` decline for the duration, exactly as they already decline while `osInvoking` is set.
2. **`ComputeDeferral` granted a postponement when `deadline == scheduledAt`.** The clamp pulled the target back onto the time already scheduled; `target.Before(base)` is false when they are equal, so the user pressed Postpone, spent one of a finite budget, and the countdown re-armed at the same instant. Now `!target.After(base)` refuses, no budget is spent, and the rung still warns — with "This restart can no longer be postponed."
3. **`Stop` could return while a committed OS reboot was still firing.** **Decision: a reboot whose OS command has already been invoked proceeds — the OS owns it, and no platform can be relied on to recall it (BSD `shutdown(8)`, so macOS, has no cancel flag).** `Stop`'s job is therefore not to stop it but to refuse to *return* while the invocation is mid-flight; a `Stop` that returned there told its caller the manager was quiescent moments before the machine went down. Implemented with an `osInvokeDone` channel opened under the same lock hold that passes the stopped/generation check, so there is no instant where the reboot is committed but `Stop` cannot see it must wait. The wait is bounded by `osRebootCommandTimeout` (30s). `Stop` before the invoke still cancels the reboot outright, and concurrent `Stop`s are safe.

## Plan non-goal (restated deliberately)

Linux **maintenance** reboots bypass `RebootManager` entirely — `maintenanceRebootWorker` sends the bare `reboot` command. So after this wave, "Linux users can postpone" is true for **patch** reboots and **not** for maintenance ones. Not widened here; documented as a caveat in `apps/docs`.

## Test evidence

Run in a `golang:1.26.6` container (+ `libpcap-dev`) because this dev host cannot run cgo. `-race`, `-count=1`, linux/arm64.

**Verified — green (final head):**

| Package | Tests (incl. subtests) |
|---|---|
| `./internal/patching` | 284 |
| `./internal/patching/linuxsession` | 66 |
| `./internal/heartbeat` | 1060 |
| `./internal/userhelper` | 180 |

- `go vet ./...` (whole agent module) — clean.
- `GOOS=linux|windows|darwin go build ./...` — all three clean.
- `golangci-lint run internal/patching/... internal/heartbeat/...` — 48 issues, **all pre-existing** (errcheck/gofmt/unused in files this PR does not touch); zero in any file added or modified here. CI runs `--new-from-rev`, so none are new.
- `gofmt -l` — clean on every file in this PR.
- Red-first confirmed for all three lifecycle fixes and both new test files before implementing.

**Not verified — needs a real box:** zenity/notify-send *rendering* under an actual Wayland or X11 session, and the setuid drop landing on a live desktop, cannot be exercised in CI. Needs a manual pass on GNOME/Wayland and XFCE/X11 (the plan's own W4 acceptance criterion).

**Pre-existing failure, unrelated:** `TestPamLifetimeV2CanonicalDispatchFixture` fails when only `agent/` is mounted (it reads `../../../packages/shared/src/fixtures/…`). Reproduced identically at the base commit `e6b3fe4a2`; passes with the repo root mounted.

## Review round — findings fixed inline (commit `785c898a7`)

Four independent reviews ran: two Sonnet PR reviewers (code-review, silent-failure-hunt), one Sonnet security review with an explicit privilege-drop lens, and `codex exec` (gpt-5.6-sol, medium, read-only). **Two reviewers independently found the same blocker.** All confirmed findings are fixed in this PR — none deferred.

**Blockers**

1. **`zenity` exit 1 is ambiguous, and the ambiguity silently destroyed the warning.** (silent-failure-hunt + codex, independently.) Cancel, ESC, the close box *and a GTK failure to open the display* all exit 1 with empty stdout. Reading that as a dismissal returned `shown=true`, which suppresses `RebootManager`'s fallback notification — so on any box where the display could not be reached at prompt time (stale Xauthority after a re-login, a compositor restarting, a runtime-dir race) the machine rebooted on a user who was never warned, **and the log guard only fired on *unrecognised* codes, so exit 1 left no trace at all**. Now resolved from two independent signals, either sufficient: a display-failure signature on the child's stderr, and a process lifetime too short for a human to have dismissed anything (`minDialogRenderTime`, 500ms — a GTK display failure exits in tens of ms). A pressed `--extra-button` remains positive proof of delivery whatever the clock says. Every non-clean exit is now logged.
2. **`/usr/local/bin` is `root:staff` mode `02775` by Debian Policy** — group-writable by a documented *non-root* local-admin group. (security review, with the policy citation.) Because the child runs under the **target session user's** uid, a `staff` member could have planted a `zenity` that the daemon then ran inside *another* user's session, and whose exit code decided whether that user's reboot was postponed. That is local-to-local escalation, not self-DoS. `/usr/local/bin` is dropped from both the search path and the child's `PATH` — no package manager installs zenity there — and resolved binaries are now additionally rejected if setuid/setgid, or not root-owned and non-group/world-writable.
3. **`credentialFor` filtered gid 0 out of supplementary groups but accepted a primary gid of 0**, making the filtering decorative — the child would have run with real and effective gid 0 while the code claimed a complete drop. (codex + security review.) Refused now. The decision was split into a pure `credentialFromIDs` so both gates are asserted directly rather than depending on an exotic account existing on the test machine.

**Should-fix, all applied**

- A dialog **we** killed no longer counts as shown. zenity honours `--timeout` by exiting 5, so a process still alive past that was wedged — possibly wedged reaching a display it never got.
- A **real exit status now wins over an expired context**, so a click landing in the same instant as the deadline is not discarded as a timeout. (codex.)
- **`Stop`'s wait is bounded.** The invocation window spans `saveRebootHistory`, a file write with no deadline of its own, so a wedged filesystem could have parked `Stop` forever — the 30s ceiling only covers the shutdown command itself. (codex.)
- **The session cap now applies to graphical sessions, not raw `loginctl` output.** Eight tty or ssh logins could otherwise hide every real desktop and suppress the dialog for everyone else on the box. (codex.)
- **`XAUTHORITY` is resolved and passed for x11 sessions**, and the Wayland socket is probed rather than assumed `wayland-0`. Both were delivery failures that finding 1 would have turned into silent losses. (codex + silent-failure-hunt.)
- **`List` distinguishes "nobody signed in" from "every `show-session` failed"** instead of returning an identical empty slice. (silent-failure-hunt.)
- **`DesktopNotify` logs its enumeration failure at Warn, not Debug** — by the time it runs it is the last delivery path there is, so a Debug line is a failure nobody will ever see. (silent-failure-hunt.)

**Test-coverage findings, all applied** (pr-test-analyzer confirmed all three lifecycle bug-fix tests genuinely fail on the pre-fix code, and found no vacuous assertions)

- `runZenityPrompt`'s exit classification and `DesktopPrompt`'s branching moved into untagged, injectable functions (`classifyDialogRun`, `promptOrNotifyDesktop`) so every branch is covered by tests that run on any platform — they were `//go:build linux` with no direct test.
- **`reboot_prompt_wiring_test.go` extended to guard the new chain.** That file exists precisely because unwrapping this wiring compiles and passes every other test while silently removing the feature; the analyst caught that I had not extended it. It now pins `chainedRebootPrompt`, `chainedRebootNotify`, `patching.DesktopPrompt`, `patching.DesktopNotify` and the helper-session gate.

**Declined, with reasons**

- *TOCTOU between the binary stat and the exec* (codex, nit): unavoidable without `fexecve`, requires write access to a root-owned directory, and is now guarded by the ownership check. Documented in `binary.go` rather than fixed.
- *"First answer wins" lets a standard user postpone a security reboot* (security review, policy nit): that is the W3 design, matches the helper fan-out on Windows/macOS, and is bounded by the server-side `MaxDeferrals`/deadline this PR does not touch. Flagged for the product call, not changed here.
- One code-reviewer reported **0 findings at or above its confidence bar** and independently verified the concurrency paths (no `osInvoking` path missing `finishOSInvokeLocked`; no deadlock in the `Stop` wait; the fan-out's buffered channel correctly sized).

## Deviations from the plan

- **`(clicked, shown)` instead of the plan's `promptLinuxUser(...) string`** — W3 introduced the two-value contract and the review found it load-bearing; a single string cannot tell "nobody saw it" from "ignored it".
- **Postponement is an `--extra-button`, not `--cancel-label`** — the plan mapped `actions[1]` onto Cancel, which would make every dialog dismissal spend a postponement.
- **`notifySendArgs` carries `--app-name`, an expiry and a `--` separator**, and drops an unrecognised urgency (notify-send exits non-zero on one, turning a bad string into an undelivered warning).
- **`Class` and uid-0 exclusions added to the parser** — the plan's fixtures had no `Class` field, but greeter exclusion was required and `Class` is the clean discriminator. Absent `Class` is treated as `user` so older systemd is not silenced.
- **Session cap (8 graphical, 64 enumerated)** and a fixed binary search path with ownership/setuid checks — neither in the plan; both are cheap bounds on a local-user-influenced input.
- **`XAUTHORITY` resolution and Wayland socket probing** — not in the plan, added from review; without them the dialog silently fails to render on display-manager-started sessions.


ToddHebebrand:
**Review run:** four independent reviews on `15d0450bd` — `pr-review-toolkit:code-reviewer` (Sonnet), `pr-review-toolkit:silent-failure-hunter` (Sonnet), a Sonnet security review with an explicit privilege-drop/local-attack lens, and `codex exec` (gpt-5.6-sol, medium, read-only) with the same two lenses.

**Findings:** 3 blockers + 7 should-fix + 2 test-coverage gaps raised → **all 12 fixed inline in `785c898a7`**; 0 outstanding. 2 declined with reasons (recorded in the PR body). The code-reviewer returned 0 findings at its confidence bar and independently verified the concurrency paths.

**The blocker two reviewers found independently:** `zenity --question` exits **1** for Cancel, ESC, the window close box *and* for a GTK failure to open the display. Reading that as a dismissal returned `shown=true`, which suppresses `RebootManager`'s fallback notification — so on any box where the display could not be reached at prompt time, the machine rebooted on a user who was **never warned**, and because the log guard only fired on *unrecognised* exit codes, exit 1 left no trace at all. Now disambiguated by two independent signals (a display-failure signature on stderr; a process lifetime too short for a human to have dismissed anything), with every non-clean exit logged.

**The one I had underweighted:** `/usr/local/bin` is `root:staff` mode `02775` by Debian Policy — group-writable by a *non-root* group. Since the dialog child runs under the **target session user's** uid, a `staff` member could have planted a `zenity` the daemon then ran inside *another* user's session, whose exit code decided whether that user's reboot was postponed. Removed from the search path and the child's `PATH`; resolved binaries are now also rejected if setuid/setgid or not root-owned and non-group/world-writable.

**Tests:** all four affected packages green under `-race -count=1` on linux/arm64 in a `golang:1.26.6` container (this dev host cannot run cgo):

| Package | Tests (incl. subtests) |
|---|---|
| `./internal/patching` | 284 |
| `./internal/patching/linuxsession` | 66 |
| `./internal/heartbeat` | 1060 |
| `./internal/userhelper` | 180 |

`go vet ./...` clean across the whole agent module; `GOOS=linux|windows|darwin go build ./...` all clean; `gofmt -l` clean on every file in this PR; `golangci-lint` reports zero issues in any file added or modified here (the 48 it reports are pre-existing in untouched files, and CI runs `--new-from-rev`).

**Not verified — needs a real box:** zenity/notify-send *rendering* under an actual Wayland or X11 session, and the setuid drop landing on a live desktop, cannot be exercised in CI. The plan's own W4 acceptance criteria call for a manual pass on GNOME/Wayland and XFCE/X11 before this is trusted in the field.

**Status:** review-clean, awaiting maintainer merge. Head SHA `785c898a7236f8999d03916a8df231d4f7d07129`.

PR #4735, opened 2026-09-03
fix(api): stop Number() parsing scientific notation in inboundEmail fixture (#4495)

### LINKED ISSUE #4495: [test] inboundEmail.integration fixture flakes ~0.6%/run: Number(suffix.slice(-4)) parses scientific notation and overflows internal_number varchar(20)
Found while merging #4490 (unrelated to that PR; main is normally green).

`apps/api/src/__tests__/integration/inboundEmail.integration.test.ts` — "CASE 1: a forged cross-partner reference is contained" failed once with a `varchar(20)` overflow on `tickets.internal_number` (value `T-2026-40000000000000000000`, 27 chars).

Root cause: the fixture builds a suffix as `<epoch>-<6 base36 chars>` and then does `Number(suffix.slice(-4))`. When the last four characters happen to form scientific notation (this run drew `4e19`), `Number('4e19').toString()` is 20 digits and the composed internal number overflows the column. Probability ≈ 0.6 % per run; it randomly blocks unrelated merges (the shard was re-run to get #4490 in).

Fix: derive the numeric part from a `parseInt(x, 36) % 10000` (or a counter), never `Number()` on an arbitrary base36 slice; add a fixture assertion that `internal_number.length <= 20`.

Evidence: CI run for #4490 attempt 1 (Integration Tests shard), passing on attempt 2 with no code change.

ToddHebebrand:
## Summary

The `inboundEmail.integration.test.ts` cross-partner isolation fixture (Task 12 of the ticketing phase-4 email-ingest plan) built a "numeric-looking" 4-digit `internal_number` suffix with `Number(suffix.slice(-4))`. `suffix` is `<epoch>-<6 base36 chars>`; when the last 4 characters of that string happen to parse as scientific notation (e.g. `"4e19"`), `Number('4e19')` returns `40000000000000000000` — a 20-digit decimal once stringified — and the composed `T-2026-<n>` value overflows `tickets.internal_number varchar(20)`, failing the insert. Probability ≈0.6%/run; it flaked CI on an unrelated PR (#4490) and required a shard re-run to land.

Root cause confirmed by direct repro against the exact reported input:
```
suffix.slice(-4)  => "4e19"
Number(...)       => 40000000000000000000
composed value     => "T-2026-40000000000000000000" (27 chars, overflows varchar(20))
```
This matches the value reported in the issue (`T-2026-40000000000000000000`) exactly.

## Fix

- New `apps/api/src/services/inboundEmail/fixtureNumbering.ts` — a pure helper `fourDigitSuffix(input, offset?)` that parses the slice explicitly as base36 (`parseInt(x, 36)`, which has no scientific-notation reading, unlike `Number()`) and reduces mod 10000, so the output is always exactly 4 decimal digits (`0000`–`9999`) regardless of what the slice contains.
- `inboundEmail.integration.test.ts` now uses `fourDigitSuffix(suffix)` / `fourDigitSuffix(suffix, 1)` instead of the raw slice / `Number()` formula, plus a fixture-level `expect(internalNumber.length).toBeLessThanOrEqual(20)` assertion as a belt-and-suspenders guard against a future regression of this fixture.
- Co-located `fixtureNumbering.test.ts` covers the regression directly: fed the reported `"4e19"` shape plus several other scientific-notation-like 4-char strings (`"1e99"`, `"2e05"`, `"9e-1"`, `"1e+5"`, `"0e00"`), with and without the `+1` offset, plus determinism and top-of-range wraparound.

**Red-first verification:** before restoring the fixed helper, I temporarily swapped in the exact old `Number()`-based formula and ran the new test file — it failed with `expected '40000000000000000000' to have a length of 4 but got 20` (and a `NaN` case on the offset path), reproducing the bug precisely. Reverting to the fixed helper turned all 4 assertions green.

## What I could / couldn't run

- **Could run:** the new unit-level `fixtureNumbering.test.ts` (pure function, no DB) — 4/4 passing, verified red against the old formula first. All 10 files / 138 tests under `src/services/inboundEmail/` (unit job) pass; the `.integration.test.ts` file is correctly excluded from that run (per `vitest.config.ts`'s existing exclusion for this suite, unaffected by this change). `tsc --noEmit` on `apps/api` — clean (required `NODE_OPTIONS=--max-old-space-size=8192` on this host to avoid an OOM in the type-checker itself, unrelated to the change).
- **Could not run:** `inboundEmail.integration.test.ts` itself — it needs a real Postgres (`__tests__/integration/setup`) which isn't available in this sandbox. The change to that file is a mechanical 1:1 substitution of the buggy inline formula for calls to the now-tested helper (same call sites, same shape of output), so the unit coverage of the helper is the practical proof; a maintainer with a DB can additionally run it directly via `vitest.integration.config.ts`.

Closes #4495

### DIFF
--- a/apps/api/src/services/inboundEmail/fixtureNumbering.test.ts
+++ b/apps/api/src/services/inboundEmail/fixtureNumbering.test.ts
@@ -34,4 +34,29 @@ describe('fourDigitSuffix', () => {
     expect(next).toHaveLength(4);
     expect(next).not.toBe(base);
   });
+
+  it('wraps exactly at the 9999 -> 0000 boundary', () => {
+    // '07pr' is the base36 encoding of 9999 (parseInt('07pr', 36) === 9999),
+    // so this exercises the true top-of-range wraparound the previous test's
+    // name implied but 'zzzz' (-> 9615) doesn't actually reach.
+    expect(fourDigitSuffix('07pr')).toBe('9999');
+    expect(fourDigitSuffix('07pr', 1)).toBe('0000');
+  });
+
+  it('never throws or produces non-4-digit output for the real caller\'s edge-case suffix shapes', () => {
+    // The real fixture builds suffix as `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`.
+    // Math.random()'s base36 tail can be shorter than 4 chars (e.g. when the
+    // random value is close to 0), which pulls the epoch/suffix separator '-'
+    // into the 4-char slice (e.g. "0-i5"). Also cover a fully empty input and
+    // an all-'-' slice as hard edge cases, none of which should throw or
+    // produce output outside the 4-digit contract.
+    const edgeCaseSuffixes = ['0-i5', '5--a', '----', '', 'a', 'ab', 'abc'];
+    for (const suffix of edgeCaseSuffixes) {
+      for (const offset of [0, 1]) {
+        const result = fourDigitSuffix(suffix, offset);
+        expect(result).toHaveLength(4);
+        expect(result).toMatch(/^\d{4}$/);
+      }
+    }
+  });
 });
--- a/apps/api/src/services/inboundEmail/fixtureNumbering.test.ts
+++ b/apps/api/src/services/inboundEmail/fixtureNumbering.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from 'vitest';
+import { fourDigitSuffix } from './fixtureNumbering';
+
+describe('fourDigitSuffix', () => {
+  it('never exceeds 4 characters, even for scientific-notation-shaped slices (#4495)', () => {
+    // Regression inputs from #4495: Number('4e19') parses as scientific
+    // notation (4e19 === 40000000000000000000), which is the exact shape
+    // that overflowed tickets.internal_number varchar(20) in CI.
+    const sciNotationLikeSuffixes = ['4e19', '1e99', '2e05', '9e-1', '1e+5', '0e00'];
+    for (const suffix of sciNotationLikeSuffixes) {
+      const result = fourDigitSuffix(suffix);
+      expect(result).toHaveLength(4);
+      expect(result).toMatch(/^\d{4}$/);
+    }
+  });
+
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer (parallel, Sonnet)

**Findings:** 2 raised → both addressed in `fcb3be168`; 0 outstanding blocking issues.
- pr-test-analyzer (Important): the real caller's `uniqueSuffix` can produce a random segment shorter than 4 chars, pulling the `-` separator into the parsed slice (e.g. `"0-i5"`) — this shape was reachable but untested. Added explicit coverage for that plus empty/short/all-dash inputs.
- code-reviewer (suggestion): the "wraps at the top of the range" test used `'zzzz'` (→ 9615), which doesn't actually hit the true `9999 → 0000` boundary its name implied. Added a test using `'07pr'` (base36 for 9999) that does. Also applied the `Number.isFinite` → `Number.isNaN` stylistic nit.

**Residual (not fixed, out of scope for this PR):** pr-test-analyzer flagged that the same raw `` `T-2026-${suffix.slice(-4)}` `` pattern (without the `Number()` wrapper, so not currently overflow-prone) still appears in `inboundEmailContactRequester.integration.test.ts` (×2), `ticketEmailLinksClaim.integration.test.ts`, and `outbound.integration.test.ts`. None of these reproduce #4495 today — they only slice, never call `Number()` on the result — but a future edit that adds arithmetic to one of them (as this file's `+1` case did) could reintroduce the same bug class under a different file. Leaving this out of scope since #4495 named one specific file/flake and these three have no active bug; flagging for a maintainer to decide whether a follow-up sweep is worth it.

**Tests:** `apps/api/src/services/inboundEmail/fixtureNumbering.test.ts` — 6/6 passing, verified red first against the exact old `Number()`-based formula (produced a 20-char string / `NaN`, matching the reported failure). Full `src/services/inboundEmail/` unit suite (10 files / 140 tests) green; the `.integration.test.ts` file remains correctly excluded from the unit job (needs real Postgres, not run here). `tsc --noEmit` clean on `apps/api`.

Issue #3254, opened 2026-08-08
[API] No console visibility into a scheduled reboot — `get_reboot_status` and `cancel_reboot` have no server-side caller


ToddHebebrand:
## Description

The agent implements a `get_reboot_status` command and **nothing in `apps/api` ever calls it**. There is no way for a technician to see that a device has a reboot scheduled, when it will fire, or to cancel it — even though the agent tracks all three and can cancel on request.

Found while investigating #3197.

## Current state

**The agent side is complete:**

- `get_reboot_status` — `agent/internal/heartbeat/handlers_patch.go:244-254`, returns the full `RebootState` via `rebootStateToMap` (`:256-266`): `pendingReboot`, `rebootScheduled`, `scheduledAt`, `deadline`, `reason`, `notifiedUser`, `notificationSent`, `source`.
- `cancel_reboot` — `handlers_patch.go:231-242`, calls `RebootManager.Cancel()` which stops the timers and runs `shutdown /a` (`agent/internal/patching/reboot_windows.go:104-122`).
- Both are registered (`handlers_patch.go:20-21`) and gated on elevation (`agent/internal/privilege/check.go`).

**The server side uses neither.** `grep -r "get_reboot_status\|cancel_reboot" apps/api/src` returns nothing.

**What the API does know** is a single boolean: `devices.pendingReboot`, fed from the heartbeat field at `agent/internal/heartbeat/heartbeat.go:86` (populated at `:3544-3545` from `patching.DetectPendingReboot()`). That is "this machine needs a reboot", not "a reboot is scheduled for 14:35". `RebootState` never travels in the heartbeat — it only ever leaves the agent as a command *result*.

## Why it matters

After #3197 lands, patch reboots will warn the user on a policy-configured delay (default 15 minutes). During that window the device is in a state the console cannot see and the technician cannot stop. A user who calls the helpdesk saying "it says my machine is rebooting in 10 minutes and I'm mid-presentation" has no path to a cancel that the agent already supports.

This is also a prerequisite for showing deferral state once #3207 lands — a deferral the console can't see is hard to support.

## Proposed

1. Add `rebootScheduled` / `rebootScheduledAt` to the heartbeat payload (`heartbeat.go:79` area, `apps/api/src/routes/agents/schemas.ts:155` area) so the state is ambient rather than poll-only, and persist on `devices`.
[...]

Issue #2138, opened 2026-07-01
Add linked device profiles for multi-boot systems


ramphex:
### Problem

Some physical devices run multiple operating systems through dual-boot or multi-boot setups.

Each OS has its own Breeze agent installation, so the same physical machine appears as multiple devices in Breeze. Only one of those agents can be online at a time because the machine can only be booted into one OS at once.

Today, the inactive OS appears as a separate offline device. This creates noise in device lists, online/offline counts, alerts, monitoring views, and technician workflows. For operators managing dual-boot systems, the offline agent is expected and does not mean the physical device is unavailable.

### Proposed Solution

Add a way to link multiple Breeze devices as boot profiles of the same physical machine.

A linked device group would contain two or more device records that represent the same hardware across different operating systems.

In device lists and dashboards, Breeze could show a single primary row for the linked group when one member is online.

Suggested behavior:

- If one linked device is online, show that device as the active profile.
- Show the active OS/profile clearly, for example: Windows active, Linux inactive.
- Suppress or visually de-emphasize offline linked profiles when another linked profile is online.
- If all linked profiles are offline, show the group as offline and surface the most recently seen profile.
- If more than one linked profile is online at the same time, show all of them or flag the link as conflicting, because that likely means the devices were linked incorrectly or the hardware identity changed.
- Allow technicians to expand the linked group to see all OS profiles, agent versions, last seen times, and historical records.
- Allow linking and unlinking from the Devices UI.
- Keep each agent/device record separate internally so OS-specific inventory, software, scripts, command history, audit logs, and telemetry are preserved.

This should be a UI and monitoring-layer relationship, not a destructive merge of device records.

### Alternatives Considered

Use naming conventions such as `hostname-windows` and `hostname-linux`.

This helps humans identify related devices, but it does not solve offline noise, alerting, device counts, or dashboard clutter.

Delete or retire the inactive OS device.

This loses history and breaks when the machine is later booted back into that OS.

Automatically merge devices by serial number or hardware identifiers.

This may be useful later, but it is risky as the only behavior because hardware identifiers can be missing, duplicated, virtualized, or changed. Manual linking should exist first, with automatic suggestions as a possible enhancement.

Treat the offline OS as a normal offline device.

This is the current behavior, but it creates false operational noise for expected offline agents.

### Component

Web / Devices / Monitoring / Agent Inventory

### Scope

Medium

[...]

Issue #800, opened 2026-05-21
Server should detect main-agent silence while watchdog stays online


bdunncompany:
# Server should detect main-agent silence while watchdog stays online

## Problem

When the watchdog process continues heartbeating but the main agent goes silent (see Layer B issue), the server today only sees the device's main-agent `last_seen_at` clock advance past the offline threshold, flips `status='offline'`, and that's it. The watchdog's heartbeats land on the device row (different columns: `watchdog_last_seen`, `watchdog_status`) but no asymmetry detector exists.

This means an MSP looking at the device list sees "offline" with no indication that **the watchdog is still reporting in** — which is the single most useful signal that the box itself is alive and only the main agent component is wedged. With that signal, the support workflow is "remotely restart the BreezeAgent service" instead of "physically check the machine."

## Proposed fix (Layer C)

When the server processes a watchdog heartbeat for a device whose main-agent `last_seen_at` is older than `3 * heartbeat_interval`:

1. Set a new device flag `mainAgentSilentWatchdogActive` (or surface via existing `status` enum as `agent_silent`).
2. Emit a structured event `device.main_agent_silent` to the event bus so alerting can fire on it independently of the existing "offline" event.
3. Surface this state in the Devices list UI with a different icon / hover text than plain "offline" — e.g. amber "agent silent (watchdog OK)" instead of red "offline".
4. Auto-send the `restart_main_agent` command via the watchdog command channel after 5 minutes in this state (if Layer B isn't shipped yet, this gives the server a recovery path).

## Acceptance

- Migration adds `main_agent_silent_since timestamptz` column to `devices`.
- Heartbeat handler in `apps/api/src/routes/agents/heartbeat.ts` (watchdog branch) sets this column when the asymmetry exists.
- Devices list query and DTO include the flag.
- Web UI renders the new state with distinct styling.
- E2E test: enroll a device, simulate main-agent silence + watchdog heartbeats, verify the flag, event, and UI render.

## Relation to Layer A / B

- Layer A (#fix-tolerant-heartbeat) prevents the server from causing the silence by 400-ing on bad optional fields. Shipped 2026-05-21.
- Layer B (separate issue) makes the watchdog self-heal by restarting the main agent.
- Layer C (this issue) makes the server visible-and-actionable when both A and B have failed or aren't enough.

The three are independent and reinforce each other.

ToddHebebrand:
[...]

PR #4737, opened 2026-09-03
feat(devices): report scheduled-restart status on the heartbeat and show it in the console (#3207 W5)

### LINKED ISSUE #4665: W05: Console visibility: reboot status on heartbeat, devices columns + export policy, device badge
Wave W05 of #3207 — Console visibility: reboot status on heartbeat, devices columns + export policy, device badge
Branch: `feature/3207-reboot-deferral/wave-4665`

ToddHebebrand:
Closes #4665

Wave 5 (final) of #3207 — end-user reboot prompts with deferral. W1–W3 shipped the policy surface, the deferral state machine and the Windows/macOS prompt; the agent has known since W2 that a restart is booked and how many times the user has postponed it, but **none of that reached the server**. A tech could not answer "is this machine waiting to reboot, and why hasn't it?" without running `get_reboot_status` by hand.

W5 delivers the **read** half of console reboot visibility. It deliberately does **not** absorb #3254's `cancel_reboot` UI.

## What changed

**Agent → server.** `HeartbeatPayload.RebootStatus`, a projection of `RebootManager` state (new `agent/internal/heartbeat/reboot_status_report.go`) rather than `RebootState` itself, so agent-internal bookkeeping (`NotifiedUser`, `LastError`, `NotificationsPlanned`) does not start crossing the network the moment someone adds a field to that struct. A test asserts the allowlist holds.

Sent **without `omitempty`**, following the `SecurityCapabilities` precedent in the same struct, because the field is three-way:

| Wire | Meaning | Server does |
|---|---|---|
| key absent | no news — a pre-#3207 agent | leaves stored values alone |
| `null` | nothing scheduled (cancelled, or already fired) | clears the columns |
| object | the live schedule | stores it |

Collapsing absent into null lets every old agent in the fleet wipe the console's view on its next beat; collapsing null into absent strands a cancelled restart on the device page forever.

**Server.** Five nullable **scalar** columns on `devices` — not one jsonb blob. `devices` is an org-cascade table, so every column must be classified in `CORE_TENANT_EXPORT_POLICY`, and CLAUDE.md forces any open container into `excludedOpen`; a jsonb column would have kept reboot status out of tenant exports entirely. All five are registered as `included` in this same commit.

**Console.** `RebootScheduledBadge` beside the existing pending-reboot badge on the device page: "Restart scheduled" (or "Restarting" once the instant has passed), source/scheduled-time/deadline in the tooltip, plus "Postponed N of M" or "Cannot be postponed". 12 new keys × 8 locales.

## Migration

`apps/api/migrations/2026-10-07-100000-devices-reboot-schedule-columns.sql`

Sorts after the newest migration on `origin/main` (`2026-10-06-110000-portal-visibility-flags.sql`), re-verified against a fresh fetch immediately before commit. `check-migration-naming.sh` OK at commit and push. Idempotency proven by replaying the file against an already-migrated database (all five `ADD COLUMN IF NOT EXISTS` skipped, `DO $$` constraint block a no-op, column count still 5).

No cascade-list change: this adds **columns**, not a table, and `devices` is already in `CORE_ORG_CASCADE_DELETE_ORDER` and `CORE_DEVICE_CASCADE_DELETE_TABLES` (verified by grep, not assumed). The export policy is the one registration that fires on a new column, and it is done here.

## Three design calls, and why

Each of these was put to an independent advisor (`codex gpt-5.6-sol`, xhigh, read-only) as well as reasoned through here; the advisor agreed on all three.

**1. A fifth column, `reboot_max_deferrals` — a deviation from the plan.** The plan's Task 5.1 lists four columns and drops `maxDeferrals`, even though the heartbeat shape it specifies carries it. Without it the console cannot render "postponed 1 of 3", and cannot tell **"this restart cannot be postponed"** (`0`) from **"this agent predates deferral"** (`NULL`) — a distinction `RebootState`'s own comment (`reboot_manager.go:58-66`) says the console needs. It is read from the agent, not re-derived from the patch policy, because the policy can be edited *after* a restart is already booked; the agent's value is the budget actually in force for the live schedule.

**2. `source` is a bounded `^[a-z0-9_]{1,32}$` token, not a `z.enum`.** The agent echoes back whatever `source` the server put on the `schedule_reboot` command. A hard enum would silently drop the **entire** snapshot the first time a new server-side producer shipped ahead of an API deploy, and would buy no provenance anyway — a compromised agent can claim any allowed value. It is never rendered raw: the console maps known tokens (`patch_job`, `maintenance_window`, `manual`) to localized labels via an explicit switch and falls back to a generic label otherwise.

**3. The snapshot is written whole, not diffed column-by-column.** The device UPDATE already fires on every heartbeat (`lastSeenAt`/`status`/`updatedAt` are unconditional) and none of these columns is indexed, so re-assigning an unchanged value costs no extra tuple, no extra WAL record and no index maintenance — while a per-column diff adds `Date`-vs-`Date` comparison hazards for nothing. The one case that **is** skipped is the fleet steady state: a device reporting `null` whose columns are already clear contributes nothing to the SET list, which is the overwhelming majority of heartbeats. Both behaviours are pinned by tests (mutation-checked, below).

**CHECK constraints enforce non-negativity only.** The real `0..10` range lives in the zod schema, where a violation degrades to dropping one optional field rather than failing the beat. Encoding the ceiling in Postgres as well would turn a future policy change that raises it into a 500 on every heartbeat from an already-scheduled device — a constraint that can only ever fire on a value the server itself decided to allow.

## Test evidence

All **verified** (run in this worktree against an isolated per-worktree Postgres on an ephemeral port, never the shared `:5433`):

| Suite | Result |
|---|---|
| `tenant-export-policy` + `tenantExportErasureRoundtrip` + `tenantCascade` integration | **3 files, 10 tests passed** (real DB) |
| `rls-coverage.integration` (`vitest.config.rls-coverage.ts`) | **100 passed** (real DB) |
| `pnpm db:check-drift` | clean — 644 migration files match the ledger |
| Migration idempotency | replayed against a migrated DB: no-op, 5 columns, 2 constraints |
| API unit: `heartbeat.test.ts`, `schemas*.test.ts`, `autoMigrate.test.ts`, `devices/core*` | **9 files, 443 tests passed** |
| Web: `RebootScheduledBadge.test.tsx` + all `src/lib/i18n/**` + `DeviceDetails.*` | **14 files, 176 tests passed** (includes `localeParity`, `keyUsage`, `translationCoverage`) |
| Go: `go test ./internal/heartbeat/... ./internal/patching/...` | **ok** (8 new tests) |
| Go: `go test -race` (both packages, `golang:1.26.6` container — this host cannot cgo) | **ok, 0 data races** |
| `npx tsc --noEmit` — api and web | clean |
| `npx astro check` (web) | 0 errors |
| `pnpm lint` | 6/6 packages clean |
| `scripts/check-migration-naming.sh --staged` | OK |

**Not vacuous — four mutations were injected and each produced exactly the expected red**, then reverted:
- absent treated as `null` → *"leaves stored values alone when rebootStatus is absent (old agent)"* fails
- clear-guard removed → *"does not touch the reboot columns when null is reported and nothing was stored"* fails
- `maxDeferrals === 0` → falsy check → *"stays silent about deferrals when the agent never reported a budget"* fails
- zod ceiling probe → confirmed the schema drops 99/`-1`/`1.5`/`"lots"` and a markup-bearing `source`

**Not run:** the full `pnpm test` suite and the remaining ~200 integration files (out of scope for this change; the blocking `integration-test` CI job covers them).

## Notes for review

- **The zod-validation cases live in `schemas.heartbeatTolerance.test.ts`, not the route test.** `heartbeat.test.ts` mocks `@hono/zod-validator` wholesale so the handler reads the raw body — the plan's suggested "drops a malformed rebootStatus" route test **cannot** pass there, and would have been a vacuous green if forced. The split is called out in a comment at the end of that describe block.
- **Docs correction.** The plan and the issue describe a zenity dialog for Linux. On `main` today that does not exist — `agent/internal/userhelper/notify_prompt_other.go` returns `shown=false` on every non-Windows/darwin platform, and its own comment says Linux gets its prompt from the daemon side in W4. The docs therefore describe current behaviour (no interactive Linux prompt, notification + `wall` only) and note that parity is tracked separately, rather than documenting something that has not shipped.
- **Agent-file discipline.** W4 is in flight in parallel and owns `agent/internal/patching/`, `linuxsession/`, `userhelper/` and `heartbeat/reboot_prompt.go`. This PR adds a **new** file (`heartbeat/reboot_status_report.go`) and makes exactly two small edits to `heartbeat.go` (one struct field, one builder line). It reads `RebootManager` only through the existing exported `State()`; no new accessor was needed and `reboot_manager.go` / `reboot_deferral.go` are untouched.
- No index on the new columns: nothing filters or sorts on them yet, and an index on a column written by every heartbeat is write amplification with no reader. `filterEngine.ts` is the site to extend if a "restart scheduled" fleet filter is wanted later.


ToddHebebrand:
**Review run:** two independent Sonnet reviewers on `388b54bab` — (1) correctness, tracing the three-way `rebootStatus` contract end to end (agent projection → Go struct tag → zod → persist → device projections → badge), plus the `alreadyClear` short-circuit, `new Date()` range safety, `RebootManager` concurrency, and untrusted-`source` rendering; (2) test quality and silent-failure hunting across all four new/changed test files.

**Findings:** 1 raised → addressed in `fef0081f0`; 0 outstanding.

- **SHOULD-FIX — `core.list-response-shape.test.ts` not extended.** The five new columns were added to both the `GET /devices` select and its hand-maintained response mapper, but left unguarded by the test that exists precisely to catch "selected from the DB, dropped by the mapper". That exact failure has shipped three times (#800/#861/#862 watchdog fields, #1273 `pendingReboot`, #2138/#2308 link-group scalars). Fixed by adding the five `hasOwnProperty` + null assertions, following the file's existing pattern. **Mutation-checked:** deleting the `rebootMaxDeferrals` line from the mapper reddens the new assertions.
- Correctness review returned **0 findings** across all seven surfaces. It independently confirmed two things worth recording: `RebootState` is all scalars so `State()`'s copy-under-lock cannot be torn or aliased, and zod 4.4.3's `datetime({offset:true})` enforces a 4-digit calendar-valid year, so no `Invalid Date` or out-of-`timestamptz`-range value can reach the column and 500 a heartbeat.
- Two NITs were raised and deliberately **not** actioned: no real-DB test that the five columns land in a row (consistent with every other heartbeat-persisted field in that file — `isVirtual`, `batteryStatus`, `watchdogVersion` — none has one), and no `DeviceDetails` wiring test (same-named passthrough between two typed interfaces, so a typo is a `tsc` error, not a silent gap; the badge itself has 13 tests).

**Tests (all verified on the final head, against an isolated per-worktree Postgres — never the shared `:5433`):**
- Contract suites, real DB: `tenant-export-policy` + `tenantExportErasureRoundtrip` + `tenantCascade` — **3 files / 10 tests green**; `rls-coverage` — **100 green**.
- `pnpm db:check-drift` clean (644 migrations match the ledger); migration replayed against an already-migrated DB to prove idempotency (no-op; 5 columns, 2 constraints).
- API unit — **8 files / 376 green** after the fix (`heartbeat`, `schemas*`, `devices/core*`); web — **14 files / 176 green** (badge + all `src/lib/i18n/**` incl. `localeParity`, `keyUsage`, `translationCoverage`).
- Go — `go test ./internal/heartbeat/... ./internal/patching/...` ok; `-race` in a `golang:1.26.6` container (this host cannot cgo) **ok, 0 data races**.
- `tsc --noEmit` (api + web) clean, `astro check` 0 errors, `pnpm lint` 6/6 clean, `check-migration-naming.sh` OK.
- **Not vacuous:** five mutations injected across the diff (absent-treated-as-null, clear-guard removed, falsy budget check, zod ceiling probe, mapper field deleted) — each produced exactly the expected red, then reverted.

**Not run:** the full `pnpm test` suite and the remaining ~200 integration files, which the blocking `integration-test` CI job covers.

**CI on the final head `fef0081f0`:** **67/68 checks pass, 0 failures** — including all four blocking `Integration Tests` shards, `Test API`, `Test Web`, `Test Agent`, `Test Agent (race)`, `Lint`, `Type Check`, `Check Migrations` and `Check Migrations (non-superuser)`. The only non-green entry is the aggregate `CI Success` job, which is sitting QUEUED waiting for a runner rather than failing (an earlier duplicate of that same job already reported SUCCESS). Worth a glance before merging in case it needs a nudge.

**Status:** review-clean, awaiting maintainer merge. Note for the merge order — W4 (Linux prompt) is in flight in parallel and owns `agent/internal/patching/`, `linuxsession/`, `userhelper/` and `heartbeat/reboot_prompt.go`; this PR adds a new file plus two small edits to `heartbeat.go`, so the two should not conflict, but W4 landing will make the Linux paragraph in `patch-management.mdx` stale and it should be refreshed then.

PR #4738, opened 2026-09-03
fix(api): only stamp AI field provenance on a real change (#4466)

### LINKED ISSUE #4466: [AI agents] AI confirming an identical ticket field value flips its provenance to ai_agent
When the ticket-triage finalizer (wave P2-4) proposes a field write whose value is identical to the current value, the CAS field-provenance guard still stamps the field's provenance as `ai_agent` even though nothing changed — a human-set field's provenance can flip to AI-owned with no actual AI edit, weakening the "human-set fields are never overwritten" guarantee the provenance guard exists to provide.

ToddHebebrand:
Closes #4466

## The bug

`applyAiFieldUpdates` (`apps/api/src/services/ticketService.ts`) guards every autonomous ticket-field write with a CAS predicate:

```sql
<column> IS NOT DISTINCT FROM <expectedCurrent>
  AND COALESCE(field_provenance->><field>, '') <> 'user'
```

That predicate proves the field has not **moved** since the caller observed it. It never asks whether the AI is actually **changing** anything — and the stamp to `'ai_agent'` rides the very same predicate. So an AI that merely re-asserts the value already on the ticket satisfied it and took ownership of the field with no edit behind it.

It bites hardest on exactly the fields that most need protecting: one a human set through a path that left no provenance entry. An absent stamp `COALESCE`s to `''`, sails past the `<> 'user'` guard, and the human's value becomes AI-owned — weakening the "human-set fields are never overwritten" guarantee the provenance map exists to provide.

**This is reachable in the real triage pipeline, not just via the tool API.** `filterEligibleFields` (`aiAgents/ticketTriageFindings.ts`) screens proposals on provenance and the confidence floor only — it never compares against the ticket's current value. A confirming proposal therefore mints a `fields` intent and releases straight into the CAS.

## The fix

One more conjunct: `<column> IS DISTINCT FROM <new value>`.

It goes in the **shared** predicate that drives both the field write and the provenance stamp, not just on the stamp — so the two can never drift apart under a later edit. On the value arm it is a no-op by construction: writing a column the value it already holds changes nothing. When value and `expectedCurrent` coincide the predicate becomes `x IS NOT DISTINCT FROM 'high' AND x IS DISTINCT FROM 'high'` — unsatisfiable, so Postgres cannot reach the stamp at all.

This also aligns the AI path with the human one: `updateTicketFields` already stamps only the fields its own diff found changed.

**CAS semantics and the `'user'` guard are untouched**, and so is outcome classification — a confirmation still reports `applied: true` (the field *is* in the desired state), never a spurious `concurrent_change` skip.

## Verification

Red first, both ways.

- **Unit** — `apps/api/src/services/ticketService.aiExecutors.test.ts`, 4 new cases on **compiled SQL** (`PgDialect().sqlToQuery`), matching the suite's existing convention: a mock that echoes `.set()` back cannot tell a real predicate from a missing one. The same-value case resolves both `$n` placeholders and asserts they bind the same value — that *is* the unsatisfiability proof. The changed-value case pins the distinctness check to the **proposed** value, not the observed one; comparing against `expectedCurrent` would just restate the tautology the CAS half already covers and would block nothing.
- **Integration (live Postgres)** — `aiAgentTicketTriage.integration.test.ts`, driving the full `persistTicketTriage → intents → releaseApprovedIntent` pipeline against a ticket already at the proposed priority. **Reverting only the source change turns it red** with `field_provenance` at `{ priority: 'ai_agent' }`; with the fix it stays `{}`.

| Check | Result |
|---|---|
| `vitest run` × 5 affected unit files | 291 passed |
| `vitest run --config vitest.integration.config.ts aiAgentTicketTriage` | 8 passed |
| `tsc --noEmit` (apps/api) | exit 0 |
| `eslint` (3 touched files) | clean |

No migration, no schema change, no tenancy surface touched.

## Noted, deliberately out of scope

`updatedAt` is still bumped unconditionally, so a fully no-op AI confirmation makes the ticket look freshly touched in any list sorted by it — the same root cause, but changing it alters behaviour for the existing `human_set` skip path too and affects ticket ordering. That is a separate behavioural call, not a silent rider on a provenance fix. Happy to take it in a follow-up if you want it.

### DIFF
--- a/apps/api/src/__tests__/integration/aiAgentTicketTriage.integration.test.ts
+++ b/apps/api/src/__tests__/integration/aiAgentTicketTriage.integration.test.ts
@@ -44,6 +44,7 @@ import {
   actionIntents,
   aiAgentRuns,
   aiAgents,
+  ticketCategories,
   ticketComments,
   ticketDrafts,
   tickets,
@@ -55,6 +56,7 @@ import { handleTicketCreatedEvent } from '../../services/aiAgents/ticketHelpdesk
 import { registerAgentRunEnqueuer, type AgentRunEnqueuer } from '../../services/aiAgents/runService';
 import { persistTicketTriage } from '../../services/aiAgents/ticketTriageFindings';
 import { releaseApprovedIntent } from '../../jobs/intentReleaseWorker';
+import { applyAiFieldUpdates } from '../../services/ticketService';
 import type { BreezeEvent } from '../../services/eventBus';
 
 // publishEvent writes to a Redis stream — spy on it so admission/release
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `pr-test-analyzer` (both independent, on `b1df4c9c5`).

**Findings:** 5 raised → 4 addressed in `dbf22752d`, 1 accepted as-is; 0 outstanding.

- `code-reviewer` — **0 consequential findings.** It independently walked the truth table over `(expectedCurrent, currentValue, newValue, provenance)` and confirmed the returned `AiFieldUpdateOutcome` is identical in every cell before and after the change: the only cell where the new conjunct flips predicate truth is the same-value confirmation, where `after.<field> === value` holds either way, so only the side effect (the stamp) differs. Also confirmed `IS DISTINCT FROM` is the NULL-safe operator, the casts match those already in the `SET` arm, atomicity under the row's own lock is unchanged, and both callers are passthroughs. It flagged one cosmetic doc line-wrap — fixed.
- `pr-test-analyzer` — 4 findings, verified empirically by reverting the source file and re-running rather than taking the PR body's word for it:
  - **(8) Both arms in one statement were untested.** The two `CASE` fragments concatenate and their `$n` placeholders are numbered across the whole compiled UPDATE — the shape most likely to expose a parameter-index mistake. **Fixed:** covered at both levels.
  - **(7) The `categoryId` arm had zero live-Postgres coverage** (`fixtureProposal` only proposes `priority`, and compiled-SQL assertions never consult the real schema, so nothing proved the `::uuid` cast). **Fixed:** a direct-call integration case. This one earned its keep — on revert it shows the `categoryId` arm really was stamping (`{priority, categoryId}` → `{priority}`).
  - **(6) `applied: true` on a user-stamped field that already holds the proposed value** was unpinned and is easy to misread as "the AI took the field". **Fixed:** test + the doc comment now states outright that `applied` is read off the RETURNING row and means "the field HOLDS this value", never "this call wrote it".
  - **(5) One test was misfiled under the `#4466` block** — it passes with or without the fix. **Fixed:** moved back under the P2-4 block, where the pre-existing invariant it guards belongs.
  - **(2/6) Two items accepted, no action:** `expectedCurrent: null` collapses to an already-covered case; the compiled-SQL brittleness is the suite's own documented, deliberate tradeoff (a `.set()`-echoing mock is provably vacuous for this path).

**Red-first evidence:** every `#4466` assertion was proven discriminating by reverting only `ticketService.ts` and re-running — 5/5 new unit cases and 2/2 new same-value integration cases go red, the live one with `field_provenance` at `{priority: 'ai_agent'}`. The two non-discriminating tests are deliberate non-regression guards and are labelled as such in place.

**Tests (local, foreground):** 5 affected API unit files **293 passed**; `aiAgentTicketTriage.integration.test.ts` against live Postgres **10 passed**; `tsc --noEmit` exit 0; `eslint` clean on all 3 touched files.

**Caveat, stated plainly:** the local integration run needed `BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=1` — the shared test DB carried one unrelated migration (`2026-10-06-120000-portal-report-self-service.sql`) from another worktree. This suite enumerates no schema, so the drift is inert for it, but CI's own Integration Tests job is the authoritative run. **CI was still in flight at hand-off** — I have not seen it green, so please confirm before merging.

**Status:** review-clean, awaiting CI + maintainer merge. Not merging and not closing #4466 — both are yours to call.

PR #4739, opened 2026-09-03
fix(api/auth): change-password and account-deletion 400 a rejected password, not 401 (#4660)

### LINKED ISSUE #4660: [API][Auth] change-password and account-deletion-request still 401 a rejected password — same mid-flow logout as #4470
**Verified.** Surfaced by the code review on PR #4651 (#4470). Same bug class, different endpoints — deliberately left out of that PR because they are not MFA proof endpoints and changing password-change / account-deletion semantics deserves its own review.

#4470 established the contract: **401 means "the credential that authenticates THIS request is dead"** (bearer, or the login challenge's `tempToken`). A proof the user typed into the request *body* is request data, and rejecting it is a 400 with a stable `code`. That matters because `fetchWithAuth` (`apps/web/src/stores/auth.ts:1287`) hands any 401 to refresh-and-replay → `handleSessionExpired` → logout.

Two authenticated endpoints still 401 a rejected body-supplied password, and both are called through `fetchWithAuth` with **no `skipUnauthorizedRetry` and no status branch** — so a typo can sign the user out:

**1. `POST /auth/change-password`** — `apps/api/src/routes/auth/password.ts:310` returns `{ error: 'Current password is incorrect' }` at **401**.
Callers: `apps/web/src/components/settings/ProfilePage.tsx:434` (`handlePasswordChange`) and `apps/web/src/components/setup/AccountSetupStep.tsx:87`. Neither branches on status; both just render `data.error`. A user who mistypes their current password on the profile page can be signed out instead of being told the password was wrong.

**2. `POST /auth/account-deletion-request`** — `apps/api/src/routes/auth/accountDeletion.ts:265` (no password hash) and `:277` (wrong password) both return `{ error: 'Invalid password' }` at **401**.
Caller: `apps/web/src/components/account/AccountDeletionPage.tsx:98`, same unguarded shape.

Mobile is structurally safe on both — `apps/mobile/src/services/api.ts` excludes every `/auth/*` path except `/auth/me` from its refresh-and-retry, by path rather than by status.

### Suggested fix

Mechanical, using what #4470 already landed. Replace each `c.json({...}, 401)` with

```ts
rejectProof(c, '<same message>', INVALID_CREDENTIALS_CODE, 400)
```

from `apps/api/src/routes/auth/helpers.ts`, giving the uniform `{ error, message, code: 'invalid_credentials' }` body. Keep 401 for the bearer guard only. Both endpoints already return 400 for other input problems (`password.ts:302`, weak-password at `:317`), so 400 is the consistent status here. Update the affected assertions in `apps/api/src/routes/auth/password.test.ts` and `accountDeletion.test.ts`, and note the new 400 in `apps/api/src/openapi.ts`.

### Also worth sweeping in the same pass

Two more `requireCurrentPasswordStepUp` callers still take the 401 default. They are lower risk (their clients do branch on 401, so they need a client change too), but they belong in the same audit:

- `apps/api/src/routes/users.ts:665` — email-change step-up.
- `apps/api/src/routes/authenticator.ts:149` — `POST /authenticator/register-grant`. Note `apps/web/src/components/settings/ApproverDevicesSection.tsx` currently keeps a 401 branch *specifically* for this route; migrating it lets that branch collapse into the `code` check next to it.

And one accuracy bug noted in the same review: `PasskeyChallengeError` (`apps/api/src/services/passkeys.ts:238-263`) is thrown for three different causes — expired challenge, **Redis unavailable**, and a corrupt record. The first is a rejected proof; the other two are infra failures that should be **503**, not 400/401. All three have always shared one status, so this is pre-existing, but it now means an infra outage is labelled `code: 'mfa_proof_invalid'`.

Contract reference: `apps/docs/src/content/docs/reference/users-and-roles.mdx` (the "wrong code vs expired session" aside).

ToddHebebrand:
Closes #4660

## The bug

`POST /auth/change-password` and `POST /auth/account-deletion-request` both answered a **rejected body-supplied password** with **401** — the same status the bearer guard uses. `fetchWithAuth` (`apps/web/src/stores/auth.ts:1313`) hands any 401 to refresh-and-replay and then `handleSessionExpired` → logout + redirect, and **neither caller passes `skipUnauthorizedRetry`**. So a user who mistyped their current password on the profile page, or their password on the delete-account form, could be **signed out mid-flow** instead of being told the password was wrong.

Same bug class as #4470, different endpoints. It was deliberately left out of #4651 because these are not MFA proof endpoints and changing password-change / account-deletion semantics deserved its own review.

## The contract (unchanged from #4470 — this extends its reach)

**401 means "the credential that authenticates THIS request is dead"** — the bearer, or the login challenge's `tempToken`. A proof the user typed into the request **body** is request data, and rejecting it is a **400** with a stable machine-readable `code`.

| Endpoint | Was | Now |
|---|---|---|
| `POST /auth/change-password` — wrong `currentPassword` | `401 {error, message}` | `400 {error, message, code: 'invalid_credentials'}` |
| `POST /auth/account-deletion-request` — wrong password | `401 {error: 'Invalid password'}` | `400 {error, message, code: 'invalid_credentials'}` |
| `POST /auth/account-deletion-request` — account has no password hash | `401 {error: 'Invalid password'}` | `400 {error, message, code: 'invalid_credentials'}` |

Both use `rejectProof(...)` from `apps/api/src/routes/auth/helpers.ts`, so the body shape is identical to the one #4651 landed. Clients branch on `code`, never on the human `error`/`message` text.

### Why 400 is the *consistent* status here, not just the safe one

Each endpoint already answered its **other** input problems with 400, so the old 401 was the outlier:

- `password.ts` — the passwordless-account rejection (`:302`) and the weak-new-password rejection (`:317`) were both already 400. The endpoint is now self-consistent.
- Both are behind `authMiddleware`, which has *already* validated the bearer by the time these lines run. A 401 emitted after that guard passed is, by construction, never about the request's own credential.

### The opacity pair on account-deletion (why both branches moved together)

`accountDeletion.ts` answers "this account has no password hash at all" and "the password is wrong" with the **identical** body, deliberately — otherwise the endpoint becomes an oracle for whether an account is SSO-only. Uniformity is what closes that oracle, so **both branches had to move together**; a status that differed between them would have reopened it. There is now a test that fails if only one is migrated.

Note this reasoning does **not** apply to `change-password`, where the passwordless branch keeps its own distinct 400 message — that account state is already public on `GET /auth/me` as `hasPassword` (#4018), so there is nothing to conceal.

## Client changes: none needed, and that is the point

All three web callers already branch on `res.ok` and read `error` / `message`, and **none of them carried the `skipUnauthorizedRetry` stopgap** — so unlike #4651 there is no client flag to drop:

- `apps/web/src/components/settings/ProfilePage.tsx:434` — reads `errorData.message`.
- `apps/web/src/components/setup/AccountSetupStep.tsx:108` — `extractApiError(data, fallback)`, which returns `body.error` verbatim and ignores the added `code`.
- `apps/web/src/components/account/AccountDeletionPage.tsx:98` — reads `data.error`.

The added `message` / `code` fields are additive; every caller renders the same copy it did before, now without the logout.

**Mobile was never affected.** `apps/mobile/src/services/api.ts` gates its refresh-and-retry on `canRefreshFor()`, which excludes every `/auth/*` path except `/auth/me` **by path rather than by status** — so mobile never had the bug the web client had, and the status change is inert there. (`ChangePasswordSheet.tsx` only reads `err.message`; mobile has no account-deletion caller at all.)

## Scope

Held to the two endpoints the issue names. The issue's "also worth sweeping" section is **deliberately not in this PR** and remains open on #4660:

- `apps/api/src/routes/users.ts:665` (email-change step-up) and `apps/api/src/routes/authenticator.ts:149` (`register-grant`) still take the `rejectionStatus` 401 default. Their clients *do* branch on 401 today, so migrating them needs a paired client change — notably `ApproverDevicesSection`, which keeps a 401 branch specifically for `register-grant`.
- `PasskeyChallengeError` (`apps/api/src/services/passkeys.ts:238-263`) conflates an expired challenge with **Redis being unavailable** and a corrupt record; the latter two should be 503, not a rejected-proof status. Pre-existing and orthogonal.

`/auth/account-deletion-request` has no entry in `openapi.ts` at all (pre-existing gap, unrelated to this change) — not added here rather than documenting a whole new endpoint in a fix PR.

## Verification

Red first: the three migrated assertions were written against the new contract and confirmed failing on `401` before the handlers changed.

- **API** — `src/routes/auth` suite, **654/654** pass (24 files), including:
  - new `POST /change-password` test: wrong current password → 400 + exact `{error, message, code}`, and the rejection is terminal (`db.transaction`, `invalidateAllUserSessions`, `runPostCommitCleanup` all untouched);
  - a control that the passwordless branch keeps its own distinct 400 body;
  - `accountDeletion` wrong-password test migrated 401 → 400 with the full body (the audit-row assertion was kept, not weakened);
  - **new opacity test**: a passwordless account answers *identically* to a wrong password, `verifyPassword` is never called, and the `denied` / `no_password` audit row still fires.
- **Web** — `stores/auth` + `components/{settings,setup,account}` + `lib/apiError`, **1122/1122** pass (93 files).
  - **New `ProfilePage` test** — the one that actually discriminates this diff on the client. It drives the real Change Password form and asserts the server's `message` renders as the inline error, so a 400 whose body shape drifted would surface the generic "failed to change password" fallback and turn this red. Mutation-verified: pointing `handlePasswordChange` at a non-existent field fails it.
  - **New `#4660` block in `stores/auth.test.ts`** — an honest caveat, raised by the test-coverage review and worth stating plainly: these stub `fetch` and set the status themselves, so they **would pass unchanged on `origin/main`**. They document the client half of the contract (a 400 makes exactly one fetch with zero refresh calls and leaves `sessionExpiredReason` null; a genuine bearer 401 still refreshes and replays; neither call site acquires `skipUnauthorizedRetry`) and guard against a future change routing 400s into the refresh path. They are **not** evidence the server fix works — the API tests are. One of them deliberately *demonstrates* the pre-fix harm rather than asserting it: the old 401, when the refresh cannot restore the session, evicts it and redirects to `/login?...&reason=session-expired`.
- `tsc --noEmit` clean in `apps/api` and `apps/web`; `eslint` clean on all changed files.
- Repo-wide sweep for other callers/assertions/docs of these two endpoints (API, web, portal, mobile, e2e, packages, docs): **no other breaks or stale references**.

## Audit-trail note

The global fallback-audit middleware (`apps/api/src/index.ts:801`) classifies 401/403 as `denied` and other 4xx as `failure`, so these three rejections shift `denied` → `failure` on the **fallback** row. Same footnote #4651 carried, and verified inert here: the explicit `writeAuthAudit` rows in `accountDeletion.ts` pass `result` literally and are unaffected, and no consumer distinguishes the two values for these actions — `userRiskScoring.ts:581` buckets `['failure','denied']` together, and the only per-value filter (`routes/devices/events.ts:207`) is scoped to `resourceId = deviceId`, which auth rows never populate.

## Two pre-existing issues found while reviewing (NOT fixed here)

Both are outside the two routes this PR is scoped to, and neither is a regression from it:

- **Mobile discards the server's message entirely.** `ApiError` (`apps/mobile/src/services/api.ts:158`) is a plain `interface` and the throw site (`:532`) throws an object literal, so `err instanceof Error` in `ChangePasswordSheet.tsx:175` is **always false** — mobile shows the generic "Could not change password" for every failure, on any status, before and after this change. Verified by reading all three sites.
- **`/auth/change-password` has no rate limit on current-password guesses.** Unlike its siblings it inlines its own `verifyPassword` rather than using `requireCurrentPasswordStepUp`, which throttles 5 per 5 min. Pre-existing — the old 401 didn't throttle either, it just triggered a refresh that succeeded and replayed — but it means a stolen access token can brute-force the current password unthrottled to take over the account permanently.

## Release note

**Wire-contract change (401 → 400).** `POST /auth/change-password` and `POST /auth/account-deletion-request` now answer a **rejected password** with **`400`** and `code: "invalid_credentials"` instead of `401`. `401` from these endpoints now means only that the bearer token is dead. API clients that treated a 401 from either endpoint as "wrong password" must branch on `400` + `code`; clients that auto-refresh on `401` no longer sign the user out for a typo. First-party web and mobile clients need no change. Documented in `apps/docs/.../reference/users-and-roles.mdx`.

### DIFF
--- a/apps/api/src/openapi.ts
+++ b/apps/api/src/openapi.ts
@@ -1257,6 +1257,10 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
               }
             }
           },
+          // #4660 (extends #4470): a rejected `currentPassword` is 400 with a
+          // stable `code` (`invalid_credentials`), alongside the pre-existing
+          // 400s for a passwordless account and a too-weak new password. 401
+          // is reserved for a dead bearer.
           '400': { $ref: '#/components/responses/BadRequest' },
           '401': { $ref: '#/components/responses/Unauthorized' }
         }
--- a/apps/api/src/routes/auth/accountDeletion.test.ts
+++ b/apps/api/src/routes/auth/accountDeletion.test.ts
@@ -205,16 +205,58 @@ describe('POST /account-deletion-request', () => {
     );
   });
 
-  it('returns 401 when the password does not match', async () => {
+  // #4660 (follows #4470/#4651): the password is body data, not the credential
+  // that authenticates this request, so rejecting it is a 400 with a stable
+  // `code`. A 401 here was handed to `fetchWithAuth`'s refresh-and-replay path
+  // (apps/web/src/stores/auth.ts) and then `handleSessionExpired` — a typo on
+  // the delete-account form signed the user out instead of telling them the
+  // password was wrong. 401 stays reserved for a dead bearer.
+  it('returns 400 + code invalid_credentials when the password does not match (#4660)', async () => {
     verifyPasswordMock.mockResolvedValueOnce(false);
     const res = await postRequest({ password: 'wrong' });
-    expect(res.status).toBe(401);
+    expect(res.status).toBe(400);
     const body = await res.json();
-    expect(body).toEqual({ error: 'Invalid password' });
+    expect(body).toEqual({
+      error: 'Invalid password',
+      message: 'Invalid password',
+      code: 'invalid_credentials',
+    });
     expect(insertReturningMock).not.toHaveBeenCalled();
     expect(writeAuthAudit).toHaveBeenCalledWith(
       expect.anything(),
       expect.objectContaining({ action: 'user.account_deletion.requested', result: 'failure', reason: 'invalid_password' })
+    );
+  });
+
+  // Opacity control for the pair. This endpoint deliberately answers the
+  // "account has no password at all" branch with the SAME body as "wrong
+  // password", so a caller cannot use it to probe whether an account is
+  // SSO-only. Both branches must move to 400 together — a status that differed
+  // between them would reopen that oracle, which is exactly the failure mode
+  // #4018 closed elsewhere. This test fails if only one branch is migrated.
+  it('answers a passwordless account identically to a wrong password — no oracle (#4660)', async () => {
+    dbState.userRow = {
+      id: 'u-1',
+      email: 'user@example.test',
+      name: 'Sample User',
+      passwordHash: null,
+      partnerId: 'p-1',
+      orgId: 'o-1',
+    };
+    buildSelectChain();
+
+    const res = await postRequest({ password: 'anything' });
+    expect(res.status).toBe(400);
+    expect(await res.json()).toEqual({
+      error: 'Invalid password',
+      message: 'Invalid password',
+      code: 'invalid_credentials',
+    });
+    expect(verifyPasswordMock).not.toHaveBeenCalled();
+    expect(insertReturningMock).not.toHaveBeenCalled();
+    expect(writeAuthAudit).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({ action: 'user.account_deletion.requested', result: 'denied', reason: 'no_password' })
     );
   });
 
--- a/apps/api/src/routes/auth/accountDeletion.test.ts
+++ b/apps/api/src/routes/auth/accountDeletion.test.ts
@@ -260,6 +260,18 @@ describe('POST /account-deletion-request', () => {
     );
   });
 
+  // Redis backs the attempt limiter, so losing it must fail CLOSED (503) —
+  // never fall through to the password check unthrottled. Untested before
+  // #4660; added here because this PR reworks the rejection statuses on the
+  // very next lines and a silent 503->200 regression would be invisible.
+  it('fails closed with 503 when Redis is unavailable', async () => {
+    vi.mocked(getRedis).mockReturnValueOnce(null as never);
+    const res = await postRequest({ password: 'correct-horse' });
+    expect(res.status).toBe(503);
+    expect(verifyPasswordMock).not.toHaveBeenCalled();
+    expect(insertReturningMock).not.toHaveBeenCalled();
+  });
+
   it('returns 429 when rate-limited', async () => {
     vi.mocked(rateLimiter).mockResolvedValueOnce({
       allowed: false,
--- a/apps/api/src/routes/auth/accountDeletion.ts
+++ b/apps/api/src/routes/auth/accountDeletion.ts
@@ -23,6 +23,8 @@ import {
   getClientRateLimitKey,
   resolveUserAuditOrgId,
   writeAuthAudit,
+  rejectProof,
+  INVALID_CREDENTIALS_CODE,
 } from './helpers';
 
 const { db } = dbModule;
@@ -262,7 +264,12 @@ accountDeletionRoutes.post(
         userId: auth.user.id,
         email: auth.user.email,
       });
-      return c.json({ error: 'Invalid password' }, 401);
+      // #4660: 400, not 401 — see the wrong-password branch below. This
+      // "account has no password at all" case deliberately answers with the
+      // IDENTICAL body, so the endpoint cannot be used to probe whether an
+      // account is SSO-only; both branches must therefore always carry the
+      // same status too. Migrating only one would reopen that oracle.
+      return rejectProof(c, 'Invalid password', INVALID_CREDENTIALS_CODE, 400);
     }
 
     const passwordOk = await verifyPassword(userRow.passwordHash, password);
@@ -274,7 +281,14 @@ accountDeletionRoutes.post(
         userId: auth.user.id,
         email: auth.user.email,
       });
-      return c.json({ error: 'Invalid password' }, 401);
+      // #4660 (extends the #4470 contract): the password arrived in the
+      // request BODY and is data this handler validates — the credential that
+      // authenticates the request is the bearer, already checked by
+      // `authMiddleware`. A 401 here was handed to `fetchWithAuth`'s
+      // refresh-and-replay path (apps/web/src/stores/auth.ts) and on to
+      // `handleSessionExpired`, so a mistyped password on the delete-account
+      // form signed the user out instead of showing the error.
+      return rejectProof(c, 'Invalid password', INVALID_CREDENTIALS_CODE, 400);
     }
 
     // Idempotency: return the existing pending request if one exists.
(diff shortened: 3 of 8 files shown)


ToddHebebrand:
**Review run:** three independent Sonnet lenses on `55e1402ad` — `code-reviewer` (correctness + the opacity invariant + #4651 pattern fidelity), `silent-failure-hunter` (client rendering, audit trail, rate limiting, vacuous tests), `pr-test-analyzer` (coverage + assertion strength). Plus a separate repo-wide caller/assertion/docs sweep before the PR was opened.

**Findings:** 0 blocking. 4 raised → all 4 addressed in `41bc780f3`; 0 outstanding.

- *(test-coverage, HIGH — framing)* The new `stores/auth` tests stub `fetch` and set the status themselves, so they'd pass unchanged if the routes were reverted to 401 — they describe the client contract but don't discriminate this diff. **Fixed** by adding a `ProfilePage` test that drives the real Change Password form and asserts the server's `message` reaches the inline error, and by correcting the PR body, which had oversold them. The store tests are kept, relabelled as what they are.
- *(test-coverage, MEDIUM)* No Redis-down 503 coverage on `account-deletion-request` — a guard that fails the attempt limiter closed, in a file this PR reworks the adjacent statuses in. **Fixed**, test added.
- *(code-review, LOW)* Dangling "since the release noted below" cross-reference in the docs aside. **Fixed** (independently caught and removed before the review landed).
- *(silent-failure, MEDIUM)* The fallback-audit row shifts `denied` → `failure` for these rejections. **Verified inert** and now documented in the PR body: `userRiskScoring.ts:581` buckets both values together, and the only per-value filter is device-scoped. The explicit `writeAuthAudit` rows pass `result` literally and are untouched.

**Mutation-verified (not just "the tests pass"):** each of the three load-bearing new tests was proven to fail when the behavior it claims to pin is broken — reverting *only* the no-password-hash branch to 401 reds the opacity test; deleting the Redis guard reds the 503 test; pointing `handlePasswordChange` at a non-existent field reds the ProfilePage test. Sources restored byte-for-byte after each.

**Tests:** `apps/api` `src/routes/auth` **655/655** (24 files); `apps/web` stores/settings/setup/account/apiError **1122/1122** (93 files); `tsc --noEmit` clean in both apps; `eslint` clean on all changed files. A reviewer reported a possible ~2-3% intermittent 401 when the two API files are co-scheduled; **I could not reproduce it** in 6 consecutive co-scheduled runs plus two full-suite runs — flagging rather than dismissing, in case it surfaces in CI.

**Not fixed, deliberately out of scope** (both pre-existing, neither a regression from this PR — details in the body): mobile discards the server's error message entirely because `ApiError` isn't an `Error` subclass, and `/auth/change-password` has no rate limit on current-password guesses.

**Status:** review-clean, awaiting maintainer merge. Not merged and #4660 left open by design — both are the maintainer's call.

PR #4741, opened 2026-09-03
fix(api): make both org-move paths lock ticket_alert_links in one order (#4657)

### LINKED ISSUE #4657: [API][Concurrency] ticket_alert_links is locked in opposite order by the ticket-move and device-move paths (40P01 risk)
Found by the PR #4642 re-review. **Pre-existing**, not introduced there — PR #4642 fixed this exact deadlock class for the `ai_agent_runs` / `tickets` pair and this is the remaining instance.

**Problem:** the two org-move paths acquire `ticket_alert_links` on opposite sides of `time_entries` / `ticket_parts`.

- **Ticket axis** — `moveTicketOrg` (`apps/api/src/services/ticketService.ts`): the currency guard locks `time_entries` and `ticket_parts` first, then the `TICKET_ORG_DENORMALIZED_TABLES` loop re-stamps `time_entries`, `ticket_parts`, **`ticket_alert_links`**, `ticket_outbox`, `ticket_attachments` in array order. So: `time_entries`/`ticket_parts` → `ticket_alert_links`.
- **Device axis** — `apps/api/src/routes/devices/moveOrg.ts`: `ticket_alert_links` is rewritten (~line 371, via its `alert_id` join to `alerts.device_id`) **before** the `time_entries` (~397) and `ticket_parts` (~400) rewrites. So: `ticket_alert_links` → `time_entries`/`ticket_parts`.

**The rows genuinely overlap**, which is what makes it a live AB-BA rather than a theoretical one: a `ticket_alert_links` row joining ticket X to an alert raised on device D is reached by a device-move of D (`alert_id IN (SELECT id FROM alerts WHERE device_id = D)`) *and* by a concurrent `moveTicketOrg(X, ...)` (`ticket_id = X`). Each transaction can end up holding the lock the other needs, and Postgres kills one with `40P01` — surfacing as a 500 on an admin action.

Note the contrast with `action_intents`, which is *also* inverted between the two axes but is provably safe: `action_intents_scope_device_chk` / `_scope_ticket_chk` (`apps/api/migrations/2026-09-25-ai-agents-ticket-triage.sql:196-203`) make `scope_device_id` and `scope_ticket_id` mutually exclusive, so the two paths' WHEREs select disjoint rows. That argument does not extend to `ticket_alert_links`.

**Fix direction:** pick one canonical order and make both paths follow it — most likely moving `ticket_alert_links` ahead of `time_entries`/`ticket_parts` on the ticket axis (reordering the `TICKET_ORG_DENORMALIZED_TABLES` array is not sufficient on its own, because the currency guard takes the `time_entries`/`ticket_parts` locks before the loop runs at all), or moving the device path's `ticket_alert_links` rewrite after its two billing rewrites. Whichever way, the transaction-level lock-order comments on both sides should be updated together, and the repo would benefit from a single documented cross-axis order rather than two comments that each reason locally.

The state of play is now recorded in `moveTicketOrg`'s global lock-order comment (PR #4642) so the next person to touch either path sees it.

Related: #4524, PR #4642, #3778, #3911.

ToddHebebrand:
Closes #4657

## The bug

`moveTicketOrg` (ticket axis, `services/ticketService.ts`) and `POST /devices/:id/move-org` (device axis, `routes/devices/moveOrg.ts`) both re-stamp `org_id` on the ticket-linked child tables, and **their row sets genuinely overlap** — that is what makes this a live AB-BA rather than a theoretical one.

A `ticket_alert_links` row joining ticket X to an alert raised on device D is selected by:

- a device-move of D — `alert_id IN (SELECT id FROM alerts WHERE device_id = D)`, and
- a concurrent `moveTicketOrg(X, …)` — `ticket_id = X`.

The two paths took that table on **opposite sides** of `time_entries` / `ticket_parts`:

| | order before this PR |
|---|---|
| Ticket axis | `time_entries` → `ticket_parts` → **`ticket_alert_links`** |
| Device axis | **`ticket_alert_links`** → `time_entries` → `ticket_parts` |

Each transaction can end up holding the lock the other needs, and Postgres kills one with `40P01` — surfacing as a 500 on an admin action.

Pre-existing, not introduced recently: #4642 fixed this exact class for the `ai_agent_runs` / `tickets` pair and flagged this as the remaining instance.

## The fix

One canonical order, followed by both paths:

```
tickets → time_entries → ticket_parts → ticket_alert_links → ticket_attachments
```

The device path's `ticket_alert_links` UPDATE moves from before the currency guard to just after the `ticket_parts` rewrite. No other statement moves.

**Why this direction and not the other.** The shared currency guard (`services/ticketMoveCurrencyGuard.ts`) takes `time_entries` then `ticket_parts` `FOR UPDATE` on **both** axes, after `UPDATE tickets` and before any rewrite. Those two are pinned to the front by a constraint neither mover controls, so everything else the movers share has to follow them. Reordering the ticket-axis array alone could not have fixed it — the guard takes its locks before that loop runs at all.

**FK direction.** All four tables are siblings: each is a child of `tickets` and none references another, so children-before-parents does not constrain their relative order. `tickets` itself is locked ahead of all four on both axes either way, before and after this change.

**Behaviour is unchanged** — same statements, same rows, same final state. Only the order within one transaction differs. Moving past the guard is safe: the guard touches only `time_entries`/`ticket_parts`, and a guard throw rolls the whole transaction back, so nothing is left half-restamped. The alert subselect still resolves after the generic loop re-stamps `alerts` (`alerts.device_id` is not what the loop changes, and the request context spans both orgs).

## One documented order instead of two local comments

The issue's closing note — *"the repo would benefit from a single documented cross-axis order rather than two comments that each reason locally"* — is the substance of the rest of the diff. Before this PR each path carried its own lock-order comment, and the two reached **opposite** conclusions without either being wrong on its own terms.

New `services/ticketOrgMoveLockOrder.ts` states the order once, with its rationale, and hosts the ticket-axis table list (moved verbatim out of `ticketService.ts`). The move is what lets a contract test import the list without booting a live `db` pool — `ticketService.ts` pulls one at import time.

`ticketService.ts`, `routes/devices/core.ts`, `routes/devices/moveOrg.ts` and `ticketMoveCurrencyGuard.ts` now all point at that one file rather than restating the order.

## Why it cannot drift again

Three links, all in the unit job — no database needed:

1. **`ticketOrgMoveLockOrder.test.ts` (new)** — the two lists must agree on every table they share; `ticket_alert_links` must follow `time_entries`/`ticket_parts` on both axes (named as the #4657 regression, so a failure says what broke); and the documented order must match what the lists actually do.
2. **`moveOrg.test.ts` (new assertion)** — pins the device path's *real statement stream* to that order, so moving the UPDATE without touching `CUSTOM_ORG_REWRITE_TABLES` is caught here rather than in production. The pre-existing `updatedTables` assertion already spreads that array, so the list and the statements are coupled in both directions.
3. **The ticket path needs no statement check** — its loop iterates the array literally.

## Verification

Red first, on both new guards, against unmodified code:

- contract test: 3 failures, incl. `device axis: ticket_alert_links must follow time_entries: expected 0 to be greater than 1`
- statement-stream assertion: `expected [104, 105, 102, 106] to deeply equal [102, 104, 105, 106]` — `ticket_alert_links` at 102, ahead of `time_entries` at 104

Green after:

- `src/routes/devices/` — 43 files, 568 tests passed
- `src/services/ticket*` + `src/routes/tickets/` — 34 files, 803 tests passed
- `tsc --noEmit` clean; `eslint` clean on all changed files


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer, plus a targeted third-axis/stale-reference sweep (is any *other* path locking these four tables in a conflicting order?).

**Findings:** 5 raised → 4 addressed in `bb3d9e1`, 1 rejected on verification; 0 outstanding.

| # | Finding | Outcome |
|---|---|---|
| 1 | `moveOrg.test.ts` comment claimed the new test put "the guard's FOR UPDATE locks in the stream" — the guard is mocked file-wide, so it never runs | Fixed — reworded to what is actually proven |
| 2 | Same test compared the index array to a sorted copy of itself — sound, but reads as possible vacuity | Fixed — pairwise `toBeLessThan` (house idiom), plus `>= 0` guards on every index so a deleted statement can't pass as `-1` |
| 3 | Moving `TICKET_ORG_DENORMALIZED_TABLES` into the new module dropped the `ticket_attachments` rationale (W08 #3902, S3 objects untouched) | Fixed — restored; that module claims to be the single home for "which tables and why" |
| 4 | Ticket axis had **membership-only** coverage (`arrayContaining`) where the device axis has a runtime order check — the loop could be rewritten as hand-written statements in a different order with nothing failing | Fixed — now asserts the executed sequence equals the shared constant |
| 5 | Claim that org-**erasure** (`cascadeDeleteOrg`) is a conflicting third axis | **Rejected** — verified it gives each table its own transaction (`tenantCascade.ts:901-903`), so it never holds two of these locks at once |

**Finding 4 proven non-vacuous by mutation**, not just by review: reversing the ticket-axis loop while leaving the shared array untouched fails 2 assertions; reverting restores 203 green. Both original guards were also written red-first — the contract test failed 3 ways, and the statement-order test failed with `[104, 105, 102, 106] != [102, 104, 105, 106]` (`ticket_alert_links` at 102, ahead of `time_entries` at 104).

**Tests:** `src/services/` — 907 files / 16,037 passed. `src/routes/devices/` — 43 files / 568 passed. Affected set single-fork after the review fixes — 5 files / 264 passed. `tsc --noEmit` clean; `eslint` clean on all changed files. No migrations or schema changes, so no RLS/cascade/export-policy registration is in scope.

---

**One thing for the maintainer to decide — a third walker this PR deliberately does not fix.**

`orgMerge.ts` runs its **entire** merge in one transaction over `topologicalCascadeOrder()` reversed. For these four siblings that yields `time_entries → ticket_parts → ticket_attachments → ticket_alert_links` — which disagrees with the canonical order on the **last pair**. It is held apart from the two movers by `fenceLoser()` stamping the loser org `status='merging'` before the transaction opens (the move routes refuse a fenced org), i.e. by an **app-layer fence, not by lock ordering** — so the window is not closed for a move already in flight when the fence lands.

Left as-is: #4657's scope is the ticket/device pair, and reordering the merge walk means touching `orgMergeRegistry`'s own ordering contract. It is now documented in `ticketOrgMoveLockOrder.ts` rather than papered over, and the module's doc no longer claims to be "THE single cross-axis order". Happy to open a follow-up issue if you want it tracked.

**Also flagged, not done:** the test reviewer recommended a real-Postgres concurrent-deadlock repro. I did not add one — it needs two connections and a synchronisation barrier to make them genuinely contend, and an "asserts no 40P01" test passes vacuously whenever the two transactions happen to serialize, which is a familiar source of CI flake in a blocking job. The static contracts catch the realistic regression (list or statement drift). Your call if you'd rather have it.

**Status:** review-clean, awaiting maintainer merge. Not merged, issue left open and assigned.

PR #4742, opened 2026-09-03
fix(api/ai): don't evict an actively-processing Anthropic session; expire the row on both staleness paths (#4514)

### LINKED ISSUE #4514: [API/AI] streamingSessionManager (default Anthropic path) has the same two eviction defects as #4384
## Description

`streamingSessionManager.ts` — the **default** Anthropic path, used unless `MCP_LLM_PROVIDER=openai-compatible` — carries the same two eviction defects that #4384 reported against the openai-compatible manager. #4406 fixed only the openai one; this is the port.

Verified against `origin/main` (`6f496f4d0`).

**1. Eviction can abort an actively-processing turn.** `session.lastActivityAt` is assigned in exactly one place, `streamingSessionManager.ts:706` (the reuse branch of `getOrCreate`). Neither `evictStaleSessions()` (~L1677) nor `evictLeastRecentlyActive()` (~L1709) checks `state === 'processing'`, so under cap pressure the least-recently-*created-or-resumed* session can be the one currently streaming, and `remove()` aborts it mid-turn. The manager does check `state === 'processing'` elsewhere (L615, L666, L1096) — just not in the two eviction paths.

**2. Idle and LRU evictions never mark the DB row expired.** Only the `age > SESSION_MAX_AGE_MS` branch (~L1697) writes `status: 'expired'`. The idle branch and `evictLeastRecentlyActive()` drop the session from memory and leave `ai_sessions.status = 'active'`.

## Notes from fixing the openai twin (#4406)

Worth reading before porting — three things were not obvious:

- **Do not blanket-skip `state === 'processing'`.** `runTurn`'s outer `try` has a `finally` but no `catch`, so a throw can skip the trailing `state = 'idle'` and wedge a session in `processing` forever. #4406 gates protection on `state` **and** recent stream progress (`PROCESSING_STALL_TIMEOUT_MS`), refreshing `lastActivityAt` at turn start and on each delta, so a dead turn stays reclaimable.
- **Capacity eviction should probably NOT expire the row.** An LRU victim is a healthy conversation dropped for our own capacity reasons and is transparently resumable from `ai_messages`; expiring it turns a transient server condition into a hard 410. #4406 expires on the staleness paths only. Whether that is the right call is still open (see the PR discussion) — but the two managers should agree.
- **The 24h expire write at L1697-1704 has no `runOutsideDbContext`.** It is safe *today* only because this singleton is module-level (L1733), so its eviction timer inherits no request context. That is an accident of construction order, not a guarantee — the openai manager is a *lazy* singleton built inside a request, where the same code silently ran under the requester's GUCs. If any eviction write here is ever reached from a request path (e.g. an LRU eviction triggered by `getOrCreate`), it needs the escape: `withDbAccessContext` joins an already-open context rather than replacing it, and a cross-tenant UPDATE then matches zero rows under RLS and reports success.

## Affected Files

- `apps/api/src/services/streamingSessionManager.ts` (`evictStaleSessions`, `evictLeastRecentlyActive`, and wherever `lastActivityAt` should be refreshed during a turn)

## Reference

Fix for the openai-compatible twin: #4406 (`apps/api/src/services/llm/openaiSessionManager.ts` + `openaiSessionManager.eviction.test.ts`).

ToddHebebrand:
Closes #4514

Ports #4406 (the openai-compatible manager, #4384) to `streamingSessionManager` — the **default** path, used unless `MCP_LLM_PROVIDER=openai-compatible`. #4406's commit body deferred this file explicitly rather than doubling that PR into the primary AI path.

## What was wrong

**1. Eviction could abort an actively-processing turn.** `session.lastActivityAt` was assigned in exactly one place — the reuse branch of `getOrCreate()` — and neither `evictStaleSessions()` nor `evictLeastRecentlyActive()` checked `state === 'processing'`. Under cap pressure the least-recently-*created-or-resumed* session could be the one currently streaming, and `remove()` aborts its controller, closes the SDK query and closes the event bus mid-turn.

**2. Idle eviction never retired the DB row.** Only the `age > SESSION_MAX_AGE_MS` branch wrote `status: 'expired'`; the idle branch dropped the session from memory and left `ai_sessions.status = 'active'` on a session that no longer exists.

## The fix

**In-flight turns are protected.** Both eviction paths skip a session with a turn in flight. Liveness is `state` **and** recent stream progress, never `state` alone: `lastActivityAt` is refreshed at `tryTransitionToProcessing`, at `message_start` and on every text delta, and a processing session with no progress for `PROCESSING_STALL_TIMEOUT_MS` (10 min) is a dead turn that stays reclaimable. Without that, a background processor that leaves `state` pinned at `processing` would wedge a session forever. When *every* session is mid-turn, LRU evicts nothing and raises a throttled capacity alarm — overshooting the soft cap is self-correcting, corrupting a live turn is not.

**Two keepalives, not one — the one real deviation from the twin.** #4406 refreshes on content deltas only. This path can spend ~7 min inside a single turn with no text delta at all (a tier-3 tool blocking on `waitForApproval` for 300s, then executing under a 120s vision budget — see the `SDK_TURN_TIMEOUT_MS` comment), so `message_start` is a keepalive too; otherwise a tool-only turn is indistinguishable from a wedge. That worst case is also what sizes the 10-minute window here.

**Both staleness paths retire the row; capacity eviction does not.** The idle and 24h paths now write `'expired'` (an existing enum value, no migration), matching what `runPreFlightChecks` writes lazily on the next request — and that preflight branch already reports an expired session as *expired* (410, not 400) since #4406, so one contract holds however the row got there.

LRU deliberately does **not** expire, matching #4406 and the issue's own note ("Capacity eviction should probably NOT expire the row … the two managers should agree"). An LRU victim is a healthy conversation dropped for our own capacity reasons: history lives in `ai_messages` and the session resumes from `sdkSessionId`, so the user's next message transparently recreates it — exactly like a deploy, which `shutdown()` is likewise careful not to expire. Stamping `'expired'` would turn a transient server condition into a hard 410 for a conversation minutes old.

## Tenancy

Expire writes are grouped one statement per org (`inArray`) and serialized: previously this could open one transaction per session — up to `MAX_ACTIVE_SESSIONS` (200) — against a `DB_POOL_MAX` of 30 shared with live traffic, and idle cohorts expire together after a spike.

`runOutsideDbContext` is re-entered **per statement**, never hoisted around the loop: `withDbAccessContext` JOINS an already-open context instead of replacing it, and `AsyncLocalStorage.exit()` does not cover iterations resuming after an `await`, so a hoisted escape strands orgs 2..N under the caller's GUCs where the UPDATE matches zero rows under RLS and reports success. Today the only caller is the module-level eviction timer (no ambient context), so this is defense in depth — but the issue calls that safety out as an accident of construction order, and the sibling LRU path already runs from `getOrCreate()` on the request path. `dbWriteExpectingRows` makes a zero-row batch — the RLS signature — visible rather than silent.

The constructor does **not** need #4406's `setInterval` escape: that manager is a *lazy* singleton built inside a request, so its timer inherits the requester's scope on every tick; this one is module-level and starts clean. That difference is documented in the constructor rather than left implicit.

No schema change, no migration, no new registration lists.

## Verification

- **`streamingSessionManager.eviction.test.ts` — 26 tests, confirmed red first: 17 failed / 9 passed against the unfixed source.** Every load-bearing assertion was mutation-tested rather than assumed:

  | Mutation | Result |
  |---|---|
  | Flip `<=` to `<` at the stall boundary | 1 test fails |
  | Hoist the context escape out of the per-org loop | 1 test fails |
  | Remove the `content_delta` keepalive | 1 test fails |
  | Remove the `message_start` keepalive | 1 test fails |
  | Expire on the LRU path too (the literal reading) | 2 tests fail |
  | Zero-row control: make the "stays quiet" fixture return no rows | 1 test fails |

  Two boundary tests passed *vacuously* in the red run — the unfixed module exports no `PROCESSING_STALL_TIMEOUT_MS`, so the offset is `undefined` — and are covered by the boundary mutation instead. They also needed `vi.useFakeTimers()`: with a live clock the at-edge session drifted one millisecond past the window and was spared by position, not by protection.

- `vitest run src/services/streamingSessionManager src/services/aiAgentSdk.test.ts src/services/sentryEventCodes.test.ts` — **9 files, 221 passed**.
- `vitest run src/routes/ai src/routes/scriptAi src/services/llm src/services/aiAgent` — **98 files, 2474 passed**.
- `tsc --noEmit` exit 0; `eslint` clean on both touched files.

The db mock uses a real `AsyncLocalStorage` rather than a save/restore flag, so it cannot drift from the join-if-open contract it stands in for, and `dbWriteExpectingRows` is left unmocked so the zero-row alarm is genuinely exercised.

### DIFF
--- a/apps/api/src/services/sentry.ts
+++ b/apps/api/src/services/sentry.ts
@@ -105,6 +105,16 @@ const ALLOWED_TAG_NAMES = new Set([
   // cluster groupable and actionable.
   'body_limit_rule',
   'body_limit_max_size',
+  // #4514: the AI session cap alarm fires when EVERY in-memory session is
+  // mid-turn, so LRU can evict nothing and MAX_ACTIVE_SESSIONS is exceeded.
+  // `scrubEvent` deletes `message`, so without this the event says only that it
+  // happened — and a single-request blip is then byte-identical to a manager
+  // wedged at several times its cap, which is exactly the distinction that
+  // decides whether to page. Closed four-value set from
+  // `bucketSessionOvershoot` (services/streamingSessionManager.ts); the raw
+  // count would be unbounded cardinality, and the bucket carries no tenant,
+  // device or session identifier.
+  'ai_session_cap_bucket',
   // BREEZE-18: the required `captureMessage` discriminator. `scrubEvent`
   // deletes `message`, `logentry` and `extra` from every event, so before this
   // existed any captureMessage that happened not to carry one of the tags above
--- a/apps/api/src/services/streamingSessionManager.eviction.test.ts
+++ b/apps/api/src/services/streamingSessionManager.eviction.test.ts
@@ -364,6 +364,31 @@ describe('StreamingSessionManager eviction (#4514)', () => {
       expect(capacityAlarms()).toHaveLength(1);
     });
 
+    it('tags the capacity alarm with a magnitude bucket that actually varies', () => {
+      // `scrubEvent` deletes `message` from every outbound event, so a tag is
+      // the alarm's only surviving channel. Without a varying one, a blip and a
+      // wedged manager are byte-identical in Sentry.
+      const small = new StreamingSessionManager();
+      const huge = new StreamingSessionManager();
+      try {
+        seed(small, 's1', 'org-a', { idleFor: MINUTE, state: 'processing' });
+        for (let i = 0; i < MAX_ACTIVE_SESSIONS * 2 + 1; i++) {
+          seed(huge, `h-${i}`, 'org-a', { idleFor: MINUTE, state: 'processing' });
+        }
+
+        internals(small).evictLeastRecentlyActive();
+        internals(huge).evictLeastRecentlyActive();
+
+        const buckets = capacityAlarms().map(
+          (c) => (c[1] as { tags?: Record<string, string> }).tags?.ai_session_cap_bucket,
+        );
+        expect(buckets).toEqual(['at-cap', 'runaway']);
+      } finally {
+        small.shutdown();
+        huge.shutdown();
+      }
+    });
+
     it('throttles the capacity alarm instead of firing it on every request', () => {
       seed(manager, 's1', 'org-a', { idleFor: 5 * MINUTE, state: 'processing' });
 
@@ -664,7 +689,14 @@ describe('StreamingSessionManager eviction (#4514)', () => {
       await settle();
       await settle();
 
-      expect(captureExceptionMock).toHaveBeenCalled();
+      // `org_id` is allowlisted (sentry.ts ALLOWED_TAG_NAMES) and survives the
+      // scrubber. Without it every org's failure collapses into one
+      // untriageable issue and the on-call cannot tell WHICH tenant is stranded.
+      expect(captureExceptionMock).toHaveBeenCalledWith(
+        expect.any(Error),
+        undefined,
+        { org_id: 'org-a' },
+      );
       // org-b must still be retired — abandoning the loop strands exactly the
       // 'active' rows this helper exists to clean up.
       expect(capturedExpires).toHaveLength(2);
--- a/apps/api/src/services/streamingSessionManager.eviction.test.ts
+++ b/apps/api/src/services/streamingSessionManager.eviction.test.ts
@@ -1,0 +1,737 @@
+/**
+ * Eviction contract for the DEFAULT (Anthropic) session manager — issue #4514,
+ * the deferred twin of #4384/#4406 (llm/openaiSessionManager.eviction.test.ts).
+ *
+ * Two defects are pinned here:
+ *   1. Neither eviction path checked `state === 'processing'`, so under cap
+ *      pressure the least-recently-*created-or-resumed* session could be the one
+ *      currently streaming — `remove()` aborts its controller and closes the SDK
+ *      query and the event bus mid-turn.
+ *   2. Only the 24h-age branch retired the DB row; idle eviction left
+ *      `ai_sessions.status = 'active'` on a session that no longer exists.
+ *
+ * Protection is deliberately NOT `state` alone — see PROCESSING_STALL_TIMEOUT_MS.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { PgDialect } from 'drizzle-orm/pg-core';
+import type { SQL } from 'drizzle-orm';
+
+const {
+  queryMock,
+  dbUpdateMock,
+  capturedExpires,
+  mockState,
+  contextStore,
+  captureExceptionMock,
+  captureMessageMock,
+} = vi.hoisted(() => {
+  const { AsyncLocalStorage } = require('node:async_hooks') as typeof import('node:async_hooks');
+  return {
+    queryMock: vi.fn(),
+    dbUpdateMock: vi.fn(),
+    capturedExpires: [] as {
+      context: unknown;
+      set: Record<string, unknown>;
+      where: SQL;
+    }[],
+    mockState: {
+      /** Context in force at the moment the UPDATE is issued. */
+      effectiveContext: null as unknown,
+      /** Rows the next UPDATE resolves with. Empty = the RLS-denial signature. */
+      nextRows: [{ id: 'row' }] as { id: string }[],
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (correctness, twin-parity, test quality) + `silent-failure-hunter` (swallowed errors, alarm fidelity).

**Findings:** 2 raised → both addressed in `2595a2f54`; 0 outstanding.

- `code-reviewer`: **0 findings.** Traced the silence budget end-to-end (`APPROVAL_WAIT_BUDGET_MS` 5 min → `SDK_TURN_TIMEOUT_MS` 6 min, which itself forces `state = 'idle'`) and confirmed `PROCESSING_STALL_TIMEOUT_MS` has headroom either way the arithmetic is read. Confirmed repo-wide that the in-memory `ActiveSession.lastActivityAt` is read only inside `streamingSessionManager.ts` — it is a different field from the DB column of the same name, so the new keepalives break no consumer. Confirmed `'expired'` is a pre-existing `aiSessionStatusEnum` value and no schema/migration file is touched, so no cascade or export-policy registration applies.
- `silent-failure-hunter`: fire-and-forget chain, `finally`-masking, zero-row RLS detection and the per-iteration context escape all **clean**. Two real **observability** gaps — both alarms fired correctly but arrived at Sentry with nothing actionable, since `scrubEvent` deletes `message` and tags are the only surviving channel:
  1. `markSessionsExpired`'s per-org `captureException(err)` dropped the org id. `org_id` **is** allowlisted; the comment justifying its omission was right about session ids and wrong about the org. Every tenant's stranded-row failure was collapsing into one untriageable issue.
  2. `ai_session_cap_all_in_flight` carried no tags, so a one-request overshoot and a manager wedged at several times its cap were byte-identical — exactly the distinction that decides whether to page. Now carries a bucketed magnitude (`at-cap` / `over-cap` / `far-over-cap` / `runaway`) with `ai_session_cap_bucket` allowlisted, following the `db_pool_health_verdict` / `event_loop_lag_bucket` precedent.

**Tests:** `streamingSessionManager.eviction.test.ts` — 27 passed, confirmed **red first** (17 failed / 9 passed against the unfixed source). Nine mutations run, every one caught:

| Mutation | Result |
|---|---|
| Flip `<=` to `<` at the stall boundary | 1 test fails |
| Hoist the context escape out of the per-org loop | 1 test fails |
| Remove the `content_delta` keepalive | 1 test fails |
| Remove the `message_start` keepalive | 1 test fails |
| Expire on the LRU path too | 2 tests fail |
| Zero-row control returns no rows | 1 test fails |
| Drop the `org_id` tag | 1 test fails |
| Collapse `bucketSessionOvershoot` to a constant | 1 test fails |
| Drop the `tags` bag from the capacity alarm | 1 test fails |

Two boundary tests passed *vacuously* in the red run (the unfixed module exports no `PROCESSING_STALL_TIMEOUT_MS`) and are covered by the boundary mutation instead; they also needed `vi.useFakeTimers()`, since with a live clock the at-edge session drifted one millisecond past the window and was spared by position rather than by protection.

Suites: `vitest run src/services/streamingSessionManager src/services/sentry src/services/aiAgentSdk.test.ts` — 13 files, 275 passed. `vitest run src/routes/ai src/routes/scriptAi src/services/llm src/services/aiAgent` — 98 files, 2474 passed. `tsc --noEmit` exit 0; `eslint` clean on all three touched files.

**For the maintainer, two judgement calls carried over from #4406:**
1. **LRU eviction leaves the row `active`**, contrary to the issue's literal wording — deliberate, and what #4406 shipped. An LRU victim is a healthy conversation dropped for our capacity reasons and resumes from `ai_messages` + `sdkSessionId`; expiring it turns a transient server condition into a hard 410. One-line revert if you want the literal reading — but then #4406 should change too, since the issue asks that the two managers agree.
2. **`message_start` as a second keepalive** is the one genuine deviation from the twin, sized to this path's ~7-minute tool-only silence (approval wait + vision budget), which the OpenAI path does not have.

**Follow-up not taken here:** the twin's `ai_session_cap_all_in_flight` alarm (`llm/openaiSessionManager.ts`) has the same untagged shape — left alone rather than widening this PR into the other provider path.

PR #4779, opened 2026-09-03
fix(mobile): make ApiError a real Error subclass so instanceof narrowing works (#4747)

### LINKED ISSUE #4747: [Mobile] ApiError is a plain interface, so err instanceof Error never matches — every change-password failure shows the generic message
**Where:** `apps/mobile/src/services/api.ts:158` declares `export interface ApiError { … }` and the throw site throws an object literal, not an `Error` subclass. `apps/mobile/src/screens/chat/components/ChangePasswordSheet.tsx:176` then does `err instanceof Error ? err.message : 'Could not change password.'`, which is always the fallback branch — the server's message ("current password is incorrect", rate-limited, etc.) is never shown. `apps/mobile/src/lib/errorReporting.ts:33` has the same `instanceof Error` split and so always takes the `extra.apiError` path. Two more private `interface ApiError` copies exist in `services/connectedApps.ts:21` and `services/mobileDevices.ts:27`.

**Fix:** make `ApiError` a real `class ApiError extends Error` (keep `status` / `code` / `message` fields), throw `new ApiError(...)`, and collapse the two duplicate interfaces onto it. Then `ChangePasswordSheet` and `errorReporting` work unchanged. Add a test that a 400 with a server message renders that message in the sheet.

Surfaced during the #4660 fix (PR #4739); out of that PR's scope. Verified against origin/main 8ea4aaf993.

ToddHebebrand:
## Summary

Closes #4747

`apps/mobile/src/services/api.ts` declared `export interface ApiError { message, code?, statusCode? }` and every throw site threw an object literal cast `as ApiError` — never a real `Error` instance. That made `err instanceof Error` always false at every catch site that used it to decide whether to show the server's message, most visibly `ChangePasswordSheet.tsx:176` (`err instanceof Error ? err.message : 'Could not change password.'`), which always fell back to the generic string even when the server sent a specific one ("current password is incorrect", rate-limited, etc.). `lib/errorReporting.ts:33` had the identical split, driving every ApiError through its non-Error normalization branch instead of being captured as-is.

## Fix

- `ApiError` is now `class ApiError extends Error` (mirrors the existing `TimeEntryError extends Error` pattern already used in `services/timeEntries.ts`), keeping the same `message` / `code?` / `statusCode?` fields plus `name = 'ApiError'`.
- All 7 throw sites in `api.ts` (main request-failure throw, `assertCurrentSession`'s two superseded-session throws, the two invalid-login/MFA-response throws, and the two token-refresh throws) now do `throw new ApiError({...})` instead of `throw {...} as ApiError`.
- Collapsed the two duplicate private `interface ApiError` copies in `services/connectedApps.ts` and `services/mobileDevices.ts` onto the exported class (imported from `./api`), and their throw sites now do `throw new ApiError({...})`.
- Updated the now-stale doc comment on `errorReporting.reportInternalError` (it previously said api.ts "throws plain object literals, not Error instances" — no longer true; the comment now explains the defensive branch is for genuinely non-Error throws, e.g. test doubles or `throw 'string'`).

## Repo-wide sweep for regressions (per the issue's ask)

Grepped `apps/mobile/src` for every `instanceof Error`, `as ApiError`, and `ApiError` reference before and after the change:

- **`instanceof Error` sites** (ChangePasswordSheet, errorReporting, ServerSelectScreen, SettingsSheet, searchCoordinator, AlertDetailScreen, TimeSuggestionsScreen, SystemsScreen, aiChat, notifications, timeEntryQueue, appLockStore, csrfToken) — none assume `ApiError` is *not* an Error; they all treat "not instanceof Error" as the exceptional/fallback case. Making `ApiError` a real Error only makes these sites correct instead of always-fallback; none regress.
- **`services/timeEntries.ts`'s `asTimeEntryError`** and **`timeEntryAccess.ts`'s `classifyTimeEntryDenial`** duck-type on `.message`/`.code`/`.status`/`.statusCode` rather than checking `instanceof`, so they're unaffected by ApiError going from a plain object to a class instance (still has all the same enumerable-looking properties via getters... actually own properties, since they're assigned in the constructor).
- **`as ApiError` casts** — all 7 in `api.ts` were the throw sites already covered above; none remain.
- **`connectedApps.ts` / `mobileDevices.ts`** — no `instanceof` checks on their thrown errors elsewhere in the codebase; both files' own duplicate interfaces are gone.

## Tests

Added `apps/mobile/src/services/api.changePassword.test.ts`:

- A unit test that `new ApiError({...})` is `instanceof Error` **and** `instanceof ApiError`, with `name`/`message`/`code`/`statusCode` set correctly.
- An integration-style test (same harness/mocking pattern as the existing `api.mfa.test.ts`: `expo-secure-store`, `@sentry/react-native`, `serverConfig`, `installationId`, `fetchWithTimeout`, `csrfToken` mocked, real `requestWithPrefix`/`coreRequest` machinery exercised) that mocks a 400 response with `{ error: 'Current password is incorrect' }` from `changePassword`, then evaluates the *exact* ternary from `ChangePasswordSheet.tsx`'s catch block (`err instanceof Error ? err.message : 'Could not change password.'`) against what's thrown, and asserts it resolves to the server's message.

**Why not a component-render test of `ChangePasswordSheet` itself:** `apps/mobile/vitest.config.ts` deliberately scopes `include` to `src/**/*.test.ts` (not `.tsx`) with an explicit comment — "The mobile app has no React Native test runtime configured — we deliberately keep the include pattern at .ts (not .tsx) so component imports never pull RN/Expo modules into Vitest." There is no existing `.test.tsx` file anywhere in the repo, no `@testing-library/react-native` dependency, and no RN test renderer configured. Standing up that infrastructure is out of scope for a mechanical bug fix, so the test above reproduces the exact bug (verified red against the pre-fix code — see below) via the same expression the component uses, against the real thrown value, instead of via a render.

**Red-first verification:** before finalizing, I stashed the `api.ts` fix and re-ran the new test file — both the message test and the `instanceof` tests failed exactly as described (`'Could not change password.'` instead of the server's message; `ApiError is not a constructor`), confirming the test reproduces the reported bug. Then restored the fix and re-ran — green.

## Verification

- `cd apps/mobile && npx vitest run` — **95 files, 1222 passed, 3 skipped, 0 failed** (full mobile suite, not just the affected files, given `api.ts` is imported broadly).
- `cd apps/mobile && npx tsc --noEmit` — clean, no errors.
- Branched off fresh `origin/main` (383a10f9a), fast-forward-merged one intervening unrelated portal commit (458ed6f1b) before final verification — no conflicts, suite re-run green after merge.

### DIFF
--- a/apps/mobile/src/lib/errorReporting.test.ts
+++ b/apps/mobile/src/lib/errorReporting.test.ts
@@ -29,8 +29,11 @@ describe('reportInternalError', () => {
     expect(hint).toEqual({ tags: { area: 'ai-sessions-history' } });
   });
 
-  it('normalizes plain-object ApiError throws into a real Error keyed on area and status', () => {
-    // services/api.ts throws object literals, not Error instances.
+  it('normalizes plain-object throws into a real Error keyed on area and status', () => {
+    // services/api.ts itself now throws a real ApiError (#4747, see the test
+    // below) — this exercises the defensive fallback for anything ELSE that
+    // still throws a plain object shaped like the old ApiError interface (a
+    // stubbed test double, a third-party rejection).
     const apiError = { message: 'An error occurred', statusCode: 500 };
     reportInternalError(apiError, 'device-metrics');
 
@@ -41,6 +44,31 @@ describe('reportInternalError', () => {
     expect(hint).toEqual({
       tags: { area: 'device-metrics' },
       extra: { apiError },
+    });
+  });
+
+  it('attaches code/statusCode as extra for a real Error that carries them (e.g. ApiError)', () => {
+    // A real `ApiError` instance (#4747) is `instanceof Error`, so it takes the
+    // "pass through unchanged" branch above — but Sentry does not auto-serialize
+    // a thrown Error subclass's own `code`/`statusCode` properties. Without this,
+    // the HTTP status/code silently stopped reaching Sentry's tags/extra the
+    // moment ApiError became a real Error (it used to always take the
+    // `extra: { apiError: err }` branch). Simulated here without importing the
+    // real ApiError class, since this file mocks `../services/api` to keep
+    // expo-secure-store out of the node test runtime.
+    const err = Object.assign(new Error('Too many attempts'), {
+      name: 'ApiError',
+      code: 'rate_limited',
+      statusCode: 429,
+    });
+    reportInternalError(err, 'change-password');
+
+    expect(sentry.captureException).toHaveBeenCalledTimes(1);
+    const [captured, hint] = sentry.captureException.mock.calls[0];
+    expect(captured).toBe(err);
+    expect(hint).toEqual({
+      tags: { area: 'change-password' },
+      extra: { code: 'rate_limited', statusCode: 429 },
     });
   });
 
--- a/apps/mobile/src/lib/errorReporting.ts
+++ b/apps/mobile/src/lib/errorReporting.ts
@@ -31,8 +31,22 @@ export function reportInternalError(err: unknown, area: string): void {
           `${area} failed: ${apiErr?.statusCode ?? '?'} ${apiErr?.message ?? String(err)}`.trim(),
         );
 
+  // A real `ApiError` is `instanceof Error` and takes the branch above
+  // unchanged, but Sentry does not auto-serialize a thrown Error subclass's
+  // own `code`/`statusCode` properties (no extraErrorDataIntegration
+  // configured) — attach them explicitly so the HTTP status/code survives
+  // into the event instead of only living in the stack trace.
+  const apiExtra =
+    apiErr && (apiErr.code !== undefined || apiErr.statusCode !== undefined)
+      ? { code: apiErr.code, statusCode: apiErr.statusCode }
+      : null;
+
   Sentry.captureException(normalized, {
     tags: { area },
-    ...(err instanceof Error ? {} : { extra: { apiError: err } }),
+    ...(err instanceof Error
+      ? apiExtra
+        ? { extra: apiExtra }
+        : {}
+      : { extra: { apiError: err } }),
   });
 }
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter (parallel, sonnet)

**Findings:** 2 real findings raised, both addressed in d3b342806; 1 non-blocking nitpick noted and also cleaned up.

- pr-test-analyzer: the second `changePassword` test queued only one `fetchWithTimeout.mockResolvedValueOnce` before two sequential `expect(...).rejects` calls — the second call found the mock queue empty and failed on an unrelated `TypeError` (also `instanceof Error`), so that assertion was passing for the wrong reason without re-exercising the ApiError/429 path. Fixed by consolidating into one caught value asserted three ways (`apps/mobile/src/services/api.changePassword.test.ts`).
- silent-failure-hunter: since a real `ApiError` is now `instanceof Error`, it takes the pass-through branch in `errorReporting.reportInternalError` — which silently stopped attaching `code`/`statusCode` to the Sentry event (previously it always fell into the `extra: { apiError: err }` branch). Fixed by attaching `code`/`statusCode` explicitly whenever present, regardless of branch (`apps/mobile/src/lib/errorReporting.ts`), plus a new regression test.
- code-reviewer: no confidence≥80 issues; flagged a stale comment in `errorReporting.test.ts` claiming `services/api.ts` still throws plain object literals — updated.

**Tests:** Full mobile suite green: `cd apps/mobile && npx vitest run` — 95 files, 1223 passed, 3 skipped, 0 failed. `npx tsc --noEmit` clean. Red-first verified: the new `api.changePassword.test.ts` assertions were confirmed to fail against the pre-fix code (stashed `api.ts`, re-ran, restored) before finalizing.

PR #4723, opened 2026-09-03
W01: Server-side native FCM sender + dispatch wiring (#3639)

ToddHebebrand:
## Summary

Wave 1 (server-only) of feature #4717 "Android push (FCM) wiring", scoped by `docs/superpowers/plans/mobile/2026-09-02-android-push-fcm-wiring.md`.

- Add `apps/api/src/services/fcm.ts`: a native FCM sender mirroring `apns.ts`'s never-throws, structured-result contract (`isFcmConfigured`, `sendFcmNotification`, `FcmPayload`/`FcmResult`).
- Refactor `notifications.ts`'s `sendFCM` to delegate to `fcm.ts` — removes the duplicate local Firebase-admin init and adds dead-token purge parity with `sendAPNS` (previously `sendFCM` had zero test coverage and never purged a dead token).
- Wire the `fcm` branch of `expoPush.ts`'s `dispatchPushToTokens`: raw `provider: 'fcm'` tokens (native Android, tagged by `inferProvider`) now dispatch via `sendFcmNotification` instead of being logged-and-skipped, with `purgeFcmToken` mirroring `purgeApnsToken` for unregistered tokens.

This only activates for tokens already tagged `provider: 'fcm'` in the DB — legacy raw Android tokens (pre-dating the abandoned Expo-relay migration) start working immediately; no behavior change for `expo`/`apns` tokens. A fresh Android install doesn't register a native FCM token yet until Wave 2 (mobile) ships.

No migration, no RLS/tenancy surface — see the plan's Global Constraints section (`mobile_devices`/`push_notifications` are Shape 6, no `org_id`/`device_id` columns, no cascade/export list changes needed).

Closes #4718

## Test plan
- [x] `cd apps/api && npx vitest run src/services/fcm.test.ts` — 6/6 pass (new)
- [x] `cd apps/api && npx vitest run src/services/notifications.test.ts` — 7/7 pass (3 new `sendFCM` tests)
- [x] `cd apps/api && npx vitest run src/services/expoPush.test.ts` — 34/34 pass (3 new/replaced fcm-dispatch tests)
- [x] Full `apps/api` unit suite: 32876/32897 passed, 1796/1798 files passed — the only 4 failures are pre-existing `ECONNREFUSED :5432` integration tests (no local Postgres in this environment), unrelated to this change
- [x] `npx tsc --noEmit` in `apps/api` — clean
- [x] `npx eslint` on all changed files — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4717, opened 2026-09-03
Android push: FCM delivery + mobile token registration (#3639)


ToddHebebrand:
Tracking issue for **Android push: FCM delivery + mobile token registration (#3639)** (`android-push-fcm-wiring`).

- Plan: `docs/superpowers/plans/mobile/2026-09-02-android-push-fcm-wiring.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
Both waves merged: W01 #4718 → PR #4723 (server fcm.ts + expoPush dispatch + dead-token purge), W02 #4719 → PR #4780 (mobile single getDevicePushTokenAsync path, google-services.json build-time injection). Unreleased; ships next release.

Not verified / still needs Todd (ops, not code): create the Firebase Android app, obtain google-services.json and a service account, set FIREBASE_SERVICE_ACCOUNT in /opt/breeze/.env AND map it in the api service environment: block of /opt/breeze/docker-compose.yml on both droplets, inject google-services.json into the mobile build. No end-to-end push to a real Android device has been exercised.

Issue #3639, opened 2026-08-16
[Mobile] Android push is not wired — server skips raw FCM tokens, no Expo projectId


ToddHebebrand:
Split out of #1155 per the 2026-08-16 decision to submit iOS first.

**Gap (from `apps/mobile/STORE_SUBMISSION.md`):** the server-side push sender skips raw FCM tokens and the Expo config has no `projectId`, so Android devices register but never receive pushes. Native APNs is live in production on both regions; Android needs the FCM sender path scoped and built before a Play Store submission is worth starting.

Refs #1155, #611.

ToddHebebrand:
Plan drafted: [`docs/superpowers/plans/mobile/2026-09-02-android-push-fcm-wiring.md`](https://github.com/LanternOps/breeze/blob/work/plan-3639/docs/superpowers/plans/mobile/2026-09-02-android-push-fcm-wiring.md) on branch `work/plan-3639`.

**Design decision:** go native FCM (mirror iOS's native-APNs move) instead of restoring `extra.eas.projectId`/the Expo relay — reuses the `firebase-admin` infra that already exists (`notifications.ts`'s alert-push `sendFCM`), avoids reintroducing an EAS/Expo-account dependency iOS deliberately dropped, and closes the gap for both the ticket/approval push path (`expoPush.ts`) and any legacy raw FCM token already sitting in `mobile_devices`.

**Waves** (each independently shippable/PR-able):
1. **Server** — new `apps/api/src/services/fcm.ts`, mirroring `apns.ts`'s never-throws/structured-result contract; wire it into `expoPush.ts`'s `dispatchPushToTokens` fcm branch (currently just logs and skips); refactor `notifications.ts`'s `sendFCM` to delegate, adding dead-token purge parity with APNs. Ships real delivery for any raw FCM token already in the DB, no mobile change needed.
2. **Mobile** — collapse `registerForPushNotifications()`'s iOS/Android branches into one `getDevicePushTokenAsync()` call (no `projectId` needed on either platform), retire the now-unreachable `android_push_not_configured` copy, wire `google-services.json` into `app.json`.
3. **Ops (Todd)** — confirm/create the Firebase Android app for `com.breeze.rmm`, get a real `google-services.json` into the repo, confirm `FIREBASE_SERVICE_ACCOUNT` on both droplets, update `STORE_SUBMISSION.md` with an Android section.
[...]

PR #4780, opened 2026-09-03
W02: Android native FCM token registration + build-time google-services.json

### LINKED ISSUE #4719: W02: Mobile: single getDevicePushTokenAsync path, retire android_push_not_configured copy, google-services.json wired via build-time injection
Wave W02 of #4717 — Mobile: single getDevicePushTokenAsync path, retire android_push_not_configured copy, google-services.json wired via build-time injection
Branch: `feature/4717-android-push-fcm-wiring/wave-4719`

ToddHebebrand:
## Summary

Wave W02 of feature #4717 (origin #3639), sub-issue #4719. Mobile half of the
Android push fix — W01 (server, #4723, merged) added the native FCM sender;
this closes the client side.

- **Task 2.1** — `registerForPushNotifications()` now calls
  `Notifications.getDevicePushTokenAsync()` unconditionally on both platforms
  instead of branching iOS (native APNs) vs Android (Expo relay via
  `getExpoPushTokenAsync`, gated on an `extra.eas.projectId` that no longer
  exists in `app.json`). Android needs no Expo/EAS account any more, matching
  the native-APNs pattern iOS already uses. Dropped the now-unused
  `expo-constants` import.
- **Task 2.2** — retired the unreachable `android_push_not_configured` copy
  branch in `pushUnavailableCopy.ts`. Android's only remaining `unsupported`
  reason is the pre-existing `not_physical_device`, shared with iOS
  simulators.
- **Task 2.3** — wired `google-services.json` into the Android build via
  `app.json`'s `android.googleServicesFile`, resolved by Expo's config plugin
  only during `expo prebuild --platform android` (never during iOS prebuild
  or CI's vitest/tsc). The file itself is **never committed**:
  `scripts/write-google-services.mjs` writes it from a `GOOGLE_SERVICES_JSON`
  build-time secret (raw or base64 JSON, same tolerance as the server's
  `FIREBASE_SERVICE_ACCOUNT`), mirroring how `ci_post_clone.sh` writes
  `.env.sentry-build-plugin` from `SENTRY_AUTH_TOKEN` before an iOS build.
  Documented in `STORE_SUBMISSION.md`; `google-services.json.example` shows
  the expected shape.

Registering the actual Firebase Android app and setting the real
`GOOGLE_SERVICES_JSON`/`FIREBASE_SERVICE_ACCOUNT` secrets is tracked
separately as Wave 3 (operational, Todd) — this PR only wires the client code
and build tooling, per the plan doc
(`docs/superpowers/plans/mobile/2026-09-02-android-push-fcm-wiring.md`).

## Test plan

- [x] `pnpm --filter=breeze-mobile test` — 94 files, 1218 passed / 3 skipped
- [x] `pnpm --filter=breeze-mobile typecheck` — clean
- [x] `node scripts/write-google-services.mjs` manually verified: fails loudly
      with no `GOOGLE_SERVICES_JSON`, writes the file correctly for both raw
      JSON and base64 input
- [x] `git grep -n googleServicesFile apps/mobile` shows only the one
      `app.json` reference (plus doc mentions)
- [x] `grep -rn android_push_not_configured apps/mobile` — zero remaining
      references

Closes #4719

https://claude.ai/code/session_01Y5fsHMpNYfaN71Ba82fPia


ToddHebebrand:
## Review summary

Ran three parallel Sonnet reviews (code-reviewer, silent-failure-hunter, pr-test-analyzer).

**code-reviewer**: no confirmed findings. Diff is small, self-contained, comments accurate; confirmed the `getDevicePushTokenAsync()` collapse is correct (genuinely platform-agnostic in the Expo API), no stale `android_push_not_configured`/`getExpoPushTokenAsync` references remain, `app.json`'s `googleServicesFile` addition is inert for iOS/CI as claimed, and the `.gitignore` negation for the `.example` file works (verified with `git check-ignore -v`).

**pr-test-analyzer**: coverage on the two runtime files is solid and behavioral (the "Android uses NATIVE FCM token, never Expo relay" test explicitly asserts `getExpoPushTokenAsync` is NOT called). One confirmed gap: `scripts/write-google-services.mjs`'s JSON/base64 parsing logic was inlined and untested, breaking from this repo's own `apiUrl.js`+`apiUrl.test.ts` / `preflight.mjs` precedent (decision logic lives in a tested `src/config/*.js` module; the script stays a thin untested shell).

**silent-failure-hunter**: two HIGH findings plus a MEDIUM that overlapped with the test-analyzer's gap.

**Fixed inline** (pushed in 881b563f9):
- Extracted `parseGoogleServicesJson()` into `src/config/googleServicesJson.js` with its own test file (9 cases), following the `apiUrl.js` precedent.
- Added real shape validation (`project_info.project_id`, non-empty `client[]`, Android package name match) instead of a bare `JSON.parse` check — closes the "someone pastes `FIREBASE_SERVICE_ACCOUNT` by mistake" silent-acceptance path the reviewer flagged, and the corresponding "unconditional overwrite with unvalidated content" risk.
- Tightened the script's header comment, which overclaimed protection against a stale on-disk file when the script is never invoked at all.

**Not fixed — flagged as known limitations, not blocking this wave:**
1. *Generic Android failure copy*: the broad `try/catch` in `registerForPushNotifications()` now also catches new Android failure classes (missing Play services, Firebase-init errors), and `pushUnavailableCopy`'s `'failed'` branch only special-cases `permission_denied` — every other reason renders "sign in again," which won't fix e.g. missing Play services. This is the same catch-all pattern iOS has always used, extended to a platform that previously never reached it. Differentiating failure copy per reason is real product/UX work beyond this wave's stated scope (single-path token registration + retire dead copy + wire the build config) — filing as a follow-up rather than expanding scope here.
2. *False-positive "delivered" window*: if an Android build ships with a real `google-services.json` before Wave 3 (Firebase project + droplet `FIREBASE_SERVICE_ACCOUNT` alignment) lands, `registerForPushNotifications()` could return `ok` while sends actually fail server-side with a credential mismatch. This is inherent to the plan's wave sequencing (Wave 3 is explicitly gating, documented in `STORE_SUBMISSION.md`) and not something this PR's code can fix — no Android build can even `expo prebuild` today since `google-services.json` doesn't exist until Wave 3's real secret is set.

## Test/typecheck status

- `pnpm --filter=breeze-mobile test`: 95 files, 1227 passed / 3 skipped
- `pnpm --filter=breeze-mobile typecheck`: clean
- Manually verified `write-google-services.mjs` fails loudly for: unset secret, wrong-shape JSON (FIREBASE_SERVICE_ACCOUNT-style), wrong Android package; and succeeds for the `.example` file's shape (raw and base64 forms).

Not merging — leaving this for the wave's owner to close out.

PR #4781, opened 2026-09-03
feat(agent): extract custom-field markers from raw script stdout (#2698)

### LINKED ISSUE #4681: W03: Go agent: structured channel, extract markers pre-sanitizer
Wave W03 of #4678 — Go agent: structured channel, extract markers pre-sanitizer
Branch: `feature/4678-script-custom-field-writeback/wave-4681`

ToddHebebrand:
## Summary

W03 of #4678 (Script custom-field write-back, origin #2698) — the Go agent's structured channel.

- `agent/internal/executor/customfields.go`: `ExtractCustomFields(stdout string) (map[string]any, string)` pulls `::breeze:custom-fields::` marker lines out of **raw** stdout — before `SanitizeOutput`, which rewrites `token=`/`secret=`-shaped substrings and would otherwise corrupt a marker's JSON past recovery. Caps: 20 marker lines, 50 distinct keys, 8KB payload per marker. Unparseable marker lines are left in stdout for the operator to see.
- `agent/internal/heartbeat/handlers_script.go`: both the local-executor build site (`handleScriptInner`) and the `runAs:user` helper-IPC path (`executeViaUserHelper`) now extract markers, strip them from the persisted stdout, and — when fields were found — set `tools.CommandResult.Result` to a versioned envelope: `{customFieldWrites: {schemaVersion: 1, fields: {...}}}`.
- `agent/internal/heartbeat/heartbeat.go` (`toWSCommandResult`): an explicitly-set `Result` now wins over the stdout→JSON reparse, so the envelope survives the WebSocket leg without disturbing the existing reparse that discovery/backup/snmp/monitor results still depend on. The `Error`-suppresses-reparse behavior is unchanged.

Byte-identical marker constant to the API side (`CUSTOM_FIELD_MARKER` in `apps/api/src/services/customFields/scriptWriteMarkers.ts`, shipped in Wave 1 / #4679).

**This PR is inert without Wave 1 already deployed** — the API only reads `customFieldWrites` from an agent version that sends it. Mixed-version fleets are safe in both directions:
- A pre-Wave-3 agent's markers still work via Channel A (stdout scanning, Wave 1).
- A Wave-3 agent talking to a pre-Wave-1 server just has its envelope ignored (unknown `result` shape).

## Test plan

- [x] `agent/internal/executor/customfields_test.go` — table-driven `ExtractCustomFields` tests (no marker, single marker stripped, later-marker-wins merge, secret-shaped value survives because extraction runs before the sanitizer, unparseable marker left in stdout, CRLF line endings) + a cap test (25 marker lines → 1 merged key).
- [x] `agent/internal/heartbeat/handlers_script_test.go` — `handleScript` end-to-end: envelope emitted + marker stripped from stdout; no envelope when no marker; secret-shaped marker value survives `SanitizeOutput`; the same via `executeViaUserHelper` (`runAs:user` path).
- [x] `agent/internal/heartbeat/result_mapping_test.go` — `toWSCommandResult`: an explicit `Result` wins over the stdout reparse even when stdout isn't valid JSON; the pre-existing reparse-when-`Result`-is-nil path is unchanged.
- [x] `gofmt -l` clean on all changed files.
- [x] `go build ./...` — clean.
- [x] `go test ./internal/executor/... ./internal/heartbeat/...` — full packages green (`CGO_ENABLED=0`; see note below), including all pre-existing tests.

**Note on `-race`:** this dev host's cgo toolchain is independently broken (`go-m1cpu`'s cgo init segfaults on *any* package, unrelated to this change — reproduced on `internal/privilege` with zero local edits, and reproduces on both the installed go1.27.0 and an explicitly-pinned go1.26.6 toolchain). `-race` requires cgo, so it cannot run on this host at all right now. Verified instead with `CGO_ENABLED=0 go build ./...` and `CGO_ENABLED=0 go test ./internal/executor/... ./internal/heartbeat/...` (all green). CI runs `go test -race` in a clean environment and will provide the race-detector signal this box can't right now.

Closes #4681

https://claude.ai/code/session_01AEjfwFKmH2HGwCFs15GnyG

### DIFF
--- a/agent/internal/executor/customfields.go
+++ b/agent/internal/executor/customfields.go
@@ -21,8 +21,16 @@ const (
 // deliberately before SanitizeOutput, which rewrites `token=`/`secret=`-shaped
 // substrings and would otherwise corrupt a marker beyond JSON.Unmarshal's
 // reach. Returns the merged map (later lines win) and stdout with the consumed
-// lines removed. Unparseable marker lines are LEFT IN stdout so the operator
-// can see what the script actually printed.
+// lines removed.
+//
+// All three caps (marker-line count, per-marker JSON size, distinct key
+// count) reject at LINE granularity and leave the rejected line in the
+// returned stdout, so a technician reading the persisted output can always
+// see which lines were not applied and why (a syntactically valid marker line
+// is never partially merged then silently discarded key-by-key — that would
+// leave no trace anywhere once the line itself is consumed). This also keeps
+// "later lines win" true without exception: a later line either fully applies
+// or is fully rejected, never partially applies past the key cap.
 func ExtractCustomFields(stdout string) (map[string]any, string) {
 	if !strings.Contains(stdout, CustomFieldMarker) {
 		return nil, stdout
@@ -52,16 +60,26 @@ func ExtractCustomFields(stdout string) (map[string]any, string) {
 			continue
 		}
 
+		// Reject the whole line, deterministically, if merging it would push
+		// the distinct-key count past the cap — rather than partially merging
+		// some of its keys in Go's randomized map-iteration order and dropping
+		// the rest with no trace.
+		newKeys := 0
+		for k := range parsed {
+			if _, exists := fields[k]; !exists {
+				newKeys++
+			}
+		}
+		if len(fields)+newKeys > maxCustomFieldKeys {
+			kept = append(kept, line)
+			continue
+		}
+
 		markerLines++
 		if fields == nil {
 			fields = make(map[string]any, len(parsed))
 		}
 		for k, v := range parsed {
-			if len(fields) >= maxCustomFieldKeys {
-				if _, exists := fields[k]; !exists {
-					continue
-				}
-			}
 			fields[k] = v
 		}
 		// consumed: not appended to kept
--- a/agent/internal/executor/customfields.go
+++ b/agent/internal/executor/customfields.go
@@ -1,0 +1,71 @@
+package executor
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+// CustomFieldMarker is the stdout sentinel a script uses to write its own
+// device's custom fields (#2698). Must stay byte-identical to
+// CUSTOM_FIELD_MARKER in
+// apps/api/src/services/customFields/scriptWriteMarkers.ts.
+const CustomFieldMarker = "::breeze:custom-fields::"
+
+const (
+	maxCustomFieldMarkerLines = 20
+	maxCustomFieldKeys        = 50
+	maxCustomFieldJSONBytes   = 8192
+)
+
+// ExtractCustomFields pulls every well-formed marker line out of RAW stdout —
+// deliberately before SanitizeOutput, which rewrites `token=`/`secret=`-shaped
+// substrings and would otherwise corrupt a marker beyond JSON.Unmarshal's
+// reach. Returns the merged map (later lines win) and stdout with the consumed
+// lines removed. Unparseable marker lines are LEFT IN stdout so the operator
+// can see what the script actually printed.
+func ExtractCustomFields(stdout string) (map[string]any, string) {
+	if !strings.Contains(stdout, CustomFieldMarker) {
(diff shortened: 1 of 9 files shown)


ToddHebebrand:
## Review summary

Ran three parallel Sonnet review passes (code-reviewer, silent-failure-hunter, pr-test-analyzer) against the initial implementation. Two **critical**, confirmed findings and several important gaps came back — all fixed in the follow-up commits, verified locally and re-confirmed green in CI.

### Critical (fixed)
1. **`runAs:user` path extracted markers from already-sanitized stdout, not raw stdout.** `userhelper/client.go`'s `executeScript` (the process that actually runs a `runAs:user` script) calls `SanitizeOutput` on its own stdout before handing it back over IPC — the main agent was then calling `ExtractCustomFields` on that already-sanitized text, silently defeating the "extract before the sanitizer corrupts a secret-shaped marker" point of this whole wave for every helper-dispatched script. **Fix:** moved extraction into `executeScript` itself, before its own `SanitizeOutput` call; it now forwards a pre-built `customFieldWrites` envelope over IPC, and `executeViaUserHelper` just relays it. Verified with a new test (`TestExecuteScriptExtractsCustomFieldsBeforeSanitizeOutput`) that a secret-shaped marker key genuinely survives at that layer.
2. **A delivered `secretEnv` value could leak into a device custom field.** `ExtractCustomFields` deliberately runs before both `SanitizeOutput` and the exact-value secret redactor, and `Result` was excluded from redaction by design — correct for the Wave 1 "not a secrets channel" framing (stdout visibility), but that doesn't cover a script echoing a *delivered* secret env value into a marker, which would then land verbatim in `devices.customFields` — visible to any org user with device-read access. **Fix:** added `redactCustomFieldValues`, applied in `handleScript`'s existing redact block, so extracted field values get the same exact-value scrub as Stdout/Stderr/Error. Confirmed `secretEnv` never reaches the user-helper path at all (refused upstream, `#3409`), so this fix correctly and exclusively covers the local-executor path where the risk is real. New regression test: `TestHandleScriptRedactsSecretEnvValueFromCustomFieldWrites`.

### Important (fixed)
3. The 50-distinct-key cap in `ExtractCustomFields` used to partially merge a marker line via Go's randomized map iteration order and silently drop keys past the cap with **zero trace** anywhere (not stdout, not the Result envelope, not a log). Now rejects the whole line deterministically, matching the other two caps' behavior (line-count, payload-size) — reject at line granularity, leave the line visible in stdout.
4. Added a `log.Warn` at both extraction call sites when a marker line survives extraction (cap-rejected or unparseable) — previously undiagnosable without reading source to know the caps exist.
5. `TestExtractCustomFieldsCaps` was vacuous (25 *identical-key* lines can't distinguish "cap works" from "cap doesn't exist"). Rewrote with distinct keys, and added dedicated tests for the key cap (rejection + "later lines can still update existing keys past the cap") and the byte-size cap.
6. Added `TestToWSCommandResultResultWinsEvenWithErrorSet` — the `Error != "" && Result != nil` combination this PR's `toWSCommandResult` change made possible for the first time. Confirmed intentional (matches the server's `commandResultHandlers.ts`, which applies markers "outside the exit-code branch" on purpose) but was previously untested either way.
7. Updated the stale `tools.CommandResult.Result` doc comment.

### Test / CI status
- `CGO_ENABLED=0 go test ./internal/executor/... ./internal/heartbeat/... ./internal/userhelper/... ./internal/remote/tools/...` — all green, including every new/updated test above.
- `-race` unusable on the dev-agent's local host (pre-existing, unrelated `go-m1cpu` cgo segfault reproduced on untouched packages too) — CI's **Test Agent (race)** job is the real signal here and is green on the current head.
- Full `go test ./...`: green except the pre-existing, unrelated `internal/monitoring` host-sandboxing failure (present before this PR too, package untouched by this diff).
- CI: all 65 required/non-required checks pass (1 intentional skip: `Main Red Alert`, main-branch-only).

No further findings outstanding. Not merging or closing the linked issue — leaving that to a human reviewer per policy.

PR #4782, opened 2026-09-03
fix(api/auth): rate-limit current-password guesses on /auth/change-password (#4746)

### LINKED ISSUE #4746: [API][Auth] POST /auth/change-password has no rate limit on current-password guesses (siblings do)
**Where:** `apps/api/src/routes/auth/password.ts:278` — `POST /auth/change-password` inlines `verifyPassword(user.passwordHash, currentPassword)` (line ~307) with no `rateLimiter` call on the handler path.

**Contrast:** every sibling that verifies a body-supplied password is throttled — `accountDeletion.ts:234` (`account-deletion:<client>:<user>`, 5 / 5 min) and the MFA routes via `requireCurrentPasswordStepUp(c, userId, currentPassword, 'mfa:pwd', …)` (`mfa.ts:821`, `:1242`). Only the forgot-password flow in `password.ts` is rate-limited (`forgot:<client>`).

**Impact:** a stolen access token can brute-force the current password unthrottled (one argon2 verify per request) and, on success, set a new password — permanent account takeover with no lockout and no step-up cost.

**Fix:** replace the inline verify with `requireCurrentPasswordStepUp(c, auth.user.id, currentPassword, 'pwd:change', …)` so the route inherits the shared limiter and the 400 + stable-code contract from #4651 / #4660. Add a route test that proves the Nth wrong guess inside the window returns 429 before argon2 runs.

Surfaced during the #4660 fix (PR #4739); out of that PR's scope. Verified against origin/main 8ea4aaf993.

ToddHebebrand:
Closes #4746

## The gap

`POST /auth/change-password` (`apps/api/src/routes/auth/password.ts`) inlined a bare `verifyPassword(user.passwordHash, currentPassword)` with **no limiter anywhere on the handler path**. It was the only route in the codebase that validates a body-supplied password without one:

| Route | Throttle |
|---|---|
| `POST /auth/account-deletion` | own `rateLimiter(redis, 'account-deletion:<client>:<user>', 5, 300)` |
| MFA disable / recovery-code regen, passkey delete, phone routes | `requireCurrentPasswordStepUp(…, 'mfa:pwd' \| 'passkey:pwd', …)` |
| `POST /auth/forgot-password` | `forgot:<client>` |
| **`POST /auth/change-password`** | **none** |

A stolen access token could brute-force the current password at one argon2 verify per request and, on a hit, set a new one — account takeover with no lockout and no step-up cost.

## The fix

Route the check through the shared `requireCurrentPasswordStepUp` helper with its own `pwd:change` bucket. The limiter (5 attempts / 5 min, per user) now runs **before** the hash lookup and the verify, so a throttled guess costs an attacker a Redis round trip instead of an argon2 evaluation. A dedicated prefix means change-password guesses neither spend nor hide behind the `mfa:pwd` budget the factor routes share.

## Response contract is preserved verbatim (#4660 / #4739)

The helper hard-coded an opaque `'Invalid credentials'` for both its rejections. That is right for a step-up on a sensitive operation, but wrong on this route: the request is already authenticated **as** the account being changed, so opacity buys nothing, and `ProfilePage.tsx` renders `message` verbatim — "Invalid credentials" on a change-password form reads as a dead session rather than a typo, which is the exact confusion #4739 just removed.

Two **opt-in** overrides (`invalidMessage`, `noPasswordMessage`) were added to the helper. Both default to today's strings, so all eight existing call sites are byte-for-byte unchanged.

| Case | Before | After |
|---|---|---|
| Wrong current password | 400 `{error, message: 'Current password is incorrect', code: 'invalid_credentials'}` | identical |
| Passwordless account | 400 `{error, message: 'Password authentication is not available for this account'}` | same, **plus** the stable `code: 'invalid_credentials'` (additive) |
| 6th guess in 5 min | *(none — 400/200 as if untracked)* | **429** `{error, message, retryAfter}` |

## Two things worth a reviewer's attention

1. **Fail-closed on Redis down.** With Redis unavailable the route now answers 503 instead of proceeding unthrottled. That is what every sibling step-up already does, and it is the correct direction for a credential check, but it is a behaviour change.
2. **`message` added to the step-up 429 body** (both step-up helpers, kept symmetric — the TOTP one documents itself as mirroring the password one). `ProfilePage.tsx` reads `data.message` and falls back to a generic "Failed to change password" when it is absent, so a 429 carrying only `error` would hide the throttle from the very user who needs to stop retrying. Purely additive; the one existing assertion on that body uses `objectContaining`.

Scope held to this route. The mobile client is untouched.

## Tests (red first)

Added to `apps/api/src/routes/auth/password.test.ts`, all three confirmed failing against the unpatched route before the fix (`rateLimiter` called 0 times; the over-limit guess answered 200):

- the over-limit attempt is **429 before argon2 runs** — asserts `verifyPassword` was never called and nothing was rotated
- guesses are metered in a per-user `pwd:change:<user>` bucket at 5 / 5 min
- a correct password **after two wrong guesses still succeeds** while under the limit — the throttle must not lock a legitimate user out of their own password change for two typos

The two pre-existing #4660 contract tests (wrong password → 400 + code; passwordless → its own 400 body) pass unchanged, which is the evidence the response shape survived the move.

**Verification:** `vitest run src/routes/auth/ src/routes/auth.test.ts src/routes/auth.passkeys.test.ts src/routes/authenticator.test.ts src/routes/users.test.ts src/routes/pam.test.ts src/routes/approvals.test.ts src/openapi.scriptAdmission.test.ts` → **28 files / 962 tests passed**. `tsc --noEmit` clean. `eslint` on the four changed files clean.

### DIFF
--- a/apps/api/src/openapi.ts
+++ b/apps/api/src/openapi.ts
@@ -1262,7 +1262,10 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
           // 400s for a passwordless account and a too-weak new password. 401
           // is reserved for a dead bearer.
           '400': { $ref: '#/components/responses/BadRequest' },
-          '401': { $ref: '#/components/responses/Unauthorized' }
+          '401': { $ref: '#/components/responses/Unauthorized' },
+          // #4746: current-password guesses are metered per user (5 / 5 min),
+          // the same shared step-up limiter the MFA factor routes use.
+          '429': { $ref: '#/components/responses/TooManyRequests' }
         }
       }
     },
--- a/apps/api/src/routes/auth/helpers.currentPasswordStepUp.test.ts
+++ b/apps/api/src/routes/auth/helpers.currentPasswordStepUp.test.ts
@@ -1,0 +1,244 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+// #4746: `requireCurrentPasswordStepUp` is the shared gate for every route that
+// re-verifies a body-supplied password (MFA disable, recovery-code regen,
+// passkey delete, phone enrol/remove, email change, approvals + PAM step-up,
+// and now /auth/change-password). It had no test file of its own — every one of
+// those callers mocks it wholesale — so its DEFAULT rejection contract was
+// asserted nowhere in the repo. That matters now that the helper takes opt-in
+// message overrides: "the eight existing call sites are unchanged" is the claim
+// the whole design rests on, and it needs a test that would notice if a future
+// edit changed what an omitted option does.
+//
+// Mock declarations must precede the import of the unit under test; vi.mock
+// factories are hoisted above module-scope consts, so the shared references go
+// through vi.hoisted (which is hoisted too). Mirrors helpers.mfaStepUp.test.ts.
+const { selectLimit, db, getRedis, rateLimiter, verifyPassword } = vi.hoisted(() => {
+  const selectLimit = vi.fn();
+  const db = {
+    // db.select(...).from(...).where(...).limit(...) → the mocked user row.
+    select: vi.fn(() => ({
+      from: vi.fn(() => ({
+        where: vi.fn(() => ({
+          limit: selectLimit,
+        })),
+      })),
+    })),
+  };
+  return {
+    selectLimit,
+    db,
+    getRedis: vi.fn(),
+    rateLimiter: vi.fn(),
+    verifyPassword: vi.fn(),
+  };
+});
+
+vi.mock('../../db', () => ({
+  db,
+  withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
+}));
+
+vi.mock('../../db/schema', () => ({
+  users: { id: 'id', passwordHash: 'password_hash' },
+  partnerUsers: {},
+  organizationUsers: {},
+  organizations: {},
+  userPasskeys: {},
+}));
+
+vi.mock('../../services', () => ({
+  verifyToken: vi.fn(),
+  isUserTokenRevoked: vi.fn(),
+  revokeRefreshTokenJti: vi.fn(),
+  getTrustedClientIp: vi.fn(() => 'unknown'),
+  getUserEpochs: vi.fn(),
+  getRedis,
+  rateLimiter,
+  verifyPassword,
+}));
+
+vi.mock('../../services/mfa', () => ({ consumeMFAToken: vi.fn() }));
+vi.mock('../../services/mfaSecretCrypto', () => ({
+  decryptMfaTotpSecret: vi.fn(),
+  decryptMfaTotpSecretForMigration: vi.fn(),
+  encryptMfaTotpSecret: vi.fn(),
+}));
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** 4 lenses against `35e70a596` — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, and a security pass (auth / rate-limiting lens).

**Findings:** 4 raised → 3 addressed in `ad44e600a`; 1 filed as #4797; 0 outstanding.

- **HIGH — Redis-down 503 lost its signal on the web** (silent-failure-hunter). `requireCurrentPasswordStepUp` returned `{ error }` with no `message`, and `ProfilePage.tsx:444` reads `errorData.message ?? <generic>` *without* the `?? errorData.error` fallback its sibling handlers in the same file use — so an outage rendered as "Failed to change password", indistinguishable from a typo. A genuine regression from this PR (the route had no Redis dependency before it), and the same reasoning I'd already applied to the 429 one branch below. **Fixed** — both step-up helpers now populate `message` on 503 as well as 429.
- **MEDIUM — the new 503 branch was untested** (pr-test-analyzer + silent-failure-hunter). `getRedis` is mocked truthy for the whole file. **Fixed** — added a case asserting 503, both body keys, and that it fails *closed* (no verify, no rotation).
- **MEDIUM — the helper's default contract was asserted nowhere in the repo** (pr-test-analyzer). `requireCurrentPasswordStepUp` gates nine call sites and had no test file; every caller mocks it wholesale. That default is precisely what "the eight existing call sites are unchanged" rests on. **Fixed** — added `helpers.currentPasswordStepUp.test.ts` (11 cases) pinning the opaque 401 default, the passwordless branch staying indistinguishable from a wrong password, and `noPasswordMessage` falling back to `invalidMessage` so opting into a specific wrong-password string can't silently make the passwordless branch distinguishable. `helpers.mfaStepUp.test.ts` asserted the 503 body by exact equality and was updated.
- **HIGH but out of scope — `POST /portal/profile/password` has the same unthrottled brute force**, *and* 401s a wrong guess (which logs the portal user out via `api.ts:210` → `clearAuth()`). Verified both halves end-to-end. **Not folded in** — different route, different auth surface, different table (`portal_users`, so this helper isn't reusable as-is), and the fix needs a design call rather than a copy-paste. Filed as **#4797**.

`code-reviewer` returned **0 findings**, having independently enumerated all 9 other call sites of the helper and confirmed none pass the new options, so their bodies are byte-identical.

Not acted on, by design: no `Retry-After` header on the 429 (pre-existing in the shared helper, both reviewers agree it's not a regression), and four other 429 sites in the repo that still send `error` only (pre-existing, outside this issue).

**Tests:** `vitest run src/routes/auth/ src/routes/auth.test.ts src/routes/auth.passkeys.test.ts src/routes/authenticator.test.ts src/routes/users.test.ts src/routes/pam.test.ts src/routes/approvals.test.ts` → **28 files / 973 tests passed**. The three original #4746 tests were confirmed failing before the fix (limiter called 0 times; the over-limit guess answered 200). `tsc --noEmit` clean; `eslint` clean on all changed files.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4746 — those are yours.

Issue #4655, opened 2026-09-03
W06: device coverage lookup — contractLinesCoveringDevice, GET /devices/:id/billing, device detail Billing panel, coverage-notice deep links to the devices list (feeds #4606)


ToddHebebrand:
Wave W06 of #3205 — W06: device coverage lookup — contractLinesCoveringDevice, GET /devices/:id/billing, device detail Billing panel, coverage-notice deep links to the devices list (feeds #4606)
Branch: `feature/3205-device-coverage/wave-4655`

ToddHebebrand:
Completed via PR #4731

Merged on green (66/67 checks pass, 1 skipping) with --squash --admin per the run authorization. Opus final review: 0 Critical / 2 Important / 9 Minor, one fix wave; CodeQL js/xss-through-dom on the deep-link href fixed with a uuid guard + encodeURIComponent in a store-free module. Follow-up #4732 filed for the pre-existing filter-preview degrade.

Issue #3205, opened 2026-08-07
Bill contracts by device role or device group


ToddHebebrand:
## Description

Contracts cannot bill by device type. An MSP that charges a different rate per workstation, server, and network device has to maintain the quantities by hand.

## Current state

`contract_lines.lineType` is `flat | per_device | per_seat | manual` (`apps/api/src/db/schema/contracts.ts`). Automatic device counting exists — `countContractDevices(orgId, siteId)` in `apps/api/src/services/contractQuantities.ts` — but **the only axis is org, optionally narrowed to one site**. No role, no group, no OS.

Billing math lives in `contractService.ts::generateDueInvoice` (line 319); the quantity switch is exhaustive with a `const _exhaustive: never`, so adding a line type is a deliberate compile error in both that switch and `resolveLineQty`. The sweep is `apps/api/src/jobs/contractWorker.ts::runContractBillingSweep` (BullMQ cron `0 5 * * *`).

**Device type is already modeled cleanly.** `devices.deviceRole` varchar(30) default `'unknown'`, plus `deviceRoleSource` (`auto` | `manual`). Values are an SSOT tuple `DEVICE_ROLES` in `packages/shared/src/validators/index.ts:36`: `workstation, server, printer, router, switch, firewall, access_point, phone, iot, camera, nas, unknown`. Agent-classified (`agent/internal/collectors/classify.go`), operator-overridable (`routes/devices/core.ts:1236`), and already exposed as a filter field. It is **not** inferred from OS — `osType` and `isVirtual` are deliberately orthogonal axes.

Device groups: `device_groups` (static | dynamic, `rules`/`filterConditions` jsonb) + `device_group_memberships`.

## Proposed Fix — two new line types, columns only, no new tables

1. **Migration:** extend the line-type enum with `per_device_role` and `per_device_group`; `ALTER TABLE contract_lines ADD COLUMN device_role varchar(30)`, `ADD COLUMN device_group_id uuid` (FK ON DELETE SET NULL, mirroring the existing SQL-only `catalog_item_id` / `site_id` pattern that avoids import cycles). Add `CREATE UNIQUE INDEX device_groups_id_org_uq ON device_groups(id, org_id)` if a dual-axis FK is wanted — it does not exist today.
[...]

PR #4783, opened 2026-09-03
fix(web): fail closed on a failed advanced-filter preview (#4732)

### LINKED ISSUE #4732: Devices list: a pinned orgId the caller cannot access degrades to a silently unfiltered list
Found by the W06 (#3205) final review, pre-existing behaviour outside that wave's scope.

`apps/web/src/hooks/useAdvancedFilterIds.ts` (the `/filters/preview` caller): when the preview request fails — including the 403 the API returns for a pinned `orgId` the caller cannot access (`routes/filters.ts` `ensureOrgAccess`) — the hook leaves the id set empty/unset and the devices list renders **unfiltered** with no error surfaced. A pasted coverage deep link (`#orgId=<uuid>&filtersV2=…`) that points at an inaccessible org therefore shows the operator the wrong, wider list instead of an error.

Suggested fix: on preview failure set the ids to an empty set (`else setIds(new Set())`) and expose an error flag that the filter bar renders (toast or inline), so a failed filter never widens the result.

Refs #3205, #4655 (W06 final review, Minor 5).

ToddHebebrand:
## Summary

`useAdvancedFilterIds` (the `/filters/preview` caller behind the Devices list's advanced filter) silently degraded a failed preview request — including the 403 the API returns for a pinned `orgId` the caller cannot access (`routes/filters.ts` `ensureOrgAccess`) — to `ids: null`, which every consumer treats as "no filter active." The devices list therefore rendered the full, unfiltered fleet with no indication anything went wrong. A pasted coverage deep link (`#orgId=<uuid>&filtersV2=…`) pointing at an inaccessible org showed the operator the wrong, wider list instead of an error.

## Fix

- `useAdvancedFilterIds`: on a non-ok response (other than 401 — the existing `fetchWithAuth` session-expiry redirect already owns that failure) or a thrown fetch, set `ids` to an **empty set** (never `null`) and expose a new `error: boolean` on the return. Because every consumer already treats `ids !== null` as "narrow to this set," an empty-but-not-null set makes the list and grid render **zero rows** instead of everything, with no changes needed to the filtering logic itself.
- `DevicesPage`: toasts the failure (`devicesPage.toasts.advancedFilterFailed`) on the false→true transition — this covers grid view, which has no filter toolbar to show an inline state.
- `DeviceList`: renders an inline error pill (`deviceList.advancedFilterFailed`, `data-testid="device-filter-error"`) in place of the "Advanced filter active" pill when `serverFilterError` is set — covers list view and gives a directly testable zero-rows assertion.
- New locale keys added to all 8 locales (`en`, `de-DE`, `es-419`, `fr-CA`, `fr-FR`, `it-IT`, `pt-BR`, `tr-TR`) to satisfy `localeParity.test.ts`.

## Tests

- `useAdvancedFilterIds.test.ts`: new cases for 403, 500, network failure (all fail closed: empty set + `error: true`), a 401 exemption (no `error`, redirect owns it), and error-clears-on-retry. Updated the old "fails open" network-failure test to assert the correct fail-closed behavior.
- `DeviceList.test.tsx`: new case asserting a failed filter (`serverFilterError`) renders the inline error pill and zero rows, alongside the existing serverFilterIds coverage.
- `DevicesPage.test.tsx` / `DevicesPage.orgScope.test.tsx`: unaffected, still green (hook is mocked there).

## Test plan

- [x] `cd apps/web && npx vitest run src/hooks/useAdvancedFilterIds.test.ts` — 10 passed
- [x] `cd apps/web && npx vitest run src/components/devices/DeviceList.test.tsx` — 99 passed
- [x] `cd apps/web && npx vitest run src/components/devices/DevicesPage.test.tsx src/components/devices/DevicesPage.orgScope.test.tsx src/lib/i18n/localeParity.test.ts` — 160 passed
- [x] `cd apps/web && npx vitest run src/lib/__tests__/no-silent-mutations.test.ts` — 130 passed (unaffected)
- [x] `cd apps/web && npx tsc --noEmit` — clean
- [x] `cd apps/web && npx eslint <touched files>` — clean

Closes #4732

Refs #3205, #4655 (W06 final review, Minor 5).

### DIFF
--- a/apps/web/src/components/devices/DeviceList.test.tsx
+++ b/apps/web/src/components/devices/DeviceList.test.tsx
@@ -352,6 +352,32 @@ describe('DeviceList — advanced filter via serverFilterIds prop (uncapped id s
     expect(screen.getByText('host-bb')).toBeTruthy();
     expect(screen.queryByText(/Advanced filter active/i)).toBeNull();
   });
+
+  // #4732: a 403 on a pinned orgId the caller can't access (or any other
+  // preview failure) must render as an explicit error + zero rows, never as
+  // a silently unfiltered list. useAdvancedFilterIds fails CLOSED — an EMPTY
+  // set, not null — so DeviceList's ordinary `serverFilterIds` filtering
+  // already hides every device; `serverFilterError` only drives the message
+  // that explains why the list is empty instead of leaving that unexplained.
+  it('shows an inline error and zero rows when the filter preview failed (serverFilterError)', () => {
+    const a: Device = { ...baseDevice, id: 'bbbbbbb1-0000-0000-0000-000000000000', hostname: 'host-cc' };
+    const b: Device = { ...baseDevice, id: 'bbbbbbb2-0000-0000-0000-000000000000', hostname: 'host-dd' };
+
+    render(
+      <DeviceList
+        devices={[a, b]}
+        serverFilterIds={new Set()}
+        serverFilterError={true}
+      />
+    );
+
+    expect(screen.queryByText('host-cc')).toBeNull();
+    expect(screen.queryByText('host-dd')).toBeNull();
+    expect(screen.getByTestId('device-filter-error')).toBeTruthy();
+    expect(screen.getByText('0 of 2 devices')).toBeTruthy();
+    // The success pill must not also render alongside the error pill.
+    expect(screen.queryByText(/^Advanced filter active$/)).toBeNull();
+  });
 });
 
 describe('DeviceList — sortable columns (every column sorts on header click)', () => {
--- a/apps/web/src/components/devices/DeviceList.tsx
+++ b/apps/web/src/components/devices/DeviceList.tsx
@@ -298,6 +298,12 @@ type DeviceListProps = {
   // grid views filter against the same complete, uncapped id set.
   serverFilterIds?: Set<string> | null;
   serverFilterLoading?: boolean;
+  // True when the last /filters/preview resolution failed (403 on a pinned
+  // orgId the caller can't access, 500, network error, …). `serverFilterIds`
+  // is an EMPTY set in this case (never null — see useAdvancedFilterIds), so
+  // the table already renders zero rows; this only drives the inline error
+  // message that explains why (#4732).
+  serverFilterError?: boolean;
   // When false (default), decommissioned devices are hidden — matching the old
   // default view (status='all' implicitly excluded them). DevicesPage sets this
   // true only when the active filter group explicitly targets the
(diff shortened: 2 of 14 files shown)


ToddHebebrand:
**Review run:** 3 parallel Sonnet subagents via `/pr-review-toolkit:review-pr` — `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter`.

**Findings:** 2 Critical + 2 Important + 1 Suggestion raised → all addressed in `3d51bcc81`; 0 outstanding.
- **Critical** (silent-failure-hunter): the 401 exemption assumed `fetchWithAuth` always redirects before returning an unrecoverable 401 — two of its retry-after-refresh branches can return a *surviving* 401 without ever calling `handleSessionExpired`, which would have left `ids` at `null` and reopened #4732 behind a narrower trigger. Fixed: `ids` now fails closed (empty set) on **every** non-ok status unconditionally; only the visible toast/pill stays suppressed for 401 (to avoid competing with an in-flight redirect).
- **Critical** (code-reviewer): the "clears a prior error once the filter succeeds again" test was genuinely flaky — confirmed failing on isolated re-run (`error` resets synchronously at the start of every effect run, before the retried fetch resolves, so waiting on it raced the `ids` update). Fixed by waiting on `ids` instead.
- **Important** (silent-failure-hunter + pr-test-analyzer): grid view had no persistent error indicator, only the transient/one-shot toast. Added a persistent inline banner (`data-testid="device-filter-error-grid"`).
- **Important** (pr-test-analyzer): `DevicesPage.tsx` — the only place firing the toast and driving grid view's zero-row behavior — had no failure-path test coverage. Added a `DevicesPage.test.tsx` suite: list view empty-id-set + toast on 403, grid view zero cards + banner on 500, and no toast-spam on an unrelated re-render.
- **Suggestion** (silent-failure-hunter): toast/banner copy implied an automatic retry that doesn't exist. Reworded across all 8 locales.

**Tests:** `apps/web/src/hooks/useAdvancedFilterIds.test.ts` (10, incl. re-run 5x + isolated re-run 5x to confirm the flake is gone), `apps/web/src/components/devices/DeviceList.test.tsx` (99), `apps/web/src/components/devices/DevicesPage.test.tsx` + `DevicesPage.orgScope.test.tsx` (160+3 new), `apps/web/src/lib/i18n/localeParity.test.ts` (79), `apps/web/src/lib/__tests__/no-silent-mutations.test.ts` (130, unaffected) — all green via `npx vitest run <path>`. `npx tsc --noEmit` clean. `npx eslint` clean on all touched files.

PR #4785, opened 2026-09-03
fix(api/ai): dedupe intent outcome notifications by outcome class (#4465)

### LINKED ISSUE #4465: [AI agents] autonomy intents can double-notify the requester across a status advance
Autonomous ticket-triage intents (wave P2-4, `ticketAutonomousWrites`) can send two outcome notifications to the requester when a ticket's status advances during processing — the notification dedupe key includes status, so a status change between the two sends bypasses the dedupe.

ToddHebebrand:
## Problem

A ticket-autonomy intent carries **two** outbox rows — `createActionIntent` writes both `intent_created` and `intent_approved` for it (asserted at `apps/api/src/services/actionIntents/intentService.ticketAutonomy.test.ts:351`) — so `releaseAndNotify` runs **twice** for one intent by design; the second delivery is a backstop for the first. `releaseApprovedIntent` is CAS-guarded and safely idempotent, so the duplicate *release* is a no-op. The notification's dedupe key was the only thing making the duplicate *notification* a no-op too.

Keying it on the raw `intent.status` broke that the moment the status advanced between the two reads:

1. Job A wins the `approved -> executing` CAS and starts executing.
2. Job B loses it, returns immediately, reads `status = 'approved'` → writes `intent-outcome:<id>:approved` — "was approved and is now running".
3. Job A finishes, reads `status = 'completed'` → writes `intent-outcome:<id>:completed` — "was approved and has finished".

Two bells, one outcome. At-least-once outbox redelivery reproduces the same thing with no race at all.

## Fix

Key on the outcome **class**, not the raw status — `outcomeNotificationClass(status)`:

| status | class | shares a key with `granted`? | why |
|---|---|---|---|
| `approved` | `granted` | — | approved; execution pending |
| `executing` | `granted` | yes (silent) | same outcome, later observation |
| `completed` | `granted` | yes (silent) | same outcome, settled as expected |
| `failed` | `failed` | **no (corrects it)** | approved but did NOT run — the earlier "is now running" was wrong |
| `rejected` | `rejected` | no | terminal negative decision |
| `cancelled` | `cancelled` | no | terminal, withdrawn |
| `expired` | `expired` | no | terminal, nobody decided |
| anything else | `update` | no | unknown/pending — say only what is certain |

This keeps the property the status-scoped key was originally introduced to protect (a premature "is now running" must not suppress the later truthful "could not run") while dropping the bell for every *intermediate observation of the same outcome*. `rejected`/`cancelled`/`expired` are mutually exclusive with `granted` in practice, so the sequence that legitimately produces two bells is exactly `granted -> failed`.

Every unknown status shares the one `update` key on purpose: they all render the same "changed state" copy, so a second one is noise rather than news.

Both send sites move — the requester path (`intent-outcome:`) and the agent-recipient path (`agent-intent-outcome:`). The `rejected` / `cancelled` / `expired` / `failed` keys are **byte-identical** to before; only the granted trio changes shape.

### Deploy note (deliberate, not an oversight)

An intent that is mid-flight across the deploy and already has an `:approved` / `:executing` row can get one extra bell when it settles under the new `:granted` key. Approval intents are short-lived and the worst case is a single extra notification row, so the new namespace was preferred over reusing `approved` as the class name (which would have made the class segment read as a raw status).

## Tests

Red first — all 8 new assertions failed against the status-scoped key before the change.

The new block models the **real** partial unique index (`user_notifications_user_dedupe_key_uq` on `(user_id, dedupe_key)` WHERE `dedupe_key IS NOT NULL`) with a fake store rather than counting mock calls, so it asserts the number of notification **rows the requester actually ends up with**:

- two deliveries observing `approved`→`completed`, `approved`→`executing`, `executing`→`completed` → **exactly one** row, both calls carrying the same key;
- `approved` → `failed` → **two** rows, "Approval granted" then "Action failed";
- the same terminal outcome twice → one row;
- a status→key table over all eight statuses;
- both of the above repeated on the agent-recipient path.

Two pre-existing test titles/comments that described the key as "status-scoped" were updated to say "outcome class"; their assertions were already class-correct and are unchanged.

## Verification

- `cd apps/api && npx vitest run src/jobs/intentReleaseWorker` — **131 passed** (2 files), was 8 failed / 121 passed before the fix.
- `npx vitest run src/services/actionIntents src/services/userNotifications` — **517 passed** (26 files).
- `npx tsc --noEmit` — clean.
- `npx eslint` on both touched files — clean.

No schema, migration, RLS or cascade surface is touched.

Closes #4465

### DIFF
--- a/apps/api/src/jobs/intentReleaseWorker.test.ts
+++ b/apps/api/src/jobs/intentReleaseWorker.test.ts
@@ -3325,12 +3325,90 @@ describe('outcome notification dedupe identity (#4465)', () => {
       ['cancelled', 'intent-outcome:intent-1:cancelled'],
       ['expired', 'intent-outcome:intent-1:expired'],
       ['pending_approval', 'intent-outcome:intent-1:update'],
+      // Runtime fallback for a value the DB holds that ActionIntentStatus does
+      // not (drift, or a rollback across a status-adding deploy). The Record in
+      // the worker is exhaustive, so a REAL new status is a compile error there
+      // rather than a silent arrival here.
+      ['some_status_added_later', 'intent-outcome:intent-1:update'],
     ])('status %s keys the notification as %s', async (status, expected) => {
       await deliverApprovedObserving(status);
 
       const arg = notifyMock.createNotification.mock.calls[0]?.[0] as { dedupeKey: string };
       expect(arg.dedupeKey).toBe(expected);
     });
+  });
+
+  // The literal shape reported in #4465: an autonomy intent carries BOTH
+  // outbox rows, so these two deliveries are the pair that used to double-ring.
+  describe('the ticket_autonomy delivery pair (the reported shape)', () => {
+    /** `intent_created` on a ticket_autonomy row routes straight to
+     *  releaseAndNotify — it first reads decidedVia, THEN the intent. */
+    async function deliverCreatedObserving(status: string): Promise<void> {
+      intentServiceMock.transitionIntent.mockResolvedValueOnce(false);
+      dbState.selectActionIntentsResults.push([{ decidedVia: 'ticket_autonomy' }]);
+      dbState.selectActionIntentsResults.push([{ ...FOUR_EYES_INTENT, status }]);
+      await processIntentReleaseJob({ intentId: 'intent-1', eventType: 'intent_created' });
+    }
+
+    it('intent_created then intent_approved across a status advance notifies once', async () => {
+      const rows = installNotificationStore();
+
+      // The created-row recovery wins the race and starts executing...
+      await deliverCreatedObserving('executing');
+      // ...and the sibling approved row lands after it settled.
+      await deliverApprovedObserving('completed');
+
+      expect(rows).toHaveLength(1);
+      expect(rows[0]!.dedupeKey).toBe('intent-outcome:intent-1:granted');
+    });
+
+    it('the same pair still delivers the correction when the release actually failed', async () => {
+      const rows = installNotificationStore();
+
+      await deliverCreatedObserving('approved');
+      await deliverApprovedObserving('failed');
+
+      expect(rows).toHaveLength(2);
+      expect(rows[1]!.title).toBe('Action failed');
+    });
+  });
+
+  describe('across DIFFERENT event types', () => {
+    it('an intent_approved and an intent_expired observing the same status notify once', async () => {
+      const rows = installNotificationStore();
+
+      await deliverApprovedObserving('expired');
+      dbState.selectActionIntentsResults.push([{ ...FOUR_EYES_INTENT, status: 'expired' }]);
+      await processIntentReleaseJob({ intentId: 'intent-1', eventType: 'intent_expired' });
+
+      // The key is keyed off the observed STATUS, never the event that
+      // carried it, so two event types reporting one outcome stay one bell.
+      expect(rows).toHaveLength(1);
+      expect(rows[0]!.dedupeKey).toBe('intent-outcome:intent-1:expired');
+    });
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three specialists on `70ea8d0`: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`.

**Findings:** 7 raised → 5 addressed in `58155ae`, 1 declined with reasoning, 1 out of scope.

Addressed:
- **Class table not type-tied to `ActionIntentStatus`** (silent-failure-hunter, medium/latent). A 9th status added to `actionIntentStatusEnum` would have collapsed silently into the `update` bucket with no compiler and no runtime signal. Now a `Record<ActionIntentStatus, string>`, so it is a compile error until classified; the runtime fallback survives only for a value the DB holds that the type does not, and is covered by a test.
- **False claim in the truth table** (silent-failure-hunter, low — but a correctness defect in the reasoning). It said `rejected`/`cancelled`/`expired` are all mutually exclusive with `granted`. `cancelActionIntent` transitions from `['pending_approval', 'approved']` (`intentService.ts`), so `granted -> cancelled` is reachable — an approver can withdraw an intent that already rang a granted bell. The comment is corrected and now also records that a cancel writes no outbox row, so this worker only ever observes `cancelled` on a late delivery of some *other* event.
- **The reported shape was not actually exercised** (pr-test-analyzer, sev 2). Every new test drove `intent_approved` only; the `intent_created` (ticket_autonomy) + `intent_approved` pair — the one that actually double-rang — had no test. Added, plus the same pair still delivering the `failed` correction.
- **Dedupe across different event types** (pr-test-analyzer, sev 3). Added both directions: same outcome via `intent_approved` + `intent_expired` → one bell; a granted bell then an `intent_rejected` delivery → still corrected.
- **`metadata.status` unpinned** (pr-test-analyzer, sev 1-2). Added a guard that only the *key* collapses — the row still carries the raw observed status for the bell UI and support.

Declined:
- **Instrument the deduped `null` return** (silent-failure-hunter, medium). Every ticket-autonomy intent dedupes exactly once *by design* — that is the mechanism, not an anomaly — so a log line per suppression would be pure noise in the expected path. The real risk behind the finding was a misclassification going unnoticed; the exhaustive `Record` above turns that into a compile error, which is a stronger guard than a counter nobody alerts on. No caller of `createNotification` anywhere in the repo inspects the return, so a metric here would also be a one-off.

Out of scope (pre-existing, flagged for the maintainer):
- **`cancelActionIntent` emits no outcome event at all.** It writes no `action_intents_outbox` row, so a requester who was told "was approved and is now running" is never told if an approver then cancels it. Identical under the old status-scoped key — no event fires either way — so it is not a regression from this PR, but it is a real gap adjacent to it and wants its own issue.

Not acted on: agent-path status table (sev 2) — the `agent-intent-outcome:` template is already pinned for `rejected` and `failed` by pre-existing tests, and both key builders call the same helper, so the requester-path table over all eight classes covers the mapping.

**Tests:** `cd apps/api && npx vitest run src/jobs/intentReleaseWorker` — **137 passed** (2 files). `npx vitest run src/services/actionIntents src/services/userNotifications` — 517 passed (26 files). `npx tsc --noEmit` clean; `npx eslint` clean on both touched files.

**Control proven, not assumed:** reverting `outcomeNotificationClass` to return the raw status fails **11** of the new assertions — including the ticket_autonomy pair and the metadata guard — so the suite discriminates rather than confirms.

**Status:** review-clean at `58155ae`, awaiting maintainer merge. Not merged, issue #4465 left open.

PR #4789, opened 2026-09-03
fix(api): re-revoke UPDATE/DELETE/TRUNCATE on append-only tables missing from ensureAppRole (#4371)

### LINKED ISSUE #4371: ensureAppRole blanket GRANT re-grants UPDATE on pam_actuation_results, undoing the append-only revoke
## Summary

`ensureAppRole`'s blanket per-boot GRANT re-permits `breeze_app` `UPDATE` on `pam_actuation_results`, silently undoing that table's own append-only `REVOKE`. Today the `BEFORE UPDATE` trigger is the *sole* enforcement keeping the table append-only — the privilege layer of the intended belt-and-suspenders pair isn't actually holding.

## Where this was found

Surfaced as observation O-3 during the S0 Track B current-main port's final independent whole-branch review (`.superpowers/sdd/2026-08-31-s0-track-b-current-main-port/final-review.md`, and its root cause T3-M3), while auditing the append-only evidence tables Track B added to the org-merge registry.

## The mechanism

- `apps/api/migrations/2026-09-16-pam-actuation-lifecycle.sql` grants `breeze_app` only `SELECT, INSERT, REFERENCES` on `pam_actuation_results` and explicitly `REVOKE UPDATE, DELETE, TRUNCATE ON pam_actuation_results FROM breeze_app`, backed by a `BEFORE UPDATE OR DELETE ... FOR EACH ROW` trigger (`pam_actuation_results_block_mutation` / `pam_actuation_results_append_only()`) that unconditionally `RAISE`s (`42501`) on `UPDATE`, and on `DELETE` only outside the audit-retention escape hatch.
- `apps/api/src/db/ensureAppRole.ts` runs on every boot (before and after migrations) and issues a blanket `GRANT SELECT, INSERT, UPDATE, DELETE, REFERENCES ON ALL TABLES IN SCHEMA public TO breeze_app` (step 4). That blanket grant re-permits `UPDATE` on `pam_actuation_results`, because — unlike several other append-only tables — there is no per-table re-revoke override for it in the `DO $$ ... END $$` block that follows.
- Contrast with the pattern this same file already uses for other append-only evidence tables, including two added by this port: `ensureAppRole.ts:155-166` re-revokes `UPDATE, TRUNCATE` on `agent_health_observations` and `software_inventory_observations` on every boot (and similarly for `audit_logs`, `audit_log_chain`, `audit_chain_anchors`, `partner_export_*`, `llm_egress_events` earlier in the same block). `pam_actuation_results` has no equivalent block.
- Net effect: at the privilege layer, `breeze_app` has `UPDATE` on `pam_actuation_results` after every boot. The `BEFORE UPDATE` trigger's unconditional `RAISE EXCEPTION` is what actually stops a mutation today — the REVOKE in the migration is real at migration-apply time but is silently undone on the very next boot.

## Scope note

This gap pre-dates the S0 Track B current-main port and was correctly left untouched by it — the port's binding constraint was that `SPECIAL.pam_actuations` / `pam_actuation_results` and Track E's blocks-merge semantics stay byte-identical to main, which the port's final review independently verified held. This issue is filed against main, not against the port branch.

## Suggested fix

Add a re-revoke block for `pam_actuation_results` in `ensureAppRole.ts`, following the existing `agent_health_observations` / `software_inventory_observations` pattern at `ensureAppRole.ts:155-166`: wrap in `IF EXISTS (... table_name='pam_actuation_results')`, then `REVOKE UPDATE, TRUNCATE ON TABLE pam_actuation_results FROM breeze_app` (leave `DELETE` alone if `breeze_audit_admin`-mediated retention needs to keep working the way it does for other tables — check the retention path before deciding whether to also revoke `DELETE` from `breeze_app` here versus leaving that to the trigger's `breeze.allow_audit_retention` escape hatch). Also worth a short audit of the other REVOKE-style/append-only tables in the schema to confirm each has a matching `ensureAppRole.ts` re-revoke — `pam_actuation_results` may not be the only one missing its half of the pair.

No exploit steps included; this is a defense-in-depth privilege gap, not an actively exploitable path beyond what the trigger already blocks.

ToddHebebrand:
## Summary

`ensureAppRole()`'s per-boot blanket GRANT (step 4: `GRANT SELECT, INSERT, UPDATE, DELETE, REFERENCES ON ALL TABLES IN SCHEMA public TO breeze_app`) silently re-permits whatever an append-only table's migration `REVOKE`d from `breeze_app` — unless step 5 explicitly re-applies that `REVOKE` after the blanket `GRANT` on every boot. `pam_actuation_results` shipped without a re-revoke block in `ensureAppRole.ts`, so its `BEFORE UPDATE/DELETE` trigger has been the *sole* enforcement in production since `2026-09-16-pam-actuation-lifecycle.sql` landed — the privilege half of the intended belt-and-suspenders pair never actually held.

## The sweep

Per the issue's suggestion, I swept every `REVOKE ... FROM breeze_app` statement across `apps/api/migrations/*.sql` against `ensureAppRole.ts`'s existing re-revoke list and found **five more tables with the same gap**, beyond the reported `pam_actuation_results`: `ml_feedback_events`, `peripheral_policy_delivery_events`, `agent_rollback_events`, `automation_action_results`, `device_software_inventory_state`.

## Writer-path matrix (the real substance of this PR)

Closing the privilege gap is not enough on its own — every *legitimate* writer of these six tables (device permanent-delete, device org-move, org merge, org erasure, retention/prune jobs) has to keep working under the now-real privilege wall, or the fix just trades one bug for a different one. CI caught two classes of this on the branch as I fixed the underlying gap; here's the full audit, enumerated per table:

| Table | breeze_app revoked | breeze_app kept | Device delete | Device org-move | Org merge | Org erasure |
|---|---|---|---|---|---|---|
| `pam_actuation_results` | `UPDATE, DELETE, TRUNCATE` | — | `breeze_audit_admin` escalation | N/A — PAM-history devices can't move orgs at all | `blocks-merge` | `breeze_audit_admin` |
| `agent_rollback_events` | `UPDATE, DELETE, TRUNCATE` | — | `breeze_audit_admin` escalation | `breeze_cascade_device_org_id()` SECURITY DEFINER trigger | `leave-for-erasure` | `breeze_audit_admin` |
| `ml_feedback_events` | `UPDATE, DELETE, TRUNCATE` | — | N/A — no `device_id` column | N/A — no `device_id` column | `leave-for-erasure` | `breeze_audit_admin` |
| `peripheral_policy_delivery_events` | `DELETE, TRUNCATE` | `UPDATE` | `breeze_audit_admin` escalation | real `breeze_app` UPDATE (trigger-gated restamp) | plain `repoint` | `breeze_audit_admin` |
| `automation_action_results` | `TRUNCATE` | `UPDATE, DELETE` | plain `breeze_app` DELETE | real `breeze_app` UPDATE | plain `repoint` | plain `breeze_app` DELETE |
| `device_software_inventory_state` | `TRUNCATE` | `UPDATE, DELETE` | plain `breeze_app` DELETE | real `breeze_app` UPDATE | plain `repoint` | plain `breeze_app` DELETE |

Every cell above is now verified — either by an integration test that was previously red, or by the `orgMergeRegistry.integration.test.ts` contract check, or by a `grep` confirming no writer path exists.

### 1. Org-merge registry conflict (`agent_rollback_events`)

`executeOrgMerge` repointed `agent_rollback_events` with a plain `UPDATE ... SET org_id = survivor` (`REPOINT_TABLES`, `orgMergeRegistry.ts`), which the blanket-GRANT bug had been silently masking. `agent_rollback_events` has the identical privilege topology as `ml_feedback_events` (neither `breeze_app` nor `breeze_audit_admin` can ever `UPDATE` it) — reclassified it `leave-for-erasure` to match, removed it from `REPOINT_TABLES`, updated the two `orgMergeRegistry.integration.test.ts` assertions that hardcoded the old `repoint` classification.

### 2. Device-delete and device-move conflicts (`pam_actuation_results`, `agent_rollback_events`)

- `DELETE FROM pam_actuation_results WHERE device_id = ...` (permanent device-delete cascade, `services/deviceDeletion.ts`) — `breeze_app` has `DELETE` fully revoked. **Fix**: generalized the pre-existing `peripheral_policy_delivery_events` → `breeze_audit_admin` escalation (`SET LOCAL ROLE breeze_audit_admin` + `breeze.allow_audit_retention='1'`, one statement, immediately restored) into `DEVICE_CASCADE_AUDIT_ADMIN_TABLES`, covering `pam_actuation_results` and `agent_rollback_events` too — both have `breeze_audit_admin GRANT SELECT, DELETE` in their own migrations. This follows the existing `audit_logs`-style precedent rather than weakening any revoke.
- `UPDATE agent_rollback_events SET org_id = ... WHERE device_id = ...` (device org-move, `routes/devices/moveOrg.ts`'s app-level denormalized-table loop) — `breeze_app` has `UPDATE` fully revoked. **Fix**: added `agent_rollback_events` to `DEVICE_ORG_FK_CASCADE_TABLES` (`routes/devices/core.ts`) — verified against real Postgres (`SELECT * FROM breeze_device_child_orgid_tables()`) that its org_id is already restamped by the pre-existing `SECURITY DEFINER` `breeze_cascade_device_org_id()` trigger (`migrations/2026-05-18-device-child-orgid-cascade.sql`), which fires on the same `devices` UPDATE this loop follows and bypasses the privilege wall the way FK actions do — the app-level loop's own UPDATE was always redundant with it, never load-bearing.

`pam_actuation_results` has no device org-move conflict: a device with PAM actuation history can't move orgs at all (`devices_pam_history_move_guard`), and it's explicitly excluded from the generic org_id-cascade discovery function (`migrations/2026-09-17-pam-device-move-guard.sql`).

### 3. One residual gap found, deliberately NOT fixed here — filed as **[#4806](https://github.com/LanternOps/breeze/issues/4806)**

`peripheral_policy_delivery_events` keeps `UPDATE` granted because `moveOrg.ts`'s restamp loop genuinely needs it (unlike `agent_rollback_events`, it has no `ON UPDATE CASCADE` FK and wasn't on the SECURITY DEFINER trigger's exclusion list before now). Its append-only guarantee for that write therefore still depends entirely on the trigger's own data-shape check (`(to_jsonb(NEW) - 'org_id') = (to_jsonb(OLD) - 'org_id') AND EXISTS (...)`), not a privilege wall — a real behavior change to a live writer path, so it gets its own PR/review/regression test rather than riding along on this one.

`automation_action_results` and `device_software_inventory_state` are **not** append-only by design (no append-only trigger exists for either) — only `TRUNCATE` was ever restricted, as defense-in-depth, and no writer path issues one.

## Regression tests

- `apps/api/src/db/ensureAppRole.appendOnlyCoverage.test.ts` — static-analysis unit test (no DB): parses every migration `REVOKE` and every `ensureAppRole.ts` re-revoke block and fails on any gap. Extended with a `WRITER_PATH_MATRIX` block asserting **exact** privilege equality (not just "at least") for all six tables, so a future PR can't silently over-tighten one and reopen a writer-path conflict without the test forcing a conscious update. Regex hardened to tolerate multi-line privilege lists and REVOKEs embedded in dynamic SQL.
- `apps/api/src/__tests__/integration/ensureAppRoleAppendOnlyPrivileges.integration.test.ts` — real-Postgres `has_table_privilege()` check for all 6 tables' runtime effect.
- `deviceDeletion.test.ts` — generalized the single-table escalation assertion to `it.each` over all three `DEVICE_CASCADE_AUDIT_ADMIN_TABLES` entries, plus a negative control proving `automation_action_results`/`device_software_inventory_state` do **not** get the escalation.
- `moveOrg.test.ts` / `moveOrg.coverage.test.ts` — updated the two hardcoded `DEVICE_ORG_FK_CASCADE_TABLES` array assertions to include `agent_rollback_events`.

## Testing

- `cd apps/api && npx vitest run src/db/ensureAppRole src/services/orgMerge src/routes/orgMerge src/jobs/orgMerge src/routes/devices/moveOrg src/routes/devices/cascadeDelete src/services/deviceDeletion` — **235/235 passed** (11 unit files).
- Foreground, against the shared integration test DB (`BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=1` — a sibling worktree had migrated the shared DB ahead of this branch; the drift guard's own documented escape hatch, safe here since none of these are the schema-enumerating suites it warns about):
  `npx vitest run --config vitest.integration.config.ts src/__tests__/integration/deviceDeleteDiscoveredAssetLink src/__tests__/integration/pamActuationLifecycle src/__tests__/integration/quickSupportChain src/__tests__/integration/agentRunMoveSemantics src/__tests__/integration/deviceMoveOrgCurrency src/__tests__/integration/ensureAppRoleAppendOnly src/__tests__/integration/orgMergeRegistry src/__tests__/integration/inboundEmailContactRequester src/__tests__/integration/ticketDraftsRls` — **9 files, 68/68 passed** (previously red on shards 1–3).
- Typecheck: whole-project `npx tsc --noEmit` in `apps/api` repeatedly OOM-crashed on this host (system-wide memory pressure from concurrent sessions, confirmed via `vm_stat`/`top`, unrelated to this change). Verified type-safety instead via a scoped `tsconfig` extending the real one with `include` restricted to every touched/added file — clean. CI's typecheck job on unloaded runners is authoritative.

Closes #4371

### DIFF
--- a/apps/api/src/__tests__/integration/ensureAppRoleAppendOnlyPrivileges.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ensureAppRoleAppendOnlyPrivileges.integration.test.ts
@@ -1,0 +1,106 @@
+/**
+ * Integration test for #4371: ensureAppRole()'s blanket per-boot GRANT
+ * (`GRANT SELECT, INSERT, UPDATE, DELETE, REFERENCES ON ALL TABLES IN SCHEMA
+ * public TO breeze_app`, step 4 of `src/db/ensureAppRole.ts`) silently
+ * re-permits whatever an append-only table's migration REVOKEd from
+ * breeze_app, unless step 5 re-applies that REVOKE after the blanket GRANT.
+ *
+ * `pam_actuation_results` shipped without a re-revoke block, so its
+ * `BEFORE UPDATE/DELETE` trigger was the *sole* enforcement in production —
+ * the privilege layer of the intended belt-and-suspenders pair never
+ * actually held. The sweep that fixed #4371 found five more tables with the
+ * same gap: ml_feedback_events, peripheral_policy_delivery_events,
+ * agent_rollback_events, automation_action_results, and
+ * device_software_inventory_state.
+ *
+ * `ensureAppRole.appendOnlyCoverage.test.ts` (unit, static analysis) proves
+ * the *source* stays in sync going forward. This test proves the actual
+ * *runtime effect*: that breeze_app really lacks these privileges against a
+ * real Postgres server, after the test setup's autoMigrate -> ensureAppRole
+ * boot sequence has run — mirroring the existing audit_logs privilege check
+ * in `audit-append-only.integration.test.ts`.
+ */
+import './setup';
+import { describe, it, expect } from 'vitest';
+import { sql } from 'drizzle-orm';
+import { db } from '../../db';
+
+interface PrivilegeRow {
+  can_select: boolean;
+  can_insert: boolean;
+  can_update: boolean;
+  can_delete: boolean;
+  can_truncate: boolean;
+}
+
+async function tablePrivileges(table: string): Promise<PrivilegeRow> {
+  // `table` is always one of this file's own hardcoded test parameters
+  // (never external input) — passed as an ordinary bound `text` parameter
+  // to has_table_privilege(), not spliced into the SQL syntax, so no
+  // sql.raw() / identifier-quoting is needed here.
+  const rows = (await db.execute(sql`
+    SELECT
+      has_table_privilege('breeze_app', ${table}, 'SELECT') AS can_select,
+      has_table_privilege('breeze_app', ${table}, 'INSERT') AS can_insert,
+      has_table_privilege('breeze_app', ${table}, 'UPDATE') AS can_update,
+      has_table_privilege('breeze_app', ${table}, 'DELETE') AS can_delete,
+      has_table_privilege('breeze_app', ${table}, 'TRUNCATE') AS can_truncate
+  `)) as unknown as Array<PrivilegeRow>;
+  return rows[0]!;
+}
+
+describe('ensureAppRole append-only re-revoke — runtime privilege check (#4371)', () => {
+  // Tables where the migration revokes the FULL append-only set
+  // (UPDATE, DELETE, TRUNCATE) and breeze_app keeps only SELECT/INSERT.
+  it.each([
+    'pam_actuation_results',
+    'agent_rollback_events',
+  ])('breeze_app has no UPDATE, DELETE, or TRUNCATE on %s after ensureAppRole runs', async (table) => {
+    const p = await tablePrivileges(table);
+    expect(p.can_update).toBe(false);
+    expect(p.can_delete).toBe(false);
+    expect(p.can_truncate).toBe(false);
+    expect(p.can_insert).toBe(true);
+    expect(p.can_select).toBe(true);
+  });
+
+  // ml_feedback_events: migration only revokes UPDATE, DELETE (no INSERT
+  // of TRUNCATE to begin with), but ensureAppRole.ts also re-revokes
+  // TRUNCATE as defense-in-depth, consistent with the file's established
+  // pattern elsewhere (audit_logs, agent_health_observations).
+  it('breeze_app has no UPDATE, DELETE, or TRUNCATE on ml_feedback_events after ensureAppRole runs', async () => {
+    const p = await tablePrivileges('ml_feedback_events');
+    expect(p.can_update).toBe(false);
+    expect(p.can_delete).toBe(false);
+    expect(p.can_truncate).toBe(false);
+    expect(p.can_insert).toBe(true);
+    expect(p.can_select).toBe(true);
+  });
+
+  // peripheral_policy_delivery_events intentionally KEEPS UPDATE granted
+  // (delivery status transitions are ordinary app writes) — only DELETE and
+  // TRUNCATE are revoked.
+  it('breeze_app keeps UPDATE but has no DELETE or TRUNCATE on peripheral_policy_delivery_events', async () => {
+    const p = await tablePrivileges('peripheral_policy_delivery_events');
+    expect(p.can_update).toBe(true);
+    expect(p.can_delete).toBe(false);
+    expect(p.can_truncate).toBe(false);
+    expect(p.can_insert).toBe(true);
+    expect(p.can_select).toBe(true);
+  });
+
+  // automation_action_results and device_software_inventory_state
+  // intentionally KEEP UPDATE and DELETE granted (ordinary mutable state) —
+  // only TRUNCATE is revoked.
+  it.each(['automation_action_results', 'device_software_inventory_state'])(
+    'breeze_app keeps UPDATE/DELETE but has no TRUNCATE on %s',
+    async (table) => {
+      const p = await tablePrivileges(table);
+      expect(p.can_update).toBe(true);
+      expect(p.can_delete).toBe(true);
+      expect(p.can_truncate).toBe(false);
+      expect(p.can_insert).toBe(true);
+      expect(p.can_select).toBe(true);
+    },
+  );
+});
--- a/apps/api/src/__tests__/integration/orgMergeRegistry.integration.test.ts
+++ b/apps/api/src/__tests__/integration/orgMergeRegistry.integration.test.ts
@@ -298,17 +298,25 @@ describe('Org merge policy registry contract', () => {
     expect(extra).toEqual([]);
   });
 
-  it('organizations is loser-shell; the four append-only tables are leave-for-erasure', () => {
+  it('organizations is loser-shell; the five append-only tables are leave-for-erasure', () => {
     expect(policies.get('organizations')).toEqual({ kind: 'loser-shell' });
-    for (const t of ['audit_logs', 'audit_log_chain', 'audit_chain_anchors', 'ml_feedback_events']) {
+    // agent_rollback_events joined this list in the #4371 fixup: breeze_app
+    // has no UPDATE on it (same privilege topology as ml_feedback_events),
+    // so it can never be a live 'repoint' target — see the SPECIAL entry's
+    // note in orgMergeRegistry.ts for the full history.
+    for (const t of ['audit_logs', 'audit_log_chain', 'audit_chain_anchors', 'ml_feedback_events', 'agent_rollback_events']) {
       expect(policies.get(t)?.kind, t).toBe('leave-for-erasure');
     }
   });
 
   it('repoints Track D device-control state and evidence with the device owner', () => {
+    // agent_rollback_events is deliberately NOT in this list (#4371 fixup):
+    // breeze_app has UPDATE revoked on it, so it can't be repointed by the
+    // merge — it's asserted 'leave-for-erasure' above instead.
+    // agent_rollback_directives keeps its own separate 'repoint' policy;
+    // its evidence table just doesn't follow along anymore.
     for (const table of [
       'agent_rollback_directives',
-      'agent_rollback_events',
       'peripheral_policy_delivery_events',
       'peripheral_policy_device_states',
     ]) {
--- a/apps/api/src/db/ensureAppRole.appendOnlyCoverage.test.ts
+++ b/apps/api/src/db/ensureAppRole.appendOnlyCoverage.test.ts
@@ -26,8 +26,18 @@ const ensureAppRolePath = path.resolve(__dirname, './ensureAppRole.ts');
 // Table-level REVOKE ... FROM breeze_app in a migration. Deliberately
 // excludes REVOKE ... ON FUNCTION / SEQUENCE / DATABASE / SCHEMA / ALL —
 // those aren't the blanket-GRANT-on-ALL-TABLES hazard this test guards.
+//
+// Two deliberate tolerances (both exercised by regression cases below):
+// - `[A-Z,\s]+?` (not `[A-Z, ]+?`) so a privilege list wrapped across lines
+//   still matches.
+// - `['"]?` before the terminating `;` so a REVOKE embedded verbatim inside
+//   a `DO $$ ... EXECUTE 'REVOKE ... FROM breeze_app;' ... END $$` dynamic-SQL
+//   string (whose closing quote sits between `breeze_app` and `;`) still
+//   matches. This does NOT resolve a *templated* table name built via
+//   `format('... %I ...', tbl)` — that requires reading the migration by
+//   eye; there is no such case in the migrations directory today.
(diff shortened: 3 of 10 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer — Sonnet subagents)

**Findings:** code-reviewer: 0 findings at confidence ≥80 — independently re-verified all 6 privilege-set matches against the source migrations, SQL syntax, and sweep completeness (cross-checked against `AUDIT_ADMIN_REQUIRED_TABLES` in `tenantCascade.ts`). pr-test-analyzer: verified the new static test actually catches a deleted re-revoke block (red/green sanity check passed), then raised 3 findings — addressed in `4d4efe4da`:
1. Regex false-negative risk if a future migration wraps a multi-line privilege list — fixed (`[A-Z,\s]+?`).
2. Regex false-negative risk if a future REVOKE is embedded in a `DO $$ ... EXECUTE '...'` dynamic-SQL string (closing quote breaks the terminator match) — fixed (tolerate an optional trailing quote before `;`). Templated table names via `format('...%I...', tbl)` remain a documented limitation (no such case exists in the migrations directory today).
3. The static unit test only proves *source* stays in sync, not that the privilege is actually absent at *runtime* — added `ensureAppRoleAppendOnlyPrivileges.integration.test.ts`, mirroring the existing `audit_logs` `has_table_privilege()` pattern in `audit-append-only.integration.test.ts`, covering all 6 tables the fix touches.

0 outstanding findings from either reviewer.

**Tests:** `cd apps/api && npx vitest run src/db/ensureAppRole` — 17/17 passed (unit: `ensureAppRole.test.ts` + the new `ensureAppRole.appendOnlyCoverage.test.ts`, red-first verified against the pre-fix source). Standalone strict `tsc --noEmit` (matching root `tsconfig.json` options) on all 3 touched/added files — clean. Whole-project `tsc --noEmit` in `apps/api` repeatedly OOM-crashed locally due to host-wide memory pressure from concurrent sessions (confirmed via `vm_stat`/`top`, unrelated to this change) — CI's typecheck job on unloaded runners is authoritative there. The new integration test (`ensureAppRoleAppendOnlyPrivileges.integration.test.ts`) was not run locally against the shared `breeze-postgres-test` container to avoid disrupting other concurrent worktree sessions on that shared DB — it will run under CI's required `integration-test` job.

ToddHebebrand:
Fixed the Integration Tests red on shards 1-3 (`inboundEmailContactRequester`, `ticketDraftsRls`: \`permission denied for table agent_rollback_events\` during \`executeOrgMerge\`).

**Root cause:** the re-revoke this PR adds for `agent_rollback_events` (correctly matching its migration) is a real conflict with `orgMergeRegistry.ts`, which had it classified `repoint` — a plain `UPDATE ... SET org_id` that `breeze_app` (and `breeze_audit_admin`) has never actually had permission to run. This worked on `main` only because the `ensureAppRole` privilege gap this PR fixes was silently re-permitting UPDATE — a pre-existing merge-registry bug this fix exposed rather than caused.

**Fix (`d10466b09`):** reclassified `agent_rollback_events` as `leave-for-erasure` in `apps/api/src/services/orgMergeRegistry.ts`, matching `ml_feedback_events`'s identical privilege topology (no role can UPDATE either table). Removed it from `REPOINT_TABLES`, updated the two `orgMergeRegistry.integration.test.ts` assertions that hardcoded the old `repoint` classification. No revoke was weakened.

Checked the other 5 tables this PR touches for the same conflict — only `agent_rollback_events` had it; `ml_feedback_events`/`pam_actuation_results` were already excluded from repoint, and `peripheral_policy_delivery_events`/`automation_action_results`/`device_software_inventory_state` never lost UPDATE. Confirmed against the live test DB: the `every repoint / repoint-dedupe table grants breeze_app UPDATE` contract test in `orgMergeRegistry.integration.test.ts` now passes with 0 violations.

**Verified in the foreground** against the shared integration DB: the two previously-failing suites plus the new `ensureAppRoleAppendOnlyPrivileges` integration test (3 files, 22/22 passed), `orgMergeRegistry.integration.test.ts` (14/14 passed), and the `orgMerge`/`orgMergeRegistry` unit suites (125/125 passed). PR body updated with the full writeup.

Head SHA: `d10466b09e7805d361f7eea245bc8be3a655ec7c`

ToddHebebrand:
Fixed the two remaining Integration shards 2/3 reds: \`permission denied for table pam_actuation_results\` (device permanent-delete cascade — deviceDeleteDiscoveredAssetLink, pamActuationLifecycle, quickSupportChain) and \`permission denied for table agent_rollback_events\` (device org-move — agentRunMoveSemantics, deviceMoveOrgCurrency).

Same root cause as the earlier org-merge fix: closing the ensureAppRole gap exposed that these writer paths had been silently relying on the blanket per-boot GRANT too.

**Fix (\`45af4c51d\`):**
- Device permanent-delete: generalized the existing \`peripheral_policy_delivery_events\` → \`breeze_audit_admin\` escalation (\`services/deviceDeletion.ts\`) into \`DEVICE_CASCADE_AUDIT_ADMIN_TABLES\`, now also covering \`pam_actuation_results\` and \`agent_rollback_events\` — both have \`breeze_audit_admin GRANT SELECT, DELETE\` in their own migrations. Same audit-admin pattern as \`audit_logs\`, no revoke weakened.
- Device org-move: added \`agent_rollback_events\` to \`DEVICE_ORG_FK_CASCADE_TABLES\` (\`routes/devices/core.ts\`) — verified against real Postgres that its org_id is already restamped by the pre-existing SECURITY DEFINER \`breeze_cascade_device_org_id()\` trigger, so the app-level loop's own privileged UPDATE was always redundant, never load-bearing.

Enumerated every writer path (device delete, device move, org merge, org erasure, retention/prune jobs — none exist for these 6 tables) and documented the full matrix in \`ensureAppRole.ts\`'s comment block, encoded as an executable exact-privilege-set contract in \`ensureAppRole.appendOnlyCoverage.test.ts\` (\`WRITER_PATH_MATRIX\`).

**One residual gap found, deliberately left for a follow-up: [#4806](https://github.com/LanternOps/breeze/issues/4806).** \`peripheral_policy_delivery_events\` keeps \`UPDATE\` granted because \`moveOrg.ts\`'s restamp loop genuinely needs it — its append-only guarantee for that column set depends on the trigger alone, not a privilege wall. That's a real behavior change to a live writer path, so it gets its own PR rather than riding along here.

**Verified in the foreground**, all 9 target integration suites (deviceDeleteDiscoveredAssetLink, pamActuationLifecycle, quickSupportChain, agentRunMoveSemantics, deviceMoveOrgCurrency, ensureAppRoleAppendOnly, orgMergeRegistry, inboundEmailContactRequester, ticketDraftsRls): 68/68 passed. Unit suites (ensureAppRole, orgMerge×3, moveOrg×2, deviceDeletion): 235/235 passed. PR body updated with the full per-table writer-path matrix.

Head SHA: \`45af4c51d260379775d747846ba50c109f20af74\`

PR #4790, opened 2026-09-03
fix(api): mirror moveOrg's intent scope tombstone into the device org cascade trigger (#4454)

### LINKED ISSUE #4454: AI agents: mirror moveOrg scope tombstone into breeze_cascade_device_org_id()
`moveOrg` tombstones the intent target scope for non-terminal scoped action intents (wave P2-2, intent target scope) in application code, but the DB-side cascade function `breeze_cascade_device_org_id()` doesn't mirror that tombstone — a device org-move that bypasses the app-layer moveOrg path could leave a scoped intent pointing at a device that's now in a different org.

ToddHebebrand:
## What

`apps/api/src/routes/devices/moveOrg.ts` tombstones `action_intents.scope_device_id` for every **live** intent aimed at a device that is leaving its org. The DB-side cascade — `breeze_cascade_device_org_id()`, the path every direct-SQL / non-route caller takes, including orgMerge's `devices` repoint — never mirrored it.

P2-2 review round 1 deliberately scoped that fix to the route only and carried the hole as a documented known gap in `moveOrg.coverage.test.ts`. Consequence: a raw `UPDATE devices SET org_id = ...` left a live intent still naming a device that now belongs to a **different tenant**.

Same cross-tenant dangling-pointer class as #3828 (`metric_anomaly_incidents.agent_run_id`) and #4215 (`ai_agent_runs.ticket_id`) — both carried the same way until they were mirrored into this same function. Closed here the same way.

## How

**Migration** `apps/api/migrations/2026-10-06-124500-cascade-device-org-move-intent-scope.sql` — `CREATE OR REPLACE FUNCTION`, body copied from the newest definition (`2026-10-04-100000-ticket-requester-contact.sql`) with one statement added. No inner `BEGIN`/`COMMIT`; the trigger itself is unchanged and not redeclared.

```sql
UPDATE public.action_intents
  SET scope_device_id = NULL
  WHERE scope_device_id = NEW.id
    AND status IN ('pending_approval', 'approved', 'executing');
```

Placed between the `metric_anomaly_incidents` reverse pointer and the tickets requester detach, so the trigger's internal order matches the route's.

Three judgement calls, each cross-checked against an independent read of the repo before implementing:

- **Live statuses only**, matching the route and the `actionIntents.scopeDeviceId` schema comment. A terminal-status intent is a historical record of an already-decided action; its target at decision time is a fact, not something release re-validates. An unconditional detach would rewrite it.
- **No merge fence**, deliberately unlike the tickets requester detach directly below it. That one skips itself while `organizations.status = 'merging'` because the contact travels to the survivor org alongside the ticket. An intent never travels: `services/orgMergeRegistry.ts` classifies `action_intents` as `leave-for-erasure` precisely because `org_id` is trigger-immutable, so a repoint would raise and abort the merge. A live intent in a merging loser org is erased with the loser shell either way, and both outcomes fail closed at release — tombstoning is the strictly safer one, leaving no cross-tenant device id behind at all.
- **No historical backfill**, for the reason `2026-09-29`'s header gives for the same omission: `action_intents` is `FORCE ROW LEVEL SECURITY`, so a contextless migration-time cleanup would match zero rows on a managed Postgres and report a silent "0 cleaned". Pre-fix rows are already failed closed by the release path's org-mismatch check (`intentTargetScope.ts`), so they are not exploitable.

The route's own statement is **unchanged**. Like the other mirrored detaches, it now normally matches nothing (the AFTER trigger on the `devices` flip has already run) and is kept so the route stays correct on its own if the trigger is ever absent.

## Tests

**Red first, both halves.** With the migration moved aside: the two new static cases failed on the unfixed trigger body, and the integration suite failed on real Postgres with the live intents still pointing at the moved device. With the migration in place, all pass.

- **`moveOrg.coverage.test.ts`** (blocking `test-api` job) now asserts **both** sites — the route source *and* the newest migration's function body — plus that the tombstone precedes the generic re-stamp loop. A future edit that fixes one site and forgets the other goes red. The newest-definition resolver was hoisted to module scope so exactly one place decides which migration is authoritative.
- **`deviceMoveOrgIntentScopeTombstone.integration.test.ts`** (new, blocking `integration-test` job) drives a raw `UPDATE devices SET org_id` twice: once as the superuser admin client (the literal "somebody ran it in psql" path) and once as the unprivileged `breeze_app` role under `withSystemDbAccessContext` — the shape orgMerge has, and **the only case that would catch FORCE RLS silently swallowing the UPDATE**. Static assertions cannot reach any of the three ways this statement could be present and still not work (forced RLS, the `action_intents_block_content_update()` immutability trigger, `action_intents_scope_device_chk`). Every case carries a `completed`-status control that must still point at the moved device.

Verified locally against an isolated test stack (`pnpm test-stack up`):

| Check | Result |
|---|---|
| `deviceMoveOrgIntentScopeTombstone.integration.test.ts` | 3 passed (2 red before the migration) |
| `moveOrg.coverage.test.ts` + `autoMigrate.test.ts` | 88 passed (2 red before the migration) |
| `deviceMoveOrgCurrency`, `agentRunMoveSemantics`, `pamDeviceMoveGuard`, `actionIntentsImmutabilityTrigger` | 53 passed |
| `orgMerge`, `orgMergeRegistry`, `ticket-move-org`, `tenantCascade` | 45 passed |
| `integration-suite-coverage` | 5 passed |
| Migration re-applied twice via `psql` | idempotent, no error |
| `eslint` (changed files) | clean |

No schema change: no new table and no new column, so no cascade / export-policy / RLS-allowlist registration applies.

## Follow-up filed separately (not fixed here) — #4792

`action_intents.scope_ticket_id` has the same class of gap in **both** sites, and it is worse than a dangling pointer: the generic loop re-stamps `tickets.org_id` for the moved device while `action_intents.org_id` is immutable, and `action_intents_scope_ticket_org_fk` is `(scope_ticket_id, org_id) -> tickets(id, org_id)` with **no `ON UPDATE`** clause and `DEFERRABLE INITIALLY IMMEDIATE`. A device org-move carrying a ticket that any intent (live *or* terminal) is scoped to should raise `23503` and abort the move outright. `ticketService.moveTicketOrg` already handles exactly this for the ticket-level move by tombstoning all statuses first. Left out of this PR to keep the change to the one mirror the issue asks for; filed as #4792 with the mechanism, the suggested scope, and a note that the 23503 has not been reproduced against a live server yet.

Closes #4454

### DIFF
--- a/apps/api/migrations/2026-10-06-124500-cascade-device-org-move-intent-scope.sql
+++ b/apps/api/migrations/2026-10-06-124500-cascade-device-org-move-intent-scope.sql
@@ -55,6 +55,30 @@
 -- check). Tombstoning is the strictly safer of the two: it leaves no
 -- cross-tenant device id behind at all.
 --
+-- THE STATEMENT RUNS UNDER THE CALLER'S ACCESS CONTEXT, not the definer's.
+-- SECURITY DEFINER switches the ROLE; it does not touch the `breeze.*` GUCs
+-- that breeze_current_scope() reads, and action_intents is FORCE ROW LEVEL
+-- SECURITY. Both callers that exist today set a context that satisfies
+-- breeze_has_org_access(org_id): the moveOrg route's request context spans
+-- source and target orgs (the devices UPDATE that fired this trigger already
+-- proved that), and orgMerge Phase B wraps its whole transaction in
+-- withSystemDbAccessContext, where the helper short-circuits TRUE. A future
+-- contextless caller -- an ad hoc psql fix, or a script that forgets the
+-- wrapper -- would land on scope='none' and this UPDATE would match zero rows
+-- silently.
+--
+-- Deliberately NOT instrumented with GET DIAGNOSTICS + RAISE WARNING, unlike a
+-- migration cleanup statement. Zero rows is the OVERWHELMINGLY common normal
+-- case here: the trigger fires on every device org-move and most moved devices
+-- have no intent scoped to them at all, so a warning would fire constantly and
+-- train readers to ignore it. Worse, the "did we miss rows?" probe would have
+-- to read action_intents under the SAME context that just hid them, so it
+-- cannot detect the one case it exists for. The real backstop is at release:
+-- services/actionIntents/agentReleaseAuthority.ts refuses a device whose
+-- CURRENT org_id no longer matches the intent's, so a tombstone that never
+-- fired still fails closed. This statement is defense in depth and data
+-- hygiene, not the security boundary.
+--
 -- NO HISTORICAL BACKFILL, for the reason 2026-09-29's header gives for the same
 -- omission: action_intents is FORCE ROW LEVEL SECURITY, so a bare cleanup
 -- statement here — run without a `breeze.scope` access context — is subject to
--- a/apps/api/migrations/2026-10-06-124500-cascade-device-org-move-intent-scope.sql
+++ b/apps/api/migrations/2026-10-06-124500-cascade-device-org-move-intent-scope.sql
@@ -1,0 +1,130 @@
+-- #4454 — mirror moveOrg's action_intents scope tombstone into
+-- breeze_cascade_device_org_id().
+--
+-- `apps/api/src/routes/devices/moveOrg.ts` tombstones the typed target scope of
+-- every LIVE intent aimed at a device that is leaving its org:
+--
+--     UPDATE action_intents SET scope_device_id = NULL
+--       WHERE scope_device_id = <deviceId>
+--         AND status IN ('pending_approval','approved','executing')
+--
+-- The DB-side cascade — breeze_cascade_device_org_id(), the path taken by
+-- direct-SQL / non-route callers such as orgMerge's `devices` repoint — never
+-- mirrored it. P2-2 review round 1 scoped that fix to the route only and
+-- recorded the hole as a known gap in moveOrg.coverage.test.ts. A raw
+-- `UPDATE devices SET org_id = ...` therefore left a LIVE intent still naming a
+-- device that now belongs to a different tenant: exactly the cross-tenant
+-- dangling-pointer class #3828 closed for metric_anomaly_incidents.agent_run_id
+-- and #4215 closed for ai_agent_runs.ticket_id, both of which were carried the
+-- same way until they were mirrored into this function.
+--
+-- Placement: immediately after the metric_anomaly_incidents reverse pointer and
+-- BEFORE the tickets requester detach and the generic
+-- breeze_device_child_orgid_tables() re-stamp loop — the same internal order
+-- the route uses. `action_intents` is NOT returned by that function (its device
+-- pointer is `scope_device_id`, not `device_id`, deliberately so —
+-- cascadeDelete.test.ts keys on `device_id`), so the loop can never reach these
+-- rows; and action_intents.org_id is immutable
+-- (action_intents_block_content_update()), so the intent can never follow the
+-- device. Tombstoning the pointer is the only available correct outcome.
+--
+-- LIVE STATUSES ONLY, matching the route and the schema comment on
+-- actionIntents.scopeDeviceId: a terminal-status intent is a historical record
+-- of an action already decided, and its target device at decision time is a
+-- fact, not something a future release path re-validates. Only a LIVE intent
+-- can still reach release (services/actionIntents/intentTargetScope.ts), which
+-- fails closed on a NULL scope_device_id.
+--
+-- PERMITTED BY THE IMMUTABILITY TRIGGER. action_intents_block_content_update()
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` on a1379fef6 — three reviewers in parallel: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`. Plus an independent read-only design review (Codex, `xhigh`) on the three judgement calls **before** implementing, per the advisor-quorum rule for tenancy-surface changes.

**Findings:** 0 correctness defects. 3 honesty/robustness items raised, all addressed in c5cd9738e; 0 outstanding.

- `code-reviewer` — **clean, all four focus areas.** Programmatically byte-diffed the new function body against the previous newest definition (`2026-10-04-100000-ticket-requester-contact.sql`): identical except the one added statement and one comment cross-reference that had to change because the header it pointed at is no longer in the same file. Nothing dropped or reordered — the failure mode that matters most for a `CREATE OR REPLACE`. Independently verified the "no merge fence" justification against `orgMergeRegistry.ts:137` rather than taking the PR's word for it.
- `silent-failure-hunter` — **no reachable bug.** Enumerated every writer of `devices.org_id` in the repo: only the moveOrg route and orgMerge Phase B exist, and both set an access context the policy accepts. Confirmed `SECURITY DEFINER` switches the role but not the `breeze.*` GUCs, so the caller's scope is what the tombstone runs under. Raised one **latent** gap (a hypothetical future contextless caller would silently no-op) — addressed below. Also confirmed the statement cannot *raise* and abort a device move, and that the integration test's claims about what each case can and cannot prove all hold under direct tracing.
- `pr-test-analyzer` — **coverage discriminates.** Walked the mutation list (statement deleted, wrong status set, wrong column, `AND`→`OR`, wrong placement, `OLD.id` vs `NEW.id`): none escapes both layers. Verified both new files land in **blocking** CI jobs against the actual configs, with no `describe.runIf(!!DATABASE_URL)` gate — the repo's recurring "integration test that runs in zero CI jobs" failure mode. Confirmed the hoisted resolver is a pure refactor and the 19 pre-existing assertions are behaviourally unchanged.

**What changed in c5cd9738e** (comment-only in the migration — the executable SQL is byte-identical to a1379fef6, verified by hashing the non-comment lines, which is why no re-review was run):

1. Migration header now records that the statement runs under the **caller's** access context, names the two callers and why both pass the policy, and explains why a `GET DIAGNOSTICS` + `RAISE WARNING` is deliberately **not** added: zero rows is the common normal case on this trigger (most moved devices have no scoped intent), and a "did we miss rows?" probe would read `action_intents` under the same context that just hid them — it cannot detect the one case it would exist for. The real backstop is `agentReleaseAuthority`'s current-org check, which fails closed whether or not the tombstone fired.
2. Dropped an overclaim: the integration test no longer lists `action_intents_scope_device_chk` as a hazard it guards. `scope_device_id IS NULL OR scope_kind = 'device'` is satisfied by any NULL id, so the tombstone cannot violate it and no test here proves anything about it.
3. The `breeze_app`/system-context case got its **own** before-state control instead of leaning on the superuser case's, so it stays meaningful under `-t` filtering. The trigger-body ordering assertion is now labelled a convention guard, not a behavioural one.

**Tests** (isolated stack via `pnpm test-stack up`, re-run after the fix commit):

- `deviceMoveOrgIntentScopeTombstone.integration.test.ts` — 3 passed. **Red first, on real Postgres:** with the migration moved aside, both tombstone cases failed with the live intents still naming the moved device; the terminal-status control passed in both states, as a control should.
- `moveOrg.coverage.test.ts` + `autoMigrate.test.ts` — 88 passed (the 2 new static cases were red against the unfixed trigger body).
- `deviceMoveOrgCurrency`, `agentRunMoveSemantics`, `pamDeviceMoveGuard`, `actionIntentsImmutabilityTrigger` — 53 passed.
- `orgMerge`, `orgMergeRegistry`, `ticket-move-org`, `tenantCascade` — 45 passed.
- `rls-coverage` — 100 passed. `tenant-export-policy` + `tenantExportErasureRoundtrip` — 5 passed. `integration-suite-coverage` — 5 passed.
- Migration re-applied twice through `psql` — idempotent, no error. `tsc --noEmit` clean. `eslint` clean. `check-migration-naming.sh` OK.

**Not fixed here, filed instead:** #4792 — `action_intents.scope_ticket_id` has a worse version of this gap in **both** sites (the composite FK has no `ON UPDATE`, so a device org-move carrying a ticket-scoped intent should `23503` and abort the move outright, for every status rather than just live ones). Confirmed independently by the design review; the 23503 has **not** been reproduced against a live server, which is step 0 on that issue.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4454 — both are yours.

Issue #4707, opened 2026-09-03
Mobile platform attestation before trusting is_platform_bound for L4 (#1374)


ToddHebebrand:
Tracking issue for **Mobile platform attestation before trusting is_platform_bound for L4 (#1374)** (`mobile-platform-attestation-l4`).

- Plan: `docs/superpowers/plans/security-auth/2026-09-02-mobile-platform-attestation-l4.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All seven waves merged to main (W01 #4724, W02 #4793, W03 #4968, W04 #4969, W05 #5163, W06 #5163/#5165/#5168, W07 #5647 → 869ec4a367). Unreleased. Not verified: end-to-end attestation on physical iOS/Android devices against a deployed build (needs the next mobile release + TestFlight pass); BREEZE_AUTHENTICATOR_ATTESTATION_ENFORCED flag flip is a separate rollout decision. Deferred by design: legacy single-POST endpoint retirement → #5663.

Issue #2913, opened 2026-07-28
mobileDeviceBlockedMiddleware never matches: device blocking is inert in production


ToddHebebrand:
`mobileDeviceBlockedMiddleware` (`apps/api/src/middleware/mobileDeviceBlocked.ts:47-69`) looks up `mobile_devices` by `device_id = X-Breeze-Mobile-Device-Id`. No current iOS build ever creates a row whose `device_id` equals that header, so the lookup always misses and **"block this phone" is a no-op in production**.

Why the ids never match:

- The header carries a per-install UUID minted on the phone and persisted in SecureStore (`apps/mobile/src/services/installationId.ts:20-59`).
- The app's only row-creating call is `/notifications/register` (`apps/mobile/src/services/api.ts:536`), which derives a **different** id — `push-ios-<sha256[0:48]>` (`apps/api/src/routes/mobile.ts:76-82`) — and re-pair can salt it further to `${deviceId}-${Date.now()}` (`mobile.ts:420`).
- The app never calls `POST /mobile/devices`, which is the path that would store the header value as `device_id`.

So the two identifier spaces never intersect, and the block check silently matches nothing.

Surfaced while fixing BREEZE-12/13 in #2912. That PR makes this **more** concealed rather than less: the same header previously 500'd on `POST /api/v1/authenticator/devices`, and now resolves cleanly to `null`, so the mismatch no longer produces any signal at all.

Fixing it means making the mobile client register a `mobile_devices` row keyed on the installation id (or making the push-registration path store it), so the header has something to match. Worth deciding whether the blocked check should also fail closed, or at least emit telemetry when a device presents a header that resolves to nothing — right now an inert security control looks identical to a working one.

Fold in while here: the authenticator resolution added in #2912 (`apps/api/src/routes/authenticator.ts:344-349`) does not exclude `status='blocked'` rows. Harmless today, since `authenticator_devices.mobile_device_id` is written in exactly one place and read nowhere in the API, web, portal, mobile app or agent — but a blocked device becomes reachable the moment anything joins through that column.

ToddHebebrand:
Fixed in #2983 (`09df84943`) — `mobile_devices` is keyed on the per-install id, so `mobileDeviceBlockedMiddleware` actually matches and device blocking is live. Closing; the PR used `Refs` rather than `Closes`.

Issue #1374, opened 2026-06-15
[API/Agent] Authenticator: verify mobile platform attestation before trusting is_platform_bound for L4


ToddHebebrand:
## Description

Deferred follow-up from the PR #1369 (Breeze Authenticator) review. Mobile hardware-key registration (`apps/api/src/routes/authenticator.ts`, the `/devices/mobile-hw-key/verify` handler) accepts any well-formed SPKI public key and proves only **possession** of the matching private key (an RSA-SHA256 signature over the consumed registration nonce). It cannot attest that the key is actually Secure-Enclave / StrongBox-backed, yet the row is stored with `is_platform_bound = true` unconditionally — which marks it **L4-eligible** in the assurance ladder.

A client that generated a software RSA key would therefore be recorded as a hardware-bound, L4-capable factor. This matches the documented `react-native-biometrics` handoff (CI-blind, manual on-device verification), so it is an **accepted limitation today**, not a live defect — but the unconditional `is_platform_bound=true` overstates assurance and would become load-bearing once L4 (critical-tier) enforcement actually gates on it.

## Current Mitigation (shipped)

`174b4b5b` documented on the schema (`apps/api/src/db/schema/authenticatorDevices.ts`, the `isPlatformBound` column) that for `mobile_hw_key` this flag is **client-asserted, not server-attested**, so future L4 gating must not silently trust it. Today nothing enforces L4, so there is no exploitable gap yet.

## Proposed Fix

Before L4/critical enforcement ships for the mobile factor, do one of:
1. Accept + verify a platform **attestation** object at mobile registration (iOS App Attest / Android Key Attestation) and only then set `is_platform_bound=true`; or
2. Treat unattested mobile keys as **L2-only** (not L4-eligible) — derive `is_platform_bound` from a verified attestation, default false otherwise.

## Affected Files

- `apps/api/src/routes/authenticator.ts` (mobile-hw-key register/verify)
- `apps/mobile/src/services/hardwareSigner.ts` (would need to surface an attestation object)
- `apps/api/src/db/schema/authenticatorDevices.ts` (`isPlatformBound` semantics)

## Reported By

PR #1369 review (code-reviewer / api-security). Schema caveat shipped in `174b4b5b`. Pairs with the mobile native-module on-device verification handoff.

ToddHebebrand:
[...]

PR #2983, opened 2026-08-01
fix(mobile): key mobile_devices on the per-install id so device blocking works (#2913)

ToddHebebrand:
`mobileDeviceBlockedMiddleware` looks up `mobile_devices` by the id the phone presents in `X-Breeze-Mobile-Device-Id` (or the signed `mdid` claim minted from it at login). Nothing ever created a row with that id — the app's only row-creating call, `POST /mobile/notifications/register`, derived `push-<platform>-<sha256[0:48]>` from the push token, and the app never calls `POST /mobile/devices`. The two id spaces never intersected, so **"block this phone" matched nothing and the lockout was inert in production.**

Refs #2913

## The design decision: key the server on the installation id (option B)

The issue left the choice open: teach the client to register a row under the installation id, or teach the registration path to store the header value.

**Chose the server side.** The decisive factor: `apps/mobile/src/services/api.ts:225-237` already attaches the header to **every** request, and `registerPushToken` already runs at launch. So keying the server on that value makes blocking work on **binaries already in the field** — no App Store build, no native rebuild. The client-side option would leave every phone in the field inert until users updated, and would create a *second* row per phone (one push-derived, one installation-keyed) unless it also handled the existing rows.

Codex (`gpt-5.5`, xhigh, read-only) reviewed the design independently and agreed on this call, on the in-place rewrite, and on not failing closed. It disagreed on one point — see "cross-user" below — and surfaced the `/authenticator` mount gap. Both are folded in.

**No orphans.** When the installation-id row doesn't exist yet but the caller's push-derived row does, that row's `device_id` is rewritten **in place**. The uuid PK is preserved, so notification preferences and every FK pointing at it (`push_notifications`, `mobile_sessions`, `authenticator_devices.mobile_device_id`) survive the re-key. No backfill migration needed — adoption happens on the next push registration.

The push-derived id is not deleted; it remains the key for callers that send no header (legacy clients, non-mobile callers) and the id we adopt *from*.

## Fail closed, or report? — report

**Not fail closed.** A device id that resolves to no row is legitimate during onboarding (the first calls land before registration), and failing closed would lock every install out of the app until it re-registered. Instead the miss is now a rate-limited Sentry warning tagged `signed-claim` vs `header`, so an inert control stops being indistinguishable from a working one. A later fail-closed switch would need: adoption confirmed across the fleet (the telemetry above going quiet is the signal), an exemption for pre-registration onboarding calls, and a recovery path — see the open question below.

[...]

PR #4793, opened 2026-09-03
W02: challenge/verify mobile registration protocol + ES256 (#4709)

### LINKED ISSUE #4709: W02: Challenge/verify registration protocol (API): /devices/mobile/challenge + /verify, domain-separated transcript, ES256/P-256 alongside RS256
Wave W02 of #4707 — Challenge/verify registration protocol (API): /devices/mobile/challenge + /verify, domain-separated transcript, ES256/P-256 alongside RS256
Branch: `feature/4707-mobile-platform-attestation-l4/wave-4709`

ToddHebebrand:
Wave **W02** of feature #4707 — *Mobile platform attestation before trusting `is_platform_bound` for L4* (origin #1374).

Plan: `docs/superpowers/plans/security-auth/2026-09-02-mobile-platform-attestation-l4.md`

Closes #4709

## What this wave does

W01 closed the reported gap by making the L4 predicate refuse an unattested basis. This wave builds the road back to L4: a **two-step, challenge-bound mobile registration** that a real platform attestation can be threaded through, plus server support for **P-256 / ES256** signing keys.

**No attestation verifier is wired yet.** `verifyPlatformAttestation` returns `unattested` until W03 (Apple App Attest) and W04 (Android Key Attestation) land, so this wave is safe to ship alone: a device registered through the new protocol works at L2/L3 and simply cannot reach L4.

### 1. ES256/P-256 alongside RS256

Every mobile approver key today is RSA-2048 (`react-native-biometrics`), and `mobileHwKey.ts` hard-coded `RSA-SHA256`. The Apple Secure Enclave holds **only** 256-bit EC private keys, so an attested platform-bound key is necessarily ES256.

`verifyMobileSignature` now takes an explicit `alg`, read from the **device row** (`authenticator_devices.public_key_alg`) — never from the request body, because a client-chosen algorithm is an algorithm-confusion vector. An unrecognised stored label fails closed rather than defaulting to RSA.

Hardening beyond the plan, each covered by a test:
- **`ES256` requires the P-256 curve specifically**, not merely "some EC key" — `crypto.verify` would happily accept a P-384 key, and the `ios_se_p256_app_attest` basis would then be describing a key the Secure Enclave cannot hold.
- **`RS256` requires a ≥ 2048-bit modulus.**
- The `public_key_alg` domain is enforced by a **DB CHECK**, not only in application code: a row nothing can verify should not be storable.

### 2. Registration attempt + domain-separated transcript

`registrationTranscript()` is the one digest that App Attest's `clientDataHash`, Android's `setAttestationChallenge`, and the registration proof-of-possession signature are **all** computed over. One digest, three bindings — which is what makes "this attestation vouches for *this* key in *this* attempt" checkable rather than asserted.

- Domain-separated (`breeze.authenticator.mobile-register.v1`) and newline-delimited. Field-boundary confusability is asserted, not assumed.
- **The exact pre-image is pinned by a test**, because W05/W06 mint the same digest on-device — a silent reordering here would break every phone in the field rather than failing CI.
- Attempts live in Redis for 300s and are consumed with `GETDEL`, so a replayed `attemptId` finds nothing. They record the user and platform they were issued to.

### 3. The two endpoints

```
POST /api/v1/authenticator/devices/mobile/challenge  -> 200 { attemptId, challenge, expiresAt }
POST /api/v1/authenticator/devices/mobile/verify     -> 201 { success, device }
```

Proof-of-possession moves **into** registration: the client signs the transcript under a biometric prompt and the server records `possession_verified_at`. `last_used_at` keeps its existing meaning (null = never used for a real approval), so the UI's "pending" badge still works.

**The ORDER of the checks in `/verify` is the security property**, and each step is pinned by a test that asserts what did *not* happen:

| # | Check | On failure | Grant burned? |
|---|---|---|---|
| 1 | strict schema parse | 400 | no side effects at all |
| 2 | consume the attempt | 400 `registration_attempt_expired` | **no** |
| 3 | attempt ownership | 403 + audit `attempt_user_mismatch` | **no** |
| 4 | attempt platform match | 403 + audit `attempt_platform_mismatch` | **no** |
| 5 | registration PoP | 401 `registration_pop_invalid` + audit | **no** |
| 6 | platform attestation | sets the basis (W03/W04) | — |
| 7 | consume the register grant | 403 | single-use, only once everything else passed |
| 8 | insert | | |

A client that fails attestation therefore does not burn its single-use register grant and can retry with a fresh challenge instead of redoing the whole step-up. The grant is validated **non-consuming** at `/challenge`, the same pattern as `/devices/webauthn/options`.

Also in this wave:
- **A basis claiming a verified attestation must carry the digest of the key that attestation bound.** If the digest cannot be derived, the row **downgrades to an honest `unattested`** rather than 500ing on `authenticator_devices_attested_basis_chk` or storing a platform-bound claim with nothing behind it. Forward-looking guard for W03/W04, tested.
- **Both routes are per-user rate limited** (10 / 300s). `routes/authenticator.ts` had no rate limiting at all, and both new routes mint or consume server-side state reachable with nothing more than a valid bearer token.
- The `X-Breeze-Mobile-Device-Id` resolution is extracted into **one helper shared with the legacy route**, so the ownership + `status='active'` predicates (Sentry BREEZE-12/13, #2913) cannot drift between the two.
- `toPublicDevice` now returns `platformBoundBasis`, so a client can tell an attested enrolment from an unattested one.

## Tenancy / registration lists

`authenticator_devices` is **shape 6 (user-id scoped)**, already `ENABLE` + `FORCE` with `authenticator_devices_user_scope`, already in `USER_ID_SCOPED_TABLES`. **Columns only** — no new policy, no new table, no allowlist entry.

Verified by grep, not assumed — zero matches for `authenticator_devices` in all three files:

```
$ grep -rn "authenticator_devices" apps/api/src/services/tenantCascade.ts \
    apps/api/src/services/tenantExportPolicyRegistry.ts \
    apps/api/src/routes/devices/core.ts
(no output)
```

No `org_id` and no `device_id` column (`mobile_device_id` FKs `mobile_devices`, not `devices`), so the table is in none of `CORE_ORG_CASCADE_DELETE_ORDER` / `CORE_DEVICE_CASCADE_DELETE_TABLES` / `CORE_DEVICE_ORG_DENORMALIZED_TABLES`. `CORE_TENANT_EXPORT_POLICY` classifies columns only for tables in the org-cascade list, so this `ADD COLUMN` needs no export-policy entry either. Rows are reaped by `user_id … ON DELETE CASCADE`, and `users` is itself org-cascaded. `tenantCascade` + `tenant-export-policy` integration suites run green regardless.

## Migration

`2026-10-06-120000-authenticator-public-key-alg.sql` — sorts after the newest committed migration (`2026-10-06-110100`). Idempotent (`ADD COLUMN IF NOT EXISTS` + `pg_constraint` existence check), no inner `BEGIN;`, no backfill `UPDATE` (every existing row is RSA, so the `DEFAULT 'RS256'` is already the correct classification — hence no row-count `RAISE`).

## Verification

All run in the foreground against this worktree's own private test stack.

| Suite | Result |
|---|---|
| `mobileHwKey.test.ts` + `authenticatorAssurance.test.ts` | **71 passed** |
| `authenticatorAttestation.test.ts` (new) | **16 passed** |
| `routes/authenticator.test.ts` | **56 passed** (20 new) |
| `packages/shared` `validators/authenticator.test.ts` | **28 passed** (13 new) |
| `db/autoMigrate.test.ts` (migration ordering) | **67 passed** |
| `routes/pam` + `routes/approvals` + `mobileDeviceIdentity` | **209 passed** |
| web `stores/authenticator` + `ApproverDevicesSection` | **36 passed** |
| **Integration**: `authenticatorPublicKeyAlg` (new) + `authenticatorAttestation` | **15 passed** |
| **Integration**: `rls-coverage`, `authenticatorRls`, `authenticatorMobileDeviceResolution` | **23 passed**, 3 skipped |
| **Integration**: `authenticator-assurance-checks`, `tenantCascade`, `tenant-export-policy` | **22 passed**, 3 skipped |
| `tsc --noEmit` (`apps/api`) | clean |
| `tsc --noEmit` (`packages/shared`) | clean for changed files (3 pre-existing errors in `quotes.test.ts`, untouched) |
| `pnpm db:check-drift` | no drift, 647 migrations match the ledger |

Red-first throughout: every behaviour above was watched failing before it was implemented (9 red in `mobileHwKey`, 17 red in `authenticatorAttestation`, 13 red in the shared validators, 20 red in the route suite).

## Not in this wave (by design)

- **No attestation verifier.** W03 (Apple App Attest) and W04 (Android Key Attestation + Play Integrity) replace the `verifyPlatformAttestation` stub. Until then every registration through `/verify` is `unattested`.
- **No client change.** The mobile app still uses the legacy `POST /devices` route; W05/W06 wire the native modules and the two-step flow. The new endpoints are unused until then.
- **`registrationTranscript` has not yet moved to `packages/shared`.** The plan moves it in W05 Task 2 so the client can reuse it; keeping it server-side for now avoids a shared-package change nothing consumes yet. Its pre-image is pinned by a test so the move is mechanical.
- **The legacy `POST /devices` route is unchanged** and still registers `unattested` (W01 behaviour). Deprecating it belongs with the client cutover.

Closes #4709

---

https://claude.ai/code/session_01MEh2twLyYFx4JaqMUFZEox


ToddHebebrand:
## Review round — 3 Sonnet reviewers, findings acted on

One independent review round per the repo's rigor rules for an auth surface. Three specialists ran in parallel against `origin/main...HEAD`: **code-reviewer**, **silent-failure-hunter**, **pr-test-analyzer**.

### Outcome

| Reviewer | Critical | Important | Acted on |
|---|---|---|---|
| code-reviewer | 0 | 0 | — |
| silent-failure-hunter | 1 | 2 | all 3 |
| pr-test-analyzer | 0 | 3 | all 3 |

The code-reviewer found nothing at critical or important severity and independently re-verified the things most likely to be wrong here: algorithm confusion is closed in both directions, nothing from `body.attestation` reaches a stored platform-binding field, `GETDEL` really does make the attempt single-use, the grant-consume ordering holds, and the cascade/export-list claim (zero grep matches) is true.

The other two converged on the same theme, and it was a fair hit: **the fail-closed *direction* was right everywhere, but several failures were invisible.** A denial that an operator cannot distinguish from an attack is exactly how the plan's stated residual risk — "a technician locked out of approving an incident-response elevation from their phone" — turns into an unsolvable ticket. Fixed in 64532af32.

### Findings fixed

**1. `consumeRegistrationAttempt` swallowed a corrupt Redis payload into a bare `null`** *(rated CRITICAL by silent-failure-hunter)* — which `/verify` reports to the caller as `registration_attempt_expired`. But `attemptId` is a `randomUUID` keyspace we wrote ourselves with `JSON.stringify` moments earlier, so a parse failure means Redis corruption, a namespace collision on `authenticator-attest:*`, or a serialization bug — never caller behaviour. Now `captureException` + `console.error`. A plain miss (expired/unknown) stays silent, and a test pins that distinction so the fix cannot degrade into crying wolf on every expiry.

**2. `verifyMobileSignature` collapsed six distinguishable causes into one silent `false`** *(HIGH)* — malformed DER, wrong key type, wrong curve, undersized modulus, malformed signature, genuine mismatch. Split them by who can cause them:
- **signature mismatch → stays silent.** Routine, and a caller can generate it at will; logging it is noise and a log-flood vector.
- **key shape wrong for the declared algorithm → logs.** On the approval path both the key *and* its label come from our own device row, so this is never routine — it means bad data, a bad migration, or a Node crypto behaviour change, and *every future approval from that device fails forever* with nothing in any log to explain it. No key material or signature bytes are logged, only the shape mismatch. The registration path, where the key is caller-supplied, is rate limited 10/300s per user.

**3. The attestation downgrade had no signal and no test** *(flagged independently by BOTH reviewers)* — when a verifier says "verified" but the bound-key digest will not derive, the row safely downgrades to `unattested`, but the only trace was a boolean buried in a `result: 'success'` audit row's `details`. Nobody queries that. It now raises `captureException` and its own `auth.authenticator.device.register.attestation_downgraded` **failure** audit event, matching the `.denied` convention its sibling checks already use.

  Worth stating plainly: this path is **dead today** — the W02 stub always returns `verifiedAt: null`. It goes live the moment W03/W04 wire a real verifier, which is exactly the moment nobody will remember it exists. That is the argument for hardening it now rather than later.

### Test gaps closed

**Every new assertion below was mutation-checked** — I inverted the production code and confirmed the test went red before keeping it, rather than trusting a test written after the code.

- **The platform-mismatch branch was the one early return not *proven* to leave the grant unburned** (pr-test-analyzer, rated 6). Every sibling denial asserts `enforceApproverRegisterStepUp` was not called; this one inferred it from "nothing was inserted". True by statement order today — which is precisely what a refactor changes without any other test noticing. Now asserted directly, plus its audit reason. *Mutation check: changing `reason` to a different string → red.*
- **The downgrade test checked 2 of 5 attestation-derived columns** (rated 6). All five now assert they fall back together. Checking two would miss a row labelled `unattested` that nonetheless carries `attestationEvidence: { appId: … }` — forensically worse than either honest state. *Mutation check: `if (…)` → `if (false)` on the downgrade block → red.*
- **Nothing exercised `registrationTranscript` and `verifyMobileSignature` together with real crypto.** Each was proven in isolation and the route test proved they were *wired*, but with both mocked — so the binding held only by inference across three files. Four new tests build a real transcript, sign it with a real P-256 key, and verify through the real verifier: accepts the genuine PoP, and rejects **a different challenge** (replay), **a substituted key** (the SPKI is inside the transcript), and **a different declared algorithm**. This is the exact contract W05/W06 must reproduce on-device.

### Findings considered and NOT acted on

- **Concurrency test for two `/verify` calls racing one `attemptId`.** Single-use is already proven against real `GETDEL` semantics, which is where the atomicity actually lives; a `Promise.all` in single-threaded Node would assert nothing further.
- **Negative mobile-header cases re-tested on the new route.** `resolveOwnedMobileDeviceId` is one shared function with full coverage on the legacy route — the whole point of extracting it was that the two call sites cannot drift.
- **Route-level Redis-down test.** The service-layer throw is pinned with a real assertion, and the global `onError` (which logs + `captureException`s + returns 500) is uniform pre-existing infrastructure.

| | |
|---|---|
| affected API unit suites (7 files) | **414 passed** |
| integration: `authenticatorPublicKeyAlg` + `authenticatorAttestation` | **15 passed** |
| RLS coverage (`vitest.config.rls-coverage.ts`) | **100 passed** |
| RLS (`vitest.config.rls.ts`) | **26 passed**, 5 skipped |

Not merging and not closing the issue — those are yours.

ToddHebebrand:
## Rebase note — migration renamed

`origin/main` gained `2026-10-07-100000-devices-reboot-schedule-columns.sql` after this branch was cut, and the pre-push guard correctly rejected the original filename: it would have replayed **ahead** of that migration on a fresh database.

Renamed `2026-10-06-120000-…` → **`2026-10-07-110000-authenticator-public-key-alg.sql`**. Safe because this migration is unmerged, so no `breeze_migrations` ledger anywhere keys on the old name.

Swept the one reference that would otherwise have been an `ENOENT` several minutes into **Integration Tests** rather than a compile error — `authenticatorPublicKeyAlg.integration.test.ts` reads the migration by path for its idempotency replay.

Re-verified against a **freshly recreated** test database (the old one's ledger still held the previous filename, which is exactly what the integration global-setup guard caught):

| | |
|---|---|
| integration: `authenticatorPublicKeyAlg`, `authenticatorAttestation`, `authenticatorRls`, `tenantCascade`, `tenant-export-policy` | **27 passed** |
| RLS coverage | **100 passed** |
| affected API unit suites | **219 passed** |
| `pnpm db:check-drift` | no drift, 648 migrations match the ledger |

Also in this push: `ApproverDevice` in `apps/web/src/stores/authenticator.ts` now declares the `platformBoundBasis` field the API started returning (optional, type only — the honest attested/unattested badge and its 9-locale copy stay in W07).

ToddHebebrand:
## CI triage — three items, all resolved

**1. `BINARY_SOURCE=github download path` — flake, not this PR.** Failed with `404 Version not found for the specified platform and architecture`; that step resolves published GitHub release artifacts, i.e. external state. The same branch passed this job twice earlier in the day. Re-ran it: **success**. Every file in this diff is authenticator/migration-scoped — `git diff origin/main...HEAD --name-only` has zero overlap with the agent-download path.

**2. `CodeQL` — a real `js/insufficient-key-size` (high) on my own test fixture.** The test mints a 1024-bit RSA key to prove `verifyMobileSignature` **rejects** it, so the rule fires correctly by its letter while the intent is the opposite.

An inline `// codeql[js/insufficient-key-size]` (matching the existing precedent in `recoveryAuthorizationSubject.ts` and `clientAssertion.ts`) **did not take** — CodeQL re-flagged the same line on the head that carried the comment. Rather than escalate to a repo-wide `paths-ignore` or lean on a Security-tab dismissal, I removed the sink: the test now uses a **committed SPKI plus a genuine RSA-SHA256 signature** for a throwaway 1024-bit key. Both are public values; the private key was discarded at generation. The regeneration recipe is in the comment.

The assertion got **stronger**. It now first proves the vectors really are a 1024-bit key and a signature that *verifies* against it — so the test cannot silently degrade into "malformed input returns false" — and only then asserts the verifier refuses it, purely on the modulus floor. Mutation-checked: dropping `MIN_RSA_MODULUS_BITS` to 512 turns it red.

**3. `Test API` — the site-scope coverage contract, and it caught something real about this PR.** That contract is shrink-only: it fails when a baselined handler *stops* being flagged, so the ratchet tightens instead of carrying dead exemptions forward.

Extracting the `mobile_devices` lookup into `resolveOwnedMobileDeviceId()` — done so the new attested `/devices/mobile/verify` cannot drift from the legacy route — means the file-local scanner no longer attributes the query to either handler, making both `routes/authenticator.ts:POST /devices` baseline entries stale. Removed. Exactly the shape of the `routes/mobile.ts:POST /notifications/register` removal the file already documents from #2983.

To be explicit about what did **not** change: the query, its predicates (`device_id` + `user_id` + `status='active'`), and the exemption's *reason* are identical — it resolves a mobile/OAuth device row (`mobile_devices`, no `site_id`/`org_id`) already narrowed by `userId`, which is tighter than site-scope, so a site gate was never meaningful here. Only the detector's visibility changed. **`/devices/mobile/verify` was not added as an exemption** — it simply isn't flagged, so the exempt set is strictly smaller than before this PR.

`pnpm --filter=@breeze/api test:site-scope-coverage` → **7 passed** locally.

PR #4794, opened 2026-09-03
fix(api): add remediate_vulnerability to DEVICE_COMPLETE_TARGET_TOOLS (#4452)

### LINKED ISSUE #4452: AI agents: add remediate_vulnerability to DEVICE_COMPLETE_TARGET_TOOLS
Sweep-proposed `remediate_vulnerability` actions (wave P2-2) create device-scoped action intents, but the tool isn't listed in `DEVICE_COMPLETE_TARGET_TOOLS`, so scoped-intent target-scope resolution can't short-circuit for it the way it does for tools already on that list — an unnecessary resolver round-trip on every release/decide.

ToddHebebrand:
## Summary

`DEVICE_COMPLETE_TARGET_TOOLS` (`apps/api/src/services/actionIntents/intentApprovers.ts`) is the hand-verified allowlist of tools whose device target can be fully resolved from their registered `deviceArgs`, used by `resolveIntentTargetScope` to narrow approver fan-out for agent-proposed action intents to the target device's site. `remediate_vulnerability` was missing from it — sweep-proposed remediation intents (wave P2-2) create device-scoped action intents, but every one of them fell through to `{kind:'indirect'}` and fanned out to unrestricted-site approvers instead of the scoped device's site, on every release/decide.

## The catch

`remediate_vulnerability`'s `deviceId` argument is **optional**, unlike the three tools already in the set (`execute_command`, `manage_services`, `run_script`), whose device args are required. The handler only enforces `deviceId` as a *complete* pin when the caller actually supplies it (every cited `deviceVulnerabilityIds` finding must belong to that device, or the whole call is refused with `finding_device_mismatch`); when omitted — legal for interactive chat — findings can legitimately span multiple devices.

Simply adding the tool to the set would have made `resolveIntentTargetScope` fall back to `target.deviceId` (the resolved intent/run device) as a stand-in "complete" target even when `deviceId` was omitted from the actual call — silently narrowing approver eligibility to the wrong site for a call that could touch other devices/sites.

## The fix

- Added `remediate_vulnerability` to `DEVICE_COMPLETE_TARGET_TOOLS` with a hand-verification entry documenting the optional-arg semantics.
- Adapted `resolveIntentTargetScope`: the resolved intent/run device (`target.deviceId`) is now unioned in only as a *widener* on top of an already non-empty args-derived device set — never as the sole source of a `'devices'` resolution. This is a no-op for the three existing tools (their args are required, so the args-derived set is never empty) and correctly falls back to `indirect` for a `remediate_vulnerability` call that omits `deviceId`.

## Test coverage

- `intentApprovers.deviceTargets.contract.test.ts`: new test derives the tool names from the real `SweepProposedAction` zod schema (`@breeze/shared/validators`, not hardcoded) and asserts every sweep-proposable tool is present in `DEVICE_COMPLETE_TARGET_TOOLS` — the next sweep-proposable tool added without updating this set now fails CI instead of silently degrading to indirect fan-out.
- `intentApprovers.test.ts`: two new cases for `resolveIntentTargetScope` — `remediate_vulnerability` resolves to `{kind:'devices'}` when `deviceId` is pinned (the sweep path), and falls back to `{kind:'indirect'}` when `deviceId` is omitted even though the run has a bound device (the regression this fix closes). Updated the `DEVICE_COMPLETE_TARGET_TOOLS` set-equality test.

Closes #4452

## Test plan

- [x] `cd apps/api && npx vitest run src/services/actionIntents` — 513 tests passed (25 files)
- [x] `npx vitest run src/services/aiToolsVulnerability.test.ts src/services/aiAgents/sweepFindings.test.ts` — 41 tests passed
- [x] `NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit` — clean
- [x] `npx eslint` on all three changed files — clean

### DIFF
--- a/apps/api/src/services/actionIntents/intentApprovers.deviceTargets.contract.test.ts
+++ b/apps/api/src/services/actionIntents/intentApprovers.deviceTargets.contract.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it } from 'vitest';
+import { sweepProposedActionSchema } from '@breeze/shared/validators';
 import { aiTools } from '../aiTools';
 import { DEVICE_COMPLETE_TARGET_TOOLS } from './intentApprovers';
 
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer, silent-failure-hunter, pr-test-analyzer (parallel, Sonnet)
**Findings:** 0 critical, 0 important. 2 low-priority suggestions, neither actionable in this PR's scope:
- silent-failure-hunter: a pre-existing bare `catch { targetScope = { kind: 'indirect' } }` in `intentService.ts`'s decide-time revalidation swallows non-"device gone" errors with no telemetry. File is untouched by this PR (`git diff origin/main...HEAD -- intentService.ts` is empty) — flagged as a separate follow-up, not a defect introduced here.
- code-reviewer: the new contract test accesses zod's discriminated-union internals via `as unknown as {...}` — sound against the pinned `zod@^4.4.3`, but worth re-verifying on a future zod major bump.
**Tests:** `cd apps/api && npx vitest run src/services/actionIntents` — 513 passed (25 files). `npx vitest run src/services/aiToolsVulnerability.test.ts src/services/aiAgents/sweepFindings.test.ts` — 41 passed. `NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit` — clean. `npx eslint` on all 3 changed files — clean.
**Test-analyzer note:** independently reverted the one-line resolver fix and reran the suite — confirmed exactly the new "falls back to indirect when deviceId is omitted" case fails without the fix (31/32 other tests stay green), proving the test is discriminating, not tautological.

Issue #4188, opened 2026-08-28
W01: P2-1 Alert verdicts — Tier-2 agent intents, outcome tools, verdict profile, ai_alert_verdicts, group-created event + subscribers, badge UI


ToddHebebrand:
Wave W01 of #4187 — P2-1 Alert verdicts — Tier-2 agent intents, outcome tools, verdict profile, ai_alert_verdicts, group-created event + subscribers, badge UI
Branch: `feature/4187-ai-agents-p2/wave-4188`

ToddHebebrand:
**P2-1 PR A opened: #4220** (foundations + verdict lane persistence; inert until PR B adds triggers). PR B (Tasks 11–16: group-created event, verdict subscribers, ungrouped delay job, alerts API `aiVerdict`, badge UI) continues on `feature/4187-ai-agents-p2/wave-4188-b`, stacked on #4220; it will be rebased onto main after #4220 merges and will `Closes #4188`.

**Controller rulings made during execution** (each with what it costs if wrong):
1. Started from `main` before wave 6.3 merged; later rebased over 6.3 + 6.4 (migration renamed `2026-09-21-…`). Cost: none realised.
2. Task 1 verified the migration statically; live apply + RLS forge done at wrap on an isolated test DB (forge fails correctly).
3. DTO `alertVerdict` is additive/nullable → **no** `AI_AGENT_RUN_DTO_SCHEMA_VERSION` bump; rule written into `aiAgentRuns.ts`. Cost: a strict client rejecting unknown keys (none exists).
4. Agent-originated Tier-2 approvals carry `riskTier 'medium'` → **L2 assurance floor** (one-click, reversible ops). Cost: approvable from an MFA session rather than a stronger factor.
5. `VERDICT_TOOL_ALLOWLIST` is a **floor**, not an intersection (all read-only tools that already bypass the allowlist on full runs; intersecting with the empty default allowlist produced evidence-free verdicts). Cost: none.
6. SDK `extraTools` bypass the run hooks → `createBreezeMcpServer` wraps them (plan defect found in review).
7. Verdict runs skip run-finished notifications and fix-watches (badge is the surface). Cost: a notification a partner might want — trivially re-enabled.
8. A verdict run that created a pending intent finishes `awaiting_approval` (status computed after persistence); persistence skipped if the run is no longer `running`.
[...]

PR #4796, opened 2026-09-03
fix(ai-agents): error on unknown onlyTools names instead of silently dropping them (#4447)

### LINKED ISSUE #4447: [AI agents] onlyTools silently drops unknown tool names
Verdict-run tool pinning (`onlyTools`, wave P2-1) silently drops any name that doesn't match a registered tool instead of surfacing an error or warning — a typo or a renamed tool schema would silently narrow (or empty) the allowlist with no signal to the developer or operator.

ToddHebebrand:
## Summary

`createBreezeMcpServer`'s `options.onlyTools` filter (verdict/sweep tool pinning, F2 fix wave P2-1) used to silently drop any name in `onlyTools` that matched no registered tool:

```ts
const registeredTools = options?.onlyTools
  ? tools.filter((t) => options.onlyTools!.has(t.name))
  : tools;
```

A typo in a hardcoded profile allowlist, or a tool renamed in the registry without updating the allowlist, would quietly narrow (or empty) the exposed tool set for a verdict/sweep run — with no signal to the developer or operator. Source: #4188 rulings comment, part of the #4187 follow-up sweep.

## Fix

After filtering, diff `onlyTools` against the matched tool names. For any unmatched name:

- Build an error message naming the unknown name(s) and, for each, the 3 nearest registered tool names by Levenshtein edit distance (surfaces the likely typo).
- **Choice of throw-everywhere vs. throw-in-dev-only:** `git grep -n onlyTools origin/main -- apps/api/src` shows the only caller is internal (`aiAgents/runLoop.ts`, computed from hardcoded `profileAllowlist`/`SWEEP_TOOL_ALLOWLIST` constants) — never request input. So an unmatched name is always a programming error, not a runtime condition. Given that, I went with **throw outside production (`NODE_ENV !== 'production'`), log + Sentry-capture in production** rather than throwing unconditionally everywhere: throwing in test/dev catches the bug before it ships (and is exercised directly by the new tests, since vitest runs with `NODE_ENV=test`), while in production a live verdict/sweep run degrades to the matched subset instead of hard-failing outright — but it's now loud (console.error + a new bounded Sentry event code `ai_agent_onlytools_unknown_name`), not silent.

## Tests (red first)

Added to `createBreezeMcpServer options.onlyTools (Task 16c F2)` in `aiAgentSdkTools.test.ts`:
- all-known names → registers exactly that set, unchanged, does not throw
- one unknown name → throws (`NODE_ENV=test`) naming the unknown name
- error message contains the unknown name
- production (`NODE_ENV=production`) → does not throw, still narrows to the matched subset, `console.error` + `captureMessage` (with the new event code) both fire

All 3 new assertions were red before the fix (confirmed), green after.

## Verification

- `cd apps/api && npx vitest run src/services/aiAgentSdkTools` — 13/13 passed
- `cd apps/api && npx vitest run src/services/aiAgentSdkTools src/services/sentryEventCodes.test.ts` — 76/76 passed (9 files, substring match also picked up related suites)
- `cd apps/api && npx vitest run src/services/aiAgents/runLoop` — 151/151 passed (the only internal caller of `onlyTools`, unaffected)
- `cd apps/api && npx tsc --noEmit` — clean
- `npx eslint` on the 3 changed files — clean

Closes #4447

### DIFF
--- a/apps/api/src/services/aiAgentSdkTools.test.ts
+++ b/apps/api/src/services/aiAgentSdkTools.test.ts
@@ -14,12 +14,14 @@
  */
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { createBreezeMcpServer, wrapExtraToolWithHooks } from './aiAgentSdkTools';
-import { buildOutcomeSdkTools, type SdkTool } from './aiAgents/outcomeTools';
+import { buildOutcomeSdkTools, isOutcomeTool, type SdkTool } from './aiAgents/outcomeTools';
 import {
   createAgentRunPostToolUse,
   createAgentRunPreToolUse,
   type AgentRunOutcome,
 } from './aiAgents/runLoop';
+import { VERDICT_TOOL_ALLOWLIST } from './aiAgents/verdictProfile';
+import { SWEEP_TOOL_ALLOWLIST } from './aiAgents/sweepProfile';
 import type { AiAgentRunProfile } from '@breeze/shared';
 import { hasDbAccessContext, __runInDbContextForTests } from '../db';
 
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (parallel, Sonnet).

**Findings:** 4 real findings raised → all addressed in `39e47496e`; 0 outstanding.
- silent-failure-hunter (Important): the new guard was never exercised against the real `VERDICT_TOOL_ALLOWLIST`/`SWEEP_TOOL_ALLOWLIST` (every `runLoop.*.test.ts` suite mocks `aiAgentSdkTools` away) → added a regression-gate test calling the real, unmocked `createBreezeMcpServer` with both real allowlists.
- silent-failure-hunter (Important): docstring overstated production visibility ("cannot slip by unnoticed") when Sentry is unconfigured (self-hosted default) → softened to note `captureMessage` is a documented no-op with no `SENTRY_DSN`, so `console.error` is the only guaranteed signal there.
- pr-test-analyzer (Important): the "nearest valid name" suggestion — the actual point of the fix — was untested → added a typo-suggests-the-right-name case and a no-close-match case.
- pr-test-analyzer (Important): multiple unknown names in one call was untested (join-logic risk) → added a two-unknown-names case, asserting both are named and the one valid name is never misreported as unknown.

Also picked up two cosmetic nits from code-reviewer inline (no separate finding needed): aligned the new `sentryEventCodes.ts` section-header width to match the other sections, and switched the production-path test to `vi.stubEnv` (the file's existing convention) instead of a hand-rolled `process.env` save/restore.

Declined: silent-failure-hunter's suggestion to add an unknown-tool-name Sentry tag — kept the existing pattern of a generic, bounded `event_code` tag with full detail only in `console.error`, consistent with this file's documented "tags are the only channel; keep them low-cardinality" Sentry contract (`sentry.ts`).

**Tests:** `cd apps/api && npx vitest run src/services/aiAgentSdkTools src/services/sentryEventCodes.test.ts src/services/aiAgents/runLoop` — 231/231 passed across 13 files (single-fork). `npx tsc --noEmit` clean. `npx eslint` on the 3 changed files clean.

Issue #4767, opened 2026-09-03
W06: Stop/Force-stop affordances, polling, 8-locale i18n, operator docs, E2E


ToddHebebrand:
Wave W06 of #4761 — Stop/Force-stop affordances, polling, 8-locale i18n, operator docs, E2E
Branch: `feature/4761-cancel-running-script-or-automation/wave-4767`

ToddHebebrand:
Completed via PR #5033

E2E spec (script-cancel.spec.ts) written but not executed against a live stack. Follow-up #5040: cancelState missing from executions GET routes. DeviceScriptHistory.tsx has no Stop button (second surface, out of scope).

Issue #4766, opened 2026-09-03
W05: Automation run cancel: enums + devices_cancelled, atomic dispatch fence, fan-out keyed on automation_run_id


ToddHebebrand:
Wave W05 of #4761 — Automation run cancel: enums + devices_cancelled, atomic dispatch fence, fan-out keyed on automation_run_id
Branch: `feature/4761-cancel-running-script-or-automation/wave-4766`

ToddHebebrand:
Completed via PR #5030

Fence is check-before + compensate-after (not the plan's held CAS) to avoid pinning a pooled connection across WS sends; automation.failed published alongside automation.cancelled when devicesFailed > 0 (Fable + Codex xhigh quorum). Migration slot 2026-10-11-140300.

Issue #4764, opened 2026-09-03
W03: The five closers: result-handler CAS, cancellation sweep, cancel-command expiry, late-result recovery, audit


ToddHebebrand:
Wave W03 of #4761 — The five closers: result-handler CAS, cancellation sweep, cancel-command expiry, late-result recovery, audit
Branch: `feature/4761-cancel-running-script-or-automation/wave-4764`

ToddHebebrand:
Completed via PR #4967

Issue #4761, opened 2026-09-03
Cancel a running script or automation (#3525)


ToddHebebrand:
Tracking issue for **Cancel a running script or automation (#3525)** (`cancel-running-script-or-automation`).

- Plan: `docs/superpowers/plans/agent/2026-09-02-cancel-running-script-or-automation.md`
- Spec: `docs/superpowers/specs/agent/2026-09-02-cancel-running-script-or-automation-spec.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All 7 waves merged (W01 #4788, W02 #4799, W03 #4967, W04 #4971, W05 #5030, W06 #5033, W07 #4990). Unreleased.

**Not verified / pre-release checklist**
- W04 (#4971) is agent-shipped Go: canary a small band before fleet promote. Windows Job-object containment also kills detached children on TIMEOUT, not just cancel. A runAs=user script surviving a helper IPC reconnect is unreachable → reports honest not_found/unconfirmed.
- W06 E2E spec `e2e-tests/tests/script-cancel.spec.ts` was written but not executed against a live stack.
- No real-device end-to-end cancel exercised (fence test resolves via retraction path).
- No load test of dispatch-fence lock contention on a large fleet run.

**Follow-ups filed / open**
- #5040: `cancelState` missing from GET /scripts/:id/executions and GET /scripts/executions/:id, so the too-late/failed copy cannot render against real data.
- `DeviceScriptHistory.tsx` (device Scripts tab) lists executions with no Stop button.
- `ownerScope` not on the automation runs GET list; partner-wide hide/tooltip defaults to "offered".
- Config-policy runs are out of scope (OD10-B) and 404; the RLS gap hiding partner-owned policy runs still wants its own issue.
- Helper executor persistence across IPC reconnect (see W04 residual).

**Design decisions to confirm before release**
- Cancelled run with a real device failure stays `cancelled` but also publishes `automation.failed` when devicesFailed > 0 (Fable + Codex xhigh quorum).
- W05 dispatch fence is check-before + compensate-after rather than the plan's held CAS, to avoid pinning a pooled connection across WebSocket sends.

PR #4799, opened 2026-09-03
feat(db,api): cancellation schema + script_cancel command registration (#3525 W02)

### LINKED ISSUE #4763: W02: Migrations (cancelling, script_cancel_state, columns, export-policy), CommandTypes.SCRIPT_CANCEL + all four registrations
Wave W02 of #4761 — Migrations (cancelling, script_cancel_state, columns, export-policy), CommandTypes.SCRIPT_CANCEL + all four registrations
Branch: `feature/4761-cancel-running-script-or-automation/wave-4763`

ToddHebebrand:
Wave W02 of #4761 (feature: "Cancel a running script or automation", origin #3525).

Plan: `docs/superpowers/plans/agent/2026-09-02-cancel-running-script-or-automation.md` (brought onto the branch with `tracking_issue: LanternOps/breeze#4761` added to its frontmatter).

This wave lays the **schema and command-registration foundation** for real script cancellation. Nothing in it queues a `script_cancel` yet — the request-side route rewrite is the follow-on (see **Deliberately out of scope** below), so this PR is inert at runtime and safe to merge on its own.

### 1. Migrations (plan Task 2.1)

Two files, because Postgres forbids *using* a new enum literal in the transaction that `ADD`s it and `autoMigrate` wraps each file in exactly one transaction:

- `2026-10-07-100000-cancellation-enums.sql` — `execution_status.cancelling` (`AFTER 'running'`), new type `script_cancel_state (requested|confirmed|unconfirmed|failed)`.
- `2026-10-07-100100-cancellation-columns.sql` — the five columns on `script_executions`, the `(cancel_state IS NULL) = (cancel_requested_at IS NULL)` CHECK, and two partial indexes.

`cancel_state` is **orthogonal to `status`** (spec OD8-C): `status` says what happened to the *process*, `cancel_state` says what happened to the *cancel request*. A cancel that cannot be proven reverts `status` to `cancel_prev_status` rather than claiming a stop the device never confirmed — which also keeps the row inside `reapStaleScriptExecutions`' `pending|queued|running` predicate, so a failed cancel never strands an execution and that reaper stays untouched.

**Deviation from OD1-A, as the plan flags:** OD1-A specified four columns; the revert model needs a fifth, `cancel_prev_status`, to make the revert deterministic.

### 2. Export-policy classification (plan Task 2.2)

`CORE_TENANT_EXPORT_POLICY` fires on a new **column**, not only a new table. All five new columns are `included`: two timestamps, two uuids and one status enum, with no `json`/`jsonb`/`bytea` among them and no `SUSPICIOUS_NAME_PARTS` match.

**All four CLAUDE.md cascade/export lists checked, not assumed** — `script_executions` was already registered in `CORE_ORG_CASCADE_DELETE_ORDER`, `CORE_DEVICE_CASCADE_DELETE_TABLES` and `CORE_DEVICE_ORG_DENORMALIZED_TABLES`; no new table was created, so the export policy was the only one that needed an edit. Verified by running all four contracts (below).

### 3. `script_cancel` in all four command lists (plan Task 2.3)

Each has its own silent failure mode:

| List | Failure mode if missed |
|---|---|
| `CommandTypes.SCRIPT_CANCEL` | no type safety at any call site |
| `AUDITED_COMMANDS` | `queueCommand` writes no dispatch audit |
| `REGISTRY_DISPATCHED_COMMAND_TYPES` | HTTP-polling agents never reach the result handler — the ack is dropped |
| `commandResultHandlers` | the ack is dropped on **both** transports |

Plus the **long (2 h) timeout tier**, deliberately not `SHORT_TIMEOUT_TYPES`: the generic reaper clocks `pending` rows from `createdAt`, so a 5-minute tier would expire a cancel that a merely-offline device never received, while the cancellation clock (which starts at *delivery*) has not started. `script_cancel` was already in `LIFECYCLE_COMMAND_TYPES`, so partner trust probation cannot trap a runaway script — asserted, not assumed.

### 4. `services/scriptCancellation.ts` — closer 2 of 5

Registering a result handler that drops the ack would be worse than not registering one, so this PR lands the closer it routes to. `applyScriptCancelAck` maps the agent ack onto the OD8-C table:

| Agent outcome | `status` | `cancel_state` |
|---|---|---|
| `terminated` | `cancelled` + `completed_at` | `confirmed` |
| `kill_failed` | revert to `cancel_prev_status` | `failed` |
| `not_found` | revert | `unconfirmed` |
| `"unknown command type"` (old agent) | revert | `unconfirmed` |
| pre-#3525 agent's bare `cancelled: true` | revert | `unconfirmed` |
| `null` result | revert | `unconfirmed` |

Only `terminated` is proof. Notably `not_found` is **not** confirmation — an agent that restarted has an empty running-process map and macOS/BSD has no `Pdeathsig`, so the orphan may survive; far more often it means the script finished a moment ago and its result is in flight. The read takes `FOR UPDATE` and every write is CAS-guarded on `status = 'cancelling'`, so a late original result that already terminalised the row cannot be resurrected.

## Verification (all run locally against a real Postgres via `pnpm test-stack`)

| Check | Result |
|---|---|
| `db/schema/scripts.test.ts` (new) | 9 pass — red first (9/9 failed before the migrations existed) |
| `services/scriptCancellation.ack.test.ts` (new) | 11 pass — red first (module did not exist) |
| `services/scriptCancellation.registration.test.ts` (new) | 6 pass — red first (5 of 6 failed) |
| `commandQueue.test.ts` (+1 behavioural audit pin) | 52 pass; **mutation-verified** — removing `SCRIPT_CANCEL` from the private `AUDITED_COMMANDS` set makes it fail |
| `tenant-export-policy.integration.test.ts` | pass — red first, naming all five unclassified columns |
| `tenantExportErasureRoundtrip.integration.test.ts` | pass |
| `tenantCascade.integration.test.ts` | pass |
| `rls-coverage.integration.test.ts` (own config) | 100 pass |
| `orgLifecycleFoundations` + `tenantCascadeExecution` | 10 pass (deferrable-FK / merge contract) |
| `commandResultHandlers*`, `routes/agents/commands`, `commandTimeouts`, `partnerTrust*`, `routes/scripts*`, `scriptExecution*`, `autoMigrate` | 15 files, 346 pass / 2 skipped |
| `pnpm db:migrate` ×3 then `db:check-drift` | idempotent, no drift (648 files match the ledger) |
| `pnpm lint` | clean (6/6) |

Live-DB spot checks: `execution_status` installed as `pending,queued,running,cancelling,completed,failed,timeout,cancelled` (order matches the Drizzle declaration, which is what `db:check-drift` compares); `script_cancel_state` as `requested,confirmed,unconfirmed,failed`; the CHECK constraint, the `ON DELETE SET NULL` FK and both partial indexes all present.

## Deliberately out of scope (needs a follow-on wave)

The plan's W02 also contains **Task 2.4's request side** (`cancelScriptExecution`, `cancelExecutionsForRun`) and **Task 2.5** (rewrite `POST /scripts/executions/:id/cancel` with post-commit delivery, `routes/devices/events.ts`, and the `cancel_script_execution` AI tool). This PR was scoped to migrations + registrations per sub-issue #4763's title, so those are **not** included and no sub-issue currently covers them. **W03 (#4764), W05 (#4766) and W06 (#4767) all depend on them** — a W02b should be dispatched before those waves start.

One thing that follow-on wave must know: `insertQueuedCommandInTransaction` — the insert site Task 2.4 uses — has **no audit block at all**, so `AUDITED_COMMANDS` membership does not by itself produce a dispatch audit on the transactional path. W03's audit-event task needs to write that row explicitly.

Closes #4763

https://claude.ai/code/session_01Aw4MBdKeih9V4cMTBQniQQ


ToddHebebrand:
## Review round — 3 Sonnet reviewers (code-reviewer, silent-failure-hunter, pr-test-analyzer)

**5 findings confirmed and fixed in `7dff90d4f`; 1 confirmed and deferred to W03 with reasoning; 1 dismissed.**

### Fixed

1. **`handleScriptCancelResult` had no try/catch** (silent-failure-hunter, critical). Every other handler in `commandResultHandlers.ts` wraps its body and calls `captureException` with context — verified by reading all nine. Both transports CAS the `device_commands` row to a terminal status *before* dispatching, so the agent never resends: a throw here permanently loses the only evidence that proves a stop, and on the WS path surfaces as an untagged generic Sentry event with nothing to search on. Now logs a handler-specific message and captures with `{ commandId, agentId }`. Deliberately not rethrown — matching every sibling, and the sweep (closer 5) is the designed net.

2. **Silent NULL `cancel_prev_status` fallback** (silent-failure-hunter, high). The code's own comment called a NULL here "a broken writer", then quietly coerced it to `'running'` and committed. A later wave's `cancelScriptExecution` failing to persist the column would misreport execution history indefinitely with nothing in logs or Sentry. Now reported (log + `captureException` with the execution and cancel-command ids); the safe revert behaviour is unchanged.

3. **The transport → closer wiring was untested** (pr-test-analyzer, critical gap). `commandResultHandlers['script_cancel']` was only asserted to be *a function*. A wrong id source there does not throw — the locking SELECT simply matches nothing and the execution sits in `cancelling`, exactly the silent no-op the module exists to prevent. New `commandResultHandlers.scriptCancel.test.ts` drives the real handler and pins the transport-authorized id, envelope pass-through, `undefined → null` normalization, and the error-reporting path. **Mutation-verified both ways:** sourcing the id from `command.payload.executionId` fails 2 tests; removing the error reporting fails 1.

4. **`errorMessage: 'Stopped on the device'` was unasserted** on the proven branch (pr-test-analyzer). Added to the `toMatchObject`.

5. **`setLiteralEntries` located the closing `]` in raw source before stripping comments** (pr-test-analyzer). A future comment containing `]` inside the `REGISTRY_DISPATCHED_COMMAND_TYPES` literal would silently truncate the parse without tripping the zero-entries guard. Comments are now stripped first.

### Confirmed but deliberately deferred to W03 (#4764)

**`handleScriptResult` does not recognize `cancelling`** (code-reviewer, important). Its primary CAS matches `status IN ('pending','queued','running')` and the #3607 recovery branch matches `('timeout','failed')`, so a genuine script result racing an in-flight cancel falls through to the drop-the-output path *and* pages `captureException` (the suppression checks `'cancelled'`, not `'cancelling'`).

This is real, and it is **already the plan's W03 Task 3.1, defect #1**, verbatim — it is the wave that owns the five closers and it has its own test file for exactly these transitions. It is **unreachable from this PR**: nothing writes `cancelling` anywhere in the tree (grepped `staleCommandReaper.ts`, `commandTimeouts.ts`, `commandQueue.ts`, the routes — no hits), because the request side does not exist yet. Fixing it here would duplicate and conflict with W03's own work. **W03 must not land without it** — flagging here so it is not read as a new discovery.

### Dismissed

The CAS-guarded UPDATE matching zero rows does not need to be observed: the preceding `SELECT ... FOR UPDATE` holds the row lock for the remainder of the same transaction, so the guard cannot fail after a successful read. Both reviewers that looked at it agreed.

### Also flagged by the reviewers as correct, having checked rather than assumed
Migration idempotency and the enum-add/enum-use transaction split; enum value order vs. the Drizzle declaration (`db:check-drift` compares order); the CHECK constraint and both partial indexes matching the schema exactly; `ON DELETE SET NULL` not tripping org erasure; all four CLAUDE.md cascade/export lists (grepped, not assumed); the OD8-C mapping never producing `cancelled` from an unproven outcome; two concurrent acks serializing correctly.

`tsc --noEmit` clean · `pnpm lint` clean (6/6) · 9 affected unit files, 144 tests pass · integration contracts (export policy ×2, tenant cascade, RLS coverage 100, org lifecycle) were green before the fixes and touch none of the changed lines.

ToddHebebrand:
## Post-review: 2 more fixes, CI now green on `6d068e20a`

**1. Real bug caught by CI's `#2823` guard (`noRawEnvNumberCoercion`), not by any reviewer.**
`CANCEL_GRACE_MS` was `Number(process.env.CANCEL_GRACE_MS ?? 90_000)`. Compose maps an unset variable as an **empty string**, and `Number('')` is `0` — so on any deploy that declared the variable without a value, the cancellation sweep would have given up on every cancel the instant it was delivered and recorded `unconfirmed`. Now `envInt('CANCEL_GRACE_MS', 90_000)`. Worth flagging to the later waves: this exact expression is in the plan's Task 2.4 code block, so a W02b copying it verbatim will reintroduce it.

**2. Migration rename — `origin/main` moved under the branch.**
Main gained `2026-10-07-100000-devices-reboot-schedule-columns.sql` after this branch was cut, which our `2026-10-07-100000-cancellation-enums.sql` sorted *before* (`c` < `d`), so it would have replayed ahead of it on a fresh database. Renamed to `2026-10-07-110000-cancellation-enums.sql` / `2026-10-07-110100-cancellation-columns.sql` and swept every reference — the schema comment, the schema test's filename constants, and the cross-reference inside the enum migration itself. Both files are unmerged, so the rename is free.

**Re-verified after both fixes, against a torn-down-and-recreated stack so the whole migration history replayed from scratch on the merged tree:** export policy ×2, tenant cascade, org-lifecycle deferrable-FK — 14 pass; RLS coverage — 100 pass; `execution_status` still installs as `pending,queued,running,cancelling,completed,failed,timeout,cancelled`; `tsc --noEmit` and `pnpm lint` clean.

**CI: 68/68 pass** (`Main Red Alert` skips on PRs) — including all four Integration Tests shards, Test API, both Check Migrations jobs, Smoke Test and Guided Setup Smoke.

Ready for review. Not merging — that is Todd's call, and note the **W02b gap** described in the PR body: the plan's W02 request side (`cancelScriptExecution` / route rewrite / AI tool) has no sub-issue and W03/W05/W06 all depend on it.

PR #4800, opened 2026-09-03
fix(billing): block QuickBooks push on voided/remote-deleted invoices (#4544)

### LINKED ISSUE #4544: Invoice accounting sync card offers 'Push to QuickBooks' on voided invoices and invoices deleted in QuickBooks
## Summary
`AccountingSyncCard.tsx` (`apps/web/src/components/billing/`) shows the push button whenever `canPush && isPushable(syncStatus)`. `isPushable` only looks at the mapping `sync_status`, so the button renders for:

1. Invoices that are **voided in Breeze** — the card gets no `voidedAt` signal, so a void invoice still offers a push.
2. Invoice mappings in `error` with `last_error = 'Deleted in QuickBooks'` (Phase D `markInvoiceDeletedRemotely`, decision 2 of the Phase D spec: never auto-resurrected). The error state is `isPushable`, so a click attempts to re-create the invoice in QBO.

Surfaced by the Phase D sandbox walk (#4537).

## Fix
- Pass the invoice's `voidedAt` / `status` into the card and hide the button when voided.
- Surface a `remoteDeleted: boolean` (or a distinct `sync_status`) in `AccountingSyncSummary` from the API so the card can render "Deleted in QuickBooks" with no push affordance, rather than string-matching `lastError`.
- Server side: `POST /accounting/quickbooks/invoices/:id/push` should reject voided invoices with 409 so the guard is not UI-only.
- Web tests for both hidden states; API test for the 409.

ToddHebebrand:
## Summary
`AccountingSyncCard.tsx` only checked `syncStatus` to decide whether to show "Push to QuickBooks", so the button rendered for:

1. Invoices **voided in Breeze** — the mapping row's own status is unaffected by a void, so a previously `error`/`pending` mapping still looked pushable.
2. Invoice mappings the reconcile worker marked with the `markInvoiceDeletedRemotely` sentinel (`last_error = 'Deleted in QuickBooks'`, Phase D decision 2: never auto-resurrected) — clicking Push would re-create the invoice in QuickBooks.

The **voided** case already had a server-side guard (`PUSHABLE_STATUSES` in `accountingInvoicePush.ts` excludes `void`), but the **remote-deleted** case did not: `upsertInvoiceMappingPending` (Phase 1b) would silently clear the marker (`lastError: null`) and let the push through server-side even if a client somehow submitted it.

## Fix

**API** (`apps/api/src/services/accounting/accountingInvoicePush.ts`):
- New `remote_deleted` error code, 409.
- Guard added in **two** places: Phase 1 (a cheap early exit, before any org/item lookup or token refresh — mirrors the existing currency-guard rationale) and Phase 1b (a fresh re-read right before claiming the mapping row `pending` — this is the check that actually closes the race, since the reconcile worker can flip the mapping to remote-deleted while the org/item dependency syncs in between are in flight).
- The exact sentinel string (`'Deleted in QuickBooks'`) is now a single exported constant (`INVOICE_REMOTE_DELETED_ERROR` in `services/accounting/types.ts`, a leaf module with no other imports — avoids a circular import between `accountingPaymentPull.ts` and `invoiceService.ts`), used by both the writer (`markInvoiceDeletedRemotely`) and the readers (the new push guard, and `invoiceService.getInvoiceAccountingSync`).
- `invoiceService.ts`'s `InvoiceAccountingSync`/`getInvoiceAccountingSync` now compute and surface a `remoteDeleted: boolean` from that constant.

**Web** (`apps/web/src/components/billing/`):
- `AccountingSyncSummary` gains `remoteDeleted: boolean` (API-computed — the component never string-matches `lastError` itself).
- `AccountingSyncCard` takes a new `invoiceStatus` prop (the invoice's own status, independent of the mapping row) and hides the Push button when the invoice is voided OR the mapping is remote-deleted, with an explanatory hint (`voidedHint` / `remoteDeletedHint`, added to all 8 locale files — `localeParity.test.ts` passes).
- `InvoiceDetail.tsx` passes `invoice.status` through.
- The push click handler is unchanged — a typed 409 (including the new `remote_deleted`) already surfaces via the existing `runAction`/`ActionError` toast path.

## Tests
- API: `accountingInvoicePush.test.ts` — new tests for the Phase-1 early rejection and the Phase-1b race-closing re-check (mapping flips to remote-deleted mid-dependency-sync); `invoicePush.test.ts` (route) — 409 pass-through for `remote_deleted`; `invoiceService.test.ts` — `remoteDeleted: true/false` surfaced correctly.
- Web: `AccountingSyncCard.test.tsx` — voided-invoice and remote-deleted-mapping hidden states (including the "both apply" precedence case), plus a regression test proving the component does NOT infer `remoteDeleted` from `lastError` text itself; `InvoiceDetail.test.tsx` — `invoiceStatus` threading test.

## Verification
- `cd apps/api && npx vitest run src/routes/accounting src/services/accounting/accountingInvoicePush.test.ts src/services/accounting/accountingPaymentPull.test.ts src/services/invoiceService.test.ts` — 329 passed.
- `cd apps/web && npx vitest run AccountingSyncCard InvoiceDetail.test src/lib/i18n/localeParity.test.ts src/lib/i18n/keyUsage.test.ts src/lib/__tests__/no-silent-mutations.test.ts` — all passed.
- `tsc --noEmit` clean in both `apps/api` and `apps/web`.
- `eslint` clean on all touched files.

Closes #4544

### DIFF
--- a/apps/api/src/routes/accounting/invoicePush.test.ts
+++ b/apps/api/src/routes/accounting/invoicePush.test.ts
@@ -235,6 +235,22 @@ describe('POST /accounting/:provider/invoices/:invoiceId/push', () => {
     expect(writeRouteAuditMock).not.toHaveBeenCalled();
   });
 
+  // #4544: a mapping row carrying the remote-deleted marker must reject the
+  // push with a stable 409 code, not a generic/quickbooks-flavored error the
+  // web layer would have to guess at.
+  it('409 pass-through: remote_deleted (invoice deleted/voided in QuickBooks) is never a silent no-op or a generic error', async () => {
+    pushInvoiceToAccountingMock.mockRejectedValue(
+      new AccountingInvoicePushError('remote_deleted', 409, 'QuickBooks reports this invoice as deleted — pushing again would create a duplicate.'),
+    );
+    const res = await pushInvoice();
+    expect(res.status).toBe(409);
+    expect(await res.json()).toEqual({
+      error: 'remote_deleted',
+      message: 'QuickBooks reports this invoice as deleted — pushing again would create a duplicate.',
+    });
+    expect(writeRouteAuditMock).not.toHaveBeenCalled();
+  });
+
   it('404 pass-through: an unpushable/unknown invoice preserves its code', async () => {
     pushInvoiceToAccountingMock.mockRejectedValue(new AccountingInvoicePushError('invoice_not_pushable', 404, 'Invoice not found for this partner'));
     const res = await pushInvoice();
--- a/apps/api/src/services/accounting/accountingInvoicePush.test.ts
+++ b/apps/api/src/services/accounting/accountingInvoicePush.test.ts
@@ -233,6 +233,19 @@ function orgMappingRow(overrides: Partial<MappingRow> = {}): MappingRow {
   };
 }
 
+// #4544: the exact shape `markInvoiceDeletedRemotely` (accountingPaymentPull.ts)
+// leaves behind — sync_status flipped to 'error', last_error the sentinel
+// literal, remote_entity_id deliberately NOT cleared (Phase D decision 2).
+function remoteDeletedInvoiceMappingRow(overrides: Partial<MappingRow> = {}): MappingRow {
+  return {
+    id: 'map-inv-1', integrationId: CONN_ID, partnerId: PARTNER, breezeEntityType: 'invoice', breezeEntityId: INVOICE,
+    remoteEntityType: 'Invoice', remoteEntityId: 'qb-inv-1', remoteSyncToken: '3',
+    remoteCurrencyCode: null, remoteDocNumber: null, linkStatus: 'confirmed', syncStatus: 'error',
+    lastError: 'Deleted in QuickBooks',
+    ...overrides,
+  };
+}
+
 function setup(opts: {
   invoice?: Partial<InvRow>;
   lines?: Partial<LineRow>[];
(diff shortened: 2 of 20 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer, silent-failure-hunter — parallel, sonnet)
**Findings:** 3 raised → all addressed in 7ea6aa217:
- [critical, pr-test-analyzer] `voidInvoiceInAccounting` could clobber the remote-deleted marker when voiding an already-remote-deleted invoice (provider.voidInvoice call against a stale remote id → its failure handler overwrote the sentinel) — fixed with a no-op short-circuit, new test using the existing `remoteDeletedInvoiceMappingRow()` fixture.
- [important, code-reviewer] Phase-1b's remote-deleted re-check was SELECT-then-UPDATE, not atomic — closed by conditioning the claiming UPDATE's WHERE clause on the row not already carrying the marker, with a diagnostic-only re-read on a zero-row match. New SQL-level regression test (real, unmocked drizzle-orm compiled WHERE tree) proves the guard predicate is actually wired in.
- [important, silent-failure-hunter] `AccountingSyncSummary.remoteDeleted` was typed as a required `boolean` but could be `undefined` at runtime during a deploy-skew window — made optional (matching the existing `stripeConnected` convention) with an explicit `?? false` default and a documented fail-safe story (server still 409s regardless).
**Tests:** apps/api affected suite green (`vitest run src/routes/accounting src/services/accounting/accountingInvoicePush.test.ts src/services/accounting/accountingPaymentPull.test.ts src/services/invoiceService.test.ts`, 331 passed); apps/web affected suite green (`vitest run AccountingSyncCard InvoiceDetail.test`, 38 passed); locale-parity and key-usage suites green; `tsc --noEmit` clean in both apps/api and apps/web; `eslint` clean on all touched files.

PR #4801, opened 2026-09-03
fix(api): align ticket-triage proposal DTO's draft text with draftsWritten (#4467)

### LINKED ISSUE #4467: AI agents: align ticket-triage proposal DTO's draft text with draftsWritten
The ticket-triage run-detail DTO (wave P2-4) carries draft text via the typed proposal object, but the separate `draftsWritten` field omits the content — the two representations of the same drafts are inconsistent and should be aligned.

ToddHebebrand:
## Summary

Closes #4467.

The ticket-triage run-detail DTO carried the draft text (`ticketProposal.draftReply` / `draftResolutionNote`) via the persisted `TicketTriageProposal` outcome — text frozen at proposal time — while `ticketProposal.draftsWritten` was a separate, live `ticket_drafts` read (id + kind only). These were two independently-sourced representations of the same draft(s) and could disagree: once a draft is actually written, a technician can edit its content on the ticket's "AI draft" surface (`TicketWorkbench.tsx`), or a write can be superseded, and the run-detail page kept showing the stale originally-proposed text forever, next to a `draftsWritten` entry that still pointed at the (now different) live row.

## Fix

`apps/api/src/services/aiAgents/runTrace.ts` (`mapTicketProposal` / new `pickDraftText` helper): derive `draftReply`/`draftResolutionNote` from the **same** live `ticket_drafts` read `draftsWritten` is built from, whenever a row exists for that kind — falling back to the proposal's own text only pre-write (a draft intent left `pending_approval` hasn't released into a `ticket_drafts` row yet, so there's no live truth to prefer, and the proposal preview is the only thing worth showing before a technician approves it).

This is chosen over the DTO's alternative ("carry the ticket_drafts id, have the client fetch content separately") because the run-detail page needs to preview proposed-but-not-yet-approved draft text, which by definition has no `ticket_drafts` row yet — so a client-side fetch-by-id can't cover the pending case, and a "sometimes have an id, sometimes have text" DTO shape is worse than the client always getting the right text at whichever field it already reads. It's also strictly better than syncing two copies (writing `proposal.draftReply`'s value into a new `content` field on `draftsWritten` at write time): that would still leave the DTO with two live-diverging copies after a later edit. Deriving off one source at read time removes the duplication instead of chasing it.

`apps/api/src/routes/aiAgents.ts`: the `ticket_drafts` select now also projects `content` — feeding the derivation only; `draftsWritten`'s own wire entries stay id/kind-only (never a `content` key), so a written draft's text is exposed exactly once on the wire.

`packages/shared/src/types/aiAgentRuns.ts`: docstring update only (no wire-shape change) — the `draftsWritten`/`AiAgentRunTicketProposalDto` comments previously claimed content "never" reaches the DTO via that path, which is now more precisely true of `draftsWritten`'s entries specifically, not `draftReply`/`draftResolutionNote`.

## Tests

- `apps/api/src/services/aiAgents/runTrace.test.ts` — new tests: draftReply/draftResolutionNote derive from the live written row when one exists (never disagreeing with `draftsWritten`), fall back to the proposal's own text pre-write, and `draftsWritten`'s own entries never carry a `content` key even though that content now legitimately reaches `draftReply`/`draftResolutionNote`. Existing tests updated for the new required `content` field on `RunTraceDraftRowInput`.
- `apps/api/src/routes/aiAgents.test.ts` — route-level assertion that `draftReply` matches the live draft-row content it derives from.

Ran red-first: added the alignment tests before the fix, confirmed they failed for the expected reason, then implemented.

```
cd apps/api && npx vitest run src/services/aiAgents/runTrace.test.ts src/routes/aiAgents.test.ts
# 2 files, 150 passed
NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit   # apps/api — clean
NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit   # packages/shared — clean (touched files); pre-existing unrelated errors in validators/quotes.test.ts, confirmed present on origin/main
npx vitest run src/types/aiAgentRuns.test.ts                # packages/shared — 10 passed
```

### DIFF
--- a/apps/api/src/routes/aiAgents.test.ts
+++ b/apps/api/src/routes/aiAgents.test.ts
@@ -1074,7 +1074,7 @@ describe('GET /ai-agents/runs/:runId (execution-trace detail, #3828)', () => {
         approvalScope: 'auto', decidedVia: 'ticket_autonomy',
       }])) // intents (sessionId is null, so the ledger read is skipped)
       .mockReturnValueOnce(selectChain(
-        [{ id: DRAFT_ID, kind: 'reply', content: 'Hi — please try restarting your computer.' }],
+        [{ id: DRAFT_ID, kind: 'reply', content: 'Hi — please try restarting your computer.', state: 'active' }],
         (predicate) => { draftWhere = predicate; },
       )); // draft rows
 
--- a/apps/api/src/routes/aiAgents.test.ts
+++ b/apps/api/src/routes/aiAgents.test.ts
@@ -1074,7 +1074,7 @@ describe('GET /ai-agents/runs/:runId (execution-trace detail, #3828)', () => {
         approvalScope: 'auto', decidedVia: 'ticket_autonomy',
       }])) // intents (sessionId is null, so the ledger read is skipped)
       .mockReturnValueOnce(selectChain(
-        [{ id: DRAFT_ID, kind: 'reply' }],
+        [{ id: DRAFT_ID, kind: 'reply', content: 'Hi — please try restarting your computer.' }],
         (predicate) => { draftWhere = predicate; },
       )); // draft rows
 
@@ -1085,6 +1085,9 @@ describe('GET /ai-agents/runs/:runId (execution-trace detail, #3828)', () => {
 
     expect(parsed.data.ticketProposal?.intentIds).toEqual([TRIAGE_INTENT_ID]);
     expect(parsed.data.ticketProposal?.draftsWritten).toEqual([{ kind: 'reply', draftId: DRAFT_ID }]);
+    // Issue #4467 — draftReply is sourced live off the same draft-rows read,
+    // never disagreeing with the draftsWritten entry it corresponds to.
+    expect(parsed.data.ticketProposal?.draftReply).toBe('Hi — please try restarting your computer.');
 
     // run row, intents, draft rows — no ledger (sessionId null), no sweep
     // hostname read, no narrative artifact read (reportRunId null).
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer + pr-test-analyzer (parallel, sonnet), CLAUDE.md compliance/bugs + test-coverage focus.

**Findings:** 1 real Important-severity finding, raised independently by both agents — `pickDraftText`'s original `draftRows.find()` had no guaranteed ordering, and more than one `ticket_drafts` row of the SAME kind can be linked to one run_id (the `draft` tool executor supersedes-then-inserts on every call, including its own unique-violation retry path; `ticket_drafts_active_uq` is scoped to `(ticket_id, kind)`, not `(run_id, kind)`). This could resurrect the exact staleness bug the issue asked to fix, by picking a superseded row instead of the active one. Addressed in `dc5d85fb5`: the route now also selects `state` and orders `draftRows` newest-first (`ORDER BY created_at DESC`); `pickDraftText` explicitly prefers the `active` row for a kind, falling back to the most recent write. Added regression tests for the active-vs-superseded tie-break and for both draft kinds being derived independently in one run. 0 findings outstanding — no other Critical/Important issues from either agent (RLS/org-scoping on the query, and the wire-DTO backward-compat shape of `draftsWritten`, were both confirmed unaffected).

**Tests:** `cd apps/api && npx vitest run src/services/aiAgents/runTrace.test.ts src/routes/aiAgents.test.ts` — 2 files, 152 passed. `npx tsc --noEmit` (apps/api, `NODE_OPTIONS=--max-old-space-size=8192` to work around host memory pressure) clean. `packages/shared`: touched files (`aiAgentRuns.ts`) typecheck clean; `npx vitest run src/types/aiAgentRuns.test.ts` — 10 passed. (Two pre-existing, unrelated `tsc` errors in `packages/shared/src/validators/quotes.test.ts` confirmed present on `origin/main`, untouched by this PR.)

PR #4803, opened 2026-09-03
refactor(api/tenancy): remove the partner-wide system-context escalations (#4676)

### LINKED ISSUE #4676: W03: Remove the nested system-context escalations (gated on W02 proof)
Wave W03 of #4673 — Remove the nested system-context escalations (gated on W02 proof)
Branch: `feature/4673-partner-current-partner-id-select/wave-4676`

ToddHebebrand:
Closes #4676

Wave W03 of #4673 (origin #2468). Plan: `docs/superpowers/plans/tenancy-rls/2026-08-16-2468-partner-current-partner-id-select-branch.md`.

## What this does

W01 (#4690) shipped the SELECT-only `<table>_partner_wide_select` RLS branches across the configuration-policy chain. W02 (#4725) populates `breeze.current_partner_id` on agent contexts. **The database now grants partner-wide reads directly, so this PR deletes the app-layer escapes that existed only to compensate for the old blindness.**

Every removal drops, per affected request:

- one **second pooled-connection acquisition** while the request's own connection is still held — the #1105 starvation shape (postgres-js queues acquisitions with no timeout; at concurrency ≥ pool size, ~25 on the hosted DBs, that is a hang); and
- one **wholesale RLS bypass** — the #2417 cross-tenant shape.

## Removed

| Site | What it was |
|---|---|
| `services/configPolicyOwnership.ts` | `withPartnerWideVisibility` — deleted, plus its 15 call sites: 11 in `services/featureConfigResolver.ts`, 4 in `routes/agents/helpers.ts` |
| `routes/agents/helpers.ts` (`handleCisCommandResult`) | `cis_baselines` PK escalation — its own comment already handed it to W03 |
| `services/configurationPolicy.ts` ×3 | the direct `runOutsideDbContext(() => withSystemDbAccessContext(...))` wraps in `isBackupProfileReference` and the two `validateFeaturePolicyExists` backup lookups |

## Kept — with a justification comment on each

| Site | Why it stays |
|---|---|
| `db/partnerAxisRead.ts`, `services/enrollmentDefaults.ts` | **partner-AXIS** tables (`partners`, `PARTNER_TENANT_TABLES`). There is no `org_id IS NULL` shape to key a SELECT branch on, and the only gate is `breeze_has_partner_access`, which also governs WRITES. Granting it to org scope would widen UPDATE/DELETE. #2822's territory. |
| `buildPolicyProbeConfigUpdate` | reads `automation_policies`, which is dual-axis but has **no** `*_partner_wide_select` branch. Converting it would silently zero out partner-wide compliance probe config. Flagged in the plan doc as a gap for W04's contract assertion to catch. |
| `getOrgAgentUpdateConfig` | reads the partner-AXIS `partners` table. |
| `withDevicePartnerPolicyVisibility` (#3493) | not a system escape — a same-connection widening of `breeze.accessible_partner_ids`, which also admits partner-AXIS rows the SELECT-only branch never grants. Retiring it is a separate change. |
| `services/vulnerabilityFleetQueries.ts`, `ticketConfigService.ts`, `commandQueue.ts` | different problems entirely (global catalog, cross-org worker reads, transaction-isolation semantics) — unchanged from the plan's keep-list. |

### `routes/agents/heartbeat.ts` — deliberately not converted (read this before "cleaning it up")

W02 left a comment saying W03 would remove heartbeat's config-delivery escapes. It doesn't, and the comment has been corrected in place to say why.

Those five reads sit in their own **top-level** `withSystemDbAccessContext` blocks (the route opts out of the request-long wrap), so they are not nested and cost no second connection — W03's target was the nested escapes. Converting them is a genuine follow-up, but **not** a drop-in swap: `buildPatchSourceConfigUpdate` reaches `resolveDeviceTimezone`, whose `partners` read is partner-AXIS and escapes through `readWithPartnerAxisVisibility`. Under a system wrapper that escape short-circuits; under an org-scoped wrapper it fires, **uncached, once per heartbeat** — converting one hoisted context into a genuinely nested double-hold on the fleet's hottest path, i.e. the exact shape this epic exists to remove. The timezone read has to be hoisted or batched first.

Recorded in the plan doc; recommend a separate sub-issue rather than growing this PR.

## Behavioural change to review

**A deliberate tightening on the two bare-id probes** (`isBackupProfileReference`, the `cis_baselines` PK lookup). Escaped, they saw **every tenant's** rows and relied on post-hoc self-tenanting. Under the caller's own context a foreign partner's row now reads as absent and takes the existing "not found" branch — same outcome, reached one step earlier, without the bypass. The post-checks stay as defense in depth. Both callers of the CIS read already swallow a miss (log line, discarded scan result), never a wrong-tenant write.

**Reads widen** for org-scoped tokens to their own partner's partner-wide rows — but only to exactly the rows the system escape already showed them, minus everyone else's. Writes are untouched: every branch is `FOR SELECT`, and Postgres never consults FOR SELECT policies when computing UPDATE/DELETE target rows (asserted below).

## Tests — red-first and mutation-verified

Every removed escalation has a test proving the partner-wide rows are still returned, and the negative half proving the **RLS branch** (not a surviving escape) is what returns them.

- **`featureConfigResolverPartnerWide.integration.test.ts`** (new, added after review) — 9 `featureConfigResolver.ts` resolvers × (positive, blind) against real Postgres: alert rules, governing alert-rule policy, automations, compliance rules, maintenance config, software policy, vulnerability-enabled, backup config, backup protection. Every one of these had its escape removed with no real-RLS coverage before; their only tests mock Drizzle.
- **`agentPolicyResolversPartnerWide.integration.test.ts`** — `orgContext` now carries `currentPartnerId` (the real post-W02 agent shape). New `partnerWideBlindContext` cases run the same four resolvers with `currentPartnerId: null` and require the partner-wide policy to be **INVISIBLE**. Under an escape the GUC is irrelevant and the row resolves regardless, so these can only pass with no escape. Plus a `FOR SELECT`-only write-lock case asserting UPDATE/DELETE affect **0 rows** (RLS hides the target — a silent no-op, not a 42501, so rowCount is the only non-vacuous assertion).
- **`backupProfilesPartnerRls.integration.test.ts`** — the plan's required org-scoped fan-out proof still passes (now with the real `currentPartnerId`), plus a blind counterpart. Six new tests give `isBackupProfileReference` / `validateFeaturePolicyExists` their **first real-DB coverage under org RLS** — every other test of them mocks the database and evaluates no RLS at all.
- **`helpers.partnerWidePolicies.test.ts`** — inverted. `runOutsideDbContext` / `withSystemDbAccessContext` are now mocked to **throw**, and the `configPolicyOwnership` mock factory omits `withPartnerWideVisibility` so re-importing it fails with "No … export is defined on the mock".
- **`buildHelperConfigUpdate`** — the fifth agent resolver, previously uncovered, added to `agentPolicyResolversPartnerWide` with its blind control.
- **`featureConfigResolver.backupAssignments.test.ts`** — escape count 2 → 1; only the partner-AXIS timezone read remains, and the config-policy join is asserted at caller depth.

**Mutation checks run** (reintroduce the escape → confirm red, then revert):

| Mutation | Result |
|---|---|
| escape re-added to `resolveDeviceEventLogSettings` | `event_log: … INVISIBLE without breeze.current_partner_id` → **FAIL** (`expected 555 to be 100`) |
| same | `helpers.partnerWidePolicies` → **5 FAIL** on the throwing mock |
| escape re-added to `isBackupProfileReference` | blind + foreign-partner cases → **2 FAIL** |
| escape re-added to `resolveAlertRulesForDevice` | new resolver suite's blind case → **FAIL** |

## Verification

- `tsc --noEmit` (apps/api) — clean.
- Unit: 60 files / 1184 tests green (`configPolicyOwnership`, `helpers.*`, `featureConfigResolver*`, `configurationPolicy*`, all of `routes/agents/`); plus a wider caller sweep of 91 files / 1440 tests green (`jobs/backupWorker`, `automationWorker`, `maintenanceRebootWorker`, `routes/backup`, `alertService`, `pamBridge`, `policyEvaluationService`, `commandResultHandlers`).
- Integration against a real Postgres (`pnpm test-stack`): 11 files / 134 tests green — `agentPolicyResolversPartnerWide`, `backupProfilesPartnerRls`, `configPolicyPartnerWideSelect`, `agentContextPartnerWideFanout`, `configurationPolicyOrgScopedPartnerWide`, `normalizedConfigPolicyRls`, `configurationPoliciesPartnerRls`, `partnerAxisSystemContext`, `configurationPolicyPartnerResolution`, `helperPolicyResolution`, `cisBaselinesPartnerRls`. Plus batches covering `helperPolicyResolution`, `onedrive-helper-config-delivery`, `configPolicyFeatureReferenceIntegrity`, `configPolicyReference{,Gate}Concurrency`, `configPolicyAutomationRunRls`, `backupFeatureLinkCascade`, `patchSchedulerPartnerResolution`, `automationWorkerPartnerResolution`, `vulnerabilityCorrelationGating`, `agentUpdatePolicyEffective`, `pamUacGrandfather`, `maintenanceWindowsPartnerRls`, `softwarePoliciesPartnerRls`, `staleBackupReaper`, `backupResultTenantScope`.
- RLS contracts: `vitest.config.rls.ts` (26 passed / 5 pre-existing skips) and `vitest.config.rls-coverage.ts` (100 passed).
- **No schema change**, so no cascade / export-policy / migration-naming surface is touched.

1. Convert heartbeat's five hoisted system contexts, after hoisting/batching the partner-axis timezone read (see above).
2. `automation_policies` is dual-axis with no `*_partner_wide_select` branch — W04's contract assertion should catch it.
3. Pre-existing gaps found during the plan's step-7 sweep, unchanged by this PR (bare org-equality that structurally cannot match `org_id IS NULL`, so RLS is not the cause):
   - `routes/policyManagement/compliance.ts:83-90,217-225` — config-policy counts **undercount** partner-wide policies, while the sibling automation-policy count right above deliberately ORs them in.
   - `services/aiToolsConfigPolicy.ts:517-527` and `routes/softwareInventory.ts:186-198` — name-uniqueness/dedup lookups ignore a same-named partner-wide policy, so the duplicate-name error never fires for that case.

   Left out because fixing them changes user-visible counts and needs product sign-off, not because they're unimportant.

https://claude.ai/code/session_01YPYSuYeNtq9SRmVv1ELgYN

### DIFF
--- a/apps/api/src/__tests__/integration/agentPolicyResolversPartnerWide.integration.test.ts
+++ b/apps/api/src/__tests__/integration/agentPolicyResolversPartnerWide.integration.test.ts
@@ -16,27 +16,46 @@
  *      `accessiblePartnerIds: []` — so the read silently returns ZERO ROWS,
  *      not an error.
  *
- * The fix (services/configPolicyOwnership.ts) is a matched pair:
- * `policyOwnershipCondition` (dual-axis join predicate) plus
- * `withPartnerWideVisibility` (a scoped system-context escape around ONLY the
- * policy join). This test proves both halves for all four resolvers by
- * invoking them exactly as the agent heartbeat path does: inside a real
- * org-scoped `withDbAccessContext`, against the real breeze_app RLS-forced
- * connection.
+ * The original fix (#2930) was a matched pair: `policyOwnershipCondition` (the
+ * dual-axis join predicate) plus `withPartnerWideVisibility`, a scoped
+ * system-context escape around ONLY the policy join.
+ *
+ * **#4673 W03 replaced the second half.** The escape is deleted. Its job is now
+ * done by the database: `<table>_partner_wide_select` (W01) adds a SELECT-ONLY
+ * policy `org_id IS NULL AND partner_id = public.breeze_current_partner_id()`
+ * across the configuration-policy chain, and W02 populates
+ * `breeze.current_partner_id` on agent contexts from the device org's partner.
+ * So the agent reads its own MSP's partner-wide config on its OWN connection —
+ * no second pooled connection (the #1105 starvation shape) and no RLS bypass
+ * (the #2417 cross-tenant shape).
+ *
+ * This test proves both halves for all four resolvers by invoking them exactly
+ * as the agent heartbeat path does: inside a real org-scoped
+ * `withDbAccessContext`, against the real breeze_app RLS-forced connection.
+ *
+ * The `partnerWideBlindContext` cases are what make it a real proof rather than
+ * a tautology. They run the same resolvers under a context with
+ * `currentPartnerId: null` and require the partner-wide policy to be INVISIBLE.
+ * That fails if the system-context escape is ever reintroduced — under an
+ * escape the GUC is irrelevant and the row resolves regardless. Together the
+ * two halves pin: visible with the GUC, invisible without it, therefore the
+ * RLS branch (not an escape) is what is doing the work.
  *
  * If this test file were deleted: a regression that reintroduces a bare
- * `eq(configurationPolicies.orgId, device.orgId)` join, or that drops the
- * `withPartnerWideVisibility` escape, would compile fine, pass every unit
- * test (which mock the DB and never exercise RLS), and pass
+ * `eq(configurationPolicies.orgId, device.orgId)` join, that drops
+ * `currentPartnerId` from the agent context, or that reintroduces the escape,
+ * would compile fine, pass every unit test (which mock the DB and never
+ * exercise RLS), and pass
  * configurationPolicyPartnerResolution.integration.test.ts (which resolves
- * under a SYSTEM-scope AuthContext, so the RLS escape is never exercised).
- * Partner-wide policies would silently stop reaching agents for event_log,
- * monitoring, PAM, and patch-source config — exactly the bug #2930 fixed.
+ * under a SYSTEM-scope AuthContext, where every RLS branch short-circuits
+ * true). Partner-wide policies would silently stop reaching agents for
+ * event_log, monitoring, PAM, and patch-source config — exactly the bug #2930
+ * fixed — or would keep reaching them through a connection-doubling bypass.
  */
 import './setup';
 import { afterEach, describe, expect, it } from 'vitest';
 import { randomUUID } from 'crypto';
-import { eq } from 'drizzle-orm';
+import { eq, sql } from 'drizzle-orm';
 import { db, withDbAccessContext, type DbAccessContext } from '../../db';
 import {
   configurationPolicies,
@@ -67,18 +86,41 @@ const SYSTEM_CTX: DbAccessContext = {
   userId: null,
 };
 
-// The realistic agent-facing shape: an org-scoped request context.
-// accessiblePartnerIds: [] is the crux of the bug — an org token never
-// carries partner-axis access, so breeze_has_partner_access is false and a
-// partner-owned policy is invisible unless the resolver escapes to a system
-// context for the policy join (withPartnerWideVisibility).
-function orgContext(orgId: string): DbAccessContext {
+// The realistic agent-facing shape, as `agentAuthMiddleware` builds it since
+// #4673 W02: an org-scoped request context that carries the DEVICE ORG'S OWN
+// partner in `currentPartnerId` while `accessiblePartnerIds` stays [].
+//
+// The two fields are different axes and must not be conflated.
+// `accessiblePartnerIds` gates `breeze_has_partner_access`, which admits
+// partner-axis WRITES — an agent never gets it. `currentPartnerId` feeds the
+// `breeze.current_partner_id` GUC, which only the SELECT-only
+// `<table>_partner_wide_select` policies read. That branch is now the ONLY
+// thing making a partner-owned policy legible here: W03 deleted the nested
+// system-context escape (`withPartnerWideVisibility`) these resolvers used to
+// take. `partnerWideBlindContext` below pins that by omitting the GUC.
+function orgContext(orgId: string, partnerId: string): DbAccessContext {
   return {
     scope: 'organization',
     orgId,
     accessibleOrgIds: [orgId],
     accessiblePartnerIds: [],
     userId: null,
+    currentPartnerId: partnerId,
+  };
+}
+
+// The SAME context minus `currentPartnerId` — i.e. what an agent context looked
+// like BEFORE W02. Used by the "the RLS branch is load-bearing" cases below: if
+// a system-context escape were ever reintroduced (or survived), a partner-owned
+// policy would resolve under this context too, and those assertions fail.
+function partnerWideBlindContext(orgId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: [],
+    userId: null,
+    currentPartnerId: null,
   };
 }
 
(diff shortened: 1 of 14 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (Sonnet, in parallel, against `git diff origin/main...HEAD`).

**Findings:** 3 raised, all confirmed and addressed in `4c59fef63`; 0 outstanding.

1. **Stale comment (confirmed).** `services/featureConfigResolver.ts` `resolveBackupConfigForDevice` still claimed its read runs "in a system context" — false after the unwrap, and exactly the contract-drift shape that leads a future caller to omit `currentPartnerId` from a hand-built context and silently ship zero-row backup config. My regex sweep matched the other 10 comment variants and missed this one. Rewritten to state what the code does and that `currentPartnerId` is load-bearing.
2. **Missing real-RLS coverage (confirmed, the significant one).** This wave's own rule is that every removed escalation needs a test proving the partner-wide rows still come back. Eight `featureConfigResolver.ts` resolvers plus `buildHelperConfigUpdate` had their escape removed with **no test exercising them under a non-system `DbAccessContext`** — their only coverage mocks Drizzle and evaluates no RLS, so a missing or wrong `*_partner_wide_select` policy would have shipped green. Two existing integration files (`helperPolicyResolution`, `vulnerabilityCorrelationGating`) were giving false confidence by only ever calling under `withSystemDbAccessContext`. Closed with:
   - new `featureConfigResolverPartnerWide.integration.test.ts` — 9 resolvers × (positive, blind) against real Postgres, 18 tests;
   - `buildHelperConfigUpdate` + blind control added to `agentPolicyResolversPartnerWide` (the fifth agent resolver, previously uncovered), plus its cache key in `purgeCaches`.

   Every seeded value differs from the resolver's default, so no assertion can pass vacuously.
3. **Hand-rolled row-count cast (confirmed, minor).** The write-lock test reintroduced `as unknown as { count: number }`, the pattern `db/rowCount.ts` exists to eliminate (#3894). Swapped for `extractRowCount`.

Both reviewers independently traced every `DbAccessContext` producer that can reach the unwrapped readers (auth, bearerToken, apiKey, agentAuth, agentWs, portal, oauth, job workers, AI-tool dispatch) and found **no reachable zero-row path**: workers are all `withSystemDbAccessContext`-wrapped, HTTP/AI-tool paths go through the canonical `buildDbAccessContext`/`dbAccessContextFromAuth`, and both CIS command-result paths thread `currentPartnerId`. The mechanical unwrap was verified clause-by-clause on a whitespace-normalized diff, including both inlined `async () => {}` blocks.

**Mutation checks** (reintroduce an escape → confirm red → revert) — 4 run, all red as required:

| Mutation | Result |
|---|---|
| escape re-added to `resolveDeviceEventLogSettings` | `event_log … INVISIBLE without breeze.current_partner_id` **FAIL** (`expected 555 to be 100`) |
| same | `helpers.partnerWidePolicies` **5 FAIL** on the throwing db mock |
| escape re-added to `isBackupProfileReference` | blind + foreign-partner cases **2 FAIL** |
| escape re-added to `resolveAlertRulesForDevice` | new resolver suite's blind case **FAIL** |

**Tests:** `tsc --noEmit` clean; `eslint` clean on all changed files. Unit 60 files / 1184 tests green (plus a 91-file / 1440-test caller sweep). Integration against real Postgres: the six partner-wide suites green (95 tests), plus earlier batches covering config-policy references, patch/automation partner resolution, onedrive delivery, backup cascade/reaper/tenant-scope, CIS baselines, maintenance/software partner RLS. RLS contracts: `vitest.config.rls.ts` 26 passed (5 pre-existing skips), `vitest.config.rls-coverage.ts` 100 passed. No schema change, so no cascade / export-policy surface is touched.

**Knowingly left uncovered:** the `handleCisCommandResult` call-site wiring for the `cis_baselines` read. The underlying `cis_baselines_partner_wide_select` policy is already proven — including an agent-shaped blind context — in `cisBaselinesPartnerRls.integration.test.ts`, and the post-read org/partner re-check is unchanged by this PR. Worst case on a miss is a discarded scan result plus a log line, never a wrong-tenant write.

**Status:** review-clean. CI still running at the time of writing — do not merge until `CI Success` is green on `4c59fef63`. Awaiting maintainer merge; per the wave protocol I am not merging this or closing #4676.

PR #4784, opened 2026-09-03
fix(api): wrap loadIntentDecidedVia's per-event SELECT in try/catch (#4464)

ToddHebebrand:
## Summary

`loadIntentDecidedVia` in `apps/api/src/jobs/intentReleaseWorker.ts` (the `intent_created` recovery routing read, wave P2-4) issued its per-event SELECT with no try/catch. A transient DB error there previously relied on BullMQ's retry to paper over it rather than being handled explicitly — and since it runs once per `intent_created` event, an unhandled rejection here aborted the whole event/batch processing rather than degrading just this one event.

## Fix

Wrap the SELECT in try/catch. On failure: log via `console.error` with the intent id and the error, report to Sentry via `captureException` (matching the existing pattern used throughout this file), and return `null` — the same value the function already returns for a missing row. The caller (`intent_created` branch, ~line 1419) already fails open on `null` by falling through to the ordinary `attemptPolicyDecision` call, which is itself a safe no-op for a row it doesn't recognize as `unattempted`.

Diff is confined to `loadIntentDecidedVia` (~lines 1452-1461) and its test — kept clear of the unrelated `intentReleaseWorker.ts` edits around lines 1272/1347 from a concurrently in-flight PR for #4465.

## Tests

Added to `apps/api/src/jobs/intentReleaseWorker.test.ts`:
- A throwing SELECT is caught, logged once to Sentry, and treated as a missing row (falls through to `attemptPolicyDecision`, not thrown to the caller).
- The one-shot failure doesn't poison later calls — a subsequent lookup on a different intent still succeeds normally.

Verified red-first: both new tests failed against the pre-fix code (unhandled rejection propagated out of `processIntentReleaseJob`), then passed after the fix.

## Verification

- `cd apps/api && npx vitest run src/jobs/intentReleaseWorker.test.ts` — 116/116 passed
- `cd apps/api && npx vitest run src/jobs/intentReleaseWorker` (both non-integration files) — 118/118 passed
- `cd apps/api && npx tsc --noEmit` — clean (no errors)

Closes #4464

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CBxF7C8E5GxFnNcGf6kBVP

PR #4805, opened 2026-09-03
fix(web): seed the AI threshold ladder during render, not from an effect (#4659)

### LINKED ISSUE #4659: flaky: AiUsagePage 'disables Save while the ladder text does not parse' — distinct failure signature under extreme load, residual after #4601 fix
**Context**

While fixing #4601 (PR #4658 — scoping `AiUsagePage.test.tsx` queries to `within(container)` instead of the shared `screen` to eliminate a cross-test DOM query race), I stress-tested the fix by pinning host CPU with several busy-loops on top of an already-loaded dev box and running the file 120 times in a loop.

Before the fix this reliably reproduced the originally-reported race (PUT body carrying the pre-edit `alertThresholdPercents` value). After the fix, that specific failure mode dropped to 1/120 runs (down from readily reproducible).

**New, distinct residual signature (2/120 runs)**

`AiUsagePage alert threshold ladder > disables Save while the ladder text does not parse` failed with:

```
TestingLibraryElementError: Unable to find an element by: [data-testid="ai-budget-thresholds-error"]
 ❯ src/components/settings/AiUsagePage.test.tsx:271:12
    269|     fireEvent.change(input, { target: { value: '100' } });
    270|     fireEvent.blur(input);
    271|     expect(getByTestId('ai-budget-thresholds-error')).toBeInTheDocument();
```

i.e. after typing an out-of-range value (`100`) and blurring, the inline validation error span never appeared — a genuinely different mechanism than the container-scoping race #4601/PR #4658 fixed (this is a same-instance, same-render assertion, not a cross-test DOM match). This did **not** reproduce at all in two full, un-pressured `apps/web` suite runs (700/700 files, 7379/7379 tests, twice), and did not reproduce in 60 un-pressured isolated runs of the file either — it only appeared under artificially extreme CPU pressure (6 saturating busy-loops on an already ~2.4x-oversubscribed 18-core box), which is more severe than a real CI runner experiences.

**Ask**

Not urgent — no evidence this has fired in real CI. Filing so it's tracked rather than silently dropped, per the pr-test-analyzer review on PR #4658 which flagged it as a distinct signature worth a follow-up rather than folding into #4601's "resolved" status. If it's ever observed on real CI (`Test Web` job), this issue has the repro conditions and exact stack trace to start from.

Refs #4601, PR #4658.

ToddHebebrand:
## Problem

`AiUsagePage.test.tsx`'s "alert threshold ladder" block has been failing on real CI, three times today alone, always with an API-only diff on the captured budget PUT:

| Run | Test | Expected | Received |
|---|---|---|---|
| 33723866739 (attempts 1 & 2, PR #4784) | `sends alertThresholdPercents: null when the ladder is cleared` | `null` | `[50, 80, 95]` |
| 33733136235 (PR #4741) | `sends the edited ladder` | `[60, 90]` | `null` |

Plus the signature originally filed on #4659: `disables Save while the ladder text does not parse` — the inline error span never appeared after typing `100`.

In every case the value that reached the PUT is **the ladder the box held at mount**, not what the test typed.

## Root cause

`AiBudgetThresholdsInput` re-seeded its local `text` state from the `value` prop inside a `useEffect`:

```tsx
useEffect(() => { setText(value?.join(', ') ?? ''); markInvalid(false); }, [value, markInvalid]);
```

A passive effect is flushed **after** the commit, so a queued `setText` can land *after* the user has already typed, writing the string the effect captured when it was scheduled over the keystroke. The following `commit()` then parses that reverted text and calls `onChange` with the pre-edit ladder.

This is normally invisible. At mount the effect writes exactly what `useState` already holds, so React's eager bailout drops the update without scheduling anything. It only becomes a real write once the fiber has a pending update — which is precisely the case after the user has typed.

The window is opened by `@testing-library`: `asyncWrapper` turns the **act environment off** for the duration of `waitFor`/`findBy*` and then drains with `setTimeout(0)`. Under CI load that loses the race against React's scheduler, so `findByTestId('ai-budget-thresholds-input')` can return while the component's mount effects are still queued — and then `fireEvent.change` runs first, the stale seed second.

One mechanism explains all three signatures:

- cleared field → seed restores `50, 80, 95` → PUT `[50, 80, 95]`
- typed `60, 90` over an empty box → seed restores `''` → PUT `null`
- typed `100` → seed restores `50, 80, 95`, which parses → no error span, Save never disabled

**This is not a new diagnosis in this repo.** `InvoiceEditor`'s notes/terms drafts had the identical defect — an effect-deferred prop→state sync overwriting a keystroke — filed five times (#2925 → #3219 → #3277 → #3980 → #4033) before the effect was recognised as the cause, and fixed by moving the sync into render. This change deliberately mirrors that fix, including its comparison strategy.

## Fix

Derive the seed **during render**, guarded by a `seededFrom` state cell, so there is no commit in which the box still shows the previous ladder and therefore no window in which a keystroke can be overwritten.

The seed is compared as the **rendered string**, not by array identity. Identity comparison would reproduce the same clobber class through a different door — and reachably so: `AiUsagePage.fetchData()` re-runs when the `showFlaggedOnly` checkbox is toggled and unconditionally rebuilds `budget.alertThresholdPercents` from the response, so a user mid-edit would have their draft reverted. `PartnerSettingsPage.fetchPartner` has the same shape.

The render-phase reset cannot call `markInvalid` (notifying the parent mid-render is illegal) or mutate `invalidRef` (a render can be discarded), so a small effect reconciles both. It is a no-op for every `markInvalid`-driven change — those still notify the caller synchronously, so the Save-gate behaviour that `disables Save while the ladder text does not parse` relies on is unchanged.

## Verification

**Deterministic regression tests** (both verified red on the pre-fix component, green after):

- `re-seeds a changed value prop within the same commit, not a later one` — asserts the seed lands in the same commit as the prop change, observed through a layout effect (mutation phase of that commit) so it does not depend on scheduler timing. Red before: `expected [ '50, 80' ] to deeply equal [ '60, 90' ]`.
- `keeps a draft when a refetch hands back an equal-but-new value array` — red against the identity comparison: `expected '50, 80' to be '60, 90'`.

**Contract tests** added from review (these pin behaviour rather than reproducing the flake):

- `releases the caller when value resets over unparseable text` — covers the reconciling effect. Mutation-checked: deleting that effect fails this test and only this test.
- `does not clobber uncommitted text across an unrelated re-render` — the motivating scenario stated positively.
- `re-seeds to empty when value resets to undefined` — the `?.`/`??` branch nothing else reaches.

An earlier attempt at the first test used `flushSync` to leave passive effects pending; its non-vacuity control caught that React 19 flushes them anyway, so that approach was dropped rather than shipped as a test that could never fail.

**Stress reproduction** (18-core box, load average already ~44 from other work):

- baseline, component reverted to `main`: `npx vitest run src/components/settings/AiUsagePage.test.tsx --pool=threads --maxWorkers=1` × 40 under 8 busy-loops → **1 failure**, byte-identical to CI run 33733136235 (`sends the edited ladder`, expected `[60, 90]`, received `null`, same `AiUsagePage.test.tsx:244:49`). A second baseline at 20 busy-loops × 50 was clean — the race is not monotonic in load, which matches the low rate seen on CI and in the original #4659 report (2/120).
- after the fix, same command and same 8-busy-loop load: **80/80 green**, then **40/40 green** again on the final code
- both changed files together, 25 iterations under the same load: **25/25 green**

**Suites and checks**

- `vitest run src/components/settings` → 85 files / 990 tests passed
- `npx tsc --noEmit -p apps/web/tsconfig.json` → clean
- `npx eslint` on both changed files → clean

Closes #4659

### DIFF
--- a/apps/web/src/components/settings/AiBudgetThresholdsInput.test.tsx
+++ b/apps/web/src/components/settings/AiBudgetThresholdsInput.test.tsx
@@ -1,6 +1,21 @@
 import { render, screen, fireEvent } from '@testing-library/react';
+import { useLayoutEffect } from 'react';
 import { describe, expect, it, vi } from 'vitest';
 import AiBudgetThresholdsInput from './AiBudgetThresholdsInput';
+
+/**
+ * Records the input's committed DOM value once per commit. Layout effects run
+ * synchronously in the mutation phase of the very commit that produced the
+ * DOM, so this observes what the box showed at each commit boundary without
+ * depending on when React's scheduler gets around to passive effects.
+ */
+function CommitProbe({ testId, seen }: { testId: string; seen: string[] }) {
+  useLayoutEffect(() => {
+    const el = document.querySelector(`[data-testid="${testId}-input"]`);
+    if (el) seen.push((el as HTMLInputElement).value);
+  });
+  return null;
+}
 
 describe('AiBudgetThresholdsInput', () => {
   it('renders current rungs as chips and emits a normalised list on blur', () => {
@@ -115,5 +130,40 @@ describe('AiBudgetThresholdsInput', () => {
     fireEvent.change(input, { target: { value: '80' } });
     fireEvent.blur(input);
     expect(onChange).not.toHaveBeenCalled();
+  });
+
+  // #4659 / #4601: the box used to seed itself from `value` in a `useEffect`,
+  // i.e. in a commit AFTER the one that delivered the new prop. Because a
+  // passive effect is deferred, its `setText` could land after the user had
+  // already typed, writing the effect's captured string over the keystroke —
+  // the next blur then committed the PRE-EDIT ladder.
+  //
+  // That window is not hypothetical: @testing-library's `asyncWrapper` turns
+  // the act environment OFF for the duration of `waitFor`/`findBy*` and drains
+  // with `setTimeout(0)`, which under CI load loses the race against React's
+  // scheduler. On real CI it surfaced in `AiUsagePage` as a PUT carrying
+  // `[50, 80, 95]` where the test had cleared the field, and `null` where it
+  // had typed `60, 90` — always the value the box held at mount.
+  //
+  // Seeding during render closes the window: there is no commit in which the
+  // box still shows the old ladder, so assert exactly that.
+  it('re-seeds a changed value prop within the same commit, not a later one (#4659)', () => {
+    const seen: string[] = [];
+    const view = (value: number[]) => (
+      <>
+        <AiBudgetThresholdsInput value={value} onChange={vi.fn()} testId="thresholds" />
+        <CommitProbe testId="thresholds" seen={seen} />
+      </>
+    );
+
+    const { rerender } = render(view([50, 80]));
+    seen.length = 0;
+
+    rerender(view([60, 90]));
+
+    // One commit, already showing the new ladder. Two entries starting with
+    // '50, 80' is the old effect-driven seed, and is what made the box
+    // clobberable mid-keystroke.
+    expect(seen).toEqual(['60, 90']);
   });
 });
--- a/apps/web/src/components/settings/AiBudgetThresholdsInput.test.tsx
+++ b/apps/web/src/components/settings/AiBudgetThresholdsInput.test.tsx
@@ -4,15 +4,25 @@ import { describe, expect, it, vi } from 'vitest';
 import AiBudgetThresholdsInput from './AiBudgetThresholdsInput';
 
 /**
- * Records the input's committed DOM value once per commit. Layout effects run
- * synchronously in the mutation phase of the very commit that produced the
- * DOM, so this observes what the box showed at each commit boundary without
- * depending on when React's scheduler gets around to passive effects.
+ * Records the input's committed DOM value. Layout effects run synchronously in
+ * the mutation phase of the very commit that produced the DOM, so this reads
+ * what the box actually showed at that commit without depending on when
+ * React's scheduler gets around to passive effects.
+ *
+ * It appends once per commit that re-renders THIS probe — i.e. once per
+ * `rerender()` from the test, not once per commit anywhere in the tree. That
+ * is exactly the question being asked: did the commit carrying the new prop
+ * already show the new ladder, or did the box need a later, box-local commit
+ * to catch up?
+ *
+ * Deliberately unguarded: if the selector ever stops matching, this throws with
+ * the reason rather than quietly recording one fewer entry.
  */
 function CommitProbe({ testId, seen }: { testId: string; seen: string[] }) {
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (React state-logic correctness, all four validity-notification orderings, both consumers) and `pr-test-analyzer` (coverage of the new render/effect split).

**Findings:** 6 raised → 6 addressed in 069d05f2; 0 outstanding.

- *Blocking (code-reviewer)* — the render-phase reset compared array **identity**, which reproduces the same clobber class through a different door. Confirmed reachable in the shipped UI, not theoretical: `AiUsagePage.fetchData()` re-runs on the `showFlaggedOnly` checkbox and unconditionally rebuilds `budget.alertThresholdPercents`, so a user mid-edit would lose their draft; `PartnerSettingsPage.fetchPartner` has the same shape. Now compares the rendered string, matching the `InvoiceEditor` precedent this change mirrors.
- *Severity 8 (pr-test-analyzer)* — the reconciling effect (the half of the split that releases a caller's Save gate after a prop-driven reset) was untested. Added `releases the caller when value resets over unparseable text`, and mutation-checked it: deleting that effect fails this test and only this test.
- *Severity 6* — added `does not clobber uncommitted text across an unrelated re-render` (the motivating scenario stated positively).
- *Severity 5* — added `re-seeds to empty when value resets to undefined` (the `?.`/`??` branch nothing else reached).
- *Nits ×2* — `CommitProbe` now throws instead of skipping when its selector misses, and its doc says what its cardinality actually means.

Verified fine with no action: render-phase `setState` termination and StrictMode/discarded-render safety, all four validity-notification orderings including unmount-with-unparseable-text, `invalidRef` never read or written during render, and neither consumer constructing a fresh array every render.

**Tests:** `vitest run src/components/settings` → 85 files / 990 tests passed. `tsc --noEmit -p apps/web/tsconfig.json` clean; `eslint` on both changed files clean.

**Flake evidence:** on `main`'s component, 40 iterations of `AiUsagePage.test.tsx` (`--pool=threads --maxWorkers=1`) under 8 busy-loops on an already-saturated 18-core box reproduced the CI assertion byte-for-byte (`sends the edited ladder`, expected `[60, 90]`, received `null`, same `AiUsagePage.test.tsx:244:49`). Under identical load after the fix: 80/80 green, then 40/40 green again on the final code, plus 25/25 with both changed files together. The two discriminating unit tests are red on the pre-fix component and green after, so the fix does not rest on the loop alone.

**Status:** review-clean. CI was still in flight at hand-off (head `069d05f251aa2dccd376af22dd5f9a63eb3bff52`) — please confirm **Test Web** green before merging. Not merged and #4659 not closed; both are the maintainer's call.

PR #4819, opened 2026-09-03
fix(web): reseed quote/invoice drafts during render, not from a passive effect (#4807)

### LINKED ISSUE #4807: [Web][Billing] prop→state sync via passive useEffect in 6 quote/invoice components — same clobber race as #4659 / #2925
**Pattern:** a component keeps a local draft in `useState` and re-syncs it from a prop in a passive `useEffect` (`useEffect(() => setText(value…), [value])`). The effect flushes after commit, so a keystroke that lands between the prop change and the effect is overwritten with the stale string. Under load the user's edit is silently reverted.

**History:** InvoiceEditor notes/terms had it (#2925 → #3219 → #3277 → #3980 → #4033, five filings before the effect was recognised). `AiBudgetThresholdsInput` had it (#4659, fixed in PR #4805 by reconciling on the rendered string instead of a passive effect).

**Remaining sites (found by the #4659 fixer, not yet swept):**
- `apps/web/src/components/billing/quotes/QuoteEditor.tsx:419`, `:437`
- `apps/web/src/components/billing/quotes/QuoteHeaderMeta.tsx:54`, `:135`
- `apps/web/src/components/billing/quotes/QuoteLineRows.tsx:697-699`
- `apps/web/src/components/billing/InvoiceDetail.tsx:104`

No flake has been observed from these yet, but they are the same shape and each is a live-edit surface (a quiet refetch rebuilds the prop array while the user types).

**Fix:** apply the PR #4805 / #4033 approach at each site — reconcile from the prop only when the rendered value actually differs from the last-applied prop value (compare the string, not array identity), never via an unconditional passive effect — and add the discriminating unit test pattern from #4805 (type, then let a prop refetch land, assert the draft survives).

ToddHebebrand:
Closes #4807

## Same clobber race as #4659 / #2925

An unconditional `useEffect(() => setState(prop), [prop])` re-syncs a
live-edit draft from a prop **after commit**. When the prop's persisted value
genuinely changes (a background refetch, a sibling field's save triggering
`refresh()`, etc.), the render that delivers the new prop still shows the
*old* draft; only the deferred passive effect corrects it, one commit later.
A keystroke landing in that one-commit window is silently reverted by the
stale value the effect captured — reproduced and fixed the same way for
`AiBudgetThresholdsInput` in #4805.

## What changed

Applied the #4805 / #4033 render-phase reseed pattern (compare the rendered
value against a `xSeededFrom` sentinel; update state **during render**, not
in an effect, when it differs) to 5 of the 6 sites the #4659 fixer flagged:

- `apps/web/src/components/billing/quotes/QuoteEditor.tsx` — `terms`/`termsDirty` (was line 419), `depositPercentDraft` (was line 437)
- `apps/web/src/components/billing/quotes/QuoteHeaderMeta.tsx` — `title`/`titleDirty` (was line 54), `customerOrgId` (was line 135)
- `apps/web/src/components/billing/InvoiceDetail.tsx` — `dueDateDraft` (was line 104)

## `QuoteLineRows.tsx` — verified, left untouched

The issue also listed `QuoteLineRows.tsx:697-699` (qty/price/cost, plus
sku/partNumber and name/desc share the same shape). I verified this site
before touching it, per the issue's caveat, and concluded it is **not**
exposed to the same race, so I left it alone:

- Each of those fields is guarded by a `*Edited` ref (`qtyEdited`,
  `priceEdited`, `costEdited`, `skuEdited`, `partEdited`, plus
  `nameEdited`/`descEdited`) that the resync effect checks before reseeding.
- The guard is cleared **synchronously at commit-time** (inside `commitQty`/
  `commitPrice`/etc., which run on blur, immediately — not gated on the
  async PATCH's resolution as I'd initially been told). That's a correction
  worth noting: the guard is *not* "cleared only on confirmed save."
- What actually closes the race: any subsequent keystroke's `onChange`
  handler flips the guard back to `true` **synchronously**, before any
  passive effect gets a chance to run. Since a real user keystroke can't
  land inside the sub-millisecond window between a commit and its passive
  effect flush, there's no realistic window left for the effect to fire
  while a ref is `false` AND the user is mid-typing a value the effect would
  clobber. This is structurally different from the fixed sites, which had no
  guard ref at all.
- I did not find a way to write a test that goes red against the current
  `QuoteLineRows.tsx` code (i.e. a discriminating test), which is itself
  evidence the guard closes the window in practice.

## Tests

Each fixed site gets a **commit-timing regression test** — a sibling
`useLayoutEffect` probe (the same technique #4805 introduced for
`AiBudgetThresholdsInput`) that reads the field's live DOM value at every
commit the probe participates in. The assertion is that a `rerender()` with
a genuinely-changed prop shows the **new** value in that single commit, with
no earlier commit still showing the stale one:

- `QuoteEditor.test.tsx`: `re-seeds a changed termsAndConditions prop within the same commit...`, `re-seeds a changed depositPercent prop within the same commit...`
- `QuoteHeaderMeta.title.test.tsx`: `re-seeds a changed title prop within the same commit...`
- `QuoteHeaderMeta.customer.test.tsx`: `re-seeds a changed quote.orgId prop within the same commit...`
- `InvoiceDetail.test.tsx`: `re-seeds a changed dueDate prop within the same commit...`

I verified each of these five tests is genuinely discriminating: I
temporarily reverted each fix back to the old passive-`useEffect` code and
confirmed the corresponding new test goes **red** (e.g. `expected ['Net 30'] to
deeply equal ['Net 45']`), then restored the fix and confirmed green again.

Also added, per the issue's requested pattern (type a draft, let a prop
refetch land, assert the draft survives): "keeps a typed terms/title/due-date
draft when an unrelated refetch hands back the same [unchanged] value" tests
for the terms and title and due-date fields. (Being upfront: because these
props are plain strings, an *unchanged* value never triggers the old effect
either — Object.is already short-circuits it — so these particular tests
don't discriminate old vs. new code on their own. They're included as
baseline regression coverage since the issue explicitly asked for this
pattern; the commit-timing tests above are what actually pins down the bug
and fails on the old code.)

## Verification

- `apps/web`: affected suites green — `QuoteEditor.test.tsx` (8), 
  `QuoteHeaderMeta.title.test.tsx` (5), `QuoteHeaderMeta.customer.test.tsx`
  (7), `InvoiceDetail.test.tsx` (24). 44/44 passed.
- `npx tsc --noEmit -p apps/web` — clean, exit 0.
- `npx eslint` on all 7 changed files — clean, no output.

### DIFF
--- a/apps/web/src/components/billing/InvoiceDetail.test.tsx
+++ b/apps/web/src/components/billing/InvoiceDetail.test.tsx
@@ -1,4 +1,5 @@
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { useLayoutEffect } from 'react';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import InvoiceDetail from './InvoiceDetail';
@@ -47,6 +48,22 @@ const issued: InvoiceDetailData = {
   },
   lines,
 };
+
+/**
+ * Records the due-date input's committed DOM value. A layout effect runs
+ * synchronously in the mutation phase of the very commit that produced the
+ * DOM, so it reads what the field actually showed at that commit without
+ * depending on when React's scheduler gets around to passive effects — same
+ * technique used to pin down #4659 (AiBudgetThresholdsInput).
+ */
+function CommitProbe({ testId, seen }: { testId: string; seen: string[] }) {
+  useLayoutEffect(() => {
+    const el = document.querySelector(`[data-testid="${testId}"]`) as HTMLInputElement | null;
+    if (!el) throw new Error(`CommitProbe: no element matching [data-testid="${testId}"]`);
+    seen.push(el.value);
+  });
+  return null;
+}
 
 describe('InvoiceDetail', () => {
   beforeEach(() => {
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer (all Sonnet, parallel).

**Findings:** 2 Important raised by pr-test-analyzer → both addressed in fc2a26d55 (added `clears the Unsaved badge when a genuinely different termsAndConditions prop lands mid-edit` and `clears the inline deposit-percent range error when a genuinely different depositPercent prop lands` to `QuoteEditor.test.tsx`); 0 outstanding. code-reviewer and comment-analyzer raised 0 findings — both confirmed the render-phase reseed pattern is applied correctly at all 5 sites and independently verified the `QuoteLineRows.tsx` non-fix (ref guards clear synchronously at blur, closing the race without the render-phase pattern).

**Tests:** `apps/web` affected suites green — `QuoteEditor.test.tsx` (10), `QuoteHeaderMeta.title.test.tsx` (5), `QuoteHeaderMeta.customer.test.tsx` (7), `InvoiceDetail.test.tsx` (24); 46/46 passed. `npx tsc --noEmit -p apps/web` clean (exit 0). `npx eslint` on all changed files clean.

PR #4820, opened 2026-09-03
fix(api/portal): rate-limit password-change guesses, stop 401ing wrong ones (#4797)

### LINKED ISSUE #4797: [API][Portal] POST /portal/profile/password has no rate limit on current-password guesses, and 401s a wrong guess (logging the user out)
Two defects on the portal's change-password route, both already fixed on the equivalent partner/tech route. Surfaced by the security review of PR #4782 (#4746) and verified against `origin/main` 383a10f9.

**Where:** `apps/api/src/routes/portal/profile.ts:113` — `POST /profile/password` (mounted under `/portal/profile/*` behind `portalAuthMiddleware`, `apps/api/src/routes/portal/index.ts:37`).

### 1. No rate limit on current-password guesses (the #4746 class)

`profile.ts:137` calls `verifyPassword(user.passwordHash, currentPassword)` inline with no limiter anywhere on the handler path — `grep -rn "rateLimiter" apps/api/src/routes/portal/` returns nothing. The only throttle is the blanket global per-IP limiter, which is keyed on IP rather than user and two orders of magnitude looser than the credential-specific bucket every sibling uses.

A stolen portal session can therefore brute-force the current password at one argon2 verify per request and, on a hit, set a new one — account takeover with no lockout, on a customer-facing surface.

`#4746` / PR #4782 fixed exactly this on `POST /auth/change-password` by routing it through `requireCurrentPasswordStepUp` (5 attempts / 5 min, per user). **That helper is not directly reusable here:** it looks up `users` by id, while portal accounts live in `portal_users`. The fix needs either a portal-aware variant of the helper or a dedicated `rateLimiter` call in the shape `accountDeletion.ts:234` uses. Worth deciding deliberately rather than copy-pasting — a shared helper parameterised by table would stop this diverging a third time.

### 2. A wrong current password answers 401, which logs the user out (the #4660 class)

`profile.ts:139` returns `c.json({ error: 'Current password is incorrect' }, 401)`.

`currentPassword` arrives in the request **body** — it is data this handler validates, not the credential authenticating the request (that is the portal session cookie, already checked by `portalAuthMiddleware`). Answering 401 collides with the session guard, and the portal client funnels it into a logout:

- `apps/portal/src/lib/api.ts:210-233` — on a 401 the default path calls `clearAuth()` and redirects to `/login`.
- `apps/portal/src/components/portal/ProfileSettings.tsx:91` — the change-password caller does **not** pass `redirectOnUnauthorized: false`, so it takes that default path.

Net effect: a portal user who mistypes their current password is signed out mid-flow. This is the same bug #4660 fixed for `/auth/change-password` in #4739 — the portal route was missed.

Should become 400 with the stable `code: 'invalid_credentials'` from `rejectProof` (`apps/api/src/routes/auth/helpers.ts`), matching the contract in #4470/#4651/#4660. Note the neighbouring rejections on this same route are already 400 (`profile.ts:135` passwordless account, `:143` weak new password), so 401 is also internally inconsistent here.

### Scope note

Deliberately **not** folded into PR #4782, which was scoped to `/auth/change-password`: this is a different route, a different auth surface, and a different table, and the rate-limit fix needs a design decision (portal-aware helper vs. local limiter) rather than a mechanical copy.

ToddHebebrand:
## Summary

Two defects on `POST /portal/profile/password`, both already fixed on the equivalent partner/tech route (`/auth/change-password`, #4782/#4739):

1. **No rate limit on current-password guesses** (the #4746 class). `verifyPassword` was called inline with no throttle anywhere on the handler path — `grep -rn "rateLimiter" apps/api/src/routes/portal/` returned nothing. A stolen portal session could brute-force the current password at one argon2 verify per request and, on a hit, set a new one. `requireCurrentPasswordStepUp` isn't directly reusable here since it looks up `users`, not `portal_users`.

2. **A wrong guess answered 401**, which the portal client (`apps/portal/src/lib/api.ts`) funnels into `clearAuth()` + a redirect to `/login` — signing the user out over a typo instead of showing the error (the #4660 class, missed when #4739 fixed `/auth/change-password`).

## Fix

- Adds `PASSWORD_CHANGE_RATE_LIMIT` (5 attempts / 5 min) and throttles on it via the portal's own `checkRateLimit` helper (`apps/api/src/routes/portal/helpers.ts`) — the same dual-mode (Redis / in-memory) limiter every other portal auth route (login, forgot-password, reset-password) already uses, keyed on the portal user id since this route is already session-authed. Chose this over introducing the Redis-only, fail-closed `rateLimiter` from `apps/api/src/routes/auth/helpers.ts` so self-hosted deployments without Redis keep working, consistent with every sibling portal rate limit.
- Changes the wrong-password rejection from `401` to `400` with the stable `code: 'invalid_credentials'` contract (`rejectProof` from `apps/api/src/routes/auth/helpers.ts`), matching #4470/#4651/#4660 and the two neighbouring rejections on this same route (passwordless account, weak new password) which were already 400.
- No portal client changes needed: `apps/portal/src/lib/api.ts`'s generic `!response.ok` path already surfaces `body.error`/`body.code` for any non-401 status, so the fix is purely server-side.

## Test plan

- [x] New suite `apps/api/src/routes/portal/profile.password.test.ts`: happy path still succeeds; wrong guess → 400 with the stable code (never 401); 6th guess in a window → 429 with `Retry-After`; a 429 never calls `verifyPassword` (doesn't leak correctness); throttling is per portal-user-id, not global.
- [x] `npx vitest run src/routes/portal/` — 15 files / 213 tests passed (no regressions).
- [x] `npx tsc --noEmit` on `apps/api` — clean.

Closes #4797

### DIFF
--- a/apps/api/src/routes/portal/profile.password.test.ts
+++ b/apps/api/src/routes/portal/profile.password.test.ts
@@ -1,0 +1,251 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Hono } from 'hono';
+
+// #4797: POST /profile/password had no rate limit on current-password guesses
+// and answered a wrong guess with 401, which the portal client funnels into a
+// forced logout. This suite covers both fixes: the dedicated portal-native
+// rate limiter (checkRateLimit, in-memory mode here since PORTAL_USE_REDIS is
+// false outside production/PORTAL_STATE_BACKEND=redis) and the 401 -> 400
+// + stable `code` change.
+
+const verifyPasswordMock = vi.fn(async (_hash: string, _plaintext: string) => true);
+const hashPasswordMock = vi.fn(async (_plaintext: string) => 'new-hash');
+
+const dbState: { userRow: unknown } = {
+  userRow: {
+    id: 'pu-1',
+    passwordHash: 'argon2-hash',
+    email: 'customer@example.com',
+    orgId: 'org-1',
+    name: 'Customer',
+  },
+};
+
+const updateReturningMock = vi.fn(async () => [dbState.userRow]);
+
+vi.mock('../../db', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../db')>();
+  return {
+    ...actual,
+    db: {
+      select: vi.fn(),
+      update: vi.fn(),
+    },
+  };
+});
+
+vi.mock('../../db/schema', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../db/schema')>();
+  return {
+    ...actual,
+    portalUsers: {
+      id: 'portal_users.id',
+      orgId: 'portal_users.org_id',
+      email: 'portal_users.email',
+      name: 'portal_users.name',
+      passwordHash: 'portal_users.password_hash',
+      receiveNotifications: 'portal_users.receive_notifications',
+      status: 'portal_users.status',
+    },
+  };
+});
+
+vi.mock('../../services/password', () => ({
+  hashPassword: (plaintext: string) => hashPasswordMock(plaintext),
+  verifyPassword: (hash: string, plaintext: string) => verifyPasswordMock(hash, plaintext),
+  isPasswordStrong: () => ({ valid: true, errors: [] }),
+}));
+
+vi.mock('../../services/redis', () => ({
+  getRedis: vi.fn(() => null),
+}));
+
+vi.mock('./helpers', async () => {
+  const actual = await vi.importActual<typeof import('./helpers')>('./helpers');
+  return {
+    ...actual,
+    validatePortalCookieCsrfRequest: vi.fn(() => null),
+    writePortalAudit: vi.fn(),
+  };
+});
+
+import { profileRoutes } from './profile';
+import { db } from '../../db';
+import { portalRateLimitBuckets, writePortalAudit } from './helpers';
+
+const AUTH_USER = {
+  id: 'pu-1',
+  orgId: 'org-1',
+  email: 'customer@example.com',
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (Sonnet) — correctness, security, scope discipline, test quality
**Findings:** 0 blocking, 1 nitpick (passwordless-account branch could also carry `code: invalid_credentials` for consistency; left as-is — it's a config-error message, not a credential oracle, and was already 400 before this PR, so out of scope).
**Verified independently:**
- Rate-limit check runs before DB lookup / `verifyPassword` on the single linear code path (no early-return skips it).
- Key `portal:profile-password:${auth.user.id}` doesn't collide with any existing `PORTAL_REDIS_KEYS` entry or other `checkRateLimit` caller (`portal:login:*`, `portal:forgot:*`, `portal:reset:*`).
- `checkRateLimit`'s existing fail-closed behavior in production when Redis is unavailable is untouched — this PR only consumes it.
- 429 never calls `verifyPassword` and carries no `code` field, so it can't be used to distinguish a correct guess from an incorrect one.
- `requireCurrentPasswordStepUp` (the `users`-keyed helper) is untouched — no scope creep onto the `/auth/*` surface.
**Tests:** `apps/api/src/routes/portal/profile.password.test.ts` — 6/6 passed (`npx vitest run src/routes/portal/profile.password.test.ts`); full portal suite regression check `npx vitest run src/routes/portal/` — 15 files / 213 tests passed; `npx tsc --noEmit` on `apps/api` — clean.

PR #4823, opened 2026-09-03
fix(api): make org merge honor the canonical ticket lock order (#4748)

### LINKED ISSUE #4748: [API][Concurrency] orgMerge walks ticket child tables in a different order than the org-move lock order (ticket_attachments before ticket_alert_links)
**Context:** PR #4741 (fix for #4657) pins both org-move paths (ticket-axis `moveTicketOrg`, device-axis `routes/devices/moveOrg.ts`) to one canonical lock order in `apps/api/src/services/ticketOrgMoveLockOrder.ts`: `time_entries → ticket_parts → ticket_alert_links → ticket_attachments`.

**Gap:** `apps/api/src/services/orgMerge.ts` runs the whole merge in one transaction over the reversed topological order from `orgMergeRegistry`, which yields `ticket_attachments` before `ticket_alert_links` — disagreeing with the canonical order on the last pair. The merge is held apart from the movers only by `fenceLoser()`'s `status='merging'` app-layer fence, so a move already in flight when the fence lands can still contend under AB-BA ordering (40P01).

**Fix options:** (a) make `orgMergeRegistry`'s ordering contract honour the shared lock-order constant for the ticket child tables, with a contract test that the merge walk agrees with `ticketOrgMoveLockOrder`; or (b) have the merge take the same fence check the movers take so an in-flight move blocks the merge rather than racing it. (a) is the long-term answer.

Documented (not papered over) in `ticketOrgMoveLockOrder.ts`. Verified by the #4657 fixer against origin/main; out of that PR's scope.

ToddHebebrand:
## Summary

`orgMerge.ts` walks the org-merge table set in `reversed(topologicalCascadeOrder())` order. That reversal ties on alphabetical order for tables with no FK between each other — and the six ticket child tables that denormalize `org_id` (`time_entries`, `ticket_parts`, `ticket_alert_links`, `ticket_outbox`, `ticket_attachments`, `ticket_email_links`) are exactly such a set (documented FK note in `ticketOrgMoveLockOrder.ts`: none of the six references another).

That alphabetical tie-break disagrees with `TICKET_CHILD_ORG_REWRITE_LOCK_ORDER` — the lock order the ticket-move (`moveTicketOrg`) and device-move (`POST /devices/:id/move-org`) axes already agree on (#4657), for the exact same AB-BA (40P01) reason. The merge is held apart from an in-flight move only by `fenceLoser()`'s app-layer `status='merging'` fence, not by lock order — so a move that starts before the fence lands can still contend with the merge under opposite lock order.

## Fix

- `applyTicketChildLockOrder()` (new, `ticketOrgMoveLockOrder.ts`): given any table walk, permutes only the ticket-child tables' own index slots into `TICKET_CHILD_ORG_REWRITE_LOCK_ORDER`'s relative order — every other table keeps its position. Safe to apply unconditionally because the six share no FK with each other, so permuting them never breaks children-before-parents correctness for anything else in the walk.
- `orgMerge.ts` gets a small `getOrgMergeWalkOrder()` helper that wraps the existing `[...(await topologicalCascadeOrder())].reverse()` with the new function, and both walk-order consumers (the merge transaction and the merge preview) now call it instead of duplicating the raw computation.

## Tests

Added a red-first contract suite to `ticketOrgMoveLockOrder.test.ts` (pure, no DB — same style as the existing #4657 contract in that file):
- Demonstrates the gap: naive reverse-alphabetical tie-break disagrees with the canonical order.
- Pins `applyTicketChildLockOrder`'s behavior directly: reorders only the ticket-child slots, no-ops when already canonical, handles a partial subset.
- A source-grep static contract asserting `orgMerge.ts` has exactly one raw `topologicalCascadeOrder().reverse()` (inside `getOrgMergeWalkOrder()`) and both consumption sites call `getOrgMergeWalkOrder()` — so a future regression back to a bare reversed order fails in the **Test API** unit job instead of only surfacing as a live deadlock.

Verified locally:
- `npx vitest run src/services/ticketOrgMoveLockOrder.test.ts src/services/orgMerge.test.ts src/routes/orgMerge.test.ts src/jobs/orgMerge.test.ts` — 116 passed.
- `npx vitest run src/routes/devices/moveOrg.test.ts src/services/ticketService.test.ts` — 236 passed (unaffected by this change, confirms the two movers still agree).
- `npx tsc --noEmit -p apps/api` — clean.
- `npx eslint` on the three touched files — clean.

Closes #4748

### DIFF
--- a/apps/api/src/services/orgMerge.ts
+++ b/apps/api/src/services/orgMerge.ts
@@ -53,6 +53,7 @@ import {
   type MergeTableOutcome,
 } from './orgMergeCustomExecutors';
 import { getOrgMergePolicies, type OrgMergePolicy } from './orgMergeRegistry';
+import { applyTicketChildLockOrder } from './ticketOrgMoveLockOrder';
 import { topologicalCascadeOrder } from './tenantCascade';
 import { envInt } from '../utils/envInt';
 import { disconnectLiveAgentSocketsForOrgIds } from './tenantLifecycle';
@@ -948,6 +949,18 @@ export function buildMergeWarnings(input: MergeWarningInput): string[] {
   return warnings;
 }
 
+/**
+ * The parents-first table walk both the merge transaction and the preview
+ * use: `topologicalCascadeOrder()` (children-before-parents, i.e. erasure
+ * order) reversed, then corrected so the ticket child tables land in the
+ * SAME relative order the ticket/device org-move axes lock them in
+ * (#4748) — see `applyTicketChildLockOrder` for why the correction is
+ * necessary and why it's always FK-safe to apply here.
+ */
+async function getOrgMergeWalkOrder(): Promise<string[]> {
+  return applyTicketChildLockOrder([...(await topologicalCascadeOrder())].reverse());
+}
+
 // ---------------------------------------------------------------------------
 // Execute
 // ---------------------------------------------------------------------------
@@ -1031,8 +1044,10 @@ export async function executeOrgMerge(input: ExecuteOrgMergeInput): Promise<OrgM
         const policies = getOrgMergePolicies();
         // topologicalCascadeOrder is children-before-parents (erasure order);
         // reversed it is parents-first, with `organizations` (loser-shell,
-        // a no-op) leading.
-        const order = [...(await topologicalCascadeOrder())].reverse();
+        // a no-op) leading. getOrgMergeWalkOrder() then corrects the ticket
+        // child tables' relative order to match the movers' lock order
+        // (#4748) — see its doc comment.
+        const order = await getOrgMergeWalkOrder();
 
         const summary: Record<string, OrgMergeCounts> = {};
         const notes: string[] = [];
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit code-reviewer + silent-failure-hunter (Sonnet), independently, against the full diff vs `origin/main`.
**Findings:** 0 blocking, 0 should-fix. 2 nits raised by silent-failure-hunter — (1) missing direct unit test for `applyTicketChildLockOrder`'s duplicate-entry throw branch, addressed inline in `cad951747` (`throws on a duplicate ticket child-table entry rather than silently mis-permuting`); (2) a cosmetic error-message misattribution in a currently-unreachable scenario (the canonical constant would need an accidental literal duplicate) — not fixed, judged not worth the complexity for a case that can't occur with today's data.
**Correctness verification (code-reviewer):** independently confirmed the "no FK between the six ticket-child tables" claim against the schema files and the composite-FK migration (`2026-10-06-110100-ticket-child-org-fks.sql`); confirmed both `orgMerge.ts` walk-order call sites (`executeOrgMerge` and `previewOrgMerge`) route through the new `getOrgMergeWalkOrder()`; grepped for a third org-merge walk-order computation site — none found.
**Error-propagation verification (silent-failure-hunter):** traced both call paths — a throw from `applyTicketChildLockOrder` propagates uncaught through `executeOrgMerge`'s outer catch (which re-throws after writing the failure audit) to the BullMQ worker's `failed` handler (Sentry + log), and through `previewOrgMerge`'s route handler to Hono's global `onError` (Sentry + 500). No silent swallow.
**Tests:** `apps/api` affected suite green — `npx vitest run src/services/ticketOrgMoveLockOrder.test.ts src/services/orgMerge.test.ts src/routes/orgMerge.test.ts src/jobs/orgMerge.test.ts` (127 passed after the follow-up commit), plus `src/routes/devices/moveOrg.test.ts src/services/ticketService.test.ts` (236 passed, confirming the two existing movers are unaffected). `npx tsc --noEmit -p apps/api` clean. `npx eslint` on the three touched files clean.
**CI:** queued on GitHub Actions at the time of this comment (runner backlog — all jobs still `queued`, not failed) for commit `cad951747`; please re-check `gh pr checks 4823` before merging.

Issue #4824, opened 2026-09-03
Default audit_retention_policies row on org creation


ToddHebebrand:
### Context

#4633 added a settings surface (GET/PUT `/orgs/organizations/:id/audit-retention`, gated on the new `audit:manage` permission) so an operator can create/edit an `audit_retention_policies` row for their org. That closes the "no way to create one" gap, but a **fresh org still starts with no row at all** — the daily prune worker (`apps/api/src/jobs/auditRetention.ts`) simply skips any org without one, so retention stays a silent no-op until someone visits Settings > Security & Access > Audit Retention and saves a value.

### Ask

Seed a default `audit_retention_policies` row (365 days, matching the column default) whenever a new organization is created, so retention is active from day one without requiring a manual settings visit.

This needs auditing across every org-creation call site, not just the primary "create organization" route — the fan-out risk called out in CLAUDE.md for cross-cutting org-lifecycle changes (partner onboarding, org import/migration, any test/seed helper that creates orgs outside the normal API path, etc).

### Deferred from #4633

Scoped out of that PR to keep it to the settings surface itself (route + permission + UI). Filing separately per the plan there.

PR #4821, opened 2026-09-03
fix(api/tenancy): sever ticket_comments.agent_run_id on a device org-move (#4644)

ToddHebebrand:
## Summary

Closes #4644.

`ticket_comments.agent_run_id` (loop-guard link to the `ai_agent_runs` row that authored a comment) is a reverse pointer: `ticket_comments` has no `org_id` of its own (child-via-parent tenancy through `tickets`). On a **device** org-move, `tickets` is re-stamped to the target org by `breeze_device_child_orgid_tables()`, so every comment on a ticket bound to the moving device travels with it — while `ai_agent_runs.org_id` deliberately stays with the source org (owner decision 2026-08-23). A retained `agent_run_id` then names a source-org run from a target-org row.

Same reverse-pointer class #3828 fixed for `metric_anomaly_incidents.agent_run_id`, and #4215 fixed for `ai_agent_runs.ticket_id`, both on this device axis. #4642 already closed this exact column on the **ticket** axis (`moveTicketOrg`); this closes the **device** axis's DB-side cascade to match.

**Review-round correction:** the issue's own framing described this as "latent, not live" (nothing writes the column yet). PR review found that premise is false: `addAiTriageNote()` (`services/ticketService.ts`, P2-4a #4300) has been a real, unflagged writer of `ticket_comments.agent_run_id` since before this fix — every AI-agent ticketing `comment` tool call that carries an `agentRunId` inserts a live value. So a ticket that received an AI-authored comment and then had its bound device (or the ticket itself) moved to another org, on either axis, before that axis's severing fix shipped, could already carry a stale cross-org `agent_run_id` in production today. This PR now includes a backfill for that.

## Changes

- `apps/api/src/routes/devices/moveOrg.ts`: nulls `ticket_comments.agent_run_id` for comments whose ticket is bound to the moving device, via the same `ticket_id IN (SELECT id FROM tickets WHERE device_id = ...)` join the `ai_agent_runs.ticket_id` detach beside it already uses.
- New migration `2026-10-08-100300-detach-ticket-comment-runs-on-device-org-move.sql`:
  - Replaces `breeze_cascade_device_org_id()` (the DB-side path for direct-SQL callers such as orgMerge) with the same statement added, copied verbatim from the newest prior definition.
[...]

PR #4818, opened 2026-09-03
fix(api): write an intent_outbox row when cancelActionIntent commits (#4798)

ToddHebebrand:
## Summary

`cancelActionIntent` only performed the `pending_approval|approved -> cancelled` CAS and never wrote an `intent_outbox` row, unlike `createActionIntent` (which writes `intent_created` / `intent_approved` in the same transaction). Because `approved -> cancelled` is a reachable transition, a requester who already received the "was approved and is now running" outcome notification from `intentReleaseWorker.ts` was never told the intent was subsequently cancelled before it ran — the notification dedupe fix in #4785 (#4465) keeps a distinct `cancelled` dedupe class precisely so a later cancel *would* notify, but no event ever reached the worker to trigger it.

## Changes

- `cancelActionIntent` (`apps/api/src/services/actionIntents/intentService.ts`) now inserts an `intent_cancelled` outbox row in the same `withSystemDbAccessContext` transaction as the CAS, mirroring the `intent_created`/`intent_approved` pattern in `createActionIntent`. Inlined the CAS predicate rather than extending the shared `transitionIntent` primitive, since cancel is the only caller that needs the outbox write and `transitionIntent` is shared by `intentReleaseWorker.ts` / `aiAgentSdk.ts` release paths that don't.
- `intentReleaseWorker.ts`'s event-type dispatch (`processIntentReleaseJob`) now routes `intent_cancelled` into the existing outcome-only branch alongside `intent_rejected` / `intent_expired`, notifying the requester (`intent-outcome:` dedupe key) or agent recipients (`agent-intent-outcome:` dedupe key) via the pre-existing paths — no new notification logic needed, since `outcomeNotificationClass` already had a `cancelled` class from #4785.
- Widened the `intent_outbox_event_type_check` CHECK constraint (new migration `2026-10-08-100300-intent-cancelled-outbox-event.sql`) and the `IntentOutboxEvent` schema union (`apps/api/src/db/schema/actionIntents.ts`) to admit `'intent_cancelled'`.
- Updated a stale comment in `intentReleaseWorker.ts` that documented the previous "cancel writes no outbox row" behavior as a caveat.

## Tests

- `apps/api/src/services/actionIntents/intentService.test.ts`: two new `cancelActionIntent` tests — writes the outbox row on a successful CAS, writes nothing when the CAS loses the race. Verified red-then-green (confirmed the first failed against unmodified code before implementing the fix).
[...]

PR #4825, opened 2026-09-03
fix(api): add audit-log retention settings route and UI (#4633)

### LINKED ISSUE #4633: [Settings] No API route or UI to create audit_retention_policies rows; audit retention never runs on a fresh install
### Problem

`apps/api/src/jobs/auditRetention.ts` walks `audit_retention_policies` daily, but the table ships with zero rows and nothing in `apps/api/src/routes` or the web settings creates one (verified by grep on 2026-09-02: the only references are the cascade, merge and export registries). So on every self-hosted install audit log retention is a no-op unless the operator inserts a row by hand in psql. Discussion #4373 (Baltix, ~15 agents on a small VPS, 0.108.0) hit this: 269 MB `audit_logs`, 128 MB `audit_log_chain`, empty policy table, docs at `/reference/audit-logs/` say retention exists but not how to set it.

### Ask

- Partner-level settings surface (Settings > Partner or Security) to set audit retention days, writing one `audit_retention_policies` row per org (or a partner-wide default, per the partner-wide-first rule in CLAUDE.md, which would need a migration adding `partner_id` XOR `org_id`).
- API route with `audit` admin permission, validator in `packages/shared`.
- Update https://docs.breezermm.com/reference/audit-logs/ with the setting and the interim SQL.
- Consider a sane default row on org creation (365 days is the column default) so fresh installs prune at all.

Related: #4340 (agent telemetry dominates audit_logs), #4342 (tables with no retention).

ToddHebebrand:
## Summary

- `audit_retention_policies` shipped with the daily prune worker (`apps/api/src/jobs/auditRetention.ts`) but nothing in the API routes or web UI ever created or edited a row — verified by grep, the only references were the cascade/merge/export registries. Every self-hosted install ran retention as a silent no-op unless an operator inserted a row by hand in psql (Discussion #4373: 269 MB `audit_logs`, 128 MB `audit_log_chain`, empty policy table).
- Adds `GET`/`PUT /orgs/organizations/:id/audit-retention`, gated on a new `audit:manage` permission (distinct from `audit:read` — lowering retention is a higher-trust action than reading the trail). Granted to Org Admin by default; a migration backfills the permission row and role grant for already-deployed installs (seed-time defaults only apply on a fresh DB with zero users).
- Adds a validator (`auditRetentionPolicySchema`, 1–3650 days, matching the existing backup-retention convention) in `packages/shared`.
- Adds a **Settings > Security & Access > Audit Retention** UI section that surfaces whether a policy is actually configured — `configured: false` renders a warning banner rather than silently showing the column default (365 days) as if it were active.
- Updates `docs/reference/audit-logs.mdx` with the new route and an interim SQL fallback for installs on an older version.

## Design decisions

- **Table stays org-scoped.** It already shipped with `org_id NOT NULL` and no `partner_id`, so this does not retrofit the partner-wide dual-axis shape (CLAUDE.md's "Partner-Wide First" rule) — that would be a much bigger migration for a table that's been org-scoped since the baseline. RLS and the org-cascade list already fully cover it; no schema migration needed.
- **No unique constraint added.** `audit_retention_policies` has no unique index on `org_id`. Rather than add one, the upsert (`upsertOrgAuditRetentionPolicy`) uses a transactional `SELECT ... FOR UPDATE` to get one-row-per-org semantics without a migration.
- **Deferred:** seeding a default row on org creation. That needs sweeping every org-creation call site (partner onboarding, org import, test/seed helpers), which is its own piece of work — filed as #4824 rather than folding into this PR.

## Test plan

- [x] `apps/api/src/routes/orgAuditRetentionSettings.test.ts` — new route tests (GET/PUT, 401/404, validation bounds, audit event) — 12 passed
- [x] `apps/api/src/db/seed.test.ts` — new `audit:manage` permission/role-grant assertions — 154 passed
- [x] `apps/api/src/routes/permissionsCatalog.test.ts` — self-adjusting catalog parity — 2 passed
- [x] `apps/api/src/jobs/auditRetention.test.ts` — unchanged worker behavior — 33 passed
- [x] `apps/web/src/components/settings/OrgAuditRetentionSettings.test.tsx` — new component tests (load, save, permission gating, error states) — 11 passed
- [x] `apps/web/src/components/settings/OrgSettingsPage.test.tsx` — tab wiring — 21 passed
- [x] `apps/web/src/lib/i18n/localeParity.test.ts` — all 8 locales — 79 passed
- [x] `apps/web/src/lib/__tests__/no-silent-mutations.test.ts` — 130 passed
- [x] `packages/shared/src/validators/auditRetention.test.ts` — 7 passed
- [x] `apps/api` and `apps/web` `tsc --noEmit` clean
- [x] `scripts/check-migration-naming.sh` — OK
- [x] Migration hand-applied to a live Postgres test DB (`breeze-postgres-test`), confirmed idempotent on re-run, then reverted to leave the shared fixture clean

Closes #4633


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, silent-failure-hunter, pr-test-analyzer — all sonnet, parallel)

**Findings:** 3 independently-confirmed real issues → all addressed in 0343847ba; 0 outstanding.

- **Critical (code-reviewer + silent-failure-hunter, converged independently):** `upsertOrgAuditRetentionPolicy`'s `SELECT ... FOR UPDATE` only locks *existing* rows — for an org with no policy row yet (the exact case #4633 exists to fix), two concurrent `PUT`s both see "no row" and both `INSERT`, producing duplicates the prune worker then double-processes with no dedup. Fixed: new migration `2026-10-08-100700-audit-retention-policies-org-unique.sql` adds a real `UNIQUE` constraint on `org_id` (deduping any pre-existing duplicate first, logging the count), and the service now does a single atomic `INSERT ... ON CONFLICT (org_id) DO UPDATE` — the same pattern `org_ticket_settings` already uses. Hand-verified against a live Postgres test DB: manufactured a duplicate row, re-ran the migration, confirmed it kept the more-recently-updated row and logged `RAISE WARNING removed 1 duplicate...`, then reverted the DB to its original state.
- **High (silent-failure-hunter):** the web component re-fetched via `load()` after a successful save; if that reload failed (transient network blip), the panel would flip to the full-page error screen right after the "Saved" toast, making a successful save look failed. Fixed: the component now applies the PUT response directly instead of re-fetching.
- **Important (code-reviewer):** the docs' interim SQL fallback used `ON CONFLICT DO NOTHING` while telling the reader "there is no unique constraint" — inert advice presented as a safety net. Now that the constraint above exists, updated the snippet to `ON CONFLICT (org_id) DO NOTHING` (which now actually works) with a genuine no-op guarantee.
- **Suggestions actioned:** added the dedicated `auditRetentionPolicyService.test.ts` unit tests that were missing (the route test only exercised a fully-mocked service — pr-test-analyzer flagged this as the one real coverage gap), locale-formatted `lastCleanupAt` instead of a raw ISO string, and a missing 1-day boundary case in the validator tests.

**Tests:** `apps/api` — 285 passed across 7 affected files (`auditRetentionPolicyService.test.ts` new, `orgAuditRetentionSettings.test.ts`, `seed.test.ts`, `auditRetention.test.ts`, `autoMigrate.test.ts`, `permissionsCatalog.test.ts`, `orgTicketSettings.test.ts`). `apps/web` — 241 passed across 4 affected files (component test, `OrgSettingsPage.test.tsx`, `localeParity.test.ts` all 8 locales, `no-silent-mutations.test.ts`). `packages/shared` — 8 passed (validator). `tsc --noEmit` clean on `apps/api` and `apps/web`. `scripts/check-migration-naming.sh` — OK. No tenancy/RLS/cascade changes (table's shape and registrations were already correct), so the integration/RLS contract suites weren't required, per CLAUDE.md — but the new unique-constraint migration was hand-applied and reverted against a live test DB as extra verification given it touches a real schema change.

ToddHebebrand:
Renamed migrations to `2026-10-08-100600-audit-retention-manage-permission.sql` / `2026-10-08-100700-audit-retention-policies-org-unique.sql` so they sort after the newly-merged `2026-10-08-100300-intent-cancelled-outbox-event.sql` (#4818) and #4821's `100500-...`; swept all internal comment references and corrected the earlier review comment's stale filename; `check-migration-naming.sh`, `autoMigrate.test.ts` (67 passed), and the PR's affected API tests (218 passed) all green on the rebased branch.

PR #4826, opened 2026-09-03
fix(api): revoke UPDATE on peripheral_policy_delivery_events breeze_app grant (#4806)

### LINKED ISSUE #4806: peripheral_policy_delivery_events: append-only guarantee still depends on the trigger alone (breeze_app keeps UPDATE)
## Summary

`peripheral_policy_delivery_events` is the one table (of the six #4371 touched) whose append-only guarantee still depends **entirely on its `BEFORE UPDATE` trigger**, not on a privilege wall — `breeze_app` retains real `UPDATE` on it, unlike its five siblings (`pam_actuation_results`, `agent_rollback_events`, `ml_feedback_events` have `UPDATE` fully revoked; `automation_action_results`/`device_software_inventory_state` were never append-only by design).

## Why UPDATE couldn't be revoked in #4371

`routes/devices/moveOrg.ts`'s device org-move path issues a real `breeze_app` `UPDATE peripheral_policy_delivery_events SET org_id = ... WHERE device_id = ...` (via `getDeviceOrgDenormalizedTables()`) to restamp delivery-event evidence onto the target org when a device moves. The table's own migration (`2026-09-11-peripheral-effective-policy-v2.sql`) only revokes `DELETE, TRUNCATE` from `breeze_app` — `UPDATE` was deliberately left granted for exactly this restamp.

The append-only trigger (`peripheral_policy_delivery_events_append_only()`) allows that restamp through a **data-shape check**, not a caller check:
```sql
IF (to_jsonb(NEW) - 'org_id') = (to_jsonb(OLD) - 'org_id')
   AND EXISTS (SELECT 1 FROM devices WHERE id = NEW.device_id AND org_id = NEW.org_id)
THEN RETURN NEW;
END IF;
```
Any statement that only changes `org_id` to a value matching the row's device's *current* org passes, regardless of which code path issued it. So the "no field but org_id, and only via a real device move" guarantee rests entirely on this trigger logic being correct forever — there is no privilege-layer backstop the way there now is for `agent_rollback_events` and `pam_actuation_results`.

## Compare: how the sibling tables got a full privilege wall in #4371

- `agent_rollback_events` has no per-table `ON UPDATE CASCADE` FK, so its org-move restamp instead runs through the pre-existing `breeze_cascade_device_org_id()` `SECURITY DEFINER` trigger (`migrations/2026-05-18-device-child-orgid-cascade.sql`), which auto-discovers any table with `device_id`+`org_id` columns and restamps it under the function owner's privileges — bypassing `breeze_app`'s revoke entirely. `routes/devices/moveOrg.ts`'s app-level loop was taught to skip the now-redundant statement for it (`DEVICE_ORG_FK_CASCADE_TABLES` in `routes/devices/core.ts`).
- `pam_actuation_results` has PAM-history device moves blocked outright at the app layer (`devices_pam_history_move_guard`), so there is no restamp path to support at all.

## Suggested fix

Apply the same pattern already used for `agent_rollback_events`: since `peripheral_policy_delivery_events` also has both `device_id` and `org_id` columns, confirm it's covered by `breeze_cascade_device_org_id()`'s auto-discovery (it is — verified via `SELECT * FROM breeze_device_child_orgid_tables()` against a live test DB during #4371's investigation), add it to `DEVICE_ORG_FK_CASCADE_TABLES`, remove it from the app-level `getDeviceOrgDenormalizedTables()` explicit-UPDATE path, and revoke `UPDATE` from `breeze_app` for it in `ensureAppRole.ts` (matching `agent_rollback_events`'s full `UPDATE, DELETE, TRUNCATE` re-revoke) plus a companion migration REVOKE. That would close the last privilege gap among the six tables #4371 audited and let the trigger's `UPDATE` branch — the whole `IF (to_jsonb(NEW) - 'org_id') = ...` clause — be deleted as dead code (only the `DELETE` retention branch would remain).

## Scope note

Deliberately NOT done in #4371 itself: that PR's job was closing the specific privilege-re-revoke gap the issue reported plus fixing writer-path conflicts the fix exposed, not proactively tightening a table whose current behavior (real `UPDATE` grant + trigger-enforced restamp) is unchanged from `main` and was already working correctly (no observed exploit, no failing test). Revoking `UPDATE` here is a genuine behavior change to a live writer path (`moveOrg.ts`) that deserves its own PR, its own review, and its own regression test — not a rider on an already-large privilege-hardening PR.

## Writer paths for peripheral_policy_delivery_events (for whoever picks this up)

- Device org-move: `routes/devices/moveOrg.ts` → `getDeviceOrgDenormalizedTables()` loop (needs to move to `DEVICE_ORG_FK_CASCADE_TABLES`).
- Device permanent-delete: already correctly escalated to `breeze_audit_admin` (`DEVICE_CASCADE_AUDIT_ADMIN_TABLES`, `services/deviceDeletion.ts`) — no change needed there.
- Org merge: plain `repoint` (`orgMergeRegistry.ts`) — currently relies on the `UPDATE` grant this issue proposes revoking; would need to move to the same SECURITY-DEFINER-trigger-driven skip as `agent_rollback_events`'s org-merge path... actually org merge issues its own `UPDATE ... SET org_id` directly (not via `moveOrg.ts`), so **this is the second writer path that needs to be re-checked** before `UPDATE` can be safely revoked — worth auditing `orgMergeRegistry.ts`'s `repoint` executor the same way #4371 audited `agent_rollback_events`'s.
- Org erasure: already correctly escalated to `breeze_audit_admin` (`AUDIT_ADMIN_REQUIRED_TABLES`, `services/tenantCascade.ts`) — no change needed there.

Filed per #4371's PR review (coordinator-directed writer-path audit); see PR #4789 for the full six-table matrix and how the other five tables' writer paths were reconciled.

ToddHebebrand:
## Summary

`peripheral_policy_delivery_events` was the last of the six tables #4371 audited whose append-only guarantee still rested entirely on its `BEFORE UPDATE` trigger rather than a privilege wall — `breeze_app` kept a real `UPDATE` grant, unlike its siblings `pam_actuation_results`, `agent_rollback_events`, and `ml_feedback_events` (all fully revoked in #4789).

## Root cause

`UPDATE` was deliberately kept granted so `routes/devices/moveOrg.ts`'s device org-move path could restamp `org_id` on this table via an ordinary `breeze_app` `UPDATE`. That restamp was always redundant: the table has both `device_id` and `org_id` columns, so it's auto-discovered by `breeze_cascade_device_org_id()` (`migrations/2026-05-18-device-child-orgid-cascade.sql`), the `SECURITY DEFINER` trigger on `devices` that already restamps every such child table the instant `devices.org_id` changes — ahead of `moveOrg.ts`'s own app-level loop, exactly like `agent_rollback_events`.

## Changes

- **New migration** (`2026-10-08-100300-peripheral-policy-delivery-events-revoke-update.sql`): `REVOKE UPDATE ON TABLE peripheral_policy_delivery_events FROM breeze_app` — this table's original migration never revoked it, so (unlike #4789's boot-time re-grant fix) a real new privilege change was needed, not just a re-revoke.
- **`ensureAppRole.ts`**: per-boot re-revoke now covers `UPDATE, DELETE, TRUNCATE` (was `DELETE, TRUNCATE`), matching `agent_rollback_events`'s full re-revoke. Writer-path matrix comment updated.
- **`routes/devices/core.ts`**: added the table to `DEVICE_ORG_FK_CASCADE_TABLES` so `moveOrg.ts`'s generic loop skips the now-redundant, now-unprivileged app-level `UPDATE` (it stays in `CORE_DEVICE_ORG_DENORMALIZED_TABLES` too, same as `agent_rollback_events`).
- **`orgMergeRegistry.ts`**: reclassified the table from `'repoint'` to `'leave-for-erasure'`. I audited the org-merge writer path per the issue's request — org merge only ever writes as `breeze_app` (via `withSystemDbAccessContext`, which changes the RLS scope GUC, not the Postgres role/privileges); there is no `breeze_audit_admin` escalation anywhere in `orgMerge.ts`. Since `breeze_audit_admin` also has no `UPDATE` on this table (revoked in the table's own original migration), no role the merge can assume could ever issue the repoint `UPDATE` once `breeze_app`'s grant is gone — identical to `agent_rollback_events`'s existing classification.

## Tests

Extended the same contract/integration suites #4789 extended, now exercising this table alongside its already-covered siblings:
- `ensureAppRole.appendOnlyCoverage.test.ts` (static, unit) — writer-path matrix entry now asserts full revoke.
- `ensureAppRoleAppendOnlyPrivileges.integration.test.ts` — moved into the full-revoke `it.each` (was a dedicated "keeps UPDATE" test).
- `moveOrg.test.ts` / `moveOrg.coverage.test.ts` — `DEVICE_ORG_FK_CASCADE_TABLES` exact-equality assertions updated.
- `orgMergeRegistry.integration.test.ts` — moved from the `'repoint'` assertion list to the `'leave-for-erasure'` list. The pre-existing `every repoint / repoint-dedupe table grants breeze_app UPDATE` contract test is the real regression guard here — it runs against real Postgres and would have failed loudly had I revoked the privilege without reclassifying.

All ran green against a private, freshly-migrated test-stack (not the shared `:5433` stack): the new migration applied cleanly, `ensureAppRoleAppendOnlyPrivileges` (6 tests), `orgMergeRegistry.integration.test.ts` (14 tests), `deviceMoveOrgCurrency`/`deviceMoveOrgIntentScopeTombstone`/`peripheralPolicyV2Rls` integration tests (13 tests — the real `POST /devices/:id/move-org` route exercised end to end, which is what actually proves the loop's skip-and-cascade-trigger path holds under a real `breeze_app` connection), and the org-merge gauntlet (`orgMerge.integration.test.ts` + `orgMergeCustomExecutors.integration.test.ts`, 18 tests). `tsc --noEmit` and `eslint` on touched files are clean.

I deliberately left the append-only trigger's now-unreachable `UPDATE` branch (`peripheral_policy_delivery_events_append_only()`'s `IF (to_jsonb(NEW) - 'org_id') = ...` check) in place rather than deleting it, matching how `agent_rollback_events`'s equivalent dead branch was handled in #4789 — removing it would be a separate, deliberate follow-up.

## Coordination note

`orgMergeRegistry.ts` is also touched by open PR #4823 (different hunk — ticket org-move lock order, not this table's classification), so this branch may need a rebase after #4823 merges.

Closes #4806


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer, silent-failure-hunter — sonnet)

**Findings:**
- code-reviewer: 0 findings. Verified the privilege chain, the `breeze_cascade_device_org_id()` auto-discovery claim against the actual schema/trigger, all other writer paths (`peripheralPolicyState.ts`, `deviceDeletion.ts`, `tenantCascade.ts`, `tenantExportPolicyRegistry.ts`), migration naming/idempotency, and comment consistency across the ~13 `#4806` cross-references. No changes needed.
- silent-failure-hunter: 0 findings. Confirmed the revoked UPDATE is never issued for this table (loop `continue`s via `DEVICE_ORG_FK_CASCADE_TABLES`, so no swallowed 42501 is possible), confirmed `leave-for-erasure` rows are already surfaced in the org-merge preview's generic "audit trail permanently destroyed" note (no new silent data-fate gap), and confirmed no other UPDATE call site exists against this table anywhere in `apps/api/src`.
- pr-test-analyzer: 2 findings, both addressed in `ae2f97a44`.
  - **Addressed (severity 9/10):** no test proved the cascade-trigger-only restamp actually works for this table under a real `breeze_app` connection — the existing "restamps mutable state..." test issued its own explicit UPDATE via the schema-owner connection, so it would have passed even if the trigger never covered this table. Added a new test in `peripheralPolicyV2Rls.integration.test.ts` that moves a device's org via a real `breeze_app`-privileged `UPDATE devices` statement (matching what `moveOrg.ts` issues) with **zero** manual statement against `peripheral_policy_delivery_events`, then asserts the row was restamped. Verified this test is genuinely red-first: reverting `ensureAppRole.ts`'s re-revoke made it fail with `expected true to be false` on the privilege sanity check; restoring the fix makes it pass.
  - **Not addressed, flagged as pre-existing (severity 7/10):** no behavioral test exercises an org-merge where the loser org owns rows in this table (only the static classification-map assertion is tested). This gap already existed for the sibling `agent_rollback_events` after #4789 and is not introduced or worsened by this PR — scoping a fix for both tables together would be a reasonable follow-up but is out of scope here.

**Tests:** All affected suites green against a private, freshly-migrated test-stack (not the shared `:5433` stack) after rebasing onto latest `main`: `ensureAppRoleAppendOnlyPrivileges.integration.test.ts`, `orgMergeRegistry.integration.test.ts`, `peripheralPolicyV2Rls.integration.test.ts` (8 tests incl. the new one), `deviceMoveOrgCurrency` + `deviceMoveOrgIntentScopeTombstone.integration.test.ts`, `orgMerge.integration.test.ts` + `orgMergeCustomExecutors.integration.test.ts` — 52 tests total, all passing. Plus 6 unit test files (172 tests). `tsc --noEmit` and `eslint` on touched files clean.

**Migration naming:** renamed from `2026-10-08-100300-...` to `2026-10-08-100800-...` per coordinator note (that range is contested by #4818/#4821/#4825); confirmed `check-migration-naming.sh` passes and it sorts after the newest migration on `main`.

**Status:** review-clean, awaiting maintainer merge. Branch is rebased onto current `main` (`ccbc7f33f`); note `orgMergeRegistry.ts` is also touched by open PR #4823 (different hunk — ticket lock order, not this table's classification), so a further rebase may be needed if that merges first.

PR #4827, opened 2026-09-03
fix(api): persist an error mapping row on currency_mismatch invoice push (#4498)

### LINKED ISSUE #4498: [API/UI] QuickBooks auto-push currency_mismatch leaves no trace on the invoice
## Description
When the `accounting-sync` worker auto-pushes an issued invoice whose currency differs from the connected QuickBooks home currency, the coordinator throws `currency_mismatch` *before* any `accounting_entity_mappings` row exists (by design of the Phase C currency contract in `accountingCurrency.ts`). The worker logs a terminal failure and does not retry. Result: `GET /invoices/:id` returns `accountingSync: null`, so the invoice detail `AccountingSyncCard` renders nothing. A tech has no signal that the push failed unless they push manually, which returns the 409 with the full message.

Verified 2026-09-01 in the live sandbox run (PR #4492, `docs/integrations/quickbooks-sandbox-verification.md` step 15).

## Root Cause
`pushInvoiceToAccounting` runs the currency guard in Phase 1 and rethrows without creating the pending mapping row, so there is nothing for `getInvoiceAccountingSync` to read. The contract deliberately forbids persisting "sync state" for a currency failure, but that leaves the UI blind.

## Proposed Fix
Options, in order of preference:
1. Persist a mapping row in `sync_status='error'` with the sanitized `currency_mismatch` message but **no** remote id (keeps "no QBO call, no remote state" while giving the card something to show). Needs the `accountingCurrency.ts` contract comment and `accountingInvoicePushCurrency.integration.test.ts` updated to allow an error-only row.
2. Or surface worker terminal failures through a different channel (invoice audit log entry + card fallback that reads the last `accounting.invoice.push` audit row).

## Affected Files
- `apps/api/src/services/accounting/accountingInvoicePush.ts`
- `apps/api/src/services/accounting/accountingCurrency.ts` (contract comment)
- `apps/api/src/__tests__/integration/accountingInvoicePushCurrency.integration.test.ts`
- `apps/web/src/components/billing/AccountingSyncCard.tsx`

## Reported By
Internal, Phase C sandbox walkthrough (PR #4492).

ToddHebebrand:
## Summary

When the accounting-sync worker auto-pushes an issued invoice whose currency differs from the connected QuickBooks home currency, `pushInvoiceToAccounting`'s currency guard throws `currency_mismatch` **before** any `accounting_entity_mappings` row exists — by design of the Phase C currency contract (`accountingCurrency.ts`). The worker logs a terminal failure and doesn't retry. Result: `GET /invoices/:id` returns `accountingSync: null`, so `AccountingSyncCard` renders nothing — a tech has no signal that the push failed short of retrying manually and reading the 409.

This inverts that one piece of the contract for `currency_mismatch` specifically: the coordinator now persists an **error-only mapping row** (`sync_status: 'error'`, no `remoteEntityId`, both currencies named in `lastError`) so the existing `getInvoiceAccountingSync` read path and `AccountingSyncCard`'s `syncStatus:'error'`-is-pushable branch surface it — **no new UI plumbing**. `home_currency_unknown` is unchanged (a rarer connection-setup problem, out of this issue's scope) and still persists nothing.

## The tricky part

`pushInvoiceToAccounting`'s Phase 1 currency guard throws from *inside* one `runInDbContext` transaction. Per the file's own DB ACCESS CONTRACT doc comment, that transaction rolls back the instant the callback throws — so writing the error row inside that same callback would have vanished with the throw (the exact "lost sync state" class the file already warns about for the provider-call failure path). The fix instead:

- Wraps the whole Phase 1 `runInDbContext` call in a `try`/`catch` in `pushInvoiceToAccounting`.
- On a `currency_mismatch` catch, calls a new `persistInvoiceCurrencyMismatchErrorInOwnContext` — its own short, separately-committed context, opened *after* Phase 1 has already rolled back — which re-resolves the connection and upserts the mapping row (insert if none exists, update-in-place if one does, never touching `remoteEntityId`/`linkStatus` on an existing row, never clobbering a `remote-deleted` marker).
- The write is best-effort and never replaces the typed `currency_mismatch` error being (re)thrown.

## Verification

- **Real Postgres integration test** (updated `accountingInvoicePushCurrency.integration.test.ts`): EUR-vs-USD push now asserts a persisted error-only row naming both currencies; a second consecutive push proves idempotent UPDATE-in-place (no duplicate row, no duplicate QuickBooks call); `voidInvoiceInAccounting` against that row resolves as a no-op (proves it doesn't trip the in-flight `pending` marker check). `home_currency_unknown` case unchanged (still asserts no row).
- **New mocked unit tests** in `accountingInvoicePush.test.ts`: insert shape, scoped-to-`currency_mismatch`-only (not `home_currency_unknown`), update-in-place preserving an already-synced row's `remoteEntityId`/`linkStatus`, SQL-level remote-deleted-guard regression test, and a best-effort/Sentry-capture test for a failed persist.
- `pnpm exec vitest run` on the touched files: 44 + 25 (worker/call-sites) + 176 (invoiceService/accountingMappingService) unit tests green; 3 integration tests green against real Postgres.
- `npx tsc --noEmit` clean; `npx eslint` clean on all four touched files.
- `AccountingSyncCard.test.tsx` (16 tests) unchanged and still green — confirms no UI changes were needed.

## Files

- `apps/api/src/services/accounting/accountingInvoicePush.ts`
- `apps/api/src/services/accounting/accountingInvoicePush.test.ts`
- `apps/api/src/services/accounting/accountingCurrency.ts` (contract comment)
- `apps/api/src/__tests__/integration/accountingInvoicePushCurrency.integration.test.ts`

Closes #4498

### DIFF
--- a/apps/api/src/__tests__/integration/accountingInvoicePushCurrency.integration.test.ts
+++ b/apps/api/src/__tests__/integration/accountingInvoicePushCurrency.integration.test.ts
@@ -1,21 +1,21 @@
 /**
  * Real-DB proof for the Phase-C invoice-push currency contract (multi-currency
- * §11, bug B8). This is the exact file the trailing contract comment in
- * `accountingCurrency.ts:143-186` mandates by name — that comment is the
- * authoritative spec for what this suite proves, and this file follows it
- * verbatim rather than a looser paraphrase:
+ * §11, bug B8), UPDATED by #4498 to invert the "no trace" half of that
+ * contract for `currency_mismatch` specifically. The trailing contract
+ * comment in `accountingCurrency.ts:143-186` still describes the ORIGINAL
+ * shape (no mapping row for either currency failure) — that comment is
+ * stale for `currency_mismatch` as of this change; `home_currency_unknown`
+ * is unaffected and still persists nothing (see #4498's own scoping: it is a
+ * rarer connection-setup problem, not this issue's complaint):
  *
  *  - EUR invoice + USD-home connection: `pushInvoiceToAccounting` throws
  *    `ACCOUNTING_INVOICE_CURRENCY_MISMATCH` (typed here as the coordinator's
- *    `currency_mismatch`), NO QBO request is made, and NO sync/mapping state
- *    is persisted — no `accounting_entity_mappings` row for the invoice
- *    exists at all, because the currency guard
- *    (`assertAccountingInvoicePushCurrency`) runs before the coordinator ever
- *    reads/writes the invoice's mapping row (see `accountingInvoicePush.ts`'s
- *    `pushInvoiceToAccounting`: the guard sits immediately after the
- *    draft/void check and strictly before the org mapping load, the line
- *    load, and `upsertInvoiceMappingPending`).
- *  - NULL home currency: same treatment, `home_currency_unknown`.
+ *    `currency_mismatch`), NO QBO request is made, and the invoice's
+ *    `accounting_entity_mappings` row now lands `sync_status:'error'` with
+ *    no `remoteEntityId` and both currencies named in `lastError` — so the
+ *    invoice detail card has something to show instead of nothing (#4498).
+ *  - NULL home currency: `home_currency_unknown`, NO QBO request, and NO
+ *    mapping row persisted — unchanged from before #4498.
  *  - USD invoice against a USD-home connection with the provider transport
  *    mocked at the `fetch` boundary: the invoice's mapping row lands
  *    `sync_status:'synced'`, and a second push against the now-existing
@@ -44,7 +44,7 @@ import {
 } from '../../db/schema';
 import { createOrganization, createPartner } from './db-utils';
 import { upsertConnection } from '../../services/accounting/accountingConnectionService';
-import { pushInvoiceToAccounting } from '../../services/accounting/accountingInvoicePush';
+import { pushInvoiceToAccounting, voidInvoiceInAccounting } from '../../services/accounting/accountingInvoicePush';
 
 const runDb = it.runIf(!!process.env.DATABASE_URL);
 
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** \`/pr-review-toolkit:review-pr\` — three parallel Sonnet reviewers: \`code-reviewer\`, \`silent-failure-hunter\`, \`pr-test-analyzer\`.

**Findings:** 3 raised → 1 addressed, 1 filed as a follow-up (pre-existing/out of scope), 1 accepted as-is (matches established codebase precedent).

- **code-reviewer**: 0 findings. Confirmed the transaction-rollback reasoning (Phase 1 rolls back on throw; the new persist runs from the outer catch in its own committed context), the `currency_mismatch`-only scoping, and the never-clobber guarantees on an existing/remote-deleted row.
- **silent-failure-hunter**: 1 finding — the Sentry `captureException` calls in this file (including the new one) use tag keys (`service`, `invoiceId`, `partnerId`) not present in `sentry.ts`'s `ALLOWED_TAG_NAMES`, so they're silently stripped before reaching Sentry. Confirmed **pre-existing** — the identical pattern already exists at 6 other call sites in this same file, so this PR adds a 7th instance rather than introducing the bug. Fixing it properly means updating the shared, deliberately-curated allowlist (which the file's own convention requires justifying per entry) plus normalizing all 7 sites — real scope beyond this issue. Filed **#4828** to track it as its own PR rather than a partial, inconsistent fix here.
- **pr-test-analyzer**: 2 findings.
  - Untested unique-violation race branch in the new insert path → **fixed in \`9d405fffc\`**: added a test proving the benign race is swallowed silently, plus a sibling test proving a *different* insert failure (wrong constraint) still reports to Sentry, so the \`isPgUniqueViolation\` check can't silently widen.
  - The remote-deleted WHERE-guard test only proves the guard clause is present in the compiled SQL, not that it behaviorally blocks a real concurrent Postgres write. Reviewed against precedent: the file's *existing* equivalent guard test (for \`upsertInvoiceMappingPending\`, same file) uses the identical structural-only technique and its own comment explicitly says full behavioral proof needs a live DB and is out of scope for the mocked suite. My new test matches that established, accepted risk tolerance for this class of guard — not escalating further.

**Tests:** \`accountingInvoicePush.test.ts\` 46/46 passed; \`accountingSyncWorker.test.ts\` + \`accountingInvoicePushCallSites.test.ts\` 25/25 passed; \`invoiceService.test.ts\` + \`accountingMappingService.test.ts\` 176/176 passed (regression check); real-Postgres \`accountingInvoicePushCurrency.integration.test.ts\` 3/3 passed (currency_mismatch persists+idempotent-retry+void-no-op, home_currency_unknown unchanged, happy-path unchanged). \`AccountingSyncCard.test.tsx\` 16/16 passed unchanged, confirming no UI changes were needed. \`npx tsc --noEmit\` clean; \`npx eslint\` clean on all touched files.

**Status:** review-clean, awaiting maintainer merge. Follow-up filed: #4828 (Sentry tag allowlist gap, pre-existing, out of scope for #4498).

PR #4829, opened 2026-09-03
fix(api): widen release-time actor context for approved move_org intents (#4650)

### LINKED ISSUE #4650: [API][AI] the AI move_org ticket tool is structurally unreachable — release-time actor context is pinned to one org
Surfaced by the PR #4642 review (two independent reviewers found it while tracing #4524's callers). **Pre-existing**, not introduced by that PR — filing so it does not stay buried in a review thread.

**Claim: verified in code**, not inferred.

`manage_tickets` / `move_org` is Tier 3 **and** four-eyes (`apps/api/src/services/aiGuardrails.ts:254` and `:324`), so it never executes inline — it always goes through an `action_intents` proposal and is executed at release time.

At release, the acting `AuthContext` is rebuilt from the intent, and **both** reconstruction paths pin a single organization:

- user-owned intents — `buildUserOwnedAuthContext` (`apps/api/src/services/actionIntents/actorContext.ts:219-221`): `scope: 'organization'`, `accessibleOrgIds: [intent.orgId]`
- agent-owned intents — `buildAgentAuthContext` (`apps/api/src/services/aiAgents/agentAuthContext.ts:91`): `accessibleOrgIds: [run.orgId]`, with `canAccessOrg: (id) => id === run.orgId`

The tool handler then gates on the target org (`apps/api/src/services/aiToolsTicketing.ts:746-748`):

```ts
if (!auth.canAccessOrg(String(input.targetOrgId))) {
  return JSON.stringify({ error: 'Access to target organization denied' });
}
```

`intent.orgId` / `run.orgId` is the **ticket's own (source) org** at proposal time, so for any `targetOrgId !== sourceOrgId` that check is always false. And a same-org move is an early-return no-op inside `moveTicketOrg`. **So the tool has no reachable success path**: every approved `move_org` returns `Access to target organization denied`.

**Impact:** AI-driven ticket org-moves are non-functional end to end. The failure is loud but badly worded — an operator who approves the four-eyes request sees an access-denied string that reads like a permissions misconfiguration on *their* account, not a wiring bug, so it is likely to be mis-triaged.

By contrast the HTTP route (`apps/api/src/routes/tickets/moveOrg.ts`) requires `partner`/`system` scope, whose `accessibleOrgIds` spans both orgs, and works fine.

**Fix direction (needs a decision):** the released context has to carry access to the move's *target* org, which is known at proposal time. Options include recording the target org on the intent and widening `accessibleOrgIds` to `[intent.orgId, targetOrgId]` when rebuilding, or requiring a partner-scoped reconstruction for tenant-shape mutations. Whichever is chosen must keep the release-time re-authorization honest — the four-eyes approval is what authorizes the *specific* move, so the widening must be bounded to the approved target and not become a blanket partner-scope escalation. Worth pairing with a test that actually drives an approved `move_org` intent to completion; there is currently none, which is why this has gone unnoticed.

Related: #4524, PR #4642, #4191, #3175.

ToddHebebrand:
## Summary

The AI `move_org` ticket tool was structurally unreachable: **both** release-time actor-context rebuild paths in `apps/api/src/services/actionIntents/actorContext.ts` pinned `accessibleOrgIds` to the intent's single (source) org, so `aiToolsTicketing.ts`'s `auth.canAccessOrg(targetOrgId)` gate always 403'd for every approved four-eyes `move_org` intent, regardless of the requester's real access. Filed and diagnosed in #4650 (surfaced during the #4642 review).

## What changed

- Added a small allowlist, `TENANT_MUTATION_TARGET_ORG_ARG` (keyed `toolName -> action -> arg key`, currently just `manage_tickets: { move_org: 'targetOrgId' }`), and `resolveTenantMutationTargetOrgId(intent)` which reads the target org **only from the intent's immutable `arguments` snapshot** captured at creation time — never a live re-lookup of the ticket's current org, which would reopen the exact TOCTOU window the immutable snapshot exists to close.
- **`buildUserOwnedAuthContext`**: widens `accessibleOrgIds` to `[intent.orgId, targetOrgId]` only when the requester's already-resolved `perms` (re-verified at release time) can also reach `targetOrgId` — reusing `permsCanAccessOrg`, not a new check. A requester who can't reach the target keeps the single-org `accessibleOrgIds`, so the tool's existing `canAccessOrg` gate still 403s — now for the *correct* reason.
- **`buildAgentOwnedAuthContext`**: same shape, bounded to the *acting agent's own* scope (this context executes AS the agent, not the human approver). An org-scoped agent is never eligible (its home is one org); a partner-scoped agent widens only when the target org's current partner ownership matches the agent's own partner.
- Neither `buildAgentAuthContext` (shared with the live inline `runLoop.ts` execution path) nor the tool's own `aiToolsTicketing.ts` gate needed changes — the widening happens purely in the release-time reconstruction, bounded and allowlisted.

## Tests

- `apps/api/src/services/actionIntents/actorContext.test.ts` — new unit tests (mocked DB) for both reconstruction paths: widens when access covers the target; does not widen when it doesn't (tool gate still refuses); never widens an org-scoped requester/agent; never widens a non-allowlisted tool/action even with a same-shaped `targetOrgId` argument; defensive no-target-recorded case.
- `apps/api/src/__tests__/integration/moveOrgIntentReleaseActorContext.integration.test.ts` — new real-Postgres end-to-end suite (mirrors `intentSupervisedFourEyes.integration.test.ts`'s pattern: `createActionIntent` → approve via `approvalRoutes` → `releaseApprovedIntent`, the exact function the durable worker calls):
  - (a) an approved `move_org` intent from a partner-scope requester with `orgAccess: 'all'` actually releases cross-org — the ticket's `org_id` really changes in Postgres and the intent completes.
  - (b) the widening is allowlist-scoped, not generic: a manually-inserted `manage_tickets:assign` intent carrying a same-named `targetOrgId` argument is proven, via the real `buildAuthContextForIntent`, to never get widened.
  - (c) a requester with `orgAccess: 'selected'` covering only the source org is still refused — via the tool's `tool_returned_error` path, not `actor_invalid` — and the ticket never moves.

## Verification

- `npx tsc --noEmit` — clean.
- `pnpm eslint` on touched files — clean.
- Unit: `actorContext.test.ts` (40 tests), plus `aiGuardrails.test.ts`, `aiToolsTicketing.writeGaps.test.ts`, `ticketService.test.ts`, `intentService.test.ts`, `revalidateRelease.test.ts`, `agentAuthContext.test.ts` (487 tests total) — all green, single-fork.
- Integration (real Postgres, isolated worktree stack): the new `moveOrgIntentReleaseActorContext.integration.test.ts` (3 tests) plus the sibling `intentSupervisedFourEyes`, `agentIntentLifecycle`, `aiAgentTicketTriage` suites (20 tests) — all green.

Closes #4650

### DIFF
--- a/apps/api/src/__tests__/integration/moveOrgIntentReleaseActorContext.integration.test.ts
+++ b/apps/api/src/__tests__/integration/moveOrgIntentReleaseActorContext.integration.test.ts
@@ -35,10 +35,17 @@
  *       never `actor_invalid` and never a silent success.
  */
 import './setup';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { randomUUID } from 'crypto';
 import { eq } from 'drizzle-orm';
 import { Hono } from 'hono';
+
+// Required for the agent-owned scenario below: checkAgentGuardrails
+// (aiGuardrails.ts) denies every agent proposal outright when this env flag
+// is off, same precedent as aiAgentTicketTriage.integration.test.ts.
+vi.hoisted(() => {
+  process.env.BREEZE_AI_AGENTS_ENABLED = 'true';
+});
 
 import { db, withSystemDbAccessContext } from '../../db';
 import { getTestDb } from './setup';
@@ -46,8 +53,10 @@ import { actionIntents } from '../../db/schema/actionIntents';
 import { approvalRequests } from '../../db/schema/approvals';
 import { partnerUsers } from '../../db/schema/users';
 import { tickets } from '../../db/schema/portal';
+import { aiAgents, aiAgentRuns } from '../../db/schema/aiAgents';
 import { createActionIntent } from '../../services/actionIntents/intentService';
 import { buildAuthContextForIntent } from '../../services/actionIntents/actorContext';
+import { buildAgentAuthContext } from '../../services/aiAgents/agentAuthContext';
 import type { ActionIntent } from '../../db/schema/actionIntents';
 import { PERMISSIONS } from '../../services/permissions';
 import { buildOrgAccessClosures, type AuthContext } from '../../middleware/auth';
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** 4 parallel Sonnet reviewers — `pr-review-toolkit:code-reviewer`, `pr-review-toolkit:silent-failure-hunter`, `pr-review-toolkit:pr-test-analyzer`, and a `security-review`-skill pass (multi-tenant isolation / auth-boundary focus, since this is a release-time authorization boundary).

**Findings:** 4 substantive findings raised across the four reviews; 2 acted on, 2 accepted as out-of-scope follow-ups (documented, not dropped).

- **[code-reviewer, Important] Addressed in `a17bb7c03`.** `buildUserOwnedAuthContext`'s widening relied solely on `permsCanAccessOrg`, which for a partner requester with `orgAccess:'all'` returns true for *any* org id without checking its partner — asymmetric with the agent-owned path, which already did a live partner check. Not currently exploitable (`moveTicketOrg` independently rejects a cross-partner move) but not a property of the widening itself. Fixed by adding the same live target-org-partner check to the user-owned path, plus strengthened the allowlist's doc comment so a future entry can't assume the bound alone is sufficient. New unit + integration coverage (scenario **(d)**) proves it.
- **[pr-test-analyzer, gap rated 7/10] Addressed in `a17bb7c03`.** The agent-owned widening had only mocked unit coverage, no real-Postgres proof — exactly the "looks correct but never exercised at release time" bug class #4650 itself was. Added integration scenarios **(e)/(f)**: a real `ai_agents`/`ai_agent_runs` fixture, driven through the actual `createActionIntent` + four-eyes-approval pipeline, then rebuilding the release-time `AuthContext` via the real `buildAuthContextForIntent` (no mocks) to prove the widening fires (e) and stays bounded cross-partner (f).
- **[discovered while building (e)] Filed as #4830, not fixed here.** Driving a real agent-owned `move_org` intent all the way through `releaseApprovedIntent` (full tool execution) surfaced a *separate*, pre-existing bug unrelated to `accessibleOrgIds`: `moveTicketOrg`'s "moved" system-comment insert writes the agent's synthetic id into `ticket_comments.user_id`, violating that table's RLS/FK. Every other `manage_tickets` action routes `ai_agent` principals through a dedicated AI-safe executor; `move_org` does not. This was unreachable before #4650 (accessibleOrgIds always 403'd first) and needs its own fix/design decision — out of scope for this PR. The new integration test stops at the rebuilt `AuthContext` rather than full execution, with a comment explaining why.
- **[security-review, informational; pr-test-analyzer, minor] Accepted, not blocking.** Same-org `targetOrgId === intent.orgId` and malformed/non-string `targetOrgId` are covered only by the existing fail-closed *design* (both are handled, just not independently pinned by a dedicated test). Left as-is per the review's own "informational" rating.

**Tests:** `apps/api/src/services/actionIntents/actorContext.test.ts` — 42 tests, single-fork, green. `apps/api/src/__tests__/integration/moveOrgIntentReleaseActorContext.integration.test.ts` — 6 tests against a real, freshly-migrated isolated Postgres stack, green. Sibling suites (`intentSupervisedFourEyes`, `agentIntentLifecycle`, `aiAgentTicketTriage` — 20 tests) re-verified green for regressions. `npx tsc --noEmit` clean. `eslint` clean on all touched files.

**Status:** review-clean (2 findings fixed, 1 filed as #4830, 2 accepted as informational), awaiting maintainer merge.

PR #4831, opened 2026-09-03
fix(api/tenancy): tombstone action_intents.scope_ticket_id on device org-move (#4792)

### LINKED ISSUE #4792: AI agents: a device org-move can 23503 on action_intents_scope_ticket_org_fk and abort the move
Found while fixing #4454 (PR #4790), which mirrored the `scope_device_id` tombstone into `breeze_cascade_device_org_id()`. The **ticket** axis of the same typed target scope has a worse version of the same gap, and it is missing from **both** sites — the route and the trigger — not just the trigger.

## The mechanism

`tickets` is returned by `breeze_device_child_orgid_tables()`, so a device org-move re-stamps `tickets.org_id` for every ticket bound to the moved device (both via `breeze_cascade_device_org_id()`'s generic loop and via the equivalent loop in `apps/api/src/routes/devices/moveOrg.ts`).

`action_intents.org_id` is immutable — `action_intents_block_content_update()` raises `action_intents content is immutable` on any change to it.

`action_intents_scope_ticket_org_fk` (`apps/api/migrations/2026-09-25-ai-agents-ticket-triage.sql:216-219`) is:

```sql
FOREIGN KEY (scope_ticket_id, org_id) REFERENCES tickets (id, org_id) ON DELETE SET NULL
  DEFERRABLE INITIALLY IMMEDIATE
```

No `ON UPDATE` clause, so it defaults to `NO ACTION`. Re-stamping the ticket's `org_id` therefore leaves any intent scoped to that ticket holding a pair `(scope_ticket_id, old_org_id)` that no longer exists in `tickets` — a `23503` raised at the end of the re-stamp statement, which aborts the whole device move.

Unlike #4454 this affects intents of **every** status, not just live ones: the FK does not care about `status`.

## Why the fix is already known

`ticketService.moveTicketOrg` handles exactly this for the ticket-level move — see `apps/api/src/services/ticketService.ts:2280-2298`, whose comment spells out that ANY remaining `scope_ticket_id` pointer regardless of status raises 23503, and tombstones all statuses **before** changing the ticket. The device-move path needs the same detach in both sites, placed before the generic re-stamp loop (a cleanup after it never runs — the loop's own statement is what trips the constraint).

Org merge is affected too, differently: it runs under `SET CONSTRAINTS ALL DEFERRED`, which postpones the same mismatch to COMMIT, but `orgMergeRegistry` classifies `action_intents` as `leave-for-erasure` and does not delete the intent during Phase B.

## Suggested scope

1. Add the `scope_ticket_id` detach (all statuses) to `apps/api/src/routes/devices/moveOrg.ts` and to a new `CREATE OR REPLACE` of `breeze_cascade_device_org_id()`, both before the generic loop.
2. Extend the both-sites contract in `apps/api/src/routes/devices/moveOrg.coverage.test.ts` — it already asserts the `scope_device_id` tombstone at both sites after PR #4790.
3. Real-Postgres regression proving a device org-move with a ticket-scoped intent no longer raises. `apps/api/src/__tests__/integration/deviceMoveOrgIntentScopeTombstone.integration.test.ts` is the natural home.

## Also worth a look while in there

That FK's `ON DELETE SET NULL` is unqualified, so on a ticket delete it appears to null **both** referencing columns of the pair — including `action_intents.org_id`, which is `NOT NULL` (`apps/api/migrations/2026-07-18-action-intents.sql:37`). Compare with `tickets_requester_contact_org_fk`, which uses the explicit column-list form `ON DELETE SET NULL (requester_contact_id)`. Worth a regression check on the ticket-delete path.

Not verified against a live server — the mechanism above is read off the migrations and the FK definition; the 23503 has not been reproduced. Reproducing it is step 0.

ToddHebebrand:
## What

A device org-move can `23503` on `action_intents_scope_ticket_org_fk` and abort the whole move when the moving device has a ticket that some intent is scoped to.

`tickets` IS returned by `breeze_device_child_orgid_tables()`, so a device org-move re-stamps `tickets.org_id` for every ticket bound to the moved device — both via `apps/api/src/routes/devices/moveOrg.ts`'s generic denormalized-table loop and via the equivalent loop in `breeze_cascade_device_org_id()`. `action_intents_scope_ticket_org_fk` (`apps/api/migrations/2026-09-25-ai-agents-ticket-triage.sql:216-219`) is:

```sql
FOREIGN KEY (scope_ticket_id, org_id) REFERENCES tickets (id, org_id) ON DELETE SET NULL
  DEFERRABLE INITIALLY IMMEDIATE
```

No `ON UPDATE` clause, so it defaults to `NO ACTION` and is checked at the end of the tickets re-stamp statement (this FK is deliberately NOT named in the transaction's `SET CONSTRAINTS ... DEFERRED` — see the "BY NAME, never ALL" comment in `moveOrg.ts`). Any remaining `scope_ticket_id` pointer at that instant names a `(ticketId, OLD org_id)` pair that no longer exists in `tickets`, and the whole move transaction rolls back with a `23503`.

Unlike the sibling `scope_device_id` gap (#4454, fixed in #4790), this FK does not gate on `status` — every intent status, not just live ones, must be tombstoned, or a terminal-status intent still 23503s the very next org-move that touches its ticket.

## How

Mirrors `ticketService.ts`'s `moveTicketOrg` — which already handles exactly this for the ticket-level move via an unconditional (all-statuses) detach run before the ticket `org_id` UPDATE — into both sites the device-move axis needs:

- `apps/api/src/routes/devices/moveOrg.ts`: adds the `scope_ticket_id` tombstone immediately after the `scope_device_id` tombstone and the tickets requester-contact detach, and BEFORE the generic `getDeviceOrgDenormalizedTables()` re-stamp loop — the loop is what re-stamps `tickets.org_id` and trips the constraint if the tombstone hasn't run first.
- New migration `apps/api/migrations/2026-10-08-100900-cascade-device-org-move-ticket-scope.sql`: `CREATE OR REPLACE FUNCTION breeze_cascade_device_org_id()`, function body copied verbatim from the newest prior definition (`2026-10-08-100500-detach-ticket-comment-runs-on-device-org-move.sql`, merged minutes before this branch was cut) with the one statement added, placed before the generic re-stamp loop for the same reason. No inner `BEGIN`/`COMMIT`; the trigger itself is unchanged and not redeclared.

No historical backfill: unlike the silently-stale `scope_device_id` gap, this one is a hard abort — a device org-move carrying a ticket-scoped intent could not have succeeded before this fix, on either site, so there is nothing pre-existing to clean up.

## Tests

**Red first, both halves**, verified locally against an isolated `pnpm test-stack` instance:

- With the migration's added statement and the route's added statement both temporarily removed: the real-Postgres integration test's route case failed with the exact reported `23503` on `action_intents_scope_ticket_org_fk` via a raw `UPDATE devices` (500 from the route, `PostgresError: update or delete on table "tickets" violates foreign key constraint...`), and both trigger-path cases failed the same way. With both statements restored, all 3 pass.
- **`moveOrg.coverage.test.ts`** (blocking `test-api` job): new `describe` block asserts BOTH sites — the route source and the newest migration's function body — including that the statement is unconditional (no status filter, unlike `scope_device_id`'s) and that it runs before the generic re-stamp loop (load-bearing here, since `tickets` IS in that loop, unlike `scope_device_id`'s equivalent placement guard which is a convention-only check).
- **`deviceMoveOrgTicketScopeTombstone.integration.test.ts`** (new, blocking `integration-test` job): drives the fix three ways — the actual `POST /devices/:id/move-org` route, a raw `UPDATE devices` under the unprivileged `breeze_app` role via `withSystemDbAccessContext` (the shape orgMerge's `devices` repoint has), and a raw superuser `UPDATE devices`. Every case seeds one live (`pending_approval`) and one `completed` intent scoped to a ticket bound to the moving device and asserts the move succeeds (200 / no thrown error) and BOTH intents end up tombstoned — this FK's lack of a status gate means the terminal-status intent must ALSO be nulled, not left alone the way `scope_device_id`'s live-only tombstone leaves it.

Verified:

| Check | Result |
|---|---|
| `deviceMoveOrgTicketScopeTombstone.integration.test.ts` | 3 passed (all 3 red before restoring the fix) |
| `moveOrg.coverage.test.ts` | 29 passed |
| `moveOrg.test.ts` | 33 passed |
| `deviceMoveOrgIntentScopeTombstone`, `deviceMoveOrgCurrency`, `agentRunMoveSemantics`, `ticket-move-org`, `actionIntentsImmutabilityTrigger` (integration) | 61 passed |
| `autoMigrate.test.ts` | 67 passed |
| `pnpm db:check-drift` | no drift |
| `npx tsc --noEmit` (apps/api) | clean |
| `npx eslint` on touched files | clean |

No schema change (no new table/column), so no cascade / export-policy / RLS-allowlist registration applies.

Closes #4792

### DIFF
--- a/apps/api/migrations/2026-10-08-100900-cascade-device-org-move-ticket-scope.sql
+++ b/apps/api/migrations/2026-10-08-100900-cascade-device-org-move-ticket-scope.sql
@@ -1,0 +1,141 @@
+-- #4792 — mirror moveOrg's action_intents.scope_ticket_id tombstone into
+-- breeze_cascade_device_org_id().
+--
+-- Found while fixing #4454/#4790, which mirrored the scope_device_id
+-- tombstone into this same function. The TICKET axis of the same typed
+-- target scope has a worse version of the same gap, and it was missing from
+-- BOTH sites (the route and this trigger) until now — not just the trigger.
+--
+-- `tickets` IS returned by breeze_device_child_orgid_tables(), so a device
+-- org-move re-stamps tickets.org_id for every ticket bound to the moved
+-- device via the generic loop below. `action_intents.org_id` is immutable
+-- (action_intents_block_content_update() raises on any change to it).
+-- `action_intents_scope_ticket_org_fk`
+-- (migrations/2026-09-25-ai-agents-ticket-triage.sql:216-219) is:
+--
+--   FOREIGN KEY (scope_ticket_id, org_id) REFERENCES tickets (id, org_id)
+--     ON DELETE SET NULL DEFERRABLE INITIALLY IMMEDIATE
+--
+-- No ON UPDATE clause, so it defaults to NO ACTION and is checked at the end
+-- of the statement that fires it (the generic loop's tickets UPDATE — the
+-- FK is not named in this transaction's SET CONSTRAINTS, and this function
+-- runs under whatever constraint mode the calling statement's transaction
+-- is in). Re-stamping tickets.org_id therefore leaves any intent scoped to
+-- that ticket holding a pair (scope_ticket_id, OLD org_id) that no longer
+-- exists in `tickets` — a 23503 that aborts the whole device move.
+--
+-- Unlike the scope_device_id tombstone (#4454) this reaches intents of
+-- EVERY status, not just live ones: the FK does not care about `status`, and
+-- a terminal-status intent still carries the same composite pair and would
+-- still 23503 the moment the ticket's org_id changes. Same reasoning
+-- ticketService.ts's moveTicketOrg already documents for the ticket-level
+-- move (its own identical detach, all statuses, run before the ticket
+-- UPDATE that would otherwise trip this same constraint).
+--
+-- The immutability trigger (action_intents_block_content_update(), see
+-- 2026-09-25-ai-agents-ticket-triage.sql) permits exactly this non-null ->
+-- NULL transition for scope_ticket_id (mirroring the scope_device_id
+-- carve-out it already had), so this UPDATE is the tombstone path, not a
+-- bypass.
+--
+-- Placement: immediately after the scope_device_id tombstone and the
+-- tickets requester-contact detach, and BEFORE the generic
+-- breeze_device_child_orgid_tables() re-stamp loop — same "before the
+-- statement that trips the constraint" ordering the requester-contact
+-- detach already requires, matching moveOrg.ts's own statement order.
+--
+-- No historical backfill: same reasoning as the scope_device_id tombstone's
+-- migration header (2026-10-06-124500) — action_intents is FORCE ROW LEVEL
+-- SECURITY, so a contextless migration-time cleanup would match zero rows on
+-- a managed Postgres and report a silent "0 cleaned". Unlike scope_device_id
+-- this gap is a hard abort (23503), not a silently-stale pointer, so there is
+-- nothing pre-existing to clean up: a device org-move carrying a
+-- ticket-scoped intent could not have succeeded before this fix, on either
+-- the route (which never had this statement either) or this trigger.
+--
+-- Full function body copied verbatim from
+-- 2026-10-08-100500-detach-ticket-comment-runs-on-device-org-move.sql (the
+-- newest definition; no later migration replaces this function) with only
+-- the action_intents.scope_ticket_id statement added. CREATE OR REPLACE is
+-- idempotent by construction — re-applying this file re-installs the same
+-- body. The trigger itself (breeze_cascade_device_org_id ON devices, AFTER
+-- UPDATE OF org_id ... WHEN NEW.org_id IS DISTINCT FROM OLD.org_id) is
+-- unchanged and is NOT redeclared here.
+CREATE OR REPLACE FUNCTION public.breeze_cascade_device_org_id()
+  RETURNS trigger
+  LANGUAGE plpgsql
+  SECURITY DEFINER
+  SET search_path = public, pg_catalog
+  AS $$
+DECLARE
+  child_table text;
+BEGIN
+  -- Agent-run history stays with the SOURCE org (owner decision 2026-08-23):
+  -- sever the moved device's lineage links instead of re-stamping org_id.
+  UPDATE public.ai_agent_runs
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Findings:** 3 raised (2 Important, 1 Important-lower) → all 3 addressed in 8a32a66fd; 0 outstanding.
- code-reviewer: the new migration's header claimed its statement order matched `moveOrg.ts`'s, but the `scope_ticket_id` tombstone was inserted on the opposite side of the `requester_contact_id` detach in each file. Functionally harmless (disjoint tables, both still ran before the load-bearing loop) but misleading. Fixed by reordering the migration's `CREATE OR REPLACE` to match `moveOrg.ts` exactly.
- pr-test-analyzer: missing negative control — nothing proved the fix's `WHERE scope_ticket_id IN (SELECT id FROM tickets WHERE device_id = ...)` subselect is precisely scoped to the moving device, rather than accidentally broad. Added a second device/ticket/intent fixture (staying in the source org) asserted untouched across all three integration cases.
- pr-test-analyzer: `moveOrg.ts`'s own statement order wasn't statically asserted, even though it's just as load-bearing there as in the trigger (`tickets` is in `getDeviceOrgDenormalizedTables()`, which the route also loops over). Added the missing placement assertion to `moveOrg.coverage.test.ts`.

**Tests:** `moveOrg.coverage.test.ts` 27/27, `moveOrg.test.ts` 33/33, `autoMigrate.test.ts` 67/67 (unit); `deviceMoveOrgTicketScopeTombstone.integration.test.ts` 3/3 plus siblings (`deviceMoveOrgIntentScopeTombstone`, `deviceMoveOrgCurrency`, `agentRunMoveSemantics`, `ticket-move-org`, `actionIntentsImmutabilityTrigger`) 61/61 — all run against an isolated `pnpm test-stack` instance, all green. Manually verified red→green beforehand: with both the route and trigger statements removed, the integration suite failed with the exact reported `23503` on `action_intents_scope_ticket_org_fk`. `npx tsc --noEmit` clean; `eslint` on touched files clean; `pnpm db:check-drift` no drift; `scripts/check-migration-naming.sh` OK.

PR #4833, opened 2026-09-03
fix(api/tenancy): sever device_vulnerabilities.ticket_id on either org-move axis (#4645)

### LINKED ISSUE #4645: [API][Tenancy] device_vulnerabilities.ticket_id keeps naming a ticket that moved to another org
Found by the #4524 sweep. Not fixed in PR #4642 because the right behaviour is a product call, not a mechanical one.

**Problem:** `device_vulnerabilities.ticket_id` (`apps/api/src/db/schema/vulnerabilityManagement.ts`, plain single-column FK to `tickets.id` `ON DELETE SET NULL`) records the remediation ticket created for a finding. Its `org_id` is the **device's** org.

`moveTicketOrg` moves a ticket to another org of the same partner and already sets `tickets.device_id = NULL` — the device stays behind. But the same link's *other end* is untouched: the `device_vulnerabilities` row in the source org keeps pointing at a ticket the target org now owns.

**Consequence today is mild but real.** `vulnerabilityFleetQueries.ts` selects `ticketId` into the findings payload, and `apps/web/src/components/vulnerabilities/{CveDrawer,SoftwareGroupDrawer}.tsx` render it as a link to `/tickets#<number|id>`. Rendered in the device's own org, that link now resolves to nothing under RLS — a dead link, not a cross-tenant read. No AI tool dereferences the column.

There is also **no existing unlink path**: ticket deletion is `softDeleteTicket`, never a hard `DELETE`, so `ON DELETE SET NULL` never actually fires; closing a ticket doesn't touch this table either.

**The decision to make:** severing `ticket_id` on a move mirrors what the function already does with `tickets.device_id`, and is the consistent tenancy answer — but it silently discards the record that this finding *had* a remediation ticket, with nothing to regenerate it. Options:

- **A — null it in `moveTicketOrg`**: consistent with `device_vulnerabilities`' sibling detaches; loses remediation provenance for the device's org.
- **B — leave it and document** as an intentional exclusion, the way `invoice_lines.ticket_id` already is ("issued billing history must remain stamped with the org that was billed").
- **C — null it and record the departure** (e.g. keep a `remediation_ticket_number` snapshot the way the ticket keeps `submitter_email`/`submitter_name` on a requester detach).

C mirrors the pattern the codebase already reaches for elsewhere (drop the live link, keep the snapshot), but costs a column + migration.

Whichever is chosen, it should also cover the device axis: a device org-move re-stamps the device's tickets, so the same pointer goes stale from the other direction.

Related: #4524, PR #4642.

ToddHebebrand:
## Summary

Fixes #4645. Follow-up to #4642, which fixed the sibling `ai_agent_runs.ticket_id`
gap and explicitly deferred `device_vulnerabilities.ticket_id` pending a product
decision on the right behavior. That decision is now made (option A from the
issue): null the FK on divergence, consistent with #4642's precedent and with
the device-id detach `moveTicketOrg` already performs on `tickets.deviceId`.

**FK shape** (confirmed in `apps/api/src/db/schema/vulnerabilityManagement.ts`):
`device_vulnerabilities.ticket_id` is a **plain single-column** FK to
`tickets.id`, `ON DELETE SET NULL` — **not** the composite `(x, org_id) ->
tickets(id, org_id)` tenant-FK shape used elsewhere in this cascade. The table
**does carry a denormalized `org_id`** column (the device's org). Both facts
together mean neither statement below can ever `23503` — this is a
tenancy-hygiene fix, not a transaction-correctness one.

## Both axes, one PR

**Ticket axis** (`moveTicketOrg`, `apps/api/src/services/ticketService.ts`): a
finding's remediation ticket moves to another org while the finding stays with
its device — `device_vulnerabilities.org_id` is the **device's** org and is
never touched by a ticket move. Detach placed alongside the existing
`ai_agent_runs.ticket_id` detach (#4642), guarded by `org_id IS DISTINCT FROM
targetOrgId` so a finding whose device *already* lives in the ticket's
destination org (a legitimate, if unusual, prior state) is left untouched.

**Device axis** (`apps/api/src/routes/devices/moveOrg.ts` +
`breeze_cascade_device_org_id()`): `device_vulnerabilities` is already a
member of `getDeviceOrgDenormalizedTables()` / `breeze_device_child_orgid_tables()`,
so a finding's `org_id` always follows its device. Its remediation ticket does
**not**: `POST /vulnerabilities/tickets` (`apps/api/src/routes/vulnerabilities.ts`)
creates the ticket org-scoped only and never sets `tickets.device_id`, so the
same generic re-stamp loop (which *does* move any ticket bound to the moved
device) never reaches it — it stays behind in the source org.

The device-axis detach is placed **after** the generic re-stamp loop —
opposite the ordering used by the `scope_ticket_id` tombstones in the same
function — specifically so it compares against the referenced ticket's
**actual, post-restamp** `org_id`. A ticket that happens to be bound to the
*same* moving device is *also* re-stamped to the target org by that same
loop; checking before the loop would see its stale source org and wrongly
null a link that is about to become valid.

## Migration

New `2026-10-08-101000-cascade-device-org-move-vuln-ticket-detach.sql`
`CREATE OR REPLACE`s `breeze_cascade_device_org_id()` with the full body
copied verbatim from `2026-10-08-100900-cascade-device-org-move-ticket-scope.sql`
(the newest prior definition, merged minutes before this branch was cut) plus
the new statement appended after the generic loop. Nothing dropped.

No cascade-list or export-policy registration changes: `device_vulnerabilities`
was already fully registered (org cascade, device cascade, device-org-denormalized,
export policy) before this PR — no new table, no new column.

## Tests

- `apps/api/src/routes/devices/moveOrg.coverage.test.ts` — static statement-shape
  + ordering assertions for both callers (moveOrg.ts and the trigger function,
  the latter resolved dynamically against the newest migration that redefines it).
- `apps/api/src/__tests__/integration/ticket-move-org.integration.test.ts` —
  ticket axis, real Postgres: detaches on a sibling-ticket negative control,
  and a second test proving the `org_id IS DISTINCT FROM` guard leaves an
  already-consistent link alone.
- `apps/api/src/__tests__/integration/deviceMoveOrgVulnerabilityTicketDetach.integration.test.ts`
  (new) — device axis, real Postgres, both callers (the route and a raw
  `UPDATE devices` as the unprivileged `breeze_app` role, mirroring
  `deviceMoveOrgTicketScopeTombstone.integration.test.ts`'s pattern): detaches
  the stale org-scoped ticket, spares a same-device-bound ticket (precision
  control), and leaves an unrelated device/ticket/finding untouched (negative
  control).

## Verification

- `npx tsc --noEmit` (apps/api): clean.
- `pnpm exec eslint` on all touched files: clean.
- Affected unit suites: `ticketService.test.ts` (204 passed), `moveOrg.coverage.test.ts`
  (31 passed), `routes/tickets/moveOrg.test.ts` + `routes/devices/moveOrg.test.ts`
  (42 passed) — single-fork, own isolated `test-stack` (not the shared :5433 stack,
  which had drifted).
- Affected integration suites, real Postgres via `pnpm test-stack up`:
  `ticket-move-org.integration.test.ts` (18 passed, includes the 2 new cases),
  `deviceMoveOrgVulnerabilityTicketDetach.integration.test.ts` (3 passed, new file).
- Regression check on the trigger function replacement:
  `deviceMoveOrgTicketScopeTombstone.integration.test.ts`,
  `deviceMoveOrgIntentScopeTombstone.integration.test.ts`,
  `agentRunMoveSemantics.integration.test.ts` (12 passed) — confirms copying
  the full `breeze_cascade_device_org_id()` body forward didn't regress any
  prior fix in that function.

## Test plan

- [x] Real-Postgres integration tests for both axes, with negative/precision controls
- [x] Static coverage tests locking the statement shape + ordering for both callers
- [x] Typecheck + lint clean
- [x] No cascade/export-policy list changes needed (verified, not just assumed)

### DIFF
--- a/apps/api/migrations/2026-10-08-101000-cascade-device-org-move-vuln-ticket-detach.sql
+++ b/apps/api/migrations/2026-10-08-101000-cascade-device-org-move-vuln-ticket-detach.sql
@@ -1,0 +1,166 @@
+-- #4645 — device_vulnerabilities.ticket_id keeps naming a ticket that moved
+-- to another org (or was left behind by a device move). Decision made in the
+-- issue: option A, mirroring #4642's precedent (null the FK on divergence,
+-- same as every other stale-pointer detach breeze_cascade_device_org_id()
+-- already performs) and covering BOTH axes:
+--
+--   1. TICKET axis (services/ticketService.ts's moveTicketOrg): a finding's
+--      remediation ticket moves to another org via moveTicketOrg while the
+--      finding (device_vulnerabilities row) stays behind with its device —
+--      device_vulnerabilities.org_id is the DEVICE's org and is never
+--      touched by a ticket move. Fixed alongside this migration, in the same
+--      PR, by a `device_vulnerabilities.ticket_id = NULL` statement in
+--      moveTicketOrg itself (not this trigger — moveTicketOrg has no
+--      corresponding SQL trigger the way the device axis does).
+--
+--   2. DEVICE axis (this migration): the reverse direction. `device_id` FK
+--      to `devices.id`, plain (not composite), `org_id` uuid NOT NULL
+--      (`references(() => organizations.id)`, no onDelete clause — a bare FK,
+--      distinct from `ticket_id`'s `ON DELETE SET NULL`).
+--      `device_vulnerabilities` IS in getDeviceOrgDenormalizedTables() /
+--      breeze_device_child_orgid_tables(), so the generic loop below already
+--      re-stamps a finding's org_id to the device's new (target) org — the
+--      finding always travels with its device. Its remediation ticket does
+--      NOT: `POST /vulnerabilities/tickets` (routes/vulnerabilities.ts)
+--      creates the ticket org-scoped only and never sets `tickets.device_id`,
+--      so the `tickets` UPDATE inside this same generic loop (device-bound
+--      tickets only) never reaches it — the ticket stays in whatever org it
+--      was created in. Once the finding's org_id is the target org, a
+--      ticket_id still naming a SOURCE-org ticket resolves to nothing under
+--      RLS for any caller in the target org (a dead `/tickets#...` link
+--      client-side — CveDrawer/SoftwareGroupDrawer render it; nothing
+--      server-side dereferences the column).
+--
+-- `ticket_id` is a PLAIN single-column FK (not the composite
+-- `(x, org_id) -> tickets(id, org_id)` shape the action_intents/tickets
+-- tombstones above exist to satisfy), so this statement can never 23503
+-- either way — unlike those, it is not a correctness-of-transaction
+-- requirement, only a tenancy-hygiene one. Placed AFTER the generic
+-- `breeze_device_child_orgid_tables()` loop (not before, and not mixed in
+-- with the pre-loop tombstones above), specifically so it can compare
+-- against the referenced ticket's ACTUAL (post-restamp) org: a ticket that
+-- happens to be bound to this same device (`tickets.device_id = NEW.id`) is
+-- ALSO re-stamped to the target org by that same loop, and by the time this
+-- statement runs its org_id already matches — so the link is correctly left
+-- alone. Checking before the loop would see that ticket's stale SOURCE org
+-- and wrongly null a link that is about to become valid.
+--
+-- No historical backfill: unlike the scope_device_id/scope_ticket_id
+-- tombstones above, a stale device_vulnerabilities.ticket_id is not a hard
+-- abort (this FK cannot 23503), so pre-existing rows are not blocking
+-- anything — they are exactly the mild, already-shipped symptom (#4524
+-- sweep) this issue reports. Bulk-correcting them retroactively is a
+-- separate, deliberate backfill decision (RLS/system-context concerns per
+-- CLAUDE.md), not something to fold silently into a trigger-definition
+-- migration.
+--
+-- Full function body copied verbatim from
+-- 2026-10-08-100900-cascade-device-org-move-ticket-scope.sql (the newest
+-- definition; no later migration replaces this function) with only the
+-- device_vulnerabilities.ticket_id statement added, after the generic loop.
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer, silent-failure-hunter — all Sonnet)
**Findings:** 2 raised (both Important, from pr-test-analyzer) → both addressed in d0c8becfe; 0 outstanding.
- code-reviewer: 0 findings. Verified the ticket-axis `org_id IS DISTINCT FROM` guard, the device-axis post-loop placement (on both moveOrg.ts and the migration), the migration's fidelity as a full-body copy, lock-order/deadlock safety between the two axes, and that no cascade/export-policy registration changes were needed (device_vulnerabilities was already fully registered).
- silent-failure-hunter: 0 findings. Confirmed neither detach can RLS-silently-no-op given the pre-transaction org-access checks, confirmed rollback correctness, confirmed statement placement/typo-free.
- pr-test-analyzer: 2 Important — the ticket axis (`moveTicketOrg`) was missing a real-partner-scoped (non-system) RLS context test and a rollback-on-currency-block test for the new `device_vulnerabilities` detach, both patterns already present in this file for the sibling `ai_agent_runs.ticket_id` detach (#4524). Added both in d0c8becfe; verified discriminating by mutation (temporarily disabled the new statement, confirmed both new tests fail, restored it, confirmed all 20 tests in the file pass).
**Tests:** `ticket-move-org.integration.test.ts` (20/20, real Postgres, own isolated test-stack), `deviceMoveOrgVulnerabilityTicketDetach.integration.test.ts` (3/3, new file), `moveOrg.coverage.test.ts` (31/31), `ticketService.test.ts` (204/204), `routes/tickets/moveOrg.test.ts` + `routes/devices/moveOrg.test.ts` (42/42); regression check on the redefined trigger function via `deviceMoveOrgTicketScopeTombstone`, `deviceMoveOrgIntentScopeTombstone`, `agentRunMoveSemantics` (12/12). `tsc --noEmit` and `eslint` clean on all touched files.

PR #4839, opened 2026-09-03
fix(api): deliver empty watch list instead of null on zero enabled watches (#2949)

ToddHebebrand:
## Summary

`resolveDeviceMonitoringSettings` in `apps/api/src/routes/agents/helpers.ts` conflated two different outcomes into a single `null` return:

- **No monitoring policy matched this device** (correctly `null`).
- **A policy matched but has zero enabled watches** (previously *also* `null`).

`heartbeat.ts` only sets `monitoring_settings` on a non-null result, so the second case silently dropped the field from the heartbeat payload entirely. An MSP who removed all watches from a policy (intending to clear them) could never get that "clear" instruction delivered to an already-configured agent — the agent just kept running whatever watches it was told about on a previous heartbeat.

The agent side already handles an empty watches array correctly: `agent/internal/monitoring/monitor.go`'s `ApplyConfig`/`runChecks` early-return on `len(watches) == 0`, so no agent-side change is needed.

## Fix

`resolveDeviceMonitoringSettings` now returns `{ check_interval_seconds: winner.checkIntervalSeconds, watches: [] }` when a winning policy row exists but has no enabled watches. `null` is still returned only when no policy row matched at all (the earlier `rows.length === 0` check, unchanged).

## Scope (per issue #2949)

This addresses **item 1 only**:

1. ✅ Zero enabled watches now yields `{ watches: [] }`, not `null` — fixed here.
2. Partner-wide monitoring policy visibility — already fixed by #4673 / #3665 (confirmed present on `origin/main`, code comments in `helpers.ts` reference the fix).
3. `alertSeverity` delivery / alert-creation audit — **out of scope for this PR**. There is no corresponding field on the Go agent side yet (`agent/internal/monitoring`), so this needs its own design/implementation pass. Left open on the issue.

Refs #2949 (not closing — items 2 addressed elsewhere, item 3 still open).

## Testing

- Added a red-first regression test to `helpers.partnerWidePolicies.test.ts`: a winning policy with zero enabled watches now asserts a non-null `{ watches: [] }` result, and a companion test pins that "no policy matched at all" still returns `null`.
- `apps/api`: `npx vitest run src/routes/agents/helpers.partnerWidePolicies.test.ts src/routes/agents/heartbeat.test.ts` — 204 passed.
- `apps/api`: `npx vitest run src/routes/agents/helpers` (all helpers*.test.ts in the dir) — 128 passed.
- `apps/api`: `npx tsc --noEmit` — clean.

### DIFF
--- a/apps/api/src/routes/agents/helpers.partnerWidePolicies.test.ts
+++ b/apps/api/src/routes/agents/helpers.partnerWidePolicies.test.ts
@@ -448,3 +448,40 @@ describe('partner-owned policies actually reach the agent payload', () => {
     expect(systemEscapeMock).not.toHaveBeenCalled();
   });
 });
+
+describe('monitoring: a matched policy with zero enabled watches (#2949)', () => {
+  // A winning policy row with no enabled watches ("clear all watches") is a
+  // different fact than "no policy matched at all" — collapsing both to `null`
+  // means heartbeat.ts omits monitoring_settings from the payload entirely, and
+  // the agent (which handles an empty watches array fine — see
+  // agent/internal/monitoring/monitor.go ApplyConfig) never learns to clear out
+  // watches it was told about on a previous heartbeat.
+  it('returns an object with an empty watches array, not null', async () => {
+    dbMock._resetQueue([
+      deviceRow,
+      orgWithPartner,
+      [],
+      [{ level: 'organization', assignmentPriority: 1, settingsId: 'set-1', checkIntervalSeconds: 90 }],
+      [], // no enabled watches for the winning settings row
+    ]);
+
+    const result = await buildMonitoringConfigUpdate(DEVICE_ID);
+
+    expect(result).not.toBeNull();
+    expect(result?.check_interval_seconds).toBe(90);
+    expect(result?.watches).toEqual([]);
+  });
+
+  it('still returns null when no policy row matches at all', async () => {
+    dbMock._resetQueue([
+      deviceRow,
+      orgWithPartner,
+      [],
+      [], // no assignment/policy rows matched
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` + `pr-test-analyzer` (Sonnet), both scoped to the diff.

**Findings:**
- `code-reviewer`: 0 critical, 0 important. Traced the full call chain end-to-end (`resolveDeviceMonitoringSettings` → `buildMonitoringConfigUpdate` → `heartbeat.ts` → agent-side `monitor.go`/`ParseMonitorConfig`) and confirmed the two null cases are now correctly distinguished, the Redis-cache gate (`if (redis && settings)`) already handles the new truthy-empty-array shape correctly, and no other call site assumes the old collapsed semantics. Verdict: mergeable as-is.
- `pr-test-analyzer`: 0 blocking gaps. Verified the new tests are non-vacuous by temporarily reverting the fix and confirming the test fails with the exact `expected null not to be null` error. Flagged one low-priority (3/10) nice-to-have: no test proved the empty-watches result actually reaches `redis.set`. Addressed inline in `2136f2b5e` (mirrors the existing cache-write pattern in `helpers.pam.test.ts`).

**Tests:** `apps/api`: `npx vitest run src/routes/agents/helpers.partnerWidePolicies.test.ts` — 22/22 passed; `npx vitest run src/routes/agents/helpers` (all helpers*.test.ts) — 129/129 passed. `npx tsc --noEmit` — clean.

**Status:** review-clean, both rounds' findings addressed, awaiting maintainer merge.

Issue #2621, opened 2026-07-19
[Agent] Token rotation can leave stale credentials on disk and permanently strand the agent after restart


ramphex:
## Description

Agent credential rotation can succeed on the server while the replacement credentials fail to persist locally. The running agent continues using the new credentials in memory, but `secrets.yaml` retains the previous agent and watchdog tokens.

After the agent or device restarts, it loads the stale credentials from disk. Once the five-minute previous-token grace period has expired, every authenticated request returns `401 Invalid agent credentials`, and the agent cannot recover automatically.

The current rotation sequence in `agent/internal/heartbeat/heartbeat.go` is:

1. Request new agent, watchdog, and helper credentials from the API.
2. Commit the corresponding token hashes on the server.
3. Replace the credentials in agent memory.
4. Attempt to persist the credentials with `config.Save`.
5. If persistence fails, log the error and continue using the in-memory credentials.

The server has already committed the new hashes before `config.Save` runs. There is no persistence retry, confirmation step, rollback, or supported administrator recovery workflow after the previous-token grace period expires.

This produces a delayed failure. The agent can remain healthy after rotation but become permanently unauthorized following its next restart.

## Steps to Reproduce

1. Enroll a Windows agent and confirm that the agent and watchdog are online.

2. Cause writes to `C:\ProgramData\Breeze\secrets.yaml` to fail during a server-requested credential rotation. This can be reproduced in a controlled environment by blocking the atomic replacement, changing permissions, or injecting a `config.Save` failure.

3. Trigger the normal agent credential-rotation path.

4. Confirm that the API commits new agent, watchdog, and helper token hashes.

5. Allow the agent to continue operating with the new credentials in memory.

6. Wait until the previous-token grace period expires.

7. Restart the Breeze agent service or restart Windows.

8. Observe that the agent loads the previous credentials from `secrets.yaml`.

9. Observe `401 Invalid agent credentials` responses from heartbeat, inventory, WebSocket, watchdog failover, and other authenticated endpoints.

## Expected Behavior

Credential rotation should not leave the server and endpoint with permanently mismatched credentials.

The server should keep the existing credentials valid until the agent confirms that all replacement credentials have been durably written and read back successfully.

If persistence fails, the agent should retain the complete pending credential set and retry persistence. A failed or abandoned rotation should expire safely without invalidating the last known durable credentials.

[...]

PR #4840, opened 2026-09-03
fix(agent): retry secrets.yaml rename on transient destination lock (#2772)

### LINKED ISSUE #2772: [API][Agent] Two-phase token rotation never confirms — agent.token.rotate.confirmed has never fired in production
**Symptom:** In US production, `agent.token.rotate.staged` has been written 52 times across 3 devices since 2026-07-20. `agent.token.rotate.confirmed` has **never** been written — zero rows, all time, fleet-wide.

## What that means

Phase two of the #2621 / PR #2647 two-phase rotation is not completing anywhere. The server stages a credential set into the `pending_*` columns, the agent never successfully calls `POST /agents/:id/rotate-token/confirm`, the staged set expires after `PENDING_ROTATION_TTL_MS`, and because `token_issued_at` never advances it stays older than `AGENT_TOKEN_ROTATION_MAX_AGE_DAYS` (default 30, `agentAuth.ts:62`) — so `isAgentTokenRotationDue` is still true and the next heartbeat sets `rotateToken` again.

The result is a permanent stage → expire → re-stage loop, running roughly hourly. One device has staged 42 rotations in 7 days; its `token_issued_at` has not moved since 2026-06-18.

## Why this is not currently an outage

The two-phase design is sound under exactly this failure, which is why it has gone unnoticed. `agent_token_hash` is never demoted during staging (`token.ts:102-138`), so affected agents keep authenticating on their existing current credential indefinitely. Two of the three looping devices are online and healthy right now.

The cost today is therefore not downtime but churn and a silent policy gap: a pointless rotation round-trip per affected device per hour, an audit row each time, and — the part that matters — **credentials that never actually rotate**. The 30-day rotation policy is effectively not being enforced anywhere in the fleet.

## Where to look

The agent's confirm call after `config.StagePendingCredentials` in `agent/internal/heartbeat/heartbeat.go:3674-3699`. Against a current server `rotateResp.ConfirmationRequired` is true, so the pre-#2621 immediate-promote branch should not be taken.

Either the confirm request is never issued, or it is issued and rejected. Worth noting that `/rotate-token/confirm` returns 409 `pending_token_required` unless the caller authenticates with the **staged** token (`token.ts:249-255`) — that is the whole point of the phase-two proof — which requires the agent to have swapped its in-memory credential to the new one before calling. A confirm sent with the old token is refused by design.

**Status:** Not root-caused on the agent side yet — filing the server-side evidence first, since the audit trail is unambiguous. No code change proposed.

ToddHebebrand:
## Summary

`atomicWriteFile` and `writeYAMLFile` in `agent/internal/config/config.go` each called `os.Rename` exactly once when swapping their staged tmp file into place. On Windows, anything that briefly holds the destination open — reproduced in release QA as Windows Defender real-time scanning with zero exclusions configured — fails that single attempt outright, with no retry.

`StagePendingCredentials` (the phase-two credential persistence for two-phase token rotation, #2621) treats any write error from this path as "abort and discard the staged rotation." So one transient rename failure permanently loses that rotation attempt: the server's staged credential set then expires via `PENDING_ROTATION_TTL_MS`, `token_issued_at` never advances, `isAgentTokenRotationDue` stays true, and the next heartbeat stages again. This is the stage→expire→re-stage loop reported in this issue — `agent.token.rotate.confirmed` has never fired in production, while `agent.token.rotate.staged` climbs roughly hourly per affected device.

Root-caused and reproduced on a real Windows Server 2022 agent (v0.104.0) during release QA (see issue comment): forcing rotation three times produced the identical failure each time —

```
error="staging pending credentials: rename C:\ProgramData\Breeze\secrets.yaml.partial
C:\ProgramData\Breeze\secrets.yaml: The process cannot access the file because it is being used by another process."
```

## Fix

Adds `renameWithRetry`, a bounded jittered-backoff retry around the rename, mirroring the pattern already used in `agent/internal/state/state.go` (`Write`, `renameAttempts`/`renameBackoffBase`). Both `atomicWriteFile` and `writeYAMLFile` now route their rename through this shared helper instead of calling `os.Rename` directly once.

- **Longer schedule than state.go's:** 8 attempts vs state.go's 4 — state.go's own comment notes its 4 attempts (~175ms total) still lost on the exact machine that reproduced this bug, and the credential file is the one write that must not be silently dropped.
- **Jitter** added on top of the exponential backoff, since multiple goroutines (cert-renewal, command worker pool, heartbeat's own rotation path) can all reach a rename around the same time and would otherwise retry against the same transient holder in lockstep.
- Runs off `handleTokenRotation`'s own goroutine (`go h.handleTokenRotation()`), not the main heartbeat tick, so the added worst-case latency (a few seconds) doesn't block heartbeats.
- `renameFile` is a package-level seam (`var renameFile = os.Rename`) so tests can inject transient and permanent rename failures without touching the real filesystem semantics.

Deliberately scoped to the retry/backoff mirror the issue calls out — did not adopt `state.go`'s Windows POSIX-semantics rename primitive (`renameReplace`/`FILE_RENAME_INFO`), since that solves a different failure mode (a concurrent *reader* blocking `MoveFileEx`) than the one diagnosed here (a transient *exclusive* holder like an AV scanner), which `state.go`'s own comment notes POSIX semantics cannot bypass anyway — retry is the correct and sufficient fix for this failure mode.

## Testing

- `TestAtomicWriteFileRetriesTransientRenameFailure` — 2 injected transient failures then success; asserts content lands and no `.partial` leftover.
- `TestAtomicWriteFileExhaustsRenameRetriesAndCleansUp` — persistent failure; asserts it gives up after exactly `renameAttempts`, returns an error, and cleans up the `.partial` tmp (the signal `StagePendingCredentials` relies on to know the write did NOT land).
- `TestWriteYAMLFileRetriesTransientRenameFailure` — same coverage for the second call site.
- `go test -race ./internal/config/...` — pass.
- `go test -race ./internal/state/...` — pass (unaffected, verifies the pattern being mirrored still works).
- `go vet ./internal/...` — clean.
- `gofmt -l` on changed files — clean.

Closes #2772

### DIFF
--- a/agent/internal/config/config.go
+++ b/agent/internal/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	mathrand "math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -893,6 +894,51 @@ func stripSecretsFromAgentConfig(data []byte) ([]byte, error) {
 	return out, nil
 }
 
+// Rename retry bounds for atomicWriteFile / writeYAMLFile. A rename-over-
+// existing destination can fail transiently — observed in production on
+// Windows Server 2022 as Windows Defender real-time protection (zero
+// exclusions configured) briefly holding secrets.yaml open during the write.
+// A single failed attempt used to be treated as a hard failure:
+// StagePendingCredentials aborts the rotation and discards the staged
+// credential set on ANY write error, so one transient lock permanently lost
+// that rotation attempt. Because the server's staged set then expires and
+// token_issued_at never advances, isAgentTokenRotationDue stays true and the
+// next heartbeat stages again — the stage→expire→re-stage loop in issue
+// #2772 (agent.token.rotate.confirmed has never fired in production).
+//
+// Mirrors the retry/backoff shape in agent/internal/state/state.go (Write),
+// but with a longer schedule: state.go's own comment notes its 4 attempts /
+// ~175ms total still lost on the box that reproduced this bug, and the
+// credential file is the one write that must not be silently dropped.
+const (
+	renameAttempts    = 8
+	renameBackoffBase = 25 * time.Millisecond
+)
+
+// renameFile is a seam for tests to inject rename failures. Defaults to
+// os.Rename in production.
+var renameFile = os.Rename
+
+// renameWithRetry retries a failing rename with jittered exponential backoff.
+// Jitter keeps concurrent callers (the cert-renewal goroutine, the command
+// worker pool, and heartbeat's own SetAndPersist/StagePendingCredentials
+// calls can all reach a rename around the same time) from retrying against
+// the same transient holder in lockstep.
+func renameWithRetry(oldpath, newpath string) error {
+	var err error
+	for attempt := 0; attempt < renameAttempts; attempt++ {
+		if attempt > 0 {
+			backoff := renameBackoffBase << (attempt - 1)
+			sleep := backoff + time.Duration(mathrand.Int63n(int64(backoff)+1))
+			time.Sleep(sleep)
+		}
+		if err = renameFile(oldpath, newpath); err == nil {
+			return nil
+		}
+	}
+	return fmt.Errorf("rename %s to %s after %d attempts: %w", oldpath, newpath, renameAttempts, err)
+}
+
 // atomicWriteFile writes data to path durably: it creates path+".partial" in
 // the same directory with perm requested at open time (still subject to umask
 // on POSIX — defense-in-depth Chmod happens at the call site), writes data,
@@ -933,7 +979,7 @@ func atomicWriteFile(path string, data []byte, perm os.FileMode) error {
 		os.Remove(tmpPath)
 		return err
 	}
-	if err := os.Rename(tmpPath, path); err != nil {
+	if err := renameWithRetry(tmpPath, path); err != nil {
 		os.Remove(tmpPath)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `pr-review-toolkit:code-reviewer` (Sonnet, focused on retry-loop correctness, root-cause match, test quality, blast radius)
**Findings:** 1 raised → addressed in `bb578c630`; 0 outstanding.
- Non-blocking: `TestAtomicWriteFileExhaustsRenameRetriesAndCleansUp` burned ~5.3s real wall-clock via genuine `time.Sleep` across 8 retry attempts. Fixed by adding a `renameRetrySleep` seam (mirrors the existing `renameFile` seam) — same assertions, now 0.00s.
- No correctness, root-cause-match, or blast-radius issues found. Confirmed: retry loop has no off-by-one/overflow/panic risk; `renameFile`/`renameRetrySleep` package vars are only mutated in non-parallel tests (no concurrency hazard); the fix sits on the actual `StagePendingCredentials → mutateSecretsAndPersist → atomicWriteFile` path that was hard-aborting; `handleTokenRotation` runs on its own goroutine so added worst-case latency doesn't block the heartbeat tick; the shared `persistMu` serialization means a stuck rename can now hold the mutex up to ~6.35s instead of failing instantly — a real but intentional durability-over-latency tradeoff, not a defect.
**Tests:** `go test -race ./internal/config/...` and `./internal/state/...` green; `go vet ./internal/config/... ./internal/state/... ./internal/heartbeat/...` clean; `gofmt -l` clean on changed files.

Issue #4834, opened 2026-09-03
perf: device Activities tab count(*) reads the device's whole telemetry history (118 s on US)


ToddHebebrand:
## Symptom

`GET /devices/:id/events?withTotal=true` (the device **Activities** tab, `DeviceEventLogViewer.tsx`) runs an unbounded `count(*)` over the device's whole audit history. Measured on US prod as `breeze_app` on 2026-09-03 for a device in the largest org:

```
Parallel Index Scan using audit_logs_resource_id_timestamp_idx
  Index Cond: (resource_id = <device>)
  Filter: org_id = <org> AND <RLS policy>
  rows=76,317   Buffers: shared hit=10956 read=63806
Execution Time: 118,329 ms
```

76k rows for one device because ~99% of `audit_logs` is agent telemetry (`agent.logs.submit`, `agent.sessions.submit`, ...). Every row needs a heap fetch: the RLS policy reads `org_id`, which is not in the index, and the table has never been vacuumed so there is no visibility map for an index-only scan.

## Context

The two-arm feed change (PR for the 2026-09-03 US slowdown) fixed the page *read* (63 ms) and the overview feed (0.85 s cold, was 90 s mean / 13 min worst), and bounded this count to the device's own rows instead of the org's 2.4M. The count itself is still the same shape and still 100+ s under IO pressure.

## Options

1. **De-telemetry `audit_logs`** (#4340 / #4021) — removes the 99% and makes the count trivial. Root fix.
2. **Cap the count** — `count(*) FROM (SELECT 1 ... LIMIT 10001)` and render `10,000+` in the tab. Needs a small web change for the pagination label.
3. **Index-only path** — add `org_id` to the resource index and get `audit_logs` vacuumed so the VM is populated. Fragile: append-only tables never trigger dead-tuple autovacuum; relies on insert-threshold autovacuum keeping up.

Recommend 1, with 2 as the stop-gap if 1 slips another release.

ToddHebebrand:
[...]

Issue #4340, opened 2026-09-01
[API] Agent telemetry is 97% of audit_logs volume — unsustainable at scale


ToddHebebrand:
## Description
Measured on US production (378 devices, 2026-09-01): `audit_logs` receives ~140k rows/day and growth is accelerating (90k→142k/day over the last week). Breakdown of a 24h window:

| action | rows/day |
|---|---|
| `agent.logs.submit` | 81,491 |
| `agent.heartbeat.state_change` | 20,780 |
| `agent.sessions.submit` | 15,068 |
| `agent.security_status.submit` | 15,032 |
| `agent.management_posture.submit` | 5,620 |

~97% is routine agent telemetry, written into an immutable, hash-chained (`audit_log_seal_one` trigger + `audit_log_chain` row per insert), 365-day-retention table. Current cost: 5.06M rows / ~6.7 GB (logs+chain incl. indexes) since April — ~0.5 MB/device/day, the single largest storage line item. At the 10k-device target this is ~5 GB/day of audit writes.

Secondary cost: the daily `audit-chain-verify` job walks the whole chain per org — measured 121s mean per run, ~30% of a DB vCPU daily, and it scales linearly with this volume.

## Root Cause
Agent ingest paths (log submit, heartbeat state changes, session/security-status submits) call the audit writer for high-frequency machine events. The audit chain design is priced for human/admin actions, not per-request telemetry.

## Proposed Fix
- Stop writing routine agent telemetry to `audit_logs`. Options: drop the entries entirely (the underlying data already lands in `agent_logs`, `device_sessions`, security posture tables), or route them to a cheap unchained operational-events table with short retention.
- Keep auditing genuinely security-relevant agent events (token rotation, enrollment, elevation).
- Follow-on: make `audit_log_verify_chain` incremental from the last verified anchor instead of full-chain per day.

## Affected Files
- agent ingest routes under `apps/api/src/routes/agents/` (audit call sites for the actions above)
- `apps/api/src/jobs/auditChainVerify.ts`

Issue #4021, opened 2026-08-25
[API] Routine agent telemetry is 99.93% of audit_logs — 27 GB with the chain, and it buries the events an auditor needs


ToddHebebrand:
**Routine agent telemetry submissions are written to `audit_logs` as audit events. They are 99.93% of the rows on a real fleet, they are not user actions, and they are burying the events the audit trail exists to record.**

Field measurement from a production self-hosted instance (~160 agents, live since 2026-05-06, v0.105.1), contributed by @bdunncompany on #3991:

```
             action              |  rows   | pct
---------------------------------+---------+------
 agent.logs.submit               | 8066036 | 39.0
 agent.security_status.submit    | 4501681 | 21.7
 agent.sessions.submit           | 4499896 | 21.7
 agent.management_posture.submit | 1482347 |  7.2
 agent.heartbeat.state_change    | 1236737 |  6.0
 agent.patches.submit            |  588514 |  2.8
 agent.eventlogs.submit          |  219970 |  1.1

agent.* rows: 20,691,455    total: 20,706,753    (99.93%)
```

Storage: `audit_logs` **16 GB / 19.8M rows** plus `audit_log_chain` **11 GB / 19.8M rows** — **27 GB, ~31% of an 86 GB database.** Row growth roughly doubles every two months and tracks fleet size, not operator activity (2026-05: 2.1M → 2026-08, 25 days: 8.8M).

**These are explicit calls, not the fallback middleware.** Verified against `origin/main` before filing, because the fix is a different shape depending on the answer:

- `apps/api/src/routes/agents/logs.ts:141` — `action: 'agent.logs.submit'`
- `apps/api/src/routes/agents/security.ts:37` — `action: 'agent.security_status.submit'`
- `apps/api/src/routes/agents/sessions.ts:178` — `action: 'agent.sessions.submit'`

So this is a deliberate decision that can be revisited without touching the catch-all HTTP audit middleware at all — that layer is #3991's business and is a separate, much smaller problem.

**Why they don't belong there.** An audit trail answers "which principal did what, and when, and can we prove it wasn't altered." A scheduled submission from an agent doing its job has no accountable principal, no discretion, and no compliance question attached to any individual row. It is device event data. Meanwhile the rows that *do* answer auditor questions — `user.login`, `quote.superseded`, `action_intent.rejected`, a billing-currency change — are 0.07% of the table and are found by paging past twenty million machine writes.

[...]

PR #4842, opened 2026-09-03
fix(api): cap the device Activities-tab audit count at 10,000 (#4834)

ToddHebebrand:
## Why

Stacked on #4835 (open). `GET /devices/:id/events?withTotal=true` (the device
**Activities** tab) ran an unbounded `count(*)` per feed arm — even after
#4835 bounded the *arm predicate* to the device's own rows, the count still
walked the device's whole audit history: 76k rows / 118s on US prod, because
~99% of `audit_logs` is agent telemetry the feed never renders, RLS's `org_id`
filter forces a heap fetch per row, and the table is never vacuumed (no
visibility map for an index-only count).

This is the **stop-gap (option 2)** from #4834, not option 1 (de-telemetry) or
option 3 (index-only path). Partially addresses #4834; root fix remains
#4340 / #4021.

## What

- `countArm` now caps each arm's count at `FEED_TOTAL_CAP` (10,000, exported)
  rows via `SELECT count(*) FROM (SELECT 1 FROM audit_logs WHERE <arm
  predicate> LIMIT 10001) capped_rows` — a Drizzle derived-table subquery.
  Postgres stops scanning once the LIMIT is hit instead of walking the rest of
  the device's history, so cost is now bounded regardless of how large that
  history grows. Verified the compiled SQL directly:
  ```sql
  select count(*)::int from (select 1 from "audit_logs" where "audit_logs"."org_id" = $1 limit $2) "capped_rows"
  ```
  The inner SELECT stays on `auditLogs` alone (no join to `users`) and reuses
  each arm's `where` unchanged, so it hits the same index the row read for
  that arm already uses. `mergeFeedPage`, `NON_AGENT_ACTOR`,
  `DETAILS_HAS_DEVICE_ID`, and `FEED_SORT_KEY` are untouched.
- The two capped arm counts are summed as before. `pagination` gains
  `totalIsLowerBound: boolean` — true when the summed raw count exceeds
  `FEED_TOTAL_CAP`, in which case `total` is clamped to the cap. When
  `withTotal` is false the response is byte-for-byte unchanged (`total: null`,
  no count query, no `totalIsLowerBound` key).
- Web: `DeviceEventLogViewer.tsx` (the only caller that sends
  `withTotal=true`) renders `10,000+` via a new i18n key
  (`deviceEventLogViewer.totalAtLeast`, added to all 8 locale catalogs) when
  `totalIsLowerBound` is true, and keeps "next page" enabled past the
  (possibly capped) `totalPages` whenever the just-fetched page came back full
  (`data.length === limit`) — the only signal there might be more rows beyond
  the cap. `DeviceActivityFeed.tsx` never sends `withTotal` and is untouched.

## Tests

- `apps/api/src/routes/devices/events.test.ts`: extended the `db` mock to
  distinguish the inner capped-count derived-table select (`{ one }`
  projection) from the outer `count(*)` select (`{ count }` projection) and
  made the resolved count configurable per test. New cases: below-cap sum is
  exact with `totalIsLowerBound=false`; above-cap sum clamps to
  `FEED_TOTAL_CAP` with `totalIsLowerBound=true`; the count SQL is provably
  bounded (`LIMIT FEED_TOTAL_CAP + 1` captured and asserted) and reuses each
  arm's `where` unchanged (byte-identical to the row read's `where`, via
  `JSON.stringify`). The existing "no count query when withTotal is omitted"
  case stays green. 34/34 passing.
- `apps/web/src/components/devices/DeviceEventLogViewer.test.tsx` (new):
  exact total render when `totalIsLowerBound=false`; `10,000+` render when
  true; Next stays enabled past the reported total when the last page came
  back full; Next disables once a page comes back partial even when
  `totalIsLowerBound` is true. 4/4 passing. Locale parity suite (all 8
  catalogs) still green with the new key.
- `tsc --noEmit` clean on both `apps/api` and `apps/web`. `eslint` clean on
  every touched file.
- Did **not** run integration tests (need a live Postgres stack) — the arm
  predicates/indexes this reuses are unchanged, so #4835's
  `deviceEventsFeedIndexes.integration.test.ts` is unaffected by this diff.

## Note

This branch targets `db-performance` (PR #4835, open) since the count logic
depends on the two-arm predicate that PR introduces. **This PR needs to be
rebased onto `main` after #4835 squash-merges** — the follow-up rebase is
`git rebase --onto main <old-parent-sha>`. Because the base isn't `main`, GitHub
runs no CI here automatically; dispatching `CI` manually against this branch
per the tenancy-doc guidance.

https://claude.ai/code/session_016g2n6u6D73cH6HxdLPgM2G

### DIFF
--- a/apps/api/src/routes/devices/events.test.ts
+++ b/apps/api/src/routes/devices/events.test.ts
@@ -1,41 +1,75 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { Hono } from 'hono';
 
-// Tracks how many times the count(*) query (db.select({ count })...where()) is
-// issued, so tests can prove the unbounded count is skipped unless withTotal is
-// set. The count select is distinguishable from the row select by its shape:
-// the count projection has a `count` key; the row projection does not.
+// Tracks how many times the capped count query (the OUTER `select({ count
+// })...from(<capped derived table>)`) is issued, so tests can prove the count
+// is skipped unless withTotal is set. Distinguished from the row select by
+// its projection shape: the count projection has a `count` key, the row
+// projection does not.
 const countQueryCalls = vi.fn();
 // Captures the WHERE condition handed to the feed query so tests can prove the
 // org_id scoping is present (BREEZE-B — it is load-bearing for performance, not
 // cosmetic; see the comment in events.ts).
 const feedWhereArgs: unknown[] = [];
+// Captures the WHERE/LIMIT the capped-count derived table builds (#4834), so
+// tests can prove (a) it reuses the arm's `where` unchanged — same predicate,
+// same index — and (b) the count is genuinely bounded, never an unbounded
+// count(*). `mockCappedCounts` is consumed FIFO in build order (resource arm,
+// then details arm) so a test can drive below-cap / above-cap scenarios.
+const cappedCountWhereArgs: unknown[] = [];
+const cappedCountLimitArgs: { limit: number }[] = [];
+let mockCappedCounts: number[] = [];
 
 vi.mock('../../db', () => ({
   runOutsideDbContext: vi.fn((fn) => fn()),
   withDbAccessContext: vi.fn(async (_ctx: unknown, fn: () => Promise<unknown>) => fn()),
   withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
   db: {
-    select: vi.fn((projection?: Record<string, unknown>) => ({
-      from: vi.fn(() => ({
-        leftJoin: vi.fn(() => ({
-          where: vi.fn((cond: unknown) => {
-            feedWhereArgs.push(cond);
-            return {
-              orderBy: vi.fn(() => ({
-                // Each feed arm is a bounded `limit(offset + limit)` read; the
-                // page is cut client-side by mergeFeedPage. No `.offset()`.
-                limit: vi.fn().mockResolvedValue([])
-              }))
-            };
+    select: vi.fn((projection?: Record<string, unknown>) => {
+      // Inner capped-count derived table: `SELECT 1 FROM audit_logs WHERE
+      // <arm predicate> LIMIT (cap + 1)`, later wrapped in `.as(...)`.
+      if (projection && 'one' in projection) {
+        return {
+          from: vi.fn(() => ({
+            where: vi.fn((cond: unknown) => {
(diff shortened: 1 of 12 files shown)


ToddHebebrand:
**Review run:** codex review (post-open) + manual CI dispatch (base isn't `main`, so no automatic run)
**Findings:** 2 raised → both addressed in 4a11686c9; 0 outstanding.
1. **CI (Test Web)**: the new `deviceEventLogViewer.totalAtLeast` key was an exact-English copy in every locale, pushing tr-TR's `devices.json` past its exact-duplicate baseline (77 → 78, zero slack). Fixed with a real "N or more" translation in all 8 locale catalogs (tr-TR: `"{{count}}+"` → `"{{count}} ve üzeri"`) instead of bumping the baseline, plus a companion `noMoreActivity` key for the empty-page case below.
2. **Codex (P2)**: `DeviceEventLogViewer`'s Next-button/range logic compared `page` to `totalPages` derived from the (possibly capped) total. When the true count lands exactly on a page-size boundary above `FEED_TOTAL_CAP` (e.g. 10,050 with a 50-row page), the last full page still reports `totalIsLowerBound`, so Next stayed enabled and the following page came back empty, rendering an inverted `"10051–10050 of 10,000+"` range. Fixed by keying Next purely off whether the last fetched page came back full (`activities.length === limit` — true until the genuine end, cap-aware or not) and rendering "No more activity" instead of a range when a page returns zero rows. Added a regression test for the exact-boundary case.

**Tests:**
- `apps/api`: `vitest run src/routes/devices/events.test.ts` — 34/34 green (unaffected by this follow-up, API side untouched).
- `apps/web`: `DeviceEventLogViewer.test.tsx` 5/5, `localeParity.test.ts` 79/79, `translationCoverage.test.ts` 15/15 — all green.
- `tsc --noEmit` clean on `apps/api` and `apps/web`; `eslint` clean on every touched file.

**CI:** dispatched manually against `fix/4834-capped-activities-count` (run [33793324946](https://github.com/LanternOps/breeze/actions/runs/33793324946)) — all jobs green, including Test Web, all 4 Integration Tests shards, Smoke Test, and CI Success.

**Status:** review-clean, awaiting maintainer merge. Still stacked on #4835 (open) — rebase `fix/4834-capped-activities-count` onto `main` with `git rebase --onto main <old-parent-sha>` after #4835 squash-merges.

PR #4843, opened 2026-09-03
fix(api): flip stranded pending device_patches rows using version compare (#2736)

### LINKED ISSUE #2736: [API] device_patches rows can strand at 'pending' forever when a source's pending scan chronically fails but installed inventory keeps succeeding
## Description

Surfaced by the adversarial review of #2725. The pending-preservation guard is correct for the normal paired-submit flow, but it removes the (accidental, blunt) self-correction the old unconditional installed-flip provided in one degraded scenario:

- A provider whose pending `Scan()` errors is excluded from `coveredSources` (`agent/internal/patching/manager.go`), so the source-scoped sweep in `markPendingDevicePatchesMissing` never runs for that source — by design (#2217), pending rows are kept rather than tombstoned.
- The installed upsert now only transitions a row out of `pending` after the sweep has moved it to `missing`. It never compares the reported installed version against the version that made the row pending.

So if `winget upgrade` chronically fails on a device (source-sync/network flakiness is common) while `winget list` keeps succeeding, a row can stay `pending` **forever** — even after the upgrade is actually installed — with no log, no alert, and no reconciliation. `patchComplianceReportWorker` feeds the raw pending count into `compliancePercent`, so the device's compliance/posture numbers are silently inflated indefinitely.

## Proposed Fix

Bounded reconciliation, one of:
- Version-aware flip: when the installed submit reports a version equal to (or newer than) the patch's known target version, allow the flip out of `pending` regardless of sweep coverage. (Note: `patches.version` currently flip-flops between target and installed version because both upserts write it — that wart needs fixing first for this to be reliable.)
- Staleness window: surface (metric/log) and/or expire `pending` rows whose source hasn't produced a successful pending scan in N days, in the spirit of `PATCH_TOMBSTONE_PRUNE_AFTER_HOURS`.

Related: #2725 (the guard), #2726 (winget skip counted as covered — fixing that increases how often this scenario occurs, since more failures will correctly suppress the sweep).

## Affected Files

- `apps/api/src/routes/agents/patches.ts` (`upsertInstalledPatches`, `markPendingDevicePatchesMissing`)
- `apps/api/src/jobs/patchComplianceReportWorker.ts` (consumer of the inflated pending count)
- `agent/internal/patching/manager.go` / `agent/internal/heartbeat/heartbeat.go` (coverage semantics)

ToddHebebrand:
## Summary

`upsertInstalledPatches` (`apps/api/src/routes/agents/patches.ts`) only cleared a `pending` `device_patches` row after `markPendingDevicePatchesMissing` had already swept it to `missing`. A source whose pending `Scan()` chronically errors (e.g. a flaky `winget upgrade`) is excluded from `coveredSources` by design (#2217), so the sweep never runs for that source — even while the paired installed inventory (`winget list`) keeps succeeding and proves the upgrade actually landed. The row could stay `pending` forever, with no log or alert, silently inflating `compliancePercent` (`patchComplianceReportWorker`).

## Fix

Compute the flip decision in JS using `compareBuilds` (`apps/api/src/services/versionCompare.ts`) against `patches.version` — the target version recorded by the earlier pending upsert, stable across scans via `fillIfNull`. When the reported installed version meets or exceeds that target, flip the `device_patches` row to `installed` regardless of sweep coverage. Otherwise, keep the existing pending-preserving `CASE` behavior unchanged.

This is a bounded, version-aware reconciliation — one of the two options proposed in the issue — and does not touch the staleness-window alternative or the `agent/` coverage-semantics angle also mentioned there.

## Changes

- `apps/api/src/routes/agents/patches.ts`: `upsertInstalledPatches` now computes `installedMeetsTarget` via `compareBuilds(version, patch.version)` and uses it to choose between the plain `'installed'` literal and the existing pending-preserving `CASE` SQL for both `status` and `installedAt`.
- `apps/api/src/routes/agents/patches.test.ts`:
  - Updated the existing "keeps a pending row pending" test to use an explicit, distinct target version (mock now supports overriding the returned `patches.version` independent of the current scan's reported version) so it exercises the real "older version" scenario rather than an accidental equal-version quirk.
  - Added a **flip-without-sweep** test: reported version meets/exceeds target → flips to `installed` via the plain literal even though the row was `pending` and no sweep ran.
  - Added an **older-version-stays-pending** case (the updated existing test) plus a **no-target-on-record** case (`patches.version` is `null`) that falls back to the preserving `CASE`.

## Test plan

- [x] `cd apps/api && npx vitest run src/routes/agents/patches.test.ts` — 31/31 passed
- [x] Confirmed red-first: reverted the implementation change only, reran — the new flip-without-sweep test failed as expected; restored and reran green
- [x] `cd apps/api && npx vitest run src/routes/agents/patches.test.ts src/routes/agents/patchIngestIdentity.test.ts src/services/versionCompare.test.ts` — 72/72 passed
- [x] `cd apps/api && npx tsc --noEmit` — clean
- [x] `npx eslint apps/api/src/routes/agents/patches.ts apps/api/src/routes/agents/patches.test.ts` — clean
- [ ] `patches.integration.test.ts` (real Postgres, `it.runIf(!!process.env.DATABASE_URL)`) not run in this sandbox (no DB); manually traced its version numbers against the new logic — behavior is unchanged for that test's scenarios (older-version preserve, then missing→installed heal both land on the same outcome the fix produces).

Closes #2736

### DIFF
--- a/apps/api/src/routes/agents.test.ts
+++ b/apps/api/src/routes/agents.test.ts
@@ -1447,6 +1447,17 @@ describe('agent routes', () => {
             where: vi.fn().mockResolvedValue(undefined)
           })
         }),
+        // No existing device_patches row for this device+patch — the
+        // installed-path version-aware flip (#2736) falls back to the global
+        // patches.version, which this fixture (installedAt-null handling) does
+        // not otherwise exercise.
+        select: vi.fn().mockReturnValue({
+          from: vi.fn().mockReturnValue({
+            where: vi.fn().mockReturnValue({
+              limit: vi.fn().mockResolvedValue([])
+            })
+          })
+        }),
         insert: vi.fn()
           .mockReturnValueOnce({
             values: pendingInsertValues
--- a/apps/api/src/routes/agents/patches.test.ts
+++ b/apps/api/src/routes/agents/patches.test.ts
@@ -158,6 +158,16 @@ function mockPatchInsertTx() {
         where: vi.fn().mockResolvedValue(undefined),
       })),
     })),
+    // No existing device_patches row in these generic fixtures — the
+    // installed-path version-aware flip (#2736) falls back to the global
+    // patches.version, which these tests don't exercise directly.
+    select: vi.fn(() => ({
+      from: vi.fn(() => ({
+        where: vi.fn(() => ({
+          limit: vi.fn().mockResolvedValue([]),
+        })),
+      })),
+    })),
     insert: vi.fn((table) => ({
       values: vi.fn((values) => ({
         onConflictDoUpdate: vi.fn(() => {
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Review run:** \`/pr-review-toolkit:review-pr\` — code-reviewer + pr-test-analyzer (Sonnet subagents)

**Findings:** 2 substantive findings raised → both addressed in `e2f369301`; 0 outstanding.
- **Critical (code-reviewer):** the original comparison (`a4cd6779e`) used the global, un-tenanted `patches.version` (frozen once via `fillIfNull`, potentially by a *different* device/org that reported the same `(source, externalId)` first) as the flip target. Confirmed against `patchApprovalEvaluator.ts:362`, which already establishes `COALESCE(devicePatches.availableVersion, patches.version)` as the correct per-device-first precedence for exactly this reason. Fixed by fetching this device's own `device_patches.available_version` and using it as the comparison target, falling back to the global column only when no per-device row exists yet.
- **Important (pr-test-analyzer):** one test exercised an unreachable state (`patches.version === null` while the same-call reported version was non-null — impossible given `fillIfNull`'s `COALESCE` semantics in the same upsert). Replaced with a reachable one (installed submit omitting `version`, which normalizes to `null` via the schema) and added dedicated tests for the per-device-precedence-over-stale-global-version scenario and the no-existing-row fallback.
- Suggestions (both agents) not addressed: `versionCompare`'s segment parser silently maps non-numeric tokens to `0`, and one `installedAt` assertion could be tightened further — pre-existing, low-risk, out of scope for this fix.

**Tests:** `apps/api/src/routes/agents/patches.test.ts` — 33/33 passed (`cd apps/api && npx vitest run src/routes/agents/patches.test.ts`); also ran alongside `versionCompare.test.ts` + `patchIngestIdentity.test.ts` — 74/74 passed. Confirmed red-first twice: reverting the implementation-only change (keeping tests) reproduced the exact failures the new tests target, both before and after the critical-finding fix. `npx tsc --noEmit` clean; `eslint` on both changed files clean. `patches.integration.test.ts` (real Postgres, `it.runIf(!!process.env.DATABASE_URL)`) not run in this sandbox — traced its version numbers by hand against the corrected logic; both its scenarios (older-version-stays-pending, then missing→installed heal) land on the same outcomes.

ToddHebebrand:
Fixed the Test API red (run 33790339315): `src/routes/agents.test.ts`'s installed-patches tx mock lacked `select`, which `upsertInstalledPatches`'s new per-device `available_version` lookup now calls — mock gap, not a behavioral regression, fixed in `c639993`; local run of both `agents.test.ts` (28/28) and `patches.test.ts` (33/33) green, `tsc --noEmit` clean.

PR #4844, opened 2026-09-03
fix(ops): default M365 JWK secret file sources to a placeholder file, not /dev/null (#2991)

### LINKED ISSUE #2991: docker compose up fails: 'remount-ro ... operation not permitted' on /dev/null-backed optional M365 signing-key secrets
Reported by **affinity_51941** in #bugs (Discord, message 1532705171576979516).

**Report (PII-redacted):**

when im trying to install using docker compose i keep getting this errorError response from daemon: remount-ro /var/lib/docker/rootfs/overlayfs/292fbc27a8737e5ecd600bcec1dd3dc5b4d03a76255fad155b75c4b9b303ddf6/run/secrets/m365_graph_actions_executor_signing_private_jwk, flags: 0x5021: operation not permitted

Any idea what this is

## Triage analysis

The reported daemon error (`remount-ro .../run/secrets/m365_graph_actions_executor_signing_private_jwk, flags: 0x5021: operation not permitted`) points directly at the file-backed Docker secret Breeze wires up for the optional M365 Graph-actions executor signing key. In docker-compose.yml the `api` service mounts three optional M365 signing-key secrets (graph-read, graph-actions, comms) via Compose's standalone (non-swarm) file-secret mechanism, which bind-mounts the host source file read-only into the container's overlay rootfs. When the corresponding `*_SOURCE_FILE` env var is left unset (the documented default — these features are described as 'off by default and safe to leave unset'), the secret's `file:` falls back to `/dev/null` (docker-compose.yml lines 559-564, mirrored in deploy/docker-compose.prod.yml lines 477-480). Docker then has to bind-mount the special character device `/dev/null` into the container and remount it read-only for the secret; on a number of host kernel/storage-driver combinations that remount-to-read-only of a device-node bind mount inside overlayfs is rejected by the kernel with EPERM, producing exactly the 'remount-ro ... operation not permitted' error the user hit. So a user who never touched the M365 write-actions feature can still fail `docker compose up` purely because of this `/dev/null` fallback default, contradicting the documented 'safe to leave unset' behavior.

## Affected code

- `docker-compose.yml:553-564` — Defines m365_graph_read/actions/comms_executor_signing_private_jwk secrets, defaulting file: to /dev/null when the *_SOURCE_FILE env var is unset.
- `docker-compose.yml:358-365` — api service mounts these three secrets (including m365_graph_actions_executor_signing_private_jwk) unconditionally regardless of whether M365_GRAPH_ACTIONS_TOOLS_ENABLED is true.
- `deploy/docker-compose.prod.yml:471-480` — Same /dev/null default pattern repeated in the production compose file.
- `.env.example:165-184` — Docs the feature as off-by-default and the SOURCE_FILE var as normally empty, i.e. the /dev/null path is the expected/common configuration, not an edge case.

## Potential fix

Avoid bind-mounting /dev/null as a Docker secret source when the M365 signing-key features are disabled. Options: (1) only declare/attach the secret in the compose file when the corresponding *_TOOLS_ENABLED flag is true (e.g. via a profile or a generated empty regular file instead of /dev/null), or (2) default the source to an empty regular placeholder file created at deploy time instead of the /dev/null character device, since regular files can be remounted read-only without the overlayfs/device-node permission issue.

ToddHebebrand:
## Summary

`docker-compose.yml` and `deploy/docker-compose.prod.yml` default the three optional M365 executor signing-key secrets' `file:` source to `/dev/null` when the corresponding `*_SOURCE_FILE` env var is left unset — the common case, since `.env.example` documents these features as off-by-default and safe to leave unset.

Docker Compose's standalone (non-swarm) file-secret support bind-mounts that source into the container's overlay rootfs and remounts it read-only. Doing that to the `/dev/null` character device is rejected by the kernel with `EPERM` on some host kernel/storage-driver combinations, producing exactly the reported error:

```
Error response from daemon: remount-ro /var/lib/docker/rootfs/overlayfs/.../run/secrets/m365_graph_actions_executor_signing_private_jwk, flags: 0x5021: operation not permitted
```

So a user who never touched the M365 write-actions feature could fail `docker compose up` purely because of this default.

## Fix

- Added `docker/secrets/.empty-jwk`, a tracked, always-empty **regular** file, plus `docker/secrets/README.md` explaining why it exists.
- Changed the three `${..._SOURCE_FILE:-/dev/null}` defaults in `docker-compose.yml` to `${..._SOURCE_FILE:-./docker/secrets/.empty-jwk}`, and the two present in `deploy/docker-compose.prod.yml` to `${..._SOURCE_FILE:-../docker/secrets/.empty-jwk}` (path relative to each compose file's own directory — verified with `docker compose config` against both `.env.example` and `deploy/compose-config-test.env`, and with a standalone smoke-test compose file confirming the empty regular file mounts and reads back as a 0-byte secret without error).
- Regular files have no remount-to-read-only restriction, so this avoids the failure while keeping the secret mount present (required by `scripts/check-supply-chain-hardening.sh`) and functionally equivalent — a zero-length "key" the API's config validator still correctly treats as absent.

## Also checked

- `scripts/security/check-supply-chain-hardening.sh` — only asserts the secret mount *lines* exist (`m365_graph_read_executor_signing_private_jwk:` etc.), never inspects the default value, so it needed no change. Confirmed it still passes.
- Guided-setup templates — there isn't a separate template; `scripts/guided-setup.sh` downloads a specific **release tag's** `docker-compose.yml` directly. Per repo precedent (guided-setup is pinned to release tags), **this fix will not reach guided-setup installs until it ships in a release** — flagging per the standing note, not a follow-up needed in this PR.

- `deploy/docker-compose.prod.yml` only defines 2 of the 3 M365 JWK secrets (no `m365_comms_executor_signing_private_jwk` — the comms executor isn't deployed to prod, only in the self-hosted `docker-compose.yml`). That asymmetry predates this PR and is unrelated to the fix; both files' existing secrets got the same placeholder-default treatment.

## Review fixes

- silent-failure-hunter: the test's default-value extractor only recognized `${VAR:-default}` and silently skipped any other `$`-containing form (e.g. `${VAR-default}`, `${VAR:=x}`) as "no default to check" — a future secret using one of those forms could reintroduce the `/dev/null` bug undetected. Replaced with `classifyFileValue()`, which recognizes both default-providing forms and both required/no-default forms, and now **fails the test** on anything else instead of skipping it.
- code-reviewer: `check-m365-graph-read-runtime.sh` / `check-m365-graph-actions-runtime.sh` each had a comment claiming production defaults to `/dev/null` — stale after this fix. Updated both to reference `docker/secrets/.empty-jwk`.

## Tests

- Added `apps/api/src/config/composeSecretFileDefaults.test.ts`: asserts every `secrets.*.file:` default in the tracked Compose files resolves to an existing **regular** file (fails on missing files, directories, or device nodes like `/dev/null`). Verified red-first (temporarily restored the old `/dev/null` defaults — the new test failed with a clear message) then green after restoring the fix.
- `pnpm exec vitest run` (via `npx vitest run` in `apps/api`) on `composeSecretFileDefaults.test.ts`, `composeBindMounts.test.ts`, `envComposeParity.test.ts`, `ciSuccessGatingContract.test.ts` — all green (45/45, then 40/40 after a lint/typecheck fix).
- `npx eslint` on the new test file — clean.
- `docker compose config` against both compose files with test env files — resolves the secret `file:` to the new placeholder correctly in both.
- `scripts/security/check-supply-chain-hardening.sh` — passes.

Closes #2991

### DIFF
--- a/apps/api/src/config/composeSecretFileDefaults.test.ts
+++ b/apps/api/src/config/composeSecretFileDefaults.test.ts
@@ -97,12 +97,32 @@ function collectSecretFileEntries(composeFile: string): SecretFileEntry[] {
   return entries;
 }
 
-/** `${VAR:-default}` -> "default"; a literal (non-interpolated) path is returned as-is. */
-function extractDefault(rawFile: string): string | null {
-  const match = /^\$\{[A-Za-z_][A-Za-z0-9_]*:-(.*)\}$/.exec(rawFile);
-  if (match) return match[1] ?? ''; // may be '' if the author wrote `${VAR:-}` (no default)
-  if (rawFile.includes('$')) return null; // `${VAR}` / `${VAR:?msg}` — no static default to check
-  return rawFile; // literal path, no interpolation at all
+type FileValue =
+  | { kind: 'default'; value: string } // `${VAR:-default}` / `${VAR-default}`, or a literal path
+  | { kind: 'no-default' } // `${VAR}` / `${VAR:?msg}` / `${VAR?msg}` — required, no static default
+  | { kind: 'unrecognized' }; // some other `$`-containing form this parser doesn't know
+
+/**
+ * Classifies a `secrets.<name>.file:` value. Deliberately fails CLOSED: any
+ * `$`-containing form that isn't one of the two recognized shapes above comes
+ * back `unrecognized` rather than being silently skipped, so a typo'd
+ * interpolation (e.g. `${VAR-default}` misread, or a future default-setting
+ * syntax this parser doesn't know) surfaces as a test failure telling a human
+ * to look at it, instead of quietly passing an unchecked default through —
+ * which is exactly how #2991's `/dev/null` default could reappear unnoticed.
+ */
+function classifyFileValue(rawFile: string): FileValue {
+  // `${VAR:-default}` (unset-or-empty) or `${VAR-default}` (unset-only) — both
+  // are real default-providing forms.
+  const withDefault = /^\$\{[A-Za-z_][A-Za-z0-9_]*:?-(.*)\}$/.exec(rawFile);
+  if (withDefault) return { kind: 'default', value: withDefault[1] ?? '' };
+
+  // `${VAR}` (plain) or `${VAR:?msg}` / `${VAR?msg}` (required, fails if
+  // unset/empty) — no static default to check either way.
+  if (/^\$\{[A-Za-z_][A-Za-z0-9_]*(:?\?.*)?\}$/.test(rawFile)) return { kind: 'no-default' };
+
+  if (rawFile.includes('$')) return { kind: 'unrecognized' };
+  return { kind: 'default', value: rawFile }; // literal path, no interpolation at all
 }
 
 const composeFiles = trackedComposeFiles();
@@ -121,8 +141,20 @@ describe('Compose secret file: defaults', () => {
     const problems: string[] = [];
 
     for (const entry of allSecretFileEntries) {
-      const defaultValue = extractDefault(entry.rawFile);
-      if (defaultValue === null || defaultValue === '') continue; // no static default to check
+      const classified = classifyFileValue(entry.rawFile);
+
+      if (classified.kind === 'unrecognized') {
+        problems.push(
+          `${entry.composeFile}: secret "${entry.secretName}" file: value "${entry.rawFile}" ` +
+            `uses an interpolation form this test doesn't recognize as either a static default ` +
+            `or a required (no-default) reference. Update classifyFileValue() in this test to ` +
+            `handle it explicitly, so its default (if any) gets checked rather than skipped.`,
+        );
+        continue;
+      }
+      if (classified.kind === 'no-default') continue; // required var — no static default to check
+      const defaultValue = classified.value;
+      if (defaultValue === '') continue; // `${VAR:-}` — author wrote no default
 
       const composeDir = path.dirname(path.join(REPO_ROOT, entry.composeFile));
       const resolved = defaultValue.startsWith('/')
@@ -158,8 +190,8 @@ describe('Compose secret file: defaults', () => {
     // host made a device-node remount succeed and the generic isFile() check
     // above stopped firing for it.
     const devNullDefaults = allSecretFileEntries
-      .map((e) => ({ ...e, defaultValue: extractDefault(e.rawFile) }))
-      .filter((e) => e.defaultValue === '/dev/null');
+      .map((e) => ({ ...e, classified: classifyFileValue(e.rawFile) }))
+      .filter((e) => e.classified.kind === 'default' && e.classified.value === '/dev/null');
 
     expect(devNullDefaults).toEqual([]);
   });
--- a/apps/api/src/config/composeSecretFileDefaults.test.ts
+++ b/apps/api/src/config/composeSecretFileDefaults.test.ts
@@ -1,0 +1,166 @@
+import { execFileSync } from 'node:child_process';
+import { existsSync, readFileSync, statSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { CORE_SCHEMA, defineMappingTag, defineScalarTag, defineSequenceTag, load } from 'js-yaml';
+import { describe, expect, it } from 'vitest';
+
+// apps/api/src/config -> repo root is 4 levels up (same as composeBindMounts.test.ts).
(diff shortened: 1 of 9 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit (code-reviewer, silent-failure-hunter, pr-test-analyzer), Sonnet, parallel
**Findings:** 2 real findings raised → both addressed in 3eef602a9; 0 outstanding.
- silent-failure-hunter (medium): the test's `${VAR:-default}`-only extractor silently skipped other `$`-containing interpolation forms as "no default to check," which could let a future default reintroduce the `/dev/null` bug undetected. Fixed: `classifyFileValue()` now recognizes both default-providing forms (`:-`/`-`) and both required/no-default forms (`${VAR}`, `:?`/`?`), and **fails the test** on anything unrecognized instead of skipping it.
- code-reviewer (suggestion): `check-m365-graph-read-runtime.sh` / `check-m365-graph-actions-runtime.sh` each had a comment claiming prod defaults to `/dev/null` — stale after this fix. Updated both to reference the new `docker/secrets/.empty-jwk` placeholder.
- pr-test-analyzer: no changes requested — confirmed the static YAML-invariant test is the right guard for this bug class (a live remount smoke test wouldn't reproduce the host-specific EPERM on the CI runner anyway, per the existing `smoke-test` job which already runs with the vulnerable default unset and didn't catch #2991), and confirmed the prod-file's 2-vs-3-secret asymmetry with the self-hosted compose file predates this PR and is unrelated.
**Tests:** `apps/api` affected suite green (`composeSecretFileDefaults.test.ts` 3/3, plus `composeBindMounts.test.ts` + `envComposeParity.test.ts` + `ciSuccessGatingContract.test.ts` 40/40 total); verified red-first (temporarily restored `/dev/null` defaults → test failed with a clear message) then green after restoring the fix. `tsc --noEmit` clean. `eslint` clean. `scripts/security/check-supply-chain-hardening.sh` passes. `docker compose config` resolves the new secret default correctly in both compose files against their respective test env files.

ToddHebebrand:
**Fixed the Guided Setup Smoke failure (run 33790143725).**

Root cause: `scripts/guided-setup.sh`'s WORK_DIR is never a full repo checkout — it (and `scripts/smoke-guided-setup.sh`'s staging) only ever place `docker-compose.yml` + `.env.example` (+ `docker/Caddyfile.prod` in packaged-Caddy mode) into it. The new `./docker/secrets/.empty-jwk` default this PR introduced doesn't exist there, so `docker compose up` failed with `bind source path does not exist` — a harder, guaranteed failure than the original `/dev/null` bug, and worse than before this PR for that install lane.

Reproduced locally against this branch with the exact staging the smoke job does:
```
docker compose --env-file .env -f docker-compose.yml up --no-deps -d api
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: .../docker/secrets/.empty-jwk
```
Confirmed the fix resolves it (same repro, placeholder created → container starts).

**Fix (pushed to this branch, `406c850de`):**
- `scripts/guided-setup.sh`: new `ensure_m365_jwk_secret_placeholder()` step (runs right after `prepare_templates`, before the stack starts) creates `${WORK_DIR}/docker/secrets/.empty-jwk` **locally** rather than downloading it as a release-pinned template. Downloading it would hard-fail the installer for *every* release before this fix ships (older releases' `docker-compose.yml` never references this path — still defaults straight to `/dev/null`, so creating an unused empty file for them is a harmless no-op). Verified in an extracted standalone harness: creates the file in a bare dir, is idempotent/self-healing on rerun, and fails closed with a clear message if the path is unexpectedly a directory.
- `scripts/smoke-guided-setup.sh`: added an explicit assertion that the installer created the placeholder, so a future regression here fails with a pinpointed message instead of a generic "api container never started."
- `scripts/prod/deploy.sh` (strict-production lane): this script already runs from a full checkout by construction (`REPO_ROOT` is derived from the script's own location, and it shells out to `pnpm db:migrate` from `REPO_ROOT` before `compose up` — both require the full monorepo), so `../docker/secrets/.empty-jwk` was never actually missing there. Added a fail-fast preflight check anyway (not a repair — a missing file there would mean the checkout itself is broken/sparse) for defense in depth and parity with the guided-setup lane.

Verified: `bash -n` + `shellcheck` clean on all three scripts; `scripts/prod/remote-access-cutover.test.sh` (deploy.sh's own library unit test) still passes; `scripts/security/check-supply-chain-hardening.sh` still passes. Not re-running the full `guided-setup-smoke` CI job manually per instructions — it will run on the next push-triggered CI cycle.

PR #4845, opened 2026-09-03
fix(api): three silent-failure paths in patch dispatch and desktop finalization (#3945)

### LINKED ISSUE #3945: Three silent-failure paths in patch job dispatch and desktop finalization (follow-ups from #3912)
Deliberately scoped out of PR #3912 to keep a three-critical fix reviewable. All three verified present on `main` at the time of filing.

**1. `routes/configurationPolicies/patchJobs.ts:249` — fire-and-forget enqueue reports success.**
`enqueuePatchJob(...).catch(err => console.error(...))`. The route returns 200 with `createdJobs` while nothing was queued. #3912 made `enqueuePatchJob` throw on a wedged queue id (previously it swallowed), so this promise can now reject — and #3912's fan-out fix does **not** cover this path, because the route calls `enqueuePatchJob`, not `processExecutePatchJob`. The reconcile sweep now deliberately skips a wedged id, so there is no backstop either.

**2. A never-resolving `stop_pending` reports once, then goes permanently silent.**
`drivePersistedIntent` re-enqueues on `stop_pending` with the same `stableJobId`. The failed job is retained by `removeOnFail: { count: 250 }`, so `queue.add({ jobId })` is a BullMQ no-op — no new job, no new attempts, no further events — while `enqueueDesktopSessionFinalization` still returns `{ acknowledged: true, jobId }`. The orphan scanner then returns `'retained'` with zero logging, every 30 s, forever. An agent that goes offline permanently mid-session yields one warning and then nothing, while the `remote_sessions` row stays non-terminal.

The durable fix belongs in the orphan scanner — escalate when an intent has been `retained` beyond some age — not in the BullMQ job.

**3. `services/desktopSessionOrphanRecovery.ts:158` — bare `catch { return 'retained'; }`** with no logging at all, swallowing any parse/canonicalise failure.

Related: #3912, and CLAUDE.md's no-silent-failures convention.

ToddHebebrand:
## Summary

Closes #3945. Three silent-failure fixes deliberately deferred from #3912 to keep that PR reviewable. Each is a separate commit with a failing test written first.

1. **`routes/configurationPolicies/patchJobs.ts`** — the fire-and-forget `enqueuePatchJob(...).catch(console.error)` returned 201/`success:true` with `createdJobs` populated even when nothing was queued (e.g. a wedged BullMQ job id that #3912 made `enqueuePatchJob` throw on). The enqueue is now awaited per job; failures are logged, reported to Sentry, and collected into an `enqueueFailures` array in both the audit-log details and the response body. `success` is now `false` whenever any created job failed to enqueue.

2. **`services/desktopSessionOrphanRecovery.ts`** — a `stop_pending` finalization intent re-adds the same stable BullMQ jobId every scan; once that job hash is retained in a terminal state, `queue.add({ jobId })` is a silent no-op with no further BullMQ-side signal, so an agent that goes offline permanently produced one warning and then total silence forever. The fix is in the scanner (not the BullMQ job, which can't un-wedge itself): track how long a given `finalizationId` has sat in `stop_pending` and escalate (log + Sentry) once past a 10-minute threshold, once per episode — same shape as `reportedWedgedJobIds` in `jobs/patchJobExecutor.ts`.

3. **`services/desktopSessionOrphanRecovery.ts`** — the bare `catch { return 'retained'; }` around the persisted-intent JSON.parse/canonicalize swallowed any parse failure with zero logging. Fail-closed behavior (retain, never reclaim) is unchanged; the catch now logs with session/finalization context and reports to Sentry, matching sibling reconcile paths.

## Test plan
- [x] `apps/api/src/routes/configurationPolicies/patchJobs.test.ts` — 28/28 passing (new: enqueue-failure surfacing test, written failing first)
- [x] `apps/api/src/services/desktopSessionOrphanRecovery.test.ts` — 14/14 passing (new: stop_pending escalation test + malformed-intent logging test, both written failing first)
- [x] `apps/api/src/jobs/patchJobExecutor.test.ts`, `apps/api/src/jobs/desktopSessionFinalizationWorker.test.ts`, `apps/api/src/services/desktopSessionFinalization.test.ts` — 71/71 passing (sibling suites, unaffected)
- [x] `npx tsc --noEmit` clean on `apps/api`

### DIFF
--- a/apps/api/src/routes/configurationPolicies/patchJobs.test.ts
+++ b/apps/api/src/routes/configurationPolicies/patchJobs.test.ts
@@ -22,7 +22,7 @@ const {
     needsRepair: rows.filter((row) => row.effectiveStatus === 'needs_repair').length,
     invalidReference: rows.filter((row) => row.effectiveStatus === 'invalid_reference').length,
   })),
-  enqueuePatchJobMock: vi.fn(async () => undefined),
+  enqueuePatchJobMock: vi.fn(async (_jobId?: string, _delayMs?: number) => undefined),
   captureExceptionMock: vi.fn(),
 }));
 
@@ -362,6 +362,7 @@ describe('configurationPolicies patchJob routes', () => {
       // response must not claim success (#3945: a fire-and-forget
       // `.catch(console.error)` previously returned 201/success:true here
       // with zero visibility that the job would never execute).
+      expect(res.status).toBe(201);
       const json = await res.json();
       expect(json.success).toBe(false);
       expect(json.enqueueFailures).toEqual([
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** 3 parallel Sonnet agents via pr-review-toolkit (code-reviewer, pr-test-analyzer, silent-failure-hunter)
**Findings:** 2 Important (both converged on the same root cause), plus test-coverage gaps → all addressed in e472a787d; 0 outstanding.

- **Important (code-reviewer + silent-failure-hunter, independently converged):** the stop_pending escalation's clock was a purely in-memory "first observed" timestamp, so a deploy/restart mid-stall silently reset the 10-minute escalation window — undermining the exact scenario ("agent never comes back across a deploy") the feature exists to catch. Fixed: age is now derived from the persisted finalization intent's own `endedAt` (restart-stable), not local first-observation.
- **Important (code-reviewer):** the "already reported" map was only cleared on a resolution path (`drivePersistedIntent` seeing the intent leave `stop_pending`) that rarely fires in practice — per `jobs/desktopSessionFinalizationWorker.ts`, the BullMQ worker resolves most intents directly and faster, so the session usually goes terminal and drops out of the scan before that branch runs. The map would grow for the process lifetime, contradicting its own doc comment. Fixed: it's now a self-pruning `Map` bounded by age, independent of whether the resolution branch ever fires.
- **Suggestion (silent-failure-hunter):** named the parse-failure exception (`PersistedFinalizationIntentParseError`) instead of wrapping in a generic `Error`, matching the BREEZE-1J pattern already used by sibling code in the same file.
- **Test gaps (pr-test-analyzer + code-reviewer):** added a multi-org partial-enqueue-failure test for `patchJobs.ts` (partner-wide request where one org's enqueue fails and another succeeds), an exact-threshold escalation boundary test, a restart-survival test, and a test proving a new finalization episode on the same session escalates independently after a prior one resolved.
- **Not a bug (verified by both code-reviewer and silent-failure-hunter):** `success:false` at HTTP 201 is intentional and consistent with `runAction`'s `success` field check independent of status code; `captureException` call sites match existing repo precedent (route handler passes `c`, background scanner omits it).

**Tests:** `apps/api/src/routes/configurationPolicies/patchJobs.test.ts` (29/29), `apps/api/src/services/desktopSessionOrphanRecovery.test.ts` (17/17), plus sibling suites `patchJobExecutor.test.ts`, `desktopSessionFinalizationWorker.test.ts`, `desktopSessionFinalization.test.ts` (117/117 total across all five files). `npx tsc --noEmit` clean on `apps/api`.

PR #4846, opened 2026-09-03
fix(api): honour partner-level config-policy assignments in warranty + event-log retention (#3963)

### LINKED ISSUE #3963: Partner-level config-policy assignments silently ignored by warranty evaluator and event-log retention reader
Two config-policy readers silently ignore `level='partner'` assignments, so a partner-wide ("all organizations") policy appears saved in the UI and never applies. Both verified by reading `origin/main`, not inferred. Same root shape as #3954 (fixed in #3962) — this is the third and fourth instance, which is why they're filed together rather than as point fixes.

**1. Partner-wide warranty policies never apply.** `apps/api/src/services/warrantyAlertEvaluator.ts:72`:

```ts
// Priority: device > device_group > site > organization > partner (closest wins)
const targetIds = [deviceId, ...groupIds, device.siteId, device.orgId].filter(Boolean) as string[];
```

The comment directly above ranks `partner` as a level. The array has no partner id, so a `level='partner'` assignment can never match. Comment and code disagree, and the comment documents the intent.

**2. Partner-level event-log retention silently falls back to 30 days.** `apps/api/src/routes/agents/helpers.ts:1961-1965` in `getOrgEventLogRetentionDays`:

```ts
.where(and(
  eq(configPolicyAssignments.level, 'organization'),
  eq(configPolicyAssignments.targetId, orgId),
  eq(configurationPolicies.status, 'active'),
))
...
return row?.retentionDays ?? 30;
```

`level='organization'` only. A partner-level event-log assignment matches nothing and the `?? 30` default takes over — so an MSP that sets fleet-wide retention gets 30 days everywhere with no error and no log line.

**Why this keeps happening:** per `CLAUDE.md` (Partner-Wide First), a partner-wide row carries `org_id NULL` + `partner_id`, so any reader written as an org-axis-only predicate returns zero rows rather than failing. It is a silent no-op by construction. `#3954` was the same bug in the update-ring device count; #3962's call-site sweep found these two while verifying it.

**Fix shape** (mirror #3962's `resolveRingAssignedDeviceIds`):
- Warranty: append the device's org's `partner_id` to `targetIds`, and keep the documented precedence walk intact so `partner` remains lowest priority.
- Event log: add a partner-level branch resolved through `organizations.partner_id`, ordered so an org-level assignment still wins.
- Gate any app-layer dual-axis read on `auth.scope === 'partner'` — org tokens carry a `partnerId` but never pass `breeze_has_partner_access`.

**Tests must not be Drizzle-mocked.** Both bugs live entirely in a `.where()` clause, so a mock that ignores `.where()` cannot catch them. Assert compiled SQL via `PgDialect`, or add a real-Postgres integration test proving the fan-out fires (the pattern #3962 used).

**Worth sweeping the rest** while in here — #3962 verified these readers as correct and left them alone: `featureConfigResolver.ts`, `automationWorker.ts`, `policyEvaluationService.ts`, `patchSchedulerWorker.ts`, `partnerApi/configuration.ts`, and the device→policy-direction readers in `routes/agents/helpers.ts` (correct by construction — they derive the device's org's partner).

ToddHebebrand:
Closes #3963

Two config-policy readers silently ignored `level='partner'` assignments, so a partner-wide ("All organizations") policy saved fine in the UI and then never applied. Third and fourth instances of the shape fixed in #3962 (#3954).

## The two axes

A partner-wide policy differs from an org policy on **two** independent axes, and each reader was broken on at least one:

1. **ASSIGNMENT** — `config_policy_assignments.target_id` is polymorphic: its referent depends on `level` (`'device'` → `devices.id` … `'partner'` → **`partners.id`**). A reader that only ever compares `target_id` against org/site/device ids can never match a `level='partner'` row.
2. **OWNERSHIP** — `configuration_policies` is org-owned (`org_id` set) XOR partner-owned (`org_id NULL`, `partner_id` set, #1724). `policyOwnershipCondition` (#2930) is what admits the second shape.

Both fail as **zero rows**, not as an error — silent by construction (CLAUDE.md, "Partner-Wide First").

## Changes

**`apps/api/src/services/warrantyAlertEvaluator.ts`** — `resolveWarrantySettings` built one `inArray(targetId, [deviceId, ...groupIds, siteId, orgId])` bag with no partner id in it, despite the comment directly above ranking `partner` as a level. Replaced with level-**paired** target conditions (mirroring `resolveDeviceEventLogSettings`, the canonical hierarchy resolver in the same codebase), including a `level='partner'` branch resolved through the device org's `partner_id`. Precedence is unchanged — partner stays lowest.

Also adds `policyOwnershipCondition`: warranty was the one hierarchy resolver that never got the #2930 predicate. That is a second, deliberate change, called out here so it can be judged on its own — it is a strict narrowing that drops policies owned by neither the device's org nor its partner (previously the query had *no* ownership predicate at all). A regression test pins that the pre-existing working case — a partner-owned policy assigned at `organization` level — still resolves.

**`apps/api/src/routes/agents/helpers.ts`** — `getOrgEventLogRetentionDays` filtered `level='organization'` only, so a partner-level event-log assignment matched nothing and `?? 30` took over: an MSP that set fleet-wide retention got 30 days on every org, with no error and no log line. Now resolves across both levels with org winning, and falls back to `EVENT_LOG_DEFAULTS.retentionDays` rather than a bare `30`. The org-only case is behaviourally unchanged: the previous `.orderBy(priority).limit(1)` is the same as the new level-DESC-then-priority-ASC sort when every row is org-level.

## Execution contexts (verified against the code, not assumed)

| Reader | Context | RLS relevance |
|---|---|---|
| `getOrgEventLogRetentionDays` | **System** — `jobs/eventLogRetention.ts` wraps each per-org call in `runOutsideDbContext(withSystemDbAccessContext(...))` | RLS bypassed; the app-layer `WHERE` was the only gate, so this was a pure predicate bug |
| `evaluateWarrantyAlerts` (worker) | **System** — `services/warrantyWorker.ts` → `warrantySync.ts` | same |
| `evaluateWarrantyAlerts` (agent) | **Org-scoped** — `PUT /agents/:id/warranty-info` → `upsertAgentWarranty`, inheriting `agentAuthMiddleware`'s organization context | RLS-gated. Works because that context carries `currentPartnerId: device.partnerId` (`agentAuth.ts:959`, #4673 W02), which the SELECT-only `<table>_partner_wide_select` policies read. No scope gate or system-context escape was added or needed. |

No schema change, so no migration and no cascade / export-policy registration.

## Tests

New `apps/api/src/__tests__/integration/configPolicyPartnerLevelAssignments.integration.test.ts` (11 cases, real Postgres via the blocking `integration-test` job). A mocked-Drizzle test **cannot** catch this class — both bugs lived entirely inside a `.where()` clause and the repo's chainable stubs ignore `.where()` arguments — so this is real-DB by necessity.

Red-first, verified: on the pre-fix code exactly the three bug-revealing cases fail (warranty system path, warranty agent path, retention partner-wide); the other eight pass before and after, which is correct — they are guards, not discriminators.

Covered: partner-wide resolution under both system and org-scoped agent contexts; a `currentPartnerId: null` negative control proving the RLS branch (not a bypass) is doing the work; org-beats-partner precedence on both readers; cross-partner isolation on both; org-level priority tie-break unchanged; default fallback; and the partner-owned/org-assigned guard described above.

`apps/api/src/services/warrantyAlertEvaluator.test.ts` needed its call-ordered Drizzle stubs re-sequenced for the added `organizations` lookup — assertions unchanged, still 8 tests.

## Sweep

The issue asked for a sweep of the remaining `config_policy_assignments` readers. #3962 had already cleared `featureConfigResolver.ts`, `automationWorker.ts`, `policyEvaluationService.ts`, `patchSchedulerWorker.ts`, `partnerApi/configuration.ts` and the device→policy readers in `routes/agents/helpers.ts`; this PR additionally audited `helperPermissions.ts`, `aiToolsConfigPolicy.ts`, `patchJobService.ts`, `softwareInventory.ts`, `pamBridge.ts` and `configurationPolicy.ts`. **No further instances of this bug class.** Everything else is either CRUD/admin listing, a delegate to an already-verified resolver, or already canonical — `configurationPolicy.ts`'s `resolveEffectiveConfigWithExecutor` has both the partner branch and an inline dual-axis ownership predicate.

One non-blocking consistency note, deliberately not changed here: `services/helperPermissions.ts` has the `level='partner'` branch but no ownership predicate at all. Its absence is *more* permissive, not less, so it cannot produce the silent-drop failure mode this PR fixes — but aligning it with `policyOwnershipCondition` would be a reasonable low-priority follow-up.

## Verification

- New integration suite: 11/11 green.
- Neighbouring integration suites (`warrantyAlertEvaluator`, `agentPolicyResolversPartnerWide`, `configurationPolicyPartnerResolution`): 23/23 green.
- Unit suites for warranty, event-log retention, config-policy ownership, and all ten `routes/agents/helpers.*` files: green.
- `tsc --noEmit` clean; `eslint` clean on the changed files.

### DIFF
--- a/apps/api/src/__tests__/integration/configPolicyPartnerLevelAssignments.integration.test.ts
+++ b/apps/api/src/__tests__/integration/configPolicyPartnerLevelAssignments.integration.test.ts
@@ -105,6 +105,8 @@ function agentBlindContext(orgId: string): DbAccessContext {
 const createdPolicies: string[] = [];
 const createdDevices: string[] = [];
 
+// Children before parents: the alert/warranty rows FK to `devices` with no
+// ON DELETE, so deleting the device first raises 23503.
 afterEach(async () => {
   await withDbAccessContext(SYSTEM_CTX, async () => {
     for (const id of createdDevices) {
@@ -121,7 +123,7 @@ afterEach(async () => {
 });
 
 type PolicyOwner = { orgId: string | null; partnerId: string | null };
-type AssignmentLevel = 'partner' | 'organization';
+type AssignmentLevel = 'partner' | 'organization' | 'site' | 'device_group' | 'device';
 
 /** ISO `YYYY-MM-DD` date `days` from now. */
 function inDays(days: number): string {
@@ -229,6 +231,7 @@ async function assign(configPolicyId: string, level: AssignmentLevel, targetId: 
   });
 }
 
+
 describe('warranty evaluator honours partner-level assignments (#3963)', () => {
   it('applies a partner-wide warranty policy from the SYSTEM-scoped warranty worker', async () => {
     const { partner, org, site } = await seedTenant();
@@ -305,6 +308,48 @@ describe('warranty evaluator honours partner-level assignments (#3963)', () => {
 
     expect(alertId).not.toBeNull();
   });
+
+  // The fix replaced ONE `inArray(targetId, [deviceId, ...groupIds, siteId, orgId])`
+  // with five level-PAIRED conditions. Only the partner branch was broken, but all
+  // five were rewritten, and `targetId` is polymorphic — pairing an id to the wrong
+  // level silently matches nothing rather than erroring. The org branch is covered
+  // by warrantyAlertEvaluator.integration.test.ts and the partner branch by the
+  // cases above; this pins a THIRD, structurally distinct branch (an id that is
+  // neither the device's org nor its partner, resolved through the device row).
+  //
+  // COVERAGE LIMIT, stated rather than implied: the `device` and `device_group`
+  // branches are still not covered here. Seeding them requires an assignment whose
+  // target is a device/group created on the app pool, and that insert is rejected
+  // in this suite by the `config_policy_assignments_target_owner_fk` statement
+  // trigger for reasons that reproduce neither in psql (the identical fixture —
+  // partner/org/site/device/group/membership/policy, then all three assignment
+  // levels — inserts cleanly as breeze_test, and both rows are visible to
+  // breeze_app under system scope) nor in the org/site/partner cases that pass
+  // here. That is a test-fixture problem, not a product one, so it is recorded
+  // instead of being papered over with a weaker assertion.
+  it('matches a site-level warranty assignment to its own level', async () => {
+    const { org, site } = await seedTenant();
+    const device = await seedDeviceWithExpiringWarranty(org.id, site.id);
+    const policyId = await seedWarrantyPolicy({ orgId: org.id, partnerId: null }, true);
+    await assign(policyId, 'site', site.id);
+
+    const alertId = await withDbAccessContext(SYSTEM_CTX, () => evaluateWarrantyAlerts(device.id));
+
+    expect(alertId).not.toBeNull();
+  });
+
+  // NOT TESTED, deliberately: "an assignment targeting this org for a policy owned
+  // by a DIFFERENT org". That row cannot be created. Postgres rejects the insert
+  // via `config_policy_assignments_target_owner_fk`
+  // (`breeze_validate_config_policy_assignment_target`) with "configuration
+  // assignment target is incompatible with the policy owner" — verified by
+  // attempting exactly this fixture and getting 23503 back. So the
+  // `policyOwnershipCondition` this PR adds to the warranty query is
+  // defense-in-depth behind a DB-enforced invariant, not a filter that any
+  // creatable row depends on. The one way ownership and target can still drift is
+  // an org MOVE (`config_policy_assignments` has no `org_id` column and is in no
+  // move/cascade list, so a device- or group-level assignment survives its device
+  // changing org), which the predicate now correctly stops from resolving.
 
   it('does not apply another partner’s partner-wide warranty policy', async () => {
     const { org, site } = await seedTenant();
--- a/apps/api/src/__tests__/integration/configPolicyPartnerLevelAssignments.integration.test.ts
+++ b/apps/api/src/__tests__/integration/configPolicyPartnerLevelAssignments.integration.test.ts
@@ -1,0 +1,385 @@
+/**
+ * Partner-LEVEL config-policy assignments reach the warranty evaluator and the
+ * event-log retention worker (#3963).
+ *
+ * Two axes decide whether an MSP's "all organizations" policy actually applies
+ * to a device or an org, and they are routinely confused:
+ *
+ *   1. OWNERSHIP — `configuration_policies` is org-owned (`org_id` set,
+ *      `partner_id NULL`) or partner-owned (`org_id NULL`, `partner_id` set),
+ *      the XOR from #1724. `policyOwnershipCondition` (#2930) is what admits the
+ *      second shape; a bare `org_id = <device org>` never matches it.
+ *   2. ASSIGNMENT — `config_policy_assignments.target_id` is POLYMORPHIC: its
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three lenses on `ad8213a1f`: `code-reviewer` (tenancy/precedence/cross-tenant focus), `pr-test-analyzer` (discrimination + coverage gaps), `silent-failure-hunter` (this PR fixes a silent zero-row failure, so the fix must not introduce another).

**Findings:** 3 raised → all 3 addressed in `7fc1345eb`; 0 outstanding.

- `code-reviewer`: **0 findings at confidence ≥80.** Independently confirmed precedence is unchanged on both readers, `or(...targetConditions)` can never be empty, and no cross-tenant widening. It also surfaced useful context (not a defect): `config_policy_assignments` has no `org_id` column and is in no move/cascade list, so a device- or group-level assignment survives an org move today — the ownership predicate this PR adds now stops that stale row from resolving.
- `silent-failure-hunter` (MEDIUM): the new `organizations` lookup both readers depend on was unguarded, so an empty result fell through to exactly the org-only resolution this PR exists to fix — one join upstream, silently. **Fixed:** both sites now log + `captureException`, then continue so resolution degrades rather than throws. The state is unreachable by schema (`devices.org_id` NOT NULL + FK; `organizations.partner_id` NOT NULL), which is the reason to make it loud: it can only mean an invariant broke.
- `pr-test-analyzer` (HIGH): the level-paired rewrite touched all five assignment levels, but only org and partner had real-Postgres coverage. **Partially fixed** — added a site-level case (a third structurally distinct branch). See the coverage limit below.

**Two things recorded rather than papered over:**
1. A cross-org ownership mismatch is **not constructible** — Postgres rejects it via `config_policy_assignments_target_owner_fk`. So `policyOwnershipCondition` here is defense-in-depth behind a DB-enforced invariant, not a filter any creatable row depends on. That is the concrete answer to "what stops matching under the narrowing": nothing that can be created through any write path.
2. The `device` and `device_group` branches remain **uncovered**. Seeding them needs an assignment targeting a device/group created on the app pool, and that insert is rejected by the target-owner trigger for reasons that reproduce neither in psql (identical fixture inserts cleanly as `breeze_test`; both rows are visible to `breeze_app` under system scope) nor in the org/site/partner cases that pass here. That is a fixture problem, not a product one — the test file states it rather than asserting coverage that does not exist.

**Tests (local, on `7fc1345eb`):**
- Integration, real Postgres: `configPolicyPartnerLevelAssignments` + `warrantyAlertEvaluator` + `agentPolicyResolversPartnerWide` — **31 passed**.
- Unit: `warrantyAlertEvaluator{,.resolveCas}`, `warrantySync`, `configPolicyOwnership`, `eventLogRetention`, `helpers.eventLogDefaults`, `helpers.partnerWidePolicies` — **64 passed**. Plus all ten `routes/agents/helpers.*` suites green on the first head.
- `tsc --noEmit` clean; `eslint` clean on all four changed files.

**Red-first, verified (not asserted):** on pre-fix code exactly the three bug-revealing cases fail (warranty system path, warranty agent path, retention partner-wide); the guards pass before and after, as they should. The new `!org` unit case was likewise confirmed red with the guard removed.

**Status:** review-clean. **CI pending at time of writing** — the local runs above are the evidence for the fix; the blocking `integration-test` and `Test API` jobs still need to go green before merge. Awaiting maintainer merge; not merged and issue not closed.

Issue #2776, opened 2026-07-25
[Web][API] No partner/org default for enrollment link device count or expiry — maxEnrollmentLinkTtlMinutes never landed


ToddHebebrand:
### Description

There is currently no way for a partner (or an org) to set defaults for enrollment link **device count** or **expiry**. Both are hardcoded server-side and can only be changed per-generated-link in the Add Device modal, or globally by an operator via env var — neither of which helps an MSP that wants "our links are always 30 days / 25 devices" as house policy.

**What exists today.** The Add Device modal exposes per-request pickers only: `deviceCount` (defaults to 1, `AddDeviceModal.tsx:124`) and a fixed "Link expires in" select of 1h/24h/7d/30d/90d/1y (`ttlMinutes`, defaults to 1440, `AddDeviceModal.tsx:133`). `EnrollmentKeyManager.tsx:723-742` similarly offers raw Max Usage + expiry inputs. Every one of those is a fresh choice each time — nothing persists.

**What the server does.** The defaults behind those blank/unset values are:

- Parent enrollment key TTL — 60 min, `routes/enrollmentKeys.ts:86`, env `ENROLLMENT_KEY_DEFAULT_TTL_MINUTES`
- Child key TTL (installer link) — 24 h, `routes/enrollmentKeys.ts:97`, env `CHILD_ENROLLMENT_KEY_TTL_MINUTES`
- Bootstrap token TTL — 24 h, `services/installerBootstrapToken.ts:30`, env `INSTALLER_BOOTSTRAP_TOKEN_TTL_MINUTES`
- `maxUsage` on any key — 1, `routes/enrollmentKeys.ts:684` / `:442`, **no env override at all**

Env vars are the wrong lever for this on hosted: they're per-deployment, so every partner on a region shares one value. Only `ENROLLMENT_KEY_DEFAULT_TTL_MINUTES` is even documented (`.env.example:322`).

**This was designed and then dropped.** `AddDeviceModal.tsx:130-132` carries the comment:

> `// "Never expires" is intentionally omitted until the partner-level cap (maxEnrollmentLinkTtlMinutes) lands in a sibling PR.`

Grepping the repo for `maxEnrollmentLinkTtl` returns that comment and nothing else — the sibling PR never landed, so the modal is permanently missing its "never expires" option waiting on a cap that doesn't exist.

### Proposed Behavior

[...]

PR #4847, opened 2026-09-03
fix(api/rls): make the remote-access launcher's partner read a real escalation (#3419)

### LINKED ISSUE #3419: [API][RLS] Remote-access launcher resolves no providers for org-scoped callers — partner read is invisible under org scope

ToddHebebrand:
Closes #3419

## The defect

`loadRemoteAccessLauncherContext` (`apps/api/src/routes/devices/core.ts`) read `partners.settings` inside a **bare `withSystemDbAccessContext`**. That wrapper looks like a privilege escalation, but it is inert inside a request: `withDbAccessContext` early-returns when a context store already exists (`db/index.ts`), and `authMiddleware` has already opened one — so the read silently retained the **caller's** scope.

`partners`' only SELECT policy is `breeze_has_partner_access(id)` (`migrations/2026-04-11-partners-rls.sql`), and `computeAccessiblePartnerIds` returns `[]` for `scope === 'organization'` (`middleware/auth.ts`). For an org-scoped technician the join therefore returned **zero rows — without raising** — and the launcher reported `no_provider_configured` for a tenant that had a provider configured.

Both entry points are `requireScope('organization', 'partner', 'system')`, so org scope is not a hypothetical caller; it is the default one. User-visible symptoms:

- **GET /devices/:id** → `hasRemoteAccessLauncher: false`, so the *Connect Desktop* button never renders.
- **POST /devices/:id/remote-access-launch** → 404 `no_provider_configured`.

Same defect class as #3391 (`readPreferredProviderId`), where the identical wrapper was inert ceremony and was removed. Here the consequence was worse than cosmetic, because this call genuinely depended on the escalation it appeared to perform.

## The fix

**Reuse, not a second implementation.** `readPartnerRemoteAccessSettings` (`apps/api/src/services/remoteAccessProviders.ts`) already performed exactly this read with the correct `runOutsideDbContext(() => withSystemDbAccessContext(...))` escape — and its own docstring already named this issue:

> `See #3419 for the same gap in the launcher's own partner read.`

So `core.ts` now delegates to it, deleting the duplicated join rather than fixing the same query in two places. That also removes the risk of the two copies drifting on skip-reason semantics — the very thing #3402 fixed for the availability/issuance split.

**Second change (same domain, deliberately in scope):** both reads in `remoteAccessProviders.ts` now go through `readWithPartnerAxisVisibility` (`apps/api/src/db/partnerAxisRead.ts`), the canonical helper for this shape (~15 other call sites). Beyond standardising, this restores the **#2776 round-4 availability skip**: a caller that is *already* system-scoped reads in its own transaction instead of borrowing a **second pooled connection** while the first is still held. That matters more after this PR, because the fix puts this read on `GET /devices/:id` — a hot path — and postgres-js has no acquire timeout.

## Sweep for the same shape

The issue asked for a sweep of other bare `withSystemDbAccessContext` calls on request paths. Verified results:

- **`routes/devices/core.ts` — the fixed call site was the only occurrence.** The one other mention in the file is a comment on the `device_commands` block explaining why that path deliberately does *not* wrap.
- **`partners.settings.remoteAccessProviders` has exactly one other reader, and it was already correct** — `services/remoteAccessProviders.ts`, whose docstring had already flagged `core.ts` as the outstanding sibling. `routes/remote/index.ts` consumes it through that service and projects via `toProviderSummaries`. After this PR there is no remaining broken read of this setting.
- **`services/pamApprovers.ts:39` (`resolveElevationApprovers`) was raised as a same-shape sibling and cleared on inspection.** It does carry a bare `withSystemDbAccessContext` over `partner_users` (a `PARTNER_TENANT_TABLES` member), but its only runtime caller — `routes/agents/elevationRequests.ts:201` — already escalates two lines above with `runOutsideDbContext(() => withSystemDbAccessContext(...))`, so the ambient scope is genuinely system by the time it runs. Not a defect. It is *fragile*, though: a future second caller that did not escalate first would silently drop partner-wide technicians from the PAM approver set.

The general lesson, and why this class keeps recurring: a bare `withSystemDbAccessContext` is neither always-broken nor always-fine. What matters is the **ambient scope at the moment it runs**, which any intermediate `runOutsideDbContext` on the call path changes — so it can only be judged by walking up to every caller, never from the call site alone.

A broader repo-wide sweep produced a longer candidate list in unrelated subsystems, but it is **not trustworthy as it stands and no part of it is claimed here as a defect**. Two of its candidates were spot-checked against the code and **both were false positives**: `services/pamApprovers.ts:39` (documented above) and `routes/partnerApi/provisioning.ts:217`. The remaining candidates were not checked either way and are neither confirmed nor cleared. The recurring error was assuming "inside a request ⇒ an ambient context exists"; that is false wherever an enclosing caller already escalated, and false for the whole partner-API write path, where `middleware/partnerApiAuth.ts:314-316` states outright that "Handlers therefore receive NO ambient DB context and must open their own bounded withDbAccessContext / withSystemDbAccessContext per operation." A bare wrapper there is correct, not inert.

Nothing from that list is folded into this PR, and none of it should be acted on until re-verified caller-by-caller.

## Known trade-off: one extra pooled connection on GET /devices/:id

Worth stating plainly, because the bug and the connection-thrift were **the same line of code**. The old wrapper was inert, so it took no second connection — it was "cheap" precisely because it was broken. A correct escalation cannot be free:

- `readWithPartnerAxisVisibility` exits the ALS store, so the system context does **not** nest — it opens a second transaction on a second pooled connection while the request's own transaction still holds the first.
- `checkRemoteAccessLauncherAvailabilityForDevice` runs on **every** `GET /devices/:id`, including for the majority of tenants who have no launcher providers configured at all — you must read the partner settings to discover that.
- Populations: **system** scope skips the escape (that is what the `readWithPartnerAxisVisibility` swap buys). **Org** scope must take it — that is the bug being fixed. **Partner** scope also takes it, though `breeze_has_partner_access` would already grant it its own partner row.

Assessed as acceptable to ship as-is:

- `DB_POOL_MAX` defaults to **30**, not 10 (`apps/api/src/db/index.ts:23-30`).
- The escaped read is a single indexed one-row join, held for milliseconds.
- `GET /devices/:id` is `requirePermission(DEVICES_READ)` on a human-paced UI path — not an agent/heartbeat firehose. Materially unlike the #2776 case, which was hundreds of concurrent installer bootstraps.
- `routes/remote/index.ts` already takes this exact escape on a user-facing endpoint today and has shipped fine.

**Deliberately not done here:** narrowing the escape so partner-scoped callers skip it too. Doing that correctly needs either an extra org→partner query on the ambient connection, or the assumption `auth.partnerId === device.org.partnerId` — and that assumption is exactly the kind of thing that should not be load-bearing for a cross-tenant read, so the `orgId`-keyed join that derives the partner from the data stays. Deriving the skip from a shared helper also beats reintroducing bespoke escape logic at one call site, which is what `readWithPartnerAxisVisibility` was extracted to prevent. Flagging it as a possible future optimization, not a defect in this PR.

## Verification

`apps/api/src/__tests__/integration/remoteAccessLauncherPartnerVisibility.integration.test.ts` drives the **real** resolver against **real Postgres** from inside an org-scoped `withDbAccessContext` — the same ambient context an org-scoped HTTP request runs under. Modelled on `enrollmentDefaultsPartnerCap.integration.test.ts` (the #2776 precedent the issue pointed at).

**This is the first runtime reproduction of #3419** — the issue was reported, carefully, as a structural defect proven only on paper. Against the pre-fix body the new test fails with exactly the reported symptom:

```
AssertionError: expected { available: false, …(2) } to deeply equal { available: true, …(2) }
-   "available": true,          +   "available": false,
-   "providerId": "rustdesk",   +   "providerId": null,
-   "skipReason": null,         +   "skipReason": "no_provider_configured",
```

Five cases, so the suite can't rot into a vacuous pass:

| Case | What it protects |
|---|---|
| org-scoped resolve | the bug itself (verified red pre-fix) |
| delegated read under org scope | the escape in the service, independently |
| raw `partners` row is invisible to that context | the RLS **premise** — if this ever returns the row, the policy changed and the file needs re-review, not green-lighting |
| unconfigured partner still reports `no_provider_configured` | discriminating negative control — the fix can't be "always available" |
| system-scoped caller reads in its own transaction | the pool-availability skip; **mutation-checked** by disabling the skip branch, which turns it red |

No mocked suite could have caught the original defect: they stage the partner row their own assertions then read back, with no RLS evaluation at all.

**Evidence**

- New integration test — **5 passed** (real Postgres, private test stack).
- `core.remoteAccessLaunch.test.ts` + `remoteAccessProviders.test.ts` — **18 passed**.
- `src/routes/devices/` — **44 files, 592 passed**.
- `src/routes/remote/`, `src/services/remoteAccess*`, `src/db/partnerAxisRead` — **7 files, 114 passed**.
- Every other suite importing `core.ts` while mocking `../../db` (`multi-tenant-isolation`, `core.decommission`, `cascadeDelete`, `systemManaged`, `core.permissions`, `moveOrg*`, `core.list-response-shape`, `ticketOrgMoveLockOrder`) — **9 files, 198 passed**; plus `routes/remote.test.ts` — **21 passed**.
- `enrollmentDefaultsPartnerCap.integration.test.ts` (the precedent suite, to confirm the shared helper is untouched) — **3 passed**.
- `integration-suite-coverage` (#4522 contract) — **5 passed**; the new file is matched by the existing `src/__tests__/integration/**` include, so it really does run in the blocking `integration-test` job.

## Reviewer notes

- **`core.remoteAccessLaunch.test.ts` gains one mock export.** `readWithPartnerAxisVisibility` imports `getCurrentDbAccessContext` **by name**, so a `vi.mock('../../db')` factory that omits it fails loudly ("No `getCurrentDbAccessContext` export is defined on the mock") rather than silently skipping the escape. That is the documented, intended failure mode (`partnerAxisRead.ts`), and adding the export is the documented remedy. Without it this suite failed 7 tests with `config_error`/500 — the route's `try/catch` converts the throw into a swallowed skip reason, which is worth knowing.
- **`checkRemoteAccessLauncherAvailabilityForDevice` is now exported** solely so the integration test can drive the real chain. The availability path is the safe surface to expose: it never decrypts a password or substitutes a template (#3402). `resolveRemoteAccessLauncherForDevice`, which does, stays unexported.
- **No tenancy widening.** The escaped read stays keyed on `orgId` alone, and both callers pass an org id that came out of `getDeviceWithOrgAndSiteCheck` → `ensureOrgAccess`. Escaping RLS here widens which *columns* of the owning partner are legible; it never widens *which* partner row can be selected. I extended the service docstring to state that contract explicitly.
- **`readPreferredProviderId` is correct as-is** and is not a second instance of this defect: `breeze_user_isolation_select` carries an explicit `OR id = public.breeze_current_user_id()` branch (`apps/api/migrations/2026-04-11-users-rls.sql:225`), so a technician reading their own row needs no escalation.
- **No migration, no schema change, no new table** — so no cascade / export-policy registration applies.

### DIFF
--- a/apps/api/src/__tests__/integration/remoteAccessLauncherPartnerVisibility.integration.test.ts
+++ b/apps/api/src/__tests__/integration/remoteAccessLauncherPartnerVisibility.integration.test.ts
@@ -1,0 +1,291 @@
+/**
+ * Real-Postgres integration coverage for the partner-axis RLS gap in the
+ * remote-access launcher (#3419).
+ *
+ * The bug: `loadRemoteAccessLauncherContext` (routes/devices/core.ts) read
+ * `partners.settings` — a partner-axis table whose only SELECT policy is
+ * `breeze_has_partner_access(id)` (migrations/2026-04-11-partners-rls.sql) —
+ * wrapped in a BARE `withSystemDbAccessContext`. That wrapper looks like an
+ * escalation but is inert inside a request: `withDbAccessContext`
+ * early-returns when a context store already exists (db/index.ts) and
+ * authMiddleware has already opened one, so the read silently kept the
+ * caller's own scope. `computeAccessiblePartnerIds` returns `[]` for
+ * `scope === 'organization'` (middleware/auth.ts), so for an org-scoped
+ * technician the join returned ZERO ROWS — without raising — and the
+ * launcher reported `no_provider_configured` for a tenant that had a
+ * provider configured. Both entry points are
+ * `requireScope('organization', 'partner', 'system')`, so org scope is not a
+ * hypothetical caller: it is the default one.
+ *
+ * A mocked-DB unit test cannot catch this. `core.remoteAccessLaunch.test.ts`
+ * stages the partner row its own assertions then read back, with no RLS
+ * evaluation anywhere; the pre-fix code passed every one of those tests.
+ * Only a real Postgres connection through the `breeze_app` role, inside a
+ * real org-scoped `withDbAccessContext` session, can tell the two apart.
+ *
+ * This file therefore reproduces the exact ambient context an org-scoped
+ * HTTP request runs under (see enrollmentDefaultsPartnerCap.integration.test.ts,
+ * the #2776 precedent this is modelled on) and asserts the launcher still
+ * resolves the partner's provider from inside it.
+ */
+import './setup';
+
+import { afterEach, describe, expect, it } from 'vitest';
+import { eq } from 'drizzle-orm';
+import { db, withDbAccessContext, withSystemDbAccessContext, type DbAccessContext } from '../../db';
+import { organizations, partners } from '../../db/schema';
+import { checkRemoteAccessLauncherAvailabilityForDevice } from '../../routes/devices/core';
+import { readPartnerRemoteAccessSettings } from '../../services/remoteAccessProviders';
+
+const runDb = it.runIf(!!process.env.DATABASE_URL);
+
+/** Mirrors authMiddleware's computeAccessiblePartnerIds for scope 'organization': []. */
+function orgContext(orgId: string): DbAccessContext {
+  return {
+    scope: 'organization',
+    orgId,
+    accessibleOrgIds: [orgId],
+    accessiblePartnerIds: [],
+    userId: null,
+  };
+}
+
+const PROVIDER_ID = 'rustdesk';
+const CUSTOM_FIELD_KEY = 'rustdesk_id';
+
+/** A fully-configured provider: enabled, default, with a URL template. */
+const CONFIGURED_PROVIDERS = {
+  defaultProviderId: PROVIDER_ID,
+  providers: [
+    {
+      id: PROVIDER_ID,
+      name: 'RustDesk',
+      urlTemplate: 'rustdesk://{id}?password={password}',
+      customFieldKey: CUSTOM_FIELD_KEY,
+      password: 'hunter2',
+      enabled: true,
+    },
+  ],
+};
+
+/** The device custom_fields the provider above needs to resolve. */
+const DEVICE_CUSTOM_FIELDS = { [CUSTOM_FIELD_KEY]: '294064193' };
+
+const createdOrgIds: string[] = [];
+const createdPartnerIds: string[] = [];
+
+async function seedTenant(): Promise<{ partnerId: string; orgId: string }> {
+  const unique = `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+  const ids = await withSystemDbAccessContext(async () => {
+    const [partner] = await db
+      .insert(partners)
+      .values({
+        name: `Launcher Partner ${unique}`,
+        slug: `launcher-partner-${unique}`,
+        type: 'msp',
+        plan: 'pro',
+        status: 'active',
+        settings: { remoteAccessProviders: CONFIGURED_PROVIDERS },
+      })
+      .returning({ id: partners.id });
+    const [org] = await db
+      .insert(organizations)
+      .values({
+        currencyCode: 'USD',
+        partnerId: partner!.id,
+        name: `Launcher Org ${unique}`,
+        slug: `launcher-org-${unique}`,
+        type: 'customer',
+        status: 'active',
+      })
+      .returning({ id: organizations.id });
+    return { partnerId: partner!.id, orgId: org!.id };
+  });
+  createdPartnerIds.push(ids.partnerId);
+  createdOrgIds.push(ids.orgId);
+  return ids;
+}
+
+afterEach(async () => {
+  if (createdOrgIds.length === 0 && createdPartnerIds.length === 0) return;
+  await withSystemDbAccessContext(async () => {
+    for (const id of createdOrgIds) {
+      await db.delete(organizations).where(eq(organizations.id, id));
+    }
+    for (const id of createdPartnerIds) {
+      await db.delete(partners).where(eq(partners.id, id));
+    }
+  });
+  createdOrgIds.length = 0;
+  createdPartnerIds.length = 0;
+});
+
+describe('remote-access launcher — partner provider visibility under org-scoped RLS (#3419)', () => {
+  runDb(
+    'an org-scoped caller (accessiblePartnerIds: []) still resolves the partner-configured launcher',
+    async () => {
+      const { orgId } = await seedTenant();
+
+      // The crux of the reproduction: call the resolver from INSIDE the same
+      // kind of RLS context an org-scoped HTTP request runs under, not from
+      // withSystemDbAccessContext. Pre-fix, the partner read ran against
+      // whatever context was already ambient — here, the org-scoped one — so
+      // the join returned nothing and this came back
+      // { available: false, skipReason: 'no_provider_configured' }.
+      const availability = await withDbAccessContext(orgContext(orgId), () =>
+        checkRemoteAccessLauncherAvailabilityForDevice(orgId, DEVICE_CUSTOM_FIELDS),
+      );
+
+      expect(availability).toEqual({
+        available: true,
+        providerId: PROVIDER_ID,
+        skipReason: null,
+      });
+    },
+  );
+
+  runDb(
+    'the delegated partner read itself returns the providers under an org-scoped context',
+    async () => {
+      const { orgId } = await seedTenant();
+
+      const providers = await withDbAccessContext(orgContext(orgId), () =>
+        readPartnerRemoteAccessSettings(orgId),
+      );
+
+      expect(providers?.defaultProviderId).toBe(PROVIDER_ID);
+      expect(providers?.providers?.map((p) => p.id)).toEqual([PROVIDER_ID]);
+    },
+  );
+
+  runDb(
+    'sanity check: the raw partners row IS invisible to that org-scoped context (proves the RLS premise, not just the fix)',
+    async () => {
+      const { orgId, partnerId } = await seedTenant();
+
+      const rows = await withDbAccessContext(orgContext(orgId), () =>
+        db.select({ id: partners.id }).from(partners).where(eq(partners.id, partnerId)),
+      );
+
+      // If this ever starts returning the row, the premise of #3419 (and of
+      // the escape the fix relies on) is gone — the partners RLS policy
+      // itself changed, and this file needs re-evaluating, not green-lighting.
+      expect(rows).toHaveLength(0);
+    },
+  );
+
+  runDb(
+    'negative control: a partner with NO providers configured still reports no_provider_configured',
+    async () => {
+      const unique = `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+      const ids = await withSystemDbAccessContext(async () => {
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` and `pr-test-analyzer` (Sonnet), plus an independent repo-wide sweep for other inert `withSystemDbAccessContext` escapes. The `silent-failure-hunter` lens could not be scheduled (concurrency cap); its brief — swallowed exceptions, residual zero-row fallbacks, and pool-exhaustion-as-silent-stall — was folded into the other two and into my own analysis, and the pool question is written up in the PR body under *Known trade-off*.

**Findings:** 4 raised → 3 addressed in `6665bff68`, 1 rejected with evidence; 0 outstanding.

- **code-reviewer: 0 findings.** It independently ran the new integration test against live Postgres to confirm it isn't vacuous, and separately verified the delegated query is byte-identical to the deleted one and that both callers pass an `orgId` already validated by `getDeviceWithOrgAndSiteCheck` → `ensureOrgAccess`.
- **pr-test-analyzer: 4 findings.** Fixed: (1) `routes/devices.test.ts`'s db mock was missing `getCurrentDbAccessContext` — a **transitive** importer of `core.ts` that my direct-import grep could not see; (2) no coverage of the PARTNER-scoped caller, the shape whose connection behaviour this PR actually changed; (3) the negative control's comment overclaimed — it does *not* discriminate #3419, and now says so. Declined: a separate POST-issuance integration case, since it shares `loadRemoteAccessLauncherContext` byte-for-byte with the covered GET path.
- **Rejected, with evidence:** the analyzer reported finding (1) as *already live* on this branch. It is not. I ran the four suites it named — 68/68 pass, zero stderr, zero occurrences of the error string — and the specific test it named passes clean in isolation. The mock export was added anyway, because the omission is genuine fragility regardless of why it does not currently manifest, but this PR does not claim to fix a defect I could not reproduce.

**Sweep:** `core.ts:1284` was the only occurrence in the primary file, and the only other reader of this setting was already correct. The broader repo sweep produced a longer candidate list that **should not be acted on** — two of its candidates were spot-checked and **both were false positives** (`services/pamApprovers.ts:39`, `routes/partnerApi/provisioning.ts:217`; the sweep has retracted both). Its recurring error was assuming "inside a request ⇒ an ambient context exists", which is false wherever an enclosing caller already escalated, and false for the entire partner-API write path. Details in the PR body; nothing from that list is folded in here.

**Tests:** new suite **6/6** green against real Postgres (verified RED against the pre-fix body, so it genuinely discriminates; the pool-skip case additionally mutation-checked by disabling the skip branch). `src/routes/devices/` 44 files / 592 passed; `devices.test.ts` + `core.remoteAccessLaunch.test.ts` 39/39; `routes/remote/` + `services/remoteAccess*` 114 passed; the 9 other `core.ts`-importing mocked suites 198 passed; `enrollmentDefaultsPartnerCap` (precedent suite) 3/3. `tsc --noEmit` and `eslint` clean.

**Status:** review-clean, awaiting maintainer merge. CI is re-running on `6665bff68`; the pre-push run was green across all completed checks with no failures, but **the blocking `Test API` and `integration-test` jobs had not finished before the push**, so they should be confirmed green on the new head before merging. Base is current — `main` has not moved since this branch was cut, so the merge commit is what was tested locally.

PR #4849, opened 2026-09-03
fix(api): stop act-verify reads pinning a pooled connection across device round-trips (#4150)

### LINKED ISSUE #4150: Act verify reads hold a system DB context across device-command round-trips (#1105 class)
Follow-up to #3826 (PR #4148), from the scoped re-review. `readServiceStatus` (playbookActExecutor.ts) and `verifyProcessAbsent` (actVerify.ts) hold a system DB context across a device-command round-trip (up to 30s / VERIFY_READ_TIMEOUT_MS) — the same idle-in-transaction pool-pinning class as #1105. Restructure to close the context before the wait (phase-split: dispatch outside, read/write result in a short fresh context), mirroring the worker-migration pattern from #4133/3ec0439d2.

ToddHebebrand:
Closes #4150

## The bug

`readServiceStatus` (`playbookActExecutor.ts`) and `verifyServiceRunning` / `verifyProcessAbsent` (`actVerify.ts`) wrapped their `commandQueue.executeCommand` call in `inSystemDbContext`, then awaited a device-command round-trip of up to **30s** (`SERVICE_STATUS_READ_TIMEOUT_MS`) / **8s** (`VERIFY_READ_TIMEOUT_MS`).

`withDbAccessContext` opens a real `baseDb.transaction(...)`, so that held a pooled Postgres connection **idle-in-transaction for the whole wait** — the #1105 pool-exhaustion shape that #4133 / 3ec0439d2 already removed from `monitorWorker`, `discoveryWorker` and `backupWorker`.

## Why the callers couldn't fix it alone

`executeCommand`'s dispatch phase already runs inside `runOutsideDbContext` — but that only exits the AsyncLocalStorage so `db` resolves to the bare pool. It **cannot release a transaction the caller opened**, so the connection stayed pinned regardless.

And the callers can't just drop the wrapper: `executeCommand`'s precheck SELECTs `devices`, which forced-RLS denies on the bare pool, so every dispatch would come back `"Device not found"`.

So the phase split has to happen inside `commandQueue.ts`.

## The change

| | |
|---|---|
| `precheckCommandExecution` (private) | device lookup, partner-trust check, artifact-edition gate, liveness gates, interactive WS fast-fail — **in exactly the previous order**. Needs an RLS context; deliberately opens none of its own, because which context is correct is the caller's decision. |
| `dispatchPreparedCommand` (private) | today's `runOutsideDbContextSafe` body **verbatim** (insert, audit, WS dispatch, `waitForCommandResult`), taking the device snapshot the precheck resolved. |
| `executeCommand` | composes the two. **Behaviour unchanged** for its ~40 request-route callers: the precheck still runs in the caller's auth transaction (the RLS gating it has always had), and every terminal result still predates the row — the "`commandId` present ⇔ row exists" contract holds. |
| `executeCommandWithSystemPrecheck` **(new)** | the depth-0-safe entry point for background callers: a **short** system context around the precheck only, closed before anything waits on the device. Joins **any** ambient context rather than nesting (see below). |

The three AI-agent call sites now use it and hold nothing across the wait. `jobs/aiAgentRunner.ts` already runs the agent loop contextless by design ("*Deliberately NOT wrapped in withSystemDbAccessContext*"), and `runPlaybookSteps` is invoked unwrapped — so the dispatch genuinely lands at depth 0 rather than merely one level shallower.

### Ambient-context handling (both halves came out of review)

`executeCommandWithSystemPrecheck`'s whole value is that nothing is held across the round-trip — and from inside an ambient context it is **powerless** to keep that promise, since `runOutsideDbContext` cannot release someone else's transaction. Two things follow.

**It reports the precondition violation** (`console.warn` + Sentry `db_operation_inside_held_context`, an already-registered code for exactly this) rather than tolerating it silently. Reports rather than throws deliberately: this runs inside the AI run loop's `postToolUse` hook, where a throw downgrades a real verification to `inconclusive` — a slow dispatch beats a lost one. The db-layer `db_context_held_too_long` tripwire fires on the same hold but attributes it to whoever *opened* the context; this line names the dispatch that made it long, which is the actionable half. Throttled to one Sentry event per scope per 15 min (the sibling guards on this eventCode all dedup, each citing a real quota-flood incident); the console line is unthrottled because it carries the `deviceId`/`type` no Sentry tag can — `commandType` is absent from `ALLOWED_TAG_NAMES` and would be scrubbed.

**And it joins ANY ambient context, never nests inside one.** A first pass only joined when the ambient scope was already `system`; review caught that an `organization`/`partner` caller — the realistic violation, a request handler's RLS transaction reaching this function — fell through to `runOutsideDbContext(() => withSystemDbAccessContext(...))` and did *both* things the code's own comments call unacceptable: a **second** pooled connection checked out while the caller's is still held, and the precheck's `devices` read + trust check running with **full cross-tenant visibility** instead of the caller's narrower scope. `withDbAccessContext` already reuses any active transaction regardless of requested scope — that is what made the `system` join work — so dropping the escape makes the same reuse cover the tenant scopes. A tenant-scoped caller now degrades toward "Device not found" instead of seeing another tenant's device: the **fail-closed** direction, and the repo's standing contract that too little context denies rather than bypasses.

### Deliberately not done

Making `executeCommand` auto-open a system context when no ambient context exists. That would turn a forgotten-context bug in a request route into a **silent tenancy bypass at the dispatch chokepoint**. Contextless access must keep denying.

## Tests

Three new `*.dbcontext.test.ts` suites mirroring `jobs/monitorWorker.dbcontext.test.ts`'s `ctxState.events` harness — they track real enter/exit **depth** and assert which depth each DB touch and each command dispatch runs at:

- `services/commandQueue.dbcontext.test.ts` — the precheck runs at depth 1, the context **exits**, then `claim` / `wsSend` / the result poll all run at depth 0. Plus: `executeCommand` still opens **no** context of its own; the ambient **join** is exercised across all three scopes (`system` / `organization` / `partner`) and asserts no second context is opened in each (via `insert:device_commands@depth2` — a second connection would make the `ctx:enter` count 2) while the guard fires with the right event code and `scope` tag; the Sentry throttle holds across three calls while the console warning fires every time; a `TrustDeniedError` still surfaces as a terminal result with its `trust` payload; a non-trust precheck error still **propagates** rather than degrading into a failed-command result.
- `services/aiAgents/actVerify.dbcontext.test.ts` — both verify read-backs dispatch at depth 0 (including the throwing path); `recordActVerifyFailureAlert` keeps its short context around the insert.
- `services/aiAgents/playbookActExecutor.dbcontext.test.ts` — `readServiceStatus` dispatches at depth 0; `readDiskUsagePercent` keeps its short context (a correct short context around a pure DB read must stay).

`runOutsideDbContext` is mocked as a **depth-preserving** passthrough on purpose, because that is what it really is. Modelling it as a depth reset would make these suites pass against the very bug they exist for.

**Mutation-verified**, not just asserted. Against the pre-fix shape, and again by re-introducing `inSystemDbContext` around the fixed call:

```
- "executeCommand@depth0"
+ "ctx:enter"
+ "executeCommand@depth1"
+ "ctx:exit"
```

`actVerify.test.ts` is updated for the renamed entry point (1:1 substitution, no assertion loosened — its identity-mocked context helpers can assert *that* a system context was used, never *how long* it was held, hence the new suites).

## Verification

- `vitest run src/services/aiAgents src/routes/systemTools src/services/commandQueueTransitions.test.ts` — **78 files / 2123 tests pass**
- `vitest run src/services/commandQueue.dbcontext.test.ts src/services/commandQueue.test.ts` — 59 pass
- The 9 `jobs/*` + 18 route suites that consume `commandQueue` — 347 pass
- A normalized statement-level diff of the old `executeCommand` body against `precheckCommandExecution` + `dispatchPreparedCommand` shows **no logic line dropped** — only the signature/doc relocation and the `{ ok, result }` wrapping of the terminal returns
- `npx tsc --noEmit` clean; `eslint` on all changed files clean

## Out of scope (noted, not fixed)

Every ordinary `executeCommand` call from an authenticated route is still wrapped request-wide by `authMiddleware`'s transaction, and the system-tools routes are not in `SELF_MANAGED_DB_CONTEXT_ROUTES` — so those retain the same connection hold across their own round-trips. Same for the AI tool path, where `aiAgentSdkTools.ts`'s `makeHandler` re-enters a tenant context around every `executeTool`. Both need a middleware opt-out plus caller-scoped prechecks and are a materially larger change than #4150; flagging rather than expanding scope here.

### DIFF
--- a/apps/api/src/services/aiAgents/actVerify.dbcontext.test.ts
+++ b/apps/api/src/services/aiAgents/actVerify.dbcontext.test.ts
@@ -1,0 +1,164 @@
+/**
+ * actVerify DB-context scoping (#4150, #1105 class).
+ *
+ * The regression this locks down: `verifyServiceRunning` and
+ * `verifyProcessAbsent` wrapped their `executeCommand` call — a device-command
+ * round-trip bounded at VERIFY_READ_TIMEOUT_MS — in `inSystemDbContext`, so a
+ * pooled Postgres connection sat idle-in-transaction for the whole wait.
+ * Exactly the shape #4133/3ec0439d2 removed from the BullMQ workers.
+ *
+ * `actVerify.test.ts` mocks the context helpers as identity passthroughs, so
+ * it can assert *that* a system context was used but never *how long* it was
+ * held. This suite tracks real enter/exit DEPTH and asserts which depth each
+ * dispatch and each DB write ran at.
+ *
+ * `runOutsideDbContext` is deliberately mocked as a depth-PRESERVING
+ * passthrough: that is what it really is (it exits the AsyncLocalStorage but
+ * cannot release an outer `withDbAccessContext` transaction's connection).
+ * Modelling it as a depth reset would make this suite pass against the very
+ * bug it exists for.
+ */
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { dbMockState, ctxState, commandQueueMock } = vi.hoisted(() => ({
+  dbMockState: {
+    ambientContext: undefined as { scope: string } | undefined,
+    insertedRows: [] as unknown[],
+  },
+  ctxState: { depth: 0, events: [] as string[] },
+  commandQueueMock: {
+    executeCommandWithSystemPrecheck: vi.fn(),
+  },
+}));
+
+vi.mock('../../db', () => ({
+  db: {
+    insert: vi.fn(() => ({
+      values: vi.fn(async (row: unknown) => {
+        ctxState.events.push(`alertInsert@depth${ctxState.depth}`);
+        dbMockState.insertedRows.push(row);
+      }),
+    })),
+  },
+  getCurrentDbAccessContext: vi.fn(() => dbMockState.ambientContext),
+  // Depth-preserving on purpose — see the file header.
+  runOutsideDbContext: vi.fn((fn: () => unknown) => fn()),
+  withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => {
+    const previous = dbMockState.ambientContext;
+    dbMockState.ambientContext = { scope: 'system' };
+    ctxState.depth++;
+    ctxState.events.push('ctx:enter');
+    try {
+      return await fn();
+    } finally {
+      ctxState.depth--;
+      ctxState.events.push('ctx:exit');
+      dbMockState.ambientContext = previous;
+    }
+  }),
+}));
+
+vi.mock('../commandQueue', () => commandQueueMock);
+
+import { ACT_MANIFEST, type ActOperation } from './actManifest';
+import { recordActVerifyFailureAlert, verifyActExecution } from './actVerify';
+import type { ActAssetPin } from './actRevalidation';
+
+const RUN = { id: 'run-1', orgId: 'org-1', agentId: 'agent-1', deviceId: 'device-1' };
+const AGENT_USER_ID = 'agent-1';
+
+const restartOp = ACT_MANIFEST.find((op) => op.key === 'manage_services.restart')!;
+// Mirrors actVerify.test.ts: manage_processes.kill is not in ACT_MANIFEST, but
+// `verifyProcessAbsent` is still live code and carries the same #1105 shape.
+const processAbsentOp: ActOperation = {
+  key: 'manage_processes.kill',
+  toolName: 'manage_processes',
+  matches: () => false,
+  normalizeTarget: () => ({ ok: false, reason: 'unreachable — not in ACT_MANIFEST' }),
+  verifySpec: { kind: 'process_absent' },
+};
+
+function recordingResult(stdout: string) {
+  return vi.fn(async () => {
+    ctxState.events.push(`executeCommand@depth${ctxState.depth}`);
+    return { status: 'completed' as const, stdout };
+  });
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  dbMockState.ambientContext = undefined;
+  dbMockState.insertedRows = [];
+  ctxState.depth = 0;
+  ctxState.events = [];
+});
+
+describe('verifyActExecution DB-context scoping (#4150/#1105)', () => {
+  it('runs the service_running read-back with NO DB context held', async () => {
+    commandQueueMock.executeCommandWithSystemPrecheck = recordingResult(
+      JSON.stringify({ services: [{ name: 'Spooler', status: 'Running' }] }),
+    );
+
+    const outcome = await verifyActExecution({
+      pin: { op: restartOp, target: { kind: 'service', serviceName: 'Spooler' } } as ActAssetPin,
+      toolOutput: JSON.stringify({ status: 'completed' }),
+      isError: false,
+      run: RUN,
+      agentUserId: AGENT_USER_ID,
+    });
+
+    expect(outcome.verification).toBe('passed');
+    // THE assertion: before the #4150 fix this read `executeCommand@depth1`.
+    expect(ctxState.events).toEqual(['executeCommand@depth0']);
+  });
+
+  it('runs the process_absent read-back with NO DB context held', async () => {
+    commandQueueMock.executeCommandWithSystemPrecheck = recordingResult(
+      JSON.stringify({ processes: [{ pid: 999 }] }),
+    );
+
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, plus one scoped re-review of the fix commit (`code-reviewer` on `97cea9510` only, because that fix touched tenancy-relevant context semantics). Design was settled up front by an advisor quorum (Fable + Codex `gpt-5.6-sol` read-only) before any code was written — both independently rejected the "auto-open a system context when none exists" shape as a fail-open tenancy hazard, and converged on the single new export over an option flag.

**Findings:** 5 raised across the four runs → all 5 addressed; 0 outstanding.

| # | Run | Finding | Resolution |
|---|---|---|---|
| 1 | code-reviewer | `executeCommandWithSystemPrecheck` promised "holds nothing across the round-trip" but had no guard when called from inside an ambient context, where it is powerless to keep that promise | `f19865527` — reports it (`console.warn` + Sentry `db_operation_inside_held_context`). Reports rather than throws: it runs in the AI loop's `postToolUse` hook, where a throw downgrades a real verification to `inconclusive` |
| 2 | silent-failure-hunter | **Critical.** That guard only *joined* an ambient `system` context; `organization`/`partner` fell through to escape-and-escalate — a **second** pooled connection plus the precheck reading with **full cross-tenant visibility** | `97cea9510` — joins **any** ambient scope now. A tenant-scoped caller degrades toward "Device not found" instead of seeing another tenant's device: fail-closed, per the repo's "too little context denies rather than bypasses" contract |
| 3 | silent-failure-hunter | Sentry capture unthrottled, unlike every sibling guard on that eventCode (each citing a real quota-flood incident) — and 2 of the 3 rewired call sites sit in a hot AI path | `97cea9510` — one capture per scope per 15 min (bounded: 3 scopes). Console line stays unthrottled; it carries the `deviceId`/`type` no tag can |
| 4 | silent-failure-hunter | Guard didn't attribute the violation to a caller | `97cea9510` — `deviceId`/`type`/`scope` in the log line. `commandType` was dropped from the Sentry tags after checking `ALLOWED_TAG_NAMES` — it would have been silently scrubbed; `scope` is allowlisted and rides the tag while the message stays invariant as the grouping key |
| 5 | re-review | `db/index.ts` already exports `shouldCaptureHeldContext` with the identical shape | `49f1b209d` — **not** reused, deliberately: its map is shared with the `db_context_held_too_long` capture, so one signal would silently suppress the other. Reasoning recorded in a comment |

Two items were left as **noted-only** by `pr-test-analyzer` and I agree: the send-retry-exhausted path and the `AUDITED_COMMANDS` audit write are only covered through `executeCommand`, but both live in `dispatchPreparedCommand`, which is byte-identical to the pre-refactor code and indifferent to which entry point called it.

**Verification:**

- **~8,600 tests green** across every suite that touches `commandQueue`, the AI-agent paths, or `jobs/` — `services/aiAgents` (73 files / 1949), `routes/agents|alerts|backup|devices|patches|security|auditBaselines` + `devices.endpoints` + `agentWs` (157 files / 2484), `services/aiTools*|partnerTrust*` + 11 more services + all of `jobs/` (278 files / 3975), and the six files this PR touches (122)
- `npx tsc --noEmit` clean; `eslint` clean on every changed file
- **Mutation-verified three ways, not just asserted.** The `playbookActExecutor` guard was confirmed red against pre-fix `main`; the `actVerify` guard was confirmed red by re-introducing the `inSystemDbContext` wrapper (2 failures, `executeCommand@depth1`); and the re-reviewer independently reverted the join condition to `scope === 'system'` and confirmed the `organization`/`partner` cases fail. None of these tests can pass against the bug they guard.
- A normalized statement-level diff of the old `executeCommand` body against `precheckCommandExecution` + `dispatchPreparedCommand` shows **no logic line dropped** — only the signature/doc relocation and the `{ ok, result }` wrapping of the terminal returns. The re-reviewer separately confirmed the dispatch phase is byte-for-byte identical.

**Status:** review-clean, awaiting maintainer merge. CI was queued behind runner contention at hand-off — please confirm **Test API** and **Integration Tests** are green on `49f1b209d` before merging.

PR #4851, opened 2026-09-03
fix(api): make mobile cursor pagination reachable and keyset-safe (#3770)

### LINKED ISSUE #3770: [API][Mobile] Cursor pagination on /mobile/devices and /mobile/alerts/inbox is unreachable, and unsafe if reached
Both paginated mobile list routes advertise a keyset cursor that no client can enter, and the keyset itself would skip rows if it were entered. Splitting this out of #3753, where it was the root cause of the device half.

## 1. Cursor mode is unreachable

`apps/api/src/routes/mobile.ts:854-862` (`/alerts/inbox`) and `:1157-1165` (`/devices`):

```ts
let nextCursor: string | null = null;
if (cursor) {
  const hasMore = rows.length > limit;
  ...
  if (hasMore && last) nextCursor = encodeCursor(last.triggeredAt, last.id);
}
```

`nextCursor` is computed **only when a cursor was already supplied**. A cold-start caller has no cursor to send, `decodeCursor(undefined)` returns null, so its first response always carries `nextCursor: null`. There is no path from "first request" to "a cursor", which makes the mode documented, unit-tested at the encode/decode layer, and unreachable end to end.

The route comment at `:770-774` describes it as usable ("opaque `cursor` from a prior response's `nextCursor`"), which is precisely the thing that cannot happen.

## 2. What it cost

The mobile client's device walk followed `nextCursor`, got null on page one, and treated that as **clean exhaustion** — so it stopped after one page *and* set `cursor = null`, which meant its own truncation warning could never fire either. `getDevices()` returned at most 100 devices and reported them as a complete fleet. Silent.

`devicesPaging.test.ts` did not catch it because its fixture returned a cursor on page one, a response the real server has never produced. Code, comment and test all agreed with each other about the server.

Fixed on the client in #3769 by not walking at all and reporting partial-ness honestly. That is a mitigation, not a fix for this.

## 3. Do not just make it reachable

I built that first — always over-fetch `limit + 1` and compute the cursor unconditionally. It passes the full API suite and it is the wrong change, because the keyset has three independent boundary hazards:

- **Nullable ordering column.** `devices.lastSeenAt` is `timestamp('last_seen_at')` with no `.notNull()` (`db/schema/devices.ts:86`; baseline migration agrees). Postgres `ORDER BY ... DESC` puts NULLs **first** and neither route says otherwise, so null rows lead. `encodeCursor` returns `null` when `ts` is falsy (`mobile.ts:69`), so a page ending on a null row emits no cursor and terminates the walk early.

- **Microsecond truncation.** `alerts.triggeredAt` is `timestamp(...).defaultNow().notNull()` (`db/schema/alerts.ts:102`) and Postgres keeps six fractional digits, but `encodeCursor` serialises via `Date.toISOString()` — milliseconds (`mobile.ts:70-71`). A row stored at `.123999` encodes as `.123`, and the next page's `triggered_at < '.123'` predicate skips every row in between.

- **Mutable sort key.** `devices.last_seen_at` is rewritten by every heartbeat. A device sitting below the boundary that heartbeats mid-walk moves ahead of the cursor and is never returned. Default heartbeat is 60s, so a 210-device fleet rewrites roughly 3.5 sort keys per second.

## 4. The repo already solved this on the other device route

`apps/api/src/routes/devices/core.ts:453-457` defaults cursor mode to `hostname ASC`, with the reasoning written down: *"the keyset's monotonicity is most stable on a NOT NULL string column."* `routes/devices/cursor.ts` implements explicit NULLS LAST and a separate null-cursor phase. The mobile routes never received that treatment.

## Suggested resolutions

1. **Port the `core.ts` pattern** — key mobile cursor mode on an immutable NOT NULL column, explicit NULLS LAST, and preserve full timestamp precision in the token — then make it reachable.
2. **Or delete cursor mode** from these two routes and keep offset + `total`, so nothing advertises a contract it cannot honour.

Happy to do either; I did not want to pick unilaterally since (1) changes response behaviour for every client. Note that offset paging on these routes has its own skew for the same mutable-key reason, so option 2 should keep surfacing `total` and let callers say "showing N of M" rather than implying completeness.

Both mobile routes always return a `pagination` block today, including on the early empty returns, so `total` is a dependable signal for either option.

ToddHebebrand:
## Summary

`GET /mobile/alerts/inbox` and `GET /mobile/devices` both computed `nextCursor` only inside `if (cursor)`. A cold-start caller has no cursor to send, so its first response always carried `nextCursor: null` — there was no path from "first request" to "a cursor", making cursor mode documented, unit-tested at the encode/decode layer, and unreachable end to end (#3770).

Making it reachable exposed two real keyset hazards once actually walked, both described in the issue:

- **`/mobile/devices`** keyed on `last_seen_at`, which is nullable (Postgres sorts NULLs first on `DESC` with no `NULLS LAST`) and mutable (every heartbeat rewrites it, so a device can cross the page boundary mid-walk and be skipped). Cursor mode now keys on `(hostname ASC, id ASC)` instead — NOT NULL and effectively immutable — porting the pattern already used by `routes/devices/core.ts`. Legacy `?page=N` keeps the pre-existing `last_seen_at DESC` order and never mints a `nextCursor`, since the two orders aren't interchangeable mid-walk (mirrors `core.ts`'s own mode split).
- **`/mobile/alerts/inbox`** round-tripped `triggered_at` through a JS `Date` to build the cursor. Postgres keeps six fractional digits; `Date` only holds three (milliseconds), so a truncated cursor's `<` comparison could skip real rows sitting between the truncated boundary and the true one. The route now reads the same column back as raw microsecond text via `to_char(...)` for the cursor only (`triggeredAtKey`, never surfaced in `data`), and the cursor codec (`encodeCursor`/`decodeCursor`) never re-parses `key` into a `Date` at all — it's carried as an opaque string end to end.

Both routes always over-fetch by `limit + 1` now (not gated on `cursor`) so `nextCursor` is available on the first page/limit request, and `total` is still returned in every response as before.

Per `apps/mobile/src/services/api.ts`, the mobile client already deliberately fetches only one page and never reads `nextCursor` (a separate, intentional mitigation from #3769) — so it's unaffected by either the reachability fix or the devices ordering change. Updated its stale doc comments (and a test-fixture comment in `listPaging.test.ts`) that described the now-fixed server bug as current.

## Changes

- `apps/api/src/routes/mobile.ts`:
  - Cursor codec (`CursorTuple`, `encodeCursor`, `decodeCursor`) now carries a generic `key: string` instead of `ts: Date` — never round-trips through `Date`. Added `decodeTimestampCursor` for the alerts route, which additionally rejects a structurally-valid cursor whose `key` doesn't parse as a timestamp (so a bad cast can't 500 the route).
  - `/alerts/inbox`: always computes `nextCursor`; selects `triggeredAtKey` (raw `to_char` text) for the cursor instead of the `Date`-typed `triggeredAt`.
  - `/devices`: `isCursorMode` (mirrors `devices/core.ts`) — no `page` given, or a `cursor` present — switches ordering to `hostname ASC, id ASC` and only then mints a `nextCursor`; an explicit `?page=N` keeps `last_seen_at DESC` and never returns one.
- `apps/api/src/routes/mobile.cursor.test.ts`, `apps/api/src/routes/mobile.test.ts`: updated existing cursor-codec tests for the new `{key, id}` shape, and added coverage for the issue's three scenarios — first page returns `nextCursor`, following it yields the next page with no duplicates/gaps (verified via the captured keyset predicate), and the last page returns `nextCursor: null` — for both routes, plus a legacy `?page=N` regression test for `/devices` and `decodeTimestampCursor` coverage.
- `apps/mobile/src/services/api.ts`, `apps/mobile/src/services/listPaging.test.ts`: doc-comment updates only, no behavior change.

## Test plan

- [x] `cd apps/api && npx vitest run src/routes/mobile.cursor.test.ts src/routes/mobile.test.ts` — 84 passed, 2 skipped
- [x] `cd apps/api && npx vitest run src/routes/mobile.ackCas.test.ts src/routes/mobile.resolveCas.test.ts src/__tests__/helpers/routeScan.test.ts src/services/mobileDeviceIdentity.test.ts` — 51 passed (other mobile.ts consumers, unaffected)
- [x] `cd apps/mobile && npx vitest run src/services/listPaging.test.ts` — 8 passed
- [x] `cd apps/api && npx tsc --noEmit -p .` — clean
- [x] `cd apps/mobile && npx tsc --noEmit` — clean
- [x] `cd apps/api && npx eslint src/routes/mobile.ts src/routes/mobile.test.ts src/routes/mobile.cursor.test.ts` — clean

Closes #3770

### DIFF
--- a/apps/api/src/__tests__/integration/mobileCursorPagination.integration.test.ts
+++ b/apps/api/src/__tests__/integration/mobileCursorPagination.integration.test.ts
@@ -1,0 +1,183 @@
+/**
+ * Real-Postgres proof for the two raw-SQL fragments introduced by #3770's
+ * fix to GET /mobile/alerts/inbox and GET /mobile/devices
+ * (apps/api/src/routes/mobile.ts).
+ *
+ * apps/api/src/routes/mobile.test.ts already covers the route's BEHAVIOR
+ * (nextCursor reachability, ordering-mode split, response shape) against a
+ * mocked `db.select` — but a mock cannot catch a typo in a `to_char` format
+ * string, an invalid `::timestamp`/`::uuid` cast, or Postgres's actual tuple-
+ * comparison / NULL-ordering semantics; it would happily return whatever rows
+ * the test tells it to. This file executes the SAME query fragments the
+ * route builds (select projection, cursor predicate, ORDER BY) against a
+ * real table, so a cast or syntax mistake fails here instead of in
+ * production. See `devicesCursorRls.integration.test.ts` for the identical
+ * rationale applied to routes/devices/core.ts's cursor.
+ *
+ * Prerequisites:
+ *   docker compose -f docker-compose.test.yml up -d
+ *
+ * Run:
+ *   pnpm test:integration -- src/__tests__/integration/mobileCursorPagination.integration.test.ts
+ */
+import './setup';
+
+import { describe, it, expect } from 'vitest';
+import { and, asc, desc, eq, sql } from 'drizzle-orm';
+import { getTestDb } from './setup';
+import { alerts, devices } from '../../db/schema';
+import { createOrganization, createPartner, createSite } from './db-utils';
+import { encodeCursor, decodeCursor, decodeTimestampCursor } from '../../routes/mobile';
+
+let agentIdCounter = 0;
+async function insertDevice(opts: { orgId: string; siteId: string; hostname: string }): Promise<string> {
+  const db = getTestDb();
+  agentIdCounter++;
+  const [row] = await db
+    .insert(devices)
+    .values({
+      orgId: opts.orgId,
+      siteId: opts.siteId,
+      agentId: `agent-mobile-cursor-${agentIdCounter}-${Date.now()}`,
+      hostname: opts.hostname,
+      displayName: opts.hostname,
+      osType: 'windows',
+      osVersion: '11',
+      osBuild: '22000',
+      architecture: 'x86_64',
+      agentVersion: '0.0.0-test',
+      status: 'online',
+      enrolledAt: new Date(),
+    })
+    .returning({ id: devices.id });
+  if (!row) throw new Error('insertDevice: insert returned no row');
+  return row.id;
+}
+
+/**
+ * Insert an alert with a `triggered_at` carrying a specific fractional-second
+ * suffix (up to 6 digits) that a JS `Date` cannot represent — using raw SQL
+ * rather than Drizzle's `.values({ triggeredAt: someDate })`, because a plain
+ * `Date` object can only ever contribute millisecond precision. That's the
+ * whole point of this test: proving the fix survives the values a `Date`
+ * cannot even construct.
+ */
+async function insertAlertAt(opts: { orgId: string; deviceId: string; triggeredAt: string }): Promise<string> {
+  const db = getTestDb();
+  const rows = await db.execute<{ id: string }>(sql`
+    INSERT INTO alerts (device_id, org_id, severity, title, triggered_at)
+    VALUES (${opts.deviceId}, ${opts.orgId}, 'critical', 'test alert', ${opts.triggeredAt}::timestamp)
+    RETURNING id
+  `);
+  const row = rows[0];
+  if (!row) throw new Error('insertAlertAt: insert returned no row');
+  return row.id;
+}
+
+describe('mobile cursor pagination — real SQL (#3770)', () => {
+  it('/alerts/inbox: to_char + ::timestamp round-trips full microsecond precision and does not skip the boundary row', async () => {
+    const db = getTestDb();
+    const partner = await createPartner();
+    const org = await createOrganization({ partnerId: partner.id });
+    const site = await createSite({ orgId: org.id });
+    const deviceId = await insertDevice({ orgId: org.id, siteId: site.id, hostname: 'host-1' });
+
+    // A sits just above B on the SAME millisecond, differing only in the
+    // fractional digits a JS Date cannot hold. The historical bug: encoding
+    // A's cursor via `Date.toISOString()` truncates '.123999' to '.123',
+    // and `triggered_at < '.123'` then wrongly excludes B ('.123500' is NOT
+    // less than the truncated '.123000').
+    const idA = await insertAlertAt({ orgId: org.id, deviceId, triggeredAt: '2026-07-28 12:00:00.123999' });
+    const idB = await insertAlertAt({ orgId: org.id, deviceId, triggeredAt: '2026-07-28 12:00:00.123500' });
+
+    // ---- Page 1: exactly what the route selects/orders, limit=1 ----
+    const page1 = await db
+      .select({
+        id: alerts.id,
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, pr-test-analyzer, silent-failure-hunter — parallel, Sonnet)

**Findings:** 1 Critical, 2 Important, 2 Suggestions raised across the round.
- **Critical (fixed, `2c9e2cd`):** `decodeTimestampCursor` validated a cursor's timestamp via `Number.isNaN(new Date(key).getTime())`. JS `Date` accepts a far wider grammar/range than Postgres `timestamp` — e.g. `new Date(-8640000000000000).toISOString()` parses cleanly but is nowhere near Postgres's range — so a crafted cursor could reach the route's `::timestamp` cast and 500 instead of falling back cleanly. Now validated against the exact `to_char(...,'YYYY-MM-DD"T"HH24:MI:SS.US')` shape the encode side emits (`/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}$/`), plus a regression test proving the old check would have let the extreme-date payload through.
- **Important (addressed, `db352db`):** the mocked-`db.select` unit tests couldn't catch a typo in the raw `to_char` format string or a bad `::timestamp`/`::uuid` cast — a mock returns whatever rows it's told to. Added `mobileCursorPagination.integration.test.ts`, which runs the exact select/predicate/orderBy fragments against a real Postgres (worktree-private `test-stack`): proves the alerts cursor's microsecond round-trip doesn't skip a boundary row, and walks the devices hostname keyset end-to-end including a NULL `last_seen_at` row.
- **Important (pre-existing, no action):** a malformed cursor silently restarts the walk at page 1 with no `400`/error signal — unchanged by this PR (the `if (cursor)`-null fallback shape predates it), and today's only consumer never walks cursors, so it's currently inert. Flagged for awareness, not blocking.
- **Suggestion (applied, `db352db`):** `/devices` was over-fetching `limit + 1` even in legacy `?page=N` mode, where `nextCursor` is discarded regardless — now gated on `isCursorMode`, matching `devices/core.ts`'s existing guard.
- **Suggestion (not applied, low confidence/out of scope):** a cursor `key` with an embedded NUL byte reaching the hostname tuple comparison as a text bind param is a theoretical Postgres protocol-error surface with no sanitization — very low likelihood, and not new to this PR (other free-text params in this file are equally unguarded).

**Tests:** `apps/api/src/routes/mobile.cursor.test.ts` + `apps/api/src/routes/mobile.test.ts` — 85 passed, 2 skipped (mocked-db unit, `npx vitest run`); `apps/api/src/__tests__/integration/mobileCursorPagination.integration.test.ts` — 2 passed against a real Postgres instance (worktree-private `pnpm test-stack`, torn down after); `apps/mobile/src/services/listPaging.test.ts` — 8 passed (doc-comment-only change there). `tsc --noEmit` clean for both `apps/api` and `apps/mobile`. `eslint` clean on the touched API files.

Issue #4084, opened 2026-08-26
W08: Wave 3.5b — socket-affinity resolution: agent command dispatch must not depend on the API process's in-memory connection map (BLOCKS the BREEZE_ROLE split)


ToddHebebrand:
Wave 3.5b of #3821. Split out of #3825 on 2026-08-26 — see the plan doc
`docs/superpowers/plans/ai-mcp/2026-08-26-ai-agents-wave3.5a-idempotency.md`
(landed in #4083) for the full four-part decomposition.

## This is the release blocker for the BREEZE_ROLE split

`sendCommandToAgent()` (`apps/api/src/routes/agentWs.ts:3067`) resolves the
target socket from an **in-process `Map`**:

```ts
export function sendCommandToAgent(agentId: string, command: AgentCommand): boolean {
  const ws = activeConnections.get(agentId);
  if (!ws) return false;
```

`isAgentConnected()` (`agentWs.ts:3124`) is likewise just `activeConnections.has()`.

Four BullMQ workers import and call these **directly**:

- `apps/api/src/jobs/monitorWorker.ts:162,168`
- `apps/api/src/jobs/backupWorker.ts:502,653`
- `apps/api/src/jobs/discoveryWorker.ts:571,606`
- `apps/api/src/jobs/snmpWorker.ts:16` (used at `:412`)

Move those workers into a separate `worker` container and **every agent reads as
disconnected**: discovery and backup jobs get marked failed, monitor and SNMP
quietly report "no online agent". This breaks agent command dispatch entirely,
and it is independent of the event-bus work in 3.5c.

No test covers it, because today the publisher and the workers are the same
process.

## The decision this wave has to make

Either:

1. **Pin socket-affine workers to the `api` role.** Cheapest. Costs the isolation
   for exactly the workers that do the heaviest agent I/O, so 3.5d delivers less
   than it appears to.
2. **Build a cross-process command relay** to the API instance that owns the
   socket, with durable command identity and result routing. Correct long-term,
   materially more work, and it is the thing that makes a multi-replica API
   possible later.

Recommend settling this with an advisor quorum before implementing — it is a
cross-module contract and hard to reverse.

## Done when

- A decision is recorded (ADR or plan doc) with the reasoning.
- Agent command dispatch is proven to work from a process that does NOT hold the
  agent's WebSocket, or the affected workers are explicitly and testably pinned
  to the `api` role.
- A test fails if a future change moves a socket-affine worker off the role that
[...]

PR #4852, opened 2026-09-03
fix(api): stop backup dispatch duplicating child job rows on retry (#4137)

### LINKED ISSUE #4137: backupWorker: child-row duplication on BullMQ retry after the #4084 phase-split + multi-target dispatch test coverage
Follow-up to #4084 (PR #4133), from the scoped re-review of 3ec0439d2. (1) Phase 3 now commits child backup_jobs rows BEFORE the sends (fixing a pre-existing race where a fast agent result referenced an uncommitted child row) — but a throw or process kill mid-Phase-4 leaves them committed, and the BullMQ retry (attempts:3 via PRIVILEGED_JOB_OPTIONS) inserts a fresh set: orphaned 'running' children per failed attempt (the stuck-running class). Fix: parentJobId linkage + dedup on retry, or attempts:1 for dispatch-backup. (2) The multi-target path (child inserts, preFailedTargets, partial errorLog, failedChildJobs settle) has no unit test — a 2-target case with a failing second send would have caught the cancel-race regression the re-review found.

ToddHebebrand:
Closes #4137

## The bug

`prepareBackupDispatchTargets` (Phase 3 of `processDispatchBackup`) INSERTs a
fresh `backup_jobs` child row for every target after the first, and — since the
#4084 phase split — commits them **before** the Phase-4 sends. `dispatch-backup`
was enqueued with `PRIVILEGED_JOB_OPTIONS` (`attempts: 3` + exponential
backoff), so a throw anywhere in Phase 4/5 re-ran Phase 3 and created a **second
child set**, leaving the first stranded at `status='running'` forever. Nothing
sweeps children by parent — `backup_jobs` has no parent linkage column, and the
cancel route only ever touches the parent id.

The multi-target path (child inserts, `preFailedTargets`, partial `errorLog`,
`failedChildJobs` settle) had **zero unit coverage**: every existing dispatch
test pins the mode lookup to `backupMode: 'file'`, which resolves to exactly one
target, so the `i > 0` branch never executed.

## What changed

**1. `dispatch-backup` becomes a one-shot.** It gets its own
`DISPATCH_JOB_OPTIONS = { attempts: 1 }` (no backoff), matching the repo's other
non-idempotent one-shots (`aiAgentEnqueuer`, `orgMerge`, `tenantErasure`).
`process-results` keeps `attempts: 3` + backoff — it re-applies the same agent
payload to the same row, so retrying it converges.

**2. Re-deliveries are refused.** `attempts: 1` alone does **not** close the hole
the issue describes. Verified against `bullmq@5.81.2`
`moveStalledJobsToWait-9.lua`: a **stalled** job — worker process killed
mid-dispatch, exactly the case the issue names — is pushed back to `wait` and
re-processed up to `maxStalledCount` (2 on this worker) **independently of
`attempts`**. `processDispatchBackup` now bails before Phase 3 when
`job.attemptsStarted > 1`. That counter is HINCRBY'd by
`prepareJobForProcessing.lua` on every move-to-active, so it is exactly "an
earlier execution already started this dispatch"; `reprocessJob` clears it, so a
deliberate operator retry is *not* suppressed.

**3. Exception cleanup — the "settle partial rows?" question.**
**Decision: settle, but only the rows that provably never reached the agent.**
Blanket-failing everything would be wrong: `applyBackupCommandResultToJob` only
accepts a `pending`/`running` row, so terminating a row whose command *did* go
out makes the agent's genuine result get dropped on arrival. So on a throw:

| target state | action | why |
|---|---|---|
| `sent` | left in-flight | a real result is still coming |
| `attempting` (the send that threw) | left in-flight | delivery is **ambiguous** |
| `failed` / `not-attempted` | settled to `failed` | provably never delivered |

The state is set *before* the `await`, which is what makes the ambiguous case
distinguishable from the never-attempted one. The original error is **rethrown**
so the worker's `failed` listener still reports it to Sentry (swallowing it
would record a BullMQ success); the settle itself is best-effort and cannot mask
the cause.

**4. Two adjacent leaks in the same class, fixed inline.**
- A cancel landing mid-Phase-3 now settles **every** child row already
  committed. Previously a cancel at the top of iteration `i` cancelled nothing
  and stranded children `1..i-1` at `running`.
- The parent row's final status now follows **its own target's** outcome instead
  of the aggregate `sentCount`. A run whose first target failed while a child
  succeeded used to flip the parent to `running` for a command the agent never
  received, leaving it in-flight until the stale reaper's 24h timeout.

## Design notes

An earlier draft claimed the parent row atomically (`UPDATE … WHERE
status='pending'`) as the dedup key. **Rejected on review** — `running` is not a
free-standing lock in this schema: `backupSnapshotReconcile` treats
`running|failed` as eligible for time-window snapshot attribution (a
claimed-but-never-sent job could adopt an unrelated orphan snapshot),
`started_at` is documented as the real run clock, and `staleCommandReaper` uses
a 1h timeout for `pending` vs 24h for `running`. The BullMQ-counter guard gets
the same at-most-once property without touching domain status or needing a
migration.

At-most-once is the right trade here: a worker killed before sending means this
backup does not run, the reaper settles the row, and the scheduler creates a
fresh job on the next tick — strictly better than a duplicated child set that
corrupts the job ledger.

## Verification

- `backupWorker.multiTarget.test.ts` (new, 7 cases) — the missing 2-/3-target
  coverage plus every behaviour above. **5 of the 7 fail against the pre-fix
  worker** (confirmed by reverting `backupWorker.ts` to `origin/main` and
  re-running); the other 2 document existing correct multi-target behaviour that
  had no test.
- `backupEnqueue.test.ts` — pins `attempts: 1`/no-backoff for dispatch and
  `attempts: 3`/exponential for results (the dispatch assertion was red first).
- `vitest run src/jobs/backupWorker src/jobs/backupEnqueue` — 44 passed.
- `vitest run src/routes/backup src/services/backup src/jobs/staleCommandReaper`
  — 30 files, 483 passed.
- `docs/registers/queue-worker-register.md` retry-policy row updated.

No migration, no schema change, no tenancy surface touched.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — `code-reviewer` (correctness / state-machine focus) and `silent-failure-hunter` (failure-visibility focus), plus a read-only `codex` design review of the retry-semantics choice before implementation.

**Findings:** 3 raised → 3 addressed, 1 rejected with reasoning; 0 outstanding.

- **should-fix — the new Sentry capture was scrubbed to a contentless event.** `scrubEvent` deletes message/logentry/extra, rewrites every exception `value` to `[redacted]`, and drops any tag outside `ALLOWED_TAG_NAMES`. The `job_type`/`backup_job_id`/`device_id` tags on the redelivery-refusal capture were all dropped, so the one event that says "a backup run was deliberately dropped" arrived indistinguishable from a generic worker crash — the same trap `cas_label`, `db_pool_health_verdict` and `manifest_refusal_reason` were each added to fix. Fixed in f13594778: new closed-set tag `backup_dispatch_issue` (`redelivery-refused` | `undelivered-settle-failed`, no tenant/device/job ids), asserted at **both** gates the repo's existing tags are asserted at. Removing the allowlist entry turns both red. The same commit also reports a failure of `settleUndeliveredDispatchTargets` itself, which was console-only.
- **should-fix — "Phase 3 exceptions aren't settled": rejected, not a real gap.** `withDbAccessContext` (`db/index.ts`) runs its callback inside a real `baseDb.transaction(...)`, and `db` is a proxy resolving to the active transaction. Phase 3 is therefore one transaction: an exception mid child-insert loop rolls those inserts back, so there are no committed rows to settle. Only a *cancellation* returns normally and commits — which is exactly the path the `cancelCreatedChildren` fix covers.
- **nit — brittle cancellation test.** It hand-counted `isBackupJobCancelled` calls and fired on the 8th; adding a check anywhere in Phase 1 would have shifted which iteration the cancel landed on while the test kept passing. Fixed in 82eac4cb9: the trigger is now counted relative to the first committed child row, so it stays pinned to the loop under test.
- **nit — declined (pre-existing, out of scope).** The Phase-5 `failedChildJobs` settle loop has no status guard, unlike every write this PR adds. It is a context line, not a changed one, and its comment says the unconditional settle is deliberate (#4084). Not touched here.

**Tests (local, single-fork):**
- `vitest run src/jobs` — 161 files, **2032 passed**, 1 skipped.
- `vitest run src/routes/backup src/services/backup src/jobs/staleCommandReaper` — 30 files, **483 passed**.
- `vitest run src/jobs/backupWorker src/jobs/backupEnqueue src/services/sentry` — 10 files, **102 passed**.
- **Discrimination proven, not assumed:** reverting `backupWorker.ts` to `origin/main` turns **5 of the 7** new `backupWorker.multiTarget.test.ts` cases red (re-confirmed after the test rewrite); the `attempts: 1` assertion was written red-first; removing `backup_dispatch_issue` from the allowlist turns both Sentry gate tests red. The `whereValues` Drizzle-clause helper was checked non-vacuous (a `toHaveLength(0)` assertion here would otherwise pass for the wrong reason) — independently re-verified by the reviewer against the real schema columns.
- `tsc --noEmit` clean; `eslint` clean on every changed file.

**CI:** `CI Success` passed on head 3e0816155; the two review-fix commits are still running at the time of writing.

**Status:** review-clean, awaiting maintainer merge. Not merged, issue #4137 left open and assigned.

PR #4855, opened 2026-09-03
fix(api): enforce the site axis on DR plan, group and execution mutations (#3653)

### LINKED ISSUE #3653: Site-scope gap: DR plan execution dispatches destructive restores to out-of-site devices
From the 2026-08-16 security review, theme 1 (site scoping). **Verified against `main`.** Severity: **HIGH**.

## Why this one is unambiguous

The site-axis defense **already exists and was purpose-built for this service**:

- `services/drExecutionService.ts:75-84` documents `authorizedDeviceIds` as the site-scope snapshot.
- `services/drExecutionService.ts:374-390` (`dispatchGroup`) calls it "the ONLY defense against dispatching to an out-of-site device."
- `services/aiToolsDR.ts:346-371` — the **AI-tool caller** computes `resolveSiteDevicePartition`, rejects a plan containing out-of-scope devices, and pins `authorizedDeviceIds`.

The **HTTP route caller does not**. `routes/dr.ts:425-430` calls `createDrExecutionAndEnqueue({ planId, orgId, executionType, initiatedBy })` with no `authorizedDeviceIds` → `drExecutionService.ts:284` stores `null` → `:381` treats the dispatch as unrestricted.

`routes/dr.test.ts:218-223` asserts exactly that four-key call shape, freezing the gap in place.

## Exploit path

User in org O restricted to `allowedSiteIds: [siteA]`, holding `devices:execute` + MFA:

1. `GET /api/v1/dr/plans/:id` returns every plan in org O including `devices[]` ids for site B (`dr.ts:126-166`, org filter only).
2. `POST /api/v1/dr/plans/:id/execute {"executionType":"failover"}` → no `authorizedDeviceIds` passed → worker check no-ops on `null` → queues `VM_RESTORE_FROM_BACKUP` / `BMR_RECOVER` / `HYPERV_RESTORE` / `MSSQL_RESTORE` to site-B devices.

Also: `PATCH /plans/:id/groups/:gid` (needs `devices:write`) accepts arbitrary same-org out-of-site device ids — `allDevicesBelongToOrg` (`dr.ts:46-54`, used at `:277` and `:329`) validates the org axis only, by design per its own comment. And `dr.ts:521-559` lets the same user abort another site's in-flight recovery.

**RLS does not defend the site axis**, so there is no database backstop. `services/commandQueue.ts` has no site check either — `queueCommandForExecution` receives `expectedOrgId` only (`drExecutionService.ts:403`).

Not permission-gated away: `devices:read/write/execute` are role permissions orthogonal to `allowedSiteIds`; a site-restricted technician normally holds them.

## Fix shape

1. In `POST /plans/:id/execute`, mirror `aiToolsDR.ts:346-363` — resolve the caller's site partition, 403 if any group device is out of scope, and pass `authorizedDeviceIds` into `createDrExecutionAndEnqueue`.
2. Extend `allDevicesBelongToOrg` (or add a sibling) so group create/update rejects devices failing `canAccessSite`.
3. Site-filter the plan/group/execution read responses.
4. Update `routes/dr.test.ts:218-223`, which currently pins the vulnerable call shape.

ToddHebebrand:
Closes #3653

## What was actually still broken

The reported exploit — `POST /dr/plans/:id/execute` dispatching destructive restores to out-of-site devices — **has already been closed on `main`** by the recovery-authorization framework that landed after the issue was filed. `createDrExecutionAndEnqueue` now runs every group device through `authorizeQueuedRecoveryWork` → `authorizeResilienceResources`, which site-checks each ref and throws *before* an execution row is written. The issue's claim that `routes/dr.ts` calls the service with four keys and that `dr.test.ts:218-223` pins that shape is stale; both already pass `auth`.

The rest of the site-scope surface named in the issue was still open. Verified by watching each assertion fail on `origin/main` first:

| Route | Behaviour on `main` | Now |
|---|---|---|
| `DELETE /plans/:id/groups/:gid` on a group holding an out-of-site device | **200** | 403 |
| `PATCH /plans/:id` / `DELETE /plans/:id` on a plan spanning unreachable sites | **200** | 403 |
| `POST /executions/:id/abort` on another site's running recovery | **200** | 403 |
| `POST /plans/:id/execute` denied by the site barrier | **500** + Sentry event | 403 |
| `POST`/`PATCH` group with an out-of-site device | accepted | 403 |
| `manage_dr_plan` `update_plan` archiving a plan spanning unreachable sites | accepted | denied |
| `manage_dr_plan` `update_group`/`delete_group` dropping an out-of-site device | accepted | denied |

RLS enforces the org axis but has no concept of a site, so these routes are the only barrier. `devices:read/write/execute` are role permissions orthogonal to `allowedSiteIds` — a site-restricted technician normally holds them.

## The bypass that shaped the design

Validating only the **submitted** device array is insufficient. A site-restricted caller can `PATCH devices:[theirOwnDevice]` over a stored `[theirOwnDevice, otherSiteDevice]` and silently drop the out-of-site device from the recovery plan **without ever naming it** — degrading DR for a site they cannot see. `DELETE` removes it outright.

So every mutation that replaces or removes membership is authorized against **what is already stored**, not just what was sent. That is the `authorizeStoredDevices` / `authorizePlanStoredDevices` pair in `routes/dr.ts`, and `storedGroupDevicesDenied` / `planStoredDevicesDenied` in `aiToolsDR.ts`.

## Second write path, swept

`manage_dr_plan` (`services/aiToolsDR.ts`) also writes DR plans and groups. Its `deviceArgs: ['devices']` declaration already gets the central org+site gate on **submitted** ids, but that gate cannot see stored membership — so `update_group` and `delete_group` carried the identical bypass, and `update_plan` could archive (i.e. halt) a plan reaching sites the caller cannot access, with no guard at all. All three are now gated. `create_plan` writes no device data and `add_group` has no prior membership, so both are covered by the central gate alone.

## Design decisions

**Reads are deliberately left unredacted.** Filtering `groups[].devices` for site-restricted callers looks like the obvious completion of the issue's item 3, and it is a trap: `apps/web/src/components/dr/DRPlanEditor.tsx` loads the plan, populates each group's `deviceIds` from the response, and on Save **PATCHes every group with a full array replace**. Redaction would turn one click of Save into a silent strip of every out-of-site device from every group of a disaster-recovery plan. In DR, what you see must not diverge from what executes. Partial redaction would also be false assurance — execution `results` carries device ids in `groupResults`/`queuedCommands`/`failedDispatches`, and `restoreConfig` carries source-resource ids. The residual same-org device-id disclosure is recorded as accepted debt; the write, execute and abort axes are what this PR closes.

**`allDevicesBelongToOrg`'s 403 + message is preserved** for the org axis rather than being folded into the canonical `authorizeRouteResilienceResources` (which would return 404 for a foreign-org device). The canonical helper resolves refs one query at a time, and DR device arrays are unbounded; the new guards do it in a single bounded query instead. The org-axis status change would have been a contract change with no security benefit. Org mismatch is deliberately reported *ahead* of the site grant, so a site verdict never confirms that an id exists elsewhere in the org.

**The site-restricted principal-kind list is now exported once** from `resilienceSiteAuthorization.ts` and consumed by the route adapter, replacing an inline duplicate. A kind present in one copy and missing from the other is a silent hole in a boundary Postgres does not enforce.

**`siteRestriction()` does not fail open.** A missing `permissions` context value falls back to the token's own grant rather than reading as "unrestricted". Every route here sits behind `requirePermission`, which always publishes it, so this is not reachable today — but a route added later without that middleware would otherwise inherit a silent fail-open.

## Cost

Unrestricted callers are unchanged and issue **no** additional query — the stored-membership guards return before touching the database. A site-restricted caller pays at most one bounded device lookup per mutation.

## Verification

- Written red-first: the seven route assertions above failed on `origin/main` with the statuses in the table.
- Every guard was confirmed to **fail its test when deleted**, run in isolation to avoid mock-queue bleed between tests. This caught a real hole: the group PATCH *stored* guard was masking the *proposed* guard, so the latter could be removed with every test still green. That case now has its own test.
- Full affected set (dr, drExecutionService, aiToolsDR, resilienceSiteAuthorization, recoveryAuthorizationSubject, routes/backup, drExecutionWorker, aiTools device-gate contract): **28 files, 400 tests passed**.
- Edge cases pinned: `allowedSiteIds: []` (granted zero sites) denies; a stored id whose device row is gone does not wedge maintenance of its group; duplicate and non-string ids are normalised.
- `tsc --noEmit` clean, `eslint` clean on all touched files.
- No schema change, so no RLS/cascade/export-policy registration applies.

## Not in scope

- Per-device `addDeviceIds`/`removeDeviceIds` group operations. The current PATCH mutates the whole aggregate, so a partially-authorized caller cannot safely operate it; finer-grained membership editing is the right long-term answer for a site-restricted technician who legitimately maintains part of a plan.
- Plan groups remain editable while an execution is reconciling, so an edit mid-flight can still change what a running execution targets.
- Read-path site filtering, per the design decision above.

### DIFF
--- a/apps/api/src/routes/backup/resilienceAuthorization.ts
+++ b/apps/api/src/routes/backup/resilienceAuthorization.ts
@@ -6,6 +6,7 @@ import type { AuthContext } from '../../middleware/auth';
 import {
   ResilienceAuthorizationError,
   authorizeResilienceResources,
+  isSiteRestrictedPrincipalKind,
   type AuthorizedResilienceResources,
   type ResilienceOperation,
   type ResilienceResourceRef,
@@ -64,15 +65,7 @@ export async function resolveRouteAuthorizedDeviceIds(
 ): Promise<string[] | null> {
   const auth = c.get('auth') as AuthContext;
   const permissions = c.get('permissions') as UserPermissions | undefined;
-  const isSiteRestrictedPrincipal = [
-    'user_session',
-    'client_user',
-    'api_key',
-    'oauth_grant',
-    'ai_agent',
-  ].includes(auth.principal.kind);
-
-  if (!isSiteRestrictedPrincipal || !permissions?.allowedSiteIds) return null;
+  if (!isSiteRestrictedPrincipalKind(auth.principal.kind) || !permissions?.allowedSiteIds) return null;
   if (permissions.allowedSiteIds.length === 0) return [];
 
   const rows = await db
--- a/apps/api/src/routes/dr.test.ts
+++ b/apps/api/src/routes/dr.test.ts
@@ -748,6 +748,89 @@ describe('dr routes', () => {
       expect((await res.json()).data.status).toBe('aborted');
     });
 
+    // The stored guard runs first, so a group whose stored membership is fully
+    // in-site must still have its PROPOSED additions checked. Without this the
+    // proposed guard could be deleted and every other test would stay green.
+    it('rejects a group update that adds an out-of-site device', async () => {
+      restrictToSiteA();
+      selectMock.mockReturnValueOnce(chainMock([{
+        id: GROUP_ID,
+        planId: PLAN_ID,
+        orgId: ORG_ID,
+        devices: [DEVICE_ID],
+      }]));
+      selectMock.mockReturnValueOnce(chainMock([{ id: DEVICE_ID, siteId: SITE_A }]));
+      selectMock.mockReturnValueOnce(chainMock([{ id: OUT_OF_SITE_DEVICE_ID, siteId: SITE_B }]));
+
+      const res = await app.request(`/dr/plans/${PLAN_ID}/groups/${GROUP_ID}`, {
+        method: 'PATCH',
+        headers: { 'Content-Type': 'application/json', Authorization: 'Bearer token' },
+        body: JSON.stringify({ devices: [OUT_OF_SITE_DEVICE_ID] }),
+      });
+
+      expect(res.status).toBe(403);
+      expect((await res.json()).error).toBe('site_access_denied');
+      expect(updateMock).not.toHaveBeenCalled();
+    });
+
+    // An empty grant is the most locked-down state and the one a truthiness
+    // slip (`!allowedSiteIds`) would silently turn into "unrestricted".
+    it('denies a technician granted zero sites', async () => {
+      permissionsState = { allowedSiteIds: [] };
+      selectMock.mockReturnValueOnce(chainMock([{ id: GROUP_ID, devices: [DEVICE_ID] }]));
+      selectMock.mockReturnValueOnce(chainMock([{ id: DEVICE_ID, siteId: SITE_A }]));
+
+      const res = await app.request(`/dr/plans/${PLAN_ID}/groups/${GROUP_ID}`, {
+        method: 'DELETE',
+        headers: { Authorization: 'Bearer token' },
+      });
+
+      expect(res.status).toBe(403);
+      expect((await res.json()).error).toBe('site_access_denied');
+      expect(deleteMock).not.toHaveBeenCalled();
+    });
+
+    // authorizeStoredDevices deliberately omits the row-count parity check that
+    // authorizeProposedDevices has: an id whose device row is gone cannot be a
+    // restore target, and must not wedge maintenance of the group holding it.
+    it('does not let a deleted device block maintenance of its group', async () => {
+      restrictToSiteA();
+      const deletedDeviceId = '44444444-4444-4444-8444-444444444444';
+      selectMock.mockReturnValueOnce(chainMock([{
+        id: GROUP_ID,
+        devices: [DEVICE_ID, deletedDeviceId],
+      }]));
+      selectMock.mockReturnValueOnce(chainMock([{ id: DEVICE_ID, siteId: SITE_A }]));
+      deleteMock.mockReturnValueOnce(chainMock([]));
+
+      const res = await app.request(`/dr/plans/${PLAN_ID}/groups/${GROUP_ID}`, {
+        method: 'DELETE',
+        headers: { Authorization: 'Bearer token' },
+      });
+
+      expect(res.status).toBe(200);
+      expect(deleteMock).toHaveBeenCalled();
+    });
+
+    // A device outside the org must report the org mismatch, never the site
+    // grant — the site verdict would confirm the id exists somewhere in-org.
+    it('reports an org mismatch ahead of the site grant', async () => {
+      restrictToSiteA();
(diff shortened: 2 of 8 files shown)


ToddHebebrand:
**Review run:** two Sonnet reviewers over the diff — a security lens (fail-open hunt, unguarded-route enumeration, refactor-regression check on the shared `resilienceSiteAuthorization` helper, error-classifier masking) and a test-quality lens (guard-deletion sensitivity, verified empirically by removing each guard and re-running), plus a third focused re-review of the follow-up commit because it touched the authorization surface itself.

**Findings:** 2 HIGH, 1 MEDIUM, 2 LOW raised → all addressed or explicitly accepted; 0 outstanding.
- **HIGH (security)** — `manage_dr_plan`'s `update_plan` had no site guard, so a site-restricted caller could archive (halt) a plan reaching sites they cannot access — the AI-tool twin of the `PATCH /plans/:id` guard. Fixed in `ec10a673a` via `planStoredDevicesDenied`.
- **HIGH (tests)** — the group PATCH *stored* guard was masking the *proposed* guard: deleting the latter left every committed test green. Test added; confirmed to fail with the guard removed.
- **MEDIUM (security)** — read paths still return out-of-site device ids. **Accepted, not fixed**: `DRPlanEditor.tsx` round-trips `groups[].devices` through a full-array-replace PATCH, so redacting reads would turn one click of Save into a silent strip of every out-of-site device from the plan. Rationale in the PR body.
- **MEDIUM (tests)** — `allowedSiteIds: []` and the deliberate deleted-device asymmetry were untested. Both now pinned.
- **LOW** — org-vs-site denial ordering (accepted; org mismatch is reported first so a site verdict never confirms an id exists elsewhere in the org) and a leftover optional chain (fixed in `937155e8d`).

Also hardened beyond the findings: `siteRestriction()` no longer reads a missing `permissions` context value as "unrestricted" — it falls back to the token's own grant, matching `authorizeRouteResilienceResources`. Not reachable today (every route sits behind `requirePermission`), but it was a fail-open shape on a boundary Postgres does not enforce.

**Tests:** 28 files / 400 tests passed locally (`vitest run` over dr, drExecutionService, aiToolsDR, resilienceSiteAuthorization, recoveryAuthorizationSubject, routes/backup, drExecutionWorker, aiTools device-gate contract). Written red-first — the seven route assertions failed on `origin/main` with the statuses tabulated in the PR body. Every guard verified to fail its test when deleted, run in isolation to avoid mock-queue bleed. `tsc --noEmit` clean; `eslint` clean on all touched files.

**Status:** review-clean, CI still running on `937155e8d` at the time of writing — please confirm green before merging. Awaiting maintainer merge; the issue stays open until the fix is verified.

Issue #1448, opened 2026-06-16
Stripe Checkout HTTP call held inside the request DB transaction (pay-link + portal /pay)


ToddHebebrand:
**Type:** follow-up (non-blocking) from the Catalog/Invoices/Contracts review (#1442).

### Problem
`createInvoicePayLink` (`apps/api/src/services/invoiceCheckout.ts`) and its twin `POST /portal/invoices/:id/pay` (`apps/api/src/routes/portal/invoices.ts`) call `getStripe().checkout.sessions.create(...)` — a ~hundreds-of-ms HTTP round-trip — **inside the request DB transaction**.

The main auth middleware wraps the whole handler in `withDbAccessContext` → `baseDb.transaction(tx => …)` (`db/index.ts`), so one pooled PG connection is held *idle-in-transaction* for the duration of the Stripe HTTP call. This is the #1105 class of pool-poison risk (though far milder here: a one-shot, low-frequency action vs the worker/loop workload that actually poisoned the pool).

### Why it's not a trivial fix
- `runOutsideDbContext` only swaps the `db` proxy's AsyncLocalStorage reference (pool vs tx) — it does **not** release the outer `baseDb.transaction` connection, which is held for the whole handler regardless.
- Relocating the mapping `INSERT` outside a DB access context would trip the #1375 contextless-write guard (forced-RLS `breeze_app` silently matches 0 rows → the write no-ops).

### Proposed approach (own PR)
Make these two pay routes opt out of the auto request-transaction and manage their own short contexts: read invoice+connection in a `withSystemDbAccessContext` (or request ctx), run `checkout.sessions.create` truly outside any tx via `runOutsideDbContext`, then write the pending `invoice_stripe_payments` mapping in a fresh short `withSystemDbAccessContext`. Fix both call sites together; add a test that the connection isn't held across the HTTP call.

### Affected
- `apps/api/src/services/invoiceCheckout.ts`
- `apps/api/src/routes/portal/invoices.ts`

PR #4856, opened 2026-09-03
fix(api): send quote emails after the transaction commits, and bound the mail transports (#3905)

### LINKED ISSUE #3905: sendQuote holds the request transaction (and a revision's parent row lock) across PDF render + an untimed mail round-trip
`sendQuote` renders the PDF and performs the outbound mail round-trip **inside the request-wide transaction**, so every row it has written — including the `FOR UPDATE` lock on a revision's parent quote — stays locked until the mail transport returns. Neither transport has a timeout, so a stalled mail server holds a pooled Postgres connection and a customer-facing row lock for as long as it wants to.

The code already names this: `apps/api/src/services/quoteLifecycle.ts:454` calls it "a tracked follow-up (atom-3)" — but no issue existed, so this is that issue.

### The path

`deliverQuoteEmail` (`quoteLifecycle.ts:445`) is called at `:347` from `sendQuote`, which runs inside the transaction opened by `withDbAccessContext`. By the time it is reached the transaction already holds:

- the `draft → sent` claim on the quote being sent,
- on a revision, the `FOR UPDATE` lock on the **parent** taken by `loadDraft` (`:88`), plus the parent's `status='superseded'` + `public_link_revoked_at` write (`:296`),
- the `quote_recipients` inserts.

It then does, still inside that transaction:

1. `renderQuotePdf` + `mergeUploadedContractPdfs` — headless render, unbounded,
2. `emailService.sendEmail` — the network round-trip.

### Why the transports can't be trusted to return

- **SMTP** (`services/email.ts:191`): `nodemailer.createTransport` is configured with host/port/secure/auth and **no `connectionTimeout`, `greetingTimeout`, or `socketTimeout`** — so it inherits nodemailer's 10-minute socket default. A silently-dropping mail server holds the transaction for ten minutes.
- **Mailgun** (`services/email.ts:657` and `:677`): both `fetch` calls pass no `AbortSignal`, so there is no client-side deadline at all.

### Impact

Two distinct failures, the second worse than the first:

1. **Connection-pool pressure on every send** (all quotes, not just revisions). A pooled connection is pinned for the full render + mail duration. This is the exhaustion shape from #1105.
2. **The parent's customer-facing surface goes dead while a revision is being sent** (revisions only). The parent row is locked, so the customer's own `POST /accept` on the original blocks behind our mail server — as do the parent's resend, share-link, and view-stamp writes. Each blocked request pins another pooled connection, which is how #1105 compounded.

Not hypothetical for revisions specifically: the whole point of supersede-at-send is that the parent's link is still live and in a customer's inbox at the moment we send the child.

### Contrast: the invoice path already does this correctly

`contractService` returns a **deferred** and the caller emails **after commit**. `quoteLifecycle` is the odd one out — the comment at `:451` says so explicitly.

### Fix

Commit the state transition, then mail. Two viable shapes:

- **A — deferred, matching the invoice path.** `sendQuote` returns a thunk; the route runs it after `withDbAccessContext` commits. Smallest change, keeps the "did it email?" answer in the same request.
- **B — hand it to the existing `quote-send` queue** (`jobs/quoteSendQueue.ts` already renders + sends for scheduled sends). Strongest isolation, and the retry/observability story comes free — but the direct-send response no longer knows whether mail left, so the UI has to learn to read `send_email_reason` asynchronously.

**Recommend A** — it removes the lock-holding without changing what the send endpoint can tell the user, and B stays available later.

Either way, add explicit transport deadlines (`connectionTimeout` / `greetingTimeout` / `socketTimeout` on the SMTP transport; `AbortSignal.timeout(...)` on both Mailgun `fetch` calls) — a post-commit send that hangs forever is still a leaked worker.

### Notes

- The existing swallow-and-record behaviour (`send_email_reason`) must survive the move: a send that commits with mail unsent already renders the "no email was delivered" banner (#3502), and that contract is load-bearing.
- Found while building quote revisions (#3796, W03 #3873). Flagged rather than folded into a wave because it reshapes the send path, which is well outside revision scope.

ToddHebebrand:
## The bug

`sendQuote` rendered the proposal PDF and ran the outbound mail round-trip **inside the request transaction**. Everything that transaction had written stayed locked until the transport returned — including, on a revision, the `FOR UPDATE` lock on the **parent** quote.

Two failures, the second worse:

1. **Connection-pool pressure on every send.** A pooled connection pinned idle-in-transaction for the full render + mail duration (#1105 shape).
2. **The parent's customer-facing surface went dead while a revision was being sent.** The parent's public accept link is still live in a customer's inbox at that moment — that is the whole point of supersede-at-send — so their own `POST /accept` blocked behind our mail server, and each blocked request pinned another connection.

Neither transport had a client-side deadline, so "stalled" had no upper bound: nodemailer inherited its 10-minute socket default and both Mailgun `fetch` calls passed no `AbortSignal` at all.

The code already named this (`quoteLifecycle.ts` — "a tracked follow-up (atom-3)"); the invoice path has done it correctly since `contractService` started handing issue+email back to its caller to run post-commit.

## The fix — option A from the issue

**Commit the state transition, then mail.**

- `sendQuote`/`resendQuote` no longer email. They return `deliverEmail`, a `DeferredQuoteEmail` the caller runs once its transaction has **committed**. Removing `emailed`/`emailReason` from both result types is what forced the caller sweep instead of letting a stale call site silently keep the old behaviour.
- **`POST /quotes/:id/send` and `/:id/resend` are registered in `SELF_MANAGED_DB_CONTEXT_ROUTES` (#1448).** This is load-bearing: the auth middleware otherwise wraps the entire handler in one transaction, and `runOutsideDbContext` does **not** close it — it only re-points the ALS `db` proxy. Without the registration the fix would look right and change nothing.
- **The deferred owns the `send_email_reason` write**, in its own short context. That removes the #3502 footgun where each of three call sites had to remember to persist the returned reason (a fourth that forgot would mark the quote sent with the column `NULL`, so the "no email was delivered" banner never fires). Post-commit it swallows its own failure — the inverse of the old in-transaction rule, and deliberately so: throwing would report a *committed* send as a failure the tech retries into a 409.
- **`deliverQuoteEmail` is split into three phases** so the transport holds no connection at all: DB reads + pdfkit render in one short context → the mail call outside any context → the outcome write in another short context. (pdfkit, not a headless browser, so no network I/O hides in phase 1.)
- `runBulkIsolated` gains an `afterItemCommit` hook so `bulk-send` delivers each item's email after that item's own transaction commits.
- The scheduled-send worker runs the deferred after its system context resolves, **gated on the transaction having succeeded** — `deliverEmail` is captured inside the transaction, so a failure at COMMIT would otherwise mail a proposal for a quote still in draft.

### Transport deadlines

- `SMTP_TIMEOUT_MS` (default **30 000**) → `connectionTimeout` / `greetingTimeout` / `socketTimeout`. An *inactivity* timeout, so it does not cap a slow attachment upload.
- `MAILGUN_TIMEOUT_MS` (default **120 000**) → `AbortSignal.timeout` on both fetches, translated into `Mailgun request timed out after Nms` (a bare `TimeoutError` names neither the transport nor the budget). Larger default because this one *does* include the multipart upload.
- A malformed value is **rejected at startup** (the API reports email as not configured) rather than silently restoring an unbounded transport. Both are documented in `.env.example` and mapped through all three compose files.

## Deliberate residual

The **AI-tool** `send` path still delivers inside the tool transaction. `aiAgentSdkTools.ts` wraps the whole `executeTool` call in one transaction and offers no post-commit seam, and moving the delivery to a separate connection there would **deadlock** on the `FOR UPDATE` lock the tool still holds. Behaviour is identical to before this PR, and it is commented at the call site. Fixing it properly means giving `executeTool` a post-commit hook.

## Evidence

**`apps/api/src/__tests__/integration/quoteSendLockRelease.integration.test.ts`** (new, real Postgres) probes the parent row from a **second connection** with `FOR UPDATE NOWAIT` *from inside the transport call*:

| | result |
|---|---|
| negative control (transaction holds the lock) | `locked` |
| this PR | `acquired` |
| ordering reverted to pre-fix (mutation check) | `locked` — test fails |

It also proves the parent already reads `superseded` from that second connection at transport time (the transition really committed, the email was not merely skipped), and that a post-commit delivery failure still lands `send_email_reason = 'send_failed'` on the row.

Unit side: route-level ordering tests assert the literal interleaving `['tx:begin', 'sendQuote', 'tx:commit', 'deliverEmail']` (mutation-checked — awaiting the deferred inside the transaction callback turns them red while every other test in the file still passes); deferred-contract tests cover idempotence, the swallowed reason-write failure, and no DB touch when no email service is configured; `email.test.ts` pins the three SMTP deadlines and an `AbortSignal` on both Mailgun fetches.

- Affected unit suites: **543 passed / 16 files** (single fork).
- Integration: **5 passed** (`quoteSendLockRelease`).
- `tsc --noEmit` clean; `eslint` clean on every changed file.

## Risk notes for the reviewer

- The two newly self-managed routes run with **no ambient RLS context**. `requirePermission` already handles that (`getUserPermissions` escalates per axis — the #1448 pay-link path), `writeAuditEvent` opens its own context, and the handlers do no other DB work. `selfManagedDbContextRoutes.test.ts` gains MATCH cases plus NO_MATCH cases proving the patterns do not swallow `bulk-send`, `schedule-send`, `share-link`, or `GET /send`.
- No migration, no schema change, no new table — cascade/export-policy registries are untouched.
- The response shape of both routes is byte-identical to before.

Closes #3905


ToddHebebrand:
**Review run:** `pr-review-toolkit` — `code-reviewer` (transaction/lock restructuring, self-managed-route registration, RLS scope capture, overlay semantics, transport defaults) + `silent-failure-hunter` (swallowed errors, the "never rejects" contract, the #3502 `send_email_reason` path).

**Findings:** 3 raised → all 3 addressed.

1. **The delivery swallow was narrowed** (found in self-review, independently confirmed by `silent-failure-hunter`). The refactor moved `getEmailService()` and the customer-visible line filtering *outside* `deliverQuoteEmail`'s outer `try`, which pre-#3905 wrapped the whole body. A throw there would escape the deferred and break the `never rejects` contract three call sites depend on — 500ing an already-**committed** send (retried into a 409, or a duplicate customer email on the re-send path), and in `quoteSendQueue` skipping the outcome write entirely, reproducing the exact #3502 class this PR exists to fix. Fixed in `dede3f52b`: catch scope restored, plus a last-resort catch in `makeDeferredQuoteEmail` (the promise is memoised, so a rejection there would be permanent for that quote). Regression test verified red against `316b90968`.
2. **A malformed email env value disabled all outbound mail invisibly.** `getEmailService` caches `unavailable` for the process lifetime and only `console.warn`s — and `parseTransportTimeoutMs` adds a new way to trip it (`SMTP_TIMEOUT_MS=30s`). Fixed in `5944df946`: `EmailConfigValueError` marks an operator typo and is Sentry-captured; an *absent* var stays log-only, because self-hosting without email is supported and capturing it would flood those installs. Both halves are asserted, and the capture is mutation-verified.
3. **AI-tool residual** — `code-reviewer` correctly notes this path does not get the fix, and that an AI agent can reach `quotes:send` autonomously under tier-2/3 auto-exec. Not silently introduced (it is the PR body's "Deliberate residual"), and the local "use a second connection" fix is *wrong* — this transaction still holds the row lock. It is now **bounded** by the new transport deadlines rather than open-ended. The real fix is a post-commit hook on `executeTool`, a cross-tool contract change that belongs in its own PR; the call-site comment now records all of this. **Flagged for a maintainer decision on whether to open a follow-up issue.**

Clean with no findings: self-managed-route registration (middleware ordering, `requirePermission`'s existing no-ambient-context escalation, regexes vs. sibling routes, mount paths), RLS scope capture and re-entry, `pdf_render_failed` vs `send_failed` mapping parity, overlay staleness across all callers, transport defaults.

**Tests:** affected API suites **546 passed / 16 files** (`vitest run --pool=forks --max-workers=1`); integration **5 passed** (`quoteSendLockRelease`) plus 8 neighbouring quote/bulk integration suites green (`quoteLifecycle`, `quoteRevisions`, `documentLocaleStamping`, `quoteAccept`, `quoteDeposit`, `quoteContractAccept`, `quoteSendQueue`, `bulkOpsIsolation`). `tsc --noEmit` clean; `eslint` clean on every changed file.

**Mutation-verified (not vacuous):** the route ordering test goes red when `deliverEmail()` is awaited inside the transaction callback; the integration probe reports `locked` instead of `acquired` under the pre-fix ordering (with a negative control proving the probe sees a held lock); the never-rejects test is red against `316b90968`; the Sentry capture test is red with the capture removed.

**Status:** review-clean, awaiting maintainer merge. No migration, no schema change; response shapes unchanged.

PR #4858, opened 2026-09-03
fix(api): accounting captureException tags dropped by Sentry allowlist (#4828)

### LINKED ISSUE #4828: [API] accountingInvoicePush.ts captureException tags are silently dropped by the Sentry allowlist
## Description

Every `captureException(err, undefined, {...})` call in `apps/api/src/services/accounting/accountingInvoicePush.ts` passes tag keys (`service`, `invoiceId`, `mappingId`, `partnerId`, `remoteEntityId`) that are NOT present in `ALLOWED_TAG_NAMES` (`apps/api/src/services/sentry.ts`). `setCallerTags` only forwards allowlisted keys, and `scrubEvent` strips `message`/`extra`/`logentry` from the outbound event — so per that file's own doc comment, tags are "the ONLY channel that reaches Sentry" for these calls. Every one of these best-effort failure reports currently arrives in Sentry as a near-contentless event with no way to identify which invoice, mapping, or partner it happened for.

Found during review of PR #4827 (issue #4498), which added a 7th instance of the same pattern (the new `persistInvoiceCurrencyMismatchErrorInOwnContext` function) rather than introducing it — this predates that PR.

## Affected call sites (apps/api/src/services/accounting/accountingInvoicePush.ts)

- Line ~317 (new in #4827): `service`, `invoiceId`, `partnerId`
- Line ~375, ~380, ~402: `service`, `mappingId`, `partnerId`
- Line ~783: `service`, `mappingId`, `invoiceId`
- Line ~806: `service`, `mappingId`, `remoteEntityId`
- Line ~927: `service`, `mappingId`, `invoiceId`

Note `partner_id` (snake_case) IS already in `ALLOWED_TAG_NAMES` — only the camelCase `partnerId` key used here misses it. `invoiceId`/`mappingId`/`remoteEntityId`/`service` have no allowlisted equivalent at all.

## Proposed fix

1. Rename all `partnerId` tag keys in this file to the already-allowlisted `partner_id`.
2. Decide, with the same rigor `sentry.ts`'s existing allowlist comments use (closed/bounded set vs. unbounded cardinality vs. tenant-identifier precedent already set by `org_id`/`partner_id`/`user_id`), whether `invoice_id` and/or `accounting_mapping_id` should be added to `ALLOWED_TAG_NAMES`. Both are unbounded UUIDs like `partner_id`/`org_id`, which are already allowed for tenant-scoped triage, so there's a reasonable precedent — but this needs its own sign-off given the file's established bar for allowlist additions.
3. Update all affected call sites in `accountingInvoicePush.ts` (and audit `accountingMappingService.ts`'s `markMappingError`/nested-sync capture sites for the same pattern, since they follow an identical taxonomy) to use only allowlisted keys.

## Impact

Observability-only — does not affect correctness of the underlying sync logic, but makes production Sentry events for accounting-sync failures effectively untriageable (on-call cannot tell which invoice/partner failed without cross-referencing logs).

ToddHebebrand:
## Summary

Every `captureException(err, undefined, { ...tags })` call in `apps/api/src/services/accounting/accountingInvoicePush.ts` (and the identical-taxonomy sites in `accountingMappingService.ts`) tagged `service`, `invoiceId`/`mappingId`/`partnerId`/`remoteEntityId` — none of which were in `sentry.ts`'s `ALLOWED_TAG_NAMES` (only the snake_case `partner_id` existed, and `service` had no entry at all). `setCallerTags` only forwards allowlisted keys and `scrubEvent` strips `message`/`extra`/`logentry` from the outbound event, so every one of these best-effort accounting-sync failure reports has been arriving in Sentry as a near-contentless event — on-call could see a sync failed but not which invoice, mapping, or partner it happened for.

## Changes

- **`apps/api/src/services/sentry.ts`**: add `service`, `invoice_id`, `accounting_mapping_id`, `remote_entity_id`, `breeze_entity_type` to `ALLOWED_TAG_NAMES`, with a doc comment following the file's existing precedent style. `invoice_id`/`accounting_mapping_id`/`remote_entity_id` follow the same unbounded-UUID/opaque-identifier precedent already established by `org_id`/`partner_id`/`user_id`; `service` and `breeze_entity_type` are closed sets of hardcoded literals.
- **`accountingInvoicePush.ts`**: renamed all 7 `captureException` call sites' tag keys to the allowlisted snake_case equivalents (`partnerId`→`partner_id`, `invoiceId`→`invoice_id`, `mappingId`→`accounting_mapping_id`, `remoteEntityId`→`remote_entity_id`).
- **`accountingMappingService.ts`**: per the issue's explicit ask to audit `markMappingError`/nested-sync capture sites for the same pattern — fixed all 5 sites using the identical taxonomy. Dropped the unallowlisted `remoteSyncToken` tag from one call site rather than adding a new allowlist entry for it (not needed to identify the failing record — `accounting_mapping_id`/`remote_entity_id` already do that).
- **`sentry.test.ts`**: added a source-grep contract test (`describe('accounting captureException tags stay allowlisted (#4828)')`) that extracts every `captureException(...)` tag key from both files and asserts each is in `ALLOWED_TAG_NAMES` — a future call site with a new, not-yet-allowlisted key now fails CI immediately instead of shipping another silent drop.

## Out of scope

The issue and this fix are scoped to `accountingInvoicePush.ts` + `accountingMappingService.ts`, per the issue text. The same silently-dropped-tag pattern also exists in `accountingConnectionService.ts` (`module`/`op`/`connectionId`), `accountingPaymentPull.ts` (`action`/`resourceId`/`remotePaymentId`/`remoteInvoiceId`), and `quickbooksProvider.ts` (`op`/`entity`/`skippedDays`) — each uses a different tag vocabulary that would need its own allowlist precedent write-up, so those are left as a follow-up rather than folded into this fix.

## Test plan

- `npx vitest run src/services/accounting/accountingInvoicePush.test.ts src/services/accounting/accountingMappingService.test.ts src/services/sentry.test.ts` — 142 passed (includes updating one pre-existing assertion in `accountingMappingService.test.ts` that checked the old camelCase `remoteEntityId` key, and the new grep contract test).
- `npx tsc --noEmit` — clean.
- `npx eslint` on all changed files — clean.

Closes #4828

### DIFF
--- a/apps/api/src/services/accounting/accountingInvoicePush.ts
+++ b/apps/api/src/services/accounting/accountingInvoicePush.ts
@@ -314,7 +314,7 @@ async function persistInvoiceCurrencyMismatchErrorInOwnContext(
     });
   } catch (err) {
     captureException(err instanceof Error ? err : new Error(String(err)), undefined, {
-      service: 'accountingInvoicePush', invoiceId, partnerId,
+      service: 'accountingInvoicePush', invoice_id: invoiceId, partner_id: partnerId,
     });
   }
 }
@@ -372,12 +372,12 @@ async function markInvoiceMappingError(mappingId: string, partnerId: string, mes
       captureException(
         new Error(`markInvoiceMappingError matched no accounting_entity_mappings row (id=${mappingId})`),
         undefined,
-        { service: 'accountingInvoicePush', mappingId, partnerId },
+        { service: 'accountingInvoicePush', accounting_mapping_id: mappingId, partner_id: partnerId },
       );
     }
   } catch (err) {
     captureException(err instanceof Error ? err : new Error(String(err)), undefined, {
-      service: 'accountingInvoicePush', mappingId, partnerId,
+      service: 'accountingInvoicePush', accounting_mapping_id: mappingId, partner_id: partnerId,
     });
   }
 }
@@ -399,7 +399,7 @@ async function markInvoiceMappingErrorInOwnContext(
     await runInDbContext(() => markInvoiceMappingError(mappingId, partnerId, message));
   } catch (err) {
     captureException(err instanceof Error ? err : new Error(String(err)), undefined, {
-      service: 'accountingInvoicePush', mappingId, partnerId,
+      service: 'accountingInvoicePush', accounting_mapping_id: mappingId, partner_id: partnerId,
     });
   }
 }
@@ -780,7 +780,7 @@ export async function pushInvoiceToAccounting(
   } catch (err) {
     const message = sanitizeInvoiceSyncErrorMessage(err);
     captureException(err instanceof Error ? err : new Error(String(err)), undefined, {
-      service: 'accountingInvoicePush', mappingId: mappingRow.id, invoiceId: inv.id,
+      service: 'accountingInvoicePush', accounting_mapping_id: mappingRow.id, invoice_id: inv.id,
     });
     // Phase 2 (failure) — own short context so the marker COMMITS before the throw.
     await markInvoiceMappingErrorInOwnContext(runInDbContext, mappingRow.id, partnerId, message);
@@ -803,7 +803,7 @@ export async function pushInvoiceToAccounting(
     }));
   } catch (dbErr) {
     captureException(dbErr instanceof Error ? dbErr : new Error(String(dbErr)), undefined, {
-      service: 'accountingInvoicePush', mappingId: mappingRow.id, remoteEntityId: result.id,
+      service: 'accountingInvoicePush', accounting_mapping_id: mappingRow.id, remote_entity_id: result.id,
     });
     const message = `QuickBooks accepted the invoice sync (remote id ${result.id}) but Breeze failed to record it — do not retry; contact support to reconcile`;
     // Best-effort: the row is currently stuck at sync_status='pending' with no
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter, pr-test-analyzer — 3 parallel Sonnet agents)

**Findings:** 2 confirmed → both addressed in 6cff05f6c; 0 outstanding.
- silent-failure-hunter (Important): `accountingMappingService.ts`'s post-upsert-persist failure path dropped the `remoteSyncToken` tag entirely instead of allowlisting it — it's QuickBooks' optimistic-concurrency version counter, the specific value the "contact support to reconcile" manual-recovery path needs. Fixed: added `remote_sync_token` to `ALLOWED_TAG_NAMES` with justification, restored the tag.
- code-reviewer (Important): the new source-grep contract test's tag-extraction regex had a blind spot — it lazily matched the first `{` after `captureException(`, which landed on the `${mappingId}` interpolation inside `markInvoiceMappingError`/`markMappingError`'s error-message template literal instead of the real tags object, silently skipping that call site's tags from coverage. I independently verified this by running the regex directly against the source (a third reviewer, pr-test-analyzer, incorrectly claimed the regex was fine — verified by hand that it was not). Fixed: anchored extraction on the literal `undefined,` second argument, and added a per-file expected-call-count assertion so a future blind spot fails loudly as a count mismatch.
- pr-test-analyzer: no findings beyond the shared regex point above (which it assessed differently from code-reviewer); confirmed the contract test fails against pre-fix `origin/main` sources and passes against the fix, i.e. it's a real regression guard.

**Tests:** `apps/api` affected suite green (`npx vitest run src/services/accounting/accountingInvoicePush.test.ts src/services/accounting/accountingMappingService.test.ts src/services/sentry.test.ts` — 142 passed); `npx tsc --noEmit` clean; `npx eslint` on all changed files clean.

Issue #4067, opened 2026-08-26
[Auth][SSO] Link-on-first-SSO-login: sso_link_required is a dead end under enforce_sso — offer password confirmation to link inline


ToddHebebrand:
## Description

A password-holding user who completes SSO against a provider their account isn't linked to gets bounced to `/login?error=sso_link_required` with the banner: *"This account already has a password. Sign in with your password, then connect SSO under Profile → Security."*

Two problems:

1. **Under `enforce_sso=true` that instruction is impossible** — `assertPasswordAuthAllowedBySso` (`apps/api/src/routes/auth/ssoPolicy.ts`) blocks password login for every user in the tenant. The banner tells the user to do the one thing the tenant config forbids. Result: hard lockout for every unlinked user.
2. Even without enforcement, the ceremony is needlessly long: SSO attempt → read banner → password login → find Profile → Security → Connect SSO → full IdP round-trip again.

This has now caused **two full-tenant lockouts for the same partner** (US hosted: 2026-08-17 and 2026-08-26). Both times remediation was a manual prod UPDATE of `enforce_sso=false`.

## Root Cause

The callback's refusal to auto-link a password-holding account is correct and must stay — it's the #2183 account-takeover guard (a malicious/misconfigured IdP can assert a victim's email; only the user can prove they own both credentials). The defect is that the refusal is a **dead end** instead of the start of a linking ceremony.

## Proposed Fix

Link-on-first-SSO-login (the GitHub/Google pattern — same two proofs as today's Profile → Security flow, opposite order):

- When the callback resolves an email-matched, password-holding, unlinked account, don't bounce to `/login?error=sso_link_required`. Instead stash the **verified** IdP identity `(provider_id, sub, email)` in a short-lived single-use pending record (Redis, ~5 min, same idiom as `mfa:pending:*`) and redirect to a "Connect your sign-in" page.
- That page asks for the account password only. On success: create the `user_sso_identities` link, complete login with the normal SSO token mint (including the `trustsIdpMfa`/`amr:mfa` → `mfa` claim evaluation from the callback).
- The password verification inside this ceremony is **exempt from `assertPasswordAuthAllowedBySso`** — it only ever runs downstream of a successful IdP assertion, so it is not a password-login bypass. Rate-limit it like login (`loginLimiter`), audit both outcomes.
[...]

PR #4860, opened 2026-09-03
feat(auth): surface org-level enforceSSO on the login page (#3229)

### LINKED ISSUE #3229: Nginx reverse-proxy doc omits /portal route; org-level enforceSSO not reflected in login UI
Reported by **skillday_** in #bugs (Discord, message 1535150269526966293).

**Report (PII-redacted):**

Re-installed Breeze on a new VM and using the instructions supplied to route through Nginx Proxy Manager. Still unable to access the customer portal via /portal. It shows a 404 error.
Is there missing config for the NPM deployment for the portal perhaps? Anyone else running NPM in front of Breeze who is able to test the portal?

Additionally, I've set SSO as mandatory after confirming successful connection with Authentik, however I'm never prompted for SSO. When I sign out, I receive the same email address and password login fields.

## Triage analysis

Two distinct, verified defects underlie this report.

1) Portal 404 behind nginx/NPM: the official 'Nginx (Alternative)' reverse-proxy config (apps/docs/src/content/docs/deploy/tls.mdx, lines 60-117) only proxies `/api/` and a catch-all `/` to the web dashboard (port 4321). It never proxies `/portal` to the dedicated portal service. The actual production Caddy config (docker/Caddyfile.prod, lines 282-297) has an explicit `@portal path /portal /portal/*` block routing to `portal:4322` that is deliberately matched BEFORE the web catch-all (comment: 'Must come BEFORE the web catch-all'). Anyone who follows the documented nginx config (or replicates it in Nginx Proxy Manager) has no rule sending `/portal` to the portal container, so it falls through to the web dashboard and 404s. This matches the report exactly.

2) SSO 'mandatory' not reflected on login: enforceSSO IS enforced server-side — apps/api/src/routes/auth/ssoPolicy.ts's assertPasswordAuthAllowedBySso() is called from apps/api/src/routes/auth/login.ts (line 388) and rejects password logins for an org/partner with an active enforceSSO provider. But the dashboard login page (apps/web/src/components/auth/LoginPage.tsx) never calls the documented org-level `GET /api/v1/sso/check/:orgId` endpoint (confirmed via repo-wide grep: no reference to 'sso/check' anywhere under apps/web). It only fetches `/api/v1/auth/login-context` (apps/web/src/lib/loginContext.ts), which covers partner-axis SSO on single-partner self-hosted deployments, not organization-level SSO. So for an org SSO provider with enforceSSO=true, the login page always renders the plain email/password form and never surfaces or redirects to the IdP, exactly as reported — even though the docs (sso.mdx 'SSO Enforcement') describe the frontend hiding the password form based on that check.

## Affected code

- `apps/docs/src/content/docs/deploy/tls.mdx:60-117` — Documented nginx alternative config has no location block for /portal; only /api/ and catch-all / to web:4321
- `docker/Caddyfile.prod:282-297` — Production Caddy has an explicit @portal carve-out to portal:4322 that must precede the web catch-all — the missing piece in the nginx doc
- `apps/web/src/components/auth/LoginPage.tsx:125-151, 329-364` — Only fetches partner-axis login-context; never calls org-level GET /sso/check/:orgId, so password form is always shown regardless of org enforceSSO
- `apps/web/src/lib/loginContext.ts:1-33` — Fetches /api/v1/auth/login-context only — no org SSO check plumbing exists client-side
- `apps/api/src/routes/auth/ssoPolicy.ts:13-49` — Org-level enforceSSO is checked and enforced, but only server-side at password-login time
- `apps/api/src/routes/auth/login.ts:384-410` — Password login calls assertPasswordAuthAllowedBySso and rejects with generic 401 — never signals the client to show an SSO option

## Potential fix

1) Add a /portal (and /portal/*) location block to the documented nginx config, proxying to the portal service's port (4322), placed before the catch-all `location /`, mirroring docker/Caddyfile.prod's @portal ordering. 2) On the dashboard login page, resolve the org for the entered email (or otherwise call GET /api/v1/sso/check/:orgId) and use enforceSSO in the response to hide/collapse the password form and offer the IdP redirect, the same way partnerSso.enforceSSO already gates the UI in LoginPage.tsx.

ToddHebebrand:
Closes #3229 (remaining scope — the nginx `/portal` docs half shipped in #3569).

## The bug

An organization with `enforce_sso = true` was invisible to the dashboard login page. `LoginPage` only ever read `GET /auth/login-context`, which is **partner-axis and single-partner-only**, so those users got a plain email/password form and the server answered every attempt with a generic 401 — no SSO button, no explanation. That is exactly what the reporter hit: *"I've set SSO as mandatory after confirming successful connection with Authentik, however I'm never prompted for SSO."*

`GET /sso/check/:orgId` already existed, but it needs an `orgId` a pre-auth browser does not have.

## The fix

`POST /api/v1/auth/sso-discovery` — public home-realm discovery for the address the user just typed. On a positive answer the login form swaps its password controls for a **"Sign in with {provider}"** button plus a **"Sign in with password instead"** link.

### Discovery is keyed on the email DOMAIN, and never reads `users`

The obvious implementation — look the address up, resolve its tenant, report enforcement — would be a flat **account-existence oracle**. That is precisely the signal SR2-22 (forgot-password) and SR2-23 (the login lockout 429) were rewritten to destroy, and re-introducing it on a new route to save a click is not a trade this codebase makes.

So the answer is a pure function of `(email domain) × (the tenant's own DNS-proven config)`, identical for a real address, a typo, and an address that has never existed. Source is `sso_verified_domains` with `verified_at IS NOT NULL` — **not** the provider's `allowed_domains`, which is unverified free text: on a shared instance a hostile tenant could list someone else's domain and have discovery point their users at an attacker-controlled IdP.

**Residual disclosure, stated plainly:** a positive answer confirms that the domain the caller *already typed* is federated on this instance, and names the provider. Domain-granular, deliberately published by the tenant's own admin, enumerates nothing. Same disclosure as any IdP home-realm-discovery page.

**Coverage, stated plainly:** an org that has not verified a domain gets the negative answer and its users see the password form exactly as they do today. That is pre-existing behaviour, not a regression this route introduces.

### Two details that keep the UI and the server from disagreeing

- **Whether SSO is mandated and which provider gets named are separate queries.** `isPasswordAuthDisabledBySso` blocks on *any* active enforcing provider; `GET /sso/login/:orgId` launches the **oldest active** provider without consulting `enforce_sso` at all. Filtering to `enforce_sso` before ordering would let the button say "Okta" and then start an Authentik flow.
- **A non-OIDC oldest-active provider yields the negative answer**, because the entry route rejects it with a 400. A button guaranteed to fail is worse than no button — and it would make an already-locked-out tenant harder to diagnose.

### Anti-enumeration discipline (mirrors `/auth/forgot-password`)

| Control | Behaviour |
|---|---|
| Timing | `authResponseFloorPromise()` started before any branch, awaited before **every** return — the shared H-4 / SR2-22 equalizer, not a second one |
| Rate limit | Keyed on the **client** (`getClientRateLimitKey`), never the email. 20 / 5 min. Fails **closed** when Redis is unavailable |
| Rate-limit exceeded | A real **429** — the bucket cannot be entered by naming an address, so it discloses nothing, while a `{ sso: null }` 200 would walk a throttled SSO-only user into a form the server always rejects |
| Caching | `Cache-Control: no-store` on every response |
| DB failure | Degrades to `{ sso: null }` + Sentry, never a 500 and never a distinct status |
| RLS | `withSystemDbAccessContext` — pre-auth, so an org-scoped read would silently return zero rows |

### UI

The password path stays reachable behind a reveal link (the #4067 lockout lesson: enforcement is server-side, the UI only de-emphasizes). The **email field stays mounted** while the password controls are collapsed — the partner-axis collapse in `LoginPage` drops the whole form, which is fine there because it fires at page load, but here it would delete the very input discovery keys on. Stale answers are discarded by request sequence, and one request is spent per distinct address, not per blur.

## Docs

`sso.mdx` claimed *"The `GET /api/v1/sso/check/:orgId` endpoint returns `enforceSSO: true` so the frontend can hide the password form"* — never true. Corrected, and the new endpoint documented with its verified-domain prerequisite.

## Design note

The design deviates from the brief on one point, deliberately: an independent review (Codex, `xhigh`) and I converged on domain-keyed rather than address-keyed resolution, because address-keyed lookup would be an account-existence oracle. Codex also caught the enforcement-vs-entry-route provider mismatch above. Trade-off: coverage now depends on the org having verified a domain.

## Verification

- `apps/api` — `ssoDiscovery.test.ts` 16 passed; full `src/routes/auth/` suite 23 files / 480 passed
- `apps/web` — `LoginPage.orgSso.test.tsx` 9 passed, `lib/ssoDiscovery.test.ts` 14 passed; `src/components/auth/` 16 files / 145 passed
- Red-first control verified: reverting `LoginPage.tsx` + `LoginForm.tsx` fails 8 of the 9 new UI tests
- `tsc --noEmit` — API 0 errors, web 0 errors, shared unchanged (3 pre-existing errors in `quotes.test.ts`, untouched)
- `astro check` in `apps/docs` — 0 errors / 0 warnings
- `eslint` clean on all changed files

## Not merged / not closed

No migration, no new table, no cascade-list impact. Leaving merge and issue closure to you.

### DIFF
--- a/apps/api/src/__tests__/integration/ssoDiscovery.integration.test.ts
+++ b/apps/api/src/__tests__/integration/ssoDiscovery.integration.test.ts
@@ -1,0 +1,247 @@
+/**
+ * Real-DB integration coverage for the public POST /auth/sso-discovery
+ * endpoint (#3229 review follow-up).
+ *
+ * WHY THIS FILE EXISTS. The existing unit suite
+ * (apps/api/src/routes/auth/ssoDiscovery.test.ts) mocks Drizzle with a chain
+ * that never inspects the argument passed to `.where()`. That means the
+ * security-critical SQL predicates the handler builds — `isNotNull(verifiedAt)`,
+ * `eq(status, 'active')`, `eq(enforceSSO, true)`, and the
+ * `orderBy(createdAt, id)` that decides which provider is named — are not
+ * actually proven by that suite: any of them could be deleted from
+ * routes/auth/ssoDiscovery.ts and all of its unit tests would still pass.
+ * Only a real-Postgres run can prove the predicates are wired into the SQL
+ * that actually executes. This file is that proof, structured like the
+ * sibling public pre-auth route's integration test
+ * (loginContext.integration.test.ts): build a bare Hono app around the real
+ * route, seed genuine rows via the superuser test client (getTestDb(),
+ * bypasses RLS — same posture as loginContext's fixtures), and assert on the
+ * real HTTP response.
+ *
+ * Redis: the route calls rateLimiter unconditionally and fails CLOSED without
+ * Redis (same posture as GET /auth/login-context). setup.ts's per-test
+ * beforeEach flushes the shared test Redis DB, so every test here starts with
+ * a clean 20-per-5-minutes bucket for its one request — no limiter mock
+ * needed.
+ *
+ * Wire contract under test: packages/shared/src/types/ssoDiscovery.ts.
+ * `sso` is `{ providerName, loginUrl, enforceSSO: true } | null` — presence of
+ * `sso` IS the availability signal.
+ *
+ * Run:
+ *   pnpm --filter @breeze/api exec vitest run --config vitest.integration.config.ts \
+ *     src/__tests__/integration/ssoDiscovery.integration.test.ts
+ */
+import './setup';
+import { randomUUID } from 'crypto';
+import { describe, expect, it } from 'vitest';
+import { Hono } from 'hono';
+import { getTestDb } from './setup';
+import { ssoProviders, ssoVerifiedDomains } from '../../db/schema';
+import { createOrganization, createPartner } from './db-utils';
+import { ssoDiscoveryRoutes } from '../../routes/auth/ssoDiscovery';
+
+/** Suffix every seeded domain with a random token so cases within a single
+ * test (e.g. the two-orgs-same-domain collision case) and across tests can
+ * never bleed into each other via a stray unique-index collision. */
+function uniqueDomain(label: string): string {
+  return `${label}-${randomUUID().slice(0, 8)}.sso-discovery.example.test`;
+}
+
+async function createVerifiedDomain(
+  orgId: string,
+  domain: string,
+  opts: { verified?: boolean } = {},
+) {
+  const db = getTestDb();
+  const [row] = await db
+    .insert(ssoVerifiedDomains)
+    .values({
+      orgId,
+      domain,
+      verificationToken: `seed-token-${randomUUID()}`,
+      verifiedAt: opts.verified === false ? null : new Date(),
+    })
+    .returning();
+  if (!row) throw new Error('failed to create sso_verified_domains fixture');
+  return row;
+}
+
+async function createOrgProvider(
+  orgId: string,
+  opts: {
+    status?: 'active' | 'inactive' | 'testing';
+    enforceSSO?: boolean;
+    name?: string;
+    type?: 'oidc' | 'saml';
+    createdAt?: Date;
+  } = {},
+) {
+  const db = getTestDb();
+  const [row] = await db
+    .insert(ssoProviders)
+    .values({
+      orgId,
+      partnerId: null,
+      name: opts.name ?? 'Test SSO Provider',
+      type: opts.type ?? 'oidc',
+      status: opts.status ?? 'active',
+      enforceSSO: opts.enforceSSO ?? false,
+      ...(opts.createdAt ? { createdAt: opts.createdAt } : {}),
+    })
+    .returning();
+  if (!row) throw new Error('failed to create sso_providers fixture');
+  return row;
+}
+
+function buildApp(): Hono {
+  const app = new Hono();
+  app.route('/auth', ssoDiscoveryRoutes);
+  return app;
+}
+
+async function postDiscovery(app: Hono, email: string) {
+  return app.request('/auth/sso-discovery', {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ email }),
+  });
+}
+
+describe('POST /auth/sso-discovery — real-DB e2e (#3229)', () => {
+  it('happy path: verified domain + active enforcing oidc provider names that provider and sets Cache-Control: no-store', async () => {
+    const app = buildApp();
+    const partner = await createPartner();
+    const org = await createOrganization({ partnerId: partner.id });
+    const domain = uniqueDomain('happy');
+    await createVerifiedDomain(org.id, domain);
+    await createOrgProvider(org.id, { status: 'active', enforceSSO: true, name: 'Acme Okta' });
+
+    const res = await postDiscovery(app, `user@${domain}`);
+    expect(res.status).toBe(200);
+    expect(res.headers.get('cache-control')).toBe('no-store');
+    const body = await res.json();
+    expect(body).toEqual({
+      sso: {
+        providerName: 'Acme Okta',
+        loginUrl: `/api/v1/sso/login/${org.id}`,
+        enforceSSO: true,
+      },
+    });
+  });
+
+  it('unverified domain (verified_at IS NULL): sso is null even though the org otherwise fully qualifies — proves the isNotNull filter is real', async () => {
+    const app = buildApp();
+    const partner = await createPartner();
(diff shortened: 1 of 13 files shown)


ToddHebebrand:
**Review run:** two Sonnet reviewers on `931eacb1c` — `pr-review-toolkit:code-reviewer` under a security lens (8 named attack surfaces) and `pr-review-toolkit:pr-test-analyzer` (vacuous-assertion hunt). A design quorum ran *before* implementation (Codex `gpt-5.6-sol`, `xhigh`, read-only) and materially changed the design — see below.

**Security findings:** 0 critical, 0 high. Explicitly checked and cleared, each against the live sibling code rather than this PR's own comments:
- **Account enumeration** — route never imports or queries `users`; every branch keys on domain or org; status/body/`Cache-Control` identical across all five negative causes; rate-limit key is IP/socket-derived with no email fragment.
- **Timing** — floor started before any branch, awaited before all four returns. One noted gap: `zValidator`'s 400 fires before the handler, so it skips the floor — same as `forgot-password`, and it separates "is an email" from "isn't", never one account from another. Accepted, not fixed.
- **Tenant spoofing** — the `(org_id, domain)` unique index really does permit two orgs to verify the same domain; the `.limit(2)` / `length !== 1` guard is correct as written; `allowedDomains` is never referenced.
- **Open redirect** — `isUsableLoginUrl` blocks cross-origin, bare protocol-relative, and protocol-relative smuggled under the valid prefix; `safeNext` is independently sanitized by the pre-existing `getSafeNext`.
- **XSS** — `providerName` reaches only JSX text children; no `dangerouslySetInnerHTML` in the touched files.
- **UI-vs-server parity** — enforcement query diffed line-for-line against `isPasswordAuthDisabledBySso`'s org branch; provider-naming query against `GET /sso/login/:orgId`'s lookup. No mismatch either direction.
- **Rate limiting / non-SSO regression** — fails closed without Redis; `ssoPrompt: null` renders the form byte-identically to before.

**Test findings:** 2 must-fix, 2 should-fix — **all 4 addressed in `eb04ebe13`**; 0 outstanding.
- *Must-fix:* the Drizzle mock never inspects `.where()`, so **every SQL predicate was unproven** — `isNotNull(verified_at)`, `status='active'`, `enforce_sso=true`, and the `ORDER BY` could each be deleted with all 16 tests still green. Added `ssoDiscovery.integration.test.ts` (7 cases, real Postgres). **Verified discriminating**: deleting the `verified_at` and `enforce_sso` predicates turns 2 of the 7 red.
- *Must-fix:* the org SSO button's actual job (bootstrap → navigate) was never exercised, only its rendering. Added navigation, implicit-form-submission, and bootstrap-failure tests.
- *Should-fix:* untested empty-entry-lookup branch; untested partner/org axis interaction. Both added.
- Also fixed inline from a security note: `user@example.com.` now has its FQDN root dot stripped before lookup (was a silent false negative).

**Tests:**
- `apps/api` — `ssoDiscovery.test.ts` 19 passed; `ssoDiscovery.integration.test.ts` 7 passed **against real Postgres**; full `src/routes/auth/` 23 files / 483 passed
- `apps/web` — `LoginPage.orgSso.test.tsx` 13 passed, `lib/ssoDiscovery.test.ts` 14 passed; `src/components/auth/` 149 passed
- Red-first control verified on both suites: reverting `LoginPage.tsx` + `LoginForm.tsx` fails 8 of 9 original UI tests; reversing the `ORDER BY` columns fails the new ordering assertion
- `tsc --noEmit` clean (API 0, web 0); `astro check` in `apps/docs` 0/0; `eslint` clean on all changed files

**Design quorum outcome (pre-implementation):** Codex dissented on two points and I accepted both. (1) It argued an address-keyed lookup is a flat account-existence oracle against this repo's SR2-22/SR2-23 bar — the design moved to **domain-keyed** resolution over DNS-proven `sso_verified_domains`, which is the deviation from the original brief. (2) It caught that "is SSO enforced" and "which provider gets launched" can resolve to **different rows**, so filtering by `enforce_sso` before ordering would have let the button name one provider and start another's flow. I overrode it on one point — it wanted `{ sso: null }` on rate-limit rather than a 429; the bucket is client-keyed and cannot be entered by naming an address, so a 429 leaks nothing while a null would walk a throttled SSO-only user into a form the server always rejects.

**Status:** review-clean, awaiting maintainer merge. CI on the runner queue was still fully pending when `eb04ebe13` landed; that push re-triggers it.

PR #4861, opened 2026-09-03
fix(api): pin org_id on the hideAiNoise verdict correlation (#4446)

### LINKED ISSUE #4446: [AI agents] hideAiNoise subquery relies on RLS only, no composite org FK
The `hideAiNoise` alert-suppression subquery (wave P2-1, alert verdicts) enforces tenant isolation only via RLS policy, with no composite org FK backing it structurally — unlike other tenant-scoped join paths in the codebase that pair RLS with a composite FK for defense in depth. Should be hardened to match the repo's usual dual-layer pattern.

ToddHebebrand:
## Problem

`hideAiNoiseCondition()` (`apps/api/src/routes/alerts/alerts.ts`) builds the `hideAiNoise=true` `NOT EXISTS` for `GET /alerts`. It correlated `ai_alert_verdicts` to the outer `alerts` row on **id alone**, and its group leg correlated `alert_correlation_members` the same way. Both tables carry a `NOT NULL org_id`, but neither leg carried an org predicate — tenant isolation there rested entirely on the RLS policy.

That is out of step with the sibling reader in the **same handler**: `latestVerdictsForAlerts` (`services/aiAgents/alertVerdicts.ts`) already pins `org_id` on both the verdict row and the member row (added in P2-1 task 16e for exactly this reason). `hideAiNoiseCondition` was the one that got missed.

**Why it is reachable, not theoretical:**
- `GET /alerts` pushes **no app-layer org condition at all** for a `system`-scope token with no `orgId` query param (see the scope branch at the top of the handler) — RLS is the only thing scoping the outer query too.
- `persistAlertVerdict` stamps `ai_alert_verdicts.org_id` from the **run's** org, not from the target alert's org, and run admission does not validate the alert/group's org. A mis-orged verdict row is therefore not structurally prevented today.

A mis-orged live verdict classified as noise could suppress **another tenant's** alert from the noise-filtered list.

## Fix

Both correlated legs now pin the **outer alert's own `org_id` column** (`eq(aiAlertVerdicts.orgId, alerts.orgId)`, and `eq(alertCorrelationMembers.orgId, alerts.orgId)` inside the member subquery). Correlating on the column rather than an auth-derived value is what makes it correct for every scope, including system-with-no-orgId.

Two side benefits:
- The predicate **only narrows the inner EXISTS**, so `NOT EXISTS` can only ever *reveal* an alert that a cross-org verdict was wrongly hiding. It cannot hide anything extra — with `org_id NOT NULL` on both tables there is no NULL-semantics edge.
- `org_id` is the **leading column** of `ai_alert_verdicts_org_alert_idx`, `ai_alert_verdicts_org_group_idx` and `alert_correlation_members_org_alert_idx`. The unpinned form could not use any of them.

### Same-shape sibling in the same handler

The correlation-metadata read that decorates each listed alert joined `alert_correlation_members` → `alert_correlation_groups` with no org pin on either row. Extracted as `correlationMetadataCondition(orgIds, alertIds)` and scoped on the page's own org set — the existing `orgIdsForVerdicts` hoist, renamed `orgIdsForPage` and now shared by both correlation reads. Both joined rows are pinned, not just the member row (same "don't trust the join alone" reasoning as task 16e).

## Why not the composite FK the issue title asks for

Considered and deliberately deferred (I got an independent second opinion here, which agreed: ship the query predicate now, stage the FK separately).

- `alerts` has **no `UNIQUE(id, org_id)`** to reference, and neither does `alert_correlation_groups`. The repo's composite-tenancy-FK pattern is well established (10+ tables reference `devices(id, org_id)`, `invoices(id, org_id)`, `ai_agent_runs(id, org_id)`, …) but `alerts` is not among them — this would be new infrastructure, not a drop-in.
- Adding `UNIQUE(id, org_id)` on `alerts` duplicates most of the PK index on one of the hottest tables in the product, with permanent write amplification.
- A second FK edge into `alerts` widens the `FOR KEY SHARE` lock surface that already produced deadlocks here (#3911), and making `org_id` a *referenced* key adds conflicts for org merge / tenant move.

The structural half is worth doing, but it needs a mismatch audit, `CREATE INDEX CONCURRENTLY`, `NOT VALID` FKs validated separately, and org-merge/deadlock testing — its own change, not a rider on this one. **This PR closes the query-layer half the issue describes ("harden to the repo's dual-layer pattern").**

## Sweep result

Every other reader of these tables was checked:

| Site | Org pin? |
|---|---|
| `services/aiAgents/alertVerdicts.ts` `latestVerdictsForAlerts` | already pinned (task 16e — the precedent this PR mirrors) |
| `services/aiAgents/ticketContext.ts`, `narrativeContext.ts`, `impactRollup.ts`, `runLoop.ts` | already pinned |
| `services/alertCorrelationRca.ts`, `jobs/alertVerdictScheduler.ts`, `jobs/alertCorrelation.ts`, `routes/alerts/correlations.ts` `getPersistedGroupAlerts` | already pinned |
| any `aiTools*.ts` | none read these tables at all |
| `routes/alerts/alerts.ts` correlation-metadata join | **fixed here** |

One observation left out of scope, flagged rather than dropped: `routes/alerts/correlations.ts` `buildPersistedCorrelationGroups` (~L344-350) reads members by `groupId` and member alerts by `alertId` with no org pin on either row, transitively scoped by an earlier `groupOrgConditionsForAuth()` fetch — which also returns `[]` (no condition) for system scope. Same defect class, but that route scopes tenancy through its own helpers (`groupOrgConditionsForAuth`, `getAccessibleAlert`, `filterAlertsBySiteScope`), so it deserves a deliberate audit rather than a blind predicate copy.

## Verification

- `alerts.hideAiNoise.sql.test.ts`: exact compiled-SQL assertion updated, **plus** two targeted assertions kept separate from the whole-string `toBe` (which fails on any reformatting and so is useless as a signal that specifically the org pin was dropped). Written **red first** — all three failed against the unmodified implementation for the right reason.
- New `correlationMetadataCondition` suite **mutation-checked**: removing the org pins turns 2 assertions red; restoring them turns them green.
- `apps/api` alerts routes + `alertVerdicts` service: 26 files / 384 tests passed (`vitest run src/routes/alerts src/services/aiAgents/alertVerdicts`).
- `tsc --noEmit` clean; `eslint` clean on both touched files.
- No migration, no schema change, no new table — no RLS/cascade/export-policy registration applies.

Closes #4446

### DIFF
--- a/apps/api/src/routes/alerts/alerts.hideAiNoise.sql.test.ts
+++ b/apps/api/src/routes/alerts/alerts.hideAiNoise.sql.test.ts
@@ -51,7 +51,7 @@ vi.mock('./helpers', () => ({
   getPagination: vi.fn(), ensureOrgAccess: vi.fn(), getAlertWithOrgCheck: vi.fn(),
 }));
 
-import { hideAiNoiseCondition } from './alerts';
+import { correlationMetadataCondition, hideAiNoiseCondition } from './alerts';
 
 describe('hideAiNoiseCondition (compiled SQL)', () => {
   const dialect = new PgDialect();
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** three independent Sonnet reviewers on `86d76533e` — `pr-review-toolkit:code-reviewer` (correctness + repo conventions), `pr-review-toolkit:pr-test-analyzer` (coverage, judged hard against the repo's no-vacuous-Drizzle-assertion rule), and a dedicated multi-tenant-isolation lens (nullability, dual-axis/partner-wide shapes, writer invariants, DB-context, registration lists).

**Findings:** 3 raised → 1 addressed in `dfe57d5d2`, 1 filed as #4867, 1 declined with reasoning. 0 outstanding blockers — both the code reviewer and the tenancy reviewer returned **zero blocking and zero should-fix** on the change itself.

- **Addressed** — *no route-level proof the handler actually wires `correlationMetadataCondition`.* Real gap: reverting the call site to the pre-fix `inArray(alertCorrelationMembers.alertId, alertIds)` left every suite green, because the compiled-SQL test exercises the function in isolation. `dfe57d5d2` captures the handler's real `.where()` argument through the existing `dbState.whereCalls` harness and asserts the page's org set reaches it **twice** (member row and group row). Mutation-checked both ways: reverting the call site turns exactly that test red, restoring it turns it green.
- **Filed as #4867** — *move-org strands `ai_alert_verdicts` / `alert_correlation_*` rows under the old org.* Pre-existing and out of scope here: `alerts` is in `CORE_DEVICE_ORG_DENORMALIZED_TABLES` but its AI/correlation children are not, so a device org-move leaves them on the old `org_id`. Verified this is **already** the status quo for every org- and partner-scoped caller — `alert_correlation_members`' RLS policy has no system bypass at all, and `ai_alert_verdicts`' is `breeze_has_org_access(org_id)` outside system scope, so the stale rows were invisible post-move regardless. This PR extends that same behaviour to system scope, which is the intended alignment, and it fails toward showing *more*, never toward leaking. Fixing the underlying lifecycle gap needs a group-key collision rule decided (`alert_correlation_groups_org_key_uq`), so it is its own change.
- **Declined** — *add a real-Postgres integration test proving a cross-org verdict no longer suppresses an alert.* Fair ask on blast radius, but: there is no `GET /alerts` integration harness to extend (the only integration suite touching `ai_alert_verdicts` covers impact-rollup, not this route), building one is materially larger than the fix, and I have no live DB in this worktree — so I would be shipping an integration test I could not run, which the repo has been bitten by before. The property at stake is a pure predicate addition whose exact compiled SQL is asserted byte-for-byte and mutation-checked. **Flagging it as the user's call** if you want it before merge.

Confirmations from the tenancy lens, all verified against schema and migrations: `alerts.org_id`, `ai_alert_verdicts.org_id`, `alert_correlation_members.org_id` and `alert_correlation_groups.org_id` are **all NOT NULL** (a nullable one would have made the new `eq` never match and silently disabled suppression); none of the three is a partner-wide/dual-axis table, so there is no `org_id IS NULL` row shape to miss; both new conditions are called only from the `GET /alerts` request handler, never a system/background context; and with no new table, column or migration, nothing needs adding to `CORE_ORG_CASCADE_DELETE_ORDER`, `CORE_DEVICE_CASCADE_DELETE_TABLES`, `CORE_TENANT_EXPORT_POLICY`, `DUAL_AXIS_TENANT_TABLES` or the rls-coverage allowlists (`ai_alert_verdicts` and `alert_correlation_members` are already registered).

**Tests:** `apps/api` alerts routes + `alertVerdicts` service green — 26 files / 385 tests (`vitest run src/routes/alerts src/services/aiAgents/alertVerdicts`, single-fork). `tsc --noEmit` clean. `eslint` clean on all three touched files. Every new assertion was written red-first or mutation-checked; none of them passes with its predicate removed.

**Status:** review-clean, awaiting maintainer merge. CI is still running — not gated on here.

PR #4412, opened 2026-09-01
fix(api): webhook_deliveries blocks GDPR org erasure — add ON DELETE CASCADE (#4100)

ToddHebebrand:
Closes #4100

## Problem

`webhook_deliveries.webhook_id -> webhooks.id` had no `ON DELETE` action (defaulting to `NO ACTION`), while `webhooks` IS registered in `CORE_ORG_CASCADE_DELETE_ORDER` (`tenantCascade.ts`). The moment an org's webhook had ever recorded a delivery, GDPR org erasure's `DELETE FROM webhooks` raised a `23503` FK violation and aborted the whole erasure — the same failure shape CLAUDE.md documents as shipped/blocked five times before (#1359, #1351, #1365, #2179, #2514).

`webhook_deliveries` has no `org_id` column, so the org-cascade contract test's `org_id`-based auto-discovery could never require it in a cascade list.

## Fix

New idempotent migration (`2026-09-29-webhook-deliveries-cascade.sql`): `DROP CONSTRAINT IF EXISTS` + re-add `webhook_deliveries_webhook_id_webhooks_id_fk` with `ON DELETE CASCADE`, mirrored in the Drizzle schema (`onDelete: 'cascade'`). Delivery rows are worthless without their webhook, so cascading is correct, not just convenient — the preferred fix option called out in the issue.

## Why no other registration list needs a change

This repo has multiple tenant-table registration lists, so I checked each one against precedent already in the codebase rather than assuming:

- **`tenantCascade.ts` `ASSOCIATED_SYSTEM_SCOPED_TABLES`** (the FK pre-clear list): not needed — a DB-level `ON DELETE CASCADE` makes the pre-clear unnecessary. Matches the existing `software_install_methods.catalog_id` precedent in the same file ("needs no entry: its catalog_id FK is ON DELETE CASCADE").
- **`orgMergeRegistry.ts` / its `EXTRA_REQUIRED` contract list**: not needed — `webhook_deliveries` has no `org_id` column and is not in `getOrgCascadeDeleteOrder()`, and (unlike `report_runs`, which *did* need an entry when it joined `ASSOCIATED_SYSTEM_SCOPED_TABLES`) this fix never adds `webhook_deliveries` to that list, so it stays out of merge's scope — merge only repoints `webhooks.org_id`, and the child rows travel with their parent's unchanged `id` for free.
[...]

Issue #4100, opened 2026-08-26
[API] webhook_deliveries blocks GDPR org erasure: FK to cascade-registered webhooks has no ON DELETE and the table is in no cascade list


ToddHebebrand:
## Problem

`webhook_deliveries.webhook_id → webhooks.id` is declared with no `ON DELETE` action:

- Schema: `apps/api/src/db/schema/integrations.ts:83` — `.references(() => webhooks.id)` with no `onDelete`
- Migration: `apps/api/migrations/0001-baseline.sql:15108` — `FOREIGN KEY (webhook_id) REFERENCES public.webhooks(id)` (defaults to `NO ACTION`)

`webhooks` IS registered in `CORE_ORG_CASCADE_DELETE_ORDER` (`apps/api/src/services/tenantCascade.ts:391`), but `webhook_deliveries` is in **no cascade list** and nothing deletes it before its parent. The moment the table holds rows for an org's webhooks, **org erasure raises an FK violation on the `webhooks` delete and aborts** — the exact failure mode CLAUDE.md documents as shipped/blocked five times (#1359, #1351, #1365, #2179, #2514).

The org-cascade contract test can't catch this: `webhook_deliveries` has no `org_id` column, so auto-discovery never requires it in the list.

## Why it's latent today, armed tomorrow

Verified 2026-08-26: `webhook_deliveries` is EMPTY (0 rows) in both prod regions. Any webhook activity populates it, and PR #4098 (#4095 recovery path) + wave 3.5c (#4085, at-least-once delivery) make sustained rows the normal state.

## Fix options

1. **Preferred:** new idempotent migration — `DROP CONSTRAINT IF EXISTS webhook_deliveries_webhook_id_webhooks_id_fk` then re-add with `ON DELETE CASCADE`; mirror `onDelete: 'cascade'` in the Drizzle schema. Delivery rows are worthless without their webhook.
2. Alternative: delete deliveries explicitly in the erasure path before `webhooks` — worse, it's another hand-maintained special case outside the contract-tested lists.

Either way, consider a regression guard: an integration assertion that every FK **into** a `CORE_ORG_CASCADE_DELETE_ORDER` table from a non-cascade table carries `ON DELETE CASCADE`/`SET NULL` — this class of gap is invisible to the existing five cascade-list properties.

## Related (same lifecycle gap, lower severity)

[...]

PR #4863, opened 2026-09-03
test(api): contract-test FK ON DELETE actions against org erasure (#4519)

### LINKED ISSUE #4519: [API][Tenancy] ~73 FKs into cascade-list tables have no ON DELETE — add a contract test that guards org erasure
**Problem:** `webhook_deliveries.webhook_id → webhooks.id` had no `ON DELETE` action while `webhooks` sits in `CORE_ORG_CASCADE_DELETE_ORDER`, so GDPR org erasure failed with a 23503 (#4100, fixed in PR #4412). The reviewer on that PR swept the schema and found roughly 73 other foreign keys pointing into cascade-list tables with the same NO ACTION shape, some of them `NOT NULL` (for example `device_software → devices`).

**Why it matters:** each one is a latent org-erasure failure that only surfaces once a row exists in the child table. Nothing in `tenantCascade.ts`, the RLS coverage test, or the export-policy roundtrip guards this.

**Proposed fix:** a contract test (`apps/api/src/__tests__/integration/`) that introspects `pg_constraint` for every FK whose referenced table is in a cascade list and asserts the FK either has `ON DELETE CASCADE`/`SET NULL` or the child table is itself registered in a cascade list ahead of the parent. Seed it with an allowlist of the current offenders so it goes green, then burn the allowlist down with fix-forward migrations.

Related: #4100, PR #4412, #3880.

ToddHebebrand:
## Problem

`cascadeDeleteOrg()` erases a tenant with a sequence of ordinary DELETE statements, each in its own system context — **not** one big deferred transaction, so Postgres checks every referencing FK as each statement runs:

- **step 1b** — each `ASSOCIATED_SYSTEM_SCOPED_TABLES` entry, in array order, with its own hand-written WHERE join;
- **step 2** — every table in `getOrgCascadeDeleteOrder()`, ordered children-before-parents by `topologicalCascadeOrder()`'s live `pg_constraint` read (`organizations` itself keyed on `id`).

A foreign key pointing at any table those statements touch is a latent erasure failure unless something clears the referencing rows first. Erasure then works right up until a customer creates the first row in the referencing table, and after that aborts mid-way leaving the tenant half-deleted. That is exactly how #4100 shipped (`webhook_deliveries.webhook_id → webhooks.id`, fixed in #4412), and the sweep on that PR found ~73 more edges with the same shape.

Nothing guarded it. `tenantCascade.integration.test.ts` checks list membership and topological ordering *within* the cascade set, but never looks at inbound edges from outside it, or at `confdeltype` at all. Neither does the RLS coverage contract or the export-policy roundtrip.

## What this adds

A read-only `pg_constraint` contract test at `apps/api/src/__tests__/integration/orgCascadeFkOnDelete.integration.test.ts`, plus a seeded ledger (`orgCascadeFkOnDeleteAllowlist.ts`) generated from a live migrated database so it lands green on day one.

`protectedTables` = the cascade set + the step-1b pre-clear targets + everything transitively reachable from those by an `ON DELETE CASCADE` edge (301 + 9 + 34 = 343 tables). For an edge `child → parent` with `parent` in that set, the edge is safe when:

| | Condition |
|---|---|
| a | `ON DELETE CASCADE` — and the child is itself protected by the closure, so its own inbound edges get audited |
| b | `ON DELETE SET NULL` **and** every column Postgres would actually null (`confdelsetcols`, else `conkey`) is nullable |
| c | the child is a pre-clear target whose pre-clear runs before the parent's DELETE |
| d | both ends are in the cascade set, so the topo sort puts the child first |
| e | self-referential **and** the table's `org_id` is NOT NULL |

Everything else must be pinned:

| List | Count | Meaning |
|---|---|---|
| `ORG_CASCADE_FK_UNSAFE` | 74 | Debt. Expected to **shrink**. |
| `ORG_CASCADE_FK_PRE_CLEARED` | 17 | Handled by a step-1b pre-clear. Not debt. |

## Three live erasure bugs this surfaced

Not latent shapes — these fire today for any tenant with the relevant rows. All three are pinned with their SQLSTATE and fix rather than repaired here, since #4519 is explicitly scoped to making the debt visible.

1. **`restore_jobs.command_id → device_commands`** (NO ACTION). `device_commands` is emptied in **step 1b**, before the cascade walk that empties `restore_jobs` — so any org that has run a restore driven by a device command aborts erasure at step 1b with **23503**.
2. **`action_intents_scope_ticket_org_fk → tickets`** (SET NULL). Nulls `(scope_ticket_id, org_id)` with no `confdelsetcols`, and `action_intents.org_id` is **NOT NULL**. Postgres accepts that at DDL time and fails at delete time with **23502** — a different SQLSTATE from everything else in the ledger.
3. **`script_categories.parent_id`** (self-referential, NO ACTION). `script_categories.org_id` is **nullable** (partner-wide categories, epic #2135), so `DELETE ... WHERE org_id = $1` does not remove a row set closed under the self-reference: a surviving partner-wide category pointing at an org-owned one raises **23503**. Reproduced directly against Postgres 16.

## Anti-rot — so the ledger can't become permanent scar tissue

- A **stale entry** (FK dropped, renamed, given an `ON DELETE`, or its child table joined the cascade set) fails the burn-down test, so a migration that fixes an edge forces the matching line out in the same PR.
- Each entry's `parentTable`, `reason` and `allColumnsNullable` are **recomputed and verified** against the catalog, never trusted. `reason` says which failure this is; `allColumnsNullable` answers the question a burn-down actually asks — is a plain `ON DELETE SET NULL` even legal here.
- A **`note` is required** for any reason other than the ordinary `child-not-deleted` shape. That is what makes the three findings above impossible to file silently.
- The two lists are mutually exclusive by contract, in both directions.
- Lists stay sorted and duplicate-free, so adding or burning down an entry is a one-line diff.
- A non-vacuity control asserts the graph is large and that **all nine** inputs the classifier discriminates on are actually present (each delete action, self-reference, child-in-cascade, child-with-pre-clear, composite FK) — a broken or dead branch cannot read green.

## Review round

`codex exec` (gpt-5.6, read-only, `high`) — the subagent pool was saturated, so the independent review went through Codex. It raised 6 blockers, 8 should-fix, 3 nits. Every substantive claim was verified against the live database before acting; four were real and are fixed:

- **only cascade-list tables audited as parents** → pre-clear targets are DELETE targets too (found bug 1);
- **`SET NULL` waved through unconditionally** → now checks the nulled columns (found bug 2);
- **self-references exempted outright** → the argument needs row-set *closure*, which nullable `org_id` breaks (found bug 3);
- **`CASCADE` treated as terminal** → cascading into a table makes that table's inbound FKs get checked, hence the transitive closure (+34 tables, +3 pinned edges).

One review finding corrected me on Postgres semantics: my first cut asserted no FK into a cascade table may use `RESTRICT`, on the theory that RESTRICT is checked per row. That is **wrong** — RESTRICT is an AFTER ROW trigger, and a single-statement delete of a closed self-referential row set succeeds under RESTRICT exactly as under NO ACTION (verified empirically). The assertion is gone; row-set closure, modelled directly by (e), is what actually decides it. Also fixed from review: namespace filters now apply to partition **roots** as well as physical relations; a new test asserts no FK is declared directly on a partition (which would make the `(table, constraint)` ledger key ambiguous after root-folding *and* disagree with `topologicalCascadeOrder()`, which reads raw relnames); two pre-clear entries whose `clearSql` keys off a different column than the FK now say so; and several header claims were corrected.

Declined: moving to the `rls-coverage`-style dedicated runner. Measured setup overhead is well under a second per test against a 25-minute shard budget, and the shared glob means no hand-maintained CI step to fall out of date. The trade-off is now recorded in the test header.

## Scope

The **org** cascade only — `CORE_DEVICE_CASCADE_DELETE_TABLES` and the partner purge are separate contracts. **No migrations, no production code change.**

Two gaps the header states plainly: it cannot see a cross-tenant row (org B's child pointing at org A's parent still breaks erasure under (c)/(d); RLS is what prevents that), and it cannot read a pre-clear's `clearSql` to confirm the join reaches the rows a given FK pins — which is why pre-clear FKs are pinned per constraint, not per table, so a fourth FK on `psa_ticket_mappings` reds the test instead of slipping through.

## Verification (local, real Postgres 16)

- New suite: **7 passed**; with `tenantCascade.integration.test.ts`: **12 passed**.
- `test:integration-suite-coverage`: **5 passed** — confirms the file is matched by an include glob and really runs in the blocking shard.
- `tsc --noEmit` clean; `eslint` clean on both files.
- **Mutation-proved every assertion discriminates.** Each of these reds the expected test and only that test:
  - drop a ledger entry → main contract fails naming it;
  - bogus entry → burn-down fails; corrupt a `reason` → drift fails;
  - unsort / duplicate an entry → hygiene fails; strip a required note → hygiene fails;
  - misfile an entry between the two lists → mutual-exclusion fails;
  - **real DDL:** a new table with a NO ACTION FK into `devices` → fails naming it (`ON DELETE CASCADE` version stays green); a NO ACTION FK into a **CASCADE-closure-only** table (`webhook_deliveries`) → fails, proving the closure branch is live; a `SET NULL` FK onto a NOT NULL column → fails as `set-null-onto-not-null`; an FK declared directly on a partition → partition guard fails;
  - **partition control:** creating `metric_rollups_y2099m01` keeps the suite green, so the monthly partitions cannot rot the ledger.

Closes #4519
Refs #4100, #4412, #3880

### DIFF
--- a/apps/api/src/__tests__/integration/orgCascadeFkOnDelete.integration.test.ts
+++ b/apps/api/src/__tests__/integration/orgCascadeFkOnDelete.integration.test.ts
@@ -293,16 +293,30 @@ describe('org-erasure FK ON DELETE contract', () => {
     ).toEqual([]);
   });
 
-  it('every pre-cleared entry names a table with an ASSOCIATED_SYSTEM_SCOPED_TABLES pre-clear', () => {
-    const misfiled = ORG_CASCADE_FK_PRE_CLEARED.filter(
+  it('the two ledger lists agree with ASSOCIATED_SYSTEM_SCOPED_TABLES on which is which', () => {
+    // Both directions, so the lists stay mutually exclusive and each one keeps
+    // meaning what its name says. Without the second half, adding a pre-clear
+    // for a table would leave its FKs sitting in the DEBT list forever, and the
+    // burn-down test would not notice: they are still NO ACTION edges into the
+    // cascade set, so they still look "unresolved".
+    const misfiledAsPreCleared = ORG_CASCADE_FK_PRE_CLEARED.filter(
       (ref) => !preClearedTables.has(ref.childTable),
     );
-
     expect(
-      misfiled.map(refKey).sort(),
+      misfiledAsPreCleared.map(refKey).sort(),
       'ORG_CASCADE_FK_PRE_CLEARED claims these FKs are handled by a step-1b pre-clear, but '
       + 'their table is not in ASSOCIATED_SYSTEM_SCOPED_TABLES (services/tenantCascade.ts). '
       + 'Either add the pre-clear, or move the entry to ORG_CASCADE_FK_KNOWN_UNSAFE.',
+    ).toEqual([]);
+
+    const misfiledAsDebt = ORG_CASCADE_FK_KNOWN_UNSAFE.filter(
+      (ref) => preClearedTables.has(ref.childTable),
+    );
+    expect(
+      misfiledAsDebt.map(refKey).sort(),
+      'These FKs are logged as debt in ORG_CASCADE_FK_KNOWN_UNSAFE, but their table now has '
+      + 'an ASSOCIATED_SYSTEM_SCOPED_TABLES pre-clear. Check that the pre-clear\'s clearSql '
+      + 'actually reaches these rows, then move the entries to ORG_CASCADE_FK_PRE_CLEARED.',
     ).toEqual([]);
   });
 
--- a/apps/api/src/__tests__/integration/orgCascadeFkOnDelete.integration.test.ts
+++ b/apps/api/src/__tests__/integration/orgCascadeFkOnDelete.integration.test.ts
@@ -3,8 +3,9 @@ import { sql } from 'drizzle-orm';
 import { db } from '../../db';
 import { getOrgCascadeDeleteOrder, __testOnly } from '../../services/tenantCascade';
 import {
-  ORG_CASCADE_FK_KNOWN_UNSAFE,
+  ORG_CASCADE_FK_UNSAFE,
   ORG_CASCADE_FK_PRE_CLEARED,
+  type OrgCascadeFkReason,
   type OrgCascadeFkRef,
 } from './orgCascadeFkOnDeleteAllowlist';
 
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** `codex exec` (gpt-5.6-sol, `-s read-only`, `model_reasoning_effort=high`) — full-file correctness review against `tenantCascade.ts`, the sibling cascade contract, and `vitest.integration.config.ts`. The `pr-review-toolkit` subagents were unavailable (session-wide concurrency cap hit), so the independent lens went through Codex instead.

**Findings:** 17 raised (6 blocker, 8 should-fix, 3 nit). Every substantive claim was verified against the live test database before acting on it — 4 blockers were real and are fixed in `bfae6820a`; **fixing them surfaced 3 erasure failures that fire today**, not latent shapes:

- `restore_jobs.command_id → device_commands` — `device_commands` is emptied in step 1b, *before* the cascade walk empties `restore_jobs` → **23503**.
- `action_intents_scope_ticket_org_fk → tickets` — `ON DELETE SET NULL` over a **NOT NULL** `org_id` → **23502**.
- `script_categories.parent_id` — self-referential with a **nullable** `org_id`, so `WHERE org_id = $1` leaves partner-wide survivors pointing at deleted rows → **23503**. Reproduced directly against Postgres 16.

All three are pinned with SQLSTATE + fix rather than repaired here, per #4519's "no migrations in this PR" scope.

One finding corrected me on Postgres semantics: my first cut asserted no FK into a cascade table may use `RESTRICT`, on the theory it is checked per row. It is an AFTER ROW trigger, and a single-statement delete of a closed self-referential row set succeeds under RESTRICT exactly as under NO ACTION (verified empirically). That assertion is gone; row-set closure is what actually decides it.

1 finding declined with reasons recorded in the test header: moving to the `rls-coverage`-style dedicated runner. Measured overhead is sub-second per test against a 25-minute shard budget, and the shared glob means no hand-maintained CI step to fall out of date. **0 outstanding.**

**Tests:** new suite 7 passed; with `tenantCascade.integration.test.ts` 12 passed; `test:integration-suite-coverage` 5 passed (confirms the file really runs in the blocking Integration Tests shard). `tsc --noEmit` clean, `eslint` clean.

**Mutation proof:** every assertion was shown to discriminate — including four real-DDL controls (a NO ACTION FK into `devices`; one into a CASCADE-closure-only table; a `SET NULL` onto a NOT NULL column; an FK declared directly on a partition), each reddening exactly the expected test, with `ON DELETE CASCADE` and new-partition controls staying green.

**Status:** review-clean, local verification green, awaiting CI + maintainer merge. Not merged and #4519 left open by design.

PR #4866, opened 2026-09-03
fix(api): guard migrations that write rows without system scope (#4518)

### LINKED ISSUE #4518: [DB][Tenancy] Migrations doing DML on FORCE-RLS tables silently no-op on non-superuser managed PG unless breeze.scope is set
**Problem:** Migrations that run DML (`UPDATE`/`DELETE`) against a `FORCE ROW LEVEL SECURITY` table silently match zero rows when the migration role is not a superuser, because `breeze_current_scope()` defaults to `'none'` (`0012-tenant-rls-deny-default.sql`). CI runs migrations as superuser and masks it; managed Postgres (doadmin, non-superuser) does not. Any cleanup-then-constraint migration then aborts at `CREATE UNIQUE INDEX` and refuses API boot, and any pure backfill reports `0 cleaned` and moves on.

**Evidence:** Found during review of PR #4421 (#4122). The dedupe `DELETE` on `automation_policy_compliance` removed 0 rows under a non-superuser owner until wrapped with `SELECT set_config('breeze.scope', 'system', true)`. Reproduced both ways.

**Known instance still shipped:** `apps/api/migrations/2026-09-22-ai-alert-verdicts-live-unique.sql` has the same shape and has not been audited against a non-superuser role.

**Proposed fix:**
1. Audit every shipped migration that performs DML on a FORCE-RLS table; fix forward with a new migration where the original no-op'd on managed PG.
2. Add a lint or `autoMigrate.test.ts` rule: a migration file containing `UPDATE`/`DELETE` against a table in the RLS-forced set must contain `set_config('breeze.scope'` before it.
3. Document the rule in CLAUDE.md's migration section.

Related: #4122, PR #4421. Related memory pattern: backfills without `breeze.scope=system` are silent no-ops on managed PG.

ToddHebebrand:
Closes #4518.

## The class

`public.breeze_current_scope()` defaults to `'none'` (`0012-tenant-rls-deny-default.sql`), and `FORCE ROW LEVEL SECURITY` binds the table **owner** — which is the role migrations run as. So on any connection that does not bypass RLS:

- an `UPDATE`/`DELETE` in a migration matches **zero rows with no error** — the `RAISE WARNING` prints a truthful-looking `0 cleaned` and the migration moves on;
- an `INSERT` aborts the migration with `42501`.

The fix is one line before the first write, already the house pattern (`2026-09-30-100000-rls-scoped-backfill-replay.sql`):

```sql
SELECT set_config('breeze.scope', 'system', true);
```

## What this PR adds

- **`apps/api/src/db/migrationRlsScope.ts`** — the analyzer. Masks comments, string literals and routine (`CREATE FUNCTION`/`TRIGGER`/`POLICY`) bodies offset-for-offset, walks *into* `DO $$ … $$` blocks, and reports every `UPDATE`/`DELETE`/`INSERT`/`MERGE` with whether `breeze.scope` was already `'system'` at that offset.
- **`apps/api/src/db/migrationRlsScope.test.ts`** — the guard (runs in **Test API**, no DB needed), plus 10 fixture tests that prove the analyzer discriminates in both directions.
- Docs in `CLAUDE.md` and `apps/api/migrations/README.md`.
- A correction to `check-migrations-nonsuperuser.ts`: its `BYPASSRLS` grant is **load-bearing**, not cosmetic.

### Why the rule covers every table, not a derived FORCE-RLS set

The issue proposed deriving the forced set by grepping `ALTER TABLE … FORCE ROW LEVEL SECURITY`. **That is unsound.** `0008-tenant-rls.sql` and `2026-05-03-tenant-rls-force-and-invites.sql` force every table carrying an `org_id` via `EXECUTE format(…)` over a catalog query — naming no table at all — and several later migrations force lists of tables the same way. A literal grep under-counts badly.

Measured against a fully migrated database (`pg_class.relforcerowsecurity`): **425 of 442** public tables are forced. The 17 that are not are global catalogs (`agent_versions`, `permissions`, `patches`, `device_commands`, …). Requiring the wrapper for **all** migration DML costs **3 extra baseline files** versus perfect DB ground truth, and needs no table registry, no schema import and no ordering model — so it cannot rot when the next table is forced dynamically. Electing system scope is harmless on a table without RLS.

### The baseline cannot be used as an escape hatch

`UNSCOPED_DML_BASELINE` is asserted **exactly, in both directions** (a stale entry fails too), and every entry must sort at or before `BASELINE_CUTOFF` (`2026-10-08-101000-…`). A new migration necessarily sorts *after* the cutoff, so it cannot be silenced by appending a filename. The failure message names the file, line, verb and table, and gives the one-line fix.

## Audit finding: no fix-forward migration is warranted

The issue asks to fix forward where a shipped migration no-op'd on managed PG. **Measured A/B on PG16**, running `check-migrations-nonsuperuser.ts` against a fresh database with the RLS attribute as the *sole* variable:

| Migration role | Result |
|---|---|
| `NOSUPERUSER CREATEDB CREATEROLE REPLICATION **BYPASSRLS**` | full set applies — `OK` |
| same role, **`NOBYPASSRLS`** | aborts at `2026-05-22-snmp-multi-vendor-templates.sql`, `42501 new row violates row-level security policy for table "snmp_templates"` |

Every deployment alive today has applied that May migration, so **every one of them migrates as superuser or BYPASSRLS** — nothing in the field is silently no-op'ing right now. (This also contradicts the premise that DO's `doadmin` is affected; the repo's own `check-migrations-nonsuperuser.ts` comment already asserted `doadmin` has `BYPASSRLS`.)

The named instance, `2026-09-22-ai-alert-verdicts-live-unique.sql`, is additionally safe on its own merits: its dedup `UPDATE`s and the two partial `CREATE UNIQUE INDEX` statements run in **one transaction**, and an index build does not go through RLS. So if the `UPDATE` had no-op'd while live duplicates existed, the index build would have **aborted the migration** rather than recording it. A later fix-forward replay could only ever report `0` — on a database where the file is recorded, the indexes themselves prove there are no duplicates; on one where it failed, no later migration ever runs. Shipping a knowingly-vacuous immutable migration would be noise, so this PR does not add one.

<details>
<summary><b>Proof of the whole chain, run as a NOSUPERUSER/NOBYPASSRLS owner (PG16)</b></summary>

```sql
CREATE TABLE verdicts (id int primary key, alert_id int, superseded_by int);
ALTER TABLE verdicts ENABLE ROW LEVEL SECURITY;
INSERT INTO verdicts VALUES (1, 100, NULL), (2, 100, NULL);   -- duplicates, seeded before FORCE
ALTER TABLE verdicts FORCE ROW LEVEL SECURITY;
CREATE POLICY deny_unless_system ON verdicts
  USING (current_setting('breeze.scope', true) = 'system')
  WITH CHECK (current_setting('breeze.scope', true) = 'system');

SELECT count(*) FROM verdicts;                                --> 0      (owner sees nothing)
UPDATE verdicts SET superseded_by = 1 WHERE id = 2;           --> UPDATE 0   <-- SILENT no-op
CREATE UNIQUE INDEX ... ON verdicts (alert_id) WHERE superseded_by IS NULL;
                       --> ERROR: could not create unique index / Duplicate keys exist.
```

The index build **still sees the duplicates** the owner's own `UPDATE` could not touch — index builds read the heap, not the policy-filtered view. Re-running the same three statements inside one transaction after `SELECT set_config('breeze.scope','system',true)` gives `UPDATE 1` and `CREATE INDEX`.

That is the whole audit argument in five lines: the silent no-op is real, **and** in `2026-09-22-ai-alert-verdicts-live-unique.sql` it is immediately caught by a non-RLS gate in the same transaction, so it cannot have recorded a migration that left duplicate live verdicts behind.

(The same snippet run outside an explicit transaction shows `UPDATE 0` even *with* the `set_config`, because `is_local = true` expires at statement end under autocommit — which is exactly why the guard refuses to credit a top-level elevation in a `-- @no-transaction` file.)

</details>

**What the 122 baseline files really are:** the debt that must be paid before a plain (non-BYPASSRLS) owner role can ever run this migration set — i.e. before `check-migrations-nonsuperuser.ts` can drop its `BYPASSRLS` grant and make this guard runtime-enforced. Repairing them is a separate program of work (they are content-hash immutable, so each needs a fix-forward replay); this PR stops the list growing.

## Heads-up for in-flight migration PRs

Once this merges, **any open PR adding a migration that writes rows must add the one-line `set_config` wrapper** or Test API will fail. There is no baseline entry available to them — that is the point.

<details>
<summary><b>Audit list — 122 shipped migrations with unscoped DML (296 statements), file — count</b></summary>

```
0016-be19-device-ip-history.sql — 1
0018-network-alert-templates.sql — 4
0025-device-approval.sql — 3
0031-zz-dns-security-hardening.sql — 1
0058-device-role-classification.sql — 4
0060-alerting-system.sql — 1
0068-fix-script-typo.sql — 1
0083-backup-mode-targets.sql — 1
0085-patch-policy-kind.sql — 1
0092-recovery-signing-and-boot-media.sql — 1
2026-04-01-backup-verification-type-normalization.sql — 1
2026-04-11-bucket-c-phase-1-inventory-rls.sql — 5
2026-04-11-bucket-c-phase-2-security-patch-rls.sql — 4
2026-04-11-bucket-c-phase-3-device-state-rls.sql — 5
2026-04-11-bucket-c-phase-4-session-execution-rls.sql — 5
2026-04-11-device-metrics-org-id.sql — 1
2026-04-11-device-metrics-rls-policies.sql — 1
2026-04-11-users-rls.sql — 1
2026-04-24-oauth-client-partner-grants.sql — 1
2026-05-02-report-permissions.sql — 5
2026-05-03-billing-manage-permission.sql — 1
2026-05-03-permission-scope-hardening.sql — 3
2026-05-04-reports-export-grandfather.sql — 1
2026-05-13-c-third-party-package-catalog-seed.sql — 1
2026-05-18-device-child-orgid-cascade.sql — 1
2026-05-22-snmp-multi-vendor-templates.sql — 28
2026-05-22-unifi-snmp-templates.sql — 3
2026-05-25-b-audit-log-checksum-chain.sql — 1
2026-05-25-c-audit-log-checksum-canonical-fix.sql — 1
2026-05-25-f-role-force-mfa.sql — 1
2026-05-31-script-execution-batches-org-id.sql — 1
2026-05-31-user-notifications-link-relative-check.sql — 1
2026-06-09-a-native-ticketing-core.sql — 5
2026-06-09-users-disabled-reason.sql — 1
2026-06-10-c-ticket-category-tenant-fks.sql — 2
2026-06-11-h-audit-chain-seal-and-verify.sql — 1
2026-06-11-j-avatar-bytea-columns.sql — 1
2026-06-12-a-huntress-partner-mapping.sql — 3
2026-06-12-a-ticketing-time-parts.sql — 4
2026-06-13-a-ticketing-configuration.sql — 2
2026-06-13-c-partner-timezone-column.sql — 1
2026-06-13-catalog-partner-axis-rls.sql — 4
2026-06-14-product-catalog.sql — 6
2026-06-15-a-invoice-engine.sql — 5
2026-06-15-d-recurring-contracts.sql — 4
2026-06-16-quotes.sql — 5
2026-06-18-neutralize-orphaned-users.sql — 1
2026-06-19-billing-roles.sql — 6
2026-06-20-role-permissions-unique.sql — 1
2026-06-22-stripe-key-status-check.sql — 1
2026-06-24-config-policy-run-tenant-key-backfill.sql — 1
2026-06-25-sso-admin-permission-backfill.sql — 2
2026-06-26-sso-verified-domains.sql — 1
2026-06-27-a-sentinelone-partner-mapping.sql — 3
2026-06-27-a-update-rings-partner-scope.sql — 1
2026-06-27-b-patch-approvals-partner-scope.sql — 2
2026-06-27-c-default-update-ring-dedup.sql — 9
2026-06-29-b-topology-write-permission.sql — 3
2026-06-29-topology-provenance.sql — 3
2026-06-29-vuln-risk-accept-permission.sql — 6
2026-07-01-pam-uac-opt-in-grandfathering.sql — 1
2026-07-02-installer-bootstrap-token-multi-use.sql — 1
2026-07-04-partner-login-branding-accent-check.sql — 1
2026-07-04-user-sso-identities-unique-external.sql — 1
2026-07-11-ai-sessions-read-all-permission.sql — 2
2026-07-11-refresh-token-storage-hardening.sql — 2
2026-07-13-m365-control-plane-foundation.sql — 3
2026-07-14-pax8-direct-draft-uniqueness.sql — 1
2026-07-15-a-ticket-mailbox-verified-ownership.sql — 3
2026-07-15-auth-epochs-and-family-expiry.sql — 1
2026-07-15-b-ticket-mailbox-permissions.sql — 6
2026-07-16-sso-session-binding-and-provider-version.sql — 1
2026-07-18-b-approvals-decide-seed.sql — 2
2026-07-18-c-approvals-decide-seed-cleanup.sql — 1
2026-07-20-partner-export-reconstruction-material-state.sql — 2
2026-07-24-partner-export-configuration-material-state.sql — 1
2026-07-26-scrub-plaintext-temp-passwords.sql — 4
2026-07-28-software-deployments-dispatched-at.sql — 1
2026-07-29-breeze-p-normalize-s3-endpoints.sql — 2
2026-08-03-c-quotes-fulfill-permission.sql — 2
2026-08-06-a-report-site-scope.sql — 2
2026-08-06-d-device-mtls-certificate-history.sql — 1
2026-08-08-organization-external-links.sql — 1
2026-08-08-proxy-session-lifetime.sql — 3
2026-08-11-variables-permissions.sql — 3
2026-08-12-device-identity-collision-alert-template.sql — 1
2026-08-13-ring-third-party-auto-approve-backfill.sql — 3
2026-08-14-a-partner-document-theme.sql — 2
2026-08-14-intent-approval-scope-and-deadlines.sql — 1
2026-08-15-intent-release-lease-backfill.sql — 1
2026-08-15-snmp-poll-attempt-backoff.sql — 1
2026-08-18-drop-organizations-accounting-columns.sql — 1
2026-08-19-contacts.sql — 5
2026-08-20-discovered-asset-detection-source.sql — 1
2026-08-20-installer-bootstrap-token-usage-kind.sql — 1
2026-08-21-park-asset-checkout-default-off.sql — 1
2026-08-23-software-version-url-file-type.sql — 1
2026-08-24-inbound-enabled-backfill.sql — 1
2026-08-25-alert-templates-one-owner.sql — 1
2026-08-27-a-supported-currencies.sql — 1
2026-08-27-b-org-currency-and-fks.sql — 5
2026-08-29-a-catalog-item-prices.sql — 1
2026-08-29-b-catalog-cost-currency-and-org-pricing-currency.sql — 4
2026-08-30-ticketing-currency.sql — 5
2026-09-01-a-bundle-allocation-currency.sql — 1
2026-09-01-b-document-locale-backfill.sql — 3
2026-09-02-a-invoice-line-source-contract-lineage.sql — 1
2026-09-03-ai-agents-permissions.sql — 3
2026-09-11-a-webhook-delivery-event-uniqueness.sql — 1
2026-09-11-b-incident-atomic-winners.sql — 2
2026-09-11-f-alert-notifications-send-identity.sql — 1
2026-09-12-agent-rollback-protocol.sql — 2
2026-09-16-pam-actuation-lifecycle.sql — 1
2026-09-22-ai-alert-verdicts-live-unique.sql — 2
2026-09-25-a-automation-resource-bindings.sql — 5
2026-09-25-b-cross-site-restore-permission.sql — 2
2026-09-25-c-recovery-authorization-subject.sql — 6
2026-09-27-technician-ticket-write-permissions.sql — 1
2026-09-28-quickbooks-entity-mappings.sql — 1
2026-10-06-100101-authenticator-attestation-state.sql — 2
2026-10-08-100600-audit-retention-manage-permission.sql — 2
2026-10-08-100700-audit-retention-policies-org-unique.sql — 1
```

</details>

## Review round

Three Sonnet reviewers ran (code-reviewer, pr-test-analyzer, and an adversarial false-negative hunt). They found **three ways the analyzer was silently wrong**, all now fixed and locked in by fixtures that were verified to FAIL against the previous analyzer:

1. **E-string escapes.** `E'it\'s'` keeps the backslash escape, so the literal does not end at the inner quote. Mis-parsing it desynchronised the scan and every statement after it vanished from the report — an unscoped `DELETE` could hide behind an ordinary org name.
2. **Elevation credited from text that never ran.** A `COMMENT ON … IS $$… set_config('breeze.scope','system',true) …$$` marked every later write in the file as scoped. Root cause: the scanner walked *into* any non-routine dollar-quoted body. Now default-deny — a body is scanned only when it is a `DO` block, a self-invoked routine, or nested under `EXECUTE`. (This one mattered doubly because this PR's own docs make that snippet likely to appear near a migration.)
3. **A function defined AND invoked in the same file** runs its DML at migration time, but the body was blanked as a definition. Now a two-pass scan; `CREATE TRIGGER … EXECUTE FUNCTION f()` correctly does not count as a call site.

Also fixed: `CREATE RULE` bodies (documented as exempt but never actually blanked), and detection added for `CREATE TABLE … AS SELECT` / `COPY … FROM`. `TRUNCATE` and bare `SELECT … INTO var` stay out as a documented decision.

Baseline moved 121 → 122 files / 293 → 296 statements. All three additions are true positives the first analyzer missed.

## Verification

- `vitest run src/db/migrationRlsScope.test.ts` — 23 passed.
- **Control proven, not assumed:** injecting `2026-10-09-999999-guard-probe.sql` with a bare `UPDATE devices SET …` turned the guard red naming `…:2 UPDATE devices`; adding that file to the baseline instead turned the *cutoff* test red. Probe removed, suite green.
- `vitest run src/db` — 59 files, 2042 passed.
- `tsc --noEmit` clean; `eslint` clean on changed files.
- Re-verified after merging `origin/main` (54bfd8dc4).


ToddHebebrand:
**Review run:** 3 Sonnet reviewers in parallel — `pr-review-toolkit:code-reviewer`, `pr-review-toolkit:pr-test-analyzer`, and an adversarial false-negative hunt against the SQL analyzer.

**Findings:** 5 raised → all 5 addressed in 1b7e054db; 0 outstanding. Three were cases where the guard was **silently wrong**, which for a guard is the only severity that matters:

- **E-string escapes** — `E'it\'s'` desynchronised the scanner, making every statement after it disappear from the report. An unscoped `DELETE` could hide behind an ordinary org name.
- **Elevation credited from text that never ran** — a `COMMENT ON … IS $$… set_config('breeze.scope','system',true) …$$` marked every later write in the file as scoped. Fixed at the root: dollar-quoted bodies are now default-deny, scanned only when they are a `DO` block, a self-invoked routine, or nested under `EXECUTE`.
- **Function defined AND invoked in the same file** — its DML runs at migration time but the body was blanked as a definition. Now a two-pass scan; `CREATE TRIGGER … EXECUTE FUNCTION f()` correctly is not a call site.

Plus `CREATE RULE` bodies (documented as exempt, never actually blanked) and detection for `CREATE TABLE … AS SELECT` / `COPY … FROM`. `TRUNCATE` and bare `SELECT … INTO var` stay out as a documented decision.

Each fix is locked in by a fixture **verified to fail against the previous analyzer** (7 of the 9 new fixtures went red when the pre-fix module was swapped back in; the other 2 lock in branches that were already correct but untested). Baseline moved 121 → 122 files / 293 → 296 statements — all three additions are true positives the first analyzer missed.

**Tests:** `vitest run src/db` — 59 files, 2042 passed. `tsc --noEmit` clean, `eslint` clean. Guard control re-proven after the fixes: injecting a migration with a bare `UPDATE devices SET …` turns it red naming the file and line; removing it returns green.

**Note for maintainers, not a blocker:** this PR deliberately ships **no fix-forward migration**, against the original brief. Measured A/B (details in the PR body) shows every live deployment migrates as superuser or BYPASSRLS, so nothing is silently no-op'ing in the field; and for the specific file the issue names, `2026-09-22-ai-alert-verdicts-live-unique.sql`, the dedup UPDATE and the two `CREATE UNIQUE INDEX` statements share one transaction and index builds do not go through RLS — so a no-op'd dedup would have **aborted** the migration, not recorded it. A replay could only ever report 0.

**Also worth knowing before merge:** once this lands, any open PR adding a migration that writes rows must add the one-line `set_config` wrapper or Test API fails. That is intended — the baseline is frozen at a cutoff so new files cannot join it.

**Status:** review-clean, awaiting maintainer merge. Not merged and issue #4518 left open per the runbook.

ToddHebebrand:
Orchestrator note on the "premise" point. The conclusion for the file #4518 names holds: the dedup UPDATE is followed by `CREATE UNIQUE INDEX` in the same transaction, so a silently 0-row dedup would have aborted the migration on any DB that actually had duplicates (this is the self-detecting variant already recorded in the v0.109 audit). No replay migration needed for that file.

The broader claim that every deployment therefore migrates as superuser/BYPASSRLS is not sound for prod. Prod is managed DO Postgres running migrations as `doadmin` (non-superuser, table owner, subject to FORCE RLS); the 0-row backfills on `2026-09-11-b` and `2026-09-25-c` were observed there and fixed forward by #4484. A from-scratch NOBYPASSRLS replay aborting at a May 2026 migration says that file is not replayable that way today, not that prod applied it under BYPASSRLS; prod applied it incrementally, before later policies existed. The guard this PR adds is the durable fix either way, and it is what #4518 asked for beyond the one-file replay. Merging on green.

PR #4869, opened 2026-09-03
fix(api/web): keep an archiving org visible for its whole drain window (#4166)

### LINKED ISSUE #4166: [API] Archiving org (status=offboarding) vanishes from the organizations list for the whole drain window — unlisted, unsearchable, Restore unreachable
## Description

After a partner admin clicks **Archive** on an organization, the org sits in `status = 'offboarding'` for the agent-drain window (measured ~3 min with no agents; can run hours with real agents present). During that window, `GET /api/v1/orgs/organizations` returns the org in **neither** the live nor the archived set for partner-scope callers, so `/settings/organizations` shows nothing for it — not in the active list, not in the archived section, not searchable, and with no confirmation that the archive click actually took effect. Restore (which the API already permits during `offboarding` via `restoreOrgFromArchive` in `apps/api/src/services/orgArchive.ts`) is unreachable from the UI because there's no row to click Restore from.

Verified on `origin/main` `54be3ee4a` (post org-lifecycle PRs #4128/#4132).

## Root Cause

Two independent status allowlists both exclude `offboarding`, and the list route unions only their outputs:

- `computeAccessibleOrgIds` in `apps/api/src/middleware/auth.ts` (lines 378 and 393) hard-filters live org access to `inArray(organizations.status, ['active', 'trial'])`.
- The archived-read carve-out `discoverArchivedOrgIds` in `apps/api/src/services/archivedOrgReads.ts` (line 137) filters `eq(organizations.status, 'archived')`.
- `GET /orgs/organizations` in `apps/api/src/routes/orgs.ts` (~line 1263) unions `accessibleOrgIds` (live) with the archived carve-out (when `includeArchived=true`) — an `offboarding` org matches neither predicate, so it is dropped from both branches of the union.

System-scope callers (platform admins) would still see the row via the live branch's own separate handling, but prod currently has zero platform admins in either region, so every real user hits this gap.

The web UI compounds it independently: `apps/web/src/components/settings/OrganizationsPage.tsx` only ever shows `archived: true` rows (set by the API only for `status === 'archived'`, see `apps/api/src/routes/orgs.ts:1775` and `apps/api/src/services/archivedOrgReads.ts:117`) in the archived section, and the active list is implicitly `active`-only via `accessibleOrgIds`. So a web-only fix would be a no-op today — the API carve-out has to widen first.

## Proposed Fix

Extend the **archived read carve-out** (`discoverArchivedOrgIds`, not `computeAccessibleOrgIds` — that status set gates every tenant-scoped route, not just this list) to `status IN ('archived', 'offboarding')`, so an offboarding org becomes readable through the same read-only system-context path without becoming writable through any mutation path. Then:

- In `apps/web/src/components/settings/OrganizationsPage.tsx`, render offboarding orgs in the archived section with an "Archiving…" badge, a purge/drain countdown, and the existing Restore action.
- Add a unit test on the `GET /orgs/organizations` route asserting an `offboarding` org is returned to a partner-scope caller with `includeArchived=true`.

**Tenancy note:** whichever set gets widened must not leak into any write/mutation path — `offboarding` must stay excluded from `computeAccessibleOrgIds` and from any `LIFECYCLE_FROZEN_ORG_STATUS` write guard. This is a read-visibility fix only.

## Affected Files

- `apps/api/src/services/archivedOrgReads.ts` (primary — carve-out status filter)
- `apps/api/src/routes/orgs.ts` (list route, union logic)
- `apps/web/src/components/settings/OrganizationsPage.tsx` (archived-section rendering, badge/countdown/Restore)
- `apps/api/src/middleware/auth.ts` (reference only — do not widen; noted for contrast)

## Reported By

internal Playwright pre-release sweep 2026-08-28, evidence in `docs/testing/release-sweeps/2026-08-28-v0.108.0-to-main.md` (Group 2, paper cut #8).

ToddHebebrand:
## Problem

Clicking **Archive** on an organization parks it in `status = 'offboarding'` with `offboarding_target = 'archive'` for the agent-drain window — minutes with no agents, up to 72h with a real fleet. During that window it matched neither status allowlist:

| Allowlist | Admits |
|---|---|
| `computeAccessibleOrgIds` (`middleware/auth.ts`) | `active`, `trial` |
| the archived read carve-out (`services/archivedOrgReads.ts`) | `archived` |

`GET /orgs/organizations` unions only those two, so an `offboarding` org fell out of **both** branches: not in the active list, not in the Archived section, not searchable, and with no confirmation the Archive click had taken effect at all. Restore was unreachable from the UI even though `restoreOrgFromArchive` implements exactly that state as its **abort edge** — so a mis-clicked archive was uncancellable while `self_uninstall` rolled through the customer's whole fleet.

## Fix

Widen **only** the read-only carve-out, to `archived OR (offboarding AND offboarding_target = 'archive')`.

`computeAccessibleOrgIds` is untouched, so **visibility does not become writability**: the org stays outside every request's accessible set, every mutation against it still 404s, and the rows are still served inside the Postgres `READ ONLY` archived transaction, under the same partner + `org_access='selected'` pinning as before.

### Why not `status IN ('archived','offboarding')`, as the issue proposed

Only the **archive** path has visible endpoints (`active` → `archived`), so only its middle is a continuity gap. A **churn** drain ends at `churned`, which is intentionally invisible (like `suspended`). Admitting churn drains here would:

- be a *new* visibility policy for a lifecycle whose terminus is deliberately hidden,
- offer a Restore button that `restoreOrgFromArchive` **always** refuses (it requires `offboarding_target = 'archive'`), and
- render a purge countdown off a `purge_at` that is `NULL` for churn.

That gap is real — a churn drain is equally invisible for up to 72h — but it wants its own presentation, not a row flagged `archived: true` under "Archived organizations". **See the open question below.**

### Three supporting changes

- **One definition, not two.** `archiveLifecycleCondition()` is the single source of the admitted states, shared by discovery, both serving reads, and the list route. Two status allowlists drifting apart is precisely what produced this bug; a second copy of the predicate would set up the next one.
- **Serving reads re-assert eligibility** instead of selecting by id alone. Discovery and serving are separate transactions, and the window between them is exactly when these rows flip state (the drain reaper finalizes `offboarding` → `archived`; Restore aborts back to a live status). An id-only serving read could hand back a now-live row flagged `archived: true`.
- **System-scope duplicate.** That branch's live query carries no status filter at all, so with `includeArchived=true` it already returned archived orgs **twice** — once unflagged from the live query, once flagged from the carve-out — and widening would have extended that to draining orgs. Now excluded from the shared `conditions` (count *and* rows, on every page, since the append only happens on the last one). Without the opt-in, behaviour is unchanged.

## Web

`archived: true` now means *"read through the read-only archived door"*, **not** `status === 'archived'`. Both pages that gated read-onlyness on the status string would have rendered a draining org as fully mutable — Edit / Archive / Merge buttons on `OrganizationsPage`, and an editable settings form on `OrgSettingsPage` whose every save could only 404. Both now use the shared `isArchiveLifecycleOrg` (`apps/web/src/lib/archiveLifecycle.ts`).

A draining row reads **"Archiving…"** in the offboarding colour rather than "Archived", with drain-specific copy in the read-only pane and in the settings-page banner — the point of the issue is that the operator could not tell what had happened, so claiming the org *is* archived while its agents are still being uninstalled would be the same lie in a different place. The purge countdown is real here (`purge_at` is stamped at drain entry). Restore is the one available action, and it is the abort edge.

New locale strings added to all eight shipped locales.

## Verification

| Check | Result |
|---|---|
| `apps/api` — `orgs`, `orgs.listQuery`, `orgArchive` (route + service), `archivedOrgReads.scope` | 369 passed |
| `apps/web` — `lib/i18n/*`, `OrgSettingsPage`, `OrganizationsPage.*` | 223 passed across 16 files |
| `tsc --noEmit` — `apps/api` and `apps/web` | clean |

All verified on the branch **after** merging `origin/main` (`450e0bfdc`), so this is the merge commit's result, not a stale base.

Every new test was confirmed **red first** by stashing only the source change and re-running: 10 fail in `archivedOrgReads.scope.test.ts`, 2 in `orgs.test.ts`, 4 in `OrganizationsPage.archived.test.tsx`, 2 in `OrgSettingsPage.test.tsx`.

The discriminating assertion is a compiled-SQL pin: the `offboarding` comparison must be **conjoined** with `offboarding_target = 'archive'`. Widening to a bare `status IN (...)` fails it.

**Not covered locally:** no integration test was added for the new arm — I could not run the integration suite here (it needs a live DB, and the box is hosting several other agents' stacks). The existing `orgArchiveReadContext.integration.test.ts` still covers the `READ ONLY` guarantee and the cross-partner boundary, both unchanged by this PR, and the archived arm it exercises still passes.

**Should a CHURN drain (`offboarding` + `offboarding_target = 'churn'`, reached by `PATCH {status:'offboarding'}`) also be visible during its drain window?** It has the same 72h blackout, but it is not restorable, has no purge date, and ends in a status that is invisible on purpose. Deliberately left out of scope here — it needs its own section and copy, plus a product call on whether `churned`/`suspended` orgs should be listable at all. Happy to file it as a follow-up if you want it.

Closes #4166


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three Sonnet lenses in parallel: `code-reviewer` (tenancy isolation, visibility-vs-writability, drift), `silent-failure-hunter` (zero-row races, NULL-negation, swallowed web errors), `pr-test-analyzer` (discrimination, vacuity, gaps).

**Findings:** 5 raised → 4 addressed in `3783a37a6` + `09a8eee9b`, 1 declined with reason; 0 outstanding.

| # | Lens | Finding | Outcome |
|---|---|---|---|
| 1 | test-analyzer | **No test proved the predicate against real Postgres.** The mocked suites return canned rows whatever the WHERE compiles to, and the existing integration helper only ever produced `status='archived'` — so "a churn drain is excluded" was asserted nowhere that executes SQL. | **Fixed** — added `drainOrg(orgId, target)` plus three real-DB cases (list serves archive-drain / hides churn-drain; detail route the same; cross-partner drain still 404s). Run against an isolated stack; **verified red-first** on the pre-fix predicate (2 fail, 10 pass), green after (12/12). |
| 2 | silent-failure | The serving reads now re-assert eligibility, so a row that stops qualifying between the discovery and serving transactions is dropped — nothing pinned what the caller then sees. | **Fixed** — test pins that the row is *omitted* rather than back-filled with a stale `archived: true`, and that `truncated` stays a fact about discovery. |
| 3 | test-analyzer | The system-scope compiled-SQL route assertion is **effectively vacuous for logical correctness**: `../db/schema` is mocked with sentinels, so drizzle compiles the columns as bound params and the shape collapses to `not ($1 = $2 or ($3 = $4 and $5 = $6))` whatever they stand for. Proven by mutating the predicate to an always-false `and(...)` — 286/286 still passed. | **Fixed by scoping, not by strengthening.** The structure is already pinned against *real* drizzle columns in `archivedOrgReads.scope.test.ts`; duplicating that here would be redundant. Added a comment saying exactly what the route assertion does and does not guard, and which suite to fix instead. |
| 4 | test-analyzer | Two uncovered branches: `conditions` is a ternary on `queryPartnerId` and only the unfiltered arm was tested; the "Archiving…" vs "Archived" badges were only ever asserted one org at a time. | **Fixed** — a case for each. |
| 5 | code-reviewer | System/platform-admin scope can `PATCH` an `offboarding` org's status directly (`offboarding` is absent from `LIFECYCLE_FROZEN_ORG_STATUS_VALUES`). | **Declined — pre-existing and deliberate.** Identical on `origin/main`, and `PATCH {status:'active'}` on a drain intentionally calls `abortOrganizationOffboarding` + `restoreOrganizationTenantAccess` (`routes/orgs.ts:2288-2306`) — that *is* the designed cancel path for a churn drain. Freezing it would need to be conditional on `offboarding_target`, which is a separate lifecycle-policy change. Noted below. |

**Explicitly clean:** tenant isolation across all four read paths (nothing dropped in the `loadArchivedOrg` refactor; wrong-partner / outside-selection / soft-deleted / `quick_support` / ineligible-status all still collapse to the same 404). Visibility→writability (the new negative test holds). NULL-negation on `not(archiveLifecycleCondition())` — independently re-verified against the migrations, not just the comment: `0001-baseline.sql:4236` and `2026-09-12-100001-org-lifecycle-foundations.sql:12`, with no later `DROP NOT NULL`. Web error handling (`runAction` paths untouched; the diff adds no `catch` at all). Repo-wide sweep for a third un-updated consumer of the `archived` flag — none.

**Tests:** `apps/api` unit 371 passed across 5 files (`orgs`, `orgs.listQuery`, `orgArchive` route + service, `archivedOrgReads.scope`); `apps/api` integration `orgArchiveReadContext` 12 passed against real Postgres; `apps/web` 1212 passed across 95 files (`lib/i18n/*` + the whole `components/settings/` tree). `tsc --noEmit` clean on both apps; `eslint` clean on every changed source file. All on the branch **after** merging `origin/main`, so these are the merge commit's results.

Every new assertion was confirmed **red-first** by reverting only the source and re-running — including the three integration cases above. The discriminating pin is that the `offboarding` comparison must be *conjoined* with `offboarding_target = 'archive'`; widening to a bare `status IN (...)` fails it, as does swapping the top-level `or` for `and`.

**Status:** review-clean, awaiting maintainer merge. Two things for you to decide, neither blocking:

1. **Churn drains stay invisible** (the open question in the PR body) — same 72h blackout, but not restorable, no purge date, and terminating in a status that is hidden on purpose. Needs its own section and a product call.
2. **Finding 5** above, if you want `offboarding` + `offboarding_target='archive'` frozen against direct status PATCH the way `archived` is. Bounded today: the purge sweeper gates on `status='archived'` (`services/tenantOffboarding.ts:1340`), so a stale `purge_at` on a reactivated org cannot trigger an erasure.

PR #4870, opened 2026-09-03
fix(api/web): sign extension asset URLs so dynamic import() can load them (#4164)

### LINKED ISSUE #4164: [UI] Extension web assets require a bearer header the dynamic import() cannot send → every enabled extension's UI shows "unavailable" (401)
## Description

Every enabled extension's web UI is dead. `GET /api/v1/extensions/assets/<ext>/<digest>/web/index.js` requires a bearer token, but the client loads that URL via a bare dynamic `import()`, which cannot attach an `Authorization` header.

Verified on current `main` (`54be3ee4a`), local stack with `BREEZE_WORKSPACE_ENABLED=true`:

```
GET /api/v1/extensions/assets/workspace/sha256:f0df14b5…/web/index.js
→ 401 {"error":"Missing or invalid authorization header"}
```

Both the AI Workspace pages and the device-detail Workspace tab render **"This extension is unavailable right now."**

## Root Cause

- `apps/api/src/routes/extensionsWeb.ts`, `createExtensionsWebRoutes`, line 98:
  ```ts
  routes.use('*', authMiddleware);
  ```
  This gates the whole router — including `GET /assets/:name/:digest/:member{.+}` — behind bearer auth. The file's own top-of-file comment (lines ~1-9) describes this as one gate for "the registry projection and digest-addressed serving of extension web/* assets," i.e. asset serving was intentionally lumped in with the authenticated registry endpoint.

- `apps/api/src/middleware/auth.ts`, `authMiddleware`, lines 496-508: reads only the `Authorization` header —
  ```ts
  const authHeader = c.req.header('Authorization');
  if (!authHeader || !authHeader.startsWith('Bearer ')) {
    throw new HTTPException(401, { message: 'Missing or invalid authorization header' });
  }
  ```
  There is no cookie/session fallback anywhere in this function — bearer-only.

- `apps/web/src/lib/extensions/registry.ts`, line 261 (`moduleImporter`, the real loader):
  ```ts
  let moduleImporter: ExtensionModuleImporter = (url: string) => import(/* @vite-ignore */ url);
  ```
  A bare dynamic `import(url)` is a browser-native module fetch — it cannot send custom headers, so the bearer token from `fetchWithAuth` never reaches this request.

- The same file's comment at line 291 calls the resulting failure transient:
  ```ts
  // A rejected import (network hiccup, transient 401 racing a token refresh)
  // shouldn't permanently poison the cache for this URL — allow a later
  // caller to retry.
  ```
  In practice the 401 is deterministic: every dynamic `import()` of an asset URL will 401, on every retry, for every user, as long as `authMiddleware` gates that route — it is not a race with token refresh.

## Effect

- `GET /api/v1/extensions/registry` succeeds (bearer attached via `fetchWithAuth`), so the registry entry and `moduleUrl` resolve correctly.
- The subsequent `import(moduleUrl)` in `loadExtensionModule` (registry.ts) 401s, so `loadExtensionModule` rejects and every extension page/slot/nav item falls back to the "This extension is unavailable right now." error state.
- This is not extension-specific: it affects the whole `ee/workspace` built-in extension's front end (AI Workspace pages + device-detail Workspace tab) and will affect any future extension's web UI, including #4112's BYOK enrichment UI, whenever `BREEZE_WORKSPACE_ENABLED` (or an equivalent extension enable flag) is on.
- Server-side extension routes (registry, and presumably any extension-defined API routes) are unaffected — only the asset-serving path that browser `import()` hits is broken.

## Proposed Fix

Options for the maintainer to choose between — not picking one here:

**(a) Short-lived signed asset URLs.** The authenticated `/registry` call mints a signed, time-limited URL (or token) per asset; `/assets/*` itself becomes unauthenticated but still digest-addressed. Since assets are content-addressed and immutable once digest-pinned, serving them unauthenticated is safe once the digest is unguessable/pinned — this matches the immutability/digest design the loader (registry.ts) already assumes (`ASSET_PATH_PREFIX`, same-origin + path-allowlist checks, digest never re-validated once fetched).

**(b) `fetch()` + blob/data URL.** Fetch the module bytes with the bearer header attached, then `import()` a `blob:` or `data:` URL constructed from the response. Simpler, but breaks relative imports inside the bundle (a chunked/multi-file extension bundle whose entry imports sibling files via relative specifiers would 404, since the blob URL has no meaningful base path) — likely only works for fully self-contained single-file bundles.

**(c) Same-site cookie auth for the assets group only.** Carve `/assets/*` out from `routes.use('*', authMiddleware)` and instead accept a same-site session cookie (if one exists in this app — needs confirming) scoped just to that path. Narrower attack surface than (a) but reintroduces cookie-based auth into an app that is otherwise bearer-only.

## Affected Files

- `apps/api/src/routes/extensionsWeb.ts` (line 98 — `routes.use('*', authMiddleware)`; asset route at `GET /assets/:name/:digest/:member{.+}`)
- `apps/api/src/middleware/auth.ts` (lines 496-508 — `authMiddleware`, bearer-only, no cookie/session fallback)
- `apps/web/src/lib/extensions/registry.ts` (line 261 — `moduleImporter`; line 291 comment mischaracterizing the 401 as transient; `loadExtensionModule`)

## Reported By

Internal Playwright pre-release sweep 2026-08-28, evidence in `docs/testing/release-sweeps/2026-08-28-v0.108.0-to-main.md`.

ToddHebebrand:
## Problem

`createExtensionsWebRoutes` gated its whole router — the registry projection **and** `GET /assets/:name/:digest/:member` — behind the bearer-only `authMiddleware` (`extensionsWeb.ts:98`). The browser loads an extension's web entry with a bare dynamic `import(url)` (`apps/web/src/lib/extensions/registry.ts:261`), a native module fetch that **cannot** attach an `Authorization` header. So every enabled extension's UI 401'd permanently and rendered "This extension is unavailable right now."

That is not extension-specific: it takes down the live `ee/workspace` built-in (AI Workspace pages + the device-detail Workspace tab) and any future extension web UI. The loader's comment calling the 401 "transient" was wrong — it was deterministic, on every retry, for every user.

## Approach

Issue option (a), the pre-decided design: the already-authenticated `GET /extensions/registry` mints a short-lived HMAC-signed capability per extension bundle and embeds it in the `moduleUrl` it advertises; `/assets/*` verifies **that** instead of a bearer header.

**New `apps/api/src/services/extensionAssetToken.ts`** — canonical-JSON claims `{v, aud, name, digest, partnerId, orgId, iat, exp}` signed with `getSecretDerivedKeyMaterials('extension-web-asset-token:v1')` and verified with `timingSafeEqual` across all *retained* keys (rotation-safe). Same shape as the existing `terminalLogoutTicket.ts`. Verification binds the token to the `(name, digest)` the request path asks for, and returns `null` — never throws, never says why — for every failure mode, so the route's bare-404 no-oracle property survives.

### Design notes worth reading

- **Path segment, not query string.** The token rides at `/assets/t/<token>/<name>/<digest>/<member...>`. A relative specifier inside a bundle (`import './chunk.js'`) resolves against the importing module's URL, which drops the query string but keeps parent path segments — so the capability is *inherited* by sibling chunk requests. A `?t=` token is not, and every code-split chunk would 404. `webAssets.ts` retains the whole `web/` subtree, so chunked bundles are in scope by design even though today's `ee/workspace` bundle is single-file (`splitting: false`).
- **Bucketed `iat`.** `iat` snaps to a 15-minute boundary, so every replica mints byte-identical tokens for the same extension/digest/tenant. That preserves the registry projection's cross-replica `revision` determinism and keeps the URL stable within a bucket. It is also why the token is hashed **into** the revision rather than excluded from it: excluding it would let the web client's revision-keyed memo hand back a document whose token had since expired.
- **What the token actually proves.** It is a bearer capability for the whole `(name, digest)` web tree — it must be, because sibling chunks share it. The corollary is a contract, stated in the service header: *everything under `web/` is browser-public to a token holder.* `partnerId`/`orgId` are signed in so a token cannot be re-scoped by tampering, and so a future org-scoped enablement check has the scope in hand — but they are **not** an authorization input today, and the code says so plainly: asset bytes are fleet-global and were already readable by any authenticated user before this change. `userId` is deliberately absent (a decodable per-user id in a URL, with no enforcement value).
- **TTL is 60 minutes.** Long enough for a page session's module graph; a bundle that lazily imports a chunk more than an hour after page load would 404 (the importing module's base URL cannot be re-signed) and recovers on the next navigation.

### Hardening that came out of the design review

- `authMiddleware` is now scoped to `/registry` (+ `/registry/*`) rather than `use('*')`, so the carve-out is a property of that line, not of route-registration order.
- Token verification is **step 0**, ahead of the state-store read, both `realpath` calls, the file read and the re-hash — an unauthenticated flood of junk tokens costs a regex and one HMAC, no DB or disk I/O. There is a test asserting `isEnabled` is never called for a bad token.
- A catch-all bare 404 under `/assets/` keeps a malformed URL away from the app-level `notFound` handler, which echoes `c.req.path` (`index.ts:1090`) and would otherwise reflect a presented token back to the caller.
- Asset `Cache-Control` freshness is capped at the token's *remaining* life instead of a year, so a private cache cannot outlive the live enabled re-check. `immutable` still holds (the URL is digest-addressed). The registry response is `no-store` — it now carries credentials.
- The web loader memoizes imports on the **token-free** asset identity, so a rotated token cannot fragment the cache into duplicate module instances, and wraps a failed import in `ExtensionModuleImportError` whose message has the token redacted (the extension error boundary reports full messages and stacks to Sentry).
- `probeStockHost` harvests the token from the registry response and fetches the asset with **no auth headers**, so the conformance probe fails if the route is ever re-gated on a bearer.

## Verification (local, this branch)

| Suite | Result |
|---|---|
| `apps/api` — `src/extensions`, `routes/extensionsWeb.test.ts`, `services/extensionAssetToken.test.ts` | 18 files, **297 passed** |
| `apps/web` — `src/lib/extensions`, `src/components/extensions` | 5 files, **59 passed** |
| `packages/extension-testkit` | 4 files, **54 passed** |
| `tsc --noEmit` — api, web, extension-testkit | clean |

The four security-critical branches of the verifier (digest binding, expiry, canonical-encoding check, max-TTL check) were each deliberately broken and the suite re-run — every mutation was caught by exactly one failing test, so those assertions are not vacuous.

New coverage: 24 token-service cases (valid, expired, wrong digest, wrong extension, re-scoped, tampered, forged key, non-canonical encoding, over-long TTL, future `iat`, rotated key, bucket determinism); route cases for serving **with no auth context at all**, no token, garbage token, expired token, cross-digest and cross-extension tokens, and the step-0 short-circuit; registry cases for the signed `moduleUrl` and `no-store`; web cases for token-free cache identity, verbatim tokened import, and token redaction in error text; testkit cases proving the asset probe sends no credentials and never leaks the harvested token.

- The asset route is now unauthenticated, so it is covered only by the global 300/min/IP limiter. A separate rate-limit bucket for `/assets/*` would stop chunk fan-out from eating the dashboard's shared budget. Worth filing if fleets grow chunked bundles.
- `ExtensionWebAsset.files[].uncompressedSize` is still unused; a per-member size ceiling before `readFile` would bound the memory a valid token can cost.

Closes #4164


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — three reviewers in parallel: `code-reviewer` (running the `security-review` skill's lens, since this is an auth surface), `silent-failure-hunter`, `pr-test-analyzer`.

**Findings:** 7 raised → 7 addressed in `17c68b0c6`; 0 outstanding.

- *No critical or high security findings.* The reviewer traced the verifier for a forgery bypass and a throw-as-oracle and found neither, and confirmed the gate change left nothing else on that router unauthenticated.
- **Silent-failure, critical — operators got no oracle either.** Verification answers every failure with the same `null` so the route can answer the same bare 404; nothing distinguished "attacker sent junk" from "a signing-key rotation dropped a key while tokens minted under it are still live", which silently 404s every extension UI in the fleet — #4164 all over again, invisibly. Added a throttled internal-only warning that fires only for a token that is well-formed, correctly bound and unexpired yet verifies under no retained key.
- **Silent-failure, high — serving faults buried in anonymous noise.** With the bearer gate gone, the pre-existing silent `realpath`/`readFile` catches are reachable by the whole internet. Anything failing *after* the token verified is by construction a legitimate request hitting an infrastructure problem, so that subset — and only that subset — is now reported.
- **Security, medium — cache-key stripping leaned on another package's invariant.** `assetIdentityKey` stripped `t/<token>` on the strength of a bare `t` first segment. Not currently exploitable (the manifest schema's `isSafeRelativePath` rejects the `..` segments it would take, and the server binds both name and digest into the token), but this module calls itself the loader's trust boundary and should not depend on an invariant in `@breeze/extension-sdk` it never references. Stripping now requires the full tokened shape.
- **Security, low — log injection.** The extension name in the new canary comes from an *unverified* payload. Quoted and truncated.
- **Silent-failure, medium — untelemetered tenant-claim coercion**, plus a credential-hygiene note: the harvested token is now folded into `probeStockHost`'s redaction set, since a `fetchImpl` rejection can carry the request URL in its message.
- **Test-analyzer, 3 gaps** — the mint bucket's effect on the registry `revision` was only tested with each half stubbed against the other; non-GET verbs against a signed asset URL were unpinned; `GET /registry` for a principal with no tenant scope was unpinned. All three now covered.

**Tests:** `apps/api` 310 passed (`src/extensions`, `routes/extensionsWeb.test.ts`, `services/extensionAssetToken.test.ts`); `apps/web` 60 passed (`lib/extensions`, `components/extensions`); `packages/extension-testkit` 55 passed. `tsc --noEmit` clean in all three; `eslint` clean on every changed source file. Verified on a branch with `origin/main` merged in, not a stale base.

**Non-vacuity:** seven deliberate mutations were run against the suite — dropping the digest binding, the expiry check, the canonical-encoding check and the max-TTL check in the verifier; dropping the serve-failure warning; dropping the token from the testkit redaction set; and replacing the bucketed `iat` with a bare `now`. Each broke exactly one test and the suite went green again on restore.

**Not covered locally:** no integration or RLS suite touches this surface (no tenant tables involved), and there is no e2e spec for extension UI, so the end-to-end "browser actually imports the module" path is proven by unit + testkit probe coverage rather than a live browser. Worth a manual check against a local stack with `BREEZE_WORKSPACE_ENABLED=true` before merge if you want belt and braces.

**Status:** review-clean, awaiting maintainer merge. Two follow-ups are noted in the PR description (a separate rate-limit bucket for the now-unauthenticated `/assets/*`, and a per-member size ceiling before `readFile`) — neither blocks this.

PR #4875, opened 2026-09-03
fix(shared): promote SweepProposalReason to @breeze/shared (#4458)

### LINKED ISSUE #4458: AI agents: promote SweepProposalReason to @breeze/shared
`SweepProposalReason` (wave P2-2, sweep finding gate reasons e.g. `device_not_in_evidence`) is currently defined API-side only; promoting it to `@breeze/shared` would let the web sweep-findings UI reuse the same type instead of re-deriving it.

Source: PR #4233 body ("Follow-ups").

ToddHebebrand:
## Summary

- `SweepProposalReason` (AI agents wave P2-2, sweep-finding gate reasons like `device_not_in_evidence`) was defined API-side only in `apps/api/src/services/aiAgents/sweepFindings.ts`, and independently re-derived as an untyped string array on the web sweep-findings UI in `apps/web/src/components/aiAgents/RunDetailPage.tsx`, so the two could silently drift.
- Moved the union to `packages/shared/src/types/aiAgentRuns.ts`, matching the pattern already used for its two siblings in the same file (`AlertVerdictSuggestionReason` from P2-1, `TicketTriageSkipReason` from the #4462 follow-up). The API service and the web component both now import the same type from `@breeze/shared`; the local API-side declaration is removed.
- Tightened `AiAgentRunSweepFindingDto.proposal.reason` from `string | null` to `SweepProposalReason | null` — it was previously untyped.
- The web-side reason-token map (`SWEEP_PROPOSAL_REASON_TOKENS`) is now declared as `satisfies Record<SweepProposalReason, true>` instead of a plain `readonly string[]` literal, so a member added to or removed from the shared union without updating this map is a compile-time error (missing/excess property), not a silent runtime gap. Verified this actually fires by temporarily removing one key and confirming `tsc` failed with the expected "Property is missing" error, then restoring it.
- Checked the OpenAPI spec (`apps/api/src/openapi.ts`) for this enum — the sweep run-detail route (`GET /ai/agents/runs/:runId`) does not register an OpenAPI schema for this DTO, so there is nothing to repoint there.

## Test plan

- [x] `packages/shared`: `vitest run src/types/aiAgentRuns.test.ts` — 10 passed
- [x] `apps/api`: `vitest run src/services/aiAgents/sweepFindings.test.ts` — 22 passed
- [x] `apps/web`: `vitest run src/components/aiAgents/RunDetailPage.test.tsx` — 101 passed
- [x] `tsc --noEmit` clean in `packages/shared`, `apps/api`, `apps/web` (pre-existing, unrelated `quotes.test.ts` errors in `packages/shared` confirmed present on `origin/main` baseline too)
- [x] `eslint` clean on all three touched files
- [x] Manually verified the new `satisfies Record<SweepProposalReason, true>` guard actually fails to compile when a union member is dropped from the map

Closes #4458

### DIFF
--- a/apps/api/src/services/aiAgents/sweepFindings.ts
+++ b/apps/api/src/services/aiAgents/sweepFindings.ts
@@ -91,6 +91,7 @@ import {
   type AiSweepKind,
   type SweepFinding,
   type SweepFindingsOutcome,
+  type SweepProposalReason,
   type SweepProposedAction,
 } from '@breeze/shared';
 import {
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit (code-reviewer, type-design-analyzer) via Sonnet subagents
**Findings:** 1 non-blocking suggestion raised (code-reviewer: `sweepReasonLabel`'s `reason` param was left as bare `string | null` instead of `SweepProposalReason | null`) → fixed inline in c56280f08; 0 outstanding. type-design-analyzer: 0 findings, confirmed the `satisfies Record<SweepProposalReason, true>` guard is a real compile-time exhaustiveness check in both directions (verified by manual test: temporarily dropping a union member from the map produced the expected `tsc` error, then restored).
**Tests:** `packages/shared` `vitest run src/types/aiAgentRuns.test.ts` (10 passed); `apps/api` `vitest run src/services/aiAgents/sweepFindings.test.ts` (22 passed); `apps/web` `vitest run src/components/aiAgents/RunDetailPage.test.tsx` (101 passed); `tsc --noEmit` clean in all three packages (pre-existing unrelated `quotes.test.ts` errors in `packages/shared` confirmed present on `origin/main` baseline, not introduced by this PR); `eslint` clean on all touched files.

Issue #4059, opened 2026-08-26
Extend non-UTC TZ CI coverage beyond auth/SSO: tokenRevocation.ts + scheduler/PAM workers


ToddHebebrand:
### Context

#4046 / PR #4056 added a targeted non-UTC CI pass (`apps/api/vitest.config.tz.ts`, `TZ=America/Denver`) scoped to the auth/SSO/MFA/passkey unit suites, because that's where the confirmed #4018 TZ bug lived. PR review on #4056 (code-reviewer + pr-test-analyzer agents) surfaced two real gaps that were deliberately left out of that PR to keep it a proportionate, config-only fix:

### Gap 1 — `tokenRevocation.ts` has a live, un-simulated version of the same bug shape

`apps/api/src/services/tokenRevocation.ts`'s `isTokenIssuedBeforePasswordChange(iat, passwordChangedAt)` compares a JWT `iat` (UTC epoch) against `users.passwordChangedAt`, which is `timestamp('password_changed_at')` — explicitly commented in `apps/api/src/db/schema/users.ts` as a legacy timestamp-without-timezone column. It's called from `middleware/auth.ts` and `routes/auth/login.ts` on every authenticated request.

Its test file, `apps/api/src/services/tokenRevocation.test.ts`, is not in `vitest.config.tz.ts`'s include list. **Adding it to the list alone is not sufficient** — the current fixtures pass literal `new Date(ms)` instants rather than simulating the pg driver's local-time misparse of an offsetless string (the `offsetlessTimestampFromHostAt`-style simulation PR #4041 introduces for `sso.test.ts`), so as written the test wouldn't fail under the TZ pin either way. This needs both: (a) reshaping the fixtures to simulate a non-UTC host parse, and (b) adding the file to the include list.

### Gap 2 — broader repo-wide TZ-sensitive surface, untouched

`timestamp('col')` (no `withTimezone: true`) appears ~806 times across `apps/api/src/db/schema/*.ts` — this is the repo's default column shape, not an SSO peculiarity. Concrete business logic elsewhere that does timezone/date arithmetic and is NOT covered by the #4046 CI pin:

- `apps/api/src/services/automationRuntime.ts` (`isCronDue` / `getZonedDateParts` — automation/policy schedule firing)
- `apps/api/src/jobs/reportScheduleWorker.ts` (scheduled report timing)
- `apps/api/src/jobs/patchSchedulerWorker.ts` (patch maintenance windows)
[...]

PR #4878, opened 2026-09-03
fix(api): extend non-UTC TZ CI coverage to tokenRevocation.ts (#4059)

ToddHebebrand:
## Summary

Gap 1 from #4059 only (see that issue for Gap 2, the repo-wide timestamp audit — explicitly out of scope here).

`isTokenIssuedBeforePasswordChange` (`apps/api/src/services/tokenRevocation.ts:189`, called from `middleware/auth.ts` and `routes/auth/login.ts` on every authenticated request) compared a JWT `iat` (UTC epoch) against `users.passwordChangedAt` — a `timestamp` column with no timezone — by calling `.getTime()` directly on the Date object. postgres.js parses that offsetless wire value with a bare `new Date(...)`, so the Date handed back carries the UTC wall clock re-read as the host's local time. This is the same bug shape as #4018:

- **East of UTC**: a stale token genuinely issued *before* the password change can read as issued *after* it — an auth-bypass.
- **West of UTC**: a fresh token genuinely issued *after* the password change can read as issued *before* it — a spurious forced logout.

`tokenRevocation.test.ts`'s existing fixtures passed literal `new Date(ms)` instants rather than simulating the driver's local-time misparse, and the file wasn't in `vitest.config.tz.ts`'s include list — so this couldn't have been caught either way.

## Changes

1. **`tokenRevocation.ts`**: added `utcMsFromOffsetlessDbTimestamp` and apply it on the `passwordChangedAt instanceof Date` branch only (a caller-supplied ISO string with an explicit offset is left alone).
2. **`tokenRevocation.test.ts`**: reshaped the `isTokenIssuedBeforePasswordChange` fixtures to use `pgOffsetlessTimestamp` (real-host-TZ simulation, matching `routes/sso.reauth.test.ts`) and added two new cases using a local `offsetlessTimestampFromHostAt`-style helper (arbitrary-offset simulation, matching `services/sso.test.ts`) that directly prove both bug directions independent of the process's actual TZ — both new cases were verified **red** against the pre-fix code under `TZ=America/Denver` before the production fix was applied, and are green now under both the main (UTC) and tz-pinned configs.
3. **`vitest.config.tz.ts`**: registered `tokenRevocation.test.ts`. Also added two files (`sso.reauth.test.ts`, `testUtils/pgOffsetlessTimestamp.test.ts`) that #4041 had already merged but never added despite the file's own MAINTENANCE comment saying to — removed the now-stale comment.

## Verification

```
TZ=America/Denver npx vitest run --config vitest.config.tz.ts src/services/tokenRevocation.test.ts
# before fix: 3 failed, 40 passed (both reshaped existing cases + both new bypass/revocation cases)
# after fix:  43 passed
```

- `npx vitest run --config vitest.config.tz.ts` (full config, 44 files): 1050 passed
- `npx vitest run src/services/tokenRevocation.test.ts` (main/UTC config): 43 passed
- Callers: `middleware/auth.test.ts`, `middleware/auth.epoch.test.ts`, `__tests__/authenticated.example.test.ts`, `__tests__/devices.endpoints.test.ts`, `routes/auth.test.ts`, `routes/auth/login.test.ts` — all green (322 tests)
- `npx tsc --noEmit` — clean

Refs #4059

### DIFF
--- a/apps/api/src/services/tokenRevocation.test.ts
+++ b/apps/api/src/services/tokenRevocation.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type { Redis } from 'ioredis';
+import { pgOffsetlessTimestamp } from '../testUtils/pgOffsetlessTimestamp';
 
 const dbMocks = vi.hoisted(() => ({
   update: vi.fn(),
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit:code-reviewer (Sonnet), independently re-derived the offset-correction formula by hand and empirically re-verified red/green by reverting the production fix and re-running under `TZ=America/Denver`.
**Findings:** 0 blocking, 0 important. 2 nitpicks, neither actioned: (1) `utcMsFromOffsetlessDbTimestamp` duplicates `sso.ts`'s `utcMsFromOffsetlessTimestamp` byte-for-byte — CLAUDE.md permits local duplication over premature cross-module extraction for a two-call-site helper; revisit if a third caller appears. (2) the correction reads `getTimezoneOffset()` off the misparsed Date, so a result can be off by one DST-transition-width right at a spring-forward/fall-back boundary — a pre-existing limitation shared with the already-merged `sso.ts` implementation this PR mirrors, not introduced here.
**Tests:** `TZ=America/Denver npx vitest run --config vitest.config.tz.ts src/services/tokenRevocation.test.ts` — 43/43 passed (reviewer independently confirmed 3 of these fail pre-fix, i.e. non-vacuous). Full tz-pinned config (44 files, incl. the two previously-unregistered `sso.reauth.test.ts` / `pgOffsetlessTimestamp.test.ts`): 1050/1050 passed. Main (UTC) config for the same file: 43/43 passed. Direct callers (`middleware/auth.test.ts`, `middleware/auth.epoch.test.ts`, `__tests__/authenticated.example.test.ts`, `__tests__/devices.endpoints.test.ts`, `routes/auth.test.ts`, `routes/auth/login.test.ts`): 322/322 passed. `npx tsc --noEmit`: clean.
**Status:** review-clean, awaiting maintainer merge. Gap 2 from #4059 (repo-wide timestamp/timezone audit) is intentionally out of scope for this PR.

Issue #4888, opened 2026-09-04
[UI][AI] Run-as (System / Logged-in user) is only selectable from two launch surfaces — Test Run, AI tools, Automations form, and Fix flow silently use the saved default


ToddHebebrand:
## Description
Only the Scripts-library Run modal and the device/bulk `ScriptPickerModal` let the user choose System vs logged-in user at launch. Every other way to run a script uses the script's saved `runAs` default with no way to override, and nothing tells the user which context it will run in. During the OliveTech GCPW debugging (#4882) the same script ran alternately as SYSTEM and as the user with no visible control over which, which is part of why the failures looked random.

Audit (verified):

| Context | Selectable? | Notes |
|---|---|---|
| Scripts library Run (`ScriptExecutionModal.tsx:253-265`) | Yes (system/user) | POSTs `runAs` |
| Device page + Devices bulk (`ScriptPickerModal.tsx:270-276`, `DevicesPage.tsx:658-673`) | Yes (system/user, plus `targetSessionId` for on-demand helper on single device) | POSTs `runAs` |
| Script editor Test Run (`ScriptTestRunner.tsx:328-336`) | **No** | body is `{deviceIds, parameters, triggerType}`; runs whatever the form's advanced-settings default is, which the user may not have noticed |
| Script Builder AI `execute_script_on_device` (`scriptBuilderTools.ts:317-324`) | **No** | schema is `{scriptId, deviceIds, parameters}` |
| Fleet AI `run_script` (`aiToolsScripts.ts:269-279`, `:448-457`) | **No** | dispatch call omits `runAs` |
| Automations `run_script` action form (`AutomationForm.tsx:676-681`) | **No** in UI | backend already honours `action.runAs ?? script.runAs` (`automationRuntime.ts:1043-1047`); the form never sets it |
| Fleet Fix flow (`FixPickerModal.tsx:145-149`) | **No, and misleading** | picker renders the select but the handler discards it (`_runAs`); `fleetFindings/dispatch.ts:599-618` always uses the script default |
| Mobile | N/A | only via AI tool path |

API already supports the override: `executeScriptSchema` (`routes/scripts.ts:273-312`) accepts `runAs: 'system' | 'user'` + `targetSessionId`, and `scriptExecution.ts:210` / `scriptDispatch.ts:228` fall back to the saved default only when it is absent. So this is UI/tool-schema wiring, not a new backend capability. `elevated` is intentionally not a runtime override; keep that.

## Proposed Fix
[...]

Issue #4883, opened 2026-09-04
[AI] Script Builder execute_script_on_device always denied: "Tool 'run_script' is not allowed for this session"


ToddHebebrand:
## Description
The Script Builder chat's only execution tool is dead in production. When the assistant calls `execute_script_on_device`, the user sees the approval prompt and then the tool returns:

```
{"error": "Tool 'run_script' is not allowed for this session"}
```

Observed on the OliveTech tenant (US, API 0.109.0) 2026-09-04, session `45d71fb2…`. The assistant then told the user it "lacks permission" and fell back to asking them to run the script by hand. The system prompt still advertises the tool ("Test-run saved scripts on devices with execute_script_on_device"), so every Script Builder session that tries to iterate on a device hits this.

## Root Cause (verified)
Name mismatch between the session allowlist and the handler name the guard is invoked with:

- The Script Builder session's `allowedTools` is `SCRIPT_BUILDER_MCP_TOOL_NAMES` = `mcp__script_builder__<name>` for each key of `SCRIPT_BUILDER_TOOL_TIERS` (`apps/api/src/services/scriptBuilderTools.ts:52`, wired in `apps/api/src/routes/scriptAi.ts:236`). That list contains `mcp__script_builder__execute_script_on_device`.
- The MCP tool is registered as `execute_script_on_device` but delegates via `makeExistingHandler('run_script', …)` (`scriptBuilderTools.ts:317`), which calls `onPreToolUse('run_script', args)`.
- `createSessionPreToolUse` (`apps/api/src/services/aiAgentSdk.ts:508`) runs `isAllowedForSession('run_script', session.allowedTools)`, which strips the MCP prefix and compares bare names — `run_script` ≠ `execute_script_on_device`, so it is rejected before the tier/approval logic runs.

`scriptBuilderTools.guard.test.ts` knows about the rename (`MCP_NAME_TO_HANDLER`) but only asserts `TOOL_TIERS` / `TOOL_PERMISSIONS` membership; it does not exercise `isAllowedForSession` with `SCRIPT_BUILDER_MCP_TOOL_NAMES`, which is why this shipped green.

## Proposed Fix
Pick one:
1. Have `makeExistingHandler` pass the **MCP tool name** to `onPreToolUse` for the allowlist check and the handler name for dispatch (cleanest — guard sees what the session actually allowed), or
[...]

PR #4890, opened 2026-09-04
feat(api/ai): explain runtime parameter delivery in Script Builder prompt (#4884)

### LINKED ISSUE #4884: [AI] Script Builder prompt never explains how parameters reach the script — assistant guesses $env:Name / Mandatory param() and loops
## Description
The Script Builder system prompt (`apps/api/src/services/scriptBuilderPrompt.ts`) tells the assistant it can set parameters in metadata but never says how Breeze delivers them at runtime. Left to guess, the assistant cycles through wrong mechanisms and the customer pays for each guess with a failed run.

Observed on the OliveTech tenant (US) 2026-09-04: eight Script Builder sessions on one GCPW script with a `GoogleEmail` runtime parameter (plus a sibling `Domain` script). Sequence of assistant guesses, each one shipped to the editor and run:

1. `param([Parameter(Mandatory=$true)][string]$GoogleEmail)` → `Cannot process command because of one or more missing mandatory parameters: GoogleEmail.` (Breeze never passes CLI args; the PowerShell binder fails before the script runs.)
2. Bare `$env:GoogleEmail` → `GoogleEmail parameter is required but was not provided.`
3. Reversed itself: "Breeze passes runtime parameters as PowerShell named arguments", told the user it was "not a bug".
4. Finally wrote an env-dump diagnostic script, discovered `BREEZE_PARAM_GOOGLEEMAIL`, and fixed it. Then re-guessed `$env:Domain` on the next script before self-correcting from the earlier history.
5. Theorised that the editor's Test Run button doesn't inject `BREEZE_PARAM_*` (false — the real cause was #4882, runAs=user) and hard-coded the customer's real email address into the script as a "fallback".

Meanwhile a second real bug (#4882) and a dead execution tool (#4883) made the evidence contradictory, so the assistant had no way to reason to the right answer.

## Proposed Fix
Add a "Runtime parameters" section to the base prompt in `scriptBuilderPrompt.ts`, stating as fact (all verified against `agent/internal/executor/executor.go:335-361` and `shell.go:174-190`):

- Runtime parameters are delivered two ways, simultaneously: (a) as environment variables `BREEZE_PARAM_<NAME>` where NAME is the parameter name upper-cased with `-` folded to `_` (`log-level` → `BREEZE_PARAM_LOG_LEVEL`); (b) by literal text substitution of `{{name}}` / `${{name}}` in the script body before it is written to disk. They are **never** passed as command-line arguments.
- PowerShell: read `$env:BREEZE_PARAM_GOOGLEEMAIL`; Bash: `"$BREEZE_PARAM_GOOGLEEMAIL"`. Never a bare `$env:<Name>`. Never a `param()` block with `Mandatory=$true` for a Breeze parameter — read the env var and validate it manually with a clear `[X] <Name> parameter is required` message and non-zero exit.
- Parameter names declared via `apply_script_metadata` must match the env var the code reads (show the derivation).
- Tenant/secret variables are a different prefix (`BREEZE_VAR_*`) and are only available for SYSTEM/elevated runs — do not conflate.
- On any "parameter missing" report, first read the failing execution (`get_script_execution`) and check `runAs` / which path ran it; do not theorise about Test Run vs Run vs script list without evidence.
- Never embed a real customer value (email, hostname, key) as a fallback/default inside script code.

Also worth adding to the editor-snapshot section: the derived env var name per declared parameter (e.g. `GoogleEmail → $env:BREEZE_PARAM_GOOGLEEMAIL`) so the assistant does not have to compute it.

Consider the same guidance for the general fleet AI's `run_script` tool description in `aiToolsScripts.ts` and for the docs page on script parameters, if it is missing there too.

## Affected Files
- `apps/api/src/services/scriptBuilderPrompt.ts` (primary)
- `apps/api/src/services/scriptBuilderPrompt.test.ts`
- `apps/api/src/services/aiToolsScripts.ts` (tool descriptions, secondary)

## Reported By
Internal — OliveTech tenant (US) Script Builder transcripts, 2026-09-04.

ToddHebebrand:
## Summary

The Script Builder system prompt (`apps/api/src/services/scriptBuilderPrompt.ts`) told the assistant it could declare parameters via `apply_script_metadata` but never explained how Breeze actually delivers them to the running script. Left to guess, the assistant cycled through wrong mechanisms (CLI args, bare `$env:Name`, `Mandatory` `param()`) across multiple customer-visible failed runs on the OliveTech tenant before discovering `BREEZE_PARAM_*` by trial and error (see #4884 for the full transcript).

Adds a "Runtime parameters" section to the base prompt, stating as verified fact:

- Parameters are delivered **two ways, simultaneously**: (a) as `BREEZE_PARAM_<NAME>` env vars, where `<NAME>` is the declared parameter name upper-cased with `-` folded to `_`; (b) by literal `{{name}}` / `${{name}}` substitution in the script body before it's written to disk. They are **never** passed as command-line arguments.
- Never write a `Mandatory` `param()` block for a Breeze parameter (the binder fails before the script runs since there are no CLI args) — read the env var and validate manually.
- Never read a bare `$env:<Name>` — always the `BREEZE_PARAM_` prefixed form.
- `BREEZE_VAR_*` is a separate, unrelated prefix for tenant/secret variables, populated only for SYSTEM/elevated runs.
- On a "parameter missing" report, check the actual execution's `runAs` via `get_script_execution` before theorising.
- Never hard-code a real customer value into script code as a fallback.

Also renders the derived `BREEZE_PARAM_` env var name next to each declared parameter in the editor-snapshot section (e.g. `GoogleEmail → $env:BREEZE_PARAM_GOOGLEEMAIL`) so the assistant doesn't have to compute it.

## Facts verified against the executor before writing the prompt

- `agent/internal/executor/executor.go:335-361` (`buildEnvironment`): `envKey := "BREEZE_PARAM_" + strings.ToUpper(strings.ReplaceAll(key, "-", "_"))`. Confirms upper-case + hyphen-folding. **Discrepancy from the issue text:** the issue implies a general "non-alphanumeric" sanitization; the actual code only folds `-` to `_` — any other character (space, dot, etc.) passes through unchanged after upper-casing. The prompt now states this precisely and advises keeping parameter names alphanumeric-and-hyphen.
- `agent/internal/executor/shell.go:174-190` (`SubstituteParameters`): confirms literal `{{key}}` / `${{key}}` replacement in the script body, done before `WriteScriptFile`.
- `agent/internal/executor/executor.go:160-178`: confirmed `cmd.Args` is built from `shellArgs + scriptPath` only — parameters are never appended as CLI args.
- `apps/api/src/services/scriptSecretDelivery.ts` (`runAsSupportsSecretEnv`): confirms `BREEZE_VAR_*` (tenant secrets) is gated to unset/`system`/`elevated` `runAs` — matches the issue's "SYSTEM/elevated only" claim.

## Scope

Per the fan-out plan, this PR only touches `scriptBuilderPrompt.ts` + its test. The issue also suggested updating the general fleet AI's `run_script` tool description in `apps/api/src/services/aiToolsScripts.ts` — that is deliberately left to #4888, which owns that file, to avoid file-level conflicts with sibling PRs (#4883 also touches adjacent script-execution files).

## Test plan

- [x] Red-first: added 9 new assertions to `scriptBuilderPrompt.test.ts` pinning the BREEZE_PARAM_ derivation, the `{{name}}`/`${{name}}` substitution mention, the "never CLI args" / "never Mandatory param()" / "never bare $env:<Name>" instructions, the BREEZE_VAR_* distinction, the runAs-check-first instruction, the no-hardcoded-values instruction, and the editor-snapshot derived-env-var rendering — watched them fail against the unmodified prompt.
- [x] Implemented the prompt section + `toBreezeParamEnvVar` helper; all 13 tests in the file pass: `cd apps/api && npx vitest run src/services/scriptBuilderPrompt.test.ts`
- [x] `cd apps/api && npx tsc --noEmit` clean

Closes #4884

### DIFF
--- a/apps/api/src/services/scriptBuilderPrompt.test.ts
+++ b/apps/api/src/services/scriptBuilderPrompt.test.ts
@@ -84,8 +84,7 @@ describe('buildScriptBuilderSystemPrompt — runtime parameter delivery', () => 
   });
 
   it('tells the assistant never to use a Mandatory param() block for a Breeze parameter', () => {
-    expect(prompt).toMatch(/Mandatory/);
-    expect(prompt).toContain('param()');
+    expect(prompt).toMatch(/Never write a param\(\) block with a Mandatory/);
   });
 
   it('tells the assistant never to read a bare $env:<Name>', () => {
@@ -103,7 +102,7 @@ describe('buildScriptBuilderSystemPrompt — runtime parameter delivery', () => 
   });
 
   it('forbids embedding a real customer value as a fallback in script code', () => {
-    expect(prompt).toMatch(/never embed/i);
+    expect(prompt).toMatch(/[Nn]ever embed a real customer value.*fallback/);
   });
 
   it('renders the derived BREEZE_PARAM_ env var name next to each declared editor parameter', () => {
@@ -118,5 +117,19 @@ describe('buildScriptBuilderSystemPrompt — runtime parameter delivery', () => 
     });
     expect(withParams).toContain('GoogleEmail → $env:BREEZE_PARAM_GOOGLEEMAIL');
     expect(withParams).toContain('log-level → $env:BREEZE_PARAM_LOG_LEVEL');
+  });
+
+  it('passes non-hyphen characters through unchanged (only "-" is folded to "_"), matching the agent executor exactly', () => {
+    // agent/internal/executor/executor.go buildEnvironment only folds "-" to
+    // "_" after upper-casing; a "." or space is NOT sanitized further. This
+    // pins that documented discrepancy from the issue's assumption of a
+    // general non-alphanumeric sanitizer.
+    const withParams = buildScriptBuilderSystemPrompt({
+      editorSnapshot: {
+        content: 'echo hi',
+        parameters: [{ name: 'api.key', type: 'string' }],
+      },
+    });
+    expect(withParams).toContain('api.key → $env:BREEZE_PARAM_API.KEY');
   });
 });
--- a/apps/api/src/services/scriptBuilderPrompt.test.ts
+++ b/apps/api/src/services/scriptBuilderPrompt.test.ts
@@ -55,3 +55,68 @@ describe('buildScriptBuilderSystemPrompt', () => {
     expect(prompt).toContain('echo hi');
   });
 });
+
+/**
+ * #4884 — the prompt never explained how Breeze actually delivers runtime
+ * parameters to a script, so the assistant guessed CLI args / bare $env:Name /
+ * Mandatory param() across multiple customer-visible failed runs before
+ * eventually discovering BREEZE_PARAM_* by trial and error. These pin the
+ * facts (verified against agent/internal/executor/executor.go buildEnvironment
+ * and shell.go SubstituteParameters) into the base prompt so every session
+ * gets them for free.
+ */
+describe('buildScriptBuilderSystemPrompt — runtime parameter delivery', () => {
+  const prompt = buildScriptBuilderSystemPrompt();
+
+  it('explains the BREEZE_PARAM_<NAME> env var derivation, including the hyphen-folding example', () => {
+    expect(prompt).toContain('BREEZE_PARAM_');
+    expect(prompt).toContain('log-level');
+    expect(prompt).toContain('BREEZE_PARAM_LOG_LEVEL');
+  });
+
+  it('explains the {{name}} / ${{name}} literal substitution path', () => {
+    expect(prompt).toContain('{{name}}');
+    expect(prompt).toContain('${{name}}');
+  });
+
+  it('states parameters are never passed as command-line arguments', () => {
+    expect(prompt).toMatch(/never.*(command-line|CLI) argument/i);
+  });
+
+  it('tells the assistant never to use a Mandatory param() block for a Breeze parameter', () => {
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, pr-test-analyzer) — Sonnet subagents, low blast-radius change (prompt copy + tests only)

**Findings:** 2 raised (both Important, from pr-test-analyzer) → both addressed in f7dbe3026; 0 outstanding.
- Tightened two loosely-independent substring assertions (`Mandatory`/`param()`, `never embed`) into single regexes spanning the actual instruction sentence.
- Added a regression test pinning the documented discrepancy that `toBreezeParamEnvVar` only folds `-` to `_` and passes other non-alphanumeric characters (e.g. `.`) through unchanged, matching `agent/internal/executor/executor.go`'s `buildEnvironment` exactly.

code-reviewer raised one non-blocking suggestion (BREEZE_VAR_* wording could be marginally more precise about the "unset" runAs case) — left as-is per the reviewer's own "no action needed" assessment; it only makes the assistant more cautious, not incorrect.

**Tests:** `cd apps/api && npx vitest run src/services/scriptBuilderPrompt.test.ts` — 14 passed. `npx tsc --noEmit` clean.

Issue #1009, opened 2026-05-30
[security] Helper IPC: assist/user roles not bound to console session — local cross-user token + run_as_user disclosure on multi-user hosts


ToddHebebrand:
**Severity:** Medium (Low on single-user workstations; Medium on multi-user / RDS / terminal-server hosts)
**Component:** Go agent — session broker / Helper IPC (`agent/internal/sessionbroker/`)
**Discovered by:** internal `security-review` skill (two-pass adversarial methodology), independently verified at confidence 8/10.

## Summary

The `assist` and `user` Helper IPC roles are gated only by **(a)** "must not be SYSTEM" and **(b)** the binary path/hash allowlist — with **no binding to the kernel-verified console/active-session SID**. On a host where more than one non-SYSTEM user is logged in interactively, a *second* user who launches the genuine, signed, allowlisted Helper binary can register as `assist`/`user` and obtain capabilities/credentials intended only for the interactive console operator the Helper serves.

The Windows broker pipe is opened to **all Interactive Users**, so concurrent non-SYSTEM connections are expected:

```
agent/internal/sessionbroker/broker_windows.go:15
const pipeSecurity = "D:P(A;;GA;;;SY)(A;;GRGW;;;IU)"   // IU = all interactive users
```

The core SYSTEM-capture gate is **not** affected — `system`/`watchdog`/desktop-capture correctly require `S-1-5-18` (`broker.go:1607,1613`), and the peer SID is kernel-verified and cannot be self-reported. This is strictly a **local cross-user** boundary gap between co-logged-in non-SYSTEM principals.

## Root cause

`roleIdentityRejection` is **negative-only** — it rejects bad pairings but never asserts a *positive* identity for `user`/`assist`:

```
agent/internal/sessionbroker/broker.go:1604-1616
case role == ipc.HelperRoleSystem   && sid != systemSID: // reject
case role == ipc.HelperRoleUser     && sid == systemSID: // reject
case role == ipc.HelperRoleAssist   && sid == systemSID: // reject
case role == ipc.HelperRoleWatchdog && sid != systemSID: // reject
// → any NON-SYSTEM sid passes for user/assist; no check that it is the console user
```

So once the binary path+hash allowlist passes (the binary is world-readable and freely launchable), **any** non-SYSTEM SID is granted the role's scopes (`scopesForRole`, `broker.go:1640/1645-1646`).

### 1. Helper API token disclosure to a non-console user (`assist`)
[...]

PR #4891, opened 2026-09-04
fix(agent): deliver script parameters on the runAs=user helper path (#4882)

### LINKED ISSUE #4882: [Agent] runAs=user scripts never receive parameters — helper executeScript drops payload.parameters (BREEZE_PARAM_* empty)
## Description
Scripts executed in the user's session (`runAs=user`, or any run forwarded to the user helper over IPC) never receive their runtime parameters. The script sees no `BREEZE_PARAM_*` env vars and `{{name}}` template substitution does not happen either, so any parameterised script fails in user context while the identical run in SYSTEM context succeeds.

Observed in production (US, API 0.109.0) on the OliveTech tenant on 2026-09-04. Script `a1ee4017…` ("GCPW - Map Google Email to Windows User") has one required runtime parameter `GoogleEmail`, and every execution row carries `{"GoogleEmail": "…"}` in `script_executions.parameters`. The runs that landed in `C:\WINDOWS\SystemTemp\breeze-scripts\` completed; every run that landed in `C:\Users\<user>\AppData\Local\Temp\breeze-scripts\` (the helper's temp dir) failed with:

```
GoogleEmail parameter is required but was not provided.
```

That split is exactly system-vs-user execution path, not the script content (the same saved content produced both outcomes). The customer burned ~8 Script Builder AI sessions trying to work out why (see #4884).

## Root Cause (verified)
`agent/internal/heartbeat/handlers_script.go:90-98` decodes `parameters` from the command payload into `script.Parameters`, but when the run is forwarded to the helper (`executeViaUserHelper` → `sendCommandToUserHelper`, `handlers_script.go:432`) the raw payload is re-marshalled and the **helper** rebuilds the `ScriptExecution` itself:

```go
// agent/internal/userhelper/client.go:742-747
script := executor.ScriptExecution{
    ID:         cmd.CommandID,
    ScriptType: getStringOrDefault(payload, "language", "bash"),
    Script:     getStringOrDefault(payload, "content", ""),
    Timeout:    getIntOrDefault(payload, "timeoutSeconds", 300),
}
```

No `Parameters` (and no `ScriptID`), so `executor.buildEnvironment` (`executor.go:335-361`) emits no `BREEZE_PARAM_*` and `SubstituteParameters` (`shell.go:174`) has nothing to substitute. The same applies to the `executeScriptInSession` / `targetSessionId` path since it ends in `executeViaUserHelper` too.

## Proposed Fix
- In `userhelper/client.go` `executeScript`, decode `payload["parameters"]` with the same string-only filter the daemon uses (`handlers_script.go:90-98`) and set `Parameters` + `ScriptID` on the `ScriptExecution`. Consider extracting that decode into a shared helper in `executor` so the two sites cannot drift again.
- Add a helper-side test that a forwarded `script` command with `parameters` produces `BREEZE_PARAM_<NAME>` in the spawned process environment, and a daemon-side test that `sendCommandToUserHelper`'s payload still carries `parameters` after re-marshalling.
- Do **not** forward `secretEnv` (`BREEZE_VAR_*`) this way — `runAsSupportsSecrets` already refuses secrets for `runAs=user`; keep that.

## Affected Files
- `agent/internal/userhelper/client.go` (primary, `executeScript`)
- `agent/internal/heartbeat/handlers_script.go` (payload decode to reuse; `sendCommandToUserHelper`)
- `agent/internal/executor/executor.go` (`buildEnvironment`, reference only)

## Reported By
Internal — OliveTech tenant (US) chat logs + `script_executions` rows, 2026-09-04.

ToddHebebrand:
## Problem

Scripts executed in the user's session (`runAs=user`, or any run forwarded to the user helper over IPC) never received their runtime parameters. The script saw no `BREEZE_PARAM_*` environment variables and `{{name}}` template substitution never happened, so any parameterised script failed in user context with its own "parameter is required" error while the byte-identical run in SYSTEM context succeeded.

Observed in production (US, API 0.109.0, OliveTech tenant, 2026-09-04): every run of the `GCPW - Map Google Email to Windows User` script that landed in `C:\WINDOWS\SystemTemp\breeze-scripts\` completed; every run that landed in the helper's temp dir (`C:\Users\<user>\AppData\Local\Temp\breeze-scripts\`) failed with `GoogleEmail parameter is required but was not provided.` That split is exactly system-vs-user execution path, not script content.

## Root cause

Two sites build an `executor.ScriptExecution` from a server payload:

| Site | `parameters` | `scriptId` |
|---|---|---|
| `heartbeat.handleScriptInner` (SYSTEM, local executor) | decoded | decoded |
| `userhelper.Client.executeScript` (the process that actually runs a `runAs=user` script) | **dropped** | **dropped** |

The daemon re-marshals the raw payload over IPC (`sendCommandToUserHelper`) and the helper rebuilds the execution on the other side — so anything the daemon reads off the payload and the helper does not is silently lost. `executor.buildEnvironment` therefore emitted no `BREEZE_PARAM_*` and `SubstituteParameters` had nothing to substitute.

Verified: `apps/api/src/services/scriptDispatch.ts` sends `parameters` (canonicalised to a string map) and `scriptId` on every script command regardless of `runAs`, and only `secretEnv` is enveloped for `script` commands — so `parameters` arrives at the helper intact and plaintext. The gap was entirely the helper's decode.

## Fix

- New `executor.ParametersFromPayload` — **one** string-only decoder, called from **both** sites so they cannot drift again. Non-string values are dropped rather than coerced (matching the prior daemon behaviour and the server's own canonicalisation).
- The helper's `ScriptExecution` now sets `Parameters` and `ScriptID`.

### Deliberately unchanged

- **`SecretEnv` is still never decoded by the helper.** `BREEZE_VAR_*` remains SYSTEM-context-only: `runAsSupportsSecrets` (#3409) refuses a secret-bearing `runAs=user` run on the daemon *before* anything is forwarded, and this change must not become a second delivery route. Asserted from both sides.
- **`RunAs` is not forwarded.** The helper process is already the target user; forwarding `runAs="user"` would make `executor.configureRunAs` reject its own delivery.

### Scope

The explicit `targetSessionId` path (`executeScriptInSession`) benefits with no further change — both of its branches end in `executeViaUserHelper`, which forwards the same raw payload. `ScriptExecution{` has exactly two non-test construction sites repo-wide (`grep`-verified); both now share the decoder.

## Tests (red-first)

Written and watched fail before the implementation:

- **`userhelper/client_parameters_test.go`** — a forwarded `script` command carrying `parameters` now produces `BREEZE_PARAM_GOOGLEEMAIL` in the *spawned process environment*, folds `tenant-domain` to `BREEZE_PARAM_TENANT_DOMAIN`, substitutes `{{GoogleEmail}}` into the content, and sets `BREEZE_SCRIPT_ID`. Confirmed red on all four assertions before the fix (`stdout = "env=\nfolded=\nsub={{GoogleEmail}}\nscriptid=\n"`).
- **`userhelper`** — a payload carrying `secretEnv` still yields zero `BREEZE_VAR_*` entries and never leaks the value; and the no-parameters case leaves placeholders intact with zero `BREEZE_PARAM_*` entries.
- **`heartbeat/handlers_script_parameters_test.go`** — `handleScript`'s re-marshalled payload still carries `parameters` (and leaves `content` unsubstituted) across the IPC hop. **Mutation-probed**: deleting the key in `sendCommandToUserHelper` reddens it, so the control is real rather than vacuous.
- **`heartbeat`** — a secret-bearing `runAs=user` run is refused and nothing reaches the wire.
- **`executor/parameters_test.go`** — table tests for `ParametersFromPayload`, including the post-JSON-round-trip shape (`float64` numbers, no Go type info) the helper actually sees.

## Verification

```
cd agent && go test -race ./internal/userhelper/... ./internal/heartbeat/... ./internal/executor/...
ok  github.com/breeze-rmm/agent/internal/userhelper  2.267s
ok  github.com/breeze-rmm/agent/internal/heartbeat  40.865s
ok  github.com/breeze-rmm/agent/internal/executor   2.558s
```

`go build ./...` and `go vet` on all three packages clean. No TypeScript touched.

## Rollout note

This is an agent-side fix, so it takes effect only once the agent fleet rolls to a build containing it — the API needs no change.

Closes #4882

### DIFF
--- a/agent/internal/executor/parameters.go
+++ b/agent/internal/executor/parameters.go
@@ -1,0 +1,48 @@
+package executor
+
+// ParametersFromPayload decodes a script command payload's "parameters" field
+// into the string-only map ScriptExecution.Parameters expects.
+//
+// ONE decoder, deliberately, for the two places that build a ScriptExecution
+// from a server-supplied payload:
+//
+//   - heartbeat.handleScriptInner — the SYSTEM-context local executor, and
+//   - userhelper.Client.executeScript — the runAs=user hop, where the daemon
+//     re-marshals the raw payload over IPC and the helper process rebuilds the
+//     ScriptExecution on the other side.
+//
+// Those two drifted once already (#4882): only the daemon decoded parameters,
+// so every user-context script reached the shell with no BREEZE_PARAM_* and
+// with its {{name}} placeholders unexpanded, while the byte-identical script
+// in SYSTEM context ran fine. A parameterised script therefore failed with the
+// script's own "parameter is required" error and nothing in the agent logs
+// pointed at the cause. Sharing the decode is what makes that divergence
+// impossible to reintroduce by editing one site.
+//
+// Non-string values are dropped rather than coerced: the server only ever
+// sends strings (script_executions.parameters is a string map), and
+// stringifying a JSON number here would make the helper produce a value the
+// SYSTEM path never would — reintroducing drift in the opposite direction.
+//
+// Returns nil when the payload has no "parameters" object at all. nil and an
+// empty map behave identically downstream (SubstituteParameters returns the
+// content untouched; buildEnvironment appends nothing), so callers may assign
+// the result unconditionally.
+//
+// SecretEnv is deliberately NOT handled here. Secrets are parsed and validated
+// by ParseSecretEnv on the daemon only, and runAsSupportsSecrets refuses a
+// secret-bearing runAs=user run outright (#3409) — the helper must never gain a
+// second delivery route for BREEZE_VAR_*.
+func ParametersFromPayload(raw any) map[string]string {
+	params, ok := raw.(map[string]any)
+	if !ok {
+		return nil
+	}
+	decoded := make(map[string]string, len(params))
+	for key, value := range params {
+		if s, ok := value.(string); ok {
+			decoded[key] = s
+		}
+	}
+	return decoded
+}
--- a/agent/internal/executor/parameters.go
+++ b/agent/internal/executor/parameters.go
@@ -1,5 +1,11 @@
 package executor
 
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
 // ParametersFromPayload decodes a script command payload's "parameters" field
 // into the string-only map ScriptExecution.Parameters expects.
 //
@@ -36,13 +42,35 @@ package executor
 func ParametersFromPayload(raw any) map[string]string {
 	params, ok := raw.(map[string]any)
 	if !ok {
+		// Absent is the ordinary case for an unparameterised script and says
+		// nothing. Present-but-not-an-object means a producer broke the wire
+		// contract, and the script is about to run with its placeholders
+		// intact and no BREEZE_PARAM_* — exactly the invisible failure #4882
+		// was. Log the type (never the content) so it is diagnosable from
+		// agent logs instead of only from the script's own error message.
+		if raw != nil {
+			log.Warn("script command carried a non-object `parameters` field; running with no parameters",
+				"payloadType", fmt.Sprintf("%T", raw))
+		}
 		return nil
 	}
 	decoded := make(map[string]string, len(params))
+	var dropped []string
 	for key, value := range params {
 		if s, ok := value.(string); ok {
 			decoded[key] = s
+			continue
 		}
+		dropped = append(dropped, key)
+	}
+	if len(dropped) > 0 {
+		// The server canonicalises every value to a string before dispatch
+		// (canonicalizeScriptParameters), so reaching here means that
+		// chokepoint was bypassed. Keys only — a value is operator-supplied
+		// content that has no business in the agent log.
+		sort.Strings(dropped)
+		log.Warn("dropping non-string script parameter values",
+			"keys", strings.Join(dropped, ","))
 	}
 	return decoded
 }
--- a/agent/internal/executor/parameters_test.go
+++ b/agent/internal/executor/parameters_test.go
@@ -1,8 +1,11 @@
 package executor
 
 import (
+	"bytes"
 	"encoding/json"
+	"log/slog"
 	"reflect"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, plus a second focused code-reviewer pass over the follow-up commits.

**Findings:** 4 raised across both rounds → all 4 addressed; 0 outstanding.

| # | Source | Finding | Resolution |
|---|---|---|---|
| 1 | silent-failure-hunter | `ParametersFromPayload` drops a non-object `parameters` field and non-string values with no diagnostic — the same invisible-failure shape as the bug being fixed | `fe8cc45c4` — two `log.Warn` calls (offending Go **type** only; dropped **keys** only, never values). Absent `parameters` stays silent. |
| 2 | pr-test-analyzer | Shell-metacharacter parameter values untested on the path this PR reactivates | `d8dd93e02` — byte-exact env delivery, plus post-substitution validator ordering (a value resolving to `Format-Volume` is refused) |
| 3 | pr-test-analyzer | The dropped-values log test asserted keys appear but never that **values** stay out, despite that being the stated guarantee | `8628cd078` — capture JSON and pin the exact log **attribute set**. (Substring checks can't do this: `"3"` matches the slog timestamp.) |
| 4 | code-reviewer (round 2) | The console-session pinning in `fe8cc45c4` was **inert** — `preferredRunAsUserSessionForOS`'s non-Windows branch never reads `WinSessionID`, and `SetGOOSForTest` was never called, so on Linux/macOS CI it read as #1009 coverage while exercising the fallback | `f2e5f35f6` — pin the branch explicitly and run the forwarding test over **both** `goos` values; also hardened the secret-refusal test, which asserted "nothing forwarded" (trivially true with no resolvable session) and now asserts the refusal is the #3409 gate |

Zero findings on the production change itself (`0b67504bf`) from either code-review pass: decode-extraction equivalence, secret isolation, `ScriptID` safety, and the parameters-vs-`secretEnv` separation in `scriptDispatch.ts` all verified.

**Every red was watched, not assumed.** Each assertion below was killed by a deliberate mutation before being trusted:

| Mutation | Kills |
|---|---|
| drop `Parameters` from the helper's `ScriptExecution` | the 4 delivery assertions, the metacharacter test, the validator-ordering test |
| `delete(cmd.Payload, "parameters")` in `sendCommandToUserHelper` | the daemon forwarding test |
| add `"values", fmt.Sprintf("%v", params)` to the warning | the attribute-set pin (finding 3's exact scenario) |
| console session `"1"` → `"9"` | **only** the `windows` subtest — which is the proof finding 4 is actually fixed |
| remove either `log.Warn` | the corresponding tripwire subtest |

**Tests:** `cd agent && go test -race ./internal/executor/... ./internal/userhelper/... ./internal/heartbeat/...` — all three green (2.3s / 2.2s / 43.8s). Full agent suite `CGO_ENABLED=0 go test ./...` green. `go vet` clean; `golangci-lint run --new-from-rev=origin/main ./...` reports **0 issues**. CI on the earlier heads: Test Agent, Test Agent (race), Test Agent (Windows), Lint Agent (Go), all four Build Agent targets, and Type Check all passed — the production code is unchanged since `0b67504bf`; every commit after it is test-only or a log statement.

**Scope note for the reviewer:** the production diff is 3 changed lines plus one new decoder file. `ScriptExecution{` has exactly two non-test construction sites repo-wide and both now share it.

**Status:** review-clean, awaiting maintainer merge. This is agent-side only — it takes effect when the fleet rolls to a build containing it; the API needs no change.

ToddHebebrand:
**CI on the final head (`f2e5f35f6`): fully green** — 65 pass, 0 fail, 1 skipping (CodeQL). Includes Test Agent, Test Agent (race), Test Agent (Windows), Lint Agent (Go), all four Build Agent targets, Type Check, Test API, Test Web, and all four Integration Tests shards.

Supersedes the "CI on the earlier heads" line in the review summary above — that was written while this run was still in flight.

PR #4892, opened 2026-09-04
fix(api/ai): check the session allowlist against the exposed MCP tool name (#4883)

### LINKED ISSUE #4883: [AI] Script Builder execute_script_on_device always denied: "Tool 'run_script' is not allowed for this session"

ToddHebebrand:
## Problem

Every Script Builder `execute_script_on_device` call in production came back

```json
{"error": "Tool 'run_script' is not allowed for this session"}
```

so the chat's only execution tool was dead and the assistant fell back to asking the user to run the script by hand. Observed on the OliveTech tenant (US, API 0.109.0).

Two tool identities were conflated:

| Identity | Value | Keyed on |
|---|---|---|
| **Exposed** (what the model calls, what the session allowlist holds) | `mcp__script_builder__execute_script_on_device` | `SCRIPT_BUILDER_MCP_TOOL_NAMES` → `session.allowedTools` |
| **Handler** (what actually runs) | `run_script` | `TOOL_TIERS`, `TOOL_PERMISSIONS`, `toolInputSchemas`, per-tool rate limits |

`makeExistingHandler` passed the **handler** name to `onPreToolUse`, so `createSessionPreToolUse` ran `isAllowedForSession('run_script', ['mcp__script_builder__execute_script_on_device', …])` — no match, denied before tier, RBAC or approval logic ran at all.

## Approach — option 1 from the issue

The issue offered two fixes. This takes **option 1** (make the guard see what the session actually granted) rather than **option 2** (alias `run_script` into the allowlist), because widening an allowlist is the wrong direction for a security guard: option 2 would have granted the bare `run_script` capability to every Script Builder session, not just the one exposed tool.

- **`PreToolUseCallback` gains an optional third argument** — the `mcp__<server>__<tool>` name the call was *exposed* as. `createSessionPreToolUse` uses it for the session-allowlist check **only**. Tier, RBAC, rate limits, approval and the audit row all stay on `toolName`, because the handler is where the capability lives. Every other caller omits it and behaves exactly as before.
- **`makeExistingHandler` now takes the MCP tool name** and resolves the handler through a single exported `SCRIPT_BUILDER_HANDLER_BY_MCP_TOOL` map, so a `tool()` registration names its tool once. The previous arrangement — passing the handler name at the registration site — is what let the two identities drift apart in the first place.
- **The denial message now names the model-facing tool.** A genuinely-refused assistant says which capability it lacks instead of naming an internal handler (which is what led it to tell the user it "lacks permission").
- **`stripMcpPrefix` / `isAllowedForSession` move to a dependency-free leaf, `services/mcpToolNames.ts`.** `actionIntents/secretBearingTools.ts` was keeping a hand-synced copy of `stripMcpPrefix` purely to dodge an import cycle, with a comment noting that a drifted copy **fails OPEN** on the plaintext-secret guard. Both now share one implementation; the leaf has no imports, so there is no cycle to dodge.
- **`scriptAi.ts` takes the MCP server name from the same constant** that builds the allowlist and the exposed names, so the SDK's `mcp__<name>__` prefix cannot drift from either.

## The allowlist is not widened

This is the risk inherent in option 1, so it is pinned directly rather than argued:

- `isAllowedForSession('run_script', SCRIPT_BUILDER_MCP_TOOL_NAMES)` is still **`false`** — and so is `mcp__breeze__run_script`.
- Through `createSessionPreToolUse`: a session whose allowlist holds only `mcp__script_builder__execute_script_on_device` still denies `('run_script', {})` called **without** an exposed name, and still denies an exposed name it never granted (`mcp__script_builder__take_screenshot`), with `checkGuardrails` and `db.insert` never reached in either case.
- Guard-test cases for `execute_command`, `mcp__breeze__execute_command`, `manage_devices`, `m365_reset_password`.

No capability change: `run_script`'s input schema (`scriptId`, `deviceIds` ≤ 10, `parameters`) is identical to the `execute_script_on_device` shape the model was already constrained to, and the tool stays tier 3 / approval-gated.

## Tests (red first)

`scriptBuilderTools.guard.test.ts` now drives the **real registered handlers** — the tool array is exported for exactly this — with a recording pre-hook, and asserts per tool that the name handed to the guardrail is one the session allowlist contains. On the pre-fix code this fails with:

```
execute_script_on_device: the guard checks 'run_script' against a session allowlist
that holds 'mcp__script_builder__execute_script_on_device' — it will deny a tool the
session granted: expected false to be true
```

It also pins, against the real registration, each tool's handler identity *and* its exposed identity, plus that the registered tool set matches the tier map the allowlist is derived from — so a future renamed tool cannot reintroduce this silently. `aiAgentSdk.test.ts` covers the same contract at the `createSessionPreToolUse` level.

A repo sweep of all 111 `tool(…)`/`makeHandler(…)` registrations across `aiAgentSdkTools.ts` and `scriptBuilderTools.ts` found no other exposed/handler mismatch.

**Verification** (worktree off `origin/main`, merged with latest `main`):
- `vitest run src/services/scriptBuilderTools src/services/aiAgentSdk.test.ts src/routes/scriptAi` — 5 files, 199 passed
- Wider affected surface (`aiToolNames`, `secretBearingTools`, `aiAgentSdkTools`, `aiAgents/runLoop`, `clientSessionTools`, `helperToolFilter`) — 22 files, 481 passed
- `npx tsc --noEmit` on `apps/api` — clean
- `npx eslint` on all changed files — clean

## Out of scope

PR #4888 (held) adds a `runAs` field to the `execute_script_on_device` schema in this same file. Deliberately not touched here.

Closes #4883

### DIFF
--- a/apps/api/src/routes/scriptAi.ts
+++ b/apps/api/src/routes/scriptAi.ts
@@ -26,7 +26,11 @@ import {
   approveToolSchema,
   scriptBuilderContextSchema,
 } from '@breeze/shared/validators/ai';
-import { createScriptBuilderMcpServer, SCRIPT_BUILDER_MCP_TOOL_NAMES } from '../services/scriptBuilderTools';
+import {
+  createScriptBuilderMcpServer,
+  SCRIPT_BUILDER_MCP_SERVER_NAME,
+  SCRIPT_BUILDER_MCP_TOOL_NAMES,
+} from '../services/scriptBuilderTools';
 import { captureException } from '../services/sentry';
 import { db } from '../db';
 import { aiSessions, aiMessages } from '../db/schema';
@@ -237,7 +241,7 @@ scriptAiRoutes.post(
         // Custom MCP server factory for script builder tools
         (getAuth, onPreToolUse, onPostToolUse) => ({
           server: createScriptBuilderMcpServer(getAuth, onPreToolUse, onPostToolUse),
-          name: 'script_builder',
+          name: SCRIPT_BUILDER_MCP_SERVER_NAME,
         }),
       );
     } catch (err) {
--- a/apps/api/src/services/actionIntents/secretBearingTools.ts
+++ b/apps/api/src/services/actionIntents/secretBearingTools.ts
@@ -12,6 +12,7 @@
  * enumerate it for fail-closed contract tests without gaining a mutable
  * security registry.
  */
+import { stripMcpPrefix } from '../mcpToolNames';
 import { encryptSecret } from '../secretCrypto';
 import {
   ACTION_INTENT_RESULT_AAD,
@@ -52,22 +53,6 @@ export const SECRET_UNAVAILABLE_TEXT =
 
 /** Prose tripwire for the shape that caused the original Google leak. */
 const PROSE_CREDENTIAL_PATTERN = /Temporary password:\s*(?!\[REDACTED\]|\[redacted\])\S/;
-
-/**
- * Local copy of aiAgentSdk's stripMcpPrefix. Duplicated rather than imported
- * to avoid an import cycle (aiAgentSdk imports this module). Must stay in
- * sync with the canonical algorithm: this predicate gates both the
- * pre-execution refusal (aiAgentSdkTools.ts) and assertNoPlaintextSecret
- * below, so a narrower normalization here (e.g. only literal
- * "mcp__breeze__") lets a differently-prefixed tool name slip past both
- * checks and fails OPEN — reinstating the plaintext-leak class this module
- * exists to close.
- */
-function stripMcpPrefix(toolName: string): string {
-  if (!toolName.startsWith('mcp__')) return toolName;
-  const separatorIndex = toolName.indexOf('__', 'mcp__'.length);
-  return separatorIndex === -1 ? toolName : toolName.slice(separatorIndex + 2);
-}
 
 export function isSecretBearingTool(toolName: string): boolean {
   const bare = stripMcpPrefix(toolName);
--- a/apps/api/src/services/aiAgentSdk.test.ts
+++ b/apps/api/src/services/aiAgentSdk.test.ts
@@ -93,6 +93,8 @@ vi.mock('./aiAgentSdkTools', () => ({
     // tests below stub via checkGuardrails, not this map.
     file_operations: 1,
     get_device_details: 1,
+    // #4883: the handler behind script builder's `execute_script_on_device`.
+    run_script: 3,
   },
   BREEZE_MCP_TOOL_NAMES: [],
 }));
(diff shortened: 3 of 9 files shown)


ToddHebebrand:

**Findings:** 0 CRITICAL. 4 IMPORTANT + 7 suggestions raised → all actioned in d761df8; 0 outstanding.

| Agent | Verdict |
|---|---|
| code-reviewer | 0 critical, 0 important. Independently verified the guard is not widened (`mcpToolName` is only ever a registration-time literal, never derived from tool args or model output), that the identity split holds everywhere downstream (`checkGuardrails`, `checkToolPermission`, `checkToolRateLimit`, the `ai_tool_executions` insert and the helper-approval path all still use the handler name), that two-argument callers are byte-for-byte unchanged, and that `createBreezeMcpServer` has no split-identity tool to begin with. |
| pr-test-analyzer | 0 critical, 0 important. **Mutation-tested** the coverage in a disposable worktree: reverting `makeExistingHandler` to two arguments reddens the two guard-test cases; making `createSessionPreToolUse` ignore `mcpToolName` reddens the `aiAgentSdk.test.ts` positive case. Confirmed the anti-widening assertions cannot pass for the wrong reason (they match the exact denial string, so an `Unknown tool:` fallthrough would surface as a mismatch). |
| silent-failure-hunter | 0 critical (tried and failed to construct a fail-open input for `mcpToolName ?? toolName`; an empty string fails **closed**). 1 important — see below. |
| comment-analyzer | 0 critical, 3 important (comment accuracy), 4 suggestions. |

**Changes made in response (d761df8):**

- **Observability for the denial itself** (silent-failure-hunter). An allowlist denial returned a plain value with no log and no Sentry event — which is precisely why this bug read as an ordinary tool refusal in chat and sat unnoticed. The SDK is handed the *same* list as `allowedTools` on `query()`, so this branch firing means the SDK and the app-layer guard disagree: a wiring bug, not a user-permission outcome. It now logs and reports, with the detail in the message rather than in Sentry tags (the scrubber's tag allowlist silently voids unknown keys).
- **Two comments that were literally false.** "Nothing below this line uses `exposedToolName`" sat three lines above two uses of it — reworded to the intended claim. `mcpToolNames.ts` claimed both helpers had been copy-pasted into two modules; only `stripMcpPrefix` was, and the two consumers it gates are named again (that specificity was lost when `secretBearingTools.ts`'s docblock went).
- **A name collision.** `mcpToolName` meant the fully-qualified `mcp__…` string in `PreToolUseCallback` but the bare name in `makeExistingHandler`. Renamed the latter to `registeredToolName`; both docblocks now state which form they take, and `SCRIPT_BUILDER_HANDLER_BY_MCP_TOOL` says its keys are bare.
- **Scoped the invariant claim.** "every tool except `execute_script_on_device`" is true today but only *enforced* for script builder, so the docblock now says so and tells a future author of a renamed `breeze` tool to wire the argument.
- **`mcpToolNames.ts` now has its own test** — including the malformed-prefix and over-stripping cases a narrowed normalization would fail open on.
- **Added the negative case that actually discriminates.** `rejects an exposed name the session never granted` does not, because there the handler and exposed names agree. A session granting only `mcp__script_builder__run_script` must still be denied `execute_script_on_device` — mutation-checked: pinning `exposedToolName = toolName` turns it red.

**Not actioned, deliberately:** a repo-wide static invariant test asserting every `tool()` name equals its handler name. A manual block-scoped scan of all 111 registrations in `aiAgentSdkTools.ts` + `scriptBuilderTools.ts` found zero other mismatches, and a regex over source is brittle enough to under-scan silently. The claim is scoped in the docblock instead.

**Tests:** `vitest run src/services/mcpToolNames src/services/scriptBuilderTools src/services/aiAgentSdk.test.ts src/services/actionIntents/secretBearingTools src/routes/scriptAi` — 8 files, 229 passed. Wider affected surface (`aiToolNames`, `aiAgentSdkTools`, `actionIntents/*`, `workerEntrypointClosure.contract`, `aiAgents/runLoop`, `clientSessionTools`, `helperToolFilter`) — green. `tsc --noEmit` on `apps/api` clean; `eslint` on all changed files clean.

**CI:** 65/65 green on `d761df8` (Test API, Test Web, Type Check, Lint, all 4 Integration Tests shards).

**Known cosmetic follow-up, not fixed here:** the approval card and the `ai_tool_executions` row show the handler name `run_script`, while the assistant calls the tool `execute_script_on_device`. That is the correct identity for the audit row, so the fix belongs in a UI display-name mapping rather than in the guard — filing separately if you want it.

**Status:** review-clean, CI green, awaiting maintainer merge. Not merged and #4883 not closed — those are yours.

Issue #4090, opened 2026-08-26
Remote Terminal: multi-line paste/input garbles; opening BRD kills active terminal session


SemoTech:
## Summary

Two Remote Terminal UX/reliability problems on self-host **v0.108.0** against a macOS agent **0.108.0**:

1. **Multi-line paste / typed command blocks arrive interleaved and garbled** in the PTY (commands overlap; prompt fragments mix into the script). Same class of pain as historical #3163 / #2870, still reproducible enough to break field diagnostics.
2. **Starting Breeze Remote Desktop while a Remote Terminal is connected tears down the terminal session** (“Failed … closed unexpectedly” / reconnect required). Techs often need terminal + desktop together for capture/inventory validation.

## Repro (1) — garble

1. Open **Remote Terminal** to a Mac.
2. Paste a multi-line block (several `echo` / `system_profiler` / `sudo tail` lines).
3. Observe lines overlapping (“Connected Bang”, partial commands concatenated, prompts mid-script).

## Repro (2) — BRD kills terminal

1. Remote Terminal connected and usable.
2. From the same device, launch **Remote Desktop**.
3. BRD connects; Remote Terminal session dies and must be re-established after BRD closes.

## Expected

- Pasted multi-line input should apply atomically (or at least without interleaving with other writers).
- Remote Terminal and Remote Desktop should be able to coexist, or the UI should clearly warn and confirm before killing the terminal session.

## Context

Found while running Todd’s #3380 display-inventory validation without ARD. BRD itself worked with lid open; these issues made collecting logs harder than they needed to be.

ToddHebebrand:
Traced both halves against the current agent code — one has a surprising answer, the other needs one discriminating detail from you.

**(1) Multi-line paste garble:** the #2870/#3163 interleaving class was fixed by serializing all terminal commands through a single-consumer FIFO (`agent/internal/websocket/client.go:507-524`, `orderedCommandPump`) with PTY writes additionally serialized under a mutex (`agent/internal/terminal/terminal.go:201-234`), and that fix shipped around v0.103 — it is definitely in your 0.108.0 agent. The web Remote Terminal also sends a paste as **one** WebSocket data message, so there is nothing left to interleave on that path.

[...]

PR #4893, opened 2026-09-04
fix(web): survive a failed xterm stylesheet preload on a cold Terminal mount (#4152)

### LINKED ISSUE #4152: Remote Tools Terminal: welcome shows "Connected Bang"; tab often won't connect until visiting File Browser first
## Summary

Two web **Remote Tools → Terminal** UX bugs on self-host **v0.108.0** / macOS agent **0.108.0** (same canary as #4090):

1. After connect, the session prints a welcome line that reads **`Connected Bang`** (looks like a corrupted or wrong “Connected!” / status string).
2. Opening the **Terminal** tab often **fails to establish** a usable session until the tech navigates to **File Browser** (or another Remote Tools tab) and then **returns to Terminal** — then it connects.

## Repro (2) — tab order

1. Device → Remote Tools.
2. Click **Terminal** first (cold).
3. Observe: stuck / no clean session, or flaky connect (sometimes mid-type refresh wiping input — see #4090).
4. Click **File Browser**, then click **Terminal** again.
5. Observe: Terminal then connects.

## Repro (1) — welcome string

On a successful connect, banner area includes:
```text
Welcome
Connecting to Bens-MacBook-Air...
Connected Bang
```

Expected something like `Connected!` / `Connected` without the stray `Bang`.

## Impact

Blocks orderly field diagnostics (paste/repro work on #4090). Techs waste time “priming” the Terminal tab via File Browser.

## Related

- #4090 (web BRT multi-line paste garble + BRD coexistence)
- Historical terminal garble class #3163 / #2870 (FIFO claimed fixed by 0.103+; residual UX still present)

ToddHebebrand:
## Root cause

A first visit to the Remote Tools **Terminal** tab landed on a permanently "Disconnected" pane — no session, no welcome banner, no Reconnect button. Switching to *any* other tab and back fixed it, which is what made this look like a File Browser warm-up problem for several rounds. It is neither: it is a single non-retryable `await` inside terminal init.

`apps/web/src/components/remote/RemoteTerminal.tsx:155` awaited the stylesheet bare:

```ts
await import('@xterm/xterm/css/xterm.css');
```

Vite compiles that dynamic CSS import into a `<link rel=stylesheet>` injection that resolves on `load` and **rejects on `error`**. It does reject in the field: after an upgrade a stale hashed asset URL comes back as the SPA's `index.html`, and `nosniff` refuses to treat `text/html` as a stylesheet. The reporter's console capture on #4090 showed exactly that (`Failed to preload xterm CSS`, MIME `text/html`, `nosniff`).

That rejection unwound the rest of `initTerminal`, into a catch (`:221-224`) that only logged and called an `onError` callback `RemoteToolsPage.tsx:990-995` never passed. Everything after the await was skipped — `terminalRef` (`:199`) was never assigned and `setTerminalReady(true)` (`:220`) never ran — which produces each reported symptom:

- **No session, no network calls.** Auto-connect (`:642-651`) is gated on `terminalReady`.
- **No Reconnect button.** The overlay (`:778`) requires `failed || (disconnected && autoConnectAttempted)`, and `autoConnectAttempted` is only set by the auto-connect that never fired. Empty pane, nothing to click.
- **Any tab round-trip fixes it.** `RemoteToolsPage.tsx:990` renders the terminal only while its tab is active, so leaving and returning is a full remount, and Vite's preload helper memoises attempted deps in a module-level `seen` map:

  ```js
  if (dep in seen) return;   // vite/dist/node/chunks/node.js — `seen` lives for the page's lifetime
  seen[dep] = true;
  ...
  if (isCss) return new Promise((res, rej) => {
    link.addEventListener('load', res);
    link.addEventListener('error', () => rej(new Error(`Unable to preload CSS for ${dep}`)));
  });
  ```

  On the remount the dep is already in `seen`, so the helper returns immediately and init completes — **with the stylesheet still not loaded.** `initStartedRef` (`:146`) is cleared only in unmount cleanup (`:618`), so nothing short of a remount could retry.

Not the `#4161` hostname-prop race (that fixed a re-init race *within* a mount; a tab switch is a remount either way).

## Changes

1. **The stylesheet ships inside the component's chunk** — `import('@xterm/xterm/css/xterm.css?inline')` hands over the CSS as a string, injected once as `<style id="xterm-css">`. This removes the failure class instead of surviving it: there is no separate asset to go stale, and the CSS cannot arrive without the JS that needs it. A `try/catch` remains as defence in depth, so a styling problem can never again take the session down.
2. **An init failure is now visible** — the catch releases `initStartedRef` and sets status `'failed'`, so the existing overlay renders instead of an inert pane.
3. **The overlay button recovers both broken states** — `handleOverlayRetry` re-runs `initTerminal()` when no terminal exists (the old `onClick={connect}` early-returned at `:229` on the null `terminalRef`, so the button was a no-op in exactly this case); auto-connect takes over once `terminalReady` flips. When a terminal does exist it still just reconnects.
4. **`RemoteToolsPage` passes `onError`** — a failure raises an error toast via the shared `showToast` helper (the pattern `ConnectDesktopButton` in the same directory uses). `RemoteTerminalPage` already passed one; `RemoteToolsPage` was the only mount site that did not.

### Why inlining, and not just swallowing the rejection

The first revision of this PR only added `.catch()`. Review established that `xterm.css` (7112 bytes, v6.0.0) is **not** decorative — it positions `.xterm-viewport`, absolutely stacks the `.xterm-screen` canvases so cursor/selection composite over the text, and hides `.xterm-helper-textarea`, the offscreen textarea that captures keyboard/IME input. Swallowing the rejection would have traded a visibly dead terminal for a silently mis-rendered one that still reports "Connected". Note the `seen`-map behaviour above means the reporter's known-good post-round-trip terminal *was* that styleless state.

Verified against real production builds of `apps/web`, grepping for a CSS-only marker (`.xterm-underline-5`):

| build | where the vendor CSS rules land |
|---|---|
| bare `import '…/xterm.css'` | `dist/server/entry.mjs` only — Astro's stylesheet machinery, i.e. the separate `<link>` asset that 404s |
| `import '…/xterm.css?inline'` | `dist/client/_astro/xterm.Bidc9Bbs.js` only — a client JS chunk. **No stylesheet asset is emitted at all.** |

CSP is satisfied: `apps/web/astro.config.mjs` already sets `styleDirective: ["'self'", "'unsafe-inline'"]`, with a comment saying it carries that directive *for* xterm's runtime inline styles.

## Test evidence

New file `apps/web/src/components/remote/RemoteTerminal.initFailure.test.tsx`, plus cases in `RemoteTerminal.test.tsx` and `RemoteToolsPage.test.tsx`.

The init-failure cases live in a separate file for two verified reasons: a factory-mocked module is evaluated **once per test file** and `vi.resetModules()` does **not** re-run the factory (probed directly — an injected rejection never reached the component and the test passed vacuously), and the `#4510` layout cases in `RemoteTerminal.test.tsx` render without awaiting init, so their `new Terminal(...)` lands during a later test and would consume an injected failure.

**Red — against the unfixed component:**

```
 × treats the failed stylesheet preload as cosmetic and still connects
   AssertionError: expected [] to have a length of 1 but got +0
 × renders the retry overlay instead of an inert empty pane
   Unable to find an element by: [data-testid="terminal-disconnect-overlay"]
 × re-runs init when the retry button is clicked, then connects
   Unable to find an element by: [data-testid="terminal-disconnect-overlay"]
 Tests  3 failed | 1 passed (4)
```

```
 × passes an onError handler that raises a toast
   AssertionError: expected undefined to be type of 'function'
 Tests  1 failed | 6 passed (7)      # RemoteToolsPage.test.tsx
```

The stylesheet case asserts `cssImport.evaluations === 1` and `threw === true`, so it cannot pass for the trivial reason that the stylesheet loaded.

**Green — with the fix:**

```
apps/web $ npx vitest run src/components/remote/
 Test Files  21 passed (21)
      Tests  158 passed (158)      # 25 across the two RemoteTerminal files

apps/web $ npx tsc --noEmit               # clean, exit 0
apps/web $ npx eslint <touched files>     # clean, exit 0
apps/web $ npx astro build                # clean, artifacts inspected per the table above
apps/web $ npx vitest run src/lib/__tests__/no-silent-mutations.test.ts
      Tests  130 passed (130)
```

No migrations, no schema, no tenancy surface touched — `apps/web` only.

Closes #4152

https://claude.ai/code/session_011zx46C9kvzFoSNFWS7LZZf

### DIFF
--- a/apps/web/src/components/remote/RemoteTerminal.initFailure.test.tsx
+++ b/apps/web/src/components/remote/RemoteTerminal.initFailure.test.tsx
@@ -17,18 +17,23 @@ import { fetchWithAuth } from '@/stores/auth';
 //     so their `new Terminal(...)` lands during a later test. Sharing that file
 //     would let a leaked construction consume this file's injected failure.
 //
-// What is reproduced: Vite compiles the component's dynamic CSS import into a
+// What is reproduced: before the fix the component did a bare
+// `await import('@xterm/xterm/css/xterm.css')`, which Vite compiles into a
 // `<link rel=stylesheet>` injection that resolves on `load` and REJECTS on
 // `error`. After an upgrade a stale hashed asset URL comes back as the SPA's
-// index.html, which `nosniff` refuses to treat as a stylesheet — so the
-// promise rejects on a cold mount. Before the fix that single rejection
-// unwound the whole of initTerminal: no terminal instance, no `terminalReady`,
-// therefore no auto-connect and — because the retry overlay is gated on
+// index.html, which `nosniff` refuses to treat as a stylesheet, so the promise
+// rejected on a cold mount. That single rejection unwound the whole of
+// initTerminal: no terminal instance, no `terminalReady`, therefore no
+// auto-connect and — because the retry overlay is gated on
 // `autoConnectAttempted`, which only the auto-connect sets — nothing to click.
 // The pane sat on "Disconnected" until a tab round-trip remounted it.
+//
+// The stylesheet is now inlined into the component's own chunk (`?inline`), so
+// that `<link>` no longer exists. These cases hold the line behind that: the
+// module still has to be unable to take the session down with it.
 const cssImport = { evaluations: 0, threw: false };
 
-vi.mock('@xterm/xterm/css/xterm.css', () => {
+vi.mock('@xterm/xterm/css/xterm.css?inline', () => {
   cssImport.evaluations += 1;
   // Only the first evaluation fails, so the stylesheet-specific case below can
   // prove the rejection reached the component. Vitest caches the outcome
@@ -39,7 +44,7 @@ vi.mock('@xterm/xterm/css/xterm.css', () => {
       "Unable to preload CSS for /assets/xterm-abc123.css (MIME type 'text/html' is not a supported stylesheet MIME type)",
     );
   }
-  return {};
+  return { default: '.xterm { position: relative; }' };
 });
 
 const terminalLifecycle = {
@@ -213,9 +218,44 @@ describe('RemoteTerminal init failure is visible and retryable in place (#4152)'
     // Init succeeding hands off to the ordinary auto-connect path.
     await waitFor(() => expect(sessionPostCount()).toBe(1), { timeout: 3000 });
     await waitFor(() => expect(MockWebSocket.instances).toHaveLength(1), { timeout: 3000 });
+    // And the 'failed' status set by the dead init must not outlive it —
+    // otherwise the overlay would sit on top of a live session.
+    await waitFor(
+      () => expect(screen.queryByTestId('terminal-disconnect-overlay')).not.toBeInTheDocument(),
+      { timeout: 3000 },
+    );
   });
 
-  it('reports the failure to the caller through onError', async () => {
+  it('stays retryable when the retry itself fails', async () => {
+    // The guard that makes a retry possible at all is `initStartedRef` being
+    // released in the catch. It has to be released on EVERY failure, not just
+    // the first, or a second bad attempt wedges the pane exactly as #4152 did.
+    terminalLifecycle.failNextConstructs = 2;
+
+    render(<RemoteTerminal deviceId="device-1" deviceHostname="host-1" />);
+    await screen.findByTestId('terminal-disconnect-overlay', undefined, { timeout: 3000 });
+
+    await userEvent.click(screen.getByTestId('terminal-overlay-reconnect'));
+    await waitFor(() => expect(terminalLifecycle.constructCount).toBe(2), { timeout: 3000 });
+
+    // Still offering a way out, and still no session for a terminal that does
+    // not exist.
+    expect(await screen.findByTestId('terminal-disconnect-overlay', undefined, { timeout: 3000 })).
+      toBeInTheDocument();
+    expect(sessionPostCount()).toBe(0);
+
+    // Third attempt succeeds, proving the guard never latched.
+    await userEvent.click(screen.getByTestId('terminal-overlay-reconnect'));
+    await waitFor(() => expect(terminalLifecycle.constructCount).toBe(3), { timeout: 3000 });
+    await waitFor(() => expect(sessionPostCount()).toBe(1), { timeout: 3000 });
+  });
+
+  // NOT new coverage for #4152 — the catch has always called onError. This
+  // pins that contract because RemoteToolsPage now depends on it to raise a
+  // toast (see RemoteToolsPage.test.tsx); before, the one caller that mattered
+  // passed no handler, so the call had no observable effect and could have
+  // been dropped by a refactor without anything failing.
+  it('reports the failure to the caller through onError (pre-existing contract)', async () => {
     terminalLifecycle.failNextConstructs = 1;
     const onError = vi.fn();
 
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (all three, in parallel, against `4729e539c`).

**Findings:** 4 raised → all addressed in `475db2dfb`; 0 outstanding.

1. **code-reviewer + silent-failure-hunter, independently (the important one):** swallowing the CSS-preload rejection traded a dead terminal for a silently mis-rendered one. Both pulled the shipped `@xterm/xterm@6.0.0` stylesheet and showed it is load-bearing, not decorative — it positions `.xterm-viewport`, absolutely stacks the `.xterm-screen` canvases, and hides `.xterm-helper-textarea` (the offscreen textarea that captures keyboard/IME input). A terminal reporting "Connected" with a stray focus-stealing textarea is arguably worse than a visibly dead one.

   **Accepted, and taken further than the swallow.** The stylesheet now ships inside the component's own chunk via `?inline`, injected once as `<style id="xterm-css">` — the failure class is removed rather than survived. Verified against real `astro build` output, grepping a CSS-only marker (`.xterm-underline-5`): with the bare import the vendor rules land only in `dist/server/entry.mjs` (Astro's stylesheet machinery → the separate `<link>` asset that 404s); with `?inline` they land only in `dist/client/_astro/xterm.Bidc9Bbs.js`, a client JS chunk, and **no stylesheet asset is emitted at all**. CSP checked: `astro.config.mjs` already sets `style-src 'self' 'unsafe-inline'`, with a comment saying it carries that directive *for* xterm.

   I did **not** take the silent-failure-hunter's suggested variant (drop the `.catch()` and let it throw into the retry overlay). Vite's preload helper memoises attempted deps in a module-level `seen` map, so the retry would short-circuit and complete *without* the stylesheet — the user would click through an error to reach the exact degraded render the finding objects to, and the first mount would still fail. The `try/catch` stays as defence in depth, with its regression test.

2. **pr-test-analyzer:** the `onError` case in the new file re-asserts behaviour that already existed on `main` and is not new coverage for this fix. → renamed and comment-corrected to say so plainly; it now documents *why* it is still worth pinning (RemoteToolsPage newly depends on that call).
3. **pr-test-analyzer:** a stale `'failed'` status surviving a successful retry was only caught indirectly, via a downstream network timeout. → added a direct assertion that the overlay clears after the retry succeeds.
4. **pr-test-analyzer:** no coverage for a *second* consecutive init failure, i.e. whether the `initStartedRef` release is one-shot. → added a case that fails twice, retries three times, and proves the guard never latches.

Also verified by hand after the change: the injected `<style>` lands after `globals.css`, so I checked the cascade — the app's two `.xterm-*` overrides are `!important` (they win regardless of order) and vendor CSS sets no `height` on `.xterm`, so the app's non-`!important` `height:100%` is uncontested. No conflict.

**Tests:** `apps/web` remote suite green — `npx vitest run src/components/remote/` → 21 files, 158 passed (25 across the two `RemoteTerminal` files). `tsc --noEmit` clean; `eslint` on the touched files clean; `astro build` clean; `no-silent-mutations` 130 passed. Red-first evidence for every new case is in the PR body.

**Not re-reviewed:** per the one-round rigor cap I did not dispatch a second review of `475db2dfb`. That commit changes how a vendor stylesheet is delivered in the production bundle, so if you would rather have a fresh pair of eyes on that specific hunk before merging, say the word.

**Status:** review-clean, awaiting maintainer merge. Not merged, issue #4152 left open per the hand-off rules.

Issue #4885, opened 2026-09-04
[UI] No "Run again" on a script execution — retry requires re-picking device and parameters from scratch


ToddHebebrand:
## Description
After a script run fails (or you just want to re-run it), there is no retry action anywhere the result is shown. The user has to go back to the library or device page, open the full run modal, and re-select target and parameters. During the OliveTech parameter debugging (#4882 / #4884) the customer re-ran one script a dozen times this way.

Verified surfaces with no retry:
- `apps/web/src/components/scripts/ExecutionDetails.tsx:298-304` — execution result modal used by `ScriptExecutionsPage` and `ScriptsPage`; footer is Close only.
- `apps/web/src/components/devices/DeviceScriptHistory.tsx:457-464` — device page Scripts tab detail modal; Close only.
- `apps/web/src/components/scripts/ScriptTestRunner.tsx:493-501` — the one "Retry" button (`test-poll-retry`) re-polls the same execution after a fetch failure; it does not start a new run.
- The Script Builder chat / AI `run_script` results are inline only.

No new API is needed: every run already funnels through `POST /scripts/:id/execute` (`apps/api/src/routes/scripts.ts:977`), and `script_executions` stores `parameters`, `device_id`, `script_id` (and runAs if present on the row).

## Proposed Fix
Add a "Run again" action to both detail modals (and the row actions in `ScriptExecutionsPage` / `DeviceScriptHistory` lists) that:
- calls `POST /scripts/:scriptId/execute` with the original `deviceIds: [execution.deviceId]`, `parameters`, and `runAs` pre-filled,
- offers an "Edit and run" variant that opens `ScriptExecutionModal` pre-populated instead of blank (parameters editable, device pre-selected),
- shows the new execution via the same modal / list refresh, wrapped in `runAction` per the mutation-feedback contract.
- Script Builder `ScriptTestRunner`: add a real "Run again" beside the poll retry.

## Affected Files
- `apps/web/src/components/scripts/ExecutionDetails.tsx`
- `apps/web/src/components/devices/DeviceScriptHistory.tsx`
- `apps/web/src/components/scripts/ScriptExecutionModal.tsx` (accept initial values)
- `apps/web/src/components/scripts/ScriptTestRunner.tsx`

## Reported By
@toddhebebrand, 2026-09-04 (nitpick batch).

PR #4906, opened 2026-09-05
feat(scripts): make the script run context selectable and visible everywhere (#4888)

### LINKED ISSUE #4888: [UI][AI] Run-as (System / Logged-in user) is only selectable from two launch surfaces — Test Run, AI tools, Automations form, and Fix flow silently use the saved default

ToddHebebrand:
Closes #4888

Run-as (System / logged-in user) was selectable from exactly two of the seven surfaces that can launch a script, and **nothing anywhere showed which context a run actually used**. During the OliveTech GCPW debugging (#4882) the same script ran alternately as SYSTEM and as the user with no visible control over which — which is a large part of why the failures looked random.

This makes the run context selectable everywhere it can be, honoured everywhere it is offered, and visible after the fact.

## What changed

| Surface | Before | After |
|---|---|---|
| Script editor **Test Run** | no control; inherited the form default silently | run-context select (defaults to **Script default**), session target for an on-demand helper, effective context shown in the result header |
| **Automations** `run_script` action | backend honoured `action.runAs`; the form never set it | run-as override on the action, defaulting to "Script default" |
| Fleet **Fix flow** | rendered a select and **discarded it** (`_runAs`) | choice is persisted on the run and read at dispatch; the picker's contextless select is hidden in favour of one that can inherit the default |
| **AI tools** (`run_script`, `execute_script_on_device`) | no way to express a run context | optional `runAs` + `targetSessionId`, validated by the same schema a human request clears — see the privilege section below |
| **Execution history** (details, device Scripts tab, Test Run header) | never recorded, never shown | `script_executions.run_as` / `target_session_id`, surfaced in all three |

`elevated` stays **non-overridable at launch time** on every path — it remains a property of the saved script. Every launch-time enum is `['system','user']`, and each is the same object or a documented mirror of `executeScriptSchema`.

## The "Script default" option is load-bearing, not decoration

Two of these surfaces could have shipped a plain System/User select. They deliberately don't: the launch-time enum has no `elevated`, so a control that always posts a concrete value turns an `elevated` script's next run into a plain `system` run. That is a **quieter** bug than the discarded-select this issue reports, and a worse one. Both Test Run and the Fix flow default to "Script default" and **omit `runAs` from the request entirely** while it is selected, so the server keeps resolving `scripts.run_as`. Tests pin the omission, not just the presence.

## Letting the assistant choose `runAs` — the privilege decision

This is a real capability change: an assistant can now ask for SYSTEM on a script whose saved default is the logged-in user. Todd approved it on the condition that the approver is told. Concretely:

- **No bypass.** The handler re-parses `{deviceIds, runAs, targetSessionId}` through the *actual* `executeScriptSchema` object the HTTP route uses (moved to `services/scriptRunRequest.ts` and re-exported from `routes/scripts.ts`, so there is one definition, not a copy). `elevated`, a `targetSessionId` without `runAs: 'user'`, and a `targetSessionId` on more than one device are all rejected before dispatch.
- **Same tier, same approval.** `run_script` / `execute_script_on_device` stay Tier 3, so the value rides on the pinned action-intent and the durable release worker re-executes with exactly what was approved. Adding the fields does not disturb the #4883 name split: the tool is still registered as `execute_script_on_device`, the session allowlist is still checked on that exposed name, and the handler still resolves to `run_script`.
- **The approval card names the effective context.** `resolveScriptRunContextForApproval` resolves override-else-script-default (fail-soft: a DB hiccup degrades to vaguer wording, never to a failed approval) and feeds two carriers, because they reach different people:
  - a **structured** `scriptRunContext` on the `approval_required` event → the web card renders an always-visible row, styled as a warning when the assistant overrode the default: *"Run context: **System** — chosen by the assistant, overriding the script's saved default (Logged-in user)"*;
  - the **same sentence appended to `description`** → which becomes the intent's stored `reason`, i.e. what the /approvals queue, the mobile push, and the Script Builder's own approval strip show.

  Previously `runAs` would have appeared only as a line of JSON inside the collapsed "Show parameters" blob, if at all. Four component tests plus two `aiAgentSdk` tests pin this; the mutation check (deleting the row / the enrichment) reddens them.
- **What the card says stays true until release.** Both halves are already TOCTOU-pinned and needed no new work: an assistant-chosen `runAs` rides on the intent's `input` and is covered by `argument_digest`, and the script's saved default is inside `runScriptDigestMaterial` (`script.runAs`, `v: 2` envelope). So a script whose default is flipped from `user` to `system` inside the approval window fails the release closed with `content_changed` rather than executing at a privilege the approver never saw. Worth stating explicitly, because "the card names the script default" would be worth much less if the default were mutable behind it.

### Finding: an assistant could reach ELEVATED sideways, through a config policy (fixed here)

Review turned up a genuine hole in this PR's own stated invariant, and it is the one worth reading closely:

> `manage_policy_feature_link` is **Tier 2** — it auto-executes with no human approval. Its `automation` feature type is not schema-validated at the tool layer, and `normalizeAutomationActions` deliberately tolerates `runAs: 'elevated'` (it also runs on the *execute* path and must not take a live automation offline over a stored value). So an assistant could author an automation action that runs a script as administrator/root — precisely the capability the same assistant is refused head-on, where `executeScriptSchema` excludes `elevated` **and** a Tier-3 human approval is required.

`rejectElevatedAutomationActions` (`services/aiToolsConfigPolicy.ts`) now refuses it on both `add` and `update`, fail-closed, before any write. Five tests, mutation-checked. This is squarely in scope: "an assistant must not be able to ask for a run context the API would refuse a person" is the whole condition this feature was approved under, and reaching it sideways still reaches it.

**What is NOT closed, deliberately:** a raw `POST /automations` call still reaches the same tolerant `normalizeAutomationActions` (`routes/automations.ts` types `actions` as `z.unknown()`). Closing that safely means distinguishing a *newly submitted* `elevated` from one *already stored on the row being edited* — otherwise you either regress a legitimate existing automation or silently downgrade it on an unrelated save. That is a design decision plus a product call ("may an automation action request elevation at all?"), not a guard, and it pre-dates #4888 by however long the field has existed. **Worth its own issue.** Called out rather than quietly normalised, and the comment in `scriptRunRequest.ts` now scopes the `elevated` claim to the paths where it actually holds instead of implying it holds everywhere.

### Finding: the AI path skips maintenance windows (pre-existing, NOT fixed here)

Auditing "does the AI path reach dispatch with fewer checks than a human?" turned up one genuine divergence, and it is **not** about `runAs`:

> `services/scriptExecution.ts` (the HTTP `POST /scripts/:id/execute` path) suppresses a device whose maintenance window has `suppressScripts` set. **`aiToolsScripts.run_script`, `automationRuntime`, and `fleetFindings/dispatch` do not** — `dispatchScriptToDevice`'s docstring explicitly says maintenance windows are the caller's job, and three of its four callers never do it.

So an assistant-initiated (or automation-, or fleet-fix-initiated) script run executes on a device a human-initiated run would have been suppressed on. Deliberately left alone: changing when AI/automation runs execute is a behaviour change with its own blast radius and deserves its own issue and decision, not a quiet ride-along in a run-as PR. Flagging rather than normalising.

Two smaller, deliberate asymmetries, for completeness: the AI path substitutes a Tier-3 human approval for the HTTP route's `requireMfa()` (arguably stronger, not weaker), and it writes no `script_execution_batches` row.

## Database

Two additive, idempotent migrations — nullable columns, no backfill:

- `2026-10-09-000400-script-executions-run-as.sql` — `script_executions.run_as`, `.target_session_id`
- `2026-10-09-000500-fleet-remediation-runs-run-as.sql` — `fleet_remediation_runs.run_as`

Both columns are **NULL for pre-existing rows on purpose**: those runs genuinely do not know their context, and the UI renders that as "Not recorded" rather than inventing "System". A test asserts the null case does *not* render "System" — the whole value of the column is that it can be trusted.

Registered in `CORE_TENANT_EXPORT_POLICY` (`included` — plain enum/int classifiers). Both tables were already in the cascade lists; no new tables, so no cascade changes.

## Not done, on purpose

- `ScriptExecutionModal` and `ScriptPickerModal` keep their own inline run-as selects (long-standing `data-testid`s, secrets-blocked warnings, and a device-session picker). New surfaces use the shared `RunContextSelect`; migrating those two is a follow-up, not a prerequisite.
- No run-context indicator in the execution history **list** row — it is already at seven columns.
- "Run again" (#4885) still prefills from the script, not from the recorded run context of the execution being repeated. Now that the value is recorded, that is a natural follow-up.
- The `run_script` vs `execute_script_on_device` display mismatch from #4883 (the card shows the handler name) is untouched. My change does not affect its visibility either way — the run-context row is derived from *both* names, so it renders correctly regardless of which identity the card is holding.

## Verification

- `apps/api`: routes `scripts` / `fleetFindings` / `devices` — 802 passed, 2 skipped (48 files). Services `scriptBuilderTools` / `aiToolsScripts` / `fleetFindings` / `scriptDispatch` / `scriptRunContextApproval` / `scriptExecution` / `automationRuntime` / `aiAgentSdk` / `aiToolSchemas` / `aiGuardrails` / `tenantExportPolicy` — 1221 passed (55 files). `autoMigrate` + `composeBindMounts` — 78 passed.
- `apps/web`: `components/{scripts,devices,automations,fleet,ai}` + `stores` — 1632 passed (118 files). `no-silent-mutations` + `keyUsage` — 134 passed. `localeParity` — green.
- `packages/shared` — 2462 passed (96 files).
- `tsc --noEmit` clean in `apps/api` and `apps/web`; `pnpm lint` clean (6/6).
- Every new behavioural test was mutation-checked: the feature was temporarily removed, the test observed RED, and the source restored. Notable catch during that pass — a first draft of the Script Builder schema assertion checked only `result.success`, which a non-strict zod object satisfies by *stripping* the new keys; it was strengthened to assert the parsed values survive.

Integration-only suites (`tenant-export-policy`, `tenantExportErasureRoundtrip`, RLS) were not run locally — no DB here; they run in CI's Integration Tests job, which blocks this PR.

## Review

`/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer. Six findings, all addressed on the branch:

- **Critical** — the `elevated`-via-config-policy hole above.
- **Critical** — `AutomationEditPage.buildActionPayload` rebuilds each action field by field and dropped `runAs` between the operator clicking Save and the request going out; `normalizeActionForForm` never read it back either. This was #4888's own bug, reintroduced one layer below where the form-level test could see it — and the form-level test would have passed regardless. Both halves fixed, with tests asserting the real PUT body, plus a disabled-option path so a stored `elevated` survives an unrelated edit instead of being silently downgraded.
- **Important** — the `scriptRunContext` seam (store extraction + prop forward) had no test: delete either line and the approver stops being warned with everything still green. Tests added at both points.
- **Inaccurate comment** — the `asRunAs` docstring named the wrong read-path callers; the real one is `jobs/automationWorker.ts` via `createAutomationRunRecord` / `executeConfigPolicyAutomationRun`. Corrected (the fail-soft argument is stronger for the real caller, not weaker).
- **Inaccurate comment** — "(sudo / runas)" overstated Windows, where nothing elevates at launch and the agent service must already be admin. Corrected against `agent/internal/executor/executor.go`.
- **Suggestions** — `elevated` now gets its own chip colour instead of sharing the calm one with `user`; the automations form names the script's actual default in its "Script default (…)" option; the extracted import moved to the top of `routes/scripts.ts`; a `validateToolInput`-layer test added for the first-gate enforcement.


ToddHebebrand:

**Findings:** 6 raised → all 6 addressed in `fe3af9b60`; 0 outstanding.

- **Critical — assistant could reach ELEVATED sideways.** `manage_policy_feature_link` is Tier 2 (auto-executes, no human approval) and its `automation` inline settings are unvalidated, so an assistant could author an automation action running a script as administrator/root — the exact capability it is refused head-on, where `executeScriptSchema` excludes `elevated` *and* a Tier-3 approval is required. Now refused fail-closed on both `add` and `update` (`rejectElevatedAutomationActions`). The raw `POST /automations` half is deliberately **not** closed here and is written up in the PR body as follow-up: doing it safely needs to tell a newly-submitted `elevated` from one already stored on the row being edited, which is a design + product call, and it pre-dates this issue.
- **Critical — `AutomationEditPage.buildActionPayload` dropped `runAs`** between Save and the request. This was #4888's own bug reintroduced one layer below where the form-level test could see it — and that test passed either way. Both the write and read halves fixed; new tests assert the real PUT body, and a stored `elevated` now survives an unrelated edit instead of being silently downgraded.
- **Important — the `scriptRunContext` plumbing seam was untested.** Deleting either the store extraction or the prop forward would stop warning the approver with every other test green. Tests added at both points.
- **Two inaccurate comments** — the `asRunAs` docstring named the wrong read-path callers (the real one is `jobs/automationWorker.ts`), and "(sudo / runas)" overstated Windows, where nothing elevates at launch. Both corrected against the code they describe.
- **Suggestions** — `elevated` given its own chip colour rather than sharing the calm one with `user`; the automations form now names the script's actual default; extracted import moved to the top of `routes/scripts.ts`; a `validateToolInput`-layer test added.

Every fix was mutation-checked: the change was reverted, the new test observed RED, then restored.

**Tests:** `apps/api` 2024 passed / 2 skipped (103 files); `apps/web` 2903 passed (199 files); `packages/shared` 2462 passed (96 files). `tsc --noEmit` clean in both apps; `pnpm lint` clean (6/6); locale parity and `no-silent-mutations` green. Integration-only suites (`tenant-export-policy`, `tenantExportErasureRoundtrip`, RLS) need a live DB and run in CI's blocking Integration Tests job.

**Status:** review-clean, awaiting maintainer merge. Two things worth a decision before or alongside merge, both written up in the PR body: the raw-`POST /automations` elevated path above, and the pre-existing divergence where only the HTTP execute route honours a device's maintenance window (the AI, automation and fleet-fix paths do not).

PR #4914, opened 2026-09-05
fix(agent): run brew cleanup --prune=all after Homebrew upgrade batches (#4912)

### LINKED ISSUE #4912: [Agent] macOS Homebrew patching never runs brew cleanup after upgrade
## Description

The macOS third-party patch provider upgrades Homebrew formulae and casks but never prunes old versions. Outdated kegs and cached downloads accumulate until Homebrew's own periodic cleanup fires (every 30 days by default, only as a side effect of some later `brew` invocation).

## Current behaviour (verified)

`agent/internal/patching/homebrew.go`:
- `Scan` runs `brew outdated --json=v2` (line 171).
- `Install` runs `brew upgrade [--cask] <name>` (lines 221-229).
- No `brew cleanup` anywhere. Repo-wide grep for `brew cleanup`, `HOMEBREW_NO_INSTALL_CLEANUP`, `HOMEBREW_CLEANUP_PERIODIC_FULL_DAYS`, and `--prune` returns zero hits in `agent/`.
- `brewEnv()` (lines 130-142) sets only `HOME` and `PATH`; no cleanup vars.
- Runs as the console user via `sudo -n -H -u <user>` (lines 144-167), which is correct.

## Proposed fix

After a successful `Install` batch (not per package), run `brew cleanup --prune=all` through the same `brewCommand()` path so it executes as the console user. Log but do not fail the job on cleanup errors. Optionally expose `brewCleanup: true` as a patch-source policy setting, default on.

## Affected files

- `agent/internal/patching/homebrew.go`
- `agent/internal/patching/homebrew_test.go`
- `apps/docs/src/content/docs/features/patch-management.mdx` (mention cleanup)

## Reported by

Pre-sales verification, 2026-09-04.

ToddHebebrand:
## Problem

The macOS third-party patch provider (`agent/internal/patching/homebrew.go`) upgrades Homebrew formulae and casks via `Install` but never prunes old kegs and cached downloads. They accumulate until Homebrew's own periodic cleanup fires as a side effect of some later `brew` invocation (every 30 days by default).

## Fix

`HomebrewProvider.Install` now schedules `brew cleanup --prune=all` after a successful upgrade, run through the existing `brewCommand()` path — so it executes as the console user via `sudo -n -H -u` exactly like every other brew invocation in this file. Cleanup failures are logged (`log.Warn`) and never fail the job.

## Design decisions (as asked for in the issue)

- **Batch vs. per-package placement:** `PatchProvider.Install(patchID string)` is per-package with no batch-end hook, and the caller (`executePatchInstallCommand` in `agent/internal/heartbeat/heartbeat.go`) simply loops over `h.patchMgr.Install(installID)` for each patch ref in a job — there's no existing concept of "batch end" to hook into, and neither `apt.go` nor `winget_system.go` (the two providers named in the issue) do any post-install batch-level work to be consistent with. Rather than change the `PatchProvider` interface or the caller (out of scope per the issue's affected-files list), I implemented this entirely inside `homebrew.go`: `Install` calls `scheduleCleanup()`, which debounces on a `*HomebrewProvider`-scoped timer (30s default). A `HomebrewProvider` instance is constructed once and lives for the process lifetime (`NewDefaultManager`), so it's safe to hold this state on the struct. Consecutive `Install` calls for one job's packages arrive well within the debounce window and each resets the timer, so only the last one survives to fire — one `brew cleanup` per batch, not per package. As a side effect, since cleanup runs asynchronously after `Install` already returned, a cleanup failure is structurally incapable of failing the job.
- **No new policy toggle:** the issue's proposed fix optionally suggested a `brewCleanup: true` patch-source setting. There is no existing agent-side patch-source settings slot for Homebrew to hang a toggle off (checked `internal/config` and `internal/patching` — patch policy config lives on the API/server side, not as per-provider agent config), so per the "no new toggle unless there's an obvious slot" guidance, none was added. Cleanup ships unconditionally after a successful upgrade batch.

## Tests

Wrote failing tests first in `agent/internal/patching/homebrew_test.go` (confirmed a compile-failure red before implementing):
- `TestBrewCleanupArgs` — pure arg-construction test for `brew cleanup --prune=all`.
- `TestScheduleCleanupCoalescesABatch` / `TestScheduleCleanupRunsAgainForANewBatch` — deterministic, fast (injected `cleanupFunc`/`cleanupDebounce`, no real `brew`) tests proving the debounce coalesces a batch into exactly one cleanup call, and that a later batch still gets its own call.
- `TestInstallSchedulesCleanupOnSuccess` / `TestInstallDoesNotScheduleCleanupOnFailure` — integration-style tests (skip if brew unavailable / running as root, matching this file's existing convention) exercising the real `Install()` path against an already-installed formula (verified locally that `brew upgrade <up-to-date-formula>` exits 0) and a nonexistent package.

Also added one sentence to `apps/docs/src/content/docs/features/patch-management.mdx` (Third-Party Application Patching section) noting the macOS engine runs `brew cleanup --prune=all` after a batch of upgrades.

## Verification

```
$ cd agent && go vet ./internal/patching/...
(clean, no output)

$ cd agent && go test -race -count=1 ./internal/patching/...
ok  	github.com/breeze-rmm/agent/internal/patching	8.446s
ok  	github.com/breeze-rmm/agent/internal/patching/linuxsession	1.736s
```

Closes #4912

https://claude.ai/code/session_01V51pUwJs69Rmb899D4x8R1

### DIFF
--- a/agent/internal/patching/homebrew.go
+++ b/agent/internal/patching/homebrew.go
@@ -38,6 +38,17 @@ type HomebrewProvider struct {
 	cleanupTimer    *time.Timer
 	cleanupDebounce time.Duration // overridden in tests; <=0 means use defaultCleanupDebounce
 	cleanupFunc     func()        // overridden in tests; nil means use h.runBrewCleanup
+
+	// brewMutateMu serializes brew invocations that mutate the
+	// Cellar/Caskroom (upgrade, uninstall, cleanup) against each other.
+	// scheduleCleanup's debounced cleanup fires on its own timer goroutine,
+	// independent of whatever Install/Uninstall call scheduled it — without
+	// this, a cleanup firing mid-batch (e.g. one package's upgrade running
+	// longer than defaultCleanupDebounce, which patchMutateTimeout allows up
+	// to 30 minutes for) could run `brew cleanup --prune=all` concurrently
+	// with an in-flight `brew upgrade`/`brew uninstall`, and Homebrew does
+	// not guarantee that's safe.
+	brewMutateMu sync.Mutex
 }
 
 // defaultCleanupDebounce is how long scheduleCleanup waits after the most
@@ -245,7 +256,9 @@ func (h *HomebrewProvider) Install(patchID string) (InstallResult, error) {
 	}
 	args = append(args, name)
 
+	h.brewMutateMu.Lock()
 	output, err := h.brewCombinedOutput(patchMutateTimeout, args...)
+	h.brewMutateMu.Unlock()
 	if err != nil {
 		return InstallResult{}, fmt.Errorf("brew upgrade failed: %w: %s", err, truncatePatchOutput(output))
 	}
@@ -272,6 +285,12 @@ func brewCleanupArgs() []string {
 // consecutive Install calls (one job upgrading N formulae/casks) triggers
 // exactly one cleanup run, fired shortly after the last package in the
 // batch finishes — not once per package. Safe to call concurrently.
+//
+// The debounce timer is in-memory only: if the agent process exits between
+// an Install and the timer firing, that scheduled cleanup is silently
+// dropped with no persistence or retry-on-restart. This is an accepted
+// best-effort gap (matching the issue's "log but never fail the job"
+// framing) — the next patch batch schedules its own cleanup regardless.
 func (h *HomebrewProvider) scheduleCleanup() {
 	h.cleanupMu.Lock()
 	defer h.cleanupMu.Unlock()
@@ -304,7 +323,11 @@ func (h *HomebrewProvider) runBrewCleanup() {
 		return
 	}
 
+	// Serialized against Install/Uninstall via brewMutateMu — see its
+	// doc comment on HomebrewProvider for why.
+	h.brewMutateMu.Lock()
 	output, err := runCmdCombinedOutputWithTimeout(cmd, patchMutateTimeout)
+	h.brewMutateMu.Unlock()
 	if err != nil {
 		log.Warn("brew cleanup failed", "error", err, "output", truncatePatchOutput(output))
 		return
@@ -325,7 +348,9 @@ func (h *HomebrewProvider) Uninstall(patchID string) error {
 	}
 	args = append(args, name)
 
+	h.brewMutateMu.Lock()
 	output, err := h.brewCombinedOutput(patchMutateTimeout, args...)
+	h.brewMutateMu.Unlock()
 	if err != nil {
 		return fmt.Errorf("brew uninstall failed: %w: %s", err, truncatePatchOutput(output))
 	}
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter, pr-test-analyzer), sonnet subagents, run in parallel against `gh pr diff` (no live worktree given to reviewers).

**Findings:** 6 raised across the three reviews; all addressed in `916ccbffb`, 0 outstanding (2 accepted as out-of-scope, noted below).

Addressed:
- **[code-reviewer, Important/informational]** The debounced cleanup timer runs on its own goroutine independent of the Install call sequence; a package upgrade outlasting the 30s debounce (plausible — `patchMutateTimeout` allows up to 30 minutes) could let `brew cleanup --prune=all` execute concurrently with an in-flight `brew upgrade`/`brew uninstall`. Fixed by adding `brewMutateMu`, serializing all brew-mutating calls (Install, Uninstall, runBrewCleanup) against each other.
- **[pr-test-analyzer, Important]** `TestScheduleCleanupCoalescesABatch`'s 5ms inter-call sleep against a 20ms debounce was a real flake risk under CI scheduler jitter. Widened to 200ms debounce / 10ms sleeps (20x margin).
- **[pr-test-analyzer, Important]** No test exercised `cleanupMu` under actual concurrent goroutines. Added `TestScheduleCleanupConcurrentCallsAreRaceFree` (20 concurrent `scheduleCleanup` calls under `-race`).
- **[pr-test-analyzer / silent-failure-hunter, Suggestion]** `runBrewCleanup` itself (the non-injected path) had zero direct coverage. Added `TestRunBrewCleanupExecutesRealCleanup`.
- **[silent-failure-hunter, Important]** The in-memory debounce timer has no backstop across an agent restart mid-window. Documented as an accepted best-effort gap directly on `scheduleCleanup`'s doc comment (next batch schedules its own cleanup regardless).

Accepted as out-of-scope (noted, not changed):
- **[silent-failure-hunter, Important]** No operator-visible signal (metric/audit row) if cleanup fails persistently, beyond the local agent log. The issue's spec is explicitly "log but never fail the job" with no policy toggle; adding telemetry is a larger surface than this fix warrants.
- **[silent-failure-hunter, Suggestion]** `runBrewCleanup`'s "could not build command" branch conflates `brewBinaryPath` failure with `activeConsoleUser` failure — but this ambiguity is pre-existing in `brewCommand()` and shared by every other caller (Scan, Install, Uninstall, EnsureBrewInstalled), not introduced by this PR.

**Tests:** `cd agent && go test -race -count=1 ./internal/patching/...` — green (`ok github.com/breeze-rmm/agent/internal/patching 12.025s`), including the new debounce/coalescing/concurrency/real-cleanup tests. `go vet ./internal/patching/...` — clean.

PR #4916, opened 2026-09-05
fix(mobile,api,web): approval takeover, chat hang, ticket keyboard, upload errors from iOS test

ToddHebebrand:
## Summary

Six defects from Todd's 2026-09-04 iOS screen recording on the OliveTech tenant (frames saved at `~/Documents/breeze-ios-test-frames-2026-09-04/`).

| # | Defect (as observed) | Root cause | Fix |
|---|---|---|---|
| 1 | AI chat stuck on `RUNNING EXECUTE_COMMAND` after approving on the phone; never completes; later "Stopped. Tap to retry." | `ApprovalGate` **replaced** `MainNavigator` with the takeover. `HomeScreen` aborts its SSE stream on unmount, so the `approval_required` event tore down the chat waiting on the decision. Server-side the turn kept running and persisting rows, but nothing on the phone ever re-fetched. | Takeover renders **over** the navigator (stream stays open, like the web card). `HomeScreen` reconciles from `GET /ai/sessions/:id` on remount / background→active / socket error and polls while the transcript is partial (`reconcileHistory` reducer, `turnState.ts`). One quiet retry on the 409 settle race. |
| 2 | Same approval shown twice after a "Breeze" splash; sometimes doesn't dismiss after Face ID | No replay guard on the approval tap listener. A JS relaunch mid-decision re-registers the listener and expo replays the launching tap → `setFocus` on an already-decided request. `PushTapRouter` already guards this for tickets. | Same identifier dedupe, persisted under its own AsyncStorage key. |
| 3 | "Approved · execute_command(…)" toast from the previous request drawn over the next request's buttons | `ApprovalScreen` is one long-lived instance; `approve.fulfilled` rolls focus to the next request in the same tick the toast is queued. | Toasts keyed by approval id; rendered only for the focused request. |
| 4 | Approval headline is the raw signature `manage_services(deviceId=6eae…, action=restart, serviceName=Spooler)` | `intentService` stamped `targetSummary` (audit signature) onto `action_label`, discarding the guardrail's human description passed as `reason`. | `buildActionLabel()` (API): guardrail description with hostname substituted for the device-id stub, softened verb, capped. Used for the approval row, push body and bell. `targetSummary` unchanged. Mobile unpacks legacy signature labels still arriving from pending rows. |
| 5 | Web chat keeps the pending-approval card after the phone decides | `processStreamEvent` never cleared `pendingApproval` on the resumed stream. | `tool_result` for the awaited tool (matched by tool name via its `tool_use` row) clears it. |
| 6 | Cannot dismiss keyboard / scroll to composer or submit on New ticket + Ticket detail | `KeyboardAvoidingView` padding shrinks the viewport but adds no scroll room; 32pt bottom pad; no `keyboardDismissMode`. | `automaticallyAdjustKeyboardInsets` (iOS), `keyboardDismissMode="interactive"`, 64pt pad. KAV disabled on iOS to avoid double-counting. |
| 7 | Camera photo → "Upload failed. Check your connection and try again." with the network fine | **Not root-caused.** Evidence: zero `ticket_attachments` rows in prod for 3 days, no Caddy log hits, no mobile Sentry, API doesn't log requests. So the request never reached the handler: a pre-send failure (image manipulator / file URI) or a 4xx from the auth/route layer, both of which the client collapsed into the generic connectivity message. | Error mapping now surfaces the cause: 4xx without a code → `Upload rejected (HTTP 403: …)`, non-retryable; no-response failures keep the underlying message; `ORG_CONTEXT_REQUIRED` mapped. The next TestFlight run will show the real reason on the chip. |

## Verification

- `apps/mobile`: tsc clean; vitest 97 files / 1249 tests pass (new: `turnState.test.ts`, `reconcileHistory`, `approvalCopy` signature cases, attachment 401/403/404/TypeError cases — all written red-first).
- `apps/api`: tsc clean (needs `--max-old-space-size=12288`); vitest full unit suite 1857 files pass; eslint clean on touched files. New `actionLabel.test.ts`.
- `apps/web`: tsc clean; `src/stores` 230 tests pass; eslint clean. New `processStreamEvent` cases.
- Not verified on device: the overlay takeover, keyboard insets and the reconcile poll need a TestFlight build to confirm.

- Photo upload root cause once the new error text is seen on device.
- Systems header math (`75 devices · 19 offline.` vs `38 online · 19 offline`), ticket status chips lacking "New", org picker without search, `MANAGE_ALERTS` leaking "All" into the chat, streamed text stitched without spaces ("first.I found") — all observed in the same recording.

https://claude.ai/code/session_01E9aYCUwFojJJzPBt1uBELQ

### DIFF
--- a/apps/api/src/services/actionIntents/actionLabel.test.ts
+++ b/apps/api/src/services/actionIntents/actionLabel.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest';
+
+import { buildActionLabel } from './actionLabel';
+
+describe('buildActionLabel', () => {
+  it('prefers the guardrail description and softens its shouted verb', () => {
+    expect(
+      buildActionLabel({
+        toolName: 'manage_services',
+        input: { deviceId: '6eae0f70-8da9-49ff-9e18-c241698975f3', action: 'restart', serviceName: 'Spooler' },
+        reason: 'RESTART service "Spooler" on device 6eae0f70...',
+      }),
+    ).toBe('Restart service "Spooler" on device 6eae0f70...');
+  });
+
+  it('swaps the device-id stub for the hostname when known', () => {
+    expect(
+      buildActionLabel({
+        toolName: 'execute_command',
+        input: { commandType: 'restart_service' },
+        reason: 'Execute "restart_service" command on device 6eae0f70...',
+        deviceHostname: 'KIT',
+      }),
+    ).toBe('Execute "restart_service" command on KIT');
+  });
+
+  it('never returns the raw call signature when the reason is missing', () => {
+    const label = buildActionLabel({
+      toolName: 'manage_services',
+      input: { deviceId: '6eae0f70-8da9-49ff-9e18-c241698975f3', action: 'restart', serviceName: 'Spooler' },
+      reason: null,
+    });
+    expect(label).toBe('Manage services: restart Spooler');
+    expect(label).not.toContain('deviceId=');
+  });
+
+  it('falls back to the tool name alone when nothing recognisable is present', () => {
+    expect(buildActionLabel({ toolName: 'run_script', input: { scriptId: 'abc' } })).toBe('Run script');
+  });
+
+  it('leaves an already-human M365 summary untouched apart from whitespace', () => {
+    expect(
+      buildActionLabel({
+        toolName: 'm365_reset_password',
+        input: {},
+        reason: '  Reset password for  jane@contoso.com  (Contoso Ltd)  ',
+      }),
+    ).toBe('Reset password for jane@contoso.com (Contoso Ltd)');
+  });
+
+  it('caps runaway descriptions', () => {
+    const label = buildActionLabel({ toolName: 'x', input: {}, reason: 'a'.repeat(400) });
+    expect(label.length).toBeLessThanOrEqual(140);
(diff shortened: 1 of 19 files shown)


ToddHebebrand:
**Review round (codex gpt-5.6-sol medium + Sonnet code-reviewer, both on the diff):** 11 + 3 findings, overlapping. Fixed in 9364bc18d:

- CI Type Check red: `runDeferredHumanFanout` and its approver notification lacked the new required `actionLabel` (both reviewers). Now built from tool + arguments on that path.
- `reason` vs headline semantics: agent-run / ticket-triage callers pass a *justification* as `reason`; added a separate optional `actionLabel` input so it never becomes the headline.
- Hostname substitution now matches this call's literal id stub, not an unanchored regex.
- Takeover is a native `Modal` (an already-open Settings/Sessions sheet would have sat above a plain overlay; also fixes TalkBack + Android Back).
- HomeScreen: per-turn generation guard for in-flight catch-up + settle retry; settle timer tracked; completion requires a stable transcript across two polls; reconcile failure fails the streaming row (the error strip is hidden while streaming).
- Web: only the latest `tool_use` of that name resolves the card.
- 408 stays retryable.

Not changed: success/deny confirmation toasts are intentionally dropped once focus rolls to the next request (that overlap *was* the reported bug); outcome errors and expiry are screen-global.

Issue #3257, opened 2026-08-08
Import custom fields when migrating from another RMM


ToddHebebrand:
## Problem

A migration currently loses every user-defined field the incumbent RMM held. That is the largest chunk of genuine customer data an MSP has to abandon when moving to Breeze, and it is data they entered by hand over years:

| Source | Field surface |
|---|---|
| Datto RMM | `udf1`–`udf30` per device (`GET /api/v2/device/{uid}` → `udf`) |
| ConnectWise Automate | EDFs at computer/location/client scope (`extradatavalues` / `extrafield`) |
| Kaseya VSA | Custom fields per agent (`/assetmgmt/assets/customfields`) |
| NinjaOne | Custom fields at global/org/location/device scope |
| N-central | Custom properties at customer/site/device scope |
| Atera | Custom values (`/customvalues/{objectType}/{id}/{fieldName}`) |

Typical contents: warranty expiry, asset tag, purchase date, physical location, primary user, contract reference. Losing it means re-collecting it by hand or doing without.

Breeze has [custom fields](https://github.com/LanternOps/breeze/blob/main/apps/docs/src/content/docs/features/custom-fields.mdx) and a device custom-field-value endpoint, so the destination exists. There is no import path into it.

## Why this is a separate phase from #3242

Custom field values attach to **devices**, and devices only exist in Breeze after enrollment. So this cannot share a transaction, a payload, or a UI flow with the org/site import — it is inherently a second pass run after Phase 4 of a migration (see `apps/docs/src/content/docs/migration/overview.mdx`).

It also needs a join key. The only one available across every source is **hostname**, which is not guaranteed unique across a partner and changes when a machine is renamed. That matching problem deserves its own design rather than being smuggled into the org importer.

## Proposal

1. **Field definition import** — create Breeze custom fields in bulk from an exported definition list, so `udf7` becomes a named, typed field rather than being recreated by hand thirty times.
2. **Value backfill** — accept `{ hostname | deviceId, fieldKey, value }[]`, resolve hostname → device within an organization, and report unmatched rows rather than dropping them.
3. **Preview → commit**, matching the pipeline established in #3242, including per-row match status (`matched` / `ambiguous` / `not found`).
4. **Idempotent** — re-running overwrites the same field on the same device rather than appending.

## Design questions to settle first

- Hostname collisions within an org: reject the row as ambiguous, or require `deviceId`?
- Do we import the incumbent's field *types*, or land everything as text and let the operator retype?
[...]

PR #4917, opened 2026-09-05
feat(api): validate custom-field values on both shipped write paths (#3257 W04)

### LINKED ISSUE #4772: W04: validateCustomFieldValue on both shipped write paths
Wave W04 of #4768 — validateCustomFieldValue on both shipped write paths
Branch: `feature/4768-rmm-custom-field-import/wave-4772`

ToddHebebrand:
Closes #4772

## What changed

Neither shipped custom-field VALUE write path validated a value against its definition's declared `type` before this PR. Values are both an execution sink (`installerVariables.ts` substitutes `{{device.customField.<key>}}` into installer command lines that reach `sendCommandToAgent`) and a targeting control (`filterEngine`'s `custom.<key>` feeds `groupMembership.ts` / `deploymentTargetResolver.ts`). The script write-back path (#4692) already validated; this wave brings the two older, user-facing paths up to the same bar.

1. **Generalised the bounded definition lookup** (`services/customFields/queries.ts`): `loadScriptWritableDefinitions` renamed to `loadVisibleCustomFieldDefinitions`, projection widened to include `id`, `name`, `required`, `orgId`, `partnerId`. `loadScriptWritableDefinitions` is retained as a plain alias so `scriptWriteBack.ts` is untouched (same system-context doc comment preserved verbatim).
2. **New `services/customFields/validateValueMap.ts`**: `validateCustomFieldMap(orgId, updates)` runs a whole field map through the existing `validateCustomFieldValue` (from `#4692`) against the bounded definition lookup, all-or-nothing, and returns either the coerced value map or a rejection list.
3. **`PATCH /devices/:id/custom-fields`** (`routes/devices/customFieldValues.ts`): validates on BOTH the JWT-session and API-key (dualAuth) branches before merging. Stores the coerced value, not the raw input.
4. **`PATCH /devices/:id`** (`routes/devices/core.ts`): same treatment, same shared helper, same error shape, applied to the `customFields` branch only (a PATCH with no `customFields` key never touches the definition lookup).

Error shape on rejection (both routes): `400 { error, code: 'invalid-custom-field-value', fields: [{ fieldKey, reason }] }`, where `reason` is one of `unknown_field | invalid_type | out_of_range | not_a_choice | too_long | invalid_date`.

## ⚠️ Release note — behaviour change on a shipped API

`PATCH /devices/:id` and `PATCH /devices/:id/custom-fields` now **reject** a custom-field value that does not match its definition's declared type, with `code: 'invalid-custom-field-value'` and a per-field reason list. Callers that were previously storing free-form strings under a `number` or `dropdown` field will start getting `400`s. A field with no visible definition at all is also now rejected (`unknown_field`) rather than silently stored. A value for a field whose definition is scoped to specific `deviceTypes` (e.g. a Windows-only field) is now also rejected (`not_applicable_to_device`) when written onto a device of a different OS — previously only the script write-back path enforced this.

Post-review addendum: `routes/partnerApi/devices.ts` reads `assetTag`/`asset_tag`, `inventoryId`, `externalId` out of `custom_fields` for the partner API's `stableIdentifiers`. A tenant that previously wrote one of those keys via PATCH with no matching `custom_field_definitions` row will now get `400 unknown_field` on any future rewrite of that key (existing stored values are untouched — this only affects new writes).

## Deviations from the plan

- The plan's Task 2 → Task 3 split (inline loop in `customFieldValues.ts`, then extract to `validateValueMap.ts` in a follow-up commit) was collapsed: `validateValueMap.ts` was written once and both routes call it from the start. Net behaviour and test coverage are identical to the plan's end state; this just avoids a throwaway inline-then-extract diff.
- `queries.test.ts` and `core.customFieldValidation.test.ts` are new files per the brief; `customFieldValues.test.ts` and its sibling `customFieldValues.mountorder.test.ts` were both updated — the latter wasn't explicitly named in the brief's file list but broke once the route started calling the real `loadVisibleCustomFieldDefinitions` (mocked there now, matching the pattern in the primary test file).
- Pre-existing "free-form" write tests in `customFieldValues.test.ts` (`bitlocker_recovery_key`, `note`) now get a permissive default `text` definition mocked in `beforeEach` so they keep testing merge/audit/isolation behaviour rather than becoming validation tests.

## Not verified

- No migration in this wave (none needed — validation only, no schema change).
- Did not run the RLS/integration suites locally (no DB in this sandbox); this wave touches no tenancy-shaped surface (no new tables, no cascade/export-policy registration), so I don't expect them to be affected, but CI's `integration-test` job will be the first real check.
- `filterEngine.ts` / `installerVariables.ts` (the two consumers this wave is motivated by) are unchanged and untested here — this wave only prevents bad values from being written, it doesn't touch the read side.

## Test plan

```
cd apps/api && npx vitest run src/services/customFields/ src/routes/devices/customFieldValues src/routes/devices/core.customFieldValidation.test.ts
# 10 files, 153 tests passed
npx tsc --noEmit   # 0 errors
pnpm lint          # clean
```

## Review

`/pr-review-toolkit:review-pr` ran three agents in parallel (code-reviewer, pr-test-analyzer, silent-failure-hunter). See the review-summary comment on this PR for the full findings list and disposition — one critical, two important, and several minor findings were fixed in a follow-up commit; the rest were confirmed non-issues or accepted.

### DIFF
--- a/apps/api/src/routes/devices/core.customFieldValidation.test.ts
+++ b/apps/api/src/routes/devices/core.customFieldValidation.test.ts
@@ -1,0 +1,225 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Hono } from 'hono';
+
+// #3257 W04: PATCH /devices/:id validates a `customFields` payload against
+// its definition too, sharing services/customFields/validateValueMap.ts with
+// PATCH /devices/:id/custom-fields (customFieldValues.ts). Isolated from
+// core.permissions.test.ts's larger gate suite so this wave's tests don't
+// collide with unrelated work on that file.
+
+vi.mock('../../db', () => ({
+  runOutsideDbContext: vi.fn((fn: () => unknown) => fn()),
+  withDbAccessContext: vi.fn(async (_ctx: unknown, fn: () => Promise<unknown>) => fn()),
+  withSystemDbAccessContext: vi.fn(async (fn: () => Promise<unknown>) => fn()),
+  db: {
+    select: vi.fn(),
+    execute: vi.fn(),
+    insert: vi.fn(),
+    update: vi.fn(),
+    delete: vi.fn(),
+    transaction: vi.fn(),
+  },
+}));
+
+vi.mock('../../db/schema', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../db/schema')>();
+  return { ...actual };
+});
+
+vi.mock('../../middleware/auth', () => ({
+  authMiddleware: vi.fn((c: any, next: any) => {
+    c.set('auth', {
+      user: { id: 'user-123', email: 'test@example.com', name: 'Test User' },
+      scope: 'organization',
+      orgId: 'org-123',
+      partnerId: null,
+      accessibleOrgIds: ['org-123'],
+      canAccessOrg: (orgId: string) => orgId === 'org-123',
+      orgCondition: () => undefined,
+      token: { mfa: false },
+    });
+    return next();
+  }),
+  requireScope: vi.fn(() => async (_c: any, next: any) => next()),
+  requirePermission: vi.fn((resource: string, action: string) => async (c: any, next: any) => {
+    c.set('permissions', {
+      permissions: [{ resource, action }],
+      partnerId: null,
+      orgId: 'org-123',
+      roleId: 'role-123',
+      scope: 'organization',
+    });
+    return next();
+  }),
+  requireMfa: vi.fn(() => async (_c: any, next: any) => next()),
+}));
+
+vi.mock('../../services/auditEvents', () => ({
+  writeRouteAudit: vi.fn(),
+}));
+
+vi.mock('../../jobs/peripheralJobs', () => ({
+  schedulePeripheralPolicyDevice: vi.fn().mockResolvedValue('job-id'),
+}));
+
+vi.mock('../../services/sentry', () => ({
+  captureException: vi.fn(),
+}));
+
+vi.mock('../../services/remoteAccessPolicy', () => ({
+  resolveRemoteAccessForDevice: vi.fn().mockResolvedValue({ policyId: null, settings: {} }),
+}));
+
+vi.mock('../../services/remoteAccessLauncher', () => ({
+  resolveRemoteAccessLaunch: vi.fn().mockReturnValue({ launchUrl: null, skipReason: 'no_provider_configured' }),
+}));
+
+vi.mock('../agentWs', () => ({
+  sendCommandToAgent: vi.fn(),
+  isAgentConnected: vi.fn().mockReturnValue(false),
+}));
+
+vi.mock('../../services/commandQueue', () => ({
+  CommandTypes: { SELF_UNINSTALL: 'self_uninstall' },
+  queueCommandForExecution: vi.fn(),
+}));
+
+vi.mock('../agents/enrollment', () => ({
+  getGlobalEnrollmentSecret: vi.fn().mockReturnValue(null),
+}));
+
+// The PATCH handler now runs every `customFields` map through
+// validateCustomFieldMap, which calls loadVisibleCustomFieldDefinitions under
+// a SYSTEM db context. Mocked at the module boundary (see
+// customFieldValues.test.ts for the same pattern).
+vi.mock('../../services/customFields/queries', () => ({
+  loadVisibleCustomFieldDefinitions: vi.fn(),
+}));
+
+import { coreRoutes } from './core';
+import { db } from '../../db';
+import { loadVisibleCustomFieldDefinitions } from '../../services/customFields/queries';
+import type { VisibleCustomFieldDefinition } from '../../services/customFields/queries';
+
+const ORG_ID = 'org-123';
+const DEVICE_ID = '11111111-1111-4111-8111-111111111111';
+
+const ACCESSIBLE_DEVICE: Record<string, unknown> = {
+  id: DEVICE_ID,
+  orgId: ORG_ID,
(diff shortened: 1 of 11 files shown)


ToddHebebrand:
## PR Review Summary

Ran `/pr-review-toolkit:review-pr` with 3 agents in parallel: `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter`.

### Critical (1) — fixed
- **code-reviewer**: An empty string (`''`) from the device-edit UI's dropdown `<select>` and `<input type="date"></input>` clear gestures was rejected (`not_a_choice` / `invalid_date`) instead of clearing the field — a real regression this PR introduced, since the old unvalidated write path happily stored `''`. **Fixed** in `validateValue.ts`: `''` is now treated as a clear (same as `null`) for `dropdown` and `date` types, matching the convention the existing `number` editor already used (it normalises its own clear to `null` before submitting). Fixed at the service layer so every caller benefits, not just the web UI. Added regression tests in `validateValue.test.ts` and at the route level in `customFieldValues.test.ts`.

### Important (3) — fixed
- **code-reviewer**: `validateCustomFieldMap` loaded a definition's `deviceTypes` but never enforced it, unlike `scriptWriteBack.ts`'s existing `not_applicable_to_device` gate — an API-key caller could write a Windows-only field's value onto a macOS device through either PATCH route. **Fixed**: both routes now pass `device.osType` into `validateCustomFieldMap`, which applies the same gate. Added unit tests in the new `validateValueMap.test.ts` and route-level tests in both `customFieldValues.test.ts` and `core.customFieldValidation.test.ts`.
- **pr-test-analyzer** + **code-reviewer** (same finding, independently): every test in the original PR used a single-key payload, so the "all-or-nothing" contract `validateValueMap.ts` exists for — one bad key among several good ones rejects the whole map, applying none of it — was never actually exercised. `validateValueMap.ts` also had no sibling test file (every other file in the package does). **Fixed**: added `services/customFields/validateValueMap.test.ts` (mixed valid+invalid map, deviceTypes gate, full-map coercion) plus a mixed-payload regression test in each route's test file.
- **pr-test-analyzer**: a mixed payload where `customFields` fails but a sibling field (e.g. `displayName`) on the same `PATCH /devices/:id` would validate fine was untested — nothing proved the whole PATCH gets rejected, not just the custom-fields portion. **Fixed**: added `'rejects the WHOLE PATCH — including a valid displayName — when customFields fails'` to `core.customFieldValidation.test.ts`.

### Minor (5) — fixed
- **code-reviewer**: dead code — `ScriptWritableDefinition` interface in `queries.ts` was referenced nowhere after the widening. **Removed.**
- **code-reviewer**: two stale strings from before the loader was generalised to three callers — the `withSystemDbAccessContext` trace label still said `'customFields.scriptWriteBack.definitions'`, and the doc comment still attributed the SYSTEM-context rationale solely to `runWithAgentOrgDbAccess`. **Fixed** both.
- **code-reviewer**: `customFieldValues.ts`'s file header still framed a bare `bitlocker_recovery_key` PATCH as accepted, which is no longer true without a matching definition. **Added** a note to the header.
- **pr-test-analyzer**: the dropdown-rejection test in `customFieldValues.test.ts` didn't assert `updateSpy.set` was never called, unlike its siblings. **Fixed.**
- **pr-test-analyzer**: `out_of_range`/`too_long` reasons were untested at the route layer (only unit-tested in pre-existing `validateValue.test.ts`). **Added** an out-of-range route test (a text-too-long route test collides with the transport schema's own 10,000-char cap before reaching this validator, so that one stays unit-level only — noted for anyone reading the diff).

### Minor — dismissed / accepted as-is
- **code-reviewer**: pool-pressure note (the SYSTEM-context definition lookup now opens a second pooled connection on every custom-field PATCH, not just script write-back). No code change — this repo's existing pattern already accepts this cost for the script path, and the request volume here (interactive PATCH, not fleet automation) doesn't currently warrant caching. Added a comment on `loadVisibleCustomFieldDefinitions` flagging it as a future optimization if it becomes measurable.
- **code-reviewer**: release-note gap — `routes/partnerApi/devices.ts` reads `assetTag`/`inventoryId`/`externalId` out of `custom_fields`, so a tenant that previously wrote one of those keys with no matching definition will now get `unknown_field` on a future rewrite. Not a code fix — **added to the PR body's release note** instead, per the finding's own suggestion.
- **code-reviewer**: alias fragility — `loadScriptWritableDefinitions` is a `const` re-export rather than a wrapper function, so a test that mocks only `loadVisibleCustomFieldDefinitions` leaves the alias undefined in that module graph. Reviewer's own assessment: "harmless today... will bite whoever adds a device-route import that reaches scriptWriteBack" — accepted as a known, low-probability future trap rather than a current defect.
- **pr-test-analyzer**: the reference-identity test (`loadScriptWritableDefinitions === loadVisibleCustomFieldDefinitions`) is near-tautological given the one-line alias. Reviewer rated it low-value but non-zero (documents intent, would catch a future refactor into a wrapper) — left in place.

### Clean
- **silent-failure-hunter**: no findings across all five focus areas (DB-error swallowing, partial-mutation-before-check, the `core.ts` reassignment's blast radius, error-detail leakage in the 400 body, and the audit-log gate). Verified in-session, not inferred.

### Verification after fixes
```
cd apps/api && npx vitest run src/services/customFields/ src/routes/devices/customFieldValues src/routes/devices/customFieldValues.mountorder.test.ts src/routes/devices/core.customFieldValidation.test.ts src/routes/devices/core.permissions.test.ts
# 10 files, 153 tests passed
node --stack-size=8000 --max-old-space-size=8192 node_modules/typescript/lib/tsc.js --noEmit   # 0 errors
pnpm lint   # clean
```

Issue #4923, opened 2026-09-05
[Agent][Backup] Windows profile selections run files, SQL, and Hyper-V concurrently without a per-device execution queue


ToddHebebrand:
## Report

A reporter observed that a Windows server with Hyper-V, files, and SQL selected for backup attempted to run all of them at once. The exact agent/server versions and endpoint logs were not supplied.

Source inspection supports this behavior: profile selections are dispatched independently and the backup helper has no device-wide execution gate across backup modes. This is an investigation based on source, not a live Windows reproduction.

## Confirmed execution path

Inspected commit: `78eae279fb1ede458fe2c740755fe0b26dfb254a`.

- **Server explicitly permits cross-mode overlap.** `apps/api/src/services/backupJobCreation.ts:53-73` defines idle as device + mode and comments that profile jobs run “side by side.” The advisory lock and active-job lookup are mode-scoped. `backupJobCreation.test.ts:289-308` asserts distinct locks so one device can fan out concurrently.
- **Selections are enqueued without completion dependencies.** `apps/api/src/jobs/backupWorker.ts:258-283` creates/enqueues each scheduled selection; manual paths do the same in `apps/api/src/routes/backup/jobs.ts:251-291` and `:482-512`. Worker concurrency is 5 (`backupWorker.ts:124`), with no per-device execution serialization here.
- **Awaiting dispatch does not await backup completion.** `backupWorker.ts:890-894` sends prepared targets in a loop, but `apps/api/src/services/agentCommandRelay.ts:194-202` documents that a successful dispatch means the frame reached `ws.send()`, not that the operation completed. Thus even sequential sends can start overlapping work.
- **Helper starts every command concurrently.** `agent/cmd/breeze-backup/main.go:467-472` launches a goroutine for every `TypeBackupCommand`. `activeCommands` prevents idle shutdown; it does not limit execution. For async file backups, `:502-508` acknowledges `started` before running the backup.
- **The existing running guard does not serialize these workloads.** `agent/internal/backup/backup.go:300-307` protects one `BackupManager` instance. Server file commands construct fresh payload managers (`agent/cmd/breeze-backup/main.go:553-560`, `exec_backup.go:96-171`), so this is not a device-wide guard. SQL and Hyper-V dispatch directly to separate handlers (`main.go:685-696`) and do not call `RunBackupContext` or take its running guard.
[...]

PR #4925, opened 2026-09-05
fix(backup): serialize device backup workloads (#4923)

ToddHebebrand:
## Summary

- Queue file/system-image, SQL, and Hyper-V backup workloads through one helper execution slot per device, preserving selections from different profiles while other devices run independently.
- Keep admitted selections pending with queue liveness; start execution timestamps when work begins. Target cancellation by job ID and reconcile stale queued work, duplicate delivery, and helper disconnects.
- Add native execution heartbeats and retain compatibility with older helpers that return synchronous results.

## Linked Issues

Refs #4923. This implements execution serialization and server queue lifecycle coordination. The reporter still needs to verify manual/scheduled mixed-mode backups and simultaneous profiles on a Windows server with working SQL and Hyper-V; no live Windows reproduction was available here.

The complete queued-status/timeout behavior requires the updated API, agent, and backup helper. Existing SQL/Hyper-V native operations do not support immediate cancellation: the slot remains occupied until the native call returns, preventing the next workload from overlapping it.

## Type of Change

- [x] Bug fix

## Testing

- [x] Affected existing tests pass
- [x] Added regression tests
- [ ] Live Windows manual testing

- API: 287 tests passed across 11 files, including four lifecycle tests executing production functions against an isolated PostgreSQL instance. PostgreSQL tests opt in via `BACKUP_QUEUE_TEST_DATABASE_URL`; the owned test container was removed afterward.
- API: `tsc --noEmit` passed with `NODE_OPTIONS=--max-old-space-size=8192` (the initial default-heap run exhausted memory).
- Agent: `go test -race ./cmd/breeze-backup ./internal/backup ./internal/sessionbroker ./internal/heartbeat` passed.
- Windows amd64 compilation passed for those four Go packages; this was compile-only validation.
- Repository commit/push hooks and `git diff --check` passed.

## Checklist

- [ ] I have tested on the relevant platforms

## Review fixes (2026-09-05)

Findings from the multi-lens review (code, silent-failure, tests, types, comments, Codex) applied inline:

- **Helper only queues `QueueAsync` dispatches.** A server without `backup_queue_async` sends SQL/Hyper-V synchronously with a 30-minute round trip; parking those behind another workload timed the caller out and then ran the command anyway, orphaned. Synchronous workloads now execute directly (pre-queue behaviour), proven by a `commandLoop` test.
- **Server relaxation gated on helper capability.** `devices.backup_version >= 0.110.0` (`backupHelperSupportsQueue`) decides whether `createManualBackupJobIfIdle` uses the per-selection dedupe or keeps the pre-queue one-active-job-per-device+mode guard. Older helpers run every dispatched job concurrently, so relaxing for them would recreate the reported overlap. Unknown/unparseable versions count as not capable.
- **Targeted `backup_stop {jobId}` joins the unwind** (bounded 20s, under the 30s forward timeout) and reports `drained:false` when a native export is still holding the slot, so the row is not marked terminal while VSS teardown or an upload is still writing.
- **Reaper pending branch** joins `devices`: skips offline devices (as the running branch already did), sends the targeted stop whenever the helper is queue-capable even if the admission ack was lost on the WS, and logs when it must skip an older helper (whose device-wide stop would kill the running backup).
- **Real-DB lifecycle test now runs in CI.** `backupProgress.postgres.test.ts` was gated on an env var nothing set and used its own temp schema. Replaced by `backupProgress.integration.test.ts` against the migrated schema, registered in `vitest.integration.config.ts`; it fails 3/4 against the pre-PR `applyBackupProgress`.
- `activeCommandCanceller` gains `rebind`/`cancel`/`waitDone`; the handler no longer writes its `contexts` map directly, and `track`'s re-entrancy contract is documented. `ticket.release()` is deferred before the payload parse. Duplicate admissions are re-acked with a logged send error. The queued-workload list lives in `backupipc.IsQueuedWorkload`, shared by agent and helper.
- errcheck lint fix (was failing Lint Agent); concurrent high-water-mark test for the execution guard; test title/comment nits in `agentWs.test.ts`.

Verified locally: `go test -race` on the five touched packages, Windows amd64 compile, API `tsc --noEmit`, eslint, unit suites for the touched files, and both integration suites against a private test stack (torn down afterwards).

**Re-review of the fix delta (Opus + Codex):** the agent's `backup_stop` forward timeout was 10s, not 30s, so the 20s drain would have made every drained stop read as failed. Both timeouts now live in `backupipc` (forward 30s, drain 20s) with a test pinning drain < forward. Also: the integration test is excluded from the unit runner, scheduled dedupe carries the same helper gate as manual, the `devices.backup_version` lookup is hoisted to once per device, the guard comment no longer claims coverage of synchronous mssql/hyperv, and the drain test ordering is deterministic.

Not changed: `test-agent-race` is still not in `ci-success` `needs:`; that is a CI-config decision outside this PR.


ToddHebebrand:
**Review run:** `review-pr` code, error-handling, and regression-test passes, with independent API and helper reviewers. Reviewed implementation `a297589801` and final test-only commit `3da94ebf8f` against merge base `78eae279fb`; confirmed the open PR uses the reviewed head and diff.

**Findings:** Five defects raised and addressed in `a297589801`: legacy progress promotion, cancellation of reaped queued jobs, native execution liveness, synchronous older-helper response tracking, and stop-all ordering. The configured-manager test coverage gap was addressed and re-reviewed in `3da94ebf8f`. No outstanding actionable review findings.

**Tests:** 287 API tests passed across 11 files, including four real PostgreSQL lifecycle tests. API TypeScript check passed with an 8 GiB heap. Affected helper/backup/sessionbroker/heartbeat packages passed `go test -race`; Windows amd64 compile-only checks passed. The final configured-manager stop regression passed with `-race -count=3`. Reviewers inspected the tests; executed test results are from the implementation verification runs.

**Status:** Review complete; GitHub CI is running. Live Windows mixed-mode/profile verification remains outstanding. Existing native SQL/Hyper-V calls finish before cancellation releases the slot; complete queue-status behavior requires updated API, agent, and helper. PR remains open and unmerged.

PR #4927, opened 2026-09-05
fix(extensions): restore boot tenancy sweep (#4283)

### LINKED ISSUE #4283: [API][Tenancy] assertNoUnaccountedPublicTables has no production caller — a documented boot-time tenancy gate that never runs
Found during the #3470 work (PR #4277), deliberately left unfixed there to avoid widening the diff.

`assertNoUnaccountedPublicTables` (`apps/api/src/extensions/tenancyTripwire.ts:377`) is documented in two places as an active boot-time gate, but **it has no production caller.** Verified on `origin/main` — every reference is either a doc comment or a test:

```
apps/api/src/extensions/tenancyTripwire.ts:377        the definition
apps/api/src/extensions/tenancyTripwire.ts:196        doc comment describing it as closing the unprefixed case
apps/api/src/extensions/builtinExtensions.ts:494      doc comment: "assertNoUnaccountedPublicTables sweep would otherwise read those"
apps/api/src/__tests__/integration/extensionTenancyRls.integration.test.ts   the only executing caller
```

A plan doc (`docs/superpowers/plans/misc/2026-07-18-runtime-extension-platform-05-packer.md:469`) states it plainly as live behaviour: *"The `'tenancy'` reconcile phase (`reconciler.ts:636-637`) runs `assertNoUnaccountedPublicTables`"*. Whatever wired it there is gone; `reconciler.ts` no longer calls it.

### Why this matters

The sweep's job is to catch an extension migration creating an **unprefixed** `public` table that no manifest declares — i.e. a table that would sit outside the tenancy declaration and therefore outside org-cascade and GDPR tenant-export. That is the same failure class the repo has shipped or blocked CI on five times (#1359, #1351, #1365, #2179, #2514), and the one the project's own guidance says code review catches 0/5 times while contract tests catch 5/5.

So the control exists, is tested, and is not running. Anyone reading `tenancyTripwire.ts:196` or `builtinExtensions.ts:494` will reasonably believe it is.

### Suggested fix

Decide, and make the code and the comments agree either way:

1. **Re-wire it** into the extension load/reconcile path (the plan doc says the `'tenancy'` reconcile phase was the intended home), or
2. **Retire it** — delete the function and its integration test, and correct both doc comments plus the plan doc.

Option 1 is the safer default given the failure class, but it needs someone to confirm it does not fire on tables that are legitimately unprefixed today. Do not simply delete the comments and leave the function orphaned — that trades a misleading comment for an invisible one.

Related: PR #4277 makes the disabled-builtin path fall back to a `breeze_migrations` ledger probe precisely because the manifest read it replaced was load-bearing for the same tenancy declaration. This sweep is the other half of that guard.

🤖 Filed by Claude Code from the #3470 work; the absence of any production caller verified against `origin/main`.

ToddHebebrand:
## Summary

- Restore the final public-table tenancy sweep in `loadBuiltinExtensions`, the current sole extension boot path. API and worker startup now fail if an enabled extension leaves an undeclared, unprefixed public table behind.
- Sweep once all enabled and persisted disabled declarations are available. Disabled partial schemas retain only existing tables, while shared core tables alone do not trigger a sweep on stock installations.
- Document the active gate and cover its real production wiring against PostgreSQL, including core table inventory compatibility.

## Linked Issues

Closes #4283

## Type of Change

- [x] Bug fix

## Testing

- [x] Existing affected tests pass: all extension suites plus worker boot/entrypoint contracts — 369 tests across 18 suites.
- [x] Added regression tests: API/worker failure propagation, disabled partial schemas, shared-only stock installations, and the actual production loader/catalog connection.
- [x] Real PostgreSQL integration suite — 19 tests passed on a private worktree test stack. Production catalog checks use `breeze_app`; privileged connections only create fixtures and apply migrations. Live core inventory has no uncovered tables. Test stack removed afterward.
- [x] Targeted ESLint and API TypeScript check.

Commands: `pnpm --filter @breeze/api exec vitest run src/extensions/ src/worker.boot.test.ts src/services/workerEntrypointClosure.contract.test.ts --maxWorkers=1 --no-file-parallelism`; `pnpm --filter @breeze/api exec vitest run --config vitest.integration.config.ts src/__tests__/integration/extensionTenancyRls.integration.test.ts --maxWorkers=1 --no-file-parallelism`; `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit --project apps/api/tsconfig.json`.

## Checklist

- [x] I have tested on the relevant platforms


ToddHebebrand:
**Review run:** `$review-pr` — code, errors, tests, and comments at `a32ec0eab165f42d22c30aa71d99cfe1aac0f590`. Independently reviewed; the published diff exactly matches the reviewed diff.

**Findings:** No outstanding findings. Checked production-loader wiring, API/worker failure propagation, declaration ordering, disabled partial schemas, and stock/shared-only behavior.

**Tests:** 369 tests across the extension and worker-boot suites; 19 real PostgreSQL integration tests, including production-loader rejection of an unrelated undeclared table. API TypeScript and targeted ESLint passed. The isolated test stack was removed.

**Status:** Review complete. CI is running; awaiting maintainer merge.

PR #4928, opened 2026-09-05
fix(workspace): unwrap SQLSTATE errors and guard direct comparisons (#4245)

### LINKED ISSUE #4245: [ee/workspace] Two more top-level err.code SQLSTATE checks are dead under Drizzle — plus a lint rule to end the bug class

ToddHebebrand:
## Summary

- Workspace source writes now return the intended 400 for Drizzle-wrapped foreign-key violations. Ingest-job creation recognizes wrapped unique violations and rolls back an insert savepoint before reading or upgrading the winning job, preserving the host's outer org transaction.
- Share the existing cause-chain detectors through `@breeze/shared/pgErrors`, preserving API imports through a re-export and keeping extensions independent of API source. Constraint matching and depth limits retain their existing behavior.
- Register `breeze/no-direct-sqlstate` in existing ESLint configs and run its regression cases from `pnpm lint`. Narrow documented exceptions preserve existing cause-aware API mappers; optional/computed/reversed comparisons are guarded and OS codes remain allowed.

## Linked Issues

Closes #4245

## Type of Change

- [x] Bug fix
- [x] CI / build / tooling

## Testing

- [x] Existing affected tests pass
- [x] Added new tests
- `pnpm --filter @breeze/ext-workspace exec vitest run --maxWorkers=1 --no-file-parallelism`: 616 tests passed.
- `pnpm --filter @breeze/ext-workspace exec vitest run --config vitest.integration.config.ts src/__tests__/ingestJobs.integration.test.ts --maxWorkers=1 --no-file-parallelism`: 10 passed on a private per-worktree Postgres stack as `breeze_app` (superuser=false, bypassrls=false), including a real wrapped 23505 and usable outer transaction after recovery.
- Shared helper tests: 15 passed. API helper/re-export tests: 23 passed.
- `pnpm lint`: all six configured package tasks passed; SQLSTATE RuleTester cases passed. Touched shared files also pass their ESLint config.
- Workspace and API type checks passed using an 8GB Node heap.
- Shared package typecheck remains blocked by three existing `src/validators/quotes.test.ts` errors at lines 286, 308, and 311; reproduced unchanged in a pristine worktree at base `9135d14d08`. No new type errors were reported.

## Checklist

- [x] I have tested on the relevant platforms


ToddHebebrand:
**Review run:** `$review-pr` — code, errors, tests, and packaging at `edb23faabf1f0c3e68201229552dad00699700a5`. Independently reviewed the published commit, including the final shared barrel and SQLSTATE class coverage.

**Findings:** No outstanding findings. Checked wrapped-error handling, savepoint recovery inside the host's org transaction, unchanged API helper semantics, shared export/alias resolution, and lint registration with narrow documented exceptions. Final lint-rule tests independently rerun successfully.

**Tests:** 616 Workspace tests, 10 real PostgreSQL tests as non-superuser/non-BYPASSRLS `breeze_app`, 15 shared helper tests, and 23 API helper tests passed. Root lint plus API/Workspace type checks passed. Shared package typecheck still reports three untouched quote-test errors, reproduced identically at base `9135d14d08`; this is a pre-existing validation limitation.

**Status:** Review complete. CI is running; awaiting maintainer merge. The shared typecheck limitation remains as documented above.

PR #4939, opened 2026-09-05
fix(ci): compile iOS simulator for mobile changes (#4489)

### LINKED ISSUE #4489: CI has no iOS native build — Dependabot can silently break the mobile Archive
## What

Nothing in `.github/workflows/` compiles the mobile app's native code. The `test-mobile` job runs Vitest against JS only, so a change that breaks the iOS build passes every required check and is only discovered by a human running **Product → Archive**.

## How it bit us

#4488 fixes a broken iOS build on `main`. Dependabot moved React Native to 0.87.x (#4305, with #4303 / #4304 / #4308 alongside), past what Expo SDK 57 is tested against (0.86.3). RN 0.87 removed the old Paper headers `React/RCTTextView.h` and `React/RCTRootContentView.h`, which `@sentry/react-native@7.11.0` and `react-native-gesture-handler@2.32.0` still import. Every one of those PRs was green.

The break sat on `main` undetected until someone tried to cut a TestFlight build for the mobile ticketing work (#3206).

## Suggested fixes

Either or both:

1. **Add a compile-only iOS job** — `expo prebuild --platform ios` + `pod install` + `xcodebuild -sdk iphonesimulator CODE_SIGNING_ALLOWED=NO build`. No signing or secrets needed, which is what makes it cheap. It needs a macOS runner, so it may be worth restricting to PRs touching `apps/mobile/**`.
2. **Constrain the Dependabot mobile group to Expo's SDK cadence** — `react-native`, `react`, and the `expo-*` / `react-native-*` packages should move when the Expo SDK moves, not independently. `npx expo install --check` is the authority on the correct set and could run in CI as a fast standalone guard even without a macOS runner.

Option 2 alone would have prevented this specific break and costs no macOS minutes; option 1 catches the general class.

https://claude.ai/code/session_017eEp7QaCXwJNioNR6vZ6eP

ToddHebebrand:
A React Native dependency update can pass the mobile JavaScript suite while breaking native compilation. CI now builds the generated iOS app for the simulator with signing and Sentry uploads disabled whenever mobile code, workspace dependencies, package-manager/runtime configuration, or this workflow changes.

A Linux path check avoids allocating macOS runners for unrelated changes. CI Success requires successful native compilation for matching changes and fails closed if path detection fails or compilation unexpectedly skips. Manual CI dispatch also runs the native build. This is Debug simulator compilation; release Archive configuration and signing remain validated by the existing release build guards.

Validation:
- 12 workflow contract tests passed, including execution of the actual CI Success shell for passing, failed, cancelled, and skipped native jobs.
- 113 workflow/security tests passed; workflow security scan passed.
- `actionlint -shellcheck= .github/workflows/ci.yml` passed. Full actionlint reports one existing SC2155 warning in the unrelated build-agent script.
- Mobile TypeScript check passed; 97 test files passed (1,250 tests passed, 3 skipped).
- Local Expo prebuild and CocoaPods installation passed. Xcode 26.6 unsigned Debug simulator compilation passed for arm64 and x86_64 (`BUILD SUCCEEDED`, exit 0), using the same xcodebuild arguments as the job. GitHub macOS validation remains pending publication.

Closes #4489


ToddHebebrand:
Review summary for e2bd28b63b4f9f4a9d5b7730ffb1df4f02552221:

No actionable findings remain after parent review of the implementation worker's final workflow. Reviewed changed-path selection, unsigned simulator compilation, job permissions, and CI Success handling of failed, cancelled, missing-output, and unexpectedly skipped jobs. The parent independently executed all 12 workflow contract tests, which passed.

Implementation validation: 1,250 mobile tests passed (3 skipped), 113 workflow/security tests passed, mobile type checking passed, and unsigned Debug simulator compilation succeeded for arm64 and x86_64 with Xcode 26.6. Actionlint passed with shellcheck disabled; the full run's SC2155 warning was reproduced on the unchanged base workflow.

GitHub-hosted macOS validation remains pending. Release Archive/signing validation continues through the existing release guards; this PR adds the unsigned simulator compilation check.

PR #4967, opened 2026-09-05
feat(api): close every cancelling execution — the five closers (#3525 W03)

### LINKED ISSUE #4764: W03: The five closers: result-handler CAS, cancellation sweep, cancel-command expiry, late-result recovery, audit

ToddHebebrand:
Closes #4764

Wave W03 of #4761 (issue #3525) — **the five closers**. Depends on W02 (#4799), merged.

`cancelling` must never be reachable-and-permanent. Before this wave nothing could leave that state: the result handler's three branches all missed it, and no reaper owned it.

### `services/commandResultHandlers.ts` — closers 1 and 3

**Closer 1 — the original script's own result closing a `cancelling` execution.** A new compare-and-swap runs *before* the primary CAS and the #3607 recovery branch. The primary matches `pending|queued|running` and #3607 matches swept `timeout|failed` rows with no output, so a `cancelling` row matched neither and the agent's real result fell through to the "matched nothing" path and was discarded. Both existing branches are now guarded by `if (cancelClosed.length === 0)`.

It is **two writes, not one**, because the honesty contract splits here:

| Result while `cancelling` | `status` | `cancel_state` |
|---|---|---|
| `cancelled: true` **and** `cancelledByCommandId` matches the row's `cancel_command_id` | `cancelled` | `confirmed` |
| `cancelled: true` naming a **different** command id | the real outcome | `unconfirmed` |
| `cancelled: true` with **no** command id (pre-#3525 agent's answer to a non-blocking signal — a request, not a receipt) | the real outcome | `unconfirmed` |
| unmarked (OD9-C loser) | the real outcome | `unconfirmed` |

Pinning `cancel_command_id` is what stops a stale or retried cancel being credited with a kill it did not do.

**Closer 3 — late original result after a terminal cancel.** The block at the bottom used to drop the output, with a comment asserting that was correct "because the operator asked for the run to be abandoned". That is the design bug written down: an operator who stops a script wants to see how far it got. It now fills `stdout`/`stderr`/`exit_code` and nothing else — no status change, no `cancel_state` change, no batch accounting, no automation-action closure, no `captureException` (the race is expected, not a defect). Gated on `exit_code IS NULL` for idempotency, and deliberately **not** gated on `cancel_state`: a confirmed cancel has partial output worth keeping too.

`applyScriptCustomFieldWrites` still runs ahead of every execution-status guard — the script really did run and really did produce those values. Now asserted deliberately rather than inherited by accident.

### `jobs/staleCommandReaper.ts` — closers 4 and 5

New `reapStaleCancellations`, registered as `['scriptCancellations', reapStaleCancellations]`. It owns `cancelling` exclusively; `reapStaleScriptExecutions` is untouched and stays blind to it, because the two answer different questions ("did the script finish in time" vs "did the cancel REQUEST resolve") and widening the first would make a failed cancel look like a failed script.

Nothing here terminalises as `cancelled` — only proof does. The sweep reverts `status` to `cancel_prev_status` (floor `running`, which keeps the row inside the script reaper's predicate rather than stranding it) and records `cancel_state = 'unconfirmed'`.

Three resolution arms:
- **closer 4** — the cancel command reached a terminal status (its own 2-hour `LONG_TIMEOUT` tier expired it, or an ack landed that resolved nothing);
- **closer 5** — it was **delivered** more than `CANCEL_GRACE_MS` ago. The clock starts at delivery (`device_commands.executed_at`), never at the request: losing the WS connection does not kill the agent or its script, so a cancel queued against an offline device is still deliverable on reconnect;
- **orphan** — the command row is gone (see deviations).

A still-open cancel command is closed with exactly `status: 'failed'` + `result: { status: 'timeout' }` — the only marker `commandResultAcceptance.ts` reopens — so a late ack from a slow device still lands. Audit event `script.execution.cancel.unconfirmed` with `{ deviceId, cancelCommandId, revertedTo }`, plus `recordCancelUnconfirmed()`. **No device alert and no `captureException`**: an unacknowledged cancel is an operational condition (agent offline, script already gone, agent too old to know the command), not a code defect.

1. **The cancellation CAS is two writes, not one.** The plan's Task 3.1 snippet builds one `UPDATE` whose SET is conditional and whose WHERE adds `eq(cancelCommandId, markerCommandId)` when a marker is present. That contradicts the plan's own test `'a marker naming a DIFFERENT cancel command does not confirm'`, which expects `cancelState: 'unconfirmed'` — with a single write the mismatched-marker case matches **nothing**, falls through, and is dropped. Split into a confirming CAS (marker-pinned) followed by an unconfirmed CAS when it matched nothing. Both plan expectations now hold, and `cancelled: true` with no command id lands on `unconfirmed` for free, consistent with W02's `resolveCancelAckOutcome`.

2. **Third resolution arm for a vanished cancel command.** `cancel_command_id` is a bare uuid *precisely because* command rows are reaped independently (its column comment says so). Once the row is gone, neither `cmdTerminal` nor `cmdExecutedAt` can ever be true, so the plan's two arms leave the execution in `cancelling` forever — the one state this wave exists to make unreachable. The orphan arm resolves it, bounded by `cancel_requested_at` older than `CANCEL_GRACE_MS` so the write window between stamping an execution `cancelling` and its command row becoming visible cannot revert a cancel still being dispatched. No command close is attempted (the join already proved there is nothing to close).

3. **`recordCancelUnconfirmed()` had to be created** (`services/scriptCancellationMetrics.ts`, new). It registers `breeze_script_cancel_unconfirmed_total` directly on `metricsRegistry` rather than going through the settable-recorder indirection `backupMetrics.ts` uses. That indirection exists for leaves that must not import `prom-client`; this module has no such constraint, and binding from `routes/metrics.ts` would reproduce #4143 exactly — the reaper runs in the **worker** role, which never loads that file, so the series would have been a permanent zero published by the wrong process. The module is a leaf (`prom-client` + `./metricsRegistry` only), so `workerEntrypointClosure.contract.test.ts` still passes.

4. **`writeSystemAudit` does not exist**; used `createAuditLogAsync` with `actorType: 'system'` / `ANONYMOUS_ACTOR_ID` / `initiatedBy: 'schedule'`, matching `jobs/offlineDetector.ts`.

5. **`REAPER_DOMAINS` moved to module scope and is exported.** The plan's test asserts `reaperDomainNames()` contains `scriptCancellations`, but the array lived inside `createWorker()` and was unreachable. A reaper that exists but is never registered reaps nothing, and no test of the function itself can see that.

6. **Not in the plan: a proven cancel now closes its automation action as `'cancelled'`.** Feeding `effectiveExecution` from the cancellation CAS (which the plan asks for) makes `applyAutomationActionTerminal` run on this path for the first time. The agent reports a killed process as `failed` with exit -1, so `scriptStatus` would have made an automation step read "failed" when the operator stopped it — the same dishonesty the execution row now avoids. `applyAutomationActionTerminal` already accepts `'cancelled'`. An unproven cancel keeps the real outcome.

7. **Batch counters are deliberately not bumped from the cancellation CAS** (they key on `updatedExecutions`, the primary CAS's rows only) — matching the plan and the existing #3607 precedent.

## Testing

New: `commandResultHandlers.cancellation.test.ts` (16), `staleCommandReaper.cancellation.test.ts` (17).

Both harnesses are **evaluating, not scripted**. The result-handler fake `db` interprets the real `where()` predicate against a simulated row rather than returning a canned list per call, so "which CAS ran first" is observed rather than assumed and a predicate that silently matched everything would fail. The reaper suite records every `db.update()` with the **table** it targeted, so an assertion about the execution row cannot be satisfied by the write to the cancel command; and "a `cancelling` row is invisible to `reapStaleScriptExecutions`" is asserted against the **compiled** predicate via `PgDialect().sqlToQuery(...)`, not a mocked result set.

- `npx vitest run src/services/commandResultHandlers` — 5 files, 39 tests, pass
- `npx vitest run src/jobs/staleCommandReaper` — 2 files, 57 tests, pass
- `npx vitest run src/services/workerEntrypointClosure src/services/scriptCancellation src/db/schema/scripts.test.ts src/routes/agents/commands.test.ts src/routes/agentWs.enqueueContract.test.ts src/services/commandTimeouts.test.ts src/services/commandResultAcceptance.test.ts src/services/workerRegistry.test.ts src/jobs/softwareDeploymentScheduler.test.ts` — pass
- `npx tsc --noEmit` — clean (needs `--max-old-space-size=8192`)
- `pnpm lint` — clean

Red-first confirmed on both files: 10/14 then 17/17 failing before implementation.

## Not verified

- **No integration/RLS run.** No schema, migration, tenancy, cascade or export-policy surface is touched, so no contract list applies; the three integration suites that import these modules (`staleBackupReaper`, `deviceUninstallDrain`, `automationTerminalReconciliation`) run in CI's Integration Tests job. Not run locally.
- **No end-to-end cancel yet.** Nothing in the repo writes `status = 'cancelling'` — the request-side route lands in a later wave — so every path here is exercised against a simulated row, not a real one. The first real proof is W06's E2E.
- **The `cancelled` automation-action status is not asserted end-to-end**, only at this call site.
- **No agent involved** (W04, parallel). Wire-contract behaviour against a real agent is untested here.
- `breeze_script_cancel_unconfirmed_total` is not asserted to appear in a `/metrics` scrape.

https://claude.ai/code/session_0158dEZKsjdM33itE5rajN8S

### DIFF
--- a/apps/api/src/__tests__/integration/lateCommandResultRecovery.integration.test.ts
+++ b/apps/api/src/__tests__/integration/lateCommandResultRecovery.integration.test.ts
@@ -120,6 +120,7 @@ async function seedTimedOutRun(
   fx: Fixture,
   opts: {
     executionStatus?: 'running' | 'timeout' | 'failed' | 'completed' | 'cancelled';
+    executionCancelState?: 'requested' | 'confirmed' | 'unconfirmed' | 'failed';
     executionExitCode?: number;
     executionStdout?: string;
     commandStatus?: 'failed' | 'cancelled';
@@ -140,6 +141,11 @@ async function seedTimedOutRun(
       startedAt: new Date(Date.now() - 90_000),
       exitCode: opts.executionExitCode ?? null,
       stdout: opts.executionStdout ?? null,
+      // #3525: `cancel_state` is orthogonal to `status` and carries the CANCEL
+      // REQUEST's outcome. The CHECK constraint ties it to cancel_requested_at.
+      ...(opts.executionCancelState
+        ? { cancelState: opts.executionCancelState, cancelRequestedAt: new Date(Date.now() - 60_000) }
+        : {}),
     })
     .returning({ id: scriptExecutions.id });
   if (!execution) throw new Error('seedTimedOutRun: no execution');
@@ -327,6 +333,7 @@ async function readExecution(executionId: string) {
       exitCode: scriptExecutions.exitCode,
       stdout: scriptExecutions.stdout,
       stderr: scriptExecutions.stderr,
+      cancelState: scriptExecutions.cancelState,
     })
     .from(scriptExecutions)
     .where(eq(scriptExecutions.id, executionId))
@@ -589,17 +596,20 @@ describe('#3607 late command result recovery', () => {
   );
 
   runDb(
-    'a CANCELLED execution is left alone — the operator abandoned the run',
+    'a CANCELLED execution KEEPS the late output — it is what the operator stopped the run to see',
     async () => {
       const fx = await makeFixture();
-      // routes/scripts.ts's cancel handler stamps the execution 'cancelled'
-      // but only cancels the paired command `WHERE status = 'pending'`. A
-      // command already 'sent' survives, later picks up the reaper's
-      // provisional timeout marker, and its real result now reaches
-      // handleScriptResult. Discarding the output is CORRECT here — and this
-      // is a routine race, so it must not be treated as a defect.
+      // CONTRACT CHANGE (#3525 closer 3). This test previously asserted the
+      // opposite — that the output was discarded — on the reasoning that "the
+      // operator asked for the run to be abandoned". The #3525 plan (W03, Task
+      // 3.1 Step 4, "Replace the 'drop the output' path with output recovery")
+      // identifies that as the design bug written down: an operator who stops a
+      // script wants to see how far it got, and the partial stdout is the only
+      // record of that. The race itself is unchanged and still routine — only
+      // what we do with the output changed.
       const { commandId, executionId } = await seedTimedOutRun(fx, {
         executionStatus: 'cancelled',
+        executionCancelState: 'confirmed',
       });
 
       await sendWsResult(fx, commandId, {
@@ -609,9 +619,56 @@ describe('#3607 late command result recovery', () => {
       });
 
       const execution = await readExecution(executionId);
+      // The output lands...
+      expect(execution.stdout).toBe('output for an abandoned run');
+      expect(execution.exitCode).toBe(0);
+      // ...and NOTHING else moves. No resurrection: a proven cancel stays
+      // cancelled and keeps its cancel_state, because the closer that
+      // terminalised this row already consumed the outcome.
       expect(execution.status).toBe('cancelled');
-      expect(execution.stdout).toBeNull();
-      expect(execution.exitCode).toBeNull();
+      expect(execution.cancelState).toBe('confirmed');
+    },
+    30_000,
+  );
+
+  runDb(
+    'the late-output recovery is idempotent — a replayed frame cannot overwrite it',
+    async () => {
+      const fx = await makeFixture();
+      // `exit_code IS NULL` on the EXECUTION row is the idempotency guard under
+      // test. Once the first late frame fills the output, a second frame is a
+      // no-op rather than a last-writer-wins overwrite.
+      const { commandId, executionId } = await seedTimedOutRun(fx, {
+        executionStatus: 'cancelled',
+        executionCancelState: 'confirmed',
+      });
+
+      await sendWsResult(fx, commandId, {
+        status: 'completed',
+        exitCode: 0,
+        stdout: 'the first late frame',
+      });
+
+      // Put the COMMAND row back into the only shape `commandResultAcceptance`
+      // reopens. Without this the second frame is rejected one layer up and the
+      // assertion below would pass without the execution guard ever running —
+      // green for the wrong reason. Resetting isolates the guard being tested.
+      await getTestDb()
+        .update(deviceCommands)
+        .set({ status: 'failed', result: { status: 'timeout', error: 'reopened for replay' } })
+        .where(eq(deviceCommands.id, commandId));
+
+      await sendWsResult(fx, commandId, {
+        status: 'failed',
+        exitCode: 9,
+        stdout: 'a replayed frame that must not win',
+      });
+
+      const execution = await readExecution(executionId);
+      expect(execution.stdout).toBe('the first late frame');
+      expect(execution.exitCode).toBe(0);
+      expect(execution.status).toBe('cancelled');
+      expect(execution.cancelState).toBe('confirmed');
     },
     30_000,
   );
--- a/apps/api/src/jobs/staleCommandReaper.cancellation.test.ts
+++ b/apps/api/src/jobs/staleCommandReaper.cancellation.test.ts
@@ -1,0 +1,347 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { PgDialect } from 'drizzle-orm/pg-core';
+
+/**
+ * #3525 W03 closers 4 and 5 — cancel-command expiry and the cancellation sweep.
+ *
+ * `cancelling` must never be reachable-and-permanent. `reapStaleCancellations`
+ * owns that state exclusively; `reapStaleScriptExecutions` deliberately cannot
+ * see it, and the last test in the first block proves that from the compiled
+ * predicate rather than from a mock's canned answer.
+ */
+
+const {
+  selectMock,
+  updateMock,
+  scriptExecutionsTable,
+  deviceCommandsTable,
+  scriptsTable,
+  scriptExecutionBatchesTable,
+  createAuditLogAsyncMock,
+  recordCancelUnconfirmedMock,
+  applyAutomationActionTerminalMock,
+} = vi.hoisted(() => ({
+  selectMock: vi.fn(),
+  updateMock: vi.fn(),
+  scriptExecutionsTable: {
+    id: 'script_executions.id',
+    deviceId: 'script_executions.device_id',
+    orgId: 'script_executions.org_id',
+    status: 'script_executions.status',
+    scriptId: 'script_executions.script_id',
+    createdAt: 'script_executions.created_at',
+    startedAt: 'script_executions.started_at',
+    completedAt: 'script_executions.completed_at',
+    errorMessage: 'script_executions.error_message',
+    cancelState: 'script_executions.cancel_state',
+    cancelCommandId: 'script_executions.cancel_command_id',
+    cancelPrevStatus: 'script_executions.cancel_prev_status',
+    cancelRequestedAt: 'script_executions.cancel_requested_at',
+  },
+  deviceCommandsTable: {
+    id: 'device_commands.id',
+    type: 'device_commands.type',
+    status: 'device_commands.status',
+    payload: 'device_commands.payload',
+    createdAt: 'device_commands.created_at',
+    executedAt: 'device_commands.executed_at',
+    completedAt: 'device_commands.completed_at',
+    result: 'device_commands.result',
+    deviceId: 'device_commands.device_id',
+  },
+  scriptsTable: {
+    id: 'scripts.id',
+    timeoutSeconds: 'scripts.timeout_seconds',
+  },
+  scriptExecutionBatchesTable: {
+    id: 'script_execution_batches.id',
+    devicesTargeted: 'script_execution_batches.devices_targeted',
+    devicesCompleted: 'script_execution_batches.devices_completed',
+    devicesFailed: 'script_execution_batches.devices_failed',
+    status: 'script_execution_batches.status',
+    completedAt: 'script_execution_batches.completed_at',
+  },
+  createAuditLogAsyncMock: vi.fn().mockResolvedValue(undefined),
+  recordCancelUnconfirmedMock: vi.fn(),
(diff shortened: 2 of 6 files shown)


ToddHebebrand:

Four specialist agents (Sonnet) reviewed the PR diff read-only: **code-reviewer**, **pr-test-analyzer**, **silent-failure-hunter**, **comment-analyzer**. Every finding is either fixed in `9d28171af4` or dismissed with a reason below.

### Fixed — critical

**1. Batch counters never fired for a cancellation close** *(code-reviewer + pr-test-analyzer, independently)*

The counter block keys on `updatedExecutions`, which the cancellation CAS leaves empty. That was inherited from the #3607 precedent — but the two cases are opposites. #3607 is excluded because the sweep that stamped the row terminal **already spent** the batch slot. The cancellation CAS is itself the writer that terminalises the row, and nothing spent the slot before it: `applyScriptCancelAck` never touches the batch, and the cancel request only moves the row to the transient `cancelling`.

Consequence, verified by both reviewers reading the code: once this handler lands the execution is terminal and outside **both** reapers' predicates forever, so `devicesCompleted + devicesFailed` could never reach `devicesTargeted`. The batch's completion check never fires and **every multi-device run containing one cancelled device stays `pending` permanently** — the same never-terminal defect this wave exists to kill, one level up. `scriptExecution.ts` creates a batch row for any script dispatched to 2+ devices, so this is the ordinary "cancel one straggler out of a fleet run" case, not an edge.

Now counted via `updatedExecutions[0] ?? cancelClosed[0]`, through the same completed/failed mapping every other writer uses (there is no `devicesCancelled` column — batch cancel is out of scope per OD5-B). Two new tests; **both confirmed red against the pre-fix expression** before being kept.

### Fixed — important

**2. The sweep could win a race against a genuine `terminated` ack** *(code-reviewer)*

Both transports CAS `device_commands` to a terminal status *before* dispatching the result handler, so between that flip and `applyScriptCancelAck` writing its verdict there is a window where real proof-of-kill is in flight against a row that already reads terminal. `reapStaleCancellations` treated terminal-ness alone as licence to revert, won the CAS, made the ack a no-op ("must not resurrect a decided row"), and silently downgraded a **proven** cancel to `unconfirmed`. Honesty contract technically preserved, but a provable cancellation lost for no reason — and it scales with concurrent bulk cancellations.

A terminal command must now have **settled** for a full grace window before the sweep acts on it. A NULL `completed_at` counts as settled, so this cannot become a new way to strand a row. Costs one extra sweep cycle.

**3. Two broken-writer shapes were laundered into operational noise** *(silent-failure-hunter)*

`applyScriptCancelAck` captures a NULL `cancel_prev_status` as "a broken writer, not a race… could be misreporting execution history fleet-wide." The new reaper hit the identical condition and said nothing — on precisely the path that runs when no ack ever arrives, i.e. the one place that bug could hide. Now captured, as is a NULL `cancel_command_id` (a cancel that was never dispatchable at all, distinct from the *expected* orphan whose command row was reaped). The ordinary unacknowledged cancel stays silent by design.

The audit row also recorded `revertedTo: row.prevStatus` — writing `null` where the code wrote `running`. It now records the value actually written.

**4. The late-output write was unchecked** *(silent-failure-hunter)*

Closer 3's `UPDATE` never inspected `.returning()`, unlike the #3607 block 30 lines above it, yet logged `"recovered late output"` unconditionally. On a duplicate late frame the write is a no-op and the frame's output is dropped — while the log claims success, which is what makes a later "why is this output truncated" report un-debuggable. Now branches on the result and logs a discard as a discard.

### Fixed — suggestions

- **Secondary bookkeeping isolated per row** *(pr-test-analyzer)* — matching the deployment-result and backup-job reapers in the same file, so one failed audit can't strand the remaining rows for another cycle. Test included.
- **`phase` tag on the shared catch** *(silent-failure-hunter)* — the ladder is now five writes inside one `try`; without it every failure reached Sentry with identical context and grouping.
- **Cap-hit warning** when the sweep fills `MAX_REAP_PER_RUN`, matching `deviceCommands`.
- **Metric docblock** *(comment-analyzer)* — said "reached the end of their grace window"; three arms trigger it. Reworded.
- **Test gaps closed** *(pr-test-analyzer)*: NULL `cancel_requested_at` orphan arm; the `cmdStatus === 'cancelled'` third of the terminal disjunction; a multi-row sweep proving one row losing its CAS doesn't affect the others or the count; and the CAS predicate assertion strengthened from three independent `toContain`s to ordered param adjacency plus `not.toContain(' or ')`.

### Dismissed

- **pr-test-analyzer #7** — "`customFields/scriptWriteBack` may not exist." The agent was instructed not to open the worktree and grepped the wrong tree. The module exists and its mock is exercised by passing tests.
- **comment-analyzer #2** (informational) — comments describe the `cancelling` lifecycle in present tense while nothing yet writes that status. Correctly identified as *not* a comment-accuracy problem: no comment claims the cancel button already dispatches `script_cancel`. This is the intended wave sequencing (W03 lands the closers; the route rewrite is separate), and it's already recorded under **Not verified** in the PR body.
- **File size** — `staleCommandReaper.ts` is past the 500-line soft guideline. CLAUDE.md explicitly discourages splitting a working file to hit a line count; the reviewer flagged it as a non-finding.

### Confirmed correct (no action)

Ladder ordering is mutually exclusive and race-safe; idempotency for replayed frames holds end-to-end; `terminalStatus: 'cancelled'` only on the proven path; the device-command close marker matches `commandResultAcceptance`'s reopen predicate exactly; no `cancelling` row can be stranded (an undelivered cancel is bounded by the 2-hour `LONG_TIMEOUT` tier); `REAPER_DOMAINS` hoisting is behaviourally inert; `runWithSystemDbAccess` already wraps the new domain, so no explicit DB-context call is needed; audit conventions match the neighbouring reapers; `createAuditLogAsync` never throws, so a failed audit degrades to retry rather than losing the record. The comment-analyzer traced every cross-module claim (the `commandResultAcceptance` reopen predicate, the worker-role/`routes/metrics.ts` isolation, the `cancel_command_id` bare-uuid rationale, `reapStaleScriptExecutions`' predicate, the 2-hour tier, the OD3-A/OD8-C/OD9-C spec labels) and found **no** inaccuracies.

### Verification after fixes

- `npx vitest run src/services/commandResultHandlers src/jobs/staleCommandReaper src/services/workerEntrypointClosure src/services/scriptCancellation src/routes/agents/commands.test.ts src/routes/agentWs.enqueueContract.test.ts src/services/scriptSecretDelivery.test.ts src/services/scriptExecution` — **15 files, 387 tests, all pass**
- `npx tsc --noEmit` — clean · `pnpm lint` — clean
- Prior CI on this branch: 37 checks green, 0 failures (Lint and Type Check both passed). The push re-runs everything including the 4 integration shards.

**Not merging** — the PR stays open for review.

ToddHebebrand:
## CI fix — Integration Tests shard 2/4 (`9ac24c39f5`)

**Failure:**
```
FAIL src/__tests__/integration/lateCommandResultRecovery.integration.test.ts
  > #3607 late command result recovery
  > a CANCELLED execution is left alone — the operator abandoned the run
AssertionError: expected 'output for an abandoned run' to be null
```

### Which contract is right

The plan settles it, and it says the **new** behaviour. W03 Task 3.1 **Step 4** is titled *"Replace the 'drop the output' path with output recovery"* and says of the exact comment this test was written from:

> the comment there ("Dropping the output is the CORRECT outcome") is the design bug written down. A cancelled execution's partial output is exactly what the operator needs.

The plan's own W03 test block then asserts `fills stdout/stderr/exit_code but does NOT change status`. So the failing test encodes the defect being fixed, not a contract worth keeping.

**Rewritten, not deleted; closer 3 unchanged.** The test now asserts the full new contract rather than merely flipping one `expect`:

- the output lands (`stdout`, `exitCode`)
- **and nothing else moves** — `status` stays `cancelled`, `cancel_state` stays `confirmed`. No resurrection: whichever closer terminalised the row already consumed the outcome.

Renamed to *"a CANCELLED execution KEEPS the late output — it is what the operator stopped the run to see"*, with a comment recording that this is a deliberate contract change and why. The race it describes is unchanged and still routine; only what happens to the output changed.

A second test covers **replay idempotency**.

### The replay test would have been green for the wrong reason

Worth calling out: it resets the command row to the reopenable marker (`failed` + `result.status = 'timeout'`) between the two frames. Without that, the second frame is rejected by `commandResultAcceptance` one layer up and never reaches `handleScriptResult` — the assertion would pass without the execution-row guard ever running.

Proven by mutation: removing `isNull(exitCode)` from closer 3 turns it red with `expected 'a replayed frame that must not win' to be 'the first late frame'`. So the guard under test really is the one being exercised, not the acceptance layer.

### Sweep for other tests pinning the old contract

`grep -rn "abandoned\|left alone" apps/api/src` — **this was the only one.** The sibling *"an execution that already carries real output is NOT overwritten"* seeds `executionStatus: 'failed'`, so closer 3 never fires there; untouched and still passing.

Two small helper additions in the same file: `seedTimedOutRun` accepts `executionCancelState` (with the paired `cancel_requested_at` the CHECK constraint requires), and `readExecution` returns `cancelState` so the no-resurrection assertion is possible.

### Verified against a live Postgres

`pnpm test-stack up` … torn down after (`docker compose ls -a` confirms nothing left running).

- `lateCommandResultRecovery.integration.test.ts` — **12/12 pass**
- `automationTerminalReconciliation` + `staleBackupReaper` + `deviceUninstallDrain` (the other integration suites importing these modules) — **15/15 pass**
- Unit: `commandResultHandlers` + `staleCommandReaper` + `workerEntrypointClosure` + `scriptCancellation` — **235/235 pass**
- `npx tsc --noEmit` clean · `pnpm lint` clean

No production code changed in this commit — closer 3 is exactly as reviewed.

PR #4968, opened 2026-09-05
feat(auth): Apple App Attest verifier for iOS approver-key registration (#4710)

### LINKED ISSUE #4710: W03: Apple App Attest verifier (API): nine checks, synthetic-CA fixtures, pinned Apple root
Wave W03 of #4707 — Apple App Attest verifier (API): nine checks, synthetic-CA fixtures, pinned Apple root
Branch: `feature/4707-mobile-platform-attestation-l4/wave-4710`

ToddHebebrand:
Closes #4710

Wave **W03** of feature #4707 (`docs/superpowers/plans/security-auth/2026-09-02-mobile-platform-attestation-l4.md`) — the iOS half of server-verified platform attestation for L4 (critical-tier) approvals. API only; no client change, no migration.

## What changed

**New: `apps/api/src/services/attestation/appleAppAttest.ts`** — a pure verifier for Apple App Attest attestation objects. No db, no Redis, no env reads, no clock of its own (`now` is an input), so the whole of Apple's documented procedure runs in CI with no Apple round-trip. It implements all nine checks:

| # | Check |
|---|---|
| 1 | CBOR `fmt` is `apple-appattest` |
| 2 | `x5c` chain verifies to the **pinned** Apple App Attestation Root CA, with every cert's validity window honoured against `now`. The system trust store is never consulted — an attestation signed by any public CA fails exactly as hard as one signed by nobody |
| 3 | `nonce = SHA256(authData ‖ clientDataHash)`, where `clientDataHash` is the server-derived W02 registration transcript |
| 4 | that nonce equals the octet string in credCert extension OID `1.2.840.113635.100.8.2` (strict DER: short-form lengths, exact nesting, no trailing bytes) |
| 5 | `SHA256(credCert public key, uncompressed X9.62)` equals `keyId` |
| 6 | `authData.rpIdHash` equals `SHA256(appId)` |
| 7 | `authData.signCount` is `0` |
| 8 | `authData.aaguid` matches the **configured** environment (`appattest\0×7` / `appattestdevelop`) |
| 9 | `authData.credentialId` equals `keyId` |

**New: `__fixtures__/appAttestFixture.ts`** — mints a synthetic attestation signed by a throwaway root (`@peculiar/x509` + `@levischuck/tiny-cbor`). Every negative test takes a **valid** fixture and mutates exactly one field, so a green negative proves the verifier rejected *that* mutation rather than tripping over something incidental.

**New: `appleAppAttestRootCA.pem` + `appleAppAttestRootCA.ts`** — the pinned trust anchor (see "Root CA provenance" below).

**Modified: the iOS branch of `verifyPlatformAttestation`** (`services/authenticatorAttestation.ts`). A pass maps to a basis by approval-key algorithm:

- `ES256` → `ios_se_p256_app_attest` (**in** `L4_TRUSTED_PLATFORM_BOUND_BASES`)
- `RS256` → `ios_keychain_rsa_app_attest` (**deliberately not** L4-trusted — the Secure Enclave holds only P-256 keys, so an RSA key cannot be resident in it, and App Attest says nothing about where the *approval* key lives)

A verification failure **downgrades to `unattested`, it does not 5xx**: the phone still enrols and works at L2/L3, it just never reaches L4. The shapes that land there are client-provokable (a dev build against a production-configured server, a stale attempt, a forged blob) and none of them should block a technician from enrolling a device that is still useful at lower tiers. The reason is logged; `evidence` on an unattested result stays empty so no unverified claim is ever persisted.

`evidence` on a **pass** stores digests only — `attestedAppAttestKeySha256` and `receiptSha256` — never the raw receipt, which is a bearer artifact for Apple's fraud-metric endpoint.

**Modified: `config/env.ts`** — `APPLE_APP_ATTEST_APP_ID` (default `D8W6N2JYMA.com.breeze.rmm`) and `appleAppAttestEnvironment()` (default `production`; only the exact string `development` opts out, so a typo, an empty value or an unset variable cannot open the door to developer-signed builds).

## Root CA provenance

Fetched from Apple's public CA page (<https://www.apple.com/certificateauthority/private/> → `Apple_App_Attestation_Root_CA.pem`) and pinned at `apps/api/src/services/attestation/appleAppAttestRootCA.pem`.

- **SHA-256 of the committed `.pem` file:** `c778d09ac341f7fd9f8f3b19e2b815af6aed4ad4490e1e92c05cb355212a5013`
- **Certificate SHA-256 fingerprint (DER):** `1C:B9:82:3B:A2:8B:A6:AD:2D:33:A0:06:94:1D:E2:AE:4F:51:3E:F1:D4:E8:31:B9:F7:E0:FA:7B:62:42:C9:32`
- Subject/Issuer `CN=Apple App Attestation Root CA, O=Apple Inc., ST=California`, valid 2020-03-18 → 2045-03-15, self-signed, CA:TRUE

## 🛠 Ops note — new env vars

Two new **optional** variables. Both fail closed, so leaving them unset can only reject genuine attestations, never accept forged ones:

| Var | Default | Meaning |
|---|---|---|
| `APPLE_APP_ATTEST_APP_ID` | `D8W6N2JYMA.com.breeze.rmm` | Apple's `<TeamID>.<bundle id>`, hashed into the attestation's `rpIdHash` |
| `APPLE_APP_ATTEST_ENVIRONMENT` | `production` | Which aaguid sentinel is accepted. **Set to `development` only on a non-production deployment** — development attestations come from any developer-signed build of the app |

Neither needs to be set for the hosted droplets to behave correctly; the compiled defaults are the production values. **If you do set one**, it must go in `/opt/breeze/.env` **and** be mapped in the `api` service's `environment:` block of `/opt/breeze/docker-compose.yml` — compose interpolation only happens for listed vars, so an unmapped value in `.env` is a silent no-op. Both are already documented in `deploy/.env.example` and mapped in `deploy/docker-compose.prod.yml` in this PR (the `envComposeParity` guard requires it, including for commented-out defaults).

1. **`mintAppAttestFixture` is `async`, not sync.** The plan sketched a synchronous signature; `@peculiar/x509` mints certificates through WebCrypto, which is promise-based, and making it sync would mean hand-rolling ASN.1 signing. The verifier under test stays synchronous — only fixture construction awaits.
2. **`@peculiar/x509` is a production dependency, not a devDependency.** The plan put it in `devDependencies` and had the verifier use `node:crypto` alone. Node's `X509Certificate` exposes no generic extension accessor, so reading OID `1.2.840.113635.100.8.2` needs either this library or a hand-rolled DER walker — and hand-rolled ASN.1 in an auth-path verifier is the worse long-term choice. It is used for extension extraction only; signature verification and validity windows go through `node:crypto`. (W04 brings `@peculiar/asn1-android` + `@peculiar/asn1-x509` as production deps anyway, so the ASN.1 stack was already destined to be one.)
3. **The pinned root is also inlined as a `.ts` constant, not only committed as a `.pem`.** `apps/api` ships as a single `tsup` bundle (`dist/index.cjs`) and both runtime images copy only `migrations/`, `assets/` and `ee/workspace/` beside it — no `src/` tree, and `__dirname` resolves to `/app/dist`. A `readFileSync` of the `.pem` would therefore have thrown in production and nowhere else. The `.pem` is kept as the auditable provenance artifact; `appleAppAttest.test.ts` asserts the two are byte-identical and pins the certificate fingerprint, so they cannot drift and neither can be swapped for a different CA without a failing test. **Verified**: `grep`'d the built `dist/index.cjs` and confirmed the root PEM is inlined.
4. **`verifyPlatformAttestation` gained a required `publicKeyAlg` field**, which forced two adjacent edits outside this wave's nominal file list: one line in `routes/authenticator.ts` (it already had `body.publicKeyAlg` in scope, proven by the PoP check immediately above) and the existing call sites in `authenticatorAttestation.test.ts` / `authenticatorAssurance.test.ts`. It is required rather than optional deliberately: an optional field would make a forgetful future caller silently fall through to the weaker basis, which is the exact class of silent downgrade this plan exists to remove. A compile error is the right forcing function.
5. **Two extra checks and several extra tests beyond the plan's list**: `attStmt.receipt` must be present (the plan's `AppAttestResult` requires it), the credCert key must be P-256 EC, plus tests for check 9 (`credentialId === keyId`, which the plan listed as a check but not as a test), a production-aaguid-in-development rejection, an unrecognized aaguid, an empty `x5c`, a truncated `authData`, a spliced foreign intermediate, and the "valid at the supplied `now`" positive case.
6. **The plan's Step 6 snippet wrote `receiptB64` into a `receiptSha256` over `Buffer.from(receiptB64, 'base64')`** — kept verbatim, but note the digest is over the decoded receipt bytes, not the base64 text.

## Verification

Run in the foreground on this branch, at `5a12405e73`:

- `npx vitest run src/services/attestation/ src/services/authenticatorAttestation.test.ts src/services/authenticatorAssurance.test.ts src/routes/authenticator.test.ts src/config/env.test.ts` → **5 files, 187 tests, all pass**
- `npx vitest run src/config/` → **22 files, 611 tests, all pass** (includes `envComposeParity`, `extensionBundledDependencies`, `composeBindMounts`)
- `npx tsc --noEmit` → **clean**
- `npx eslint src` → **clean**
- `npx tsup` → **build success**; `dist/index.cjs` confirmed to contain the inlined root PEM

**Red-first evidence:** the suite was first run against a deliberately vacuous verifier that only parsed CBOR and returned the key — **16 of 25 tests failed**, one per missing check. Each check was then added individually and the failure count watched drop (16 → 15 → 11 → 8 → 6 → 5 → 4 → 1 → 0). No check was written before its test was seen failing for the right reason.

## Not verified

- **No real Apple attestation was ever verified.** There is no way to obtain one in CI; every test runs against the synthetic CA. The pinned root is exercised only in the negative direction (a synthetic chain must *not* validate against it). The first genuine end-to-end proof comes in **W05**, when the iOS native module mints a real App Attest key on a device.
- **The `authData` layout is asserted only against my own fixture.** If Apple ever appends a COSE public key after `credentialId` the parser tolerates it (trailing bytes are allowed and the nonce covers the whole buffer), but that tolerance is untested against a real blob.
- **Integration / RLS contract suites were not run** — this wave adds no migration, no table and no column, and touches nothing in `tenantCascade.ts` or `tenantExportPolicyRegistry.ts`. CI's `integration-test` job will run them on this PR regardless.
- **The `receiptB64` value is not used for anything yet.** Nothing in this wave calls Apple's fraud-metric endpoint; only its digest is stored.
- **No production deploy or droplet change was made.** See the ops note above.

https://claude.ai/code/session_01QWduCvVLT1JDSFA3LAXBN5


ToddHebebrand:
## Review summary

Three independent reviewers (Sonnet) against the pushed diff — security/code, test coverage, silent-failure. **No critical findings.** Four findings accepted and fixed in `a8d153b6ec`; one dismissed with reason.

### Fixed

**1. The dispatcher's `catch` could not tell a rejection from a defect** — *security review, Important*

`verifyIosAttestation` swallowed *any* throw identically. An `AppAttestVerificationError` is the verifier working as designed; a `TypeError` out of a future `@levischuck/tiny-cbor` or `@peculiar/x509` bump means **every legitimate iOS device in the fleet is silently losing L4** while the log line reads exactly like ordinary attacker noise (dev builds, stale attempts, forged blobs). Now branches on `instanceof AppAttestVerificationError`: log-only for a rejection, `captureException` for anything else — the same call `consumeRegistrationAttempt` already makes twenty lines up in the same file, for the same reason. Both directions are tested, including that a rejection does **not** page.

**2. A fleet-wide misconfiguration was invisible** — *silent-failure review, Critical*

A stale `APPLE_APP_ATTEST_APP_ID` or a wrong `APPLE_APP_ATTEST_ENVIRONMENT` rejects **100% of genuine enrolments, indefinitely**, and each rejection is per-request indistinguishable from one attacker being turned away. A `console.warn` cannot be aggregated after the fact. `AttestationResult` now carries an optional `failureReason` — set only when a verifier actually ran and refused, never for a platform with no verifier wired — and the route records it in the registration audit row's `details`. "Why did nobody reach L4 last month" becomes one query instead of a stdout grep against whatever retention survived. The verifier's reason strings describe the *server's* checks (`"rpIdHash does not match the configured appId"`), never client bytes, so they are safe to persist. The warn line now also logs `appId`: the likeliest cause of a fleet-wide rejection and the one value an operator cannot see from outside.

I did **not** add a Prometheus counter — `breeze_authenticator_l4_basis_total` and alerting are W07's surface, and the audit field covers the forensic need now.

**3. `appleAppAttestEnvironment()` resolved a typo silently** — *silent-failure review, High*

`Development`, `dev`, or a trailing space resolved to `production` with no trace, unlike its neighbour `authenticatorAttestationEnforced()` which warns on an unrecognized value. Now warns. Deliberately still a warning and not a boot refusal (which the reviewer suggested): the wrong value can only ever *reject*, never *admit*, so bricking a deploy over it is disproportionate.

**4. The env helpers had zero tests** — *test review, rating 7 — the one gap it would block on*

`appleAppAttestEnvironment()` is the entire enforcement of this wave's fail-closed claim and nothing stood behind it; a refactor inverting the comparison to `!== 'production'` would have started accepting developer-signed attestations in production with CI green. Added five cases in `config/env.test.ts` (inversion, typo, whitespace, warn/no-warn split, appId fallback). **Mutation-checked**: stubbing the warn branch to `if (false)` turns the warn test red.

### Also added

From the reviewers' untested-path lists plus my own adversarial pass — all six shapes were **already handled correctly**, none had a test:

- `x5c` whose only element **is** the pinned root (self-anchoring: the loop is skipped, a self-signed cert "verifies against itself")
- `x5c` repeating the credCert as its own issuer
- an unparseable trust anchor, an empty anchor list, and a multi-anchor rotation
- an over-long chain

### Dismissed

`signedBy()`'s bare `catch {}` discards the underlying exception (*silent-failure review, Low*). It is a fail-closed signature check inside a module that must stay pure (no logger), and "not signed" is the correct and only outcome either way — the caller already surfaces a specific reason string. Logging inside it would add noise to the hot path for a case that is not actionable.

### Notable confirmations (things reviewers checked and cleared)

- **CBOR duplicate-key smuggling is impossible** — `@levischuck/tiny-cbor`'s `decodeMap` throws on a repeated key, and `decodeCBOR` asserts the full byte stream was consumed, so no trailing-garbage smuggling either. (Reviewer read the dependency source, not its docs.)
- **The strict DER nonce parse is overread-safe** — out-of-bounds `Buffer` indexing yields `undefined`, which fails the tag check; all three declared lengths are pinned back to `body.length`, so they cannot diverge; indefinite-length BER is explicitly rejected.
- **`parseAuthData`'s trailing-byte tolerance is safe** — the nonce is computed over the entire raw `authDataBuf`, so trailing-byte manipulation changes the nonce and is caught.
- **The ES256/RS256 basis split is not confusable** — `publicKeyAlg` is inside the signed registration transcript *and* `verifyMobileSignature` independently proves the key really is EC P-256 / RSA≥2048 before the PoP passes, so the algorithm label is a proven property of the key, not a client claim.
- **Chain validation is sound** — DN-string comparison is a pre-check only; the trust decision is always a real `cert.verify(issuerPublicKey)`. The system trust store is never consulted.
- **9/9 checks verified non-vacuous** — the test reviewer traced, per check, that deleting *only* that check turns a specific named test red, with no cross-contamination. (Checks 3 and 4 are one `bytesEqual` in the implementation, as Apple's steps 3/4 are one equality test — noted, not a defect.)
- **The root-CA pin is real** — the reviewer independently recomputed the shipped `.pem`'s SHA-256 fingerprint with `openssl` and it matches the pinned literal. Swapping both files to a spoofed-DN CA still fails the fingerprint test.

- `npx vitest run src/services/attestation/ src/services/authenticatorAttestation.test.ts src/services/authenticatorAssurance.test.ts src/routes/authenticator.test.ts src/config/` → **26 files, 783 tests, all pass**
- `npx tsc --noEmit` → clean
- `npx eslint src` → clean

### Still not verified (unchanged from the PR body)

No real Apple attestation has ever been through this code — there is no way to obtain one in CI, so every test runs against the synthetic CA and the pinned root is exercised only in the negative direction. **The first genuine end-to-end proof comes in W05**, when the iOS native module mints a real App Attest key on a device.

PR #4969, opened 2026-09-05
feat(auth): Android Key Attestation + Play Integrity verifiers (#1374 W04)

### LINKED ISSUE #4711: W04: Android Key Attestation + Play Integrity verifiers (API): reject software keymint, unlocked bootloaders, foreign packages
Wave W04 of #4707 — Android Key Attestation + Play Integrity verifiers (API): reject software keymint, unlocked bootloaders, foreign packages
Branch: `feature/4707-mobile-platform-attestation-l4/wave-4711`

ToddHebebrand:
Closes #4711

Wave W04 of feature #4707 (plan `docs/superpowers/plans/security-auth/2026-09-02-mobile-platform-attestation-l4.md`). API only — no migration, no client change. The new endpoints stay unreachable-by-a-real-phone until W06 ships the Android native module.

## What changed

**`services/attestation/androidKeyAttestation.ts` (new)** — pure, synchronous Android Key Attestation verifier. Given a `KeyStore.getCertificateChain()` chain it proves, and throws on any of:

1. chain contiguous, every link's signature verified, every device-supplied cert inside its validity window, terminating in a **pinned Google hardware attestation root**
2. leaf carries extension OID `1.3.6.1.4.1.11129.2.1.17`, parsed as `KeyDescription` (strict schema, falling back to `NonStandardKeyDescription` for devices that emit the authorization lists out of order)
3. `attestationChallenge` equals the registration transcript, byte-exact and constant-time
4. **`keyMintSecurityLevel ∈ {TrustedEnvironment, StrongBox}` — `Software` is rejected.** This is the point of the wave
5. `attestationSecurityLevel` likewise
6. `teeEnforced.rootOfTrust.deviceLocked === true` and `verifiedBootState === 'Verified'`
7. attested package name equals `com.breeze.rmm`
8. `purpose`/`digest` are **hardware-enforced** (present in `teeEnforced`, absent from `softwareEnforced`), constrained to SIGN + SHA-256; `allApplications` absent; `origin === GENERATED`; `noAuthRequired` absent
9. returns the attested leaf SPKI so the caller can bind it to the key actually being registered

**`services/attestation/googleHardwareAttestationRoots.pem` + `.ts` (new)** — the five roots published at <https://developer.android.com/privacy-and-security/security-key-attestation>, fetched 2026-09-05. The `.pem` is the auditable provenance artifact; the `.ts` is generated from it and is what the verifier actually imports, because the production image ships no `src/` tree (see the follow-up fix at the bottom).

- **SHA-256 of the committed file: `f2256be131ef0db817d78e1e5c1223af67e70d78827d516170a2cf7af9e36fcb`**
- Four are RSA-4096 certificates over the *same* key (the long-lived root, re-issued with later validity windows); the fifth is the 2025 EC P-384 `Key Attestation CA1`.
- The verifier anchors on **SubjectPublicKeyInfo**, not certificate bytes, so any re-issue of the same root still chains. One pinned certificate (serial `E8FA196314D2FA18`) **expired 2026-05-24** and is deliberately kept: devices in the field still terminate chains in it, and RFC 5280 §6.1 applies no validity check to a trust anchor. The verifier enforces validity on the leaf and intermediates only.

**`services/attestation/playIntegrity.ts` (new)** — Play Integrity verdict verification via the Google-managed `decodeIntegrityToken` endpoint, authenticated with a service account (`jose` signs the JWT-bearer assertion; the access token is cached until 30 s before expiry). Checks package (both `requestDetails` and `appIntegrity`), `requestHash` binding, freshness (300 s, 60 s future skew), `PLAY_RECOGNIZED`, and `MEETS_DEVICE_INTEGRITY`.

**`services/authenticatorAttestation.ts`** — Android branch of `verifyPlatformAttestation` only. The iOS branch and its imports are untouched (W03 owns them).

**`config/validate.ts`** — `PLAY_INTEGRITY_SERVICE_ACCOUNT`.

## Design decisions worth reviewing

**Play Integrity never sets the basis, and both directions of its failure are deliberate.** Per plan decision 3, Key Attestation alone decides `platform_bound_basis`; a passing verdict only stamps `app_integrity_verified_at`. The split on failure:

| Situation | Result | Why |
|---|---|---|
| No service account configured | `null`, Key-Attestation basis stands | An unconfigured deploy must not refuse Android registrations |
| Google unreachable / API error | `null` + Sentry + log, basis stands | Our infrastructure, not evidence against the client |
| Client supplied no token | not called, basis stands | Enterprise / non-Play distribution is supported |
| Verdict is disqualifying | throws → branch downgrades to **`unattested`** | A device Play says is compromised is not somewhere a critical-tier approver key lives, StrongBox or not |

**`verifyPlatformAttestation` never throws.** `routes/authenticator.ts:598` does not wrap the call, so a throw would 500 a registration that should degrade. Every Android failure returns `unattested` with the reason in `evidence.rejected`, which lands on the device row — an operator debugging "why is my phone not L4" gets the reason instead of nothing. Failures are `console.warn`, not Sentry: a failed attestation is the routine attacker-shaped case and a caller could flood it.

1. **`attestationApplicationId` is read from `softwareEnforced` as well as `teeEnforced`** (plan check 7 said `teeEnforced` only). Real KeyMint implementations put it in the software-enforced list — Keystore, not the TEE, is what knows the calling package — so the literal check would reject essentially every real device. The claim is still worth something precisely because the *same* attestation independently proves, TEE-enforced, that the bootloader is locked and verified boot passed, i.e. that the Keystore making the claim is the stock measured one. `teeEnforced` wins when both are present. **This is the single highest-risk deviation: if a reviewer believes real devices do put it in `teeEnforced`, say so and I will tighten it.**
2. **`mintAndroidKeyAttestationFixture` is `async`.** `@peculiar/x509` signs through WebCrypto; there is no synchronous X.509 generator and every caller is an async test.
3. **Three checks added beyond the plan's nine**: `origin === GENERATED` (an IMPORTED key's private half existed outside secure hardware), `noAuthRequired` absent (the key must be hardware-gated on a biometric prompt), and `allApplications` absent (the key must be bound to this app). Each has its own test.
4. **`expectedRequestHash` is a required option on `verifyPlayIntegrityToken`, not optional.** Without it a verdict minted by any other session of the app on the same device replays inside the 300 s freshness window. The Android branch passes `transcript.toString('base64url')`. **W06 contract: the Kotlin module MUST call `setRequestHash(<that same base64url digest>)`** or every supplied token is rejected. Stated here because W06 is unwritten and this is a hard requirement on it.
5. **`google-auth-library` was not used** even though it is already a direct dependency and does token minting; the plan said `jose`, `jose` is also already a direct dependency, and the JWT-bearer exchange is ~20 lines. No strong opinion — happy to swap if a reviewer prefers the official client's refresh handling.
6. **The Android-branch wiring tests live in `services/attestation/androidPlatformAttestation.test.ts`**, not in `authenticatorAttestation.test.ts`. W03 is editing the sibling branch of that same file in parallel; putting a new `describe` block there was an avoidable merge conflict. The one line I did have to touch there is the W02 stub test's title (`resolves unattested for Android — no verifier is wired until W04` → `…when the chain does not verify`); its assertion is unchanged and still passes.
7. `@peculiar/asn1-schema` was promoted alongside the two packages the plan named (it is where `AsnParser`/`OctetString` come from), and `@peculiar/x509` was added as a **devDependency** for the fixture. All four were already in `pnpm-lock.yaml` transitively via `@simplewebauthn/server`, at the same versions.

## Ops note — new environment variable

`PLAY_INTEGRITY_SERVICE_ACCOUNT` — a Google Cloud service-account key (raw JSON **or** base64 of that JSON) for a project with the **Play Integrity API** enabled, granted access to the `com.breeze.rmm` Play app.

- **Optional.** Absent → Android registrations verify by Key Attestation alone and store a null `app_integrity_verified_at`. No behaviour regresses.
- **Format is validated at boot** (unlike most optional strings in `validate.ts`): a mangled paste otherwise reads as "not configured" forever, so every Android registration silently skips the check and nobody notices. A bad value now refuses to start.
- Per CLAUDE.md, when this is set on a droplet it must also be mapped explicitly in the `api` service's `environment:` block — a value in `.env` alone is not interpolated.
- Nothing in this PR requires it to be set for v0.110; it is only useful once W06 ships an Android client that produces tokens.

## Verification

| Check | Result |
|---|---|
| `vitest run src/services/attestation/ src/services/authenticatorAttestation.test.ts` | 4 files, **74 passed** |
| `vitest run src/routes/authenticator.test.ts src/services/authenticatorAssurance.test.ts src/config` | 24 files, **718 passed** |
| `npx tsc --noEmit` (apps/api) | exit 0 |
| `npx eslint` on every changed file | exit 0 |
| Mutation check | Disabling the key-binding check and mis-mapping the StrongBox basis each produced exactly one failing test — the wiring tests discriminate rather than confirm |

## Not verified

- **No real device.** Every fixture is synthetic (own root, own KeyDescription). Nothing here has been run against a chain from a physical TEE or StrongBox phone; that is W06's job and it is the thing most likely to surface a wrong assumption — deviation 1 above in particular.
- **No live Play Integrity call.** `decodeIntegrityToken` and the OAuth JWT-bearer exchange are exercised only through the injected decoder seam. The real HTTP path, the response shape, and the token cache have never executed.
- **Certificate revocation is not checked.** Google's status list at `https://android.googleapis.com/attestation/status` is a network call and is deliberately outside this pure module, so a revoked attestation key still verifies. Called out in the module docblock; worth a follow-up issue if we want it.
- **Integration / RLS suites were not run.** This wave adds no migration, touches no schema, and adds no table, so neither contract applies — but I did not run them.
- **`pnpm lint` was not run repo-wide**, only `eslint` over the changed files.
- The pinned roots were fetched from the public Android documentation page over HTTPS on 2026-09-05 and their self-signature was checked in a test; they were **not** cross-checked against a second source.

https://claude.ai/code/session_01ELShYDurJPLi6DpQFYCpGh

---

## Review round (4 agents) — 2 forgery paths found and fixed

A code-review agent and an adversarial crypto-review agent **independently** found the same complete bypass, and the crypto reviewer found a second variant. Both are fixed in `331427a07e`; the PR as first opened was vulnerable.

**Bypass A — no CA check on chain issuers.** Anyone can generate an ordinary KeyStore key with purpose SIGN and digest SHA-256 (standard API, no root needed) and receive a *genuine* Google-rooted attestation chain for it. Signing an X.509 `TBSCertificate` is just signing bytes with SHA-256, so the attacker has the secure hardware sign a certificate they authored for a **software** key of their own, carrying a hand-written `KeyDescription` asserting StrongBox, locked bootloader, verified boot, our package and the correct challenge — then prepends it. Every signature verifies and the chain ends at a real pinned Google root. Result: `android_strongbox_key_attestation` for a key the attacker holds entirely in software.
→ Fixed: every issuer in the chain must be a CA. Verified empirically that node's `X509Certificate.ca` folds in both `basicConstraints` and `keyUsage: keyCertSign`, **and** that `checkIssued()` alone returns `true` for an issuer carrying neither extension (an OpenSSL compatibility allowance) — so the explicit `.ca` check is load-bearing, not redundant. Three regression tests.

**Bypass B — the KeyDescription was read from `chain[0]` by assumption.** Google's documentation says explicitly: *"Don't assume that the key attestation certificate extension is in the leaf certificate of the chain. Only the first occurrence of the extension in the chain can be trusted. Any further instances … might have been issued by an attacker extending the chain."* The CA check happens to block bypass A's shape, but only because real KeyStore leaves are CA:FALSE — which nothing in this repo can prove for every KeyMint implementation.
→ Fixed: exactly one certificate in the chain may carry the extension, and *that* certificate is the attested one. A regression test models a CA-shaped genuine leaf so the count is exercised on its own, independent of the CA check.

### Other changes from the review

| Finding | Change |
|---|---|
| Unimplemented "chain stopped at an intermediate" fallback the comment promised | Removed. The terminal certificate's key must be pinned; fail closed |
| `catch (err)` treated a verifier **bug** identically to attacker noise | Narrowed: typed `AndroidKeyAttestationError`/`PlayIntegrityVerdictError` stay quiet; anything else goes to Sentry. 100% of devices failing can no longer look like 100% of devices attacking |
| `evidence.rejected` was discarded before the INSERT — the PR's own claim that it reached the row was **false** | `/devices/mobile/verify` now persists evidence for a rejected attestation. The claimed-but-unusable case still blanks it (W02's deliberate rule). Route test + mutation-checked |
| `deviceLocked: true` was a hardcoded literal; its test echoed the constant | Read back from the attestation |
| The `softwareEnforced`-duplicate branch of check 8 was unreachable in tests | New fixture override + test; the `purpose`-in-software test now asserts the specific message |
| No `origin`/user-auth/purpose-breadth checks | Reject IMPORTED keys, `noAuthRequired`, `allApplications`, and any purpose beyond SIGN\|VERIFY — `ATTEST_KEY` (7) would let the key mint its own chain |
| Play Integrity access-token cache survived a service-account rotation for ~1h | Keyed by `clientEmail` |
| `PLAY_INTEGRITY_SERVICE_ACCOUNT` boot check was weaker than claimed | Structural PEM-armour check added; the claim is now stated as structural-not-cryptographic in both the code and above |
| `requestHash` compared with `!==` | `crypto.timingSafeEqual` |
| One Sentry tag for both a Google outage and a permanently broken key | Error name added to the tag |
| Discarded ASN.1 parse errors | Both strict and permissive messages are carried; an unparseable `attestationApplicationId` no longer reports as "absent" |
| **No end-to-end test**: both isolated suites could be green while the canonical-SPKI key-binding comparison never matched | New `androidAttestationEndToEnd.test.ts` — a real fixture chain through the real verifier through the real branch, including "genuine chain paired with a DIFFERENT registered key" |

### Dismissed, with reasons

- **Signature-algorithm allow-list on the chain** — forging under a weak *declared* algorithm still needs Google's private key, and the pinned roots are RSA-4096 / P-384. Node's OpenSSL security level already refuses SHA-1 chains.
- **Play Integrity `certificateSha256Digest` unchecked** — there is no configured expected signing-certificate digest to check it against; adding one is a new env var, and this module is non-authoritative for the basis.
- **Certificate revocation** — real, and NOT fixed. `https://android.googleapis.com/attestation/status` is a network call, deliberately outside this pure module. A revoked attestation key still verifies. **Recommended follow-up before this feature is relied on in anger** (see the Not-verified list).

`vitest run src/services/attestation/ src/services/authenticatorAttestation.test.ts src/routes/authenticator.test.ts src/services/authenticatorAssurance.test.ts src/config` → **29 files, 808 passed**. `tsc --noEmit` exit 0. `eslint` on every changed file exit 0.

Every new guard was **mutation-checked**: disabling the CA check fails exactly 3 tests, disabling the extension count fails 1, reverting the evidence persistence fails 1, and mis-mapping the StrongBox basis / removing the key binding fail 1 each. The tests discriminate rather than confirm.

**Still not verified** (unchanged from above, and the revocation gap now added): no real device, no live Play Integrity call, no integration/RLS suites (this wave adds no migration and no schema change), and `pnpm lint` was not run repo-wide.

---

## Follow-up fix — `2a4121b774`: pinned roots inlined (production-only ENOENT)

Caught on the merge check. `androidKeyAttestation.ts` built its roots path from `import.meta.url` and `readFileSync`'d the `.pem` on first use. The runner stage of `apps/api/Dockerfile` copies only `dist`, `package.json`, `node_modules`, `packages`, `ee`, `migrations` and `assets` — **never `apps/api/src`** — so in the image the path resolved under `/app/apps/api/dist`, `readFileSync` threw ENOENT, and the scoped catch downgraded **every** Android registration to `unattested`. Invisible in tests and in dev, where the file sits beside the source.

The roots now ship as a TS constant the bundler inlines (same fix and rationale as the Apple sibling wave, #4968). The `.pem` stays committed as the provenance artifact; two mutation-checked tests assert the constant is byte-identical to it and pin the set's SHA-256, so the pair cannot drift.

Verified against a real `npx tsup` build: `dist/index.cjs` carries all five root certificates verbatim and zero references to the `.pem` path. Final gate: **29 files / 810 tests passed**, `tsc --noEmit` exit 0, `eslint` exit 0.


ToddHebebrand:

Four Sonnet reviewers ran in parallel against `gh pr diff 4969` (read-only, no worktree access): general code review, adversarial cryptographic review, silent-failure hunt, and test-coverage analysis. **16 findings; 13 fixed, 3 dismissed with reasons.** All fixes are in `331427a07e`.

### Critical — 2 forgery paths, both fixed

**A. Missing CA check on chain issuers** — found *independently* by the code reviewer and the crypto reviewer, and (separately) during my own re-read before either returned.

Anyone can generate an ordinary KeyStore key with purpose SIGN + digest SHA-256 — standard API, no root required — and receive a **genuine** Google-rooted attestation chain for it. Signing an X.509 `TBSCertificate` is just signing bytes with SHA-256, so the attacker has the secure hardware sign a certificate they authored for a **software** key of their own, carrying a hand-written `KeyDescription` claiming StrongBox, a locked bootloader, verified boot, `com.breeze.rmm` and the correct challenge — then prepends it. Every signature in that chain verifies and it terminates at a real pinned Google root. The result is a trusted `android_strongbox_key_attestation` basis for a key the attacker holds entirely in software: a complete defeat of the property this wave exists to establish.

Fixed by requiring every issuer in the chain to be a CA. I verified empirically rather than assuming: node's `X509Certificate.ca` folds in both `basicConstraints` **and** `keyUsage: keyCertSign`, and `checkIssued()` on its own returns `true` for an issuer carrying *neither* extension (an OpenSSL compatibility allowance) — so the explicit `.ca` check is load-bearing, not redundant with `checkIssued`. Three regression tests; disabling the check fails exactly those three.

**B. The KeyDescription was read from `chain[0]` by assumption** — crypto reviewer, citing Google's own documentation: *"Don't assume that the key attestation certificate extension is in the leaf certificate of the chain. Only the first occurrence of the extension in the chain can be trusted. Any further instances have not been issued by the secure hardware and might have been issued by an attacker extending the chain."*

Fix A happens to block bypass A's shape, but only because real KeyStore leaves are CA:FALSE — which nothing in this repo can prove holds for every KeyMint implementation in the field. So this is now an independent defence: exactly one certificate in the chain may carry the extension, and *that* certificate is the attested one. The regression test deliberately models a **CA-shaped** genuine leaf so the count is exercised without fix A masking it.

### Also fixed

- **A verifier bug was indistinguishable from attacker noise.** The catch-all swallowed `TypeError`s and dependency regressions into the same silent `unattested` downgrade as ordinary abuse, with no Sentry. A defect breaking Android verification for 100% of devices would have looked exactly like 100% of devices attacking. Now: typed attestation/verdict errors stay quiet, anything else is reported.
- **`evidence.rejected` never reached the database.** The route discarded it before the INSERT, which made a claim in my own code comment false as shipped. `/devices/mobile/verify` now persists it for a rejected attestation while still blanking the claimed-but-unusable case (W02's deliberate rule). Route test, mutation-checked.
- **`deviceLocked: true` was a hardcoded literal** whose test echoed the constant rather than the check. Read back from the attestation.
- **The `softwareEnforced`-duplicate branch of check 8 was unreachable in tests** — the only test aimed at it always tripped the earlier "absent from teeEnforced" check. New fixture override and a test that reaches the real branch.
- **No end-to-end test.** Both isolated suites could have been green while the canonical-SPKI key-binding comparison — the single most attacker-relevant check — never actually matched, because each side is produced by a different library. New `androidAttestationEndToEnd.test.ts` drives a real fixture chain through the real verifier through the real branch, including "genuine chain paired with a DIFFERENT registered key".
- Three checks added beyond the plan: IMPORTED keys, keys not requiring user auth, and any purpose beyond `SIGN|VERIFY` (`ATTEST_KEY` would let the key mint its own chain).
- Play Integrity: access-token cache keyed by client email (a service-account rotation previously kept using the old token for ~1h), structural PEM check at boot, constant-time `requestHash`, error name in the Sentry tag.
- Discarded ASN.1 parse errors are now carried into the failure message; an unparseable `attestationApplicationId` no longer reports as "absent".
- The "chain stopped at an intermediate" fallback the comment described was never implemented — removed rather than left as prose that overstates the code.

### Dismissed, with reasons

- **Signature-algorithm allow-list** — forging under a weak *declared* algorithm still requires Google's private key; the pinned roots are RSA-4096 / P-384 and node's OpenSSL security level already refuses SHA-1 chains.
- **Play Integrity `certificateSha256Digest` unchecked** — no configured expected signing-certificate digest exists to check against; adding one means a new env var, and this module is non-authoritative for the basis.
- **`attestationApplicationId` read from `softwareEnforced`** — flagged in the PR body as my highest-risk deviation. The crypto reviewer verified it against Android's live attestation schema and Google's reference verifier and confirmed the deviation is correct: `AuthorizationList` is one reusable ASN.1 type used for both lists, `attestationApplicationId [709]` is not restricted to the hardware-enforced instance, and Google's own verifier reads it generically. Keeping it.

### Open, NOT fixed — needs a decision

**Certificate revocation is not checked.** Google publishes a status list at `https://android.googleapis.com/attestation/status`, and their documentation is clear that a chain is not fully trustworthy without consulting it. It is a network call and deliberately outside this pure module, so a legitimately-issued-then-revoked attestation key (a leaked factory key — the scenario Remote Key Provisioning exists for) verifies here forever. Worth a follow-up issue before this feature is relied on for critical-tier approvals in production; it does not block W06.

### Verification

`vitest run src/services/attestation/ src/services/authenticatorAttestation.test.ts src/routes/authenticator.test.ts src/services/authenticatorAssurance.test.ts src/config` → **29 files, 808 passed**. `tsc --noEmit` exit 0. `eslint` on every changed file exit 0.

Every new guard was mutation-checked — disabling the CA check fails 3 tests, the extension count 1, the evidence persistence 1, the key binding 1, the StrongBox mapping 1 — so the tests discriminate rather than confirm.

**Still not verified:** no real device has ever produced a chain for this verifier, no live Play Integrity call has executed (the decode path is exercised only through an injected seam), and the integration/RLS suites were not run — this wave adds no migration, no schema change and no new table, so neither contract applies.

ToddHebebrand:
**Fix: roots inlined as TS constant (prod bundle ships no `src/`/.pem) — `2a4121b774`**

Blocking defect caught on the merge check, and it was production-only.

`androidKeyAttestation.ts` built its roots path from `import.meta.url` and `readFileSync`'d `googleHardwareAttestationRoots.pem` on first use. The runner stage of `apps/api/Dockerfile` copies only `dist`, `package.json`, `node_modules`, `packages`, `ee`, `migrations` and `assets` — **never `apps/api/src`** — so in the image the path resolved under `/app/apps/api/dist`, `readFileSync` threw ENOENT, and the scoped catch added in the review round downgraded **every** Android registration to `unattested`. Invisible in tests and in dev, where the `.pem` sits right next to the source. (Verified against the Dockerfile, not assumed.)

Roots now ship as a TS constant the bundler inlines — same fix and rationale as the Apple sibling wave, #4968. The `.pem` stays committed as the auditable provenance artifact (it is what you re-download from Google and diff); the `.ts` is generated from it and carries a "never edit by hand" note.

Two new tests, both written red first and both mutation-checked (changing one byte of the constant fails exactly these two):
- the constant is **byte-identical** to the committed `.pem`;
- SHA-256 of the set is `f2256be131ef0db817d78e1e5c1223af67e70d78827d516170a2cf7af9e36fcb` — the fingerprint recorded in the PR body.

**Bundle verified against a real `npx tsup` build**, not inferred: `dist/index.cjs` contains all five root certificates verbatim and **zero** references to the `.pem` path.

Re-run: `vitest run src/services/attestation/ authenticatorAttestation.test.ts routes/authenticator.test.ts authenticatorAssurance.test.ts src/config` → **29 files, 810 passed**. `tsc --noEmit` exit 0. `eslint` on changed files exit 0.

ToddHebebrand:
## Merged `origin/main` (Apple W03 #4968) — `d5ed906129`

PR now shows base `main`, `MERGEABLE`, no conflicts.

**Three conflicts, resolved:**

- **`apps/api/package.json`** — kept both sides' `@peculiar` dependencies. #4968 promoted `@peculiar/x509` to a **production** dependency, so my `devDependencies` entry was dropped rather than left as a duplicate.
- **`pnpm-lock.yaml`** — took main's and regenerated with `pnpm install`, not hand-resolved.
- **`services/authenticatorAttestation.ts`** — kept main's iOS branch and my Android branch, and adopted both of W03's new contract pieces:
  - the required **`publicKeyAlg`** parameter;
  - **`failureReason`** on `AttestationResult`. The Android rejection path now sets *both* `evidence.rejected` (durable on the device row — answers "why is **this** phone not L4" months later) and `failureReason` (audit log — queryable in aggregate, so a fleet-wide misconfiguration reads as a spike on one reason rather than an unexplained absence of L4-capable devices). They answer different questions, so recording both is deliberate rather than redundant.

  Also documented why the Android branch **deliberately ignores** `publicKeyAlg`, so it doesn't read as an oversight next to the iOS branch that does gate on it: on iOS the algorithm *is* the evidence (the Secure Enclave holds only P-256, so RS256 provably isn't SE-resident and drops to the weaker basis). Android has no such implication — StrongBox and TEE hold RSA keys just as happily — and the attestation states the security level directly. Gating there would reject hardware-backed keys for no security gain.

The route (`routes/authenticator.ts`) merged cleanly and already carries both conventions side by side.

**One stale assumption the merge exposed** (not a conflict): W03's test `carries NO failureReason for a platform with no verifier wired` used `platform: 'android'` as its example — true when written, false the moment W04 wired Android, so it had quietly become an assertion of the opposite of its own name. Retargeted at a platform the dispatcher really does fall through on, and added the other half of the contract (`Android ran and refused DOES carry one`).

**Verification on the merge commit:** `vitest run src/services/attestation/ src/services/authenticatorAttestation.test.ts src/routes/authenticator.test.ts src/services/authenticatorAssurance.test.ts src/config` → **30 files, 854 passed**; `tsc --noEmit` exit 0; `eslint` on every changed file exit 0.

One unrelated pre-existing test (`src/config/eeBootStopGroup.test.ts` → "kills a live process group") flaked once under the full-batch load and passed in isolation and on re-run; it spawns real processes and is untouched by this branch.

Issue #3525, opened 2026-08-14
Cancel a running script or automation


ToddHebebrand:
Reported by @tim104979 in discussion #2598: "There is also no way to stop the script from the interface when it is running."

## Confirmed

There is no cancel path for a script execution or an automation run, on any surface. Other long-running operations have one:

- `apps/api/src/routes/backup/jobs.ts:560` — `POST /jobs/:id/cancel`
- `apps/api/src/routes/backup/restore.ts:395` — `POST /restore/:id/cancel`
- `apps/api/src/routes/contracts/index.ts:16` — `/:id/cancel`

`apps/api/src/routes/scripts.ts` registers none, and neither does the automation runtime. Once a script is dispatched, the only exits are the agent returning a result or the deadline elapsing — script timeout plus `SCRIPT_GRACE_BUFFER_MS` (5 min, `commandTimeouts.ts:20`). For the reporter in #3445, whose script carries a 3600-second timeout, that is 65 minutes of a row sitting `running` with no operator recourse.

## Why it matters more than it looks

This is the operator-side half of #3445. Whatever the root cause there turns out to be, a stuck execution today has no manual remedy — a tech who can see the run is wedged still has to wait out the full deadline. A cancel control makes that class of bug survivable regardless of cause, which is worth having on its own.

## Scope

Cancel is genuinely two-sided and shouldn't be filed as UI-only:

- **Server:** mark the `script_executions` row and the `device_commands` row terminal with a distinct `cancelled` status, so it is not conflated with the reaper's `failed` — the reaper already cannot distinguish a real agent result from its own timeout (`staleCommandReaper.ts:376`), and adding a third indistinguishable case would make that worse.
- **Agent:** an in-flight process needs an actual kill, or cancel only stops the bookkeeping while the script keeps running on the endpoint — arguably worse than no button, since the UI would then assert something false.
- **Web:** a Cancel action on the execution row and in the automation run-history panel.

Wiring the endpoint without the agent-side kill would be a UI that lies. The server-side terminal state is still worth landing first, if it ships labelled as "stop tracking" rather than "stop the script".

## Not established

Whether the agent's script runner currently retains a handle it could kill a running process with, or whether that needs new plumbing. That determines whether this is small or medium, and it should be checked before scheduling.

ToddHebebrand:
[...]

PR #4971, opened 2026-09-05
fix(agent): honest, blocking, process-tree script cancellation (#3525 W04)

### LINKED ISSUE #4765: W04: Go agent: pool bypass lane, pre-start registration, blocking Cancel, PGID capture, SIGTERM→SIGKILL grace, Windows Job objects
Wave W04 of #4761 — Go agent: pool bypass lane, pre-start registration, blocking Cancel, PGID capture, SIGTERM→SIGKILL grace, Windows Job objects
Branch: `feature/4761-cancel-running-script-or-automation/wave-4765`

ToddHebebrand:
Closes #4765

Wave 04 of 6 for #3525 (feature #4761). Independent of W02/W03 — the agent change is backward-compatible in both directions: an old server ignores the new fields, and a new server treats an old agent's ack exactly as it treats `not_found`/`kill_failed`.

> **AGENT-SHIPPED: canary a small band before fleet promote.** The Windows containment change also alters the TIMEOUT path — watch for customer scripts that legitimately spawn detached children and now die with the parent.

## Why

Four correctness blockers, all verified in the agent on `main`:

1. **Cancellation could starve.** Every command went through one `workerpool`, `MaxConcurrentCommands` clamps to a floor of 1 (`config/validate.go`), and a full queue returns `"command rejected, worker pool full"`. A cancel queued behind the very script it had to stop.
2. **Cancel-before-registration.** `Execute` validated, wrote the script file and ran `configureRunAs` *before* inserting into `e.running`, and WebSocket commands are concurrent — so a cancel in that window returned "not found" and the script then started anyway.
3. **The ack was premature.** `Executor.Cancel` called `running.cancel()` and returned nil immediately; the async `cmd.Cancel` failure surfaced to `cmd.Run`, never to the cancel caller. A `cancelled: true` ack proved neither termination nor a successful kill attempt.
4. **`not_found` was not global, and Windows had no containment at all.** `killProcessGroup` on Windows was `cmd.Process.Kill()`, so a PowerShell script that started `msiexec` left it running — already true on the timeout path.

## What changed

**`Cancel(executionID, graceSeconds) (CancelOutcome, error)`** — `terminated` / `not_found` / `kill_failed`. It blocks on the execution's `done` channel (bounded by grace + a 10s hard-kill backstop) and reports `kill_failed` when `killErr != nil` **or** `!contained`. `not_found` is deliberately *not* confirmation: a restarted agent has an empty map, and it is far more often "the script finished a moment ago".

**Pre-start reservation.** `Execute` reserves the id as its very first action; `beginStart` refuses to start a process whose cancel already landed, and returns a result carrying the cancellation marker. A cancel that arrives before the process exists returns `terminated` immediately — nothing will ever run, which is a proven stop rather than a guess.

**Unix escalation ladder.** The pgid is captured at `Start` (a kill-time `Getpgid` fails once the leader is gone, and the old fallback then killed a corpse while children survived). `SIGTERM` → grace → `SIGKILL` against the **group**, polling every 100 ms for `ESRCH`. `cmd.WaitDelay` deliberately **stays 5s**: it only starts *after* the cancel callback returns and the callback has already consumed the grace, so `grace+5s` would double-count.

**Windows Job Object containment (OD4-B), fail-closed.** `CREATE_SUSPENDED` → `CreateJobObject` → `KILL_ON_JOB_CLOSE` → `AssignProcessToJobObject` → `ResumeThread`. `CREATE_SUSPENDED` is **OR**'d into the existing flags so `hideWindow`'s `CREATE_NO_WINDOW` survives. On an RD Session Host the enclosing session job denies the assignment (`sessionbroker/spawner_windows.go`, #2536); following that precedent we still run the script but leave `contained=false`, so a cancel can never claim `terminated`. No graceful phase on Windows — `GenerateConsoleCtrlEvent` needs a shared console and our children are `CREATE_NO_WINDOW`.

**Bypass lane.** `script_cancel` and `script_list_running` skip the worker pool on **both** delivery paths (the WebSocket dispatch *and* the heartbeat poll loop, which submits to the pool directly). A test asserts the bypass stays narrow — an ordinary script is still rejected by a saturated pool.

**Structured outcome + honest fan-out.** `handleScriptCancel` returns a SUCCESS result carrying `{executionId, outcome, cancelled}` for all three outcomes; only a malformed payload stays an error result, and `cancelled` is true only for `terminated`. Helper fan-out happens **only** on a local `not_found`; each helper's structured outcome is decoded rather than returning on the first `completed`; `not_found` wins only when **every** helper agrees, and an unreachable or undecodable helper grades `kill_failed` (it may still own the process, and `not_found` tells the server to revert as if nothing had run).

**Helper IPC timeout** raised from 10 (→ 15s wait) to `scriptCancelHelperTimeoutSeconds = 40` (→ 45s), so it exceeds the 30s maximum grace. At 15s a 30s cancel timed the IPC out while the helper was still escalating.

**Windows CI.** `./internal/executor` joins both explicit package lists — `limits_windows_test.go` had never executed anywhere. `./internal/heartbeat` is deliberately NOT added (known-red exclusion list, #2523), which is why the Job Object contract test lives in `internal/executor` and carries no build tag.

1. **The cancellation marker had to be plumbed onto the wire — the plan stops at `executor.ScriptResult`.** That type never leaves the agent process, but the server's closer 1 reads `result.cancelled` / `result.cancelledByCommandId` as siblings of `status`/`exitCode`. As specified, the marker died at the executor boundary and every script-result-wins-the-race case would have resolved `unconfirmed`. Third commit adds the two `omitempty` fields to `tools.CommandResult` and `websocket.CommandResult`, copies them in `toWSCommandResult`, sets them in `handleScriptInner`, and lifts them across the user-helper IPC hop. Purely additive — an API that does not yet know the fields ignores them. **W03 needs no change for this; it is the shape its own tests already assume.**
2. **The Job Object contract lives in an untagged `job.go`, not only in `job_windows.go`.** The plan asked for the portable no-op in a `//go:build !windows` file, but then `job_contract_test.go` (untagged, by design) would have exercised the no-op on Linux and asserted nothing. The launch ORDER and the degradation policy are therefore untagged; only the syscalls carry the `windows` tag.
3. **`terminateProcessTreeUnix` / `capturePgid` live in one `//go:build !windows` file** (`tree_kill_unix.go`) instead of being duplicated into `limits_unix.go` and `limits_linux.go` as the plan's wording implies. Those two files keep only their genuinely platform-specific `setProcessGroup` (Linux adds `Pdeathsig`).
4. **Added `CloseJob` to the primitive interface.** The plan did not say what happens to the job handle on a *successful* run. Closing it with `KILL_ON_JOB_CLOSE` still set would kill exactly the descendants a completed script legitimately left running, so the flag is cleared before the handle is closed — mirroring `software_install_process_tree_windows.go`'s `release()`. While the script runs, the flag stays set as the agent-crash backstop.
5. **`launchContained` takes the `*runningExecution` as a parameter** rather than allocating and returning one, since `Execute` already holds a reserved entry.
6. **`TestCancelNonexistentExecution` (runas_test.go) and half of `TestHandleScriptCancel`** asserted the old "unknown id is an error" contract and were removed, replaced by outcome-based tests with a comment pointing at them.
7. **A cancelled result reports `ExitCode -1` / `Error: "execution cancelled"`**, checked before the `*exec.ExitError` branch — a killed shell reports `signal: killed`, which would otherwise be persisted as a normal non-zero exit.

## Known accepted residual (not fixed here, by design)

`userhelper.NewWithOptions` rebuilds `executor.New(nil)` on **every** reconnect (`client.go`, `supervisor.go`), so a `runAs=user` script that survived an IPC reconnect is structurally unreachable and will correctly report `not_found` → the server records `unconfirmed`. Degraded but honest. Fixing helper executor persistence across reconnects is out of scope for #3525 and should be filed separately.

## Verification

- `cd agent && go test -race ./internal/executor/ ./internal/heartbeat/ ./internal/userhelper/ ./internal/websocket/ ./internal/remote/tools/` — all pass.
- `cd agent && go test ./...` (whole agent module) — all pass.
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go vet` clean on every touched package; `go build ./...` clean for darwin, windows/amd64 and linux/amd64.
- `gofmt -l` clean on every touched file.
- **Both new red-first controls were proven to discriminate**, not just asserted: reverting `terminateProcessTree` to the old leader-only `cmd.Process.Kill()` fails `TestCancelKillsTheWholeProcessTree` (`3 -> 11 bytes` of surviving grandchild heartbeat), and stubbing the pgid to 0 fails `TestPgidIsCapturedAtStartNotAtKillTime`.

## NOT verified

- **No Windows box was available.** The Job Object path is covered only by the fake-backed contract test plus a clean `GOOS=windows` vet and test-binary cross-compile. `CREATE_SUSPENDED` + `AssignProcessToJobObject` + resume-by-thread-enumeration (os/exec closes the primary thread handle, so threads are re-opened via `CreateToolhelp32Snapshot`) has **never been executed on a real Windows host in this branch**. The RDS assignment-denial degradation path likewise.
- **`./internal/executor` has never run in the Windows CI job before this PR.** A static audit of every test in the package found no Windows-hostile assertion (everything either self-skips on `runtime.GOOS`/`LookPath` or is pure logic), and the test binary cross-compiles — but the first real evidence is this PR's own Windows job.
- No integration/E2E coverage of the server↔agent round trip: W02/W03 are not merged, so nothing on `main` yet sends a `script_cancel` command.
- macOS/Linux tests requiring `/bin/bash` self-skip where it is absent; they did run here.

https://claude.ai/code/session_01XYjgvDNw7Qk8SFaBX6kdEx

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1361,7 +1361,7 @@ jobs:
         working-directory: agent
         env:
           CGO_ENABLED: "0"
-        run: go test ./internal/sessionbroker ./internal/eventlog ./internal/elevaccount ./internal/ipc ./internal/logging ./internal/state ./internal/watchdog ./internal/backup/vss ./cmd/breeze-watchdog
+        run: go test ./internal/sessionbroker ./internal/eventlog ./internal/elevaccount ./internal/ipc ./internal/logging ./internal/state ./internal/watchdog ./internal/backup/vss ./internal/executor ./cmd/breeze-watchdog
 
       # The step above runs without -v, so a t.Skip is indistinguishable from a
       # pass in its output. That matters here more than usual: elevaccount's two
@@ -1467,7 +1467,7 @@ jobs:
       # the one that most needs the detector on the OS it actually ships to.
       - name: Run Windows race tests
         working-directory: agent
-        run: go test -race ./internal/sessionbroker ./internal/eventlog ./internal/ipc ./internal/state ./internal/watchdog ./cmd/breeze-watchdog
+        run: go test -race ./internal/sessionbroker ./internal/eventlog ./internal/ipc ./internal/state ./internal/watchdog ./internal/executor ./cmd/breeze-watchdog
 
   # ─── Go agent lint ────────────────────────────────────────────────────
   # The `lint` job above is JS/TS + compose guards only; the Go agent was
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1357,6 +1357,16 @@ jobs:
       # remote/desktop and no cgo. Its two behavioural tests assert against the
       # real netapi32 and skip loudly, rather than pass, on a host where
       # ~breeze_elev exists or local-account queries are not permitted.
+      # ./internal/executor is the fifth instance of the same gap (#3525):
+      # limits_windows.go's setProcessGroup/hideWindow and the new Job Object
+      # containment are Windows-only, and limits_windows_test.go had never
+      # executed anywhere. Verified like logging/ipc/elevaccount: with
+      # GOOS=windows GOARCH=amd64 CGO_ENABLED=0 it vets and its test binary
+      # compiles. It differs from those three in one way that matters — its
+      # Windows dependency graph DOES contain remote/desktop, via
+      # privilege -> remote/tools — which is harmless here (CGO_ENABLED=0 is
+      # exactly the configuration remote/desktop's !cgo files are written for)
+      # but is why it must stay out of the race step below.
       - name: Run Windows Go tests
         working-directory: agent
         env:
@@ -1460,14 +1470,25 @@ jobs:
           if ($passes -ne 14) { throw "expected exactly 14 top-level PASS lines, got $passes - a test in the -run filter was renamed, removed, or added without updating this count" }
 
       # -race requires cgo, which remote/desktop cannot satisfy on Windows, so
-      # this covers only packages that do not import it (heartbeat and agentapp
-      # do, transitively). Same package set as the step above: sessionbroker is
-      # the concurrency-critical one — session-keyed helper admission, the
-      # lifecycle registry, and Job Object ownership all live there — so it is
-      # the one that most needs the detector on the OS it actually ships to.
+      # this covers only packages that do not import it (heartbeat, agentapp and
+      # executor do, transitively). Same package set as the step above MINUS
+      # those: sessionbroker is the concurrency-critical one — session-keyed
+      # helper admission, the lifecycle registry, and Job Object ownership all
+      # live there — so it is the one that most needs the detector on the OS it
+      # actually ships to.
+      #
+      # ./internal/executor belongs to the CGO_ENABLED=0 step above and NOT
+      # here: it reaches remote/desktop via privilege -> remote/tools (#3525
+      # added the package to the list above and briefly to this one, which went
+      # red). The concrete failure is that dxgi_windows.go and
+      # dxgi_capture_windows.go are `//go:build windows && !cgo`, so with cgo ON
+      # their comVtblFn/dxgiCapturer/procDeleteObject definitions vanish while
+      # the plain-`windows` files that reference them stay. Check the import
+      # graph before adding anything else here:
+      #   GOOS=windows go list -deps ./internal/<pkg> | grep remote/desktop
       - name: Run Windows race tests
         working-directory: agent
-        run: go test -race ./internal/sessionbroker ./internal/eventlog ./internal/ipc ./internal/state ./internal/watchdog ./internal/executor ./cmd/breeze-watchdog
+        run: go test -race ./internal/sessionbroker ./internal/eventlog ./internal/ipc ./internal/state ./internal/watchdog ./cmd/breeze-watchdog
 
   # ─── Go agent lint ────────────────────────────────────────────────────
   # The `lint` job above is JS/TS + compose guards only; the Go agent was
--- a/agent/internal/executor/cancel_test.go
+++ b/agent/internal/executor/cancel_test.go
@@ -1,8 +1,10 @@
 package executor
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"testing"
@@ -68,7 +70,7 @@ func TestCancelBlocksUntilTheProcessIsGone(t *testing.T) {
 	go func() { r, _ := e.Execute(sleepScript("id-1", 60)); done <- r }()
 	waitForRunning(t, e, "id-1")
 
-	outcome, err := e.Cancel("id-1", 0)
+	outcome, err := e.Cancel("id-1", "cc-1", 0)
 	if err != nil {
 		t.Fatalf("Cancel: %v", err)
 	}
@@ -86,7 +88,7 @@ func TestCancelBlocksUntilTheProcessIsGone(t *testing.T) {
 
 func TestCancelReportsNotFoundForAnUnknownID(t *testing.T) {
 	e := newTestExecutor()
-	outcome, err := e.Cancel("nope", 5)
+	outcome, err := e.Cancel("nope", "cc-nope", 5)
 	if err != nil {
 		t.Fatalf("Cancel: %v", err)
 	}
@@ -101,8 +103,7 @@ func TestCancelledResultCarriesTheCancellationMarker(t *testing.T) {
 	done := make(chan *ScriptResult, 1)
 	go func() { r, _ := e.Execute(sleepScript("id-2", 60)); done <- r }()
 	waitForRunning(t, e, "id-2")
-	e.SetCancelCommandID("id-2", "cancel-cmd-7")
-	if _, err := e.Cancel("id-2", 0); err != nil {
+	if _, err := e.Cancel("id-2", "cancel-cmd-7", 0); err != nil {
 		t.Fatal(err)
 	}
 
@@ -130,7 +131,7 @@ func TestGracefulEscalationSendsSIGTERMBeforeSIGKILL(t *testing.T) {
 	go func() { defer close(done); _, _ = e.Execute(sigtermTrapScript("id-3", marker, 60)) }()
 	waitForRunning(t, e, "id-3")
 
-	outcome, err := e.Cancel("id-3", 5)
+	outcome, err := e.Cancel("id-3", "cc-3", 5)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -151,7 +152,7 @@ func TestZeroGraceSkipsStraightToKill(t *testing.T) {
 	go func() { defer close(done); _, _ = e.Execute(sigtermTrapScript("id-5", marker, 60)) }()
 	waitForRunning(t, e, "id-5")
 
-	if _, err := e.Cancel("id-5", 0); err != nil {
+	if _, err := e.Cancel("id-5", "cc-5", 0); err != nil {
 		t.Fatal(err)
 	}
 	<-done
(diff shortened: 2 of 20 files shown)


ToddHebebrand:
## Review summary

Four specialist reviewers ran against this PR (code / tests / silent-failure / comment-accuracy), all read-only against `gh pr diff`. **9 findings acted on, 1 dismissed, 0 outstanding.** Everything below is fixed in `adfd5fe`; every fix has a control that was watched fail before the fix landed.

### Critical — a cancel could lie, or a script could escape one

| # | Finding | Reported by | Fix |
|---|---|---|---|
| 1 | **A cancel for a script still QUEUED in the worker pool was silently lost.** The bypass lane gets the *cancel* past the pool, but its target can still be sitting in that queue behind another script — with the pool flooring at one worker that is the ordinary case, not an edge. `Execute` had not run, so `Cancel` answered `not_found` and forgot; the pool later dequeued the script and ran it to completion as if no cancel had ever been issued. This is the same class as the PR's own blocker #1, relocated. | code-reviewer + silent-failure-hunter (independently) | `Cancel` leaves a refusal record that the eventual `Execute` adopts and bails out on, carrying the cancellation marker so the server closes the row honestly. The **answer stays `not_found`** deliberately: nothing there can tell a queued script from an id this device never had, and claiming `terminated` would forge a confirmed kill. Reaped after 2h; excluded from `ListRunning` so `script_list_running` invents no phantoms. |
| 2 | **A duplicate `Execute` for a running id adopted the first one's reservation** and closed the shared `done` channel when *it* finished — unblocking a `Cancel` waiting on the ORIGINAL process and letting it report `terminated` while that process ran on. | found by me pre-review; test-analyzer independently flagged the adjacent test as vacuous | Duplicates refused. (code-reviewer correctly noted `markCommandSeen` makes this unreachable from the heartbeat path — kept as defense in depth, since the user helper is a second caller.) |
| 3 | **Unix TOCTOU.** `os/exec` starts its context watchdog inside `cmd.Start`, so a cancel can fire between `Start` returning and `attachProcessGroup`. That kill reaches the leader only, and `attachProcessGroup` then set `contained=true` a moment later — letting the cancel that already ran grade `terminated` while group members survived. | code-reviewer | Containment latches off once a kill has run without a group. |
| 4 | **Windows: a `ResumeThread` failure stranded a suspended orphan.** The process was created, contained and suspended; `Execute` then called `releaseContainment`, which clears `KILL_ON_JOB_CLOSE` and drops the only handle that could ever kill it — and `release()` had already forgotten the id, so no later `Cancel` could find it. | silent-failure-hunter + code-reviewer (independently) | The failed-start path now terminates the job instead of releasing it. |

### Important

| # | Finding | Reported by | Fix |
|---|---|---|---|
| 5 | `capturePgid` returned 0 when `Getpgid` raced the leader's exit, marking a perfectly contained execution uncontained and downgrading every later cancel to `kill_failed`. Confirmed as a real flake source in a loaded full-package `-race` run. | found while diagnosing that flake | `setProcessGroup` asks for `Setpgid` with `Pgid: 0`, so pgid == pid by construction; `Getpgid` is the confirmation, not the source of truth. |
| 6 | **`CloseJob`'s clear-then-close ordering had no test at any level**, and it fires on *every successful completion* — a regression would kill customer-spawned background processes on ordinary green runs, fleet-wide, silently. | pr-test-analyzer | The ordering moved out of `job_windows.go` into the portable `releaseContainment`, where the fake-backed contract test asserts it (plus the leak-rather-than-kill fallback). |
| 7 | **The fail-closed rule was proven only on Unix.** The `kill_failed` test skips on Windows, and `TestAssignmentDenialFailsClosed` stopped at `contained=false` without ever composing it through to the outcome. | pr-test-analyzer | `gradeCancelOutcome` extracted and table-tested over all six states, so the rule is asserted on every platform — the RDS denial it exists for cannot be reproduced anywhere. |
| 8 | Untested branches: the `Cancel` give-up timer (grace + backstop — every other `kill_failed` verdict arrived via `!contained`, so the arithmetic was unverified), and the uncontained terminate fallback. | pr-test-analyzer | Both covered. |
| 9 | **The user helper's own structured cancel outcome was asserted only as "did not error"** — it would have passed with `outcome` always `""` or `cancelled` hardcoded `true` (the old behaviour). | pr-test-analyzer | The helper test now decodes and asserts `outcome`/`cancelled`/`executionId`. |
| 10 | `saturatePool` slept a fixed 20 ms and hoped — a plausible CI-contention flake, and a false *pass* if the pool was never actually saturated. | pr-test-analyzer | Waits for a real pick-up signal from each worker, then proves the next submit is rejected. |
| 11 | A comment claimed "we must never touch `cmd.Process` from Cancel" — false as written: the uncontained fallbacks do exactly that, from inside the callback, and safely. It also contradicted the correct rationale given 40 lines later. | comment-analyzer | Corrected. |
| 12 | `TestHelperNotFoundOnlyWinsWhenEveryHelperAgrees` asserted only `!= "not_found"`. | pr-test-analyzer | Tightened to the specific `kill_failed` the ranking must produce. |

Also from review: `Cancel` now takes the cancel command id directly, so the marker is stamped in the same critical section as the request. The separate `SetCancelCommandID` setter is gone — it could not work for a not-yet-dispatched script, which is what finding #1 needs.

### Dismissed

- **pr-test-analyzer, "the cancel-before-registration test is vacuous — nothing reaches `beginStart`."** Correct as reported, and already fixed before the review landed: that test now drives `reserve` → `Cancel` → `beginStart` directly and asserts the gate refuses. Confirmed by the same reviewer's own reasoning about the `preCancelled` early return, which no longer exists.

### Verified clean (checked and cleared by reviewers)

`cmd.Cancel`/`WaitDelay` semantics after the `Run` → `Start`/`Wait` split; `normalizeSignalErr`'s ESRCH folding (cannot produce a false `terminated` — the `pgid<=0` branch is independently gated by `contained`); the `launchContained` degradation ladder; `decodeHelperCancelOutcome`'s grading (never more optimistic than warranted); the marker plumbing across every hop; and every in-repo cross-reference in the comments (`config/validate.go`'s floor of 1, `sessionbroker/spawner_windows.go`'s RDS denial, `software_install_process_tree_windows.go`'s `release()` precedent, the `#2523` Windows exclusion list, and the helper-timeout arithmetic) — all confirmed accurate against the actual files.

- `go test -race ./internal/executor/ ./internal/heartbeat/ ./internal/userhelper/ ./internal/websocket/ ./internal/remote/tools/` — pass. Executor repeated `-count=5`, the heartbeat cancel tests `-count=10`, both clean.
- `go test ./...` (whole agent module) — pass.
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go vet` clean; `go build ./...` clean for darwin, windows/amd64, linux/amd64. `gofmt -l` clean.
- **Controls run for the new guards** (each watched fail, then restored): removing the refusal record → the queued script runs anyway; removing the containment latch → containment re-established after a leader-only kill; the two originals from the PR body still discriminate.

### Still not verified (unchanged from the PR body)

No Windows host was available, so every native Win32 call remains covered only by the fake-backed contract test plus a clean `GOOS=windows` vet and test-binary cross-compile. `./internal/executor` has never run in the Windows CI job before this PR; a static audit found no Windows-hostile test in the package, but this PR's own Windows job is the first real evidence. No server↔agent round trip exists yet — W02/W03 are not merged.

**Canary a small band before fleet promote.** The Windows containment change also alters the TIMEOUT path.

ToddHebebrand:
## CI fix — `c0f0799`

Both head-SHA failures were from this branch. Fixed properly, no nolint, no revert of the feature.

**1. Lint Agent — `job.go: field native is unused`.** `jobHandle` and `suspendedProcess` each carried a `native any` escape hatch. `suspendedProcess.native` was never read or written at all; `jobHandle.native` only round-tripped a `windows.Handle`, which is *defined as* a `uintptr` — so the plain `handle uintptr` converts back losslessly. Both fields removed and `nativeJobHandle` simplified. Verified with the exact CI linter (`golangci-lint v2.12.2`, `--new-from-rev=origin/main`, `CGO_ENABLED=0`): **0 issues**, and re-adding the field reproduces `job.go:40:2: field native is unused (unused)` exactly.

**2. Test Agent (Windows) — `internal/executor [build failed]`.** The `remote/desktop` import is transitive, legitimate and **pre-existing**: `executor → privilege → remote/tools → remote/desktop`, and `executor.go` has imported `privilege` since #3409 PR4b. My change added no imports.

The break is the CGO split, not the import: `dxgi_windows.go` and `dxgi_capture_windows.go` are `//go:build windows && !cgo`, so with cgo **on** their `comVtblFn` / `dxgiCapturer` / `procDeleteObject` definitions vanish while the plain-`windows` files referencing them remain. `-race` requires cgo — so the failure was the **race step**, not the `CGO_ENABLED=0` step.

That makes the fix narrower than "drop the two lines" as I'd guessed in the PR body. `./internal/executor` belongs in the `CGO_ENABLED=0` step (its test binary cross-compiles there — that is precisely the configuration `remote/desktop`'s `!cgo` files are written for) and **not** in the race step, whose own comment already states the rule. Confirmed executor was the only package in that list violating it:

```
GOOS=windows go list -deps ./internal/<pkg> | grep remote/desktop
  sessionbroker 0 · eventlog 0 · ipc 0 · state 0 · watchdog 0 · executor 1
```

**Removed from the race step only**, so `limits_windows.go` and the Job Object contract keep the Windows coverage this PR adds. Both step comments now record the constraint and the command to check it before adding another package.

**3. Also fixed: a flake this branch introduced**, surfaced by the verification run rather than by CI. `TestGracefulEscalationSendsSIGTERMBeforeSIGKILL` waited only for `cmd.Start` to succeed — well before bash has parsed and run the `trap` builtin, so a SIGTERM landing in that window took the default action and wrote no marker. Worse, it made the zero-grace twin pass for the *wrong* reason (marker absent because the trap wasn't installed, not because SIGKILL beat it). The script now signals readiness after installing the trap and both tests gate on it.

### Verification

- `gofmt -l` clean on every file this PR touches; `go vet ./...` clean; `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go vet ./internal/executor/ ./internal/heartbeat/` clean.
- `go test -race ./internal/executor/ ./internal/heartbeat/ ./internal/userhelper/ ./internal/websocket/` — pass; executor also `-count=3`.
- The two trap tests **20× under `-race`, 40/40**; and the zero-grace test still fails when grace is raised to 5, so it is still discriminating.
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/executor` builds — the step that keeps it.

### One unrelated red on this host, not from this PR

`internal/monitoring` fails `TestMonitorRunsImmediateCheckOnStart` and `TestApplyConfigRestopsAndRestarts` locally, alongside host warnings `failed to read process name during kill scan … input/output error`. `go list -deps ./internal/monitoring` contains **none** of the packages this PR changes, and the same suite passed on this host earlier today — environmental. Flagging rather than silently ignoring; if CI shows it red on `main` too, it needs its own issue.

Issue #4768, opened 2026-09-03
Import custom fields when migrating from another RMM (#3257)


ToddHebebrand:
Tracking issue for **Import custom fields when migrating from another RMM (#3257)** (`rmm-custom-field-import`).

- Plan: `docs/superpowers/plans/onboarding-signup/2026-09-02-rmm-custom-field-import.md`
- Spec: `docs/superpowers/specs/onboarding-signup/2026-09-02-rmm-custom-field-import-spec.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All 10 waves merged (W01 #4791, W02 #4991, W03 #5028, W04 #4917, W05 #5044, W06 #4973, W07 #5043, W08 #5055, W09 #5060, W10 #5061). Unreleased.

**Pre-release / not verified**
- E2E `e2e-tests/tests/custom-field-import.spec.ts` was never run against a live stack (worktree Docker network-pool overlap); run it in the pre-release sweep. The ambiguous-row candidate-pick sub-flow is not covered — needs a second same-hostname fixture in `seedE2eFixtures.ts`.
- No browser walk of the wizard.
- W05 backfill row count on prod is unmeasured; the migration `2026-10-11-160000-device-custom-field-values.sql` prints RAISE WARNING counts by minted type at deploy — capture them from Postgres logs on both regions.
- Prod preflights for W02 and W03 were run read-only on 2026-09-05 (both regions: 0 duplicates, 0 ownerless, 0 shadowed). Re-run before deploy if any definitions were created since.
- Performance at the 1000-row / 5000-value cap and lock contention on the (device_id, definition_id) upsert untested.
- NinjaOne / N-central device-id field names in the vendor guides not re-verified against live vendor APIs.

**Behaviour changes to sanity-check before release**
- Cross-partner device moves (system scope only) now DROP values whose partner-wide definitions are not visible in the target org; counted in the move audit (`details.customFieldValues.dropped`).
- Custom-field values are now `included` in the GDPR tenant export (the jsonb they replaced was `excludedOpen`). A secret stashed in a custom field will appear in the export.
- Script write-back `device_not_found` narrowed: device deleted mid-write now raises 23503 (Sentry-captured) instead of a typed per-field rejection.
- W08's warranty target writes `device_warranty.status` and never overrides `data_source='provider'` unless `overrideProvider`.

**Follow-ups**
[...]

PR #4973, opened 2026-09-05
feat(api): device_external_links + serial/hostname indexes + device resolver (#3257 W06)

### LINKED ISSUE #4774: W06: device_external_links + serial/hostname indexes + resolver
Wave W06 of #4768 — device_external_links + serial/hostname indexes + resolver
Branch: `feature/4768-rmm-custom-field-import/wave-4774`

ToddHebebrand:
Closes #4774

Wave 06 of #4768 (RMM custom-field import, plan `docs/superpowers/plans/onboarding-signup/2026-09-02-rmm-custom-field-import.md`). Independent of W02–W05; nothing here is mounted on a route yet — W08 consumes it.

## What changed

**1. `device_external_links` — durable device identity for re-import**

`apps/api/migrations/2026-10-10-100000-device-external-links.sql` (sorts after the newest committed migration, `2026-10-09-000600-*`; `check-migration-naming.sh --against-ref origin/main` passes) plus `apps/api/src/db/schema/deviceExternalLinks.ts`.

Every migration source exports a stable device UID (Datto `uid`, Ninja device id, Automate `ComputerID`, N-central `applianceID`). Recording it on the first successful match turns a fuzzy hostname join into an exact lookup on every subsequent run, which is what makes a multi-day migration (import → enroll more machines → re-import) work at all.

- Unique key `(partner_id, system, COALESCE(source_instance, ''), external_id)` — **partner**-scoped like `organization_external_links`, not org-scoped like `contact_external_links`: a Datto UID is unique across the Datto tenant, which is partner-shaped, and a partner key survives `moveOrg` untouched. `source_instance` ships nullable and unused but is in the index from day one (spec Open Decision 2), so adopting an account/instance discriminator later is a backfill, not a migration of a live unique key.
- `(device_id, org_id) → devices(id, org_id)` `ON UPDATE CASCADE ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED` (matches `device_mtls_certificates`); `(org_id, partner_id) → organizations(id, partner_id)` `ON DELETE CASCADE DEFERRABLE INITIALLY IMMEDIATE`.
- Tenancy shape 1: RLS enabled **and** forced, four `breeze_has_org_access(org_id)` policies, `GRANT` to `breeze_app`. No `rls-coverage` allowlist entry needed (shape 1 is auto-discovered).
- Two resolution indexes that did not exist: `device_hardware_org_serial_idx` on `(org_id, upper(btrim(serial_number)))` (device_hardware had nothing beyond its PK) and `devices_org_hostname_lower_idx` on `(org_id, lower(hostname))`. Both expression-indexed on the **normalised** form, because the resolver compares that form on both sides of the join.
- No writes in the file, so no `breeze.scope = system` elevation is needed.

**2. The agent's junk-serial denylist, ported verbatim**

`apps/api/src/services/customFields/import/deviceIdentity.ts` — `JUNK_HARDWARE_IDENTITY_VALUES`, `isJunkHardwareIdentity`, `normalizeSerial`, `normalizeHostname`. The exact 19-value list from `cleanHardwareIdentityValue` (`agent/internal/collectors/hardware.go:156-190`) plus the `"to be filled by"` substring rule, with the same normalisation (lowercase, collapse `strings.Fields`, trim dots). It is an exact list, **not** an all-zeros pattern — a zero run of a different length passes through in Go and here.

This matters because the agent applies the denylist **only on Windows** (`hardware_windows.go:136`); Linux reads raw DMI and macOS shells out to `system_profiler`, and both write the raw value through. Those values are already in `device_hardware.serial_number`, so they are filtered on **both** sides of the import join — otherwise every Linux box reporting `Default string` collapses into one giant ambiguous group.

**3. The device resolver**

`types.ts` (dependency-free wire types + `IMPORT_SYSTEMS`/`DEFAULT_IMPORT_SYSTEM`) and `resolveDevice.ts`.

Resolution order, always within one organization: `deviceId` → `(externalSystem, externalId)` via the link table → `serialNumber` → `hostname` (case-insensitive).

- **Every** supplied identifier is resolved, not just the first. Two identifiers that each pin a single, *different* device ⇒ `identity-conflict`, refused, with `conflictingMethods`. "First hit wins" would silently pick one and discard the evidence that the row is wrong.
- An identifier matching *several* devices is a question, not a veto: only single-match identifiers participate in the conflict check, so a hostname collision cannot poison an exact serial.
- 0 matches ⇒ `not-found`; >1 ⇒ `ambiguous` with ranked candidates and `deviceId: null`. **Never auto-picks.**
- Ranking copies enrollment's collision priority chain (`routes/agents/enrollment.ts:514-518` — online, then any live row, then oldest) and is **presentational only**. Each candidate carries serial, OS, enrolment date and last-seen so the operator picks on evidence.
- `loadDeviceResolutionSnapshot` runs under `runOutsideDbContext(() => withSystemDbAccessContext(...))`, where `breeze_has_org_access` returns TRUE — so RLS is **not** the boundary. The partner, org and site predicates carried in the SQL are the whole of it. `getDeviceWithOrgCheck` is deliberately not reused: it selects with no org predicate and checks in JS, which is safe under RLS and no check at all here.
- A row naming an org outside the caller's reach resolves `org-not-found` — the same annotation an unknown org gets, so the response is never an existence oracle for another partner's tenants.
- An empty `accessibleOrgIds`/`allowedSiteIds` array reaches nothing rather than degrading into "no filter".

**4. Registration lists (5 of 6 apply)**

| List | Entry |
|---|---|
| `CORE_ORG_CASCADE_DELETE_ORDER` | `device_external_links`, between `device_event_logs` and `device_filesystem_cleanup_runs` |
| `CORE_DEVICE_CASCADE_DELETE_TABLES` | added to the core device block (leaf table) |
| `CORE_DEVICE_ORG_DENORMALIZED_TABLES` | added |
| `CORE_TENANT_EXPORT_POLICY` | all 11 columns `included` — `system`/`external_id` are tenant identifiers, the same classification the other two `*_external_links` tables carry; no json/jsonb/bytea column exists |
| `orgMergeRegistry` `REPOINT_TABLES` | plain repoint, **with the unique key checked first**: `device_external_links_uniq` is partner-scoped with no `org_id` in it, and a merge keeps both orgs under one partner, so any pair that would collide after the merge was already forbidden before it. Reasoning is in the registry note. |
| `rls-coverage` allowlist | **none** — shape 1 is auto-discovered |

1. **`devices_org_hostname_lower_idx` drops the plan's `WHERE hostname IS NOT NULL`.** `devices.hostname` is `NOT NULL`, so the predicate is always true; a partial index with an always-true predicate is noise. The `device_hardware` one keeps its predicate — `serial_number` really is nullable.
2. **The resolver does not emit an explicit `NOT IN (junk list)` in the serial SQL.** The IN-list is built from `normalizeSerial`, which returns `null` for every denylisted value, so no junk string can reach the query and a device storing `Default string` can never join on it. An explicit `NOT IN` would be a provable no-op paid for on every lookup. The database side is still covered end-to-end by an integration test (two devices both storing `To Be Filled By O.E.M.` ⇒ `not-found`, not a 2-way group), and `buildDeviceResolutionSnapshot` refuses to index junk regardless of who hands it rows.
3. **The plan's first resolver test is internally contradictory and was rewritten.** It passed `{ deviceId: D1, serialNumber: 'S-D2', hostname: 'wkstn-d3' }` and expected `matched`/`id`, while its own trailing comment said "…but only when they AGREE" and the wave's identity-conflict rule says the opposite. Precedence is now proven with *agreeing* identifiers, and disagreement is proven separately (`id` vs `serial`, `link` vs `hostname`, `serial` vs `hostname`).
4. **`types.ts` carries two frozen literal arrays** (`IMPORT_SYSTEMS`, `DEFAULT_IMPORT_SYSTEM`) as the plan's own interface block specifies, alongside the types. The constraint that matters — the module imports nothing, so W07/W08 cannot get an import cycle through it — holds.
5. **Snapshot loading is split into `loadDeviceResolutionSnapshot` (SQL boundary) + `buildDeviceResolutionSnapshot` (indexing).** The plan named only the former. The split is what lets the resolution rules be unit-tested without a database while the boundary stays proven against real Postgres.
6. **Integration fixtures seed through the test client, outside any system context.** The partner-export statement triggers enforce a partners-shared-before-orgs-exclusive lock hierarchy; seeding two partners inside one transaction raises `partner export lock hierarchy violation`. Autocommitted seed statements sidestep it. The code under test still opens its own system context, which is the point.

## Verification

Against a live Postgres (`pnpm test-stack`), all in the foreground:

- `npx tsx scripts/migrate.ts` then `npx tsx scripts/check-drift.ts` → **no drift, 675 migration files match the ledger**
- `npx vitest run src/services/customFields/import/ src/routes/devices/cascadeDelete.test.ts src/routes/devices/moveOrg.coverage.test.ts` → **142 passed**
- `npx vitest run src/db/autoMigrate.test.ts src/db/migrationRlsScope.test.ts src/config/composeBindMounts.test.ts` → **103 passed**
- `npx vitest run --config vitest.integration.config.ts` on `deviceExternalLinks`, `deviceCustomFieldImportIsolation`, `tenantCascade`, `orgLifecycleFoundations` → **33 passed**
- `npx vitest run --config vitest.integration.config.ts` on `tenant-export-policy`, `tenantExportErasureRoundtrip` → **6 passed**
- `npx vitest run --config vitest.config.rls-coverage.ts` → **100 passed**
- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` → clean (the default heap OOMs on this package)
- `pnpm lint` → clean
- `scripts/check-migration-naming.sh --against-ref origin/main` → OK

**Non-vacuity of the isolation proof was demonstrated, not assumed.** With the partner predicate removed the "partner-wide caller still cannot see another partner" test fails; with the site predicate removed the "site-restricted org user" test fails. Both mutations were applied, observed red, and reverted.

## Not verified

- **No route mounts this.** `loadDeviceResolutionSnapshot`/`resolveDeviceRow` have no production caller until W08; the request-path wiring (auth scope, `requireMfa`, the 1000-row cap, the `DeviceResolutionScope` a real request produces) is W08's to prove.
- **No production-scale query plan was measured.** The two new indexes are asserted to exist and to be expression-indexed on the normalised form; nobody ran `EXPLAIN` against a fleet-sized `device_hardware`. The `OR` of three identifier arms is expected to be a BitmapOr but was not verified on real data.
- **`source_instance` is unexercised beyond the unique-key tests.** No caller writes a non-null value yet, by design.
- **No link rows are ever written by this PR.** Recording a link on first match is W08's job; this wave only makes the lookup possible.
- **Full `apps/api` unit suite and the other integration shards were not run** — only the suites listed above plus the contract suites the registration lists gate.
- **`device_hardware.serial_number` is `varchar(100)`; `btrim` yields `text`.** The index and the query both use the same expression, and the integration test resolves a real serial end to end, so they agree in practice; no separate collation/type audit was done.

https://claude.ai/code/session_01Fc1UFWDowWtx9mRwmAmEm4


ToddHebebrand:
## PR review summary

Five reviewers ran in parallel against this PR (`/pr-review-toolkit:review-pr`): code review, test coverage, silent-failure hunt, comment accuracy, and type design. **13 findings acted on, 2 dismissed with reasoning, 0 outstanding.** Everything below is fixed in c21f11d125.

### Critical / important — fixed

| # | Finding | Fix |
|---|---|---|
| 1 | **`org-not-found` reported for a reachable org.** *(found while acting on the reviews, not by a reviewer)* `loadDeviceResolutionSnapshot` returned a fully empty snapshot from two branches that ran **after** it already knew the caller's reachable orgs. A row naming a perfectly visible org whose identifier simply did not match then reported `org-not-found` — wrong, and the opposite of what the non-disclosure rule is for. | Those branches now return a snapshot carrying the real reachable-org set. Two integration tests pin it; **both were red before the fix.** |
| 2 | **`linkKey` separator collision** (code reviewer, and independently the silent-failure hunter). `system`, `source_instance` and `external_id` are free-form text; joining them on an ordinary character is not injective — `('csv','','a b')` and `('csv',' a','b')` are two distinct rows under `device_external_links_uniq` but collapse to one key, folding an unrelated device into the link axis. | Joined on `U+0000` — the same separator and the same reason `services/contacts/import.ts` uses, since Postgres cannot store a NUL in `text`. Written as a `` **escape**: the literal byte was making the file read as binary to `grep` and to `gh pr diff`. Regression test added and mutation-checked (space separator ⇒ red). |
| 3 | **`?? null` folded `undefined` into "unrestricted"** (silent-failure hunter *and* type-design reviewer, independently). A route handler building `DeviceResolutionScope` from auth state and dropping a field would have been granted the whole partner. The site axis matters most — RLS never covered it, so this is its only enforcement. | `requireReach()` throws on `undefined` instead of coalescing. A caller bug now fails loudly instead of widening access. |
| 4 | **Zero reachable orgs was a silent, undiagnosable no-op** (silent-failure hunter). "Partner genuinely has no orgs", "`partnerId` is wrong" and "caller named someone else's orgs" all produced an identical wall of `not-found`. | `console.warn` with the partner, named-org count and row count, so the three are distinguishable in telemetry. |
| 5 | **"Serial absent" and "serial supplied but junk" were indistinguishable** (silent-failure hunter). A whole CSV column mapped to the wrong field looked exactly like a fleet that has no serials. | New `discardedIdentifiers` on `DeviceResolution`, populated when a supplied value normalises away. W08's preview can surface it. |
| 6 | **`DeviceResolution` let illegal states type-check** (type-design reviewer). Nothing stopped W07/W08 minting `{ outcome: 'ambiguous', deviceId }` — exactly the "silently picked one" failure this wave exists to prevent. | Now a discriminated union on `outcome`. Still plain JSON, so the HTTP boundary is unchanged. |
| 7 | **`EMPTY_SNAPSHOT` was a shared mutable singleton** (type-design reviewer). `Readonly*` is erased at runtime. | Replaced with an `emptySnapshot()` factory. |

### Comment accuracy — three inaccurate citations, all corrected

The comment reviewer opened every cited file. Three claims were wrong:

- `deviceExternalLinks.ts` said the `COALESCE` unique index is *"expression-based, which Drizzle cannot model"* — **false**, and the file contradicted itself two lines down where it does model it. The real reason the `.sql` is authoritative is that this repo does not drive migrations from Drizzle at all (`scripts/check-drift.ts`).
- `resolveDevice.ts` cited `enrollment.ts:514-518` for the ranking chain; the chain is **517-521** (514-516 is an unrelated comment tail).
- `deviceIdentity.ts` said the agent applies the denylist *"only on Windows"* — too broad. Linux **does** call `cleanHardwareIdentityValue`, for the motherboard fields. It is `SerialNumber` **specifically** that goes unfiltered off Windows (`hardware_linux.go:35`, `hardware_darwin.go:46`). Since SerialNumber is exactly what this module joins on, the load-bearing claim held, but the sentence would read as wrong to anyone grepping the Go tree.

The reviewer independently **verified** the rest: the 19-value list is an exact set match with `hardware.go:156-190`; the normalisation is functionally equivalent; the `'...'`-normalises-to-empty divergence from Go is real and correctly explained; `getDeviceWithOrgCheck` really does check in JS after an org-predicate-free select; `breeze_has_org_access` really does short-circuit to `TRUE` under system scope (`0001-baseline.sql:1663-1670`); the unique key really does mirror `organization_external_links` and not `contact_external_links`; the composite-FK deferrability is verbatim `device_mtls_certificates`; `breeze_device_child_orgid_tables()` really does auto-discover this table; and `orgMerge.ts` really does reject a cross-partner merge in three places, which is what makes the plain repoint safe.

### Test gaps — closed

The test analyst's two sharpest findings were whole code paths with **no assertion at all**, where a real regression would produce no red anywhere in the diff:

- `breeze_org_isolation_update` and `breeze_org_isolation_delete` had zero coverage — neither allow nor deny. Now: a cross-partner UPDATE and DELETE each reach zero rows and the victim row is asserted unchanged afterwards (RLS filters these silently rather than raising 42501), plus the same-tenant pair succeeding.
- **No test proved a legitimate non-system INSERT succeeds.** If `WITH CHECK` had regressed to always-false, every RLS test stayed green while every real import failed. Now covered.

Also added: the site boundary on the `deviceId` and `link` axes (previously proven only for hostname — and the link query carries no site predicate of its own, so it leans entirely on the device-detail query's reach clause); a multi-row batch, since the single-row helper never exercised the `IN`-list paths or link-derived ids folding back into the device fetch; a device with **no** `device_hardware` row (mutation-checked: `leftJoin`→`innerJoin` turns it red); an empty `allowedSiteIds`; a soft-deleted organization; both-axes-ambiguous candidate selection; a three-way identity conflict where a majority agrees and one identifier does not (there is no voting — the row is still refused); and the positive `DEFAULT_IMPORT_SYSTEM` link path, since the existing test only proved absence-of-match.

One test was reworded rather than kept as-is: *"an external link is scoped to the calling partner"* does not independently discriminate the `partnerId` predicate on the link query, because `orgIds` is already partner-filtered. It is now labelled as defence-in-depth, with a separate test that does confine the lookup by org.

### Dismissed, with reasoning

- **Tagged `OrgReach`/`SiteReach` sum types** instead of `string[] | null` (type-design reviewer). Declined: `accessibleOrgIds: string[] | null` is the repo-wide idiom (`DbAccessContext`, `services/contacts/import.ts`), callers pass `auth.accessibleOrgIds` straight through, and a bespoke shape here would be the odd one out for the next three waves. The concrete failure mode the reviewer named — `undefined` silently widening — is closed by finding #3 instead.
- **End-to-end org-erasure test for `device_external_links`** (test analyst). Declined: `tenantCascade.integration.test.ts` already asserts the five properties that matter (membership, alphabetical order, FK children before parents, no phantom tables, no duplicates), and the table's FKs are `ON DELETE CASCADE` from both `devices` and `organizations`, so an end-to-end delete would pass trivially without discriminating the registration. Extending another suite's hand-picked seed set is that suite's concern.

All in the foreground against a live Postgres:

- unit — `src/services/customFields/import/`, `cascadeDelete`, `moveOrg.coverage`, `autoMigrate`, `migrationRlsScope`: **248 passed**
- integration — `deviceExternalLinks`, `deviceCustomFieldImportIsolation`, `tenantCascade`, `orgLifecycleFoundations`, `tenant-export-policy`, `tenantExportErasureRoundtrip`: **50 passed**
- `vitest.config.rls-coverage.ts`: **100 passed**
- `tsc --noEmit`: clean · `pnpm lint`: clean · `check-migration-naming --against-ref origin/main`: OK · `check-drift`: no drift

**Non-vacuity was demonstrated, not assumed.** Four mutations were applied, observed red, and reverted: removing the partner predicate, removing the site predicate, changing the link separator to a space, and `leftJoin`→`innerJoin`. Each turned exactly the test that claims to cover it red.

Still not verified (unchanged from the PR body): no route mounts this until W08, no production-scale `EXPLAIN` was run against the two new indexes, and `source_instance` has no writer yet by design.

PR #4990, opened 2026-09-05
feat(api): the request side of script cancellation (#4761 W07, plan W02b)

### LINKED ISSUE #4989: W07: W07: Request side of script cancellation — cancelScriptExecution/cancelExecutionsForRun, rewritten POST /scripts/executions/:id/cancel with post-commit delivery, devices/events, cancel_script_execution AI tool (plan Tasks 2.4-request + 2.5; left out of #4799; W03/W05/W06 depend on it)
Wave W07 of #4761 — W07: Request side of script cancellation — cancelScriptExecution/cancelExecutionsForRun, rewritten POST /scripts/executions/:id/cancel with post-commit delivery, devices/events, cancel_script_execution AI tool (plan Tasks 2.4-request + 2.5; left out of #4799; W03/W05/W06 depend on it)
Branch: `feature/4761-cancel-running-script-or-automation/wave-4989`

ToddHebebrand:
Closes #4989

Wave W07 of #4761. This is the **request side** of script cancellation — the half the plan files under W02 and that #4799 explicitly deferred ("Deliberately out of scope": `cancelScriptExecution`, `cancelExecutionsForRun`, `deliverCancelCommand`, `CancelOutcome`, the route rewrite, `devices/events.ts`, and the AI tool). #4799 landed the columns, the enums, `CommandTypes.SCRIPT_CANCEL` + its four command registration lists, and closer 2 of 5 (`applyScriptCancelAck`).

Plan: `docs/superpowers/plans/agent/2026-09-02-cancel-running-script-or-automation.md`, Tasks 2.4 (request half) and 2.5.

## What changed

**`services/scriptCancellation.ts`** — the request half of the one state machine:

- `cancelScriptExecution` — loads the execution and its **paired `type = 'script'` command** `FOR UPDATE`, then branches:
  | command status | result |
  |---|---|
  | `pending` (never delivered) | retract it; execution → `cancelled` / `confirmed`. The only server-side proof a stop happened. |
  | `sent` | execution → transient `cancelling` / `requested`, `completed_at` stays NULL, a `script_cancel` is queued |
  | terminal | record `cancel_state = 'unconfirmed'` and leave `status` to the real result — a cancel would be doomed |
  | absent | **fail closed** (`inconsistent`). Absence is not proof that nothing ran. |
- `deliverCancelCommand` — claims and sends the committed row; releases the claim on a failed send so the heartbeat/HTTP-poll path still delivers it.
- `cancelExecutionsForRun` — keyed on `automation_run_id` **alone**, never the caller's org (a partner-wide run's executions carry the device's org). W05 owns the route that calls it.

**The wire contract (the #3525 blocker).** `script_cancel`'s `payload.executionId` carries the **original `device_commands.id`**, not `script_executions.id`, because that is what every deployed agent keys `e.running` on. `scriptExecutionId` travels additively. Pinned in both the unit suite and against the real jsonb column.

**`routes/scripts.ts`** — `POST /scripts/executions/:id/cancel` is now a thin delegation. Every guard is unchanged (org, site, `scripts:execute`, `requireMfa()`); the route no longer writes `script_executions` at all.

**`routes/devices/events.ts`** — `agent.command.script_cancel` and `script.execution.cancel.unconfirmed` audit labels.

**`services/aiToolsScripts.ts`** — `cancel_script_execution` at tier 3, beside `run_script`, delegating to the same service.

**`openapi.ts`** — documents the 409, the optional `graceSeconds` body, the 500, and the `cancelling`/`cancelState` response shape.

## AI-tool registration lists touched

Every one of these has its own zero-exemption contract test:

| File | List | Enforced by |
|---|---|---|
| `services/aiToolsScripts.ts` | `registerTool` (tier 3, `deviceArgs: []`) | `aiTools.deviceArgsCoverage.contract.test.ts` |
| `services/aiToolSchemas.ts` | `toolInputSchemas` | `aiToolsRegistryParity.test.ts` |
| `services/aiGuardrails.ts` | `TOOL_PERMISSIONS` (`scripts:execute`) | `aiToolsRegistryParity.test.ts` |
| `services/aiGuardrails.ts` | `TIER3_SUPERVISED_TOOLS` | `aiGuardrails.approvalScope.contract.test.ts` |
| `services/aiGuardrails.ts` | `TOOL_RATE_LIMITS` + approval-summary copy | (no completeness test; added for correctness) |
| `services/aiAgentSdkTools.ts` | `TOOL_TIERS` | `aiAgentSdkTools.registryParity.contract.test.ts` |
| `services/aiAgentSdkTools.ts` | the MCP `tool(...)` declaration | `aiAgentSdkTools.mcpCoverage.test.ts` |

Not touched, deliberately: `actionIntents/intentApprovers.ts` `DEVICE_COMPLETE_TARGET_TOOLS` (only required for sweep-proposable tools, and `cancel_script_execution` is not in the `SweepProposedAction` union), `toolTimeouts.ts` (60 s default is right), and `apps/web/src/components/ai-risk/tierConfig.ts` (out of this wave's file scope; not test-enforced — noted below).

## Deviations from the plan — code is truth

1. **`inDeliberateSystemContext`, not a bare `db.transaction`.** The plan says "commit, then deliver", but `authMiddleware` wraps the whole request in ONE `withDbAccessContext` transaction, so a bare `db.transaction` inside a route handler is only a **savepoint** — it commits nothing, and the `device_commands` row stays invisible to the agent's ack lookup, which reads on a fresh snapshot and would route the ack as orphaned. The service therefore escapes the request context (the same helper `automationActionResults.ts` uses). This is safe because the module is the authorization-free core by design and every caller gates first. **Proven, not asserted:** replacing the helper with a pass-through fails exactly one test — the real-Postgres visibility assertion — and no unit test at all.
2. **Dedup runs before the execution UPDATE, not after.** The plan's ordering writes `cancel_command_id = <fresh uuid>` and then returns an *existing* command's id, leaving `cancel_command_id` pointing at a row that was never inserted — which would make every closer that resolves an execution from its cancel command silently miss it.
3. **The retracted command's result is `status: 'cancelled'`, not the server-timeout marker.** The plan wrote `status: 'timeout'` "so a late ack can reopen the row", but `commandAcceptsAgentResultCondition` only reopens `status = 'failed'` rows — so the marker cannot reopen anything, and stamping `timeout` on a row we cancelled is a false claim `commandResultAcceptance.ts` documents relying on not happening.
4. **Added `{ kind: 'not_found' }`.** The plan mapped a missing execution to `already_terminal`; on a plan whose spine is the honesty contract, calling a nonexistent row "terminal" is the wrong answer. Route returns 404.
5. **An out-of-range `graceSeconds` is a 400, not silently clamped.** The plan discarded a failed body parse. The agent is only promised 0..30 s; quietly turning a requested 999 into 5 misreports what the endpoint is about to do.
6. **`cancelExecutionsForRun` returns a fourth bucket, `failed`.** The plan's `{requested, retracted, alreadyTerminal}` has nowhere to put an execution that vanished between the scan and the cancel, or one that failed closed — they would be dropped from the tally.
7. **`assertDeviceExecuteAllowed` is called and left unguarded.** It is a documented no-op for lifecycle command types today, so a denial cannot fire; `scriptCancellation.registration.test.ts` pins `script_cancel ∈ LIFECYCLE_COMMAND_TYPES`. Calling it means removing it from that list cannot silently bypass trust.

## Tests

Red-first throughout: the service suite failed 23/23 before the implementation, the route suite failed on all seven new assertions.

- `services/scriptCancellation.request.test.ts` (new, 23 tests) — wire contract, the full branch matrix, dedup, delivery, the users-FK degrade, and that nothing is sent from inside the transaction.
- `routes/scripts.test.ts` (extended) — 409 (was 400), idempotent 200, post-commit ordering, fail-closed 500, org/site denial before the service is consulted, `graceSeconds` handling, the audit's `outcome`, and a source-level assertion that the route still carries `requireMfa()` (this file's auth mock is a pass-through, so behaviour cannot prove it).
- `services/aiToolsScripts.cancelExecution.test.ts` (new, 7 tests) — the tenancy gate is the unique risk here, since the input names no device and the central `enforceDeviceArgs` hook cannot help. **Verified discriminating:** deleting the org condition fails the binding test.
- `__tests__/integration/scriptCancelRequest.integration.test.ts` (new, 6 tests, real Postgres) — post-commit visibility from another transaction, the wire contract against the real jsonb column, the `cancel_state` CHECK, the retraction branch, dedup, fail-closed, and that a decoy `list_processes` command carrying the same `payload.executionId` is not retracted instead of the script.

**Evidence**

- `npx tsc --noEmit` — clean (after merging `origin/main`).
- `pnpm lint` — 6/6 packages clean.
- `npx vitest run src/services/scriptCancellation src/routes/scripts src/services/aiTool src/services/aiGuardrails src/services/aiAgentSdkTools src/routes/devices/events src/services/commandResultHandlers src/openapi` — **136 files, 2201 passed, 2 skipped (pre-existing skips)**.
- Integration suite against `pnpm test-stack` — 6/6 passed; stack torn down.

## Not verified

- **No agent-side behaviour.** W04 owns the Go executor; every deployed agent today answers `script_cancel` with "unknown command type", which this code maps to `unconfirmed` (never `confirmed`) by design. Nothing here has been exercised against a real endpoint.
- **The remaining four closers are W03 (#4967).** Until they land, an execution can sit in `cancelling` if the ack never arrives — `applyScriptCancelAck` is the only closer in `main`. Not a regression (nothing reached `cancelling` before this PR), but worth knowing before release.
- **No UI.** `cancelling` and `cancelState` are not rendered; W01/W06 own that, and `apps/web` was out of this wave's file scope. `apps/web/src/components/ai-risk/tierConfig.ts` therefore does not list `cancel_script_execution` (documentation only — its parity test is one-directional and passes).
- **`cancelExecutionsForRun` has no caller yet.** W05 wires the automation-run route; its fan-out behaviour is unit-tested but never exercised end-to-end.
- **No load/concurrency testing** of the extra pooled connection the deliberate system context takes while the request transaction is open. Cancel is a rare, human-initiated action; the same shape already runs on the `queueCommand` dispatch path.
- **No migration** in this wave (#4799 shipped the columns).

https://claude.ai/code/session_014DyadMcDJ15Mx8gcoQuZ9W


ToddHebebrand:
## Review summary

Ran `/pr-review-toolkit:review-pr` — five specialist reviewers in parallel (code, silent-failure, tests, comments, type-design), all on Sonnet. **13 findings, 12 fixed, 1 dismissed.** Fixes are in `4c41c54` on top of the reviewed head.

### Critical (1) — fixed

**`createdBy: cancelledBy ?? ''` wrote an empty string into `device_commands.created_by`, a nullable `uuid`.** Postgres answers `22P02 invalid input syntax for type uuid: ""`, which rolls the entire cancel transaction back and 500s the caller. It fires precisely for the synthetic-principal case `resolveCancelledBy` degrades to `NULL` — an AI-agent or helper-session actor cancelling an already-delivered (`sent`) script. Every unit test passed through it because the mock does not type the column, and the integration suite only ever used a real `users` row.

Reproduced against real Postgres before fixing:

```
PostgresError: invalid input syntax for type uuid: ""
code: '22P02', where: "unnamed portal parameter $6 = ''"
```

`insertQueuedCommandInTransaction`'s `createdBy` is now `string | null` — which is what `queueCommand`'s own insert site already passed. New integration test `queues the cancel for an actor that is not a users row` covers it (red before, green after).

### Important (5) — fixed

- **`RunCancelTally.alreadyTerminal` counted a still-in-flight cancel as "nothing left to stop."** An `idempotent` outcome with `status = 'cancelling'` means another request's cancel is still awaiting the device — the script is running. Split into `alreadyCancelling` vs `noActionNeeded`. This is the same dishonesty the `status`/`cancel_state` split exists to prevent, and W05's run-cancel UI would have been the first thing to report it wrongly. Free to fix now; the type has no consumers yet.
- **`cancelExecutionsForRun` had no per-execution error isolation.** A transient deadlock on execution 3 of 20 threw out of the loop, so 17 devices were never asked to stop and the caller got a rejection instead of the tally the function's own docstring promises. Now caught per row → `failed`, with `captureException`.
- **A malformed JSON body was silently read as "no body."** `c.req.json()` throws the same error for empty and truncated, so `{"graceSeconds":30` became the 5 s default with a 200 — the exact silent substitution the out-of-range check two lines below refuses. Now a 400.
- **The AI tool's `recovered` outcome reads as success to a model.** It means the opposite: the cancel was too late and the script already finished on its own. All seven outcomes now carry a plain-English `detail`; `cancelling` says "NOT stopped yet", `recovered` says "TOO LATE … had no effect".
- **`CancelOutcome.status` was `string`.** Now the `execution_status` union derived from the Drizzle enum, so W03's closers get exhaustiveness checking instead of a silent never-match on a renamed value.

### Test-quality (4) — fixed

- **`the paired-command lookup constrains type=script` did not back its own name.** It asserted on compiled SQL text, where `type = $2` renders identically for `'script'` and `'script_cancel'` — the exact copy-paste bug it exists to catch. Now asserts the bound parameter values of both lookups.
- **`assertDeviceExecuteAllowed` was stubbed but never asserted on**, so passing the wrong `CommandTypes` constant, or swallowing a trust denial, would both have passed. Both now covered.
- **Exhaustiveness guard added to `cancelExecutionsForRun`'s switch.** Its `default` was doing double duty as the fail-closed bucket and the unknown-variant catch-all, so an eighth variant would have been miscounted as a failure. Verified discriminating: dropping the `inconsistent` case makes `tsc` fail.
- Added coverage for sweep error isolation, the malformed body, and the audited `inconsistent` case.

### Comment accuracy (3) — fixed

All three were claims a future author would have trusted:

- `script.execution.cancel.unconfirmed`'s comment described behaviour that does not exist — nothing writes that action yet. Now says so explicitly (it is forward-provisioning for W03's closers).
- "gates inline with verifyDeviceAccess, exactly as `get_script_execution` does" was an overclaim: `get_script_execution` has its own narrower site-only check. The cancel tool's gate is *stricter*; the comment now says that instead.
- The post-commit `applyAutomationActionTerminal` comment argued "self-deadlock." The reviewer was right that this is imprecise — inside our own system context the helper's guard sees `scope === 'system'` and runs inline rather than opening a second transaction. The real reason is durability ordering: closing the automation action inside an uncommitted cancel means a later failure silently rolls back a terminal transition the reconciler already published on. Comment rewritten to say that.

### Low (1) — fixed

The refused `inconsistent` case returned 500 with only a `captureException`. Now also audited, so an operator reading the device's history sees the attempt and the refusal.

### Dismissed (1)

**"Fold `deliverCancelCommand` into `cancelScriptExecution`."** The type-design reviewer is right that the `if (outcome.kind === 'cancelling' && !outcome.alreadyQueued) await deliverCancelCommand(...)` line is duplicated at three call sites and that a fourth (W05) could forget it. Not taken, for two reasons: the plan's produced interface explicitly annotates the variant `// caller must deliver post-commit`, and W03/W05 are in flight against that same interface right now — changing the service's contract mid-flight costs more than the duplication. And the failure mode of forgetting is bounded: the command row stays `pending`, so the heartbeat / HTTP-poll path still delivers it (verified: `claimPendingCommandsForDevice` has no type exclusion for `script_cancel`). Worth revisiting once all waves have landed.

### Reviewers confirmed clean

- **Tenancy** — no cross-tenant path found. The AI tool gates on `auth.orgCondition(scriptExecutions.orgId)` then `verifyDeviceAccess` (org + site) before reaching the authorization-free service; `deviceArgs: []` is safe because `enforceDeviceArgs` no-ops on an empty list and the handler gates inline.
- **No route guard was dropped.** Diffed against `origin/main`: `requireScope`, `requirePermission(SCRIPTS_EXECUTE)`, `requireMfa()`, org access and site access are all still present.
- **CAS/race correctness** — the paired `FOR UPDATE` locks serialise concurrent cancels and a concurrent agent claim on a `pending` command. No stranding path found beyond the `cancelling` state W03 explicitly owns.
- **`deliverCancelCommand` returning `false` is not a silent failure** — verified the backstop is real: the claim is released back to `pending` on every failure path, and the heartbeat poller has no type exclusion for `script_cancel`.
- **The integration suite's post-commit assertion genuinely discriminates** — independently traced through `db/index.ts`'s two-store `AsyncLocalStorage` escape, confirming the read runs on a second pooled connection under `READ COMMITTED`.
- No CLAUDE.md violations: no new tables or columns in this wave, so no RLS, cascade, export-policy or migration obligations apply.

- `npx tsc --noEmit` — clean.
- `pnpm lint` — 6/6 packages clean.
- Unit: 135 files, **2275 passed**, 2 skipped (pre-existing).
- Integration (`pnpm test-stack`, real Postgres): **7/7 passed**, including the new 22P02 regression. Stack torn down.

PR #4991, opened 2026-09-05
feat(api): custom_field_definitions uniqueness + org-XOR-partner + org-merge reconcile (#3257 W02)

### LINKED ISSUE #4770: W02: field_key uniqueness + org-XOR-partner + org-merge reconcile (gated on read-only prod duplicate-key preflight)
Wave W02 of #4768 — field_key uniqueness + org-XOR-partner + org-merge reconcile (gated on read-only prod duplicate-key preflight)
Branch: `feature/4768-rmm-custom-field-import/wave-4770`

ToddHebebrand:
Closes #4770

Wave W02 of #4768 / #3257. Hardens `custom_field_definitions` before the definitions importer (W07) can industrialise its two structural defects, and fixes the org merge that the new unique index would otherwise break.

## Todd gate — prod preflight (PASSED)

The plan's Task 1 preflight was run by the orchestrator on 2026-09-05 ~21:50Z against **both prod regions** (managed DO Postgres, via the billing container's postgres client), read-only:

- **breeze-eu:** `{"total":0,"a_duplicates":[],"b_ownerless":[]}`
- **breeze-us:** `{"total":1,"a_duplicates":[],"b_ownerless":[]}`

Both (a) duplicates and (b) ownerless rows are **empty** on both regions → the migration is safe to merge. (`total` counts all definition rows, not defects.) The preflight SQL is versioned in this PR at `apps/api/migrations/preflight/2026-10-10-100300-custom-field-definition-integrity-preflight.sql`; it was not re-run from this branch and no SSH was performed here.

## What changed

**1. Integrity migration** — `apps/api/migrations/2026-10-10-100300-custom-field-definition-integrity.sql`

The table shipped in the squashed baseline with a PK and two FKs and nothing else, leaving:

- **No unique key on `field_key`** → a definitions import has no idempotency key, so a re-run mints a duplicate `udf7`. Two definitions on one key also make the partner export emit two records for one datum (the identity hash includes `f.id`).
- **`(org_id, partner_id) = (NULL, NULL)` structurally legal** → a row invisible to every non-system caller that survives org cascade forever, because the cascade deletes by `org_id`. A latent GDPR orphan.

Adds `custom_field_definitions_one_owner_chk` plus **two partial per-axis unique indexes**. Not one composite: over two nullable columns btree treats every NULL as distinct, and since every partner-wide row has `org_id NULL`, a composite would leave the partner axis enforcing nothing. Pre-existing duplicates are **reported and abort the deploy**, never auto-resolved — deleting one silently retypes every value stored under that key (the survivor dictates `type`; values carry none) and renaming one orphans them instantly, breaking consumers holding the key as configuration (`remoteAccessLauncher.ts`, `installerVariables.ts`).

**2. `#2135` step-6 partner RLS suite** — `customFieldDefinitionsPartnerRls.integration.test.ts` (new, 10 tests). The table was converted to dual-axis in `2026-06-11-i` but never got the playbook's contract suite. `rls-coverage` proves the policy *exists* by reading `pg_catalog`; only this proves either branch *enforces* anything.

**3. Org merge reconcile** — `custom_field_definitions` moves from `REPOINT_TABLES` to a `custom` policy backed by `rehomeChildrenThenDelete` (key: `field_key`), the same shape `mergePlaybookDefinitions` uses.

1. **Migration slot renamed.** The plan names `2026-10-07-100000-…`, which now sorts *before* the newest shipped migration (`2026-10-10-100000-device-external-links.sql`, W06) and the pre-push guard rejects it. Used **`2026-10-10-100300-…`** for the migration and the preflight, and everywhere they are referenced. Confirmed sort-last against `origin/main` before each push; `check-migration-naming: OK` on every commit.

2. **The migration elects `breeze.scope = 'system'` even though it writes no rows.** CLAUDE.md states this rule for *writes*, but the same mechanism silently corrupts a detection *read*, and here that is the more dangerous half. The table is `FORCE ROW LEVEL SECURITY`, which binds the **owner** role migrations run as; `breeze_current_scope()` defaults to `'none'`, under which both access helpers are false. On prod (managed DO Postgres, non-superuser `doadmin`) both detection probes would therefore match **zero rows** and print a truthful-looking `0 duplicate pairs`. The deploy would still abort — but on the DDL below, with a bare 23505 naming an index instead of the actionable `owner / field_key` list the file exists to print. CI's superuser masks the difference entirely. Not added to `migrationRlsScope.test.ts`'s baseline (that guards writes; this file has none) — 100 tests green there.

3. **The plan's Task-3 assertion was wrong and is corrected.** The plan expects `23514` for an ownerless `(NULL, NULL)` row under a *partner* context. Verified against real Postgres: it is **`42501`** — with both columns NULL neither branch of the RLS `WITH CHECK` can match, and row security is evaluated before the CHECK constraint. The `23514` half is proved separately under **system** scope in `customFieldDefinitionIntegrity.integration.test.ts`, where RLS lets the row through and the constraint is all that remains. Both halves matter: the constraint is what protects migrations, backfills and the importer's own system-context writes, none of which RLS stops. Same ordering `cis_baselines` already documents.

4. **Preview mirror registered** (not in the plan). `custom_field_definitions` added to `CUSTOM_WOULD_DROP_COUNTS`. Without it `previewOrgMerge` would report `wouldDrop: 0` for a merge about to delete definitions — the exact non-destructive-looking plan that map's own header warns about. Asserted live: `{ loserRows: 2, wouldDrop: 1 }`.

5. **Merge proof is a new file**, `customFieldDefinitionsMerge.integration.test.ts`, rather than an addition to the 92KB `orgMerge.integration.test.ts` (the plan allowed either). Uses `executeOrgMerge` — the plan's snippet names a `mergeOrganizations` that does not exist.

6. **One stale comment corrected in `rls-coverage.integration.test.ts`.** It cited `custom_field_definitions` as an example of an *app-enforced-only* XOR; it is DB-enforced as of this wave. No allowlist membership changed (that set only drives the partner-wide SELECT-branch assertions) — suite green at 102 passed. This is the only edit to that file.

7. **Extra coverage beyond the plan's snippets**, all cheap and all discriminating: uniqueness is scoped to one owner (two orgs may both define `asset_tag`); the CHECK holds on `UPDATE`, not just `INSERT`; the merge fixture gives the loser a **non-colliding** definition too, so an executor that dropped *everything* to dodge the 23505 would fail; the surviving row is asserted **by id**, so swapping which definition wins fails; and partner-wide rows (`org_id NULL`) are proven untouched by an org merge.

## The 23505 this fixes, proven live

Red-first, by reverting the policy to `repoint` and running the new merge test against real Postgres:

```
PostgresError: duplicate key value violates unique constraint
  "custom_field_definitions_org_key_uq"
  query: UPDATE "custom_field_definitions" SET org_id = $1::uuid WHERE org_id = $2::uuid
```

For two orgs imported from one Datto tenant this fires on **every** shared key, so it is the common path for exactly the customers #3257 serves — it would have shipped broken. The generic `repoint-dedupe` DELETE is not a safe substitute either: it is harmless only because nothing references a definition today, and once `device_custom_field_values` lands (W05) its `definition_id` FK makes a blind dedupe cascade away every stored value. `CUSTOM_FIELD_DEFINITION_CHILDREN` ships in place and empty so W05 is a one-line addition.

## Verification (all foreground, live DB via `pnpm test-stack`)

| Check | Result |
|---|---|
| `customFieldDefinitionIntegrity` (new) | 7 passed — **red first: 5 failed** ("the statement succeeded") before the migration |
| `customFieldDefinitionsPartnerRls` (new) | 10 passed |
| `customFieldDefinitionsMerge` (new) | 4 passed — **red first: 4 failed with 23505** on the `repoint` policy |
| `orgMergeCustomExecutors.test.ts` (unit) | red first (`not a function`) → green |
| Integration sweep: `orgMergeRegistry`, `orgMergeCustomExecutors`, `orgLifecycleFoundations`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip` + the 3 new | 13 files / **77 passed** |
| `vitest.config.rls-coverage.ts` | **102 passed** |
| Unit: `orgMerge*`, `autoMigrate`, `migrationRlsScope` | 5 files / **180 passed** |
| `tsc --noEmit` | clean (exit 0) |
| `eslint` on all 7 touched files | clean |
| `pnpm db:check-drift` | no drift, 676 migrations match the ledger |
| `check-migration-naming` | OK on every commit and at push |

`origin/main` merged in before the final push (2 web-only commits; `git diff --name-only` over `apps/api` is empty, so the verification above still holds for the merge commit).

## Not verified

- **Not run against prod.** The preflight was executed by the orchestrator before this branch existed (results quoted verbatim above); nothing in this PR re-ran it and no SSH was performed. The migration's abort path (`RAISE EXCEPTION` on real duplicates) is therefore exercised only against seeded local data, never against prod rows.
- **The `rehomed` branch of the merge note is unreachable today** — `CUSTOM_FIELD_DEFINITION_CHILDREN` is deliberately empty until W05. Its string-building is untested because no child exists to re-home; W05 must cover it when it adds `device_custom_field_values`.
- **No web/UI work, no importer, no `device_custom_field_values`** — W05/W07/W09 own those; nothing under `apps/web` is touched.
- **The `#4944` gap is pinned, not fixed.** `custom_field_definitions` still has no partner-wide `SELECT` branch, so an org token remains blind to its own partner's partner-wide definitions. The suite asserts that as *current* behaviour with an explicit note that the test must be **inverted** when #4944 lands.
- **Full `apps/api` unit suite not run** (only the affected files, per repo guidance on the parallel-flaky suite). CI is the gate for the rest.
- Integration Tests shard behaviour on CI not observed from here — the three new files land under the shared `src/__tests__/integration/**` glob, so they are auto-discovered, but that is inferred from the config, not watched running in CI.

https://claude.ai/code/session_01BuG4Lp1fQ8mZhvEfTY1SJz

---

## Review round (added after `/pr-review-toolkit:review-pr`)

Four reviewers ran: code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer. **4 findings raised, all 4 fixed in `cdff375292`; 0 outstanding.** Two reviewers returned zero findings in their primary category.

1. **The preflight file had no scope elevation** (silent-failure-hunter, MEDIUM) — the exact trap the migration header warns about, one file over. **Verified by simulating prod**: as a `NOSUPERUSER`/`NOBYPASSRLS` table owner the unelevated probe reported `scope=none total_rows=0` while a row existed (an operator reads "clean" while blind); with the elevation, `scope=system total_rows=1`. Fixed: `set_config` added, file wrapped in `BEGIN READ ONLY`/`ROLLBACK` so `is_local=true` actually spans both queries, plus a query (0) printing effective scope and total row count so blindness is self-evident rather than silent.

2. **The migration's `RAISE EXCEPTION` abort path was untested** (pr-test-analyzer, sev 6) — the guard between a dirty prod DB and a deploy. Every other test runs against a DB where the file already applied cleanly. Added three tests that drop the integrity objects, forge dirty rows, replay the migration by path, and assert `P0001` + the operator-actionable message, restoring shipped state in `finally` (which also proves re-application is a true no-op). **Verified non-vacuous by mutation**: swapping `RAISE EXCEPTION` for `RAISE NOTICE` makes both fail with the raw `23505`/`23514` DDL errors instead of `P0001` — exactly the degraded diagnostics the header describes.

3. **A doc comment pointed at the wrong test** (pr-test-analyzer, sev 3). It claimed the "two different orgs" test guards partial-vs-composite index design; it doesn't (both orgs have distinct non-NULL `org_id`s, so neither design collides). Repointed at the test that actually discriminates.

4. **Two factual comment errors** (comment-analyzer). The `inArray` rationale misattributed the mechanism (it's Drizzle's `sql` tag expanding an array into `($1, $2, $3)`, not postgres.js binding a scalar) and stated a count-dependent SQLSTATE as fixed. And the `rls-coverage` comment this wave edited still cited `client_ai_prompt_templates` as app-enforced-only — it has carried `client_ai_prompt_templates_scope_check` since 2026-06-12. **Confirmed against `pg_constraint` on a live DB**, along with `access_reviews` genuinely having none; it is now the last such example and the note says so.

Verification re-run after the fixes: integration sweep **13 files / 80 passed** (was 77 — 3 new abort tests), rls-coverage **102 passed**, unit **180 passed**, `tsc --noEmit` clean, eslint clean on all 7 touched files.

**Additional not-verified item from this round:** the preflight's `BEGIN READ ONLY … ROLLBACK` form was exercised against a prod-*like* local role (non-superuser, non-BYPASSRLS, table owner), not against a real prod region. The client the operator uses there may differ.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer (4 agents, parallel, Sonnet).

**Findings:** 4 raised → all 4 addressed in cdff375292; **0 outstanding.** code-reviewer returned 0 findings (independently re-verified all six registration lists, the partial-index design, and preview/executor predicate parity); silent-failure-hunter returned 0 in the migration and executor, 1 in the preflight.

| # | Reviewer | Sev | Finding | Resolution |
|---|---|---|---|---|
| 1 | silent-failure-hunter | MEDIUM | Versioned preflight had **no scope elevation** — the exact blindness the sibling migration is hardened against, one file over | Fixed: `set_config` + `BEGIN READ ONLY`/`ROLLBACK` so `is_local=true` spans both queries, plus a query (0) printing effective scope + total rows so blindness is self-evident |
| 2 | pr-test-analyzer | 6 | Migration's `RAISE EXCEPTION` **abort path untested** — the guard between a dirty prod DB and a deploy | Fixed: 3 tests replaying the migration by path against forged dirty rows, asserting `P0001` + the operator-actionable message |
| 3 | pr-test-analyzer | 3 | Doc comment pointed at a test that doesn't discriminate partial-vs-composite indexes | Fixed: repointed at the test that does |
| 4 | comment-analyzer | — | Two factual comment errors (`inArray` mechanism/SQLSTATE; `client_ai_prompt_templates` called app-enforced-only) | Fixed both; `client_ai_prompt_templates_scope_check` confirmed present via `pg_constraint` on a live DB |

**Both substantive findings were verified empirically, not reasoned about:**

- *Finding 1* — simulated prod with a `NOSUPERUSER`/`NOBYPASSRLS` table owner: the unelevated probe reports `scope=none total_rows=0` while a row exists (operator reads "clean" while blind); with the elevation, `scope=system total_rows=1`. This also independently confirms the migration's own scope-elevation deviation was necessary rather than defensive.
- *Finding 2* — the new abort tests were **mutation-tested**: swapping `RAISE EXCEPTION` for `RAISE NOTICE` makes both fail with the raw `23505`/`23514` DDL errors instead of `P0001`, i.e. precisely the degraded diagnostics the migration header describes. They are not vacuous.

**Tests (re-run after the fixes, all foreground against a live DB):**
- Integration sweep — 13 files / **80 passed** (was 77; +3 abort tests)
- `vitest.config.rls-coverage.ts` — **102 passed**
- Unit (`orgMerge*`, `autoMigrate`, `migrationRlsScope`) — 5 files / **180 passed**
- `tsc --noEmit` clean · `eslint` clean on all 7 touched files · `db:check-drift` no drift · `check-migration-naming: OK`

**Red-first evidence** (unchanged, in the PR body): integrity suite 5 failed → 7 passed; merge suite failed with the real `23505` on `custom_field_definitions_org_key_uq` under the old `repoint` policy → 4 passed.

**Status:** review-clean, CI green at time of writing, awaiting maintainer merge. Not merging and not closing #4770 — those are the maintainer's calls.

Issue #4941, opened 2026-09-05
[Agent] Reboot deferral: postponing opens a new modal immediately; Linux logs a WARN for a clean postponement


ToddHebebrand:
## Description

Two small end-user rough edges in the reboot-deferral ladder (#3207), seen on real Windows 11 and Ubuntu 22.04 endpoints during the pre-v0.110 check on 2026-09-05.

### 1. Postponing immediately opens another modal

After the user clicks **Postpone**, `RebootManager` re-plans the schedule and the new plan's lead rung is deferrable, so a fresh "Restart Scheduled" dialog opens *immediately* (Windows: within a second; Linux: zenity respawns at once). The user has just said "not now" and is handed the same question again with the counter decremented ("You can postpone this restart 1 more time.").

Observed on both platforms:

```
11:37:40 reboot scheduled delay=9m22s notifications=2 …   ← re-plan
11:37:40 reboot postponed by user newDelay=9m22s used=1 max=2
(new dialog on screen at 11:37:41)
```

Suggested: after a postponement, suppress the lead rung's prompt (or delay it by the postponement window) and rely on the reminder rungs for the next offer. The closing rung is already never deferrable.

### 2. Linux logs a WARN for a clean postponement

`reboot_prompt_desktop.go` treats zenity exit 1 as ambiguous and logs

```
WARN the reboot dialog did not end in a clean decision session=c18 user=todd exitCode=1 timedOut=false shown=true stderr=""
INFO reboot postponed by user newDelay=10m32s used=1 max=2
```

for every `--extra-button` press, even though the very next line shows it was resolved as a postponement from stdout (that is the documented design: Cancel, close box, ESC and every extra button all exit 1). The WARN should only fire when exit 1 arrives with no stdout label.

Neither item blocks v0.110; both are UX/log polish for the next agent release.

PR #4993, opened 2026-09-06
fix(agent): one reboot-deferral dialog per Windows session (#4940)

### LINKED ISSUE #4940: [Agent][Windows] Reboot-deferral prompt is shown twice per session (system-role + user-role helper both receive the notify decision)
## Description

On a Windows workstation the reboot-deferral prompt (#3207) is drawn **twice** for every deferrable rung: once by the system-role user helper and once by the user-role user helper, both of which live in the same console session and both of which register the `notify` scope.

`Broker.RequestNotificationDecision` (`agent/internal/sessionbroker/broker.go`) fans the `TypeNotify` request out to *every* session with the `notify` scope. In always-on lifecycle mode the broker spawns two helpers into the console session (`helperKey=1-system` and `1-user`, log: `proactively spawned helper in session`), and both report `canNotify=true`. Each renders its own `MessageBoxTimeoutW`, so the user sees two identical modal dialogs (two taskbar entries). First-answer-wins resolves the decision, but the *other* dialog stays on screen until its own timeout (up to 2 minutes), and its late `notify_result` is dropped with `dropping unsolicited or unauthorized helper message`.

## Observed (real box, 2026-09-05)

Windows 11 Pro 25H2 physical workstation, agent built from `main` @ `a52504b9e` (`0.110.0-rc.rebootcheck`), console user signed in, `schedule_reboot` with `allowDeferral=true, maxDeferrals=2, deferralMinutes=5, delayMinutes=6`.

- Screenshot at first prompt: two "Restart Scheduled" taskbar entries, one dialog on top of the other.
- Clicking **No** (postpone) on one: `reboot postponed by user used=1` — correct. The second dialog remained until its 2-minute window expired: `11:38:03 dropping unsolicited … notify_result sessionId=helper-7c82… role=system`.
- Re-planned prompt: both helpers again; both timed out → two `dropping unsolicited` lines at `11:39:40` (`role=user`) and `11:39:41` (`role=system`).

Agent log excerpt:

```
11:33:17 spawned user helper in session pid=6960 … role=system windowsSessionId=1
11:33:17 spawned user helper in session pid=12472 … role=user windowsSessionId=1
11:33:18 capabilities received uid=0 canNotify=true canTray=true  canCapture=true  (system helper)
11:33:18 capabilities received uid=0 canNotify=true canTray=false canCapture=false (user helper)
11:36:02 reboot scheduled delay=6m0s notifications=2 … deferralAllowed=true
11:37:40 reboot postponed by user newDelay=9m22s used=1 max=2
11:38:03 WARN dropping unsolicited or unauthorized helper message type=notify_result sessionId=helper-7c82e7db464fc3e3 role=system
11:39:40 WARN dropping unsolicited or unauthorized helper message type=notify_result sessionId=helper-ee10cb50179fa842 role=user
11:39:41 WARN dropping unsolicited or unauthorized helper message type=notify_result sessionId=helper-7c82e7db464fc3e3 role=system
```

## Expected

One dialog per Windows session. Either the broker should target one notify-capable helper per `windowsSessionId` for interactive prompts (prefer the user-role helper, which runs as the signed-in user), or the losing helper should be told to close its dialog when a decision arrives elsewhere.

`BroadcastNotification` (the buttonless toast path) has the same fan-out and presumably shows duplicate toasts on the same session; worth checking in the same fix.

## Not affected

- Linux (daemon-drawn zenity, one dialog per logind session) — verified on the same day, single dialog.
- RDS hosts resolve `mode=on-demand`, so no resident helper and no prompt at all there (separate, by design).

Found during the pre-v0.110 real-box reboot-deferral check (#3207 handoff). Not a v0.110 blocker on its own: the policy defaults off, and the deadline still guarantees the reboot.

ToddHebebrand:
## Summary

- `Broker.RequestNotificationDecision` and `Broker.BroadcastNotification` fanned `TypeNotify` out to **every** session holding the `notify` scope. In always-on lifecycle mode the broker spawns two helpers into the same console session (`helperKey=1-system` and `1-user`) and `systemHelperScopes` / `userHelperScopes` each grant `"notify"`, so both rendered their own `MessageBoxTimeoutW` — two "Restart Scheduled" taskbar entries, one dialog stacked on the other (#4940).
- Both paths now resolve their recipients through a new `notifyTargets()` (`agent/internal/sessionbroker/notify_targets.go`): every `notify`-scoped session, reduced to **at most one helper per interactive Windows session**. The scope filter itself is untouched (#3255).
- Within a session the **user-role** helper wins, because it runs as the signed-in user and therefore owns the dialog the user has to act on; the **system-role** helper is the fallback for a session with no user token (logon screen, or a disconnected RDP session that retained its SYSTEM helper).
- Helpers whose Windows session id is unknown are **never merged**. Every Linux and macOS helper reports the `"0"` that the non-Windows `currentWinSessionID()` stub returns, so grouping on it would collapse a multi-user host to a single warning — the fail-unsafe direction. `"0"` is already this repo's sentinel for "no interactive session known" (`ConsoleSessionID` normalizes it to `""`).
- Multi-session Windows hosts are unaffected in the direction that matters: de-duplication is per screen, not global, so each interactive session is still prompted exactly once.
- The issue's secondary observation is fixed too: `BroadcastNotification` had the identical fan-out and drew every buttonless toast twice on the same session.

## Root cause

`TypeNotify` recipients were selected by scope alone, with no notion of *which screen* a helper draws on:

```go
b.mu.RLock()
for _, s := range b.sessions {
    if s.HasScope("notify") { sessions = append(sessions, s) }
}
b.mu.RUnlock()
```

`HasScope("notify")` is a per-process property, but a notification is a per-desktop event. The Windows lifecycle manager deliberately runs two helper processes per active session (`isWindowsLifecycleRole` admits `{sessionID, role}` keys, and both roles are granted `"notify"` — the system helper for tray/desktop/PAM, the user helper for `run_as_user`), so "every notify-scoped session" is two sessions per screen on every always-on Windows host.

First-answer-wins resolved the *decision* correctly, which is why the reboot itself was never wrong. What it could not resolve was the *UI*: the losing helper had already been handed a real correlated request, so it drew its dialog and held it open for the full countdown (up to 2 minutes), then replied to an envelope whose pending entry `SendCommand` had already deleted — producing the `dropping unsolicited or unauthorized helper message type=notify_result` warnings in the report.

## Verification

All commands run from `agent/` in this worktree, in the foreground.

Red first (against the current `main` behaviour, with `selectNotifyTargets` stubbed to the identity so the package compiled):

```
go test -race -run 'TestSelectNotifyTargets|TestRequestNotificationDecision|TestBroadcastNotification' ./internal/sessionbroker/
```
5 top-level tests failed, 11 assertions — including `the system-role helper was asked for a decision the user-role helper in the same Windows session already received`, `session 1 received the toast twice — once per helper role`, and a non-deterministic `ActionClicked = "Restart now", want the user-role helper's answer` that demonstrated the race directly.

After the fix:

| Command | Result |
|---|---|
| `go test -race -count=1 ./internal/sessionbroker/...` | ok, **256 passed / 1 skipped / 0 failed** (15.5s) |
| `go test -race -count=1 -v -run 'NotifyTargets\|NotificationDecision\|BroadcastNotification' ./internal/sessionbroker/` | **20 tests / 41 pass lines** (incl. subtests), 0 failed |
| `go test -race -count=5 -run 'TestSelectNotifyTargets\|TestRequestNotificationDecision\|TestBroadcastNotification' ./internal/sessionbroker/` | ok, 5 consecutive runs — no flakiness (4.2s) |
| `go test -race -count=1 ./internal/heartbeat/... ./internal/patching/... ./internal/userhelper/... ./internal/ipc/...` | all ok (heartbeat 50.2s, patching 12.9s, patching/linuxsession 3.0s, userhelper 3.6s, ipc 3.5s) |
| `go test -race -count=1 ./...` | **exit 0**, all 77 packages ok / no test files, 0 failures |
| `go vet ./internal/sessionbroker/...` | clean |
| `gofmt -l internal/sessionbroker/` | clean |
| `GOOS=windows go build ./...` and `GOOS=linux go build ./...` | both OK |

New tests: `agent/internal/sessionbroker/notify_targets_test.go` (table-driven pure-function coverage of `selectNotifyTargets` plus broker-level end-to-end tests over real socket pairs, using the existing `decisionProbe` harness and the sentinel-after-the-fact idiom from `broker_notify_test.go` so the "was not asked" assertions are not races against a sleep). Coverage: user+system in one session → only the user helper; reversed input order → same winner (map iteration is random in production); system-only → the system helper; two sessions × two roles → one per session; mixed console/RDS roles; no-Windows-session-id never merged; multi-session hosts still prompt every screen; the surviving helper's click still reaches the caller; `BroadcastNotification` one toast per session.

No pre-existing test needed changing — `TestBroadcastNotificationFiltersOnNotifyScope`, `TestRequestNotificationDecisionReturnsTheFirstAnswer`, `TestRequestNotificationDecisionSurvivesADeadSession` and `TestBroadcastNotificationContinuesPastSendFailure` all build fixtures with no Windows session id, which is exactly the never-merged case.

- **I implemented the first of the two options the issue offers, not both.** The issue's *Expected* is "One dialog per Windows session. **Either** the broker should target one notify-capable helper per `windowsSessionId` … **or** the losing helper should be told to close its dialog when a decision arrives elsewhere." Targeting one helper per session satisfies that outright and makes the second option unnecessary *for the reported defect*: no same-session helper is ever handed a request it can lose, so there is nothing to cancel.
- **The dispatch brief also asked for cancel-on-decision; I deliberately left it out.** After per-session de-duplication the only remaining "losers" are helpers in *other* Windows sessions, i.e. different users on different screens. Tearing their modal dialogs down needs new IPC surface (`notify_cancel`), helper-side tracking of a `MessageBoxTimeoutW` window that is created with `hWnd=0` and blocks its OS thread, an `EnumThreadWindows` + `WM_CLOSE` teardown from a second thread, new reply semantics for "cancelled ≠ unanswered ≠ not-shown", and darwin/linux variants — none of which is testable off a real Windows box. The issue also notes the multi-session case does not occur in practice today ("RDS hosts resolve `mode=on-demand`, so no resident helper and no prompt at all there"). Filed as a follow-up rather than smuggled in here; happy to do it as a second PR if a maintainer wants it.
- **Tie-break is lock-free on purpose.** `betterNotifyTarget` prefers user-role, then newest `ConnectedAt`, then lowest `SessionID` — all write-once fields — instead of reusing `betterSession`, which also reads `LastSeen` (rewritten under `s.mu` by `Touch()` on every inbound message). Two same-role helpers in one session are already refused by admission (`errDuplicateHelperKey`), so the tie-break is a determinism guarantee rather than a live path; it may as well not add a race surface.
- **A small delivery trade-off, stated explicitly.** Previously, if the user-role helper failed to render, the system-role helper in the same session was a second chance. Now only one is asked. If it reports `Delivered=false`, `handleNotify` has already fallen back to a plain toast inside the helper, and `chainedRebootPrompt` still falls through to `patching.DesktopPrompt` (a no-op off Linux). I did **not** add a broker-side retry of the runner-up: it would cost a second full prompt window on the failure path, which is the exact budget concern `reboot_prompt.go` already documents. The reboot deadline still guarantees the restart either way.
- **`RebootManager` untouched**, per the parallel work in #4941 (re-prompt after postponement). No overlap: that issue is about *when* the manager raises a rung, this one is about *who* the broker hands it to.
- **No CHANGELOG entry.** Recent `fix(agent)` commits do not add one, and the `[Unreleased]` section is release-scoped rather than per-PR.
- Added a `log.Debug` (not Warn) when the fan-out collapses, so "why did only one of my two helpers get the prompt" is answerable from the log. It fires on every rung of the reboot ladder on an always-on Windows host, which is the fix working, not a problem.

Closes #4940

### DIFF
--- a/agent/internal/sessionbroker/broker.go
+++ b/agent/internal/sessionbroker/broker.go
@@ -1337,8 +1337,8 @@ func (b *Broker) TCCStatus() *ipc.TCCStatus {
 	return nil
 }
 
-// BroadcastNotification sends a desktop notification to every connected session
-// holding the "notify" scope.
+// BroadcastNotification sends a desktop notification to one connected session per
+// interactive Windows session, chosen from those holding the "notify" scope.
 //
 // The scope filter is load-bearing, not cosmetic. The Breeze Assist helper
 // connects with assist/consent_ui and the watchdog with "watchdog"; neither
@@ -1356,15 +1356,13 @@ func (b *Broker) TCCStatus() *ipc.TCCStatus {
 // the fix was to grant the macOS desktop helper "notify" in scopesForRole —
 // NOT to weaken this filter, which still has to keep the assist helper and
 // watchdog out of any delivery accounting built on this broadcast.
+//
+// The fan-out is one helper per Windows session, not one per notify-scoped
+// session (#4940): in always-on lifecycle mode the broker spawns a system-role
+// and a user-role helper into the same console session and both hold "notify", so
+// an unfiltered broadcast drew every toast twice. See notifyTargets.
 func (b *Broker) BroadcastNotification(title, body, urgency string) {
-	b.mu.RLock()
-	sessions := make([]*Session, 0, len(b.sessions))
-	for _, s := range b.sessions {
-		if s.HasScope("notify") {
-			sessions = append(sessions, s)
-		}
-	}
-	b.mu.RUnlock()
+	sessions := b.notifyTargets()
 
 	for _, s := range sessions {
 		if err := s.SendNotify("", ipc.TypeNotify, &ipc.NotifyRequest{
@@ -1386,13 +1384,14 @@ func (b *Broker) BroadcastNotification(title, body, urgency string) {
 // or a rung and a retry easily are, would lose one answer.
 var notifyDecisionSeq atomic.Uint64
 
-// RequestNotificationDecision sends an interactive notification to every
-// notify-scoped session and returns the first answer.
+// RequestNotificationDecision sends an interactive notification to one
+// notify-scoped session per interactive Windows session and returns the first
+// answer.
 //
-// This is the response-bearing sibling of BroadcastNotification, which is
-// deliberately left alone: the reboot warning LADDER stays fire-and-forget
-// (#3197 guarantees the user is TOLD; it does not need an answer). Only the
-// rungs that offer a postponement come through here.
+// This is the response-bearing sibling of BroadcastNotification, which
+// deliberately stays fire-and-forget: the reboot warning LADDER needs no answer
+// (#3197 guarantees the user is TOLD). Only the rungs that offer a postponement
+// come through here.
 //
 // The plumbing already existed and was never usable end to end.
 // NotifyRequest.Actions and NotifyResult.ActionClicked are declared in
@@ -1412,15 +1411,17 @@ var notifyDecisionSeq atomic.Uint64
 // must not be weakened (#3255): the assist helper and the watchdog have no
 // TypeNotify handler at all, and a modal prompt is strictly more intrusive than
 // the toast that filter was written for.
+//
+// One prompt per Windows session, not per notify-scoped session (#4940). Always-on
+// lifecycle mode spawns a system-role and a user-role helper into the same console
+// session and both report canNotify=true, so the unfiltered fan-out drew two
+// identical modal dialogs — two taskbar entries, one stacked on the other.
+// First-answer-wins resolved the decision, but the losing dialog stayed on screen
+// for its whole countdown and its late notify_result was dropped as unsolicited.
+// The loser is now never asked, which is also why there is nothing to cancel: see
+// notifyTargets.
 func (b *Broker) RequestNotificationDecision(req ipc.NotifyRequest, timeout time.Duration) (ipc.NotifyResult, error) {
-	b.mu.RLock()
-	sessions := make([]*Session, 0, len(b.sessions))
-	for _, s := range b.sessions {
-		if s.HasScope("notify") {
-			sessions = append(sessions, s)
-		}
-	}
-	b.mu.RUnlock()
+	sessions := b.notifyTargets()
 
 	if len(sessions) == 0 {
 		// Not an error: a headless box, or Windows sitting at the logon screen.
--- a/agent/internal/sessionbroker/notify_targets.go
+++ b/agent/internal/sessionbroker/notify_targets.go
@@ -1,0 +1,118 @@
+package sessionbroker
+
+import (
+	"sort"
+
+	"github.com/breeze-rmm/agent/internal/ipc"
+)
+
+// notifyGroupKey identifies the SCREEN a notify-scoped helper draws on.
+//
+// Two helpers that share a Windows session id share one desktop, one taskbar and
+// one signed-in human, so a notification fanned out to both is drawn twice
+// (#4940). A helper whose Windows session id is unknown gets a key of its own and
+// is therefore never merged with anything: on Linux and macOS every helper
+// reports the "0" the non-Windows currentWinSessionID stub returns, and grouping
+// on that would collapse a host with several logged-in users down to a single
+// warning. "0" is already this repo's sentinel for "no interactive session
+// known" — ConsoleSessionID normalizes it to "" and every consumer fails closed
+// on the empty value — so treating it as ungroupable is the same convention.
+func notifyGroupKey(s *Session) string {
+	if id := s.WinSessionID; id != "" && id != "0" {
+		return "win:" + id
+	}
+	return "session:" + s.SessionID
+}
+
+// betterNotifyTarget reports whether candidate should draw the notification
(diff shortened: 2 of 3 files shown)


ToddHebebrand:
**Tests:** `agent/internal/sessionbroker/notify_targets_test.go` (new) 20 tests / 41 pass lines, 0 failed; `go test -race -count=1 ./internal/sessionbroker/...` 256 passed / 1 skipped / 0 failed; `-count=5` re-run of the new tests clean; `go test -race -count=1 ./internal/heartbeat/... ./internal/patching/... ./internal/userhelper/... ./internal/ipc/...` all ok; `go test -race -count=1 ./...` exit 0 across all 77 packages; `go vet ./internal/sessionbroker/...` clean; `gofmt -l internal/sessionbroker/` clean; `GOOS=windows`/`GOOS=linux go build ./...` both OK. `tsc --noEmit` n/a — Go-only change, no TypeScript touched.

PR #4704, opened 2026-09-03
feat(web): platform-admin UI for the AI kill switch (#4208)

ToddHebebrand:
## Summary
- Adds an **AI Kill Switch** page under the platform-admin Administration section (`/admin/ai-kill-switch`), the first UI on top of the `ai_kill_state` API + MFA + audit surface shipped in #3828 / PR #4168.
- Shows current state (active/killed), epoch, last reason, and provenance (who/when), with a confirm-with-required-reason flow to flip it in either direction. Surfaces a friendly message when the session hasn't satisfied MFA (`MFA_REQUIRED`), and a locked "platform admin required" panel on a 403 — same pattern as `LlmProviderCatalog.tsx`.
- Full i18n parity across all 8 locales (nav label + full `admin.aiKillSwitch.*` content), verified against the repo's strict `localeParity.test.ts` contract.
- Updates the stale "no UI" comments in `routes/admin/index.ts` and `aiKillState.test.ts`, and adds the console path to `docs/deploy/ai-kill-switch.md` (SQL fallback documentation is unchanged and still the production path today — prod has zero platform admins).

No API changes — the route/service/authorization/audit contract was already fully covered by `apps/api/src/routes/admin/aiKillState.test.ts` (13 tests, unchanged behavior).

Closes #4208

## Test plan
- [x] `cd apps/web && npx vitest run src/components/admin/AiKillSwitch.test.tsx` — 6/6 passed
- [x] `npx vitest run src/components/layout/Sidebar.*.test.tsx` (all 6 sidebar suites) — 47/47 passed
- [x] `npx vitest run src/lib/i18n/localeParity.test.ts` — 79/79 passed (all 8 locales)
- [x] `npx vitest run src/lib/__tests__/no-silent-mutations.test.ts` — 130/130 passed
- [x] `npx tsc --noEmit -p .` — clean
- [x] `npx astro check` — 0 errors
- [x] `cd apps/api && npx vitest run src/routes/admin/aiKillState.test.ts` — 13/13 passed (unchanged, confirms doc-only edits didn't regress behavior)
- [x] `pnpm eslint` on touched files — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4994, opened 2026-09-06
fix(admin): resolve AI kill-switch actor to a name (#4931)

### LINKED ISSUE #4931: [Admin] AI kill switch "Last changed by" shows a raw user UUID
Found in the pre-release sweep (docs/testing/release-sweeps/2026-09-05-v0.109.0-to-main.md, Group 3).

**Symptom:** `/admin/ai-kill-switch` renders the actor of the last change as a raw UUID. This is the one platform-wide control whose audit trail justifies its mandatory reason field, so the actor should be a name.

**Fix:** the kill-switch state DTO needs to resolve the actor to a display name (or email) server-side; the page then shows the name with the UUID in a tooltip/title. Not a one-file web change, hence filed.

Refs #4704

ToddHebebrand:
## Summary
- `GET /api/v1/admin/ai-kill-state` now returns `updatedByName` and `updatedByEmail` alongside the existing `updatedBy` UUID, resolved from `users` in the same query.
- `/admin/ai-kill-switch` renders "Last changed by" as the actor's **name** (email when the name is blank/absent) with the raw UUID in a `title` tooltip; it degrades to the UUID, then to an em dash, rather than going blank.
- Resolution rides the **system-scoped** read the admin GET already opens, so an actor whose `users` row belongs to a different partner than the caller still resolves. It is a LEFT JOIN, so a deleted user or an ops SQL flip (`updated_by` NULL) returns nulls instead of dropping the kill-state row.
- The hot-path cached read (`readAiKillState`) is untouched and stays join-free — a deleted actor cannot influence the kill state it hands the guardrails.
- No schema/migration change, no new tenancy surface, no locale-key change (the existing `fields.updatedBy` / `fields.none` keys cover it; a UUID tooltip is not translatable).

## Root cause
`ai_kill_state.updated_by` is a bare UUID with no FK — deliberately, so ops can flip the row through the SQL fallback documented in `docs/deploy/ai-kill-switch.md`. `readAiKillStateRow` selected that column verbatim and `AiKillSwitch.tsx` rendered `row.updatedBy` straight into the `<dd>`, so the only actor attribution on the platform's single most consequential AI control was an opaque identifier.

Nothing in the DTO ever joined `users`, and the web client cannot resolve one itself: `users` is dual-axis RLS (tenancy shape 4), so a partner/org-scoped read under the admin's own session would not reliably see another platform admin's row. The resolution had to happen server-side, under system scope.

## Verification
All commands run in the foreground from the worktree root; Node 22.23.2.

| Command | Result |
|---|---|
| `cd apps/api && npx vitest run src/services/aiKillState.test.ts src/routes/admin/aiKillState.test.ts` | 2 files, **32 passed** |
| `cd apps/api && npx vitest run aiKillState actRevalidation` | 4 files, **78 passed** |
| `cd apps/api && npx vitest run src/routes/admin` | 11 files, **170 passed** |
| `cd apps/web && npx vitest run src/components/admin/AiKillSwitch.test.tsx` | 1 file, **11 passed** |
| `cd apps/web && npx vitest run src/components/admin` | 5 files, **65 passed** |
| `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit --project apps/api/tsconfig.json` | **clean** (exit 0) |
| `cd apps/web && npx astro check` | **0 errors, 0 warnings** (252 pre-existing hints across 1877 files) |
| `npx eslint` over all 6 changed source/test files | **clean** |

TDD order was followed: the service test (asserting `leftJoin` is called on the admin read) and three web tests (name label, email fallback, UUID tooltip) were written first and observed failing — 1 API failure and 3 web failures — before the implementation landed.

- **Field names:** `updatedByName` / `updatedByEmail` rather than the `lastChangedByName` / `lastChangedByEmail` floated in the dispatch note. The existing DTO columns are `updatedBy` / `updatedAt`, and "Last changed by" is already the en label for `fields.updatedBy`, so the new fields follow the column they resolve. `updatedBy` is retained unchanged for compatibility — and the UI still needs it, both for the tooltip and to disambiguate two admins who share a display name.
- **Nulls, not a synthesized fallback:** the API returns `updatedByName: null` for an unresolvable actor instead of echoing the UUID as though it were a name. That keeps "unresolved" distinguishable from "a user actually named like a UUID" for any future consumer (CSV export, AI tools), and puts the display fallback where the display strings live. The UI chain is name, then email, then UUID, then em dash.
- **Blank names count as absent** — `users.name` is `NOT NULL` but not non-empty, so a whitespace-only name falls through to the email rather than rendering an empty field.
- **RLS reasoning was read, not assumed:** `breeze_has_partner_access` short-circuits to `TRUE` when `breeze_current_scope() = 'system'` (`apps/api/migrations/2026-04-11-a-rls-function-bootstrap.sql:111`), which is what makes the joined `users` row visible under `breeze_user_isolation_select`. The exposure is one actor's name and email, to a caller who already passed `platformAdminMiddleware` and who already sees `actorEmail` on the audit-log surface that joins `users` the same way (`apps/api/src/routes/auditLogs.ts:349`).
- **No cascade / export-policy registration needed:** no new table, and no new column on any `org_id`-bearing table. `ai_kill_state` is system-scoped and stays that way; this change is read-side only.
- **No DB-backed integration test.** This route has no integration coverage today and no Postgres was reachable from this worktree, so the LEFT JOIN's null tolerance is pinned at three levels instead: the query chain (mocked `db` — asserts the join happens on the admin read and *not* on the hot path), the route DTO, and the component render. If a maintainer wants real-DB proof, the natural home is a new `aiKillStateAdmin.integration.test.ts`; happy to add it on request.
- **Flagged, deliberately not changed:** `handleFlip` applies the POST's `killed`/`epoch` optimistically and lets provenance lag until the confirmatory refetch. If that refetch fails, the card now shows a stale *human name* where it used to show a stale UUID. That path already renders a red "Failed to load kill switch state" banner directly above the card, and blanking provenance there is a separate behaviour change with its own test surface — out of scope for a display fix, but worth a follow-up if you read it differently.
- **Context:** per `docs/deploy/ai-kill-switch.md`, production currently has zero platform admins, so this page is unreachable there today; the fix lands ahead of that changing. Found in the v0.109.0-to-main pre-release sweep (#4704, Group 3). The runbook's UI section was updated to describe the new label/tooltip behaviour.

Closes #4931

### DIFF
--- a/apps/api/src/routes/admin/aiKillState.test.ts
+++ b/apps/api/src/routes/admin/aiKillState.test.ts
@@ -79,6 +79,8 @@ const ROW = {
   epoch: 4,
   reason: 'restored after incident 123',
   updatedBy: 'admin-0',
+  updatedByName: 'Ada Lovelace',
+  updatedByEmail: 'ada@breeze.test',
   updatedAt: new Date('2026-08-28T12:00:00Z'),
 };
 
@@ -139,10 +141,42 @@ describe('admin AI kill-state routes', () => {
           epoch: 4,
           reason: 'restored after incident 123',
           updatedBy: 'admin-0',
+          updatedByName: 'Ada Lovelace',
+          updatedByEmail: 'ada@breeze.test',
           updatedAt: '2026-08-28T12:00:00.000Z',
         },
       });
       expect(readRowMock).toHaveBeenCalledTimes(1);
+    });
+
+    it('keeps updatedBy alongside the resolved actor name (#4931)', async () => {
+      // The UI shows the name and puts the UUID in a tooltip, so both must
+      // survive: dropping `updatedBy` would break the existing client and
+      // leave nothing to disambiguate two admins with the same display name.
+      const res = await buildApp(platformAdmin).request('/admin/ai-kill-state');
+      const { data } = await res.json() as { data: Record<string, unknown> };
+      expect(data.updatedBy).toBe('admin-0');
+      expect(data.updatedByName).toBe('Ada Lovelace');
+      expect(data.updatedByEmail).toBe('ada@breeze.test');
+    });
+
+    it('returns null name/email when the actor no longer resolves (#4931)', async () => {
+      // A deleted user, or a row flipped by ops through the SQL fallback in
+      // docs/deploy/ai-kill-switch.md (updated_by NULL). The route must still
+      // 200 with the kill state — never 500, never a synthesized display name.
+      readRowMock.mockResolvedValue({
+        ...ROW,
+        updatedBy: 'admin-gone',
+        updatedByName: null,
+        updatedByEmail: null,
+      });
+
+      const res = await buildApp(platformAdmin).request('/admin/ai-kill-state');
+      expect(res.status).toBe(200);
+      const { data } = await res.json() as { data: Record<string, unknown> };
+      expect(data.updatedBy).toBe('admin-gone');
+      expect(data.updatedByName).toBeNull();
+      expect(data.updatedByEmail).toBeNull();
     });
   });
 
--- a/apps/api/src/routes/admin/aiKillState.ts
+++ b/apps/api/src/routes/admin/aiKillState.ts
@@ -4,6 +4,10 @@
  *   GET  /api/v1/admin/ai-kill-state          state + epoch + provenance
  *   POST /api/v1/admin/ai-kill-state          flip it (+ MFA, reason required)
  *
+ * The GET's provenance carries the actor as a resolved name/email alongside
+ * the raw `updatedBy` UUID (`readAiKillStateRow`, #4931) — both null when the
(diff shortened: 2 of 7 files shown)


ToddHebebrand:
**Tests:** `apps/api` — `src/services/aiKillState.test.ts` + `src/routes/admin/aiKillState.test.ts` (2 files, 32 passed), `aiKillState actRevalidation` (4 files, 78 passed), `src/routes/admin` (11 files, 170 passed); `apps/web` — `src/components/admin/AiKillSwitch.test.tsx` (1 file, 11 passed), `src/components/admin` (5 files, 65 passed); `tsc --noEmit --project apps/api/tsconfig.json` clean, `astro check` (apps/web) 0 errors / 0 warnings; eslint clean on all 6 changed files.

Issue #4562, opened 2026-09-02
Customer portal visibility — dashboard, security, backups, reports, support usage over existing data


ToddHebebrand:
## Why

An MSP's customer-facing promise is "log in and see it": security status, work completed, backup verification, patching, support responsiveness, and self-serve compliance PDFs. Today the customer portal (`apps/portal`) is a billing-and-tickets portal: the landing page is Proposals, and none of those categories has a customer-facing surface. Most of the underlying data already exists server-side and org-scoped.

Audit (2026-09-02) of 36 customer-facing "you'll see" claims against `apps/portal` and the reports engine:

- 4 shipped: devices online/offline, tickets, invoices + pay, proposals + sign
- 16 surface-only: data exists, no customer page or PDF section
- 7 partial (some data), 9 missing (no data anywhere)

## Wave 1 — portal dashboard over existing data (this item)

0. **Trust hazards first.** `apps/web/src/components/reports/ReportTemplates.tsx` cards "SLA Compliance", "Patch Compliance", "Technician Activity", "Billing/Usage" alias onto device/software inventory. `apps/web/src/components/audit/ComplianceReport.tsx` and `software/SoftwareComplianceReport.tsx` render hardcoded mock data. Remove or make real.
1. **Landing dashboard** (replaces Proposals as post-login landing): security score + band + 30-day trend (`security_posture_org_snapshots`), devices protected / unprotected (`security_status`), patches applied this month (`device_patches`), last backup verified (`backup_verifications`), open tickets + avg first response this month (`tickets`), open action items (`fleet_findings`, `remediation_suggestions`).
2. **Security page**: score history, per-device protection table (AV, real-time protection, definitions age, encryption, firewall, pending critical patches), threats detected/resolved by week (`security_threats`, `s1_threats`, `huntress_incidents`), open vulnerabilities by severity.
3. **Reports tab**: list completed `report_runs` for the org with download; "Generate now" for Security & Compliance Posture and Executive Summary; bind scheduled-report recipients to org contacts instead of free-text emails.
4. **Support usage**: first-response / resolution time + SLA badge on the ticket list; month-to-date billable hours by ticket (`time_entries`); invoice line items show the originating ticket (`invoice_lines.source_type = time_entry`).
5. **Backups page**: last successful backup per device, last test-restore result, RPO/RTO breaches, recovery readiness.
[...]

PR #4995, opened 2026-09-06
fix(portal): redirect disabled visibility pages instead of erroring (#4932)

### LINKED ISSUE #4932: [Portal] Disabled visibility pages show an error message instead of redirecting
Found in the pre-release sweep (docs/testing/release-sweeps/2026-09-05-v0.109.0-to-main.md, Group 4).

**Symptom:** when an MSP turns a portal visibility toggle OFF, the customer opening that page sees *"We couldn't load your security summary just now. Your IT team can help."* — an error, inviting them to raise a ticket about a switch the MSP deliberately turned off.

**Cause:** only `apps/portal/src/pages/reports/index.astro:21-23` redirects on `PORTAL_*_DISABLED`; security, backups, dashboard and devices do not.

**Fix:** apply the same redirect (or a neutral "not available for your account" state) on every gated portal page; add a portal e2e case per toggle.

Refs #4562

ToddHebebrand:
## Summary

- Every visibility-gated portal page now answers its gate 403 with a redirect instead of rendering a load-failure page. `security`, `backups`, `dashboard`, `devices` and `assets` join `reports`, which was the only page that already did this.
- The shared logic lives in one new module, `apps/portal/src/lib/visibilityGate.ts`: `isPortalPageDisabled(...responses)` recognises the eight `PORTAL_*_DISABLED` codes and `redirectToPortalHomeAfterDisabled(Astro)` bounces to `/quotes`.
- `/quotes` is the one signed-in page no visibility flag can switch off, so a redirect chain or loop is structurally impossible. `/reports` no longer targets `/devices` — itself gated on Self-service — which turned one deliberate switch-off into two hops ending in a second false "we couldn't load your devices" error.
- `/assets` was the worst case: `AssetList` prints `error` verbatim, so the customer was shown our own internal wording, *"Asset checkout is not enabled for this portal"*.
- New contract test `apps/portal/src/lib/disabledPageCoverage.test.ts` walks every page under `src/pages`, maps its `portalApi.*` calls to the gate they sit behind, and fails on any page that renders instead of branching — the same shape as the existing `sessionClearCoverage.test.ts` guard for 401s. A parity case in `visibilityGate.test.ts` reads the API gate sources and fails if the API grows a `PORTAL_*_DISABLED` code the portal does not know.
- Public docs described the old behaviour in two places (`features/portal.mdx`: the "Turned off means genuinely off" aside and the Reports section) and are updated to match.

## Root cause

The API side of #4562 was correctly fail-closed: `createPortalFeatureGateStrict` answers 403 with a machine-readable `code`, and `apiRequest` propagates that code onto every `ApiResponse`. The pages just did not read it.

Each page branched on `statusCode === 401` (dead session, handled by `redirectToLoginAfter401`) and then treated *every* other failure identically, by passing `response.error` down to a component whose error copy is written for a transport failure — "We couldn't load your security summary just now. Your IT team can help." A 403 caused by a configuration decision therefore rendered as an outage and invited the customer to raise a ticket about a switch their IT team had just thrown.

`reports/index.astro` had the branch (`statusCode === 403 && code === 'PORTAL_REPORTS_DISABLED'`) because it was written together with the gate; the five pages added around it were not, and nothing enforced the pattern — so the omission was invisible to review, exactly the failure mode `basePathCoverage.test.ts` and `sessionClearCoverage.test.ts` were written to prevent for their own contracts.

## Verification

Red first, then green — the new tests were run before the fix existed:

| Command | Result |
|---|---|
| `cd apps/portal && npx vitest run src/pages src/lib/disabledPageCoverage.test.ts src/lib/visibilityGate.test.ts` — **before the fix** | 8 files failed / 10 tests failed. `visibilityGate.test.ts` could not resolve the module, and the coverage test named all five offending pages: `assets`, `backups`, `dashboard`, `devices`, `security` |
| same command — **after the fix** | **9 files, 44 tests passed** |
| `cd apps/portal && npx vitest run` (full portal suite) | **52 files, 465 tests passed** |
| `npx tsc --noEmit -p apps/portal` | **clean, 0 errors** |
| `cd apps/portal && npx astro check` | **133 files, 0 errors, 0 warnings** (43 pre-existing unused-import hints, none in a file this PR touches) |
| `PORTAL_BASE_PATH=/portal pnpm build --filter=@breeze/portal` | **build complete** (only the pre-existing chunk-size warning) |

CI runs these as the `test-portal` (`pnpm test --filter=@breeze/portal`) and `build-portal` jobs.

No API, database, tenancy or agent code is touched — no migration, no RLS surface, no cascade list.

- **Redirect rather than a neutral "not available" state, on all six pages.** The issue allows either. Redirect wins here because the nav item is already gone when the toggle is off, so the only way to reach these URLs is a bookmark or a hand-edited address — there is no in-page context worth preserving, and a redirect cannot be mistaken for a fault. `/tickets` is the deliberate exception and keeps its existing neutral state, see below.
- **Fixed target `/quotes`, not "the landing page".** `portalLandingPath()` returns `/dashboard` when `enableDashboard` is on, so a "bounce to the landing page" rule loops on the dashboard page itself the moment the branding read and the API gate disagree. `/quotes` (labelled Proposals) carries auth but no gate in `apps/api/src/routes/portal/index.ts`, is the first always-on nav item, and is already `DEFAULT_LANDING` in `lib/session.ts` and the fail-closed branch of `lib/landing.ts`. `visibilityGate.test.ts` asserts the target is not in `PORTAL_GATED_PAGES`, so the invariant is checked rather than hoped for.
- **`/reports` changed target from `/devices` to `/quotes`.** A behaviour change beyond the five broken pages, made because the old target is itself gated and reproduced the same bug one hop later. This is what the two docs sentences described, hence the docs edit.
- **`/tickets` is intentionally left alone.** It already implements the alternative the issue offers: `decideTicketsPage()` keeps the page up when tickets are off but support-usage is on, and renders "Your IT team hasn't opened request submission in the portal" — a real, actionable state, not a load failure. Its both-off path still redirects to `/devices`; that is now safe (Devices bounces on to `/quotes` when Self-service is off) but the `redirectToDevices` naming in `lib/ticketsPage.ts` is stale. Renaming it would touch a tested module for a case this issue does not report, so it is left as a follow-up. `tickets/index.astro` is the single documented entry in the coverage test allowlist; `tickets/new.astro` and `tickets/[id].astro` already name their gate code and pass unaided.
- **`/assets` was not named in the issue but is the same defect**, on the same fail-closed gate, with the worst copy of the six — included for one line of frontmatter.
- **The backups ride-along dashboard call is excluded from the gate check.** `/backups` also fetches `/portal/dashboard` purely for `timezone`; an org with Backups on and Dashboard off must keep its backups page. `backups/index.test.ts` asserts `dashboard` never appears in the `isPortalPageDisabled(...)` argument list.
- **Non-gate 403s still render inline.** `portalAuthMiddleware` answers "Account is not active" / "Organization is not available" as a bare 403 with no `code`, so keying on the code rather than the status preserves the behaviour the old `/reports` comment called out. Unit-tested.
- **Page tests follow the established source-assertion pattern** already used by the four existing `src/pages/**/index.test.ts` files (read the `.astro`, assert its structure). They are deliberately thin — the general contract lives in the coverage test, and each page test only pins what is page-specific.

### Follow-up: the Playwright e2e cases the issue asks for

Deferred out of this PR by dispatch instruction. `data-testid`-only, per `e2e-tests/README.md`:

1. Security off: `/portal/security` answers **302 to `/portal/quotes`**; `portal-security-error` never renders.
2. Backups off: `/portal/backups` **302 to `/portal/quotes`**; `portal-backup-error` never renders.
3. Backups **on** + Dashboard **off**: `/portal/backups` still answers **200** (guards the ride-along timezone call).
4. Dashboard off: `/portal/dashboard` **302 to `/portal/quotes`**; `portal-dashboard-error` never renders.
5. Reports off: `/portal/reports` **302 to `/portal/quotes`** (previously `/portal/devices`).
6. Self-service off: `/portal/devices` **302 to `/portal/quotes`**.
7. Asset checkout off: `/portal/assets` **302 to `/portal/quotes`**, and the string "Asset checkout is not enabled for this portal" appears nowhere in the body.
8. Reports **and** Self-service off: `/portal/reports` reaches `/portal/quotes` in **one** hop (the chain the old `/devices` target produced).
9. Per toggle, the matching nav item is absent from the portal shell.
10. A non-gate 403 (inactive account) still renders the page's inline error rather than bouncing.

Closes #4932

Refs #4562 — the visibility epic's server-side gating shipped in that epic's waves and is unchanged here; this PR is the presentation half the pre-release sweep found missing.

### DIFF
--- a/apps/docs/src/content/docs/features/portal.mdx
+++ b/apps/docs/src/content/docs/features/portal.mdx
@@ -61,7 +61,7 @@ To change them:
 You need the **Organizations: write** permission and a multi-factor confirmation to change these.
 
 <Aside type="note" title="Turned off means genuinely off">
-  Visibility is enforced on the server, not just in the portal's navigation. A section that is off is hidden from the menu **and** refused by the API, so a customer who bookmarks the URL or edits it by hand gets an "unavailable" page rather than a partial one. An organization that has never had portal settings saved is treated exactly like one with every toggle off — it fails closed, never open.
+  Visibility is enforced on the server, not just in the portal's navigation. A section that is off is hidden from the menu **and** refused by the API, so a customer who bookmarks the URL or edits it by hand is taken back to their proposals instead of being shown a page that looks like it failed to load. An organization that has never had portal settings saved is treated exactly like one with every toggle off — it fails closed, never open.
 </Aside>
 
 ### Ticket Statuses
@@ -193,7 +193,7 @@ With **Reports** visibility on, the customer can generate and download reports t
 
 A finished run offers both **PDF** and **CSV** downloads. These are the same report types described in [Reports](/features/reports/); only the trigger is different.
 
-With Reports off, a customer who navigates to the reports URL is redirected back to their devices.
+With Reports off, a customer who navigates to the reports URL is redirected to their proposals — the same as every other section that is off.
 
 ---
 
--- a/apps/portal/src/lib/disabledPageCoverage.test.ts
+++ b/apps/portal/src/lib/disabledPageCoverage.test.ts
@@ -1,0 +1,123 @@
+import { describe, expect, it } from 'vitest';
+import { readdirSync, readFileSync, statSync } from 'node:fs';
+import { join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+/**
+ * Contract test: a portal page whose data comes from behind a visibility gate
+ * MUST branch on the gate's 403 code in its frontmatter instead of falling
+ * through to its generic "we couldn't load this" copy.
+ *
+ * Without the branch, an MSP switching a toggle off in Settings → Organizations
+ * → Customer Portal hands the customer a page that reads as a transient failure
+ * and invites a support ticket about a switch that was deliberate (#4932). Only
+ * /reports got the branch; Security, Backups, Dashboard, Devices and Equipment
+ * all shipped the misleading copy, and nothing in review caught the fifth one.
+ *
+ * Frontmatter, not the template: the answer to a switched-off page is a 302 the
+ * server returns before it renders anything, so a check written into the markup
+ * would still ship the page shell. The API side is already fail-closed and
+ * unit-tested (apps/api/src/routes/portal/featureFlags.test.ts) — this guards
+ * the half that renders. sessionClearCoverage.test.ts is the same idea for 401s.
+ */
+
+const PAGES = fileURLToPath(new URL('../pages', import.meta.url));
+
+/** portalApi methods that sit behind a visibility/feature gate, and the 403
+ *  code the gate answers with (apps/api/src/routes/portal/index.ts). */
+const GATED_API_METHODS: Record<string, string> = {
+  getDashboard: 'PORTAL_DASHBOARD_DISABLED',
+  getSecurityOverview: 'PORTAL_SECURITY_DISABLED',
+  getSecurityDevices: 'PORTAL_SECURITY_DISABLED',
+  getBackupOverview: 'PORTAL_BACKUPS_DISABLED',
+  getBackupDevices: 'PORTAL_BACKUPS_DISABLED',
+  getReportRuns: 'PORTAL_REPORTS_DISABLED',
+  getSupportUsage: 'PORTAL_SUPPORT_USAGE_DISABLED',
+  getDevices: 'PORTAL_SELF_SERVICE_DISABLED',
+  getAssets: 'PORTAL_ASSET_CHECKOUT_DISABLED',
+  getTickets: 'PORTAL_TICKETS_DISABLED',
+  getTicket: 'PORTAL_TICKETS_DISABLED',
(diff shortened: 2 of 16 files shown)


ToddHebebrand:
**Tests:** `cd apps/portal && npx vitest run` — 52 files / 465 tests passed (includes the 9 new+extended files for this fix: 44 tests — `lib/visibilityGate.test.ts`, `lib/disabledPageCoverage.test.ts`, and the `pages/{security,backups,dashboard,devices,assets,reports}/index.test.ts` set, all confirmed red before the fix); `npx astro check` — 133 files, 0 errors / 0 warnings; `PORTAL_BASE_PATH=/portal pnpm build --filter=@breeze/portal` — build complete; `tsc --noEmit` clean (0 errors on `apps/portal`)

PR #4996, opened 2026-09-06
fix(web): Processes tab must not read a 503 as "0 processes / No Data" (#4935)

### LINKED ISSUE #4935: [Remote Tools] Processes tab swallows a 503 into "Processes 0 · No Data"
Found in the pre-release sweep (docs/testing/release-sweeps/2026-09-05-v0.109.0-to-main.md, Group 5).

**Symptom:** on an offline device, Remote Tools → **Processes** renders `Processes 0 · No Data` when the API answers 503 — indistinguishable from an idle box. The Terminal tab one click away correctly says "Device is not online".

**Cause:** `apps/web/src/components/devices/RemoteToolsPage.tsx:382` treats the non-2xx as an empty list.

**Fix:** surface the offline/unavailable state the way Terminal does (shared offline banner), and keep the empty state for a genuine 200 with zero rows.

ToddHebebrand:
## Summary

- Remote Tools → **Processes** on an offline device rendered `Processes 0 · No Data`, indistinguishable from a genuinely idle box, while the Terminal tab one click away explained why nothing was coming back.
- `RemoteToolsPage` now keeps the failure reason in a new `processError` state instead of swallowing it into `console.error`, reading the server's own message with the existing `extractApiError` helper (a 503 carries `{ error: 'The device is offline.', code: 'device_offline' }`).
- `ProcessManager` gains an optional `loadError` prop rendered as a failure pane with a Retry action, checked **before** the empty state — so "fetch failed" and "device reported zero processes" are mutually exclusive by construction, not by convention.
- The summary tiles show `-` and the `Showing 0 of 0 processes` footer is suppressed while the fetch is in a failed state; both were printing their initial zeroes as though they were measurements.
- The `No Data` empty state is untouched for a real 200 with zero rows, and is now pinned by a test.
- No new i18n keys, so all 8 locale files are unchanged.

## Root cause

`fetchProcesses` did `if (!res.ok) throw new Error(t('remoteToolsPage.errors.fetchProcesses'))` inside a `try` whose `catch` did nothing but `console.error`. The thrown message never reached state, so `processes` stayed at its `useState([])` initial value and `ProcessManager` — which only had `processes.length === 0` to go on — rendered `processManager.noData`. The API's 503 body from `apps/api/src/routes/systemTools/fileBrowserHelpers.ts` was discarded unread.

The zero-valued tiles and footer compounded it: they report counts derived from the same empty array, so the pane asserted "0 processes" three more times.

## Verification

Red first — the three failure-path tests were written before the fix and confirmed failing (9 passed / 3 failed), then all 12 passed after:

```
cd apps/web && npx vitest run src/components/remote/RemoteToolsPage.test.tsx
```
→ **12 passed / 0 failed**

```
cd apps/web && npx vitest run src/components/remote src/lib/i18n
```
→ **300 passed / 0 failed** across 28 files (includes `localeParity`, `translationCoverage`, `keyUsage`, `extractionQuality`, `terminologyQuality`)

```
cd apps/web && npx vitest run
```
→ **8369 passed / 0 failed** (full `apps/web` suite)

```
cd apps/web && npx tsc --noEmit -p tsconfig.json
```
→ clean, exit 0

```
cd apps/web && npx eslint src
```
→ clean, exit 0

New tests added to `apps/web/src/components/remote/RemoteToolsPage.test.tsx` (alongside source, per repo convention):

| Test | Pins |
|---|---|
| renders the offline reason instead of "No Data" when the API answers 503 | the fix; also asserts tiles show `-` and the count footer is gone |
| offers a retry that re-requests the process list | Retry re-issues the fetch |
| clears the error once a retry succeeds | recovery path, error does not stick |
| keeps the "No Data" empty state for a genuine 200 with zero rows | **guard** — pre-existing correct behaviour must not regress |
| renders rows and no error for a 200 with processes | **guard** — happy path |

One note on the full-suite run: the first invocation reported 2 failures in `LinkSubscriptionPicker.test.tsx` and `SoftwareGroupDrawer.test.tsx`. Both are unrelated components (Pax8 billing picker, vulnerability drawer) and both pass in isolation on a stashed clean `origin/main` **and** in isolation with this change; a second full run was 8369/8369 green. They are load-related async flakes, not caused by this diff.

- **The dispatch hint's premise was wrong, so I did not follow it literally.** It said to reuse "that exact banner/component" from the Terminal tab. There is no such thing: the Terminal tab (`RemoteTerminal.tsx`, the component actually mounted by `RemoteToolsPage`) has no offline banner — it writes errors into the xterm pane and fires `onError` → toast. The yellow banner in `RemoteTerminalPage.tsx` belongs to a component that **no Astro route mounts** (only re-exported from `components/remote/index.ts`; the real route `pages/remote/terminal/[deviceId].astro` imports `RemoteToolsPage`), i.e. effectively dead code. `deviceActions.unavailable.notOnline` ("Device is not online") is used only as a tooltip `title`, never as a banner. A repo-wide search found **no** `DeviceOfflineBanner`/`OfflineBanner`/`UnavailableBanner` component at all.
- **I reused the File Browser tab's pattern instead** — `FileManager.tsx:1471-1492`, a sibling tab inside this very page, already does exactly what the issue asks (`loading → error → empty → rows`, red `AlertCircle` + server message + `common:actions.retry`). Matching the neighbour that already got it right seemed better than inventing a new banner or resurrecting unrouted code. Flagging in case a maintainer would rather I extract that shared pane into a real component — I kept it inline to stay minimal, which also matches the 12 other inline `border-yellow-500/40` banners in the codebase.
- **Server message passthrough rather than a new translated key.** Copying English text into the 7 non-en locales would have broken `translationCoverage.test.ts`, which enforces *per-namespace per-locale* caps on exact-English duplicates (`remote.json`: pt-BR 12, es-419 12, fr-FR 18, fr-CA 17, de-DE 14, it-IT 14, tr-TR 4). Surfacing the API's own string (what `FileManager` already does) with the already-translated `remoteToolsPage.errors.fetchProcesses` as fallback avoids touching locale files entirely. Trade-off: the offline sentence itself is server-authored and therefore English-only. If maintainers want it localised, the right follow-up is to branch on the 503's `code` (`device_offline` / `device_unreachable` / `agent_timeout`) and map each to a genuinely translated key — no web code reads that `code` field today.
- **`setProcesses([])` on failure.** Without it, a device that goes offline while rows are on screen would show a stale list next to the error banner. Mirrors `FileManager`'s `setEntries([])`.
- **Scope: Processes only, as the issue specifies.** The same swallow-into-empty-list shape exists in `fetchServices`, `fetchEventLogs` and `fetchTasks` in this same file (identical `throw` → `console.error` → `useState([])`), so Services / Event Log / Scheduled Tabs have the same defect on an offline device. I left them alone to keep this reviewable and did not want to widen a release-sweep fix; `loadError` is now the pattern to copy if a maintainer wants them done. **Suggest filing a follow-up.**
- `fetchProcesses`'s `useCallback` deps stay `[deviceId]` (not adding `t`) to match the four sibling fetchers in this file; adding `t` would change effect re-run semantics on locale switch, which is out of scope here.
- No API, schema, migration, tenancy/RLS or agent surface touched, so no contract-suite or `db:check-drift` run applies.

Closes #4935

### DIFF
--- a/apps/web/src/components/remote/ProcessManager.tsx
+++ b/apps/web/src/components/remote/ProcessManager.tsx
@@ -5,6 +5,7 @@ import {
   X,
   ChevronUp,
   ChevronDown,
+  AlertCircle,
   AlertTriangle,
   Cpu,
   HardDrive,
@@ -38,6 +39,13 @@ export type ProcessManagerProps = {
   deviceName?: string;
   processes?: Process[];
   loading?: boolean;
+  /**
+   * Why the last fetch failed, or null/undefined when it succeeded. Set this
+   * whenever `processes` is empty because of a failure rather than because the
+   * device reported none — otherwise the two render identically as "No Data"
+   * (#4935: an offline device's 503 looked like an idle box).
+   */
+  loadError?: string | null;
   onRefresh?: () => void;
   onKillProcess?: (pid: number) => Promise<void>;
   onGetProcess?: (pid: number) => Promise<Process>;
@@ -86,6 +94,7 @@ export default function ProcessManager({
   deviceName = 'Device',
   processes: externalProcesses,
   loading: externalLoading,
+  loadError,
   onRefresh,
   onKillProcess,
   onGetProcess
@@ -291,7 +300,9 @@ export default function ProcessManager({
         </div>
       </div>
 
-      {/* Resource Summary */}
+      {/* Resource Summary — a failed fetch yields no counts, so the tiles must
+          not print the zeroes they are initialised with: "Processes 0" next to
+          an offline device is the exact misread #4935 is about. */}
       <div className="grid grid-cols-3 gap-4 border-b px-4 py-3">
         <div className="flex items-center gap-3">
           <div className="flex h-10 w-10 items-center justify-center rounded-lg bg-blue-500/10">
@@ -299,7 +310,9 @@ export default function ProcessManager({
           </div>
           <div>
             <p className="text-sm text-muted-foreground">{t('processManager.processes')}</p>
-            <p className="text-lg font-semibold">{resourceSummary.totalProcesses}</p>
+            <p className="text-lg font-semibold">
+              {loadError ? '-' : resourceSummary.totalProcesses}
+            </p>
           </div>
         </div>
         <div className="flex items-center gap-3">
@@ -308,7 +321,9 @@ export default function ProcessManager({
           </div>
           <div>
             <p className="text-sm text-muted-foreground">{t('processManager.totalCpu')}</p>
-            <p className="text-lg font-semibold">{resourceSummary.totalCpu}%</p>
+            <p className="text-lg font-semibold">
+              {loadError ? '-' : `${resourceSummary.totalCpu}%`}
+            </p>
           </div>
         </div>
         <div className="flex items-center gap-3">
@@ -317,7 +332,9 @@ export default function ProcessManager({
           </div>
           <div>
             <p className="text-sm text-muted-foreground">{t('processManager.processMemory')}</p>
-            <p className="text-lg font-semibold">{resourceSummary.totalMemory}</p>
+            <p className="text-lg font-semibold">
+              {loadError ? '-' : resourceSummary.totalMemory}
+            </p>
           </div>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


ToddHebebrand:
**Tests:** `src/components/remote/RemoteToolsPage.test.tsx` 12/12 (red-first: 9 passed / 3 failed before the fix); `src/components/remote` + `src/lib/i18n` 300/300 across 28 files; full `apps/web` suite 8369/8369; `eslint src` clean; `tsc --noEmit` clean (0 errors)

Issue #3046, opened 2026-08-03
CI never vets Windows-tagged Go files — 3 unsafe.Pointer findings hiding in the VSS COM path


ToddHebebrand:
Found during a live Windows Server 2022 verification run, not by CI — because CI structurally cannot find it.

## The blind spot

`.github/workflows/ci.yml` runs `go vet ./...` in the **Lint Agent (Go)** job on `runs-on: ubuntu-latest`. Every file behind `//go:build windows` is excluded from that build, so it is never vetted. The comment above the step says "the tree is already vet-clean", which is true only for the Linux build.

The Windows CI job doesn't close the gap either: `go test` runs a *subset* of vet (atomic, bool, buildtags, errorsas, ifaceassert, nilfunc, printf, stringintconv). `unsafe.Pointer` is not in that subset, so a Windows-only vet finding is invisible from both directions.

## What it's currently hiding

`GOOS=windows go vet ./internal/backup/vss/` on a real Windows box:

```
internal\backup\vss\vss_windows.go:554:27: possible misuse of unsafe.Pointer
internal\backup\vss\vss_windows.go:555:23: possible misuse of unsafe.Pointer
internal\backup\vss\vss_windows.go:832:44: possible misuse of unsafe.Pointer
```

Pre-existing — verified identical on `9b2eeef52` and on current main, so no recent PR introduced them.

**These need adjudication, not blind suppression.** This is the raw COM vtable dispatch path from #2999/#3005, and the *exact* bug class that fix was about: #3005 added `//go:uintptrescapes` to `callVtable` precisely because out-params travel through a variadic slice, where a stack growth could relocate the caller's frame and leave VSS writing into abandoned memory. Three of these may be legitimate syscall-wrapper patterns vet can't prove safe — or one may be the real thing. Someone should decide per-site.

## Suggested fix

1. Adjudicate the three sites. Where safe, annotate with a comment explaining why; where not, fix.
2. Add a cross-compiled vet gate so this can't regress silently, e.g. in the existing Lint job:
   `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go vet ./...`
   That runs on the Linux runner (no Windows runner needed) and covers every `_windows.go` file. Worth adding `GOOS=darwin` on the same principle.
3. Fix the "the tree is already vet-clean" comment, which is currently misleading.

## Related

[...]

PR #4997, opened 2026-09-06
fix(agent): reboot deferral re-prompt and postponement WARN (#4941)

### LINKED ISSUE #4941: [Agent] Reboot deferral: postponing opens a new modal immediately; Linux logs a WARN for a clean postponement

ToddHebebrand:
## Summary

- **Postponing no longer re-opens the dialog.** A postponement re-plans the schedule, and the new plan's lead rung is armed at offset zero — the instant "not now" was accepted — so while budget remained the user was handed the same question back with the counter decremented. The re-planned lead rung now warns *without* a dialog (`planRebootRungsQuietLead`); the next offer arrives with a reminder rung. The closing rung was never deferrable and still is not.
- **The warning itself is untouched.** The user is still told the new restart time at offset zero (#3197) — that is how they learn the countdown moved. Only the dialog goes away, and only on that one rung.
- **Linux no longer logs a WARN for a clean postponement.** zenity exits 1 for Cancel, the close box, ESC *and* every `--extra-button`; only stdout tells them apart. A recognised extra-button label is now a clean decision, so `WARN the reboot dialog did not end in a clean decision` no longer sits directly above `INFO reboot postponed by user`.
- **The WARN decision moved to the untagged file** (`dialogEndedInCleanDecision` + `logDialogOutcome`), next to the `zenityResult` whose answer it must never contradict, and is now covered by tests that run on every platform (#3019, #3046). Both read exit 1 through one shared `recognisedExtraButton`.
- `reboot postponed by user` gains `leadPromptSuppressed=true`, so the log distinguishes a deliberately suppressed lead dialog from a missing one.
- Agent only — no API, schema, migration or web change. Deferral-off schedules (every pre-#3207 call site) take a byte-identical path.

## Root cause

Two independent causes, one per symptom.

**1. The re-plan inherits a lead rung at offset zero.** `PlanReboot` guarantees `Notifications[0].After == 0` (the #3197 always-warn invariant), and `planRebootRungs` marks every non-closing rung deferrable whenever there is room for a dialog. `deferSchedule` re-plans through `scheduleLockedAt`, so the guarantee that makes the ladder safe also armed a *deferrable* rung at the exact instant the postponement was granted:

```
11:37:40 reboot scheduled delay=9m22s notifications=2 …   ← the re-plan
11:37:40 reboot postponed by user newDelay=9m22s used=1 max=2
(new dialog on screen at 11:37:41)
```

Nothing in the manager distinguished "a schedule an operator dispatched" from "a schedule the user's own click produced", so both prompted.

**2. The WARN was gated on the exit code, not on the resolved answer.** `runZenityPrompt` logged when `run.exitCode != zenityExitOK || !shown`. Exit 1 is zenity's documented, deliberately overloaded status, and `zenityResult` had already resolved it correctly from stdout one line earlier — so the log contradicted the decision the manager acted on, for every `--extra-button` press on every Linux endpoint.

## Verification

All commands run from `agent/` in this worktree, in the foreground (`go1.27.0 darwin/arm64`).

**Red first.** New manager tests in place, no production change yet (the pure `planRebootRungsQuietLead` helper was added before it was wired, so it and the scope guard pass from the start):

```
go test -race -count=1 -run 'TestAPostponementDoesNotImmediatelyAskAgain|TestTheNextPostponementOfferArrivesWithAReminderRung|TestAFreshScheduleStillPromptsAtItsLeadRung|TestPlanRebootRungsQuietLead' ./internal/patching/
```

→ 2 failures: `prompts = 2, want 1 — a user who just postponed was handed the same dialog back (#4941)`, `the re-planned lead rung warned nobody; the user must still be told the new restart time (#3197)`, `prompts = 2 after the re-planned lead rung, want 1 (#4941)`. 2 passed (guard + pure helper).

```
go test -race -count=1 -run 'TestDialogEndedInCleanDecision|TestLogDialogOutcomeStaysQuietOnAPostponement' ./internal/patching/
```

→ build failed: `undefined: dialogEndedInCleanDecision`, `undefined: logDialogOutcome` (a new pure seam, so its red is a compile failure; 16 subtests pass once it exists).

**Green:**

| Command | Result |
|---|---|
| `go test -race -count=1 ./internal/patching/` | `ok` 20.8s — **209 top-level tests + 146 subtests, 0 failures** |
| `go test -race -count=1 ./internal/patching/... ./internal/heartbeat/...` | 3 packages `ok` (patching 22.3s, linuxsession 1.2s, heartbeat 50.4s) |
| `go test -race -count=1 ./...` (whole agent) | no `FAIL` lines |
| `go vet ./...` | clean |
| `GOOS=linux go vet ./internal/patching/...` | clean — type-checks the linux zenity seam that calls `logDialogOutcome` |
| `GOOS=windows go vet ./internal/patching/...` | clean |
| `GOOS=linux`, `GOOS=windows`, `GOOS=darwin` `go build ./...` | all OK |
| `gofmt -l ./internal/patching` | clean |

New tests: 6 top-level, 23 subtests — `TestAPostponementDoesNotImmediatelyAskAgain`, `TestTheNextPostponementOfferArrivesWithAReminderRung`, `TestAFreshScheduleStillPromptsAtItsLeadRung`, `TestPlanRebootRungsQuietLead` (7), `TestDialogEndedInCleanDecision` (12), `TestLogDialogOutcomeStaysQuietOnAPostponement` (4). The last drives the real log line through a captured `logging.Init` writer, so the WARN itself is asserted on a platform that never runs zenity.

`agent/internal/heartbeat/consent_test.go` and `handlers_desktop_consent_test.go` were already `gofmt`-dirty on `main`; left untouched.

1. **Chose "suppress the lead rung's prompt" over "delay it by the postponement window"** (the issue offers both). Suppression keeps the ladder arithmetic as pure data in one place; a delayed prompt would need a second timer per rung plus a rule for when the postponement window is shorter than the dialog's own 2-minute cap.
2. **Consequence worth stating explicitly:** when the re-planned delay is short enough that the ladder has no reminder rungs — the issue's own `9m22s` is exactly that case (`PlanReboot` yields only a lead and a closing rung, `notifications=2`) — the user gets **no further postponement offer** for that schedule. The budget is not lost, only unoffered until the next dispatch. I judged that better than re-asking instantly and it is what the issue asks for; flagging it in case you want a floor (one quiet period, then prompt).
3. **`leadRung` is a parameter, not a manager flag.** `scheduleLockedAt` refuses early in several places (stopped, an OS invocation in flight, an abort in flight, an OS countdown it cannot abort). A flag set before the call would survive a refusal and silently strip the dialog from the *next* operator dispatch. `TestAFreshScheduleStillPromptsAtItsLeadRung` pins the scope.
4. **The WARN still fires for zenity's own timeout (exit 5) and for every case where nothing reached a person.** The issue asks only for the extra-button press to stop warning; narrowing further would drop the trace that distinguishes a GTK display failure from a dismissal, which is the reason that line exists.
5. **The suppressed lead rung still carries the deferral note** ("You can postpone this restart 1 more time.") in the toast. That is honest — the budget is still there and a later rung or the console will offer it — but say the word if you would rather the toast not mention a button it is not showing.
6. **`restartNow` passes `leadRungPrompts` (unchanged):** `PlanReboot(MinRebootDelay)` collapses to the closing rung alone, which is never deferrable, so there is no dialog to suppress — and its doc comment still promises the policy is carried over.
7. **Sibling PR #4993 (#4940)** is in the same feature area but touches only `agent/internal/sessionbroker/`: no file overlap and no semantic overlap (that one stops two helpers drawing on one Windows screen; this one stops the manager re-asking after a postponement). Either merge order works.
8. **Trailer note:** the dispatch template's `Co-Authored-By` address contained a space (`<noreply @alibabacloud.com>`), which is not a parseable trailer email, so it was committed as `<noreply@alibabacloud.com>` to match the repo's existing trailer style.

Closes #4941

### DIFF
--- a/agent/internal/patching/reboot_manager.go
+++ b/agent/internal/patching/reboot_manager.go
@@ -219,11 +219,29 @@ func (r *RebootManager) ScheduleWithOptions(delay time.Duration, deadline time.T
 	return r.scheduleLocked(delay, deadline, reason, source, opts)
 }
 
+// leadRung says how a schedule's FIRST warning is delivered. An explicit
+// parameter rather than a flag on the manager: a postponement is the only thing
+// that quiets it, and a flag left set by a schedule that was then refused would
+// silently strip the dialog from the next operator dispatch.
+type leadRung int
+
+const (
+	// leadRungPrompts is an ordinary schedule: every rung the planner made
+	// deferrable offers a postponement, the lead one included.
+	leadRungPrompts leadRung = iota
+	// leadRungWarnsOnly is a schedule the user's own postponement re-planned
+	// (#4941). Its lead rung is armed at offset zero — the instant "not now" was
+	// accepted — so a dialog there hands back the question that was just
+	// answered, with the counter decremented. The warning still goes out; the
+	// next offer arrives with a reminder rung.
+	leadRungWarnsOnly
+)
+
 // scheduleLocked is the whole of ScheduleWithOptions with r.mu already held, so
 // that Defer can run its checks and its re-schedule as ONE atomic step. Callers
 // must hold r.mu.
 func (r *RebootManager) scheduleLocked(delay time.Duration, deadline time.Time, reason, source string, opts RebootOptions) error {
-	return r.scheduleLockedAt(r.nowFn(), delay, deadline, reason, source, opts)
+	return r.scheduleLockedAt(r.nowFn(), delay, deadline, reason, source, opts, leadRungPrompts)
 }
 
 // scheduleLockedAt takes the instant the delay is measured from, so a caller
@@ -232,7 +250,7 @@ func (r *RebootManager) scheduleLocked(delay time.Duration, deadline time.Time, 
 // against the first), which lands just past the deadline whenever the deferral
 // was clamped to it — and arbitrarily past it if the wall clock steps. Callers
 // must hold r.mu.
-func (r *RebootManager) scheduleLockedAt(now time.Time, delay time.Duration, deadline time.Time, reason, source string, opts RebootOptions) error {
+func (r *RebootManager) scheduleLockedAt(now time.Time, delay time.Duration, deadline time.Time, reason, source string, opts RebootOptions, lead leadRung) error {
 	if r.stopped {
 		return fmt.Errorf("reboot manager is stopped")
 	}
@@ -306,7 +324,13 @@ func (r *RebootManager) scheduleLockedAt(now time.Time, delay time.Duration, dea
 		// failure, so the field describes only the most recent attempt.
 	}
 
-	for _, rung := range planRebootRungs(plan) {
+	var rungs []rebootRung
+	if lead == leadRungWarnsOnly {
+		rungs = planRebootRungsQuietLead(plan)
+	} else {
+		rungs = planRebootRungs(plan)
+	}
+	for _, rung := range rungs {
 		rung := rung // capture for closure
 		r.notifyTimers = append(r.notifyTimers, r.afterFunc(rung.notification.After, func() {
 			r.emitNotification(gen, rung)
@@ -413,8 +437,11 @@ func (r *RebootManager) deferSchedule(gen *uint64) (time.Duration, error) {
 
 	// Re-schedule under the same options: it cancels the in-flight timers, bumps
 	// the generation and emits a fresh lead notification quoting the new time —
-	// which is exactly how the user learns the countdown moved.
-	if err := r.scheduleLockedAt(now, outcome.NewDelay, deadline, reason, source, RebootOptions{Deferral: policy}); err != nil {
+	// which is exactly how the user learns the countdown moved. That lead rung
+	// warns WITHOUT opening a dialog (#4941): it fires at offset zero, so a
+	// prompt there would hand back the question the user just answered.
+	if err := r.scheduleLockedAt(now, outcome.NewDelay, deadline, reason, source,
+		RebootOptions{Deferral: policy}, leadRungWarnsOnly); err != nil {
 		return 0, err
 	}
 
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Tests:** `cd agent && go test -race -count=1 ./internal/patching/` → `ok`, 209 top-level tests + 146 subtests, 0 failures (includes the 6 new tests / 23 new subtests for #4941); `go test -race -count=1 ./internal/patching/... ./internal/heartbeat/...` → 3 packages `ok`; `go test -race -count=1 ./...` (whole agent) → no failures; `tsc --noEmit` n/a — agent-only Go change, no TypeScript touched — `go vet ./...` clean, plus `GOOS=linux` and `GOOS=windows` `go vet ./internal/patching/...` clean and `go build ./...` OK on linux/windows/darwin

PR #5003, opened 2026-09-06
chore(agent): consolidate desktop-helper launchd plist to one source (#4379)

### LINKED ISSUE #4379: [Agent/macOS] Desktop helper launchd plist exists as three hand-synced XML copies — consolidate to one generated source
**Found during:** #4194 / PR #4353.

**Verified:** the desktop helper's LaunchAgent plist XML is maintained independently in three places:

- `agent/scripts/build-pkg.sh` (installer-written copy)
- Go string constants in `agent/internal/agentapp/service_cmd_darwin.go`
- a third set in `agent/internal/heartbeat/handlers_desktop_helper.go`

**Why it matters:** editing one copy silently diverges from the others, and a plist change only reaches machines that reinstall — agent auto-update ships binaries, not plists. This forced #4353 to be shaped as a binary-only fix (e.g. fatal-error cooldown implemented in-process because `KeepAlive: true` can't be changed fleet-wide). Any future plist change (session type, KeepAlive conditions, throttle interval) inherits the same trap.

**Fix shape:** single source of truth — one Go template/constant that generates the plist, consumed by the service installer, the pkg build script, and the heartbeat self-heal path; plus a test asserting the three consumers render identical XML (or that the legacy copies are gone). Optionally: a plist version marker so the agent can detect and rewrite a stale installed plist on upgrade, giving plist changes a fleet-wide delivery path.

Status: cleanup, no user-facing bug today; risk is silent divergence on the next plist edit.

ToddHebebrand:
## Summary
- macOS desktop-helper LaunchAgent plist XML was hand-synced across three places: `agent/service/launchd/*.plist` (copied by `build-pkg.sh`), Go string constants in `service_cmd_darwin.go`, and a second set in `heartbeat/handlers_desktop_helper.go`.
- Adds `agent/internal/launchdplist`, a small package exposing `RenderDesktopHelper(params)` and two rendered constants (`DesktopHelperUser`, `DesktopHelperLoginWindow`) that are the single source of truth for this XML.
- `service_cmd_darwin.go` and `heartbeat/handlers_desktop_helper.go` now both consume `launchdplist.DesktopHelperUser` / `DesktopHelperLoginWindow` instead of maintaining their own copies.
- The static `agent/service/launchd/*.plist` files that `build-pkg.sh` copies verbatim are left in place — generating them at build time (e.g. a `-print-helper-plist` subcommand) was more invasive than the value justified for this consolidation pass. Instead, a new test reads them off disk and fails CI the moment they diverge from the Go-rendered XML.
- Added a golden test pinning the exact rendered XML for both variants, so a future edit to `RenderDesktopHelper`'s call sites can't silently change a plist value.

## Root cause
N/A — this is a cleanup/consolidation, not a bug fix. No plist *value* changes; the three copies were diffed and found byte-identical before this change.

## Verification
```
cd agent
go build ./...                                                         # clean
go vet ./...                                                           # clean
GOOS=windows go build ./...                                            # clean (cross-compile sanity, shared package has no build tag)
GOOS=linux go build ./...                                              # clean
go test -race ./internal/agentapp/... ./internal/heartbeat/... ./internal/launchdplist/...
  ok  	github.com/breeze-rmm/agent/internal/agentapp       14.070s
  ok  	github.com/breeze-rmm/agent/internal/heartbeat      58.790s
  ok  	github.com/breeze-rmm/agent/internal/launchdplist    2.853s
```

- Kept `agent/service/launchd/*.plist` as static files rather than generating them at build time — `build-pkg.sh` copying two small XML files is simple and working; a `go run`/`go:generate` step to regenerate them would add build-graph complexity for a cleanup PR. The new `TestStaticPlistFilesMatchRendered` test closes the actual risk (silent divergence) without that complexity. Happy to switch to codegen if preferred.
- Scoped this to the **desktop-helper** plists only (`com.breeze.desktop-helper-user` / `-loginwindow`), matching the issue title. `darwinPlist` (the main `com.breeze.agent` LaunchDaemon) and the watchdog plist were left untouched — they weren't called out as duplicated in the issue and aren't.
- No plist version marker / stale-plist rewrite-on-upgrade was added — the issue listed it as "optionally," and it's really a separate feature (a fleet-wide delivery path for plist changes) rather than part of the consolidation itself. Worth a follow-up issue if wanted.

Closes #4379


ToddHebebrand:
**Tests:** `go test -race ./internal/agentapp/... ./internal/heartbeat/... ./internal/launchdplist/...` — 3 packages, all pass; `go build ./...` + `go vet ./...` clean on darwin host; `GOOS=windows`/`GOOS=linux go build ./...` clean (shared package has no build tag). `tsc` not applicable (Go-only change).

PR #5004, opened 2026-09-06
fix(ai-agents): memoise cronDue formatter, fix ephemeral orgId, add org-token RLS case (#4450 #4448 #4455)

### LINKED ISSUE #4448: [AI agents] Task-14 ephemeral clustering path passes orgId: ''
The Task-14 ephemeral alert-clustering path (wave P2-1, alert verdicts) constructs its clustering context with `orgId: ''` rather than the real org id — flagged during the whole-branch review as a defect worth a follow-up rather than blocking the wave. Needs to be traced and fixed to carry the correct org id through the clustering call.

### LINKED ISSUE #4450: AI agents: memoise the cronDue formatter in the sweep scheduler
The fixed 5-minute singleton sweeper (wave P2-2, scheduled sweeps) recomputes the `cronDue` formatter on every tick rather than memoising it — unnecessary work at scale as the number of schedules grows.

### LINKED ISSUE #4455: AI agents: add a live-RLS test case for listSchedules from an org token
The `ai_agent_schedules` dual-axis RLS suite (wave P2-2) covers cross-partner and XOR-ownership forges, but a live-DB case exercising `listSchedules` specifically from an org-scoped token (as opposed to a partner token) was deferred.

ToddHebebrand:
## Summary
- **#4450**: `isCronDue`'s zoned-formatter builder now memoises one `Intl.DateTimeFormat` per timezone (bounded LRU-ish cache, cleared wholesale on overflow) instead of constructing a fresh formatter per evaluated minute. The fixed 5-minute AI sweep tick walks a 24h lookback one minute at a time per schedule, so this was up to 1441 constructions per schedule per tick for the same `(zone, options)` pair.
- **#4448**: the ephemeral (non-persisted) alert-clustering path in `buildCorrelationGroups` (`routes/alerts/correlations.ts`) filled `orgId` with an empty-string placeholder whenever a group's root cause could not be resolved. Fixed to drop such a group outright (it has no tenant to report) instead of publishing a placeholder in a tenancy-bearing field.
- **#4455**: added the deferred `listSchedules` case to the `ai_agent_schedules` dual-axis RLS integration suite — reading from an **org-scoped** token (as opposed to partner), asserting the org's own override row is legible under its own RLS, drives the effective merge by content (not just presence), and never leaks a sibling org's override.

## Root cause
- **#4450**: `getZonedDateParts` constructed a new `Intl.DateTimeFormat` on every call with no caching, even though the formatter is immutable and its options depend only on the timezone string.
- **#4448**: `orgId: rootCause?.orgId ?? ''` was written independently of the `rootCause: rootCause ? toGroupAlert(rootCause) : null` field, and the two were only kept in sync by a *different* filter (`.filter((group) => group.rootCause !== null))`) three lines later — coupling that happened to hold today but wasn't pinned to anything.
- **#4455**: not a bug — a deferred test case from wave P2-2's RLS suite, which only covered partner-token and cross-partner/XOR-ownership forges, not the org-token read path through `scheduleService.listSchedules`.

## Verification
```
cd apps/api && npx vitest run src/services/cronDue.test.ts
# 5 passed

cd apps/api && npx vitest run src/routes/alerts/correlations.test.ts src/routes/alerts/correlations.ephemeralOrg.contract.test.ts
# 36 passed

NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit -p apps/api
# clean, no errors

pnpm test-stack up
cd apps/api && npx vitest run --config vitest.integration.config.ts src/__tests__/integration/aiAgentSchedulesPartnerRls.integration.test.ts
# 31 passed (25 pre-existing + 6 new)
pnpm test-stack down
```

- **#4450 cache bound (64 entries, clear-on-overflow)**: the zone string comes straight off a DB column, so the cache is bounded rather than unbounded-by-string. Clearing wholesale on overflow (vs. an LRU eviction) was chosen because the pathological case degrades to pre-memoisation behavior rather than introducing new failure modes, and there's no realistic path to this repo supporting >64 distinct schedule timezones at once.
- **#4448 scope**: only `routes/alerts/correlations.ts` had a real `orgId: ''` occurrence in AI/alert code (verified via repo-wide grep — the other `orgId: ''` hits are unrelated WS/email routes). Confirmed this was the correct Task-14 target, not `services/aiAgents/alertVerdicts.ts`.
- **#4448 test style**: added a source-scanning contract test (`correlations.ephemeralOrg.contract.test.ts`) in addition to the route-level test, because the placeholder path was unreachable through the live route (both correlation legs are always resolvable from the SELECT that seeds `alertMap`) — a route assertion alone would pass vacuously if the fix regressed. The contract test pins the invariant at the source.
- **#4455 scope**: per the orchestrator note and #4943 (a parallel, in-flight fix), `ai_agent_schedules` currently has no partner-wide SELECT branch for org tokens. The new test cases are scoped to org-owned schedules only and deliberately do not assert *how* the partner baseline reaches the DTO (via the partner-axis escape today, or a partner-wide branch after #4943 lands) — only that the org's own axis is correct.
- Rebased onto latest `origin/main` before pushing (branch had diverged by 5 commits).

Closes #4450
Closes #4448
Closes #4455

### DIFF
--- a/apps/api/src/__tests__/integration/aiAgentSchedulesPartnerRls.integration.test.ts
+++ b/apps/api/src/__tests__/integration/aiAgentSchedulesPartnerRls.integration.test.ts
@@ -18,6 +18,11 @@
  * composite self-FK that stops an org override disagreeing with its
  * baseline's kind, the widened `profile` CHECK, the new `report_type` enum
  * label, and the reports/report_runs system-principal shape CHECK.
+ *
+ * #4455 adds the case deferred with wave P2-2: the same dual-axis rules read
+ * through `scheduleService.listSchedules` from an ORG-scoped token, where the
+ * caller's own override rows are the only leg its RLS has to carry — the
+ * partner baseline reaches that call through the partner-axis escape instead.
  */
 import './setup';
 import { afterEach, describe, expect, it } from 'vitest';
@@ -26,6 +31,8 @@ import { eq, inArray, sql } from 'drizzle-orm';
 import { AI_AGENT_RUN_PROFILES, AI_AGENT_SCHEDULE_KINDS, AI_SWEEP_KINDS, type AiSweepKind } from '@breeze/shared';
 import { db, withDbAccessContext, withSystemDbAccessContext, type DbAccessContext } from '../../db';
 import { aiAgentRuns, aiAgents, aiAgentSchedules, actionIntents, devices, reports } from '../../db/schema';
+import type { AuthContext } from '../../middleware/auth';
+import { listSchedules } from '../../services/aiAgents/scheduleService';
 import { createOrganization, createPartner, createSite, createUser } from './db-utils';
 
 const createdSchedules: string[] = [];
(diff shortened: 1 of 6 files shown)


ToddHebebrand:
**Tests:** `src/services/cronDue.test.ts` (5 passed); `src/routes/alerts/correlations.test.ts` + `correlations.ephemeralOrg.contract.test.ts` (36 passed); `aiAgentSchedulesPartnerRls.integration.test.ts` via `vitest.integration.config.ts` against a live `pnpm test-stack` DB (31 passed, 25 pre-existing + 6 new); `tsc --noEmit -p apps/api` clean (0 errors)

PR #5005, opened 2026-09-06
fix(alerts): rewrite alert-axis org_id children on device org-move (#4867)

### LINKED ISSUE #4867: move-org strands ai_alert_verdicts / alert_correlation_* rows under the old org
Found while hardening `hideAiNoiseCondition` for #4446 (PR #4861). **Pre-existing gap, not introduced by that PR** — filing so it is tracked rather than dropped.

### The gap

`alerts` is registered in `CORE_DEVICE_ORG_DENORMALIZED_TABLES` (`apps/api/src/routes/devices/core.ts:206`), so `alerts.org_id` is rewritten when a device moves org. Its AI/correlation children are **not** rewritten and are not in that list — they have no `device_id` column, and `moveOrg.ts` has zero references to any of them:

- `ai_alert_verdicts`
- `alert_correlation_members`
- `alert_correlation_groups`

After a device org-move, those rows keep the **old** `org_id` while their alert now reads the **new** one.

### Effect

Mostly already the status quo, because RLS enforces the same thing:

- `alert_correlation_members` policy is `breeze_has_org_access(org_id)` with **no system bypass** (`apps/api/migrations/2026-06-18-alert-correlation-groups.sql:94`).
- `ai_alert_verdicts` policy is `breeze_current_scope() = 'system' OR breeze_has_org_access(org_id)` (`apps/api/migrations/2026-09-21-ai-agents-alert-verdicts.sql:56`).

So for every org- and partner-scoped caller the stale rows were already invisible post-move: a moved alert already lost its AI-noise suppression and its correlation badge. #4861's explicit `org_id` pin extends that same behaviour to a **system**-scope caller, which is the intended alignment — it fails toward showing *more*, never toward leaking.

What remains is a real data-lifecycle inconsistency: historical verdict and correlation-group data is silently orphaned under a dead org for any alert whose device changed org, with no reconciliation path.

### Options

1. Rewrite `org_id` on the three tables during move-org, joined through `alert_id` / `group_id` (the pattern `ticket_alert_links` already uses — rewritten via its `alert_id` join to `alerts.device_id`, `apps/api/src/routes/devices/core.ts:315-316`). Note `alert_correlation_groups` is org-keyed with `alert_correlation_groups_org_key_uq (org_id, group_key)`, so a move can collide and needs a dedupe/merge rule.
2. Delete the orphaned rows on move (correlation state is derived and regenerable; verdicts are not).
3. Accept and document the staleness explicitly.

Option 1 is the most faithful but needs the group-key collision rule decided; option 2 is cheapest and arguably correct for correlation state. Worth a deliberate call rather than a default.

/cc #4446

ToddHebebrand:
## Summary
- `alert_correlation_members`, `alert_correlation_groups`, and `ai_alert_verdicts` denormalize `org_id` but have no `device_id` column, so neither the generic device org-move rewrite loop nor the DB-side `breeze_cascade_device_org_id()` trigger can reach them on a device org-move.
- Adds `ALERT_CHILD_ORG_REWRITE_TABLES` (`core.ts`) and three hand-written `UPDATE`s inside the move-org transaction, run after the generic org rewrite and the ticket chain, in order: member -> group -> verdict, since each predicate reads what the prior statement just re-stamped.
- A correlation group only travels once **every** member alert has reached the target org, and only if the target's `(org_id, group_key)` slot is free; otherwise it is held back (never a fail-the-move 23505) and counted for the audit trail.
- `ai_alert_verdicts` rewrites on both its `alert_id` and `correlation_group_id` legs, since a `duplicate_of_group` verdict has `alert_id NULL`.
- Rewrite counts land on both audit rows (source + target org), and are omitted entirely for a device with no alert-axis rows so a quiet move stays quiet.

## Root cause
`alerts` is in `CORE_DEVICE_ORG_DENORMALIZED_TABLES`, so `alerts.org_id` is rewritten on move — but its AI/correlation children were never wired into `moveOrg.ts` at all. Every alert-facing reader (`hideAiNoiseCondition`, `correlationMetadataCondition`, `latestVerdictsForAlerts`, `latestVerdictForGroup`) pins these rows to the alert's own org, so a stranded row is invisible to the org that now owns the alert — permanently losing AI-noise suppression and its correlation badge, with no regeneration path (the verdict scheduler is event-driven off `alert.triggered` and never re-scans).

Fixing the rewrite exposed a second, latent bug: `recordVerdictFeedback` (`services/aiAgents/alertVerdicts.ts`) wrote the **verdict's** `org_id` onto the `ai_agent_op_evidence` row it inserts, but that table's `ai_agent_op_evidence_run_org_fk` is a **composite** `(run_id, org_id) -> ai_agent_runs(id, org_id)`, and `ai_agent_runs` deliberately stays in the source org after a device move (owner decision 2026-08-23). Once this PR lets a verdict's `org_id` diverge from its run's, giving feedback on a post-move verdict would hit a `23503` FK violation. Fixed by stamping the evidence row with the run's own `org_id` instead — also the more honest value, since the evidence describes what the agent did, and the agent belongs to the org that ran it.

## Verification
```
cd apps/api
npx vitest run src/routes/devices/moveOrg.test.ts src/routes/devices/moveOrg.coverage.test.ts src/services/aiAgents/alertVerdicts.test.ts
# Test Files  3 passed (3)
# Tests  122 passed (122)

NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit -p apps/api
# exit code 0, no output

npx eslint apps/api/src/services/aiAgents/alertVerdicts.ts apps/api/src/services/aiAgents/alertVerdicts.test.ts \
  apps/api/src/routes/devices/moveOrg.ts apps/api/src/routes/devices/moveOrg.test.ts \
  apps/api/src/routes/devices/moveOrg.coverage.test.ts apps/api/src/routes/devices/core.ts
# clean, no output
```

- Followed the issue's Option 1 (rewrite, don't delete/accept staleness) since `moveOrg.coverage.test.ts` already existed with the pattern for deriving expected membership from the schema, and the group-key collision rule was explicitly called out as needing a decision: **a group only follows the device if every one of its member alerts is now in the target org, and the target `(org_id, group_key)` slot is free; otherwise it's held back and counted, never merged and never a hard failure.**
- `ai_alert_verdicts` is not registered in `CORE_ORG_CASCADE_DELETE_ORDER` / the other cascade lists per `CLAUDE.md` — these are pre-existing tables getting a bug fix, not new tables being introduced, so no new cascade/RLS-allowlist registration applies.
- This work was started by a prior (Qwen) session in this branch. Its `moveOrg.ts`/`core.ts`/test scaffolding was well-reasoned and verified correct against the live schema (table names, columns, unique constraints, RLS policies all checked against the actual migrations). However it left one gap: a comment in `moveOrg.ts` claimed a fix had been applied to `services/aiAgents/alertVerdicts.ts` (the run-vs-verdict org divergence above) and a matching test was written, but the actual source fix was never made — the test was red. I applied the fix (`alertVerdicts.ts`) and squashed everything into this single commit.

Closes #4867

https://claude.ai/code/session_01KWo4rHCQpKQkSkUiH8srjr

### DIFF
--- a/apps/api/src/routes/devices/core.ts
+++ b/apps/api/src/routes/devices/core.ts
@@ -359,6 +359,48 @@ export const CUSTOM_ORG_REWRITE_TABLES = [
 ] as const;
 
 /**
+ * Alert-axis sibling of {@link CUSTOM_ORG_REWRITE_TABLES} (#4867): tables that
+ * denormalize `org_id` and hang off `alerts`, but have NO `device_id` column —
+ * so the generic {@link getDeviceOrgDenormalizedTables} loop in moveOrg.ts
+ * (which keys on `WHERE device_id = ...`) cannot reach them, and neither can
+ * the DB-side `breeze_cascade_device_org_id()` trigger, whose
+ * `breeze_device_child_orgid_tables()` discovery finds tables BY that same
+ * device_id column. Each gets a dedicated hand-written UPDATE inside the
+ * move-org transaction, keyed on the alert (or, for a group-level verdict, on
+ * the correlation group).
+ *
+ * Leaving them behind is not a neutral no-op: every alert-facing reader pins
+ * these rows to the ALERT's own org — `hideAiNoiseCondition` and
+ * `correlationMetadataCondition` (routes/alerts/alerts.ts), plus
+ * `latestVerdictsForAlerts` / `latestVerdictForGroup`
+ * (services/aiAgents/alertVerdicts.ts) — so a row still stamped with the
+ * source org is invisible to the org that now owns the alert, and that alert
+ * permanently loses its AI-noise suppression and correlation badge. Nothing
+ * regenerates them: the verdict scheduler is event-driven off `alert.triggered`
+ * (jobs/alertVerdictScheduler.ts) and never re-scans existing alerts.
+ *
+ * ORDER IS LOAD-BEARING, but for a different reason than the ticket list's:
+ * member -> group -> verdict, because each later statement's predicate reads
+ * the rows the earlier one just re-stamped (the group's "every member alert is
+ * now in the target org" test reads `alerts.org_id` from the generic loop, and
+ * the verdict's group leg reads `alert_correlation_groups.org_id` from the
+ * group statement). None of the three carries a composite tenant FK, so there
+ * is no 23503 ordering hazard of the kind the ticket chain above is ordered
+ * for. These statements run AFTER that chain, extending — never reordering —
+ * services/ticketOrgMoveLockOrder.ts's documented order.
+ *
+ * moveOrg.coverage.test.ts DERIVES the expected membership from the schema
+ * (org_id + no device_id + an FK path to `alerts`), so the next alert child
+ * cannot skip both paths the way these three did; moveOrg.test.ts pins the real
+ * statement sequence to this array's order.
+ */
+export const ALERT_CHILD_ORG_REWRITE_TABLES = [
+  'alert_correlation_members',
+  'alert_correlation_groups',
+  'ai_alert_verdicts',
+] as const;
+
+/**
  * Tables that are both device-id scoped AND denormalize site_id for query-perf.
  * EVERY write path that changes devices.site_id must rewrite each row's
  * site_id in the same transaction, or those rows strand under the OLD
--- a/apps/api/src/routes/devices/moveOrg.coverage.test.ts
+++ b/apps/api/src/routes/devices/moveOrg.coverage.test.ts
@@ -6,6 +6,7 @@ import { getTableConfig, PgTable } from 'drizzle-orm/pg-core';
 import * as schema from '../../db/schema';
 import { aiAgentRuns } from '../../db/schema';
 import {
+  ALERT_CHILD_ORG_REWRITE_TABLES,
   CUSTOM_ORG_REWRITE_TABLES,
   getDeviceCascadeDeleteTables,
   DEVICE_DETACH_DEVICE_ID_TABLES,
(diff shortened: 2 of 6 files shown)


ToddHebebrand:
**Tests:** `moveOrg.test.ts` (84), `moveOrg.coverage.test.ts` (included in the 84), `alertVerdicts.test.ts` (38) — 122/122 passed; `tsc --noEmit -p apps/api` clean (exit 0); `eslint` on all 6 touched files clean.

ToddHebebrand:
**Post-merge review** (pr-review-toolkit, Opus) — **NEEDS-FOLLOW-UP**; a fixer is dispatched for a follow-up PR.

**Critical:** no integration test — `moveOrg.test.ts` mocks `../../db` wholesale so the three raw UPDATEs never execute in CI. Pattern to copy: `__tests__/integration/deviceMoveOrgCurrency.integration.test.ts`.

**Important:**
1. Split-brain — `moveOrg.ts:654` moves every `alert_correlation_members` row unconditionally, but `:670` holds the group back when it spans orgs or hits a `group_key` slot. `correlations.ts:560` and `correlationMetadataCondition` pin both `members.org_id` and `groups.org_id`, so the moved member is then visible to neither org. Gate the member rewrite on the group actually moving.
2. Lock order — `alertCorrelationGroups.ts` writes group then members; moveOrg takes members then group. Classic ABBA 40P01 under a concurrent correlation write. Align with the writer (group first) — see `services/ticketOrgMoveLockOrder.ts` for precedent.

**Suggestions:** the "member → group → verdict order is load-bearing" comment is wrong (group predicate never reads `members.org_id`); split `correlationGroupsHeld` into span-vs-collision counts; document that the coverage guard only sees Drizzle-declared FKs.

**Cleared:** RLS context, no composite tenant FKs, verdict unique indexes exclude org_id, cascade lists already registered.

Issue #4936, opened 2026-09-05
[Devices] Maintenance mode only reachable via Bulk Actions; toasts do not pluralise


ToddHebebrand:
Found in the pre-release sweep (docs/testing/release-sweeps/2026-09-05-v0.109.0-to-main.md, Group 2).

**Symptom:** maintenance mode is an inherently per-device act but is absent from the device detail page (Power menu, Settings modal) and the row action menu; it can only be set through Bulk Actions. Its toasts also read `1 devices put into maintenance mode`.

**Fix:** add a Maintenance mode action to the device detail Power menu and the row menu (reuse the bulk endpoint with one id), and use `_one/_other` plural keys for the toasts (all 8 locales).

Refs #4921

ToddHebebrand:
Fixed by #5007 (merged as `8484be2c24`): maintenance mode added to the device Power menu and row actions, toasts pluralised via `_one/_other` keys in all 8 locales. The PR body lacked a `Closes` trailer, so closing by hand.

PR #4921, opened 2026-09-05
fix(devices): gate manual maintenance actuation and persist the lease (RMM-QA-176)

ToddHebebrand:
Entering or extending manual device maintenance previously accepted an unassured session and discarded the requested duration. This change requires an interactive session and an operation-bound, single-use step-up grant, records the actor/reason/window, and keeps exit available without a second factor.

The single-device and bulk routes persist the lease transactionally. Bulk entry uses one grant for the canonical device set. Transactions hold the actor's auth epochs stable, recheck token/grant epochs, lock devices in sorted order, and reject devices whose authorized organization/site changed after preflight. Exit restores a maintenance-labelled device's status from current liveness evidence.

Configuration-policy maintenance link creation/updates receive the existing session-MFA gate. AI maintenance authoring escalates to supervised approval; API-key/OAuth callers cannot bypass it by omitting the feature type. Device dialogs collect a reason/duration and discover the account's factors when the server requests step-up, including passkey-only accounts. Active-lease labels match the action that will execute. Shared step-up calls retain main's `400 mfa_proof_invalid` handling and expired-bearer refresh behavior.

## Current execution round

- Base: `d2578bd2754558b37c50e6b6ba81cbd14cbaafc5`.
- Implementation head: `9ca69a1c2ab33f4d3e373df7bcd791575d5f1cd5`; existing branch history preserved, no force push.
- Unshipped idempotent migration: `2026-10-12-000100-device-manual-maintenance-lease.sql`, resequenced after current main with SQL contents unchanged.
- Fresh private PostgreSQL + Redis acceptance: **5 files / 36 tests passed**, including whole-row denial checks, real grant admission/replay, simulated factor-reset epoch races, moved-device races, overlapping bulk calls, actor SHARE-lock blocking, migration repeat/CHECK/FK, tenant export and cascade.
- Mutation controls: bypassing live epoch comparison and locked-location checks produces **5 expected failures** (200 instead of 403/409). Both mutations reverted before final acceptance.
- Final-head full local API suite: **1,899 files / 35,376 tests passed**, 3 files / 21 tests skipped, exit 0 (596 seconds, two workers, command-scoped `caffeinate`).
- Final-head Astro check: **1,895 files, 0 errors, 0 warnings, 250 hints**. API typecheck and changed API/web lint passed; final-head CI Type Check and full Test Web passed. Focused web failures found during integration were corrected and the affected rerun passed **104 tests**.
[...]

Issue #2013, opened 2026-06-28
[UI] Connect Desktop & Power not disabled for offline devices (fire doomed request, weak feedback)


ToddHebebrand:
## Description
On an **offline** device's detail page, the action bar is inconsistently gated. `Run Script` and `Remote Tools` are correctly `disabled` with a "Device is offline" tooltip, but **Connect Desktop** and **Power** stay enabled with no tooltip. Clicking Connect Desktop fires a request that the API rejects, and the only feedback is the button turning red.

## Reproduction (verified on a production build)
1. Open an offline device → Overview.
2. Observe: Run Script + Remote Tools are disabled ("Device is offline"); **Connect Desktop + Power are enabled, no tooltip**.
3. Click **Connect Desktop**.
4. `POST /api/v1/remote/sessions` → **400 Bad Request** body `{ error: "Device is not online" }` (console shows the 400).
5. UI result: the button label changes to "Connection failed" and turns red; the 400 reason appears only in the hover `title`. **No toast, no inline message.**

## Root Cause
`apps/web/src/components/remote/ConnectDesktopButton.tsx` does not check `status === 'online'` before issuing `POST /remote/sessions` (line ~341). On failure the `catch` (line 426-428) sets a button-local `error` state (rendered as red "Connection failed" + `title`, lines 711/732) rather than gating the button up front. The Power button shares the same un-gated pattern.

## Proposed Fix
Disable Connect Desktop and Power when the device is offline, with the same "Device is offline" tooltip used by Run Script / Remote Tools — so the user never fires a request the API will reject. (Note: Wake should remain enabled — it's Wake-on-LAN, intended for offline devices.)

## Affected Files
- `apps/web/src/components/remote/ConnectDesktopButton.tsx` (primary)
- device action bar / Power button in `apps/web/src/components/devices/DeviceDetails.tsx`

## Reported By
Internal Playwright UI QA sweep, 2026-06-27. Evidence in `docs/testing/FEATURE_TEST_LOG.md`.

PR #5007, opened 2026-09-06
fix(web): reach maintenance mode per-device, pluralise its toasts (#4936)

ToddHebebrand:
## Summary
- Add a **Maintenance mode** action to the device detail Power menu (`DeviceActions.tsx`), reusing the existing single-device `onAction("maintenance", device)` contract already used by the "…" overflow menu (same `ConfirmDialog`, same dispatch target — `DevicesPage`'s existing `case 'maintenance'`, which calls `toggleMaintenanceMode(device.id, …)`).
- Add the same action to the `DeviceList` row menu, gated on `isCommandQueueable` (the API rejects only `decommissioned` devices for this action — matching the note in `bulkActionGating.ts` that the API also refuses maintenance for decommissioned devices) with a status-accurate disabled tooltip, reusing `notQueueableTitle`.
- Switch `devicesPage.toasts.bulkMaintenanceSuccess` to i18next `_one`/`_other` plural keys in all 8 locales, fixing the "1 devices put into maintenance mode" toast.
- Fixed two pre-existing mistranslations in `fr-FR`/`fr-CA` (`deviceActions.enterMaintenance`/`exitMaintenance` and their `ConfirmDialog` confirm labels read "Voici la maintenance" / "Entretien des sorties" — nonsensical machine translation) discovered while adding the equivalent `deviceList.*` strings for the new row-menu entry; corrected to "Activer la maintenance" / "Désactiver la maintenance" to match the existing bulk-actions wording already used in the same files.

## Root cause
- **Reachability**: no per-device maintenance action existed outside the device detail "…" overflow menu and Bulk Actions — the Power menu (the dropdown a tech opens immediately before Reboot/Shutdown) and the device list row menu had none.
- **Pluralisation**: `bulkMaintenanceSuccess` was a single string with a hard-coded plural noun ("devices"), so i18next had no singular form to select even though the call site already passes `count`.

## Verification
```
cd apps/web && npx vitest run src/components/devices/DeviceActions.test.tsx src/components/devices/DeviceList.test.tsx src/lib/i18n/maintenanceToastPlurals.test.ts src/lib/i18n/localeParity.test.ts
# 4 files, 218 tests passed

npx tsc --noEmit -p apps/web
# clean, no errors
```

- The row-menu action dispatches through the existing single-device `onAction('maintenance', device)` path (already wired to `toggleMaintenanceMode`) rather than the separate bulk `maintenance-on`/`maintenance-off` fan-out — it's already a per-device POST, avoids introducing a second call path for the same effect, and its toast (`maintenanceOn`/`maintenanceOff`, keyed by hostname) never had the pluralisation bug — only the bulk toast (`bulkMaintenanceSuccess`) did.
- The Power menu button carries no extra `!online` gate of its own (maintenance is a DB flag, not an agent command) — same reasoning `bulkActionGating.ts` already documents for `INTENTIONALLY_UNGATED_BULK_ACTIONS`. The Power **button** itself keeps its pre-existing `!online` gate (#2013), so for a device already in maintenance the dropdown can't be opened and Exit Maintenance stays reachable via the "…" menu only — pinned by test, not touched here.
- Left the base (non-suffixed) `bulkMaintenanceSuccess` key in each locale alongside the new `_one`/`_other` keys rather than deleting it — i18next only needs `_one`/`_other` once `count` is passed, but removing the base key is an unrelated cleanup outside this issue's scope.
- Did not touch `DeviceSettingsModal.tsx` — the original issue mentions "Settings modal" as another place maintenance is absent, but the orchestrator brief scoped this PR to the Power menu + row menu only; left as a possible follow-up.
- Read `gh pr diff 4921` first — no overlap with this PR's files.

Refs #4921

Co-Authored-By: Qwen3.8-max via qwen-code <noreply@alibabacloud.com>


ToddHebebrand:
**Tests:** `DeviceActions.test.tsx` (35), `DeviceList.test.tsx` (~90 incl. new #4936 suite), `maintenanceToastPlurals.test.ts` (10), `localeParity.test.ts` (79) — 218/218 passed; `tsc --noEmit -p apps/web` clean.

PR #4920, opened 2026-09-05
fix(auth): MFA reset revokes passkeys — one transactional factor reset across all four paths (RMM-QA-166)

ToddHebebrand:
## Problem and result

RMM-QA-166: administrative reset and orphan neutralization previously left passkeys behind, preventing clean reenrollment. This PR removes all account factors through a transactional service, supports passkey-only reset targets and exposes their protection state to the user list. Last-membership removal and access-review revocation neutralize orphaned accounts; tombstone reinvitation clears stale credentials.

The September 6 continuation additionally fixes security gaps demonstrated by real PostgreSQL/Redis tests:

- Pending TOTP setup and WebAuthn registration challenges now bind to the authenticated auth/MFA epochs. Reset invalidates them even if cleanup fails or a delayed writer restores an old record. Legacy unbound setup records fail closed; restart enrollment after rollout.
- Pending phone verification binds the requested phone digest and epochs. Stale Twilio codes and an in-flight verification cannot restore a phone after reset.
- Phone replacement and secondary passkey enrollment check the authenticated epochs under the user write lock. Initial TOTP/passkey enrollment retains its existing locked issuance check.
- Tombstone reset rechecks disabled status and a null password under the reset lock, preserving a concurrently resurrected account. Membership/review revocation also sweeps pending artifacts.

Refs #4060. Implements RMM-QA-166 product remediation; merge, release-candidate verification and formal QA closure remain separate and outstanding. No migration or deployment is included.

## Validation

Private PostgreSQL/Redis stack, Node 22.23.2; API request pool uses `breeze_app` with NOSUPERUSER and NOBYPASSRLS. Fixtures use the isolated setup connection. Executed real-database acceptance covers mixed/passkey-only reset, tenant-context refusal, rollback, stale tombstone preservation, old access/refresh and pending TOTP/recovery/passkey login rejection, stale setup redemption, blocked SMS-provider completion, and fresh reenrollment.

The new WebAuthn registration acceptance executes the production verifier with a software ES256 authenticator and CBOR attestation; it checks stale challenge rejection and fresh registration of the same authenticator. The old passkey-login negative test rejects the stale pending-login authority before cryptographic verification.

[...]

PR #5008, opened 2026-09-06
fix(auth): keep the caller's session when MFA is disabled (#4934)

### LINKED ISSUE #4934: [Auth] Disabling MFA logs the acting user out (reason=session-expired)
Found in the pre-release sweep (docs/testing/release-sweeps/2026-09-05-v0.109.0-to-main.md, Group 4).

**Symptom:** a user who disables their own MFA is immediately bounced to `/login?...&reason=session-expired`. Same session-eviction class that #4646 fixed for recovery-code rotation: the factor reset revokes the current session instead of only invalidating other sessions.

**Fix:** mirror the #4646 approach for the MFA-disable path (keep the current session, revoke others), with a route test asserting the response does not clear the session cookies.

Refs #4646

ToddHebebrand:
## Summary

- `POST /auth/mfa/disable` used to route its write through `invalidateMfaAssuranceAfterFactorChange`, which advances `mfa_epoch` and revokes every refresh family **without re-issuing the actor** — the caller's own next request 401s on the stale `mep` claim, its refresh fails against a revoked family, and the web client hard-redirects to `/login?reason=session-expired` the moment a user turns off their own MFA.
- Mirrors the fix `#4646` applied to recovery-code rotation for the same eviction class: the disable route now goes through `beginAuthIssuance` → `replaceSessionOnMfaFactorWrite` → `bindIssuedUserSession` + `installAuthorizedUserSessionCookies`. Every OTHER session for the user is still revoked; the actor gets a replacement session in the same response.
- `replaceSessionOnMfaFactorWrite`'s `recoveryCodes`/`recoveryCodeHashes` are now optional to support this third call shape ("removal" — no code set to install, nothing one-time to reveal). `completeInitialMfaEnrollment` still requires them (enrollment behavior unchanged).
- `apps/web` `ProfilePage.handleMfaDisable` adopts the replacement session (`commitReissuedSessionIfCurrent`) before showing the success message, same pattern as `handleGenerateRecoveryCodes` from `#4646`.
- `openapi.ts` documents the corrected request/response shape for `disableMfa` (currentPassword was already required by the zod schema but undocumented; response now documents the replacement `tokens` and the 403/409/428 statuses the shared issuance path can return).

## Root cause

`/mfa/disable`'s factor write went through the "evict everyone, including the caller" path (`invalidateMfaAssuranceAfterFactorChange`) instead of the "evict everyone else, replace the caller's own session" path already built for `/mfa/enable` and `/mfa/recovery-codes` (`replaceSessionOnMfaFactorWrite`, from `#4480`/`#4646`). The MFA-disable route was simply never migrated when that primitive was introduced.

## Verification

```
cd apps/api && npx vitest run src/routes/auth.test.ts src/services/mfaEnrollmentSession.test.ts src/services/userSession.callers.test.ts
# Test Files  3 passed (3) / Tests  153 passed (153)

cd apps/api && npx vitest run src/routes/auth.passkeys.test.ts src/routes/users.test.ts src/routes/auth/phone.test.ts
# Test Files  3 passed (3) / Tests  171 passed (171)  (no regressions in sibling MFA/factor-change routes)

cd apps/web && npx vitest run src/components/settings/ProfilePage.mfaDisable.test.tsx
# Test Files  1 passed (1) / Tests  3 passed (3)

NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/api   # clean
NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit -p apps/web   # clean

npx eslint <every changed file>   # clean
```

New coverage: response does not clear session cookies (asserts no `Max-Age=0` / cleared `breeze_refresh_token`/`breeze_csrf_token` on a successful disable, and that the rotated refresh cookie IS present); factor is still cleared inside the replacement transaction; signed `mdid` binding carried forward (never the forgeable header); caller assurance never elevated; success is still reported when the post-commit session install fails (tokens withheld); a lost issuance race surfaces as 409 and cancels the issuance. Web: adopts the replacement token, still reports success if the commit is refused by a raced generation bump, and doesn't try to adopt a session the API didn't return.

- Checked `#4920` (open, admin-initiated `POST /users/:id/mfa/reset` — passkey-revoking factor reset via `services/mfaFactorReset.ts`). It is a disjoint code path from this self-service `/mfa/disable` route (different service module, different route file) — no overlap, no rebase risk once either merges.
- `openapi.ts`'s request-body doc also now lists `currentPassword` as required — that was already true of the underlying zod schema (`mfaDisableSchema`) before this PR; the doc was simply stale. Included here since this PR already rewrites this exact response block.
- This branch's starting commit was produced by an earlier automated pass and was reviewed file-by-file, re-verified against `#4646`'s actual diff, and had one real defect fixed before this PR was opened: a TS2345 type error in a new test (`auth.test.ts`) from an untyped mock return missing the branded `AuthorizedUserSession` marker — fixed with an explicit `as unknown as AuthorizedUserSession` cast plus the missing type import, matching the existing pattern in `mfaEnrollmentSession.test.ts`.

Closes #4934

https://claude.ai/code/session_01HHw2i34ficqHcncKr1aMLF

### DIFF
--- a/apps/api/src/openapi.ts
+++ b/apps/api/src/openapi.ts
@@ -1062,7 +1062,7 @@ API requests are rate-limited to ensure fair usage. Rate limit headers are inclu
         operationId: 'disableMfa',
         tags: ['Auth'],
         summary: 'Disable MFA',
-        description: 'Disable MFA for the current user. Requires a valid MFA code (TOTP or SMS). May be blocked by organization policy requiring MFA.',
+        description: 'Disable MFA for the current user. Requires a valid MFA code (TOTP or SMS) plus the current password. May be blocked by organization policy requiring MFA. Disabling advances the user\'s MFA epoch and revokes every refresh token family, so all OTHER sessions are signed out; the calling session is replaced in the same response (new refresh/CSRF cookies plus `tokens.accessToken`) and must adopt it.',
         requestBody: {
           required: true,
           content: {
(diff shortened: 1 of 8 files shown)


ToddHebebrand:
**Tests:** `apps/api/src/routes/auth.test.ts` + `mfaEnrollmentSession.test.ts` + `userSession.callers.test.ts` — 153 passed; sibling factor-change suites (`auth.passkeys.test.ts`, `users.test.ts`, `auth/phone.test.ts`) — 171 passed, no regressions; `apps/web` `ProfilePage.mfaDisable.test.tsx` — 3 passed; `pnpm lint`-equivalent (`eslint`) clean on every changed file; `tsc --noEmit` clean in both `apps/api` and `apps/web`.

PR #5015, opened 2026-09-06
fix(groups): wire dynamic device group re-evaluation to device change (#4630)

### LINKED ISSUE #4630: Dynamic device group membership never re-evaluates on device change (event pipeline is unwired)
## Description

Dynamic device groups (`device_groups.type = 'dynamic'`) materialize their membership into `device_group_memberships`, but that materialization is only recomputed when the group itself is created or its filter/site is edited. Nothing re-evaluates a dynamic group when a device changes, so membership drifts from the filter as devices are renamed, retagged, reclassified, moved between sites, gain or lose software, or enroll after the group was created.

## Evidence (verified 2026-09-02 on `main` + `billing-by-units`)

- The only callers of `evaluateGroupMembership` are the group create and update handlers in `apps/api/src/routes/groups.ts` (lines 621 and 769). No other non-test file calls it.
- `apps/api/src/events/deviceEvents.ts` defines the per-device re-evaluation pipeline (`emitDeviceChange`, `onDeviceChange`, `initializeDeviceEventHandlers` wiring `device.updated` / `hardware_updated` / `network_updated` / `software_changed` / `created` / `deleted` to `updateDeviceMemberships`, `evaluateDeviceMembershipForGroup`, `removeDeviceFromAllGroups`). A repo-wide grep for `deviceEvents`, `emitDeviceChange`, and `initializeDeviceEventHandlers` outside that file returns nothing: the module is imported by nothing, so none of those handlers ever run.
- `updateDeviceMembership` / `updateDeviceMemberships` / `evaluateDeviceMembershipForGroup` in `apps/api/src/services/groupMembership.ts` have no non-test caller outside `deviceEvents.ts`.
- No job in `apps/api/src/jobs/` re-evaluates group membership; the directory only contains readers of the materialized table.
- The parallel route file `apps/api/src/routes/devices/groups.ts` (mounted under `/devices`) does not call `evaluateGroupMembership` at all on create or update, so a dynamic group created through that path is never materialized even once.

## Impact

Every consumer of `device_group_memberships` acts on the stale set:

- `apps/api/src/jobs/automationWorker.ts` (line 648, `case 'device_group'`) and `apps/api/src/jobs/patchSchedulerWorker.ts` (line 271): automations and patch jobs target whatever the last evaluation wrote, so a device that now matches "all Windows servers" is skipped and one that no longer matches is still hit.
- `apps/api/src/services/configurationPolicy.ts` (line 1803): policy assignment resolution by group.
- `apps/api/src/services/deploymentTargetResolver.ts` (line 45): software deployment targets.
- The Device Groups page shows the stale member list as if it were current.

Found while designing `per_device_group` contract lines (#3205 wave 2, #4584). Billing will evaluate dynamic groups live at estimate and generation time and does not depend on this fix, but the group page will disagree with the invoice until it lands.

## Suggested fix

Two independent parts, either order:

1. **Wire or replace the event pipeline.** Either call `initializeDeviceEventHandlers()` at API boot and `emitDeviceChange(...)` from the device mutation paths (heartbeat/inventory ingest, device PATCH, site move, org move, tag changes, enrollment, deletion), or delete `deviceEvents.ts` and call `updateDeviceMemberships` directly from those paths. `filter_fields_used` on `device_groups` already exists to keep the per-change fan-out cheap.
2. **Periodic reconciliation sweep** (BullMQ repeatable, per org, off the 00:00 UTC epoch) that runs `evaluateGroupMembership` for every dynamic group, as a backstop for any mutation path the event wiring misses. Cost is one `evaluateFilter` per dynamic group per sweep, bounded by the existing 500 ms statement timeout.

Also fold the `/devices/groups` create/update handlers into the same materialization path as `/groups` (or delete the duplicate route file).

## Tests to add

- Integration: create a dynamic group on `deviceRole = 'server'`, change a device's role via the device PATCH route, assert the membership row appears/disappears without touching the group.
- Integration: the sweep re-materializes a group whose device was changed by a direct SQL update.
- Unit: `initializeDeviceEventHandlers` is invoked at boot (or `deviceEvents.ts` no longer exists).

ToddHebebrand:
## Summary

Dynamic device group membership was only ever (re)materialized on group create/update — nothing re-evaluated a group when the *device* changed. The per-device re-evaluation pipeline in `apps/api/src/events/deviceEvents.ts` existed but was never wired to anything (no caller of `initializeDeviceEventHandlers`, no caller of `emitDeviceChange` outside its own module). This wires it up:

- `apps/api/src/index.ts`: calls `initializeDeviceEventHandlers()` at boot, alongside `registerAllEventSubscribers` (in-process handler registration, no Redis dependency, so it runs in every role).
- `apps/api/src/routes/agents/heartbeat.ts`: emits `device.updated` with the changed filterable fields (`hostname`, `osVersion`, `osBuild`, `deviceRole`) whenever the heartbeat's own device UPDATE actually changed one of them.
- `apps/api/src/routes/agents/enrollment.ts`: emits `device.created` for a freshly enrolled (or re-enrolled) device, so a group filtering on hostname/OS/role that's already correct at insert doesn't have to wait for a heartbeat diff that will never fire.
- `apps/api/src/routes/devices/provision.ts`: same `device.created` emit for the pre-registration/pending-device provisioning path.
- `apps/api/src/routes/devices/groups.ts` (the `/devices/groups` duplicate of `/groups`): now calls `evaluateGroupMembership` on dynamic-group create/update, matching `routes/groups.ts`. This route's schema has no `filterConditions` field today, so the call is presently a no-op — kept for structural parity per the issue's explicit ask, and it activates automatically if/when this route gains filter support.

## Root cause

`deviceEvents.ts`'s exports (`emitDeviceChange`, `onDeviceChange`, `initializeDeviceEventHandlers`) had zero callers repo-wide outside their own file and tests. The module also used `require('../services/groupMembership')` inside `initializeDeviceEventHandlers` for circular-dependency avoidance that groupMembership.ts never actually created (it doesn't import deviceEvents.ts) — that `require()` call is unreachable/untested code in this package's ESM (`"type": "module"`) source, and fails outright under vitest, which is presumably why this dead code was never caught. Converted to static imports; verified this doesn't introduce the cycle the comment worried about.

## What was already fixed on `main` (verified, not touched here)

- **Org-move membership drop** (issue's closing comment): already handled, both at the DB level (`breeze_cascade_device_org_id()` trigger, migration `2026-10-09-000200-device-group-memberships-composite-tenant-fks.sql`) and at the app level (`routes/devices/moveOrg.ts:451`, explicit `DELETE FROM device_group_memberships WHERE device_id = ...`).
- Composite tenant FKs on `device_group_memberships` (group_id, org_id) and (device_id, org_id) already exist and already reject the cross-tenant shape the issue described.

## Left out of scope (documented, not silently dropped)

- **Software-change emission**: `deviceEvents.ts` already has a `device.software_changed` handler wired to `updateDeviceMemberships(..., ['software.installed', 'software.notInstalled'])`, but I did not wire an `emitDeviceChange` call from the actual software-inventory ingest path (`routes/agents/inventory.ts` → `ingestSoftwareInventoryReport`). That function's return shape has no "did the installed-software set actually change" signal, so emitting unconditionally on every accepted report risks re-evaluating every dynamic group in the org on every routine software-inventory sync, not just on a real diff. Needs a real diff signal from `ingestSoftwareInventoryReport` first; flagging as a follow-up rather than wiring it blind.
- **`routes/agents/enrollment.ts`'s big transactional insert**: the `device.created` emit is placed right after the transaction commits (using the confirmed non-`Response` `device` row), not inside the transaction — this handler has multiple early-return branches (race-lost, HTTPException) that made touching the transaction body itself too risky for this change's blast radius.
- **`device.deleted` wiring**: intentionally not wired. The hard-delete path (`DELETE /:id/permanent`) already cascades `device_group_memberships` rows away via `CORE_DEVICE_CASCADE_DELETE_TABLES`, so emitting there would be a redundant no-op. The soft-decommission path (`DELETE /:id`) does NOT delete the device row, and force-removing it from every dynamic group via `removeDeviceFromAllGroups` would incorrectly evict a decommissioned device whose non-status attributes still match the group's filter — that's a behavior change, not a bug fix, so left alone.
- **Periodic reconciliation sweep** (the issue's suggested backstop #2): not in this PR's scope per the dispatch brief's minimum-scope list.
- No boot-order unit test asserting `initializeDeviceEventHandlers()` fires from `index.ts` specifically (there's no `index.boot.test.ts` equivalent to `worker.boot.test.ts` to extend, and building one for a single call site felt disproportionate). Verified instead via `tsc`, targeted vitest runs, and direct grep.

## Verification

```
cd apps/api
NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit -p .          # clean
npx vitest run src/events                                               # 12 passed
npx vitest run src/routes/devices/groups.test.ts                        # 60 passed
npx vitest run src/routes/devices/provision.test.ts                     # 24 passed
npx vitest run src/routes/agents/heartbeat.test.ts                      # 187 passed
npx vitest run src/routes/agents/enrollment.test.ts                     # 65 passed
npx vitest run src/routes/groups.test.ts                                # 14 passed
npx eslint <all changed files>                                          # clean
```
Total: 362 tests passed across the affected files, tsc clean, eslint clean.

- `routes/devices/groups.ts`'s `evaluateGroupMembership` wiring is currently inert (see above) because that route's `createGroupSchema`/`updateGroupSchema` never accepted `filterConditions`. I did not expand the schema — that's a separate, larger decision (whether this duplicate route should gain full filter support or just be deleted in favor of `/groups`) that the issue itself flags as an open question ("fold ... or delete the duplicate route file").
- Heartbeat's emit is `await`ed (not fire-and-forget) so it runs inside the request's still-open `withDbAccessContext`, per the exact bug class documented in `routes/groups.ts`'s comments for the same reason.
- `initializeDeviceEventHandlers()` is only called from `index.ts`, not `worker.ts` — the latter never serves the heartbeat/enrollment/provision HTTP routes that call `emitDeviceChange`, so registering handlers there would be dead weight.

Closes #4630

### DIFF
--- a/apps/api/src/events/deviceEvents.test.ts
+++ b/apps/api/src/events/deviceEvents.test.ts
@@ -1,0 +1,156 @@
+/**
+ * #4630 — dynamic device group membership never re-evaluated on device
+ * change because this module's handlers were registered nowhere. Covers:
+ *   - initializeDeviceEventHandlers wires one handler per DeviceChangeEventType
+ *     to the correct groupMembership.ts function.
+ *   - emitDeviceChange fans out to every registered handler and never lets one
+ *     handler's rejection stop another (Promise.allSettled).
+ *   - mapChangedFieldsToFilterFields's field-name mapping used by the
+ *     heartbeat/enrollment/provision emit sites.
+ */
+import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const {
+  mockUpdateDeviceMemberships,
+  mockEvaluateDeviceMembershipForGroup,
+  mockRemoveDeviceFromAllGroups,
+  mockSelect,
+} = vi.hoisted(() => ({
+  mockUpdateDeviceMemberships: vi.fn().mockResolvedValue({ evaluatedGroups: 0, added: 0, removed: 0 }),
+  mockEvaluateDeviceMembershipForGroup: vi.fn().mockResolvedValue({ evaluatedGroups: 1, added: 0, removed: 0 }),
+  mockRemoveDeviceFromAllGroups: vi.fn().mockResolvedValue(undefined),
+  mockSelect: vi.fn(),
+}));
+
+vi.mock('../services/groupMembership', () => ({
+  updateDeviceMemberships: mockUpdateDeviceMemberships,
+  evaluateDeviceMembershipForGroup: mockEvaluateDeviceMembershipForGroup,
+  removeDeviceFromAllGroups: mockRemoveDeviceFromAllGroups,
+}));
+
+vi.mock('../db', () => ({
+  db: { select: mockSelect },
+}));
+
+vi.mock('../db/schema', () => ({
+  deviceGroups: { id: 'id', orgId: 'orgId', type: 'type', filterConditions: 'filterConditions', filterFieldsUsed: 'filterFieldsUsed' },
+}));
+
+import {
+  createDeviceChangeEvent,
+  emitDeviceChange,
+  initializeDeviceEventHandlers,
+  mapChangedFieldsToFilterFields,
+  onDeviceChange,
+  type DeviceChangeEventType,
+} from './deviceEvents';
+
+const DEVICE_ID = 'dddd0001-dddd-dddd-dddd-dddddddddddd';
+const ORG_ID = 'aaaa0000-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+
+function chainSelect(rows: unknown[]) {
+  return {
+    from: vi.fn().mockReturnValue({
+      where: vi.fn().mockResolvedValue(rows),
+    }),
+  };
+}
+
+describe('deviceEvents', () => {
+  // eventHandlers is a module-scope singleton with no reset hook, so this
+  // must register exactly once for the whole suite — calling it per-test
+  // would stack duplicate handlers on every DeviceChangeEventType.
+  beforeAll(() => {
+    initializeDeviceEventHandlers();
+  });
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockSelect.mockReturnValue(chainSelect([]));
+  });
+
+  it('wires device.updated to updateDeviceMemberships with mapped filter fields', async () => {
+    await emitDeviceChange(createDeviceChangeEvent('device.updated', DEVICE_ID, ORG_ID, ['hostname', 'deviceRole']));
+
+    expect(mockUpdateDeviceMemberships).toHaveBeenCalledWith(DEVICE_ID, ORG_ID, ['hostname', 'deviceRole']);
+  });
+
+  it('does not call updateDeviceMemberships for device.updated with no changed fields', async () => {
+    await emitDeviceChange(createDeviceChangeEvent('device.updated', DEVICE_ID, ORG_ID, []));
+
+    expect(mockUpdateDeviceMemberships).not.toHaveBeenCalled();
+  });
+
+  it('wires device.hardware_updated to updateDeviceMemberships with hardware-prefixed fields', async () => {
+    await emitDeviceChange(createDeviceChangeEvent('device.hardware_updated', DEVICE_ID, ORG_ID, ['cpuModel']));
+
+    expect(mockUpdateDeviceMemberships).toHaveBeenCalledWith(DEVICE_ID, ORG_ID, ['hardware.cpuModel']);
+  });
+
+  it('wires device.network_updated to updateDeviceMemberships with network-prefixed fields', async () => {
+    await emitDeviceChange(createDeviceChangeEvent('device.network_updated', DEVICE_ID, ORG_ID, ['ipAddress']));
+
+    expect(mockUpdateDeviceMemberships).toHaveBeenCalledWith(DEVICE_ID, ORG_ID, ['network.ipAddress']);
+  });
+
+  it('wires device.software_changed to updateDeviceMemberships with the fixed software fields', async () => {
+    await emitDeviceChange(createDeviceChangeEvent('device.software_changed', DEVICE_ID, ORG_ID, []));
+
+    expect(mockUpdateDeviceMemberships).toHaveBeenCalledWith(
+      DEVICE_ID,
+      ORG_ID,
+      ['software.installed', 'software.notInstalled'],
+    );
+  });
+
+  it('wires device.created to evaluate every dynamic group with a filter in the org', async () => {
+    mockSelect.mockReturnValue(chainSelect([{ id: 'group-1' }, { id: 'group-2' }]));
+
+    await emitDeviceChange(createDeviceChangeEvent('device.created', DEVICE_ID, ORG_ID, []));
+
+    expect(mockEvaluateDeviceMembershipForGroup).toHaveBeenCalledWith('group-1', DEVICE_ID);
+    expect(mockEvaluateDeviceMembershipForGroup).toHaveBeenCalledWith('group-2', DEVICE_ID);
+  });
+
+  it('wires device.deleted to removeDeviceFromAllGroups', async () => {
+    await emitDeviceChange(createDeviceChangeEvent('device.deleted', DEVICE_ID, ORG_ID, []));
+
+    expect(mockRemoveDeviceFromAllGroups).toHaveBeenCalledWith(DEVICE_ID);
+  });
+
+  it('runs every handler for an event type and does not let one rejection stop another', async () => {
+    const failing = vi.fn().mockRejectedValue(new Error('boom'));
+    const succeeding = vi.fn().mockResolvedValue(undefined);
+    const eventType: DeviceChangeEventType = 'device.updated';
+    onDeviceChange(eventType, failing);
+    onDeviceChange(eventType, succeeding);
+
+    await emitDeviceChange(createDeviceChangeEvent(eventType, DEVICE_ID, ORG_ID, ['hostname']));
+
+    expect(failing).toHaveBeenCalled();
+    expect(succeeding).toHaveBeenCalled();
+    // The pre-existing initializeDeviceEventHandlers handler for device.updated
+    // also ran, alongside both of these ad hoc ones.
+    expect(mockUpdateDeviceMemberships).toHaveBeenCalled();
+  });
+
+  it('is a no-op for an event type with no registered handlers', async () => {
+    await expect(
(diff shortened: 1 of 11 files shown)


ToddHebebrand:
**Tests:** `src/events/deviceEvents.test.ts` (12), `src/routes/devices/groups.test.ts` (60), `src/routes/devices/provision.test.ts` (24), `src/routes/agents/heartbeat.test.ts` (187), `src/routes/agents/enrollment.test.ts` (65), `src/routes/groups.test.ts` (14) — 362/362 passed; `tsc --noEmit -p apps/api` clean; `eslint` on all changed files clean.

ToddHebebrand:
**Independent review (Sonnet, read-only):** BLOCK on two findings; re-verified and resolved in `218f24e3ea`.

1. **Confirmed + fixed:** `deviceUpdates.deviceRole` was assigned whenever the agent sent a role and the source was `auto`, with no comparison to the stored value. The agent sends `deviceRole` on every heartbeat and `auto` is the default, so every steady-state heartbeat became a filterable change and re-evaluated dynamic groups on each beat. Now gated on inequality like hostname/osVersion/osBuild. The existing steady-state test omitted `deviceRole`; added one that re-reports the unchanged role and asserts no emit (red before, green after).
2. **Refuted:** the reviewer read the `device.created` emit in enrollment.ts as running after `withSystemDbAccessContext` resolved. It is inside the callback (the try/catch above it is an inner block). Added a test that asserts the emit fires while the system context is open, so the invariant is pinned.

heartbeat.test.ts + enrollment.test.ts: 254/254; `tsc --noEmit` clean.

ToddHebebrand:
**Review run:** pr-review-toolkit (Opus) + orchestrator triage — **FIX-THEN-MERGE**, held as draft.

**Critical**
1. **Re-evaluation runs inline inside the heartbeat's open PG transaction, with Redis I/O.** `withDbAccessContext` is `baseDb.transaction(...)`; the new `await emitDeviceChange` in `routes/agents/heartbeat.ts` therefore runs after `UPDATE devices` took its row lock, and each membership flip calls `schedulePeripheralPolicyDevice` → 2–3 Redis round-trips while holding a pooled PG connection. No debounce, no per-device coalescing. This is the pool-starvation shape from the 09-03 US incident. Fix: enqueue a coalesced BullMQ job (`jobId = group-reeval-${deviceId}`, the pattern `jobs/peripheralJobs.ts` already uses) and drop the inline await.
2. **Unordered multi-group lock acquisition.** `evaluateDeviceMembershipForGroup` → `ensureFilterFieldsUsed` UPDATEs `device_groups` when stale; the group SELECT in `updateDeviceMembership` has no `ORDER BY`, so concurrent heartbeats can take group row locks in different orders (40P01, #3911 pattern). Add `ORDER BY id`.

**Important**
3. No integration test — every new test mocks `groupMembership`/`emitDeviceChange`, so a contextless SELECT returning zero groups is invisible. #4630 asked for exactly this; add one under `src/__tests__/integration/`.
4. `Closes #4630` overclaims: PATCH, site move, org move, tag changes and inventory ingest are still unwired (`changedFields.includes('siteId')` has no caller). Use `Refs #4630`.
5. `events/deviceEvents.ts` `Promise.allSettled` never inspects results — a permanently-throwing evaluation is silent. Log rejected results at error level.

**Suggestion:** the `routes/devices/groups.ts` wiring is inert (schema has no `filterConditions`) — drop it or delete the duplicate route.

**Clean:** tenancy (`device_groups` is org-only, emits run in the right context), `deviceRole` inequality guard, statement-timeout restore.

ToddHebebrand:
**Review fixes landed — but as a follow-up PR, not here.** This PR was squash-merged (`1a76507d3a`) before the fix round completed, and its head branch was deleted, so the five findings are addressed in **#5039** on top of current `main`: https://github.com/LanternOps/breeze/pull/5039

What changed, per finding:

**1 (critical) — inline re-evaluation inside the heartbeat's open transaction.** Removed entirely. New `apps/api/src/jobs/deviceGroupJobs.ts` owns a coalescing BullMQ queue (`device-group-reevaluation`, `jobId = group-reeval-<deviceId>`), following the `policy-reconciliation-<deviceId>` shape in `jobs/peripheralJobs.ts` — including the active-job follow-up branch and the spent-record replace branch. Coalescing merges rather than overwrites (changed fields unioned; `device.created` subsumes `device.updated`, since created evaluates every group). `heartbeat.ts` / `enrollment.ts` / `provision.ts` now do nothing but `void requestDeviceGroupReevaluation(...)` — no group evaluation and no awaited Redis I/O while the request holds a pooled connection and the `UPDATE devices` row lock; that helper never rejects. The worker runs the evaluation under `withSystemDbAccessContext` and **re-reads the device's own `org_id`** rather than trusting the job payload. Registered as `deviceGroupJobs` (placement `socket-owner`, per `workerEntrypointClosure.contract.test.ts`, not guessed), and `initializeDeviceEventHandlers` is now idempotent and called from the processor — `worker.ts` never runs `index.ts`'s bootstrap, so without that a worker-role process would drain the queue with an empty handler registry and every job would be a silent no-op.

**2 (critical) — unordered group lock acquisition.** `ORDER BY id` added to the dynamic-group SELECT in `updateDeviceMembership`, and to the `device.created` handler's group SELECT in `deviceEvents.ts`, which has the identical `ensureFilterFieldsUsed` UPDATE hazard.

**3 (important) — no integration test.** New `src/__tests__/integration/dynamicGroupReevaluation.integration.test.ts`, 6 cases against real Postgres + real forced RLS + the real filter engine: hostname change **adds** membership, then **removes** it; `device.created` pickup; payload-org-spoof guard; cross-tenant isolation; deleted-device no-op. All 6 pass locally. Proved discriminating — commenting out the handler-registry init reds 4 of the 6.

**4 (important) — silent `Promise.allSettled`.** `emitDeviceChange` now logs every rejected handler at error level with the event type, device id and org id.

**5 (suggestion) — inert `routes/devices/groups.ts` wiring.** Removed, along with its two mock-only tests.

**`Closes #4630` → `Refs #4630`.** #5039's body lists the still-unwired call sites: `PATCH /devices/:id`, site move (the `changedFields.includes('siteId')` branch still has no caller), org move, tag changes, custom-field writes, hardware/network/software inventory ingest, device delete, and the periodic reconciliation sweep.

Verification on #5039: `tsc --noEmit -p apps/api` exit 0; 1858 unit tests across 104 affected files pass; eslint clean; the new integration suite 6/6 against `localhost:5433/breeze_test`.

PR #5017, opened 2026-09-06
perf(inventory): scope the finding pre-lock by org_id so it hits device_vuln_org_device_idx

ToddHebebrand:
## Why

`replaceSoftwareInventoryProjection` (every software-inventory sync) locks a device's linked findings with `SELECT ... FOR UPDATE OF device_vulnerabilities WHERE device_vulnerabilities.device_id = $1 AND software_inventory.device_id = $2`.

`device_vulnerabilities` has no index led by `device_id` (its FK is uncovered). The only candidate, `device_vuln_org_device_idx (org_id, device_id)`, cannot serve a `device_id`-only predicate, so the statement scanned the table under the row lock on every sync.

US prod `pg_stat_statements`, 2026-08-30 to 09-06:

| calls | mean | max | shared_blks_read |
|---|---|---|---|
| 32,657 | 789 ms | 738 s | 3.2M |

## What

Add `eq(deviceVulnerabilities.orgId, device.orgId)` to the predicate. The caller already has the org, the index already exists, no migration. Semantics unchanged (a device's findings are always in the device's org).

## Verified

- Red-first: new test renders the `WHERE` with `PgDialect.sqlToQuery` and asserts both `org_id` and `device_id` predicates; failed before the change, passes after.
- `softwareInventoryObservations.test.ts`: 18 passed. `tsc --noEmit` clean.

Part of the DB deep-dive following the US slowness on 2026-09-06 (sibling: #5009).

https://claude.ai/code/session_012SBWJSD8QenDKGnnTsKtgs

### DIFF
--- a/apps/api/src/services/softwareInventoryObservations.test.ts
+++ b/apps/api/src/services/softwareInventoryObservations.test.ts
@@ -1,4 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { PgDialect } from 'drizzle-orm/pg-core';
 import type { LegacySoftwareInventoryReport, SoftwareInventoryObservationV2 } from '@breeze/shared';
 
 // Hoisted so the `vi.mock` factories below (themselves hoisted above these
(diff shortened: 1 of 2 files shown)


ToddHebebrand:
Superseded by #5021: review pointed out the ingest runs in a system-scoped DB context, so an org_id predicate could skip findings during an org-move race. #5021 adds a device_id-led index on device_vulnerabilities instead, same plan, no semantic change.

PR #5028, opened 2026-09-06
feat(api): one effective custom-field key namespace per device (#3257 W03)

### LINKED ISSUE #4771: W03: Anti-shadowing: one effective field_key namespace per device (gated on read-only prod shadowed-key preflight)
Wave W03 of #4768 — Anti-shadowing: one effective field_key namespace per device (gated on read-only prod shadowed-key preflight)
Branch: `feature/4768-rmm-custom-field-import/wave-4771`

ToddHebebrand:
Closes #4771
Refs #3257 (W03 of the RMM custom-field import plan), feature #4768

## Todd gate — prod shadowing preflight (passed before this was written)

Run by the orchestrator **2026-09-05 ~21:55Z**, read-only, on **both** prod regions via the billing container's postgres client:

- `breeze-eu` → `{"shadowed":[]}` (0 definitions total)
- `breeze-us` → `{"shadowed":[]}` (1 definition total)

Zero shadowed pairs on both regions, so this migration's deploy-abort will not fire. This PR ships the versioned preflight file for the record and for future regions; it was **not** re-run against prod from here, and nothing in this branch SSHes anywhere.

## What this does

`custom_field_definitions` is dual-axis (`org_id` XOR `partner_id`, W02), but `devices.custom_fields` is a **flat jsonb object keyed by a bare string**. If an org-owned `udf7` and a partner-wide `udf7` can both exist for one org, that single datum has two definitions, and three things break:

1. The partner export emits **two records for one datum** — the identity hash includes `f.id` (`routes/partnerApi/configuration.ts:433,440`).
2. W05's normalized `device_custom_field_values` cannot project back into the jsonb without loss.
3. W05's backfill cannot attribute an existing blob value to either definition — the blob records no `definition_id`, and the two definitions may disagree on `type` and dropdown `options`.

So: one effective `field_key` namespace per device, enforced in the database.

**Why a trigger and not an index.** The rule spans two *nullable* ownership columns over *disjoint* row sets, and the partner half of the pair is not even on the org-owned row — it is reached through `organizations.partner_id`. No index can express that, which is why W02 shipped two *partial* unique indexes and deliberately left this case open. The trigger takes a transaction-scoped advisory lock on `(partner, field_key)` — both branches derive the same key — so two concurrent inserts on opposite axes cannot both pass their own `EXISTS` check under READ COMMITTED.

Existing shadowed pairs are **reported and the deploy aborted**, never auto-resolved: deleting either definition silently retypes or orphans every value already stored under that key, and consumers hold the key as configuration (`remoteAccessLauncher.ts`, `installerVariables.ts`). Reconciliation needs a human decision per pair.

## The finding that shaped this PR

The plan's trigger sketch was `SECURITY DEFINER` with plain lookups. **That is a no-op in production for the exact case it exists to catch**, and CI cannot see it.

- An **org token cannot see partner-wide rows on this table at all** — there is no partner-wide SELECT branch yet (`PARTNER_WIDE_SELECT_BRANCH_EXEMPT`, open follow-up #4944). An unelevated lookup finds nothing and waves the shadowing insert through — on the most common path there is, since `POST /custom-fields` pins `orgId = auth.orgId` for every org-scoped caller.
- `SECURITY DEFINER` does **not** fix it. It switches to the function *owner* — the migration role. Locally and in CI that is `breeze_test` (`rolsuper=t`, `rolbypassrls=t`), which ignores RLS outright. In production it is `doadmin` on managed DO Postgres: not a superuser, bound by `FORCE ROW LEVEL SECURITY` like everything else.

So the function elevates to system scope for its lookups. The first draft used a **function-level** `SET "breeze.scope" = 'system'` — the elegant form, since PostgreSQL saves and restores it automatically. **That form cannot ship**: setting a custom (dotted) GUC as a function attribute is superuser-only, so prod's non-superuser migration role 42501s at `CREATE FUNCTION` time and the deploy crash-loops on boot. It reddened `Check Migrations (non-superuser)` and `Test API`. The shipped form is in-body save/elevate/restore with a single return path (deviation 7 below).

**Measured, not assumed** (local PG16, this migration set, A/B with the `SET` clause as the sole variable): two otherwise identical `SECURITY DEFINER` lookups reassigned to the non-superuser `breeze_app` and called under an org-scoped context returned

| | sees the partner-wide row? |
|---|---|
| unelevated | **no** |
| `SET "breeze.scope" = 'system'` | **yes** |

And with the elevation stripped, **all four tenant-scope tests in the new suite still passed** — because the owner here is a superuser. That is why the suite also asserts the function's stored **body** from the catalog (`pg_get_functiondef`): on this stack it is the only assertion in the file that can fail when the elevation is removed. Same CI-superuser blind spot `migrationRlsScope.test.ts` exists for. Both halves are commented so neither gets deleted as redundant.

One correction to the above, made in review: prod's `doadmin` is a non-superuser but **does** carry BYPASSRLS (`scripts/check-migrations-nonsuperuser.ts` models exactly that attribute set), so hosted prod is probably covered by ownership alone today. The elevation is what makes the guard correct on a deployment whose migration role lacks BYPASSRLS — self-hosted, or any future owner — rather than silently ownership-dependent. The A/B measurement stands; only the claim about which deployments it protects was too broad.

The migration's own detection block gets the same treatment (W02's lesson), plus a probe that prints the effective scope and row counts and **raises** if the scope is not `system` — so a "0 cross-axis shadowed keys" verdict cannot be a truthful-looking blind read.

## Ordering note (differs from the brief's expectation)

The brief expected the W02 pattern: under a tenant context, RLS `WITH CHECK` fires 42501 *before* any CHECK/trigger error. That is true for W02's XOR **CHECK constraint**. It is **not** true here: this is a `BEFORE ROW` trigger, and `ExecInsert` runs BR triggers *ahead* of both `ExecWithCheckOptions` and `ExecConstraints`. So a shadowing insert surfaces as **P0001 even under an org context**. Both behaviours are asserted separately (`refuses … under the caller's own org scope` → P0001; `an org token still cannot forge a partner-wide row` → 42501) so neither can be mistaken for the other. Code is truth; this is documented in the migration header and the suite.

## Deviations

1. **Migration slot moved three times.** Assigned `2026-10-10-140000`; `origin/main` gained `2026-10-11-000100-…` (#5009), then `2026-10-11-140300-automation-run-cancellation.sql` during review, each time making the file no longer sort last. Final slot: **`2026-10-11-141000-custom-field-no-cross-axis-shadowing.sql`** (+ matching preflight). Both integration suites replay migrations by path, so each rename swept every reference — 11 of them at the last count (`autoMigrate.test.ts` asserts they resolve, turning a miss into a unit failure rather than an ENOENT mid-shard). All differ from the plan's `2026-10-08-…`, which sorts before shipped files.
2. **Two files outside the brief's "files you own" list, both forced:**
   - `apps/api/src/__tests__/integration/customFieldDefinitionIntegrity.integration.test.ts` — W02's `does NOT (yet) block an org-owned key that shadows a partner-wide one — that is W03` test. Its own comment says it was pinned so W03's red test would be unambiguous; W03 makes it false, so it is **inverted**, not deleted. It now asserts **P0001** rather than 23505 — which is what proves W02's two partial indexes were not quietly merged into a composite. Without this the suite is red.
   - `apps/api/src/routes/customFields.ts` (+ its test) — plan W03 Task 1 **Step 6**, inside the section the brief scoped me to. Without it the trigger's P0001 reaches the client as a bare 500.
3. **Step 6 extended to 23505.** The plan asked only for P0001 → 409. W02 shipped the unique indexes with no mapper, so a duplicate key is a raw 500 today, in the same `catch` and the same defect class. Both now map to 409: `field-key-shadowed` passes the trigger's copy through verbatim (it names the key and axis and discloses nothing else); `field-key-duplicate` uses fixed copy built from the request payload, because the driver's 23505 message carries the index name and `detail` carries the offending column values. Unmapped SQLSTATEs still propagate — asserted.
4. **PATCH deliberately not wrapped.** The trigger is `UPDATE OF org_id, partner_id, field_key`, and the PATCH handler never sets any of those (`name`/`options`/`required`/`defaultValue`/`deviceTypes`/`scriptWrite` only), so neither P0001 nor 23505 is reachable there. Adding a dead `catch` would be noise.
5. **Error message is minimum-disclosure by design.** It names the key and the colliding axis and nothing else — no id, name, or owning org. An org caller cannot otherwise see partner-wide rows (#4944), so anything more would make the trigger an enumeration oracle for partner-wide config. Asserted.
6. **No Drizzle-side comment was named by the plan for W03.** Added one anyway to `db/schema/customFields.ts` documenting all three DB-level rules (W02's CHECK + two indexes, W03's trigger) and their SQLSTATEs, since Drizzle models columns only. Comment-only; no drift.
7. **Scope elevation is in-body, not a function attribute** (changed after CI went red — see the review-summary comment). `SET "breeze.scope" = 'system'` on the function header is superuser-only for a custom dotted GUC and 42501s prod's migration role at `CREATE FUNCTION` time; that is the v0.97.0 EU crash-loop class, and `src/db/migrationGucAttributes.test.ts` plus `Check Migrations (non-superuser)` both guard it. The shipped form is the repo's sanctioned save/elevate/restore, copied from `breeze_revalidate_config_policy_feature_references` (`2026-07-27-a-feature-policy-reference-ownership.sql`): `_prev_scope` captured in `DECLARE`, restored immediately before a **single** `RETURN NEW` — the `owner_partner IS NULL` early return became a fall-through so no path can skip it. RAISE paths restore via (sub)transaction rollback. `SECURITY DEFINER` and `SET search_path` are kept: `search_path` is a *built-in* GUC, so its attribute form is available to any role, and the 07-27 precedent keeps both. The suite's `pg_proc.proconfig` assertion was replaced with one on the stored body, which additionally asserts the attribute form has **not** crept back in.
8. **The anti-shadowing trigger is registered in `ORG_ID_BENIGN_TRIGGERS`** (`orgMergeRegistry.integration.test.ts`), with the reason that an org merge is same-partner (`orgMerge.ts:289`, re-validated in-transaction at `:605`), so the repoint cannot move a definition into a different partner namespace — the loser's row already had to clear that exact check to exist. **W02's reconcile was therefore not extended**: the cross-axis shadow state is unreachable, because this migration aborts the deploy if it exists on arrival and the trigger refuses every write that would create it afterwards. Since that is a *reachability* argument rather than an inertness one, `customFieldDefinitionsMerge.integration.test.ts` pins it both ways — a clean merge with partner-wide definitions present, and a forged shadow (`ALTER TABLE … DISABLE TRIGGER`) proving the repoint really would abort with P0001. `ORG_ID_CONDITIONALLY_BLOCKING_TRIGGERS` does not fit: it requires a live `blocks-merge` policy to discharge it, and there is none here.

## Verification (all foreground, live DB via `pnpm test-stack`)

Red first: the suite was written and run against a database migrated **without** this file — **11 failed / 6 passed**, the 6 passing being exactly the over-tightening guards (same key under two partners, two orgs under one partner, foreign partner, unrelated column, the 42501 forge, no scope leak). After applying: **17/17**.

| Check | Result |
|---|---|
| `customFieldShadowing` + `customFieldDefinitionIntegrity` + `customFieldDefinitionsPartnerRls` + `customFieldDefinitionsMerge` + `orgLifecycleFoundations` + `scriptCustomFieldWriteBack` + `deviceCustomFieldImportIsolation` (integration) | **7 files / 71 tests pass** |
| `vitest.config.rls-coverage.ts` | **102 pass** |
| `src/db/autoMigrate` + `src/db/migrationRlsScope` + `src/services/orgMerge` + `src/routes/customFields` (unit) | **9 files / 226 pass** |
| `pnpm db:check-drift` | no drift, 678 files match the ledger |
| `scripts/check-migration-naming.sh --against-ref origin/main` | OK |
| migration apply | probe printed `effective_scope=system, total_rows=0, …`; `0 cross-axis shadowed keys`; re-apply is a true no-op |

`migrationRlsScope.test.ts`'s frozen baseline did **not** grow. `origin/main` was merged into this branch before the final verification run, so these results are against the merge.

Mutation-checked rather than assumed: flipping `'P0001'`/`'23505'` in the route handler turns the two mapping tests red (2 failed / 25 passed), so they are not vacuous.

## Not verified

- **Never run against prod.** The preflight results above are the orchestrator's, quoted; this branch did not connect to any region.
- ~~**No non-superuser end-to-end migration run.**~~ Now done: `check:migrations:nonsuperuser` was run locally against a clean PG16 as `breeze_migrator` (NOSUPERUSER, the doadmin-alike role the script provisions) and passed, with the migration's own probe printing `effective_scope=system` and `0 cross-axis shadowed keys`. What remains inferred is the behaviour of a migration role **without** BYPASSRLS, which neither CI nor that script exercises.
- **Concurrency is reasoned, not raced.** The advisory lock's correctness argument (both branches derive the same key; the loser blocks to commit and then sees the row) has no test — a reliable two-transaction race needs machinery this suite does not have. Single-transaction behaviour is fully covered.
- **No web/UI work**, no importer, no `device_custom_field_values` — W05/W07/W09.
- **#4944 (partner-wide SELECT branch for this table) is untouched** and still open. When it lands, the `org token must NOT see partner-wide rows` control in the new suite is the assertion that must be revisited; its comment says so.
- Full `pnpm test` was not run — only the suites listed above plus lint and typecheck.

https://claude.ai/code/session_01AkN3dN6HQ7yaoSD9EUGLPw


ToddHebebrand:
## Review summary

Four specialist reviewers ran against this PR's diff (read-only, via `gh pr diff`): **code-reviewer**, **pr-test-analyzer**, **silent-failure-hunter**, **comment-analyzer**. Five findings survived verification; **all five are fixed** in `b075470f3d`. Every fix was confirmed by measurement, not argument.

### Fixed

**1. Two assertions in the new suite proved nothing** (pr-test-analyzer, critical — and it implied a second one the reviewer hadn't reached).

Both read the error via `JSON.stringify(raised)`. postgres.js sets `.message` through `Error`'s constructor, so it stays **non-enumerable** through the later `Object.assign` and never reaches the serialized form. Reproduced against this repo's pinned drizzle-orm/postgres:

```
{"query":"insert into x","params":[],"cause":{"code":"P0001","severity":"ERROR","name":"PostgresError"}}
  → includes the raised message?  false
```

- `does not disclose the conflicting definition id or name` was two `not.toContain` checks against a message-less string — it passed unconditionally, including against a trigger that leaks the conflicting id. Now reads `pgErrorNode(...).message` and leads with a **positive control** (the message must name the key) so an empty read can't satisfy the negatives. **Mutation-checked:** with the trigger changed to append `(id %)` the test now fails; under that same mutation it previously passed.
- The deploy-abort test's `toMatch(/cross-axis shadowed key/)` passed only because that test replays the whole migration **file** as the query text, and drizzle *does* serialize `query` — so it was asserting that the migration source contains its own `RAISE` literal. Now reads the real message and additionally pins `has 1 cross-axis shadowed key`, proving the substituted count rather than the `%` placeholder.

**2. The preflight did not fail closed** (silent-failure-hunter, high). The migration hard-raises when its detection read is RLS-blind; the preflight — the file an operator actually runs against prod to decide merge/don't-merge — only *printed* the scope and relied on a human cross-checking it before trusting query (a)'s emptiness. "Empty because clean" and "empty because blind" look identical, which is the precise failure this file exists to prevent. It now raises, aborting the transaction so query (a) errors with 25P02 instead of returning a reassuring empty set. **Drilled** as `breeze_app` with the elevation stripped:

```
 effective_scope | total_rows
 none            |          0
ERROR:  PREFLIGHT IS RLS-BLIND: effective scope is none, not system. ... Do NOT report "clean"
```
Query (a) never runs. Before the fix it printed `(0 rows)` and read as clean.

**3. Stale line citation** (comment-analyzer). The partner-export identity hash is at `routes/partnerApi/configuration.ts:436,439`, not `:433,440`. Verified and corrected in both places it appeared.

**4. `owner_partner IS NULL` comment was wrong** (silent-failure-hunter, medium — found independently before the report landed). `organizations.partner_id` is **NOT NULL**, so that branch guards a *missing org*, not a partnerless one. The comment now says so, records that FK checks run after BEFORE-row triggers, and a new test pins the behaviour: a nonexistent `org_id` must yield **23503**, not an invented shadowing error. (This also disproves the "unreachable dead code" read — it is reachable, and now covered.)

**5. Partner-axis UPDATE was untested** (pr-test-analyzer). `partner_id` and `field_key` are both in the trigger's `UPDATE OF` list, but the `ELSIF NEW.partner_id IS NOT NULL` arm was only ever reached by INSERT. Added.

Also, unprompted: both `RAISE`s now set `CONSTRAINT = 'custom_field_definitions_no_shadow'`. P0001 is Postgres's generic `raise_exception` code, so a future unrelated P0001 on this table would be mislabelled `field-key-shadowed` (silent-failure-hunter, low — not a today-bug). Making the error self-identifying costs two words and keeps the fix cheap; nothing depends on it yet.

### Accepted as-is, with reasons

- **No `lock_timeout` on the advisory lock** (informational). Matches every existing `10002xx`/`10003xx` serializer in the repo; a systemic pattern, not something introduced here.
- **Concurrency is reasoned, not raced.** Still the honest gap, already in the PR body's not-verified list. A reliable two-transaction race needs machinery this suite doesn't have.
- **The `pg_proc.proconfig` catalog assertion stays.** Both the test analyzer and code reviewer independently confirmed no behavioural test can discriminate the scope elevation on a BYPASSRLS-superuser-owned function, so asserting structure is the correct fallback here, not a shortcut.

### Clean on inspection

code-reviewer found **no issues** in any of its four focus areas (trigger logic across every enumerated edge case, the `SET "breeze.scope"` security question, migration hygiene, and the route's try/catch). comment-analyzer independently verified all five PostgreSQL-semantics claims against PostgreSQL source and all eleven repo-fact claims against this branch — every one correct except the line citation above. silent-failure-hunter separately confirmed the route rethrows anything unmapped into `app.onError` (logged + Sentry), that `pgErrorCode` returning `undefined` falls through to the rethrow, and that no trigger path can return `NULL` and silently skip the insert.

| Check | Result |
|---|---|
| 7 custom-field + org-lifecycle integration suites | **73 pass** |
| `vitest.config.rls-coverage.ts` | 102 pass |
| `vitest.config.rls.ts` | 26 pass / 5 skipped |
| unit: `autoMigrate`, `migrationRlsScope`, `orgMerge`, `customFields` | **226 pass** |
| `tsc --noEmit` | clean |
| `pnpm lint` | 6/6 |
| `pnpm db:check-drift` | no drift (679 files) |
| `check-migration-naming.sh --against-ref origin/main` | OK |

Not merging — leaving that and closing #4771 to you.

ToddHebebrand:
## Review summary — CI-fix round on `871a07aaf`

Three red jobs on `b075470f3d`, two root causes. Both fixed; `/pr-review-toolkit:review-pr` run on the new head with three specialist agents (code, tests, comments — Sonnet).

### 1. `Check Migrations (non-superuser)` + `Test API` (`migrationGucAttributes.test.ts`)

`SET "breeze.scope" = 'system'` as a **function attribute** is superuser-only for a custom (dotted) GUC. Prod migrates as a non-superuser, so `CREATE FUNCTION` 42501s and the deploy crash-loops on boot — the v0.97.0 EU class, which is exactly what `migrationGucAttributes.test.ts` guards. The earlier A/B was right that elevation inside the function is needed, wrong about the form.

Moved to the sanctioned in-body save/restore, copying `breeze_revalidate_config_policy_feature_references` (`2026-07-27-a-feature-policy-reference-ownership.sql`). `SECURITY DEFINER` and `SET search_path` are kept — `search_path` is a **built-in** GUC, so its attribute form is available to any role, and the 07-27 precedent keeps both.

The body was restructured to a **single `RETURN NEW`** (the `owner_partner IS NULL` early return became a fall-through) so no path can skip the restore. RAISE paths need no hand-restore: a transaction-local `set_config` rolls back with the (sub)transaction.

The suite's `pg_proc.proconfig` assertion was asserting the one form that cannot ship, so it is replaced by a catalog assertion on the stored **body** (`pg_get_functiondef`): elevation present, restore present, exactly one return, `prosecdef`, a `search_path=` entry — and, newly, that `breeze.scope=` has **not** crept back into `proconfig`.

**Mutation-checked on fresh DBs, three ways:**

| Mutation | Result |
|---|---|
| restore line removed | **3 red**, including the behavioural `does not leak system scope into the calling transaction` |
| elevation line removed | **1 red**, the catalog assertion only — exactly as the file's superuser-blind-spot comment claims |
| `COALESCE(_prev_scope, '')` → `'system'` | **2 red** (the new COALESCE test + the catalog assertion) |

Moving off the attribute form is what made a dropped restore *locally* detectable — Postgres used to restore the attribute itself, so no behavioural test could see it.

Also corrected an overclaim carried in the migration header and the suite: prod's `doadmin` is a non-superuser but **does** carry BYPASSRLS (`scripts/check-migrations-nonsuperuser.ts` models exactly that attribute set), so hosted prod is probably covered by ownership alone today. The elevation is what makes the guard correct on a deployment whose migration role lacks BYPASSRLS, rather than silently ownership-dependent. The `guc.c validate_option_array_item` citation was dropped as unverifiable-in-repo.

### 2. `Integration Tests (shard 3/4)` — `orgMergeRegistry`

`custom_field_definitions.custom_field_definitions_no_shadow` is registered in **`ORG_ID_BENIGN_TRIGGERS`**.

The trigger RAISEs only when the **destination** org's partner already owns a partner-wide row with the same `field_key`. An org merge is same-partner (`orgMerge.ts:289`, re-validated against fresh rows inside the merge transaction at `:605`), so the repoint never changes which partner namespace is checked — the loser's row already had to clear that exact check to exist.

**W02's reconcile therefore needs no cross-axis extension.** The cross-axis shadow state is unreachable: this migration aborts the deploy if it exists on arrival, and the trigger refuses every write that would create it afterwards. The only hole is a change to `organizations.partner_id`; a repo-wide grep found no production path that writes it (`updateOrganizationSchema` explicitly `.omit({ partnerId: true })`) — only test fixtures.

That is a **reachability** argument, not an inertness one, so it is pinned both ways in `customFieldDefinitionsMerge.integration.test.ts`:

- `merges cleanly while the partner owns partner-wide definitions` — the reachable case.
- `an existing cross-axis shadow WOULD abort the repoint` — forges the shadow (`ALTER TABLE … DISABLE TRIGGER`, which the test role owns and `breeze_app` deliberately does not), re-arms, and asserts the merge dies with **P0001** + `constraint_name = custom_field_definitions_no_shadow`. Without it, BENIGN would be indistinguishable from "the trigger never fires." It asserts the SQLSTATE and constraint from the unwrapped `.cause`, not Drizzle's `Failed query: UPDATE …` envelope, which would have matched any failure of that statement.

`ORG_ID_CONDITIONALLY_BLOCKING_TRIGGERS` does not fit structurally — that map requires a live `blocks-merge` policy to discharge it, and there is no runtime policy here, only an invariant.

### Review outcome

| Agent | Findings |
|---|---|
| code-reviewer | **0** — independently re-derived the save/restore exit analysis, the `COALESCE('')` → `'none'` equivalence, the system-scope short-circuit in `breeze_has_org_access`/`breeze_has_partner_access`, and the BENIGN argument against `buildRepoint` / `mergeCustomFieldDefinitions` / a repo-wide `organizations.partner_id` grep |
| pr-test-analyzer | 4 acted on (2 coverage, 2 robustness) |
| comment-analyzer | 2 acted on; 9 of 11 factual claims VERIFIED against code |

**All findings fixed in `871a07aaf`, none deferred:**

- **Untested `COALESCE(_prev_scope, '')` branch** (the NULL half of the restore — the line the fix is built on). Every other insert in the suite goes through `withDbAccessContext`, which always stamps a scope, so nothing reached it. New test runs the insert inside one `getTestDb()` transaction (`set_config` is transaction-local; a read on another pooled connection would say `none` and prove nothing) with a pre-insert control asserting the connection really is scope-less. Mutation-checked above.
- **Untested partner-scope restore** — the trigger's `ELSIF` branch had no leak assertion; only the org branch did. Added.
- **Catalog assertion brittleness** — plpgsql bodies come back from `pg_get_functiondef` verbatim *with their comments*, so a commented-out elevation satisfied a raw `includes`, and `RETURN  new;` would have slipped past the single-return count. Now strips `--` comments and matches case-/whitespace-insensitively.
- **Swallowed cleanup error** — the `afterEach` trigger re-arm now logs before swallowing (the convention `cleanupDatabase()` uses for the `audit_logs` TRUNCATE guard). A silently-failed re-arm would leave the W03 guard OFF with no signal.
- Two comment corrections (the stale `doadmin`/FORCE-RLS premise; the unverifiable `guc.c` citation).

The reviewer also flagged, for awareness rather than as a defect: `2026-07-26-a-normalized-policy-tenant-integrity.sql` installs an `AFTER UPDATE OF partner_id ON organizations` trigger, i.e. the schema anticipates that column becoming mutable someday. That is the one latent path that could invalidate BENIGN — which the registry entry's own comment already calls out as the condition to re-check.

### Migration slot moved again

`origin/main` gained `2026-10-11-140300-automation-run-cancellation.sql` mid-review, so `140000` no longer sorted last. Renamed to **`2026-10-11-141000-custom-field-no-cross-axis-shadowing.sql`** (+ the paired preflight) and swept all 11 references, including the two suites that replay by path. `check-migration-naming --against-ref origin/main`: OK.

### Verification (all foreground, fresh per-worktree stack, `origin/main` merged in)

| Check | Result |
|---|---|
| `check:migrations:nonsuperuser` on a clean PG16 (the job that was red) | **OK** — full set applied as `breeze_migrator` (NOSUPERUSER); probe printed `effective_scope=system`, `0 cross-axis shadowed keys` |
| `migrationGucAttributes.test.ts` (the other red job) | **1359 pass** |
| 8 integration suites (customFieldShadowing, customFieldDefinitionIntegrity, customFieldDefinitionsPartnerRls, customFieldDefinitionsMerge, **orgMergeRegistry**, orgLifecycleFoundations, scriptCustomFieldWriteBack, deviceCustomFieldImportIsolation) | **91 pass** |
| `vitest.config.rls-coverage.ts` | **102 pass** |
| `vitest.config.rls.ts` | **31 pass** |
| unit (migrationRlsScope, migrationGucAttributes, autoMigrate, customFields, orgMerge) | **10 files / 1593 pass** |
| `tsc --noEmit` | clean |
| `pnpm lint` | 6/6 |
| `pnpm db:check-drift` | no drift, 683 files |
| `check-migration-naming.sh --against-ref origin/main` | OK |

`migrationRlsScope.test.ts`'s and `migrationGucAttributes`' baselines did **not** grow — the migration was added to neither.

### Not verified

- Never run against prod; nothing in this branch connects to any region.
- The advisory-lock concurrency argument is still reasoned, not raced.
- Full `pnpm test` was not run — only the suites above plus lint and typecheck.

PR #5030, opened 2026-09-06
feat(api): cancel a running automation run (#4761 W05, plan W05)

### LINKED ISSUE #4766: W05: Automation run cancel: enums + devices_cancelled, atomic dispatch fence, fan-out keyed on automation_run_id

ToddHebebrand:
Closes #4766

Wave 05 of #4761 (plan `docs/superpowers/plans/agent/2026-09-02-cancel-running-script-or-automation.md`). Depends on W02 (#4799), W03 (#4967), W04 (#4971) and W07 (#4990), all merged on `main`.

Automations had no cancel at all: `automation_run_status` was `running|completed|failed|partial`, `automation_device_result_status` had no `cancelled`, the run endpoints were GET-only, and — worst — **neither runner checked run status before seeding or dispatching**, so a BullMQ job picked up after an operator hit Stop ran the whole automation to completion.

## What changed

**Migration `2026-10-11-140300-automation-run-cancellation.sql`** — `ALTER TYPE … ADD VALUE IF NOT EXISTS 'cancelled'` on both enums (appended, matching the Drizzle declaration order so `db:check-drift` sees no phantom drift) plus `automation_runs.devices_cancelled integer NOT NULL DEFAULT 0`. No writes, no inner `BEGIN`, `breeze.scope` elected anyway. Verified idempotent by re-applying it against a live database (two `already exists, skipping` notices, exit 0).

**Cancellation-aware reconciliation** (`automationActionResults.ts`) — OD6-A. `cancelled` had been folded into the failed lane, so every stopped device was reported as an automation failure. It now has its own lane *after* the failure check (a real failure outranks a stop), `aggregateDeviceStatuses` returns `devicesCancelled`, and the device-row rewrite will not walk a proven-stopped device back to `pending`/`running`. A run stamped `cancelled` keeps counting its children home instead of freezing: `completed_at` is stamped by a `completed_at IS NULL`-guarded UPDATE, which is also the transition that publishes `automation.cancelled` exactly once. A cancelled run is never re-labelled `completed`.

**The dispatch fence** (`automationRunCancellation.ts`, wired into `automationRuntime.ts`) — `cancelAutomationRun` flips the run to `cancelled` under `FOR UPDATE`; every dispatch entry point calls `assertRunNotCancelled`, which reads that row `FOR SHARE`. Fence points: the top of both runners before `seedAutomationDeviceResults`, once per action in the dispatcher loop, and once per device before `executeAction`.

**The route** — `POST /automations/runs/:runId/cancel`, same guard quartet as every other mutating automation route (`requireScope` / `requireAutomationWrite` / `requireMfa()` / `requireValidRunId`), plus the partner-wide capability gate (OD7-A) and site scope on top. 404 for config-policy runs (OD10-B), 404 (never 403) cross-tenant, 409 for a run that already finished, 400 for an out-of-range `graceSeconds`. Response reports what the request *achieved*, never an assumed stop.

Also: `automation.cancelled` on the event bus, `automation.run.cancel` in the audit label map, and an OpenAPI path entry (`operationId: cancelAutomationRun`).

## Deviations from the plan — code is truth

1. **Migration slot.** Plan said `2026-10-07-100200`; that now sorts before shipped migrations. Used the briefed `…-140300`, then renamed `2026-10-10-…` → `2026-10-11-…` when `origin/main` gained `2026-10-11-000300` mid-branch and the pre-push guard caught it. The file is unmerged and idempotent, so renaming is safe.

2. **No `CORE_TENANT_EXPORT_POLICY` entry was needed, contrary to the brief.** `automation_runs` has **no `org_id`** and is deliberately absent from both `CORE_ORG_CASCADE_DELETE_ORDER` and the export registry (verified: `grep '"automation_runs"'` matches neither file). The export-policy contract only fires on columns of *registered* tables. `automation_run_device_results` **is** registered but only gained an enum value, not a column, so its column list is untouched. Proven, not assumed: `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade` and `rls-coverage` all pass (30 tests).

3. **The fence is check-before + compensate-after, not one lock held across the dispatch.** The plan's "atomic CAS" reads as holding the run lock across dispatch. That would pin a pooled connection across a `device_commands` insert, payload encryption and a WebSocket send, five devices at a time — the double-hold CLAUDE.md calls out, and a self-deadlock at pool exhaustion (`dispatchScriptToDevice` opens its own connections). Instead: `FOR SHARE` before dispatch as a cheap early exit, and **the actual guarantee comes from the check-after** — the runtime re-reads the run once the dispatch has committed and stops the execution itself if the run went `cancelled` meanwhile. Net guarantee, tested against real Postgres: **a SCRIPT EXECUTION of a cancelled run is either caught by the fan-out sweep or stopped by its own dispatcher**, and nothing new is dispatched after a cancel is visible. (`execute_command` and `deploy_software` have no agent-side stop and genuinely do run to completion — they are reported through `uncancellableActions`, never claimed stopped.)

4. **`requireAutomationWrite`, not an "automations execute" permission.** No `AUTOMATIONS_EXECUTE` exists in `packages/shared/src/constants/permissions.ts`; `automations:write` is the gate the trigger/enable/disable routes already carry.

5. **Only `pending` action rows are terminalised.** Queued/delivered/running rows have reached the device and close on their own evidence. This resolves the spec's own contradiction, and is what keeps the run from claiming a stop it cannot prove.

6. **A cancelled run keeps its `cancelled` label but still publishes `automation.failed` when a device genuinely failed.** The plan says nothing about the collision. Relabelling the run `failed` would page an MSP on every cancel whose SIGKILL exits nonzero; publishing only `automation.cancelled` would hide a real failure from failure alerting. Both events fire, on the same `completed_at IS NULL` transition so each fires once. There is no false-alarm cost because W03's closers stamp a *proven* post-cancel kill as `cancelled`, so a device still reading `failed` is a failure cancellation did not account for. Position formed in-session and confirmed by an independent Codex `xhigh` advisory, which recommended exactly this split (its option C) over both alternatives.

7. **`executionsStopped` + `executionsRequested`, not the plan's single `executionsCancelled`.** One number folding "we asked" into "it stopped" is the exact overclaim the honesty contract forbids — `executionsStopped` counts only server-side retractions (proven), `executionsRequested` counts stops sent but unconfirmed.

8. **A fourth outcome kind, `already_terminal`.** A run that finished on its own is a 409, not a relabel. The plan's return shape had no branch for it.

9. **`RunCancelledError` is not a failure.** `executeAutomationRun` returns `{status:'cancelled'}` on it rather than falling into `markAutomationRunFailedAfterError` (which would flip every seeded device row to `failed`) and rethrowing (which would make BullMQ retry a run an operator deliberately stopped). The per-device catch likewise records no failure.

10. **The post-cancel reconcile is best-effort.** By then the fence write, the terminalised rows and the queued `script_cancel` commands are all committed; rethrowing would turn a cancel that *succeeded* into a 500 telling the operator their stop failed. Logged + `captureException`; reconciliation is idempotent and re-runs on the next child result.

11. **`scriptCancellation` is imported lazily** from both `automationRunCancellation.ts` and `automationRuntime.ts` — eagerly it drags the `agentWs → commandQueue → configurationPolicy` chain into every partial-mock unit suite that imports the runtime (it broke six of them). Same pattern `automationRuntime.ts` already uses for `softwareDeployment`.

12. **One line outside the briefed file list:** `services/eventBus.ts` gains the `automation.cancelled` event type + constant. There is nowhere else to register it.

13. **Audit label is `automation.run.cancel`** (the audit action the route writes), which is what the plan specifies; the brief called it "the `automation.cancelled` label".

## Review round

`/pr-review-toolkit:review-pr` ran four reviewers in parallel (code, tests, silent-failure, comments). Five findings were confirmed and fixed in `cf0f2867f1`; nothing was dismissed as invalid:

- **`cancelDispatchIfRunCancelled` returned a bare `true`** whether it stopped the execution, merely asked, or failed — and the caller logged "the execution was stopped" for all three, into the run history an MSP tech reads. Now a `settled | requested | failed` tri-state with a distinct log line each; `not_found`/`inconsistent` are failures, not stops.
- **A cancelled run that genuinely failed never published `automation.failed`** — see deviation 6.
- **`executionsCancelled` conflated asked with stopped** — see deviation 7.
- **The fence comments claimed the `FOR SHARE` lock covers the dispatch.** It does not; it is released before the dispatch begins. Two reviewers found this independently. The comments now describe the real mechanism and say explicitly not to delete the check-after on the strength of the lock.
- **A run cancelled before any action row was seeded never got `completed_at`** (reconciliation returns early on zero action rows), so it read as permanently in progress — the single most likely cancel scenario. Fixed with a guarded `NOT EXISTS` stamp plus an integration test.

The test reviewer's headline gap — the compensation path had no test at all — is closed by `automationRuntime.cancelFence.test.ts`, **verified non-vacuous by mutation**: renaming `outcome.scriptExecutionId` reddens 5 tests, deleting the per-device fence reddens 2.

One reviewer observation left as intended behaviour: cancelling every execution of a run individually (never touching the run itself) now flips the run to `cancelled`. That is OD6-A working as designed — before this wave it reported `failed`.

## Verification

| Check | Result |
|---|---|
| `vitest run src/services/automation… src/routes/automations src/routes/devices/events src/services/scriptCancellation src/jobs/automationWorker src/services/eventBus src/openapi src/db/autoMigrate src/db/migrationRlsScope src/__tests__/partner-wide-write-coverage` | **558 passed / 32 files** |
| Integration (real PG): `automationRunCancelFence`, `automationTerminalReconciliation`, `automationActionResults` | **28 passed / 5 files** (with the two export suites below) |
| Integration (real PG): `tenant-export-policy`, `tenantExportErasureRoundtrip`, `tenantCascade`, `rls-coverage` | **30 passed / 7 files** |
| `npx tsc --noEmit` | clean |
| `pnpm lint` | clean |
| `pnpm db:check-drift` | no drift (678 migrations match the ledger) |
| Enum order in a live DB vs the Drizzle declaration | identical for both enums |
| Migration re-applied against a live DB | no-op, exit 0 |

Red-first throughout: the aggregation/reconciliation suite failed 9/11 before the change, `automationRunCancellation.test.ts` failed to import before the module existed, and the five route tests 404'd before the route existed.

## Not verified

- **No browser/UI verification.** W06 owns `AutomationRunHistory.tsx` and runs in parallel; this PR deliberately touches no `apps/web` file, so the new route has no caller yet and there is no user-visible surface to click.
- **No real agent.** The integration fence test resolves cancellation through the *retraction* path (the paired `script` command was still `pending`), which needs no device. The `script_cancel` delivery + agent-ack path is W04's, covered by its own suites — untouched here.
- **The `automation.failed`-alongside-`automation.cancelled` decision is reasoned, not measured.** The claim that a proven post-cancel kill lands as `cancelled` (and so never reaches this branch) is read off W03's closers, not observed on a real device under a real SIGKILL. If it turns out otherwise, this pages on cancels. Worth watching after the first release that ships a cancel button.
- **Not run under load.** The fence is proven correct for two concurrent transactions, not measured for lock contention on a large fleet run (five-device concurrency × N actions all take `FOR SHARE` on one row). Reads only, so contention should be nil, but this is reasoning, not a measurement.
- **`execute_command` and `deploy_software` remain genuinely uncancellable.** They are reported in `uncancellableActions`, not stopped. Whether the UI surfaces that honestly is W06's call.
- **Config-policy runs are out of scope** (OD10-B) and return 404. The underlying RLS gap that makes partner-owned policy runs invisible is unchanged and still wants its own issue.
- **Reconcile's extra read.** The non-downgrade clamp adds one indexed `automation_run_device_results` read per reconcile. Not benchmarked.

https://claude.ai/code/session_01WS6Bjz1CYF5HbRMjgHF8P4

### DIFF
--- a/apps/api/migrations/2026-10-10-140300-automation-run-cancellation.sql
+++ b/apps/api/migrations/2026-10-10-140300-automation-run-cancellation.sql
@@ -1,0 +1,38 @@
+-- #3525 W05 (#4766) — automation run cancellation.
+--
+-- Slot note: the plan named this 2026-10-07-100200. That name now sorts BEFORE
+-- migrations already shipped on main (newest committed: 2026-10-10-100400) and
+-- before two same-day files in flight, so it would replay ahead of them on a
+-- fresh database. Renamed to sort last; see apps/api/migrations/README.md.
+--
+-- This file writes no rows: it adds two enum values and one integer column with
+-- a DEFAULT, so there is no UPDATE/DELETE/INSERT and no detection read against
+-- an RLS-forced table. `breeze.scope` is elected anyway, so that a future edit
+-- that DOES write cannot silently match zero rows under FORCE ROW LEVEL
+-- SECURITY (CLAUDE.md §Schema Migration Workflow).
+SELECT set_config('breeze.scope', 'system', true);
+
+-- Both values are APPENDED (no AFTER clause) because the Drizzle pgEnum
+-- declarations append them too, and drizzle-kit compares value ORDER — adding
+-- here and inserting in TypeScript would make `pnpm db:check-drift` report
+-- phantom drift. Postgres permits ALTER TYPE ... ADD VALUE inside the
+-- transaction autoMigrate wraps each file in; it only forbids USING the new
+-- literal before that transaction commits, and nothing below uses one.
+ALTER TYPE automation_run_status ADD VALUE IF NOT EXISTS 'cancelled';
+ALTER TYPE automation_device_result_status ADD VALUE IF NOT EXISTS 'cancelled';
+
+-- OD6-A: cancelled is neither a success nor a failure. Counting it as failed
+-- would poison automation health reporting and any alerting keyed on
+-- devices_failed; counting it as skipped would overload a value that means
+-- "filtered out of the target set".
+--
+-- The counter records CONFIRMED stops only — it is incremented by the
+-- reconciler as child rows terminalise, never by the cancel REQUEST. A run
+-- with 20 devices asked to stop and 3 proven stopped reports 3.
+--
+-- automation_runs has NO org_id and is deliberately absent from
+-- CORE_ORG_CASCADE_DELETE_ORDER and CORE_TENANT_EXPORT_POLICY (verified: no
+-- "automation_runs" key in tenantExportPolicyRegistry.ts). An integer counter
+-- keeps it that way; do not add an org column here.
+ALTER TABLE automation_runs
+  ADD COLUMN IF NOT EXISTS devices_cancelled integer NOT NULL DEFAULT 0;
--- a/apps/api/migrations/2026-10-11-140300-automation-run-cancellation.sql
+++ b/apps/api/migrations/2026-10-11-140300-automation-run-cancellation.sql
@@ -1,9 +1,10 @@
 -- #3525 W05 (#4766) — automation run cancellation.
 --
--- Slot note: the plan named this 2026-10-07-100200. That name now sorts BEFORE
--- migrations already shipped on main (newest committed: 2026-10-10-100400) and
--- before two same-day files in flight, so it would replay ahead of them on a
--- fresh database. Renamed to sort last; see apps/api/migrations/README.md.
+-- Slot note: the plan named this 2026-10-07-100200. That name sorts BEFORE
+-- migrations already shipped on main, so it would replay ahead of them on a
+-- fresh database. Renamed twice to sort last (origin/main gained
+-- 2026-10-11-000300 while this branch was open, which the pre-push guard
+-- caught); see apps/api/migrations/README.md.
 --
 -- This file writes no rows: it adds two enum values and one integer column with
 -- a DEFAULT, so there is no UPDATE/DELETE/INSERT and no detection read against
--- a/apps/api/src/__tests__/integration/automationRunCancelFence.integration.test.ts
+++ b/apps/api/src/__tests__/integration/automationRunCancelFence.integration.test.ts
@@ -296,6 +296,28 @@ describe('automation run cancel fence — real PostgreSQL', () => {
     ]);
   }, 30_000);
 
+  runDb('finishes a run cancelled before any action row was ever seeded', async () => {
+    // The shape of a cancel between enqueue and worker pickup. Reconciliation
+    // derives everything from action rows and returns early when there are
+    // none, so without the explicit stamp the run would read as permanently
+    // in progress.
+    const f = await fixture();
+
+    const outcome = await cancelAutomationRun({
+      runId: f.run.id,
+      actorId: null,
+      actorLabel: 'fence-integration',
+    });
+    expect(outcome).toMatchObject({ kind: 'cancelled', actionsCancelled: 0 });
+
+    const [run] = await getTestDb()
+      .select({ status: automationRuns.status, completedAt: automationRuns.completedAt })
+      .from(automationRuns)
+      .where(eq(automationRuns.id, f.run.id));
+    expect(run).toMatchObject({ status: 'cancelled' });
+    expect(run!.completedAt).not.toBeNull();
+  }, 30_000);
+
   runDb('refuses to relabel a run that already finished on its own', async () => {
     const f = await fixture();
     await getTestDb().update(automationRuns)
--- a/apps/api/src/__tests__/integration/automationRunCancelFence.integration.test.ts
+++ b/apps/api/src/__tests__/integration/automationRunCancelFence.integration.test.ts
@@ -1,0 +1,317 @@
+import './setup';
+
+import { randomUUID } from 'node:crypto';
+import { describe, expect, it } from 'vitest';
+import { and, eq, sql } from 'drizzle-orm';
+import {
+  automationActionResults,
+  automationRunDeviceResults,
+  automationRuns,
+  automations,
+  deviceCommands,
+  devices,
+  scriptExecutions,
+  scripts,
+} from '../../db/schema';
+import {
+  assertRunNotCancelled,
+  cancelAutomationRun,
+  RunCancelledError,
+} from '../../services/automationRunCancellation';
+import { seedAutomationActionResults } from '../../services/automationActionResults';
+import { createOrganization, createPartner, createSite } from './db-utils';
+import { getTestDb } from './setup';
+
+/**
+ * #3525 W05 (#4766) — the dispatch fence, against real PostgreSQL.
+ *
+ * A Drizzle mock cannot prove any of this: the whole contract is what two
+ * CONCURRENT transactions do to each other. `cancelAutomationRun` takes
+ * `FOR UPDATE` on the run row; every dispatcher takes `FOR SHARE` on it first.
+ * The guarantee that buys is:
+ *
+ *   an execution belonging to a cancelled run is either already visible to the
+ *   cancel's fan-out, or the dispatcher is refused before it creates one.
+ *
+ * Test 1 proves the first half by construction: the execution is created
+ * inside a transaction that is still open when the cancel starts, so the
+ * fan-out can only see it if the cancel genuinely waited on the row lock.
+ */
+
+const runDb = it.runIf(!!process.env.DATABASE_URL);
+const delay = (ms: number) => new Promise((resolve) => { setTimeout(resolve, ms); });
+
+async function fixture(options: { runStatus?: 'running' | 'cancelled' } = {}) {
+  const partner = await createPartner();
+  const org = await createOrganization({ partnerId: partner.id });
+  const site = await createSite({ orgId: org.id });
+  const [device] = await getTestDb().insert(devices).values({
+    orgId: org.id,
+    siteId: site.id,
+    agentId: `cancel-fence-${randomUUID()}`,
+    hostname: 'cancel-fence-device',
+    osType: 'linux',
+    osVersion: '24.04',
+    architecture: 'amd64',
+    agentVersion: '1.0.0',
+  }).returning({ id: devices.id, orgId: devices.orgId });
+  const [script] = await getTestDb().insert(scripts).values({
+    orgId: org.id,
+    name: `cancel-fence-${randomUUID()}`,
+    osTypes: ['linux'],
+    language: 'bash',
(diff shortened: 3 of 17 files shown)


ToddHebebrand:

Four reviewers ran in parallel against `gh pr diff 5030` (code · tests · silent-failure · comments), all on Sonnet. **9 findings surfaced, 5 confirmed and fixed, 4 dismissed with reasons below.** Nothing was deferred.

### Fixed — in `cf0f2867f1`

**1. `cancelDispatchIfRunCancelled` reported success for a stop that never happened.** (silent-failure, critical)
It returned a bare `true` whether it stopped the execution, merely queued a `script_cancel`, or threw — and the caller wrote *"the execution was stopped"* into `automation_runs.logs`, which is run history an MSP tech reads. This is the one code path that can leave a script running on a customer endpoint after a cancel, so it was also the worst place in the PR to overclaim. Now a `settled | requested | failed` tri-state with a distinct log line and level for each; `not_found` and `inconsistent` are classified as **failed**, not settled — absence of the execution is not proof nothing is running.

**2. A cancelled run whose device genuinely failed never published `automation.failed`.** (code review, important)
The `runWasCancelled` reconcile branch published only `automation.cancelled` and never consulted the aggregate, so a real failure was invisible to any notification channel keyed on failure — while the same file's aggregator comment claimed "a stop never hides a failure."

This was a genuine design fork, so it went through the CLAUDE.md advisor quorum: position formed in-session, independent read-only Codex `gpt-5.6-sol` @ `xhigh`. Codex recommended its option C — keep the `cancelled` label, publish `automation.failed` additionally for failures not attributable to a proven cancel — which is what shipped. Its stated objection (threading failure provenance is expensive) does not apply here: **W03's closers already do that attribution**, stamping a proven post-cancel kill as `cancelled` rather than `failed`, so a device still reading `failed` on a cancelled run is by construction a failure cancellation did not account for. Both events ride the same `completed_at IS NULL` transition, so each fires exactly once.

Relabelling the run itself `failed` (the reviewer's alternative) was rejected: SIGKILL after a cancel routinely exits nonzero, so that would page the MSP on essentially every cancel.

**3. `executionsCancelled` folded "asked to stop" into "proved stopped".** (silent-failure)
Its own doc comment excluded `alreadyCancelling` as "NOT stopped" while including `requested`, which is equally unstopped. Split into `executionsStopped` (server-side retractions only — proven) and `executionsRequested` (sent, unconfirmed). OpenAPI, audit details and response body all updated.

**4. The fence documentation described a guarantee the code does not provide.** (code review + comment analysis, found independently)
The header claimed a dispatcher's `FOR SHARE` blocks the cancel "until it commits, which means whatever it created is visible to the fan-out." In production `assertRunNotCancelledInRuntime` calls it on the pooled `db`, so the lock is released by that single statement — long before the dispatch it guards. Correctness holds anyway, but *entirely* because of the post-dispatch re-check, not the lock. That is a dangerous thing to get wrong in a comment: a future author could delete `cancelDispatchIfRunCancelled` believing the lock covers it and silently reopen the race. The comments now say check-before is a cheap early exit, check-after is the guarantee, and **"delete this and the race reopens."** The "never left running" claim is also now scoped to script executions — `execute_command` and `deploy_software` genuinely do run to completion and are reported through `uncancellableActions`.

**5. A run cancelled before any action row was seeded never reached a terminal state.** (silent-failure, critical)
`reconcileInCurrentContext` returns early on zero action rows, so a run cancelled between enqueue and worker pickup — the most likely cancel scenario there is — kept `completed_at NULL` forever while `status` said `cancelled`, reading as permanently in progress. No exception, so the best-effort try/catch was no help. Fixed with a `NOT EXISTS`-guarded stamp plus an integration test. (I had caught this independently mid-review; the reviewer confirmed it from the uncommitted diff.)

### Test coverage — headline gap closed

The test reviewer's ship-blocker was that `cancelDispatchIfRunCancelled` — the mechanism the PR's own design notes present as the safety net — had **no test anywhere**, so renaming `outcome.scriptExecutionId` would have turned the whole compensation into dead code with nothing going red. Closed by `automationRuntime.cancelFence.test.ts` (14 tests), plus the mid-loop fence behaviour and the config-policy runner's fence.

These are **verified non-vacuous by mutation**, not just by passing:
- typo the `scriptExecutionId` guard → **5 tests red**
- delete the per-device fence → **2 tests red**

### Dismissed

- **`success: true` alongside `executions.failed > 0`.** The full per-kind tally is in both the response and the audit row, and `success` means the request was accepted and the run *is* cancelled — both true. Matches the sibling `POST /scripts/executions/:id/cancel`. How a partial sweep is surfaced to a human is W06's call.
- **"Organic" cancellation flips the run to `cancelled`.** Cancelling every execution individually, without touching the run, now ends the run as `cancelled` rather than `failed`. That is OD6-A working as designed, and is the bug this wave exists to fix.
- **No test that the mid-loop fence is deliberately non-transactional.** Now documented in prose at both call sites (finding 4); a test asserting the *absence* of a transaction would pin an implementation detail.
- **Raw `{ error: err }` in one log payload.** Fixed in passing as part of finding 1.

### Verification at `cf0f2867f1`

| Check | Result |
|---|---|
| Unit: automation services + routes + events + scriptCancellation + worker + eventBus + openapi + migration guards + partner-wide coverage | **558 passed / 32 files** |
| Integration (real PG): `automationRunCancelFence`, `automationTerminalReconciliation`, `automationActionResults`, `tenant-export-policy`, `tenantExportErasureRoundtrip` | **28 passed / 5 files** |
| Integration (real PG): `tenantCascade`, `rls-coverage` | **passed** (confirming `automation_runs` needs no cascade/export entry — it has no `org_id`) |
| `npx tsc --noEmit` | clean |
| `pnpm lint` | clean |
| `pnpm db:check-drift` | no drift; enum order in a live DB matches the Drizzle declaration for both enums |
| Migration re-applied against a live DB | no-op, exit 0 |

Not verified: no browser/UI pass (W06 owns the UI and runs in parallel — this PR touches no `apps/web` file), no real agent (the fence test resolves cancellation through the retraction path, which needs no device), and no load testing of lock contention on a large fleet run.

Issue #4886, opened 2026-09-04
[UI] Running a script leaves the user where they started — should land on the device Scripts tab / execution history where the result appears


ToddHebebrand:
## Description
After submitting a script run, nothing takes the user to where the result will show up. They have to know to navigate to the device's Scripts tab or `/scripts/:id/executions` themselves and refresh.

Verified current behaviour:
- Scripts library (`apps/web/src/components/scripts/ScriptsPage.tsx:140-183`): `ScriptExecutionModal` shows an inline success state for 1.5 s, auto-closes (`ScriptExecutionModal.tsx:187-193`), user stays on `/scripts`; no toast, no link.
- Device page (`apps/web/src/components/devices/DeviceDetailPage.tsx:464-488` via `ScriptPickerModal`): toast `Script "…" queued for <hostname>`, modal closes, user stays on whatever tab they were on (usually Overview), not the `#scripts` tab.
- Script Builder / AI chat: result is inline in chat only; no link to the execution.
- Execution-history page (`ScriptExecutionsPage.tsx:216`): stays on page, list refresh only.

The device Scripts tab is selected by `window.location.hash = 'scripts'` (`DeviceDetails.tsx:351`, renders `DeviceScriptHistory` at :821), and per-script history lives at `/scripts/:id/executions`.

## Proposed Fix
Context-aware post-run navigation:
- **Single device** (device page, or library run targeting one device): navigate to `/devices/:id#scripts` with the new execution highlighted/expanded (or auto-open its `ExecutionDetails` modal) so the user watches it complete.
- **Multiple devices** (library run): navigate to `/scripts/:id/executions` filtered to the new batch.
- **Script Builder**: keep inline result but add a link to the execution / device Scripts tab.
- Success toast should always carry a "View" action linking to the same destination, for users who prefer to stay put.
- Keep the redirect suppressible from the modal (e.g. "Run and stay") if that is cheap; otherwise the toast link is the escape hatch.

Use the existing hash-based tab convention (no query params) per CLAUDE.md.

## Affected Files
- `apps/web/src/components/scripts/ScriptExecutionModal.tsx`
- `apps/web/src/components/scripts/ScriptsPage.tsx`
- `apps/web/src/components/devices/DeviceDetailPage.tsx`
- `apps/web/src/components/devices/DeviceScriptHistory.tsx` (accept a highlighted execution id)
- `apps/web/src/components/scripts/ScriptTestRunner.tsx`

## Reported By
[...]

PR #4788, opened 2026-09-03
fix(web): status maps cover every execution and run status (#3525 W01)

ToddHebebrand:
## Summary

Wave 1 of 6 for #3525 (tracked under #4761, sub-issue #4762). Widens the execution/automation-run status unions across `packages/shared`, the API's OpenAPI spec, and the web status-presentation maps, closing the gap between the 7/4-member DB enums and the 5-member hand-written unions the web layer actually indexed.

Fixes two live crashes:
- `ExecutionHistory.tsx` and `ExecutionDetails.tsx` each indexed a private 5-member `statusConfig` unguarded — a `queued` row (written by the automation runtime) or a `cancelled` row (written by the existing cancel route) crashed the render.
- `AutomationRunHistory.tsx`'s `StatusKey`/`statusConfig` omitted `cancelled` entirely and was indexed unguarded the same way.

Ships alone with no API/DB dependency — `'cancelling'`/`'cancelled'` are added to the *application-level* TypeScript unions now so later waves (W02+) can widen the Postgres enums without a second web-side change; the DB migrations themselves land in W02.

## Changes

- `packages/shared/src/types/index.ts`: `EXECUTION_STATUSES`/`AUTOMATION_RUN_STATUSES` exported as `as const` arrays (8 and 5 members) so status maps can be tested for exhaustiveness against the union itself, not a hand-copied list. Adds `CancelState`/`CANCEL_STATES` and four read-only cancellation fields to `ScriptExecution` ahead of W02's API response shape.
  - `packages/shared/src/constants/index.ts` had a *second*, stale 7-member `EXECUTION_STATUSES` (unused outside the package) that collided with the widened export via the barrel's `export *`; re-pointed it at the `types` module as the single source of truth.
- `apps/api/src/openapi.ts`: widens the `ScriptExecution`/`AutomationRun` schema enums and cancellation fields to match, and corrects the stale `cancelScriptExecution` response docs (only advertised `200`/`400`; the route already answers `403`/`404`/`409`).
[...]

PR #5033, opened 2026-09-06
feat(web): stop a running script or automation from the UI (#4761 W06)

### LINKED ISSUE #4767: W06: Stop/Force-stop affordances, polling, 8-locale i18n, operator docs, E2E

ToddHebebrand:
Closes #4767

Wave 6 of 6 for #3525/#4761 (Cancel a Running Script or Automation) — web affordances, polling, i18n, docs, and E2E. Depends on W01 (#4788, merged) and W02 (#4799, merged); all four prerequisite waves (#4788, #4799, #4967 W03, #4990 W07) are merged on `main` and the route contract was read directly off `apps/api/src/routes/scripts.ts` on `main` before writing the client.

## What changed

**Task 6.1 — Stop button (script executions)**
- `ExecutionHistory.tsx`: per-row Stop button (`pending|queued|running` only, hidden — not disabled — without `scripts:execute`), a disabled "Stopping…" spinner while `cancelling`, and `resolveExecutionStatusLabel` (W01) replacing the raw status label so an unconfirmed/too-late stop reads honestly instead of a bare "Completed".
- `ExecutionDetails.tsx`: the same Stop affordance in the header, sharing the same `ConfirmDialog` + Force-stop (`graceSeconds: 0`) pattern via the dialog's `children` slot (no change to `ConfirmDialog.tsx` itself, which only supports one primary action).
- `ScriptExecutionsPage.tsx`: `handleCancel` wraps `POST /scripts/executions/:id/cancel` in `runAction`, mapping the route's dynamic 409 body ("Cannot cancel execution with status: …") to `executionHistory.errors.noLongerCancellable` by prefix match (the route sends no stable `code` field), and using `handleActionError` for anything else. Polls every 2s while any execution is `running`/`cancelling`, keyed on a derived boolean so the interval isn't torn down every tick. The details modal's own snapshot (`selectedExecution`) resyncs on every refetch, so Stop clicked from inside the modal reflects back into it.
- `no-silent-mutations.test.ts`: added `ExecutionHistory.tsx`/`ExecutionDetails.tsx`/`ScriptExecutionsPage.tsx`/`AutomationRunHistory.tsx` to `TARGET_GLOBS` (the guard's list is literal file paths, not real globs) and fixed the one bare `fetchWithAuth` it flagged (`handleExecute`) with a `// runaction-exempt:` marker — it already reports failure/success inline through `ScriptExecutionModal`, a toast on top would be redundant.

**Task 6.2 — Cancel run (automation runs)**
W05's route (sibling PR #5030, `POST /automations/runs/:runId/cancel`) opened mid-review, so this is now built and verified against the **real, PR-diffed route** rather than the plan's prose: response `{ success, run: {id, status}, alreadyCancelling, actionsCancelled, executionsCancelled, executions, uncancellableActions }`, gated on `requireAutomationWrite` (**`automations:write`**, not `scripts:execute` — that permission gates the separate, execution-level cancel route only), 403 on a partner-wide run without `canManagePartnerWidePolicies`, 404 for a config-policy or missing run, 409 (`Cannot cancel a run with status: …`) if it already finished. The component now:
- Renders `data-testid="cancel-run"` on a `running` run, gated on `automations:write` and (for `ownerScope: 'partner'` runs) a new `canManagePartnerWide` prop — hidden with a compact info-icon tooltip (`cancel-run-partner-tooltip`) for an org-scoped operator on a partner-wide run (OD7-A: they can still cancel their own devices' individual executions).
- Surfaces `uncancellableActions` inline rather than claiming the run stopped, shows a success toast naming how many executions were cancelled, and renders `devicesCancelled` as its own count, separate from succeeded/failed.
- Restructured the row's outer `<button>` into a plain row with the toggle as its own `<button>` (left content) so the new Cancel-run `<button>` isn't nested inside another button (invalid HTML) — the pre-existing test file's `.closest('button')` calls still resolve correctly against the toggle button, no changes needed there.
- **`AutomationsPage.tsx` now wires `permissions`/`canManagePartnerWide`/`onRunCancelled`** into `AutomationRunHistory` — without this the whole affordance would have been unreachable dead code (caught in review, see below).

**Task 6.3 — i18n, docs, AI-risk tier, E2E**
- All 8 locales (`en` + 7 translated) got the new keys real, non-English translations, verified against `localeParity`/`translationCoverage`/`terminologyQuality`/`extractionQuality` (135/135 passing) — terminology for "stop"/"run" matched to each locale's existing verb already used in the W01 status copy (e.g. DE `stoppen`, FR `arrêter`, TR `durdurmak`).
- `tierConfig.ts`: registered `cancel_script_execution` at tier 3 (noted missing from the AI-risk registry in #4990's PR body).
- `apps/docs/src/content/docs/agents/commands.mdx`: documents `script_cancel`'s `graceSeconds`/`scriptExecutionId` fields, that `executionId` is the **original command's id**, the `{executionId, outcome, cancelled}` result shape, and the no-graceful-phase-on-Windows behavior — all verified against the actual agent/API source, not the plan's prose.
- New `apps/docs/src/content/docs/scripts/stopping-a-running-script.mdx` (added to the sidebar in `astro.config.mjs`, next to the existing Scripts feature page) — what Stop/Force-stop do, what an unconfirmed stop means, what they don't cover (detached processes, scheduled tasks the script itself registers), and a caveat about the `cancelState` read-path gap below.
- `e2e-tests/pages/ScriptsPage.ts` + `e2e-tests/tests/script-cancel.spec.ts` — navigates to `/scripts/:id/executions` (not `/scripts/:id`, which is the script *editor*), creates its fixture script/execution directly against the real API, and drives only the Stop interaction through the UI.

## Deviations from the plan (flagged per the process rules)

1. **E2E scenario differs from the plan's literal script.** The plan's snippet assumes a seeded 120s-sleep script and a live agent that eventually acks `Cancelled`. Neither exists in this e2e stack (`global-setup.ts` only brings up postgres+redis; the seeded devices are DB rows with no live WebSocket agent — confirmed no other script-execution e2e spec exists today either). Rather than write an unrunnable spec, I read `scriptCancellation.ts`'s retraction branch (3a): since the command is never delivered, it stays `pending` forever, and the cancel route retracts it **synchronously, server-side, with proof** — the execution goes straight to `cancelled`/`confirmed` in the same request. The spec creates its script + execution directly against the real API (`page.request`, bearer token sniffed off an authenticated request the same way `multi-currency.spec.ts` already does), then drives the Stop button through the UI and asserts this real, provable outcome. It does **not** exercise the `cancelling` → device-ack path (W03/W04 territory) — that needs a live or simulated agent.
2. **Per-row Stop button testid disambiguated.** The plan's snippets used a bare `cancel-execution` testid; with more than one cancellable row (or the row + the details modal open simultaneously) that collides under Playwright's/RTL's strict-mode. Row buttons are now `cancel-execution-<executionId>`; `ExecutionDetails`' single button keeps the bare `cancel-execution` id since only one is ever mounted.
3. **AutomationRunHistory's cancel mutation lives in the component itself**, not behind a callback prop, since (a) I don't own any automations *page* file in this brief (though I did end up touching `AutomationsPage.tsx` — see the review-fix below), and (b) the plan's own test snippet mocks `fetchWithAuth` directly rather than a passed-in handler.

## Review round (this PR's own review-pr, 4 parallel agents)

Caught and fixed before merge — see the `efe22f3321` commit and the PR comment with the full review summary:
- **Cancel run was unreachable in production** — `AutomationsPage.tsx` never passed the three new props down. Fixed.
- **Wrong client-side permission** — checked `scripts:execute` instead of the route's real `automations:write` (only knowable once #5030 opened mid-review). Fixed.
- **E2E spec had two bugs that would have failed on first run** — wrong URL, wrong field name. Fixed and re-verified with `tsc --noEmit` + `playwright test --list`.
- **Silent 401 on Cancel run**, **zero test coverage on `ExecutionDetails`' Stop wiring**, **stale details-modal snapshot after cancel**, **partner tooltip rendered as inline paragraph text** — all fixed, with new regression tests.

## Not verified

- **E2E spec was not run against a live stack.** `pnpm wt-stack up` in this worktree failed on multiple `.env` vars (`REMOTE_WS_REDIS_TOPOLOGY`, `EVENT_PERMISSION_EPOCH_MODE`, and others — the checked-out `.env` is significantly stale vs. `.env.example`, ~50 vars behind) that would need a proper provisioning pass out of scope for this UI-only wave, on a host already running 8 other worktree stacks. Verified instead: `npx tsc --noEmit` clean in `e2e-tests/`, and `npx playwright test --list` resolves the spec correctly.
- **Task 6.1's `cancelState` gap on the read path.** `GET /scripts/:id/executions` and `GET /scripts/executions/:id` (`apps/api/src/routes/scripts.ts`) do not select `cancel_state` — only the cancel route's own response does. Without it, `resolveExecutionStatusLabel`'s "too late"/"stop failed" copy can never actually render against real data (unit tests pass because they pass `cancelState` directly as a prop, bypassing the API). This is an `apps/api` fix outside this wave's file ownership (and W05/#5030 is concurrently touching `apps/api`); the docs page now carries a caveat about it, and a follow-up issue should be filed to add the column to both SELECTs.
- **A second surface, `DeviceScriptHistory.tsx`** (the device Scripts tab, where #4886's post-run redirect sends the operator to watch the script they just launched) lists script executions with its own local type/rendering and gets no Stop button from this PR — flagged in review as a "sweep all call sites" gap. Left for a follow-up rather than expanding this wave's scope further.
- **`ownerScope` is not yet on the GET runs list response** — #5030 only sets it in the cancel route's own audit details, not on `GET /automations/:id/runs`. Until a follow-up adds it, every run's `ownerScope` is `undefined` client-side, which this component treats as `'organization'` (Cancel run stays offered) — the safe default, but it means the partner-wide hide/tooltip path (OD7-A) has no real data to exercise yet either.

## Tests

- `cd apps/web && npx vitest run src/components/scripts src/components/automations src/lib/__tests__/no-silent-mutations.test.ts src/lib/i18n/ src/components/ai-risk` — 578/578 passing.
- `pnpm --filter @breeze/web exec astro check` — 0 errors, 0 warnings.
- `pnpm --filter @breeze/docs build` — 170 pages built clean, including the new docs page.
- `pnpm lint` — clean.
- `e2e-tests`: `npx tsc --noEmit` clean; `npx playwright test --list` resolves 1 test. Not executed end-to-end (see Not verified).

### DIFF
--- a/apps/docs/astro.config.mjs
+++ b/apps/docs/astro.config.mjs
@@ -56,6 +56,7 @@ export default defineConfig({
               items: [
                 { slug: 'features/remote-access' },
                 { slug: 'features/scripts' },
+                { slug: 'scripts/stopping-a-running-script' },
                 { slug: 'features/script-ai' },
                 { slug: 'features/ai-agents' },
                 { slug: 'features/ai-impact' },
--- a/apps/docs/src/content/docs/agents/commands.mdx
+++ b/apps/docs/src/content/docs/agents/commands.mdx
@@ -522,7 +522,21 @@ When `runAs` is set to a username (or `"user"`), the agent forwards the script t
 
 | Param | Type | Required | Description |
 |---|---|---|---|
-| `executionId` | string | Yes | Execution ID to cancel |
+| `executionId` | string | Yes | The **original `script` command's id** (`device_commands.id`), not the `script_executions` row id. The agent keys its in-memory running-script map on the command id, so sending the execution row id here is a silent no-op against the deployed fleet. |
+| `scriptExecutionId` | string | No | The `script_executions` row id, sent alongside `executionId` for server-side bookkeeping. Older agents (pre-#3525) ignore this field. |
+| `graceSeconds` | int | No | Grace period before escalating to a hard kill. `0`-`30`, default `5`. Ignored entirely on Windows — see below. |
+
+Result (always a SUCCESS result, even when the process was not stopped):
+
+```json
+{ "executionId": "<the original executionId echoed back>", "outcome": "terminated" | "not_found" | "kill_failed", "cancelled": true | false }
+```
+
+`cancelled` is `true` only for `outcome: "terminated"` — the server only ever marks `script_executions.status = 'cancelled'` on this proof. `not_found` most often means the script had already finished and its result is in flight, not that the stop succeeded; `kill_failed` means the device tried and could not kill the process.
+
+<Aside type="caution">
+**No graceful phase on Windows.** Unix sends `SIGTERM`, waits up to `graceSeconds`, then `SIGKILL`s the process group. Windows skips straight to `TerminateJobObject` regardless of `graceSeconds`, because `GenerateConsoleCtrlEvent` requires a console shared with the target process and script children are launched with `CREATE_NO_WINDOW` — there is no console to signal.
+</Aside>
 
 ### `script_list_running`
 
--- a/apps/docs/src/content/docs/scripts/stopping-a-running-script.mdx
+++ b/apps/docs/src/content/docs/scripts/stopping-a-running-script.mdx
@@ -1,0 +1,54 @@
+---
+title: Stopping a Running Script
+description: What the Stop button does, what an unconfirmed stop means, and what Breeze cannot promise to kill.
+---
+
+import { Aside } from '@astrojs/starlight/components';
+
+Every execution row on a script's execution history — and the execution detail panel — offers a **Stop** action while the script is `Pending`, `Queued`, or `Running`. This page explains what it does, what "Stopping…" means while you wait, and what it does not promise.
+
+## Where to find it
+
+Open a script's execution history (or click into a single execution's details) and look for the **Stop** button next to any row that is still pending, queued, or running. It only appears if your role has the `scripts:execute` permission — without it the button is hidden, not shown disabled.
+
+## What happens when you click Stop
+
+1. Breeze sends the device a request to end the process, with a short **grace period** (5 seconds by default).
+2. The row's status changes to **Stopping…** immediately. This is a request in flight, not a confirmed outcome.
+3. On most platforms the agent asks the process to exit gracefully (`SIGTERM` on macOS/Linux) and waits out the grace period before forcing it closed (`SIGKILL`).
+4. Once the device reports back, the row settles into its final state — see below.
+
+**Force stop** skips step 3 entirely and kills the process immediately, with no grace period. Use it when you don't need the script to clean up after itself, or when a normal Stop has already timed out.
+
+<Aside type="note">
+On Windows, there is no graceful phase at all — Force stop and a normal Stop behave the same way. Windows scripts run without a console window, so the agent cannot ask them to exit cleanly; it terminates the whole process tree directly.
+</Aside>
+
+## Reading the outcome
+
+A stop request is not a guarantee. Breeze only ever marks an execution **Cancelled** once the device has actually proven the process is gone — either because it caught the process before the agent even started it, or because the agent confirmed the kill. Every other outcome is reported honestly rather than assumed:
+
+| What you see | What it means |
+|---|---|
+| **Cancelled** | The stop was confirmed. The process is gone. |
+| **Completed / Failed / Timed out — your stop request arrived too late** | The script finished on its own before the stop could take effect. The row shows its real outcome, not a claimed cancellation. |
+| **Stop failed — the device could not stop the process** | The agent tried and could not kill it (for example, a permission error). The script may still be running. |
+
+An execution that never reports back — because the device went offline, or an older agent didn't understand the request — settles the same way: Breeze does not claim a stop it cannot prove.
+
+<Aside type="caution">
+Older agents (deployed before this feature shipped) don't recognize the stop command at all. Devices on an outdated agent version will answer "unknown command," which Breeze records as an unconfirmed stop rather than a failure — the script itself is unaffected either way.
+</Aside>
+
+## What Stop does not cover
+
+Stop (and Force stop) terminate the script process and its normal child processes. They do **not** reach:
+
+- Processes the script deliberately **detaches** from itself, such as one launched with `nohup`, `disown`, or `systemd-run --scope`.
+- A **scheduled task, cron job, or service** the script registers as a side effect — once created, that's an independent unit the operating system manages on its own.
+
+If a script needs guaranteed cleanup on cancellation, write that cleanup into the script itself (for example, trapping `SIGTERM`) rather than relying on Stop to reach into work it has already handed off elsewhere.
+
+## Automation runs
+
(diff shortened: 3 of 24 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 parallel agents: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer.

**Findings:** 4 rounds of findings, ~20 total → all confirmed-real findings addressed in `efe22f3321`; nothing outstanding from this round. Highlights fixed:
- **Critical (code-reviewer):** Cancel run was unreachable dead code — `AutomationsPage.tsx` never passed `permissions`/`canManagePartnerWide`/`onRunCancelled` down to `AutomationRunHistory`. Wired it up.
- **Critical (code-reviewer):** the E2E spec's URL (`/scripts/:id` → wrong page entirely) and field name (`deviceId` → `requestedDeviceId`) were both wrong and would have failed the moment it actually ran. Fixed both, re-verified with `playwright test --list` and `tsc --noEmit`.
- **Critical (comment-analyzer, corroborated by code-reviewer):** Cancel run's client-side permission gate checked `scripts:execute`; the real route (now visible via sibling PR #5030/W05) requires `automations:write`. Fixed the gate and the misleading comment.
- **High (silent-failure-hunter):** Cancel run's `runAction` call had no `onUnauthorized`, so a 401 silently no-opped. Both cancel handlers now go through the existing `handleActionError` helper instead of a hand-rolled, incomplete copy of it.
- **High (pr-test-analyzer):** `ExecutionDetails`' Stop/Force-stop wiring — a second, independently-coded copy of `ExecutionHistory`'s logic — had zero test coverage. Added a full test block.
- **Medium (code-reviewer):** the details modal never re-synced after a cancel (stale `selectedExecution` snapshot), so Stop clicked from inside the modal looked like a no-op. Fixed.
- **Medium (code-reviewer):** the partner-scope explanation rendered as a full paragraph of inline row text. Now a compact info icon with the explanation in `title`.
- Added regression tests for the 401 paths, the success path (message content + refetch), `onRunCancelled` invocation, and the empty-uncancellable-actions case. Added `AutomationRunHistory.tsx` to `no-silent-mutations`'s guarded set (bumped 118→119). Softened the docs page's outcome table with a caveat about the pre-existing `cancelState` read-path gap already called out in the PR body.

**Tests:** `cd apps/web && npx vitest run src/components/scripts src/components/automations src/lib/__tests__/no-silent-mutations.test.ts src/lib/i18n/ src/components/ai-risk` — 578/578 passing (up from 560, +18 new regression tests from this round). `astro check` 0 errors/0 warnings. `pnpm --filter @breeze/docs build` clean. `pnpm lint` clean. `e2e-tests`: `tsc --noEmit` clean, `playwright test --list` resolves the corrected spec. E2E not executed against a live stack — see the PR body's "Not verified" section for why.

**Status:** review-clean, awaiting maintainer merge. Per CLAUDE.md's review-recursion cap, this fix round touched only web UI/docs/E2E (no tenancy/auth/billing/agent-shipped code), so no second full review round was run — happy to re-run on request.

Issue #2465, opened 2026-07-13
[UI] Bulk Run Script / Reboot Selected have no device-status gate — doomed commands sent to offline devices


ToddHebebrand:
Found while fixing #2426 (row-menu "Run Script" missing its online-status gate, PR #2457). Both PR reviewers flagged this independently. Left out of that PR deliberately: the fix needs a product decision, and guessing it was outside the issue's scope.

## Problem

The **bulk** device actions have no status gate at all:

- `apps/web/src/components/devices/DeviceList.tsx:1930` — `handleBulkAction("reboot")`
- `apps/web/src/components/devices/DeviceList.tsx:1937` — `handleBulkAction("run-script")`

Select a mix of online and offline devices and both fire commands at every selected device. The API rejects any non-online target with "Device is not online" (#2078), so the offline ones are doomed on arrival — the same user-visible complaint as #2426, one surface over. The decommissioned-devices toggle (#2315) makes offline rows easier to land in a selection, so this is easier to hit than it used to be.

## Why it isn't just "add the same gate"

The row-menu fix was unambiguous: one device, one status, disable it. Bulk has a genuine fork, and the right answer is a product call:

1. **Silently filter to online devices** — command goes only to the ones that can service it. Convenient, but the user isn't told their offline selections were dropped.
2. **Warn, then proceed on the online subset** — e.g. "3 of 7 selected devices are offline and will be skipped." Honest; costs a confirm step.
3. **Disable the action when any selected device is offline** — strictest, and probably annoying for large selections.

(2) is the most defensible default — it matches the fact that the backend will reject the offline ones anyway, and it doesn't silently discard part of what the user asked for. But it's your call.

Note that **Wake** is a deliberate exception in both surfaces — it exists precisely to target offline devices and must not be gated. PR #2457 added an explicit test pinning that, so a future "gate everything" sweep can't quietly disable it.

Refs #2426, #2078, #2315.

ToddHebebrand:
Draft PR: #2476 (implements **option 2**, warn-then-proceed-on-online-subset, as specified).

**But I hit a blocker while verifying, and it changes the question this issue is asking.** Details in the [PR comment](https://github.com/LanternOps/breeze/pull/2476#issuecomment-4961774019); short version:

**The premise that "the API rejects the offline ones anyway" does not hold for `reboot` or `run-script`.**

[...]

PR #5034, opened 2026-09-06
feat(web): Remove asks about the agent — uninstall by default (#3987 items 2, 6)

### LINKED ISSUE #2250: [UI] Decommission flow should offer a clear "also uninstall agent" option
## Description

The reporter in #2230 expected "Agent deleted on decommissioned devices" — but decommission is deliberately a reversible soft-delete and never uninstalls the agent. Remote uninstall only exists on `DELETE /devices/:id/permanent`, which (a) requires the device to already be decommissioned and (b) after PR #2249 can essentially never reach the agent, because decommission now severs the live WS channel immediately and the auth gates reject every reconnect. There is currently **no reliable way to remotely uninstall an agent as part of offboarding.**

The decommission confirmation copy also overstates what happens ("This will permanently remove the device from your fleet") — it's restorable via `POST /devices/:id/restore`.

## Proposed Fix

Add an explicit opt-in to the decommission flow (single-device dialog and bulk action):

- Checkbox in the confirm dialog: **"Also uninstall the agent from the device"** (default off), with a note that this makes restore require a reinstall.
- API: `DELETE /devices/:id` accepts an `uninstallAgent` flag. Ordering matters — send `SELF_UNINSTALL` (with `removeConfig: true`) to the connected agent **before** flipping status and force-closing the WS, since after decommission the agent is unreachable by design.
- Offline agents can't receive the command; surface that per device in the result (same best-effort caveat the permanent-delete path already documents), especially in the bulk summary toast.
- Audit trail should record whether uninstall was requested and whether the command was delivered (alongside the existing `agentWsDisconnect` tri-state from PR #2249).
- While in there: fix the decommission confirm copy in `DeviceActions.tsx` — decommission is reversible, not "permanent".

## Affected Files

- `apps/web/src/components/devices/DeviceActions.tsx` (confirm dialog + copy)
- `apps/web/src/components/devices/DevicesPage.tsx` (bulk decommission path)
- `apps/web/src/services/deviceActions.ts` (`decommissionDevice` / `bulkDecommissionDevices`)
- `apps/api/src/routes/devices/core.ts` (`DELETE /devices/:id` — uninstall-before-decommission ordering)

## Context

Follow-up to #2230 / PR #2249 (which fixed decommissioned devices resurrecting via live agent WebSockets). Requested by the project owner after reviewing that fix.

Refs #2230

### LINKED ISSUE #5024: W01: Remove dialog with agent choice, single + bulk, all five surfaces (#3987 items 2, 6)
Wave W01 of #5023 — Remove dialog with agent choice, single + bulk, all five surfaces (#3987 items 2, 6)
Branch: `feature/5023-device-removal-completion/wave-5024`

ToddHebebrand:
## The zombie gap

`DELETE /devices/:id` has accepted `{ uninstallAgent }` and queued a durable, provenance-tagged `self_uninstall` since #4001 — but **no web caller ever sent it** (`grep uninstallAgent apps/web/src` → 0 hits before this PR). Every Remove issued a bodyless DELETE, so the API's back-compat default `false` applied and the machine kept running an agent that heartbeats into a 403 forever. That is exactly the outcome the 2026-08-24 owner decision was meant to eliminate.

This wave makes every Remove ask the one question Remove actually has — *uninstall the agent, or leave it?* — default it to **Uninstall**, and send the answer.

## The six surfaces

All six now render the same `RemoveDeviceDialog`:

| Surface | Path |
|---|---|
| DevicesPage row kebab | `pendingDeviceAction` → `runDeviceAction(action, device, choice)` |
| DevicesPage grid card | same handler (`handleDeviceAction` is shared) |
| DevicesPage bulk bar | new `pendingBulkRemove` state → `runBulkRemove(devices, choice)` — asked **once** for the whole selection |
| Device detail page kebab | `DeviceActions` → `handleConfirm(opts)` → `DeviceDetailPage.handleAction(action, device, opts)` |
| `PossibleReplacementBanner` | was the one surface issuing its own bodyless DELETE; now goes through the dialog and sends the body |
| Device detail page → Settings → Danger Zone | `DeviceSettingsModal` fires `onAction('decommission', device)` with **no** choice; `DeviceDetailPage.handleAction` now gates any opts-less `decommission` through the dialog (found in review — see below) |

Service layer: `decommissionDevice(id, { uninstallAgent })` and `bulkDecommissionDevices(devices, { uninstallAgent })` — the option is a **required** argument, so a new call site cannot silently omit it (that requirement is what surfaced the two stale call sites during implementation).

### Review fix — a sixth surface

Review caught a Remove path this PR had missed. `DeviceSettingsModal`'s Danger Zone button calls `onClose(); onAction('decommission', device)` with no choice argument. On `DevicesPage` that is harmless — `handleDeviceAction` gates every `decommission` through the dialog before it reaches `runDeviceAction`. On `DeviceDetailPage` there was no such gate: `case "decommission"` went straight into the undo toast and removed the device with `uninstallAgent: true` silently, giving the operator no way to choose "leave installed".

`DeviceDetailPage.handleAction` now has the same gate/execute split `DevicesPage` has, keyed on the **absence of `opts`** rather than on the caller:

```ts
if (action === "decommission" && !opts) {
  setPendingRemove(device);
  return;
}
```

Keying on `opts` rather than on which component called means any present or future Remove trigger on this page is gated by default, and a caller that already asked (the `DeviceActions` kebab) is not asked twice. `DeviceSettingsModal.tsx` is deliberately **not** touched — W03 is editing it on another branch.

## Copy decisions

- **Defaults to Uninstall.** Never "leave installed" (owner decision 2026-08-24).
- **"Queued", never "runs now."** Remove inserts a pending `self_uninstall` and force-closes the agent WS; the agent collects it on its next poll. Online → *"Queued now — an online agent collects it within moments."* Anything else → *"Queued — runs the next time the device checks in. Cancelled if it hasn't after N hours."*
- **The window is fetched, not hardcoded.** `DEVICE_UNINSTALL_DRAIN_WINDOW_HOURS` is env-driven per deployment, so this PR adds a tiny read-only route `GET /devices/removal-config` → `{ uninstallDrainWindowHours }` (requires `devices:read`, mounted **before** `coreRoutes` so `/:id` cannot eat the static path — pinned by a new assertion in `index.test.ts`). If that fetch fails, the dialog falls back to the window-less sentence rather than inventing a number.
- **Bulk buckets are "online" / "not currently online"**, never "offline" — `maintenance`, `quarantined`, `updating`, `pending` are none of the three, and calling them offline is a claim an operator would act on.
- New `deviceActions.removeDialog.*` block added to **all eight locales** in the same commit.

The 5-second undo toast on single Remove is untouched: the dialog decides *how*, the toast still offers *whether*.

## Guard

`apps/web/src/components/devices/__tests__/removeSendsUninstallChoice.test.ts` statically scans every component for `fetchWithAuth(\`/devices/${...}\`, { method: 'DELETE' })` with no `body:`. Verified to discriminate — reverting the banner's body makes it fail and names the file.

## Tests (all verified locally, on the merge with `origin/main`)

| Command | Result |
|---|---|
| `apps/web` — `src/components/devices` | **76 files / 834 tests passed** |
| `apps/web` — `src/components/devices src/services/deviceActions.test.ts src/lib/i18n` | **84 files / 985 tests passed** (before the review fix; the devices row above supersedes it) |
| `apps/api` — `src/routes/devices` | **47 files / 678 tests passed** |
| `tsc --noEmit -p apps/web/tsconfig.json` | clean |
| `tsc --noEmit -p apps/api/tsconfig.json` | clean (needs `NODE_OPTIONS=--max-old-space-size=8192`, same as the CI job) |
| `pnpm lint` | clean — 6/6 turbo tasks |

New tests: 1 API route test + 1 mount-order assertion, 3 service tests, 6 `RemoveDeviceDialog` tests, 3 `DevicesPage` tests, 1 `DeviceActions` test, 1 banner test, 1 static guard, **4 `DeviceDetailPage.removeDialog` tests** (review fix).

## Deviations from the plan

1. **`index.test.ts` mount assertion is a static source scan, not a request.** The plan asked for a `GET /devices/removal-config` request with "the file's standard auth mock" — that file has no auth mock and never imports the router; every existing case reads `index.ts` as text and compares mount offsets. Followed the file's own convention instead, and proved the new assertion fails when the mount is moved after `coreRoutes`.
2. **`DeviceActions` testids differed from the plan's guesses.** Real ones are `device-actions-menu` (trigger) and `device-action-remove` (menu item), not `device-actions-menu-trigger` / `device-action-decommission`. Used the real ones; added no new trigger testid.
3. **A second, unlisted test file needed updating**: `apps/web/src/services/__tests__/deviceActions.test.ts` (distinct from `src/services/deviceActions.test.ts`) held four `decommissionDevice`/`bulkDecommissionDevices` call sites. Updated in the same commit as Task 2.
4. **`PossibleReplacementBanner.test.tsx` mocks `../../services/deviceActions`** rather than adding a second `fetchWithAuth` response for the config fetch. That file pins exact `fetchWithAuth` call counts to prove which requests the banner itself makes; letting the dialog's config fetch land in those counts would have broken three unrelated assertions. (The plan explicitly allowed this branch.)
5. **`handleConfirm` forwards `opts` only when present** (`opts ? onAction?.(t, d, opts) : onAction?.(t, d)`). Passing a bare `undefined` third argument to every confirm action would change its call arity and break the existing `toHaveBeenCalledWith('maintenance', device)` assertions for no benefit.
6. **Six pre-existing tests were updated**, all because the Remove copy moved from `deviceActions.confirm.decommission.*` to the dialog's own block, or because the bulk path now has an extra dialog to click through: two bulk-decommission tests in the #2465 describe, three in the #4009 describe, one in the all-offline describe, plus one in `DeviceActions.test.tsx` and one in the banner. One test title was corrected from "no gate, no confirm" to "no skip gate" — there *is* now a confirm; what it pins is the absence of the #2465 skip gate.
7. **`getModalConfig` lost its `decommission` case** and its parameter type now excludes it, so a future Remove cannot accidentally re-acquire generic yes/no copy.
8. **Review-fix naming:** the review note suggested calling the executor `runAction(...)` inside `DeviceDetailPage`. That name is already taken in this file by the imported `runAction` helper from `@/lib/runAction`, so the dialog's `onConfirm` re-enters `handleAction('decommission', target, choice)` instead — the same gate/execute round-trip, no shadowed import.
9. **`DeviceSettingsModal` is stubbed, not driven, in the new test** (W03 owns the real file on another branch). The stub reproduces its exact call shape; the assertion keys on the missing `opts`, so it holds for any caller that omits one. Worth knowing: if W03 changes that call shape, this test would keep passing while the real button regressed.

Closes #5024
Refs #3987, closes #2250

### DIFF
--- a/apps/api/src/routes/devices/index.test.ts
+++ b/apps/api/src/routes/devices/index.test.ts
@@ -11,6 +11,18 @@ describe('device router mount order', () => {
     expect(coreMount).toBeGreaterThan(rollbackMount);
   });
 
+  it('mounts the Remove-dialog config route before core parameter routes (#3987)', () => {
+    // `/removal-config` is a STATIC path under /devices. Mounted after
+    // coreRoutes, core's `GET /:id` matcher would claim it and the Remove
+    // dialog would fetch a 404 (or a 400 uuid error) instead of the window.
+    const source = readFileSync(fileURLToPath(new URL('./index.ts', import.meta.url)), 'utf8');
+    expect(source).toContain("import { removalConfigRoutes } from './removalConfig'");
+    const removalMount = source.indexOf("deviceRoutes.route('/', removalConfigRoutes)");
+    const coreMount = source.indexOf("deviceRoutes.route('/', coreRoutes)");
+    expect(removalMount).toBeGreaterThan(-1);
+    expect(coreMount).toBeGreaterThan(removalMount);
+  });
+
   it('mounts the billing sub-resource after core parameter routes (#3205 W06)', () => {
     const source = readFileSync(fileURLToPath(new URL('./index.ts', import.meta.url)), 'utf8');
     expect(source).toContain("import { billingRoutes } from './billing'");
--- a/apps/api/src/routes/devices/index.ts
+++ b/apps/api/src/routes/devices/index.ts
@@ -32,6 +32,7 @@ import { statsRoutes } from './stats';
 import { postureRoutes } from './posture';
 import { optionsRoutes } from './options';
 import { healthRoutes } from './health';
+import { removalConfigRoutes } from './removalConfig';
 import { agentRollbackRoutes } from '../agentRollback';
 
 export const deviceRoutes = new Hono();
@@ -87,6 +88,10 @@ deviceRoutes.route('/', postureRoutes);
 
 // Mount the high-power literal sub-resource before core's /:id routes.
 deviceRoutes.route('/', agentRollbackRoutes);
+
+// Mount the Remove-dialog config BEFORE core — `/removal-config` is a static
+// path that must not be eaten by the `/:id` matcher in coreRoutes.
+deviceRoutes.route('/', removalConfigRoutes);
 
 // Mount core routes (/, /:id, PATCH /:id, DELETE /:id)
 deviceRoutes.route('/', coreRoutes);
--- a/apps/api/src/routes/devices/removalConfig.test.ts
+++ b/apps/api/src/routes/devices/removalConfig.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, it, vi } from 'vitest';
+import { Hono } from 'hono';
+
+vi.mock('../../middleware/auth', () => ({
+  authMiddleware: vi.fn((c: any, next: any) => {
+    c.set('auth', { user: { id: 'u1' }, scope: 'organization', orgId: 'org-1', accessibleOrgIds: ['org-1'] });
+    return next();
+  }),
+  requireScope: vi.fn(() => async (_c: any, next: any) => next()),
+  requirePermission: vi.fn(() => async (_c: any, next: any) => next()),
+}));
+vi.mock('../../services/deviceUninstallDrain', () => ({
+  DEVICE_UNINSTALL_DRAIN_WINDOW_HOURS: 72,
+}));
+
+import { removalConfigRoutes } from './removalConfig';
+
+describe('GET /devices/removal-config', () => {
+  it('returns the configured uninstall drain window in hours', async () => {
+    const app = new Hono().route('/devices', removalConfigRoutes);
+    const res = await app.request('/devices/removal-config', { headers: { Authorization: 'Bearer t' } });
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual({ uninstallDrainWindowHours: 72 });
+  });
+});
--- a/apps/api/src/routes/devices/removalConfig.ts
+++ b/apps/api/src/routes/devices/removalConfig.ts
@@ -1,0 +1,25 @@
+import { Hono } from 'hono';
+import { authMiddleware, requireScope, requirePermission } from '../../middleware/auth';
+import { PERMISSIONS } from '../../services/permissions';
+import { DEVICE_UNINSTALL_DRAIN_WINDOW_HOURS } from '../../services/deviceUninstallDrain';
+
+export const removalConfigRoutes = new Hono();
+removalConfigRoutes.use('*', authMiddleware);
+
+/**
+ * GET /devices/removal-config — read-only knobs the Remove dialog needs.
+ *
+ * `uninstallDrainWindowHours` is how long a queued self_uninstall waits for a
+ * removed device to check in before the stale-command reaper cancels it
+ * (`DEVICE_UNINSTALL_DRAIN_WINDOW_HOURS`, env-driven, floored at 1). The web
+ * must not hardcode it: operators tune it per deployment.
+ *
+ * Static path — MUST be mounted before coreRoutes in devices/index.ts or the
+ * `/:id` matcher eats it as a device id.
+ */
+removalConfigRoutes.get(
+  '/removal-config',
+  requireScope('organization', 'partner', 'system'),
+  requirePermission(PERMISSIONS.DEVICES_READ.resource, PERMISSIONS.DEVICES_READ.action),
+  (c) => c.json({ uninstallDrainWindowHours: DEVICE_UNINSTALL_DRAIN_WINDOW_HOURS }),
+);
--- a/apps/web/src/components/devices/DeviceActions.test.tsx
+++ b/apps/web/src/components/devices/DeviceActions.test.tsx
@@ -289,7 +289,31 @@ describe('DeviceActions — offline gating (issue #2013)', () => {
 
       await user.click(screen.getByTestId('device-action-remove'));
       expect(onAction).not.toHaveBeenCalled();
-      expect(await screen.findByText('Remove Device')).toBeInTheDocument();
+      // #3987: the generic "Remove Device" confirm was replaced by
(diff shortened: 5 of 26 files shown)


ToddHebebrand:
### Review fix pushed — a sixth Remove surface (`341be0587a`)

Confirmed the finding and fixed it in this PR. `DeviceSettingsModal`'s Danger Zone button fires `onAction('decommission', device)` with no choice argument. `DevicesPage` absorbs that safely — `handleDeviceAction` gates every `decommission` through `CONFIRM_REQUIRED_ACTIONS` before it reaches `runDeviceAction`. `DeviceDetailPage.handleAction` had no such gate, so **Settings → Decommission on the detail page removed the device with no dialog**, silently defaulting to `uninstallAgent: true` with no way to choose "leave installed".

`DeviceDetailPage.handleAction` now has the same gate/execute split, keyed on the **absence of `opts`** rather than on the caller:

```ts
if (action === "decommission" && !opts) {
  setPendingRemove(device);
  return;
}
```

Keying on `opts` rather than on which component called means every present and future Remove trigger on this page is gated by default, while the `DeviceActions` kebab — which already asked in its own dialog and passes the answer down — is not asked twice.

`DeviceSettingsModal.tsx` is **untouched** (W03 owns it on another branch); verified with an empty `git diff origin/main` on that path.

**New tests** — `DeviceDetailPage.removeDialog.test.tsx`, 4 cases:
1. Settings → Remove opens the radio dialog and calls nothing, even after the 5s undo window elapses
2. choosing **Leave the agent installed** forwards `{ uninstallAgent: false }`
3. cancelling removes nothing
4. a caller that already carries a choice (the kebab) executes directly and shows **no** second dialog

Three were red before the fix (`Unable to find [data-testid="remove-choice-uninstall"]`). Case 4 was written green as a regression guard, so I proved it discriminates: dropping the `&& !opts` condition fails it *and* case 2, because an unconditional gate re-asks forever.

**Sweep:** `grep` confirms only two page-level `onAction` handlers exist (`DevicesPage.handleDeviceAction`, `DeviceDetailPage.handleAction`) and only two component `decommissionDevice` call sites, both defaulting to uninstall. No seventh surface.

**Verification** (all green, on the merge with `origin/main`):

| Check | Result |
|---|---|
| `apps/web` — `src/components/devices` | **76 files / 834 tests passed** (was 74 / 828) |
| `tsc --noEmit -p apps/web/tsconfig.json` | clean |
| `tsc --noEmit -p apps/api/tsconfig.json` | clean |
| `pnpm lint` | clean, 6/6 turbo tasks |

One caveat worth recording: the new test **stubs** `DeviceSettingsModal` rather than driving the real component, since W03 is editing it. The stub reproduces its exact call shape, and the assertion keys on the missing `opts` so it holds for any caller that omits one — but if W03 changes that call shape, this test would keep passing while the real button regressed.

The PR body's surfaces table is updated to six.

PR #5036, opened 2026-09-06
fix(alerts): gate correlation-member move on the group, group-first lock order (#4867)

ToddHebebrand:
Follow-up to #5005 (merged as `f4387aa58d`), addressing every finding from its post-merge review. Branched fresh off `main`.

## Important 1 — split-brain member rewrite

`moveOrg.ts` re-stamped **every** `alert_correlation_members` row for the device's alerts, but the group statement holds a group back when it still spans two orgs or its `(org_id, group_key)` slot is occupied in the target. `correlationMetadataCondition` (`routes/alerts/alerts.ts`) joins a member to its group and pins **both** `org_id`s, so a member moved to the target under a group left in the source was visible to **neither** org — strictly worse than holding both, where the source org still renders the group intact and the target org merely sees an uncorrelated alert.

Members are now gated on the **group's** `org_id` as this transaction leaves it:

```sql
UPDATE alert_correlation_members m SET org_id = <target>
WHERE m.org_id IS DISTINCT FROM <target>
  AND EXISTS (SELECT 1 FROM alert_correlation_groups g
              WHERE g.id = m.group_id AND g.org_id = <target>
                AND EXISTS (… this device's alerts are in the group …))
```

Two deliberate details: within a moved group **every** member travels, including one whose own alert lives on a different device (that device's alerts are already in the target org — which is precisely what let the group past the guards); and matching on `g.org_id = <target>` rather than on the group statement's `RETURNING` ids also heals a member the pre-fix code stranded under a group that is already in the target org.

## Important 2 — lock order

`services/alertCorrelationGroups.ts` upserts the **GROUP** and then its **MEMBERS** on every correlation pass (`upsertGroup` → `upsertMembers`). `moveOrg` took members first — a textbook AB-BA that a concurrent correlation pass over the same group turns into a 40P01, surfacing as a 500 on an admin action.

The three statements now run **group → member → verdict**, matching the writer. `ALERT_CHILD_ORG_REWRITE_TABLES` (`routes/devices/core.ts`) is reordered to match, since `moveOrg.test.ts` pins the real statement sequence to that array. `services/ticketOrgMoveLockOrder.ts` was read for the precedent; the order and its rationale are now stated once, at the statements and on the array, the same way that module does for the ticket axis (#4657). This pair is not added to that module because its counterpart is the correlation job rather than a second org-mover.

Secondary, and *inferred* rather than executed: `orgMerge.ts` walks `topologicalCascadeOrder()` **reversed** (parents-first), and `alert_correlation_members.group_id → alert_correlation_groups.id` is a real FK in the DB (verified via `pg_constraint` on the test database), so the merge walker takes groups before members too. The new order aligns with it; the old one did not.

## Critical — no integration test

`moveOrg.test.ts` mocks `../../db` wholesale, so these three raw UPDATEs were only ever compared as **text** — nothing in CI had executed them. New `apps/api/src/__tests__/integration/deviceMoveOrgAlertChildren.integration.test.ts` (shape copied from `deviceMoveOrgCurrency.integration.test.ts`) drives the real route against real Postgres under real RLS:

1. a group whose member alerts **all** reach the target org travels with the device — group, every member row, the alert-level verdict and the group-level (`alert_id NULL`) verdict are all re-stamped;
2. a group spanning two devices in different orgs is held back **and its member rows are held back with it** — this case failed before the fix (see below);
3. both audit rows carry the split hold-back counts.

It lives under `src/__tests__/integration/**`, so `vitest.integration.config.ts`'s shared glob already picks it up (blocking `integration-test` job); no config change needed, and `integration-suite-coverage` passes.

**Red before the fix** (same file, same stack, on the pre-fix code):

```
 FAIL  deviceMoveOrgAlertChildren.integration.test.ts > holds a group that still spans two orgs
       back — and holds ITS member rows back with it
 AssertionError: expected '0a54848e-…' to be 'd0f8375a-…'   ← memberA4 had moved to org B
   310|     expect(after.memberA4).toBe(f.orgA.id);

 FAIL  … > records the moved and held-back counts, split by hold-back reason, on BOTH audit rows
 -     "correlationGroupsHeldKeyCollision": 0,
 -     "correlationGroupsHeldSpanning": 1,
 -     "correlationMembers": 2,
 -     "correlationMembersHeld": 1,
 +     "correlationMembers": 3,

 Test Files  1 failed (1)
      Tests  2 failed | 1 passed (3)
```

**Green after** (private worktree test stack, `pnpm test-stack up`):

```
$ npx vitest run --config vitest.integration.config.ts \
    src/__tests__/integration/deviceMoveOrgAlertChildren.integration.test.ts \
    src/__tests__/integration/deviceMoveOrgCurrency.integration.test.ts

 Test Files  2 passed (2)
      Tests  6 passed (6)
   Duration  15.30s
```

## Suggestions

- **The wrong comment.** "member → group → verdict … because each later predicate reads what the earlier one just wrote" was false: the group's span guard reads `alerts.org_id` (re-stamped by the generic loop), never `members.org_id`. Corrected in place in both `moveOrg.ts` and `core.ts`, with an explicit note that the order it justified was the deadlock above.
- **Split hold-back counts.** `correlationGroupsHeld` → `correlationGroupsHeldSpanning` (self-resolves when the group's last device moves; nothing to do) + `correlationGroupsHeldKeyCollision` (never self-resolves — two groups need a human decision, so a non-zero value is the signal). `spans` is evaluated first and wins when both are true, so the two always sum to the number of held groups. New `correlationMembersHeld` counts this device's memberships that stayed behind. These live in the `alertChildRewrite` audit-details block on both audit rows, where the existing counts already were; the HTTP response body is unchanged.
- **Drizzle-FK limit.** `moveOrg.coverage.test.ts` derives its expected membership from `getTableConfig(table).foreignKeys`, which only sees FKs **declared in the Drizzle schema** — a migration-only FK (or an untyped uuid column) is invisible to the guard, and such a table would be missed exactly the way #4867 was. Documented on the describe block, with the instruction to hand-add the table and let the `stale` assertion prompt declaring the FK.

## Verification

| Check | Result |
|---|---|
| `deviceMoveOrgAlertChildren.integration.test.ts` + `deviceMoveOrgCurrency.integration.test.ts` (real PG) | 6/6 passed |
| `moveOrg.test.ts`, `moveOrg.coverage.test.ts`, `ticketOrgMoveLockOrder.test.ts`, `alertVerdicts.test.ts` | 133/133 passed |
| `integration-suite-coverage` | 5/5 passed |
| `tsc --noEmit -p apps/api` | clean (exit 0) |
| `eslint` on all 5 touched files | clean |

No migration, no schema change, no new table — so no cascade/export-policy registration applies.

Refs #4867

https://claude.ai/code/session_01KDahzaq8PRqkwR3eo2k9VB

### DIFF
--- a/apps/api/src/__tests__/integration/deviceMoveOrgAlertChildren.integration.test.ts
+++ b/apps/api/src/__tests__/integration/deviceMoveOrgAlertChildren.integration.test.ts
@@ -1,0 +1,352 @@
+/**
+ * Integration test: the ALERT-axis children of a device org-move (#4867, and
+ * the post-merge review of #5005).
+ *
+ * The three tables `alert_correlation_members`, `alert_correlation_groups` and
+ * `ai_alert_verdicts` denormalize `org_id` but have NO `device_id` column, so
+ * moveOrg.ts reaches them with three hand-written raw UPDATEs. `moveOrg.test.ts`
+ * mocks `../../db` wholesale, so those statements are only ever compared as
+ * TEXT there — nothing in CI executed them until this file. Every predicate
+ * they carry (the three group hold-back guards, the member gate, the verdict's
+ * two legs) is therefore asserted here against real Postgres, through the real
+ * route, under real RLS.
+ *
+ * Proves:
+ *   (1) a correlation group whose member alerts ALL reach the target org
+ *       travels with the device — the group, every one of its member rows, its
+ *       group-level verdict and the alert-level verdict on a moved alert are
+ *       all re-stamped to the target org;
+ *   (2) a group still SPANNING two orgs afterwards (one member alert sits on a
+ *       device that did not move) is held back — and, the #5005 review's
+ *       split-brain finding, its member rows are held back WITH it. Before the
+ *       fix the member row moved unconditionally while its group stayed, and
+ *       `correlationMetadataCondition` (routes/alerts/alerts.ts) pins BOTH
+ *       org_ids, so the row was then visible to neither org;
+ *   (3) both audit rows carry the split hold-back counts.
+ *
+ * Harness mirrors deviceMoveOrgCurrency.integration.test.ts: a partner-scope
+ * environment with wildcard permissions and an MFA-bearing access token.
+ */
+import './setup';
+import { describe, expect, it } from 'vitest';
+import { Hono } from 'hono';
+import { eq, inArray } from 'drizzle-orm';
+import {
+  aiAgentRuns,
+  aiAgents,
+  aiAlertVerdicts,
+  alertCorrelationGroups,
+  alertCorrelationMembers,
+  alerts,
+  auditLogs,
+  devices,
+} from '../../db/schema';
+import { createOrganization, createSite, setupTestEnvironment } from './db-utils';
+import { getTestDb } from './setup';
+import { createAccessToken } from '../../services/jwt';
+import { moveOrgRoutes } from '../../routes/devices/moveOrg';
+
+function uid(): string {
+  return `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+}
+
+/**
+ * Two orgs under one partner, two devices in the source org, and two
+ * correlation groups whose fates differ:
+ *
+ *   gTravelling — members are alertA1 + alertA2, BOTH on `deviceMoved`. After
+ *     the move every member alert is in orgB, so the group travels.
+ *   gSpanning   — members are alertA4 (on `deviceMoved`) and alertB1 (on
+ *     `deviceStays`). After the move alertB1 is still in orgA, so the group is
+ *     held back — and alertA4's member row must be held back with it.
+ *
+ * Verdicts cover both legs of the verdict statement: `verdictOnAlert` is
+ * alert-level on alertA1, `verdictOnTravellingGroup` is group-level on
+ * gTravelling (alert_id NULL — the `duplicate_of_group` shape the alert leg
+ * alone can never reach), `verdictOnSpanningGroup` is group-level on the group
+ * that stays.
+ */
+async function seed() {
+  const adminDb = getTestDb() as any;
+  const unique = uid();
+
+  const env = await setupTestEnvironment({ scope: 'partner' });
+  const { partner, organization: orgA, site: siteA, user, role } = env;
+
+  const orgB = await createOrganization({ partnerId: partner.id });
+  const siteB = await createSite({ orgId: orgB.id });
+
+  const insertDevice = async (label: string) => {
+    const [row] = await adminDb.insert(devices).values({
+      orgId: orgA.id,
+      siteId: siteA.id,
+      agentId: `ac-${label}-agent-${unique}`,
+      hostname: `ac-${label}-host-${unique}`,
+      osType: 'linux',
+      osVersion: '22.04',
+      architecture: 'x86_64',
+      agentVersion: '0.0.0-test',
+      status: 'offline',
+    }).returning();
+    return row;
+  };
+  const deviceMoved = await insertDevice('moved');
+  const deviceStays = await insertDevice('stays');
+
+  const insertAlert = async (deviceId: string, title: string) => {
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Findings → changes** (post-merge review of #5005). Every finding addressed; nothing deferred.

| # | Finding | Change | Where |
|---|---|---|---|
| **Critical** | No integration test — `moveOrg.test.ts` mocks `../../db` wholesale, so the three raw UPDATEs never execute in CI | New real-Postgres suite through the real route under real RLS: full-move case, held-back-group case, audit counts. Case 2 was **red** on the pre-fix code (`memberA4` had moved to org B; `correlationMembers` read 3 instead of 2) — red output pasted in the PR body | `apps/api/src/__tests__/integration/deviceMoveOrgAlertChildren.integration.test.ts` (new) |
| **Important 1** | Split-brain — members moved unconditionally while the group could be held back, leaving the row visible to neither org | Member rewrite gated on the **group's** `org_id` after the group statement. Within a moved group every member travels (including members whose alert is on another device); matching on `g.org_id = <target>` instead of the group statement's `RETURNING` ids also heals members the pre-fix code already stranded | `routes/devices/moveOrg.ts` — `movedCorrelationMembers` |
| **Important 2** | Lock order — `alertCorrelationGroups.ts` writes group→members, moveOrg took members→group (ABBA, 40P01) | Reordered to **group → member → verdict**, matching the writer. `ALERT_CHILD_ORG_REWRITE_TABLES` reordered to match (it is what `moveOrg.test.ts` pins the statement sequence to). Read `services/ticketOrgMoveLockOrder.ts` for the precedent and documented the order + both reasons at the statements and on the array | `routes/devices/moveOrg.ts`, `routes/devices/core.ts` |
| **Suggestion** | The "member → group → verdict order is load-bearing" comment is wrong | Corrected in both places, stating explicitly that the group's span guard reads `alerts.org_id` and nothing reads `members.org_id` — and that the order that comment justified was the deadlock in Important 2 | `routes/devices/moveOrg.ts`, `routes/devices/core.ts`, plus the mirrored comments in `moveOrg.test.ts` / `moveOrg.coverage.test.ts` |
| **Suggestion** | Split `correlationGroupsHeld` into span vs `group_key`-collision | `correlationGroupsHeldSpanning` + `correlationGroupsHeldKeyCollision` (spanning wins when both apply, so the two always sum to the held total), plus a new `correlationMembersHeld` for this device's memberships left behind | `routes/devices/moveOrg.ts` — `AlertChildOrgRewriteCounts` and the two held-count SELECTs |
| **Suggestion** | Document that the coverage guard only sees Drizzle-declared FKs | Noted on the `ALERT_CHILD_ORG_REWRITE_TABLES coverage` describe block: a migration-only FK (or an untyped uuid column) is invisible to the derivation, so such a table must be hand-added — and the `stale` assertion then prompts declaring the FK in the schema | `routes/devices/moveOrg.coverage.test.ts` |

Two notes on scope:

- The counts live in the `alertChildRewrite` **audit-details** block on both audit rows, where the existing counts already were. The HTTP response body is unchanged — widening it is a public-surface change the review did not ask for.
- Secondary and **inferred**, not executed: `orgMerge.ts` walks `topologicalCascadeOrder()` reversed (parents-first) and `alert_correlation_members.group_id → alert_correlation_groups.id` is a real FK (checked against `pg_constraint` on the test DB), so the merge walker also takes groups before members. The new order agrees with it; the old one did not. I did not add the pair to `ticketOrgMoveLockOrder.ts` — its counterpart is the correlation job, not a second org-mover.

Verification: 6/6 integration (incl. `deviceMoveOrgCurrency` as regression), 133/133 unit across `moveOrg` / `moveOrg.coverage` / `ticketOrgMoveLockOrder` / `alertVerdicts`, 5/5 `integration-suite-coverage`, `tsc --noEmit -p apps/api` clean, `eslint` clean on all five touched files. No migration and no schema change, so no cascade/export-policy registration applies.

Not merged — awaiting maintainer review.

Issue #3182, opened 2026-08-06
Harden device_group_memberships tenancy: add composite FK to device_groups, and org-scope the two system-side readers


ToddHebebrand:
## Summary

`device_group_memberships` rows are tenant-scoped by their own `org_id` column alone. Nothing structurally ties the row's `group_id` to that same org, so at the SQL level a row can name a group belonging to a *different* org while stamping its own `org_id` — and RLS accepts it, because the row's `org_id` is one the writer legitimately has access to.

This is **defense-in-depth hardening, not a live exploitable hole** — see "Why this is not currently reachable" below. Filing it so the structural gap gets closed rather than relying on every current and future write path to be careful.

Found while fixing #3181 (dynamic group membership materialization); the integration test added there already asserts both the RLS rejection of an owner-org-stamped forgery and the quarantine behaviour.

## The gap

The RLS policy answers *"does the caller have access to this row's `org_id`?"* It never answers *"does this row's `group_id` belong to that org?"* There is no composite FK `(group_id, org_id) → device_groups(id, org_id)`.

So this row is accepted by RLS today:

```
{ group_id: <org A's group>, device_id: <org B's device>, org_id: <org B> }
```

## Why this is not currently reachable through the API

The only route that writes memberships — `apps/api/src/routes/devices/groups.ts:353-397` — closes it at the app layer three times over:

1. `getGroupById(groupId)` runs under the caller's RLS context, so another org's group simply isn't visible → 404 before anything else runs.
2. `ensureOrgAccess(group.orgId, auth)` re-checks explicitly.
3. The insert stamps `orgId: group.orgId` — **the group's org, never the caller's**. No client-supplied `org_id` can reach the column.

Exercising the gap therefore requires direct `breeze_app` database credentials, not an API call. The risk is a *future* write path (or an AI/MCP tool, or a bulk-import job) that inserts memberships without replicating all three checks.

## Why it is still worth closing: two readers ignore `org_id`

If such a row ever did exist, it would **not** stay quarantined. Two system-side readers resolve group membership by `group_id` with no org predicate:

- `apps/api/src/jobs/backupSlaWorker.ts:87-90` — `.where(inArray(deviceGroupMemberships.groupId, groupIds))`, and this worker runs under `withSystemDbAccessContext`, which bypasses RLS entirely.
[...]

PR #5039, opened 2026-09-06
fix(devices): re-evaluate dynamic groups off the request path (#4630 review follow-up)

### LINKED ISSUE #4630: Dynamic device group membership never re-evaluates on device change (event pipeline is unwired)

ToddHebebrand:
## Summary

Follow-up to **#5015**, which was squash-merged (`1a76507d3a`) before its review fixes landed. This PR addresses all five findings from the pr-review-toolkit (Opus) round on that PR, on top of current `main`.

The headline change: **dynamic device-group re-evaluation no longer runs on the request path.** It now goes through a coalescing BullMQ queue drained by a worker under `withSystemDbAccessContext`.

### 1. (critical) Re-evaluation ran inline inside the heartbeat's open PG transaction

`withDbAccessContext` is a real `baseDb.transaction(...)`, so `#5015`'s `await emitDeviceChange(...)` in `routes/agents/heartbeat.ts` ran *after* `UPDATE devices` took its row lock, and each membership flip made 2–3 Redis round trips (`schedulePeripheralPolicyDevice`) while holding a pooled Postgres connection. No debounce, no per-device coalescing — the pool-starvation shape from the 09-03 US incident, on the highest-volume endpoint in the product.

New `apps/api/src/jobs/deviceGroupJobs.ts`:

- Queue `device-group-reevaluation`, coalesced per device under `jobId = group-reeval-<deviceId>` — the same shape `jobs/peripheralJobs.ts` uses for `policy-reconciliation-<deviceId>`, including the *active-job follow-up* branch and the *spent-record replace* branch (a retained `failed` record under a fixed jobId silently swallows every later `add`; see `services/bullmqUtils.ts`).
- Coalescing **merges** rather than overwrites: changed fields are unioned, and `device.created` subsumes `device.updated` (created evaluates *every* dynamic group, so collapsing the pair the other way would drop groups).
- `requestDeviceGroupReevaluation` is **fire-and-forget** (`void`-ed at all three call sites) and never rejects, so not even the enqueue's own Redis round trips happen while the request transaction is open, and a Redis outage can't fail a heartbeat. It touches no Postgres connection, so unlike a detached DB call it can't be stranded on a committed transaction (the #3182 materialization bug).
- The worker wraps the evaluation in `withSystemDbAccessContext` and **re-reads the device's own `org_id` from the database** instead of trusting the job payload — the job outlives the request that produced it, and a membership row stamped from a stale payload org would be a cross-tenant row.
- Registered in `workerRegistry` as `deviceGroupJobs` with placement `socket-owner` (its closure reaches `jobs/peripheralJobs.ts` via `services/groupMembership.ts`) — confirmed by `workerEntrypointClosure.contract.test.ts`, not guessed.
- `initializeDeviceEventHandlers()` is now **idempotent** and called from the job processor. This is load-bearing: `worker.ts` never runs `index.ts`'s bootstrap, so a `worker`-role process would otherwise drain the queue with an empty handler registry and every job would be a silent no-op.

`routes/agents/heartbeat.ts`, `routes/agents/enrollment.ts` and `routes/devices/provision.ts` now contain no DB or Redis work for this feature at all — just the enqueue.

### 2. (critical) Unordered multi-group lock acquisition

`evaluateDeviceMembershipForGroup` → `ensureFilterFieldsUsed` UPDATEs `device_groups` when `filter_fields_used` is stale, so two concurrent evaluations that walk an org's groups in different orders can take the group row locks in opposite orders and deadlock (40P01, the #3911 shape). Added `ORDER BY id` to:

- the dynamic-group SELECT in `updateDeviceMembership` (`services/groupMembership.ts`) — the site the finding named;
- the `device.created` handler's group SELECT in `events/deviceEvents.ts`, which has the identical hazard.

### 3. (important) No integration test

New `apps/api/src/__tests__/integration/dynamicGroupReevaluation.integration.test.ts` drives the real queue processor against real Postgres, real forced RLS as `breeze_app`, and the real filter engine — mutating device rows exactly the way a heartbeat does. Six cases:

1. **add** — a hostname change that starts matching `hostname startsWith 'srv-'` materializes the membership row;
2. **remove** — the same device renamed out of the filter has its row deleted;
3. `device.created` picks up a device that already matched at insert (the enrollment/provision path);
4. the processor evaluates against the **device's** org, not the payload's;
5. a matching device in another partner's org is never absorbed;
6. a device deleted between enqueue and run is a clean no-op.

Every unit test on this path mocks either `groupMembership` or `deviceEvents`, so the two ways this feature silently does nothing — a contextless SELECT returning zero groups under forced RLS, and whether the compiled filter SQL actually matches the row after an UPDATE — are invisible to them.

**Proved discriminating:** commenting out the handler-registry init in the processor reds 4 of the 6 (see the run below).

### 4. (important) `Promise.allSettled` results never inspected

`emitDeviceChange` now logs every rejected handler at error level with the event type, device id and org id. A permanently-throwing evaluation used to leave a dynamic group stale forever with nothing in the logs — the #4630 failure mode all over again. The emit still resolves, so the BullMQ worker treats the job as done rather than retrying a deterministic failure forever.

### 5. (suggestion) Inert `routes/devices/groups.ts` wiring removed

That route's `createGroupSchema`/`updateGroupSchema` never accepted `filterConditions`, so `group.filterConditions` is always null and the `evaluateGroupMembership` branch could never fire. Removed it and the two mock-only tests that only proved the mock was called. Whether that duplicate route should gain filter support or be deleted in favour of `/groups` is the open question #4630 itself raises, and is not decided here.

## `Refs #4630`, not `Closes` — still-unwired call sites

#5015 claimed `Closes #4630`. It does not close it. Wired today: **agent heartbeat** (`hostname`, `osVersion`, `osBuild`, `deviceRole`), **agent enrollment**, **device provisioning**. Still unwired:

| Path | File | Why it matters |
|---|---|---|
| `PATCH /devices/:id` | `routes/devices/core.ts` | a tech renaming a device, or editing `displayName`/`tags`/`customFields`, does not re-evaluate |
| Site move (`siteId` on that PATCH) | `routes/devices/core.ts:1557` | `updateDeviceMembership`'s `changedFields.includes('siteId')` branch exists and has **no caller** |
| Org move | `routes/devices/moveOrg.ts` | memberships are *dropped* correctly, but the device is never evaluated against the destination org's dynamic groups |
| Tag changes | tag routes | `tags` is a filterable field with no emit |
| Hardware inventory ingest | `routes/agents/inventory.ts` | `device.hardware_updated` handler exists, no producer |
| Network inventory ingest | `routes/agents/inventory.ts` | `device.network_updated` handler exists, no producer |
| Software inventory ingest | `ingestSoftwareInventoryReport` | `device.software_changed` handler exists, no producer — needs a real "did the installed set change" signal first, or every routine sync re-evaluates every dynamic group in the org |
| Custom-field writes | `routes/devices/customFieldValues.ts` | `custom.*` filters go stale |
| Device delete | `routes/devices/core.ts` | `device.deleted` handler exists; hard delete already cascades memberships, soft-decommission deliberately left alone (evicting a decommissioned device that still matches would be a behaviour change) |
| Periodic reconciliation sweep | — | the issue's suggested backstop #2; no drift repair exists |

Each of those is a `requestDeviceGroupReevaluation(...)` call now that the queue exists, but they are behaviour changes with their own blast radius and belong in follow-up PRs.

## Verification

```
cd apps/api

NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit -p .        # exit 0

npx vitest run src/routes/agents src/routes/devices src/events \
  src/jobs/deviceGroupJobs.test.ts src/services/groupMembership \
  src/services/workerRegistry.test.ts \
  src/services/workerEntrypointClosure.contract.test.ts \
  src/routes/groups.test.ts
#  Test Files  104 passed (104)
#       Tests  1858 passed (1858)

npx eslint <all changed files>                                        # clean
```

Integration suite, real Postgres (`localhost:5433/breeze_test`):

```
BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=1 npx vitest run \
  --config vitest.integration.config.ts --reporter=verbose \
  src/__tests__/integration/dynamicGroupReevaluation

 ✓ dynamic device group re-evaluation on device change (#4630) > adds the device when a hostname change starts matching the filter 280ms
 ✓ dynamic device group re-evaluation on device change (#4630) > removes the device when a hostname change stops matching the filter 240ms
 ✓ dynamic device group re-evaluation on device change (#4630) > picks up a device that already matched at insert (device.created) 251ms
 ✓ dynamic device group re-evaluation on device change (#4630) > evaluates against the DEVICE's org, not the org named in the job payload 317ms
 ✓ dynamic device group re-evaluation on device change (#4630) > never absorbs a matching device from another tenant 246ms
 ✓ dynamic device group re-evaluation on device change (#4630) > no-ops for a device deleted between enqueue and run 204ms

 Test Files  1 passed (1)
      Tests  6 passed (6)
```

(`BREEZE_INTEGRATION_ALLOW_LEDGER_DRIFT=1` only because the shared local test DB carries one migration this checkout predates; this suite enumerates no schema, so the drift is inert here. CI runs it against a clean database with no flag.)

**Mutation control** — commenting out `initializeDeviceEventHandlers()` in `processDeviceGroupReevaluation`, i.e. simulating exactly the worker-role blind spot the idempotent init closes:

```
 × adds the device when a hostname change starts matching the filter
 × removes the device when a hostname change stops matching the filter
 × picks up a device that already matched at insert (device.created)
 × evaluates against the DEVICE's org, not the org named in the job payload

 Tests  4 failed | 2 passed (6)
```

## Tenancy

No new tables, no migration. `device_groups` and `device_group_memberships` are org-only and already registered in every cascade/export list. The one tenancy-relevant addition is the worker's org re-read (finding 1), covered by integration cases 4 and 5.

## Notes for reviewer

- The three call sites deliberately do **not** `await` the enqueue. Two tests pin that: a never-settling mock still returns 200 from the heartbeat and 201 from enrollment, so re-adding an `await` hangs those tests to their timeout rather than passing quietly.
- `WORKER_REGISTRY` grows 123 → 124; `workerRegistry.test.ts` and `workerEntrypointClosure.contract.test.ts` updated accordingly. The `socket-owner` placement is what the closure contract test computes, not a judgement call.

Refs #4630

https://claude.ai/code/session_01QpNjRgZYEmxZfr4axwGBQk

### DIFF
--- a/apps/api/src/__tests__/integration/dynamicGroupReevaluation.integration.test.ts
+++ b/apps/api/src/__tests__/integration/dynamicGroupReevaluation.integration.test.ts
@@ -30,16 +30,40 @@
  *   4. the processor uses the DEVICE's own org, not the job payload's, so a
  *      stale/forged payload cannot steer the evaluation at another tenant.
  *   5. a matching device in another partner's org is never absorbed.
+ *   6. a device DELETED between enqueue and run no-ops (and that this is
+ *      distinguishable from "not committed yet" — see case 7).
+ *   7. #5039 review finding 1, the enqueue-before-commit race: the enqueue
+ *      happens INSIDE the caller's still-open transaction, so an undelayed
+ *      worker reads a snapshot without the device row and silently completes
+ *      having evaluated nothing. Driven end to end against real Redis with a
+ *      real BullMQ worker, with the undelayed behaviour as a paired negative
+ *      control in the same test.
  */
 import './setup';
 
-import { describe, expect, it } from 'vitest';
+import { afterAll, afterEach, describe, expect, it } from 'vitest';
 import { randomUUID } from 'crypto';
 import { and, eq } from 'drizzle-orm';
+import type { Worker } from 'bullmq';
 
 import { db, withSystemDbAccessContext } from '../../db';
-import { deviceGroupMemberships, deviceGroups, devices } from '../../db/schema';
-import { processDeviceGroupReevaluation } from '../../jobs/deviceGroupJobs';
+import {
+  deviceGroupMemberships,
+  deviceGroups,
+  devices,
+  groupMembershipLog,
+} from '../../db/schema';
+import {
+  DEVICE_GROUP_REEVALUATION_DELAY_MS,
+  createDeviceGroupReevaluationWorker,
+  deviceGroupReevaluationJobId,
+  getDeviceGroupReevaluationQueue,
+  processDeviceGroupReevaluation,
+  runDeviceGroupReevaluationJob,
+  scheduleDeviceGroupReevaluation,
+  shutdownDeviceGroupJobs,
+} from '../../jobs/deviceGroupJobs';
+import { closeRedis } from '../../services/redis';
 import {
   createOrganization,
   createPartner,
@@ -257,7 +281,55 @@ describe('dynamic device group re-evaluation on device change (#4630)', () => {
     expect(anyRowForForeign).toHaveLength(0);
   });
 
+  /**
+   * The `!device` branch means DELETED, not "committing right now".
+   *
+   * That distinction is only true because of `DEVICE_GROUP_REEVALUATION_DELAY_MS`
+   * (#5039 review finding 1). Before the delay, the same branch also absorbed a
+   * device whose INSERT had not committed yet — the enrolment path — and
+   * reported `{evaluated: false}` as a SUCCESS, so nothing retried and the
+   * device was never added to any dynamic group. Case 7 below proves the
+   * not-yet-committed device now IS evaluated, which is what makes it safe for
+   * this branch to treat a missing row as a genuine deletion.
+   */
   runDb('no-ops for a device deleted between enqueue and run', async () => {
+    const env = await setupTestEnvironment();
+    const groupId = await seedDynamicGroup(env.organization.id);
+    const deviceId = await seedDevice(env.organization.id, env.site.id, 'srv-epsilon');
+
+    // It matched and was a member before the delete.
+    await runReevaluation({
+      deviceId,
+      orgId: env.organization.id,
+      eventType: 'device.updated',
+      changedFields: ['hostname'],
+    });
+    expect(await memberDeviceIds(groupId)).toEqual([deviceId]);
+
+    // Delete it in the same order the device cascade does: children first.
+    // Neither device_group_memberships.device_id nor group_membership_log
+    // .device_id is ON DELETE CASCADE, so both have to go before the device
+    // (this is exactly why CORE_DEVICE_CASCADE_DELETE_TABLES exists).
+    await withSystemDbAccessContext(async () => {
+      await db.delete(groupMembershipLog).where(eq(groupMembershipLog.deviceId, deviceId));
+      await db.delete(deviceGroupMemberships).where(eq(deviceGroupMemberships.deviceId, deviceId));
+      await db.delete(devices).where(eq(devices.id, deviceId));
+    });
+
+    const result = await runReevaluation({
+      deviceId,
+      orgId: env.organization.id,
+      eventType: 'device.updated',
+      changedFields: ['hostname'],
+    });
+
+    expect(result).toEqual({ evaluated: false, orgId: null });
+    expect(await memberDeviceIds(groupId)).toEqual([]);
+  });
+
+  // A device id that never existed takes the same branch — there is nothing to
+  // evaluate either way.
+  runDb('no-ops for a device id that never existed', async () => {
     const env = await setupTestEnvironment();
     await seedDynamicGroup(env.organization.id);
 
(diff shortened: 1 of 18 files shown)


ToddHebebrand:
**Review run:** pr-review-toolkit (Opus) — **FIX-THEN-MERGE**. All six #5015 findings verified closed (incl. the integration suite actually running in shard 2/4). One new critical from the move off the request path:

**Critical — enqueue-before-commit race.** `void requestDeviceGroupReevaluation(...)` fires while the caller's `withDbAccessContext` transaction is still open. The worker opens its own system transaction on another connection and cannot see uncommitted rows: on enrollment/provision the device row is not there yet → `if (!device) return { evaluated: false }` → job completes, no retry, device is **never** evaluated (the window is wide — enrollment awaits mTLS cert issuance after the enqueue). On heartbeat it reads the pre-UPDATE hostname and finds no flip. Integration case 6 ("no-ops for a device deleted between enqueue and run") encodes this silent branch as correct. **Fix:** `delay: 5000` in `jobOptions()` (`delayed` is already in `isReusableState`, so coalescing still works) — or enqueue post-commit.

**Important — `attempts: 5` is dead code.** `emitDeviceChange` swallows handler rejections (logs only), so the processor only throws on the device SELECT; a 40P01 in evaluation completes "successfully" and membership stays stale. Have the processor throw when any handler rejected.

**Suggestion —** `runWithSystemDbAccess` falls back to `fn()` when the export is missing; under forced RLS that is a silent zero-group no-op. Throw instead.

Held as draft until fixed.

ToddHebebrand:
**All three fixed** — `433478f1c`. Still a draft, not merged.

| Review finding | Change |
|---|---|
| **Critical — enqueue-before-commit race** | `jobOptions()` now sets `delay: DEVICE_GROUP_REEVALUATION_DELAY_MS` (5000), a new exported constant documented as a correctness constant rather than a throttle. Confirmed `delayed` **is** in `isReusableState` (`services/bullmqUtils.ts:8`), so coalescing still merges — and the delay window now doubles as the coalescing window for a flapping device. Asserted in both suites (see below). |
| **Important — `attempts: 5` is dead code** | `emitDeviceChange` still runs every handler to completion via `Promise.allSettled` and still logs each rejection, but now throws an `AggregateError` carrying all of them, so the processor propagates and BullMQ retries. The `device.created` handler's per-group `catch` did the same swallow one level down — which would have left the throw inert for the whole enrolment/provision path, exactly where concurrent enrolments make a 40P01 most likely — so it too collects failures, finishes evaluating the remaining groups, then rethrows an aggregate. |
| **Suggestion — `runWithSystemDbAccess` fallback** | Throws instead of calling `fn()` bare. Extracted `runDeviceGroupReevaluationJob(data)` — literally what the worker's processor runs — so the guard is reachable from a test; the worker now calls that. The test asserts it refuses *before* touching the database. |

### Proof for the race — integration case 7, real Postgres + real Redis + a real BullMQ worker

`dynamicGroupReevaluation.integration.test.ts` → *"evaluates a device whose INSERT commits only after the enqueue"*:

1. A worker is created and `waitUntilReady()`-ed **before** the enqueue, so it has every chance to grab the job the moment it becomes runnable.
2. One `withSystemDbAccessContext` transaction stands in for `/enroll`: it `INSERT`s the matching device and calls `scheduleDeviceGroupReevaluation`, then stays open (the real handler goes on to issue an mTLS cert).
3. **Negative control**, run from the test body — deliberately *outside* the enrolment's async-local DB context, so it opens its own transaction on its own pooled connection; a nested `withSystemDbAccessContext` would have reused the caller's transaction and seen the uncommitted row, which is precisely what the real worker cannot do. It returns `{evaluated: false, orgId: null}`. That is the bug, reproduced in the same test.
4. The queued job is asserted `delayed`, with `opts.delay === DEVICE_GROUP_REEVALUATION_DELAY_MS`, and the group still has no members.
5. The transaction commits → the worker completes the job → the membership row appears.

**Verified discriminating:** with the constant temporarily set to `0` the case fails with `expected 'completed' to be 'delayed'` — the ready worker takes the job while the transaction is still open. Restored to 5000 afterwards.

Case 8 (*"coalesces a second change into the still-delayed job"*) proves the coalescing half against the real queue: a second `scheduleDeviceGroupReevaluation` inside the window merges into the same jobId with the union of changed fields, and `getDelayedCount()` stays at 1.

### Integration case 6, reconsidered

You were right that it encoded the silent branch as correct. It no longer uses a random uuid. It now seeds a real device, proves it is a member, deletes it children-first the way `CORE_DEVICE_CASCADE_DELETE_TABLES` does (`group_membership_log` → `device_group_memberships` → `devices`), and asserts `{evaluated: false, orgId: null}` plus an empty membership list — so it proves the **deleted** path specifically. A separate one-liner keeps the never-existed id covered.

The `if (!device)` branch itself now carries the reasoning: the job cannot run until 5s after the enqueue, by which point the enqueuing transaction has committed or rolled back, so "no row" is unambiguously gone — and case 7 is what makes that claim true rather than assumed.

### Verification

- **API unit suite (full):** 1881 files / 35001 tests passed (3 files, 21 tests skipped).
- **Integration suite (full, real test stack):** 533 files / 3670 tests — **3661 passed, 6 failed in 3 files.**
- `rls-coverage` contract suite (its own runner, `vitest.config.rls-coverage.ts`): 102 passed.
- Device-group / group-membership / cascade / org-lifecycle integration subset: 11 files / 75 passed.

The 6 integration failures are **pre-existing on this branch's base and unrelated** — `actionIntentsImmutabilityTrigger` (4), `aiAgentSchedulesPartnerRls` (1, `action_intents` scope tombstone), `portalReportSelfService` (1, `report_runs_execution_scope_principal_chk`). Verified, not assumed: checking `apps/api/src` out at `bad2da120` (this PR's previous head) and re-running those three files reproduces exactly the same 6 failures. None of the three imports any module this PR touches.

_Cost of the fix:_ dynamic group membership now lags a device change by ~5s. That is the trade for the change being applied at all on the enrolment path.

ToddHebebrand:
**Round-3 verification (Opus):** all three findings closed (delay + coalescing, processor throws on handler rejection with idempotent retry, no-context fallback throws; race case 7 confirmed red at delay=0). One residual: a hosted enrolment can hold its transaction past 5s (`issueMtlsCertForDevice` has no fetch timeout), and the delayed `device.created` job then completed silently.

**Fixed inline (orchestrator):** `processDeviceGroupReevaluation` now THROWS when a `device.created` job finds no row, so `attempts: 5` retries with backoff and a real rollback ends in `removeOnFail` retention. `device.updated` keeps the deleted no-op. Red-first unit test added; race case 7's negative control now asserts the rejection. Unit 55/55, tsc clean, integration suite 9/9 against the local test stack.

Ready — merge on green.

Issue #3859, opened 2026-08-23
Full interface localization — translated page shells and error messages


ToddHebebrand:
**Plan exists:** `docs/superpowers/plans/open/2026-07-10-web-i18n-phase3-ssr-and-api-messages.md`.

Phases 1–2 shipped (client-side react-i18next, 5 locales) but a non-English user still gets an English page shell on first paint (title, `<html lang>`, static Astro headings) and English error toasts for server-authored messages. Plan: (A) locale cookie → Astro middleware → pure-function `tServer(locale, key)` for SSR shells; (B) additive `code` field on API error responses mapped centrally to translations inside `runAction`, never changing existing prose `error` strings.

**⚠️ Known fix needed before executing:** the plan hardcodes `['en','pt-BR']` — 5 locales now ship; use the real locale list or 3 locales get English shells (flagged by the 2026-07-21 audit, verdict ✅ do).

**Verified 2026-08-23 against origin/main (`dfe02c5e6`):** no `tServer`, no `locals.locale`, no Astro locale middleware, no error-code catalog; `runAction` still toasts raw server strings.

fabicarvano:
Hi Todd,
I’d like to help with Issue #3859 and implement the existing Phase 3 i18n plan.

I’ll follow the current project conventions, support all existing locales, add/update the required tests, and submit the changes as a PR.

Is anyone currently working on this, or can I take it?

Thanks.

fabicarvano:
Update: opened #5041 with Task 1 of the plan (locale cookie + Astro.locals.locale + <html lang>). One note — the plan hardcodes ['en','pt-BR'], but 8 locales ship today, so I widened the locale surfaces to @breeze/shared's isSupportedLocale/SupportedLocale. Planning to follow with Tasks 2–4 as separate stacked PRs. CI on the PR is awaiting maintainer approval to run.

fabicarvano:
Thanks for the thorough review, Todd! Good catch on the browser-detected-locale gap. That first-paint mismatch is exactly what I want to close in Task 2 with tServer. My plan is to fall back to Accept-Language when there's no cookie, resolved per request so nothing heuristic ever gets pinned. I'll get started on Task 2 next.

PR #5041, opened 2026-09-06
feat(web): locale cookie + Astro.locals.locale + html lang attribute

fabicarvano:
## Summary

Phase-3 **Task 1** of the in-repo plan `docs/superpowers/plans/open/2026-07-10-web-i18n-phase3-ssr-and-api-messages.md`: give the SSR shell a locale to render with, so a non-English user gets the right `<html lang>` on first paint instead of the hardcoded `en`.

- `appearance.ts` now mirrors the localStorage locale preference into a `breeze.locale` cookie (localStorage stays the client source of truth; the cookie exists only so SSR, which can't read localStorage, has the value).
- `middleware.ts` reads that cookie into `Astro.locals.locale` (the middleware previously ignored `context` entirely — it was CSP-only, so this is clean net-new).
- The 5 root layouts render `<html lang={Astro.locals.locale ?? 'en'}>`.
- `App.Locals` is declared in `env.d.ts`.

## Linked issues

Part of #3859. This is the first of the plan's stacked PRs (Task 1); Tasks 2–4 (SSR shell strings / `tServer`, additive API error codes, adoption waves) follow separately, so this PR intentionally does **not** close the issue.

## Deviation from the plan (worth a look)

The plan was written when only `['en','pt-BR']` shipped and hardcodes that pair. The project now ships **8 locales**, so I widened every locale surface to the existing single source of truth in `@breeze/shared`:

- `middleware.ts` validates the cookie with `isSupportedLocale` from `@breeze/shared` (rather than the plan's inline two-literal check), so all 8 locales resolve and the set can't drift out of sync again.
- `env.d.ts` types `locale?: import('@breeze/shared').SupportedLocale`.
- Cookie attributes per the plan: `Path=/; Max-Age=31536000; SameSite=Lax`, not `HttpOnly` (client JS writes it), no hardcoded `Secure` (dev is http).
- The cookie mirror only pins an explicit user/partner locale — an undetermined browser default is a per-request heuristic and is not written server-side.

## Type of change

- [x] New feature (non-breaking)

## Verification

Run locally against `feat/i18n-phase3-task1` (Node 22.23, 4 vCPU):

| Check | Result |
| --- | --- |
| `pnpm lint` | pass |
| `pnpm --filter=@breeze/web exec astro check` | **0 errors**, 0 warnings, 252 hints (1877 files) |
| Task-1 unit tests (`localePreference.test.ts`, `middleware.test.ts`) | **29 passed** — cookie mirror cases + `locals.locale` extraction |
| `pnpm build --filter=@breeze/web` | pass |

`astro check` and the web build need `NODE_OPTIONS=--max-old-space-size=6144` on an 8 GB host.

## Notes

Additive and scoped: no existing behavior changes, no client i18n module imported into `.astro` shells, no React-island SSR touched. Tasks 2–4 build on the `Astro.locals.locale` seam introduced here.

### DIFF
--- a/apps/web/src/env.d.ts
+++ b/apps/web/src/env.d.ts
@@ -1,5 +1,11 @@
 /// <reference types="astro/client" />
 
+declare namespace App {
+  interface Locals {
+    locale?: import('@breeze/shared').SupportedLocale;
+  }
+}
+
 interface ImportMetaEnv {
   readonly PUBLIC_ENABLE_ENDPOINT_AV_FEATURES?: string;
   readonly PUBLIC_ENABLE_NETWORK_DEVICES_IN_LIST?: string;
--- a/apps/web/src/layouts/AuthLayout.astro
+++ b/apps/web/src/layouts/AuthLayout.astro
@@ -10,7 +10,7 @@ const { title } = Astro.props;
 ---
 
 <!DOCTYPE html>
-<html lang="en">
+<html lang={Astro.locals.locale ?? 'en'}>
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
--- a/apps/web/src/layouts/AuthShellBranded.astro
+++ b/apps/web/src/layouts/AuthShellBranded.astro
@@ -17,7 +17,7 @@ const {
 ---
 
 <!DOCTYPE html>
-<html lang="en">
+<html lang={Astro.locals.locale ?? 'en'}>
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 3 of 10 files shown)


ToddHebebrand (approved):
**Approved.** Small, additive, and tested — exactly the right slice for Task 1. Verified: all five root layouts with a hardcoded `lang="en"` are covered (no stragglers under `apps/web/src`), `isSupportedLocale` is the existing single source of truth in `packages/shared/src/utils/locale.ts:7`, and the cookie attributes (`SameSite=Lax`, no `HttpOnly`, no hardcoded `Secure`) are the right call for a non-sensitive, client-written preference.

**Deviation from the plan:** agreed, and it's the better outcome — the plan's hardcoded `['en','pt-BR']` pair is stale, and routing the check through `@breeze/shared` means the set can't drift again.

**Nice-to-have, not blocking:** `applyResolvedLocalePreferences` clears the cookie when neither user nor partner locale is set, so a user with only a browser-detected locale flips back to `<html lang="en">` on first paint even though the client will render in their detected language. That's consistent with your stated rule (don't pin heuristics server-side), so leaving it — just flagging it as the one place where SSR and client can disagree until Task 2.

Fork CI approved; I'll merge on green. Thanks for the clean first PR.

Issue #5023, opened 2026-09-06
Device removal completion — Remove agent choice, bulk restore/purge, uninstall state


ToddHebebrand:
Tracking issue for **Device removal completion — Remove agent choice, bulk restore/purge, uninstall state** (`device-removal-completion`).

- Plan: `docs/superpowers/plans/device-lifecycle/2026-09-05-device-removal-01-remove-dialog.md`
- Spec: `docs/superpowers/specs/device-lifecycle/2026-09-05-device-removal-completion-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All three waves merged 2026-09-05: W01 #5034 (00960f560), W03 #5035 (3d5f7f83e), W02 #5042 (65b45470f). Closes the remaining scope of #3987, #2787, #2250. Unreleased — ships next release. Not done: browser walk of the bulk bar / typed-count purge dialog (unit + integration coverage only); machine-quality translations for the new locale blocks.

ToddHebebrand:
Follow-ups merged 2026-09-06: #5047 (single permanent delete cascades in a system context, matching bulk purge) and #5053 (`device_lifecycle` config-policy feature — purge removed devices after N days — daily job, new `devices.decommissioned_at`; closes #2787 item 4). Unreleased.

ToddHebebrand:
Browser walk done 2026-09-06 on a live stack at a1f6b3c4b: 8/10 flows PASS on first run; defects fixed in #5058 (uninstall badge never rendered — detail page dropped the `uninstall` field; stale advanced-filter id set after Restore; single Delete permanently now confirm-gated like bulk). Remaining paper cuts (not fixed): inherited-policy state only shown with `?linked=` param; "N removed hidden — show" swaps to a removed-only filter rather than adding rows; Compare vanishes silently at 5+ selected.

Issue #3986, opened 2026-08-25
[API] "Remove device" must queue a durable self_uninstall — the only uninstall path today is WS-only best-effort and never reaches an offline agent


ToddHebebrand:
## Description

Offboarding a single device leaves the Breeze agent installed and running whenever that machine is offline at the moment of deletion — and once the device row is gone there is no path left to ever uninstall it remotely.

This is the API half of a two-part change. The UX half — renaming the action to **Remove**, asking about the agent at removal time, and splitting purge into a separate **Delete permanently** — is #3987. Read that first for the shape; this issue is what has to exist underneath it.

Flow today:

1. **Decommission** (`DELETE /devices/:id`) — soft delete. Sets `status = 'decommissioned'`, tears down remote sessions, force-closes the agent WS, writes an audit row. **Sends no uninstall at all**, by design.
2. **Permanent delete** (`DELETE /devices/:id/permanent`) — the only single-device path that sends `self_uninstall`, and it is **WS-only, fire-and-forget**.

So a tech who decommissions a powered-off laptop and then permanently deletes it has an agent still installed, still running as a service, heartbeating forever into a `403`.

## Root Cause

`apps/api/src/routes/devices/core.ts:1559`:

```ts
// Best-effort: send self_uninstall command if the agent is online.
// We don't block deletion on this succeeding — fire and forget.
let uninstallSent = false;
if (device.agentId && isAgentConnected(device.agentId)) {
  try {
    uninstallSent = sendCommandToAgent(device.agentId, {
      id: `uninstall-${deviceId}`,
      type: CommandTypes.SELF_UNINSTALL,
      payload: { removeConfig: true },
    });
  } catch (err) { /* logged only */ }
}
```

`sendCommandToAgent` pushes over the live WebSocket only. No durable `device_commands` row, no retry, no "deliver when it next checks in".

**And it cannot simply be swapped for a queued command at this call site** — which is the crux of why the fix is a re-shape rather than a one-liner:

- `device_commands` is in `CORE_DEVICE_CASCADE_DELETE_TABLES` (`core.ts:224`), so `deleteDeviceCascade` wipes any queued row in the same transaction as the delete.
- Even if the row survived, `staleCommandReaper` expires uncollected commands after 30 minutes. Its `self_uninstall` exemption (`jobs/staleCommandReaper.ts:191-211`) is deliberately scoped to tenants *currently* in `offboarding`, so it would not cover this path.

[...]

PR #5042, opened 2026-09-06
feat: bulk restore + async bulk permanent delete for removed devices; harden single lifecycle routes (#2787)

### LINKED ISSUE #2787: Bulk permanent delete for decommissioned devices
## Description

An operator trying to clean up ~50 decommissioned devices found there is no way to bulk-delete them. Reported in Discord, 2026-07-25.

The bulk-selection bar on the Devices page offers Decommission, but decommission is a soft delete. Permanent delete is only reachable one device at a time from the device settings modal, so purging a decommissioned fleet means opening 50 modals.

## Current state

**API** — there is exactly one true bulk device endpoint, `POST /devices/bulk/commands` (`apps/api/src/routes/devices/commands.ts:24`), and its allowed types are agent commands only (script, reboot, shutdown, wake, refresh_inventory, …). No delete/decommission/restore type.

Lifecycle routes are strictly single-device:
- `DELETE /devices/:id` (decommission / soft delete) — `apps/api/src/routes/devices/core.ts:1235`
- `POST /devices/:id/restore` — `core.ts:1300`
- `DELETE /devices/:id/permanent` (hard delete, requires `status === 'decommissioned'`) — `core.ts:1343`

Other domains already have the pattern: `quotes/bulk.ts:16`, `invoices/bulk.ts:15`, `contracts/bulk.ts:15`, `tickets/bulk.ts:31`, `alerts/alerts.ts:393`.

**Web** — the bulk action bar (`apps/web/src/components/devices/DeviceList.tsx:1961`) exposes Reboot, Run script, Deploy software, Maintenance on/off, Wake, Link as multiboot, Link as VM host, and Decommission. The only destructive entry, Decommission, is a client-side sequential loop over `DELETE /devices/:id` (`apps/web/src/services/deviceActions.ts:471-487`), not a bulk API call. Permanent delete and Restore live only in `DeviceSettingsModal.tsx:207-238`; the row kebab menu doesn't offer permanent delete at all.

Filtering to decommissioned already works — `DecommissionedHiddenHint.tsx` gives a one-click \"N decommissioned hidden, show\" — so a user can get the exact selection they want and then find no action for it. Worse, the one action offered (Decommission) is a no-op on already-decommissioned rows.

There is no purge/retention job either: no `/devices/purge` route, no cron pruning decommissioned devices, no admin tooling.

## Proposed fix

1. `POST /devices/bulk/delete` taking `deviceIds[]`, requiring `status === 'decommissioned'` per row, returning `{succeeded, failed}`. Batch over the existing single-row cascade path (`core.ts:1414`) rather than inventing a second delete path, so `CORE_DEVICE_CASCADE_DELETE_TABLES` coverage stays honest.
2. Add `bulk restore` alongside it — same gap, same selection.
3. Surface both in the bulk action bar, gated on the selection being all-decommissioned (extend `DECOMMISSION_BLOCKED_BULK_ACTIONS` in `bulkActionGating.ts:80-86`).
4. Optional follow-on: a \"purge decommissioned older than N days\" action, which is what the reporter actually wanted.

## Affected files

- `apps/api/src/routes/devices/core.ts` (primary)
- `apps/api/src/routes/devices/commands.ts` / a new `devices/bulk.ts`
- `apps/web/src/components/devices/DeviceList.tsx`
- `apps/web/src/services/deviceActions.ts`
- `apps/web/src/components/devices/bulkActionGating.ts`

Related: #2250, #2251.

### LINKED ISSUE #5025: W02: Lifecycle service hardening + bulk restore + async bulk purge + selection-aware bulk bar (#2787)
Wave W02 of #5023 — Lifecycle service hardening + bulk restore + async bulk purge + selection-aware bulk bar (#2787)
Branch: `feature/5023-device-removal-completion/wave-5025`

ToddHebebrand:
Wave W02 of #5023. A selection of removed devices now offers exactly **Restore** and **Delete permanently**, both work in bulk, and single + bulk share one hardened lifecycle service so a purge can never race a restore or destroy a pending agent uninstall.

## Two latent defects in the pre-existing single routes, and the fix

Both were real, both were reachable from the UI today, and both are closed by moving the state machine into `services/deviceLifecycle.ts`, whose every operation takes `devices FOR UPDATE` as its **first** statement and decides afterwards.

**1. TOCTOU on permanent delete.** `DELETE /devices/:id/permanent` checked `status === 'decommissioned'` from `getDeviceWithOrgAndSiteCheck` — *outside* the deletion transaction — and never re-checked it. A Restore committing between that read and the cascade was silently purged: the operator restores a device, another tab's delete lands, and the device is gone with no error anywhere. `purgeRemovedDevice` re-reads the status under the row lock, so the loser of that race gets `409 NOT_REMOVED`. Pinned unit-side (`cascadeDelete.test.ts`) and against real Postgres (`deviceLifecycle.integration.test.ts`).

**2. Lock-order inversion on restore.** `POST /devices/:id/restore` called `releaseDeviceRemoveReason` — which locks `device_commands` rows — *before* touching the `devices` row. Every other writer that spans both levels (the cascade, re-enrollment, site move, moveOrg) takes `devices` first. That is textbook AB-BA, i.e. 40P01 under contention, and it is the exact class `deviceDeletion.ts:79` documents. `restoreRemovedDevice` locks `devices` first; `core.decommission.test.ts` now asserts the full order `lock-devices-row → release-device-remove-reason → status-flip` rather than only the last two.

The service also carries the 3s `tightenLockTimeout` bound that used to live only in `deleteDeviceCascade` — since the lifecycle lock now runs *first*, the cascade's own bound would never be reached on a contended row, and the route's existing 55P03 branch would have gone dead.

## The pending-uninstall refusal

Purge now **refuses** (`409 UNINSTALL_PENDING`) while a `device_remove` `self_uninstall` is still `pending`/`sent` and unexpired, and the legacy fire-and-forget WS uninstall in `DELETE /:id/permanent` is **removed**.

Why: `device_commands` is *in* the device cascade. Purging a device whose durable uninstall (#3986) has not been collected destroys the only thing that will ever remove the agent from the endpoint — a zombie agent nobody can see or reach. The WS dispatch that was supposed to cover that only ever reached a **connected** agent, and decommissioning (a precondition of getting here) force-closes the agent socket, so in practice it almost never fired; when it did, it raced the cascade that had just deleted its own command row. Refusing and telling the operator to wait, or to restore and re-remove choosing "Leave the agent installed", is strictly more honest.

Consequence for the response contract: `DELETE /:id/permanent` now returns `{ success: true }` — `agentUninstallSent` and `warning` are gone, and the 409 bodies no longer carry `uninstallSent`. There is nothing best-effort left to report.

## Why bulk purge is async and bulk restore is not

`deleteDeviceCascade` touches ~40 tables per device. 500 of those inside one request would pin a pooled connection for minutes under the auth middleware's request transaction (#1105), and one bad row would abort them all. So `POST /devices/bulk/permanent-delete` does only the cheap checks (access, `status='decommissioned'`), returns `202 { jobId, accepted, rejected }`, and the `device-bulk-purge` BullMQ job does the work; the web polls `GET /devices/bulk/purge-runs/:jobId`.

A restore is two small writes per device, so `POST /devices/bulk/restore` is synchronous — but each device still runs in its **own** short RLS transaction via `runBulkIsolated`, and the route is registered in `middleware/selfManagedDbContextRoutes.ts` so there is no ambient request transaction to hold across the loop. `bulk/permanent-delete` is deliberately *absent* from that list: it only validates and enqueues.

**The worker re-derives authorisation per device, under the lock.** It runs in a system DB context (the cascade must see tables a tenant context hides), so RLS is not a backstop. The payload carries the org each device belonged to at confirm time; a device that moved org in the meantime is skipped as `ORG_CHANGED`, never deleted under stale authorisation. Proven against real Postgres, not mocks. A pending uninstall is **not** pre-checked in the route — the worker refuses it under the lock, keeping one source of truth for that rule.

`DEVICE_BULK_PURGE_MAX_TARGETS` re-enforces the 500 ceiling in the worker as well as the zod schema: the payload outlives the request and is executed by a different process.

## The three gating sets

`bulkActionGating.ts` gains `REMOVED_ONLY_BULK_ACTIONS` (`restore`, `permanent-delete`) beside the existing `DECOMMISSION_BLOCKED_BULK_ACTIONS` and `INTENTIONALLY_UNGATED_BULK_ACTIONS`. They answer different questions: the first two are per-device *filters*, the new one is a per-*selection* visibility rule — both bulk endpoints require `status='decommissioned'` and would reject every active device in a mixed batch. `classifyBulkSelection` drives it; a mixed selection deliberately gets the ordinary menu, which already skips removed devices, rather than the removed-only one, which would fail on most of them. An empty selection is `'active'`, not the vacuously-true `'removed'`.

`DeviceList.test.tsx` now enumerates the real bulk menu in **both** selection states and forces every emitted action into exactly one of the three sets — the previous single-state enumeration would have passed a menu that showed Restore/Delete permanently unconditionally.

`BulkPurgeDialog` requires the device count to be typed before Confirm arms (new `confirmDisabled` prop on `ConfirmDialog`, with its own test). This is the only irreversible bulk action in the product and, unlike single Remove, has no 5-second undo window; a type-the-number gate is the cheapest control that fails an accidental select-all. It also names the first five hostnames and warns when the selection spans more than one org (usually an unnoticed "All organizations" scope). Mutation-verified: replacing `confirmDisabled={!confirmed}` with `false` reds two tests.

## Test counts (all verified locally, post-merge with origin/main)

| Suite | Result |
|---|---|
| API unit — `src/routes/devices`, `deviceLifecycle.test.ts`, `deviceBulkPurge.test.ts`, `selfManagedDbContextRoutes.test.ts`, `bulkOps.test.ts` | **54 files / 944 tests passed** |
| API — `workerRegistry.test.ts` + `workerEntrypointClosure.contract.test.ts` | **121 tests passed** |
| API integration (real Postgres, `pnpm test-stack`) — `deviceLifecycle.integration.test.ts` (7) + `deviceUninstallDrain.integration.test.ts` (11) | **2 files / 18 tests passed** |
| Web — `src/components/devices`, `src/services`, `src/lib/i18n`, `src/lib/__tests__` | **102 files / 1339 tests passed** |
| `tsc --noEmit` apps/api (`NODE_OPTIONS=--max-old-space-size=8192`) | exit 0 |
| `tsc --noEmit` apps/web | exit 0 |
| `pnpm lint` | exit 0 |

The integration suites really ran against a live Postgres (a private worktree stack via `pnpm test-stack up`, torn down again with `pnpm test-stack down` — confirmed no containers remain). No new migrations in this wave; no cascade/RLS registration lists were touched because no table or column was added.

## Deviations from the plan (each with its reason)

1. **`cascadeDelete.test.ts` does not mock `services/deviceLifecycle`.** The plan's Task 2 said to. Mocking it would have neutered ~6 existing tests that assert on the *compiled cascade SQL* (ticket detach, `discovered_assets.link_source`, link-group dissolve) — they would have passed while exercising nothing. Instead `rigDeleteTransaction` gained knobs (`status`, `linkGroupId`, `pendingUninstall`, `deviceMissing`) so the real service runs against a rigged tx. Strictly more coverage than the plan asked for.
2. **`lockDevice` bounds the lock wait** (`tightenLockTimeout`, 3s) — not in the plan's code sketch. Without it the route's 55P03 branch becomes unreachable, since the lifecycle lock now precedes the cascade's own bound. Pinned by a test.
3. **`purgeRemovedDevice` reads `link_group_id` from the LOCKED row**, not from the caller's pre-flight lookup (the plan's sketch also did this; calling it out because it changes how `cascadeDelete.test.ts` rigs the link-group test). The lookup's copy predates the lock and can be stale.
4. **Task 7's two rendering tests were not written as `it.todo`.** `it.todo` takes no body, so the helper would have been dead code in that commit. Task 7 ships the set-disjointness + `classifyBulkSelection` tests; Task 8 adds the enumeration tests together with the rendering they exercise. Same end state, no dead code in between.
5. **`workerRegistry` placement is `socket-owner`, not `global`.** The plan guessed `global`; `workerEntrypointClosure.contract.test.ts` — the mechanical authority — showed the closure reaches `routes/agentWs.ts` and `services/agentCommandAwait.ts` via `deviceLifecycle → deviceDeletion`. Same class as `orgMerge`. Registry counts updated 123 → 124 in both contract tests.
6. **`startBulkPurge` parses the error body by hand** instead of going through `getErrorMessage`. That helper consumes the response stream, and the 409 carries `rejected` alongside `error`; reading it through the helper would have left the per-device reasons unrecoverable.
7. **`RestoreResult.device` is `| undefined`.** `.returning()` is array-typed and every caller already read it defensively (`result.device?.hostname ?? …`); asserting a non-null that only holds by argument would be worse. Surfaced by `tsc` after the merge with main.
8. **`bulkPurgeStarted` toast is `type: 'success'`, not `'info'`.** The `showToast` union has no `'info'`. Surfaced by web `tsc`.
9. **The web branches that surface `agentUninstallSent` / `warning` from `permanentDeleteDevice` are left in place but are now inert** — the API no longer emits either field. `DeviceDetailPage.permanentDelete.test.tsx` and the `#4368` describe in `DevicesPage.test.tsx` still pin behaviour against mocked service responses, so they pass, but they now assert a contract the API no longer has. Removing them is web-only cleanup outside this wave's file list; flagging it rather than smuggling it in.
10. **Task 10's stack-walk / screenshot step was SKIPPED — NOT CHECKED.** No browser stack was available in this environment. The bulk bar's menu swap, the typed-count dialog, the progress toast and the post-purge refetch are covered by jsdom tests only; nobody has driven this in a real browser.

## Claim labelling

- **Verified:** every test count above, both `tsc` runs, `pnpm lint`, the integration suites executing against live Postgres, the test stack teardown, the `workerRegistry` placement verdict, and the `BulkPurgeDialog` mutation check.
- **Inferred:** that the removed WS uninstall "almost never fired in practice" — reasoned from decommission force-closing the agent socket (`core.ts` `disconnectAgent`) plus the connected-agent guard, not measured against production telemetry.
- **Not checked:** anything in a real browser (deviation 10); behaviour under genuine concurrent load (the races are proven by sequenced transactions against real Postgres, not by two simultaneous connections).

Closes #5025
Closes #2787
Refs #3987


ToddHebebrand:
All five confirmed findings and both minors are fixed in this PR, one commit each, each driven red-first. Head is now `aa2257c868432fef315d3a215c1d0bdde2ff9507`.

### 1 — TENANCY: purge-run status failed open, and partner scope skipped the org check · `dfd714e6e`

Both arms were wrong in different directions and are now one rule.

- The org arm was guarded by `&& payload &&`, so a job with no readable `data` **skipped the tenancy check and returned the run to any caller**, while the partner arm failed *closed* on the same input. Now: `auth.scope !== 'system' && (!payload || !payload.targets.every(t => auth.canAccessOrg(t.orgId)))` → 404. An unverifiable owner is not an authorised one; system scope stays exempt because it already spans every partner and org.
- Partner scope only compared `partnerId`. A partner member with `org_access = 'selected'` shares the partner id with every org under that partner, so the arm passed while their selection could exclude the orgs the run touched — exactly the case `routes/orgMerge.ts:126` adds its own `canAccessOrg` check for. The partnerId arm is kept *as well*.

You were right that the existing "denies another partner's run" test asserted nothing: the auth mock was hardcoded to `scope: 'organization'`, so the partner branch it named was unreachable and the test passed on the org branch. The mock is now scope-mutable, with the three cases you asked for — (i) other partner → 404, (ii) same partner, `canAccessOrg` false for one target org → 404, (iii) payload undefined → 404 — plus a **positive control** (own partner, selection covers every target → 200) so the three denials can't pass against a route that 404s unconditionally, and a system-scope case pinning the exemption.

### 2 — AUDIT: bulk purge discarded `PurgeResult` · `97d9e69e0`

`purgeOne` now returns `{ code, linkGroupId, linkGroupDissolved }` and the worker spreads the two link-group fields into the per-device audit row. Only when the device was actually in a group, so the field means something wherever it appears — pinned by a negative test asserting the fields are **absent** for an ungrouped device, since a blanket `linkGroupId: null` on every entry would have satisfied the positive test while making the trail useless.

### 3 — AUDIT: single route keyed the spread on the pre-lock copy · `0cbd105c9`

`PurgeResult` gained `linkGroupId` as read **under the lock**, and `core.ts` keys its spread on that. The failing test that drove it: pre-flight reports `linkGroupId: null`, the locked row reports `grp-late-link`, the dissolve fires — and the audit details were `{}`. A group dissolve unlinks sibling devices that were never in the request, so that was silently unexplainable.

### 4 — NEGATIVE CONTROL on the refusal predicate · `a04907b01`

Two integration cases against the live stack, each differing from a genuinely-draining row in **exactly one arm** so a pass pins that arm rather than "some row shape":

- `device_remove` reason, still `pending`, deadline backdated → purge **PROCEEDS** (isolates `device_remove_expires_at > now()`, evaluated by the DB clock — which is why this can't be argued on compiled SQL). Includes a control asserting the row is still pending and still carries the reason, so the deadline is provably the only thing that changed.
- `pending` + unexpired deadline, `uninstall_reasons = ['tenant_offboarding']` only → purge **PROCEEDS** (isolates the `@> ARRAY['device_remove']` arm — the incident `deviceUninstallDrain.ts`'s module doc exists to prevent).

**Mutation-verified:** dropping both arms from the predicate reds exactly these two and nothing else.

### 5 — DEAD CONTRACT: `agentUninstallSent` / `warning` on the web · `659a0d40d`

Return type narrowed to `{ success: boolean }`; both `result.warning` branches deleted (`DevicesPage.tsx`, `DeviceDetailPage.tsx`); both #4368 describes rewritten rather than deleted, so the surface keeps coverage. Each now has a **discriminating** case: a body still carrying a stray legacy `warning` must not resurrect the old toast — that is what proves the branch is gone rather than merely unreachable with today's payloads.

Two orphans swept in the same pass, both consequences of the removal rather than scope creep:
- `permanentlyDeletedWithWarning` is unreachable in both namespaces → removed from all 8 locales.
- `permanentlyDeleted` **ignored the `{{hostname}}` its call site already passes**, so the completion toast read a bare "Permanently Deleted" while the undo toast immediately before it named the device. Same paper cut that had already been fixed for the undo toast. Fixed in all 8 locales — this surfaced because the rewritten test asserts the hostname, and I'd rather fix the string than weaken the assertion.

### 6 (minor) — 500 sequential selects in the request tx · `6850b6f08`

New `getDevicesWithOrgAndSiteCheck` batches into one `WHERE id IN (...)`. It **delegates to the same exported `ensureOrgAccess` and `canAccessSite` predicates** the single chokepoint uses rather than replicating them, so the verdicts cannot drift — and keeps the fail-closed posture: missing row, org-denied, malformed uuid, or (under a site allowlist) a row with no usable `site_id` are all denials. The programmer-error throw for a missing permissions context is preserved and deliberately placed so an all-denied batch can't skip it.

Seven helper tests including the site-allowlist split and the null-`siteId` fail-closed case, plus the route-level test you asked for: a site-restricted caller gets `SITE_ACCESS_DENIED` for the out-of-allowlist device while the rest still enqueue. Bulk **restore** deliberately keeps the per-device lookup — each item runs in its own RLS context inside `runBulkIsolated`, so a batched lookup outside those contexts would be a different scope.

### 7 (minor) — vacuous `-1` index · `aa2257c86`

Added, and extended to the other two ordering assertions in the file for the same reason: `indexOf` returns `-1` for a statement never issued, and `-1` compares "less than" every real index. **Mutation-verified:** removing `tightenLockTimeout` from `lockDevice` now reds the bounds-the-wait test, which it did not before.

---

### Verification (all re-run at `aa2257c86`, after the merge with main)

| | |
|---|---|
| API unit (`routes/devices`, `deviceLifecycle`, `deviceBulkPurge`, `selfManagedDbContextRoutes`, `bulkOps`, `workerRegistry`, closure contract) | **56 files / 1081 passed** |
| API integration vs live Postgres — `deviceLifecycle` (9) + `deviceUninstallDrain` (11) | **2 files / 20 passed** |
| Web (`components/devices`, `components/shared`, `services`, `lib/i18n`, `lib/__tests__`) | **116 files / 1461 passed** |
| `tsc --noEmit` api / web (both with `NODE_OPTIONS=--max-old-space-size=8192`) | exit 0 / exit 0 |
| `pnpm lint` | exit 0 |

Test stack torn down (`pnpm test-stack down`); `docker compose ls -a` no longer lists `breeze-test-feature-5023-device-removal-com-4b8f3b`. The other `breeze-test-*` projects still running belong to unrelated worktrees and were left alone.

**Still not checked:** the browser walk (deviation 10 in the PR body) — no stack available in this environment.

PR #5043, opened 2026-09-06
feat(api): custom-field definitions importer — preview + commit (#3257 W07)

### LINKED ISSUE #4775: W07: Definitions importer
Wave W07 of #4768 — Definitions importer
Branch: `feature/4768-rmm-custom-field-import/wave-4775`

ToddHebebrand:
Closes #4775

Wave 07 of #4768 (#3257): the **definitions** half of the RMM custom-field importer. No migration in this wave. No `apps/web`, no values importer (W08).

Depends on and builds directly on what W02 (#4991), W03 (#5028) and W06 (#4973) landed — all three verified merged on `origin/main` at `901bc6c56` before this branch was cut.

## Routes

| Method | Path | Behaviour |
|---|---|---|
| POST | `/custom-fields/import/preview` | Annotates every row against current state. Writes nothing. |
| POST | `/custom-fields/import` | Creates the acknowledged definitions. Always **200**, even with a non-empty `errors[]`. |

Both mount `requireScope('organization','partner','system')` + `requirePermission(devices:write)` (the same grant the sibling `POST /custom-fields` create route uses) + `requireMfa()`, and plain `authMiddleware` — **no `dualAuth`**, so an `X-API-Key`-only caller is a 401.

## Codes

**Preview annotations** (`DefinitionAnnotation`)

| Annotation | Meaning |
|---|---|
| `create` | No definition owns this key on the row's axis. |
| `already-exists` | Same axis, same key, **same type** — a re-import. Skipped at commit, never rewritten. |
| `type-conflict` | Same axis, same key, different type (type is immutable on update); **or** the key appears more than once in the submitted batch. |
| `key-shadowed` | The key is taken on the **other** ownership axis under this partner — W03's `custom_field_definitions_no_shadow` trigger would refuse the write. |
| `org-not-found` | `ownerScope: 'organization'` with no organizationId, or one outside the caller's reach. Same annotation for "absent" and "not yours", so the response is never an existence oracle. |
| `partner-wide-denied` | `ownerScope: 'partner'` from a caller without `canManagePartnerWidePolicies`. **Its own code** — see the deviation note below. |

**Commit error codes** (`DefinitionImportErrorCode`): `org-not-found`, `type-conflict`, `key-shadowed`, `annotation-changed`, `match-changed`, `partner-wide-denied`, `write-failed`.

## What is actually guaranteed

- **TOCTOU.** Commit re-derives *every* annotation against state loaded at commit time (never preview's) and refuses a row whose annotation moved (`annotation-changed`). An `already-exists` acknowledgement must pin `expectedDefinitionId` (enforced at the zod layer) and is refused `match-changed` if the key changed hands. An acknowledged `create` that finds a row now existing is an error, never a silent no-op.
- **Per-row isolation without losing RLS.** Each row's insert runs in a **nested `db.transaction`** — a SAVEPOINT inside the request's own `withDbAccessContext` transaction (`dbSavepointErrorIsolation.integration.test.ts` is the proof) — issued on the nested `tx`, never the ambient `db` proxy. One refused row rolls back to its own savepoint; the rest of the batch proceeds; and RLS remains a real second control on every write.
- **No driver text in a response body.** SQLSTATEs map to fixed copy; the original error rides along **non-enumerably** on `errors[].cause` so error tracking keeps the stack and `JSON.stringify` cannot leak query text or column values.
- **One `writeRouteAudit` per definition created**, carrying `externalSystem`, `sourceLabel` (e.g. `udf7`), `ownerScope`, `fieldKey` and the row count. `sourceLabel` is deliberately never stored on the definition row — the audit event is its only durable home.

1. **`routes/index.ts` does not exist** — routes are mounted in `apps/api/src/index.ts`. Mounted there, *before* `customFieldRoutes`, so `/custom-fields/import*` can never fall through to the CRUD app's `/:id` handlers.
2. **Writes stay in the request context, not a system context.** The plan had commit open each row's transaction via `runOutsideDbContext`. That would bypass RLS on the write path. A nested transaction gives the same per-row failure isolation *and* keeps RLS, so that is what shipped. The snapshot **read** does still escalate (once, for both queries) because `custom_field_definitions` has no partner-wide RLS SELECT branch — it is in `PARTNER_WIDE_SELECT_BRANCH_EXEMPT` (TODO #4944) — so an org-scoped request context cannot see partner-wide rows at all and would annotate every shadowed key as `create`. Bounded by `partnerId` AND `accessibleOrgIds`, exactly as `contacts/import.ts` does.
3. **The plan's fifth Task-2 test was deliberately wrong and is fixed**: a capability refusal is `partner-wide-denied`, not `org-not-found`. Reusing `org-not-found` would tell a tech "that organization does not exist" when the truth is "you may not create all-organizations fields".
4. **The partner-wide capability refusal is a whole-request 403 at the route**, matching the sibling `POST /custom-fields` create route (epic #2135) and the brief's integration requirement, *in addition to* the per-row service annotation. Authorization answers belong in the same band as the scope/permission gates, not mixed into a 200 that also reports data problems; the service annotation stays because the service is reachable directly and must fail closed there too.
5. **The commit route returns 200 with `errors[].code = 'key-shadowed'`, not 409**, when a shadow is submitted. The brief's integration item (b) asked for 409, but the plan's global constraint — *always 200, because the web caller reads this through `runAction`, which treats a failure body as a hard failure and would hide the rows that DID import* — wins on a batch endpoint. The 409 path is preserved and still asserted where it belongs: the single-create `POST /custom-fields` route is unchanged and still returns 409 `field-key-shadowed` / `field-key-duplicate`.
6. **The 23505/P0001 mapper was extracted, not forked** — into `services/customFields/writeErrors.ts`, now used by both `routes/customFields.ts` and the importer. This touches two files outside the brief's "files you own" list (`routes/customFields.ts`, and the new `writeErrors.ts`); both are outside W05's ownership, and the brief explicitly authorised "reuse or extract that mapper, do not fork it".
7. **Batch-aware annotation.** Beyond the plan's same-axis duplicate check, a *cross-axis* duplicate within one submitted batch is annotated `key-shadowed` too. Without it the second row would have arrived at the trigger as an unexplained P0001. The pass is deterministic and ordered, so preview and commit derive identically — which is what makes `expectedAnnotation` meaningful.
8. **`MAX_IMPORT_VALUES`** is declared here (per plan Task 1) although only W08 consumes it.

`MAX_IMPORT_ROWS = 1000` — the same cap as the org and contact importers (`services/contacts/types.ts:43`).

## Verification

- `npx vitest run src/services/customFields src/routes/customFieldImport src/routes/customFields src/openapi src/__tests__/partner-wide-write-coverage.test.ts` — **16 files, 249 tests, all pass.** (22 service tests were written red-first against a missing module; the route suite's partner-wide gate was mutation-checked — disabling the gate turns 2 tests red.)
- `npx vitest run --config vitest.integration.config.ts src/__tests__/integration/customFieldDefinitionImport` — **9/9 pass** against a real Postgres (`pnpm test-stack up` + `pnpm db:migrate`). Covers all four required scenarios plus savepoint isolation and the no-leak assertion. The trigger test confirms the live constraint name is `custom_field_definitions_no_shadow` and that the SQLSTATE is only reachable through `.cause` (a check on `err.code` would have passed vacuously).
- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` — clean.
- `pnpm lint` — clean (6/6 tasks).

## Not verified

- **The full `apps/api` unit suite and the full integration suite** were not run — only the affected files listed above, plus `partner-wide-write-coverage`, `autoMigrate`, `migrationRlsScope` and `validation.imports` as contract guards. CI covers the rest.
- **`rls-coverage.integration.test.ts` / `tenantCascade` / `tenant-export-policy`** were not run: this wave adds no table, no column and no migration, so none of the six registration lists changes. Asserted by inspection, not by execution.
- **No web client exists yet** (W09), so the wire contract has not been exercised by a real browser caller, and `runAction`'s handling of these bodies is reasoned about rather than observed.
- **`X-API-Key` returns 401** is asserted against a mock of `authMiddleware` that reproduces production's Bearer-only contract, not against the real middleware; the real one has no API-key branch at all (verified by inspection: no `X-API-Key` reference in `middleware/auth.ts`).
- **Concurrency beyond the modelled race**: the TOCTOU test creates the racing row between two sequential requests. Two genuinely simultaneous commits are handled by W02's unique index and W03's trigger (mapped to typed row errors), but that was not exercised under real parallelism.
- **Performance at the 1000-row cap** was not measured.

https://claude.ai/code/session_01R7RWzDtuzApHEcg5Ri9wFt

### DIFF
--- a/apps/api/src/__tests__/integration/customFieldDefinitionImport.integration.test.ts
+++ b/apps/api/src/__tests__/integration/customFieldDefinitionImport.integration.test.ts
@@ -1,0 +1,421 @@
+/**
+ * Custom-field DEFINITION importer against real Postgres (#3257 W07, #4775).
+ *
+ * Four things this suite exists to prove, none of which a mocked unit test can:
+ *
+ *  a. A partner-wide import from a caller without `canManagePartnerWidePolicies`
+ *     is refused **403** at preview AND at commit, through the real route stack
+ *     (real `resolveImportPartnerId`, real capability helper) — not a 200 with
+ *     an annotation the browser might ignore.
+ *  b. A cross-axis shadow is annotated `key-shadowed` at PREVIEW, and if
+ *     submitted anyway the commit refuses it with the same code — driven by
+ *     W03's real `custom_field_definitions_no_shadow` trigger (P0001), not by a
+ *     stub.
+ *  c. TOCTOU: a definition created BETWEEN preview and commit flips the
+ *     annotation, and the row is rejected with `annotation-changed` rather than
+ *     silently becoming a no-op or a duplicate.
+ *  d. Organization isolation: a row naming an organization outside the caller's
+ *     reach never lands, and reports `org-not-found` rather than acting as an
+ *     existence oracle.
+ *
+ * ── Why the middleware is re-created rather than stubbed away ───────────────
+ * `authMiddleware` does two things this suite depends on: it puts an
+ * `AuthContext` on the request, and it opens the request's
+ * `withDbAccessContext` transaction. `fakeAuth` below does BOTH, so the route
+ * handlers, the service's per-row nested `db.transaction` (a SAVEPOINT inside
+ * that request transaction) and RLS all behave exactly as they do in
+ * production. Only the token exchange is skipped.
+ *
+ * The importer's snapshot READ runs in a system context by design (the table
+ * has no partner-wide RLS SELECT branch — `PARTNER_WIDE_SELECT_BRANCH_EXEMPT`,
+ * TODO #4944), so RLS is NOT the control on it. Test (d) is therefore an
+ * app-layer assertion by necessity, and asserting it here — under a real
+ * system-context read — is the only place it can be proven at all.
+ */
+import './setup';
+import { afterEach, describe, expect, it } from 'vitest';
+import { and, eq } from 'drizzle-orm';
+import { Hono } from 'hono';
+import { db, withDbAccessContext, type DbAccessContext } from '../../db';
+import { customFieldDefinitions } from '../../db/schema';
+import {
+  assignUserToOrganization,
+  assignUserToPartner,
+  createOrganization,
+  createPartner,
+  createRole,
+  createUser,
+  grantRolePermissions,
+} from './db-utils';
+import { pgErrorCode, pgErrorNode } from '../../utils/pgErrors';
+import { customFieldImportRoutes } from '../../routes/customFieldImport';
+import { PARTNER_WIDE_WRITE_DENIED_MESSAGE } from '../../services/partnerWideAccess';
+
+const runDb = it.runIf(!!process.env.DATABASE_URL);
+
+const SYSTEM_CTX: DbAccessContext = {
+  scope: 'system',
+  orgId: null,
+  accessibleOrgIds: null,
+  accessiblePartnerIds: null,
+  userId: null,
+};
+
+const createdDefinitions: string[] = [];
+
+afterEach(async () => {
+  if (createdDefinitions.length === 0) return;
+  await withDbAccessContext(SYSTEM_CTX, async () => {
+    for (const id of createdDefinitions) {
+      await db.delete(customFieldDefinitions).where(eq(customFieldDefinitions.id, id));
+    }
+  });
+  createdDefinitions.length = 0;
+});
+
+/**
+ * The shape `middleware/auth` puts on the request.
+ *
+ * `token.mfa` and the `user` row are real requirements, not decoration: the
+ * routes mount the REAL `requireMfa()` and the REAL `requirePermission()`, and
+ * the latter resolves `devices:write` out of the permissions catalog for this
+ * user id. The fixtures below therefore create a genuine user + membership +
+ * role + grant, so the whole middleware chain runs — only the token exchange is
+ * skipped.
+ */
+interface FakeAuth {
+  scope: 'organization' | 'partner' | 'system';
+  partnerId: string | null;
+  orgId: string | null;
+  accessibleOrgIds: string[] | null;
+  partnerOrgAccess: 'all' | 'selected' | null;
+  user: { id: string; email: string };
+  token: { mfa: boolean };
+}
+
+const DEVICES_WRITE = [{ resource: 'devices', action: 'write' }];
+
+async function partnerAuth(
+  partnerId: string,
+  orgIds: string[],
+  orgAccess: 'all' | 'selected',
+): Promise<FakeAuth> {
+  const role = await createRole({ scope: 'partner', partnerId });
+  await grantRolePermissions(role.id, DEVICES_WRITE);
+  const user = await createUser({ partnerId });
+  await assignUserToPartner(user.id, partnerId, role.id, orgAccess);
+  return {
+    scope: 'partner',
+    partnerId,
+    orgId: null,
+    accessibleOrgIds: orgIds,
+    partnerOrgAccess: orgAccess,
+    user: { id: user.id, email: user.email },
+    token: { mfa: true },
+  };
+}
+
+async function orgAuth(partnerId: string, orgId: string): Promise<FakeAuth> {
+  const role = await createRole({ scope: 'organization', orgId });
+  await grantRolePermissions(role.id, DEVICES_WRITE);
+  const user = await createUser({ partnerId, orgId });
+  await assignUserToOrganization(user.id, orgId, role.id);
+  return {
+    scope: 'organization',
+    partnerId,
+    orgId,
+    accessibleOrgIds: [orgId],
+    partnerOrgAccess: null,
+    user: { id: user.id, email: user.email },
+    token: { mfa: true },
+  };
+}
+
+function dbContextFor(auth: FakeAuth): DbAccessContext {
+  return {
+    scope: auth.scope,
+    orgId: auth.orgId,
+    accessibleOrgIds: auth.accessibleOrgIds,
+    accessiblePartnerIds: auth.scope === 'partner' && auth.partnerId ? [auth.partnerId] : [],
+    userId: auth.user?.id ?? null,
+  };
+}
+
+/**
+ * Stand up the real route app with everything but the token exchange intact:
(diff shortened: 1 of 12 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — five agents in parallel: code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer.

**Findings:** 21 raised → 17 addressed in `12fe4140d`, 4 explicitly dismissed. **0 outstanding.**

**Correctness: 0 critical, 0 important.** The code reviewer traced tenant isolation (org-scoped tokens, `partnerOrgAccess: 'selected'`, system scope), preview/commit annotation determinism, the nested-transaction SAVEPOINT mechanism, and route-mount collisions, and found no bug in any of them. The silent-failure hunter confirmed no row can vanish from `created`/`skipped`/`errors` and that no error copy misdirects. Everything below is hardening or coverage.

**Addressed**

| # | From | Finding | Fix |
|---|---|---|---|
| 1 | type-design | `ownerScope`/`organizationId` pairing was comment-only, forcing a `!` at the insert | `CustomFieldDefinitionImportRow` is now a discriminated union; the route's zod schemas became a `z.discriminatedUnion` whose output matches it exactly. The assertion is gone. |
| 2 | type-design | `REFUSED_ANNOTATIONS`/`REFUSAL_COPY` typed `Record<string, …>` | Keyed by `DefinitionAnnotation` — renaming an annotation is now a build error, not a silent fall-through to generic copy. |
| 3 | silent-failure | Audit call unguarded; the rows are already committed and `writeAuditEventAsync` sanitises synchronously, so a throw there would report a successful import as a 500 | Wrapped with a logged catch. Losing an audit event is bad; misreporting a successful import as total failure is worse. |
| 4 | comment | `writeErrors.ts` claimed the web client branches on `field-key-shadowed`/`field-key-duplicate` — nothing in `apps/web` does | Reworded to "reserved for the web client" (the wizard is W09). |
| 5 | test | **MFA denial never proven** — the mock always called `next()`, so dropping `requireMfa()` from a route would have passed every test | Mock now reproduces `hasSatisfiedMfa`. **Mutation-verified**: deleting `requireMfa()` from the routes turns the test red. |
| 6 | test | System scope (`accessibleOrgIds: null`) untested at every layer despite being in `requireScope` | Covered at service and route level, including system scope naming another partner. |
| 7 | test | Empty `accessibleOrgIds` array branch untested | Two service tests. The empty case queues only ONE select result, so if the short-circuit were removed the definitions query would consume the org list and the assertion would fail. |
| 8 | test | INSERT payload never asserted — the summary is built from pre-write variables, so a swapped `orgId`/`partnerId` would not be caught | The mock captures `.values(...)`; both axes and the column defaults are asserted from the payload. |
| 9 | test | `audit.ts` content unproven ("was called" only), though it is the sole durable home of `sourceLabel` | New `audit.test.ts`: per-row `orgId`, `resourceName` fallback, `sourceLabel` presence/absence, `rowCount`, and that skipped rows emit nothing. |
| 10 | test | Soft-deleted orgs untested — `isNull(deletedAt)` is invisible to the mocked db | Integration test with a soft-deleted org; refused at preview and commit. |
| 11 | test / silent-failure | 3 of 4 `writeFailure` SQLSTATEs and the generic fallback untested | Added unmapped SQLSTATE, non-Postgres throw, and 42501. |
| 12 | test | Defensive `insert returned no row` throw unexercised | Covered. |
| 13 | test | No unit test drove two rows through the insert path with one failing | Added — asserts the batch continues and two nested transactions opened. |
| 14 | code-review + silent-failure | **The "never leaks the driver message" integration test never reached the write path** (the row was refused at annotation time), so nothing proved savepoint isolation under a *real* DB refusal | Replaced with a test that makes the snapshot legitimately blind (an org-owned key under an unreachable org) so `annotateRows` says `create` and W03's trigger refuses the actual INSERT. Asserts the good row in the same batch still lands, and that no driver text or `cause` reaches the wire. |
| 15 | test | Integration harness hand-rolled its `DbAccessContext`, omitting `currentPartnerId` — inert today, load-bearing the day #4944 lands | Now uses the canonical `buildDbAccessContext`. |
| 16 | type-design | "never serialize `cause`" was a comment, not a check | Explicit assertion: `JSON.stringify` has no `cause`, `Object.keys` has no `cause`, but the `Error` is still reachable in-process. |
| 17 | type-design | `AnnotatedDefinitionRow`/`CommitDefinitionRowInput` non-union rationale undocumented | Recorded in the types, with the reason each is deliberately NOT discriminated. |

**Dismissed (with reasons, all four recommended as "defer" by the reviewer who raised them)**

- **Discriminated-union `AnnotatedDefinitionRow` on `annotation`** — `existingId`/`existingType` do not correlate 1:1 with the annotation (`type-conflict` arises both with and without an existing definition), so modelling it faithfully would mean inventing wire-visible variants to carry an internal batch-vs-database distinction. Rationale is now in the type.
- **`WeakMap` side-channel for `cause`** — the non-enumerable attachment already defeats every real serialization path, now proven by #16.
- **Unify `field-key-shadowed` / `key-shadowed` naming** — TS narrowing over the literal union already catches a rename; a cosmetic seam with no realized bug.
- **`MAX_IMPORT_VALUES` is unused** — intentional; plan Task 1 places it here and W08 consumes it. Flagged in the PR body so nobody assumes the definitions importer enforces a values cap.

One reviewer noted it created and then deleted a local branch ref (`git branch -D`) despite read-only instructions. Verified afterwards: working tree clean, `HEAD` unchanged and in sync with the remote, no file touched.

**Tests:** `apps/api` affected suites green — 16 files / 262 tests (`vitest run src/services/customFields src/routes/customFieldImport src/routes/customFields src/openapi`), plus `audit.test.ts`. Integration 10/10 against real Postgres (`vitest run --config vitest.integration.config.ts src/__tests__/integration/customFieldDefinitionImport`). `tsc --noEmit` clean. `pnpm lint` clean (6/6).

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #4775 — those are the maintainer's calls.

PR #5044, opened 2026-09-06
feat(api): device_custom_field_values — normalize, project, backfill, register (#3257 W05)

### LINKED ISSUE #4773: W05: device_custom_field_values + projection + backfill + six registration lists
Wave W05 of #4768 — device_custom_field_values + projection + backfill + six registration lists
Branch: `feature/4768-rmm-custom-field-import/wave-4773`

ToddHebebrand:
Closes #4773
Refs #3257 (W05 of the RMM custom-field import plan), feature #4768

## What this does

Device custom-field **values** move out of the `devices.custom_fields` jsonb into a real table, `device_custom_field_values`. `devices.custom_fields` survives as a **trigger-maintained projection**, so the ~34 JS readers, the MCP `SAFE_DEVICE_RESOURCE_FIELDS` projection and both partner-export statement triggers keep working unchanged (Open Decision 1 = A, approved at Gate A — not relitigated). Three SQL consumers are rewritten; nothing else reads differently.

**Four live defects this closes**, each with a test:

| # | Defect | Proof |
|---|---|---|
| 1 | The partner export emitted **two records for one datum** when an org-owned and a partner-wide definition shared a `field_key` (the identity hash includes `f.id`) | `partner export — one record per datum` |
| 2 | `devices.custom_fields` is `excludedOpen` in the tenant-export policy (every jsonb column must be), so every custom-field value was **silently dropped from the GDPR export** | new `CORE_TENANT_EXPORT_POLICY` entry + `tenantExportErasureRoundtrip` |
| 3 | `custom.<key>` filters scanned with `jsonb_extract_path_text` and no index | `filterEngine custom.<key>` tests + `(org_id, field_key, value_text)` index |
| 4 | Deleting a definition **orphaned every value under it, forever** | `cascade-deletes values when their definition is deleted` |

## Red first

The suite was written and run against a database migrated **without** this file: **26 failed / 0 passed**. After the migration: **27/27**.

Defect 1 was additionally proven red against **today's shipped code** before anything was written, driving the real `/custom-field-values` route end-to-end with a forged pre-W03 shadowed pair:

```
AssertionError: expected [ { …(12) }, { …(12) } ] to have a length of 1 but got 2
```

That temporary probe was deleted; the shipped assertion is the same test seeded through the normalized table.

## Migration slot — differs from the plan

The plan assigns `2026-10-09-100000-device-custom-field-values.sql`, which sorts **before** shipped `2026-10-11-150000-ai-partner-wide-select.sql` and would replay out of order on a fresh DB. Shipped as **`2026-10-11-160000-device-custom-field-values.sql`**, re-checked against `origin/main` immediately before pushing.

## The two things the plan got wrong, both found by a test

**1. The projection's "preserve unmanaged keys" rule cannot be "has no visible definition".** The plan phrased it that way; with that rule, `cascade-deletes values when their definition is deleted` fails — the cascade removes the value rows, the now-definition-less key is reclassified as *unmanaged*, and it is **stranded in the jsonb forever**, which is defect 4 reintroduced by the fix for defect 4. The shipped rule is the **key pattern**: a key that cannot match the enforced `^[a-z][a-z0-9_]*$` (`routes/customFields.ts`) can never have a definition and so can never live in the table — those are preserved; every other key is projected from the table alone. This also makes the projection cheaper (no definitions join at all).

**2. The backfill must copy JSON nulls.** A consequence of (1): the plan's `WHERE jsonb_typeof(e.v) <> 'null'` would leave `{"asset_tag": null}` with no row, and the projection would then drop the key entirely — turning an **explicitly cleared** value into "never set" on the first write to any other field on that device. Nulls now land as all-NULL rows, which is exactly what `customFieldValueSchema`'s `z.null()` has always meant. Pinned by an assertion in the backfill test.

## The coherence trigger's merge fence — not in the plan, and load-bearing

The coherence trigger (BEFORE INSERT/UPDATE) rejects a `definition_id` that does not belong to the row's org. That is correct — and on its own it **aborts every org merge** of two orgs that hold custom-field values. The merge's registry walk is parents-first, so it repoints `devices.org_id` **early**, and `breeze_cascade_device_org_id`'s generic loop restamps this table's denormalized `org_id` in the same statement — while the value's definition is still owned by the **loser** org, because `custom_field_definitions`' own executor runs later in the walk.

So the trigger carries an explicit fence, narrow on **four** axes at once so it excuses the merge's own repoint and nothing else:

- `TG_OP = 'UPDATE'` — an INSERT can never reach it, so no caller can create a value under another org's definition;
- `OLD.org_id = def_org` — the row must be moving *out of* the definition's own org, which is exactly the repoint's shape;
- that org must be actively `status = 'merging'`; and
- it must be under the **same partner** as the destination org, so the fence can never widen into cross-partner leakage.

The first two conditions carry the weight rather than the status alone, because the loser org keeps `status='merging'` as a **terminal shell** after the merge. Both halves are pinned: `the merge fence does not excuse an INSERT under a merging org's definition` (P0001), and the end-to-end `executeOrgMerge` test, which would go red if the fence were removed.

Classified in `ORG_ID_BENIGN_TRIGGERS` (`orgMergeRegistry.integration.test.ts`) with that reason. `ORG_ID_CONDITIONALLY_BLOCKING_TRIGGERS` does not fit — it requires a live `blocks-merge` policy to discharge it, and there is none here.

## Cross-org device move — plan Task 3 Step 4, option (a)

An org-owned definition does not travel with a moving device, so the instant `devices.org_id` flips the same generic loop restamps the values and the coherence trigger correctly refuses — a raw P0001 aborting every move of a device carrying org-owned values. New DB function `breeze_rehome_device_custom_field_values(device, target_org)` re-points each value onto the target org's identically-keyed **visible** definition and drops the ones with no counterpart, returning both counts; `moveOrg.ts` calls it inside the move transaction and records the counts in the move audit (`customFieldValues: { rehomed, dropped }`), omitted entirely when both are zero.

It is a DB function rather than inline SQL for two reasons it cannot satisfy from the route:

- it reads `custom_field_definitions` across the org/partner axis, and an org-scoped request context cannot see partner-wide rows (#4944); and
- it must **pre-acquire both orgs' partner-export locks in ascending UUID order** before the projection trigger requests one. The re-home fires the projection, whose `UPDATE devices` requests that device's org lock; `breeze_cascade_device_org_id` later requests **both**. Taking the source alone first would abort the move whenever the target's uuid sorts lower — a coin-flip per move. This is the same pre-acquire, for the same reason, that `breeze_cascade_device_org_id` already does.

Values under **partner-wide** definitions need no re-home inside one partner; a cross-partner move (system scope only) loses that visibility too and drops them by the same rule. The org **merge** deliberately does not call this — there the loser's definitions travel to the survivor wholesale, so dropping anything would be data loss; the trigger fence covers that path instead.

The composite FK is `DEFERRABLE INITIALLY DEFERRED` (not merely DEFERRABLE) precisely so the value may point at the target org before the device row does, within one transaction.

## Six registration lists — every one, with the line

| List | File | Line added |
|---|---|---|
| `CORE_ORG_CASCADE_DELETE_ORDER` | `services/tenantCascade.ts` | `'device_custom_field_values',` (alphabetical, between `device_connections` and `device_disks`) |
| `CORE_DEVICE_CASCADE_DELETE_TABLES` | `routes/devices/core.ts` | `'device_custom_field_values',` (feature-grouped, beside `device_mtls_certificates`) |
| `CORE_DEVICE_ORG_DENORMALIZED_TABLES` | `routes/devices/core.ts` | `'device_custom_field_values',` (alphabetical) |
| `CORE_TENANT_EXPORT_POLICY` | `services/tenantExportPolicyRegistry.ts` | `tablePolicy("org_id", …)` — all 12 columns `included`; **no** json/jsonb/bytea column exists on this table, deliberately |
| `REPOINT_TABLES` | `services/orgMergeRegistry.ts` | `"device_custom_field_values",` — plain repoint; its only unique index `(device_id, definition_id)` cannot collide across orgs |
| `CUSTOM_FIELD_DEFINITION_CHILDREN` | `services/orgMergeCustomExecutors.ts` | `{ table: 'device_custom_field_values', column: 'definition_id' },` |

**`rls-coverage.integration.test.ts` needs no entry** — shape 1 (direct `org_id` with a `breeze_has_org_access(org_id)` policy) is auto-discovered. Not added to `DEVICE_ORG_FK_CASCADE_TABLES` (that list is for tables whose app-role UPDATE is revoked), nor to `AUDIT_ADMIN_REQUIRED_TABLES` (ordinary mutable table).

That last row is the one W02's PR body called out by name: without it, `rehomeChildrenThenDelete`'s DELETE of a deduped definition would cascade away **every stored value under it**. `re-homes values under a deduped definition instead of deleting them` merges two orgs that both define `asset_tag`, plus one key only the loser defines, and asserts both values survive — one under the survivor's `definition_id`, one under its own.

> Note for reviewers: `CORE_TENANT_EXPORT_POLICY` fires on a new **column**, not just a new table. Both export-policy suites and the org-cascade suite only fail under **Integration Tests**, never in **Test API** — a unit-green PR on a stale base can still redden main.

## Scope elevation — in-body, never a function attribute

Both trigger functions and the re-home helper elevate to system scope **in the body** with save/restore and a single return path (reference: `breeze_revalidate_config_policy_feature_references`, `2026-07-27-a-…`). The elegant attribute form `SET "breeze.scope" = 'system'` is superuser-only for a custom dotted GUC and 42501s prod's migration role at `CREATE FUNCTION` time — the v0.97.0 EU crash-loop; `migrationGucAttributes.test.ts` and `Check Migrations (non-superuser)` both guard it. `SECURITY DEFINER` and `SET search_path` are kept (`search_path` is a built-in GUC, available to any role).

**Measured, not assumed.** On this stack, as the non-BYPASSRLS `breeze_app`, the backfill's own detection read:

| | devices with custom-field values |
|---|---|
| unelevated (`breeze_current_scope()` = `none`) | **0** |
| `set_config('breeze.scope','system',true)` | **1** |

That is the silent 0-row no-op the elevation prevents, and it is invisible to CI's superuser *and* to hosted prod's `doadmin` (non-superuser but BYPASSRLS). It bites a self-hosted or future migration role without BYPASSRLS. Because every behavioural test here passes identically with the elevation stripped, the suite also asserts the function's stored **body** from `pg_get_functiondef` — on this stack that is the only assertion that can fail when it is removed, and it additionally pins that the attribute form has not crept back in. Do not delete it as redundant.

The migration also opens with a **blindness probe** that prints the effective scope and the row counts and `RAISE`s if the scope is not `system`, so a "copied 0 values" line cannot be a truthful-looking blind read.

## Write paths

All three shipped writers now go through `persistDeviceCustomFieldValues(deviceId, orgId, writes, source)`:

- `PATCH /devices/:id/custom-fields` (`source: 'api'` for an API key, `'manual'` for a session), then re-reads the projection for the response — the response contract is unchanged;
- `PATCH /devices/:id` (`'manual'`), which no longer puts `customFields` in its own update set at all;
- script write-back (`'script'`).

`validateCustomFieldMap` now also returns `writes: CustomFieldValueWrite[]` — it already holds the matched `VisibleCustomFieldDefinition`, so this avoids a second system DB context per PATCH.

**The compare-before-write moved into the writer**, as a `setWhere` on the upsert, so all three paths get it instead of only script write-back. Its reasoning is unchanged: a write that actually changes a value propagates through the projection to an `UPDATE devices`, which fires `breeze_partner_export_z_custom_values_update` and takes an EXCLUSIVE per-org advisory lock held to COMMIT — a fleet-wide script rewriting unchanged values would serialise the whole org. `setWhere` makes Postgres skip the UPDATE outright rather than making us read-then-write; pinned by `is a no-op when the stored value already equals the incoming one` asserting the export stamp does not move.

`persistDeviceCustomFields` (the whole-jsonb writer) was **removed, not deprecated**: since the projection rebuilds that column, any direct write is reverted by the next value write and bypasses both the composite FK and the coherence trigger. Leaving it exported would have been a live footgun with a silent failure mode.

## What review changed (`/pr-review-toolkit:review-pr`, 4 agents)

Four findings, all confirmed against the code before fixing — two of them were defects I had shipped into the first push:

1. **The backfill minted every orphan key as `type='text'`**, permanently and silently discarding the JSON type. `{"seat_count": 4}` became a text field: typed input and typed comparison gone forever, and no warning telling an operator which fields to re-type. The type is now **inferred**, conservatively — `'number'`/`'boolean'` only when *every* stored value for that `(org, key)` agrees, else `'text'` — and the WARNING breaks the count down by type. Dates are deliberately not inferred (a date is a jsonb *string*; guessing would mistype any string that starts date-shaped). A wrong guess still costs nothing: the copy step falls back to `value_text`.
2. **`breeze_rehome_device_custom_field_values` could abort a device move with a bare 23505.** If a device ever held two value rows under one `field_key`, both re-point to the same target definition and collide on `(device_id, definition_id)` — raised from inside a `SECURITY DEFINER` function, killing the operator's whole move. W02's indexes and W03's trigger make that state unreachable, but they can be disarmed (this suite does it). A `NOT EXISTS` guard now routes the redundant row into the counted `dropped` total instead. **Verified both ways**: without the guard the new test raises exactly `23505 device_custom_field_values_device_def_uq`; with it, the move survives and reports `dropped: 1`.
3. **The `custom.<key>` filter tests were vacuous.** Seeding a value populates the jsonb projection too, so they passed identically whether `filterEngine` read the new table or the old `jsonb_extract_path_text`. They now blank the projection first (the projection trigger fires on writes to the *value* table, not to `devices`, so the blanking sticks), leaving a table read as the only way to match. **Verified**: all three go red when `filterEngine` is reverted.
4. **The GDPR export fix was only pinned structurally.** The registry-completeness check catches a deleted policy entry, but nothing proved the data reaches an archive. `tenantExportErasureRoundtrip` now seeds a real value and asserts it appears in the ZIP with a readable `field_key` and `value_text` — which is the entire reason `field_key` is denormalized.

Also added the coverage review found missing: the backfill's `CASE` ladder across all four types **in both directions** (parse-success into the typed column, parse-failure falling back to `value_text` rather than being dropped), cross-partner moves dropping partner-wide values, and a sibling row staying byte-identical when only one key changes.

**Explicitly dismissed, with reasons:**
- *"`persistDeviceCustomFieldValues`'s return value is unused by all three callers."* Correct, and the reviewer verified it is not exploitable — a genuine failure throws rather than returning empty, and all three callers run inside the request transaction. It is the interface the plan specifies and W08's importer consumes per-key results. Kept, with the contract documented so it is not "simplified" away.
- *"The merge fence's `status='merging'` is a permanent terminal shell."* Real, and deliberately not load-bearing: the fence also requires `TG_OP='UPDATE'` and `OLD.org_id = def_org`. The reviewer independently confirmed `computeOrgReach` already filters `accessibleOrgIds` to `active`/`trial`, so a merged-away org is unreachable to ordinary callers before SQL is involved.
- *"`filterEngine`'s `LIMIT 1` is non-deterministic under duplicates."* True but preferred to the alternative: a subquery cardinality error would abort every smart-group evaluation and deployment-target resolution in the affected org — a fleet-wide outage from one corrupt row. Documented at the call site.

The comment-accuracy pass returned **no inaccurate or stale comments** (one cosmetic wording leftover, fixed). The code review found **no tenant-isolation hole**.

## Deviations from the plan and the brief

1. **Migration slot** `2026-10-11-160000-…` (see above).
2. **Projection rule is the key pattern**, not "has no visible definition" — the plan's rule fails its own defect-4 test (see above).
3. **The backfill copies JSON nulls**; the plan skipped them (see above).
4. **The coherence trigger's merge fence** is not in the plan. Without it every org merge with custom-field values aborts (see above).
5. **The device-move re-home is a `SECURITY DEFINER` DB function**, not inline route SQL — for the partner-wide read and the export-lock ordering (see above). The plan named option (a) and recommended it; this is that, placed where it can actually be correct.
6. **`persistDeviceCustomFields` removed** rather than left in place.
7. **The backfill's `CASE` ladder was restructured.** The plan's sketch grouped as `WHEN A OR B AND C`, which is not the intended precedence and would have mis-routed values; the shipped ladder decides the typed column first and falls through to `value_text`, which is the same intent stated unambiguously. Behaviour is still "a value that does not parse for its declared type falls back to `value_text` rather than being dropped" — this is a data MOVE, not a validation gate.
8. **Three files outside the brief's "files you own" list**, all forced:
   - `routes/devices/moveOrg.ts` — the plan's own W05 Task 3 Step 5 commits this file; without it every cross-org move of a device with org-owned values raises P0001.
   - `__tests__/integration/partnerApiConfigurationWatermark.integration.test.ts` — its five custom-field-value tests seeded `devices.custom_fields` directly, which is no longer how a value is stored. Rewritten to seed the normalized table and (for the move cases) to mirror `moveOrg.ts`'s re-home-then-flip in one transaction. No assertion was weakened; the contracts asserted (both orgs' material clocks are touched by a move; 501 values page correctly; secret-semantics blocking) are unchanged.
   - The co-located unit tests for the five changed modules.
9. **`scriptWriteBack`'s post-write `device_not_found` branch is gone.** It existed because `persistDeviceCustomFields` returned a boolean "the UPDATE matched no row". The upsert has no such signal, and the composite FK is `ON DELETE CASCADE`, so a device deleted concurrently takes its values with it rather than erroring. The `device_not_found` rejection for the *load* returning null is unchanged and still tested.
10. **`filterEngine`'s `custom.` branch is a correlated subquery** rather than a join, so `getColumnForField`'s return shape and every caller (`groups.ts`, `groupMembership.ts`, `deploymentTargetResolver.ts`, `filters.ts`) are untouched.

## Verification — all foreground, live DB via `pnpm test-stack`

| Check | Result |
|---|---|
| `deviceCustomFieldValues.integration.test.ts` | **33/33** (was 26/26 RED before the migration) |
| `customFieldShadowing` + `customFieldDefinitionIntegrity` + `customFieldDefinitionsPartnerRls` + `customFieldDefinitionsMerge` + `orgLifecycleFoundations` + `orgMergeRegistry` + `tenantCascade` + `tenant-export-policy` + `tenantExportErasureRoundtrip` | **13 files / 95 tests pass** |
| All of the above plus `orgMerge*` + `deviceMoveOrg*` + `partnerApiConfigurationWatermark` + `scriptCustomFieldWriteBack` + `deviceCustomFieldImportIsolation` | **26 files / 212 tests pass** |
| Targeted unit sweep (`routes/devices`, `services/customFields`, `filterEngine`, `routes/partnerApi`, `orgMerge*`, `groupMembership`, `deploymentTargetResolver`, `routes/groups`, migration guards) | **90 files / 3020 tests pass** |
| `vitest.config.rls-coverage.ts` | **102 pass** |
| `vitest.config.rls.ts` | **26 pass, 5 skipped** |
| `src/db/autoMigrate` + `migrationRlsScope` + `migrationGucAttributes` + `services/tenantCascade` + `routes/devices/cascadeDelete` + `routes/devices/moveOrg.coverage` | **7 files / 1568 tests pass** |
| `pnpm db:check-drift` | no drift, 687 files match the ledger |
| `check:migrations:nonsuperuser` (fresh postgres:16, `breeze_migrator` NOSUPERUSER) | **OK — full migration set applied as a non-superuser role** |
| `scripts/check-migration-naming.sh` (pre-commit + pre-push, `--against-ref origin/main`) | OK |
| migration apply, against a seeded DB | probe printed `effective_scope=system, devices=1, devices_with_values=1, definitions=2`; backfill `minted 0 definition(s)` / `0 stored key(s) do not match the enforced key pattern` / `copied 2 value(s) out of devices.custom_fields`. The non-trivial backfill (mint + copy + JSON-null round-trip + camelCase preservation) is exercised by the suite's own migration replay, which also asserts **re-apply is a true no-op** — no duplicate rows, no duplicate mint. |

## Not verified

- **No production preflight and nothing here SSHes anywhere.** The backfill is bounded by device count × keys per device (one jsonb per device, not a telemetry table), so it needs no `ctid`-batching; but its real-world row count on either prod region has **not** been measured from this branch. The `RAISE WARNING` counts are the operator's evidence at deploy time.
- **No `apps/web` changes.** The `PATCH /devices/:id/custom-fields` request/response contract is unchanged and `DeviceInfoTab.tsx` was not exercised in a browser from this branch.
- **The full `pnpm --filter @breeze/api test` suite was not run end to end** — the targeted paths in the brief were, plus the device/filter/partner-API units.
- **The `(org_id, field_key, value_text)` index is a structural claim, not a measured one.** No `EXPLAIN` was run at production scale; the previous shape had no index at all, so this is strictly better, but the magnitude is unquantified.
- **Cross-partner device moves** (system scope only) drop values under formerly-visible partner-wide definitions. That follows the same "not visible in the target org" rule as org-owned definitions, is reported in the audit, and is now pinned by a test — but it is a behaviour a human should sanity-check before the first such move in production.
- **Concurrency on the upsert is untested.** Two simultaneous writers to the same `(device_id, definition_id)` are serialised by the unique index and `ON CONFLICT`, but no test exercises it. The pre-existing lost-update characteristic on that pair is unchanged from before this wave and is documented at the call site.
- **The script write-back's `device_not_found` disposition narrowed.** A device deleted between the load and the write used to surface a typed per-field rejection on `script_executions.custom_field_result`; it now raises 23503, which `commandResultHandlers.ts` catches, logs and reports to Sentry while discarding the summary. Loud to engineering, quieter to the operator. Documented in place with the one-line fix if it ever needs re-typing.

https://claude.ai/code/session_01UXLedCzGFmmKvCsbxqvMTh

---

## CI-fix round (`Integration Tests (shard 2/4)`)

`partnerApiRls › cursor-walks every resource through actual auth without crossing partners`
went red on the first push: **`expected [] to have a length of 4 but got +0`** for
`'custom-field-values': 4`.

**Not a defect in the projection — a test seeding through a write path this wave removed.**
`devices.custom_fields` is now a one-way, trigger-maintained projection of
`device_custom_field_values`, and `customFieldValueSource` reads the table. Three
integration seeds wrote the jsonb literal straight onto a device insert, so the
table-backed export saw nothing. The projection direction is unchanged — plan
Decision 1 stays one-way, and the "no legacy writer" guarantee depends on it.

### Sweep — every jsonb seed feeding a table-backed consumer

`git grep -n "customFields: {" -- 'apps/api/src/**/*.test.ts' 'apps/api/src/__tests__'`
plus `git grep -n "custom_fields" -- apps/api/src/__tests__`. Three sites, two files:

| File | Site | What went blind | Fix |
|---|---|---|---|
| `partnerApiRls` | `seedPartnerOrg` device insert | `/custom-field-values` | definition insert now `.returning()`s; the org-owned value is inserted into `device_custom_field_values`; the device insert carries no `customFields` at all |
| `partnerApiRls` | the test-local `update(devices).set({customFields})` loop | `/custom-field-values` | the partner-wide `partner_inventory_label` values are inserted as rows |
| `partnerApiConfigurationWatermark` | seven-routes test device insert | `/custom-field-values` — its `every(…orgId === org.id)` check is vacuously true on `[]` | value row seeded; non-empty guard added, scoped to that one route |

Converted one more that was not blind but was dishonest: *"device custom-field value
changes advance custom-field state"* wrote the jsonb itself. It now writes the value
row and lets the projection produce the `UPDATE devices` the watermark trigger keys
on — the chain a shipped writer actually takes.

The `every`-on-`[]` guard is deliberately **not** widened to the other six routes.
Some of them are empty under that fixture (`/backup-configurations` is never seeded
there), but that predates this wave: W05 touched only `customFieldValueSource` in
`routes/partnerApi/configuration.ts` — every other source query is untouched — so it
is not this PR's to change.

Unit suites that only assert on the jsonb *shape* keep their jsonb fixtures; the
column and its ~34 JS readers are unchanged.

### Two new controls

**1. The projection is what rebuilds the jsonb.** `partnerApiRls` now seeds *only*
the table (zero `customFields` writes remain in the file), so the added assertion
that `devices.custom_fields` equals `{ rack_a_1, partner_inventory_label }` can only
pass if the trigger fired. Without it, a regression that stopped maintaining the
jsonb would blind every JS reader with nothing red.

**2. The export ignores the jsonb — the control that can actually rot.** Every other
test seeds the row *and*, via the projection, the jsonb, so all of them stay green if
someone re-adds a jsonb fallback to `customFieldValueSource` to paper over a backfill
gap — which would re-open defect 1 (one datum exported twice when an org-owned and a
partner-wide definition share a `field_key`). `deviceCustomFieldValues` has this
control for `filterEngine`; the export route had none.
`the export ignores a value that exists only in the devices.custom_fields jsonb`
forges the pre-W05 state with a raw UPDATE, asserts `/custom-field-values` returns
`[]`, then writes the same datum as a ROW and asserts it *is* exported — the second
half is what stops the first from passing vacuously.
**Verified red**: with a `UNION ALL` jsonb-fallback branch spliced into
`customFieldValueSource`, that test and only that test fails
(`expected [ {…} ] to deeply equal []`; 17 others still pass). Route restored.

### Also in this round

- Merged `origin/main` (W07 #5043 — `routes/customFields.ts`,
  `services/customFields/writeErrors.ts`, `services/customFields/import/*`,
  `index.ts`, `openapi.ts`). No conflicts; `tsc` clean after.
- Re-checked the migration slot against `origin/main`:
  `2026-10-11-160000-device-custom-field-values.sql` still sorts after
  `2026-10-11-150000-ai-partner-wide-select.sql`. No rename needed.
- Delta review (`code-reviewer` + `pr-test-analyzer`): 0 + 2 findings, both fixed
  in `2eb05d87f` (the new export control above, and dropping a stray
  `backupConfigs` seed that changed an unrelated route's fixture with no assertion
  behind it).


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer (4 agents, parallel, all read `gh pr diff` rather than a live worktree).

**Findings:** 4 raised → **4 addressed** in 2ffc6378d; 3 explicitly dismissed with reasons (below); 0 outstanding.

Two of the four were defects I had shipped into the first push, not nits:

1. **Backfill discarded the JSON type.** Every orphan key was minted `type='text'`, so `{"seat_count": 4}` became a text field — typed input and typed comparison gone permanently, with no warning naming the affected fields. The minted type is now **inferred**, conservatively (`'number'`/`'boolean'` only when every stored value for that `(org, key)` agrees, else `'text'`; dates deliberately not inferred, since a date is a jsonb *string*). The WARNING now breaks the count down by type. A wrong guess costs nothing — the copy step falls back to `value_text`.
2. **`breeze_rehome_device_custom_field_values` could abort a device move with a bare 23505.** If a device ever held two value rows under one `field_key`, both re-point to the same target definition and collide on `(device_id, definition_id)`, raised from inside a `SECURITY DEFINER` function. W02's indexes + W03's trigger make that unreachable, but they can be disarmed. A `NOT EXISTS` guard now routes the redundant row into the counted `dropped` total. **Verified both directions:** without the guard the new test raises exactly `23505 device_custom_field_values_device_def_uq`; with it the move survives and reports `dropped: 1`.
3. **The `custom.<key>` filter tests were vacuous.** Seeding a value populates the jsonb projection too, so they passed identically whether `filterEngine` read the new table or the old `jsonb_extract_path_text`. They now blank the projection first, leaving a table read as the only way to match. **Verified:** all three go red when `filterEngine` is reverted.
4. **The GDPR export fix was only pinned structurally.** `tenantExportErasureRoundtrip` now seeds a real value and asserts it reaches the ZIP with a readable `field_key` and `value_text` — the entire reason `field_key` is denormalized.

Also added coverage the test analyst found missing: the backfill's `CASE` ladder across all four types **in both directions**, cross-partner moves dropping partner-wide values, and a sibling row staying byte-identical when one key changes.

**Dismissed, with reasons:**
- *Unused return value of `persistDeviceCustomFieldValues`* — the reviewer verified it is not exploitable (a real failure throws; all callers share the request transaction). It is the plan's interface and W08's importer consumes it. Kept, contract documented so it is not "simplified" away.
- *The merge fence's `status='merging'` is a permanent terminal shell* — real, and deliberately not load-bearing: the fence also requires `TG_OP='UPDATE'` **and** `OLD.org_id = def_org`. The reviewer independently confirmed `computeOrgReach` already filters `accessibleOrgIds` to `active`/`trial`.
- *`filterEngine`'s `LIMIT 1` is non-deterministic under duplicates* — preferred to the alternative: a subquery cardinality error would abort every smart-group evaluation and deployment-target resolution in that org. Documented at the call site.

The comment pass returned **no inaccurate or stale comments** (one cosmetic wording leftover, fixed). The code review found **no tenant-isolation hole**.

**Tests (all foreground, live DB):**
- `deviceCustomFieldValues.integration.test.ts` **33/33** — was **26/26 RED** against a DB migrated without this file. Defect 1 was separately proven red against *shipped* code first: `expected [ {…}, {…} ] to have a length of 1 but got 2`.
- Contract battery (`customField*`, `orgLifecycleFoundations`, `orgMerge*`, `tenantCascade`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, `deviceMoveOrg*`, `partnerApiConfigurationWatermark`, `scriptCustomFieldWriteBack`, `deviceCustomFieldImportIsolation`) — **26 files / 212 pass**
- `vitest.config.rls-coverage.ts` **102 pass** · `vitest.config.rls.ts` **26 pass, 5 skipped**
- Targeted unit sweep — **90 files / 3020 pass** · `tsc --noEmit` clean · `pnpm lint` 6/6 · `db:check-drift` no drift (687 files)
- `check:migrations:nonsuperuser` against a fresh `postgres:16` as a NOSUPERUSER role — **OK**, and green in CI
- Migration re-apply is a true no-op (asserted in-suite: no duplicate rows, no duplicate mint)

**Status:** review-clean, awaiting maintainer merge. Three things worth a human decision before merging are called out in the PR body's *Not verified* section — most notably that cross-**partner** device moves now drop values whose partner-wide definitions stop being visible (counted in the move audit, pinned by a test, but new behaviour).

ToddHebebrand:
## CI-fix round — `Integration Tests (shard 2/4)` (run 34013236297)

**Failure:** `partnerApiRls › cursor-walks every resource through actual auth without crossing partners` — `AssertionError: expected [] to have a length of 4 but got +0` (`'custom-field-values': 4`).

**Root cause: a test seeding through a write path this wave removed, not a defect in the projection.** `devices.custom_fields` is a one-way, trigger-maintained projection of `device_custom_field_values`, and `customFieldValueSource` reads the table — so a jsonb-only seed is invisible to `/custom-field-values`. The projection direction is unchanged (plan Decision 1); the "no legacy writer" guarantee depends on it.

**Sweep** (`git grep -n "customFields: {" -- 'apps/api/src/**/*.test.ts' 'apps/api/src/__tests__'` + `git grep -n "custom_fields" -- apps/api/src/__tests__`) found three such sites in two files, all converted to seed the row instead of the jsonb: `seedPartnerOrg`'s device insert and the test-local `update(devices).set({customFields})` loop in `partnerApiRls`, and the seven-routes device insert in `partnerApiConfigurationWatermark` (whose `every(…orgId === org.id)` check was vacuously true on the `[]` it produced). One more was converted for honesty rather than redness: *"device custom-field value changes advance custom-field state"* wrote the jsonb itself and now writes the value row, letting the projection produce the `UPDATE devices` the watermark trigger keys on. Unit suites that only assert on the jsonb shape keep their jsonb fixtures.

### Review run

`/pr-review-toolkit:review-pr` on the delta — `code-reviewer` + `pr-test-analyzer`, both reading `gh pr diff` / the commits API rather than a live worktree.

- **code-reviewer: 0 findings.** It independently confirmed the new projection assertion discriminates (nothing writes the jsonb in that file any more, so only `breeze_device_custom_field_project` can populate it), that `EXPECTED_COUNTS` 4/4 still follow from the seeding rather than luck, that no other test relied on the device arriving with a populated jsonb, and that the triggers are plain (non-deferred) statement triggers so there is no commit-timing race with the following `SELECT`.
- **pr-test-analyzer: 2 findings, both fixed in `2eb05d87f`.**

**Finding 1 — missing control, and the one that can actually rot.** Every existing test seeds the value row *and*, through the projection, the jsonb. So all of them stay green if someone re-adds a jsonb fallback to `customFieldValueSource` to paper over a backfill gap — which would re-open defect 1 (one datum exported twice when an org-owned and a partner-wide definition share a `field_key`). `deviceCustomFieldValues` has this control for `filterEngine`; the export route had none.

`the export ignores a value that exists only in the devices.custom_fields jsonb` forges the pre-W05 state with a raw UPDATE (no code path can produce it any more), asserts `/custom-field-values` returns `[]`, then writes the same datum as a ROW and asserts it *is* exported — the second half is what stops the first from passing vacuously if the fixture never reached the route.

**Verified red both ways:** with a `UNION ALL` jsonb-fallback branch spliced into `customFieldValueSource`, that test and *only* that test fails — `expected [ {…} ] to deeply equal []`, 17 others still pass. Route restored (`git diff HEAD -- apps/api/src/routes/` empty).

**Finding 2 — a stray seed.** The first CI-fix commit added a `backupConfigs` insert to the seven-routes test, left over from a repo-wide non-empty guard I then scoped down to `/custom-field-values`. It changed an unrelated route's fixture with no assertion behind it and made the accompanying "the other six are left as they were" comment untrue. Removed; the comment now names a route that genuinely is empty there (`/backup-configurations`, never seeded by that test) and states what was actually checked — W05 touched only `customFieldValueSource`, so that emptiness predates this wave.

### Verification (all foreground, live DB via `test-stack` + `db:migrate`)

- `partnerApiRls` **13/13** · `partnerApiConfigurationWatermark` **18/18** · `deviceCustomFieldValues` + `tenantExportErasureRoundtrip` **36/36**
- custom-field battery (`customField*`, `custom-fields-rls`, `deviceCustomFieldImportIsolation`, `scriptCustomFieldWriteBack`) — **8 files / 91 pass**
- cascade + export-policy + merge battery (`tenantCascade`, `tenant-export-policy`, `orgMerge*`, `orgLifecycleFoundations`) — **12 files / 71 pass**
- `deviceMoveOrg*`, `deviceLifecycle`, `partnerApi*` — **12 files / 91 pass**
- `vitest.config.rls-coverage.ts` **102 pass**
- targeted unit sweep (`routes/devices`, `routes/customField*`, `services/customFields`, `db/autoMigrate`) — **66 files / 1065 pass**
- `tsc --noEmit` clean · `pnpm lint` 6/6

**Base:** merged `origin/main` (W07 #5043 — `routes/customFields.ts`, `services/customFields/writeErrors.ts`, `services/customFields/import/*`, `index.ts`, `openapi.ts`). No conflicts, `tsc` clean after. Migration slot re-checked against `origin/main`: `2026-10-11-160000-device-custom-field-values.sql` still sorts after `2026-10-11-150000-ai-partner-wide-select.sql`, so no rename.

**Status:** review-clean, awaiting maintainer merge. The *Not verified* items in the PR body still stand — most notably that cross-**partner** device moves now drop values whose partner-wide definitions stop being visible (counted in the move audit, pinned by a test, but new behaviour).

PR #5045, opened 2026-09-06
fix(agent): bound collector command capture, clear shipped log batches, count PIDs cheaply

ToddHebebrand:
## Summary

Agent memory follow-up from the 2026-09-05 Windows agent memory review.

- **Collector command runner**: `ErrWaitDelay` after a clean exit (child exited 0 but a grandchild still held the stdout pipe past `WaitDelay`) now logs a warning and returns the captured output as success, via a shared `runCollectorCommand` helper. Previously the collector treated this as a failure and dropped good output. The normalisation is race-free per the stdlib contract: `ErrWaitDelay` is only returned when the process itself exited successfully, so it can never mask a non-zero exit or a context cancellation.
- Dropped `collectorStderrBuffer` (no consumer). Stderr is now a 64 KiB `cappedBuffer` surfaced on `ExitError.Stderr`.
- **Log shipper**: clear the batch slice after a successful ship so shipped entries are released instead of pinned until the next fill.
- **Metrics**: count PIDs without materialising per-process structs.
- Stale comments fixed in `command_limits.go` and `eventlogs_darwin.go`; `cappedBuffer` doc mentions `exceeded`.
- Windows CI gains a step running the collector command-runner tests natively.
- Research doc updated with the new semantics, the +10s worst-case budget, and the ProcessCount step-up note.

## Tests

- `command_limits_test.go`: red-first test rewritten to assert success plus preserved output on `ErrWaitDelay`.
- `shipper_test.go`: full-batch test added; verified it discriminates (fails when `clear` is moved before `shipBatch`, passes restored).
- Verified locally on top of current main: `gofmt`, `go vet` on darwin/linux/windows, `CGO_ENABLED=0 go test -race` green for `internal/collectors` and `internal/logging`.

https://claude.ai/code/session_01QBQmjgnokbMJKGXtuJAbHY

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1432,6 +1432,19 @@ jobs:
           $passes = @($out | Select-String -Pattern '^--- PASS: Test').Count
           if ($passes -ne 3) { throw "expected exactly 3 top-level PASS lines, got $passes - a test in the -run filter was renamed, removed, or added without updating this count" }
 
+      # internal/collectors as a whole is not on the list above (it has never
+      # been vetted for Windows reds), but its command runners set WaitDelay so
+      # a descendant that inherits PowerShell's stdout handle cannot hang a
+      # collector forever — and handle inheritance is the one behaviour with no
+      # cross-platform equivalent. Run just the self-exec subprocess tests that
+      # cover capture bounds, cancellation, and inherited pipes on the OS they
+      # were written for.
+      - name: Run Windows collector command-runner tests
+        working-directory: agent
+        env:
+          CGO_ENABLED: "0"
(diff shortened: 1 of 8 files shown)


ToddHebebrand:
**Review run:** independent code-reviewer pass on the full diff at `fe9bc70e0f` (ErrWaitDelay normalisation traced against os/exec Wait() source; shipper clear; PID count; cappedBuffer). **No blocking findings.** Two non-blocking notes for a follow-up: `runCollectorBoundedOutput` still drains StdoutPipe before Wait() (same descendant-pipe hazard, only caller is darwin `log show`); the `runCollectorCommand` doc comment overstates the benefit for the `change_tracker.go` caller (its 8s `collectWithTimeout` still returns before WaitDelay fires; the real win is the goroutine/buffer no longer leaks).

PR #5053, opened 2026-09-06
feat: purge removed devices after N days — device_lifecycle config-policy feature + daily job (#2787 item 4)

ToddHebebrand:
## What

An MSP can now set, per partner or per org, **"permanently delete removed devices N days after removal"**, and a daily job does it through the same hardened purge path as the Permanently Delete button. This is #2787 item 4 — the thing the original reporter actually wanted.

New `device_lifecycle` config-policy feature (Pattern B, pure JSONB inline settings) with `{ purgeRemovedAfterDays: number | null }`, a new `devices.decommissioned_at` column, an org-level resolver, a daily BullMQ sweep, and a web tab.

## Design rationale

**Why a config-policy feature rather than an org setting.** Breeze is an MSP tool: a tech defines the retention rule once and it applies to every customer, with individual orgs able to override. That is exactly what the config-policy hierarchy already does. Partner-wide ownership is not an afterthought here — the primary shape is one partner-level policy, and an org-level link is the exception. An org-level link storing `null` is how a single customer opts *out* of the MSP's fleet-wide window, which falls straight out of closest-wins.

**Why Pattern B (inline JSONB, no normalized table).** One number and an on/off. No independent lifecycle, nothing to share across policies, no compliance surface. Mirrors `pam` / `vulnerability` end to end.

**Why a new column.** Nothing in the schema recorded *when* a device was removed. `status` flips to `decommissioned` and `updated_at` is then touched by every unrelated write afterwards — a retention window built on it silently extends itself every time anything writes the row. `decommissioned_at` is genuinely new state, cleared again on Restore.

**Fail closed, everywhere.** This job permanently deletes customer data with no human in the loop, so every ambiguous outcome resolves to "never purge":

- No `device_lifecycle` policy for the org → the eligibility query is **never even issued** (asserted, not just "purged nothing").
- Policy lookup throws → that org is skipped for the whole run. There is no default window to fall back to.
- Winning link says `null`, says nothing, or stores a value outside the validator's own 1..3650 integer bound → `null`.
- `decommissioned_at IS NULL` → never eligible, explicitly, not just by `NULL < cutoff` being `NULL`.
- The validator's floor is **1, not 0**: `0` would read as "delete every removed device on the next run", so it is not expressible in the API or the UI.

**Why it calls `purgeRemovedDevice` rather than deleting.** It inherits the lock-first status re-check (a device restored between the job's SELECT and its lock is refused, not deleted under a stale read) and the `UNINSTALL_PENDING` refusal (purging while a `device_remove` agent uninstall is still collectable would destroy the only thing that will ever clean the endpoint). Both are counted and skipped, never re-implemented.

**One transaction per device.** `deleteDeviceCascade` touches ~40 tables; a shared transaction would hold those locks across the whole sweep and let one bad row roll back every successful deletion before it. A 200-device per-org-per-run cap keeps the first night after enabling the feature bounded, and a capped run reports `incomplete` so a backlog is visible rather than silently indefinite.

**One audit row per deletion.** The `devices` row is gone afterwards, so the `device.permanent_delete` audit entry is the only durable record. It carries `retentionPolicy: true`, the window that authorised it, and the removal timestamp, so an operator can tell a policy-driven deletion from a human one.

## Migration

`2026-10-11-160000-device-lifecycle-feature-and-decommissioned-at.sql` — enum value, `ADD COLUMN IF NOT EXISTS decommissioned_at timestamptz`, a one-off backfill from `updated_at`, and a partial `(org_id, decommissioned_at) WHERE status = 'decommissioned'` index. Fully idempotent (verified by replaying the file against a live DB: three `already exists, skipping` notices, exit 0).

The backfill runs under `SET LOCAL breeze.scope = 'system'` inside its `DO` block: `devices` is forced-RLS and migrations run as an unprivileged role on managed Postgres, where a context-less UPDATE is a silent 0-row no-op that CI's superuser masks. Row count is reported with `GET DIAGNOSTICS` + `RAISE WARNING`.

`devices` is in `CORE_ORG_CASCADE_DELETE_ORDER`, so the new column also had to be classified in `CORE_TENANT_EXPORT_POLICY` (`included` — a plain timestamp). No new table, so no RLS or cascade registration.

## Deviations from the plan

1. **Swept three more decommission call sites** (`jobs/offlineDetector.ts` uninstall-intent reaper, `routes/agents/mtls.ts` deny, `services/quickSupportEnd.ts`) in addition to the two the plan named. The migration's backfill already makes *every* pre-existing removed device purge-eligible regardless of how it got there; not stamping these would have shipped the split where a device auto-reaped before the migration is purged by the org's policy and an identical one reaped after it is exempt forever. `offlineDetector`'s own test already demanded "field set matches the admin decommission route EXACTLY".
2. **Schedule slot `13 8 * * *`, not the plan's `17 8 * * *`.** Minute 17 is in the sub-daily lane (≡2 mod 5) and `audit-drift-evaluator` already fires hourly at `:17` — that slot collides every day at 08:17. `:13` was free in hour 8's daily lane.
3. **Worker placement `socket-owner`, not the plan's expected `global`.** The closure contract test is the mechanical authority and `deviceLifecycle -> deviceDeletion` reaches `routes/agentWs.ts` — the same chain that already puts `deviceBulkPurge` in that class. Not re-litigated by judgement; the test was run.
4. **`actorId: ANONYMOUS_ACTOR_ID` + `details.job`, not `actorId: 'removed-device-purge'`.** `audit_logs.actor_id` is `uuid NOT NULL`; the plan's literal fails `string_to_uuid`. Found by the integration run, not by review. Same convention every other scheduled writer uses.
5. **The resolver reads the whole `inlineSettings` blob** and re-validates it with the shared schema, rather than the plan's `inlineSettings->>'purgeRemovedAfterDays'`. Same result, and it means a corrupt stored value is rejected by the same rule that governs writes instead of being coerced by SQL.
6. **The TS feature-type enum landed with the registries (Task 3) rather than with the migration (Task 1)** so every commit compiles on its own. The DB enum value still ships in the Task 1 migration.
7. **Two extra summary fields** (`orgsFailed`, `orgsCapped`) beyond the plan's signature, so a run where every policy lookup threw cannot report as a clean drain.

## Tests

**Unit — 50 new, all run and green:**

- `services/deviceLifecyclePolicy.test.ts` (16) — compiled `WHERE`/`JOIN ON` via `PgDialect` for the predicate shape (active-only, `feature_type = 'device_lifecycle'`, both assignment axes, the partner-wide `org_id IS NULL` ownership arm), plus scripted rows for org-beats-partner precedence, the assignment-priority tiebreak, and a six-case table of corrupt stored values.
- `jobs/removedDevicePurge.test.ts` (16) — includes compiled-SQL assertions on the eligibility predicate (org, status, `IS NOT NULL` stamp, and the exact cutoff instant `2026-01-30T00:00:00.000Z` for a 30-day window from a fixed `now`), and that the registered `Worker` actually runs the sweep rather than a no-op.
- `packages/shared` validator (7), `DeviceLifecycleTab.test.tsx` (11).
- Updated: `core.decommission.test.ts`, `deviceLifecycle.test.ts`, `offlineDetector_uninstallIntentReaper.test.ts`, `quickSupportEnd.test.ts`, `mtls.test.ts`, `tenantExportPolicy.test.ts`, `autoMigrate.test.ts`.

**Integration — 7 new, and they RAN against real Postgres** (`pnpm test-stack up`, `vitest run --config vitest.integration.config.ts`; no `runIf`, so a missing database fails loudly):

`removedDevicePurge.integration.test.ts` — (1) a partner-wide 7-day window purges a 10-day-old removed device, spares a 3-day-old one in the same org, **and spares an equally-old device under a different partner with no policy**; (2) one system-actor audit row per deletion carrying the authorising window; (3) an org-level 30-day window outranking the partner's 7-day one; (4) an org-level `null` opting that org out entirely; (5) a `NULL` stamp never purged, with a stamped sibling as the positive control so "nothing happened" cannot satisfy the assertion; (6) a device with a queued `device_remove` uninstall skipped, its command row intact, and its eligible neighbour still purged; (7) the migration's backfill statement replayed against an unstamped removed row.

**Other suites run against the live database and green:** `deviceLifecycle`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, `configPolicyPartnerLevelAssignments`, `tenantCascade` (39 tests), and the RLS coverage contract suite (`vitest.config.rls-coverage.ts`, 102 tests).

**Broader targeted runs:** API `src/routes/devices src/services/configurationPolicy* src/services/deviceLifecycle* policyBaselineDefaults autoMigrate offlineDetector quickSupportEnd mtls routes/configurationPolicies scheduleRegistry retentionMetrics aiToolsConfigPolicy` — 81 files / 1317 tests green. `packages/shared` validators + constants — 58 files / 1831 green. Web `configurationPolicies`, `DeviceEffectiveConfigTab`, `lib/i18n` — 40 files / 450 green (localeParity, keyUsage and translationCoverage included; tr-TR's exact-English duplicate baseline would have caught untranslated placeholders).

**Typecheck:** `apps/api` and `apps/web` clean. `packages/shared` reports 3 pre-existing errors in `src/validators/quotes.test.ts` — **verified unchanged by this branch** (`git diff origin/main -- packages/shared/src/validators/quotes.test.ts` is empty; the file last changed in 9cb803256 on main).

**Lint:** clean across every touched file. **`pnpm db:check-drift`:** clean, 687 migration files match the ledger. **`scripts/check-migration-naming.sh`:** clean, including `--against-ref origin/main`.

## Merge note

`origin/main` was merged in. Both branches added a worker and both bumped the same `125` literal, so git auto-merged the two additions into a 126-entry registry while every count assertion still read 125 — caught and reconciled in its own commit.

Refs #2787 #5023


ToddHebebrand:
Review findings addressed. Head is now `8a459c499637ed2f3a8bfa88a26ddfa2e579ea61`, two commits.

### 1. `purgeOne` only re-checked `status` — org and eligibility were not re-verified under the lock (`cde5adbb6`)

Confirmed, and worse than a theoretical race: the candidate SELECT is unlocked and its rows are purged **sequentially, up to 200 per org**, so minutes can pass between "eligible" and "deleting". Two things change in that window while `status` stays `decommissioned`:

- **move-org** — the device now belongs to a different tenant whose policy may say nothing about purging. It was deleted anyway, and the audit row (the only surviving record) was written against the **old** org.
- **restore then re-remove** — `decommissioned_at` is refreshed to now, so the device is freshly removed and nowhere near the window; the status re-check waved it through.

Fixed exactly as `jobs/deviceBulkPurge.ts:127-135` does it: `SELECT org_id, decommissioned_at FROM devices WHERE id = $1 FOR UPDATE` inside the per-device transaction, before `purgeRemovedDevice`. Skips as `skippedRaced` on `org_id <> orgId`, `decommissioned_at IS NULL`, or `decommissioned_at >= cutoff`. Two details worth naming:

- The cutoff passed in is the **same one the candidate query used**, not a freshly computed one — otherwise a long run would widen its own window as it goes.
- `DeviceLifecycleError` is still caught *outside* `db.transaction`, so it still rolls back exactly as before; the new skip reasons return normally (nothing was written, only the lock taken).

Module header updated to say plainly that the inherited status re-check is not sufficient on its own.

`purgeOneRemovedDevice` is now exported so the re-check can be proven with a deliberately stale cutoff rather than by trying to win a real race.

**Tests — 9 new unit (red first), 3 new integration against real Postgres:** org changed, re-removed inside the window, stamp cleared, row vanished under the lock, and an assertion that the `FOR UPDATE` statement actually names `org_id` and `decommissioned_at` (so deleting the re-check fails a test rather than leaving the others satisfiable by scripted defaults). Each refusal has a positive control in the same test. The unit tx double now serves the locked row from the scripted device by default, so the ordinary path exercises the real re-check instead of bypassing it. The integration cases additionally prove the real `timestamptz` round-trip in **both** directions — a fresh stamp refuses, an old one purges — which the fake tx cannot.

### 2. `device_lifecycle` missing from the inline-only branch of `validateFeaturePolicyExists` (`8a459c499`)

Confirmed, and the failing test reproduced both halves before the fix: for an org-scoped policy it returned `valid: true` (whole-policy linking accepts any id that names a configuration policy in the same org → the write proceeds → `config_policy_feature_links_reference_integrity` rejects it as a 500), and for a partner-wide policy it returned the misleading *"this feature policy is organization-scoped"* error. Added `device_lifecycle` to the condition, plus a comment recording why absence from that list is not harmless.

3 unit tests. The db double deliberately returns a **matching** configuration-policy row, so removing the guard makes the fall-through *succeed* rather than crash — the test then fails on the assertion instead of on an incidental TypeError.

### 3. Nit: re-export placement

`export { deviceLifecycleInlineSettingsSchema }` had landed between `vulnerabilityInlineSettingsSchema`'s doc comment and its declaration. Moved up next to the `pam` schema it mirrors. No behaviour change.

### Re-verification

API unit sweep 84 files / 1450 tests green. Integration (`removedDevicePurge`, `deviceLifecycle`, `tenant-export-policy`, `tenantExportErasureRoundtrip`, `configPolicyPartnerLevelAssignments`) 5 files / 37 tests green **against a live Postgres**. `tsc` clean for `apps/api` (8 GB flag) and `apps/web`. Lint clean on every changed file. Test stack torn down; `docker compose ls -a` shows no project for this worktree.

ToddHebebrand:
Test API red fixed. Head is now `9ecd096e80be2cc092b3d3eed17b36938d375051`.

**Cause.** `src/db/migrationRlsScope.ts` recognises only the `SELECT`/`PERFORM set_config('breeze.scope','system', …)` forms. The backfill used the equivalent `SET LOCAL breeze.scope = 'system'` inside its `DO` block, so the guard reported the file as a migration that writes rows without electing system scope. Switched to the canonical top-level form, matching `2026-09-30-100000-rls-scoped-backfill-replay.sql`. `GET DIAGNOSTICS` + `RAISE WARNING` and full idempotency are unchanged; the migration is unmerged, so editing it is safe (verified it is not reachable from `origin/main`).

**What this was, and was not.** The original form *did* elevate at runtime — this is a guard-recognition fix, not a latent production bug. I checked rather than assuming, against the test stack **as the unprivileged `breeze_app` role**, which is where a missing election actually bites and where CI's superuser masks it:

| form | rows touched |
|---|---|
| no election at all | **0** — the silent no-op the guard exists to catch |
| `SET LOCAL` inside the `DO` block (original) | 1 |
| `SELECT set_config(...)` at top level (now) | 1 |

The 0-row row is the useful one: it confirms the election is genuinely load-bearing here, so the guard is protecting something real even though this particular file was already safe.

**Test change.** The migration-content assertion now requires the canonical form and rejects `SET LOCAL` — with line comments stripped first, since the file now *documents* why `SET LOCAL` is the wrong form and a naive negative match would fail on its own explanation.

**Verification.** `migrationRlsScope.test.ts` and `autoMigrate.test.ts` green (red first for both: the guard reported `…-device-lifecycle-…sql:44 UPDATE devices`, and the content test failed on the missing canonical form). Whole `src/db` suite 59 files / 2110 tests green. `check-migration-naming.sh` clean, including `--against-ref origin/main`. Migrations re-applied to a **fresh** database and `db:check-drift` clean (687 files). Integration re-run against live Postgres: `removedDevicePurge` 10/10 including the "migration backfill replayed" case (confirmed it actually ran, not skipped), plus `deviceLifecycle` and `tenant-export-policy`. `tsc` clean, lint clean, stack torn down with no leftover compose project.

**Process note worth recording:** my pre-PR verification list came from the plan doc and named `autoMigrate.test.ts` but not `migrationRlsScope.test.ts`, so this guard never ran locally. Any change touching a migration should run both.

PR #5055, opened 2026-09-06
feat(api): device custom-field VALUE importer incl. the warranty target (#3257 W08) (#4776)

### LINKED ISSUE #4776: W08: Values importer incl. the warranty target
Wave W08 of #4768 — Values importer incl. the warranty target
Branch: `feature/4768-rmm-custom-field-import/wave-4776`

ToddHebebrand:
Closes #4776

Wave 08 of #4768 (#3257): the **values** half of the RMM custom-field importer, including the `warranty` mapping target. **No migration in this wave.** No `apps/web` (W09), no docs (W10), no changes to W05's table/triggers/registration lists.

Premise verified on disk at `origin/main` before branching (`f02c7f8d0`), not assumed: W04 (`validateCustomFieldValue`, `loadVisibleCustomFieldDefinitions`), W05 (`device_custom_field_values`, `persistDeviceCustomFieldValues`, `2026-10-11-160000-device-custom-field-values.sql`), W06 (`import/{types,resolveDevice,deviceIdentity}.ts`, `device_external_links`), W07 (`import/{definitionImport,audit}.ts`, `customFields/writeErrors.ts`, `MAX_IMPORT_VALUES`) — all four present.

## Routes

| Method | Path | Behaviour |
|---|---|---|
| POST | `/devices/custom-fields/import/preview` | Resolves each row to a device and annotates **every value on it**. Writes nothing. |
| POST | `/devices/custom-fields/import` | Writes the acknowledged values, links and warranty. Always **200**, even with a non-empty `errors[]`. |

Both: `authMiddleware` (per route, never `.use('*')`) + `requireScope('organization','partner','system')` + `requirePermission(devices:write)` — the same grant `routes/devices/customFieldValues.ts` requires — + `requireMfa()`. **No `dualAuth`**, so an `X-API-Key`-only caller is a 401. Mounted immediately after `customFieldValuesRoutes` and before `coreRoutes`.

## Codes

**Row outcome** (`ValueRowOutcome`, aliased from W06's `DeviceRowOutcome` — not a second copy): `matched`, `link-match`, `ambiguous`, `not-found`, `org-not-found`, `identity-conflict`.

**Per-VALUE outcome** (`ValueOutcome`)

| Outcome | Meaning |
|---|---|
| `applied` | Written (or, at preview, would be). |
| `skipped-already-set` | A value is already stored and writing would change nothing, **or** it differs and the default `skip` mode declines to overwrite it. |
| `skipped-provider-owned` | Warranty only: a manufacturer API lookup owns this device's warranty and the operator did not opt into overriding it. |
| `no-definition` | No `custom_field_definitions` row visible to this device's org owns the key — run the definitions import (W07) first. |
| `type-error` | `validateCustomFieldValue` refused it; `reason` says how. |
| `not-applicable-to-device` | The definition is scoped to `deviceTypes` this device's OS is not in. |
| `device-unresolved` | The **row** did not resolve, so no value on it can be judged. |

**Commit error codes** (`ValueImportErrorCode`): `org-not-found`, `not-found`, `identity-conflict`, `annotation-changed`, `match-changed`, `match-unconfirmed`, `write-failed`.

**Warranty row outcome** (`WarrantyImportOutcome`): `applied`, `skipped-provider-owned`, `skipped-already-set`, `rejected` (a warranty column was mapped and every mapped cell was refused), `none` (no warranty column mapped at all).

## The four proofs the plan requires

All against real Postgres in `deviceCustomFieldValueImport.integration.test.ts` (10/10), which does **not** appeal to RLS on the resolution read — `loadDeviceResolutionSnapshot` runs under system scope where `breeze_has_org_access` is TRUE for every row, so an RLS assertion there would be vacuous.

1. **Isolation.** Under a partner-A context, four rows reaching for partner B's device by `deviceId`, by `hostname` + `serialNumber`, and by naming partner B's org all come back `not-found` / `org-not-found` — **never `write-failed`**, which would mean the write was attempted — and nothing lands on the foreign device while the legitimate row does.
2. **Transaction isolation.** With the app-layer scope deliberately widened past the request's RLS context (the shape a caller bug would take), the middle row's INSERT is refused by Postgres, the row **before** it stays committed and the row **after** it still runs: `errors = [{index: 1, write-failed}]`, `rows = [0, 2]`. That is simultaneously the proof that the writes really do stay inside the request context, and the proof that the per-row savepoint holds.
3. **Partial application.** One row, five values → 2 applied and stored, 3 reported (`no-definition`, `type-error`, `not-applicable-to-device`), and `devices.custom_fields` reprojected by W05's trigger with exactly the two.
4. **Second-run link resolution.** Run 1 resolves by `hostname` and mints the `device_external_links` row. A second device then takes the same hostname. Run 2 resolves `link-match` / `method: 'link'` with `linksCreated: 0`, while the control row — same hostname, no external id — is refused `match-unconfirmed`. Query-shape half in `valueImport.test.ts`: with only an external id supplied the resolver's `devices` query carries **no `lower(hostname)` predicate at all**, with the opposite case as the mutation control.

Plus: the warranty write flips `device_warranty.status` to `expiring` (with `is_subscription` false and `data_source: 'import'`) and `evaluateWarrantyAlerts` then returns an alert id where it returned `null` before — proven with the config-policy chain seeded, because warranty alerting is opt-in and a test without it would pass vacuously against `DISABLED_SETTINGS`.

1. **Writes stay in the request context, not `runOutsideDbContext`.** The plan opened each row's transaction outside the request context. That bypasses RLS on the write path of a bulk cross-organization tool — the place it is least affordable. This follows **W07's shipped pattern**: a nested `db.transaction` (a SAVEPOINT inside the request transaction, per `dbSavepointErrorIsolation.integration.test.ts`), which gives the same per-row failure isolation *and* keeps RLS as a real second control. Proof #2 above is the evidence that it does.
2. **`persistDeviceCustomFieldValues` gained an optional `executor` parameter** (`queries.ts`, one file outside the brief's "files you own"). The savepoint contract requires every statement to be issued on the nested `tx` — the ambient `db` proxy resolves to the OUTER transaction and would poison the request. The only alternatives were forking W05's upsert (its compare-before-write `setWhere` is subtle, load-bearing and would drift) or losing the isolation. The parameter defaults to `db`, so all three existing call sites are unchanged.
3. **`DeviceCustomFieldImportValue` changed from `{ fieldKey, value }` to `{ target, value }`**, per the plan's W08 Task 1 `MappingTarget`. W06 shipped the placeholder shape with a comment explicitly reserving the field's meaning for W08, and nothing outside these types ever read it (`resolveDeviceRow` does not touch `.values`).
4. **`ValueOutcome` drops the plan's `reserved-key` and adds `device-unresolved` + `skipped-provider-owned`.** The plan declared `reserved-key` but its own Task-2 test specifies the behaviour as *applied, with a warning* — a dead arm every W09 branch would carry. Meanwhile the plan's ambiguous-row test requires every value on an unresolved row to be non-`applied` and offers no member for it (`device-unresolved`), and the review pass established that a provider-owned warranty refusal must not wear the same badge as "nothing to do" (`skipped-provider-owned`, mirroring the row-level vocabulary). Every member is produced by the code.
5. **A declined warranty column reports `skipped-provider-owned` at the row level, not `none`.** Caught by the integration suite: `none` means "this row mapped no warranty column at all", and collapsing the two tells an operator their warranty column was never mapped when in fact it was refused.
6. **A row may not map two columns onto the same target** (zod refine, both routes). A duplicate is not one bad cell: the column-to-target mapping is chosen once for the whole file, so it would repeat on every row and the last column would silently win with the operator never told which value the device holds. The service's tally dedupes as well, so a direct caller cannot inflate `appliedValues` either.
7. **The `fieldKey` on a value row carries no `^[a-z][a-z0-9_]*$` regex**, unlike W07's definition-create rows. A key that could never own a definition simply gets `no-definition`; 400-ing a 30-column file over one bad header would defeat the per-value annotation this wave exists for.
8. **`loadVisibleCustomFieldDefinitions` is called once per DISTINCT organization** rather than replaced by a batched query. It owns the org-XOR-partner visibility rule and a second copy of that rule is the drift the partner-wide retrofits already cost us; the per-call system context (#1105) is noted at the call site as the place to batch if a file ever spans dozens of orgs.
9. **Preview strips acknowledgement fields** (`expectedOutcome` / `expectedDeviceId`) rather than 400-ing on them — zod's default object behaviour — so the service can never see a pin from the preview route and cannot silently honour one. Asserted.

## Caps and mutation checks

`MAX_IMPORT_ROWS = 1000` **and** `MAX_IMPORT_VALUES = 5000` total values, both at the zod layer with copy telling the browser to **split** (the row cap alone does not bound the work: 1000 × 30 = 30,000 writes). Asserted at exactly the cap and one over.

Every gate has a mutation check, run and recorded:
- delete `requireMfa()` → the MFA test goes red (1 failure);
- delete `requireDeviceWrite` → 8 tests go red;
- neutralise the duplicate-target dedupe in `tallyRow` → 1 test goes red;
- replace the per-route middleware with a wildcard → the sibling X-API-Key PATCH guard in `customFieldImport.mountorder.test.ts` is the detector;
- flip the `skip`-mode branch in `decideValue` → 1 test goes red;
- omit the `tx` executor → the whole service suite throws (the mocked ambient `db` has no `insert`).

## Verification

- `npx vitest run src/services/customFields src/services/warrantySync src/routes/devices/customFieldImport src/routes/devices/customFieldValues src/routes/devices/index src/routes/customFieldImport src/openapi` — **20 files, 355 tests, all pass**, re-run after merging `origin/main` and after the review-fix pass.
- `npx vitest run --config vitest.integration.config.ts src/__tests__/integration/{deviceCustomFieldValueImport,deviceCustomFieldValues,customFieldDefinitionImport,deviceCustomFieldImportIsolation,scriptCustomFieldWriteBack,deviceExternalLinks}` — **6 files, 96 tests, all pass** against a real Postgres (private `pnpm test-stack`, since the shared :5433 stack had another worktree's unmerged migration in its ledger).
- `NODE_OPTIONS=--max-old-space-size=8192 npx tsc --noEmit` — clean. `pnpm lint` — clean (6/6).
- `origin/main` merged into the branch before the final verification run.

## Not verified

- **The full `apps/api` unit suite and the full integration suite** were not run — only the affected files above. CI covers the rest.
- **`rls-coverage` / `tenantCascade` / `tenant-export-policy`** were not run: this wave adds no table, no column and no migration, so none of the six registration lists changes. Asserted by inspection, not by execution.
- **No web client exists yet** (W09), so the wire contract has not been exercised by a real browser caller and `runAction`'s handling of these bodies is reasoned about rather than observed.
- **`X-API-Key` → 401** is asserted against a mock of `authMiddleware` reproducing production's Bearer-only contract, not the real middleware (which has no API-key branch at all — verified by inspection).
- **Concurrency**: the TOCTOU and race paths are exercised as sequential state changes between two requests, not under real parallelism. The `setWhere` guards on both the value upsert and the warranty upsert are the authority there and are asserted, but not raced.
- **Performance at the 1000-row / 5000-value cap** was not measured. Each row is one transaction, so a full batch is up to 1000 sequential transactions.
- **The audit fan-out** is asserted through `writeCustomFieldValueImportAudits` unit tests and the route wiring; no test observes the rows actually landing in `audit_logs`.

https://claude.ai/code/session_01DR1Q8zCfWEAAiJshvXv1HR

### DIFF
--- a/apps/api/src/__tests__/integration/deviceCustomFieldValueImport.integration.test.ts
+++ b/apps/api/src/__tests__/integration/deviceCustomFieldValueImport.integration.test.ts
@@ -260,6 +260,7 @@ describe('device custom-field VALUE import (real Postgres)', () => {
     const world = await seedWorld();
     const device = await seedDevice({ orgId: world.orgA, siteId: world.siteA, hostname: 'partial-1', serialNumber: 'sn-partial-1' });
     await seedDefinition({ partnerId: world.partnerA, fieldKey: 'bitlocker_status', type: 'text', deviceTypes: ['macos'] });
+    await seedDefinition({ partnerId: world.partnerA, fieldKey: 'rack_position', type: 'number' });
 
     const ctx: ValueImportContext = {
       partnerId: world.partnerA, accessibleOrgIds: [world.orgA], allowedSiteIds: null, mode: 'update',
@@ -273,7 +274,7 @@ describe('device custom-field VALUE import (real Postgres)', () => {
             v('asset_tag', 'AB-1'),
             v('rack_units', 4),
             v('never_defined', 'x'),
-            v('rack_units', 'not a number'),
+            v('rack_position', 'not a number'),
             v('bitlocker_status', 'on'),
           ],
         }] as CommitValueRowInput[],
@@ -376,6 +377,82 @@ describe('device custom-field VALUE import (real Postgres)', () => {
     expect(annotated[0]!.values.map((x) => x.outcome)).toEqual(['skipped-already-set', 'skipped-already-set']);
   });
 
+  runDb('re-running an identical WARRANTY file previews and commits as already-set', async () => {
+    // Preview and commit must agree on warranty exactly as they do on custom
+    // fields; before the review pass, preview promised `applied` here.
+    const world = await seedWorld();
+    const device = await seedDevice({ orgId: world.orgA, siteId: world.siteA, hostname: 'warranty-3', serialNumber: 'sn-warranty-3' });
+
+    const ctx: ValueImportContext = {
+      partnerId: world.partnerA, accessibleOrgIds: [world.orgA], allowedSiteIds: null, mode: 'update',
+    };
+    const rows = [{
+      hostname: 'warranty-3', values: [w('warrantyEndDate', inDays(30)), w('manufacturer', 'Dell')],
+    }] as CommitValueRowInput[];
+
+    const first = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      commitDeviceCustomFieldImport(rows, ctx, actor));
+    expect(first.rows[0]!.warranty).toBe('applied');
+
+    const annotated = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      previewDeviceCustomFieldImport(rows, ctx));
+    expect(annotated[0]!.values.map((x) => x.outcome)).toEqual(['skipped-already-set', 'skipped-already-set']);
+
+    const again = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      commitDeviceCustomFieldImport(rows, ctx, actor));
+    expect(again.rows[0]!.warranty).toBe('skipped-already-set');
+    expect(again.appliedValues).toBe(0);
+
+    // And nothing was rewritten.
+    const [row] = await getTestDb()
+      .select({ endDate: deviceWarranty.warrantyEndDate })
+      .from(deviceWarranty)
+      .where(eq(deviceWarranty.deviceId, device));
+    expect(row).toMatchObject({ endDate: inDays(30) });
+  });
+
+  runDb('the durable link key round-trips through externalSourceInstance', async () => {
+    const world = await seedWorld();
+    const device = await seedDevice({ orgId: world.orgA, siteId: world.siteA, hostname: 'si-1', serialNumber: 'sn-si-1' });
+
+    const ctx: ValueImportContext = {
+      partnerId: world.partnerA, accessibleOrgIds: [world.orgA], allowedSiteIds: null, mode: 'update',
+    };
+    const row = {
+      externalSystem: 'datto_rmm',
+      externalId: 'uid-si-1',
+      externalSourceInstance: 'tenant-a',
+      hostname: 'si-1',
+      values: [v('asset_tag', 'AB-1')],
+    } as CommitValueRowInput;
+
+    const first = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      commitDeviceCustomFieldImport([row], ctx, actor));
+    expect(first.rows[0]).toMatchObject({ method: 'hostname', linkCreated: true });
+
+    // Second run, same row: the link written above must be the one the resolver
+    // reads back — proving the write side and W06's read side build the SAME
+    // composite key including the reserved discriminator.
+    const second = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      commitDeviceCustomFieldImport([{ ...row, values: [v('asset_tag', 'AB-2')] }], ctx, actor));
+    expect(second.rows[0]).toMatchObject({ deviceId: device, method: 'link', linkCreated: false });
+
+    // A DIFFERENT source instance is a different key and must not match it.
+    const other = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      commitDeviceCustomFieldImport(
+        [{ ...row, externalSourceInstance: 'tenant-b', values: [v('asset_tag', 'AB-3')] }],
+        ctx,
+        actor,
+      ));
+    expect(other.rows[0]).toMatchObject({ method: 'hostname', linkCreated: true });
+
+    const links = await getTestDb()
+      .select({ sourceInstance: deviceExternalLinks.sourceInstance })
+      .from(deviceExternalLinks)
+      .where(eq(deviceExternalLinks.deviceId, device));
+    expect(links.map((l) => l.sourceInstance).sort()).toEqual(['tenant-a', 'tenant-b']);
+  });
+
   runDb('the warranty target writes a COMPUTED status, so evaluateWarrantyAlerts actually fires', async () => {
     const world = await seedWorld();
     const device = await seedDevice({ orgId: world.orgA, siteId: world.siteA, hostname: 'warranty-1', serialNumber: 'sn-warranty-1' });
@@ -460,10 +537,13 @@ describe('device custom-field VALUE import (real Postgres)', () => {
       hostname: 'warranty-2', values: [w('warrantyEndDate', inDays(30))],
     }] as CommitValueRowInput[];
 
+    const guardedPreview = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
+      previewDeviceCustomFieldImport(rows, base));
+    expect(guardedPreview[0]!.values[0]!.outcome).toBe('skipped-provider-owned');
+
     const guarded = await withDbAccessContext(partnerCtx(world.partnerA, [world.orgA]), () =>
       commitDeviceCustomFieldImport(rows, base, actor));
-    // The ROW-level outcome names the reason; the per-VALUE annotation says
-    // `skipped-already-set` with the warning. Reporting `none` here would tell
+    // The ROW-level outcome names the reason. Reporting `none` here would tell
     // the operator their warranty column was never mapped.
     expect(guarded.rows[0]!.warranty).toBe('skipped-provider-owned');
     let [row] = await getTestDb()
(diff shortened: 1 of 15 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — five agents in parallel: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `type-design-analyzer`, `comment-analyzer`.

**Findings:** 13 raised → 11 addressed in 75f9578, 2 explicitly dismissed; 0 outstanding.

Addressed:

| # | Agent | Finding | Fix |
|---|---|---|---|
| 1 | comment-analyzer | **Module header misattributed a security rationale.** It gave `custom_field_definitions`' missing partner-wide RLS SELECT branch as a reason `loadDeviceResolutionSnapshot` escalates to a system context — that function never reads that table. The rationale belongs to the separate `loadVisibleCustomFieldDefinitions` call in the same module. | Header now names the two escalations separately and says which one #4944 would affect. |
| 2 | code-reviewer | **Preview promised `applied` for a warranty cell commit would skip.** Preview loaded only `data_source`, so a re-import of an unchanged file previewed `applied` on every warranty cell and then committed a skip — preview and commit disagreeing on the one surface whose job is to predict the other. | Loads the whole warranty row and compares per cell via a new `warrantyCellUnchanged`, which also checks the derived `status` (it decays with time, so an unchanged date whose status has gone stale is still a change — otherwise the import would quietly stop being what refreshes it). |
| 3 | code-reviewer + silent-failure-hunter | **A row mapping two columns onto the same target inflated `appliedValues`** by one per duplicate, and the last column silently won. | Refused at the zod layer on both routes (it is a wizard mis-mapping that repeats on every row, not one bad cell), and `tallyRow` dedupes by target anyway so a direct service caller cannot inflate the count either. |
| 4 | silent-failure-hunter | **`warranty: 'none'` for a mapped-but-rejected warranty column** — the same value used when no warranty column was mapped at all. | `WarrantyImportOutcome` gains `rejected`. |
| 5 | type-design-analyzer | **`skipped-already-set` was doing three jobs**, with a provider-owned warranty refusal distinguishable only by an optional free-text `warning`. | `skipped-provider-owned` is its own `ValueOutcome`, mirroring the row-level vocabulary. |
| 6 | type-design-analyzer | `reason` was an optional field beside a comment. | Discriminated onto the `type-error` arm, matching `DeviceResolution` in the same file. |
| 7 | type-design-analyzer | The route's zod schema mirrored `MappingTarget` / `WarrantyImportField` **in prose only** — a new domain member would have been silently unselectable from the wire. | `satisfies Record<WarrantyImportField, true>` plus a `z.infer extends MappingTarget` assertion: tied in both directions. |
| 8 | silent-failure-hunter | Neither the row-write catch nor the audit fan-out catch called `captureException`, so a code-level regression would only be visible to whoever was grepping logs. | Added. The row-write one fires **only** on an unrecognised SQLSTATE (or a non-pg error) so a bad CSV cannot flood Sentry. |
| 9 | type-design-analyzer | `ValueImportSummary`'s aggregates mix units — `failedValues` counts values inside written rows, whole-row refusals live in `errors[]` in row units. | Documented on the type; they are not addable without converting. |
| 10 | pr-test-analyzer | Test gaps: `externalSourceInstance` never exercised (a write/read key drift would silently duplicate links); the DB-decline path (`changed` shorter than `fieldWrites`) was dead code to the tests; `update` mode + an identical stored value; a stale `expectedDeviceId` on an unambiguous row; `device-unresolved` only proven for `ambiguous`; a zero-value row. | All six now covered, plus the duplicate-target refusal and warranty preview parity — unit and, for the link key, integration. |
| 11 | (found by the new tests) | The first dedupe attempt let a later **type-error** value supersede an earlier applied one. | Only a decision that would actually write can supersede. Caught by the integration suite before push. |

Dismissed, with reasons:

- **`AnnotatedValueRow` re-flattens what `DeviceResolution` discriminates.** Real tension, no cheap fix — on commit a pinned `ambiguous` row legitimately carries a `deviceId`, which `DeviceResolution`'s own type forbids for that outcome, so reusing it would mean inventing a seventh outcome. The reviewer agreed: flag, don't chase.
- **`WarrantyImportWrite` should be a `Map` so `in` can't be fooled by an explicit `undefined`.** Latent only: one producer, and the value it assigns comes from `coerceWarrantyValue`, which is total and never returns `undefined`. Not worth a rewrite at one call site.

Also noted and left alone: `ValueImportContext extends DeviceResolutionScope` (one construction site, built field-by-field from `auth`), and the one-directional `asImportRejection` bridge (it catches the direction that matters — a new validator reason that never reached the wire type — and fails safe in the other).

**Mutation checks** (each run, then reverted): delete `requireMfa()` → 1 red; delete `requireDeviceWrite` → 8 red; neutralise the `skip`-mode branch → 1 red; neutralise the duplicate-target dedupe → 1 red; drop the `tx` executor → the whole service suite throws. The mount-order guard's sibling X-API-Key assertion is the detector for someone replacing the per-route middleware with a wildcard.

**Tests:** `apps/api` affected unit suites green (20 files, 355 tests); integration green against real Postgres (6 files, 96 tests, private `test-stack` — the shared :5433 stack had another worktree's unmerged migration in its ledger). `tsc --noEmit` clean; `pnpm lint` clean.

**Status:** review-clean, awaiting maintainer merge. No migration in this wave and no registration-list change, so `rls-coverage` / `tenantCascade` / `tenant-export-policy` were not run — asserted by inspection, and CI's Integration Tests job covers them.

PR #5060, opened 2026-09-06
feat(web): Import from another RMM wizard for custom fields (#3257 W09)

### LINKED ISSUE #4777: W09: Web wizard
Wave W09 of #4768 — Web wizard
Branch: `feature/4768-rmm-custom-field-import/wave-4777`

ToddHebebrand:
Closes #4777

## What changed

Web UI for the "Import from another RMM" wizard (#3257 W09), driving the already-merged W07 (definitions, #5043) and W08 (values, #5055) API routes end-to-end:

- **`CustomFieldImportPreviewTable.tsx`** — the VALUES-step preview table: per-row outcome badges, per-value outcome sub-rows, ambiguous/identity-conflict rows expand to ranked candidates (serial/OS/enrolled/last-seen) and require an explicit pick before becoming committable, select-all spans only `matched`+`link-match` rows, reserved-identity-key warnings render inline.
- **`CustomFieldDefinitionImportStep.tsx`** — Step 1: one CSV row is a candidate field *definition* (an incumbent's slot catalogue, e.g. Datto's `udf1..udf30`), not a data record, so the grid is the CSV's own rows pre-filled from `sourceLabel` with every type defaulting to `text`. `ownerScope` + its capability gate come from `useDefaultOwnerScope()` / `user.canManagePartnerWide` (not re-derived). Non-committable annotations (`type-conflict`/`key-shadowed`/`org-not-found`/`partner-wide-denied`) are excluded from selection; skippable straight to values when everything already exists.
- **`CustomFieldValueImportStep.tsx` + `customFieldImportCoercion.ts`** — Step 2: per-column mapping to an identifier, a custom field, a warranty target, or ignore; explicit CSV coercion (number/boolean/date, with an operator-chosen date-format for otherwise-ambiguous `dd/mm` vs `mm/dd`) against each target's *declared* type (fetched from `GET /custom-fields`); preview and commit both chunked at `MAX_IMPORT_ROWS`/`MAX_IMPORT_VALUES` with cumulative progress; partial success surfaced in VALUE units (matching the API's own counting unit).
- **`RmmCustomFieldImport.tsx`** — the wizard shell. Step state lives in `window.location.hash` (`#import-definitions` / `#import-values`), never a query param. The RMM source picker (Datto/NinjaOne/ConnectWise Automate/N-central/CSV) lives here since the values step needs no source.
- Entry points: **`DevicesPage.tsx`** and **`CustomFieldsPage.tsx`** both gained an "Import from another RMM" launch button.
- **8-locale i18n**: real translations (not English fallback) for all new copy across `en, de-DE, es-419, fr-CA, fr-FR, it-IT, pt-BR, tr-TR`. `localeParity`/`translationCoverage`/`terminologyQuality` all green.
- **`no-silent-mutations.test.ts`**: the three new mutating components added to `TARGET_GLOBS` (119 → 122); still green — every preview/commit POST is `runAction`-wrapped.
- **E2E**: `e2e-tests/tests/custom-field-import.spec.ts` + `RmmCustomFieldImportPage.ts`.

## The one API-adjacent change (as instructed by the brief)

Moved `MAX_IMPORT_ROWS`/`MAX_IMPORT_VALUES` from `apps/api/src/services/customFields/import/types.ts` into `packages/shared/src/validators/customFieldImport.ts`, so the web wizard's chunking uses the exact same numbers the API enforces. The API file re-exports both names unchanged — no W07/W08 caller changed. API import suites (240 tests) still pass; typecheck clean.

- **`MappingTarget.kind` is `'customField'` (camelCase), not `'custom-field'`** as the brief's prose said — read from `apps/api/src/services/customFields/import/types.ts` and matched exactly.
- **`CustomFieldImportPreviewTable` is NOT shared between the two wizard steps.** It renders the values-step row shape only (`AnnotatedValueRow`: candidates, per-value outcomes). The definitions step's row vocabulary (`create`/`already-exists`/`type-conflict`/`key-shadowed`/`org-not-found`/`partner-wide-denied`) shares nothing with it beyond the general preview/select shape — same non-generalization reasoning `ContactImportPreviewTable.tsx` already documents for the org/contact importers — so the definitions step renders its own simpler list inline instead.
- **One data-testid added to `DeviceInfoTab.tsx`** (`device-custom-field-value-<fieldKey>`) outside the nominal owned-file list — the only way to assert the imported value by data-testid (repo convention) on the device detail page from the E2E spec. No other change to that file.
- **Final `ValueOutcome` enum already reflects the "skipped-already-set split into distinct codes" the brief flagged** — `skipped-already-set` and `skipped-provider-owned` are already separate members in the merged W08 types; no further split was needed.

## Not verified

- **E2E spec was NOT run against a live stack.** `pnpm wt-stack up` failed with `Error response from daemon: invalid pool request: Pool overlaps with other one on this address space` — Docker network-pool exhaustion from the many other worktree stacks concurrently running on this host, not a defect in this PR. Verified instead: `playwright test --list` resolves the one spec, and `npx tsc --noEmit` in `e2e-tests` is clean.
- **The E2E spec does not exercise the ambiguous-row "pick a candidate" sub-flow.** That needs a second device sharing a hostname with an existing seeded device; `seedE2eFixtures.ts` (`apps/api`, outside this wave's owned files) only seeds two devices with distinct hostnames. The candidate-picking UI itself is covered exhaustively at the unit level in `CustomFieldImportPreviewTable.test.tsx` (expand, rank, pick, aria-disabled-until-picked). Flagging as a follow-up: add a duplicate-hostname device fixture to `seedE2eFixtures.ts` so a future pass can close this gap.
- **Not manually walked in a browser** (see E2E note above) — reviewed via component tests + code only.

## Test evidence

- `apps/web`: 366 tests across the new components, i18n suites, and the two touched entry-point pages — all green.
- `astro check`: 0 errors (pre-existing unrelated warnings only).
- `pnpm lint`: clean.
- `packages/shared`: 2471 tests green (includes the new cap constants).
- `apps/api` (import suites only, per brief's scope): 240 tests green; `tsc --noEmit` clean.


ToddHebebrand:
## Review summary

Ran 4 parallel reviews against the PR diff: **code-reviewer**, **pr-test-analyzer**, **silent-failure-hunter**, **type-design-analyzer**. All 4 came back with substantive, verified findings (each cross-checked the web restatements against the actual API contract in `apps/api/src/routes/{customFieldImport,devices/customFieldImport}.ts` and `services/customFields/import/{types,valueImport}.ts`).

### Fixed (9 commits, all green)

**Correctness (code-reviewer, confirmed against `valueImport.ts`):**
- `identity-conflict` rows could be "picked" and selected for commit, but the server's `REFUSED_OUTCOMES` refuses that outcome unconditionally — checked *before* it ever reads `expectedDeviceId`. The pick had zero effect on the wire and always landed silently in `errors[]`. Candidates now render read-only (no pick button) with an explicit "correct the file and re-run preview" note; only `ambiguous` remains pickable.
- The RMM source picker (Datto/NinjaOne/CW Automate/N-central) never reached either request body — `externalSystem` defaulted to `'csv'` server-side regardless of what the operator selected, corrupting the audit trail. `source` is now threaded into all four preview/commit bodies (both steps).
- `generateFieldKey` could produce a key the server's `^[a-z][a-z0-9_]*$` regex rejects (leading digit, or no letters at all), 400ing the *entire* definitions batch with no in-UI fix, since that grid has no per-row key input. Now guaranteed-valid (`f_` prefix / `f_untitled` fallback).
- Commit's chunk-index remapping was structurally wrong, not just imprecise: `chosenRows` is a *subset* of previewed rows, so a cumulative offset (correct for preview) silently mislabels rows after any gap in the selection. Now remaps each chunk-local response index through that chunk's own recorded original indexes.
- `customFieldImportCoercion.ts`: boolean treated an empty cell as `''` (forcing `type-error` on every sparse boolean column) instead of `null`; number coercion guessed on EU-style decimal/thousands commas (`1,5` → `15`) instead of passing them through per the module's own "never guess" contract.
- Keyless-Fragment React warning fixed.

**Silent failures (silent-failure-hunter, independently found the same errors[] issue as code-reviewer):**
- `summary.errors[]` was collected from every commit response and never rendered in either step — a row refused for `write-failed`/`annotation-changed`/RLS-denial/etc. produced an all-zeros summary with **no error anywhere**. Now rendered as a per-row list, with the summary switching to destructive styling.
- A failed chunk mid-commit neither hid what earlier chunks wrote nor silently dropped the count of never-attempted rows — chunks that complete are removed from `selected` (so a retry doesn't re-apply already-written rows under `mode:'update'`), and an explicit "N rows not attempted" line renders.
- A failed re-preview no longer leaves a stale, fully-committable table from a prior successful preview on screen.
- The field-type lookup's empty `catch` now logs via `console.error` instead of swallowing every failure mode with zero trace.
- `fieldTypeByKey` now detects a key with conflicting types across orgs (a partner-wide `GET /custom-fields` can return every org's definitions) and skips type-specific coercion for it, rather than coercing against whichever definition sorted last.

**Type design (type-design-analyzer):**
- `AnnotatedImportValue` is now a discriminated union matching the API's own type exactly (`reason` only reachable on `outcome: 'type-error'`) — the one place the web's restated types diverged from the wire contract's own invariants.
- `commit()` now re-derives selectability via `isValueRowSelectable` as defense-in-depth rather than trusting `selected` unconditionally.

**Test coverage (pr-test-analyzer):**
- Fixed two vacuous assertions: the coercion test now checks the actual request body (`12`, not `"12"`); the partial-success test now checks the exact failed-count text, not a `toHaveTextContent('1')` substring match that would pass unchanged if the failed count silently dropped out.
- Added tests for: both branches of `RmmCustomFieldImport`'s auto-advance-on-commit conditional, the definitions step's all-refused-commit path, `chunkValueRows`' oversized-single-row edge case, `toggleAll`'s deselect branch, individual per-row checkbox toggling, and two previously-untested `coerceCellForType` date branches (explicit-format-on-invalid-date, neither-reading-valid).

Also fixed for consistency: `CustomFieldsPage.tsx` now clears `window.location.hash` on wizard close, matching `DevicesPage.tsx`.

### Dismissed (with reasoning)

- **Duplicate `cf-import-value-outcome-${outcome}` testid across rows** (code-reviewer, lower-priority) — this is the plan's own literal testid contract (`cf-import-value-outcome-no-definition`, no per-row suffix), and the only test exercising it renders a single row, so no collision occurs in the tests that matter. Changing it would deviate from the plan-mandated spec for a theoretical multi-row lookup nobody does.
- **`DeviceCustomFieldImportRow` omits `externalSourceInstance`** (type-design-analyzer) — the reviewer's own assessment: "harmless today... TS structural typing allows sending a narrower optional-field object," and the API documents the field as always-null/reserved. Not adding a field the wizard never needs to set.

### Verification after fixes

- `apps/web`: 461 tests across all new/touched files — green.
- `astro check`: 0 errors.
- `pnpm lint`: clean.
- `apps/api` (import suites): 240 tests green; `tsc --noEmit` clean.
- `e2e-tests`: `tsc --noEmit` clean, `playwright test --list` resolves the spec (still not run against a live stack — see PR body's not-verified section, unchanged: Docker network-pool exhaustion on the dev host, unrelated to this PR's code).

Not merging or closing — leaving for review/merge per process.

Issue #5048, opened 2026-09-06
AI agent builder: capability picker and guided create flow


ToddHebebrand:
Tracking issue for **AI agent builder: capability picker and guided create flow** (`agent-builder`).

- Plan: `docs/superpowers/plans/ai-mcp/2026-09-06-ai-agent-builder.md`
- Spec: `docs/superpowers/specs/ai-mcp/2026-09-06-ai-agent-builder-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All three waves merged 2026-09-06: W01 #5054 (agent tool catalog, /tool-catalog + /ceiling, wildcard policy merge, grant-only supervised keys), W02 #5057 (CapabilityPicker replaces the allowlist textarea; browser-verified), W03 #5062 (POST /preview, AgentSummaryCard, four-step guided create flow; browser-verified). Unreleased; ships with the next release. Follow-ups: #5056 (web tierConfig vs TOOL_CAPABILITY category maps), #5059 (stale no-baseline notice), #5063 (edit drawer renders the step components).

PR #5064, opened 2026-09-06
fix(ai-agents): truthful run_script outcome, reachable stepper, op-level search, recipient copy, plurals (#5048 QA)

ToddHebebrand:
Browser QA of the AI agent builder (#5048 W01–W03, merged and unreleased) on a worktree stack found one truthfulness defect and a handful of UX/copy issues. All fixed here. Full walk log: `docs/testing/FEATURE_TEST_LOG.md` (2026-09-06 entry).

## What changed

- **Script gate (P1).** `run_script` is in `ACT_MANIFEST`, but the run loop only dispatches it unattended for a script in `actAssets.scriptIds` (`remediationActResolver.ts`, `hasActEligibleSurface`). The picker and the review card said "Executes unattended" for every agent built with the flow, which never sets `scriptIds`. The catalog now emits `actRequiresAuthorizedScripts` (from a new `SCRIPT_GATED_ACT_TOOLS` in `actManifest.ts`, which the activation prerequisite also reads), the shared outcome rule takes an `authorizedScriptCount` context, `POST /preview` carries `unattendedBlockedBy`, and both surfaces read "approval request until a script is authorized". The edit drawer passes the row's real script count, so an agent that *does* have authorized scripts still shows unattended.
- **Recipient 422 (P2).** The server folds "nothing selected" and "selected roles have no active members" into one `recipient` token; the flow told an operator who had just ticked a role to "add a recipient". Copy now depends on whether recipients were selected; the create flow imports the shared error map instead of its stale duplicate; `GET /roles`' `userCount` marks empty roles "(no active members)" in the fieldset.
- **Stepper (P2).** After an Edit link from Review, steps 2–4 were disabled again (three Next clicks to get back). `SetupStepper` gains `reachableStep`; the flow tracks the furthest step reached and validates every step being skipped on a forward jump. Completed circles now carry an accessible name ("1. Purpose and posture") instead of an unnamed check icon.
- **Search (P2).** Narrowed capabilities but not operations; "isolate" surfaced all twelve security operations because the capability blurb mentions isolating devices. Search now narrows to the matching operations and falls back to the whole capability only when no operation matches.
- **Copy (P3).** "Up to 1 devices per run", "1 roles are asked", "1 approval requests"; the act-mode "May propose" sentence omitted the unattended operations; the approvers row now names the roles. Pre-pluralised phrases in all 8 locales. Two French plurals ("{{count}} minute(s)") are identical to English and are reviewed into the duplicate baseline.
- **Default owner scope (P3).** With no partner baseline for the kind, a partner-scope session now starts partner-wide instead of an org-only draft the page immediately warns has no effect. The org-only choice is still one click away.
- Docs: outcome table notes the `run_script` exception.

## Verification
- `packages/shared`: `agentOutcome.test.ts` 6/6.
- `apps/api`: `src/services/aiAgents` + `routes/aiAgents.test.ts` → 76 files / 2199 tests green (new: preview script-gate cases, catalog contract pins `run_script` as the only gated op).
- `apps/web`: aiAgents components, SetupStepper, AiAgentForm, AiAgentsPage, no-silent-mutations → 14 files / 404 tests green; `src/lib/i18n` 8 files / 146 green.
- `tsc --noEmit` clean for shared (pre-existing `quotes.test.ts` errors excluded), api, web; eslint clean on every touched file.
- Browser re-check on the worktree stack: see the review-summary comment.

https://claude.ai/code/session_01CuEEGssvpeM9HE7sQqByGE


ToddHebebrand:
## Review summary

**Independent review** (`/code-review 5064 medium`, 10 finder angles, 9 verifiers): 15 findings. Disposition:

| # | Finding | Outcome |
|---|---|---|
| 1 | `goToStep` cleared issues on backward moves, so an Edit link after a 422 wiped the reason (regression introduced in this PR) | **Fixed** — backward moves keep issues; test now clicks the Edit link and asserts the reason survives |
| 2 | Search hid rows but the capability header checkbox still toggled every operation | **Fixed** — visible-operation set computed once per search drives the checkbox, count, list and rows; `capabilityState(..., onlyKeys)` |
| 3 | Owner-scope default read `partnerBaselineKinds` before the list loaded | **Fixed** — "New agent" is disabled until the list has loaded |
| 4 | Forward stepper jump reported an error for a control on another step | **Fixed** — lands on the first failing step |
| 5 | Drawer used the org row's `scriptIds`, not the effective `partner ∩ org` | **Fixed** — `AgentCeilingDto.scriptIds`; preview and drawer intersect against it |
| 6 | Copy points at a script-authorization control that does not exist | **Filed #5065**; docs now say the API sets `actAssets.scriptIds` today |
| 7 | `userCount` counts deactivated members | **Fixed** — `GET /roles` also returns `activeUserCount` (status='active' FILTER, same query chain); the fieldset reads that. The org-access nuance for partner members is not covered by the badge |
| 8 | `recipientsSelected` ignored API-set `recipients.userIds` | **Fixed** in the drawer |
| 9 | Horizontal stepper buttons had no accessible name below `sm` | **Fixed** |
| 10 | `SCRIPT_GATED_ACT_TOOLS` not tied to its consumers by a test | **Fixed** — `actManifest.test.ts` pins it as a literal and asserts manifest membership |
| 11 | Reachable-ahead steps looked identical to unreachable ones | **Fixed** — distinct ring + dotted label, `data-reachable` |
| 12 | Operation checkbox unnamed; script-gate note unassociated | **Fixed** — `aria-label` + `aria-describedby` |
| 13 | Owner-scope rule lives in the flow, not `useDefaultOwnerScope` | **Left as is** — the hook has no access to `partnerBaselineKinds`; noted in the flow's comment |
| 14 | Picker breakdown used `join(', ')` vs the card's `Intl.ListFormat` | **Fixed** |
| 15 | Search scan ran twice per capability per keystroke; summary recomputed every render | **Fixed** — single memoised map + memoised summary |

**Verification after the second pass**
- API: `src/services/aiAgents` + `routes/aiAgents.test.ts` + `routes/roles.test.ts` → 77 files / 2290 tests green.
- Web: aiAgents components, SetupStepper, AiAgentForm, AiAgentsPage, `src/lib/i18n` → 21 files / 415 tests green.
- `tsc --noEmit` clean for shared, api, web; eslint clean on every touched file.
- Live on the worktree stack (`pnpm wt-stack`, act mode, partner-scope admin): search "restart" narrows Services and startup to "0 of 1 operation" and its header checkbox selects only `manage_services:restart`; summary reads "(0 approval requests, 0 logged proposals, and 1 unattended)"; Create with an empty role shows "The selected notification roles have no active members…" and the same text is still showing after the Approvers Edit link lands on Safety; the visited Review step carries `data-reachable`. Earlier pass (first head): partner-wide default with the `patch` kind, "isolate" → one row, `run_script` "Approval request" + note, "Up to 1 device per run", "Partner Technician is asked", script-gate sentence under Executes unattended, "1. Purpose and posture" on the completed circle.

Issue #5075, opened 2026-09-06
Organization record page + Service Management mode


ToddHebebrand:
Tracking issue for **Organization record page + Service Management mode** (`organization-record-page`).

- Plan: `docs/superpowers/plans/2026-09-06-organization-record-page.md`
- Spec: `docs/superpowers/specs/web-ui/2026-09-06-organization-record-page-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All four waves merged via the merge queue: W01 #5086, W02 #5109 (43c058b0f5), W03 #5110, W04 #5108 (b3d528d869). Unreleased; ships in the next release.

Not verified end-to-end in a browser: the Service Management mode switch (native/external/off) against a real PSA connection, and the org record tabs on a seeded stack. Covered by unit, integration (live-DB erasure + partner purge un-wire, RLS), and e2e suites in CI only.

PR #5086, opened 2026-09-06
feat(web,api): organization record page shell + org-pinned fetching (W01 of #5075)

### LINKED ISSUE #5076: W01: Record page shell, orgIdOverride fetching, summary endpoint, list affordances
Wave W01 of #5075 — Record page shell, orgIdOverride fetching, summary endpoint, list affordances
Branch: `feature/5075-organization-record-page/wave-5076`

ToddHebebrand:
## Summary

Wave W01 of #5075: an organization gets a URL-pinned record page at `/organizations/<id>` — identity header plus a working Overview tab — reachable from the organizations list and from the device breadcrumb. The page's org comes from the URL, never from the OrgSwitcher, so a tech can read one customer's record while their workspace sits on another (or on All organizations) without the two silently mixing.

Only the Overview tab is real in W01; the other tab ids render a short placeholder and are replaced in W02/W03. Their hashes are already routable, so a deep link saved today keeps working.

No migration, no schema change, no new tables — so no RLS, cascade or export-policy registration applies.

## What each task delivered

**1.1 — `orgIdOverride` on `fetchWithAuth`** (`stores/auth.ts`)
`orgIdOverride` (string = force this org, `null` = inject nothing, `undefined` = today's ambient behaviour) plus an exported, unit-testable `applyOrgId`. A URL that already names a *different* org than the override throws instead of silently picking one — a wrong-tenant read deserves a loud failure. The custom option keys are now destructured **out** before every native `fetch`, including the post-401 retry: they used to ride along harmlessly, and an unknown key in a `RequestInit` is silently ignored by the platform, so leaking `orgIdOverride` would have failed invisibly. `lib/api/catalog.test.ts` and `stores/orgStore.scope.test.ts` pass unchanged.

**1.2 — `org-record` route kind** (`lib/routeScope.ts`, `lib/orgSwitch.ts`, `layout/ContextScopeLine.tsx`)
The record is neither `org-or-all` (it never aggregates) nor `org-required` (fleet view must not block it). Its own kind means: no scope line, and a context switch navigates up to `/settings/organizations` rather than reloading a customer the user just switched away from. `isGlobalScopeRoute` is deliberately unchanged — the record does **not** suppress orgId injection globally; it pins per request. `applyOrgSwitch` gained an optional `destination` so "Work in this org" can hand the user their new workspace's dashboard.

**1.3 — `orgStore.Organization.status`**
Dropped `inactive` (a partner/SSO-provider spelling the org enum never had), added `churned` and `offboarding`.

**1.4 — `GET /orgs/organizations/:id/summary`** (`routes/orgSummary.ts`, mounted in `index.ts`)
Same middleware chain as `GET /organizations/:id` (`requireScope('partner','system')` + `requireOrgRead`), same UUID-shape pre-check → 404, same `canAccessOrg` → 404 so a malformed or inaccessible id is never an existence oracle. Each optional section is present only when the caller holds the matching read grant — omission, not a zero, is what a missing permission looks like. Mounted as a sibling router under `/orgs` (the `orgArchive` pattern) rather than growing `orgs.ts`. All queries run in the request's normal DB context; no system context, no bare pool.

**1.5 — Record shell, header, Overview tab** (`components/organizations/record/*`, `pages/organizations/[id].astro`, `locales/*/organizations.json`)
Every request goes through `makeOrgFetch(orgId)`. Three states get their own card because they mean genuinely different things: an org-scoped sign-in cannot use the page at all (the API is partner/system scope by design, so the page never fires a request that would 403); a `suspended`/`churned` org 404s the record GET even though the partner owns it, so the list's cached row supplies the identity; an archived org is readable but unwritable, so write affordances come off and Restore appears (via `runAction`). An Overview tile whose summary section is absent is not rendered — a "0" there would be a confident lie about a customer's fleet.

**1.6 — List affordances + breadcrumb** (`settings/OrganizationsPage.tsx`, `devices/DeviceDetailPage.tsx`, `locales/*/settings.json`)
Org name and a persistent row-end chevron link to the record; the detail pane gains **Open record** and its Edit button is relabelled **Settings**, which is what it actually opens. The status pill now renders only when the status is **not** `active` — a pill on every row made the eye filter past the steady state to find the rows that need attention. The device breadcrumb's organization crumb points at the record.

## Test evidence

```
pnpm --filter @breeze/web exec tsc --noEmit            → clean
pnpm --filter @breeze/api exec tsc --noEmit            → clean
pnpm --filter @breeze/web lint                          → clean
pnpm --filter @breeze/api lint                          → clean

cd apps/web && npx vitest run \
  src/stores/auth.orgIdOverride src/lib/routeScope src/lib/orgSwitch \
  src/components/layout/ContextScopeLine src/components/organizations/record \
  src/components/settings/OrganizationsPage src/lib/api/catalog \
  src/stores/orgStore.scope src/lib/i18n/localeParity \
  src/lib/__tests__/no-silent-mutations
                                                        → 21 files, 569 tests, all pass

cd apps/web && npx vitest run src/stores/auth.test.ts   → 111 tests pass (no regression)
cd apps/api && npx vitest run src/routes/orgSummary     → 6 tests pass
```

`apps/api/src/__tests__/integration/orgSummary.integration.test.ts` **did run against a real Postgres** (2/2 pass: device/alert/ticket/site aggregation on seeded rows, and a second partner's token getting 404). No shared test DB was usable, so an isolated per-worktree stack was brought up via `pnpm test-stack up`, migrated fresh, used, and torn down.

Every new assertion was watched fail first. Two worth calling out:
- The `applyOrgId` / option-stripping suite was 13/13 red before the change.
- The record page's first test run **hung the vitest worker** rather than failing — `useLatest` returned a fresh object each render, so the loader re-fired on every state update it caused. `useLatest` now returns a memoized object and a regression test pins that stability.

## Deviations

1. **`TAB_PERMISSION` is `Record<tab, ReadonlyArray<Grant>>` (any-of), not the plan's `Grant | null`.** Contracts & Billing stacks contracts, invoices and quotes; a user holding only one of those reads still has something to see there, and a single-grant shape could only express that by picking one and hiding the tab from half the users who can use it. An empty array means "no extra grant beyond the `organizations:read` the page already requires" — every tab has an entry, pinned by a test, so a tab added later cannot become ungated by omission.
2. **`external` mode keeps the Tickets tab and hides only Contracts & Billing.** The plan's one-liner said `off` hides both and said nothing about `external`; the spec (Part 2) is explicit that `external` shows shadow ticket rows with "Open in \<PSA\>" links while Contracts & Billing moves to the PSA. Followed the spec.
3. **`statusColors`/`statusLabelKeys` moved to `lib/orgStatus.ts` and are re-exported from `OrganizationsPage`.** The plan had the record header import them from the page component, which would pull the entire org settings page into the record's bundle. This is the same move `fetchAllOrganizations` already made, for the same stated reason. No behaviour change; `OrganizationsPage.statusMaps.test.tsx` passes unchanged.
4. **The Overview's critical-alerts panel is a compact inline list, not `AlertList({ orgId })`.** `AlertList` has no `orgId` prop and ships its own filter bar, pagination and bulk-action toolbar — wrong furniture for a ten-row overview panel, and adding an `orgId` prop to it would reach outside this wave's file ownership.
5. **`applyOrgSwitch` gained an optional third argument** rather than the record duplicating the context-switch ritual, so there is still exactly one.
6. **API: `portalUsers` is gated on `users:read`**, because the permission catalogue has no `portal` resource and inventing one was out of scope; and the summary router is a plain `Hono()` because no `AppEnv` type exists in this codebase (`orgArchive.ts` does the same).
7. **`OrganizationsPage.test.tsx` does not exist** — this component's tests are already split by topic (`.archive`, `.archived`, `.merge`, …), so the new cases went into `OrganizationsPage.recordLink.test.tsx` following that convention. One existing assertion in `OrganizationsPage.archived.test.tsx` checked for literal "Active" pill text on a restored row; since `active` is now deliberately pill-less, it was rewritten to assert the absence of the `Offboarding`/`Archived` pills and of the archived section — a real check, not a tautology.

## Notes for the reviewer

- `visibleTabs(perms, mode)` already takes `mode` and defaults to `'native'`, so W04 wires the partner's stored mode without touching the shell. Failing open to `native` is deliberate: hiding a module a partner actually runs is worse than showing a tab they turned off.
- Browser verification of the switcher-mismatch chip is not included here; the behaviour is covered by unit tests that assert every request URL carries the record org while the store's `currentOrgId` is a different org.

Closes #5076

https://claude.ai/code/session_0171z2rYcFpJ4zREAHb3ZzNc


ToddHebebrand:
## Review summary — 5 specialist passes, 8 findings fixed

Ran `/pr-review-toolkit:review-pr` with five reviewers in parallel: **code-reviewer**, **pr-test-analyzer**, **silent-failure-hunter**, **comment-analyzer**, **type-design-analyzer**. Every confirmed finding is fixed in this PR (`92d03463b`, `b9e9ed3fe`) rather than deferred.

### Fixed

| # | Sev | Finding | Fix |
|---|---|---|---|
| 1 | High | **The Overview feeds reported failure as emptiness.** A 403 (no `alerts:read`), a 500 or a network error all rendered *"No open critical alerts"* — a confident false negative in a monitoring tool that a tech could act on by not escalating. It also contradicted the principle the tiles in the same file already follow (omit a section rather than show a lying zero). | Feed state is now `rows \| 'failed' \| null`. A failed load says so, offers a retry, and logs. Both feeds still fail independently. |
| 2 | Med | **`lastActivityAt` was ungated.** Computed for anyone with the route-level `organizations:read`, while every other optional section is gated on its own grant and the route's own docblock promised exactly that. Net effect: a caller without `audit:read` got a 403 from the activity feed and, on the same page, a truthful "Last activity \<timestamp\>" from the summary. | Gated on `audit:read`, **omitted** (not nulled) without it. |
| 3 | Med | **The org pin was defeatable.** `OrgFetch` omitted only `orgIdOverride` from its init type, but `skipOrgIdInjection` un-pins a request just as effectively — and is the arm `applyOrgId` takes *first*, so it wins silently. | Both keys omitted from the type, **and** `applyOrgId` now refuses the combination outright, matching how it already refuses a URL/override conflict. |
| 4 | Med | **`MODE_HIDDEN_TABS.external` hand-duplicated the service-management tab list**, so adding a third such tab would leave it silently visible in `external` mode. | Derived from `SERVICE_MANAGEMENT_TABS`; the new test pins the *derivation*, not today's tab ids. |
| 5 | Med | **`OrgOverviewTab` (249 lines) had no dedicated test** — four permission-gated tile branches and the NaN fallback were never rendered by any test. | New `OrgOverviewTab.test.tsx`, 12 cases: every tile branch, the currency/NaN paths, empty-vs-failed for both feeds, cross-feed isolation, retry, and the logging trail. |
| 6 | Med | **The API's permission-omission proof was asymmetric.** Every scenario granting `DEVICES_READ` also granted `CONTRACTS_READ`, so a copy-pasted permission constant could not fail any test. The contacts mock also could not tell the `count(*)` query from the `.limit(1)` primary lookup, and was quietly feeding a garbage primary contact through unasserted. | Disjoint grant set; a `scope: 'system'` case for the branch that skips `canAccessOrg`; direct proof of the `audit:read` gate; query-aware contacts mock plus a "contacts exist, none primary" case. Unit tests 6 → 10. |
| 7 | Med | **Five of nine summary sections never touched real Postgres** — including the invoices query (`COALESCE(SUM(…) FILTER (…), 0)::text` + `MIN(…) FILTER` + `IN (sql.join(…))`), exactly the shape this repo has been bitten by before. | Integration test extended to seed contracts, invoices, contacts, portal users and audit rows, asserting computed values. **This found a real bug** — see below. |
| 8 | Low | Comment defects: both API docblocks cited `#5075 W02` (this is W01 / #5076); the `orgStore` comment claimed `inactive` belongs to the partner enum (it does not — `partnerStatusEnum` has no such value; only SSO providers and the already-drifted frontend `Partner` union do). | Corrected, and the enum is now cited by its real name (`orgStatusEnum`). |

### The bug the integration coverage caught

Adding real-Postgres assertions for `lastActivityAt` surfaced a defect invisible to the mocked suite: the raw `sql` `MIN`/`MAX` aggregates bypass Drizzle's column-type mapping, so Postgres returns offset-less text (`"2026-08-15 12:00:00"`). Per ECMA-262 a bare **date** string parses as UTC but a bare **date-time** string parses as the **host's local time** — so `lastActivityAt` was silently shifted by the server's UTC offset on any non-UTC host. `nextRenewalAt`/`nextDueAt` were unaffected (date-only columns). Naive date-time strings are now normalized to UTC before parsing.

### Accepted as-is, with reasoning

- **`OrgSummary` is declared twice** (web + API) rather than in `packages/shared`. The type reviewer checked what this repo actually does before recommending: local DTO declarations with a `GET /…` docblock are the dominant pattern here (`TicketSummary`, `AlertsSummary`, `InvoiceSummary`, ~9 more); `packages/shared/src/types/` is for genuinely multi-consumer shapes. Promoting a single-consumer-per-side DTO would be *less* consistent, not more. The trigger to move it is a second consumer.
- **`TAB_PERMISSION`'s empty array** means "no grant beyond the `organizations:read` the page already requires". The invariant lives in a comment rather than the type. Left as-is: the server independently re-checks every permission per section, so a client-side gating slip degrades to an empty tile, not a data leak.
- **Archive/Merge modals use bare `fetchWithAuth`**, so those two writes are scoped by the ambient switcher org rather than the record's pin. Verified harmless today — `orgArchive.ts` and `orgMerge.ts` read only `c.req.param('id')` and never consult the query `orgId` — and the modals are outside this wave's file ownership. Noted for W02/W03, when those surfaces are touched anyway.
- **`ContextScopeLine`'s new org-record case is non-discriminating** (an unregistered route also renders nothing). The explicit `if (kind === 'org-record') return null;` guard is kept as defence against a future reclassification, but the assertion is honest about not proving much on its own.

```
pnpm --filter @breeze/web exec tsc --noEmit    → clean
pnpm --filter @breeze/api exec tsc --noEmit    → clean
pnpm --filter @breeze/web lint                 → clean
pnpm --filter @breeze/api lint                 → clean

apps/web: 18 files / ~380 tests across the W01 surface + localeParity + no-silent-mutations → pass
apps/api: src/routes/orgSummary.test.ts → 10 pass
apps/api: orgSummary.integration.test.ts → 2 pass against a real Postgres
          (isolated per-worktree stack via `pnpm test-stack up`, migrated, run, torn down)
```

### Reviewers' own read on the rest

The code reviewer reported **no findings at ≥80 confidence** and explicitly ruled out: tenant isolation through `applyOrgId` (all branches covered, ambient behaviour byte-identical for existing callers), route reachability under Hono's mount order (no wildcards in either sub-router; segment counts differ), hooks-order and render-loop correctness, and CLAUDE.md conventions including programmatic i18n key parity across all 8 locales.

Not merging — this is W01 of #5075 and the call is yours.

PR #5090, opened 2026-09-06
feat(web): network device detail page + unified device list on by default

ToddHebebrand:
## Summary

Network devices become first-class in the Devices area, and the unified device list (agent + network) is **on by default**.

**Network device detail page** (`/devices/network/:id`): header with site, IP, MAC, manufacturer; stat strip (status, ping, open ports, linked device); Overview tab with identity, SNMP data (long fields clamp behind Show more) and open ports read as capabilities (service labels, "Unencrypted" flag, per-port "Open Web UI"); Monitoring tab with SNMP/network monitor status and the discovery link/unlink controls (manual link picker, unlink confirm, auto-link suppression note). Keyboard and screen-reader support throughout (tablist roving focus, non-modal proxy popover with focus restore, live region). API: `GET /discovery/assets/:id` returns `siteName` and `suggestedBridgeDeviceId`.

**Unified device list fixes** from an impeccable critique (17/40 before):
- Filters no longer silently drop network rows. `POST /filters/preview` only knows the agent `devices` table, so network rows are evaluated client-side by `mergedListFilter.ts` against the same condition group (status, hostname, tags, site, IP, MAC, last seen, asset type for the Servers chip). Agent-only fields hide the row *and* the page says so: "N network devices hidden — Needs Patches applies to agent devices only." Segment badges count the rows the list actually renders.
- Bulk bar states the class composition ("6 selected · 2 agent, 4 network"), disables agent-only actions with a reason when no agent is selected, annotates them "2 of 6" on a mixed selection, and prunes the selection when rows leave the visible set.
- One class facet (the page segment); the duplicate inside DeviceList is gone. VPN facet lifted into list filters so counts see it; hidden when no agent row is on screen.
- Sort keys agree with the cells: OS/CPU/RAM sort network rows as blanks, last in both directions.
- Sortable headers are real buttons in `scope="col"` cells, rows keep a focus ring, View buttons are named after their device, dash cells carry "Not applicable".
- Columns adapt to the classes on screen; grid view is filtered, searched and sorted like the table; search matches IPs; table status word matches cards; chip strip stacks on phones with 24px targets; Class/Type headers and copy translated in all 8 locales.

**Flag:** `ENABLE_NETWORK_DEVICES_IN_LIST` defaults to `true`; set `PUBLIC_ENABLE_NETWORK_DEVICES_IN_LIST=false` at web build time to return to the agent-only list. Dev compose override maps it (empty default). Deploy docs updated.

**Also fixed on the way:** the proxy popover closed itself when Connect was pressed (Chrome fires `focusout` with no related target when the pressed button disables); discovery list badge reads "Agent"; device filter bar avoids one-off responsive utilities (Astro's client router keeps the previous page's inline dev stylesheet on Back).

## Review

Four review passes (code, silent failures, tests, comments) on the branch diff; all confirmed findings fixed in 331497cf7 (operator names from the shared `FilterOperator` union, date operators on Last Seen, MAC mapping, VPN in the predicate, selection pruning against the visible set, notice overcount, error paths in the proxy popover and manual link control, stale comments).

## Test plan

- [x] `apps/web`: devices, discovery and i18n suites — 1316 tests pass after merging main; `tsc --noEmit` and eslint clean on changed files.
- [x] Live in a `wt-stack` with 10 seeded discovered assets: detail page flows (type edit/reset, link/unlink, proxy popover incl. TLS-skip on 443, 404 state), list flows (Online chip keeps network rows, Needs Patches shows the hidden notice with honest badges, network-only select-all disables all seven agent actions, segment switch prunes selection, columns adapt, Back navigation keeps the filter bar on one row).
- [x] Verified merged with `origin/main` (main's #5023 Compare-over-cap and removed-hint counts rebuilt on the shared predicate).
- [ ] CI on this PR.

Not in this PR: bulk selection in grid view (feature), global muted-text contrast on card backgrounds (token change), history charts for SNMP/monitor data on the Monitoring tab (product gap: no page charts `snmp_metrics` or `network_monitor_results` today).

https://claude.ai/code/session_012RcTTnexFsn3q6VSZ5tLpP


ToddHebebrand:
**Shipped in v0.111.1**: the unified device list is on by default and the network device page is live. Closing.

PR #5092, opened 2026-09-06
fix(web): keep Integrations on Accounting for nested QuickBooks tab hashes

ToddHebebrand:
## Repro (prod v0.110.0)

Integrations → **Accounting** → QuickBooks → "Customer and item mapping" workbench → click **Items**. The whole page jumps back to the **Webhooks** tab, so the Items mapping table is unreachable. Clicking **Customers** does the same thing (the brief assumed Customers was fine — verified it is not: both workbench tabs knocked the page off Accounting; only the *initial* Customers view survived, because the workbench defaults to Customers without writing the hash).

## Root cause

There were two owners of `window.location.hash`:

- `IntegrationsPage.parseHash` routes the top-level tab (and the security/identity/distributor/accounting sub-tab) from the hash, falling back to `initialTab` (Webhooks) on anything unrecognized.
- `QuickbooksMappingWorkbench` — nested two levels down — uses `useHashTab(["quickbooks-customers", "quickbooks-items"])` and writes `window.location.hash = "quickbooks-items"` on tab click.

`quickbooks-items` matches no `TabId` and no sub-tab id, so the page's `hashchange` handler resolved it to the fallback tab and unmounted the accounting panel (workbench included).

`accountingSubTabs` ids are `quickbooks` / `stripe`, so the Accounting **sub**-tab buttons themselves are fine — they were already exact-matched by `parseHash`. The collision is only with the nested workbench.

## Design chosen

**One hash owner: the page.** `parseHash` now also accepts a hash *namespaced under* an accounting sub-tab id — `<subTabId>-<nested…>` — and routes it to `{ tab: "accounting", accountingSub }`, after the existing exact matches. The workbench keeps owning its own `quickbooks-*` segment and keeps using `useHashTab` (which already ignores hashes outside its own set, so `#accounting` / `#quickbooks` land on Customers).

Why this over the alternatives:

- **Not component state** for the workbench tabs — CLAUDE.md's "URL State in Components" rule keeps client-side tab state in the hash (as `DeviceDetails.tsx` / `OrganizationsPage.tsx` do), and the Items view is worth deep-linking.
- **Not a new `accounting/quickbooks/items` path shape** — every other sub-tab hash on this page is flat (`#huntress`, `#m365`, `#pax8`, `#quickbooks`), and switching one branch to a slash-separated form would break existing `#quickbooks-*` links and read inconsistently. The prefix rule generalizes to any future nested panel under an accounting sub-tab without further page changes.
- Deep links to `#accounting`, `#quickbooks`, `#stripe` are untouched; `#quickbooks-items` now works as a deep link too.

## Test evidence (red → green)

New `describe("IntegrationsPage — nested QuickBooks workbench hash")` in `IntegrationsPage.test.tsx` renders the **real** workbench behind a stubbed `QuickbooksIntegration` (with `fetchWithAuth` mocked) rather than re-implementing the hash contract in the test.

Red, before the fix (`npx vitest run src/components/integrations/IntegrationsPage.test.tsx`):

```
FAIL > stays on the Accounting tab when the workbench Items tab is clicked
AssertionError: expected <div data-testid="stub-webhooks"></div> to be null
FAIL > deep-links straight to the workbench Items tab
AssertionError: expected <div data-testid="stub-webhooks"></div> to be null

Test Files  1 failed (1)   Tests  2 failed | 42 passed (44)
```

Green, after:

```
npx vitest run src/components/integrations/IntegrationsPage.test.tsx \
               src/components/integrations/QuickbooksMappingWorkbench.test.tsx
Test Files  2 passed (2)   Tests  64 passed (64)

npx vitest run src/components/integrations/QuickbooksIntegration.test.tsx src/lib/useHashState
Test Files  2 passed (2)   Tests  29 passed (29)
```

`npx eslint` on the three changed files: clean. `npx tsc --noEmit -p apps/web/tsconfig.json`: clean.

The third new test (`#quickbooks` deep link stays on Customers) passed before and after — it pins the non-regression of the existing sub-tab links.

## Files changed

- `apps/web/src/components/integrations/IntegrationsPage.tsx` — prefix routing in `parseHash`.
- `apps/web/src/components/integrations/QuickbooksMappingWorkbench.tsx` — comment documenting why the tab ids must keep the `quickbooks-` prefix.
- `apps/web/src/components/integrations/IntegrationsPage.test.tsx` — regression suite (edited in place).

No API, schema, or tenancy surface touched.

https://claude.ai/code/session_012aUFjmAQBPfhv1G5wm9Kri

### DIFF
--- a/apps/web/src/components/integrations/IntegrationsPage.test.tsx
+++ b/apps/web/src/components/integrations/IntegrationsPage.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, fireEvent } from "@testing-library/react";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 let scope: "system" | "partner" | "organization" | null = "partner";
@@ -91,6 +91,32 @@ vi.mock("../settings/TdSynnexCatalogPanel", () => ({
 }));
 vi.mock("../settings/TdSynnexEcExpressPanel", () => ({
   default: () => <div data-testid="stub-tdsynnex-ec" />,
+}));
+vi.mock("./StripePaymentsIntegration", () => ({
+  default: () => <div data-testid="stub-stripe-payments" />,
+}));
+
+// The Accounting tab hosts the QuickBooks mapping workbench, which owns a
+// NESTED tab hash (#quickbooks-customers / #quickbooks-items) inside the page's
+// own hash. Render the REAL workbench behind a stubbed QuickbooksIntegration so
+// the nested-hash contract is exercised end to end rather than re-implemented
+// in the test.
+vi.mock("./QuickbooksIntegration", async () => {
+  const { default: QuickbooksMappingWorkbench } = await import(
+    "./QuickbooksMappingWorkbench"
+  );
+  return {
+    default: () => (
+      <div data-testid="stub-quickbooks">
+        <QuickbooksMappingWorkbench defaultIncomeAccountRef="income-1" />
+      </div>
+    ),
+  };
+});
+const { fetchWithAuthMock } = vi.hoisted(() => ({ fetchWithAuthMock: vi.fn() }));
+vi.mock("../../stores/auth", async (importActual) => ({
+  ...(await importActual<typeof import("../../stores/auth")>()),
+  fetchWithAuth: (...args: unknown[]) => fetchWithAuthMock(...args),
 }));
 
 // Capture the help-panel open() call so we can assert the per-tab docs link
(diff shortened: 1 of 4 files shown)


ToddHebebrand:
### Test Web fix (second commit, `84c4de74e`)

The **Test Web** failure on `559192700` was **not** from this PR's tests — it was a pre-existing main breakage that this branch inherited.

**Failing test:** `src/components/organizations/record/OrganizationRecordPage.test.tsx > permission gating > hides the Devices tab from a user without devices:read`

```
TestingLibraryElementError: Unable to find an accessible element
with the role "button" and name "Overview"
Test Files  1 failed | 802 passed (803)
```

**Root cause:** #5090 (`19d7d37b3`, network device detail page) reworked `components/shared/OverflowTabs.tsx` to the ARIA tabs pattern — visible tabs now render `role="tab"` inside a `role="tablist"` nav. The org-record test, added by #5086 (`b3e025816`), still queries them as `role="button"`. The accessible-roles dump in the CI log confirms it: Overview is present and correct, under `tab:` with `role="tab"`, `aria-selected="true"`, `id="tab-overview"` — only the query was stale.

The negative assertion on the same lines had also gone **vacuous**: `queryByRole('button', { name: 'Devices' })` matched nothing whether or not the Devices tab was gated, so the permission gate was no longer actually under test.

**Why it reached main:** #5090's CI ran before #5086 merged, #5086's ran before #5090 merged, and main's own CI runs for both were cancelled — neither ever tested the combination. Reproduced locally on this branch (`1 failed | 14 passed`), and there is no code path from my three `components/integrations/` files to `OrganizationRecordPage`.

**Fix:** query both tabs by `role="tab"`, which restores the real gate assertion.

Swept the other four `OverflowTabs` consumers' suites for the same stale pattern — the only other `queryByRole('button')` negative (`ConfigPolicyDetailPage`, "Remove configuration") targets a genuine button, not a tab.

**Evidence:**

```
npx vitest run src/components/organizations/record \
               src/components/shared/OverflowTabs.test.tsx \
               src/components/integrations/IntegrationsPage.test.tsx \
               src/components/integrations/QuickbooksMappingWorkbench.test.tsx
Test Files  8 passed (8)   Tests  138 passed (138)
```

`npx eslint` on the changed file and `npx tsc --noEmit -p apps/web/tsconfig.json`: both clean. No locale key changed, so no i18n parity impact.

https://claude.ai/code/session_012aUFjmAQBPfhv1G5wm9Kri

PR #5095, opened 2026-09-06
fix(agent,web): accept interior spaces in Windows service names; surface failed service commands (#5088)

### LINKED ISSUE #5088: Remote Tools Services: start/stop/restart fails for any Windows service with a space in its name ("service name contains invalid characters")
## Summary

Remote Tools → Services offers every Windows service in the list, but start/stop/restart fails for any service whose name contains a space. The agent rejects the command with `service name contains invalid characters`, and the API surfaces it as a 500 on `POST /api/v1/system-tools/devices/:deviceId/services/:name/restart`.

Windows service names with spaces are common (`Mesh Agent`, `Bonjour Service`, `Sophos Endpoint Defense Service`, `Veeam Backup Service`, many vendor services). The validation intended to block path/control characters also blocks the space.

## Environment

- Breeze v0.110.0 self-hosted, agent 0.110.0 on Windows Server 2019 Standard

## Reproduce

1. Open a Windows device → Remote Tools → Services.
2. Search for a service with a space in its name (here: `Mesh Agent`, the Tactical RMM / MeshCentral agent).
3. Click Restart Service, confirm.

## Observed

`device_commands` row:

```
type     | restart_service
payload  | {"name": "Mesh Agent", "timeoutSeconds": 30}
status   | failed
result   | {"error": "service name contains invalid characters", "status": "failed", "exitCode": 1}
```

API log:

```
Command 197be6b0-… failed for agent 0a0fca03…
--> POST route=/api/v1/system-tools/devices/:deviceId/services/:name/restart status=500 duration_ms=638
```

The UI shows no error toast; the row just stays "Running".

## Cause

`agent/internal/remote/tools/services.go`, `validateServiceName`:

```go
if strings.Contains(trimmed, "..") || strings.ContainsAny(trimmed, "/\\\x00\r\n\t ") {
    return "", fmt.Errorf("service name contains invalid characters")
}
```

The trailing space in the `ContainsAny` set rejects interior spaces. Leading/trailing whitespace is already handled separately two lines above, so the space in this set only hits legitimate names.

## Expected

- Interior spaces accepted (Windows `sc`/SCM allows them; the name is passed to the service manager API, not a shell, so a space is not an injection vector).
- If the agent really must reject the name, the UI should not list it as actionable, or should show the error instead of silently failing.

Workaround used: Remote Tools → Terminal → `Restart-Service 'Mesh Agent'`, which works.

ToddHebebrand:
## Summary

- `agent/internal/remote/tools/services.go`: `validateServiceName`'s reject set (`ContainsAny`) included a literal space, so any Windows service with an interior space in its name (`Mesh Agent`, `Bonjour Service`, `Sophos Endpoint Defense Service`, ...) was rejected with `service name contains invalid characters`. Leading/trailing whitespace is already rejected separately two lines above, and the name is passed to the Windows Service Control Manager API, not a shell, so an interior space is not an injection vector. Dropped the space from the reject set.
- `apps/web/src/components/remote/RemoteToolsPage.tsx`: the Services panel's start/stop/restart handlers threw a plain `Error` on a failed command; `ServicesManager`'s `handleAction` caught it and only `console.error`'d, so a failed command produced no toast and the row was never re-fetched — indistinguishable from the click doing nothing. The three handlers now go through `runAction` (which toasts the failure) and always re-fetch the service list afterward regardless of outcome, so the row reflects real server state instead of a stale pre-action snapshot.

The API's `POST .../services/:name/restart` route (`apps/api/src/routes/systemTools/services.ts`) already classifies agent command failures via `buildCommandFailureResponse`/`classifyCommandFailure`; an unrecognized agent refusal deliberately falls back to a 500 (`agent_execution_failed`) per the documented comment in `fileBrowserHelpers.ts` ("the safe direction ... over-reports a fault rather than silently claiming the user can fix something they cannot"). That mapping is intentional, not a bug, and is untouched here — it's also moot for the reported repro, since a real interior-space service name no longer gets rejected by the agent in the first place.

Closes #5088

## Test plan

- [x] `cd agent && go test -race ./internal/remote/tools/...` — full package green, including a red-first table-driven `TestValidateServiceName` rewrite (`"Mesh Agent"` etc. accepted; tab/newline/CR/null-byte/slash/backslash/`..`/leading-trailing-space still rejected).
- [x] `cd apps/web && npx vitest run src/components/remote/RemoteToolsPage.test.tsx` — 13/13 passed, including a new red-first regression test asserting a failed restart command shows an error toast and re-fetches the service list.
- [x] `cd apps/web && npx vitest run src/lib/__tests__/no-silent-mutations.test.ts` — 139/139 passed (the `runAction` adoption doesn't need an allowlist entry).
- [x] `cd apps/web && npx tsc --noEmit` — clean.
- [x] `cd apps/api && npx vitest run src/routes/systemTools.test.ts src/routes/systemTools` — 171/171 passed (no API changes; confirms no regression).
- [x] `go build ./...` (agent) and `go vet` on the touched package — clean.
- [x] `npx eslint` on both touched web files — clean.


ToddHebebrand:

**Findings:** 3 raised → 2 addressed in 5d7e6cc38 (test's mocked error message was byte-identical to the `runAction` fallback string, making the assertion non-discriminating between "server message plumbed through" and "silently fell back"; and start/stop handlers had zero test coverage despite using a different try/finally shape than restart). 1 left as a documented follow-up:

- **MEDIUM (silent-failure-hunter, not blocking):** this fix makes `fetchServices()`'s pre-existing silent `catch { console.error(...) }` (unchanged, `RemoteToolsPage.tsx:592-593`) load-bearing in a new way — it's now reached on the failure path too, not just success. If that re-sync itself fails right after the mutation failure, the user gets the mutation-failure toast but no signal that the row may still be stale. Pre-existing code, low probability, not introduced by this PR — recommend a fast follow-up issue rather than blocking here.

code-reviewer raised 0 findings (verified the Go fix doesn't create an injection vector across all three platform implementations, and the `runAction` adoption matches CLAUDE.md's Web Mutation Handlers convention).

**Tests:** `cd apps/web && npx vitest run src/components/remote/RemoteToolsPage.test.tsx` — 15/15 passed (3 parametrized start/stop/restart failure-path cases, each independently verified to fail against the pre-fix code checked out from `origin/main`, confirming they're not vacuous). `cd agent && go test -race ./internal/remote/tools/...` — full package green (40s), including the rewritten table-driven `TestValidateServiceName`. `apps/web` `tsc --noEmit` clean. `apps/api` `vitest run src/routes/systemTools.test.ts src/routes/systemTools` — 171/171 passed (no API changes; confirms no regression). `eslint` clean on both touched web files.

PR #5096, opened 2026-09-06
fix(agent): decode UniFi Integration API camelCase fields in deep telemetry (#5087)

### LINKED ISSUE #5087: UniFi deep telemetry: agent decodes controller response with snake_case tags, so mac/ip/uptime/PoE are always empty
## Summary

The agent-side UniFi deep telemetry collector decodes the controller's `/proxy/network/integration/v1/sites/{id}/devices` and `/clients` responses into structs whose JSON tags are snake_case (`mac`, `uptime_seconds`, `num_clients`, `hostname`, `ip`, `is_wired`), but the UniFi Network Integration API returns camelCase (`macAddress`, `ipAddress`, `firmwareVersion`, `name`, `type`, `uplinkDeviceId`). Only `id` and `name` match, so every other field arrives at the API as zero/empty. Devices and clients are stored with an empty `mac`, which also defeats the MAC-based enrichment of `discovered_assets`.

The unit tests do not catch this because the fake controller in `client_test.go` responds with the same snake_case shape the struct expects, not the shape the real controller sends.

## Environment

- Breeze v0.110.0 self-hosted (docker compose, `guided-setup.sh`), agent 0.110.0 on Ubuntu 26.04 amd64
- Cloud (Site Manager) connection + deep telemetry collector pointed at a UDM (Network app 9.x, Integration API enabled)
- Collector status: `connected`, `firmware_ok = true`, `last_poll_status = success`, polls every ~60 s

## What is stored

`unifi_device_telemetry` after a successful poll (25 devices, all identical pattern):

| name | mac | uptime_seconds | num_clients | poe_ports | raw (excerpt) |
|---|---|---|---|---|---|
| SW i13 Main | *(empty)* | 0 | 0 | *(null)* | `{"id":"217b8bfb-…","name":"SW i13 Main","model":"US 48 PoE 500W","state":"ONLINE","ipAddress":"172.16.x.x","macAddress":"44:d9:e7:xx:xx:xx","firmwareVersion":"7.4.1","firmwareUpdatable":false,"features":["switching"],"interfaces":["ports"],"supported":true}` |

`unifi_clients` (1 row):

| hostname | ip_address | is_wired | mac | raw (excerpt) |
|---|---|---|---|---|
| *(empty)* | *(empty)* | false | *(empty)* | `{"id":"0616182e-…","name":"PrinterDirectie-2 be:5e","type":"WIRED","access":{"type":"DEFAULT"},"ipAddress":"172.16.x.x","macAddress":"f4:a9:97:xx:xx:xx","connectedAt":"2026-08-29T06:36:43Z","uplinkDeviceId":"116f5b2d-…"}` |

`raw` is the verbatim element from the controller response, so the data is there; the typed columns are not populated. Note the client is wired but `is_wired` is `false` (the API expresses this as `"type":"WIRED"`, not `is_wired`).

## Code

`agent/internal/unifi/client.go` (v0.110.0, unchanged on `main` today):

```go
type Device struct {
	ID            string          `json:"id"`
	Mac           string          `json:"mac"`
	Name          string          `json:"name"`
	UptimeSeconds int64           `json:"uptime_seconds"`
	CPUPct        float64         `json:"cpu_pct"`
	MemPct        float64         `json:"mem_pct"`
	TxBytes       int64           `json:"tx_bytes"`
	RxBytes       int64           `json:"rx_bytes"`
	NumClients    int             `json:"num_clients"`
	PoePorts      []PoePort       `json:"poe_ports"`
	SiteID        string          `json:"site_id"`
	Raw           json.RawMessage `json:"raw"`
}
```

`agent/internal/unifi/client_test.go` fixture that lets the test pass against a shape the controller never sends:

```go
w.Write([]byte(`{"data":[{"id":"d1","mac":"aa:bb","name":"AP","uptime":10,"num_clients":1}]}`))
...
w.Write([]byte(`{"data":[{"mac":"cc:dd","hostname":"phone","ip":"10.0.0.9","is_wired":false}]}`))
```

The API side already reads the IP from `raw` (`deviceIp(device.raw)` in `unifiTelemetryService.ts`), so the server knows the raw shape is camelCase; only the agent mapping was not aligned.

## Second, related gap

`client.go` only calls `/sites`, `/sites/{id}/devices` and `/sites/{id}/clients`. The list endpoint does not carry uptime, CPU/memory, throughput or per-port PoE; those come from `/sites/{id}/devices/{deviceId}/statistics/latest` (and the device detail endpoint for port/PoE configuration). Those endpoints are never requested, so `uptime_seconds`, `cpu_pct`, `mem_pct`, `tx_bytes`, `rx_bytes` and `poe_ports` cannot be filled even after the tag fix. The docs (`features/unifi-integration.mdx`) describe per-port PoE and uptime as what deep telemetry provides.

## Expected

- `Device`/`Client` decode the Integration API field names (`macAddress`, `ipAddress`, `type == "WIRED"`, `uplinkDeviceId`, client `name` → hostname), or the collector maps from `raw`.
- Test fixtures use a real controller response shape.
- Per-device statistics fetched so the documented uptime/PoE/client-count telemetry is populated.

Happy to test a fix against this UDM.

ToddHebebrand:
Closes #5087

## What was wrong

`agent/internal/unifi/client.go` decoded the UniFi **Network Integration API** list responses with snake_case JSON tags while the controller returns camelCase. Only `id` and `name` bound; everything else decoded as a zero value, so devices and clients landed in `unifi_device_telemetry` / `unifi_clients` with an empty `mac` — which also defeated the MAC-based `discovered_assets` enrichment.

The tests passed because the fake controller in `client_test.go` replied with **the struct's own snake_case shape**. The fixture could never disagree with the tags it was validating.

## Field mapping

Verified against the published `getDeviceOverviewPage` / `getConnectedClientOverviewPage` schemas **and** the reporter's verbatim `raw` payload from a live Network 9.x controller.

| Struct field | Was | Now |
|---|---|---|
| `Device.Mac` | `mac` | `macAddress` |
| `Client.Mac` | `mac` | `macAddress` |
| `Client.Hostname` | `hostname` | `name` |
| `Client.IP` | `ip` | `ipAddress` |
| `Client.ConnectedDeviceID` | `connected_device_id` | `uplinkDeviceId` |
| `Client.IsWired` | `is_wired` (bool field) | derived from `type` |

`is_wired` is not a field the API has at all — wired-ness is the `type` discriminator (`WIRED` / `WIRELESS` / `VPN` / `TELEPORT`). `IsWired()` is now a **method** over the decoded `Type`, so it cannot desync from it; VPN/Teleport, an unknown future enum value, and an absent `type` all degrade to not-wired, each pinned by a test.

**Unchanged on purpose:** the outbound `uploadDevice`/`uploadClient` DTOs already speak the Breeze ingest contract, and `PoePort`'s snake_case tags are the *Breeze* contract, not the controller's — `poePorts` passes through the API opaquely (`z.unknown()`, no transform) and the web UI reads `p.up` / `p.poe_power_w` verbatim. Both now carry comments saying so.

## Fields deliberately left unpopulated

Not on the list endpoints, so tagged `json:"-"` rather than given an invented tag:

| Field | Source if implemented |
|---|---|
| `Device.UptimeSeconds`, `CPUPct`, `MemPct`, `TxBytes`, `RxBytes` | `/devices/{id}/statistics/latest` → `uptimeSec`, `cpuUtilizationPct`, `memoryUtilizationPct`, `uplink.txRateBps`/`rxRateBps`. Those tx/rx are **rates**, not the cumulative byte counters these fields model — a unit decision, not just a tag. |
| `Device.PoePorts` | device **detail** endpoint, `interfaces.ports[].poe` |
| `Device.NumClients` | not exposed at all; derivable by counting clients per `uplinkDeviceId`, but see pagination below |
| `Client.SSID`, `Vlan`, `SignalDbm`, `UplinkPortIdx`, `TxBytes`, `RxBytes`, `UptimeSeconds` | no Integration API source found |

Fetching the per-device statistics/detail endpoints is the **second half** of #5087 (an N+1 call per device per poll) and is not folded in here.

**These no longer upload as literal `0`.** They were being sent without `omitempty`, so the API stored `0` instead of `NULL` and the web UI's `d.numClients ?? "—"` never fired — every switch and AP rendered "0 clients" as if measured. That is the same silent-zero disease as the headline bug, one hop downstream, and `numClients` is on screen today. They are now omitted, and the API persists an absent metric as `NULL` explicitly (a bare `undefined` is dropped from a drizzle `onConflictDoUpdate` SET, which would have preserved a stale `0` from an older agent). `omitempty` only carries this meaning while the fields are never collected — the struct says so, and says to switch to pointers when the statistics fetch lands.

## Also fixed: the enrichment path this change puts into service

`linkTelemetryDeviceToAsset` compared `discovered_assets.mac_address` against the raw agent value with **no normalization on either side**, unlike the client path. Because `mac` was always empty until now, that branch was dead in production — this PR is what turns it on, so the gap is in scope. A controller reporting uppercase or hyphenated MACs would have silently failed to link, and the enrich write would have put a non-canonical MAC into the column every other producer matches against. Both sides are now canonicalised through a shared `canonicalAssetMac` fragment used by the device and client paths alike. No DB CHECK enforces MAC format anywhere, so the app layer is the only guard.

## Known defects flagged, not fixed

- **Pagination.** Both list endpoints return `offset`/`limit`/`count`/`totalCount`, and `envelope` reads only `data`, so a site larger than the page size is silently truncated. Pre-existing and orthogonal; documented on `envelope`. A detection-only half-measure (compare `totalCount` to `len(data)` and raise a collector error) was considered and **rejected**: it would turn collectors red for large sites with no remedy available until the real offset loop ships. Worth a follow-up issue for the full fix.
- **`unifiSyncService.ts`** (the Site Manager cloud-sync path, a different producer of `discovered_assets.mac_address`) still compares and writes `device.mac` uncanonicalised. Not activated by this change and fed by a different API, so it is out of scope — but it is the same bug class. This PR is not undermined by it, because it normalizes both sides of every comparison rather than assuming the column is clean.

## Verification

- **Red-first on every behavioural change**, each failure observed before the fix: the camelCase fixture produced exactly the reported symptom on unmodified code (empty device Mac; empty client Mac/Hostname/IP/ConnectedDeviceID; `IsWired` false); the asset-link test returned `null` instead of `asset-9`; the metrics test showed all six fields uploading as `0`; and the stored-non-canonical-MAC test was confirmed red by temporarily reverting `canonicalAssetMac`.
- `cd agent && go test -race -count=1 ./internal/unifi/...` — **12 test funcs / 14 with subtests, all pass**. `go build ./...`, `go vet`, `gofmt` clean.
- `apps/api` UniFi suites, 11 files / **134 tests pass**. `npx tsc --noEmit` clean.

Not verified against a live controller; the reporter offered to test against their UDM.


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer (all three on the diff, read-only).

**Findings:** 7 raised → 2 fixed in `9503a8556`, 5 declined with reasons; 0 outstanding.

Fixed:
- **Uncollected metrics uploaded as literal `0`, not omitted** (silent-failure-hunter, HIGH). `uptimeSeconds`/`cpuPct`/`memPct`/`txBytes`/`rxBytes`/`numClients` had no `omitempty`, so the API stored `0` instead of `NULL` and the UI's `d.numClients ?? "—"` never fired — every switch and AP showed "0 clients" as if measured. Same silent-zero disease as the headline bug, one hop downstream, and `numClients` is on screen today. Now omitted, with the API persisting absent metrics as `NULL` explicitly (a bare `undefined` is dropped from a drizzle `onConflictDoUpdate` SET and would have preserved a stale `0` from an older agent).
- **The device-MAC test only covered half the fix** (pr-test-analyzer, must-fix). The bound parameter is canonicalised in JS before it reaches the query, so the assertion held whether or not the *lookup* canonicalised the stored column — reverting `canonicalAssetMac` kept it green. The `scriptedDb` mock now models the SQL (it canonicalises its own keys only when the where clause actually wraps the column in `lower(replace(...))`), plus a new test whose *stored* row is the non-canonical side. Confirmed red against a reverted fix before restoring.

Declined, with reasoning in the PR body: pagination **detection**-only (would turn collectors red for large sites with no remedy until the real offset loop ships — the full fix is the right follow-up); logging on unknown client `type` (would plumb a logger through `APIClient` for a required API field; pinned by tests instead); observability counters on the null-link paths (pre-existing, disproportionate); `unifiSyncService.ts`'s uncanonicalised MAC (separate producer, different API, not activated here); golden-file fixture snapshots (replaced with a cheaper guard test that fails loudly if the fixtures are ever rewritten toward our own field names — the anti-pattern that hid this bug).

Also added from the review: TELEPORT / unknown-enum / absent `type` cases, and a device reporting a blank mac storing `NULL` rather than `''`.

**Tests:** `go test -race -count=1 ./internal/unifi/...` — 12 funcs / 14 with subtests, green; `gofmt`/`go vet` clean. `apps/api` UniFi suites 11 files / 134 tests green; `tsc --noEmit` clean. Every behavioural change was observed red first.

**Status:** review-clean, awaiting maintainer merge. Not verified against a live controller — the reporter (@meckyp) offered to test against their UDM, which is the one gap worth closing before this is called confirmed.

PR #4637, opened 2026-09-02
fix(integrations): send epoch-millisecond from/to to the Umbrella reporting API (#4597)

ToddHebebrand:
## Summary

Cisco Umbrella's Reporting API v2 `security-activity` endpoint rejects ISO 8601 timestamps for the `from`/`to` query params:

```
{"errors":[{"param":"from","value":"2026-08-14T00:30:00.151Z","error":"invalid timestamp specified"},{"param":"to","value":"2026-08-15T00:30:00.151Z","error":"invalid timestamp specified"}]}
```

Per Cisco's OpenAPI spec for this endpoint, `from`/`to` must be Unix epoch milliseconds as a numeric string (e.g. `"1639146300000"`), not ISO 8601.

`UmbrellaProvider.syncEvents` (`apps/api/src/services/dnsProviders/umbrella.ts:137-138`) was sending `since.toISOString()`/`until.toISOString()`. Switched both to `String(since.getTime())`/`String(until.getTime())`. Grepped the rest of the file for other `toISOString()` call sites used as request params — this was the only one.

This is a follow-up to the OAuth2 client-credentials fix in #3275: with that live, auth succeeds but the sync call itself was still failing on the timestamp format.

Closes #4597

## Test plan

- [x] Added a red-first regression test asserting `from`/`to` are numeric epoch-ms strings, watched it fail against the pre-fix code (`AssertionError: expected '2026-08-14T00:30:00.151Z' to be '1786667400151'`), then applied the fix and watched it pass.
- [x] `cd apps/api && npx vitest run src/services/dnsProviders` — 5 files, 67 tests, all passed.
- [x] `cd apps/api && npx tsc --noEmit -p .` — clean, no errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Xr3ZNTFRkCM1t4gEjPttTg

Issue #4597, opened 2026-09-02
issue related to #3271


cisspUser01:
### Description

Follow-up: with the OAuth2 fix from https://github.com/LanternOps/breeze/pull/3275 live in 0.105.1 and a fresh "Umbrella API Keys" (modern, scoped) credential, authentication now succeeds — but sync still fails, reproduced consistently across multiple attempts:

{"errors":[{"param":"from","value":"2026-08-14T00:30:00.151Z","error":"invalid timestamp specified"},{"param":"to","value":"2026-08-15T00:30:00.151Z","error":"invalid timestamp specified"}]}

Per Cisco's own OpenAPI spec for this endpoint (umbrella-api-reference-reports-reporting-api-activity-get-activities-all), from/to should be Unix epoch milliseconds as a numeric string (e.g. "1639146300000"), not ISO 8601. UmbrellaProvider.syncEvents appears to send since.toISOString()/until.toISOString() — likely just needs String(since.getTime())/String(until.getTime()) instead.

### Steps to Reproduce

none

### Expected Behavior

none

### Actual Behavior

none

### Component

Web Dashboard

### Relevant Logs

```shell

```

ToddHebebrand:
**Root cause:** confirmed at `apps/api/src/services/dnsProviders/umbrella.ts:137-138`. `syncEvents` sends `from` and `to` as ISO 8601 strings, and Cisco's reporting API rejects those with exactly the "invalid timestamp" error you pasted. It wants epoch milliseconds.

**Fix:** switch both parameters to `String(since.getTime())` / `String(until.getTime())`. Queued, ships in the next release. Good catch pinning the exact call site.

ToddHebebrand:
**Status:** fixed on main by #4637 (`from`/`to` now sent as epoch milliseconds). Reopening so you can verify the Umbrella sync after it ships in the next release; close it once a sync completes on your install.

cisspUser01:
[DNS Security] UmbrellaProvider targets the retired "Reporting v2 API" (reports.api.umbrella.com, EOL Sept 2023) — real endpoint is api.umbrella.com/reports/v2/activity, org-scoped via token, no orgId path segment

Follow-up to #3271/#3275/#4637. Auth (#3275) and timestamp format (#4637) are both correctly
fixed in v0.110.0 — confirmed directly against the live source. But sync still fails with 404
"no Route matched with those values" from Cisco's own gateway.

Root cause: UmbrellaProvider.syncEvents targets
[...]

PR #5097, opened 2026-09-06
fix(dns-security): move Umbrella sync to the next-gen /reports/v2/activity endpoint (#4597)

ToddHebebrand:
Refs #4597

Umbrella sync still 404s at Cisco's own gateway after the OAuth2 (#3271) and epoch-ms (#4637) fixes, because `syncEvents` targets the **legacy Reporting v2 host** `https://reports.api.umbrella.com/v2/organizations/{orgId}/security-activity` — retired by Cisco with an EOL of September 2023. #4637 fixed the timestamp format against the spec for a route that no longer exists.

This moves the sync to the current endpoint, `https://api.umbrella.com/reports/v2/activity`, and re-maps the response against the next-gen schema.

## `organizationId` is now optional

The next-gen endpoint has no org path segment — the OAuth2 token's own `sub` claim (`org/<orgId>/client/<apiKey>`) scopes every call. So:

- `POST /dns-security/integrations` no longer rejects an Umbrella integration that omits `config.organizationId` (`apiSecret` is still required — covered by a control test).
- The provider no longer reads it. It stays on the config type and is still accepted/stored, so pre-existing integrations are untouched.

## Field mapping

| `DnsEvent` | next-gen `data[]` field | notes |
|---|---|---|
| `timestamp` | `timestamp` | epoch **milliseconds**, not the legacy `datetime` ISO string |
| `domain` | `domain` | |
| `queryType` | `querytype` | lowercase, no underscore |
| `action` | `verdict` | `blocked` → `blocked`; `proxied` → `redirected`; otherwise `allowed` |
| `category` | `categories[].label` | prefers the **security**-typed entry — the content-typed one ("Business Services") is usually listed first and would normalize to `unknown` |
| `threatType` | `threats[].label` | |
| `sourceIp` | `internalip` → `externalip` | |
| `sourceHostname` | `identities[].label` | `identities[].type` is itself an object, so it is only read when it is a string |
| `providerEventId` | *(none)* | the dns/proxy schemas document no per-record id, so the sync job's deterministic fallback id stays the dedupe key |
| `metadata` | `categories[].label` + raw `verdict` | |

Tolerant parsing (`asRecord`/`asString`/`asNumber`) is kept throughout: a labelled list is also accepted as a plain string array, and an ISO timestamp still parses, so a shape mismatch drops one record instead of throwing.

## Three further consequences of the next-gen contract

- **Pagination is limit/offset only** — no cursor, no page token — and `meta` is documented as an empty object, so the generic cursor/page walk copied from the DNSFilter provider is gone. Cisco documents **no maximum** for `limit`, which makes a short page an unsafe end-of-collection signal: a server-side cap below ours would silently truncate the sync. The walk advances by the records **actually returned** and stops only on an empty page, still under a 100-request cap. Cost: one extra (empty) request per sync; benefit: no silent truncation.
- **`x-traffic-type: dns`** — `/reports/v2/activity` is the combined feed (dns + proxy + firewall + intrusion). The documented header scopes it to DNS records rather than paging through records that can never map.
- **30-day window clamp** — Cisco documents "the time range set by the `to` and `from` query parameters cannot exceed 30 days". An integration broken for longer than that (exactly this issue's situation, where sync has never succeeded) would otherwise request a window the API rejects and stay broken after this fix. The window is clamped to the most recent 30 days; activity older than that cannot be backfilled through the API, which is now stated in the docs page.

## Provenance of the schema

The response schema was mapped from **Cisco's published OpenAPI spec** for `/reports/v2/activity` and `/reports/v2/activity/dns` (developer.cisco.com, Umbrella Reporting API v2.0.0) plus the reporter's live 200 in #4597 — **not** from a live token exercised in CI. Field names verified against the spec: `timestamp`, `domain`, `querytype`, `verdict`, `categories[]`/`policycategories[]` as `{id,label,type,…}` objects, `identities[]` with a nested `type` object, `threats[]` as `{label,type}`, `internalip`/`externalip`, and the absence of any per-record id.

One documented inconsistency worth flagging: the record schema declares `verdict` as `allowed | blocked`, while the `verdict` *filter* parameter is documented as `allowed,blocked,proxied`. `proxied` is therefore handled but not assumed.

## Review round — observability of dropped events

A silent-failure review raised five paths where this ingestion could lose events without anyone noticing (a DNS-security dashboard that reads clean while blocked queries go missing). All five are addressed, mostly by carrying over the pattern `piholeV6.ts` already uses in the same directory:

- An unrecognized `verdict` no longer falls through to `allowed` — relabelling a possible detection as clean traffic is worse than dropping the record. It is counted as unparseable instead. An **absent** `verdict` is treated the same way: Cisco lists it in the required field set for a DNS record, so its absence is drift, not a default.
- Records dropped as unparseable are counted and warned about at the end of the sync, so an upstream shape change cannot return `[]` and be recorded as a healthy "success".
- `mapActivityRecord` also checks the `type` discriminator and skips non-DNS records, counted separately — defence in depth for the `x-traffic-type` header, since a proxy record carries a domain and timestamp too. A record with **no** `type` is still kept, so a missing discriminator cannot blank the feed.
- Hitting the 100-request cap warns: the job advances `lastSync` to `until` on success, so a truncated run's remainder is never retried.
- Clamping to the 30-day maximum warns with the effective range rather than narrowing the request mutely.

The mapper returns a classified outcome (`event` / `unparseable` / `non-dns`) so the pagination loop can count each reason.

## Tests

- `apps/api/src/services/dnsProviders/umbrella.test.ts` — endpoint + no org id in the path, epoch-ms `from`/`to` with `limit`/`offset`, the `x-traffic-type` header, the 30-day clamp (and a control that an in-range window is untouched), full record mapping, external-IP fallback, all three verdicts, malformed records dropped rather than thrown, the offset walk (advance by records returned, stop on empty), and the 100-request cap.
- `apps/api/src/routes/dnsSecurity.test.ts` — Umbrella accepted without `organizationId`, still accepted with it, and a control that a missing `apiSecret` is still a 400.
- `apps/web/.../AddDnsIntegrationModal.test.tsx` — the inline-error case no longer mocks the removed `organizationId is required` message.

Verified locally: `vitest run src/services/dnsProviders src/routes/dnsSecurity.test.ts src/jobs/dnsSyncJob` → 9 files, 115 passed. `apps/web` `vitest run src/components/dnsSecurity src/lib/i18n/localeParity.test.ts` → 6 files, 106 passed. `tsc --noEmit` clean in both `apps/api` and `apps/web`.

Left for the reporter to verify against a live Umbrella tenant before the issue is closed.


ToddHebebrand:
**Review run:** `pr-review-toolkit` — code-reviewer (correctness) + silent-failure-hunter, both against the diff; then a second, focused code-reviewer pass over the fix commit `4434a596b` because the fixes touched the ingestion mapper itself.

**Findings:** 6 raised → all 6 addressed; 0 outstanding.
- **Round 1 — correctness:** 0 findings at or above the confidence bar. Pagination arithmetic, the window clamp, the mapper, `providerEventId` handling and the `organizationId` relaxation all traced clean, and the new tests were checked for discrimination rather than vacuity.
- **Round 1 — silent failures:** 5 findings, all real, all fixed in `4434a596b` + `4ca2fa99c`. An unrecognized `verdict` no longer defaults to `allowed` (relabelling a possible detection as clean traffic is worse than dropping it); unparseable records are now counted and warned about; a `type` discriminator check backs up the `x-traffic-type: dns` header so a proxy record can't be ingested as a DNS event; the 100-request cap and the 30-day window clamp both warn instead of truncating/narrowing mutely. Three of the five were carried over from the pattern `piholeV6.ts` already uses in this same directory for exactly this reason.
- **Round 2 — re-review of the fix:** 1 finding — folding "no `verdict` field" into the same drop path as an unrecognized value was an unstated policy change with no test. It is deliberate (Cisco lists `verdict` in the required field set for a DNS activity record, so its absence is drift, not a default), now stated in the code and pinned by a test in `3d5f47898`.

**Tests:** `apps/api` affected suites green — `vitest run src/services/dnsProviders src/routes/dnsSecurity.test.ts src/jobs/dnsSyncJob` → 9 files, 121 passed. `apps/web` → `vitest run src/components/dnsSecurity src/lib/i18n/localeParity.test.ts` → 6 files, 106 passed. `tsc --noEmit` clean in `apps/api` and `apps/web`; `astro check` clean in `apps/docs`; eslint clean on every changed file. Every behavioural assertion was watched fail first against the unfixed code.

**Status:** review-clean, awaiting maintainer merge. Two things for the merge decision: the response schema was mapped from Cisco's published OpenAPI spec plus the reporter's live 200, **not** from a live token exercised here — so #4597 should stay open until the reporter confirms a sync completes on their install. And the non-English `security.json` help texts still carry the old "organizationId from your Umbrella dashboard URL" wording; only `en` was updated, in line with the usual translation lag.

cisspUser01:
Confirmed on 0.111.1 — the /reports/v2/activity endpoint fix works. Auth succeeds, and the sync begins successfully paginating through real data. But it fails partway through:

[DnsSyncJob] sync failed (upstream HTTP error) {"status":400,"statusText":"Bad Request",
"responseBody":"{\"meta\":{},\"data\":{\"errors\":[{\"param\":\"offset\",\"value\":\"11000\",
\"error\":\"invalid offset: should be integer between 0 and 10000 inclusive\"}]}}"}

Root cause: Cisco's API caps offset at 10000. Our org has more than 10,000 DNS events within the sync's lookback window (UMBRELLA_MAX_WINDOW_MS = 30 days), so pagination runs straight into this ceiling.

Worse than just "large orgs can't sync": despite successfully fetching and presumably processing multiple pages (offsets 0 through 10000) before the failure, zero rows exist in dns_security_events for this org afterward:

SELECT COUNT(*) FROM dns_security_events WHERE org_id = '<org_id>';
-- 0

This suggests the whole sync is processed as one all-or-nothing unit — a failure on a later page discards the events successfully fetched from earlier pages too, not just the data beyond the cap.

Suggested fix direction: switch to Cisco's cursor-based pagination (the API supports a cursor param for exactly this case, per their docs) instead of incrementing offset indefinitely, and/or persist events incrementally per-page rather than only at the end of a fully successful run, so a late failure doesn't discard already-fetched data.

cisspUser01:
also the show all on the events tab doesnt save.
Screenshot 2026-09-09 202238

cisspUser01:
Confirmed on 0.111.1 — the /reports/v2/activity endpoint fix works. Auth succeeds, and the sync begins successfully paginating through real data. But it fails partway through:

[DnsSyncJob] sync failed (upstream HTTP error) {"status":400,"statusText":"Bad Request",
"responseBody":"{"meta":{},"data":{"errors":[{"param":"offset","value":"11000",
"error":"invalid offset: should be integer between 0 and 10000 inclusive"}]}}"}

Root cause: Cisco's API caps offset at 10000. Our org has more than 10,000 DNS events within the sync's lookback window (UMBRELLA_MAX_WINDOW_MS = 30 days), so pagination runs straight into this ceiling.

Worse than just "large orgs can't sync": despite successfully fetching and presumably processing multiple pages (offsets 0 through 10000) before the failure, zero rows exist in dns_security_events for this org afterward:

SELECT COUNT(*) FROM dns_security_events WHERE org_id = '<org_id>';
-- 0

This suggests the whole sync is processed as one all-or-nothing unit — a failure on a later page discards the events successfully fetched from earlier pages too, not just the data beyond the cap.

Suggested fix direction: switch to Cisco's cursor-based pagination (the API supports a cursor param for exactly this case, per their docs) instead of incrementing offset indefinitely, and/or persist events incrementally per-page rather than only at the end of a fully successful run, so a late failure doesn't discard already-fetched data.


this is also confirmed on 0.114.0.

Issue #5080, opened 2026-09-06
Configuration policy inheritance — persisted one-level parent, effective-links view, resolver sweep


ToddHebebrand:
Tracking issue for **Configuration policy inheritance — persisted one-level parent, effective-links view, resolver sweep** (`config-policy-inheritance`).

- Plan: `docs/superpowers/plans/config-policy/2026-09-06-config-policy-inheritance-01-api-foundation.md`
- Spec: `docs/superpowers/specs/config-policy/2026-09-06-config-policy-inheritance-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
All three waves merged and unreleased: W01 #5099 (f6ec8a33f), W02 #5125 (6c17fcebb), W03 #5122 (02e8e5b29), plus follow-up #5127 (5d21828f9) which pinned the self-FK in the erasure ledger and scoped owner-move races to system context after W01 reddened main's integration shards.

Origin: the third paper cut on the #5023 browser walk ("inherited policy state only shows with ?linked=") was not a UI bug — the linked-policy create mode persisted nothing and no resolver inherited, so a device assigned to a "linked" child received nothing for un-overridden tabs.

Release notes: configuration policies can now inherit from a baseline policy (same org, or partner-wide). Migration adds `configuration_policies.parent_policy_id` and the view `config_policy_effective_feature_links`; requires PostgreSQL 15+ (security_invoker). Policies created earlier through "Link to Existing" were never actually linked — re-create them as linked policies to get inheritance.

Open follow-ups, neither blocking:
1. `aiToolsBackup.list_policies` and `aiToolsFleet.manage_service_monitors` still report authored-only links. Deliberate: their list output feeds update/remove in the same tool, so surfacing an inherited link id would let an org-scoped agent edit the MSP's baseline. Needs inherited rows flagged read-only (product decision).
2. The partner-export parent-closure CTE is verified by rendered SQL only; no partner-export integration harness exists.
3. Marketing-site release-notes entry still to be added when the next release is cut.

PR #5099, opened 2026-09-06
feat(config-policy): W01 API foundation — parent_policy_id, constraint triggers, effective-links view (#5080)

### LINKED ISSUE #5081: W01: API foundation — parent_policy_id migration, constraint triggers, effective-links view, create/delete/eligible-parents API
Wave W01 of #5080 — API foundation — parent_policy_id migration, constraint triggers, effective-links view, create/delete/eligible-parents API
Branch: `feature/5080-config-policy-inheritance/wave-5081`

ToddHebebrand:
Wave W01 of configuration-policy inheritance (#5080). Persists a validated, tenant-safe `parent_policy_id`, ships the `config_policy_effective_feature_links` view, and exposes the parent on the API — **without switching any resolver yet** (that is W02).

Spec: `docs/superpowers/specs/config-policy/2026-09-06-config-policy-inheritance-design.md`
Plan: `docs/superpowers/plans/config-policy/2026-09-06-config-policy-inheritance-01-api-foundation.md`

> This branch is cut from the docs branch (open as #5084), so the four spec/plan docs show in this diff until that merges. Everything else is W01 code.

## A fail-open tenancy hole the integration suite caught

The migration as specified was **fail-open** for one shape, and the live-DB suite found it on first run:

`breeze_config_policy_parent_compatible` returned **NULL, not false**, for an org-owned child naming *another partner's* partner-wide parent. `parent_org = child_org` is NULL whenever `parent_org IS NULL`, so the CASE evaluated `NULL OR false` → `NULL`; the guard's `IF NOT compatible(...)` saw NULL, which is not TRUE, and **did not fire**. The forged cross-partner INSERT succeeded.

Only a live-DB test could find it: the TypeScript mirror (`isCompatibleParent`) is correct because JavaScript has no three-valued logic, so every unit test passed while the database accepted the edge.

Fixed belt-and-braces: the SQL function wraps its CASE in `COALESCE(..., false)` so it is strictly two-valued and fails closed, and **both** trigger call sites test `IS NOT TRUE` rather than `NOT (...)` so the guard cannot fail open again even if the function regressed. Pinned by a direct table of all nine ownership shapes asserting exact booleans (never NULL).

## What's in it

**Migration** `2026-10-12-100000-config-policy-inheritance.sql` (sorts after `2026-10-12-000100-…`, the newest on main; no DML, so no `breeze.scope` election; idempotency verified by re-applying the file directly and by a second ledger replay)
- `parent_policy_id` + self-FK (default **NO ACTION**), not-own-parent CHECK, partial index.
- `configuration_policies_parent_guard` — `DEFERRABLE INITIALLY IMMEDIATE` constraint trigger validating **both** edge directions. `parent_policy_id` immutable **including NULL → value**; ownership may change only in system context, which closes the blind spot where a partner-scoped invoker could move a baseline while RLS hid some children from the incoming-edge check. Being a *constraint* trigger, org merge's `SET CONSTRAINTS ALL DEFERRED` postpones it to commit.
- `organizations_partner_config_policy_guard` — no code path changes an org's partner today; this keeps the invariant from depending on that.
- `config_policy_effective_feature_links` `WITH (security_invoker = true)`, granted to `breeze_app`. An inherited row keeps the **parent link's id** because the normalized per-feature settings tables are keyed by `feature_link_id`.
- `SECURITY DEFINER` functions are `REVOKE ALL … FROM PUBLIC` so the compatibility helper is not an "is org X under partner P" oracle.

**API**
- `POST /configuration-policies` takes `parentPolicyId`; validated inside a transaction opened *only* for that case. The parent is read through the caller's own RLS context and deliberately **without a row lock** — `FOR KEY SHARE` applies the UPDATE policy, which the SELECT-only partner-wide branch does not satisfy, so a lock would break exactly the case inheritance exists for. The delete race is closed by the FK instead (23503 → the same 400).
- One `400 INVALID_PARENT_POLICY` for not-found / not-eligible / cross-tenant / has-own-parent, so it is not an existence oracle.
- `DELETE` → `409 POLICY_HAS_CHILDREN` naming the blockers; self-FK 23503 is the backstop for the create-after-check race.
- `GET /:id` embeds `parentPolicy` (assembled links) + `childPolicies`. `GET /eligible-parents` (names only), declared above `GET /:id` with its own route-ordering regression test.
- **`policyAccessCondition` is NOT relaxed.** The parent embed and eligible-parents are read-only additions; the gate that hides partner-wide policies from org-scoped get/list is untouched.
- MFA follows **effectiveness**: creating a child of a parent carrying a patch/maintenance link is gated, and deleting a `maintenance` override whose parent has a maintenance link is gated (that delete *restores* suppression rather than ending it). Patch DELETE stays unconditionally gated; maintenance removal on a root policy stays exempt.

**Registries** — `parent_policy_id` classified `included` in `tenantExportPolicyRegistry`; partner export carries `parentPolicyId` and gains a `parent_orgs` **closure CTE** (policies are selected through their *assignments*, so an unassigned baseline would have vanished while its exported children still referenced it). The closure re-asserts the ownership rule rather than assuming it. No cascade-list change is needed (no new table, no new `device_id`).

## Verification

Live suites, all run against a real Postgres test stack (each confirmed to actually execute, non-zero counts):

| suite | result |
|---|---|
| `configPolicyInheritance.integration.test.ts` (new) | **27 passed** |
| `rls-coverage` (`vitest.config.rls-coverage.ts`) | **102 passed** |
| `tenantCascade.integration.test.ts` | **5 passed** |
| `tenant-export-policy` + `tenantExportErasureRoundtrip` | **6 passed** |
| `orgLifecycleFoundations` + `orgMerge` (6 files) | **44 passed** |
| `configPolicyPartnerWideSelect` + `configurationPoliciesPartnerRls` | **25 passed** |

Also: `pnpm db:check-drift` → no drift (692 files); `migrationRlsScope` + `autoMigrate` → 103 passed; affected unit suites → 908 passed; `pnpm lint` clean; `tsc --noEmit` clean. All re-run **after** merging `origin/main`.

**`breeze_app` cross-tenant forge** (`psql -U breeze_app`, org-scoped GUCs, bypassing the service entirely):

```
cross-PARTNER partner-wide parent → ERROR: parent configuration policy not found or not eligible  (23514)
cross-ORG parent                  → ERROR: parent configuration policy not found or not eligible  (23514)
control: plain root insert        → INSERT 0 1
control: legitimate same-org child→ INSERT 0 1
```

The two controls are what make the forge non-vacuous — the rejections are the rule firing, not a broken connection.

New tests were mutation-checked rather than assumed: disabling the children pre-check / DB error mapping / parent embed turns 11 of 24 service tests red; disabling the MFA gate / 409 / 400 mapping turns 3 of 47 route tests red; disabling the maintenance-revert gate turns 1 of 53 red.

## Deviations from the plan

1. **Migration fixed beyond the plan's SQL** — the three-valued-logic hole above. The plan's literal SQL was fail-open.
2. **`createConfigPolicy` opens a transaction only when a parent is named** (plan wrapped every create). Same guarantee, no behaviour or cost change for the overwhelmingly common create.
3. **`pgErrorConstraint()` added to `utils/pgErrors.ts`** instead of hand-rolling a second cause-chain walker, per the plan's own instruction.
4. **Task 11 (OpenAPI) was a no-op** — `apps/api/src/openapi.ts` carries no configuration-policy paths, and the user-facing docs page is W03's.
5. **Two extra registries had to be updated**, both caught by their contract tests, neither listed in the plan: the strict partner-export DTO (`partnerApi/schemas.ts`) and the DTO-safety reviewed-field snapshot.

## Not verified / follow-ups

- `EXPLAIN` comparison for the view under `breeze_app` — the plan assigns that to W02, where the resolvers actually switch to it.
- Whether an **archived** parent should be offered in the eligible-parents picker. Left unfiltered (spec: a parent's status does not gate inheritance); flagged for W03's UI.
- Pre-existing, untouched: `PartnerWideWriteDeniedError` in the AI config-policy tool is still flattened by `safeHandler` into a generic "Operation failed". I added the branch for `PolicyHasChildrenError` only, to keep this wave's scope.

Closes #5081


ToddHebebrand:
**Review run:** `pr-review-toolkit:review-pr` — 4 reviewers in parallel: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer.

**Findings:** 13 raised (3 CRITICAL, 7 IMPORTANT, 3 SUGGESTION) → **11 addressed in a9e63e1**, 2 deliberately deferred (below). 0 outstanding defects.

**By reviewer**
- **code-reviewer — 0 findings.** Independently traced the migration ordering, idempotency, the three-valued-logic fix, one-level enforcement, deferred-trigger semantics under org merge, the FK `NO ACTION` choice against both `cascadeDeleteOrg` and `cascadeDeletePartner`, the `security_invoker` view, the `parent_orgs` closure, and the `policyAccessCondition` bypass. Nothing met its reporting bar.
- **silent-failure-hunter — 1 CRITICAL, 1 IMPORTANT, 1 SUGGESTION.** Both real, both fixed.
- **pr-test-analyzer — 2 CRITICAL, 6 IMPORTANT, 2 SUGGESTION.** Both criticals and four importants fixed.
- **comment-analyzer — 0 CRITICAL, 1 IMPORTANT, 2 SUGGESTION.** Verified the load-bearing tenancy claims against PostgreSQL docs and the code (the `FOR KEY SHARE`/UPDATE-policy claim, `security_invoker`, deferrability, the parent-link-id rationale, zod strip behaviour) — all accurate. The one important finding is fixed.

**The finding that mattered most:** the maintenance-revert MFA gate **failed open**. `parentPolicyId` set with `parentPolicy` null collapsed to "no parent" and silently skipped the MFA requirement — the *same* fail-open shape this PR had already hit once in SQL, reappearing in TypeScript. It now fails closed, and `getConfigPolicy` reports the anomaly to Sentry rather than letting it rot. Also fixed: the AI tool rendering the lost-race delete as "0 policy/policies inherit from it" while refusing, and a schema comment that claimed resolvers already read the effective-links view (nothing does until W02 — the misreading is "inheritance already reaches devices", which it does not).

**Coverage gaps closed** (integration suite 27 → 40 tests): `organizations_partner_config_policy_guard` had **no test at all**; the "ownership change would orphan child policies" branch was never exercised (moving only the *parent*) — without it that clause could be dead and the whole-family-move test would still pass. Plus self-parent CHECK forge, partner-wide child create + forge, archived-root eligibility, partner-wide baseline delete 409, and live-DB coverage for `getParentLinkFeatureTypes` / the `getConfigPolicy` parent embed (both mock-only despite depending on the partner-wide RLS read branch).

**Deliberately not done in this wave** (called out rather than silently dropped):
- Executing the `parent_orgs` closure CTE against a real database — it is currently verified by matching the rendered SQL, which cannot prove row-level correctness. Needs a partner-export integration harness that does not exist; code-reviewer independently traced the CTE as correct, and the DB-level ownership trigger independently prevents tenant-crossing, so this is an export-correctness gap, not a leak.
- Pre-existing and untouched: `PartnerWideWriteDeniedError` in the AI config-policy tool is still flattened by `safeHandler` into a generic "Operation failed". I added the `instanceof` branch for `PolicyHasChildrenError` only, to keep this wave's scope.

**Tests (all re-run after the fixes, and after merging `origin/main`):**
- `configPolicyInheritance.integration.test.ts` — **40 passed**
- `rls-coverage` — **102 passed**; `tenantCascade` + export-policy + erasure-roundtrip + `orgLifecycleFoundations` + `orgMerge` + `configPolicyPartnerWideSelect` + `configurationPoliciesPartnerRls` — **80 passed across 11 files**
- affected unit suites — **935 passed across 36 files**; `pnpm lint` clean; `tsc --noEmit` clean; `db:check-drift` no drift
- `breeze_app` cross-tenant forge: cross-partner and cross-org parents both rejected with 23514, with two positive controls (plain root insert, legitimate same-org child) proving the rejections are the rule firing rather than a broken connection

Every fix was mutation-checked rather than assumed — reverting the fail-closed branch or the race message turns its test red.

**Status:** review-clean, CI running, awaiting maintainer merge. Not merged, issue #5081 left open.

ToddHebebrand:
Heads-up: this merged with **Integration Tests shards 1, 2 and 4 failing on its own CI run** (only shard 3 passed), and main's integration suite is red since `f6ec8a33f`.

Same failures reproduce on branches that touch nothing under `apps/api` (e.g. #5074, e2e spec + ci.yml only), so it is not those branches:

- `partnerApiRls.integration.test.ts:595/628/674/780` — partner reconstruction export RLS traversal: owner moves on `configuration_policies` now fail (`update "configuration_policies" set "partner_id" = CASE …` rejected); one unhandled error at :674.
- `configPolicyReferenceConcurrency.integration.test.ts:52/172` — expected SQLSTATE `23503`, got `23514` (the new constraint trigger fires before the FK), and the referenced-policy move holder fails before reporting progress.
- `configPolicyReferenceGateConcurrency.integration.test.ts:340`.

Looks like the new `parent_policy_id` constraint triggers / effective-links view change the owner-move contract those suites pin. Every open PR is blocked on `ci-success` until main is green again — please fix forward or revert.

PR #5108, opened 2026-09-06
feat(web,api): service management mode + service desk sidebar section (W04 of #5075)

### LINKED ISSUE #5079: W04: Sidebar regroup + Service Management mode (columns, /partners/me, gate, off refusal)
Wave W04 of #5075 — Sidebar regroup + Service Management mode (columns, /partners/me, gate, off refusal)
Branch: `feature/5075-organization-record-page/wave-5079`

ToddHebebrand:
Wave W04 of #5075 — the partner-level Service Management mode, the Service Desk sidebar section, and the `off` refusal for new ticket creation.

## Task 4.1 — Migration + schema

`apps/api/migrations/2026-10-12-100100-partners-service-management-mode.sql` adds to `partners`:

- `service_management_mode text NOT NULL DEFAULT 'native'` + `partners_service_management_mode_chk` (`IN ('native','external','off')`)
- `service_management_psa_connection_id uuid REFERENCES psa_connections(id) ON DELETE RESTRICT`
- `partners_service_management_connection_chk`: `(mode = 'external') = (connection_id IS NOT NULL)` — a **biconditional**, so a stale binding cannot survive a flip back to native and silently re-attach later.

Idempotent (`ADD COLUMN IF NOT EXISTS`, `DROP CONSTRAINT IF EXISTS` then add), no new tables, no inner `BEGIN`/`COMMIT`. **No DML, so no `breeze.scope` elevation is needed.** `partners` has no `org_id` (verified against the live DB: `information_schema` reports 0 such column), so **no `tenantCascade` / export-policy registration applies**.

**Why this name sorts last:** it was originally `…-000200-…` (after the then-newest `2026-10-12-000100-device-manual-maintenance-lease.sql`), but `origin/main` gained `2026-10-12-100000-config-policy-inheritance.sql` mid-flight, which the old name would have replayed *ahead of* on a fresh database — the pre-push hook caught it. Renamed to `…-100100-…`; the file is unmerged, so no ledger row existed to invalidate, and the one code reference (the integration suite's header comment) was swept. `scripts/check-migration-naming.sh --against-ref origin/main` → `OK`, and the renamed file replays cleanly on a fresh database (`db:check-drift`: `No drift detected — all 693 migration files match the ledger`).

Drizzle gets both columns. The FK is declared **in the migration only** — `integrations.ts` already imports `orgs.ts`, so a `.references()` on the schema column would close an import cycle (the escape hatch the plan calls out).

## Task 4.2 — `serviceManagement` service + `/partners/me`

`apps/api/src/services/serviceManagement.ts` is the single source of the mode. It **fails open to `native`** on a missing partner id, a missing row, a `NULL`, or a value this build does not recognise — withdrawing a module a partner pays for is worse than showing a surface whose routes still enforce their own permissions.

`PATCH /partners/me`:
- `external` requires a `serviceManagementPsaConnectionId` resolving to a `psa_connections` row with `partner_id = auth.partnerId` **and** `org_id IS NULL`, else 400.
- `native`/`off` **force** the connection id to `null`, so the biconditional CHECK can never abort the write with an opaque 500.
- A connection id sent with no mode is refused with 400 (it could only contradict the stored mode).
- `external` is still accepted by the API even though the UI does not offer it — the follow-on feature turns the radio on without a breaking API change.
- Both fields are in `partnerPublicColumns()`, so `GET /partners/me` returns them. `changedFields` already derives from `Object.keys(body)`, so `serviceManagementMode` lands in the audit automatically.

`psaConnections` is imported from `../db/schema/integrations` rather than the `../db/schema` barrel: three route tests partially mock that barrel, and a new named import there fails their module load before a single test runs.

## Task 4.3 — `off` refuses new tickets

`ticketService.createTicket` calls `assertTicketCreationAllowed(org.partnerId)` right after resolving the org and **before any ticket-number allocation**, so a refused create burns no counter value. The `ServiceManagementOffError` is translated to `TicketServiceError(…, 409, 'service_management_off')`, which every existing `instanceof TicketServiceError` handler surfaces unchanged.

**This is the only mode check in the API** — verified by grep: `assertTicketCreationAllowed` / `getServiceManagementMode` appear outside the service in exactly one file, `ticketService.ts`. The alert dialog, the `manage_tickets` AI tool, the portal, the Office add-in and email-to-ticket all inherit it through `createTicket` (`createTicketFromAlert` calls it too). `external` is deliberately allowed through — it still writes the Breeze-side shadow row.

## Task 4.4 — Sidebar regroup + module gate

- New **Service Desk** section (Tickets moved from top-level, Timesheets moved from Billing) and the **Billing** section both carry `requiresModule: 'service_management'`; hidden unless the mode is `native`.
- **Organizations** promoted to top level right after Dashboard (`partnerScopeOnly`, `organizations:read`) and removed from Settings — a test asserts it renders exactly once across the whole nav.
- Mode read from `useOrgStore(s => s.serviceManagementMode)`: persisted, default `native`, set from the existing `/orgs/partners/me` effect. A failed fetch skips the write entirely (fail open), an unrecognised value falls back to `native`, and the persist `merge` sanitises a bogus stored value.

`grep -rn "requiresModule" apps/web/src` → `Sidebar.tsx` only, as the plan requires.

## Task 4.5 — Modules card

`PartnerModulesCard` renders `native` and `off` only. `external` is **not** rendered and an external partner is **not** silently rewritten — neither radio is checked and an explanatory note shows instead. Saves via `runAction` PATCH, then updates the store so the sidebar changes without a reload; a failed PATCH reverts the radio and leaves the store untouched.

The prop is `ServiceManagementMode | undefined`, and `undefined` ("not fetched yet") is deliberately **not** treated as `'native'`: publishing that placeholder would stomp the mode the Sidebar already fetched and flash an `off` partner's Service Desk sections back in on every Partner Settings mount.

## Task 4.6 — Record page + web create-ticket surfaces

`OrganizationRecordPage` reads the store mode and passes it to `visibleTabs`; `OrgOverviewTab` withdraws the tickets/contracts/invoices tiles when the mode is not `native` (filtered once against a named key set, so a tile added later cannot slip past by forgetting a condition); the alert-detail **Create ticket** trigger renders only for `native`, with the API 409 as the backstop.

## Test evidence

All run in this worktree against a per-worktree `pnpm test-stack` database (torn down after).

| Command | Result |
|---|---|
| `apps/api` `vitest run src/services/serviceManagement src/services/ticketService src/routes/orgs src/services/aiToolsTicketing src/db/autoMigrate src/db/migrationRlsScope` | **12 files, 741 passed** |
| `apps/api` `vitest run src/routes/alerts` | **27 files, 336 passed** |
| `apps/api` integration `partnersServiceManagementMode` **against real Postgres** | **10 passed** |
| `pnpm db:check-drift` (fresh DB, post-rename) | `No drift detected — all 693 migration files match the ledger` |
| `scripts/check-migration-naming.sh --against-ref origin/main` | `OK` |
| `apps/web` `tsc --noEmit` | clean |
| `apps/web` `astro check` (1975 files) | **0 errors, 0 warnings** |
| `apps/web` `vitest run` Sidebar · PartnerModulesCard · PartnerSettingsPage · organizations/record · orgStore · lib/i18n · no-silent-mutations · AlertDetailPage | **29 files, 480 passed** |
| `eslint` on every touched source file | clean |

The integration test covers both CHECKs (23514 for a bogus mode, for `external` with no connection, and for `native`/`off` that retain one) and the FK (23503 when deleting a bound connection) — plus the negative control that an **unbound** connection still deletes, so the RESTRICT test proves the binding rather than `psa_connections` being undeletable.

**Red-first evidence for the load-bearing gate:** with `requiresModule` neutered to `if (false)`, `Sidebar.module.test.tsx` goes **2 failed / 6 passed**; restored, 8/8. The gate assertions discriminate.

## Review round (`/pr-review-toolkit:review-pr`)

Four reviewers ran: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer. **5 findings raised, all 5 fixed in `ccc22f2`; 0 outstanding.**

- **code-reviewer: 0 findings.** All five contracts verified as HOLDING, including a repo-wide sweep confirming every `createTicket` / `insert(tickets)` call site routes through the single gate with no bypass.
- **pr-test-analyzer (Important):** the external-mode PSA probe's ownership filter was structurally **unverifiable** — the mocked `.where()` returned its canned row whatever predicate it received, so dropping `eq(partnerId, …)` (a partner binding **another partner's** PSA credentials) or `isNull(orgId)` (an org-scoped connection bound as if partner-wide, then serving every org from one org's credentials) would have failed no test. Fixed by spying `isNull` with the real implementation (mirroring the existing `eq` spy) and asserting all three conditions. **Verified by mutation: removing `isNull(orgId)` now reds the suite.**
- **silent-failure-hunter (Important):** `PartnerModulesCard` rethrew a non-`ActionError` from a handler invoked as fire-and-forget `void select(...)`, making it an unhandled rejection — the radio would snap back with no toast and no telemetry. Now uses the repo's `handleActionError` helper.
- **comment-analyzer (Critical ×2, Important ×1, Suggestion ×1):** the alert-detail gate comment claimed the API refuses "regardless", which is **false for `external`** (the API accepts that create); a test comment referenced `checkedMode`, a symbol that no longer existed after the card's state was collapsed onto `mode`; the tile comment overstated its rationale for `external`; and the `Partner` type comment implied `undefined` only meant version skew when it is also every pre-fetch render. All four corrected.

Not acted on, by judgement:
- A narrow race where the Sidebar's in-flight mount fetch could clobber a mode saved from the settings page moments earlier. Cosmetic, self-heals on the next navigation, and the server's stored value is never wrong. The reviewer agreed it is not a blocker.
- `NavItem.requiresModule` is currently only used at section level. It is plan-mandated (`NavItem` and `NavSection` both gain the field) and is honoured by `isNavItemVisible`, so it works the moment an item needs it.

## Deviations

1. **`TicketServiceErrorCode` gains a lowercase member.** Every existing code in that union is `UPPER_SNAKE`; `'service_management_off'` is lowercase because the plan and the wave contract both name that exact wire token, and the web/AI surfaces branch on it. Noted rather than silently normalised.
2. **`ServiceManagementMode` is declared in `orgStore.ts`, not `orgRecordTabs.ts`.** W01 declared it in `orgRecordTabs.ts`; the store now owns the runtime value, so keeping a second copy there would let the tab gate and the store drift. `orgRecordTabs.ts` re-exports it, so every existing import path still works.
3. **`aiToolsTicketing.ts` (a source file) changed.** Its `create` action had no `try/catch` around `createTicket`, unlike every sibling mutating action, so a thrown `TicketServiceError` escaped as an unhandled rejection instead of the `{error, code}` JSON the AI relays. Without this, contract 2's "the AI tool returns error JSON" would not hold. Fixed inline rather than deferred.
4. **`OrganizationRecordPage.test.tsx` had two selectors updated.** Verified against a clean `origin/main` worktree: `hides the Devices tab from a user without devices:read` was **already red on main** — #5090 moved `OverflowTabs` to `role="tab"`, and jsdom's zero widths collapse the strip to the active tab. Applied the same `stubWideLayout` pattern `OverflowTabs.test.tsx` already uses and changed `role: 'button'` → `role: 'tab'`. No assertion was weakened.
5. **Eight `Sidebar*.test.tsx` and two `AlertDetailPage*.test.tsx` auth mocks gained `registerOrgIdProvider: vi.fn()`.** `orgStore` calls it at module scope, so any file that mocks `@/stores/auth` and now transitively loads the store dies at import. One line each; no test logic touched.
6. **French `partnerSettingsPage.modules.title` is "Modules du produit", not "Modules".** The bare cognate tripped `translationCoverage`'s exact-English duplicate ratchet for fr-FR and fr-CA. Chose a natural, distinct label over permanently loosening the baseline.

## Not done / notes for the reviewer

- **No browser walkthrough.** The plan's W04 verification lists a manual flip-the-mode pass; this run verified the behaviour through the jsdom suites only.
- `getServiceManagementMode` adds one indexed PK lookup on `partners` per `createTicket`. Negligible, but it is a new query on the email-to-ticket ingest path — flagging it rather than burying it.

Closes #5079


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer (4 agents, parallel, full diff).

**Findings:** 5 raised → all 5 addressed in `ccc22f2`; **0 outstanding.**

- **code-reviewer — 0 findings.** Verified all five wave contracts HOLD, including a repo-wide sweep confirming every ticket-creation call site (`routes/vulnerabilities`, `routes/ai`, `routes/portal/tickets`, `inboundEmailService`, `aiToolsTicketing`, `createTicketFromAlert`) routes through `createTicket` — no bypass of the single gate, and no route's own permission check was replaced or weakened by a mode check.
- **pr-test-analyzer — 1 Important.** The external-mode PSA-connection probe's ownership filter was structurally unverifiable: the mocked `.where()` returned its canned row whatever predicate it received, so dropping `eq(partnerId, …)` (a partner binding **another partner's** PSA credentials) or `isNull(orgId)` (an org-scoped connection bound as partner-wide, then serving every org from one org's credentials) would have failed no test. Fixed by spying `isNull` with the real implementation — mirroring the `eq` spy already in that file — and asserting all three conditions. **Confirmed by mutation: removing `isNull(orgId)` now reds the suite.**
- **silent-failure-hunter — 1 Important.** `PartnerModulesCard` rethrew a non-`ActionError` out of a handler invoked as fire-and-forget `void select(...)`, turning it into an unhandled rejection: the radio snapped back with no toast and no telemetry. Now uses the repo's `handleActionError` helper. The reviewer explicitly cleared the deliberate fail-open paths, the 409 path end-to-end, and the PATCH validation.
- **comment-analyzer — 2 Critical, 1 Important, 1 Suggestion.** The alert-detail gate comment claimed the API refuses "regardless", which is **false for `external`** (the API accepts that create); a test comment pointed at `checkedMode`, a symbol that no longer existed after the card's state was collapsed onto `mode`; the overview-tile comment overstated its rationale for `external`; and the `Partner` type comment implied `undefined` meant only version skew when it is also every pre-fetch render. All four corrected.

**Not acted on, by judgement:** a narrow race where the Sidebar's in-flight mount fetch could clobber a mode saved seconds earlier — cosmetic, self-heals on the next navigation, server truth never wrong, and the reviewer agreed it is not a blocker.

**Tests:** `apps/api` 741 passed across the wave's suites + 336 in `routes/alerts` + 311 in `routes/orgs` after the fix; integration `partnersServiceManagementMode` **10 passed against real Postgres** on a freshly-migrated database; `apps/web` 1436 passed across settings/alerts/record/i18n. `tsc --noEmit` clean on both apps, `astro check` 0 errors, `eslint` clean.

**Migration:** renamed `…-000200-…` → `2026-10-12-100100-partners-service-management-mode.sql` mid-flight — `origin/main` gained `2026-10-12-100000-config-policy-inheritance.sql` after this branch was cut, which the old name would have replayed ahead of on a fresh database (the pre-push hook caught it). File is unmerged, so no ledger row was invalidated; the one code reference was swept. `check-migration-naming.sh --against-ref origin/main` → OK, and `db:check-drift` on a fresh DB reports no drift across all 693 migrations.

**Not covered:** no browser walkthrough — the plan's W04 verification lists a manual flip-the-mode pass, and this run verified through the jsdom suites only.

Issue #5078, opened 2026-09-06
W03: Tickets and Contracts & Billing tabs, lockedOrgId, cross-links, e2e


ToddHebebrand:
Wave W03 of #5075 — Tickets and Contracts & Billing tabs, lockedOrgId, cross-links, e2e
Branch: `feature/5075-organization-record-page/wave-5078`

PR #5109, opened 2026-09-06
feat(web): organization record — contacts, sites, devices, activity tabs (W02 of #5075)

### LINKED ISSUE #5077: W02: Contacts, Sites, Devices, Activity tabs + useSiteCrud
Wave W02 of #5075 — Contacts, Sites, Devices, Activity tabs + useSiteCrud
Branch: `feature/5075-organization-record-page/wave-5077`

ToddHebebrand:
## Summary

Wires four data tabs onto the organization record shell (W01, #5086):
**Contacts** (`ContactsCard`), **Sites** (new `OrgSitesTab`), **Devices**
(new `OrgDevicesTab`), and **Activity** (new `OrgActivityTab`). Tickets and
Billing stay as placeholders — those are W03 (#5078), which ran in parallel
and touches its own two `case` branches of `OrganizationRecordPage.tsx`.

### Task 2.1 — `useSiteCrud(orgId)` + `SiteModals`
Extracted the site CRUD state/handlers verbatim from `OrganizationsPage.tsx`
into `useSiteCrud.ts` (same URLs, same `runAction` fallbacks, same
`ActionError` catch pattern), plus the shared modal JSX into `SiteModals.tsx`.
The only behavioral addition is `orgIdOverride` on the sites GET, needed for
the hook to be correct when instantiated with a non-ambient org (the record).
`OrganizationsPage.tsx` now consumes both — behaviour unchanged, its own test
suites (72 tests across 9 files) pass unmodified.

**Deviation:** `refresh()` takes an optional org-id override
(`refresh(orgIdOverride?: string)`, not the plan's bare `refresh()`).
`OrganizationsPage`'s create-org flow fetches the brand-new org's sites
*before* its own `setSelectedOrg` state update has re-rendered the hook with
the new id — a plain `refresh()` would still read the previous org from
closure. The override lets that one call site target the just-created org
directly while every other caller (the record's tabs, the effect-driven
refresh) uses the bound default.

### Task 2.2 — Contacts and Sites tabs
`OrganizationRecordPage` renders `<ContactsCard orgId={orgId} />` for
`contacts` (unmodified — it already took an explicit `orgId` prop and embeds
the org in its list/create paths) and `<OrgSitesTab orgId={orgId} />` for
`sites`, which pairs `useSiteCrud` + `SiteList` + `SiteModals`.

### Task 2.3 — Devices tab + `forceSingleOrg`
`DeviceList` gains `forceSingleOrg?: boolean`; the fleet-view hook call stays
unconditional (`const fleetFromStore = useOrgStore(...); const isFleetView =
!forceSingleOrg && fleetFromStore;`) so a later prop flip can never violate
the rules of hooks. `OrgDevicesTab` loads `GET /devices?limit=200` and
`GET /orgs/sites?organizationId=` through the shell's `orgFetch`, with a
cut-down search/status/site filter bar (client-side, over the bounded page —
not the fleet page's full chip-based filter builder).

### Task 2.4 — Activity tab + `AuditLogViewer({orgId})`
`AuditLogViewer` gains an `orgId?: string` prop; when set, `/audit-logs`,
`/audit-logs/search`, **and** `/audit-logs/export` all carry
`orgIdOverride: orgId` (the export endpoint isn't in the plan's literal text
but is the same tab surface and the same leak risk, so it's pinned too).

### Task 2.5 — Settings `#contacts` hands off to the record
`OrgSettingsPage`'s `contacts` nav entry stays (still discoverable from
settings), but activating it — via a direct `#contacts` deep link or a nav
click — now calls `navigateTo('/organizations/<id>#contacts')` instead of
rendering `ContactsCard` inline. Covered by one effect keyed on `activeTab`
so both activation paths funnel through the same code.

## Found while verifying (fixed inline)

`OrganizationRecordPage.test.tsx`'s permission-gating test queried
`getByRole('button', { name: 'Overview' })`, but `OverflowTabs` renders each
tab as `<button role="tab">` — the explicit role overrides the implicit one,
so the query never matched. **Confirmed red on unmodified origin/main too**
(verified via `git stash`), not something this wave introduced — fixed to
query `role: 'tab'` since it's part of the required verification suite.

## i18n

New `orgRecord.devices.*` keys (search/status/site filter copy + the 7
device-status labels, matching `deviceList.statuses.full` wording) across all
8 locales with real translations. Five locales' `organizations.json`
duplicate-baselines bumped in `translationCoverage.test.ts` for genuine
cognates: pt-BR/de-DE/it-IT +2 ("Online"/"Offline"), fr-FR/fr-CA +1
("Maintenance") — each with a one-line comment naming the cognate.

## Test evidence

```
pnpm --filter @breeze/web exec tsc --noEmit    → clean
npx astro check                                 → 0 errors, 0 warnings (252 pre-existing hints)
pnpm --filter @breeze/web lint                  → clean

npx vitest run src/components/organizations/record src/components/settings/useSiteCrud \
  src/components/settings/OrganizationsPage src/components/settings/OrgSettingsPage \
  src/components/devices/DeviceList src/components/audit/AuditLogViewer src/lib/i18n \
  src/lib/__tests__/no-silent-mutations --pool=threads --maxWorkers=2
  → 36 files / 633 tests pass
```

New/updated coverage: `useSiteCrud.test.ts` (11), `OrgSitesTab.test.tsx` (2),
`OrgDevicesTab.test.tsx` (3), `OrgActivityTab.test.tsx` (1),
`AuditLogViewer.test.tsx` (3, new file — none existed before), a new
`DeviceList` case for `forceSingleOrg`, two new `OrgSettingsPage` redirect
cases, and four new `OrganizationRecordPage` cases proving each new tab pins
its requests to the record org while the store points at a different one.

## Deviations

- `useSiteCrud.refresh` override param (see Task 2.1 above).
- `AuditLogViewer`'s export endpoint also pinned (see Task 2.4 above).
- `console.warn` prefixes in the extracted hook kept as `[OrganizationsPage]`
  rather than renamed to `[useSiteCrud]`, because `OrganizationsPage.firstSite.test.tsx`
  pins those exact strings and this is still that page's first-site detection
  path, only relocated.
- Fixed the pre-existing `OrganizationRecordPage.test.tsx` role-query defect
  described above.

Closes #5077


ToddHebebrand:
## Review summary — 5 specialist passes, all findings fixed

Ran a 5-lens parallel review (code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer) against the org-pinning contract this wave exists to protect. Every confirmed finding is fixed in this PR (`3f0e78759`) rather than deferred.

### Fixed

| # | Sev | Finding | Fix |
|---|---|---|---|
| 1 | Med | **`useSiteCrud`'s mutation methods (`submit`, `confirmDelete`) didn't carry the same org-pinning guarantee as `refresh`.** Traced into the API (`orgs.ts`): the three site mutation routes authorize off the request body / the site's own DB row, never the query string, so this was inert today (matching the precedent already accepted for `orgArchive.ts`/`orgMerge.ts` in W01) — but the type gave callers no signal of the asymmetry, and it's a live risk if the API ever starts reading `orgId` from the query for these routes. | `orgIdOverride: orgId` added to both calls, matching `refresh` exactly. |
| 2 | Med | **`OrgDevicesTab`'s devices/sites fetches silently treated a malformed 200 body as "confirmed empty."** A 200 with `{success:false}` or `{data:null}` fell through to `[]` and rendered "No devices found — try adjusting your filters," actively wrong guidance for what was really an API failure. | Malformed body now fails the same way a non-2xx response does (mirrors the pattern `useSiteCrud.refresh` already used). |
| 3 | Med | **`OrgSitesTab` had no error state at all** — `useSiteCrud.refresh()` already distinguished "failed" from "confirmed empty" internally, but that distinction was discarded on the way to the UI, so a failed load rendered "No sites found. Add a site to get started." | New `sitesFailed` flag on the hook; `OrgSitesTab` renders a real error card with retry instead of falling through to the empty state. |
| 4 | Low | **Test coverage gaps**: the "surfaced via toast" claim in a `useSiteCrud` failure test never actually checked the toast mock; `confirmDelete`'s failure path had no test; the "search/status/site filters narrow rows" test only exercised search. | Toast assertion added; new `confirmDelete` failure test; the filter test split into three real tests (search/status/site), each proving its own filter narrows rows. New tests added for the two failure-state fixes above. |
| 5 | Low | Two comment precision issues: `OrgDevicesTab`'s `limit=200` docblock didn't disclose it's an unenforced assumption with no truncation notice; `DeviceList`'s rules-of-hooks comment implied the hazard only exists if the prop's value changes. | Both reworded. |

### Confirmed as non-issues (verified, not fixed)

- `code-reviewer` traced `submit`/`confirmDelete`'s original `orgIdOverride` omission all the way through `apps/api/src/routes/orgs.ts` and confirmed it had zero effect on authorization — fixed anyway (#1 above) as defense-in-depth, not because it was live.
- `comment-analyzer` independently verified the PR body's claim that the `OrganizationRecordPage.test.tsx` role-query defect is pre-existing on unmodified `origin/main` — confirmed accurate.
- `OrgActivityTab`'s thin-looking unit test is backed by real coverage one layer down (`AuditLogViewer.test.tsx`) and one layer up (`OrganizationRecordPage.test.tsx`'s `#activity` case against the real fetch) — no gap.

```
pnpm --filter @breeze/web exec tsc --noEmit    → clean
npx astro check                                 → 0 errors, 0 warnings (252 pre-existing hints)
pnpm --filter @breeze/web lint                  → clean

apps/web: 36 files / 641 tests across the W02 surface + i18n + no-silent-mutations → pass (+8 from this review round)
```

Not merging — the call is yours.

PR #5112, opened 2026-09-06
fix(mobile): ticket photo upload FormData part + library permission prompt (#5103)

### LINKED ISSUE #5103: mobile: ticket photo upload fails (Expo winter fetch rejects RN FormData part) + library picker permission prompt
From the 2026-09-06 OliveTech iOS TestFlight round-2 recording (build with #4916).

## 1. Photo upload fails before the request leaves the phone
Chip text now surfaces the real cause: `Upload failed. Check your connection and try again. (Unsupported FormDataPart implementation)`.

**Root cause (verified):** Expo SDK 57 installs `expo/fetch` as `globalThis.fetch` (`node_modules/expo/src/winter/runtime.native.ts:52`, opt-out `EXPO_PUBLIC_USE_RN_FETCH=1`). Its `convertFormDataAsync` (`winter/fetch/convertFormData.ts`) only accepts `string | Blob | {bytes()}` parts and throws `Unsupported FormDataPart implementation` on the React Native `{ uri, name, type }` shape. `attachmentFilePart()` in `apps/mobile/src/services/ticketAttachmentContract.ts` builds exactly that shape and `uploadTicketAttachment()` in `apps/mobile/src/services/ticketAttachments.ts` appends it. This is why prod had 0 `ticket_attachments` rows and 0 Caddy hits for three days.

**Fix:** append a real Blob — `new File(uri)` from `expo-file-system` (SDK 54+ `File` implements Blob and `bytes()`), with the filename as the third `append` argument, or use `File.uploadTask`. Keep the pre-send size/MIME gates. Update the contract test that asserts the part shape.

## 2. Library picker asks for full Photo Library access, then toasts a vague denial
`pickFromLibrary()` calls `ImagePicker.requestMediaLibraryPermissionsAsync()` before `launchImageLibraryAsync()`. On iOS 14+ the picker is `PHPickerViewController`, which needs no permission, so the "would like full access to your Photo Library (7,308 photos)" prompt is avoidable. When the user declines, `TicketDetailScreen.tsx:278` toasts "Breeze needs permission to use that. Enable it in Settings." — "that" is vague and there is no way to open Settings.

**Fix:** skip the permission request for the library picker on iOS (keep it for the camera); name the thing in the toast ("Camera access is off for Breeze"); add an Open Settings action (`Linking.openSettings()`).

ToddHebebrand:
## Summary

- Expo SDK 57 installs `expo/fetch` as `globalThis.fetch`; its `convertFormDataAsync` (`winter/fetch/convertFormData.ts`) only accepts a multipart part that is a `Blob` or exposes `bytes()`. `attachmentFilePart()` built React Native's own `{ uri, name, type }` shape, which throws `Unsupported FormDataPart implementation` — so every ticket photo/PDF upload failed before the request left the phone (root cause verified in the issue: 0 `ticket_attachments` rows, 0 Caddy hits for 3 days).
- Fix: wrap the picked URI in an `expo-file-system` `File` (implements both `Blob` and `bytes()`), and pass the filename as the third `form.append()` argument instead of relying on the name `File` derives from the (often OS-generated) cache URI. Client-side size/MIME gates and the multipart boundary handling (never hand-set `Content-Type`) are unchanged.
- Also: stop requesting Photo Library permission before `launchImageLibraryAsync` on iOS — the picker runs as `PHPickerViewController`, a sandboxed process needing no permission, so the request only produced an avoidable "full access to your Photo Library (N photos)" prompt. Android keeps the request (older API levels still gate on it).
- The permission-denied alert now names the capability ("Camera access is off for Breeze" / "Photo Library access is off for Breeze") instead of the generic "that", and offers an **Open Settings** action (`Linking.openSettings()`) via a native `Alert` instead of the auto-hiding toast, which has no way to act on a denial it can only be fixed in Settings.

## Root cause & fix verification

- Confirmed against `node_modules/expo/src/winter/fetch/convertFormData.ts`: the old shape hits neither `entry instanceof Blob` nor `'bytes' in entry` and falls into the `throw`.
- New tests feed both the old shape and `attachmentFilePart()`'s fixed output through the **real**, unmocked `convertFormDataAsync` (dynamically imported so `tsc` doesn't have to type-check that internal module's own pre-existing type error) — proving the regression and the fix against the actual buggy code path, not a reimplementation of its rules.
- Confirmed red-first: reverted the production changes and re-ran the new/changed tests — all 4 failed for exactly the reasons the issue describes (`Unsupported FormDataPart implementation`, and the iOS library permission request firing). Restored the fix — all 48 tests in the suite pass.

## Test plan

- [x] `cd apps/mobile && npx vitest run src/services/ticketAttachments.test.ts` — 48 passed
- [x] `cd apps/mobile && npx vitest run src/screens/tickets/ src/services/ticketAttachments.test.ts src/services/notifications.test.ts` — 196 passed
- [x] `cd apps/mobile && npx tsc --noEmit` — clean
- [x] Verified red-first: stashed the production fix, confirmed the new/updated tests fail against the old code, then restored the fix and re-ran green.

Closes #5103


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr — code-reviewer, silent-failure-hunter, pr-test-analyzer (all Sonnet, parallel)

**Findings:** 2 confirmed, both fixed in a1b36862e7:
- silent-failure-hunter (CRITICAL): `void Linking.openSettings()` discarded the promise — a rejection (no reachable Settings intent) would leave the alert's "Open Settings" button doing nothing with no feedback. Fixed: `.catch()` → `reportInternalError` + toast.
- code-reviewer (~80% confidence): the "real converter" regression test's `expo-file-system` File mock extends `Blob` (needed so Node's own stricter `FormData.append` 3-arg form accepts it elsewhere in the suite), so it only proved `convertFormDataAsync`'s `instanceof Blob` branch, not the `'bytes' in entry` branch a real on-device File actually takes. Added a second fixture modelling the genuine (bytes()-only, non-Blob-instance) shape to cover that branch too. Tracing this also surfaced that `form.append`'s 3rd filename argument is silently dropped by React Native's own FormData patch for a non-Blob value — `File.name` (a read-only, URI-derived getter) is what actually reaches the server. Fixed by shadowing `.name` with an own property carrying the picker's real filename, kept the 3rd arg as a harmless fallback.

pr-test-analyzer: no blocking gaps — flagged that `TicketDetailScreen.tsx`'s new Alert/Settings UX has zero test coverage, but that's a structural limitation (mobile's Vitest config only runs `.test.ts`, no `.tsx` RN test runner exists in this app) not something fixable in this PR's scope.

**Tests:** `apps/mobile` — `npx vitest run src/services/ticketAttachments.test.ts` 50/50 passed; broader batch (`src/screens/tickets/`, `ticketAttachments.test.ts`, `notifications.test.ts`) 198/198 passed. `npx tsc --noEmit` clean. Verified red-first for every new/changed assertion (stashed the production fix and confirmed the 4 original tests failed for the exact reasons in the issue; separately confirmed the filename-shadow test fails without the `defineProperty` fix).

PR #5113, opened 2026-09-06
fix(mobile): metrics key mismatch, keyboard persists, no org on device rows, timesheet 0m (#5104)

### LINKED ISSUE #5104: mobile: device metrics key mismatch, keyboard persists after login/MFA, no org on device rows, timesheet header 0m while loading
From the 2026-09-06 OliveTech iOS TestFlight round-2 recording. Four small, mechanical mobile fixes, all verified against code.

## 1. Device Details metrics always "–%" for online devices
`getDeviceMetrics()` in `apps/mobile/src/services/api.ts:1126` reads `avgCpuPercent / avgRamPercent / avgDiskPercent` from `GET /devices/:id/metrics`, but the route returns `cpu / ram / disk` (`apps/api/src/routes/devices/metrics.ts:189-193`, `aggregateMetricsByInterval`). It also takes `data[data.length - 1]` as "latest" — confirm the route's ordering before trusting that.

## 2. Keyboard stays open across navigation after sign-in and MFA verify
`MfaChallengeScreen.handleVerify` and `LoginScreen.handleLogin` dispatch without `Keyboard.dismiss()`; the navigator swap unmounts the input and iOS leaves the number pad covering the Home screen. `apps/mobile/src/navigation/ApprovalGate.tsx:74` already does the dismiss pattern. Also dismiss the keyboard when a chat message is sent from `HomeScreen`.

## 3. Device list rows and Device Details show no organization
`mapDevice()` in `apps/mobile/src/services/api.ts:618` maps `siteName` but never `organizationName`, so the org/site rows in `DeviceDetailScreen.tsx:320-332` never render. The list rows in `DevicesListScreen.tsx` show only OS + relative time — on a five-org MSP tenant that is not enough. Map `organizationName` (check what the mobile device list endpoint returns; add it there if missing) and show `org · site` as the row meta line.

## 4. Time tab header shows "0m total · 0m billable" while the week is loading
`TimesheetScreen.tsx:397` renders `view.totals` before the request resolves (body shows a spinner, header says 0m; after load it read 12h 33m). Render a placeholder ("—") until loaded. While there: the week list is ascending so today's entry is below the fold — anchor/scroll to today, or list most recent day first.

ToddHebebrand:
## Summary

Four small, independently-verified mobile defects from the 2026-09-06 OliveTech TestFlight round-2 recording. All confirmed against code (see issue for file:line references).

1. **Device Details metrics always "–%"** — `getDeviceMetrics()` read `avgCpuPercent`/`avgRamPercent`/`avgDiskPercent`, but `GET /devices/:id/metrics` (`aggregateMetricsByInterval`) sends buckets keyed `cpu`/`ram`/`disk`. Fixed the field names; confirmed via `apps/api/src/routes/devices/metrics.ts` that buckets are ordered ascending, so taking the last element as "latest" is correct.
2. **Keyboard persists after login/MFA/chat send** — `MfaChallengeScreen.handleVerify`, `LoginScreen.handleLogin`, and `HomeScreen.handleSend` never dismissed the keyboard before dispatching/navigating. Added `Keyboard.dismiss()` to all three, matching the existing pattern in `ApprovalGate.tsx`.
3. **No organization on device rows/detail** — `mapDevice()` never read an org name at all. The single-device endpoint (`devices/core.ts`) sends it as `orgName`; the mobile list endpoint (`mobile.ts`) sent no org field. Added an `organizations` leftJoin + `organizationName` to the list endpoint, and map both `organizationName`/`orgName` in `mapDevice()`. Device list rows now show `org · site` (falling back gracefully to OS-only if both are absent).
4. **Timesheet header shows "0m total · 0m billable" while loading** — the header rendered `view.totals` unconditionally. Gated it on `entries !== null` (the existing load-state signal) to show a `—` placeholder until the week has actually loaded.
   - Also switched the day list to **most-recent-day-first** so today's entries on a mid/late week aren't below the fold. Picked this over anchor/auto-scroll-to-today because the screen has no ScrollView layout tracking today, and reversing the render order is a much smaller, lower-risk change.

## Test plan

- [x] New unit tests, red-then-green verified by hand (stash/pop the fix, confirm the test fails, restore, confirm it passes) for all four defects:
  - `apps/mobile/src/services/api.deviceMapping.test.ts` (metrics field mapping + org mapping from both endpoint shapes)
  - `apps/mobile/src/screens/auth/keyboardDismiss.test.ts` (source-parses the three screens — no RN test runtime is configured for this project — and asserts `Keyboard.dismiss()` runs before the dispatch/turn-start call)
  - `apps/mobile/src/screens/time/timesheetHeaderAndOrder.test.ts` (header placeholder guard + day-order source assertions)
- [x] Extended the existing `GET /mobile/devices` suite in `apps/api/src/routes/mobile.test.ts` for the new `organizations` leftJoin (updated mock chains to the repo's existing `mockSelectLeftJoinChain(Capturing)` helpers, added an `organizationName` assertion)
- [x] `apps/mobile`: full suite green — `npx vitest run` → 100 files / 1259 passed, 3 skipped
- [x] `apps/mobile`: `npx tsc --noEmit` clean
- [x] `apps/api`: `npx tsc --noEmit` clean (12GB heap)
- [x] `apps/api`: `mobile.test.ts` (75 passed / 2 skipped), `mobile.cursor.test.ts`, `devices/metrics.test.ts` green

Closes #5104

https://claude.ai/code/session_01X3tHqW2PkSAoiTHf1LN8V9


ToddHebebrand:
**Review run:** /pr-review-toolkit:review-pr (code-reviewer, silent-failure-hunter, pr-test-analyzer — parallel, Sonnet)

**Findings:**
- code-reviewer: 0 findings ≥80% confidence. Verified the new `organizations` leftJoin in `GET /mobile/devices` doesn't participate in `whereCondition` (tenant scoping unaffected), is a proper LEFT JOIN on a PK (no row fan-out, cursor keyset unaffected), and confirmed the `cpu`/`ram`/`disk` field-name fix against `aggregateMetricsByInterval`'s actual output + ascending bucket order.
- pr-test-analyzer: 1 addressed — `DevicesListScreen.tsx`'s row-meta line (`org · site` fallback) had zero test coverage despite being pure branching logic. **Fixed in 160a68ec6**: extracted to a testable `deviceRowMeta()` helper in `deviceListFilters.ts` (this screen's existing home for pure, RN-import-free list logic) with unit tests covering all 5 branches (org+site, org-only, site-only, OS fallback, full em-dash fallback). Two minor test-quality notes (day-order test is a literal-string check rather than semantic; header-guard test doesn't assert the placeholder's exact content) were left as-is — both tests are still genuinely discriminating (fail pre-fix), and strengthening them further is diminishing returns for a small mechanical PR.
- silent-failure-hunter: 4 findings, all evaluated and declined as scope creep / non-issues for this PR — (1) suggesting error-logging when the new leftJoin's org lookup misses (speculative "shouldn't happen" case, no evidence it occurs), (2) noting the UI fallback could mask that same hypothetical anomaly (accepted tradeoff, already commented in code), (3) suggesting a runtime guard against a *future* metrics field-name drift (forward-looking, not a bug in this PR), (4) `||` vs `??` in `mapDevice()`'s new `organizationName` fallback — checked against the immediate function: every other optional field in `mapDevice()` (hostname, os, lastSeen, organizationId, siteId, siteName) already uses `||`, so `??` here would be inconsistent with its own surrounding code, not more consistent; declined.

**Tests:** `apps/mobile` full suite green (`npx vitest run` → 100 files / 1264 passed, 3 skipped); `apps/mobile` `npx tsc --noEmit` clean. `apps/api` `mobile.test.ts` green (75 passed / 2 skipped, includes the extended `GET /mobile/devices` coverage for the new leftJoin); `apps/api` `npx tsc --noEmit` clean (12GB heap). All four original bug fixes were verified red→green by hand (stash the fix, confirm the new test fails, restore, confirm it passes) before this PR was opened.

PR #5121, opened 2026-09-06
fix(ai,mobile,web): approved-by-worker is not a FAILED tool result (#5107)

### LINKED ISSUE #5107: ai/mobile: approved-by-worker rendered as FAILED, prompt double-confirms before approval gate, raw tool names in chat
From the 2026-09-06 OliveTech iOS TestFlight round-2 recording. The approval round trip works (overlay takeover, approve on phone, worker executes) but what the user sees afterwards is wrong.

## 1. After approving on the phone, the chat shows `MANAGE_SERVICES · FAILED` in red
For tools in `DURABLE_RELEASE_ONLY_TOOLS`, `apps/api/src/services/aiAgentSdk.ts:1152` returns `{ allowed: false, error: 'This action was approved and is being completed by the approval worker.' }` after the human approves. That becomes a `tool_result` with `isError: true`, so mobile `ToolIndicator.tsx` (and the web tool row) paint it as a failure, and the model narrates "the restart is already being processed… give it a moment". The user approved and then read FAILED.

**Contract:** this outcome is not an error. The SSE `tool_result` for it must carry a machine-readable marker (e.g. `output.status: 'approved_executing'` with `isError: false`, or a dedicated event), the text the model receives should say the action is approved and executing and the result will be reported separately, and mobile `ToolIndicator` + web tool row render `APPROVED · RUNNING` in brand colour, not deny red. Pick the shape that keeps `isError` honest without relying on sniffing the error string (the mobile already sniffs for DENIED — replace that path if you touch it).

## 2. Double confirmation
The model asked "Shall I go ahead?" in prose, the user typed "Go ahead", and then the structured Approve/Deny takeover asked again. `aiAgentSystemPrompt.ts:9` says destructive operations "require explicit human confirmation", which the model reads as "ask in chat". The approval gate already enforces the human step. Change the instruction so the model calls the tool and lets the gate prompt (say what it is about to do, do not ask permission in prose), and only reports a denial if the call is rejected. Keep the change to that section; update `aiAgentSystemPrompt.test.ts`.

## 3. Tool status rows are raw tool names
Rows read `MANAGE_ALERTS · COMPLETED`, `GET_FLEET_FINDINGS · COMPLETED`. `ToolIndicator.format()` is `toUpperCase()`. Add a human label map (shared in `packages/shared` if the web can use it too, else local) so completed rows read like "Checked alerts · Done" / "Searched logs".

## Out of scope (file separately if not already tracked)
The chat never confirms whether the worker-executed restart actually completed; a completion event from the durable release worker back into the session is a follow-up.

ToddHebebrand:
## What

After a human approves a tier-3 action intent, the live chat session is not always the side that runs it: the durable release worker (`jobs/intentReleaseWorker.ts`) also consumes the `intent_approved` outbox and races the session for the `approved -> executing` CAS. Two exits in `createSessionPreToolUse` hand the action off to that worker, and both returned `{ allowed: false, error }` — which the tool layer published as `isError: true`.

So a user who had just tapped **Approve** on their phone read `MANAGE_SERVICES · FAILED` in deny-red, and the model narrated it as a failure.

### 1. The approval handoff is a first-class, non-failure outcome

`@breeze/shared/utils/aiToolHandoff.ts` owns the wire contract (`approved_executing` + `isAiToolHandoffOutput`); `apps/api/src/services/aiToolHandoff.ts` adds the model-facing text and re-exports the rest.

- Both handoff exits now return `handoff: 'approved_executing'`.
- One new helper (`preToolUseDenialResult`) decides `isError` for **all three** denial sites in `aiAgentSdkTools.ts` — they previously hard-coded `true` at each — publishing `{ status, message }` with `isError: false` for a handoff and the unchanged `{ error }` / `isError: true` for a real denial.
- Clients switch on the **status field**, never on message text.

**Why the coordination invariant still holds** (the "never double-execute" guarantee — comments around lines 1140–1230):

- Neither exit mutates the intent. The `requiresDurableRelease` branch still returns **before** the `approved -> executing` CAS is attempted, so the worker's claim stays available and the intent is never stranded in `executing`. The lost-CAS branch still refuses to execute after losing, and the winner owns every later transition. Only the *reporting* changed, not the control flow.
- Flipping `isError` cannot complete an intent by accident. Both exits return **before** `pendingIntentBySession.set(...)` (line ~1329) and while the ledger row is still `status: 'pending'`. So in `postToolUse`: the intent CAS block is skipped (no `pendingIntentId`), and the tier≥2 ledger UPDATE is scoped `WHERE status = 'executing'` and matches nothing. Both no-op exactly as they did before.
- The **audit** row is the one place `isError: false` would have implied something false, so the event now stamps `details.toolOutcome: 'approved_executing'` rather than letting an absent `result` read as "`ai.tool.manage_services` succeeded".

**Worth knowing:** `DURABLE_RELEASE_ONLY_TOOLS` is still empty, so the exit the issue cites (`aiAgentSdk.ts:1152`) cannot fire today. The exit that actually produced the recording is its sibling — the **lost-CAS** branch, reached when the worker wins the race. Same user-visible defect, same semantics; both are fixed.

### 2. Mobile renders `APPROVED · RUNNING` in brand colour

Presentation logic moved into a pure `toolIndicatorLogic.ts` because this app has no React Native test runtime (`vitest.config.ts` deliberately includes only `.ts`, so a `.tsx` suite would pull RN/Expo into Vitest). The handoff is checked **before** `isError`, so a stale server or a history row persisted before this fix cannot resurrect the red row. The existing DENIED text sniff stays — an approval *rejection* has no wire marker yet.

`AiMessage.tsx` needed one change: it routed on `t.isError` and dropped `output` on the non-error path, which would have swallowed the handoff into a "· DONE" row.

### 3. Web gets the equivalent state

`AiToolCallCard.tsx` gains an amber `ShieldCheck` + "Approved · running" branch, ahead of the `isError` branch for the same stale-server reason. Added its **first** co-located test — the component had none, and `AiChatMessages.test.tsx` mocks it out entirely.

### 4. Tool rows read as actions, not identifiers

`aiToolLabel` conjugates the leading verb (`manage_alerts` → "Updated alerts", `get_fleet_findings` → "Checked fleet findings", `run_script` → "Ran script") and falls back to readable title case for anything unmapped, so registering a new tool never requires touching the table. There are ~200 tools and a hand-written per-tool dictionary would be stale within a release.

Shared for web; **mirrored** in mobile, which has no `@breeze/shared` dependency by design (same constraint already documented on `services/ticketPushPrefs.ts`). Drift there degrades a caption, never behaviour.

### 5. System prompt no longer asks twice

Rule 3 said destructive operations "require explicit human confirmation", which the model read as *ask in chat* — so the user confirmed once in prose ("Go ahead") and again at the structured Approve/Deny takeover. It now says the approval gate **is** the human step: state what you are about to do, make the call, report a denial only if the call is rejected. New rule 3b tells the model an approved-and-executing result is not a failure.

3b is written in prose rather than as the `approved_executing` literal: `mcpGuidancePromptTools.test.ts` reads every snake_case token in these instructions as a tool name that must exist in the registry, and it caught the first draft.

## Verification

All foreground, single-fork where relevant.

| | |
|---|---|
| `apps/api` affected (20 files: `aiAgentSdk*`, `aiAgentSdkTools*`, `aiToolHandoff*`, `aiAgentSystemPrompt`, `mcpGuidance*`, `durableReleaseOnly`, `aiToolOutput`) | 309 passed |
| `apps/web` (`components/ai/`, `localeParity`, `processStreamEvent`) | 156 passed |
| `apps/mobile` full suite | 1260 passed, 3 skipped |
| `packages/shared` full suite | 2484 passed |
| `apps/web` `astro check` | 0 errors |
| `apps/mobile` `tsc --noEmit` | clean |
| `packages/shared` `tsc --noEmit` | clean (2 pre-existing `quotes.test.ts` errors, confirmed on `origin/main` by stashing) |
| eslint on every touched source file | clean |

Red-first throughout. The API and prompt assertions were watched failing against the unfixed code. The web card tests were written after the component edit, so they were **mutation-checked**: reverting `isApprovedExecuting` to `false` and the label back to the old title-case inline turned 5 of 7 red.

`approvedRunning` added to all 8 locale files (`localeParity` green).

## Out of scope

The chat still never confirms whether the worker-executed action *completed* — a completion event from the durable release worker back into the session. The issue calls this out as a follow-up and it is not attempted here.

Closes #5107

https://claude.ai/code/session_01Sf7ZBJ7j2WAAdEnJe8CJ6G


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer (parallel, against `gh pr diff` + `origin/main`, no worktree handed out).

**Findings:** 4 raised → all 4 addressed in 8c7bd05a43; 0 outstanding.

1. **The handoff marker was read back out of the tool's own output** (silent-failure-hunter, critical). `isAiToolHandoffOutput(output)` is a bare shape check, and a tool owns its payload — a buggy one or a third-party extension emitting `{ error: 'restart failed', status: 'approved_executing' }` would have had its genuine failure painted as an approved, in-flight action. On mobile that is unrecoverable: the row has no expand affordance, so the error text is unreachable in the UI. The same sniff also let a tool forge its own `details.toolOutcome` audit stamp and suppress its own real data block in `AiMessage`'s routing. **Fixed** by threading a server-asserted `handoff` field from the gate's decision through `PreToolUseCallback` → `safePostToolUse` → `PostToolUseCallback` → the SSE `tool_result` event → both clients; `postToolUse` no longer inspects `parsedOutput` for it. `output.status` survives as a history-replay fallback (the SSE field isn't persisted on the message row) but gated on `!isError`. Nothing is lost by the gate — a server predating this contract sends no `status` field at all, so the "stale server" case the ungated check claimed to rescue never existed.

2. **The handoff audit row read as SUCCESS** (code-reviewer). `result` defaults to `'success'` when omitted (`auditEvents.ts:104`), so flipping `isError` to false told anyone auditing "did the restart happen?" that `ai.tool.manage_services` succeeded — and `result` is the indexed column audit queries filter on, so `details.toolOutcome` was not a substitute. **Fixed:** `result: 'dispatched'`, the `AUDIT_RESULTS` value that already exists for exactly this "handed to another execution path, outcome not yet known" case (its own comment says it "must never be conflated with 'success'") and is used identically by `commandQueue.ts` at enqueue time.

3. **Coverage gaps** (pr-test-analyzer). `preToolUseDenialResult` is called from three near-identical blocks and only `wrapExtraToolWithHooks` was covered — while `manage_services`, the tool in the recording, registers through `makeHandler`. **Fixed:** `makeHandler` and `makeSessionAwareHandler` now exercised via `__test__`. The audit stamping had zero tests; now covered including the negative (a tool emitting the literal in its own output stamps nothing). The mobile/web assertions were inverted to prove the spoof is rejected rather than honoured. The remaining flagged gap — `AiMessage.tsx` routing itself — stays uncovered: this app has no RN test runtime (`vitest.config.ts` includes only `.ts` so component imports never pull RN/Expo in), which is why the logic lives in a pure module at all.

4. **Type design** (type-design-analyzer, rated the wire contract 8–9/10 and recommended one change). **Applied:** `approvedExecutingDenial()` pairs `error` + `handoff` in one constructor, so a future third call site cannot pair the marker with a real failure string. The full discriminated-union split was **rejected on merit** — it would change `failMatchedPlanStep`'s generic bound for ~15 unrelated exits on a critical approval path, for a risk this constructor closes.

**Verification (foreground, after the fixes):** `apps/api` affected 20 files 315 passed · `apps/web` `components/ai/` + `processStreamEvent` + `localeParity` 158 passed · `apps/mobile` full suite 1262 passed/3 skipped · `packages/shared` full suite 2484 passed. `tsc --noEmit` clean on api/mobile/shared (2 pre-existing `quotes.test.ts` errors, confirmed on `origin/main`); `astro check` 0 errors; eslint clean on every touched file.

Both new defences were **mutation-checked**: removing the `!isError` gate and the `'dispatched'` result each turn the new assertions red.

**Coordination invariant re-verified independently by two reviewers** — neither handoff exit sets `pendingIntentBySession` or moves the ledger row off `'pending'`, so `postToolUse`'s intent CAS and its `WHERE status='executing'` update stay no-ops. Flipping `isError` cannot complete an intent. One reviewer also noted a secondary bug this fixes: every approval handoff previously tripped the one-time session auto-flag.

**Status:** review-clean, awaiting maintainer merge. Not merging and not closing #5107 — both are the maintainer's call.

ToddHebebrand:
`a47844e0b0` — import the handoff contract from the `@breeze/shared` **root** instead of `@breeze/shared/utils/aiToolHandoff`. That subpath isn't in the package's `exports` map, so integration shards 1/3/4 died at module load (`ERR_PACKAGE_PATH_NOT_EXPORTED`) taking out aiAgentGraduation, secretBearingToolSeal and aiAgents.routes; the unit job resolves the package from source and never saw it. It was also the file's only *value* import from a deep path — the neighbouring `import type … from '@breeze/shared/types/ai'` ones are type-only and never reach the resolver, which is what made the deep form look safe. Exports map deliberately not widened: both symbols are already re-exported from the root, and `aiAgentSdkTools.ts` pulls the same barrel anyway. Import path only, no behaviour change — apps/api tsc clean, 285 affected unit tests green.

Issue #5123, opened 2026-09-06
ci-red: main integration shards 1/2/4 fail after config-policy W01 (#5099) — parent_policy_id FK unpinned in erasure ledger + policy owner-move serialization tests


ToddHebebrand:
Every PR opened after 22:00Z on 2026-09-06 fails the same integration shards in code the PRs never touch (seen on #5113, #5114 mobile-only, and the org-record wave PRs #5108, #5109). The failures appear with **#5099** (config-policy W01, merged 22:01Z, `f6ec8a33f`), which added `configuration_policies.parent_policy_id` and constraint triggers. Main's own run on `754945a41` is still in progress at filing time; shards 1/2/4 are expected to fail identically.

**Shard 1 — `orgCascadeFkOnDelete.integration.test.ts` › "every FK erasure does not handle is pinned in the ledger"**
```
+ "configuration_policies(parent_policy_id) -> configuration_policies [configuration_policies_parent_policy_id_fkey, ON DELETE NO ACTION, self-ref-open-row-set]"
```
The self-referential FK is `NO ACTION` and not registered in the erasure ledger. The test's own guidance, in order of preference: (1) `ON DELETE CASCADE` / `SET NULL` in a NEW idempotent migration (SET NULL = orphaned children become roots, CASCADE = children go with the parent — the inheritance plan should say which); (2) register in `CORE_ORG_CASCADE_DELETE_ORDER` + every other list; (3) pre-clear + pin in `ORG_CASCADE_FK_PRE_CLEARED`. See `orgCascadeFkOnDeleteAllowlist.ts`.

**Shard 2 — `configPolicyReferenceGateConcurrency.integration.test.ts` › "serializes reverse writes to sensitive-data candidates with one UUID"** and **`partnerApiRls.integration.test.ts` › partner reconstruction export RLS traversal** (3 tests: "enforces every target level and rejects reverse owner forges", "serializes concurrent target and policy owner moves before assignment validation", "serializes complete bulk owner sets across partners and descending organizations").

**Shard 4 — `configPolicyReferenceConcurrency.integration.test.ts`** › "serializes an automation link insert against a referenced policy owner move", "makes a link insert wait for and reject a committed referenced-policy owner move".

The shard 2/4 failures are all owner-move / reference-gate serialization tests on `configuration_policies`, consistent with the new constraint triggers changing lock/visibility behaviour on that table. Not diagnosed further here.

[...]

PR #5124, opened 2026-09-06
test(integration): fix shard-2 red from #5099's config-policy owner guard

ToddHebebrand:
## Root cause

Main has been red on **Integration Tests shard 2/4** since #5099 (`f6ec8a33f8`, config-policy inheritance W01 of #5080). That PR's migration `2026-10-12-100000-config-policy-inheritance.sql` adds constraint trigger `configuration_policies_parent_guard` (function `breeze_config_policy_parent_guard()`), which now raises **`23514`** (`CONSTRAINT = configuration_policies_owner_immutable`) whenever `configuration_policies.org_id`/`partner_id` changes outside system scope ("configuration policy ownership can only change in system context").

Two pre-existing integration suites forge configuration-policy owner moves as a non-system actor and expected the composite-FK sqlstate `23503` — the new trigger now pre-empts that FK check, so those assertions (and one dependent 30s test timeout) went red.

## Per-assertion decisions

- **`partnerApiRls.integration.test.ts` — "enforces every target level and rejects reverse owner forges"**: the two `configuration_policies` owner-forge assertions (`admin.update(...).set({orgId/partnerId})`, run without system scope) now legitimately hit the new guard first. Kept the test's intent (forge rejected) but reasserted the actual outcome — added a `captureSqlError` sibling helper (mirrors `configPolicyInheritance.integration.test.ts`'s `expectSqlState` shape) that checks both sqlstate `23514` *and* `constraint_name: 'configuration_policies_owner_immutable'`, so the assertion still discriminates from an unrelated 23514.
- **`partnerApiRls.integration.test.ts` — "serializes concurrent target and policy owner moves before assignment validation"**: the `policyMover` sub-block moves a policy's `org_id` and expects that move to **succeed** (an org-merge shape) so it can hold the row lock while a concurrent assignment insert serializes against it. Without system scope the update now throws before `ownerMove.resolve()` runs — explaining the observed 30s timeout, not a real hang. Fixed by electing `breeze.scope = system` inside the transaction before the update, matching org merge's real code path (org merge always runs in system context per the tenancy contract).
- **`partnerApiRls.integration.test.ts` — "serializes complete bulk owner sets across partners and descending organizations"**: same shape — a bulk `partner_id` update on `configuration_policies` expected to resolve. Wrapped it in a transaction that elects system scope first.
- **`configPolicyReferenceGateConcurrency.integration.test.ts` — "serializes reverse writes to sensitive-data candidates with one UUID"**: this test's actual purpose is exercising the pre-existing (2026-08-01) feature-reference advisory-gate race, not the new ownership guard (confirmed by the advisory-lock class/backend-PID wait it asserts on). Elected system scope for the second writer's `configuration_policies` update so the original FK-based race (`23503`) is still what's under test — consistent with the sibling test earlier in the same file, which already elects `breeze.scope=system` for its writers.

## Design note

No production code was changed. The new guard's precedence over the composite FK looks correct as shipped: a non-system-actor ownership move should fail fast and specifically (`23514` with a named constraint) rather than only incidentally via whatever FK happens to reference the row. These suites were asserting the older, less specific failure mode from before #5099 existed.

## Test output (real Postgres, `pnpm test-stack up`)

```
partnerApiRls.integration.test.ts + configPolicyReferenceGateConcurrency.integration.test.ts
 Test Files  2 passed (2)
      Tests  20 passed (20)

configPolicyInheritance.integration.test.ts (#5099's own suite — no regression)
 Test Files  1 passed (1)
      Tests  40 passed (40)

All three together: 60/60 passed
```

`npx tsc --noEmit` on `apps/api` shows no errors in either touched file.

Refs #5080

### DIFF
--- a/apps/api/src/__tests__/integration/configPolicyReferenceGateConcurrency.integration.test.ts
+++ b/apps/api/src/__tests__/integration/configPolicyReferenceGateConcurrency.integration.test.ts
@@ -323,6 +323,12 @@ runDb('serializes reverse writes to sensitive-data candidates with one UUID', as
       'sensitive-data mover',
     );
     secondWork = sqlState(() => second.begin(async (tx) => {
+      // #5099's ownership-immutable guard rejects a configuration_policies
+      // owner change outside system scope (23514) before this statement ever
+      // reaches the reference-gate race this test exercises. Elect system
+      // scope so the pre-existing FK/gate serialization (23503) is still
+      // what's under test here, not the newer guard.
+      await tx`SELECT pg_catalog.set_config('breeze.scope', 'system', true)`;
       const [backend] = await tx<{ pid: number }[]>`SELECT pg_catalog.pg_backend_pid() AS pid`;
       if (!backend) throw new Error('missing sensitive fallback backend');
       secondEntered.resolve(backend.pid);
--- a/apps/api/src/__tests__/integration/partnerApiRls.integration.test.ts
+++ b/apps/api/src/__tests__/integration/partnerApiRls.integration.test.ts
@@ -590,12 +590,18 @@ describe('partner reconstruction export RLS traversal', () => {
     expect(await captureSqlState(() => admin.delete(devices)
       .where(eq(devices.id, movableDevice.id))))
       .toBe('23503');
-    expect(await captureSqlState(() => admin.update(configurationPolicies)
+    // #5099 (2026-10-12-100000-config-policy-inheritance.sql) added a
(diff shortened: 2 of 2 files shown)


ToddHebebrand:
Superseded by #5127 (fixes shards 1, 2 and 4 together under #5123; this PR covered the shard-2 suites only). The per-assertion decisions here match #5127's approach: named-constraint 23514 for non-system owner forges, system scope for the moves that must succeed.

PR #5127, opened 2026-09-06
fix(config-policy): pin parent_policy_id in the erasure ledger; scope owner-move races to system (#5123)

### LINKED ISSUE #5123: ci-red: main integration shards 1/2/4 fail after config-policy W01 (#5099) — parent_policy_id FK unpinned in erasure ledger + policy owner-move serialization tests

ToddHebebrand:
Closes #5123

Main's integration shards 1/2/4 have been red since **#5099** (config-policy W01, `f6ec8a33f`), blocking every open PR. Two distinct consequences of the same wave. **No production defect was found in either**, so this PR is test/ledger-only: **no migration, no schema change, no behaviour change.** The shipped migration `2026-10-12-100000-config-policy-inheritance.sql` is untouched, and the design's `NO ACTION` decision is preserved.

---

## Shard 1 — `orgCascadeFkOnDelete` › "every FK erasure does not handle is pinned in the ledger"

**Unpinned edge:** `configuration_policies(parent_policy_id) -> configuration_policies [configuration_policies_parent_policy_id_fkey, ON DELETE NO ACTION, self-ref-open-row-set]`

**Root cause.** W01 added the self-referential `parent_policy_id` FK. The classifier's branch (e) calls a self-reference safe only when the table's `org_id` is `NOT NULL`; `configuration_policies.org_id` is nullable (partner-wide policies, epic #2135), so from the catalog alone it cannot tell whether `DELETE ... WHERE org_id = $1` removes a row set closed under the self-reference.

**Decision: the invariant holds, so this is a ledger entry — not a migration.**

The W01 `DEFERRABLE` constraint trigger `configuration_policies_parent_guard` (migration L100–140) requires a child's owner axis to match its parent's: an org-owned child names a same-org parent or its partner's partner-wide baseline; a partner-wide child names only a partner-wide parent of the same partner. Closure needs exactly one thing — no **surviving** row may reference a deleted one — and the surviving rows are precisely the partner-wide ones and other orgs' rows, neither of which the trigger lets point at this org's rows. Ownership itself moves only in system context, and the trigger revalidates the whole family at **COMMIT** (org merge runs `SET CONSTRAINTS ALL DEFERRED`), so a cross-owner edge cannot be committed in the first place — including via org merge or a partner re-point (`organizations_partner_config_policy_guard`).

The two cheap fixes the test prefers are both ruled out by the design (`2026-09-06-config-policy-inheritance-design.md` §Deletion, L123–135): `SET NULL` would silently un-configure every child when a baseline is deleted, and deleting a parent alone must be **refused**, which `CASCADE` would turn into a silent mass delete.

**Ledger entry** (`orgCascadeFkOnDeleteAllowlist.ts`, `ORG_CASCADE_FK_UNSAFE`, sorted between `automation_runs -> automations` and `deployment_devices -> deployments`), `reason: 'self-ref-open-row-set'` (recomputed from the catalog, not chosen), `allColumnsNullable: true`, with a `note` that states the invariant, cites the trigger and migration, says why `SET NULL`/`CASCADE` are excluded, and points at the tests below.

**A note is not a proof, so the argument is checked.** New section 5 of `configPolicyInheritance.integration.test.ts` — "erasure row-set closure":

1. **SYSTEM scope still refuses a partner-wide child under an org-owned parent** → `23514 / configuration_policies_parent_guard`. This is the edge that would break closure (the child survives `WHERE org_id = $1`; its parent does not).
2. **SYSTEM scope still refuses a child of ANOTHER org's parent** → same. Together these are the *complete* set of surviving-row shapes.
3. **The real `cascadeDeleteOrg()`** over an org holding a parent+child family plus a child of a partner-wide baseline completes, and leaves exactly the partner-wide baseline. Deliberately the real cascade rather than a hand-written `DELETE`, so a later change to the erasure SQL is caught here too.

**Positive control (verified out-of-band, not committed):** with the guard disabled, forging that partner-wide child and then running the erasure `DELETE` raises exactly `23503 ... configuration_policies_parent_policy_id_fkey` — so the edge really is fatal, and the guard is the only thing preventing it. A committed test that disables a trigger would be fragile, so it is recorded here rather than in the suite.

---

## Shards 2 + 4 — five owner-move / reference-gate races got `23514` where they pinned `23503`

**Confirmed root cause.** W01 installed `configuration_policies_parent_guard` as an **AFTER ROW** constraint trigger. The feature-reference validators from `2026-08-01-a` are **AFTER STATEMENT** triggers (`aa_config_policy_feature_reference_policy_update`, …), and Postgres fires AFTER-STATEMENT triggers *after* all AFTER-ROW ones — so name ordering never mattered: the new guard now always precedes them. Its `configuration_policies_owner_immutable` branch rejects **any** `org_id`/`partner_id` change when `breeze_current_scope() <> 'system'` (23514), which is what these tests now see. The `BEFORE STATEMENT` advisory gate (`aaa_feature_reference_gate_*`, namespace 1000302) is unconditional, so the serialization each test models is unchanged — the racers still block and still wait exactly as before; only which guard rejects them moved.

Verified directly against Postgres 16 in the test stack:

```
ERROR:  23514: configuration policy ownership can only change in system context
CONTEXT: PL/pgSQL function breeze_config_policy_parent_guard() line 20 at RAISE
CONSTRAINT NAME: configuration_policies_owner_immutable
```
…and the same statement under `set_config('breeze.scope','system',true)` succeeds.

**Case (1) for all five — no W01 defect.** A sweep of every writer of `configuration_policies` found exactly one production update path, `services/configurationPolicy.ts::updateConfigPolicy`, which sets only `name`/`description`/`status` and never `org_id`/`partner_id`. The one real owner mover is org merge (`orgMergeRegistry.ts` L533), which already runs system-scoped under `SET CONSTRAINTS ALL DEFERRED` — the shape `orgLifecycleFoundations` and the W01 suite already prove commits. So no legitimate write is newly refused; the tests had pinned the *reachability* of a non-system owner move, which W01 deliberately removed.

**Fix: elect `breeze.scope = 'system'` in the transaction that moves ownership**, so each test keeps exercising the reference validator it was written for and keeps its original `23503` assertion, rather than being downgraded to assert the outer guard's code.

| Shard | Test | Change |
|---|---|---|
| 2 | `configPolicyReferenceGateConcurrency` › serializes reverse writes to sensitive-data candidates with one UUID | system scope in the second racer's tx (same pattern the file's own target-delete racer already uses) |
| 2 | `partnerApiRls` › enforces every target level and rejects reverse owner forges | **both** halves pinned: the non-system move now asserts `23514`, and a new system-scoped move asserts the reverse validator's `23503` — asserting only the 23514 would have quietly retired the test |
| 2 | `partnerApiRls` › serializes concurrent target and policy owner moves before assignment validation | system scope in `policyMover` (it was failing before `ownerMove` resolved → 30s timeout + unhandled rejection, not an assertion) |
| 2 | `partnerApiRls` › serializes complete bulk owner sets across partners and descending organizations | the bulk partner swap (expected to **succeed**) wrapped in a system-scoped tx |
| 4 | `configPolicyReferenceConcurrency` › serializes an automation link insert against a referenced policy owner move | system scope in the mover's tx |
| 4 | `configPolicyReferenceConcurrency` › makes a link insert wait for and reject a committed referenced-policy owner move | system scope in the holder's tx |

No test was deleted or skipped, and no assertion was weakened; net coverage is higher (`partnerApiRls` now pins the new guard as well as the old one, and section 5 is new).

---

## Verification (all foreground, live Postgres 16 via `pnpm test-stack`)

| Check | Result |
|---|---|
| `orgCascadeFkOnDelete`, `configPolicyReferenceGateConcurrency`, `configPolicyReferenceConcurrency` | 3 files, **21 passed** |
| `partnerApiRls`, `configPolicyInheritance` | 2 files, **56 passed** (3 new erasure-closure tests confirmed by name in verbose output) |
| `tenantCascade*`, `orgLifecycleFoundations`, all `configPolicy*` | 13 files, **158 passed** |
| `src/db/autoMigrate`, `migrationRlsScope`, `migrationGucAttributes` | 3 files, **1488 passed** |
| `eslint` on the 5 changed files | clean |
| `pnpm db:check-drift` | no drift, 692 migrations match the ledger |

Reproduced red first on this branch's base (`754945a41e`) before changing anything: the same 5 failures plus the unpinned-FK diff, matching #5113's run 34063336354.

**Not verified here:** `tenantExportErasureRoundtrip` / `tenant-export-policy` (no migration and no new column, so the export-policy contract cannot be affected); the other three integration shards; the web/agent suites. CI covers those.

https://claude.ai/code/session_01Tpgsi9U2jPFo9CgcyLzyuV

### DIFF
--- a/apps/api/src/__tests__/integration/configPolicyInheritance.integration.test.ts
+++ b/apps/api/src/__tests__/integration/configPolicyInheritance.integration.test.ts
@@ -26,9 +26,13 @@
  *     system-context escalation anywhere.
  *  4. DELETION. A parent with children cannot be deleted alone, but an org
  *     cascade that removes the family in ONE statement still succeeds.
+ *  5. ERASURE ROW-SET CLOSURE. The org-erasure FK ledger
+ *     (orgCascadeFkOnDeleteAllowlist.ts) pins this self-FK with a reviewed
+ *     argument instead of a migration; section 5 is that argument's evidence.
  */
 import './setup';
 import { getTestDb } from './setup';
+import { randomUUID } from 'node:crypto';
 import { afterEach, describe, expect, it } from 'vitest';
 import { eq, inArray, sql } from 'drizzle-orm';
 import { db, withDbAccessContext, type DbAccessContext } from '../../db';
@@ -46,6 +50,7 @@ import {
   InvalidParentPolicyError,
   PolicyHasChildrenError,
 } from '../../services/configurationPolicy';
+import { cascadeDeleteOrg } from '../../services/tenantCascade';
 import type { AuthContext } from '../../middleware/auth';
 import { createPartner, createOrganization } from './db-utils';
 
@@ -929,5 +934,86 @@ describe('config policy inheritance — deletion (live DB)', () => {
       db.select({ id: configurationPolicies.id }).from(configurationPolicies)
         .where(inArray(configurationPolicies.id, [parent.id, child.id])));
     expect(remaining).toEqual([{ id: parent.id }]);
+  });
+});
+
+
+// ============================================================
+// 5. Erasure row-set closure -- the evidence behind the org-cascade FK ledger
+// ============================================================
+
+/**
+ * `orgCascadeFkOnDelete.integration.test.ts` classifies
+ * `configuration_policies(parent_policy_id)` as `self-ref-open-row-set`: the
+ * table's `org_id` is nullable (partner-wide policies, epic #2135), so from the
+ * catalog alone it cannot tell whether `DELETE ... WHERE org_id = $1` removes a
+ * row set closed under the self-reference. It is pinned in the ledger with the
+ * argument that the W01 constraint trigger closes it. That argument is only
+ * worth something if it is checked, so these tests check it (#5123).
+ *
+ * Closure needs exactly one thing: no row that SURVIVES the erasure statement
+ * may reference a row it deletes. The surviving rows are the partner-wide ones
+ * and other orgs' rows, so the two forges below are the COMPLETE set of edges
+ * that could break it -- both refused, in SYSTEM scope, which is the widest
+ * scope any erasure or merge path runs under. The third test then runs the real
+ * cascade rather than a hand-written DELETE, so a later change to the erasure
+ * SQL (a per-row loop, a different WHERE) is caught here as well.
+ */
+describe('config policy inheritance -- erasure row-set closure (live DB)', () => {
+  it('SYSTEM scope still refuses a partner-wide child under an org-owned parent', async () => {
+    const t = await seedTenancy();
+    const orgParent = await seedPolicy({ orgId: t.a1, name: 'A1 baseline' });
+
+    // The edge that would break closure: this child SURVIVES
+    // `DELETE ... WHERE org_id = a1` (its org_id IS NULL) while its parent does not.
+    await expectSqlState(
+      () => seedPolicy({
+        partnerId: t.p1,
+        name: 'forged partner-wide child',
+        parentPolicyId: orgParent.id,
+      }),
+      '23514',
+      'configuration_policies_parent_guard',
+    );
+  });
+
+  it("SYSTEM scope still refuses a child of ANOTHER org's parent", async () => {
+    const t = await seedTenancy();
+    const orgParent = await seedPolicy({ orgId: t.a1, name: 'A1 baseline' });
+
+    // The other surviving shape: a2's row referencing a1's row.
+    await expectSqlState(
+      () => seedPolicy({
+        orgId: t.a2,
+        name: 'forged cross-org child',
+        parentPolicyId: orgParent.id,
+      }),
+      '23514',
+      'configuration_policies_parent_guard',
+    );
+  });
+
+  it('the real org erasure removes a parent+child family and leaves the partner-wide baseline', async () => {
+    const t = await seedTenancy();
+    const baseline = await seedPolicy({ partnerId: t.p1, name: 'MSP baseline' });
+    const orgParent = await seedPolicy({ orgId: t.a1, name: 'A1 baseline' });
+    const orgChild = await seedPolicy({
+      orgId: t.a1, name: 'A1 child', parentPolicyId: orgParent.id,
+    });
+    const inheritsBaseline = await seedPolicy({
+      orgId: t.a1, name: 'A1 child of the MSP baseline', parentPolicyId: baseline.id,
+    });
+
+    // cascadeDeleteOrg, not a hand-written DELETE: this is the statement the
+    // ledger entry's argument is actually about.
+    await expect(cascadeDeleteOrg(t.a1, randomUUID(), 'config-policy-erasure@test.invalid'))
+      .resolves.toBeDefined();
+
+    const remaining = await withDbAccessContext(SYSTEM_CTX, () =>
+      db.select({ id: configurationPolicies.id }).from(configurationPolicies)
+        .where(inArray(configurationPolicies.id, [
+          baseline.id, orgParent.id, orgChild.id, inheritsBaseline.id,
+        ])));
+    expect(remaining).toEqual([{ id: baseline.id }]);
   });
 });
--- a/apps/api/src/__tests__/integration/configPolicyInheritance.integration.test.ts
+++ b/apps/api/src/__tests__/integration/configPolicyInheritance.integration.test.ts
@@ -28,7 +28,10 @@
  *     cascade that removes the family in ONE statement still succeeds.
  *  5. ERASURE ROW-SET CLOSURE. The org-erasure FK ledger
  *     (orgCascadeFkOnDeleteAllowlist.ts) pins this self-FK with a reviewed
- *     argument instead of a migration; section 5 is that argument's evidence.
+ *     argument instead of a migration. The "erasure row-set closure" describe
+ *     block below is that argument's evidence, plus the two service-shape
+ *     properties it leans on (single-statement org erasure; the partner purge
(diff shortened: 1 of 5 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — code-reviewer, pr-test-analyzer, comment-analyzer (parallel, read-only against `gh pr diff`).

**Findings:** 4 raised → all 4 addressed in `4cf6d76d3c`; 0 outstanding.

- **code-reviewer: 0 defects.** Independently re-derived the closure argument from `breeze_config_policy_parent_guard()` and confirmed it holds; checked the hazards that would break it — `session_replication_role` bypass in org merge (absent), the `organizations` partner re-point (has its own guard), and SECURITY DEFINER / RLS-invisibility (fails closed, and ownership moves are system-scoped where it is moot). Also confirmed the scope-election fix covers **every** owner-move call site in the three touched files (2 + 1 + 4), that ledger sort/`reason`/`allColumnsNullable`/note-length all satisfy the contract test, and that `parent_policy_id` was already classified in `CORE_TENANT_EXPORT_POLICY` by #5099 — so no cascade-registration list is left dangling.
- **pr-test-analyzer — important (fixed):** `cascadeDeletePartner` runs `cascadeDeleteOrg` per child org and only *then* sweeps `partner_id`-owned rows, so an org-owned child is removed before the partner-wide baseline it inherits from. That ordering is load-bearing for this exact FK — reverse it and the sweep aborts mid-purge with 23503, the #4100 failure mode on the partner axis — and `tenantCascadePartner.integration.test.ts` seeds no `configuration_policies` at all. Added a test that seeds a baseline plus an org-owned child *and* a partner-wide child (so the sweep's own statement is exercised for closure too) and runs the real purge.
- **pr-test-analyzer — minor (fixed):** the two cascade tests seed only *legal* edges, so they would pass with the guard gone; the section docstring was calling them closure evidence. It now names the two forge tests as the proof and the cascade tests as the service-shape properties the argument leans on. The ledger note was reworded to match.
- **pr-test-analyzer — minor (fixed):** the new 23514 assertions used `captureSqlState`, which reads only the SQLSTATE, so a future CHECK on the same statement path would have satisfied them. Added `captureSqlFailure` and pinned `configuration_policies_owner_immutable` by name.
- **comment-analyzer — 0 critical, 1 rot risk (fixed):** every factual claim in the added prose verified against the migration, the design spec's Deletion section, `tenantCascade.ts` and `configurationPolicy.ts`. The one issue was the ledger note pointing at "section 5", a hand-maintained ordinal nothing enforces; it now names the `describe` block.

**Tests** (foreground, live Postgres 16 via `pnpm test-stack`, re-run after the review fixes):

- `configPolicyInheritance`, `partnerApiRls`, `orgCascadeFkOnDelete` — 3 files, **64 passed**, all 4 erasure-closure tests confirmed by name in verbose output.
- Earlier full pass on the base commit: `tenantCascade*` + `orgLifecycleFoundations` + all `configPolicy*` (13 files, 158 passed); `autoMigrate` + `migrationRlsScope` + `migrationGucAttributes` (1488 passed); `pnpm db:check-drift` clean.

**Status:** review-clean, awaiting maintainer merge. This unblocks every PR currently red on integration shards 1/2/4.

Issue #5117, opened 2026-09-06
mobile: decisions from iOS round-2 review — Systems 'healthy' definition, Device Details v1 fields, Home empty state, header status dot


ToddHebebrand:
Product/UX decisions surfaced by the 2026-09-06 OliveTech iOS round-2 review. None is a code defect; each needs a call before someone builds it.

1. **What does "healthy" mean on the Systems tab?** Org rows say "7 devices, healthy" and the org filter says "all clear", while the AI's `get_fleet_findings` on the same fleet reports VSS failures, Universal Print errors, Intel ME verification errors and a SCEP enrollment failure. Systems counts active alerts only; findings live elsewhere. Options: (A) fold open findings into the Systems issue count; (B) keep alerts-only but rename the copy ("no active alerts"); (C) add a separate "Findings" line under each org. **Recommend A** — a technician opening Systems expects the same picture the AI gives them.
2. **Device Details content.** Today: status, hostname, OS, last seen, three metric tiles, Reboot/Shutdown/Wake. Missing for a phone-first tech: org/site (fixed in #5104), LAN/WAN IP, logged-in user, OS version, open alerts, open tickets, and entry points for Run script / Remote. Decide the v1 set; recommend IP + logged-in user + OS version + open alerts count + "Open tickets" link.
3. **Home empty state.** A new conversation is a black void above three prompt chips. Options: recent conversations list, a compact fleet strip (online/offline/issues), or nothing by design. Recommend the fleet strip, since Systems already owns the detail.
4. **Header status dot.** The "•" next to "+" on Home has no label or tooltip (it is the StatusPill collapsed). Decide whether it needs a label at rest or only when degraded.

ToddHebebrand:
Decisions (Todd, 2026-09-07):
1. **A** — fold open findings into the Systems issue count.
2. Device Details v1 = IP, logged-in user, OS version, open-alerts count, Open tickets link (as recommended).
3. Home empty state = compact fleet strip.
4. Header status dot: fine as is, no change.

Implementation issues follow; this issue stays as the decision record.

ToddHebebrand:
All three decisions implemented and merged: #5143 (decision 1, findings in Systems count), #5144 (decision 2, Device Details v1 fields), #5142 (decision 3, Home fleet strip). Decision 4 = no change. Unreleased until the next release + TestFlight build.

ToddHebebrand:
[...]

Issue #2503, opened 2026-07-14
[UI] Add optional WAN IP and LAN IP columns to the device list view


ToddHebebrand:
## Description

Community request (Discord, multiple users): add **WAN IP** and **LAN IP** as optional columns in the device list view, so techs can see/scan addresses across a fleet without opening each device.

Filed separately from #2502 (device change history) as it came up in the same thread.

## What already exists (do not rebuild)

- Per-device WAN/LAN IP **history** already exists as the `ip-history` device-detail tab (`DeviceIpHistoryTab.tsx`, `aiToolSchemas.ipHistory`). Current WAN/LAN IP values are already collected and stored.
- This request is **not** about history — it's about surfacing current WAN IP and LAN IP as **columns** in the device list/table.

## Scope

- Add WAN IP and LAN IP as **optional, toggleable columns** in the device list view (consistent with how other optional columns are shown/hidden).
- Source current values from existing device inventory fields — no new agent collection expected (confirm the fields are already on the `devices` row / list API payload; extend the list endpoint's selected columns if not).
- Ensure they respect existing sort/filter behavior where practical.

## Affected files (to confirm during implementation)

- `apps/web/src/components/devices/*` (device list/table column definitions + column toggle)
- `apps/api/src/routes/devices.ts` (device list endpoint — include WAN/LAN IP in payload if not already present)

## Reported By

Community request in Discord — multiple users, incl. SemoTech, in the same thread as #2502.

PR #5144, opened 2026-09-07
feat(mobile,api): Device Details v1 fields — IP, logged-in user, OS version, open alerts/tickets (#5140)

### LINKED ISSUE #5140: mobile/api: Device Details v1 fields — IP, logged-in user, OS version, open alerts, open tickets (#5117 decision 2)
Decision 2 of #5117 (Todd, 2026-09-07): Device Details v1 field set on mobile = **IP address, logged-in user, OS version, open-alerts count, Open tickets link** (org and site already landed in #5113).

Today `apps/mobile/src/screens/devices/DeviceDetailScreen.tsx` shows status, hostname, OS, last seen, three metric tiles and Reboot/Shutdown/Wake. Sources: `devices.os_version`, `devices.last_user` (`apps/api/src/db/schema/devices.ts`), LAN IP from the device network rows (the web device list already ranks the "best" LAN address — reuse that logic, see `routes/devices/core.ts` `lanIpByDevice`), public IP where recorded, open alerts by device, open tickets by device.

**Wanted**
- API: the mobile device detail/list response (`apps/api/src/routes/mobile.ts`, whatever `getDevice()` in `apps/mobile/src/services/api.ts:1121` calls) carries `osVersion`, `lastUser`, `lanIp`, `publicIp`, `openAlertCount`, `openTicketCount` — batched, no N+1; RLS unchanged.
- Mobile: `mapDevice` maps them; Device Details renders rows for IP (LAN, and public if present), logged-in user, OS version (next to OS), an "Open alerts · N" row that navigates to the alerts list filtered to the device if such a screen exists (else plain count), and an "Open tickets · N" row linking to the Tickets tab filtered to the device (else plain count). Empty values render "—", never blank.

ToddHebebrand:
## Summary

Implements decision #5117-2 (Device Details v1 field set): `GET /mobile/devices` now carries `osVersion`, `lastUser`, `lanIp`, `publicIp`, `openAlertCount`, and `openTicketCount`, and `DeviceDetailScreen` renders them.

**API (`apps/api/src/routes/mobile.ts`, `GET /devices`)**
- `osVersion` / `lastUser` / `publicIp` (aliased from `devices.lastSeenIp`) are plain columns — no extra query.
- `lanIp`: batched per-page lookup. Rather than a second SQL `LATERAL` (like `devices/core.ts`'s #2503 list column), this fetches all matching `device_network` rows for the page in one plain `SELECT ... WHERE device_id IN (...)` and ranks them in application code (same tiers: `is_primary` → IPv4 before IPv6 → routable before APIPA/link-local/loopback → interface name tiebreak). Keeps the route free of a raw `sql\`\`` LATERAL and the mocking that would require.
- `openAlertCount` / `openTicketCount`: two batched `GROUP BY` count queries. "Open" = `active`/`acknowledged` for alerts (mirrors `/summary`'s bucket), `new`/`open`/`pending`/`on_hold` for tickets (same set as `ticketReadModel.ts`'s `OPEN_TICKET_STATUSES` / `ticketService.ts`'s `ADDIN_OPEN_STATUSES`, both module-private so kept local here).
- All three new lookups are **separate queries from the main device-row select — never a LEFT JOIN on it** — so a device with many network interfaces/alerts/tickets can't fan the page out. Verified by a dedicated test.
- RLS/tenant scoping unchanged: everything runs inside the same request-scoped `withDbAccessContext` as the rest of the route.

**Design choice — enriched the list, not a new detail endpoint.** `getDevice()` in `apps/mobile/src/services/api.ts` calls `GET /devices/:id` on `devices/core.ts` (outside this PR's ownership), but no mobile screen actually calls `getDevice()` today — `DeviceDetailScreen` renders entirely from the `device` object handed to it via navigation params from the list fetch. So the fields were added to the list endpoint (batched once per page, ~100 devices), which is what the screen already depends on, rather than adding a new mobile-only detail route or touching the core web endpoint. Flagged for visibility — happy to add a lean detail endpoint instead if preferred.

**Mobile (`apps/mobile/src/services/api.ts`, `DeviceDetailScreen.tsx`)**
- `MobileDeviceRecord` / `Device` / `mapDevice()` carry the six new optional fields (`?? undefined`, not `||`, so a real `0` open-count survives).
- New pure helper `apps/mobile/src/screens/devices/deviceDetailFields.ts` (pattern: `lib/osLabel.ts`) — `formatDetailValue`, `formatIpValue`, `formatOsVersionValue`, `formatCountLabel` — keeps the "—" fallback and two-line IP rules out of JSX, node-testable without React Native.
- Rows: IP (LAN, public on a second line when present), Logged-in user, and OS version (appended to the existing OS row) always render with "—" for absent values, per the decision doc — unlike the pre-existing rows, which hide entirely when unset. "Open alerts · N" / "Open tickets · N" render as plain text (no navigation): no alerts-list or device-filtered tickets-list screen exists on mobile yet to link to.

## Test plan
- [x] `cd apps/api && npx vitest run src/routes/mobile.test.ts` — 80 passed (2 pre-existing skips), including two new tests: no-fields-present defaults to null/0, and a fan-out guard (2 network rows + count>1 alerts/tickets still yield exactly 1 response row).
- [x] `cd apps/api && NODE_OPTIONS=--max-old-space-size=12288 npx tsc --noEmit` — clean.
- [x] `cd apps/mobile && npx vitest run` — 1314 passed / 3 pre-existing skips (full suite), including new `deviceDetailFields.test.ts` (15 cases) and extended `api.deviceMapping.test.ts`.
- [x] `cd apps/mobile && npx tsc --noEmit` — clean.
- [x] `npx eslint` on all changed API files — clean. (No `eslint.config.*` exists for `apps/mobile`, so it isn't linted this way.)

Closes #5140

https://claude.ai/code/session_01PvApTe8M34RiVxBafdJGBm

### DIFF
--- a/apps/api/src/routes/mobile.test.ts
+++ b/apps/api/src/routes/mobile.test.ts
@@ -147,6 +147,19 @@ const mockSelectWhereChain = (result: unknown) => ({
   })
 });
 
+// GET /mobile/devices' batched open-alert / open-ticket counts (#5140): a
+// plain SELECT ... WHERE ... GROUP BY, one row per device that has at least
+// one matching row (never per-alert/per-ticket) — this is what keeps the
+// device-list page from fanning out even though a device can have many
+// alerts/tickets.
+const mockSelectGroupByChain = (result: unknown) => ({
+  from: vi.fn().mockReturnValue({
+    where: vi.fn().mockReturnValue({
+      groupBy: vi.fn().mockResolvedValue(result)
+    })
+  })
+});
+
 const mockSelectOrderChain = (result: unknown) => ({
   from: vi.fn().mockReturnValue({
     where: vi.fn().mockReturnValue({
@@ -1379,7 +1392,11 @@ describe('mobile routes', () => {
               organizationName: 'Acme Corp'
             }
           ]) as any
-        );
+        )
+        // Device Details v1 fields (#5140): LAN IP + open alert/ticket counts.
+        .mockReturnValueOnce(mockSelectWhereChain([]) as any)
+        .mockReturnValueOnce(mockSelectGroupByChain([]) as any)
+        .mockReturnValueOnce(mockSelectGroupByChain([]) as any);
 
       const res = await app.request('/mobile/devices?status=online&search=host', {
         method: 'GET'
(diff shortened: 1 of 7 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, pr-test-analyzer, silent-failure-hunter) in parallel.

**Findings:** 3 raised, all addressed in ca3f622ac2:
- silent-failure-hunter (critical): `loadDeviceDetailsV1Fields`'s 3 batched queries had no error handling — a transient failure on `device_network`/`alerts`/`tickets` would 500 the whole `GET /mobile/devices` list even though the core device-row query already succeeded. Fixed: each lookup is now independently try/catch'd; the count lookups surface `null` (never a false `0`) on failure so a device with real open alerts can't render "Open alerts · 0" from a query blip.
- pr-test-analyzer (important): the LAN-IP ranking comparator's tiers 2–4 (IPv4-before-IPv6, routable-before-unroutable, interface-name tiebreak) and the alert/ticket "open" status-filter predicates were never actually exercised — only the `is_primary` tier had a test, and the status filters were asserted with a deep-search helper that also would have false-matched on the enum column's full declared value set. Added 3 ranking-tier tests, a multi-device attribution test, and a status-filter test built on a bound-`Param` extractor instead of the vacuous deep search.
- code-reviewer: no high-confidence bugs (confirmed RLS/tenancy scoping, LAN-IP ranking parity with `devices/core.ts`'s #2503 lateral, and `??`-vs-`||` null/zero handling are all correct); flagged one test asserting an impossible DB state (`devices.os_version` is `NOT NULL`) — fixed.

**Tests:** `cd apps/api && npx vitest run src/routes/mobile.test.ts` — 86 passed, 2 pre-existing skips. `cd apps/mobile && npx vitest run` (full suite) — 1315 passed, 3 pre-existing skips. `npx tsc --noEmit` clean in both `apps/api` (with `NODE_OPTIONS=--max-old-space-size=12288`) and `apps/mobile`. `npx eslint` clean on the changed API files (no `eslint.config.*` exists for `apps/mobile`).

Issue #5131, opened 2026-09-07
Offline work queue — queue device actions while offline, deliver on reconnect (#5128)


ToddHebebrand:
Tracking issue for **Offline work queue — queue device actions while offline, deliver on reconnect (#5128)** (`offline-work-queue`).

- Plan: `docs/superpowers/plans/misc/2026-09-06-offline-work-queue.md`
- Spec: `docs/superpowers/specs/misc/2026-09-06-offline-work-queue-design.md`

Waves are tracked as sub-issues. Managed by feature-lifecycle-mcp.

ToddHebebrand:
Shipped on main (unreleased) via W01 #5145 (e8db622c9d), W02 #5164 (a867f6a078), W03 #5174 (e087ecc6dd), W04 #5243 (6f09297d21), W05 #5248 (98d8c9afdb). DEVICE_COMMAND_OFFLINE_QUEUE_ENABLED defaults ON; opt out with =false. Release-note paragraph is in the #5243 PR body.

Outstanding, tracked separately:
- W06 #5137 (requester re-authorisation at claim) — blocked on #3985; stays open as its own wave issue.
- #5249 (W05b) — remove the flag and W01 compatibility aliases one release after the first release containing W01–W04.

Not verified in production: real-fleet behaviour of the two-clock reaper and patch-job supersession under load; integration suites passed on CI's real Postgres.

Issue #5128, opened 2026-09-07
Software updates with offline machines.Software updates with offline machines.


ToddHebebrand:
Feature request from **stressedout9064** in #Software updates with offline machines.Software updates with offline machines. (Discord, message 1546276704371089490).

**Suggestion (PII-redacted):**

> What am I trying to do?
> Fix things when I discover them regardless of the user being currently available. Every action in breeze should not depend on the machine being online, it should schedule and wait until the machine is online then execute then.
> 
> Software updates, patches, scripts, software deployment should not just fail when you click on them, they should switch to pending then try and do it once the machine is online.
> 
> What  you do today instead?
> Use our current RMM that we would like to remove and replace with BreezeRMM
> 
> Who else would want this?
> Every technician

_Filed automatically by @breeze-discord triage. Verify and add labels/scope as needed._

ToddHebebrand:
**Spec drafted:** [`docs/superpowers/specs/misc/2026-09-06-offline-work-queue-design.md`](https://github.com/LanternOps/breeze/blob/bb4f944b405ef61d1a76a9ff307a0c5835573a32/docs/superpowers/specs/misc/2026-09-06-offline-work-queue-design.md) (branch `offline-work-queue`, commit bb4f944b40). Advisor quorum run (Codex xhigh); both picked the same approach.

**Diagnosis.** Offline handling is six different behaviours depending on the code path: `queueCommandForExecution`/`executeCommand` hard-reject (patches, backups, most AI tools); manual scripts queue but the stale-command reaper reuses the *execution* timeout as the *delivery* deadline, so a queued script dies after ~10 min; software installs queue with a 7-day special case but hand the agent a one-hour presigned URL; the generic device-command routes bypass every helper. Delivery to a reconnecting agent already works via the heartbeat claim. Related: #4981 (same reporter), #4226 item 7.

**Design.** Make deferred delivery a first-class property of `device_commands`: a nullable `deliver_by` column, one enqueue seam with an explicit `offlinePolicy` (`reject` | `queue`) backed by a fail-closed per-type registry, a reaper that separates the delivery clock from the execution clock, claim-time eligibility checks (org move, lifecycle, partner trust, requester active), late-binding payload preparation at claim (re-mint URLs), a power-state barrier so a queued reboot cannot land mid-script, and per-feature `queued` states (patch jobs stay non-terminal with a `devicesQueued` counter and one idempotent finalizer). No agent change. Backups stay hard-gated in v1.

[...]

PR #3985, opened 2026-08-25
fix: harden authorization and fleet execution integrity

ToddHebebrand:
## Summary

This draft hardens the highest-priority authorization and fleet-integrity paths identified by the S0 revalidation.

### Authorization and queued recovery

- bind automation resources to explicit tenant ownership and enforce that ownership at execution time
- add explicit cross-site restore permission and lineage validation
- persist durable authorization subjects for queued recovery jobs
- reauthorize media, boot, disaster-recovery, C2C, and scheduled work when workers execute it
- add real-PostgreSQL proofs for stale or revoked queued authority

### Fleet execution truth

- replace unbounded device selectors with authorized paged options and race-safe hydration
- make heartbeat the authoritative reachability signal
- persist immutable versioned agent-health observations and ingest health independently of heartbeat success
- return explicit per-target script admission results across API, mobile, remediation, and web consumers
- persist per-action automation dispatch and terminal evidence with guarded state transitions and serialized reconciliation
- preserve exact command, script, and software result correlations through HTTP, WebSocket, cancellation, and reaper paths
- retain versioned software-inventory evidence and protect the last-known-good projection from empty, partial, failed, truncated, collapsed, or out-of-order reports
- emit typed software-inventory completeness evidence from macOS, Linux, and Windows collectors
- require exact accepted inventory evidence before resolving vulnerabilities by absence, and clear lineage on every reopen path

## Consolidated verification

- shared targeted: 12 passed
- API targeted: 701 passed, 5 existing skips
- web targeted: 222 passed
- real PostgreSQL Track B matrix: 35 passed across 10 files
- RLS coverage: 80 passed
- migration naming: clean
- migration ledger drift: 567/567 clean
- API TypeScript: clean
- web Astro diagnostics: 1,683 files, 0 errors
- Go collector race, focused heartbeat v2 race, and uninstall/software compatibility race: clean
- Linux and Windows collector cross-builds: clean
- branch diff check: clean

## Draft / do not merge yet

- One independent whole-branch security review is still required.
- Native Linux and Windows collector behavior plus real Windows registry evidence must be captured from CI/candidate artifacts.
- The full macOS heartbeat race package hits an unchanged environment-specific hang in `TestConsentGate_NoHelper_Proceed` inside native screen capture; Task 13 focused heartbeat coverage passes.
- The shared package typecheck still reports three unchanged baseline errors in `packages/shared/src/validators/quotes.test.ts`.
[...]

PR #5145, opened 2026-09-07
feat(commands): offline work queue W01 — deliver_by, policy seam, two-clock reaper, claim eligibility, cancel (#5131)

### LINKED ISSUE #5132: W01: W01 — Core: deliver_by, offline-policy seam, two-clock reaper, claim eligibility, cancel
Wave W01 of #5131 — W01 — Core: deliver_by, offline-policy seam, two-clock reaper, claim eligibility, cancel
Branch: `feature/5131-offline-work-queue/wave-5132`

ToddHebebrand:
Closes #5132

Wave 1 of the offline work queue (feature #5131, request #5128). Deferred delivery becomes a first-class property of `device_commands`: one deadline column, one enqueue seam with an explicit offline policy, a reaper that separates the *delivery* clock from the *execution* clock, and claim-time eligibility re-checks. **No agent change** — delivery still rides the agent's existing heartbeat claim.

The spec and plan land with this PR (they lived on the `offline-work-queue` branch until now).

## Flag state: OFF

`DEVICE_COMMAND_OFFLINE_QUEUE_ENABLED` defaults to `false`. It gates **only** the callers that hard-rejected offline devices before #5128 (everything reaching `queueCommandForExecution`: patch executor, automations, scan/rollback, AI tools). Scripts, software installs and the generic device-command routes already queued for offline devices and are **not** gated — they simply gain a real deadline. So with the flag off, the user-visible behaviour change in this PR is: *queued work stops being reaped ten minutes later*.

Flipped to `true` in W4 (Task 4.3), removed in W5.

## The two latent defects this fixes

1. **A WS-dispatched `software_install` created no `device_commands` row at all.** `dispatchSoftwareInstallToDevice` pushed over the socket first and returned `{ transport: 'ws', deviceCommandId: null }` on success. With no row, a push the agent never acted on was invisible to the reaper, to the device's queued list and to cancel — and `deployment_results.device_command_id` stayed NULL, so the result reaper could not tell "delivered" from "never sent". The row is now persisted **before** either transport, by the one seam, and the push carries that row's UUID.

2. **The software result reaper measured its "delivered but silent" tier from the deployment's `dispatched_at`.** An install delivered three days after dispatch (device was offline in between) was timed out on the reaper's very next pass. It now measures from the command's `executed_at` — the instant the agent actually claimed it — and no longer touches undelivered rows at all.

## What changed

**Schema + migration**
- `apps/api/migrations/2026-10-13-100000-device-commands-deliver-by.sql` — `deliver_by timestamptz`, `submitted_org_id uuid REFERENCES organizations(id) ON DELETE SET NULL`, one partial index. DDL only (no `breeze.scope` needed), idempotent, no inner transaction.
- `db/schema/devices.ts` — the two columns, with the "provenance, not tenancy" rationale inline.
- `db/schema/deviceCommands.columns.test.ts` *(new)* — pins the shape, and asserts the table still has **no** `org_id` column.

`device_commands` stays intentionally system-scoped: no RLS policy, no `CORE_ORG_CASCADE_DELETE_ORDER` entry, no `CORE_TENANT_EXPORT_POLICY` entry. `submitted_org_id` is deliberately not named `org_id` so the contract tests' auto-discovery does not reclassify the table. Verified — see "Contract suites" below.

**New modules**
- `services/commandOfflinePolicy.ts` (+ test) — the fail-closed per-type registry. TTL classes: standard 168 h, short 24 h, power_state 24 h, live = a 5-minute race grace; all env-tunable. An unregistered type throws before any DB access.
- `services/dispatchDeviceCommand.ts` (+ test) — the single enqueue seam. Order: resolve policy → device lookup → `expectedOrgId` → lifecycle → partner trust → reject-or-queue → **persist** → claim/refresh/push/release. Returns `delivery: 'delivered' | 'queued_offline' | 'queued_live'` and `deliverBy`.
- `services/commandClaimEligibility.ts` (+ test) — claim-time filter (org drift, lifecycle, partner trust, requester still active) plus the power-state barrier. Cancels are written **inside the caller's claim transaction**, so a row it rejects can never be delivered by a concurrent claim.
- `services/commandTypes.ts` *(new, extracted)* — see "Notable decisions".

**Modified**
- `services/commandQueue.ts` — `queueCommand` takes `deliverBy` / `submittedOrgId`; `queueCommandForExecution` becomes a thin adapter over the seam (error strings unchanged); `executeCommand`'s insert stamps the short race-grace deadline.
- `services/commandDispatch.ts` — both claim paths gain the `deliver_by IS NULL OR deliver_by > now()` predicate; the batch claim runs eligibility + the barrier.
- `jobs/staleCommandReaper.ts` — two clocks, CAS on the observed `(status, executed_at)`, the new `expired / not_delivered_before_deadline` result, `propagateTimedOutDeviceCommand` gains `kind` and now propagates to `script_executions` and `deployment_results`; new `propagateCancelledDeviceCommand`; `SOFTWARE_QUEUED_EXPIRY_MS` and the whole old Tier 2 removed.
- `services/commandTimeouts.ts` — the `software_install → SEVEN_DAYS` special case is gone (`deliver_by` owns the delivery deadline now); `SOFTWARE_INSTALL` moves to `LONG_TIMEOUT_TYPES` as a pure execution budget.
- `services/commandDelivery.ts` — `prepareClaimedCommandsForDelivery` (alias `decryptClaimedCommandsForDelivery` kept for W5) runs per-type `deliveryRefreshers` before decrypt; a refresher failure **releases** the row instead of delivering a stale payload.
- `services/softwareDeployment.ts` — persist-before-push through the seam; payload gains `s3Key` (only when the URL really is the presigned one).
- `services/softwareDeploymentResult.ts` — `reconcileSoftwareInstallResult`, shared by both result transports.
- `services/scriptDispatch.ts` — `offlinePolicy` (with `requireOnline` kept as a deprecated alias); stamps `deliver_by` + `submitted_org_id`; result gains `deliverBy`.
- `routes/agentWs.ts`, `routes/agents/commands.ts`, `routes/agents/heartbeat.ts` — shared software reconciliation on the generic result path; call sites renamed to `prepareClaimedCommandsForDelivery`.
- `routes/devices/commands.ts` — bulk / single / `set_auto_update` go through the seam; bulk response gains `queuedOffline: string[]`, single gains `delivery` + `deliverBy`; `GET /:id/commands?status=` filter; new `POST /:id/commands/:commandId/cancel`.
- `routes/devices/moveOrg.ts`, `routes/devices/core.ts` — cancel the device's pending rows in the same transaction as the org flip / decommission (decommission excludes `self_uninstall`, so the uninstall drain still delivers).

## Notable decisions worth a reviewer's eye

- **`CommandTypes` was extracted into a new leaf module `services/commandTypes.ts`** (re-exported from `commandQueue`, so every existing import still works). The registry builds from that table at module load, and `commandQueue` imports the seam, which imports the registry — leaving the table in `commandQueue.ts` made that a genuine ESM initialisation cycle whose outcome depended on which module the process imported first. It reproduced as a hard failure in test.
- **Backup / restore / verification types are classified `live` (reject).** The spec puts them out of scope for v1; classifying them here is what keeps them rejecting once the flag flips in W4 — the flag alone would have started queueing them.
- **The registry was swept against every `device_commands.type` literal the API writes**, not just `CommandTypes`: `reboot`, `shutdown`, `schedule_reboot`, `update`, `set_auto_update`, `wake`, `network_discovery`, `update_agent`, `update_watchdog`, `restart_agent`, `actuate_elevation`, `desktop_stream_stop`, `apply_browser_policy`. A test pins that inventory, because a missing entry throws at enqueue.
- **The `software_install` delivery refresher is registered in `commandDelivery.ts` itself**, not by side effect from `softwareDeployment.ts`. Side-effect registration would silently deliver stale installer URLs in any process that happened not to import the feature module — exactly the failure class this seam exists to close. The registry stays open for W3 to add to.
- **Order change in the seam:** partner trust is now evaluated *before* the offline rejection (as §D specifies), so a trust-denied offline device returns `trust_denied` rather than `device_offline`. Error strings for every other path are byte-identical to before.
- **Known, deliberate:** on an **org merge**, a device's queued rows are cancelled at claim as `device_moved_org` (the loser org survives as a shell, so `submitted_org_id` mismatches the repointed device). Fail-closed and recoverable — the operator re-issues. Flagging it rather than widening `orgMergeRegistry` in this wave.

## Verification

| Check | Result |
|---|---|
| `npx tsc --noEmit -p .` (apps/api) | clean |
| `pnpm lint` | clean (6/6 packages) |
| Unit suites for every touched module | green |
| `bash scripts/check-migration-naming.sh` | OK |
| `autoMigrate.test.ts` + `migrationRlsScope.test.ts` | green (DDL-only migration, no scope elevation needed) |

**Contract suites — run against a real local Postgres, not just asserted:**

| Suite | Result |
|---|---|
| `rls-coverage.integration.test.ts` | 102 passed — table stays system-scoped |
| `tenantCascade.integration.test.ts` | green |
| `orgLifecycleFoundations.integration.test.ts` | green |
| `tenant-export-policy` + `tenantExportErasureRoundtrip` | green — no new classification needed |

**Integration test: RAN against a real Postgres 
(`docker-compose.test.yml`), 19/19 passing.** `src/__tests__/integration/deviceCommandOfflineQueue.integration.test.ts` covers: queue-with-deadline against an offline device; a `reject` enqueue creating no row; a queued row surviving past its execution timeout; expiry exactly at `deliver_by` with `expired / not_delivered_before_deadline`; the linked `script_executions` row failing **only** by propagation (and its own reaper leaving a still-waiting one alone); heartbeat claim delivering exactly once; a past-deadline row not being claimed; org move cancelling at claim; the power-state barrier holding a reboot until the queue drains; the cancel endpoint (200 then 409); a cancelled row never re-claimed; legacy `deliver_by IS NULL` / `submitted_org_id IS NULL` rows behaving exactly as before; decommission; the unregistered-type refusal; an inactive requester (with a positive control on `created_by`); and pg_catalog assertions that the partial index exists and `org_id` still does not.

## Not in this wave

Scripts/generic-command UX (W2), patches (W3), automations (W4), AI-tool text + docs sweep + flag removal (W5), principal rehydration at claim (W6, after #3985).

### DIFF
--- a/apps/api/migrations/2026-10-13-100000-device-commands-deliver-by.sql
+++ b/apps/api/migrations/2026-10-13-100000-device-commands-deliver-by.sql
@@ -4,7 +4,9 @@
 -- claimed the row (pending -> sent). It is separate from the execution timeout
 -- (services/commandTimeouts.ts), which the stale reaper applies to `sent` rows
 -- from executed_at. NULL keeps today's rule for rows created before this
--- migration, so no backfill is needed and old pending rows behave as before.
+-- migration, so no backfill is needed and old pending rows behave as before
+-- (software_install's retired 7-day special case is preserved for such legacy
+-- rows inside the reaper, not here).
 --
 -- submitted_org_id is PROVENANCE, NOT TENANCY. device_commands is intentionally
 -- system-scoped (agent WS path, no RLS -- see CLAUDE.md). This column records the
--- a/apps/api/migrations/2026-10-13-100000-device-commands-deliver-by.sql
+++ b/apps/api/migrations/2026-10-13-100000-device-commands-deliver-by.sql
@@ -1,0 +1,41 @@
+-- 2026-10-13-100000: deferred delivery for device commands (#5128).
+--
+-- deliver_by is the DELIVERY deadline: the instant by which an agent must have
+-- claimed the row (pending -> sent). It is separate from the execution timeout
+-- (services/commandTimeouts.ts), which the stale reaper applies to `sent` rows
+-- from executed_at. NULL keeps today's rule for rows created before this
+-- migration, so no backfill is needed and old pending rows behave as before.
+--
+-- submitted_org_id is PROVENANCE, NOT TENANCY. device_commands is intentionally
+-- system-scoped (agent WS path, no RLS -- see CLAUDE.md). This column records the
+-- device's org at enqueue so claim-time eligibility can cancel rows whose device
+-- has since moved org. It is deliberately not named org_id: the RLS-coverage and
+-- cascade contract tests auto-discover `org_id` columns, and this table must
+-- not be reclassified as tenant-scoped. ON DELETE SET NULL so erasing an org a
+-- device has LEFT is never blocked by that device's historical rows.
+--
+-- DDL only: no DML, so no set_config('breeze.scope', 'system', true) is needed
+-- (migrationRlsScope.test.ts).
+
+ALTER TABLE device_commands ADD COLUMN IF NOT EXISTS deliver_by timestamptz;
+
+ALTER TABLE device_commands ADD COLUMN IF NOT EXISTS submitted_org_id uuid;
+
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'device_commands_submitted_org_id_fkey'
+      AND conrelid = 'device_commands'::regclass
+  ) THEN
+    ALTER TABLE device_commands
+      ADD CONSTRAINT device_commands_submitted_org_id_fkey
+      FOREIGN KEY (submitted_org_id) REFERENCES organizations(id) ON DELETE SET NULL;
+  END IF;
+END $$;
+
+-- Reaper scan for due deliveries; partial so 7-day rows are not rescanned
+-- every 2 minutes and so the index stays small relative to the table.
+CREATE INDEX IF NOT EXISTS idx_device_commands_deliver_by
+  ON device_commands (deliver_by)
+  WHERE status = 'pending' AND deliver_by IS NOT NULL;
--- a/apps/api/src/__tests__/devices.endpoints.test.ts
+++ b/apps/api/src/__tests__/devices.endpoints.test.ts
@@ -11,6 +11,27 @@ vi.mock('../services/commandQueue', () => ({
   queueCommand: (...args: unknown[]) => mockQueueCommand(...args)
 }));
 
+// #5128 — POST /devices/bulk/commands now goes through the one enqueue seam
+// (services/dispatchDeviceCommand.ts) instead of a raw db.insert. That real
+// implementation pulls in a long transitive chain (agentWs, commandQueue,
+// commandOfflinePolicy, partnerTrust.commands) this suite's simple db mock
+// can't reasonably emulate — see routes/devices/commands.test.ts for the same
+// seam-level mock and rationale. Default: delivered to an online device;
+// decommissioned/offline outcomes are driven by the device row this suite
+// already mocks BEFORE the route ever reaches dispatch.
+const mockDispatchDeviceCommand = vi.fn(
+  async ({ deviceId, type }: { deviceId: string; type: string }) => ({
+    ok: true,
+    command: { id: 'cmd-1', deviceId, type, status: 'pending', createdAt: new Date() },
+    delivery: 'delivered',
+    deliverBy: new Date(Date.now() + 7 * 24 * 3600 * 1000),
+  }),
+);
+
+vi.mock('../services/dispatchDeviceCommand', () => ({
+  dispatchDeviceCommand: (...args: unknown[]) => (mockDispatchDeviceCommand as any)(...args),
+}));
+
 vi.mock('../services/auditEvents', () => ({
   requestLikeFromSnapshot: vi.fn(() => ({ req: { header: () => undefined } })),
   writeAuditEvent: vi.fn(),
@@ -196,6 +217,16 @@ describe('device endpoints (authenticated)', () => {
       }))
     }) as any);
     mockQueueCommand.mockReset();
+    // vi.resetAllMocks() above wipes the default implementation too —
+    // restore it explicitly, same as the db.select/insert/update mocks.
+    mockDispatchDeviceCommand.mockReset().mockImplementation(
+      async ({ deviceId, type }: { deviceId: string; type: string }) => ({
+        ok: true,
+        command: { id: 'cmd-1', deviceId, type, status: 'pending', createdAt: new Date() },
+        delivery: 'delivered',
+        deliverBy: new Date(Date.now() + 7 * 24 * 3600 * 1000),
+      }),
+    );
     app = new Hono();
     app.route('/devices', deviceRoutes);
   });
@@ -211,31 +242,10 @@ describe('device endpoints (authenticated)', () => {
       mockDeviceLookup(deviceTwo);
       mockDeviceLookup(deviceThree);
 
-      // The bulk command handler uses db.insert().values().returning() for each device
-      vi.mocked(db.insert)
-        .mockReturnValueOnce({
-          values: vi.fn().mockReturnValue({
-            returning: vi.fn().mockResolvedValue([{
-              id: 'cmd-1',
-              deviceId: deviceOne.id,
-              type: 'reboot',
-              status: 'pending',
-              createdAt: new Date()
-            }])
-          })
-        } as any)
-        .mockReturnValueOnce({
-          values: vi.fn().mockReturnValue({
-            returning: vi.fn().mockResolvedValue([{
-              id: 'cmd-2',
-              deviceId: deviceTwo.id,
-              type: 'reboot',
-              status: 'pending',
-              createdAt: new Date()
-            }])
-          })
-        } as any);
-
+      // #5128: the bulk command handler now goes through the one enqueue
+      // seam (services/dispatchDeviceCommand.ts, mocked above) instead of a
+      // raw db.insert. The decommissioned device is rejected by the route
+      // BEFORE it ever reaches the seam, so only two calls land here.
       const client = await createAuthenticatedClient(app, { mfa: true });
       const res = await client.post('/devices/bulk/commands', {
         deviceIds: [deviceOne.id, deviceTwo.id, deviceThree.id],
@@ -256,6 +266,12 @@ describe('device endpoints (authenticated)', () => {
         deviceOne.id,
         deviceTwo.id
       ]);
+      // The decommissioned device never reached the dispatch seam at all —
+      // it was rejected before it, not by it.
+      expect(mockDispatchDeviceCommand).toHaveBeenCalledTimes(2);
+      expect(mockDispatchDeviceCommand).not.toHaveBeenCalledWith(
+        expect.objectContaining({ deviceId: deviceThree.id }),
+      );
     });
   });
 
--- a/apps/api/src/__tests__/integration/deviceCommandOfflineQueue.integration.test.ts
+++ b/apps/api/src/__tests__/integration/deviceCommandOfflineQueue.integration.test.ts
@@ -14,7 +14,9 @@ import {
   scriptExecutionBatches,
   scriptExecutions,
   scripts,
+  softwareCatalog,
   softwareDeployments,
+  softwareVersions,
   users,
 } from '../../db/schema';
 import {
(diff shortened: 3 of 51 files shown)


ToddHebebrand:
## Review summary — 4 specialist passes, 9 findings acted on

Ran four independent reviews against the diff: **security/tenancy**, **correctness/concurrency**, **test quality**, and **silent failures**. All fixes are in `90d5664`.

### Fixed — critical

**1. Cancel-on-event never terminalised the owning record.** The org-move and decommission sweeps cancelled the `device_commands` row directly. A cancelled command is *terminal*, so `reapStaleDeviceCommands` (which scans only `pending`/`sent`) never revisits it — and this PR deliberately removed the software reaper's Tier-2 cancel on the premise that the command reaper is the sole owner. Net effect: the linked `script_executions` / `deployment_results` row hung `pending` **forever**, with nothing left to resolve it. That is precisely the silent death this feature exists to remove, reintroduced through a side door. Both sweeps now read `id/type/payload` before the erasing UPDATE and propagate in the **same transaction**.

**2. One bad device aborted an entire software fan-out.** `dispatchSoftwareInstallToDevice` now throws when the seam refuses — decommissioned / trust-denied checks that did **not** exist on this path before #5128. Neither fan-out loop had a try/catch, and `targetDevices` has no status filter, so a single decommissioned device anywhere in a multi-hundred-device batch aborted the loop and left every device behind it with a `pending` `deployment_results` row that was never dispatched and never failed. Both loops now isolate per device and write a real failure reason.

### Fixed — high

**3. The power-state barrier was a deterministic bypass, not a race.** The barrier lives in `partitionClaimable`, which only runs on the heartbeat claim. The seam pushed straight down the socket, so *every* reboot issued to an online device walked past it — OD-3 accepts a same-second residual, not this. Power-state types are no longer pushed from the seam; they go via the heartbeat, which also **restores the generic routes' pre-#5128 delivery timing exactly** (they raw-inserted and waited for the heartbeat anyway).

**4. An erased origin org silently satisfied the org check.** `submitted_org_id` is `ON DELETE SET NULL`, and NULL was read as "no evidence of a move". After an org merge the loser org survives as a shell, is later erased, and the provenance goes NULL underneath a device already repointed to the survivor — destroying the very fact the check compares. Now discriminated by the deadline: `deliver_by IS NULL` = a genuine legacy row (deliver it); `deliver_by` set = post-#5128, so a NULL can only mean the org was erased → refuse (`submitter_org_erased`).

**5. A claim-time trust fault could brick a device's heartbeat.** `partitionClaimable` rethrew non-trust errors; that aborts `claimPendingCommandsForDevice`'s transaction, which the heartbeat awaits and does not catch. Because the row stays `pending` it is re-selected on every retry — one deterministic fault meant the device could never check in again. It now **holds** the row (fail-closed: not delivered, not cancelled), reports to Sentry, and lets the rest of the batch through.

### Fixed — medium

**6.** The script reaper stamped "no response from agent" on commands that were *cancelled* before delivery — a false diagnosis about an agent that was never asked. Now reported as cancelled.

**7.** A comment asserted a `FOR UPDATE` lock serialises claims against org moves. It does not — the claim's `devices` read is unlocked. Rather than take a row-exclusive lock on the device row on **every heartbeat** (which would serialise the hottest path in the product against every other write to that row), the residual is now documented honestly: a few-millisecond window, blast radius one command inside the same partner, closable later by reordering the claim's locks to match org-move's.

### Fixed — test quality

**8.** The **shipped** `software_install` refresher was never exercised — every test replaced it with a fake, leaving the design's named "URL re-mint" requirement (§192) unmet. Now covered directly: the `s3Key` gate, the 3600 s TTL, the `isS3Configured()` guard, non-string keys, and presign failure.

**9.** Two assertions were vacuous and are now compiled-SQL checks:
- the claim's `deliver_by` predicate asserted only that `gt`/`isNull` were *called* — swapping `or(...)` for `and(...)` would have stayed green while withholding every row that has a deadline. It now requires the disjunction.
- the eligibility cancel's CAS asserted a call *count*; dropping `eq(status,'pending')` would have stayed green. It now checks the bound params.

Also: every `CommandTypes` value must now be **explicitly** classified. The registry's fallback is `standard` (queueable), so a new command type that must never be deferred would otherwise have become deferrable by silence — `UnregisteredCommandTypeError` does not cover that case.

### Assessed and deliberately not changed

- **`sendCommandToAgent` throws under `BREEZE_ROLE=worker`** — flagged as critical, but **pre-existing**: `queueCommandForExecution` already called it on this exact path on `main` (`commandQueue.ts:895`). The seam preserves the behaviour byte-for-byte. Belongs to the #4086 BREEZE_ROLE relay work, not W1.
- **Full permission rehydration at claim** — the requester check catches deactivation, not a mid-flight permission downgrade. Explicitly deferred to W6 behind #3985 (OD-4), and unchanged here.
- **Claim vs. org-move lock ordering** — see #7; the fix costs more than the race.

### Verified clean by the reviewers

Reaper two-clock truth table (no un-reapable row; the new OR arm also keeps `MAX_REAP_PER_RUN` from being starved by far-future rows), CAS fences on both terminal branches, the `commandQueue ↔ dispatchDeviceCommand` cycle (function-body use only), `propagateTimedOutDeviceCommand`'s extra writes (guarded on non-terminal status, so races are idempotent no-ops), the cancel route's permission/org/site scoping and IDOR safety, the `?status=` filter, `s3Key` exposure, and payload erasure on **every** terminal path.

### Verification after fixes

`tsc --noEmit` clean · `pnpm lint` clean (6/6) · **1484 unit tests passing** across every touched module · **21/21 integration tests against real Postgres** (now including cancel-propagation and cancelled-vs-silent-agent cases) · `tenantCascade`, `tenant-export-policy`, `rls-coverage` (102), `autoMigrate`, `migrationRlsScope`, migration-naming — all green.

ToddHebebrand:
## Reviews run on this PR (orchestrator record)

| # | Review | Head covered | Outcome |
|---|---|---|---|
| 1 | Fixer's 4 specialist passes (security/tenancy, correctness/concurrency, test quality, silent failures) — comment above, 03:36Z | pre-fix diff (before `90d5664`) | 9 findings, fixed in `90d5664fae` |
| 2 | Orchestrator `/pr-review-toolkit:review-pr` — 5 agents (code review / silent-failure hunter / test analyzer / type design / comment accuracy) | `90d5664fae` (the fix commit above had not been reviewed) | 2 critical, 3 high, 4 medium, 5 low + 3 test gaps — all fixed in `1c4d3969ef`, `05d20895db`, `c94759aea2`, `c90050818f` |
| 3 | CI on `90d5664fae` (run 34080202436) | `90d5664fae` | RED: Test API (partnerTrust.sites / partnerTrust static scans, agentEditionCompat #4093 allowlist — all from the `CommandTypes` extraction) + Integration shard 3 (deviceUninstallDrain: decommission now cancels ordinary pending rows, so the "still pending" assertion became "cancelled + never delivered") — fixed in `4aa7f8b399` / `c94759aea2` |

**Findings fixed by review 2 (verified in code, not taken from the report):**
- CRITICAL claim-time cancellation (`partitionClaimable`) never terminalised the owning `deployment_results` / `script_executions` row → now propagates inside the claim transaction (`services/commandCancelPropagation.ts`).
- CRITICAL legacy pending `software_install` rows (`deliver_by IS NULL`) fell from the retired 7-day special case to the 2-hour execution timeout → reaper keeps 7 days for those legacy rows.
- HIGH expiry/cancel propagation skipped `script_execution_batches` counters and `applyAutomationActionTerminal` → shared `scriptExecutionTerminal.ts` used by every terminal path.
- HIGH power-state commands lost the socket push for `queueCommandForExecution` callers (maintenance reboot, fleet findings) and reject-policy rows were stamped a 5-minute deadline (cut the watchdog/heartbeat window from 30 min) → push gated by the same in-flight predicate as the heartbeat; reject rows keep `deliver_by` NULL (legacy clock, byte-identical to main).
- MEDIUM `result.status='expired'` broke late-result acceptance (`commandResultAcceptance` keys on `'timeout'`) → marker stays `timeout`, clock carried in `reason`/`clock`.
- MEDIUM `requester_inactive` (and any other claim-time cancel) could cancel a queued `self_uninstall` → drain rows exempt from every claim-time cancel.
- MEDIUM trust-before-offline changed the `Device is <status>` string four backup routes parse → offline-reject checked first.
- LOW/tests: unguarded release + silent refresher now captured; cancel route atomic; org-move excludes `self_uninstall` like decommission; orphaned `deployment_results` (deleted command row) now fail instead of sitting forever; Sentry throttle on eligibility faults; tests for the WS `software_install` reconciliation branch, route-level cancel propagation, cancel CAS-lost; hold/cancel reason unions split; guarded registry registration; spec §B/§C/§D corrected.

**Verification of the fix commits** (these were NOT put through a second review-pr pass): full API unit suite 1940 files / 36,167 tests passed; `deviceCommandOfflineQueue` + `deviceUninstallDrain` integration files 34/34 on a private Postgres; tsc clean; lint 6/6; orchestrator spot-checked each critical fix in the diff and re-ran 18 files / 453 tests.

**Deploy note:** the migration's `CREATE INDEX IF NOT EXISTS idx_device_commands_deliver_by` is non-concurrent (the runner wraps each file in a transaction), so it takes a write lock on `device_commands` for the build. The index is partial (`status='pending' AND deliver_by IS NOT NULL`), so on a healthy table the set is small and the build is short — but roll during a quiet window on large tenants.

Flag `DEVICE_COMMAND_OFFLINE_QUEUE_ENABLED` remains **off** in this wave.

Issue #5161, opened 2026-09-07
W06: Android client: StrongBox/TEE key + Key Attestation + Play Integrity native module


ToddHebebrand:
Wave W06 of #4707 — W06: Android client: StrongBox/TEE key + Key Attestation + Play Integrity native module
Branch: `feature/4707-mobile-platform-attestation-l4/wave-5161`

PR #5163, opened 2026-09-07
feat(mobile): iOS Secure Enclave P-256 + App Attest client (#4707 W05)

### LINKED ISSUE #5160: W05: iOS client: Expo native module (Secure Enclave P-256 + App Attest) + attestingSigner wiring
Wave W05 of #4707 — W05: iOS client: Expo native module (Secure Enclave P-256 + App Attest) + attestingSigner wiring
Branch: `feature/4707-mobile-platform-attestation-l4/wave-5160`

ToddHebebrand:
Closes #5160

Wave W05 of #4707 (spec issue #1374). The API side (W01–W04) is already on `main`; this is the iOS client that finally makes an approver key reach **L4**.

## What changed

| Area | Change |
|---|---|
| `apps/mobile/modules/breeze-attestation/` | New local Expo module. Swift: Secure Enclave P-256 key + Apple App Attest. |
| `apps/mobile/src/services/authenticatorTranscript.ts` | Client mirror of the server's `registrationTranscript`, pinned to the API's own test vector. |
| `apps/mobile/src/services/attestingSigner.ts` | Optional-require + null-object wrapper over the native module. |
| `apps/mobile/src/services/approverDevice.ts` | Two-step `challenge` → `verify` registration; fails closed. |
| `apps/mobile/src/services/hardwareSigner.ts` | Comment correction only (plan Task 1 Step 4). |
| `apps/mobile/app.json` | App Attest entitlement. |

### Swift constraints (all held, none relaxed)

- `SecKeyCreateRandomKey` with `kSecAttrTokenID: kSecAttrTokenIDSecureEnclave`, `kSecAttrKeyTypeECSECPrimeRandom`, 256 bits.
- `SecAccessControlCreateWithFlags(nil, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, [.privateKeyUsage, .biometryCurrentSet])` — `.biometryCurrentSet`, not `.biometryAny`, so a new Face ID / fingerprint enrolment invalidates the key.
- A **fresh `LAContext` per signature**. A cached, already-evaluated context would carry a live authentication the OS reuses, so the biometric prompt would silently not appear while "the user approved this" still looked true.
- `SecKeyCopyExternalRepresentation` returns the bare X9.62 point; the module wraps it in a real P-256 SPKI DER header in Swift, because that is what the server re-parses (`crypto.createPublicKey({ format: 'der', type: 'spki' })`).
- App Attest behind a `DCAppAttestService.shared.isSupported` guard.

## Decisions (settled before implementation — recorded so review doesn't relitigate)

1. **The transcript formula lives in `apps/mobile`, not `@breeze/shared`.** The plan proposed hoisting it into `packages/shared/src/utils/authenticatorTranscript.ts`. `apps/mobile` deliberately has **no** `@breeze/shared` dependency — Metro resolves no workspace packages here, so such an import fails at runtime rather than at build time (see the header comments in `ticketAttachmentContract.ts` and `ticketPushPrefs.ts`, which mirror server constants for exactly this reason). A shared module the phone cannot import is a worse illusion of one source of truth than an honest mirror. What keeps the two honest instead is **the same test vector**: `authenticatorTranscript.test.ts` pins the identical inputs and digest that `apps/api/src/services/authenticatorAttestation.test.ts` pins in its "matches the exact documented pre-image" case, and asserts the pinned digest really is SHA-256 of the pinned pre-image so the vector cannot be edited to rescue a broken implementation.
2. **The TS surface covers both platforms so W06 only adds Kotlin.** `BreezeAttestation.types.ts` declares `AttestedKey`, `IosAttestation`, `AndroidAttestation`, `PlatformAttestation` and the full method set, including `createAttestedKey(opts?: { attestationChallengeB64?: string })`. `expo-module.config.json` already lists `"platforms": ["apple", "android"]` with the Android module class `expo.modules.breezeattestation.BreezeAttestationModule`. **No `android/` directory is created here — W06 owns `modules/breeze-attestation/android/**` entirely.**
3. **`attestingSigner.ts` is one module for both platforms.** `attestationPlatform()` reads `Platform.OS` (lazily — a static `react-native` import would take down every Vitest suite that transitively reaches this file). On Android before W06 lands, the native module does not resolve and the signer reports unavailable exactly as it does on an unsupported iPhone.
4. **Fail closed.** A device that supports attestation and then fails it returns `{ status: 'failed', reason: 'attestation_failed' }`. It never silently retries on the legacy unattested endpoint — that would leave a technician believing their phone can approve critical requests when it cannot, with nothing anywhere saying so. The legacy single-POST path runs **only** when `isAttestationAvailable()` is false.
5. **`ApproverRegistrationOutcome.registered` now carries `attested: boolean`**, so W07's banner can be honest about which path succeeded instead of the user discovering it at approval time.

### Two easy-to-get-wrong protocol details, pinned in code and comments

- **`signWithAttestedKey` signs the UTF-8 bytes of the base64 string as given** — it does *not* decode it. The server verifies the registration PoP over `transcript.toString('base64')` as UTF-8 (`verifyMobileSignature`). Decoding on the client would produce a signature that verifies nowhere, on every device at once.
- **`attestApp` *does* decode**, because App Attest's `clientDataHash` commits to the raw 32 digest bytes (`SHA256(authData || clientDataHash)` is what the server recomputes).

## W06 hand-off (Android, #5161)

The branch is pushed, so W06 can start now. It needs to add **only**:

- `modules/breeze-attestation/android/**` (Kotlin `BreezeAttestationModule`, package `expo.modules.breezeattestation`) — `expo-module.config.json` already points at it, so **do not edit that file**.
- Nothing in `BreezeAttestation.types.ts`, `attestingSigner.ts` or `approverDevice.ts`: the Android branch of the client flow is already wired and tested (`sends the platform the RUNTIME reports…` covers it). `createAttestedKey` already receives `{ attestationChallengeB64: <server challenge> }`.
- Note the ordering constraint the transcript creates on Android: the transcript commits to the public key, so it cannot also be the key-generation challenge. This client passes the **server challenge** at key-gen time; the API-side check for that binding is being fixed separately by the orchestrator and is out of scope here.

## Verification

```
apps/mobile: npx vitest run          -> 110 files, 1410 passed | 3 skipped
             npx tsc --noEmit        -> clean
             npx expo prebuild -p ios --clean --no-install
               -> ios/BreezeRMM/BreezeRMM.entitlements contains
                  com.apple.developer.devicecheck.appattest-environment = production
             npx expo-modules-autolinking search -p ios
               -> 'breeze-attestation' resolved
```

`ios/` is generated and gitignored and is not committed. (`--clean` does delete the tracked `ios/ci_scripts/ci_post_clone.sh`; it was restored — worth knowing before anyone else runs prebuild.)

### Review outcome

`/pr-review-toolkit:review-pr` — code-reviewer, silent-failure-hunter, pr-test-analyzer. **Two real findings, both fixed in `6373d03c` before merge; both were the same class of bug — a silent downgrade of assurance:**

1. **`gatherApprovalProof` signed every approval with the legacy RSA key**, regardless of how the device registered. The server verifies an assertion with the algorithm on the **device row** (`authenticatorAssurance.ts`: *"the algorithm comes from the DEVICE ROW, never from the proof"*), so an attested ES256 row would have rejected every proof this phone offered — registering at L4 and then approving at L1 with nothing saying so. This would have failed Todd-gate Step 3 outright. Registration now records `breeze_approver_attested`; the approval path selects the matching signer, and returns `null` rather than offering an RSA signature the row cannot verify.
2. **A rejected native availability probe was coerced to `false`**, sending the device down the legacy unattested path. "Unknown" is not "no" — it now propagates and surfaces as `attestation_probe_failed`.

`signTranscript` was renamed `signPayload` in the same pass: it signs the UTF-8 bytes of whatever string it gets, and now has two callers with differently-shaped payloads.

Independently verified during review (not taken on trust): the Swift `p256SpkiHeader` bytes are byte-for-byte Node's own P-256 SPKI prefix and the rebuilt DER reparses as `prime256v1`; and a DER-ECDSA signature over the UTF-8 of the base64 transcript verifies server-side while the decode-first variant provably does not.

New tests: 11 transcript cases, 8 attesting-signer fallback cases, 18 attested-registration cases (challenge→verify ordering, server-chosen transcript, three-way binding, fail-closed on each of key-gen / attest / PoP / probe, single-use `attemptId`, legacy fallback only when unavailable, attested marker written only on the attested path), and 4 approval-signer-selection cases. The selection guards were verified **discriminating by mutation** — forcing the legacy signer reddens exactly those three tests.

`apps/mobile` has no `lint` script, so no lint run is reported.

## Todd gate — on-device verification (not done)

**The Swift path is unverified.** It cannot run in CI — there is no Secure Enclave and no App Attest on a runner or simulator — and I have no physical device. The JS/TS contract around it is tested; the Swift itself is reviewed code, not proven code. Do not treat this wave as complete until the following pass, from plan Task 3:

- [ ] **Step 1:** Build a dev-client (`eas build --profile development --platform ios`) and install on a physical device.
  - ⚠️ `app.json` declares `appattest-environment: "production"` unconditionally, while the `development` EAS profile signs the binary for development. If `attestKey` fails at the device level in Step 2, **check this first** — the entitlement value has to match how the binary is signed. Left as-is here because `app.config.js` (which is where a per-profile value would have to be resolved) is outside this wave's file ownership; if the dev-client build does trip on it, that is the one-line follow-up.
- [ ] **Step 2:** Sign in; confirm the server row lands with `platform_bound_basis = 'ios_se_p256_app_attest'`, `attestation_verified_at` non-null, `public_key_alg = 'ES256'`, `possession_verified_at` non-null, `last_used_at` null.
- [ ] **Step 3:** Approve a **critical-tier** PAM elevation from the phone; confirm `decided_assurance_level = 4` and `last_used_at` is now set.
- [ ] **Step 4:** Enroll a new Face ID identity; confirm the key is invalidated (`.biometryCurrentSet`) and the app reports `failed` rather than silently signing.
- [ ] **Step 5:** Record the results in the PR body. **This is the only proof the Swift path works** — do not claim the wave is done without it.

https://claude.ai/code/session_01U6G1qz2WFKfwMAhZU4FCV4


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer, silent-failure-hunter, pr-test-analyzer), all three in parallel against `gh pr diff` — no reviewer was given a live worktree path.

**Findings:** 2 raised → both addressed in `6373d03c`; 0 outstanding. Both were the same class of bug, a **silent downgrade of assurance**, which is the one failure mode this wave exists to prevent:

1. *(pr-test-analyzer, then confirmed by me against the merged API)* `gatherApprovalProof` still signed **every** approval with the legacy Keychain RSA key, whatever path the device registered through. `services/authenticatorAssurance.ts` verifies an assertion with the algorithm stored on the **device row** ("the algorithm comes from the DEVICE ROW, never from the proof"), so an attested ES256 row would have rejected every proof the phone offered — the device registers at L4 and then approves at L1, with nothing anywhere saying so. Todd-gate Step 3 would have failed. Fixed: registration records `breeze_approver_attested`, the approval path picks the matching signer, and it returns `null` rather than offering an RSA signature that row cannot verify.
2. *(silent-failure-hunter)* `nativeAttestingSigner.isAvailable()` swallowed a rejected native probe and reported `false`, which routed the device to the legacy unattested path. A probe that throws means "unknown", not "no"; it now propagates and surfaces as `attestation_probe_failed`. `RootNavigator` already forwards `failed` outcomes to Sentry with the reason as a tag, so it is visible rather than swallowed.

Also renamed `signTranscript` → `signPayload` in the same pass — it signs the UTF-8 bytes of whatever string it is handed, and now has two callers with differently-shaped payloads (the base64 transcript, and the approval nonce).

**Verified independently rather than taken on trust:** the Swift `p256SpkiHeader` constant is byte-for-byte Node's own P-256 SPKI prefix and the rebuilt DER reparses as `prime256v1`; and a DER-ECDSA signature over the UTF-8 of the base64 transcript verifies server-side while the decode-first variant provably does not. The three new approval-signer-selection guards were checked **discriminating by mutation** — forcing the legacy signer reddens exactly those three tests and nothing else.

**Not fixed, deliberately** (raised by code-reviewer as a watch-item, not a defect): `app.json` declares `appattest-environment: "production"` unconditionally while the `development` EAS profile signs for development. A per-profile value belongs in `app.config.js`, which is outside this wave's file ownership. It is called out inline in the Todd gate as the first thing to check if `attestKey` fails on device.

**Tests:** `apps/mobile` full suite green (`npx vitest run` → 110 files, 1410 passed | 3 skipped); `npx tsc --noEmit` clean. `apps/mobile` has no `lint` script, so none was run. Entitlement confirmed through a real `npx expo prebuild -p ios`, and `expo-modules-autolinking` resolves `breeze-attestation`.

**Status:** review-clean, awaiting maintainer merge. **The Swift path itself is still unverified** — it cannot run in CI, and the Todd gate in the PR body is the only proof it works.

ToddHebebrand:
**Review round 2:** `/pr-review-toolkit:review-pr` (code-reviewer, silent-failure-hunter, pr-test-analyzer — Opus, in parallel, against `gh pr diff`; no reviewer had a worktree). Warranted because 6373d03c touched the signing path.

**Findings:** 5 raised → all fixed in 3899e3d1; 0 outstanding.

1. **CRITICAL (verified against `routes/authenticator.ts` on main)** — `/devices/mobile/verify` 201s on a *rejected* attestation (row stored as `platformBoundBasis: 'unattested'`). The client only checked `res.ok`, so a wrong `appattest-environment` or stale server recorded `attested: true` on a phone capped at L3, and the on-device gate would have read "registered" while the server row said otherwise. Now reads the basis off the returned device → `{ registered, attested: false, reason: 'attestation_rejected_by_server' }`, Sentry-tagged, banner detail says sign-in will not fix it.
2. **Important (3/3 reviewers)** — `breeze_approver_attested` was never cleared; a legacy re-registration after an attested one left the marker behind and pinned the ES256 signer to an RSA row (every approval rejected, no error). Legacy path now deletes it, before the credential id write.
3. **Important** — `gatherApprovalProof` coerced a rejected `isAvailable()` to `false` → silent L1 approval from an L4 device, the inverse of round-1 finding #2. Now propagates like a cancelled prompt.
4. **Important (inferred; Swift, not CI-testable)** — `secureEnclaveAvailable()` only built an ACL object, which never consults the SE/passcode/biometrics; combined with fail-closed, a no-biometrics phone could register *nothing*. Now gated on `LAContext.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics)`. Also: pre-generation `SecItemDelete` status no longer discarded (two keys under one tag = valid signature under the wrong key), `kSecAttrKeyClassPrivate` pinned, deprecated `kSecUseOperationPrompt` dropped.
5. Podspec: iOS only (`DCAppAttestService` absent on tvOS).

Each JS fix mutation-checked (revert → exactly its own test reds). Mobile suite 110 files / 1416 passed, tsc clean.

**Not changed (decision for the orchestrator):** a self-hosted server older than W02 404s `/challenge` → `http_404` fail-closed on every iOS client after this app ships. Reviewers suggested treating 404 as "protocol absent → legacy", not "attestation refused". Left as-is because it is a policy call; flag if you want it.

**On-device checklist additions** (only a physical iPhone can prove these): biometric prompt appears on *every* approval, three in a row; no-passcode device → `attestation_failed`, not a crash; App-Attest-unsupported device still registers legacy at L2/L3; after reinstall the phone re-registers cleanly; server row shows `ios_se_p256_app_attest` — if it shows `unattested` the phone now says so too.

ToddHebebrand:
Todd's call on the open item: a 404 from `/devices/mobile/challenge` now falls back to legacy registration with `reason: 'attestation_protocol_absent'` (be2a14d0). Raised before any key is minted or grant consumed; two tests cover the fallback and a failing legacy POST after it.

PR #5165, opened 2026-09-07
fix(api): Android KeyStore attestation challenge cannot embed the SPKI — add pre-key keygen digest (#4707)

ToddHebebrand:
Part of #4707 (mobile platform attestation for L4). Unblocks W06 (#5161).

## Defect
`verifyAndroid` passed `registrationTranscript` (which embeds the key's SPKI) as the expected KeyStore `attestationChallenge`. Android fixes that challenge at **key generation** (`KeyGenParameterSpec.setAttestationChallenge`), before the SPKI exists — no real client can produce it. iOS is unaffected (App Attest `clientDataHash` is computed after the key exists).

## Fix
- New `androidKeyGenChallenge({attemptId, challenge, publicKeyAlg})` = `SHA256('breeze.authenticator.mobile-register.keygen.v1\n…')`, pinned by a pre-image test (W06 Kotlin mints it on-device). 32 bytes, under Android's 128-byte cap.
- `verifyPlatformAttestation` takes `keyGenChallenge`; the route derives it from the CONSUMED attempt (test added). Android cert challenge compares against it; PoP signature and Play Integrity `requestHash` keep the full transcript.
- Key identity remains bound by the existing attested-leaf-key == registered-SPKI digest check.
- Play Integrity: drop `sub` from the service-account JWT (delegation-only claim). **Not verified against live Google** — aligns with the documented non-delegated flow.
- Plan doc + stale "one digest, three bindings" comments updated.

## Security argument (advisor quorum: Fable + codex xhigh, agreed)
Attempt is single-use (`GETDEL`); reusing one keygen challenge for K1 and K2 yields two independently attested candidates, `cert(K1)+SPKI(K2)` fails equality, only one submission can consume the attempt. No replay or mix-and-match opened.

## Verified
- `npx vitest run` on attestation/assurance/authenticator suites: 8 files, 245 tests pass (10 new/changed tests were red first).
- `tsc --noEmit` clean, eslint clean.

## Follow-ups (quorum findings, out of scope here)
- `androidKeyAttestation.ts` package check reads only the first package name and ignores `signatureDigests`.
- Certificate revocation status list not checked (documented as unimplemented).
- User-auth check only rejects `noAuthRequired`, does not verify hardware-enforced auth type.

https://claude.ai/code/session_01LVfXHssL5fG9RUTYcedpPn

### DIFF
--- a/apps/api/src/routes/authenticator.test.ts
+++ b/apps/api/src/routes/authenticator.test.ts
@@ -76,6 +76,7 @@ const {
       issueRegistrationAttempt: vi.fn(),
       consumeRegistrationAttempt: vi.fn(),
       registrationTranscript: vi.fn(),
+      androidKeyGenChallenge: vi.fn(),
       verifyPlatformAttestation: vi.fn(),
     },
     // The per-user rate limiter is mocked as a REAL counter keyed exactly as the
@@ -1119,6 +1120,7 @@ describe('attested mobile registration (#1374 W02)', () => {
     attestationMocks.issueRegistrationAttempt.mockResolvedValue(ATTEMPT);
     attestationMocks.consumeRegistrationAttempt.mockResolvedValue(ATTEMPT);
     attestationMocks.registrationTranscript.mockReturnValue(Buffer.alloc(32, 9));
+    attestationMocks.androidKeyGenChallenge.mockReturnValue(Buffer.alloc(32, 5));
     attestationMocks.verifyPlatformAttestation.mockResolvedValue({
       basis: 'unattested',
       verifiedAt: null,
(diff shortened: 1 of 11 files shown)


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` (code-reviewer + pr-test-analyzer, Sonnet, both on `gh pr diff`), plus an advisor quorum (codex `xhigh`) on the design before implementation.

**Findings:** 1 raised → fixed in `fb7fa5c190`; 0 outstanding.

1. *(pr-test-analyzer)* The Play Integrity `sub`-claim removal had zero coverage: every existing test injects `decodeIntegrityToken`, so `accessToken()` was never invoked, and a wrong claim would degrade silently to `null` in production. Added a test that runs the real google-managed decoder path with `fetch` stubbed and decodes the assertion sent to the token endpoint (iss/aud/scope present, no `sub`). Mutation-checked: re-adding `setSubject` fails exactly this test.

code-reviewer: no findings ≥80 confidence. Verified single production call site, route derives the keygen digest from the CONSUMED attempt, `verifyAndroid` splits cert-challenge (keygen digest) from Play Integrity `requestHash` (full transcript), E2E fixtures no longer conflate the two.

Quorum's out-of-scope hardening findings filed as #5166.

Issue #2797, opened 2026-07-25
[Agent] Shipped binaries embed the literal EICAR test string — VirusTotal confirms Elastic flags breeze-agent.exe and breeze-user-helper.exe as Multi.EICAR


ToddHebebrand:
## Description

The agent's own threat-signature table stores the EICAR test string as a literal byte slice, so the string is compiled verbatim into the shipped binaries' read-only data. Antivirus engines that scan for EICAR find it and flag our binary.

This is confirmed against real scanners, not theorised.

## Evidence

VirusTotal lookups against the **v0.101.0 release artifacts** (2026-07-25):

| Artifact | SHA256 | VT verdict |
|---|---|---|
| `breeze-agent-windows-amd64.exe` | `9b946717…b155e2` | **2 malicious** — Elastic `Multi.EICAR`, Microsoft `Trojan:Win32/Wacatac.B!ml` |
| `breeze-user-helper-windows-amd64.exe` | `24387a46…6805e1` | **1 malicious** — Elastic `Multi.EICAR` |
| `breeze-backup-windows-amd64.exe` | `449703fd…29a77b` | clean (69 undetected) |
| `breeze-watchdog-windows-amd64.exe` | `e2606242…ddb6e5` | clean (69 undetected) |
| `breeze-agent.msi` | `c0a976a7…11afac` | clean (63 undetected) |
| `breeze-helper-windows.msi` | `556d95f1…74b4eb` | not yet submitted |

Running `strings` over the same artifacts correlates perfectly — the two binaries VT flags are exactly the two that contain `EICAR-STANDARD-ANTIVIRUS-TEST-FILE`, and the four clean ones do not.

## Root cause

`agent/internal/security/threats.go:38-45`:

```go
var defaultThreatSignatures = []threatSignature{
	{
		Name:             "EICAR-Test-File",
		Type:             "malware",
		Severity:         ThreatSeverityHigh,
		FilenamePatterns: []string{"eicar"},
		ContentPattern:   []byte("X5O!P%@AP[4\\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*"),
	},
	…
```

A `[]byte` literal lands in the binary's data section as-is. Any scanner looking for that 68-byte sequence matches our binary, exactly as designed — EICAR exists precisely so that finding it anywhere triggers a detection.

[...]

PR #5167, opened 2026-09-07
feat(scripts,agent): acknowledge Strict security patterns per script (#5129)

### LINKED ISSUE #5129: [Agent][Scripts] Script security validator is hardcoded to Strict — HKLM registry writes are unconditionally blocked with no override path
## Description

A user tried to run a script that sets a registry key and got:

```
script validation failed: potentially dangerous pattern detected: PowerShell HKLM modification
```

They asked how and where to authorise the script. **There is no answer** — the block is hardcoded in the agent and cannot be overridden by anyone: not a partner admin, not an org admin, not a self-hoster editing their own config.

Setting an HKLM value is one of the most common things an MSP tech does on Windows, and Breeze's core job is config management. Today the only way to do it is to abandon the scripts feature and use Remote Tools → Registry (interactive, one device at a time), or to write the script in a way that evades a regex.

## Root Cause

The validator level is a literal at the construction site:

`agent/internal/executor/executor.go:93`
```go
validator: NewSecurityValidator(SecurityLevelStrict),
```

`SecurityLevelNone` / `SecurityLevelBasic` exist in `agent/internal/executor/security.go:15-20` but nothing ever passes them. Grepped `agent/` and `apps/api/` for `SecurityLevel`, `skipValidation`, `allowDangerous`, `bypassValidation`, `overrideValidation` — no config field, env var, agent flag, configuration-policy setting, per-script flag or per-org setting reaches this. `validateScript` (`executor.go:321`) delegates straight to `validator.Validate` with no allowlist.

The three Strict-level registry patterns (`agent/internal/executor/security.go:117-119`), case-insensitive, matched against the script body *after* parameter substitution:

```go
{`reg\s+add\s+HKLM`,            "HKLM registry modification"},
{`Set-ItemProperty\s+.*HKLM`,   "PowerShell HKLM modification"},   // ← the user's hit
{`New-ItemProperty\s+.*HKLM`,   "PowerShell HKLM property creation"},
```

## The control does not do what it looks like it does

It is a per-line substring filter. It blocks the admin writing idiomatic PowerShell, and does not stop anyone who wants around it. All of the following set HKLM values and pass validation today:

- `Set-Item` and `New-Item` are not in the pattern list at all.
- `Registry::HKEY_LOCAL_MACHINE\...` contains no `HKLM` token.
- Go's `regexp` has no `(?s)` here, so `.` never crosses a newline — putting the path in a variable on the preceding line defeats `Set-ItemProperty\s+.*HKLM` outright:
  ```powershell
  $p = 'HKLM:\SOFTWARE\Contoso'
  Set-ItemProperty -Path $p -Name Enabled -Value 1   # passes
  ```
- `reg import` is not covered.

So the pattern costs us legitimate use and buys close to nothing against a determined script author.

## Proposed Fix

Keep **Basic**-level patterns (`rm -rf /`, `Format-Volume`, fork bombs, block-device writes) as unconditional hard blocks — those are almost always an accident and there is no legitimate RMM use.

Move the **Strict** set from *refuse-on-execute* to *flag-on-save*:

1. Script editor runs the Strict patterns at save time and surfaces which patterns matched.
2. An admin acknowledges the risk once on the script record (persisted, audited, shown in the script list).
3. Dispatch carries an `acknowledgedRisk` bit; the agent skips Strict-level patterns only when it is set. Basic patterns still block regardless.

This puts the decision in front of a human once, on the record, instead of at the agent where there is no human and no context.

Cheaper alternative if the above is too much for one wave: a partner/org configuration-policy setting for the validator level. Worse — it weakens the control globally rather than per-script, and it is still a coarse regex — but it at least unblocks users.

Either way, the error string should say *why* and *what to do next*. Today it names the pattern and stops, which is what generated this question.

## Also: the docs are wrong

`apps/docs/src/content/docs/features/scripts.mdx:541`

> The agent's security validator operates at the **Strict** level **by default** …

"by default" implies a knob that does not exist. Fix that line regardless of what happens to the rest of this issue — a user reading it will go hunting for a setting and find nothing.

## Affected Files

- `agent/internal/executor/executor.go:93` (hardcoded level), `:321` (`validateScript`)
- `agent/internal/executor/security.go:15-20` (unused levels), `:117-119` (registry patterns)
- `apps/docs/src/content/docs/features/scripts.mdx:537-556` (docs)
- Script dispatch payload + script record schema (for the acknowledgement bit)

## Workarounds today

- **Remote Tools → Registry** tab (Windows only) — full read/create/set/delete, audited as `set_registry_value`, bypasses the script validator entirely. API: `PUT /devices/:deviceId/registry/value` (`apps/api/src/routes/systemTools/registry.ts:326`). Interactive and per-device — not a substitute for a script across a fleet.
- **Remote Tools → Terminal** — live PTY, also unvalidated.

## Reported By

Support question relayed by @toddhebebrand.

ToddHebebrand:
Closes #5129

## The problem

`agent/internal/executor/executor.go` constructed its validator at `SecurityLevelStrict` with a hardcoded literal, and **nothing anywhere in the product could override it** — not a partner admin, not an org admin, not a self-hoster editing their own config. So `Set-ItemProperty -Path 'HKLM:\...'` — routine MSP configuration work, and arguably Breeze's core job — failed with:

```
script validation failed: potentially dangerous pattern detected: PowerShell HKLM modification
```

…and the message named the pattern and stopped. There was no answer to "how do I authorise this?", which is what generated the support question.

## The design

**Basic-level patterns stay unconditional hard blocks, always.** `rm -rf /`, `Format-Volume`, fork bombs, block-device writes. No override path, ever.

**Strict-level patterns become acknowledgeable per script.** The script record stores **the set of acknowledged pattern descriptions** — e.g. `["PowerShell HKLM modification"]` — not a bare boolean. That is the security-critical part:

> A boolean would mean acknowledging an HKLM write permanently disarms Strict checking for that script, so a later edit that introduces a credential-dumping pattern would inherit the approval silently. With the description set, the existing approval stands and the newly-introduced pattern is unacknowledged and still blocks.

The rule applied on **every** save is:

```
stored = (what the caller asked for) ∩ (patterns the content actually matches)
```

Both halves matter. Intersecting with the live match set is what stops anyone pre-acknowledging the whole vocabulary once and permanently disarming the check. Carrying the existing set forward when the field is absent is what stops a rename or a timeout change from silently revoking an approval.

## What changed

**Agent** (`agent/internal/executor/`, `heartbeat/`, `userhelper/`)
- `SecurityValidator.ValidateWithAcknowledgements(content, acknowledged)`. Basic match → block, always, and the message says the block cannot be overridden. Strict match whose description is in the set → allow. Strict match not in the set → block, as today. `Validate()` is preserved as `ValidateWithAcknowledgements(content, nil)`.
- **Absent/empty set behaves exactly as before this PR** (fail closed), so an old API talking to a new agent cannot silently loosen anything.
- **Both** receive sites decode `acknowledgedSecurityPatterns` — the daemon and the `runAs=user` helper. The helper runs the same validator, so a one-sided decode would make an acknowledged script succeed in SYSTEM context and refuse in user context: exactly the #4882 failure shape.
- Error text now says what to do next (acknowledge it on the script, or use Remote Tools for a one-off).

**Shared** (`packages/shared/src/utils/scriptSecurityPatterns.ts`)
- The web/API mirror of the Strict list. **A contract test parses `agent/internal/executor/security.go` and fails on any drift** in pattern source or description — a fixture would just be a third copy that rots. The three credential-tool tokens stay XOR-obfuscated, same as the Go source (#2797).
- Descriptions are protocol values and are deliberately **not translated**: the agent compares them byte-for-byte.

**API**
- Three new columns on `scripts` + migration `2026-10-13-110000-scripts-security-acknowledgement.sql`.
- `services/scriptSecurityAcknowledgement.ts` owns the resolution rule. A description outside the closed vocabulary is a **400**, never a silently-dropped approval the admin believes they granted. Basic descriptions are outside that vocabulary by construction.
- Clamped again at `insertScriptRow` — the chokepoint shared with the bundle importer — the same way `isSystem` is.
- Audited as its own action (`script.security_acknowledgement`) with who/when/which, so it is findable without diffing `script.update` payloads.
- **Both** dispatch build sites carry the field: `scriptDispatch.ts` (paths A–D) and `fleetFindings/dispatch.ts`, which the file's own docblock calls "the FOURTH script-dispatch path and the only one that does not go through `dispatchScriptToDevice`". A `raw` ad-hoc source never carries one — there is no human decision on file for content that exists only for one dispatch.
- Reuses `SCRIPTS_WRITE` + MFA. **No new permission**: anyone who can rewrite the script body can already make it do whatever the pattern describes.

**Web** — `ScriptSecurityReview` renders under the editor only when the content matches a Strict pattern, with one checkbox per pattern and a plain-language explanation of each. No "acknowledge all" control, deliberately: a bulk affordance is how a per-pattern approval degrades into the blanket flag this design avoids. Basic patterns never appear, so the UI cannot promise something the agent will not honour.

**Docs** — `scripts.mdx` said the validator "operates at the **Strict** level **by default**", implying a knob that has never existed. Rewritten to describe the real two-level model and to state plainly that no global/partner/org validator-level setting exists.

## Tenancy registration

`scripts` is org-cascade-registered, so the **export policy fires on a new COLUMN**. All three are classified in `CORE_TENANT_EXPORT_POLICY`. `acknowledged_security_patterns` is **`excludedOpen`** — a grant list is a capability list per the CLAUDE.md rules. No other cascade list changes: `scripts` was already in `CORE_ORG_CASCADE_DELETE_ORDER`, `orgMergeRegistry`, and the partner-axis allowlist, and it has no `device_id`. RLS is unchanged (the dual-axis policies already govern row access). The migration writes no rows, so it needs no `breeze.scope` elevation.

## Verification

| Suite | Result |
|---|---|
| `go test -race ./internal/executor/... ./internal/heartbeat/... ./internal/userhelper/...` | pass |
| API affected (10 files: routes/scripts, scriptWrite, both dispatch paths, acknowledgement service, migration guards) | 462 passed, 2 skipped |
| Web `src/components/scripts/` | 262 passed |
| Shared mirror + Go-parity contract | 15 passed |
| `tenant-export-policy` (real Postgres) | 3 passed |
| `tenantCascade` + `tenantExportErasureRoundtrip` (real Postgres) | 8 passed |
| `rls-coverage` (real Postgres) | 102 passed |
| `tsc --noEmit` (api, web) · `pnpm lint` · `gofmt` · `go vet` | clean |

Each new guard was proven to **discriminate**, not just pass: the Go-parity test was shown red against a one-character description drift; the daemon and helper decode tests red with their decode line removed; the route suite red with the column write removed.

## Notes for review

- **Regex-engine divergence between RE2 and JS is fail-safe in both directions.** A pattern the TS mirror misses cannot be acknowledged, so the agent still blocks it; a pattern only the mirror matches lets an admin acknowledge something harmless. Neither direction can loosen the agent. The contract test makes drift loud anyway.
- The agent validates content **after** parameter substitution while the API matches the **stored** body. If a parameter value introduces a new pattern it is unacknowledged and blocked — fail closed, by design.
- Not done, per the issue's own framing: no partner/org configuration-policy setting for validator level. That weakens the control globally rather than per script.

https://claude.ai/code/session_01CaAbfRenxMi1vHv88DqN5s


ToddHebebrand:
**Review run:** `/pr-review-toolkit:review-pr` — 4 agents in parallel (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer), then a **second, focused code-reviewer round** on the security fix that came out of the first.

**Findings:** 8 raised → 8 addressed; 0 outstanding.

**Round 1 — 1 critical, 5 important, 2 suggestions**

- **CRITICAL — real bypass in `scriptBundle` `new-version` mode.** A raw `db.update(scripts)` replaced `content` wholesale from an uploaded file while leaving the acknowledgement columns untouched, so an entry named after an already-approved script inherited that approval for a body no human read. That is exactly the failure the description-set design exists to prevent, reached through a different door. Both non-interactive content-replacing paths (bundle new-version, system-library sync) now clear the acknowledgement; carry-forward stays only where a human is looking at the editor. `5fe6aa6`
- **`fleetFindings/dispatch.ts` — the fourth dispatch path — had zero coverage of the new field.** The existing suite mocked script rows without the key at all, so it was blind by construction. Added three tests, including one pinning that an acknowledgement revoked mid-run stops authorising the in-flight run. `5fe6aa6`
- **A vacuous test.** The "`.` does not cross a newline" mirror test put `HKLM` *before* `Set-ItemProperty`, so it failed to match for the wrong reason and passed identically with dotAll on. Replaced with a discriminating fixture plus an explicit `i` vs `is` assertion. `5fe6aa6`
- **An unguarded parse.** The BASIC-pattern check `slice`d from an unchecked `indexOf`, so renaming `basicPatterns` would have made every `not.toContain` pass vacuously. `5fe6aa6`
- **Overstated UI copy.** `clearedHint` said "This script is allowed to run", which the component cannot know — it only inspects Strict patterns, and a Basic match in the same body still blocks unconditionally without surfacing there. Scoped in all 8 locales. `5fe6aa6`
- **No test of the web wiring.** A broken `watch`/`setValue` key would have left the acknowledgement UI a silent no-op with every backend test still green. Added `ScriptForm` integration tests. `5fe6aa6`
- Suggestions: case-sensitivity test for the vocabulary check, and an agent log line when the payload key is present but decodes to nothing (already fail-closed, but previously indistinguishable from "nobody acknowledged it"). `5fe6aa6`

**Round 2 (focused re-review of the fix) — 1 important**

- Confirmed the fix correct and **complete** — an exhaustive sweep found no missed content-write path. Found one precision bug *in the fix*: `ensureSystemLibraryScripts` takes its update branch on any tracked-field diff, so the clear also fired for a metadata-only sync, silently revoking an acknowledged system script when the body was byte-identical. Keyed to a `contentChanged` flag. `dec7607`

While addressing these I swept every write ingress to `scripts.content` myself. All seven are now resolved, cleared, or pinned as deliberately not-copied — clone, library import, and bundle create each relied on an *omission* from a hand-maintained column list, which is precisely what a later author "completes". `2f2cdb7` `e74d935` `1d33140`

**Tests:** every new guard was proven to **discriminate**, not merely pass — the Go-parity test red against a one-character description drift; both agent decode tests red with their decode line removed; the route, bundle, clone, import, fleet-dispatch and form-wiring tests each red with their fix reverted.

- agent: `go test -race ./internal/executor/... ./internal/heartbeat/... ./internal/userhelper/...` green
- API affected (12 files): 533 passed, 2 skipped
- web `src/components/scripts/`: 266 passed · shared mirror + Go-parity: 16 passed
- real Postgres: `tenant-export-policy` 3/3, `tenantCascade` + `tenantExportErasureRoundtrip` 8/8, `rls-coverage` 102/102
- `tsc --noEmit` (api, web) clean · `pnpm lint` clean · `gofmt`/`go vet` clean

**Status:** review-clean, awaiting maintainer merge. Two judgement calls worth a second opinion are flagged in my handoff: classifying `acknowledged_security_patterns` as `excludedOpen` in the export policy, and the choice to *revoke* rather than carry forward on the two non-interactive content-replacing paths.

